PR #21645, opened 2024-02-23
[bazel] unify cw310_params, cw310/hyper310_jtag_params

nbdd0121:
The functionality is integrated to cw310_params as extra parameters. The intention is that one can choose to enable JTAG without having to clear the bitstream.

For backward compatibility of the rules, *_jtag_params are set to clear the bitstream.

If test_harness is specified, then `--bitstream={bitstream}` is automatically added to test_cmd.

PR #21646, opened 2024-02-23
[bazel] Rework bitstream clearing in tests

pamaury:
At the moment, some tests manually clear the bitstream *before* the tests but really we ought to be doing the opposite which is to clear the bitstream *after* the test if it touches the OTP. This way, such a test can be safely followed by a "normal" tests.

This commit introduces the notion of a post_test_{harness,cmd} that is run after the normal test_harness (on success and failure). The fpga310 parameters are modified so that they take an extra "clear_bitstream" argument that uses this post test command to clear the bitstream if requested.

This should be combined with #21645

### DIFF
--- a/rules/opentitan/fpga_cw310.bzl
+++ b/rules/opentitan/fpga_cw310.bzl
@@ -30,6 +30,17 @@ load("//rules/opentitan:toolchain.bzl", "LOCALTOOLS_TOOLCHAIN")
 _TEST_SCRIPT = """#!/bin/bash
 set -e
 
+function cleanup {{
+  echo "cleanup: {post_test_harness} {post_test_cmd}"
+  {post_test_harness} {post_test_cmd}
+}}
+
+# Bazel will send a SIGTERM when the timeout expires and will
+# wait for a short amount of time (local_termination_grace_seconds)
+# before it kills the process. Therefore the trap will run even
+# on timeout.
+trap cleanup EXIT
+
 TEST_CMD=({test_cmd})
 echo Invoking test: {test_harness} {args} "$@" "${{TEST_CMD[@]}}"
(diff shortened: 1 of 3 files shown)


a-will:
It doesn't seem bad to have the command that occurs afterwards, but I'm not sure we can count on it to always occur. Tests that time out will not run the command, right?

Mostly, we'll want to be careful about our claims of safety, hehe.

pamaury:
Hum, you are right that if bazel stops the test then the command will not run. But if the timeout is due to the test harness then it should run.

nbdd0121:
Apparently bazel will send a SIGTERM, wait for a termination grace period and then send SIGKILL. There's `local_termination_grace_seconds` option to tune it, and it's 15 seconds, which should be sufficient for clearing the bitstream.

I ran a test locally and force it to timeout, and I do indeed see a cleanup message popping up *after* timeout.

rules/opentitan/fpga_cw310.bzl
```diff
@@ -37,6 +37,13 @@ load("//rules/opentitan:toolchain.bzl", "LOCALTOOLS_TOOLCHAIN")
 _TEST_SCRIPT = """#!/bin/bash
 set -e
 
+function cleanup {{
+  echo "cleanup: {post_test_harness} {post_test_cmd}"
+  {post_test_harness} {post_test_cmd}
+}}
+
+trap cleanup EXIT
```
nbdd0121, line 45: We might want to skip this if there's nothing to do for cleanup?

pamaury: We could skip entirely but it would make the code more complex for little gain?

a-will (approved):
LGTM

timothytrippel (approved):

nbdd0121 (approved):

PR #21649, opened 2024-02-23
Bazel jtag improve

pamaury:


rules/opentitan/cc.bzl
```diff
@@ -422,6 +422,19 @@ opentitan_test = rv_rule(
...
             cfg = "exec",
         ),
+        "need_jtag": attr.bool(
+            default = False,
+            doc = "JTAG is required for this test",
+        ),
+        "jtag_data_deps": attr.label_list(
+            allow_files = True,
+            default = [],
+            doc = "Extra data files to depend on when JTAG is to be used.",
```
pamaury, line 432: FIXME: it should say that it's an override

rules/opentitan/cc.bzl
```diff
@@ -422,6 +422,19 @@ opentitan_test = rv_rule(
...
+        ),
+        "jtag_test_cmd": attr.string(
+            default = "",
+            doc = "Extra test command arguments for JTAG.",
```
pamaury, line 436: Same

rules/opentitan/defs.bzl
```diff
@@ -231,6 +231,7 @@ def opentitan_test(
             rsa_key = rsa_key,
             spx_key = spx_key,
             manifest = manifest,
+            need_jtag = getattr(tparam, "need_jtag", False),
```
nbdd0121, line 234: I think this is the part that I am missing?

rules/opentitan/cc.bzl
nbdd0121: Do we need this override at all?

rules/opentitan/defs.bzl
pamaury: Yes, but I think you were also missing the first change in `exec_env.bzl` (adding fields to `_FIELDS` so that they get copied to the exec env provider)

pamaury:
I think the overrides are not strictly speaking necessary but might be nice to have?

PR #21659, opened 2024-02-23
[dv] Make mem_model's compare_byte function less chatty

rswarbrick:
This spits out quite an impressive ream of text, making --verbosity=medium less useful than it might be.

Looking at the history of the file, I think this message got promoted from UVM_HIGH to UVM_MEDIUM as part of commit 39b2692. I suspect this was a debug print that survived by accident.

Demote it again!


hcallahan-lowrisc (approved):

eshapira (approved):
LGTM

PR #21661, opened 2024-02-24
[fpga_cw310_sival_rom_ext] Fix key mismatch

a-will:
Change the generic ROM_EXT target to be signed by the fake prod key, so it may be used in the prod life cycle state.

Change the fpga_cw310_sival_rom_ext execution environment to optionally use the generic ROM_EXT with a sival OTP configuration. A build setting allows it to be built with the real keys, however.

This fixes a key mismatch due to earlier changes.

### DIFF
--- a/hw/top_earlgrey/BUILD
+++ b/hw/top_earlgrey/BUILD
@@ -214,8 +214,14 @@ fpga_cw310(
         "//sw/device/lib/arch:fpga_cw310",
     ],
     otp = "//hw/ip/otp_ctrl/data/earlgrey_a0_skus/sival_bringup:otp_img_prod_manuf_personalized",
-    rom_ext = "//sw/device/silicon_creator/rom_ext/sival:rom_ext_fake_prod_signed_slot_a",
-    rsa_key = {"//sw/device/silicon_creator/rom_ext/keys/fake:rom_ext_prod_private_key_0": "prod_key_0"},
+    rom_ext = select({
+        "//signing:test_keys": "//sw/device/silicon_creator/rom_ext:rom_ext_slot_a",
+        "//conditions:default": "//sw/device/silicon_creator/rom_ext/sival:rom_ext_fake_prod_signed_slot_a",
+    }),
+    rsa_key = select({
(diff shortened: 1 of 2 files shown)


hw/top_earlgrey/BUILD
```diff
@@ -214,8 +218,14 @@ fpga_cw310(
...
     otp = "//hw/ip/otp_ctrl/data/earlgrey_a0_skus/sival_bringup:otp_img_prod_manuf_personalized",
-    rom_ext = "//sw/device/silicon_creator/rom_ext/sival:rom_ext_fake_prod_signed_slot_a",
-    rsa_key = {"//sw/device/silicon_creator/rom_ext/keys/fake:rom_ext_prod_private_key_0": "prod_key_0"},
+    rom_ext = select({
+        "//signing:test_keys": "//sw/device/silicon_creator/rom_ext:rom_ext_slot_a",
+        "//conditions:default": "//sw/device/silicon_creator/rom_ext/sival/binaries:rom_ext_real_prod_signed_slot_a",
+    }),
+    rsa_key = select({
+        "//signing:test_keys": {"//sw/device/silicon_creator/rom_ext/keys/fake:rom_ext_prod_private_key_0": "prod_key_0"},
+        "//conditions:default": {"//sw/device/silicon_creator/rom/keys/real/rsa:keyset": "earlgrey_a0_dev_0"},
```
a-will, line 227: Is this actually the correct key for someone with the sival SKU?

a-will:
I think I still need to fix something up. This isn't working for me with Cloud KMS and the sival SKU.

hw/top_earlgrey/BUILD
a-will, line 227: Oops, copy-paste bug. This is a ROM key, not a ROM_EXT key. Will fix.

a-will:
Took me awhile to understand the labels, but now the Cloud KMS variant for the sival SKU also works.

moidx (approved):
This makes sense to me. Thanks for fixing this @a-will!

PR #21662, opened 2024-02-24
[dice] fix certificate array field endianness

timothytrippel:
Som certificate fields are expected to be big endian formatted in order for the ASN1 to be encoded correctly. This fixes some integer byte arrays that were incorrect, causing certificates not to parse correctly.


cfrantz (approved):

pamaury:
I suspected this problem might arise but I wasn't sure what the endianness of the otbn library was. Good catch!

pamaury (approved):

PR #21544, opened 2024-02-18
[RV_DM]rv_dm_progbuf_read_write_execute_vseq

BasitMehmood:
This commit contains vseq and other realted files and verify the functioanlity of program buffer.

PR #21663, opened 2024-02-24
[RV_DM]rv_dm_base_vseq

BasitMehmood:
This commit add up the request_halt() task in base_vseq for reusability of this task in other vseqs.


rswarbrick (commented):
Thanks, this makes sense.

Could you extend this commit so that we also use the new task in the places where its body is currently inlined?

In case this saves you some searching, I think I made a big list in [this comment](https://github.com/lowRISC/opentitan/pull/21544#discussion_r1500762945) on PR. #21544

Also, it might make sense to expand the title of the commit and to write a commit message. The title could be something like "[rv_dm,dv] Add request_halt task to base vseq and use it".

The idea is that someone looking through a list of commits can then understand/remember what each one was about without having to look at the diff each time.

rswarbrick (approved):
This looks good, thanks.

The failing CI checks are SiVal ROM checks that this can't possibly have broken, so I'm happy to merge.

PR #21665, opened 2024-02-25
[RV_DM]rv_dm_rom_read_access_vseq

BasitMehmood:
This commit contains rom_read_access_vseq which is to verify the read accessability of ROM.


BasitMehmood:
@rswarbrick kindly review it. Thanks

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_rom_read_access_vseq.sv
```diff
@@ -0,0 +1,38 @@
...
+
+  constraint lc_hw_debug_en_c {
+    lc_hw_debug_en == lc_ctrl_pkg::On;
+  }
+  constraint scanmode_c {
+    scanmode == prim_mubi_pkg::MuBi4False;
+  }
+  task body();
+    uvm_reg_data_t rdata_o;
+    uvm_reg_data_t expected_data[40] = {'h0180006f, 'h00000013, 'h0840006f, 'h00000013,
```
rswarbrick, line 19: Couldn't we do some sort of backdoor access to figure out what we're expecting? I'm thinking `uvm_hdl_read` with paths like `tb.dut.u_dm_top.i_dm_mem.gen_rom_snd_scratch.i_debug_rom.mem[i]`. That should avoid us having to copy the contents of a vendored module.

That way, we won't be checking that the ROM works (it will!), but we are still checking that the register-based access to the ROM works correctly.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_rom_read_access_vseq.sv
```diff
@@ -0,0 +1,38 @@
...
+                                        'h00147413, 'h02041c63, 'hf1402473, 'h00a40433,
+                                        'h40044403, 'h00247413, 'hfa0418e3, 'hfd5ff06f,
+                                        'h00000517, 'h00c55513, 'h00c51513, 'h10052c23,
+                                        'h7b302573, 'h7b202473, 'h00100073, 'h10052423,
+                                        'h7b302573, 'h7b202473, 'ha79ff06f, 'hf1402473,
+                                        'h10852823, 'h7b302573, 'h7b202473, 'h7b200073};
+    repeat (2) begin
+      for (int i = 0; i < 40; i++) begin
+        mem_rd(.ptr(tl_mem_ral.rom), .offset('h00+i), .data(rdata_o));
+        `uvm_info(`gfn, $sformatf(" \n Access ROM Addr 0x%0h", rdata_o), UVM_LOW)
```
rswarbrick, line 32: Nit: I'd probably put these lines the other way around. That way, if something explodes when accessing a particular address, you'll see the message before the simulation stops...

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_rom_read_access_vseq.sv
```diff
@@ -0,0 +1,36 @@
...
+  }
+  constraint scanmode_c {
+    scanmode == prim_mubi_pkg::MuBi4False;
+  }
+  task body();
+    string         path;
+    bit [31:0]     lower_mem_val;
+    bit [31:0]     upper_mem_val;
+    uvm_reg_data_t mem_val;
+    uvm_reg_data_t rdata_o;
```
rswarbrick, line 22: This name looks a bit strange to me: it sounds like the name of an output port. Probably something like `rom_data` would work?

rswarbrick (commented):
Thanks for adding the backdoor reads. This looks much cleaner now!

As a general note, the commit message seems to have got a little bit mangled here. Is this the commit message of the final fixup commit that you added?

BasitMehmood:
> Thanks for adding the backdoor reads. This looks much cleaner now!
> 
> As a general note, the commit message seems to have got a little bit mangled here. Is this the commit message of the final fixup commit that you added?

I have updated the commit message in latest commit.

rswarbrick (approved):
Thanks for the tidy-ups.

rswarbrick:
Merging without waiting for CI: the only difference from a version that passed is a change of variable name. The previous version passed everything except for CW310 tests (which this code can't have broken), so I'm happy for this to go in.

PR #21667, opened 2024-02-26
[hw,spi_host,i2c,soc_proxy] Fix CDC issue for lsio_triggers

Razer6:
This PR flops the LSIO triggers at the source IP with their respective peripheral clock and then uses a synchronizer cell at the soc_proxy IP with the fast SoC clock to fix a CDC issue between both domains.

The LSIO trigger of the I2C IP is already flopped.

### DIFF
--- a/hw/ip/spi_host/rtl/spi_host.sv
+++ b/hw/ip/spi_host/rtl/spi_host.sv
@@ -589,7 +589,14 @@ module spi_host
     end
   end
 
-  assign lsio_trigger_o = tx_wm_q | rx_wm_q;
+  // Flop trigger signal to avoid glitches on the output
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      lsio_trigger_o <= 1'b0;
+    end else begin
+      lsio_trigger_o <= tx_wm_q | rx_wm_q;
+    end
+  end
 
   prim_intr_hw #(.Width(1)) intr_hw_spi_event (
     .clk_i,
--- a/hw/ip/uart/rtl/uart_core.sv
+++ b/hw/ip/uart/rtl/uart_core.sv
@@ -321,7 +321,14 @@ module uart_core (
     endcase
   end
 
-  assign lsio_trigger_o = tx_watermark_prev_q | rx_watermark_prev_q;
+  // Flop trigger signal to avoid glitches on the output
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      lsio_trigger_o <= 1'b0;
+    end else begin
+      lsio_trigger_o <= tx_watermark_prev_q | rx_watermark_prev_q;
+    end
+  end
 
   assign event_tx_watermark = tx_watermark_d & ~tx_watermark_prev_q;
 
--- a/hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
+++ b/hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
@@ -305,12 +305,42 @@ module soc_proxy
     .filter_o (rst_req_external_o)
   );
 
+  // CDC sync of LSIO trigger signals between the peripheral and the high-speed clock domain
+  logic uart_lsio_trigger_sync, spi_host_lsio_trigger_sync, i2c_lsio_trigger_sync;
+
+  prim_flop_2sync #(
+    .Width(1)
+  ) u_uart_lsio_trigger_sync (
+    .clk_i,
+    .rst_ni,
+    .d_i      (uart_lsio_trigger_i),
+    .q_o      (uart_lsio_trigger_sync)
+  );
+
+  prim_flop_2sync #(
+    .Width(1)
+  ) u_spi_host_lsio_trigger_sync (
+    .clk_i,
+    .rst_ni,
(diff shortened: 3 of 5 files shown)


hw/ip/spi_host/rtl/spi_host.sv
```diff
@@ -589,7 +589,14 @@ module spi_host
...
 
-  assign lsio_trigger_o = tx_wm_q | rx_wm_q;
+  // Flop trigger signal for CDC to SOC clock domain
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      lsio_trigger_o <= 1'b0;
+    end else begin
+      lsio_trigger_o <= tx_wm_q | rx_wm_q;
+    end
+  end
```
andreaskurth, line 599: `tx_wm_q` and `rx_wm_q` are already in the `clk_i`/`rst_ni` domain; do we really need to flop them again after they have propagated through a single OR gate? I mean, this doesn't introduce glitches and we don't care if the propagation delays out of `tx_wm_q` and `rx_wm_q` are different, right? Or am I missing sth?

hw/ip/uart/rtl/uart_core.sv
```diff
@@ -321,7 +321,14 @@ module uart_core (
...
 
-  assign lsio_trigger_o = tx_watermark_prev_q | rx_watermark_prev_q;
+  // Flop trigger signal for CDC to SOC clock domain
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      lsio_trigger_o <= 1'b0;
+    end else begin
+      lsio_trigger_o <= tx_watermark_prev_q | rx_watermark_prev_q;
+    end
+  end
```
andreaskurth, line 331: Similar question here: `tx_watermark_prev_q` and `tx_watermark_prev_q` are already in the `clk_i`/`rst_ni` domain; why flop again?

hw/ip/spi_host/rtl/spi_host.sv
andreaskurth: Also, I wouldn't use "SOC clock domain" because `clk_i`/`rst_ni` are part of the peripheral clock domain

hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
```diff
@@ -305,12 +305,42 @@ module soc_proxy
     .filter_o (rst_req_external_o)
   );
 
+  // CDC sync of LSIO trigger signals between the peripheral and the SoC clock domain
```
andreaskurth, line 308: "SoC clock domain" isn't appropriate here, I think, because we don't call OpenTitan a SoC and its clock domain is / may be different from the SoC it is integrated into. How about "high-speed domain" as in the block diagram? Or just "clock domain of this module"?

hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
```diff
@@ -305,12 +305,42 @@ module soc_proxy
...
+    .Width(1)
+  ) u_i2c_lsio_trigger_sync (
+    .clk_i,
+    .rst_ni,
+    .d_i    (i2c_lsio_trigger_i),
+    .q_o    (i2c_lsio_trigger_sync)
+  );
+
   // Collate LSIO trigger inputs into signal for DMA
   assign dma_lsio_trigger_o = {
```
andreaskurth, line 339: As all components of this signal are now synchronous, I suggest adding a ", synchronous" to the signal's description in the Hjson file:
https://github.com/lowRISC/opentitan/blob/9fb9dbc43f97b9d0c3d7c9d81252998f0e6d21b9/hw/top_darjeeling/ip/soc_proxy/data/soc_proxy.hjson#L223

andreaskurth (commented):
Thx for this change, @Razer6. I agree that the trigger signals should be `flop_2sync`ed in `soc_proxy`, but I'm not convinced about flopping them (again) in the source modules. Could you please check / explain that?

hw/ip/spi_host/rtl/spi_host.sv
Razer6: Today I was told that even a single OR can glitch, thus the flop.

msfschaffner (approved):

msfschaffner:
CHANGE AUTHORIZED: hw/ip/uart/rtl/uart_core.sv
CHANGE AUTHORIZED: hw/top_darjeeling/ip/soc_proxy/data/soc_proxy.hjson
CHANGE AUTHORIZED: hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv

hw/ip/spi_host/rtl/spi_host.sv
andreaskurth: Okay, it can glitch 1 -> 0 -> 1, and these are watermark signals, hence level-encoded triggers, so we do care about it. I agree with adding a flop

andreaskurth (approved):

andreaskurth:
CHANGE AUTHORIZED: hw/ip/uart/rtl/uart_core.sv
CHANGE AUTHORIZED: hw/top_darjeeling/ip/soc_proxy/data/soc_proxy.hjson
CHANGE AUTHORIZED: hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv

Issue #21666, opened 2024-02-26
[hmac/rtl] Fix suffixes of `hmac_core`'s port names


andreaskurth:
`hmac_core`'s ports currently don't have the `_i` or `_o` suffixes [mandated by our style guide:](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#suffixes-for-signals-and-types):

https://github.com/lowRISC/opentitan/blob/49a27e136c7b79a3c738dd77cdae3764e19210c0/hw/ip/hmac/rtl/hmac_core.sv#L7-L44

andreaskurth:
Note that changes to signal names may have implications on side-band information such as waivers and tool configuration files. It may thus be desirable to postpone this cleanup until after the current release.

rswarbrick:
I did a bit of grepping to see whether any of the ports appear in exclusion files. I'm sad to say that they definitely do! I've appended the hits that I could find.

I'd suggest splitting the change into multiple commits (maybe one port per commit). And each commit could tweak the exclusion files with the obvious rename. Practically speaking, there are only two exclusion files (`hmac_cov_excl.el` and `hmac_unr_excl.el`) and they aren't ridiculously huge, so a manual edit might just work.

- wipe_v verilator waiver in hmac.vlt
- sha_rvalid and sha_rready mentioned in hmac_cov_excl.el
  + This is a waiver about hmac_core.sv, line 113. Maybe worth hacking
    on the exclusion (on the off chance that it still works), but
    hmac_cor.sv has changed anyway. There's a good chance we'll need
    to re-generate the exclusion list either way.
- fifo_rvalid, fifo_rready, fifo_rdata
  + These get mentioned in hmac_unr_excl. Lines 110, 132, 284
- fifo_wsel
  + Gets mentioned in hmac_unr_excl, lines 106, 108, 128, 130
- fifo_wvalid
  + Gets mentioned in hmac_unr_excl, lines 106, 108, 122, 132
- fifo_wdata_sel and fifo_wready
  + Get mentioned in hmac_cov_excl, line 24
- idle
  + Goodness knows! Probably not worth grepping for...

As a more general comment, we've made a couple of other changes to `hmac_core.sv` since those exclusion files were generated last January. There's a good chance they'll need tweaking either way, so I'd be inclined to just do the rename now.

gdessouky:
[...]

PR #21670, opened 2024-02-26
[hmac,verilator] Remove lint waivers for nonexistent file

rswarbrick:
sha2_pad was removed from hmac with 2290dcc11, because the code now lives in prim_sha2_pad. I suspect we'll want to put in a lint waiver there at some point too, but there's no real advantage to keeping this stale one.

I found this because I was searching through the repository for the various ports of `hmac` when thinking about #21666 (and whether we could "just do it"). Here's a waiver that definitely wouldn't need renaming...


andreaskurth (approved):
LGTM, thanks

PR #21238, opened 2024-02-07
[usbdev] Support Si device type in USBDEV tests

alees24:
Minor modifications to support use of kDeviceSilicon as a build target, and BUILD file changes for SiVal.

PR #21145, opened 2024-02-01
[usbdev] usbdev_config_host test

alees24:
This PR creates a simple test of USBDEV detection, identification and configuration by the host/DPI model, indicating that bidirectional communication with the host is working.

Control Transfers from the host are received and processed, identifying the USB device as a single 'simpleserial' device to the host, and a Linux-based host will usually present the device as '/dev/ttyUSBx'.

No further communication with the host occurs during this test and no additional software beyond the driver stack is required on the host side for this test to pass. The test has been checked on the CW310 and in chip simulation.

See [#19966](https://github.com/lowRISC/opentitan/issues/19966)

PR #19249, opened 2023-07-20
Initial USB logging code

alees24:
Prototype code for logging over USB. This PR details an extension to the usbdev testing code which supports conventional textual logging as if over a UART, but because usbdev operates at 12Mbps it is able to offer much higher throughput and - perhaps more importantly - lower impact CPU performance and latency in servicing other events/devices. The UART offers only 115.2kbaud by contrast. Redirection of stdout is achieved using base_set_stdout() and then LOG_x() etc may be used unmodified.

Also included is a top-level test that exercises a number of concurrent serial streams (presented as ttyUSBx devices on a Linux host), allowing us to model the Google use case of multiple UART streams simultaneously.

Any feedback and suggestions welcome.

UPDATE: Thanks all for the feedback and suggestions; I've spent a little time addressing all points whilst implementing the proposed change to dynamically allocate buffers among the active streams. That reduces the memory footprint, and hopefully makes the code easier to understand since buffers are now separated from logging streams and hold their own state.

Added instructions to the top of usbdev_logging_test for anyone wishing to run the test:
```
bazel test --test_output=streamed //sw/device/tests:usbdev_logging_test_fpga_cw310_test_rom
```
Test will pass with 4 'cat /dev/ttyUSBn' commands on the host-side serial connections established by the test.

@a-will, @engdoreis, @luismarques FYI.

PR #21671, opened 2024-02-26
Cherry-pick USB sival tests to earlgrey_es_sival

jwnrt:
This PR cherry picks the following from master to earlgrey_es_sival:

* https://github.com/lowRISC/opentitan/pull/21238
* https://github.com/lowRISC/opentitan/pull/21145
* https://github.com/lowRISC/opentitan/pull/19249
* https://github.com/lowRISC/opentitan/pull/21363

I see that we've also had lots of changes to the USB device DIFs because the RTL has changed. I haven't cherry-picked those over, so unsure whether every single one of these will be okay just yet.


pamaury (approved):
Looks good to me. Would be good to have @alees24 double check

PR #21675, opened 2024-02-26
[cryptotest] Add NIST AES-KW testvector parser

sadmansakib-git:
This PR pulls in the NIST testvectors for AES-KW and includes the associated parser along with the JSON schema.

### DIFF
--- a/sw/host/cryptotest/testvectors/data/BUILD
+++ b/sw/host/cryptotest/testvectors/data/BUILD
@@ -157,3 +157,102 @@ run_binary(
         ("kmac256_no_customization_test.json", "256", "wycheproof_kmac_256"),
     ]
 ]
+
+[
+    run_binary(
+        name = "nist_cavp_aes_{}_{}_json".format(
+            function_name.lower(),
+            keylen_with_tmode,
+        ),
+        srcs = [
+            "@nist_cavp_aes_kw_sp_800_38f//:{}_{}.txt".format(
+                function_name,
+                keylen_with_tmode,
+            ),
+            "//sw/host/cryptotest/testvectors/data/schemas:aes_kw_schema.json",
+        ],
+        outs = [":nist_cavp_aes_{}_{}.json".format(
+            function_name.lower(),
+            keylen_with_tmode,
+        )],
+        args = [
+            "--src",
+            "$(location @nist_cavp_aes_kw_sp_800_38f//:{}_{}.txt)".format(function_name, keylen_with_tmode),
+            "--dst",
+            "$(location :nist_cavp_aes_{}_{}.json)".format(
+                function_name.lower(),
+                keylen_with_tmode,
+            ),
+            "--operation",
+            "{}".format(operation),
+            "--key_len",
+            "{}".format(key_len),
+            "--padding",
+            "{}".format(padding),
+            "--transformation_mode",
+            "{}".format(transformation_mode),
+            "--schema",
+            "$(location //sw/host/cryptotest/testvectors/data/schemas:aes_kw_schema.json)",
+        ],
+        tool = "//sw/host/cryptotest/testvectors/parsers:nist_cavp_aes_kw_parser",
+    )
+    for function_name, operation, padding in [
+        ("KW_AD", "decrypt", False),
+        ("KW_AE", "encrypt", False),
+        ("KWP_AD", "decrypt", True),
+        ("KWP_AE", "encrypt", True),
+    ]
+    for keylen_with_tmode, key_len, transformation_mode in [
+        ("128_inv", "128", "inverse"),
+        ("128", "128", "null"),
+        ("192_inv", "192", "inverse"),
+        ("192", "192", "null"),
+        ("256_inv", "256", "inverse"),
+        ("256", "256", "null"),
+    ]
+]
+
+[
+    run_binary(
+        name = "nist_cavp_aes_{}_{}_json".format(
+            function_name.lower(),
+            key_len,
+        ),
+        srcs = [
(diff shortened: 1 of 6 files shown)


sw/host/cryptotest/testvectors/data/schemas/aes_kw_schema.json
```diff
@@ -0,0 +1,67 @@
...
+        "maximum": 255
+      }
+    }
+  },
+  "type": "array",
+  "items": {
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+      "test_case_id": {
```
RyanTorok, line 21: Along with the test case ID, I usually include a `vendor` field, with one of the values `[nist, wycheproof, manual, ...]`.

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_kw_parser.py
```diff
@@ -0,0 +1,99 @@
...
+                tmode = "forward" if args.operation == "decrypt" else "inverse"
+            testcase = {
+                "test_case_id": test_case_id,
+                "operation": args.operation.lower(),
+                "transformation_mode": tmode,
+                "padding": args.padding,
+                "key_len": args.key_len,
+                "key": str_to_byte_array(test_vec["K"]),
+                "ciphertext": str_to_byte_array(test_vec["C"]),
+                "plaintext": str_to_byte_array(test_vec["P"]) if "P" in test_vec else [],
```
RyanTorok, line 35: Just to let you know, when you include an optional parameter like this, you'll need to add `#[serde(deafult)]` to the variable in the test case type in the Rust test harness once you get there, otherwise the deserialization will fail for test cases that lack this parameter.

sw/host/cryptotest/testvectors/data/BUILD
```diff
@@ -128,3 +128,100 @@ genrule(
...
+            ),
+            "//sw/host/cryptotest/testvectors/data/schemas:aes_kw_schema.json",
+        ],
+        outs = [":nist_cavp_aes_{}_{}.json".format(
+            function_name.lower(),
+            keylen_with_tmode,
+        )],
+        args = [
+            "--src",
+            "$(location @nist_cavp_aes_kw_sp_800_38f//:kwtestvectors/{}_{}.txt)".format(function_name, keylen_with_tmode),
```
RyanTorok, line 151: If all the test vectors you need are in this directory, it's cleaner to add `strip_prefix = "kwtestvectors"` in the `http_archive` pull for the test vector archive, and then you can reference all the test vector files like they are in the root directory of the bazel collection.

third_party/nist_cavp_testvectors/repos.bzl
```diff
@@ -31,3 +31,9 @@ def nist_cavp_repos():
         sha256 = "debfebc3157b3ceea002b84ca38476420389a3bf7e97dc5f53ea4689a16de4c7",
         url = "https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/sha3/shakebytetestvectors.zip",
     )
+    http_archive(
+        name = "nist_cavp_aes_kw_sp_800_38f",
+        build_file = Label("//third_party/nist_cavp_testvectors:BUILD.nist_cavp_common.bazel"),
+        sha256 = "04a4a82e4de65bca505125295003f9c75a5a815afda046dc83661b8b580dfdf3",
```
RyanTorok, line 37: See comment above about `strip_prefix`.

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_kw_parser.py
sadmansakib-git: Is the code selection part for your comment correct? I can't see any optional parameter in this code selection. Or you meant this:

```python
    parser.add_argument(
        "--padding",
        type = bool,
        default = False,
        help = "Existence of padding."
    )
    ```

Comment (RyanTorok) [reply sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_kw_parser.py:35]: Never mind, I see you're passing an array as the default. I was referring to if you did not include `plaintext` at all in the JSON test vector when it does not exist in the source test vector.

**Review by RyanTorok — APPROVED** [review]

File: sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_kw_parser.py [hunk 5]
```diff
@@ -0,0 +1,99 @@
...
+
+from cryptotest_util import parse_rsp, str_to_byte_array
+
+
+def parse_testcases(args) -> None:
+    raw_testcases = parse_rsp(args.src)
+    testcases = list()
+    test_case_id = 1
+    for test_key in raw_testcases.keys():
+        for test_vec in raw_testcases[test_key]:
```
RyanTorok, line 23: We'll have to change this code to remove the double loop to merge #21728, but that shouldn't block this PR.

sadmansakib-git: Locally I have already verified it by removing the nested loop to follow the new parsing logic. It works fine. I will create another PR to change this.

sadmansakib-git:
@milesdai , rebased and checked this too. Works fine!

milesdai (approved):
LGTM! Thanks! Looks like there's a rebase conflict with master. Can you rebase again and then we can get this merged?

milesdai:
Merging since the broken lint was introduced on master and has been fixed. All other tests are passing.

PR #21680, opened 2024-02-26
[cryptotest] Test vectors for SPHINCS+

RyanTorok:
1. Adds an `http_archive` pull rule for the test vectors for SPHINCS+, removing the need to vendor the test vectors in `/sw/device/tests/crypto/testvectors/sphincsplus_kat`.
2. Adds a python parser and JSON schema to integrate the SPHINCS+ test vectors with the rest of the cryptotest framework, allowing automated testing of the implementation in `/sw/device/silicon_creator/lib/sigverify/sphincsplus`

### DIFF
--- a/sw/host/cryptotest/testvectors/data/BUILD
+++ b/sw/host/cryptotest/testvectors/data/BUILD
@@ -396,3 +396,23 @@ run_binary(
         ("verify", "Ver", "rsp"),
     ]
 ]
+
+run_binary(
+    name = "rsp_sphincsplus_shake256_128s_simple_json",
+    srcs = [
+        "//sw/host/cryptotest/testvectors/data/schemas:sphincsplus_schema.json",
+        "@sphincsplus_kat//:sphincs-shake256-128s-simple/PQCsignKAT_64.rsp",
+    ],
+    outs = [":rsp_sphincsplus_shake256_128s_simple.json"],
+    args = [
+        "--src",
+        "$(location @sphincsplus_kat//:sphincs-shake256-128s-simple/PQCsignKAT_64.rsp)",
+        "--dst",
+        "$(location :rsp_sphincsplus_shake256_128s_simple.json)",
+        "--schema",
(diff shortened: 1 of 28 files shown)


jadephilipoom (commented):
I think you might need to leave the old test vectors in place for now, since they currently run in CI and we probably want to keep them running there to ensure the sigverify implementation is working.

sw/host/cryptotest/testvectors/data/BUILD
```diff
@@ -131,3 +131,31 @@ run_binary(
...
+            "--hash_alg",
+            hash_alg,
+        ],
+        tool = "//sw/host/cryptotest/testvectors/parsers:rsp_sphincsplus_parser",
+    )
+    for src_name, cryptotest_name, hash_alg in [
+        ("sphincs-sha256-128s-simple/PQCsignKAT_64.rsp", "rsp_sphincsplus_sha256_128s_simple", "sha-256"),
+        ("sphincs-sha256-128s-robust/PQCsignKAT_64.rsp", "rsp_sphincsplus_sha256_128s_robust", "sha-256"),
+        ("sphincs-shake256-128s-simple/PQCsignKAT_64.rsp", "rsp_sphincsplus_shake256_128s_simple", "shake-256"),
+        ("sphincs-shake256-128s-robust/PQCsignKAT_64.rsp", "rsp_sphincsplus_shake256_128s_robust", "shake-256"),
```
jadephilipoom, line 159: For what it's worth, we'll probably never need `robust`; NIST already announced that it won't be standardized and it has a ~2x speed penalty. We also currently do not use the `sha256` variant either, so only `shake256-128s-simple`.

jadephilipoom (approved):

sw/host/cryptotest/testvectors/data/BUILD
RyanTorok: Removing these make the build rule much simpler, thanks!

RyanTorok:
Closing in favor or #21976, which passed CI and includes all of the commits in this PR.

PR #21113, opened 2024-01-30
[cryptotest] Add ECDSA host-side test harness

RyanTorok:
This PR adds the host-side Rust test harness for the ECDSA `Verify` operation tests and the associated uJSON C header file. The device-side firmware will be included in #20853.

~Dependent on #21112 and #21114~

PR #20853, opened 2024-01-16
[cryptotest] add device-side test firmware for ECDSA `Verify` operation

RyanTorok:
This PR adds the device-side firmware handler and host-side Rust driver for performing the ECDSA cryptotests using the test vectors added in #20793 . 

The tests in this PR require the patch in #21115 in order to pass.

~Dependent on #21113~

PR #21681, opened 2024-02-26
[cryptotest] Test harness for SPHINCS+ verify operation w/ SHA-2

RyanTorok:
This PR is a follow-up to #21680. Adds a test harness for SPHINCS+ verify operation tests w/ SHAKE-256 in the style of the test harnesses for the algorithms in cryptolib. Structurally, the test harness has a lot in common with the ECDSA harness in #21113 and #20853, except that there is no elliptic curve parameter.

[Resolved] The dependency on #22953 exists because this PR switches the test vectors to use the `consistent-basew` version of SPHINCS+.

~Dependent on #21680~
~Dependent on #22953~

### DIFF
--- a/sw/device/tests/crypto/cryptotest/BUILD
+++ b/sw/device/tests/crypto/cryptotest/BUILD
@@ -365,6 +365,25 @@ cryptotest(
     test_vectors = KMAC_TESTVECTOR_TARGETS,
 )
 
+SPHINCSPLUS_TESTVECTOR_TARGETS = [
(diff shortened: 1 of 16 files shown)


RyanTorok:
Updated this PR to use the test vectors for SPX w/ SHA-2 instead of SHAKE, since this was changed in ROM in #23732 .

jadephilipoom (approved):

cfrantz (approved):

PR #21684, opened 2024-02-26
OTBN ISA Test

HU90m:
Test re-using the `smoke_test.s` sequence to check the OTBN instructions are working as expected. Loads from the `RND` and `KEY_*` registers have been removed so the sequence is deterministic. These loads are OK to remove because the load instruction is used in the OTBN randomness test, which is also run in SiVAl

Covers: https://github.com/lowRISC/opentitan/issues/20114

### DIFF
--- a/hw/ip/otbn/dv/smoke/smoke_test.s
+++ b/hw/ip/otbn/dv/smoke/smoke_test.s
@@ -5,7 +5,7 @@
 # OTBN Smoke test, runs various instructions which are expected to produce the
 # final register state see in smoke_expected.txt
 
-.section .text
+.section .text.start
 
 # x2 = 0xd0beb513
 lui x2, 0xd0beb
@@ -125,13 +125,18 @@ csrrs x23, mod5, x0
 # w1 = mod = 0x78fccc06_2228e9d6_89c9b54f_887cf14e_c79af825_69be586e_9866bb3b_53769ada
 bn.wsrr w1, 0x0 /* MOD */
 
-# Request an RND value with a write to CSR RND_PREFETCH
-csrrw x0, rnd_prefetch, x0
-
-# sim environment provides a fixed value for RND (in other environment RND isn't
-# fixed so this test will have a different final state)
 # w2 = rnd = 0xAAAAAAAA_99999999_AAAAAAAA_99999999_AAAAAAAA_99999999_AAAAAAAA_99999999
-bn.wsrr w2, 0x1 /* RND */
+.ifdef deterministic
+  la x25, rnd_value
+  li x26, 2
+  bn.lid x26, 0(x25)
+.else
+  # Request an RND value with a write to CSR RND_PREFETCH
+  csrrw x0, rnd_prefetch, x0
+  # sim environment provides a fixed value for RND (in other environment RND isn't
+  # fixed so this test will have a different final state)
+  bn.wsrr w2, 0x1 /* RND */
+.endif
 
 # w3 = w1 + w2 = 0x23a776b0_bbc28370_34745ffa_22168ae8_7245a2d0_0357f208_431165e5_ed103473
 bn.add w3, w1, w2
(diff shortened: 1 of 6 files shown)


hw/ip/otbn/dv/smoke/smoke_test.s
```diff
@@ -5,7 +5,7 @@
 # OTBN Smoke test, runs various instructions which are expected to produce the
 # final register state see in smoke_expected.txt
 
-.section .text
+.section .text.start
```
rswarbrick, line 8: I don't think I quite understand this change. The existing `.section` lines plonked everything into `.text`. Since it's a single compilation unit, this first instruction ended up at the location that you're getting from `.text.start`.

Why does this need to change? And if it does need to change, is there any reason to add the `.text` bit lower down? I'd have thought that everything could go into `.text.start`...

hw/ip/otbn/dv/smoke/smoke_test.s
```diff
@@ -125,13 +128,18 @@ csrrs x23, mod5, x0
...
 bn.wsrr w1, 0x0 /* MOD */
 
-# Request an RND value with a write to CSR RND_PREFETCH
-csrrw x0, rnd_prefetch, x0
-
-# sim environment provides a fixed value for RND (in other environment RND isn't
-# fixed so this test will have a different final state)
 # w2 = rnd = 0xAAAAAAAA_99999999_AAAAAAAA_99999999_AAAAAAAA_99999999_AAAAAAAA_99999999
-bn.wsrr w2, 0x1 /* RND */
+.ifdef deterministic
```
rswarbrick, line 132: I suggest you add a `nop` to the "else branch". That way, the instruction addresses will be unchanged by the value of `deterministic`.

Similarly, I'd suggest sticking a bunch of `nop`s on the other side for the stuff below where there's `.ifnotdef deterministic`.

hw/ip/otbn/dv/smoke/smoke_test.s
```diff
@@ -356,12 +367,85 @@ test_fn_2:
...
+
+.global gpr_state
+.balign 32
+gpr_state:
+  .zero (32 / 4) * 30 # not including x0 and x1
+
+.global wdr_state
+.balign 32
+wdr_state:
+  .zero (256 / 4) * 32
```
rswarbrick, line 451: Can this stuff go in an `ifdef deterministic` block? I'm just thinking that a `rnd_value` in the image is going to be a bit confusing if we are actually using RND! :-)

sw/device/tests/otbn_isa_test.c
```diff
@@ -0,0 +1,144 @@
...
+OTBN_DECLARE_APP_SYMBOLS(smoke_test);
+OTBN_DECLARE_SYMBOL_ADDR(smoke_test, gpr_state);
+OTBN_DECLARE_SYMBOL_ADDR(smoke_test, wdr_state);
+
+static const otbn_app_t kAppSmokeTest = OTBN_APP_T_INIT(smoke_test);
+static const otbn_addr_t kGprState = OTBN_ADDR_T_INIT(smoke_test, gpr_state);
+static const otbn_addr_t kWdrState = OTBN_ADDR_T_INIT(smoke_test, wdr_state);
+
+// These expected values are taken from
+// `hw/ip/otbn/dv/smoke/smoke_expected.txt`.
```
rswarbrick, line 34: I'd suggest expanding this comment to make it clear that we're talking about both GPRs and WDRs.

sw/device/tests/otbn_isa_test.c
```diff
@@ -0,0 +1,144 @@
...
+#include "sw/device/lib/testing/otbn_testutils.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+/**
+ * This test runs every instruction in OTBN's ISA and checks the result.
+ *
+ * It reuses the otbn's `smoke_test.s`,
```
jadephilipoom, line 15: nit: Coming to this without DV context, I had to do a little digging to figure out where `smoke_test.s` originally came from. Maybe:
```suggestion
 * It reuses the OTBN simulator's `smoke_test.s`,
```

sw/device/tests/otbn_isa_test.c
```diff
@@ -0,0 +1,144 @@
...
+            0x23631fe5, 0xbbb1704f, 0xd25666ac},
+    [29] = {0xeb0953c2, 0xe0654fef, 0x63388709, 0x5763bcdf, 0x26628bdb,
+            0x64341d3c, 0x9f24f0c1, 0x4f0d4b81},
+    [30] = {0x68ba2fa1, 0xb55098e0, 0x4efa2ec9, 0xaee49292, 0xab123192,
+            0xffa3d88b, 0xe9ee7ac7, 0x2167f87d},
+    [31] = {0x0f09b7c8, 0x25769434, 0x6978ad1b, 0x67a8c221, 0x5466a52c,
+            0x73880075, 0xf9dbff5e, 0x37adadae},
+};
+
+bool test_main(void) {
```
jadephilipoom, line 111: You could also maybe check the instruction count and error bit registers with `dif_otbn_get_insn_cnt` and `dif_otbn_get_err_bits`.

jadephilipoom (approved):

hw/ip/otbn/dv/smoke/smoke_test.s
HU90m: The `.text.start` is needed because `.text` is not a GC root when using `hw/ip/otbn/data/otbn.ld.tpl`. This wasn't needed before because the linking in the DV test doesn't make use of garbage collection.

Yes the second is `.text` is a little superfluous; I'll remove it.

sw/device/tests/otbn_isa_test.c
HU90m: I've added a instruction count check. The error bits should already be checked by `otbn_testutils_wait_for_done`.

hw/ip/otbn/dv/smoke/smoke_test.s
HU90m: I don't think any `nop`s are needed. The address changes was due to setting a `.text` section.

hw/ip/otbn/dv/smoke/smoke_test.s
rswarbrick: Thanks for explaining. This looks sensible to me.

rswarbrick (approved):

Issue #20947, opened 2024-01-24
[entropy_src] Restarting physical noise source through rng_en


johannheyszl:
### Description

We will restart the physical noise source for cases of a persistent error. This ability also helps with SP 800-90B restart testing to simplify test harnesses. An[ rng_en_i signal](https://github.com/lowRISC/opentitan/tree/f83506bbdbeffbf02110f8a7197caad64b9d55ee/hw/top_earlgrey/ip/ast#random-number-generator) into AST is driving this off/on reset, and originates in [entropy_src.](https://github.com/lowRISC/opentitan/blob/0a0d043d44fd64a7bf9979634d60f9e833247f30/hw/ip/entropy_src/rtl/entropy_src_core.sv#L744-L746)

We need to make sure this rng_en is deasserted carefully and in sync with the entropy_src to maintain FIPS-compatibility. Startup testing shall be repeated upon each restart for instance. 
- For example, the signal shall not be deasserted upon the FIFO full event, even though this is expected to be very rare or ‘should never happen’ (@vogelpi to pls add context here). Beware that for startup testing we might need to fill the FIFO up to capacity and certainly do not want a deassert of the signal as a result.
- Also check other conditions for deassertion.

Issue created after discussion b/w: @moidx @johannheyszl VadimS @vogelpi @zi-v @h-filali

johannheyszl:
cc @msfschaffner @andreaskurth

vogelpi:
We've discussed this in the Security WG meeting today. Even thought the 2 x 4-bit input FIFO should never become full during regular operation we should probably change the design to:
1. Notify software if this FIFO or any of the following ones get full (except for the Observe FIFO for which this is already correctly handled). And, because software needs to know if entropy is dropped.
2. Not lower the `rng_enable_o` signal if those FIFOs get full because the AST should not be disabled/enabled because of that.

PR #21685, opened 2024-02-26
[entropy_src] Remove line diabling TRNG when esrng FIFO is full

h-filali:
This PR removes the line that disables the noise source whenever the esrng FIFO is full.
I added assertions to see whether the esrng, esbit and postht FIFO were pushed to when full and ran tests both with and without this change.
Both with and without this change, the assertions showed that there are some pushes to the postht FIFO when it is full but it seems like this change doesn't have an effect on any of the three FIFOs mentioned above.
This PR does not implement the notifications to the SW for the instances where entropy is dropped because we still need to determine when and for which FIFOs it is okay to drop entropy.

Closes [#20947](https://github.com/lowRISC/opentitan/issues/20947)


hw/ip/entropy_src/rtl/entropy_src_core.sv
```diff
@@ -2921,6 +2921,7 @@ module entropy_src_core import entropy_src_pkg::*; #(
   assign unused_sha3_state = (|sha3_state[0][sha3_pkg::StateW-1:SeedLen]);
   assign unused_entropy_data = (|reg2hw.entropy_data.q);
   assign unused_fw_ov_rd_data = (|reg2hw.fw_ov_rd_data.q);
+  assign unused_sfifo_esrng_not_full = (|sfifo_esrng_not_full);
```
vogelpi, line 2924: Nit: `sfifo_esrng_not_full` is already a single-bit signal, there is no point in ORing all bits together here. :-)

vogelpi (approved):
LGTM, thanks @h-filali !

Thanks for reporting the other issue and for doing this test in the first place. We have to analyze the different FIFOs separately and decide what to do for each of them.

hw/ip/entropy_src/rtl/entropy_src_core.sv
h-filali: Oh yes you're right.

PR #21693, opened 2024-02-26
[otp] Increase OWNER_SW_CFG_ROM_ALERT_CLASSIFICATION by 4

Razer6:
This synchronizes the OTP layout with the internal version. Internally, we have another alert for the socdbg module, which requires to increase the particular OTP field. Since we upstream that module in the future, we want to go ahead with the OTP changes


andreaskurth (approved):

andreaskurth:
CHANGE AUTHORIZED: hw/ip/otp_ctrl/data/otp_ctrl.hjson
CHANGE AUTHORIZED: hw/ip/otp_ctrl/rtl/otp_ctrl_part_pkg.sv
CHANGE AUTHORIZED: hw/ip/otp_ctrl/rtl/otp_ctrl_reg_pkg.sv

PR #21695, opened 2024-02-26
[top/pads] Switch CC pins to 5V tolerant pad cells

msfschaffner:
We should double check whether the CC pads should be changed to a 5V tolerant pad type in order to fulfill the spec requirements.


a-will:
5V-tolerant is actually 5.5V-tolerant, right? The spec provides for a host to power Vconn through the CC pins that OT might be snooping.

a-will (approved):

meisnere (approved):

Issue #19425, opened 2023-08-11
[doc, pinmux] Broken links to pinout tabels in pinmux documentations


mazurek-michal:
### Description

There are broken links to pinout tables in `Pinmux` documentations: 
[Pinout and Pinmux Mapping](https://opentitan.org/book/hw/ip/pinmux/doc/programmers_guide.html#pinout-and-pinmux-mapping)

broken links:
ASIC: [Pinout Table](https://opentitan.org/book/hw/top_earlgrey/ip/pinmux/doc/autogen/pinout_asic/index.html)
CW310: [Pinout Table](https://opentitan.org/book/hw/top_earlgrey/ip/pinmux/doc/autogen/pinout_cw310/index.html)
CW340: [Pinout Table](https://opentitan.org/book/hw/top_earlgrey/ip/pinmux/doc/autogen/pinout_cw340/index.html)

jrvanwhy:
I believe this is a dupe of #18535 (though #18535 describes a second issue as well).

mazurek-michal:
@jrvanwhy Ticket number that you provide is already merged, and current web page have broken links. 
Can you describe how it is related to current documentation state.

jrvanwhy:
Oops, I meant #18585 .

msfschaffner:
CC @hcallahan-lowrisc @HU90m

Issue #12848, opened 2022-05-24
[top/doc] Autogenerated tables for top-specific IP


msfschaffner:
We have several top-level specific IPs for which the configuration is not evident from the technical IP specification alone.
Examples are the plic, alert handler, pinmux, pwrmgr, rstmgr, clkmgr.

For some of these, we do already generate top-level specific documentation, such as for example the pinout and pinmux configuration tables [here](https://docs.opentitan.org/hw/ip/pinmux/doc/#pinout-and-pinmux-mapping).

It would probably be good to enhance documentation for other IPs to follow this example so that SW developers can find that information more readily.

andreaskurth:
Triaged for `pwrmgr`. Assigning https://github.com/lowRISC/opentitan/labels/Type%3AIcebox as I think this is not needed for M2.5.

GregAC:
Triaged for `rstmgr`, added https://github.com/lowRISC/opentitan/labels/Type%3AFutureRelease

GregAC:
Triaged for `pinmux` same conclusion as above

andreaskurth:
Triaged for `clkmgr`, ditto.

matutem:
The description of this issue is vague, and the link meant to show an example of the desired tables is broken. Please update so we can determine what really needs to be done.

msfschaffner:
I fixed the link. The issue has been left open intentionally since it's not fully clear what we want to generate yet, and we can discuss what would be useful. Information that I think could be useful (especially for the top-level documentation):
- table of reset requests and associated indices
- table of wakeup requests and associated indices
- map of alert IDs
- map of IRQ IDs
- pinmux configuration
- pinout (already exists to some degree)
- memory map (already exists to some degree)
etc

PR #21696, opened 2024-02-26
[topgen] Fix broken pinmux table link

### LINKED ISSUE #18585: [doc] Links to pinmux configuration tables are broken
### Description

The links in the pinout config tables [here](https://opentitan.org/book/hw/ip/pinmux/doc/programmers_guide.html#pinout-and-pinmux-mapping) do not seem to work anymore.

msfschaffner:
This is a short term fix and addresses #18585 and #19425.
Longer term, we should align how top-level documentation is generated, see #12848.

### DIFF
--- a/hw/ip/pinmux/doc/programmers_guide.md
+++ b/hw/ip/pinmux/doc/programmers_guide.md
@@ -108,11 +108,7 @@ A typical programming sequence for the wakeup detectors looks as follows:
 
 ## Pinout and Pinmux Mapping
 
-The tables below summarize the pinout and pinmux connectivity for certain top-level designs.
-
-### Top Earlgrey
-
-{{#include ../../../top_earlgrey/ip/pinmux/doc/autogen/targets.md}}
+Please see the top-level specific documentation for detailed pinout and pinmux mapping tables (e.g. for [Top Earlgrey](https://opentitan.org/book/hw/top_earlgrey/ip/pinmux/doc/autogen/targets.html)).
 
 ## Device Interface Functions (DIFs)
 


a-will (approved):

hw/top_earlgrey/ip/pinmux/doc/autogen/targets.md
```diff
@@ -4,8 +4,8 @@ DO NOT EDIT THIS FILE DIRECTLY.
...
-|:-------------:|:-----------:|:-------------:|:--------------:|:--------------:|:-------------:|:---------------------------------:|
-|     ASIC      |      4      |      47       |       14       |       10       |      71       | [Pinout Table](./pinout_asic.md)  |
-|     CW310     |      4      |      47       |       14       |       14       |      75       | [Pinout Table](./pinout_cw310.md) |
-|     CW340     |      4      |      47       |       14       |       13       |      74       | [Pinout Table](./pinout_cw340.md) |
+|  Target Name  |  #IO Banks  |  #Muxed Pads  |  #Direct Pads  |  #Manual Pads  |  #Total Pads  |                           Pinout / Pinmux Tables                            |
+|:-------------:|:-----------:|:-------------:|:--------------:|:--------------:|:-------------:|:---------------------------------------------------------------------------:|
+|     ASIC      |      4      |      47       |       14       |       10       |      71       | [Pinout Table](../../../top_earlgrey/ip/pinmux/doc/autogen/pinout_asic.md)  |
```
a-will, line 9: Hm... actually, is this the right number of directory changes? It looks to me like this new link wouldn't resolve.

Also, did you find a broken link on the website? It was working for me.

a-will: It seems like this would break the top-level documentation here: https://opentitan.org/book/hw/top_earlgrey/ip/pinmux/doc/autogen/targets.html

msfschaffner: There are two scripts, I believe. One that generates top-level docs, and one for the IP itself. The latter is broken:
https://opentitan.org/book/hw/ip/pinmux/doc/programmers_guide.html#top_earlgrey-pinmux-targets

msfschaffner: Ah I see now that this table is reused here:
https://opentitan.org/book/hw/top_earlgrey/ip/pinmux/doc/autogen/targets.html

That was not how this was initially intended.

a-will: I'm feeling like the top-level documentation is already fine, but it's the IP-level that is broken (which is referenced in the issue you linked). Should we consider eliminating the top-level-specific pin-out table from the IP docs instead?

https://github.com/lowRISC/opentitan/blob/c3d9a993d78670d45f9085466e80fe6d263d829e/hw/ip/pinmux/doc/programmers_guide.md?plain=1#L109-L115

msfschaffner: Yeah that may be the easier fix. Let me do that.

msfschaffner: Updated - PTAL

a-will (approved):

PR #21697, opened 2024-02-26
[dv,sram_ctrl] Fix a few failing tests

matutem:
Improve handling of ram initialization: refine the CSR expectations, and
move most processing outside process_tl_access since it is not called for
the CIP common tests; also, ram initialization blocks RAM accesses so
there could be a long time with transactions outstanding, which can cause
random reset timeouts.

Handle re-scramble during test execution, letting the push_pull agent
become active. In case of re-scrambling, invalidate the expected memory
state since it becomes invalid.

Add missing calls to cfg.stop_transaction_generators.

### DIFF
--- a/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
+++ b/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
@@ -678,6 +678,8 @@ class cip_base_vseq #(
                                             bit          do_tl_err = 1,
                                             uint         reset_delay_bound = 10_000_000);
     `DV_CHECK_FATAL(seq != null)
+    `uvm_info(`gfn, $sformatf("running run_seq_with_rand_reset_vseq for sequence %s",
+                               seq.get_full_name()), UVM_MEDIUM)
 
     for (int i = 1; i <= num_times; i++) begin
       bit ongoing_reset;
@@ -704,6 +706,8 @@ class cip_base_vseq #(
                   dv_vseq.do_apply_reset = 0;
                   dv_vseq.set_sequencer(p_sequencer);
                   `DV_CHECK_RANDOMIZE_FATAL(dv_vseq)
+                  `uvm_info(`gfn, $sformatf("Starting sequence %s", dv_vseq.get_full_name()),
+                            UVM_MEDIUM)
                   dv_vseq.start(p_sequencer);
                 end
               join
@@ -769,12 +773,17 @@ class cip_base_vseq #(
 
     cfg.set_intention_to_reset();
 
-    `uvm_info(`gfn, $sformatf("Waiting for %0d cycles with a run of no accesses", wait_cycles),
+    `uvm_info(`gfn, $sformatf(
+              "Waiting up to %0d cycles for a long enough run of no accesses", wait_cycles),
               UVM_MEDIUM)
     for (; cycles_waited < wait_cycles || cycles_with_no_accesses > 0; ++cycles_waited) begin
       if (!has_outstanding_access()) begin
+        `uvm_info(`gfn, "A cycle with no outstanding accesses", UVM_MEDIUM)
         ++cycles_with_no_accesses;
         if (cycles_with_no_accesses > CyclesWithNoAccessesThreshold) begin
+          `uvm_info(`gfn, $sformatf(
+                    "Finally no outstanding accesses after %d cycles", cycles_waited),
+                    UVM_MEDIUM)
           break;
         end
       end else begin
--- a/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
+++ b/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
@@ -22,6 +22,8 @@ class sram_ctrl_base_vseq #(parameter int AddrWidth = `SRAM_ADDR_WIDTH) extends 
     super.pre_start();
     void'($value$plusargs("partial_access_pct=%0d", partial_access_pct));
     `DV_CHECK_LE(partial_access_pct, 100)
+    // Wait for ram initialization to be done, since it blocks memory accesses.
+    `DV_WAIT(cfg.in_init == 1'b0, "Timed out waiting for initialization done")
   endtask
 
   virtual task dut_init(string reset_kind = "HARD");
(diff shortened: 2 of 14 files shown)


hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -704,6 +706,7 @@ class cip_base_vseq #(
                   dv_vseq.do_apply_reset = 0;
                   dv_vseq.set_sequencer(p_sequencer);
                   `DV_CHECK_RANDOMIZE_FATAL(dv_vseq)
+                  `uvm_info(`gfn, $sformatf("Starting sequence %s", dv_vseq.get_full_name()), UVM_MEDIUM)
```
rswarbrick, line 709: Nit: Missing a newline here, to break the long line?

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -678,6 +678,8 @@ class cip_base_vseq #(
                                             bit          do_tl_err = 1,
                                             uint         reset_delay_bound = 10_000_000);
     `DV_CHECK_FATAL(seq != null)
+    `uvm_info(`gfn, $sformatf("running run_seq_with_rand_reset_vseq for sequence %s",
+                                seq.get_full_name()), UVM_MEDIUM)
```
rswarbrick, line 682: Nit: Indentation is off here.

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -740,10 +743,25 @@ class cip_base_vseq #(
     end
   endtask
 
+  // This can be overriden in derived classes to wait a longer time for no outstanding accesses
```
rswarbrick, line 746: Genuine SV question: Can this just be a field in the class, which gets a different initialiser in subclasses? It might be simpler than messing around with virtual functions...

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -754,13 +772,15 @@ class cip_base_vseq #(
...
-    // accesses. If the wait doesn't clear something is probably wrong: perhaps some loop
-    // sending CSR transactions is missing a break if stop_transaction_generators() is set.
-    for (int i = 0; i < 10000 || cycles_with_no_accesses > 0; i++) begin
+    `uvm_info(`gfn, $sformatf("Waiting for %0d cycles with a run of no accesses", wait_cycles),
+              UVM_MEDIUM)
+    for (; cycles_waited < wait_cycles || cycles_with_no_accesses > 0; ++cycles_waited) begin
       if (!has_outstanding_access()) begin
+        `uvm_info(`gfn, "A cycle with no outstanding accesses", UVM_MEDIUM)
         ++cycles_with_no_accesses;
         if (cycles_with_no_accesses > CyclesWithNoAccessesThreshold) begin
```
rswarbrick, line 781: I think we're currently waiting for `min(CyclesWithNoAccessesThreshold, wait_cycles)` cycles. Is that correct? Maybe it's worth saying that in a comment just above the for loop. Also, I'm not sure that the message that's printed is guaranteed to be true: do we need to tweak it slightly?

rswarbrick, line 781: I'm not sure that the code is currently waiting for consecutive cycles with no access (with, or without this PR).

Don't we need something that sets `cycles_with_no_accesses` to zero if `has_outstanding_access` returned true?

rswarbrick (commented):
I've gone through the `cip_base_vseq` changes and think there are some tweaks that we need. Happy to read more later.

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
```diff
@@ -80,7 +106,7 @@ class sram_ctrl_base_vseq #(parameter int AddrWidth = `SRAM_ADDR_WIDTH) extends
     cfg.disable_d_user_data_intg_check_for_passthru_mem = 1;
     ral.ctrl.renew_scr_key.set(1);
     csr_update(.csr(ral.ctrl));
-    if (wait_valid) csr_spinwait(.ptr(ral.status.scr_key_valid), .exp_data(1));
+//    if (wait_valid) csr_spinwait(.ptr(ral.status.scr_key_valid), .exp_data(1));
```
jdonjdon, line 109: Can you remove this?

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
rswarbrick: Oops, I hadn't read this properly. Thanks for checking!

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
matutem: If you can help me: it seems most vseq's don't have an explicit constructor, and also where would the constructor be called? Or are you suggesting to just set this.wait_cycles_<whatever we call this variable> in pre_start for example?

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
matutem: We can wait for up to wait_cycles + CyclesWithNoAccessesThreshold, and I changed the log message to wait for this many cycles for a run, and it doesn't say for such run to complete. The comment above he function is meant to explain it. By the way, I changed this code to wait for this many cycles for such a run to start (not to end) in a previous PR after your suggestion :)

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
rswarbrick: I was wondering what happens if you do something like this:
```systemverilog
class A;
  int x = 123;

  task foo();
    $display("foo(): %0d", x);
  endtask
endclass;

class B extends A;
  int x = 234; // I'm overwriting the base class here

endclass;

initial
  B b = new;
  b.foo();
end
```

Unfortunately, I just tried. It prints
```
foo(): 123
```

Not what I was hoping for! :-)

I just looked in the 1800 standard, and this sort of initialiser in the definition of a class property is only shown for const properties. I'm couldn't quickly see anything that dictates how my example should behave, but Xcelium, at least, doesn't do what I was hoping for...

I guess we could do it off the constructor like this:
```systemverilog
class A;
  int x;

  function new;
    x = 123;
  endfunction

  task foo();
    $display("foo(): %0d", x);
  endtask
endclass;

class B extends A;
  function new;
    super.new();
    x = 234;
  endfunction
endclass;
```
This works, but it's pretty ugly! I guess it's equivalent to doing it in `pre_start` like you describe.

So, to answer my question, "No!" :-)

Let's keep the virtual function approach that you suggested.

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
```diff
@@ -57,16 +59,31 @@ class sram_ctrl_base_vseq #(parameter int AddrWidth = `SRAM_ADDR_WIDTH) extends
     req_mem_init();
   endtask
 
-  // Request a memory init.
-  //
+  // Incease the number of cycles to wait for no outstanding accesses, since it is hard
+  // to sequence all invocations of pre_start, and that can trigger ram init which blocks
+  // accesses for a long time. The RAM has 32K words, and it updates one per cycle, so
+  // 50,000 cycles should be okay.
```
rswarbrick, line 65: I think I agree with the calculation. Should the value in the function be a factor of ten smaller? It's currently 500_000.

rswarbrick (approved):
Nitty arithmetic question, but this looks good to me.

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
matutem: Indeed. The 500_000 number was leftover from developing the change. Done.

PR #21701, opened 2024-02-27
[spi_device] Clean up TODOs and add a mechanism to reinit the read buffer

a-will:
- Remove outdated TODOs / ICEBOXs.
- DIFs only: Add a mechanism to clear outstanding requests to the status registers
- Add a mechanism to reinit the read buffer without needing to reset the entire spi_device IP
- Document the SRAM layout and its relationship to 1r1w memories.

### DIFF
--- a/hw/ip/spi_device/data/spi_device.hjson
+++ b/hw/ip/spi_device/data/spi_device.hjson
@@ -487,6 +487,22 @@
           tags: [// This CSR only briefly pulses to anything other than 0.
                  "excl:CsrNonInitTests:CsrExclWrite"]
         },
+        { bits: "1",
+          name: "FLASH_READ_BUFFER_CLR",
+          swaccess: "rw1s",
+          hwaccess: "hrw",
+          desc: '''Set to clear the read buffer state.
+
+            When set to 1, resets the flash read buffer state that tracks the host read address.
+            The reset should only be used when the upstream SPI host is known to be inactive.
+            This function is intended to allow restoring initial values when the upstream SPI host is reset.
+
+            This CSR automatically resets to 0.
+            '''
+          resval: "0",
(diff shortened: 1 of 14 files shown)


a-will:
Apologies to my DV partners for adding these offhanded sub-block reset features at this stage of the game. It's been a random walk leading to occasional discovery of obstacles to a quick warm boot of the upstream host.

hw/ip/spi_device/data/spi_device.hjson
```diff
@@ -487,6 +487,22 @@
           tags: [// This CSR only briefly pulses to anything other than 0.
                  "excl:CsrNonInitTests:CsrExclWrite"]
         },
+        { bits: "1",
+          name: "FLASH_READ_BUFFER_CLR",
+          swaccess: "rw1s",
```
msfschaffner, line 492: is this a single-cycle operation or do we need the ability to poll clearing status?
if it is the former, it may make sense to make this a `swaccess: wo` and `hwaccess: hro` register.

msfschaffner (approved):
Thanks @a-will, just one suggestion.

hw/ip/spi_device/data/spi_device.hjson
a-will: It's a single cycle, the same as the field above. wo / hro works for me, and I wasn't sure which way best represented the intention. Just need to add the qe in that case.

msfschaffner: Ah I should have expanded that in the viewer :). 
wo/hro seems more concise, but this is likely incompatible with the higher `MODE` bits present.
So in order to avoid further changes to the CSR layout, I think it's ok to merge this as-is.

PR #21604, opened 2024-02-21
[hmac] Implementation of wider digest & configurable key length

gdessouky:
This brings the implementation of the wider digest size and configurable key length for HMAC (issue #20968).  This should allow the HMAC IP to compute SHA-2 and keyed HMAC using either SHA-256, SHA-384 or SHA-512. This also allows configuring the key length to one of the supported 128-bit/256-bit/384-bit/512-bit/1024-bit.  If a key length larger than the block size (which is twice the digest length) of a configured mode is picked, this gets capped to the block size, i.e., if 1024-bit key is selected with SHA-256, this will get capped at 512-bit to simplify the HW implementation. When a larger key length is required, the SW is expected to manage this by hashing the key and then supplying that hashed key to the HW. 

SW usage requires additionally configuring the digest size and key length required - DIFs will need to be modified accordingly. Unsupported key lengths and digest sizes are not handled or tested at the moment; most conservative/correct approach is to not proceed with the operation and flag an error, and leave the digest CSRs untouched.

This PR also brings changes to the DV environment to provide preliminary support for the extended features and to pass the basic smoke test. I haven't modified other tests yet, and definitely don't see the full regression tests running without extending them all appropriately.

Note that this will need to be merged together with the save & restore implementation in PR  #21307 and tested again to pass basic smoke.

Follow-up PRs needed:
- [ ] update documentation and programmer's guide
- [ ] update DIFs
- [x] update HMAC version number 
- [ ] handle unsupported digest size and key length configuration: flag an error and leave digest CSRs retaining previous digest. These are not dealt with at the moment in the RTL and DV.
- [ ] 1 or 2 DV checks/assertions trigger issues which I have commented them for now (labelled with TODO); I believe they are FIFO status issues that have come up because of the integration of the HMAC core with the 32-bit wrapper for the wider SHA-384/512 engine, and will need DV adaptations. 
- [ ] extend DV support for the rest of the existing tests
- [ ] extend testplan to test new cases; e.g., supply unsupported digest size or key length configuration, configuration changes during runtime, etc.

PR #21307, opened 2024-02-12
[hmac] Initial implementation of save & restore

andreaskurth:
This is an initial implementation of save & restore for HMAC (issue #49), which should allow SW to switch between different parallel message streams. In short, SW usage follows a pattern like this:
1. Start processing message stream A by configuring HMAC and then setting the `CMD.hash_start` bit.
2. Write an arbitrary number of message blocks to HMAC's `MSG_FIFO`.
3. Stop HMAC by setting the `CMD.hash_stop` bit and wait for the `hmac_done` interrupt (or poll the interrupt status register).
4. Save the context by reading the `DIGEST_0`..`7` and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers. (The values in the `CFG` register must also be preserved, but that is purely SW-controlled so doesn't need to be read from HW.)
5. Repeat steps 1-4 for message stream B.
6. Restore the context of message stream A by writing the `CFG`, `DIGEST_0`..`7`, and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers.
7. Continue processing message stream A by setting the `CMD.hash_continue` bit.
8. Write an arbitrary number of message blocks to HMAC's `MSG_FIFO`.
9. Continue this with as many message blocks and parallel message streams as needed. The final hash for any message stream can be obtained at any time (no need for complete blocks) as before by setting `CMD.hash_process` and waiting for the `hmac_done` interrupt / status bit, finally reading the digest from the `DIGEST` registers.

A noteworthy limitation of this implementation is that context can only be switched on complete message blocks (512 bit for SHA-256). This is so that no additional internal state (i.e., working variables and round counter) needs to be exposed to and controlled from SW. As noted in [issue #49](https://github.com/lowRISC/opentitan/issues/49#issuecomment-1935538297), this shouldn't be a major limitation for SW, though.

Although this PR adds preliminary DV support for the changes, save & restore does not yet get checked by the tests (the `save_and_restore_pct` randomization knob is currently 0 for all tests). I have manually verified that save & restore works under the basic operating conditions for HMAC and SHA but this needs to be tested much more thoroughly, especially the corner cases. When one enables save & restores in the tests, one will get errors from DV. My first assessment is that these are DV limitations, not RTL bugs, though.

I ran a full regression on HMAC to check that this doesn't break existing functionality or DV, and I got the same pass rates as reported in nightlies -- so no breakage of existing RTL & DV expected.

Opens before this PR can be merged:
[...]

PR #21703, opened 2024-02-27
[hmac] Bump version to 2.0.0 and reset development stages

andreaskurth:
With the addition of *save & restore* (PR #21307) and *wider digest and configurable key length* (PR #21604), HMAC will see major changes.  For this reason, this commit gives HMAC a major version bump to 2.0.0 and resets its development stages to L1 (Development), D1 (Functional), V0 (*Initial Work* because the DV environment may not be completely functional for some time), and S0 (*Initial Work* because the DIF may not be completely functional for some time).

The previous version 1.0.0 got assigned a commit ID for future reference even though it had not reached L2, V3, and S3.

### DIFF
--- a/hw/ip/hmac/data/hmac.hjson
+++ b/hw/ip/hmac/data/hmac.hjson
@@ -33,6 +33,15 @@
       design_stage:       "D3",
       verification_stage: "V2S",
       dif_stage:          "S2",
+      commit_id:          "2290dcc119067538aaf3fb5fd43c3dc2418458d1",
+      notes:              "",
+    }
+    {
+      version:            "2.0.0",
+      life_stage:         "L1",
+      design_stage:       "D1",
+      verification_stage: "V0",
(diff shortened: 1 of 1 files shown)


andreaskurth:
CC @msfschaffner: as discussed

vogelpi (approved):
This looks good to me but my understanding is that we really would like the smoke test to pass, i.e., V1 for M2.

andreaskurth:
> my understanding is that we really would like the smoke test to pass, i.e., V1 for M2.

Agreed, and we'll likely get there very soon and hopefully ahead of M2.

andreaskurth:
_CW310 SiVal Tests_ and _CW310 ROM Tests_ failed, but I don't see how this version bump could cause that. Merging.

gdessouky:
looks good to me, thanks @andreaskurth!

Issue #19455, opened 2023-08-15
[manuf] Earlgrey ES FT provisioning binary and test harness for generic SKU


timothytrippel:
A provisioning device binary and host test harness need to be written to run in FT for provisioning Earlgrey ES chips for the generic OpenTitan SKU. The details / requirements for both are listed below:

## Device Binaries

There should be two device binaries:
1. one that is JTAGed into SRAM that executes in a TEST_UNLOCKED* state (this is to program the *_SW_CFG OTP partitions, since ROM cannot boot until specific values like ROM_EXEC_EN here are programmed, and without ROM execution there is no bootstrap mechanism to get code into flash)
2. one that this is bootstrapped into flash that executes in a mission mode state (DEV, PROD, PROD_END)

The first is also known as the _"individualization"_ binary, as this binary configures OTP setting defined by a chip's SKU, and the second is know as the _"personalization"_ binary, as this binary will instantiate a chip's cryptographic identity.

### Individualization

The individualization device binary should:
- [x] be loadable into SRAM over JTAG
- [x] check the LC state is in TEST_UNLOCKED*
- [x] setup a UART/SPI console
- [x] perform the corresponding **individualization**  operation based on generic OpenTitan SKU requirements to provision the following OTP partitions:
    - [x] CREATOR_SW_CFG
    - [x] OWNER_SW_CFG
    - [x] HW_CFG
- [x] lock the OTP partitions programmed above (SW partitions should be locked by computing a SHA256 hash of the partitions)
- [x] indicate success/failure over the console
- [x] halt the CPU since ROM execution should now be active

### Personalization

The personalization device binary should:
- [x] be bootstrapped into flash
- [x] check the LC state is in DEV/PROD/PROD_END
- [x] check the OTP partitions provisioned by the SRAM program have been locked
- [x] setup a UART/SPI console
- [x] self-report hash of itself over console
- [x] perform the corresponding **personalization** operation that should provision the following flash info pages /OTP partitions:
    - [x] manufacturing entropy for attestation keygen (flash info)
    - [x] creator root secret (flash)
    - [x] owner root secret (flash)
    - [x] flash and SRAM scrambling keys (OTP - secret1); perform software reset after to activate scrambling of flash flash-info pages
    - [x] root key (OTP - secret2)
    - [x] RMA unlock token (OTP - secret2)
- [x] lock the OTP secret1 and 2 partitions and issue a soft reset to activate them
[...]

PR #21704, opened 2024-02-27
[dice] endorse UDS TBS cert and parse all certs with OpenSSL

timothytrippel:
This updates the personalization provisioning flow to endorse the UDS TBS certificate on the host (using a fake endorsement key added to the repo for testing). Additionally, this adds code to parse each attestation certificate with OpenSSL to ensure each certificate is valid.

In a follow up PR, the endorsed UDS certificate will be transmitted back to the device over the console and written to flash.

This partially addresses #19455.


pamaury (approved):
Looks great. Nice that you can reuse the der generator to create the certificate from the TBS!

timothytrippel:
> Looks great. Nice that you can reuse the der generator to create the certificate from the TBS!

Yes, this was very convenient, thanks for adding it!

PR #21705, opened 2024-02-27
[prim_alert_sender,cdc] Ability to sync alert inputs

Razer6:
OpenTitan integrated supports external alerts. These alerts are generated using the same prim_alert_sender primitives. However, since the alerts are coming from an external source, the alert signals might be in a different clock domain.

This PR adds the ability to enable a 2-FF synchronizer stage for the alert sender inputs, i.e., alert_req_i and alert_test_i.

### DIFF
--- a/hw/ip/prim/rtl/prim_alert_sender.sv
+++ b/hw/ip/prim/rtl/prim_alert_sender.sv
@@ -48,7 +48,10 @@ module prim_alert_sender
   parameter bit AsyncOn = 1'b1,
   // alert sender will latch the incoming alert event permanently and
   // keep on sending alert events until the next reset.
-  parameter bit IsFatal = 1'b0
+  parameter bit IsFatal = 1'b0,
+  // Synchronize alert inputs with a 2-FF synchronizer when being used
+  // externally, where the inputs might be in a different clock domain
+  parameter bit SyncAlertInput = 1'b0
 ) (
   input             clk_i,
   input             rst_ni,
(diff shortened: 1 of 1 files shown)


msfschaffner:
My recommendation would be to do this outside of the primitive, since we have FPV and cycle accurate scoreboards that use this primitive, and aligning those will be a lot of work.

Also, if you were to use `prim_alert_sender` and `prim_alert_receiver`, please note that we designed them so that they can handle asynchronous transitions natively in the protocol itself. I.e., the sender and receiver primitives can be in separate clock and reset domains.

Can you let me know how this is used outside of the RoT? Because the channels that are exposed to the SoC via `soc_proxy` are not compatible with this primitive (it's a different protocol).

Razer6:
Thanks for those insights. We mostly use the `prim_alert_sender` in the following way:

```systemverilog
prim_alert_sender #(
  .AsyncOn(1'b1),
  .IsFatal(1'b1)
) u_alert_sender (
  .clk_i         ( ext_clk                ),
  .rst_ni        ( ext_rst                ),
  .alert_test_i  ( 1'b0                   ),
  .alert_req_i   ( alert_req              ),
  .alert_ack_o   (                        ),
  .alert_tx_o    ( alert_to_socproxy_tx   ), 
  .alert_rx_i    ( alert_to_socproxy_rx   ),
  .alert_state_o (                        )
);
```

`alert_req` might come from a different clock domain. So from your last coment if `ext_rst`/`ext_clk` are different to the one of top_darjeeling, we should be good to go as the CDC is handled on protocol level. We only need to deal with aligning the request to the sender's domain or move that one the `alert_req`'s domain.

msfschaffner:
yes it would be preferred to just align the sender side so that `alert_req` is synchronous to `ext_clk` and `ext_rst`.

one more thing: where do you place the receivers and how are they wired up?
do you use the integrity and alert ping features?

there are a few gotchas there and I would like to make sure you are using this the right way.

Razer6:
> one more thing: where do you place the receivers and how are they wired up?

I grepped the codebase. The senders, as instantiated above, go to the soc_proxy or they enter a local alert manager. What gotchas do you have on your radar?

Razer6:
We discussed that topic offline and came to the conclusion that this approach is not needed and external senders use a different way to notify top_darjeeling for an alert.

Issue #20964, opened 2024-01-24
[spi_device] Scoreboard loses track of pending Read Buffer interrupts triggered at transaction end


a-will:
### Description

When predicting values for interrupts that are triggered during SPI transactions, the scoreboard needs to wait a few (main / TL-UL) clock cycles for the triggering event to make its way over to the main clock domain. However, if this event occurs at the end of a transaction, the thread waiting to sync the update can get killed, losing the event.

https://github.com/lowRISC/opentitan/blob/17d5a97c3bcd838470596d6fa0457e88de3b3e17/hw/ip/spi_device/dv/env/spi_device_scoreboard.sv#L880-L894

`predict_read_buffer_intr()` leads to a forked thread waiting for those clock cycles at https://github.com/lowRISC/opentitan/blob/17d5a97c3bcd838470596d6fa0457e88de3b3e17/hw/ip/spi_device/dv/env/spi_device_scoreboard.sv#L932-L941

However, `item.mon_item_complete` goes high shortly afterwards (when CSB de-asserts), and if that isn't enough time for the 4 main clock cycles, the pending interrupt update is lost, causing a mismatch with the predictor of a subsequent TL-UL access.

a-will:
Thanks @antmarzam!

PR #21712, opened 2024-02-27
Spi device readbuf interrupt tb fix

antmarzam:
The TB model of flash_mode interrupts isn't complete.
The task which updates a predicted interrupt pending gets killed on CSB=1.

Solution was to move the interrupt pending prediction out of DV_SPINWAIT to avoid thread being killed on CSB assertion.

In addition, There are situations in which the TB wasn't keeping track and clearing interrupt pending state correctly. 
For example, if the TB has interrupts pending before a read to intr_state and there is another interrupt of the same kind from the moment the CSR is read until the moment the CSR is written to clear interrupts, the TB was still believing after the write it had predicted interrupts when it didn't. 

The TB now clears interrupt pending ('intr_trigger_pending') on write to intr_status to avoid any later mismatch.

### DIFF
--- a/hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
+++ b/hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
@@ -35,6 +35,7 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
 
   // when interrupt is triggered, it may take a few cycles before it's reflected in a TL read
   local bit [NumSpiDevIntr-1:0] intr_trigger_pending;
+  bit [NumSpiDevIntr-1:0]       intr_exp_read_mirrored;
 
   // tx/rx async fifo, size is 2 words
   local bit [31:0] tx_word_q[$];
@@ -871,6 +872,8 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
       spi_item item;
       uint payload_idx;
       bit [31:0] start_addr, offset, read_buffer_addr;
+      event      interrupt_update_ev;
+      bit        reading_readbuffer=1;
 
       upstream_spi_req_fifo.get(item);
       if (cfg.spi_host_agent_cfg.spi_func_mode == SpiModeTpm) begin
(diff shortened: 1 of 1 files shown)


hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -887,6 +890,36 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
         continue;
       end
 
+      `uvm_info(`gfn, "Process read buffer command", UVM_DEBUG)
+
+      //Use fork...join_any and lock until we return from 'compare_mem_byte' below.
```
rswarbrick, line 895: Nit: Missing space after `//`, and similar for some of the other comments.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -35,6 +35,7 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
 
   // when interrupt is triggered, it may take a few cycles before it's reflected in a TL read
   local bit [NumSpiDevIntr-1:0] intr_trigger_pending;
+  bit [NumSpiDevIntr-1:0]       intr_exp_read_mirrored;
```
rswarbrick, line 38: I don't think it really matters, but maybe it's worth making this `local` to match the surrounding fields.

rswarbrick, line 38: I wonder whether this name should be tweaked a bit. When I first saw it, I thought this might be a boolean flag that said whether "intr_exp" had been read. I don't think that's quite what we want :-)

rswarbrick (commented):
I've not read through this carefully yet (sorry), but here are a couple of silly tidyups that might be worth doing anyway.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -887,6 +890,36 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+      // 'item.mon_item_complete=1' which occurs the moment CSB transitions 0->1.
+      // If there was a read-buffer interrupt which occurred shortly before CSB=1
+      // The DV_SPINWAIT could kill the thread (depends on timing), without update
+      // to the interrupt predictions.
+      fork
+        begin
+          while(reading_readbuffer) begin
+            `uvm_info(`gfn, "Blocking on SV event 'interrupt_update_ev'", UVM_DEBUG)
+            wait(interrupt_update_ev.triggered);
+            if(reading_readbuffer) begin
```
rswarbrick, line 906: Could we drop this `if` test entirely and put the UVM_DEBUG message just outside of the `while` loop? I think the code would end up slightly simpler:
```systemverilog
        begin
          wait (interrupt_update_ev.triggered);
          while (reading_readbuffer) begin
            `uvm_info(`gfn, "Blocking on SV event 'interrupt_update_ev'", UVM_DEBUG)
            `uvm_info(`gfn,
                      $sformatf("'interrupt_update_ev' event is triggered (buffer_addr=0x%0x)",
                                read_buffer_addr + 1 ), UVM_DEBUG)
            predict_read_buffer_intr(read_buffer_addr + 1, item.opcode);

            // The task above is delay free since it's within a fork...join_any.
            // Adding some "SPI-side" clk_delay to ensure the triggering events are noticed.
            #(cfg.spi_host_agent_cfg.sck_period_ps/2 * 1ps);

            wait(interrupt_update_ev.triggered);
          end
          `uvm_info(`gfn, "Exiting interrupt update fork", UVM_DEBUG)
        end
```
(although the debug messages probably need a bit of tweaking after my rearrangement!)

rswarbrick (commented):
Some more nitty comments about the first commit (sorry!)

antmarzam:
This PR deals with:
- #20964

rswarbrick (approved):

PR #21713, opened 2024-02-27
[cryptotest] ECDSA Sign operation tests

RyanTorok:
Adds support for testing the `sign` operation in ECDSA to the cryptotest framework.

~Dependent on #20853~


pamaury (approved):
I had an overall look at the PR and it seems to make sense. I can't say I checked everything single detail because that's a lot of code.

RyanTorok:
A few structural changes were made to #20853 to make the code easier to follow, and those will need to be propagated to this PR as well.

sw/host/tests/crypto/ecdsa_kat/src/main.rs
```diff
@@ -240,12 +310,147 @@ fn run_ecdsa_testcase(
...
+                let mut output_signature =
+                    CryptotestEcdsaSignature::recv(&*uart, opts.timeout, false)?;
+                // Truncate signature values to correct size for curve and convert to big-endian
+                output_signature.r.truncate(output_signature.r_len);
+                output_signature.s.truncate(output_signature.s_len);
+                output_signature.r.reverse();
+                output_signature.s.reverse();
+                match test_case.curve.as_str() {
+                    "p256" => {
+                        // Convert qx and qy to big-endian (resize in case to full length beforehand
```
milesdai, line 333: Extra "in case"?

sw/host/tests/crypto/ecdsa_kat/src/main.rs
```diff
@@ -240,12 +310,147 @@ fn run_ecdsa_testcase(
...
+                        hash_oid.expect("Unsupported hash algorithm for Sign verification"))
+                        .expect("Invalid signature format"),
+                    )
+                    .is_ok()
+                    }
+                    "p384" => {
+                        // Zero-fill the remaining space for the public key parameters, because
+                        // p384 requires an exact size
+
+                        // Convert qx and qy to big-endian (resize in case to full length beforehand
```
milesdai, line 388: Same as above - extra "in case"?

sw/host/tests/crypto/ecdsa_kat/src/main.rs
```diff
@@ -240,12 +310,147 @@ fn run_ecdsa_testcase(
...
+                        // in case there are leading zero(es))
+                        qx.resize(ECDSA_CMD_MAX_COORDINATE_BYTES_P256, 0u8);
+                        qy.resize(ECDSA_CMD_MAX_COORDINATE_BYTES_P256, 0u8);
+                        qx.reverse();
+                        qy.reverse();
+                        // Zero-fill the remaining space for the public key parameters, because
+                        // p256 requires an exact size
+                        // Verify the signature with the public key
+                        p256::ecdsa::VerifyingKey::from_encoded_point(
+                        &p256::EncodedPoint::from_affine_coordinates(
```
milesdai, line 343: Is there in an indentation problem here? The call to `from_affine_coordinates(` should should be shifted right by one right? I think this continues all the way down to the end of the `p256` match arm.

sw/host/tests/crypto/ecdsa_kat/src/main.rs
```diff
@@ -240,12 +310,147 @@ fn run_ecdsa_testcase(
...
+        d0: ArrayVec::try_from(d0.as_slice()).unwrap(),
+        d0_len: d0.len(),
+        d1: ArrayVec::try_from(d1.as_slice()).unwrap(),
+        d1_len: d1.len(),
+        unmasked_len: d0.len(),
+    }
+    .send(&*uart)?;
+    let success =
+        match operation {
+            CryptotestEcdsaOperation::Sign => {
```
milesdai, line 323: The `Sign` arm of this match statement is getting quite long and the nesting is pretty deep. Can this be split out into its own function to improve readability?

sw/host/tests/crypto/ecdsa_kat/src/main.rs
```diff
@@ -63,6 +72,11 @@ const ECDSA_CMD_MAX_COORDINATE_BYTES_P256: usize = 32;
 const ECDSA_CMD_MAX_SIGNATURE_SCALAR_BYTES_P384: usize = 48;
 const ECDSA_CMD_MAX_COORDINATE_BYTES_P384: usize = 48;
 
+// These values were generated randomly for testing purposes.
+// Each value must be less than the modulus for its respective curve.
+const RANDOM_MASK_P256: &str = "9665cc1c4e9e15354e3fc319ca7de255f2122dc3c16da15e6a83ce1d7a912df0";
```
milesdai, line 77: Is there a reason we're doing the masking on the host side? I'm not sure of the performance differences, but if the current bottleneck in these tests is the UART, could we benefit by not sending twice as many bytes for the private key?

milesdai (commented):
This generally lgtm, but I have a few questions and a some style comments.

sw/host/tests/crypto/ecdsa_kat/src/main.rs
RyanTorok: Unlike in AES, masking for asymmetric crypto is a rather complex operation. The two shards of the private key need to add to the actual key, mod the size of the curve. Here, we use a known-good Rust crypto library to perform the modular subtraction necessary to extract the two shards. If we did this on the device-side, we would have the implement the modular arithmetic entirely from scratch.

sw/host/tests/crypto/ecdsa_kat/src/main.rs
RyanTorok: This is fixed by moving the sign verification code to helper functions, per your below comment. Must have hit some strange behavior in `rustfmt.`

sw/host/cryptotest/testvectors/parsers/wycheproof_ecdsa_parser.py
```diff
@@ -71,6 +71,10 @@ def parse_test_vectors(raw_data):
                 # consumers may incorrectly handle large integers.
                 "qx": key["wx"],
                 "qy": key["wy"],
+                # Wycheproof does not provide tests for the `sign` operation,
+                # so we set the value of d (private key) to be zero as a
+                # placeholder.
+                "d": "00"
```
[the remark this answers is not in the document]
RyanTorok: The schema doesn't, but a previous version of the Rust test harness in this PR required `d` to be included for all test vectors. I recently changed this to do the argument validation later to remove this restriction, but I had only fixed this in the NIST test vector parser. This can now be removed.

milesdai (approved):
LGTM! Thanks for the changes!

milesdai:
kokoro failure is unrelated to this PR (fixed by #22147)

PR #21714, opened 2024-02-27
DO NOT MERGE][ci] Improve handling of bitstream jobs

pamaury:
With this commit, the bitstream jobs will now only use the EDA pool if they need to build a bitstream. If the bitstream can be downloaded from the cache, then the download will use a normal runners which avoids waiting uselessly for an EDA machine.

My suggestion is to only convert the CW340 job first, make sure that it works over a week and then convert the other jobs. Any thoughts?

**Comments welcome but note that this is a WIP!**

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -450,19 +450,13 @@ jobs:
       top_name: earlgrey
       design_suffix: cw310_hyperdebug
 
-- job: chip_earlgrey_cw340
-  displayName: CW340's Earl Grey Bitstream
-  # Build CW340 variant of the Earl Grey toplevel design using Vivado
-  dependsOn:
-    - lint
-  condition: and(succeeded(), eq(dependencies.lint.outputs['DetermineBuildType.onlyDocChanges'], '0'), eq(dependencies.lint.outputs['DetermineBuildType.onlyDvChanges'], '0'), eq(dependencies.lint.outputs['DetermineBuildType.onlyCdcChanges'], '0'))
-  pool: ci-public-eda
-  timeoutInMinutes: 150
-  steps:
-  - template: ci/fpga-template.yml
-    parameters:
-      top_name: earlgrey
-      design_suffix: cw340
+- template: ci/fpga-job.yml
+  parameters:
+    top_name: earlgrey
+    design_suffix: cw340
+    display_name: CW340's Earl Grey Bitstream
+    job_name: chip_earlgrey_cw340
+    build_timeout: 150
 
 - job: chip_englishbreakfast_cw305
   displayName: CW305's Bitstream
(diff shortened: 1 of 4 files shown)


a-will:
Just a heads-up that this might have some funny effects. Because the determination of the bitstream strategy happens on the EDA machines, queuing there can increase the probability of a cache hit when we are actively developing the RTL -- It gives time for a `master` build to upload its cached bitstreams to GCP.

With this fast lane created, we may end up in a "hurry up and wait" situation, where more PRs end up running bitstream builds due to deciding much earlier that there was no bitstream available. In addition, it may cause greater fairness issues for accessing FPGAs for tests, as RTL changes could end up way down the queue after the SW changes bypass them and cause a traffic jam at the FPGA test job.

The former issue could be alleviated by checking the strategy _again_ on the EDA machine. The latter issue can only be resolved by not having a bottleneck for FPGA test jobs in the first place. Right now, do the FPGA test agents retire an entire PR's jobs tremendously faster than FPGA build agents?

moidx:
Hi @pamaury, is it possible to discuss this bottleneck issues in the next software WG meeting? I think the current fan out of FPGA execution environments is causing a lot of strain in CI. We can probably agree on reducing the number of exec_envs drastically for pre-submit and continuous integration jobs, and run the bulk of the configurations in a nightly regression. 

I have some ideas on which exec_envs may be interesting to run in pre-submit, and how to support the more complicated test environments (e.g. manufacturing tests). 

Thanks

pamaury:
Sure, I am happy to discuss this at the next software meeting. I also agree that we are running too many tests in CI.

I understand your point @a-will but the problem that this PR is trying to solve is that we have many (software) PR where the strategy is always 'cached' because no RTL was changed, but those jobs still have to wait sometimes hour just to run 2min on an EDA machine to reach that conclusion. This is clearly a huge loss in latency and sometimes leads to under-utilization of the FPGA if the EDA machines are under stress. We could run the strategy determination again on EDA potentially if you are worried about it, this way this PR would just add an optimization ewssentially?

a-will:
> As to fairness, I don't think that's a real issue: if you end up queueing every PR for an EDA machine and let the FPGA sit idle, it seems worse to me than letting at least some FPGA job go through when they can, even if it's slightly unfair.

That isn't what happens, though. Again, when there is actually substantial load that causes long queuing in CI, there is _also_ a traffic jam at FPGA test. Right now, the PR throughput of the FPGA test nodes is not substantially higher than the average throughout of the build nodes. There is no idle FPGA, except in times of relatively moderate load.

Moreover, when we are in the high load state and there is active RTL development, the critical PRs are _not_ software. The deadline that is rapidly approaching is for hardware. We don't get excessive delays at FPGA build if there aren't changes that cause FPGA builds.

That means that this PR amounts to a _disastrous_ scheduling algorithm change, whereby software PRs are allowed to skip the line for FPGA tests, blocking the most important work of the time, when the scheduling algorithm has the most impact on project-wide outcomes.

elliotb-lowrisc:
Closing as no longer relevant (Azure, do-not-merge)

Issue #21706, opened 2024-02-27
[sival,triag] Some SiVal ROM_EXT are broken


pamaury:
### Description

The following tests have been found to be broken in `//hw/top_earlgrey:fpga_cw310_sival_rom_ext` execution environment. See #21580 for details. This issue tracks the lists for further triaging.
- `edn_kat_fpga_cw310_sival_rom_ext`
- `flash_ctrl_mem_protection_test_fpga_cw310_sival_rom_ext`
- `flash_ctrl_test_fpga_cw310_sival_rom_ext`
- `pmp_smoketest_tor_fpga_cw310_sival_rom_ext`
- `rv_core_ibex_address_translation_test_fpga_cw310_sival_rom_ext`
- `sram_ctrl_execution_test_fpga_cw310_sival_rom_ext`
- `uart_tx_rx_test_fpga_cw310_sival_rom_ext`
- `keymgr_test_fpga_cw310_sival_rom_ext`
- `boot_data_functest_fpga_cw310_sival_rom_ext`
- `irq_asm_functest_fpga_cw310_sival_rom_ext`
- `alert_functest_fpga_cw310_sival_rom_ext`
- `keymgr_functest_fpga_cw310_sival_rom_ext`
- `rnd_functest_fpga_cw310_sival_rom_ext`
- `rstmgr_functest_fpga_cw310_sival_rom_ext`
- `watchdog_functest_fpga_cw310_sival_rom_ext`
- `mod_exp_otbn_functest_hardcoded_fpga_cw310_sival_rom_ext`
- `rsa_verify_functest_fpga_cw310_sival_rom_ext`
- `sigverify_dynamic_functest_hardcoded_fpga_cw310_sival_rom_ext`
- `chip_power_idle_load_fpga_cw310_sival_rom_ext`
- `chip_power_sleep_load_fpga_cw310_sival_rom_ext`
- `csrng_edn_concurrency_test_fpga_cw310_sival_rom_ext`
- `csrng_kat_test_fpga_cw310_sival_rom_ext`
- `entropy_src_csrng_test_fpga_cw310_sival_rom_ext`
- `entropy_src_edn_reqs_test_fpga_cw310_sival_rom_ext`
- `flash_ctrl_clock_freqs_test_fpga_cw310_sival_rom_ext`
- `flash_ctrl_ops_test_fpga_cw310_sival_rom_ext`
- `keymgr_key_derivation_test_fpga_cw310_sival_rom_ext`
[...]

Issue #21548, opened 2024-02-19
[silicon_creator] implement ECDSA sigverify operations


timothytrippel:
Implement the ECDSA-equivalent sigverify operations on OTBN:
- [x] OTBN (#21824)
- [ ] https://github.com/lowRISC/opentitan/pull/23680
- Ibex (dropped)

Additionally, integrate these into the sigverify silicon_creator device library.

moidx:
Hi @jadephilipoom, @cfrantz, @timothytrippel, @vogelpi, @andreaskurth 

Currently we have been using OTBN execution in both silicon and FPGA environments in `earlgrey_es_sival` and `master` branches. 

`top_earlgrey_es` has an OTP configuration word that supports reverting to a software implementation of RSA in case there were hardware issues with OTBN during silicon bring-up. The option was not used in SiVal, and all our testing has thus far been based on OTBN performing signature verification.

Given that OTBN has so far been proven to work, I propose we drop the request to have support for ECDSA in Ibex. This is mainly due to the following reasons: 

1. Code size space constraints. We are currently with about 1KB of room left in the ROM, and we still have to implement some fixes for `Earlgrey-PROD.ROM`.
2. We can explore the option of running SLH-DSA and ECDSA in parallel to help optimize boot performance.

The 2. point above does not need to be part of `Earlgrey-PROD.ROM`, but can potentially be considered for `Earlgrey-PROD.ROM_EXT`. 

To further mitigate risk of not having an ECDSA software backup, we have focused on re-enabling ROM e2e test cases in DV. We will run a sub set of these tests in gate level simulation to include them to the tape-out sign-off criteria.

Please add your comments here and feel free to add this topic to any relevant working group agendas for further discussion.

jadephilipoom:
This sounds okay to me. It's nice to have redundancy in ROM, but the SPHINCS+ implementation gives us a little bit of protection already. If there was a security-critical issue with the ECDSA implementation that meant it accepted invalid signatures (or was vulnerable e.g. to a fault injection attack that made it accept invalid signatures), then we could always enable SPHINCS+ as an emergency mitigation. If the ECDSA implementation rejected valid signatures, then we wouldn't have an alternative, but this scenario is much less dangerous and seems unlikely to occur in a way that would affect large numbers devices/keys. For example, if there was a fault-injection attack that made the ECDSA implementation reject a valid signature, that wouldn't really be a big problem; it would just make the device fail to boot.

[...]

PR #21717, opened 2024-02-27
[boot_services] Add functional tests for the OTBN boot services library.

jadephilipoom:
These tests caught a few bugs and places where comments were needed; the first commits fix bugs and the last commit adds the test code.

Thanks to @ballifatih for finding one bug in the OTBN code, which brought to my attention that we were missing tests for this! Once I started writing them I found more, so it's really good that this came up.

### DIFF
--- a/sw/device/silicon_creator/lib/otbn_boot_services_functest.c
+++ b/sw/device/silicon_creator/lib/otbn_boot_services_functest.c
@@ -1,0 +1,321 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/device/lib/crypto/drivers/entropy.h"
+#include "sw/device/lib/dif/dif_keymgr.h"
+#include "sw/device/lib/dif/dif_kmac.h"
+#include "sw/device/lib/testing/flash_ctrl_testutils.h"
+#include "sw/device/lib/testing/keymgr_testutils.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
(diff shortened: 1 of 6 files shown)


sw/device/silicon_creator/lib/otbn_boot_services_functest.c
```diff
@@ -0,0 +1,321 @@
...
+      keymgr_testutils_check_state(&keymgr, kDifKeymgrStateCreatorRootKey));
+  CHECK_STATUS_OK(keymgr_testutils_advance_state(&keymgr, &kOwnerIntParams));
+
+  // Run endorsement (should overwrite the key with randomness when done).
+  hmac_digest_t digest;
+  hmac_sha256(kEndorseTestMessage, kEndorseTestMessageLen, &digest);
+  attestation_signature_t sig;
+  RETURN_IF_ERROR(otbn_boot_attestation_endorse(&digest, &sig));
+
+  // TODO: run an ECDSA signature verification here once we have code for that.
```
timothytrippel, line 216: maybe we could file an issue to track? or tie to #21548 ?

timothytrippel (approved):

timothytrippel:
Thanks for catching this @jadephilipoom !

jadephilipoom:
CI was failing on `sival_rom_ext` with a load access fault from this read in `dif_otp_ctrl`:
https://github.com/lowRISC/opentitan/blob/5cb6b899d73004a1bddf33d707a3a24757b3e54f/sw/device/lib/dif/dif_otp_ctrl.c#L768

This function gets called by `keymgr_testutils_startup`; it tries to lock the OTP partitions, but it doesn't look like it's allowed to do so in ROM_EXT. I think this falls into the same category as the other tests in https://github.com/lowRISC/opentitan/issues/21706, so I've marked it broken for now and tagged it with that issue.

PR #21709, opened 2024-02-27
[i2c] Increase maximum `FifoDepth` from 127 to 4095

andreaskurth:
This is a prerequisite for increasing the depth of some of the FIFOs
beyond 127.

This implies that the four FIFO levels can no longer be exposed to SW in
the single 32-bit `FIFO_STATUS` register.  Instead, each of the two
host-mode FIFO levels is now a 12-bit field in the `HOST_FIFO_STATUS`
register, and each of the two target-mode FIFO levels is now a 12-bit
field in the `TARGET_FIFO_STATUS` register.

PR #21723, opened 2024-02-27
[i2c] Add `AcqFifoDepth` parameter

andreaskurth:
**This PR currently depends on https://github.com/lowRISC/opentitan/pull/21709 and includes commits from that PR. Only the last commit is unique to this PR. Other than this dependency, this PR is ready for review but kept as draft for now until it can be rebased on `master` once #21709 got merged.**

This commit splits the depth of the ACQ FIFO from the `FifoDepth`
parameter into a separate, new parameter named `AcqFifoDepth`.  This
allows instantiators to make the ACQ FIFO of any `i2c` instance deeper
or shallower than the other FIFOs.

This commit also makes initial DV changes to cover the new parameter,
although for now `AcqFifoDepth` in DV is simply set to a fixed value
that is different from that of `FifoDepth`.  Later work should cover
different values of the parameter.

The three `i2c` instances in Earlgrey have the new parameter set to 64,
which is equal to the value of the `FifoDepth` parameter, for now.

### DIFF
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -100,7 +100,10 @@
   ],
   param_list: [
     { name: "FifoDepth",
-      desc: "Depth of FMT, RX, TX, and ACQ FIFOs",
+      desc: '''
+            Depth of FMT, RX, TX, and ACQ FIFOs.
+            The maximum supported value is 2^12-1, although much lower values are recommended to keep area requirements reasonable.
+            ''',
       type: "int",
       default: "64",
     }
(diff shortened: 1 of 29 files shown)


andreaskurth:
Similar to #21709 this is technically ready for review, but I'm leaving it as a draft PR to prevent it from getting merged accidentally as some decisions related to this are still open.

a-will (approved):

msfschaffner (approved):

andreaskurth:
This PR had already been approved, but I'm about to add a commit that sets the value of `AcqFifoDepth` for all three FIFOs, which is beyond the original intention of this PR. Thus I'm closing it and will open a new one.

PR #21724, opened 2024-02-27
[pinmux] Add missing SEC_CM labels and move to V2S

msfschaffner:
This adds missing labels for some of the countermeasures present in pinmux.
They are not complex and basically describe:
- the standardized bus integrity error (this was already present in the list)
- all multibit encoded LC inputs
- the LC signal-gated TAP mux/demux

These features are already covered in our FPV environment, but we currently do not have a good way to indicate that on the SEC_CM testplan. This is noted in the V2S checklist.


a-will (approved):

rswarbrick (approved):
Looks good to me.

PR #21725, opened 2024-02-27
[dv,otp_ctrl] Fix failure in otp_ctrl_mem_partial_access test

matutem:
Add call to `stop_transaction_generators` in the common run_mem_partial_access_vseq_sub sequence.


msfschaffner (approved):
Thanks for fixing this, Matute!

PR #21727, opened 2024-02-27
[top_earlgrey] Relocate DFT_STRAP0/1 pins

msfschaffner:
The DFT_STRAP pins are currently mapped onto muxed IOs that are allocated to the main UART console used in ROM.

This can be inconvenient when testing the chip in DFT unlocked life cycle states, and we hence remap those DFT_STRAP pins onto other muxed IOs that are not used in ROM during early boot.


msfschaffner:
@arnonsha @meisnere @sha-ron please let me know if this reallocation of the DFT straps works for you.

FYI @moidx @timothytrippel @cfrantz

a-will (approved):

timothytrippel (approved):
Thanks @msfschaffner ! I think once this is merged, we can remove the opentitanlib hack that sidestepped this issue: https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/opentitanlib/src/app/mod.rs;drc=025dd9f55ff027cf7c5d63b44f6658a028a35354;l=973 

We should leave it on the ES sival branch though.

msfschaffner:
> Thanks @msfschaffner ! I think once this is merged, we can remove the opentitanlib hack that sidestepped this issue: https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/opentitanlib/src/app/mod.rs;drc=025dd9f55ff027cf7c5d63b44f6658a028a35354;l=973
> 
> We should leave it on the ES sival branch though.

Sounds good! Can you take care of that part once this has been merged?

meisnere:
The suggested pads are open-drain while the original ones aren't.
Please switch to regular pads instead, maybe IOR5, IOR7?

moidx (approved):

PR #21731, opened 2024-02-28
Usbdev invalid data1 data0 toggle test sequence

Nadia-Hanif:
This draft PR contains invalid data1 data0 toggle test sequence.


Nadia-Hanif:
@alees24  Kindly review this. Thanks.

rswarbrick:
Hi there. This seems to have 3 commits in it with different sequences. Could you re-push with just the sequence you're intending to add?

rswarbrick (changes requested):
Please could you tidy up the commits here? At the moment, I think we're adding some repeated include items in `usbdev_vseq_list.sv` and we're also adding a `usbdev_enable` test in `usbdev_sim_cfg.hjson`.

hw/ip/usbdev/dv/env/seq_lib/usbdev_vseq_list.sv
```diff
@@ -15,6 +15,10 @@
 `include "usbdev_in_trans_vseq.sv"
 `include "usbdev_in_iso_vseq.sv"
 `include "usbdev_nak_trans_vseq.sv"
+`include "usbdev_invalid_data1_data0_toggle_test_vseq.sv"
+`include "usbdev_random_length_out_transaction_vseq.sv"
+`include "usbdev_min_length_out_transaction_vseq.sv"
+`include "usbdev_max_length_out_transaction_vseq.sv"
```
rswarbrick, line 21: Please could you tidy this up? These are not related to the code you're trying to add.

hw/ip/usbdev/dv/env/seq_lib/usbdev_invalid_data1_data0_toggle_test_vseq.sv
```diff
@@ -0,0 +1,33 @@
...
+    for (int i = 0; i < 6; i++) begin
+      call_token_seq(PidTypeOutToken);
+      inter_packet_delay();
+      call_data_seq(PidTypeData0, .randomize_length(1'b1), .num_of_bytes(0));
+      get_response(m_response_item);
+      $cast(m_usb20_item, m_response_item);
+      get_out_response_from_device(m_usb20_item, PidTypeAck);
+    end
+    // Read link_out_err interrupt
+    csr_rd(.ptr(ral.intr_state.link_out_err), .value(link_out_err));
```
rswarbrick, line 28: Rather than using `intr_state`, could you take a look at the interrupt itself? I think it should be in `intr_vif`

hw/ip/usbdev/dv/env/seq_lib/usbdev_invalid_data1_data0_toggle_test_vseq.sv
```diff
@@ -0,0 +1,33 @@
...
+    // Enable link_out_err interrupt
+    csr_wr(.ptr(ral.intr_enable.link_out_err), .value(1'b1));
+    // Out token packet followed by a data packet
+    for (int i = 0; i < 6; i++) begin
+      call_token_seq(PidTypeOutToken);
+      inter_packet_delay();
+      call_data_seq(PidTypeData0, .randomize_length(1'b1), .num_of_bytes(0));
+      get_response(m_response_item);
+      $cast(m_usb20_item, m_response_item);
+      get_out_response_from_device(m_usb20_item, PidTypeAck);
```
rswarbrick, line 25: Don't you need an inter-packet delay between a Data0 packet and the following token packet?

hw/ip/usbdev/dv/env/seq_lib/usbdev_invalid_data1_data0_toggle_test_vseq.sv
```diff
@@ -0,0 +1,33 @@
...
+
+  task body();
+    bit link_out_err;
+
+    // Configure out transaction
+    configure_out_trans();
+    // Enable link_out_err interrupt
+    csr_wr(.ptr(ral.intr_enable.link_out_err), .value(1'b1));
+    // Out token packet followed by a data packet
+    for (int i = 0; i < 6; i++) begin
```
rswarbrick, line 19: Why is 7 iterations the magic value here? (It might be right, but maybe it needs some explanation?)

Won't we get an interrupt after only 2?

rswarbrick (changes requested):

hw/ip/usbdev/dv/env/seq_lib/usbdev_invalid_data1_data0_toggle_test_vseq.sv
Nadia-Hanif: i have just added random number 6 here.

hw/ip/usbdev/dv/env/seq_lib/usbdev_invalid_data1_data0_toggle_test_vseq.sv
Nadia-Hanif: i have added inter_packet_delay between token and data packet.

rswarbrick:
Please can you take a look at the results from CI? At the moment, CI is failing. (I just checked, and the results should be visible without any extra credentials)

rswarbrick (changes requested):
I just tried to run this test locally after rebasing and it fails with every seed. Please could you do some debugging work? The errors look like the scoreboard is calling the packetiser's `pack_pkt` with a pid of `0x38`. The bottom two bits are zero, which means "PID Type" is special (in fact, I think this PID is "SPLIT").

I don't think this is supposed to be what happens.

Nadia-Hanif:
> I just tried to run this test locally after rebasing and it fails with every seed. Please could you do some debugging work? The errors look like the scoreboard is calling the packetiser's `pack_pkt` with a pid of `0x38`. The bottom two bits are zero, which means "PID Type" is special (in fact, I think this PID is "SPLIT").
> 
> I don't think this is supposed to be what happens.

this sequence is failing after changes done in "pack_pkt task" in packetiser because of that default statement.

elliotb-lowrisc:
Closing as no longer relevant

PR #21738, opened 2024-02-28
[dv] fix rom_ctrl dvsim output path

nbdd0121:
Currently it uses the default output path, which does not have the variant in it. This causes the files of ROM_CTRL/32KB run to be overwritten by ROM_CTRL/64KB run silently, and will cause error if parallelism > 3.

Follow what other blocks are doing and specify an output path with variant name in it.


rswarbrick (approved):

andreaskurth (approved):
Thanks for debugging and fixing this, @nbdd0121 :+1:

PR #21739, opened 2024-02-28
[dv/otp_ctrl] increased spinwait timeout duration for otp_ctrl_parallel_lc_req_vseq

dkabely-ot:
Consistency and integrity checks performed by `otp_ctrl_smoke` in parallel to LC state transition requests are halted until the LC partition is fully programmed, which might cause the [spinwait on OTP_CTRL.STATUS.CHECK_PENDING](https://github.com/lowRISC/opentitan/blob/master/hw/ip/otp_ctrl/dv/env/seq_lib/otp_ctrl_base_vseq.sv#L463) to raise false negative timeout due to short duration of the default spinwait timeout, as real OTP PROG operations take much longer in reality then the generic memory array in the OS implementation. Hence, this request scales up the default spinwait timeout duration to solves the issue.


eshapira (approved):

dkabely-ot:
Hi @msfschaffner,

Can you please help with the CI failures?

msfschaffner (approved):
Thanks! The ROM CI failure is unrelated and has been fixed separately, so we can merge this.

PR #21740, opened 2024-02-28
[otp] Fix OWNER_SW_CFG_ROM_ALERT_DIGEST_PROD_END value

Razer6:
It seems something messed up for the values. I just re-generated the digest values and got the diff.
Thats the output of the opentitantool:
```
{
  partitions: [
    {
      name: "OWNER_SW_CFG",
      items: [
        {
          name: "OWNER_SW_CFG_ROM_ALERT_DIGEST_PROD",
          value: 3837441892
        },
        {
          name: "OWNER_SW_CFG_ROM_ALERT_DIGEST_PROD_END",
          value: 284251105
        },
        {
          name: "OWNER_SW_CFG_ROM_ALERT_DIGEST_DEV",
          value: 2316472459
        },
        {
          name: "OWNER_SW_CFG_ROM_ALERT_DIGEST_RMA",
          value: 1321476032
        }
      ]
    }
  ]
}
```

### DIFF
--- a/hw/ip/otp_ctrl/data/otp_ctrl_img_owner_sw_cfg.hjson
+++ b/hw/ip/otp_ctrl/data/otp_ctrl_img_owner_sw_cfg.hjson
@@ -233,7 +233,7 @@
                 },
                 {
                     name: "OWNER_SW_CFG_ROM_ALERT_DIGEST_PROD_END",
-                    value: "0x8a12908b",
+                    value: "0x10f153e1",
                 },
                 {
                     name: "OWNER_SW_CFG_ROM_ALERT_DIGEST_RMA",


andreaskurth (approved):
Thanks for this fix, @Razer6. I cross-checked the changed value with OTT and it matches.

andreaskurth:
Brief explanation why this didn't get caught by CI: These OTP values get checked in ROM code, and ROM end-to-end tests in simulation are too slow for CI. If we could run it on FPGA on Darjeeling, it would be fast enough for CI. In simulation, it can only be tested periodically ('nightly').

Another way to catch this problem in CI in the future would be to add a script that runs `opentitantool` and checks its output against the values in `hw/ip/otp_ctrl/data/otp_ctrl_img_owner_sw_cfg.hjson`. Maybe worth doing, @Razer6?

Razer6:
Thanks for double-checking. We also run such a test at our end to confirm this, i.e., configured the alert handler from the OTP config and then computed the digest over it.

> Another way to catch this problem in CI in the future would be to add a script that runs `opentitantool` and checks its output against the values in `hw/ip/otp_ctrl/data/otp_ctrl_img_owner_sw_cfg.hjson`. Maybe worth doing, @Razer6?

Yes such a tool might indeed be a good idea.

PR #21741, opened 2024-02-28
[pentest] Add support for AES SCA measurements on chip

nasahlpa:
On the FPGA, the AES block automatically raises the trigger allowing to perform SCA analysis. As this is not available on the chip, this PR extends the AES SCA code to manually handling the trigger.

The arming of the trigger is implemented in the `sca_call_and_sleep_trigger` to keep the trigger window as narrow as possible (this reduces the number of samples needed and helps to easier setup the SCA gear).

The host code is located in lowRISC/ot-sca#329.

### DIFF
--- a/sw/device/sca/lib/sca.c
+++ b/sw/device/sca/lib/sca.c
@@ -309,7 +309,8 @@ void sca_set_trigger_low(void) {
   OT_DISCARD(dif_gpio_write(&gpio, trigger_bit_index, false));
 }
 
-void sca_call_and_sleep(sca_callee callee, uint32_t sleep_cycles) {
+void sca_call_and_sleep(sca_callee callee, uint32_t sleep_cycles,
+                        bool sw_trigger) {
   // Disable the IO_DIV4_PERI clock to reduce noise during the actual capture.
   // This also disables the UART(s) and GPIO modules required for
   // communication with the scope. Therefore, it has to be re-enabled after
@@ -328,7 +329,15 @@ void sca_call_and_sleep(sca_callee callee, uint32_t sleep_cycles) {
   OT_DISCARD(dif_rv_timer_counter_set_enabled(&timer, kRvTimerHart,
                                               kDifToggleEnabled));
 
+  if (sw_trigger) {
+    sca_set_trigger_high();
+  }
+
   callee();
+
+  if (sw_trigger) {
+    sca_set_trigger_low();
+  }
 
   wait_for_interrupt();
 
--- a/sw/device/sca/aes_serial.c
+++ b/sw/device/sca/aes_serial.c
@@ -245,7 +245,7 @@ static void aes_encrypt(const uint8_t *plaintext, size_t plaintext_len) {
   // Using the SecAesStartTriggerDelay hardware parameter, the AES unit is
   // configured to start operation 40 cycles after receiving the start trigger.
   // This allows Ibex to go to sleep in order to not disturb the capture.
-  sca_call_and_sleep(aes_manual_trigger, kIbexAesSleepCycles);
+  sca_call_and_sleep(aes_manual_trigger, kIbexAesSleepCycles, false);
 }
 
 /**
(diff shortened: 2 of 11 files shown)


sw/device/sca/lib/sca.c
```diff
@@ -337,6 +337,35 @@ void sca_call_and_sleep(sca_callee callee, uint32_t sleep_cycles) {
...
+  // Start timer to wake Ibex after the callee is done.
+  uint64_t current_time;
+  // Return values of below functions are ignored to improve capture
+  // performance.
+  OT_DISCARD(dif_rv_timer_counter_read(&timer, kRvTimerHart, &current_time));
+  OT_DISCARD(dif_rv_timer_arm(&timer, kRvTimerHart, kRvTimerComparator,
+                              current_time + sleep_cycles));
+  OT_DISCARD(dif_rv_timer_counter_set_enabled(&timer, kRvTimerHart,
+                                              kDifToggleEnabled));
+  sca_set_trigger_high();
```
vogelpi, line 358: As far as I can tell, this function is identical to `sca_call_and_sleep()` except for the calls to the trigger functions. To reduce maintenance overhead, would it be better to extend the existing function with an argument to decide whether the trigger needs to be handled? Or would the branch this involves negatively impact the measurement?

vogelpi (approved):
LGTM with one question. Maybe it would be better to extend the existing function instead of adding a new one to reduce code duplication but I might have overseen something.

sw/device/sca/lib/sca.c
nasahlpa: Thanks for the comment. Done!

vogelpi: Nice, thanks!

vogelpi:
`//sw/device/tests/crypto:drbg_functest_fpga_cw310_sival_rom_ext` keeps failing for this PR. I believe it shouldn't be related to this PR but since it keeps failing, it might be worth investigating. Question: is this code here ending up in the same firmware binary as `drbg_functest_fpga_cw310_sival_rom_ext` or is it completely different/unrelated?

nasahlpa:
> `//sw/device/tests/crypto:drbg_functest_fpga_cw310_sival_rom_ext` keeps failing for this PR. I believe it shouldn't be related to this PR but since it keeps failing, it might be worth investigating. Question: is this code here ending up in the same firmware binary as `drbg_functest_fpga_cw310_sival_rom_ext` or is it completely different/unrelated?

As CI in other PRs  (e.g. #21911) also seem to fail at the same test, I assume it is not related to my changes. But I will have a look.

sw/device/sca/lib/sca.c
```diff
@@ -309,7 +309,8 @@ void sca_set_trigger_low(void) {
   OT_DISCARD(dif_gpio_write(&gpio, trigger_bit_index, false));
 }
 
-void sca_call_and_sleep(sca_callee callee, uint32_t sleep_cycles) {
+void sca_call_and_sleep(sca_callee callee, uint32_t sleep_cycles,
+                        bool trigger) {
```
johannheyszl, line 313: nit: maybe "sw_trigger" instead of "trigger" for readability; but not mandatory

johannheyszl:
Thanks @nasahlpa .
As noted in https://github.com/lowRISC/ot-sca/pull/329 I wonder if we could avoid the case separation alltogether, but if not it looks good

sw/device/sca/lib/sca.c
nasahlpa: Good point, I've renamed the parameter.

nasahlpa:
> Thanks @nasahlpa . As noted in [lowRISC/ot-sca#329](https://github.com/lowRISC/ot-sca/pull/329) I wonder if we could avoid the case separation alltogether, but if not it looks good

Thanks for the review! The SCA library, where the trigger is managed, does not support avoiding the case separation. We could do this in the future but for now I would prefer having this trigger selection.

nasahlpa:
> `//sw/device/tests/crypto:drbg_functest_fpga_cw310_sival_rom_ext` keeps failing for this PR. I believe it shouldn't be related to this PR but since it keeps failing, it might be worth investigating. Question: is this code here ending up in the same firmware binary as `drbg_functest_fpga_cw310_sival_rom_ext` or is it completely different/unrelated?

CI now passes - it seems that this was a CI internal issue.

Issue #21716, opened 2024-02-27
[rv_dm] Address Verilator lint warnings


rswarbrick:
### Description

The nightly lint check for rv_dm contains 5 warnings.

It looks like 4 of these are width bugs in the vendored code. No big deal, but probably worth waiving in a .vlt file. There's also an UNOPTFLAT warning about `tlul_lc_gate.sv`, which we wrote. Address it.

See https://reports.opentitan.org/hw/ip/rv_dm/lint/verilator/latest/report.html for the latest report.

rswarbrick:
The two PRs linked in above should fix this. I don't think I know how to use PR links to say "please close this issue when both PRs have been merged", so we might have to do it manually... :-)

rswarbrick:
The remaining portion of this is fixed by #22486 (but I don't seem to be able to create the links through GitHub that would put this in the "development" box on the right)

andreaskurth:
Closing as completed by #21736 and #22486.

PR #21742, opened 2024-02-28
[tlul,lint] Tell Verilator to split a process to avoid an UNOPTFLAT

rswarbrick:
This fixes the "non-vendored half" of #21716.


rswarbrick:
Thanks for the review, @andreaskurth. I was being a bit lazy with my comment! I've updated it to be (I think) more precise now. Does this make a bit more sense?

PR #21743, opened 2024-02-28
Testplan formatter

engdoreis:
Auto format the tests plans using the python hjson library.
The script reads the hjson file to a python `dict` and write it back to guarantee a standard format.
The downside is that we lose comments.

```
 python3 ./util/lint_testplan.py --formatting-fix  --dir hw/top_earlgrey/data/ip/
```

### Note
The first commit contains the script and the second commit contains the testplans formatted by the script.

### DIFF
--- a/hw/top_earlgrey/data/ip/chip_adc_ctrl_testplan.hjson
+++ b/hw/top_earlgrey/data/ip/chip_adc_ctrl_testplan.hjson
@@ -4,10 +4,10 @@
 {
   name: chip_adc_ctrl
   testpoints: [
-    // ADC_CTRL (pre-verified IP) integration tests:
     {
       name: chip_sw_adc_ctrl_debug_cable_irq
-      desc: '''Verify that the ADC correctly detects the voltage level programmed for each channel.
+      desc: '''
+            Verify that the ADC correctly detects the voltage level programmed for each channel.
 
             - Program both ADC channels to detect mutually exclusive range of voltages. Setting only
               one filter CSR is sufficient.
@@ -24,7 +24,7 @@
               filter caused the interrupt to fire. Read the ADC channel value register to verify the
               correctness of the detected value that was forced in the AST for each channel.
             '''
-      features:["ADC_CTRL.MODE.LOW_POWER"]
+      features: ["ADC_CTRL.MODE.LOW_POWER"]
       stage: V2
       si_stage: SV3
       lc_states: ["PROD"]
@@ -32,7 +32,8 @@
     }
     {
       name: chip_sw_adc_ctrl_sleep_debug_cable_wakeup
-      desc: '''Verify that in deep sleep, ADC ctrl can signal the ADC within the AST to power down.
+      desc: '''
+            Verify that in deep sleep, ADC ctrl can signal the ADC within the AST to power down.
 
             - Read the reset cause register in rstmgr to confirm we are in POR reset phase.
             - Follow the same steps as chip_adc_ctrl_debug_cable_irq, but instead of programming the
@@ -49,7 +50,7 @@
               that was forced in the AST.
             - Repeat for both ADC channels.
             '''
-      features:["ADC_CTRL.MODE.LOW_POWER"]
+      features: ["ADC_CTRL.MODE.LOW_POWER"]
       stage: V2
       si_stage: SV3
       lc_states: ["PROD"]
@@ -57,8 +58,8 @@
     }
     {
       name: chip_sw_adc_ctrl_normal
-      desc:"Verifiy that ADC can sample from channel 0 and 1 in normal mode."
-      features:["ADC_CTRL.MODE.NORMAL"]
+      desc: Verifiy that ADC can sample from channel 0 and 1 in normal mode.
+      features: ["ADC_CTRL.MODE.NORMAL"]
       stage: V3
       si_stage: SV2
       lc_states: ["PROD"]
@@ -66,8 +67,8 @@
     }
     {
       name: chip_sw_adc_ctrl_oneshot
-      desc:"Verifiy that ADC can sample from channel 0 and 1 in normal mode with oneshot filter config."
-      features:["ADC_CTRL.ONESHOT", "ADC_CTRL.MODE.NORMAL"]
+      desc: Verifiy that ADC can sample from channel 0 and 1 in normal mode with oneshot filter config.
+      features: ["ADC_CTRL.ONESHOT", "ADC_CTRL.MODE.NORMAL"]
       stage: V3
       si_stage: SV3
       lc_states: ["PROD"]
--- a/hw/top_earlgrey/data/ip/chip_aon_timer_testplan.hjson
+++ b/hw/top_earlgrey/data/ip/chip_aon_timer_testplan.hjson
@@ -6,7 +6,8 @@
   testpoints: [
     {
       name: chip_sw_aon_timer_wakeup_irq
-      desc: '''Verify the AON timer wake up interrupt in normal operating state.
+      desc: '''
+            Verify the AON timer wake up interrupt in normal operating state.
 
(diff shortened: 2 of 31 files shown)


hw/top_earlgrey/data/ip/chip_adc_ctrl_testplan.hjson
```diff
@@ -3,71 +3,94 @@
 // SPDX-License-Identifier: Apache-2.0
 {
   name: chip_adc_ctrl
-  testpoints: [
-    // ADC_CTRL (pre-verified IP) integration tests:
```
HU90m, line 7: Might we want to be able to have comments?

hw/top_earlgrey/data/ip/chip_adc_ctrl_testplan.hjson
```diff
@@ -3,71 +3,94 @@
...
+        ADC_CTRL.ONESHOT
+        ADC_CTRL.MODE.NORMAL
+      ]
       stage: V3
       si_stage: SV3
       tests: []
     }
   ]
-}
+}
```
HU90m, line 96: The formatted files are missing a final newline

hw/top_earlgrey/data/ip/chip_aon_timer_testplan.hjson
```diff
@@ -3,137 +3,254 @@
...
+        AON_TIMER.WAKEUP.INTERRUPT
+      ]
+      tests:
+      [
+        chip_sw_aon_timer_irq
+      ]
+      bazel:
+      [
+        "//sw/device/tests:aon_timer_irq_test"
+      ]
```
HU90m, line 39: I'm not sure I prefer this formatting

HU90m (changes requested):
This would be lovely to have

hw/top_earlgrey/data/ip/chip_aon_timer_testplan.hjson
pamaury: I also find this formatting to be very verbose/hard to read, is there a way to control the formatter ?

hw/top_earlgrey/data/ip/chip_adc_ctrl_testplan.hjson
pamaury: Loosing comments seems like a big problem, do we have many of them? Are there no parsers that can save the comments?

hw/top_earlgrey/data/ip/chip_aon_timer_testplan.hjson
engdoreis: I also don't like it, and there's no way to control it. I'll have to write a simple serializer for the python `dict` I think. I'll investigate.

engdoreis: I changed the formatting to be very similar to the one mostly used.

hw/top_earlgrey/data/ip/chip_adc_ctrl_testplan.hjson
engdoreis: I still need to find a solution for the comments.

hw/top_earlgrey/data/ip/chip_edn_testplan.hjson
```diff
@@ -117,18 +110,17 @@
...
-      features: [
-        EDN.MODE.SW
-      ]
+      features: ["EDN.MODE.SW"]
       stage: V3
       si_stage: SV3
       lc_states: ["PROD"]
       tests: ["chip_sw_edn_sw_mode"]
       bazel: ["//sw/device/tests:edn_sw_mode"]
-    },
```
engdoreis, line 128: Should we keep the comma?

hw/top_earlgrey/data/ip/chip_spi_device_testplan.hjson
```diff
@@ -18,20 +17,15 @@
...
-      lc_states: [ "PROD" ]
-      features: [
-        "SPI_DEVICE.MODE.FLASH_EMULATION",
-        "SPI_DEVICE.MODE.FLASH_EMULATION.COMMANDS",
-        "SPI_DEVICE.MODE.FLASH_EMULATION.READ_COMMAND_PROCESSOR",
-        "SPI_DEVICE.HW.FLASH_EMULATION_BLOCKS",
-        "SPI_DEVICE.HW.LAST_READ_ADDR",
-      ]
+      lc_states: ["PROD"]
+      features: ["SPI_DEVICE.MODE.FLASH_EMULATION", "SPI_DEVICE.MODE.FLASH_EMULATION.COMMANDS", "SPI_DEVICE.MODE.FLASH_EMULATION.READ_COMMAND_PROCESSOR", "SPI_DEVICE.HW.FLASH_EMULATION_BLOCKS", "SPI_DEVICE.HW.LAST_READ_ADDR"]
```
HU90m, line 21: I'm not sure this is an improvement

hw/top_earlgrey/data/ip/chip_spi_device_testplan.hjson
```diff
@@ -159,10 +138,11 @@
                Notes for silicon targets:
                This test currently does not exercise any features of the spi_device block itself,
                which is why the "features" list is empty. It may be reclassifed in the future.
-             '''
+
+            '''
       stage: V3
       si_stage: None
-      lc_states: [ "PROD" ]
+      lc_states: ["PROD"]
```
HU90m, line 145: :heart:

hw/top_earlgrey/data/ip/chip_spi_host_testplan.hjson
```diff
@@ -4,10 +4,10 @@
 {
   name: chip_spi_host
   testpoints: [
-    // SPI_HOST (pre-verified IP) integration tests:
     {
       name: chip_sw_spi_host_tx_rx
-      desc: '''Verify the transmission of data on the chip's SPI host port.
+      desc: '''
+            Verify the transmission of data on the chip's SPI host port.
```
HU90m, line 10: :heart:

hw/top_earlgrey/data/ip/chip_sram_ctrl_testplan.hjson
```diff
@@ -139,7 +134,8 @@
              - Read the location from the SRAM and check that the data has been wiped.
              - Repeat the steps above for the retention SRAM.
 
-             '''
+             
+            '''
```
HU90m, line 138: It's nice it spotted the incorrect indenting, but a shame it has to keep the same semantics so it has to add a newline

hw/top_earlgrey/data/ip/chip_uart_testplan.hjson
```diff
@@ -160,7 +168,6 @@
       features: ["UART.FIFO_INTERRUPTS"]
       tests: ["chip_sw_uart_tx_rx"]
       bazel: ["//sw/device/tests:uart_tx_rx_test"]
-    },
-
+    }
```
HU90m, line 171: nit: I quite like trailing commas

HU90m (commented):
hjson is a very flexible format. Creating formatter is far from trivial. I'm not sure whether this is worth the development and maintenance time. I think the more important thing is to verify the fields are correct.

That being said, this formatter works OK already and I wouldn't be strongly oppose it going in as is

hw/top_earlgrey/data/ip/chip_uart_testplan.hjson
engdoreis: Let's keep it, though.

engdoreis:
After some discussion with the Lowrisc SW team, we decided to don't proceed with a formatter for the testplans as hjson is too loose, and the maintenance cost may not be worth.

PR #21744, opened 2024-02-28
[kmac] Fix driving conflict in kmac_reduced.sv used for SCA evaluation

vogelpi:
This should fix the driving conflict mentioned by @meisnere. Thanks again for bringing this up Eran and sorry for the inconvenience.

It's surprising that none of the tools I've used flagged this. Maybe synthesis was smart and realized that both packers are always in sync and then optimized away the redundant control signals.

### DIFF
--- a/hw/ip/kmac/rtl/kmac_reduced.sv
+++ b/hw/ip/kmac/rtl/kmac_reduced.sv
@@ -99,6 +99,8 @@ module kmac_reduced
   ///////////////////////////////////
   // Message packer FIFO
   logic [sha3_pkg::MsgWidth-1:0] msg [NumShares];
+  logic [NumShares-1:0] msg_valid_shares;
+  logic [NumShares-1:0] msg_ready_shares;
   logic msg_valid, msg_ready;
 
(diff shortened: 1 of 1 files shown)


a-will:
Would it make sense for kmac_reduced.sv to be in a separate fusesoc core? It kind of looks like there are 2 top-level SV files for one core.

Anyway, that isn't relevant for this PR; it's just a thought prodded by the discussion.

andreaskurth (approved):

vogelpi:
> Would it make sense for kmac_reduced.sv to be in a separate fusesoc core? It kind of looks like there are 2 top-level SV files for one core.
> 
> Anyway, that isn't relevant for this PR; it's just a thought prodded by the discussion.

That's a good idea and it was easier to do this than I thought. I've now added a second commit for this change. Thanks for the suggestion!

vogelpi:
I am merging this to not unnecessarily occupy CI resources. The files touched by this PR are not relevant for any of the CW310 ROM / SiVal ROM / SiVal non-ROM_EXT / CW310 Test ROM tests.

PR #21749, opened 2024-02-29
[RV_DM]rv_dm_jtag_dmi_debug_disabled_vseq

BasitMehmood:
This commit is to make some change in vseq that is failing in regression. I just increase the wait upto 2 clock edges to has effect of LC signal. Previously it was 1 clock edge.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_jtag_dmi_debug_disabled_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_jtag_dmi_debug_disabled_vseq.sv
@@ -69,8 +69,8 @@ class rv_dm_jtag_dmi_debug_disabled_vseq extends rv_dm_base_vseq;
       cfg.m_jtag_agent_cfg.vif.do_trst_n(2);
       cfg.rv_dm_vif.cb.lc_hw_debug_en <= lc_ctrl_pkg::On;
 
-      // Wait a clock edge to make sure the LC signal has an effect
-      cfg.clk_rst_vif.wait_clks(1);
+      // Wait atleast two clock edges to make sure the LC signal has an effect
+      cfg.clk_rst_vif.wait_clks(2);
 
       // Read the contents of abstractdata[0] and check they are what we set at the start.
       read_abstractdata(value0);


BasitMehmood:
@rswarbrick

BasitMehmood:
> Hmm, that's surprising! It looks like my [331ca9e](https://github.com/lowRISC/opentitan/commit/331ca9e1ea900c55608a7a636337c2951e8c96c5) broke things here. I'm not keen on unexplained magic waits though :-) Could you run this with waves and see why a single cycle wait wasn't enough? Presumably, something is taking a while to flow through. But it would be good to see exactly what's happening.
> 
> If you're struggling to see what's going on, maybe you could give an SV seed that is making it fail? (Or maybe it fails for most seeds?)

I have given that seed on which it was failing. Now after changes when I checked on that seed, it is passing.

rswarbrick:
I'm sorry but I don't completely understand what you just wrote. Do you mean that this test fails for a seed and then passes when you make the change in this PR?

If so, what is the seed?

BasitMehmood:
> I'm sorry but I don't completely understand what you just wrote. Do you mean that this test fails for a seed and then passes when you make the change in this PR?
> 
> If so, what is the seed?

No . It was failing without giving any seed. But when i change the wait to 2 then it was passing. Then I enter the seed that was given in regression failure bucket. So after changes it was passing.

BasitMehmood:
So basically it was failing with and without giving any seed. When i change the wait to 2 then it was passing. Then I also checked with seed and it was passing with that seed also.

rswarbrick:
I've just cloned the head of this PR and ran 100 seeds (with the 2 cycle wait). I get a 50% failure rate. The pass rate is similarly middling without this commit. I think you might need to do a bit more "root causing" to figure out what change needs to be made.

BasitMehmood:
> I've just cloned the head of this PR and ran 100 seeds (with the 2 cycle wait). I get a 50% failure rate. The pass rate is similarly middling without this commit. I think you might need to do a bit more "root causing" to figure out what change needs to be made.

Ok sure.

BasitMehmood:
> I've just cloned the head of this PR and ran 100 seeds (with the 2 cycle wait). I get a 50% failure rate. The pass rate is similarly middling without this commit. I think you might need to do a bit more "root causing" to figure out what change needs to be made.

I ran this sequence without/dropping repeat loop with 2 cycle wait and it is passing. But with the repeat loop it is failing.
In repeat loop, when it repeats second time rdata and expected data does not match. So it fails on that check.

PR #21752, opened 2024-02-29
[crypto] Check for zero in P256 modular inverse.

jadephilipoom:
A zero input (such as a point-at-infinity in the conversion to affine coordinates) can cause the variable-time modular inverse algorithm to run forever.

@wettermo I looked at the P384 code and it seems not to be a problem there because we use only the constant-time modular inverse for code size reasons, but let me know if I missed something.

Thanks to @RyanTorok for finding this bug via the wycheproof tests!

### DIFF
--- a/sw/otbn/crypto/p256_verify.s
+++ b/sw/otbn/crypto/p256_verify.s
@@ -278,6 +278,9 @@ p256_verify:
  * show a data dependent timing and execution profile. Only use in situations
  * where a full white-box environment is acceptable.
  *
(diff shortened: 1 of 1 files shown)


wettermo (approved):
LGTM!

Yes exactly! To reduce code size in the P384 code, only the time constant modular inverse is used. It just uses a square and multiply loop with a fixed number of iterations, so a zero input shouldn't be a problem.

moidx (approved):

PR #21580, opened 2024-02-20
[ci] Avoid running duplicate hyperdebug-based tests

pamaury:
At the moment, the sival job runs only `cw310_sival` jobs but not `cw310_sival_rom_ext` which is not right. On the other hand, the hyperdebug job runs all `hyper310` tests and that includes a number of `cw310_sival` and `cw310_sival_rom_ext` tests so we run some tests twice.

This commit changes the sival job to run all `cw310_sival` and `cw310_sival_rom_ext` tests, and the hyperdebug runs only the hyperdebug tests that are not in sival (if any).

PR #20836, opened 2024-01-15
[ci] Try to use GCP Bazel cache for more jobs

jwnrt:
## Summary

This change lets more jobs _read_ from the cache, and also allows more jobs to _write_ to the cache when running for the `master` branch.

It's being tested in [this Azure run][azure-run] which will write to the cache. A second run will need to be triggered to see if we actually read from the cache.

## Details

This change:

1. Switches usage of `./bazelisk.sh` to `ci/bazelisk.sh` in pipelines.
   * This script connects Bazel to our GCP bucket containing a Bazel cache.
   * This allows CI to read from the cache.
   * If the `GCP_BAZEL_CACHE_KEY` variable is set, it also writes to the cache.
2. Adds downloads for the file containing cache write credentials to more jobs.
   * This file is stored encrypted in Azure and cannot be downloaded by PRs from forks.
   * The location to this file is automatically stored in the `GCP_BAZEL_CACHE_KEY_SECUREFILEPATH` environment variable.
   * The file is only downloaded for the `master` branch, however I have changed this for [this test run][azure-run] to show that the cache can be written to.
   * In summary: having this downloaded allows the master branch to write to the cache.

[azure-run]: https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=138706&view=results

Issue #8440, opened 2021-09-29
[tracking] Use ipgen to produce templated IPs


imphil:
The following IP blocks are actually templates at the moment and need to be converted to be using ipgen ([slides discussing motivation and general approach](https://docs.google.com/presentation/d/1x6YRgxdP1w9T-i0oApwelb782H4B5Pbyo4HnR6_Ei4s/edit?usp=sharing)).

- [x] alert_handler: https://github.com/lowRISC/opentitan/pull/8513
- [x] clkmgr
  - [x] Create ipgen files: https://github.com/lowRISC/opentitan/pull/21202
  - [x] Transition to ipgen files and remove old files: https://github.com/lowRISC/opentitan/pull/21321
- [x] flash_ctrl
  - [x] Create ipgen files: https://github.com/lowRISC/opentitan/pull/21170
  - [x] Transition to ipgen files and remove old files: https://github.com/lowRISC/opentitan/pull/22518
- [x] pinmux: https://github.com/lowRISC/opentitan/pull/24808
- [x] pwrmgr: https://github.com/lowRISC/opentitan/pull/19801
- [x] rstmgr: https://github.com/lowRISC/opentitan/pull/21041
- [x] rv_plic: #8431

When the list above is done
- [x] Remove topgen support for "templated" module "attr" attribute: https://github.com/lowRISC/opentitan/pull/25129
- [x] Remove topgen quirks due to templated path searches, as for countermeasures: https://github.com/lowRISC/opentitan/pull/25129

Blocks which were unclear, and their resolution:
- [x] ast: doesn't need generation
- [x] xbar: xbars have their own dedicated generation
- [x] xbar_main: xbars have their own dedicated generation
- [x] xbar_peri: xbars have their own dedicated generation

msfschaffner:
CC @alphan

GregAC:
Triaging for `flash_ctrl`. This is a significant chunk of work and not required for M2.5, however it is likely useful for integrated.

andreaskurth:
Triaged for `pwrmgr` - agree

GregAC:
Triaged for 'rstmgr'

GregAC:
Triaged for `pinmux` - Same conclusion as above

GregAC:
Triaged for `sensor_ctrl` - Same conclusion as above

a-will:
Should templated DIFs also accompany templated IPs? Or do we feel that the current mechanisms with C preprocessor macros is enough?

[...]

PR #21754, opened 2024-02-29
Backport ci changes

pamaury:
This is a backport of #21580, #21645 and #20836 but I had to fix every single commit with conflicts so please double-check carefully @nbdd0121 and @jwnrt 

I also re-enabled the rom and rom_test jobs but added a condition so that they only run on the master branch. The idea is that we can back-backport this change to master and avoid divergences between the azure files (thanks @nbdd0121 for the suggestion)


pamaury:
Just to help review: with this change, the only diff between master and earlgray_es_sival for the azure pipelines are:
- condition on rom and rom_ext dependent on branch -> will be backported to master
- branch and bucket for GCP -> different names on master and earlgrey_es_sival so that's fine
- changes introduced by #https://github.com/lowRISC/opentitan/issues/8440 related to ip_gen (probably not needed for  es_sival branch since RTL does not change?)

jwnrt (approved):
The changes here LGTM, good idea to conditionally re-enable those ROM FPGA tests. I'll have another look at the logs of the CI run once it's finished to make sure it's doing what we expect.

nbdd0121 (approved):

pamaury:
I had to do so more manual fixing because we have a couple more tests in es_sival and those still had `--bitstream`. I also early-backported part of #21757 to remove manual-tagged tests to avoid running pmod tests.

nbdd0121:
The log says the ROM job is doing bitstream build now

pamaury:
I added a commit to remove the `spi_device_ottf` test, see commit for details.

jwnrt (approved):
I think removing the SPI device OTTF console test is fine since we don't have it in master anymore.

PR #21757, opened 2024-02-29
[ci] Avoid running ROM tests that have a corresponding sival test

pamaury:
This PR reduces the number of ROM tests and favors the Sival environment for tests that exists both in ROM and SiVal.

### DIFF
--- a/ci/scripts/run-fpga-tests.sh
+++ b/ci/scripts/run-fpga-tests.sh
@@ -82,7 +82,42 @@ then
         )
     " \
     > "${pattern_file}"
-    test_args="${test_args} --test_tag_filters=cw310_sival,-broken,-skip_in_ci"
+    # We need to remove tests tagged as manual since we are not using a wildcard target.
+    test_args="${test_args} --test_tag_filters=cw310_sival,-broken,-skip_in_ci,-manual"
+elif [ "${fpga_tags}" == "cw310_rom_but_not_manuf_and_sival_tests" ]
+then
+    # Only consider tests that are tagged `cw310_rom_with_fake_keys` or `cw310_rom_with_real_keys`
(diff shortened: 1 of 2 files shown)


pamaury:
The CI failure is unrelated. This PR reduces the number of tests quite a bit: 297 instead of 375 previously. The time gain does not seem to be massive (maybe 5 to 10 minutes) but I think that's worth it.

ci/scripts/run-fpga-tests.sh
```diff
@@ -50,41 +50,66 @@ test_args=""
...
+            kind(
+                \"test_suite\",
+                //...
+            )
+            except
+            `# Remove all test suites depending on a test tagged cw310_sival_rom_ext`
+            rdeps(
+                //...,
+                `# Find all tests tagged cw310_sival_rom_ext`
+                attr(\"tags\",\"cw310_sival_rom_ext\", //...),
```
nbdd0121, line 75: should this check for broken?

pamaury: Hum, that's a good point, we might not want to ignore a test if the corresponding `cw310_sival_rom_ext` is marked as broken? I will update this.

moidx (approved):

pamaury:
Thanks @moidx for the comment, indeed I think that we should filter out `manual` tests because previously we were using a wildcard pattern but now we are listing targets explicitely. I will change that.

pamaury:
Can we merge this?

pamaury:
> Reference in new iss

I agree, although I think this will be hard to achieve with our current approach. I was thinking that maybe a better approach would be to have a "test dispatcher": a small program (python) that queries all the tests and assign them to the various test job in the CI. This would make it much easier to make sure that every test is dispatched and that there is minimal redundancy between execution environments.

Issue #18492, opened 2023-05-10
[i2c] Clock speed mismatch


johngt:
Promoting this issue as this was captured in [I2C test plan](https://github.com/lowRISC/ot-discrete-eg/issues/11) previously.
During testing of power monitor, a mismatch between the expected clock timings and those captured has been observed.

This was using a USB Logic analyser. Possible that there is a setup or test issue. Capturing here for more full discussion.

marnovandermaas:
My understanding of this issue is that the timing parameters are calculated in a conservative way. So if you say the bus clock should be 100 kHz it gives you the timing parameters that make sure you do not *exceed* this frequency. We need to document this discrepancy.

a-will:
https://github.com/lowRISC/ot-discrete-eg/issues/11#issuecomment-1575126104 explains where some of the mismatch with expectation comes from.

In short, the round-trip latency of ~3 cycles for the IP's output change to observing responses adds to `thigh` time. Due to the much larger peripheral clock period on the FPGA, the effect here is much more noticeable.

Why is there a round-trip? It's because the I2C IP must monitor for clock stretching, so it needs to wait to see SCL go high before counting out high time.

moidx:
Moving to M2.5.3 as this is something that we should capture in the programming guide as part of the datasheet release.

marnovandermaas:
The documentation of the speed mismatches that I am aware of are captured in this PR: https://github.com/lowRISC/opentitan/pull/18964

msfschaffner:
@jesultra @a-will what is your opinion on this discrepancy - are we ok with the documented behavior, or do we need to optimize this behavior for PROD?

a-will:
I'd recommend at least getting rid of the double-counted t_fall, since that is just a quick parameter update. From https://github.com/lowRISC/opentitan/issues/18958

I forget if the problems were more than that + the 3 added cycles to t_high from needing to detect clock stretching (https://github.com/lowRISC/opentitan/issues/18962), but it's a considerable reduction in speed for fast-plus if both effects are present.

For a back-of-the-envelope calculation for fast-plus, we take 24 MHz and target a minimum of _0.26 us_ t_high, _0.5 us_ t_low, and _0.12 us_ for t_rise and t_fall. Since we'll have to round up all factors, that leaves us with counts of 7, 12, 3, and 3 cycles, respectively, resulting in a rate of 24 MHz / (7 + 12 + 3 + 3) = 960 kHz.

[...]

PR #21765, opened 2024-02-29
[i2c,rtl] Remove double-counted t_f

### LINKED ISSUE #18958: [i2c,rtl] Fall time counted twice
The fall time in the I2C FSM is counted twice for the host clock pulse:
https://github.com/lowRISC/opentitan/blob/7545bb6eac5207c29c5413fa596b9e219403f200/hw/ip/i2c/rtl/i2c_fsm.sv#L161-L162

This issue was raised in this comment: https://github.com/lowRISC/ot-discrete-eg/issues/11#issuecomment-1536763517 (restricted access).

hcallahan-lowrisc:
As titled, basically.

This change caused a lot of breakage in the i2c block level DV, which I have not been able to fully mitigate.
This is primarily around breaking the START/RSTART detection routines in the monitor, which causes the monitor to lose it's lock 
on the bus traffic.
For now, I have weakened the monitor checks for the correct Setup/Holds around these events, which mostly fixes the breakage.
There is still one test that fails systematically with this change, `i2c_host_rx_oversample`.
(On a related note, this test seems to be misnamed, as the testplan description doesn't gel with its implementation IMO. It could be a candidate for removal)

I proposed merging this RTL change with the DV mitigations for now, and removing the broken test from the regression for now. Then to create an action item to come back and fix it properly later.

Closes #18958
Goes towards #18492

### DIFF
--- a/hw/dv/sv/i2c_agent/i2c_if.sv
+++ b/hw/dv/sv/i2c_agent/i2c_if.sv
@@ -64,16 +64,21 @@ interface i2c_if(
     repeat (dly) @(posedge clk_i);
   endtask : wait_for_dly
 
+  // TODO(#21887) Re-strengthen checks when detecting S/Sr/P conditions on the bus
+  // Currently these monitor tasks observe the derived timing parameters before
+  // checking for the expected bus behaviour.
+  // These delays were disabled as removal of some extra slack in the system
+  // caused the tasks to cease functioning reliably, and the monitor would lose its
+  // lock on the bus traffic.
+
   task automatic wait_for_host_start(ref timing_cfg_t tc);
     forever begin
       @(negedge sda_i);
-      if (scl_i) begin
-      wait_for_dly(tc.tHoldStart);
-      end else continue;
+      if (!scl_i) continue;
       @(negedge scl_i);
       if (!sda_i) begin
-      wait_for_dly(tc.tClockStart);
-      break;
+        // wait_for_dly(tc.tClockStart);
+        break;
       end else continue;
     end
   endtask: wait_for_host_start
@@ -83,10 +88,10 @@ interface i2c_if(
     rstart = 1'b0;
     forever begin
       @(posedge scl_i && sda_i);
-      wait_for_dly(tc.tSetupStart);
+      // wait_for_dly(tc.tSetupStart);
       @(negedge sda_i);
       if (scl_i) begin
-        wait_for_dly(tc.tHoldStart);
+        // wait_for_dly(tc.tHoldStart);
         @(negedge scl_i) begin
           rstart = 1'b1;
           break;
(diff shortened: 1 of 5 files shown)


hw/dv/sv/i2c_agent/i2c_if.sv
```diff
@@ -63,17 +63,21 @@ interface i2c_if(
...
-      if (scl_i) begin
-      wait_for_dly(tc.tHoldStart);
-      end else continue;
+      if (!scl_i) continue;
       @(negedge scl_i);
       if (!sda_i) begin
-      wait_for_dly(tc.tClockStart);
-      break;
+        // wait_for_dly(tc.tClockStart);
+        break;
```
a-will, line 80: This part of the task actually isn't correct anyway. The start symbol is _only_ a falling edge on SDA while SCL is high. The monitor probably ought to follow this up with a different task that holds until SCL falls low and checks that SDA is still low.

Then, on the side, maybe some other task should ensure timing requirements are met, instead of trying to squash it into the functional stuff in this awkward way.

a-will (approved):

hw/dv/sv/i2c_agent/i2c_if.sv
hcallahan-lowrisc: I have a whole bunch of local changes to separate the timing-checks from the functional part of the monitor. I hope to upstream it all after the pending deadlines, and things will become much simpler :)

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -158,7 +158,7 @@ module i2c_fsm import i2c_pkg::*;
         tSetupData  : tcount_d = 20'(t_r_i) + 20'(tsu_dat_i);
         tClockStart : tcount_d = 20'(thd_dat_i);
         tClockLow   : tcount_d = 20'(tlow_i) - 20'(thd_dat_i);
-        tClockPulse : tcount_d = 20'(t_r_i) + 20'(thigh_i) + 20'(t_f_i);
+        tClockPulse : tcount_d = 20'(t_r_i) + 20'(thigh_i);
```
marnovandermaas, line 161: RTL changes look good to me. `tClockPulse` is always followed by `tHoldBit` which already counts the fall time.

marnovandermaas (approved):
Thanks for making these changes! I agree that a full DV rework can be done at a later point.

andreaskurth:
Rebasing this PR on `master` should fix the failing _CW310 ROM Tests_ check (`//sw/device/tests:aon_timer_wdog_lc_escalate_test_fpga_cw310_rom_with_fake_keys` failed). The failure in _Test an airgapped Bazel build_ looks spurious.

hcallahan-lowrisc:
Rebased, plus added links to #21887

andreaskurth:
Passes CI except for the `uart_tx_rx_test_fpga_cw310_sival` test, which is unrelated to this PR. Merging.

Issue #15227, opened 2022-09-30
[i2c] Configurable support between stretch and nak.


tjaychen:
When in target mode, the i2c AddrRead state also writes to the acq_fifo to indicate a start has been seen.
However, this write is not qualified by acq_wready_i, so in the event the `acq_fifo` is still full from the previous transaction, would just cause an overflow. 

The overflow does not stop the fsm from continuing or cause the device to issue a NAK, so it will just get silently dropped in the background. If software is expecting to read the "start" indication in the acquisition fifo, this may become a problem. 

@drewmacrae 
@a-will

tjaychen:
It turns out nack-ing while fifo full was not considered as part of the original features.
This will be moved to M3 since the focus of M2 is that basic functionality of i2c is present.

There is an effort to slightly restructure i2c's fsm so that nack-ing can be more easily added in the near future.

tjaychen:
@jesultra @a-will 
can you guys have a look at this one and let me know if this is an option we have to get in?

@jdonjdon fyi

a-will:
Does the original presentation of the issue still apply? Because I do think the FSM should not continue if the ACQ FIFO cannot take in the command (does the start/re-start condition get co-located with the command, or is it an entirely separate entry?).

tjaychen:
the command gets co-located into the same entry as the start (but not the restart).
So the way this currently works is that we always reserve at least 1 entry of space for a restart / stop.
Then when it cycles back around, if there is no space for the start / command, we then stretch the address. 

But I think the original spirit of this is that we could have configurable nak or stretches, for those situations where stretch is not supported.

a-will:
I'd recommend changing the title if we're going to make the conversation about handling systems without support for stretching.

And for that, I suspect it's not a requirement. For our most prominent use case, the TPM i2c spec requires support for stretching. Are there other critical use cases for an OT i2c target?

However, our i2c IP would be more complete if we had a "stretch timeout to NAK" feature (where no stretching would imply an immediate timeout). Note that there is no way to have anything but undefined behavior for stretching that has timed out in the middle of a target-transmitter sending data--If the FIFO is empty, garbage data goes out until the host stops clocking it out.

tjaychen:
yes let me clarify the original intent of this issue.
It was mostly meant to be a binary decision between "use nak" or "stretch".

[...]

PR #21766, opened 2024-02-29
I2C NACK when full

marnovandermaas:
This PR adds NACK functionality to the I2C target. It adds a mode bit in the control register to control whether target stretches the clock when ACQ FIFO is full or sends a non-acknowledge instead.

Closes: https://github.com/lowRISC/opentitan/issues/15227

### DIFF
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -213,6 +213,14 @@
           tags: [// Exclude from write-checks: writing 1'b1 to this bit causes interrupts unexpectedly asserted
                 "excl:CsrAllTests:CsrExclWrite"]
         }
+        { bits: "3"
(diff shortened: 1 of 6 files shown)


marnovandermaas:
The last force push was a rebase on top of the latest I2C changes.

andreaskurth (commented):
Thanks for this PR, @marnovandermaas. This implements NACK-instead-of-stretch when the FIFO is full, and I think it does so correctly based on comparison with the ACK implementation. However, as the discussion whether we want NACK-instead-of-stretch or stretch-to-NACK is still open (#15227), let's hold this for now.

marnovandermaas:
I'm closing this PR since it seems like another more feature-full one: https://github.com/lowRISC/opentitan/pull/21857

Issue #19435, opened 2023-08-14
[usbdev] Support required for retracting IN packet robustly


alees24:
### Description

This is a very specific case, although I would contend that support for Isochronous IN traffic could benefit too.

It would be useful if there were a clean and robust way to retract an IN packet from the SW/FW side once it has been presented to usbdev for collection by the host. This relates to section `8.5.3.3 Error Handling on the Last Data Transaction` in the USB 2.0 protocol specification:

- During the Data Stage of a Control Transfer Read, the final IN transaction should be ACKnowledged by the host in the usual manner.
- If the ACK gets lost/corrupted, the normal Data Toggle Synchronization mechanism cannot be employed to ensure that the device/host agree.
- The device is supposed to take the commencement of the Status Stage OUT transaction as evidence that the host has successfully received the final Data Stage transaction, and shall advance automatically.
- usbdev will have retained the IN packet, which the host _has_ already collected. usbdev has no awareness of the packet having been successfully collected.
- SW/FW will need to remove the IN packet manually, because otherwise a subsequent Control Transfer will collect it instead.

With the current implementation, under this very specific case, SW/FW is able to retract the IN packet safely at the point of receiving the OUT transaction of the Status stage since the host shall make no further attempt to collect that IN packet.

SW/FW _must_ retract the IN packet in a 'timely' fashion, however, because the receipt of the Status stage OUT transaction is automatically mediated by hardware, and nothing prevents the commencement of a subsequent Control Transfer. Retraction of an IN packet may be achieved by clearing the `rdy` bit of the `configin` register for the endpoint; in general this is most safely achieved by leaving the other register fields unaltered, in the event that the device is simultaneously transmitting the packet. In this specific case, if the SW/FW is timely in responding, transmission will not occur at the same time as retraction.

Re Isochronous IN traffic, since the priority is always to present the most recent data in the most timely manner, rather than delivering data reliably, there is an argument that usbdev should support replacement of IN packet data for Isochronous endpoints. Note that removal of an IN packet is not generally safe and may result in the transmission of truncated packets, even when the `rdy` field of the `configin` register is the only field modified; for Isochronous traffic this is probably inconsequential because there is no error detection/retrying in normal use, and the recovery of host-side software by other mechanisms is required. It should usually be avoided for non-Isochronous endpoints, however.

a-will:
For control endpoints, would something like clearing the `rdy` upon receipt of a SETUP packet do the trick?

[...]

Issue #18942, opened 2023-06-16
[usbdev] Iso IN support


alees24:
### Description

Investigate whether usbdev properly supports Isochronous IN transmission. The periodic provision of packets via configX_in registers may suffer from an inability to replace a buffer that has not yet been collected by the host controller. Ideally, with Isochronous streams prioritizing real-time delivery over reliability, and with the device having no direct control over when the packets are collected, it should be possible to replace the previous buffer in a hazard-free fashion.

This may be possible with the current interface by careful coding and clearing of the RDY bit before presenting the new buffer. Alternatively, it may be sufficient to rely upon the pkt_sent interrupt being processed in time and/or using the timing of the frame interrupt to replace buffers safely.

@johngt 

> estimate 4

a-will:
Hm... if we wanted more hardware support, what might that look like? Perhaps a TTL (time to live) in config_inX that decrements on every SOF received, then clears ready (and possibly asserts an interrupt?) when TTL reaches 0?

alees24:
Subsumed into #19435

PR #21771, opened 2024-02-29
[usbdev] Support the safe retraction of IN packets

alees24:
This ~~draft~~ PR relates to a discussion of perfective changes in #19435 (open) and #18942 (closed and subsumed).

There are two distinct cases when we may want to retract IN packets that have been presented for collection by the USB host, and which are still marked as ready for transmission ('rdy' bit set in 'configin'):

- Isochronous IN traffic, when there is new data available (for Isochronous streams the more recent data is more important than older data)
- Loss of ACKnowledgement of the final IN packet within the Data Stage of a Control Transfer; in this case, the missing acknowledgement is to be detected by the lack of a 'pkt_sent' interrupt for the IN Endpoint and the arrival of the zero-length OUT DATA packet that is the Status Stage. (See '8.5.3.3 Error Handling on the Last Data Transaction')
  - this requires us to retract the final IN packet since it has indeed been sent (just not successfully), and invert the IN Data Toggle on that endpoint, to realign with the host's state.

There may be other cases, especially in handling error cases/edge conditions, that this form of IN packet retraction is required.

This PR proposes minimal changes to the IN-side hardware that involve an additional status bit in the 'configin' register. The 'sending' bit becomes set when the hardware first attempts to transmit the IN packet, and it remains set until such time as the packet is successfully transmitted (when both 'rdy' and 'sending' are cleared) or the packet is canceled by the hardware ('rdy' and 'sending' both cleared, and instead 'pend' is set).

Packet retraction operates as follows:

1. FW writes to the 'configin' register with the 'rdy' bit clear. The intention would be to write the 'buf' number and 'size' with their current values, but this PR incorporates logic to capture these properties of the current IN transaction so that they cannot be changed throughout the current transaction. It is therefore safe just to write '0' to the 'rdy' bit.
2. FW reads back from the 'configin' register to check the 'sending' and 'pend' bits.
3. If 'sending' is set then FW must wait until either successful transmission is communicated via the 'pkt_sent' interrupt, or the 'pend' bit becomes set because of transaction cancelation, as per normal IN packet transmission.
4. If 'pend' is already set, or 'sending' is clear, then the buffer may safely be reallocated and the new packet supplied immediately.

~~This remains a draft PR presently to encourage discussion and~~ because, although I have reasoned through the logic, I have not yet had a chance to test the implementation properly. Basic simulations and inspection of the logic behavior in waveforms are okay, but there are obviously no new tests to exercise this functionality at this point.

@a-will

### DIFF
--- a/hw/ip/usbdev/rtl/usbdev.sv
+++ b/hw/ip/usbdev/rtl/usbdev.sv
@@ -349,7 +349,7 @@ module usbdev
   logic [3:0]            in_endpoint;
   logic                  in_endpoint_val;
   logic [NEndpoints-1:0] in_rdy;
-  logic [NEndpoints-1:0] clear_rdybit, set_sentbit, update_pend;
+  logic [NEndpoints-1:0] clear_rdybit, set_sentbit, update_pend, set_sending;
   logic                  setup_received, in_ep_xact_end;
   logic [NEndpoints-1:0] ep_out_iso, ep_in_iso;
   logic [NEndpoints-1:0] enable_out, enable_setup, in_ep_stall, out_ep_stall;
@@ -403,6 +403,25 @@ module usbdev
       in_rdy[i] = reg2hw.configin[i].rdy.q;
     end
   end
+
+  // Captured properties of current IN buffer, maintained throughout packet collection as
+  // protection against change during packet retraction by FW.
+  logic [NBufWidth-1:0] in_buf_q, in_buf_d;
+  logic [SizeWidth:0] in_size_q, in_size_d;
+  logic       in_xact_starting;
+  logic [3:0] in_xact_start_ep;
+
+  always_ff @(posedge clk_i or negedge rst_n) begin
+    if (!rst_n) begin
+      in_buf_q  <= '0;
+      in_size_q <= '0;
+    end else begin
+      in_buf_q  <= in_buf_d;
+      in_size_q <= in_size_d;
+    end
+  end
+  assign in_buf_d  = in_xact_starting ? in_buf[in_xact_start_ep]  : in_buf_q;
+  assign in_size_d = in_xact_starting ? in_size[in_xact_start_ep] : in_size_q;
 
   // OUT data toggles are maintained with the packet engine but may be set or
   // cleared by software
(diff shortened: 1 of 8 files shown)


a-will:
> Isochronous IN traffic, when there is new data available (for Isochronous streams the more recent data is more important than older data)

For what it's worth, I'm not sure we can serve the isochronous use case _perfectly_ without double-buffering and/or a TTL. It's really a mismatch for the functions a RoT would ordinarily serve anyway, but isochronous endpoints are only supposed to be used for data that is aligned to a particular frame number (or window of frame numbers), with a specific advertised polling rate.

The reason is all having to do with the tight timing. Isochronous transfers are performed very early in the frame, and if someone wants to implement polling for every frame, then all they have is the SOF interrupt and almost no time to swap buffers for the retraction case. The TTL would guarantee packet retraction at the particular frame, and the double-buffering would immediately replace it for the next transmission.

Or... something requiring fewer bits than a TTL would be a "retract at next SOF" bit.

Probably this is only necessary for the fastest polling rate, though. Once we get to every-other-frame or slower, there is plenty of time to retract and swap (that is, after software learns which frame numbers will have transmission scheduled -- That means it also knows which frames will _not_ be scheduled, when it is safe to swap buffers). ...We could just specify a max polling rate and be done with it, heh.

Sorry for opening with a random aside!

alees24:
I think the idea of a TTL has merit but I''m aiming here for a minimal change, given the time frame, and of course this doesn't preclude the possibility of improving Isochronous support later. It's 'push driven' in that the software can at any point switch over to the new buffer and our control software already implements double-buffering, ie. the software completes the new buffer within the packet memory before then just hitting the 'configin' register to make it current.

There's no reason, in principle, that the software cannot use the frame interrupts (1ms intervals) to set up a timer and that tracks the frame interrupts but itself fires at say 80% of the way through the bus frame. It can thus switch to the new buffer in preparation for the new frame start to implement 'pull-driven' streaming.

As you say, Isochronous support really isn't a priority, though.

a-will (approved):
LGTM

hw/ip/usbdev/rtl/usbdev.sv
```diff
@@ -404,6 +404,25 @@ module usbdev
...
 
+  // Captured properties of current IN buffer, maintained throughout packet collection as
+  // protection against change during packet retraction by FW.
+  logic [NBufWidth-1:0] in_buf_q, in_buf_d;
+  logic [SizeWidth-1:0] in_size_q, in_size_d;
+  logic       in_xact_starting;
+  logic [3:0] in_xact_start_ep;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_n) begin
```
a-will, line 415: Ah, little typo (don't you just _love_ when the `default_nettype is wire?):

```suggestion
    if (!rst_ni) begin
```

jesultra:
I think this feature could be useful also for the UART to USB forwarding used by ChromeOS. Assuming the OT chip has received a few characters on the UART and put them in a buffer ready for an IN transaction, but the USB host has not yet retrieved them, when more characters arrive via UART. In such case, it would be desirable to "augment" the prepared IN buffer (rather than enqueue another buffer with just a few bytes), which could be done if we have this feature.

hw/ip/usbdev/rtl/usbdev.sv
alees24: Thanks for finding this. In this case it's actually the sensitivity list that is wrong, because usbdev.sv has this oddity of 'rst_ni' being used only for the logic that's present even in a Stubbed implementation. So, 'rst_n' actually does exist; it's not the unbelievably poor 'implicit nets' language feature, in this case.

alees24:
DV smoke test failure (chip_sw_spi_host_tx_rx) unrelated.

PR #21772, opened 2024-02-29
[silicon] Add proda verification key.

moidx:
This commit adds an RSA verification key to be used with the proda SKU.


moidx:
Will move this change to `earlgrey_es_sival`.

PR #21774, opened 2024-02-29
[dv,rstmgr] Handle CSRs reset by POR

### LINKED ISSUE #18258: [dv/csr_utils] Fix handling of reset for rstmgr alert_info_ctrl
### Description

When attempting to enable read checks for alert_info_ctrl in the rstmgr scoreboard we get a mismatch soon after a reset.
On reset only bit zero of alert_info_ctrl should be cleared, but apparently the full CSR was.

Notice the rstmgr cpu_info_ctrl is similar, in that only the lowest bit is cleared on reset, and we have no problem doing read checks for it.

To diagnose this, enable read checks for this CSR in rstmgr_scoreboard and run a few iterations of the smoke test.

matutem:
There is a handful of registers that are reset on POR only, but the dv_base classes will clear all mirrored values on reset. Rather than changing all that code to handle resets more accurately here we just capture the mirrored values of all such CSRs before reset, and apply them back once reset is handled.

Fixes #18258

### DIFF
--- a/hw/ip_templates/rstmgr/dv/env/rstmgr_scoreboard.sv
+++ b/hw/ip_templates/rstmgr/dv/env/rstmgr_scoreboard.sv
@@ -12,6 +12,10 @@ class rstmgr_scoreboard extends cip_base_scoreboard #(
   // local variables
   static const string sw_rst_ctrl_n_preffix = "sw_rst_ctrl_n_";
 
+  // This is used to capture and restore upon a regular reset the CSR values that are reset
+  // by POR only, since the CIP and lower layers will reset all CSRs.
+  rstmgr_values_of_por_csr_fields_t por_fields;
+
   // TLM agent fifos
 
   // local queues to hold incoming packets pending comparison
@@ -25,6 +29,8 @@ class rstmgr_scoreboard extends cip_base_scoreboard #(
 
   task run_phase(uvm_phase phase);
     super.run_phase(phase);
+    // Initial capture to be able to restore after POR.
+    capture_por_csr_fields();
     fork
       monitor_por();
       monitor_capture();
@@ -75,6 +81,8 @@ class rstmgr_scoreboard extends cip_base_scoreboard #(
           `uvm_info(`gfn, "tl got out of reset", UVM_MEDIUM)
           under_reset = 0;
           clear_outstanding_access();
+          cfg.clk_rst_vif.wait_clks(1);
+          restore_por_csr_fields();
         end
       end
   endtask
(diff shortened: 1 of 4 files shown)


hw/ip_templates/rstmgr/dv/env/rstmgr_scoreboard.sv
```diff
@@ -243,7 +247,59 @@ class rstmgr_scoreboard extends cip_base_scoreboard #(
...
+    `uvm_info(`gfn, $sformatf(
+              "Restoring cpu_info_ctrl.index to 0x%x", por_fields.cpu_info_ctrl_index), UVM_MEDIUM)
+    csr_wr(.ptr(ral.cpu_info_ctrl.index), .value(por_fields.cpu_info_ctrl_index), .backdoor(1),
+           .predict(1));
+    `uvm_info(`gfn, $sformatf(
+              "After restoring cpu_info_ctrl mirrored value 0x%x", `gmv(ral.cpu_info_ctrl)),
+              UVM_MEDIUM)
+  endtask
+
+  // There is a handful of registers that are reset on POR only, but the dv_base classes
```
jdonjdon, line 295: There `are` ?

hw/top_earlgrey/ip_autogen/rstmgr/dv/env/rstmgr_scoreboard.sv
```diff
@@ -243,7 +247,59 @@ class rstmgr_scoreboard extends cip_base_scoreboard #(
...
+    `uvm_info(`gfn, $sformatf(
+              "Restoring cpu_info_ctrl.index to 0x%x", por_fields.cpu_info_ctrl_index), UVM_MEDIUM)
+    csr_wr(.ptr(ral.cpu_info_ctrl.index), .value(por_fields.cpu_info_ctrl_index), .backdoor(1),
+           .predict(1));
+    `uvm_info(`gfn, $sformatf(
+              "After restoring cpu_info_ctrl mirrored value 0x%x", `gmv(ral.cpu_info_ctrl)),
+              UVM_MEDIUM)
+  endtask
+
+  // There is a handful of registers that are reset on POR only, but the dv_base classes
```
jdonjdon, line 295: Same here is -> are

jdonjdon (approved):
LG with nits

hw/top_earlgrey/ip_autogen/rstmgr/dv/env/rstmgr_scoreboard.sv
matutem: Done

PR #21776, opened 2024-03-01
[manuf] send endorsed UDS cert back to the device

timothytrippel:
This updates the FT provisioning flow to send the endorsed UDS cert back to the device and writes it to flash. This partially addresses #19455.

### DIFF
--- a/sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize_3.c
+++ b/sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize_3.c
@@ -48,6 +48,7 @@ static manuf_cert_perso_data_out_t out_data = {
     .cdi_1_certificate = {0},
     .cdi_1_certificate_size = kCdi1MaxCertSizeBytes,
 };
+static manuf_endorsed_certs_t endorsed_certs;
 
 static const flash_ctrl_perms_t kCertificateFlashInfoPerms = {
     .read = kMultiBitBool4True,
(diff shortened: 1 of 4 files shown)


moidx (approved):

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize_3.c
```diff
@@ -138,7 +139,19 @@ static status_t personalize(ujson_t *uj) {
...
 
-  // TODO(#19455): Load endorsed UDS certificate and write it to flash.
+  // Import endorsed certificates from the provisioning appliance.
+  LOG_INFO("Importing certificates ...");
+  TRY(ujson_deserialize_manuf_endorsed_certs_t(uj, &endorsed_certs));
+
+  // Write the endorsed UDS certificate to flash and ack to host.
+  TRY(flash_ctrl_info_erase(&kFlashCtrlInfoPageUdsCertificate,
+                            kFlashCtrlEraseTypePage));
+  TRY(flash_ctrl_info_write(
```
pamaury, line 149: Should the certificate size be written somewhere? I don't think this writes it to the flash

pamaury (approved):

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize_3.c
timothytrippel: This is encoded in the ASN.1, no? Ultimately we will need functions to read out the ASN.1 from flash to extract various values. But for now I think we can just write the bytes to the page. When we read out the bytes, we can just read the maximum number of cert bytes from the flash info page, b/c each cert is isolated to its own page currently. We may be some zeros at the end, but that is ok.

PR #21777, opened 2024-03-01
[opentitanlib] update default JTAG speed

timothytrippel:
This updates the default JTAG speed to 1MHz to descrease test time in CI for tests that make use of JTAG.


pamaury (approved):
Looks good to me if that's stable enough to work reliably.

PR #21778, opened 2024-03-01
[opentitantool] Increase TPM timeout to match ChromeOS

jesultra:
Some TPM operations, such as generating RSA keypairs can take several minutes (at least on Google's legacy Titan chips).  This PR increases the time that opentitantool is willing to wait for the TPM status register to indicate that a command is done being processed.


engdoreis (approved):
LGTM

PR #21779, opened 2024-03-01
[alert_handler_testutils] Revise default ping timeout

a-will:
Add a default ping timeout for tests using alert_handler testutils. A number of tests previously used 0 here, but that value risks timeouts, since alert_handler_testutils_configure_all() enables the ping feature unconditionally. Bump that up to the recommended value for earlgrey.

### DIFF
--- a/sw/device/lib/testing/alert_handler_testutils.h
+++ b/sw/device/lib/testing/alert_handler_testutils.h
@@ -14,6 +14,10 @@
 
 #include "alert_handler_regs.h"
 
+enum {
+  kAlertHandlerTestutilsDefaultPingTimeout = 256,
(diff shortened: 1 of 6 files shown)


a-will:
CC @vogelpi

vogelpi (approved):
Many thanks @a-will for debugging and fixing this!

vogelpi:
I am merging this now to unblock CI. Since yesterday morning, the `//sw/device/tests:aon_timer_wdog_lc_escalate_test_fpga_cw310_rom_with_fake_keys` of the CW310 ROM Test CI job fails for basically every PR and this one here seems to solve it.

PR #21780, opened 2024-03-01
[RV_DM]rv_dm_progbuf_exception_vseq

BasitMehmood:
This commit contains vseq which verify that writing on tl_mem_ral.exception register causes the jtag_dmi_ral.abstractcs.cmderr set to 3.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_progbuf_exception_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_progbuf_exception_vseq.sv
@@ -1,0 +1,30 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class rv_dm_progbuf_exception_vseq extends rv_dm_base_vseq;
+  `uvm_object_utils(rv_dm_progbuf_exception_vseq)
+
+  `uvm_object_new
+
+  constraint lc_hw_debug_en_c {
+    lc_hw_debug_en == lc_ctrl_pkg::On;
+  }
+  constraint scanmode_c {
+    scanmode == prim_mubi_pkg::MuBi4False;
+  }
+  task body();
+    uvm_reg_data_t data;
+    uvm_reg_data_t r_data;
+    repeat (2) begin
+      data = $urandom;
+      request_halt();
+      csr_wr(.ptr(jtag_dmi_ral.command), .value(data));
+      csr_wr(.ptr(tl_mem_ral.exception), .value(data));
+      csr_rd(.ptr(jtag_dmi_ral.abstractcs), .value(r_data));
+      `DV_CHECK_EQ(AbstractCmdErrException, get_field_val(jtag_dmi_ral.abstractcs.cmderr,
+                                                          r_data));
+    end
+  endtask
+
+endclass:rv_dm_progbuf_exception_vseq
--- a/hw/ip/rv_dm/dv/env/rv_dm_env.core
+++ b/hw/ip/rv_dm/dv/env/rv_dm_env.core
@@ -50,6 +50,7 @@ filesets:
       - seq_lib/rv_dm_progbuf_busy_vseq.sv: {is_include_file: true}
       - seq_lib/rv_dm_abstractcmd_status_vseq.sv: {is_include_file: true}
       - seq_lib/rv_dm_rom_read_access_vseq.sv: {is_include_file: true}
+      - seq_lib/rv_dm_progbuf_exception_vseq.sv: {is_include_file: true}
     file_type: systemVerilogSource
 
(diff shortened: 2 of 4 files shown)


BasitMehmood:
@rswarbrick kindly review it.thanks

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_progbuf_exception_vseq.sv
```diff
@@ -0,0 +1,31 @@
...
+  constraint lc_hw_debug_en_c {
+    lc_hw_debug_en == lc_ctrl_pkg::On;
+  }
+  constraint scanmode_c {
+    scanmode == prim_mubi_pkg::MuBi4False;
+  }
+  task body();
+    uvm_reg_data_t data;
+    uvm_reg_data_t r_data;
+    repeat ($urandom_range(1, 10)) begin
```
rswarbrick, line 20: Any particular reason to randomise the number of times we do this? If not, I'd just pick some small number greater than 1 (like 2, for example).

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_progbuf_exception_vseq.sv
```diff
@@ -0,0 +1,31 @@
...
+      data = $urandom_range(0,31);
+      request_halt();
+      // Access Register command with postexec bit set.
+      csr_wr(.ptr(jtag_dmi_ral.command), .value('h00250000));
```
rswarbrick, line 24: Please can you construct this value using the datatypes in `jtag_rv_debugger_pkg.sv`. That should avoid us needing magic hex numbers.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_progbuf_exception_vseq.sv
```diff
@@ -0,0 +1,31 @@
...
+      csr_wr(.ptr(tl_mem_ral.exception), .value(data));
+      csr_rd(.ptr(jtag_dmi_ral.abstractcs), .value(r_data));
+      `DV_CHECK_EQ(3, get_field_val(jtag_dmi_ral.abstractcs.cmderr, r_data));
```
rswarbrick, line 27: This magic value is named in `abstrct_cmd_err_e` as `AbstractCmdErrException`. Let's use the name.

BasitMehmood:
> I think this might need a bit of reworking. Looking at the testplan, I think we are supposed to be "writing on tl_mem_ral.exception register while an abstract command is executing".
> 
> Unfortunately, it's a bit fiddly to get the timing right and be certain that an abstract command is still executing while we write to the exception register. So I think this probably needs to be something like the following:
> 
> * Write a command over the JTAG DMI interface
> * Write something over TileLink to `exception`
> * Be certain the command is still running (how?)
> * Check over DMI that `cmderr` now contains `AbstractCmdErrException`
> 
> Structuring it like this will avoid an unchecked race between DMI and TL access.
> 
> This still means we need to work out how to check if the command was still running. To avoid needing a backdoor, the easiest approach is probably to choose an abstract command that will take unbounded time (because it requires the nonexistent core to respond in some way).

Actually if there is hart present,  then if we try to access the register which is not present in hart then cmderr will set to 3 to show exception. In our case we don't have hart so we are just checking that if we write on tl_mem_ral.exception register on behalf of hart, we got cmderr=3.

BasitMehmood:
I don't know why (Run DV smoke test) is failing?

rswarbrick:
Thanks for the responses. I'm afraid that I've got a bit confused about what you're actually trying to argue here. I suggest that you rebase onto the head of master and re-push (which will fix the failing smoke test).

But tweak the code to explain in comments what we're doing and why we are convinced that the second `csr_wr` happened while there was an abstract command in flight.

BasitMehmood:
> Thanks for the responses. I'm afraid that I've got a bit confused about what you're actually trying to argue here. I suggest that you rebase onto the head of master and re-push (which will fix the failing smoke test).
> 
> But tweak the code to explain in comments what we're doing and why we are convinced that the second `csr_wr` happened while there was an abstract command in flight.

1- I think abstract command is completed before we write to tl_mem_ral.exception register. Actually what we are doing is just writing on exception and read status in abstractcs.cmderr=3. So i think abstract command is not necessary to implement to check status of exception.
2- I just realized that this test-case is already implemented in "rv_dm_cmderr_exception" test case. We can skip this test because this functionality is already covered.

rswarbrick:
Ok, I think I understand. So are you saying that this is no longer needed? If so, would you mind opening a PR with a testplan change to remove the testpoint?

BasitMehmood:
> Ok, I think I understand. So are you saying that this is no longer needed? If so, would you mind opening a PR with a testplan change to remove the testpoint?

Ok i will open a PR to remove testpoint.

PR #21784, opened 2024-03-01
[pentest] Fix bugs in AES SCA

nasahlpa:
This PR fixes three bugs in the AES used for SCA in separate commits:
- Set the seed to the correct state
- Set the unused key shares
- Configure entropy complex for SCA measurements


vrozic (approved):
LGTM
Thanks for debugging this.

vogelpi (approved):
LGTM, thanks for fixing these bugs!

PR #21786, opened 2024-03-01
[sw/i2c_testutils] Take I2C instance ID into account for pins

andreaskurth:
The previous `kI2cPlatformPins` array mapped to pins solely based on the platform.  This works for CW310 (with PMOD and HyperDebug) where I2C is mapped to one pair of pins for all I2C instances.  In DV, however, each I2C instance has its own agent, which is connected to its own pair of (muxed) pins.  Coincidentally, when the I2C instance ID (i.e., 0 to 2) is (incorrectly) used to index the `kI2cPlatformPins` array, that would result in the correct pair of pins for DV.  When pins change for one of the platforms or the number of I2C instances changes, however, this will no longer work.

This commit takes the I2C instance ID into account when mapping to pins. The result is the same as before on the condition that the platform argument to a call of `i2c_testutils_select_pinmux` in `i2c_device_tx_rx_test.c` gets fixed (which this commit also does).

This got split out of the closed PR #21676.

### DIFF
--- a/sw/device/lib/testing/i2c_testutils.c
+++ b/sw/device/lib/testing/i2c_testutils.c
@@ -91,46 +91,72 @@ static const i2c_pinmux_pins_t kI2cPinmuxPins[] = {
 };
 
 /**
- * This table store i2c pin mappings of different platforms.
- * This is used to connect i2c instances to mio pins based on the platform.
+ * Map the combination of platform and I2C ID to pins for SCL and SDA.
  */
-static const i2c_platform_pins_t kI2cPlatformPins[] = {
-    // Hyper310 bitstream.
-    [I2cPinmuxPlatformIdHyper310] =
-        {.sda =
-             {
-                 .mio_out = kTopEarlgreyPinmuxMioOutIoa7,
-                 .insel = kTopEarlgreyPinmuxInselIoa7,
-             },
-         .scl =
-             {
-                 .mio_out = kTopEarlgreyPinmuxMioOutIoa8,
-                 .insel = kTopEarlgreyPinmuxInselIoa8,
-             }},
-    // DV.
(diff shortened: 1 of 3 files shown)


sw/device/lib/testing/i2c_testutils.c
```diff
@@ -91,46 +91,72 @@ static const i2c_pinmux_pins_t kI2cPinmuxPins[] = {
...
+                                    .mio_out = kTopEarlgreyPinmuxMioOutIoa8,
+                                    .insel = kTopEarlgreyPinmuxInselIoa8,
+                                }};
+      break;
+    case I2cPinmuxPlatformIdDvsim:  // DV
+      // In DV, there's one agent for each I2C instance, with a fixed set of
+      // muxed pins.
+      switch (i2c_id) {
+        case 0:  // I2C0 uses the same pins as CW310 HyperDebug
+          TRY(map_platform_to_pins(I2cPinmuxPlatformIdHyper310, i2c_id, pins));
```
a-will, line 118: I feel a little iffy about tying the platforms together like this, since DV could evolve into something different while hyper310 could not change. However, no need to change it on my account if you don't feel similarly.

a-will (approved):

sw/device/lib/testing/i2c_testutils.c
andreaskurth: I see your point -- but this intertwining existed in the code already before this commit implicitly, this commit just makes it explicit. If the pins change in DV, this will no longer hold and tests will fail. I think the 'perfect' solution would be to describe the DV environment in an Hjson and generate RTL and headers from it, but in the meantime I think this is as good a solution as we can get (without duplicating code).

Issue #21686, opened 2024-02-26
[entropy_src] Determine when and what FIFOs can drop entropy


h-filali:
### Description

Currently several FIFOs drop entropy on certain occasions e.g. the postht FIFO, the precon FIFO and the bypass FIFO.
We need to evaluate for each FIFO when and in what configuration it is allowed to drop data.
For the instances where FIFOs are not allowed to drop data we need to add alerts to notify the firmware that entropy has been dropped. We should also determine whether those alerts should be fatal or not. There are already fatal alerts for the esrng FIFO, the observe FIFO and the esfinal FIFO in place. However, the observe FIFO is the only one of those three that has a write error. The other two only have a read and a state error.

To determine whether there are writes to full FIFOs I wrote assertions which trigger when the FIFOs are full and they are written to.
I was able to find writes when full for the postht FIFO, the bypass FIFO and the precon FIFO. I did this both before and after [this](https://github.com/lowRISC/opentitan/pull/21685/commits/d23d43ea543f6b83d27e7487eb0c4a5edf5bd82a) commit. For both before and after the commit I get the same result.

vogelpi:
Thanks for the draft PR @h-filali . I had a look at this again, also because Andreas spent quite some time in the past writing all the SVAs for tracking when entropy is lost in the hardware pipeline. Based on what I've found now, I think there are some misunderstandings. In my view, the key points are:
- Both the `prim_fifo_sync` and the `prim_packer_fifo` primitives ignore pushes if they're full. So it's not possible to push more data if those FIFOs are already full. Unfortunately, in the ENTROPY_SRC RTL it is only described (and a bit ambiguously in my view) for two FIFOs that this only needs to be considered when popping the previous FIFO. But e.g. the observe FIFO still has the error connected to the error register even though this doesn't make sense really. To understand better whether the observed behavior is problematic, I would need to see the SVAs you used before creating this issue.

[...]

Issue #20953, opened 2024-01-24
[entropy_src] Make sure unaltered raw data can be extracted for FIPS randomness validation


johannheyszl:
### Description

Our current understanding is that no health tests will alter the data that is accumulated in the observe FIFO. FIPS validation testing, we can use the Firmware Override / Bypass Mode to read it (see [Programmer's Guide - OpenTitan Documentation](https://opentitan.org/book/hw/ip/entropy_src/doc/programmers_guide.html#firmware-override--bypass-modes)). An interrupt is generated on reaching a predefined threshold in the FIFO.

Issue created after discussion b/w: @moidx @johannheyszl VadimS @vogelpi @zi-v @h-filali

johannheyszl:
cc @msfschaffner @andreaskurth

vogelpi:
@johannheyszl am I correct in assuming that this is mainly a block-level DV tasks? I.e., is it sufficient to check that the entropy arriving in the observe FIFO is indeed what is received at the input?

vogelpi:
I wanted to give a quick update on this @johannheyszl . While @h-filali has been working towards solving this issue, he noticed that some FIFOs aren't connected to alerts that fire if the FIFO is pushed when full already (while other FIFOs have such alerts). Adding some SVAs specifically for this, seemed to indicate that the postht FIFO, the precon FIFO and the bypass FIFO can be pushed when they're actually full, so entropy can get lost. The good news is that this doesn't affect the firmware override modes important for validation.

As for pushing full FIFOs, I am currently thinking that there is still a misunderstanding here and those FIFOs aren't really pushed. Because there are quite some SVAs by @andreaskurth  in the design ensuring that no entropy is systematically dropped where it shouldn't be dropped. Now that we no longer disable the noise source if the pipeline gets full (wasn't compliant with the spec) it seems it can happen that entropy is dropped at the very input. This is what needs to be analyzed for M2  as it we might need to add another FIFO to buffer entropy when the conditioner is busy. (see #21686)

I think the DV task this PR is about can be done in M3 instead.

zi-v:
Regarding "it seems it can happen that entropy is dropped at the very input ... need to add another FIFO ...": I am not sure about the context, but if this is to support continuous health test requirement, than all you need to do is move the overflow samples drop point to after the health test. Note that the entropy rate from analog RNG is slower by orders of magnitude compared to the core clock so it shouldn't be a challenge to keep health test serving the incoming analog samples.

[...]

PR #21787, opened 2024-03-01
[entropy_src, sha3] Fix CS AES Halt interface and entropy_src_rng_max_rate test

vogelpi:
This PR contains a couple of commits to fix the configuration of the `entropy_src_rng_max_rate` test (commits 1 and 2) and the known-to-be-broken CS AES halt interface intended to prevent power spikes at the chip level by only letting either the AES core inside CSRNG or the SHA3 core inside the entropy source run, but not both together (commits 3 and 4).

The `entropy_src_rng_max_rate` test is important to analyze the back pressure behavior of the pipeline and the CS AES halt interface (if properly working) is the main source of back pressure inside the hardware pipeline of ENTROPY_SRC.

All this is related to #21686 and #20953.


vogelpi:
I ran a regression with `--reseed-multiplier 0.1` for this PR. All tests pass except for the `entropy_src_rng_max_rate` test which now has a pass rate of only 12.5 %. However, this is expected as the DUT now experiences heavy backpressure during that test which the DUT cannot properly handle at the moment. This has to be fixed with a follow-up PR.

hw/ip/entropy_src/dv/env/entropy_src_env.sv
```diff
@@ -41,6 +41,9 @@ class entropy_src_env extends cip_base_env #(
     cfg.m_rng_agent_cfg.zero_delays = 0;
     cfg.m_rng_agent_cfg.host_delay_min = 1;
     cfg.m_rng_agent_cfg.host_delay_max = 12;
+    if (cfg.rng_max_delay_specified) begin
+      cfg.m_rng_agent_cfg.host_delay_max = cfg.rng_max_delay;
+    end
```
andreaskurth, line 46: Couldn't we unconditionally make this assignment? When the plusarg is not provided, the value would come from `entropy_src_env_cfg`, right?

vogelpi: Yes we can do that it will simplify the code a bit.

hw/ip/kmac/rtl/sha3.sv
```diff
@@ -155,9 +161,34 @@ module sha3
...
+  logic keccak_run_req_d, keccak_run_req_q;
+  logic keccak_triggered_d, keccak_triggered_q;
   logic keccak_complete;
 
-  assign keccak_run = sha3pad_keccak_run | sw_keccak_run;
+  // Announce that we want to run the Keccak core and tell other blocks to go
+  // quiet. Keep holding the REQ until the Keccak core is done with the
+  // processing. The keccak_complete signal is received once the Keccak core
+  // is back in the Idle state and again susceptible to keccak_run.
+  assign run_req_o = sha3pad_keccak_run | sw_keccak_run | keccak_run_req_d;
```
andreaskurth, line 172: Can't we simply assign from `keccak_run_req_d` here?

andreaskurth (approved):
Thanks for this work, @vogelpi! It's really nice how much simpler the *halt request* mechanism is to implement inside SHA3 than in entropy_src's FSM. I only have a small suggestion

hw/ip/kmac/rtl/sha3.sv
vogelpi: Absolutely, good point!

vogelpi:
Thanks for your review @andreaskurth ! Yes, it's quite cool how simple things are now. :-)

Issue #21755, opened 2024-02-29
[i2c,dv] Failing block level DV tests as a result of changes.


alees24:
### Description

This issue tracks the state of block level I2C tests, broken or made less reliable by DD work for prod.

Probably failing/unreliable already:

i2c_target_stress_all_with_rand_reset

PR [#21621](https://github.com/lowRISC/opentitan/pull/21621) at this time is thought to be responsible for breaking the following:

- i2c_host_fifo_reset_fmt
- i2c_host_fifo_watermark
- i2c_target_smoke
- i2c_target_stress_rd

Others:

- i2c_intr_test : cip_base_vseq needs 'run_intr_test' changes. (PR #21804)
- cip_base_vseq also requires a change to 'clear_all_interrupts' which is still pending merge and will cause a number of other vseqs to fail, unless it or the final commit of PR #21621 is present.

andreaskurth:
`i2c_host_fifo_watermark` is another victim, I think. It passes before the merge of #21755 but fails after (seed `9596103540527566890202921826073542621372745652001006134958711591620971245237` on VCS T-2022.06-SP2). I added it to the list.

andreaskurth:
`i2c_target_stress_rd` as well: it passes before the merge of #21621 but fails after (seed `1123967404029634992796962205584092120176243998109053670291080899048047786851` on VCS T-2022.06-SP2). This test simply extends `i2c_target_smoke`, which may explain why it fails as well.

Added to the list.

a-will:
These all seem to be passing now--I fixed up several of these along the way.

We're not at 100% at this point, though:

- i2c_csr_bit_bash overflows FIFOs and triggers assertions inside the newer FIFO RTL
- i2c_host_perf has a bizarre calculation for speed that gets affected by stretching that randomly occurs
- i2c_target_unexp_stop appears to not generate correct stimulus
- i2c_target_perf generates incorrect stimulus, with SDA and SCL switching simultaneously
- i2c_target_stretch I haven't looked at

hcallahan-lowrisc:
Moving tracked effort from this issue to #23077

moidx:
Discussed moving as P1 to M5 due to RTL risk analysis performed during last issue triage.

hcallahan-lowrisc:
This PR doesn't really accurately represent the state of the I2C regressions anymore. Hence I'm going to close it.

PR #21789, opened 2024-03-01
[i2c] Increase depth of ACQ FIFO to 268 bytes 

andreaskurth:
This change allows the I2C HW IP module to absorb a max-length SMBus Block Write,
which consists of 1 command code byte, 1 byte defining the number of data bytes in the
block write, up to 255 data bytes, and 1 optional Packet Error Code
(PEC) byte; thus up to 258 bytes in total (from SMBus Specification
version 3.2, section 6.5.7).  Our I2C HW IP module additionally puts the
byte after the Start condition (which contains the address and the R/Wn
bit) and a byte for the Stop condition into the FIFO; thus the total
increases to 260 bytes.

To account for any oversights we may have made in the analysis of the
protocol, we add an extra 8 bytes as margin, thus 268 bytes.

Note that this is larger than required by MCTP, which would be 75 bytes:
- "The size of a transmission unit is defined as the size of the packet
  payload that is carried in an MCTP packet.  The [...] minimum
  transmission unit size for MCTP is 64 bytes.  A message terminus that
  supports MCTP control messages shall always accept valid packets that
  have a transmission unit equal to or less than the baseline
  transmission unit." (from MCTP Base Specification version 1.3.0,
  section 8.3)
- MCTP packets have a 4 byte header followed by the packet payload and
  they are encapsulated between a header and a tailer that is specific
  to the physical medium (from MCTP Base Specification version 1.3.0,
  section 8.1).  For MCTP on SMBus, header and tailer consist of 4 and 1
  byte, respectively (see above).
- Thus 4 + 4 + 64 + 1 = 73 bytes plus two bytes for Start and Stop
  (based on our HW IP module; see above) gives 75 bytes.

If the ACQ FIFO cannot absorb a full SMBus Block Write or MCTP control
message, real-time constraints to process bytes from the ACQ FIFO are
imposed on SW controlling the I2C HW module.  The maximum allowed delays
depend on the protocol:
- SMBus allows target devices to stretch the clock by up to 25 ms per
  message (t_LOW:CEXT in Table 14, 15, and 16 of the SMBus Specification
  version 3.2).
- MCTP allows target devices to stretch the clock by at most 250 us per
  byte (from MCTP SMBbus/I2C Transport Binding Specification version
  1.2.0, section 6.18).

In order to not impose such real-time constraints on SW for top-level
designs that use I2C with SMBus or MCTP, this commit makes the ACQ FIFO
sufficiently deep that any max-length transaction can be absorbed.  If
SW doesn't drain the ACQ FIFO once it's full before the next message,
that message should get NACK'd and the controller has to retry later --
but protocol compliance holds and the I2C lines are not blocked.
Currently our I2C HW IP module doesn't implement this NACKing, but this
will be resolved in a separate commit (see the PR of this commit for
details and links).

Block-level tests pass, except for the known failures listed in #21755, thus not caused by this PR.

### DIFF
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -105,10 +105,18 @@
   param_list: [
     { name: "FifoDepth",
       desc: '''
-            Depth of FMT, RX, TX, and ACQ FIFOs.
+            Depth of FMT, RX, and TX FIFOs.
             The maximum supported value is 2^12-1, although much lower values are recommended to keep area requirements reasonable.
             ''',
       type: "int",
+      default: "64",
+    }
+    { name: "AcqFifoDepth",
+      desc: '''
+            Depth of ACQ FIFO.
+            The maximum supported value is 2^12-1, although much lower values are recommended to keep area requirements reasonable.
+            ''',
+      type: int
       default: "64",
     }
   ],
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -117,7 +117,7 @@
             The maximum supported value is 2^12-1, although much lower values are recommended to keep area requirements reasonable.
             ''',
       type: int
-      default: "64",
+      default: "268",
     }
   ],
   features: [
--- a/hw/ip/i2c/dv/env/i2c_env_pkg.sv
+++ b/hw/ip/i2c/dv/env/i2c_env_pkg.sv
@@ -72,7 +72,7 @@ package i2c_env_pkg;
   parameter uint I2C_FMT_FIFO_DEPTH = i2c_reg_pkg::FifoDepth;
   parameter uint I2C_RX_FIFO_DEPTH  = i2c_reg_pkg::FifoDepth;
   parameter uint I2C_TX_FIFO_DEPTH  = i2c_reg_pkg::FifoDepth;
-  parameter uint I2C_ACQ_FIFO_DEPTH = i2c_reg_pkg::FifoDepth;
+  parameter uint I2C_ACQ_FIFO_DEPTH = i2c_reg_pkg::AcqFifoDepth;
 
   // alerts
   parameter uint NUM_ALERTS = i2c_reg_pkg::NumAlerts;
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_glitch_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_glitch_vseq.sv
@@ -17,7 +17,7 @@ class i2c_glitch_vseq extends i2c_target_smoke_vseq;
   // Number of cycles sequence wait before all of the read states are executed
   parameter uint READ_STATE_WAIT_TIMEOUT_CYCLES = 40;
   // ACQ FIFO size in bytes
-  parameter uint ACQ_FIFO_SIZE = 64;
+  import i2c_env_pkg::I2C_ACQ_FIFO_DEPTH;
   // Period of SCL clock depending on timing parameters
   uint scl_period;
 
@@ -316,7 +316,7 @@ class i2c_glitch_vseq extends i2c_target_smoke_vseq;
       // Add data items to transaction enter StretchAddr state
       if (addr_states[i] == StretchAddr) begin
         // one transaction for start address and another for stop condition
-        repeat(ACQ_FIFO_SIZE - 2) begin
+        repeat(I2C_ACQ_FIFO_DEPTH - 2) begin
           `uvm_create_obj(i2c_item, req)
           append_data(req, m_i2c_target_seq.req_q);
         end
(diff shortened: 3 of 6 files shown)


a-will:
> For MCTP control messages, clock stretching is only allowed by the
requester (from MCTP SMBbus/I2C Transport Binding Specification
version 1.2.0, section 6.19).

FYI, I think this was a mistake in understanding. The particular line reads

> Response specifications are given assuming that the requester is able to operate at full speed on the bus.
> That is, clock stretching, if any, is solely generated by the requester.

In other words, it is qualifying that the timing requirements do not account for clock stretching being done by the target, not that it requires clock stretching to only be done by the requester. (The latter would be a requirement that goes too far, since the bus could have non-MCTP devices on it.)

msfschaffner (approved):
RTL changes LGTM

a-will (approved):

hcallahan-lowrisc (approved):
Having not read the spec(s) myself and going by the details in the PR message, all of these changes seem reasonable to me.

alees24 (commented):
The implementation of the changes looks fine, however I think the depth of 258 is inappropriate:
- the ACQ FIFO must also accommodate at least one START entry and one STOP entry as I understand things.
- is there a risk that RESTARTs will also need to be captured, eg. a host may choose to decompose the full message into smaller chunks internally and use RESTARTs between those chunks of the full message?
- if we're taking the hit of deepening the FIFO as far as 258 entries, I would - aside from the above points - suggest padding a little further to allow for errors in calculation/protocol undertanding at this stage anyway.

One further point, an open question since I'm afraid I haven't yet read/understood enough - can you explain the claim that any subsequent message will be NAKed, please? Do you mean in the I2C target hardware; I couldn't see evidence of that, looking quickly.

andreaskurth:
Thanks for your feedback, @hcallahan-lowrisc and @alees24.

> -  the ACQ FIFO must also accommodate at least one START entry and one STOP entry as I understand things.

Good catch, so that will be two additional entries, totaling to 260 bytes. I'll amend the PR.

> -  is there a risk that RESTARTs will also need to be captured, eg. a host may choose to decompose the full message into smaller chunks internally and use RESTARTs between those chunks of the full message?

In general I2C, I think that would be possible because a controller can use a REPEATED START (Sr) to 'chain' data transfers to the same or different devices (from I2C Specification rev. 7.0, section 3.1.10). However, as I2C doesn't limit the length of data transfers ("The number of bytes that can be transmitted per transfer is unrestricted" -- from I2C Specification rev. 7.0, section 3.1.5), I don't think we should take this into account for the depth of the ACQ FIFO.

In SMBus, a Sr is "used to switch from write mode to read mode in a combined format protocol (e.g., Byte Read)" (from SMBus Specification version 3.2, section 2.4). The SMBus protocols that make use of Sr are (exhaustive list, from SMBus Specification version 3.2, section 6.5):
(a) Read Byte and Read Word (6.5.5)
(b) Process Call (6.5.6)
(c) Block Read (6.5.7)
(d) Block Write-Block Read Process Call (6.5.8)

In cases (a), (b), and (c), the controller wants to read data from the target and always sends a fixed number of bytes (3 to 5) to the target before switching the direction to read. In case (d), the controller first writes up to M (>= 0) bytes to the target before reading up to N (>= 0) bytes; and M + N must be <= 255. So in the worst case for the ACQ FIFO, M = 255 and N = 0, which means 1 start byte, 1 command code byte, 1 byte count byte, 255 data bytes, and 1 repeated start byte (no PEC byte from controller); total 259 bytes. For the number of words to store in the ACQ FIFO, it doesn't matter if the last word is a Sr or a P byte.

> -  if we're taking the hit of deepening the FIFO as far as 258 entries, I would - aside from the above points - suggest padding a little further to allow for errors in calculation/protocol undertanding at this stage anyway.

I agree, but we probably don't want to add too many extra entries. How about 272?

a-will:
> For MCTP on SMBus, however, there is a window during which a target device is allowed to NACK (see MCTP SMBbus/I2C Transport Binding Specification version
1.2.0, section 6.14), and I think implementing that cannot be done in SW alone

The timing of the NACKing behavior is the primary reason MCTP scoping provided for a special higher-level hardware controller for Block Write protocols, hehe.

What we have now is effectively a logical layer for signaling and transfer framing with a really barebones frontend data / packet layer. The core does signals to bits/bytes and bits/bytes to signals, but it only bolts on FIFO status to drive that layer's control inputs. An abstraction with a more formalized "nack_byte_i" might be a useful input to the FSM, so a higher layer could choose to reject data independently and direct some of the transitions.

In turn, the logical layer module would expose a little of its state (when a relevant transfer starts + transaction boundaries), so the packet module(s) know when a new transfer started, whether this was a read transfer that immediately followed a write transfer in the same transaction (repeated-start style), etc.

andreaskurth:
Merging as just discussed with @msfschaffner

PR #21631, opened 2024-02-22
[DO NOT MERGE][quality] Change the way rustfmt_check and rustfmt_fix work

pamaury:
**This PR depends on https://github.com/lowRISC/rules_rust/pull/5**

Previously, rustfmt_fix was calling a custom script but actually, while the intent was to call the official `rustfmt`, it was calling `@rules_rust//tools/rustfmt` which is a different thing: it's a driver that does bazel queries to find out which files to change. The idea being this driver is that one can already just run
```
bazel run @rules_rust//tools/rustfmt
```
and it will fix all the local rust files in the repository (and ignores the ones tagged as `norustfmt`). This commit changes `rustfmt_fix` to essentially do that.

Unfortunately, `@rules_rust//tools/rustfmt` does do check-only in the upstream version. This commit relies on a custom modification of rules_rust so that the driver accepts a `--check` option. This way, rustfmt_check and rustfmt_fix consider exactly the same set of files.

PR #21790, opened 2024-03-01
[ci] add missing invocation for `rustfmt`

RyanTorok:
* Currently not ready for review * I needed to mark as ready so I can test the new CI check.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -95,6 +95,7 @@ jobs:
   - bash: ci/scripts/rust-format.sh $SYSTEM_PULLREQUEST_TARGETBRANCH
     condition: eq(variables['Build.Reason'], 'PullRequest')
     displayName: rustfmt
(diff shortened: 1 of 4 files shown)


pamaury:
We are already running this script in CI:
https://github.com/lowRISC/opentitan/blob/master/azure-pipelines.yml#L95

The problem is that the `//quality:rustfmt_check` target manually list dependencies to check which is not sustainable. I have some PRs that would enable to get rid of this:
- https://github.com/lowRISC/rules_rust/pull/5
- https://github.com/lowRISC/opentitan/pull/21631

But they require changes to `rules_rust`.

RyanTorok:
The invocation was missing in `ci/jobs/quick-lint.sh`. Is this because of the issue you mentioned?

RyanTorok:
@pamaury 's explanation makes sense. The issue is with `rustfmt_check`, not any missing script. Closing.

PR #21791, opened 2024-03-01
[usbdev] Software resets for internal buffer FIFOs

### LINKED ISSUE #17300: [usbdev] FIFO state inaccessible to software
### Description
Correct operation of the usbdev block relies upon the software knowing the buffer IDs passed to the Available Buffer FIFO (avbuffer) and not using those buffers (queueing them again, or attempting to use them for transmission of IN packets). Software can neither inspect the contents of the avbuffer (which would be tricky), nor flush the FIFO.

In normal use this is fine; my concern is that in the event of suspend/resume, or unexpected bus behavior, there may be a case in which buffer IDs have been supplied to the Available Buffer FIFO but it is difficult/impossible for the software and hardware to recover synchronization without forcing a reset of the 'usbdev' block (via the 'ast' Analog Sensor Top, which is not supposed to be reprogrammed after first ROM boot). The risk would be that the first few Control Transfers from the host to set the device address/configuration would not be properly processed, and the device would not be set up.

RFC:

Since the 'avbuffer' and 'rxfifo' implementations already have synchronous resets ('clr_i') I propose that these should be tied to a software controlled register that would be driven only upon startup/reconfiguration when the device is disconnected and the link inactive.

Even if the implementation is as simple as the register flop driving the signal 'clr_n' directly' and requires a couple of register writes, it could be a valuable contingency measure; the rest of the hardware state is (probably) recoverable already.

Aside: with our current DIF code, it is by this token not possible to call usbdev_configure() a second time after reset because doing so will render the buffer pool inconsistent with the hardware state. If there is any case when the exact PHY configuration (single ended vs differential, swapped vs unswapped pins etc) needs to change during run time, this is a problem.

> estimate 8

alees24:
This PR introduces low-impact software resets for the internal buffer FIFOs, controlled by a new register, as a contingency against unanticipated error conditions, and for future deployments of the USB device where the use of the asynchronous block-level reset is not a viable option. Block-level reset may impact other IP blocks and/or firmware within other chips that contain the USB device, and it may be very time-consuming.

Use of the software resets is only safe when the USB device is not connected to the USB, because packets may be received at any time when the USB is active. Therefore, the pull up should be disabled first. Documentation update to follow in a separate PR, accompanying documentation updates for other recent changes.

The second commit introduces a simple test sequence that randomly resets the FIFO to check that the FIFO level and status indications are updated as expected in response to the resets. Some further DV work will be required later to extend this sequence to exercise resetting of the Rx FIFO.

Fixes #17300

### DIFF
--- a/hw/ip/usbdev/data/usbdev.hjson
+++ b/hw/ip/usbdev/data/usbdev.hjson
@@ -1158,7 +1158,6 @@
         }
       ]
     }
-
     { name: "wake_control",
       desc: "USB wake module control for suspend / resume",
       swaccess: "wo",
@@ -1197,7 +1196,6 @@
         },
       ]
     }
-
     { name: "wake_events",
       desc: "USB wake module events and debug",
       swaccess: "ro",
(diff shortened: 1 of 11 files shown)


a-will (approved):

alees24:
CI failures unrelated

Issue #18511, opened 2023-05-10
[adc_ctrl] Issues with the current design


msfschaffner:
### Description

@zi-v has identified a few potential issues with the current `adc_ctrl` design and it would be good to get feedback from the firmware devs to whether these issues require hardware alignments, or whether SW workarounds are good enough. Note however that we do not plan to push any RTL changes for the shuttle, hence any alignments would have to happen after M2.5.1.

## 1) ADC Analog Range

The `adc_ctrl` spec always refers to an ADC range of 0 - 2.2V to calculate the configuration constants.
The physical ADC implementation however has a range of 0 - 2.3V.
It has been designed that way so that there is 0.1V of margin in the range.

~~**No immediate action should be necessary other than that the documentation should be updated to reflect this.**~~

This is fixed for Earlgrey.PROD in #21796

##  2) No Noise Filter in `adc_ctrl`

The current implementation of `adc_ctrl` does not have a noise filter.
Instead, it just debounces the measured values against a set of thresholds, and declares a match if the measured value has fulfilled the comparison for a specified amount of cycles.

There is a concern that this may be problematic from a noise perspective, since the thresholds may have to be relaxed to reliably detect the relevant scenarios.
This in turn may lead to more false positives, and when used in low-power mode, lead to more spurious wake-up requests, thus affecting power consumption.

~~**This does not seem critical for the shuttle tapeout, but we should assess whether this is a problem for production.**~~

This has been assessed and we determined that this does not have to be fixed for Earlgrey.PROD.

## 3) `adc_ctrl` FSM issues

The `adc_ctrl` FSM has a low-power mode where it periodically samples the ADC at a low rate.
If a matching condition has been detected in that mode, the FSM transitions into a normal power mode where ADC sampling frequency is increased in order to confirm the matching condition.

The concern here is that the current FSM implementation does not seem to have a way to automatically fall back from normal power to low-power without software intervention.
This is problematic if the system is in a sleep state, since the wakeup request will only be triggered if the match is confirmed.
That means that there could be scenarios where the `adc_ctrl` gets stuck in the normal power mode, thus affecting the overall power consumption during any of the sleep modes.

~~**This does not seem critical for the shuttle tapeout, but we should modify the FSM for production silicon so that this cannot happen. A potential workaround would be to set the debounce counter to 0 or 1 (to be confirmed) in order to minimize the amount of extra samples that need to be taken before sending out a wakeup request, thereby reducing the likelihood that this can happen.**~~

[...]

PR #21796, opened 2024-03-01
[adc_ctrl/doc] Update programming guidance for 0.0V 2.3V full range

msfschaffner:
As mentioned on #18511, the full range of the ADC in Earlgrey is 0.0V - 2.3V. 
This patch updates the programming guidance accordingly.

### DIFF
--- a/hw/ip/adc_ctrl/doc/programmers_guide.md
+++ b/hw/ip/adc_ctrl/doc/programmers_guide.md
@@ -45,13 +45,14 @@ The debug accessory voltage range of interest is shown in the diagram below:
 The ADC can be used to detect debug cable connection / disconnection in the non-overlapping regions.
 As an example use case of the two channel filters they can be used for detection of a USB-C debug accessory.
 The ADC must meet some minimum specifications:
-* Full scale range is 0.0V to 2.2V
-* If the signal is below 0.0V the ADC value will be zero.
-* If the signal is above 2.2V the ADC value will be maximum (i.e. same as 2.2V)
-* Absolute maximum error +/- 15 mV in the 0.25 - 0.45 V range
-* Absolute maximum error +/- 30 mV in the rest of the 0.0 - 2.2 V range
+* Full scale range must be 0.0V to 2.2V (note: the ADC in Earlgrey has a full range of 0.0V - 2.3V)
+* If the signal is below 0.0V the ADC value must be zero.
+* If the signal is above the full scale value, the ADC value must be maximum
+* Absolute maximum error must be +/- 15 mV in the 0.25 - 0.45 V range
+* Absolute maximum error must be +/- 30 mV in the rest of the 0.0 - 2.2 V range
 
 The following assumes:
+* Full scale range is 0.0V to 2.3V
 * The slow clock runs at 200kHz or 5 us.
 * The ADC requires 30 us to power on.
 * The ADC takes a single sample in 44 clocks (220 us)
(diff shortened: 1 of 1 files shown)


hw/ip/adc_ctrl/doc/programmers_guide.md
```diff
@@ -45,13 +45,14 @@ The debug accessory voltage range of interest is shown in the diagram below:
...
 The ADC must meet some minimum specifications:
-* Full scale range is 0.0V to 2.2V
-* If the signal is below 0.0V the ADC value will be zero.
-* If the signal is above 2.2V the ADC value will be maximum (i.e. same as 2.2V)
-* Absolute maximum error +/- 15 mV in the 0.25 - 0.45 V range
-* Absolute maximum error +/- 30 mV in the rest of the 0.0 - 2.2 V range
+* Full scale range must be 0.0V to 2.2V (note: the ADC in Earlgrey has a full range of 0.0V - 2.3V)
+* If the signal is below 0.0V the ADC value must be zero.
+* If the signal is above 2.2V the ADC value must be maximum (i.e. same as 2.2V)
+* Absolute maximum error +/- 15 mV must be in the 0.25 - 0.45 V range
```
mdhayter, line 51: Nit: I think this would be better with the must be earlier?

```suggestion
* Absolute maximum error must be +/- 15 mV in the 0.25 - 0.45 V range
```

hw/ip/adc_ctrl/doc/programmers_guide.md
```diff
@@ -45,13 +45,14 @@ The debug accessory voltage range of interest is shown in the diagram below:
...
+* Absolute maximum error +/- 30 mV must be in the rest of the 0.0 - 2.2 V range
```
mdhayter, line 52:
```suggestion
* Absolute maximum error must be +/- 30 mV in the rest of the 0.0 - 2.2 V range
```

mdhayter (commented):
Nit for making more readable

hw/ip/adc_ctrl/doc/programmers_guide.md
msfschaffner: ah yes - what I wrote did not make any sense :D

hw/ip/adc_ctrl/doc/programmers_guide.md
msfschaffner: dito

hw/ip/adc_ctrl/doc/programmers_guide.md
```diff
@@ -45,13 +45,14 @@ The debug accessory voltage range of interest is shown in the diagram below:
 The ADC can be used to detect debug cable connection / disconnection in the non-overlapping regions.
 As an example use case of the two channel filters they can be used for detection of a USB-C debug accessory.
 The ADC must meet some minimum specifications:
-* Full scale range is 0.0V to 2.2V
-* If the signal is below 0.0V the ADC value will be zero.
-* If the signal is above 2.2V the ADC value will be maximum (i.e. same as 2.2V)
-* Absolute maximum error +/- 15 mV in the 0.25 - 0.45 V range
-* Absolute maximum error +/- 30 mV in the rest of the 0.0 - 2.2 V range
+* Full scale range must be 0.0V to 2.2V (note: the ADC in Earlgrey has a full range of 0.0V - 2.3V)
```
zi-v, line 48: How about: Full scale range must be 0.0V to <at least> 2.2V ...

hw/ip/adc_ctrl/doc/programmers_guide.md
```diff
@@ -45,13 +45,14 @@ The debug accessory voltage range of interest is shown in the diagram below:
...
+* If the signal is below 0.0V the ADC value must be zero.
+* If the signal is above 2.2V the ADC value must be maximum (i.e. same as 2.2V)
```
zi-v, line 50: How about: If the signal is above  the full scale value, the ADC ...?

msfschaffner: Fixed!

moidx (approved):

jettr (commented):
Sorry, I thought I posted these comment earlier

msfschaffner:
> Full scale range must be 0.0V to 2.2V

Are you referring to the "The following assumes:" section?
That was changed to 2.3V as it states the assumptions that were used to create the table below, which would not be of much use if generated for 2.2V as then this would not match with the Earlgrey design.

hw/ip/adc_ctrl/doc/programmers_guide.md
zi-v: Sorry, I meant to say "... 0.0V to <AT LEAST> 2.2V ..."

mdhayter: The list is about what is needed for the detection of the USB-C debug accessory. Where it is important to resolve down to 0.0V and up to 2.2V.

It should probably say "The full scale range must cover 0.0V to 2.2V" or "The full scale range must include 0.0V to 2.2V" or "The ADC must produce valid results in the range 0.0V to 2.2V". Then the parenthetical comment shows that the ADC in the Earlgrey part covers this range and a little more. (Of course if someone else was building another discrete chip from the Earlgrey confing and they had an ADC that connected to the controller and worked from 0.0V to 2.7V that would be fine too -- although they would need a different adjustment to the table)

Or split into two
* The minimum voltage for the ADC to produce a valid result is 0.0V or less
* The maximum voltage for the ADC to produce a valid result is 2.2V or more

jettr:
> > Full scale range must be 0.0V to 2.2V
> 
> Are you referring to the "The following assumes:" section? That was changed to 2.3V as it states the assumptions that were used to create the table below, which would not be of much use if generated for 2.2V as that would not match with the Earlgrey design.
> 
> The list here only enumerates the minimum requirements.
> 
> I can expand the parenthetical or place another comment to make that clear.

I like Mark's suggestion of changing the text to say that the full scale range must __include__ 0.0V to 2.2V (instead of must __be__) or some variation of that.

hw/ip/adc_ctrl/doc/programmers_guide.md
msfschaffner: I tweaked the list as suggested in #21848 - PTAL

PR #21800, opened 2024-03-02
[flash_ctrl/dv] carry over DV fix to ip_autogen

msfschaffner:
This needs to be carried over to ip_autogen since the IPgen transition has not been fully completed yet.


matutem (approved):
LGTM

PR #21802, opened 2024-03-02
[hw,soc_proxy,rtl] Synchronize external alert signals

Razer6:
Use a synchronizer primitive for differentially encoded external alert signals. External alerts don't use the custom alert protocol. Instead, they are just differentially encoded signals, which require a sync cell to pass the clock domain crossing.

I removed the sec anchor cell sincy we now have a flop here anyways.

### DIFF
--- a/hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
+++ b/hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
@@ -148,41 +148,142 @@ module soc_proxy
                                            reg2hw.alert_test.recov_alert_external_3.q;
 
   // Handle fatal external alert requests
-  logic [NumFatalExternalAlerts-1:0] soc_fatal_alert_p, soc_fatal_alert_n, fatal_alert_external;
+  logic [NumFatalExternalAlerts-1:0] fatal_alert_external, fatal_alert_external_async;
+  logic [NumFatalExternalAlerts-1:0] soc_fatal_alert_p, soc_fatal_alert_n;
+  logic [NumFatalExternalAlerts-1:0] soc_fatal_alert_ack_p_d, soc_fatal_alert_ack_n_d;
+
+  // Acknowledge alert based on request.
+  // Ensure that Ack has always a valid encoding and we swallow a staggered _p/_n signal
+  always_comb begin
+    for(int i = 0; i < NumFatalExternalAlerts; i++) begin
+      // Acknowledge alert based on request.
+      // Ensure that Ack has always a valid encoding and we swallow a staggered _p/_n signal
+      if (soc_fatal_alert_p[i] ^ soc_fatal_alert_n[i]) begin
+        soc_fatal_alert_ack_p_d[i] = soc_fatal_alert_p[i];
+        soc_fatal_alert_ack_n_d[i] = soc_fatal_alert_n[i];
+      end else begin
+        soc_fatal_alert_ack_p_d[i] = soc_fatal_alert_o[i].ack_p;
+        soc_fatal_alert_ack_n_d[i] = soc_fatal_alert_o[i].ack_n;
+      end
+    end
+  end
+
   for (genvar i = 0; i < NumFatalExternalAlerts; i++) begin : gen_fatal_alert_handling
-    // Buffer/anchor incoming signals to prevent optimization
-    prim_sec_anchor_buf #(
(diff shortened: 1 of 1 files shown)


hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
```diff
@@ -168,21 +180,59 @@ module soc_proxy
...
+    ) u_prim_flop_ack_p (
+      .clk_i,
+      .rst_ni,
+      .d_i    (soc_alert_ack_p_d),
+      .q_o    (soc_recov_alert_o[i].ack_p)
+    );
+
+    prim_flop #(
+      .Width(1),
+      .ResetValue(1'b<)
```
andreaskurth, line 229: RTL code typo
```suggestion
      .ResetValue(1'b1)
```

hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
```diff
@@ -168,21 +180,59 @@ module soc_proxy
...
     );
     // Treat any positive value on `alert_p` and any negative value on `alert_n` as alert.
     assign recov_alert_external[i] = soc_recov_alert_p[i] | ~soc_recov_alert_n[i];
+
     // Acknowledge alert based on request.
-    assign soc_recov_alert_o[i] = '{
-      ack_p: soc_recov_alert_p[i],
-      ack_n: soc_recov_alert_n[i]
-    };
+    // Ensure that Ack has always a valid encoding and we swally a staggered _p/_n signal
```
andreaskurth, line 208:
```suggestion
    // Ensure that Ack has always a valid encoding and we swallow a staggered _p/_n signal
```

andreaskurth, line 208:
```suggestion
    // Ensure that Ack has always a valid encoding and we swallow a staggered _p/_n signal
```

hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
```diff
@@ -150,12 +150,24 @@ module soc_proxy
...
-    ) u_prim_sec_anchor_buf (
-      .in_i ({soc_fatal_alert_i[i].alert_p, soc_fatal_alert_i[i].alert_n}),
-      .out_o({soc_fatal_alert_p[i], soc_fatal_alert_n[i]})
+    // Synchronize external differentially encoded alert to internal clk domain
+    prim_flop_2sync #(
+      .Width(1),
+      .ResetValue(1'b0)
+    ) u_prim_flop_2sync_fatal_alert_p (
+      .clk_i,
+      .rst_ni,
```
andreaskurth, line 159: I think we have to use the AON domain here and for the other `prim_flop_2sync`s added in this PR, otherwise external alerts will no longer cause a wakeup if OT's non-AON clock is gated. Prior to this commit, external alerts went into the `async_wkup` signal combinatorially; now that there are flops on this path, these flops need to be in the AON domain

Razer6: External are alerts coming from the external fast clock domain, which is the same as the internal fast domain, though, this might be gated. To avoid crossing that many clock domains (Fast -> AON -> Fast for the alert) for the alerts, would it make sense to keep it like this, but additionally compute `fatal_alert_external` combinational from the inputs and feed that to `async_wkup `?

Razer6:
@msfschaffner @andreaskurth As discussed yesterday, here is an implementation as we discussed. Please can you review that as it is a priority for us to freeze upstream.

hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
msfschaffner: yes, computing that combinatorially and feeding into `async_wkup` LGTM.

msfschaffner (approved):
LGTM

msfschaffner:
CHANGE AUTHORIZED: hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv

msfschaffner:
CHANGE AUTHORIZED: hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv

PR #21807, opened 2024-03-03
[dv,usbdev] Testplan extensions for usbdev

alees24:
Additional test points for USBDEV, particularly stress testing and conditions that are impractical or difficult to simulate at top level/FPGA such as random disconnections, bus resets and unexpected/invalid traffic.
These test points are not required for V1, but should go some way towards scoping out the requirements for V2.

Introduce the recently-created FIFO reset test following RTL change in #21791


rswarbrick (approved):

PR #21808, opened 2024-03-03
[dv,sram_ctrl] Improve detection of init_done and key_scr_valid

matutem:
The older scheme for this detection used csr_spinwait for the duration of init and key request. This only uses csr_spinwait once the request is satisfied, to detect the actual CSR updates. This is just as accurate as the older scheme, and by avoiding csr_spinwait for the full time it simplifies handling of random reset.

### DIFF
--- a/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_smoke_vseq.sv
+++ b/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_smoke_vseq.sv
@@ -100,7 +100,6 @@ class sram_ctrl_smoke_vseq extends sram_ctrl_base_vseq;
         req_mem_init();
       end
 
-
       // Do some random memory accesses
       `uvm_info(`gfn,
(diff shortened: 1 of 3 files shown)


rswarbrick (approved):
This looks like an improvement, I think. (And I like the way we're losing some magic waits!)

As a more general point, I'm slightly surprised to see something in the scoreboard driving TL transactions (`csr_spinwait` added here / `csr_wr` that was there already). I'm not really convinced this is the right way to do things, but it's also not changed in this PR, so I guess it's a question for another day :-)

matutem:
@rswarbrick This csr_spinwait is to synchronize the scoreboard to a hardware update of the corresponding CSRs, since dv can not accurately predict when that will happen, and if it tried it can easily end up as one of those very flaky mechanisms.

rswarbrick:
Thinking again this morning, I thought it would be "more passive" with `.backdoor(1)`... which I see is what you did! Sorry for the noise :-)

PR #21809, opened 2024-03-03
[hmac] Change fifo_empty interrupt type to status

### LINKED ISSUE #21206: [hmac] Change fifo_empty interrupt from event to status type
### Description

This is analogous to #21049:
- hmac_done: is fine as event
- fifo_empty: should be changed to status
- hmac_err: is fine as event

vogelpi:
This resolves #21206.

This converts the HMAC Message FIFO empty interrupt from type event to status. Similarly to KMAC (#21657), the design is further changed to make the interrupt fire only if the message FIFO has run full previously. This can happen e.g. if the SHA2 engine is busy and cannot accept more data currently. In other cases, hardware will consume the data faster than software can provide it and there is no point in interrupting the software to tell it to work faster.

Also, the interrupt is not firing if 1) the block is running in HMAC mode and doing the second phase where the Message FIFO is managed entirely by the hardware, or 2) if software triggered the Process or Stop command already (no more input data is accepted).


hw/ip/hmac/rtl/hmac.sv
```diff
@@ -299,10 +291,70 @@ module hmac
...
+  //   processing a block, it always empties the message FIFO faster than software can fill it up,
+  //   meaning the message FIFO is empty most of the time. Note, the empty status is signaled only
+  //   once after the FIFO was completely full. The FIFO needs to be full again for the empty
+  //   status to be signaled again next time it's empty.
+
+  // FIFO empty interrupt
+  // The message FIFO empty interrupt is not useful for software if:
+  // - The HMAC block is currently feeding the result of the first round message into the FIFO
+  //   for the second round. This step is managed by the hardware.
+  // - The HMAC block is completely idle. The FIFO should anyway be empty in that case.
```
andreaskurth, line 319: Is this second _FIFO empty interrupt_ comment still relevant? I think the relevant points are listed in the first comment, right? If not all points have been integrated into the first comment, I suggest doing so

hw/ip/hmac/rtl/hmac.sv
```diff
@@ -299,10 +291,70 @@ module hmac
...
+
+  // Track whether the FIFO was full after being empty. We clear the tracking:
+  // - When receiving the Start, Process or Stop command. This is to start over for the next
+  //   message.
+  // - When seeing a negative edge on the empty signal. This signals that software has reacted to
+  //   the interrupt and is filling up the FIFO again.
+  assign fifo_full_seen_d =
+      fifo_full                                           ? 1'b 1 :
+      fifo_empty_negedge                                  ? 1'b 0 :
+      reg_hash_start || reg_hash_process || reg_hash_stop ? 1'b 0 : fifo_full_seen_q;
```
andreaskurth, line 333: Since this gets cleared on _start_, should it also get cleared on _continue_?

andreaskurth (approved):
Thanks for this work, @vogelpi :+1: Glad to see that the DV env can be updated to model and check this behavior without very intrusive modifications. I left two comments/questions, but otherwise this LGTM!

hw/ip/hmac/rtl/hmac.sv
vogelpi: Ah, sorry this is a mistake. When rebasing, I've failed to remove this comment from an earlier version of the new implementation.

hw/ip/hmac/rtl/hmac.sv
vogelpi: Yes, I think so!

vogelpi:
> Thanks for this work, @vogelpi 👍 Glad to see that the DV env can be updated to model and check this behavior without very intrusive modifications. I left two comments/questions, but otherwise this LGTM!

Thanks for your review @andreaskurth . unfortunately, it turns out the DV env changes are much more complicated. With what was in the PR before, the pass rate was not satisfactory and I currently don't have the time to bring it to an acceptable level. Since there are still other more important HMAC changes for which DV is in a better shape. I've now decided to disable verification for the FIFO empty status interrupt. This should be acceptable as no software is currently using this interrupt (previously, one test used the interrupt but the test would have to be refactored anyway to make use of the status type interrupt which is not a priority).

The verification of the status type FIFO empty interrupt is tracked in #21815.

andreaskurth:
SGTM :+1:

PR #21810, opened 2024-03-04
[sival] Backport entropy_src_fw_override_test to earlgrey_es_sival

jwnrt:
There are two main changes from the version on `master` required due to RTL differences:

1. `entropy_testutils_drain_observe_fifo` must manually clear the FIFO overflow flag.
2. The test "payload" does not run OTBN because it requires the FIPS bit to be set which isn't possible from software. Instead, 3 repeats of the AES and KMAC payloads are run.


pamaury (approved):

vogelpi (approved):
LGTM!

vogelpi:
Out of interest, have you ever counted how many seeds/interrupts the entropy_src generates per test run?

jwnrt:
I haven't collected those stats, but the test does write `19488` words total to the FW override FIFO for the `master` branch version (so that's 10 x the AES, KMAC, and OTBN payloads).

vogelpi:
Thanks, I think that should be fine. Just wanted to get an idea for the amount of entropy that's actually requested by the consumers.

PR #21811, opened 2024-03-04
[dv] Update spi-device testplan for V1 sign-off.

antmarzam:
The testplan has been updated to directly mention the new RTL features for V1 sign-off:
- equivalent r/w access when using 1r1w SRAMs as opposed to the 2PSRAM
- Mention of read_pipeline feature as a testpoint
- new covergroup to sample the type of SRAM instantiated
- addition of read_pipeline_mode coverpoint to existing flash_cmd_info_cg

### DIFF
--- a/hw/ip/spi_device/data/spi_device_testplan.hjson
+++ b/hw/ip/spi_device/data/spi_device_testplan.hjson
@@ -205,6 +205,17 @@
       tests: ["spi_device_intercept", "spi_device_flash_all"]
     }
     {
+      name: cmd_read_pipeline
+      desc: '''
+            - Configure proper fast read command info slot with the read data pipeline feature.
+            - Issue fast read command.
+            - Check propagation of fast read command, including dummy cycles for read data pipeline.
+            - Initiate response to the fast read.
+            - Check proper reception of response.'''
+      stage: V2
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
The dual vs. single point change won't quite work like this.

We'll essentially need to parameterise the testbench into two different versions (2P vs. 1R1W). Would you mind dropping that part from this PR? (1st and 3rd hunk) I think it might even be on my plate to get this sorted out (eek...)

The read pipeline part looks sensible to me.

antmarzam:
> The dual vs. single point change won't quite work like this.
> 
> We'll essentially need to parameterise the testbench into two different versions (2P vs. 1R1W). Would you mind dropping that part from this PR? (1st and 3rd hunk) I think it might even be on my plate to get this sorted out (eek...)
> 
> The read pipeline part looks sensible to me.
 
Ok - I'll do that. @hcallahan-lowrisc and I had a brief discussion on this last Friday and concluded that doubling the regression for a new RTL configuration (with the alternative SRAM parameter) would probably be overkill and a a comprehensive test using the alternative 1r1w parameter would be enough. 
Nonetheless, I'll trim those 1r1w bits of the testplan from this PR and we'll figure out how we tackle the SRAM parameter testing!

rswarbrick (approved):

PR #21812, opened 2024-03-04
[spi_device,dv] Split sim config into separate 1r1w and 2p versions

rswarbrick:
This follows the scheme used for the aes block (where we select between masked/unmasked implementations).

The "implementation" is that we pass a define (SRAM_TYPE). This essentially does the same job as the SRAM_TYPE parameter that was in spi_device_env_pkg.

But now we have a pair of _sim_cfg.hjson files that implement the two modes. This is equivalent to aes_masked_sim_cfg.hjson / aes_unmasked_sim_cfg.hjson.

@msfschaffner: Is this what you were imagining? The notes from the meeting on Friday talked about noting this in the testplan, but the AES testplan (from which I nicked the design) doesn't get split up like that. If there is a better way, I'm happy to rejig it.


hw/ip/spi_device/dv/base_sim_cfg.hjson
```diff
@@ -47,6 +47,15 @@
...
   run_opts: ["+cdc_instrumentation_enabled=1"]
 
+  build_modes : {
+    spi_device_1r1w: {
+      build_opts: ["+define+SRAM_TYPE=spi_device_pkg::SramType1r1w"]
+    }
+    spi_device_2p: {
+      build_opts: ["+define+SRAM_TYPE=spi_device_pkg::SramType2p"]
+    }
+  }
```
msfschaffner, line 57: Can you remove the build modes and pipe the plusdefine in via buildopts as done for lc_ctrl?

https://github.com/lowRISC/opentitan/blob/7e196e6f072791aa17817b7691aed6b258399eeb/hw/ip/lc_ctrl/dv/lc_ctrl_volatile_unlock_disabled_sim_cfg.hjson#L13-L14

I ran into issues collecting coverage with this variant previously.

msfschaffner (approved):
Yes that's what I meant!

a-will (commented):
Seems good to me, but I'll duck out again and let you experts work. I only dabble here and am unfamiliar with the spectrum of choices and their effects. :smile:

hw/ip/spi_device/dv/base_sim_cfg.hjson
rswarbrick: Yep, happily. I've just pushed up a version that I *think* should work. Leaving this for you to glance at before it gets merged but please feel free to hit merge if it looks sensible to you.

msfschaffner:
LGTM, although it seems like it needs another `make -C hw`...

msfschaffner (approved):

rswarbrick:
Oh! I think that rebasing should suffice (so that I pick up 7771831eae41ccf0e2c2a0cdac5bef933bb6d62e). I've just done so. I'll merge if the quick lint check passes.

PR #21813, opened 2024-03-04
[i2c,rtl] Predict target clock stretching in HOST mode

### LINKED ISSUE #18962: [i2c,rtl] Clock stretch causes timing mismatch
When entering the ClockPulse state from the ClockLow state in host mode, `scl_d` goes from 0 to 1. However, the counter that determines the timing of `t_high` only starts counting down when the `scl_i` sees this output. This is necessary for detecting clock stretching, but in the non-clock stretching case, this adds effectively adds another 3 cycles to the `t_high` time.

https://github.com/lowRISC/opentitan/blob/7545bb6eac5207c29c5413fa596b9e219403f200/hw/ip/i2c/rtl/i2c_fsm.sv#L169-L170

https://github.com/lowRISC/opentitan/blob/7545bb6eac5207c29c5413fa596b9e219403f200/hw/ip/i2c/rtl/i2c_fsm.sv#L200-L201

This issue is based on this comment: https://github.com/lowRISC/ot-discrete-eg/issues/11#issuecomment-1575126104 (restricted access).

hcallahan-lowrisc:
> Ignore first commit for reviewing purposes, comes from #21765 

From the commit message : 

> Previously, the cycle counter for any 'thigh' state would only decrement once
the FSM had observed scl_i = 1'b1 on it's input. When the FSM releases SCL to
create the clock pulse, a minimum of 3 cycles is required to observe this effect
on it's input. (1-cycle output flop, 2-cycle input synchronizer)
> This change allows the HOST-mode state machine to proceed if it does not
observe SCL being stretched 4 cycles after releasing SCL to try and create the
next clock pulse. This removes the 3-cycle delay from every 'thigh' state, and
in the absence of clock-stretching, brings the performance of the block in-line
with a user's calculations based on the given timing parameters.
> This imposes a minimum 'thigh' of 4 cycles, which may limit the possible
performance of the block when the frequency of clk_i is not significantly
greater than scl.

This approach seemed the simplest way to achieve the intended result, but one cost is the minimum 'thigh' of 4 as described above. I could not come up with a solution that avoided this without greatly increasing complexity, and I'm not even sure one is possible. If we advance on too quickly and reach the "next" tlow, the FSM drives SCL low again, and at this point if the target is stretching we cannot distinguish it from our own driver.

There are some DV gremlins from this change that I have still to resolve (similar to #21765). I am again proposing to fix these up as a follow-up item.

Testing the performance with this change + that from #21765, I see clock frequencies within the 4-cycle possible float as a result of the calculation `tlow + t_r + thigh + t_f`. This is as good as we can do without stacking the remainders of the quantization calcs, which may be possible in some cases to squeeze out an extra cycle or two.

Closes #18962
Goes toward #18492


hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -921,9 +932,15 @@ module i2c_fsm import i2c_pkg::*;
...
+          if (!scl_i && stretch_predict_cnt_expired) begin
+            // The target is (still) stretching the clock.
+            state_stretch_d = ClockPulseAck;
+            state_d = ClockStretching;
+          end else begin
+            state_d = HoldDevAck;
+            load_tcount = 1'b1;
+            tcount_sel = tHoldBit;
+          end
         end
```
a-will, line 944: For what it's worth, this probably could be reduced to a form that doesn't add another state:

```systemverilog
        if (!scl_i && stretch_predict_cnt_expired) begin
          // Saw stretching. Remain in this state and don't count down pulse until we see clock high.
          load_tcount = 1'b1;
          tcount_sel = tClockPulse;
        end else if (tcount_q == 20'd1) begin
          state_d = HoldDevAck;
          load_tcount = 1'b1;
          tcount_sel = tHoldBit;
        end
```

There might be an off-by-one error there with the cycle when `stretch_predict_cnt_expired` lands. I didn't give this a thorough evaluation.

Does that seem like it would work and be simpler, though?

hw/ip/i2c/rtl/i2c_pkg.sv
```diff
@@ -12,4 +12,48 @@ package i2c_pkg;
     AcqRestart = 2'b11
   } i2c_acq_byte_id_e;
 
+  // State definitions
+  typedef enum logic [5:0] {
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Good point, I'll double check what others are doing. This was just to make it easier for some local DV code to reuse the type.

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -85,6 +85,8 @@ module i2c_fsm import i2c_pkg::*;
   logic        load_tcount;   // indicates counter must be loaded
   logic [31:0] stretch_idle_cnt; // counter for clock being stretched by target
                                  // or clock idle by host.
+
+  // This bit is active when the FSM is in a state where a TARGET might be trying to stretch
```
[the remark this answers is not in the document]
hcallahan-lowrisc: SGTM, I'll do a quick review of the spec's for terminology and try to align with them.

hw/ip/i2c/rtl/i2c_fsm.sv
hcallahan-lowrisc: Ah this seems much simpler, I think it should work to just hold onto `load_tcount` while stretching. Let me try that.

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -166,10 +171,10 @@ module i2c_fsm import i2c_pkg::*;
         tNoDelay    : tcount_d = 20'h00001;
         default     : tcount_d = 20'h00001;
       endcase
-    end else if (stretch_idle_cnt == '0 || target_enable_i) begin
+    end else if (host_enable_i || target_enable_i) begin
       tcount_d = tcount_q - 1'b1;
     end else begin
-      tcount_d = tcount_q;  // pause timer if clock is stretched
+      tcount_d = tcount_q;
```
marnovandermaas, line 177: Do we still need this else statement? It only covers the case when host and target mode are both disabled in which case we are in the Idle state anyways.

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -204,6 +210,30 @@ module i2c_fsm import i2c_pkg::*;
...
+  // The TARGET can stretch the clock during any time that the host drives SCL to 0.
+  // However, we (the HOST) cannot know it is being stretched until we release SCL,
+  // usually trying to create the next clock pulse.
+  // Due to the 3-cycle round trip (1-cycle output flop, 2-cycle input synchronizer),
+  // even if the TARGET is not stretching the clock, we cannot observe the effect of
+  // releasing the clock until 3 cycles later.
+  //
+  // 'stretch_predict_cnt_expired' becomes active once we have observed 4 cycles of
+  // delay, and scl_i = 1'b0 at this point we know that the Target is stretching the
+  // clock. This requires 'thigh >= 4' to guarantee we don't miss stretching, which
```
marnovandermaas, line 222: Can we add an assert to check that `thigh` is always greater than 3?

marnovandermaas (commented):
Thanks for making these changes. I agree with Alex that it would be nice to remove the second state variable if possible. Other than that just a few minor comments.

hw/ip/i2c/rtl/i2c_pkg.sv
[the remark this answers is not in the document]
matutem: We sometimes add state enums in unit packages for the sake of DV (see pwrmgr_pkg.sv for example), but this may need a prefix to associate the states to a specific FSM. I guess that is an RTL change that should not change functionality.

hw/ip/i2c/rtl/i2c_fsm.sv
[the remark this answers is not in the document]
andreaskurth: I agree with keeping terminology aligned with the spec. Please don't extend your terminology alignment efforts to beyond the things you're changing, though, because we may have to go through the entire spec and change wording soon anyway.

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -204,6 +210,30 @@ module i2c_fsm import i2c_pkg::*;
...
+  // is the 3-cycle round trip +1 for potential metastability on the input.
+  logic stretch_predict_cnt_expired;
+  always_ff @ (posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      stretch_predict_cnt_expired <= 1'b0;
+    end else begin
+      if (stretch_idle_cnt == 32'd3) begin
```
a-will, line 229: Oh, tClockPulse includes the rise time, so we need to allow for that as well, right?

```suggestion
      if (stretch_idle_cnt == (32'd3 + t_r_i)) begin
```

hw/ip/i2c/rtl/i2c_fsm.sv
hcallahan-lowrisc: > An optimization for this:

Agreed, I've implemented this in the latest push.

hw/ip/i2c/rtl/i2c_fsm.sv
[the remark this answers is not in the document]
a-will: You don't have the latest with the cleaning of sensitive terms, hehe. Make sure you're looking at 7.0.

hw/ip/i2c/rtl/i2c_pkg.sv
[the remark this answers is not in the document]
hcallahan-lowrisc: This has now been refactored to not add any new states, and hence I'll just leave the enum where it is. I plan to submit some DV changes in the future which make use of this enum, so I will rehome it them.

hw/ip/i2c/rtl/i2c_fsm.sv
hcallahan-lowrisc: I think I agree with you, but I need to think about it a bit more to be sure. That would imply that we document a minimum 
 value `thigh >= t_r + 4` to be safe I think. As before, probably only a downside for very low `clk_i` input frequencies.

hw/ip/i2c/rtl/i2c_fsm.sv
hcallahan-lowrisc: We could do that, but there strictly isn't any reason why a user couldn't program it with the value of 0. That's the reset value after all, so it's probably worth testing in some capacity for error recovery. There's the (unused) switch `program_incorrect_regs` in the base class which probably anticipated testing this at some point. I guess we could always add the assertion and then disable it for that sort of testing.

hw/ip/i2c/rtl/i2c_fsm.sv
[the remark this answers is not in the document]
hcallahan-lowrisc: Ah yes of course. We still link to the Rev 6. version in the docs. Yet another thing to update!

hw/ip/i2c/rtl/i2c_fsm.sv
a-will: Oh, no, it's still `thigh >= 4`. Remember, `tClockPulse` _already_ includes `t_r` as well. The condition you were satisfying was that the timeout happens before we could possibly change out of the state. In other words...

```
tStretchTimeout < tClockPulse
-------------------------------------
        t_r + 3 < t_r + t_high
-------------------------------------
              3 < t_high
```

a-will (approved):
LGTM. Thanks!

marnovandermaas:
Kokoro seems unhappy about:
`E   ARITH_CONTEXT:   i2c_fsm.sv:227   Bitlength of arithmetic operation '16'd3 + t_r_i' is self-determined in this context`

I guess it means that the `16'` is unnecessary because `t_r_i` is already 16 bits?

hw/ip/i2c/rtl/i2c_fsm.sv
marnovandermaas: Ok, that makes sense.

marnovandermaas (approved):
These changes make sense to me. Bar the Kokoro Lint error.

andreaskurth:
Rebased on `master` to use the first commit from the now merged PR #21765 and clean failing CI checks (which should have been fixed on `master` in the meantime).

andreaskurth:
The failing CW310 tests are phony (sourceforge.net problem). The failing `chip_sw_i2c_host_tx_rx` test, which is part of smoke tests, is real and happens before and after the rebase.

hcallahan-lowrisc:
The failure of the `chip_sw_i2c_host_tx_rx` test was due to the monitor losing it's lock on the bus again, similar to failures initially encountered in #21765 and patched to be rethought later as part of closing #21887. This time however we were losing the lock around the ACK bit.
I've weakened the checks to remove the setup time here, which fixes that particular test. Assuming the remaining test suite also passes, I think it may be best to expand #21887 to cover these changes too.

andreaskurth:
> I've weakened the checks to remove the setup time here, which fixes that particular test. Assuming the remaining test suite also passes, I think it may be best to expand #21887 to cover these changes too.

I'm OK with that; however now the `i2c_host_smoke` test fails, apparently due to an RTL assertion that fails: `(i2c_fsm.sv:1423) [ASSERT FAILED] SclInputGlitch_A`.

andreaskurth (approved):
LGTM, thanks for this work, @hcallahan-lowrisc. If CI passes (modulo the currently known unrelated problems), please go ahead with merging this PR.

hcallahan-lowrisc:
CI failures are known-issues, and are unrelated to this PR. Hence merging.

PR #21816, opened 2024-03-04
[usbdev,dv] Drop some arguments from the call_token_seq task

rswarbrick:
The values are the same at every call site. Get rid of the argument, since it's just confusing!


hcallahan-lowrisc (approved):
LGTM, thanks for the cleanup :+1:

rswarbrick:
Huh, I've just realised that this doesn't quite work the way I imagined (because of the way subclasses with initialisers for shadowed names work in SystemVerilog). I'll tidy it up and re-request a review when I'm convinced.

rswarbrick:
After looking more carefully, this was just confusing code which I suspect shows how different things got written / merged in parallel. Anyway, the only change needed is the first commit in the PR. Everything else is just rebased. (@hcallahan-lowrisc, would you mind taking another look?)

hcallahan-lowrisc (approved):
Yeah sounds like a merge issue to me, there's no point redefining a variable/constraint if you're not going to change it! Thanks

PR #21818, opened 2024-03-04
[usbdev,dv] Set num_of_bytes earlier in usbdev_in_trans_vseq

rswarbrick:
The existing code was pretty unclear. The m_data_pkt.data queue is created inside call_data_seq() in the base class, where its size is constrained to num_of_bytes: a variable that is passed in uninitialised.

I think that the existing implementation was picking a random 7-bit number to use as an argument, but it was *very* unclear.

This implementation is a little more obvious. I suspect there are further improvements possible, but at least it's now obvious what this does.

(@MubashirSaleem775: I think this code came from you originally. Would you mind checking that I've understood the behaviour correctly?)

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_in_trans_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_in_trans_vseq.sv
@@ -11,6 +11,9 @@ class usbdev_in_trans_vseq extends usbdev_base_vseq;
   virtual task body();
     uvm_reg_data_t rxfifo;
     bit in_sent;
+
+    num_of_bytes = $urandom_range(1, 64);
+
     super.dut_init("HARD");
     clear_all_interrupts();
     ral.intr_enable.pkt_sent.set(1'b1); // Enable pkt_sent interrupt
@@ -31,7 +34,6 @@ class usbdev_in_trans_vseq extends usbdev_base_vseq;
     // Make sure buffer is availabe for next trans
     ral.avoutbuffer.buffer.set(out_buffer_id + 1);
     csr_update(ral.avoutbuffer);
-    num_of_bytes = m_data_pkt.data.size();
     // Note: data should have been written into the current OUT buffer by the above transaction
     configure_in_trans(out_buffer_id);  // register configurations for IN Trans.
     // Token pkt followed by handshake pkt
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_sent_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_sent_vseq.sv
@@ -11,8 +11,10 @@ class usbdev_pkt_sent_vseq extends usbdev_base_vseq;
   virtual task body();
     uvm_reg_data_t read_rxfifo;
 
-     super.dut_init("HARD");
-     clear_all_interrupts();
+    num_of_bytes = $urandom_range(1, 64);
+
+    super.dut_init("HARD");
+    clear_all_interrupts();
 
     // OUT TRANS
     // -------------------------------
@@ -38,7 +40,6 @@ class usbdev_pkt_sent_vseq extends usbdev_base_vseq;
     // IN TRANS
     // --------------------------------
     // Configure in transaction
-    num_of_bytes = m_data_pkt.data.size();
     // Note: data should have been written into the current OUT buffer by the above transaction
     configure_in_trans(out_buffer_id);
     // Token pkt followed by handshake pkt


MubashirSaleem775:
@rswarbrick , there has been a modification in the implementation of these sequences. The variable "num_of_bytes" is designated for cases where a specific number of bytes is necessary for a transaction, such as in certain control transfers. In all other transaction, random data will be generated, constrained to a maximum of 64 bytes as defined in the "seq_item" class. The variable "rand_or_not" is employed to manage the randomness of the data. By default, its value is set to 1, indicating that the data will be random within the range of 0 to 64 bytes. when we need a specific number of bytes we have to initialize num_of_bytes in extended class.

rswarbrick:
Concentrating on `usbdev_pkt_sent_vseq`, I think this is what happens with the current head of master:

- We do "OUT TRANS" with a random value in `num_of_bytes`.
- This uses `call_data_seq`, which randomises `m_data_pkt`.
- The constraint on the randomisation is that the length of `m_data_pkt` should equal `num_of_bytes`.
- We then set `num_of_bytes` to equal the length of `m_data_pkt`.

With my proposed change, the sequence is:

- We pick a random value in `num_of_bytes` and pass it to `call_data_seq` (as before)
- We don't change `num_of_bytes` again because it already has the right value.

Have I misunderstood something?

hw/ip/usbdev/dv/env/seq_lib/usbdev_in_trans_vseq.sv
```diff
@@ -11,6 +11,9 @@ class usbdev_in_trans_vseq extends usbdev_base_vseq;
   virtual task body();
     uvm_reg_data_t rxfifo;
     bit in_sent;
+
+    num_of_bytes = $urandom_range(1, 127);
```
MubashirSaleem775, line 15: OT USB 2.0 accommodates packet lengths ranging from 0 to 64 bytes, with 0 being the minimum and 64 being the maximum packet size. Therefore, could you modify the 'num_of_bytes' within the range of (0, 64)?

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_sent_vseq.sv
```diff
@@ -11,8 +11,10 @@ class usbdev_pkt_sent_vseq extends usbdev_base_vseq;
   virtual task body();
     uvm_reg_data_t read_rxfifo;
 
-     super.dut_init("HARD");
-     clear_all_interrupts();
+    num_of_bytes = $urandom_range(1, 127);
```
MubashirSaleem775, line 14: same here as described above.

MubashirSaleem775:
> Concentrating on `usbdev_pkt_sent_vseq`, I think this is what happens with the current head of master:
> 
> * We do "OUT TRANS" with a random value in `num_of_bytes`.
> * This uses `call_data_seq`, which randomises `m_data_pkt`.
> * The constraint on the randomisation is that the length of `m_data_pkt` should equal `num_of_bytes`.
> * We then set `num_of_bytes` to equal the length of `m_data_pkt`.
> 
> With my proposed change, the sequence is:
> 
> * We pick a random value in `num_of_bytes` and pass it to `call_data_seq` (as before)
> * We don't change `num_of_bytes` again because it already has the right value.
> 
> Have I misunderstood something?

Yes,  this approach will work also.

hw/ip/usbdev/dv/env/seq_lib/usbdev_in_trans_vseq.sv
rswarbrick: Cool, I'll sort that now. (What an annoying size! Just big enough to need 7 bits!)

Thanks for the feedback.

hw/ip/usbdev/dv/env/seq_lib/usbdev_in_trans_vseq.sv
```diff
@@ -11,6 +11,9 @@ class usbdev_in_trans_vseq extends usbdev_base_vseq;
   virtual task body();
     uvm_reg_data_t rxfifo;
     bit in_sent;
+
+    num_of_bytes = $urandom_range(1, 64);
```
Saad22386, line 15: before this you have to set rand_or_not = 1'b0

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_sent_vseq.sv
```diff
@@ -11,8 +11,10 @@ class usbdev_pkt_sent_vseq extends usbdev_base_vseq;
   virtual task body();
     uvm_reg_data_t read_rxfifo;
 
-     super.dut_init("HARD");
-     clear_all_interrupts();
+    num_of_bytes = $urandom_range(1, 64);
```
Saad22386, line 14: same here as mentioned above

hw/ip/usbdev/dv/env/seq_lib/usbdev_in_trans_vseq.sv
rswarbrick: Hmm, I've definitely done something silly here. I'm convinced that there's an easier way to do this and will open a new PR that does it more sensibly next week.

PR #21819, opened 2024-03-04
[usbdev,dv] Remove unneeded arguments from call_data_seq

rswarbrick:
This is analoguous to #21816. Remove some unneeded arguments and sort out some slightly confusing subclass usage. It should all end up a bit clearer now.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_av_buffer_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_av_buffer_vseq.sv
@@ -20,7 +20,7 @@ class usbdev_av_buffer_vseq extends usbdev_base_vseq;
     // Out token packet followed by a data packet of 8 bytes
     call_token_seq(PidTypeOutToken);
     cfg.clk_rst_vif.wait_clks(20);
-    call_data_seq(PidTypeData0, num_of_bytes);
+    call_data_seq(PidTypeData0);
     get_response(rsp_item);
     $cast(item, rsp_item);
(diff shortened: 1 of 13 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_av_buffer_vseq.sv
```diff
@@ -11,9 +11,10 @@ class usbdev_av_buffer_vseq extends usbdev_base_vseq;
   usb20_item      item;
   RSP             rsp_item;
   bit      [6:0]  num_of_bytes = 8;
-  bit             rand_or_not = 0;
 
   task body();
+    rand_or_not = 1'b0;
```
hcallahan-lowrisc, line 15: Very unimportant suggestion: I tend to like `pre_start()` for this sort of stuff. Not for any functional reason, but just to separate it from the functional part of the seq.

hcallahan-lowrisc (approved):
LGTM

hw/ip/usbdev/dv/env/seq_lib/usbdev_av_buffer_vseq.sv
rswarbrick: I agree, but was being lazy and avoiding adding a new task! :-) I should be less lazy though... I've just pushed up a version that moves things around properly. Would you mind re-reviewing to check that I haven't done anything silly?

Saad22386:
@hcallahan-lowrisc, @rswarbrick Could you proceed with this PR so that the upcoming SEQ PR should follow the suggested stuff?

PR #21822, opened 2024-03-04
[cryptolib] Add a top-level `otcrypto.h` header

cfrantz:
Add a top-level header file that includes all other cryptolib headers.

### DIFF
--- a/sw/device/lib/crypto/BUILD
+++ b/sw/device/lib/crypto/BUILD
@@ -25,12 +25,34 @@ ot_static_library(
     ],
 )
 
+# This library imports the archive created by the static library rule
+# above. This can be used within the codebase to test the cryptolib
+# as built.
 cc_import(
     name = "crypto",
     static_library = ":otcrypto",
     target_compatible_with = [OPENTITAN_CPU],
     deps = [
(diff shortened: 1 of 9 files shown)


cfrantz:
@vsukhoml Does this header work for you?

jadephilipoom (approved):

vsukhoml:
yes, looks good!

sw/device/lib/crypto/BUILD
```diff
@@ -34,6 +34,15 @@ cc_import(
     ],
 )
 
+cc_import(
```
jadephilipoom, line 37: nit: I think a comment here might be helpful to clarify the contexts in which this target should (or should not) be used. Is it only for tests within the OT repo of the freestanding/exported lib?

jadephilipoom (approved):

sw/device/lib/crypto/BUILD
cfrantz: Fixed: added a comment.

This target is used to test that the freestanding version of the library & headers work correctly on their own, whereas the `:crypto` target can be used to test cryptolib within the codebase.

Issue #21550, opened 2024-02-19
[rom,rom_ext] integrate `otbn_boot_services` ECDSA sigverify into ROM/ROM_EXT


timothytrippel:
Integrate the ECDSA sigverify, embedded in `otbn_boot_services` library, in the ROM/ROM_EXT boot flow.
This will replace the existing RSA secure boot implementation  and trigger the OTBN boot services program to already be loaded by the time the ROM_EXT executes.

a-will:
@moidx Fixed by https://github.com/lowRISC/opentitan/pull/22744 perhaps?

moidx:
Yes. This is correct.

Issue #20024, opened 2023-10-13
[rom_ext] Add instruction count checks for OTBN operations.


jadephilipoom:
### Description

See https://github.com/lowRISC/opentitan/pull/20011#discussion_r1357822273

OTBN's instruction count register is a useful protection against fault attacks, especially for constant-time code. In the OTBN boot-services code, we should check the instruction counts against expected values. This may require expanding `hw/ip/otbn/util/get_instruction_count_range.py` to accept certain fixed-constant registers, so we can set the mode. For ECDSA keypair generation and signatures, the instruction count should never vary. For RSA signature verification, it will need to be a range.

This can follow the model of `mod_exp_otbn` from ROM: https://github.com/lowRISC/opentitan/blob/619ce83526bd38018d6b5c2972d07fa6e1dcef65/sw/device/silicon_creator/lib/sigverify/mod_exp_otbn.c#L91

moidx:
Hi @jadephilipoom, are you comfortable if we introduce this to the ROM_EXT first to get some test mileage before adding it to the ROM? If you agree, please move to the ROM_EXT milestone. Thanks.

jadephilipoom:
Yes, I think that's prudent. In case there's a mistake in the min/max calculation, we don't want to crash.

siemen11:
Please see https://github.com/lowRISC/opentitan/pull/30111

Issue #20023, opened 2023-10-13
[rom_ext] Test sideload-clear timing for keymgr.


jadephilipoom:
### Description

See https://github.com/lowRISC/opentitan/pull/20011#discussion_r1357813813

Currently we spin for 100 microseconds before stopping keymgr clearing; we should thoroughly test to make sure this is enough, and possibly shorten it to save time if it's more than enough. An alternative would be to leave the clearing on until the next time we need to sideload something, but that could consume a lot of entropy unnecessarily.

PR #21824, opened 2024-03-04
[boot_services] Switch boot services library to use ECDSA verify.

jadephilipoom:
This required making the P256 code slightly smaller so that sign and verify would fit together on the same binary. Luckily I knew where to find a very slightly faster exponentiation chain for P-256 modular inversion, which I had tried implementing before but hadn't gone to the effort of merging because the speed improvement was so modest. However, the code size improvement is pretty good, so that plus a few more adjustments helped me not only fit everything on the same binary but also shave a cool ~3700 cycles off our top-level ECDSA sign operation. (That's a 0.5% speed improvement, so not all that exciting, but every little bit helps!)

The first three commits are P-256 code size optimizations; the last two are the ones that actually swap over the boot services library to ECDSA.

Should help with https://github.com/lowRISC/opentitan/issues/21548 -- the ECDSA verify op exposes a pretty similar interface to the RSA sigverify's `mod_exp`, where the caller is responsible for the final comparison.

### DIFF
--- a/sw/device/silicon_creator/lib/attestation.h
+++ b/sw/device/silicon_creator/lib/attestation.h
@@ -36,9 +36,22 @@ enum {
   kAttestationPublicKeyCoordWords =
       kAttestationPublicKeyCoordBytes / sizeof(uint32_t),
   /**
+   * Size of an attestation signature component in bits.
+   */
+  kAttestationSignatureComponentBits = 256,
+  /**
+   * Size of an attestation signature component in bytes.
+   */
+  kAttestationSignatureComponentBytes = kAttestationSignatureComponentBits / 8,
+  /**
+   * Size of an attestation signature component in 32b words.
+   */
+  kAttestationSignatureComponentWords =
+      kAttestationSignatureComponentBytes / sizeof(uint32_t),
+  /**
    * Size of an attestation signature in bits.
    */
-  kAttestationSignatureBits = 512,
+  kAttestationSignatureBits = kAttestationSignatureComponentBits * 2,
   /**
    * Size of an attestation signature in bytes.
    */
@@ -82,8 +95,8 @@ typedef struct attestation_public_key {
  * Holds an attestation signature (ECDSA-P256).
  */
 typedef struct attestation_signature {
-  uint32_t r[kAttestationSignatureWords / 2];
-  uint32_t s[kAttestationSignatureWords / 2];
+  uint32_t r[kAttestationSignatureComponentWords];
+  uint32_t s[kAttestationSignatureComponentWords];
 } attestation_signature_t;
 
 #ifdef __cplusplus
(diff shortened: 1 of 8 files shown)


moidx (approved):
🎉

timothytrippel (approved):
Thanks @jadephilipoom !

jadephilipoom:
Thanks for linking the issues, Tim! But just to be clear, to completely switch over `sigverify` I think there are still a few tasks remaining:

1. We need to implement an `ecdsa_verify.c` that calls `otbn_boot_sigverify` and then does the final comparison in a FI-resistant way. For RSA this comparison is here:
https://github.com/lowRISC/opentitan/blob/ffab8da7381c66b3d2e2b78a382a9f6937e5482e/sw/device/silicon_creator/lib/sigverify/rsa_verify.c#L128-L140
For SPX+ we do something similar here:
https://github.com/lowRISC/opentitan/blob/ffab8da7381c66b3d2e2b78a382a9f6937e5482e/sw/device/silicon_creator/lib/sigverify/spx_verify.c#L85-L97
In both cases it's a matter of constructing some shares such that all the shares XORed together are equal to the "success" value, and then iterating through the two values we're comparing and XORing them with each other as well as the shares so that we only get the success value if everything is equal.
2. We need to update the ROM code to call `ecdsa_verify` instead of `rsa_verify`, including updating the test keys etc.
3. It would be good to shore up the side-channel and FI protections a bit by checking the OTBN instruction count register like we do for RSA (there are TODOs about this in the boot services lib atm). Also might be a good idea to call `verify` after `sign` when we generate attestation endorsements. For that one I'm not certain where to put it -- we could put this in the attestation lib, which would provide more control over performance since it's a costly countermeasure, but it might be better to do it on OTBN immediately when we compute the signature so that if it's bad, we don't expose the bad signature where an attacker could see it and potentially glean information from the faulty structure. What do you think @timothytrippel ?

timothytrippel:
Thanks for clarifying @jadephilipoom! Regarding:
1. I think we are tracking that in #21550 (though feel free to update the issue to be more explicit),
2. tracking in #21550 and #21548 (again feel free to make issues more explicit, sorry they are a little brief), and
3. perhaps we could start by putting the verify after sign (for cert endorsement) directly on OTBN, then we can measure its performance and make optimizations if needed? wdyt? Also, regarding the TODOs in otbn_boot_services and this sign then verify cleanup, could we file another issue to track this and tag the ROM label (so it shows up in the ROM tracking [board](https://github.com/orgs/lowRISC/projects/13/views/65?filterQuery=label%3A%22SW%3AROM%22+-Status%3ADone&groupedBy%5BcolumnId%5D=2972208&visibleFields=%5B%22Title%22%2C2972208%2C7206618%2C%22Assignees%22%2C%22Status%22%2C%22Milestone%22%5D)?

jadephilipoom:
Okay, created https://github.com/lowRISC/opentitan/issues/21943 to track the verify-after-sign.

For (1) and (2), thanks, I didn't catch those issues before! I think your comment linked them to this discussion so that's probably enough detail added.

For the TODOs in OTBN boot services, there was already a tracking issue #20024 (the TODOs are mistakenly linked to #20023 in the code, my bad :woman_facepalming: ). I added the `SW:ROM` label to #20024 and #20023 (which happens to also be ROM-relevant) as well, so both of them now show up in the dashboard.

timothytrippel:
Awesome, thanks! I added the "opentitan" project to all of the issues, so with that + the "SW:ROM" label I confirmed they show up in the dashboard now.

Issue #21823, opened 2024-03-04
[bazel] bazelisk.sh complains of unbound variable error when invoked with dvsim.py to run TLT


gdessouky:
### Description

`dvsim.py` invokes bazelisk.sh when running top-level-tests. When exercising a TLT, e.g. `/util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_hmac_smoketest`, (or bazelisk.sh alone) this errors out with 

**./bazelisk.sh: line 172: pre_cmd_args[@]: unbound variable** 

I suspect this is a side effect of this last commit https://github.com/lowRISC/opentitan/commit/4d88b9516c7f0774afed0012e50d8082c2715833, since once I undo it, the test run proceeds further.

jwnrt:
Thanks for flagging, I didn't expect these changes to break in an older Bash versions.

I could reproduce this in Bash 4.2 but I'm not sure why. I've opened https://github.com/lowRISC/opentitan/pull/21826 as a workaround but it doesn't fix the actual issue. Would be great to get some help from somebody who knows a bit more about Bash, but I'm not sure who that could be.

PR #21826, opened 2024-03-04
[bazel] Disable `set -u` flag in bazelisk.sh

jwnrt:
For some reason Bash 4.2 complains about `pre_cmd_args` being unset when as far as I can tell it is set. Disabling `set -u` as a workaround.

This isn't a problem with Bash 5+.

### DIFF
--- a/bazelisk.sh
+++ b/bazelisk.sh
@@ -10,7 +10,7 @@
 # CI jobs should use ci/bazelisk.sh instead, which performs CI-friendly additional
 # setup.
 
-set -euo pipefail
+set -eo pipefail
 
 # Change to this script's directory, as it is the location of the bazel workspace.
 cd "$(dirname "$0")"


jwnrt:
See https://github.com/lowRISC/opentitan/issues/21823 for the error being worked around. I'm not enough of a Bash expert to know exactly what's going on here in version 4.2 and it's harder than expected to find a clear list of changes between 4 and 5 so I haven't tracked down exactly what's happened here.

jwnrt:
Thanks to @a-will for finding the problem: https://git.savannah.gnu.org/cgit/bash.git/tree/CHANGES?id=3ba697465bc74fab513a26dea700cc82e9f4724e#n878

Looks like the array expansion works in both 4.2 and 5.0 but the error produced by `set -u` was fixed in 4.4. We could replace the array expansions with `"${pre_cmd_args[@]+"${pre_cmd_args[@]}"}"` to make `set -u` continue working but I personally don't think that complexity is worth it.

pamaury (approved):
Not a bash expert but this seems to be the simplest workaround.

PR #21827, opened 2024-03-04
[i2c,rtl] Add user toggle TXRST_ON_COND

### LINKED ISSUE #17750: [hw,i2c] TXDATA FIFO reset with RSTART/STOP in data byte transmission
### Description

With the update to `i2c_target_hrst_vseq` in https://github.com/lowRISC/opentitan/pull/17746, sequence introduces RStart or Stop condition during active I2C transaction.

In Target mode, during Read data byte transmission, if `RStart` condition is issued as shown in figure below, design is transitioning internal FSM state to `AcquireStart` as expected.

Since this is a glitch and the data in TXDATA fifo is no longer relevant for the request starting with `RStart`, should reset to TXDATA fifo happen via 
- write to FIFO_CTL register ?
- or DUT resets the FIFO internally ? 

image

hcallahan-lowrisc:
By setting the bit CSR.TARGET_FIFO_CONFIG.TXRST_ON_COND, the block will then automatically reset the TXFIFO (using TXRST) if it observes a STOP/RSTART condition in Target-Mode.

From the commit message:
> This behaviour may be useful to software, as any remaining data in the TXFIFO
> after a Sr/P condition is probably no-longer applicable to the next transfer, so
> will likely have to be cleared out anyway. Keeping this behaviour as a toggle
> allows software to observe the fifo state before resetting it, which may be
> useful to understand how much of the previous transfer completed if that
> information is helpful or relevant.

Reset state is disabled.

Closes #17750

### DIFF
--- a/hw/ip/i2c/rtl/i2c_fsm.sv
+++ b/hw/ip/i2c/rtl/i2c_fsm.sv
@@ -73,6 +73,7 @@ module i2c_fsm import i2c_pkg::*;
   input logic [6:0] target_address1_i,
   input logic [6:0] target_mask1_i,
 
+  output logic target_sr_p_cond_o,       // Saw RSTART/STOP in Target-Mode.
   output logic event_target_nack_o,      // this target sent a NACK (this is used to keep count)
   output logic event_nak_o,              // target didn't Ack when expected
   output logic event_scl_interference_o, // other device forcing SCL low
@@ -1591,6 +1592,11 @@ module i2c_fsm import i2c_pkg::*;
       state_d = Idle;
     end
   end
+
+  // TARGET-Mode -> RSTART or STOP condition observed
+  // This output is used to (optionally) reset the TXFIFO, as any data populated by
+  // software is now highly-unlikely to be applicable to the next transaction.
+  assign target_sr_p_cond_o = target_enable_i && !target_idle && (stop_det | start_det);
 
   // Synchronous state transition
   always_ff @ (posedge clk_i or negedge rst_ni) begin : state_transition
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -388,6 +388,14 @@
                 '''
           resval: "0"
         }
+        { bits: "15"
+          name: "TXRST_ON_COND"
+          desc: '''If set, automatically reset the TX FIFO (TXRST) upon seeing a RSTART/STOP condition
+                during an active transaction in Target Mode. This may be useful if the remaining data
+                in the TX FIFO becomes no longer applicable to the next transaction.
+                '''
+          resval: "0"
+        }
         { bits: "27:16"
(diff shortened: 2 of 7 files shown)


andreaskurth:
@hcallahan-lowrisc and @marnovandermaas: What is the status and plan for this PR and issue #17750? Does that issue need a HW fix, and if so, is this the proposed fix?

marnovandermaas:
So it is possible to delegate this fix to software since on a restart or stop an interrupt happens. This means that software must manually empty the transmit FIFO. I think we discussed this issue and decided that it would be good to make this an automatic option, maybe @hcallahan-lowrisc has more context?

hcallahan-lowrisc:
Yes so I linked this PR to close #17750, but as discussed in that issue, there is a software workaround, though I understand that it remained open so a potential hardware fix could be implemented when the time was right. The hardware fix was proposed in a comment [in that issue](https://github.com/lowRISC/opentitan/issues/17750#issuecomment-1488815735), which is what this PR implements.
It should still be considered lower-priority, but was triaged into M2.

I still want to move the control bit into a different CSR, but otherwise this PR is complete.
It will require some DV effort for D2.

andreaskurth (approved):
Thanks @hcallahan-lowrisc, this LGTM, though it would be valuable to get feedback from the other reviewers too

andreaskurth:
CI checks show some real failures that need to be addressed (probably low effort)

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -1474,6 +1475,9 @@ module i2c_fsm import i2c_pkg::*;
     end
   end
 
+  // TARGET-Mode -> unexpected RSTART or STOP
+  assign unexp_cond_o = (bit_idx != 0) && (stop_det | start_det);
```
a-will, line 1479: Note that this doesn't actually address what I meant in the issue. The generalized form is a case where there is data still left in the TX FIFO at the end of a data transfer, and the i2c IP doesn't know what to do with it.

You could probably drop the term for the bit_idx and be handling that more general case.

hcallahan-lowrisc: Okay I understand what you mean now. I think I just read the title of that issue a bit too literally.
I updated the logic to trigger TXRST for the generic case now.

andreaskurth:
@a-will: WDYT after the latest update?

a-will (approved):

hw/ip/i2c/rtl/i2c_fsm.sv
a-will: Thanks for taking care of that!

andreaskurth:
The _CW310 ROM Tests_ CI check fails with `Error: Found no USB device`. This also happens for other PRs and is most likely unrelated to this PR. All other CI checks pass --> merging

PR #21829, opened 2024-03-04
[adc_ctrl] Updates to LP -> NP transition

msfschaffner:
This contains the following patches:
1) the adc_ctrl FSM has one more arc now that allows it to go back to LP mode iff LP mode was enabled in the first place, and a match could not be confirmed after transitioning from LP -> NP mode.
2) an FSM state observation register is added. his is intended for debug purposes only.
3) another internal interrupt and wakeup source is added that allows to wake up or interrupt the system if an LP -> NP transition has occurred. this is intended for debug purposes only.


jesultra:
From the description, it sounds exactly as what we need.

hw/ip/adc_ctrl/data/adc_ctrl.hjson
```diff
@@ -312,24 +312,6 @@
       ]
     }
 
-    { name: "filter_status",
```
msfschaffner, line 315: @moidx @timothytrippel PTAL

msfschaffner:
The DIF has been aligned, but the DV environment still needs patches and fails currently.

hw/ip/adc_ctrl/rtl/adc_ctrl_intr.sv
```diff
@@ -22,83 +31,115 @@ module adc_ctrl_intr import adc_ctrl_reg_pkg::*; (
...
+     if (!rst_aon_ni) begin
+        aon_req_hold_q <= '0;
+     end else if (aon_ld_req) begin
+        aon_req_hold_q <= staging_reqs_q;
+     end else if (aon_ack) begin
+        aon_req_hold_q <= '0;
+     end
+  end
+
+  // delay register to hold the wakeup request until the bus-side status has been re-synced.
```
a-will, line 83: Shall we add a little more explanation so the reader knows we've done our homework?

```suggestion
  // delay register to hold the wakeup request until the bus-side status has been re-synced.
  // This covers the uncertainty of the relative timing for CDC between these two events:
  //   - dst_ack -> (2FF sync) -> aon_ack
  //   - adc_intr_status_i -> (2FF sync) -> aon_irq_status
```

That is what this delay register is about, right?

hw/ip/adc_ctrl/rtl/adc_ctrl_fsm.sv
```diff
@@ -347,7 +351,11 @@ module adc_ctrl_fsm
           // 3, the sample count is greater than the threshold -> this is a continued stable match,
```
mvertescher, line 351: Just to confirm, if we're sampling a continued stable filter match like a [debug cable filter or the disconnected filter](https://opentitan.org/book/hw/ip/adc_ctrl/doc/programmers_guide.html#use-case), we're always in NP mode? Do we not want to track stable filter matches in LP mode?

hw/ip/adc_ctrl/rtl/adc_ctrl_intr.sv
msfschaffner: yes that is correct! I will add this.

hw/ip/adc_ctrl/data/adc_ctrl.hjson
```diff
@@ -307,25 +307,28 @@
       fields: [
         { bits: "NumAdcFilter-1:0",
           name: "EN",
-          desc: "0: filter match wil not generate wakeupe; 1: filter match will generate wakeup",
+          desc: '''
+          Filter wakeup source.
+
+          0: filter match wil not generate wakeup;
```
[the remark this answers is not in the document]
msfschaffner: fixed, thanks

hw/ip/adc_ctrl/rtl/adc_ctrl_fsm.sv
msfschaffner: That is correct.

I fear this might require a more significant rewrite of the FSM and control logic, which I was trying to avoid due to DV impact. The fix I currently made is very localized and addresses the main problem (which is to avoid high power consumption while the chip is in either of the sleep modes).

Handling continued filter match with NP sampling should not be that problematic I think, since presumably the entire system will be pulled out of sleep anyway when that happens.

sw/device/lib/dif/dif_adc_ctrl.h
```diff
@@ -73,7 +73,7 @@ typedef enum dif_adc_ctrl_channel {
  * instances that may be configured.
  */
 typedef enum dif_adc_ctrl_filter {
-  DIF_ADC_CTRL_FILTER_LIST(DIF_ADC_CTRL_FILTER_ENUM_INIT_)
+  DIF_ADC_CTRL_FILTER_LIST(DIF_ADC_CTRL_FILTER_ENUM_INIT_) kDifAdcCtrlTrans,
```
timothytrippel, line 76: Nit: maybe add a new line between the x-macro instantiation and this additional item in the enum?

timothytrippel (approved):
DIF changes LGTM.

sw/device/lib/dif/dif_adc_ctrl.h
msfschaffner: ok will do!

msfschaffner: ah that's a clang formatter thing. it only accepts this when leaving a blank line in between (which I think is still better for readability)

msfschaffner:
Ok the latest force push updates the block-level DV and models the new behavior so that it is implicitly tested by randomized sequences like `adc_ctrl_filters_wakeup`. The FSM state output register for debug is just tested with an SVA right now.

mvertescher (commented):
LGTM, thanks for the ADC debug changes!

moidx (approved):

PR #21830, opened 2024-03-04
[silicon] Add proda signing and execution environment

moidx:
* Add signing token configuration
* Add Bazel execution environment 
* Update relevant device tests to support the proda execution environment

### DIFF
--- a/signing/README.md
+++ b/signing/README.md
@@ -88,6 +88,32 @@ example:
 }
 ```
 
+## Configuration for silicon_owner PRODA SKU signing
+
+The production PRODA SKU expects a signing key held in Google's cloud-kms
+service.
+
+See the previous section for more details on how to update the `hsmtool`
+configuration.
+
+```json
+{
+  "earlgrey_a0": {
+    "token": "earlgrey_a0_000",
+    "user": "user",
+    "pin": "XXXXXX"
+  },
+  "earlgrey_z0_sival": {
+    "token": "ot-earlgrey-z0-sival",
+    "user": "user"
+  },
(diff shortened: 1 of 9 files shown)


signing/README.md
```diff
@@ -88,6 +88,32 @@ example:
...
+{
+  "earlgrey_a0": {
+    "token": "earlgrey_a0_000",
+    "user": "user",
+    "pin": "XXXXXX"
+  },
+  "earlgrey_z0_sival": {
+    "token": "ot-earlgrey-z0-sival",
+    "user": "user"
+  },
```
cfrantz, line 109: These aren't really necessary for the `proda` setup (although they won't hurt either).

sw/device/tests/BUILD
```diff
@@ -425,6 +425,7 @@ opentitan_test(
         "//hw/top_earlgrey:silicon_creator": None,
         "//hw/top_earlgrey:silicon_owner_sival_rom_ext": "silicon_owner",
         "//hw/top_earlgrey:silicon_owner_prodc_rom_ext": "silicon_owner",
+        "//hw/top_earlgrey:silicon_owner_proda_rom_ext": "silicon_owner",
```
cfrantz, line 428: Do we really want to add `proda` to all the tests?

cfrantz (approved):

sw/device/tests/BUILD
moidx: In the past it has been useful to allow the silicon creator to test the sku before delivery. I think this could be restricted to just smoketests.

PR #21831, opened 2024-03-05
[rom] remove 64k ROM_EXT code size restriction

### LINKED ISSUE #21553: [rom] remove 64k code size restriction on first boot stage
This restriction has made writing provisioning firmware challeging.
Once this is done, the FT personalization firmware can be consolidated into a single binary.

timothytrippel:
This fixes #21553.

Note, this simply removes the restriction. Additional changes will be required to truly support dynamic ROM_EXT / Owner slot sizes (if that is deemed desirable in the future).

### DIFF
--- a/sw/host/opentitanlib/src/image/image.rs
+++ b/sw/host/opentitanlib/src/image/image.rs
@@ -15,7 +15,7 @@ use thiserror::Error;
 use crate::crypto::rsa::Modulus;
 use crate::crypto::rsa::Signature as RsaSignature;
 use crate::crypto::sha256;
-use crate::image::manifest::Manifest;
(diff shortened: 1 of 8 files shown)


cfrantz (approved):

sw/host/opentitanlib/src/image/image.rs
```diff
@@ -109,6 +109,13 @@ impl Image {
         ensure!(manifest.entry_point < len);
         ensure!(manifest.extensions.entries.iter().all(|x| x.offset < len));
 
+        // We allow
```
pamaury, line 112: unfinished sentence?

moidx (approved):

sw/host/opentitanlib/src/image/image.rs
timothytrippel: ah thanks. Good catch.

PR #21833, opened 2024-03-05
[rom] Add ECDSA P256 support to manifest.h

moidx:
This commit adds support for ECDSA public key and signatures to the manifest. Both signature and public key fields were converted to union types to be able to support either RSA or ECDSA.

`kManifestVersionMajor1` is associated with RSA.
`kManifestVersionMajor2` is associated with ECDSA P256.

### DIFF
--- a/sw/device/lib/testing/test_rom/BUILD
+++ b/sw/device/lib/testing/test_rom/BUILD
@@ -122,6 +122,7 @@ cc_library(
         "//sw/device/silicon_creator/lib:keymgr_binding",
         "//sw/device/silicon_creator/lib/base:chip",
         "//sw/device/silicon_creator/lib/base:sec_mmio",
+        "//sw/device/silicon_creator/lib/sigverify:ecdsa_p256_key",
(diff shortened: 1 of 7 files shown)


sw/device/silicon_creator/lib/manifest_def.c
```diff
@@ -28,7 +28,7 @@ static manifest_t kManifest_ = {
     .manifest_version =
         (manifest_version_t){
             .minor = kManifestVersionMinor1,
-            .major = kManifestVersionMajor1,
+            .major = kManifestVersionMajor2,
```
cfrantz, line 31: I'm not sure we should advance the version number just yet -- even though we've defined what the manifest structure is with ECDSA keys, we aren't actually using them yet.

moidx: Agreed. I reverted this change. Will update once we enable ECDSA verify.

timothytrippel (approved):

PR #21837, opened 2024-03-05
[usbdev,dv] usbdev_dv_env_changes

MubashirSaleem775:
Adds changes in DV env to support error cases like timeout, PID error, CrC error and bitstuff error.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_agent_cfg.sv
+++ b/hw/dv/sv/usb20_agent/usb20_agent_cfg.sv
@@ -13,4 +13,13 @@ class usb20_agent_cfg extends dv_base_agent_cfg;
 
   `uvm_object_new
 
+  // If this is set then the driver will generate a bit stuffing error.
+  bit usb_bitstuff_error = 1'b0;
+  // This flag serves to evaluate the device's functionality,
+  // particularly regarding the recognition of a single SE0 bit as an end-of-packet,
+  // which necessitates two successive bits otherwise. Once set, it is accompanied by
+  // its respective test sequence, followed by the transmission of a single-bit SE0 as EOP.
+  bit single_bit_SE0 = 1'b0;
+  
+  bit time_out = 1'b0;
 endclass
--- a/hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
+++ b/hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
@@ -17,28 +17,51 @@ package usb20_agent_pkg;
   // Packet IDentifiers; the 4 LSBs identify the Packet type, and are reflected in inverted
   //   form as the 4 MSBs to provide some internal error checking.
   typedef enum bit [7:0] {
-      // Token PIDs
-      PidTypeOutToken = 8'b1110_0001, PidTypeInToken    = 8'b0110_1001,
-      PidTypeSofToken = 8'b1010_0101, PidTypeSetupToken = 8'b0010_1101,
-      // Data PIDs
-      PidTypeData0 = 8'b1100_0011, PidTypeData1 = 8'b0100_1011,
-      PidTypeData2 = 8'b1000_0111, PidTypeMData = 8'b0000_1111,
-      // Handshake PIDs
-      PidTypeAck   = 8'b1101_0010, PidTypeNak  = 8'b0101_1010,
-      PidTypeStall = 8'b0001_1110, PidTypeNyet = 8'b1001_0110,
-      // Special PIDs
-      PidTypePre   = 8'b0011_1100, PidTypeSplit = 8'b0111_1000,
-      PidTypePing  = 8'b1011_0100} pid_type_e;
+    // Token PIDs
+    PidTypeOutToken   = 8'b1110_0001,
+    PidTypeInToken    = 8'b0110_1001,
+    PidTypeSofToken   = 8'b1010_0101,
+    PidTypeSetupToken = 8'b0010_1101,
+    // Data PIDs
+    PidTypeData0 = 8'b1100_0011,
+    PidTypeData1 = 8'b0100_1011,
+    PidTypeData2 = 8'b1000_0111,
+    PidTypeMData = 8'b0000_1111,
+    // Handshake PIDs
+    PidTypeAck   = 8'b1101_0010,
+    PidTypeNak   = 8'b0101_1010,
+    PidTypeStall = 8'b0001_1110,
+    PidTypeNyet  = 8'b1001_0110,
+    // Special PIDs
+    PidTypePre   = 8'b0011_1100,
+    PidTypeSplit = 8'b0111_1000,
+    PidTypePing  = 8'b1011_0100,
+    PidTypeTimeOut = 8'b1100_1100
+  } pid_type_e;
 
-  typedef enum byte {bmRequestType0 = 8'b0_00_00000, bmRequestType1 = 8'b0_00_00001,
-      bmRequestType2 = 8'b0_00_00010, bmRequestType3 = 8'b1_00_00000,
-      bmRequestType4 = 8'b1_00_00001, bmRequestType5 = 8'b1_00_00010} bmrequesttype_e;
+  typedef enum byte {
+    bmRequestType0  = 8'b0_00_00000,
+    bmRequestType1  = 8'b0_00_00001,
+    bmRequestType2  = 8'b0_00_00010,
+    bmRequestType3  = 8'b1_00_00000,
+    bmRequestType4  = 8'b1_00_00001,
+    bmRequestType5  = 8'b1_00_00010
+  } bmrequesttype_e;
 
-  typedef enum byte {bRequestGET_STATUS = 8'h00, bRequestCLEAR_FEATURE = 8'h01,
-      bRequestSET_FEATURE = 8'h03, bRequestSET_ADDRESS = 8'h05, bRequestGET_DESCRIPTOR = 8'h06,
-      bRequestSET_DESCRIPTOR = 8'h07, bRequestGET_CONFIGURATION = 8'h08,
-      bRequestSET_CONFIGURATION = 8'h09, bRequestGET_INTERFACE = 8'h0A,
-      bRequestSET_INTERFACE = 8'h0B, bRequestSYNCH_FRAME = 8'h0C} brequest_e;
+  typedef enum byte {
+    bRequestGET_STATUS        = 8'h00,
+    bRequestCLEAR_FEATURE     = 8'h01,
+    bRequestSET_FEATURE       = 8'h03,
+    bRequestSET_ADDRESS       = 8'h05,
+    bRequestGET_DESCRIPTOR    = 8'h06,
+    bRequestSET_DESCRIPTOR    = 8'h07,
+    bRequestGET_CONFIGURATION = 8'h08,
+    bRequestSET_CONFIGURATION = 8'h09,
+    bRequestGET_INTERFACE     = 8'h0A,
+    bRequestSET_INTERFACE     = 8'h0B,
+    bRequestSYNCH_FRAME       = 8'h0C
+  } brequest_e;
+
   typedef enum bit [2:0] {CtrlTrans, IsoTrans, IntrptTrans, BulkTrans} usb_transfer_e;
   // local types
   // forward declare classes to allow typedefs below
--- a/hw/dv/sv/usb20_agent/usb20_driver.sv
+++ b/hw/dv/sv/usb20_agent/usb20_driver.sv
@@ -75,7 +75,7 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     `uvm_info(`gfn, $sformatf("Complete Token_Packet = %p", comp_token_pkt), UVM_DEBUG)
     drive_pkt(comp_token_pkt);
     if (req_item.m_pid_type == PidTypeInToken) begin
-      get_dut_response(rsp_item);
+      device_response(rsp_item);
       seq_item_port.item_done(rsp_item);
       `uvm_info (`gfn, $sformatf("In drive afer In packet : \n %0s", rsp_item.sprint()), UVM_DEBUG)
     end else begin
@@ -107,13 +107,10 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     end
     `uvm_info(`gfn, $sformatf("Complete Data_Packet = %p", comp_data_pkt), UVM_DEBUG)
     drive_pkt(comp_data_pkt);
-    `uvm_info(`gfn, $sformatf("\n\nTransfer Type = %s", req_item.m_usb_transfer), UVM_HIGH)
     if (req_item.m_usb_transfer == IsoTrans) begin
       seq_item_port.item_done();
-      `uvm_info(`gfn, $sformatf("\n\nTransfer Type = %s", req_item.m_usb_transfer), UVM_HIGH)
     end else begin
-      `uvm_info(`gfn, $sformatf("\n\nTransfer Type = %s", req_item.m_usb_transfer), UVM_HIGH)
-      get_dut_response(rsp_item);
+      device_response(rsp_item);
       seq_item_port.item_done(rsp_item);
     end
   endtask
@@ -195,8 +192,12 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   // EOP Task
   // -------------------------------
   task end_of_packet();
-    for (int j = 0; j < 2; j++) begin
-      drive_bit_interval(EOP[j], EOP[j]);
+    if (~cfg.single_bit_SE0) begin
+      for (int j = 0; j < 2; j++) begin
+        drive_bit_interval(EOP[j], EOP[j]);
+      end
+    end else begin
+       drive_bit_interval(EOP[0], EOP[0]);
     end
     `uvm_info(`gfn, "\n After EOP Idle state", UVM_DEBUG)
     drive_bit_interval(1'b1, 1'b0);
@@ -207,17 +208,20 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   task bit_stuffing(input bit packet[], output bit bit_stuff_out[]);
     int consecutive_ones_count = 0;
     bit stuffed[$];
-    for (int i = 0; i < packet.size(); i++) begin
-      stuffed.push_back(packet[i]);
-      if (packet[i] == 1'b1) begin
-        consecutive_ones_count = consecutive_ones_count + 1;
-        if (consecutive_ones_count >= 6) begin
-          consecutive_ones_count = 0;
-          stuffed.push_back(1'b0);
-        end
-      end else consecutive_ones_count = 0;
+   if (~cfg.usb_bitstuff_error) begin
+     for (int i = 0; i < packet.size(); i++) begin
+       stuffed.push_back(packet[i]);
+       if (packet[i] == 1'b1) begin
+         consecutive_ones_count = consecutive_ones_count + 1;
+         if (consecutive_ones_count >= 6) begin
+           consecutive_ones_count = 0;
+           stuffed.push_back(1'b0);
+         end
+       end else consecutive_ones_count = 0;
     end
     bit_stuff_out = stuffed;
+   end else
+     bit_stuff_out = packet;
   endtask
 
   // Returns 1 in the event of detecting a bit stuffing error, output is
@@ -282,8 +286,8 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   // -------------------------------
   virtual task reset_signals();
     // Bus is unpowered and inactive.
-    cfg.bif.drive_vbus = 1'b0;
-    cfg.bif.usb_rx_d_i = 1'b0;
+    cfg.bif.drive_vbus = 1'b1;
+    cfg.bif.usb_rx_d_i = 1'b1;
     cfg.bif.drive_p  = 1'b0;
     cfg.bif.drive_n = 1'b0;
     @(posedge cfg.bif.rst_ni);
@@ -317,7 +321,37 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
 
   // Get_DUT_Response
   // -------------------------------
-  task get_dut_response(ref usb20_item rsp_item);
+  // This task verifies the device status following the transmission of an OUT packet or IN token.
+  // Upon receiving the IN token/OUT packet, the device is expected to initiate a response.
+  // This task monitors whether the device initiates the response in the form of a
+  // handshake or data packet within the specified timeout period.
+  // If no response is detected within timeout frame that is 18 bit times(from section 7.1.19.1),
+  // it send timeout response to sequence.
+  task device_response(ref usb20_item rsp_item);
+    bit timed_out = 1'b0;
+    `uvm_info(`gfn, "After drive Packet in wait to check usb_dp_en_o signal", UVM_LOW)
+    fork begin : isolation_fork
+      fork
+        begin
+          repeat (18) begin
+            repeat (4) @(posedge cfg.bif.clk_i);
+            if (cfg.bif.usb_dp_en_o) break;
+          end
+          if (!cfg.bif.usb_dp_en_o) begin
+            timed_out = 1'b1;
+            disable get_device_response;
+          end
+        end
+        get_device_response(rsp_item);
+      join
+      if (timed_out) begin
+        // this bit will indicate that device didn't repond within timeout period.
+        cfg.time_out = 1'b1; 
+      end
+    end join
+  endtask
+
+  task get_device_response(ref usb20_item rsp_item);
     bit received_pkt[];
     bit nrzi_out_pkt[];
     bit decoded_received_pkt[];
--- a/hw/ip/usbdev/dv/env/usbdev_env.sv
+++ b/hw/ip/usbdev/dv/env/usbdev_env.sv
@@ -18,7 +18,7 @@ class usbdev_env extends cip_base_env #(
     super.build_phase(phase);
     // get clock/resets
     if (!uvm_config_db#(virtual clk_rst_if)::get(this, "", "aon_clk_rst_vif",
-        cfg.aon_clk_rst_vif)) begin
+                                                 cfg.aon_clk_rst_vif)) begin
       `uvm_fatal(get_full_name(), "failed to get aon_clk_rst_if from uvm_config_db")
     end
     if (!uvm_config_db#(virtual clk_rst_if)::get(this, "", "host_clk_rst_vif",
@@ -28,6 +28,10 @@ class usbdev_env extends cip_base_env #(
     // get usb20_if handle
     if (!uvm_config_db#(virtual usb20_if)::get(this, "", "vif", cfg.usb20_usbdpi_vif)) begin
       `uvm_fatal(`gfn, "failed to get usb20_if handle from uvm_config_db")
+    end
+    // get usb20_block_if handle
+    if (!uvm_config_db#(virtual usb20_block_if)::get(this, "", "bif", cfg.bif)) begin
+      `uvm_fatal(`gfn, "failed to get usb20_block_if handle from uvm_config_db")
     end
 
     // Use the configured USB speed for the main clock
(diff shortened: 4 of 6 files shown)


hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -296,22 +318,43 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+    // Wait for 18 bit times for timeout.
+    for ( int i = 0; i < 18; i++) begin
+      @(posedge cfg.bif.usb_clk)
+      if (cfg.bif.usb_dp_en_o) begin
+        disable time_out;
+      end
+    end
+    rsp_item.m_pid_type = pid_type_e'(PidTypeTimeOut);
+    disable get_response_pid;
+  endtask
```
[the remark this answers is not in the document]
rswarbrick: Looking a bit more carefully at your code, I think the `time_out` task can complete without killing `get_response_pid`. Would this be correct?
```systemverilog
    bit timed_out = 1'b0;

    fork begin : isolation_fork
      fork
        begin
          repeat (18) begin
            @(posedge cfg.bif.usb_clk);
            if (cfg.bif.usb_dp_en_o) break;
          end
          if (!cfg.bif.usb_dp_en_o) timed_out = 1'b1;
        end
        get_response_pid(rsp_item);
      join_any
      disable fork;

      if (timed_out) begin
        rsp_item.m_pid_type = pid_type_e'(PidTypeTimeOut);
      end
    end join
```

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -296,22 +318,43 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
-    `uvm_info(`gfn, $sformatf("Received Packet = %p", received_pkt), UVM_LOW)
-    nrzi_decoder (received_pkt, decoded_received_pkt);
-    `uvm_info(`gfn, $sformatf("Decoded Received Packet = %p", decoded_received_pkt), UVM_LOW)
-    for (int i = 0 ; i < 8; i++) begin
+    `uvm_info(`gfn, $sformatf("Received Packet = %p", received_pkt), UVM_DEBUG)
+    nrzi_decoder(received_pkt, decoded_received_pkt);
+    `uvm_info(`gfn, $sformatf("Decoded Received Packet = %p", decoded_received_pkt), UVM_DEBUG)
+    for (int i = 0; i < 8; i++) begin
       received_pid[i] = decoded_received_pkt[i + 8];
     end
```
rswarbrick, line 359: Is this just the same as:
```systemverilog
received_pid = decoded_received_pid[15:8];
```

If so, maybe that would be simpler?

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -296,22 +318,43 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
-    `uvm_info(`gfn, "After drive Packet in wait to check usb_dp_en_o signal", UVM_DEBUG)
-    wait(cfg.bif.usb_dp_en_o);
-    while (cfg.bif.usb_dp_en_o) begin
-      received_pkt = new[received_pkt.size() + 1](received_pkt);
-      @(posedge cfg.bif.usb_clk)
-      received_pkt[receive_index] = cfg.bif.usb_p;
-      receive_index = receive_index + 1;
+    wait (cfg.bif.usb_dp_en_o);
+    if (cfg.bif.usb_dp_en_o) begin
+      while (cfg.bif.usb_dp_en_o) begin
```
rswarbrick, line 347: I think you can get rid of the `if` here: the `while` loop won't do anything if `cfg.bif.usb_dp_en_o` is already false.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -296,22 +318,43 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+        received_pkt = new[received_pkt.size() + 1](received_pkt);
```
rswarbrick, line 348: Why aren't we just using a queue for `received_pkt`? I think it would be rather simpler.

hw/ip/usbdev/dv/tb/tb.sv
```diff
@@ -150,7 +150,7 @@ module tb;
     uvm_config_db#(intr_vif)::set(null, "*.env", "intr_vif", intr_if);
     uvm_config_db#(virtual tl_if)::set(null, "*.env.m_tl_agent*", "vif", tl_if);
     uvm_config_db#(virtual usb20_if)::set(null, "*.env.m_usb20_agent*", "vif", usb20_if);
-    uvm_config_db#(virtual usb20_block_if)::set(null, "*.env.m_usb20_agent*","bif",usb20_block_if);
+    uvm_config_db#(virtual usb20_block_if)::set(null, "*.env*","bif",usb20_block_if);
```
rswarbrick, line 153: I know they were missing before, but... missing spaces between the arguments here.

hw/ip/usbdev/dv/env/usbdev_env_cfg.sv
```diff
@@ -5,6 +5,7 @@
 class usbdev_env_cfg extends cip_base_env_cfg #(.RAL_T(usbdev_reg_block));
 
   virtual clk_rst_if  aon_clk_rst_vif;
+  virtual usb20_block_if bif;
```
rswarbrick, line 8: I see that the `tb.sv` change is setting this virtual interface instead of one in the agent. Don't we also have to delete a handle from the agent class?

rswarbrick (changes requested):

hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
```diff
@@ -18,7 +18,7 @@ package usb20_agent_pkg;
       PidTypeSofToken = 8'b0101_1010, PidTypeSetupToken = 8'b1101_0010,
       PidTypeData0 = 8'b0011_1100, PidTypeData1 = 8'b1011_0100, PidTypeData2 = 8'b0111_1000,
       PidTypeMData = 8'b1111_0000, PidTypeAck = 8'b0010_1101, PidTypeNak = 8'b1010_0101,
-      PidTypeStall = 8'b1110_0001, PidTypeNyet = 8'b0110_1001} pid_type_e;
+      PidTypeStall = 8'b1110_0001, PidTypeNyet = 8'b0110_1001, PidTypeTimeOut = 8'b1100_1100} pid_type_e;
```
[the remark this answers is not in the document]
MubashirSaleem775: okay i will do .

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -180,45 +181,66 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
-      cfg.bif.drive_p =  EOP[j];
-      cfg.bif.drive_n =  EOP[j];
+      cfg.bif.drive_p <=  EOP[j];
+      cfg.bif.drive_n <=  EOP[j];
+    end
+    @(posedge cfg.bif.usb_clk) begin
+      `uvm_info(`gfn, "\n After EOP Idle state", UVM_DEBUG)
+      cfg.bif.drive_p <= 1'b1;
+      cfg.bif.drive_n <= 1'b0;
     end
```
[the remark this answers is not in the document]
MubashirSaleem775: okay, l will try this way.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -180,45 +181,66 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+    @(posedge cfg.bif.usb_clk)
+    cfg.bif.drive_p <=  1'b0;
+    cfg.bif.drive_n <=  1'b0;
     @(posedge cfg.bif.usb_clk) begin
       `uvm_info(`gfn, "\n After EOP Idle state", UVM_DEBUG)
-      cfg.bif.drive_p = 1'b1;
-      cfg.bif.drive_n = 1'b0;
+      cfg.bif.drive_p <= 1'b1;
+      cfg.bif.drive_n <= 1'b0;
     end
```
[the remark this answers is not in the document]
MubashirSaleem775: Okay

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -180,45 +181,66 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
-    for (int i = 0; i < packet.size(); i++) begin
-      if (packet[i] == 1'b1) begin
-        consecutive_ones_count = consecutive_ones_count + 1;
-        if (consecutive_ones_count == 6) begin
-          packet = new[packet.size() + 1](packet);
-          for (int j = packet.size() ; j > i; j = j - 1) begin
-            packet[j] = packet[j - 1];
+    if (cfg.usb_bitstuff_error) begin
+      bit_stuff_out = packet;
+    end else begin
```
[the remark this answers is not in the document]
MubashirSaleem775: sure

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: we can't do this because The type of the target is 'bit[7:0]', while the type of the 
  source is 'bit$[]

hw/ip/usbdev/dv/env/usbdev_env_cfg.sv
MubashirSaleem775: Yes, however, this virtual is utilized in the agent, which is why we require this handle within the agent.

hw/ip/usbdev/dv/tb/tb.sv
MubashirSaleem775: okay

hw/dv/sv/usb20_agent/usb20_driver.sv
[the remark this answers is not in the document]
MubashirSaleem775: I attempted this, but it didn't work for both scenarios, namely timeout and get_response_pid. That's why I made some modifications in the updated commit.

MubashirSaleem775:
@rswarbrick can you please log of DV_smoke failue.

rswarbrick:
As a general comment, please could you rebase over recent changes to the code? There are currently merge conflicts with some tidyups that @alees24 merged last week.

hw/dv/sv/usb20_agent/usb20_driver.sv
rswarbrick: Oh, how annoying! :-) Thanks for the explanation.

hw/dv/sv/usb20_agent/usb20_driver.sv
rswarbrick: Thanks for tidying this up. Something seems to have got a little unstuck with the indentation though. Would you mind re-aligning?

hw/ip/usbdev/dv/env/usbdev_env_cfg.sv
rswarbrick: Ahah, I see. Thanks for the explanation.

hw/dv/sv/usb20_agent/usb20_driver.sv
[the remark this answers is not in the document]
rswarbrick: Thanks for tidying this up: I think it looks much better.

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: okay

hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
```diff
@@ -14,23 +14,45 @@ package usb20_agent_pkg;
 
   // usb20_item enums
   typedef enum bit [2:0] {PktTypeSoF, PktTypeToken, PktTypeData, PktTypeHandshake} pkt_type_e;
-  // TODO: these are presently nibble-swapped relative to the specification.
```
rswarbrick, line 17: This is not mergeable yet: you are dropping a TODO comment which is still true. Please could you put it back?

hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
```diff
@@ -14,23 +14,45 @@ package usb20_agent_pkg;
...
+  typedef enum byte {
+    bmRequestType0  = 8'b0_00_00000,
+    bmRequestType1  = 8'b0_00_00001,
+    bmRequestType2  = 8'b0_00_00010,
+    bmRequestType3  = 8'b1_00_00000,
+    bmRequestType4  = 8'b1_00_00001,
+    bmRequestType5  = 8'b1_00_00010
+  } bmrequesttype_e;
+
+  typedef enum byte {bRequestGET_STATUS = 8'h00,
```
rswarbrick, line 43: Nit: Missing newline after the `{`

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -198,11 +199,28 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     end
   endtask
 
+   // Single bit SE0 EOP
+  // -------------------------------
+  task single_se0_end_of_packet();
```
rswarbrick, line 204: Isn't this equivalent to `end_of_packet` but with just one iteration around the loop? Probably better to encode this with an optional argument in the existing task?
```
task end_of_packet(bit single_se0 = 1'b0);
```

hw/dv/sv/usb20_agent/usb20_agent_cfg.sv
```diff
@@ -14,4 +14,7 @@ class usb20_agent_cfg extends dv_base_agent_cfg;
 
   `uvm_object_new
 
+  bit usb_bitstuff_error = 1'b0;
+  bit single_bit_SE0 =1'b0;
```
rswarbrick, line 18: These fields need comments that explain what they do.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -309,7 +327,30 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   // Get_DUT_Response
   // -------------------------------
   task get_dut_response(ref usb20_item rsp_item);
-    bit received_pkt[];
+   bit timed_out = 1'b0;
```
rswarbrick, line 330: Nit: indentation is off here.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -309,7 +327,30 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   // Get_DUT_Response
   // -------------------------------
   task get_dut_response(ref usb20_item rsp_item);
```
rswarbrick, line 329: This task needs a comment that explains what it does. (I know that this is true about other tasks in the same file, but we're touching this one!)

Can you describe the behaviour in a sentence or two and put it into a comment above the task?

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -309,7 +327,30 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+      fork
+        begin
+          repeat (18) begin
+            @(posedge cfg.bif.usb_clk);
+            if (cfg.bif.usb_dp_en_o) break;
+          end
+          if (!cfg.bif.usb_dp_en_o) begin
+            timed_out = 1'b1;
+            disable get_response_pid;
+          end
```
rswarbrick, line 342: Can't this be written slightly more simply with something like this? (untested)
```systemverilog
    fork begin : isolation_fork
      fork
        get_response_pid();
        begin
          repeat (18) begin
            @(posedge cfg.bif.usb_clk);
            if (cfg.bif.usb_dp_en_o) break;
          end
          timed_out = 1'b1;
        end
      join_any
      disable fork;
      if (timed_out) begin
        rsp_item.m_pid_type = pid_type_e'(PidTypeTimeOut);
      end
    end join
```

Note that we no longer have to give a process to the `disable` call, which seems a bit simpler to understand.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -335,9 +375,7 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
       else @(posedge cfg.bif.usb_clk);
       // Detect SE0 signaling which indicates End Of Packet
       if (cfg.bif.usb_p === 1'b0 && cfg.bif.usb_n === 1'b0) break;
-      received_pkt = new[received_pkt.size() + 1](received_pkt);
-      received_pkt[receive_index] = cfg.bif.usb_p;
-      receive_index = receive_index + 1;
+        received_pkt.push_back(cfg.bif.usb_p);
```
rswarbrick, line 378: Nit: Indentation.

hw/ip/usbdev/dv/env/usbdev_env.sv
```diff
@@ -18,9 +18,12 @@ class usbdev_env extends cip_base_env #(
     super.build_phase(phase);
     // get vifs
     if (!uvm_config_db#(virtual clk_rst_if)::get(this, "", "aon_clk_rst_vif",
-        cfg.aon_clk_rst_vif)) begin
+                                                 cfg.aon_clk_rst_vif)) begin
       `uvm_fatal(get_full_name(), "failed to get aon_clk_rst_if from uvm_config_db")
     end
+    if (!uvm_config_db#(virtual usb20_block_if)::get(this, "", "bif", cfg.bif)) begin
+        `uvm_fatal(`gfn, "failed to get usb20_block_if handle from uvm_config_db")
```
rswarbrick, line 25: Nit: Two spaces too many here.

(But thanks for fixing the indentation a couple of lines further up)

rswarbrick (changes requested):

hw/dv/sv/usb20_agent/usb20_agent_cfg.sv
MubashirSaleem775: ok

hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
MubashirSaleem775: okay

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: sure

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: okay

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: ok

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: ok

hw/ip/usbdev/dv/env/usbdev_env.sv
MubashirSaleem775: ok

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: i tested but it works only when there is a time out case for device response it not works.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -306,10 +308,37 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+        begin
+          repeat (18) begin
+            @(posedge cfg.bif.usb_clk);
+            if (cfg.bif.usb_dp_en_o) break;
+          end
+          if (!cfg.bif.usb_dp_en_o) begin
+            timed_out = 1'b1;
+            disable get_device_response;
+          end
+        end
```
rswarbrick, line 331: Wouldn't it be simpler to put the 18 clock wait into the top of `get_device_response`? That would also allow us to get rid of the complicated fork/join stuff. Something like this maybe? (untested!)
```systemverilog
...
    bit bitstuff_err;
    bit use_negedge;
    int time_before_response = 0;

    // Wait up to 18 clock cycles for the start of a response from the device, matching section
    // 7.1.19.1 of the USB specification.
    //
    // TODO: We currently spot this by looking at the driver enable of the DUT. We shouldn't really
    //       be stealing access to that signal and should be inferring that the device is responding
    //       by just looking at the USB_P/USB_N signals.
    for (time_before_response = 0; time_before_response < 18; time_before_response++) begin
      if (cfg.bif.usb_dp_en_o) break;
      @(posedge cfg.bif.usb_clk);
    end
    if (time_before_response >= 18) begin
      rsp_item.m_pid_type = pid_type_e'(PidTypeTimeOut);
      return;
    end

    // TODO: Operating on a div 4 clock is inherently fraught and runs into sampling problems;
    // a simple solution would be to use the 4x oversampling and detect the initial transition away
    // from Idle state, then choose an appropriate sampling phase [0,3] on the 4x clock.
    //
    // Here, for now, we just choose the clock edge to sidestep sampling errors.
...
```

@alees24: Have I understood the spec correctly here? I'm definitely not an expert... In particular, I *think* I've understood your TODO comment, but I had to think quite hard about it! Is my expansion right?

hw/dv/sv/usb20_agent/usb20_agent_cfg.sv
```diff
@@ -14,4 +14,13 @@ class usb20_agent_cfg extends dv_base_agent_cfg;
...
   `uvm_object_new
 
+  bit usb_bitstuff_error = 1'b0; // This flag will be utilized in error scenarios to verify
+  // the functionality of bit stuffing errors in the device. Set this flag in error cases,
+  // check in the driver if it's set, then generate a bit stuffing error.
+
+  bit single_bit_SE0 =1'b0; // This flag serves to evaluate the device's functionality,
+  // particularly regarding the recognition of a single SE0 bit as an end-of-packet,
+  // which necessitates two successive bits otherwise. Once set, it is accompanied by
+  // its respective test sequence, followed by the transmission of a single-bit SE0 as EOP.
```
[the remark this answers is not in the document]
MubashirSaleem775: okay

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -203,6 +201,10 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   task bit_stuffing(input bit packet[], output bit bit_stuff_out[]);
     int consecutive_ones_count = 0;
     bit stuffed[$];
+    if (cfg.usb_bitstuff_error) begin
+      bit_stuff_out = packet;
+      return;
+    end
```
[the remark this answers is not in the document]
MubashirSaleem775: sure, i will do that way. Thanks

hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
```diff
@@ -15,22 +15,45 @@ package usb20_agent_pkg;
...
-      bmRequestType2 = 8'b0_00_00010, bmRequestType3 = 8'b1_00_00000,
-      bmRequestType4 = 8'b1_00_00001, bmRequestType5 = 8'b1_00_00010} bmrequesttype_e;
-
-  typedef enum byte {bRequestGET_STATUS = 8'h00, bRequestCLEAR_FEATURE = 8'h01,
-      bRequestSET_FEATURE = 8'h03, bRequestSET_ADDRESS = 8'h05, bRequestGET_DESCRIPTOR = 8'h06,
-      bRequestSET_DESCRIPTOR = 8'h07, bRequestGET_CONFIGURATION = 8'h08,
-      bRequestSET_CONFIGURATION = 8'h09, bRequestGET_INTERFACE = 8'h0A,
-      bRequestSET_INTERFACE = 8'h0B, bRequestSYNCH_FRAME = 8'h0C} brequest_e;
+  typedef enum bit [7:0] {
+    PidTypeOutToken   = 8'b0001_1110,
```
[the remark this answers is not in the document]
MubashirSaleem775: yes its doable and its done in other PR by @alees24

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: This not works properly i have tested this. its works well for timeout case. for other scenario it didn't works.

rswarbrick: Ok, thanks. Could you explain what doesn't work?

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -212,6 +215,13 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
         end
       end else consecutive_ones_count = 0;
     end
+    if (cfg.usb_bitstuff_error) begin
+      // Pick a random 6-bit region and set all those bits to one for bitstuff err.
+    int start_index = $urandom_range(16, stuffed.size() - 7);
+    for (int j = start_index; j < start_index + 7; j++) begin
+        stuffed[j] = 1'b1;
+    end
```
rswarbrick, line 223: Please fix the indentation here. (I think you need two spaces more on each line)

rswarbrick (changes requested):
I've spent a vast amount of time reviewing this code and I'm *still* finding silly indentation problems.

And your practice of hiding conversations means that I have to wade through all the (MANY!) reviews that I have already written on this PR to figure out what is going on.

Reviewing this code is taking a significant chunk of my time and it's very unclear that it's worth the effort. Please tidy things up and respond in sentences to the reviews that I'm leaving.

hw/dv/sv/usb20_agent/usb20_agent_pkg.sv
```diff
@@ -17,28 +17,51 @@ package usb20_agent_pkg;
...
+    // Handshake PIDs
+    PidTypeAck   = 8'b1101_0010,
+    PidTypeNak   = 8'b0101_1010,
+    PidTypeStall = 8'b0001_1110,
+    PidTypeNyet  = 8'b1001_0110,
+    // Special PIDs
+    PidTypePre   = 8'b0011_1100,
+    PidTypeSplit = 8'b0111_1000,
+    PidTypePing  = 8'b1011_0100,
+    PidTypeTimeOut = 8'b1100_1100
```
alees24, line 39: I'm not happy with introducing invented PID values because there is nothing that stops the invented value being transmitted/received over the USB in the presence of communication errors, device malfunction etc.

The PID is specified in terms of its 4 LSBs, which are then complemented to provide the full 8 bits that are transmitted/received over the wire.

By introducing additional values, for DV convenience, we compromise the error detection as well as increasing the risk of the DV code misbehaving, not least because we presently do not even perform a validity check on the received/decoded PID.

Please use an additional variable/state information to indicate exceptional conditions such as timeout, since there shall be additional error condition over time. The additional information probably needs to be stored in the 'usb20_item' object.

alees24 (changes requested):

hw/dv/sv/usb20_agent/usb20_driver.sv
alees24: A few observation here:

    this approach may stop the packet being decoded by the USB device, leading to spurious test failures, because the randomized range overlaps the device address and endpoint number too.
    the code here is choosing a 7-bit region not the 6 bits annotated in the comment.
    the final bit (a '1') of the SYNC pattern is included in the bit stuffed sequence, although a valid PID cannot induce bit stuffing until the affected 6-bit sequence involves the final 3 bits (the MSBs) of the PID and the first 3 bits (LSBs) of the device address.
    a more robust approach, if the test is expected to guarantee a bit stuffing violation would be to have it control the device address, endpoint number and/or contents of the data packet, and then deactivate the bit stuffing functionality of the usb20_driver when driving that packet onto the USB.

hw/dv/sv/usb20_agent/usb20_driver.sv
MubashirSaleem775: The suggested code isn't functioning properly. In a fork-join scenario, there are two possibilities: either the device responds with a packet based on the current transaction, or it doesn't respond at all, and we must wait up to 18 bit times for the transaction to time out and end the test. Your suggested code handles the timeout scenario well, but when the device responds with packets, it often results in incorrect packet identification (PID) or packet sizes smaller than expected. However, the monitor still accurately captures the packets. The main issue is that the code doesn't initiate at the right time to capture the device's response.

MubashirSaleem775:
@rswarbrick @alees24  I am going to close this PR because there a few changes required in dv Env not all in this PR. I have tried to drop extra changes but not able to drop. So  I have added a new PR (#22722). Kindly review. Thanks

PR #21838, opened 2024-03-05
[uart] Further limit baud in uart_rx_oversample

elliotb-lowrisc:
Constrain the baud rate to 9600.

At high baud rates the oversample bit capture window is too small to reliably drive every oversample bit correctly due to the timing variation from CDC sync and inexact clk/baud ratios.

Also includes an improvement to `find_oversampled_clk_center` to adjust for the register read delay using a fixed offset rather than a proportion of the baud clock. Not required at 9600 baud, but the increased accuracy might be needed if an intermediate baud rate is introduced later.


hw/ip/uart/dv/env/seq_lib/uart_rx_oversample_vseq.sv
```diff
@@ -87,7 +94,7 @@ class uart_rx_oversample_vseq extends uart_tx_rx_vseq;
       #(get_oversampled_baud_clk_period_ns() * 1ns);
     end
     cfg.m_uart_agent_cfg.vif.uart_rx = 1; // back to default value
-    // launch check in a parallel thread
+    // launch check in a parallel thread to avoid loosing clock alignment in the main thread
```
rswarbrick, line 97: Nit: I think you probably want "losing" here :-)

rswarbrick (commented):
This looks really good, thanks! Minor nitty grammar note, but well done!

rswarbrick (approved):
This looks good, thanks.

Merging before CI completes because the only change from a version that has passed CI is the text in a comment!

PR #21843, opened 2024-03-05
[sival,usbdev] Remove test of usbdev pin-flipping from SiVal

alees24:
With the present SiVal test hardware it is not possible to test pin-flipping fully without a compensatory external flip of the pins.
Change the set of pin configurations to match that tested on the CW310/340 FPGAs.


pamaury (approved):

alees24:
Change affects only usbdev_pincfg_test which now passes.

Issue #16693, opened 2022-12-05
[uart] Convert edge based interrupt to level 


tjaychen:
Split out from #15378:

- [x] tx_watermark
- [x] rx_watermark
- [x] tx_empty should remain event-type

eunchan:
CC: @weicaiyang

tjaychen:
do you guys think this should be a backlog issue? It would be very nice to fix, but UART is pretty close to V3 already...

tjaychen:
actually...an interrupt change from edge triggered to status might be considered a major change if semantic versioning were to go through.. @msfschaffner

weicaiyang:
I'm OK with either way. If we want to fix it, it's nice to fix all the watermarks, so that we have a consistent behavior across the project.

msfschaffner:
If it's easy to fix I would do it now, otherwise we will have to open this up again. And I have a feeling that the status kind is more amenable to SW drivers... 

CC @a-will @timothytrippel

GregAC:
Need to determine if this is desired for M2.5

GregAC:
Given comments in https://github.com/lowRISC/opentitan/issues/15378 this isn't needed for M2.5

msfschaffner:
FYI @jesultra @moidx

GregAC:
Total effort upped to 1.5 days. PR has been done https://github.com/lowRISC/opentitan/pull/21632 though still needs review and need to check if there's anything to adjust in DIFs/top-level tests.

jettr:
The two watermark interrupt moving to status SGTM.

For the empty interrupt, it would be nice for there to be consistency across other empty interrupts, e.g. fifo_empty. If fifo_empty is status, then it seems like uart's tx_empty should be status as well  (see also #15378)

GregAC:
PR here to switch `tx_empty` to a status type: https://github.com/lowRISC/opentitan/pull/21844

@jesultra @jettr  it turns out the existing `tx_empty` was actually the "transmission complete" interrupt you describe it triggers when the FIFO is empty and the TX is idle. From a comment in the RTL this was done to avoid an immediate `tx_empty` interrupt when you write a single byte to the FIFO (as it will be immediately popped off and transmitted).

We could alter this to a 'proper' `tx_empty` that only cares about the state of the FIFO and rename the existing one to `tx_done` or similar or just leave it as is. I'd prefer the latter as it's less change. What are your thoughts?

jettr:
[...]

PR #21844, opened 2024-03-05
[uart, rtl] Switch tx_empty to status type interrupt

GregAC:


GregAC:
From discussion in https://github.com/lowRISC/opentitan/issues/16693 this is on hold for now. The desire is for `tx_empty` to only take into account the FIFO level. The current version fires with the FIFO empty and TX to be idle. An event type `tx_done` interrupt is wanted that has that same behaviour which is actually what we currently call `tx_empty`.

So as we're out of time for M2 RTL changes this work is on hold, `tx_empty` remains an event type. For M4 we'll introduce a `tx_done` and make `tx_empty` status type but only considering the FIFO level.

elliotb-lowrisc:
Closing as overtaken by events

andreaskurth:
For context: This PR was superseded by https://github.com/lowRISC/opentitan/pull/23409.

PR #21203, opened 2024-02-05
[opentitantool] Support for bitbanging

jesultra:
Some tests need to generate a pulse of a precise duration (when testing e.g. debouncing logic), generate edges on separate GPIOs with a particular delay between the two, or initiate I2C or SPI communication at a precise time after RESET has been released.

For such cases, better precision can be achieved if the HyperDebug EC does the timing, rather than Rust (or go-language) scripts managing the clock, and sending separate USB requests with millisecond latency for each operation.

This PR adds a feature to produce arbitrary waveforms on multiple GPIO pins, it is easy to generate pulses or edges on multiple pins.  For I2C/SPI communication with precise timing, the callers will have to themselves construct the waveform, which will be bit-banged by HyperDebug, (rather than generated using the internal peripherals of HyperDebug).  Still, this can be useful for testing whether software/hardware responds properly to requests arriving exactly at the most inopportune times.

Example uses below.

Generate a keypress (falling edge followed by rising edge) of a precise
duration:
```
opentitantool gpio bitbang --clock "0.1 ms" PWR_BTN_L -s "0 9.5ms 1"
```

Generate a break condition of 40 bit times, followed by 0x33 character:
```
opentitantool gpio bitbang --clock "115.2 kHz" UART_TX -s "1 100us 0 40ticks 10110011001"
```

Reset Ti50, then generate I2C start condition after exactly 250ms,
followed by incomplete address byte:
```
opentitantool gpio bitbang --clock "100 kHz" RESET DEVICE_I2C_SDA DEVICE_I2C_SCL \
  -s "011 50ms 111 250ms 101 100 101 100 101 100 101 100 111 111"
```

Corresponding change to HyperDebug firmware:
https://chromium-review.googlesource.com/c/chromiumos/platform/ec/+/5262904

PR #21845, opened 2024-03-05
[sival] Add UART parity test

### LINKED ISSUE #19865: [chip-test] chip_sw_uart_parity
### Test point name

chip_sw_uart_parity

### Description

Verify transmission of data in both directions with parity enabled.

This is very similar to chip_sw_uart_tx_rx except that parity (generation and checking)
should be enabled. Control expected word parity by configuring CTRL.PARITY_ODD register.

Test the rx_parity_err interrupt by sending a word with incorrect parity.

### Host side component

Rust

### OpenTitanTool infrastructure implemented

Yes

### Silicon Validation (SiVal)

Yes

### Emulation Targets

- [ ] None
- [ ] CW310
- [ ] Hyperdebug + CW310

### Contact person

@rswarbrick

### Checklist

Please fill out this checklist as items are completed. Link to PRs and issues as appropriate.
- [ ] Check if existing test covers most or all of this testpoint (if so, either extend said test to cover all points, or skip the next 3 checkboxes)
- [ ] Device-side (C) component developed
- [ ] Bazel build rules developed
- [ ] Host-side component developed
- [ ] Test added to dvsim nightly regression (and passing at time of checking)
- [ ] For SiVal test cases, test is running relevant FPGA or silicon regression

> estimate 3

jwnrt:
This tests the UART's parity bit feature:

1. Data is sent and received for both parities on all four UARTS.
2. The host checks that the transmitted data's parity.
3. The device checks the received data's parity.
4. The host sends some data with the incorrect parity and the device checks for the `rx_parity_err` interrupt.

The codes to send to the HyperDebug USB interface were taken from [this firmware source](https://chromium.googlesource.com/chromiumos/platform/ec/+/refs/heads/main/chip/stm32/usart.c#177).

Resolves #19865


sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,233 @@
...
+// updated according to the uart we select.
+static volatile uint32_t uart_peripheral_id;
+static volatile uint32_t uart_irq_rx_parity_err_id;
+
+// Whether we expect and have received the RX parity error interrupt.
+static volatile bool uart_irq_rx_parity_err_expected = false;
+static volatile bool uart_irq_rx_parity_err_fired = false;
+
+enum {
+  kCommandTimeout = 5000000,  // microseconds
```
pamaury, line 44: nit: maybe kCommandTimeoutMicros?

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,233 @@
...
+
+// Some random bytes that will be sent and received to check the parity.
+static const uint8_t kUartData[32] = {
+    0x3f, 0x39, 0xb0, 0x4e, 0xa6, 0xce, 0xe5, 0xb7, 0x94, 0x48, 0xec,
+    0xb5, 0x48, 0x5c, 0x08, 0x5b, 0xcd, 0x47, 0xae, 0x80, 0xbb, 0x49,
+    0xa1, 0x7c, 0x39, 0x20, 0xd1, 0x6d, 0x2f, 0x4f, 0x94, 0xd8,
+};
+
+// Configure the UART under test selected by `uart_idx` with `parity`.
+static status_t configure_uart(void) {
```
pamaury, line 64: There is a similar (but more complex) function in the uart tx/rx test, maybe this could be moved to a test util?

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,233 @@
...
+  // Claim the interrupt.
+  dif_rv_plic_irq_id_t plic_irq_id;
+  CHECK_DIF_OK(dif_rv_plic_irq_claim(&rv_plic, kTopEarlgreyPlicTargetIbex0,
+                                     &plic_irq_id));
+
+  // Check the interrupt fired on the correct UART.
+  top_earlgrey_plic_peripheral_t peripheral = (top_earlgrey_plic_peripheral_t)
+      top_earlgrey_plic_interrupt_for_peripheral[plic_irq_id];
+  CHECK(peripheral == uart_peripheral_id,
+        "Interurpt from unexpected peripheral: %d", peripheral);
```
pamaury, line 125: nit: "Interurpt"

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,233 @@
...
+
+  // Check it was the parity error that fired and that we expected it.
+  CHECK(plic_irq_id == uart_irq_rx_parity_err_id);
```
pamaury, line 128: maybe add a message? `"Unexpected interrupt from UART: %d", plic_irq_id`

sw/device/tests/uart_parity_test.c
jwnrt: I did think about this since it's copied from that test, however I wasn't a fan of having the test_util share its global state (the statics containing the interrupts etc) with tests. Maybe all of the interrupts and base addresses could be put into some structs which the testutil returns, but that was also a little complex.

Do you mind if I leave this duplicated in the test for now? I'm writing another test for line breaks which might need something similar so I will consider extracting it for that test

sw/device/tests/BUILD
```diff
@@ -5315,3 +5315,45 @@ opentitan_test(
         "//sw/device/lib/testing/test_framework:ottf_main",
     ],
 )
+
+opentitan_test(
+    name = "uart_parity_test",
+    srcs = ["uart_parity_test.c"],
+    cw310 = new_cw310_params(
+        otp = "//hw/ip/otp_ctrl/data/earlgrey_a0_skus/sival:otp_img_prod_manuf_personalized",
```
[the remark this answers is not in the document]
jwnrt: Good catch, I lazily copied this test definition from the `uart_tx_rx_test`. I've also removed dependencies on `clkmgr` files that aren't needed anymore.

pamaury (approved):
Looks good to me

sw/device/tests/uart_parity_test.c
pamaury: Sure, I was thinking a struct but maybe it's a bit overkill.

jwnrt: On second thought, I've added it to this PR instead.

pamaury:
Should this test be enabled for silicon targets as well?

jwnrt:
Yes, I've added the exec env and support in the teacup config.

jwnrt:
This is passing CI but I'm not confident in its robustness. Sometimes it gets stuck for reasons I don't understand. I'd like to hold off on merging until I understand the reasons it was getting stuck.

jwnrt:
We found the likely cause of the flakiness: configuring the parity in HyperDebug causes it to invent a NULL byte sometimes that definitely wasn't on the wire. I've worked around this by clearing the RX buffer _after_ we set the parity. Hopefully that's enough to avoid flakiness.

jwnrt:
Sorry, just updated again to make the change in `uart_tx_rx` too since that's been flaky.

jesultra:
I will look into the spurious NULL bytes from HyperDebug.

In the meantime, could you please implement the new `set_parity()` trait function also in `transport/proxy/uart.rs`.  It should be straightforward how to add declarations to the `proxy/protocol.rs` similar to how it is done for e.g. `set_baudrate()`.

jesultra:
And a side note.  If you want to test how the OpenTitan chip reacts to various framing errors, then maybe the bitbanging functionality of HyperDebug could be useful.  It allows precise control of timing between characters, or between a break condition and the following character.

https://github.com/lowRISC/opentitan/pull/21203

Example: Generate a break condition of 40 bit times, followed by 0x33 character:
```
opentitantool gpio bitbang --clock "115.2 kHz" UART_TX -s "1 100us 0 40ticks 10110011001"
```

jwnrt:
> And a side note. If you want to test how the OpenTitan chip reacts to various framing errors, then maybe the bitbanging functionality of HyperDebug could be useful. It allows precise control of timing between characters, or between a break condition and the following character.
> 
> #21203
> 
> Example: Generate a break condition of 40 bit times, followed by 0x33 character:
> 
> ```
> opentitantool gpio bitbang --clock "115.2 kHz" UART_TX -s "1 100us 0 40ticks 10110011001"
> ```

Thanks, this would actually be super useful for the line break test where we need to check various level triggers so the timing matters

sw/host/opentitanlib/src/io/uart.rs
```diff
@@ -9,6 +9,7 @@ use std::time::Duration;
 use anyhow::Result;
 use clap::Args;
 use serde::{Deserialize, Serialize};
+use serialport::Parity;
```
jesultra, line 12: Could you make this a `pub use`, so that `opentitantool` or other users of `io::uart` do not need to directly refer to the `serialport` library, when using our io traits.

jesultra:
> > Example: Generate a break condition of 40 bit times, followed by 0x33 character:
> > ```
> > opentitantool gpio bitbang --clock "115.2 kHz" UART_TX -s "1 100us 0 40ticks 10110011001"
> > ```
> 
> Thanks, this would actually be super useful for the line break test where we need to check various level triggers so the timing matters

Just remember that you have to change the mode of the HyperDebug pin from "Alternate" into "PushPull", before you can do bitbanging.  (And change it back afterwards, in order to use normal UART function again.)

I have so far not been able to reproduce the issue of getting stray 0x00 characters on changing parity, and I have some other urgent issues to look at, so I hope that your workaround of clearing the receive buffer after each switch is sufficient.  If you craft me some sort of stress test that can reliably reproduce the issue, then I could take another look.

sw/host/opentitanlib/src/io/uart.rs
jwnrt: That's a good idea, will do

sw/device/lib/testing/uart_testutils.h
```diff
@@ -55,4 +68,14 @@ OT_WARN_UNUSED_RESULT
 status_t uart_testutils_detach_pinmux(const dif_pinmux_t *pinmux,
                                       uint8_t uart_idx);
 
+/**
+ * Look up the configuration parameters for a given UART instance.
+ *
+ * @param uart_idx The index of the UART being configured.
+ * @param params Pointer to the struct to write the parameters to.
```
engdoreis, line 75:
```suggestion
 * @param[out]  Pointer to the struct to write the parameters to.
```

sw/host/opentitanlib/src/proxy/protocol.rs
```diff
@@ -142,6 +143,9 @@ pub enum UartRequest {
     SetBaudrate {
         rate: u32,
     },
+    SetParity {
```
engdoreis, line 146: Nit. I would use a tuple since there's only one attribute.
```rust
SetParity(Parity)
```

hw/top_earlgrey/data/ip/chip_uart_testplan.hjson
```diff
@@ -93,6 +93,7 @@
       lc_states: ["PROD"]
       features: ["UART.PARITY"]
       tests: []
+	  bazel: ["//sw/device/tests:uart_parity_test"]
```
engdoreis, line 96: The indentation seems wrong.

sw/device/tests/BUILD
```diff
@@ -5576,3 +5576,44 @@ opentitan_test(
...
+    cw310 = new_cw310_params(
+        test_cmd = " ".join([
+            "--bootstrap=\"{firmware}\"",
+            "--firmware-elf=\"{firmware:elf}\"",
+        ]),
+        test_harness = "//sw/host/tests/chip/uart:uart_parity",
+    ),
+    exec_env = {
+        "//hw/top_earlgrey:fpga_cw310_sival": None,
+        "//hw/top_earlgrey:fpga_cw310_sival_rom_ext": None,
```
engdoreis, line 5592: I think that we should add the cw340 if the exec_env for it is merged before this one.

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,211 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/device/lib/arch/device.h"
+#include "sw/device/lib/base/mmio.h"
+#include "sw/device/lib/dif/dif_base.h"
+#include "sw/device/lib/dif/dif_clkmgr.h"
+#include "sw/device/lib/dif/dif_lc_ctrl.h"
```
engdoreis, line 9: These two headers are not used and can be dropped.
Please also check the Bazel file.

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,211 @@
...
+#include "sw/device/lib/testing/uart_testutils.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+static dif_pinmux_t pinmux;
+static dif_rv_plic_t rv_plic;
+static dif_uart_t uart;
+
+// Parameters of the test to be set at the beginning.
+static volatile uint8_t parity = 0xff;
```
engdoreis, line 31:
```suggestion
static volatile uint8_t parity = UINT8_MAX;
```

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,211 @@
...
+
+  LOG_INFO("Receiving data with correct parity...");
+  while (bytes_to_recv > 0) {
+    size_t len = 0;
+    TRY(dif_uart_bytes_receive(&uart, bytes_to_recv, recv_ptr, &len));
+    bytes_to_recv -= len;
+    recv_ptr += len;
+  }
+
+  CHECK_ARRAYS_EQ(recv_buf, kUartData, ARRAYSIZE(kUartData));
```
engdoreis, line 155:
```suggestion
  TRY_CHECK_ARRAYS_EQ(recv_buf, kUartData, ARRAYSIZE(kUartData));
```

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,211 @@
...
+
+  // Wait for host to sync and then receive data but now expecting the parity
+  // error interrupt to trigger.
+  uart_irq_rx_parity_err_expected = true;
+  OTTF_WAIT_FOR(test_phase == kTestPhaseRecvErr, kCommandTimeoutMicros);
+
+  LOG_INFO("Receiving data with wrong parity");
+  IBEX_SPIN_FOR(uart_irq_rx_parity_err_fired, 1 * 1000 * 1000);
```
engdoreis, line 163: Can you define a const like `kCommandTimeoutMicros` for this timeout?

sw/device/tests/uart_parity_test.c
```diff
@@ -0,0 +1,211 @@
...
+
+bool test_main(void) {
+  mmio_region_t base_addr;
+  base_addr = mmio_region_from_addr(TOP_EARLGREY_PINMUX_AON_BASE_ADDR);
+  CHECK_DIF_OK(dif_pinmux_init(base_addr, &pinmux));
+  base_addr = mmio_region_from_addr(TOP_EARLGREY_RV_PLIC_BASE_ADDR);
+  CHECK_DIF_OK(dif_rv_plic_init(base_addr, &rv_plic));
+
+  // Wait for host to tell us which parity and UART to test.
+  OTTF_WAIT_FOR(uart_idx != 0xff && parity != 0xff, kCommandTimeoutMicros);
```
engdoreis, line 178: Nit
```suggestion
  OTTF_WAIT_FOR(uart_idx != UINT8_MAX && parity != UINT8_MAX, kCommandTimeoutMicros);
```

engdoreis (commented):
Thanks for the PR, I left some minor comments.

hw/top_earlgrey/data/ip/chip_uart_testplan.hjson
jwnrt: Thanks, I'd added a literal tab character somehow.

sw/device/tests/uart_parity_test.c
jwnrt: Thanks, this test was originally part of `uart_tx_rx` and I forgot to remove these.

PR #21848, opened 2024-03-05
[adc_ctrl/doc] Tweak ADC requirements

msfschaffner:
Some tweaks as suggested in https://github.com/lowRISC/opentitan/pull/21796#discussion_r1512198507

### DIFF
--- a/hw/ip/adc_ctrl/doc/programmers_guide.md
+++ b/hw/ip/adc_ctrl/doc/programmers_guide.md
@@ -45,11 +45,14 @@ The debug accessory voltage range of interest is shown in the diagram below:
 The ADC can be used to detect debug cable connection / disconnection in the non-overlapping regions.
 As an example use case of the two channel filters they can be used for detection of a USB-C debug accessory.
 The ADC must meet some minimum specifications:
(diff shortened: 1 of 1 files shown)


hw/ip/adc_ctrl/doc/programmers_guide.md
```diff
@@ -45,12 +45,15 @@ The debug accessory voltage range of interest is shown in the diagram below:
 The ADC can be used to detect debug cable connection / disconnection in the non-overlapping regions.
 As an example use case of the two channel filters they can be used for detection of a USB-C debug accessory.
 The ADC must meet some minimum specifications:
-* Full scale range must be 0.0V to 2.2V (note: the ADC in Earlgrey has a full range of 0.0V - 2.3V)
-* If the signal is below 0.0V the ADC value must be zero.
+* The minimum voltage for the ADC to produce a valid result is 0.0V or less
```
jettr, line 48:
```suggestion
* The minimum voltage for the ADC to produce a valid result is 0.0V
```
What do you think about dropping `or less` since the 3rd requirement states anything less than 0.0V should be clipped to a `0` value?

msfschaffner: sounds good - fixed!

jettr (approved):

PR #21850, opened 2024-03-06
[sram_ctrl] Minor alignment of init and key request behavior

msfschaffner:
Previously, it was possible to re-request a new key while a memory init was ongoing. This can result in problematic behavior since the init done indication would still be set to true afterwards, which is misleading since part of the array has then been initialized with the a different key.

This patch changes that and ignores key requests that are issued while a memory initialization is pending.

On a similar note, it was possible to re-trigger a memory initialization while one was already ongoing, causing the existing initialization to restart. This is now also changed so that initialization requests are ignored while an initialization is still pending.

### DIFF
--- a/hw/ip/sram_ctrl/rtl/sram_ctrl.sv
+++ b/hw/ip/sram_ctrl/rtl/sram_ctrl.sv
@@ -205,10 +205,12 @@ module sram_ctrl
 
   // A write to the init register reloads the LFSR seed, resets the init counter and
   // sets init_q to flag a pending initialization request.
-  logic init_trig;
-  assign init_trig = reg2hw.ctrl.init.q & reg2hw.ctrl.init.qe;
+  logic init_trig, init_q;
+  assign init_trig = reg2hw.ctrl.init.q &&
+                     reg2hw.ctrl.init.qe &&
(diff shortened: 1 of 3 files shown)


moidx (approved):
Thanks @msfschaffner. QQ, are there any conditions where sram init is triggered by hardware? I assume that this is currently controlled by ROM, so no issues here as we can control the init sequence in firmware.

msfschaffner:
That is correct, only firmware can trigger SRAM init at the moment

hw/ip/sram_ctrl/rtl/sram_ctrl.sv
```diff
@@ -205,10 +205,12 @@ module sram_ctrl
 
   // A write to the init register reloads the LFSR seed, resets the init counter and
   // sets init_q to flag a pending initialization request.
-  logic init_trig;
-  assign init_trig = reg2hw.ctrl.init.q & reg2hw.ctrl.init.qe;
+  logic init_trig, init_q;
+  assign init_trig = reg2hw.ctrl.init.q &&
```
matutem, line 209: And thanks for using && vs & for boolean operations :)

matutem (approved):
LGTM

PR #21326, opened 2024-02-13
[usbdev] rx_fifo_sequence

Nadia-Hanif:
This is the draft PR for rx_fifo sequence. It is just for code review.
@jdonjdon kindly review it. Thanks

PR #21852, opened 2024-03-06
In status stage sequence

Nadia-Hanif:
This PR contains in status stage sequence.
The sequence verifies that the data size is zero, as this test validates the functionality of sending data with a zero length, followed by acknowledging the data through ACK pid. And we are verifying this by reading the data size from rxfifo.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_in_status_stage_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_in_status_stage_vseq.sv
@@ -1,0 +1,36 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// In status stage vseq
+class usbdev_in_status_stage_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_in_status_stage_vseq)
+
+  `uvm_object_new
+
+  task body();
+    // Configure out transaction
+    configure_out_trans();
+    // Out token packet followed by a zero length data packet
+    call_token_seq(PidTypeOutToken);
+    inter_packet_delay();
+    call_data_seq(PidTypeData0, .randomize_length(1'b0), .num_of_bytes(1'b0));
+    inter_packet_delay();
+    // Get response from DUT
+    get_response(m_response_item);
+    $cast(m_usb20_item, m_response_item);
+    get_out_response_from_device(m_usb20_item, PidTypeAck);
+    // The sequence verifies that the data size is zero, as this test validates the functionality
+    // of sending data with a zero length, followed by acknowledging the data through ACK pid.
+    // And we are verifying this by reading the data size from rxfifo.
+    check_rxfifo_empty();
+  endtask
+
+  task check_rxfifo_empty();
+    bit read_rxfifo_size;
+    // Get rx fifo size status
+    csr_rd(.ptr(ral.rxfifo.size), .value(read_rxfifo_size));
+    // DV_CHECK on data size
+    `DV_CHECK_EQ(read_rxfifo_size, 0);
+  endtask
+endclass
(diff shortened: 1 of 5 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_in_status_stage_vseq.sv
```diff
@@ -0,0 +1,37 @@
...
+  task body();
+    num_of_bytes = 0;
+    rand_or_not = 0;
+
+    super.dut_init("HARD");
+    // Configure out transaction
+    configure_out_trans();
+    // Out token packet followed by a zero length data packet
+    call_token_seq(PidTypeOutToken);
+    cfg.clk_rst_vif.wait_clks(10);
```
rswarbrick, line 20: (Following up on: https://github.com/lowRISC/opentitan/pull/21326#discussion_r1511332046)

Thanks for explaining how we have to wait for a couple of bit times here. Please could you factor this out into a `wait_two_bit_times` task? (feel free to choose a better name!) That way, it should be a bit more obvious what's going on, and we'll slightly decouple the USB clock and the main clock in the code, at least.

rswarbrick (changes requested):
Please can you squash the two commits together? I don't think I can see any reason that we'd want to keep them separate.

hw/ip/usbdev/dv/env/seq_lib/usbdev_in_status_stage_vseq.sv
```diff
@@ -0,0 +1,37 @@
...
+class usbdev_in_status_stage_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_in_status_stage_vseq)
+
+  `uvm_object_new
+
+  task body();
+    num_of_bytes = 0;
+    rand_or_not = 0;
+
+    super.dut_init("HARD");
```
rswarbrick, line 15: I can't see why this is necessary. Note that `dut_init` already gets called before the start of `body`.

hw/ip/usbdev/dv/env/seq_lib/usbdev_in_status_stage_vseq.sv
```diff
@@ -0,0 +1,37 @@
...
+
+// in status stage vseq
+class usbdev_in_status_stage_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_in_status_stage_vseq)
+
+  `uvm_object_new
+
+  task body();
+    num_of_bytes = 0;
+    rand_or_not = 0;
```
rswarbrick, line 13: I'd suggest pulling the configuration into a separate task like this:
```systemverilog
  task pre_start();
    super.pre_start();
    num_of_bytes = 0;
    rand_or_not = 0;
  endtask

  task body();
    // Configure out transaction
    configure_out_trans();
    // Out token packet followed by a zero length data packet
    call_token_seq(PidTypeOutToken);
...
```

hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
```diff
@@ -74,8 +74,13 @@
...
     {
-      name: usbdev_fifo_rst
-      uvm_test_seq: usbdev_fifo_rst_vseq
+      name: usbdev_in_status_stage
+      uvm_test_seq: usbdev_in_status_stage_vseq
+    }
+    {
+      name: usbdev_in_trans
+      uvm_test_seq: usbdev_in_trans_vseq
+>>>>>>> [usbdev] in status stage sequence
```
rswarbrick, line 83: Nit: Something has gone a bit wrong with the merging here. Could you tidy it up?

rswarbrick (commented):
There's a minor git problem that needs sorting out.

Otherwise, the vseq seems sensibly written, but I'm not an expert on USB and I don't properly understand the testpoint. @alees24: Would you mind taking a look at this? I don't properly understand the description in the testpoint, and I'm not sure I understand how the vseq is implementing it. Can you help?

alees24:

PR #21854, opened 2024-03-06
[usbdev,dv] usbdev resume link active seq PR

Saad22386:
This PR contains usbdev_resume_link_active_vseq which implements the resume functionality of usbdev from link_powered_state.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_resume_link_active_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_resume_link_active_vseq.sv
@@ -1,0 +1,31 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class usbdev_resume_link_active_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_resume_link_active_vseq)
+
+  `uvm_object_new
+
+  task body();
+    bit [2:0] powered_link_state;
+    bit       resume;
+
+    super.dut_init("HARD");
+    // Perform no operation for more than 3ms (3ms = 144,000 clk cycle at 48MHz)
+    cfg.clk_rst_vif.wait_clks(154000);
+    // Read link_state it should be in link powered state
+    csr_rd(.ptr(ral.usbstat.link_state), .value(powered_link_state));
+    `DV_CHECK_EQ(powered_link_state, 1);
+
+    // Set resume link state in usbctrl register
+    csr_wr(.ptr(ral.usbctrl.resume_link_active), .value(1'b1));
+    cfg.clk_rst_vif.wait_clks(5);
+
+    // Read link_resume interrupt
+    csr_rd(.ptr(ral.intr_state.link_resume), .value(resume));
+    // DV_CHECK on link_resume interrupt
+    `DV_CHECK_EQ(resume, 1);
+  endtask
+
+endclass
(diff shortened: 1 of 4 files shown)


Saad22386:
@rswarbrick, kindly review it
Thanks

hw/ip/usbdev/dv/env/seq_lib/usbdev_resume_link_active_vseq.sv
```diff
@@ -0,0 +1,31 @@
...
+class usbdev_resume_link_active_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_resume_link_active_vseq)
+
+  `uvm_object_new
+
+  task body();
+    bit [2:0] powered_link_state;
+    bit       resume;
+
+    super.dut_init("HARD");
```
rswarbrick, line 14: I don't think we actually need to do this extra initialisation. Is it required for the test at your end?

hw/ip/usbdev/dv/env/seq_lib/usbdev_resume_link_active_vseq.sv
```diff
@@ -0,0 +1,31 @@
...
+    // Perform no operation for more than 3ms (3ms = 144,000 clk cycle)
+    cfg.clk_rst_vif.wait_clks(154000);
```
rswarbrick, line 16: That's reasonable but (obviously) depends on the clock speed. I'd probably add "... at 48MHz" to the end of the comment.

rswarbrick (changes requested):
Looking at the testplan, I see:
```
            - Read the link state from usbstate register to see if it is in powered state and then send the
              resume signal (K state) for 20 ms and set the usbctrl.resume_link_active.
```

I don't think I can see anything in the code that waits 20ms. Can you explain how your code matches the testplan? I suspect that we either need to fix the testplan or there's a minor bug in the code itself.

Saad22386:
> Looking at the testplan, I see:
> 
> ```
>             - Read the link state from usbstate register to see if it is in powered state and then send the
>               resume signal (K state) for 20 ms and set the usbctrl.resume_link_active.
> ```
> 
> I don't think I can see anything in the code that waits 20ms. Can you explain how your code matches the testplan? I suspect that we either need to fix the testplan or there's a minor bug in the code itself.

We don't need to wait for 20ms i think description need to rewrite. To resume from link_powered state just need to set the usbctrl.resume_link_active.

hw/ip/usbdev/dv/env/seq_lib/usbdev_resume_link_active_vseq.sv
Saad22386: Yes it's required at our end

hw/ip/usbdev/dv/env/seq_lib/usbdev_resume_link_active_vseq.sv
Saad22386: Ok I'll add

Saad22386:
> Ok, that makes sense. Could you make that change? The best thing is probably to add another commit that comes first in this PR and tweaks the testpoint description.
> 
> If you think there might be some discussion, the other option would be to open a separate PR to track it.

yes i have made that change You can look at it.

rswarbrick:
I'm sorry: we're talking past each other slightly. I'm proposing that you change the testpoint description in a separate commit (and possibly separate PR). That way, the reviews of the two changes can proceed in parallel.

Saad22386:
> I'm sorry: we're talking past each other slightly. I'm proposing that you change the testpoint description in a separate commit (and possibly separate PR). That way, the reviews of the two changes can proceed in parallel.

can you look into #22129 that contains the change in test description?

PR #21857, opened 2024-03-06
[i2c,rtl] NACK after timing out on a stretch

marnovandermaas:
This is a change to our I2C target. It essentially adds another timeout control register that can be set in the target that keeps track of how long it has been stretching for and otherwise NACK. The first 6 commits have no functional changes, so you can focus your reviews on the latter 10 commits (some of these have just auto-generated changes which are labelled accordingly).

There are many design parameters here and you can read a more detailed discussion in the issue: https://github.com/lowRISC/opentitan/issues/15227 But here's my take on a few of these:

### When to stretch
There are two options here, stretch before the ACK or stretch afterwards. After discussion with @andreaskurth and @a-will we decided to stretch after the ACK to be closer to the current design and to align with I2C high-speed mode.

### Do we NACK an address?
To be compatible with SMBus, @andreaskurth suggested we should not NACK an address and instead NACK subsequent data bytes. This works fine for writes because the host sends bytes after the address and the acknowledge is in the control of the target. However, for reads we still need to NACK the address because after a read request the acknowledge is in control of the host.

### How to communicate a NACK to the OpenTitan software
We need to NACK when the ACQ FIFO is full, but there also needs to be a way to report this incident to OpenTitan software. I've chosen to reserve an extra spot in the ACQ FIFO so that we can always send an `AcqNack` in there. I've also added a `AckNackStart` for the case when we have actually acknowledged an address when our ACQ FIFO is full but will unconditionally NACK the next data byte. This means the state machine needs to continue for a little while so that is why `AcqNackStart` needs to be different from `AcqNack`. The subsequent NACK'ed data byte will not be put in the ACQ FIFO so we cannot put a `AcqNack` entry in there anymore that is why `AckNackStart` needs to be different from `AcqStart`.

Another request by @hcallahan-lowrisc was that we still record the byte that was NACK'ed in the ACQ FIFO. This is done automaticaly at the moment through the concatenation of `{AcqNack, input_byte}`. However, this may not be possible in the future if we compress the entries in the ACQ FIFO.

### NACK'ing subsequent bytes
At the moment, I go directly to idle after sending a NACK and do not explicitly NACK any subsequent bytes. @andreaskurth mentioned that in idle mode the pull-up on SDA will essentially act as a NACK.

### Draft status
I've marked this PR as draft because I've only done limited testing on this. Some tests are passing, but I have not investigated the failures. Let's first converge on the functionality and then we can do more thorough testing.

Closes: https://github.com/lowRISC/opentitan/issues/15227

### DIFF
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -570,7 +570,10 @@
       ]
     }
     { name: "TIMEOUT_CTRL"
-      desc: "I2C clock stretching timeout control"
+      desc: '''
+            I2C clock stretching timeout control.
+            This is used in I2C host mode to detect whether a connected target is stretching for too long.
+            '''
       swaccess: "rw"
       hwaccess: "hro"
       fields: [
@@ -660,6 +663,24 @@
       hwaccess: "hro"
       fields: [
         { bits: "31:0" }
+      ]
+    }
+    { name: "TARGET_TIMEOUT_CTRL"
+      desc: '''
+            I2C target internal stretching timeout control.
+            When the target has stretched beyond this time it will send a NACK.
+            '''
+      swaccess: "rw"
+      hwaccess: "hro"
+      fields: [
+        { bits: "30:0"
+          name: "VAL"
+          desc: "Clock stretching timeout value (in units of input clock frequency)"
+        }
+        { bits: "31"
+          name: "EN"
+          desc: "Enable timeout feature and send NACK once the timeout has been reached"
+        }
       ]
     }
   ]
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -621,7 +621,7 @@
           name: "ABYTE"
           desc: "Address for accepted transaction or acquired byte"
         }
-        { bits: "9:8"
+        { bits: "10:8"
           name: "SIGNAL"
           desc: "Host issued a START before transmitting ABYTE, a STOP or a RESTART after the preceeding ABYTE"
           enum: [
@@ -640,6 +640,14 @@
             { value: "3",
               name: "RESTART",
               desc: "ABYTE contains junk, START with address will follow"
+            },
+            { value: "4",
+              name: "NACK",
+              desc: "ABYTE contains either the address or data that was NACK'ed"
+            },
+            { value: "5",
+              name: "NACKSTART",
+              desc: "ABYTE contains the I2C address which was ACK'ed, but the block will continue and NACK the next data byte that was received: this only happens for writes"
             },
           ]
         }
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -691,5 +691,17 @@
         }
       ]
     }
+    { name: "TARGET_NACK_COUNT"
+      desc: '''
+            Number of times the I2C target has NACK'ed a new transaction since the last read of this register.
+            Reading this register clears it.
+            This is useful because when the ACQ FIFO is full the software know that a NACK has occurred, but without this register would not know how many transactions it missed.
+            '''
+      swaccess: "rc"
+      hwaccess: "hrw"
+      fields: [
+        { bits: "31:0" }
+      ]
+    }
   ]
 }
--- a/hw/ip/i2c/doc/registers.md
+++ b/hw/ip/i2c/doc/registers.md
@@ -492,30 +492,33 @@ I2C target address and mask pairs
 I2C target acquired data
 - Offset: `0x58`
 - Reset default: `0x0`
-- Reset mask: `0x3ff`
+- Reset mask: `0x7ff`
 
 ### Fields
 
 ```wavejson
-{"reg": [{"name": "ABYTE", "bits": 8, "attr": ["ro"], "rotate": 0}, {"name": "SIGNAL", "bits": 2, "attr": ["ro"], "rotate": -90}, {"bits": 22}], "config": {"lanes": 1, "fontsize": 10, "vspace": 80}}
+{"reg": [{"name": "ABYTE", "bits": 8, "attr": ["ro"], "rotate": 0}, {"name": "SIGNAL", "bits": 3, "attr": ["ro"], "rotate": -90}, {"bits": 21}], "config": {"lanes": 1, "fontsize": 10, "vspace": 80}}
 ```
 
 |  Bits  |  Type  |  Reset  | Name                       |
 |:------:|:------:|:-------:|:---------------------------|
-| 31:10  |        |         | Reserved                   |
-|  9:8   |   ro   |    x    | [SIGNAL](#acqdata--signal) |
+| 31:11  |        |         | Reserved                   |
+|  10:8  |   ro   |    x    | [SIGNAL](#acqdata--signal) |
 |  7:0   |   ro   |    x    | [ABYTE](#acqdata--abyte)   |
 
 ### ACQDATA . SIGNAL
 Host issued a START before transmitting ABYTE, a STOP or a RESTART after the preceeding ABYTE
 
-| Value   | Name    | Description                                         |
-|:--------|:--------|:----------------------------------------------------|
-| 0x0     | NONE    | ABYTE contains ordinary data byte as received       |
-| 0x1     | START   | ABYTE contains the 8-bit I2C address (R/W in lsb)   |
-| 0x2     | STOP    | ABYTE contains junk                                 |
-| 0x3     | RESTART | ABYTE contains junk, START with address will follow |
+| Value   | Name      | Description                                                                                                                                              |
+|:--------|:----------|:---------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 0x0     | NONE      | ABYTE contains ordinary data byte as received                                                                                                            |
+| 0x1     | START     | ABYTE contains the 8-bit I2C address (R/W in lsb)                                                                                                        |
+| 0x2     | STOP      | ABYTE contains junk                                                                                                                                      |
+| 0x3     | RESTART   | ABYTE contains junk, START with address will follow                                                                                                      |
+| 0x4     | NACK      | ABYTE contains either the address or data that was NACK'ed                                                                                               |
+| 0x5     | NACKSTART | ABYTE contains the I2C address which was ACK'ed, but the block will continue and NACK the next data byte that was received: this only happens for writes |
 
+Other values are reserved.
 
 ### ACQDATA . ABYTE
 Address for accepted transaction or acquired byte
@@ -570,5 +573,6 @@ When the target has stretched beyond this time it will send a NACK.
 |:------:|:------:|:-------:|:-------|:-----------------------------------------------------------------------|
 |   31   |   rw   |   0x0   | EN     | Enable timeout feature and send NACK once the timeout has been reached |
 |  30:0  |   rw   |   0x0   | VAL    | Clock stretching timeout value (in units of input clock frequency)     |
+
 
 <!-- END CMDGEN -->
--- a/hw/ip/i2c/doc/registers.md
+++ b/hw/ip/i2c/doc/registers.md
@@ -3,34 +3,35 @@
 <!-- BEGIN CMDGEN util/regtool.py -d ./hw/ip/i2c/data/i2c.hjson -->
 ## Summary
 
-| Name                                              | Offset   |   Length | Description                                                                          |
-|:--------------------------------------------------|:---------|---------:|:-------------------------------------------------------------------------------------|
-| i2c.[`INTR_STATE`](#intr_state)                   | 0x0      |        4 | Interrupt State Register                                                             |
-| i2c.[`INTR_ENABLE`](#intr_enable)                 | 0x4      |        4 | Interrupt Enable Register                                                            |
-| i2c.[`INTR_TEST`](#intr_test)                     | 0x8      |        4 | Interrupt Test Register                                                              |
-| i2c.[`ALERT_TEST`](#alert_test)                   | 0xc      |        4 | Alert Test Register                                                                  |
-| i2c.[`CTRL`](#ctrl)                               | 0x10     |        4 | I2C Control Register                                                                 |
-| i2c.[`STATUS`](#status)                           | 0x14     |        4 | I2C Live Status Register for Host and Target modes                                   |
-| i2c.[`RDATA`](#rdata)                             | 0x18     |        4 | I2C Read Data                                                                        |
-| i2c.[`FDATA`](#fdata)                             | 0x1c     |        4 | I2C Host Format Data                                                                 |
-| i2c.[`FIFO_CTRL`](#fifo_ctrl)                     | 0x20     |        4 | I2C FIFO control register                                                            |
-| i2c.[`HOST_FIFO_CONFIG`](#host_fifo_config)       | 0x24     |        4 | Host mode FIFO configuration                                                         |
-| i2c.[`TARGET_FIFO_CONFIG`](#target_fifo_config)   | 0x28     |        4 | Target mode FIFO configuration                                                       |
-| i2c.[`HOST_FIFO_STATUS`](#host_fifo_status)       | 0x2c     |        4 | Host mode FIFO status register                                                       |
-| i2c.[`TARGET_FIFO_STATUS`](#target_fifo_status)   | 0x30     |        4 | Target mode FIFO status register                                                     |
-| i2c.[`OVRD`](#ovrd)                               | 0x34     |        4 | I2C Override Control Register                                                        |
-| i2c.[`VAL`](#val)                                 | 0x38     |        4 | Oversampled RX values                                                                |
-| i2c.[`TIMING0`](#timing0)                         | 0x3c     |        4 | Detailed I2C Timings (directly corresponding to table 10 in the I2C Specification).  |
-| i2c.[`TIMING1`](#timing1)                         | 0x40     |        4 | Detailed I2C Timings (directly corresponding to table 10 in the I2C Specification).  |
-| i2c.[`TIMING2`](#timing2)                         | 0x44     |        4 | Detailed I2C Timings (directly corresponding to table 10 in the I2C Specification).  |
-| i2c.[`TIMING3`](#timing3)                         | 0x48     |        4 | Detailed I2C Timings (directly corresponding to table 10, in the I2C Specification). |
-| i2c.[`TIMING4`](#timing4)                         | 0x4c     |        4 | Detailed I2C Timings (directly corresponding to table 10, in the I2C Specification). |
-| i2c.[`TIMEOUT_CTRL`](#timeout_ctrl)               | 0x50     |        4 | I2C clock stretching timeout control.                                                |
-| i2c.[`TARGET_ID`](#target_id)                     | 0x54     |        4 | I2C target address and mask pairs                                                    |
-| i2c.[`ACQDATA`](#acqdata)                         | 0x58     |        4 | I2C target acquired data                                                             |
-| i2c.[`TXDATA`](#txdata)                           | 0x5c     |        4 | I2C target transmit data                                                             |
-| i2c.[`HOST_TIMEOUT_CTRL`](#host_timeout_ctrl)     | 0x60     |        4 | I2C host clock generation timeout value (in units of input clock frequency)          |
-| i2c.[`TARGET_TIMEOUT_CTRL`](#target_timeout_ctrl) | 0x64     |        4 | I2C target internal stretching timeout control.                                      |
+| Name                                              | Offset   |   Length | Description                                                                                        |
+|:--------------------------------------------------|:---------|---------:|:---------------------------------------------------------------------------------------------------|
+| i2c.[`INTR_STATE`](#intr_state)                   | 0x0      |        4 | Interrupt State Register                                                                           |
+| i2c.[`INTR_ENABLE`](#intr_enable)                 | 0x4      |        4 | Interrupt Enable Register                                                                          |
+| i2c.[`INTR_TEST`](#intr_test)                     | 0x8      |        4 | Interrupt Test Register                                                                            |
+| i2c.[`ALERT_TEST`](#alert_test)                   | 0xc      |        4 | Alert Test Register                                                                                |
+| i2c.[`CTRL`](#ctrl)                               | 0x10     |        4 | I2C Control Register                                                                               |
+| i2c.[`STATUS`](#status)                           | 0x14     |        4 | I2C Live Status Register for Host and Target modes                                                 |
+| i2c.[`RDATA`](#rdata)                             | 0x18     |        4 | I2C Read Data                                                                                      |
+| i2c.[`FDATA`](#fdata)                             | 0x1c     |        4 | I2C Host Format Data                                                                               |
+| i2c.[`FIFO_CTRL`](#fifo_ctrl)                     | 0x20     |        4 | I2C FIFO control register                                                                          |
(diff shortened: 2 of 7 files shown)


marnovandermaas:
In terms of testing these are my current results. Some target smoke tests are passing, but it would be nice if this percentage was higher. Surprisingly `i2c_target_stretch` is passing reliably although it would be good to extend this to include testing the timeout.

```
### Test Results
|  Stage  |         Name         | Tests                     |  Max Job Runtime  |  Simulated Time  |
Passing  |  Total  |  Pass Rate  |
|:-------:|:--------------------:|:--------------------------|:-----------------:|:----------------:|:---------:|:-------:|:-----------:|
|   V1    |     target_smoke     | i2c_target_smoke          |      5.633m       |    100.000ms     |   1     |   10    |   10.00 %   |
|   V1    |                      | **TOTAL**                 |                   |                  |   1     |   10    |   10.00 %   |
|   V2    |  target_error_intr   | i2c_target_unexp_stop     |      13.050s      |     17.108ms     |  10     |   10    |  100.00 %   |
|   V2    |    target_glitch     | i2c_target_glitch         |      6.990s       |     2.961ms      |   0     |   10    |   0.00 %    |
|   V2    |     target_perf      | i2c_target_perf           |      7.490s       |    829.233us     |  10     |   10    |  100.00 %   |
|   V2    |  target_fifo_empty   | i2c_target_stress_rd      |      10.451m      |    100.000ms     |   5     |   10    |   50.00 %   |
|         |                      | i2c_target_intr_smoke     |      11.900s      |     1.987ms      |  10     |   10    |  100.00 %   |
|   V2    |  target_fifo_reset   | i2c_target_fifo_reset_acq |      1.038m       |     10.134ms     |  10     |   10    |  100.00 %   |
|         |                      | i2c_target_fifo_reset_tx  |      1.127m       |     10.061ms     |  10     |   10    |  100.00 %   |
|   V2    |   target_fifo_full   | i2c_target_stress_wr      |      21.560s      |     23.100ms     |   3     |   10    |   30.00 %   |
|         |                      | i2c_target_stress_rd      |      10.451m      |    100.000ms     |   5     |   10    |   50.00 %   |
|         |                      | i2c_target_intr_stress_wr |      15.070s      |     11.895ms     |   2     |   10    |   20.00 %   |
|   V2    |    target_timeout    | i2c_target_timeout        |      12.170s      |     26.951ms     |  10     |   10    |  100.00 %   |
|   V2    | target_clock_stretch | i2c_target_stretch        |      13.465m      |     34.208ms     |  10     |   10    |  100.00 %   |
|   V2    |     bad_address      | i2c_target_bad_addr       |      9.120s       |     5.739ms      |  10     |   10    |  100.00 %   |
|   V2    |  target_mode_glitch  | i2c_target_hrst           |      5.580s       |    732.399us     |  10     |   10    |  100.00 %   |
|   V2    |                      | **TOTAL**                 |                   |                  |  100    |   130   |   76.92 %   |
|         |                      | **TOTAL**                 |                   |                  |  101    |   140   |   72.14 %   |
```

andreaskurth:
> ### Do we NACK an address?
> 
> To be compatible with SMBus, @andreaskurth suggested we should not NACK an address and instead NACK subsequent data bytes. This works fine for writes because the host sends bytes after the address and the acknowledge is in the control of the target. However, for reads we still need to NACK the address because after a read request the acknowledge is in control of the host.

I agree about NACKing our address on a read because we cannot serve the read (there's no space in the ACQ FIFO to inform SW that it should provide data) and we cannot NACK any subsequent bytes in this case (those NACKs are under the control of the controller). Note that SMBus protocols *always* start with a write, so there we'll ACK the address of the write and then the higher-level controller should NACK the command code _before_ the Sr and the read. Thus no risk of violating the SMBus spec IMO. MCTP anyway only does block writes, thus no risk of violating that either.

> ### How to communicate a NACK to the OpenTitan software
> 
> We need to NACK when the ACQ FIFO is full, but there also needs to be a way to report this incident to OpenTitan software. I've chosen to reserve an extra spot in the ACQ FIFO so that we can always send an `AcqNack` in there. I've also added a `AckNackStart` for the case when we have actually acknowledged an address when our ACQ FIFO is full but will unconditionally NACK the next data byte. This means the state machine needs to continue for a little while so that is why `AcqNackStart` needs to be different from `AcqNack`. The subsequent NACK'ed data byte will not be put in the ACQ FIFO so we cannot put a `AcqNack` entry in there anymore that is why `AckNackStart` needs to be different from `AcqStart`.

I don't understand this, sorry. How does reserving a spot in the ACQ FIFO for an `AcqNack` solve the problem in general? Suppose you put the `AcqNack` into the reserved spot and then another transfer that we need to NACK follows while the FIFO is still full (now without the reserved spot). What do you do now? 

> ### NACK'ing subsequent bytes
> 
> At the moment, I go directly to idle after sending a NACK and do not NACK any subsequent bytes. I think @a-will and @andreaskurth thought it might be good to NACK subsequent bytes as well. My take on this is that it is the host's responsibility to observe the NACK and send a stop or restart. However, this can still be changed if there is a good reason to.

When the FSM goes to idle, it will no longer ACK any subsequent bytes, right? So the pull-ups will let SDA look like NACK. So from an I/O perspective that should be okay, though it may or may not be fine for FSM-internal purposes.

marnovandermaas:
> I don't understand this, sorry. How does reserving a spot in the ACQ FIFO for an AcqNack solve the problem in general? Suppose you put the AcqNack into the reserved spot and then another transfer that we need to NACK follows while the FIFO is still full (now without the reserved spot). What do you do now?

Hmmm, good point. I guess we would have to continue NACK'ing on subsequent starts, and by that point completely drop them on the ground. I still think it's useful to record the fact that we NACK'ed in the ACQ FIFO. I was expecting the I2C host to do something sane here which is send a repeated start or a stop which should raise the command complete interrupt and hopefully get the OT software to empty the ACQ FIFO. After having a closer look at this, this interrupt does not get fired in Idle mode, so I should be sending that command complete event at the end of the NackHold state. Does that sound reasonable?

> When the FSM goes to idle, it will no longer ACK any subsequent bytes, right? So the pull-ups will let SDA look like NACK. So from an I/O perspective that should be okay, though it may or may not be fine for FSM-internal purposes.

Oh, that's good news. From the FSM's perspective this is fine. It only gets out of the idle state when it sees a start condition.

andreaskurth:
> ### When to stretch
> 
> Before our target would stretch after having already acknowledged a received address or byte. I think it's better to stretch before the acknowledge in case the ACQ FIFO is full so that hopefully some space will appear in the meantime. Then we can NACK if this is not the case.
> 
> I think @a-will expressed that we should keep the stretch after the ACK, but this means we have to unconditionally NACK a byte if the ACQ FIFO is full and I prefer to stretch at the latest possible moment. Both solutions could work though and I can change this to the other behavior

There seem to be two possible positions for stretching; either of:
(a) we ACK, our ACQ FIFO is now full, we **stretch until our timeout**, let controller clock data bits, we NACK
(b) we ACK, our ACQ FIFO is now full, let controller clock data bits, **we stretch until our timeout**, we NACK

Both options could potentially work. I see the following advantages for (b):
1. It gives SW a bit more time to free a slot in the ACQ FIFO.
2. It means that the stretch for write transactions (by us as target-receiver) happens at the same point as for read transfers (by the controller-receiver). (Note that for read transfers we as target-transmitter will no longer stretch because we cannot NACK. Instead, we will NACK our address + R bit.)

and the following advantages for (a):
1. As you wrote this is where we currently stretch (indefinitely), so keeping the stretch there means less changes to DD and DV.
2. It is certainly permitted by the I2C spec, which in section 3.1.9 states "Targets can then hold the SCL line LOW after reception and acknowledgment of a byte to force the controller into a wait state until the target is ready for the next byte transfer [...]." Also figure 6 shows clock stretching at this position, and I didn't find any other figure in the I2C spec that shows stretching at another position. Interestingly (and confusingly!) enough, SMBus allows target devices to stretch "[...] during any bit transfer including the clock provided prior to the ACK clock pulse." as well as "[...] after the reception and acknowledgment of a byte." (section 5.3.3). Nonetheless, I think an implementation of an I2C controller could implement the spec such that stretching for targets is only allowed in position (a).

So I think (a) is the better choice for us.

Thanks also to @a-will for the out-of-band inputs on this topic.

marnovandermaas:
@andreaskurth and @a-will thanks for this input. I understand your reasoning and you've convinced me to stretch after the ACK. I'll rework this PR accordingly.

andreaskurth:
> I guess we would have to continue NACK'ing on subsequent starts, and by that point completely drop them on the 
ground.

Yes, so it just shifts the problem from the first NACK to subsequent NACKs, and I think this makes things more complicated because the behavior of our I2C IP would differ once the ACQ FIFO is full.

> I still think it's useful to record the fact that we NACK'ed in the ACQ FIFO.

I agree that it's useful to record NACK'ing and inform SW, but I don't think we should use the ACQ FIFO for that, due to the problems above. Instead, how about we raise an IRQ (similar to the `nak` host-mode interrupt)? To record how many NACKs have happened, we could add a counter exposed in a CSR that HW increments on every NACK and SW can clear (probably on read to prevent access races). 

This would allow SW to handle 'the occasional NACK': it gets a NACK IRQ, the NACK counter is 1, and SW drains the ACQ FIFO until a data byte is followed by a Start condition or the ACQ FIFO is empty: that's the last data byte before the NACK and the corresponding transfer is incomplete. SW can reliably discard the incomplete transfer, and it may even be able to recover from this situation (depending on the higher-level protocol).

This would also allow SW to detect and flag 'repeated NACKs': it gets a NACK IRQ and the counter is greater than 1. In this case, SW not only knows that the last transfer in the ACQ FIFO is incomplete but also other transfers got dropped. The counter value can be useful during development: likely the maximum delay until the `acq_full` ISR needs to make space in the ACQ FIFO has to be reduced -- depending on the value in the counter only by a bit or substantially.

marnovandermaas:
It does create different behavior if the ACQ FIFO is full but I don't think that is necessarily a problem. I think it is useful to record the NACK inside the ACQ FIFO especially if the ACQ FIFO is quite deep. For example, if the software is notified that the ACQ FIFO is full, then it starts emptying the FIFO, what if it receives a start condition and its address before software reaches the end of the ACQ FIFO. The hardware would be acknowledging these bytes, but the software would have to disregard them because it does know where the NACK happened.

I could implement a NACK interrupt, but at the moment there is already an ACQ FIFO full interrupt, which is pretty much the same thing because we only NACK when the FIFO is full.

In terms of a NACK counter, I think that's a good idea even in my design especially with the clear-on-read property. Let me see if I can implement that.

marnovandermaas:
Ok, I've just pushed updates to this PR to stretch after an ACK again, and then NACK the next byte. I've also made sure at the end of a NACK it sends a command complete interrupt. This is on top of the ACQ FIFO full interrupt that will have fired when the stretching happened. Finally, I've implemented @andreaskurth suggestion of having a NACK counter that clears on a read so that software can tell how many NACKs happened in the meantime.

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -1166,8 +1195,14 @@ module i2c_fsm import i2c_pkg::*;
             if (acq_fifo_empty_or_last_space) begin
               // We must have stretched before, and software has been notified
               // through and ACQ FIFO full event. Now we must NACK
-              // unconditionally.
-              state_d = NackWait;
+              // unconditionally for reads and for writes we should NACK the
+              // next byte unconditionally.
```
andreaskurth: Suggest one sentence for the read case and one for the write case

hw/ip/i2c/rtl/i2c_pkg.sv
```diff
@@ -16,7 +16,22 @@ package i2c_pkg;
...
+                          //    byte is still in the data field for inspection.
+    AcqNackStart = 3'b101 // This means that we got a write request to our
+                          // address, we sent an ACK to back to the host so
+                          // that we can be compatible with SMBus, but now we
+                          // must unconditionally NACK the next byte. We
+                          // cannot record that NACK'ed byte because there is
+                          // no space in the ACQ FIFO. The OpenTitan software
+                          // must know this distinction from a normal AcqNack
+                          // because the state machine must still continue
+                          // through the AcquireByte and Nack* states.
```
andreaskurth, line 34: Suggest putting the comments above each of the two values rather than next to them, due to limited horizontal space

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -1317,6 +1357,7 @@ module i2c_fsm import i2c_pkg::*;
         if (tcount_q == 20'd1 && !scl_i) begin
           state_d = NackSetup;
         end
+        clear_nack_next_byte = 1'b1;
```
andreaskurth, line 1364: This can be removed without changing behavior, right? Because it will get cleared as soon as we're back in the Idle state. I'm asking because I find it confusing as this NACKing is supposed to be sticky.

hw/ip/i2c/rtl/i2c_core.sv
```diff
@@ -187,6 +188,11 @@ module i2c_core import i2c_pkg::*;
   assign hw2reg.target_fifo_status.acqlvl.d = MaxFifoDepthW'(acq_fifo_depth);
   assign hw2reg.acqdata.abyte.d = acq_fifo_rdata[7:0];
   assign hw2reg.acqdata.signal.d = acq_fifo_rdata[AcqFifoWidth-1:8];
+  // Add one to the target NACK count if this target has sent a NACK.
+  assign hw2reg.target_nack_count.de = 1'b1;
```
andreaskurth, line 192: Why not
```
.de = event_target_nack;
.d = .q + 1;
```
? That way the D enable is only high if there's an update from HW.

andreaskurth:
> It does create different behavior if the ACQ FIFO is full but I don't think that is necessarily a problem. I think it is useful to record the NACK inside the ACQ FIFO especially if the ACQ FIFO is quite deep. For example, if the software is notified that the ACQ FIFO is full, then it starts emptying the FIFO, what if it receives a start condition and its address before software reaches the end of the ACQ FIFO. The hardware would be acknowledging these bytes, but the software would have to disregard them because it does know where the NACK happened.

Ah, I think a Start wouldn't be problematic because it would have to be preceded by a Stop for the previous transfer to be complete. But SW wouldn't know whether a transfer ended in a NACK or not if the next transfer starts with a Repeated Start, right? OK, storing the NACK SGTM.

> I could implement a NACK interrupt, but at the moment there is already an ACQ FIFO full interrupt, which is pretty much the same thing because we only NACK when the FIFO is full.

SG

hw/ip/i2c/rtl/i2c_pkg.sv
```diff
@@ -5,11 +5,18 @@
...
+  // FIFO does not need to contain both an 8 bit data field and a 3 bit
+  // identifier. We should have the ACQ FIFO be 9 bits wide where the MSB
+  // indicates whether it is a data byte or a control byte. This way we can
+  // add more values to this enum without having to widen the ACQ FIFO width.
+  typedef enum logic [2:0] {
+    AcqData      = 3'b000,
+    AcqStart     = 3'b001,
+    AcqStop      = 3'b010,
+    AcqRestart   = 3'b011,
+    AcqNack      = 3'b100
```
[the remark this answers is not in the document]
andreaskurth: And can you please create an issue for the encoding enhancement and link it as `TODO(#<issue number>)`?

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -572,8 +572,8 @@ module i2c_fsm import i2c_pkg::*;
         host_idle_o = 1'b0;
         scl_d = 1'b0;
         if (bit_index == '0 && tcount_q == 20'd1) begin
-          rx_fifo_wdata_o = read_byte;  // transfer read data to rx_fifo
```
andreaskurth, line 575: I would prefer if the first four commits, all of which make no functional changes, were one commit and the details are in the commit message. Four non-functional commits is too much noise on the Git log IMO

hw/ip/i2c/doc/registers.md
```diff
@@ -3,33 +3,34 @@
...
+| i2c.[`TIMING4`](#timing4)                         | 0x4c     |        4 | Detailed I2C Timings (directly corresponding to table 10, in the I2C Specification). |
+| i2c.[`TIMEOUT_CTRL`](#timeout_ctrl)               | 0x50     |        4 | I2C clock stretching timeout control.                                                |
+| i2c.[`TARGET_ID`](#target_id)                     | 0x54     |        4 | I2C target address and mask pairs                                                    |
+| i2c.[`ACQDATA`](#acqdata)                         | 0x58     |        4 | I2C target acquired data                                                             |
+| i2c.[`TXDATA`](#txdata)                           | 0x5c     |        4 | I2C target transmit data                                                             |
+| i2c.[`HOST_TIMEOUT_CTRL`](#host_timeout_ctrl)     | 0x60     |        4 | I2C host clock generation timeout value (in units of input clock frequency)          |
+| i2c.[`TARGET_TIMEOUT_CTRL`](#target_timeout_ctrl) | 0x64     |        4 | I2C target internal stretching timeout control.                                      |
```
andreaskurth: Can this commit be squashed into the previous commit (with the commit messages merged), please?

hw/ip/i2c/doc/registers.md
```diff
@@ -492,30 +492,33 @@ I2C target address and mask pairs
...
+| 0x0     | NONE      | ABYTE contains ordinary data byte as received                                                                                                            |
+| 0x1     | START     | ABYTE contains the 8-bit I2C address (R/W in lsb)                                                                                                        |
+| 0x2     | STOP      | ABYTE contains junk                                                                                                                                      |
+| 0x3     | RESTART   | ABYTE contains junk, START with address will follow                                                                                                      |
+| 0x4     | NACK      | ABYTE contains either the address or data that was NACK'ed                                                                                               |
+| 0x5     | NACKSTART | ABYTE contains the I2C address which was ACK'ed, but the block will continue and NACK the next data byte that was received: this only happens for writes |
```
andreaskurth, line 520: Can this commit be squashed into the previous commit (with the commit messages merged), please?

hw/ip/i2c/doc/registers.md
```diff
@@ -3,34 +3,35 @@
...
+| i2c.[`TARGET_ID`](#target_id)                     | 0x54     |        4 | I2C target address and mask pairs                                                                  |
+| i2c.[`ACQDATA`](#acqdata)                         | 0x58     |        4 | I2C target acquired data                                                                           |
+| i2c.[`TXDATA`](#txdata)                           | 0x5c     |        4 | I2C target transmit data                                                                           |
+| i2c.[`HOST_TIMEOUT_CTRL`](#host_timeout_ctrl)     | 0x60     |        4 | I2C host clock generation timeout value (in units of input clock frequency)                        |
+| i2c.[`TARGET_TIMEOUT_CTRL`](#target_timeout_ctrl) | 0x64     |        4 | I2C target internal stretching timeout control.                                                    |
+| i2c.[`TARGET_NACK_COUNT`](#target_nack_count)     | 0x68     |        4 | Number of times the I2C target has NACK'ed a new transaction since the last read of this register. |
```
andreaskurth, line 34: Can this commit be squashed into the previous commit (with commit messages merged), please?

andreaskurth (commented):
Thanks for these changes, @marnovandermaas! This mostly LGTM, though I have a few suggestions and questions.

I appreciate that you split your changes into commits, but I think this is too fine-grained and commits are no longer atomic changes (e.g., when Hjson modifications are split from autogen runs). This may make sense when hundreds of files get autogen'd but at the IP level I think we should keep those in one commit.

hw/ip/i2c/rtl/i2c_core.sv
```diff
@@ -44,6 +44,10 @@ module i2c_core import i2c_pkg::*;
   // Maximum number of bits required to represent the level/depth of any FIFO.
   localparam int unsigned MaxFifoDepthW = 12;
 
+  // Width of each entry in the ACQ FIFO with enough space for an 8-bit data
+  // byte and an identifier defined by i2c_acq_byte_id_e.
+  localparam int unsigned AcqFifoWidth = 10;
```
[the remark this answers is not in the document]
marnovandermaas: I've done this now by adding a parameter to the package.

hw/ip/i2c/rtl/i2c_pkg.sv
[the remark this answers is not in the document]
marnovandermaas: Yes, I've opened issue https://github.com/lowRISC/opentitan/issues/22028
And I've also moved it to the other commit.

hw/ip/i2c/rtl/i2c_fsm.sv
marnovandermaas: Done.

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -355,6 +374,7 @@ module i2c_fsm import i2c_pkg::*;
   // This is used for acq_fifo_wready_o to send the ACQ FIFO full alert to
   // software.
   assign acq_fifo_plenty_space = acq_fifo_remainder > AcqFifoDepthWidth'(2);
+  assign acq_fifo_empty_or_last_space = acq_fifo_remainder <= AcqFifoDepthWidth'(1);
```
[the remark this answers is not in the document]
marnovandermaas: Yes, thanks for spotting this.

hw/ip/i2c/rtl/i2c_pkg.sv
marnovandermaas: Yes, that looks better.

hw/ip/i2c/rtl/i2c_core.sv
marnovandermaas: Good point! This should now be fixed.

hw/ip/i2c/rtl/i2c_fsm.sv
marnovandermaas: Ok, I've squashed the first four commits into one.

hw/ip/i2c/doc/registers.md
marnovandermaas: Yes :)

hw/ip/i2c/doc/registers.md
marnovandermaas: Yes :smile:

hw/ip/i2c/doc/registers.md
marnovandermaas: Yes :+1:

hw/ip/i2c/rtl/i2c_fsm.sv
marnovandermaas: Yes, the behavior should not change. Clearing in the Idle and AcquireStart states should be enough. I agree that it is confusing to clear it here as well.

marnovandermaas:
The only failure in CI is the USB device smoke test, which is not affected by this PR.

hw/ip/i2c/data/i2c.hjson
```diff
@@ -691,5 +691,17 @@
...
+    { name: "TARGET_NACK_COUNT"
+      desc: '''
+            Number of times the I2C target has NACK'ed a new transaction since the last read of this register.
+            Reading this register clears it.
+            This is useful because when the ACQ FIFO is full the software know that a NACK has occurred, but without this register would not know how many transactions it missed.
+            '''
+      swaccess: "rc"
+      hwaccess: "hrw"
+      fields: [
+        { bits: "31:0" }
```
andreaskurth, line 703: Isn't a 32-bit counter way too large? I think an 8-bit counter, which saturates when it reaches its max value, would suffice. That can be done in a follow-up PR, though.

andreaskurth (approved):
Thanks for this work and addressing all our feedback, @marnovandermaas. This PR LGTM now; the remaining points can be addressed in follow-up PRs.

PR #21859, opened 2024-03-06
[dv] spi-monitor collect_trans unused phase reference removed

antmarzam:
Spi-monitor collect_trans task was receiving the UVM_PHASE handle as a dead task parameter since it was doing nothing with it


rswarbrick (commented):
This is actually overriding `dv_base_monitor::collect_trans` (which does have the argument).

So I *think* we actually need to stay the same shape. :-(

That said, I'm pretty surprised that the `phase` argument is needed in any version of the class! Maybe we can rip it out throughout...

antmarzam:
Closing since this is handled in a separate PR

PR #21861, opened 2024-03-06
[uart] Correct timeout comment units

elliotb-lowrisc:
The timeout value and corresponding comment seem to be saying different things. Let's update the comment to reflext the value, as doing it the other way around results is many more test failures.


eshapira (approved):
LGTM

PR #21865, opened 2024-03-06
[cherry-pick] #21776 onto earlgrey_es_sival

timothytrippel:


timothytrippel:
Looks like there is an issue with the `//sw/device/tests/crypto:drbg_functest_fpga_cw310_sival_rom_ext` that is causing it to fail across many PRs to the `earlgrey_es_sival` branch, so that is not related to this PR.

moidx (approved):

PR #21866, opened 2024-03-06
[opentitantool] Update hyperdebug_teacup.json to match hardware

jesultra:
It seems that on the Teacup board IOB11 is routed to CN9.21 and IOB12 is routed to CN9.19, the opposite of how there were declared.  As far as I recall, the two signals were swapped at some point during Teacup design review, in order for HyperDebug I2C capabilities to match how ChromeOS intends to use the pins of OpenTitan.


timothytrippel (approved):
nice catch, thanks @jesultra !

PR #21867, opened 2024-03-06
[ci] Avoid fusesoc builds in CI

pamaury:
See individual commits. Essentially this adds a new switch to the workspace bitstream cache so that required but unavailable bitstream can use a configurable package. In CI, we can then used a newly created "fake" unavailable bitstream that errors out if used.

NOTE: I have left a few debugging statements on purpose to double-check with what happens in CI. Will remove them afterwards.

### DIFF
--- a/ci/scripts/run-fpga-tests.sh
+++ b/ci/scripts/run-fpga-tests.sh
@@ -28,7 +28,7 @@ fi
 mkdir -p "${BIT_CACHE_DIR}"
 cp -rt "${BIT_CACHE_DIR}" "${BIT_SRC_DIR}"/*
 
-export BITSTREAM="--offline --list ci_bitstreams"
+export BITSTREAM="--offline --list --unavailable=@//hw/bitstream/unavailable ci_bitstreams"
 
 # We will lose serial access when we reboot, but if tests fail we should reboot
 # in case we've crashed the UART handler on the CW310's SAM3U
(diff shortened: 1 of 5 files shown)


nbdd0121:
How about removing .core files for all non-bitstream CI tasks?

a-will:
Just so I understand the rationale, these changes are needed chiefly to enable a fast failure in CI on the FPGA test nodes?

pamaury:
> Just so I understand the rationale, these changes are needed chiefly to enable a fast failure in CI on the FPGA test nodes?
> 
> I guess technically, the FPGA builds would succeed there if given enough time, but they take long enough that the CI job would time out if bitstreams started building there.

Yes, when a fusesoc build happens by mistake, it will run until the timeout (60 or 90 depending on the job) which is a *huge* waste of FPGA time and we already have queues on the FPGA runners. That's why I think it would be much more preferable to fail fast. It also provides a much clearer error.

I guess we could remove the core files as suggested by @nbdd0121 but I find this a bit too hacky.

jwnrt:
I think this change is okay, though it is quite a bit of code for this failure.

How would you feel about adding our own `fusesoc` to the front of the `PATH` which gave a clear failure message? e.g.

```sh
#!/bin/sh

echo "ERROR: fusesoc should not be run in this CI job." >&2
exit 1
```

Is this less hacky?

rswarbrick:
@pamaury: I suspect this no longer matches the way the bitstream cache works, but I may be wrong? If it is still active, could it be rebased?

pamaury:
The problem potentially still exists but improvements in the CI mean that we are much less likely to trigger it.

PR #21869, opened 2024-03-06
[dv,otp_ctrl] Fix issues related to random resets

matutem:
- Add some missing calls to cfg.stop_transaction_generators
- Fix use of wait_to_issue_reset override


msfschaffner:
I took care of the lint issues due to tab characters.

msfschaffner (approved):

msfschaffner:
Another thing that messed with CI was that some of the OTP sequences were templated - this is fixed now as well.

Issue #20798, opened 2024-01-09
[sival,sensor_ctrl] Fix unexpected alert 5


matutem:
### Description

Sensor_ctrl is generating spurious alert at bit 5, which can cause failures in any test that examines SENSOR_CTRL RECOV_ALERT expecting specific bits to be on.

One such test is pwrmgr_random_sleep_all_wake_ups, which for sensor_ctrl wakeup the test sets ALERT_TRIGGER bit 0 to trigger a recoverable alert.

Upon wakeup it clears bit 0 for both ALERT_TRIGGER and RECOV_ALERT CSRs, but that doesn't clear the pwrmgr's WAKE_STATUS CSR because RECOV_ALERT bit 5 is still on.

Once this bit is also cleared WAKE_STATUS is finally clear and the test passes. We need clarification from the AST owner that this behavior is as expected.

In order to test the behavior of SENSOR_CTRL, or once this extra alert is not generated, comment out the code that clears bit 5 in //sw/device/tests/sim_dv/rstmgr_alert_info_test.c, indicated by a TODO referring to lowrisc/opentitan#20798.

msfschaffner:
We should probably add a masking mechanism for AST alerts to `sensor_ctrl` as a fall back. 
That way flaky alerts can be stopped from creating spurious wakeup requests and/or alert events.

matutem:
So a new 11 bit wide alert_en sort of CSR? I expect the logic should only disable recoverable events. It probably should also be locked by cfg_regwen.

msfschaffner:
We can discuss adding it on the OT side. Nevertheless, the issue should be root-caused on the AST side as well...

meisnere:
Let me rephrase it, the issue will be fixed on ast side on next TO.
I want to understand what is the suggested fix on this thread, aren't we double fixing the issue? ast &ot side?

msfschaffner:
Ok great that is what I wanted to know.

I think the proposal is to:
- fix the actual issue on the AST side
- make the OT side more generic so that any future issues that are similar could be masked with a generic masking mechanism that is unrelated to the specific nature of this issue.

Does that make sense?

meisnere:
It make sense if it doesn't require ast changes, meaning the described extra masking will be an ot feature while handling false alerts from ast (even though there shouldn't be any on next TO)

msfschaffner:
This as been addressed in open source via https://github.com/lowRISC/opentitan/pull/21870.

Issue #20678, opened 2023-12-19
[sensor_ctrl/ast] Sensor controls RECOV_ALERT bit 10 should be masked on ES


meisnere:
### Description

Sensor controls RECOV_ALERT bit 10 should be masked on ES as false alert might rise.

@msfschaffner @timothytrippel @moidx

msfschaffner:
CC @matutem

timothytrippel:
@benjbender will this change the AST calibration values written to flash / OTP?

msfschaffner:
CC @arnonsha

johngt:
@msfschaffner - this has not been captured in any discussions for a while - have added a section in Top Level. Will need to consider if Nuvoton needs to work on this.

andreaskurth:
Re Earlgrey-ES: The issue to track masking flaky alerts in SW is #22778.

Re Earlgrey-PROD: sensor_ctrl now has CSRs to mask individual alerts from AST. All alerts are enabled after reset and there's a section in OTP `OWNER_SW_CFG` to hold the post-ROM configuration (#23149). ROM will read these OTP values and configure sensor_ctrl accordingly (#23150).

I'm closing this issue as no work beyond that tracked in the mentioned issues is left AFAIK.

PR #21870, opened 2024-03-06
[sensor_ctrl] Add enable register for AST alerts

msfschaffner:
This allows us to mask flaky alerts outside of AST in sensor_ctrl, before the alerts are collated into a single fatal or recoverable alert channel where they could potentially mask other alerts.

This should help with issues like #20678 and #20798.


hw/top_earlgrey/ip/sensor_ctrl/data/sensor_ctrl.hjson
```diff
@@ -168,12 +168,37 @@ registers: [
       },
     },
 
+    { multireg: {
+        cname: "SENSOR_CTRL",
+        name: "ALERT_EN",
```
msfschaffner, line 173: @moidx @timothytrippel do you think it is ok to handle them as "opt in"?

Any ROM changes needed for this?

moidx: Hi @msfschaffner, I added https://github.com/lowRISC/opentitan/issues/21871 to track changes for the ROM.

a-will (approved):

hw/top_earlgrey/ip/sensor_ctrl/data/sensor_ctrl.hjson
```diff
@@ -168,12 +168,37 @@ registers: [
...
+        fields: [
+            { bits: "0",
+              name: "VAL",
+              desc: '''
+                1 - An alert event is enabled.
+                0 - An alert event is disabled.
+              ''',
+              resval: "0"
+            },
+        ],
```
andreaskurth, line 190: Would it make sense to use MuBi here (to protect the value written by SW and at rest in the register)? I noted that `FATAL_ALERT_EN` isn't MuBi, but the `ALERT_EN` register has even more impact

msfschaffner: I can make this a mubi, yes.

msfschaffner: Ok I have converted all of these to MuBi4 registers.

hw/top_earlgrey/ip/sensor_ctrl/rtl/sensor_ctrl.sv
```diff
@@ -190,8 +194,12 @@ module sensor_ctrl
   for (genvar i = 0; i < NumAlertEvents; i++) begin : gen_ast_alert_events
 
     // when there is a valid alert, set the alert state
-    assign recov_event[i] = event_vld[i] & ~reg2hw.fatal_alert_en[i];
-    assign fatal_event[i] = event_vld[i] & reg2hw.fatal_alert_en[i];
+    assign recov_event[i] = mubi4_test_true_strict(mubi4_t'(reg2hw.alert_en[i])) &&
+                            event_vld[i] &&
+                            !reg2hw.fatal_alert_en[i];
+    assign fatal_event[i] = mubi4_test_true_strict(mubi4_t'(reg2hw.alert_en[i])) &&
```
matutem, line 200: It seems it should be a software error to have an alert configured as fatal, yet have it disabled. I suppose this is subject to alert_en only for symmetry's sake.

hw/top_earlgrey/ip/sensor_ctrl/data/sensor_ctrl.hjson
```diff
@@ -168,12 +168,39 @@ registers: [
...
+        compact: "false",
+        fields: [
+            { bits: "3:0",
+              name: "VAL",
+              mubi: "true"
+              desc: '''
+                kMultiBitBool4True - An alert event is enabled.
+                kMultiBitBool4False - An alert event is disabled.
+              ''',
+              resval: "False"
```
matutem, line 190: Reset to True to prevent accidentally leaving them off? As in the tests you needed to modify, or more importantly, in firmware.

msfschaffner:
> We are missing some unit tests, and I suggest we reset these as True (enabled). I also wonder if we really want to mask fatal alerts with this, since that seems like a software misconfiguration.

Indeed, it would be good to add a negative test for this as well.

Regarding masking of fatal alerts, the thing is that AST is a bit special in this regard. I.e., the fatality is configurable, and since the alerts can be flaky (as we've seen in the past), there needs to be some way of masking these.

Regarding the default, I actually asked the ROM folks above but did not get feedback on that yet.
In general, all alerts in the system are opt in and have to be enabled (also at the alert handler), hence I defaulted them to false.
Another reason why this is in my view the preferred default is again flakyness of analog/mixed signal alerts: if an alert is flaky, it has a chance to pollute an alert channel (potentially until next reset) if the mask resets to enabled.

hw/top_earlgrey/ip/sensor_ctrl/rtl/sensor_ctrl.sv
msfschaffner: Yes this is for symmetry reasons, but also due to the fact that we do not want to risk a flaky analog alert polluting any of the two alert channels.

hw/top_earlgrey/ip/sensor_ctrl/data/sensor_ctrl.hjson
msfschaffner: I thought about that, but as explained below, I think this is the safer way due to the fact that these alerts can be flaky and we don't want a known bad alert pollute any of the two alert channels by default.

sw/device/tests/sensor_ctrl_alerts.c
```diff
@@ -100,6 +100,10 @@ static void check_alert_state(dif_toggle_t fatal) {
  *  Finally, check for correct capture of cause in alert handler.
  */
 static void test_event(uint32_t idx, dif_toggle_t fatal) {
+  // Enable the alert on the sensor_ctrl side
```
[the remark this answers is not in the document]
msfschaffner: agreed, it would be good to add such a negative test.

we can do that as part of the V signoff for this block (the block is only tested at the top-level anyway).

hw/top_earlgrey/ip/sensor_ctrl/rtl/sensor_ctrl.sv
```diff
@@ -286,8 +305,8 @@ module sensor_ctrl
   // be used to trigger wake from low power.
   // Fatal alerts are not used here because they do not ever ack, meaning
   // the originating event can never disappear.
-  assign async_wake = (|async_alert_event_p)  |
-                      (~&async_alert_event_n) |
+  assign async_wake = (|(async_alert_event_p & alert_en_buf))  |
+                      (~&(async_alert_event_n | ~alert_en_buf)) |
```
msfschaffner, line 309: I forgot to mask the wakeup request, where it is arguably just as important to mask out a flaky alert so that the system is not constantly woken up. due to the mubi -> single bit translation, I have used a sec_buf anchor further above.

a-will (approved):
LGTM

PR #21872, opened 2024-03-06
[adc_ctrl] Change collated IRQ to status type

### LINKED ISSUE #21832: [sysrst_ctrl, adc_ctrl] Convert collated edge based interrupt to level
### Description

Both `sysrst_ctrl` and `adc_ctrl` have collated interrupt registers, where all local causes are reported and then sent out via one interrupt signal.

This is not ideal from an API perspective, since multiple IRQ registers have to be cleared for the same cause, potentially leading to a race.

The recommendation is thus to convert the collated IRQs to status type for both IPs.

- [x] sysrst_ctrl #21934
- [x] adc_ctrl https://github.com/lowRISC/opentitan/pull/21872

msfschaffner:
This fixes the `adc_ctrl` portion of #21832

### DIFF
--- a/hw/ip/adc_ctrl/data/adc_ctrl.hjson
+++ b/hw/ip/adc_ctrl/data/adc_ctrl.hjson
@@ -45,8 +45,9 @@
     }
   ],
   interrupt_list: [
-    { name: "match_done",
-      desc: "ADC match or measurement event done",
+    { name: "match_pending",
+      desc: "ADC match or measurement event has occurred",
+      type: "status"
     }
   ],
   alert_list: [
--- a/hw/ip/adc_ctrl/doc/interfaces.md
+++ b/hw/ip/adc_ctrl/doc/interfaces.md
@@ -18,9 +18,9 @@ Referring to the [Comportable guideline for peripheral device functionality](htt
 
 ## Interrupts
 
-| Interrupt Name   | Type   | Description                         |
-|:-----------------|:-------|:------------------------------------|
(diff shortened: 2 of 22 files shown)


a-will (approved):

jettr:
Can we rename `match_done` to `match_pending`? I am also assuming that `match_pending` collates `adc_intr_status[7:0]` or does is also collate `adc_intr_status[9:8]` too?

a-will:
> I am also assuming that `match_pending` collates `adc_intr_status[7:0]` or does is also collate `adc_intr_status[9:8]` too?

It gangs _all_ adc_ctrl interrupts sources into one signal for the PLIC. (Aside: Grouping before sending to the PLIC is how I'd expect most peripherals to handle their interrupts, but we currently mirror almost all interrupt registers independently into the PLIC, with a correspondingly much larger PLIC as a result, haha).

msfschaffner:
Yes it collates all interrupts.

msfschaffner:
Changed the interrupt name and tweaked the modeling in DV in the last force-push.

msfschaffner:
I removed the patch for `sysrst_ctrl` since it needs some more restructuring to support this (similar to what had to be done for `adc_ctrl`). I will push that as a separate fix.

msfschaffner:
Merging since the CI failures are unrelated and will be fixed separately:
- SiVal non_ROM_EXT tests: //sw/device/tests:uart_tx_rx_test_fpga_cw310_sival
- DV smoke test (chip-level): chip_sw_spi_host_tx_rx

PR #21873, opened 2024-03-06
[usbdev] Usbdev block level smoke test improvements for V1

alees24:
This PR offers improvements to the testing performed by the block level smoke test for the USB device in two commits:

1. Ensure that SETUP, OUT and IN data packets are transmitted and checked at the receiver for a random endpoint, exercising both transmission and reception of packets at both the CSR side and the USB side. Crude initial implementation of full packet decoding to check the contents of IN data packets; further validation is required and in particular the calculation and checking of CRC16. Implement the removal of bit stuffed '0' bits. Introduce workaround for the sampling issues that arise from operating on a clock frequency that matches the bit rate; the DV should preferably be employing an oversampled clock and not stealing access to the USB device driver enable output; this would then permit testing of frequency and phase mismatches and operation using just the true USB signals, for example.

2. Employ a deque whilst constructing the bit stuffed output packet, which is simpler and more efficient, and addresses the out-of-bounds access in the previous implementation.

The pass rate is unaffected. It is still not 100% with either EDA tool because for some seeds each simulator finds itself stuck without simulation time advancing. It appears to be something to do with the structure of the test environment and certainly does not indicate a DUT failure.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -100,6 +100,7 @@ virtual task dut_init(string reset_kind = "HARD");
   if (do_usbdev_init) usbdev_init();
 endtask
 
+  // Construct and transmit a token packet to the USB device
   virtual task call_token_seq(input pid_type_e pid_type);
     `uvm_create_on(m_token_pkt, p_sequencer.usb20_sequencer_h)
     m_token_pkt.m_pkt_type = PktTypeToken;
@@ -111,6 +112,7 @@ endtask
     finish_item(m_token_pkt);
   endtask
 
+  // Construct and transmit a DATA packet to the USB device
   virtual task call_data_seq(input pkt_type_e pkt_type, input pid_type_e pid_type,
                              input bit rand_or_not, input bit [6:0] num_of_bytes);
     `uvm_create_on(m_data_pkt, p_sequencer.usb20_sequencer_h)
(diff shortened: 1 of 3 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -148,6 +150,85 @@ endtask
...
+
+    // Check buffer content against packet data
+    for (int unsigned idx = 0; idx < exp_byte_data.size(); idx += 4) begin
+      bit [31:0] act_data;
+      bit [31:0] exp_data;
+      // Expected value of this word
+      exp_data = {exp_byte_data[idx + 3], exp_byte_data[idx + 2],
+                  exp_byte_data[idx + 1], exp_byte_data[idx]};
+
+      mem_rd(.ptr(ral.buffer), .offset(offset), .data(act_data));
```
rswarbrick, line 188: Nit: I'd be tempted to change `offset` to something like `buf_offset` and then use `buf_offset + idx/4` here, avoiding needing the increment that we've currently got on line 192.

It should be equivalent, but might be slightly more obvious to a reader.

hw/ip/usbdev/dv/env/seq_lib/usbdev_smoke_vseq.sv
```diff
@@ -8,83 +8,112 @@ class usbdev_smoke_vseq extends usbdev_base_vseq;
...
-      finish_item(m_token_pkt);
-      if (pid_type == PidTypeInToken) begin
-        get_response(rsp_item);
-        $cast(rsp_itm, rsp_item);
-        get_response_from_device(rsp_itm, PidTypeData1);
-      end
+    // TODO: Presently the act of sending a data packet, destructively modifies it!
+    // Restore the data packet to its original state.
+    exp_data = {<<8{m_data_pkt.data}};
+    exp_data = {<<{exp_data}};
```
rswarbrick, line 43: This took me a while to understand! (which is why I had to come and chat to you in the office...)

I think the end result is that `exp_data` is `m_data_pkt.data`, but with each byte reversed in place. Would it make sense to say that explicitly in a comment to avoid the next reader getting as confused as me? :-)

Similarly for the repeat of this code on line 66. Thinking about it, maybe the best solution is to put this in a function? (This would give a nice place to hang the comment...)

rswarbrick (approved):
This looks dramatically cleaner, thanks. A couple of nitty questions, but I'm happy either way :-)

hw/ip/usbdev/dv/env/seq_lib/usbdev_smoke_vseq.sv
alees24: Moved to a vseq-local function with explanatory comments as an interim, and hoping for it to be deleted soon.

alees24:
CI failures unrelated and USBDEV tests within DV smoke test are passing.

PR #21881, opened 2024-03-07
[cryptotest] Fix AES-GCM schema and test vector parser

sadmansakib-git:
Fix 1:
Currently, there are no FAIL cases in the GCM encryption test vectors. However, for decryption test vectors, if a test vector lacks a Plaintext (PT) field, the result of that particular test vector will be marked as FAIL.

In the initial PR, we accidentally marked the expected result as PASS if
there is no Plaintext (PT) field in the test vector, regardless it's
operation mode.

Fix 2:
In this PR, we include the missing auth tag from the test vectors into
the schema and parser.

### DIFF
--- a/sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
+++ b/sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
@@ -17,9 +17,9 @@ from cryptotest_util import parse_rsp, str_to_byte_array
 
 def parse_testcases(args) -> None:
     raw_testcases = parse_rsp(args.src)
-    testcases = list()
+    test_cases = list()
     for test_case_id, test_vec in enumerate(raw_testcases):
-        testcase = {
+        test_case = {
             "test_case_id": test_case_id,
             "vendor": "nist",
             "mode": "gcm",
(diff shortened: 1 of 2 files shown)


RyanTorok:
Just for documentation, NIST has a good description of this behavior in [https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/gcmvs.pdf](this document).

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
```diff
@@ -30,19 +36,19 @@ def parse_testcases(args) -> None:
             "iv": str_to_byte_array(test_vec["IV"]),
             "ciphertext": str_to_byte_array(test_vec["CT"]),
             "plaintext": str_to_byte_array(test_vec["PT"]) if "PT" in test_vec else [],
-            "result": True if "PT" in test_vec else False,
+            "result": decrypt_fail_check(args.operation.lower(), test_vec),
```
RyanTorok, line 39: Nit: this could be one-lined: `"result": operation != "decrypt" or "PT" in test_vec`

sadmansakib-git: Done

RyanTorok (approved):
LGTM, thanks!

milesdai (approved):
The fix lgtm. Can we add the reasoning and a link to the document (and maybe cite a specific section number) in a comment next to this logic? Otherwise the code on its own looks pretty suspicious 😅

sadmansakib-git:
> The fix lgtm. Can we add the reasoning and a link to the document (and maybe cite a specific section number) in a comment next to this logic? Otherwise the code on its own looks pretty suspicious 😅

I agree! Reasonable suggestion.

Done!

PR #21886, opened 2024-03-07
[RV_DM] Add seq_names in Test plan

BasitMehmood:
Adding names in testplan as these two vseqs has been merged and not added to regression yet.


rswarbrick (approved):
Thanks!

PR #21889, opened 2024-03-07
[uart] Speed up uart_intr test

elliotb-lowrisc:
A combination of tweaks to reduce the maximum length of time the uart_intr test takes to run. These seem to reduce max runtime by about half. Improves fail rate slightly by reducing the chance of hitting the timeout set by `+test_timeout_ns`. Did not appear to worsen coverage when tested locally with 50 random tests each for baseline vs. change.

The first change simply limits `num_trans` to a lower maximum. The second change forbids slower baud rates (9600 baud) when performing a long test (`num_trans` > 2).
The third change tweaks some probabilities to reduce the time it takes to empty the RX FIFO when checking the `RxTimeout` interrupt, which could take a long time for large values of `timeout_ctrl.val`.

### DIFF
--- a/hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
+++ b/hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
@@ -7,13 +7,15 @@ class uart_intr_vseq extends uart_base_vseq;
   `uvm_object_utils(uart_intr_vseq)
 
   constraint num_trans_c {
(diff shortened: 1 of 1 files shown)


elliotb-lowrisc:
Tweaked for further reduction of RxTimeout max runtime. Combined with a rebase it gave me a 100% pass rate on a 200-wide local test.

rswarbrick (approved):

PR #21890, opened 2024-03-07
[dv,usbdev] Test plan additions for V1

alees24:
- Include some further stress tests for invalid traffic, clock frequency adjustment to match the USB host traffic, and retraction of IN packets (improved Isochronous support).
- Additional tests for aon/wake module. Previously excluded from usbdev block level DV because it lies outside of the usbdev IP block. It should be incorporated into the usbdev TB and DV environment.

### DIFF
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -323,12 +323,14 @@
     {
       name: host_lost
       desc: '''
-            Verify that if link is active but SOF was not received from host for 4.096 ms then host lost
-            will raise.
+            Verify that if link is active (ie. not completely idle) but SOF was not received from
+            host for 4.096 ms, the 'host lost' interrupt becomes asserted.
 
-            - Send SOF from host after 4.096 ms while  INTR_ENABLE [3] is set.
-            - Verify that the corresponding interrupt pin goes high once the host lost.
-            - It should come after every 1 ms.
+            - Note that there must be _some_ non-idle signaling on the USB during this time because
+              otherwise the link state will become suspended, and 'host lost' will not be signaled.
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
This looks good. Thanks for working on filling out this testplan. (And, gosh, interacting with the outside world is much harder than an internal block!)

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -815,6 +841,91 @@
...
     }
+    {
+      name: aon_wake_resume
+      desc: '''
+            Verify that the AON/Wake module is able to detect Resume Signaling and indicate that
+            via the USB device register interface.
+
+            - Connect the USB device to the USB host model.
+            - Send Suspend Signaling to the USB device.
+            - Verify that the 'host_lost' interrupt is raised indicating the lack of SOF signaling
```
a-will, line 852: Wouldn't this be the `link_suspend` interrupt, not `host_lost`? Am I miscalculating? `host_lost` was supposed to take 4 frames, while we suspend in 3, I thought (which brings the link inactive, zeroing the missed SOF count).

a-will (approved):
Thanks for the hard work! Just one question about a little bullet point.

hw/ip/usbdev/data/usbdev_testplan.hjson
a-will: In other words, `host_lost` only triggers when the link is actually active / switching, but we never find SOF packets.

a-will: The phrase about the associated cause also now doesn't match, since it still describes the `host_lost` interrupt. :)

Thanks for the alignments!

a-will (approved):

PR #21892, opened 2024-03-07
[uart,dv] Fix uart TX watermark interrupt test

elliotb-lowrisc:
Fix occasional failures of the `uart_intr` test by adding a cycle of delay when checking the `TxWatermark` interrupt with the lowest watermark level.

Also update the uart "Theory of Operation" documentation to reflect the switch to status-type interrupts for these watermark interrupts.

### DIFF
--- a/hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
+++ b/hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
@@ -56,10 +56,14 @@ class uart_intr_vseq extends uart_base_vseq;
         int watermark_bytes = get_watermark_bytes_by_level(level);
         if (!en_tx) return;
         // First byte is immediately popped from TX FIFO (for transmission) and watermark based upon
-        // TX FIFO level excluding in-tranmission byte. Add 1 to watermark_bytes here to give the
+        // TX FIFO level excluding in-transmission byte. Add 1 to watermark_bytes here to give the
         // number of bytes required to move over the watermark threshold.
         watermark_bytes += 1;
         drive_tx_bytes(.num_bytes(watermark_bytes - 1));
+        if (level == 0) begin
+          // Need a brief delay while the data byte is popped and the interrupt returns high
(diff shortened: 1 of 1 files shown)


hw/ip/uart/doc/theory_of_operation.md
```diff
@@ -169,15 +169,10 @@ and unexpected event interrupts.
...
-This is especially important in the tx_watermark case.
-When the TX FIFO is empty, it by default satisfies all the watermark conditions.
-In order for the interrupt to trigger then, it is required that software initiates a write burst that is greater than the programmed watermark value.
-
-For example, assume TX watermark is programmed to be less than 4 bytes, and software programs one byte at a time, waits for it to finish transmitting, before supplying the next byte.
-Under these conditions, the TX watermark interrupt will never trigger because the size of the FIFO never exceeds the watermark level.
-
+Note that the watermark interrupts are level-based status interrupts.
+They will stay asserted for as long as the FIFO levels are in violation of the configured level
+and cannot be cleared by writing to the status register.
```
rswarbrick, line 174: Nit: For markdown docs, we follow the style of one sentence per line. It's not beautiful in the source, but the end result is that you don't get enormous diffs from things getting reflowed.

There's some documentation / a "markdown style guide" at https://opentitan.org/book/doc/contributing/style_guides/markdown_usage_style.html (rendered from `doc/contributing/style_guides/markdown_usage_style.md`).

This is different for README files, which are intended to be readable without rendering. These get flowed at 80 characters (I think, but I can't immediately find a reference).

rswarbrick (commented):
This looks sensible, thanks.

If you're re-pushing for the theory_of_operation tweak, could you split it into two commits? That way, we can separate the documentation update from the change to the code. (Not really important here, but it's always nice to make things more granular!)

elliotb-lowrisc:
Removed documentation changes

rswarbrick (approved):
Merging without waiting for CI (the previous change with the documentation tweak passed earlier).

PR #21898, opened 2024-03-07
Make .core files pass FuseSoC 2 schema validator

olofk:
FuseSoC 2 introduced a schema for CAPI2 and a validator. These changes make all .core files syntactically compatible with FuesSoC 2

### DIFF
--- a/hw/ip/adc_ctrl/adc_ctrl.core
+++ b/hw/ip/adc_ctrl/adc_ctrl.core
@@ -28,18 +28,12 @@ filesets:
       # common waivers
       - lowrisc:lint:common
       - lowrisc:lint:comportable
-    files:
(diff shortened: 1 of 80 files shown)


olofk:
Looks like private CI is failing, but I presume I can't see the reason for that. Any pointers for what should be fixed?

rswarbrick (approved):
Thank you very much for the tidyup. (I'm not sure my sed skills are up to it :-D)

rswarbrick:
About the private CI: this is caused by a change yesterday that has messed up things with spi_host. I've just checked and this PR isn't causing any of the failures, so I've approved and merged it now. Thanks!

olofk:
Thanks! Btw, that wasn't sed. That was me manually going through 80 files :D

PR #21905, opened 2024-03-08
[usbdev,dv] correction required in smoke test

Saad22386:
This PR contains a little amendment to the USBdev smoke test. Before checking the integrity of 'in_transaction' we need to restore the data packet to its original state.


Saad22386:
@alees24 @rswarbrick can you look at it.

PR #21338, opened 2024-02-13
[spi_host,rtl] Use sd_en_o[0] to only drive sd_o[0] during TX (Standard speed mode)

hcallahan-lowrisc:
Change the behaviour as suggested/proposed in https://github.com/lowRISC/opentitan/issues/14969.

The fix suggested [here](https://github.com/lowRISC/opentitan/issues/14969#issuecomment-1501348019) does indeed seem to make logical sense, and after observing the waveforms I can see the expected masking behaviour.

TODO
- [x] ~~Broken block-level DV~~ (spi_host_speed, spi_host_spien)
<details>
  <summary>..failures appear to be just spurious, same as nightlies.</summary>

      There appears to be a sampling issue on the final bit of the final byte of a segment.
```
UVM_FATAL @ 373563549 ps: (spi_host_scoreboard.sv:155) uvm_test_top.env.scoreboard [uvm_test_top.env.scoreboard] 
	 WRITE: actual data did not match exp data 
|	 (len 4) 
|	 byte      actual     expected
|	 [0] 			 3 			 3
|	 [1] 			 37 			 37
|	 [2] 			 8a 			 8a
|	 [3] 			 cd 			 cc
```

</details>

Closes https://github.com/lowRISC/opentitan/issues/14969

PR #21906, opened 2024-03-08
[spi_host,rtl] Revert "[spi_host,rtl] Use sd_en_o[0] to only drive sd_o[0] during TX (std_mode)"

andreaskurth:
The original change cause the `chip_sw_spi_host_tx_rx` test to fail and thus causes failing CI checks on `master`.
For now, revert the commit.

A proper fix is being worked on.
See https://github.com/lowRISC/opentitan/pull/21338#issuecomment-1984908004 for details.

### DIFF
--- a/hw/ip/spi_host/rtl/spi_host_fsm.sv
+++ b/hw/ip/spi_host/rtl/spi_host_fsm.sv
@@ -564,7 +564,7 @@ module spi_host_fsm
     end else begin
       unique case (speed_o)
         Standard: begin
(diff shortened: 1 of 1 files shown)


hcallahan-lowrisc (approved):
Approving now, can be merged as soon as CI passes.

andreaskurth:
Merging as the failing `//sw/device/tests/crypto:kdf_kmac_functest_fpga_cw310_test_rom` test is likely not caused by this PR and this fixes a failing CI check on `master`.

PR #21907, opened 2024-03-08
[sival] Disable uart_tx_rx test in CI

jwnrt:
This test has had some flaky behaviour with the UART communication between the FPGA and HyperDebug. I haven't yet debugged the cause, so disabling until it's fixed.

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -5105,7 +5105,8 @@ opentitan_test(
         test_harness = "//sw/host/tests/chip/uart:uart_tx_rx",
     ),
     exec_env = {
-        "//hw/top_earlgrey:fpga_cw310_sival": None,
+        # FIXME has some flaky behaviour.
+        "//hw/top_earlgrey:fpga_cw310_sival": "broken",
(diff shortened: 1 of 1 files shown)


sw/device/tests/BUILD
```diff
@@ -5105,7 +5105,8 @@ opentitan_test(
         test_harness = "//sw/host/tests/chip/uart:uart_tx_rx",
     ),
     exec_env = {
```
pamaury, line 5107: You need to create a `broken = new_cw310_params(tags = ["broken"])` entry

jwnrt: It already exists for the _rom_ext exec env

pamaury: Ah sorry I missed it!

pamaury (approved):

andreaskurth:
This passed CI except for the `spi_host` DV smoke test, which is unrelated and got fixed separately. Merging.

andreaskurth:
Thanks for this CI stability improvement, @jwnrt. Could you please create an issue to track the debug and re-enablement of this test?

PR #21908, opened 2024-03-08
[aon_timer] Move to v2.0/D2/V1

GregAC:
Major version bump due to change in wake-up timer to become 64-bit.

### DIFF
--- a/hw/ip/aon_timer/data/aon_timer.hjson
+++ b/hw/ip/aon_timer/data/aon_timer.hjson
@@ -19,10 +19,10 @@
   dv_doc:             "../doc/dv",
   hw_checklist:       "../doc/checklist",
   sw_checklist:       "/sw/device/lib/dif/dif_aon_timer",
-  version:            "1.0.0",
+  version:            "2.0.0",
   life_stage:         "L1",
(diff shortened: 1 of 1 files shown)


msfschaffner (approved):

andreaskurth:
Merging as actions from the discussions above can be taken later (if necessary) and otherwise there seems to be consensus. The failing CI checks cannot be caused by this version bump.

hw/ip/aon_timer/data/aon_timer.hjson
```diff
@@ -19,10 +19,10 @@
   dv_doc:             "../doc/dv",
   hw_checklist:       "../doc/checklist",
   sw_checklist:       "/sw/device/lib/dif/dif_aon_timer",
-  version:            "1.0.0",
+  version:            "2.0.0",
   life_stage:         "L1",
-  design_stage:       "D3",
-  verification_stage: "V2S",
+  design_stage:       "D2",
```
msfschaffner, line 24: @GregAC is there any reason not to go to D2S directly? i.e., any outstanding S reviews?
Otherwise we could go directly to D2S.

PR #21909, opened 2024-03-08
[uart,doc] uart watermark documentation update

elliotb-lowrisc:
Update the uart "Theory of Operation" documentation to reflect the switch to status-type interrupts for these watermark interrupts.


rswarbrick (approved):
Thanks for this: it looks good.

Merging without waiting for CI (since it's just a documentation change).

PR #21913, opened 2024-03-08
[bazel] Fix bitstream splicing with skip configuration

### LINKED ISSUE #21668: [bazel] Bitstream skip does not work
### Description

It seems that  the switch to the "universal" splicer has broken bitstream skip because it wants to splice `skip.bit` which obviously is not possible. I was able to fix it using the following:
```diff
diff --git a/rules/opentitan/splice.bzl b/rules/opentitan/splice.bzl
index 5ee0c9ca77..6294eb777a 100644
--- a/rules/opentitan/splice.bzl
+++ b/rules/opentitan/splice.bzl
@@ -106,6 +106,9 @@ def _bitstream_splice_impl(ctx):
     exec_env = ctx.attr.exec_env[ExecEnvInfo]
     src = ctx.file.src if ctx.file.src else exec_env.base_bitstream
 
+    if src.path.endswith("skip.bit"):
+        return [DefaultInfo(files = depset([src]))]
+
     # Splice in a ROM image if we have one either in attrs or the exec_env.
     if not ctx.attr.rom or ctx.attr.rom.label.name == "none":
         rom = exec_env.rom
```
We should consider fixing this properly (maybe the above can be sufficient).

pamaury:
Fixes #21668, see issue for details

### DIFF
--- a/rules/opentitan/splice.bzl
+++ b/rules/opentitan/splice.bzl
@@ -106,6 +106,11 @@ def _bitstream_splice_impl(ctx):
     exec_env = ctx.attr.exec_env[ExecEnvInfo]
     src = ctx.file.src if ctx.file.src else exec_env.base_bitstream
 
+    # If configured to skip bitstream loading, do not splice and forward the file:
+    # opentitantool skips loading if the file is named `skip.bit`.
+    if ctx.attr.skip:
+        return [DefaultInfo(files = depset([src]))]
+
     # Splice in a ROM image if we have one either in attrs or the exec_env.
     if not ctx.attr.rom or ctx.attr.rom.label.name == "none":
         rom = exec_env.rom
@@ -159,7 +164,7 @@ def _bitstream_splice_impl(ctx):
     )
     return DefaultInfo(files = depset([output]))
 
-bitstream_splice = rule(
+bitstream_splice_ = rule(
     implementation = _bitstream_splice_impl,
     attrs = {
         "src": attr.label(allow_single_file = True, doc = "The bitstream to splice"),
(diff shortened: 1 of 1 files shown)


jwnrt (approved):

rules/opentitan/splice.bzl
```diff
@@ -106,6 +106,11 @@ def _bitstream_splice_impl(ctx):
     exec_env = ctx.attr.exec_env[ExecEnvInfo]
     src = ctx.file.src if ctx.file.src else exec_env.base_bitstream
 
+    # If configured to skip bitstream loading, do not splice and forward the file:
+    # opentitantool skips loading if the file is named `skip.bit`.
+    if src.path.endswith("skip.bit"):
```
a-will, line 111: Hm... Special file names feel a little awkward compared to using configuration from a build setting and provider.

pamaury: Yes I agree, I am still planning on cleaning that up. In the mean time, I am keeping the PR alive just in case some people need it (since skip is broken).

It would also be nice if we had a simple way of passing a specific bitstream to bazel so you can e.g. get a bitstream build from the CI and let bazel splice and use it.

pamaury: @a-will what do you think of this new implementation?

a-will: > It would also be nice if we had a simple way of passing a specific bitstream to bazel so you can e.g. get a bitstream build from the CI and let bazel splice and use it.

I'm not sure you're going to get "simple" with that particular function, hehe. However, it's possible to do now by creating a test bitstream cache, placing the desired bitstream output (the tar file with the manifest) in a subdirectory with some name `<test-bit-dir>`, pointing `BAZEL_BITSTREAMS_CACHE` to the bitstream cache directory, and modifying the `BITSTREAM` environment variable to contain something like "--offline `<test-bit-dir>`"

This is exactly what CI does, and it's how I test against bitstreams that haven't entered the cache (either built by me or CI).

pamaury: I know it is doable but this is not exactly "simple". I was thinking doing something like:
```bazel test --define bitstream_tgz=/path/to/bitstream.tar ...`
and it automatically takes care of extracting the bitstream and mmi files.

a-will (approved):
LGTM

At some point, it'd be nice to remove the funky `__skip__` magic value in opentitanlib and just remove the `--bitstream` argument when generating the test command script, plus convert this legacy bazel `bitstream` flag to a Starlark build attribute. Neither of those things needs to be done in this PR, though.

rules/opentitan/splice.bzl
pamaury: yes agreed

PR #21914, opened 2024-03-08
[dv,pwrmgr] Use reset_asserted instead of setting under_reset

matutem:
Calling reset_asserted propagates the reset to the csr package.


rswarbrick:
I'm not sure what `cdf` is here... Cumulative distribution function? :-)

More generally doesn't this do the same thing as it did before, setting the `reset_asserted` variable? (I assume not, but I don't understand it yet! Can you help me understand?)

matutem:
@rswarbrick the dv_base_env_cfg reset_asserted function sets under_reset, clears will_reset, and calls csr_utils_pkg::reset_asserted, so this change is necessary.

I am not sure what `cdf` you are referring to.

rswarbrick (commented):
Thanks for explaining. This all looks sensible to me.

(Note: the failed CI is definitely nothing to do with this change - feel free to merge it :-))

rswarbrick (approved):

PR #21915, opened 2024-03-08
[usbdev,dv] Minor cleanups to vseqs and call_data_seq task

rswarbrick:
The various commits explain the changes in detail. This is part of the work needed to move data out of the base class and into arguments, which should be much easier to reason about.


alees24 (approved):
LGTM, thanks.

PR #21916, opened 2024-03-08
[hw,prim_tlul_gate,rtl] Parametrize the number of outstanding transfers

Razer6:
If the sink port takes longer to respond, the internal counter for outstanding transfers would overflow. This PR adds the ability to parametrize the counter for the number of outstanding transactions

### DIFF
--- a/hw/ip/tlul/rtl/tlul_lc_gate.sv
+++ b/hw/ip/tlul/rtl/tlul_lc_gate.sv
@@ -23,7 +23,9 @@ module tlul_lc_gate
   // By default we return a TL-UL bus error response if the bus is gated. However, in some special
   // cases we need to be able to return valid, all-zero responses instead (e.g. for the RV_DM). In
   // those cases, ReturnBlankResp can be set to 1.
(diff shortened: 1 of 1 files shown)


a-will (approved):

msfschaffner (approved):

msfschaffner:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_lc_gate.sv

msfschaffner:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_lc_gate.sv

andreaskurth:
> If the sink port takes longer to respond, the internal counter for outstanding transfers would overflow.

This sounds like a bug to me; I would expect `prim_tlul_gate` to stop letting transactions through as long as the counter is saturated. If you agree, @msfschaffner, I'll open an issue to get this resolved.

msfschaffner:
Yes that sounds like something we need to take a closer look at.
Filing an issue for PROD.M3/M4 sounds good to me.

PR #21920, opened 2024-03-08
[dv] Fix more timeout comments with wrong units

elliotb-lowrisc:
A few more instances where the comment uses milliseconds but the value is in whole seconds.


rswarbrick (approved):
Looks good. Merging without waiting for CI because this is just a comment change.

PR #21922, opened 2024-03-08
[usbdev,dv] Tighten constraint on AON clock frequency

rswarbrick:
This fixes a rather confusing bug that I introduced last July with 23f4afe772b. My change there was reasonable, except that we were now only constraining aon_clk_freq_mhz after it was multiplied by 2 or 3.

Since these are 32-bit unsigned integers, weird things happened if the initial guess for the value was "2^31 + N". Then the constraints were:

 1.) aon_clk_freq_mhz * 3 >= usb_clk_freq_mhz

     This would check that 3 * (2^31 + N) was large enough. But that
     wraps to 2^31 + 3*N, which is pretty big.

 2.) aon_clk_freq_mhz * 2 <= usb_clk_freq_mhz;

     This would check that 2 * (2^31 + N) was small enough. But that
     wraps to 2*N which can be nice and small.

Oops!

Check that this sort of thing isn't happening by explicitly asking aon_clk_freq_mhz (not multiplied by anything) is less than usb_clk_freq_mhz.


alees24 (approved):
Thanks for investigating and fixing this. At last we have an apparent 100% pass rate for 'usbdev_smoke_test!'

PR #21923, opened 2024-03-08
[ci/cryptotest] Add rule to run crypto tests in nightly CI

RyanTorok:
Adds a rule to run the cryptotest framework tests in the nightly CI.

This is a draft PR because we currently have 4 unresolved test failures in ECDSA, whose test harness is already in `master`.


ci/azure-pipelines-nightly.yml
```diff
@@ -124,6 +124,28 @@ jobs:
     displayName: "Run the tests"
   - template: ../ci/publish-bazel-test-results.yml
 
+- job: cryptotest_cw310
+  displayName: "Cryptotest Framework CW310 Tests"
+  timeoutInMinutes: 300
```
RyanTorok, line 129: This is an estimate of what the timeout will be once all of the crypto tests are included. The algorithms that are currently merged at the time of this writing (ECDSA, SHA-2, SHA-3, SHAKE) should complete in < 90 minutes.

ci/azure-pipelines-nightly.yml
```diff
@@ -124,6 +124,28 @@ jobs:
...
+  steps:
+  - template: ../ci/checkout-template.yml
+  - template: ../ci/install-package-dependencies.yml
+  - bash: |
+      set -x
+      ci/bazelisk.sh test \
+        --define DISABLE_VERILATOR_BUILD=true \
+        --define bitstream=gcp_splice \
+        --build_tests_only \
+        --test_output=errors \
```
RyanTorok, line 143: I wrote `--test-output=errors` in order to not burden the CI with piping the output from the cryptotest framework (which there is a lot of) to the console, but we will need a way to communicate which tests failed, if any. Currently this is printed to the test log, but I'm not sure if there's a way to recover that file from CI. This does not affect the random seed printed for random tests.

RyanTorok: Edit: outdated; getting the desired test summary to print required some more in-depth handling of logging in each of the test harnesses.

ci/azure-pipelines-nightly.yml
RyanTorok: Edit: HMAC, DRBG, and ECDH are now included in master as well (see #21976)

elliotb-lowrisc:
Closing as no longer relevant (Azure)

PR #21925, opened 2024-03-08
[usbdev,dv] Remove num_of_bytes from base vseq

rswarbrick:
It turns out that we almost never need the sequence to have an opinion about num_of_bytes. The only time it matters is when we send an OUT packet to populate a buffer and then try to read the data back out again with an IN packet.

In those situations, we pick up the correct num_of_bytes value for configuring the IN packet by looking at the data packet that we just sent (see usbdev_pkt_sent_vseq for an example).


Saad22386:
@rswarbrick, can you proceed with this PR?

rswarbrick:
Only when someone reviews it! :-) @alees24, @hcallahan-lowrisc: would you mind taking a look?

alees24 (approved):
Thanks.

PR #21928, opened 2024-03-08
Out_iso_vseq

MubashirSaleem775:
[usbdev] out_iso_sequence

Adds usbdev_out_iso_sequence

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -159,10 +159,13 @@ endtask
 
   // Construct and transmit a DATA packet to the USB device
   virtual task call_data_seq(input pid_type_e pid_type,
-                             input bit randomize_length, input bit [6:0] num_of_bytes);
+                             input bit randomize_length, input bit [6:0] num_of_bytes,
+                             input bit isochronous_transfer = 1'b0);
     `uvm_create_on(m_data_pkt, p_sequencer.usb20_sequencer_h)
     m_data_pkt.m_pkt_type = PktTypeData;
     m_data_pkt.m_pid_type = pid_type;
+    if (isochronous_transfer)
+      m_data_pkt.m_usb_transfer = usb_transfer_e'(IsoTrans);
     `DV_CHECK_RANDOMIZE_WITH_FATAL(m_data_pkt,
                                    !randomize_length -> m_data_pkt.data.size() == num_of_bytes;)
     m_usb20_item = m_data_pkt;
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
@@ -1,0 +1,50 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class usbdev_out_iso_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_out_iso_vseq)
+
+  `uvm_object_new
+
+  task body();
+    // Expected data content of packet
+    byte unsigned exp_data[];
+    bit pkt_received;
+    configure_out_trans();
+    // ISO EP OUT
+    csr_wr(.ptr(ral.out_iso[0].iso[endp]), .value(1'b1));
+    // write to clear pkt_received interrupt. Verify this after transaction.
+    csr_wr(.ptr(ral.intr_state.pkt_received), .value(1'b1));
+    call_token_seq(PidTypeOutToken);
+    inter_packet_delay();
+    call_data_seq(PidTypeData0, .randomize_length(1'b0), .num_of_bytes(64),
+                  .isochronous_transfer(1'b1));
+    // Wait until usbdev generates an interrupt to show the packet has been received. This should
+    // happen without needing an ACK from the host because the device is configured for ISO traffic.
+    for (int i = 0; i < 10; i++) begin
+      csr_rd(.ptr(ral.intr_state.pkt_received), .value(pkt_received));
+      if (pkt_received) break;
+    end
+    if (!pkt_received) begin
+      `uvm_error(`gfn, "usbdev didn't generate expected pkt_received interrupt")
+    end
+    // verify that after OUT Iso device ports is on IDLE J statmeans not sending any ACK pkt.
+    check_device_status();
+    check_rx_packet(endp, 1'b0, out_buffer_id, m_data_pkt.data);
+  endtask
+
+  task check_device_status();
+    for (int i=0; i < 18; i++) begin
+      bit prev_usb_p;
+      bit prev_usb_n;
+      @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+      if (cfg.m_usb20_agent_cfg.bif.usb_p != prev_usb_p &&
+          cfg.m_usb20_agent_cfg.bif.usb_n != prev_usb_n) begin
+        `uvm_error(`gfn, "usbdev is not Idle") break;
+      end
+        prev_usb_p = cfg.m_usb20_agent_cfg.bif.usb_p;
+        prev_usb_p = cfg.m_usb20_agent_cfg.bif.usb_p;
+    end
+  endtask
+endclass
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_vseq_list.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_vseq_list.sv
@@ -16,6 +16,7 @@
 `include "usbdev_in_trans_vseq.sv"
 `include "usbdev_in_iso_vseq.sv"
 `include "usbdev_nak_trans_vseq.sv"
+`include "usbdev_out_iso_vseq.sv"
 `include "usbdev_out_stall_vseq.sv"
 `include "usbdev_out_trans_nak_vseq.sv"
 `include "usbdev_phy_config_usb_ref_disable_vseq.sv"
--- a/hw/ip/usbdev/dv/env/usbdev_env.core
+++ b/hw/ip/usbdev/dv/env/usbdev_env.core
@@ -42,6 +42,7 @@ filesets:
       - seq_lib/usbdev_stall_trans_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_min_length_out_transaction_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_max_length_out_transaction_vseq.sv: {is_include_file: true}
+      - seq_lib/usbdev_out_iso_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_out_stall_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_out_trans_nak_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_fifo_rst_vseq.sv: {is_include_file: true}
--- a/hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
+++ b/hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
@@ -85,6 +85,10 @@
       uvm_test_seq: usbdev_nak_trans_vseq
     }
     {
+      name: out_iso
+      uvm_test_seq: usbdev_out_iso_vseq
+    }
+    {
       name: usbdev_out_stall
       uvm_test_seq: usbdev_out_stall_vseq
     }


MubashirSaleem775:
@rswarbrick @alees24 please review this PR.

rswarbrick (changes requested):
I believe that this PR was just supposed to be for one of the vseqs. Please could you force-push to drop the other commits?

MubashirSaleem775:
> I believe that this PR was just supposed to be for one of the vseqs. Please could you force-push to drop the other commits?

Okay, I have removed spare commit. Please review it.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+class usbdev_out_iso_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_out_iso_vseq)
+
+  `uvm_object_new
+
+  task body();
+    // Expected data content of packet
+   // byte unsigned exp_data[];
+    bit pkt_received;
+    super.dut_init("HARD");
```
rswarbrick, line 14: I understand that this is required because of a more general testbench problem. Probably the best thing is to fix that first, but we'll want to drop this line before merging this PR.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+class usbdev_out_iso_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_out_iso_vseq)
+
+  `uvm_object_new
+
+  task body();
+    // Expected data content of packet
+   // byte unsigned exp_data[];
```
rswarbrick, line 12: Please remove the commented code here.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+    bit pkt_received;
+    super.dut_init("HARD");
+    cfg.clk_rst_vif.wait_clks(10);
```
rswarbrick, line 15: I doubt we need to wait here, especially once we can drop the `dut_init` above.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+    configure_out_trans();
+    cfg.clk_rst_vif.wait_clks(10);
```
rswarbrick, line 17: If this is waiting for a specific period, please could you pull the wait out into a named task? (Similarly for the waits below)

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+    csr_wr(.ptr(ral.out_iso[0].iso[endp]),  .value(1'b1)); // ISO EP OUT
+    call_token_seq(PidTypeOutToken);
+    cfg.clk_rst_vif.wait_clks(10);
+    call_data_seq(PidTypeData0, .randomize_length(1'b0), .num_of_bytes(64));
+    //call_data_seq(PktTypeData, PidTypeData0, 0, 64);
```
rswarbrick, line 22: Please delete the commented code.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+    cfg.clk_rst_vif.wait_clks(10);
+    csr_rd(.ptr(ral.intr_state.pkt_received), .value(pkt_received));
+    // For completion of OUT transaction no ACK will comes from device .
```
rswarbrick, line 25: How are we checking that no ACK comes from the device? I can't actually see that here, and it's kind of the point of this test! (Maybe it's being checked in the scoreboard? But it probably makes sense to leave a comment explaining what's going on)

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+    `uvm_create_on(m_data_pkt, p_sequencer.usb20_sequencer_h)
+    m_data_pkt.m_pkt_type = PktTypeData;
+    m_data_pkt.m_pid_type = pid_type;
+    m_data_pkt.m_usb_transfer = usb_transfer_e'(IsoTrans);  // Defines Transfer types
+    `DV_CHECK_RANDOMIZE_WITH_FATAL(m_data_pkt,
+                                   !randomize_length -> m_data_pkt.data.size() == num_of_bytes;)
+    m_usb20_item = m_data_pkt;
+    start_item(m_data_pkt);
+    finish_item(m_data_pkt);
+  endtask
```
rswarbrick, line 42: This doesn't feel like the right way to do this. I'd suggest you do one of the following approaches:
- Add an optional argument to the task in the base class that lets the caller specify an isochronous transfer.
- Split the task in the base class so that it calls a helper function to create `m_data_pkt`. Then override that helper function in this vseq. This will avoid us having to copy the rest of the task.

hw/ip/usbdev/dv/env/seq_lib/usbdev_vseq_list.sv
```diff
@@ -20,7 +20,7 @@
 `include "usbdev_pkt_sent_vseq.sv"
 `include "usbdev_random_length_out_transaction_vseq.sv"
 `include "usbdev_setup_trans_ignored_vseq.sv"
-
+`include "usbdev_out_iso_vseq.sv"
```
rswarbrick, line 23: The newline here was to separate "all the vseqs, in alphabetical order" from "stuff that extends usbdev_random_length_out_transaction_vseq".

I suggest moving this new vseq to come just before `usbdev_phy_pins_sense_vseq.sv`.

rswarbrick (changes requested):

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: Okay

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: no we don't need wait here after dut_init removal.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: ok

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: ok

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: okay

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: Okay i will try this.

hw/ip/usbdev/dv/env/seq_lib/usbdev_vseq_list.sv
MubashirSaleem775: ok

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
rswarbrick: I don't believe that this has been addressed yet.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,50 @@
...
+
+  `uvm_object_new
+
+  task body();
+    // Expected data content of packet
+    byte unsigned exp_data[];
+    bit pkt_received;
+    uvm_reg_data_t rxfifo;
+    configure_out_trans();
+    inter_packet_delay();
```
rswarbrick, line 16: Why do we need a delay here? I don't think we're following a packet?

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,50 @@
...
+    csr_wr(.ptr(ral.out_iso[0].iso[endp]),  .value(1'b1)); // ISO EP OUT
+    call_token_seq(PidTypeOutToken);
+    inter_packet_delay();
+    call_data_seq(PidTypeData0, .randomize_length(1'b0), .num_of_bytes(64));
+    cfg.clk_rst_vif.wait_clks(10); // Wait for register update after transaction
```
rswarbrick, line 21: 10 cycles is a bit of a magic number! :-) Don't we need to wait with a timeout for the interrupt (or change to `intr_state`)? Something like this, maybe? (untested)
```systemverilog
...
    inter_packet_delay();
    call_data_seq(PidTypeData0, .randomize_length(1'b0), .num_of_bytes(64));

    // Wait until usbdev generates an interrupt to show the packet has been received. This should
    // happen without needing an ACK from the host because the device is configured for ISO traffic.
      csr_rd(.ptr(ral.intr_state.pkt_received), .value(pkt_received));
      if (pkt_received) break;
    end
    if (!pkt_received) begin
      `uvm_error(`gfn, "usbdev didn't generate expected pkt_received interrupt")
    end

    // Check the contents of the packet buffer memory against the OUT packet that was sent.
    recover_orig_data(m_data_pkt.data, exp_data);
    check_rx_packet(endp, 1'b0, out_buffer_id, exp_data);
...
```

Note that this only works if `intr_state.pkt_received` was zero before we start. We probably need to ensure that with a write to clear the interrupt first.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,50 @@
...
+    // bit without waiting for an ACK handshake from the host. In other transfers,
+    // such as control transfers, the device will wait for the ACK from the host.
+    // Upon receiving the ACK, it will then assert the 'pkt_received' interrupt.
+    // DV_CHECK on pkt_received interrupt
+    `DV_CHECK_EQ(pkt_received, 1)
+    // Check the contents of the packet buffer memory against the OUT packet that was sent.
+    recover_orig_data(m_data_pkt.data, exp_data);
+    check_rx_packet(endp, 1'b0, out_buffer_id, exp_data);
+  endtask
+  function recover_orig_data(input byte unsigned in[], output byte unsigned out[]);
```
rswarbrick, line 33: Nit: Can you add a blank line between the functions/tasks?

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: yes, its added by mistake.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: its addressed now.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: okay, i have changed.

rswarbrick:
I've just force-pushed to the PR to resolve some rebasing conflicts. No other change.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -112,10 +112,13 @@ endtask
 
   // Construct and transmit a DATA packet to the USB device
   virtual task call_data_seq(input pid_type_e pid_type,
-                             input bit randomize_length, input bit [6:0] num_of_bytes);
+                             input bit randomize_length, input bit [6:0] num_of_bytes,
+                             input bit isochronous_transfer = 1'b0);
     `uvm_create_on(m_data_pkt, p_sequencer.usb20_sequencer_h)
     m_data_pkt.m_pkt_type = PktTypeData;
     m_data_pkt.m_pid_type = pid_type;
+    if(isochronous_transfer)
```
rswarbrick, line 120: Nit: Missing space after `if`.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+    // happen without needing an ACK from the host because the device is configured for ISO traffic.
+    for (int i = 0; i < 10; i++) begin
+      csr_rd(.ptr(ral.intr_state.pkt_received), .value(pkt_received));
+      if (pkt_received) break;
+    end
+    if (!pkt_received) begin
+      `uvm_error(`gfn, "usbdev didn't generate expected pkt_received interrupt")
+    end
+     // DV_CHECK on pkt_received interrupt
+    `DV_CHECK_EQ(pkt_received, 1)
```
rswarbrick, line 33: Thanks for taking my suggestion, but we're now checking the value of `pkt_recieved` in two places! Can probably drop the last two lines here.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+    // Check the contents of the packet buffer memory against the OUT packet that was sent.
+    recover_orig_data(m_data_pkt.data, exp_data);
+    check_rx_packet(endp, 1'b0, out_buffer_id, exp_data);
+  endtask
+
+  function recover_orig_data(input byte unsigned in[], output byte unsigned out[]);
```
rswarbrick, line 39: I think that @alees24 has done a bit of tidying up here and we don't need to do the shuffling any more. Please can you run the test locally? I think it should probably pass without this function.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
```diff
@@ -0,0 +1,43 @@
...
+  `uvm_object_new
+
+  task body();
+    // Expected data content of packet
+    byte unsigned exp_data[];
+    bit pkt_received;
+    configure_out_trans();
+    // ISO EP OUT
+    csr_wr(.ptr(ral.out_iso[0].iso[endp]), .value(1'b1));
+    // write to clear pkt_received interrupt.Verify this after transaction.
```
rswarbrick, line 17: Nit: Missing space after "."

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: Yes, i will remove these.

hw/ip/usbdev/dv/env/seq_lib/usbdev_out_iso_vseq.sv
MubashirSaleem775: Yes, now we don't need this.

hw/ip/usbdev/dv/env/usbdev_env.core
```diff
@@ -39,14 +39,15 @@ filesets:
...
+      - seq_lib/usbdev_out_iso_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_out_stall_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_out_trans_nak_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_fifo_rst_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_phy_pins_sense_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_in_stall_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_stall_priority_over_nak_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_in_iso_vseq.sv: {is_include_file: true}
-    file_type: systemVerilogSource
+          file_type: systemVerilogSource
```
rswarbrick, line 50: This change of indentation breaks things in CI (we've messed up the YAML format). Please fix.

MubashirSaleem775: oops. i am totally unaware of these errors as we don't have complete tool set. i have fixed this please review this again. Thanks

rswarbrick:
Thanks for this: the code looks really good. I've just re-read the testplan item, which ends with "Verify that the device does not send any handshake packet in response."

I guess we're not exactly checking that no handshake packet came out? If that's right, I think it's reasonable! Would it also make sense to wait a little bit (100 USB cycles or something) and then check that usbdev hasn't sent a handshake packet while we were waiting?

MubashirSaleem775:
> Thanks for this: the code looks really good. I've just re-read the testplan item, which ends with "Verify that the device does not send any handshake packet in response."
> 
> I guess we're not exactly checking that no handshake packet came out? If that's right, I think it's reasonable! Would it also make sense to wait a little bit (100 USB cycles or something) and then check that usbdev hasn't sent a handshake packet while we were waiting?

Now i have coded to verify that device does not send any handshake packet after receiving OUT ISO trans until timeout condition of usb device. can you please review this. Thanks

alees24:

PR #21934, opened 2024-03-08
[sysrst_ctrl] Rework IRQ sync mechanism and convert IRQ to status type

msfschaffner:
Reworks the IRQ mechanism in `sysrst_ctrl` and converts the collated interrupt to status type.
To that end, this moves the IRQ tracking into the bus domain in a similar way as done in `adc_ctrl` so that the IRQ can be created from existing IRQ status bits more readily.

Fixes https://github.com/lowRISC/opentitan/issues/21832


msfschaffner:
The design side should be ready now, but DV still needs alignments.
My goal is to patch everything up so that V2S requirements are still fulfilled after this patch (so that we do not have to regress the stages, since this block has already been signed off).

msfschaffner:
The block-level DV should work now - a few delays had to be adjusted due to different CDC latency.

hw/ip/sysrst_ctrl/dv/env/seq_lib/sysrst_ctrl_ultra_low_pwr_vseq.sv
```diff
@@ -172,7 +172,7 @@ class sysrst_ctrl_ultra_low_pwr_vseq extends sysrst_ctrl_base_vseq;
                                   enable_ulp, pwrb_cycles, get_pwrb_timer),
                         $sformatf("ac_cycles=%0d, get_ac_timer=%0d", ac_cycles, get_ac_timer),
                         $sformatf("lid_cycles=%0d, lid_timer=%0d", lid_cycles, get_lid_timer)},
-                 UVM_MEDIUM)
+                 UVM_LOW)
```
a-will, line 175: Is that a debug session leftover?

Picking exactly the right severity for messages is not something I pursue with strong interest; this is just a poke in case you see a need for change, haha.

hw/ip/sysrst_ctrl/rtl/sysrst_ctrl_intr.sv
```diff
@@ -0,0 +1,197 @@
...
+
+  // staging portion takes on the value of the incoming event match
+  // and clears when it is snapshot into request hold.
+  always_ff @(posedge clk_aon_i or negedge rst_aon_ni) begin
+     if (!rst_aon_ni) begin
+        aon_staging_reqs_q <= '0;
+     end else if (aon_ld_req && |aon_reqs) begin
+        aon_staging_reqs_q <= aon_reqs;
+     end else if (aon_ld_req) begin
+        aon_staging_reqs_q <= '0;
```
a-will, line 66: Is this not the same as...

```suggestion
     end else if (aon_ld_req) begin
        aon_staging_reqs_q <= aon_reqs;
```

hw/ip/sysrst_ctrl/rtl/sysrst_ctrl_intr.sv
```diff
@@ -0,0 +1,197 @@
...
+  // Holding reg after the CDC. Note that aon_req_hold_q does not change until the handshake has
+  // been completed, hence we can sample it safely upon a dst_ack pulse.
+  logic dst_ack_q;
+  logic [NumAonIntrEvents-1:0] req_hold_q;
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      req_hold_q <= '0;
+      dst_ack_q <= 1'b0;
+    end else begin
+      dst_ack_q <= dst_ack;
```
a-will, line 111: Random comment: This line surprised me at first, but then it hit me that we must convert back to a pulse to satisfy downstream logic.

I feel unsure if we have more encoding changes than are strictly necessary, but how to reduce it doesn't immediately come to mind. At least this works!

a-will (approved):
LGTM. Thanks!

a-will:
Oh, but actually, we must've missed something, since //sw/device/tests:sysrst_ctrl_in_irq_test_fpga_cw310_sival is failing in CI.

msfschaffner:
> Oh, but actually, we must've missed something, since //sw/device/tests:sysrst_ctrl_in_irq_test_fpga_cw310_sival is failing in CI.

Yeah I saw that one. Let me check what's wrong.

hw/ip/sysrst_ctrl/dv/env/seq_lib/sysrst_ctrl_ultra_low_pwr_vseq.sv
msfschaffner: ah - indeed, this diff is not needed. I'll revert this.

hw/ip/sysrst_ctrl/rtl/sysrst_ctrl_intr.sv
msfschaffner: looks like it, we can simplify that (the adc_ctrl does that as well)

hw/ip/sysrst_ctrl/rtl/sysrst_ctrl_intr.sv
msfschaffner: yeah unfortunately we need to sync over some sort of pulse to snapshot the IRQ bits and trigger events on the bus clock side (these are still event based)...

hw/ip/sysrst_ctrl/rtl/sysrst_ctrl_intr.sv
msfschaffner: I removed this and changed it in adc_ctrl as well

a-will:
I force-pushed software changes for sysrst_ctrl_in_irq_test. The IRQ handling and synchronization should be all patched up now.

sw/device/tests/sysrst_ctrl_in_irq_test.c
```diff
@@ -213,22 +210,42 @@ void sysrst_ctrl_key_combo_detect(dif_sysrst_ctrl_key_combo_t key_combo,
...
+  const uint32_t kPlicTarget = kTopEarlgreyPlicTargetIbex0;
+  dif_rv_plic_irq_id_t plic_irq_id = 0;
+  CHECK_DIF_OK(dif_rv_plic_irq_claim(&plic, kPlicTarget, &plic_irq_id));
+
+  top_earlgrey_plic_peripheral_t peripheral = (top_earlgrey_plic_peripheral_t)
+      top_earlgrey_plic_interrupt_for_peripheral[plic_irq_id];
+
+  switch (peripheral) {
+    case kTopEarlgreyPlicPeripheralUart0:
+      if (!ottf_console_flow_control_isr(exc_info)) {
```
a-will, line 222: I wonder how many tests have flow control enabled, yet don't properly handle the interrupt...

pamaury: That's a good catch, I probably modified this sysrst_ctrl test for sival and enabled control flow without realizing that it required the ISR... It's also kind of nasty that every test has to call the ottf console flow isr manually..

a-will: > It's also kind of nasty that every test has to call the ottf console flow isr manually..

Probably, we would want to have OTTF provide the top-level ISR, and it could insert handlers for its enabled features, then call into a hook provided by the downstream test. If I have missed something, that might risk interfering with how some tests might want to work, though.

msfschaffner:
> I force-pushed software changes for sysrst_ctrl_in_irq_test. The IRQ handling and synchronization should be all patched up now.

Awesome, thanks!

PR #21946, opened 2024-03-11
Disable ROM chip info stamping

pamaury:
**Context**

The ROM contains a `chip_info` structure that records the git hash at the build time. Unfortunately, this also prevents caching in CI and it provides little benefit in general since it does not record if the git hash is clean or not.

**Proposed change**

This PR disables bazel stamping by default so that the ROM chip_info will contain a **fixed** hash by default. The original behaviour can be restore by explicitley building with `--stamp`, e.g. `./bazelisk.sh --stamp ...`. The only instance that I can think of where we want this hash is probably for releases anyway.

### DIFF
--- a/rules/autogen.bzl
+++ b/rules/autogen.bzl
@@ -82,6 +82,12 @@ def _chip_info_src(ctx):
         stamp_files = [ctx.version_file]
         stamp_args.append("--ot_version_file")
         stamp_args.append(ctx.version_file.path)
+    else:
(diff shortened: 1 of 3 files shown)


jwnrt (approved):
Thanks, looks good

nbdd0121 (approved):

moidx (approved):
FYI @cfrantz for ROM release documentation.

timothytrippel (approved):

Issue #13316, opened 2022-06-22
[usbdev] Add some basic event counters


a-will:
Add some saturating counters to track events that might otherwise be invisible to software, such as...
- Number of NAKs due to missing AV buffer
- Number of NAKs due to full RX buffer
- Number of dropped packets with bad token / CRC
- Number of attempted IN packets that didn't get a handshake

These can be useful for detecting marginal cables / boards and performance issues.

It would also be nice to detect the number of IN requests that don't get a response when we _expect_ to keep up, but it's harder to distinguish between "Oops, we were too slow" and "There was no data right now, and that's fine."

a-will:
I am realizing that only the device side knows what happens to isochronous OUT transactions. Do we need endpoint-specific information or is the device-wide counter enough?

There is a similar situation for isochronous IN transactions, but this still has the problem of unknown intentions, as above. We don't currently require users to write to usbdev to indicate whether they intentionally want to NAK / drop an IN request.

msfschaffner:
CC @alees24

msfschaffner:
@alees24 @a-will how easy would it be to add this?
@jesultra is this something you'd consider a must for the production chip or not?

johngt:
@moidx / @jesultra - could you please confirm if this is a req for production chip.

a-will:
FYI, this was merely a nice-to-have idea when I wrote this issue. It'd be a new, late-breaking requirement if it escalated. :smile:

jesultra:
These are not strict requirements from ChromeOS. Nice to have for sure, but we can certainly use the production chips without this feature.

alees24:
Fixed by  #22118

PR #21950, opened 2024-03-11
[usbdev] Diagnostic and performance counters

alees24:
This ~~draft~~ PR proposes a specification for a set of diagnostic/performance monitoring event counters as discussed in #13316.

These would be purely passive monitors/counters that do not modify the behavior of the existing logic.

- Many of the counters may be enabled/disabled on a per-endpoint basis, given software more fine-grained control of the events that it wishes to monitor.
- All counters saturate at their maximum values and remain there until reset (IP block reset or software reset via bit 31 of the counter register itself).
- An explicit enable is provided for those event counters that cannot be enabled/disabled on a per-endpoint basis, where per-endpoint counting is not meaningful/useful.
- CRC5 errors and CRC16 errors are separated, because when there is a significant frequency delta between the host and the device, it is anticipated that Invalid PID and CRC5 errors may be observed. When the delta is smaller, longer packets will be required in order to demonstrate communication problems; CRC16 errors may be observed on long packets whilst Invalid PID and CRC5 errors are not observed.
- Counting of timed out IN transactions is performed, separately from host NAK responses to IN traffic.
- Counting of attempted OUT transactions with mismatched Data Toggles are counted; these are taken, according to the specification, to be retransmissions of earlier packets, but there may be causes other than a missed handshake response leading to a resend.
- The register fields have been specified to permit the endpoint set to be changed over time, whilst the counters are operating, allowing them to be switched among multiple endpoints/pipes.

~~This PR has been created to discuss whether the set of event counters is appropriate, and whether implementation should proceed.~~ These are not considered an essential feature for production Si, although verification/testing of frequency tracking is limited at this stage, not being open sourced.

### DIFF
--- a/hw/ip/usbdev/data/usbdev.hjson
+++ b/hw/ip/usbdev/data/usbdev.hjson
@@ -884,6 +884,8 @@
                   '''
           }
         ]
+        tags: [// 'rdy', 'pend' and 'sending' bits interact with each other.
+               "excl:CsrNonInitTests:CsrExclWrite"]
       }
     }
     { multireg: {
--- a/hw/ip/usbdev/data/usbdev.hjson
+++ b/hw/ip/usbdev/data/usbdev.hjson
@@ -1301,6 +1301,450 @@
         }
       ]
     }
+    { name: "count_ign_avsetup",
+      desc: "Count of SETUP packets ignored because there was no buffer available.",
+      resval: "0",
+      hwext: "true",
+      fields: [
+        {
+          bits: "15:0",
+          name: "count",
+          desc: '''
+                Number of SETUP packets ignored because there was no buffer in the Av SETUP FIFO.
+                '''
+          swaccess: "ro",
+          hwaccess: "hwo"
+        }
+        {
+          bits: "27:16"
+          name: "endpoints",
+          desc: "Set of endpoints for which this counter is enabled.",
+          swaccess: "rw",
+          hwaccess: "hrw",
+          hwqe: "true"
+        }
+        {
+          bits: "31",
+          name: "rst",
+          desc: "Write 1 to reset the counter.",
+          swaccess: "wo",
+          hwaccess: "hro",
+          hwqe: "true",
+          tags: [// Prevent interaction with the count field.
+                 "excl:CsrNonInitTests:CsrExclWrite"]
+        }
+      ]
+    }
+    { name: "count_drop_avout",
+      desc: "Count of OUT packets dropped/NAKed because there was no buffer available.",
+      resval: "0",
+      hwext: "true",
+      fields: [
+        {
+          bits: "15:0",
+          name: "count",
+          desc: '''
+                Number of OUT packets that could not be accepted because there was no buffer in the
+                Av OUT FIFO. Non-Isochronous OUT packets have been NAKed.
+                Isochronous OUT packets were ignored.
+                '''
+          swaccess: "ro",
+          hwaccess: "hwo"
+        }
+        {
+          bits: "27:16"
+          name: "endpoints",
+          desc: "Set of endpoints for which this counter is enabled.",
+          swaccess: "rw",
+          hwaccess: "hrw",
+          hwqe: "true",
+        }
+        {
+          bits: "31",
+          name: "rst",
+          desc: "Write 1 to reset the counter.",
+          swaccess: "wo",
+          hwaccess: "hro",
+          hwqe: "true",
+          tags: [// Prevent interaction with the count field.
+                 "excl:CsrNonInitTests:CsrExclWrite"]
+        }
+      ]
+    }
+    { name: "count_drop_rx",
+      desc: "Count of SETUP/OUT packets ignored, dropped or NAKed because the RX FIFO was full.",
+      resval: "0",
+      hwext: "true",
+      fields: [
+        {
+          bits: "15:0",
+          name: "count",
+          desc: '''
+                Number of SETUP/OUT packets ignored, dropped or NAKed because the RX FIFO was full.
+                SETUP packets have been ignored, Isochronous OUT packets have been dropped, and
+                non-Isochronous OUT packets have been NAKed.
+                '''
+          swaccess: "ro",
+          hwaccess: "hwo"
+        }
+        {
+          bits: "27:16"
+          name: "endpoints",
+          desc: "Set of endpoints for which this counter is enabled.",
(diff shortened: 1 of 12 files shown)


a-will:
The CSRs look pretty cool to me!

alees24:
Pushed to try to overlap CI run with development of a basic test sequence.

@a-will It occurs to me that we probably want explicit counters for the times that packets are dropped without a handshake response, in either direction, since that's likely to be the chief consequence of communications errors and the correct behavior when the receiver detects invalid packets. On the OUT side we can infer this to some extent from the existing bitstuff/crc/pid checks, but I'm thinking a 'count_timeout_in' for IN transactions that timeout in USBDEV would be a good idea.

a-will:
> @a-will It occurs to me that we probably want explicit counters for the times that packets are dropped without a handshake response, in either direction, since that's likely to be the chief consequence of communications errors and the correct behavior when the receiver detects invalid packets. On the OUT side we can infer this to some extent from the existing bitstuff/crc/pid checks, but I'm thinking a 'count_timeout_in' for IN transactions that timeout in USBDEV would be a good idea.

That makes sense to me too.

I'm kind of curious how these will play out. Is there significant chance of communication errors at full-speed signaling rates? ...but if we don't measure it, how will we know? :smile:

hw/ip/usbdev/data/usbdev.hjson
```diff
@@ -1286,6 +1286,346 @@
...
+          desc: "Write 1 to reset the counter.",
+          swaccess: "wo",
+          hwqe: "true",
+          tags: [// Prevent interaction with the count field.
+                 "excl:CsrNonInitTests:CsrExclWrite"]
+        }
+      ]
+    }
+    { name: "count_nodata_in0",
+      desc: "Count of IN transactions for which no packet data was available.",
```
mdhayter, line 1424: Do these two counters need a common enable and a companion counter of the number of USB frames that they are enabled for? Maybe you can extract that all by software (we have the frame count stamp somewhere already) but since these are mostly counters to make the monitoring more efficient and effective the combo with enable seems better.

alees24: I'm aiming to leave software to decide the policy of how the counters are used, and build only the mechanism in hardware. Much of the motive for having two 'nodata_in' registers is that we can separate Isochronous IN traffic from non-Isochronous, and the set of endpoints for each counter may be changed at any moment, eg. in response to frame interrupts.

Indeed there's questionable utility to the 'pid_invalid' and 'bitstuff' counters, given that those events could be counted in software.

In situations where heavy traffic/load is _not_ the cause of the events that we're trying to monitor (eg. dropped packets), then it seems reasonable to expect the interrupt servicing to be prompt.

alees24:
Counting of IN side timeouts is now included, as is counting of Data Toggle mismatches on the OUT side, because these are - as per specification - silently dropped and ACKnowledged, but the software visibility of such is limited without the diagnostics. We've seen this happen, as you may remember, after returning from Deep Sleep.

There's a basic vseq to test the counters, which I'm holding back for a separate PR because it has to work within the present limitations of the DV environment and has a number of 'TODO:' comments presently.

@a-will One question about the RTL changes, though. Do you think we're okay not having visibility of STALL responses to transactions on the IN and OUT side? It's something else that happens automatically, albeit under conditions when the software has specifically requested stalling. Of course, there's always the option of a logic analyzer on the USB as a fallback.

a-will:
> @a-will One question about the RTL changes, though. Do you think we're okay not having visibility of STALL responses to transactions on the IN and OUT side? It's something else that happens automatically, albeit under conditions when the software has specifically requested stalling. Of course, there's always the option of a logic analyzer on the USB as a fallback.

Since it's something that is supposed to happen under software's direction, I think we're okay not having that visibility. Software would direct usbdev to continue issuing STALL until a control transaction comes in to clear it, and I feel like the user won't particularly care about how many transactions were rejected.

This is recently-discovered info, but because of the state we're in with area limits (spawning an issue like https://github.com/lowRISC/opentitan/issues/22025 , for example), there might be pushback for everything under this feature umbrella to be admitted into the next release, though. Just an FYI!

@msfschaffner @moidx @andreaskurth for any input here.

alees24:
The additions are small, and I think quite important from a software perspective, but should they become a focus regarding area concerns we could pare it back by reducing the counter widths and/or dropping the counting of bit stuffing errors and invalid PIDs in hardware, if necessary. We would just need to rely upon software being sufficiently responsive to interrupts.

GregAC:
@a-will @alees24 FYI we are going to be seeking area savings for PROD. It would be useful to get a rough idea of what's getting added here in terms of total flops, numbers and widths of adders/incrementers etc. It may this has to be shelved for Prod or gated behind a parameter so we can include it in FPGA builds and other top-levels but it isn't in the Earl Grey ASIC.

alees24:
The counters and register configuration amount fo 'Width x 12 + 96' flops, with 12 'Width'-bit incrementers, so presently with Width == 16, that's 288 flops.

Narrowing the counters to 12- or even 8 bits would probably be fine, and I think there's some validity to making them available only in FPGA builds as far as USBDEV is concerned, in that the logic/USB are the same in FPGA as the the chip; same clock frequency, same timings.

My concern would be that the CPU/chip clocking scheme in the FPGA Is different from that of the chip and so CPU/software response times, particularly interrupt latencies and servicing the USBDEV are impacted. Also, electrically the two targets could be very different because the external transceiver/PHY circuitry is very different.

GregAC:
Great thanks for the info @alees24. That's roughly half of the flop we just ditched from Ibex getting rid of performance counters though the area from the surrounding logic may well be different.

@vogelpi @andreaskurth @moidx please note the above for the area discussions and let @alees24 know if these are suitable to be included or not.

andreaskurth:
Adding extra area at this point is indeed a concern. Two thoughts:
- If one can observe the external USB wires with a protocol analyzer, which of the counters contain information that cannot be extracted from measurements with the external analyzer?
- Instead of adding one counter _per event_, would it be possible to have one or two counters and mux events into them based on SW configuration? Similar to how one can mux events to counters in the [RISC-V hardware performance monitor (page 34 ff of the priv spec)](https://drive.google.com/file/d/1EMip5dZlnypTk7pt4WWUKmtjUKTOkBqh/view).

andreaskurth:
This PR has been superseded by #22118, so I think we can close it.

PR #21953, opened 2024-03-11
[otp_ctrl] Increase Hamming distance in OTP commands

### LINKED ISSUE #21948: [otp_ctrl] Use sparse encoding for `cmd_e` in `prim_otp_pkg.sv`
### Description

When doing the D2(S) review for `otp_ctrl` in #20983 @andreaskurth and I noticed that the `cmd_e` enum in 
`hw/ip/prim/rtl/prim_otp_pkg.sv` is currently not using a sparse encoding:
```sv
  typedef enum logic [CmdWidth-1:0] {
    Read     = 3'b000,
    Write    = 3'b001,
    // Raw commands ignore integrity
    ReadRaw  = 3'b010,
    WriteRaw = 3'b011,
    Init     = 3'b111
  } cmd_e;
```
This is non-ideal as it is flipping one bit now is theoretically sufficient for ignoring integrity bits. Depending on the underlying macro / wrapper implementation, a stuck-at fault could be lead to integrity being completely disabled for certain partitions.

We've discussed with @msfschaffner that for Earl Grey this isn't a problem due to how the underlying wrapper interprets the commands. But we've also concluded that this should still be fixed to not depend on the closed source implementation, e.g., used in other platforms.

msfschaffner:
This a security patch for closing out D2S for OTP_CTRL.

Fixes #21948

### DIFF
--- a/hw/ip/prim/rtl/prim_otp_pkg.sv
+++ b/hw/ip/prim/rtl/prim_otp_pkg.sv
@@ -6,16 +6,37 @@
 
 package prim_otp_pkg;
 
-  parameter int CmdWidth = 3;
+  // The command is sparsely encoded to make it more difficult to tamper with.
+  parameter int CmdWidth = 7;
   parameter int ErrWidth = 3;
 
+  // Encoding generated with:
+  // $ ./util/design/sparse-fsm-encode.py -d 4 -m 5 -n 7 \
+  //     -s 696743973 --language=sv
+  //
+  // Hamming distance histogram:
+  //
+  //  0: --
+  //  1: --
+  //  2: --
+  //  3: --
+  //  4: |||||||||||||||||||| (100.00%)
+  //  5: --
+  //  6: --
+  //  7: --
+  //
+  // Minimum Hamming distance: 4
+  // Maximum Hamming distance: 4
+  // Minimum Hamming weight: 3
+  // Maximum Hamming weight: 5
+  //
   typedef enum logic [CmdWidth-1:0] {
-    Read     = 3'b000,
-    Write    = 3'b001,
+    Read     = 7'b1000101,
+    Write    = 7'b0110111,
     // Raw commands ignore integrity
-    ReadRaw  = 3'b010,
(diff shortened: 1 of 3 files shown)


msfschaffner:
@arnonsha @meisnere FYI

This is a security patch that touches the primitive wrapper interface.
We will have to update the encoding in the closed source wrapper once you ingest a newer M2 release candidate than RC0.

hw/ip/prim/rtl/prim_otp_pkg.sv
```diff
@@ -6,16 +6,35 @@
 
 package prim_otp_pkg;
 
-  parameter int CmdWidth = 3;
+  parameter int CmdWidth = 6;
   parameter int ErrWidth = 3;
 
+  // Encoding generated with:
+  // $ ./util/design/sparse-fsm-encode.py -d 3 -m 5 -n 6 \
```
matutem, line 13: I thought we typically aimed for higher hamming distances?

msfschaffner: For FSM encodings, yes, we would typically do 4 or 5, but at least 3.
I am not sure this is really required in this setting for the command.
We can do 4, but the command needs to be 7bit wide then for this to work.

msfschaffner: I increased the HD to 4

andreaskurth (approved):
LGTM, thanks!

vogelpi (approved):
LGTM, thanks @msfschaffner !

In case you need to push once more, there is also a comment in `hw/ip/prim_generic/rtl/prim_generic_otp.sv` mentioning the now outdated 3-bit commands.

For refernce, commit dc9da971498948e75213089c49b7b3d8b45d9567 added the raw commands ignoring the integrity.

msfschaffner:
> In case you need to push once more, there is also a comment in `hw/ip/prim_generic/rtl/prim_generic_otp.sv` mentioning the now outdated 3-bit commands.

I patched that up, thanks!

Issue #7895, opened 2021-08-23
[mask_rom] Decide retention SRAM scrambling strategy.


mundaym:
Currently we aren't scrambling the retention SRAM because it will result in the retention SRAM being effectively wiped when the device experiences a non-low power reset. More detailed discussion is [here](https://github.com/lowRISC/opentitan/pull/7220#issuecomment-873117531).

We should decide whether we want to scramble the retention SRAM or not, and if not what, if any, mitigations are required.

@moidx (in [comment](https://github.com/lowRISC/opentitan/pull/7885/commits/bd0b5f7ec2f7654b07c1c4e6332618bae92714de#r694327657)):
If we leave retention ram unscrambled, we should add a warning to the c struct, and also add a checksum or digest to protect the integrity as an attacker may be able to manipulate the data to abuse boot services. Or perhaps the multibit encoding of boot_services is sufficient.

mundaym:
(cc @tjaychen @cfrantz who originally discussed this)

mundaym:
If we do decide to scramble retention SRAM then we can use the code from this [PR](https://github.com/lowRISC/opentitan/pull/7182).

alphan:
@tjaychen closing this since IIRC we agreed that using the ephemeral key is what we want since otherwise the scrambling key would be gone after reset.

tjaychen:
yeah that's right. But just to clarify, we're going to be using the "fixed" scrambling key for FI protection since that will be preserved across reset domains.

msfschaffner:
indeed, I was also confused by the word `ephemeral`. the default key is a fixed (but secret) netlist constant.

alphan:
Yeah, my bad, the "fixed" scrambling key, (edited my comment above) apologies for the confusion.

PR #21960, opened 2024-03-11
[rom_ext] Initialize ownership 

cfrantz:
1. Initialize ownership during ROM_EXT init.
2. Process the OwnershipUnlock message in boot services.
3. Update the serial rescue implementation to encode and send the
   ownership unlock command.
4. Add a top-level `ownership` command for creating the ownership unlock
   command.
5. Add a rescue command to send the ownership unlock command.
6. Add a test for the OwnershipUnlock command.

### DIFF
--- a/sw/device/silicon_creator/rom_ext/rom_ext.c
+++ b/sw/device/silicon_creator/rom_ext/rom_ext.c
@@ -37,6 +37,8 @@
 #include "sw/device/silicon_creator/lib/manifest.h"
 #include "sw/device/silicon_creator/lib/manifest_def.h"
 #include "sw/device/silicon_creator/lib/otbn_boot_services.h"
+#include "sw/device/silicon_creator/lib/ownership/ownership.h"
+#include "sw/device/silicon_creator/lib/ownership/ownership_unlock.h"
 #include "sw/device/silicon_creator/lib/shutdown.h"
 #include "sw/device/silicon_creator/lib/sigverify/sigverify.h"
 #include "sw/device/silicon_creator/rom_ext/rescue.h"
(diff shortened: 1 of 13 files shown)


sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -541,47 +545,49 @@ static rom_error_t boot_svc_min_sec_ver_handler(boot_svc_msg_t *boot_svc_msg,
...
+        return ownership_unlock_handler(boot_svc_msg, boot_data);
       case kBootSvcNextBl0SlotResType:
+      case kBootSvcPrimaryBl0SlotResType:
+      case kBootSvcMinBl0SecVerResType:
+      case kBootSvcOwnershipUnlockResType:
         // For response messages left in ret-ram we do nothing.
         break;
       default:
-        HARDENED_TRAP();
+        dbg_printf("Unknown boot_svc: %x\r\n", msg_type);
```
moidx, line 577: 1) should we return an error here? Or are we confident that an attacker will not be able to leverage this fall-through condition.

2) An attacker may be able to use this print statement to characterize faults to the SRAM holding the boot service command (e.g. [fault parameters] == [observed %x behavior]). The gained information then could then later be used to mount an attack at runtime. @johannheyszl, is this something we should be concerned about, or do you think the SRAM re-scrambling will help to mitigate concerns here.

A less attacker friendly message would be `"Unknown boot_srv message type"` at the cost of usability.

moidx (approved):

sw/device/silicon_creator/rom_ext/rom_ext.c
johannheyszl: Thx @moidx - I think you are right it would be better not output the value.

AFAIK, the retention SRAM is scrambled using a default, global netlist key (#7895 ), so there is no rescrambling that would help.

cfrantz: Agreed: we shouldn't print anything for this case.  Removed.

PR #21967, opened 2024-03-12
[RV_DM]remove progbuf_exception testpoint entry

BasitMehmood:
This commit is to remove rv_dm_progbuf_exception testpoint entry as this functionality is already covered in rv_dm_cmderr_exception vseq. So this testpoint is no longer needed.

### DIFF
--- a/hw/ip/rv_dm/data/rv_dm_testplan.hjson
+++ b/hw/ip/rv_dm/data/rv_dm_testplan.hjson
@@ -525,7 +525,7 @@
             - Verify that jtag_dmi_ral.abstractcs.cmderr is set to 3.
             '''
       stage: V1
-      tests: []
+      tests: ["rv_dm_cmderr_exception"]
     }
     {
       name: rom_read_access
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -157,7 +157,7 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     uvm_reg_data_t rdata;
     repeat ($urandom_range(1, 10)) begin
       data = $urandom;
-      csr_wr(.ptr(tl_mem_ral.halted), .value(0));
+      request_halt();
       cfg.clk_rst_vif.wait_clks($urandom_range(0, 1000));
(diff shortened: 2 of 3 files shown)


BasitMehmood:
@rswarbrick

rswarbrick:
Sorry for giving rather bad advice when writing on a different PR yesterday. I guess the point is that the check is still needed, but is fortunately being implemented by `rv_dm_cmderr_vseq`! So I think the correct change is just to change the `tests` entry to be `["rv_dm_cmderr_exception"]`. Would you mind re-jigging this PR to match?

BasitMehmood:
> Sorry for giving rather bad advice when writing on a different PR yesterday. I guess the point is that the check is still needed, but is fortunately being implemented by `rv_dm_cmderr_vseq`! So I think the correct change is just to change the `tests` entry to be `["rv_dm_cmderr_exception"]`. Would you mind re-jigging this PR to match?

You mean change this test entry name to rv_dm_cmderr_exception rather than excluding this vseq?

rswarbrick:
Yes, that's right. We'll end up with two different testpoint entries which both point at the `rv_dm_cmderr_exception` test.

BasitMehmood:
> Yes, that's right. We'll end up with two different testpoint entries which both point at the `rv_dm_cmderr_exception` test.

Ok I have changed the name of test entry to cmderr_exception.
Another thing I want to mention that i don't know why abstractcmd_status vseq is failing in regression?

rswarbrick:
Oh, sorry, this isn't what I mean. I mean that the change should just be to amend the existing testpoint and change the empty list that's currently there to be `["rv_dm_cmderr_exception"]` instead.

For the failing test in the nightly regression: this isn't the right place to discuss it. I've just diagnosed what's going wrong and will push a PR to fix the rather silly code in a sec.

rswarbrick:
Failing nightly will be fixed by #22016.

rswarbrick (approved):
Thanks, this makes sense. Changing the testpoint name isn't quite correct, but it's a silly nit: I'll force-push with a fix in a second.

For next time, this is doing three things:
- Explicitly adding a test to a testpoint
- Switching some `csr_wr` calls to call a specialised `request_halt` task
- Using a named enum value in a vseq

These are all sensible changes! But they should probably be in separate commits next time: there is no particular reason for them to be merged together atomically.

Issue #21936, opened 2024-03-08
[cryptolib] support for EC key generation from provided DRBG


vsukhoml:
### Description

There is no way to implement "deterministic" key generation from DRBG. This is a common use case for TPM2 when object is recreated from provided data. This is usually done by seeding DRBG with some entropy stored on device and additional data from key public info, and then withdrawing random candidates from DRBG.
`otcrypto_ecdsa_keygen`, `otcrypto_x25519_keygen`,  `otcrypto_ecdh_keygen`, ` otcrypto_ed25519_keygen`

@moidx, @cfrantz, @jadephilipoom

jadephilipoom:
I think the solution here depends on the exact use-case. Do you need:
(a) some pathway where you can input the provided data to keygen, so you can get a key that is always the same for the same data and different for different data, or
(b) specifically a pathway where you can do this by seeding a DRBG, because you need the output key to be the exact value you would get from seeding the DRBG and then running keygen and no other bitstring will do?

If (a), then I recommend using `otcrypto_hw_backed` to generate a keymgr handle for the key based on your provided data. Then all the ECC keygen operations will give you a keypair that is derived from this diversification data (specifically, passing it through KMAC-KDF as the diversification input with the keymgr state as the key).

If (b), it's a far more controversial discussion, because exposing an interface like that would leave a major footgun in the API (the caller could improperly seed the DRBG) and also substantially weaken the SCA picture. Right now OTBN pulls randomness directly from its own CSRNG instance, and Ibex cannot interfere with that pathway. If we were to include a route from Ibex's CSRNG instance, the only one that can be manually/deterministically instantiated, then we'd have to pass the seed from CSRNG to Ibex and then from Ibex to OTBN, which involves a lot more secret data passing over the bus. And if it has to be a super specific bitstring that corresponds, for example, to the first 32 bytes that come from DRBG, then we would be forced to handle those bits unmasked at least briefly when they are produced (as opposed to pulling enough bytes for two shares and assuming that the secret value is a combination of the two that we never handle directly).

vsukhoml:
I think option (a) describes what I need better, but devil is in details. What is missing in documentation @ https://opentitan.org/book/doc/security/cryptolib/cryptolib_api.html is the example of use.

[...]

PR #19714, opened 2023-09-21
[crypto] Change the way hardware-backed keys are represented.

jadephilipoom:
Move hardware-backed key diversification data into the keyblob and add a function to translate it.

Previously, there were separate fields in key structs for key data (for non-keymgr keys) and the keymgr diversification data (for keymgr-generated keys). One or the other was always empty. This PR moves the diversification data into the buffer that is otherwise used for key material, which simplifies the representation and removes boilerplate empty buffers.

Part of #19549

PR #21975, opened 2024-03-12
[crypto] Fix outdated cryptolib docstrings.

jadephilipoom:
These comments refer to the old method of representing hardware-backed keys (which used a field in the key configuration instead of the keyblob). This was changed in https://github.com/lowRISC/opentitan/pull/19714

Instead, adds guidance that directs callers to `otcrypto_hw_backed_key` for initializing these keys.

Thanks to @vsukhoml for finding this mistake! See https://github.com/lowRISC/opentitan/issues/21936#issuecomment-1989531303


cfrantz (approved):

vsukhoml:
lgtm! thanks for fixing.

moidx (approved):

Issue #21111, opened 2024-01-30
[spi_device/dv] Update scoreboard to track dynamically-updated status registers


a-will:
### Description

This is a DV companion issue to #20754. When the SPI domain begins updating status registers at every byte beat (8th posedge of SPI clock), the Read Status command will no longer return a static value. Consequently, validation of Read Status commands will need to be updated to support tracking changes that occur at every byte beat.

The design PR will initially disable this section of the code: https://github.com/lowRISC/opentitan/blob/efc45b71517e6a504f4ca86478b643118bcd3797/hw/ip/spi_device/dv/env/spi_device_scoreboard.sv#L587-L593

In addition, the core clock (sys_clk) domain will continue to update its view only at CSB de-assertion time. As a result, the scoreboard's current strategy of using `csr_rd()` to determine the resolved value (of software and hardware updates at commit time) will no longer work. `latch_flash_status()` currently represents what happens to _both_ the SPI domain source-of-truth and the sys_clk domain's view, but the update rate for these two domains is no longer the same.

Note that the passthrough gate's "view" of the BUSY bit also continues to only update at CSB de-assertion time.

Here's a list of timings to keep in mind:
- (byte beat) SPI domain commits a newly-resolved flash status register value.
- (byte beat) The latest committed value of the flash status register is prepared for the next response byte for any ongoing Read Status command.
- (CSB posedge) The current value of the BUSY bit becomes the committed value for the passthrough gate enable.
- (CSB posedge) Delayed by a synchronizer, the core clock domain picks up the current value of the flash status register.

antmarzam:
This ticket can now be closed since the flash_status TB changes have now been merged in master in PR #22763

PR #21977, opened 2024-03-12
Flash status synchronisation tb fix

antmarzam:
TB has been updated to check for flash_status bits correctness.

SPI-monitor now triggers SV events on each byte sampled on the SPI bus as well as when the CSB is deasserted.

The intercept VSEQ has been updated so reads/writes to flash_status won't occur just sequentially before/after a SPI-txn but also concurrently, as well as sending two flash_status writes in a row (less likely) where the first  writes the upper bits of the flash_status register, and the second would clear the busy/wel bits.

The Scoreboard has been updated to be in sync with the latest  flash_status behaviour changes in the RTL:
- SPI reads to FLASH_STATUS (READ_STATUS#1/2/3 commands ) no longer wrap around the register if the CSB was left "active" after the first byte is returned to the host by the RTL
- Writes to flash_status make their way towards the CDC SPI-side at each of the bytes sent to/from the host.
- TL-UL view of flash_status updates on CSB being inactive.

### DIFF
--- a/hw/dv/sv/spi_agent/spi_item.sv
+++ b/hw/dv/sv/spi_agent/spi_item.sv
@@ -32,6 +32,9 @@ class spi_item extends uvm_sequence_item;
   // of wait until the entire item is collected. This indicates item has collected all data.
   bit mon_item_complete;
 
+  //Triggered for each byte sampled and when CSB becomes inactive
+  event byte_sampled_ev, item_finished_ev;
+
   // constrain size of data sent / received to be at most 64kB
   constraint data_size_c { data.size() inside {[0:65536]}; }
 
--- a/hw/dv/sv/spi_agent/spi_item.sv
+++ b/hw/dv/sv/spi_agent/spi_item.sv
@@ -32,7 +32,7 @@ class spi_item extends uvm_sequence_item;
   // of wait until the entire item is collected. This indicates item has collected all data.
   bit mon_item_complete;
 
-  //Triggered for each byte sampled and when CSB becomes inactive
+  // Triggered for each byte sampled and when CSB becomes inactive
   event byte_sampled_ev, item_finished_ev;
 
   // constrain size of data sent / received to be at most 64kB
--- a/hw/dv/sv/spi_agent/spi_monitor.sv
+++ b/hw/dv/sv/spi_agent/spi_monitor.sv
@@ -23,19 +23,22 @@ class spi_monitor extends dv_base_monitor#(
   // Analysis port for the collected transfer.
   uvm_analysis_port #(spi_item) host_analysis_port;
   uvm_analysis_port #(spi_item) device_analysis_port;
+  uvm_analysis_port #(spi_item) csb_active_analysis_port; //Sends txn on CSB deassertion
 
   `uvm_component_new
 
   function void build_phase(uvm_phase phase);
     super.build_phase(phase);
-    host_analysis_port   = new("host_analysis_port", this);
-    device_analysis_port = new("device_analysis_port", this);
+    host_analysis_port       = new("host_analysis_port", this);
+    device_analysis_port     = new("device_analysis_port", this);
+    csb_active_analysis_port = new("csb_active_analysis_port", this);
   endfunction
 
   virtual task run_phase(uvm_phase phase);
     forever collect_trans();
   endtask
 
+  //TODO: Remove the UVM_PHASE reference - It's not being used
   // collect transactions
   virtual protected task collect_trans();
     bit flash_opcode_received;
@@ -46,9 +49,9 @@ class spi_monitor extends dv_base_monitor#(
 
     cfg.vif.sck_polarity = cfg.sck_polarity[active_csb];
     cfg.vif.sck_phase = cfg.sck_phase[active_csb];
-
     host_item   = spi_item::type_id::create("host_item", this);
     device_item = spi_item::type_id::create("device_item", this);
+    csb_active_analysis_port.write(host_item);
     fork
       begin : isolation_thread
         fork
@@ -75,6 +78,11 @@ class spi_monitor extends dv_base_monitor#(
         disable fork;
       end : isolation_thread
     join
+
+    -> host_item.item_finished_ev;
+    `uvm_info(`gfn, "Triggering 'host_item.item_finished' since CSB just became inactive",
+              UVM_DEBUG)
+
     // write to host_analysis_port
     case (cfg.spi_func_mode)
       SpiModeFlash: begin
@@ -90,7 +98,9 @@ class spi_monitor extends dv_base_monitor#(
         end
       end
       default: ; // do nothing, in SpiModeGeneric, it writes to fifo for each byte
-    endcase
+    endcase // case (cfg.spi_func_mode)
+
+
   endtask : collect_trans
 
   virtual protected task collect_curr_trans();
@@ -238,17 +248,25 @@ class spi_monitor extends dv_base_monitor#(
         num_addr_bytes, item.write_command, item.num_lanes, item.dummy_cycles);
     `uvm_info(`gfn, $sformatf("sampled flash opcode: 0x%0h", item.opcode), UVM_HIGH)
 
+    `uvm_info(`gfn, "Triggering 'host_item.byte_sampled' after sampling opcode", UVM_DEBUG)
+    -> host_item.byte_sampled_ev;
     sample_address(num_addr_bytes, item.address_q);
-
+    if(item.address_q.size > 0) begin
+      `uvm_info(`gfn, "Triggering 'host_item.byte_sampled' after sampling address", UVM_DEBUG)
+      -> host_item.byte_sampled_ev;
+    end
     repeat (item.dummy_cycles) begin
       cfg.wait_sck_edge(SamplingEdge, active_csb);
     end
+    `uvm_info(`gfn, "Sending {opcode,address} on the 'req_analysis_port'", UVM_DEBUG)
     req_analysis_port.write(item);
 
     forever begin
       logic[7:0] byte_data;
       sample_and_check_byte(item.num_lanes, !item.write_command, byte_data);
       item.payload_q.push_back(byte_data);
+      `uvm_info(`gfn, "Triggering 'host_item.byte_sampled' after sampling a data byte", UVM_DEBUG)
+      -> host_item.byte_sampled_ev;
     end
   endtask : collect_flash_trans
 
@@ -262,6 +280,9 @@ class spi_monitor extends dv_base_monitor#(
     decode_tpm_cmd(cmd, item.write_command, size);
     if (!item.write_command) item.read_size = size;
     `uvm_info(`gfn, $sformatf("Received TPM command: 0x%0x", cmd), UVM_MEDIUM)
+    `uvm_info(`gfn, "Triggering 'host_item.byte_sampled' after sampling opcode", UVM_DEBUG)
+    -> host_item.byte_sampled_ev;
+
     // read 3 bytes address
     fork
       begin : tpm_sio_0
@@ -279,6 +300,12 @@ class spi_monitor extends dv_base_monitor#(
       end
     join
     `uvm_info(`gfn, $sformatf("Received TPM addr: %p", item.address_q), UVM_MEDIUM)
+
+    if(item.address_q.size > 0) begin
+      `uvm_info(`gfn, "Triggering 'host_item.byte_sampled' after sampling address", UVM_DEBUG)
+      -> host_item.byte_sampled_ev;
+    end
+
     req_analysis_port.write(item);
 
     // poll TPM_START
@@ -297,6 +324,9 @@ class spi_monitor extends dv_base_monitor#(
                             .data(item.data[i]), .check_data_not_z(1));
       `uvm_info(`gfn, $sformatf("collect %s data for TPM, idx %0d: 0x%0x",
                               item.write_command ? "write" : "read", i, item.data[i]), UVM_MEDIUM)
+      `uvm_info(`gfn, "Triggering 'host_item.byte_sampled' after sampling a data byte", UVM_DEBUG)
+      -> host_item.byte_sampled_ev;
+
     end
   endtask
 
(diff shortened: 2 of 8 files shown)


antmarzam:
This PR takes care of  issue #21111

hw/dv/sv/spi_agent/spi_item.sv
```diff
@@ -32,6 +32,9 @@ class spi_item extends uvm_sequence_item;
   // of wait until the entire item is collected. This indicates item has collected all data.
   bit mon_item_complete;
 
+  //Triggered for each byte sampled and when CSB becomes inactive
```
rswarbrick, line 35: Nit: Would you mind adding some spaces after "//"? I think the usual syntax in this repo is "// a comment here".

hw/dv/sv/spi_agent/spi_monitor.sv
```diff
@@ -23,19 +23,22 @@ class spi_monitor extends dv_base_monitor#(
...
+    host_analysis_port       = new("host_analysis_port", this);
+    device_analysis_port     = new("device_analysis_port", this);
+    csb_active_analysis_port = new("csb_active_analysis_port", this);
   endfunction
 
   virtual task run_phase(uvm_phase phase);
     forever collect_trans(phase);
   endtask
 
+  //TODO: Remove the UVM_PHASE reference - It's not being used
```
rswarbrick, line 41: Hopefully most of the work is done for this! See #22175. I'd probably drop the TODO: it will conflict with that PR, which resolves it anyway :-)

hw/dv/sv/spi_agent/spi_monitor.sv
```diff
@@ -90,7 +98,9 @@ class spi_monitor extends dv_base_monitor#(
         end
       end
       default: ; // do nothing, in SpiModeGeneric, it writes to fifo for each byte
-    endcase
+    endcase // case (cfg.spi_func_mode)
+
+
```
rswarbrick, line 103: Not a big deal, but we seem to have gained a couple of blank lines here. Maybe drop them?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -23,9 +23,11 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
   uvm_tlm_analysis_fifo #(spi_item) upstream_spi_host_fifo;
   uvm_tlm_analysis_fifo #(spi_item) upstream_spi_device_fifo;
   // connect with req_analysis_port where it receives item when opcode and address are received
-  uvm_tlm_analysis_fifo #(spi_item) upstream_spi_req_fifo;
+  uvm_tlm_analysis_fifo #(spi_item) upstream_spi_req_fifo; //Just sends addr+opcode
```
rswarbrick, line 26: I think I understand, but this all gets a bit confusing with the comment on the line above as well: which comment applies to `downstream_spi_host_fifo`? Maybe the best thing would be to expand the comment above slightly so that it's clear what we're saying.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -23,9 +23,11 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
   uvm_tlm_analysis_fifo #(spi_item) downstream_spi_host_fifo;
 
+  uvm_tlm_analysis_fifo #(spi_item) upstream_csb_active_fifo; //RX a SPI-txn the when CSB is active
```
rswarbrick, line 29: Again, I'd be tempted to move this comment to the line above. That way you get more space! :-) "Receives an SPI transaction when CSB is active" should then fit...

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -52,12 +54,32 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
   bit read_buffer_watermark_triggered;
 
-  flash_status_t flash_status_q[$];
-  flash_status_t flash_status_settle_q[$];
-  // this queue contains the previous value of flash_status
-  // it's ok the readback value matches to either this one or the mirror value
-  // after a flash_status access, delete this queue as the uncertain window is only a few cycles
-  flash_status_t flash_status_tl_pre_val_q[$];
+  flash_status_t spi_side_flash_status; //Predicted spi-side flash status
+  flash_status_t tl_ul_side_flash_status_q[$]; // flash_status value written on TL-UL. There can be
```
rswarbrick, line 58: I'm sorry: I keep repeating myself! But I think it might be easier to read if these comments were above the definition lines. That way, they can be full width, giving much more space for description.

rswarbrick, line 58: About the names: eugh! Using "_q" to mean "queue" is a little bit at odds with the d/q naming that we use in the RTL! :-) I realise that you are following the existing style, but I wonder whether there is another naming scheme that we could use. I guess this particular name could be something like "tl_pending_flash_status" (where "tl_pending" means "something to be applied, from TL side"). But there might be a better option!

Any change to this names could be a separate PR: it definitely doesn't need to hold up this PR.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -52,12 +54,32 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                                               // CDC side before it's seen on the SPI CDC side.
+
+  flash_status_t tl_ul_old_flash_status_q[$];  // Needed for when TL-UL read occurs before CDC
+                                               // transition from SPI->UL after CSB is inactive.
+                                               // will have up to 2 items max. That's because once
+                                               // CSB goes high, if SW reads flash_status we may get
+                                               // the "old old" flash_status, or the new "old"
+                                               // value depending on when the read arrives
+
+  flash_status_t tl_ul_fuzzy_flash_status_q[$]; //This queue will have 1 entry at most, and it'll
```
rswarbrick, line 69: If this queue has at most one entry, I wonder whether a different type would be better. Maybe this should just be `flash_status_t`. It could be initialised to the reset value and set to any value that gets written in over TL.

Similarly for the next variable.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -127,65 +151,70 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
             // downstream item should be in the queue at the same time, add small delay
             #1ps;
             cmd_type = triage_flash_cmd(item.opcode, set_busy);
-            `uvm_info(`gfn, $sformatf("Triage flash cmd: %s, set_busy: %0d", cmd_type, set_busy),
+            `uvm_info(`gfn,
+                      $sformatf("Triage flash cmd: %s, set_busy: %0d", cmd_type.name, set_busy),
                       UVM_MEDIUM)
+            if(set_busy)
```
rswarbrick, line 156: Nit: Missing space after `if`. (And similarly on next `if` line)

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -127,65 +151,70 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
-                check_internal_processed_read_sfdp(item);
-              end
-              InternalProcessReadCmd: begin
-                spi_item downstream_item;
-                if (is_opcode_passthrough(item.opcode)) begin
-                  `DV_CHECK_EQ_FATAL(spi_passthrough_downstream_q.size, 1)
-                  downstream_item = spi_passthrough_downstream_q[0];
+            //InternalProcessStatus is handled by its own process
+            if(cmd_type != InternalProcessStatus) begin
+              case (cmd_type)
```
rswarbrick, line 162: I realise that this code was pretty complicated before you this PR. But I wonder: would it make sense to try to pull a chunk of this task into a sub-task? It's not immediately obvious to me how this all works!

(Again, this shouldn't gate this PR: I can see that you're just touching lots of lines here because the indentation is changing)

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -864,6 +769,308 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
     end
   endtask
 
+  //process_flash_tl2spi_updates: receives TXN on CSB being active and each of the sampled
```
rswarbrick, line 771: Nit: I probably wouldn't bother giving the name of the task in the comment. It's on the next line anyway :-)

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -864,6 +769,308 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+  // bytes of data as well as CSB becoming inactive
+  virtual task process_flash_tl2spi_updates();
+    forever begin
```
rswarbrick, line 774: Would it make sense to make a task for the body of this loop? That would eat 2 spaces of indentation and might make things a bit easier to read at a glance. Maybe the outer task could just be popping from `ubstream_csb_active_fifo`.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -864,6 +769,308 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+      //Received TXN is used to know when to update TL-UL and SPI flash_status values
+      // The SPI side updates on each 8th clock edge whereas the TL-UL side updates on
+      // CSB deassertion
+      `uvm_info(`gfn, "CSB is now active (low)", UVM_DEBUG)
+      fork
+        begin: isolation_thread
+          fork
+            begin
+              //Exit if CSB becomes inactive
+              wait(spi_txn.item_finished_ev.triggered);
```
rswarbrick, line 788: Yet another style nit: I think this should probably be `wait (spi_txn...);`. Without the space, it looks like `wait` is a function that we're calling.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -864,6 +769,308 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+              `uvm_info(`gfn,
+              "[spi_txn.item_finished_ev triggered] - CSB has gone inactive - killing thread",
```
rswarbrick, line 790: Nit: indentation

rswarbrick (commented):
I've left lots of nitty style comments (sorry) on some of the first commit. Sorry that it has taken so long.

I'm a big fan of the extra UVM_DEBUG logs: I've never really looked at code where that turns out to be useful. It's cool to see them in use!

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: As discussed, I'm leaving SV queues with "_q" on the TB code side only just to make it obvious at a glance these are SV queues

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: The idea behind using a SV queue (even for just 1 item) instead of a static variable was just so that we could use ".size>0" to determine if there are valid entries. I thought of adding a "valid" bit to `flash_status_t`, but it's defined as:
``` 
  typedef struct packed {
    bit [21:0] other_status;
    bit wel;
    bit busy;
  } flash_status_t;
```
And it's also used to directly read/write the flash status CSR, so figured that rather than wrapping it around another struct or another data type I'd just use a SV queue which is valid if it has 1 entry.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: I see what you mean, but once you understand internally processed flash_commands I think it makes sense to have them laid out in a case like this. 
Nonetheless, I'm happy to amend for readability either in this PR or a new one maybe with a sub-task called `internally_processed_flash_cmd` or similar.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: I'm inclined to leave it as-is, just because this task only handles the prediction of the flash_status bits.
It receives a TXN the moment CSB becomes active, and keeps updating the predicted spi_side flash_status at every SPI beat. 
Then on CSB assertion updates the TL-UL side. 
If the task was dealing with more than just flash_status modelling I agree it would be handy to have it split, but I'm not sure I see the benefit. 
But as above, LMK if I'm not convincing as I'm happy to edit for readability either in this PR or a different one!

rswarbrick (approved):
This looks good, thanks. As a minor nit, I think you addressed some of my style comments in the second commit, so we sometimes get a sequence of "... mess up style; fix style ...". But amending exactly the right commit is a pain!

On the plus side, magit has a reasonably nice way of doing this: either @hcallahan-lowrisc or I can probably show you in person at some point. Another win for Emacs :-)

But it's definitely not worth re-doing everything for. Thank you very much for tidying things up.

a-will:
Was a significant subset of tests run on this? It looks like regression results have taken a big dive between before this change (https://reports.opentitan.org/hw/ip/spi_device_2p/dv/2024.03.29_04.21.20/report.html) and after (https://reports.opentitan.org/hw/ip/spi_device_2p/dv/2024.04.01_04.52.26/report.html).

Are those real failures?

antmarzam:
> Was a significant subset of tests run on this? It looks like regression results have taken a big dive between before this change (https://reports.opentitan.org/hw/ip/spi_device_2p/dv/2024.03.29_04.21.20/report.html) and after (https://reports.opentitan.org/hw/ip/spi_device_2p/dv/2024.04.01_04.52.26/report.html).
> 
> Are those real failures?

I run quite a bunch of tests for the flash_status changes. I had some spurious failures of tests timing out which passed once I rerun them. 
I'll investigate!

PR #21979, opened 2024-03-12
[aon_timer,sival] Fix aon_timer_smoketest in silicon

pamaury:
In non-TestROM environements, the ROM enables the NMI for the watchdog which is not caught by the test. This commit makes the code handle the NMI in non-testrom.

One could say that with this change, the test becomes quite close to the aon IRQ test but it is still simpler I think.

### DIFF
--- a/sw/device/tests/aon_timer_smoketest.c
+++ b/sw/device/tests/aon_timer_smoketest.c
@@ -6,6 +6,7 @@
 
 #include "sw/device/lib/base/mmio.h"
 #include "sw/device/lib/dif/dif_aon_timer.h"
+#include "sw/device/lib/dif/dif_rv_core_ibex.h"
 #include "sw/device/lib/runtime/hart.h"
 #include "sw/device/lib/runtime/log.h"
 #include "sw/device/lib/testing/aon_timer_testutils.h"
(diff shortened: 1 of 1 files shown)


sw/device/tests/aon_timer_smoketest.c
```diff
@@ -16,6 +17,10 @@
 
 OTTF_DEFINE_TEST_CONFIG();
 
+static dif_aon_timer_t aon;
+static dif_rv_core_ibex_t rv_core_ibex;
+static bool wdog_fired = false;
```
jwnrt, line 22: Should this be volatile since it crosses the ISR boundary?

jwnrt: I guess it's only read once so I'm not sure if volatile is needed, but it can't hurt

sw/device/tests/aon_timer_smoketest.c
```diff
@@ -36,6 +41,31 @@ static void aon_timer_test_wakeup_timer(dif_aon_timer_t *aon) {
...
+  // The watchdog bark external interrupt is also connected to the NMI input
+  // of rv_core_ibex. We therefore expect the interrupt to be pending on the
+  // peripheral side.
+  CHECK_DIF_OK(dif_aon_timer_irq_is_pending(&aon, kDifAonTimerIrqWdogTimerBark,
+                                            &is_pending));
+  if (is_pending) {
+    wdog_fired = true;
+    CHECK_DIF_OK(dif_aon_timer_watchdog_stop(&aon));
+    // In order to handle the NMI we need to acknowledge the interrupt status
+    // bit it at the peripheral side.
```
engdoreis, line 60:
```suggestion
    // bit at the peripheral side.
```

engdoreis (approved):
LGTM.

sw/device/tests/aon_timer_smoketest.c
pamaury: Hum, I think it does not technically need to but you are right that it doesn't hurt, in case the compiler would like to do fancy optimization.

PR #21980, opened 2024-03-12
[bazel] Change default environments for test targets.

jadephilipoom:
By default, `EARLGREY_TEST_ENVS` used to run tests on both test ROM and regular ROM. This results in a lot of duplicated runs on CI. This commit removes the test ROM from the defaults list, leaving only ROM. If a test has explicitly overridden this default value (and e.g. specified it only runs on test ROM), then it will be unaffected by this change.

The second commit updates documentation: since `uart_smoketest` and other targets no longer include the
`test_rom` environment, it updates the Bazel commands in the documentation to use the ROM targets instead. It also updates a few references to `_rom` targets (which no longer exist afaict, having been replaced by `_rom_with_fake_keys` targets) and clarifies that not all tests will support both test ROM and ROM.

If this works like we hope, it should mean the test ROM portion of CI is a lot shorter.

### DIFF
--- a/doc/getting_started/build_sw.md
+++ b/doc/getting_started/build_sw.md
@@ -117,7 +117,7 @@ You may find it useful to use wildcards to build/test all targets in the OpenTit
 If a target (a test or build artifact) relies on optional parts of the "Getting Started" guide they should be tagged so they can be filtered out and users can `bazelisk.sh test //...` once they filter out the appropriate tags.
 We maintain or use the following tags to support this:
 * `broken` is used to tag tests that are committed but should not be expected by CI or others to pass.
-* `cw310`, `cw310_test_rom`, and `cw310_rom` are used to tag tests that depend on a correctly setup cw310 "Bergen Board" to emulate OpenTitan.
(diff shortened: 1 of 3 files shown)


jwnrt (approved):

pamaury (approved):
Thanks for the change

cfrantz (approved):

jadephilipoom:
The test ROM run still takes 23 minutes -- that's down from 35 minutes before, but still much higher than I expected. I looked into what's going on there, and it looks like a lot of ROM crypto tests are running in CI. In particular, the large Wycheproof test set (238 tests) is running for RSA and the full test set (100 tests) is running for SPHINCS+. Both of these routines have shorter tests that also run, e.g. one or two test vectors.

I think both the Wycheproof RSA tests and the 100 SPHINCS+ tests are good candidates to move to a nightly build; most issues should be caught in the short tests, and there's really no reason we need to run everything on every PR. But I think that's a different category of change, so my suggestion for now is to merge this PR (once CI passes, I think `master` has some issues right now) for at least the 12min of time saved, and then make a separate PR that moves the crypto tests to nightly.

jadephilipoom:
This finally passed CI after a rebase; I'll merge it now.

Issue #16516, opened 2022-11-22
[csrng] Define how CSRNG status response should operate


marnovandermaas:
With the current RTL, the `sw_cmd_sts.cmd_sts` is never high and the bits `hw_exc_sts` are also never high. This is due to the current design of the RTL not being designed to pass back anything but a successful status. For example, when an illegal command is received the main state machine simply hangs until it is reset and does not return any status.

In the documentation it states that there are three reasons for there to be an illegal status: 
1. https://github.com/lowRISC/opentitan/blob/b15fa805be3bae4528310e614372d91761faef28/hw/ip/csrng/doc/_index.md?plain=1#L416
1. https://github.com/lowRISC/opentitan/blob/b15fa805be3bae4528310e614372d91761faef28/hw/ip/csrng/doc/_index.md?plain=1#L417
1. https://github.com/lowRISC/opentitan/blob/b15fa805be3bae4528310e614372d91761faef28/hw/ip/csrng/doc/_index.md?plain=1#L418

We should decide whether the current design is desirable and verify that each of these scenarios have the correct behavior.

Assuming no changes to RTL:
> estimate 2

marnovandermaas:
This issue is also related to: https://github.com/lowRISC/opentitan/issues/16499

vogelpi:
Thanks for filing this @marnovandermaas and @ctopal . I agree with you, we should discuss how the RTL should behave and then also verify that. I don't think any of this is super critical for ES but it seems that there is at least the need for some clarification. Maybe it's all about aligning the documentation, maybe some RTL changes should be done. It would be best to discuss this with @mwbranstad after the break to get his view and the original intention. 

Regarding the second reason for an failure status "Attempts to use an instance which has not been properly instantiated" I think I've seen something in the RTL but it was more checking the RTL instantiation of the command stage, i.e., SV parameters. From the spec, I would have expected the design to report a failing status if e.g. a generate command is received without prior instantiate command.

We've also discussed to return a failing status bit upon a fatal alert but that wouldn't be in line with how we handle fatal alerts in other places in the system and it might need big design changes. But it could be useful to get some more immediate feedback in the case of misconfiguration. E.g. if an EDN or SW provides an illegal command, there is no response and CSRNG and downstream modules just hang. Software will have to figure out what caused this. This could maybe be simplified if also EDN/SW would see the failure via `sts` bit more directly. Let's discuss with Mark after the break.

vogelpi:
[...]

PR #21981, opened 2024-03-12
[csrng/rtl] Propagate csrng errors to edn through cmd sts signal

h-filali:
This PR adds functionality to signal status errors along a command acknowledgement.
For further details have a look at the individual commit messages.

Resolves [#16516](https://github.com/lowRISC/opentitan/issues/16516)

### DIFF
--- a/hw/dv/sv/csrng_agent/csrng_agent_cfg.sv
+++ b/hw/dv/sv/csrng_agent/csrng_agent_cfg.sv
@@ -15,13 +15,13 @@ class csrng_agent_cfg extends dv_base_agent_cfg;
 
   `uvm_object_new
 
-  uint   min_cmd_ack_dly, max_cmd_ack_dly,
-         min_genbits_dly, max_genbits_dly,
-         min_genbits_rdy_dly, max_genbits_rdy_dly,
-         min_cmd_rdy_dly, max_cmd_rdy_dly,
-         reseed_cnt, generate_cnt, generate_between_reseeds_cnt;
-  bit    cmd_zero_delays;
-  bit    under_reset;
-  bit    rsp_sts_err;
+  uint              min_cmd_ack_dly, max_cmd_ack_dly,
+                    min_genbits_dly, max_genbits_dly,
+                    min_genbits_rdy_dly, max_genbits_rdy_dly,
+                    min_cmd_rdy_dly, max_cmd_rdy_dly,
+                    reseed_cnt, generate_cnt, generate_between_reseeds_cnt;
+  bit               cmd_zero_delays;
+  bit               under_reset;
+  csrng_cmd_sts_e   rsp_sts_err;
 
 endclass
--- a/hw/ip/csrng/data/csrng.hjson
+++ b/hw/ip/csrng/data/csrng.hjson
@@ -325,20 +325,45 @@
       tags: [// Internal HW can modify status register
                  "excl:CsrNonInitTests:CsrExclCheck"]
       fields: [
-        { bits: "0",
-          name: "CMD_RDY",
-          desc: '''This bit indicates when the command interface is ready to accept commands.
-                '''
-          resval: "1"
-        }
         { bits: "1",
+          name: "CMD_RDY",
+          desc: '''
+                This bit indicates when the command interface is ready to accept commands.
+                Before starting to write a new command to !!SW_CMD_REQ, this field needs to be polled.
+                0b0: CSRNG is not ready to accept commands or the last command hasn't been acked yet.
+                0b1: CSRNG is ready to accept the next command.
+                '''
+          resval: "0"
+        }
+        { bits: "2",
+          name: "CMD_ACK",
+          desc: '''
+                This one bit field indicates when a SW command has been acknowledged by the CSRNG.
+                It is set to low each time a new command is written to !!CMD_REQ.
+                The field is set to high once a SW command request has been acknowledged by the CSRNG.
+                0b0: The last SW command has not been acknowledged yet.
+                0b1: The last SW command has been acknowledged.
+                '''
+          resval: "0"
+        }
+        { bits: "4:3",
           name: "CMD_STS",
           desc: '''
-                This one bit field is the status code returned with the application command ack.
+                This field represents the status code returned with the application command ack.
                 It is updated each time a command ack is asserted on the internal application
                 interface for software use.
-                0b0: Request completed successfully
-                0b1: Request completed with an error
+                To check whether a command was succesful, wait for !!INTR_STATE.CS_CMD_REQ_DONE or
+                !!SW_CMD_STS.CMD_ACK to be high and then check the value of this field.
+                0x0: Request completed successfully.
+                0x1: Request completed with an invalid application command error.
+                     This error indicates that the issued application command doesn't represent a valid operation.
+                     If this error appears, the main state machine will hang and the entropy complex has to be restarted.
+                0x2: Request completed with an invalid state parameter error.
+                     This error indicates that the state wasn't zeroized properly after an uninstantiate command.
+                     In this case the entropy complex should be restarted to properly zeroize the state.
+                0x3: Request completed with an invalid counter drbg generation command error.
+                     This error indicates that CSRNG entropy was generated for a command that is not a generate command.
+                     In this case the entropy should not be considered as valid.
                 '''
           resval: "0"
         }
--- a/hw/ip/csrng/doc/registers.md
+++ b/hw/ip/csrng/doc/registers.md
@@ -157,30 +157,51 @@ Command request register
 ## SW_CMD_STS
 Application interface command status register
 - Offset: `0x1c`
-- Reset default: `0x1`
-- Reset mask: `0x3`
+- Reset default: `0x0`
+- Reset mask: `0x1e`
 
 ### Fields
 
 ```wavejson
-{"reg": [{"name": "CMD_RDY", "bits": 1, "attr": ["ro"], "rotate": -90}, {"name": "CMD_STS", "bits": 1, "attr": ["ro"], "rotate": -90}, {"bits": 30}], "config": {"lanes": 1, "fontsize": 10, "vspace": 90}}
+{"reg": [{"bits": 1}, {"name": "CMD_RDY", "bits": 1, "attr": ["ro"], "rotate": -90}, {"name": "CMD_ACK", "bits": 1, "attr": ["ro"], "rotate": -90}, {"name": "CMD_STS", "bits": 2, "attr": ["ro"], "rotate": -90}, {"bits": 27}], "config": {"lanes": 1, "fontsize": 10, "vspace": 90}}
 ```
 
 |  Bits  |  Type  |  Reset  | Name                            |
 |:------:|:------:|:-------:|:--------------------------------|
-|  31:2  |        |         | Reserved                        |
-|   1    |   ro   |   0x0   | [CMD_STS](#sw_cmd_sts--cmd_sts) |
-|   0    |   ro   |   0x1   | [CMD_RDY](#sw_cmd_sts--cmd_rdy) |
+|  31:5  |        |         | Reserved                        |
+|  4:3   |   ro   |   0x0   | [CMD_STS](#sw_cmd_sts--cmd_sts) |
+|   2    |   ro   |   0x0   | [CMD_ACK](#sw_cmd_sts--cmd_ack) |
+|   1    |   ro   |   0x0   | [CMD_RDY](#sw_cmd_sts--cmd_rdy) |
 
 ### SW_CMD_STS . CMD_STS
-This one bit field is the status code returned with the application command ack.
+This field represents the status code returned with the application command ack.
 It is updated each time a command ack is asserted on the internal application
 interface for software use.
-0b0: Request completed successfully
-0b1: Request completed with an error
+To check whether a command was succesful, wait for [`INTR_STATE.CS_CMD_REQ_DONE`](#intr_state) or
+[`SW_CMD_STS.CMD_ACK`](#sw_cmd_sts) to be high and then check the value of this field.
+0x0: Request completed successfully.
+0x1: Request completed with an invalid application command error.
+     This error indicates that the issued application command doesn't represent a valid operation.
+     If this error appears, the main state machine will hang and the entropy complex has to be restarted.
+0x2: Request completed with an invalid state parameter error.
+     This error indicates that the state wasn't zeroized properly after an uninstantiate command.
+     In this case the entropy complex should be restarted to properly zeroize the state.
+0x3: Request completed with an invalid counter drbg generation command error.
+     This error indicates that CSRNG entropy was generated for a command that is not a generate command.
+     In this case the entropy should not be considered as valid.
+
+### SW_CMD_STS . CMD_ACK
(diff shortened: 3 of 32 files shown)


hw/ip/edn/data/edn.hjson
```diff
@@ -351,6 +339,26 @@
...
+                0b00: Request completed successfully.
+                0b01: Request completed with an invalid application command error.
+                      This error indicates that the issued application command doesn't represent a valid operation.
+                      If this error appears, the main state machine will hang and the entropy complex has to be restarted.
+                0b10: Request completed with an invalid state parameter error.
+                      This error indicates that the state wasn't zeroized properly after an uninstantiate command.
+                      In this case the entropy complex should be restarted to properly zeroize the state.
+                0b11: Request completed with an invalid counter drbg generation command error.
+                      This error indicates that CSRNG entropy was generated for a command that is not a generate command.
+                      In this case the entropy should not be considered as valid.
```
andreaskurth, line 358: Instead of duplicating this information from CSRNG, would it makes sense to refer to the CSRNG doc (like for `HW_CMD_STS.CMD_TYPE` below)?

hw/ip/edn/data/edn.hjson
```diff
@@ -379,17 +387,16 @@
...
+        { bits: "5:2",
+          name: "CMD_TYPE",
           desc: '''
-                This one bit field represents the status code returned with the CSRNG application command ack.
-                It is updated each time a HW command is acknowledged by CSRNG.
-                0b0: Request completed successfully.
-                0b1: Request completed with an error.
+                This field contains the application command type of the hardware controlled command issued last.
+                The application command selects one of five operations to perform.
+                A description of the application command types can be found [here](https://opentitan.org/book/hw/ip/csrng/doc/theory_of_operation.html#command-description).
```
andreaskurth, line 395: Suggest to use relative paths (like `../../csrng/doc/theory_of_operation.md#command-description`) instead, because that will work with local files, and it will work if we change the web interface to the documentation

hw/ip/edn/data/edn.hjson
```diff
@@ -400,12 +407,21 @@
...
+                0b00: Request completed successfully.
+                0b01: Request completed with an invalid application command error.
+                      This error indicates that the issued application command doesn't represent a valid operation.
+                      If this error appears, the main state machine will hang and the entropy complex has to be restarted.
+                0b10: Request completed with an invalid state parameter error.
+                      This error indicates that the state wasn't zeroized properly after an uninstantiate command.
+                      In this case the entropy complex should be restarted to properly zeroize the state.
+                0b11: Request completed with an "invalid counter drbg generation command" error.
+                      This error indicates that CSRNG entropy was generated for a command that is not a generate command.
+                      In this case the entropy should not be considered as valid.
```
andreaskurth, line 424: Ditto referencing the CSRNG doc?

hw/dv/sv/csrng_agent/csrng_agent_cfg.sv
```diff
@@ -15,13 +15,13 @@ class csrng_agent_cfg extends dv_base_agent_cfg;
...
 
-  uint   min_cmd_ack_dly, max_cmd_ack_dly,
-         min_genbits_dly, max_genbits_dly,
-         min_genbits_rdy_dly, max_genbits_rdy_dly,
-         min_cmd_rdy_dly, max_cmd_rdy_dly,
-         reseed_cnt, generate_cnt, generate_between_reseeds_cnt;
-  bit    cmd_zero_delays;
-  bit    under_reset;
-  bit    rsp_sts_err;
+  uint              min_cmd_ack_dly, max_cmd_ack_dly,
```
andreaskurth, line 18: Wouldn't it be better to combine this commit with the first one, so that tests continue to pass despite the RTL changes?

hw/ip/csrng/doc/registers.md
```diff
@@ -157,30 +157,51 @@ Command request register
 ## SW_CMD_STS
 Application interface command status register
 - Offset: `0x1c`
-- Reset default: `0x1`
-- Reset mask: `0x3`
+- Reset default: `0x0`
```
andreaskurth, line 160: I'd also combine this commit with the first one because otherwise documentation will be outdated

sw/device/lib/crypto/drivers/entropy.c
```diff
@@ -450,9 +450,9 @@ static status_t csrng_send_app_cmd(uint32_t base_address,
         reg = abs_mmio_read32(kBaseCsrng + CSRNG_INTR_STATE_REG_OFFSET);
       } while (!bitfield_bit32_read(reg, CSRNG_INTR_STATE_CS_CMD_REQ_DONE_BIT));
 
-      // Check the "status" bit, which will be 1 only if there was an error.
+      // Check the "status" bit, which will be 0 unless there was an error.
```
andreaskurth, line 453: I'd also combine this commit with the first one, because otherwise SW tests will fail

andreaskurth (commented):
Thanks for this work, @h-filali. I think this is almost ready, just a few questions and suggestions from my side

hw/ip/csrng/rtl/csrng_ctr_drbg_cmd.sv
```diff
@@ -318,8 +319,11 @@ module csrng_ctr_drbg_cmd import csrng_pkg::*; #(
   // block ack
   assign ctr_drbg_cmd_ack_o = sfifo_keyvrc_pop;
 
-  assign ctr_drbg_cmd_sts_o = sfifo_keyvrc_pop && (ctr_drbg_cmd_ccmd_o == UNI) &&
+  // Return a status error when the state is not zeroized properly during a UNI command.
+  assign ctr_drbg_cmd_sts_err = sfifo_keyvrc_pop && (ctr_drbg_cmd_ccmd_o == UNI) &&
          ((KeyLen == '0) && (BlkLen == '0) && (CtrLen == '0));
```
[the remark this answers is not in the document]
h-filali: I tried to change as little as possible but I agree with your point. I'll adapt the check such that it actually checks if we are zeroizing properly.

hw/ip/csrng/rtl/csrng_ctr_drbg_gen.sv
```diff
@@ -588,13 +591,17 @@ module csrng_ctr_drbg_gen import csrng_pkg::*; #(
...
 
-  assign ctr_drbg_gen_sts_o = sfifo_genbits_pop && (
+  // Return a status error when the genbits FIFO is popped while ctr_drbg_gen_ccmd_o is not
+  // equal to GEN.
+  assign ctr_drbg_gen_sts_err = sfifo_genbits_pop && (
          (ctr_drbg_gen_ccmd_o == INV) ||
          (ctr_drbg_gen_ccmd_o == INS) ||
          (ctr_drbg_gen_ccmd_o == RES) ||
          (ctr_drbg_gen_ccmd_o == UPD) ||
          (ctr_drbg_gen_ccmd_o == UNI));
```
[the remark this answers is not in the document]
h-filali: Good point! I'll change it.

hw/ip/csrng/rtl/csrng_state_db.sv
```diff
@@ -37,30 +37,30 @@ module csrng_state_db import csrng_pkg::*; #(
...
-  output logic               state_db_sts_sts_o,
+  output csrng_cmd_sts_e     state_db_sts_sts_o,
   output logic [StateId-1:0] state_db_sts_id_o
 );
 
   localparam int InternalStateWidth = 2+KeyLen+BlkLen+CtrLen;
   localparam int RegInternalStateWidth = 30+InternalStateWidth;
   localparam int RegW = 32;
-  localparam int StateWidth = 1+1+KeyLen+BlkLen+CtrLen+StateId+1;
+  localparam int StateWidth = 1+1+KeyLen+BlkLen+CtrLen+StateId+2;
```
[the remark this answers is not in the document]
h-filali: Excellent point, I totally missed that.

hw/ip/edn/data/edn.hjson
h-filali: Yes it would.

hw/dv/sv/csrng_agent/csrng_agent_cfg.sv
h-filali: Sure, I'll squash.

hw/ip/csrng/data/csrng.hjson
```diff
@@ -325,20 +325,45 @@
...
           desc: '''
-                This one bit field is the status code returned with the application command ack.
+                This field represents the status code returned with the application command ack.
                 It is updated each time a command ack is asserted on the internal application
                 interface for software use.
-                0b0: Request completed successfully
-                0b1: Request completed with an error
+                To check whether a command was succesful, wait for !!INTR_STATE.CS_CMD_REQ_DONE or
+                !!SW_CMD_STS.CMD_ACK to be high and then check the value of this field.
+                0b00: Request completed successfully.
```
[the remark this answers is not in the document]
h-filali: Sure, done!

h-filali:
@andreaskurth I addressed all your points and rebased to the current master. Thanks a lot for reviewing!

hw/ip/csrng/rtl/csrng_ctr_drbg_cmd.sv
```diff
@@ -318,8 +320,11 @@ module csrng_ctr_drbg_cmd import csrng_pkg::*; #(
   // block ack
   assign ctr_drbg_cmd_ack_o = sfifo_keyvrc_pop;
 
-  assign ctr_drbg_cmd_sts_o = sfifo_keyvrc_pop && (ctr_drbg_cmd_ccmd_o == UNI) &&
-         ((KeyLen == '0) && (BlkLen == '0) && (CtrLen == '0));
+  // Return a status error when the state is not zeroized properly during a UNI command.
+  assign ctr_drbg_cmd_sts_err = sfifo_keyvrc_pop && (ctr_drbg_cmd_ccmd_o == UNI) &&
+         (sfifo_keyvrc_wdata >> (KeyVRCFifoWidth-UniZeroizeWidth)) == '0;
```
vogelpi, line 325: I am not 100% sure but I think this is another error condition in the design which doesn't really make sense as it's never going to happen (similarly to receiving an INS command on an instance which is not instantiated in the RTL ). Here we have in the same file:
```sv
  assign sfifo_keyvrc_wdata = (rcstage_ccmd == UNI) ?
         {{(KeyLen+BlkLen+CtrLen+1+SeedLen){1'b0}},rcstage_glast,upd_cmd_inst_id_i,upd_cmd_ccmd_i} :
```
meaning when the command seen is UNI, we set the relevant `wdata` bits to 0 anyway. This might be a coverage hole and synthesis might optimize this away. Things like that should be tested with SVAs. Did I miss something @h-filali ?

vogelpi: Ah, now I understood your comment in another discussion with @andreaskurth about "doing minimal changes". I think agree with you but I think in the case of CSRNG we should maybe sit together and discuss what we really want.

vogelpi (commented):
Thanks for the PR @h-filali ! It looks mostly good to me but I have some questions. Regarding the zeroizing check, I am more in favor of removing that entirely but let's discuss once you're back from vacation and exams.

hw/ip/csrng/rtl/csrng_pkg.sv
```diff
@@ -22,17 +23,25 @@ package csrng_pkg;
     logic                       genbits_ready;
   } csrng_req_t;
 
+  typedef enum logic [CSRNG_CMD_STS_WIDTH-1:0] {
+    CMD_STS_SUCCESS              = 2'h0,
+    CMD_STS_INVALID_ACMD         = 2'h1,
+    CMD_STS_INVALID_STATE_PARAM  = 2'h2,
+    CMD_STS_INVALID_GEN_CMD      = 2'h3,
+    CMD_STS_UNDRIVEN             = 'z
```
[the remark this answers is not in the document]
h-filali: I think you mean the undriven value here not the entire enum type.
I spent quite some time getting the CSRNG interface to work. Xcelium doesn't like it if we assign 'z to a struct that has a substruct. There's different ways you can do this and this is the only way I found that works both in the RTL and for the block level simulation.

sw/device/lib/dif/dif_csrng.c
```diff
@@ -147,90 +147,12 @@ dif_result_t dif_csrng_get_cmd_interface_status(
...
+  uint32_t cmd_sts = bitfield_field32_read(reg, CSRNG_SW_CMD_STS_CMD_STS_FIELD);
 
-  // The function prioritizes error detection to avoid masking errors
-  // when `cmd_ready` is set to true.
-  if (cmd_error) {
+  status->cmd_sts = cmd_sts;
+  // If the command did not execute successfully, return the error status.
+  if (cmd_sts != 0) {
     status->kind = kDifCsrngCmdStatusError;
-    uint32_t reg =
```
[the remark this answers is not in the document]
h-filali: I think the function description should be fine but let's check in person.

sw/device/lib/dif/dif_edn.c
```diff
@@ -157,26 +157,26 @@ dif_result_t dif_edn_get_status(const dif_edn_t *edn, dif_edn_status_t flag,
...
+      *set = bitfield_bit32_read(reg, EDN_SW_CMD_STS_CMD_REG_RDY_BIT);
       break;
     case kDifEdnStatusReady:
-      bit = EDN_SW_CMD_STS_CMD_RDY_BIT;
+      *set = bitfield_bit32_read(reg, EDN_SW_CMD_STS_CMD_RDY_BIT);
       break;
     case kDifEdnStatusCsrngStatus:
-      bit = EDN_SW_CMD_STS_CMD_STS_BIT;
+      field_val = bitfield_field32_read(reg, EDN_SW_CMD_STS_CMD_STS_FIELD);
+      *set = field_val ? true : false;
```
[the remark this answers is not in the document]
h-filali: Yes I think we can do that in a separate issue. Let's talk about it and create an issue.

hw/ip/csrng/rtl/csrng_ctr_drbg_cmd.sv
h-filali: Yes, let's discuss this in person.

sw/device/lib/dif/dif_edn.c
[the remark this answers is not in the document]
vogelpi: Discussed and agreed to create a separate software issue for this. This isn't critical for the TO.

sw/device/lib/dif/dif_csrng.c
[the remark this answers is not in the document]
vogelpi: Checked and agreed that no further changes are needed here.

hw/ip/csrng/rtl/csrng_pkg.sv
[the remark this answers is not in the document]
vogelpi: Discussed and agreed in person that this is the way to handle this. Already previously, something very similar was done but only now we need to define an enum value for this case. This is okay.

vogelpi (approved):
LGTM after discussing in person.

PR #21983, opened 2024-03-12
[usbdev] Fixes to IN packet engine

alees24:
This PR addresses a couple of defects in the IN packet engine:

- Handling of IN token packet reception whilst awaiting the ACK from a previous IN transaction did not check whether the endpoint is active. This aspect of the endpoint configuration is not expected to change whilst the USB device is connected, and the host should not be attempting to retrieve packets from an endpoint that is not being used. The logic should be consistent, however.

- Recent change to maintain IN buffer properties throughout the transaction (#21771) was incomplete; the logic needs qualifying with the two IN transaction states to capture the buffer properties only at the start of an IN transaction. verification effort still pending. Spotted by inspection when implementing event counters.


a-will (approved):
Nice catch!

alees24:
Failures unrelated.

PR #21985, opened 2024-03-12
[usbdev,dv] Remove usb20_if uses from usbdev block-level DV

rswarbrick:
There's a slightly complicated situation where we currently have two USB interfaces (usb20_if and usb20_block_if).

At the moment, usb20_if is used at chip-level (interfacing over DPI with a USB reference model) and usb20_block_if is used at block level.

Keeping them separate avoids block-level DV engineers who can't run chip-level tests breaking chip-level sims. This is a slightly silly situation and the right eventual solution would probably be to merge the two interfaces again.

But a short-term cleanup is to remove the chip-level-only interface from the block-level testbench.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_agent.sv
+++ b/hw/dv/sv/usb20_agent/usb20_agent.sv
@@ -15,10 +15,6 @@ class usb20_agent extends dv_base_agent #(
 
 function void build_phase(uvm_phase phase);
   super.build_phase(phase);
-  // get usb20_if handle
-  if (!uvm_config_db#(virtual usb20_if)::get(this, "", "vif", cfg.vif)) begin
-    `uvm_fatal(`gfn, "failed to get usb20_if handle from uvm_config_db")
-  end
   // get usb20_block_if handle
   if (!uvm_config_db#(virtual usb20_block_if)::get(this, "", "bif", cfg.bif)) begin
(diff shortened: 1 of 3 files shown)


alees24:
After we discussed this, I've given it a bit more thought and I think this is really not the right time to be performing this tidy up when things aren't actually broken, so I'd rather park it for now, please.

The DPI model does an awful lot more verification than our DV block level presently does, including:

- reads descriptors using complete Control Transfers as per a real USB host, and configure the device
- streaming and checking of LFSR-generated test to/from maximal endpoints, most transfer types
- injection of faults
- Low Speed traffic, invalid traffic, unexpected device addresses and endpoints
- Suspend, Resume signaling, Disconnect, Reconnect, Bus Resets

The main limitations of the DPI model are not in the DPI model itself but rather the way that it is used; at chip level, it takes a long time to build the chip model, and then to simulate.

At block level, if connected into the DV environment it could more readily offer randomization, particularly signal timing/inter-packet delays etc, and would likely give us greatly improved coverage even without modification, since it is already the product of months of work that still remain to be done in the block level SV/UVM code.

rswarbrick:
Ok, that makes sense. My long-term aim is to merge the two interfaces, and switch the block-level DV to use the merged version. But I don't have strong feelings about the ordering! Let's leave things as they are for now.

PR #21992, opened 2024-03-12
[rom_ctrl] Switch to overridable memory mechanism

msfschaffner:
This is a carryover commit from `integrated_dev`.

It will make parameterization of the ROM_CTRL via the `top_earlgrey.hjson` much more straightforward using the same mechanism used for SRAM_CTRL.

It will also allow us to have multiple ROM_CTRL instances in a design, just like we do in Darjeeling.

---------------------------------

Unfortunately we can't override window sizes at the top-level, which would be necessary in order to instantiate two different ROM_CTRL parameterizations.

To that end, this patch switches to the same topgen override mechanism that the SRAM_CTRL uses. That however means that we have to remove the window and second reg_top, which in turn affects how the DV environment is constructed.

Most edits there have basically been carried over from the SRAM_CTRL testbench.

This patch also adds simulation configurations for the relevant ROM sizes.


a-will:
> Unfortunately we can't override window sizes at the top-level, which would be necessary in order to instantiate two different ROM_CTRL parameterizations.

Out of curiosity, is this going to make sense as an ipgen type someday? It kind of feels like we're avoiding our infrastructure, but I can't claim familiarity with the decisions that led to SRAM_CTRL's mechanism. :)

msfschaffner:
> > Unfortunately we can't override window sizes at the top-level, which would be necessary in order to instantiate two different ROM_CTRL parameterizations.
> 
> Out of curiosity, is this going to make sense as an ipgen type someday? It kind of feels like we're avoiding our infrastructure, but I can't claim familiarity with the decisions that led to SRAM_CTRL's mechanism. :)

I believe that mechanism was implemented at a time where Ipgen wasn't a thing yet and we did not want to have yet another templated IP that uses some custom script for generation ;). 

Converting to Ipgen could make sense, although maybe we need some sort of intermediate mechanism where overriding Hjson parameters is enough (i.e., no templating is needed since the SV parameters offer enough flexibility).

msfschaffner:
The `CI (Verilated English Breakfast (Execute))` failure is unrelated and currently known broken. 
It should not gate this PR from getting merged.

a-will (approved):

PR #21994, opened 2024-03-12
[top_earlgrey] Double the ROM size to 64kB 

msfschaffner:
This PR stages the change for doubling the ROM size - it is not decided yet whether we are really going to accept this change, though.

Note that only the last commit is relevant in this PR, since it is dependent on #21992.

CC @meisnere @arnonsha @sha-ron


hw/top_earlgrey/data/top_earlgrey.hjson
```diff
@@ -747,15 +747,15 @@
...
+      base_addrs: {rom: "0x00010000", regs: "0x411e0000"},
       memory: {
         rom: {
           label:              "rom",
           swaccess:           "ro",
           data_intg_passthru: "true",
           exec:               "True",
           byte_write:         "False",
-          size:               "0x8000",
+          size:               "0x10000",
```
msfschaffner, line 758: This and the edits above in this file are the only changes in this PR - the rest is all autogenerated!

hw/top_earlgrey/data/autogen/top_earlgrey.gen.hjson
```diff
@@ -7268,7 +7268,7 @@
       }
       base_addrs:
       {
-        rom: 0x00008000
+        rom: 0x00010000
```
timothytrippel, line 7271: is it possible/would it be better to start the ROM at 0x0, then the rv_dm mem would not move?

a-will: I think we prefer not to have anything map to 0x0, so it can be an obvious failure for NULL references.

cfrantz: Where is the RvDM RAM located?  I seem to recall that it is at 0x10000.

a-will: With this PR, it was moved up to `0x00080000`. I'm surprised there aren't any obvious signs of changes to the DM firmware or some Ibex vector or something, though... (nothing I spotted, at least)

a-will: Ah, it plugs in via a parameter, which references the xbar / tlul generation script outputs:

https://github.com/lowRISC/opentitan/blob/3152b444b36843d4f4e8769643fb29b53f347e1f/hw/top_earlgrey/rtl/autogen/top_earlgrey.sv#L118-L119

msfschaffner:
We decided not to move forward with this change.

PR #21998, opened 2024-03-12
[dv,top_level] Fix chip_sw_sysrst_ctrl_ulp_z3_wakeup test

matutem:
The UVM code in the test sequence should not wait before updating the phase memory location via backdoor when transitioning to phase 3, or the update can be too late and the test will fail.

More details: In phase 2 the SV side causes sysrst_ctrl to detect the lid opened, which trigger a wakeup while the CPU is in deep sleep. SV then calls sync_with_sw, which delays updating the phase after the CPU enters test_main, and can cause phase 2 to rerun. Normally this test passes, but the delay can be configured via sim_dv runtime plusargs and set large enough for this to happen, and that is what happens for this failure.

The wait in phase 2 is unnecessary since the chip is asleep and opening the lid causes it to wakeup, so updating the phase can be done anytime before the CPU restarts running the test code.

### DIFF
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_sw_sysrst_ctrl_ulp_z3_wakeup_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_sw_sysrst_ctrl_ulp_z3_wakeup_vseq.sv
@@ -2,14 +2,19 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// Test hand shake sequence between sv and c
-// 1. sv: `write_test_phase`(PHASE)
-//    and waiting for test status change from c
+// Running `write_test_phase` from SV causes the C side to perform
+// the checks for the new phase, so any preparation from SV needs to be
+// done prior to updating the phase. Therefore the test sequences the sv
+// and c side as follows:
+// 1. sv: Prepare any external settings,
+//        `write_test_phase`(PHASE),
+//        wait for test status change from c.
 // 2. c: check kTestPhase in switch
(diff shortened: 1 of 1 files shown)


pamaury:
I am a bit suspicious of this change, wouldn't it be better to write the test phase before driving the pins? This way there is not risk with the update? Something like
```
write_test_phase(PHASE_DRIVE_ZERO);
drive_zero_pads();
```
Probably repeat for the various phases.

sha-ron (commented):
I guess the delay in sync_with_sw() was added to allow some more time for the system to stabilize after reset/deep sleep and it might be needed for certain seeds.
I am not  sure I fully understand the test but shouldn't write_test_phase(PHASE_WAIT_WAKEUP); be before sync_with_sw() below it?
@matutem can you please indicate at which stage of the sequence the test failed?

matutem:
Sorry I have not responded: I am dealing with other more urgent issues...

matutem:
Regarding the ordering of write_test_phase, the C side will check for any side-effect of the current step, and then just wait for the test phase to change. This means any preparation for the next step should be done while the CPU is idle waiting, which is the current order of events. I updated the UVM sequence comments.

As for why the delay was there, I really see no good reason, and if some wait was needed for some particular step I think it would be cleaner to add it for that step only, with a comment explaining why. The Nuvoton team had problems with this test that required this change.

Clearly this change only affects sim_dv.

pamaury (approved):
I will approve since you have probably spent a lot of time with this and understand the test better than us. That being said, I am still quite confused by the fix: since the setup is done before changing the phase, I would not expect this delay to have any influence on the test.

sha-ron (approved):

matutem:
Unrelated failures. Merging.

PR #21999, opened 2024-03-12
[dif,flash_ctrl] Avoid reading potentially uninitialized memory

matutem:
Initialize returned objects from objects built by initialization to avoid read-modify-write on potentially uninitialized memory. Without this some tests were getting ECC errors from uninitialized sram. Specifically this addresses failures Nuvoton saw for the chip_sw_pwrmgr_sysrst_ctrl_reset test.


pamaury:
The dif_flash_ctrl unittest seems to be failing, it looks related

matutem:
@pamaury I fixed the unittest failures.

PR #21682, opened 2024-02-26
[sival, rv_dm] implement rv_dm_csr_rw and rv_dm_mem_access

nbdd0121:
Fix #19877
Fix #19878

PR #22002, opened 2024-03-13
Revert "[sival, rv_dm] implement rv_dm_csr_rw and rv_dm_mem_access"

jwnrt:
Reverts lowRISC/opentitan#21682.

This broke the verilated english breakfast CI job.


jwnrt:
Fixed by #22004 instead.

PR #22006, opened 2024-03-13
[uart,dv] Correct uart time unit mishaps & typos

elliotb-lowrisc:
Fix the incorrect naming and usage of time-related variables. Mostly this stems from the `_ns` suffix of `uart_clk_period_ns`, despite being assigned a `realtime` value rather than an integer number of nanosecond. Correct this by renaming to `uart_clk_period` and adjusting downstream usage as seemed most appropriate.

Also correct some comment typos in the wider uart DV code.


rswarbrick (approved):
Thanks for the tidy-up!

PR #21379, opened 2024-02-15
Use LLD for linking

luismarques:
This is something that we already wanted to enable, to support LTO, ePIC, etc.
Enabling this now also solves a problem when upgrading the Rust toolchain.

PR #22008, opened 2024-03-13
Prepare for LLD

luismarques:
To make cherry-picking and potential reverts easier, this PR extracts and applies the commits from #21379 that ensure LLD compatibility but doesn't yet enable the use of LLD. The former commit will then just enable LLD.

Thanks to @a-will for his suggestion on how to fix the linker script to make it compatible with LLD while keeping it compatible with the BFD linker.

### DIFF
--- a/sw/device/lib/testing/test_framework/ottf_common.ld
+++ b/sw/device/lib/testing/test_framework/ottf_common.ld
@@ -257,33 +257,20 @@ SECTIONS {
    _non_volatile_counter_3_start = _non_volatile_counter_3_end - _non_volatile_counter_size;
 
   /**
-   * Non-volatile scratch area in flash.
-   */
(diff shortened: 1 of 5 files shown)


sw/device/lib/testing/test_framework/ottf_common.ld
```diff
@@ -274,28 +272,39 @@ SECTIONS {
    * twice, tests that require non-volatile counters can use this section by
    * striking flash words to count up to `_non_volatile_counter_max_count`.
    */
-  .non_volatile_counter_0 _non_volatile_counter_0_start (NOLOAD) : ALIGN(2048) {
```
pamaury, line 277: Maybe add a comment here that those sections were reordered so that the order corresponds to memory order?

pamaury (approved):

jwnrt (approved):
LGTM, though my experience with linker scripts is limited

PR #22010, opened 2024-03-13
[dv,usbdev] Initial coverage for usbdev

alees24:
Initial set of covergroups and coverpoints for usbdev DV.
Disclaimer: coverage DV is very unfamiliar at this point, and none of the DV code collects samples presently.

### DIFF
--- a/hw/ip/usbdev/dv/env/usbdev_env_cov.sv
+++ b/hw/ip/usbdev/dv/env/usbdev_env_cov.sv
@@ -2,20 +2,247 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-/**
- * Covergoups that are dependent on run-time parameters that may be available
- * only in build_phase can be defined here
- * Covergroups may also be wrapped inside helper classes if needed.
- */
-
 class usbdev_env_cov extends cip_base_env_cov #(.CFG_T(usbdev_env_cfg));
   `uvm_component_utils(usbdev_env_cov)
 
-  // the base class provides the following handles for use:
-  // usbdev_env_cfg: cfg
+  // Have observed all PIDs about which we care; a number of the available PIDs are ignored
+  // by the USB device because they related only to High Speed operation and/or hubs.
+  covergroup pids_cg with function sample(pid_type_e pid);
+    cp_pid: coverpoint pid {
+      bins sof   = {PidTypeSofToken};
+      bins setup = {PidTypeSetupToken};
+      bins out   = {PidTypeOutToken};
+      bins in    = {PidTypeInToken};
+      bins data0 = {PidTypeData0};
+      bins data1 = {PidTypeData1};
+      bins ack   = {PidTypeAck};
+      bins nak   = {PidTypeNak};
+      // Not directly relevant to USBDEV function, but we must appropriately ignore it and the
+      // ensuing Low Speed traffic.
+      bins pre   = {PidTypePre};
+    }
+  endgroup
 
-  // covergroups
-  // [add covergroups here]
+  // Have observed all significant/extreme frame numbers being reported via CSRs; this just
+  // ensures that all bits have been decoded and presented; the frame number has no significance
+  // to USBDEV but it's important to software.
+  covergroup framenum_rx_cg with function sample(bit [10:0] frame);
+    cp_frame: coverpoint frame {
+      bins zero = {0};
+      bins one  = {1};
+      bins range_2_to_15 = {[2:15]};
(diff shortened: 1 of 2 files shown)


hw/ip/usbdev/dv/env/usbdev_env_cov.sv
```diff
@@ -2,20 +2,153 @@
...
- * Covergoups that are dependent on run-time parameters that may be available
- * only in build_phase can be defined here
- * Covergroups may also be wrapped inside helper classes if needed.
- */
-
 class usbdev_env_cov extends cip_base_env_cov #(.CFG_T(usbdev_env_cfg));
   `uvm_component_utils(usbdev_env_cov)
 
   // the base class provides the following handles for use:
   // usbdev_env_cfg: cfg
```
rswarbrick, line 9: Nit: I'd be tempted to drop this noise while we're in the file :-)

hw/ip/usbdev/dv/env/usbdev_env_cov.sv
```diff
@@ -2,20 +2,153 @@
...
+    }
+    cp_out_enable:     coverpoint out_enable;
+    cp_in_enable:      coverpoint in_enable;
+    cp_rxenable_setup: coverpoint rxenable_setup;
+    cp_rxenable_out:   coverpoint rxenable_out;
+    cp_set_nak_out:    coverpoint set_nak_out;
+    cp_out_iso:        coverpoint out_iso;
+    cp_in_iso:         coverpoint in_iso;
+
+    cr_pid_x_epconfig: cross
```
rswarbrick, line 67: I think we're getting 3*2^7 = 384 bins here. That's not an enormous number, but I thought it might be worth checking: do all of these features make sense for each pid type?

hw/ip/usbdev/dv/env/usbdev_env_cov.sv
```diff
@@ -2,20 +2,153 @@
...
+      cp_avout,
+      cp_rx,
+      cp_pid;
+  endgroup
+
+  // Length of DATA packets in each direction
+  covergroup data_pkt_cg with function sample(bit dir_in, bit [6:0] pkt_len);
+    // Packets are read/written from/to memory as 32-bit quantities.
+    cp_pkt_len: coverpoint pkt_len {
+      bins lens[] = {0, 1, 2, 3, 4, 5, 63, 64};
```
rswarbrick, line 115: I *think* this is equivalent to `bins lens[] = {[0:5], [63:64]};`, which might be a bit less typing.

hw/ip/usbdev/dv/env/usbdev_env_cov.sv
```diff
@@ -2,20 +2,153 @@
...
+  endgroup
+
+  // All of SETUP, IN, and OUT for each endpoint
+  covergroup pid_type_endp_cfg with function sample(pid_type_e pid, bit [3:0] endp);
+    cp_pid: coverpoint pid {
+      bins pkt_types[] = {PidTypeSetupToken, PidTypeOutToken, PidTypeInToken};
+    }
+    cp_endp: coverpoint endp {
+      bins endpoints[] = {[0:11]};
+    }
```
rswarbrick, line 150: Does it make sense to also have an "invalid endpoint" bin for values 12-15?

rswarbrick (approved):
A couple of nitty questions, but I think I'm convinced by the general idea. Thanks for working through this.

hw/ip/usbdev/dv/env/usbdev_env_cov.sv
alees24: I've added explanatory comments to the plan now, but your question made be realize that it's probably safe to separate the IN and OUT side configuration bits and test them independently at least.

There are a lot of configuration bits for each endpoint, some of them constant throughout the operation of the USBDEV but some of them changing dynamically in response to traffic, and the concern is that there may be logic expressions within the USB device that are not properly qualified with all of the pertinent configuration bits.

The plan is to hit all of these combinations using a directed test that runs through all permutations, doing a small amount of traffic with the endpoint in each of its possible configurations and check the response/non-response against predictions.

hw/ip/usbdev/dv/env/usbdev_env_cov.sv
alees24: Explanatory comment added here too.

hw/ip/usbdev/dv/env/usbdev_env_cov.sv
alees24: Yes, indeed. We do already test this at top level to some degree, but it's a good idea.

PR #21543, opened 2024-02-18
[RV_DM]rv_dm_abstractcmd_status_vseq

BasitMehmood:
This commit contains test sequence and other related files of #21381 testplan entry.

PR #22016, opened 2024-03-13
[rv_dm,dv] Fix typo in rv_dm_abstractcmd_status_vseq

rswarbrick:
With this bug, the code failed on every execution when it tried to access the (nonexistent) abstractcmd[10].

@BasitMehmood: This is causing the failing nightly. It looks like you never actually ran the code. I didn't notice the mistake when reviewing #21543.


BasitMehmood:
> With this bug, the code failed on every execution when it tried to access the (nonexistent) abstractcmd[10].
> 
> @BasitMehmood: This is causing the failing nightly. It looks like you never actually ran the code. I didn't notice the mistake when reviewing #21543.

Oh i am so sorry. I ran the code several times on synopsys and it was passing. I think before synopsys gave us git access, we used to do PRs from our local server. So we had to move our code from synopsys to local machine for PR. I think then this typo occur in doing so. I am regretful for that.

vogelpi (approved):

PR #22018, opened 2024-03-13
[doc] Make it clear that DEV -> RMA and PROD -> RMA is allowed

jesultra:
Transitions from either DEV or PROD to RMA are the intended way to use the RMA state.  The documentation falsely stated that we attempted to make them hard.

### DIFF
--- a/hw/ip/lc_ctrl/doc/theory_of_operation.md
+++ b/hw/ip/lc_ctrl/doc/theory_of_operation.md
@@ -355,7 +355,7 @@ The purpose of this encoding is to ensure the following
 
 - It is difficult to jump from PROD / PROD_END / SCRAP into DEV
 - It is difficult to jump from DEV / PROD / PROD_END / SCRAP into TEST*
-- It is difficult to jump from DEV / PROD / PROD_END / SCRAP into RMA
+- It is difficult to jump from PROD_END / SCRAP into RMA
 
 Further, the encoding has been chosen to minimize the probability of successful glitch attacks attempting to alter the value of bits in the life cycle state.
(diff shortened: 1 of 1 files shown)


hw/ip/lc_ctrl/doc/theory_of_operation.md
```diff
@@ -355,7 +355,7 @@ The purpose of this encoding is to ensure the following
 
 - It is difficult to jump from PROD / PROD_END / SCRAP into DEV
 - It is difficult to jump from DEV / PROD / PROD_END / SCRAP into TEST*
-- It is difficult to jump from DEV / PROD / PROD_END / SCRAP into RMA
+- It is difficult to jump from PROD_END / SCRAP into RMA
```
timothytrippel, line 358: I think this is supposed to be as it was. The idea is that the actual encoding of the LC FSM states is such that it is difficult to glitch the chip to think it is in a different LC state than it is actually in. Since transfers from DEV and PROD to RMA require a token, we want to make this a difficult transition to glitch. Similarly, once a chip is in the PROD_END / SCRAP states, since these are terminal states, we want to make it difficult for the chip be glitched into thinking it is now in RMA. 

Does that make sense?
CC: @msfschaffner

jesultra: I see, some of the other transitions mentioned are not allowed at all, so it makes sense that what is supposed to be "difficult" here, is to glitch the chip between the states.  This section does not describe, as I initially thought, what transitions can be made knowing the appropriate tokens.

PR #19270, opened 2023-07-21
DRAFT [usb_aon] Deassert pullups if SENSE lost when active

alees24:
This (resurrected) PR includes two proposed changes to the 'usbdev_aon_wake' module:

- Deassert any active pullups if VBUS/SENSE is lost through disconnection when AON/Wakeup module is active.
Avoid potential confusion between host and device, and give the software longer to reinitialize when waking. Related to #18562.

- Introduce explicit reporting of the 'bus not idle' event, which previously had to be inferred by the absence of any other reported reason for waking from Deep Sleep.

Regarding the sequence of operations when resuming from Deep Sleep, software should surely consult the 'wake_events' register before deciding what to do. If a Disconnect has been detected and reported, then the bus and open pipes must be regarded as being in an indeterminate state, and it seems preferable for the aon_wake module to deassert the pull ups to prevent automatic reconnection and then leave software to decide how to resume communications upon return from sleep.

Both changes give software greater visibility into events occurring on the bus and better error handling/recovery in the event that bus signaling does not proceed as expected. For example, software presently has no way to detect the USB becoming non-idle during a transition to a sleep state. Whilst the USB protocol specification requires that resume signaling have a minimum duration long enough for the chip to sleep and resume from sleep, relying upon the correct USB behavior being observed seems undesirable.

@a-will

PR #22019, opened 2024-03-13
[usb,aon_wake] Report 'bus not idle' status to usbdev

alees24:
The 'bus not idle' condition had to be inferred by software, from the absence of Bus Reset and Disconnect indications; less than ideal from a robustness perspective even when returning from Deep Sleep, but it afforded no opportunity to indicate a non-idle bus state whilst still suspending/entering sleep. Introduce an explicit indication for increased robustness and error recovery.

This has been separated out from PR #19270 because of the more contentious, still-to-be-resolved change to pull up behavior and handover sequencing in that PR.


pamaury (approved):
Looks good to me. I concur that having this bit seems more robust than trying to infer something that could turn out to be wrong.

PR #22021, opened 2024-03-13
[kmac] Switch PRNG stream cipher primitive from Trivium to Bivium

vogelpi:
To reduce area, this PR switches the stream cipher primitive inside the PRNG of KMAC from Trivium to Bivium. It was found that the security posture of Bivium is sufficient in this application scenario. This should give us back around 10 kGE. 

Please don't merge this yet. The SCA evaluation is currently running and I will update the PR with the results.

### DIFF
--- a/hw/ip/kmac/data/kmac.hjson
+++ b/hw/ip/kmac/data/kmac.hjson
@@ -155,7 +155,7 @@
     }
     { name:    "NumSeedsEntropy",
       type:    "int",
-      default: "9",
+      default: "6",
       desc:    "Number of words for the PRNG seed used for entropy generation",
       local:   "true"
     }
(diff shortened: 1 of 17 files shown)


johannheyszl (approved):

andreaskurth (approved):
LGTM from an arch & impl perspective. I leave the sec review to you and Johann. Thanks!

msfschaffner (approved):

vogelpi:
The current results I have look okay. To be on the safe side, I'll have to repeat some more experiments, too. But this can happen later. I am confident that this is okay. Merging!

vogelpi:
Update: I've now also finished the FPGA TVLA experiments for SHA3 and KMAC. The results do not significantly change between using Trivium or Bivium. So, it's all good :-)

cdgori (commented):
Catching up - LGTM and this great results!

PR #22029, opened 2024-03-14
[usbdev,dv] Specify num_of_bytes for call to configure_in_trans

rswarbrick:
This is in usbdev_in_stall_vseq, where we don't actually expect to transfer any data (we're going to stall!). So specify any old value: 10 is pretty arbitrary.

This fixes a silly bug that I introduced by merging #21926, which conflicts with some base sequence tidy-ups I've done in the last couple of days. Sorry!


rswarbrick:
Fixed separately in #22032.

PR #22030, opened 2024-03-14
[usbdev,dv] adds inter_pkt_delay() task in usbdev_base_seq

MubashirSaleem775:
Introduces an inter-packet delay task in the base sequence, which will be utilized between packets in all sequence transactions.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -284,4 +284,13 @@ virtual task call_sof_seq(input pkt_type_e pkt_type, input pid_type_e pid_type);
   start_item(m_sof_pkt);
   finish_item(m_sof_pkt);
 endtask
+
+virtual task inter_packet_delay(int delay = 0);
+  // From section 7.1.18.1 Time interval between packets is between 2 to 6.5 bit times.
+  // Multiply with 4 because usb  samples on 4 clks.
+  if (delay == 0) delay = $urandom_range(8, 28);
+  else delay = delay * 4;
+  // for max/min inter pkt delay it can be changed with task argument
+  cfg.clk_rst_vif.wait_clks(delay);
+endtask
 endclass : usbdev_base_vseq


MubashirSaleem775:
@rswarbrick please review this PR.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -284,4 +284,14 @@ virtual task call_sof_seq(input pkt_type_e pkt_type, input pid_type_e pid_type);
   start_item(m_sof_pkt);
   finish_item(m_sof_pkt);
 endtask
+
+virtual task inter_packet_delay(int delay = 12);
```
rswarbrick, line 288: Would it make sense to change this argument so that it's measured in bit times? Then we could end up calling `wait_clks(delay * 4)`.

The reason I'm asking is that I *think* there's never going to be a reason to call this function with a delay that isn't an integer number of bit times.

MubashirSaleem775: Okay, it make sense. i will change this argument as you suggested. Thanks

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -284,4 +284,11 @@ virtual task call_sof_seq(input pkt_type_e pkt_type, input pid_type_e pid_type);
   start_item(m_sof_pkt);
   finish_item(m_sof_pkt);
 endtask
+
+virtual task inter_packet_delay(int delay);
+  // The interpacket delay ranges from 2 to 18 bit times.
```
alees24, line 289: This range has confused two different timing parameters within the specification. From section 7.1.18.1 the 'Inter-packet Delay' which is the time interval from the end of one packet to the start of the next packet _in the same direction_, must be at least 2 bit times and at most 6.5 bit times at the device port.
This means that the task needs to have a 'delay' value in the range [2,6] inclusive if specified in terms of usb20_driver clocks (= bit intervals).

The number 18 comes from the maximum time interval for the device to wait before deciding to timeout and give up on receiving a response from the host eg. with ACK or NAK in response to a DATA packet that the USB device sent as an IN transmission.

I think it would also be worth having this task randomize by default so that everywhere that it is enabled may benefit from the extra testing. If a specific, fixed delay is required then the code that enables the task may override the default.

Thanks.

alees24 (changes requested):
Please see comment.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
MubashirSaleem775: Thank you for your review. I have made the necessary updates based on your feedback.

alees24 (changes requested):
This looks okay thank you except for the use of 6.5 here, which will sometimes yield a delay of 7 cycles because randomizing across a range of 'real' values yields a result of type 'real' which is then rounded to nearest when casting to an integral data type.

Since the code is about to multiply by 4 to achieve USB device clocks, randomizing with $urandom_range(8, 26) in the case where 'delay' has defaulted to 0 would work, please.

alees24 (approved):

PR #22037, opened 2024-03-14
[veriblelint] Update to v0.0-3622-g07b310a3

msfschaffner:
The version we have been using was quite outdated and did not catch some legacy always @(*) statements that it should have warned about.

This updates to v0.0-3622-g07b310a3, and updates the installation instructions.

We may want to consider holding off merging, though, if this causes any infrastructure issues.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,7 @@ variables:
   #
   VERILATOR_VERSION: 4.210
   TOOLCHAIN_PATH: /opt/buildcache/riscv
-  VERIBLE_VERSION: v0.0-2135-gb534c1fe
+  VERIBLE_VERSION: v0.0-3622-g07b310a3
   # Release tag from https://github.com/lowRISC/lowrisc-toolchains/releases
   TOOLCHAIN_VERSION: 20220210-1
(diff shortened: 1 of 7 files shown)


a-will:
Need to update this too: https://github.com/lowRISC/opentitan/blob/e844018f2c29ee58e95340e264f519dbc8271c80/ci/install-package-dependencies.sh#L99

msfschaffner:
> Need to update this too (no more Ubuntu-specific builds):
> 
> https://github.com/lowRISC/opentitan/blob/e844018f2c29ee58e95340e264f519dbc8271c80/ci/install-package-dependencies.sh#L99

Good catch - fixed!

msfschaffner:
It appears that Private CI will fail until we merge a corresponding fix over there: https://github.com/lowRISC/opentitan-private-ci/pull/35

This is because the URL has changed for newer releases, and Private CI relies on scripts from the main repo.

timothytrippel (approved):

andreaskurth:
Reopening as I think this got closed accidentally. (If you disagree, please close again.)

andreaskurth:
Merging to keep public CI in sync with private CI. Bypassing failed checks as this addresses a CI problem on master.

Issue #15123, opened 2022-09-26
[rv_dm,rtl] Fix acknowledgment of reset in `dmstatus.allhavereset`/`dmstatus.anyhavereset`


andreaskurth:
https://github.com/pulp-platform/riscv-dbg/issues/118:
> Currently the debug module seems to assume any assertion of `ndmreset_o` will instantly cause a reset. The `dmstatus.allhavereset`/`dmstatus.anyhavereset` bits are set purely on `ndmreset_o` being asserted rather than any acknowledgement a reset has occurred.

This is incorrect behavior and should be fixed, but it is currently not urgent for us, because the [`rstmgr` resets the system reset domain](https://docs.opentitan.org/hw/ip/rstmgr/doc/index.html#system-reset-tree) unconditionally whenever `ndmreset_req` is asserted (a few cycles after asserting this signal).

> estimate 12

andreaskurth:
Triaged for `rv_dm`. Assigning to M2.5 with https://github.com/lowRISC/opentitan/labels/Priority%3AP3 because there are only a few clock cycles (between the assertion of `ndmreset_o` and the instant `rstmgr` resets the system) during which those bits are not correct -- so I think this can be easily worked around in software without noticeable impact, whereas the RTL changes involved in proper reset acknowledgment are out of scope for this release IMHO.

msfschaffner:
Depending on how we intend to implement the delayed debug unlock feature, this may be something that we have to fix as well.

CC @rswarbrick

PR #22038, opened 2024-03-14
[rv_dm] WARL behavior and NDM tracking fixes

### LINKED ISSUE #19750: [rv_dm,rtl] Potential issue with hartsel WARL behaviour
### Description

 The debug spec states:

> While the spec allows for 20 hartsel bits, an implementation may choose to implement fewer than that. The actual width of hartsel is called HARTSELLEN. It must be at least 0 and at most 20. A debugger should discover HARTSELLEN by writing all ones to hartsel (assuming the maximum size) and reading back the value to see which bits were actually set.

The rv_dm implementation implements the full 20 bits and doesn't appear to limit the number you can write based upon the number of harts available. For OpenTitan with a single hart I believe this field should act as a read-only field with a constant value of 0. Instead you can write any 20 bit number there and then read it back, any number other than 0 will select a non-existent hart. 

The above is based on RTL inspection rather than testing the behaviour so I may be missing something. I am uncertain what the consequences of using rv_dm functionality against a non-existent hart is.

This is not high priority but could cause issues with certain tools that are doing some kind of autoprobe to determine what harts are available. It should be a simple fix so likely worth doing for prod.

msfschaffner:
This fixes #19750 and #15123.

TODO:  we should file issues for BLT/TLT extensions.
@rswarbrick can you take care of the BLT part? 
I filed an issue for TLTs here: #22044


hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -268,16 +273,80 @@ module rv_dm
...
+        ndmreset_pending_q <= 1'b1;
+      end else if (ndmreset_ack && ndmreset_pending_q) begin
+        ndmreset_pending_q <= 1'b0;
+      end
+      // We only track lc resets that are asserted during an active ndm reset request..
+      if (ndmreset_pending_q && lc_rst_asserted) begin
+        lc_rst_pending_q <= 1'b1;
+      end else if (ndmreset_ack && lc_rst_pending_q) begin
+        ndmreset_pending_q <= 1'b0;
+      end
```
a-will, line 335: Was this supposed to be...

```suggestion
      if (ndmreset_pending_q && lc_rst_asserted) begin
        lc_rst_pending_q <= 1'b1;
      end else if (ndmreset_ack && lc_rst_pending_q) begin
        lc_rst_pending_q <= 1'b0;
      end
```

msfschaffner: yes, absolutely! this is still a bit drafty, need to go through the CI failures...

msfschaffner: Fixed!

a-will (approved):

msfschaffner:
If anyone with an FPGA setup could help me debug the `rv_dm_access_after_wakeup_*` SiVal non-ROM_EXT tests that would be appreciated.

rswarbrick:
@msfschaffner: About the testing side of things: I've filed an issue (#22048) to track the work that needs doing. Assigning it to a PROD milestone is clearly the next job! I assume M4, but will leave that to someone with more knowledge of the roadmap/timescale.

msfschaffner:
> @msfschaffner: About the testing side of things: I've filed an issue (#22048) to track the work that needs doing. Assigning it to a PROD milestone is clearly the next job! I assume M4, but will leave that to someone with more knowledge of the roadmap/timescale.

Thanks! PROD.M4 sounds right!

PR #22040, opened 2024-03-14
[spi_host,rtl] Use sd_en_o[0] to only drive sd_o[0] during TX (std_mode)

### LINKED ISSUE #14969: [spi_host] output line during RX only segments
It appears that when the spi_host is configured for an RX only segment, its TX line (this is a standard single lane scenario) may just toggle garbage. 

In `spi_host_tx_rx_test`, the test transmits 4 bytes, does 60 bytes of TX/RX, and then finally 4 bytes of RX.
During the final 4 bytes of receive, the TX line appears to just shift a combination of what was already in the shift register plus new bits being shifted in (this is a nice optimization to re-use to same shift register for both TX/RX).

Generally speaking, there is nothing wrong with this behavior.  For something like a spi flash protocol, during the cycles where we are reading back, no one should care what's on the TX line. 

However, it's been pointed out that there are a lot of non-compliant devices out there, and it would be better to not try our luck with random toggling on the line because we don't know for sure what we can expect. 

The suggestion then is to simply stop driving the line when there is nothing to drive. This will ensure we don't needlessly toggle, but also that we do not come into conflict with the other side (since this period is don't care...we can't say for the other device is not actively contending on the line as well).

It's not immediately clear how high a priority this is, but it feels like something that is FutureRelease to me. 

> estimate 8

hcallahan-lowrisc:
Redux of #21338

This is effectively the same change, but fixes the problem as diagnosed/described [in this comment](https://github.com/lowRISC/opentitan/pull/21338#issuecomment-1984908004).

Closes #14969 (again)

### DIFF
--- a/hw/ip/spi_host/rtl/spi_host_fsm.sv
+++ b/hw/ip/spi_host/rtl/spi_host_fsm.sv
@@ -556,6 +556,25 @@ module spi_host_fsm
     end
   end : gen_csb_gen
 
+  // When 'cmd_wr_en_q' deasserts, we need to stretch the enable used
+  // to drive sd_en_o[0] until the end of the bit period. This is because
+  // 'cmd_wr_en_q' can change up to 1/2 cycle before the end of the period
+  // (depending on cpol/cpha).
+  // The signal 'last_bit' is active in the final period, so use that to start
+  // stretching the enable, and 'byte_ending' marks the period end.
+  logic cmd_wr_en_last_bit;
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      cmd_wr_en_last_bit <= 1'b0;
+    end else if (cmd_wr_en_q & last_bit) begin
+      cmd_wr_en_last_bit <= 1'b1;
+    end else if (byte_ending) begin
+      cmd_wr_en_last_bit <= 1'b0;
+    end else begin // Stretching..
+      cmd_wr_en_last_bit <= cmd_wr_en_last_bit;
+    end
+  end
+
   assign csb_o = csb_q;
 
   always_comb begin
@@ -564,7 +583,8 @@ module spi_host_fsm
     end else begin
       unique case (speed_o)
         Standard: begin
-          sd_en_o[0]   = 1'b1;
+          // Observing 'last_bit' ensures we do not deassert the enable too early
+          sd_en_o[0]   = cmd_wr_en_q | cmd_wr_en_last_bit;
           sd_en_o[1]   = 1'b0;
           sd_en_o[3:2] = 2'b00;
         end


hcallahan-lowrisc:
Chip level test failure (`//sw/device/silicon_creator/rom/e2e/jtag_inject:sram_program_fpga_cw310_test_otp_rma_fpga_cw310_rom_with_fake_key`) would appear to be unrelated.

hw/ip/spi_host/rtl/spi_host_fsm.sv
```diff
@@ -564,7 +583,8 @@ module spi_host_fsm
     end else begin
       unique case (speed_o)
         Standard: begin
-          sd_en_o[0]   = 1'b1;
+          // Observing 'last_bit' ensures we do not deassert the enable too early
+          sd_en_o[0]   = cmd_wr_en_q | cmd_wr_en_last_bit;
```
a-will, line 587: Curious: Is this actually a special property of the `Standard` "speed"? Or is it more that we don't test `Dual` and `Quad` writes?

hcallahan-lowrisc: I suspect it is the latter. I will peek at some waves and check

andreaskurth (approved):
The RTL change makes sense to me, and if you've verified that it solves #14969 without breaking other tests, this LGTM

hw/ip/spi_host/rtl/spi_host_fsm.sv
hcallahan-lowrisc: I spent some time looking at this, and the block level testbench seems to be fairly inflexible at generating transactions where we are switching directions between segments. The constraints in `spi_segment_item.sv` mean we only generate transactions which are fairly representative of standard spi operations, and don't e.g. change direction during the DATA phase.
I hacked around a bit and haven't been able to generate stimulus which might generate the suspected behaviour. However I still think it might be possible. My feeling is that is not likely to happen when constructing standard SPI commands with the prescriptive CMD->[DUMMY]->DATA structure of segments. The chip level test that got broken by the first PR does a slightly unusual `~CSB -> Tx/Std -> Bidir/Std -> Rx/Std -> CSB` which is not stimulus our block level environment would currently create.
I'm going to merge this PR as-is, and create an issue to come back and address this properly for V2.

hcallahan-lowrisc:
CI failures are known and unrelated to this PR. Hence merging.

hw/ip/spi_host/rtl/spi_host_fsm.sv
a-will: Thanks @hcallahan-lowrisc !

FYI, the "standard SPI operations" of interest might be DUAL I/O and QUAD I/O read types, like this:

```
QUAD I/O READ (1-4-4):
CMD (1x OUT) -> ADDR (4x OUT) -> DUMMY -> DATA (4x IN)
```

hcallahan-lowrisc: Yes the block level DV just smooshes the first CMD+ADDR into one segment only. Extending to the 'I/O' ops will be useful, as well as more random combinations of segments.
https://github.com/lowRISC/opentitan/blob/e3ca274e77a7f2bf22f6d460f7e14734f32f1570/hw/ip/spi_host/dv/env/spi_host_env_pkg.sv#L51-L55

PR #22041, opened 2024-03-14
[entropy_src] Rework swread packer FIFO, reduce esfinal FIFO depth

vogelpi:
This PR contains two separate commits to reduce the overall area footprint of ENTROPY_SRC by reducing the number of full 384-bit seeds buffered in the design:
- The first commit reduces the depth of the esfinal FIFO from 4 to 3. This should not impact performance as the CSRNG instance connected to ENTROPY_SRC has 3 contexts. This means the esfinal FIFO has still enough space to have one seed ready per CSRNG instance.
- The second commit reworks the swread packer FIFO to not buffer the seed readable by software. This works because whenever SW can read the final seeds, no entropy is forwarded to CSRNG anyway. This means, the seeds can directly read from the esfinal FIFO.

In total, this PR should reduce the area by around 768 FFs.

I've run the `chip_sw_entropy_src_kat_test` chip level test to explicitly verify the swread path still works (second commit) and a block-level regression with a reduced number of seeds is currently running. So far, the results look okay.


vogelpi:
Update: the block-level regression looks fine. This can be merged once it's approved and CI ran through.

msfschaffner (approved):

PR #22042, opened 2024-03-14
[dv,pwrmgr] Fix some test failures

matutem:
This has three commits:
- Fix cip_base_scoreboard function set_exp_alert to avoid forking a time-consuming thread
  since that was giving trouble in VCS and is perhaps at the fringe of the LRM.
- Handle rapid changes in escalation clock gating changes.
- Fix a few miscellaneous failures related to detection of the fast fsm becoming inactive
  and handling changes in rst_lc_req and rst_lc_src_n.


msfschaffner (approved):

hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_base_vseq.sv
```diff
@@ -158,9 +163,9 @@ class pwrmgr_base_vseq extends cip_base_vseq #(
     // There is tb lock up case
     // when reset come while rom_ctrl = {false, false}.
     // So we need rom_ctrl driver runs in parallel with
-    // wait_for_fast_fsm_active
+    // wait_for_fast_fsm(FastFsmActive)
```
rswarbrick, line 166: Nit: I'm not sure you need to explicitly give the argument value in the comment here. (I spy a `sed` call :-D)

rswarbrick (approved):

hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_base_vseq.sv
matutem: Well, the comment is accurate since the call is done with that argument.

PR #22026, opened 2024-03-14
[uart] Reduce TX FIFO depth from 128B to 32B

msfschaffner:
The UART FIFOs had been increased from 32B to 128B for Earlgrey ES, predominantly due to the fact that there was a requirement for 128B deep RX FIFOs. Such a requirement does however not exist for the TX FIFO and it can be just 32B deep.

Since the design only supported a homogeneous arrangement, both RX and TX FIFOs had been set to the same depth. This patch reworks that and makes both depths individually configurable.

PR #22046, opened 2024-03-15
[uart] Reduce RX FIFO depth from 128B to 64B

msfschaffner:
After more discussion with customers, it became clear that a 64B FIFO size is sufficient.

### DIFF
--- a/hw/ip/uart/README.md
+++ b/hw/ip/uart/README.md
@@ -20,7 +20,7 @@ top level system.
 - 2-pin full duplex external interface
 - 8-bit data word, optional even or odd parity bit per byte
 - 1 stop bit
-- 128 x 8b RX buffer
+- 64 x 8b RX buffer
 - 32 x 8b TX buffer
 - Programmable baud rate
 - Interrupt for transmit empty, receive overflow, frame error, parity error, break error, receive
--- a/hw/ip/uart/dv/env/uart_scoreboard.sv
+++ b/hw/ip/uart/dv/env/uart_scoreboard.sv
@@ -248,7 +248,7 @@ class uart_scoreboard extends cip_base_scoreboard #(.CFG_T(uart_env_cfg),
             if (ral.ctrl.slpbk.get_mirrored_value()) begin
               // if sys loopback is on, tx item isn't sent to uart pin but rx fifo
               uart_item tx_item = tx_q.pop_front();
-              if (rx_enabled && (rx_q.size < TxFifoDepth)) begin
+              if (rx_enabled && (rx_q.size < RxFifoDepth)) begin
                 rx_q.push_back(tx_item);
                 predict_rx_watermark_intr();
               end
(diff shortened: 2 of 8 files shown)


msfschaffner:
Can someone run 
```
util/dvsim/dvsim.py -t vcs hw/top_earlgrey/dv/top_earlgrey_sim_cfgs.hjson --select-cfgs uart -i all_once
```
on this PR and post the results? I lost access to my VM for some reason...

If this is all-green except for `uart_stress_all_with_rand_reset` this can be merged (see also https://github.com/lowRISC/opentitan/pull/22026#issuecomment-1996322346).

andreaskurth:
I'm running the regression asked for above; will post the results here.

msfschaffner:
I fixed something and reran the full nightly regression.
There are some sporadic interrupt failures that should be looked at for a V2 signoff, but all tests pass with >90% of seeds except for that stress-all test that is known to fail.
I think we can move forward with this change and patch things up later for the V2 signoff.

hw/ip/uart/README.md
```diff
@@ -20,7 +20,7 @@ top level system.
 - 2-pin full duplex external interface
 - 8-bit data word, optional even or odd parity bit per byte
 - 1 stop bit
-- 128 x 8b RX buffer
+- 64 x 8b RX buffer
```
vogelpi, line 23: Thanks for aligning that!

hw/ip/uart/dv/env/uart_scoreboard.sv
```diff
@@ -248,7 +248,7 @@ class uart_scoreboard extends cip_base_scoreboard #(.CFG_T(uart_env_cfg),
             if (ral.ctrl.slpbk.get_mirrored_value()) begin
               // if sys loopback is on, tx item isn't sent to uart pin but rx fifo
               uart_item tx_item = tx_q.pop_front();
-              if (rx_enabled && (rx_q.size < TxFifoDepth)) begin
+              if (rx_enabled && (rx_q.size < RxFifoDepth)) begin
```
vogelpi, line 251: This looks like a drive-by bug fix to me. Correct?

sw/device/lib/dif/dif_uart.h
```diff
@@ -98,11 +98,7 @@ typedef enum dif_uart_watermark {
   /**
    * Indicates a sixty-four-byte watermark.
```
vogelpi, line 99: Nit: this comment is outdated now.

vogelpi: But to be honest, I would leave it as is.

vogelpi (approved):
LGTM, thanks @msfschaffner !

hw/ip/uart/dv/env/uart_scoreboard.sv
msfschaffner: yes, I think I introduced that bug when I separated the rx/tx depth parameters

moidx (approved):

vogelpi:
I've now rebased this on master as I saw the same unrelated chip-level test failing repeatedly.

PR #22047, opened 2024-03-15
[rv_core_ibex,rtl] Reduce number of performance counters

GregAC:
On top of the RV32I mandatory counters this just leaves the counters for data memory and instruction memory waits. The others are counting different types of instructions being executed so are purely a property of the architectural behaviour of the program so could be found another way (e.g. via simulation) where the wait counters cannot be easily modelled.

This will drop 8 * 32 * 2 (counters are duplicated due to lockstep) = 512 flops out of the design along with associated logic for incrementers and reduce inputs on muxes in the Ibex CSR block.

For reference the details on performance counters in Ibex can be found here: https://ibex-core.readthedocs.io/en/latest/03_reference/performance_counters.html

With this change only events 0-4 from that documentation will have counters.

Before merging we need to confirm this is acceptable for software

@msfschaffner @andreaskurth @vogelpi @jesultra @cfrantz  PTAL

### DIFF
--- a/hw/top_earlgrey/data/autogen/top_earlgrey.gen.hjson
+++ b/hw/top_earlgrey/data/autogen/top_earlgrey.gen.hjson
@@ -7466,7 +7466,7 @@
         PMPEnable: "1"
         PMPGranularity: "0"
         PMPNumRegions: "16"
-        MHPMCounterNum: "10"
+        MHPMCounterNum: "2"
         MHPMCounterWidth: "32"
         RV32E: "0"
         RV32M: ibex_pkg::RV32MSingleCycle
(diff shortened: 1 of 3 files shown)


andreaskurth (approved):
I think this makes a lot of sense to save area, thanks @GregAC. I agree with your reasoning, but it's really important to have buy-in from the SW team as well (like you wrote).

vogelpi (approved):
Good idea @GregAC ! This looks good to me but as Andreas said, let's first discuss this also with the right people before merging it.

jesultra:
I am not aware of any plans by the Ti50 team to use these performance counters, it is fine with us to remove them.

moidx:
This SGTM wrt other use cases on our end for A1.

moidx (approved):

PR #22049, opened 2024-03-15
[i2c,rtl] Halt controller FSM when unexp. NAK is seen

### LINKED ISSUE #18917: [i2c] Change FSM to stop processing the FMT_FIFO after receiving a NAK
### Description

From #18764:
Despite the IP registering a NAK, the transaction continues normally, when it should really stop and wait for software to clear the NAK condition. It turns out that there is nothing in the i2c FSM to prevent continuing with the next entry in the FIFO.

Fix this and document the expected handling from software.

hcallahan-lowrisc:
If the I2C bus NAKs any of our transfers when operating as a Controller-Transmitter in Byte-Formatted Programming Mode, the current behaviour is to proceed to the next FDATA entry as normal, disregarding the NAK entirely. The only side-effect of this event is the block raising the 'nak' interrupt (though this can be suppressed by the use of NAKOK in the FDATA entry).

This PR changes the behaviour so the Controller-FSM halts upon generation of the 'nak' interrupt, and proceeds once the interrupt has been acknowledged using the standard 'Event-Type' interrupt acknowledgement (W1C to INTR_STATE).
The next item in the FMTFIFO is not popped until this acknowledgement is received.
If the NAKOK bit is set for a particular Format Indicator and the bus NAKs our byte, the 'nak' interrupt is not asserted and the FSM continues to the next indicator in the fifo.

While the FSM is halted, SCL is released (we halt in the 'ClockPulseAck' state), and the bus is observed as having an elongated 9th clock period.

Two additional features are also added alongside this change.

- "HOST_NACK_HANDLER_TIMEOUT" CSR
This CSR controls a timeout mechanism where if SW does not handle a pending 'nak'
interrupt before the configured timeout elapses, the hardware will automatically
end the active transaction with a STOP condition and disable Host-Mode.

- "STATUS.HOST_DISABLED_NACK_TIMEOUT" bit
This bit is set when the block times out with the "HOST_NACK_HANDLER_TIMEOUT"
mechanism, and clears when SW re-enables Host-Mode. This is intended to inform
software that the timeout occurred.

The I2C agent / infrastructure does not currently support the generation of NAK's when operating as a Target-Receiver, so this PR also adds rudimentary support for this with a new Agent sequence, plus a test to exercise it (`i2c_host_may_nack`). The testbench currently models the SW interrupt handler latency with a fixed 1000-cycle delay. This should be expanded in the future.

TODO

- [x] Update documentation
- [ ] Check & Update TLTs

---

This new behaviour will need more comprehensive testing to achieve V2 quality. Our infrastructure to test this is currently incomplete, and will need to be expanded. I have observed some of the expected behaviour in an ad-hoc simulation, but it should not be considered fully robust until further testing is completed.

---

Closes #18917

### DIFF
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -204,7 +204,8 @@
     { name: "CTRL"
       desc: "I2C Control Register"
       swaccess: "rw"
-      hwaccess: "hro"
+      hwaccess: "hrw"
+      hwqe:     "true",
       fields: [
         { bits: "0"
           resval: "0"
@@ -283,6 +284,13 @@
           name: "ACQEMPTY"
           desc: "Target mode receive FIFO is empty"
           resval: "1"
+        }
+        { bits: "10"
+          name: "HOST_DISABLED_NACK_TIMEOUT"
+          desc: '''
+                A Host-Mode active transaction has been ended by the !!HOST_NACK_HANDLER_TIMEOUT mechanism.
+                This bit is cleared when !!CTRL.ENABLEHOST is set by software to start a new transaction.
+                '''
         }
       ]
       tags: [// Updated by the hw. Exclude from write-checks.
@@ -726,6 +734,38 @@
       hwaccess: "hrw"
       fields: [
         { bits: "7:0" }
+      ]
+    }
+    { name: "HOST_NACK_HANDLER_TIMEOUT"
+      desc: '''
+            Timeout in Host-Mode for an unhandled NACK before hardware automatically ends the transaction.
+            (in units of input clock frequency)
+
+            In Host-Mode during an active Controller-Transmitter transfer, if the Target NACKs a byte
+            the 'nak' interrupt is asserted and the Byte-Formatted Programming Mode FSM halts awaiting
+            software intervention. Software must acknowledge the interrupt (CIP 'Event-Type') to allow
+            the state machine to continue, typically after clearing out the FMTFIFO to start a new transfer.
+            While halted, the active transaction is not ended (no STOP (P) condition is created), and the
+            block releases both SCL and SDA.
+
+            This timeout can be used to automatically disable the block if software does not handle the
+            'nak' interrupt in a timely manner. If the timeout expires, (!!CTRL.HOSTMODE) will be disabled
+            which creates a STOP (P) condition on the bus ending the active transaction. Additionally, the
+            !!STATUS.HOST_DISABLED_NACK_TIMEOUT bit is set to alert software.
+
+            The enable bit must be set for this feature to operate.
+            '''
+      swaccess: "rw"
+      hwaccess: "hro"
+      fields: [
+        { bits: "30:0"
+          name: "VAL"
+          desc: "Unhandled NAK timeout value (in units of input clock frequency)"
+        }
+        { bits: "31"
+          name: "EN"
+          desc: "Timeout enable"
+        }
       ]
     }
   ]
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -683,7 +683,14 @@
       ]
     }
     { name: "HOST_TIMEOUT_CTRL"
-      desc: "I2C host clock generation timeout value (in units of input clock frequency)"
+      desc: '''
+            I2C host clock generation timeout value (in units of input clock frequency).
+
+            In an active transaction in Target-Mode, if the Controller ceases to send SCL pulses
+            for this number of cycles then the "host_timeout" interrupt will be asserted.
+
+            Set this CSR to 0 to disable this behaviour.
+            '''
       swaccess: "rw"
       hwaccess: "hro"
       fields: [
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
@@ -462,7 +462,10 @@ class i2c_base_vseq extends cip_base_vseq #(
     print_format_flag(item, msg, do_print);
   endtask : program_format_flag
 
-  // read interrupts and randomly clear interrupts if set
+  // Read and clear interrupts
+  //
+  // - Clear all interrupts that are active
+  // - Add a delay of 'clear_intr_dly' cycles before the W1C operation
   virtual task process_interrupts();
     bit [TL_DW-1:0] intr_state, intr_clear;
 
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
@@ -466,6 +466,7 @@ class i2c_base_vseq extends cip_base_vseq #(
   //
   // - Clear all interrupts that are active
   // - Add a delay of 'clear_intr_dly' cycles before the W1C operation
+  // - If 'nak' irq is active, add an extra delay of 1000 cycles
   virtual task process_interrupts();
     bit [TL_DW-1:0] intr_state, intr_clear;
 
@@ -489,6 +490,11 @@ class i2c_base_vseq extends cip_base_vseq #(
     end
     if (bit'(get_field_val(ral.intr_state.tx_stretch, intr_clear))) begin
       `uvm_info(`gfn, "\n  clearing tx_stretch", UVM_DEBUG)
+    end
+    if (bit'(get_field_val(ral.intr_state.nak, intr_clear))) begin
+      // Add a longer delay to mimic a software handler clearing the NAK condition
+      cfg.clk_rst_vif.wait_clks(1_000);
+      `uvm_info(`gfn, "Clearing 'NAK' interrupt, allowing FSM to continue...", UVM_DEBUG)
     end
 
     `DV_CHECK_MEMBER_RANDOMIZE_FATAL(clear_intr_dly)
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_host_may_nack_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_host_may_nack_vseq.sv
@@ -1,0 +1,40 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class i2c_host_may_nack_vseq extends i2c_rx_tx_vseq;
+  `uvm_object_utils(i2c_host_may_nack_vseq)
+  `uvm_object_new
+
+  virtual task i2c_init(if_mode_e mode = Host);
+    super.i2c_init();
+    // Mask some interrupts that are just noise for this test.
+    ral.intr_enable.fmt_threshold.set(1'b0);
+    ral.intr_enable.tx_threshold.set(1'b0);
+    ral.intr_enable.sda_interference.set(1'b0);
+    ral.intr_enable.scl_interference.set(1'b0);
+    ral.intr_enable.sda_unstable.set(1'b0);
+    ral.intr_enable.stretch_timeout.set(1'b0);
+    csr_update(ral.intr_enable);
+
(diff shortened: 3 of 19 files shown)


andreaskurth:
Thanks for working on this, @hcallahan-lowrisc. Two points to consider:
- After we as Controller-Transmitter get a NACK from the Target-Receiver, we may either generate a Stop condition to abort the transfer or a Repeated Start to start a new transfer (I2C spec, section 3.1.6). Just continuing with the next byte isn't an option AFAIK. I think right now SW would be able to disable the Host mode in the I2C IP to generate a Stop condition, so this should be covered, but SW doesn't have a way to decide on a Repeated Start, right?
- Our I2C IP shouldn't stretch the clock indefinitely. If SW fails to handle the interrupt within a certain time, our HW should stop stretching (and as fallback generate a Stop condition, I think). Higher-level protocols such as SMBus define upper limits for clock stretching, so our I2C IP could have a CSR that allows SW to set the upper limit (similar to what was added in PR #21857.

@a-will: When you created #18917, what did you have in mind SW would want to do when it gets the NACK interrupt?

a-will:
> Thanks for working on this, @hcallahan-lowrisc. Two points to consider:
> 
> * After we as Controller-Transmitter get a NACK from the Target-Receiver, we may either generate a Stop condition to abort the transfer or a Repeated Start to start a new transfer (I2C spec, section 3.1.6). Just continuing with the next byte isn't an option AFAIK. I think right now SW would be able to disable the Host mode in the I2C IP to generate a Stop condition, so this should be covered, but SW doesn't have a way to decide on a Repeated Start, right?
> * Our I2C IP shouldn't stretch the clock indefinitely. If SW fails to handle the interrupt within a certain time, our HW should stop stretching (and as fallback generate a Stop condition, I think). Higher-level protocols such as SMBus define upper limits for clock stretching, so our I2C IP could have a CSR that allows SW to set the upper limit (similar to what was added in PR [[i2c,rtl] NACK after timing out on a stretch #21857](https://github.com/lowRISC/opentitan/pull/21857).
> 
> @a-will: When you created #18917, what did you have in mind SW would want to do when it gets the NACK interrupt?

I think SW would effectively need to start with a fresh/empty FMT FIFO, but it might be useful to allow SW to clear it. The intended transfer did not succeed, but perhaps SW would want to know how many bytes made it before the NACK. Then SW can decide whether to issue Stop or Start after clearing the FMT FIFO.

I'd leave SCL high while waiting for SW, though. In a multi-controller context, if SW takes too long and the transaction is forced to end, it might be good to communicate the info to SW. I think in most cases, a Stop would be forthcoming anyway.

hcallahan-lowrisc:
> * After we as Controller-Transmitter get a NACK from the Target-Receiver, we may either generate a Stop condition to abort the transfer or a Repeated Start to start a new transfer (I2C spec, section 3.1.6). Just continuing with the next byte isn't an option AFAIK. I think right now SW would be able to disable the Host mode in the I2C IP to generate a Stop condition, so this should be covered, but SW doesn't have a way to decide on a Repeated Start, right?

So I think we can handle the RSTART case, as we would expect SW to clear the FMT FIFO then add new entries which are popped upon acknowledging the NAK irq. If the first entry added was a START then that [should issue a RSTART](https://opentitan.org/book/hw/ip/i2c/doc/theory_of_operation.html#byte-formatted-programming-mode:~:text=This%20flag%20may%20also%20be%20used%20to%20issue%20a%20repeated%20start%20condition.) on the bus. The STOP case should work as you say by disabling Host Mode. (maybe not so elegant but it does work).

> * Our I2C IP shouldn't stretch the clock indefinitely. If SW fails to handle the interrupt within a certain time, our HW should stop stretching (and as fallback generate a Stop condition, I think). Higher-level protocols such as SMBus define upper limits for clock stretching, so our I2C IP could have a CSR that allows SW to set the upper limit (similar to what was added in PR [[i2c,rtl] NACK after timing out on a stretch #21857](https://github.com/lowRISC/opentitan/pull/21857).

This seems reasonable, I can easily add an additional timeout counter for this that just disables Host mode in this case, which should generate the STOP condition. However, we should probably have a think about how much hardware we implement to be a backstop for badly-behaved software, as there are some other pitfalls. e.g. 1. if the RXFIFO becomes full (and software ignores the interrupts), currently we would just drop any additional received bytes. 2. if the FMTFIFO becomes empty where the final entry did not contain a STOP entry, just looking at the code we would go back to Idle without creating a STOP condition (Just ran a quick sim and we hang in the Idle state with SDA=1/SCL=0).

> I'd leave SCL high while waiting for SW, though. In a multi-controller context, if SW takes too long and the transaction is forced to end, it might be good to communicate the info to SW. I think in most cases, a Stop would be forthcoming anyway.

SCL high wait-state makes sense. I'll also add a status bit that gets set if we return to Host-Mode disabled due to the timeout mechanism, and clears the next time Host-Mode is enabled.

hcallahan-lowrisc:
rebased

hw/ip/i2c/data/i2c.hjson
```diff
@@ -728,5 +736,37 @@
         { bits: "7:0" }
       ]
     }
+    { name: "HOST_NACK_HANDLER_TIMEOUT"
```
andreaskurth, line 739: In the message of the commit that contains this change, "This CSR controls a timeout mechanism where if SW does handle a pending 'nak'" should instead say "if SW does **not** handle a pending 'nak'", right?

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -491,6 +491,11 @@ class i2c_base_vseq extends cip_base_vseq #(
     if (bit'(get_field_val(ral.intr_state.tx_stretch, intr_clear))) begin
       `uvm_info(`gfn, "\n  clearing tx_stretch", UVM_DEBUG)
     end
+    if (bit'(get_field_val(ral.intr_state.nak, intr_clear))) begin
+      // Add a longer delay to mimic a software handler clearing the NAK condition
+      cfg.clk_rst_vif.wait_clks(1_000);
```
andreaskurth, line 496: Suggest randomizing the delay here, also so that the timeout and the non-timeout case are covered

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -491,6 +491,11 @@ class i2c_base_vseq extends cip_base_vseq #(
...
+      `uvm_info(`gfn, "Clearing 'NAK' interrupt, alllowing FSM to continue...", UVM_DEBUG)
```
andreaskurth, line 497: Typo: allowing

hw/ip/i2c/dv/env/seq_lib/i2c_host_may_nack_vseq.sv
```diff
@@ -0,0 +1,40 @@
...
+    ral.intr_enable.tx_threshold.set(1'b0);
+    ral.intr_enable.sda_interference.set(1'b0);
+    ral.intr_enable.scl_interference.set(1'b0);
+    ral.intr_enable.sda_unstable.set(1'b0);
+    ral.intr_enable.stretch_timeout.set(1'b0);
+    csr_update(ral.intr_enable);
+
+    // Enable the 'nak' timeout feature, but with a longer timeout than our handler latency.
+    ral.host_nack_handler_timeout.en.set(1'b1);
+    ral.host_nack_handler_timeout.val.set(30'd5000);
```
andreaskurth, line 22: Also this value would be good to randomize, I think

andreaskurth (approved):
LGTM overall, thanks for these changes @hcallahan-lowrisc. Please give @a-will a chance to review it as well before you merge, though.

I left a few smaller comments. If you think the DV suggestions are out of scope, feel free to create an issue to address them as part of M4.

hw/ip/i2c/data/i2c.hjson
hcallahan-lowrisc: Indeed it should! ty

hw/ip/i2c/rtl/i2c_core.sv
```diff
@@ -168,12 +175,20 @@ module i2c_core import i2c_pkg::*;
   logic        unused_alert_test_q;
   logic        unused_txrst_on_cond_qe;
 
+  assign hw2reg.ctrl.enablehost.d = 1'b0;
+  assign hw2reg.ctrl.enablehost.de = host_disable;
```
a-will, line 179: Are we sure we can avoid all the race conditions that arise from automatically sending a Stop? There are at least hazards present if software is not interrupted when the transaction ends. We've got two streams:

SW wants to continue transaction:

0. Clear FMT FIFO
1. Push Start/Addr/Cmd to FMT FIFO.
2. Check if transaction already timed out by reading the status CSR.
3. Clear the NAK interrupt.

HW races with SW:

2. TIMEOUT: Disable host and send a Stop.

If the sequence is `[S1, S2, H2, S3]`, how does SW know that there is data sitting in the FMT FIFO that will never go anywhere? Is it enough for SW to read the `ctrl.host_enable` bit after clearing the NAK interrupt, then abort if it went low?

hw/ip/i2c/rtl/i2c_fsm.sv
```diff
@@ -1122,10 +1154,26 @@ module i2c_fsm import i2c_pkg::*;
...
+            // is not popped until the NACK is acknowledged by software.
+            load_tcount = 1'b1;
+            tcount_sel = tClockHigh;
+            if (unhandled_nak_cnt_expired) begin
+              // If our timeout counter expires, generate a STOP condition and disable Host-Mode.
+              state_d = ClockStop;
+              load_tcount = 1'b1;
+              tcount_sel = tClockStop;
+              host_disable_o = 1'b1;
+            end
```
a-will, line 1171: I feel kind of iffy about having an automatic Stop sent out here. Generally, I'd expect the timeout to represent the bus going back to idle (which might as well make the FSM go directly to the Idle state), so automatically driving more symbols out can kind of violate rules in some contexts (especially for multi-controller SMBus cases). At least the timeout feature is optional.

The alternative would be to have the FSM go to Idle on the timeout and allow software to decide what to do. In the multi-controller case, losing the bus / timing out a transaction is probably worthy of its own interrupt bit (or an independently-cleared event bit contributing to a "controller stopped" status interrupt, so (1) an unexpected NAK or (2) lost arbitration or (3) timing out to bus idle could all simply gang up on that one interrupt).

a-will (approved):
The timeout mechanism and controls feel a little awkward, but we could choose to clean this up in the multi-controller work, a context that might help inform which directions work best.

hw/ip/i2c/rtl/i2c_core.sv
hcallahan-lowrisc: I agree that the mechanism feels awkward as a whole, and it may make sense to rip this out and re-architect for the multi-controller feature. Let's discuss it in the D2(S) review meeting.

I think there is a race-free flow that SW could use with this implementation, though it's maybe not the definition of ergo:
1. SW gets 'nak' irq.
2. SW checks STATUS.HOST_DISABLED_NACK_TIMEOUT, if 0 we are still halted. (optional)
3. SW clears HOST_NACK_HANDLER_TIMEOUT.EN, if we are halted but not yet timed out this prevents the automatic stop.
4. SW reads STATUS.HOST_DISABLED_NACK_TIMEOUT again to ensure we didn't race the automatic stop.
5. SW can now do whatever it needs to before clearing the 'nak' irq.

LMK if that makes sense.

hw/ip/i2c/rtl/i2c_fsm.sv
hcallahan-lowrisc: I agree that this doesn't really easily mesh with the multi-controller application. I deliberately did not consider that case too closely, as for this milestone we chose to consider all those grouped extensions as a separate work package. It might make sense to merge this as-is, then come back and refine when we have a plan for how to architect for the MCTP/SMBus compat. Again we can discuss at D2.

hcallahan-lowrisc:
We reviewed this issue in the D2(S) signoff meeting, and decided that the implementation here was acceptable.
For future multi-controller support, this implementation may prove to be insufficient, and it can be addressed then as part of a larger re-architecture.

PR #22050, opened 2024-03-15
[tlul,rtl] Weaken "known" assertion in tlul_adapter_sram.sv

rswarbrick:
The existing assertion was actually false because the response from the SRAM might have been unknown. Thank you, Jasper, for leading me to this!


vogelpi (approved):
This looks good to me. Maybe we should wait with merging until the timing changes of @GregAC have been merged. These changes might touch the same files.

rswarbrick:
I think it has landed and we got lucky :-) Merging.

PR #22051, opened 2024-03-15
[prim] Weaken "known" assertion in prim_fifo_sync

rswarbrick:
The existing assertion was false because it essentially asserted that we never write an unknown value into the FIFO. Not something that can be asserted from within the FIFO (obviously) and not actually true in the design, because we use the FIFO on the output from memories.

This assertion is slightly weaker. Instead of saying "rdata is never X", it tracks whether we've seen an X as input since the last time the FIFO was empty. If we haven't, it asserts that rdata is not X.

Running locally, Jasper doesn't quickly find a proof for this new statement (so I'm absolutely sure there's a better way to do this). But it also doesn't find a counterexamples, so I'm convinced this is an improvement!

I'm not sure whether the "ifdef FPV_ON" is needed here. It makes it more obvious that this definitely won't appear in the synthesized design, but I don't have strong feelings: maybe it's something that makes sense to track in simulations too.

(This is related to #22050, which is "downstream" of the property that we're weakening here)

### DIFF
--- a/hw/ip/prim/rtl/prim_fifo_sync.sv
+++ b/hw/ip/prim/rtl/prim_fifo_sync.sv
@@ -183,6 +183,30 @@ module prim_fifo_sync #(
     end
 
     `ASSERT(depthShallNotExceedParamDepth, !empty |-> depth_o <= DepthW'(Depth))
+
+
+    // "X tracking"
+    //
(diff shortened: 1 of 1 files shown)


rswarbrick:
In hindsight, the right approach is probably to do a bit more work to make something provable. Let's drop this work-around for now.

PR #22052, opened 2024-03-15
Prepare for LTO

luismarques:
Update the CRT to a version that includes the LTO feature.
Adds `OT_USED` to variables that would seem unreferenced and thus could be discarded by LTO.

### DIFF
--- a/sw/device/silicon_creator/lib/base/static_critical_boot_measurements.c
+++ b/sw/device/silicon_creator/lib/base/static_critical_boot_measurements.c
@@ -10,4 +10,4 @@
 // This is placed at a fixed location in memory within the .static_critical
 // section. It will be populated by the ROM before the jump to ROM_EXT.
 OT_SET_BSS_SECTION(".static_critical.boot_measurements",
-                   boot_measurements_t boot_measurements;)
+                   OT_USED boot_measurements_t boot_measurements;)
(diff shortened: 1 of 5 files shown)


sw/device/lib/base/BUILD
```diff
@@ -120,33 +120,12 @@ cc_library(
     ],
 )
 
-cc_library(
```
cfrantz, line 123: Can you comment on why this is being removed?  I don't necessarily object, but:

The original motivation here was to prevent the compiler provided 64-bit division functions from landing in our code. Things like `__udivdi3` are rather large by comparison to how much ROM space we have available.

luismarques: If you do a 64-bit division you'll get a linker error anyway: `undefined symbol: __udivdi3`. But I'll see if I can avoid adding that commit.

jwnrt (approved):

jwnrt:
Skipping the failed test as `//sw/device/tests/crypto:drbg_functest_fpga_cw310_sival_rom_ext` is known to be flaky in `sival_rom_ext` - see https://github.com/lowRISC/opentitan/pull/22075.

sw/device/lib/base/BUILD
luismarques: @cfrantz I have checked that these `__udivdi3` etc functions are hard-coded into the compiler and won't be removed during LTO, even if they aren't referenced. So we can't keep the current solution. In any case, I don't think we need it anyway, as we still get a linker error if we unintentionally do a 64-bit division, etc.

PR #22054, opened 2024-03-15
Increase OTBN imem size to 8 KiB

GregAC:
Only to be merged if we choose to have the imem size increase.

@moidx @jadephilipoom please note with the larger imem size the branch instructions can no longer branch to any other instruction, it's limited to a 4 KiB window either side of the instruction (PC relative addressing). The jump instructions can continue to branch anywhere in imem. Hopefully this doesn't present any issues.

@rswarbrick I've seen some failures in the RIG, I believe due to this line of the `BadBranch` class:

https://github.com/lowRISC/opentitan/blob/53c6c771ed140efeefccf942ce5d875134ff3679/hw/ip/otbn/dv/rig/rig/gens/bad_branch.py#L42

This will no longer work as depending on which PC the bad branch is located at the top or bottom of memory may not be reachable. Could you look at a fix?

### DIFF
--- a/hw/ip/otbn/data/otbn.hjson
+++ b/hw/ip/otbn/data/otbn.hjson
@@ -728,7 +728,7 @@
     // Imem size (given as `items` below) must be a power of two.
     { window: {
         name: "IMEM",
-        items: "1024", // 4 kB
+        items: "2048", // 8 kB
         swaccess: "rw",
         data-intg-passthru: "true",
         byte-write: "false",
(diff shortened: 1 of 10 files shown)


jadephilipoom:
> @moidx @jadephilipoom please note with the larger imem size the branch instructions can no longer branch to any other instruction, it's limited to a 4 KiB window either side of the instruction (PC relative addressing). The jump instructions can continue to branch anywhere in imem. Hopefully this doesn't present any issues.

Does this show up as a runtime error or an assembler error? And if the former, could we add a check to the assembler to catch these cases quickly? It's doable to organize code so that the branches are closer together, but I'm concerned about some not-often-used code path triggering the error if our tests don't happen to cover it.

GregAC:
@jadephilipoom it will be a assembler error, it's a static property of the branch (is its target more than 4 KiB away from it's location in imem)

msfschaffner:
I rebased this on master, which should hopefully fix a CI issue.

msfschaffner (approved):
RTL LGTM

msfschaffner:
@meisnere @OTshimeon FYI

vogelpi (approved):
LGTM, thanks @GregAC !

moidx (approved):

vogelpi:
I am merging this as agreed in the OT sync weekly meeting today to make sure it's included in the next RC.

PR #22056, opened 2024-03-15
[aes, csrng] Reduce area of unmasked AES core inside CSRNG

vogelpi:
This PR contains mainly two changes:
1. A new compile-time parameter is added to the AES cipher core to only implement the forward operation of the cipher. Enabling this parameter inside CSRNG allows to completely trim away unused data paths only relevant for the inverse operation. Previously some logic wasn't optimized out due to synthesis optimization barriers placed around the control FSM (synthesis was no longer able to infer that some mux control signals never take on certain values). This allows optimizing away 256 FFs, one 256-bit 2-input mux, one 128-bit 2-input mux and some more inputs on other existing 256-bit and 128-bit muxes.
2. The generation of the PRD signals for register clearing are moved to the outside of the cipher core. This allows connecting the initial state and key inputs inside CSRNG (where we don't care about SCA leakage and never clear the key registers).

These two changes allow saving around 4 kGE inside CSRNG.


andreaskurth (approved):
RTL changes LGTM, thanks @vogelpi :+1:

msfschaffner (approved):

PR #22057, opened 2024-03-15
[dice] add signature vars to UDS cert HJSON description

timothytrippel:
If we do not include the signature variables in the UDS cert HJSON description then the `kUdsMaxCertSizeBytes` constant that is automatically generated via the certificate codegen infra does not accound for the signature bytes. If these bytes are not accounted for, then the ujson struct that holds the final endorsed UDS certificate will not allocate an ArrayVec of adequate size on the host side.

### DIFF
--- a/sw/device/lib/testing/json/provisioning_data.h
+++ b/sw/device/lib/testing/json/provisioning_data.h
@@ -116,7 +116,7 @@ UJSON_SERDE_STRUCT(ManufDiceCerts, \
  */
 // clang-format off
 #define STRUCT_MANUF_ENDORSED_CERTS(field, string) \
-    field(uds_certificate, uint8_t, 596) \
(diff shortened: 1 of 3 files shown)


moidx (approved):

pamaury:
I don't have anything against this change but do you think it would be better to fix the code to compute the size even when those fields are not specified? I am happy to have a look at it today.

timothytrippel:
> I don't have anything against this change but do you think it would be better to fix the code to compute the size even when those fields are not specified? I am happy to have a look at it today.

I didn't view this as the ot_certs code being incorrect, it generates exactly what it was instructed to generate. I think it will probably be more confusing if the code starts generating space for fields that are not specified. Plus as is, it allow us to better test corner cases in the future.

pamaury (approved):

cfrantz (approved):

PR #22059, opened 2024-03-15
[silicon_creator] add OTP DAI read functions to driver lib

timothytrippel:
This adds several functions to the OTP silicon_creator driver library to enable reading data from non-memory mapped, readable, partitions. This is required to enable measuring said OTP partitions for attestation key and cert generation purposes.


moidx (approved):

timothytrippel:
Merging this as the failures are unrelated.

PR #22070, opened 2024-03-18
[aon,ci] Remove broken tag for silicon and use standard exec env

pamaury:


jwnrt (approved):
So this test is working now? I remember it giving us problems in CI previously (or was that `aon_wakeup`)

pamaury:
It was fixed by #22003

PR #22073, opened 2024-03-18
[spi_device,dv] Generalise description of mem_cfg testpoint

### LINKED ISSUE #21863: [spi_device,dv] Drop explicit mentions of 2-port RAMs in testplan
### Description

The spi_device testplan currently talks explicitly about 2-port RAMs. The test in question will also work for the `1r1w` configuration. Make the text less specific.

rswarbrick:
This testpoint was written when we only had the two-port implementation. It makes sense in the 1r1w configuration too. Generalise the text a bit so that it is sensible either way.

Fixes #21863.

### DIFF
--- a/hw/ip/spi_device/data/spi_device_testplan.hjson
+++ b/hw/ip/spi_device/data/spi_device_testplan.hjson
@@ -41,9 +41,13 @@
     }
     {
       name: mem_cfg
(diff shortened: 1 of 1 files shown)


a-will (approved):

rswarbrick:
CI looks a little wonky here, but I don't believe it can be related to the change in this commit so I'm merging now.

PR #22074, opened 2024-03-18
[spi_device,dv] Lower the reseed count for spi_device_ram_cfg

rswarbrick:
This test randomises a configuration many times and then checks that the randomised values make it to the RAM itself. That's reasonable, but there's not really any benefit from running it several times. Cut down to avoid the wasted work.


marnovandermaas (approved):
I think this is fine. Good to not load CI unnecessarily.

PR #22078, opened 2024-03-18
[bazel] Add some cw340 exec_envs

jwnrt:
This PR adds `cw340_test_rom` and `cw340_rom_with_fake_keys` execution environments. The `cw340_rom_with_fake_keys` one is tested through `uart_smoketest` in CI. They're both using the `hyper340` transport.

Unlike the CW310, CW340 boards can only be in either the HyperDebug or SAM3X configurations at any given time.

Our CI currently only has a board in the HyperDebug configuration. Adding exec_envs for the SAM3X configuration would be simple enough but would not get tested in CI.

### DIFF
--- a/hw/top_earlgrey/BUILD
+++ b/hw/top_earlgrey/BUILD
@@ -249,20 +249,6 @@ fpga_cw305(
 ###########################################################################
 fpga_cw340(
     name = "fpga_cw340",
-    design = "earlgrey",
-    exec_env = "fpga_cw340",
-    libs = [
-        "//sw/device/lib/arch:boot_stage_rom_ext",
-        "//sw/device/lib/arch:fpga_cw340",
-    ],
-    linker_script = "//sw/device/lib/testing/test_framework:ottf_ld_silicon_creator_slot_a",
-    rom_scramble_config = "//hw/top_earlgrey/data:autogen/top_earlgrey.gen.hjson",
-    test_cmd = "testing-not-supported",
-)
-
-fpga_cw340(
-    name = "fpga_cw340_test_rom",
-    testonly = True,
     args = [
         "--rcfile=",
         "--logging=info",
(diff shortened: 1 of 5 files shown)


jwnrt:
Thanks @engdoreis that worked great. I've added a `cw340_rom_with_fake_keys` target that will get tested in CI, but the `test_rom` target works too.

jwnrt:
@engdoreis do you think we should remove the `uart_smoketest_signed` target in this PR as well? It's using the old `opentitan_functest` macro but otherwise looks the same as `uart_smoketest`. I think maybe it was left here since it was the only target being tested on the CW340.

CC @cfrantz

a-will:
> @engdoreis do you think we should remove the `uart_smoketest_signed` target in this PR as well? It's using the old `opentitan_functest` macro but otherwise looks the same as `uart_smoketest`. I think maybe it was left here since it was the only target being tested on the CW340.
> 
> CC @cfrantz

Yes, the cw340 was the only target using it. With this PR, it should be obsolete.

hw/top_earlgrey/BUILD
```diff
@@ -249,39 +249,61 @@ fpga_cw305(
...
-        "--logging=info",
-        "--interface={interface}",
-    ] + select({
-        "@//ci:lowrisc_fpga_cw340": ["--uarts=/dev/ttyACM_CW340_1,/dev/ttyACM_CW340_0"],
-        "//conditions:default": [],
-    }),
     base = ":fpga_cw340",
     base_bitstream = "//hw/bitstream/cw340:bitstream",
     exec_env = "fpga_cw340_test_rom",
     otp_mmi = "//hw/bitstream/cw340:otp_mmi",
```
a-will, line 282: It would be wise not to assume the bitstream's built-in OTP image is correct.

```suggestion
    otp_mmi = "//hw/bitstream/cw340:otp_mmi",
    otp = "//hw/ip/otp_ctrl/data:img_rma",
```

engdoreis (approved):

PR #22079, opened 2024-03-18
[ast,rtl] Fix verible lint check for pgd modules

rswarbrick:
I managed to merge e8bfccf76c without realising that it breaks Verible lint in CI. Fortunately, the fix is easy!


rswarbrick:
Hmm. That ain't gonna work. Things like this aren't allowed in `always_comb` or `always_ff`. Rather than trying to get the original author's code to work, let's revert that change and sort it out properly.

Wrong:
```systemverilog
always_comb (* xprop_off *) begin
  ...
    vcc_pok_o = #(ast_bhv_pkg::VCC_POK_RDLY) gen_supp_a;
```

PR #21617, opened 2024-02-22
[cryptotest] Add NIST AES-GCM testvector parser

sadmansakib-git:
This PR pulls in the NIST testvectors for AES-GCM, and includes the associated parser and JSON schema.

PR #22080, opened 2024-03-18
[cryptotest] Test harness for AES-GCM

sadmansakib-git:
This PR adds the test harness for the AES-GCM, using the NIST test vectors added in https://github.com/lowRISC/opentitan/pull/21617. The test implementation contains a Rust host-side test driver, a device-side firmware handler, and a uJSON API allowing the two to communicate.


jwnrt (approved):
The test harness LGTM, thanks :)

PR #22082, opened 2024-03-18
[ast] Revert dependency on top-specific rstmgr_pkg

a-will:
ast is used in the englishbreakfast build, so it must depend on the virtual rstmgr_pkg.


a-will:
It might be good to design ast out of englishbreakfast at some point. Now probably isn't the time, though, hehe.

andreaskurth (approved):

vogelpi (approved):

PR #22083, opened 2024-03-18
[cryptotest] Add generator for random ECDSA test vectors

RyanTorok:
Adds a generator for random ECDSA tests. Per the crypto testplan, the randomness is sourced using a seed generated using python's internal randomness source. This seed is printed to `stderr` to allow for reproducible results in CI.

This PR also includes a few small changes to the ECDSA test harness to allow the random tests to work properly. The changes have been regression-tested against the existing test vectors.

~Dependent on #21713~

### DIFF
--- a/sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
+++ b/sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import json
+import logging
+import math
+import sys
+
+import jsonschema
+from Crypto.Hash import SHA256, SHA384, SHA512, SHA3_256, SHA3_384, SHA3_512
+from Crypto.PublicKey import ECC
+from Crypto.Signature import DSS
+from cryptotest_util import rng
+
+MESSAGE_LENGTH = 512
+
+
+def generate_test_vectors(args):
+    test_vectors = []
+    random = rng()
+    # Parse tests within the group
+    for count in range(args.count):
+        # Generate random ECC key pair
+        key_pair = ECC.generate(curve=args.curve, randfunc=random)
+        public = key_pair.pointQ
+        msg = random(MESSAGE_LENGTH)
+        signature = []
+        # Sign digest of message using private key
+        if args.hash_alg == "sha-256":
+            digest = SHA256.new(msg)
+        elif args.hash_alg == "sha-384":
+            digest = SHA384.new(msg)
+        elif args.hash_alg == "sha-512":
+            digest = SHA512.new(msg)
+        elif args.hash_alg == "sha3-256":
+            digest = SHA3_256.new(msg)
+        elif args.hash_alg == "sha3-384":
+            digest = SHA3_384.new(msg)
+        elif args.hash_alg == "sha3-512":
+            digest = SHA3_512.new(msg)
+        else:
+            raise ValueError("Unsupported hash algorithm " + args.hash_alg)
+        signer = DSS.new(key_pair, 'fips-186-3', randfunc=random)
+        signature = signer.sign(digest)
+
+        signature = list(signature)
+
+        verify_result = True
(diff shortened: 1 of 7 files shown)


sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
```diff
@@ -0,0 +1,149 @@
...
+            signer = DSS.new(key_pair, 'fips-186-3', randfunc=random)
+            signature = signer.sign(digest)
+        elif args.hash_alg == "sha3-384":
+            digest = SHA3_384.new(msg)
+            signer = DSS.new(key_pair, 'fips-186-3', randfunc=random)
+            signature = signer.sign(digest)
+        elif args.hash_alg == "sha3-512":
+            digest = SHA3_512.new(msg)
+            signer = DSS.new(key_pair, 'fips-186-3', randfunc=random)
+            signature = signer.sign(digest)
```
jadephilipoom, line 55: nit: I think you can simplify this if/elif chain:
```suggestion
        if args.hash_alg == "sha-256":
            digest = SHA256.new(msg)
        elif args.hash_alg == "sha-384":
            digest = SHA384.new(msg)
        elif args.hash_alg == "sha-512":
            digest = SHA512.new(msg)
        elif args.hash_alg == "sha3-256":
            digest = SHA3_256.new(msg)
        elif args.hash_alg == "sha3-384":
            digest = SHA3_384.new(msg)
        elif args.hash_alg == "sha3-512":
            digest = SHA3_512.new(msg)
        signer = DSS.new(key_pair, 'fips-186-3', randfunc=random)
        signature = signer.sign(digest)
```

sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
```diff
@@ -0,0 +1,149 @@
...
+                signature[math.floor(idx / 8)] ^= 1 << (idx % 8)
+                verify_result = False
+
+        r = signature[:int(len(signature) / 2)]
+        s = signature[int(len(signature) / 2):]
+
+        d = hex(int(key_pair.d))[2:]
+        # pad private value d to even number of characters
+        if len(d) % 2 == 1:
+            d = "0" + d
```
jadephilipoom, line 80: Is this not needed for x and y as well?

sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
```diff
@@ -0,0 +1,149 @@
...
+            if len(signature) == 0:
+                # To change an empty signature, make it a single random byte
+                signature = list(random(1))
+            else:
+                # Flip a random bit in the signature
+                idx = int.from_bytes(random(4), "big") % (len(signature) * 8)
+                signature[math.floor(idx / 8)] ^= 1 << (idx % 8)
+                verify_result = False
+
+        r = signature[:int(len(signature) / 2)]
```
jadephilipoom, line 74: nit: you can do the integer division with `//` (here and on the line below)
```suggestion
        r = signature[:len(signature) // 2]
```

jadephilipoom (approved):

sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
RyanTorok: Doing the padding in the generator script is obsolete now, thanks to improvements made to the Rust test harness in #21713.

Previously, this was required because the library used by the Rust test harness to perform the modular subtraction for the private key masking was particularly picky about the private key being the right size, which is why this was not needed for `x`/`y`. However, #21713 fixed this to perform the padding in the test harness, so the python scripts do not have to worry about it.

sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
RyanTorok: I recall making this change locally at some point but it must have fallen through the cracks...

sw/host/cryptotest/testvectors/parsers/random_ecdsa_generator.py
RyanTorok: A new operator a day keeps the `ValueError`s away :)

RyanTorok:
Fixed requested changes and rebased to master to avoid merge conflict with #20900

PR #19559, opened 2023-08-31
[ast] Reword some pgd/pok blocks in AST

rswarbrick:
This code is in an ifndef synthesis block and works by asserting a
signal (e.g. vcc_pok_o) shortly after the start of time, using a
"#RDLY" delay to get the timing right.

Using "always @*" to structure this causes Verible to warn because of
its "always-comb" lint rule. There's no real reason to use this style
here (and the code goes back to the initial AST import in 9ee56b2 from
2020).

Make the ordering more explicit, and stick everything in the same
initial block: it should be easier to understand and does the same
thing.

Fixes #19556.

PR #22085, opened 2024-03-18
Revert "[ast] fixing always_comb lint"

rswarbrick:
This reverts commit e8bfccf76c80a082870348b4f2acfec2a27b3c17. The change here doesn't work at all (you can't put #(...) delays inside an always_comb).

Undo the change for now: we can try to fix things properly when CI is working again.

### DIFF
--- a/hw/top_earlgrey/ip/ast/rtl/vcaon_pgd.sv
+++ b/hw/top_earlgrey/ip/ast/rtl/vcaon_pgd.sv
@@ -31,7 +31,7 @@ initial begin
   init_start = 1'b0;
 end
 
-always_comb begin
+always @( * ) begin
   if ( init_start ) begin
     vcaon_pok_o <= 1'b0;
   end else if ( !init_start && gen_supp_a ) begin
--- a/hw/top_earlgrey/ip/ast/rtl/vcc_pgd.sv
+++ b/hw/top_earlgrey/ip/ast/rtl/vcc_pgd.sv
@@ -31,7 +31,7 @@ initial begin
   init_start = 1'b0;
 end
 
-always_comb (* xprop_off *) begin
+always (* xprop_off *) @( * ) begin
   if ( init_start ) begin
     vcc_pok_o <= 1'b0;
   end
--- a/hw/top_earlgrey/ip/ast/rtl/vcmain_pgd.sv
+++ b/hw/top_earlgrey/ip/ast/rtl/vcmain_pgd.sv
@@ -31,7 +31,7 @@ initial begin
   init_start = 1'b0;
 end
 
-always_comb begin
+always @( * ) begin
   if ( init_start ) begin
     vcmain_pok_o <= 1'b0;
   end else if ( !init_start && gen_supp_a ) begin
--- a/hw/top_earlgrey/ip/ast/rtl/vio_pgd.sv
+++ b/hw/top_earlgrey/ip/ast/rtl/vio_pgd.sv
@@ -31,7 +31,7 @@ initial begin
   init_start = 1'b0;
 end
 
-always_comb (* xprop_off *) begin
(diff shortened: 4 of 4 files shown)


hw/top_earlgrey/ip/ast/rtl/vcaon_pgd.sv
```diff
@@ -31,7 +31,7 @@ initial begin
   init_start = 1'b0;
 end
 
-always_comb begin
+always @( * ) begin
```
msfschaffner, line 34: should this be converted to `initial` then, or does that cause other problems?

GregAC (approved):
Whilst I agree with @msfschaffner that doing this with an `always @(*)` may not be the best solution approving on the basis that 
 - This is just reverting us back to a previous repository state
 - It's needed to fix CI

GregAC:
Though I note this commit itself is failing CI, shouldn't we see all green if this is fixing a CI issue?

msfschaffner:
@rswarbrick which tool failed with the ast change?
Because https://github.com/lowRISC/opentitan/pull/22087 seems to be passing xcelium.
I.e., I believe private CI only failed because the main repo and the private CI repo were out of sync for a while (regarding tooling versions).

rswarbrick:
Oh! I thought I'd broken things royally! I'm running home now and will leave this open overnight in case it's needed. But I'm thrilled if I misunderstood and all is fine... :-D

msfschaffner:
Yeah I think the only fix we need at this point is this here: https://github.com/lowRISC/opentitan/pull/22097

hw/top_earlgrey/ip/ast/rtl/vcaon_pgd.sv
rswarbrick: I think the problem is that the `gen_supp_*` signals get driven by a backdoor and the contents of this block are supposed to reflect those changes. So I fear that `initial` won't get what we want.

Switching to an initial block was the whole point of my proposed change in #19559. This got blocked Nuvoton engineers explained that my change was wrong and that I shouldn't care about this block because it's just a behavioural model anyway. I gave up, so I'm feeling really really stupid for merging their change that broke things by editing the same file. Grr.

andreaskurth:
@rswarbrick: While this reduces the number of warnings, it doesn't completely solve the problem, right? I see this in the CI log:
```
#### Lint Warnings
../src/lowrisc_systems_ast_0.1/rtl/vcc_pgd.sv:34:1-6: Use 'always_comb' instead of 'always @*'. [Style: combinational-logic] [always-comb]

../src/lowrisc_systems_ast_0.1/rtl/vio_pgd.sv:34:1-6: Use 'always_comb' instead of 'always @*'. [Style: combinational-logic] [always-comb]

../src/lowrisc_systems_ast_0.1/rtl/vcaon_pgd.sv:34:1-6: Use 'always_comb' instead of 'always @*'. [Style: combinational-logic] [always-comb]

../src/lowrisc_systems_ast_0.1/rtl/vcmain_pgd.sv:34:1-6: Use 'always_comb' instead of 'always @*'. [Style: combinational-logic] [always-comb]
```

Since this means the _Quality (in-depth lint)_ CI check is still broken on `master`, should we add waivers until we have found a permanent solution?

rswarbrick:
Eugh. I'm back to confused: why wasn't this broken yesterday morning?!

But I think I agree: it's probably waivers for now.

andreaskurth:
Thx -- would you be able to add the waivers and create an issue to get this resolved properly later? :slightly_smiling_face:

rswarbrick:
Done in #22123, I hope.

msfschaffner:
@meisnere FYI - we are reverting this for now and will add waivers.

PR #22086, opened 2024-03-18
[dv,usbdev] Fixes to DV sequences for usbdev

alees24:
Some recent RTL changes shown below have not been mirrored with DV changes and thus sequences have been failing more often:

- the 'pkt_received' interrupt is now a Status type interrupt and will become deasserted when the RX FIFO entry is read.
- the 'configin' registers must now be excluded from automated CSR testing because there is an interaction among the register bits.

Additionally, since checking of the RX packet properties and content is now available within the base sequence, modify the OUT side sequences to benefit from this more extensive testing.

(Note that whilst the 'recover_orig_data' function is local to each affected sequence for now, rather than shared in the base class, this is because the function should soon be unnecessary. The action of transmitting a data packet should not cause the object to be modified in the present manner.)

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -153,6 +153,7 @@ endtask
   task check_rx_packet(bit [3:0] endp, bit setup, bit [4:0] buffer_id,
                        byte unsigned exp_byte_data[]);
     uvm_reg_data_t rx_fifo_read;
+    int unsigned   aligned_size;
     int unsigned   offset;
 
     // Read RX FIFO which should contain a buffer description matching the expectations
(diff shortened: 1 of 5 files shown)


rswarbrick:
Hi @alees24! I'm really sorry but I think I've merged some other PRs that conflict with this one. Would you mind rebasing this to sort out the mess?

rswarbrick (approved):
Sorry about the annoyance with the merge conflicts. I've just looked through these changes, though, and they look good to me.

alees24:
Rebased and re-tested. Known linting issues at present are unrelated.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -227,6 +232,33 @@ endtask
...
+    csr_rd(.ptr(ral.intr_state), .value(intr_state));
+    // Get pkt_received interrupt status
+    pkt_received = bit'(get_field_val(ral.intr_state.pkt_received, intr_state));
+    // DV_CHECK on pkt_received interrupt; this is a Status type interrupt,
+    // so it will be asserted only whilst the RX FIFO is non-empty.
+    `DV_CHECK_EQ(pkt_received, 1);
+
+    // Read rx_fifo reg, clearing the interrupt condition.
+    check_rx_packet(endp, setup, buffer_id, exp_byte_data);
+    csr_rd(.ptr(ral.rxfifo), .value(read_rxfifo));
```
a-will, line 252: Is this an intentional doubled-up read of the RX FIFO? It looks like it already happens in `check_rx_packet()`.

alees24: Nope, that's an editing error as I imagine you'd guessed, thanks! Overlooked because in the present context and use the additional read is harmless.

a-will (approved):
Looks good to me!

Issue #12301, opened 2022-04-23
[prim] DV infra for additional standardized countermeasures


msfschaffner:
We just recently added two additional standardized countermeasure primitives. Also, I believe we have not yet added common tests for the double LFSR (similar to double counter). Hence, we should consider adding common tests for these:
- [ ] Spurious WE checks in regfiles ([prim_reg_we_check](https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/rtl/prim_reg_we_check.sv)), see #12113
- [ ] Duplicated RR arbiter ([prim_arbiter_tree_dup](https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/rtl/prim_arbiter_tree_dup.sv))
- [ ] Duplicated LFSR ([prim_double_lfsr](https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/rtl/prim_double_lfsr.sv))

msfschaffner:
CC @eunchan @tjaychen @jeoongp

msfschaffner:
We should probably add the countermeasure ID label `CONFIG.GLITCH_DETECT` to all IPs once the common test for spurious WE checks is available (similar to `BUS.INTEGRITY`).

msfschaffner:
@weicaiyang we added common tests for the WE checker, right?

weicaiyang:
> @weicaiyang we added common tests for the WE checker, right?

yes, just marked it done.

vogelpi:
I've discussed this offline with @rswarbrick and our DV team is going to look at this. For now, I am assigning @rswarbrick . We need this for M3.

vogelpi:
@moidx Changing the labels could be done after TO.
@vogelpi DV should be done for M3/4.

andreaskurth:
Deprioritized from Earlgrey-PROD

PR #22091, opened 2024-03-18
[flash_ctrl] Use redundant arbiters in scrambling module

msfschaffner:
I am not sure but I believe we do not yet have CM assertions similar to the sparse FSM or prim_count assertions for the duplicated arbiter.

We may want to fix that for PROD.M3 or PROD.M4: https://github.com/lowRISC/opentitan/issues/12301


vogelpi (approved):
LGTM, I wasn't aware we have hardened arbiters available!

vogelpi:
We need to run cmdgen again for this PR.

vogelpi:
Update: I've run cmdgen and force-pushed. Will merge the PR once CI finishes.

vogelpi:
The CI (Quality (in-depth lint)) failure is unrelated to this PR and due to the Verible style lint version change having landed too soon.

PR #22094, opened 2024-03-19
[spi_device] Match OE with the passthrough read pipeline data

a-will:
Add delays to the output-enable when the read pipeline is enabled in passthrough mode, so the drivers are left off for the full dummy cycle period.

@antmarzam discovered I left this out.

### DIFF
--- a/hw/ip/spi_device/rtl/spi_passthrough.sv
+++ b/hw/ip/spi_device/rtl/spi_passthrough.sv
@@ -618,6 +618,9 @@ module spi_passthrough
   logic [3:0] half_cycle_sampled_sd;
   logic [3:0] read_pipeline_stg1_d, read_pipeline_stg1_q;
   logic [3:0] read_pipeline_stg2_d, read_pipeline_stg2_q;
+  logic [3:0] host_s_en_muxed;
+  logic [3:0] read_pipeline_oe_stg1_d, read_pipeline_oe_stg1_q;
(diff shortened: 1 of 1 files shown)


hw/ip/spi_device/rtl/spi_passthrough.sv
```diff
@@ -657,30 +660,57 @@ module spi_passthrough
     .q_o       (read_pipeline_stg2_q)
   );
 
+  prim_flop #(
+    .Width         ($bits(read_pipeline_oe_stg1_d)),
+    .ResetValue    ('0)
+  ) u_read_pipe_oe_stg1 (
```
antmarzam, line 666: Just a minor comment to leave it with the same naming convention  as the `cio_sd_en_o` signal for flash_mode where the flop is declared as `u_read_en_pipe_stg1/2` and the signals are called `internal_sd_en_stg1/2` respectively.

Looks good to me otherwise!

a-will: If the only change suggestions come down to harmony of names across modules, I think I'm going to leave this as-is to save both myself and CI time. The names are descriptive and not ambiguous in this module's context, and for internal signals like this, the value added for that level of naming similarity is quite low.

I agree that it'd be nicer to have the same naming, but I need to prioritize other tasks.

antmarzam (approved):

a-will:
The failures are unrelated. Merging...

PR #22097, opened 2024-03-19
[ast/lint] Fix Verible linting errors

msfschaffner:
This should fix the Verible linting issue we're currently seeing in CI.

@meisnere FYI - this is now needed since `always_comb` should only contain blocking assingments.

### DIFF
--- a/hw/top_earlgrey/ip/ast/rtl/vcaon_pgd.sv
+++ b/hw/top_earlgrey/ip/ast/rtl/vcaon_pgd.sv
@@ -33,11 +33,11 @@ end
 
 always_comb begin
   if ( init_start ) begin
-    vcaon_pok_o <= 1'b0;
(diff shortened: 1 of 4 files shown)


andreaskurth:
Currently doesn't simulate due to delay statements in `always_comb` blocks; needs to be edited

rswarbrick (changes requested):
Ha. Unfortunately, that's what I tried yesterday in #22079 before realising my approach was completely wrong.

This is why I've opened #22085, which will undo the broken change that messed things up in the first place. I'm going to merge that now. We can always "revert the reversion" when we come up with a better fix.

msfschaffner:
Ah I see - I missed that one.

I believe we would have to rewrite this a bit to make it work, e.g. by using `initial` blocks instead.

I am closing this for now.

PR #22100, opened 2024-03-19
[dv, uart] Update RX/TX watermark randomisation

elliotb-lowrisc:
Adjust randomisation range of `rxilvl` and `txilvl` to automatically track `RxFifoDepth` and `TxFifoDepth` (assuming the relationship between FIFO depth and the number of levels remains the same).
    
Should restore `uart_intr` pass rate following recent reduction in RX and TX FIFO sizes.

### DIFF
--- a/hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
+++ b/hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
@@ -102,9 +102,8 @@ class uart_base_vseq extends cip_base_vseq #(.CFG_T               (uart_env_cfg)
       `DV_CHECK_RANDOMIZE_FATAL(ral.timeout_ctrl.en)
       csr_update(ral.timeout_ctrl);
 
-      `DV_CHECK_RANDOMIZE_FATAL(ral.fifo_ctrl.rxilvl)
-      `DV_CHECK_RANDOMIZE_WITH_FATAL(ral.fifo_ctrl.txilvl,
-                                     value <= 6;)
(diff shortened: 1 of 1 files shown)


hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
```diff
@@ -102,9 +102,10 @@ class uart_base_vseq extends cip_base_vseq #(.CFG_T               (uart_env_cfg)
       `DV_CHECK_RANDOMIZE_FATAL(ral.timeout_ctrl.en)
       csr_update(ral.timeout_ctrl);
 
-      `DV_CHECK_RANDOMIZE_FATAL(ral.fifo_ctrl.rxilvl)
-      `DV_CHECK_RANDOMIZE_WITH_FATAL(ral.fifo_ctrl.txilvl,
-                                     value <= 6;)
+      // Note: update randomisation range for RXILVL and/or TXILVL below when RxFifoDepth
+      // and/or TxFifoDepth (and matching FIFO_CTRL register) are changed.
+      `DV_CHECK_RANDOMIZE_WITH_FATAL(ral.fifo_ctrl.rxilvl, value <= 6;)
+      `DV_CHECK_RANDOMIZE_WITH_FATAL(ral.fifo_ctrl.txilvl, value <= 4;)
```
hcallahan-lowrisc, line 108: Can we pull the limits from a variable, such as a parameter or a package constant? That way this code wouldn't need to be updated as a separate step.

elliotb-lowrisc:
Updated changes to be derived from FIFO depth rather than constant

rswarbrick (approved):

Issue #17496, opened 2023-03-10
[ujson] implement integrity check mechanism


timothytrippel:
# Background

For [manufacturing tests](https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/manuf/data/manuf_testplan.hjson), we will use the [ujson framework](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/ujson/?q=ujson&ss=opentitan%2Fopentitan) to stream out data from the device, see [example test](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/tests/chip/gpio/src/main.rs) that already uses this in a command-control (host-device) setup.

# Problem

There are no integrity checks on ujson messages. This could cause issues when streaming out critical provisioning data (e.g., public keys) from a new chip.

# Solution / Tasks

- [ ] investigate how to add integrity checks to ujson structs
    - will crc / checksum of binary representation work?
    - will it be valid in C and Rust? (might need rust representation of structs to be #[repr(c)])
- [ ] implement integrity check mechanism in ujson framework

CC: @cfrantz

a-will:
Suggestion: Encapsulate the data in another protocol that provides a checksum.

This is more an organizational suggestion than anything functionally different, but the ujson layer itself probably does not need to provide this / can remain focused. Instead, it seems like something that we'd want to assign to an I/O layer.

engdoreis:
I agree, this would also address the case where the ujson message is larger than the SPI FIFO described here https://github.com/lowRISC/opentitan/issues/17494.

cfrantz:
> So, if we are only worried about bits being flipped in transport, we could just use a crc32 of the serialised struct. We could then send the crc32 as a seperate JSON object with one field following the message, so we can reuse the JSON serialiser to convert the crc32 to valid UTF-8.

I'd suggest going this route: checksum over the serialized form of the struct.  All ujson output is handled through `ujson_putbuf`, which itself uses a `ujson_t` instance to actually direct the IO.   All ujson outputs intended to be consumed by host-based programs are emitted from the firmware via the `RESP_OK` and `RESP_ERR` macros in `ujson_ottf.h`.

I think that `ujson_t` provides an ideal place for carrying enable/disable information for checksums as well as any intermediate data for the calculation.  `RESP_OK/_ERR` look like a reasonable place to calculate and emit the checksums:
```
#define RESP_OK(responder_, uj_ctx_, data_)
  ({
[...]

PR #22101, opened 2024-03-19
[ottf] Add support for CRC protected ujson messages from host to device

pamaury:
It turns out that we only support CRC-protected ujson messages when sending from the device to the host and not the other way around. Since some important data (e.g. certificates, manuf, ...) will be sent by the host, it makes sense to protect them as well. This PR introduces a mecanism to do so.

To keep the changes to a mimum, I decided to make this mecanism **opt-in**. On the device side, it is just a matter of transforming a call like
```c
TRY(ujson_deserialize_mem_read_req_t(uj, &op));
```
into
```c
TRY(UJSON_WITH_CRC(ujson_deserialize_mem_read_req_t, uj, &op));
```
On the device side, I added a new `send_with_crc` method to the `UartSend` trait so one simply needs to call the new method.

This PR only enables CRC for the memory backdoor messages at the moment. Since those are used by `OTTF_WAIT_FOR`, this means that all users for this macros are automatically CRC protected now. Conversion of other callers will be done in a follow-up PR.


sw/device/lib/testing/test_framework/ujson_ottf.h
```diff
@@ -20,6 +21,36 @@ extern "C" {
...
+ * @param deser_fn_ A pointer to the deserialization function generated by
+ * ujson_derive.
+ * @param uj_ctx_ A `ujson_t` representing the IO context.
+ * @param data_ A pointer to the data to send.
+ *
+ */
+#define UJSON_WITH_CRC(deser_fn_, uj_ctx_, data_)                    \
+  ({                                                                 \
+    ujson_crc32_reset(uj_ctx_);                                      \
+    LOG_INFO("waiting for usjon message...");                        \
```
jwnrt, line 39:
```suggestion
    LOG_INFO("waiting for ujson message...");                        \
```

Also, I wonder if these logs will be too noisy

pamaury: Good point, I will remove those print.

jwnrt (approved):
I think this LGTM. I was wondering what would be the downside of enabling this for all messages rather than having it opt-in. What could be broken by that?

jwnrt:
Yeah that makes sense, I didn't fully understand that because this is a macro it will need to be updated at all the call sites

sw/device/lib/testing/test_framework/ujson_ottf.h
```diff
@@ -20,6 +21,34 @@ extern "C" {
...
+        LOG_ERROR("ujson CRC mismatch: %x vs %x", msg_crc, crc.crc); \
+        err = DATA_LOSS();                                           \
+      }                                                              \
+    }                                                                \
+    err;                                                             \
+  })
+
 /**
  * Add a CRC to the response.
  * Should not be used directly.
```
HU90m, line 54: It would be nice to use the `OttfCrc` struct for device to host as well, i.e. in `RESP_CRC`, but it would likely be an unnecessary complication. *See https://github.com/lowRISC/opentitan/issues/17496#issuecomment-1523732105*

HU90m (approved):
Looks good

timothytrippel (approved):

timothytrippel:
Looks like merging this broke CI. #22277 should fix.

PR #22105, opened 2024-03-19
[spi_device,dv] Fix dvsim output paths for variant configs

### LINKED ISSUE #22104: [spi_device,dv] The separate spi_device configs are currently causing test failures in CI
### Description

I can reproduce this by running the following command from a checkout:
```
util/dvsim/dvsim.py hw/top_earlgrey/dv/top_earlgrey_sim_cfgs.hjson --select-cfgs=spi_device -i smoke --fixed-seed=1
```

I think the problem is that the two runs are both writing their output to the same directory and the stuff that looks at the results then gets confused because things have become corrupted.

rswarbrick:
Closes #22104.


rswarbrick:
It turns out that Gary had done this last week and I'd even reviewed it, but not hit merge. Oops! That's now merged, so we can abandon this one :-)

PR #22115, opened 2024-03-19
[otbn, rtl] Squash call stack read data to 0 when stack is empty

GregAC:
Previous to this change we were seeing X prop issues where a branch happened using a call stack as one or both of the compare operands when the stack was empty. Whilst OTBN raises an error due to the use of the empty stack the branch result goes to X (as the comparison inputs are X) which produces Xs on the imem request interface.

With this change output from the stack gets squashed to 0 when its empty.

Draft PR for as I don't think this is an essential fix. When running a regression I'm seeing some new failures though I think they were hidden behind previous X check failures (overall pass rate is similar). Further investigation should be done before merging this.

### DIFF
--- a/hw/ip/otbn/rtl/otbn_rf_base.sv
+++ b/hw/ip/otbn/rtl/otbn_rf_base.sv
@@ -89,7 +89,7 @@ module otbn_rf_base
   logic rd_stack_b;
 
   logic                     stack_full;
-  logic [BaseIntgWidth-1:0] stack_data_intg;
+  logic [BaseIntgWidth-1:0] stack_data_intg, stack_data_intg_raw;
   logic                     stack_data_valid;
 
   logic state_reset;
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
This looks sensible to me.

I wonder whether it would make sense to add an assertion that would explain why this change is sensible. I'm wondering about the following statements (which I *think* should be true):
- `stack_data_intg` is never unknown (because `stack_data_valid` is false if there are unknown bits in `stack_data_intg_raw`)
- If there are unknown bits in `stack_data_intg_raw` then `intg_err_o` will be high.

GregAC:
Thanks for the review @rswarbrick 

I did try adding the assertion that `stack_data_intg` is never unknown but I did see a couple of failures. It is possible for a program to push unknown bits into the stack (read from memory that has unknown data in then push to stack) so perhaps not a great assertion.

So no new assertions for now, I can look into the failure I found with the new `stack_data_intg` is never unknown assertion if you think that's a concern.

rswarbrick:
Good point about the assertion. No need to tweak it: this looks sensible to me.

rswarbrick (approved):

PR #22120, opened 2024-03-19
[dv,pwrmgr] Fix failures

matutem:
- Wait for a few slow clock cycles between iterations of pwrmgr_aborted_low_power to avoid consecutive pulses in CDC due to cfg_cdc_sync being too close.
- Move wait_for_csr_to_propagate_to_slow_domain into update_control_csr to avoid missing calling it.
- Fix incorrect check in fast_check_wake_info.
- Change cycles waited before low power transition to slow clock.


rswarbrick (approved):
This all looks sensible to me!

PR #22123, opened 2024-03-19
[lint,verible] Waive Verible's always-comb rule for AST blocks

rswarbrick:
This waiver is described properly in the code. The reason to bother is that the rule failure is breaking CI at the moment. Eek!


a-will (approved):

msfschaffner:
@meisnere FYI

PR #22129, opened 2024-03-20
[usbdev] Rewrite usbdev resume link active seq description

Saad-Anees1:
This PR includes the modification to the description of the usbdev resume_link_active sequence.

### DIFF
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -431,8 +431,8 @@
             then write will only have an effect when the device is in the LinkPowered state.
 
             - Perform no operation in J state for more than 3 ms.
-            - Read the link state from usbstate register to see if it is in powered state and then send the
-              resume signal (K state) for 20 ms and set the usbctrl.resume_link_active.
+            - Read the link state from usbstate register to see if it is in powered state and set the
+              usbctrl.resume_link_active.
             - Verify that the resume_link_active interrupt goes high and device will be resumed.
             '''
       stage: V2


Saad22386:
@rswarbrick

rswarbrick:
@alees24: Since you know rather more about USB than I do... Would you mind taking a look at this and checking it makes sense? (I suspect it does, but want to get an experienced pair of eyes!)

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -431,8 +431,8 @@
             then write will only have an effect when the device is in the LinkPowered state.
 
             - Perform no operation in J state for more than 3 ms.
-            - Read the link state from usbstate register to see if it is in powered state and then send the
-              resume signal (K state) for 20 ms and set the usbctrl.resume_link_active.
+            - Read the link state from usbstate register to see if it is in powered state and set the
+              usbctrl.resume_link_active.
             - Verify that the resume_link_active interrupt goes high and device will be resumed.
```
alees24, line 436: The description change is okay but the register name is 'usbstat' without a trailing 'e.'

Also, whilst this test description is being changed, please note that there is no 'resume_link_active interrupt' and the test should instead be looking at the link state in 'usbstat' to ascertain whether the 'resuming' state has been entered.

alees24 (changes requested):

hw/ip/usbdev/data/usbdev_testplan.hjson
Saad22386: > The description change is okay but the register name is 'usbstat' without a trailing 'e.'
> 
> Also, whilst this test description is being changed, please note that there is no 'resume_link_active interrupt' and the test should instead be looking at the link state in 'usbstat' to ascertain whether the 'resuming' state has been entered.

Are you suggesting that the "link_resume" interrupt in the intr_state register doesn't monitor the resuming state?

Saad22386:
When I set usbctrl.resume_link_active to 1 after the usbdev is in the powered state, it randomly transitions to either the active_nosof state or the resuming state. However, according to the usbctrl.resume_link_active bit description, it should be in the resuming state. Why does this occur?

alees24:
> When I set usbctrl.resume_link_active to 1 after the usbdev is in the powered state, it randomly transitions to either the active_nosof state or the resuming state. However, according to the usbctrl.resume_link_active bit description, it should be in the resuming state. Why does this occur?

I suspect that the usb20_driver is interfering with the test by driving a Bus Reset or Idle signal into the USB device.
The 'resume_link_active' bit is supposed to be set by software when the USB device is powered up whilst the usbdev_aon_wake module still has control of the bus. Software sets this bit and then enables the pull up, which causes the link state to transition to Powered and then into Resuming.

During this 'Resume Signaling' period (7.1.7.7 in the spec) there should be at least 20ms of K state, but the usb20_driver is presently unaware of suspend/resume behavior and it automatically drives a Bus Reset and Idle period onto the USB when activated.

alees24:

PR #22130, opened 2024-03-20
[topgen] Support if last module does not contain a parameter

Razer6:
The generator for the parameter section of the top-level module only generates the last comma correctly if it is a module that has parameters. If the last module has no parameter, the generator only prints a comment for that module but a comma is already created before, causing a syntax error as shown in the following snippet:

```systemverilog
  parameter int unsigned RvCoreIbexDmExceptionAddr =
      tl_main_pkg::ADDR_SPACE_RV_DM__MEM + dm::ExceptionAddress[31:0],
  parameter bit RvCoreIbexPipeLine = 0,
  // parameters for mbx0
  // parameters for mbx1
  // parameters for mbx2
) (
 ...
  ```
  
This change adds a new helper function to determine if there are any modules after the current one that have no parameters to determine if it is the last parameter.

### DIFF
--- a/util/topgen/lib.py
+++ b/util/topgen/lib.py
@@ -244,6 +244,12 @@ def get_pad_list(padstr):
     return pads
 
 
+def is_last_module_with_params(top, idx):
+    modules_after = top["module"][idx + 1:]
+    num_params = sum(len(m["param_list"]) for m in modules_after)
+    return num_params == 0
+
+
 # Template functions
 def ljust(x, width):
     return "{:<{width}}".format(x, width=width)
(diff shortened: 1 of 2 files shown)


rswarbrick:
Thanks for tidying this up. I was inspired when I read the code here and thought it was a bit sad that we're computing whether to add a comma in two different places. So I've made a modified version of this PR with you as a co-author and have pushed it up as #22164. I'm kind of happy either way but I *think* the refactored version is worth taking. (But I would say that, given that I wrote it!)

Razer6:
Closing in favor of https://github.com/lowRISC/opentitan/pull/22164

PR #22033, opened 2024-03-14
[WIP] Enable LTO

luismarques:

PR #22132, opened 2024-03-20
[test] Ensure SHA3 is idle before using KMAC

jwnrt:
In some situations software seems to work fast enough that it can try to use KMAC before SHA3 is idle after the previous use. Our testutils check for this condition but error out rather than try to recover from it. This change uses a very small timeout since it should become idle very quickly ideally.

The error condition was being observed when #22033 was applied. With this change, the error does not fire.

### DIFF
--- a/sw/device/tests/entropy_src_fw_override_test.c
+++ b/sw/device/tests/entropy_src_fw_override_test.c
@@ -263,6 +263,15 @@ static status_t entropy_config(
 }
 
 /**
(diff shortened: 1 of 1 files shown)


luismarques:
Why don't you instead check if it's idle and if so you wait for it to be? Then if it's still not idle after a reasonable period do error out.

luismarques (approved):

PR #22133, opened 2024-03-20
[pentest] Ibex SCA tests

nasahlpa:
This commit adds the following penetration tests:
- ibex.sca.register_file_read
- ibex.sca.register_file_write
- ibex.sca.tl_read
- ibex.sca.tl_write

This code is used to check whether data written/read to/from the tegister file or transmitted over TL-UL leaks.


johannheyszl (approved):
thanks @nasahlpa

moidx (approved):

PR #22139, opened 2024-03-20
[dv] TB update to cope and enable new read data pipeline enhancement

antmarzam:
- SPI host agent now waits a couple of extra clock_cycles when the read_pipeline delay is enabled.

- The read_pipeline feature has been added and enabled (randomly) to the cmd_info configuration routine as well as the seq_items.

- Scoreboard has been updated to understand the DS item payload won't be the same as the US item since when the CSB is deasserted two clock cycles worth of data will be latched in the read pipe

### DIFF
--- a/hw/dv/sv/spi_agent/spi_flash_cmd_info.sv
+++ b/hw/dv/sv/spi_agent/spi_flash_cmd_info.sv
@@ -11,13 +11,13 @@ class spi_flash_cmd_info extends uvm_sequence_item;
   // number of lanes when sending payload, set to 0 if no payload is expected
   rand bit [2:0] num_lanes;
   rand int dummy_cycles;
+  // 2-stage read_pipeline enabled when >0
   rand bit [1:0] read_pipeline_mode;
 
   constraint addr_mode_c {
     // for dual/quad mode, it always contains address
     num_lanes > 1 -> addr_mode != SpiFlashAddrDisabled;
   }
-
 
   constraint num_lanes_c {
     write_command -> num_lanes == 1;
--- a/hw/dv/sv/spi_agent/spi_flash_cmd_info.sv
+++ b/hw/dv/sv/spi_agent/spi_flash_cmd_info.sv
@@ -11,11 +11,13 @@ class spi_flash_cmd_info extends uvm_sequence_item;
   // number of lanes when sending payload, set to 0 if no payload is expected
   rand bit [2:0] num_lanes;
   rand int dummy_cycles;
+  rand bit [1:0] read_pipeline_mode;
 
   constraint addr_mode_c {
     // for dual/quad mode, it always contains address
     num_lanes > 1 -> addr_mode != SpiFlashAddrDisabled;
   }
+
 
   constraint num_lanes_c {
     write_command -> num_lanes == 1;
@@ -23,6 +25,10 @@ class spi_flash_cmd_info extends uvm_sequence_item;
   }
 
   constraint dummy_cycles_c {
+    solve write_command before read_pipeline_mode;
+    write_command  -> read_pipeline_mode == 0;
+    !write_command -> read_pipeline_mode <= 2;
+
     // for dual/quad read, need at least 2 dummy cycles
     num_lanes > 1 && !write_command -> dummy_cycles >= 2;
     // write or no payload doesn't have dummy cycle.
@@ -40,6 +46,7 @@ class spi_flash_cmd_info extends uvm_sequence_item;
     `uvm_field_int(write_command, UVM_DEFAULT)
     `uvm_field_int(dummy_cycles,  UVM_DEFAULT)
     `uvm_field_int(num_lanes,     UVM_DEFAULT)
+    `uvm_field_int(read_pipeline_mode, UVM_DEFAULT)
   `uvm_object_utils_end
 
   `uvm_object_new
--- a/hw/dv/sv/spi_agent/spi_monitor.sv
+++ b/hw/dv/sv/spi_agent/spi_monitor.sv
@@ -251,13 +251,13 @@ class spi_monitor extends dv_base_monitor#(
                                      item.dummy_cycles, item.read_pipeline_mode);
     `uvm_info(`gfn, $sformatf("sampled flash opcode: 0x%0h", item.opcode), UVM_HIGH)
     item.terminated_before_dummy_cycles = 1;
-    if(item.read_pipeline_mode>0)
+    if (item.read_pipeline_mode>0)
       item.terminated_before_read_pipeline = 1;
 
     `uvm_info(`gfn, "Triggering 'host_item.byte_sampled' after sampling opcode", UVM_DEBUG)
     -> host_item.byte_sampled_ev;
     sample_address(num_addr_bytes, item.address_q);
-    if(item.address_q.size > 0) begin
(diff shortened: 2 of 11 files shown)


hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -667,12 +667,39 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
     start_addr = convert_addr_from_byte_queue(up_item.address_q);
 
-    if (dn_item != null) `DV_CHECK_EQ(up_item.payload_q.size, dn_item.payload_q.size)
+    if (dn_item != null) begin
+      // If the read_pipeline is enabled, there will be data latched on the pipe which won't
+      // "come out" to the host since CSB has become inactive. This can be 2-bit, 4-bit or 8-bit
+      // depending whether we're doing a "normal" READ, READDUAL, READQUAD.
+      // Hence it's ok for the DS queue to be 1 item larger.
+
+      if(up_item.read_pipeline_mode && up_item.payload_q.size==0) begin
```
rswarbrick, line 676: Missing space after `if` (and similarly on next two `if` statements). I also think the conventional style is to add some spaces around `==` operators.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -667,12 +667,39 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+        // If the read_pipeline is enabled is possible the DS item has sampled some data
+        // since it had 2 "extra cycles", it really depends on when the txn terminated.
+        if(!((up_item.payload_q.size==dn_item.payload_q.size) ||
+             (up_item.payload_q.size==dn_item.payload_q.size-1)))
+          `uvm_error(`gfn, $sformatf(
+                           "DS and US items don't have the same payload size (US - %0d) (DS - %0d)",
+                                     up_item.payload_q.size,dn_item.payload_q.size ))
```
rswarbrick, line 683: Nit: Indentation. I'd probably stick a newline before `$sformatf` but not one after, then use `{"...", "..."}` with a line break if it still doesn't fit.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -667,12 +667,39 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+      end
+      else begin
+        if(!up_item.read_pipeline_mode ||
+           up_item.payload_q.size==0) begin //Read terminated before any data was returned
+          `DV_CHECK_EQ(up_item.payload_q.size, dn_item.payload_q.size)
+        end
+        else
+          `DV_CHECK_EQ(up_item.payload_q.size, dn_item.payload_q.size-1)
```
rswarbrick, line 691: I'd probably suggest restructuring this a bit, which will also allow us to avoid the comment after part of an expression. Maybe something like this? (I don't know anything about SPI! So feel free to correct me if I've got this wrong :-D)
```systemverilog
        // We expect the payload for up_item to be one byte shorter than the payload for dn_item,
        // unless dn_item had an empty payload or we're in read_pipeline mode.
        int unsigned exp_up_payload_size = dn_item.payload_q.size;
        if (exp_up_payload_size > 0 && up_item.read_pipeline_mode) exp_up_payload_size--;

        `DV_CHECK_EQ(up_item.payload_q.size, exp_up_payload_size)
      end
```

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -833,7 +860,34 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
         `DV_CHECK_EQ(upstream_item.address_q[i], downstream_item.address_q[i])
       end
     end else begin
-      if (!downstream_item.compare(upstream_item)) begin
+
+      if(upstream_item.read_pipeline_mode>0 &&
```
rswarbrick, line 864: Missing space after `if`. I'd also suggest spaces around `>` here and around `==` and `+` on the next line.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -833,7 +860,34 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+         (downstream_item.payload_q.size==upstream_item.payload_q.size+1)) begin
+        string print_str;
+        // Read pipeline is enabled, so if the US item is 1 item short of payload bytes (due to CSB
+        // becoming inactive and the "missing" data still in the read pipeline), we need to
+        // "manually" compare the items (opcode, address, payload)
+        print_str = "Comparing only the number of payload items in US item (1 less than DS item)";
+        print_str = {print_str, " ","since read_pipeline is enabled"};
```
rswarbrick, line 871: I think we can use `{"..", ".."}` to do this more easily with something like this:
```systemverilog
        `uvm_info(`gfn,
                  {"Comparing only the number of payload items in US item ",
                   "(1 less than DS item) since read_pipeline is enabled"},
                  UVM_DEBUG);
```

Similarly for the other places we're using `print_str` below. Maybe we don't need that variable at all?

hw/dv/sv/spi_agent/spi_flash_cmd_info.sv
```diff
@@ -11,18 +11,24 @@ class spi_flash_cmd_info extends uvm_sequence_item;
   // number of lanes when sending payload, set to 0 if no payload is expected
   rand bit [2:0] num_lanes;
   rand int dummy_cycles;
+  rand bit [1:0] read_pipeline_mode;
```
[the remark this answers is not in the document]
antmarzam: read_pipeline_mode is a new field in the cmd_info registers. I've added a comment to aid reading. Will resolve once I push the changes

hw/dv/sv/spi_agent/spi_monitor.sv
```diff
@@ -234,20 +235,38 @@ class spi_monitor extends dv_base_monitor#(
...
+    end
+    else begin
+      repeat ((item.read_pipeline_mode>0 ? 2 : 0)) begin
+        cfg.wait_sck_edge(SamplingEdge, active_csb);
+      end
+    end
+
+    `uvm_info(`gfn, $sformatf(
+      "TB waited for (%0d) dummy_cycles + (%0d) cycles due to read_pipeline",
+      item.dummy_cycles, (item.read_pipeline_mode && cfg.if_mode != dv_utils_pkg::Device)  ? 2 : 0),
```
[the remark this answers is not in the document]
antmarzam: It's intentional not to match previous ternary. That's because the previous only applies when the agent is used in host_mode, but not on device_mode - the pass through interface won't see the extra read_pipeline delay.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: I've tweaked a bit your suggestion a bit, but switched the code to look as your suggestion - thank you :)

PR #22144, opened 2024-03-20
[spi_host,dv] Fix some typing problems in spi_host DV code

rswarbrick:
There's some description in commit messages. This is about trying to avoid warnings coming out when building spi_host DV code with VCS (or, I suspect, Xcelium).

### DIFF
--- a/hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
+++ b/hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
@@ -298,11 +298,10 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
         end
 
         "configopts": begin
-          string      csr_str;
-          int         csr_idx;
-
-          csr_str = csr_name.getc(csr_name.len());
-          csr_idx = csr_str.atoi();
+          // Note: CONFIGOPTS is actually a multireg, but we've only got a count of 1, which means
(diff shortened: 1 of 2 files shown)


rswarbrick:
@hcallahan-lowrisc: Would you mind taking a look at this? (It's been sitting in my queue for a while and I just found it!)

rswarbrick:
Huh, that's not quite going to work. I'll sort it out properly tomorrow, but the original code was bogus. This bit:
```systemverilog
        "configopts": begin
          string      csr_str;
          int         csr_idx;

          csr_str = csr_name.getc(csr_name.len());
          csr_idx = csr_str.atoi();
```
This is a bit strange because `csr_str` is "configopts". Then `csr_str` ends up as the empty string and `csr_idx` equals zero. My proposed change subtracts '0' from "" and gets an index of -48. Oops! I guess I'm pleased that the silly code came to light eventually :-)

hcallahan-lowrisc:
Could you just tag me again when it's green on CI? I'll review it then. Cheers

rswarbrick:
Well that was embarrassing: I missed a semicolon! I think everything should probably go through cleanly this time. Will ping you in person if I get a tick :-)

rswarbrick:
@hcallahan-lowrisc: I think things are "green" now (since the failures are IT problems and not caused by this PR)

hcallahan-lowrisc (approved):

PR #22147, opened 2024-03-20
[lint] Demote licence warning in AscentLint parser

rswarbrick:
This should avoid spurious CI failures caused by the kokoro AscentLint instance warning that the licence will expire soon.

### DIFF
--- a/hw/lint/tools/dvsim/common_lint_cfg.hjson
+++ b/hw/lint/tools/dvsim/common_lint_cfg.hjson
@@ -29,11 +29,12 @@
                "{fusesoc_core}"]
 
   // Determines which message severities to print into report summaries.
-  report_severities: ["warning", "error"]
+  report_severities: ["info", "warning", "error"]
   // Determines which message severities lead to a pass/fail.
   fail_severities: ["warning", "error"]
   // Define message bucket categories and severities.
   message_buckets: [
+    {category: "flow", severity: "info",     label: ""},
     {category: "flow", severity: "warning",  label: ""},
     {category: "flow", severity: "error",    label: ""},
     {category: "lint", severity: "info",     label: ""},
--- a/util/dvsim/LintParser.py
+++ b/util/dvsim/LintParser.py
@@ -17,6 +17,7 @@ class LintParser():
 
     def __init__(self) -> None:
         self.buckets = {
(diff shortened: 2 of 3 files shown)


marnovandermaas (dismissed):
Just looking at the code, this looks fine to me. have you been able to test this? Should Kokoro on this PR pass or does it use the Python files from the main branch?

timothytrippel:
Looks like this change causes the following dvsim error, which crashes dvsim before it can run, causing the report to never be generated.

Screenshot 2024-03-20 at 11 30 02 PM

Overall, I like the idea of filtering out this warning to get things green again though :)

rswarbrick:
Huh, that's odd. It turns out that the list of buckets in the Python code in `LintParser.py` must match the list of buckets in the config file `common_lint_cfg.hjson`. I didn't realise this: oops!

I suspect the code could be refactored to be a little less spaghetti-ish, but maybe that's a problem for another day.

rswarbrick:
I've just force-pushed *yet* again, because it turns out that the license expiry message wasn't appearing in the HTML at all. Oops! I'm hoping that the most recent version will work properly. I've updated the commit message to describe how this all works. This is what it says about the changes we needed:

  1. Modify LintParser so that if there are matches for two different
     patterns that both start at the same character then we only
     attach a match to the first matching bucket.

  2. Tweak common_lint_cfg so that it knows about a new "info"
     severity.

  3. Tweak common_lint_cfg so that it knows about a new bucket
     for the "flow" category, with "info" severity.

  4. Declare this new "flow_info" bucket in LintParser.py

  5. Add a "license will expire" pattern in
     ascentlint-report-parser.py and assign matches to the new
     "flow_info" severity.

This was... much harder than I expected! But I hope it will work now.

rswarbrick:
Latest force-push avoids accidentally sticking the entire tool log in the output (it turns out that if `m` is a `Match` object, then `m.string` gives you the string that it was matching on, rather than the match itself: oops! `group(0)` seems to give the thing we actually wanted).

hcallahan-lowrisc (approved):
Looking at the kokoro results I can now see a pass is reported and the license expiry message is shown under the 'Infos' heading, so it would seem to be working as intended.

PR #22148, opened 2024-03-20
[cryptotest] Test harness for AES-GCM

sadmansakib-git:
This PR adds the test harness for the AES-GCM, using the NIST test vectors added in https://github.com/lowRISC/opentitan/pull/21617. The test implementation contains a Rust host-side test driver, a device-side firmware handler, and a uJSON API allowing the two to communicate.


sw/device/tests/crypto/cryptotest/firmware/aes.c
```diff
@@ -27,6 +32,312 @@ static const uint32_t kKeyMask[8] = {
...
+
+  // Set the auth_tag and tag_len
+  size_t tag_num_words = ceil_div(uj_data.tag_length, sizeof(uint32_t));
+  uint32_t tag_buf[tag_num_words];
+  memcpy(tag_buf, uj_data.tag, uj_data.tag_length);
+  otcrypto_word32_buf_t auth_tag = {
+      .data = tag_buf,
+      .len = tag_num_words,
+  };
+  otcrypto_aes_gcm_tag_len_t tag_len = get_tag_length(tag_num_words * sizeof(uint32_t));
```
RyanTorok, line 152: Possibly return `INTERNAL(..)` if `get_tag_length` returns zero?

sw/device/tests/crypto/cryptotest/firmware/aes.c
```diff
@@ -27,6 +32,312 @@ static const uint32_t kKeyMask[8] = {
...
+
+  // Set the auth_tag and tag_len
+  size_t tag_num_words = ceil_div(uj_data.tag_length, sizeof(uint32_t));
+  uint32_t tag_buf[tag_num_words];
+  memcpy(tag_buf, uj_data.tag, uj_data.tag_length);
+  otcrypto_const_word32_buf_t auth_tag = {
+      .data = tag_buf,
+      .len = tag_num_words,
+  };
+  otcrypto_aes_gcm_tag_len_t tag_len = get_tag_length(tag_num_words * sizeof(uint32_t));
```
RyanTorok, line 224: See comment above about return an error here if `get_tag_length` returns 0.

sw/device/tests/crypto/cryptotest/firmware/aes.c
```diff
@@ -27,6 +32,312 @@ static const uint32_t kKeyMask[8] = {
...
+  size_t nblocks = ceil_div(uj_data.input_length, kAesBlockNumBytes);
+  size_t text_buffer_length = nblocks * kAesBlockNumBytes;
+  
+  // Create buffer to store the input. For decryption, it is the ciphertext.
+  uint8_t input_buf[text_buffer_length];
+  memcpy(input_buf, uj_data.input, uj_data.input_length);
+
+  // Set the ciphertext field
+  otcrypto_const_byte_buf_t ciphertext = {
+      .data = input_buf,
```
RyanTorok, line 205: I think you can just put `uj_data.input` here and get rid of `input_buf`.

sw/device/tests/crypto/cryptotest/firmware/aes.c
```diff
@@ -27,6 +32,312 @@ static const uint32_t kKeyMask[8] = {
...
+  size_t nblocks = ceil_div(uj_data.input_length, kAesBlockNumBytes);
+  size_t text_buffer_length = nblocks * kAesBlockNumBytes;
+ 
+  // Create buffer to store the input. For encryption, it is the plaintext.
+  uint8_t input_buf[text_buffer_length];
+  memcpy(input_buf, uj_data.input, uj_data.input_length);
+
+  // Set the plaintext field
+  otcrypto_const_byte_buf_t plaintext = {
+      .data = input_buf,
```
RyanTorok, line 133: I think you can just put `uj_data.input` here and get rid of `input_buf`.

sw/device/tests/crypto/cryptotest/firmware/aes.c
```diff
@@ -27,6 +32,312 @@ static const uint32_t kKeyMask[8] = {
...
+
+  // Test stepwise API
+  otcrypto_aes_gcm_context_t ctx;
+  status = otcrypto_aes_gcm_decrypt_init(&key, iv, &ctx);
+  if (status.value != kOtcryptoStatusValueOk) {
+    return INTERNAL(status.value);
+  }
+
+  size_t plaintext_bytes_written;
+  TRY(stream_gcm(&ctx, aad, ciphertext, plaintext, &plaintext_bytes_written));
```
RyanTorok, line 247: When I wrote the test for stepwise SHA-2, @jadephilipoom suggested splitting the input into two parts to get better coverage of the intermediate call. Not sure if that would work with GCM.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+        test_case.mode,
+        test_case.operation,
+        test_case.key_len,
+        test_case.test_case_id
+    );
+    let uart = transport.uart("console")?;
+
+    CryptotestCommand::AesGcm.send(&*uart)?;
+
+    assert!(
```
RyanTorok, line 77: Some of my earlier test harnesses added these checks, completely out of paranoia that UART would be messed up if we panicked after sending a partial test case. This turns out to have no negative side effects (aside from the test failing, of course), so another option I have used more recently is to add the error message in an `expect` call when creating the `ArrayVec` for each input value, rather than including an explicit length check. The other nice thing about it is that `ArrayVec`'s length is a generic parameter that can be extrapolated by the compiler, so we wouldn't need to copy the max data lengths from the uJSON type definitions. Up to you whether to make this change.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+         ("stepwise", gcm_output.stepwise_output)]
+    .into_iter()
+    .for_each(|(mode, output)|{
+        let target_output = match test_case.operation.as_str() {
+            "encrypt" => &test_case.ciphertext,
+            "decrypt" => &test_case.plaintext,
+            _ => panic!("Invalid operation type")
+        };
+        let success = if target_output.len() > gcm_output.output_length {
+            // If we got a shorter digest back then the test asks for, we
```
RyanTorok, line 158: These comments looks like they were copied from the hash algorithm tests. Update to AES-specific message?

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+            failed = true;
+        }
+    });
+    if failed {
+        *fail_counter += 1;
+    }
+    Ok(())
+}
+
+fn is_input_empty(aes_gcm_test: &AesGcmTestCase) -> bool {
```
RyanTorok, line 185: What's the purpose of this function? Does cryptolib not accept empty inputs? If that's the case, those tests should be thrown out in the test vector parser, rather than forcing Rust to filter out unused tests (adding a change to the test vector parser to this PR is okay if that's what needs to happen).

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+        for aes_gcm_test in &aes_gcm_tests {
+            test_counter += 1;
+            log::info!("Test counter: {}", test_counter);
+            if is_input_empty(&aes_gcm_test) {
+                continue;
+            }
+            run_aes_gcm_testcase(aes_gcm_test, opts, transport, &mut fail_counter)?;
+        }
+    }
+    assert_eq!(
```
RyanTorok, line 215: Just FYI: I change the output format for the test harnesses in #21923, that way it's possible to see which tests failed in CI while still suppressing all the UART output in Azure pipelines. Nothing there should block this PR, but if you're so inclined you can update this test harness to use the format we use over there to avoid having to rebase later :)

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
sakib-inspektors: I need some discussion over it.

`cryptolib` accepts empty inputs, but there are around 10% empty input test vectors out of 48k test vectors. Parsing each test vector roughly takes 0.75 seconds time so I was wondering whether we really should pass them to cryptolib or not.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
sakib-inspektors: Yes. I copied some portion of the codes as I am not familiar with rust. :)

I'll update the comments accordingly.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
sadmansakib-git: I need some discussion over it.

cryptolib accepts empty inputs, but there are around 10% empty input test vectors out of 48k test vectors. Parsing each test vector roughly takes 0.75 seconds time so I was wondering whether we really should pass them to cryptolib or not.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
sadmansakib-git: Yes. I copied some portion of codes from the earlier test harness references as I am new to rust.

I'll update the comment accordingly. Actually missed it to modify.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
sadmansakib-git: Removing this function at this moment. We can continue our discussion here and if needed we can make desired changes.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
sadmansakib-git: I think I will go with the new test harness format. Thanks for the heads up!

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
sadmansakib-git: Could you refer to any existing test harness that reflects your suggestion? I guess that would be better.

sw/device/tests/crypto/cryptotest/firmware/aes.c
sadmansakib-git: Tested with partition size 2 and looks like it works.

I noticed we split the AES-GCM test vectors into more than two parts (10 for AAD and 20 for input strings) in unit tests. I am okay going in any direction though.

sw/host/tests/crypto/aes_gcm_nist_kat/src/main.rs
RyanTorok: You can look at what I do in `sw/host/tests/crypto/hash_kat/src/main.rs` on line 95:
```
CryptotestHashMessage {
        message: ArrayVec::try_from(test_case.message.as_slice()).unwrap(),
        message_len: test_case.message.len(),
    }
    .send(&*uart)?;
```
We create the `ArrayVec` with the message and just `unwrap()` it, which will panic if the input is too long. The `ujson_rust` Bazel rule automatically generates the `CryptotestHashMessage` type, where the `message` field is specified to be an ArrayVec with generic constant length equal to the size specified in the uJSON C header (`hash_commands.h`).

RyanTorok: If you want to keep the error message you had before, you can replace the `unwrap()` with `expect("<message>")`.

sadmansakib-git: Ah I see. Yes I saw that file but was not sure you are referring to this as there's no `expect` wording there.

Sounds like a reasonable suggestion. Updating it soon. Thanks!

RyanTorok: Yeah, sorry about the confusion there; I had forgotten I only used `unwrap()`.

jadephilipoom (commented):
Thanks @sadmansakib-git ! Because of the timeouts, CI won't be running these tests, so: did you run the test vectors locally and did they pass?

nasahlpa:
Closing as done in https://github.com/lowRISC/opentitan/pull/28484.

PR #22150, opened 2024-03-20
[ottool] add command to read SAM3X firmware version

timothytrippel:
This adds a top-level ottool command to read the SAM3X firmware version.

You can use it by running `bazel run //sw/host/opentitantool -- --interface=cw310 fpga get-sam3x-fw-version`


sw/host/opentitantool/src/command/fpga.rs
```diff
@@ -11,7 +11,8 @@ use opentitanlib::app::command::CommandDispatch;
 pub enum FpgaCommand {
     LoadBitstream(crate::command::load_bitstream::LoadBitstream),
     ClearBitstream(crate::command::clear_bitstream::ClearBitstream),
-    ResetSam3x(crate::command::reset_sam3x::ResetSam3x),
+    GetSam3xFwVersion(crate::command::sam3x::GetFwVersion),
```
a-will, line 14: May I recommend `TransportCommand::GetFwVersion` instead? This isn't really related to the FPGA.

timothytrippel: Yeah I thought about that, but then I was unsure how to handle the hyper310 case. The nice thing is the hyp firmware version is a direct bazel dep that is pinned in the codebase, so I figured this would be less useful (otlib currently drops a warning if hyp fw is old).

timothytrippel: Also, I wanted to put it next to the resetSam3X command... maybe that should be moved over to the transport subcommand too since it is not the FPGA itself perse?

a-will: Haha, yeah, I think it ought to be under the transport subcommand too

sw/host/opentitanlib/src/transport/chip_whisperer/mod.rs
```diff
@@ -166,6 +166,12 @@ impl<B: Board + 'static> Transport for ChipWhisperer<B> {
             usb.spi1_enable(false)?;
             usb.clear_bitstream()?;
             Ok(None)
+        } else if action.downcast_ref::<GetSam3xFwVersion>().is_some() {
+            let usb = self.device.borrow();
+            //let (x, y, z) = usb.get_firmware_version()?;
+            //println!("Version {:?}.{:?}.{:?}", x, y, z);
```
a-will, line 172: Leftover comments, perhaps?

a-will (approved):
I'll go ahead and approve. I think it's where they belong, but it's up to you whether you want to organize the new commands under the transport command group.

sw/host/opentitanlib/src/transport/chip_whisperer/mod.rs
timothytrippel: Oops, meant to clean that up. Thanks. fixed.

sw/host/opentitantool/src/command/fpga.rs
pamaury: Since the output of the command can depend on the transport, I think hyperdebug could use this command as well and simply print both the SAM3x and the hyperdebug firmware version?

pamaury (approved):
I think I would also put it under the transport group of commands and make it more generic so that other transport can decide to implement it (no need to do in this PR).

timothytrippel:
Sg, I will move both of these commands over to the transport subcommand in a followup. Thanks all!

PR #22151, opened 2024-03-20
[opentitanlib] ECDSA/RSA/SPX raw key helper functions

cfrantz:
1. Add a `Raw` struct type for each of the key types that is both serde serializable and convertible to/from binary form.
2. Add `TryFrom` conversions between the native key types and the `Raw` structs.
3. Implement `FromStr` conversions that load the Raw key material from a file.
4. Add a serde helper to allow serde deserialization to use the `FromStr` trait to initialize structs based either on their `FromStr` conversion or a json key-value map.

### DIFF
--- a/sw/host/opentitanlib/BUILD
+++ b/sw/host/opentitanlib/BUILD
@@ -212,6 +212,7 @@ rust_library(
         "src/util/printer.rs",
         "src/util/raw_tty.rs",
         "src/util/rom_detect.rs",
+        "src/util/serde.rs",
         "src/util/status.rs",
         "src/util/testing.rs",
         "src/util/unknown.rs",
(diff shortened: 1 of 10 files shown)


moidx (approved):
Thanks @cfrantz!

Does this also support printing the keys as bignum little-endian strings?

cfrantz:
> Thanks @cfrantz!
> 
> Does this also support printing the keys as bignum little-endian strings?

The json-serialized form will print the key material as little-endian hex-strings:
```
opentitantool ecdsa key show foo.der
{
  x: "6d1b408188f2b22120250294e22bafe7d543fd5229e0a55109317c304fb8f7b3",
  y: "8996a09cd07ff74345caa284f9851873f68e24cb4cb1919e0e8cd208d13f0e4c"
}
```

timothytrippel (approved):

PR #22156, opened 2024-03-20
[bazel] And KeyInfo provider and associated rules.

moidx:
Introduce the following Bazel rules:

- `key_ecdsa`: Defines an ECDSA object that can be used for sign and verify operations.
- `key_sphincs_plus`: Defines a SPHINCS+ object that can be used for sign and verify operations.

This PR also introduces `key_sphincs_plus` targets for ROM SPX+ keys.

### DIFF
--- a/rules/opentitan/keyutils.bzl
+++ b/rules/opentitan/keyutils.bzl
@@ -10,24 +10,38 @@ KeyInfo = provider(
     Used to capture all required information about a key.
     """,
     fields = {
-        "key_info": "Info associated with the key.",
+        "id": "Identifier used by the consumers of the provider to determine the key algorithm.",
+        "config": "The config of the key. Specific to the key algorithm.",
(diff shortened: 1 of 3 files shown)


rules/opentitan/keyutils.bzl
```diff
@@ -1,8 +1,100 @@
...
+    key_obj = {
+        "id": "sphincs_plus",
+        "config": ctx.attr.config,
+        "method": ctx.attr.method,
+        "pub_key": ctx.file.pub_key.path,
+        "type": ctx.attr.type,
+    }
+    if ctx.attr.private_key:
+        key_obj["private_key"] = ctx.file.private_key.path
+    return [KeyInfo(key_info = key_obj)]
```
cfrantz, line 27: Rather than using a dict to hold the fields, should we just make each of the properties provider-level fields?

cfrantz (approved):

PR #7025, opened 2021-06-17
[sw/silicon_creator] Add entry_point, code_start, and code_end fields to the manifest

alphan:
This change
- Adds entry_point, code_start, and code_end to the manifest along with getters and unit tests,
- Updates the signer tool to set these fields by parsing ELF files, and
- Removes the extra entry in ROM_EXT interrupt vector since it is no longer needed.

This change also adds a new rust dependency (`object` crate) for parsing ELF files.

PR #22158, opened 2024-03-20
[sw] Require mere 2-byte alignment of entry point

jesultra:
Risc-V instructions can either be 2 bytes or 4 bytes, and can be aligned on 2-bytes boundaries.  It seems that Clang will align functions to 2 bytes.

Is there any reason to require 4-byte alignment of the entry point? Adhering to such restriction will likely require manual linker script tuning by owners.  I propose allowing any 2-byte aligned address to serve as the entry point.

### DIFF
--- a/sw/device/silicon_creator/lib/manifest.h
+++ b/sw/device/silicon_creator/lib/manifest.h
@@ -481,10 +481,10 @@ inline rom_error_t manifest_check(const manifest_t *manifest) {
     return kErrorManifestBadCodeRegion;
   }
 
-  // Entry point must be inside the executable region and word aligned.
+  // Entry point must be inside the executable region and be an even address.
   if (manifest->entry_point < manifest->code_start ||
       manifest->entry_point >= manifest->code_end ||
-      (manifest->entry_point & 0x3) != 0) {
+      (manifest->entry_point & 0x1) != 0) {
     return kErrorManifestBadEntryPoint;
   }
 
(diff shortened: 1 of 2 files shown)


cfrantz:
I don't recall the exact rationale, but I believe we specifically chose a 4-byte alignment for the entry point.

The choice may have either been related to hardening or to the way we handle code fetches out of flash.

jesultra:
> > I don't recall the exact rationale, but I believe we specifically chose a 4-byte alignment for the entry point.
> > The choice may have either been related to hardening or to the way we handle code fetches out of flash.
> 
> It seems to be related to the ePMP: [#7025 (comment)](https://github.com/lowRISC/opentitan/pull/7025#discussion_r658708427)

Code segment start and end must by 4-byte aligned, in order for the ePMP to be able to be configured to allow executing that range, and nothing more.  However, I do not see how ePMP restrictions would affect the alignment of the entry point within that aligned range.   The linked comment explicitly mention start and end only, and not the entry point.  I suspect that the same alignment requirements were applied to the entry point without much thought.  (In fact, I do not see any security issue with ROM_EXT allowing odd addresses for the entry point, at worst, it will trigger immediate trap due to unaligned read, but we might as well report a nice error code.)

jesultra:
While I still think that requiring 2-byte alignment is sufficient, I also do not want to change something that works, and adhering to 4-byte alignment is not a huge burden on owner code.

PR #22164, opened 2024-03-21
[topgen] Avoid adding a comma after the last parameter in the list

rswarbrick:
The previous code added a comma after a parameter unless this was the last parameter of the module and this was the last module in the list. This works fine unless the last module in the list doesn't happen to have any parameters.

In that situation, you got something like

     ...
      parameter int unsigned RvCoreIbexDmExceptionAddr =
          tl_main_pkg::ADDR_SPACE_RV_DM__MEM + dm::ExceptionAddress[31:0],
      parameter bit RvCoreIbexPipeLine = 0,
      // parameters for mbx0
      // parameters for mbx1
      // parameters for mbx2
    ) (
     ...

which is a syntax error (the parser sees something like "a, b, )").

Tweak the logic slightly so that we add a comma unless this is the last parameter in the module and this is the last module *with parameters* in the list

The diagnosis and main work was done by Robert Schilling. I just tweaked the code to make it a bit easier for me to follow.

(And I stole the idea from him! See #22130 for the work that prompted me to take a look)

### DIFF
--- a/util/topgen/templates/toplevel.sv.tpl
+++ b/util/topgen/templates/toplevel.sv.tpl
@@ -37,6 +37,8 @@ for m in top['memory']:
   if m['type'] == 'rom':
     has_toplevel_rom = True
 
+last_modidx_with_params = lib.idx_of_last_module_with_params(top)
+
 %>\
 `include "prim_assert.sv"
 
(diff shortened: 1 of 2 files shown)


util/topgen/templates/toplevel.sv.tpl
```diff
@@ -37,6 +37,8 @@ for m in top['memory']:
   if m['type'] == 'rom':
     has_toplevel_rom = True
 
+last_module_with_params = lib.idx_of_last_module_with_params(top)
```
Razer6, line 40: maybe factor `idx` into the name to indicate its just an integer and not a full module.

Razer6 (approved):
Awesome that you picked that up. Happy to merge yours.

rswarbrick:
Thanks for the responses. Good point about the commit author: looks like I was overly enthusiastic about stealing @Razer6's work :-)

Also good point about the variable rename: I'll do that now.

PR #22167, opened 2024-03-21
[uart] Bump UART to v2.0.0 and back to D2S due to interrupt changes

GregAC:


andreaskurth (approved):
LGTM, though the `commit_id` isn't necessary and we'll probably remove it in the longer term (and instead rely on tooling to extract the Git hash at which an IP was tagged as L3, D3, or V3 -- which removes a manual, error-prone step).

andreaskurth:
CI failures are unrelated to this PR --> merging

PR #22169, opened 2024-03-21
Revert "[dv] Remove phase argument from monitor's collect_trans"

rswarbrick:
This reverts commit d39c54d91dad7ef1b9dd2c617dccc2c354c0939f.

The change was reasonable, but it touched some vendored code and some unrelated CI breakage meant that I didn't spot the CI check that would have told me about it. Oops!

The easiest way to fix this properly is probably to implement the same change on the Ibex side first, so let's back out the change at this end for now.


hcallahan-lowrisc (approved):
As it's a revert, LGTM.

PR #22174, opened 2024-03-21
[vendor] Apply patches to "properly do" local edits to vendored Ibex

rswarbrick:
Embarrassingly, there are two places where I have accidentally edited vendored code from Ibex. Unfortunately, our CI didn't catch the mistake... until today!

### DIFF
--- a/hw/vendor/lowrisc_ibex.vendor.hjson
+++ b/hw/vendor/lowrisc_ibex.vendor.hjson
@@ -4,6 +4,7 @@
 {
   name: "lowrisc_ibex",
   target_dir: "lowrisc_ibex",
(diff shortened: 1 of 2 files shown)


GregAC (commented):
Is this missing a 0001 patch that didn't get added to the commit?

rswarbrick:
Sort of: I cheated and that patch is in #22175, but isn't needed for the initial "getting stuff aligned" work :-)

rswarbrick:
Hmm, this is all rather confusing. I think I got completely confused yesterday! Let's drop this and forget about it for now: I can always try again if I work out what on earth I was seeing...

PR #22175, opened 2024-03-21
[dv] Remove phase argument from monitor's collect_trans

rswarbrick:
The path to this PR was a bit of a mess (caused by me). But the end result is hopefully reasonably clean. The commit message of the single commit left is:

**[dv] Remove phase argument from monitor's collect_trans**

This argument isn't actually used anywhere it's passed. Let's clean it
up.

This patch is *almost* trivial, except that it touches some vendored
code from Ibex. Apply a patch to the vendored code to make it match
the function signature that we expect.


hcallahan-lowrisc (approved):
Test failures should be unrelated.

PR #22177, opened 2024-03-21
[dv,uart] Aid `uart_stress_all_with_rand_reset`

elliotb-lowrisc:
Get the `uart_stress_all_with_rand_reset` test pass-rate higher by breaking out of sequence loops and increasing polling delays when we want to reset (as indicated by `cfg.stop_transaction_generators()`).

The key to getting the pass rate higher is to avoid the case where `wait_to_issue_reset()` times-out before it sees a large enough gap (defined by `CyclesWithNoAccessesThreshold`) between register accesses. The primary approach to achieving this idleness is to stop the test sequences early by breaking out of loops when they can do so cleanly. The secondary approach taken here is to increase register polling delays, which covers the cases where the test sequence would not reach a clean break point before the reset timeout. Low uart baud rates and sequences with long transactions make this additional method critical to getting the pass rate above 90%.

### DIFF
--- a/hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
+++ b/hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
@@ -134,20 +134,51 @@ class uart_base_vseq extends cip_base_vseq #(.CFG_T               (uart_env_cfg)
     cfg.m_uart_agent_cfg.set_baud_rate(baud_rate);
   endtask
 
-  virtual task spinwait_txidle();
-    if (ral.ctrl.tx.get_mirrored_value()) begin
-      // use a very big timeout as it takes long time to flush all the items
-      csr_spinwait(.ptr(ral.status.txidle), .exp_data(1'b1),
-                   .timeout_ns(TxFifoDepth * 1_250_000),
-                   .spinwait_delay_ns($urandom_range(0, 1000)));
+  task wait_if_stop_transaction_generators();
+    if (cfg.stop_transaction_generators()) begin
+      // Increase polling interval to allow requested reset without breaking test sequence
+      cfg.clk_rst_vif.wait_clks(CyclesWithNoAccessesThreshold * 2);
     end
   endtask
 
+  task reset_aware_simple_spinwait(input uvm_object ptr,
+                                   input uvm_reg_data_t exp_data,
+                                   input uint spinwait_delay_ns = 0,
+                                   input uint timeout_ns = 10_000_000);
+    fork
+      begin : isolation_fork
+        bit [TL_DW-1:0] rdata;
+        csr_field_t csr_or_fld = decode_csr_or_field(ptr);
+        fork
+          while (!cfg.under_reset) begin
+            if (spinwait_delay_ns) #(spinwait_delay_ns * 1ns);
+            wait_if_stop_transaction_generators();
+            csr_rd(.ptr(ptr), .value(rdata));
+            if (rdata == exp_data) break;
+          end
+          begin
+            `DV_WAIT_TIMEOUT(timeout_ns, `gfn,
+                             $sformatf("timeout %0s (addr=0x%0h) == 0x%0h",
+                                       ptr.get_full_name(), csr_or_fld.csr.get_address(), exp_data))
+          end
+        join_any
+        disable fork;
+      end : isolation_fork
+    join
+  endtask
+
+  virtual task spinwait_txidle();
+    if (!ral.ctrl.tx.get_mirrored_value()) return;
+    // use a very big timeout as it takes long time to flush all the items
+    reset_aware_simple_spinwait(.ptr(ral.status.txidle), .exp_data(1'b1),
+                                .timeout_ns(TxFifoDepth * 1_250_000),
+                                .spinwait_delay_ns($urandom_range(0, 1000)));
+  endtask
+
   virtual task spinwait_rxidle();
-    if (ral.ctrl.rx.get_mirrored_value()) begin
-      csr_spinwait(.ptr(ral.status.rxidle), .exp_data(1'b1),
-                   .spinwait_delay_ns($urandom_range(0, 1000)));
-    end
+    if (!ral.ctrl.rx.get_mirrored_value()) return;
+    reset_aware_simple_spinwait(.ptr(ral.status.rxidle), .exp_data(1'b1),
+                                .spinwait_delay_ns($urandom_range(0, 1000)));
   endtask
 
   // task to send a byte of data out of dut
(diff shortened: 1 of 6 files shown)


hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
```diff
@@ -134,20 +134,46 @@ class uart_base_vseq extends cip_base_vseq #(.CFG_T               (uart_env_cfg)
...
+            if (cfg.stop_transaction_generators()) begin
+              // Increase polling interval to allow requested reset without breaking test sequence
+              cfg.clk_rst_vif.wait_clks(CyclesWithNoAccessesThreshold * 2);
+            end
+            csr_rd(.ptr(ptr), .value(rdata));
+            if (rdata == exp_data) break;
+          end
+          begin
+            `DV_WAIT_TIMEOUT(timeout_ns, `gfn, $sformatf("timeout %0s (addr=0x%0h) == 0x%0h",
+                ptr.get_full_name(), csr_or_fld.csr.get_address(), exp_data))
```
rswarbrick, line 157: Nit: I think the indentation is off here.

hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
```diff
@@ -134,20 +134,46 @@ class uart_base_vseq extends cip_base_vseq #(.CFG_T               (uart_env_cfg)
...
+        csr_field_t csr_or_fld = decode_csr_or_field(ptr);
+        fork
+          while (!cfg.under_reset) begin
+            if (spinwait_delay_ns) #(spinwait_delay_ns * 1ns);
+            if (cfg.stop_transaction_generators()) begin
+              // Increase polling interval to allow requested reset without breaking test sequence
+              cfg.clk_rst_vif.wait_clks(CyclesWithNoAccessesThreshold * 2);
+            end
+            csr_rd(.ptr(ptr), .value(rdata));
+            if (rdata == exp_data) break;
```
rswarbrick, line 153: Would it make more sense to move this before the initial delay? I feel like "Wait until X is true" should take no time if X is currently true.

hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
```diff
@@ -239,6 +265,10 @@ class uart_base_vseq extends cip_base_vseq #(.CFG_T               (uart_env_cfg)
         if (cfg.under_reset) break;
         `DV_CHECK_MEMBER_RANDOMIZE_FATAL(dly_to_access_fifo)
         cfg.clk_rst_vif.wait_clks(dly_to_access_fifo);
+        if (cfg.stop_transaction_generators()) begin
+          // Increase polling interval to allow requested reset without breaking test sequence
+          cfg.clk_rst_vif.wait_clks(CyclesWithNoAccessesThreshold * 2);
+        end
```
rswarbrick, line 271: Would it make sense to pull this out into a separate task called something like `wait_if_transaction_generators_stopped`? I notice that we've got the same lines in `reset_aware_simple_spinwait` as well.

hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
```diff
@@ -25,10 +25,12 @@ class uart_intr_vseq extends uart_base_vseq;
     // will inject parity/stop error in this case
     cfg.m_uart_agent_cfg.en_rx_checks = 0;
     for (int i = 1; i <= num_trans; i++) begin
+      if (cfg.stop_transaction_generators()) break;
```
rswarbrick, line 28: Is this the right behaviour? I think it means we'll finish the `body` task instantly if the configuration bit becomes true. (Genuine question:) Is this what we want?

(Similar question for some of the other vseqs)

hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
elliotb-lowrisc: Ah perhaps, but this is the way that `csr_spinwait` does it. Delay first, then test. My preference would be to maintain behaviour consistent with `csr_spinwait`. I have no strong feelings wither way beyond that.

hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
elliotb-lowrisc: Hmm, this also seems to be from me mimicking `csr_spinwait`. Happy to change this if you think aligning with the contents of the brackets would be better.

hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
elliotb-lowrisc: Sounds like a good idea to me

hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
elliotb-lowrisc: The intent with these `if (cfg.stop_transaction_generators()) break;` lines is that they will stop the sequence from generating new transactions at points in loops where they might naturally choose to stop if their loop condition were no longer met. I got the idea from the `sram` and `otp` code, which does this in similar places.

For compactness, I suppose in the case of a `for` loop I could combine lines to get:
```
for (int i = 1; i <= num_trans && !cfg.stop_transaction_generators(); i++) begin
```

I also wonder if some of these could be removed now that we have the register polling interval increases, but I don't think there is a strong reason to remove them. It feels like they should be there and are not worth the time to figure out which are which are required and which are not.

hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
rswarbrick: I think that's probably cleaner. Something like this, maybe?
```systemverilog
          begin
            `DV_WAIT_TIMEOUT(timeout_ns, `gfn,
                             $sformatf("timeout %0s (addr=0x%0h) == 0x%0h",
                                       ptr.get_full_name(), csr_or_fld.csr.get_address(), exp_data))
          end
```

hw/ip/uart/dv/env/seq_lib/uart_base_vseq.sv
rswarbrick: Good point! And that behaviour goes back to the beginning of time! Maybe you're right and it makes sense to keep consistent. We can always file a follow-up change that rejigs both sides to (what I think is) a more sensible order afterwards.

hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
rswarbrick: Definitely no need to collapse onto a single line: I prefer the style you've got at the moment. I've just looked through the repo more closely and I'm convinced: this looks consistent to me :-)

elliotb-lowrisc:
Put conditional register polling delay in a task.
Adjusted indentation of a line.
Rebased.

rswarbrick:
Looks really good, thanks. The CI failure is FPGA tests (definitely not related to this change!) and Kokoro (which looks like a tooling issue).

rswarbrick (approved):

PR #22180, opened 2024-03-21
[doc] Update Technical Committee members

mundaym:


doc/project_governance/technical_committee.md
```diff
@@ -28,5 +28,5 @@ The OpenTitan Technical Committee membership is:
 * Neeraj Upasani
 * Nir Tasher
 * Arnon Sharlin
-* Sivakumar
 * Tim Trippel
+* Miles Dai
```
andreaskurth, line 32: Somewhat nitty, but I think the members are currently sorted by last name, so I think we should keep it that way

rswarbrick (approved):

PR #22183, opened 2024-03-21
[cryptotest] Filter out invalid AES-GCM test vectors

sadmansakib-git:
The cryptolib API for AES-GCM supports IV length of strictly 3 or 4 words (hex length between [24, 32]).

Reference: https://github.com/lowRISC/opentitan/blob/9b69f99be2bf3ba72e1b74fec0ce6d8197bbd1ce/sw/device/lib/crypto/impl/aes_gcm/aes_gcm.c#L308

Out of existing 47250 test vectors, 31500 test vectors do not satisfy this condition. It leads us having 15750 valid test vectors. We should only keep the valid ones.


sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
```diff
@@ -15,12 +15,27 @@
...
     test_cases = list()
-    for test_case_id, test_vec in enumerate(raw_testcases):
+    valid_test_case_id = 0
+    for test_vec in raw_testcases:
+        if is_iv_len_invalid(len(test_vec["IV"])):
+            continue
+        valid_test_case_id += 1
         test_case = {
-            "test_case_id": test_case_id,
+            "test_case_id": valid_test_case_id,
```
RyanTorok, line 38: For NIST test vector files that don't include a `COUNT` variable, I usually count the invalid tests, too. The test harness already prints a test counter (counting only valid tests). The purpose of the test case ID is to make it easier to find the right test vector in the source rsp file when debugging a failure e.g. on CI, so it's perfectly fine to skip values.

sadmansakib-git: I used the `test_case_id` a little bit differently.

In NIST AES-GCM rsp files, we do have COUNT variable but it resets back to 0 each time some other parameters (i.e. tag length/plaintext length/ciphertext length etc) change. So the value of this variable is not consistent across the file.

To better debug and maintain a universal test case count, I only included the test cases having valid iv length so that we can have a consistent test case id in the parsed json files (the main food for the harness).

After that if anything bad happens from the test harness side, I can easily check the parsed json file and navigate directly to the test vector with the test_case_id, as it now acts like a primary key.

RyanTorok: If the `COUNT` values repeat, I usually do so with the test case IDs, too. The values don't need to be unique.

sadmansakib-git: But a unique value might come more handy at debugging, no? Or are we supposed to keep it consistent with the "rsp files"?

RyanTorok: When `COUNT` repeats, I usually print some other info to uniquely identify the test (i.e. `test_case_id` is what test it is within its group_). If you look at #21923, I update most of the test harnesses to print out other parameters to ensure the test can be uniquely identified. For GCM, I would include the IV size in the output.

sadmansakib-git: Applied the suggestion. Skipped the values even if the test vector is invalid.

I will include the IV size in the debug log along with the key length in the test harness PR.

RyanTorok (approved):

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
sadmansakib-git: Updated the test harness with detailed print-outs in #22148. Noticed that only IV and key length are not enough, as they are not unique across groups.

RyanTorok: Not surprising. I was just listing those values as examples.

sadmansakib-git: Yeah. Just to make it clear here for the future readers.

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
```diff
@@ -15,10 +15,29 @@
 from cryptotest_util import parse_rsp, str_to_byte_array
 
 
+def is_iv_len_invalid(iv_len) -> bool:
+    """IV length checker for AES-GCM testvectors.
+
+    According to cryptolib API implementation, the only supported
+    IV lengths are either 3 or 4 words (hex length between [24, 32]).
+    Reference: sw/device/lib/crypto/impl/aes_gcm/aes_gcm.c
```
jadephilipoom, line 23: nit: could you add a comment in this file that links here and says that if that behavior changes to accept more IV lengths, then this code should change too?

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
```diff
@@ -15,10 +15,29 @@
 from cryptotest_util import parse_rsp, str_to_byte_array
 
 
+def is_iv_len_invalid(iv_len) -> bool:
+    """IV length checker for AES-GCM testvectors.
+
+    According to cryptolib API implementation, the only supported
+    IV lengths are either 3 or 4 words (hex length between [24, 32]).
```
jadephilipoom, line 22:
```suggestion
    IV lengths are either 3 or 4 words (hex length exactly 24 or 32).
```

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
```diff
@@ -15,10 +15,29 @@
...
+    Reference: sw/device/lib/crypto/impl/aes_gcm/aes_gcm.c
+
+    """
+    return iv_len < 24 or iv_len > 32
```
jadephilipoom, line 26: I don't think this actually matches the behavior in cryptolib; an `iv_len` between 24 and 32 would also not be accepted.
```suggestion
    return iv_len != 24 and iv_len != 32
```

sw/host/cryptotest/testvectors/parsers/nist_cavp_aes_gcm_parser.py
```diff
@@ -15,10 +15,29 @@
...
     test_cases = list()
-    for test_case_id, test_vec in enumerate(raw_testcases):
+    count = 0
+    for test_vec in raw_testcases:
+        if "Count" in test_vec:
+            test_case_id = int(test_vec["Count"])
+        else:
+            test_case_id = count
+        count += 1
+        if is_iv_len_invalid(len(test_vec["IV"])):
```
jadephilipoom, line 39: I'm concerned about just reading the length of the hex string to get the length here. For example, if the hex string starts with 0, then Python might not print the leading 0 and it would show up as invalid. Most test vectors probably include the length as a separate field; I'd save that value and use it to filter instead if possible.

milesdai (approved):
LGTM modulo Jade's comments

PR #22184, opened 2024-03-21
[dv,usbdev] Fixes for usbdev DV sequences

alees24:
This PR addresses the following issues three separate issues in individual commits:

- IN side sequences were conflicting with the usb20_driver behavior by needlessly trying to clear interrupts whilst the drive is causing interrupts by performing a Bus Reset on the USB. This led to seed-dependent failures.
- Warnings in the vcs build log had become quite numerous; eliminate them to lessen the risk of losing warnings that indicate functional issues.
- Avoid the use of csr_update to write to the FIFOs because this can lead to the csr/uvm logic deciding that update is not required, and thus the write of a buffer number into the FIFO does not occur. When buffer number 0 is chosen this leads to test failures, but that is a perfectly valid buffer number and does/will get selected by randomization.


rswarbrick (approved):
Looks good to me. And I've just learned about `uvm_object::pack`, which was rather new to me!

PR #22185, opened 2024-03-21
[python] Fix FuseSoC version

andreaskurth:
Prior to this commit, the FuseSoC version downloaded when running

    pip install -r python-requirements.txt

would identify itself as **version `0.4.dev0`** to Pip as well as in `fusesoc --version`.  This is problematic in at least two ways:

1. Because Pip gets told that it downloaded version 0.4, it can prefer to use a locally cached archive (and I've seen it doing that!) that is *really* version 0.4, causing the update to become a NOP even though the user thinks they just updated FuseSoC.

2. Users have no way of distinguishing between versions 0.4 and 0.5 from the command line, which makes debugging issues that got fixed in version 0.5 difficult.

To work around this problem, the PR lowRISC/fusesoc#6 changed `fallback_version`, from which our FuseSoC gets its version, to `0.5.dev0`, and this version was subsequently tagged as such.  This commit now updates our Python requirements to load the fixed version.

Run `pip install -r python-requirements.txt` after this commit has landed to update FuseSoC to version `0.5.dev0` for good.

### DIFF
--- a/python-requirements.in
+++ b/python-requirements.in
@@ -61,7 +61,7 @@ types-tabulate==0.8.11
 anytree==2.8.0
 
 # Development version with OT-specific changes
-https://github.com/lowRISC/fusesoc/archive/refs/tags/ot-0.5.zip
+https://github.com/lowRISC/fusesoc/archive/refs/tags/ot-0.5.dev0.zip
 
 # Development version with OT-specific changes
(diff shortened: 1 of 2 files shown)


a-will (approved):

andreaskurth:
Failing CI checks are unrelated. Merging

PR #22186, opened 2024-03-21
[dv,usbdev] DV sequence for packet buffer memory

alees24:
This PR adds a DV test sequence for exercising the packet buffer memory:

- Randomized IN and OUT transactions of random length/content, to randomized buffers and randomized endpoints.
- CSR-side randomized packet buffer reads/writes, ensuring no collision with USB-side access.
- Checking of all data transfers, in both directions and on both CSR and USB sides.
- Aim to induce access collisions (same cycle, don't need to worry about same address); CSR side should be stalled because the packet buffer is implemented using a Single Port RAM now.
- CSR-side access to the packet buffer memory only worked if the scoreboard/UVM was capable of predicting the memory contents; this is not the case when the DUT hardware may write to the packet buffer memory at any moment.
- usbdev_scoreboard is therefore updated to avoid predictions, making writing possible and relying upon the vseq itself to check the returned read data.
- Constraint on 4n packet length is eliminated, and the usbdev_base_vseq tasks 'write_buffer' and 'read_check_buffer' should now be as generic as required for any test sequence.

The changes in the 'usbdev_base_vseq' are more forward-thinking and some of the functionality is not yet required, but as the testing moves to being more thorough, covering more endpoints etc. these tasks should prove useful in the base sequences.

Note that some of the constants used within the DV environment should be lifted from the DUT package/constants in the RTL but presently non-essential changes to the RTL are to be discouraged.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -98,6 +98,26 @@ virtual task dut_init(string reset_kind = "HARD");
   if (do_usbdev_init) usbdev_init();
 endtask
 
+  // Wait for the DUT to enter one of the given link states, with an optional timeout
+  //   (the default of -1 disables the timeout).
+  // Link state transitions occur in response to bus activity from the host driving the USB.
+  virtual task wait_for_link_state(usbdev_link_state_e states[], int timeout_clks = -1,
+                                   bit fatal = 1);
+    // Wait until the given state is attained or the (optionally
+    usbdev_link_state_e rd;
+    do begin
+      uvm_reg_data_t data;
+      // Delay and if timeout was requested, decrease the cycle count.
+      cfg.clk_rst_vif.wait_clks(128);
+      if (timeout_clks >= 0) timeout_clks -= (timeout_clks > 128) ? 128 : timeout_clks;
+      csr_rd(.ptr(ral.usbstat.link_state), .value(data));
+      rd = usbdev_link_state_e'(data);
+    end while (!(rd inside {states}) && (timeout_clks != 0));
+    if (fatal && !(rd inside {states})) begin
+      `dv_fatal($sformatf("DUT did not enter expected link state (still in %d)", rd))
+    end
+  endtask
+
   // Construct and transmit a token packet to the USB device
   virtual task call_token_seq(input pid_type_e pid_type);
     `uvm_create_on(m_token_pkt, p_sequencer.usb20_sequencer_h)
@@ -148,13 +168,91 @@ endtask
     finish_item(m_handshake_pkt);
   endtask
 
+  // Read and check the content of the given packet buffer against expectations.
+  task read_check_buffer(bit [4:0] buffer_id, byte unsigned exp_data[]);
+    int unsigned exp_size = exp_data.size();
+    byte unsigned pkt_data[];
+    bit mismatch = 0;
+
+    // Collect the packet data from the buffer within the DUT.
+    read_buffer(buffer_id, exp_size, pkt_data);
+    `DV_CHECK_EQ_FATAL(pkt_data.size(), exp_size, "Unexpected packet size")
+
+    // Check buffer content against packet data.
+    for (int unsigned idx = 0; idx < exp_size; idx++) begin
+      `uvm_info(`gfn, $sformatf("Checking %2d: 0x%02x against 0x%02x",
+                                idx, pkt_data[idx], exp_data[idx]), UVM_HIGH)
+      // `DV check on actual and exp data
+      mismatch |= (pkt_data[idx] !== exp_data[idx]);
+    end
+    // Now that all data has been reported, stop if anything mismatched.
+    `DV_CHECK_EQ(mismatch, 0, "Received buffer contents do not match expected data")
+  endtask
+
+  // Write some data into the USB device packet buffer memory.
+  task write_buffer(bit [4:0] buffer_id, byte unsigned pkt_data[]);
+    int unsigned pkt_size = pkt_data.size();
+    int unsigned offset;
+    `DV_CHECK(pkt_size <= MaxPktSizeByte);
+
+    // Calculate start address (in 32-bit words) of buffer.
+    offset = buffer_id * 'h10;
+
+    `uvm_info(`gfn, $sformatf("write_buffer id %d with %d bytes", buffer_id, pkt_size), UVM_MEDIUM)
+    for (int unsigned idx = 0; idx < pkt_size; idx++) begin
+      `uvm_info(`gfn, $sformatf("%2d: 0x%02x", idx, pkt_data[idx]), UVM_HIGH)
+    end
+
+    // Note that unfortunately we can write only complete data words to the DUT; partial word
+    // writes would raise an error if attempted using 'tl_access' because the packet buffer memory
+    // itself does not offer lane enables.
+    for (int unsigned idx = 0; idx < pkt_size; idx += 4) begin
+      // Number of valid bytes within this word.
+      int unsigned nb = (pkt_size - idx >= 4) ? 4 : (pkt_size - idx);
+      bit [TL_DBW-1:0] mask = TL_DBW'((1 << nb) - 1);
+      // Populate only the valid parts of the write data word, and importantly avoid
+      // out-of-bounds accesses to 'pkt_data.'
+      bit [TL_DW-1:0]  wdata;
+      wdata[7:0] = pkt_data[idx];
+      if (mask[1]) wdata[15:8]  = pkt_data[idx + 1];
+      if (mask[2]) wdata[23:16] = pkt_data[idx + 2];
+      if (mask[3]) wdata[31:24] = pkt_data[idx + 3];
+
+      mem_wr(.ptr(ral.buffer), .offset(offset), .data(wdata));
+      offset++;
+    end
+  endtask
+
+  // Read and return data from the given packet buffer.
+  task read_buffer(bit [4:0] buffer_id, int unsigned len, output byte unsigned data[]);
+    // Calculate start address (in 32-bit words) of buffer
+    int unsigned offset = buffer_id * 'h10;
+
+    `DV_CHECK_FATAL(len <= MaxPktSizeByte);
+    data = new [len];
+
+    for (int unsigned idx = 0; idx < len; idx += 4) begin
+      // Number of valid bytes within this word
+      int unsigned nb = (len - idx >= 4) ? 4 : (len - idx);
+      bit [TL_DBW-1:0] mask = TL_DBW'((1 << nb) - 1);
+      bit [31:0] act_data;
+
+      // We can only read complete bus words from the packet buffer memory, but the USB packet
+      // lengths have byte-level granularity.
+      mem_rd(.ptr(ral.buffer), .offset(offset), .data(act_data));
+      data[idx] = act_data[7:0];
+      if (mask[1]) data[idx + 1] = act_data[15:8];
+      if (mask[2]) data[idx + 2] = act_data[23:16];
+      if (mask[3]) data[idx + 3] = act_data[31:24];
+      offset++;
+    end
+  endtask
+
   // Check that a SETUP/OUT data packet with the given properties was received and stored by
   // the USB device in its packet buffer memory.
   task check_rx_packet(bit [3:0] endp, bit setup, bit [4:0] buffer_id,
                        byte unsigned exp_byte_data[]);
     uvm_reg_data_t rx_fifo_read;
-    int unsigned   aligned_size;
-    int unsigned   offset;
 
     // Read RX FIFO which should contain a buffer description matching the expectations
     csr_rd(.ptr(ral.rxfifo), .value(rx_fifo_read));
@@ -164,36 +262,7 @@ endtask
     `DV_CHECK_EQ(get_field_val(ral.rxfifo.buffer, rx_fifo_read), buffer_id);
     `DV_CHECK_EQ(get_field_val(ral.rxfifo.size, rx_fifo_read), exp_byte_data.size());
 
-    // Calculate start address (in 32-bit words) of buffer
-    offset = buffer_id * 'h10;
-
-    // TODO: This is partly a limitation of this test code and partly because we may expect to
-    // run into troubles trying to perform complete word reads from a packet buffer memory that
-    // has not been initialized, resulting in the transfer of undefined data even if it is
-    // ultimately unused.
-    aligned_size = exp_byte_data.size() & ~3;
-    if (aligned_size < exp_byte_data.size()) begin
-      `uvm_info(`gfn, $sformatf("WARNING: Packets must be 4n presently; final byte(s) not tested"),
-                UVM_LOW)
-    end
-    for (int unsigned idx = 0; idx < exp_byte_data.size(); idx++) begin
-      `uvm_info(`gfn, $sformatf("%d: 0x%x", idx, exp_byte_data[idx]), UVM_DEBUG)
-    end
-
-    // Check buffer content against packet data
-    for (int unsigned idx = 0; idx < aligned_size; idx += 4) begin
-      bit [31:0] act_data;
-      bit [31:0] exp_data;
-      // Expected value of this word
-      exp_data = {exp_byte_data[idx + 3], exp_byte_data[idx + 2],
-                  exp_byte_data[idx + 1], exp_byte_data[idx]};
-
-      mem_rd(.ptr(ral.buffer), .offset(offset), .data(act_data));
-      `uvm_info(`gfn, $sformatf("Checking 0x%x against 0x%x", act_data, exp_data), UVM_DEBUG)
-      // `DV check on actual and exp data
-      `DV_CHECK_CASE_EQ(act_data, exp_data, "Received buffer contents do not match data packet")
-      offset++;
-    end
+    read_check_buffer(buffer_id, exp_byte_data);
   endtask
 
   // Check that the IN DATA packet transmitted by the USB device matches our expectations
@@ -203,9 +272,10 @@ endtask
     int unsigned len = (act_len < exp_len) ? act_len : exp_len;
 
     `uvm_info(`gfn, $sformatf("IN packet PID 0x%x len %d expected PID 0x%x len %d",
-                              pkt.m_pid_type, act_len, exp_pid, exp_len), UVM_DEBUG)
+                              pkt.m_pid_type, act_len, exp_pid, exp_len), UVM_HIGH)
     `DV_CHECK_EQ(pkt.m_pid_type, exp_pid);
 
+    // Report all of the actual and expected bytes before checking, to aid diagnosis.
     for (int unsigned i = 0; i < exp_data.size(); i++) begin
       `uvm_info(`gfn, $sformatf("%d: 0x%x", i, exp_data[i]), UVM_HIGH)
     end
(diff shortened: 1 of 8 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -98,6 +98,25 @@ virtual task dut_init(string reset_kind = "HARD");
...
+    usbdev_link_state_e rd;
+    do begin
+      uvm_reg_data_t data;
+      // Delay and if timeout was requested, decrease the cycle count.
+      cfg.clk_rst_vif.wait_clks(128);
+      if (timeout >= 0) timeout -= (timeout > 128) ? 128 : timeout;
+      csr_rd(.ptr(ral.usbstat.link_state), .value(data));
+      rd = usbdev_link_state_e'(rd);
+    end while (!(rd inside {states}) && (timeout != 0));
+    if (fatal && !(rd inside {states})) begin
```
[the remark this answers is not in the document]
alees24: Apparently. The tool complained without it, and it is specified in the grammar. I expect it avoids a shift-reduce conflict in LALR(1) parsing, but it does leave it with such ugly contortions as 'inside {[1:10]}'

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -148,13 +167,81 @@ endtask
...
+    // Calculate start address (in 32-bit words) of buffer
+    offset = buffer_id * 'h10;
+
+    `uvm_info(`gfn, $sformatf("write_buffer id %d with %d bytes", buffer_id, pkt_size), UVM_MEDIUM)
+    for (int unsigned idx = 0; idx < pkt_size; idx++) begin
+      `uvm_info(`gfn, $sformatf("%d: 0x%x", idx, pkt_data[idx]), UVM_HIGH)
+    end
+
+    // Note that unfortunately we can write only complete data words to the DUT; partial word
+    // writes raise an error.
```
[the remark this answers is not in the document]
alees24: I need to update that comment, thanks. I tried to use 'tl_access' directly which accepts lane enables, but the USBDEV packet buffer memory does not presently support partial word writes, so I switched it back to mem_wr/rd which supports only full 32-bit (two-state) words.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -148,13 +167,81 @@ endtask
...
+    // Calculate start address (in 32-bit words) of buffer
+    offset = buffer_id * 'h10;
+
+    // Check buffer content against packet data; note that the final word may be incomplete
+    // and that the values of the remaining bytes of that final word are dependent upon the
+    // internal implementation details of the DUT.
+    for (int unsigned idx = 0; idx < exp_size; idx += 4) begin
+      // Number of valid bytes within this word
+      int unsigned nb = (exp_size - idx >= 4) ? 4 : (exp_size - idx);
+      bit [TL_DBW-1:0] mask = TL_DBW'((1 << nb) - 1);
```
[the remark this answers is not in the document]
alees24: Indeed (aside from the use of TL_DBW rather than TL_DW) and it has a certain elegance, although I've just pushed changes containing a restructuring of these buffer-related tasks because sometimes we shall need to separate the reading from the checking.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -98,6 +98,25 @@ virtual task dut_init(string reset_kind = "HARD");
...
+    // Wait until the given state is attained or the (optionally
+    usbdev_link_state_e rd;
+    do begin
+      uvm_reg_data_t data;
+      // Delay and if timeout was requested, decrease the cycle count.
+      cfg.clk_rst_vif.wait_clks(128);
+      if (timeout >= 0) timeout -= (timeout > 128) ? 128 : timeout;
+      csr_rd(.ptr(ral.usbstat.link_state), .value(data));
+      rd = usbdev_link_state_e'(rd);
+    end while (!(rd inside {states}) && (timeout != 0));
```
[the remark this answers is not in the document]
rswarbrick: I guess the behaviour would stay the same if the "if ... break" stuff went to the bottom of the loop. But that would be really ugly! I'm convinced that your code is better :-)

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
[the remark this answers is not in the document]
rswarbrick: Oh! How bizarre! I didn't realise that the `{}` bit was actually part of the `inside` syntax (looking at the 1800 doc). What an ugly language! :-D

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+          end
+        end
+      join
+    end
+  endtask
+
+  // TODO: Presently the act of sending a data packet, destructively modifies it!
+  // Restore the data packet to its original state. This just bit-reverses each byte
+  // within the input array.
+  function void recover_orig_data(input byte unsigned in[], output byte unsigned out[]);
```
rswarbrick, line 179: Can the output be a return value? It might make the call site a bit easier to understand.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+    // Check the contents of the packet buffer memory against the OUT packet that was sent.
+    recover_orig_data(m_data_pkt.data, pkt_data);
+    check_rx_packet(endp, 1'b0, usb_buf, pkt_data);
+
+    // This is now the expected content of the packet buffer
+    valid_len[usb_buf] = pkt_data.size();
+    exp_data[usb_buf]  = pkt_data;
+  endtask
+
+  // IN (sending) component of the USB side.
```
rswarbrick, line 76: Hmm, I'm not sure I understand the comments here :-) You wrote "OUT (sending) ..." above the previous task. Can you expand things a bit, to something like "IN component of the USB side, which is receiving data from the device"? (And similarly for the previous task)

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+    // Given that the packet has been accepted, we must flip the OUT Data Toggle.
+    exp_out_toggle[endp] ^= 1;
+
+    // Check the contents of the packet buffer memory against the OUT packet that was sent.
+    recover_orig_data(m_data_pkt.data, pkt_data);
+    check_rx_packet(endp, 1'b0, usb_buf, pkt_data);
+
+    // This is now the expected content of the packet buffer
+    valid_len[usb_buf] = pkt_data.size();
+    exp_data[usb_buf]  = pkt_data;
```
rswarbrick, line 73: Does it make sense to allow things like "Write 100 bytes to buf 0; Write 10 bytes to buf 0; Read 100 bytes from buf 0" and have an opinion about the upper 90 bytes?

If so maybe we should update `exp_data` in place and take a maximum when computing `valid_len`?

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+    csr_update(ral.configin[endp]);
+
+    // Token pkt followed by handshake pkt
+    call_token_seq(PidTypeInToken);
+    get_response(m_response_item);
+    $cast(response, m_response_item);
+    `DV_CHECK_EQ(response.m_pkt_type, PktTypeData);
+    $cast(in_data, response);
+    check_tx_packet(in_data, exp_in_toggle[endp] ? PidTypeData1 : PidTypeData0, pkt_data);
+    cfg.clk_rst_vif.wait_clks(20);
```
rswarbrick, line 106: Again, should this be `inter_packet_delay()`?

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+      int unsigned csr_buf;
+      int unsigned usb_buf;
+      // Choose a random buffer and direction for the CSR side access.
+      `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(csr_buf, csr_buf inside {[0:NumBuffers-1]};)
+      // Choose a random packet buffer and direction for the USB side, ensuring that it does not
+      // access the same buffer as the CSR side.
+      `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(usb_buf,
+                                         usb_buf inside {[0:NumBuffers-1]}; usb_buf != csr_buf;)
+      fork
+        begin : csr_side
```
rswarbrick, line 137: It's a bit of a nit, but would this be easier to read if the contents of `csr_side` and `usb_side` were pulled into named tasks? It would make the fork/join structure trivial to read, which would be nice.

rswarbrick, line 137: Would it make sense to add a small random delay at the start of each process here? That way, we'd see the situation where one side started when the other side was in the middle of a transaction.

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
```diff
@@ -168,12 +168,24 @@ class usbdev_scoreboard extends cip_base_scoreboard #(
...
+    // Do nothing
+  endtask
+  virtual task process_mem_read(tl_seq_item item, string ral_name);
+    // Do nothing
+  endtask
+
   virtual task process_tl_access(tl_seq_item item, tl_channels_e channel, string ral_name);
     uvm_reg csr;
     bit     do_read_check   = 1'b1;
     bit     buffer_mem      = 1'b0;
```
rswarbrick, line 188: Would you mind adding a comment above this line to explain what the `buffer_mem` flag means? I think I understood eventually, but it's not super obvious.

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
```diff
@@ -472,15 +484,16 @@ class usbdev_scoreboard extends cip_base_scoreboard #(
...
     if (!write && channel == DataChannel) begin
-      if (do_read_check & ~buffer_mem) begin
-        `DV_CHECK_EQ(csr.get_mirrored_value(), item.d_data,
-                     $sformatf("reg name: %0s", csr.get_full_name()))
-      end
-      else if (do_read_check & buffer_mem) begin
-        // do-nothing
+      if (do_read_check) begin
+        if (buffer_mem) begin
+          // do-nothing
```
rswarbrick, line 489: I know this code was here already, but would you mind expanding the comment slightly? I think we probably want something like "Do nothing: we do not model packet buffer memory in the scoreboard."

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+    // usb20_driver module.
+    wait_for_link_state({LinkActive, LinkActiveNoSOF}, 10 * 1000 * 48);  // 10ms timeout, at 48MHz
+
+    for (int unsigned txn = 0; txn < num_trans; txn++) begin
+      int unsigned csr_buf;
+      int unsigned usb_buf;
+      // Choose a random buffer and direction for the CSR side access.
+      `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(csr_buf, csr_buf inside {[0:NumBuffers-1]};)
+      // Choose a random packet buffer and direction for the USB side, ensuring that it does not
+      // access the same buffer as the CSR side.
```
rswarbrick, line 133: Thinking again about this, do we want to cover a situation where USB and CSR both read from a buffer at the same time? I think that would be a reasonable thing to do. Maybe we have to pick buffer numbers and also directions here?

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+    usb20_item response;
+
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(usb_len, usb_len inside {[0:MaxPktSizeByte]};)
+    `uvm_info(`gfn, $sformatf("USB sending an OUT packet of 0x%x byte(s) to buffer 0x%x",
+                              usb_len, usb_buf), UVM_MEDIUM)
+    // Supply the buffer
+    csr_wr(.ptr(ral.avoutbuffer), .value(usb_buf));
+
+    call_token_seq(PidTypeOutToken);
+    cfg.clk_rst_vif.wait_clks(10);
```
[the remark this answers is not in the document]
alees24: Yes, it hadn't been merged at that point.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,184 @@
...
+      join
+    end
+  endtask
+
+  // TODO: Presently the act of sending a data packet, destructively modifies it!
+  // Restore the data packet to its original state. This just bit-reverses each byte
+  // within the input array.
+  function void recover_orig_data(input byte unsigned in[], output byte unsigned out[]);
+    out = {<<8{in}};  // Reverse the order of the bytes
+    out = {<<{out}};  // Bit-reverse everything
```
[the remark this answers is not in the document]
alees24: I'm hopeful that we may very soon drop all of those contortions, so for now let's please leave them unchanged so they're at least consistent. It's in a few places already, sadly.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
alees24: Done.

alees24: I'm relying upon the CSR timing variation and any randomization on the usb20_driver side for now. There probably is none of the latter, but introducing it there would improve all sequences.
It is something that I'm aware could definitely be useful in all sequences as the DV structure is developed.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
alees24: It's not really because it assumes the physical properties of the packet buffer implementation. There's no reason that the set of OUT side (host->device) buffers could not be disjoint from the set of IN side (device->host) buffers in the memory implementation.

Presently it just happens to be implemented as a single SRAM, but the only way that the USB should be reading from a buffer is if the CSR side has previously written data into it, and similarly the CSR side should only be reading from a buffer that has recently been populated by USB writing. In effect each of the packet buffers is actually used as a FIFO.

(At some point the USB device may even act as a host on the TileLink bus and thus shuffle packet data between main memory and any small internal storage on demand, for higher performance and lower CPU involvement.)

alees24:
Thanks the review, and sharing the usbdev burden! I think I've addressed everything.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,190 @@
...
+
+// USBDEV packet buffer test sequence
+//
+// Employing randomization, sufficient transactions and coverage:
+//
+// - read and write the entire packet buffer memory from the CSR side
+// - read and write from/to every buffer on the USB side
+// - produce temporal collisions between the CSR side and the USB side;
+//   - The DUT does not need to handle simultaneous access to the same packet
+//     buffer from both sides because the hardware is responsible for ensuring
```
rswarbrick, line 13: Sorry to be annoying, but isn't the responsibility on the SW? :-)

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
```diff
@@ -0,0 +1,190 @@
...
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(usb_len, usb_len inside {[0:MaxPktSizeByte]};)
+    `uvm_info(`gfn, $sformatf("USB sending an OUT packet of 0x%x byte(s) to buffer 0x%x",
+                              usb_len, usb_buf), UVM_MEDIUM)
+    // Supply the buffer
+    csr_wr(.ptr(ral.avoutbuffer), .value(usb_buf));
+
+    call_token_seq(PidTypeOutToken);
+    inter_packet_delay();
+    call_data_seq(exp_out_toggle[endp] ? PidTypeData1 : PidTypeData0, 0, usb_len);
+    cfg.clk_rst_vif.wait_clks(20);
```
rswarbrick, line 56: What do these 20 clocks measure? I think the previous delay was 10 cycles (and 10 != 20!): is this a different sort of wait? Would you mind adding a comment that explains what the number means?

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
rswarbrick: Does the "let's not mess around with this: it's going away soon" logic apply here too?

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
rswarbrick: That makes sense. Thanks for thinking about it.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
rswarbrick: Ah, ok. Thanks for explaining.

rswarbrick (commented):
Thanks for the tidy ups. I think there are a few extra questions in the middle of the previous review (which GitHub helpfully hides!)

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
alees24: Fair enough.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
alees24: That would be another case of assuming that the packet buffer behaves like a regular memory, rather than a means of conveying packet data in a single direction.

The USB device would be quite within its rights to do anything with the remainder of the buffer contents really, and in fact it does write some additional data into the buffer, although software should not rely upon it. (It happens to write in the CRC16 bytes if they fit, and since it can only write complete 32-bit words it will corrupt 0-3 bytes of the final word, but that's implementation detail/artefact).

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
alees24: I'm afraid at the moment it's there through pragmatic mimicry. As per the 'inter_packet_delay' tasks, we need some kind of 'host_response_delay' task to collect (randomized) delays in the host responding to the packets it has received.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
rswarbrick: Cool, thanks for explaining.

hw/ip/usbdev/dv/env/seq_lib/usbdev_pkt_buffer_vseq.sv
rswarbrick: Cool, that makes sense. One day, I'll learn enough about the USB protocol to figure this out on my own :-)

rswarbrick (approved):
Looks great! Thank you for responding to my (many!) comments.

Issue #22089, opened 2024-03-18
[flash_ctrl] Patch up corner cases in DV after merging the scrambling module


msfschaffner:
I still seem to be running into issues with error case handling. 
For some reason the shared module seems to be causing issues with
```
flash_ctrl_oversize_error
flash_ctrl_ro_derr
flash_ctrl_rw_derr
flash_ctrl_integrity
flash_ctrl_rw_serr
```

I am a bit hesitant merging this without fully understanding these errors, and it is a critical block for the system.

_Originally posted by @msfschaffner in https://github.com/lowRISC/opentitan/issues/22045#issuecomment-2000916780_

msfschaffner:
This should probably be looked at in order to close down #21012

matutem:
I see integrity errors flagged by flash_ctrl_oversize_error with seed 101526864083573723368910117507542387818288841624698717015324241278126026539731.

The flash_phy_rd block generates integrity errors around time 51787ps in https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/flash_ctrl/rtl/flash_phy_rd.sv;drc=8046c2896fa50aaf3a186a7ce8c0570db9f99eaf;l=677, but I see the top 4 bits in data_out_muxed don't change with the data contents. There is something weird in https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/flash_ctrl/rtl/flash_phy_rd.sv;drc=8046c2896fa50aaf3a186a7ce8c0570db9f99eaf;l=586: forward == 0 and hint_descram == 1, the integrity bits are from fifo_data, which is constant, while the actual data is from descramble_data_i and mask, which change.

msfschaffner:
There were two exclusive instances of the scrambling module - one per bank.
We changed that and shared one scrambling module among both banks.
Specifically, there is a cipher datapath and a galois multiplier inside the scrambling module. Both operate independently, hence we used two RR arbiters that independently arbitrate the bank requests for these two sub-datapaths.

matutem:
I think the problem is that the scrambler is missing creating a calc response for bank 0, so the key sequence gets messed up. The waves show both banks requesting masks at 51267, and there is no ack for bank 0. Once the mask is wrong everything fails.

![Screenshot from 2024-03-21 11-43-28](https://github.com/lowRISC/opentitan/assets/80294896/db1f7604-7501-4953-9d26-2384465d02f1)

matutem:
[...]

PR #22191, opened 2024-03-22
[flash_ctrl] Revert back to using per-bank, dedicated scrambling modules

vogelpi:
An inspection of the DV issues arising from the change to share a single scrambling module between Flash banks revealed that there are indeed some corner cases that are not properly handled by the current design. Fixing those corner cases is tricky and may lead to further issues.

For this reason, this commit reverts the following commits for the time being:

- [doc,flash_ctrl] Fix inconsistency and typo in documentation 3a35e26c7a1e6e26aa38f4ae899670274f69b1d1

- [flash_ctrl] Use redundant arbiters in scrambling module a63c2d9a7abeef6411b9746f7540f444217aefa4

- [flash_ctrl] Share flash_scrambling module among banks c95dc32e6235cd0dd3beaee313dc54e25cebd8b4

This is related to lowRISC/OpenTitan#22089.


vogelpi:
Closing as we managed to fix the corner cases (an actual RTL bug).

PR #22192, opened 2024-03-22
[usbdev,dv] Remove repeated entry in usbdev_sim_cfg

rswarbrick:


alees24 (approved):
Thanks. I'm sure I've deleted that locally somewhere too!

PR #22193, opened 2024-03-22
[usbdev,dv] Simplify usbdev_enable_vseq

rswarbrick:
This sequence carefully disables the automatic usbctrl.enable write, and then does exactly the same thing by hand! Remove the configuration that disables the base class and use that instead. The results should be the same.


rswarbrick:
Hmm. In hindsight, maybe this is a bit silly. Long term, we're going to have to tidy it up a bit more drastically than this would do. Let's step back and think a bit :-)

PR #22195, opened 2024-03-22
[usbdev,dv] Move covergroup descriptions to testplan

### LINKED ISSUE #22194: [usbdev,dv] Describe covergroups in the testplan
### Description

There are some covergroups defined in `usbdev_env_cov.sv`, along with comments explaining what they do. This is very sensible! But the comments (and list of covergroups) should probably be in the testplan instead. Move them accordingly.

rswarbrick:
This is a pretty mechanical change: it is basically copying the list of covergroups from usbdev_env_cov.sv into usbdev_testplan.hjson. I've taken the long comment above each covergroup and turned it into the description of the covergroup in the testplan.

The only nontrivial change is to a covergroup name (pid_type_end_cg), which had a stray "f".

@alees24: Would you mind glancing over this? I *think* it should be equivalent to the text you wrote, but want to make sure I'm not mangling anything.

Closes #22194.

### DIFF
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -1113,4 +1113,82 @@
       tests: ["usbdev_fifo_rst"]
     }
   ]
+
+  covergroups: [
+    {
+      name: pids_cg
+      desc: '''
+        Observe all PIDs about which we care; a number of the available PIDs are ignored by the USB
+        device because they related only to High Speed operation and/or hubs.
+      '''
+    }
+    {
+      name: framenum_rx_cg
+      desc: '''
+        Observe all significant/extreme frame numbers being reported via CSRs; this just ensures
+        that all bits have been decoded and presented; the frame number has no significance to
+        USBDEV but it's important to software.
+      '''
+    }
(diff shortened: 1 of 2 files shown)


hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -1113,4 +1113,82 @@
...
+    {
+      name: ep_in_cfg_cg
+      desc: '''
+        Receipt of IN and PRE packets with different IN endpoint configurations
+      '''
+    }
+    {
+      name: fifo_lvl_cg
+      desc: '''
+        Received SETUP/OUT packets within FIFO levels of interest
```
alees24, line 1168: That was me, but it's supposed to be simply 'with FIFO...'

alees24 (approved):
The formatting of the multi-line comments is unconventional, but not enforced as I understand it. Otherwise, it looks fine thanks.

hw/ip/usbdev/data/usbdev_testplan.hjson
rswarbrick: Cool, I'll fix that now, thanks.

PR #22197, opened 2024-03-22
[ot_certs] Refactor extension handling

### LINKED ISSUE #22062: [ot_certs] update `ot_certs` lib to enable creating new custom cert extensions
Currently, the certificate [Rust struct representation](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/ot_certs/src/template/mod.rs;drc=47fe7f75a75dec7f7d2e57830aad2db898526e3d;l=57) is tightly coupled to the DiceTcbInfo custom extension. We would like to decouple this to enable creating different certificates with different custom extensions. Specifically we would like to eventually create a TPM certificate (with the appropriate custom extension. We should:

- [ ] Refactor the `Certificate` struct and separate out a `CustomExtension` struct to represent various custom extensions.
- [ ] Refactor the certificate codegen to use the above new struct layouts
- [ ] Refactor the HJSON certificate descriptions for the DICE certs (UDS, CDI_0, CDI_1)  to use above struct representation.

pamaury:
This PR refactor ot_certs extension handling to make it more generic. This will pave to the way to add support for more extensions like TPM. This PR also fixes a couple of issues found when refactoring and also adds an integration tests so that we parse certificate with openssl and check that openssl interprets the certificates as expected.

Fixes #22062


sw/host/ot_certs/src/template/mod.rs
```diff
@@ -67,22 +69,37 @@ pub struct Certificate {
...
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum CertificateExtension {
+    /// DICE TCB extension.
+    DiceTcb(DiceExtension),
+}
+
+/// DICE TCB extension.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct DiceExtension {
```
timothytrippel, line 88: Why define a new struct here? can you not use the `DiceTcbInfo` struct in `dice_tcb.rs`?

sw/host/ot_certs/src/template/mod.rs
```diff
@@ -67,22 +69,37 @@ pub struct Certificate {
...
+    pub extensions: Vec<CertificateExtension>,
+    /// X509 certificate's signature.
+    pub signature: Signature,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum CertificateExtension {
+    /// DICE TCB extension.
+    DiceTcb(DiceExtension),
```
timothytrippel, line 82:
```suggestion
    DiceTcbInfo(DiceExtension),
```

or 

```suggestion
    DiceTcbInfoExt(DiceExtension),
```
?

sw/host/ot_certs/src/x509/extension.rs
```diff
@@ -113,13 +199,25 @@ impl<'a> asn1::SimpleAsn1Readable<'a> for Flags {
     }
 }
 
-const DICE_TCB_EXT_OID: &str = "2.23.133.5.4.1";
+pub const DICE_TCB_EXT_OID: &str = "2.23.133.5.4.1";
```
timothytrippel, line 202:
```suggestion
pub const DICE_TCB_INFO_EXT_OID: &str = "2.23.133.5.4.1";
```

sw/host/ot_certs/src/x509/extension.rs
```diff
@@ -94,6 +134,52 @@ pub struct DiceTcbInfo<'a> {
...
 
+fn asn1utf8_to_str(s: &asn1::Utf8String) -> Value<String> {
+    Value::literal(s.as_str().to_string())
+}
+
+fn asn1bigint_to_bn(bn: &asn1::BigInt) -> Value<BigUint> {
+    Value::literal(BigUint::from_bytes_be(bn.as_bytes()))
+}
+
+impl DiceTcbInfo<'_> {
```
timothytrippel, line 145: Out of curiosity, why are there two DiceTcbInfo structs (one in `extension.rs` and one in `dice_tcb.rs` and an additional `DiceExtension` struct in `template/mod.rs`? Can we not collapse these into one struct representation within this lib and impl the appropriate methods on it to instantiate it from asn1 and conversely gen asn1 from it?

sw/host/ot_certs/tests/generic.hjson
```diff
@@ -0,0 +1,135 @@
+// Copyright lowRISC contributors.
```
timothytrippel, line 1: do you need another "generic.hjson" here for the test? can you reuse the one in sw/device/silicon_creator/lib/cert/generic.hjson ?

sw/device/silicon_creator/lib/cert/cdi_0.hjson
```diff
@@ -68,19 +68,24 @@
...
-        ],
-        flags: {
-            not_configured: false,
-            not_secure: false,
-            recovery: false,
-            debug: false,
-        }
+        extensions: [
+            {
+                type: "dice_tcb",
```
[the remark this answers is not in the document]
pamaury: Sure, I'll change it

sw/host/ot_certs/src/template/mod.rs
pamaury: I haven't finished refactoring yet. It cannot be used as-is for now because the structure in `dice_tcb.rs` only contains reference and does not own the data. Also the one in `dice_tcb.rs` is more general, as it contains a few fields that we do not use so I am undecided if I should remove them entirely or not? But eventually those two structure will be merge, probably the one in `dice_tcb` will disappear.

sw/host/ot_certs/src/x509/extension.rs
pamaury: The one in `extension.rs` cannot be removed because it is the one used to derive the parser so it contains stuff like tag and asn1 specific marking. It really ought to be private really (might change that in that PR). The other two will be merged.

sw/host/ot_certs/tests/generic.hjson
pamaury: The generic.hjson in `sw/device` will be remove to point to the one in `ot_certs`. I am also thinking about moving the example hsjon as well.

sw/host/ot_certs/src/template/mod.rs
timothytrippel: sg. I think consolidating at least these two should make things more maintainable.

sw/host/ot_certs/src/x509/extension.rs
```diff
@@ -64,14 +106,20 @@ pub fn x509_get_ext_by_obj<'a>(x509: &'a X509, obj: &Asn1Object) -> Result<&'a A
...
-pub struct Fwid<'a> {
+struct Fwid<'a> {
     pub hash_alg: asn1::ObjectIdentifier,
     pub digest: &'a [u8],
 }
 
+// This is an internal structure used to parsed an extension using the ASN1
+// extension. We cannot use the `DiceTcbInfoExtension` in `template` since we
+// need to use specific annotations and types so that the `asn` library can
+// derive a parser.
```
timothytrippel, line 120: nit:
```suggestion
// This is an internal structure used to parse a DiceTcbInfo extension using the ASN1
// crate. We cannot use the `DiceTcbInfoExtension` in `template` since we
// need to use specific annotations and types so that the `asn` library can
// derive an ASN1 parser.
```

timothytrippel (approved):
Thanks for the update @pamaury ! Looks great.

PR #22201, opened 2024-03-22
[dv,mem_bkdr] Fix digest calculation for hw_cfg0

### LINKED ISSUE #22200: [dv,mem_bkdr_util] Fix digest calculation for hw_cfg0
### Description

When splitting the hw_cfg into hw_cfg0 and hw_cfg1 the digest calculation for hw_cfg0 erroneously added an extra 32 bits set to 0. This is only needed for the hw_cfg1 partition, since the contents of hw_cfg0 are 64-bit aligned, and hw_cfg1 is not.

matutem:
This bug was introduced when splitting the hw_cfg partition.

Fixes #22200


sha-ron (approved):

sha-ron:
CC @dkabely-ot

eshapira (approved):

PR #22202, opened 2024-03-22
[spi_host] Correct TXFULL description

kupiakos:
Cc @timothytrippel

### DIFF
--- a/hw/ip/spi_host/data/spi_host.hjson
+++ b/hw/ip/spi_host/data/spi_host.hjson
@@ -230,7 +230,7 @@
         { bits: "29",
           name: "TXFULL",
           desc: '''When high, indicates that the transmit data fifo is full.
-                   Any further writes to !!RXDATA will create an error interrupt.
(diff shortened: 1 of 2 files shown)


kupiakos:
Looks like I need to change the HJSON register definition? Not sure where to start here

rswarbrick (commented):
This is rather confusing to fix because we check in some generated code (grr!). The source for this line is actually `hw/ip/spi_host/data/spi_host.hjson` and we have a Makefile that regenerates the Markdown docs to match. The CI check is failing because the Markdown docs no longer match :-)

I'll push a fixed version in a second (getting this right is a real pain, especially because the failing part of CI in a bit that isn't visible to external contributors yet). I'll ask someone else who knows more about `spi_host` to review it technically.

hcallahan-lowrisc (approved):
Thanks for the fix @kupiakos :)

hcallahan-lowrisc:
CW310 Test Failures in CI are known and unrelated to this documentation-only change. Hence merging.

PR #22205, opened 2024-03-22
[spi_host] Fix "bute"/"byte" typo in register docs

kupiakos:


rswarbrick (approved):
Thanks for fixing this! As with the other spi_host doc fix that you sent, the problem is that the real source for this text is an hjson file (`hw/ip/spi_host/data/spi_host.hjson`). You can get everything lined up by editing that file and then running `./util/cmdgen.py -u '**/*.md'`.

I've just done it for this PR and force-pushed. I'll try to come back and look again (and merge the result) once CI has finished.

Thank you very much for the work you are doing to tidy up the documentation here.

PR #22213, opened 2024-03-23
[prim_fifo_sync] Move address and depth calculation to `prim_fifo_sync_cnt`

andreaskurth:
This will make it easier to reuse `prim_fifo_sync_cnt` to build other synchronous FIFO modules (e.g., one that doesn't use FFs as storage). The functionality of `prim_fifo_sync` is not changed by this PR. This PR also improves the documentation/comments around how FIFO pointers are calculated.

### DIFF
--- a/hw/ip/prim/rtl/prim_fifo_sync_cnt.sv
+++ b/hw/ip/prim/rtl/prim_fifo_sync_cnt.sv
@@ -12,7 +12,9 @@ module prim_fifo_sync_cnt #(
   // Width of the read and write pointers for the FIFO
   parameter int Width = 16,
   // Whether to instantiate hardened counters
-  parameter bit Secure = 1'b0
+  parameter bit Secure = 1'b0,
+  // Width of the 'current depth' output
+  localparam int unsigned DepthW = prim_util_pkg::vbits(Depth+1)
 ) (
   input clk_i,
   input rst_ni,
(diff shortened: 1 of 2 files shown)


andreaskurth:
I ran block-level regressions for i2c, which instantiates four `prim_fifo_sync`, and did not notice any problems. I'm currently running all top-level tests (one seed per test) locally to double-check that this doesn't break anything in other modules.

andreaskurth:
CI checks passed except for _kokoro_ (environment error) and _CW310 ROM Tests_, which is currently known to be flaky.

vogelpi (approved):

hw/ip/prim/rtl/prim_fifo_sync_cnt.sv
```diff
@@ -8,7 +8,7 @@
 
 module prim_fifo_sync_cnt #(
   // Depth of the FIFO, i.e., maximum number of entries the FIFO can contain
-  parameter int Depth = 4,
+  parameter int unsigned Depth = 4,
```
vogelpi, line 11: Nit: I think you could remove some signed to unsigned conversions in this file now that were needed before.

vogelpi (commented):
Thanks for the PR @andreaskurth , this looks definitely good!

I've thoroughly reviewed the code and I can confirm that this PR doesn't change functional behavior of the FIFO primitive.

Thanks for your comments around the code manipulating the pointers inside `prim_fifo_sync_cnt.sv` by the way. They are very useful and made the review a lot easier.

hw/ip/prim/rtl/prim_fifo_sync_cnt.sv
andreaskurth: Thx, done in PR #22221

Issue #22025, opened 2024-03-14
[top_earlgrey] Area optimizations for production


msfschaffner:
### Description

This is a tracker for area optimizations that are in flight for Earlgrey PROD.

Planned for RC1:
- [x] *kmac*: Switch PRNG from Trivium to Bivium #22021 
- [x] *entropy_src*: reduce depth of esfinal FIFO #22041
- [x] *uart*: reduce TX buffer size from 128B to 32B #22026
- [x] *uart*: reduce RX buffer size from 128B to 64B #22046
- [x] *rv_core_ibex*: remove 8 performance counters #22047
- [x] *csrng*: more fine-grained parametrization of AES to remove unneeded datapaths #22056
- [x] *csrng*: remove redundant state copies in state database #22064

Planned for RC2:
- [x] *flash_ctrl*: Share the scrambling module between two banks #22045
- [x] *entropy_src*: reduce depth of observe FIFO #22087 

Planned for RC3:6000
- [x] #22159

Deferred:
~- [ ] Reverting of #22045 due to problems found in #22089; deferred because likely not a bug, see #22191~ It turned out to be an RTL bug but this got fixed by #22276 

The list only contains optimizations that have been approved. More ideas are captured in [this document](https://docs.google.com/spreadsheets/d/1uZ79DLd08nX7gYJlmsNF_QhiHITyCkgwTxe4lKnuGsg/edit).

msfschaffner:
FYI @zi-v @arnonsha @msfschaffner @OTshimeon 
This is where we tracked approved optimizations.

andreaskurth:
Closing as completed (at least for now).

PR #22216, opened 2024-03-23
[i2c/rtl] Replace FFs for FIFOs with a single SRAM

andreaskurth:
This PR replaces most FFs used to store FIFO data in i2c with one single-port SRAM that is 13 bit wide and 472 entries deep. This optimization is part of the area saving efforts for Earlgrey-PROD (https://github.com/lowRISC/opentitan/issues/22025).


andreaskurth:
I tested this locally by running all i2c block-level tests with `-rx 0.1`. I noticed that:
- The `i2c_csr_bit_bash` test fails more often because it doesn't take the `wready` of FIFOs into account, so assertions in `i2c_fifos` fail. I propose fixing this for M4 in DV by disabling the assertions for that test.
- The `i2c_target_stress_wr` test fails more often because the FSM writes the ACQ FIFO even though it is full, so an assertion in `i2c_fifos` fails. I propose fixing this for M4 in RTL with a small FSM change.

Otherwise, the pass rate for all tests seems to be as can be expected based on [current nightly results](https://reports.opentitan.org/hw/ip/i2c/dv/latest/report.html).

I'm now running a full block-level regression and will post the results here later.

andreaskurth:
All CI checks except _kokoro_, which just flagged an inconsequential lint error (unused `localparam`), and _CW310 Test ROM Tests_, which is currently known to be flaky, passed. I'll fix the unused `localparam`.

image

vogelpi (approved):

andreaskurth:
_kokoro_ finally passed, and all other CI checks (except _CW310 Test ROM tests_ passed before. This is urgent, merging

PR #22220, opened 2024-03-25
[uart,dv] Import generic testplan for stress_all_with_reset

rswarbrick:
This mirrors most of the other blocks in the design. One consequence is that the testpoint moves to V3, but we probably believe that's a better place for it to live anyway.


andreaskurth (approved):
Agreed, thx @rswarbrick :+1:

PR #22226, opened 2024-03-25
[reggen] Reserve CIP IDs for Integrated IPs

andreaskurth:
Four HW IP modules have been created on the `integrated_dev` branch, which hasn't been reconverged with `master` yet.  This commit reserves their Comportable IP IDs on `master` to ensure we don't accidentially assign their ID to another new IP, which would make reconvergence more difficult.


vogelpi (approved):
LGTM, thanks @andreaskurth !

andreaskurth:
Failure of _CW310 Test ROM Tests_ is unrelated; merging

PR #22229, opened 2024-03-25
[DO NOT MERGE][ot_certs] Add fake TPM extension

pamaury:
This PR shows how to add an X509 extension support to `ot_certs`. Only the last commit matters, the rest is from #22197.


timothytrippel:
@vbendeb : if you take a look at just the last commit in this PR, @pamaury has added some stubs to enable you to easily add the TPM cert extension you need to add to the existing OT certificate generation infrastructure we have. He is going to add some more documentation as well that describes the process of adding more custom extensions. Once this merges, you should be able to update the stubs to add the entire TPM cert definition.

timothytrippel (approved):
TPM stub LGTM, thanks @pamaury !

PR #22231, opened 2024-03-25
[ci] Print SAM3X firmware version for FPGA jobs

jwnrt:
This PR prints the SAM3X firmware version before running FPGA tests. This is useful even for hyperdebug tests as we still program the bitstream etc through the SAM3X.

### DIFF
--- a/ci/scripts/run-fpga-tests.sh
+++ b/ci/scripts/run-fpga-tests.sh
@@ -44,6 +44,9 @@ trap 'ci/bazelisk.sh run //sw/host/opentitantool -- --rcfile= --interface=${fpga
 ci/bazelisk.sh run //sw/host/opentitantool -- --rcfile= --interface="$fpga" --logging debug fpga set-pll || true
 ci/bazelisk.sh run //sw/host/opentitantool -- --rcfile= --interface="$fpga" fpga clear-bitstream
 
+# Print the SAM3X firmware version. HyperDebug transports don't currently support this, so we ignore errors.
+ci/bazelisk.sh run //sw/host/opentitantool -- --rcfile= --interface="$fpga" fpga get-sam3x-fw-version || true
+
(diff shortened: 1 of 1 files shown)


pamaury (approved):
Good idea, thanks James

ci/scripts/run-fpga-tests.sh
```diff
@@ -44,6 +44,9 @@ trap 'ci/bazelisk.sh run //sw/host/opentitantool -- --rcfile= --interface=${fpga
 ci/bazelisk.sh run //sw/host/opentitantool -- --rcfile= --interface="$fpga" --logging debug fpga set-pll || true
 ci/bazelisk.sh run //sw/host/opentitantool -- --rcfile= --interface="$fpga" fpga clear-bitstream
 
+# Print the SAM3X firmware version.
+ci/bazlisk.sh run //sw/host/opentitantool -- --rcfile= --interface="$fpga" fpga get-sam3x-fw-version
+
```
luismarques, line 49: `bazelisk` is misspelled. Otherwise looks good to me. You could also echo something before printing the version, if the output itself isn't sufficiently self-descriptive.

jwnrt: Oops, impressive to typo a one line change. Maybe I should have found a way to run this locally. Thanks.

Bazel prints the command it's running so I think it's fairly self descriptive.

luismarques (approved):
Good enough:

```
(09:51:28) INFO: Running command line: bazel-bin/sw/host/opentitantool/opentitantool '--rcfile=' '--interface=cw310' fpga get-sam3x-fw-version
[
  1,
  2,
  0
]
```

pamaury:
You might want to just add a `|| true` after the command so that transports that don't support it don't break the test.

PR #22235, opened 2024-03-25
[dvsim] Be more explicit about seeds when running tests

rswarbrick:
This should avoid ugly paths all over the place and tests that are hard to reproduce. We still get the long build seeds that were the motivation for the 256 bit constants that we had before.

The base seeds are now all 16 bit values, so should be a bit more convenient to type on the command line. But the build seeds should be much longer (256 bits) if needed.

### DIFF
--- a/hw/dv/tools/dvsim/common_modes.hjson
+++ b/hw/dv/tools/dvsim/common_modes.hjson
@@ -69,7 +69,7 @@
     {
       name: build_seed
       is_sim_mode: 1
-      build_opts: ["+define+BUILD_SEED={seed}"]
+      build_opts: ["+define+BUILD_SEED={long_seed}"]
     }
   ]
 
--- a/util/dvsim/Deploy.py
+++ b/util/dvsim/Deploy.py
@@ -472,9 +472,10 @@ class RunTest(Deploy):
         self.test_obj = test
         self.index = index
         self.build_seed = sim_cfg.build_seed
-        self.seed = RunTest.get_seed()
-        # Systemverilog accepts seeds with a maximum size of 32 bits.
-        self.svseed = int(self.seed) & 0xFFFFFFFF
+
+        self.base_seed, self.short_seed, self.long_seed = \
+            RunTest.get_test_seed()
+
         self.simulated_time = JobTime()
         super().__init__(sim_cfg)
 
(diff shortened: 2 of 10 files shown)


vogelpi (commented):
Thanks for the PR @rswarbrick . The change looks good to me but I don't think it's a good idea to merge this right now. There are too many things in flight and several people under lots of pressure heavily pushing to hit the next deadline.

If we decide we want to move forward with this, we will need to make sure all partners have visibility and are aligned, that nothing in open and closed source DV breaks because of this etc., track coverage and pass/fail rates. Right now is not the right time for this IMO.

rswarbrick:
Ok. Let's leave this to mature gently :-) We can always reconsider things and tidy this up after next time we take a release branch (where we'll be at a rather different place in the release cycle)

andreaskurth:
> Ok. Let's leave this to mature gently :-) We can always reconsider things and tidy this up after next time we take a release branch (where we'll be at a rather different place in the release cycle)

Agreed, thanks for your work on this, @rswarbrick. This will be a useful improvement but now doesn't seem to be the right time to make this change.

util/dvsim/Deploy.py
```diff
@@ -472,9 +472,12 @@ def __init__(self, index, test, build_job, sim_cfg):
         self.test_obj = test
         self.index = index
         self.build_seed = sim_cfg.build_seed
-        self.seed = RunTest.get_seed()
-        # Systemverilog accepts seeds with a maximum size of 32 bits.
-        self.svseed = int(self.seed) & 0xFFFFFFFF
+
+        base_seed, short_seed, long_seed = RunTest.get_test_seed()
```
matutem, line 476:
```suggestion
        self.base_seed, self.short_seed, self.long_seed = RunTest.get_test_seed()
```
And remove the 3 lines below.

util/dvsim/Deploy.py
```diff
@@ -561,17 +564,35 @@ def post_finish(self, status):
...
+        if RunTest.seeds:
+            base = RunTest.seeds.pop(0)
+        elif RunTest.fixed_seed is not None:
+            base = RunTest.fixed_seed
+        else:
             for i in range(1000):
-                seed = random.getrandbits(256)
-                RunTest.seeds.append(seed)
-        return RunTest.seeds.pop(0)
+                base = random.getrandbits(16)
```
matutem, line 586: I thought you intended to use 32 bits?

hw/dv/tools/dvsim/common_modes.hjson
```diff
@@ -52,19 +52,19 @@
     // in two ways. One of them is setting the pre-processor macro `BUILD_SEED` to the seed value,
     // which is done below. The SystemVerilog testbench sources can use the `BUILD_SEED` macro
     // value to set some design constants (such as parameters) upon instantiation. The `BUILD_SEED`,
-    // if not set externally (by passing the --build-seed switch) is set to 1 in 
-    // `hw/dv/sv/dv_utils/dv_macros.svh`. The other way is by passing the {seed} value to utility 
-    // scripts that generate packages that contain randomized constants. These utility scripts can 
-    // be invoked as a `pre_build_cmd`, wrapped within the `build_seed` sim mode in the DUT 
+    // if not set externally (by passing the --build-seed switch) is set to 1 in
+    // `hw/dv/sv/dv_utils/dv_macros.svh`. The other way is by passing the {seed} value to utility
```
matutem, line 56:
```suggestion
    // is set to 1 in `hw/dv/sv/dv_utils/dv_macros.svh` if not set externally by passing the --build-seed switch).
    //. The other way is by passing the {seed} value to utility
```

Also, remove the trailing `,' in the line above.

By the way, otp also uses a "build" image.

util/dvsim/Deploy.py
rswarbrick: It definitely mustn't be more than 32 bits, but I'm unconvinced making it that long is necessary (and the shorter options are easier to type / remember...). 65k options is still quite a lot! What do you think?

hw/dv/tools/dvsim/common_modes.hjson
rswarbrick: Amusingly, the change actually landed in April(!) with f4c2bb9 and the hunk will go away when I rebase in a sec.

Issue #19199, opened 2023-07-14
[rom] Reactivate ROM tests with real keys for PROD


msfschaffner:
### Description

Reactivate ROM tests with real keys for the next ROM release, see discussion here: https://github.com/lowRISC/opentitan/pull/17117#issuecomment-1635253025

msfschaffner:
@timothytrippel @cfrantz

moidx:
@cfrantz is going to look at enabling //sw/device/silicon_creator/rom/e2e/release:rom_e2e_self_hash_test.

timothytrippel:
Now that keys are in OTP, we should only need to reactive the ROM self hash test. The other presigned binary test serves no purpose anymore and will be removed.

Issue #17177, opened 2023-01-27
[topgen] Topgen fails if all source clocks are free running


AWHoppe:
### Description

I was trying no generate a new toplevel in which the source clocks would be external and free running, although derived clocks would still be managed. 
The topgen however cannot populate the parent_child_clks dict because of  lines 321 and 322 of util/topgen/clocks.py :

`if not clk.aon:`
`    parent_child_clks[name] = [name];` 

I'm not sure if the field 'aon' is being used correctly or if external free-running source clocks are not allowed in the design as-is since there is an overlap in the meaning of the hjson field 'aon' for free-running clocks and the defined low-power 'aon' always-on clock.

Would just like to understand the rationale.
Thanks

johngt:
Assigning @andreaskurth to have a first look at this and then either answer this or to (re-)assign accordingly

andreaskurth:
Thanks for your question, @AWHoppe.

In the current design, the [clock manager](https://docs.opentitan.org/hw/ip/clkmgr/doc/) manages the `main`, `io`, and `usb` clocks. In that sense, they are not free running. The `aon` field of the `clocks.srcs` data structure indicates this:
https://github.com/lowRISC/opentitan/blob/7bf43b1d56a998a465308eb683b7d8413bc5f1e6/hw/top_earlgrey/data/top_earlgrey.hjson#L42-L53
(The field name `aon` may not be the most descriptive, but the comment above defines its meaning.)

AWHoppe:
Hi @andreaskurth,

Thanks for the feedback. I'm aware of the HJSON field.
I'm trying to achieve a design where the sources besides 'aon' are also free-running.
I.e., the source clocks will be external to the device (waving any security considerations for the moment).
Basically all the source clocks have 'aon: "yes"' .
However these clocks still have derived clocks of their own.

On the topgen script, the lines I mention above do not let the script populate the dict parent_child_clks with free-running sources, so the scripts gives an errors when it tries to parse the derived clocks.

I'm not sure if the overlap in naming with aon was noticed here, since aon does not have derived clocks of its own, or if there's another rationale behind it.

andreaskurth:
Ah, I think I understand your point about the difference between 'aon' as the abbreviation for the always-on clock, which is linked to the power-on reset, and `aon` as the field in the above struct, which defines whether the on-chip clock manager manages the clock or not.

[...]

PR #22238, opened 2024-03-25
[rom_e2e,dv] disable ROM self hash test in DV

timothytrippel:
The ROM E2E self hash test was disabled on FPGA in #17177. The test should also have been disabled in DV. #19199 tracks reenabling the test when the ROM is released for the Earlgrey PROD backend integration.


pamaury (approved):
Looks good to me. Thanks Tim

timothytrippel:
Merging as failures have nothing to do with this PR.

PR #22239, opened 2024-03-25
[usbdev] Packet receiver refinements

alees24:
This PR proposes two alterations to the receiver side logic of the USB device:

1. As per the assertion in section `8.4.5 Handshake Packets` of the protocol specification, handshake packets that contain additional bits between the valid PID and the EOP shall cause the handshake packet to be ignored. Local modifications to the USBDPI model have demonstrated that an arbitrary number of bits may be introduced between the handshake PID and the EOP, and the handshake will still be accepted at present. Indeed the DPI model is erroneously doing this in a couple of places where it transmits 'USB_PID_ACK' as a complete token packet.

2. As per section `8.7.1 Packet Error Categories` a SOF packet that has a valid PID shall not be ignored completely; instead the presence of a CRC5 error and/or bit stuffing shall cause the Frame Number to be ignored, but the packet shall still be used for synchronization purposes. This may prove important in adjusting the USB clock frequency to match that of the USB Host.

In the first commit, it should also be noted that Data packets must be an integral number of bytes (`8.4.4 Data Packets`) and that there is presently no validity check on the packet length, including whether there is a sufficient number of bits to encode the CRC16.

Local changes to the USBDPI model to improve its error injection capabilities have shown - by waveform inspection and test failure - that additional bits may be inserted before the EOP for each of (i) token packets, (ii) data packets and (iii) handshake packets and the packet is presently still accepted. With the changes in this PR, such packets are now ignored as intended.

The change in 2 really requires additional DV effort, since block level DV presently does not emit SOF packets (excepting the nascent sequence in draft PR #22199).

Relates to #22240.

### DIFF
--- a/hw/ip/usbdev/rtl/usb_fs_nb_pe.sv
+++ b/hw/ip/usbdev/rtl/usb_fs_nb_pe.sv
@@ -88,7 +88,8 @@ module usb_fs_nb_pe #(
   input  logic  [NumInEps-1:0]   in_ep_iso_i, // Configure endpoint in isochronous mode
 
   // sof interface
-  output logic                   sof_valid_o,
+  output logic                   sof_detected_o, // SOF detected, PID valid
+  output logic                   sof_valid_o, // SOF packet valid; frame number may be trusted
   output logic [10:0]            frame_index_o,
 
   // RX line status
@@ -140,6 +141,7 @@ module usb_fs_nb_pe #(
   logic [10:0] rx_frame_num;
   logic rx_data_put;
   logic [7:0] rx_data;
+  logic rx_pid_valid;
   logic rx_pkt_valid;
 
   // tx mux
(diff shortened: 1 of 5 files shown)


hw/ip/usbdev/rtl/usb_fs_nb_pe.sv
```diff
@@ -159,6 +161,7 @@ module usb_fs_nb_pe #(
   logic usb_oe;
 
   // sof interface
+  assign sof_detected_o = rx_pkt_end & rx_pid_valid & (usb_pid_e'(rx_pid) == UsbPidSof);
```
a-will, line 164: I had to stare at that for awhile to see the difference with sof_valid_o's assignment, lol

a-will (approved):
Nice work!

hw/ip/usbdev/rtl/usb_fs_nb_pe.sv
alees24: Good point. I've added some explanatory comments.

PR #22245, opened 2024-03-26
[siva] Add cryptotest targets to chip_kmac_testplan.

moidx:
Adds cryptotest targets to KMAC stress chip test points.

## Sival Test Issues

* https://github.com/lowRISC/opentitan/issues/21406
* https://github.com/lowRISC/opentitan/issues/21407
* https://github.com/lowRISC/opentitan/issues/21408
* https://github.com/lowRISC/opentitan/issues/21409


moidx:
Hi @engdoreis, some of these tests take a long time to run. I assume that the nightly regression will only run them once and then the automation will take care of mapping the results to each chip level test point? Thanks

timothytrippel (approved):

PR #22248, opened 2024-03-26
[usbdev,env] DV env changes to avoid csr_test failures

MubashirSaleem775:
This draft pull request introduces changes to the ubsdev environment to facilitate the CSR tests failures, specifically the csr_bit_bash and mem_walk.

Changes include:

Addition of the driver_enable task in usb20_block_if to manage the driver by enabling and disabling it. The driver is now set to a disabled state by default, ensuring it remains inactive when not required, such as during CSR tests. To activate the driver when necessary, the driver_enable function can be called in a specific sequence, within the pre_task method, enabling the driver for USB transactions.


MubashirSaleem775:
@rswarbrick please review this draft PR

hw/dv/sv/usb20_agent/usb20_block_if.sv
```diff
@@ -28,7 +28,15 @@ interface usb20_block_if (
   logic usb_ref_val_o;
   logic usb_ref_pulse_o;
   logic usb_clk;             // signal used to divide clock or send J/K symbols for 4 clock cycles
+  logic drive_vbus; // to drive usb_vbus from driver
+  bit connected = 0;
```
rswarbrick, line 32: Nit: indentation

rswarbrick, line 32: Could you make the name a bit more specific here? I think you want it to be a direction signal that says we're treating `usb_vbus` as an output. Maybe something like `driving_vbus`?

hw/dv/sv/usb20_agent/usb20_block_if.sv
```diff
@@ -28,7 +28,15 @@ interface usb20_block_if (
...
 
+  // Enable/disable the output drivers
+  function automatic void enable_driver(bit enabled);
+    connected = enabled;
+  endfunction
+
+  assign usb_vbus = connected ? drive_vbus : 1'bZ;
```
rswarbrick, line 39: I think this makes sense, but don't we want the port type to be `inout`?

hw/dv/sv/usb20_agent/usb20_block_if.sv
```diff
@@ -28,7 +28,15 @@ interface usb20_block_if (
   logic usb_ref_val_o;
   logic usb_ref_pulse_o;
   logic usb_clk;             // signal used to divide clock or send J/K symbols for 4 clock cycles
+  logic drive_vbus; // to drive usb_vbus from driver
```
rswarbrick, line 31: Can you change this name? At the moment, it looks like it means "Please drive vbus", but that's not what you mean :-) Maybe just switching to `vbus_drive` would avoid the confusion?

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -278,14 +278,15 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
     cfg.bif.drive_n = 1'b0;
     @(posedge cfg.bif.rst_ni);
-    cfg.bif.usb_vbus = 1'b0;
+    cfg.bif.drive_vbus = 1'b0;
     repeat(usb_idle_clk_cycles) @(posedge cfg.bif.usb_clk);
-    cfg.bif.usb_vbus = 1'b1;
-    bus_reset();
+    cfg.bif.drive_vbus = 1'b1;
+    if (cfg.bif.connected) bus_reset(); // check that driver is enabled then starts bus_reset.
+     // this change is introduced to avoid csr test failures due to reactive driver.
```
rswarbrick, line 289: Nit: Probably best to move these comments to the line before. At the moment, it's not completely obvious that the second line is part of the same comment as the first line.

hw/dv/sv/usb20_agent/usb20_block_if.sv
```diff
@@ -28,7 +28,15 @@ interface usb20_block_if (
...
+  bit connected = 0;
 
+  // Enable/disable the output drivers
+  function automatic void enable_driver(bit enabled);
```
rswarbrick, line 35: Can you change this name slightly? At the moment, a call that looks like `enable_driver(0)` *looks* like it's enabling a driver, but that's not what it means. Maybe a name like `set_driver_enable` or something would make sense?

rswarbrick:
Looking at the failing CI, it *is* reporting a problem with this PR. Specifically, the DV smoke test is failing on usbdev items. Specifically, `usbdev_csr_hw_reset` and `usbdev_csr_rw` both fail.

hw/dv/sv/usb20_agent/usb20_block_if.sv
MubashirSaleem775: for usb_vbus it don't need to be inout.

MubashirSaleem775:
> Looking at the failing CI, it _is_ reporting a problem with this PR. Specifically, the DV smoke test is failing on usbdev items. Specifically, `usbdev_csr_hw_reset` and `usbdev_csr_rw` both fail.

I have tested both are passing on my end.

hw/dv/sv/usb20_agent/usb20_block_if.sv
rswarbrick: Sorry, but I'm back to being a little confused. Thinking again, I guess it makes sense for this to be an output port from the interface (it's not like there's anything else that's going to drive it!) But why would we ever want `connected=0`? If we haven't got a use in mind, I suggest getting rid of the signal.

rswarbrick:
Hmm, I've just read this more carefully and thought about #22199, which does something similar. I think we'll probably merge that change (in some form). Since you're touching roughly the same lines, would it make sense to rebase this change over that one?

MubashirSaleem775:
> Hmm, I've just read this more carefully and thought about #22199, which does something similar. I think we'll probably merge that change (in some form). Since you're touching roughly the same lines, would it make sense to rebase this change over that one?

Yes, Its similar to #22199 and you can merge that first and i will do these left changes after that one is merged.  Thanks

hw/dv/sv/usb20_agent/usb20_block_if.sv
MubashirSaleem775: We've implemented a connected signal feature to activate the driver when necessary. Otherwise, it remains in a disconnected state, such as in cases like CSR tests where the USB20_driver isn't needed.

PR #22253, opened 2024-03-26
[usbdev,dv] usbdev stall_trans seq

Saad22386:
This PR includes the usbdev stall_tran_vseqs sequence, designed to validate the stall functionality of usbdev. Upon setting the out_stall bit for a OUT transaction, the DUT is expected to respond with PidTypeStall.

This builds on #22308. Merge that first, which should just leave the final commit from this PR.

### DIFF
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -671,7 +671,7 @@
             - Verify that we will get STALL response when we try to attempt OUT transactions.
             '''
       stage: V2
-      tests: []
+      tests: ["usbdev_stall_trans"]
     }
     {
       name: setup_priority_over_stall_response
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_stall_trans_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_stall_trans_vseq.sv
@@ -1,0 +1,28 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class usbdev_stall_trans_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_stall_trans_vseq)
+
+  `uvm_object_new
+
+  task body();
+    // Configure out transaction
+    configure_out_trans();
+    // Set stall on endp
+    csr_wr(.ptr(ral.out_stall[0].endpoint[endp]), .value(1'b1));
+
+    // Out token packet followed by a data packet
+    call_token_seq(PidTypeOutToken);
+    inter_packet_delay();
+    call_data_seq(PidTypeData0, .randomize_length(1'b1), .num_of_bytes(0));
+
+    // Check that the DUT reponds with the PidTypeStall
+    get_response(m_response_item);
+    $cast(m_usb20_item, m_response_item);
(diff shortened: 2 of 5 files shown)


Saad22386:
@rswarbrick kindly review it.

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -661,7 +661,7 @@
             - Verify that we will get STALL response when try to attempt IN/OUT transactions.
             '''
       stage: V2
-      tests: []
+      tests: [usbdev_stall_trans]
```
rswarbrick, line 664: This is missing some quotes, which is causing CI failures. Other testpoints have what you need (e.g. `pkt_received`)

hw/ip/usbdev/dv/env/seq_lib/usbdev_stall_trans_vseq.sv
```diff
@@ -0,0 +1,29 @@
...
+    // Configure out transaction
+    configure_out_trans();
+    // Set stall on endp
+    csr_wr(.ptr(ral.out_stall[0].endpoint[endp]), .value(1'b1));
+
+    // Out token packet followed by a data packet
+    call_token_seq(PidTypeOutToken);
+    inter_packet_delay();
+    call_data_seq(PidTypeData0, .randomize_length(1'b1), .num_of_bytes(0));
+    inter_packet_delay();
```
rswarbrick, line 20: Is this delay required? I don't think I can see why it would be.

Saad22386: If I remove this delay, it will also work. So, I am going to remove this delay.

rswarbrick (commented):
This looks sensible to me. Looking at the testpoint, it talks about both `in_stall` and `out_stall`. Your sequence is just using `out_stall` and an output transaction. Don't we want to have an `in_stall`/`IN` mode as well?

Saad22386:
> This looks sensible to me. Looking at the testpoint, it talks about both `in_stall` and `out_stall`. Your sequence is just using `out_stall` and an output transaction. Don't we want to have an `in_stall`/`IN` mode as well?

It could either be out_stall or in_stall. The functionality of in_stall has already been tested in another sequence (#21926) that has been merged.

rswarbrick:
Ok. If I understand correctly, we're checking stalls for IN transactions in the `usbdev_in_stall` test, which implements the `in_stall` testpoint. Am I right in thinking that there is no point in the `stall_trans` mentioning OUT transactions? If so, I suggest you add an extra commit to this PR (probably before the existing one) which drops the generic nature of the testpoint.

Saad22386:
> Ok. If I understand correctly, we're checking stalls for IN transactions in the `usbdev_in_stall` test, which implements the `in_stall` testpoint. Am I right in thinking that there is no point in the `stall_trans` mentioning OUT transactions? If so, I suggest you add an extra commit to this PR (probably before the existing one) which drops the generic nature of the testpoint.

ok I'm going to drop the generic nature of the test point.

rswarbrick:
Thanks for the refactor to the testplan. You've also split it out into a separate PR (which seems sensible) and it is at #22308.

Reviewers: Merge that one first! :-)

rswarbrick (approved):
This looks sensible to me and I think this should be good to merge once #22308 has landed.

Saad22386:
I've just rebased over some recent testplan tweaks that collide with it.

Saad22386:
@rswarbrick can you merge this one?

PR #22259, opened 2024-03-26
[usbdev,dv] usbdev phy_config_usb_ref_disable sequence

MubashirSaleem775:
Added usbdev_phy_config_usb_ref_disable sequence to evaluate the functionality of the device's usb_ref_disable feature. Once it's set to disabled, the device will refrain from asserting the ref_value_output signal for synchronization after receiving the Start of Frame (SoF) packet.


hw/ip/usbdev/dv/env/seq_lib/usbdev_phy_config_usb_ref_disable_vseq.sv
```diff
@@ -0,0 +1,25 @@
...
+  `uvm_object_new
+
+  task body();
+    configure_out_trans();
+    // Tx Start of Frame (SoF) packet to the device.
+    // The device will receive the SoF and in response, it will assert the usb_ref_val_o signal.
+    call_sof_seq(PidTypeSofToken);
+    inter_packet_delay();
+    // Verify that the device has asserted the usb_ref_val_o signal after receiving the SoF packet.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o , 1)
```
rswarbrick, line 17: Nit: Extra space before the comma

hw/ip/usbdev/dv/env/seq_lib/usbdev_phy_config_usb_ref_disable_vseq.sv
```diff
@@ -0,0 +1,25 @@
...
+    // De-assert usb_ref_disable signal from phy_config register.
+    csr_wr(.ptr(ral.phy_config.usb_ref_disable),  .value(1'b1));
```
rswarbrick, line 19: Nit: Extra space after the comma.

rswarbrick, line 19: Don't we have to check that `usb_ref_val_o` is false here as well? I think we expect it to drop as a result of writing `usb_ref_disable`?

hw/ip/usbdev/dv/env/seq_lib/usbdev_phy_config_usb_ref_disable_vseq.sv
```diff
@@ -0,0 +1,25 @@
...
+    call_sof_seq(PidTypeSofToken);
+    // Confirm that after setting usb_ref_disable to 1,
+    // the device will not assert the synchronous usb_ref_val_o signal.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o , 0)
```
rswarbrick, line 23: Nit: Extra space before comma

rswarbrick (commented):
There are some nitty things that need tidying up, but this looks pretty sensible to me.

hw/ip/usbdev/dv/env/seq_lib/usbdev_phy_config_usb_ref_disable_vseq.sv
MubashirSaleem775: Yes, we are checking usb_ref_val_o is false in second check after setting this register field on line 23.

rswarbrick: Sorry, I don't think we're quite understanding each other. I think the sequence should probably be:
1. Do some stuff and recognise SoF packet. Check `usb_ref_val_o` is 1.
2. Write `usb_ref_disable`
3. Check that `usb_ref_val_o` is now zero.
4. Now send another SoF packet (which shouldn't be noticed)
5. Check that `usb_ref_val_o` is still zero.

I think that item 3 on the list above is not currently addressed in your code. As you say, item 5 *is* addressed.

MubashirSaleem775: yes, item 3 was not addressed. But now its addressed. Thanks

rswarbrick (approved):
This looks great, thanks.

PR #22277, opened 2024-03-26
[test,ci] fix broken `spi_passthru_test` in CI by adding UJSON CRC checks

timothytrippel:
Since #22101 merged, the `spi_passthru_test` was broken due to ujson CRC's missing between Host-->Device communication. This fixes said test by adding the CRC checks.


pamaury (approved):
Good catch. This might have domino effect: any test that uses memory ujson needs CRC because the ottf framework uses it. So now you are adding CRC to the spi command and i2c targets, so all tests that use them might need some fixing. It's possible you'll end up adding CRC to all tests ^^

timothytrippel:
Yeah, I realized that ... we will see how this goes. May have to do more search and replacing across the code base.

timothytrippel:
Looks like just these changes resolved the issue.

PR #22282, opened 2024-03-27
[spi_host,doc] Clarify that the 'spi_event' interrupt is Status-Type

hcallahan-lowrisc:
It is Status-Type in terms of signalling and handling for software. The appropriate section of comportability docs is linked to clarify this point.


rswarbrick (approved):
Makes sense. It's fun that we have an interrupt called "event" which... is not an event interrupt :-)

PR #22257, opened 2024-03-26
[ci] Update GitHub actions

jwnrt:
Very minor PR which updates the deprecated `actions/checkout@v3` to `@v4` in preparation for Node 16 being removed from the runners.

It also fixes the verible lint action version to v2 as opposed to `main` for some stability.

PR #22287, opened 2024-03-27
[ci] Fix verible lint action version

jwnrt:
Fixes the CI that was broken by #22257

Unlike other actions, the tag is called v2.0 and not v2. See https://github.com/chipsalliance/verible-linter-action/tags for tags.

I've checked the fix on my own fork here: https://github.com/jwnrt/opentitan/pull/13


nbdd0121 (approved):

jwnrt:
Superseded by https://github.com/lowRISC/opentitan/pull/22287 which will also prevent this kind of breakage happening again (for this action at least)

PR #22291, opened 2024-03-27
[hw] Remove PMOD kicad files

jwnrt:
I cannot find the provenance of these files and don't know who owns the copyright and what license they are under. Inside some of these files is "Copyright 2016 Accelerated Designs. All rights reserved." but I cannot find the original files or a license anywhere on the Internet.

Unless anybody is still using this PMOD board, I think we should remove it from the repository to resolve the licensing issue.

It was added in https://github.com/lowRISC/opentitan/pull/4330.

---

Open question: should we remove https://github.com/lowRISC/opentitan/tree/master/hw/ip/usbdev/pmod/dualpmod-kicad as well if it's also unused? This one has a `README` which lists a copyright license (but not a copyright assignment to lowRISC) and the files don't contain a copyright assignment to some other party, so maybe we do own these files.


vogelpi (approved):
Removing those files sounds okay to me and is probably the right thing to do given the Copyright notice + unclear license situation for these. I think we're not using them anymore. Those were mostly useful when working the the NexysVideo.

As for the dualpmod-kicad: I think we should keep those. It's currently unused but this may be of interest to the community. Whenever using an FPGA board other than the CW310/CW340 (these have USB PHY onboard), some kind of board is needed to interface a USB connector directly to the FPGA pins if people want to use USB. This is what these PMOD boards are for.

I am pretty sure we do have the copyright for these files. They were developed by @mdhayter during his sabbatical at the Cambridge Computer Lab around 2018/2019. That's also when Mark made the majority of his USBDEV and regtool contributions in the OpenTitan repo for which we have the copyright.

PR #22294, opened 2024-03-27
[ci] Simplify exec-check script

jwnrt:
This script checks for files with extensions that should not be executable.

### DIFF
--- a/ci/scripts/exec-check.sh
+++ b/ci/scripts/exec-check.sh
@@ -3,34 +3,41 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+# Find executable files that should not be executable.
+
 set -e
 
-allowed_suffixes=(
+# Files with these extensions may be executable.
+allowed_extensions=(
     py
     sh
 )
 
-suff_re=""
-for suff in "${allowed_suffixes[@]}"; do
-    suff_re="${suff_re}$suff\\|"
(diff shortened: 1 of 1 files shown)


ci/scripts/exec-check.sh
```diff
@@ -5,32 +5,23 @@
 
 set -e
 
-allowed_suffixes=(
-    py
-    sh
```
nbdd0121, line 10: I'd still prefer to have an array upfront.

ci/scripts/exec-check.sh
```diff
@@ -5,32 +5,23 @@
...
-}
-trap 'rm -f "$TMPFILE"' EXIT
-
-find -name vendor -prune -o \
-     -name .git -prune -o \
-     -type f -executable -name '*.*' -not -regex "$path_re" \
-     -print >"$TMPFILE"
-if [ -s "$TMPFILE" ]; then
+# Fail if any exist.
+if [ -s "$bad_files" ]; then
```
nbdd0121, line 22: `-s` expects a path. You probably want to use `-n`?

jwnrt: Thanks, this was working when I tested with one file because `"$bad_files"` was the name of a file. It doesn't work with multiple files. Changed to `-n`.

rswarbrick (approved):
This looks good to me. And I like how you've avoided `mktemp` (I think I was being lazy when I wrote the first version here...)

nbdd0121 (approved):

PR #22284, opened 2024-03-27
[ot_certs] Fix substitution bug and add integration test

pamaury:
Fix a small bug in integer parsing and add an end to end substitution test.

PR #22299, opened 2024-03-27
[ot_certs] Add support for validity dates

pamaury:
Only review the last commit, the rest is from #22284

Tagging @jwnrt and @nbdd0121 for unsafe review.

### DIFF
--- a/sw/device/silicon_creator/lib/cert/cdi_1.hjson
+++ b/sw/device/silicon_creator/lib/cert/cdi_1.hjson
@@ -63,6 +63,13 @@
         subject: {
             serial_number: { var: "owner_pub_key_id", convert: "lowercase-hex" },
         },
+        // From the Open Profile for DICE specification:
+        // https://pigweed.googlesource.com/open-dice/+/refs/heads/main/docs/specification.md#x_509-cdi-certificates
+        // The certificate expiration time is set to a fixed date in the future.
+        // The "not before" date is chosen to be in the past (publication date of the DICE TCG) since the device does not
+        // have a reliable way to get the time.
+        not_before: "20180322235959Z",
+        not_after: "99991231235959Z",
         subject_public_key_info: {
             algorithm: "ec-public-key",
             curve: "prime256v1",
--- a/sw/host/ot_certs/src/x509.rs
+++ b/sw/host/ot_certs/src/x509.rs
@@ -6,7 +6,8 @@ use anyhow::{bail, Context, Result};
 use num_bigint_dig::BigUint;
 use std::collections::HashMap;
 
-use openssl::asn1::{Asn1IntegerRef, Asn1OctetStringRef, Asn1StringRef};
+use foreign_types::ForeignTypeRef;
+use openssl::asn1::{Asn1IntegerRef, Asn1OctetStringRef, Asn1StringRef, Asn1TimeRef};
 use openssl::bn::{BigNum, BigNumContext, BigNumRef};
 use openssl::ec::{EcGroupRef, EcKey};
 use openssl::ecdsa::EcdsaSig;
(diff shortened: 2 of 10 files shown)


sw/device/silicon_creator/lib/cert/cdi_1.hjson
```diff
@@ -63,6 +63,13 @@
         subject: {
             serial_number: { var: "owner_pub_key_id", convert: "lowercase-hex" },
         },
+        // From the Open Profile for DICE specification:
+        // https://pigweed.googlesource.com/open-dice/+/refs/heads/main/docs/specification.md#x_509-cdi-certificates
+        // The certificate expiration time is set to a fixed date in the future.
+        // The "not before" date is chosen to be in the past since the device does not
+        // have a reliable way to get the time.
+        not_before: "20230101000000Z",
```
timothytrippel, line 71: Just curious: how did you select the "not_before" value? Looks like the [open profile](https://pigweed.googlesource.com/open-dice/+/refs/heads/main/docs/specification.md?pli=1#x_509-cdi-certificates) had a different value (which corresponded to the date of publication of the TCG DICE spec. Not that it matters, just was curious :)

timothytrippel (approved):

sw/device/silicon_creator/lib/cert/cdi_1.hjson
pamaury: The DICE specification said any date in the past is acceptable so I just took a random date. But it's true that the publication date of the spec makes the most sense ^^ I am happy to change if you think that's worth it, it's now it's trivial to change in the hjson template.

sw/host/ot_certs/src/x509.rs
```diff
@@ -75,6 +76,35 @@ fn asn1octets_to_vec(s: &Asn1OctetStringRef) -> Value<Vec<u8>> {
     Value::literal(s.as_slice().to_vec())
 }
 
+fn asn1time_to_string(time: &Asn1TimeRef) -> Result<Value<String>> {
+    // SAFETY: OpenSSL guarantees that an ASN1_TIME is in fact just a typedef for ASN1_STRING
```
nbdd0121, line 80: nit, as this does not have an unsafe block:
```suggestion
    // OpenSSL guarantees that an ASN1_TIME is in fact just a typedef for ASN1_STRING
```

sw/host/ot_certs/src/x509.rs
```diff
@@ -75,6 +76,35 @@ fn asn1octets_to_vec(s: &Asn1OctetStringRef) -> Value<Vec<u8>> {
...
+        bail!("time uses type {time_type} but only GeneralizedTime is supported")
+    }
+    // SAFETY: the above pointer is guaranteed to be valid since `time` is valid reference.
+    // From the OpenSSL documentation: ASN1_STRING_get0_data() returns an internal pointer to
+    // the data of x. Since this is an internal pointer it should not be freed or modified in any way.
+    let time_str = unsafe {
+        let len = openssl_sys::ASN1_STRING_length(time_str);
+        let data = openssl_sys::ASN1_STRING_get0_data(time_str);
+        std::slice::from_raw_parts(data, len as usize)
+    };
```
nbdd0121, line 100: How about just use
```rust
let time_str = unsafe { Asn1StringRef::from_ptr(time_str) }.as_str();
```

pamaury: I should I thought about that ^^ Thanks for the suggestion.

sw/device/silicon_creator/lib/cert/cdi_1.hjson
timothytrippel: no need to change, was just curious

PR #22300, opened 2024-03-27
[dv,usbdev] Test plan refinements

alees24:
Drop a few unnecessary tests such as those which relate to Control Transfers above the level of packet transactions.

Clear up the descriptions in a number of tests and clarify the required behavior in some other tests.

Modifications for the altered RTL in places, particularly the separation of the single 'Available Buffer FIFO' into two separate buffer FIFOs for SETUP and OUT packets.

### DIFF
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -76,19 +76,6 @@
       tests: ["usbdev_phy_pins_sense"]
     }
     {
-      name: wake_events
-      desc: '''
-            Verify the read functionality of wake_event register via TLUL interface.
-
-            - Allow the usbdev to go to deep sleep mode then read the wake_events register
-              from the TLUL interface.
-            - Verify that the read value of the register should reflect the correct value of module_aon
-              as well as bus_disconnected and bus_reset during deep sleep.
-            '''
-      stage: V2
-      tests: []
-    }
-    {
       name: av_buffer
       desc: '''
             Verify the functionality of availability of buffer.
@@ -413,46 +400,6 @@
       tests: []
     }
     {
-      name: enable
-      desc: '''
-            Verify that USB connect to interface when enable is set.
-
-            - set USB_CTRL_enable pin high.
-            - Make sure phy_config_pinflip isn't set.
-            - Verify that the usb_dp_pullup_o pin goes high.
-            '''
-      stage: V2
-      tests: ["usbdev_enable"]
-    }
-    {
-      name: resume_link_active
-      desc: '''
-            Verify that when Write a 1 to this bit to instruct usbdev to jump to the LinkResuming state
-            then write will only have an effect when the device is in the LinkPowered state.
-
-            - Perform no operation in J state for more than 3 ms.
-            - Read the link state from usbstate register to see if it is in powered state and then send the
-              resume signal (K state) for 20 ms and set the usbctrl.resume_link_active.
-            - Verify that the resume_link_active interrupt goes high and device will be resumed.
-            '''
-      stage: V2
-      tests: []
-    }
-    {
-      name: device_address
-      desc: '''
-            Verify that the device ignores all the packets that have invalid address.
-
-            - Assigned device address to (usbctrl[22:16]) and send packets to a valid address.
-            - Send packets to invalid address.
-            - Verify that the device responds to all packets sent by the DV environment (USB side)
-              that are sent to its assigned address.
-            - Verify that the device ignores all packets that are sent to other address.
-            '''
-      stage: V2
-      tests: []
-    }
-    {
       name: link_in_err
       desc: '''
             Verify that if a packet to an IN endpoint started to be received but was then dropped due to an error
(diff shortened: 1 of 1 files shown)


a-will (approved):

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -511,12 +492,52 @@
...
+            - Verify that the 'usb_dp_pullup_o' pin goes high.
+            '''
+      stage: V2
+      tests: ["usbdev_enable"]
+    }
+    {
+      name: resume_link_active
+      desc: '''
+            Verify that when Write a 1 to this bit to instruct usbdev to jump to the LinkResuming state
+            then write will only have an effect when the device is in the LinkPowered state.
```
rswarbrick, line 511: I'm not completely sure I understand this description. Is this saying that we should verify that setting the `resume_link_active` bit causes usbdev to jump to the `LinkResuming` state? (I think there should probably be a full stop at this point). Then are we saying a write will not have any effect until the device gets to LinkPowered?

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -511,12 +492,52 @@
...
+
+            - Perform no operation in J state for more than 3 ms.
```
rswarbrick, line 513: Grammar nit: I'd suggest rephrasing this as something like "Sit in J state for more than 3ms, performing no operation." (otherwise it looks like we need to make sure that we're ensuring none of the operations we perform takes more than 3ms :-D)

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -511,12 +492,52 @@
...
+            - Read the link state from usbstate register to see if it is in powered state and then send the
+              resume signal (K state) for 20 ms and set the usbctrl.resume_link_active.
```
rswarbrick, line 515: This feels rather woolly. Presumably we expect the link not to be in the powered state at this point? Maybe we should say that here.

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -511,12 +492,52 @@
...
+            - Verify that the resume_link_active interrupt goes high and device will be resumed.
```
rswarbrick, line 516: (Genuine USB question): After the interrupt goes high, I guess we want to see that the device resumes? Maybe the testpoint should say this?

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -511,12 +492,52 @@
...
+            '''
+      stage: V2
+      tests: []
+    }
+    {
+      name: device_address
+      desc: '''
+            Verify that the device ignores all packets intended for other addresses.
```
rswarbrick, line 524: I'd possibly rephrase this: the address tells us where the packet is intended to go :-) Maybe "Verify that the device ignores all packets whose addresses do not match the device's address."?

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -511,12 +492,52 @@
...
+
+            - Assigned device address to (usbctrl[22:16]) and send packets to this address.
```
rswarbrick, line 526: Nit (grammar): I think this should be "Assign"?

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -511,12 +492,52 @@
...
+            - Send packets to other addresses.
+            - Verify that the device responds to all packets sent by the DV environment (USB side)
+              that are sent to its assigned address.
```
rswarbrick, line 529: As we're looking at this text anyway... This could be a lot shorter! Probably "Verify that the device responds to all packets sent to its assigned address." ?

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -529,8 +550,8 @@
             Verify the SETUP stage of enumeration.
 
             - Send a setup token with a valid setup packet containing the address(0), endpoint number(0)
```
rswarbrick, line 552: Do we have an opinion about the address and endpoint here? If not, maybe it makes sense to make this test more general?

hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -696,9 +692,9 @@
             Verify that the a Link Reset or SETUP transaction cancels any waiting IN transactions
             by clearing the rdy bit in the configin register of all endpoints.
 
-            - configure an endpoint for transmission by setting field in configin_ep register and
-              set the rdy bit 1.Then send a SETUP transaction to that endpoint or issue a link reset.
-            - Verify that the rdy bit in the configin_ep will be cleared and pend bit will be set to 1.
+            - Configure an endpoint for transmission by setting fields in 'configin[ep]' register and
+              set the 'rdy' bit to 1. Then send a SETUP transaction to that endpoint or issue a link reset.
```
rswarbrick, line 696: Nit(grammar): I think this should be "setting"

rswarbrick (approved):
This definitely improves things. A couple of comments, but I'm happy for this to go in with or without the tweaks :-)

hw/ip/usbdev/data/usbdev_testplan.hjson
alees24: I'll leave this test description to be addressed in #22129.

hw/ip/usbdev/data/usbdev_testplan.hjson
alees24: As noted in #22129, there is no such interrupt; this description is somewhat confused.

PR #22306, opened 2024-03-28
[usbdev] usbdev predictor amendment

Saad22386:
This PR includes modifications to the usbdev_packet_classifier class, removes unnecessary stuff, and adjusts the usbdev_transaction_manager class to accommodate changes made in the pkt_classifier class

### DIFF
--- a/hw/ip/usbdev/dv/env/usbdev_packet_classifier.sv
+++ b/hw/ip/usbdev/dv/env/usbdev_packet_classifier.sv
@@ -5,118 +5,67 @@
 class usbdev_packet_classifier extends uvm_object;
   `uvm_object_utils(usbdev_packet_classifier)
 
-  // Define packet types
-  typedef enum logic [3:0] {
-  TOKEN_SETUP = 4'b1101,
-  TOKEN_IN = 4'b1001,
-  TOKEN_OUT = 4'b0001,
-  SOF = 4'b0101,
-  DATA_0 = 4'b0011,
-  DATA_1 = 4'b1011,
-  HANDSHAKE_ACK = 4'b0010,
-  HANDSHAKE_NACK = 4'b1010,
-  HANDSHAKE_STALL = 4'b1110,
-  UNSUPPORTED_PACKET = 4'b0000
-  }usbpackettype_e;
   // Packet fields
-  bit [3:0] pid_iden;         // Packet ID identifier
-  logic [7:0] pid;            // Packet ID identifier
-  logic [6:0] address;        // Address field for token packets
-  logic [3:0] endpoint;       // Endpoint field for token packets
-  logic [10:0] frameNumber;   // Frame number for SOF packets
-  bit data[];                 // Data fields for data packets
-  logic [4:0] crc5;           // CRC5 field for token packets
-  logic [15:0] crc16;         // CRC16 field for data packets
-  usbpackettype_e packetType; // Type of packet
+  logic [7:0]  pid;            // Packet ID identifier
+  logic [6:0]  address;        // Address field for token packets
+  logic [3:0]  endpoint;       // Endpoint field for token packets
+  logic [10:0] frameNumber;    // Frame number for SOF packets
+  bit          data[];         // Data fields for data packets
+  logic [4:0]  crc5;           // CRC5 field for token packets
+  logic [15:0] crc16;          // CRC16 field for data packets
+  pid_type_e   m_pid_type;     // Type of pid
 
   // Constructor
-  function new(string name  = "packet_classifier");
+  function new(string name = "packet_classifier");
     super.new(name);
-    pid_iden = 4'b0000;
-    pid = 8'b00000000;
-    address = 7'b0000000;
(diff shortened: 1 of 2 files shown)


Saad22386:
@rswarbrick and @alees24, kindly review it.
Thnaks

rswarbrick:
I'm sorry but I don't think I can review this change: I don't properly understand what the code here is supposed to be doing. You'll need to add some comments to explain e.g. the functions that are being used. I'm willing to believe that the improvements in #22270 broke stuff here, but we *definitely* shouldn't change the code in `usbdev_packet_classifier` without documenting what it's supposed to be doing.

Saad22386:
I'm sorry, but I don't think I can review this change. I don't properly understand what the code here is supposed to be doing. You'll need to add some comments to explain, e.g., the functions that are being used. I'm willing to believe that the improvements in #22270 broke stuff here, but we definitely shouldn't change the code in `usbdev_packet_classifier` without documenting what it's supposed to be doing.

The first four bits of the PID are used by the usbdev_packet_classifier class to identify packets. Based on this identification, the class extracts fields from the packet.

Saad22386:
> I'm sorry but I don't think I can review this change: I don't properly understand what the code here is supposed to be doing. You'll need to add some comments to explain e.g. the functions that are being used. I'm willing to believe that the improvements in #22270 broke stuff here, but we _definitely_ shouldn't change the code in `usbdev_packet_classifier` without documenting what it's supposed to be doing.

The first four bits of the PID are used by the usbdev_packet_classifier class to identify packets. Based on this identification, the class extracts fields from the packet.

Saad22386:
> I'm sorry but I don't think I can review this change: I don't properly understand what the code here is supposed to be doing. You'll need to add some comments to explain e.g. the functions that are being used. I'm willing to believe that the improvements in #22270 broke stuff here, but we _definitely_ shouldn't change the code in `usbdev_packet_classifier` without documenting what it's supposed to be doing.

The first four bits of the PID are used by the usbdev_packet_classifier class to identify packets. Based on this identification, the class extracts fields from the packets.

Saad22386:
If we don't alter the PID bits after #22270, the predictor shouldn't proceed because packet identification fails at this point.

Saad22386:
@alees24, please take a look at this. The suggested change is causing the predictor in the current environment to halt its progress.

alees24:
> @alees24, please take a look at this. The suggested change is causing the predictor in the current environment to halt its progress.

Presumably we are seeing incomplete code in the scoreboard at this stage?

Checking against expectations and then reporting any mismatches using only a UVM_DEBUG-level with a uvm_info message means that they typically don't even appear in logs; how much work remains to be done before the loop can be closed and a mismatch in the scoreboard will yield a report and test failure?

hw/ip/usbdev/dv/env/usbdev_packet_classifier.sv
```diff
@@ -43,7 +43,7 @@ class usbdev_packet_classifier extends uvm_object;
   endfunction
   // Function to check the packet
   function void checkPacket(input bit  pkt[]);
-    for(int i = 0; i <= 3; i++) begin
+    for(int i = 4; i <= 7; i++) begin
       pid_iden = {pid_iden, pkt[i]};
     end
```
alees24, line 48: The member variable `pid_iden` appears not to be used anywhere, so the `packetType` assignment could be done directly from pid[3:0], and the first decode loop be dropped entirely. That would be easier to understand when compared with the specification, and it would probably be helpful to explain in comments the relationship between the PID and the 'usbpackettype_e'

Saad22386: I am going to do a PR that will make these thing a lot more clearer.

Saad22386:
> > @alees24, please take a look at this. The suggested change is causing the predictor in the current environment to halt its progress.
> 
> Presumably we are seeing incomplete code in the scoreboard at this stage?
> 
> Checking against expectations and then reporting any mismatches using only a UVM_DEBUG-level with a uvm_info message means that they typically don't even appear in logs; how much work remains to be done before the loop can be closed and a mismatch in the scoreboard will yield a report and test failure?

Presently, the scoreboard prioritizes checking for PID errors. If a PID error is detected, the packet is dropped from the queue without undergoing packet comparison. Subsequently, if there is no PID error, the scoreboard proceeds to examine for CRC errors. If a CRC error is present, the packet is also dropped from the queue without undergoing packet comparison. However, when neither error is detected, the scoreboard proceeds with packet comparison.

Furthermore, it's important to clarify that "IN" transactions currently rely on "OUT" transactions in the predictor. This dependency arises because we store data in the predictor memory through "OUT" transactions and then read that data via "IN" transactions. If data were directly written to the buffer in sequence, the predictor would fail to retrieve data directly written into the buffer without an "OUT" transaction.

alees24 (approved):
Thanks for tidying this up.

Saad22386:
@alees24 could you proceeds with this PR?

PR #22307, opened 2024-03-28
[usbdev] Rewrite usbdev setup stage seq description

Saad22386:
This PR incorporates changes to the description of the usbdev setup_stage sequence to define a more generic behavior.


rswarbrick:
Thanks for tweaking this. I've just rebased over some recent testplan tweaks that collide with it.

rswarbrick (approved):

PR #22314, opened 2024-03-28
[dv,usbdev] Reliability fixes

alees24:
This PR addresses a couple of reliability issues with the current usbdev DV sequences:

- The usb20_monitor suffers the same sampling/aliasing issues as the usb20_driver code when attempting to receive and decode packets on a 12MHz clock, with no ability to adjust the phase of the clock. As an interim workaround this adjusts the sampling to use the positive- or negative-edge of the clock, but a cleaner and more general solution should be implemented at some point to permit testing of frequency- and phase-differences between host and device.
- The unnecessary startup code in the usbdev_nak_trans_vseq and usbdev_av_buffer_vseq conflicted with the current Bus Reset behavior of the usb20_driver, particularly the interrupt clearing because bus resets set interrupt state bits.


rswarbrick (approved):
This looks sensible and thank you for the detailed comments!

PR #22317, opened 2024-03-28
[keymgr_dpe, dv] fixes issue where exp_kmac_key for sideload keys

bobby-rivos:
When a sideload key for kmac was never generated in tests due to randomization there was no mechanism to reset the exp_kmac_key to 0 as it should be. So in some seeds after disabling keymgr_dpe and checking sideload keys a comparison was being done against the exp_kmac key and the kmac_key in rtl which was a 0. The exp_kmac key would not be properly set to 0 and a false fail was happening. This commit resets the exp_kmac key to 0 when there is no sideload key generated.

### DIFF
--- a/hw/ip/keymgr_dpe/dv/env/keymgr_dpe_if.sv
+++ b/hw/ip/keymgr_dpe/dv/env/keymgr_dpe_if.sv
@@ -585,6 +585,7 @@ interface keymgr_dpe_if(input clk, input rst_n);
           kmac_key_exp <= '{1'b1, kmac_sideload_key_shares};
           is_kmac_key_good <= 1;
         end else begin
+          kmac_key_exp <= '0;
           kmac_key_exp.valid <= 0;
           is_kmac_key_good   <= 0;
         end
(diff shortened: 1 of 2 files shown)


hw/ip/keymgr_dpe/dv/env/keymgr_dpe_if.sv
```diff
@@ -585,6 +585,7 @@ interface keymgr_dpe_if(input clk, input rst_n);
           kmac_key_exp <= '{1'b1, kmac_sideload_key_shares};
           is_kmac_key_good <= 1;
         end else begin
+          kmac_key_exp <= '0;
           kmac_key_exp.valid <= 0;
```
ballifatih, line 589: nit: we can remove `kmac_key_exp.valid <= 0` assignment.

ballifatih (approved):
LGTM, thanks @bobby-rivos.

hw/ip/keymgr_dpe/dv/env/keymgr_dpe_if.sv
```diff
@@ -585,6 +585,7 @@ interface keymgr_dpe_if(input clk, input rst_n);
           kmac_key_exp <= '{1'b1, kmac_sideload_key_shares};
           is_kmac_key_good <= 1;
         end else begin
+          kmac_key_exp <= '0;
```
rswarbrick, line 588: This looks sensible, but is it at the right time? Would it be easier to qualify the places we look at `kmac_key_exp` with `kmac_sideload_status == SideLoadAvail`? I think this is just a pair of assertions (`CheckKmacKey` and `CheckKmacKeyValid`)?

rswarbrick (commented):
Sorry for not seeing this review request before. Is the PR still relevant? If so, I wonder whether there's a tweak that would do the change more simply. I've left a comment on the code.

PR #22324, opened 2024-03-28
[entropy_src] Improve documentation

vogelpi:
This PR contains a series of commits to correct and clarify the documentation of the ENTROPY_SRC hardware block.

PR #22327, opened 2024-03-29
[entropy_src] Add missing SEC_CM labels and increase version number

vogelpi:


andreaskurth (approved):
Approved based on the signoff analysis in #20973 on the condition that #22324 gets merged first. Thx @vogelpi

PR #22328, opened 2024-03-29
[dv,mem_bkdr] Fix handling of multiple tiles in sram

matutem:
Pass the number of bits selecting a tile to various sram backdoor functions. This was incompletely done in the past, which had no impact in the open-source ram configuration since it has a single tile, but causes failures in configurations with multiple tiles.


andreaskurth (approved):
LGTM, thx for this fix @matutem!

PR #22329, opened 2024-03-29
[sysrst_ctrl,sival] Various improvements

pamaury:
Read the commits for a description. In short:
- fix `ulp_z3_wakeup`
- introduce gpio monitor test util and use it in tests

### DIFF
--- a/sw/device/tests/sysrst_ctrl_ulp_z3_wakeup_test.c
+++ b/sw/device/tests/sysrst_ctrl_ulp_z3_wakeup_test.c
@@ -14,6 +14,7 @@
 #include "sw/device/lib/testing/flash_ctrl_testutils.h"
 #include "sw/device/lib/testing/pwrmgr_testutils.h"
 #include "sw/device/lib/testing/rstmgr_testutils.h"
+#include "sw/device/lib/testing/sysrst_ctrl_testutils.h"
 #include "sw/device/lib/testing/test_framework/check.h"
 #include "sw/device/lib/testing/test_framework/ottf_main.h"
 #include "sw/device/lib/testing/test_framework/ottf_utils.h"
@@ -105,6 +106,8 @@ static void pinmux_setup(void) {
     CHECK_DIF_OK(dif_pinmux_output_select(&pinmux, kOutputPads[i],
                                           kPeripheralOutputs[i]));
   }
+  sysrst_ctrl_testutils_setup_dio(&pinmux);
+  sysrst_ctrl_testutils_release_dio(&sysrst_ctrl, true, true);
 }
 
 static uint8_t read_phase_pins(void) {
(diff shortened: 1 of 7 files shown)


jwnrt (approved):
This LGTM, thanks especially for making that nice VCD dumping module. Regarding sending the test phase over GPIO, that looks like it will work quite nicely but FYI I've seen other tests use the reset reason from the `rstmgr` to restore the correct test phase when reset during the test. I'm not sure whether that works for deep sleep, but your solution is good as well.

sw/host/tests/chip/sysrst_ctrl/sysrst_ctrl_ulp_z3_wakeup.rs
```diff
@@ -86,45 +110,95 @@ fn chip_sw_sysrst_ctrl_ulp_z3_wakeup(
...
     log::info!("Go to PhaseDone");
-    MemWriteReq::execute(uart, test_phase_addr, &[TestPhase::Done as u8])?;
+    set_test_phase(transport, TestPhase::Done)?;
+
+    // Stop monitoring and check that there were no glitches.
+    let events = gpio_mon.read(false)?;
+    let mut events_iter = events.iter();
+    let Some(first_event) = events_iter.next() else {
+        bail!("Expected at least one GPIO event during reset");
+    };
```
engdoreis, line 170:
```suggestion
    let first_event = events_iter.next().context("Expected at least one GPIO event during reset");
```

engdoreis (approved):
LGTM

sw/device/tests/sysrst_ctrl_ulp_z3_wakeup_test.c
```diff
@@ -102,6 +106,18 @@ static void pinmux_setup(void) {
...
+  sysrst_ctrl_testutils_release_dio(&sysrst_ctrl, true, true);
+}
+
+static uint8_t read_phase_pins(void) {
+  uint32_t gpio_state;
+  CHECK_DIF_OK(dif_gpio_read_all(&gpio, &gpio_state));
+  // Since the host may not be able to change all pins atomically, we use
+  // a Gray code encoding so that it suffices to change one pin to go to
+  // the next phase.
+  static const uint8_t kGrayCode[] = {0, 1, 3, 2, 7, 6, 4, 5};
```
jettr, line 119: Note the jump between `2` and `7` is 2 bits, the graycode should be:

`static const uint8_t kGrayCode[] = {0, 1, 3, 2, 6, 7, 5, 4};`

to only flip one bit at a time

pamaury: This is for decoding the Gray code, not encoding, so it's the "reversed" mapping.

jettr: Ah, I see. You might want to add a comment saying that it is an inverse look up, but the code it correct. Thanks!

PR #22345, opened 2024-03-29
[i2c] Split the target and controller FSMs

a-will:
Split the target and controller FSMs both to simplify the code for better maintainability / readability and to lay the groundwork for multi-controller support. Note that there will likely be some increase in area for duplicated structures here, but it turns out that much of the state was already independent.

Also fix a bug where the controller-mode stretch counter would continue counting cycles even when the target had already released SCL.

### DIFF
--- a/hw/ip/i2c/rtl/i2c_target_fsm.sv
+++ b/hw/ip/i2c/rtl/i2c_target_fsm.sv
@@ -4,11 +4,9 @@
 //
 // Description: I2C finite state machine
 
-module i2c_fsm import i2c_pkg::*;
+module i2c_target_fsm import i2c_pkg::*;
 #(
-  parameter int FifoDepth = 64,
   parameter int AcqFifoDepth = 64,
-  localparam int FifoDepthWidth = $clog2(FifoDepth+1),
   localparam int AcqFifoDepthWidth = $clog2(AcqFifoDepth+1)
 ) (
   input        clk_i,  // clock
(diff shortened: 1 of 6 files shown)


a-will:
At least for now, this is just a draft to show how it could be done.

a-will:
Added a commit to implement the controller halt events CSR as described before. It also at least attempts to eliminate some race conditions with software modifying CSRs during automatic Stop generation.

I did switch back to having SCL held low. If the Stop is going to happen automatically anyway, this seems acceptable, since it puts SCL in the correct place to prepare for a Stop (or Start, if software handles it and prefers).

hw/ip/i2c/rtl/i2c_target_fsm.sv
```diff
@@ -1359,8 +629,15 @@ module i2c_fsm import i2c_pkg::*;
         if (bit_ack) begin
           if (address_match) begin
             if (acq_fifo_full_or_last_space) begin
+              // FIXME: Not sure about this logic here. Might want to remove
+              // this. The first read ends up proceeding beyond this point anyway,
+              // even if it times out waiting for the ACQ FIFO.
+              // After viewing the SMBus specs, it looks like we can't NACK
+              // here anyway. We'll just have to return garbage data (which is
+              // what we would've done for the first read that timed out).
```
a-will, line 637: This is only true for SMBus mode, whereas I2C TPM uses the address NACK to indicate that the TPM isn't ready for internal reasons. So this may need to be configurable.

a-will: Moved this to a discussion to be handled later.

a-will:
For a gauge of the area impact, here are the FPGA results:

| Resource | Before | After |
| - | - | - |
| Flops | 1029 | 1102 |
| LUTs | 1612 | 1588 | 

The LUTs decrease as the FSMs get a little simpler, and the flops increase. Most likely, the flop increase is largely due to the duplication of 2 counters:

- 20-bit `tcount_q`
    - Used to count out hold times on driven bits and setup times for stretching on the target.
    - Used extensively on the host.
    - This counter is larger than necessary: It only need to be capable of a max count of 5 us.
        -  Even if we support up to a 1 GHz core clock, 13 bits would be enough.
- 32-bit `stretch_idle_cnt`
    - For the target, used to count a timeout for bus idle
    - For the host, used to count how long a target has stretched the clock
    - This counter is needlessly large: It only need to be capable of counting to 35 ms.
        - Even if we support up to a 1 GHz core clock, 26 bits would be enough.

a-will:
The DV failure is unrelated (in LC_CTRL block level in the VOLATILE_UNLOCK_RAW configuration).

hcallahan-lowrisc (approved):
Code changes LGTM, glad this task ended up being not too difficult, as there wasn't too much shared state after all.
I only have one meta question before merging. Do we want to merge these changes into master? I understood there was some discussion about derisking by keeping the M2 design for some instances, and then working on all extensions as a separate block. In that case, maybe we want to target these changes into a separate branch, or maybe even a new copy of the block itself? @andreaskurth @mundaym may have some input here.

a-will:
> Code changes LGTM, glad this task ended up being not too difficult, as there wasn't too much shared state after all. I only have one meta question before merging. Do we want to merge these changes into master? I understood there was some discussion about derisking by keeping the M2 design for some instances, and then working on all extensions as a separate block. In that case, maybe we want to target these changes into a separate branch, or maybe even a new copy of the block itself? @andreaskurth @mundaym may have some input here.

I don't think that strategy makes sense, since it suggests the current i2c block is acceptable in any way. There isn't some proven IP to fall back on here. Make no mistake: The current code has numerous bugs that I have started fixing, and the regression results reveal an IP that just is not at tape-out quality.

Trying to have two separate IPs only adds overhead in a time when we can't afford it. We don't have the resources.

Instead, I am cleaning up the IP in ways that make it more robust for single controller environments. In the series of PRs that follow, there are fixes to existing features we targeted for D2 sign-off, and the DV tests substantially improve. Splitting the FSM into two simplifies the code and makes it easier to work with, and it enables us to share fixes.

Meanwhile, it looks like area will not be significantly impacted, as the following cleanup PRs reduce it from here (have to check the area usage again after the last commit for the target mode cleanups, though).

In any case, please don't hold up my PRs for that discussion if it needs to continue. The older version can be retrieved from the git history if necessary.

hcallahan-lowrisc:
I'm happy to merge this, and we can discuss the meta items later and change the workflow if needed. Approved.

andreaskurth:
Thanks @a-will for this work and @hcallahan-lowrisc for reviewing. :+1:

These changes can go into the i2c HW IP block on `master`.

PR #22351, opened 2024-03-29
[dice] move keysave out of keygen function

timothytrippel:
We only want to save the current stage's attestation private key to OTBN DMEM _after_ we have used the previous stage's private key to endorse our current stage's certificate. This was a slight refactoring error in #22303, since previously, we were always overwriting the prior stage's private key in OTBN DMEM before using it.


timothytrippel:
CC: @vbendeb this might explain the cert chain verification issue

vbendeb:
> CC: @vbendeb this might explain the cert chain verification issue

Can you generate some updates certs?

jadephilipoom (approved):

Issue #22352, opened 2024-03-29
[edn] Investigate FSM coverage


vogelpi:
### Description

During M2, we did quite some modifications in EDN and while the majority of coverage metrics remained above the 90% bar, the FSM coverage dropped significantly and is now at 82%. This is not sufficient for V2(S) and we should investigate this as part of M4. It may be that adjusting and reviewing coverage exclusions is sufficient (there were some modifications to the FSMs in M2).

vogelpi:
I've filed PR #23092 to increase the FSM coverage through the `edn_err` test (and fix an RTL bug). Together with @h-filali , I've discussed that we should maybe modify the `edn_alert` test to directly signal the required CSRNG error status after arriving in the desired state. Right now, some more operations are done meaning the FSM likely moves out of short-lived states.

vogelpi:
With #23092 applied, FSM coverage goes up to 85%. The missing transitions are mostly from various states into `RejectCsrngEntropy`. I think we need to change the alert test as discussed previously (inject error response from CSRNG directly after entering the expected state).

vogelpi:
The latest regression results show that the FSM coverage is now back above the threshold. I am thus closing this issue. We can proceed with the V2(S) signoff.

Issue #20818, opened 2024-01-12
[edn/dv] edn_disable_auto_req_mode test sequence failing


h-filali:
### Description

edn_disable_auto_req_mode fails sometimes. It looks like this has something to do with the disablement of the EDN.
If the EDN is disabled via backdoor during a TL transaction, the TL sequencer hangs and is thus killed automatically.
To reproduce use this git hash ``f3af826941`` and run ``util/dvsim/dvsim.py hw/ip/edn/dv/edn_sim_cfg.hjson -i edn_disable_auto_req_mode -t xcelium --fixed-seed 61070996226483181682937165685756564904249297810070954908397241823715099635498``
Two more failing seeds I could find are:
``83106884606859576866395536417688321183151628139158240429607179727476328981642``
``93668555437452971905965393254453301839962811678600888621534390713802600691300``

A possible solution could involve something like killing and restarting the tl sequence before and after disabling the EDN.

Error message:
``UVM_ERROR @   3783561 ps: (uvm_sequencer_base.svh:757) uvm_test_top.env.m_tl_agent_edn_reg_block.sequencer [SEQREQZMB] The task responsible for requesting a wait_for_grant on sequencer 'uvm_test_top.env.m_tl_agent_edn_reg_block.sequencer' for sequence 'm_tl_host_base_seq' has been killed, to avoid a deadlock the sequence will be removed from the arbitration queues``

vogelpi:
This still seems to be an issue. The `edn_disable_auto_req_mode` test has currently a pass rate of roughly 85%. This is not sufficient for V2(S). We need to investigate this for M4.

vogelpi:
Update: #22752 fixed quite some test failures. The pass rate of edn_disable_auto_req_mode is now back up at 98%. The remaining failures have a predictable pattern though (see below). We should investigate what the issue is here.

Example failure pattern:
```
6.edn_disable_auto_req_mode.54433158439156122443312458540656012130830202329242750771247128625030336802488
Line 261, in log /container/opentitan-public/scratch/os_regression/edn-sim-vcs/6.edn_disable_auto_req_mode/latest/run.log

  UVM_ERROR @  35973176 ps: (edn_scoreboard.sv:313) [uvm_test_top.env.scoreboard] Check failed sw_cmd_sts == item.d_data (4 [0x4] vs 1 [0x1]) reg name: edn_reg_block.sw_cmd_sts
[...]

PR #22353, opened 2024-03-29
[edn] Increase version number, revert verification stage to V1

vogelpi:
Since ES tapeout, a couple of API breaking changes had to be implemented. The DV side was mostly updated in sync but there are two reasons why we cannot meet V2S currently:
- The edn_disable_auto_req_mode test has a pass rate below 90%. See lowRISC/OpenTitan#20818.
- The FSM coverage dropped to around 82%. See lowRISC/OpenTitan#22352.


andreaskurth (approved):
Agreed, thx @vogelpi :+1:

Issue #22354, opened 2024-03-30
[csrng] Investigate and fix branch and expression coverage


vogelpi:
### Description

Most likely due to the area optimization changes for M2, some coverage metrics like branch and expression coverage dropped to below 90%.

We need to investigate what exactly happened and how we can get the metrics back up again.

vogelpi:
This is a duplicate of #19033.

PR #22355, opened 2024-03-30
[csrng] Increase version number, revert verification stage to V1

vogelpi:
For the production tapeout, a couple of API breaking changes are soon being merged which requires a major version bump. Indepentenly, some area optimization changes for M2 negatively impacted the code coverage: branch and expression coverage are now below the 90% bar. Meaning, we should revert the verification stage to V1.

Getting the coverage metrics back up again is tracked in lowRISC/OpenTitan#22354.


andreaskurth (approved):
Agreed, thanks @vogelpi

Issue #20995, opened 2024-01-25
[aon_timer] D2(S) Signoff


msfschaffner:
### Description

Ensure D2(S) signoff criteria are still maintained (this is not a focus area block).

GregAC:
@msfschaffner PTAL

msfschaffner:
Agreed, thanks for the summary @GregAC.

The Hjson still mentions D3 and v1.0.0:
https://github.com/lowRISC/opentitan/blob/7e196e6f072791aa17817b7691aed6b258399eeb/hw/ip/aon_timer/data/aon_timer.hjson#L22-L26

Given the interface change, I believe we should move this to v2.0.0 D2S / V1 instead (potentially V2S if we can sign off on the DV part at the same time).

GregAC:
Good point @msfschaffner I'll make the change. I'll move to V1 as I've actually uncovered some issues in the testbench that need addressing (which I'll open an issue about).

GregAC:
PR here to switch to versions: https://github.com/lowRISC/opentitan/pull/21908

andreaskurth:
Closing as completed in #21908.

PR #22359, opened 2024-04-01
[aon_timer] Sign off at D2S

andreaskurth:
aon_timer only has the standardized `BUS.INTEGRITY` security countermeasure, and it had been signed off at D2S before the bump to version 2.0.0.  Since aon_timer is not a security-focused block and we have not changed its security design since the previous version, we can directly sign it off at D2S again.  See #20995 for the D2 signoff analysis and PR #21908 for the version change, where going to D2S directly has been suggested post merge.


vogelpi (approved):
LGTM!

PR #22361, opened 2024-04-01
[usbdev] pending IN transaction sequence

MubashirSaleem775:
Adds pending IN transaction sequence to validate device functionality

This commit adds a sequence to check the device functionality. It verifies that a SETUP transaction effectively cancels any pending IN transactions by clearing the 'rdy' bit in the configin register of all endpoints.


rswarbrick (approved):

rswarbrick:
CI has failed here, but it seems to be an IT problem in a chip-level test. Definitely not this PR's fault. I'm going to rebase and force-push, then merge.

rswarbrick:
Oh! Please could you force-push after fixing up the commit message to sort out the quick lint check? (It seems that our CI is very European and believes everyone should have the form "firstname lastname")

I don't believe the CI is really correct here, but I see that you are called "Mubashir Saleem" in GitHub, so hopefully your name can match the regex comfortably.

MubashirSaleem775:
@rswarbrick please review.

rswarbrick:
I've just looked and the CI failures are not caused by this PR (one problem in LC_CTRL, and another problem with ROM tests).

I'm going to force-push to this PR with a version that sorts out the merge conflict, then we should be good to merge. Thanks!

PR #22370, opened 2024-04-01
Replace struct/command identifiers with "FourCC" codes.

cfrantz:
Replace struct, command and slot identifiers with [FourCC](https://en.wikipedia.org/wiki/FourCC) codes.

FourCC codes are easily visible in hexdumps and can be printed by our `printf` implementation's custom `%C` specifier.

1. Consolidate all boot slot definitions into `BootSlot`.
2. Use the "FourCC" codes `AA__` and `__BB` to represent the boot slot values.  These values are easily visible in a hexdump, can be printed with our `printf`'s custom `%C` specifier and have a hamming distance of 16 bits.
3. Update the `boot_svc` request/response identifiers to FourCC codes.  The response identifiers are the request identifiers in reverse order (ie: `NEXT` -> `TXEN`).

### DIFF
--- a/sw/device/silicon_creator/lib/boot_log.c
+++ b/sw/device/silicon_creator/lib/boot_log.c
@@ -83,7 +83,7 @@ void boot_log_check_or_init(boot_log_t *boot_log, uint32_t rom_ext_slot,
   boot_log->chip_version.scm_revision_high =
       info->scm_revision.scm_revision_high;
   boot_log->rom_ext_slot = rom_ext_slot;
-  boot_log->bl0_slot = kBootLogUninitialized;
+  boot_log->bl0_slot = 0;  // Unknown: no BL0 slot selected yet.
   for (size_t i = 0; i < ARRAYSIZE(boot_log->reserved); ++i) {
     boot_log->reserved[i] = 0;
   }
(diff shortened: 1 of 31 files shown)


sw/device/silicon_creator/lib/boot_log.c
```diff
@@ -83,7 +83,7 @@ void boot_log_check_or_init(boot_log_t *boot_log, uint32_t rom_ext_slot,
   boot_log->chip_version.scm_revision_high =
       info->scm_revision.scm_revision_high;
   boot_log->rom_ext_slot = rom_ext_slot;
-  boot_log->bl0_slot = kBootLogUninitialized;
+  boot_log->bl0_slot = 0;
```
timothytrippel, line 86: nit:
```suggestion
  boot_log->bl0_slot = 0;  // Uninitialized.
```
?

timothytrippel (approved):

sw/device/silicon_creator/lib/boot_log.c
cfrantz: Fixed.

milesdai (approved):
Thanks for cleaning up all the boot slot identifiers!

moidx (approved):

PR #19171, opened 2023-07-11
[opentitantool] Add RESET gpio to RESET strap for C2D2

jettr:
Flesh out the built-in config for C2D2 debugger to include holding the `RESET` gpio low for the `RESET` strapping option.

PR #22371, opened 2024-04-01
[opentitantool] Improvements to C2D2 debugger support

jesultra:
The C2D2 debugger is used by ChromeOS for communicating with the security chip among other things.  It offers basically only access to the RESET signal and the two UART lines.  This is enough to perform rescue firmware updating, or other operations.

It turns out that the default configuration file for C2D2 had not been updated with proper RESET strapping, and also instead declared empty strappings, which is confusing as it leads to silent errors.

This change properly declares the RESET strapping, and removes the ROM_BOOTSTRAP strapping (as C2D2 does not have wires to enter bootstrap mode.)

### DIFF
--- a/sw/host/opentitanlib/src/app/config/h1dx_devboard_c2d2.json
+++ b/sw/host/opentitanlib/src/app/config/h1dx_devboard_c2d2.json
@@ -12,10 +12,13 @@
   ],
   "strappings": [
     {
-      "name": "RESET"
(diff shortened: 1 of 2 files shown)


jettr (approved):
I wish I would have pushed harder to land #19171 so this functionality would already be there :)

cfrantz (approved):

PR #22376, opened 2024-04-02
[usbdev,dv] usbdev rx crc err seq

Saad-Anees1:
This PR incorporates the rx_crc_err sequence, where we send a packet with a corrupted CRC and subsequently check the CRC status after sending the packet.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_rx_crc_err_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_rx_crc_err_vseq.sv
@@ -1,0 +1,35 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class usbdev_rx_crc_err_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_rx_crc_err_vseq)
+
+  `uvm_object_new
+
+  task body();
+    // Configure transaction
+    configure_out_trans();
+
+    // Enable rx_crc_err interrupt
+    csr_wr(.ptr(ral.intr_enable.rx_crc_err), .value(1'b1));
+
+    // Out Token packet with corrupted CRC5
+    call_token_seq(PidTypeOutToken);
+    // Adding a delay ensures that the CRC5 status is updated and triggers the
+    // interrupt to go high.
+    cfg.clk_rst_vif.wait_clks(4);
+
+    // In this check we are checking the rx crc error interrupt bit because we are sending corrupt
+    // crc and after getting corrupt crc the rx crc err bit should be high.
+    `DV_CHECK_EQ(cfg.intr_vif.sample_pin(IntrRxCrcErr), 1'b1)
+  endtask
+
+  // Override the create_token_pkt function to corrupt the crc5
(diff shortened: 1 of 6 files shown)


Saad22386:
@rswarbrick, kindly review it.

hw/ip/usbdev/dv/env/seq_lib/usbdev_rx_crc_err_vseq.sv
```diff
@@ -0,0 +1,45 @@
...
+    // Configure transaction
+    configure_out_trans();
+
+    // Out Token packet with corrupted CRC5
+    call_token_seq(PidTypeOutToken);
+    // Adding a 1-bit time delay ensures that the CRC5 status is updated following the transaction.
+    cfg.clk_rst_vif.wait_clks(4);
+
+    // Verify the status of CRC5 after sending a token packet with a corrupted CRC5.
+    check_crc5_err();
```
rswarbrick, line 20: Can we make this slightly simpler? I think we really want something like (in C pseudo-code)
```
seen_crc_err = 0;
for (i = 0; i < 10; i++) {
  if (rx_crc_err) {
    seen_crc_err = 1;
    break;
  }
}
if (!seen_crc_err) die("oh no...");
```
Doing it with this structure avoids us encoding the exact timing into the test.

hw/ip/usbdev/dv/env/seq_lib/usbdev_rx_crc_err_vseq.sv
```diff
@@ -0,0 +1,45 @@
...
+  endtask
+
+  // Override the call_token_seq task to corrupt the crc5
+  virtual task call_token_seq(input pid_type_e pid_type);
```
rswarbrick, line 24: I don't think that we should override a base task like this: you've ended up having to copy-paste the contents and then change a bit. Aargh!

I'd suggest a refactor: Change the base class so that it looks something like this (untested!):
```systemverilog
virtual task call_token_seq(input pid_type_e pid_type);
  create_token_pkt();
  m_usb20_item = m_token_pkt;
  start_item(m_token_pkt);
  finish_item(m_token_pkt);
endtask

virtual task create_token_pkt();
  `uvm_create_on(m_token_pkt, p_sequencer.usb20_sequencer_h)
  m_token_pkt.m_pkt_type = PktTypeToken;
  m_token_pkt.m_pid_type = pid_type;
  assert(m_token_pkt.randomize() with {m_token_pkt.address inside {7'b0};
                                       m_token_pkt.endpoint == endp;});
endtask
```

Now the derived class can just do this:
```systemverilog
virtual task create_token_pkt();
  super.create_token_pkt();
  // Flip the crc5 bits to corrupt the crc5
  m_token_pkt.crc5 = ~m_token_pkt.crc5;
endtask
```

rswarbrick (changes requested):

hw/ip/usbdev/dv/env/seq_lib/usbdev_rx_crc_err_vseq.sv
Saad22386: I am not getting your point here.
Are you saying to replace check_crc5_err() task with your suggested approach?
Also, why are we using a loop to iterate 10 times to check for a CRC error?
Isn't it enough to check the error one time after sending the packet?

alees24:

PR #22379, opened 2024-04-02
[usbdev] usbdev_disconnected seqeunce

MubashirSaleem775:
Adds 'usbdev_disconnected' sequence to validate device functionality upon disconnection. Upon loss of usb_vbus, the device generates a disconnected interrupt.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
@@ -1,0 +1,34 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class usbdev_disconnected_vseq extends usbdev_base_vseq;
+  `uvm_object_utils(usbdev_disconnected_vseq)
+
+  `uvm_object_new
+
+  task body();
+    uvm_reg_data_t rd_data;
+    bit disconnected;
+    // Enable disconnected interrupt
+    csr_wr(.ptr(ral.intr_enable.disconnected),  .value(1'b1));
+    // Clear disconnected interrupt to make sure its on reset state.
+    csr_wr(.ptr(ral.intr_state.disconnected),  .value(1'b1));
+    // send pkt to device
+    configure_out_trans();
+    call_sof_seq(PidTypeSofToken);
+    @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+    // Verify that the device has asserted the usb_ref_val_o signal after receiving the SoF packet.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o, 1)
+    // Check the disconnected interrupt isn't yet asserted
+    `DV_CHECK_EQ(cfg.intr_vif.pins[IntrDisconnected], 0)
+    // Drive 0 on usb_vbus signal to tell device that its has been disconnected.
+    cfg.m_usb20_agent_cfg.bif.drive_vbus = 1'b0;
+    // Expect the disconnected interrupt within a short time
+    for (int i = 0; i < 10; i++) begin
+      @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+      if (cfg.intr_vif.pins[IntrDisconnected]) break;
+    end
+    `DV_CHECK_EQ(cfg.intr_vif.pins[IntrDisconnected], 1)
+  endtask
+endclass
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -347,7 +347,7 @@
             - Verify that the corresponding Interrupt pin goes high after the link disconnected.
             '''
       stage: V2
-      tests: []
+      tests: ["usbdev_disconnected"]
     }
     {
       name: host_lost
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_vseq_list.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_vseq_list.sv
@@ -9,6 +9,7 @@
 
 `include "usbdev_av_buffer_vseq.sv"
 `include "usbdev_csr_test_vseq.sv"
+`include "usbdev_disconnected_vseq.sv"
 `include "usbdev_dpi_config_host_vseq.sv"
 `include "usbdev_enable_vseq.sv"
 `include "usbdev_fifo_rst_vseq.sv"
--- a/hw/ip/usbdev/dv/env/usbdev_env.core
+++ b/hw/ip/usbdev/dv/env/usbdev_env.core
@@ -31,6 +31,7 @@ filesets:
       - seq_lib/usbdev_smoke_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_csr_test_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_pkt_received_vseq.sv: {is_include_file: true}
+      - seq_lib/usbdev_disconnected_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_av_buffer_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_setup_trans_ignored_vseq.sv: {is_include_file: true}
       - seq_lib/usbdev_pkt_buffer_vseq.sv: {is_include_file: true}
--- a/hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
+++ b/hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
@@ -81,6 +81,10 @@
       uvm_test_seq: usbdev_min_length_out_transaction_vseq
     }
     {
+      name: usbdev_disconnected
+      uvm_test_seq: usbdev_disconnected_vseq
+    }
+    {
       name: usbdev_nak_trans
       uvm_test_seq: usbdev_nak_trans_vseq
     }


hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,37 @@
...
+    csr_wr(.ptr(ral.intr_state.disconnected),  .value(1'b1));
+    // send pkt to device
+    configure_out_trans();
+    call_sof_seq(PidTypeSofToken);
+    inter_packet_delay();
+    // Verify that the device has asserted the usb_ref_val_o signal after receiving the SoF packet.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o, 1)
+    // Read intr_state register and veify that disconncted interrupt is 0.
+    csr_rd(ral.intr_state, .value(rd_data));
+    `DV_CHECK_EQ(get_field_val(ral.intr_state.disconnected, rd_data), 0)
```
rswarbrick, line 25: Instead of using CSRs for this, please use the interrupt pin itself (visible through `intr_vif`). Similarly for the check in the loop below.

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,37 @@
...
+    uvm_reg_data_t rd_data;
+    bit disconnected;
+    // Enable disconnected interrupt
+    csr_wr(.ptr(ral.intr_enable.disconnected),  .value(1'b1));
+    // Clear disconnected interrupt to make sure its on reset state.
+    csr_wr(.ptr(ral.intr_state.disconnected),  .value(1'b1));
+    // send pkt to device
+    configure_out_trans();
+    call_sof_seq(PidTypeSofToken);
+    inter_packet_delay();
```
rswarbrick, line 20: Why is this delay required?

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,37 @@
...
+
+  `uvm_object_new
+
+  task body();
+    uvm_reg_data_t rd_data;
+    bit disconnected;
+    // Enable disconnected interrupt
+    csr_wr(.ptr(ral.intr_enable.disconnected),  .value(1'b1));
+    // Clear disconnected interrupt to make sure its on reset state.
+    csr_wr(.ptr(ral.intr_state.disconnected),  .value(1'b1));
```
rswarbrick, line 16: You probably don't actually need to write the `intr_state` here, but it makes sense to check that the interrupt is false at this point. As I mentioned later on, you can look directly at the interrupt line through `intr_vif`.

rswarbrick (changes requested):

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
MubashirSaleem775: We need to introduce this delay because we're awaiting the assertion of the usb_ref_val_o signal, which occurs after a certain delay from the clock signal.

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
MubashirSaleem775: Okay , I will read from intr_vif.

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
MubashirSaleem775: Okay

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,36 @@
...
+    // send pkt to device
+    configure_out_trans();
+    call_sof_seq(PidTypeSofToken);
+    @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+    // Verify that the device has asserted the usb_ref_val_o signal after receiving the SoF packet.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o, 1)
+    // Read intr_state register and veify that disconncted interrupt is 0.
+    `DV_CHECK_EQ(cfg.intr_vif.sample_pin(IntrDisconnected), 1'b0)
+    // Disconnect the device vbus
+    cfg.m_usb20_agent_cfg.bif.usb_vbus = 1'b0;
```
rswarbrick, line 26: This is going to need tweaking after recent changes to `usb20_block_if.sv`. I think this will work:
```systemverilog
    cfg.m_usb20_agent_cfg.bif.drive_vbus = 1'b0;
```

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,36 @@
...
+    csr_wr(.ptr(ral.intr_state.disconnected),  .value(1'b1));
+    // send pkt to device
+    configure_out_trans();
+    call_sof_seq(PidTypeSofToken);
+    @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+    // Verify that the device has asserted the usb_ref_val_o signal after receiving the SoF packet.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o, 1)
+    // Read intr_state register and veify that disconncted interrupt is 0.
+    `DV_CHECK_EQ(cfg.intr_vif.sample_pin(IntrDisconnected), 1'b0)
+    // Disconnect the device vbus
```
rswarbrick, line 25: Can you rephrase this? I think we're driving 0 on VBUS in order to tell the device that it has been disconnected. (We're not disconnecting the vbus signal...)

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,36 @@
...
+    cfg.m_usb20_agent_cfg.bif.usb_vbus = 1'b0;
+    // delay required so that device monitored vbus disconnected and assert disconnected interrupt.
+    for (int i = 0; i < 10; i++) begin
+      @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+      if (cfg.intr_vif.sample_pin(IntrDisconnected)) break;
+    end
+    if (!cfg.intr_vif.sample_pin(IntrDisconnected)) begin
+      `uvm_error(`gfn, "usbdev didn't generate disconnected interrupt")
+    end
```
rswarbrick, line 34: I think this can be slightly simpler:
```systemverilog
    // Expect the disconnected interrupt within a short time
      @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
      if (cfg.intr_vif.pins[IntrDisconnected]) break;
    end
    `DV_CHECK_EQ(cfg.intr_vif.pins[IntrDisconnected], 1)
```

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,36 @@
...
+    csr_wr(.ptr(ral.intr_enable.disconnected),  .value(1'b1));
+    // Clear disconnected interrupt to make sure its on reset state.
+    csr_wr(.ptr(ral.intr_state.disconnected),  .value(1'b1));
+    // send pkt to device
+    configure_out_trans();
+    call_sof_seq(PidTypeSofToken);
+    @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+    // Verify that the device has asserted the usb_ref_val_o signal after receiving the SoF packet.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o, 1)
+    // Read intr_state register and veify that disconncted interrupt is 0.
```
rswarbrick, line 23: Can you tweak this text, now that we're looking directly at the pin? I think something like this would work:
```
    // Check the disconnected interrupt isn't yet asserted
    `DV_CHECK_EQ(cfg.intr_vif.pins[IntrDisconnected], 1)
```
(I've also switched to looking at `pins[..]` directly: I don't think there's any need for explicitly calling the `sample_pin` function here)

Also nit: Missing "e" in "disconnected"

rswarbrick (changes requested):

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
MubashirSaleem775: Yes, i will update

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
MubashirSaleem775: Yes

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
MubashirSaleem775: Okay , i will change this as you suggested. thanks

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
rswarbrick: I agree that there's a slight delay. Here's a screenshot of the waves from a run:
image
The red line is the time where `call_sof_seq` completes.

I've taken a proper look at the signals. The required wait is a single cycle of the main clock. Can we switch over to the correct clock, then I think it makes sense.

rswarbrick: Er. I take that back! It's 2 cycles of the main clock :-) I've just taken a more careful look and this timing comes from the "packet end" signal coming out of `usb_fs_rx.sv`, which waits to see an EOP signal. It's a bit sad to have to snoop on the USB clock itself (which we're currently doing), but I guess it's a simple way to keep everything in sync for now.

hw/ip/usbdev/dv/env/seq_lib/usbdev_disconnected_vseq.sv
```diff
@@ -0,0 +1,34 @@
...
+    // Clear disconnected interrupt to make sure its on reset state.
+    csr_wr(.ptr(ral.intr_state.disconnected),  .value(1'b1));
+    // send pkt to device
+    configure_out_trans();
+    call_sof_seq(PidTypeSofToken);
+    @(posedge cfg.m_usb20_agent_cfg.bif.usb_clk);
+    // Verify that the device has asserted the usb_ref_val_o signal after receiving the SoF packet.
+    `DV_CHECK_EQ(cfg.m_usb20_agent_cfg.bif.usb_ref_val_o, 1)
+    // Check the disconnected interrupt isn't yet asserted
+    `DV_CHECK_EQ(cfg.intr_vif.pins[IntrDisconnected], 1)
```
rswarbrick, line 24: Ah, this was my fault: I had a typo in my initial suggestion! I think this probably needs to be `0` (otherwise the test will fail :-D)

alees24:

PR #22380, opened 2024-04-02
[dvsim] General clean-up in modes.py

rswarbrick:
I spent some time looking carefully at this code back in December and it's much more complicated than we need it to be. The commits in this PR are simplifications. There are some other tidy-ups that they enable, but the PR was turning into a bit of a mammoth and I thought this was a sensible place to start!

Most of the commits here are generic tidyups. The only significant change is probably in the code that cleans up how we print `Modes` objects (see the "Strengthen typing ..." commit). This removes use of some generic "object to string" code which the following commit can delete.

### DIFF
--- a/util/dvsim/Modes.py
+++ b/util/dvsim/Modes.py
@@ -5,6 +5,7 @@
 from copy import deepcopy
 import logging as log
 import sys
+from typing import List, Optional
 
 from utils import VERBOSE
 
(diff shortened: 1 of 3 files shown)


util/dvsim/Modes.py
hcallahan-lowrisc, line 1: This is much easier to understand! Thanks

hcallahan-lowrisc (approved):
LGTM, thanks for the clean ups!

PR #22383, opened 2024-04-02
[dvsim] Further tidyup in `modes.py`

rswarbrick:
Lots of these changes are "generic tidyups". The one that's a bit awkward is the 4th commit ("Treat build_modes as a dictionary..."). This change avoids a rather messy situation where lists were getting merged (and I thought it might have been taking lots of memory).

Unfortunately, we vendor some DV code from Ibex and that DV code uses dvsim vendored from OpenTitan, giving a circular dependency. The commit "fixes" things by applying a patch to the vendored Ibex hjson file, ensuring that everything stays consistent at this end.


util/dvsim/Modes.py
hcallahan-lowrisc, line 1: Looks like `util/dvsim/BUILD` needs to be updated as well

rswarbrick: Bazel is still newfangled in my book :-) I've just done the obvious change, which should hopefully sort it out.

Issue #22322, opened 2024-03-28
[crypto] Outstanding test failures in ECDSA and RSA


RyanTorok:
### Description

Several outstanding test failures exist for RSA and ECDSA in the cryptotest framework.

RSA:

- [ ] `wycheproof RSA-2048 pss verify sha-384 #64`
- [ ] `wycheproof RSA-3072 pss verify sha-256 #94` 
- [ ] `wycheproof RSA-4096 pss verify sha-256 #31`
- [ ] `wycheproof RSA-4096 pss verify sha-384 #9`
- [ ] `wycheproof RSA-4096 pss verify sha-384 #33`
- [ ] `wycheproof RSA-4096 pss verify sha-512 #164`
- [ ] `NIST CAVP RSA-3072 Verify PSS SHA-{256, 384, 512}` unconditionally returns "fail"

ECDSA: 

- [ ] `wycheproof p256 verify sha-256 #350` 
- [ ] `wycheproof p256 verify sha-512 #420` 
- [ ] `wycheproof p256 verify sha3-256 #358` 
- [ ] `wycheproof p256 verify sha3-512 #424`

vsukhoml:
I requested NIST ACVP tests for many of the combinations and got sample test vectors. I can add those in addition to wycheproof if this helps.

RyanTorok:
ECDSA test failures resolved by #22393

moidx:
Removing the `ROM` label as we are no longer using RSA for secure boot.

johannheyszl:
@andrea-caforio to PTAL

nasahlpa:
Closing as the ECDSA tests are passing and the remaining failing RSA verify tests are documented in https://github.com/lowRISC/opentitan/issues/28656.

PR #22393, opened 2024-04-02
[crypto] Fix a bug in ECDSA-P256.

jadephilipoom:
Fixes the failing wycheproof ECDSA tests in #22322.

When computing the recovered x coordinate, we get a value that's reduced modulo p (the coordinate field modulus) and we then need to reduce this value modulo n (the curve order). This operation is almost always a no-op, since p and n are very close together; there's roughly a 2^-130 probability of hitting this range by chance. However, the code implementing the reduction was *always* a no-op. Essentially, by using `bn.subm` instead of `bn.addm` here, the code was conditionally *adding* the modulus n if `w19 - 0` underflowed, which it never would. Instead, we need `bn.addm`, which will conditionally *subtract* the modulus n if `w19 + 0` is greater than n.

### DIFF
--- a/sw/otbn/crypto/p256_verify.s
+++ b/sw/otbn/crypto/p256_verify.s
@@ -244,7 +244,7 @@ p256_verify:
   la        x3, p256_n
   bn.lid    x0, 0(x3)
   bn.wsrw   MOD, w0
-  bn.subm   w24, w19, w31
+  bn.addm   w24, w19, w31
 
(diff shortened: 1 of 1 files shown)


RyanTorok (approved):
LGTM, thanks Jade!

wettermo (approved):
LGTM!

moidx (approved):

moidx:
Thanks @jadephilipoom! Please ignore the Kokoro failure.

PR #22395, opened 2024-04-02
[ci] Set up a nightly tag and use it for ROM crypto tests.

jadephilipoom:
Draft while I check if this has the expected effect on CI.


jadephilipoom:
Comparing to [1e9b334f](https://github.com/lowRISC/opentitan/commit/1e9b334ff19ea817c8e5267975316198b066ea6b), which is the last successful run on `master` that shares history with this branch, the `CW310 Test ROM Tests` job took 11m20s instead of 22m25s. Might be helpful with the FPGA contention, although it wasn't as much speedup as I was hoping for.

For full disclosure, the `CW310 ROM Tests` job took a few minutes longer on this PR than on the master run, but when I looked at the build log I think it was just on a slightly slower runner (they ran the same number of tests, but this PR had to re-run a couple flaky tests that timed out 1/2 times, and seemed to take very slightly longer on all the tests). If anyone has a better way of confirming whether I accidentally affected the ROM job somehow, I'm all ears!

Also, I don't think I have permissions to trigger a manual nightly run -- we should definitely do that before merging this change to make sure the changes didn't break it, :warning:

ci/azure-pipelines-nightly.yml
```diff
@@ -108,8 +108,8 @@ jobs:
     displayName: "Run all ROM E2E tests"
   - template: ../ci/publish-bazel-test-results.yml
 
-- job: slow_otbn_crypto_tests
-  displayName: "Slow OTBN Crypto Tests"
+- job: nightly_tagged_sw_tests
+  displayName: "Software tests explicitly tagged as 'nightly'"
```
rswarbrick: I'm not sure about this name change. It's a job in the nightly pipeline, so it's kind of "nightly" already! I think the tag on the test means "Only run this test in the nightly run" and the name here means "The stuff that wouldn't be run otherwise".

I don't have strong feelings, but I'm not sure this name change is needed.

ci/azure-pipelines-nightly.yml
```diff
@@ -108,8 +108,8 @@ jobs:
     displayName: "Run all ROM E2E tests"
   - template: ../ci/publish-bazel-test-results.yml
 
-- job: nightly_tagged_sw_tests
-  displayName: "Software tests explicitly tagged as 'nightly'"
+- job: nightly_tagged_tests
```
rswarbrick, line 111: I think this second commit has two parts:
- Restrict the nightly pipeline less, so that it's not just running software tests
- Avoid running nightly tests in `run-fpga-tests.sh`

Do those changes need to be in the same commit? If not, I'd be tempted to split the commit in two.

sw/device/silicon_creator/lib/sigverify/BUILD
```diff
@@ -110,6 +110,10 @@ opentitan_test(
     srcs = ["mod_exp_ibex_functest.c"],
     cw310 = cw310_params(
         timeout = "long",
+        tags = [
+            # The wycheproof suite has a lot of tests; run it only on the nightly CI.
+            "nightly",
```
rswarbrick, line 115: I wonder whether it would make sense to change this so that these tests were tagged "whycheproof" and then the "only run whycheproof tests in nightly regressions" decision would only need to be done and documented in one place.

I'm not a Bazel expert though: this might be really silly!

(Same question below for the known answer tests)

sw/device/tests/BUILD
```diff
@@ -181,7 +181,11 @@ opentitan_test(
 opentitan_test(
     name = "alert_handler_lpg_clkoff_test",
     srcs = ["alert_handler_lpg_clkoff_test.c"],
-    cw310 = cw310_params(timeout = "moderate"),
+    cw310 = cw310_params(
+        timeout = "moderate",
+        # This test is slow; run it only on the nightly CI.
```
rswarbrick, line 186: Nit: I'm not sure that "nightly CI" makes sense. Maybe "nightly regression" or something work work better?

PR #22397, opened 2024-04-02
[bazel,autogen_hjson] Split C and rust header generation

matutem:
Create separate rules to generate the C and rust register collateral for each IP. This is cleaner, and enables generating rust collateral only as required by tock.


moidx (approved):
Thanks @matutem!

PR #22403, opened 2024-04-03
[prim,dv] Tweak ASSERT_FINAL to be a no-op if FPV enabled

rswarbrick:
Jasper just emits a warning if it sees a final block (and ignores the contents). Don't emit the final block if we're targeting Jasper, which avoids the warning.


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #22404, opened 2024-04-03
[prim,tlul,rtl] Explicitly cast a "1" to specific number of bits

rswarbrick:
This avoids a warning from Jasper (and, I suspect, some lint tools). The problem is that "1" is a 32-bit value, so "1 << xyz" is also a 32-bit value and the tool warns when we try to put it in something with fewer bits bits.

Explicitly making a "one" that's the right width seems to avoid the warnings and is obviously equivalent.


hcallahan-lowrisc (approved):
lgtm

PR #22406, opened 2024-04-03
Sival update testplans

engdoreis:
This PR adds bazel targets to the testplans which are already running on the nightlies but not linked.


moidx (approved):

engdoreis:
I'm merging this PR because the CI failure is not related, the failed test also fails on the earlgrey_es_sival branch.

PR #22408, opened 2024-04-03
[reggen,fpv] Fix up some bugs in indices in reggen's FPV code

rswarbrick:
These were found by looking at Verisium warnings that come out of running a Jasper proof. The proof still fails with the same percentages, but these changes silence some warnings and (I think) fix some genuine bugs.

### DIFF
--- a/util/reggen/fpv_csr.sv.tpl
+++ b/util/reggen/fpv_csr.sv.tpl
@@ -204,8 +204,11 @@ module ${mod_base}_csr_assert_fpv import tlul_pkg::*;
 
   `ASSERT(TlulOOBAddrErr_A, oob_addr_err |-> s_eventually(d2h.d_valid && d2h.d_error))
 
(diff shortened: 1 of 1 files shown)


util/reggen/fpv_csr.sv.tpl
```diff
@@ -110,21 +137,22 @@ module ${mod_base}_csr_assert_fpv import tlul_pkg::*;
 
 % if num_hro_regs > 0:
   // Assign regwen to registers. If the register does not have regwen, it will default to value 1.
-  logic [${num_hro_regs}-1:0] regwen;
+  logic [${num_hro_regs}:0] regwen;
```
vogelpi, line 140: Based on the comment above, I guess this must have been one of the bugs :-)

vogelpi (approved):
Thanks @rswarbrick  for the PR and your explanatory comments. I am no formal expert but this all looks good and very reasonable.

PR #22409, opened 2024-04-03
[host,usbdev] Host-side component of usbdev streaming tests

alees24:
The existing streaming test code requires a host-side component to verify communications with the USB device in FPGA or ES when connected to a physical host. This PR includes the host-side software that cooperates with the following device-side test software already in the repository:

 - sw/device/tests/**usbdev_stream_test**
 - sw/device/tests/**usbdev_iso_test**
 - sw/device/tests/**usbdev_mixed_test**

The code has been used extensively on different hosts and with a number of different FPGA boards (CW310, CW340, Sonata, ArtyA7) over a period of many months, and is required for SiVal testing of the USB device.

It's perhaps also worth mentioning again that the three tests above are running in regressions with the `usbdpi` model implementing the same functionality and acting as a virtual host.

The PR is formed of 3 commits:

1. Initial source files to build either of the following two variants:
  - a libusb-dependent Linux console application called `stream_test` which locates the USB device automatically using Vendor/Product ID and then reads a test descriptor from the device-side software in order to establish the required test behavior/transfer types etc.
  - a standalone Linux console application that is not dependent upon libusb and that may be used in conjunction with just usbdev_stream_test using direct serial port access to exercise Bulk Transfers. This may be used on other target platforms where libusb is not presently available/supported.
2. This small commit adds support for Isochronous Transfers and is required for either 'usbdev_iso_test' or 'usbdev_mixed_test' to operate correctly.
3. **Temporary** example build files for testing. **Not intended for inclusion.** (See below*) These may be used to build the host-side application 'stream_test' (libusb-dependent) or the reduced application  'serial_test'  that uses just serial port file descriptors.
  - This commit also includes some interim modifications of the BUILD file as a sketch of what is required for regression runs/SiVal testing, but they are not yet operational.

The goal:

In order for the above 3 device-side tests to be integrated into regression runs and SiVal testing, as I understand it we require opentitantool/test_harness to kick off the automatically-built 'stream_test' application. The host-side software will wait until the USB device is connected and configured by the device-side software, and requires no command-line parameters for its default behavior, since it responds to the device-side software.

Upon test completion/failure, the device-side software signals in the normal manner to opentitantool/lib, at which point the test harness may need to terminate the `stream_test` process, because in the case of both 'usbdev_iso_test' and 'usbdev_mixed_test' the host-side component has no awareness of test completion and will necessarily keep running. These two tests employ Isochronous Transfers for which packets may be dropped at any point, so only the device-side test software know when it has received and checked the amount of data required for successful test completion. When `usbdev_stream_test` is run, the host-side code should terminate automatically because that employs only reliable Bulk Transfers with retrying where necessary.

** Help would be appreciated from those who better understand our build system and host-side execution environment, please.

A README.md file has been included to try to explain the behavior and structure of what has necessarily become a fairly extensive bit of test software over the course of its development.

### DIFF
--- a/sw/host/tests/usbdev/usbdev_stream/BUILD
+++ b/sw/host/tests/usbdev/usbdev_stream/BUILD
@@ -1,0 +1,41 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "usbdev_stream",
+    srcs = [
+        "stream_test.cc",
+        "usb_device.cc",
+        "usbdev_int.cc",
+        "usbdev_iso.cc",
+        "usbdev_serial.cc",
+        "usbdev_stream.cc",
+        "usbdev_utils.cc",
+    ],
+    hdrs = [
+        "stream_test.h",
+        "usb_device.h",
+        "usbdev_int.h",
+        "usbdev_iso.h",
+        "usbdev_serial.h",
+        "usbdev_stream.h",
+        "usbdev_utils.h",
+    ],
+    # Not a local define since targets that depend on this need to use
+    # the same configuration.
+    defines = [
+        "STREAMTEST_LIBUSB=1",
+    ],
+    linkopts = ["-lusb-1.0"],
+)
+
+cc_binary(
+    name = "stream_test",
+    srcs = ["stream_test.cc"],
+    deps = [
+        ":usbdev_stream",
+    ],
+)
--- a/sw/host/tests/usbdev/usbdev_stream/build.sh
+++ b/sw/host/tests/usbdev/usbdev_stream/build.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+g++ -Wall -Werror -std=c++14 -c -o stream_test.o -DSTREAMTEST_LIBUSB=1 stream_test.cc
+g++ -Wall -Werror -std=c++14 -c -o usbdev_iso.o -DSTREAMTEST_LIBUSB=1 usbdev_iso.cc
+g++ -Wall -Werror -std=c++14 -c -o usbdev_int.o -DSTREAMTEST_LIBUSB=1 usbdev_int.cc
+g++ -Wall -Werror -std=c++14 -c -o usbdev_serial.o -DSTREAMTEST_LIBUSB=1 usbdev_serial.cc
+g++ -Wall -Werror -std=c++14 -c -o usbdev_stream.o -DSTREAMTEST_LIBUSB=1 usbdev_stream.cc
+g++ -Wall -Werror -std=c++14 -c -o usbdev_utils.o -DSTREAMTEST_LIBUSB=1 usbdev_utils.cc
+g++ -Wall -Werror -std=c++14 -c -o usb_device.o -DSTREAMTEST_LIBUSB=1 usb_device.cc
+
+g++ -g -O2 -o stream_test stream_test.o usbdev_iso.o usbdev_int.o usbdev_serial.o usbdev_stream.o usbdev_utils.o usb_device.o -lusb-1.0
--- a/sw/host/tests/usbdev/usbdev_stream/build_standalone.sh
+++ b/sw/host/tests/usbdev/usbdev_stream/build_standalone.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+g++ -std=c++14 -Wall -Werror -g -O2 -o serial_test stream_test.cc usbdev_serial.cc usbdev_stream.cc usbdev_utils.cc usb_device.cc


a-will:
Hm... I'm not exactly sure how to approach this one.

Our test environment is built around opentitanlib being the test_harness that orchestrates everything. There could be a Rust+opentitanlib binary that mostly calls into a C library via FFI to reuse most of the stimulus generation and checking of output, and that would get you support for all the "standard arguments" and non-USB interfaces and functions, like programming a bitstream, bootstrapping the binary, etc.

Or as you said, what might work is even a Rust binary that just forks off a new process that executes stream_test and monitors stdout and/or signals. The mechanisms would probably look similar to how we fork off the simulator built atop the Verilator output, with the stream_test arguments passing through the Rust binary's arguments. Then you can use the usual framework we have to set up the test.

pamaury:
I have already anticipated this and created a usb test harness: https://github.com/lowRISC/opentitan/blob/master/sw/host/tests/chip/usb/usb.rs
The idea is to extend it to take a binary as argument (the C host component) and spawn it after USB detection.
I will create a PR to add support for that.

alees24:
> I have already anticipated this and created a usb test harness: https://github.com/lowRISC/opentitan/blob/master/sw/host/tests/chip/usb/usb.rs The idea is to extend it to take a binary as argument (the C host component) and spawn it after USB detection. I will create a PR to add support for that.

Yes, I found that after I'd temporarily extended the usbdev_aon_wake code on master to support the streaming tests! It worked well enough to do initial testing on ES. Thanks for the usb_harness work; I have switched over to using that locally.

sw/host/tests/usbdev/usbdev_stream/README.md
```diff
@@ -0,0 +1,174 @@
+# Host-side streaming software
+
+To verify the USB device (USBDEV) with a physical USB host requires special-purpose software to
+communicate with the device, checking all of the data transferred. This `stream_test` program forms
+the host-side component of the verification effort, complementing the following device-side tests:
+
+- usbdev_stream_test
+  - Device-side software that streams Bulk Transfer data to/from multiple endpoints, generating the
+    source data using a LFSR, and then checking the returned data that is received from the
```
pamaury, line 9: missing end of sentence

sw/host/tests/usbdev/usbdev_stream/usbdev_utils.cc
```diff
@@ -0,0 +1,169 @@
...
+    ReportSyntax();
+    return -1;
+  }
+
+  // We need to ensure that we can send full 8-bit binary data with no character
+  // translations and no character echo etc.
+  struct termios tty;
+  if (tcgetattr(fd, &tty) != 0) {
+    report_error("Failed getting terminal attributes");
+    return -1;
```
pamaury, line 41: file descriptor is not closed

sw/host/tests/usbdev/usbdev_stream/usbdev_utils.cc
```diff
@@ -0,0 +1,169 @@
...
+
+  // Set in/out baud rate to be as high as possible; just in case, but it has
+  // no impact upon the measured transfer speed.
+  cfsetispeed(&tty, B4000000);
+  cfsetospeed(&tty, B4000000);
+
+  // Save tty settings, also checking for error.
+  if (tcsetattr(fd, TCSANOW, &tty) != 0) {
+    report_error("Failed setting terminal attributes");
+    return -1;
```
pamaury, line 70: file descriptor is not closed

sw/host/tests/usbdev/usbdev_stream/stream_test.h
```diff
@@ -0,0 +1,79 @@
...
+#else
+#define STREAMS_MAX 11U
+#endif
+
+/**
+ * Simple container for test configuration.
+ */
+struct TestConfig {
+  /**
+   * Initialize test settings.
```
pamaury, line 21: Maybe document arguments?

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,574 @@
...
+// Note that the mapping from device endpoints to USB port number is not
+// guaranteed, and  when multiple streams are used, it is _not_ necessarily the
+// case that ascending streams/endpoints in usbdev_stream_test are mapped to
+// a contiguous range of ascending ttyUSBi port names.
+//
+// Either or both of the initial input port and the initial output port may be
+// overridden using command line parameters.
+//
+// Usage:
+//   stream [-v<bool>][-c<bool>][-r<bool>][-s<bool>]
```
pamaury, line 28: The usage string does not contains several flags described below (`-z`, `-t`, `-v`)

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,574 @@
...
+//          [[-d<bus>:<address>] | [--device <bus>:<address>]]
+//          [<input port>[ <output port>]]
+//
```
pamaury, line 31: Specify what is a `<bool>`, I think you accept `0`, `1`, `y` and `n`.

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,574 @@
...
+      *pp = p;
+      return true;
+    }
+  }
+  return false;
+}
+
+// Parse a command line option specifying the bus number and device address.
+bool GetDevice(const char *p, uint8_t &busNumber, uint8_t &devAddress) {
+  return GetByte(&p, busNumber) && (*p++ == ':') && GetByte(&p, devAddress);
```
pamaury, line 130: You don't check that `p` is the end of the string

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,574 @@
...
+//
+// Note that the mapping from device endpoints to USB port number is not
+// guaranteed, and  when multiple streams are used, it is _not_ necessarily the
+// case that ascending streams/endpoints in usbdev_stream_test are mapped to
+// a contiguous range of ascending ttyUSBi port names.
+//
+// Either or both of the initial input port and the initial output port may be
+// overridden using command line parameters.
+//
+// Usage:
```
pamaury, line 27: I don't know if it is necessary to repeat the usage here given that there is the usage below in the code

sw/host/tests/usbdev/usbdev_stream/usb_device.h
```diff
@@ -0,0 +1,354 @@
...
+   * Return the test number from the test descriptor.
+   *
+   * @return test number
+   */
+  uint8_t TestNumber() const { return testNumber_; }
+  /**
+   * Return the specified test argument from the test descriptor,
+   *
+   * @param  arg     Argument number.
+   * @return test argument
```
pamaury, line 263: Maybe specify somewhere, preferably in the form of a named constant, that there are 4 arguments and explain what they are?

sw/host/tests/usbdev/usbdev_stream/README.md
alees24: Fair point, I'm good at that. Thanks.

sw/host/tests/usbdev/usbdev_stream/stream_test.h
alees24: Added a comment; should suffice since it's only a simple aggregation of the various configuration settings.

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
alees24: Okay.

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
alees24: I think I'll leave it. It was more useful before, granted, because it started as a single-file program. I like being able to look at the top of the source to a simple program and see the command line used to invoke/build it. Of course, over a period of some months of verification effort it has become quite a bit more involved.

sw/host/tests/usbdev/usbdev_stream/usb_device.h
alees24: I've added a comment. A bit like 'argc/argv', though, it's not really specified and could potentially be test number-dependent. Presently the set of tests remains fairly small.

alees24:
Thanks for the review and the oversights that you caught. I think I've addressed them all.

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,580 @@
...
+              << std::endl;
+
+    bool opened(false);
+#if STREAMTEST_LIBUSB
+    bool bulk(true);
+#endif
+    switch (streamType) {
+      case USBDevStream::StreamType_Serial: {
+        USBDevSerial *s;
+        s = new USBDevSerial(idx, transfer_bytes, cfg.retrieve, cfg.check,
```
pamaury, line 273: Should an error be reported if s is null?

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,580 @@
...
+                             cfg.send, cfg.verbose);
+        if (s) {
+          opened = s->Open(in_port, out_port);
+          if (opened) {
```
pamaury, line 277: Here as well

pamaury: Ah I see that error handling is below, ignore this comment

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,580 @@
...
+            // Notify all of the streams that no more traffic shall be
+            // initiated.
+            for (unsigned idx = 0U; idx < nstreams; idx++) {
+              streams[idx]->Resume();
+            }
+            std::cout << "Resuming streaming..." << std::endl;
+            // Start of Running interval.
+            start_time = time_us();
+          }
+        }
```
pamaury, line 406: do we ever go to `USBDevice::StateSuspending` ?

sw/host/tests/usbdev/usbdev_stream/stream_test.cc
```diff
@@ -0,0 +1,580 @@
...
+          // Start of Suspended interval.
+          start_time = time_us();
+          std::cout << "Suspended" << std::endl;
+        }
+        break;
+
+      case USBDevice::StateSuspended:
+        if (elapsed_time(start_time) >= kSuspendedInterval) {
+          dev->Resume();
+          dev->SetState(USBDevice::StateResuming);
```
pamaury, line 422: I think `Resume()` already sets the state to `StateResuming`.

sw/host/tests/usbdev/usbdev_stream/usb_device.cc
```diff
@@ -0,0 +1,361 @@
...
+  std::cout << "Reading Test Descriptor" << std::endl;
+
+  if (!Open()) {
+    return false;
+  }
+#if STREAMTEST_LIBUSB
+  // Send a Vendor-Specific command to read the test descriptor.
+  uint8_t testDesc[0x10u];
+  int rc =
+      libusb_control_transfer(devh_, 0xc2u, kVendorTestConfig, 0u, 0u, testDesc,
```
pamaury, line 227: Maybe replace `0xc2` by suitable constants from libusb?

sw/host/tests/usbdev/usbdev_stream/usb_device.cc
```diff
@@ -0,0 +1,361 @@
...
+    for (unsigned idx = 0u; idx < sizeof(testDesc); idx++) {
+      printf("%u: 0x%02x\n", idx, testDesc[idx]);
+    }
+  }
+
+  // Validate the received test descriptor.
+  const uint8_t test_sig_head[] = {0x7eu, 0x57u, 0xc0u, 0xf1u};
+  const uint8_t test_sig_tail[] = {0x1fu, 0x0cu, 0x75u, 0xe7u};
+  const uint8_t *dp = testDesc;
+  if (!memcmp(dp, test_sig_head, 4) && 0x10u == get_le16(&dp[4]) &&
```
pamaury, line 246: I wonder if having a `struct` would make this a bit easier to follow?

sw/host/tests/usbdev/usbdev_stream/usb_device.cc
alees24: Done.

sw/host/tests/usbdev/usbdev_stream/usb_device.cc
alees24: I think it's best left as a simple byte stream, and that is how it's constructed on the device side (sw/device/lib/testing/usb_testutils_controlep.c). It's intended only to be a simple description of the test so that the host side can create appropriate streams, without running into potential endianness issues.

sw/host/tests/usbdev/usbdev_stream/usbdev_int.cc
```diff
@@ -0,0 +1,275 @@
...
+  // Stream remains operational, even if it presently has no work on the OUT
+  // side.
+  return true;
+}
+
+bool USBDevInt::Service() {
+  if (failed_) {
+    return false;
+  }
+  // (Re)start Isochronous IN traffic if not already in progress.
```
pamaury, line 173: I think the comment is wrong because this class handles interrupt and bulk tranfers?

sw/host/tests/usbdev/usbdev_stream/usbdev_stream.h
```diff
@@ -0,0 +1,426 @@
...
+  };
+  /**
+   * Record whether or not the stream is in the process of closing
+   * (shutting down or suspending).
+   *
+   * @param  closing   Whether or not the stream is now closing.
+   */
+  void SetClosing(bool closing) { closing_ = closing; }
+  /**
+   * Indicate whether an additional I/O transfer may safely be scheuled
```
pamaury, line 77: nit: scheuled

sw/host/tests/usbdev/usbdev_stream/usbdev_stream.cc
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// Detect a stream signature within the byte stream;
+// simple restartable parser that drops all bytes until we find a header
+// signature.
+//
+// The signature permits usbdev_stream_test to pass test parameters to this
+// program as well as overcoming the issue of the mapping from device
+// endpoints/streams to USB ports not being under our control.
+uint32_t USBDevStream::SigDetect(usbdev_stream_sig_t *sig, const uint8_t *sp,
```
pamaury, line 65: I find this function to be quite confusing and hard to follow. Would the following work:
- accumulate bytes in `sig` until you have `sizeof(usbdev_stream_sig_t)` bytes
- at this point, verify if this is a valid signature, if so then great
- if not, drop the first byte, shift everyelse one byte

The only downside it would have compared to yours is that it takes more bytes before you starts discarding. But I think it would also make it much easier to follow.

sw/host/tests/usbdev/usbdev_stream/usbdev_iso.cc
```diff
@@ -0,0 +1,320 @@
...
+          sizeof(usbdev_stream_sig_t) <= pack->actual_length - dropped) {
+        // Pick up information from this packet signature.
+        SigProcess(sig);
+
+        // Valid packet received; payload includes the signature which we
+        // retain and propagate to the caller to permit synchronization.
+        uint32_t payload = pack->actual_length - dropped;
+        pktLen_.push(payload);
+
+        // Since packets amy have been dropped we must use the supplied values
```
pamaury, line 234: s/amy/may

pamaury (approved):
Overall, this looks ok to me modulo the minor comments  have left.

pamaury:
The CI errors are strange, can you try to rebase?

sw/host/tests/usbdev/usbdev_stream/BUILD
```diff
@@ -0,0 +1,35 @@
...
+        "stream_test.cc",
+        "usb_device.cc",
+        "usbdev_int.cc",
+        "usbdev_iso.cc",
+        "usbdev_serial.cc",
+        "usbdev_stream.cc",
+        "usbdev_utils.cc",
+    ],
+    hdrs = [
+        "stream_test.h",
```
pamaury, line 19: you need to add some missing headers here.

PR #22416, opened 2024-04-04
[pentest] Add timeout for extclk config

nasahlpa:
This commit adds a timeout when configuring the EXTCLK for SCA/FI penetration tests.


sw/device/tests/crypto/cryptotest/firmware/extclk_sca_fi.c
```diff
@@ -32,13 +46,16 @@ status_t handle_extclk_sca_fi_configure(ujson_t *uj) {
...
     // Enable external clock.
     TRY(dif_clkmgr_external_clock_set_enabled(&clkmgr, is_low_speed));
+    // Wait for the external clock to become active.
+    IBEX_SPIN_FOR(did_extclk_settle(&clkmgr), kSettleDelayMicros);
+    LOG_INFO("External clock enabled.");
   } else {
     // Disable external clock.
     TRY(dif_clkmgr_external_clock_set_disabled(&clkmgr));
+    // Wait for the external clock to become inactive.
+    IBEX_SPIN_FOR(!did_extclk_settle(&clkmgr), kSettleDelayMicros);
```
johannheyszl, line 56: Nit: 
I think this function and underlying reg probably do not help to get a status on a switch back to internal clock. I think the switching back might not even be working properly.

johannheyszl (approved):
Thanks @nasahlpa this helps when switching to external clk
Left one comment re switching back which I think is probably a path we will not need to use though

sw/device/tests/crypto/cryptotest/firmware/extclk_sca_fi.c
nasahlpa: Thanks for your feedback. Why do you think that switching the clock does not work? According to the [documentation](https://opentitan.org/book/hw/ip/clkmgr/doc/registers.html#extclk_ctrl--sel):
`When the current value is kMultiBitBool4True, writing a value of kMultiBitBool4False selects internal clock as clock for the system. Writing any other value during this stage has no impact.`

nasahlpa: As discussed offline with @johannheyszl, I've removed the possibility of disabling the external clock and instead added a print to advise the user to manually reset the device. Reason for that is that we are not sure whether switching back to the internal clock works.

johannheyszl: thank you

jadephilipoom (approved):

nasahlpa:
Thanks for the review! I think this can be merged as CI failure is not related to this PR.

PR #22420, opened 2024-04-04
[tlul,rtl] Explicitly cast +1 widths in tlul_adapter_host.sv

rswarbrick:
No functional change, but this silences some lint warnings which were treating a "1" literal as a 32-bit number, so e.g. "source_q + 1" was also being expanded to 32 bits.


hcallahan-lowrisc (approved):
It looks a bit funny to my eye, but if it makes the linter happy ;)

PR #22422, opened 2024-04-04
[usbdev,dv] Adds usbdev_link_in_err_vseq

MubashirSaleem775:
This PR introduces a new sequence, usbdev_link_in_err. This addition focuses on USB device functionalty, particularly where packet reception at an IN endpoint is initiated but subsequently dropped due to an error condition, thus facilitating accurate error reporting via link_in_err.


alees24:

PR #22423, opened 2024-04-04
[spi_device] Remove no longer used prim_ram_2p_adv dependency

vogelpi:
The latest version of the SPI_DEVICE block uses one instance of of prim_ram_1r1w_async_adv which depends on prim_ram_2p_pkg but not prim_ram_2p_adv.

Reading the prim_ram_2p_adv dependency without having a representation in the target technology (it's currently not used anywhere) leads to issues in the back-end flow.


vogelpi:
@a-will , it seems spi_device has supports both 2p and 1r1w memory types. The selection is based on a parameter which for Earlgrey is pointing to the 2p config. Is this intentional?

a-will:
> @a-will , it seems spi_device has supports both 2p and 1r1w memory types. The selection is based on a parameter which for Earlgrey is pointing to the 2p config. Is this intentional?

Yes. The 2p config is the original, the one used in ES, and the selection was not changed because the 2p variant is the one available in the technology.

1r1w is used in Darjeeling, and it's the opposite situation for the technology associated with that top.

a-will (approved):
LGTM

vogelpi:
Ah, now I understand. Thanks @a-will for explaining!

PR #22425, opened 2024-04-04
[prim, rom_ctrl] Increase number of PRINCE rounds for improved security

vogelpi:
This commit increases the number of half rounds of the PRINCE cipher used for memory scrambling from 2 (5 effective rounds) to 3 (7 effective rounds).

This change affects the main and retention SRAMs, the OTBN I- and DMEM, the I-Cache memories of Ibex, and ROM_CTRL.

~NOTE: this is currently just a draft as we first need to determine if increasing the number of half rounds is realistic from a timing perspective.~
UPDATE: we now got that feedback from @meisnere . It looks like that the area impact of this is low and that there is no negative timing impact.

I've now also made the required changes to the tooling and also increased the number of PRINCE rounds for ROM_CTRL to align everything.


vogelpi:
@meisnere Note that the CI failure is because I haven't touched the software side yet to modify the generation of the images for back door loading. Nevertheless, this should be sufficient for the synthesis experiment.

rswarbrick (approved):
This looks good to me. I almost raised review comments a couple of times and (each time) understood what was going on a couple of hunks later. Nice!

PR #22427, opened 2024-04-04
[pentest] Add SCA tests for HMAC

nasahlpa:
This commit adds the following SCA penetration tests for HMAC:
- hmac.sca.batch_random
- hmac.sca.batch_fvsr
- hmac.sca.single

The tests are based on the cryptolib and cryptotest applications and are extended with the SCA specific functionalities (on-chip generation of random and FvsR data as well as setting and unsetting the trigger).


sw/device/tests/penetrationtests/firmware/hmac_sca.c
```diff
@@ -0,0 +1,222 @@
...
+    }
+    prng_rand_bytes(batch_messages[it], kMessageLength);
+    sample_fixed = batch_messages[it][0] & 0x1;
+  }
+
+  // Invoke HMAC for each data set.
+  uint32_t tag_buf[kTagLengthWord];
+  for (size_t it = 0; it < uj_it.num_iterations; it++) {
+    TRY(trigger_hmac(batch_keys[it], batch_masks[it], batch_messages[it],
+                     tag_buf));
```
vogelpi, line 150: IIRC, we do or did chain the results of operations together in batch mode by xoring ciphertexts / digests. As it is now, you'll not notice mismatches happened for the hash operations inside the batch, you only check the last one.

vogelpi (approved):
LGTM, but I would XOR results together in batch mode to check the entire batch instead of just the last operation.

sw/device/tests/penetrationtests/firmware/hmac_sca.c
nasahlpa: Thanks for pointing this out!

The reason why I implemented this was to make sure that the host and the device compute the same random data for the HMAC operation. By sending the last digest, I know when one of the two parties messed up the generation of the random data.

My assumption is that when the host and the device HMAC code get the same random data at its input for each trace, the output should be valid also for each trace.

I know that we do this differently for HMAC (but not for AES). Does it make sense for you to keep it as it is or should I take the XOR of all operations together?

vogelpi: The thing is that if the last digest is from the fixed set, it won't tell you anything about the generation of the random data. I would follow the approach taken for KMAC here and XOR them together.

vogelpi: Ah, now I notice that this is fvsr_key and that the message is always random. Since you use the same software PRNG for random keys and the always random message, checking the last digest is sufficient actually.

vogelpi: No need to make changes then :-)

PR #22430, opened 2024-04-04
[usbdev,dv] Avoid duplicate file when building usbdev tests

rswarbrick:
Before this change, usb20_if.sv appeared in two cores: usb20_usbdpi and usb20_agent. This causes silly warnings from EDA tools like VCS which point out that we have a spare copy of the file.

Avoid the duplication and leave a comment that explains what's going on.


rswarbrick:
Take two! I've just force-pushed with a version that should make a bit more sense and still avoids the duplicate file. @alees24, would you mind taking a look at the new improved version? :-)

alees24 (approved):
Thanks.

PR #22434, opened 2024-04-04
[lc_ctrl,dv] Fix constraint in lc_ctrl_errors_vseq

rswarbrick:
I don't think the types in this constraint made sense (and VCS warns when asked to compile it). Change to types that actually match.


marnovandermaas (approved):
Yes, I think you're right. It should use values from `lc_state_e` and not `dec_lc_state_e`.

PR #22436, opened 2024-04-04
[lc_ctrl,dv] Fix width of coverage binding

rswarbrick:
The existing binding expected a single object of type lc_tx_t, but the thing that it was sampling changed type with commit fa447d0cdc9: it is now NumRmaAckSigs different lc_tx_t objects.

VCS correctly complains at build time about the width mismatch. I couldn't easily see how to name the instances sensibly and "bind things in a loop", so I've used a simple approach where the generate module inside the thing that gets bound.


vogelpi (approved):
Thanks @rswarbrick , this looks good to me!

PR #22269, opened 2024-03-26
[pentest] Add Ibex FI & SCA tests

nasahlpa:
This commit adds the following SCA and FI penetration tests:
- ibex.address_translation
- ibex.address_translation_config
- ibex.char.csr_write
- ibex.char.csr_read
- ibex.sca.key_sideloading

PR #22450, opened 2024-04-05
[pentest] Separate from cryptotest

nasahlpa:
This commit separates the penetration testing firmware from the cryptotest. The reason for that is that we are running out of instruction memory when having both tests in one firmware.

The penetration testing firmware for the FPGA can be build with:
`./bazelisk.sh build //sw/device/tests/penetrationtests/firmware:firmware_fpga_cw310_test_rom`


RyanTorok (approved):
LGTM. Just out of curiosity, why have some `UJSON_CHECK_DIF_OK` checks been changed to `TRY`?

nasahlpa:
> LGTM. Just out of curiosity, why have some `UJSON_CHECK_DIF_OK` checks been changed to `TRY`?

Thanks Ryan!
I've added the `UJSON_CHECK_DIF_OK` macro in a previous commit because I was not aware that a similar functionality is already provided in the codebase. Tim pointed this out in a previous PR (#22269) and I took the opportunity to remove it now.

engdoreis (approved):
LGTM

engdoreis:
Merging this PR because the test that failed on CI is not related and also fails on the earlgrey_es_sival branch.

PR #22451, opened 2024-04-05
[dvsim] Split and rename Modes.py

rswarbrick:
The capitalised filenames were named after the classes they defined (confusingly, I renamed "Modes" to "Mode" in 5c99c6ff4e). Also, this file defined lots of different classes.

This commit pulls out the more interesting subclasses, keeping with the CapitalisedName style and then renames the remaining file to modes.py so that we're no longer claiming that it contains a class called Modes.


hcallahan-lowrisc (approved):
LGTM!

PR #22455, opened 2024-04-05
[dvsim] Tweak mode find/merge to be simpler to understand

rswarbrick:
The existing code worked through a list of desired names and found
modes with these names. If the merge_modes flag was true then it also
merged each of the hits into the object supplied as the "mode"
argument. It turns out that this merging behaviour was actually only
used at one call-site.

This commit simplifies things. We now have a function called
find_mode_list which just does the filtering operation. The merging
part of things gets moved up to the call site.


hcallahan-lowrisc (approved):

rswarbrick:
Oops, I didn't realise but I think I opened two copies of this! A different copy got merged in #23371.

PR #21533, opened 2024-02-16
[Sival, i2c] Repeated start

engdoreis:
Implement the i2c repeated start test using bitbanging.

Fix: https://github.com/lowRISC/opentitan/issues/19840

PR #22457, opened 2024-04-05
Manual Backport 21533 to master

engdoreis:
This is a manual backport of https://github.com/lowRISC/opentitan/pull/21533
The conflict was related to the introduction of a CRC on the ujson.

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -4843,7 +4843,6 @@ opentitan_test(
         "i2c_target_test.c",
     ],
     cw310 = cw310_params(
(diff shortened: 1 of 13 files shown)


sw/device/tests/BUILD
```diff
@@ -4843,15 +4843,17 @@ opentitan_test(
         "i2c_target_test.c",
     ],
     cw310 = cw310_params(
```
pamaury, line 4845: Not your fault but the comment above the `opentitan_test` looks obsolete?

pamaury (approved):
I had a cursory glance and it looks good to me.

sw/device/tests/BUILD
engdoreis: Good catch, I'll remove it in a follow-up PR. I don't want to trigger CI again. Thanks.

PR #22458, opened 2024-04-05
[sival] Add checks for RX line breaks to UART parity test

jwnrt:
This PR extends the UART parity test to also expect a line break after receiving a message with the wrong parity.

The line break is detected at each of the thresholds that Earl Grey supports.

We could use HyperDebug's bitbanging feature to create breaks of exact timings for each threshold, but looking at that feature I'm not sure it's necessary for the complexity it adds to the test.


sw/device/lib/dif/dif_uart.c
```diff
@@ -124,12 +125,19 @@ dif_result_t dif_uart_configure(const dif_uart_t *uart,
     return kDifBadArg;
   }
 
+  // Check requested RXBLVL is within bounds.
+  uint32_t rxblvl = config.rx_break_level;
+  if ((rxblvl & UART_CTRL_RXBLVL_MASK) != rxblvl) {
```
engdoreis, line 130: Would you mind testing this branch on the unittest. Also adding a new test for the function below.

sw/device/tests/uart_parity_break_test.c
```diff
@@ -162,6 +206,13 @@ static status_t execute_test(void) {
   LOG_INFO("Receiving data with wrong parity");
   IBEX_SPIN_FOR(uart_irq_rx_parity_err_fired, 1 * 1000 * 1000);
 
+  // Wait for host to sync and then expect to receive a line break.
+  uart_irq_rx_break_err_expected = true;
+  OTTF_WAIT_FOR(test_phase == kTestPhaseBreakErr, kCommandTimeoutMicros);
+
+  LOG_INFO("Waiting for line break");
+  IBEX_SPIN_FOR(uart_irq_rx_break_err_fired, 1 * 1000 * 1000);
```
engdoreis: I would be wiser to use `ATOMIC_WAIT_FOR_INTERRUPT`

jwnrt: Thanks, I'll remember this one

timothytrippel (approved):

PR #22459, opened 2024-04-05
[i2c] Controller mode cleanup of NACK handling and transaction boundaries

a-will:
Consolidate NACK handling events into a single "controller halt" status
interrupt. The event bits in the CONTROLLER_EVENTS CSR combine to drive
the interrupt. Clear the bits to acknowledge the interrupt.

Change up the NACK handling to finish the current byte, force transition
to the Idle state, and hold there, as well as beginning the timeout
counter's incrementing there. SCL will be held low if the transaction
was not stopped already.

Track automatic Stop generation to eliminate races with software, since
changes to the control bits could occur while the Stop symbol was being
generated, leading to a transition to the PopFmtFifo state and the
discarding of an unused FMT FIFO entry. Proceed directly to Idle if the
current Stop symbol was automatically generated.


hw/ip/i2c/doc/interfaces.md
```diff
@@ -29,7 +29,7 @@ Referring to the [Comportable guideline for peripheral device functionality](htt
...
 | acq_threshold    | Status | target mode interrupt: asserted whilst the ACQ FIFO level is above the high threshold. This is a level status interrupt.                                                                                             |
 | rx_overflow      | Event  | host mode interrupt: raised if the RX FIFO has overflowed.                                                                                                                                                           |
-| nak              | Event  | host mode interrupt: raised if there is no ACK in response to an address or data write                                                                                                                               |
+| controller_halt  | Status | host mode interrupt: raised if the controller FSM is halted. Check there is no ACK in response to an address or data write.                                                                                          |
```
a-will, line 32: Oops, I think I didn't finish editing this line

hw/ip/i2c/doc/theory_of_operation.md
```diff
@@ -87,23 +87,22 @@ Typically every byte transmitted must also receive an ACK signal, and the IP wil
...
+If the `controller_halt` interrupt is asserted, the Byte-Formatted Programming Mode FSM will halt until the interrupt condition has been cleared in the [`CONTROLLER_EVENTS.NACK`](registers.md#controller_events) register.
 Software will likely want to do one of two things in handling this irq:
-1. End the current transaction by setting [`CTRL.ENABLEHOST`](registers.md#ctrl) to `1'b0`, and clear the FIFOs ready to begin a new transaction.
-2. Clear the FIFOs, then begin a new transfer without ending the transaction by add new FDATA indicators to the FMTFIFO.
+1. End the current transaction by setting [`CTRL.ENABLEHOST`](registers.md#ctrl) to `1'b0`, and clear the FIFOs to be ready to begin a new transaction.
+2. Clear the FIFOs, then begin a new transfer without ending the transaction by adding new FDATA indicators to the FMTFIFO.
    In this case, the first entry written to the FMTFIFO should have [`FDATA.START`](registers.md#fdata) set to `1'b1` to create a repeated start condition.
 
 While the FSM is halted, SCL/SDA are released and the bus is observed as having an elongated 9th clock period.
```
a-will, line 99: Ah, this is no longer true. Need to update.

hw/ip/i2c/data/i2c.hjson
```diff
@@ -204,8 +209,7 @@
     { name: "CTRL"
       desc: "I2C Control Register"
       swaccess: "rw"
-      hwaccess: "hrw"
-      hwqe:     "true",
+      hwaccess: "hro"
```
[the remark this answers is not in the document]
a-will: Not really. It was an integral mechanism to how the unhandled NACK timeout previously worked. Commits should generally be atomic, and that's why you see all of this at once--Aside from a few nitty renames, it's all required simultaneously to work.

hw/ip/i2c/rtl/i2c_controller_fsm.sv
```diff
@@ -30,6 +29,7 @@ module i2c_controller_fsm import i2c_pkg::*;
   input                      fmt_flag_read_continue_i,// host to send Ack to final byte read
   input                      fmt_flag_nak_ok_i,       // no Ack is expected
   input                      unhandled_unexp_nak_i,
+  input                      unhandled_nak_timeout_i, // NACK handler timeout event not cleared
```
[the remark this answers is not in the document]
a-will: Oh, I definitely agree, but I will defer these software-invisible renames until I have nothing higher priority to do. We've run out enough of the clock in the scoping and resourcing phase that we're now going to need to race a bit to complete all the required functionality.

We can create an issue for this to track it, though!

a-will:
> Thanks for making these changes! I find this PR a bit difficult to review because of a few reasons:
> 
> * Autogenerated files are mixed in with functional changes.

Note that the autogenerated changes are related and _should_ be in the same commit for atomicity. You don't generally have to look at files like interfaces.md, registers.md, _reg_top.sv, and _reg_pkg.sv, though, unless you have reason to suspect the reggen output, hehe.

> * Reorganizational changes are mixed in with functional changes. (For example implementing the unhandled NACK counter and moving the unexpected NACK.)

Sorry about that, but recent commits ignored the file's stated structure. One always block was supposed to be _only_ about the state transition (my bad for not catching it on review!). This PR redoes how those NACK features work, in the file's stated style.

The current file structure doesn't follow our coding guidelines, and I think you have the same complaints I do: It's hard to read! We ought to collapse the two always blocks down to one, so state transitions are present in the same screen as state outputs. However, I won't do that until at least after I finish cleaning the DV regressions, so we can rely on them again to tell us if we have broken something.

andreaskurth (approved):
Thanks for this work @a-will -- LGTM

hw/ip/i2c/data/i2c.hjson
```diff
@@ -741,17 +738,14 @@
...
 
-            In Host-Mode during an active Controller-Transmitter transfer, if the Target NACKs a byte
-            the 'nak' interrupt is asserted and the Byte-Formatted Programming Mode FSM halts awaiting
-            software intervention. Software must acknowledge the interrupt (CIP 'Event-Type') to allow
-            the state machine to continue, typically after clearing out the FMTFIFO to start a new transfer.
-            While halted, the active transaction is not ended (no STOP (P) condition is created), and the
-            block releases both SCL and SDA.
+            If an active Controller-Transmitter transfer receives a NACK from the Target, the !!CONTROLLER_EVENTS.NACK bit is set.
+            In turn, this causes the Controller FSM to halt awaiting software intervention, and the 'controller_halt' interrupt may assert.
+            Software must clear the !!CONTROLLER_EVENTS.NACK bit  to allow the state machine to continue, typically after clearing out the FMTFIFO to start a new transfer.
```
hcallahan-lowrisc, line 743: nit
```suggestion
            Software must clear the !!CONTROLLER_EVENTS.NACK bit to allow the state machine to continue, typically after clearing out the FMTFIFO to start a new transfer.
```

hw/ip/i2c/doc/theory_of_operation.md
```diff
@@ -87,23 +87,27 @@ Typically every byte transmitted must also receive an ACK signal, and the IP wil
...
 Software will likely want to do one of two things in handling this irq:
-1. End the current transaction by setting [`CTRL.ENABLEHOST`](registers.md#ctrl) to `1'b0`, and clear the FIFOs ready to begin a new transaction.
-2. Clear the FIFOs, then begin a new transfer without ending the transaction by add new FDATA indicators to the FMTFIFO.
+1. End the current transaction by setting [`CTRL.ENABLEHOST`](registers.md#ctrl) to `1'b0`, and clear the FIFOs to be ready to begin a new transaction.
+2. Clear the FIFOs, then begin a new transfer without ending the transaction by adding new FDATA indicators to the FMTFIFO.
    In this case, the first entry written to the FMTFIFO should have [`FDATA.START`](registers.md#fdata) set to `1'b1` to create a repeated start condition.
 
-While the FSM is halted, SCL/SDA are released and the bus is observed as having an elongated 9th clock period.
+Software should then clear the [`CONTROLLER_EVENTS.NACK`](registers.md#controller_events) to proceed with the next operation.
```
hcallahan-lowrisc, line 99: I might just add a clarification here that the `controller_halt` interrupt is CIP status-type, and hence does not need to be acknowledged, but will deassert on it's own when software clears the `CONTROLLER_EVENTS` register.

hw/ip/i2c/rtl/i2c_core.sv
```diff
@@ -321,17 +312,12 @@ module i2c_core import i2c_pkg::*;
...
   // may NACK our bytes. In Byte-Formatted Programming Mode, each FDATA format indicator
   // can set the 'NAKOK' bit to ignore the Target's NACK and proceed to the next item in
-  // the FMTFIFO. If 'NAKOK' is not set, the 'nak' interrupt is asserted, and the FSM
-  // halts until software intervenes.
-  // To acknowledge the 'NACK', software should acknowledge the interrupt using the
-  // standard W1C mechanism to the INTR_STATE register.
-  assign unhandled_unexp_nak = reg2hw.intr_state.nak.q;
+  // the FMTFIFO. If 'NAKOK' is not set, the 'controller_halt' interrupt is asserted, and the FSM
+  // halts until software intervenes. In addition, the 'controller_events.nack' status bit is set.
+  // To acknowledge the 'NACK', software should clear the status bit using the standard W1C
```
hcallahan-lowrisc, line 317: nit: I might remove the 'standard' wording here, as that was intended as a callback to how Event-type interrupts are acknowledged.

hcallahan-lowrisc (approved):
LGTM, some tiny doc nits that can be wrapped up later on.

hw/ip/i2c/rtl/i2c_core.sv
a-will: Done. Thanks!

PR #22463, opened 2024-04-08
[pentest] Add tweaks to FI & SCA code

nasahlpa:
Based on feedback during the SCA & FI penetration testing, this commit changes the following:
- Avoid code optimization for conditional branch test
- Optimize init of different tests
- Performance optimizations

### DIFF
--- a/sw/device/tests/penetrationtests/firmware/ibex_fi.c
+++ b/sw/device/tests/penetrationtests/firmware/ibex_fi.c
@@ -34,6 +34,11 @@ static str_fn_t increment_100x1_remapped = (str_fn_t)increment_100x1;
 // Interface to Ibex.
 static dif_rv_core_ibex_t rv_core_ibex;
 
+// Indicates whether flash already was initialized for the test or not.
+static bool flash_init;
+// Indicates whether flash content is valid or not.
+static bool flash_data_valid;
+
 // Make sure that this function does not get optimized by the compiler.
 uint32_t increment_counter(uint32_t counter) __attribute__((optnone)) {
   uint32_t return_value = counter + 1;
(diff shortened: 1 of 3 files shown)


sw/device/tests/penetrationtests/firmware/ibex_fi.c
```diff
@@ -102,7 +107,8 @@ static dif_flash_ctrl_device_info_t flash_info;
 // we can do the write/read test without the risk of clobbering data
 // used by the program.
 OT_SECTION(".data")
-static volatile uint32_t sram_main_buffer[32];
+
+static volatile uint32_t sram_main_buffer[256];
```
m-temp, line 111: As discussed offline: Could you add a small comment, why this window is increased, but only 32 different entries are used later: `(i % 32)` in the code

m-temp (commented):
LGTM, but as always: I'm not a device software expert

sw/device/tests/penetrationtests/firmware/ibex_fi.c
nasahlpa: Thanks @m-temp - good point. I've added a description to this.

m-temp (approved):
LGTM

nasahlpa:
Thanks, I've rebased it.

sw/device/tests/penetrationtests/firmware/ibex_fi.c
```diff
@@ -547,26 +568,32 @@ status_t handle_ibex_fi_char_sram_read(ujson_t *uj) {
       mmio_region_from_addr(sram_main_buffer_addr);
 
   // Write reference values into SRAM.
-  for (int i = 0; i < 32; i++) {
+  for (int i = 0; i < 256; i++) {
     mmio_region_write32(sram_region_main_addr, i * (ptrdiff_t)sizeof(uint32_t),
-                        ref_values[i]);
+                        ref_values[i % 32]);
```
johannheyszl, line 573: it may be helpful to not mod 32 to be able to distinguish the 256 values, but on the other hand, if address is corrupt it will read other initialized mem with zeros (?) which we cannot dinstinguish anyways

nasahlpa: Yeah exactly, reading from a faulty address would, with a high probability, return something from memory.

PR #22464, opened 2024-04-08
[usbdev,dv] Get rid of get_out_response_from_device

rswarbrick:
This task didn't actually do anything interesting! It was actually just checking that the supplied usb20_item had the PID that was provided. There were lots of other lines as well, but these had no effect.

Probably better to ditch the extra noise.


alees24 (approved):
Thanks.

PR #22476, opened 2024-04-08
[i2c,dv,cleanup] Refactor timing-related configuration into timing_pkg

hcallahan-lowrisc:
> This change creates a common home for all timing related routines, constraints and constants. It contains a class 'i2c_timing_cfg' which can be instantiated and randomized directly to create new configurations of timing parameters. This class should be extended to add/disable/redefine constraints by child sequences and other consumers.
>
> Timing parameters are still randomized as part of the base vseq, but the whole class instance is declared as rand.
>
> Previously, the timing parameters (and their constraints) were defined in the base sequence, and sub-sequences then override parts of these constraints, or added new ones. This often led to code-duplication, and subtly different constraints often with the same comment which caused confusion.

This will be platform for further cleaning up and extending the DV in the future. In particular, I would like to implement an assertion-based bus checker for timing compliance, to then greatly simplify the monitor.

---

The new print output whenever the class is randomized looks like this:
```
UVM_INFO @ 0 ps: reporter [RNTST] Running test i2c_base_test...
UVM_INFO @         0 ps: (i2c_base_test.sv:22) uvm_test_top [uvm_test_top] set i2c agent mode to Host
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:423) [] post_randomize() : Newly randomized timing properties:
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:424) [] speed_mode                       =   FastPlus
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:425) [] scl (frequency) (desired)        =       1.00 MHz
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:426) [] scl (period)    (desired)        =    1000000 ps
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:427) [] clk (frequency)                  =      97.00 MHz
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:428) [] clk (period)                     =      10309 ps
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:429) [] scl/clk period ratio             =      97.00
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:430) [] scl/clk period ratio (quantized) =         97
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:431) [] scl (period)    (coerced)        =        117 cycles
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:432) [] scl (period)    (coerced)        =    1206186 ps
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:433) [] scl (frequency) (coerced)        =     829059 Hz
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:562) [i2c_timing_pkg] 
-------------------------------------
timing_properties
thigh     : 46
tlow      : 50
t_r       : 11
t_f       : 10
thd_sta   : 13
tsu_sta   : 13
tsu_sto   : 13
tsu_dat   : 13
thd_dat   : 13
t_buf     : 39
eTimeout  : 0
tTimeout  : 35
-------------------------------------
UVM_INFO @         0 ps: (i2c_timing_pkg.sv:602) [i2c_timing_pkg] 
-------------------------------------
derived_timing_properties
tSetupStart       : 32
tHoldStart        : 23
tSetupData        : 37
tClockStart       : 11
tClockLow         : 1
tClockPulse       : 57
tHoldBit          : 21
tClockStop        : 49
tSetupStop        : 26
tHoldStop         : 29
eTimeOut          : 0
tTimeOut          : 35
tSetupBit         : 37
tStretchHostClock : 61
-------------------------------------
timing_error_properties
tSdaUnstable          : 0
tSclInterference      : 0
tSdaInterference      : 0
-------------------------------------

```


hw/ip/i2c/dv/env/seq_lib/i2c_target_perf_vseq.sv
```diff
@@ -5,30 +5,36 @@
...
-    thd_dat   == 1;
-    t_timeout == 1;
-    e_timeout == 1;
-    tsu_sta   == 1;
+  class target_perf_timing_cfg extends i2c_timing_cfg;
+    // Fast timing values programmed to registers
+    constraint target_perf_c {
+      tc.t_r      == 1;
+      tc.t_f      == 1;
+      tc.thd_sta  == 1;
```
a-will, line 14: Note that we still require...

```
thd_dat > 0
thd_dat + 1 < thd_sta
```

Thus, for minimum values, the 3 here was correct.

hcallahan-lowrisc: Yep good point, that one got lost in a rebase. Cheers

hw/dv/sv/i2c_agent/i2c_agent.core
```diff
@@ -9,7 +9,10 @@ filesets:
     depend:
       - lowrisc:dv:dv_utils
       - lowrisc:dv:dv_lib
+      - lowrisc:ip:i2c
```
a-will, line 12: We should probably bring the constants / enums up to the agent's packages. It's kind of backwards for the agent to depend on the DUT.

Then the DV environment for the IP can reference the agent's packages for those values, if needed. Meanwhile, the RTL will continue to do its own thing.

If we wanted to create shared enums for even the RTL to use, I guess the place for that would be a core in the `prim` core library (prim_i2c_pkg? haha), since it's kind of our only common library.

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  // DUT Implementation Limitations
+  constraint implementation_c {
+    solve tc.t_r, tc.tsu_dat, tc.thd_dat before tc.tlow;
+    solve tc.t_r, tc.tsu_sta             before tc.t_buf;
+
+    // > 'tlow' must be at least 2 greater than the sum of (tsu_dat + t_r + thd_dat)
+    // > because the flopped outputs of the FSM (see #15003) reduces tClockLow by 1.
+    tc.tlow >= (tc.tsu_dat + tc.t_r + tc.thd_dat) + 2;
```
a-will, line 97: Hm... this is awkward. We have IP-specific timing constraints in the agent. This feels misplaced to me.

a-will (commented):
I only very briefly looked so far, but it looks like we've got some bleeding across abstraction boundaries. It feels to me like the agent should have its functionality encapsulated such that IP requirements aren't part of its code.

I'll defer to the opinions of DV experts here, though!

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+
+  // Assertion probability (in percentages)
+  uint i2c_prob_sda_unstable     = 30;
+  uint i2c_prob_sda_interference = 30;
+  uint i2c_prob_scl_interference = 70;
+
+  // These variables create signalling errors to test error-detection features.
+  rand uint t_sda_unstable;     // sda     unstable time during the posedge_clock
+  rand uint t_sda_interference; // sda interference time during the posedge_clock
+  rand uint t_scl_interference; // scl interference time during the posedge_clock
```
rswarbrick, line 59: These variables seem to match ones in `i2c_timing_pkg.sv` with names like `tSdaUnstable`. Can we get rid of those? (We probably also want to move the more informative documentation comments in that package to live here)

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  constraint spec_minimums_c {
+    tc.tsu_sta == limcyctc.min.tsu_sta;
+    tc.thd_sta == limcyctc.min.thd_sta;
+    tc.tsu_dat == limcyctc.min.tsu_dat;
+    // thd_dat only has non-zero specifed minimum value in one configuration:
+    // - Standard-mode for CBUS-compatible controllers
+    tc.thd_sta == limcyctc.min.thd_sta;
+    tc.tsu_sto == limcyctc.min.tsu_sto;
```
rswarbrick, line 138: This looks slightly surprising to me. The block below sets lower bounds for some signals, but this block fixes signals to their minimal values. Is that intended?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+    t_sda_unstable     inside {[0 : tc.t_r + tc.thigh + tc.t_f - 1]};
+    t_sda_interference inside {[0 : tc.t_r + tc.thigh + tc.t_f - 1]};
+    t_scl_interference inside {[0 : tc.t_r + tc.thigh + tc.t_f - 1]};
+  }
+
+  constraint t_timeout_c {
+    tc.tTimeout inside {[i2c_min_timing : i2c_max_timing]};
+  }
+
+  constraint spec_minimums_c {
```
rswarbrick, line 131: I see that this constraint and `spec_maximums_c` are both disabled by default. Could you add a comment above the constraints that explain in words what they are meant to be doing?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+
+  /////////////
+  // METHODS //
+  /////////////
+
+  function new (string name = "");
+    super.new(name);
+    // The following constraints are disabled by default.
+    spec_minimums_c.constraint_mode(0);
+    spec_maximums_c.constraint_mode(0);
```
rswarbrick, line 168: I'm not a huge fan of `constraint_mode()`: it's a bit tricky to debug because the object no longer has the "enable constraint FOO" flag. I wonder whether it would make sense to have e.g. a `use_minimum_timings` flag and condition the constraints on that flag with the `->` operator?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+  endfunction : new
+
+  function void pre_randomize();
+    super.pre_randomize();
+
+    // Set the nominal frequency.
+    // - Ensure the input clock frequency is high enough to meet minimum
+    //   thigh and tlow cycle counts for a given mode.
```
rswarbrick, line 176: This code took me quite a while to understand. In hindsight, I would have got there faster if this comment block was *after* the bit where we're picking `speed_mode` based on the input clock frequency. Would you mind moving it?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+    // For FastPlus, need at least 8 MHz:
+    // - tlow  = 4
+    // - thigh = 3
+    // - t_r   = 1
+    // - t_f   = 1
+    // ...to satisfy all specs with high enough thigh and tlow to satisfy
+    // the round-trip time (both greater than 3 cycles).
+    // That yields a nominal I2C frequency of 889 kHz (889 / 1000 kHz == 88.9%).
+
+    // The true, coerced value will be determined in post_randomize().
```
rswarbrick, line 197: I think this comment should go after the randomisation of `speed_mode` and before the case statement.

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+
+    // Get the spec'd limits for this speedmode.
+    get_spec_timing_cfg(speed_mode);
+  endfunction : pre_randomize
+
+  function void post_randomize();
+    super.post_randomize();
+    // Derived Timing Parameters
+    dtc = get_derived_timing_cfg();
+    get_derived_error_stimulus();
```
rswarbrick, line 218: The flow doesn't seem quite right here: the first function returns a partially completed object. We store that object into a class variable, then the second function finishes the object there.

Maybe this is cleanest if everything operates on the class variable?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+
+    // Ensure these parameters must be greater than zero
```
rswarbrick, line 220: Maybe a better comment would be "These parameters must be greater than zero because ...". A reader can hopefully see what the three lines do :-)

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+  time t_clk_i; // Input clock period (can only store down to the `timescale precision, 1ns)
+
+  speed_mode_e speed_mode;
+
+  // Desired Values
+  uint scl_frequency;
+  time scl_period;
+  uint scl_period_ratio;
+  // Actual/Expected Values
+  uint coerced_scl_frequency;
```
rswarbrick, line 24: This guy is derived from `coerced_scl_period_cycles` with a flooring operation (in `post_randomize`).

It's probably cleaner to represent this as a real number, but I'm not even convinced we actually need the class variable! The only read happens just after the value is set in `post_randomize`. Maybe make it local to that function?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+    `DV_CHECK_GT_FATAL(dtc.tHoldStop, 0)
+
+    // Coerce value after quantization. Actual frequency is different from the
+    // randomized setting, due to the granularity of the dividers.
+    coerced_scl_period_cycles = tc.tlow + tc.t_r + tc.thigh + tc.t_f;
+    coerced_scl_frequency    = f_clk_i / coerced_scl_period_cycles;
+
+    `uvm_info(`gfn, "post_randomize() : Newly randomized timing properties:", UVM_LOW)
+    `uvm_info(`gfn, $sformatf("speed_mode                       = %10s",
+      speed_mode.name()), UVM_LOW)
```
rswarbrick, line 232: Is this indentation following a style guide? I think my preference is with Emacs if not :-)

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+      scl_period_ratio), UVM_LOW)
+    `uvm_info(`gfn, $sformatf("scl (period)    (coerced)        = %10d cycles",
+      coerced_scl_period_cycles), UVM_LOW)
+    `uvm_info(`gfn, $sformatf("scl (period)    (coerced)        = %10t",
+      (coerced_scl_period_cycles * (1s / f_clk_i))), UVM_LOW)
+    `uvm_info(`gfn, $sformatf("scl (frequency) (coerced)        = %10d Hz",
+      coerced_scl_frequency), UVM_LOW)
+
+    print_time_property(tc);
+    print_derived_time_property(dtc);
```
rswarbrick, line 253: I wonder whether it would be cleaner to turn e.g. `cycles_timing_cfg_t` into a `class` and give it a `print()` function.

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+    abstc.min.tsu_sto = get_tsusto_min(sm);
+    abstc.min.t_buf   = get_tbuf_min  (sm);
+    abstc.max.t_r     = get_tr_max    (sm);
+    abstc.max.t_f     = get_tf_max    (sm);
+
+    // Calculate the timing parameters in terms of the input clock
+    // Here is where we apply appropriate rounding.
+
+    // Minimums
+    limcyctc.min.thigh   = $ceil(real'(abstc.min.thigh)   / real'(1s / f_clk_i));
```
rswarbrick, line 275: A possibly silly question, but...

At the moment, we're calculating `$ceil(A / (1 / B))`. Isn't that the same as `$ceil(A*B)`? And both values are integers, so that's just `A*B`! (Is there some rounding that I'm not spotting?)

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+
+  function automatic timing_cfg_t get_derived_timing_cfg();
+    timing_cfg_t dtc;
+
+    // The derived parameters for the agent are largely the same as the parameters programmed
+    // into the DUT, as our if/agent is also synchronous to the source clk_i.
+    // Therefore, these calculations largely match the calculations in `i2c_fsm.sv`.
+    begin
+      timing_cfg_t fsmdtc;
+      // IDENTICAL calcs as `i2c_fsm.sv`
```
rswarbrick, line 299: I'd suggest dropping this line or merging it into the comment above.

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+      fsmdtc.tHoldBit    = tc.t_f + tc.thd_dat;
+      fsmdtc.tClockStop  = tc.t_f + tc.tlow - tc.thd_dat;
+      fsmdtc.tSetupStop  = tc.t_r + tc.tsu_sto;
+      fsmdtc.tHoldStop   = tc.t_r + tc.t_buf - tc.tsu_sta;
+      fsmdtc.eTimeOut    = tc.eTimeout;
+      fsmdtc.tTimeOut    = tc.tTimeout;
+      dtc = fsmdtc;
+    end
+
+    // > The following params are used by the TB/Agent only
```
rswarbrick, line 316: Stray ">" ?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+    dtc.tStretchHostClock = gen_num_stretch_host_clks(dtc);
+
+    // Shrink tClockLow because SCL coming from the host side is now flopped. (See #15003)
+    // This means, relative to the expectation of the DUT, we have "1 less"
+    // cycle of clock time. Specifically, if the DUT drives the clock at cycle 1,
+    // the device does not see the clock until cycle 2. This means, the device
+    // expectation of how long "low" is now shrunk by 1, since the end point is
+    // still fixed.
+    dtc.tClockLow -= (dtc.tSetupBit + 1);
+    //            -= (t_r + tsu_dat + 1);
```
rswarbrick, line 330: Commented line of code?

hw/dv/sv/i2c_agent/i2c_timing_cfg.sv
```diff
@@ -0,0 +1,358 @@
...
+    // Error Conditions //
+    dtc.tSclInterference = (!en_scl_interference) ? 0 : prob_scl_interference * t_scl_interference;
+    dtc.tSdaInterference = (!en_sda_interference) ? 0 : prob_sda_interference * t_sda_interference;
+    dtc.tSdaUnstable     = (!en_sda_unstable)     ? 0 : prob_sda_unstable     * t_sda_unstable;
+  endfunction : get_derived_error_stimulus
+
+  function int gen_num_stretch_host_clks(timing_cfg_t tc);
+    // By randomly pulling down scl_o "offset" within [0:2*tc.tTimeOut],
+    // 'intr_stretch_timeout_o' interrupt would be generated uniformly.
+    // To test this feature more regressive, there might need a dedicated vseq (V2)
```
rswarbrick, line 345: I realise that you didn't write this comment in the first place, but the grammar is a bit wonky here. Maybe "If we decide we need to test this feature more aggressively, we might need a dedicated vseq in which TIMEOUT_CTRL.EN is always set." ?

rswarbrick (commented):
Only a partial review I'm afraid! I got to the bottom of `i2c_timing_cfg.sv`.

hw/dv/sv/i2c_agent/i2c_agent.core
hcallahan-lowrisc: Yeah I felt like the RTL's package was the lowest common denominator for shared enums from what we had. A standalone prim pkg is probably a better abstraction though! I'll split it out.

PR #22483, opened 2024-04-08
[sival] fix GPIO tests on silicon

nbdd0121:
The current test failure is caused by:
* "teacup" interface not being recongised by test harness.
* Slight difference in pin availability compared to hyper310.

### DIFF
--- a/sw/device/lib/testing/pinmux_testutils.c
+++ b/sw/device/lib/testing/pinmux_testutils.c
@@ -140,11 +140,20 @@ const dif_pinmux_index_t kPinmuxTestutilsGpioMioOutPins[kDifGpioNumPins] = {
     kTopEarlgreyPinmuxMioOutIor12, kTopEarlgreyPinmuxMioOutIor13};
 
 uint32_t pinmux_testutils_get_testable_gpios_mask(void) {
-  if (kDeviceType == kDeviceFpgaCw310) {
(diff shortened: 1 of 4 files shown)


pamaury (approved):
Is the pin list exhaustive or does it just need to be large enough?

nbdd0121:
It's not exhaustive. For the gpio_pinmux_test I just remove ones that not usable for GPIO testing from the hyper310 list, and for the testutil change I use the gpio_pinmux_test list to generate the mask.

sw/device/lib/testing/pinmux_testutils.c
```diff
@@ -140,11 +140,20 @@ const dif_pinmux_index_t kPinmuxTestutilsGpioMioOutPins[kDifGpioNumPins] = {
...
+      // Only IOR6, IOR7, and IOR10 to IOR13 are available for use as GPIOs.
+      return 0xfc000000;
+    case kDeviceSilicon:
+      // IOA3/6, IOB6, IOC9-12, IOR5-7 and IOR10-13.
+      return 0xfe0f0248;
+    default:
+      CHECK(false);
+      return 0;
   }
 }
```
luismarques, line 158: This part was hard to review but it isn't immediately obvious how to automate those masks without paying some code size / execution time penalty, so I guess it's fine as-is.

luismarques (approved):
LGTM.

PR #22485, opened 2024-04-08
[rv_dm,lint] Waive unused port in rv_dm.sv

rswarbrick:
This was connected up by commit b3c72069c2e. The point is that Topgen (which wires stuff up) doesn't know how to connect the reset without connecting the clock as well.

This is already waived for AscentLint in rv_dm.waiver.


andreaskurth (approved):

andreaskurth:
Failing CI check is spurious -- merging

PR #22489, opened 2024-04-09
[ast] Add dependency in fileset_partner to select correct ast_pkg

sha-ron:
Add missing condition to use the correct ast_pkg.


sha-ron:
@matutem can you please help to figure out if CI CW310 failure is related or it is ok to merge?

eshapira (approved):

PR #22491, opened 2024-04-09
[prim,fpv] Only allow unconstrained counters in prim_count FPV

rswarbrick:
The counters in prim_count have a magic delta signal (fpv_force), which can be used to make them jump around. This causes all manner of problems in FPV tests for other blocks that happen to include prim_count.

Tighten the ifndef line so that fpv_force is only allowed to be nonzero when we're doing FPV for prim_count on its own. This way, the outputs of the prim_count instances won't randomly jump around in blocks that want to use them.

Honestly, I suspect that the fpv_force signal isn't sufficiently constrained to be useful even in that situation. But that's probably an issue for a different commit!

This change will fix a bunch of failures in e.g. `entropy_src` SEC_CM FPV testing. I'm not sure whether it gets us to 100%, but I was seeing instant counterexamples before, and now it doesn't find any for at least 5 minutes... :-)

### DIFF
--- a/hw/top_earlgrey/formal/top_earlgrey_fpv_prim_cfgs.hjson
+++ b/hw/top_earlgrey/formal/top_earlgrey_fpv_prim_cfgs.hjson
@@ -93,7 +93,7 @@
                dut: prim_count_tb
                fusesoc_core: lowrisc:fpv:prim_count_fpv
                import_cfgs: ["{proj_root}/hw/formal/tools/dvsim/common_fpv_cfg.hjson"]
-               build_opts: ["-define ResetValue -1"]
+               build_opts: ["-define ResetValue -1 -define PrimCountFpv 1"]
                rel_path: "hw/ip/prim/prim_count/{name}/{sub_flow}/{tool}"
                cov: true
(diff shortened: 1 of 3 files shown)


hw/top_earlgrey/formal/top_earlgrey_fpv_prim_cfgs.hjson
```diff
@@ -103,7 +103,7 @@
                dut: prim_count_tb
                fusesoc_core: lowrisc:fpv:prim_count_fpv
                import_cfgs: ["{proj_root}/hw/formal/tools/dvsim/common_fpv_cfg.hjson"]
-               build_opts: ["-define ResetValue 0"]
+               build_opts: ["-define ResetValue 0 -define PrimCountFpv"]
```
vogelpi, line 106: Nit: do you need to specify a value here as well? Above, you set `PrimCountFpv` to 1.

vogelpi (approved):
LGTM, but I have one question. Thanks @rswarbrick !

hw/top_earlgrey/formal/top_earlgrey_fpv_prim_cfgs.hjson
rswarbrick: Nuts, you're right. Thanks!

Issue #21204, opened 2024-02-06
[otp_ctrl, rom] Move ROM keys to OTP


moidx:
### Description

This is a [draft RFC doc](https://docs.google.com/document/d/1N0-Wvp73-rzWcpDGFwxt3FAa4oUsOFOgB9SneoOqPyU/edit?pli=1) for consideration of the TC after discussion with the EarlGrey subgroup and the Software WG.

Tasks 

- [x] https://github.com/lowRISC/opentitan/pull/23320
- [x] #22697
- [X] https://github.com/lowRISC/opentitan/pull/21270
- [X] https://github.com/lowRISC/opentitan/pull/22494
- [X] https://github.com/lowRISC/opentitan/pull/22743
- [X] https://github.com/lowRISC/opentitan/pull/22741
- [X] https://github.com/lowRISC/opentitan/pull/22744
- [X] https://github.com/lowRISC/opentitan/pull/22950
- [x] https://github.com/lowRISC/opentitan/pull/23300
- [x] https://github.com/lowRISC/opentitan/pull/23309
- [x] https://github.com/lowRISC/opentitan/pull/23315
- [x] https://github.com/lowRISC/opentitan/pull/23316
- [x] https://github.com/lowRISC/opentitan/pull/23318
- [x] https://github.com/lowRISC/opentitan/pull/23319

msfschaffner:
The OTP side has now been merged, so the remaining parts of this change are SW only.

timothytrippel:
Thanks @msfschaffner !

msfschaffner:
This RFC has been approved by the TC today.

PR #22494, opened 2024-04-09
[rom] Switch SPX+ keys to be loaded from OTP instead of ROM memory

moidx:
1. Integrate `sigverify_otp_keys` module into ROM, initializing its context before performing any key operations as part of secure boot.
2. Remove unused fake/real SPX+ keys. With the OTP implementation we are going down to four key slots. The ROM implementation supported seven slots.
3. Configure the standard OTP overlay configuration to include the ROM fake keys used for `dv_sim` and `verilator` and `FPGA` testing.

This is part of https://github.com/lowRISC/opentitan/issues/21204


timothytrippel (approved):

moidx:
The following test case is failing in this PR, but also in master. Proceeding with merge.
```shell
TIMEOUT: //sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup:ft_provision_fpga_cw310_sival
```

Issue #22462, opened 2024-04-08
[top_earlgrey] Timing optimizations for production


andreaskurth:
This is a tracker for timing optimizations for Earlgrey-PROD.

Planned for M3:
- [x] Decouple `sram_ctrl`'s `a_ready` TL-UL signal from `a_valid`
  --> #22497
- [x] Decouple `sram_ctrl`'s `a_ready` TL-UL signal from `a_opcode` (through `tlul_we`)
  --> #22497
- [x] Break path from `tl_i` to `tl_o` in `tlul_sram_adapter`
  --> #22588
- [x] Rework `tlul_err_resp`
  --> #22588
- [x] Reduce the number of PRINCE rounds for the ICache scrambling back to 5 effective rounds.
  --> #22948

vogelpi:
Update: PRs to fix all feedthrough paths between Ibex and SRAM_CTRL (the critical path of the design) were filed and merged, an RC was tagged and sent to the PD team to trigger a new synthesis run and the generation of a timing report. We're now waiting for feedback.

andreaskurth:
By @vogelpi: Discussed with PD team yesterday, feedback pending.

By @andreaskurth: Timing closure for M3 is not required. Can be moved to M4 if needed.

By @moidx: That would be my recommendation as well, given the timeline.

vogelpi:
Update: we got feedback from the PD team. There are some violations left between the ICache and the main SRAM that we can hopefully get rid of by reducing the number of scrambling rounds in the ICache again. This is acceptable from a security viewpoint.

PR #17072, opened 2023-01-13
[sram_ctrl] Fix RMW corner case when escalating

msfschaffner:
If a global escalation request and a RMW transaction come in at the same time, the RMW transaction will not be able to complete even though the TL gate allows for transactions that are already pending inside the device to respond back (to make the bus behavior less erratic for a better SW debug experience).

The reason for this is the write part of the RMW transaction, since the scrambling device blocks any transaction in case the keys have been invalidated due to escalation.

This patch fixes this corner case by specifically allowing pending writes to complete upon escalation.

The rationale is that a write won't give an attacker any advantage, especially since the scrambling keys are already clobbered at that point, and the TL gate is not accepting any additional transactions from the bus.

Fixes #17064.

Signed-off-by: Michael Schaffner <msf@google.com>

PR #11863, opened 2022-04-02
[sram_ctrl] Switch to tlul_lc_gate for transaction blocking

msfschaffner:
Signed-off-by: Michael Schaffner <msf@opentitan.org>

PR #22497, opened 2024-04-09
[sram_ctrl/rtl] Timing fixes

GregAC:
I plan to put together a hacky block synthesis for sram_ctrl so I can check I've eliminated all of the feedthroughs that were causing issues. Will work on that tomorrow.

Contributes to resolving #22462.

### DIFF
--- a/hw/ip/sram_ctrl/rtl/sram_ctrl.sv
+++ b/hw/ip/sram_ctrl/rtl/sram_ctrl.sv
@@ -478,9 +478,16 @@ module sram_ctrl
     .rmw_in_progress_o(sram_rmw_in_progress)
   );
 
+  logic key_valid;
+
   // Interposing mux logic for initialization with pseudo random data.
   assign sram_req        = tlul_req | init_req;
-  assign tlul_gnt        = sram_gnt & ~init_req;
+  // This grant signal acts more like a ready internally in tlul_adapter_sram. In particular it's
+  // fine to assert it when tlul_req is low (it has no effect). So here tlul_gnt is asserted when
+  // a request from tlul_req will be granted regardless of whether a request exists. This is done
+  // for timing reasons so that the output TL ready isn't combinatorially connected to the incoming
+  // TL valid. In particular we must not use `sram_gnt` in the expression to avoid this.
+  assign tlul_gnt        = key_valid & ~init_req;
   assign sram_we         = tlul_we | init_req;
   assign sram_intg_error = |bus_integ_error[2:1] & ~init_req;
   assign sram_addr       = (init_req) ? init_cnt          : tlul_addr;
@@ -496,7 +503,6 @@ module sram_ctrl
   // read-modify-write transaction in the SRAM adapter. In that case we force key_valid high to
   // enable that to complete so it returns a response, the TL gate won't accept any new
   // transactions and the SRAM keys have been clobbered already.
-  logic key_valid;
   assign key_valid =
     (key_req_pending_q)         ? 1'b0 :
     (reg2hw.status.escalated.q) ? (tl_gate_resp_pending & sram_rmw_in_progress) : 1'b1;
@@ -528,6 +534,11 @@ module sram_ctrl
     .raddr_o     ( ),
     .cfg_i
   );
+
+  logic unused_sram_gnt;
+  // Ignore sram_gnt signal to avoid creating a bad timing path, see comment on `tlul_gnt` above for
+  // more details.
+  assign unused_sram_gnt = sram_gnt;
 
   ////////////////
   // Assertions //
--- a/hw/ip/sram_ctrl/rtl/sram_ctrl.sv
+++ b/hw/ip/sram_ctrl/rtl/sram_ctrl.sv
@@ -445,6 +445,8 @@ module sram_ctrl
   logic [AddrWidth-1:0] sram_addr;
   logic [DataWidth-1:0] sram_wdata, sram_wmask, sram_rdata;
 
+  logic sram_rmw_in_progress;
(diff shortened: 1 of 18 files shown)


GregAC:
Got the block level synth running I've identified another feedthrough via the `error_internal` signal in `tlul_adapter_sram`: https://github.com/lowRISC/opentitan/blob/2feef77d3a04bb4281d70fdeff9c1cdf458b6320/hw/ip/tlul/rtl/tlul_adapter_sram.sv#L311

I think it'll be shallower feedthrough than the ones I've eliminated but best we get rid of them all. I can get rid of it but it will be a little more work, in particular may involve some extra DV work.

vogelpi:
As this second path is expected to be shallower and involves more DV work, should we first try to finalize the fix for the first and see how this impacts the actual chip timing?

GregAC:
In parallel I'm fixing up the CI issues for these current fixes so they should be good to go once they've been reviewed. Happy to trial how that impacts the chip timing alone but I would like to continue with this other fix. Long feedthroughs starting at Ibex going through the crossbar and then through SRAM and all the way back in Ibex generally aren't good. With them gone the implementation flow may find it has more placement freedom and wants to put in fewer buffers giving better QoR. In particular could give us some modest area savings that we'll appreciate having.

GregAC:
After a bit more investigation this new path may not be that much shallower than the previous ones anyway, so I think worth fixing it before doing a new implementation trial.

hw/ip/sram_ctrl/rtl/sram_ctrl.sv
```diff
@@ -489,13 +492,14 @@ module sram_ctrl
...
+  // If we have escalated, but there is a pending request in the TL gate, we may have a pending
+  // read-modify-write transaction in the SRAM adapter. In that case we force key_valid high to
+  // enable that to complete so it returns a response, the TL gate won't accept any new
   // transactions and the SRAM keys have been clobbered already.
   logic key_valid;
-  assign key_valid = (key_req_pending_q)         ? 1'b0 :
-                     (reg2hw.status.escalated.q) ? (tl_gate_resp_pending & tlul_we) : 1'b1;
+  assign key_valid =
+    (key_req_pending_q)         ? 1'b0 :
+    (reg2hw.status.escalated.q) ? (tl_gate_resp_pending & sram_rmw_in_progress) : 1'b1;
```
andreaskurth, line 508: Do we really need to take `tlul_we` or `sram_rmw_in_progress` into account for `key_vaiid` in the escalated case? As far as I can see, in the escalated case `key_valid` is only used to make `prim_ram_1p_scr` grant requests -- can't we make it grant all requests, including reads?

The reason I'm asking is because the change would be much simpler, and it would also mean less code to maintain

hw/ip/sram_ctrl/rtl/sram_ctrl.sv
```diff
@@ -478,9 +478,16 @@ module sram_ctrl
...
+
   // Interposing mux logic for initialization with pseudo random data.
   assign sram_req        = tlul_req | init_req;
-  assign tlul_gnt        = sram_gnt & ~init_req;
+  // This grant signal acts more like a ready internally in tlul_adapter_sram. In particular it's
+  // fine to assert it when tlul_req is low (it has no effect). So here tlul_gnt is asserted when
+  // a request from tlul_req will be granted regardless of whether a request exists. This is done
+  // for timing reasons so that the output TL ready isn't combinatorially connected to the incoming
+  // TL valid. In particular we must not use `sram_gnt` in the expression to avoid this.
+  assign tlul_gnt        = key_valid & ~init_req;
```
andreaskurth, line 490: This change LGTM

hw/ip/sram_ctrl/rtl/sram_ctrl.sv
GregAC: I guess it's a defence in depth? With this setup we've both got a TLUL gate and the clearing of `key_valid` from stopping a transaction acting on the SRAM after escalation. If `key_valid` is always true other than when `key_req_pending_q` is set then after an escalation any transaction that reaches the `tlul_sram_adapter` will get acted upon.

I note there's an explicit clone of the escalation signal, with one clone setting the escalated bit (which causes `key_valid` is go low) and the other clone closing the TLUL gate:

https://github.com/lowRISC/opentitan/blob/9f4903e77a8567a289e22d4c788822afb3a89cc6/hw/ip/sram_ctrl/rtl/sram_ctrl.sv#L164-L172

There's some discussion on these PRs: https://github.com/lowRISC/opentitan/pull/11863 and https://github.com/lowRISC/opentitan/pull/17072

There's also the argument we want to change behaviour as little as possible.

andreaskurth: Thanks for the explanation. I think defense in depth and changing behavior as little as possible are sufficient arguments for making the change this way

andreaskurth (approved):
LGTM based on the discussions in this PR. Thanks @GregAC!

vogelpi:
Please don't merge this yet, AFAIK Greg found another path he is currently working on improving.

GregAC:
Yes I've got two more fixes inbound, currently investigating some regression failures and then will update this PR with them.

andreaskurth:
Since we are agreed on this, we could just merge this one for now and create a follow-up PR for the other improvements (tracked against #22462)?

GregAC:
Regression issues I'm looking at may be related to these changes so I'd prefer to at least understand those before we merge this.

GregAC:
Failures look to be related to my new work so merging this now

Issue #22271, opened 2024-03-26
[dv,usbdev] USBDEV DV improvements/suggestions


alees24:
### Description

A container for proposed improvements to the USBDEV block level DV effort:

- [x] Modify the usb20_driver and usb20_monitor to operate on a higher clock frequency; attempting to operate on a 12MHz clock and decode 12Mbps data is fraught with issues and it works presently only because the DV code steals information from the usbdev output driver enable, which happens to be asserted synchronously with the first output bit.
- [x] Operate on a clock frequency that is independent of the usbdev clock and may thus be varied with respect to that clock, to verify operation under conditions of frequency and phase mismatch.
- [x] Decode data transmitted by the usbdev using only the USB_DP/DN signals themselves, responding to the deviation from Idle (J) state.
- [x] Collect the various sections of code that prefix 'SYNC' pattern onto transmitted packets into a single function, rather than replicate the code for each packet type in usb20_driver.
- [x] Make it possible to transmit packets that have an invalid or incomplete SYNC prefix.
- Decoding of response packets in the usb20_driver:
  - [x] Be robust against the absence of a packet; often this is the correct behavior according to the protocol specification.
  - [x] Make zero assumptions about the received data, particularly with respect to validity and length, duration of EOP etc.
  - [x] Calculate and check the CRC16 values of received Data packets. 
- [x] usb20_monitor needs to be robust against packets not matching expectations; presently there are errors when attempting to cast the raw bit data to the packet types `sof_p`, `token_p` etc if the bit data does not have exactly the expected length.
- [x] usb20_monitor does not check the EOP (End Of Packet), but rather assumes its validity.
- [x] data_pkt.set_payload() constructs the SETUP DATA packet incorrectly; Least Significant Bytes (eg. wVL) should go first because USB is Little Endian, but it should probably also be accepting the values as 16-bit quantities.
- [x] usb20_monitor does not handle only a single bit interval of Idle between packets; this admittedly should not happen (specification requires at least two bit intervals of J between EOP and start of a new packet) but the usb20_monitor ought to cope.
- [x] usb20_monitor, like the usb20_driver treats Bus Reset as a special event to be detected at the start of a sequence, rather than something that may occur at any time; similarly it assumes that the departure from Idle state indicates a packet, and will not detect Bus Resets, Resume Signaling, Disconnects (VBUS dropping) and so forth.

@rswarbrick

alees24:
[...]

PR #22498, opened 2024-04-09
[dv,usbdev] Clocking changes to usbdev block level DV

alees24:
This PR proposes a number of improvements to the block level DV for USBDEV:

- Modify the frequency of the AON clock to better match the real use case, and more importantly prevent spurious reports of events such as bus resets; the design assumes that it can simply count clock cycles.
- Separate the clock/reset of the host from that of the DUT.
- Introduce a variable frequency 4x oversampling clock for the usbdpi and usb20_agent.
- Introduce the `usbdev_aon_wake` module since this is verified only at top level thus far.
- Modify the usb20_driver and usb20_monitor to operate on the new clock, and lose the clock division to 12MHz because sampling at such low frequencies causes issues and makes frequency tracking infeasible.

Related to #22271


hw/ip/usbdev/dv/env/usbdev_env_cfg.sv
```diff
@@ -4,33 +4,52 @@
...
-  // speed. (This isn't as big a ratio as we'll have on the chip, but the tests are a bit more
-  // efficient when the clocks are similar speeds)
-  rand uint aon_clk_freq_mhz;
-  constraint aon_clk_freq_mhz_c {
-    aon_clk_freq_mhz >= usb_clk_freq_mhz / 3 &&
-    aon_clk_freq_mhz <= usb_clk_freq_mhz / 2;
+  // Constrain the AON clock to be slower than the USB clock. For the usbdev_aon_wake logic
+  // to operate correctly and not generate spurious reports of bus resets when detecting Low Speed
+  // signaling, it requires a clock frequency lower than 2MHz cf the usbdev clock at 48MHz, so the
+  // ratio should exceed 24.
```
a-will, line 26: Now I am curious. This would be a case where it's supposed to be a K to resume, but a low-speed EOP has somehow made its way to our device instead, right?

It's illegal for a hub to drive / forward anything when the port is suspended (and only K or SE0 in follow-up states), but this seems fine to prevent us from simulating something nonsensical.

a-will (approved):
LGTM

PR #22499, opened 2024-04-09
[dv,prim] Clarification of reset behavior

alees24:
Clarify behavior when a single domain is reset; the auto-generated register logic appears to be robust against the spurious pulses generated by the use of 'prim_pulse_sync' and 'prim_sync_reqack' (NRZ mode) during a single-domain reset, provided that there is no in-progress transaction (register read/write from SW).

Correct a couple of typographical errors.

It is perhaps worth also noting that the `SrcAckBusyChk_A` assertion will fire if the source domain is reset after an odd number of SW accesses to the CDC register, which could perhaps be prevented by changes either to the logic (now is perhaps not the time given the chip-wide impact) or the use of more-involved assertion checking that attends to reset behavior.

### DIFF
--- a/hw/ip/prim/rtl/prim_pulse_sync.sv
+++ b/hw/ip/prim/rtl/prim_pulse_sync.sv
@@ -7,6 +7,9 @@
 // cycle of its respective clock domain. Consecutive pulses need to be spaced
 // appropriately apart from each other depending on the clock frequency ratio
 // of the two clock domains.
+//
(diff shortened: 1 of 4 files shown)


vogelpi (approved):
LGTM, thanks for the clarifications and fixes @alees24 .

I agree that we shouldn't change the behavior or assertions here at this point unless necessary.

PR #22500, opened 2024-04-09
[pinmux/doc]: Fix the description for the keeper_en register

kupiakos:
It previously copied the pull_en description, which is incorrect. I consulted with hweng on what a correct description would be.

Cc: @moidx @msfschaffner


kupiakos:
Looks like the kokoro check is failing. I suspect that the failure is due to the `Warning: License will expire in 6 days` flow warnings, rather than an actual linting violation.

hcallahan-lowrisc:
Yes the ascentlint failure is unrelated, it doesn't check .hjson/.md files.

PR #22502, opened 2024-04-10
[ci] increase SiVal ROM_EXT test timeout

timothytrippel:
This was timing out in CI.


moidx (approved):

a-will:
I'd wager the machine _frank_ is just overloaded when running 8 simultaneous jobs, where some of those jobs are spawning many heavier tasks during the build phase, competing for CPU time with the FPGA test tasks.

If that's actually the case, this would just be a patch, but it'll do for now. ;)

PR #22504, opened 2024-04-10
[fpv] Minor tidy-ups to SecCm FPV code

rswarbrick:
No big change, but this makes the code a bit easier to follow. (It took me a while to understand!). Proper comments on each different commit.

The only bug fix is in the second commit ("[sram_ctrl,fpv] Fix incorrect assertion") where the assertion wasn't strong enough (and I only noticed because I was going through the code trying to understand everything).


vogelpi (approved):

PR #22506, opened 2024-04-10
[entropy_src,rtl] Factor sfifo errors into fatal alert

rswarbrick:
These signals come from:

- u_prim_fifo_sync_distr (the distribution FIFO)
- u_prim_fifo_sync_esrng (the RNG bus input FIFO)
- u_prim_fifo_sync_observe
- u_prim_fifo_sync_esfinal

when one sees an internal error. We want any of these errors to be transformed into a fatal alert.

### DIFF
--- a/hw/ip/entropy_src/rtl/entropy_src_core.sv
+++ b/hw/ip/entropy_src/rtl/entropy_src_core.sv
@@ -983,7 +983,9 @@ module entropy_src_core import entropy_src_pkg::*; #(
   end : gen_err_code_test_bit
 
   // alert - send all interrupt sources to the alert for the fatal case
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
Thanks for working on this and fixing it @rswarbrick !

For background: these errors got already factored into `event_es_fatal_err` so lead to a `fatal_alert_o`. However, `event_es_fatal_err` is kept low while ENTROPY_SRC is disabled. This may be desirable for some error types but it's not for FI countermeasures like the hardened FIFO counters. And that's why FPV complained in this case. But I agree, we should always flag this also if the module is disabled. LGTM!

rswarbrick:
Thanks for the explanation: that makes sense to me.

PR #22507, opened 2024-04-10
[sival] Add `sram_ctrl_lc_escalation` test

jwnrt:
This test:

1. Reads and writes to main and retention SRAMs from ibex and through the debug module.
2. Triggers a lifecycle controller bus integrity error alert which is configured to escalate to locking the SRAMs.
3. Tries to read and write the SRAMs through the debug module again which should fail.
4. Resets the chip.
5. Tries to read and write the SRAMs again which should pass.

We can't check that the SRAMs lock up from within ibex because ibex also locks up at the same escalation level.

Instead, I have moved the test from the PROD to RMA lifecycle state and tried to access the SRAMs through the debug module.

There's a chance for a false positive with this test where actually the debug module (or something else) locks up / breaks during the escalation and not the SRAMs. I can't think of any way around that, though.

### DIFF
--- a/sw/host/tests/chip/sram_ctrl/src/sram_ctrl_lc_escalation.rs
+++ b/sw/host/tests/chip/sram_ctrl/src/sram_ctrl_lc_escalation.rs
@@ -1,0 +1,143 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! This is the test harness for the `sram_ctrl_lc_escalation` test which
+//! checks that SRAMs can no longer be accessed after an alert is escalated.
+//!
+//! We have to check SRAM access over the debugger and not from within Ibex
+//! because the core also locks up on escalation.
+
+use std::fs;
+use std::path::PathBuf;
+use std::time::Duration;
+
+use anyhow::{ensure, Context, Result};
+use clap::Parser;
+
+use opentitanlib::app::TransportWrapper;
+use opentitanlib::execute_test;
+use opentitanlib::io::jtag::{Jtag, JtagTap};
+use opentitanlib::io::uart::Uart;
+use opentitanlib::test_utils;
+use opentitanlib::test_utils::init::InitializeTest;
+use opentitanlib::test_utils::mem::{MemRead32Req, MemWriteReq};
+use opentitanlib::uart::console::UartConsole;
+
+#[derive(Debug, Parser)]
+struct Opts {
+    #[command(flatten)]
+    init: InitializeTest,
+
+    /// Console receive timeout.
+    #[arg(long, value_parser = humantime::parse_duration, default_value = "600s")]
+    timeout: Duration,
+
+    /// Path to the ELF file being tested on the device.
+    #[arg(long)]
+    firmware_elf: PathBuf,
+}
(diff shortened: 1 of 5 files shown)


jwnrt:
cc @nbdd0121 - have I used the debug test utils correctly here?

sw/host/tests/chip/sram_ctrl/src/sram_ctrl_lc_escalation.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    log::info!("-------------------------------");
+    log::info!("EXPECTING JTAG ACCESSES TO FAIL");
+    log::info!("-------------------------------");
+
+    // Check that we can no longer access the SRAMs after escalation.
+    // These operations should time out.
+    if write_read(&mut *jtag, main_addr, 0xcccc_cccc).is_ok() {
+        bail!("expected main SRAM access to fail");
+    }
```
nbdd0121, line 109:
```suggestion
    ensure!(write_read(&mut *jtag, main_addr, 0xcccc_cccc).is_err(), "expected main SRAM access to fail");
```

sw/host/tests/chip/sram_ctrl/src/sram_ctrl_lc_escalation.rs
```diff
@@ -0,0 +1,141 @@
...
+    if write_read(&mut *jtag, ret_addr, 0xdddd_dddd).is_ok() {
+        bail!("expected retention SRAM access to fail");
+    }
+
+    // Reset the chip and try again - the SRAMs should now be unlocked.
+    Jtag::disconnect(jtag)?;
```
nbdd0121, line 115:
```suggestion
    jtag.disconnect()?;
```

jwnrt: I didn't realise you could change the type of `self` like this and still use `.function` syntax, thanks

sw/host/tests/chip/sram_ctrl/src/sram_ctrl_lc_escalation.rs
```diff
@@ -0,0 +1,143 @@
...
+    // Check that we can no longer access the SRAMs after escalation.
+    // These operations should time out.
+    ensure!(
+        write_read(&mut *jtag, main_addr, 0xcccc_cccc).is_err(),
+        "expected main SRAM access to fail"
+    );
+    ensure!(
+        write_read(&mut *jtag, ret_addr, 0xdddd_dddd).is_err(),
+        "expected retention SRAM access to fail"
+    );
```
pamaury, line 114: Is it worth checking using register that there was an escalation alert?

jwnrt: Good idea

Do you mean trying to read the alert handler's registers to see whether the alert was triggered, or just trying to read any generic register or CSR to make sure the debugger still works?

pamaury: Read the alert handler's register

jwnrt: This turned out not to be as easy as I expected, do you mind if I do it in a follow-up PR?

PR #22520, opened 2024-04-11
[docker] use CI verilator build in docker instead of custom apt source

nbdd0121:


nbdd0121:
No, it's not tested by the CI. You can do a local build with `docker build -t opentitan -f util/container/Dockerfile .`, and then `docker run -it opentitan bash` to enter it and see `verilator` is indeed available in the path.

jwnrt (approved):
Thanks a lot, that works. Using the version from CI seems more reliable.

LGTM

andreaskurth (approved):
LGTM (though I didn't run it through Docker myself), thx Gary

jwnrt:
@nbdd0121 was the container image rebuilt and uploaded to the store after this change? It seems we're still depending on this repository and the keys have expired, causing CI failure

PR #22521, opened 2024-04-11
[ci] Mark csrng_edn_concurrency_test_fpga_cw310_sival broken

jwnrt:
All other variants are marked broken but this one seems to have been missed.


jwnrt:
Closing now that we have https://github.com/lowRISC/opentitan/pull/22523 which fixes the test.

PR #22522, opened 2024-04-11
[sival] Re-enable UART Baud test and disable high Bauds

jwnrt:
This test was disabled in CI due to problems with the HyperDebug which appear to be fixed now.

This PR also removes the two highest Baud rates (1MBd and 1.5MBd) since they're too high for us to test. Previously they were only enabled in Silicon (where the clock was thought to run fast enough) but it seems it's either not fast enough or HyperDebug doesn't go that fast.

Open question: should I change the test plan to only require the lower Baud rates? @rswarbrick is this an acceptible change to the test?

### DIFF
--- a/sw/device/tests/uart_baud_rate_test.c
+++ b/sw/device/tests/uart_baud_rate_test.c
@@ -24,16 +24,11 @@ static const uint32_t kBaseAddrs[4] = {
     TOP_EARLGREY_UART2_BASE_ADDR,
     TOP_EARLGREY_UART3_BASE_ADDR,
 };
-static uint32_t kBauds[7] = {
-    9600, 115200, 230400, 128000, 256000, 1000000, 1500000,
-};
+static const uint32_t kBauds[9] = {4800,   9600,   19200,  38400, 57600,
+                                   115200, 230400, 128000, 256000};
 
 enum {
   kTestTimeoutMillis = 500000,
-  // Number of Bauds to test from `kBauds`.
-  kBaudCountSilicon = 7,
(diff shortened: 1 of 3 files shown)


sw/device/tests/uart_baud_rate_test.c
```diff
@@ -24,16 +24,10 @@ static const uint32_t kBaseAddrs[4] = {
     TOP_EARLGREY_UART2_BASE_ADDR,
     TOP_EARLGREY_UART3_BASE_ADDR,
 };
-static uint32_t kBauds[7] = {
-    9600, 115200, 230400, 128000, 256000, 1000000, 1500000,
-};
+static uint32_t kBauds[5] = {9600, 115200, 230400, 128000, 256000};
```
engdoreis, line 27: Could you make it `const` to justify the name pattern?
I think that we should test at least a sub set of the standard baud `4800, 9600, 19200, 38400, 57600, 115200, 230400, 460800,` Hyperdebug should be able to handle it.

engdoreis (commented):
LGTM. I only have a suggestion, though.

sw/device/tests/uart_baud_rate_test.c
jwnrt: I've added most of these, except `460800` was rejected by the DIF so I've left that one out.

engdoreis (approved):

jesultra:
FYI HyperDebug firmware has recently been enhanced to be able to support sustained 1.5Mbaud UART communication (see PR #23379).  Maybe the high-speed tests can be re-enabled.

PR #22523, opened 2024-04-11
[csrng,sival] Fix `csrng_edn_concurrency`

pamaury:
The problem is that recoverable alert (`ACMD_FLAG0_FIELD_ALERT`) is triggered *before* the test, but the test does not clear it before starting. This is problem triggered in the ROM_EXT or the ROM, and it happens on both master and es_sival.

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -928,20 +928,16 @@ opentitan_test(
 opentitan_test(
     name = "csrng_edn_concurrency_test",
     srcs = ["csrng_edn_concurrency_test.c"],
-    broken = cw310_params(tags = ["broken"]),
     exec_env = dicts.add(
         EARLGREY_TEST_ENVS,
         {
             "//hw/top_earlgrey:fpga_cw310_sival": None,
-            "//hw/top_earlgrey:fpga_cw310_sival_rom_ext": "broken",
+            "//hw/top_earlgrey:fpga_cw310_sival_rom_ext": None,
             "//hw/top_earlgrey:silicon_creator": None,
-            "//hw/top_earlgrey:silicon_owner_sival_rom_ext": "silicon_owner",
-            "//hw/top_earlgrey:silicon_owner_prodc_rom_ext": "silicon_owner",
-            "//hw/top_earlgrey:silicon_owner_proda_rom_ext": "silicon_owner",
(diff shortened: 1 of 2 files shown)


jwnrt (approved):
Really nice find, if this passes CI then LGTM

vogelpi (approved):
Nice work, thanks for fixing this!

Out of interest, which alert did fire? I suspect its the `CS_BUS_CMP_ALERT` because in case of `CS_MAIN_SM_ALERT` the module would need to be restarted which this PR doesn't do. The `CS_BUS_CMP_ALERT` can fire when doing some known answer testing on the software application port. I suspect ROM_ext might do something like that.

pamaury:
> Nice work, thanks for fixing this!
> 
> Out of interest, which alert did fire? I suspect its the `CS_BUS_CMP_ALERT` because in case of `CS_MAIN_SM_ALERT` the module would need to be restarted which this PR doesn't do. The `CS_BUS_CMP_ALERT` can fire when doing some known answer testing on the software application port. I suspect ROM_ext might do something like that.

It is `ACMD_FLAG0_FIELD_ALERT`.

vogelpi:
I think I've found the issue. Cryptolib configures EDN0 and EDN1 with `kHardenedBoolFalse` for flag0.
```
sw/device/lib/crypto/include/freestanding/hardened.h
36:  kHardenedBoolFalse = 0x1d4,
```
But acceptable values are in fact `kMultiBitBool4True` and `kMultiBitBool4False` only.

pamaury:
The issue is [here](https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/crypto/drivers/entropy.c#L418):
```c
if (launder32(cmd.disable_trng_input) == kHardenedBoolTrue) {
    reg = bitfield_field32_write(reg, kAppCmdFieldFlag0, kMultiBitBool4True);
  }
```
If `disable_trng_input` is false, then `flag0` is left to zero instead of `kMultiBitBool4False`. I will create a PR for that.

What should we do about this PR? I think it's still good practice to clear the recoverable error before the test?

vogelpi:
I've flagged an issue here #22524.

vogelpi:
Yes, I think clearing the alerts doesn't hurt especially as the test only passes if no alert is triggered :-)

PR #22525, opened 2024-04-11
[cryptolib,entropy] Fix `csrng_send_app_cmd`

### LINKED ISSUE #22524: [cryptolib] Correct flag0 values for CSRNG and EDN configurations
### Description

While debugging ENTROPY_SRC/CSRNG test failures, it has been discovered that CSRNG triggeres a recoverable alert and sets the ACMD_FLAG0_FIELD_ALERT alert status bit during ROM_ext.

The reason for this is most likely that cryptolib doesn't configure the EDNs correctly. More precisely, inside `entropy_complex_init()` which is called in ROM_ext, the entropy complex is configured with the settings defined in `kEntropyComplexConfigIdContinuous` (see `entropy.c`).

I am not exactly sure where the misconfiguration happens. I've found two things so far:

1. The configuration struct has configurations for both EDN0 and EDN1. For EDN1, there the `disable_trng_input` struct member for the Generate command is not defined (it's defined for the EDN0 Generate command though.):
```c
                .edn1 =
                    {
                        .base_address = kBaseEdn1,
                        .reseed_interval = 4,
                        .instantiate =
                            {
                                .id = kEntropyDrbgOpInstantiate,
                                .disable_trng_input = kHardenedBoolFalse,
                                .seed_material = NULL,
                                .generate_len = 0,
                            },
                        .generate =
                            {
                                .id = kEntropyDrbgOpGenerate,
                                .seed_material = NULL,
                                .generate_len = 1,
                            },
                        .reseed =
                            {
                                .id = kEntropyDrbgOpReseed,
                                .disable_trng_input = kHardenedBoolFalse,
                                .seed_material = NULL,
                                .generate_len = 0,
                            },
                    },
```
2. `ENTROPY_CMD` macro leaves the flag0 field for the register empty unless `kHardenedBoolTrue` true is provided via the above struct.
```c
  if (launder32(cmd.disable_trng_input) == kHardenedBoolTrue) {
    reg = bitfield_field32_write(reg, kAppCmdFieldFlag0, kMultiBitBool4True);
  }
```
So 0 is written which is not a valid MuBi4Value expected by CSRNG.

pamaury:
When asked to enable the TRNG input (disable_trng_input = false), the code would incorrectly set `flag0` in the command to 0 instead of `kMultiBitBool4False` which triggers an alert.

Fixes #22524

### DIFF
--- a/sw/device/lib/crypto/drivers/entropy.c
+++ b/sw/device/lib/crypto/drivers/entropy.c
@@ -414,6 +414,7 @@ static status_t csrng_send_app_cmd(uint32_t base_address,
   reg = bitfield_field32_write(0, kAppCmdFieldCmdId, cmd.id);
   reg = bitfield_field32_write(reg, kAppCmdFieldCmdLen, cmd_len);
   reg = bitfield_field32_write(reg, kAppCmdFieldGlen, cmd.generate_len);
+  reg = bitfield_field32_write(reg, kAppCmdFieldFlag0, kMultiBitBool4False);
 
   if (launder32(cmd.disable_trng_input) == kHardenedBoolTrue) {
     reg = bitfield_field32_write(reg, kAppCmdFieldFlag0, kMultiBitBool4True);


vogelpi (approved):
LGTM, thanks for the quick fix!

jadephilipoom (approved):
Nice catch!

ballifatih (approved):
LGTM.

PR #22528, opened 2024-04-11
[rv_plic,fpv] Fix failing FpvSecCmBusIntegrity_A assertion

rswarbrick:
This was... surprisingly tricky to get right! There's a long comment above the assertion that explains how to read it.

Once merged, this should hopefully fix the TOP_EARLGREY_IP_FPV Formal FPV regression (current results [here](https://reports.opentitan.org/hw/top_earlgrey/formal/ip/summary/latest/report.html))

### DIFF
--- a/hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic.sv
+++ b/hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic.sv
@@ -424,11 +424,30 @@ module rv_plic import rv_plic_reg_pkg::*; #(
   // Assume
   `ASSUME(Irq0Tied_A, intr_src_i[0] == 1'b0)
 
- // RV_PLIC does not have a block-level DV environment, hence we add an FPV assertion to test this.
+  // This assertion should be provable in FPV because we don't have a block-level DV environment. It
+  // is trying to say that any integrity error detected inside the register block (u_reg) will cause
+  // an alert to be asserted within at most _SEC_CM_ALERT_MAX_CYC cycles.
+  //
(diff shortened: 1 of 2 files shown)


hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic.sv
```diff
@@ -424,11 +424,30 @@ module rv_plic import rv_plic_reg_pkg::*; #(
...
+  //    must go high". To encode this cleanly in SVA, we actually say "We can't have neither an
+  //    integrity error nor an alert signal for too many cycles".
   `ASSERT(FpvSecCmBusIntegrity_A,
-          $rose(u_reg.intg_err)
+          ($rose(u_reg.intg_err) &&
+           gen_alert_tx[0].u_prim_alert_sender.state_q == gen_alert_tx[0].u_prim_alert_sender.Idle)
           |->
-          ##[0:`_SEC_CM_ALERT_MAX_CYC] (alert_tx_o[0].alert_p))
+          not ((!gen_alert_tx[0].u_prim_alert_sender.sigint_detected && !alert_tx_o[0].alert_p)
+               [*`_SEC_CM_ALERT_MAX_CYC]))
```
vogelpi, line 450: This is really hard to understand, your comment above is very helpful. Thanks for adding it!

I am no expert with assertions and their syntax. Did you check that this assertion really fires if the design is hacked such that the alert doesn't fire early enough?

vogelpi (approved):
Thanks @rswarbrick , this looks sensible but I am no SVA expert. I think it would be worth to check that the SVA correctly fires if the alert doesn't raise in time (e.g. by hacking the design up locally).

PR #21226, opened 2024-02-06
[flash_ctrl] Change FIFO interrupts to status type

msfschaffner:
The first two commits patch up common infrastructure:
1) The clear_all_interrupts task in the `cip_base_vseq` needs to be adjusted so that it does not fail when status interrupts remain asserted (since they cannot be cleared via a write to the `INTR_STATE` register.
2) Another optional field `default` is added to the interrupt declaration in the IP Hjson file. This allows us to define what the default expectation of the interrupt should be during automatically generated tests such as the `chip_plic_all_irqs` top-level test. That test is aligned to treat status interrupts differently now. In particular, the ones that are asserted by default will not be enabled until that specific interrupt is being tested.

The last two commits are the changes in `flash_ctrl`, aligning the interrupt type behavior according to #21211.

This fixes #21211

PR #22530, opened 2024-04-11
Cherrypick to earlgrey_es_sival: [topgen] Add field to specify status IRQ default behavior

nbdd0121:
Cherry-picks a single commit in #21226 which seems to fix the rv_plic sival tests on silicon.


pamaury (approved):
This change looks ok. I don't know quite enough to be sure but I assume that the change to `top_earlgrey.gen.hjson` is just a "documentation" change, ie it doesn't actually change any RTL file?

engdoreis:
The commit matches the original PR.

engdoreis (approved):

PR #22534, opened 2024-04-11
[rv_plic,fpv] Fix index widths

rswarbrick:
No functional change really, but it silences some warnings that come out of Jasper at build time (because the type of the index is a 32-bit variable and it reasonably worries that this might mean we're going to to an out-of-bounds access).


vogelpi (approved):
This makes sense, thanks @rswarbrick !

Issue #8120, opened 2021-09-09
[keymgr] Security enhancement


tjaychen:
- for the random values used to wipe keys in key manager, we can consider additional permutation on one of the shares to ensure the values are diverse.

msfschaffner:
I wasn't sure whether this is supposed to be an enhancement, or something that we need to do - hence I labeled for D2S.
Please relabel if this is inaccurate.

tjaychen:
yeah this is meant to be marked as a future enhancement, since i didn't want to alter the existing keymgr more.

ballifatih:
Triaged for `keymgr`. It is unclear what could go wrong if we do not implement randomization. Temporarily marking this as `Type:FutureRelease`.

ballifatih:
For the first case (inside `keymgr_ctrl`), I am not sure I understood the question exactly. Are you suggesting that we have a secret netlist constant value, and reset the value of state_d to this fixed value instead of '0 vector? The reset value of state_d (which is effectively '0) lingers until the first advance command is executed (which latches the OTP root key).

Otherwise, as you also pointed out, because of the XORing behavior, we need to load equal-share randomness. If we do not load equal-value shares, then after XORing, we would not get the correct key value (i.e. lose the value of OTP root key).  Therefore the effective key value of pre-populated randomness must be '0. This value lingers in state_d for few clock cycles (8 clock cycles, if EDN/LFSR is not blocking), so I do not think it would be a big issue. No SW interaction is required for switching from this randomness to OTP-loaded state_d value.

For the second one (inside `keymgr_kmac_if`), I think this value is likely to be propagated to KMAC as the key value. I agree that it would be better if the effective key behind the two shares is also random. The easiest solution is to use `entropy_i[0]` and `entropy_i[1]` respectively for two shares. This is still not a very good solution because the same randomness `entropy_i[1]` would be shared for two different purposes.

Another point about these relatively small entropy is that if KMAC can be invoked with these random keys, it would leak the state of internal LFSR to some point. 32-bit space is small enough that we can brute-force through it.

vogelpi:
> For the first case (inside `keymgr_ctrl`), I am not sure I understood the question exactly. Are you suggesting that we have a secret netlist constant value, and reset the value of state_d to this fixed value instead of '0 vector? The reset value of state_d (which is effectively '0) lingers until the first advance command is executed (which latches the OTP root key).
> 
[...]

PR #22535, opened 2024-04-11
[keymgr] Use non-deterministic decoy values for KMAC and sideload ports

vogelpi:
Previously, the two shares of the masked values forwarded to KMAC and the sideload ports would be equal when not used. This is non-ideal as the effective value observed on the sideload ports and by KMAC becomes then 0 which may potentially be exploited.

This resolves lowRISC/OpenTitan#8120.


andreaskurth (approved):
LGTM, thx @vogelpi

PR #22539, opened 2024-04-12
[pentest] Add new alert handler to tests

nasahlpa:
This commit adds the new alert handler format to the Ibex FI tests.


engdoreis:
Merging this PR since ci just failed due to a flaky test.

Issue #20747, opened 2024-01-02
[rom_ext,csrng] Fix rom_ext for pending csrng interrupt


matutem:
### Description

The test //sw/device/tests/autogen:plic_all_irqs_test_0 fails for both fpga_cw310_sival_rom_ext and silicon_owner_sival_rom_ext. 

The failure is for peripheral 19 (kTopEarlgreyPlicPeripheralAdcCtrlAon): rather than 19 the test gets peripheral 27 (kTopEarlgreyPlicPeripheralCsrng), and interrupt 174 (kTopEarlgreyPlicIrqIdCsrngCsEntropyReq).

Apparently the interrupt happens right after peripherals_irqs_enable https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/tests/autogen/plic_all_irqs_test.c;drc=e5b074b89ee185a45a1008415cc6839a14a033ad;l=2111 , since a call to LOG_INFO after that and before peripheral_irqs_trigger is not run. It seems there is a pending interrupt from csrng upon test entry, so I suspect the rom_ext leaves it pending.

ADC is the first peripheral in the test, so it makes sense it would be the one to fail, and the fact that both fpga_cw310 and silicon running with rom_ext fail the same way suggests it is a rom_ext issue.

pamaury:
Can you check if #22525 fixes this issue? I am not entirely sure it's related but it's worth a try.

vogelpi:
The root cause for this bug is most likely #22877 (similar as for #22819).

vogelpi:
I just re-ran the test on the FPGA `//sw/device/tests/autogen:plic_all_irqs_test_0_fpga_cw310_sival_rom_ext`. It's working both on `master` (622018b25190bc6129f6495edc98dec642b6c037) and `earlgrey_es_sival` (7f1ef1c112fbdb0e5b3dab8030044a4c21d2c946). It seems that the keymgr related fixes linked in the previous post indeed solve this issue.

I am closing this issue.

vogelpi:
FYI: @nasahlpa also check the test on silicon where it's working now as well :-)

PR #22543, opened 2024-04-12
Fix entropy src csrng test

pamaury:
See commits for explanations.

### DIFF
--- a/sw/device/tests/entropy_src_csrng_test.c
+++ b/sw/device/tests/entropy_src_csrng_test.c
@@ -118,13 +118,7 @@ static void plic_interrupts_enable(void) {
  * `ottf_external_isr()` routine.
  */
 static void irq_block_wait(irq_flag_id_t isr_id) {
-  // The interrupt can come before we enter sleep, so we check beforehand.
-  while (true) {
-    if (irq_flags[isr_id]) {
-      break;
-    }
-    wait_for_interrupt();
-  }
+  ATOMIC_WAIT_FOR_INTERRUPT(irq_flags[isr_id]);
   switch (isr_id) {
     case kTestIrqFlagIdCsrngEntropyReq:
       CHECK_DIF_OK(dif_csrng_irq_set_enabled(&csrng, kDifCsrngIrqCsEntropyReq,
--- a/sw/device/tests/entropy_src_csrng_test.c
+++ b/sw/device/tests/entropy_src_csrng_test.c
@@ -248,16 +248,26 @@ static void test_edn_cmd_done(const dif_edn_seed_material_t *seed_material) {
 
   plic_interrupts_enable();
 
-  // Generate enough entropy for the otbn randomness test.
-  // The len provided here is in number of words as opposed to num of 128b
-  // blocks. The requested len **must** also be equal to the amount of entroy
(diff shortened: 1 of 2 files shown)


sw/device/tests/entropy_src_csrng_test.c
```diff
@@ -248,16 +248,26 @@ static void test_edn_cmd_done(const dif_edn_seed_material_t *seed_material) {
...
+  // more otherwise the Generate command will
   // not be fully executed, causing a hang in the `irq_block_wait()` calls
-  // following the end of the OTBN test.
-  // EDN0: 20 words = 5x128b blocks
-  // EDN1: 44 words = 11x128b blocks
+  // following the end of the OTBN test. The OTBN test reads `RND` 5 times
+  // and each read consumes 256b of entropy. The EDN1 generates entropy per
+  // blocks of 128b so we need to generate 10 blocks.
+  const size_t kEdnBlockSizeBits = 128;     // Each EDN block contains 128b.
+  const size_t kOtbnRequestSizeBits = 256;  // Each OTBN request requires 256b.
```
jwnrt, line 261: As far as I can see these are always converted to words - maybe it would be better to define them like that?

```suggestion
  const size_t kEdnBlockSizeWords = 4;     // Each EDN block contains 128b.
  const size_t kOtbnRequestSizeWords = 8;  // Each OTBN request requires 256b.
```

pamaury: I would rather avoid this because the OTBN code can request either words or side register and each requests uses 256b independently of the request (ie request a word of entropy will consume 256b). So it's really the number of requests that counts.

vogelpi (approved):

sw/device/tests/entropy_src_csrng_test.c
jwnrt: Hmm okay, fair enough

sw/device/tests/entropy_src_csrng_test.c
```diff
@@ -243,6 +238,7 @@ static void edn_configure(const dif_edn_t *edn, irq_flag_id_t irq_flag_id,
  */
 static void test_edn_cmd_done(const dif_edn_seed_material_t *seed_material) {
   CHECK_STATUS_OK(entropy_testutils_stop_all());
+  CHECK_DIF_OK(dif_csrng_clear_recoverable_alerts(&csrng));
```
moidx, line 241: Do we now what recoverable alerts are been triggered?

moidx (approved):

sw/device/tests/entropy_src_csrng_test.c
pamaury: Yes, I fixed it in https://github.com/lowRISC/opentitan/pull/22525 but the SiVal ROM_EXT is signed so at least this avoid releasing a new signed ROM_EXT just for that.

moidx: Thanks!

Does this somehow also help with this interrupt https://github.com/lowRISC/opentitan/issues/20747? (I suspect this is a separate issue).

pamaury: It could be related, I'll add comment there.

Issue #22443, opened 2024-04-05
[rtl,flash_ctrl] Fix scrambler interface


matutem:
### Description

The fix for the flash_ctrl scrambler interface is not complete, as we still get some unexpected ecc failures in the flash_ctrl_phy_arb_redun unit test. Specifically, running it with the fix in #22444 with seed 18977237425298456515035329733300663925435727379005790363212818518281951604587 the test shows an ECC error for a read to address 0x08a0 from bank 0 around time 269437 x 100 ps.

From the test waves, the data from the flash primitive and the memory backdoor read is 0xcdf_0523ab48_aa8035de (the top 3 nibbles are both eccs), and the lower 64 bits are never sent to the descrambler, as you can see from https://github.com/lowRISC/opentitan/blob/master/hw/ip/flash_ctrl/rtl/flash_phy_core.sv#L554. 

It can be helpful to add instrumentation in the descramble function at https://github.com/lowRISC/opentitan/blob/master/hw/ip/flash_ctrl/dv/env/flash_otf_item.sv#L173 to show the raw scrambled data, the ecc, and the descrambled data.
![Screenshot from 2024-04-04 23-50-19](https://github.com/lowRISC/opentitan/assets/80294896/139613f0-7963-4e90-ac3a-046c0b2560a4)

vogelpi:
Thanks for filing the issue @matutem ! I've now investigated this a bit and noted the following:
- Your branch `[flash_ctrl_failures](https://github.com/matutem/opentitan/tree/flash_ctrl_failures)` doesn't include the latest fix to the scrambling.
- When rebasing the your branch on current master, I still see a failure with the same seed, but it seems to be slightly different.
- I also see an unexpected ECC error being generated. It's generated for a controller request. To arbitrate between controller and host requests, there is another `prim_arbiter_tree_dup` instance in the design. This one has been there before sharing the scrambling unit and it's the one that the test actually forces.
- The transaction producing the ECC error goes into the scrambling arbitration right after releasing the force signal on the controller / host arbiter. 

I am not yet sure what exactly happens and why the ECC error is produced. It seems to be triggered by the test. Maybe the forcing violates one of the assumptions stated in the design for controller requests (no back-to-back requests, only forwarded if the pipeline is completely empty), or the injected fault / signal force leaves the read pipeline in a weird state. But I think we should fix the RTL here. I'll continue my investigation and report back.

matutem:
[...]

PR #22549, opened 2024-04-12
[dv,flash_ctrl] Create test the generates ecc faults

matutem:
The flash_ctrl_phy_arb_redun_no_fault test is configured like flash_ctrl_phy_arb_redun except it does not inject arbiter faults. It demonstrates the ecc faults are not related to arbiter faults.

Addresses #22443


vogelpi (approved):
This looks good to me. Question: Do we also need to make a change in the testplan? That's where I'd usually look for test descriptions.

PR #22556, opened 2024-04-15
[pentest] Return device ID to host

nasahlpa:
This commit extends the init function of the SCA and FI tests to read out the DEVICE_ID register in the life cycle controller. This can be used by the host to categorize test results.

### DIFF
--- a/sw/device/tests/penetrationtests/firmware/BUILD
+++ b/sw/device/tests/penetrationtests/firmware/BUILD
@@ -177,13 +177,16 @@ cc_library(
     deps = [
         "//sw/device/lib/base:csr",
         "//sw/device/lib/dif:alert_handler",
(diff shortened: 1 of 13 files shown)


m-temp (approved):
LGTM

pamaury (approved):
The code looks ok to me, that's a lot of duplication though, I wonder if this could be part of some common routine?

nasahlpa:
Thanks for your review @m-temp and @pamaury!

Regarding redundancy, for each class of test (AES SCA, KMAC SCA, Ibex FI, ...) the corresponding "Init" sub-command needs to send back the device ID to the host. I've now moved the sending of the data into the `sca_read_device_id()` function - this should reduce the redundancy a bit.

PR #22562, opened 2024-04-15
[pentest] Add SRAM Write Static test

nasahlpa:
This commit adds the ibex.fi.char.sram_write_static FI test to the codebase. The aim of this test is to fault one of the 64 consecutive SW instructions used to write a static value into SRAM. The host code is located in lowRISC/ot-sca#359

### DIFF
--- a/sw/device/tests/penetrationtests/firmware/ibex_fi.c
+++ b/sw/device/tests/penetrationtests/firmware/ibex_fi.c
@@ -102,7 +102,7 @@ static dif_flash_ctrl_device_info_t flash_info;
 // we can do the write/read test without the risk of clobbering data
 // used by the program.
 OT_SECTION(".data")
-static volatile uint32_t sram_main_buffer[32];
+static volatile uint32_t sram_main_buffer[256];
 static volatile uint32_t sram_main_buffer_large[4000];
 
 status_t handle_ibex_fi_address_translation(ujson_t *uj) {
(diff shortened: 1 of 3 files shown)


sw/device/tests/penetrationtests/firmware/ibex_fi.c
```diff
@@ -586,6 +586,183 @@ status_t handle_ibex_fi_char_sram_read(ujson_t *uj) {
...
+  uintptr_t sram_main_buffer_addr = (uintptr_t)&sram_main_buffer;
+  mmio_region_t sram_region_main_addr =
+      mmio_region_from_addr(sram_main_buffer_addr);
+
+  // Initialize SRAM region with inverse ref_values to avoid that data from a
+  // previous run are still in memory.
+  for (int i = 0; i < 64; i++) {
+    mmio_region_write32(sram_region_main_addr, i * (ptrdiff_t)sizeof(uint32_t),
+                        ~ref_values[0]);
+  }
```
timothytrippel, line 603: Is there a problem with:
```suggestion
  // Initialize SRAM region with inverse ref_values to avoid that data from a
  // previous run are still in memory.
  for (size_t i = 0; i < 64; i++) {
    sram_main_buffer[i] = ~ref_values[0];
  }
```
?

sw/device/tests/penetrationtests/firmware/ibex_fi.c
```diff
@@ -586,6 +586,183 @@ status_t handle_ibex_fi_char_sram_read(ujson_t *uj) {
...
+
+  // FI code target.
+  // Unrolled to create 64 consecutive SW instructions as a FI target.
+  sca_set_trigger_high();
+  mmio_region_write32(sram_region_main_addr, 0 * (ptrdiff_t)sizeof(uint32_t),
+                      ref_values[0]);
```
timothytrippel, line 609: similarly, can this just be: `sram_main_buffer[0] = ref_values[0];` ?

sw/device/tests/penetrationtests/firmware/ibex_fi.h
```diff
@@ -169,6 +169,25 @@ status_t handle_ibex_fi_char_sram_static(ujson_t *uj);
  */
 status_t handle_ibex_fi_char_sram_read(ujson_t *uj);
 
+/**
+ * ibex.fi.char.sram_write_static command handler.
+ *
+ * This FI penetration tests executes the following instructions:
+ * - Set the trigger.
+ * - Add 10 NOPs to delay the trigger
+ * - Write 32 static values into SRAM.
```
timothytrippel, line 178: are there 10 NOPs I am missing? and aren't you writing 64 values?

sw/device/tests/penetrationtests/firmware/ibex_fi.c
nasahlpa: Yes, you are right, this works too and is much simpler. I've copied this from an SRAM test to make sure that the buffer is in SRAM.

sw/device/tests/penetrationtests/firmware/ibex_fi.c
nasahlpa: Fixed, thanks.

sw/device/tests/penetrationtests/firmware/ibex_fi.h
nasahlpa: Ah thanks, yes. I've now added the nops and updated the description.

sw/device/tests/penetrationtests/firmware/ibex_fi.c
```diff
@@ -586,6 +586,111 @@ status_t handle_ibex_fi_char_sram_read(ujson_t *uj) {
...
 }
 
+status_t handle_ibex_fi_char_sram_write_static(ujson_t *uj) {
+  // Clear registered alerts in alert handler.
+  sca_registered_alerts_t reg_alerts = sca_get_triggered_alerts();
+
+  // Initialize SRAM region with inverse ref_values to avoid that data from a
+  // previous run is still in memory.
+  for (size_t i = 0; i < 64; i++) {
+    sram_main_buffer[i] = ~ref_values[0];
```
timothytrippel, line 596: Alternatively we could also use `memset` here I think? even simpler. Only works if the buffer is initialized all to the same value.

timothytrippel (approved):

PR #22568, opened 2024-04-15
[alert_handler,rtl] Correct width of prim_count instance

rswarbrick:
This width wasn't quite right for `u_prim_count_esc_cnt` in the design (and Jasper correctly points out that we can't get to the maximal count value).

We pick a target to ping by incrementing that target's index every time we went around the FSM that drove the pings. Of course, there are only finitely many targets (`N_ESC_SEV` of them) and we wrap at `N_ESC_SEV-1`.

With the current code (before or after this commit), this is done by driving an `esc_cnt_clr` signal. Before we switched to use `prim_count` (in commit b5eb150), the wrapping was implemented directly in this file.

The switch to `prim_count` was reasonable, but instantiated the counter with a width of `PING_CNT_DW`, which is 16, instead of the number of bits needed to represent `N_ESC_SEV-1 = 3`, which is just 2.

### DIFF
--- a/hw/ip_templates/alert_handler/rtl/alert_handler_ping_timer.sv
+++ b/hw/ip_templates/alert_handler/rtl/alert_handler_ping_timer.sv
@@ -58,6 +58,9 @@ module alert_handler_ping_timer import alert_pkg::*; #(
   localparam int unsigned ReseedLfsrExtraBits = 3;
   localparam int unsigned ReseedLfsrWidth = PING_CNT_DW + ReseedLfsrExtraBits;
 
+  // The number of bits needed for an index into the esc senders
+  localparam int unsigned EscSenderIdxWidth = $clog2(N_ESC_SEV);
+
   // A few smoke checks for the DV mask:
   // 1) make sure the value is a right-aligned mask.
   //    this can be done by checking that mask+1 is a power of 2.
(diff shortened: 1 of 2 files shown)


vogelpi (approved):
Thanks for fixing this @rswarbrick . This looks good to me!

hw/ip_templates/alert_handler/rtl/alert_handler_ping_timer.sv
```diff
@@ -265,7 +268,7 @@ module alert_handler_ping_timer import alert_pkg::*; #(
 
   // generate ping enable vector
   assign alert_ping_req_o = NAlerts'(alert_ping_en) << id_to_ping_q;
-  assign esc_ping_req_o   = N_ESC_SEV'(esc_ping_en) << esc_cnt;
+  assign esc_ping_req_o   = EscSenderIdxWidth'(esc_ping_en) << esc_cnt;
```
a-will, line 271: @qmn discovered this one.

Hm.. pretty sure this particular line change wasn't right. `esc_ping_req_o` is of size `N_ESC_SEV`. Here we are saved by context-determined expressions widening the result compared to self-determined expressions, though (a dubious feature of SystemVerilog :sweat_smile: )

rswarbrick: Er. Looks like someone else is more efficient! Never mind :-D

PR #22573, opened 2024-04-15
[certs] implement TPM cert template generation

vbendeb:


pamaury:
Hi Vadim, this is a good start but I wonder if the subject alt name handling could be made more general, in a similar way to `subject`? Concretely, this would add a field to the main certificate:
```rust
pub subject_alt_name: HashMap<AttributeType, Value<String>>,
```
and the hjson would look like:
```hjson
certificate: {
   ...
   subject: {
            country: "USA",
            state: "CA",
            organization: "Google",
            organizational_unit: "Engineering",
            common_name: "ChromeOS OT TPM EK",
        },
   subject_alt_name: {
        tpm_manufacturer: {var: "tpm_manuf" },
        tpm_model: {var: "tpm_model" },
        tpm_version: {var: "tpm_version" },
   }
}
```
For that you need to add support for the TPM OIDs in [`AttributeType`](https://github.com/lowRISC/opentitan/blob/master/sw/host/ot_certs/src/asn1/x509.rs#L36). When producing the extension, you can probably reuse the [builder code for pushing names](https://github.com/lowRISC/opentitan/blob/master/sw/host/ot_certs/src/asn1/x509.rs#L169). It also has the advantage that when parsing a certificate, you can parse any subject alt name, not just TPM ones.

timothytrippel:
> Hi Vadim, this is a good start but I wonder if the subject alt name handling could be made more general, in a similar way to `subject`? Concretely, this would add a field to the main certificate:
> 
> ```rust
> pub subject_alt_name: HashMap<AttributeType, Value<String>>,
> ```

And by `HashMap`, I think we mean `IndexMap`? (for build reproducibility)

sw/device/silicon_creator/lib/cert/tpm_ek.hjson
```diff
@@ -0,0 +1,99 @@
...
+        },
+        authority_key_identifier: { var: "auth_key_key_id" },
+        subject_key_identifier: { var: "creator_pub_key_id" },
+        basic_constraints: { ca: false },
+        subject_alt_name: {
+            vendor: { var: "sal_manufacturer" },
+            model: { var: "sal_model" },
+            version: { var: "sal_version" },
+        },
+        extensions: [],
```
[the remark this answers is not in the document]
timothytrippel: @pamaury : just a thought for better clarity / consistency with the [X.509 spec](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2), perhaps in a follow up PR we should name the `extensions` field here `private_extensions` since the spec states "_The X.509 v3 certificate format also allows communities to define private extensions to carry information unique to those communities._"  Thoughts?

sw/device/silicon_creator/lib/dice.c
```diff
@@ -95,40 +98,15 @@ static void curr_pubkey_le_to_be_convert(attestation_public_key_t *pubkey) {
...
+    const sc_keymgr_diversification_t *diversifier, attestation_key_seed_t seed,
+    sc_keymgr_state_t desired_keymgr_state, hmac_digest_t *pubkey_id,
+    attestation_public_key_t *pubkey) {
+  HARDENED_RETURN_IF_ERROR(sc_keymgr_state_check(desired_keymgr_state));
 
   // Generate / sideload key material into OTBN, and generate the ECC keypair.
-  HARDENED_RETURN_IF_ERROR(otbn_boot_attestation_keygen(
-      otbn_ecc_keygen_seed, keymgr_diversifier, pubkey));
+  HARDENED_RETURN_IF_ERROR(
+      otbn_boot_attestation_keygen(seed, *diversifier, pubkey));
```
timothytrippel, line 109: We will need to refactor the `otbn_boot_attestation_keygen` function to:
1. Allow selecting which side of the key ladder we are generating a key from. For the TPM key case we want to generate from the `sealing` side, but as is currently written, the `otbn_boot_attestation_keygen` function calls `sc_keymgr_generate_attestation_key_otbn` [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/otbn_boot_services.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=132), which under the hood, selects the `attestation` side of the key ladder [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/drivers/keymgr.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=216).
2. rename this function from `otbn_boot_attestation_keygen` to `otbn_boot_ecc_p256_keygen`, or something like that

I can turn this into a separate issue, and we can submit a follow up PR to address.

sw/device/silicon_creator/lib/attestation.h
```diff
@@ -75,6 +75,18 @@ typedef enum {
    * The CDI_1 attestation key seed.
    */
   kCdi1AttestationKeySeed = 2,
+  /**
+   * The TPM EK attestation key seed.
+   */
+  kTpmEkAttestationKeySeed = 3,
```
timothytrippel, line 81: We will need to update the provisioning firmware to provision these seeds into a flash info page during manufacturing. We can put them on the same page as the attestation key seeds if we would like. The current provisioning firmware only provisions the attestation key seeds [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/lib/personalize.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=373). These seeds are loaded from flash info into OTBN DMEM during key generation by `otbn_boot_attestation_keygen`  [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/otbn_boot_services.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=141).

I can turn this into a separate issue, and we can address in a follow up PR.

sw/device/silicon_creator/lib/dice.c
```diff
@@ -141,6 +119,62 @@ rom_error_t dice_attestation_keygen(dice_key_t desired_key,
...
+    otbn_ecc_keygen_seed = kCdi1AttestationKeySeed;
+    break;
+  default:
+    return kErrorDiceInvalidArgument;
+  };
+  return common_attestation_keygen(keymgr_diversifier, otbn_ecc_keygen_seed,
+                                   desired_state, pubkey_id, pubkey);
+}
+
+rom_error_t tpm_attestation_keygen(tpm_key_t desired_key,
```
timothytrippel, line 152: any reason you split this into a separate function instead of just adding additional cases to the original `dice_attestation_keygen` function and just renaming the `dice_attestation_keygen` to `dice_ecc_keygen` or something like that?

sw/device/silicon_creator/lib/dice.h
```diff
@@ -107,4 +116,23 @@ rom_error_t dice_cdi_1_cert_build(manuf_certgen_inputs_t *inputs,
                                   attestation_public_key_t *cdi_1_pubkey,
                                   uint8_t *cert, size_t *cert_size);
 
-#endif  // OPENTITAN_SW_DEVICE_SILICON_CREATOR_LIB_DICE_H_
+OT_WARN_UNUSED_RESULT
+rom_error_t tpm_attestation_keygen(tpm_key_t desired_key,
+                                   hmac_digest_t *pubkey_id,
+                                   attestation_public_key_t *pubkey);
+
```
timothytrippel, line 123: These are all missing docstrings.

sw/device/silicon_creator/lib/dice.c
```diff
@@ -319,3 +353,77 @@ rom_error_t dice_cdi_1_cert_build(manuf_certgen_inputs_t *inputs,
 
   return kErrorOk;
 }
+
+rom_error_t dice_tpm_ek_cert_build(manuf_certgen_inputs_t *inputs,
+                                   attestation_public_key_t *tpm_ek_pubkey,
+                                   uint8_t *tpm_ek_tbs, size_t *tpm_ek_tbs_size) {
+  hmac_digest_t digest;
```
timothytrippel, line 360: What is this used for? if you just need an array of 0 bytes maybe create one instead?

sw/device/silicon_creator/lib/dice.c
```diff
@@ -319,3 +353,77 @@ rom_error_t dice_cdi_1_cert_build(manuf_certgen_inputs_t *inputs,
...
+
+  tpm_ek_tbs_values_t tpm_ek_tbs_params = {
+      .auth_key_key_id = inputs->auth_key_key_id,
+      .auth_key_key_id_size = kDiceCertKeyIdSizeInBytes,
+      .creator_pub_key_ec_x = (uint8_t *)tpm_ek_pubkey->x,
```
timothytrippel, line 365: We cast with `unsigned char *` so as not to violate strict aliasing rules (https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf; Section 6.5, Item 7) and for consistency with the above code.

Applies below too.

sw/device/silicon_creator/lib/dice.c
```diff
@@ -319,3 +353,77 @@ rom_error_t dice_cdi_1_cert_build(manuf_certgen_inputs_t *inputs,
...
+      .sal_version_len = 5,
+      .sal_manufacturer = "Nuvoton",
+      .sal_manufacturer_len = 7,
+      .sal_model = "Ti50",
+      .sal_model_len = 4,
+  };
+
+  HARDENED_RETURN_IF_ERROR(
+      tpm_ek_build_tbs(&tpm_ek_tbs_params, tpm_ek_tbs, tpm_ek_tbs_size));
+
```
timothytrippel, line 381: What do we need to do with the private key? should these be stored somewhere so they are accessible by owner firmware later? Should we store them in OTBN DMEM along side the DICE CDI_1 private key?

timothytrippel:
> Hi Vadim, this is a good start but I wonder if the subject alt name handling could be made more general, in a similar way to `subject`? Concretely, this would add a field to the main certificate:
> 
> ```rust
> pub subject_alt_name: HashMap<AttributeType, Value<String>>,
> ```
> 
> and the hjson would look like:
> 
> ```
> certificate: {
>    ...
>    subject: {
>             country: "USA",
>             state: "CA",
>             organization: "Google",
>             organizational_unit: "Engineering",
>             common_name: "ChromeOS OT TPM EK",
>         },
>    subject_alt_name: {
>         tpm_manufacturer: {var: "tpm_manuf" },
>         tpm_model: {var: "tpm_model" },
>         tpm_version: {var: "tpm_version" },
>    }
> }
> ```
> 
> For that you need to add support for the TPM OIDs in [`AttributeType`](https://github.com/lowRISC/opentitan/blob/master/sw/host/ot_certs/src/asn1/x509.rs#L36). When producing the extension, you can probably reuse the [builder code for pushing names](https://github.com/lowRISC/opentitan/blob/master/sw/host/ot_certs/src/asn1/x509.rs#L169). It also has the advantage that when parsing a certificate, you can parse any subject alt name, not just TPM ones.

Also I think that if the HJSON description of the cert contains the Subject Alt Name extension, there are a few more constraints to enforce as described here: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6

sw/device/silicon_creator/lib/dice.c
timothytrippel: Added an Issue to track this: https://github.com/lowRISC/opentitan/issues/22622

sw/device/silicon_creator/lib/cert/tpm_cek.hjson
```diff
@@ -0,0 +1,93 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+{
+    name: "tpm_cek",
+
+    variables: {
+        // Creator public key: this is an EC point on
+        // the prime256v1 curve so it has two components
+        // (x and y) which are 32-bytes integers.
+        creator_pub_key_ec_x: {
```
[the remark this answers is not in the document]
vbendeb: yes, damn copy and paste, fixed.

sw/device/silicon_creator/lib/cert/tpm_cek.hjson
```diff
@@ -0,0 +1,93 @@
...
+            type: "integer",
+            size: 32,
+        },
+        creator_pub_key_ec_y: {
```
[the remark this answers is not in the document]
vbendeb: fixed.

sw/device/silicon_creator/lib/cert/tpm_cek.hjson
```diff
@@ -0,0 +1,93 @@
...
+            type: "integer",
+            size: 32,
+        },
+        // Creator public key ID: this is a 20-byte hash
+        // derived from the creator public key.
+        creator_pub_key_id: {
```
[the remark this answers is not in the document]
vbendeb: fixed

sw/device/silicon_creator/lib/cert/tpm_cek.hjson
```diff
@@ -0,0 +1,93 @@
...
+            size: 100,
+        },
+        sal_model: {
+            type: "string",
+            size: 100,
+        },
+        sal_version: {
+            type: "string",
+            size: 100,
+        },
```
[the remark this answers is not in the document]
vbendeb: another damn copy and paste, fixed.

sw/device/silicon_creator/lib/cert/tpm_cik.hjson
```diff
@@ -0,0 +1,93 @@
...
+            size: 100,
+        },
+        sal_model: {
+            type: "string",
+            size: 100,
+        },
+        sal_version: {
+            type: "string",
+            size: 100,
+        },
```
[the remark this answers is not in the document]
vbendeb: fixed.

sw/device/silicon_creator/lib/dice.h
vbendeb: added

sw/device/silicon_creator/lib/dice.c
vbendeb: `tpm_ek_build_tbs()` returns a digest of the pubkey of the certificate, it is not used in case of TPM certs, as there is no chain, but the function expects the digest space to fill, so I use the `hmac_digest_t` to make sure that the space is there. I could define it as `unsigned char buf[sizeiof(hmac_diget_t)];` instead, should I?

sw/device/silicon_creator/lib/dice.c
vbendeb: fixed

sw/device/silicon_creator/lib/dice.c
vbendeb: I was under impression that the private key is not stored anywhere and is generated at run time when necessary?

But the whole issue of storing the generated certificates is open, let's discuss it once we get the generation going.

sw/device/silicon_creator/lib/cert/tpm_cek.hjson
```diff
@@ -0,0 +1,93 @@
+// Copyright lowRISC contributors.
```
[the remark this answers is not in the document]
vbendeb: fixed

vbendeb:
Sorry guys, I am not entirely sure how I update the pull request. 

I fixed most of the comments and updated my own branch: https://github.com/vbendeb/opentitan/tree/tpm_certs

sw/device/silicon_creator/lib/dice.c
vbendeb: I had to split it because the key type enums are different for DICE and TPM, as discussed, let's consolidate them in the next round.

sw/host/ot_certs/src/template/mod.rs
```diff
@@ -75,7 +75,9 @@ pub struct Certificate {
     pub subject_key_identifier: Value<Vec<u8>>,
     // X509 basic constraints extension, optional.
     pub basic_constraints: Option<BasicConstraints>,
-    /// X509 certificate extensions.
+    /// X509 Subject Alternative Name extension, optional.
+    pub subject_alt_name: Option<IndexMap<AttributeType, Value<String>>>,
```
vbendeb, line 79: As discussed, I started with declaring this as `IndexMap<>`, but the hjson parser started failing on all templates where this field was not persent.

pamaury: You can remove the `Option` by diing:
```rust
#[serde(default)]
    subject_alt_name: IndexMap<...>,
```

pamaury:
> Also I think that if the HJSON description of the cert contains the Subject Alt Name extension, there are a few more constraints to enforce as described here: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6

That's a good point, although I am not entirely sure if ot_certs is the right place to enforce that. I see ot_certs as tool to produce certificates, but checking that they are valid/satisfy the spec is a nontrivial task. We could potentially add a test somewhere that tries to load, say, a generated TPM cert and ask OpenSSL to validate it as much as possible? (assuming OpenSSL enforces that)

pamaury:
> Sorry guys, I am not entirely sure how I update the pull request.
> 
> I fixed most of the comments and updated my own branch: https://github.com/vbendeb/opentitan/tree/tpm_certs

Once you have created a PR, any push to your branch will update the PR :)

sw/device/silicon_creator/lib/dice.c
timothytrippel: Hmm, `tpm_ek_build_tbs()` does not return a digest of the pubkey, it needs the digest of the pubkey as an **_input_**, as the digest of the pubkey is what we use for the serial number for the DICE certs. If you are using the digest of the pubkey as the serial number for the TPM certs, then you need to pass in the digest of the pubkey to `dice_tpm_ek_cert_build()` since it is generated via the `tpm_attestation_keygen()` keygen function. Does that make sense?

sw/device/silicon_creator/lib/dice.h
timothytrippel: Hmm I don't see the docstrings, did you forget to upload them in your latest push?

timothytrippel (commented):
I think this is getting close, just a couple more tweaks and we should be able to get this in.

sw/device/silicon_creator/lib/dice.h
vbendeb: yes, have not saved the latest edit, fixed.

sw/device/silicon_creator/lib/cert/tpm_cek.hjson
```diff
@@ -0,0 +1,81 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+{
+    name: "tpm_cek",
+
+    variables: {
+        // Creator public key: this is an EC point on
```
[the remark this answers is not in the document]
vbendeb: done

vbendeb:
I *think* I addressed all outstanding comments, here one can see the difference between the previous version of the patch and the latest one.

https://github.com/vbendeb/opentitan/compare/64d18d1f84fc2a2f8c3ccb856f19566dba5859f4..14ad6b7a7b633b69bd52450d47a4653061921725

Unfortunately due to constant rebases there are many unrelated differences, but you can search for file names (dice.h, dice.c, etc).

This lack of clear differences between versions of the same PR is a pretty significant github shortcoming compared to gerrit.

sw/host/ot_certs/src/asn1/x509.rs
```diff
@@ -230,6 +237,48 @@ impl X509 {
...
 
+    pub fn push_subject_alt_name_ext<B: Builder>(
+        builder: &mut B,
+        subject_alt_name: &IndexMap<AttributeType, Value<String>>,
+    ) -> Result<()> {
+        if subject_alt_name.is_empty() {
+            // Subject ALternative Name is an optional extension, it's ok not to be present.
+            return Ok(());
+        }
+        let required_attributes = [
```
pamaury, line 248: I would not add a list of required attributes, subject alt names can be used for everything and not just TPM. I think it's the responsability of the tpm template to set it up correct.

sw/host/ot_certs/src/template/mod.rs
```diff
@@ -86,6 +89,13 @@ pub struct BasicConstraints {
     pub ca: Value<bool>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct SubjectAltName {
```
pamaury, line 93: Is this structure used?

sw/host/tests/chip/rv_dm/src/control_status.rs
```diff
@@ -49,7 +49,10 @@ fn test_control_status(opts: &Opts, transport: &TransportWrapper) -> Result<()>
     dmi.dmi_write(consts::DMCONTROL, dmcontrol)?;
     // Since this target only supports 1 hart, the WARL behavior of this register is such that no
     // bits should be set at this point, except for dmactive.
-    assert_eq!(dmi.dmi_read(consts::DMCONTROL)?, consts::DMCONTROL_DMACTIVE_MASK);
+    assert_eq!(
```
pamaury, line 52: If you want to correct the formatting, maybe make that part of a different commit in the PR?

sw/host/tests/chip/sysrst_ctrl/sysrst_ctrl_ec_rst_l.rs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::{bail, Context, ensure, Result};
+use anyhow::{bail, ensure, Context, Result};
```
pamaury, line 5: If you want to correct the formatting, maybe make that part of a different commit in the PR?

pamaury:
Hi Vadim, the rust part is looking good now :) You just need to add the parsing bit so thjat ot_certs can extract the subject alt names from the certificate.

sw/host/tests/chip/sysrst_ctrl/sysrst_ctrl_ec_rst_l.rs
vbendeb: ah, a  rustfmt_fix escape, how did the code make it into the tree despite presubmit checks?

I reverted unrelated files.

sw/host/tests/chip/rv_dm/src/control_status.rs
vbendeb: reverted

sw/host/ot_certs/src/template/mod.rs
vbendeb: I guess not, a leftover from the original implementation, dropped.

sw/host/ot_certs/src/asn1/x509.rs
vbendeb: ok, after all it is easy to check the produced certificates to see that the included fields are what's expected, dropped the checks.

vbendeb:
changes from the reviewed version:

https://github.com/lowRISC/opentitan/compare/14ad6b7a7b633b69bd52450d47a4653061921725..893f20bebcbb08cd42e6c284e474097c37419622

sw/host/tests/chip/sysrst_ctrl/sysrst_ctrl_ec_rst_l.rs
pamaury: The rustfmt_check is know to be flawed because we need to manually point to targets to verify. There is a pending PR to fix it but I requires an upstream change to rules_rust that I haven't gotten around to submit.

vbendeb:
> Hi Vadim, the rust part is looking good now :) You just need to add the parsing bit so thjat ot_certs can extract the subject alt names from the certificate. It would also be nice to update `generic.hson` and `example.hjson` in the integration tests once you have implemented the parsing.

Can this be in  a separate PR?

pamaury:
> > Hi Vadim, the rust part is looking good now :) You just need to add the parsing bit so thjat ot_certs can extract the subject alt names from the certificate. It would also be nice to update `generic.hson` and `example.hjson` in the integration tests once you have implemented the parsing.
> 
> Can this be in a separate PR?

Yes I think that's fine.

timothytrippel (approved):
Thanks @vbendeb ! I think you just have to rebase this on master one more time to fix the conflicts and get it to pass CI (since there were some CI fixes that went in today). Then I can merge. Sorry for the toil on this and thanks for all the fixes and your patience!

pamaury (approved):

vbendeb:
> Thanks @vbendeb ! I think you just have to rebase this on master one more time to fix the conflicts and get it to pass CI (since there were some CI fixes that went in today). Then I can merge. Sorry for the toil on this and thanks for all the fixes and your patience!

rebased, fixed a conflict and dropped an unused import.

Thank you for your guidance and patience guys, I'm just learning the ropes, hopefully the following ones will move smoother :)

pamaury:
The CI failures are unrelated so I will merge this.

PR #22576, opened 2024-04-16
[bazel/opentitantool] Add support for ECDSA public keys and signatures

moidx:
## opentitantool changes

Updates the image commands in opentitantool to support ECDSA public keys
and signatures. The current manifest implementation uses an union to
store RSA or ECDSA public keys and signatures. In the RSA case, the
public key/signature takes up the entire union space. In the ECDSA
case the key/signature occupy the first 256bits of the data
container and the rest is padded with a known pattern.
    
This is a temporary measure until we remove support for RSA.
    
The opentitantool manifest module was updated to use public key and
signature fields in favor over the RSA specific names.

## Bazel changes

Add support for ECDSA in Bazel infrastructure under signing.bzl. At the
moment, the signing infrastructure supports RSA and ECDSA. This is so
that we can migrate to ECDSA via incremental changes.
    
ECDSA signing can be tested as follows:
    
```shell
$ bazel build //signing/examples:digests_ecdsa
```
   
The `opentitantool image manifest verify` command can be used to verify
the signature over the pre-signed artifacts.


sw/host/opentitanlib/src/crypto/ecdsa.rs
```diff
@@ -115,6 +117,15 @@ impl EcdsaRawSignature {
         Ok(sig)
     }
 
+    pub fn read_from_file(path: &Path) -> Result<EcdsaRawSignature> {
+        let err = |e: std::io::Error| Error::ReadFailed {
+            file: path.to_owned(),
+            source: anyhow!(e),
+        };
+        let mut file = File::open(path).map_err(err)?;
```
cfrantz, line 125: It would be considered idiomatic rust to stuff the closure expression into the `mapp_err` argument:
```
let mut file = File::open(path).map_err(|e| Error::ReadFailed {
    file: path.to_owned(),
    source: anyhow!(e),
})?;
```

One other note: since this isn't a trait implementation with specific error type constraints, you could also just let `?` auto-coerce the `io::Error` into `anyhow::Error`.  You could attach a context message to the error to be friendly:
```
let mut file = File::open(path).context(format!("Failed to read {path:?}"))?;
````

File: sw/host/opentitanlib/src/image/image.rs [hunk 2]
```diff
@@ -120,6 +147,41 @@ impl Image {
...
+        {
+            let rsa_key = RsaPublicKey::new(Modulus::from_le_bytes(pub_key)?)?;
+            let rsa_sig = RsaSignature::from_le_bytes(signature)?;
+            return Ok(SigverifyParams::Rsa(rsa_key, rsa_sig));
+        }
+
+        let ecdsa_pub_key = EcdsaRawPublicKey {
+            x: pub_key[0..32].try_into()?,
+            y: pub_key[32..64].try_into()?,
+        };
```
cfrantz, line 175: There is also the "reader" API of `EcdsaRaw{PublicKey,Signature}`:

```
let ecdsa_pub_key = EcdsaRawPublicKey::read(std::io::Cursor::new(pub_key))?;
let ecdsa_sig = EcdsaRawSignature::read(std::io::Cursor::new(signature))?;
```

sw/host/opentitanlib/src/crypto/ecdsa.rs
cfrantz: Or use `with_context` to defer the string formatting to only the error case:
```
use anyhow::Context;
let mut file = File::open(path).with_context(|| format!("Failed to read {path:?}"))?;
```

cfrantz (approved):

sw/host/opentitanlib/src/image/image.rs
```diff
@@ -244,26 +306,101 @@ impl Image {
...
+        let signature_bytes = signature
+            .r
+            .iter()
+            .chain(signature.s.iter())
+            .copied()
+            .collect::<Vec<u8>>();
+        let sig_padding = vec![0xa5a5a5a5u32; (384 - 64) / 4]
+            .iter()
+            .flat_map(|&x| x.to_le_bytes())
+            .collect::<Vec<u8>>();
```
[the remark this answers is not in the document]
moidx: Fixed!

sw/host/opentitanlib/src/crypto/ecdsa.rs
moidx: I went for the last option. Thanks!

Done.

sw/host/opentitanlib/src/crypto/ecdsa.rs
```diff
@@ -165,6 +176,25 @@ impl EcdsaPublicKey {
...
+
+        x.reverse();
+        y.reverse();
+
+        let key = VerifyingKey::from_encoded_point(&p256::EncodedPoint::from_affine_coordinates(
+            &GenericArray::from(<[u8; 32]>::try_from(x.as_slice()).unwrap()),
+            &GenericArray::from(<[u8; 32]>::try_from(y.as_slice()).unwrap()),
+            false,
+        ))
+        .expect("Failed to create verifying key from raw public key");
```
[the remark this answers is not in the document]
moidx: Fixed.

sw/host/opentitanlib/src/image/image.rs
moidx: Fixed.

moidx:
The errors in this PR are not relevant to this change. i2c_target_test_fpga_cw310_sival test harness CRC error.

PR #22577, opened 2024-04-16
[i2c] Reduce counter sizes to save area

a-will:
Reduce counter sizes to only the widths needed to support Standard mode at a 1 GHz core clock.

In the FPGA implementation, this saves 44 flops (~4%) and over 100 LUTs (~6.6%). The granularity of the FPGA area is large, so the size of the savings can be misleading, but compared to what we had prior to the FSM split, this brings us to +29 flops, -59 LUTs for the whole chain of commits.

This builds on #22551. Only the last commit is new.

### DIFF
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -547,13 +547,19 @@
       swaccess: "rw"
       hwaccess: "hro"
       fields: [
-        { bits: "15:0"
+        { bits: "12:0"
           name: "THIGH"
-          desc: "The actual time to hold SCL high in a given pulse: in host mode, when there is no stretching this value is 3 cycles longer as tracked in issue #18962"
+          desc: '''
+                The actual time to hold SCL high in a given pulse.
+                This field is sized to have a range of at least Standard Mode's 4.0 us max with a core clock at 1 GHz.
+                '''
         }
-        { bits: "31:16"
+        { bits: "28:16"
           name: "TLOW"
-          desc: "The actual time to hold SCL low between any two SCL pulses"
+          desc: '''
+                The actual time to hold SCL low between any two SCL pulses.
+                This field is sized to have a range of at least Standard Mode's 4.7 us max with a core clock at 1 GHz.
+                '''
         }
       ]
     }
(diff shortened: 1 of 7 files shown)


hw/ip/i2c/data/i2c.hjson
```diff
@@ -506,11 +547,11 @@
       swaccess: "rw"
       hwaccess: "hro"
       fields: [
-        { bits: "15:0"
+        { bits: "12:0"
```
a-will, line 550: With a maximum core clock of 1 GHz, all the 13-bit counters are sized to handle at least 5 us.

hw/ip/i2c/data/i2c.hjson
```diff
@@ -524,11 +565,11 @@
       swaccess: "rw"
       hwaccess: "hro"
       fields: [
-        { bits: "15:0"
+        { bits: "9:0"
```
a-will, line 568: With a maximum core clock of 1 GHz, the rise and fall times are only 10-bit counters to accommodate a maximum of 1000 ns (and really, the fall time could be reduced further, since the max allowed by spec is only 300 ns).

andreaskurth (approved):
Thanks for these optimizations, @a-will. Area matters these days :-)

LGTM; I just have suggestions about capturing this in the documentation. (And obviously please only rebase & merge this after the PR that this depends on got merged.)

hw/ip/i2c/data/i2c.hjson
a-will: > So we can reduce the width of `T_F` by another bit without loss, no?

We can. I simply got lazy and matched rise and fall widths.

a-will: Done. I also removed some bits from TSU_DAT, which had 4 more bits than necessary. THD_DAT likely also has more bits than necessary, but there's a line about CBUS-compatible controllers needing 5 us, so I left it alone for now.

hw/ip/i2c/data/i2c.hjson
a-will: Done.

a-will:
Ah, kokoro provided results. I'll fix the lint errors from the last PR(s).

a-will:
Given the content of this PR, it is very highly unlikely to be the cause, though.

jwnrt:
I think you're right. Maybe it's just due to the bitstreams available in the cache before and after this PR merged. Apologies for the noise.

PR #22579, opened 2024-04-16
Fix entropy src edn reqs

pamaury:
The first commit enables `opentitan_test` to take per-exec-env defines.
The second commit fixes the test, the test has two problems in SiVal:
- the SiVal ROM_EXT blocks direct access to the OTP registers so the
  OTP test triggers an exception. Disable this test in SiVal as it
  simply cannot be done.
- the SiVal ROM_EXT cranks the key manage to the owner state whereas
  the ROM does not touch the key manager. Change the test to expect
  a different state in SiVal.

### DIFF
--- a/rules/opentitan/defs.bzl
+++ b/rules/opentitan/defs.bzl
@@ -206,7 +206,7 @@ def opentitan_test(
             kind = kind,
             deps = deps,
             copts = copts,
-            defines = defines,
+            defines = defines + getattr(tparam, "defines", []),
             local_defines = local_defines,
             includes = includes,
             linker_script = linker_script,
--- a/rules/opentitan/fpga_cw310.bzl
+++ b/rules/opentitan/fpga_cw310.bzl
@@ -228,6 +228,7 @@ def cw310_params(
         needs_jtag = False,
         test_cmd = "",
         data = [],
+        defines = [],
         **kwargs):
     """A macro to create CW310 parameters for OpenTitan tests.
 
@@ -277,4 +278,5 @@ def cw310_params(
         param = kwargs,
         post_test_cmd = post_test_cmd,
         post_test_harness = post_test_harness,
+        defines = defines,
     )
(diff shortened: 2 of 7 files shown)


pamaury:
@cfrantz can you check the bazel part?

engdoreis:
I think that this test may never be able to run with ROM_EXT, therefore we should remove the execution environments that use the ROM_EXT from this test which would automatically remove it from the nightly. 
Keeping the test with ROM and TEST_ROM would allow it to run on the FPGA and NON-PROD silicon.

pamaury:
> I think that this test may never be able to run with ROM_EXT, therefore we should remove the execution environments that use the ROM_EXT from this test which would automatically remove it from the nightly. Keeping the test with ROM and TEST_ROM would allow it to run on the FPGA and NON-PROD silicon.

This test can run fine in ROM_EXT, the OTP is really just a tiny part of the test to exercise the EDN; the only issue is that the code was making a few assumptions of the key manage state and that's easily fixed. I would suggest not to disable it with ROM_EXT.

jwnrt (approved):
This LGTM, thanks @pamaury 

Do you think we should automatically set a define whose value is the exec env for each exec env?

pamaury:
> This LGTM, thanks @pamaury
> 
> Do you think we should automatically set a define whose value is the exec env for each exec env?

I would say no because if you do that then two targets could not compile the same code just because of that define.

jwnrt:
Oh, the binary still changes even if you don’t use the define? Okay, makes sense not to then.

pamaury:
> Oh, the binary still changes even if you don’t use the define? Okay, makes sense not to then.

Well for one you cannot be sure how an extra define will influence the compiler but more importantly, adding an extra define to a bazel rule will make bazel think it's a different action so it won't cache.

engdoreis (approved):
LGTM.

PR #22584, opened 2024-04-16
[rom_ctrl,lc_ctrl,dv] Specialise scratch_path/rel_path for variants

### LINKED ISSUE #22122: [lc_ctrl,dv] Regression results lost for one config of lc_ctrl tests
https://reports.opentitan.org/hw/top_earlgrey/dv/summary/latest/report.html ended up running tests for both configs for lc_ctrl, but one of them seems to have no results for the majority of tests:

https://reports.opentitan.org/hw/ip/lc_ctrl/dv/2024.03.18_04.36.41/report.html

_Originally posted by @a-will in https://github.com/lowRISC/opentitan/issues/22104#issuecomment-2007877974_

rswarbrick:
Without this change, there is a problem where we have two different variants of each block that are both getting built for nightly tests and both variants write results to the same place. This fix nicks the equivalent code from the aes block (which has to solve the same problem).

Fixes #22122.

### DIFF
--- a/hw/ip/lc_ctrl/dv/lc_ctrl_base_sim_cfg.hjson
+++ b/hw/ip/lc_ctrl/dv/lc_ctrl_base_sim_cfg.hjson
@@ -57,6 +57,15 @@
       name: tl_dbw
       value: 8
     }
+    // Override the default scratch directory and rel_path to take variant into account
+    {
+      name: scratch_path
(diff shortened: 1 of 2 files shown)


vogelpi (approved):
LGTM, thanks @rswarbrick !

Issue #22385, opened 2024-04-02
[fpv] Inspect formal FPV results and get them back up to 100% pass rate


vogelpi:
### Description

For a couple of weeks during the M2 push, the result publishing for the FPV regressions was broken (see #21694) . The tooling has now been fixed and it seems that during M2 some failures made their way into the system (previously, everything was green and at 100%). Some failures are systematic and prevent FPV from running at all, see e.g. the edn_sec_cm results [here](https://reports.opentitan.org/hw/top_earlgrey/formal/sec_cm/summary/latest/report.html).

For M3, we need to inspect the failures and get things back up to 100% pass rate.

The following regressions should be inspected:
- https://reports.opentitan.org/hw/top_earlgrey/formal/sec_cm/summary/latest/report.html
- https://reports.opentitan.org/hw/top_earlgrey/formal/ip/summary/latest/report.html
- https://reports.opentitan.org/hw/top_earlgrey/formal/prim/summary/latest/report.html

vogelpi:
I recommend starting with the sec_cm regression: https://reports.opentitan.org/hw/top_earlgrey/formal/sec_cm/summary/latest/report.html

vogelpi:
@rswarbrick stuff seems to have been broken for longer than we thought. One issue is the black boxing of prim_count in various places.
@moidx could ask @cindychip on this.
@rswarbrick this would be helpful.

rswarbrick:
On the sec_cm side:

- The failures in `entropy_src` should hopefully be fixed by PR #22506.
- The failures in `lc_ctrl` are more mystifying! I've opened issue #22509 with some more details about the test that is actually failing. Unfortunately, I don't yet understand what's going on with this.

rswarbrick:
A bit of extra background: I don't believe that we've ever had a 100% pass rate for these tests (so I quibble with the "back to" in the title :-D). It's probably still worth putting some serious effort into tidying things up though.

vogelpi:
Thanks for your work on this @rswarbrick . I have three questions:
1. The latest sec_cm regression results for EDN show all N/A. But I am pretty sure we had tests running for this in the past. Would you mind investigating this please?
[...]

PR #22587, opened 2024-04-16
[edn,rtl] Fix indices in assertion

rswarbrick:
The existing code wasn't quite right and actually causes a compilation error when you try to push it through Jasper. For a bit more detail, edn_enable_fo is the "edn enable fanout". This is indexed by entries from FatalErr(=1) to LastEdnEntry-1, and doesn't have anything at index zero. (This is what causes the compilation error)

The bug was introduced by a974108, which removed an "enable_q" signal from edn.sv. Looking at the assertions, I think the property that we're talking about is "some of EDN is enabled". To get what I hope is the intended behaviour, I've used a unary OR to combine together all of the enable signals.

The bug is causing N/A entries in the "nightly" FPV tests, discussed here: https://github.com/lowRISC/opentitan/issues/22385#issuecomment-2058225120.

### DIFF
--- a/hw/ip/edn/rtl/edn.sv
+++ b/hw/ip/edn/rtl/edn.sv
@@ -117,15 +117,16 @@ module edn
   for (genvar i = 0; i < NumEndPoints; i = i+1) begin : gen_edn_if_asserts
     `ASSERT_KNOWN(EdnEndPointOut_A, edn_o[i])
 
-    // These assertions check that EDN data will be stable from edn_ack until the next EDN request
-    // or until next EDN enablement.
+    // Check that EDN data stays stable from edn_ack until the next EDN request or until EDN
+    // disablement.
     `ASSERT(EdnDataStable_A,
-            ($rose(edn_o[i].edn_ack) && $past(u_edn_core.edn_enable_fo[0])) |=>
-                $stable(edn_o[i].edn_bus) throughout edn_i[i].edn_req[->1],
-            clk_i, !rst_ni)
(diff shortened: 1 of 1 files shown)


rswarbrick:
Hmm, my (somewhat guessed) translation of the assertion that was there before is clearly not true! It's causing CI to fail.

@h-filali: Would you mind taking a look at this? (I'm hoping you'll be able to figure out what's going on rather more quickly than me!)

vogelpi:
Thanks for looking into this @rswarbrick and reporting your findings here! I should be able to take care of this (anyway need to do some work on EDN for M3).

vogelpi:
I think I've managed to solve the issue @rswarbrick . EDN intentionally clears the FIFOs upon re-enablement this assertions are observing. This behavior didn't change in the past. What the referenced commit also changed is the condition to disable the assertion. Previously, it stopped after disabling EDN. With the other commit (and your fix) the SVA would only stop after seeing a new EDN request. Obviously, a disablement/re-enablement can come before the next EDN request.

I think I've now managed to fix the issue and directly pushed into your PR. I will now approve the PR but leave the merging to you. Feel free to edit or merge tomorrow.

vogelpi (approved):
LGTM, but I've further edited the assertion (there was a second bug introduced in the same commit you referenced). I'll leave the merging up to you :-)

rswarbrick:
Thanks for this. I've looked through the change and it looks sensible to me.

PR #22588, opened 2024-04-16
SRAM timing fixes

GregAC:
With my hacky block synthesis with these fixes and the previous work (https://github.com/lowRISC/opentitan/pull/22497) I believe I have eliminated all feedthroughs from the incoming tilelink signals to the outgoing tilelink signals in sram_ctrl.

These fixes will also have a positive effect on timing elsewhere as `tlul_sram_adapter` and `tlul_err_resp` are used in various places. In turn this could give us some modest area improvements as buffers are removed and cell drive strengths can be lowered. Though note this also means increases verification risk in more places.

I'm doing a final synthesis to confirm everything now, block level regressions (on `sram_ctrl` other blocks are effect, haven't run regressions for them yet) are looking healthy. I've had issues running the sram_ctrl top-level smoke test (not related to these changes) but `chip_sw_sram_ctrl_scrambled_access` is passing successfully.

I'm going to do some block level regressions of the other blocks that use the altered primitives and check some other sram related top-level tests.


hw/ip/tlul/rtl/tlul_adapter_sram.sv
```diff
@@ -310,7 +328,7 @@ module tlul_adapter_sram
       d_data   : d_data,
       d_user   : '{default: '0, data_intg: data_intg},
       d_error  : d_valid && d_error,
-      a_ready  : (gnt_i | error_internal) & reqfifo_wready & sramreqfifo_wready
+      a_ready  : (gnt_i | missed_err_gnt_q) & reqfifo_wready & sramreqfifo_wready
```
marnovandermaas, line 331: One thing I had noticed while working on Sonata is that a missed error causes an `a_ready` to be sent but the error response never gets written to the request FIFO.

marnovandermaas (approved):
This looks reasonable to me based on just a code review. It would be good to get someone else to have a look at this as well.

hw/ip/tlul/rtl/tlul_err_resp.sv
```diff
@@ -32,24 +32,24 @@ module tlul_err_resp (
 
   always_ff @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
-      err_req_pending <= 1'b0;
+      err_rsp_pending <= 1'b0;
       err_source      <= {top_pkg::TL_AIW{1'b0}};
       err_opcode      <= Get;
       err_size        <= '0;
       err_instr_type  <= MuBi4False;
+    end else if(err_rsp_pending && tl_h_i.d_ready) begin
```
andreaskurth, line 40: Not semantically relevant, but there's a space missing here
```suggestion
    end else if (err_rsp_pending && tl_h_i.d_ready) begin
```

andreaskurth (approved):
Thx Greg, small suggestion but overall LGTM

hw/ip/tlul/rtl/tlul_err_resp.sv
GregAC: Well spotted @andreaskurth, gotta give the electrons some extra gaps to squeeze through :)

GregAC:
Kick off a kokoro rerun, assuming that's all good I will bypass CI for a new version that adds the missing space @andreaskurth highlighted given this is an important fix to get in.

GregAC:
Local block level regression runs are all looking good

hw/ip/tlul/rtl/tlul_adapter_sram.sv
```diff
@@ -300,6 +300,24 @@ module tlul_adapter_sram
...
+  // performing an SRAM request. It may be the transaction receives a ready the first cycle it is
+  // seen, but if not we force a ready the following cycle. This avoids factoring the error
+  // calculation into the outgoing ready preventing a feedthrough path from the incoming tilelink
+  // signals to the outgoing tilelink signals.
+  logic missed_err_gnt_d, missed_err_gnt_q;
+
+  // Track whether we've seen an incoming transaction with an error that didn't get a ready
+  assign missed_err_gnt_d = error_internal & tl_i_int.a_valid & ~tl_o_int.a_ready;
+
+  always @(posedge clk_i or negedge rst_ni) begin
```
vogelpi, line 313: Needs needs to be changed to always_ff. Kokoro is producing a lint error here.

vogelpi (changes requested):

hw/ip/tlul/rtl/tlul_adapter_sram.sv
vogelpi: I am quickly doing this.

vogelpi: Done

vogelpi (approved):

vogelpi:
I am merging this, the `i2c_target_test_fpga_cw310_sival` test keeps failing but it's also failing for other PRs. I am going to mark it as broken.

PR #22592, opened 2024-04-16
[tock] Add Tock UART tests.

jrvanwhy:
This test the ability of a Tock app to perform console writes and reads over the UART. It is an integration tests that tests the Console syscall driver, the UART virtualizer, and the UART peripheral driver.

### DIFF
--- a/hw/top_earlgrey/BUILD
+++ b/hw/top_earlgrey/BUILD
@@ -130,6 +130,26 @@ fpga_cw310(
 )
 
 fpga_cw310(
+    name = "fpga_hyper310_test_rom",
+    testonly = True,
+    args = [
+        "--rcfile=",
+        "--logging=info",
+        "--interface={interface}",
+    ],
+    base = ":fpga_cw310_test_rom",
+    base_bitstream = "//hw/bitstream/hyperdebug:bitstream",
+    exec_env = "fpga_hyper310_test_rom",
+    otp_mmi = "//hw/bitstream/hyperdebug:otp_mmi",
+    param = {
+        "interface": "hyper310",
+        "exit_success": DEFAULT_TEST_SUCCESS_MSG,
+        "exit_failure": DEFAULT_TEST_FAILURE_MSG,
+    },
+    rom_mmi = "//hw/bitstream/hyperdebug:rom_mmi",
+)
+
+fpga_cw310(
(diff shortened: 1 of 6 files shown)


jrvanwhy:
I think the CI error is a true positive, debugging now.

hw/top_earlgrey/BUILD
```diff
@@ -129,6 +129,26 @@ fpga_cw310(
     rsa_key = {"//sw/device/silicon_creator/rom_ext/keys/fake:rom_ext_test_private_key_0": "test_key_0"},
 )
 
+fpga_cw310(
+    name = "fpga_hyper310_test_rom",
```
a-will, line 133: Is there a compelling reason to add a test ROM environment for hyper310 and not just use `//hw/top_earlgrey:fpga_cw310_sival_rom_ext` or `//hw/top_earlgrey:fpga_hyper310_rom_ext`?

If not, it'd be nice if we didn't expand the maintenance surface, and the changes to this file and rules/opentitan/defs.bzl can be removed.

sw/device/silicon_owner/tock/tests/uart/BUILD
```diff
@@ -0,0 +1,50 @@
...
+    apps = [":tab"],
+    kernel = "//sw/device/silicon_owner/tock/kernel",
+)
+
+opentitan_test(
+    name = "uart_test",
+    exec_env = {
+        "//hw/top_earlgrey:fpga_hyper310_test_rom": None,
+    },
+    hyper310 = cw310_params(
```
a-will, line 43: Consider using one of the existing environments:

```suggestion
    exec_env = {
        "//hw/top_earlgrey:fpga_cw310_sival_rom_ext": None,
        "//hw/top_earlgrey:fpga_hyper310_rom_ext": "cw310",
    },
    cw310 = cw310_params(
```

hw/top_earlgrey/BUILD
jrvanwhy: Short answer: no.

Long answer: I was originally seeing test failures on cw310 that I believed were a result of a cw310 issue, so I moved this test over to hyper310. However, the issues eventually showed up on hyper310 as well and they were traced to a Tock driver bug. Now that the bug is fixed, we can switch back to cw310.

I don't think Tock will run with a ROM EXT yet, so those particular environments aren't an option.

PR #22593, opened 2024-04-16
[rom] Avoid touching watchdog config on wake from sleep

jesultra:
When waking from sleep, the watchdog IP will not have been reset, so it is not necessary to re-configure it.  In fact, if watchdog REGWEN has been used to disable writes to its registers, attempting to do so can generate an exception causing chip reset, which will reset not only the watchdog, but other IP blocks that the owner code wanted retained across the period of sleep.

### DIFF
--- a/sw/device/silicon_creator/rom/rom.c
+++ b/sw/device/silicon_creator/rom/rom.c
@@ -49,6 +49,7 @@
 
 #include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
 #include "otp_ctrl_regs.h"
+#include "rstmgr_regs.h"
 
 /**
  * Table of forward branch Control Flow Integrity (CFI) counters.
@@ -85,6 +86,8 @@ CFI_DEFINE_COUNTERS(rom_counters, ROM_CFI_FUNC_COUNTERS_TABLE);
 lifecycle_state_t lc_state = (lifecycle_state_t)0;
 // Boot data from flash.
 boot_data_t boot_data = {0};
+// Whether we are "simply" waking from low power mode.
+static hardened_bool_t waking_from_low_power = 0;
 
 OT_ALWAYS_INLINE
 OT_WARN_UNUSED_RESULT
@@ -131,8 +134,23 @@ OT_WARN_UNUSED_RESULT
 static rom_error_t rom_init(void) {
   CFI_FUNC_COUNTER_INCREMENT(rom_counters, kCfiRomInit, 1);
   sec_mmio_init();
-  // Initialize pinmux configuration so we can use the UART.
-  pinmux_init();
+  uint32_t reset_reasons = rstmgr_reason_get();
+  if (reset_reasons != (1U << RSTMGR_RESET_INFO_LOW_POWER_EXIT_BIT)) {
+    // The above compares all bits, rather than just the one indication "low
+    // power exit", because if there is any other reset reason, besides
+    // LOW_POWER_EXIT, it means that the chip did full reset while coming out of
+    // low power.  In that case, the state of AON IP blocks would have been
+    // reset, and the ROM should not treat this as "waking from low power".
+    waking_from_low_power = kHardenedBoolFalse;
+
+    // Initialize pinmux configuration so we can use the UART, (except if waking
+    // up from low power, as the pinmux will in such case have retained its
+    // previous configuration.)
+    pinmux_init();
+  } else {
+    waking_from_low_power = kHardenedBoolTrue;
+  }
+
   // Configure UART0 as stdout.
   uart_init(kUartNCOValue);
 
@@ -157,9 +175,14 @@ static rom_error_t rom_init(void) {
   // Update epmp config for debug rom according to lifecycle state.
   rom_epmp_config_debug_rom(lc_state);
 
-  // Re-initialize the watchdog timer.
-  watchdog_init(lc_state);
-  SEC_MMIO_WRITE_INCREMENT(kWatchdogSecMmioInit);
+  if (launder32(waking_from_low_power) != kHardenedBoolTrue) {
+    HARDENED_CHECK_EQ(waking_from_low_power, kHardenedBoolFalse);
+    // Re-initialize the watchdog timer, if the RESET was caused by anything
+    // besides waking from low power (which would have left the watchdog in its
+    // previous configuration).
+    watchdog_init(lc_state);
+    SEC_MMIO_WRITE_INCREMENT(kWatchdogSecMmioInit);
+  }
 
   // Initialize the shutdown policy.
   HARDENED_RETURN_IF_ERROR(shutdown_init(lc_state));
@@ -175,7 +198,6 @@ static rom_error_t rom_init(void) {
 
   // Initialize the retention RAM based on the reset reason and the OTP value.
   // Note: Retention RAM is always reset on PoR regardless of the OTP value.
-  uint32_t reset_reasons = rstmgr_reason_get();
   uint32_t reset_mask =
       (1 << kRstmgrReasonPowerOn) |
       otp_read32(OTP_CTRL_PARAM_CREATOR_SW_CFG_RET_RAM_RESET_MASK_OFFSET);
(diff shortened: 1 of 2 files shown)


moidx (commented):
This is a great find @jesultra!

We should also inspect rom_start.s https://github.com/lowRISC/opentitan/blob/1c75f24e9920e60db74f8c0811b89ed06422f88f/sw/device/silicon_creator/rom/rom_start.S#L182-L187

sw/device/silicon_creator/rom/rom.c
```diff
@@ -156,9 +157,14 @@ static rom_error_t rom_init(void) {
   // Update epmp config for debug rom according to lifecycle state.
   rom_epmp_config_debug_rom(lc_state);
 
-  // Re-initialize the watchdog timer.
-  watchdog_init(lc_state);
-  SEC_MMIO_WRITE_INCREMENT(kWatchdogSecMmioInit);
+  uint32_t reset_reasons = rstmgr_reason_get();
+  if (reset_reasons & ~(1U << RSTMGR_RESET_INFO_LOW_POWER_EXIT_BIT)) {
```
moidx, line 161: Should we also check the lock bit status?

a-will: That raises an interesting question at a higher layer: Should the lock bits actually survive deep sleep? Shouldn't ROM be able to configure the IPs it needs?

jesultra: I do not think that we should base the decision on whether owner code has locked the watchdog configuration registers, I think it makes sense that ROM configures watchdog (non-locked) on boot, and then refrains from touching it when coming out of deep sleep later (as the watchdog IP is not being reset, and does not need re-configuring.)

To answer Alex.  I find it reassuring that e.g. watchdog, sysrst keyboard combinations, pinmux and other things, once configured by Ti50 on first cold boot, cannot be accidentally reconfigured by a bug in Ti50 even after going in and out of deep sleep many times.  I agree that it is weird if owner code reconfigures pinmux such that e.g. serial console cannot work, and locks pinmux that way that, but I am OK if it would take an explicit reset to get out of such a situation.

jesultra:
> We should also inspect rom_start.s

That is a good point.  This means I suppose that the reading of the reset cause register needs to be re-written in assemply, and moved up before the watchdog setup in `rom_start.S`.

I am not sure how this worked for me, with the code still left in `rom_start.S` trying to re-configure the watchdog, but then again, I do not know what the difference between `abs_mmio_write32()` and `sec_mmio_write32()` is, or how they differ from the `sw` instruction used in `rom_start.S`

sw/device/silicon_creator/rom/rom.c
a-will: Personally, I lean the other way, as ROM is generally written for only a limited set of known states, constructing a narrow gate to pass through and jump to the next stage. More branching paths creates a larger surface for coverage closure and increases the risk of bugs to something that can't be updated. Meanwhile, the applications likely would have fewer risks of deep-sleep-specific init bugs if they simply had to redo init, instead of having two paths, but that wouldn't be as fast, hehe.

Ultimately, I guess either option gives us ways to shoot ourselves in the foot. They both still put it on the application developer to be compatible with whatever ROM does. Perhaps the most important thing is to precisely document that interface.

a-will:
> I am not sure how this worked for me, with the code still left in `rom_start.S` trying to re-configure the watchdog, but then again, I do not know what the difference between `abs_mmio_write32()` and `sec_mmio_write32()` is, or how they differ from the `sw` instruction used in `rom_start.S`

`sec_mmio_write32()` checks that the value became what was written with `HARDENED_CHECK_EQ()`. If the check fails, you get an illegal instruction exception.

`abs_mmio_write32()` is just the write.

rom_start.S just has raw loads and stores, like `abs_mmio_write32()`. It doesn't generally check the results.

sw/device/silicon_creator/rom/rom.c
```diff
@@ -156,9 +176,14 @@ static rom_error_t rom_init(void) {
...
 
-  // Re-initialize the watchdog timer.
-  watchdog_init(lc_state);
-  SEC_MMIO_WRITE_INCREMENT(kWatchdogSecMmioInit);
+  if (launder32(waking_from_sleep) != kHardenedBoolTrue) {
+    HARDENED_CHECK_EQ(waking_from_sleep, kHardenedBoolFalse);
+    // Re-initialize the watchdog timer, if the RESET was caused by anything
+    // besides waking from low power (which would have left the watchdog in its
+    // previous configuration).
+    watchdog_init(lc_state);
```
moidx, line 184: If we leave this as implemented here, we probably need to document what is the minimum counter configuration supported by the ROM, since we don't have any WD pet commands as of now.

sw/device/silicon_creator/rom/rom.c
```diff
@@ -130,10 +133,27 @@ OT_WARN_UNUSED_RESULT
 static rom_error_t rom_init(void) {
   CFI_FUNC_COUNTER_INCREMENT(rom_counters, kCfiRomInit, 1);
   sec_mmio_init();
-  // Initialize pinmux configuration so we can use the UART.
-  pinmux_init();
-  // Configure UART0 as stdout.
-  uart_init(kUartNCOValue);
+
+  uint32_t reset_reasons = rstmgr_reason_get();
```
moidx, line 137: Is there any meaningful difference if we do the following?

```c 
waking_from_sleep = kHardenedBoolFalse;
if (launder32(reset_reasons) & (1 << kRstmgrReasonLowPowerExit)) {
  waking_from_sleep = kHardenedBoolTrue;
}
```

I can't think of a case where the attacker could benefit from skipping the branch, but if you can think of one, then it may be worth it to add a redundant check later on.

```c
if (launder32(rstmgr_reason_get()) & (1 << kRstmgrReasonLowPowerExit)) {
  HARDENED_CHECK_EQ(waking_from_sleep, kHardenedBoolTrue);
}
```

moidx:
Re `sec_mmio` transactions. The idea behind them is to catch any potential faults by applying redundant reads to registers previously accessed via `sec_mmio` functions. The redundant checks are executed by the following functions:

```c
  sec_mmio_check_values(rnd_uint32());
  sec_mmio_check_counters(/*expected_check_count=*/n);
```

timothytrippel:
Can this be marked non-draft and merged?

sw/device/silicon_creator/rom/rom.c
jesultra: Yes, I think that will be necessary.  I assume we want to let the owner code configure watchdog, and even lock its configuration, so we will need to state that if such owner code want to make use of deep sleep, how many ticks could elapse between waking up and control being handed back to owner code (or at least how much time could pass between ROM/ROM_EXT petting the watchdog, until handing over control to owner code.)

sw/device/silicon_creator/rom/rom.c
jesultra: If we wake up from low power mode, but before executing to the point of claring the reset reason register, the external POR_N is asserted (or we hit a watchdog, undervoltage or any other reset source), then as execution reaches this point we might find TWO or more bits in the register set, one of them being low power exit.  We want to treat the above situations as hard reset (not waking up from low power), since even the always on IP blocks would have had their state reset.  So that is why my logic seem to look at every bit except the "low power exit" one.

a-will: Hehe, just being pedantic here, but POR_N specifically would reset the reset reason register (rstmgr will be reset). There would only be the one bit set in that particular case.

jesultra: Good to know. But a watchdog reset or plain software reset (trap) while waking up from low power would result in AON IPs having their state reset, and a combination of bits in the reset cause register, so I still think that the logic of looking at every other bit in the register is valid.

jesultra:
> Yep it makes sense that you don't see any issues with the code from `rom_start.s`

Actually, I did eventually see issues with `rom_start.S` unconditionally attempting to put initial values for watchdog, before the "main" logic in `rom.c` applied settings based on OTP fuses.  After making the `rom.c` logic conditional on not waking from low power, it turns out that tests which do not lock watchdog settings, but wake from low power, will expect the OTP settings to apply also after wakeup, which previously was the case as `rom_start.S` first applied different settings, but then `rom.c` would set them back.  With this CL, `rom.c` no longer does so.  I have also made the logic in `rom_start.s` conditional, such that no part of ROM will touch watchdog settings on wakeup from low power.

PR #22594, opened 2024-04-16
[rv_dm,dv] Connect rst_lc_ni in tb.sv

rswarbrick:
This PR depends on #22591 (no real impact, but they are touching related lines and rebasing is annoying!). The second commit is the only one that's special to this PR and can reviewed on its own. The other PR should be merged first.

The commit that's unique to this PR just makes an extra clock/reset interface to drive the signals and wires it up to the block.


rswarbrick:
Sorry for the slow reply. The answer is "kind of"! The first step to wiring things up in existing tests is done in #22596, which depends on this PR. There are things that still aren't wired up, and there's a TODO comment (which moves with that PR) reminding us that they are pending.

marnovandermaas (approved):
Thanks for the explanation.

PR #22595, opened 2024-04-16
[rv_dm,rtl] Remove next_dm_addr_i port

rswarbrick:
This port was added to the RISC-V debug module that we vendor from pulp in order to support having a chain of debug modules. I don't believe we're planning to build anything like this in OpenTitan in the near term, so let's just tie it to zero rather than exposing the port from the block.


a-will:
This is needed for darjeeling, though. Is it a problem to leave it exposed and tied to zero as it is now?

rswarbrick:
Oh, I'm sorry: I hadn't realised that. I'll drop this and let's tie it off as you suggest.

PR #22596, opened 2024-04-16
[rv_dm,dv] Expose new ports in rv_dm_if

rswarbrick:
This builds on top of #22594 (which should be reviewed and merged first). Only the final commit is unique to this PR. Its commit message is:

*[rv_dm,dv] Expose new ports in rv_dm_if*

This change shouldn't cause any change to behaviour, but it's
connecting up the ports separately in tb.sv so that they can be
controlled by vseqs.

The logic that wires everything together (and the TODO comment) have
now been moved to the base vseq.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -52,6 +52,13 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     cfg.rv_dm_vif.scanmode <= scanmode;
     cfg.rv_dm_vif.unavailable <= unavailable;
 
+    // TODO(#23096): We're currently wiring all the enable signals to match lc_hw_debug_en and
+    //               hard-coding the late debug enable flag to be true. These eventually need to be
+    //               separately controlled.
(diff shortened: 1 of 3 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
```diff
@@ -52,6 +52,13 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     cfg.rv_dm_vif.scanmode <= scanmode;
     cfg.rv_dm_vif.unavailable <= unavailable;
 
+    // TODO: We're currently wiring all the enable signals to match lc_hw_debug_en and hard-coding
```
hcallahan-lowrisc, line 55: Just to avoid un-tagged TODOs if possible, is there anything (open issues, testpoints etc) that this could be linked to?

hcallahan-lowrisc (approved):

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
rswarbrick: Good point. There isn't yet, but I'll add one now.

PR #22598, opened 2024-04-16
[rv_dm,doc] Expand table in theory_of_operation.md

rswarbrick:
No functional change, but the resulting table should be much easier to read without needing to render it and then open in a browser.


vogelpi (approved):
LGTM!

Issue #21155, opened 2024-02-01
[edn] Consider modifying FIFO instantiations


vogelpi:
### Description

When reviewing #21142, I notices that inside EDN the FIFO primitives are not hardened and they output zero when empty. This is not ideal and could be changed at low design, area and DV overhead. Each FIFO already has error signals used for detecting cases when an empty FIFO is popped etc.

andreaskurth:
@vogelpi: It may not be possible to turn on hardening due to area optimizations, but we should disable output zero

vogelpi:
I turns out there are just two sync FIFOs that can be hardened and the area impact should be very low for both of them. I've filed a PR to do this here #22599.

PR #22599, opened 2024-04-16
[edn] Harden FIFO instantiations

vogelpi:
This resolves lowRISC/OpenTitan#21155.


vogelpi:
I am currently running a full EDN regression and will debug potential test failures / look into extending the edn_err sequence on THU.

rswarbrick:
Hi @vogelpi! This was failing quick lint for silly reasons (cmdgen blocks). I've just regenerated those and force-pushed to the branch. (No other change, so hit "compare" to see how little work I've done :-D)

hw/ip/edn/data/edn.hjson
```diff
@@ -603,9 +606,7 @@
         { bits: "30",
           name: "FIFO_STATE_ERR",
           desc: '''
-                This bit will be set to one when any of the source bits (bits 0 through 1 of this
-                this register) are asserted as a result of an error pulse generated from
-                any FIFO where both the empty and full status bits are set.
+                This bit will be set to one when any of the source bits (bits 0 through 1 of this register) are asserted as a result of an error pulse generated from any FIFO where both the empty and full status bits are set or in case of error conditions inside the hardened counters.
```
rswarbrick, line 609: I don't believe we've got a hard "one sentence to a line" rule for hjson files, so I'm not sure this is necessary. No problem, but it might be more comfortable to edit without the unwrapping step.

rswarbrick (approved):
Nitty comment about markdown in hjson files, but no change needed. This looks good to me.

hw/ip/edn/data/edn.hjson
vogelpi: You're right, we don't have a hard rule here (wasn't aware we have a separate style guide for hjson, see https://opentitan.org/book/doc/contributing/style_guides/hjson_usage_style.html) but I always tell people to do 1 sentence per line to ease future diffs.

vogelpi:
I am merging this despite the `i2c_target_test_fpga_cw310_sival` test failing. This PR doesn't impact the timing of anything in the system and I've seen this I2C test failing also for other PRs. I'll check if we should mark the test as flaky or broken.

PR #22600, opened 2024-04-17
[prim,fpv] Fix width of FPV variable in prim_arbiter_ppc.sv

rswarbrick:
A minor cleanup that avoids some tool warnings. Fortunately, it turns out that we had a width parameter floating around already so it was really easy to use!


vogelpi (approved):
LGTM!

PR #22558, opened 2024-04-15
Implement `chip_sw_rom_ctrl_digests` as part of `chip_sw_kmac_app_rom`

luismarques:

PR #22604, opened 2024-04-17
Implement chip_sw_rom_ctrl_digests as part of chip_sw_kmac_app_rom

pamaury:
Improvement on #22558, using this to check CI, will see if we keep this one or the other.


luismarques:
I'm happy to merge this PR instead of my original one.
It LGTM but I'm biased, other people should have a look at it.

The testplan should be updated as well. Do you want to do it in this PR or on a separate one, @pamaury?

moidx:
Hi @timothytrippel, can you review this PR since you are the person that last looked at the scrambling logic integration for Z1?

Thanks

pamaury:
@luismarques I might create another PR for the testplan, unless I need to rework this PR to fix issues.

hw/top_earlgrey/dv/chip_sim_cfg.hjson
```diff
@@ -1533,7 +1533,7 @@
     {
       name: chip_sw_kmac_app_rom
       uvm_test_seq: chip_sw_base_vseq
-      sw_images: ["//sw/device/tests:kmac_app_rom_test:1:new_rules"]
+      sw_images: ["//sw/device/tests:kmac_app_rom_test_sim_dv:1:new_rules"]
```
timothytrippel, line 1536: are you sure you need the suffix `_sim_dv`? I believe this gets automatically here: https://cs.opensource.google/opentitan/opentitan/+/master:hw/dv/tools/dvsim/sim.mk;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=87

rules/opentitan.bzl
```diff
@@ -426,9 +427,16 @@ def _elf_to_scrambled_rom_impl(ctx):
                 src.basename.replace("." + src.extension, ""),
             ),
         )
+        digest = ctx.actions.declare_file(
+            "{}.39.scr.vmem.digest.h".format(
+                # Remove ".elf" from file basename.
+                src.basename.replace("." + src.extension, ""),
```
timothytrippel, line 433: is it a digest of the VMEM file or a digest of the scrambled ROM? if the latter maybe: `"{}.39.scr.digest.h" ?

timothytrippel: Moreover, do we need to modify the bazel rules, or can you just use this target here: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/rom/BUILD;drc=742df4f26d8ca6a4a5b0877dc0d6e72ab388bdcd;l=505

And add one for the test rom if you need it?

timothytrippel (approved):
mostly LGTM, but dropped a couple comments/suggestions

hw/top_earlgrey/dv/chip_sim_cfg.hjson
pamaury: It is necessary in this case because we have to manually create one `opentitan_test` per exec-env. I'll add a comment for avoid any confusion.

rules/opentitan.bzl
pamaury: I will double check, @luismarques did the hash part, I just helped with the bazel.

hw/top_earlgrey/dv/chip_sim_cfg.hjson
timothytrippel: hmm ok, just to confirm this works b/c the sim.mk file I referenced is super brittle, did you try to run this in DV with the current label?

pamaury: Yes I ran it already to check

rules/opentitan.bzl
pamaury: I checked with @luismarques and this is different: this test checks that the ROM digest reported by ROM_CTRL matches what is computed by the scrambler. This is a cSHAKE hash. The code you pointed to uses SHA256 and the test is also different since the actually computes the hash instead of relying on the ROM_CTRL digest.
Another difference is this test is automated whereas the one you pointed to is a manual process for release I think?

hw/top_earlgrey/data/ip/chip_kmac_testplan.hjson
```diff
@@ -68,6 +68,9 @@
       stage: V2
       si_stage: NA
       tests: ["chip_sw_kmac_app_rom"]
+      bazel: [
+        "//sw/device/tests:kmac_app_rom_test",
+      ]
```
luismarques, line 73: Do we want to merge the `chip_sw_kmac_app_rom` and `chip_sw_rom_ctrl_digests` testplan tests? If not, what's the best way to update them?

pamaury: It would be useful to have the opinion of someone doing DV, are those tests doing the same thing in DV? On real device, there isn't much to do since if there is a mismatch, it would not be running!

luismarques: @rswarbrick any opinion about this?

rswarbrick: They can definitely be merged, I think. The `rom_ctrl` test has a little bit of extra work (to check the digest that's reported in the `rom_ctrl` CSRs), but I can't see any other real difference.

I don't really have an opinion about which testplan the resulting testpoint should sit in :-)

pamaury:
This PR will need to be manually cherry-pickd to master due to the changes of the ROM targets (mask_rom)

PR #22606, opened 2024-04-17
[bazel] Rename `cw310_params` to `fpga_params`

jwnrt:
This PR renames `cw310_params` to `fpga_params` in preparation for better CW340 support.

`fpga_params` was chosen because we're going to be using all FPGA boards roughly the same and therefore they will share the same parameters.

This change will allow us to add CW340 execution environments to tests without duplicating their `cw310_parameters` to `cw340_parameters`.

This PR adds an alias from `cw310_params -> fpga_params` and falls back to a parameter named `cw310 =` if `fpga =` is missing. This is used to prevent merge skew errors if pull requests are merged after this one which use the old names. After enough time, we can revert the last commit here to be fully on the `fpga_params` name.


pamaury (approved):
Thanks James, this looks great.

a-will:
For cases where the CW340 and CW310 are different (e.g. substantially worse SPI signal quality for the former vs. the latter, limiting I/O speeds*), is the thinking that we would split their configs off into different named parameters for opentitan_test() targets on a case-by-case basis?

Is the thinking that most tests are driven similarly in opentitanlib, so it's better to default to the same config vs. requiring every target to individually handle the two boards?

*This probably would be handled in opentitanlib config files anyway, but pretend I thought of something that might change the params ;)

rules/opentitan/defs.bzl
```diff
@@ -124,10 +124,8 @@ CLEAR_TEST_CMD = " "
 def _parameter_name(env, pname):
     if not pname:
         (_, suffix) = env.split(":")
-        if "cw310" in suffix:
-            pname = "cw310"
-        elif "cw340" in suffix:
-            pname = "cw340"
+        if "fpga" in suffix:
+            pname = "fpga"
```
nbdd0121, line 128: I wonder if we should try use cw310/cw340 if they are present, and then try "fpga" as a fallback. If we decide in the future to future merge fpga_params and silicon_params, then it can try cw310->fpga->common.

jwnrt: So `fpga` would only be a fallback and not merged with `cw310`/`cw340`?
I think that could work, but I'll do it in a follow up.

jwnrt: I've added an extra commit which makes `cw310` the fallback parameter name if `fpga` isn't provided, but I've only written it to avoid merge skew errors and it's not intended as a full solution.

We can revisit this later if we want to fallback behaviour properly.

engdoreis (approved):

Issue #20119, opened 2023-10-19
[chip-test] chip_sw_otbn_mem_scramble


GregAC:
### Test point name

[chip_sw_otbn_mem_scramble](https://github.com/lowRISC/opentitan/blob/d1676c071ab51d2f4921beaa798fe7148b890c90/hw/top_earlgrey/data/ip/chip_otbn_testplan.hjson#L101-L129)

### Host side component

None

### OpenTitanTool infrastructure implemented

None

### Silicon Validation (SiVal)

Yes

### Emulation Targets

- [ ] None
- [X] CW310
- [ ] Hyperdebug + CW310

### Contact person

@GregAC

### Checklist

Please fill out this checklist as items are completed. Link to PRs and issues as appropriate.
- [ ] Check if existing test covers most or all of this testpoint (if so, either extend said test to cover all points, or skip the next 3 checkboxes)
- [ ] Device-side (C) component developed
- [ ] Bazel build rules developed
- [ ] Host-side component developed
- [ ] HJSON test plan updated with test name (so it shows up in the dashboard)
- [ ] Test added to dvsim nightly regression (and passing at time of checking)
- [ ] For SiVal test cases, test is running relevant FPGA or silicon regression

> estimate 0.5

HU90m:
The existing `//sw/device/tests:otbn_mem_scramble_test` hasn't run in CI for as far back as the history goes.

After scrambling, ECC errors when reading values from IMEM should trigger interrupts but they don't appear to be:

```
I00051 otbn_mem_scramble_test.c:180] Mismatch without integrity error: Entry 49, address 0x29f, data 0x6984800c
E00052 otbn_mem_scramble_test.c:244] CHECK-fail: Expecting at least 48 IMEM integrity errors, got 0
I00053 status.c:34] FAIL!
```

luismarques:
This test works everywhere except the FPGA, and *very occasionally* also fails on dvsim.

I found a couple of dvsim VCS seeds that fail:

- 105706649916084870360541638213197649019961057817534949222100493257430194194456
- 84178660973699573356153380528591368847312397780491565840024634280044710784108

dvsim reproducer:

```
./util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_otbn_mem_scramble --tool vcs --reseed 1 --seeds 105706649916084870360541638213197649019961057817534949222100493257430194194456
```

[...]

PR #22610, opened 2024-04-17
[sival] Port `sram_ctrl_scrambled_access_test` to sival

jwnrt:
This PR ports the `sram_ctrl_scrambled_access_test` to work outside of simulation by making the following changes:

* Making the test-bench synchronisation code no-ops. We don't have a test bench outside simulation.
* Replacing the backdoor comparison:
  * The simulation version of this test scrambles SRAM and then receives the expected scrambled contents from the simulation test bench.
  * We cannot provide this outside of simulation.
  * Instead, we check that the SRAM no longer contains the pattern we wrote to it, indicating it was scrambled.
* Disabled the ECC error checking expectations when running on an FPGA.
  * This code checks that we receive the expected number of ECC errors that should have been triggered by the scrambling of SRAM.
  * We cannot seem to catch these errors in the interrupt handler on the FPGA.
  * See #20119 for more details.
  * Since this was an extra feature added to the test, it has been disabled outside of DVSIM and silicon environments.


luismarques (approved):
I think all of the issues we discussed in person have been addressed. LGTM.

engdoreis (approved):
LGTM

pamaury (approved):

PR #22612, opened 2024-04-17
[pentest] Fix extclk speed config

nasahlpa:
This commit fixes the external clock speed selection configuration. Accidentally, a MUBI value was passed instead of a boolean to select the slow or high speed clock.


timothytrippel (approved):

vogelpi:
The CI failure is not related to this PR (not touching SiVal code). I am merging this.

PR #22613, opened 2024-04-17
[pentest] Constant trigger window for OTBN

nasahlpa:
For side-channel and fault injection analysis, a constant trigger window is required. However, currently, for OTBN analysis, this is not the case as entropy complex reseeds are issued during OTBN execution.

This commit adds the SCA & FI specific sca_configure_entropy_source_otbn() function. This function is a copy of entropy_testutils_auto_mode_init() but sets the reseed_interval of EDN0...4 to the max. available value, i.e., 0xffffffff.

With this change, the trigger window stays constant for OTBN measurements.

Thanks @jadephilipoom for helping to debug this!


sw/device/tests/penetrationtests/firmware/sca_lib.c
```diff
@@ -21,6 +27,112 @@ static dif_rv_plic_t plic;
 static dif_rstmgr_t rstmgr;
 static dif_alert_handler_t alert_handler;
 
+status_t sca_configure_entropy_source_otbn(void) {
```
jadephilipoom, line 30: nit: I'd drop the `_otbn`, because this routine will configure EDN for all blocks, not just OTBN. Maybe just `sca_configure_entropy_source`, or `sca_configure_entropy_source_max_reseed_interval`?

jadephilipoom (approved):

sw/device/tests/penetrationtests/firmware/sca_lib.c
nasahlpa: Thanks, good point! I've renamed the function to `sca_configure_entropy_source_max_reseed_interval`

PR #22616, opened 2024-04-17
Add a bazel rule to start a jtag server

cfrantz:
Add a bazel rule to configure the chip for jtag and start an OpenOCD instance as a jtag server:

```
bazel run //hw/top_earlgrey:jtag_server
```

### DIFF
--- a/hw/top_earlgrey/BUILD
+++ b/hw/top_earlgrey/BUILD
@@ -14,6 +14,7 @@ load(
     "sim_dv",
     "sim_verilator",
 )
+load("//rules/opentitan:jtag.bzl", "jtag_server")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -41,6 +42,29 @@ filegroup(
     ],
 )
 
+label_flag(
+    name = "jtag_env",
+    build_setting_default = ":silicon_creator",
+)
+
+jtag_server(
+    name = "jtag_server",
+    testonly = True,
+    chip_setup = [
+        "gpio write TAP_STRAP0 false",
+        "gpio write TAP_STRAP1 true",
+        "gpio apply RMA_BOOTSTRAP",
+        "gpio apply RESET",
+        "no-op --delay 50ms",
+        "gpio remove RESET",
+    ],
+    exec_env = ":jtag_env",
+    openocd_chip_config = "//util/openocd/target:lowrisc-earlgrey.cfg",
+    openocd_commands = [
+        "adapter speed 500; transport select jtag; reset_config trst_only",
+    ],
+)
+
 ###########################################################################
 # FPGA CW310 Environments
 ###########################################################################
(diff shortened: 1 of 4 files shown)


cfrantz:
Q: Where is an appropriate location for this rule?  I have half a mind to locate it in `//third_party/openocd`, as in:
```
bazel run //third_party/openocd:jtag_server --//third_party/openocd:jtag_env=//hw/top_earlgry:silicon_creator
```

However, the chip initialization sequence (setting tap straps, etc) feels a bit more earlgrey specific, so I located it in `//hw/top_earlgrey`.

moidx (approved):
Nice thanks! 

Can you cross reference this in the following docs?

* doc/getting_started/using_openocd.md
* doc/getting_started/setup_fpga.md

timothytrippel (approved):

cfrantz:
> Nice thanks!
> 
> Can you cross reference this in the following docs?
> 
> * doc/getting_started/using_openocd.md
> * doc/getting_started/setup_fpga.md

Done.

cfrantz:
What, if anything, do we need to change or adjust to allow this to start a JTAG (openocd) server for the CW340?

hw/top_earlgrey/BUILD
```diff
@@ -141,6 +165,7 @@ fpga_cw310(
     base_bitstream = "//hw/bitstream/hyperdebug:bitstream",
     exec_env = "fpga_hyper310_rom_with_fake_keys",
     manifest = "//sw/device/silicon_creator/rom_ext:manifest_standard",
+    openocd_adapter_config = "//third_party/openocd:jtag_cmsis_dap_adapter_cfg",
```
a-will, line 168: `:fpga_cw340` likely needs this assignment too. Otherwise, I believe it should behave like the hyper310.

engdoreis: I've never used the Hyperdebug CMSIS-DAP interface, but I believe it should work just like on the CW310.
Additionally, the CW340 has an embedded FTDI, and this is the config file.
https://github.com/lowRISC/opentitan/blob/master/util/openocd/board/cw340_ftdi.cfg
This is how you connect:
https://github.com/lowRISC/opentitan/blob/master/doc/getting_started/setup_fpga.md#connecting-openocd

a-will: Ah, for CMSIS-DAP to work, the hyperdebug would need to be connected to the IOR pins via the ribbon cable. It depends on the configuration in CI.

cfrantz: This rule isn't meant to run in CI - it's meant to be used by developers who want to start an OpenOCD server:

```
bazel run //third_party/openocd:jtag_server --//third_party/openocd:jtag_env=<some env>
```

You can then use `gdb` (or `telnet`) to go debug.  The rule executes a script that automates the necessary setup to get JTAG connected.

Adding the adapter config to the rules isn't too much of a big deal.  Hopefully that's all that will be needed to make this rule useful to CW340 users.

a-will: > Adding the adapter config to the rules isn't too much of a big deal. Hopefully that's all that will be needed to make this rule useful to CW340 users.

Yeah, selecting the adapter config might work best for CW340. It's a funky one where the ribbon cable's position determines which adapter to select, and that's sort of orthogonal to the execution environment. I'm not sure it's in our interest to support every permutation in general, but for the JTAG server w/ CW340, maybe it's nicer to be able to select any config directly, without the execution environment proxy.

PR #22617, opened 2024-04-17
[pentest] Return alerts as array

nasahlpa:
As for the pen. testing evaluating an array instead of individual variables for the triggered alerts is easier, this commit switches the format of how the alerts are returned to the host.


pamaury (approved):
The change looks ok to me but once again I wonder if some refactor could be done, it looks a bit suspicious that this piece of code is duplicated so many times.

nasahlpa:
Thanks for your review @pamaury! Not sure how to avoid the duplicated code. Each function inside, for example, `ibex_fi.c` is an independent test that gets executed standalone. Each of these tests need to return different test results, only the `err_status` and the `alerts` field stay the same. For the test framework initiating the fault injections it is important that only a single uJSON message gets transmitted back from the host.

PR #22619, opened 2024-04-17
Setup priority over stall response

Nadia-Hanif:
This Pr contains sequence that checks the setup and stalling priority.
Specific Endpoint is stalled in OUT transaction than after wards setup transaction will be sent.
instead of stalling the endpoint setup transaction will take priority.
This priority will be check using the out_stall and in_stall register.


elliotb-lowrisc:
Closing as no longer relevant

PR #22621, opened 2024-04-17
Device timeout sequence

Nadia-Hanif:
This PR contain device timeout sequence in which token packet is sent and than data packet sent after 18 bit times so that device timed out. This will check if we get the timeout pid in response.


elliotb-lowrisc:
Closing as no longer relevant

PR #22627, opened 2024-04-17
[cert] add cert lib with util functions / structs

timothytrippel:
This adds a cert library in the silicon_creator dir that contains helper functions and structs for managing OpenTitan certificates. Specifically a function is added to check if a cert's serial number matches what is expected. This function will be used to determine whether CDI0/1 certs need to be updated in the ROM_EXT (if the ROM_EXT or OWNER firmware is ever updated).

### DIFF
--- a/sw/device/silicon_creator/lib/cert/cert.h
+++ b/sw/device/silicon_creator/lib/cert/cert.h
@@ -1,0 +1,58 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef OPENTITAN_SW_DEVICE_SILICON_CREATOR_LIB_CERT_CERT_H_
+#define OPENTITAN_SW_DEVICE_SILICON_CREATOR_LIB_CERT_CERT_H_
+
+#include "sw/device/lib/base/hardened.h"
+#include "sw/device/silicon_creator/lib/drivers/flash_ctrl.h"
+#include "sw/device/silicon_creator/lib/error.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum {
+  /**
+   * Offset to the ASN.1 DER encoded serial number of an X.509 certificate.
+   */
+  kCertX509Asn1SerialNumberByteOffset = 13,
+
+  /**
+   * Sizes of the ASN.1 DER encoded serial number of an X.509 certificate.
+   */
+  kCertX509Asn1SerialNumberSizeInBytes = 20,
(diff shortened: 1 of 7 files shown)


sw/device/silicon_creator/lib/cert/cert_unittest.cc
```diff
@@ -0,0 +1,215 @@
...
+          uint32_t *out_words = static_cast<uint32_t *>(out);
+          for (size_t i = 0; i < num_words; ++i) {
+            out_words[i] = data[i];
+          }
+          return error;
+        });
+  }
+  void ExpectFlashCtrlErrorCodeGet(flash_ctrl_error_code_t *in_error_code) {
+    EXPECT_CALL(flash_ctrl_, ErrorCodeGet(_))
+        .WillOnce([in_error_code](flash_ctrl_error_code_t *out_error_code) {
```
cfrantz, line 107: Will `SetArgPointee` work here?
```
using ::testing::SetArgPointee;
...
        .WillOnce(SetArgPointee<0>(*in_error_code));
```

sw/device/silicon_creator/lib/cert/cert.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @param info_page Pointer to the flash info page the certificate is on.
+ * @param expected_sn_words Expected serial number words (in big endian order).
+ * @param[out] matches True if expected serial number found. False otherwise.
+ * @return The result of the operation.
+ */
+OT_WARN_UNUSED_RESULT
+rom_error_t cert_x509_asn1_check_serial_number(
+    const flash_ctrl_info_page_t *info_page, uint32_t *expected_sn_words,
+    multi_bit_bool_t *matches);
```
cfrantz, line 52: Why `mutli_bit_bool_t` instead of `hardened_bool_t`?

cfrantz (approved):

sw/device/silicon_creator/lib/cert/cert.h
timothytrippel: Fixed.

sw/device/silicon_creator/lib/cert/cert_unittest.cc
timothytrippel: Fixed.

moidx (approved):

pamaury (approved):
This looks good to me, although potential a bit fragile with the offset being hardcoded. An alternative (which is more complex) would be that the asn1 generator outputs the offset and size of this field and this information could be stored in the flash.

timothytrippel:
Good point @pamaury. I will capture your comment in a separate issue and track in the manufacturing board.

PR #22629, opened 2024-04-17
[ownership] Parse the owner block

cfrantz:
1. Parse the owner block, extracting useful configuration structures.
2. Configure the flash and flash-info regions based on the configuration.
3. Add tests for the parser and configuration functions.
4. Add several basic owner blocks with errors to test the parser.


sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -0,0 +1,175 @@
...
+  }
+  return kErrorOk;
+}
+
+#define MUBI(ex_) (ex_) ? kMultiBitBool4True : kMultiBitBool4False
+
+rom_error_t owner_block_flash_apply(const owner_flash_config_t *flash,
+                                    uint32_t config_side,
+                                    uint32_t primary_side) {
+  if (!flash)
```
moidx, line 78: Is there a case where this evaluates to true in production environments? If not, It would probably be good to replace it for a hardened check. Same for other functions.

We can chat to @johannheyszl about the likelihood of faults to zero in the current implementation.

sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -0,0 +1,175 @@
...
+               sizeof(owner_flash_region_t);
+  if (len >= 8) {
+    return kErrorOwnershipFlashConfigLenth;
+  }
+
+  const owner_flash_region_t *config = flash->config;
+  for (size_t i = 0; i < len; ++i, ++config) {
+    if (config->start >= start && config->start + config->size <= end) {
+      flash_ctrl_cfg_t cfg = {
+          .scrambling = MUBI(config->properties & kOwnerFlashPropertyScramble),
```
moidx, line 96: If the hardware supports MUBI, maybe we should consider using MUBI in the data structure as well. This way we don't have to worry about faults here. 

Alternatively, we could argue if leaving as-is would be beneficial for an attacker. The read path should be protected here. In which case, the macro here just makes the attack exposure at the dual lockstep side. 

Out of all these settings, I think the most critical ones are the read/write access.

sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -0,0 +1,175 @@
...
+        perm.write = kMultiBitBool4False;
+        perm.erase = kMultiBitBool4False;
+      }
+      flash_ctrl_data_region_protect(i, config->start, config->size, perm, cfg);
+    }
+  }
+  return kErrorOk;
+}
+
+static inline bool is_owner_page(const owner_info_page_t *config) {
```
moidx, line 116: `hardened_bool_t` here to avoid comparisons to zero in the calling side.

sw/device/silicon_creator/lib/ownership/owner_block_unittest.cc
```diff
@@ -0,0 +1,325 @@
...
+#include "gtest/gtest.h"
+#include "sw/device/lib/testing/binary_blob.h"
+#include "sw/device/silicon_creator/lib/boot_data.h"
+#include "sw/device/silicon_creator/lib/drivers/mock_flash_ctrl.h"
+#include "sw/device/silicon_creator/lib/error.h"
+#include "sw/device/silicon_creator/lib/ownership/datatypes.h"
+#include "sw/device/silicon_creator/testing/rom_test.h"
+
+namespace {
+#include "sw/device/silicon_creator/lib/ownership/testdata/basic_owner.h"
```
moidx, line 19: nit: `"basic_owner_testdata.h"` or `"basic_owner_testdata.inc"`

sw/device/silicon_creator/lib/ownership/owner_block.c
cfrantz: Done.

sw/device/silicon_creator/lib/ownership/owner_block_unittest.cc
cfrantz: Done.

sw/device/silicon_creator/lib/ownership/owner_block.c
johannheyszl: @cfrantz and I have chatted

sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -0,0 +1,175 @@
...
+  config->unknown_tags = 0;
+  config->sram_exec = block->sram_exec_mode;
+  config->flash = NULL;
+  config->info = NULL;
+  config->rescue = NULL;
+  uint32_t remain = sizeof(block->data);
+  uint32_t offset = 0;
+  while (remain) {
+    const tlv_header_t *item = (const tlv_header_t *)(block->data + offset);
+    if (item->tag == 0 || item->length == 0) {
```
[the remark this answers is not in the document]
cfrantz: Fixed.  I'm using the pattern `0x5a5a5a5a` (aka `ZZZZ`) as the end condition.

sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -0,0 +1,175 @@
...
+      break;
+    }
+    if (item->length > remain) {
+      return kErrorOwnershipInvalidTagLength;
+    }
+    remain -= item->length;
+    offset += item->length;
+    switch (item->tag) {
```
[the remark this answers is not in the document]
cfrantz: Done.

sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -0,0 +1,175 @@
...
+          return kErrorOwnershipDuplicateItem;
+        config->info = (const owner_flash_info_config_t *)item;
+        break;
+      case kTlvTagRescueConfig:
+        if (config->rescue)
+          return kErrorOwnershipDuplicateItem;
+        config->rescue = (const owner_rescue_config_t *)item;
+        break;
+      default:
+        /* unrecognized item. skip. */
```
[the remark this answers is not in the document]
cfrantz: I've changed this to be an error condition: an invalid tag will `return kErrorOwnershipInvalidTag`.

sw/device/silicon_creator/lib/ownership/owner_block.c
cfrantz: I've changed the use of NULL to use `kHardenedBoolFalse`.

sw/device/silicon_creator/lib/ownership/owner_block.c
cfrantz: I've migrated the flash config fields to use MUBI4 and to xor each of the configuration words with `0x11111111 * index` to create a data-dependency between the index of the item and the configuration data.

moidx:
Hi @cfrantz. LGTM.

PR #22659, opened 2024-04-17
[rom_ext] Lock the address translation windows

cfrantz:
Lock the address translation windows before jumping to owner code so they cannot be abused to read/write memory regions disabled by the ePMP.


sw/device/silicon_creator/lib/drivers/ibex.c
```diff
@@ -51,6 +52,13 @@ void ibex_addr_remap_1_set(uint32_t matching_addr, uint32_t remap_addr,
   icache_invalidate();
 }
 
+void ibex_addr_remap_lockdown(uint32_t index) {
+  HARDENED_CHECK_LT(index, 2);
+  index *= sizeof(uint32_t);
+  abs_mmio_write32(kBase + RV_CORE_IBEX_IBUS_REGWEN_0_REG_OFFSET + index, 0);
+  abs_mmio_write32(kBase + RV_CORE_IBEX_DBUS_REGWEN_0_REG_OFFSET + index, 0);
```
moidx, line 59: Can you read back the values to make sure they were written? It may be worth doing the read back check as a separate inline call so that we can call `ibex_addr_remap_lockdown` before `rom_ext_sram_exec()` in the rom followed by the read-back check. This is to deter against faults aiming to skip this function call.

moidx (approved):
Thanks @cfrantz! Should we aim to update the ROM_EXT major, minor and SVN versions in the manifest?

sw/device/silicon_creator/lib/drivers/ibex.c
cfrantz: Done:  I used `sec_mmio` to lock out the translation windows and then checked values before jumping to owner code.

nbdd0121:
I've found that this change means that rv_core_ibex_address_translation_test cannot be made to work in cw310_sival_rom_ext anymore (which I was trying to do in #22826)

cfrantz:
> I've found that this change means that rv_core_ibex_address_translation_test cannot be made to work in cw310_sival_rom_ext anymore (which I was trying to do in #22826)

That test should be disqualified from the sival_rom_ext environment and just run under the `fpga_cw310_sival` environment.  Unfortunately, the address translation windows are far too flexible to be left open for owner code as they can arbitrarily remap to any address in the CPU's address space (ie: they can be used to defeat the ePMP lockouts or any other ePMP restrictions).

PR #22703, opened 2024-04-18
[sw] Dispatch SAM3X commands from the hyper310 transport

jwnrt:
This allows the hyper310 transport to still reset the SAM3X etc. when connected to a ChipWhisperer FPGA.


pamaury (approved):
This looks good to me. Although if we are going down this route, we might consider having more general `reset` and `get-fw-version`, not just for the SAM3x. But maybe can be done in another PR.

jwnrt:
I agree, will take a look after this merges

PR #22706, opened 2024-04-18
[sival, rv_dm] fix rv_dm_mem_access flakiness

nbdd0121:
The test is randomised and sometimes will try to access the last 32 bytes of the ROM which is not accessible via SBA and fail. Exclude these 32 bytes from the test.

### DIFF
--- a/sw/host/tests/chip/rv_dm/src/mem_access.rs
+++ b/sw/host/tests/chip/rv_dm/src/mem_access.rs
@@ -12,6 +12,7 @@ use opentitanlib::app::TransportWrapper;
 use opentitanlib::execute_test;
 use opentitanlib::io::jtag::JtagTap;
 use opentitanlib::test_utils::init::InitializeTest;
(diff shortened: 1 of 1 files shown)


sw/host/tests/chip/rv_dm/src/mem_access.rs
```diff
@@ -26,12 +27,15 @@ struct Opts {
...
 
     /// Seed for random number generator.
-    #[arg(long)]
+    #[arg(long, value_parser = u64::from_str)]
     seed: Option<u64>,
 }
 
 const NUM_ACCESSES_PER_REGION: usize = 32;
 
+// The last 32 bytes of ROM are not accessible.
```
luismarques, line 36: maybe explain that's where the ROM digest is, for some context?

luismarques (approved):

jwnrt (approved):

PR #22708, opened 2024-04-18
Fix chip_power_idle_load for SiVal ROM_EXT environment

luismarques:
~~The first commit (cherr-picked from #22579) enables opentitan_test to take per-exec-env defines.~~ (now committed)
The second commit fixes the test, by disabling OTP access in SiVal ROM_EXT environments.

### DIFF
--- a/sw/device/tests/chip_power_idle_load.c
+++ b/sw/device/tests/chip_power_idle_load.c
@@ -4,6 +4,7 @@
 
 #include <stdint.h>
 
+#include "sw/device/lib/arch/boot_stage.h"
 #include "sw/device/lib/base/mmio.h"
 #include "sw/device/lib/dif/dif_alert_handler.h"
 #include "sw/device/lib/dif/dif_gpio.h"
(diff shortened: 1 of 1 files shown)


cfrantz (approved):

sw/device/tests/chip_power_idle_load.c
```diff
@@ -269,6 +269,7 @@ bool test_main(void) {
 
   // OTP
 
+#ifndef SIVAL_ROM_EXT
```
jwnrt, line 272: Could we use `if (kBootStage >= kBootStageRomExt)` here instead? It means we can use the same binary for all these test environments.

The constants are quite new and are in `sw/device/lib/arch/boot_stage.h`.

jwnrt (approved):
LGTM but I've added a suggestion

sw/device/tests/chip_power_idle_load.c
moidx: There are a couple of reasons in favor for assuming OTP in locked in test cases running in post ROM_EXT stages:

1. test management: Supporting OTP mutability in SiVal test cases becomes difficult to maintain in silicon targets. The current approach is to only support the `silicon_creator` exec_env for test cases requiring OTP mutability. Such test cases need to be run by hand inside a secure environment. The by hand part due to limited access to silicon (some tests require a new chip per run), and the secure environment part required by the classification of AST configuration values. 
2. security: Devices available outside secure environments are required to be provisioned. This results in the OTP partition being locked. Having a prod-signed ROM_EXT with access to OTP enabled is currently considered as a vulnerability from an asset classification perspective. 

We also have some style guide restrictions wrt this, we should try to avoid conditionally compilation of code as it becomes difficult to maintain over time. More reasoning on this is captured in the style guide: https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros

Given the way `SIVAL_ROM_EXT` definition is used, I recommend to just switch to the use of `kBootStage`. All test cases running post ROM_EXT stage will most likely have to: 

1. Be aware of the stage they are running at; and,
2. Opportunistically check the IP status before interacting with it. Rather than always doing this, we need to be careful to only apply these checks when it is reasonable to do it, and without compromising DV coverage for the `sim_dv` configuration.

PR #22716, opened 2024-04-19
[entropy] reduce reseed interval in cryptolib

timothytrippel:
This reduces the reseed interval for EDN0 and 1 which was causing OTBN program execution to appear to hang in the `ft_personalize` binary during an attestation key generation operation.

Specifically, when working on #22714, I noticed that OTBN execution would hang during the UDS attestation (ECC P256) key generation. It would not error out, rather it would just hang, causing `//sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup:ft_provision_fpga_cw310_sival` to timeout eventually. (After attaching a JTAG/OpenOCD/GDB setup, I was able to determine the location of the hang.)

A few things we noticed were:
1. [moving the loading of the OTBN Boot Services program to before the entropy complex was reconfigured for keymgr operations](https://github.com/lowRISC/opentitan/pull/22714/commits/9deb11db10bdce8b5764e2691471a7a2d52f3873) inside the `ft_personalize` binary made the problem go away, and
2. making the modification made in this PR also made the problem go away.

As result, the thought is that entropy requests are getting dropped somewhere in the pipeline of FIFOs between entropy_src --> CSRNG --> EDN. This should probably be investigated further via waveforms. If we increase the reseed intervals (opposite of what is done in this PR) we may be able to trigger some DV testcases to fail. Alternatively, I could try to create a failing DV testcase based on parts of the `ft_personalize` test mentioned above.

CC: @vogelpi

### DIFF
--- a/sw/device/lib/crypto/drivers/entropy.c
+++ b/sw/device/lib/crypto/drivers/entropy.c
@@ -259,7 +259,7 @@ static const entropy_complex_config_t
                                 .id = kEntropyDrbgOpGenerate,
                                 .disable_trng_input = kHardenedBoolFalse,
                                 .seed_material = NULL,
-                                .generate_len = 8,
+                                .generate_len = 32,
                             },
                         .reseed =
                             {
@@ -284,7 +284,7 @@ static const entropy_complex_config_t
                             {
                                 .id = kEntropyDrbgOpGenerate,
                                 .seed_material = NULL,
-                                .generate_len = 1,
+                                .generate_len = 4,
                             },
                         .reseed =
                             {


timothytrippel:
CC: @h-filali this may fix the issue you saw in #22114. Could you try applying this on top to see if it fixes it?

vogelpi:
Thanks for your investigations @timothytrippel and for reporting your findings!

Dropping entropy requests really shouldn't be happening. We have quite some block-level DV in place now to properly verify all interfaces between the blocks. Based on your findings and after having a look at the software code I believe what's really happening is that the entropy complex is being reconfigured (using `entropy_complex_init()`) without being shutdown first: The `entropy_complex_stop_all()` defined in the cryptolib is called nowhere in the code. I have a strong suspicion that ROM_EXT configures the entropy complex in auto mode and during the test it's reconfigured without a proper shutdown first. This is when requests of the EDNs get into CSRNG and are left hanging once CSRNG is reconfigured. Changing the reseed interval changes the frequency of requests but doesn't solve the problem properly - we might just be lucky as the reconfiguration now happens between the issuing of requests between the EDNs and CSRNG.

Later today, I will try to push a fix into @h-filali 's PR and see if it really works. I'll report back here.

vogelpi:
Oh, no! I just noticed that we indeed shutdown the entropy complex before re-configuring it:
```c
status_t entropy_complex_init(void) {
  entropy_complex_stop_all();

  const entropy_complex_config_t *config =
      &kEntropyComplexConfigs[kEntropyComplexConfigIdContinuous];
  if (launder32(config->id) != kEntropyComplexConfigIdContinuous) {
    return OTCRYPTO_RECOV_ERR;
  }

  HARDENED_TRY(entropy_src_configure(&config->entropy_src));
  csrng_configure();
  HARDENED_TRY(edn_configure(&config->edn0));
  return edn_configure(&config->edn1);
}
```

timothytrippel:
One thing to clarify @vogelpi is the test I was observing this behavior on (`//sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup:ft_provision_fpga_cw310_sival`) does not have a ROM_EXT in place. It runs at the ROM_EXT stage itself (using the OTTF). So the entropy complex configuration that is in place at the start of the test should be that of what the ROM sets up (i.e. in boot mode).

Did this change help with #22114?

moidx (approved):

vogelpi:
Thanks a lot @timothytrippel for this. @h-filali checked #22114 with this and it works. I am merging the PR now and will create an issue to debug the underlying issue before M4.

vogelpi:
Note: I've merged this PR despite the following CI failures:
- CW310 SiVal non-ROM_EXT Tests: i2c_target_test_fpga_cw310_sival failed, PR #22729 marks this as broken.
- EDN smoke test failed: this got fixed by https://github.com/lowRISC/opentitan/pull/22736

PR #22717, opened 2024-04-19
[pentest] Split firmware into SCA & FI

nasahlpa:
Due to the rising number of tests added to the pen testing framework, the combined image exceeds the available IMEM. Hence, this commit splits up the binary into a separate FI and SCA binary for the chip. On the FPGA, still the combined image can be used.


timothytrippel:
Looks like there are conflicts that are preventing merging?

nasahlpa:
> Looks like there are conflicts that are preventing merging?

Thanks, PR is now rebased.

PR #22723, opened 2024-04-19
Fix chip_sw_sensor_ctrl_ast_alerts for SiVal silicon

luismarques:
On silicon the test would fail because there would be events being triggered before the (expected) ones being triggered by the test. This patch clears prior events, to start from a clean slate and prevent the test from failing.

I don't know if the fact that there are any prior events at all reveals an unexpected problem, and thus if this fix is sweeping that problem under the rug...? Does anyone have any insights into that?

### DIFF
--- a/sw/device/tests/sensor_ctrl_wakeup.c
+++ b/sw/device/tests/sensor_ctrl_wakeup.c
@@ -88,6 +88,11 @@ bool test_main(void) {
   dif_pwrmgr_domain_config_t sleep_config =
       kDifPwrmgrDomainOptionMainPowerInLowPower;
 
+  // Clear any prior events before we start the test
+  for (size_t i = 0; i < sensor_ctrl_events; ++i) {
+    CHECK_DIF_OK(dif_sensor_ctrl_clear_recov_event(&sensor_ctrl, i));
+  }
+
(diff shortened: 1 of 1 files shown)


nbdd0121 (approved):
Not familiar with sensor_ctrl block,  but the change LGTM

pamaury (approved):
This makes sense but might indicate a problem, like with the csrng when we discovered that the ROM_EXT contained an error. It's probably worth investigating a bit more.

Tagging @andreaskurth @moidx

andreaskurth:
AFAIK we don't expect any alerts from sensor_ctrl before this code
https://github.com/lowRISC/opentitan/blob/2906ca5fa4ca47894222df83539136a6d416a32c/sw/device/tests/sensor_ctrl_wakeup.c#L98-L100
in this test, right?

So I think if there are alerts from sensor_ctrl before that, we need to know where they come from and potentially update ROM / ROM_EXT to handle these. Just clearing the alerts in this test masks only masks the problem.

luismarques:
We probably want to fix this not by using this PR but by masking the event in the ROM_EXT, per solution (b) of #22708 22778. I'm closing this PR, at least for now.

PR #22726, opened 2024-04-19
[usbdev,dv] Adds usbdev_phy_config_eop_single_bit_handling_vseq

MubashirSaleem775:
Adds new sequence to verify usb device funtionality of receiving single bit End of packet (EoP) instead of 2 bits in normal.


MubashirSaleem775:
@alees24

alees24:

PR #22730, opened 2024-04-19
[Sival, pwrmgr] Fix flakyness on Silicon

engdoreis:
The test was `pwrmgr_all_reset_reqs_test_fpga_cw310_sival_rom_ext` was flaky because occasionally an unexpected reset (0x40 = `kDifRstmgrResetInfoEscalation`) would occur and mess up with the test.
This PR moves the alert handler configuration to phase 3 of the test.


timothytrippel (approved):

engdoreis:
Merging because the CI failure is not related.

Issue #22733, opened 2024-04-19
[test-triage] The chip_sw_uart_smoketest_signed test fails to run


rswarbrick:
### Hierarchy of regression failure

Chip Level

### Failure Description

This fails at run time, but the failure is a Bazel build step. This tries to run `uart_smoketest_signed_sim_dv` but that test was removed by commit 2cfed61d6.

I suspect we just need to revert that change, but there might be another way to get this working. If we revert the change, I suggest leaving a comment above the `opentitan_functest` line that explains what this is used for.

### Steps to Reproduce

- GitHub Revision: ec711524
- dvsim invocation command to reproduce the failure, inclusive of build and run seeds:
  ./util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_uart_smoketest_signed --fixed-seed 1

rswarbrick:
Assigning @jwnrt because the change that breaks the build was from you. If there's a better person, please feel free to reassign.

a-will:
Rather than bring back another user of the old opentitan_functest() rule, perhaps we ought to create a DV execution environment that runs atop ROM w/ signed tests.

jwnrt:
I've made a reverting PR for now but if you know how to get that working @a-will I'd be happy to review.

Sorry for the breakage @rswarbrick

a-will:
I think I've got something that could work, but the test currently times out doing an uber slow validation of something (might be that it doesn't like how things are signed yet...).

I suppose we could start with the revert (but only with the sim_dv target remaining).

a-will:
I've posted my work at https://github.com/lowRISC/opentitan/pull/22739

rswarbrick:
Sadly, the test is still failing, but I'm not sure whether it's related to the build problem. The failure mode is:
```
Test chip_sw_uart_smoketest_signed has 3 failures.

    0.chip_sw_uart_smoketest_signed.2910630305322550404265939249394086456704288511190492659161499184475528669513
    Line 934, in log /container/opentitan-public/scratch/os_regression/chip_earlgrey_asic-sim-vcs/0.chip_sw_uart_smoketest_signed/latest/run.log

      UVM_FATAL @ 10010.280001 us: (chip_sw_uart_smoke_vseq.sv:24) [uvm_test_top.env.virtual_sequencer.chip_sw_uart_smoke_vseq] wait timeout occurred!
[...]

PR #22734, opened 2024-04-19
[bazel] Revert "Remove opentitan_functest UART smoketest"

jwnrt:
This reverts commit 2cfed61d65eb9a7e0055af6fcfcff5697d28361c.

The Bazel target was removed because it was assumed to only be there for CW340 use, but it was also used by some DVSIM tests.

Re-adding until we work out how DVSIM can use the `opentitan_test` version.

See https://github.com/lowRISC/opentitan/issues/22733

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -17,6 +17,12 @@ load(
     "RSA_ONLY_KEY_STRUCTS",
 )
 load(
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
Thanks, @jwnrt!

jwnrt:
CI failure is an unrelated i2c test.

Issue #22697, opened 2024-04-18
ROM e2e Test cases running in DV with ECDSA implementation


moidx:
Implement ROM e2e Test cases with ECDSA keys.

PR #22739, opened 2024-04-20
[dv] Add sim_dv_rom_with_fake_keys exec env

a-will:
Because execution environments only describe how to sign the last stage and provide direct opentitan_binary() targets with signing and key info embedded for all prior stages, require specifying the target for the ROM directly (and for test_in_rom images).

Add a uart_smoketest target that is signed for DV.

For some reason, the OTP image provided by DV tooling is different from the OTP images associated with the execution environment for the sim_dv_rom_with_fake_keys target, causing
`kErrorSigverifyBadAuthPartition` errors. Provide a "custom" OTP from //hw/ip/otp_ctrl/data:img_rma instead, so the key digests and validity match.

Now we get `kErrorBootPolicyBadIdentifier`, even though the ROM and images in the scratch directory appear to have the expected values. However, when ROM reads the image, it comes back erased (0xffffffff). What now...?

### DIFF
--- a/hw/dv/tools/dvsim/sim.mk
+++ b/hw/dv/tools/dvsim/sim.mk
@@ -83,7 +83,7 @@ ifneq (${sw_images},)
 		index=`echo $$sw_image | cut -d: -f 3`; \
 		flags=(`echo $$sw_image | cut -d: -f 4- --output-delimiter " "`); \
 		bazel_label="`echo $$sw_image | cut -d: -f 1-2`"; \
-		if [[ $${index} != 4 && $${index} != 5 ]]; then \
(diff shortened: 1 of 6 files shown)


a-will:
I'm mostly making this available so others might take it further (or possibly direct me to the flow that needs modification). It seems like DV's OTP generation is doing its own thing somehow, but even after accounting for that, the reading of the manifest doesn't appear to match the actual image on the file system.

At some point, it'd be nice to connect the new opentitan_test() rules with DV targets that use signed binaries.

moidx:
This is probably already covered by this issue: https://github.com/lowRISC/opentitan/issues/22697

a-will:
Superseded.

PR #22744, opened 2024-04-20
[rom] Switch to ECDSA signature verification

moidx:
This commit switches the ROM to ECDSA signature verification. We were
unable execute a gradual transition away from RSA due to code size
limitations in the ROM.

Summary of changes:

1. Integrate otbn_boot_services in the ROM, remove RSA signature
   verification and associated keys and integrate ECDSA sigverify
   module.
2. Update manifest to major version 2.
3. Migrate all ROM based execution environments in hw/top_earlgrey to
   use ECDSA keys.

This change is part of https://github.com/lowRISC/opentitan/issues/21204.


sw/device/silicon_creator/lib/manifest.h
```diff
@@ -463,7 +463,7 @@ typedef struct manifest_ext_spx_signature {
 OT_WARN_UNUSED_RESULT
 inline rom_error_t manifest_check(const manifest_t *manifest) {
   // Major version must be `kManifestVersionMajor1`.
-  if (manifest->manifest_version.major != kManifestVersionMajor1) {
+  if (manifest->manifest_version.major != kManifestVersionMajor2) {
```
timothytrippel, line 466: nit: comment above

sw/device/silicon_creator/rom/BUILD
```diff
@@ -206,7 +205,6 @@ opentitan_binary(
     linker_script = ":linker_script",
     deps = [
         ":rom_without_keys",
-        "//sw/device/silicon_creator/rom/keys/real/rsa",
```
timothytrippel, line 209: should we probably remove this target entirely? since there will only be one ROM now that keys are in OTP?

sw/device/silicon_creator/rom/BUILD
```diff
@@ -175,7 +175,6 @@ opentitan_binary(
     linker_script = ":linker_script",
     deps = [
         ":rom_without_keys",
-        "//sw/device/silicon_creator/rom/keys/fake/rsa",
```
timothytrippel, line 178: do we need the `rom_with_fake_keys` target now that keys are in OTP? maybe we can remove it in a follow up and just rename `rom_without_keys` to `rom`?

sw/device/silicon_creator/rom/BUILD
timothytrippel: ok to do in a follow up if that was already your plan (if so pls disregard my comment :) )

timothytrippel (approved):
🥳

sw/device/silicon_creator/lib/manifest.h
moidx: Done.

sw/device/silicon_creator/rom/BUILD
moidx: Will do it in a follow up PR :)

sw/device/tests/BUILD
```diff
@@ -4398,8 +4398,10 @@ test_suite(
         cw310 = cw310_params(
             needs_jtag = True,
             otp = _SIVAL_OTP_IMAGE[lc_state],
+            # TODO(#22823): Remove "broken" tag once the test is fixed.
             tags = [
                 "lc_{}".format(lc_state),
+                "broken",
```
nbdd0121, line 4404: I think the `broken` tag should not be used when the test is not broken but the bitstream is.

Perhaps the `skip_in_ci` tag is better to skip these tests in CI.

timothytrippel: Hm, I think I would be prefer broken here, just because when we grep for "skip_in_ci" these are tests that should still work locally and are often (or should ideally be) run in a regular cadence regression, but are not run in pre-submit intentionally, usually due to resource constraints. To me, "Broken" indicates it is actually broken locally too. In this case, given our bitstream caching mechanism, this test is indeed broken locally too. Whether it is a software or hardware issue that is the cause of the brokenness is not as important IMO.

PR #22751, opened 2024-04-22
[usbdev,dv] usbdev disable endp seq

Saad-Anees1:
This PR includes usbdev_disable_endpoint_vseq, in which we send OUT/IN transactions when endpoints are disabled. Also, check the status of the transaction after sending it.


Saad22386:
@alees24 kindly review it.
Thanks

alees24:

PR #22752, opened 2024-04-22
[edn/dv] Correct prediction of sw_cmd_reg_rdy for auto mode Instantiate

vogelpi:
The sw_cmd_reg_rdy also goes high after the last additional data word and low again after receiving the ACK from CSRNG. Previously, the scoreboard expected this bit not to go high again after the last additional data word causing around 20 - 40% of edn_disable_auto_req_mode tests to fail.

### DIFF
--- a/hw/ip/edn/dv/env/edn_scoreboard.sv
+++ b/hw/ip/edn/dv/env/edn_scoreboard.sv
@@ -664,7 +664,8 @@ class edn_scoreboard extends cip_base_scoreboard #(
 
       // Each time a handshake happens in SW mode, the cmd_reg_rdy signal goes high.
       // In auto mode the signal also goes high but only for the initial instantiate command.
-      if ((!auto_mode || (!instantiated && clen_cntr)) && !boot_mode && predict_sts) begin
+      if ((!auto_mode || (!instantiated && clen_cntr) || (instantiated && inst_ack_outstanding))
+          && !boot_mode && predict_sts) begin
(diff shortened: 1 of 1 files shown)


rswarbrick:
This looks reasonable, and thanks for the careful commit message.

Trying to review this, I found the surrounding code quite hard to understand: I had to do quite a bit of reading around to understand things. I *think* the logic is something like this:
```systemverilog
      // Predict the sw_cmd_reg_rdy flag to be set as long as
      //
      //   - Predict_sts is true (so we're not starting an uninstantiate command)
      //   - We're not in boot mode.
      //   - We're finishing a handshake
      //   - One of these statements is true:
      //     +     We're in SW mode
      //     + OR: We're not yet instantiated and clen_cntr is positive (so there is still some more
      //           of the command to come)
      //     + OR: We are instantiated but we're waiting for an ACK to acknowledge this.
```
I wonder whether it would make sense to extend the comment a little to make it a little more obvious what's going on. I also wonder whether it makes sense to split `process_cs_cmd_fifo` so that we have a separate task for the first/later words in a command. It might be a way of giving names and splitting up a long line of code.

(Definitely not required as part of this PR, but it might be worth thinking about)

rswarbrick (approved):

vogelpi:
Thanks for your review @rswarbrick .

I agree that we could maybe improve the documentation or even code structure here. On the other hand, I also think it's a lot better already than what we had before (in all possible dimensions). :-)

PR #22762, opened 2024-04-22
Backport fix entropy src edn req

pamaury:
Manual backport of #22579 due to a minor conflict in the BUILD file.

### DIFF
--- a/sw/device/tests/entropy_src_edn_reqs_test.c
+++ b/sw/device/tests/entropy_src_edn_reqs_test.c
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+#include "sw/device/lib/arch/boot_stage.h"
 #include "sw/device/lib/base/memory.h"
 #include "sw/device/lib/base/mmio.h"
 #include "sw/device/lib/dif/dif_aes.h"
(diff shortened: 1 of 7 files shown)


sw/device/tests/entropy_src_edn_reqs_test.c
```diff
@@ -45,6 +45,17 @@ enum {
   kFpgaLoop = 5,
 };
 
+#ifndef SIVAL_ROM_EXT
```
moidx, line 48: Hi @pamaury, a couple of comments:

1. The following was added to perform a more dynamic test configuration based on the `keymgr` state:

https://github.com/lowRISC/opentitan/blob/dcaf19e3feee5ef4947c7ce8fd457057a3ae2b7c/sw/device/lib/testing/keymgr_testutils.c#L111-L131

2. The following was added to give test cases the ability to detect in which boot stage they are running:

https://github.com/lowRISC/opentitan/blob/dcaf19e3feee5ef4947c7ce8fd457057a3ae2b7c/sw/device/lib/arch/boot_stage.h#L28-L47

moidx: The approach followed in this PR can be implemented using the `boot_stage.h` approach.

moidx (approved):

sw/device/tests/entropy_src_edn_reqs_test.c
pamaury: Regarding the `keymgr`, I am not sure the approach you are suggesting applies: the test perform a keymgr transition. The issue is not the startup of the keymgr but that the expected stage is different between the ROM_EXT and Owner stages.

jwnrt (approved):
Did you mean to leave the `defines` changes in this PR even if they're unused? I haven't checked if the Bazel rules support was backported separately.

pamaury:
Since we now have the defines in the es_sival branch, we might as well keep it. What do you think?

jwnrt:
Sure, I guess that's fine. I could see it being useful and it doesn't really add any code.

sw/device/tests/entropy_src_edn_reqs_test.c
moidx: Hi @pamaury, 

I added some comments on this here: https://github.com/lowRISC/opentitan/pull/22708#discussion_r1578290691

Short answer is I believe so.

pamaury:
I believed I have addressed all comments

PR #22763, opened 2024-04-22
Spi device flash status tb fix

antmarzam:
This PR includes a few spi_device TB changes related to flash_status modelling:

* spi_monitor: seq_item events aren't triggered if in TPM mode since flash_status only updates during flash transactions
* spi_device_intercept_vseq: Leaving 'two_writes' case out since after chatting with Alex it's not correct SW operation to send two separate flash_status writes. Nonetheless,  leaving the code in there for when the time for coverage closure comes, since in theory the RTL supports a pair of writes in the following manner: 1st - set upper flash_status bits, 2nd - clear WEL/BUSY bits.
* spi_device_pass_base_seq: 'clear_flash_busy_bit' didn't take into consideration the RTL may clear busy/wel bits on TL-UL a-channel completion. However, it can happen the D-channel may be back-pressured and the 'csr_update' function comes back after TL-UL handshake completion. By the time 'csr_update' returns there could have been other flash commands which set the busy bit again.
* spi_device_env_cfg: new Flag set by SCB on flash command.
* spi_device_scoreboard: The scoreboard now takes hints of the flash_status values committed in the RTL, rather than trying to predict when a specific flash_status value would have crossed CDC boundaries


antmarzam:
I've not squash all the commits because I figured they'd make more sense like this, but happy to squash them if you think that's best!

a-will:
> spi_device_intercept_vseq: Leaving 'two_writes' case out since after chatting with Alex it's not correct SW operation to send two separate flash_status writes. Nonetheless, leaving the code in there for when the time for coverage closure comes, since in theory the RTL supports a pair of writes in the following manner: 1st - set upper flash_status bits, 2nd - clear WEL/BUSY bits.

Yeah, the RTL would support it, but to avoid confusion with the timing of when those writes will be absorbed and avoid the interface complexity of trying to manage the FIFO, we should tell SW users that they should perform a single write. If it's split into two, the host may end up needing to send one more SPI command to bring the second write across, depending on the timing. It's also desirable from a host visibility standpoint to update the registers atomically, so operation completion is synchronous with bit update.

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
```diff
@@ -613,18 +613,41 @@ class spi_device_pass_base_vseq extends spi_device_base_vseq;
...
+
     csr_update(ral.flash_status);
     // The intent here is to check the flash_status after csr_wr and then read the register
     // after end of SPI transaction (since flash_status gets update after CSB is deasserted)
     // First, ensure enough SPI clock cycles occur to make the transition into
     // the SPI domain.
+
+
+    //Note: it can happen the written flash_status is already synced with both SCK and SYS domains.
+    // This could happen in cases where the written_value in the A-channel occurss before CSB is
```
rswarbrick, line 633: Nit: Stray s in "occurs"

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
```diff
@@ -613,18 +613,41 @@ class spi_device_pass_base_vseq extends spi_device_base_vseq;
...
+    // active, but the D-channel response being back-pressured until the next inactive CSB
+
     while (wait_sck_count > 0) begin
       @(posedge cfg.spi_host_agent_cfg.vif.sck);
       if (cfg.spi_host_agent_cfg.vif.csb[FW_FLASH_CSB_ID] == 0) begin
+        if (wait_sck_count == num_sck_pulses) begin
+          // On the first SCK clock, we check if the value has already made it
+          csr_rd(.ptr(ral.flash_status), .value(new_flash_status), .backdoor(1));
```
rswarbrick, line 641: Isn't this the value we just wrote? (Or are we explicitly coping with the situation where it changed?)

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
```diff
@@ -613,18 +613,41 @@ class spi_device_pass_base_vseq extends spi_device_base_vseq;
...
+          // Note: it can happen the written flash_status is already synced with both SCK and SYS
+          // domains. This could happen in cases where the written_value in the A-channel occurss
+          // before CSB is active, but the D-channel response being back-pressured until the next
+          //  inactive CSB
+
+          if(new_flash_status != old_flash_status)
```
rswarbrick, line 647: Nit: Missing space after "if"

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
```diff
@@ -613,18 +613,41 @@ class spi_device_pass_base_vseq extends spi_device_base_vseq;
...
+    // active, but the D-channel response being back-pressured until the next inactive CSB
+
     while (wait_sck_count > 0) begin
       @(posedge cfg.spi_host_agent_cfg.vif.sck);
       if (cfg.spi_host_agent_cfg.vif.csb[FW_FLASH_CSB_ID] == 0) begin
+        if (wait_sck_count == num_sck_pulses) begin
+          // On the first SCK clock, we check if the value has already made it
+          csr_rd(.ptr(ral.flash_status), .value(new_flash_status), .backdoor(1));
+          // Note: it can happen the written flash_status is already synced with both SCK and SYS
+          // domains. This could happen in cases where the written_value in the A-channel occurss
```
rswarbrick, line 643: Nit: Stray s in "occurs"

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
```diff
@@ -613,18 +613,41 @@ class spi_device_pass_base_vseq extends spi_device_base_vseq;
...
+          // before CSB is active, but the D-channel response being back-pressured until the next
+          //  inactive CSB
```
rswarbrick, line 645: Would it make sense to move this comment to the bit where we're deciding whether to do the DV_CHECK_EQ?

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
```diff
@@ -640,7 +663,8 @@ class spi_device_pass_base_vseq extends spi_device_base_vseq;
     // If busy bit is not cleared, check once more else raise an error
     if (busy) begin
       get_flash_status_busy(busy);
-      `DV_CHECK_EQ(busy, 0, "flash_status.busy == 1 expected to be 0")
+      if (!busy_unset_early && !cfg.last_flash_cmd_set_busy)
```
rswarbrick, line 666: I think some commits are out of order: isn't the `last_flash_cmd_set_busy` variable only declared in the next commit?

hw/dv/sv/spi_agent/spi_monitor.sv
```diff
@@ -86,7 +86,7 @@ class spi_monitor extends dv_base_monitor#(
     // write to host_analysis_port
     case (cfg.spi_func_mode)
       SpiModeFlash: begin
-        -> host_item.item_finished_ev; // 
+        -> host_item.item_finished_ev;
```
rswarbrick, line 89: Could you move this fix to the first commit?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -62,7 +62,8 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
   flash_status_t tl_ul_side_flash_status_q[$];
   // Needed for when TL-UL read occurs before CDC transition from SPI->UL after CSB is inactive.
   // It'll have up to 2 items max. That's because once CSB goes high, if SW reads flash_status we
-  // may get the "old old" flash_status, or the new "old" value depending on when the read arrives
+  // may get the "old old" flash_status (no longer the active one, but had already been pushes to
```
rswarbrick, line 65: Nit: I think this should be "pushed"

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -128,6 +130,14 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
     forever begin
       bit is_status, is_jedec, is_sfdp, is_mbx_read;
       bit update_wel, wel_val;
+      //Note: depending on the timing of flash_status writes, some WREN/WRDI commands may be
```
rswarbrick, line 133: Nit: Missing space after "//"

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -128,6 +130,14 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+      // 'supressed' by the flash_status write. For example, if a write was clearing WEL, whilst
```
rswarbrick, line 134: Nit: Spelling of "suppressed"

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -128,6 +130,14 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+      // there is a flash WREN command the TB doesn't know when will the write value cros CDC. So,
```
rswarbrick, line 135: Nit: Missing s in cross. But maybe it should be something like "... value with go through CDC" to avoid saying cross twice.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -128,6 +130,14 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+      // TB just reads RTL's committed value to know if the write crossed CDCs before/after flash
+      // command
+      uvm_hdl_data_t 	rtl_sck_committed_status;
```
rswarbrick, line 138: Nit: I'm not sure we need the extra spaces here but (if we do add them), we should probably line up with the following line.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -191,13 +199,91 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                    // on that written value, and in others it won't.
+                    // SW note: This is not an actual concern since SW won't operate this way. In
+                    // reality, SW won't be changing busy/wel bits unless the host is known to be
+                    // waiting for busy/wel to drop (via sending flash READ_STATUS commands) os the
+                    // host is held in reset.
+                    `uvm_info(`gfn, "Updating WEL (0x1)", UVM_DEBUG)
                     update_wel = 1;
                     wel_val = 1;
                     spi_side_flash_status.wel = 1;
+                    rtl_committed_flash_status  = fetch_flash_status_rtl_committed_value();
```
rswarbrick, line 214: Nit: Stray space before "="

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -191,13 +199,91 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+
+                    // Update flash_status.wel predicted value
+                    if (!ral.flash_status.wel.predict(.value(spi_side_flash_status.wel),
+                                                     .kind(UVM_PREDICT_READ)))
```
rswarbrick, line 218: Nit: Indentation

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -191,13 +199,91 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                      `uvm_error(`gfn, "ral.flash_status.predict did not succeed")
```
rswarbrick, line 219: I think I'd suggest using `DV_CHECK` here.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -191,13 +199,91 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                    else begin
+                      `uvm_info(`gfn,$sformatf("Updated the TL-UL flash_status.wel CSR to: %p",
+                                               spi_side_flash_status.wel), UVM_DEBUG)
+                    end
+
+                    foreach (tl_ul_fuzzy_flash_status_q[i]) begin
+                      if (rtl_committed_flash_status == tl_ul_fuzzy_flash_status_q[i]) begin
+                        `uvm_info(`gfn, {$sformatf("The last flash_status written value (0x%0x)",
+                                                  tl_ul_fuzzy_flash_status_q[i]),
```
rswarbrick, line 228: Nit: Indentation

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -191,13 +199,91 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                                         "ocurred before this WREN command - but hasn't crossed",
+                                         "CDC boundaries until now and hence WEL won't be set"
+                                         }
+                                  ,
```
rswarbrick, line 232: Nit: I'd probably drop the newlines and move these above, so the line would end `... set"},`

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -191,13 +199,91 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                                  UVM_DEBUG)
+                        break;
+                      end // if (rtl_committed_flash_status == tl_ul_fuzzy_flash_status_q[i])
+                      if ( (rtl_committed_flash_status.other_status ==
```
rswarbrick, line 236: Nit: Stray space between the `( (` parens.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -191,13 +199,91 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                      if (rtl_committed_flash_status == tl_ul_fuzzy_flash_status_q[i]) begin
+                        `uvm_info(`gfn, {$sformatf("The last flash_status written value (0x%0x)",
+                                                  tl_ul_fuzzy_flash_status_q[i]),
+                                         "ocurred before this WREN command - but hasn't crossed",
+                                         "CDC boundaries until now and hence WEL won't be set"
+                                         }
+                                  ,
+                                  UVM_DEBUG)
+                        break;
+                      end // if (rtl_committed_flash_status == tl_ul_fuzzy_flash_status_q[i])
```
rswarbrick, line 235: Is this `foreach` loop really needed? I think it's just for improving a UVM_DEBUG message?

If it *is* useful, I notice that we've got essentially the same loop just below. Could we factor out a function that finds the first index `i` where `tl_ul_fuzzy_flash_status_q[i]` equals a given value? I think that would allow us to avoid repeating the foreach.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -214,8 +300,32 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
 
             // if busy, passthrough is blocked
             if (is_opcode_passthrough(item.opcode) && !`gmv(ral.flash_status.busy)) begin
+              `uvm_info(`gfn, "This is a Passthrough command and busy is not set", UVM_DEBUG)
               compare_passthru_item(.upstream_item(item), .is_intercepted(is_intercepted));
             end
+            else if(is_opcode_passthrough(item.opcode)) begin
```
rswarbrick, line 306: Can you restructure this to something like this?
```systemverilog
if (is_opcode_passthrough(...)) begin
  if (!`gmv(ral.flash_status.busy) begin
    ...
  end else begin
    ...
  end
end
```

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -214,8 +300,32 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+              if (!ral.flash_status.busy.predict(.value(spi_side_flash_status.busy),
+                                                 .kind(UVM_PREDICT_READ)))
+                `uvm_error(`gfn, "ral.flash_status.predict did not succeed")
+              else begin
+                `uvm_info(`gfn,$sformatf("Updated the TL-UL flash_status.busy CSR to: %p",
+                                         spi_side_flash_status.busy), UVM_DEBUG)
+              end
+            end
+
+            if(set_busy) begin
```
rswarbrick, line 324: Nit: Maybe collapse `if` with previous one?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -214,8 +300,32 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                                        spi_passthrough_downstream_q.size)}, UVM_DEBUG)
+            end
+
+            //Updating BUSY below, otherwise it compromises the previous comparison
+            if (set_busy) begin
+              spi_side_flash_status.busy = 1;
+              if (!ral.flash_status.busy.predict(.value(spi_side_flash_status.busy),
+                                                 .kind(UVM_PREDICT_READ)))
+                `uvm_error(`gfn, "ral.flash_status.predict did not succeed")
+              else begin
```
rswarbrick, line 318: Again, I'd probably suggest `DV_CHECK_FATAL` here too.

rswarbrick (commented):
I've only gone through half of the last commit so far. Would it make sense to try to split it out a bit? Maybe it makes sense to add the debug prints separately from the commit that alters the scoreboard behaviour?

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
antmarzam: It can be, but not neccessarily- It depends on what's happened from the moment csr_update returned. 
For example, You could have the TL-UL flash_status write A-channel being handshaked, but the D-channel being back-pressured (for maybe like 1 flash_command or more). 
The TL-UL flash_status CSR only updates on CSB assertion: The written value propagates towards the SPI-side at each SPI byte beat, and the RTL commits the written value towards the TL-UL side on CSB becoming inactive

In the case I describe above, We just set the `busy_unset_early` flag if the value has propagated already towards the TL-UL side because of CSB assertion.
In a "golden" case without backpressure on the D-channel, by the time we reach the first SPI clock cycle the TL-UL side will still hold the "old_written_value", until CSB goes inactive.

hw/ip/spi_device/dv/env/seq_lib/spi_device_intercept_vseq.sv
```diff
@@ -57,23 +57,28 @@ class spi_device_intercept_vseq extends spi_device_pass_cmd_filtering_vseq;
...
 
     access_option = wrong;
     randcase
       25: access_option = sequential_access;
       70: access_option = concurrent_access;
-      5 : access_option = two_writes;
+      // Advice is for stimulus not to operate with 'two_writes' access option. However, the option
+      //  is left here in case we'd need to exercise any RTL feature which becomes
+      // difficult/impossible to exercise with the 'two_writes' configuration
+      //      5 : access_option = two_writes;
```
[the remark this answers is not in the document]
antmarzam: No it doesn't but see comment above from Alex where he says SW should be constrained to a 1-write only: https://github.com/lowRISC/opentitan/pull/22763#issuecomment-2070927701

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
antmarzam: I've left another command at the DV_CHECK_EQ, but since the written flash_status may be committed by the time we reach the back-door csr_rd I thought it made sense to leave it there rather than at a later point in time.

hw/ip/spi_device/dv/env/seq_lib/spi_device_pass_base_vseq.sv
antmarzam: Ooops - You're right. I'm going to squash all the commits

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: I personally find it more readable by using UVM reporting macros directly. Is this because of coding guidelines? 
Also, DV_CHECK doesn't allow for an else if the check goes correctly, right?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: It's not just for the debug message, the UVM_INFO helps to know in which cases a WREN command doesn't set the WEL bit.
But also, there are cases in which the WEL bit is set in the second if of the foreach.  I didn't consider making a function for this case because the conditions and printouts are different for the WREN/WRDI commands.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: I replaced the uvm_errors with DV_CHECK_EQ_FATAL in the last commit

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                // In which case, we need to see if the tl_ul_fuzzy_q has any value different from
+                // the GMV (which means there's been a write)
+
+
+                // Note: depending on the clocks relationship, it may happen the RTL may commit a
+                // value before the TB notices. The TB isn't trying to accurately predict when a
+                // given value will be committed. Instead, the TB just checks from the set of
+                // written values one of those has been comitted
+
+                comitted_value_predicted  = 0;
```
rswarbrick, line 955: Nit: Maybe this line can be dropped, since it's overwritten by the next statement anyway? Perhaps combine the two blocks of comments with a commented empty line?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                // It can happen busy/wel bits may get set as the SPI txn goes through. In this
+                // case we don't compare WEL/busy, as the value might've changed since the
+                // flash_status write occurred
+                comitted_value_predicted = flash_sts_committed_value_in_q(tl_ul_side_flash_status_q,
+                                                                          0, 0, tl_ul_side_idx);
+                comitted_value_predicted |= flash_sts_committed_value_in_q(
+                                                                         tl_ul_fuzzy_flash_status_q,
+                                                                         0, 0,tl_ul_fuzzy_idx);
```
rswarbrick, line 963: To avoid the updates, I wondered about something like
```systemverilog
                comitted_value_predicted = |{flash_sts_committed_value_in_q(tl_ul_side_flash_status_q,
                                                                            0, 0, tl_ul_side_idx),
                                             flash_sts_committed_value_in_q(tl_ul_fuzzy_flash_status_q,
                                                                            0, 0,tl_ul_fuzzy_idx)};
```
But that doesn't work because the lines get too long. I wonder if a shorter name would work for the function?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                if (comitted_value_predicted) begin
+                  uvm_hdl_data_t 	rtl_sck_status_committed_value;
```
rswarbrick, line 965: Nit: stray spaces

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                  flash_status_t  f_status_rtl_committed_value;
+                  // Queue rather than a plain variable to check emptiness before updating
+                  // spi_side_flash_status
+                  flash_status_t  flash_status_q[$];
+                  flash_status_q = {};
```
rswarbrick, line 970: Doesn't a queue get initialised empty anyway?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                    flash_status_q.push_back(tl_ul_side_flash_status_q[tl_ul_side_idx]);
+                    `uvm_info(`gfn, {"The RTL committed value is found in ",
+                                     $sformatf("tl_ul_side_flash_status_q[i=%0d]=0x%0x",
+                                               tl_ul_side_idx,
+                                               tl_ul_side_flash_status_q[tl_ul_side_idx])},
+                                     UVM_DEBUG)
+                    // removing all old items up to the one which made it to spi_side
+                    for (int i=0; i <= tl_ul_side_idx; i++) begin
+                      `uvm_info(`gfn, $sformatf("Popping 'tl_ul_side_flash_status_q[i=%0d]=0x%0x",i,
+                                                tl_ul_side_flash_status_q.pop_front()), UVM_DEBUG)
```
rswarbrick: I'd suggest avoiding side-effects in the print routines because they're a bit hard to notice. Maybe just call `front()` here and then `pop_front()` as a separate statement on the next line?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                    end
+                  end // if (tl_ul_side_idx != -1)
+                  else if (tl_ul_fuzzy_idx != -1) begin
+                    if(ongoing_read_status_cmd) begin
```
rswarbrick, line 989: Nit: Missing space after "if"

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                      `uvm_info(`gfn, {"The RTL committed value is found in ",
+                                       $sformatf("tl_ul_fuzzy_flash_status_q[i=%0d]=0x%0x",
+                                                 tl_ul_fuzzy_idx,
+                                                 tl_ul_fuzzy_flash_status_q[tl_ul_fuzzy_idx])},
+                                UVM_DEBUG)
+                    end
+                  end // if (tl_ul_fuzzy_idx != -1)
+
+                  // Predicted queues non-empty and value committed ? -> we update spi_side_flash_status
+                  if(flash_status_q.size > 0) begin
```
rswarbrick, line 1010: Nit: Missing space after "if". As a quick way to find these, I see that `git grep "if("` returns lots of hits in this directory.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -775,36 +886,212 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
...
+                    // Just for easier readability of the comparison
+                    flash_status_t flash_rtl_committed;
+                    bit tweak_wel_busy;
+                    // 8th SCK clk (byte beat) and we update SPI side
+                    `uvm_info(`gfn, {"Updating the SPI-side. Old spi-side flash_status value:",
+                                     $sformatf(" 0x%0x",spi_side_flash_status)}, UVM_DEBUG)
+
+                    //Updating SPI-side with committed value:
```
rswarbrick, line 1018: Nit: Missing space after "//". I think you can get the list with something like: `git grep '//[a-zA-Z]' -- 'hw/ip/spi_device/dv/**/*.sv'`

rswarbrick (commented):
I've gone through the scoreboard changes a bit more carefully now. Crikey, this was a big change (and looks like it was rather confusing to make!)

Lots of nitty comments, but I think the code mostly makes sense.

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: Renaming the function `flash_sts_committed_in_q`  and your vector concatenation suggestion did the trick, thank you :)

rswarbrick (approved):
Thanks for all the tidy-ups!

PR #22765, opened 2024-04-22
cert: make sure fake CA matches TPM certificates' Issuer

vbendeb:
For openssl x509 certificate chain verification to succeed subject/issuer key ID and subject/issuer Distinguished Name fields must match. Moreover, the Distinguished name components must match literally, i.e. not just the component values, but also the components layout in the certificate.

The country name field must be exactly two characters (US for the USA).

The hjson cert parser seems to be changing the order of the DN components listed in the input files, changing the openssl configuration file allows to create a CSR which results in the expected layout of the generated certificate's DN field.

After these changes the generated certificates can be successfully verified, i.e.
 $ openssl verify -CAfile fake_ca.pem tpm_ek.full.pem
 tpm_ek.full.pem: OK


timothytrippel (approved):
Looks like values are always sorted by OID to get a deterministic order: https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/ot_certs/src/asn1/x509.rs;drc=166179b6e280ea8aa8bdd119dda71af849e86a2e;l=198

timothytrippel:
Nice find @vbendeb !

pamaury:
Unfortunately, the order in which fields are listed in the hjson cannot be retrieved easily because it is a map so the order is not guaranteed, which is why I sort them by OID so we get a deterministic order. I was not aware that even the layout would matter which is a bit unfortunate. If we want to preserve the order, we could either switch to a list (this would make the syntax ugly though), or maybe a custom serde deserializer could preserve the order.
What do you think?

vbendeb:
> Unfortunately, the order in which fields are listed in the hjson cannot be retrieved easily because it is a map so the order is not guaranteed, which is why I sort them by OID so we get a deterministic order.

So it is deterministic as in the same hjson will produce the same cert order every time and the order can be predicted, which might be good enough.

All we need to do in this case is to ensure that the intermediate CA has the fields of DN listed in the same order. This is easy to achieve since when gnerating the CSR openssl strictly follows the order of the fileds in the cofiguration file.

> I was not aware that even the layout would matter which is a bit unfortunate.

This was kind of a surprise to me too, but the last thing I changed was the order of the fields in the CA cert and the verification started succeeding.

> If we want to preserve the order, we could either switch to a list (this would make the syntax ugly though), or maybe a custom serde deserializer could preserve the order. What do you think?

I think to be inline with openssl the deserailizer should mainatin the order such that the user does not have to think twice when encoding the templates. But as long as we can enforce the Intermediate CA cert format we can just leave the rest as is.

timothytrippel:
Yeah I think perhaps we can just drop a comment in the config file explaining the ordering and leave as is for now.

vbendeb:
can we push this patch in the meantime so that the end-to-end openssl tests can be run?

pamaury (approved):
I am going to approve this but as for end to end tests, we prefer to have them in the repository. We don't like the opensource repo chasing some unknown e2e tests.

vbendeb:
I thought of a test where certificates are generated, UDS ad TPM are certs endorsed with the fake_ca.pem key, and then `openssl verify..` is invoked to validate the chains  - is using openssl for such a test acceptable?

vbendeb:
can this be merged? The erros do not seem to be related to this patch

PR #22766, opened 2024-04-22
[manuf] provision TPM attestation key seeds in flash info

### LINKED ISSUE #22623: We will need to update the provisioning firmware to provision these seeds into a flash info page during manufacturing. We can put them on the same page as the attestation key seeds if we would like. The current provisioning firmware only provisions the attestation key seeds [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/lib/personalize.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=373). These seeds are loaded from flash info into OTBN DMEM during key generation by `otbn_boot_attestation_keygen`  [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/otbn_boot_services.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=141).
We will need to update the provisioning firmware to provision these seeds into a flash info page during manufacturing. We can put them on the same page as the attestation key seeds if we would like. The current provisioning firmware only provisions the attestation key seeds [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/lib/personalize.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=373). These seeds are loaded from flash info into OTBN DMEM during key generation by `otbn_boot_attestation_keygen`  [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/otbn_boot_services.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=141).

I can turn this into a separate issue, and we can address in a follow up PR.

_Originally posted by @timothytrippel in https://github.com/lowRISC/opentitan/pull/22573#discussion_r1569129546_

timothytrippel:
This fixes #22623.

### DIFF
--- a/sw/device/silicon_creator/manuf/lib/flash_info_fields.c
+++ b/sw/device/silicon_creator/manuf/lib/flash_info_fields.c
@@ -83,6 +83,27 @@ const flash_info_field_t kFlashInfoFieldCdi1AttestationKeySeed = {
     .byte_offset = (2 * kAttestationSeedBytes),
 };
 
+const flash_info_field_t kFlashInfoFieldTpmEkAttestationKeySeed = {
+    .partition = 0,
+    .bank = 0,
+    .page = 4,
+    .byte_offset = (3 * kAttestationSeedBytes),
+};
+
+const flash_info_field_t kFlashInfoFieldTpmCekAttestationKeySeed = {
+    .partition = 0,
+    .bank = 0,
+    .page = 4,
+    .byte_offset = (4 * kAttestationSeedBytes),
+};
+
+const flash_info_field_t kFlashInfoFieldTpmCikAttestationKeySeed = {
+    .partition = 0,
+    .bank = 0,
+    .page = 4,
+    .byte_offset = (5 * kAttestationSeedBytes),
+};
+
 status_t manuf_flash_info_field_read(dif_flash_ctrl_state_t *flash_state,
                                      flash_info_field_t field,
                                      uint32_t *data_out, size_t num_words) {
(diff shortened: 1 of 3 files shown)


sw/device/silicon_creator/manuf/lib/flash_info_fields.c
```diff
@@ -83,6 +83,27 @@ const flash_info_field_t kFlashInfoFieldCdi1AttestationKeySeed = {
...
+    .bank = 0,
+    .page = 4,
+    .byte_offset = (4 * kAttestationSeedBytes),
+};
+
+const flash_info_field_t kFlashInfoFieldTpmCikAttestationKeySeed = {
+    .partition = 0,
+    .bank = 0,
+    .page = 4,
+    .byte_offset = (5 * kAttestationSeedBytes),
```
vbendeb, line 104: I think this scheme would be more robust to define a structure describing the info page layout, verifying that the structure fits into the page, etc.

Could be a future improvement, this patch naturally follows the existing approach.

vbendeb (approved):

sw/device/silicon_creator/manuf/lib/flash_info_fields.c
timothytrippel: Good idea, I captured in a separate issue: #22768.

moidx (approved):

PR #22770, opened 2024-04-22
[rom_ext] Relax ePMP settings.

cfrantz:
Per discussion in the tape-out sync meeting on 2024-04-22, relax the ePMP settings:

Unlock the configuration of ePMP entries 12, 14 and 15, allowing the owner to reconfigure as needed.

As detailed in the [ePMP Proposal](file:///home/cfrantz/Downloads/Smepmp.pdf) (see section 2.2), this change will allow for CPU execution over the entire flash region, the MMIO region and main RAM because we're currently operating with MMWP=1 and MML=0.  We prevent RAM execution at the SRAM controller.

### DIFF
--- a/sw/device/silicon_creator/lib/epmp_state.h
+++ b/sw/device/silicon_creator/lib/epmp_state.h
@@ -70,8 +70,14 @@ typedef enum epmp_perm {
   /** M mode: read/write/execute. U mode: read only. */
   kEpmpPermReadOnly = EPMP_CFG_R,
 
+  /** M mode: read/write/execute. U mode: read/write. */
+  kEpmpPermReadWrite = EPMP_CFG_R | EPMP_CFG_W,
+
   /** M mode: read/write/execute. U mode: read/execute. */
   kEpmpPermReadExecute = EPMP_CFG_R | EPMP_CFG_X,
+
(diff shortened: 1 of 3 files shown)


mundaym:
If you want to be more conservative wrt executable regions then we could leave the locked flash region as-is and just make the MMIO and main RAM regions unlocked since IIRC execution in those regions is not possible anyway (at least in PROD).

cfrantz:
> If you want to be more conservative wrt executable regions then we could leave the locked flash region as-is (though maybe move it to slot 15) and just make the MMIO and main RAM regions unlocked since IIRC execution in those regions is not possible anyway (at least in PROD).

Why would we want to relax the settings on the last entries?  From a security perspective, I don't really have a problem with it (e.g. we'd enable execution anywhere in flash, but the owner can always set the lock bit or mess with MML), but I don't understand what additional flexibility we're giving to an owner.  I'm having trouble imagining a scenario where you'd want to change the flash, mmio or ram mappings.

cfrantz:
I've adjusted the configuration as discussed this morning: entries 12-15 are now unlocked.

moidx (approved):

cfrantz:

PR #22772, opened 2024-04-23
[ot_certs] Rework Names and rename extensions to private_extensions

### LINKED ISSUE #22620: @pamaury : just a thought for better clarity / consistency with the [X.509 spec](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2), perhaps in a follow up PR we should name the `extensions` field here `private_extensions` since the spec states "_The X.509 v3 certificate format also allows communities to define private extensions to carry information unique to those communities._"  Thoughts?
@pamaury : just a thought for better clarity / consistency with the [X.509 spec](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2), perhaps in a follow up PR we should name the `extensions` field here `private_extensions` since the spec states "_The X.509 v3 certificate format also allows communities to define private extensions to carry information unique to those communities._"  Thoughts?

_Originally posted by @timothytrippel in https://github.com/lowRISC/opentitan/pull/22573#discussion_r1569105674_

pamaury:
See commits for explanations. There is an issue when parsing names with openssl as openssl-sys is misses a binding. I will ask upstream to add it.

Fixes #22620
Hopefully fixes #22765 ?


sw/host/ot_certs/src/template/mod.rs
```diff
@@ -77,13 +77,19 @@ pub struct Certificate {
...
+    #[serde(default)]
+    pub private_extensions: Vec<CertificateExtension>,
     /// X509 certificate's signature.
     pub signature: Signature,
 }
 
+/// An X501's Name: this is an ordered list of sets of attributes.
+/// The order of the list matters but the order within the sets of
+/// attributes is arbitrary.
+pub type Name = Vec<IndexMap<AttributeType, Value<String>>>;
```
timothytrippel, line 91: Just curious, why is this a Vec of IndexMaps? Doesn't the IndexMap itself  give you ordering: https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#order ?

Also, if you do need the Vec, do the IndexMap's only ever have on key/value pair? If so would it be better to use a Vec of Structs or Vec of Tuples here?

pamaury: I have updated the comment, is that better?

timothytrippel: LGTM, thanks!

timothytrippel:
I don't think the CI are failures are relevant, shall we merge?

timothytrippel:
Fixes #22620 as well.

pamaury:
Yes the failure looks unrelated, you can merge if you are happy with the code

nbdd0121 (approved):

PR #22777, opened 2024-04-23
[rom] Always set ret sram version on boot

jwnrt:
This PR changes the ROM to write the retention SRAM version on every boot, not just after initialization. The version is expected by the ROM_EXT to boot, but is cleared by scrambling.

An alternative would be for the `ROM_EXT` to assume that unknown version indicate scrambling and re-populate the retention sram.

This change fixes the `sram_ctrl_sleep_sram_ret_contents_scramble_test` which ensures we can wake up from deep sleep after scrambling.

---

The previous version of this PR disabled this part of the test:

> Deep sleep requires a reset to wake up again. Resetting runs through the ROM_EXT. The ROM_EXT will not boot if the retention SRAM is scrambled.

> We therefore cannot test this part under those conditions, so I have disabled it.

### DIFF
--- a/sw/device/silicon_creator/rom/rom.c
+++ b/sw/device/silicon_creator/rom/rom.c
@@ -181,9 +181,13 @@ static rom_error_t rom_init(void) {
       otp_read32(OTP_CTRL_PARAM_CREATOR_SW_CFG_RET_RAM_RESET_MASK_OFFSET);
   if ((reset_reasons & reset_mask) != 0) {
     retention_sram_init();
-    retention_sram_get()->version = kRetentionSramVersion3;
     retention_sram_get()->creator.last_shutdown_reason = kErrorOk;
   }
+
+  // Always store the retention RAM version so the ROM_EXT can depend on its
+  // accuracy even after scrambling.
+  retention_sram_get()->version = kRetentionSramVersion3;
+
   // Store the reset reason in retention RAM and clear the register.
   retention_sram_get()->creator.reset_reasons = reset_reasons;
   rstmgr_reason_clear(reset_reasons);
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -3541,20 +3541,15 @@ opentitan_test(
 opentitan_test(
     name = "sram_ctrl_sleep_sram_ret_contents_scramble_test",
     srcs = ["sram_ctrl_sleep_sram_ret_contents_scramble_test.c"],
-    broken = cw310_params(tags = ["broken"]),
     exec_env = dicts.add(
         EARLGREY_TEST_ENVS,
         {
             "//hw/top_earlgrey:fpga_cw310_sival": None,
-            # FIXME broken in sival ROM_EXT, remove this line when fixed. See #21706.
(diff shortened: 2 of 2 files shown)


a-will:
>  The ROM_EXT will not boot if the retention SRAM is scrambled.

Did @cfrantz or others doing ROM_EXT happen to confirm that this isn't a bug in ROM_EXT? (just checking that we aren't disabling part of a test on master that _should_ pass)

a-will:
It takes a little effort to go from the hex to the module code and then the particular error, but in case you hadn't found the definitions yet, the BFV error constants are defined in this file:

https://github.com/lowRISC/opentitan/blob/41bc3e0c7f3f69618505ead494725c93a5bb8727/sw/device/silicon_creator/lib/error.h#L25

jwnrt:
> Did @cfrantz or others doing ROM_EXT happen to confirm that this isn't a bug in ROM_EXT? (just checking that we aren't disabling part of a test on master that _should_ pass)

This is the error that triggers on the boot after deep sleep. It seems intentional: we can't read the version from the RET SRAM since it's scrambled.

https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/lib/drivers/retention_sram.c#L59

jwnrt:
> It takes a little effort to go from the hex to the module code and then the particular error, but in case you hadn't found the definitions yet, the BFV error constants are defined in this file:
> 
> https://github.com/lowRISC/opentitan/blob/41bc3e0c7f3f69618505ead494725c93a5bb8727/sw/device/silicon_creator/lib/error.h#L25

Thanks, we did find this eventually but only after debugging on a CW340 and getting to the actual point where the error fires. BFVs aren't documented very well I don't think.

a-will (commented):
Oh, I see what's going on here. This test can't clear the ROM_EXT stage after deep sleep if the scrambling key is renewed in a later boot stage, since that obliterates silicon owner values.

I think I'd suggest there may be a bug in the ROM, though--ROM is not initializing a value expected by ROM_EXT unless ROM is also wiping the entire contents of the retention sram:

https://github.com/lowRISC/opentitan/blob/41bc3e0c7f3f69618505ead494725c93a5bb8727/sw/device/silicon_creator/rom/rom.c#L175-L185

jwnrt:
Yeah, I think you're right. I've changed this PR to make the ROM always write the retention SRAM version regardless of whether it's a boot that initializes it.

cfrantz:
> > Did @cfrantz or others doing ROM_EXT happen to confirm that this isn't a bug in ROM_EXT? (just checking that we aren't disabling part of a test on master that _should_ pass)
> 
> This is the error that triggers on the boot after deep sleep: https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/lib/drivers/retention_sram.c#L59
> 
> It seems intentional: we can't read the version from the RET SRAM since it's scrambled. We didn't check with any ROM_EXT developers, but it would be good to see what @cfrantz thinks about this failure mode.

I agree that this is an error and we should always initialize the version number in the ROM.

cfrantz (approved):

timothytrippel (approved):

PR #22781, opened 2024-04-23
doc/theory_of_operation.md: "under" should be "over"

cousteaulecommandant:
"If the baud rate is (too low), care should be taken when the system clock frequency is (too high)." - It currently says "when the clock frequency is **under** X MHz"; should say "over".

I also changed the rounding of frequency values in that paragraph (after doing the math to make sure it's right) so that they're rounded down instead of up, which makes more sense after this change (otherwise, it would imply that 31.7 MHz is safe for 1200 baud since it's <32, but it's not - max safe frequency is 31.45).

### DIFF
--- a/hw/ip/uart/doc/theory_of_operation.md
+++ b/hw/ip/uart/doc/theory_of_operation.md
@@ -154,10 +154,10 @@ $$ {{baud} < {{40 * f\_{pclk}} \over {2^{$bits(NCO)+4}}}} \qquad OR \qquad
 {{f\_{pclk}} > {{{2^{$bits(NCO)+4}} * {baud}} \over {40}}} $$
 
 Using rounded frequencies and common baud rates, this implies that
-care is needed for 9600 baud and below if the system clock is under
-250MHz, with 4800 baud and below if the system clock is under 125MHz,
-2400 baud and below if the system clock us under 63MHz, and 1200 baud
-and below if the system clock is under 32MHz.
+care is needed for 9600 baud and below if the system clock is over
+250MHz, with 4800 baud and below if the system clock is over 125MHz,
+2400 baud and below if the system clock is over 62MHz, and 1200 baud
+and below if the system clock is over 31MHz.
 
 
 ### Interrupts


rswarbrick (dismissed):
I don't believe the proposed change is correct. The paragraph is pointing out that running a very fast UART baud rate with a slow system clock won't work, and giving regions where this will be a concern.

cousteaulecommandant:
> The paragraph is pointing out that running a very fast UART baud rate with a slow system clock won't work

Not really, it is pointing out that running a very **slow** UART may result in inaccurate baud rates because the UART frequency granularity is too coarse.  This happens with very slow baud rates at very high clock frequencies, and as the formula above that paragraph shows, the region of concern is when

$$ {{f\_{pclk}} > {{{2^{$bits(NCO)+4}} * {baud}} \over {40}}} $$

i.e., when $f\_{pclk}$ is **greater** than a certain value.

This is unrelated to the other end of the spectrum, where the baud rate cannot be too fast (higher than 1/16 of the system clock frequency).  This means that a baud rate of 9600 bauds and **above** cannot be achieved with a system clock frequency **under** 0.1536 MHz.  (In that case, the baud rate is not a "concern", it's simply impossible to achieve.)  Note that the paragraph talks about 250 MHz, not 0.15.

In no case is the combination "baud rate **below**" + "system clock **under**" a reason of concern, but that's what the paragraph states.

rswarbrick:
You are completely right, sorry! I've just read through things carefully (and scribbled on a piece of paper to convince myself). Thank you very much for the proposed change.

Annoyingly, it is failing the "quick lint" test because the commit author is "cousteau" and the script (rather stupidly) assumes that every full name will contain a space. We should probably fix that! Fortunately, I suspect you are a French speaker and have a reasonably European name. Do you have a full name that you could use as an author in the commit message?

Error message:
```
Commit f018c6d72e3e64de48aa162a054e63d3474054f2: The commit author name 'cousteau' contains no space. Use "git config user.name 'Johnny English'" to set your real name, and update the commit with "git rebase -i " and/or "git commit --amend --signoff --reset-author". This command will also sign off your commit indicating agreement to the Contributor License Agreement. See CONTRIBUTING.md for more details.
```

cousteaulecommandant:
I amended the commit.  Is it OK now?
(In hindsight, it was naïve of me to believe that I could get away with just using the github interface to make this change...)

rswarbrick (approved):
This looks good to me, and thank you for all the effort getting it merged :-)

PR #22789, opened 2024-04-24
ot_certs: add an app for signing TBS blobs

vbendeb:
The app accepts three command line parameters

 - file names of the signing key, a private P256 key in DER form
 - file name of the TBS blob of the certificate
 - file name to save the properly endorse certificate

This was tested in the end to end test where the TBS blobs were created by the generated code, signed with
sw/device/silicon_creator/manuf/keys/fake/cert_endorsement_key.sk.der and then the certificate chain was validated using sw/device/silicon_creator/manuf/keys/fake/fake_ca.pem as the CA.

### DIFF
--- a/sw/host/ot_certs/BUILD
+++ b/sw/host/ot_certs/BUILD
@@ -2,7 +2,7 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-load("@rules_rust//rust:defs.bzl", "rust_doc", "rust_library", "rust_test", "rust_test_suite")
+load("@rules_rust//rust:defs.bzl", "rust_doc", "rust_library", "rust_binary", "rust_test", "rust_test_suite")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -62,6 +62,23 @@ rust_library(
         # is merged.
         "@crate_index//:openssl-sys",
         "@crate_index//:foreign-types",
+    ],
+)
+
+rust_binary(
+    name = "tbs_sign",
+    testonly = True,
+    srcs = ["src/tbs_sign.rs"],
+    deps = [
+        "//sw/host/provisioning/ft_lib:ft_lib",
+        "//sw/host/opentitanlib",
+        "//sw/host/ot_certs",
+        "@crate_index//:anyhow",
+        "@crate_index//:clap",
+        "@crate_index//:elliptic-curve",
+        "@crate_index//:num-bigint-dig",
+        "@crate_index//:p256",
+        "@crate_index//:sha2",
     ],
 )
 
--- a/sw/host/ot_certs/src/tbs_sign.rs
+++ b/sw/host/ot_certs/src/tbs_sign.rs
@@ -1,0 +1,75 @@
+use anyhow::Result;
+use clap::Parser;
+use elliptic_curve::pkcs8::DecodePrivateKey;
+use elliptic_curve::SecretKey;
+use num_bigint_dig::BigUint;
+use p256::ecdsa::SigningKey;
+use p256::NistP256;
+use sha2::{Digest, Sha256};
+use std::fs::File;
+use std::io::{Read, Write};
+use std::path::PathBuf;
+
+use ot_certs::template::{EcdsaSignature, Signature, Value};
+use ot_certs::x509::generate_certificate_from_tbs;
+
+/// TBS signing command-line parameters.
+#[derive(Debug, Parser)]
+struct TbsSignInputs {
+    /// Certificate endorsement ECC (P256) private key DER file.
+    #[arg(long)]
+    signing_key: PathBuf,
+
+    /// Plain binary file containing the certificate TBS section.
+    #[arg(long)]
+    tbs: PathBuf,
+
+    /// Output file to store the signed certificate.
+    #[arg(long)]
+    signed_cert: PathBuf,
+}
+
+fn run_ft_tbs_signer(host_ecc_sk: &PathBuf, tbs: &PathBuf) -> Result<Vec<u8>> {
+    let mut file = File::open(tbs)?;
+
+    // Read the file contents into a byte vector
+    let mut tbs_bytes = Vec::new();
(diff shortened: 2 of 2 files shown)


sw/host/ot_certs/BUILD
```diff
@@ -65,6 +65,23 @@ rust_library(
     ],
 )
 
+rust_binary(
+    name = "tbs_sign",
+    testonly = True,
+    srcs = ["src/tbs_sign.rs"],
+    deps = [
+        "//sw/host/provisioning/ft_lib:ft_lib",
```
pamaury, line 73: I might have missed it but I don't think you actually use anything in `ft_lib`?

pamaury (approved):
The code looks good @vbendeb . I just have a couple of questions: do you think it would make sense to make this a subcommand of `opentitantool certificate` instead of a binary on its own? Or if you don't like that, we could consider doing it the other way around: move the `opentitantool certificate` command to its own binary in `ot_certs` that would include yours? (this is just to avoid having 10 different binaries).

The advantage of your current code is that it treats the TBS as a binary blob but I wonder if (by default, it could have an option to disable) it would make sense in the future for the tool to run some basic checks like making sure that the signing key matches thet key specific in the TBS?

timothytrippel (commented):
I agree with @pamaury , I don't think we should create a separate binary just to sign TBS certs. This should be implemented as subcommands of the `opentitantool certificate` subcommand: https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/opentitantool/src/command/certificate.rs;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=36

We try to group related functionality together as @pamaury says.

timothytrippel:
Also, I think the best way to end-to-end test cert chains would be to make it part of the [ `ft_provision` test flow](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/BUILD;drc=69a9afcd9afed484fb32f4c115833084fad603bb;l=259) as we discussed. It is just a few modifications to the existing perso flow and we will then be testing everything (cert gen/parsing/chain-validation etc) all in one go.

Specifically, I think we should add the generation of TPM TBS certs [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c;drc=b1694b9990cfa4dd9b3a73a66ca786041dd8eae3;l=279), send them to host by adding them to the UJSON struct [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/json/provisioning_data.h;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=104), then parse and endorse them [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/ft_lib/src/lib.rs;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=250), then add calls to openssl to verify the chain [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/ft_lib/src/lib.rs;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=260), then send them back to the device [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/ft_lib/src/lib.rs;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=267) (by adding them to the UJSON struct [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/json/provisioning_data.h;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=118)).

timothytrippel:
Having these commands in opentitantool is an added bonus though, since you have already written the code.

vbendeb:
> Also, I think the best way to end-to-end test cert chains would be to make it part of the [ `ft_provision` test flow](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/BUILD;drc=69a9afcd9afed484fb32f4c115833084fad603bb;l=259) as we discussed. It is just a few modifications to the existing perso flow and we will then be testing everything (cert gen/parsing/chain-validation etc) all in one go.
> 
> Specifically, I think we should add the generation of TPM TBS certs [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c;drc=b1694b9990cfa4dd9b3a73a66ca786041dd8eae3;l=279), send them to host by adding them to the UJSON struct [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/json/provisioning_data.h;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=104), then parse and endorse them [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/ft_lib/src/lib.rs;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=250), then add calls to openssl to verify the chain [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/ft_lib/src/lib.rs;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=260), then send them back to the device [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/ft_lib/src/lib.rs;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=267) (by adding them to the UJSON struct [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/json/provisioning_data.h;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=118)).

this would require some hardware hookup to run the test, which is ideal, but not readily available.  I  think a test flow  where the test can run solely on the build host, without actual hardware is also valuable.

vbendeb:
ok, as discussed elsewhere, let's concentrate on making the HW supported flow work first, I'll get back to this patch later.

PR #22790, opened 2024-04-24
fix rust format violations.

vbendeb:
This patch is a result of running

$ touch sw/device/silicon_owner/tock/kernel/src/tests.rs $ basel run //quality:rustfmt_fix
$ rm sw/device/silicon_owner/tock/kernel/src/tests.rs

The first line is necessary to address the issue where rustfmt fails to process sw/device/silicon_owner/tock/kernel/src/main.rs due to the missing test module.


vbendeb:
I'm not sure why these files are not formatted properly, suggesting an automated fix just in case.

pamaury:
This is a known issue with `rustfmt_check` only checking manually specified targets. There is a fix in #21631 but it requires a rules_rust change so it's low priority. Thanks for this PR.

pamaury (approved):

Issue #16855, opened 2022-12-15
[kmac] Expecting error after clearing sideload key


gdessouky:
When KMAC is configured in sideload mode, and it receives an invalid key at the sideload port after clearing (see test #16492), this ought to be indicated in the `ERR_CODE`, also in line with the documentation https://docs.opentitan.org/hw/ip/kmac/doc/#keynotvalid-0x01.

This doesn't happen however probably because it is a SW-initiated operation, although the sideload key is invalid. It is probably consuming that invalid (random) key generated at the sideload port to compute the digest. Is this intentional but an error/ambiguity in the documentation or a bug in the error reflection in `ERR_CODE` ?

cc @ballifatih @johannheyszl  @cindychip @eunchan @jadephilipoom

ballifatih:
We had an offline discussion with @gdessouky, and the issue might be due to `kmac_app` FSM not checking `valid` bit of the sideload key within `StSw` state.

Here is waveform from kmac sideload test where we try to use the sideload key after clearing (`st` is the FSM of `kmac_app`):
![Screenshot from 2022-12-15 15-43-09](https://user-images.githubusercontent.com/7850700/207910431-c8bc4a82-64eb-4b2b-9dec-868446427a97.png)

tjaychen:
ah good catch, it does seem like it should be error'd out. Unless we explicitly decided at some point to make it somewhat "hidden" (although we don't do the same with the app interface...).

@cdgori / @eunchan do you have any recollection here? I feel like an error would be better here.

cdgori:
No specific recollection, but it seems as if it should be an error to me as well.

tjaychen:
@vogelpi / @eunchan i'm not completely sure who is handling kmac D3 right now, but would this be a pretty simple fix?

ballifatih:
Marking this as P1, because if an attacker clears the sideload port and then initiates a KMAC operation, then it can leak information on keymgr's internal LFSR state.

It is better to prevent his altogether rather than doing the security analysis of "What can go wrong if an attack recovers LFSR state".

cindychip:
Related issue: https://github.com/lowRISC/opentitan/issues/14596

cindychip:
I believe there is a smiliar issue reported here:  https://github.com/lowRISC/opentitan/issues/10704

ballifatih:
[...]

Issue #10704, opened 2022-02-08
[kmac] kmac does not check key validation in sw initiated sequence


cindychip:
Hi Eunchan,

I have two questions regarding kmac ErrKeyNotValid error:
1). Right now this error ErrKeyNotValid is specifically triggered when design is configured to KMAC mode.
     However, if design is in other mode (such as Cshake128), user can still use keymgr key input when configure the `sideload` field to 1 right?
    In that case, if keymgr's key is not valid, shall we flag a ErrKeyNotValid error? (I think current design does not flag any error)

2). Follow up with the previous question, in that case, design seems to take the invalid input `keymgr.key` from the interface, and continues the cshake operation.
     However, in DV we will drive `X` as data in the sideload interface when key is not valid.
    Then the `X` will propagates to the data, and eventually I saw an assertion error.
    How do you think we can proceed with this issue?
    a). Drive all 0s or 1s as key output when keymgr is not valid?
    b). Constrain DV to not drive `X`  when key is not valid? (I think this is less desirable)
    c). You probably have some better ideas..

Please corrent me if I misunderstand anything :)

smart ID: smart:dc11c7ec-e875-47bf-8d06-9e129c1668ed
path:  /workspace/mnt/opentitan/scratch/fix_kmac_err/kmac_masked-sim-vcs/0.kmac_masked_error/out/
![image](https://user-images.githubusercontent.com/11466553/153093991-0d2aad94-494a-4dc4-96d6-ab591f32eb6c.png)

Thanks,
Cindy

eunchan:
> Hi Eunchan,
> 
> I have two questions regarding kmac ErrKeyNotValid error:
> 1). Right now this error ErrKeyNotValid is specifically triggered when design is configured to KMAC mode.
> However, if design is in other mode (such as Cshake128), user can still use keymgr key input when configure the sideload field to 1 right?
> In that case, if keymgr's key is not valid, shall we flag a ErrKeyNotValid error? (I think current design does not flag any error)
> 
Only when the key is needed, KMAC checks the side load key is valid or not. cSHAKE, SHAKE, SHA mode does not use the key.
> 2). Follow up with the previous question, in that case, design seems to take the invalid input keymgr.key from the interface, and continues the cshake operation.
> However, in DV we will drive X as data in the sideload interface when key is not valid.
> Then the X will propagates to the data, and eventually I saw an assertion error.
> How do you think we can proceed with this issue?
[...]

PR #22794, opened 2024-04-24
[kmac] Abort when sideload key is invalid during operation

andreaskurth:
The KMAC HW IP block features an option to load keys from Key Manager
via a HW key sideload interface.  Prior to this commit, KMAC would:
- when used via the SW application interface: *not check at all* if the
  sideload key is valid (issue https://github.com/lowRISC/opentitan/issues/10704, https://github.com/lowRISC/opentitan/issues/16855);
- when used via a HW application interface: check if the sideload key is
  valid *only for a single cycle* when the application interface gets
  configured (state `StAppCfg` in `kmac_app`).

This could lead to cases where KMAC would use an invalid sideload key. This PR fixes that by raising an error in these cases.

To enable that, this PR moves the `err_processed` bit from the `CFG_SHADOWED` CSR to the `CMD` CSR, and this PR fixes three problems with KMAC's error handling (some of which would only occur with the newly introduced error).

Please see the commit messages for details on the changes in this PR.

Block-level tests maintain their pass rate (tested with `-i all -rx 0.1`).

### Follow-up tasks
- [ ] DV: Cover sideload key becoming invalid during operation in all possible states (part of #16855, although we could create a new issue for clarity).
- [ ] Arch/RTL: Decide if we need a way to recover from an error if the HW application interface does not provide the last data item.

### DIFF
--- a/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
+++ b/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
@@ -99,6 +99,7 @@
     {
       name: "{name}_sideload"
       uvm_test_seq: kmac_sideload_vseq
+      run_opts: ["+test_timeout_ns=1_000_000_000"]
     }
     {
       name: "{name}_burst_write"
--- a/hw/ip/kmac/rtl/kmac_app.sv
+++ b/hw/ip/kmac/rtl/kmac_app.sv
@@ -108,6 +108,8 @@ module kmac_app
   // SW sets err_processed bit in CTRL then the logic goes to Idle
   input err_processed_i,
 
+  output prim_mubi_pkg::mubi4_t clear_after_error_o,
+
   // error_o value is pushed to Error FIFO at KMAC/SHA3 top and reported to SW
   output kmac_pkg::err_t error_o,
 
@@ -170,43 +172,31 @@ module kmac_app
   };
 
   // Encoding generated with:
-  // $ ./util/design/sparse-fsm-encode.py -d 3 -m 11 -n 10 \
-  //      -s 155490773 --language=sv
+  // $ ./util/design/sparse-fsm-encode.py -d 3 -m 14 -n 10 \
+  //     -s 2454278799 --language=sv
   //
   // Hamming distance histogram:
   //
   //  0: --
   //  1: --
   //  2: --
-  //  3: ||||||||||||| (14.55%)
-  //  4: |||||||||||||||||||| (21.82%)
-  //  5: |||||||||||||||||| (20.00%)
-  //  6: |||||||||||||||||||| (21.82%)
-  //  7: |||||||||||||||||| (20.00%)
-  //  8: | (1.82%)
+  //  3: |||||||||| (14.29%)
+  //  4: |||||||||||||||||||| (27.47%)
+  //  5: ||||||||||||| (18.68%)
+  //  6: |||||||||||||||| (21.98%)
+  //  7: |||||||| (10.99%)
+  //  8: |||| (6.59%)
   //  9: --
   // 10: --
   //
   // Minimum Hamming distance: 3
   // Maximum Hamming distance: 8
-  // Minimum Hamming weight: 2
-  // Maximum Hamming weight: 9
+  // Minimum Hamming weight: 3
+  // Maximum Hamming weight: 8
   //
   localparam int StateWidth = 10;
-  // States
-  //  StIdle                  = 10'b0101110011,
-  //  StAppCfg                = 10'b0001010000,
-  //  StAppMsg                = 10'b0001011111,
-  //  StAppOutLen             = 10'b1011001111,
-  //  StAppProcess            = 10'b1000100110,
-  //  StAppWait               = 10'b0010010110,
-  //  StSw                    = 10'b0111111111,
-  //  StKeyMgrErrKeyNotValid  = 10'b1001110100,
-  //  StError                 = 10'b1101011101,
-  //  StServiceRejectedError  = 10'b1110000110,
-  //  StTerminalError         = 10'b0010001001
   typedef enum logic [StateWidth-1:0] {
-    StIdle = 10'b0101110011,
+    StIdle = 10'b1010111110,
 
     // Application operation.
     //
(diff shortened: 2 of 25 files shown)


hw/ip/kmac/rtl/kmac_app.sv
```diff
@@ -530,36 +518,69 @@ module kmac_app
         // In this state, the state machine flush out the request
         st_d = StError;
 
-        fsm_data_ready = 1'b 1;
+        // Absorb data on the app interface.
+        fsm_data_ready = 1'b1;
+
+        unique casez ({err_processed_i, app_i[app_id].valid && app_i[app_id].last})
```
rswarbrick, line 555: Nit: (Me checking my SystemVerilog knowledge!) Is there any reason to use `casez` instead of `case` here? I think they should be equivalent if none of the cases have `?` characters?

hw/ip/kmac/rtl/kmac_app.sv
```diff
@@ -530,36 +518,69 @@ module kmac_app
...
+            service_rejected_error_clr = 1'b1;
+            st_d = StIdle;
+          end
+          default: st_d = StError;
+        endcase
+      end
 
+      StErrorAwaitSw: begin
+        // Just wait for SW to process the error.
+        st_d = StErrorAwaitSw;
```
rswarbrick, line 595: Is this line needed? Don't we already have `st_d = st` above?

hw/ip/kmac/rtl/kmac_app.sv
```diff
@@ -530,36 +518,69 @@ module kmac_app
...
-          // Send garbage digest to the app interface to complete transaction
-          fsm_digest_done_d = 1'b 1;
-
-          if (service_rejected_error) begin
-            // If service was rejected, it is assumed the SW is not loaded
-            // yet. In this case, return the request with error and keep
-            // moving, hoping SW may handle the errors later
-            st_d = StServiceRejectedError;
-          end
+          // Send garbage digest to the app interface to complete the transaction.
```
rswarbrick, line 608: Nit: Maybe these lines need 1. / 2. prefixes to match the equivalent comments a few lines earlier?

hw/ip/kmac/rtl/kmac_app.sv
```diff
@@ -530,36 +518,69 @@ module kmac_app
...
+          fsm_digest_done_d = 1'b1;
+          // Clear internal variables and return to idle.
+          clr_appid = 1'b1;
```
vogelpi, line 611: I fear this doesn't work:
- `clr_appid` causes `app_id` to be 0 in the next clock cycle, while
- `fsm_digest_done_d` causes `app_o[app_id].done` to raise in the next clock cycle. 

So this code will always send a `done` pulse on app interface 0.

vogelpi (commented):
Thanks for investigating and fixing this @andreaskurth ! The code looks mostly good but there seems to be a bug in one of the new states which may cause KMAC to send a done pulse on the wrong app interface.

As for the new issue of misbehaving HW app interfaces stalling KMAC: I think the current behavior may be acceptable but it's probably worth to have a separate discussion. Would you mind creating a new GitHub issue please?

hw/ip/kmac/rtl/kmac_app.sv
andreaskurth: No, it's not. Sorry, I forgot to clean this up from a first implementation. I'll fix it

hw/ip/kmac/rtl/kmac_app.sv
andreaskurth: Indeed, thanks for catching this! Fixed

hw/ip/kmac/rtl/kmac_app.sv
andreaskurth: Thx, changed

hw/ip/kmac/rtl/kmac_app.sv
```diff
@@ -435,8 +435,9 @@ module kmac_app
 
           service_rejected_error_set = 1'b 1;
 
-        end else if ((AppCfg[app_id].Mode == AppKMAC) &&
-          !keymgr_key_i.valid) begin
+        end else if ((AppCfg[app_id].Mode == AppKMAC) && !keymgr_key_i.valid) begin
```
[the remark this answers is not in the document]
andreaskurth: Thx for this suggestion. I found out that I needed to extend error handling anyway, so I implemented your suggestion along with it.

hw/ip/kmac/rtl/kmac_app.sv
```diff
@@ -530,36 +521,87 @@ module kmac_app
...
-          if (service_rejected_error) begin
-            // If service was rejected, it is assumed the SW is not loaded
-            // yet. In this case, return the request with error and keep
-            // moving, hoping SW may handle the errors later
-            st_d = StServiceRejectedError;
-          end
+          // Send garbage digest to the app interface (in the next cycle) to complete the
+          // transaction.
+          fsm_digest_done_d = 1'b1;
+          cmd_o = CmdProcess;
```
vogelpi, line 576: Is it worth to add a comment here while we send this Process command (I believe it's to flush the SHA3 engine).

hw/ip/kmac/rtl/kmac_app.sv
```diff
@@ -530,36 +521,87 @@ module kmac_app
...
+          default: st_d = StError;
+        endcase
+      end
 
-          // clear internal variables
-          clr_appid = 1'b 1;
+      StErrorAwaitSw: begin
+        // Just wait for SW to process the error.
+        if (err_processed_i) begin
+          cmd_o = CmdProcess;
```
vogelpi, line 564: Should we add a comment to say we need to flush the SHA3 engine here?

vogelpi (approved):
This looks great, thanks @andreaskurth for fixing this and explaining what's going on. Nice work!

hw/ip/kmac/rtl/kmac_app.sv
andreaskurth: Yes, thx, done

ballifatih (approved):
Thanks for the fix @andreaskurth. LGTM, but I must admit I could not fully review the full FSM transitions.

Restoring KMAC from an error has always been a bit tricky from SW side. I will have to keep that in mind that for the cryptolib. This PR really does not touch that clearing mechanism, so I think we should be fine (except for relocating `err_processed` bit, for which this PR also includes changes).

The following top level tests might provide some coverage (although they are not targeting this particular fix):

- sw/device/tests/keymgr_sideload_kmac_test.c
- sw/device/tests/kmac_entropy_test.c

andreaskurth:
Thx for the review @ballifatih. Actually the `keymgr_sideload_kmac_test` flags an issue that I'm now resolving.

andreaskurth:
Alright, I think I fixed the problems indicated by the failing `keymgr_sideload_kmac_test` TLT:
- SW expected KMAC to complete (with a bogus output) when the sideload key from keymgr is invalid, whereas it now raises an error that needs to be handled;
- key stability assertions in `kmac_core` no longer made sense when `kmac_app` is able to abort upon an invalid key;
- the state of the `kmac_errchk` module did not get cleared when error handling was complete;
- `kmac_app` waited for data from a HW app interface even when the error occurred during SW operation.

vogelpi:
Update: rebased on top of #22940.

andreaskurth:
Thx! With that, CI finally passes. Merging

cdgori (commented):
LGTM modulo the one hjson bit that might have snuck in (?)

hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
```diff
@@ -99,6 +99,7 @@
     {
       name: "{name}_sideload"
       uvm_test_seq: kmac_sideload_vseq
+      run_opts: ["+test_timeout_ns=1_000_000_000"]
```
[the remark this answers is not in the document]
andreaskurth: Yes, this is necessary because kmac now drains all data beats on the HW app interface when there's an error.

Re this being very long -- time is relative, right? :-) kmac block level DV simulates at a relatively high rate (simulated time over wall clock time), and other tests have an even higher timeout (e.g., `kmac_long_msg_and_output` above has 10 s).

Issue #20589, opened 2023-12-07
[test-triage] Fix rstmgr_alert_info_test_fpga_cw310_sival


matutem:
### Hierarchy of regression failure

Chip Level

### Failure Description

The rstmgr_alert_info_test_fpga_cw310_test_rom passes, but the _sival and _sival_rom_ext variants
fail. The failure is on round 3, when the test gets an unexpected flash_ctrl fatal alert. To reproduce it,
uncomment these tests from https://github.com/lowRISC/opentitan/blob/master/hw/top_earlgrey/data/ip/chip_rstmgr_testplan.hjson
and run the test on a hyperdebug board with
bazel test //sw/device/tests:rstmgr_alert_info_test_fpga_cw310_sival

### Steps to Reproduce

bazel test //sw/device/tests:rstmgr_alert_info_test_fpga_cw310_sival

### Tests with similar or related failures

- [ ] //sw/device/tests:rstmgr_alert_info_test_fpga_cw310_sival_rom_ext

matutem:
I find that upon entry to test_main the flash_ctrl has RD_ERR set in the FAULT_STATUS CSR, which seems wrong.

matutem:
An update on this strange test:
- this test was expecting an otp_ctrl fatal integrity error in round 2, but apparently this is not happening in sival_rom_ext. I cannot find why this was expected, but perhaps this was fixed in SW?
- In plain fpga_cw310_sival the otp_ctrl fatal integrity error still happens.
When I remove the expectation of an otp_ctrl alert.
- in sival_rom_ext it is only failing in fpga_cw310 due to an unexpected flash_ctrl alert.

So it seems sival_rom_ext avoids the otp_ctrl fatal integrity error, but in fpga_cw310_sival_rom_ext we get an unexpected flash_ctrl alert.

jdonjdon:
@matutem from rom_ext, creator and owner partitions are locked down.
(https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/rom_ext/rom_ext.c;drc=049de361f312a71e7f82e4dcc6b9c9106f404597;l=284)
So, I guess [this line](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/tests/rstmgr_alert_info_test.c;drc=e5b074b89ee185a45a1008415cc6839a14a033ad;l=752) can causes failure with rom_ext

matutem:
[...]

PR #22795, opened 2024-04-24
[Sival, rstmgr] Fix rstmgr_alert_info_test on ROM_EXT

engdoreis:
Fixes https://github.com/lowRISC/opentitan/issues/20589

The Root cause was that the ROM_EXT locks up the OTP registers with the ePMP, any attempt to write to them would generate an exception.
Therefore, we can't test the OTP alert when the test runs after the ROM_EXT.

### DIFF
--- a/sw/device/tests/rstmgr_alert_info_test.c
+++ b/sw/device/tests/rstmgr_alert_info_test.c
@@ -77,15 +77,12 @@
 OTTF_DEFINE_TEST_CONFIG();
 
 enum {
-  kWdogBarkMicros = 200,          // us
-  kWdogBiteMicros = 200,          // us
-  kEscalationPhase0Micros = 300,  // us
-  kEscalationPhase1Micros = 200,  // us
-  kEscalationPhase2Micros = 100,  // us
-  kRoundOneDelay = 100,           // us
-  kRoundTwoDelay = 100,           // us
-  kRoundThreeDelay = 1000,        // us
-  kEventCounter = 0               // the retention sram counter tracking events
+  kWdogBarkMicros = 200,
+  kWdogBiteMicros = 200,
+  kRoundOneDelay = 100,
+  kRoundTwoDelay = 100,
+  kRoundThreeDelay = 1000,
+  kEventCounter = 0  // the retention sram counter tracking events
 };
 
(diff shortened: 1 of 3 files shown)


sw/device/tests/rstmgr_alert_info_test.c
```diff
@@ -701,6 +701,12 @@ static void init_expected_cause(void) {
       .alert_info.alert_cause[kTopEarlgreyAlertIdI2c0FatalFault] = 1;
   kExpectedInfo[kRound3]
       .alert_info.alert_cause[kTopEarlgreyAlertIdSpiHost0FatalFault] = 1;
+
+  if (kBootStage == kBootStageOwner &&
```
engdoreis, line 705: On Round 3 all the alerts are enabled, after the reset is caused by the alert handler escalation, a `kTopEarlgreyAlertIdFlashCtrlFatalErr` happens during the boot, only with FGPA + ROM_EXT. 
This doesn't happen on Silicon, probably because the Creator and Owner secrets are there, but I'm not sure why it happens on the FPGA. 
@cfrantz Could you check that this is not a bug with ROM_EXT?

sw/device/tests/rstmgr_alert_info_test.c
```diff
@@ -790,10 +802,17 @@ bool test_main(void) {
...
+      if (kBootStage != kBootStageOwner) {
+        // The ROM_EXT locks up the OTP register in the ePMP, any attempt to
+        // write to it would generate an exception. Therfore we can't test the
+        // OTP alert when the test runs after the ROM_EXT.
+        CHECK_DIF_OK(dif_otp_ctrl_alert_force(
+            &otp_ctrl, kDifOtpCtrlAlertFatalBusIntegError));
+        CHECK_DIF_OK(dif_alert_handler_irq_set_enabled(
+            &alert_handler, kDifAlertHandlerIrqClassb, kDifToggleEnabled));
+      } else {
+        LOG_INFO("Skipping OTP test due to ROM_EXT");
```
moidx, line 814: nit: `Skipping OTP test due to ROM_EXT ePMP configuration`

moidx (approved):

PR #22796, opened 2024-04-24
[fpv,prim] Generalise FPV parameter that constrains what actions a prim_count is asked for

rswarbrick:
This is all about avoiding some vacuous assertions that are reported in prim FPV tests. The first commit generalises a `DecrNeverTrue` parameter that I added recently and allows a user to specify exactly which actions are possible in the given instance.

The second commit uses the new `PossibleActions` parameter in the alert_handler's ping timer, where we have two different instances of `prim_count` which are used in two different ways.

The end result should be to improve the "pass rate" in the nightly prim FPV tests.

### DIFF
--- a/hw/ip/prim/lint/prim_count.vlt
+++ b/hw/ip/prim/lint/prim_count.vlt
@@ -4,6 +4,6 @@
 
 `verilator_config
 
-// This parameter is only used in DV/FPV.
+// These parameters are only used in DV/FPV.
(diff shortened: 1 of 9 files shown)


vogelpi (approved):
Thanks @rswarbrick , this looks good!

rswarbrick:
The force-push with a tidyup seems to fix things. The Kokoro build is still failing, but that's an unrelated ascon issue which is being fixed elsewhere.

PR #22798, opened 2024-04-24
[alert_handler,rtl] Slightly simplify decrementing counter

rswarbrick:
The FSM in the ping timer uses a prim_count to count down from a timeout value (ping_timeout_cyc_i) to zero. It then avoids telling the prim_count module to decrement further.

But the prim_count saturates at zero already, so that check isn't actually needed.

(Spotted by debugging why an FPV property about decrementing when equal to zero was unreachable)

Note that the FPV property in question has its coverage property waived in PR #22797. If both PRs land, the second one to land should be amended to drop the waiver.


vogelpi (approved):
LGTM!

PR #22800, opened 2024-04-24
tock: comment out nonexisting test module reference

vbendeb:
This makes it possible to run rustfmt acroll the entire tree.

### DIFF
--- a/sw/device/silicon_owner/tock/kernel/src/main.rs
+++ b/sw/device/silicon_owner/tock/kernel/src/main.rs
@@ -42,8 +42,6 @@ use rv32i::csr;
 
 pub mod io;
 mod otbn;
-#[cfg(test)]
-mod tests;
 
 const NUM_PROCS: usize = 4;
 


sw/device/silicon_owner/tock/kernel/src/main.rs
```diff
@@ -42,9 +42,12 @@ use rv32i::csr;
 
 pub mod io;
 mod otbn;
+/*
+ * This could be reinstated if test code is brought in, see
+ * https://github.com/lowRISC/opentitan/issues/22737
 #[cfg(test)]
 mod tests;
-
+*/
```
jrvanwhy, line 50: We will not put tests into this crate for the forseeable future, so delete this entirely rather than commenting it out.

Unit tests for drivers/components within the kernel belong in the driver/component crates, and integration tests that run on hardware belong in separate test crates under `sw/device/silicon_owner/tock/tests`.

vbendeb: sure, done.

jrvanwhy (approved):

vbendeb:
so, now as this patch has been approved how does it get merged? Is there a subset of users who have the merge privileges?

jrvanwhy:
> so, now as this patch has been approved how does it get merged? Is there a subset of users who have the merge privileges?

Yes, and I am not one of those users. Tim is, and he was automatically added as a reviewer.

timothytrippel (approved):

PR #22804, opened 2024-04-24
[dv,flash_ctrl] Enable random read-write tests with ecc enabled

### LINKED ISSUE #22559: [flash_ctrl] Extend block-level DV test cases
### Description

Related to #22443, #22444, #22549, we've realized that we should probably extend block-level DV to cover some more cases (where they are not yet covered):
- [x] Change the `flash_ctrl_phy_arb_redun` to inject faults into both copies of all four redundant arbiters. In case the right copy is glitched to 1, this may inject new transactions into the read pipeline as a side effect trigger ECC and other errors. It may be necessary to ignore some of these errors during the fault injection (as we have been ignoring the access permission errors or mp_err bit already). --> #22444 
- [ ] Stress test both flash banks in parallel (without FI) but with ECC enabled.
- [ ] Enable ECC more widely (it seems to be disabled with the original scoreboard).
- [ ] Fix DV sequence handling of program resolution errors.

### LINKED ISSUE #22746: [dv,flash_ctrl] Fix handling program resolution
### Description

The flash program sequence generator doesn't have a constraint that avoids program requests crossing program window boundaries (which are aligned at 64 bytes (8 flash words). Rather than detecting this and expect an error and drop the update, `prog_flash` rewrites the request. Unfortunately this rewrite generates multiple writes to the same locations when multiple consecutive writes are requested.

I think if anything the rewrite should instead break a request that spills into two portions, each fully contained within a single window. There are dedicated tests that check that crossing errors are generated.

matutem:
We didn't have extensive testing of reads to adddresses peviously written with ECC and scrambling enabled. This PR enables ECC for flash_ctrl_rw_vseq and some others.

This requires a number of other changes:
- Writes must be done for full flash words, so the 32-bit chunks must be even and their address 8-byte aligned.
- Writes cannot overwrite contents already written or most likely some bit will not be able to flip from 0 to 1. This maintains a collection of addresses that have been already written to avoid overwrites.
- A bug is fixed for writes that span multiple program windows (aligned at 64 bytes): the dv code was replacing the sequence of writes by repeated writes to the start of the program window. Instead, this breaks any request that spills across windows in two.

Fixes #22746
Fixes #22559

### DIFF
--- a/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
+++ b/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
@@ -192,7 +192,7 @@ class cip_base_vseq #(
                       tl_sequencer_h, tl_intg_err_type);
   endtask
 
-  // this tl_access can input req_abort_pct (pertentage to enable req abort) and output status for
+  // this tl_access can input req_abort_pct (percentage to enable req abort) and output status for
   // seq to update predicted value
   virtual task tl_access_w_abort(
       input bit [BUS_AW-1:0]  addr,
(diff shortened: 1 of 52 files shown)


eshapira (approved):
LGTM, thanks!

PR #22814, opened 2024-04-25
[dv,clkmgr] Used synchronized sw enables in SVA

matutem:
The SVAs for peripheral clock gating need to use inputs synchronized to the clock they control, or they will be unrealistic. It is possible that the gating controls will change too quick and a pulse will be missed, causing assertion failures. In reality this is a corner case since the expectation is that gating controls will change infrequently.


matutem:
This was the cause of the one test failing in block level regressions.

rswarbrick (approved):

PR #22815, opened 2024-04-25
[ascon] Fix AscentLint warning

vogelpi:
The Ascon hardware block is very much work in progress. Unfortunately, lint warnings and errors inside Ascon cause the whole AscentLint CI job to fail.

### DIFF
--- a/hw/ip/ascon/rtl/ascon_core.sv
+++ b/hw/ip/ascon/rtl/ascon_core.sv
@@ -42,7 +42,7 @@ assign hw2reg = 'b0;
 
 // TODO
 logic d, q;
-assign d = 1'b1;
+assign d = keymgr_key_i.valid;
 
 logic unused_q;
 assign unused_q = q;


m-temp (approved):

m-temp:
Thx @vogelpi for fixing this. I was not aware, that an error or warning inside Ascon causes the whole AscentLint CI to fail...

Should we disable ascnentlint for now, or do we/I need to pay more attention an ascentlint?
I think it should be more relaxed in the future to commit error/warning free code, as there will be less parts that depend on each other.

vogelpi:
> Thx @vogelpi for fixing this. I was not aware, that an error or warning inside Ascon causes the whole AscentLint CI to fail...
> 
> Should we disable ascnentlint for now, or do we/I need to pay more attention an ascentlint? I think it should be more relaxed in the future to commit error/warning free code, as there will be less parts that depend on each other.

I think we should aim to keep it enabled for Ascon and keep Ascon lint clean right from the beginning.

timothytrippel:
Thanks @vogelpi !

PR #22820, opened 2024-04-25
[prim,rtl,fpv] Fix typo in assertion in prim_alert_receiver

rswarbrick:
This precondition for the assertion was never possible because !state_q was never InitReq or InitAckWait.

Unfortunately, fixing this makes the assertion false. It turns out that the problem was that the init_trig_i signal squashes the ping. Add a condition based on this.

Annoyingly the result is very repetitive! To make things a bit tidier, move the condition into the __rst argument of the `ASSERT macro.

Finally, drop the initial "##1" (which avoided the property triggering just after reset). The $rose(...) condition implies this anyway (because it's false on the first cycle after reset), so we can drop the cycle. This makes the property marginally stronger (and Jasper still proves it!)


vogelpi (approved):
LGTM!

PR #22825, opened 2024-04-25
[fpv,alert_handler] Remove two unnecessary cover waivers

rswarbrick:
These were added in 4668da7 but the waivers were needed because of something fixed in the design by cd303a4. Remove the waivers accordingly.

(I meant to squash the change that drops these waivers into the second commit above, but forgot. That PR is #22798)


marnovandermaas (approved):
Looks reasonable to me.

PR #22826, opened 2024-04-25
[sival, rv_core_ibex] fix ibex address translation test

nbdd0121:
Due to PMP lockdowns, we cannot map arbitrary address spaces, so instead we remap within .text sections. The illegal access fault is replaced with illegal instruction fault when the address is not mapped.


jwnrt:
The Azure logs have gone from this run so I can't tell if it's related (and annoyingly Azure won't let me rerun).

Please could you rebase?

PR #22829, opened 2024-04-25
[prim,rtl] Avoid unnecessary check in prim_esc_receiver.sv

rswarbrick:
The code in prim_esc_receiver tells the instantiated prim_count to increment the counter unless the value is '1. But prim_count is already designed to saturate, so there's no need to condition incr_en_i like this.

Found by an unreachable cover property when running FPV tests.


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #22834, opened 2024-04-25
[sram_program] Rename `sram_main` to `test_main`

nbdd0121:
This makes it easier to use the a test as both sram main and a normal test.

The rv_core_ibex_isa_test currently uses a0 for returning error code, which I have changed it to a1 so that a0 is used as `bool test_main` return code.

This is a preparatory PR to run some rv_core_ibex SRAM program tests as normal tests when LC state is PROD.


pamaury (approved):
Looks good to me

cfrantz (approved):

PR #22840, opened 2024-04-25
[rom_ext] Improve `boot_log` reporting

cfrantz:
1. Add an `ownership_transfers` member which will report the total number of ownership transfers the chip has had.
2. Use the remap window to detect in which flash slot the ROM_EXT is running.
3. Test all variations of owner slots and ROM_EXT slots.


jettr (approved):
LGTM, but would probably benefit with someone more familiar with code to ensure nothing is missing :)

moidx (approved):
LGTM

PR #22841, opened 2024-04-25
[bazel] Remove opentitantest references from build files

moidx:
1. Remove unused `opentitan_functest` references from BUILD files.
2. Remove `uart_smoketest_signed` as this is equivalent to the `rom_e2e_smoketest`.

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -17,12 +17,6 @@ load(
     "RSA_ONLY_KEY_STRUCTS",
 )
 load(
-    "//rules:opentitan_test.bzl",
-    "ROM_BOOT_FAILURE_MSG",
-    "opentitan_functest",
-    dv_params_functest = "dv_params",
-)
-load(
     "//rules:otp.bzl",
     "STD_OTP_OVERLAYS",
     "otp_image",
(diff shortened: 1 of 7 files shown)


cfrantz (approved):

sw/device/tests/BUILD
```diff
@@ -3613,16 +3607,19 @@ opentitan_test(
 )
 
 # Used by DVSIM test-points to test signed binaries.
-# Please do not remove until those tests are updated to use `uart_smoketest`.
-opentitan_functest(
+# TODO(#21204): Please do not switch to opentitan_functest as this will cause
+# issues with ongoing transition to ECDSA keys.
+opentitan_test(
     name = "uart_smoketest_signed",
```
a-will, line 3613: Should this just merge with `uart_smoketest` below?

a-will: Also... it actually doesn't run in DV. I just tried. :)

https://github.com/lowRISC/opentitan/pull/22739 was getting part of the way there, but there are still things I hadn't figured out.

moidx: Hi @a-will, for the OTP issue you were seeing, something like this should fix it: 

https://github.com/lowRISC/opentitan/blob/d20279803b6b475fab8e320e541cc400f7f5095a/hw/top_earlgrey/dv/chip_rom_tests.hjson#L14-L16

We should eventually migrate all DV tests to use OTP images generated by Bazel.

a-will: Sounds good to me! `rom_e2e_smoke` is also broken, but it already was.

Issue #16202, opened 2022-11-10
[aes, kmac, sca] Repeat formal masking verification experiments prior to tapeout


vogelpi:
@johannheyszl  brought it up in today's SCA meeting that we need to repeat the formal masking verification experiments close to tapeout. We have been running these experiments in the past (as well as with FPGA measurements) but there is the risk that something is introduced into the the RTL which doesn't show on FPGA but only in the ASIC netlist.

Related to that, we should try to port he formal masking verification flow into the production environment.

> estimate 64
> remaining 2023-03-23 63

vogelpi:
Triaged for aes, needed for M2.5 because the masking countermeasures are very important for TO.

vogelpi:
Update: @arnonsha , @zi-v , @ballifatih , @johannheyszl and I have been discussing how we can achieve this. We've agreed on the following procedure and sub tasks: 
- [ ] @vogelpi to share a how-to and a pre-synthesized netlist of the AES S-Box obtained from Yosys as well as tool prerequisites
- [ ] @zi-v to get the formal tool and requirements installed locally and to perform a dry run using the Yosys netlist
- [ ] @vogelpi to contact formal tool authors and clarifying how exactly to add stdcell library information to the formal tool
- [ ] @ballifatih to integrate the closed-source stdcell library information to a closed copy of the tool and perform a test run using a netlist generated from the closed-source flow
- [ ] @zi-v and @arnonsha to repeat the experiment using this closed tool version an AES S-Box netlist obtained from the real backend flow

After that, this pre-verified and synthesized AES S-Box will be loaded into the final netlist.

moidx:
Planning to work on this post M2.5, but leaving as part of the milestone as is something that needs to be done with the netlist after we sign-off the block.

GregAC:
@vogelpi @moidx is this actually needed for M2.5.2?

johannheyszl:
@msfschaffner is talking to Nuvoton about running this for AES. @ballifatih has prepared the tooling.

johannheyszl:
IMO the consensus is that we cannot make it a requirement and it is best effort for M2.5.3.

GregAC:
Great, in that case I'll shift this to M2.5.3

johngt:
Please note change of status and milestone movement @vogelpi

vogelpi:
Thanks for the heads up @johngt . This sounds good to me.

[...]

PR #22844, opened 2024-04-25
[aes] Randomly advance the masking PRNG when not used during processing

### LINKED ISSUE #21167: [aes] Potential minor PRNG-related modifications for production
### Description

Following the big PRNG change in #20852, there are some minor changes that we might want to implement, also based on the penetration testing results:
- The 8-bit buffers for the PRD input inside the DOM S-Boxes should be moved up in the hierarchy. Ideally it's placed inside the `aes_cipher_core.sv` module. This doesn't change SCA properties but it simplifies control. This further allows taking the randomness for masking the initial state input from the buffers rather than from the PRNG directly (might be glitchy). This should have a beneficial impact on SCA hardening (to be evaluated on FPGA).
- The PRNG should be advanced in every cycle (but not the FF-based buffers of course) to increase the noise level. This needs the change above.

This won't have a noticeable timing or area impact. I'm thus putting in M3.

vogelpi:
This PR contains a couple of changes, but only two of them have functional/SCA impact:
- [[aes] Randomnly advance the masking PRNG when not used during processing](https://github.com/lowRISC/opentitan/commit/eea56e4ba21bec9d005989e8588336e578e0bcc7) changes as follows:
  - The PRNG output width is slightly increased to create some auxiliary randomness that is fed into the cipher core FSM.
  - There, we use this randomness to decide whether the masking PRNG is advanced during processing cycles where no fresh randomness is required. This is beneficial for SCA hardening as it introduced additional noise. When idling, the behavior of AES doesn't change. This resolves #21167 and paves the way to solve #16202.
- [[aes] Use the buffered PRNG output data for input data masking](https://github.com/lowRISC/opentitan/commit/76e7d69deee704102eea2d41c749b7b568466d86) changes the design to use the buffered PRNG output for masking the input data. This reduces the amount of XOR operations with the plaintext when advancing the PRNG output.

I suggest disabling showing whitespace changes during the review as this PR contains some alignment changes to port lists. 

For the other commits:
- [[aes] Move PRD input buffers out of S-Boxes](https://github.com/lowRISC/opentitan/commit/e97e68541801bf539b251796ae8a8321713e98f4) moves around some FFs to enable the relevant changes above.
- [[aes, rtl] Uniquify type names for PRD structs inside DOM S-Box](https://github.com/lowRISC/opentitan/commit/2aee53dfdd29901d5706bb535364181f5faca039) is just a cherry-pick from integrated_dev.
- [[aes] Correct some comments regarding the masking PRNG implementation](https://github.com/lowRISC/opentitan/commit/4451178bf9e9f063281da75ae499fcde9b8f1a3b) is just changing comments in the code.


vogelpi:
I still need to run the FPGA experiments which will happen tomorrow most likely.

nasahlpa (approved):
Thanks, looks good for me from a security point of view!

Are you planning to share the results from the FPGA measurement? Would be interesting for me to see the effect of advancing the PRNG.

hw/ip/aes/rtl/aes_cipher_control_fsm.sv
```diff
@@ -479,6 +483,38 @@ module aes_cipher_control_fsm import aes_pkg::*;
...
+    always_ff @(posedge clk_i or negedge rst_ni) begin : reg_prd
+      if (!rst_ni) begin
+        prd_q <= '0;
+      end else begin
+        prd_q <= prd_d;
+      end
+    end
+
+    // Advance the masking PRNG based on the FSM and the shift register.
+    assign prng_update_o = prng_update | prd_q[0];
```
johannheyszl, line 507: @vogelpi please let's discuss this and L497 shortly

hw/ip/aes/rtl/aes_cipher_core.sv
```diff
@@ -316,9 +316,29 @@ module aes_cipher_core import aes_pkg::*;
...
+    //    for remasking.
+    // 2. Prevent glitches originating from inside the masking PRNG from propagating into the
+    //    masked S-Boxes.
+    always_ff @(posedge clk_i or negedge rst_ni) begin : prd_sub_bytes_reg
+      if (!rst_ni) begin
+        prd_sub_bytes_q <= '0;
+      end else if (state_we == SP2V_HIGH) begin
+        prd_sub_bytes_q <= prd_sub_bytes_d;
+      end
+    end
```
andreaskurth, line 344: Just to make sure I understand this correctly: the flops that get added here ...
(see next comment)

hw/ip/aes/rtl/aes_sbox_dom.sv
```diff
@@ -1068,19 +1066,9 @@ module aes_sbox_dom
...
-  assign prd1_d = prd_we_i ? prd_i[7:0] : prd1_q;
-  prim_flop #(
-    .Width      ( 8  ),
-    .ResetValue ( '0 )
-  ) u_prim_flop_prd1_q (
-    .clk_i  ( clk_i  ),
-    .rst_ni ( rst_ni ),
-    .d_i    ( prd1_d ),
-    .q_o    ( prd1_q )
-  );
```
andreaskurth, line 1082: ... get removed here, right?

(see previous comment)

hw/ip/aes/rtl/aes_key_expand.sv
```diff
@@ -225,18 +218,50 @@ module aes_key_expand import aes_pkg::*;
...
+    //    for remasking.
+    // 2. Prevent glitches originating from inside the masking PRNG from propagating into the
+    //    masked S-Boxes.
+    always_ff @(posedge clk_i or negedge rst_ni) begin : prd_reg
+      if (!rst_ni) begin
+        prd_q <= '0;
+      end else if (prd_we) begin
+        prd_q <= prd_i;
+      end
+    end
```
andreaskurth, line 244: Oh, and flops for PRD also get added here -- so in total there are now more flops to buffer PRD? If so, how many?

vogelpi: Before we had 8 FFs inside every S-Box to buffer the PRD. Now we have 128 FFs inside aes_cipher_core + 32 FFs inside aes_key_expand. So in total that's 160 FFs = 20 S-Boxes * 8 FFs / S-Box. The number of FFs don't change.

However, we add a 4 FFs per FSM copy (that's 12 FFs in total) to buffer the auxiliary randomness to decide whether we run the PRNG during th remaining cycles of the round.

andreaskurth: I see, thanks for the explanation

andreaskurth (approved):

vogelpi:
Thanks for your reviews @johannheyszl, @nasahlpa and @andreaskurth . SCA results look all good but I've discussed with @johannheyszl and @nasahlpa that it's probably better to always let the PRNG advance when some processing is going on. I'll change the design accordingly.

johannheyszl:
Thanks @vogelpi !

cdgori (commented):
More catch-up, LGTM (I especially like the extra noise from randomly advancing the masking PRNG)

PR #22846, opened 2024-04-25
[dv,pwrmgr] Fix corner case unit block test failures

matutem:
- Fix wait times for pwrmgr_escalation_timeout test.
- Don't expect escalation timeout to generate fatal alerts when clk_esc_i is gated by io_ip_clk_en.
- Stop fast_check_wake_status upon reset, and improve the comparisons.
- Add a couple slow clock cycles after reset_en propagates to the slow domain for an extra slow synchronizer before it gets to the fast fsm.
- Add a couple more assertions for the escalation timeout path to generate a reset.
- The assertion that an incoming reset request generates a corresponding outgoing request is blocked by rst_lc_ni, since that clears escalation.
- Improve comments in pwrmgr_sec_cm_checker_assert.


hw/ip_templates/pwrmgr/dv/env/pwrmgr_scoreboard.sv
```diff
@@ -51,38 +51,40 @@ class pwrmgr_scoreboard extends cip_base_scoreboard #(
...
-                begin
-                  cfg.clk_rst_vif.wait_clks(121);
-                  if (cfg.esc_clk_rst_vif.clk_gate && cfg.pwrmgr_vif.pwr_ast_req.io_clk_en &&
-                    cfg.pwrmgr_vif.pwr_clk_req.io_ip_clk_en) begin
-                    `uvm_info(`gfn, "clk_esc_i has been inactive enough to trigger an alert",
-                              UVM_MEDIUM)
-                    `uvm_info(`gfn, "set_exp_alert from monitor_escalation_timeout clock gated",
-                              UVM_MEDIUM)
-                    set_exp_alert("fatal_fault", 1, 500);
+	  if (cfg.pwrmgr_vif.pwr_clk_req.io_ip_clk_en) begin
```
rswarbrick, line 60: I agree with verible here :-)

hw/ip_templates/pwrmgr/dv/env/pwrmgr_scoreboard.sv
```diff
@@ -51,38 +51,40 @@ class pwrmgr_scoreboard extends cip_base_scoreboard #(
...
+            // A timeout could be triggered after more than 121 clk_i cycles, but this number
+            // is somewhat unpredictable for reasons explained in the pwrmgr_escalation_timeout
+            // sequence, so this relies on the sequence to avoid unpredictable stoppages.
+            `uvm_info(`gfn, "Detected unexpected clk_esc_i stop", UVM_MEDIUM)
+            fork
+              begin : isolation_fork
+                // This fork is so we can wait for a number of cycles with the clock inactive,
```
rswarbrick, line 67: This might be worth rephrasing. I think your point is that we're waiting for N cycles on the main clock when the esc clock is gated? (My initial reading was that we were waiting for several of a gated clock's cycles, which is obviously not correct!)

hw/ip_templates/pwrmgr/dv/env/pwrmgr_scoreboard.sv
```diff
@@ -51,38 +51,40 @@ class pwrmgr_scoreboard extends cip_base_scoreboard #(
...
+                // and stop waiting if the escalation clock gate is re-opened.
+                fork
+                  begin
+                    cfg.clk_rst_vif.wait_clks(121);
```
rswarbrick, line 71: I know this isn't new, but would it be reasonable to add a note somewhere around here to explain the magic 121 number?

hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_base_vseq.sv
```diff
@@ -688,7 +691,11 @@ class pwrmgr_base_vseq extends cip_base_vseq #(
                    "wakeup_status wait timeout exp:%x init:%x", expected_wakeups, init_wakeup_status
                    ), 15_000)
       // The various bits of wakeup_status could have different sync delays, so wait some more.
-      cfg.clk_rst_vif.wait_clks(2);
+      // If reset is active all bets are off.
```
rswarbrick, line 694: I think I'd suggest expanding this comment a little bit. (It took me a minute to understand what's going on). Something like this?
```systemverilog
      // The various bits of wakeup_status could have different sync delays, so wait some more to
      // make sure everything has synchronised properly. Give up immediately if we go into reset.
```

(The tabs also need converting to spaces here)

hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_escalation_timeout_vseq.sv
```diff
@@ -9,7 +10,15 @@ class pwrmgr_escalation_timeout_vseq extends pwrmgr_base_vseq;
 
   `uvm_object_new
 
-  localparam int TIMEOUT_THRESHOLD = 128;
+  // The number of clock cycles with a stopped escalation clock before raising escalation is 128.
+  // But the logic can wait for up to 7 more cycles before it starts the counter, and there is
+  // some additional fast clock cycles of delay in the logic that triggers the escalation to
+  // be signalled. This adds 12 extra cycles to be conservative.
```
rswarbrick, line 16: This looks reasonable, but could you prefix it with a sentence that explains what the param means? We're using it in `check_stopped_esc_clk`, so maybe the name could be tweaked? Also I think localparams are usually CamelCase, so maybe `EscClkStopMainCyclesTimeout`?

(Similar notes for the following parameter)

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
```diff
@@ -60,8 +65,8 @@ module pwrmgr_sec_cm_checker_assert
   // For any assertions involving state transitions, also allow cases where the fsm
   // transitions to an invalid state, since we inject invalid encodings at random.
 
-  // Check that unless rom_intg_chk_done is mubi true the fast state machine will
-  // stay in FastPwrStateRomCheckDone.
+  // Check that while rom_intg_chk_done is mubi false the fast state machine will
```
rswarbrick, line 68: I don't think the first change is right here. The assertion is `... != MubBi4True`, which matches what the comment said before.

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
```diff
@@ -89,6 +96,9 @@ module pwrmgr_sec_cm_checker_assert
           clk_i,
           reset_or_disable)
 
+  // Check that when either rom_intg_chk_good is mubi true or rom_intg_chk_dis is mubi true
```
rswarbrick, line 99: Grammar nit: I think this might be clearer (and shorter): "Check that when one of rom_intg_chk_good and rom_intg_chk_dis is mubi true ..."

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
```diff
@@ -98,13 +108,28 @@ module pwrmgr_sec_cm_checker_assert
           clk_i,
           reset_or_disable)
 
+  // sec_cm_esc_rx_clk_bkgn_chk, sec_cm_esc_rx_clk_local_esc
```
rswarbrick, line 111: These are the names of testpoints. Could you expand slightly to make it obvious what is going on here? (I had to grep!) Maybe this can just be "For testpoints ABC, DEF." ?

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
```diff
@@ -98,13 +108,28 @@ module pwrmgr_sec_cm_checker_assert
...
+
+  // sec_cm_esc_rx_clk_bkgn_chk, sec_cm_esc_rx_clk_local_esc
+  // If escalation timeout is detected a reset request will be generated.
+  `ASSERT(EscTimeoutTriggersReset_A, esc_timeout |=> ##[1:3] slow_esc_rst_req,
+	  clk_slow_i, !rst_slow_ni || disable_sva)
+
   // pwr_rst_o.rstreqs checker
   // sec_cm_esc_rx_clk_bkgn_chk, sec_cm_esc_rx_clk_local_esc
-  // if esc_timeout, rstreqs[ResetEscIdx] should be asserted
+  // if esc_timeout, rstreqs[ResetEscIdx] should be asserted unless rst_lc_n
```
rswarbrick, line 127: I know this isn't new with this PR, but I'm slightly surprised by the condition. Isn't this "On an ESC reset request" or similar, instead of something about `esc_timeout`?

hw/ip_templates/pwrmgr/dv/env/pwrmgr_scoreboard.sv
matutem: I missed this one. Done.

hw/ip_templates/pwrmgr/dv/env/pwrmgr_scoreboard.sv
matutem: Done.

hw/ip_templates/pwrmgr/dv/env/pwrmgr_scoreboard.sv
matutem: I think this should be better now.

hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_base_vseq.sv
matutem: Good suggestion. Done.

hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_escalation_timeout_vseq.sv
matutem: Done.

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
matutem: Reverted it.

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
matutem: Much better.

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
matutem: Good idea. Done.

hw/ip_templates/pwrmgr/dv/sva/pwrmgr_sec_cm_checker_assert.sv
matutem: Comments improved. Done.

matutem:
@rswarbrick, thanks for the thoughtful review. PTAL.

rswarbrick (approved):
Thanks for the tweaks. This looks good to me.

PR #22847, opened 2024-04-26
[i2c] Add multi-controller and broadcast command support

a-will:
Add support for...
- Bus arbitration for both the controller (multi-controller environments) and the target (broadcast commands)
- Consolidated transaction boundary checking in a bus monitor
  - Start / Stop detection
  - Bus idle timeouts (SCL and SDA continuously high)
  - Bus active timeouts (SCL continuously low)
  - Bus free status signal
  - Adjust controller and target FSMs to respond to bus monitor's outputs.
- Clock synchronization
- Simultaneously active host and target modules
- SMBus's clock extension timeout
  - Accumulate count over an entire transaction; don't reset per instance

Change fault reporting (SDA interference, SDA instability, SCL interference) to harmonize with multi-controller support.
- "SDA interference" no longer triggers from SDA being 0 at the first controller enablement. Just read the RX Oversample CSR instead.
- "SCL interference" should be ignored in multi-controller environments, as this is simply clock synchronization. Leave the interrupt masked.

Rework the functionality meant to prevent stale data in the TX FIFO from transmitting from the target. Avoid races with software via an optional mode that simply halts the FSM and stretches the clock on the first bit for every read transfer.

Fix up address matching to not trigger if the mask is completely clear.

Fix up "fault" injection timing so pulses do not get deleted by CDC instrumentation and so they don't trigger assertions.


a-will:
Whoops, forgot that the bus monitor can't be active right away at chip-level -- The pads aren't necessarily connected to the IP's pins immediately, and on earlgrey, those input muxes start with SCL and SDA low!

PR #22848, opened 2024-04-26
[util,reggen] Trigger errors for countermeasures missing in RTL

### LINKED ISSUE #10071: [reggen] Add tooling for checking in-line countermeasure annotations
As discussed in the `entropy_src` D2S review meeting today, it would be great if we could annotate the code with the countermeasure IDs. 

These would not be exact annotations, but rather serve as hints where an auditor should look in order to audit a specific countermeasure. I.e., certain countermeasures can be pinpointed exactly  - think prim like instances like `prim_count`, whereas algorithmic countermeasures like masking schemes are a bit less well localized.

The tentative format for such annotations should be `// SEC_CM [INSTANCE].ASSET.CM_TYPE` in order to facilitate checks in reggen.

In particular, reggen should:
- Extract all lines starting with `// SEC_CM` from `ipname/rtl/*.sv` files
- Match them up against the Hjson list defined, and throw an error if
  - the countermeasure annotation does not exist in the Hjson list
  - a countermeasure from the Hjson list cannot be matched up with an RTL annotation (it is allowed to use the same countermeasure ID multiple times in RTL)

Possible future enhancements:
- Instead of just grepping files under `ipname/rtl/*.sv`, fusesoc could be invoked to build the file list
- In the documentation, we can append a list of links to each countermeasure ID that takes the reviewer/auditor to the code section with the annotation (on GitHub).

matutem:
This has two commits:
- Move the countermeasure checks to topgen.py.
- Trigger the countermeasure check in CI.

Fixes #10071

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -142,6 +142,10 @@ jobs:
   steps:
   - template: ci/checkout-template.yml
   - template: ci/install-package-dependencies.yml
+  # Check that all earlgrey countermeasures are implemented.
+  - bash: ci/scripts/check-countermeasures.sh earlgrey
+    displayName: Check countermeasures for top_earlgrey
+    continueOnError: True
   # Bazel test suites are a common cause of problematic tags. Check test suites
   # before checking for other tag issues.
   - bash:  ci/scripts/check_bazel_test_suites.py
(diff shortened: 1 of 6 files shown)


matutem:
@rswarbrick you are right: I woke up this morning thinking to do the same. Done.
PTAL

rswarbrick (commented):
This looks like a good idea but I see it's failing in CI. Would you mind debugging?

matutem:
I will put this PR on hold until we deal with the pinmux failures: they are caused by this still using the legacy generation low. The simplest way to handle it is to move pinmux to ipgen.

matutem:
The check was moved to topgen, and removed from regtool, and is ready to go. I tested it by both adding a spurious countermeasure in the RTL and by adding an extra countermeasure in an hjson file, both of which result in errors being flagged.

rswarbrick (approved):
Looks good to me, and I don't believe it could have triggered the CI failures that we're seeing here.

vogelpi (approved):
This looks good, thanks for taking care of this @matutem !

PR #22852, opened 2024-04-26
[prim_sha2/hmac,rtl/dv] HMAC/SHA-2 secure wiping fix

### LINKED ISSUE #22555: [hmac/rtl] Overwrite internal values with WIPE_SECRET instead of XOR'ing with it
### Doc
The wipe secret register, in the intro of the doc (here, last paragraph: https://opentitan.org/book/hw/ip/hmac/index.html), it says that the internal variables will be reset to the written value from this register WIPE_SECRET. But the register description (https://opentitan.org/book/hw/ip/hmac/doc/registers.html#wipe_secret) is saying that a XOR is done instead. So which description should be considered?

### RTL
Within the RTL I can see that:
- hmac.sv: the secret key is replaced by the wipe_secret register value
- prim_sha2: all the internal values are XORed with the wipe_secret register value

gdessouky:
This PR fixes the secure wiping feature in `prim_sha2` engine such that internal variables are overwritten with the WIPE_SECRET value rather than XORed. This change makes it more secure and also consistent with how the secret key is wiped in HMAC. This change has ~~no impact on DV, given its current state (value overwritten to variables is not checked atm)~~ no impact on block-level DV - `hmac_smoke` and `hmac_wipe_secret` pass. A TLT has been updated in this PR to correspond with this change.

Once merged, this would close #22555.

### DIFF
--- a/hw/ip/prim/rtl/prim_sha2.sv
+++ b/hw/ip/prim/rtl/prim_sha2.sv
@@ -16,7 +16,7 @@ module prim_sha2 import prim_sha2_pkg::*;
   input rst_ni,
 
   input               wipe_secret_i,
-  input sha_word64_t  wipe_v_i,
+  input sha_word32_t  wipe_v_i,
   // control signals and message words input to the message FIFO
   input               fifo_rvalid_i, // indicates that the message FIFO (prim_sync_fifo) has words
                                      // ready to write into the SHA-2 padding buffer
@@ -66,7 +66,7 @@ module prim_sha2 import prim_sha2_pkg::*;
       assign unused_digest_upper[i] = ^digest_i[i][63:32];
     end
     logic unused_signals;
-    assign unused_signals = ^{wipe_v_i[63:32], shaf_rdata[63:32], unused_digest_upper};
+    assign unused_signals = ^{shaf_rdata[63:32], unused_digest_upper};
   end
 
   // Most operations and control signals are identical no matter if we are starting or continuing
@@ -87,7 +87,7 @@ module prim_sha2 import prim_sha2_pkg::*;
     always_comb begin : compute_w_multimode
       w_d = w_q;
       if (wipe_secret_i) begin
-        w_d = w_q ^ {16{wipe_v_i[63:0]}};
+        w_d = {32{wipe_v_i}};
       end else if (!sha_en_i || hash_go) begin
         w_d = '0;
       end else if (!run_hash && update_w_from_fifo) begin
@@ -118,9 +118,7 @@ module prim_sha2 import prim_sha2_pkg::*;
     always_comb begin : compression_multimode
       hash_d = hash_q;
       if (wipe_secret_i) begin
-        for (int i = 0; i < 8; i++) begin
-          hash_d[i] = hash_q[i] ^ wipe_v_i;
-        end
+        hash_d = {16{wipe_v_i}};
       end else if (init_hash) begin
         hash_d = digest_q;
       end else if (run_hash) begin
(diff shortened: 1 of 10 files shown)


martin-velay (approved):
Sounds good to me

hw/ip/prim/rtl/prim_sha2.sv
```diff
@@ -87,7 +87,7 @@ module prim_sha2 import prim_sha2_pkg::*;
     always_comb begin : compute_w_multimode
       w_d = w_q;
       if (wipe_secret_i) begin
-        w_d = w_q ^ {16{wipe_v_i[63:0]}};
+        w_d = {32{wipe_v_i[31:0]}};
```
andreaskurth, line 90: Indexing is no longer necessary here, right? By removing it, linting would flag an error if we were to change the width of `wipe_v_i`
```suggestion
        w_d = {32{wipe_v_i}};
```

hw/ip/prim/rtl/prim_sha2.sv
```diff
@@ -119,7 +119,7 @@ module prim_sha2 import prim_sha2_pkg::*;
       hash_d = hash_q;
       if (wipe_secret_i) begin
         for (int i = 0; i < 8; i++) begin
-          hash_d[i] = hash_q[i] ^ wipe_v_i;
+          hash_d[i] = {wipe_v_i, wipe_v_i};
         end
```
andreaskurth, line 123: We could now simplify this
```suggestion
        hash_d = {16{wipe_v_i}};
```

hw/ip/prim/rtl/prim_sha2.sv
```diff
@@ -144,7 +144,7 @@ module prim_sha2 import prim_sha2_pkg::*;
       digest_d = digest_q;
       if (wipe_secret_i) begin
         for (int i = 0 ; i < 8 ; i++) begin
-          digest_d[i] = digest_q[i] ^ wipe_v_i;
+          digest_d[i] = {wipe_v_i, wipe_v_i};
         end
```
andreaskurth, line 148: We could now simplify this
```suggestion
        digest_d = {16{wipe_v_i}};
```

hw/ip/prim/rtl/prim_sha2.sv
```diff
@@ -249,7 +249,7 @@ module prim_sha2 import prim_sha2_pkg::*;
       digest256_d = digest256_q;
       if (wipe_secret_i) begin
         for (int i = 0 ; i < 8 ; i++) begin
-          digest256_d[i] = digest256_q[i] ^ wipe_v_i[31:0];
+          digest256_d[i] = wipe_v_i;
         end
```
andreaskurth, line 253: We could now simplify this
```suggestion
        digest256_d = {8{wipe_v_i}};
```

andreaskurth (commented):
Thanks for this change, @gdessouky. It mostly LGTM, though I left a few suggestions.

I think the three commits should be squashed into one, because the documentation would be outdated after the first commit and the TLT would fail after the first and before the third commit.

hw/ip/prim/rtl/prim_sha2.sv
gdessouky: had missed removing it here, thanks for catching!

gdessouky:
@andreaskurth thanks for the suggestions to simplify; addressed them in my upcoming forced push. Additionally, I remove the secure wipe input (enable and value) from `hmac_core.sv`; they are not used there and are waived as "not currently used, but keeping them attached for future use" - is there any reason now to still keep them?

andreaskurth:
> Additionally, I remove the secure wipe input (enable and value) from `hmac_core.sv`; they are not used there and are waived as "not currently used, but keeping them attached for future use" - is there any reason now to still keep them?

As `hmac_core` only stores non-confidential values in its flops and doesn't instantiate any modules, I don't see a need for a secure wipe input into that module -- removing that input SGTM

gdessouky:
Rebased and pushed again to see if we go green again :(

gdessouky:
The tests failing again don't seem to have anything to do with the changes in this PR.

andreaskurth:
> The tests failing again don't seem to have anything to do with the changes in this PR, but checking again to confirm.

I agree that the failures don't seem to be related to the changes in this PR.

cdgori (commented):
This LGTM

PR #22861, opened 2024-04-26
[hjson,format] Re-format some hjson files per style guide

matutem:
Deals mostly with indentation and description fields.

### DIFF
--- a/hw/ip/aes/data/aes.hjson
+++ b/hw/ip/aes/data/aes.hjson
@@ -583,388 +583,388 @@
     },
 ##############################################################################
 # control and status registers
-  { name: "CTRL_SHADOWED",
-    desc: '''
-      Control Register.
+    { name: "CTRL_SHADOWED",
+      desc: '''
+        Control Register.
 
-      Can only be updated when the AES unit is idle.
-      If the AES unit is non-idle, writes to this register are ignored.
-      This register is shadowed, meaning two subsequent write operations are required to change its content.
-      If the two write operations try to set a different value, a recoverable alert is triggered (See Status Register).
-      A read operation clears the internal phase tracking: The next write operation is always considered a first write operation of an update sequence.
-      Any write operation to this register will clear the status tracking required for automatic mode (See MANUAL_OPERATION field).
-      A write to the Control Register is considered the start of a new message.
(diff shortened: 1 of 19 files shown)


moidx:
Hi @engdoreis can you review this PR? I am not sure what will be the impact on the automation script you maintain.

Hi @matutem, I think it would be good to quickly cherry pick this change into `earlgrey_es_sival` as soon as you get this merged to avoid future cherry-pick conflicts.

hw/top_earlgrey/ip/pinmux/data/autogen/pinmux.hjson
```diff
@@ -434,718 +434,739 @@
...
+    { multireg: {
+        name:     "MIO_PERIPH_INSEL_REGWEN",
+        desc:     "Register write enable for MIO peripheral input selects.",
+        count:    "NMioPeriphIn",
+        compact:  "false",
+        swaccess: "rw0c",
+        hwaccess: "none",
+        cname:    "MIO_PERIPH_INSEL",
+        fields: [
+          { bits:   "0",
```
engdoreis, line 446: Maybe this should be: 
```suggestion
        fields: [{
            bits:   "0",
```
For consistency.

hw/ip/aes/data/aes.hjson
```diff
@@ -583,388 +583,388 @@
...
+            },
+            { value: "2",
+              name: "AES_DEC",
+              desc: '''
+                2'b10: Decryption.
+              '''
+            },
+          ]
+          tags: ["shadowed_reg_path:u_aes_core.u_ctrl_reg_shadowed.u_ctrl_reg_shadowed_operation"]
+        }
```
engdoreis, line 631: Should we put a comma here to be consistent with the other lists?

engdoreis (commented):
I like the Idea of trying to standardize the formatting of the configuration files, but I still notice some inconsistencies. I believe the only way to achieve consistency and prevent formatting loss in the future due to edits is by using tooling. 
I've worked on a hjson formatting script in this [PR](https://github.com/lowRISC/opentitan/pull/21743) that could be applied here. But I also want to hear what @HU90m has to say.

PR #22862, opened 2024-04-26
[otbn, rtl] Qualify secure wipe control signals with secure wipe running

GregAC:
This introduces an extra check for all secure wipe control signals to ensure secure wipe is running.

This still has some work left to do. Smoke tests pass but I haven't run any further tests.

Currently it doesn't trigger any alert when one of the secure wipe control signals is asserted but secure wipe isn't actually running. This could be added.

Finally there's a TODO in the commit that needs attending to.

### DIFF
--- a/hw/ip/otbn/dv/uvm/sva/otbn_idle_checker.sv
+++ b/hw/ip/otbn/dv/uvm/sva/otbn_idle_checker.sv
@@ -20,11 +20,13 @@ module otbn_idle_checker
   input logic otbn_dmem_scramble_key_req_busy_i,
   input logic otbn_imem_scramble_key_req_busy_i,
 
-  input logic [7:0] status_q_i,
-  input logic busy_secure_wipe,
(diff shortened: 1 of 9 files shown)


vogelpi:
Thanks Greg for your draft PR. I've taken this as a source of inspiration for #22924. The main differences between your draft PR and my implementation are:
- I've flopped `secure_wipe_running_o` inside the start stop controller to ensure we have separate drivers for the signals we want to protect and the signal we check against. This allows reducing the code changes as we don't need to introduce prim_bufs.
- Instead of qualifying, my PR does a comparison and flags an error in case of mismatches. The error is factored into the fatal internal error signal (already factors in a different secure wipe error around spurious secure wipe ack detection).
- Also the flopped signals used for wiping GPR and WDRs with zero or URND inside otbn_core are now protected. This is also pretty critical in my view.

GregAC:
Superseded by @vogelpi's work: https://github.com/lowRISC/opentitan/pull/22924

PR #22863, opened 2024-04-26
Make check_tool_requirements compatible with newer FuseSoC

olofk:
Newer versions of FuseSoC (at least 2.4) fails to export files that start with ./ in the filename.


olofk:
Some kind of failure. A bit unclear what went wrong though. For the untrained eye, it seems unlikely to be related this PR

a-will (approved):

rswarbrick (approved):

PR #22865, opened 2024-04-27
[i2c] Add target FSM responses to timeouts and read commands

a-will:
Change target FSM's stretch timeout to count across an entire transaction, for SMBus support.

Add a CSR to report fault events that occurred during a read, and gate
the next read (stretch clock at the first bit) until the event latch is
cleared. This is for reads to this IP's target module.

The gate helps prevent stale data from emerging from the FIFO.

Also roll in some of the functionality the TXRST_ON_COND CSR was meant
to provide, and make this a catch-all gate for releasing the TX FIFO on
a read instead. The prior mechanism had some corner cases that produced
races with software, such as software being able to inadvertently write
a new (yet stale) byte of data after the clear yet before being notified
that the command completed. The separate gating CSR helps avoid such
races by not allowing software's normal write path to bypass event
handling.

### DIFF
--- a/hw/ip/i2c/data/i2c.hjson
+++ b/hw/ip/i2c/data/i2c.hjson
@@ -291,6 +291,19 @@
                 When 0, the bus monitor will report that the bus is always free, so the controller FSM is never blocked from transmitting.
                 '''
         }
+        { bits: "6"
+          resval: "0"
+          name: "TX_STRETCH_CTRL_EN"
+          desc: '''
+                If set to 1, this bit causes a read transfer addressed to this target to set the corresponding bit in !!TARGET_EVENTS.
+
+                While !!TARGET_EVENTS.TX_PENDING is 1, subsequent read transactions will stretch the clock, even if there is data in the TX FIFO.
+
+                If enabled, this function allows software to confirm the data in the TX FIFO should be released for the current read.
+                This may be useful for cases where the TX FIFO has data that does not apply to the current transfer.
+                For example, the transaction could've targeted an alternate function via another address.
+                '''
+        }
       ]
     }
     { name:     "STATUS"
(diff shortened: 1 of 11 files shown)


hw/ip/i2c/rtl/i2c_target_fsm.sv
```diff
@@ -627,7 +635,7 @@ module i2c_target_fsm import i2c_pkg::*;
   //
```
hcallahan-lowrisc, line 635: nit: stale comment (L631-L637)

hw/ip/i2c/data/i2c.hjson
```diff
@@ -291,6 +291,20 @@
                 When 0, the bus monitor will report that the bus is always free, so the controller FSM is never blocked from transmitting.
                 '''
         }
+        { bits: "6"
+          resval: "0"
+          name: "TX_STRETCH_CTRL_EN"
+          desc: '''
+                Enable setting !!TARGET_EVENTS.TX_PENDING to gate target read transfers.
```
hcallahan-lowrisc, line 298: nit: imo this first line of the description is a bit too terse, I had to think quite hard to parse it the first time reading. L300 would be a better first line for the description I think.

hcallahan-lowrisc (approved):
LGTM, just minor doc nits that can be wrapped up into another PR if desired.

hw/ip/i2c/rtl/i2c_target_fsm.sv
a-will: Added info about the new term in the comment.

hw/ip/i2c/data/i2c.hjson
a-will: Works for me. I deleted this line and made the one you selected be first.

andreaskurth (approved):
LGTM, thx @a-will!

andreaskurth:
You ran this quickly through block-level DV to ensure that existing tests don't fall apart (pass rates roughly maintained), right?

a-will:
> You did run this quickly through block-level DV to ensure that existing tests don't fall apart (pass rates roughly maintained), right?

Always. My process has been to fix DV tests first, then ensure they continue to pass--Otherwise, I make more fixes.

PR #22866, opened 2024-04-27
[hw,mbx,rtl] Move SOC_INTR_ADDR/DATA registers to the end

Razer6:
A mailbox con either be a PCIe mailbox or a FW mailbox. In case of the PCIe setting, the SOC reagisters 0x0 and 0x4 implement the capability header functionality, but this is done outside of the mbx RTL. PCIe mailboxes do not use the SOC_INTR_DATA/ADDR registers. FW-based mailboxes on the other hand only use the SOC_INTR_DATA/ADDR registers but no capability headers. Previously, it was decided to alias both registers as they are used orthogonally. However, this deceission was revised.

This PR moves the SOC_INTR_DATA/ADDR registers to the end of the SOC register interface to not alias with any capability header registers for PCIe mailboxes

### DIFF
--- a/hw/ip/mbx/data/mbx.hjson
+++ b/hw/ip/mbx/data/mbx.hjson
@@ -336,36 +336,7 @@
       }
     ]
     soc: [
-      { name: "SOC_DOE_INTR_MSG_ADDR"
-        desc: '''
-              Utilized by the mailbox responder to send an interrupt message to the requester via a write to the configured address.
-              Defined only for FW-to-FW mailboxes.
-              '''
-        swaccess: "rw"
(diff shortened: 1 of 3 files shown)


andreaskurth (approved):

andreaskurth:
CHANGE AUTHORIZED: hw/ip/mbx/data/mbx.hjson
CHANGE AUTHORIZED: hw/ip/mbx/rtl/mbx_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/mbx/rtl/mbx_soc_reg_top.sv

PR #22872, opened 2024-04-29
[i2c] Add multi-controller support

a-will:
Add support for...
- Bus arbitration for both the controller (multi-controller environments) and the target (broadcast commands)
- Clock synchronization
- Simultaneously active host and target modules

Change fault reporting (SDA interference, SDA instability, SCL interference) to harmonize with multi-controller support.
- "SDA interference" no longer triggers from SDA being 0 at the first controller enablement. Just read the RX Oversample CSR instead.
- "SCL interference" should be ignored in multi-controller environments, as this is simply clock synchronization. Leave the interrupt masked.

Add clock stretching and lost arbitration detection for control symbols. These were overlooked, and potential clock stretching applies even in single-controller environments.

Fix up "fault" injection timing so pulses do not get deleted by CDC instrumentation and so they don't trigger assertions.

Builds on https://github.com/lowRISC/opentitan/pull/22865. The last 9 commits are new, but the very last commit will likely be dropped.


hw/ip/i2c/rtl/i2c_core.sv
```diff
@@ -418,6 +449,74 @@ module i2c_core import i2c_pkg::*;
...
+      bus_event_detect_cnt <= '1;
+    end else if ((scl_fsm != scl_fsm_q) || (sda_fsm != sda_fsm_q)) begin
+      // Wait for the round-trip time on SCL changes or on SDA changes. The
+      // latter handles Start and Stop conditions, so changes while SCL is high
+      // are allowed to propagate. The rise time is used here because it
+      // should be the longer value. The (-1) term here is to account for the
+      // delay in the counter.
+      // Note that there are limits to this method of detecting arbitration.
+      // A separate, buggy device that drives clock low faster than the counter
+      // can expire would not trigger loss of arbitration.
```
a-will, line 479: There might be other options here. Perhaps it would be better to put these checks in the specific modules, instead. The check here loses some of the information we had with the particular state in the controller and target modules.

a-will:
I'm not sure what's going on with the Xcelium tests. They don't seem to actually run.

a-will:
Note that DV pass rates are maintained. The last commit is there just to show that I've tested responses to multi-controller "fault" conditions while the target module was enabled. We can delete that commit and have proper sequences for simultaneous operation.

Technically, the i2c controller module should even be able to communicate with its own sibling target module, but that hasn't been tested. ;)

a-will:
Here's how the FPGA utilization numbers look compared to before the split FSM PR:

| branch         | flops | luts |
| -------------- | ----- | ---- |
| before-split   |  1029 | 1612 |
| this-pr        |  1040 | 1715 |

a-will:
I made a guess that Xcelium poorly handles dependency tracking between always_comb blocks and created an infinite loop that prevented forward progress with simulation. Let's see if moving `log_start` and `log_stop` into the same always_comb block as `ctrl_symbol_failed` brings the Xcelium CI jobs to pass.

hw/ip/i2c/rtl/i2c_bus_monitor.sv
```diff
@@ -23,6 +23,7 @@ module i2c_bus_monitor import i2c_pkg::*;
   input                            bus_active_timeout_en_i,
   input [19:0]                     bus_inactive_timeout_i, // SCL held high (~50 us)
 
+
```
andreaskurth, line 26:
```suggestion
```
Stray empty line ;-)

hw/ip/i2c/data/i2c.hjson
```diff
@@ -133,6 +133,16 @@
...
+    { name:    "InputDelayCycles",
+      type:    "int",
+      default: "0",
+      desc:    '''
+               Maximum number of cycles of propagation delay between a change on the cio_scl_en_o or cio_sda_en_o pins and sensing the new values on the corresponding input pins, not including the rise/fall times.
+               For the purposes of this calculation, an input delay of 0 cycles means an output pin changing at the beginning of clock edge N will be sampled and observed on the input pins at clock edge N+1.
+               '''
+      local:   "false",
+      expose:  "true"
+    }
```
andreaskurth, line 145: This change should already be in the first commit, right? Because the parameter is referred to there

andreaskurth (approved):
Thx @a-will, this LGTM overall, though I must admit that I don't have time for an in-depth review.

How did you check that the implementation works (at least in the common case)?

Oh, and don't forget to remove or edit the _DELETE ME_ commit (maybe track it with a TODO item in the PR description?).

hw/ip/i2c/rtl/i2c_controller_fsm.sv
```diff
@@ -631,16 +655,28 @@ module i2c_controller_fsm import i2c_pkg::*;
...
+        if (!trans_started && !scl_i) begin
+          // This was the start of a transaction, but another device beat us to access. Go back to
+          // Idle, and wait for the next turn.
+          state_d = Idle;
+        end else if (trans_started && !scl_i && !scl_i_q && stretch_predict_cnt_expired) begin
+          // Saw stretching. Remain in this state and don't count down until we see SCL high.
+          state_d = SetupStart;
+          load_tcount = 1'b1;
+          // This double-counts the rise time, unfortunately.
+          tcount_sel = tSetupStart;
```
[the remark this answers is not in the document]
a-will: No, it's not a problem. It merely makes the module a little slower than it might need to be. I chose to keep it safe + not introduce yet another load value.

a-will:
> How did you check that the implementation works (at least in the common case)?

Mostly, I made sure that DV regression test (+ top-level test) pass rates did not drop for all the known use cases. I also fixed failures in `i2c_host_error_intr` when control symbols would get squashed due to the timing of the interference. Finally, I manually checked that the target module's FSM responded appropriately to "interference" cases when it wasn't being addressed--That last commit turns on both FSMs simultaneously.

There are still features from this PR to validate:
- The full responses to loss of arbitration and clock synchronization should be checked. The sequence would produce the same interference stimulus from `i2c_host_error_intr`, but it shouldn't reset the full IP as soon as the interrupt goes high. These are events that should not prevent the IP from continuing operation, and checking and working with `CONTROLLER_EVENTS` is the proper route.
- Loss of arbitration for the target needs to be checked. This can manifest in a way that is similar to `i2c_host_error_intr`, but with the checks for proper responses in `TARGET_EVENTS` and the ACQ FIFO.
- Controller and target sequences should all be able to pass in a multi-controller configuration (e.g. where both modules are enabled). The only one I've checked so far is `i2c_host_error_intr`.


Then from prior PRs...
- Wait for idle after init in multi-controller monitor mode
- Bus timeouts
- Target stretch timeout counter accumulation across multiple bits / bytes
- Stale data in the TX FIFO properly being blocked by software clearing the `TARGET_EVENTS.TX_PENDING` bit

andreaskurth:
SG, thx @a-will for the elaboration. CC @hcallahan-lowrisc for the test points listed above.

hw/ip/i2c/data/i2c.hjson
a-will: Sure, that works. Squashed.

PR #22873, opened 2024-04-29
[crypto] Change P-384 Pointer Handling

wettermo:
This PR removes `dptr_<x>` variables and adapts the code to function without it.

Previously, to reference a value in the data memory, the memory address of the value was stored in addition to the actual value. 
However, this is not absolutely necessary to reference values in the data memory. It is also possible to do this without extra pointer variables. 
This means that instead of storing the pointer (i.e. the memory address of the value) in the data memory, the pointer is only used in registers without writing it to the data memory. 
This reduces the number of instructions required in the code and therefore saves instruction and data memory.

### DIFF
--- a/sw/otbn/crypto/p384_base_mult.s
+++ b/sw/otbn/crypto/p384_base_mult.s
@@ -19,10 +19,8 @@
  * Sets up context and calls the internal scalar multiplication routine.
  * This routine runs in constant time.
  *
- * @param[in]       x17: dptr_d0, pointer to location in dmem containing
- *                                1st private key share d0
- * @param[in]       x19: dptr_d1, pointer to location in dmem containing
- *                                2nd private key share d1
+ * @param[in]  dmem[d0]: 1st private key share d0 in dmem
+ * @param[in]  dmem[d1]: 2nd private key share d1 in dmem
  * @param[out]  dmem[x]: x-coordinate in dmem
  * @param[out]  dmem[y]: y-coordinate in dmem
  *
@@ -32,7 +30,7 @@
  * Flags: When leaving this subroutine, the M, L and Z flags of FG0 correspond
  *        to the computed affine y-coordinate.
  *
- * clobbered registers: x2, x3, x9 to x13, x18 to x21, x26 to x30
+ * clobbered registers: x2, x3, x9 to x13, x17 to x21, x26 to x30
  *                      w0 to w30
  * clobbered flag groups: FG0
  */
@@ -50,6 +48,12 @@ p384_base_mult:
 
   /* set dmem pointer to scratchpad */
   la        x30, scratchpad
+
+  /* set dmem pointer to 1st private key share d0 */
+  la        x17, d0
+
+  /* set dmem pointer to 1st private key share d0 */
+  la        x19, d1
 
   /* load domain parameter n (order of base point)
      [w11, w10] = n = dmem[p384_n] */
@@ -94,6 +98,18 @@ p384_base_mult:
 .section .data
 
 .balign 32
+
+/* 1st private key share d0 */
+.globl d0
+.weak d0
+d0:
+  .zero 64
+
+/* 2nd private key share d1 */
+.globl d1
+.weak d1
+d1:
+  .zero 64
 
 /* buffer for x-coordinate */
 .globl x
--- a/sw/otbn/crypto/p384_base_mult.s
+++ b/sw/otbn/crypto/p384_base_mult.s
@@ -19,12 +19,12 @@
  * Sets up context and calls the internal scalar multiplication routine.
  * This routine runs in constant time.
  *
- * @param[in]  dmem[0]: dptr_d0, pointer to location in dmem containing
- *                      1st private key share d0
- * @param[in]  dmem[4]: dptr_d1, pointer to location in dmem containing
- *                      2nd private key share d1
- * @param[in]  dmem[20]: dptr_x, pointer to result buffer for x-coordinate
- * @param[in]  dmem[24]: dptr_y, pointer to result buffer for y-coordinate
+ * @param[in]       x17: dptr_d0, pointer to location in dmem containing
+ *                                1st private key share d0
+ * @param[in]       x19: dptr_d1, pointer to location in dmem containing
+ *                                2nd private key share d1
+ * @param[out]  dmem[x]: x-coordinate in dmem
+ * @param[out]  dmem[y]: y-coordinate in dmem
  *
  * 384-bit quantities have to be provided in dmem in little-endian format,
  * 512 bit aligned, with the highest 128 bit set to zero.
@@ -44,14 +44,6 @@ p384_base_mult:
 
   /* set dmem pointer to y-coordinate of base point */
   la        x21, p384_gy
-
-  /* set dmem pointer to 1st scalar share d0 */
-  la        x17, dptr_d0
-  lw        x17, 0(x17)
-
-  /* set dmem pointer to 2nd scalar share d1 */
-  la        x19, dptr_d1
-  lw        x19, 0(x19)
 
   /* set dmem pointer to domain parameter b */
   la        x28, p384_b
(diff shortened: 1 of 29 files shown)


sw/otbn/crypto/p384_curve_point_valid.s
```diff
@@ -26,25 +26,17 @@ start:
   unimp
 
 validate_point:
+  /* Fill gpp registers with pointers to coordinates */
```
jadephilipoom, line 29: nit: I think this is a typo?
```suggestion
  /* Fill gpr registers with pointers to coordinates */
```

There are a few more repetitions of the same comment below.

sw/otbn/crypto/p384_base_mult.s
```diff
@@ -19,12 +19,12 @@
...
- * @param[in]  dmem[0]: dptr_d0, pointer to location in dmem containing
- *                      1st private key share d0
- * @param[in]  dmem[4]: dptr_d1, pointer to location in dmem containing
- *                      2nd private key share d1
- * @param[in]  dmem[20]: dptr_x, pointer to result buffer for x-coordinate
- * @param[in]  dmem[24]: dptr_y, pointer to result buffer for y-coordinate
+ * @param[in]       x17: dptr_d0, pointer to location in dmem containing
+ *                                1st private key share d0
+ * @param[in]       x19: dptr_d1, pointer to location in dmem containing
+ *                                2nd private key share d1
```
jadephilipoom, line 25: nit: I'd maybe just change the code below to load the addresses from `dmem[d0]` and `dmem[d1]` (that is, put the `la` instructions here instead of in the wrapper code).

sw/otbn/crypto/p384_isoncurve.s
```diff
@@ -145,14 +131,10 @@ p384_isoncurve:
...
- * @param[in]  dmem[16]: dptr_lhs, pointer to dmem location where left hand
- *                               side result lhs will be stored
- * @param[in]  dmem[20]: dptr_x, pointer to dmem location containing affine
- *                               x-coordinate of input point
- * @param[in]  dmem[24]: dptr_y, pointer to dmem location containing affine
- *                               y-coordinate of input point
+ * @param[in]  x20: dptr_x, pointer to dmem location containing affine
+ *                          x-coordinate of input point
+ * @param[in]  x21: dptr_y, pointer to dmem location containing affine
+ *                          y-coordinate of input point
```
jadephilipoom, line 137: Same note here about the pointers; I'd put the `la` instructions inside the subroutine and use `dmem[x]` and `dmem[y]` directly here.

sw/otbn/crypto/p384_keygen.s
```diff
@@ -151,10 +151,8 @@ p384_random_scalar:
  *
  * Flags: Flags have no meaning beyond the scope of this subroutine.
  *
- * @param[in]  dmem[0]: dptr_d0, pointer to location in dmem containing
- *                      1st private key share d0
- * @param[in]  dmem[4]: dptr_d1, pointer to location in dmem containing
- *                      2nd private key share d1
+ * @param[in]  x20: dptr_d0, pointer to bufffer of 1st private key share d0
+ * @param[in]  x21: dptr_d1, pointer to bufffer of 2nd private key share d1
```
jadephilipoom, line 155: Same note about indirect pointers here.

sw/otbn/crypto/p384_keygen.s
```diff
@@ -191,10 +185,8 @@ p384_generate_random_key:
  *
  * Flags: Flags have no meaning beyond the scope of this subroutine.
  *
- * @param[in]  dmem[0]: dptr_k0, pointer to location in dmem containing
- *                      1st scalar share k0
- * @param[in]  dmem[4]: dptr_k1, pointer to location in dmem containing
- *                      2nd scalar share k1
+ * @param[in]  x20: dptr_k0, pointer to bufffer of 1st scalar share k0
+ * @param[in]  x21: dptr_k1, pointer to bufffer of 2nd scalar share k1
```
jadephilipoom, line 189: Same note about indirect pointers here.

sw/otbn/crypto/p384_scalar_mult.s
```diff
@@ -19,12 +19,12 @@
...
- * @param[in]   dmem[4]: dptr_k1, pointer to location in dmem containing
- *                       2nd scalar share k1
- * @param[in]  dmem[20]: dptr_x, pointer to affine x-coordinate in dmem
- * @param[in]  dmem[22]: dptr_y, pointer to affine y-coordinate in dmem
+ * @param[in]   x20:     dptr_x, pointer to affine x-coordinate in dmem
+ * @param[in]   x21:     dptr_y, pointer to affine y-coordinate in dmem
+ * @param[in]   x17:     dptr_k0, pointer to location in dmem containing
+ *                                1st scalar share k0
+ * @param[in]   x19:     dptr_k1, pointer to location in dmem containing
+ *                                2nd scalar share k1
```
jadephilipoom, line 27: Same note about indirect pointers here.

sw/otbn/crypto/p384_sign.s
```diff
@@ -22,19 +22,19 @@
...
+ *                            2nd scalar share k1
+ * @param[in]   x6:  dptr_msg, pointer to the message to be signed in dmem
+ * @param[in]  x14:  dptr_r, pointer to dmem location where s component
+ *                           of signature will be placed
+ * @param[in]  x15:  dptr_s, pointer to dmem location where r component
+ *                           of signature will be placed
+ * @param[in]   x4:  dptr_d0, pointer to location in dmem containing
+ *                            1st private key share d0
+ * @param[in]   x5:  dptr_d1, pointer to location in dmem containing
+ *                            2nd private key share d1
```
jadephilipoom, line 37: Same note about indirect pointers here.

sw/otbn/crypto/tests/p384_curve_point_valid_test.s
```diff
@@ -3,54 +3,28 @@
...
- * the curve point are contained in the .data section below.
- * The x coordinate of the resulting curve point is masked arithmetically
- * with a random value. As the x coorodinate represents the actual
- * shared key, the x coordinate and its mask are then converted from an
- * arithmetic to a boolean masking scheme.
- *
- * The result of boolean unmasking is then compared with the expected shared
- * key value.
+ * Runs the P-384 curve point test to check whether a point (given in affine
+ * space) is a valid P-384 curve point.
```
jadephilipoom, line 9: Thanks for catching and fixing the misleading comment! :+1:

sw/otbn/crypto/tests/p384_ecdsa_verify_test.s
```diff
@@ -135,6 +116,6 @@ pub_y:
   .word 0xaaafcad2
   .zero 16
 
-/* signature verification result x_res */
-sig_xres:
+/* signature verification result x_res (rnd) */
+rnd:
   .zero 64
```
jadephilipoom, line 121: This bit is a little confusing to me -- why is the signature verification result called `rnd`? When I search for it in the code, it only ever seems to be used for the returned x-coordinate. I'd call it maybe `x_r`, to match the p256 code.

jadephilipoom (approved):
Thanks for the refactor, Moritz! Nice to see all this unneeded indirection go away :slightly_smiling_face:

sw/otbn/crypto/p384_base_mult.s
wettermo: You're right, makes it more clear. Thx!!

sw/otbn/crypto/tests/p384_ecdsa_verify_test.s
wettermo: I don't know why it's called `rnd`. I just stuck to the naming in `p384_verify.s`, where it's also named `rnd`. This was already there before I started working on the code. But tbh it confused me too. Maybe it's the right moment now to change this to `x_r` throughout the code.

wettermo:
@jadephilipoom Thanks for the review Jade! I changed everything you commented on. As soon as the CI completes, from my perspective, this PR should be mergable.

PR #22874, opened 2024-04-29
[doc] Add SCA guidelines for OTBN from COCO-Alma work.

jadephilipoom:
This text is from an OpenTitan-domain [doc](https://docs.google.com/document/d/1kX2fr5lH9xzYcpu2ycsac4JZ2j8L0Vl_0UNmoN3-UlY/edit) that @abdullahvarici wrote a while back, but it seems useful to have on the public documentation website as well.

### DIFF
--- a/doc/contributing/style_guides/otbn_style_guide.md
+++ b/doc/contributing/style_guides/otbn_style_guide.md
@@ -146,6 +146,168 @@ Prefer `.word` over alternatives such as `.quad`.
 
 The following guidelines address cryptography-specific concerns for OTBN assembly.
 
+### Handling secret shares
+
+The following guidelines were determined with the [help of the COCO-Alma tool](https://github.com/lowRISC/opentitan/tree/master/hw/ip/otbn/pre_sca/alma) and help protect against SCA attacks.
+
+**1. Do not overwrite a share with its counterpart share.**
+
+A register or memory location which contains one share must not be overwritten with its counterpart. The following lines would each violate this rule and cause transient leakage:
+```armasm
+/* Leakage (if w0, w1 contain two shares of a secret) */
+bn.and  w1, w0, w0
+bn.addi w1, w0, 0x0
+bn.mov  w1, w0
+bn.sel  w1, w0, w0, FG0.C
+```
+
+**2. Do not use multiple shares in a single instruction without masking.**
+
+Shares may interact with each other in the ALU and cause stable leakage at a gate or a register.
+The following lines would each violate this rule and cause stable leakage:
+```armasm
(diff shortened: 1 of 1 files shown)


nasahlpa (approved):
Thanks, makes sense to add this to the documentation!

doc/contributing/style_guides/otbn_style_guide.md
```diff
@@ -146,6 +146,168 @@ Prefer `.word` over alternatives such as `.quad`.
...
+bn.xor w4, w0, w0
+bn.xor w5, w1, w1
+
+/* No leakage */
+bn.xor w4, w0, w0
+bn.xor w9, w9, w9 /* dummy instruction */
+bn.xor w5, w1, w1
+```
+
+**4. Do not have shares even in different bits of a register.**
```
rswarbrick, line 191: Nit: I slightly struggled to understand this. Do we mean "Don't combine shares in a register, even in different bits" ?

rswarbrick (approved):
Eek, these look like hard guidelines to follow.

I wonder whether it would make sense to add some checking tool that could look at a trace and figure out whether the guidelines are all being followed. Practically speaking, we'd need to be able to tell it things like "w0 and w1 are two shares of a secret". But I don't think we'd need much more, so maybe this is a plausible tool to build?

doc/contributing/style_guides/otbn_style_guide.md
jadephilipoom: Yes, that's my understanding.

jadephilipoom:
> I wonder whether it would make sense to add some checking tool that could look at a trace and figure out whether the guidelines are all being followed. Practically speaking, we'd need to be able to tell it things like "w0 and w1 are two shares of a secret". But I don't think we'd need much more, so maybe this is a plausible tool to build?

I agree! It would be good to be able to check these automatically. I think we'd need a bit more than just marking the initial share registers -- we'd have to track the information flow a little bit through the program to make sure no values that _flow from_ one share are combined with any that flow from the other share. As a short-term solution, we could indeed potentially program a quick and dirty Python checker to look for violations of these specific rules, since there's already information-flow logic and similar checks in Python for OTBN. Fundamentally it's not _too_ different of a problem than the constant-time checker we already have, it just needs to check two threads of "secret" instead of one and have different failure conditions.

Longer-term though, I think it would be really nice to have a formalized contract between OTBN hardware and software that specifies which values' interactions might be visible through side channels, so that hardware and software can both be tested and verified against the same concrete set of statements -- especially relevant as OTBN evolves, since the contract might change between versions. Curious what you think of that @rswarbrick ?

PR #22875, opened 2024-04-29
[topgen] Support for multiple rv_dm instances

Razer6:
There might be multiple rv_dm instances in the system. Therefore, use the module property to determine there is no DIF for rv_dm modules rather than the name

### DIFF
--- a/util/topgen/c_test.py
+++ b/util/topgen/c_test.py
@@ -125,8 +125,7 @@ class TopGenCTest(TopGenC):
             if inst_name not in self.top["alert_module"]:
                 continue
 
-            # RV_DM module does not have DIF.
-            if inst_name == "rv_dm":
+            if not entry['generate_dif']:
                 continue
 
             for item in self.top['module']:
--- a/hw/top_darjeeling/data/autogen/top_darjeeling.gen.hjson
+++ b/hw/top_darjeeling/data/autogen/top_darjeeling.gen.hjson
@@ -529,6 +529,7 @@
           hart: 0x30010000
         }
       }
+      generate_dif: true
     }
     {
       name: gpio
@@ -627,6 +628,7 @@
           hart: 0x30000000
         }
       }
+      generate_dif: true
     }
     {
       name: spi_device
@@ -727,6 +729,7 @@
           hart: 0x30310000
         }
       }
+      generate_dif: true
     }
     {
       name: i2c0
@@ -791,6 +794,7 @@
           hart: 0x30080000
         }
       }
+      generate_dif: true
     }
     {
       name: rv_timer
@@ -841,6 +845,7 @@
           hart: 0x30100000
         }
       }
+      generate_dif: true
     }
     {
       name: otp_ctrl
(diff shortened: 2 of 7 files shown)


rswarbrick (approved):
Looks sensible to me, and presumably the fact that the generated code hasn't changed is a good sign that the change works :-)

andreaskurth (approved):
Agree with @rswarbrick; thx @Razer6

Razer6:
@rswarbrick @andreaskurth Thanks for looking into this. I had to update the condition to also include `rv_dm_v1`, which is used in Earlgrey.

util/topgen/c_test.py
```diff
@@ -126,7 +126,7 @@ def _get_alert_peripherals(self):
                 continue
 
             # RV_DM module does not have DIF.
-            if inst_name == "rv_dm":
+            if entry['type'] == "rv_dm" or entry['type'] == "rv_dm_v1":
```
andreaskurth, line 129: Hm, is there a cleaner solution than to hard-code the version suffix here? Maybe
```suggestion
            if entry['type'].startswith("rv_dm")
```
?

a-will: A "cleaner solution" would probably add a property to the hjson files and have that control what topgen generates, instead of hard-coding exceptions with specific named IP types :wink:

andreaskurth: Hehe, agreed. @Razer6 would you have time to look into this?

Razer6: I thought about the .startswith() but decided against it to be explicit.

What about adding a new property .generate_dif, set that to true of not specified, and set that to false for the RV_DM instances?

andreaskurth: SGTM!

Razer6:
@a-will @andreaskurth Could you take another look?

util/topgen/c_test.py
a-will: So now it'd be this, right?

```suggestion
            if entry['generate_dif']:
```

Razer6: Oh no. I missed saving the file. I think it is:

```python
if not entry['generate_dif']:
    continue
```

Razer6: There is still some diff generated, which I need to investigate. Will be a bit async since I'm on vacation this week.

Razer6: Seems the valie from the hjson is not picked as a bool but as a string.

Razer6:
@a-will @andreaskurth Finally CI is green. Can you take another look now?

a-will:
CHANGE AUTHORIZED: hw/top_darjeeling/data/top_darjeeling.hjson

a-will (approved):
LGTM

I feel like the key could probably be named for something structural in the top-level design, rather than for a specific output to generate, but I'm having writer's block and can't think of a good example. This is probably good enough for now :smile:

andreaskurth:
CHANGE AUTHORIZED: hw/top_darjeeling/data/top_darjeeling.hjson

PR #22882, opened 2024-04-29
[dv] fix chip_sw_flash_ctrl_write_clear bazel target

nbdd0121:
The test was moved from sw/device/tests/sim_dv to sw/device/tests and the hjson hasn't been updated.


rswarbrick (approved):
Sorry for the slow review! This looks sensible to me.

Issue #16499, opened 2022-11-21
[csrng] Investigate reseed interval


marnovandermaas:
Together with @ctopal, we were trying to figure out when the bits of `hw_exc_sts` are set, and the same for `sw_cmd_sts.cmd_sts`. Neither of which are getting coverage.

@ctopal traced the chain of assignments as `ctr_drbg_gen_ccmd_o <- rcstage_ccmd <- sfifo_bencack_ccmd <- bencack_ccmd_modified` through a number of FIFOs. The following line could lead to an error status:
https://github.com/lowRISC/opentitan/blob/8ef845b4d5a7d90fb5acb394869e9bfdb49e5cd9/hw/ip/csrng/rtl/csrng_ctr_drbg_gen.sv#L461

However, we think that if `block_encrypt_ccmd_i != GENB` then we never push to the FIFO that will eventually lead to an error status being set.

More worryingly, currently, I can find a place where the reseed counter is incremented:
https://github.com/lowRISC/opentitan/blob/8ef845b4d5a7d90fb5acb394869e9bfdb49e5cd9/hw/ip/csrng/rtl/csrng_ctr_drbg_gen.sv#L561
However, I cannot find the place where the overflow is checked to make sure that we do not exceed the reseed interval.
We need to make sure this check exists otherwise this poses a security risk.

> estimate 4

marnovandermaas:
I've tracked the assignments and uses of the reseed value, but so far have not found an overflow detection.
This is the assignment loop I have currently found:
`csrng_ctr_drbg_cmd::sfifo_keyvrc -> csrng_ctr_drbg_gen::(sfifo_genreq -> sfifo_adstage -> sfifo_rcstage -> sfifo_genbits) -> csrng_state_db::state_db_rc -> csrng_ctr_cmd::(sfifo_cmdreq -> sfifo_rcstage -> sfifo_keyvrc)`
Note: the first and last FIFOs are the same completing the loop.

As far as I can tell this is a pass through except for a zero assignment:
https://github.com/lowRISC/opentitan/blob/21db8a1708191d19ffa2510e9d055c0a9bc7417b/hw/ip/csrng/rtl/csrng_ctr_drbg_cmd.sv#L301
And a plus 1:
https://github.com/lowRISC/opentitan/blob/8ef845b4d5a7d90fb5acb394869e9bfdb49e5cd9/hw/ip/csrng/rtl/csrng_ctr_drbg_gen.sv#L561

[...]

PR #22883, opened 2024-04-29
[csrng/rtl] add reseed intervall sts err

h-filali:
Please see the commit messages for info.

Resolves [#16499](https://github.com/lowRISC/opentitan/issues/16499)

I still need to do some cleanup and debug the last part of the alert test vseq.
There seems to be an issue in DV where the wait for an acknowledgement doesn't behave as 
expected.
I also need to rebase.
The first commit can be ignored since it is already merged.

### DIFF
--- a/hw/ip/csrng/data/csrng.hjson
+++ b/hw/ip/csrng/data/csrng.hjson
@@ -317,6 +317,29 @@
         }
       ]
     },
+    { name: "RESEED_INTERVAL",
+      desc: "CSRNG maximum number of generate requests allowed between reseeds register",
+      swaccess: "rw",
+      hwaccess: "hro",
+      hwqe: "true",
+      fields: [
+        { bits: "31:0",
+          name: "RESEED_INTERVAL",
+          desc: '''
+                Setting this field will set the number of generate requests that can be
+                made to CSRNG before a reseed request needs to be made.
+                This register supports a maximum of 2^32 requests between reseeds.
+                This register will be compared to a counter, which counts the number of
+                generate commands between reseed or instantiate commands.
+                If the counter reaches the value of this register, the violating command
+                will be acknowledged with a status error.
+                If the violating command was issued by a HW instance, an interrupt will
+                be triggered.
+                '''
+          resval: 0xffff_ffff
+        }
+      ]
+    },
     {
       name: "SW_CMD_STS",
       desc: "Application interface command status register",
@@ -346,7 +369,7 @@
                 '''
           resval: "0"
         }
-        { bits: "4:3",
+        { bits: "5:3",
           name: "CMD_STS",
           desc: '''
                 This field represents the status code returned with the application command ack.
@@ -364,6 +387,8 @@
                 0x3: This error indicates that the last command was issued out of sequence.
                      This happens when a command other than instantiate was issued without sending an instantiate command first.
                      This can also happen when an uninstantiate command is sent without instantiating first.
+                0x4: This error indicates that the number of generate commands between reseeds exceeded the maximum number allowed.
+                     This happens only for generate commands.
                 '''
           resval: "0"
         }
@@ -538,6 +563,15 @@
                 instantiated or when any command other than instantiate is sent for a state that
                 wasn't instantiated yet.
                 The invalid command is ignored and CSRNG continues to operate.
+                Writing a zero resets this status bit.
+                '''
+        }
+        { bits: "15",
+          name: "CMD_STAGE_RESEED_CNT_ALERT",
+          desc: '''
+                This bit is set when the maximum number of generate requests between reseeds is
+                exceeded.
+                The invalid generate command is ignored and CSRNG continues to operate.
                 Writing a zero resets this status bit.
                 '''
         }
--- a/hw/ip/csrng/dv/env/csrng_scoreboard.sv
+++ b/hw/ip/csrng/dv/env/csrng_scoreboard.sv
@@ -266,6 +266,8 @@ class csrng_scoreboard extends cip_base_scoreboard #(
           end
         end
       end
+      "reseed_interval": begin
+      end
       "sw_cmd_sts": begin
         do_read_check = 1'b0;
       end
@@ -447,7 +449,7 @@ class csrng_scoreboard extends cip_base_scoreboard #(
     cfg.key[app] = 'h0;
     cfg.v[app]   = 'h0;
     ctr_drbg_update(app, seed_material);
-    cfg.reseed_counter[app] = 1'b1;
+    cfg.reseed_counter[app] = 1'b0;
     cfg.compliance[app]     = fips;
     cfg.status[app]         = 1'b1;
     cov_vif.cg_csrng_state_db_sample(cfg.compliance[app], compliance_previous, app);
@@ -464,7 +466,7 @@ class csrng_scoreboard extends cip_base_scoreboard #(
     `uvm_info(`gfn, $sformatf("Reseed of app %0d", app), UVM_MEDIUM)
     seed_material = entropy_input ^ additional_input;
     ctr_drbg_update(app, seed_material);
-    cfg.reseed_counter[app] = 1'b1;
+    cfg.reseed_counter[app] = 1'b0;
     cfg.compliance[app]     = fips;
     cov_vif.cg_csrng_state_db_sample(cfg.compliance[app], compliance_previous, app);
   endfunction
--- a/hw/ip/csrng/rtl/csrng_cmd_stage.sv
+++ b/hw/ip/csrng/rtl/csrng_cmd_stage.sv
@@ -18,6 +18,9 @@ module csrng_cmd_stage import csrng_pkg::*; #(
   input logic [StateId-1:0]          cmd_stage_shid_i,
   input logic [CmdFifoWidth-1:0]     cmd_stage_bus_i,
   output logic                       cmd_stage_rdy_o,
+  // Command checking interface.
+  input logic                        reseed_cnt_reached_i,
+  output logic                       reseed_cnt_alert_o,
   // Command to arbiter.
   output logic                       cmd_arb_req_o,
   output logic                       cmd_arb_sop_o,
@@ -72,16 +75,19 @@ module csrng_cmd_stage import csrng_pkg::*; #(
   logic                        sfifo_genbits_not_empty;
 
   // Command signals.
-  logic [3:0]              cmd_len;
-  logic                    cmd_fifo_zero;
-  logic                    cmd_fifo_pop;
-  logic                    cmd_len_dec;
-  logic                    cmd_gen_cnt_dec;
-  logic                    cmd_gen_1st_req;
-  logic                    cmd_gen_inc_req;
-  logic                    cmd_gen_cnt_last;
-  logic                    cmd_final_ack;
+  logic [3:0]                  cmd_len;
+  logic                        cmd_fifo_zero;
+  logic                        cmd_fifo_pop;
+  logic                        cmd_len_dec;
+  logic                        cmd_gen_cnt_dec;
+  logic                        cmd_gen_1st_req;
+  logic                        cmd_gen_inc_req;
+  logic                        cmd_gen_cnt_last;
+  logic                        cmd_final_ack;
+  logic                        cmd_err_ack;
   logic [GenBitsCntrWidth-1:0] cmd_gen_cnt;
+  csrng_cmd_sts_e              err_sts;
+  logic                        reseed_cnt_exceeded;
 
   // Flops.
   logic                    cmd_ack_q, cmd_ack_d;
@@ -255,6 +261,8 @@ module csrng_cmd_stage import csrng_pkg::*; #(
     cmd_arb_mop_o = 1'b0;
     cmd_arb_eop_o = 1'b0;
     cmd_stage_sm_err_o = 1'b0;
+    cmd_err_ack = 1'b0;
+    reseed_cnt_exceeded = 1'b0;
 
     if (state_q == Error) begin
       // In case we are in the Error state we must ignore the local escalate and enable signals.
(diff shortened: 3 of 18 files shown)


h-filali:
@vogelpi I rebased now. I think this can be reviewed now, although I still get some failing tests. I'm working on resolving the failures.
I also removed part of the DV that doesn't need to be reviewed yet.

hw/ip/csrng/data/csrng.hjson
```diff
@@ -317,6 +317,29 @@
...
+      fields: [
+        { bits: "31:0",
+          name: "RESEED_INTERVAL",
+          desc: '''
+                Setting this field will set the number of generate requests that can be
+                made to CSRNG before a reseed request needs to be made.
+                This register supports a maximum of 2^32 requests between reseeds.
+                This register will be compared to a counter, which counts the number of
+                generate commands between reseed or instantiate commands.
+                If the counter reaches the value of this register the violating command
```
vogelpi, line 334: Nit: comma after "register"

hw/ip/csrng/data/csrng.hjson
```diff
@@ -364,6 +387,8 @@
                 0x3: This error indicates that the last command was issued out of sequence.
                      This happens when a command other than instantiate was issued without sending an instantiate command first.
                      This can also happen when an uninstantiate command is sent without instantiating first.
+                0x5: This error indicates that the number of generate commands between reseeds exceeded the maximum number allowed.
+                     This happens only for generate commands.
```
vogelpi, line 391: 0x5 -> 0x4?

hw/ip/csrng/dv/env/csrng_scoreboard.sv
```diff
@@ -266,6 +266,8 @@ class csrng_scoreboard extends cip_base_scoreboard #(
           end
         end
       end
+      "reseed_interval": begin
+      end
```
vogelpi, line 270: Should we do a read check here as well? Or if not, explain why we don't do it?

hw/ip/edn/doc/registers.md
```diff
@@ -3,25 +3,26 @@
...
-| csrng.[`INT_STATE_NUM`](#int_state_num)     | 0x28     |        4 | Internal state number register                         |
-| csrng.[`INT_STATE_VAL`](#int_state_val)     | 0x2c     |        4 | Internal state read access register                    |
-| csrng.[`HW_EXC_STS`](#hw_exc_sts)           | 0x30     |        4 | Hardware instance exception status register            |
-| csrng.[`RECOV_ALERT_STS`](#recov_alert_sts) | 0x34     |        4 | Recoverable alert status register                      |
-| csrng.[`ERR_CODE`](#err_code)               | 0x38     |        4 | Hardware detection of error conditions status register |
-| csrng.[`ERR_CODE_TEST`](#err_code_test)     | 0x3c     |        4 | Test error conditions register                         |
-| csrng.[`MAIN_SM_STATE`](#main_sm_state)     | 0x40     |        4 | Main state machine state debug register                |
+| Name                                        | Offset   |   Length | Description                                                                |
+|:--------------------------------------------|:---------|---------:|:---------------------------------------------------------------------------|
```
vogelpi, line 7: This is just the auto-generated `registers.md`, you also have to commit the hjson.

vogelpi: Actually, this is the CSRNG registers.md which we now have in EDN? This doesn't make sense at all. It's also like that on master, got introduced with 903b319.

I've now filed #22899 to revert this.

hw/ip/csrng/rtl/csrng_cmd_stage.sv
```diff
@@ -310,12 +325,15 @@ module csrng_cmd_stage import csrng_pkg::*; #(
...
-            cmd_gen_cnt_dec= 1'b1;
+            cmd_gen_cnt_dec = 1'b1;
           end
           state_d = CmdAck;
         end
         CmdAck: begin
-          if (cmd_ack_i) begin
+          if (cmd_sts_err_q) begin
+            state_d = Idle;
+            cmd_sts_err_release = 1'b1;
```
vogelpi, line 335: I think you should only release the error when getting the ack. Right now you drop the error upon entering the CmdAck state. What's worse, cmd_sts_err_release doesn't seem to have an effect other than delaying the branch checking `cmd_ack_i` by one clock cycle. Does this make sense?

hw/ip/csrng/rtl/csrng_ctr_drbg_cmd.sv
```diff
@@ -213,8 +213,8 @@ module csrng_ctr_drbg_cmd import csrng_pkg::*; #(
          '0;
 
   assign prep_rc =
-         (cmdreq_ccmd == INS) ? {{(CtrLen-1){1'b0}},1'b1} :
-         (cmdreq_ccmd == RES) ? {{(CtrLen-1){1'b0}},1'b1} :
+         (cmdreq_ccmd == INS) ? {{(CtrLen-1){1'b0}},1'b0} :
+         (cmdreq_ccmd == RES) ? {{(CtrLen-1){1'b0}},1'b0} :
```
vogelpi, line 217: ?

hw/ip/csrng/rtl/csrng_pkg.sv
```diff
@@ -28,6 +28,7 @@ package csrng_pkg;
     CMD_STS_INVALID_ACMD         = 'h1,
     CMD_STS_INVALID_GEN_CMD      = 'h2,
     CMD_STS_INVALID_CMD_SEQ      = 'h3,
+    CMD_STS_RESEED_CNT_OVERFLOW  = 'h4,
```
vogelpi, line 31: The way I understand the implementation is that this doesn't flag an overflow but just reaching the limit defined by software.

vogelpi (changes requested):
Thanks for working on this @h-filali . There are a couple of issues in this PR I would like to discuss with you 1 by 1. Let's set up a meeting.

hw/ip/csrng/data/csrng.hjson
h-filali: You are correct. This happened during the rebase, I already fixed it locally.

hw/ip/csrng/data/csrng.hjson
h-filali: Thanks!

hw/ip/csrng/dv/env/csrng_scoreboard.sv
h-filali: I think it should do the read checks by default. If you don't want to do read checks you have to set `do_read_check = 1'b0;`

vogelpi: Oh god, yes you're totally right! Sorry for the noise.

hw/ip/csrng/rtl/csrng_pkg.sv
h-filali: You're right, the naming wasn't the greatest. I changed it now.

hw/ip/csrng/rtl/csrng_cmd_stage.sv
h-filali: I changed it such that the check now happens directly in the cmd_stage. I think it should be better now.

hw/ip/edn/doc/registers.md
h-filali: Rebased, thanks a lot!

hw/ip/csrng/rtl/csrng_ctr_drbg_cmd.sv
vogelpi: Yes, I agree that this makes sense.

hw/ip/csrng/rtl/csrng_core.sv
```diff
@@ -868,6 +876,12 @@ module csrng_core import csrng_pkg::*; #(
       .cmd_stage_sm_err_o           (cmd_stage_sm_err[ai])
     );
 
+    // Set reseed_cnt_reached_d to true if the max number of generate requests between reseeds
+    // has been reached for the respective counter.
+    assign reseed_cnt_reached_d[ai] = (shid_q == ai) ?
+        (state_db_rd_rc == reg2hw.reseed_interval.q) :
+        reseed_cnt_reached_q[ai];
+
```
vogelpi, line 884: I fear this is not going to work properly when more than 1 instance is busy. Because:
- `shid_q` always updates 1 cycle after `acmd_sop` which itself is high 1 cycle after the command stage of interest has won arbitration.
- But inside the command stage we check `reseed_cnt_reached` before initiating arbitration.

As a result, the command stage might use `reseed_cnt_reached` of the wrong instance.

To solve this, I see the following two options:
1) Feed the reseed counter values for all instances out of the state database to compare them to the limit "live" inside the command stage. This means you need 3 comparators instead of  1, but no muxes and no FFs. It's easy to do and will help us to with upcoming changes for M4.
2) The check is performed upon writing to the state database instead. This has the same area requirements but there is a risk of the comparison getting out of sync when SW updates the reseed limit in the meantime.

I would go for Option 1.

hw/ip/csrng/rtl/csrng_cmd_stage.sv
```diff
@@ -433,13 +449,18 @@ module csrng_cmd_stage import csrng_pkg::*; #(
 
   assign cmd_stage_ack_o = cmd_ack_q;
```
vogelpi, line 450: What's missing here is to also set `cmd_stage_ack_o` in case of `cmd_err_ack`. Right now, just the status is set but the ack is not going out. As a result CSRNG just sucks the command without doing anything. The app interface dies.

vogelpi (commented):
Thanks for the update @h-filali . This is going in the right direction but it's not yet working as expected. Please see my comments for details.

hw/ip/csrng/rtl/csrng_core.sv
h-filali: There are 3 different reseed count reached values. One for each endpoint. They all start off at 0 and for them to be increased, a generate command must have been sent at some point, which means `shid_q == ai` must have been true at some point and the value must have been updated. I probably should have added a comment explaining this or maybe I'm missing something here. What do you think @vogelpi

hw/ip/csrng/rtl/csrng_cmd_stage.sv
h-filali: Ah yes totally missed that one. Let me fix that.

h-filali:
@vogelpi Thanks for reviewing, I addressed your points. Hopefully this works now

hw/ip/csrng/rtl/csrng_core.sv
vogelpi: Hm, yes thanks for clarifying. However, there is still an issue that the comparison can be out of date. What can happen:

1. The last Generate command ends and the state database is updated.
2. The next Generate is received, the command stage checks `reseed_cnt_reached_q`. It's showing value from when that reseed counter was last read, not when it was written.
3. It's below the limit, so it's requesting arbitration.
4. It wins arbitration, `shid_q` is updated and we read the state database and update `reseed_cnt_reached_q`.
5. Now it reached the limit but the Generate is already in the pipeline and we will ack it at the end.

So we are doing one Generate more than we should.

h-filali: I think an example for when the state is read, is for the error response or the acknowledgement. The command stage needs to read from the state db every time a command is acked, so the reseed_cnt_reached_d variables will be updated any time this happens. For a second generate to be processed, the previous one has to have read from state db first.

h-filali: I'm now also convinced, that doing it on state_db writes instead of state_db reads is safer. I now changed it to writes.

vogelpi: Thanks @h-filali !

vogelpi (approved):
This looks now good, thanks @h-filali ! Let's see what CI has to say to this :-)

hw/ip/csrng/rtl/csrng_core.sv
```diff
@@ -868,6 +876,13 @@ module csrng_core import csrng_pkg::*; #(
       .cmd_stage_sm_err_o           (cmd_stage_sm_err[ai])
     );
 
+    // Set reseed_cnt_reached_d to true if the max number of generate requests between reseeds
+    // has been reached for the respective counter.
+    assign reseed_cnt_reached_d[ai] =
+        state_db_wr_req && state_db_wr_req_rdy && (state_db_wr_inst_id == ai) ?
+            (state_db_wr_rc == reg2hw.reseed_interval.q) :
```
marnovandermaas, line 883: Do we want this to be `>=`? Maybe that case can be reached if we are reseeding based on a large reseed value and then when a lower reseed value is put in the reseed count is already beyond it.

marnovandermaas (commented):
Looks very reasonable to me. Just a small question on the condition in which the exceeded reseed is triggered.

hw/ip/csrng/rtl/csrng_core.sv
h-filali: This count goes up for each generate command that is completed. When the `reseed_interval` is reached no further generates should be accepted. I guess we could add an assertion for that to make sure it doesn't happen. However, I already go another PR cooking that tests this.

h-filali:
> Looks very reasonable to me. Just a small question on the condition in which the exceeded reseed is triggered.

Thanks @marnovandermaas for having a look!

hw/ip/csrng/rtl/csrng_core.sv
marnovandermaas: Ok, it's not a huge deal. I was just thinking it might happen if your counter already has a higher value when you write your reseed interval. We can just put some text in the reseed interval register description saying you should reseed the block to guarantee the new reseed interval takes effect.

h-filali: Ah now I see, thanks for clarifying. I guess I can do it in a follow up PR once this one is merged.

vogelpi:
`clkmgr_jitter_test_fpga_cw310_sival_rom_ext` seems to constantly and reliably timeout with this PR (but not with any others) we should probably investigate this before merging. The test itself is very simple and I suspect the test isn't even reached.

hw/ip/csrng/rtl/csrng_core.sv
marnovandermaas: Yup, that's fine!

Issue #22884, opened 2024-04-29
[rv_core_ibex] Reads using address_translation modes do not work if read is in a different 64kb aligned address


jettr:
### Description

When we enabled [address_translation] in our manifest, we started to see errors, and it took a little while to figure out what was going wrong. I traced it down to the loading of initial values of the static variables in early code. The problem seems to be in the `lw` read of addresses that are not within the same 64kb (`0x1_0000`) boundary as the program counter. I have demonstrated this with the following simplified `_start` function attached to a debugging on the FPGA

The code I am running is as follows:

```
a0010798 <.Ltmp3>: 
a0010798: 98 42        lw·····a4, 0(a3)
a001079a: 91 06        addi···a3, a3, 4
a001079c: f5 bf        j······0xa0010798 <.Ltmp3>
```

where I loaded up `a3` with and address `0x8` before the `0xa0020000` boundary. Then I just ran the loop three times while dumping out flash and register contents below:

```
(gdb) p/x *($a3 + 0x80000000) == $a4
$1 = 0x1
(gdb) i r a3 a4
a3             0xa001fff8       -1610481672
a4             0x44924422       1150436386
(gdb) si 3
0xa001079a in ?? ()
(gdb) p/x *($a3 + 0x80000000) == $a4
$2 = 0x1
(gdb) i r a3 a4
a3             0xa001fffc       -1610481668
a4             0x80820141       -2138963647
(gdb) si 3
0xa001079a in ?? ()
(gdb) p/x *($a3 + 0x80000000) == $a4
$3 = 0x0
(gdb) i r a3 a4
a3             0xa0020000       -1610481664
a4             0x4238282        69436034
(gdb) p/x *($a3 + 0x80000000)
$4 = 0xce061101
(gdb) x/3x 0x2001fff8
0x2001fff8:     0x44924422      0x80820141      0xce061101
```

If `p/x *($a3 + 0x80000000) == $a4` returns true (`0x1`), that means the `lw` read operation worked and the contents in register `a4` matches the value read directly from the memory mapped flash location (e.g. `0x2001fff8`) -- since gdb cannot read the address_translated address for some reason.

I then ran the entire loop `si 3` (single step 3 times), a few times and performed the check again while dumping out the register contents.

[...]

PR #22887, opened 2024-04-29
[rom, rom_ext] Fix address translation

cfrantz:
1. The address translation windows do not properly mask off low bits in hardware.  Mask them off in software before programming.
2. The address translation windows must be power-of-two aligned with respect to the window size.  Forcibly align the remap address.

This partially addresses #22884.

### DIFF
--- a/sw/device/silicon_creator/lib/drivers/ibex.c
+++ b/sw/device/silicon_creator/lib/drivers/ibex.c
@@ -22,9 +22,14 @@ uint32_t ibex_fpga_version(void) {
 
 void ibex_addr_remap_0_set(uint32_t matching_addr, uint32_t remap_addr,
                            size_t size) {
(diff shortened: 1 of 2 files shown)


jettr (approved):
Does this change need to land on master as well?

Also does this workaround issues for both rom-> rom_ext and rom_ext->owner_firmware?

cfrantz:
> Does this change need to land on master as well?

Yes.  Automation will automatically cherry-pick to master when I submit this.

> Also does this workaround issues for both rom-> rom_ext and rom_ext->owner_firmware?

Yes, although because of the position and alignment of ROM_EXT, we'd never observe the buggy behavior with the ROM_EXT.  Nevertheless, the functions which write values to the registers should now be fixed for both cases.

Issue #21583, opened 2024-02-20
[rom,rom_ext] move first two attestation keygen stages (UDS, CDI_0) from ROM_EXT to ROM


timothytrippel:
### Description

For DICE compliance, the UDS attestation key pair should be generated in ROM according to the diagram below. Additionally, since the UDS attestation key is derived from the ROM, and the CDI_0 key derived from the ROM_EXT, it is also (performance-wise) optimal to generate both key pairs in the ROM, since the ROM_EXT measurement is already taken in the ROM as part of the secure boot flow.

![OpenTitan DICE Key Ladder](https://github.com/lowRISC/opentitan/assets/5633066/57aa86ad-75c7-4f77-a211-4819aa02d725)

# Tasks
- [ ] move UDS attestation key/cert gen to the ROM 
- [ ] move CDI_0 attestation key/cert gen to the ROM 
- [ ] update website documentation with the above diagram

timothytrippel:
Note: we considered moving UDS and CDI_0 key / cert gen to the ROM for a better security posture, but decided against it due to a hard 32k ROM size constraint (I.e., we ran out of room in the ROM to implement this). Moreover, performing singing operations (for the CDI_0 attestation certificate endorsement) was deemed potentially dangerous.

This issue has been updated to track the updated plan of record.

johannheyszl:
Thanks @timothytrippel, I am removing the Hotlist label due to reached convergence on the topic. 

We have recorded security discussions for both variants in an (internal) [doc](https://docs.google.com/document/d/1zcXd0KnYZBRsNjPZE1x2V1fuSMjxQX9rxd7yyoYBx74/edit?usp=sharing&resourcekey=0-_mOEdXyu3E6cyLfH4tKj7Q) for later reference.

jadephilipoom:
I just reviewed the doc today (I only heard of it yesterday) and as far as I can tell, the proposed solution still means that the ECDSA keygen/signing is fixed after manufacturing. To me, this raises the same concerns as before about side-channel vulnerabilities we can't patch. It buys only a few months of time to find and fix such issues, at the cost of more complex provisioning. This wasn't at all clear to me before reading the doc today, so I think it would be good to raise this in the security meeting one more time -- sorry, I know it's been a lot of rounds on this! I just think the new proposal is significantly different from previous discussions.

timothytrippel:
[...]

PR #22890, opened 2024-04-29
[rom] measure OTP and set attestation binding CSRs  

timothytrippel:
This updates the ROM to measure the software readable OTP partitions, and use the measurement as input to the attestation binding CSRs of keymgr. This partially addresses #21583.


sw/device/silicon_creator/rom/e2e/keymgr/rom_e2e_keymgr_init_test.c
```diff
@@ -22,6 +23,8 @@
 
 OTTF_DEFINE_TEST_CONFIG();
 
+static uint32_t otp_state[1024] = {0};
```
pamaury, line 26: Is there a named constant that could be used instead of 1024? Or at least check in the code that `kOtpPartitions[p].size` is smaller than the size of this array.

sw/device/silicon_creator/rom/rom.c
```diff
@@ -379,6 +399,34 @@ static void rom_pre_boot_check(void) {
...
+    keymgr_binding_value_t *measurement) {
+  memset(measurement, (int)rnd_uint32(), sizeof(keymgr_binding_value_t));
+  hmac_sha256_init();
+  for (size_t p = 0; p < kOtpPartitionNumPartitions; ++p) {
+    // Rather than calling `hmac_sha256_update()` update here, we copy the OTP
+    // data directly into the HMAC FIFO to reduce code size.
+    HARDENED_CHECK_LE(kOtpPartitions[p].size / sizeof(uint32_t),
+                      HMAC_MSG_FIFO_SIZE_WORDS);
+    HARDENED_CHECK_EQ(kOtpPartitions[p].size % sizeof(uint32_t), 0);
+    otp_dai_read(
```
pamaury, line 418: I guess the FIFO window is large enough but is it worth adding a check against the partition size?

pamaury (approved):
Looks good to me, just small comments.

sw/device/silicon_creator/rom/rom.c
timothytrippel: I do, that is what `HARDENED_CHECK_LE(kOtpPartitions[p].size / sizeof(uint32_t), HMAC_MSG_FIFO_SIZE_WORDS);` `HMAC_MSG_FIFO_SIZE_WORDS` is autogenerated from the fifo window size: https://opentitan.org/book/hw/ip/hmac/doc/registers.html#msg_fifo

sw/device/silicon_creator/rom/e2e/keymgr/rom_e2e_keymgr_init_test.c
timothytrippel: Good ideas. Fixed.

timothytrippel:
Another thought: if we did not want to read out each entire partition (for each software readable partition) on every boot, to minimize boot time we could just read out the digests stored along for each partition, hash those together and lock that in the keymgr attestation binding CSRs. The only downside is for the non-buffered SW partitions, we only write the [least significant 64-bits of SHA256 digest to the digest registers](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=110) so this may not be desirable.

timothytrippel:
I took some _rough_ ROM execution measurement times on the CW310 FPGA by reading out the watchdog count during execution of two tests:
- NO OTP measurements in the ROM (`//sw/device/silicon_creator/rom/e2e/keymgr:rom_e2e_keymgr_init_otp_no_meas`): 76ms
- WITH OTP measurements in the ROM (`//sw/device/silicon_creator/rom/e2e/keymgr:rom_e2e_keymgr_init_otp_meas`): 236ms

Not: this is in no way accurate of what timing will be on actual silicon, since the FPGA clks are must slower and ratios between clks are different. But does highlight the added execution time of performing these OTP readouts and measurements in the ROM is non-neglible. @cfrantz @moidx

sw/device/silicon_creator/rom/rom.c
```diff
@@ -259,35 +272,42 @@ static rom_error_t rom_verify(const manifest_t *manifest,
...
          sizeof(boot_measurements.rom_ext));
 
   CFI_FUNC_COUNTER_INCREMENT(rom_counters, kCfiRomVerify, 2);
-  // Swap the order of signature verifications randomly.
+
+  /**
+   * Verify the RSA/SPX+ signatures of ROM_EXT.
+   *
+   * We swap the order of signature verifications randomly.
+   */
```
moidx, line 310: nit: In-code comments should follow the `// ..` format following the C/C++ style guide.

sw/device/silicon_creator/rom/rom.c
```diff
@@ -398,16 +445,23 @@ static rom_error_t rom_boot(const manifest_t *manifest, uint32_t flash_exec) {
...
-  if (launder32(use_rom_ext_measurement) == kHardenedBoolTrue) {
-    HARDENED_CHECK_EQ(use_rom_ext_measurement, kHardenedBoolTrue);
-    attestation_measurement =
-        (const keymgr_binding_value_t *)&boot_measurements.rom_ext;
+  uint32_t use_otp_measurement =
+      otp_read32(OTP_CTRL_PARAM_OWNER_SW_CFG_ROM_KEYMGR_OTP_MEAS_EN_OFFSET);
+  if (launder32(use_otp_measurement) == kHardenedBoolTrue) {
+    HARDENED_CHECK_EQ(use_otp_measurement, kHardenedBoolTrue);
+    rom_measure_otp_partitions(&otp_measurement);
+    SEC_MMIO_WRITE_INCREMENT(kOtpSecMmioDaiReadPartitionCreatorSwCfg +
```
moidx, line 456: We should chat offline to see if it makes sense to track OTP reads separately to minimize the number of reads triggered by SEC_MMIO. There is probably a way to capture an OTP hash at manufacturing time and just compare that against our in memory OTP copy.

moidx (approved):

sw/device/silicon_creator/rom/rom.c
timothytrippel: fixed.

cdgori:
> Another thought: if we did not want to read out each entire partition (for each software readable partition) on every boot, to minimize boot time we could just read out the digests stored along for each partition, hash those together and lock that in the keymgr attestation binding CSRs. The only downside is for the non-buffered SW partitions, we only write the [least significant 64-bits of SHA256 digest to the digest registers](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=110) so this may not be desirable.

In theory reading all the partitions (again, since presumably the otp controller does this as well to verify the digests?) on every boot can slightly stress the OTP over time - probably not significantly in this case, but other processes / OTP technologies might like it less that what we've got.  It's likely that the read endurance of the OTP is going to be pretty high in any case though so this is likely a secondary issue.

If we had a way to improve the digest width that would be the obvious solution to incorporating the OTP measurement.

Issue #22312, opened 2024-03-28
[hmac,rtl] Flag error for invalid digest size/key length configuration


gdessouky:
### Description

Unsupported/invalid digest sizes and key lengths are now mapped to SHA-2 256 and 256-bit key. We should instead flag an error to SW and not start the HMAC/SHA-2 operation; this is the most conservative and secure approach.

andreaskurth:
@gdessouky I think you're working on this in #22580, right?

PR #22892, opened 2024-04-30
[hmac/dv] Cleanups and coverage fixes/improvements

martin-velay:
- fix and improve coverage model
- fix couple of minor SCB issues
- rename couple objects to be more relevant

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_env_cov.sv
+++ b/hw/ip/hmac/dv/env/hmac_env_cov.sv
@@ -6,12 +6,40 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
   `uvm_component_utils(hmac_env_cov)
 
   covergroup cfg_cg with function sample (bit [TL_DW-1:0] cfg);
-    hmac_en:     coverpoint cfg[HmacEn];
+    hmac_en: coverpoint cfg[HmacEn] {
+      bins disabled = {1'b0};
+      bins enabled  = {1'b1};
+    }
+    sha_en: coverpoint cfg[ShaEn] {
+      bins disabled = {1'b0};
+      bins enabled  = {1'b1};
+    }
     endian_swap: coverpoint cfg[EndianSwap];
     digest_swap: coverpoint cfg[DigestSwap];
-    digest_size: coverpoint cfg[DigestSize];
-    key_length : coverpoint cfg[KeyLength];
-    cfg_cross:   cross hmac_en, endian_swap, digest_swap;
+    digest_size: coverpoint cfg[DigestSizeMsb:DigestSizeLsb] {
+      bins sha2_256     = {4'h1};
+      bins sha2_384     = {4'h2};
+      bins sha2_512     = {4'h4};
+      bins sha2_none    = {4'h8};
+      bins sha2_invalid = digest_size with (!$onehot0(item));
+    }
+    key_length: coverpoint cfg[KeyLengthMsb:KeyLengthLsb] {
+      bins key_128     = {6'h01};
+      bins key_256     = {6'h02};
+      bins key_384     = {6'h04};
+      bins key_512     = {6'h08};
+      bins key_1024    = {6'h10};
+      bins key_none    = {6'h20};
+      bins key_invalid = key_length with (!$onehot0(item));
+    }
+    cfg_cross: cross hmac_en, endian_swap, digest_swap;
+    hmac_dis_x_sha_en: cross hmac_en, sha_en {
+      bins b0 = binsof(hmac_en.disabled) && binsof(sha_en.enabled);
+    }
+    key_x_digest_mismatch: cross key_length, digest_size {
+      bins b0 = binsof(key_length.key_1024) && binsof(digest_size.sha2_256);
+    }
+    key_length_x_digest_size: cross key_length, digest_size;
   endgroup : cfg_cg
 
   covergroup status_cg with function sample (bit [TL_DW-1:0] sta, bit [TL_DW-1:0] cfg);
@@ -20,8 +48,8 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
     digest_swap:      coverpoint cfg[DigestSwap];
     sta_fifo_empty:   coverpoint sta[HmacStaMsgFifoEmpty];
     sta_fifo_full:    coverpoint sta[HmacStaMsgFifoFull];
-    sta_fifo_depth:   coverpoint sta[HmacStaMsgFifoDepth + 4 : HmacStaMsgFifoDepth] {
-      bins fifo_depth[] = {[0:16]};
+    sta_fifo_depth:   coverpoint sta[HmacStaMsgFifoDepthMsb:HmacStaMsgFifoDepthLsb] {
+      bins fifo_depth[] = {[0:2^(HmacStaMsgFifoDepthMsb+1-HmacStaMsgFifoDepthLsb)-1]};
     }
     fifo_empty_cross: cross sta_fifo_empty, hmac_en, endian_swap, digest_swap;
     fifo_full_cross:  cross sta_fifo_full, hmac_en, endian_swap, digest_swap;
@@ -37,50 +65,83 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
       bins update_secret_key_in_process = {SwUpdateSecretKeyInProcess};
       bins hash_start_when_active       = {SwHashStartWhenActive};
       bins push_msg_when_idle           = {SwPushMsgWhenIdle};
+      bins key_length_higher_blk_size   = {SwInvalidConfig};
       illegal_bins illegalvalue         = default;
     }
   endgroup : err_code_cg
 
-  covergroup msg_len_cg with function sample (bit [TL_DW-1:0] msg_len_lower, bit [TL_DW-1:0] cfg);
+  covergroup msg_len_cg with function sample (logic [TL_DW-1:0] msg_len_lower,
+                                              logic [TL_DW-1:0] msg_len_upper,
+                                              logic [TL_DW-1:0] cfg          );
     hmac_en: coverpoint cfg[HmacEn];
-    msg_len: coverpoint (msg_len_lower / 8) {
-      bins len_0         = {0};
-      bins len_1         = {1};
-      bins len_256       = {256};
-      bins len_257       = {257};
-      bins len_512       = {512};
-      bins len_513       = {513};
-      bins len_768       = {768};
-      bins len_769       = {769};
-      bins len_1024      = {1024};
-      bins len_1025      = {1025};
-      bins len_1280      = {1280};
-      bins len_1281      = {1281};
-      bins len_1536      = {1536};
-      bins len_1537      = {1537};
-      bins len_1792      = {1792};
-      bins len_1793      = {1793};
-      bins len_2048      = {2048};
-      bins len_2049      = {2049};
-      bins len_2304      = {2304};
-      bins len_2305      = {2305};
-      bins len_2560      = {2560};
-      bins len_2561      = {2561};
-      bins len_2816      = {2816};
-      bins len_2817      = {2817};
-      bins len_3072      = {3072};
-      bins len_3073      = {3073}; // up to 3000 for the most SW usage
-      bins auto_lens[50] = {[1:10000]};
-     }
-    msg_len_cross: cross hmac_en, msg_len;
+    // Register is in bits but we are interested in the number of bytes => /8
+    msg_len_lower_cp: coverpoint (msg_len_lower / 8) {
+      bins len_0         = {0};     // Empty message
+      bins len_1         = {1};     // One byte message
+      bins len_511       = {511};   // One block in SHA-2 256, -1 byte
+      bins len_512       = {512};   // One block in SHA-2 256
+      bins len_513       = {513};   // One block in SHA-2 256, +1 byte
+      bins len_1023      = {1023};  // One block in SHA-2 384/512 or two in SHA-2 256, -1 byte
+      bins len_1024      = {1024};  // One block in SHA-2 384/512 or two in SHA-2 256
+      bins len_1025      = {1025};  // One block in SHA-2 384/512 or two in SHA-2 256, +1 byte
+      bins len_2047      = {2047};  // Two blocks in SHA-2 384/512, -1 byte
+      bins len_2048      = {2048};  // Two blocks in SHA-2 384/512
+      bins len_2049      = {2049};  // Two blocks in SHA-2 384/512, +1 byte
+      // Any others than the one defined above
+      bins auto_lens[50] = {[0:$]} with (!(item inside {0, 1, 511, 512, 513, 1023, 1024, 1025,
+                                                        2047, 2048, 2049}));
+    }
+    // Ensure that message length upper register has been used once at least
+    msg_len_upper_cp: coverpoint (msg_len_upper) {
+      bins len_upper     = {[1:$]};
+    }
+    msg_len_lower_cross: cross hmac_en, msg_len_lower_cp;
+    msg_len_upper_cross: cross hmac_en, msg_len_upper_cp;
   endgroup : msg_len_cg
+
+  covergroup wr_config_during_hash_cg with function sample (logic wr_config_during_hash);
+    cp: coverpoint wr_config_during_hash {bins true = {1'b1};}
+  endgroup : wr_config_during_hash_cg
+
+  covergroup wr_key_during_hash_cg with function sample (logic wr_key_during_hash);
+    cp: coverpoint wr_key_during_hash {bins true = {1'b1};}
+  endgroup : wr_key_during_hash_cg
+
+  covergroup wr_key_during_sha_only_cg with function sample (logic wr_key_during_sha_only);
+    cp: coverpoint wr_key_during_sha_only {bins true = {1'b1};}
+  endgroup : wr_key_during_sha_only_cg
+
+  covergroup wr_msg_during_hash_cg with function sample (logic wr_msg_during_hash);
+    cp: coverpoint wr_msg_during_hash {bins true = {1'b1};}
+  endgroup : wr_msg_during_hash_cg
+
+  covergroup trig_rst_during_hash_cg with function sample (logic trig_rst_during_hash);
+    cp: coverpoint trig_rst_during_hash {bins true = {1'b1};}
+  endgroup : trig_rst_during_hash_cg
+
+  covergroup rd_digest_during_hmac_en_cg with function sample (logic rd_digest_during_hmac_en);
+    cp: coverpoint rd_digest_during_hmac_en {bins true = {1'b1};}
+  endgroup : rd_digest_during_hmac_en_cg
+
+  covergroup save_and_restore_cg with function sample (save_and_restore_e sar_ctxt);
+    save_and_restore_cp : coverpoint sar_ctxt {
+      bins same_context       = {SameContext};
+      bins different_context  = {DifferentContext};
+    }
+  endgroup : save_and_restore_cg
 
   function new(string name, uvm_component parent);
     super.new(name, parent);
-    cfg_cg      = new();
-    status_cg   = new();
-    msg_len_cg  = new();
-    err_code_cg = new();
+    cfg_cg                      = new();
+    status_cg                   = new();
+    msg_len_cg                  = new();
+    err_code_cg                 = new();
+    wr_config_during_hash_cg    = new();
+    wr_key_during_hash_cg       = new();
+    wr_key_during_sha_only_cg   = new();
+    wr_msg_during_hash_cg       = new();
+    trig_rst_during_hash_cg     = new();
+    rd_digest_during_hmac_en_cg = new();
+    save_and_restore_cg         = new();
   endfunction : new
-
 endclass
--- a/hw/ip/hmac/dv/env/hmac_env_pkg.sv
+++ b/hw/ip/hmac/dv/env/hmac_env_pkg.sv
@@ -21,12 +21,13 @@ package hmac_env_pkg;
   `include "dv_macros.svh"
 
   // local parameters and types
-  parameter uint32 HMAC_MSG_FIFO_DEPTH       = 32;
-  parameter uint32 HMAC_MSG_FIFO_DEPTH_BYTES = HMAC_MSG_FIFO_DEPTH * 4;
-  parameter uint32 HMAC_MSG_FIFO_SIZE        = 2048;
-  parameter uint32 HMAC_MSG_FIFO_BASE        = 32'h1000;
-  parameter uint32 HMAC_MSG_FIFO_LAST_ADDR   = HMAC_MSG_FIFO_BASE + HMAC_MSG_FIFO_SIZE - 1;
-  parameter uint32 HMAC_HASH_SIZE            = 64;
+  parameter uint32 HMAC_MSG_FIFO_DEPTH        = 32;
+  parameter uint32 HMAC_MSG_FIFO_DEPTH_BYTES  = HMAC_MSG_FIFO_DEPTH * 4;
+  parameter uint32 HMAC_MSG_FIFO_SIZE         = 2048;
+  parameter uint32 HMAC_MSG_FIFO_BASE         = 32'h1000;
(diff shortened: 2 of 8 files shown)


andreaskurth:
(I'm aware that this is currently a draft, just assigned Rupert and me for awareness and so we can get back to you quickly once the PR is ready for review.)

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -6,12 +6,38 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
   `uvm_component_utils(hmac_env_cov)
 
   covergroup cfg_cg with function sample (bit [TL_DW-1:0] cfg);
-    hmac_en:     coverpoint cfg[HmacEn];
+    hmac_en: coverpoint cfg[HmacEn] {
```
rswarbrick, line 9: Learning question from me! Does this coverpoint change make any change to what we cover? Or is it just ensuring that e.g. "0" is reported as "disabled"?

(I'm perfectly happy with the change, but want to make sure I understand it too!)

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -6,12 +6,38 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
...
+      bins key_256      = {5'h02};
+      bins key_384      = {5'h04};
+      bins key_512      = {5'h08};
+      bins key_1024     = {5'h10};
+      bins key_invalid  = default;
+    }
+    cfg_cross: cross hmac_en, endian_swap, digest_swap;
+    hmac_dis_x_sha_en: cross hmac_en, sha_en {
+      bins b0 = binsof(hmac_en.disabled) && binsof(sha_en.enabled);
+    }
```
rswarbrick, line 37: Is this equivalent to adding a coverpoint that tracks `{hmac_en, sha_en}` with a bin of `{1'b0, 1'b1}`? I don't really have strong feelings, but using a cross for a single point feels like the code is looking more complicated than it needs to.

(Similar question for the cross below)

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -20,8 +46,8 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
     digest_swap:      coverpoint cfg[DigestSwap];
     sta_fifo_empty:   coverpoint sta[HmacStaMsgFifoEmpty];
     sta_fifo_full:    coverpoint sta[HmacStaMsgFifoFull];
-    sta_fifo_depth:   coverpoint sta[HmacStaMsgFifoDepth + 4 : HmacStaMsgFifoDepth] {
-      bins fifo_depth[] = {[0:16]};
+    sta_fifo_depth:   coverpoint sta[HmacStaMsgFifoDepthMsb:HmacStaMsgFifoDepthLsb] {
+      bins fifo_depth[] = {[0:2^(HmacStaMsgFifoDepthMsb+1-HmacStaMsgFifoDepthLsb)]};
```
rswarbrick, line 50: The power of two for the upper limit here looks a bit odd. I think we're sampling a signal with `HmacStaMsgFifoDepthMsb - HmacStaMsgFifoDepthLsb` bits, so the maximum possible value is `2^(HmacStaMsgFifoDepthMsb - HmacStaMsgFifoDepthLsb + 1) - 1`. Have we just got an extra bin at the top?

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -37,50 +63,68 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
       bins update_secret_key_in_process = {SwUpdateSecretKeyInProcess};
       bins hash_start_when_active       = {SwHashStartWhenActive};
       bins push_msg_when_idle           = {SwPushMsgWhenIdle};
+      bins key_length_higher_blk_size   = {KeyLengthHigherBlkSize};
       illegal_bins illegalvalue         = default;
     }
   endgroup : err_code_cg
 
-  covergroup msg_len_cg with function sample (bit [TL_DW-1:0] msg_len_lower, bit [TL_DW-1:0] cfg);
+  // Declared as logic with x as default to avoid undesired sampling when register not written
```
rswarbrick, line 71: I wonder whether it would be simpler to just declare the arguments without defaults and to pass `'x` at the call site if there is no value. (It might be easier to follow than looking at the call site and noticing that there were fewer arguments?)

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -37,50 +63,68 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
...
+  covergroup msg_len_cg with function sample (logic [TL_DW-1:0] msg_len_lower = 'x,
+                                              logic [TL_DW-1:0] msg_len_upper = 'x,
+                                              logic [TL_DW-1:0] cfg           = 'x);
     hmac_en: coverpoint cfg[HmacEn];
-    msg_len: coverpoint (msg_len_lower / 8) {
+    // Register is in bits but we are interested in the number of bytes => /8
+    msg_len_lower_cp: coverpoint (msg_len_lower / 8) {
```
rswarbrick, line 77: Where do the numbers in the bins below come from? Maybe it makes sense to add something in the comment that explains why they are the right lengths to track?

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -37,50 +63,68 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
...
-      bins len_2817      = {2817};
-      bins len_3072      = {3072};
-      bins len_3073      = {3073}; // up to 3000 for the most SW usage
-      bins auto_lens[50] = {[1:10000]};
-     }
-    msg_len_cross: cross hmac_en, msg_len;
+      bins len_2056      = {2056};
+      bins auto_lens[50] = default;
+    }
+    // Ensure that message length upper regsiter has been used once at least
```
rswarbrick, line 91: Nit: Spelling of "register"

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -37,50 +63,68 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
...
+    msg_len_upper_cp: coverpoint (msg_len_upper) {
```
rswarbrick, line 92: (SystemVerilog question:) Would it make sense to use e.g. `coverpoint (msg_len_upper > 0)` ? It might be shorter?

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -37,50 +63,68 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
...
+      bins len_upper     = {[1:$]};
+    }
+    msg_len_lower_cross: cross hmac_en, msg_len_lower_cp;
+    msg_len_upper_cross: cross hmac_en, msg_len_upper_cp;
   endgroup : msg_len_cg
 
+  // Declared as logic with x as default to avoid undesired sampling when register not written
+  covergroup do_during_cg with function sample (logic wr_config_during_hash     = 'x,
```
rswarbrick, line 100: Wouldn't it be easier to declare a separate covergroup for each coverpoint? Looking at the code below, you only seem to sample this with one argument each time. Maybe it would be clearer if the coverpoints weren't artificially grouped?

rswarbrick (commented):
Lots of nitty questions about ways the covergroups might be easier to express, but this looks reasonable to me.

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: I have added the bins "disabled" and "enabled" to be able to make the proper cross below (see hmac_dis_x_sha_en)

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: Yes it would be the same. I could change it if it bothers you, but I feel it's clearer with the bins names rather than value concatenation in my opinion.

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: Oh yes, a -1 is missing there. Thanks

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: The purpose was to be able to call the sample function without having to specify all the arguments:
``cov.msg_len_cg.sample(.msg_len_lower(item.d_data), .cfg(ral.cfg.get_mirrored_value()));``

But I can change it if you want.

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: You're right, I'll do that

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: I'll break them into separate covergroups then

hw/ip/hmac/dv/env/hmac_env_cov.sv
rswarbrick: I don't feel strongly but wanted to make sure I understood what this was doing. Seems sensible to me.

rswarbrick (commented):
This looks sensible. I've left a few more nitty comments, but nothing major.

However: For your next PR, please could you file more than one commit? This PR is pretty enormous (+438/-450) and contains a bunch of unrelated changes that all need reviewing at once. It's not particularly efficient for the reviewer and getting things merged would be much simpler if you could put a bit of work into submitting it in smaller pieces.

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -41,41 +42,45 @@ package hmac_env_pkg;
...
   parameter string LIST_OF_ALERTS[] = {"fatal_fault"};
 
-  typedef enum {
-    HmacDone,
-    HmacMsgFifoEmpty,
-    HmacErr
+  typedef enum int {
+    HmacDone          = 0,
+    HmacMsgFifoEmpty  = 1,
+    HmacErr           = 2
```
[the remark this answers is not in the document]
martin-velay: Done

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -41,41 +42,45 @@ package hmac_env_pkg;
...
   } hmac_intr_e;
 
-  typedef enum {
-    HmacStaMsgFifoEmpty,
-    HmacStaMsgFifoFull,
-    HmacStaMsgFifoDepth = 4
+  typedef enum int {
```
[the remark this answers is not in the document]
martin-velay: Done

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -94,8 +96,7 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
 
                 `uvm_info(`gfn, $sformatf(
                           "Setting previous digest and digest swap: %4b",
-                           previous_digest_size), UVM_HIGH)
-
+                          previous_digest_size), UVM_HIGH)
```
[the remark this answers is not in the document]
martin-velay: OK
Just a parenthesis on this because it slightly degrade our time. In my opinion, this 100 characters limit is making the code less readable. Do you reckon to increase it up to 150 for example for the DV only maybe?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -123,11 +124,19 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
             // Do nothing
           end
           "cfg": begin
-            if (hmac_start) return; // won't update configs if hash start
+            if (hmac_start) begin
+              if (cfg.en_cov) cov.wr_config_during_hash_cg.sample(1);
+              return; // won't update configs if hash start
+            end
             if (cfg.en_cov) cov.cfg_cg.sample(item.a_data);
+            // Changing the key length should be seemless while HMAC disabled
```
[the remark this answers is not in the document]
martin-velay: Yes, thanks.

andreaskurth:
This PR now contains commits that aren't related to HMAC. Please rebase on `master` and drop them.

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -41,41 +42,45 @@ package hmac_env_pkg;
...
-    SwHashStartWhenActive      = 32'h 0000_0004,
-    SwPushMsgWhenIdle          = 32'h 0000_0005
+  typedef enum int {
+    NoError                    = 0,
+    SwPushMsgWhenShaDisabled   = 1,
+    SwHashStartWhenShaDisabled = 2,
+    SwUpdateSecretKeyInProcess = 3,
+    SwHashStartWhenActive      = 4,
+    SwPushMsgWhenIdle          = 5,
+    KeyLengthHigherBlkSize     = 99   // TODO MVy: TBD
```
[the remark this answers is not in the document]
martin-velay: It is related to this issue #22312, I'll add it to the comment

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -147,14 +157,14 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+          "digest_0", "digest_1", "digest_2", "digest_3", "digest_4", "digest_5", "digest_6",
           "digest_7", "digest_8", "digest_9", "digest_10", "digest_11", "digest_12", "digest_13",
           "digest_14", "digest_15", "status", "msg_length_upper", "msg_length_lower": begin
             // Predict updated value coming from write iff SHA core is disabled.
             do_predict = !sha_en;
           end
           "status": begin
-            `uvm_error(`gfn, $sformatf("this reg does not have write access: %0s",
-                                       csr.get_full_name()))
+            do_predict  = 0;
```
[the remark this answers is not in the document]
martin-velay: Yes it's true ! I'll remove this case statement

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -179,19 +189,19 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
-          bit             hmac_fifo_empty  = hmac_fifo_depth == 0;
-          bit [TL_DW-1:0] hmac_status_data = (hmac_fifo_empty << HmacStaMsgFifoEmpty) |
-                                             (hmac_fifo_full  << HmacStaMsgFifoFull) |
-                                             (hmac_fifo_depth << HmacStaMsgFifoDepth);
-          void'(ral.status.predict(.value(hmac_status_data), .kind(UVM_PREDICT_READ)));
-        end else if (csr_name == "intr_state") begin
-          if (fifo_empty && ral.intr_state.fifo_empty.get_mirrored_value() != 1) begin
-            void'(ral.intr_state.fifo_empty.predict(.value(1), .kind(UVM_PREDICT_READ)));
-          end
+      if (csr_name == "status") begin
```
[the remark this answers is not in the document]
martin-velay: Understood

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -343,23 +365,25 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
       //if (do_read_check) begin
       //  `uvm_info(`gfn, $sformatf("%s reg is checked with expected value %0h",
       //                            csr_name, csr.get_mirrored_value()), UVM_LOW)
-      //  `DV_CHECK_EQ(csr.get_mirrored_value(), item.d_data, csr_name)
+      //  `DV_CHECK_EQ(item.d_data, csr.get_mirrored_value(), csr_name)
```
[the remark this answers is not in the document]
martin-velay: This shouldn't be commented, Ghada or I will enable this check soon. And about the swap, what the macro will print is commutative but there is still a definition of it containing the words "ACT" and "EXP", which confused me first. I'll keep my modification

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -291,41 +294,39 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
...
+        // Expected error as we may not push message into the FIFO while SHA is disabled
+        if (!ral.cfg.sha_en.get_mirrored_value()) begin
           check_error_code();
         end
       end else begin // remaining msg is smaller than the burst_wr_length
         wr_msg(msg_q);
         break;
       end
-    csr_utils_pkg::wait_no_outstanding_access();
-    if ($urandom_range(0, 1)) rd_msg_length();
```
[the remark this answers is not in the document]
martin-velay: Interesting, I think I have removed it by mistake while I have added `read_status_intr_clr`. I'll add it back, thanks for spotting it!

hw/ip/hmac/dv/env/seq_lib/hmac_datapath_stress_vseq.sv
```diff
@@ -9,17 +9,22 @@ class hmac_datapath_stress_vseq extends hmac_smoke_vseq;
...
   }
 
   constraint msg_c {
-    msg.size() == 1 + msg_size_base * HMAC_HASH_SIZE;
+    solve digest_size before msg;
+    if (digest_size == SHA2_256) {
+      msg.size() == 1 + nb_blk_msg * HMAC_BLK_SIZE_SHA2_256;
+    } else {
+      msg.size() == 1 + nb_blk_msg * HMAC_BLK_SIZE_SHA2_384_512;
+    }
```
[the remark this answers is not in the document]
martin-velay: Shorter yes, but also harder to read IMO.
For the same number of lines:
```
    if (digest_size == SHA2_256) {msg.size() == 1 + nb_blk_msg * HMAC_BLK_SIZE_SHA2_256;}
    else                         {msg.size() == 1 + nb_blk_msg * HMAC_BLK_SIZE_SHA2_384_512;}
```
@rswarbrick, is that compliant with the coding rules? Do you prefer this version?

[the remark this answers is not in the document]
rswarbrick: Not particularly: your original code looks perfectly sensible. My point was that we're just changing a block size, so it might be nice to avoid repeating all the other code around. But I agree that the result might get ugly and hard to read.

hw/ip/hmac/dv/env/hmac_env_cov.sv
rswarbrick: I'm sorry for all the noise here. I see that you've dropped the values that made things implicit arguments. Doesn't that mean that we should drop this comment?

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -37,50 +63,82 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
...
-      bins len_3072      = {3072};
-      bins len_3073      = {3073}; // up to 3000 for the most SW usage
-      bins auto_lens[50] = {[1:10000]};
-     }
-    msg_len_cross: cross hmac_en, msg_len;
+    // Register is in bits but we are interested in the number of bytes => /8
+    msg_len_lower_cp: coverpoint (msg_len_lower / 8) {
+      bins len_0         = {0};     // Empty message
+      bins len_8         = {8};     // One byte message
+      bins len_504       = {504};   // One block in SHA-2 256, -1 byte
```
rswarbrick, line 80: I'm a bit confused about the numbers here. We've divided `msg_len_lower` by 8 above. Doesn't that mean that this byte length should be 1 less than the next one? (Similar question for the other triples of lengths below)

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -63,6 +63,8 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
         end else if (hmac_start && !cfg.under_reset) begin
           bit [7:0] bytes[4];
           bit [7:0] msg[$];
+          // During compression rounds, SW should be able to write into the message FIFO
```
rswarbrick, line 66: If I understand correctly, the code is saying "Oh, we've seen a write to the message FIFO during a compression round". That's not quite the same as the fact that doing so is allowed. Maybe this comment should be something like "Register coverage for the fact that we have seen a write to the message FIFO during a compression round"?

rswarbrick (commented):
Thanks for the responses. I spotted a couple of other minor nits, but this looks pretty sensible to me.

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: You're right. Good catch, thanks!

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -6,12 +6,38 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
...
+      bins key_512      = {5'h08};
+      bins key_1024     = {5'h10};
+      bins key_invalid  = default;
+    }
+    cfg_cross: cross hmac_en, endian_swap, digest_swap;
+    hmac_dis_x_sha_en: cross hmac_en, sha_en {
+      bins b0 = binsof(hmac_en.disabled) && binsof(sha_en.enabled);
+    }
+    key_length_x_digest_size: cross key_length, digest_size {
+      bins b0 = binsof(key_length.key_1024) && binsof(digest_size.sha2_256);
```
andreaskurth, line 39: Why only this bin?

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -41,41 +42,49 @@ package hmac_env_pkg;
...
-  typedef enum {
-    HmacStaMsgFifoEmpty,
-    HmacStaMsgFifoFull,
-    HmacStaMsgFifoDepth = 4
+  // HMAC status register indices
+  typedef enum int {
+    HmacStaMsgFifoEmpty     = 0,
+    HmacStaMsgFifoFull      = 1,
+    HmacStaMsgFifoDepthLsb  = 4,
+    HmacStaMsgFifoDepthMsb  = 9
```
andreaskurth, line 57: This will break if someone changes `hmac.hjson` without changing these values, right?

I think ideally we would expose these constants in a generated SV file (probably `hmac_reg_pkg.sv`), so that we don't have to redefine them here. If that makes sense to you, please create an issue and track this as future enhancement. No need to change anything in this PR, though.

andreaskurth (approved):
Thanks for these fixes and improvements, @martin-lueker! This LGTM overall, I just have one question and one comment.

andreaskurth:
You probably have to rebase on `master` to include CI fixes that should get the failing CI checks to pass

hw/ip/hmac/dv/env/hmac_env_pkg.sv
martin-velay: Yes, it will break.
From a DV perspective, that's not a good practice to share files with DD even though it's auto generated, as it can hide some issues. I know it's done like that in this project, but that's probably something we should think of changing in the future. The DD and DV should "fork" into 2 distinct branches based on the specification.
Do you still want to go down this route?

martin-velay:
Hi @andreaskurth,
It is now ready to be merged according to me.

andreaskurth:
CI checks were failing. Although I don't think the failures are caused by this PR, it would still be better to have all checks pass before merging. I rebased on `master` to check if that resolves the CI problem.

Other than that, this LGTM

andreaskurth:
As this PR only touches DV code, it cannot cause the observed CI check failures on FPGA. All other CI checks pass. Merging

PR #22897, opened 2024-04-30
[sival, rv_dm] improve jtag_tap_sel to support multiple lc states

nbdd0121:
This test now tests PROD states and check in RMA state the tap is continuously sampled. This makes it cover chip_sw_tap_strap_sampling test.

### DIFF
--- a/sw/host/tests/chip/rv_dm/src/jtag_tap_sel.rs
+++ b/sw/host/tests/chip/rv_dm/src/jtag_tap_sel.rs
@@ -42,54 +42,103 @@ fn test_jtag_tap_sel(opts: &Opts, transport: &TransportWrapper) -> Result<()> {
     let lc_state =
         DifLcCtrlState::from_redundant_encoding(jtag.read_lc_ctrl_reg(&LcCtrlReg::LcState)?)?;
 
+    // For test_unlocked* and RMA state, test that we can switch TAP without reset.
+    let (needs_reset, rv_dm_enabled) = match lc_state {
+        DifLcCtrlState::TestUnlocked0
+        | DifLcCtrlState::TestUnlocked1
+        | DifLcCtrlState::TestUnlocked2
+        | DifLcCtrlState::TestUnlocked3
+        | DifLcCtrlState::TestUnlocked4
+        | DifLcCtrlState::TestUnlocked5
+        | DifLcCtrlState::TestUnlocked6
+        | DifLcCtrlState::TestUnlocked7
+        | DifLcCtrlState::Rma => (false, true),
+        DifLcCtrlState::Dev => (true, true),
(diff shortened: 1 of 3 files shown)


sw/host/tests/chip/rv_dm/src/jtag_tap_sel.rs
```diff
@@ -42,54 +42,103 @@ fn test_jtag_tap_sel(opts: &Opts, transport: &TransportWrapper) -> Result<()> {
...
     // Test DFT TAP
     transport.pin_strapping("PINMUX_TAP_DFT")?.apply()?;
-    transport.reset_target(opts.init.bootstrap.options.reset_delay, true)?;
-
-    // For DFT test we need to do raw OpenOCD command.
+    if needs_reset {
+        transport.reset_target(opts.init.bootstrap.options.reset_delay, true)?;
+    }
+
+    // If RV-DM is disabled, strapping DFT tap will result in LC TAP being selected.
```
pamaury, line 96: Why is there a link between selecting the DFT tap and rv_dm being enabled? I thought DFT and RV_DM used different strappings and also are not enabled in the same LC states (DFT is not enabled in DV but RV_DM is)

nbdd0121: DFT is encoded as strapping both LC and RV_DM strapping pins. The pinmux gates rv_dm access by gating on the strapping pin, instead of first use the pin to determine which tap to access and then gate the tap.

Arguably the pinmux logic could be improved but I am not sure if RTL is still changeable?

pamaury: Hum, I thought the strapping pins were just a 2-bit encoded value (so 4 possible values). The [documentation](https://opentitan.org/book/hw/ip/pinmux/doc/theory_of_operation.html?highlight=strap#strap-sampling-and-tap-isolation) does not suggest that each pin is associated with a particular tap.
If the RTL cannot be changed (likely) then the documentation probably should be.

moidx: Yes, I think this will probably require a documentation update. The behavior of selecting LC TAP by default is probably ok when `TAP_STRAP = 2'b11`. 

@timothytrippel, what is the behavior you have observed wrt the `DFT_STRAP` pins. Do they override the `TAP_STRAP` pins? Or are we just always pulling the `TAP_STRAP` pins by default in the transport? From inspecting the transport configuration it seems that the default `TAP_STRAP = 2'b00`, which makes me think that `DFT_STRAP` override `TAP_STRAP`.

CC @andreaskurth

moidx (approved):

PR #22898, opened 2024-04-30
[sival] add field to testplan to indicate required boot stages

nbdd0121:
Due to rom_ext locking down a bunch of stuff, some tests cannot be run in silicon owner stage. This is orthogonal to the lifecycle states, so adding a new field to testplan to indicate this.

The idea is that the new `boot_stages` field is used to indicate which boot stages the test can run in. For example,
```
boot_stages: ["rom_ext"]
```
limits the test to rom_ext stage, where
```
boot_stages: ["silicon_owner"]
```
limits the test to silicon owner stage. I suppose for tests that use SRAM injection they can have `"rom"` as boot stage since they run before ROM is executed.

When nothing is specified the test should be treated as having
```
boot_stages: ["rom_ext", "silicon_owner"]
```

### DIFF
--- a/hw/top_earlgrey/data/chip_testplan.hjson
+++ b/hw/top_earlgrey/data/chip_testplan.hjson
@@ -509,7 +509,6 @@
               "chip_sw_hmac_smoketest",
               "chip_sw_kmac_smoketest",
               "chip_sw_otbn_smoketest",
-              "chip_sw_otp_ctrl_smoketest",
               "chip_sw_pwrmgr_smoketest",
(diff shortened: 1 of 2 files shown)


moidx (approved):
Hi @nbdd0121, this sounds good to me. 

It would be good to document somewhere, maybe under, `sw/device/tests/doc/sival/devguide.md` the mapping of `boot_stages` to Bazel `exec_env`?

I initially thought we could just maintain the list of `exec_env` in the Bazel targets, but in retrospect, I think it is useful to have this level of information in the test plan.

pamaury (approved):

PR #22900, opened 2024-04-30
[sival,sram_ctrl] Fix sram_ctrl_sleep_contents_ret test

jwnrt:
The retention SRAM contains a "version" field which the ROM_EXT expects to be valid in order to boot. The ES version of the ROM does not set this field on every boot. Since the version is invalid after a wipe / scrambling of the retention SRAM, this test was failing due to boot errors.

This is fixed in the production ROM, but we've had to work around it for ES.

The scrambling version of this test:

1. Changes the scrambling key from the default.
2. Writes a pattern to retention SRAM and goes into a normal sleep. After waking up, we check that the pattern reads back the same.
3. Goes into deep sleep and resets using the AON timer.
4. With this kind of reset, the ROM should restore the scrambling key we changed to. We check this by reading back the pattern again.
5. Resets "normally" using the watchdog timer.
6. With that kind of reset, the ROM restores the default scrambling key. We check this happened by trying to read back the pattern and seeing that it has changed.

The workaround:

1. Swaps the order of the deep sleep reset and the "normal" watchdog reset. This is necessary to make the next change work.
2. When scrambling before the normal reset, we manually restore the `version` field into retention SRAM after the wipe but _before_ the scrambling key is changed. This means it's stored using the default scrambling key and the next boot can read it correctly.
3. When scrambling before the deep sleep reset, we manually restore the `version` field _after_ the scrambling key is changed. This means it's stored using our new scrambling key and can be read back correctly at the next boot.

### DIFF
--- a/sw/device/tests/sram_ctrl_sleep_sram_ret_contents_impl.c
+++ b/sw/device/tests/sram_ctrl_sleep_sram_ret_contents_impl.c
@@ -190,7 +190,7 @@ bool execute_sram_ctrl_sleep_ret_sram_contents_test(bool scramble) {
       // retention SRAM when it boots after being scrambled. The ROM_EXT
       // expects a valid version and refuses to boot otherwise.
       //
-      // This is fixed in the prod ROM, but we can must work around it in ES
+      // This is fixed in the prod ROM, but we must work around it in ES
       // tests by restoring the version after the wipe.
       //
       // We write the version pre-scrambling since we want it to be readable
--- a/sw/device/tests/sram_ctrl_sleep_sram_ret_contents_impl.c
+++ b/sw/device/tests/sram_ctrl_sleep_sram_ret_contents_impl.c
@@ -38,6 +38,7 @@ static dif_rv_plic_t rv_plic;
 static dif_aon_timer_t aon_timer;
 static dif_pwrmgr_t pwrmgr;
 static dif_rstmgr_t rstmgr;
+static dif_sram_ctrl_t ret_sram;
 static plic_isr_ctx_t plic_ctx = {.rv_plic = &rv_plic,
                                   .hart_id = kTopEarlgreyPlicTargetIbex0};
 static pwrmgr_isr_ctx_t pwrmgr_isr_ctx = {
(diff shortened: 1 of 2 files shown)


sw/device/tests/sram_ctrl_sleep_sram_ret_contents_impl.c
```diff
@@ -188,32 +185,59 @@ bool execute_sram_ctrl_sleep_ret_sram_contents_test(bool scramble) {
...
+
       LOG_INFO("Wiping ret_sram...");
       CHECK_STATUS_OK(sram_ctrl_testutils_wipe(&ret_sram));
       LOG_INFO("Scrambling ret_sram...");
       CHECK_STATUS_OK(sram_ctrl_testutils_scramble(&ret_sram));
+
+      // Restore the retention SRAM version from before it was scrambled.
+      retention_sram_get()->version = version;
     }
-    test_ret_sram_in_normal_sleep();
```
nbdd0121, line 200: Is this removal intentional?

jwnrt: Nope, it was a mistake. Thanks for catching.

nbdd0121 (approved):

sw/device/tests/sram_ctrl_sleep_sram_ret_contents_impl.c
```diff
@@ -188,32 +185,61 @@ bool execute_sram_ctrl_sleep_ret_sram_contents_test(bool scramble) {
...
-      dif_sram_ctrl_t ret_sram;
-      addr =
-          mmio_region_from_addr(TOP_EARLGREY_SRAM_CTRL_RET_AON_REGS_BASE_ADDR);
-      CHECK_DIF_OK(dif_sram_ctrl_init(addr, &ret_sram));
+      // NOTE:
+      // The ES version of the ROM does not set the version field of the
+      // retention SRAM when it boots after being scrambled. The ROM_EXT
+      // expects a valid version and refuses to boot otherwise.
+      //
+      // This is fixed in the prod ROM, but we can must work around it in ES
```
engdoreis, line 193:
```suggestion
      // This is fixed in the prod ROM, but we must work around it in ES
```

engdoreis (approved):

jwnrt:
CI failures unrelated.

PR #22902, opened 2024-04-30
[rv_dm,rtl] Fix location of SEC_CM comment

rswarbrick:
This comment is supposed to be explaining that we're using a mubi value for the late debug signal. It was not quite in the right place: move it to fix that.


vogelpi (approved):
When doing the D2S RTL review I also noted this. Thanks, this looks good to me.

PR #22920, opened 2024-05-01
[manuf] add TPM cert generation to personalization binary

vbendeb:
This patchset adds generatiinig of TPM certificates during personalization, storing  all three TPM certificates in a dedicated flash INFO page and then hashing the certificates contents on the device after reading them back from flash, and comparing with the hash calculated on the host.


vbendeb:
There are some style inconsistencies which slipped through while I was splitting the patch, I'll fix then along with other review comments.

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -244,6 +245,52 @@ static uint32_t get_cert_size(const uint8_t *header) {
...
+ *
+ * If the caller passed a pointer, save there the certificate size.
+ */
+static status_t hash_certificate(const flash_ctrl_info_page_t *page,
+                                 size_t offset, sha256_state_t *state,
+                                 size_t *size) {
+  uint8_t buffer[1024];  // 1K should be enough for the largest possible
+                         // certificate.
+  uint32_t cert_size;
+  uint32_t bytes_read;  // Number of bytes already read.
```
timothytrippel, line 261: is this variable need or can you just pass `offset + 1` to the second call to `flash_ctrl_info_read()`?

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -274,25 +362,126 @@ static status_t personalize_dice_certificates(ujson_t *uj) {
...
+  // string, two bytes per digit plus trailing zero.
+  char hash_string[sizeof(hash) * 2 + 1];
+  memset(hash_string, 0, sizeof(hash_string));
+  size_t room = sizeof(hash_string);
+
+  for (i = 0; i < sizeof(hash); i++) {
+    room -= base_snprintf(hash_string + sizeof(hash_string) - room, room,
+                          "%02x", ((unsigned char *)&hash)[i]);
+  }
+  LOG_INFO("Certs hash: %s", hash_string);
```
timothytrippel, line 480:
```suggestion
  LOG_INFO("Sha256Sum of Certificates: 0x%08x%08x%08x%08x%08x%08x%08x%08x", hash[7],hash[6],hash[5],hash[4],hash[3],hash[2],hash[1],hash[0]);
```?

**Comment by timothytrippel** [comment]
One thing I realized (though it is ok to address in a follow up PR): the HJSON definitions of each TPM cert are almost identical, with the exception of the subject common name. If we made that a variable, then couldn't we just define one TPM HJSON cert template and instantiate the 3 different versions (EK, CIK, CEK)? Or would the variability in the sized of the subject common name's cause an issue? @pamaury

**Comment by timothytrippel** [comment]
Just a few cleanup suggestions otherwise this looks great @vbendeb ! Thanks for fixing my bugs too :)

**Comment by timothytrippel** [comment]
Also, is the cert chain validation step on the host side missing? or were you planning on adding that in a follow up PR? @vbendeb

File: sw/host/provisioning/ft_lib/src/lib.rs [hunk 3]
```diff
@@ -247,25 +247,63 @@ pub fn run_ft_personalize(
...
+        .clone()
+        .into_iter()
+        .take(certs.tpm_cik_tbs_certificate_size)
+        .collect();
 
     // Log certificates to console and check they are parsable with OpenSSL.
     let cert_endorsement_sk = SecretKey::<NistP256>::read_pkcs8_der_file(cert_endorsement_ecc_sk)?;
     let uds_cert_bytes = parse_and_endorse_uds_cert(uds_tbs_cert_bytes, &cert_endorsement_sk)?;
+    let tpm_ek_cert_bytes =
+        parse_and_endorse_uds_cert(tpm_ek_tbs_cert_bytes, &cert_endorsement_sk)?;
```
pamaury, line 275: maybe `parse_and_endorse_uds_cert` should be rename now that it is used for other certificates than the UDS?

pamaury:
> One thing I realized (though it is ok to address in a follow up PR): the HJSON definitions of each TPM cert are almost identical, with the exception of the subject common name. If we made that a variable, then couldn't we just define one TPM HJSON cert template and instantiate the 3 different versions (EK, CIK, CEK)? Or would the variability in the sized of the subject common name's cause an issue? @pamaury

Absolutely, the generator has no problem handling variable sizes, just make sure the variable in the template is defined with the maximum size possible. It would also save some code size so probably nice to have.

pamaury (approved):
Looks good to, couple of minor things.

sw/host/provisioning/ft_lib/src/lib.rs
```diff
@@ -247,27 +249,103 @@ pub fn run_ft_personalize(
...
     endorsed_certs.send(&*uart)?;
-    let _ = UartConsole::wait_for(&*uart, r"Imported DICE UDS certificate.", timeout)?;
-
-    Ok(())
+    let _ = UartConsole::wait_for(&*uart, r"Final hash:", timeout)?;
+    let received_hash = SerdesSha256Hash::recv(&*uart, timeout, false)?;
+    let _ = UartConsole::wait_for(&*uart, r"Finished importing certificates.", timeout)?;
+
+    // Compare host and device generated certificates' hash values.
+    if received_hash
```
pamaury, line 334: I think for debugging it would be useful to construct the two hashs in array and then compare them. Also zipping them seems a bit overkill. Shouldn't the follow just work?
```rust
let actual_hash = hash.finalize();
if data.as_bytes() == actual_hash.as_slice() { // I think the as_slice might be necessary but I didn't check the types
}
```

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -274,25 +315,87 @@ static status_t personalize_dice_certificates(ujson_t *uj) {
       dice_certs.cdi_1_certificate, &dice_certs.cdi_1_certificate_size));
   TRY(flash_ctrl_info_write(
       &kFlashCtrlInfoPageCdi1Certificate,
-      /*offset=*/0, dice_certs.cdi_1_certificate_size / sizeof(uint32_t),
+      /*offset=*/0, size_to_words(get_cert_size(dice_certs.cdi_1_certificate)),
       dice_certs.cdi_1_certificate));
   LOG_INFO("Generated CDI_1 certificate.");
 
+  {
```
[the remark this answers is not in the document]
vbendeb: some editing leftover, removed.

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -274,25 +315,87 @@ static status_t personalize_dice_certificates(ujson_t *uj) {
...
+    hmac_digest_t pubkey_id;
+    attestation_public_key_t pubkey;
```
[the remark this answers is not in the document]
vbendeb: done

sw/device/silicon_creator/manuf/keys/fake/fake_ca.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
vbendeb: done

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -274,25 +315,87 @@ static status_t personalize_dice_certificates(ujson_t *uj) {
...
+  // Now TPM certificates, get saved in the same flash INFO page.
+  uint32_t page_offset = 0;
+  size_t i;
+  unsigned char *tpm_certs[] = {
+      endorsed_certs.tpm_ek_certificate,
+      endorsed_certs.tpm_cek_certificate,
+      endorsed_certs.tpm_cik_certificate,
+  };
+
+  for (i = 0; i < ARRAYSIZE(tpm_certs); i++) {
```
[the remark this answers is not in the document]
vbendeb: done

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -274,25 +315,87 @@ static status_t personalize_dice_certificates(ujson_t *uj) {
...
+    const char *names[] = {"EK", "CEK", "CIK"};
```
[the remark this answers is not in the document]
vbendeb: why? It is not used outside the loop, since it's const I don't think there is any additional overhead.

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/BUILD
```diff
@@ -221,6 +221,7 @@ opentitan_binary(
     linker_script = "//sw/device/lib/testing/test_framework:ottf_ld_silicon_creator_slot_a",
     deps = [
         "//sw/device/lib/crypto/drivers:entropy",
+        "//sw/device/lib/crypto/impl/sha2:sha256",
```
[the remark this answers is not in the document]
vbendeb: done

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
vbendeb: Buffer pointer passed to flash_crtl_info_read() also needs to be updated, I think having this variable and a comment makes the code a bit better readable.

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
vbendeb: amazing, it even prints it the same as the host :)
done

vbendeb:
> > One thing I realized (though it is ok to address in a follow up PR): the HJSON definitions of each TPM cert are almost identical, with the exception of the subject common name. If we made that a variable, then couldn't we just define one TPM HJSON cert template and instantiate the 3 different versions (EK, CIK, CEK)? Or would the variability in the sized of the subject common name's cause an issue? @pamaury
> 
> Absolutely, the generator has no problem handling variable sizes, just make sure the variable in the template is defined with the maximum size possible. It would also save some code size so probably nice to have. Great idea!

There are in fact two different TPM certificate forms, CEK and CIK certificates do not include some extensions TPM EK has. This could be consolidated, but would require some fine tuning, let's leave it out of this PR.

sw/host/provisioning/ft_lib/src/lib.rs
vbendeb: after changing the hashing function on the device the generated hash endianness needs to change, which is easier when using this zipping pattern. Also, if hashes don't match the hash received from the device is printed out later.

vbendeb:
> Also, is the cert chain validation step on the host side missing? or were you planning on adding that in a follow up PR? @vbendeb

yes, I mean to do this in the follow up PR. We also need a check that the device can re-generate the private key when owner FW is runing.

vbendeb:
I went through the comments and made modifications, please take another look.

vbendeb:
while working on openssl chain validation I fouond a bug in size calculation when hashing certificate contents, PR  updated.

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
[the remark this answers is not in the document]
timothytrippel: ah sorry, miss parsed.

timothytrippel (approved):
LGTM, thanks @vbendeb !

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -205,6 +215,81 @@ static void compute_keymgr_owner_binding(manuf_certgen_inputs_t *inputs) {
...
+  // Read the first word of the certificate.
+  TRY(flash_ctrl_info_read(page, offset, 1, buffer));
+  bytes_read = sizeof(uint32_t);  // One word has been read.
+  cert_size = get_cert_size(buffer);
+  if (cert_size == 0) {
+    LOG_ERROR("Inconsistent certificate header %02x %02x page %x:%x", buffer[0],
+              buffer[1], page->base_addr, offset);
+    return DATA_LOSS();
+  }
+  if (cert_size > sizeof(buffer)) {
```
timothytrippel, line 271: Sorry one other thing: I think you want to use `ARRAYSIZE()` here: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/base/macros.h;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=45 no?

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
timothytrippel: ah sorry missed that.

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
timothytrippel: never mind. pls disreged. cert_size is bytes b/c it is coming out of the ASN1 directly.

Issue #22917, opened 2024-05-01
[test-triage] chip_sw_aes_masking_off fails after AES RTL changes


antmarzam:
### Hierarchy of regression failure

Block level

### Failure Description

UVM_FATAL @ 3321.421894 us: (chip_sw_aes_masking_off_vseq.sv:162) [uvm_test_top.env.virtual_sequencer.chip_sw_aes_masking_off_vseq] Check failed state_init_share1 == '0 (261008556511606517935595337728332530022 [0xc45c6b6290143840c51c1bbea8020166] vs 0 [0x0]) Share 1 of the input not equal zero.

There's also a TL-UL failing assertion:
../src/lowrisc_tlul_common_0.1/rtl/tlul_assert.sv", 301: tb.dut.top_earlgrey.u_xbar_main.tlul_assert_host_rv_core_ibex__cored.gen_assert_final[1].noOutstandingReqsAtEndOfSim_A: started at 3321421894ps failed at 3321421894ps
	Offending '((pend_req[1].pend == 1'b0) || $test$plusargs("disable_assert_final_checks"))'
UVM_ERROR @ 3321.421894 us: (tlul_assert.sv:301) [ASSERT FAILED] noOutstandingReqsAtEndOfSim_A

### Steps to Reproduce

- GitHub Revision: [0cb61fc7e7](https://github.com/lowrisc/opentitan/tree/0cb61fc7e7b9f6e2a29de3de0671c2605964ee86)
- dvsim invocation command to reproduce the failure, inclusive of build and run seeds:
./util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_aes_masking_off --build-seed 29629349767786988748941369645310183062873507656225682712521573681396210883738  --fixed-seed 50872679030684457411525866978198029813162651732304102301881216380340976679238

### Tests with similar or related failures

Fixed Seed `62014364074305549766088787075294037528818904623409476815904894661659903874324` was also failing

vogelpi:
I've now filed a PR to get this test going again, see https://github.com/lowRISC/opentitan/pull/22930.

PR #22930, opened 2024-05-02
[aes/sw] Load all-zero vector into PRNG buffer stage for masking off

vogelpi:
The input masks are now taken from the PRNG buffer stage. This means after loading the magic seed into the PRNG and having the PRNG output the all-zero vector, the buffer stage needs to be updated as well. This can be achieved by triggering via a clearing operation of the data output registers.

This resolves lowRISC/OpenTitan#22917.

### DIFF
--- a/sw/device/sca/aes_serial.c
+++ b/sw/device/sca/aes_serial.c
@@ -776,8 +776,13 @@ bool test_main(void) {
     LOG_INFO("Initializing entropy complex.");
     CHECK_STATUS_OK(aes_testutils_masking_prng_zero_output_seed());
     CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerPrngReseed));
+    bool idle = false;
+    do {
+      CHECK_DIF_OK(dif_aes_get_status(&aes, kDifAesStatusIdle, &idle));
+    } while (!idle);
   }
 #endif
+  CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerDataOutClear));
 
   LOG_INFO("Starting simple serial packet handling.");
   while (true) {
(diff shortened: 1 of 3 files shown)


nasahlpa (approved):
Thanks, LGTM!

sw/device/sca/aes_serial.c
```diff
@@ -776,8 +776,13 @@ bool test_main(void) {
     LOG_INFO("Initializing entropy complex.");
     CHECK_STATUS_OK(aes_testutils_masking_prng_zero_output_seed());
     CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerPrngReseed));
+    bool idle = false;
+    do {
+      CHECK_DIF_OK(dif_aes_get_status(&aes, kDifAesStatusIdle, &idle));
+    } while (!idle);
   }
 #endif
+  CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerDataOutClear));
```
rswarbrick, line 785: I notice that the equivalent line in `aes_sca.c` is skipped for EnglishBreakfast. Is putting this line outside of the if/endif intentional?

rswarbrick (approved):
Looks sensible, with a question about aes_serial.c

sw/device/sca/aes_serial.c
vogelpi: Yes it is. Only `aes_serial.c` runs on English Breakfast (where we use a different way to 0 masks because there is no CSRNG). `aes_sca.c` only runs on the actual chip and the CW310/340.

vogelpi:
Thanks for your reviews. There are some CI FPGA failures but this is all spurious. The code in this PR is NOT run on the FPGA in any form in CI. I am thus merging this.

PR #22931, opened 2024-05-02
[doc] Fix explanations of the wr_en signal

### LINKED ISSUE #22911: regtool: reg_top triggers write error when not writing to a permitted byte, instead of when writing to a forbidden byte
https://github.com/lowRISC/opentitan/blob/e9925bdf1464e9cec8ab8931c108cee4acb2709b/util/reggen/reg_top.sv.tpl#L655-L659

Shouldn't this code check `|(reg_be & ~PERMIT[i])`, not `|(PERMIT[i] & ~reg_be)`?
I.e., "if a byte is being written (`reg_be[j]` = 1) and it is not permitted (`PERMIT[i][j]` = 0, or `~PERMIT[i][j]` = 1), then there's an error."

Currently the error signal seems to be activated when any of the permitted bytes is not being written, instead of when any of the non-permitted bytes is being written.

rswarbrick:
This fixes #22911, and the root of the confusion that led to that issue was a backwards comment that I added in 2021. Oops!

The first commit in the PR fixes the backwards comment (to avoid confusing anyone else). The second commit extends the regtool documentation a little bit to explain why the signals work like this.

### DIFF
--- a/util/reggen/README.md
+++ b/util/reggen/README.md
@@ -640,8 +640,10 @@ Other error responses include for the following reasons:
 
 * TL-UL `a_opcode` illegal value
 * TL-UL writes of size smaller than register size
-  * I.e. writes of size 8b to registers > 8b will cause error (explicitly: if it has field bits within `[31:08]`)
-  * I.e. writes of size 16b to registers > 16b will cause error (explicitly: if it has field bits within `[31:16]`)
+  * Disallowing such writes makes the implementation simpler.
+  * It also strengthens the security of the system, by making it hard to use a fault injection to cause only part of a register to be updated.
+  * This can be seen with an 8b write to a register which is larger than 8 bits, or a 16b write to a register which is larger than 16 bits.
+  * The width of the register is determined by the location of its fields.
 * TL-UL writes of size smaller than 32b that are not word-aligned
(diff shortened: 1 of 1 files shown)


util/reggen/reg_top.sv.tpl
```diff
@@ -652,9 +652,9 @@ ${finst_gen(sr, field, finst_name, fsig_name, fidx)}
 
   % if regs_flat:
 <%
-    # We want to signal wr_err if reg_be (the byte enable signal) is true for
-    # any bytes that aren't supported by a register. That's true if a
-    # addr_hit[i] and a bit is set in reg_be but not in *_PERMIT[i].
+    # We want to signal wr_err for a write if the write is not enabled for every byte supported by
+    # its target register. That happens for register i if if addr_hit[i] and a bit is set in
+    # *_PERMIT[i] (so there is a field covering this bit) but the bit is not set in reg_be.
```
cousteaulecommandant, line 658: Maybe it's helpful to add a SystemVerilog comment as well, in case someone wants to have a look at the generated file?  Like
```sv
// Ensure all relevant bytes are written (partial writes are not supported)
```

PR #22936, opened 2024-05-02
[crypto/hmac] Add unit tests for HMAC driver

wettermo:
This PR adds unit tests for each function in the HMAC crypto library driver. 
It uses `sw/device/silicon_creator/lib/drivers/hmac_unittest.cc` as a basis and changes it according to the style of `sw/device/lib/crypto/impl/keyblob_unittest.cc`.


jadephilipoom (approved):
LGTM, thanks!

PR #22937, opened 2024-05-02
[spi_device] Track intercept_en along the read pipeline

a-will:
intercept_en is used to select between the passthrough and internal read paths, but when the fast read pipeline is enabled, this selector needs to follow the data values that it is meant to select. Add intercept_en to the set of signals that pass through the read pipeline, so the correct data can be selected on the cycle in which it is used.

Without this change, transitions from the mailbox to passthrough may cause passthrough data to appear when the last mailbox data should be transmitted.


antmarzam (approved):
LGTM - Thank you! Very quick turnaround as well :)

a-will:
The failures are unrelated.

hcallahan-lowrisc (approved):

Issue #22941, opened 2024-05-02
[rom] The ROM_EXT selection is based on the security version


cfrantz:
### Description

The ROM currently uses only the `security_version` field in the manifest to select which ROM_EXT to boot.

This makes it impossible to build and distribute a new ROM_EXT that doesn't advance the security version.  This is at odds with some testing and roll-out models where a new ROM_EXT (with new features or bugfixes) is created and rolled out as a test before rolling out a version that protects against rollback.

The ROM should use the manifest version fields to select the preferred ROM_EXT to boot.  The security version should be used to prevent rollback to known-bad ROM_EXTs.

PR #22942, opened 2024-05-02
[rom_ext] Advance the security version in boot_data

cfrantz:
1. When the security version of the just-booted ROM_EXT is greater than the ROM_EXT minimum security version in boot data, move the value forward.
2. Provide a mechanism in the manifest to prevent moving foward even though the version is newer.  This is to release test ROM_EXTs without running afoul of #22941.

This should NOT be cherry-picked to master.  We will fix this on master by adjusting the selection criteria as discussed in #22941.

### DIFF
--- a/rules/manifest.bzl
+++ b/rules/manifest.bzl
@@ -112,7 +112,7 @@ def _manifest_impl(ctx):
         mf["extensions"] = [
             "spx_key",
             "spx_signature",
-            None,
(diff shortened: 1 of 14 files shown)


timothytrippel (approved):

moidx (approved):

cfrantz:
The test failure is unrelated to this change.

Issue #22877, opened 2024-04-29
[keymgr] Reseed interval timer modfications


vogelpi:
### Description

While debugging a ROM_EXT issue #22819, we've noticed that the reseed interval timer inside keymgr is active all the time and doesn't take into account whether keymgr is actually busy.

This means, once software has started keymgr, it will continuously request fresh entropy from EDN0 (by default 64 bits every 256 clock cycles). What's worse, if software decides to lock the [reseed interval register](https://opentitan.org/book/hw/ip/keymgr/doc/registers.html#reseed_interval_shadowed), there is no way to stop this afterwards. The max interval is 65536, meaning every 0.65 ms, a reseed is happening anyway. Because the reseed frequency is then not tied to activity, there is no way to program this adequately.

I am proposing to change the design as follows: The counter should not count the clock cycles but instead the number of PRNG updates. The PRNG is currently updated using `ctrl_lfsr_en | data_lfsr_en | sideload_lfsr_en`. I am suggesting to just feed this signal also into the reseed interval timer.

Any objections @moidx @johannheyszl @andreaskurth ?

johannheyszl:
sgtm.
cc @ballifatih

vogelpi:
Thanks for the feedback @johannheyszl and @ballifatih .

`reg2hw.start.q` was actually low in the waves I saw. keymgr de-asserts that after running an operation. So, using keymgr once and then leaving alone will trigger this behavior. I think we should investigate how to resolve / workaround the DV issue. A simple solution could be to not make the reseed counter count when start is disabled, i.e., after running a command also stop the counter.

ballifatih:
I agree that it would be great if we can drop `~reg2hw.start.q` from `seed_en_i` signal. Then, we can use `reg2hw.start.q` to gate the counter.

This line of code was introduced very early on #4448, I had a look at the PR, but I could not figure out what it actually prevents on DV side.

vogelpi:
Thanks for digging up that old PR - it seems I also reviewed that :-D 

I've now removed that `~reg2hw.start.q` qualifier for the seed enable and didn't get additional failures in block-level DV. I'll remove that as part of the PR. I think I'll try to do an implementation counting the number of LFSR updates as this is a more accurate measurement of how "old" the previous seed is. How, long KeyMgr operations take (as measured by  `~reg2hw.start.q`) can depend on other factors as well including the entropy complex (in case KMAC needs to perform a reseed operation as well).

[...]

PR #22944, opened 2024-05-02
[rom_ext] Relax EDN0 configuration.

moidx:
This change aims to mitigate some of the side effects caused by lowRISC/opentitan#22877.

1. Increase the reseed interval in EDN0 to reduce the number of requests going to entropy_src.
2. Increase the reseed interval in keymgr to reduce the number of requests going to edn0.

Both configuration values will be updated later on based on recommendations derived from penetration testing and pre-silicon evaluation.

### DIFF
--- a/sw/device/lib/crypto/drivers/entropy.c
+++ b/sw/device/lib/crypto/drivers/entropy.c
@@ -246,7 +246,7 @@ static const entropy_complex_config_t
                 .edn0 =
                     {
                         .base_address = kBaseEdn0,
-                        .reseed_interval = 32,
+                        .reseed_interval = 128,
                         .instantiate =
                             {
                                 .id = kEntropyDrbgOpInstantiate,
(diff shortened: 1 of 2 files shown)


timothytrippel (approved):

sw/device/lib/crypto/drivers/entropy.c
```diff
@@ -246,7 +246,7 @@ static const entropy_complex_config_t
                 .edn0 =
                     {
                         .base_address = kBaseEdn0,
-                        .reseed_interval = 32,
+                        .reseed_interval = 0x10000000,
```
vogelpi, line 249: This seems very aggressive and I am not sure if it would be better to leave it as is.

With the current generate length of 32 128-bit block and the max reseed interval for keymgr, keymgr will request 64 bits every 65535 clock cycles. At 100 MHz, that's 0.65 ms. So 1 Generate will last for 32 * 128 / 64 * 0.65ms = 41.6 ms. And 32 Generates will last for 1.3 seconds. This is already longer than some tests run for.

I think with 0x1000000 we are already in the years range.

vogelpi (approved):
Thanks for the PR, the keymgr change looks good, the reseed interval change seems a bit too aggressive to me. Most tests might not see a single reseed on EDN0 with this.

sw/device/lib/crypto/drivers/entropy.c
vogelpi: This sounds good to me, thanks @moidx .

vogelpi: BTW, I have now created a PR to change the hardware https://github.com/lowRISC/opentitan/pull/22951
But I still need to run the full regression and there might be some TLT fallout. So we shouldn't wait with this PR here.

PR #22946, opened 2024-05-02
[keymgr] Abort into Invalid when Creator Root Key is not valid during initialization

### LINKED ISSUE #17776: [test-triage, keymgr] KMAC request inputs do not match

Block level

### Failure Description

```
UVM_ERROR (keymgr_scoreboard.sv:1011) [scoreboard] Check failed act != exp (* [*] vs * [*]) cdi_type: Attestation has 3 failures:
Test keymgr_kmac_rsp_err has 3 failures.
2.keymgr_kmac_rsp_err.2066803726
Line 231, in log /container/opentitan-public/scratch/os_regression/keymgr-sim-vcs/2.keymgr_kmac_rsp_err/latest/run.log 

UVM_ERROR @ 209262498 ps: (keymgr_scoreboard.sv:1011) [uvm_test_top.env.scoreboard] Check failed act != exp (3448233742551261233350829039715943372291253566479285849790124933425161349841910563173666674697844162097279452497933417303135164423955581644432887423676992950725997964640470922335688219765526800548978811829166820560503504502464571941177108845978525249654317321220978727586821025825863080101494279265254667601687257966967121625515820888366008035314717776276455917157881911666113789762834026365826720842116093422499135407033351 [0x7c9fd2cc40f73072d2537475ceb57077867ef72b01e440e80b14596565c1d44d3a0a6d73cd50897de4d744bd65ebdb3837ea77087d878651c517c18a5742b2f9e2ff13b9c2e5ca370bafd64b3e25b1ab96dfebe0c3a5b28e155ca633f34d14061d1022bdad5190ba27206340bb548155d0cf21769a5ac3545253e457375fa4b826db834d11c8fe6e9165c12f6bb2a15a0ff55132e7481cebacf2f653dbc8c0fc3113c8904863625d3c8a023addac8c07] vs 3448233742551261233350829039715943372291253566479285849790124933425161349841910563173666674697844162097279452497933417303135164423955581644432887423676992950725997964640470922335688219765526800548978811829166820560503504502464571941177108845978525249654317321220978727586821025825863080101494279265254667601687257966967121625515820888366008035314717776276455917157881911666113789762834026365826720842116093422499135407033351 [0x7c9fd2cc40f73072d2537475ceb57077867ef72b01e440e80b14596565c1d44d3a0a6d73cd50897de4d744bd65ebdb3837ea77087d878651c517c18a5742b2f9e2ff13b9c2e5ca370bafd64b3e25b1ab96dfebe0c3a5b28e155ca633f34d14061d1022bdad5190ba27206340bb548155d0cf21769a5ac3545253e457375fa4b826db834d11c8fe6e9165c12f6bb2a15a0ff55132e7481cebacf2f653dbc8c0fc3113c8904863625d3c8a023addac8c07]) cdi_type: Attestation
  
   DiversificationKey act: 0xff55132e7481cebacf2f653dbc8c0fc3113c8904863625d3c8a023addac8c07, exp: 0xff55132e7481cebacf2f653dbc8c0fc3113c8904863625d3c8a023addac8c07
   RomDigest act: 0xd0cf21769a5ac3545253e457375fa4b826db834d11c8fe6e9165c12f6bb2a15a, exp: 0xd0cf21769a5ac3545253e457375fa4b826db834d11c8fe6e9165c12f6bb2a15a
   HealthMeasurement act: 0x1d1022bdad5190ba27206340bb548155, exp: 0x1d1022bdad5190ba27206340bb548155
```

In this particular test, OTP root key is not valid while loading the key. Keymgr ends up loading invalid key (all zeroes and all 1s), but it does not move to Disabled state. It accepts further advance operations. I am not sure if this is relevant for M2.5.

### Steps to Reproduce

- GitHub Revision: [af46a9fd1](https://github.com/lowrisc/opentitan/tree/af46a9fd1429e67a213338b5eec28a93c557c708)
  ./util/dvsim/dvsim.py hw/ip/keymgr/dv/keymgr_sim_cfg.hjson -i keymgr_kmac_rsp_err --build-seed 2952949999 --fixed-seed 2066803726 --waves fsdb

- [ ] keymgr_stress_all_with_rand_reset
- [ ] keymgr_stress_all

andreaskurth:
Prior to this PR, keymgr could be initialized and would open itself
to operations even if the Creator Root Key shares provided by OTP in the
`root_key_i` input were invalid (although keymgr did write all-zeros and
all-ones instead of the input from OTP into its key shares if that was
the case).  However, the specification dictates that kemygr must signal
an invalid operation, though it did not specify the next state.

This PR fixes the problem by letting keymgr raise an *invalid
operation* error in this case and then transition to the terminal
Invalid state (going through a wipe state to clear any secrets with
entropy).  The rationale for this implementation is that the Creator
Root Key is meant to be static after reset, so SW doesn't have to be
able to retry the initialization.  Letting keymgr abort into a terminal
state thus is an additional HW security countermeasure to deter FI
attacks aimed at loading a manipulated Creator Root Key.

This PR fixes that behavior in its second-last commit, thereby closes https://github.com/lowRISC/opentitan/issues/17776.

This PR contains a second RTL fix in its last commit:
The keymgr specification states that no operation is allowed in the
Invalid state.  Prior to that commit, keymgr did not raise an *invalid
operation* error, raise an alert, nor update the `OP_STATUS` and
`ERR_CODE` CSRs.

The last commit fixes keymgr to do all these things.

Only the last two commits contain RTL changes, which are relatively simple. Those commits also align the DV. All commits before are fixes and extensions to DV and SW that are necessary and sufficient to prepare for the RTL changes. Please see the commit messages for details.

## TODO Items (pre-merge)

- [x] Run related TLTs and check their result.
  --> I ran `util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_keymgr_{key_derivation{,_prod,_jitter_en{,_reduced_freq}},sideload_{kmac,aes,otbn}} -rx 2` and got results that are in-line with those reported in the current nightly regression results: there are some failures but only in `(chip_sw_keymgr_key_derivation_vseq.sv:122) [uvm_test_top.env.virtual_sequencer.chip_sw_keymgr_key_derivation_vseq] wait timeout occurred!` (as in the nightlies).
- [x] Debug and fix failing CW310 ROM test.
- [x] Update documentation.
- [x] Check if other steps in the key ladder need similar fix.
  --> I think they don't. The inputs to the other key ladder steps don't have _valid_ signals associated with them (which could be improved in a future revision) but they get checked by `keymgr_input_checks` if they aren't all-zeros or all-ones. The result of that check then feeds into `keymgr_ctrl`'s `op_err`, which prevents the key ladder from advancing and any keys from being updated, and the error gets reported as _invalid input_ to the `err_code` CSR.
- [x] Split RTL commit because it fixes two orthogonal problems in keymgr.

### DIFF
--- a/hw/ip/keymgr/rtl/keymgr_ctrl.sv
+++ b/hw/ip/keymgr/rtl/keymgr_ctrl.sv
@@ -633,7 +633,7 @@ module keymgr_ctrl
       end
 
       StCtrlInvalid: begin
-        op_req = op_start_i;
+        invalid_op = op_start_i;
         invalid = 1'b1;
       end
 
--- a/hw/ip/keymgr/rtl/keymgr_ctrl.sv
+++ b/hw/ip/keymgr/rtl/keymgr_ctrl.sv
@@ -518,7 +518,7 @@ module keymgr_ctrl
       StCtrlRootKey: begin
         init_o = 1'b1;
         initialized = 1'b1;
-        state_d = en_i ? StCtrlInit : StCtrlWipe;
+        state_d = (en_i && root_key_valid_q) ? StCtrlInit : StCtrlWipe;
       end
 
       // Beginning from the Init state, operations are accepted.
--- a/hw/ip/keymgr/doc/theory_of_operation.md
+++ b/hw/ip/keymgr/doc/theory_of_operation.md
@@ -238,18 +238,23 @@ The error is reported in [`FAULT_STATUS`](registers.md#fault_status) and the key
 When a fatal error is encountered, the key manager transitions to the `Invalid` [state](#invalid-entry-wiping).
 The following are a few examples of when the error occurs and how the key manager behaves.
 
-#### Example 1: Fault During Operation
+#### Example 1: Fault During Initialization
+The key manager is in the `Reset` state and receives an advance operation.
+After it has reseeded its internal PRNG with entropy, it will try to load the Creator Root Key from OTP.
+If not both shares of the Creator Root Key are valid at that point, key manager will wipe its secrets and transition to the `Invalid` state.
+
+#### Example 2: Fault During Operation
 The key manager is running a generate operation and a non-onehot command was observed by the KMAC interface.
 Since the non-onehot condition is a fault, it is reflected in [`FAULT_STATUS`](registers.md#fault_status) and a fatal alert is generated.
 The key manager transitions to `Invalid` state, wipes internal storage and reports an invalid operation in [`ERR_CODE.INVALID_OP`](registers.md#err_code).
 
-#### Example 2: Fault During Idle
+#### Example 3: Fault During Idle
 The key manager is NOT running an operation and is idle.
 During this time, a fault is observed on the regfile (shadow storage error) and FSM (control FSM integrity error).
 The faults are reflected in [`FAULT_STATUS`](registers.md#fault_status).
 The key manager transitions to `Invalid` state, wipes internal storage but does not report an invalid operation.
 
-#### Example 3: Operation after Fault Detection
+#### Example 4: Operation after Fault Detection
 Continuing from the example above, the key manager now begins an operation.
 Since the key manager is already in `Invalid` state, it does not wipe internal storage and reports an invalid operation in [`ERR_CODE.INVALID_OP`](registers.md#err_code).
 
@@ -259,6 +264,7 @@ What is considered invalid input changes based on current state and operation.
 
 When an advance operation is invoked:
 - The internal key is checked for all 0's and all 1's.
+- During the `Reset` state, both shares of the the Creator Root Key provided by OTP are checked to be valid.
 - During `Initialized` state, creator seed, device ID and health state data is checked for all 0's and all 1's.
 - During `CreatorRootKey` state, the owner seed is checked for all 0's and all 1's.
 - During all other states, nothing is explicitly checked.
--- a/hw/ip/keymgr/dv/env/keymgr_env_pkg.sv
+++ b/hw/ip/keymgr/dv/env/keymgr_env_pkg.sv
@@ -72,8 +72,8 @@ package keymgr_env_pkg;
       keymgr_pkg::keymgr_working_state_e current_state);
 
     uint next_state = int'(current_state) + 1;
-    if (next_state >= int'(keymgr_pkg::StDisabled)) begin
-      return keymgr_pkg::StDisabled;
+    if (next_state > int'(keymgr_pkg::StDisabled)) begin
+      return current_state;
     end else begin
       `downcast(get_next_state, next_state, , , msg_id);
     end
--- a/hw/ip/keymgr/dv/env/keymgr_if.sv
+++ b/hw/ip/keymgr/dv/env/keymgr_if.sv
@@ -229,6 +229,7 @@ interface keymgr_if(input clk, input rst_n);
                                               );
     keymgr_env_pkg::key_shares_t trun_key_shares = {key_shares[1][keymgr_pkg::KeyWidth-1:0],
                                                     key_shares[0][keymgr_pkg::KeyWidth-1:0]};
+    if (state == keymgr_pkg::StInvalid) return;
     case (dest)
       keymgr_pkg::Kmac: begin
         if (kmac_sideload_status != SideLoadClear) begin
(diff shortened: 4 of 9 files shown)


hw/ip/keymgr/rtl/keymgr_ctrl.sv
```diff
@@ -518,7 +519,7 @@ module keymgr_ctrl
       StCtrlRootKey: begin
         init_o = 1'b1;
         initialized = 1'b1;
-        state_d = en_i ? StCtrlInit : StCtrlWipe;
+        state_d = (en_i && root_key_valid_d) ? StCtrlInit : StCtrlWipe;
```
vogelpi, line 522: I think you want to use the _q here, i.e., the synchronized version of the signal no?

vogelpi (commented):
Thanks for working on this! It's quite amazing that the RTL is basically a one line change and the hard work is on the DV side.

hw/ip/keymgr/rtl/keymgr_ctrl.sv
andreaskurth: Ah, right. I originally used `root_key_valid_d` because this FSM state (`StCtrlRootKey`) drives the mux selector to update `key_state_d` to the _non-synced_ `root_key_i.creator_root_key_share{0,1}` in this cycle (via `init_o` and `update_sel`). However, that mux also takes `root_key_valid_q` into account, so the FSM should do the same (also from a CDC perspective).

What is not yet clear to me is the CDC on `root_key_i.creator_root_key_share{0,1}`: They don't get sync'ed and they are multi-bit signals, so syncing would be tricky anyway. As long as the key shares are stable once the valid bits are high, this should be fine. We should double-check that there are constraints on the key shares *and* the valid bits to propagate from otp_ctrl to keymgr within one clock cycle of the keymgr clock. Through that, we can ensure that once keymgr has synced the AND-ed valid bit, the non-synced key shares are stable and can be flopped into standard flops. Do you agree?

andreaskurth:
> Prior to this PR, keymgr could be initialized and would open itself
to operations even if the Creator Root Key shares provided by OTP in the
root_key_i input were invalid (although keymgr did write all-zeros and
all-ones instead of the input from OTP into its key shares if that was
the case). However, the specification dictates that kemygr must signal
an invalid operation. Since the Creator Root Key is meant to be static after
reset, SW doesn't have to be able to retry the initialization. As a security
defense mechanism, keymgr should abort into the Invalid state in this case.

@moidx @ballifatih just wanted to double-check with you that you agree this this reasoning, since I think we haven't specified the _next state_ part of keymgr's behavior when the Creator Root Key is invalid during initialization yet.

hw/ip/keymgr/rtl/keymgr_ctrl.sv
vogelpi: Because software has full control over when we move into the `StCtrlRootKey` state, I think it is fine to not use synced signals for the keys themselves. I think we may not need  to place constraints on these signals.

vogelpi (approved):
This looks good to me from the RTL side point of view.

hw/ip/keymgr/rtl/keymgr_ctrl.sv
andreaskurth: Okay, sounds good

ballifatih:
> Prior to this PR, keymgr could be initialized and would open itself
> to operations even if the Creator Root Key shares provided by OTP in the
> root_key_i input were invalid (although keymgr did write all-zeros and
> all-ones instead of the input from OTP into its key shares if that was
> the case). However, the specification dictates that kemygr must signal
> an invalid operation. Since the Creator Root Key is meant to be static after
> reset, SW doesn't have to be able to retry the initialization. As a security
> defense mechanism, keymgr should abort into the Invalid state in this case.

I think moving right away to Invalid makes more sense. It also aligns with our earlier decision in keymgr_dpe:

https://github.com/lowRISC/opentitan/blob/f4962c85dd1fb69a26c6fa2aeb9d59d6b832f330/hw/ip/keymgr_dpe/rtl/keymgr_dpe_ctrl.sv#L480-L481

In general I favor Keymgr behaviors that immediately mark invalid/erroneous operations instead of resuming operation with invalid data (as a form of obfuscation). I think the latter behavior was chosen early in the design specs to conceal the results of fault attempts. However, it only prevents shallow attempts by the attacker, and there is always further ways to deduce that fault attempt worked (I.e. by computing the attestation identities or cryptographic keys that is supposed to be same in the absence of a fault).

PR #22950, opened 2024-05-02
[rom] Remove real and fake key targets.

moidx:
Now that ROM keys have moved to OTP, there is no longer a need for maintaining fake and real keys rom targets. The execution environments should be able to handle the key configuration based on the OTP profile. This change contains the following updates:

1. Introduce the //sw/silicon_creator/rom/mask_rom target, replacing the rom_with_fake_keys and rom_with_real_keys targets. Update all dependencies previously consuming the old targets.
2. Reduce the number of bitstreams to only support OTP configurations with either `test_rom` and `mask_rom`. All changes are reflected in //hw/bitstreams and consuming targets.
3. Update DV targets to use a single mask_rom target. This is to ensure it is clear in the DV target configuration which ROM target is being used, without making any claims on the OTP configuration.
4. Leave all other FPGA targets as is. For example fpga_cw310_rom_with_fake_keys after this change means that the bitstream has been configured with the new mask_rom target as well as an OTP image containing fake signing keys.


timothytrippel (approved):

cfrantz (approved):

sw/device/silicon_creator/manuf/lib/BUILD
```diff
@@ -141,7 +141,7 @@ opentitan_test(
     name = "individualize_functest",
     srcs = ["individualize_functest.c"],
     cw310 = cw310_params(
```
pamaury, line 143: I thought `opentitan_test` was supposed to use `otp` and not `bitstream`?

PR #22952, opened 2024-05-03
[dif,alert_handler] Fix dif_alert_handler_get_class_state

matutem:
A state indicating an fsm error was added about 3 years ago, but apparently the dif code was not updated, and the code would flag an error if it ever hits that state. This state indicates some tampering.


moidx (approved):

matutem:
The CI failure is unrelated. Merging this.

Issue #21944, opened 2024-03-11
[rom] Change the base-w representation in SPHINCS+


jadephilipoom:
### Description

Our current SPHINCS+ implementation is from the [reference implementaion](https://github.com/sphincs/sphincsplus) and compatible with the round 3 NIST competition submission. However, the NIST [draft standard](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.ipd.pdf) for SPHINCS+ incorporates one change that makes it backwards-incompatible with round 3 (some brief discussion on the pqc-forum mailing list [here](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/88tuvtb7nN4/m/U_fBkPBMAAAJ)).

This change is implemented in the `consistent-basew` branch of the reference implementation, but has not been merged into the main branch. The change itself is not hard to implement; it's essentially a one-line edit to an internal routine that flips the endianness of some values. We'd also need to update our test vectors.

To me it's frustratingly unclear if we should implement this or not for EarlGrey PROD. On one hand, it's of course preferable to be compatible with the eventual NIST standard, which presumably is what everyone will standardize on when it comes out. On the other, there's some risk that NIST will decide last-minute on some other backwards-incompatible change before finalizing the standard, and then instead of "compatible with round 3 SPHINCS+" we're stuck in a no-man's-land in between round 3 and the final standard where we're compatible with neither (this is why we didn't implement the change for ES, even though it was announced just barely before ES ROM freeze). Admittedly by now the risk of a new change is probably small; I've not heard of other backwards-incompatible changes being considered, the comments period is closed, and the final standards are [expected in 2024](https://www.nist.gov/news-events/news/2023/08/nist-standardize-encryption-algorithms-can-resist-attack-quantum-computers). I want to solicit opinions here about which course we should take; perhaps this would be a good security WG discussion topic as well.

CC @moidx @cfrantz @johannheyszl 

Edit: This issue is now tracking a few SPHINCS+ updates that I want to get in before ROM freeze, so here's a checklist:
- [x] Base-w endianness change (i.e. [this change](https://github.com/sphincs/sphincsplus/commit/129b72c80e122a22a61f71b5d2b042770890ccee))
- [x] Widen the key pair address field (see [here](https://github.com/sphincs/sphincsplus/issues/59) -- doesn't affect the parameters we're using but needed for alternate parameters) - will be addressed by https://github.com/lowRISC/opentitan/pull/22894
[...]

PR #22953, opened 2024-05-03
[rom] Change the endianness of FORS indices for SPHINCS+.

jadephilipoom:
This backwards-incompatible change is part of the changes between the round 3 SPHINCS+ submission to the NIST PQC competition and the SLH-DSA draft standard. Part of https://github.com/lowRISC/opentitan/issues/21944

Corresponds to this commit from the SPHINCS+ reference implementation (`consistent-basew` branch):
https://github.com/sphincs/sphincsplus/commit/129b72c80e122a22a61f71b5d2b042770890ccee

Test vectors were re-generated from the `consistent-basew` branch of the reference implementation.

~Note that I think this might break the cryptotest SPHINCS+ setup, because they are right now hard-coded to fetch a specific self-hosted file for the round3 submission: https://github.com/lowRISC/opentitan/blob/5dbbd252ce8531680fde18361614ed07c783b5c7/third_party/sphincsplus/repos.bzl#L19~
~@RyanTorok @milesdai do either of you know if there is another file there, or how we can update it? I don't think NIST is actually hosting a corresponding KAT file for the draft standard yet, so we could also potentially wait until they release final test vectors (but in that case we should probably add a comment that this setup will fail tests on master, and explain why). The round3 tests are still useful for ES sival.~

Edit: now addressed, thanks Ryan!


RyanTorok:
If we find a new archive with KAT tests for the new version, it would be easy to upload it to the cloud storage and use it (although this would cause the test to break on any branches that do not include this PR).

The archive file in the GCP storage for SPHINCS+ is a bit special compared to the rest of the cryptotest framework. Due to size considerations, the SPHINCS+ archive is manually created by us (which is why the GCP storage is primary, unlike the other algorithms, which use the NIST site as primary and GCP as backup). "Manually created" here just means we un-compressed the archive, removed the KAT tests for all the variations of SPHINCS+ we don't support, and re-compressed it.

That original zip came from the NIST round 3 submission available directly on the [SPHINCS+ site](https://sphincs.org/resources.html), and did not involve the [SPHINCS+ repo](https://github.com/sphincs/sphincsplus) at all. I took a look at the [new branch](https://github.com/sphincs/sphincsplus/tree/consistent-basew) matching the new draft standard, and it does include a `./vectors.py` script that generates the test vectors. However, because it uses a temporary directory to do so, I don't think this can be directly used to hermetically build the test vectors in the OT repo (not to mention there's no release to fetch using an `http_archive` to begin with).

We may be able to create a new archive file to store in the GCP using the temporary files generated by `vectors.py`, but the process to create the archive would be more complicated than simply removing a bunch of irrelevant files.

RyanTorok:
I updated #21681 to switch the source of the test vectors to the new GCP-backed archive for the `consistent-basew` version of SPHINCS+.

moidx (approved):
Thanks @jadephilipoom. I assume this will break the current rom e2e `sigverify_spx` tests. 

Is there an equivalent update in the opentitantool side? 

@cfrantz for visibility as this will also require updates on the HSM integration side.

jadephilipoom:
> Thanks @jadephilipoom. I assume this will break the current rom e2e `sigverify_spx` tests.
> 
> Is there an equivalent update for opentitantool?

Good point. I looked at the `pqcrypto` rust crate we're using, and it looks like it pulls from `PQClean`, which pulls from the `master` branch of the reference implementation :disappointed: so it's hard to just change something "easy" there and update it. There's an [open issue](https://github.com/PQClean/PQClean/issues/526) on PQClean to update to the to-be-standardized version, but the timeline is unclear. We could try switching to a different crate, but the ones I can see that advertise "slh-dsa" don't seem very well-established. Not sure what we should do about this. We could depend on one of the newer crates, since it's just for testing anyway, or we could always hold off on merging this PR until the last minute or disable the tests that would fail until `pqcrypto` updates, but it would be good to run end-to-end tests before ROM freeze.

jadephilipoom:
> Thanks for looking into the host side tooling @jadephilipoom. Should we be considering to do a similar thing to what we did on the device side an maintain our own host-side implementation? It seems that there are still additional changes in the pipeline which are going make it difficult for us to run integration testing.
> 
> Should we raise this issue to the Software WG?

Yes, seems like a good topic for Software WG (and good timing too!)

mundaym:
This issue was discussed in the software working group today. The consensus was that we should write our own Rust wrapper for the reference implementation.

jadephilipoom:
For visibility: I've cherry-picked the last commit from https://github.com/lowRISC/opentitan/pull/23326; this puts the host-side tooling in sync with the device for the endianness swap so that tests work. If it passes CI, I'll just merge this and close #23326,. Since both sets of changes have been reviewed and approved, I don't think it requires any additional review; just wanted to put an explanation for the new commit post-approve.

jadephilipoom:
There are a couple remaining test failures because (I think) some pre-signed binaries need to be manually re-generated. Can anyone tell me what the procedure is for that? I can see that commit c36ad25e8fbe158314dbe57f9f33a1576b71281d re-generated them due to `opentitantool` updates, which looks like the same thing I need, but I can't figure out which commands to run to do that again. @timothytrippel , do you remember?

jadephilipoom:
Update: after talking to @cfrantz and @moidx , I understand that none of the pre-signed binaries should need to be regenerated, and the tests are likely failing because the signatures opentitantool are producing are incorrect. I added a KAT on the opentitantool impl and it failed, so that is definitely the case and I'm working on figuring out what the bug is.

jadephilipoom:
Update: CI result has two failing tests, but both appear to be failing on all CI jobs at the moment and I think they're unrelated.
- `//sw/device/silicon_creator/rom/e2e/boot_policy_flash_ecc_error:a_valid_b_corrupt_manifest_security_version_fpga_cw310_rom_with_fake_keys`
- `//sw/device/tests:uart_smoketest_fpga_cw340_rom_with_fake_keys`

I don't have a CW340, but I can replicate the first one on `master` and both are also failing on https://github.com/lowRISC/opentitan/pull/23598, which only adds test data that's not used by anything. So I think it's safe to say that re-generating the keys fixed all the failures (although I'm not satisfied until I know why).

sw/device/silicon_creator/rom/keys/README.md
```diff
@@ -0,0 +1,27 @@
+## Re-generating ROM keys
+
+The following steps generate a new fake test SPHINCS+ key pair and update the
+various places that reference it. The instructions can be modified to generate
+other types of keys:
+- To generate ECDSA keys, replace every occurrence of `spx` with `ecdsa`.
+- To generate prod or dev keys, replace every occurrenct of `test_key` with `prod_key` or `dev_key`.
+- To generate real keys, use the `keys/real` directory instead of `keys/fake`.
```
moidx, line 8: nit: For real keys we recommend using an offline HSM, so this procedure is only applicable to fake keys.

cfrantz (approved):

sw/device/silicon_creator/rom/keys/README.md
jadephilipoom: Do we have a way of generating those keys ahead of time and checking in simulation that the chip definitely can boot using them, so that we don't end up with a subtle encoding mismatch between the HSM and the ROM?

jadephilipoom:
Fixed to address @moidx's comment on real keys and also to re-generate "unauthorized" keys. I'll consider this OK to merge once CI passes, unless anyone objects.

PR #22957, opened 2024-05-03
[top, fpga] Enable KMAC masking on CW340 FPGA board

vogelpi:
Unlike the CW310, the CW340 has the FPGA resources to enable both AES and KMAC masking. Having the masking enabled gives us better test coverage for ROM_EXT tests running on FPGA. Without KMAC masking, the timing and entropy consumption in the system is less representative.

In addition this PR also disables FPGA-specific SCA optimizations in the KMAC RTL to be as close as possible to the actual chip.


vogelpi:
FYI, once this PR is in. Any FPGA tests involving KMAC and KEYMGR should be run on the CW340 if possible because the CW340 configuration is much more representative of the real chip in terms of timing and entropy consumption.

andreaskurth (approved):
LGTM thx @vogelpi

timothytrippel (approved):
Thanks @vogelpi I added #22971 to track bringing up the manuf flows on the 340 since they make use of these blocks.

vogelpi:
Thanks @timothytrippel for keeping track of moving relevant tests over!

PR #22963, opened 2024-05-03
[usbdev,sival] Enable streaming test

pamaury:
This PR builds on #22409, it cherry picks the USB harness from the sival branch and extends it so it can run the C++ host streaming code and monitor for success.

At the moment, I had to reduce the number of bytes sent in those tests because they are way too long (5min for each test). @alees24 is that fine or should we find a way to make this configurable? I think for CI it should be sufficient to aim for maybe a 30s test?

Still TODO:

- [x] set the number of transferred bytes depending on the target (FPGA vs silicon)

### DIFF
--- a/sw/device/tests/usbdev_stream_test.c
+++ b/sw/device/tests/usbdev_stream_test.c
@@ -35,10 +35,12 @@
 #define NUM_STREAMS USBUTILS_STREAMS_MAX
 #endif
 
-// This takes about 256s presently with 10MHz CPU in CW310 FPGA and physical
+#define TRANSFER_BYTES_SILICON (0x10U << 20)
+// 16MiB takes about 256s presently with 10MHz CPU in CW310 FPGA and physical
 // USB with randomized packet sizes and the default memcpy implementation;
-// The _MEM_FASTER switch drops the run time to 187s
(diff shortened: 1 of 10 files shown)


pamaury:
@alees24 can you double-check the FPGA vs silicon transfer change?

sw/device/tests/usbdev_stream_test.c
```diff
@@ -35,10 +35,11 @@
 #define NUM_STREAMS USBUTILS_STREAMS_MAX
 #endif
 
+#define TRANSFER_BYTES_SILICON (0x10U << 20)
 // This takes about 256s presently with 10MHz CPU in CW310 FPGA and physical
 // USB with randomized packet sizes and the default memcpy implementation;
 // The _MEM_FASTER switch drops the run time to 187s
-#define TRANSFER_BYTES_FPGA (0x10U << 20)
+#define TRANSFER_BYTES_FPGA (0x10U << 16)
```
alees24, line 42: The comment needs modifying since it relates to the _FPGA definition. Suggest replacing 'This takes' with '16MiB takes..' and then extending the comment with an annotation that the transfer size has been reduced to keep CI jobs completing in under 30s.

alees24 (approved):
The changes look fine; just minor twiddles suggested, please.

PR #22964, opened 2024-05-03
[prim] Fix typo'd loop increment

jwnrt:
Happened to see this bit of code that looks like a typo.

Not sure about SystemVerilog, but in C this loop would never terminate. Better safe than sorry.


rswarbrick (approved):
This looks sensible, but it's a bit weird that we haven't found the problem before. I think the reason might be that we don't actually instantiate the module anywhere!

Hunting with `git log -S` doesn't find any hits either. So I'm going to file a follow-up PR that removes the prim entirely. Thanks for finding this.

vogelpi (approved):
This looks good, thanks @jwnrt !

PR #22977, opened 2024-05-05
[manuf] add TPM cert generation to personalization binary 

vbendeb:
This is an attempt to cherry-pick master branch patches submitted under https://github.com/lowRISC/opentitan/pull/22920 and described as
"This patchset adds generatiinig of TPM certificates during personalization, storing all three TPM certificates in a dedicated flash INFO page and then hashing the certificates contents on the device after reading them back from flash, and comparing with the hash calculated on the host."


vbendeb:
This patchset does not pass the test, publishing for review and ducsuccion of the issue.

timothytrippel:
I think we can close this? since I am refactoring this into an ft_personalize_4 binary due to ROM enforced size constraints on the ROM_EXT boot slot.

PR #22989, opened 2024-05-07
[sival,sram_ctrl] Fix scrambled_access ECC error counts

jwnrt:
Previously the code preserved both the `pattern` and `backdoor` fields when we scramble main SRAM. The `pattern` field is the one that we expect to be scrambled and trigger ECC errors. In DV-sim environments the `backdoor` field is magically corrected so that it's scrambled with the new key and thus doesn't trigger ECC errors.
    
This of course does not (and cannot) happen outside of DV sim, so the number of ECC errors was doubled.
    
Fixed by changing the layout of `scramble_test_frame` so that we can skip copying the `backdoor` field outside of DV-sim.

If this description makes no sense, let me know and I'll try fix up the explanation better.


engdoreis (approved):

jwnrt:
CI failure (`uart_baud_rate_test`) unrelated.

PR #22991, opened 2024-05-07
[sival,sram_ctrl] Fix SRAM execution test in ROM_EXT environments

jwnrt:
This test was failing in `ROM_EXT` environments because the ROM_EXT sets the `MMWP` bit in ePMP.

This changes the behaviour for unconfigured ePMP regions from them being ignored (everything allowed) to blocked (everything disallowed).

The test was removing configurations including seemingly for the region that the test was running out of flash from. This caused Ibex to just stop in the middle of the test.

This PR changes this part of the test to change to configuration of the actual region we care about and leave the others alone.

### DIFF
--- a/sw/device/tests/sram_ctrl_execution_test.c
+++ b/sw/device/tests/sram_ctrl_execution_test.c
@@ -75,11 +75,14 @@ bool test_main(void) {
 
   // Unlock the entire address space for RWX so that we can run this test with
   // both rom and test_rom.
-  CSR_SET_BITS(CSR_REG_PMPADDR15, 0x7fffffff);
-  CSR_SET_BITS(CSR_REG_PMPCFG3, kEpmpPermLockedReadWriteExecute << 24);
-  CSR_WRITE(CSR_REG_PMPCFG2, 0);
-  CSR_WRITE(CSR_REG_PMPCFG1, 0);
-  CSR_WRITE(CSR_REG_PMPCFG0, 0);
+  CSR_WRITE(CSR_REG_PMPADDR7, 0x7fffffff);
+
+  uint32_t pmpcfg1 = 0;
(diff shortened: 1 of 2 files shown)


sw/device/tests/sram_ctrl_execution_test.c
```diff
@@ -75,11 +75,9 @@ bool test_main(void) {
 
   // Unlock the entire address space for RWX so that we can run this test with
   // both rom and test_rom.
-  CSR_SET_BITS(CSR_REG_PMPADDR15, 0x7fffffff);
-  CSR_SET_BITS(CSR_REG_PMPCFG3, kEpmpPermLockedReadWriteExecute << 24);
-  CSR_WRITE(CSR_REG_PMPCFG2, 0);
-  CSR_WRITE(CSR_REG_PMPCFG1, 0);
-  CSR_WRITE(CSR_REG_PMPCFG0, 0);
+  CSR_SET_BITS(CSR_REG_PMPADDR7, 0x7fffffff);
```
nbdd0121, line 78: Use `CSR_WRITE` instead

sw/device/tests/sram_ctrl_execution_test.c
```diff
@@ -75,11 +75,9 @@ bool test_main(void) {
...
+  CSR_SET_BITS(CSR_REG_PMPCFG1,
```
nbdd0121, line 79: For pmpcfg1, I recommend do a CSR read to confirm that none of the regions are being configured (i.e. have content 0), and do a `CSR_WRITE`.

nbdd0121 (changes requested):

moidx (approved):

nbdd0121 (approved):

PR #22906, opened 2024-04-30
[csrng/rtl] csrng move cmd checks into the cmd stage

h-filali:
Please see the commit message for a description.

I ran a regression with 0.1 multiplier and a few more tests than usual are failing (~5 more than expected).

I didn't have time to debug all of them but I think this PR is ready to at least be looked at.

depends on #22883

PR #22992, opened 2024-05-07
[csrng/dv] Add DV for the command status responses

h-filali:
This PR adds some basic DV for the recently added status response changes.
For further information check out the individual commit messages.

This PR depends on #22906.

I ran 10% of a full regression multiple times and around 9 tests fail consistently.
None of them are alert tests, but some of the tests that fail don't fail before my changes.
I would have to do some further debugging to get back to the baseline number of failures.
However, I decided to create this PR anyways such that I can move on to some more urgent tasks.

### DIFF
--- a/hw/ip/csrng/dv/env/seq_lib/csrng_alert_vseq.sv
+++ b/hw/ip/csrng/dv/env/seq_lib/csrng_alert_vseq.sv
@@ -16,29 +16,32 @@ class csrng_alert_vseq extends csrng_base_vseq;
   rand bit [3:0]  clen;
   rand bit [11:0] glen;
 
-  task body();
+
+  virtual task reset_csrng_alert();
+
+    `uvm_info(`gfn, $sformatf("Checking RECOV_ALERT_STS reset"), UVM_MEDIUM)
+    // Toggle enable to put main FSM back into legal state.
+    ral.ctrl.enable.set(prim_mubi_pkg::MuBi4False);
+    csr_update(.csr(ral.ctrl));
+    cfg.clk_rst_vif.wait_clks(100);
+    ral.ctrl.enable.set(prim_mubi_pkg::MuBi4True);
+    csr_update(.csr(ral.ctrl));
+
+    // Clear recov_alert_sts register.
+    csr_wr(.ptr(ral.recov_alert_sts), .value(32'b0));
+    cfg.clk_rst_vif.wait_clks(100);
+    // Check recov_alert_sts register has cleared.
+    csr_rd_check(.ptr(ral.recov_alert_sts), .compare_value(0));
+
+  endtask
+
+  virtual task test_field_alerts();
+
     int           first_index;
     string        reg_name;
     string        fld_name;
     uvm_reg       csr;
     uvm_reg_field fld;
-
-    // Values for the CMD_STAGE_INVALID_ACMD_ALERT test.
-    `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(illegal_command, illegal_command inside {INV, GENB,
-                                                                                   GENU};)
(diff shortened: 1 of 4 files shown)


hw/ip/csrng/dv/env/seq_lib/csrng_alert_vseq.sv
```diff
@@ -205,43 +212,181 @@ class csrng_alert_vseq extends csrng_base_vseq;
...
+    // Send the first max_reseed_count generate commands, which should succeed.
+    cs_item.acmd  = csrng_pkg::GEN;
+    cs_item.clen  = 'h0;
+    cs_item.flags = get_rand_mubi4_val(.t_weight(4), .f_weight(4), .other_weight(0));
+    cs_item.glen  = 'h1;
+    for (int i = 0; i < cfg.max_reseed_count; i++) begin
+      send_cmd_req(cfg.which_app_err_alert, cs_item);
+    end
+
+    // Send the last generate command, which should fail.
```
vogelpi, line 344: Here the comment is correct. This second part of the task seems almost identical and it could be factored out into another task which is called twice to reduce duplication. But I guess this is not really urgent.

vogelpi (approved):
Thanks @h-filali , this looks good to me modulo some minor things.

Even if this isn't perfect and the pass rate isn't 100% yet, it's still a good improvement. The fact that now some more tests fail, may indicate that we've found a bug. I think it's fine to merge this before debugging all failures.

hw/ip/csrng/dv/env/seq_lib/csrng_alert_vseq.sv
```diff
@@ -205,43 +212,181 @@ class csrng_alert_vseq extends csrng_base_vseq;
...
+    exp_recov_alert_sts[ral.recov_alert_sts.cmd_stage_invalid_cmd_seq_alert.get_lsb_pos()] = 1;
+    csr_spinwait(.ptr(ral.recov_alert_sts), .exp_data(exp_recov_alert_sts));
+    // Since we already did a backdoor check, sampling with this value is sufficient.
+    cov_vif.cg_recov_alert_sample(.recov_alert(exp_recov_alert_sts));
+
+    reset_csrng_alert();
+
+  endtask
+
+  // Here we set the reseed interval to 1 to see if the corresponding alert is triggered
```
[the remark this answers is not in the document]
h-filali: Oh yes, good catch. Forgot to change this.

hw/ip/csrng/dv/env/seq_lib/csrng_alert_vseq.sv
h-filali: Fair point. I'll factor it out.

hw/ip/csrng/dv/env/seq_lib/csrng_alert_vseq.sv
```diff
@@ -205,43 +212,181 @@ class csrng_alert_vseq extends csrng_base_vseq;
     // Check recov_alert_sts register has cleared.
     csr_rd_check(.ptr(ral.recov_alert_sts), .compare_value(0));
 
+  endtask
+
+  // Here we send an illegal command to CSRNG to check that CMD_STAGE_INVALID_ACMD_ALERT
+  // is triggered. Sending an illegal command does not get a response from CSRNG.
```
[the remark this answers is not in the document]
h-filali: I changed it, thanks.

vogelpi (approved):
LGTM!

hw/ip/csrng/dv/env/seq_lib/csrng_alert_vseq.sv
vogelpi: Thanks!

PR #22994, opened 2024-05-07
[sival,alert] Enable `alert_handler_lpg_sleep_mode_pings` test in sival environments

jwnrt:
* Added the sival test environments, but kept Verilator disabled.
* Changed the platform-dependent handling to a switch/case to make the verilator case clearer.
* Added a case for silicon.

### DIFF
--- a/sw/device/tests/alert_handler_lpg_sleep_mode_pings.c
+++ b/sw/device/tests/alert_handler_lpg_sleep_mode_pings.c
@@ -80,23 +80,31 @@ static void init_peripherals(void) {
  */
 void wait_enough_for_alert_ping(void) {
   // wait enough
-  if (kDeviceType == kDeviceFpgaCw310) {
-    // 2*margin_of_safety*(2**DW)*(1/kClockFreqPeripheralHz)
-    // 2*4*(2**16)*(400ns) = 0.2s
-    busy_spin_micros(1000 * 200);
-  } else if (kDeviceType == kDeviceSimDV) {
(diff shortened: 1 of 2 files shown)


sw/device/tests/alert_handler_lpg_sleep_mode_pings.c
```diff
@@ -80,23 +80,36 @@ static void init_peripherals(void) {
...
+      busy_spin_micros(1000 * 200);
+      break;
+    case kDeviceSilicon:
+      // 2*margin_of_safety*(2**DW)*(1/kClockFreqPeripheralHz)
+      // 2*4*(2**16)*(42ns) = 22ms
+      busy_spin_micros(1000 * 22);
+      break;
+    case kDeviceSimDV:
+      // NUM_ALERTS*2*margin_of_safety*(2**DW)*(1/kClockFreqPeripheralHz)
+      // 2*4*(2**3)*(40ns) = 3us
```
moidx, line 97: Shouldn't this be `(2**16)`? The clock and RTL params configuration between `kDeviceSilicon` and `kDeviceSimDV` should be the same.

jwnrt: Updated

jwnrt:
Failing on the FPGA because of https://github.com/lowRISC/opentitan/issues/23038

jwnrt:
Closed because it’s against an old branch, will need redoing on `earlgrey_1.0.0` when the splicing issue is fixed.

PR #22996, opened 2024-05-07
[sival,alert] Enable `alert_handler_reverse_ping_in_deep_sleep` in sival

jwnrt:
This PR enables the `alert_handler_reverse_pin_in_deep_sleep` test in sival environments.

A change was needed for FPGA environments to work around either the ROM or ROM_EXT always triggering a `FlashCtrlFatalErr` when the chip boots. This alert is ignored.

There are a few tests affected by this alert and we're not 100% sure we want to add the workaround to all of them, so we can hold off on merging this for now.

### DIFF
--- a/sw/device/tests/alert_handler_reverse_ping_in_deep_sleep_test.c
+++ b/sw/device/tests/alert_handler_reverse_ping_in_deep_sleep_test.c
@@ -191,7 +191,37 @@ static void chip_sw_reset(void) {
  * line to the CPU, which results in a call to this OTTF ISR. This ISR
  * overrides the default OTTF implementation.
  */
-void ottf_external_isr(uint32_t *exc_info) { interrupt_serviced = true; }
+void ottf_external_isr(uint32_t *exc_info) {
+  dif_rv_plic_irq_id_t irq_id;
+  CHECK_DIF_OK(dif_rv_plic_irq_claim(&plic, kPlicTarget, &irq_id));
+
+  // Only handle `alert_handler` interrupts.
+  top_earlgrey_plic_peripheral_t peripheral =
+      top_earlgrey_plic_interrupt_for_peripheral[irq_id];
+  CHECK(peripheral == kTopEarlgreyPlicPeripheralAlertHandler);
+
+  // Check whether the alert was caused by the `FlashCtrlFatalErr` alert which
+  // we must ignore on FPGAs.
+  if (kDeviceType == kDeviceFpgaCw310 || kDeviceType == kDeviceFpgaCw340) {
+    bool is_cause;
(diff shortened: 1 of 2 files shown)


sw/device/tests/alert_handler_reverse_ping_in_deep_sleep_test.c
```diff
@@ -191,7 +191,37 @@ static void chip_sw_reset(void) {
...
+  CHECK_DIF_OK(dif_rv_plic_irq_claim(&plic, kPlicTarget, &irq_id));
+
+  // Only handle `alert_handler` interrupts.
+  top_earlgrey_plic_peripheral_t peripheral =
+      top_earlgrey_plic_interrupt_for_peripheral[irq_id];
+  CHECK(peripheral == kTopEarlgreyPlicPeripheralAlertHandler);
+
+  // Check whether the alert was caused by the `FlashCtrlFatalErr` alert which
+  // we must ignore on FPGAs.
+  if (kDeviceType == kDeviceFpgaCw310 || kDeviceType == kDeviceFpgaCw340) {
```
moidx, line 205: Let's look at what error is being triggered here, and discuss how to handle it in a GitHub issue. 

Having an interrupt filter like this at the OTTF level can result in masking real errors in the future, so we should try to avoid this as much as possible.

jwnrt: I'll open a proper issue, but so far I just know `flash_ctrl` has `read_err = 1` in its error code bits and fault bits. The address of the read error is `0x00081800` which I think is bank 1 page 3: `kFlashCtrlInfoPageOwnerSlot1` (based on [this address map](https://opentitan.org/book/hw/ip/flash_ctrl/#default-address-map) and [this header](https://github.com/lowRISC/opentitan/blob/596d86e3d0be7cbc5a71d86f79cc976029c6e493/sw/device/silicon_creator/lib/drivers/flash_ctrl.h#L89)) but I don't know where it's being read in the ROM_EXT or why it fails.

moidx: This particular error seems to be related to the owner blob not being initialized. I believe @cfrantz is looking at defining a weak function mechanism for the FPGA target to initialize this page at boot time. 

I will file a bug to capture some concerns related to info flash page initialization for FPGA targets.

moidx: https://github.com/lowRISC/opentitan/issues/23038

jwnrt:
Closed because it's targeting an old branch. Will need to be recreated for a newer branch once flash splicing is implemented.

PR #22998, opened 2024-05-07
Cherry pick usb

pamaury:
Manual backport for several usbdev commits, replaces #21671 that was closed


jwnrt (approved):
Will need rebasing for the licence header change, but otherwise LGTM.

PR #23002, opened 2024-05-07
[jtag,dv] Lots of tidyups in jtag_driver.sv

rswarbrick:
The main impetus for this came from trying to sort out an rv_dm test which was failing because of incorrect use of a clocking block (see "Remove use of HOST_CB clocking block"), but I spent a while looking at the rather complicated code and trying to tidy it up a bit.

The commit described above should be the only one that changes behaviour (and it changes it for the better!). The others should all be tidyups that make the code clearer without changing behaviour.

### DIFF
--- a/hw/dv/sv/jtag_agent/jtag_driver.sv
+++ b/hw/dv/sv/jtag_agent/jtag_driver.sv
@@ -161,8 +161,8 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
     // ShiftIR
     `HOST_CB.tms <= 1'b0;
     `HOST_CB.tdi <= 1'b0;
+    @(`HOST_CB);
     for(int i = 0; i < len; i++) begin
-      @(`HOST_CB);
       // ExitIR if end of addr
       `HOST_CB.tms <= (i == len - 1) ? 1'b1 : 1'b0;
       `HOST_CB.tdi <= ir[i];
@@ -175,8 +175,8 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
            UVM_MEDIUM)
         jtag_pause(pause_count, dout);
       end
+      @(`HOST_CB);
     end
-    @(`HOST_CB);
     // go to RTI either via
     // - PauseIR -> exit2IR -> UpdateIR -> RTI or
     // - Exit1IR -> UpdateIR -> RTI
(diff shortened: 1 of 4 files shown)


rswarbrick:
Hmm, I'm (embarrassingly) struggling to get this solid enough to get through CI. I'm converting to draft now, and will convert back once I think it's worthy of review!

rswarbrick:
I've just re-pushed to include the contents of #22974, which avoids collecting an empty packet in the rv_dm smoke sequence with seed 1 (and failing a CI check). Merge that first.

rswarbrick:
Right. I believe this is now worthy of reviewing! :-)

hw/dv/sv/jtag_agent/jtag_driver.sv
```diff
@@ -13,9 +13,11 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
 
   `uvm_component_new
 
-  // If the same IR was already selected earlier, then don't resend IR based on seq item knob.
+  // The most recently selected IR. If a new request has the same IR and has the skip_reselected_ir
```
hcallahan-lowrisc, line 16: Based on the discussion we had about this in-person, probably worth stressing the point here that this is 'state' the driver maintains, and the driven traffic may be different depending on it. Maybe it's worth adding a short description to `hw/dv/sv/jtag_agent/README.md` as well?

hw/dv/sv/jtag_agent/jtag_driver.sv
```diff
@@ -79,6 +72,21 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
     end
   endtask
 
+  // Drive TMS, TDI to the given values and wait for a single edge of TCK.
+  task tms_tdi_step(bit tms, bit tdi);
+    // We normally expect this task to be called at the negedge of tck (synchronous with HOST_CB).
```
hcallahan-lowrisc, line 77: Is there any sort of DV_CHECK we can put in place here to test if the task is being called at the right time? Or based on the comment, does that mean we're fairly confident it can now be called safely anytime?

hcallahan-lowrisc (approved):
This seems much cleaner and easier to follow to me. Thanks! (Always nice to see the reds(-) beats the greens(+) too)
Just a couple minor questions about doc clarifications.

hw/dv/sv/jtag_agent/jtag_driver.sv
rswarbrick: I believe that it can be safely called at any time. Most of the time it will be on the negedge of TCK, but I don't think this is guaranteed if `tck_en` was low and now goes high. But it shouldn't matter: it will definitely before the first posedge of TCK after the posedge of `tck_en`.

hw/dv/sv/jtag_agent/jtag_driver.sv
rswarbrick: Good point. I'll probably do this by amending the follow-up PR that adds knowledge of interface disconnections, because I think the explanation would make more sense there. Thanks for the heads up.

hcallahan-lowrisc: Maybe it’s worth using ‘protected’ for state variables, as that highlights to me somewhat that they are internal, and not meant to be, for example, control knobs.

PR #20349, opened 2023-11-15
[RV_DM]JTAG_dtm_idle_hint_vseq

BasitMehmood:
This test verify that jtag debugger does not need to stay any longer than advertized idle time indicated in dtmcs register.
@jdonjdon. kindly review it.Thanks

PR #23003, opened 2024-05-07
[rv_dm,dv] Rewrite dtm_idle_hint_vseq to use DMI helpers

rswarbrick:
This seems less complicated than the code that was there before, which had compiled the various DMI operations into their underlying JTAG writes without comments. Undoing this means I'm getting steadily better with a hex calculator...

There shouldn't be any functional change here, except that we're now using the DMI frontdoor driver, which is probably a bit more robust than what was there before.

Rather confusingly, this is somewhat reverting the change that landed in this file as fde2e0c. That change had an identical commit message to the previous change to the file and there wasn't really a proper discussion on the pull request (#20349). I'm not entirely sure why Basit thought that change was required, but I propose to undo things and revisit it if necessary.

This test should get the rv_dm_jtag_dtm_idle_hint test working in the nightlies again, but the fix will also require the changes in #23002.


hcallahan-lowrisc (approved):
Much more digestible! LGTM

Issue #2111, opened 2020-05-02
Statistical Arguments on entropy_src health check definitions


martin-lueker:
I'm approving #1943, pending a more detailed review of the statistical arguments around how to define health checks when merging several (assumedly) independent data streams.   Any one with strong opinions on statistics should weigh in on this mathematically motivated point.  This discussion should at least include: @zi-v, @martin-lueker, @felixmiller, @cyrilguyot.  Everyone should feel free to weigh in. 

This deliverables should be at least.
- a detailed definition of the test definition
- A formula relating the RCT and APT thresholds, queue depths, and the nominal miniumu entropy/bit  to the rate of false positives in the entropy/bit of an given source.
- Guidelines for setting the thresholds
- An outline of the assumptions: Notably: "How weak a correlation can we tolerate between analog channels?"
- Limits of the entropy/bit for the interleaved entropy stream, given our assumptions about the individual entropy streams.
- A second review of the RTL, confirming that the implementation matches the more clearly defined test.

mateenfaisal:
@martin-lueker, @mwbranstad - is this part of Entropy D2 backlog or already merged?

martin-lueker:
Hi Faisal,
   This is largely a documentation task, and is more relevant to the security review, than the silicon readiness review.  I don't think there are any gating factors here for Silicon.

martin-lueker:
Update: Need to add Symbol RepCnt

msfschaffner:
This appears to be mostly a documentation exercise that we can take care of when signing off to D3. As such, it does not have to go into V2S / M2. Let me relabel that accordingly.

CC @moidx @tjaychen let me know if you think otherwise.

tjaychen:
since others are helping with some of the directed M2 tasks of `entropy_src`, it would be nice if @martin-lueker could help out with the documentation side things just in case it has any impact on V2S/

martin-lueker:
Hi All,
   This issue is really intended as a guideline for firmware and software developers defining the health test thresholds and window sizes.   The real impacts will come in defining the thresholds:
   1. Boot thresholds to be written to OTP
   2. "FIPS-mode" thresholds written as part of RomExt or even later.
Any changes to the DUT as a result of these documentation would be an optimaization, so I agree it can be deferred to M3.

vogelpi:
[...]

PR #23006, opened 2024-05-07
entropy_src: add test to retrieve entropy via UART and run NIST tools

vsukhoml:
Added a test to retrieve entropy from entropy_src and dump in on UART as hex, scripts to parse log and extract it and prepare for NIST Entropy Assessment tools.

Entropy is captured in 2 modes:
- Continuous, in chunks of 80KB (limited to memory buffer), which is 655360 1-bit samples or 163840 4-bit samples.
- Restart mode in chunks of 1000 1-bit or 4-bit samples alternated with module disable/enable for restart test (check that entropy doesn't degrade if entropy source is power cycled).

./bazelisk.sh run //sw/host/opentitantool -- transport init

./bazelisk.sh test --test_output=streamed --cache_test_results=no \
 //sw/device/tests:entropy_src_fw_observe_retrieve_test_silicon_owner_prodc_rom_ext

./entropy.py bazel-testlogs/sw/device/tests/entropy_src_fw_observe_retrieve_test_silicon_owner_prodc_rom_ext/test.log

./nist_test.sh test_0

Issue https://github.com/lowRISC/opentitan/issues/2111

### DIFF
--- a/entropy.py
+++ b/entropy.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import binascii
+import sys
+import traceback
+
+def extract_hex_between_delimiters(filename):
+    """Reads a file, extracts hexadecimal digits between "+++" delimiters."""
+
+    hex_dumps = []
+    hex_data = ""
+    capture = False
+    counter = int(0)
+    with open(filename, "r", errors="ignore") as file:
+        for line in file:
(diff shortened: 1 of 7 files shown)


vogelpi:
Hi @vsukhoml , somehow I did not get tagged for review for this PR but I believe this is very relevant and it addresses a long standing issue (#2111). I believe we should re-visit this again for Gen2 / probably after M2.

PR #23007, opened 2024-05-07
[manuf] cleanup cert sizes and fix slight bug

timothytrippel:
This cleans up the DICE/TPM certificate sizes and fixes a small bug in the TPM CIK certificate's subject country string. This is in preparation for combining the DICE certs into a single flash info page.

### DIFF
--- a/sw/device/silicon_creator/lib/cert/tpm_cik.hjson
+++ b/sw/device/silicon_creator/lib/cert/tpm_cik.hjson
@@ -53,7 +53,7 @@
         not_before: "20230101000000Z",
         not_after: "20500101000000Z",
         subject: [
-            { country: "USA" },
+            { country: "US" },
             { state: "CA" },
             { organization: "Google" },
             { organizational_unit: "Engineering" },
--- a/sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
+++ b/sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
@@ -4,7 +4,6 @@
 
 #include "sw/device/lib/arch/device.h"
 #include "sw/device/lib/crypto/drivers/entropy.h"
-#include "sw/device/lib/crypto/impl/sha2/sha256.h"
 #include "sw/device/lib/dif/dif_flash_ctrl.h"
 #include "sw/device/lib/dif/dif_lc_ctrl.h"
 #include "sw/device/lib/dif/dif_otp_ctrl.h"
@@ -16,6 +15,7 @@
 #include "sw/device/lib/testing/test_framework/check.h"
 #include "sw/device/lib/testing/test_framework/ottf_main.h"
 #include "sw/device/lib/testing/test_framework/ottf_test_config.h"
+#include "sw/device/lib/testing/test_framework/status.h"
 #include "sw/device/lib/testing/test_framework/ujson_ottf.h"
 #include "sw/device/silicon_creator/lib/attestation_key_diversifiers.h"
 #include "sw/device/silicon_creator/lib/base/boot_measurements.h"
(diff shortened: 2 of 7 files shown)


vbendeb (commented):
> Additionally, this updates the sizes of all the DICE/TPM certs and their
UJSON buffers.

One thing I ran into: when we save serts in the flash we need to specify the number of flash words to use, so if the cert size is not evenly divisible by 4 a few extra trailing bytes would be read from RAM and saved in flash. While accessing memory beyond the cert is not a big deal technially, it is still good to make sure that the buffer is long enoug to cover the trailing bytes. 

Is this change accounting for this need to have extra bytes?

Also, with this change, does the end-to end test up to comparing cert hashes succeed?

timothytrippel:
> > Additionally, this updates the sizes of all the DICE/TPM certs and their
> > UJSON buffers.
> 
> One thing I ran into: when we save serts in the flash we need to specify the number of flash words to use, so if the cert size is not evenly divisible by 4 a few extra trailing bytes would be read from RAM and saved in flash. While accessing memory beyond the cert is not a big deal technially, it is still good to make sure that the buffer is long enoug to cover the trailing bytes.
> 
> Is this change accounting for this need to have extra bytes?

Ah good catch. I will update this to round up, and add appropriate static asserts.

> Also, with this change, does the end-to end test up to comparing cert hashes succeed?

Yes it passed.

sw/device/silicon_creator/lib/cert/tpm_cik.hjson
```diff
@@ -53,7 +53,7 @@
         not_before: "20230101000000Z",
         not_after: "20500101000000Z",
         subject: [
-            { country: "USA" },
+            { country: "US" },
```
timothytrippel, line 56: @vbendeb: FYI

moidx (approved):

sw/device/silicon_creator/lib/cert/tpm_cik.hjson
vbendeb: yes, this is a good catch (would have been highligted by openssl verify ;) )

vbendeb (approved):

timothytrippel:
@vbendeb see the latest commit for the static assert refactorings. The UJSON buffer sizes should now be based entirely off of autogenerated values. Due to limitations with the UJSON framework, we cannot make the size values enum constants in `provisioning_data.h` but the static asserts should ensure things stay in sync.

sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup/ft_personalize.c
```diff
@@ -40,40 +41,64 @@
...
+  kTpmEkCertBufferSizeInBytes =
+      (kTpmEkMaxCertSizeBytes + k32BitWordAlignMask) & ~k32BitWordAlignMask,
+  kTpmCekCertBufferSizeInBytes =
+      (kTpmCekMaxCertSizeBytes + k32BitWordAlignMask) & ~k32BitWordAlignMask,
+  kTpmCikCertBufferSizeInBytes =
+      (kTpmCikMaxCertSizeBytes + k32BitWordAlignMask) & ~k32BitWordAlignMask,
+};
+
+// Check TBS certificate buffer sizes.
+OT_ASSERT_MEMBER_SIZE_AS_ENUM(manuf_certs_t, uds_tbs_certificate,
```
vbendeb, line 80: I actually meant to ask earlier: why are these asserts needed in the first place? Shouldnt there just be a check when certs are stored that they fit, and another check confirming their validity?

The size of the signed cert is not guaranteed - signature R and S components are expected to be 32 bytes, but could be up to 33 in case there is a need to make sure the value is positive, or shorter than 32 if there are leading zeros...

timothytrippel: The static asserts are needed because we need to hard code the sizes of the struct fields in the UJSON certificate struct definitions in `provisioning_data.h` (due to limitations with the UJSON framework), and we want to ensure these struct definitions stay in sync with the autogenerated enum constants that get generated from the HJSON by the ot_certs framework.

pamaury:
Didn't review in time but I really like the assert change, it's much cleaner!

PR #23012, opened 2024-05-07
[manuf] fix provisioning orchistrator

timothytrippel:
This fixes the provisioning orchestrator script and the CP/FT host binary build rules that was triggering errors at run time.


timothytrippel:
CC: @BenBenderOT

PR #23016, opened 2024-05-08
[Sival, spi] Fix TPM test on teacup boad.

engdoreis:
This PR makes the teacup config for SPI TPM exactly the same as the CW310.

Testing:
```
bazel test  --//signing:token=//signing/tokens:cloud_kms_sival \
       //sw/device/tests:spi_device_tpm_tx_rx_test_silicon_owner_sival_rom_ext
```


jwnrt (approved):

a-will (approved):
LGTM

The three additional pin definitions in the SPI_TPM strapping were the critical bit--The OCTOSPI's I/Os needed to be set to high-impedance to avoid multiple drivers.

PR #23018, opened 2024-05-08
[hmac, rtl] Fix digest reloading bug

### LINKED ISSUE #23014: [hmac, rtl] Loading SHA-256 back thru DIGEST registers.
### Description

I am running SHA-256 engine in streaming mode with the following vector (by updating [`sha256_functest`](https://github.com/lowRISC/opentitan/blob/07352ce012de17dde18313a8a1b3814375d02106/sw/device/tests/crypto/sha256_functest.c#L4)):

```
static const uint8_t kLongMessage[] = {
0x45, 0x11, 0x01, 0x25, 0x0e, 0xc6, 0xf2, 0x66,
0x52, 0x24, 0x9d, 0x59, 0xdc, 0x97, 0x4b, 0x73,
0x61, 0xd5, 0x71, 0xa8, 0x10, 0x1c, 0xdf, 0xd3,
0x6a, 0xba, 0x3b, 0x58, 0x54, 0xd3, 0xae, 0x08,
0x6b, 0x5f, 0xdd, 0x45, 0x97, 0x72, 0x1b, 0x66,
0xe3, 0xc0, 0xdc, 0x5d, 0x8c, 0x60, 0x6d, 0x96,
0x57, 0xd0, 0xe3, 0x23, 0x28, 0x3a, 0x52, 0x17,
0xd1, 0xf5, 0x3f, 0x2f, 0x28, 0x4f, 0x57, 0xb8,
// here ends the first 512b block
0x5c, 0x8a, 0x61, 0xac, 0x89, 0x24, 0x71, 0x1f,
0x89, 0x5c, 0x5e, 0xd9, 0x0e, 0xf1, 0x77, 0x45,
0xed, 0x2d, 0x72, 0x8a, 0xbd, 0x22, 0xa5, 0xf7,
0xa1, 0x34, 0x79, 0xa4, 0x62, 0xd7, 0x1b, 0x56,
0xc1, 0x9a, 0x74, 0xa4, 0x0b, 0x65, 0x5c, 0x58,
0xed, 0xfe, 0x0a, 0x18, 0x8a, 0xd2, 0xcf, 0x46,
0xcb, 0xf3, 0x05, 0x24, 0xf6, 0x5d, 0x42, 0x3c,
0x83, 0x7d, 0xd1, 0xff, 0x2b, 0xf4, 0x62, 0xac,
// here ends the second 512b block
0x41, 0x98, 0x00, 0x73, 0x45, 0xbb, 0x44, 0xdb,
0xb7, 0xb1, 0xc8, 0x61, 0x29, 0x8c, 0xdf, 0x61,
0x98, 0x2a, 0x83, 0x3a, 0xfc, 0x72, 0x8f, 0xae,
0x1e, 0xda, 0x2f, 0x87, 0xaa, 0x2c, 0x94, 0x80,
0x85, 0x8b, 0xec,
// the partial block ends here (35 bytes)
};
static const uint8_t kLongMessageDigest[] = {
0x3c, 0x59, 0x3a, 0xa5, 0x39, 0xfd, 0xcd, 0xae, 0x51, 0x6c, 0xdf, 0x2f, 0x15, 0x00, 0x0f, 0x66, 0x34, 0x18, 0x5c, 0x88, 0xf5, 0x05, 0xb3, 0x97, 0x75, 0xfb, 0x9a, 0xb1, 0x37, 0xa1, 0x0a, 0xa2,
};
```
Here is what [my driver implementation](https://github.com/ballifatih/opentitan/blob/mv_hmac_impl/sw/device/lib/crypto/drivers/hmac.c) is doing:

- Initialize CFG registers, start operation and feed the first two 512b blocks continuously (which is done by `hmac_init` and `hmac_update` functions).
- Issue stop to HMAC, wait for interrupt and then store the context: DIGESTs, LOWER and UPPER.
- Later, reload the configuration and context regs DIGESTs, LOWER, UPPER and issue continue to HMAC.
- Feed the remaining 35 bytes and then issue process.
- Poll the interrupt register, and then read the result from the digest.

My local branch is https://github.com/lowRISC/opentitan/compare/master...ballifatih:opentitan:mv_hmac_impl
To reproduce with verilator:
`./bazelisk.sh test --test_output=streamed //sw/device/tests/crypto:sha256_functest_sim_verilator`
I am also running this as a TLT, with the following on my local branch:
`./util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_hmac --fixed-seed 1 --waves fsdb`

From the waveform, I observed that when I am reloading DIGEST values (that I confirmed by looking at TL-UL transactions), they do not end up in the internal [`digest_o = digest_q`](https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/prim/rtl/prim_sha2.sv;l=159?q=prim_sha2&ss=opentitan) register of `prim_sha2.sv`. Also there are actually 7 pulses from `digest_sw_we` signals instead of 8:

image

I suspect that the indexing part in the expression `reg2hw.digest[i+1]` might be the root cause:

https://github.com/lowRISC/opentitan/blob/07352ce012de17dde18313a8a1b3814375d02106/hw/ip/hmac/rtl/hmac.sv#L235-L242

cc: @gdessouky @martin-velay @moidx

ballifatih:
Fixes #23014 and updates the programmers guide.

The gist is that we need deassert `sha_en` as a clean-up after HMAC operation/streaming is done. `sha_en` is only asserted once `DIGEST` registers are restored.


hw/ip/hmac/doc/programmers_guide.md
```diff
@@ -88,14 +88,17 @@ To switch from one message stream to another, set the `CMD.hash_stop` bit, wait
...
-9. Continue this with as many message blocks and parallel message streams as needed. The final hash for any message stream can be obtained at any time (no need for complete blocks) by setting `CMD.hash_process` and waiting for the `hmac_done` interrupt / status bit, finally reading the digest from the `DIGEST` registers.
+1. Write an arbitrary number of message blocks to HMAC's `MSG_FIFO`.
+1. Stop HMAC by setting the `CMD.hash_stop` bit and wait for the `hmac_done` interrupt (or poll the interrupt status register).
+1. Save the context by reading the `DIGEST_0`..`15` and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers. (The values in the `CFG` register must also be preserved, but that is purely SW-controlled so doesn't need to be read from HW.)
+1. Disable `sha_en` by updating `CFG` register, in order to clear the digest from stream A.
+1. Repeat steps 1-5 for message stream B.
+1. Before restoring context, make sure that `sha_en` in `CFG` remains disabled.
+1. Restore the context of message stream A by writing the `CFG`, `DIGEST_0`..`15`, and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers.
+1. Enable `sha_en` of CFG, which locks down `DIGEST_0`..`15` registers.
```
gdessouky, line 98: How about  "Enable `sha_en` of CFG, which enables HMAC again and locks down writing to the `DIGEST_0`..`15` registers from SW."?

hw/ip/hmac/doc/programmers_guide.md
```diff
@@ -88,14 +88,17 @@ To switch from one message stream to another, set the `CMD.hash_stop` bit, wait
...
-5. Repeat steps 1-4 for message stream B.
-6. Restore the context of message stream A by writing the `CFG`, `DIGEST_0`..`15`, and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers.
-7. Continue processing message stream A by setting the `CMD.hash_continue` bit.
-8. Write an arbitrary number of message blocks to HMAC's `MSG_FIFO`.
-9. Continue this with as many message blocks and parallel message streams as needed. The final hash for any message stream can be obtained at any time (no need for complete blocks) by setting `CMD.hash_process` and waiting for the `hmac_done` interrupt / status bit, finally reading the digest from the `DIGEST` registers.
+1. Write an arbitrary number of message blocks to HMAC's `MSG_FIFO`.
+1. Stop HMAC by setting the `CMD.hash_stop` bit and wait for the `hmac_done` interrupt (or poll the interrupt status register).
+1. Save the context by reading the `DIGEST_0`..`15` and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers. (The values in the `CFG` register must also be preserved, but that is purely SW-controlled so doesn't need to be read from HW.)
+1. Disable `sha_en` by updating `CFG` register, in order to clear the digest from stream A.
```
gdessouky, line 94: I would add here that "This is necessary to also prevent leakage of intermediate context of one SW thread to another."

gdessouky (approved):
thanks Fatih, this LGTM! I only left minor comments for more clarification in the programmer's guide.

ballifatih:
Thanks @gdessouky, I made the editorial changes in the doc as you suggested.

hw/ip/hmac/doc/programmers_guide.md
```diff
@@ -88,14 +88,17 @@ To switch from one message stream to another, set the `CMD.hash_stop` bit, wait
...
-5. Repeat steps 1-4 for message stream B.
-6. Restore the context of message stream A by writing the `CFG`, `DIGEST_0`..`15`, and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers.
-7. Continue processing message stream A by setting the `CMD.hash_continue` bit.
-8. Write an arbitrary number of message blocks to HMAC's `MSG_FIFO`.
-9. Continue this with as many message blocks and parallel message streams as needed. The final hash for any message stream can be obtained at any time (no need for complete blocks) by setting `CMD.hash_process` and waiting for the `hmac_done` interrupt / status bit, finally reading the digest from the `DIGEST` registers.
+1. Write an arbitrary number of message blocks to HMAC's `MSG_FIFO`.
+1. Stop HMAC by setting the `CMD.hash_stop` bit and wait for the `hmac_done` interrupt (or poll the interrupt status register).
+1. Save the context by reading the `DIGEST_0`..`15` and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers. (The values in the `CFG` register must also be preserved, but that is purely SW-controlled so doesn't need to be read from HW.)
```
martin-velay, line 93: I think the `key` registers should also be preserved as values can be changed

hw/ip/hmac/doc/programmers_guide.md
```diff
@@ -88,14 +88,17 @@ To switch from one message stream to another, set the `CMD.hash_stop` bit, wait
...
+1. Disable `sha_en` by updating `CFG` register, in order to clear the digest from stream A. This is necessary to also prevent leakage of intermediate context of one SW thread to another.
+1. Repeat steps 1-5 for message stream B.
+1. Before restoring context, make sure that `sha_en` in `CFG` remains disabled.
+1. Restore the context of message stream A by writing the `CFG`, `DIGEST_0`..`15`, and `MSG_LENGTH_`{`LOWER`,`UPPER`} registers.
```
martin-velay, line 97: And the `key` registers?

martin-velay (commented):
I think we should also mention the possibility to change the secret key between contexts. I feel that's an important feature. See my comments.

Otherwise, the rest LGTM

hw/ip/hmac/doc/programmers_guide.md
andreaskurth: I think you're right, because the key not only gets hashed *before* the message (in the first round) but also *after* the message (in the second round) before the final digest is computed.

andreaskurth (approved):
LGTM modulo @martin-velay's suggestion. Thx for these improvements / fixes, @ballifatih :+1:

hw/ip/hmac/doc/programmers_guide.md
gdessouky: indeed, and `CFG` includes the `key_length` field.

ballifatih:
Thanks @martin-velay for pointing out that `KEY` registers also need to be stored/restored during context switches for keyed HMAC (but not SHA, as far as I see). I updated the doc to clarify that.

PR #23028, opened 2024-05-08
[manuf] cleanup `ft_personalize`

timothytrippel:
This cleans up the `ft_personalize` binary to consolidate operations into sub-functions, and increase readability. Additionally this fixed an initialization bug in the size of the TPM TBS certificate buffers. Lastly, this updates the certificate hash comparison on the host side to print both hashes if the comparison fails (so that it is easier to see how the host interpretted the bytes, including in the order of said bytes).


vbendeb:
Looks OK, a good test would be confirming that openssl verification still works.

moidx (approved):

PR #23030, opened 2024-05-09
[dv,flash_ctrl] Fix failures for prog with ecc enabled

### LINKED ISSUE #22973: [dv,flash_ctrl] Fix breakage for prog_flash
### Description

A big breakage of flash_ctrl tests performing flash_prog was introduced in #22804.

That PR missed running a full regression, or it would have found these breakages. My apologies. The change in that PR fixed the randomization for writes in a specific sequence by adjusting randomization. However, the changes also affected all other sequences, most of which were not prepared for it.

The best fix is conceptually simple but delicate: currently the class randomization happens before, and independently of choosing the flash op, and the error modes. The pattern is, for example

this.randomize()
randcase
  cfg.seq_cfg.otf_wr_pct: prog_flash(...);
  cfg.seq_cfg.otf_rd_pct: read_flash(...);
  ...
endcase

And the `flash_op_t` op field is set inside either prog_flash, read_flash, or others.

This means the class randomization cannot set per-op variables correctly, like preventing program resolution faults, half word updates, or preventing multiple programs to the same word with no erase between.

The preferred solution is to create a constraint for rand_op.op like

rand_op.op dist  { FlashOpProgram := cfg.otf_wr_pct, FlashOpRead := cfg.otf_rd_pct, ...}

And solve op before fractions and address.

Unfortunately this causes many inconsistent randomized variables, suggesting the constraint solver is faulty. Given the turnaround time for tool support, this won't be addressed in time for M4, so we will pursue a different strategy, and if time permits we will raise an issue for VCS to look into this.

The proposed solution is to adapt tests that use prog_flash to use the strategy employed in flash_ctrl_rw_vseq.

matutem:
This fixes failures introduced in #22804. That PR didn't run a full flash_ctrl block regression, and it caused lots of breakages in tests that launched program flash ops but were not instrumented as the flash_ctrl_rw_vseq was.

This also fixes other issues found while developing this:
- The detection of program resolution violations needs to consider all address bits past the resolution size.
- The workaround to avoid program resolution window faults needs to be extended to handle multiple splits.
- The data to be sent with a program request needs to be initialized for each flash control start.

Fixes #22973

### DIFF
--- a/hw/ip/flash_ctrl/dv/env/flash_ctrl_otf_scoreboard.sv
+++ b/hw/ip/flash_ctrl/dv/env/flash_ctrl_otf_scoreboard.sv
@@ -331,7 +331,7 @@ class flash_ctrl_otf_scoreboard extends uvm_scoreboard;
     // So each pop becomes 8 times of fqs.
     int col_sz = exp.fq.size / 8;
     `uvm_info("process_write", $sformatf("process_write: addr:0x%x bank:%0d colsz:%0d ffsz:%0d",
-                       exp.cmd.otf_addr, bank, col_sz, eg_rtl_ctrl_fifo[bank].used()), UVM_MEDIUM)
+              exp.cmd.otf_addr, bank, col_sz, eg_rtl_ctrl_fifo[bank].used()), UVM_MEDIUM)
     eg_rtl_ctrl_fifo[bank].get(item);
 
     `uvm_create_obj(flash_otf_item, obs)
--- a/hw/ip/flash_ctrl/dv/env/flash_ctrl_seq_cfg.sv
+++ b/hw/ip/flash_ctrl/dv/env/flash_ctrl_seq_cfg.sv
@@ -72,6 +72,12 @@ class flash_ctrl_seq_cfg extends uvm_object;
   // Avoid creating flash program operations that cross program resolution boundaries of 64 byte.
   bit avoid_prog_res_fault;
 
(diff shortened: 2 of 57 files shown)


matutem:
I may update this with a fix for some remaining failures. A full flash_ctrl regresion is running. The results will be added here when available.

vogelpi (approved):
This looks good, thanks @matutem for fixing this and further improving flash_ctrl DV.

I see there are still some lint errors. Once these are fixed and the full regression confirmed we have again reasonable passrates and coverage this should be merged.

matutem:
The block level regression summary:
|         |                                           | **TOTAL**                                |                   |                  |   1236    |  1278   |   96.71 %   |

## Coverage Results

### [Coverage Dashboard](/home/matute/work/jot_g/scratch/flash_ctrl_extend_rw/flash_ctrl-sim-vcs/cov_report/dashboard.html)

|  SCORE  |  LINE   |  COND   |  TOGGLE  |   FSM   |  BRANCH  |  ASSERT  |  GROUP  |
|:-------:|:-------:|:-------:|:--------:|:-------:|:--------:|:--------:|:-------:|
| 96.50 % | 95.84 % | 94.36 % | 98.85 %  | 91.84 % | 98.27 %  | 98.10 %  | 98.21 % |

I will merge this, and look into some of the remaining failures at a lower priority.

vogelpi:
Thanks @matutem for posting the regression results, this looks very good!

PR #23037, opened 2024-05-09
[sival] Clarify SiVal steps of flash_ctrl LC tests

jwnrt:
In DV we can transition the chip and check these signals. This is not possible in SiVal. We will still have a set of chips in these LC states that we check the accessibilities of the OTP partitions from without having to do the transition.

Also updates the `lc_states` to the steps specified by the plan as needing to be checked.


hw/top_earlgrey/data/ip/chip_flash_ctrl_testplan.hjson
```diff
@@ -147,11 +147,15 @@
             - Transition from TEST_LOCKED to DEV/PROD to ESCALATION/SCRAP state via OTP and verify
               that this LC signal transitions from 0 to 1 and back to 0. Verify that the SW
               accessibility of the corresponding partition depending on the signal value.
+
+            In SiVal, we cannot transition the chip or check this signal. The
```
engdoreis, line 151:
```suggestion
            In Silicon, we cannot transition the chip or check this signal. The
```
To be consistent with other testplans.

PR #23042, opened 2024-05-09
[top_earlgrey,pinmux] Add input disable pad attribute

andreaskurth:
This PR adds an _input disable_ attribute for pads that are controlled through pinmux. A later PR will add the attribute to manually instantiated pads (probably through sensor_ctrl).


hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
```diff
@@ -53,7 +53,7 @@ module prim_xilinx_pad_wrapper
 
     // input inversion
     logic in;
-    assign in_raw_o = (ie_i) ? in  : 1'bz;
+    assign in_raw_o = (ie_i & attr_i.input_en) ? in  : 1'bz;
```
a-will, line 56: High-impedance doesn't actually make sense inside the FPGA:

```suggestion
    assign in_raw_o = (ie_i & attr_i.input_en) ? in  : 1'b0;
```

a-will: Actually, on second thought, we should probably do this properly (if we choose to support it on FPGAs at all):

```systemverilog
    logic ie_n;
    assign ie_n = ~(ie_i & attr_i.input_en);
    assign in_raw_o = in;

    IBUF_IBUFDISABLE u_ibuf (
      .I ( inout_io ),
      .IBUFDISABLE( ie_n ),
      .O ( in       )
    );
```

hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
```diff
@@ -67,7 +67,7 @@ module prim_xilinx_pad_wrapper
...
```
a-will, line 70: For the Bidir* family, we could either skip, do the proper support, or leave it with logical emulation.

"Proper" support requires knowing whether the connected port is in a high-performance or high-range bank, since the primitives allowed are different. IOBUF_DCIEN is only allowed in high-performance banks, whereas IOBUF_INTERMDISABLE is only allowed in high-range banks.

andreaskurth: Thanks for flagging this, @a-will. I had hoped that behavioral code would let Vivado figure out the correct cell to insert for each pad. But that may not be supported (I haven't checked), so it's probably better to explicitly instantiate the correct cell. However, I wonder if we could use the logic emulation approach that is currently in the PR for the short term and create an issue to instantiate the proper cells when there's less time pressure to finish the RTL that goes into the chip (post M4/M5)?

a-will: Yeah, logic emulation seems fine to me, but maybe we should not use `1'bz` for the fabric side of an FPGA's prim pad wrapper. The IOBUF primitives in 7-series will output `1'b1` if the input buffer is disabled. In Ultrascale, they will output `1'b0`.

hw/ip/pinmux/data/pinmux.hjson.tpl
```diff
@@ -566,7 +566,7 @@
                   regwen:       "MIO_PAD_ATTR_REGWEN",
                   regwen_multi: "true",
                   cname:        "MIO_PAD",
-                  resval:       0
+                  resval:       "0x80",
```
a-will, line 569: If the intention is to not introduce software changes, the logic might've been a bit simpler if the pad attribute were `input_disable` (i.e. with logic levels that map a 1 to disable).

`input_en` is totally fine too, though. Just figured I'd present that option in case you hadn't considered it.

andreaskurth: True; that's probably less surprising from a SW/programming perspective

hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
```diff
@@ -51,23 +51,33 @@ module prim_xilinx_pad_wrapper
...
-      .O ( in       )
-    );
   end else if (PadType == BidirTol ||
                PadType == DualBidirTol ||
                PadType == BidirOd ||
                PadType == BidirStd) begin : gen_bidir
 
-    // input inversion
+    // Input buffer with input disable
+    // TODO(#xTODO): This should be implemented with an instance of `IOBUF_DCIEN` (for pads in
```
andreaskurth, line 72: TODO (pre merge): Create issue for this and insert number here

hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_pad_wrapper.sv
```diff
@@ -51,23 +51,33 @@ module prim_xilinx_ultrascale_pad_wrapper
...
-      .O ( in       )
-    );
   end else if (PadType == BidirTol ||
                PadType == DualBidirTol ||
                PadType == BidirOd ||
                PadType == BidirStd) begin : gen_bidir
 
-    // input inversion
+    // Input buffer with input disable
+    // TODO(#xTODO): This should be implemented with an instance of `IOBUF_DCIEN` (for pads in
```
andreaskurth, line 72: TODO (pre merge): Create issue for this and insert number here

hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
andreaskurth: SG, thx. Updated

hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
andreaskurth: Thx, updatde

hw/ip/pinmux/data/pinmux.hjson.tpl
andreaskurth: Fixed; it's now `input_disable`

hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
a-will: Whoops, I missed that there is a compatibility matrix:

```
ibuf_IOSTANDARD_not_compatible_with_IBUFDISABLE: u_padring/gen_dio_pads[11].u_dio_pad/gen_xilinx.u_impl_xilinx/gen_input_only.u_ibuf is using input pin IBUFDISABLE which is not supported for IOSTANDARDs LVCMOS12,LVCMOS15,LVCMOS18,LVCMOS25,LVCMOS33,MOBILE_DDR,DIFF_MOBILE_DDR,LVDCI_18,LVDCI_15,LVDCI_DV2_18,LVDCI_DV2_15,PCI33_3,LVTTL
```

I guess we'll have to do logic values and no true buffer disablement after all.

hw/ip/pinmux/fpv/vip/pinmux_assert_fpv.sv
```diff
@@ -322,7 +323,7 @@ module pinmux_assert_fpv
                                                {bid_pad_types} ? 1 : 0;
   assign dio_pad_attr_mask.keep_en     = 0;
   assign dio_pad_attr_mask.schmitt_en = 0;
-  assign dio_pad_attr_mask.od_en = 0;
```
a-will, line 325: Did you mean to get rid of this line?

It doesn't particularly matter to me, since the default should already have it set to 0, but it is the odd one out.

a-will (approved):
IBUF_IBUFDISABLE turns out to only be particularly useful for HSTL, SSTL, and other V_REF-using IOSTANDARDs, so we can just drop the primitive, I think. If CI passes, though, leaving it in also seems fine.

It was my mistake to encourage actually supporting turning off buffers on the FPGA.

hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
```diff
@@ -44,59 +44,79 @@ module prim_xilinx_pad_wrapper
...
                            oe_i,
                            attr_i.virt_od_en};
 
-    // input inversion
+    // Input buffer with input disable
+    // TODO(#xTODO): This should be implemented with an instance of `IBUF_IBUFDISABLE`, which has
+    // parameters `IBUF_LOW_PWR` and `IOSTANDARD`, which depend on the physical FPGA pad that this
+    // module is connected to.  However, this module currently doesn't have that information,  so
+    // the logic below instead emulates this behavior (disabled inputs get internally driven to 1
+    // for 7 Series devices) without really disabling the input driver.
```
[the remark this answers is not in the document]
andreaskurth: Are you sure we can't? From how I understand Tables 1-12 and 1-13 in [7 Series FPGAs SelectIO Resources User Guide (UG471)](https://docs.amd.com/v/u/en-US/ug471_7Series_SelectIO), the IOSTANDARDs we use (LVCMOS33 and LVCMOS18) are supported for IBUF. It doesn't explicitly say that this also applies to IBUF_IBUFDISABLE, but since there is no separate table for the latter, I'd assume it does.

hw/ip/pinmux/fpv/vip/pinmux_assert_fpv.sv
andreaskurth: Oops, no that was a mistake. I'll revert that

hw/ip/prim_xilinx/rtl/prim_xilinx_pad_wrapper.sv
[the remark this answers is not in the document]
andreaskurth: Ah, but we saw this error message
```
ibuf_IOSTANDARD_not_compatible_with_IBUFDISABLE: u_padring/gen_dio_pads[11].u_dio_pad/gen_xilinx.u_impl_xilinx/gen_input_only.u_ibuf is using input pin IBUFDISABLE which is not supported for IOSTANDARDs LVCMOS12,LVCMOS15,LVCMOS18,LVCMOS25,LVCMOS33,MOBILE_DDR,DIFF_MOBILE_DDR,LVDCI_18,LVDCI_15,LVDCI_DV2_18,LVDCI_DV2_15,PCI33_3,LVTTL`
```
which explicitly states that `IBUFDISABLE` is not supported for LVCMOS18 and LVCMOS33. I'll update the comment, and I'll also change the "disabled" value to 0 to align it with the behavior on UltraScale devices, for which we may eventually enable this.

PR #21348, opened 2024-02-13
[sival, spi] Passthru

engdoreis:
This PR fixes the spi_passthru on silicon and link it to the sival testplan

PR #23044, opened 2024-05-10
Backport [sival, spi] Passthru

engdoreis:
This is a backport PR of https://github.com/lowRISC/opentitan/pull/21348#issuecomment-2094051293 to master.
There was a conflict on the bazel BUILD file.

### DIFF
--- a/hw/top_earlgrey/BUILD
+++ b/hw/top_earlgrey/BUILD
@@ -429,7 +429,7 @@ silicon(
     name = "silicon_owner_sival_rom_ext",
     testonly = True,
     args = [
-        "--rcfile=",
+        "--rcfile=__builtin__/hyperdebug_teacup.json",
         "--logging=info",
         "--interface={interface}",
     ],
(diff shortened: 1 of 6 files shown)


a-will (approved):

hw/top_earlgrey/BUILD
```diff
@@ -429,7 +429,7 @@ silicon(
     name = "silicon_owner_sival_rom_ext",
     testonly = True,
     args = [
-        "--rcfile=",
+        "--rcfile=__builtin__/hyperdebug_teacup.json",
```
cfrantz, line 432: I think this is an error.
`--rcfile=...` tells opentitantool which user-level config file to read, which contains commandline options like `--interface=foo` that the user would like set by default.

As far as I recall, `--rcfile` cannot parse a JSON file.

engdoreis: I think you are right. I'll file a PR fixing it.

PR #23046, opened 2024-05-10
[rv_dm,dv] Drive correct enable signal in dmi_disabled vseq

rswarbrick:
This was using the lc_hw_debug_en signal (which controls whether the debugger actually does anything), but we should have been using the pinmux_hw_debug_en, which controls whether DMI access is enabled.

This fixes the rv_dm_jtag_dmi_debug_disabled test, which was failing because... DMI wasn't disabled! Oops :-)


marnovandermaas (approved):
Looks reasonable to me, please take a look at whether the CI fails for any related reason before merging.

PR #23047, opened 2024-05-10
[test] Make e2e_bootstrap_rma test more resilient

jwnrt:
There was an assert checking that the LC controller's status was `INITIALIZED`. This failed if the status was `INITIALIZED | READY`.

Changed to use the `wait_for_status` function which allows unions and also has retries built in.


andreaskurth (approved):
LGTM, thx for the fix

PR #23049, opened 2024-05-10
[deps/crypto] add skeleton code for self-hosted SPHINCS+ library

RyanTorok:
Follow-up to #22953 

Background: the existing SPHNICS+ crate that `opentitantool` relies on does not support the `consistent_basew` update to the algorithm in the newest NIST draft standard that is planned for earlgrey prod.

This issue was discussed in the Software WG, and the consensus was that we would develop our own SPHINCS+ crate that imports the correct version of the SPHINCS+ C code.

This PR adds a bazel toolchain to fetch the correct branch of the SPHINCS+ repository (which unfortunately does not have a release, so we fall back to using the `bare_repository` bazel rule). From there, `rules_rust` provides some nice tooling to import the Rust library generated by the `bindgen` tool into our code.

### DIFF
--- a/sw/host/sphincsplus/BUILD
+++ b/sw/host/sphincsplus/BUILD
@@ -1,0 +1,29 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+load("@rules_rust//bindgen:defs.bzl", "rust_bindgen_library")
+load("@rules_rust//rust:defs.bzl", "rust_library", "rust_test")
+
+rust_bindgen_library(
+    name = "sphincsplus_bindgen",
+    bindgen_flags = [
+        "--allowlist-function=crypto_.*",
+        "--allowlist-var=CRYPTO_.*",
+        "--allowlist-var=SPX_.*",
+    ],
+    cc_lib = "@sphincsplus_fips205_ipd//:sphincs_random_shake_128s_simple",
+    clang_flags = ["-DPARAMS=sphincs-shake-128s"],
+    header = "@sphincsplus_fips205_ipd//:api.h",
+)
+
+rust_library(
+    name = "sphincsplus",
+    srcs = ["lib.rs"],
+    deps = [":sphincsplus_bindgen"],
+)
+
+rust_test(
+    name = "sphincsplus_test",
+    crate = ":sphincsplus",
+)
--- a/third_party/sphincsplus/repos.bzl
+++ b/third_party/sphincsplus/repos.bzl
@@ -2,9 +2,10 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+load("//rules:repo.bzl", "http_archive_or_local")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-def sphincsplus_repos():
+def sphincsplus_repos(local = None):
     http_archive(
         name = "sphincsplus_kat",
         build_file = Label("//third_party/sphincsplus:BUILD.sphincsplus_common.bazel"),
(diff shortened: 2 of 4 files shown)


third_party/sphincsplus/repos.bzl
```diff
@@ -19,3 +20,96 @@ def sphincsplus_repos():
             "https://storage.googleapis.com/ot-crypto-test-vectors/sphincsplus/sphincsplus_shake256_128s_round3.zip ",
         ],
     )
+    bare_repository(
+        name = "sphincsplus_fips205_ipd",
+        url = "https://github.com/sphincs/sphincsplus/archive/129b72c80e122a22a61f71b5d2b042770890ccee.tar.gz",
+        strip_prefix = "sphincsplus-129b72c80e122a22a61f71b5d2b042770890ccee",
+        additional_files_content = {
+            "ref/BUILD": """exports_files(glob(["**"]))
```
RyanTorok, line 28: Due to the size of this injected BUILD file, I would like to make it a separate file, but I was unable to get the `additional_files` argument to `bare_repository` to work correctly. @cfrantz do you know how I have to format the path in order to get this to work?

cfrantz: You should consider `http_repository` (or `//rules/repo.bzl%http_repository_or_local`) instead of `bare_repository`.

`bare_repository` was developed to create a bazel repo that is a symlinked hierarchy with a bunch of BUILD files injected into the hierarchy.  Specifically, `bare_repository` was developed to inject an entire hierarchy of BUILD files into tock and libtock-rs.

I think the sphincsplus reference implementation is sufficiently simple in terms of directory hierarchy that you can probably write an appropriate `cc_library` rule at the root of the and simply add `build_file = "..."` the the http_repository rule.

In the interest of not being obtuse and answering your question, the `additional_build_files` argument accepts a [string:string] dict mapping the desired foreign-repo file location to the local (opentitan) repo file location:
```
additional_files = {
  "ref/BUILD": "third_party/sphincsplus/BUILD.spx-ref.bazel",
  "ref/params/BUILD": "third_party/sphincsplus/BUILD.spx-params.bazel",
  ...
}
```

I don't think the existing users of `bare_repository` actually make use of `additional_files`, so its not exactly well tested.

cfrantz: Edit: you might need to prefix the local paths with `//` and/or call the `Label` constructor:
```
"foo/BUILD": Label("//third_party/foo/BUILD.foo.bazel"),
```

RyanTorok: Thanks for the feedback. I knew the standard `http_archive` doesn't work for fetching specific git commits, but I didn't know that `http_repository` would work.

RyanTorok: > Edit: you might need to prefix the local paths with `//` and/or call the `Label` constructor:
> 
> ```
> "foo/BUILD": Label("//third_party/foo/BUILD.foo.bazel"),
> ```

I've already tried this. It complains about the path not being a string.

cfrantz: Sorry, I typo-ed that.  It's `http_archive`.

You need to construct the correct github URL to fetch the exact commit you want.

Something like:
```
https://github.com/$user/$repo/archive/$git_hash.tar.gz
```

RyanTorok: I got this working with `http_archive_or_local`.

I think I had incorrectly remembered that we couldn't use `http_archive` with specific git commits. When we added wycheproof as a non-vendored dependency we explicitly avoided depending on a specific git commit because we wanted to keep up with the repo.

jadephilipoom (approved):
This LGTM, but I'm not sure I know the details well enough to sign off; I'd wait for Chris's more authoritative review before merging :slightly_smiling_face:

sw/host/sphincsplus/BUILD
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+load("@rules_rust//bindgen:defs.bzl", "rust_bindgen_library")
+load("@rules_rust//rust:defs.bzl", "rust_library", "rust_test")
+
+rust_bindgen_library(
+    name = "sphincsplus_bindgen",
+    cc_lib = "@sphincsplus_fips205_ipd//:sphincs_random_shake_128s_simple",
+    clang_flags = ["-DPARAMS=sphincs-shake-128s"],
+    header = "@sphincsplus_fips205_ipd//:api.h",
```
cfrantz, line 12: The library generated by this rule is pretty permissive about what gets bindings.  You should use some flags to be more restrictive (ie: we don't care about most of the junk that gets incidentally included like `uint128_t` or `foff_t`, etc).

You could try something like this:
```
    bindgen_flags = [
        "--allowlist-function=crypto_.*",
        "--allowlist-var=CRYPTO_.*",
        "--allowlist-var=SPX_.*",
    ],
```
(see also the [bindgen manual](https://rust-lang.github.io/rust-bindgen/allowlisting.html))

RyanTorok: From a cursory look at what gets generated, this looks fine.

cfrantz (approved):

PR #23059, opened 2024-05-10
[rv_dm,dv] Temporarily avoid scanmode in rv_dm_base_vseq

rswarbrick:
Without doing this, things get rather confused if the scanmode happens to be true. The problem is that the jtag driver has to model the current FSM state, and it ends up counting a different clock from the one in the design. This is rather confusing!

The proper long-term fix is probably to make the jtag driver able to consume its TCK passively when configured to do so.


hcallahan-lowrisc (approved):
Can we create a TODO item to track this? Or link it to another item that tracks rv_dm testbench items? Other than that, LGTM.

PR #23062, opened 2024-05-10
[rescue] Send only the application firmware

cfrantz:
The initial implementation of the rescue command sent the entire file as the rescue payload.  However, the output of many software builds is an image containing a ROM_EXT _and_ an application image.

1. Parse the complete image into sub-images.
2. Update the `image manifest show` command to show all manifests present in an image.
3. Update the `rescue firmware` command to select the first application sub-image inside of an image.


timothytrippel (approved):

jettr (approved):
Thank you!

PR #23064, opened 2024-05-10
[rom_ext] Use Ibex for RSA sigverify

cfrantz:
Since we are planning to do away with RSA signatures in favor of ECDSA signatures, we can use Ibex for RSA verification until we fully eliminate the use of RSA.  This important temporary measure allows us to bring the full ECDSA OTBN program into the ROM_EXT and use it for ECDSA verification of ownership transfer operations.

1. Switch RSA sigverify of the owner application firmware over to Ibex-based verification.
2. Cherry-pick jadephilipoom's OTBN P-256 implementation.
3. Use the OTBN implementation for ownership-related verification, thus eliminating CryptoC and saving 12K of flash space.

### DIFF
--- a/sw/device/silicon_creator/lib/ownership/BUILD
+++ b/sw/device/silicon_creator/lib/ownership/BUILD
@@ -33,13 +33,19 @@ cc_library(
     name = "ecdsa",
     srcs = ["ecdsa.c"],
     hdrs = ["ecdsa.h"],
-    defines = ["USE_CRYPTOC=1"],
+    defines = [
+        "USE_OTBN=1",
+        #"USE_CRYPTOC=1",
+    ],
     deps = [
         ":datatypes",
         "//sw/device/lib/base:hardened",
+        "//sw/device/lib/base:hardened_memory",
         "//sw/device/lib/base:macros",
+        "//sw/device/silicon_creator/lib:error",
         "//sw/device/silicon_creator/lib/drivers:hmac",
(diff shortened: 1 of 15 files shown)


sw/device/silicon_creator/lib/ownership/BUILD
```diff
@@ -33,13 +33,19 @@ cc_library(
     name = "ecdsa",
     srcs = ["ecdsa.c"],
     hdrs = ["ecdsa.h"],
-    defines = ["USE_CRYPTOC=1"],
+    defines = [
+        "USE_OTBN=1",
+        #"USE_CRYPTOC=1",
```
jadephilipoom, line 38: nit: are these two commented-out lines (here and a few lines below) left in intentionally? I see there are also some `#ifdefs` with `USE_CRYPTOC` still left in; will that still work, or should these be removed?

jadephilipoom (approved):

sw/device/silicon_creator/lib/ownership/BUILD
cfrantz: They've been left in intentionally, as both implementations should be functional.

Once I get this merged and tested with the rest of the WIP ownership transfer code & tests, I'll eliminate CryptoC entirely.

timothytrippel (approved):

moidx (approved):

cfrantz:
The test failure in CI is unrelated to this change.

PR #23066, opened 2024-05-11
[manuf]: add code to create Cloud KMS signed CA certificate

vbendeb:
The modification makes use of the openssl ability to interact with Google Cloud KMS when the appropriate configuration set.

When the appropriate environment variables are set to point to a local instance of Google KMS integration library
(https://github.com/GoogleCloudPlatform/kms-integrations) and Cloud KMS configuration, the second certificate, signed by the configured external crypto engine is also generated.


vbendeb:
hi Tim, I know you worked on CA cert name changes, I can update this if necessary. Not even sure we want to merge it, but this is an example of how the KMS signed CA can be created.

timothytrippel (approved):

timothytrippel:
> hi Tim, I know you worked on CA cert name changes, I can update this if necessary. Not even sure we want to merge it, but this is an example of how the KMS signed CA can be created.

No problem, I will wait to do the renaming until we get this and the following PR (#23069) in to avoid blocking you.

timothytrippel:
Note, since this is passing CI, I think you can cleanup the comments I mentioned above in #23609, since I made other suggestions on that PR too.

vbendeb:
sorry for the typos I should set up vscode to spellcheck bash comments. Will update in the other PR

PR #22460, opened 2024-04-06
[i2c] Clean up target mode transaction boundaries and NACK handling

a-will:
Change recording of Start symbols to always accompany the address data,
instead of recording a separate R.Start, then Start+address. This helps
maintain certain availability invariants needed for reliable stretching
and NACK handling.

Note that a controller that chooses to do a Repeated Start and address a
*different* target is an unusual case, and it leads to a situation where
the transfer closing symbol won't appear until the Stop or a later
Repeated Start that *does* address this target. While this transaction
style is not explicitly forbidden in the I2C specification, it is not
typically supported in the wild. In any case, software should be able to
resolve this case, and a closing symbol will eventually appear.

Fix up the DV transaction generation to match the expected behavior.

Remove the extra NACK-specific states, generate ACQ FIFO entries at the
point of the NACK timeout, and go to WaitForStop to release SDA and SCL.
Return 0xFF on reads for NACK'd transactions.

Add NackStop as an ACQ FIFO value to indicate the end of transactions
that had errors (generally, NACK'd transfers).

Fix up the glitch sequence to better handle the various states.

Finally, add a StretchAcqSetup state to force the target FSM to maintain the required setup time for the ACK bit it started driving.

PR #23073, opened 2024-05-13
[i2c,testplan] Add testpoints for new features

### LINKED ISSUE #22895: [i2c,dv] Update testplan for MCTP/SMBus M4 features

hcallahan-lowrisc:
This PR adds new testpoints to the testplan for newly-merged I2C features

Currently split into commits with new testpoints per-PR/commit that was merged, just to make reviewing and discussion easier. This will be squashed down before merging.

Changes from:
- [x] #22345 
- [x] #22459 
- [x] #22460 
- [x] #22551 
- [x] #22864 
- [x] #22865 
- [x] #22872 

Closes #22895

### DIFF
--- a/hw/ip/i2c/data/i2c_testplan.hjson
+++ b/hw/ip/i2c/data/i2c_testplan.hjson
@@ -398,11 +398,11 @@
             Stimulus:
               - Configure DUT/Agent to Target/Host mode respectively
               - Send enough read and write requests to acq_fifo
-              - Slow down acq FIFO read process to trigger acq_full interrupt
+              - Slow down acq FIFO read process to trigger 'acq_stretch' interrupt
 
             Checking:
               - Check FIFO full states by reading status register
-              - Whether acq_full interrupt raised
+              - Whether 'acq_stretch' interrupt raised
             '''
       stage: V2
       tests: ["i2c_target_stress_wr", "i2c_target_stress_rd", "i2c_target_intr_stress_wr"]
@@ -410,7 +410,7 @@
     {
       name: target_timeout
       desc: '''
-            Test host_timeout interrupts.
+            Test the 'host_timeout' interrupt.
 
             Stimulus:
               - Configure DUT/Agent to Host/Target mode respectively
@@ -418,7 +418,7 @@
               - Agent stops sending clock during an ongoing transaction
 
             Checking:
-              - Ensure host_timeout is asserted and a correct number is received
+              - Ensure the 'host_timeout' irq is asserted and a correct number is received
 
             '''
       stage: V2
@@ -576,20 +576,28 @@
             such as:
               - NACK when ACQFIFO is full.
               - NACK when TARGET_TIMEOUT_CTRL threshold has elapsed.
-              - ACQDATA.NACK/ACQDATA.NACKSTART is correctly recorded in the ACQFIFO.
+                - The timeout is measured against the cumulative stretching time within a single
+                  transaction. This accumulation resets at the end of the current transaction.
+                - All further bytes are NACK'd unconditionally, awaiting the end of the transaction.
+              - ACQDATA.{NACK,NACKSTART,NACKSTOP} is correctly recorded in the ACQFIFO.
               - TARGET_NACK_COUNT recording the number of NACK'd bytes.
+              - NACK on address-byte from stretch-timeout (CTRL.NACK_ADDR_AFTER_TIMEOUT)
 
             Stimulus
               - Configure DUT/Agent to Target/Host mode respectively.
-              - Do not proactively fill the TXFIFO, to elicit clock stretching
-              - Stimulate transactions that will fill the ACQFIFO
+              - Elicit clock stretching by
+                - Allowing the ACQFIFO to become full
+                - (reads) starving the TXFIFO for data
               - Only read the ACQFIFO when it is full, possibly adding extra delays so
                 that new transactions while full will be automatically NACK'd
 
             Checking:
-              - Check all bytes (except address bytes) are NACK'd when the ACQFIFO is full.
+              - Check *all* bytes are NACK'd when the ACQFIFO is full.
+                - Address bytes are an exception, as they may/may not be be NACK'd depending
+                  on the configuration of CTRL.NACK_ADDR_AFTER_TIMEOUT.
               - Check that NACKs are generated when the clock is stretched past the threshold
-                configured in TARGET_TIMEOUT_CONTROL.
+                configured in TARGET_TIMEOUT_CTRL.
+                - Check that ACQDATA.NACK_STOP is set in this case.
               - Check that ACQDATA.{NACK,NACKSTART} are set correctly when automatic NACKs have
                 space to add a new entry to the ACQFIFO.
               - Check that ACQDATA.ABYTE contains the NACKd byte when automatic NACKs have space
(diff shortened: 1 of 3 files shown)


hw/ip/i2c/data/i2c_testplan.hjson
```diff
@@ -608,29 +613,83 @@
...
+
+            Stimulus
+            - Configure DUT/Agent as Target/Controller respectively.
+            - Configure TIMEOUT_CTRL with TIMEOUT_CTRL.MODE == BUS_TIMEOUT.
+            - Start a normal transaction using the agent, but halt midway-through and
+              hold SCL low.
+
+            Checking
+            - Both TARGET_EVENTS.BUS_TIMEOUT and CONTROLLER_EVENTS.BUS_TIMEOUT fields should be
+              set upon timeout.
```
a-will, line 691: Note that these two have independent conditions:

`TARGET_EVENTS.BUS_TIMEOUT` only occurs if the bus times out during an active read transfer from our target.
`CONTROLLER_EVENTS.BUS_TIMEOUT` only occurs if the bus times out while our controller is actively engaged in a transaction.

If the DUT is configured as a Target only or if the agent is driving the transaction, then `CONTROLLER_EVENTS.BUS_TIMEOUT` will not be set.

hw/ip/i2c/data/i2c_testplan.hjson
```diff
@@ -608,29 +613,83 @@
...
-              - Data validity continues to match expectation when software uses the 'nak' interrupt
-                to change up the dataflow.
-              - During halt, the bus should be idle (both signals released)
-              - If the 'nak' interrupt is unhandled, after the HOST_NACK_HANDLER_TIMEOUT delay the DUT
-                should be observed to have returned to disabled, and a Stop condition should have been
-                generated to end the active transaction.
-              - Check that the STATUS.HOST_DISABLED_NACK_TIMEOUT bit is set when the timeout is reached.
+              - Data validity continues to match expectation when software uses the 'controller_halt'
+                interrupt to change up the dataflow.
+              - During halt, the bus should be have SCL asserted, SDA released.
```
a-will, line 635: nit:

```suggestion
              - During halt, the bus should have SCL asserted, SDA released.
```

hw/ip/i2c/data/i2c_testplan.hjson
```diff
@@ -608,29 +613,83 @@
...
+    }
+    {
+      name: target_mode_n_byte_ack_control
+      desc: '''
+            Test the (Target-Mode) N-Byte ACK control mode
+            When this mode is enabled, we should expect:
+            - So long as there is space in the ACQ_FIFO, write data bytes will be ACK'd if the
+              auto-decremeting ACK Counter (TARGET_ACK_CTRL.NBYTES) contains a positive value.
+            - If the Auto ACK Counter is zero, the clock is stretched up to the timeout configured in
+              TARGET_TIMEOUT_CTRL. If the timeout is reached (or is configured to zero/disabled), the
```
a-will, line 653: Something like this would be more accurate:

```suggestion
              TARGET_TIMEOUT_CTRL, if the timeout is enabled. If the timeout is reached (or is configured to zero), the
```

If TARGET_TIMEOUT_CTRL is disabled, the target FSM waits / stretches indefinitely.

a-will (approved):
Thanks for putting it all together, and nice work!

Just some nits are in the comments.

hw/ip/i2c/data/i2c_testplan.hjson
```diff
@@ -604,33 +612,328 @@
...
+
+            Stimulus
+            - Configure DUT/Agent as Controller/Agent respectively.
+            - Populate FMTFIFO/RXFIFO for a Controller-Mode transaction.
+            - Configure HOST_TIMEOUT_CTRL with the number of idle cycles.
+            - Enable CTRL.MULTI_CONTROLLER_MONITOR_EN before enabling CTRL.ENABLEHOST
+
+            Checking
+            - Check that the DUT-Controller does not begin transmitting until at least
+              the number of idle cycles in HOST_TIMEOUT_CTRL has elapsed.
```
a-will, line 741: Hm... I wonder if this could be done with an assertion.

hw/ip/i2c/data/i2c_testplan.hjson
```diff
@@ -604,33 +612,328 @@
...
+            - Common I2C bus, with the following connected devices:
+              - DUT
+              - x1 Agent-Controller
+              - x1 Agent-Target
+
+            Stimulus
+            - Configure both controllers to use different timing parameters
+            - Configure both controllers to begin a transaction, where:
+              - A significant portion of the data is identical near the start of the transaction.
+              - The latter data is random, so either may win arbitration.
```
a-will, line 783: This is nice test to have, but if you wanted to short-circuit this, you could probably have a Target Agent that pulls the clock low a little early. Arbitration is technically a separate subject, though synchronization and arbitration loss may both be observed in the same transaction.

Up to you if you want the separate agent in the testbench for a more complete model, though. Compared to that short-circuited approach, it may actually be easier to think about and work with three modeled entities instead of a bunch of special-case "fault" injection code. Hm... but I haven't really thought about how yet another agent's set of monitors should work... Would that be a problem? :thinking:

hw/ip/i2c/data/i2c_testplan.hjson
```diff
@@ -604,33 +612,328 @@
...
+    {
+      name: multi_controller_arbitration_lost_interference
+      desc: '''
+            The DUT-controller should detect other controllers driving the bus simultaneously by observing
+            SDA, and correctly back-off after determining it has lost arbitration for this transaction.
+            After software has addressed the HALT-state due to lost arbitration, and the DUT has awaited the
+            bus to become free again, the DUT-controller should then be able to attempt to send a new transaction.
+
+            This test manually yanks on SDA to induce the DUT into determining it has lost arbitration (There is
+            no multi-controller-aware bus agent in use). Without another controller on the bus to create a valid
```
a-will, line 806: Technically, this can be done with the current "SDA interference" injection in the agent, but as it is now, it's quite unwieldy.

hw/ip/i2c/data/i2c_testplan.hjson
```diff
@@ -604,33 +612,328 @@
...
+            Target-Transmitter with another Target, and observing the conflicting value of SDA. This
+            is possible during read transactions when two connected bus Targets have the same address.
+            When this occurs, the DUT behaves as though it has determined it should NACK, then going
+            on to NACK (release bus) for the remainder of the transaction, effectively awaiting the
+            transaction to end.
+
+            This testpoint uses the TB to manually induce an arbitration loss by pulling SDA low during
+            the transmit-phase of a read transaction. Once the DUT-target has determined it has lost
+            arbitration, it should go on to NACK, causing the transaction to end. Once the Agent-Controller
+            sends the stop condition, the DUT-Idle will return to idle, awaiting the next transaction.
```
a-will, line 874: Did you mean "DUT-Target"?

```suggestion
            sends the stop condition, the DUT-Target will return to idle, awaiting the next transaction.
```

a-will (approved):
LGTM. Most of my responses are comments, with one little nit.

a-will:
> It's difficult to know how far to go with the testplan at this point. I've included testpoints that take the interference-approach to stimulating loss-of-arbitration behaviours, but also testpoints that describe a fuller DV environment that makes use of a multi-controller aware agent on the bus. I feel like we could cross many events with e.g. timing parameter combinations and see the test space explode, so I would be interested if you have any ideas about particularly interesting crosses that may inform the testplan. Though maybe it's worth starting that discussion as a follow-up to this.

I think we should probably limit combinations to the spec'd limits for each mode, then maybe open up a small amount of freedom for some of the parameters that are chip-specific and/or board-specific. This probably will take more thought and discussion, though, haha.

hw/ip/i2c/data/i2c_testplan.hjson
hcallahan-lowrisc: Indeed! Thanks

hcallahan-lowrisc:
CW310 tests in CI seem to be held up, so since this is a documentation-only PR and the lints have passed, I'm going to merge it.

PR #23074, opened 2024-05-13
[hmac, rtl] Fix digest with wipe secret

martin-velay:
- Digest upper part should be zeroed for digest 256 and 384


martin-velay:
Invalid changes

PR #23079, opened 2024-05-13
[ci] Add DV sim_cfg sw_images linter

pamaury:
The goal of this script is to check that all sw_images specified in DV sim_cfg hjson files are valid bazel targets. This avoid DV tests getting out of sync with bazel BUILD files.

Note that this is just a first pass because it only checks that the bazel targets exists but a more thorough check would be to check that the actual output files of the bazel target contain the expected files (e.g. flash image, otp image).

### DIFF
--- a/ci/scripts/check_dv_sw_images.py
+++ b/ci/scripts/check_dv_sw_images.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""Checks for Bazel targets in sim DV cfg files.
+"""
+
+import argparse
+import sys
+import hjson
+from enum import Enum
+from pathlib import Path
+
+from lib.bazel_query import BazelQueryRunner
+
+# List of fake targets added to please the linker:
+FAKE_TARGETS = [
+    "@manufacturer_test_hooks//:example_test_sim_dv",
+]
+
+
+# Copied from hw/top_earlgrey/dv/env/chip_env_pkg.sv for the type definitions.
+class SwType(Enum):
+    ROM = 0  # Ibex SW - first stage boot ROM.
+    TEST_SLOT_A = 1  # Ibex SW - test SW in (flash) slot A.
+    TEST_SLOT_B = 2  # Ibex SW - test SW in (flash) slot B.
+    OTBN = 3  # Otbn SW
+    OTP = 4  # Customized OTP image
+    DEBUG = 5  # Debug SW - injected into SRAM.
+
+
+def check_sw_image(name: str, sw_image: str, valid_bazel_targets: list[str]) -> list[str]:
+    """
+    Check that a sw_image is valid.
+
+    Parameters:
+        name: Name of the test.
+        sw_image: Value of "sw_images" to check.
+        valid_bazel_targets: List of valid bazel targets.
+
+    Returns: list of errors detected.
+    """
+    # The format is sw_image is target:type[:flags]...
+    # where the target is a bazel name so it also contains a ':', for example:
+    #     //sw/device/silicon_creator/rom/e2e:rom_e2e_shutdown_exception_c:1:new_rules
+    #     //sw/device/silicon_creator/rom/e2e:empty_test_slot_b_corrupted:2:ot_flash_binary:signed:fake_ecdsa_prod_key_0
+    # The type indicate whether it's a flash image, otp image, and so on.
+    # The flags can be used for other purposes such as indicating if the DV
+    # requires a signed image.
+    #
+    # See:
+    # - hw/top_earlgrey/dv/env/chip_env_cfg.sv for the actual parsing code
+    # - hw/dv/tools/dvsim/sim.mk for the makefile that copies files
+    # - hw/top_earlgrey/dv/env/chip_env_pkg.sv for the type definitions.
+    try:
+        bazel_module, bazel_label, other = sw_image.split(':', 2)
+        other = other.split(':')
+    except ValueError:
+        return [f"{name}: invalid sw_image '{sw_image}'"]
+    img_type = int(other[0])
+    # If the bazel target does not start with '//' or '@', then prepend '//'
+    # because all targets in valid_bazel_targets use an absolute path.
+    if not bazel_module.startswith('//') and not bazel_module.startswith('@'):
+        bazel_module = '//' + bazel_module
+    bazel_target = f'{bazel_module}:{bazel_label}'
+    # If the type is not OTP or Debug, then add the sim_dv suffix.
+    if SwType(img_type) not in [SwType.OTP, SwType.DEBUG]:
+        bazel_target = f'{bazel_target}_sim_dv'
+
+    if bazel_target not in valid_bazel_targets:
+        return [f"{name}: non-existent target '{bazel_target}' in sw_image={sw_image}"]
+    return []
+
+
+def check_links(cfg_hjson: Path, valid_bazel_targets: list[str]) -> list[str]:
+    """
+    Check all sw_images links in a sim DV cfg hsjon file.
+
+    Parameters:
+        cfg_hjson: Path to the file to check.
+        valid_bazel_targets: List of valid bazel targets.
+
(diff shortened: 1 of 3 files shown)


ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python3
+
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""Checks for Bazel targets that contain banned characters.
+"""
```
jwnrt, line 7: I think this needs updating

jwnrt (approved):
Thanks, this has caught me in the past. LGTM though I can't comment on the quality of the Python.

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+from lib.bazel_query import BazelQueryRunner
+
+# List of fake targets added to please the linker:
+FAKE_TARGETS = [
+    "@manufacturer_test_hooks//:example_test",
+]
+
+
+def check_sw_image(name, sw_image, valid_sw_images):
+    # The format is target:type[:flags]...
```
rswarbrick, line 22: Nit: Probably worth saying "The format of sw_image is..." (it took me a couple of seconds to figure out what was going on with this comment).

The extra colon is also a bit confusing: would it make sense to give an example value in the comment?

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+    # where the target is a bazel name so it also contains a ':'
+    # and the type indicate whether it's a flash image, otp image, and so on.
+    try:
+        bazel_module, bazel_label, other = sw_image.split(':', 2)
+    except ValueError:
+        print(f"- {name}: invalid sw_image '{sw_image}'")
```
rswarbrick, line 28: I think I'd suggest tweaking `check_sw_image` so that it returns a (hopefully empty) list of problems. The `print` lines here could then be turned into `acc += "my message"` and the true/false checks would be replaced with checking whether a list of failures was empty.

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def check_links(cfg_hjson, valid_sw_images):
+    sim_cfg = hjson.load(Path(cfg_hjson).open())
+    ok = True
+    if 'tests' not in sim_cfg:
+        print("{} looks invalid: there are no tests")
+        return False
+    for test in sim_cfg['tests']:
+        name = test.get('name', '<invalid test name>')
```
rswarbrick, line 44: Do we have examples of tests that have no `name:` field? If not, I'd make this a "first level" check and bail instantly if the test was not named.

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+        sw_images = test.get('sw_images', [])
+        for sw_image in sw_images:
+            ok = check_sw_image(name, sw_image, valid_sw_images) and ok
+    return ok
+
+
+def usage():
+    print("usage: check_dv_sw_image.py [hjson cfg files]...")
```
rswarbrick, line 52: I'd really strongly suggest using `argparse`. It's no more complicated than doing this by hand, and it's probably more obvious what's going on.

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+if __name__ == '__main__':
```
rswarbrick, line 55: I'd probably suggest sticking the stuff from here in a `main()` function and calling it in the `if __name__ == '__main__':` block. The resulting code is slightly easier to glance at.

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+    valid_sw_images |= set(FAKE_TARGETS)
+    print("{} sim_dv targets identified".format(len(valid_sw_images)))
+
+    # Verify all files passed as argument
+    if len(sys.argv) == 1:
+        usage()
+        sys.exit(1)
+
+    ok = True
+    for f in sys.argv[1:]:
```
rswarbrick, line 69: I don't massively like this. I'd probably suggest making the Python script take a single hjson file and putting the loop into `check_dv_sw_images.sh`.

ci/scripts/check_dv_sw_images.py
pamaury: Sure, I'll update the comment

ci/scripts/check_dv_sw_images.py
pamaury: One reason to do it this way is that potentially query the list of all bazel targets is an expensive operation. By doing things this way, we can factor the cost and really processing one or ten files does not make a difference in code complexity.

rswarbrick: Oh, good point. In which case, this makes loads of sense. Thanks for explaining.

ci/scripts/check_dv_sw_images.py
rswarbrick: Not really: I just suspect it's actually easier. No strong feelings though.

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+
+from lib.bazel_query import BazelQueryRunner
+
+# List of fake targets added to please the linker:
+FAKE_TARGETS = [
+    "@manufacturer_test_hooks//:example_test",
+]
+
+
+def check_sw_image(name, sw_image, valid_sw_images):
```
engdoreis, line 21: It would be nice to add type annotation to the functions' signature.

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,73 @@
...
+if __name__ == '__main__':
+    bazel = BazelQueryRunner()
+
+    # Search for all targets in device software.
+    valid_sw_images = set(bazel.query("//sw/device/..."))
+    valid_sw_images |= set(FAKE_TARGETS)
+    print("{} sim_dv targets identified".format(len(valid_sw_images)))
+
+    # Verify all files passed as argument
+    if len(sys.argv) == 1:
```
engdoreis, line 64: I suggest using ArgumentParser Lib
https://github.com/lowRISC/opentitan/blob/8b15c287f8b8e5618681b4d9d1a28ea7d400ed72/util/make_new_dif.py#L62

pamaury: done

ci/scripts/check_dv_sw_images.py
pamaury: I added a few annotations.

ci/scripts/check_dv_sw_images.py
pamaury: oops, removed :)

ci/scripts/check_dv_sw_images.py
engdoreis: I think that this function can be deleted now that you are using argparse?

ci/scripts/check_dv_sw_images.py
```diff
@@ -0,0 +1,129 @@
...
+
+from lib.bazel_query import BazelQueryRunner
+
+# List of fake targets added to please the linker:
+FAKE_TARGETS = [
+    "@manufacturer_test_hooks//:example_test_sim_dv",
+]
+
+# Copied from hw/top_earlgrey/dv/env/chip_env_pkg.sv for the type definitions.
+SwTypeRom = 0  # Ibex SW - first stage boot ROM.
```
engdoreis, line 22: Nit.
You can use enums
```python
from enum import Enum
class SwType(Enum):
    ROM = 0
    TEST_SLOT_A = 1

type = SwType.ROM
```

engdoreis (approved):
LGTM.

moidx (approved):
Very nice!

pamaury:
Can we merge this?

jwnrt:
Please could you rebase first so it runs through CI again?
The lint could have broken since it was last run.

pamaury:
sure

PR #23085, opened 2024-05-13
[Sival, spi host] Fix PMOD pin mapping for teacup board.

engdoreis:
This commit fixes the PMOD tests on teacup board.
The teacup board has the SPI_HOST1 sd0 and sd1 pins inverted, so this PR adds a different mapping to work around this problem.

### DIFF
--- a/sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
+++ b/sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
@@ -36,6 +36,7 @@
     },
     {
       "name": "IOA4",
+      "mode": "Input",
       "alias_of": "CN8_10"
     },
     {
--- a/sw/device/lib/testing/BUILD
+++ b/sw/device/lib/testing/BUILD
@@ -467,6 +467,7 @@ cc_library(
     hdrs = ["spi_host_testutils.h"],
     target_compatible_with = [OPENTITAN_CPU],
     deps = [
+        ":pinmux_testutils",
         "//hw/top_earlgrey/sw/autogen:top_earlgrey",
         "//sw/device/lib/base:status",
         "//sw/device/lib/dif:pinmux",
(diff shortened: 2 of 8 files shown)


sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
```diff
@@ -36,6 +36,7 @@
     },
     {
       "name": "IOA4",
+      "mode": "Input",
       "alias_of": "CN8_10"
     },
```
jwnrt, line 41: Aside, is there any way we can give these semantic names in these configs?
Either through comments or `alias_of` etc?

jwnrt (approved):
These changes LGTM, thanks. Weird that sd0 and sd1 were swapped on the teacup board. Was that intentional?

sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
engdoreis: This pin is not used by hyperdebug during this test, but if it's not configured as input it would disturb the PMOD pins as they are shared.

jesultra: It is entirely possible to use `alias_of` to give a name such as `PMOD1_2`, which could maybe better explain why it needs to be configured as `Input`.

But this change causes trouble for ChromeOS, as we use `hyperdebug_teacup.json` in combination with further json files that declare how we want HyperDebug pins to default.   (And one of our files declares `IOA4` as `Alternate`.

I have created PR #23202, which separates this file into one that tries to merely declare physical routing of the Teacup board, which all first and third party users would want to include, and then a separate file with declarations such as this one, that OpenTitan test suites expect HyperDebug to have some default configuration of specific pins, which third party users would not include.  (Even in the OpenTitan repo, some other test suite might want to have HyperDebug drive CN8_10, in order to test something else.)

PR #23091, opened 2024-05-13
[rom_ext_e2e] Document ownership transfer testplan

cfrantz:


sw/device/silicon_creator/rom_ext/data/rom_ext_e2e_testplan.hjson
```diff
@@ -216,5 +216,250 @@
...
+
+    {
+      name: rom_ext_e2e_transfer_any_test
+      desc:
+        '''
+        Verify that ROM_EXT can unlock and transfer ownership to a new owner.
+
+        - Start in an locked & owned state with the `fake` owner as owner.
+        - Unlock the chip into the UnlockedAny state using the `fake` owner unlock key.
+        - Upload an owner block for the `dummy` owner.
```
timothytrippel, line 228: what is the difference between "fake" and "dummy"? is it the same as "fake_owner_1" and "fake_owner_2"?

timothytrippel (approved):

sw/device/silicon_creator/rom_ext/data/rom_ext_e2e_testplan.hjson
cfrantz: Yes: its just terminology to distinguish between different fake (test-only) owners.

Among synonyms for fake, there are several choices:
- dummy
- phony
- sham
- fraud
- hoax
- pirate
- knockoff

moidx: It seems like something like `test-only-owner` or `test-owner` would work right? And the initial state can be `uninitialized-owner` or `default-test-owner`?

sw/device/silicon_creator/rom_ext/data/rom_ext_e2e_testplan.hjson
```diff
@@ -216,5 +216,250 @@
...
+      tags:
+      [
+        rom_ext
+        fpga
+      ]
+      stage: V3
+      tests: []
+    }
+    {
+      name: rom_ext_e2e_flash_permission_test
```
moidx, line 442: Should we add corner cases here to cover:

1. An attempt to override settings for a creator partition.
2. Switching from non-scrambled to scrambled + ECC (and any other interesting permutations). This is to show that the implementation will provide a mechanism for the silicon owner to complete the ownership transfer without requiring bootstrap.

Maybe the two cases above are worth to keep in separate test cases.

moidx (approved):

sw/device/silicon_creator/rom_ext/data/rom_ext_e2e_testplan.hjson
cfrantz: I wanted to have names a bit more distinct than `test-owner-1`/`test-owner-2`, but if you prefer, I can change the use of `fake`and `dummy` to something else.

sw/device/silicon_creator/rom_ext/data/rom_ext_e2e_testplan.hjson
cfrantz: I'll add a test to check that a creator page can't be reconfigured by a bad/malicious owner configuration.

This test (flash_permission_test) supplies a flash configuration that changes the scrambling/ecc page configuration.  I'll amend the text to explicitly state that.

PR #23081, opened 2024-05-13
[rv_dm,dv] Handle jtag disconnection more robustly

rswarbrick:
This PR looks a bit enormous, but this is because it builds on top of two existing PRs:
- #23002 
- #22596 
and these two need merging first.

With those merged, there are four commits that are unique to this PR:
- aee48eb756 `[jtag,dv] Add a mon_vif interface to jtag_agent`
- 9531946853 `[jtag,dv] Add jtag_if_connected event`
- 072396baa6 `[rv_dm,dv] Wire up mon_jtag_if, supporting debug disable & scanmode`
- e4dccd1dc5 `[rv_dm,dv] Trigger jtag_if_connected event on debug enable`

PR #23097, opened 2024-05-14
[jtag,dv] Improve driver sync with TCK

rswarbrick:
This fixes some silliness that crops up when the jtag driver ends up out of sync with the edges of TCK. We could probably make it more robust, but this should be the simplest tweak that gets everything working more predictably.

Note: this PR looks bigger than it is! The first commit is from PR #23081. Only these three commits are unique to this PR:
- c318b028bc `[jtag,dv] Handle trst_n at start of time in jtag_driver.sv`
- 0c5a40cc13 `[jtag,dv] Sync up with TCK before calling drive_jtag_req`
- a52617fd0b `[jtag,dv] Add a check about synchronisation in jtag driver`

### DIFF
--- a/hw/dv/sv/jtag_agent/jtag_driver.sv
+++ b/hw/dv/sv/jtag_agent/jtag_driver.sv
@@ -114,13 +114,9 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
 
   // Drive TMS, TDI to the given values and wait for a single edge of TCK.
   task tms_tdi_step(bit tms, bit tdi);
-    // We normally expect this task to be called at the negedge of tck (synchronous with HOST_CB).
-    // But that won't quite be true if the clock has been paused for a while because tck=1 when
-    // idle. We can spot that happening because tck will be 1. In that situation, wait for HOST_CB
-    // so we can get back in sync.
-    if (cfg.vif.tck) begin
-      @(`HOST_CB);
-    end
+    // When everything is in sync, this task will be called just after the negedge of TCK
+    // (synchronous with HOST_CB). In particular, TCK should be low: check that this is true.
+    `DV_CHECK_FATAL(!cfg.vif.tck);
 
     `HOST_CB.tms <= tms;
     `HOST_CB.tdi <= tdi;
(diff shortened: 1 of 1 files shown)


hw/dv/sv/jtag_agent/jtag_driver.sv
```diff
@@ -57,18 +57,33 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
     `DV_CHECK_FATAL(cfg.if_mode == Host, "Only Host mode is supported", "jtag_driver")
 
     forever begin
+      // Wait until we either go into reset or we get an item to drive.
+      fork : isolation_fork
```
hcallahan-lowrisc, line 61: In OT I normally see the isolation done with a single begin-end block inside the outer fork that contains all the processes we wish to disable. Not sure if it is behaviourally different or not without the begin-end, though I would imagine not?
e.g.
```
fork begin: iso_fork
...
end: iso_fork join
```

hcallahan-lowrisc (commented):
This makes sense to me, but there seems to be a number of chip-level tests failing in the regression as a result.

rswarbrick:
> This makes sense to me, but there seems to be a number of chip-level tests failing in the regression as a result.

Woohoo! I've finally managed to get stuff working again, with a change which is now the first commit in the PR. Would you mind taking another look?

hw/dv/sv/jtag_agent/jtag_driver.sv
```diff
@@ -74,49 +74,82 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
...
+      join
+
       if (!cfg.vif.trst_n) begin
         `DV_WAIT(cfg.vif.trst_n)
         cfg.vif.wait_tck(1);
         drive_jtag_test_logic_reset();
+      end else begin
+        // Since trst_n is not 0, the wait_for_sequences() task must have completed and there must
+        // be an item that we can handle. Retrieve it.
+        seq_item_port.peek(req);
```
hcallahan-lowrisc, line 95: As discussed in person, I think we need to .get() here to consume the item, after peeking that it is not null?

rswarbrick: Well, it didn't quite work the way we both thought(!). It turns out that the item gets consumed by the `item_done` call later, but it also turns out that `peek` didn't work in the way I guessed, and actually blocks until an item is present.

Now fixed, I think.

hcallahan-lowrisc (approved):
This LGTM now, happy to merge when CI passes. Thanks @rswarbrick!

PR #23103, opened 2024-05-14
[rv_dm,dv] Tidy up how we disable scan mode

rswarbrick:
Rather than explicitly setting it to MuBi4False in each vseq, let's move the constraint to the base class. This fixes things for the stress sequence which didn't have the individual copied constraint, and hopefully makes the code a bit clearer.


marnovandermaas (approved):
Yes, thank you.

PR #23104, opened 2024-05-14
[host] Write Rust library wrapper for SPHINCS+ implementation.

jadephilipoom:
Builds on the bindgen library from https://github.com/lowRISC/opentitan/pull/23049 to create more Rust-friendly wrappers for the C functions and a few quick tests.

Next steps: integrate this wrapper with the host-side SPHINCS+ code under `opentitanlib`, remove the `pqcrypto` dependency, and finally unblock https://github.com/lowRISC/opentitan/pull/22953

### DIFF
--- a/sw/host/sphincsplus/lib.rs
+++ b/sw/host/sphincsplus/lib.rs
@@ -2,16 +2,168 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use sphincsplus_bindgen::SPX_BYTES;
+use sphincsplus_bindgen::crypto_sign_keypair;
+use sphincsplus_bindgen::crypto_sign_seed_keypair;
+use sphincsplus_bindgen::crypto_sign_signature;
+use sphincsplus_bindgen::crypto_sign_verify;
+use thiserror::Error;
 
-pub fn do_something() -> u32 {
-    SPX_BYTES
+// SPHINCS+ secret key byte length.
+pub const SPX_SECRET_KEY_BYTES: usize = sphincsplus_bindgen::CRYPTO_SECRETKEYBYTES as usize;
+pub const SPX_PUBLIC_KEY_BYTES: usize = sphincsplus_bindgen::CRYPTO_PUBLICKEYBYTES as usize;
+pub const SPX_SIGNATURE_BYTES: usize = sphincsplus_bindgen::CRYPTO_BYTES as usize;
+pub const SPX_SEED_BYTES: usize = sphincsplus_bindgen::CRYPTO_SEEDBYTES as usize;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SpxPublicKey([u8; SPX_PUBLIC_KEY_BYTES]);
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SpxSecretKey([u8; SPX_SECRET_KEY_BYTES]);
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SpxSignature([u8; SPX_SIGNATURE_BYTES]);
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SpxSeed([u8; SPX_SEED_BYTES]);
+
+#[derive(Debug, Error)]
+pub enum SpxError {
+    #[error("SPHINCS+ key generation failed with error code {0}")]
+    KeyGen(i32),
+
+    #[error("SPHINCS+ signature generation failed with error code {0}")]
+    SigGen(i32),
+
+    #[error("Unexpected signature length {0}")]
+    BadSigLength(usize),
+
+    #[error("Signature did not pass verification")]
+    BadSignature,
+}
+
+// Generate a new keypair from a seed.
+pub fn spx_keypair_from_seed(seed: &SpxSeed) -> Result<(SpxPublicKey, SpxSecretKey), SpxError> {
+    let mut pk = [0u8; SPX_PUBLIC_KEY_BYTES];
+    let mut sk = [0u8; SPX_SECRET_KEY_BYTES];
+    let err_code =
+        // SAFETY: the buffers here are all fixed-length arrays of the size expected by the C code.
+        unsafe { crypto_sign_seed_keypair(pk.as_mut_ptr(), sk.as_mut_ptr(), seed.0.as_ptr()) };
+    if err_code != 0 {
+        return Err(SpxError::KeyGen(err_code));
+    }
+    Ok((SpxPublicKey(pk), SpxSecretKey(sk)))
+}
+
+// Generate a new random keypair.
+pub fn spx_keypair_generate() -> Result<(SpxPublicKey, SpxSecretKey), SpxError> {
+    let mut pk = [0u8; SPX_PUBLIC_KEY_BYTES];
+    let mut sk = [0u8; SPX_SECRET_KEY_BYTES];
+    let err_code =
+        // SAFETY: the buffers here are all fixed-length arrays of the size expected by the C code.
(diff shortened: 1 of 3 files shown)


sw/host/sphincsplus/lib.rs
```diff
@@ -2,16 +2,111 @@
...
 
-pub fn do_something() -> u32 {
-    SPX_BYTES
+// SPHINCS+ secret key byte length. 
+pub const SPX_SECRET_KEY_BYTES : usize = sphincsplus_bindgen::CRYPTO_SECRETKEYBYTES as usize;
+pub const SPX_PUBLIC_KEY_BYTES : usize = sphincsplus_bindgen::CRYPTO_PUBLICKEYBYTES as usize;
+pub const SPX_SIGNATURE_BYTES : usize = sphincsplus_bindgen::CRYPTO_BYTES as usize;
+pub const SPX_SEED_BYTES : usize = sphincsplus_bindgen::CRYPTO_SEEDBYTES as usize;
+
+#[derive(Clone, Debug, PartialEq)]
```
jwnrt, line 16: Nit:
```suggestion
#[derive(Clone, Debug, PartialEq, Eq)]
```

sw/host/sphincsplus/lib.rs
```diff
@@ -2,16 +2,111 @@
...
+// Generate a new keypair from a seed. 
+pub fn spx_keypair_from_seed(seed: &SpxSeed) -> (SpxPublicKey, SpxSecretKey) {
+    let mut pk = [0u8; SPX_PUBLIC_KEY_BYTES];
+    let mut sk = [0u8; SPX_SECRET_KEY_BYTES];
+    let err = unsafe { crypto_sign_seed_keypair(pk.as_mut_ptr(), sk.as_mut_ptr(), seed.0.as_ptr()) };
+    if err != 0 {
+        panic!("SPHINCS+ seeded key pair generation failed with error code {}.", err);
+    }
+    (SpxPublicKey(pk), SpxSecretKey(sk))
+}
```
jwnrt, line 37: Nit: would be nice to make this function fallible instead of panicking

```suggestion
pub fn spx_keypair_from_seed(seed: &SpxSeed) -> anyhow::Result<(SpxPublicKey, SpxSecretKey)> {
    let mut pk = [0u8; SPX_PUBLIC_KEY_BYTES];
    let mut sk = [0u8; SPX_SECRET_KEY_BYTES];
    let err = unsafe { crypto_sign_seed_keypair(pk.as_mut_ptr(), sk.as_mut_ptr(), seed.0.as_ptr()) };
    anyhow::ensure!(err != 0, "SPHINCS+ seeded key pair generation failed with error code {err}.");
    Ok((SpxPublicKey(pk), SpxSecretKey(sk)))
}
```

jwnrt (commented):
LGTM, thanks

I think it might need to go through `bazel run //quality:rustfmt_fix` and may also need adding to `RUST_TARGETS` in `quality/BUILD`

Also we don't really require this in OpenTitan, but if you feel like you could add a `// SAFETY: ...` comment to these unsafe blocks [explaining why they're safe](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks) (e.g. the pointer arguments are valid and point to buffers of the right size) that would be helpful!

RyanTorok (approved):
LGTM. I would take care of @jwnrt 's comment about returning `Result` instead of panicking, and we should be good to go.

sw/host/sphincsplus/lib.rs
jadephilipoom: Added `Eq` to all wrapper structs.

sw/host/sphincsplus/lib.rs
jadephilipoom: Fixed, thanks!

sw/host/sphincsplus/lib.rs
```diff
@@ -2,16 +2,111 @@
...
+    if err != 0 {
+        panic!("SPHINCS+ random key pair generation failed with error code {}.", err);
+    }
+    (SpxPublicKey(pk), SpxSecretKey(sk))
+}
+
+// Generate a detached signature for the message using the secret key. 
+pub fn spx_sign(sk: &SpxSecretKey, msg: &Vec<u8>) -> SpxSignature {
+    let mut sig = [0u8; SPX_SIGNATURE_BYTES];
+    let err = unsafe { crypto_sign_signature(sig.as_mut_ptr(), &mut sig.len(), msg.as_ptr(), msg.len(), sk.0.as_ptr()) };
```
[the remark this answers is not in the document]
jadephilipoom: I added a new mutable int for the size returned by the C code, and I check it now against `sig.len()` later. I also added this to the new `// SAFETY:` comment above this line.

jadephilipoom:
> I think it might need to go through `bazel run //quality:rustfmt_fix` and may also need adding to `RUST_TARGETS` in `quality/BUILD`

Both of these should now be addressed.

> Also we don't really require this in OpenTitan, but if you feel like you could add a `// SAFETY: ...` comment to these unsafe blocks [explaining why they're safe](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks) (e.g. the pointer arguments are valid and point to buffers of the right size) that would be helpful!

Done!

jwnrt (approved):

sw/host/sphincsplus/lib.rs
```diff
@@ -2,16 +2,149 @@
...
+pub fn spx_keypair_generate() -> anyhow::Result<(SpxPublicKey, SpxSecretKey)> {
+    let mut pk = [0u8; SPX_PUBLIC_KEY_BYTES];
+    let mut sk = [0u8; SPX_SECRET_KEY_BYTES];
+    let err_code =
+        // SAFETY: the buffers here are all fixed-length arrays of the size expected by the C code.
+        unsafe { crypto_sign_keypair(pk.as_mut_ptr(), sk.as_mut_ptr()) };
+    anyhow::ensure!(
+        err_code == 0,
+        "SPHINCS+ random key pair generation failed with error code {err_code}."
+    );
```
cfrantz, line 52: Please return structured errors from library functions.  The `anyhow` ensure/bail functions, when taking a string as the error argument, basically are returning a "stringly-typed" error.  Should a user of this library want to handle or otherwise make behavior decisions based on the error, stringly-typed errors then require string parsing.

My recommendation is to avoid the use of the `anyhow` crate in low-level base libraries.  You can use the `thiserror` crate to easily create nice error types and return those errors via `std::result::Result`.  Higher-level code is then free to handle such errors or thunk those errors into an `anyhow::Result` via `?` coercion.

Based on what I see in this PR, I'd perhaps start like this:
```
use thiserror::Error;

#[derive(Debug, Error)]
pub enum SpxError {
    #[error("Key generation failed with error {0}")]
    KeyGen(u32),

    #[error("Unexpected length {0}")]
    BadLength(u32),
}
```

sw/host/sphincsplus/lib.rs
```diff
@@ -2,16 +2,149 @@
...
+        unsafe {
+        crypto_sign_verify(
+            sig.0.as_ptr(),
+            sig.0.len(),
+            msg.as_ptr(),
+            msg.len(),
+            pk.0.as_ptr(),
+        )
+    };
+    err_code == 0
```
cfrantz, line 101: You may want to look at the verification functions in some of the RustCrypto libraries we use.  IIRC, it's fairly typical to return `Result<(), Error>` from those functions, with `Ok(())` indicating verification success.

sw/host/sphincsplus/lib.rs
jadephilipoom: Fixed!

sw/host/sphincsplus/lib.rs
jadephilipoom: You're right, should be addressed now.

cfrantz (approved):

PR #23105, opened 2024-05-14
[Sival, pwrmgr] Fix chip_power_sleep_load_test with ROM_EXT 

engdoreis:
Adds the chip_power_sleep_load_test to SiVal.
The test was marked as broken because it was attempting to write to OTP regs, which are block by ROM_EXT in the ePMP.


pamaury (approved):
Why was it marked as broken before?

engdoreis:
> Why was it marked as broken before?

Segfault caused by the OTP.
https://github.com/lowRISC/opentitan/pull/23105/commits/875f521b3ef46b9e1076586c2a61da0d3cd49b05

PR #23106, opened 2024-05-14
[rv_dm,dv] Add a slight wait when enabling debug in rv_dm_base_vseq

rswarbrick:
This is a small tweak that gets things to line up for some seeds in the FSM smoke sequence.


hcallahan-lowrisc (approved):
Makes sense to me. Thanks :)

Issue #22990, opened 2024-05-07
[pinmux] Remove AST_MISC pad, add IE bit for all pads


vogelpi:
### Description

As per email by @meisnere , we should make the following two changes:
- [x] Remove the AST_MISC pad (will be replaced by a VSS supply pad) --> #22995
- [x] Add an IE control bit for each pad for leakage control --> #23042
  - [x] including also some manual pads (see email) --> #23107

andreaskurth:
Closing as completed in #22995, #23042, and #23107.

PR #23107, opened 2024-05-14
 [top_earlgrey,sensor_ctrl] Add input disable attribute for manual pads

andreaskurth:
This PR adds _input disable_ control for the so-called manual pads CC1, CC2, FLASH_TEST_MODE0, and FLASH_TEST_MODE1.

This contributes to resolving #22990.

Instead of going through pinmux, which is auto-generated, this is implemented in sensor_ctrl, which is written specifically for top_earlgrey. If we see the need to implement this functionality also for other top-level designs, we can extend pinmux accordingly and remove this from sensor_ctrl.

## TODOs (post merge --> create issue)
- [ ] Add sensor_ctrl DIF functions.


a-will (approved):
LGTM

There are maybe some nits in the use of `manual` in the CSR name and using the `hwext` type (unclear to me what it's used for), but they don't seem important.

PR #23109, opened 2024-05-14
[sival] fix pmp_smoketest_{napot, tor} tests in rom_ext

nbdd0121:
Instead of unlocking all address space by unsetting all PMP registers, these tests only need 1 or 2 spare PMP regions. Choose regions 6/7 for this purpose. A test is added to ensure that these regions are not configured.

### DIFF
--- a/sw/device/lib/runtime/pmp.h
+++ b/sw/device/lib/runtime/pmp.h
@@ -286,6 +286,17 @@ pmp_region_configure_result_t pmp_region_configure_tor(
     uintptr_t address_start, uintptr_t address_end);
 
 /**
+ * Check if the requested region is configured.
+ *
+ * @param region PMP region to query the information for.
+ * @param[out] configured Whether the PMP region is configured.
+ * @return `pmp_region_configure_result_t`.
+ */
+OT_WARN_UNUSED_RESULT
+pmp_region_configure_result_t pmp_region_is_configured(
+    pmp_region_index_t region, bool *configured);
+
+/**
  * Queries the lock status for the requested region.
  *
  * @param region PMP region to query the lock information for.
--- a/sw/device/tests/pmp_smoketest_napot.c
+++ b/sw/device/tests/pmp_smoketest_napot.c
@@ -57,12 +57,13 @@ static void pmp_configure_load_napot(void) {
 OTTF_DEFINE_TEST_CONFIG();
 
 bool test_main(void) {
-  // Check that `PMP_LOAD_REGION_ID` region is not already used.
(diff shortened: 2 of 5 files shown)


sw/device/tests/pmp_smoketest_napot.c
```diff
@@ -54,6 +57,12 @@ static void pmp_configure_load_napot(void) {
 OTTF_DEFINE_TEST_CONFIG();
 
 bool test_main(void) {
+  // Check that `PMP_LOAD_REGION_ID` region is not already used.
+  uint32_t pmpcfg1 = 0;
+  CSR_READ(CSR_REG_PMPCFG1, &pmpcfg1);
+  CHECK((pmpcfg1 & 0xff000000) == 0,
```
jwnrt, line 63: Nit: I'm not sure I like that `0xff000000` is hard-coded here but is tied to `PMP_LOAD_REGION_ID` being `7`. I understand it would be annoying to select the register and byte offset based on this constant, though.

If you don't want to calculate it, could you add a `static_assert(PMP_LOAD_REGION_ID == 7)` or `&&` that into the `CHECK` so it always stays up to date?

jwnrt (approved):
LGTM, thanks

sw/device/tests/pmp_smoketest_napot.c
nbdd0121: I added a `pmp_region_is_configured` function to `runtime/pmp`

jwnrt: Very nice, thanks.

Would you mind sticking this in `sram_ctrl_execution_test` while we're here?

nbdd0121: The current pmp API doesn't seem to be able to handle the napot usage in `sram_ctrl_execution_test`. `0x7fffffff` napot means that `2**34` bytes and this overflows the 32-bit integer API.

sw/device/lib/runtime/pmp.h
```diff
@@ -285,6 +285,17 @@ pmp_region_configure_result_t pmp_region_configure_tor(
     pmp_region_index_t region_end, const pmp_region_config_t config,
     uintptr_t address_start, uintptr_t address_end);
 
+/**
+ * Check if the requested region is configured.
+ *
+ * @param region PMP region to query the information for.
+ * @param configured Whether the PMP region is configured.
```
engdoreis, line 292:
```suggestion
 * @param[out] configured Whether the PMP region is configured.
```

engdoreis (approved):
LGTM, just a nit.

PR #23058, opened 2024-05-10
[sival, spi_dev] Move test harness to a common dir

engdoreis:
Move test harness to a common folder as the UART tests are.

PR #23111, opened 2024-05-14
[sival, spi_dev] Move test harness to a common dir

engdoreis:
This is a Backport PR of https://github.com/lowRISC/opentitan/pull/23058


timothytrippel (approved):

sw/host/tests/chip/spi_device/BUILD
```diff
@@ -0,0 +1,56 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+load("@rules_rust//rust:defs.bzl", "rust_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_binary(
+    name = "spi_device_flash_smoketest",
+    srcs = [
+        "src/spi_device_flash_smoketest.rs",
```
a-will, line 12: Should the named binaries like this one be located at `src/bin/*.rs`? It looks like that would follow the cargo's conventions for package layout (and maybe satisfy clippy, though I'm not sure if we need a default executable too...).

a-will (commented):
Interesting. It looks like clippy demands specific layouts, and the default binary is supposed to be at `src/main.rs`.

PR #23112, opened 2024-05-14
[runtime,log] Make logging usable in DIFs

pamaury:
Presently, it is impossible to use the logging macros in DIFs for somewhat obscure reasons. This is because the DIFs are compiled on host for opentitanlib bindgen, and it fails with a really difficult to debug error about signing select having wrong values. On the other hand, being able to print from the DIFs is really useful when debugging.

This commit change the log library so that it can compile on host, hence enabling usage in DIFs. It requires a bit of ugly conditional compilation because logging in DV relies on magic symbols.

### DIFF
--- a/sw/device/lib/runtime/BUILD
+++ b/sw/device/lib/runtime/BUILD
@@ -62,7 +62,6 @@ cc_library(
     name = "log",
     srcs = ["log.c"],
     hdrs = ["log.h"],
-    target_compatible_with = [OPENTITAN_CPU],
     deps = [
         "//sw/device/lib/arch:device",
         "//sw/device/lib/base:macros",
--- a/sw/device/lib/runtime/log.c
+++ b/sw/device/lib/runtime/log.c
@@ -2,13 +2,18 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-#include "sw/device/lib/runtime/log.h"
+/* On host, we need access to memrchr */
+#ifndef OT_PLATFORM_RV32
+#define _GNU_SOURCE
+#include <string.h>
+#endif
 
 #include <assert.h>
 
 #include "sw/device/lib/arch/device.h"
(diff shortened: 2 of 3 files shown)


luismarques (approved):

jwnrt (approved):

a-will:
Contrary opinion: I think the DIFs should not have logging. Logs have costs, and DIFs are the innermost loops of inner loops, making those costs potentially be born at a very high impact. (One way around that: Do you plan to create a conditional compilation setup where these logging statements are not ordinarily compiled in?)

Moreover, there should not be any complex logic in DIFs that makes it necessary to print anything out inside the DIFs. The user of the DIFs can print out what was returned and get complete information about what happened. If there is some complex logic, then those DIFs need to be rewritten, and the logic should be split out into testutils.

jwnrt:
I somewhat agree that DIFs shouldn't log, but it is helpful for debugging. I guess `dbg_printf` can replace that use-case though, so not a huge issue.

engdoreis:
I'm also contrary to have logs in the DIFs because it would make one DIF depend on the other (UART). If a DIF is suspected to have a bug, we should use the unittests to debug it with GDB.

luismarques:
Amaury made this PR after I tried to use logging in a DIF for debugging purposes. I also don't think we should commit dif code that logs, but having that functionality for debugging purposes seems valuable. The alternative is to basically cherry-pick this code manually every time we want to "printf debug" with `LOG_INFO`s, etc.

jwnrt:
I would recommend trying to use the `dbg_printf` functions. I think they work from DIFs and all you're really losing is the line number that the log happened at (which you can manually print yourself).

pamaury:
I should point out that:
- the goal of this PR was partly to fix an issue that if you ever tried to use `log` from within a DIF, you will get an inscrutable build error (bazel being really unhelpful here),
- you can't call `dbg_printf` in a test because it is only defined in the ROM
- some DIF functions are actually nontrivial but at the same time not necessary super easy to split into several functions.

PR #23115, opened 2024-05-14
[Sival] Map unmapped tests in testplan.

engdoreis:


hw/top_earlgrey/data/ip/chip_otbn_testplan.hjson
```diff
@@ -157,5 +157,107 @@
...
+      desc: '''Check that the OTBN can run RSA.
+
+            Check that OTBN can run RSA to:
+             - Encrypt data.
+             - Decrypt data.
+             - Sign data.
+             - Verify sined data.
+            With the keys sizes 2048, 3072 and 4096 bits.
+      '''
+      stage: V3
```
moidx, line 171: Is it possible to mar these test cases for SiVal only? I am not sure we can support them in DV given the test run time.

moidx: The same applies to other test cases here. I recommend to tracking cryptolib test cases in a separate testplan al together. Similar to how we did with the ROM e2e testplan.

jadephilipoom: Agreed, the RSA tests (especially keygen below) take quite a long time with the larger keys -- to be more precise, ~tens of millions of cycles for signing or decryption  (the `*_signature_functest` targets do sign-then-verify, and the `*_encryption_functest` ones do encrypt-then-decrypt) , and keygen can easily go to hundreds of millions of cycles.

jadephilipoom (approved):

hw/top_earlgrey/data/ip/chip_otbn_testplan.hjson
engdoreis: Moved the new testpoints to `sw/device/lib/crypto/data/crypto_testplan.hjson` and changed `stage` to `NA`.

moidx (approved):

hw/top_earlgrey/data/ip/chip_pwrmgr_testplan.hjson
```diff
@@ -62,7 +62,10 @@
         "PWRMGR.LOW_POWER.WAKE_INFO"
       ]
       tests: ["chip_sw_pwrmgr_random_sleep_all_wake_ups"]
-      bazel: ["//sw/device/tests:pwrmgr_random_sleep_all_wake_ups"]
+      bazel: [
+        "//sw/device/tests:pwrmgr_random_sleep_all_wake_ups",
+        "//sw/device/tests:pwrmgr_sleep_wake_5_bug_test"
```
jwnrt, line 67: Should this test have been mapped in ES @engdoreis?

It's a known bug on this branch that will always fail.

engdoreis: In that case, we should remove the `silicon_owner_sival_rom_ext` exec_env and unmap it.

jwnrt: Okay, I've opened https://github.com/lowRISC/opentitan/pull/23344 to remove the test from this branch entirely. It will persist as a regression test in `master`.

The test isn't useful with the silicon environments removed so I've deleted the whole thing.

PR #23116, opened 2024-05-14
Sar dv

martin-velay:
This PR is only for debug purpose, it shouldn't be merged !


martin-velay:
@andreaskurth and @gdessouky
I have pushed my ongoing work related to the Save and Restore feature. This PR goal is only to share this code with you to reproduce the following issue at your end if you have time to help.

image

As you can see on this image:
1. we are triggering a hash_stop as message_length counter is on the 512 boundary (we are in SHA-256 mode).
2. interrupt hmac_done is happening
3. But the digest is not updated as sha_hash_done is not raised

To reproduce this issue you can use the following command (if you are using Xcelium you can run it in GUI mode and I have enabled the breakpoints):
util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 114132373073648776435301108104832965337850972304806816025000095479785483310546

Note: you'll see the same issue whatever the seed as I have added some constraints

andreaskurth:
Thx for providing this snapshot of your work, @martin-velay. I ran `hmac_smoke` with the seed you posted and could reproduce your waves. I noticed a few things that I think are problematic on the DV side:
1. When HMAC is stopped for a context swap (i.e., the `CMD.hash_stop` bit is set), the content of the digest registers is expected to be the _intermediate_ digest of the HMAC operation, not the _final_ digest. To visualize this in an example:
image
  When HMAC is stopped at the red arrow, the intermediate digest includes the inner padded key and N message blocks (two in this example), but it does _not_ include hash of the entire first round and the outer padded key (blue cross). The _expected_ digest in the scoreboard, however, seems to come from a DPI call that computes the _final_ digest, i.e., if HMAC was told to _process_ at the red arrow. I think this is why the values read from the digest registers don't match the expectation. Now ideally we would have golden model functions to compute the intermediate digest, and I think the cryptoc DPI code can be extended to expose this -- though @gdessouky likely has much more knowledge on this. (When I did basic verification while implementing the feature, I did not check the intermediate values. Instead, I just saved, cleared, and restored the digest registers and then told HMAC to continue, and in the end I checked that the final digest matched the golden model.) 
2.

martin-velay:
Hi Andreas,
Thanks a lot for your analysis, I understand better now what is happening. I'll continue from this point. And sorry for making you work that late !

gdessouky:
Thanks for sharing @martin-velay and your feedback @andreaskurth, yes indeed these are the intermediate digest values, so comparing with predicted values from the C model will not work. This is where the C model is https://github.com/lowRISC/opentitan/tree/master/hw/ip/hmac/dv/cryptoc_dpi. I'm sure we can adapt it to generate and compare intermediate digest values for thorough confidence on functionality, but this will take some work and debugging still. I think if we compare final digest values for now and confirm they match prediction, then we already have very good confidence on functionality. My suggestion would be to de-prioritize this at the moment in favor of higher-prio changes required, and we can spin out an issue to track this for DV improvements once we all have more bandwidth, WDYT?

martin-velay:
Thanks again @andreaskurth for your support, it helped me a lot to be able to move forward. I have pushed some fixes, now the first S&R with same context is working. Still with the same seed I have another S&R which triggered and failing. I'll investigate on my own and keep you informed.

@gdessouky about your last comment above: no problem to update the C model later. You'll see in my last commit that I am skipping the comparison of the intermediate digest for the moment 👌

martin-velay:
My last commit [#af69f1b](https://github.com/lowRISC/opentitan/pull/23116/commits/af69f1bceba4a7da0858dfa5fc27511833897231) is addressing the point 1 mentioned in my comment from yesterday. Now the digest is not matching because the digest is not updated at the right time as mentioned yesterday. I'll now try to understand if the reason is the RTL/TB.

martin-velay:
Here is the issue I am observing with the latest commit and still with the same command:

util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 114132373073648776435301108104832965337850972304806816025000095479785483310546

image

The following line is triggered as the message_length is indeed =512.
https://github.com/lowRISC/opentitan/blob/349bab6601c24b1d01b7264296bfc1c3bcd6c631/hw/ip/hmac/rtl/hmac.sv#L173
Then, as hash_running is 0, the hash_done_event is triggered but the digest has been updated too early.
I guess something is wrong there, I don't know enough the RTL to say precisely what's going on.

Could you have a look @gdessouky / @andreaskurth ?

martin-velay:
After the lasts commits, verification is slightly moving forward. There is now another issue, but I haven't found yet the root cause. I can just tell you that intermediate digest seems to match (check with debug prints in the C files added with Andreas help) but the final digest is not matching. To be investigated.

martin-velay:
@andreaskurth I am now comparing the waves when it's working with a similar S&R against the non-working one. Both of them are triggering a S&R while the `message_length` counter is equal to 512.
I could notice 2 main differences so far, but still no conclusion:
- the intermediate HMAC hash is using some pieces of the message when working properly, while empty message when failing but I could see that intermediate digest is matching.
- the `round_q` signal from `prim_sha2`, is always counting up to 63, but when not working it counts only up to 26 and goes back to 0 because `sha_en_i` goes to 0.
I don't have any idea if this might be an explanation so far. I am sharing it in case you are working on it in parallel.

andreaskurth:
> * the `round_q` signal from `prim_sha2`, is always counting up to 63, but when not working it counts only up to 26 and goes back to 0 because `sha_en_i` goes to 0.

Ah, interesting! Yes, bringing `sha_en_i` to zero indeed will clear the round counter. This `sha_en` signal comes from the CSR, right? Does the vseq disable it before the digest computation is complete (`hmac_done_o`)?

martin-velay:
The vseq is waiting for the `intr_hmac_done_o` to be high, but I feel that the `intr_hmac_done_o` is happening too early. I guess this 2nd round should happen.

image

martin-velay:
With this new fact, I am now almost convinced that we have another RTL bug.
I haven't got a solution yet, but we need to find a way to trigger the `intr_hmac_done_o` only once SHA core has processed all the current message after the stop. The `round_q` shouldn't be interrupted. Maybe wait until it moves to Idle or something else.
If you have time to have a look @andreaskurth, this might be part of your [PR](https://github.com/lowRISC/opentitan/pull/23224) as it is linked somehow.

andreaskurth:
Ah great, thanks for the pointer @martin-velay! I'll take a look

andreaskurth:
I found and fixed multiple problems in `prim_sha2_pad`, and I added them to PR #23224. With those fixes, the case you mentioned above (`util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 114132373073648776435301108104832965337850972304806816025000095479785483310546`) passes! I pushed the commits onto this branch as well for your convenience.

I even got 100% pass rate when I ran `util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium`! So things look much better for the interaction we currently cover, but I'm sure you have a few tricks up your sleeve to stress HMAC's save & restore feature even more :wink:

martin-velay:
Excellent! Thanks a lot for your support @andreaskurth. I'll now start to prepare for merging this part with the master. And then I'll show my sleeves tricks.

martin-velay:
Last commit is making message bursts working better while using S&R. I will start a regression now and continue to fix all the fails before unleashing more the TB (as very constrained at the moment).

martin-velay:
@andreaskurth 
I think there is an issue with the HMAC Idle.

Could you run this simu and have a look? It will fail on the 4th message. Take the latest database from this branch and run this command:
`util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 1`

This is what you'll see:
image

andreaskurth:
> @andreaskurth I think there is an issue with the HMAC Idle.
> 
> Could you run this simu and have a look? It will fail on the 4th message. Take [this database](https://github.com/lowRISC/opentitan/pull/23116/commits/219acdb3db1b2af89db381dd60ba84db628c9614) from this branch and run this command: `util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 1`
> 
> This is what you'll see: image

Just looking at the waves you posted, I don't think `idle_o` behaves incorrectly: hmac is told to stop, then waits until the intermediate digest is computed, after which it raises the `hmac_done` interrupt as well as the `idle_o` status output. `idle_o` is correct not to raise immediately after hmac is told to stop, because it's used for clock gating so it must only rise when hmac doesn't have ongoing operations before it can be clock-gated.

martin-velay:
image> > @andreaskurth I think there is an issue with the HMAC Idle.
> > Could you run this simu and have a look? It will fail on the 4th message. Take [this database](https://github.com/lowRISC/opentitan/pull/23116/commits/219acdb3db1b2af89db381dd60ba84db628c9614) from this branch and run this command: `util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 1`
> > This is what you'll see: image
> 
> Just looking at the waves you posted, I don't think `idle_o` behaves incorrectly: hmac is told to stop, then waits until the intermediate digest is computed, after which it raises the `hmac_done` interrupt as well as the `idle_o` status output. `idle_o` is correct not to raise immediately after hmac is told to stop, because it's used for clock gating so it must only rise when hmac doesn't have ongoing operations before it can be clock-gated.

The problem is exactly that: the `hmac_done` interrupt is raised but not the `idle_o`. This is visible on the screenshot I've shared, `idle_o` remains on `_MuBi4False_` state.

martin-velay:
I have done RTL changes, it looks better now but I'll run a regression to be sure of my changes.
See this #23336

gdessouky:
Thanks @martin-velay, I've looked into the waveforms at this commit, and indeed there is a bug with the computation of `txcnt_eq_blksz` that was masked up until the SAR feature because it is actually irrelevant in signalling the finalization of the computation and asserting idle, but it becomes relevant here.

martin-velay:
I have pushed my DV modifs to be able switch between different contexts. It's not complete yet.

But now we should be able to debug this [issue #23352](https://github.com/lowRISC/opentitan/issues/23352).

This command may reveal it with [this database](https://github.com/lowRISC/opentitan/pull/23116/commits/0dfc8e56bfaa9daa1d166dd2a2ad9f75fbcf09e4):
_util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 1_

@gdessouky, I let you have a look if you have bandwidth today. Please update me if you find something.

@ballifatih FYI

gdessouky:
> I have pushed my DV modifs to be able switch between different contexts. It's not complete yet.
> 
> But now we should be able to debug this [issue #23352](https://github.com/lowRISC/opentitan/issues/23352).
> 
> This command may reveal it with Xcelium and with [this database](https://github.com/lowRISC/opentitan/pull/23116/commits/0dfc8e56bfaa9daa1d166dd2a2ad9f75fbcf09e4): _util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --gui --fixed-seed 1_
> 
> @gdessouky, I let you have a look if you have bandwidth today. Please update me if you find something.
> 
> @ballifatih FYI

Thanks @martin-velay for the DV setup to reproduce the issue!  I've pushed an RTL fix https://github.com/lowRISC/opentitan/pull/23432 that passes this DV smoke test, but I wasn't sure if you wanted me to push it to this branch here, so I leave that up to you.

martin-velay:
Thanks a lot @gdessouky for your fix #23432 ! I'll continue to test this feature and keep you informed.

hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
```diff
@@ -147,7 +151,10 @@ class hmac_smoke_vseq extends hmac_base_vseq;
         end
 
         // msg stream in finished, start hash
-        if (do_hash_start) trigger_process();
+        if (do_hash_start) begin
+          `uvm_info(`gfn, $sformatf("triggering process because msg stream in finished"), UVM_LOW)
```
gdessouky, line 155: minor typo: do you mean "is finished" here?

martin-velay: I guess yes, this modif is from Andreas actually. I'll change it

gdessouky:
> ```
> util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --fixed-seed 114185448211150958355807046043715683092785711447270148054926510481266628632264
> ```

thanks @martin-velay for raising, looking into it now.

gdessouky:
> I am currently releasing some constraints after having in place the S&R with different contexts (not complete yet, still remain to add some additional checks). This is leading me into a potential issue where the HMAC Done is never triggered. I am currently trying to understand if it's truly an RTL bug or not.
> 
> To reproduce it with Xcelium:
> 
> ```
> git checkout b242d6c90327aed83a1f67fbf1518bf4734741d8
> util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --fixed-seed 114185448211150958355807046043715683092785711447270148054926510481266628632264
> ```
> 
> With VCS, use seed: `83812634503382603449251114596099188419278643638756659410240524018062663081103`
> 
> @gdessouky please could you have a look?

thanks @martin-velay looking into it now.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -431,4 +463,219 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
     end
   endtask
 
+  // Stop hash, save current context, do something/nothing and restore context
+  //    - Test with context A saved and restored
+  //    - Test with context A and B, alternatively saved and restored. Ensure to randomize again:
+  //      key length, digest size, digest swap, endian swap and secret key
+  virtual task save_and_restore();
```
martin-velay, line 470: @ballifatih could you review this piece of code, to be sure that my seq is as the SW will do?

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -431,4 +463,219 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
...
+      `uvm_info(`gfn, $sformatf("Stop and trigger continue only"), UVM_LOW)
+      sar_ongoing = 1;
+      // Stop hash operations.
+      trigger_hash_stop();
+      // Expose ongoing Save and Restore triggered to avoid to request a new hash process
+      save_ctx_ongoing = 1;
+      // Wait for hash to be done so the digest is updated.
+      csr_spinwait(.ptr(ral.intr_state.hmac_done), .exp_data(1'b1));
+      // Clear the interrupt.
+      csr_wr(.ptr(ral.intr_state.hmac_done), .value(1'b1));
```
ballifatih, line 502: Since the last changes, I am using `idle` bit to wait HMAC IP, but `hmac_done` can also be used. I am using `idle` because I wanted to test that it works. I just wanted to share that, but I do not have any strong reason for the one or the other.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -431,4 +463,219 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
...
+      save_ctx_ongoing = 0;
+      trigger_hash_continue();
+      sar_stop_continue_ev.trigger();
+    end else if (sar_ctxt == SameContext) begin
```
ballifatih, line 506: I do not know if `KEY` and `CFG` registers are reset when we stop.

If they are resetting, we need to reconfigure them during restoring.

If they are kept intentionally, then it is good that we are not restoring them, so that we have better testing coverage.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
martin-velay: Actually when `hmac_done` I am then checking that `idle` is reached in the SCB. Thanks for the remark anyway.

hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
```diff
@@ -74,8 +78,8 @@ class hmac_smoke_vseq extends hmac_base_vseq;
       `DV_CHECK_RANDOMIZE_FATAL(this)
       `uvm_info(`gfn, $sformatf("starting seq %0d/%0d, message size %0d bits, hmac=%0d, sha=%0d",
                 i, num_trans, msg.size()*8, hmac_en, sha_en), UVM_LOW)
-      `uvm_info(`gfn, $sformatf("digest size=%4b, key length=%6b",
-                digest_size, key_length), UVM_LOW)
+      `uvm_info(`gfn, $sformatf("digest size=%s, key length=%0d",
+                get_digest_size(digest_size), get_key_length(key_length)), UVM_LOW)
```
antmarzam, line 82: I think you may wanna increase the verbosity for the UVM infos here and below? Not sure if there's further reason to have such low verbosity. 
I'm personally inclined to have this kind of printouts at UVM_MEDIUM/HIGH/DEBUG.

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -20,30 +18,27 @@ class hmac_test_vectors_sha_vseq extends hmac_base_vseq;
...
+    //   0 :/ 2   // Illegal -> should get casted to Key_None in HW
+    // };
+  }
+
+  constraint digest_size_c {
+    digest_size == 'h1;  // test only with SHA2-256
+    // $countones(digest_size) == 1 dist {
+    //   1 :/ 8,  // SHA2_256/SHA2_384/SHA2_512/SHA2_None
+    //   0 :/ 2   // Illegal -> should get casted to SHA2_None in HW
+    // };
```
antmarzam, line 36: Is the commented code intentional?

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -44,33 +56,18 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
...
       1 :/ 10,  // Key_128/Key_256/Key_384/Key_512/Key_1024/Key_None
       0 :/ 0    // Illegal -> should get casted to Key_None in HW
     };
+  }
 
+  constraint digest_size_c {
     $countones(digest_size) == 1 dist {
       1 :/ 10,  // SHA2_256/SHA2_384/SHA2_512/SHA2_None
       0 :/ 0    // Illegal -> should get casted to SHA2_None in HW
     };
```
antmarzam, line 70: I'm not sure I understand the constraint, can you explain what it does?
I've always seen dist used as:
```
constraint c{
variable_name dist {
value_a :/ weight // or := instead
};
}
```

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -10,14 +10,15 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
 
   bit             sha_en, fifo_empty, hmac_idle;
   bit [7:0]       msg_q[$];
-  bit             hmac_start, hmac_process;
+  bit [7:0]       msg_part[$];  // Queue containing partial piece of the message if HASH interrupted
```
antmarzam, line 13: I'd suggest to rename to `msg_part_q[$]` just so the name itself indicates this is a SV queue

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -110,14 +121,17 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
                 end
               end else if (item.a_data[HashStart]) begin
                 {hmac_process, hmac_start} = item.a_data[1:0];
+                // Clear this variable when context has to be skipped as HASH continue won't be set
+                if (cfg.sar_skip_ctxt) hmac_stopped = 0;
```
antmarzam, line 125: Nit: In order to help with live-debugging is better practice not to use this if structure. It blocks you to distinguish if you'd like to set a breakpoint on the if, or in the variable assignment itself because both statements are on the same line. 
Definitely not a deal breaker, but if you wanted to set a breakpoint in `hmac_stopped = 0` you'd need to go and hack the source code...

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -467,7 +548,7 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
     fork
       begin : process_hmac_key_pad
         forever begin
-          wait(!cfg.under_reset);
+          if (cfg.under_reset) @(negedge cfg.under_reset);
```
antmarzam, line 551: I'd be inclined to go back towards using `wait` which is a more DV like statement in a scoreboard. 
Perhaps something like:
```
if (cfg.under_reset)
  wait (!cfg.under_reset)
```

hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
martin-velay: You're probably right, I think that's a general comment for this IP. I'll create an issue for later.

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
martin-velay: There is a comment just above with a TODO
```
  // TODO (#22932): remove this constraint and use the constraints from hmac_base_vseq once
  // this test is adapted for all digest sizes and key lengths
```

gdessouky: A PR coming out now will be removing that :-)

rswarbrick (commented):
Thanks for how you've split this into small commits. I think that some of the commit messages might want tweaking though: I don't believe that the first commit's message is really correct :-)

image

rswarbrick:
As a separate note, I'd also suggest splitting up the PR if possible. The 22 commits that are currently in the PR won't be "atomic" after merge (we don't distinguish in the git history between "internal" and "head of PR" commits), so it probably makes sense to split things out if possible.

That will allow the review to go much more in parallel.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
martin-velay: Here we have one-hot encoded values (see [here](https://opentitan.org/book/hw/ip/hmac/doc/registers.html#cfg--digest_size)). That's why I am forcing the randomization to have only one '1' with the `$countones`. At it is now, as we have issues with a test we are not testing any illegal values, that's why its weight is null, maybe that's your where your confusion comes from. We should have :
```
    $countones(key_length) == 1 dist {
      1 :/ 10,  // Key_128/Key_256/Key_384/Key_512/Key_1024/Key_None
      0 :/ 1    // Illegal -> should get casted to Key_None in HW
    };
```

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: You're right, thanks

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Yes, BTW it's very hard to live debugging with the compare macro we have !
I'll change it here and in some other places

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -487,10 +568,13 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
-                // break if hmac is done or if invalid config error is triggered
+                // break if hmac is done or if invalid config error is triggered or if SHA is
+                // being disabled as HMAC enable configuration could be changed and thus
+                // key_process_cycles might need to be updated
                 if (ral.intr_state.hmac_done.get_mirrored_value() == 1 ||
                    (ral.intr_state.hmac_err.get_mirrored_value()  == 1 &&
-                    ral.err_code.get_mirrored_value() == SwInvalidConfig)) break;
+                    ral.err_code.get_mirrored_value() == SwInvalidConfig) ||
+                    !sha_en) break;
                 cfg.clk_rst_vif.wait_clks(1);
```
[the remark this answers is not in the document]
martin-velay: Yes, this is reinforced by the nice rule of 100 characters per line 🤯
I have changed it, and I also adopted the `gmv` macro

martin-velay:
@antmarzam thanks for you feedback. I have done most of the suggested changes

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
antmarzam: Sorry - I had missed your response.
I still don't understand what the constraint does, or at least the order of the resolution, are you sure randomization is doing what you want? 
I guess I'd be expecting a variable to be controlled via `dist`, but not the result of a `$countones` constraint

If you just want `key_lenght/digest` to have only 1-bit set, why not just doing:
```
constraint one_hot_c {
$countones(key_length) == 1;
}
```

Perhaps I'm not understanding the usecase fully, but if you wanted to also allow illegal values by allowing all zeroes, can't you just do:
```
constraint one_hot_and_illegal_c {
solve illegal_allowed before key_length; //Or perhaps set the `illegal_allowed` variable in a test by test basis?
...
// Then you can randomize `illegal_length` to have whatever chances it makes sense
..
//and depending on that flag you either enforce one-hot, or not:
illegal_allowed -> key_length == 0;
!illegal_allowed -> $scountones(key_length) == 1;
}

```

martin-velay: After a talk with Antonio, we feel better to add a comment to explain this constraint. But it has been tested and it's working as expected.

gdessouky: Just to help clarify a bit, illegal values are all-zero values as well as values that are not one-hot, so all values where countones(x) != 1 and we want to be able to test both, but illegal values with a much lower probability.

hw/dv/tools/dvsim/xcelium.hjson
```diff
@@ -199,8 +199,8 @@
     {
       name: xcelium_gui
       is_sim_mode: 1
-      build_opts: ["-createdebugdb", "-access +c"]
-      run_opts: ["-gui"]
+      build_opts: ["-createdebugdb", "-access +c", "-uvmlinedebug", "-linedebug"]
+      run_opts: ["-gui", "-input '@database -open waves -into waves.shm -default;probe -create tb -depth all -packed 32k -unpacked 32k -all -dynamic -database waves'"]
```
rswarbrick: This change is definitely not something that's needed. You undo it again in a later commit in the PR. Please squash the two for this file. (Easiest approach is probably to do an interactive rebase and pause after this commit. Then `git checkout my-work-branch hw/dv/tool/dvsim/xcelium.hjson` and `git commit --amend` and `git rebase --continue`).

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -115,6 +115,9 @@ package hmac_env_pkg;
   typedef class hmac_env_cov;
   typedef cip_base_virtual_sequencer #(hmac_env_cfg, hmac_env_cov) hmac_virtual_sequencer;
   typedef virtual pins_if #(1) d2h_a_ready_vif;
+  // typedef uvm_event uvm_event_sar;
+  // typedef uvm_event #(int) uvm_event_sar;
```
rswarbrick: This commented code (and the `get_key_csr_name` commented function below) gets removed by your later commit "[hmac/dv] Test vectors SHA and cleans". Again, please squash the change and avoid this being in the history at all.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
rswarbrick: Your change that renames this variable is currently in the PR as the commit "[hmac,dv] Break if conditions to ease debug". Please tidy this up: if you prefer the new name, please rewrite history to avoid committing the original name at all.

rswarbrick (changes requested):
I've got a small way through the first commit of this enormous PR.

Please could do some tidying up in the git history (as suggested in the comments I'm leaving here). More importantly, please could you try to split up the PR a bit as I suggested [last week](https://github.com/lowRISC/opentitan/pull/23116#issuecomment-2154619275)? This behemoth is not going to get merged today as a single blob, but we might manage to merge some of the changes if a more incremental approach is possible.

rswarbrick:
Thanks for the response. If you really believe that this can only be reviewed as a single block (+789/-306) then you should probably squash all the commits together.

I'd be quite surprised if that was true. Maybe it's worth trying to pull a few commits out to the front of the PR and rebase the rest on top of them. It will probably make things a bit easier to reason about for you, as well as a potential reviewer.

hw/ip/hmac/dv/env/hmac_env_cfg.sv
```diff
@@ -19,6 +19,9 @@ class hmac_env_cfg extends cip_base_env_cfg #(.RAL_T(hmac_reg_block));
   // continue hashing.
   int save_and_restore_pct;
 
+  // Flag to notify the scoreboard to skip the current message writes and don't flush its variables
+  bit sar_skip_ctxt = 0;
```
antmarzam, line 23: I see how it's convenient to use `cfg.sar_skip_ctxt` from VSEQs. However, that seems hacky. 
Is there a way for the monitor itself to distinguish different message writes? 

To me it seems like this should be a `uvm_sequence_item` field that gets set in the monitor. Then the SCB would ditch the transactions which have the flag set

martin-velay:
Thanks a lot for your help to clean this up @hcallahan-lowrisc

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -177,6 +179,40 @@ package hmac_env_pkg;
...
+  endfunction : get_digest_size
+
+  // Return block size in bits according to the digest size
+  function automatic int get_block_size(int digest_size_reg);
+    case (digest_size_reg)
+      'h01:    return HMAC_BLK_SIZE_SHA2_256*8;      // SHA-2 256: 512-bit block
+      'h02:    return HMAC_BLK_SIZE_SHA2_384_512*8;  // SHA-2 384: 1024-bit block
+      'h04:    return HMAC_BLK_SIZE_SHA2_384_512*8;  // SHA-2 512: 1024-bit block
+      'h08:    return 0;                             // Unsupported/invalid and all-zero values
+      default: return 0;
```
rswarbrick, line 212: Is this the behaviour you want for invalid values? I just grepped for the function and found lines like this:
```systemverilog
if ((bits_written % get_block_size(digest_size) == 0) && (msg_q.size() > 0)) begin
```
(from `hmac_base_vseq`).

I think this will currently give a divide by zero behaviour if `digest_size` happens to be something other than 1, 2 or 4. Would it make sense to explode in `get_block_size`? Or maybe we need to explicitly calculate the block size and check it's nonzero before using it with `%` in the vseq?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
rswarbrick: I think that you've dropped the variable rename. Maybe worth changing it eventually, but I don't think it should block merging this PR. Thanks for taking a look at this.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -42,7 +43,10 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
     bit     do_cycle_accurate_check = 1'b1;
     bit     write                   = item.is_write();
     string  csr_name;
-
+    int     block_size;
```
rswarbrick, line 46: I'd suggest moving some of these variables to smaller scopes. I think that all four of these variables can move to the `if` block where they are used.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -133,6 +137,38 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
                 update_err_intr_code(SwHashStartWhenActive);
               end
             end
+            // Detect when stop has been triggered to store the current state, for Save and Restore
+            // check
+            if (item.a_data[HashStop]) begin
+              hmac_stopped = 1;
+
+              // Retrieve block size (in bits) according to configured digest size
+              block_size = get_block_size(`gmv(ral.cfg.digest_size));
```
rswarbrick, line 156: Probably worth checking that this is nonzero here to avoid dividing by zero on lines 152/153?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -133,6 +137,38 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+
+              // Round up message counter to the block boundary to predict message being processed
+              // so far by the DUT.
+              // hmac_rd_cnt is multiplied by 32 to express it in bits.
+              // block_size is divided by 8 to express it in bytes.
+              mult  = (hmac_rd_cnt * 32) / block_size;
+              mod   = (hmac_rd_cnt * 32) % block_size;
+              msg_slice_length = mod > 0 ? (mult + 1) * block_size / 8 : mult * block_size / 8;
```
rswarbrick, line 164: I think you might be able to do this by rounding with something like
```systemverilog
              mult = (hmac_rd_cnt * 32 + block_size - 1) / block_size;
              msg_slice_length = mult * block_size / 8;
```
as it's shorter and maybe easier to understand.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -133,6 +137,38 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+
+              // In case the remaining message is less than a block size, push only this remaining
+              // message into the queue. Divide by 8 as message queue is taking data in bytes.
```
rswarbrick, line 167: I think this comment might need a tweak. The division by 8 has happened a couple of lines earlier, I think?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -133,6 +137,38 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+              msg_slice_length = msg_slice_length > msg_q.size() ? msg_q.size() : msg_slice_length;
```
rswarbrick, line 168: This is fine, but I think an if block might be more obvious:
```systemverilog
              if (msg_slice_length > msg_q.size()) begin
                msg_slice_length = msg_q.size();
              end
```
(it's a pity that SystemVerilog doesn't have a `min` operator!)

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -291,23 +330,31 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
-          if (!ral.cfg.digest_swap.get_mirrored_value()) begin
-            // digest_swap is set to true to match spec vectors/expected digest word (big-endian)
-            // when digest swap is not true, then we need to swap the digest word
-            // to be able to compare it with the expected digest word (big-endian)
-            real_digest_val = {<<8{item.d_data}};
-          end else begin
-            real_digest_val = item.d_data;
-          end
+          // Predict intermediate digest values when S&R has been triggered
+          // TODO later, modify the C model to be able to provide intermediate digest (issue #23240)
```
rswarbrick, line 374: This looks really sensible. I think people have standardised on the syntax "// TODO(#23240): ..." and I wouldn't be surprised if we've got some helper tools that extract stuff in that format. Not a big deal, but I see that you've got an issue number already!

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -65,22 +65,28 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
-              // endian_swap bit 1 operates at big endian data
-              msg = (ral.cfg.endian_swap.get_mirrored_value()) ? {msg, bytes[i]} : {bytes[i], msg};
+        // Only push message into the FIFO when intended, as in case of S&R triggered with another
+        // context, we don't want to process the message of this other context.
+        if (!cfg.sar_skip_ctxt) begin
+          if (!sha_en) begin
+            update_err_intr_code(SwPushMsgWhenDisallowed);
+          end else if (!hmac_start && !hmac_continue) begin
+            update_err_intr_code(SwPushMsgWhenDisallowed);
+          end else if ((hmac_start || hmac_continue) && !cfg.under_reset) begin
```
rswarbrick, line 75: I'd be tempted to fold the first two conditions together to give something like
```systemverilog
          if ((!sha_en) || (!hmac_start && !hmac_continue)) begin
            update_err_intr_code(SwPushMsgWhenDisallowed);
          end else if ((hmac_start || hmac_continue) && !cfg.under_reset) begin
```
but then I notice that we're repeating a check, so I think it could actually be:
```systemverilog
          if ((!sha_en) || (!hmac_start && !hmac_continue)) begin
            update_err_intr_code(SwPushMsgWhenDisallowed);
          end else if (!cfg.under_reset) begin
```

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -65,22 +65,28 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+            bit [7:0] bytes[4];
+            bit [7:0] msg[$];
+            // Register coverage for the fact that a write occurs to the message FIFO during
+            // a compression round
+            if (cfg.en_cov) cov.wr_msg_during_hash_cg.sample(1);
+            {<<byte{bytes}} = item.a_data;
```
rswarbrick, line 81: This is clever! Is it possible to move the streaming operator so that it applies to the rvalue instead? It might be easier to understand. (For me, at least...)

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -65,22 +65,28 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+            // TODO (#): could add here cov sampling of nb_bytes valid per word and its transitions
+            // do endian swap in the word according to the mask, then push to the msg queue
+            foreach (item.a_mask[i]) begin
+              if (item.a_mask[i]) begin
+                // endian_swap bit 1 operates at big endian data
+                msg = (`gmv(ral.cfg.endian_swap)) ?
+                      {msg, bytes[i]} : {bytes[i], msg};
```
rswarbrick, line 88: Would it be easier to do this reversal with a streaming operation after the `foreach`? It might be a bit simpler to write?

rswarbrick (commented):
Some comments on the first 3 commits in the PR. Sorry that I haven't made it to the end yet.

hw/ip/hmac/dv/env/hmac_env_cfg.sv
martin-velay: What you are saying make sense, I don't deny. The monitor could probably do it, and transmit this info to the SB via the `seq_item`, but this stuff is temporary as I have an open issue #23562 to make the SB behaving the same way for all the messages.
I can add a comment with this issue tag along this bit.

hw/ip/hmac/dv/env/hmac_env_pkg.sv
martin-velay: Good catch, I have added a condition to check if the function is returning a value different from 0 in the `hmac_base_vseq` and also in `hmac_scoreboard`

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Done

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Done (see https://github.com/lowRISC/opentitan/pull/23116#discussion_r1642409815)

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Indeed that's simpler. Thanks

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: True!

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: That's maybe something we could implement as an additional macro in our DV pkg. What do you think?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Oh OK, I had no idea, I'll change it there and on the other locations in the HMAC DV code as well

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: This change has been moved into another PR https://github.com/lowRISC/opentitan/pull/23685.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Agree, that's better

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: That's not coming from me, I have checked the history it's there since the beginning of this file.
Should I still edit it ?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Same here, I haven't changed anything. That's here here since ages. Should I edit?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
rswarbrick: Maybe a good idea: I'm not sure. I know that `#define MIN(a,b) ...` is very common in C codebases.

martin-velay: Let's keep it in mind if this come up again

martin-velay:
Yesterday I have pushed after having taken into account all the comments above. I ran a regression over the night, all is good. And all checks have passed. I think this PR is ready to be merged. It requires approvals now.

Any other comments?
@gdessouky @rswarbrick @ballifatih @antmarzam @hcallahan-lowrisc

hw/ip/hmac/dv/env/hmac_scoreboard.sv
rswarbrick: Ah, sorry: I didn't realise that this wasn't new. Definitely no need to hack on it! Sorry for the noise.

rswarbrick (approved):
Thanks for all the tidy-ups.

PR #23120, opened 2024-05-14
[edn/dv] Add transition to reject entropy state to alert test

h-filali:
This PR is related to [#22352](https://github.com/lowRISC/opentitan/issues/22352)

I wasn't able to confirm with the coverage yet if this fills the FSM coverage hole since I was working on debugging.
Through this PR I was able to find some potential minor issues.
Some might not even be issues but are still worthy of discussion.

These include:
- HW_CMD_STS only shows boot/auto mode after the INS command (For auto mode this probably makes sense since INS is SW controlled).
- After reenabling EDN the HW_CMD_STS still shows the last command before EDN was disabled.
- Two assertoffs I added need to be checked if they can be added or if there is a bug.

Furthermore, the alert test does not have a check for SW_MODE since the probability of having both auto mode and boot mode disabled in the CTRL reg is 0 for this test. This hole needs to be filled.

### DIFF
--- a/hw/ip/edn/dv/env/seq_lib/edn_alert_vseq.sv
+++ b/hw/ip/edn/dv/env/seq_lib/edn_alert_vseq.sv
@@ -30,14 +30,13 @@ class edn_alert_vseq extends edn_base_vseq;
     join_none
   endtask
 
-  task test_edn_cs_sts_alert();
+  task test_edn_cs_hw_sts_alert();
     string state_path;
-    string genbits_valid_path;
     state_e exp_state;
-    bit [31:0] exp_hw_cmd_sts;
+    bit [31:0] exp_cmd_sts;
     bit send_generate;
-    state_path = cfg.edn_vif.sm_err_path("edn_main_sm");
-    genbits_valid_path = cfg.edn_vif.genbits_valid_path();
+    bit state_reached_during_cs_handshake;
+    state_path = cfg.edn_vif.sm_err_path("edn_main_sm_next");
     // Re-randomize config without invalid mubi values and without clearing the FIFOs.
     cfg.use_invalid_mubi  = 0;
     cfg.cmd_fifo_rst_pct  = 0;
@@ -55,14 +54,15 @@ class edn_alert_vseq extends edn_base_vseq;
                                          exp_state inside {BootLoadIns, BootInsAckWait,
                                                            BootLoadGen, BootGenAckWait,
                                                            BootPulse, BootDone,
-                                                           BootLoadUni};)
+                                                           BootLoadUni, BootUniAckWait};)
       // Set the boot generate command to request the minimal amount of entropy.
       wr_cmd(.cmd_type(edn_env_pkg::BootGen), .acmd(csrng_pkg::GEN), .clen(0), .flags(MuBi4False),
              .glen(1), .mode(edn_env_pkg::BootReqMode));
     end else if (cfg.auto_req_mode == MuBi4True) begin
       // If instead EDN is configured in auto mode, randomly select one of the auto mode states.
       `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(exp_state,
-                                         exp_state inside {AutoAckWait, AutoDispatch,
+                                         exp_state inside {AutoLoadIns, AutoFirstAckWait,
+                                                           AutoAckWait, AutoDispatch,
                                                            AutoCaptGenCnt, AutoSendGenCmd,
                                                            AutoCaptReseedCnt, AutoSendReseedCmd};)
       // In auto mode, minimize number of requests between reseeds and set the reseed command.
@@ -90,6 +90,14 @@ class edn_alert_vseq extends edn_base_vseq;
         `uvm_info(`gfn, $sformatf("State reached"), UVM_HIGH)
         // The next acknowledgement should return an error status.
         cfg.m_csrng_agent_cfg.rsp_sts_err = cfg.which_cmd_sts_err;
+        cfg.m_csrng_agent_cfg.cmd_zero_delays = 1;
+        cfg.m_csrng_agent_cfg.cmd_force_ack = 1;
+        cfg.clk_rst_vif.wait_n_clks(2);
+        // If the state was reached during the handshake,
+        // the hw_cmd_type in hw_cmd_sts should not have been recorded yet.
+        state_reached_during_cs_handshake =
+            cfg.m_csrng_agent_cfg.vif.mon_cb.cmd_req.csrng_req_valid &&
+            cfg.m_csrng_agent_cfg.vif.mon_cb.cmd_rsp.csrng_req_ready;
       )
     join_none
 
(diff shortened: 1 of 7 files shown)


h-filali:
I fixed all of the above issues.
I removed the assertions. Out of a full regression there is still one failing test left that fails due to the assertion which should be checked.
I added some RTL changes in a second commit.
I rebased and merged a conflict in `edn_core.sv` and `edn_main_sm.sv`.
There are probably still some errors that will be detected by the CI which need to be fixed.
@vogelpi please have a look at the two latest pushes (On the second one just the files I resolved a conflict for).

hw/ip/edn/dv/env/seq_lib/edn_alert_vseq.sv
```diff
@@ -121,39 +127,100 @@ class edn_alert_vseq extends edn_base_vseq;
...
-    exp_hw_cmd_sts[hw_cmd_type+:CMD_TYPE_SIZE] =
-        (exp_state inside {BootLoadIns, BootInsAckWait}) ? csrng_pkg::INS :
-        (exp_state inside {BootPulse, BootDone, BootLoadUni}) ? csrng_pkg::UNI :
-        (exp_state inside {AutoCaptReseedCnt, AutoSendReseedCmd}) ? csrng_pkg::RES :
-        csrng_pkg::GEN;
-    csr_rd_check(.ptr(ral.hw_cmd_sts), .compare_value(exp_hw_cmd_sts));
+    // Check the hw_cmd_sts.
+    exp_cmd_sts[hw_cmd_boot_mode] =
+        (exp_state == BootLoadIns)       ? 1'b0 :
+        (exp_state == BootInsAckWait) && !state_reached_during_cs_handshake ? 1'b0 :
```
vogelpi, line 156: Can you please document why `state_reached_during_cs_handshake` is required?

hw/ip/edn/dv/env/seq_lib/edn_alert_vseq.sv
```diff
@@ -90,6 +90,12 @@ class edn_alert_vseq extends edn_base_vseq;
         `uvm_info(`gfn, $sformatf("State reached"), UVM_HIGH)
         // The next acknowledgement should return an error status.
         cfg.m_csrng_agent_cfg.rsp_sts_err = cfg.which_cmd_sts_err;
+        cfg.m_csrng_agent_cfg.cmd_zero_delays = 1;
+        cfg.m_csrng_agent_cfg.cmd_force_ack = 1;
+        cfg.clk_rst_vif.wait_n_clks(2);
+        state_reached_during_cs_handshake =
+            cfg.m_csrng_agent_cfg.vif.mon_cb.cmd_req.csrng_req_valid &&
+            cfg.m_csrng_agent_cfg.vif.mon_cb.cmd_rsp.csrng_req_ready;
```
vogelpi, line 98: Are there better ways to infer these signals than looking at the monitor? Maybe a question to discuss with @martin-velay ?

vogelpi (commented):
Thanks for the PR @h-filali . I've added some comments.

hw/ip/edn/dv/env/seq_lib/edn_alert_vseq.sv
martin-velay: I don't have the feeling that's bad, at least not more than than performing checks from the sequence itself 🤯

hw/ip/edn/rtl/edn_core.sv
```diff
@@ -547,11 +549,11 @@ module edn_core import edn_pkg::*;
   assign hw2reg.sw_cmd_sts.cmd_rdy.d = cmd_rdy;
   assign cmd_rdy = !sw_cmd_req_load && cmd_rdy_d && cmd_reg_rdy_d;
   // We accept SW commands only in SW or auto mode.
-  // In auto mode, sw_cmd_valid will transition to low after the initial instantiate command.
+  // In auto mode, sw_cmd_mode will transition to low after the initial instantiate command.
   // In SW mode, cmd_rdy is low when a previous command has not been acked yet.
   assign cmd_rdy_d =
          !edn_enable_fo[SwCmdSts] ? 1'b0 :
-         !sw_cmd_valid ? 1'b0 :
+         !sw_cmd_mode ? 1'b0 :
```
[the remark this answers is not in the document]
h-filali: sure

hw/ip/edn/dv/env/seq_lib/edn_alert_vseq.sv
h-filali: Will do!

hw/ip/edn/dv/env/seq_lib/edn_alert_vseq.sv
h-filali: Thanks @martin-velay !

hw/ip/edn/rtl/edn_core.sv
```diff
@@ -582,51 +586,53 @@ module edn_core import edn_pkg::*;
...
   // Set the cmd_ack signal to high whenever a hardware command is acknowledged and set it
   // to low whenever a new hardware command is issued to the CSRNG.
   assign hw2reg.hw_cmd_sts.cmd_ack.de = 1'b1;
   assign hw2reg.hw_cmd_sts.cmd_ack.d = csrng_hw_cmd_ack_d;
   assign csrng_hw_cmd_ack_d =
-         !edn_enable_fo[HwCmdSts] ? 1'b0 :
-         sw_cmd_valid ? csrng_hw_cmd_ack_q :
-         (cs_cmd_req_vld_out_q && csrng_cmd_i.csrng_req_ready) ? 1'b0 :
-         csrng_cmd_i.csrng_rsp_ack ? 1'b1 :
+         main_sm_idle ? 1'b0 :
```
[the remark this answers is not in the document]
vogelpi: Thanks for changing this. This looks better now.

hw/ip/edn/rtl/edn_core.sv
```diff
@@ -582,51 +586,53 @@ module edn_core import edn_pkg::*;
...
+         cs_hw_cmd_handshake ? 1'b0 :
          csrng_hw_cmd_ack_q;
   // Set the cmd_type to the application command type value of the hardware controlled
   // command issued last.
   assign hw2reg.hw_cmd_sts.cmd_type.de = 1'b1;
   assign hw2reg.hw_cmd_sts.cmd_type.d = cmd_type_d;
-  assign cmd_type_d = (edn_enable_fo[HwCmdSts] && !sw_cmd_valid && cs_cmd_req_vld_out_q
-                       && csrng_cmd_i.csrng_req_ready) ? cs_cmd_req_out_q[3:0] : cmd_type_q;
+  assign cmd_type_d =
+      main_sm_idle ? {1'b0, csrng_pkg::INV} :
```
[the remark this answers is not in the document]
vogelpi: I really think we don't want to clear this field upon entering idle. Because this means we can't see which command was executed last before leaving auto mode. I've now changed the design / this PR accordingly.

vogelpi:
Update: I've minimally modified the RTL in this PR (see https://github.com/lowRISC/opentitan/pull/23120#discussion_r1626731901) and rebased the PR on top of master (second force push).

vogelpi (approved):
Thanks for the PR and for implementing the feedback. One comment wasn't addressed and I've now implemented that change.

PR #23122, opened 2024-05-14
[topgen] Improve support for multiple address spaces

Razer6:
This PR uses a different approach than #22961 to support multiple address spaces better and generate their C and Rust collateral. 

In this PR, `topgen` iterates over all address spaces available and individually calls `TopGenC` or `TopGenRust` on it. Consequently, all address spaces are treated separately, and all outputs are generated in separate files. Therefore, the address space becomes part of the file name as a suffix after the top name. One address space is a default address space (annotated in the HJSON, e.g., the `hart`) that doesn't face this change in file name.  That means the currently existing file names stay the same (without suffixes) and describe the `hart` address space.

The C and Rust headers generate the definition for alerts and interrupts. Right now, these definitions will only be generated for the default address space.  This will be improved in the future since multiple address spaces with multiple CPUs might have their own PLIC. Consequently, the interrupt definitions for different address spaces or interrupt domains will be generated in the corresponding file.


hw/top_darjeeling/sw/autogen/top_darjeeling_soc_mbx.h
```diff
@@ -0,0 +1,211 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef OPENTITAN_SOC_MBX_HW_TOP_DARJEELING_SW_AUTOGEN_TOP_DARJEELING_H_
+#define OPENTITAN_SOC_MBX_HW_TOP_DARJEELING_SW_AUTOGEN_TOP_DARJEELING_H_
+
+#include "top_darjeeling_flash_ctrl_dummy.h"
```
Razer6, line 8: That might be something to remove from the generation.

Razer6: And the order of the top and address apce might be changed in the include guard.

util/topgen/c_test.py
```diff
@@ -59,13 +59,15 @@ def __init__(self, name: str, inst_name: str, base_addr_name: str,
...
-        super().__init__(top_info, name_to_block)
+    def __init__(self, top_info, name_to_block: Dict[str, IpBlock], addr_space: str):
+        super().__init__(top_info, name_to_block, addr_space)
 
         self.irq_peripherals = self._get_irq_peripherals()
         self.alert_peripherals = self._get_alert_peripherals()
 
     def _get_irq_peripherals(self):
+        if self.addr_space != self.default_addr_space:
+            return []
```
a-will, line 70: I'm not totally sure whether this restriction would generally hold, but it doesn't hurt to start with it.

a-will (approved):
Generally looks good to me. I fixed up some stylistic issues to get it to pass lint, but you might want to have a more strict Python reviewer than me.

a-will:
Looks like englishbreakfast may need similar attention :)

util/topgen/c_test.py
Razer6: No, in the future, I would like to determine if there is a PLIC in the address space and then yield the devices that are connected to that one - keyword `interrupt_domain`.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.c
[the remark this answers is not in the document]
Razer6: Yeah, these address spaces only have definitions in the header. Not sure if we can determine if there is nothing to generate. Probably the length of all internal structures is 0?

Razer6:
CI is happy now :) @andreaskurth Who be eligible to review that PR form a python perspective?

andreaskurth:
> @andreaskurth Who be eligible to review that PR form a python perspective?

I think getting @hcallahan-lowrisc's feedback would be valuable

util/topgen.py
```diff
@@ -776,30 +776,38 @@ def render_template(template_path: str, rendered_path: Path, **other_info):
...
+        addr_space_suffix = ""
+        if not default_addr_space:
+            addr_space_suffix = "_" + addr_space['name']
 
-    # Creating Rust output directory
-    rsformat_dir = out_path / "sw/autogen/chip/"
-    rsformat_dir.mkdir(parents=True, exist_ok=True)
+        # The Rust file needs some complex information, so we initialize this
+        # object to store it.
+        rs_helper = TopGenRust(completecfg, name_to_block, version_stamp, addr_space['name'])
```
rswarbrick, line 788: I know that the rest of this file has the same problem, but maybe it makes sense to do some sanity checking of `addr_space['name']` here? (Make sure it's a string, maybe? And how does `default_addr_space` relate to the allowed values here? (Or maybe that sort of check belongs somewhere earlier, rather than being Rust-specific?)

util/topgen.py
```diff
@@ -1286,7 +1294,7 @@ def render_template(template_path: str, rendered_path: Path,
                 fout.write(template_contents)
 
         # Header for SV files
-        gencmd = warnhdr + """//
+        gencmd_sv = warnhdr + """//
```
rswarbrick, line 1297: I really approve of name changes like this! :-)

util/topgen.py
```diff
@@ -1297,83 +1305,102 @@ def render_template(template_path: str, rendered_path: Path,
...
-                            memory_cheader_path,
-                            helper=c_helper)
+                        gencmd=gencmd_sv)
+
+        for addr_space in topcfg['addr_spaces']:
+            default_addr_space = addr_space.get('default', False)
+
+            addr_space_suffix = ""
+            if not default_addr_space:
+                addr_space_suffix = "_" + addr_space['name']
```
rswarbrick, line 1330: Given that this appears in several places, maybe it makes sense to pull it out into a function?

rswarbrick (commented):
Some nitty comments, but this looks really good to me.

util/topgen.py
Razer6: I agree, it might be checked that `addr_space['name']` is a string. There is no custom validation for these kind of dicts, right? `default_addr_space` is determined on demand from the list of address spaces, the one that specifies `default = true`.

util/topgen.py
Razer6: Agree.

util/topgen.py
rswarbrick: I don't think there is (not code I wrote...). And good point about `default_addr_space`: I totally misread your code!

sw/host/opentitanlib/src/chip/autogen/mod.rs
```diff
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// The "english breakfast" variant of the chip doesn't have the same
```
Razer6, line 5: That code has been manually edited and now got overwritten my the template rendering.

Razer6:
Closing since https://github.com/lowRISC/opentitan/pull/25894 is merged

PR #23126, opened 2024-05-15
[rv_dm,dv] Revert "double request" logic in DMI driver and monitor

rswarbrick:
I strongly suspect that I misunderstood what was going on when I added this code, and it just makes things more complicated. The failures that I was "working around" were probably due to bugs in the jtag driver, which I believe I have now fixed.


hcallahan-lowrisc (approved):
I don't fully understand all the in's and out's of this, but since the checks look good then I think this should be good to go.

PR #23127, opened 2024-05-15
[rom_ext] Communicate minimum security version in the boot_log

cfrantz:
1. Add `min_sec_ver` fields to the `boot_log` structure.
2. Initialize the minimum security versions to the values contained in the `boot_data` structure.

### DIFF
--- a/sw/device/silicon_creator/lib/boot_log.h
+++ b/sw/device/silicon_creator/lib/boot_log.h
@@ -45,8 +45,12 @@ typedef struct boot_log {
   uint32_t ownership_state;
   /** Number of ownership transfers this chip has had. */
   uint32_t ownership_transfers;
+  /** Minimum security version permitted for ROM_EXT payloads */
+  uint32_t rom_ext_min_sec_ver;
+  /** Minimum security version permitted for application payloads */
+  uint32_t bl0_min_sec_ver;
   /** Pad to 128 bytes. */
-  uint32_t reserved[12];
+  uint32_t reserved[10];
 } boot_log_t;
 
 OT_ASSERT_MEMBER_OFFSET(boot_log_t, digest, 0);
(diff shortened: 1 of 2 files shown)


jettr (approved):
We will be able to use the ROM_EXT version number to determine which new fields are populated in the future (it isn't really a big deal now, but wondering how we add fields in the future).

What does ROM_EXT set the `reserved` field to (or is it left uninitialized)?

cfrantz:
> We will be able to use the ROM_EXT version number to determine which new fields are populated in the future (it isn't really a big deal now, but wondering how we add fields in the future).

I will be incrementing the ROM_EXT version to 0.2 for the next release.  However, I consider the structure of the `boot_log` and `boot_svc` messages to be part of the silicon_creator retention RAM, and thus, are identified by the ret_ram version (which, I think is currently `RR04`).

> 
> What does ROM_EXT set the `reserved` field to (or is it left uninitialized)?

The `reserved` fields of the boot_log are explicitly initialized to zero.

While we're in development mode, I don't really want to be incrementing the retram version for trivial non-breaking changes, so I'm inclined to leave it at `RR04` until we make a more major change to the structure of retram.

jettr:
> I will be incrementing the ROM_EXT version to 0.2 for the next release. However, I consider the structure of the `boot_log` and `boot_svc` messages to be part of the silicon_creator retention RAM, and thus, are identified by the ret_ram version (which, I think is currently `RR04`).

That sounds good to me. Where am I able to look up the ret_ram version?

> While we're in development mode, I don't really want to be incrementing the retram version for trivial non-breaking changes, so I'm inclined to leave it at `RR04` until we make a more major change to the structure of retram.

Agreed. I just wanted to see what the plan was long-term :)

jettr:
> > That sounds good to me. Where am I able to look up the ret_ram version?
> 
> It should be the first word of the ret_ram region (`0x4060_0000`).

My bad, I was thinking about old code. Thanks!

timothytrippel (approved):

PR #23128, opened 2024-05-15
[usbdev,sival] Fix config_host and aon_pullup on silicon

pamaury:
These tests were missing a harness to enable vbus sensing which caused the test to fail.


jwnrt (approved):
LGTM

PR #23130, opened 2024-05-15
[dv/otp_ctrl] dont compare direct_access_rdata_* on on non-integrity partitions

dkabely-ot:
For non-integrity partitions (s.a. VENDOR_TEST), the OTP scoreboard backdoor reads the OTP in raw form (doesn't apply ECC corrections) and compares with actual data returned on the DAI interface. However, in case `MacroEccUncorrErr` is injected, the IP might try and correct the errors, and the result on the DAI won't match the actual raw data in that particular address.

This issue causes two failures:

1. [Memory comparison](https://github.com/lowRISC/opentitan/blob/master/hw/ip/otp_ctrl/dv/env/otp_ctrl_scoreboard.sv#L612C1-L616C44). This has been mitigated in CS.

2. [Register comparison for `OTP_CTRL.DIRECT_ACCESS_RDATA_*`](https://github.com/lowRISC/opentitan/blob/master/hw/ip/otp_ctrl/dv/env/otp_ctrl_scoreboard.sv#L1119-L1130). This is a bit trickier to override in CS, hence this PR is created to address the issue. Just skip the register comparison in case a `MacroEccCorrErr` or `MacroEccUncorrErr` happens when we access one of the non-integrity partitions.

### DIFF
--- a/hw/ip/otp_ctrl/data/otp_ctrl_scoreboard.sv.tpl
+++ b/hw/ip/otp_ctrl/data/otp_ctrl_scoreboard.sv.tpl
@@ -807,6 +807,9 @@ class otp_ctrl_scoreboard #(type CFG_T = otp_ctrl_env_cfg)
                     predict_no_err(OtpDaiErrIdx);
                     predict_rdata(is_secret(dai_addr) || is_digest(dai_addr),
                                   read_out0, read_out1);
+                    // do not check direct_access_rdata_* on ECC errors in
+                    // non-integrity partitions
+                    check_dai_rd_data = 0;
                   end else begin
                     predict_no_err(OtpDaiErrIdx);
                     predict_rdata(is_secret(dai_addr) || is_digest(dai_addr),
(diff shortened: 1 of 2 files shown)


sha-ron (approved):

andreaskurth:
@matutem: Could you review this please?

vogelpi (approved):
Thanks for the PR and clear explanation @dkabely-ot . To me, this looks good.

andreaskurth:
Great, thanks for the review @vogelpi.

@dkabely-ot if you could address this failing CI check, we should then be able to merge this afterwards:
> Auto-generated OTP memory map not up-to-date. Regenerate with 'make -k -C hw otp-mmap'.

dkabely-ot:
Hi @andreaskurth 

Thanks!
It seems that this CI failure isn't related to the PR itself, however I rebased and re-pushed.

andreaskurth:
> Thanks! It seems that this CI failure isn't related to the PR itself, however I rebased and re-pushed to check if it helps.

Ah, you also need to update `hw/ip/otp_ctrl/data/otp_ctrl_scoreboard.sv.tpl`, otherwise the file that you changed will get overwritten during the generation of autogenerated files

I took the liberty to do that

dkabely-ot:
Hi @andreaskurth,



> > Thanks! It seems that this CI failure isn't related to the PR itself, however I rebased and re-pushed to check if it helps.
> 
> Ah, you also need to update `hw/ip/otp_ctrl/data/otp_ctrl_scoreboard.sv.tpl`, otherwise the file that you changed will get overwritten during the generation of autogenerated files
> 
> I took the liberty to do that

Thanks!

andreaskurth:
I don't see how this DV-only change could cause FPGA tests to fail.. merging

dkabely-ot:
Thanks!

PR #23133, opened 2024-05-15
[dv,usbdev] Fix regressions

alees24:
This PR just updates some of the existing block level DV sequences following these changes to the base sequence:
- use a randomized address by default for better testing/coverage
- wait for the USB link to become active so that the Bus Reset performed by the usb20_driver does not interfere with sequence behavior.

A small subset of the existing sequences has been failing in regressions, mostly in a seed-invariant manner, through not being updated following those changes.


jwnrt (approved):
Thanks for the fixes

rswarbrick (approved):

PR #23134, opened 2024-05-15
[rv_dm,dv] Add missing waits to rv_dm_jtag_dmi_debug_disabled_vseq

rswarbrick:
This fixes failures for the occasional seed in the rv_dm_jtag_dmi_debug_disabled test. Those failures also cause lots of failures in the rv_dm_stress_all test, which it also fixes.


marnovandermaas (approved):
Looks reasonable to me. Adding these waits can't have any adverse effects as far as I can tell.

Issue #23077, opened 2024-05-13
[i2c] V2(S) checklist items


hcallahan-lowrisc:
### Unimplemented Testpoints
- [x] target_fifo_watermark #23142
- [x] host_mode_config_perf #23543 
- [x] host_mode_clock_stretching #23543 
- [x] target_mode_tx_stretch_ctrl #23186
- [x] target_mode_nack_generation #23192 
- [x] host_mode_halt_on_nak #23940

### Regression
https://reports.opentitan.org/hw/ip/i2c/dv/2024.05.10_04.25.29/report.html 
- [x] https://github.com/lowRISC/opentitan/pull/23987
- A sampling of testpoints with low passrates. This is no-longer accurate for master, as it was observed on 10/05/24, but the same thresholds for V2 signoff still apply. The above PR should bring everything back up to the requisite level.
  - i2c_target_perf 0%
  - i2c_target_unexp_stop 8%
  - i2c_target_stress_all 12%

### Cov
- [ ] FSM 83.3%

### Open Issues (effort tracked against each issue)
Also see [#21755](https://github.com/lowRISC/opentitan/issues/21755) for previous discussion.
- [x] #23078
- [ ] ~~#17822~~ -> M7 
- [x] #23078 This issue was split out from the following:
  - [ ] ~~#21887~~ -> M7
- [x] #22347 

### Signoff Process (effort tracked against this issue)
#22108

PR #23142, opened 2024-05-15
[i2c,dv] I2C target mode fifo watermark tests

hcallahan-lowrisc:
This PR adds two new directed testcases for the I2C target-mode watermark interrupts added in #21621.
See the commit message + vseq header comment for more details about each test.

Some test machinery changes and cleanups have been done to support these new testcases, which comprises the first 4 commits. The final two commits add each testcase.

Best reviewed commit-by-commit.

Goes towards #23077

### DIFF
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
@@ -429,25 +429,21 @@ class i2c_base_vseq extends cip_base_vseq #(
     ral.timeout_ctrl.en.set(e_timeout);
     ral.timeout_ctrl.val.set(t_timeout);
     csr_update(.csr(ral.timeout_ctrl));
+    ral.host_fifo_config.rx_thresh.set(rx_thresh);
+    ral.host_fifo_config.fmt_thresh.set(fmt_thresh);
+    csr_update(ral.host_fifo_config);
+    ral.target_fifo_config.acq_thresh.set(acq_thresh);
+    ral.target_fifo_config.tx_thresh.set(tx_thresh);
+    csr_update(ral.target_fifo_config);
+
+    `uvm_info(`gfn, $sformatf("Wrote following register values :\n%s",
+      ral.sprint()), UVM_MEDIUM)
+
     // configure i2c_agent_cfg
     cfg.m_i2c_agent_cfg.timing_cfg = timing_cfg;
     `uvm_info(`gfn, $sformatf("\n  cfg.m_i2c_agent_cfg.timing_cfg\n%p",
         cfg.m_i2c_agent_cfg.timing_cfg), UVM_MEDIUM)
 
-    //*** program Host mode FIFO thresholds
-    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(rx_thresh)
-    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(fmt_thresh)
-    ral.host_fifo_config.rx_thresh.set(rx_thresh);
-    ral.host_fifo_config.fmt_thresh.set(fmt_thresh);
-    csr_update(ral.host_fifo_config);
-    //*** program Target mode FIFO thresholds
-    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(acq_thresh)
-    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(tx_thresh)
-    ral.target_fifo_config.acq_thresh.set(acq_thresh);
-    ral.target_fifo_config.tx_thresh.set(tx_thresh);
-    csr_update(ral.target_fifo_config);
-    //*** set FIFO_CTRL register
-    csr_update(ral.fifo_ctrl);
   endtask : program_registers
 
   virtual task program_format_flag(i2c_item item, string msg = "", bit do_print = 1'b0);
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
@@ -1277,4 +1277,28 @@ class i2c_base_vseq extends cip_base_vseq #(
     return (addr == target_addr0 || addr == target_addr1);
   endfunction
 
+  // Check an interrupt matches our expectation
+  // - Check that csr.intr_state[idx] == exp_val
+  // - Check the wired irq intr_vif.pins[idx] == exp_val
+  //
+  virtual task check_one_intr(i2c_intr_e intr, bit exp_val);
+    bit [TL_DW-1:0] obs_intr_state;
+    bit exp_pin;
+
+    csr_rd(.ptr(ral.intr_state), .value(obs_intr_state));
+    `DV_CHECK_EQ(obs_intr_state[intr], exp_val)
+    `DV_CHECK_EQ(cfg.intr_vif.pins[intr], exp_val)
+  endtask : check_one_intr
+
+  virtual task count_edge_intr(i2c_intr_e intr, ref uint poscnt, ref uint negcnt);
+    forever begin
+      @(cfg.intr_vif.pins[intr]);
+      case (cfg.intr_vif.pins[intr])
+        1'b0: negcnt++;
+        1'b1: poscnt++;
+        default:;
+      endcase
+    end
+  endtask: count_edge_intr
+
 endclass : i2c_base_vseq
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
@@ -942,57 +942,42 @@ class i2c_base_vseq extends cip_base_vseq #(
     return cnt;
   endfunction
 
-  // This task needs to set do_clear_all_interrupts = 0
+  virtual task proc_intr_acqthreshold; endtask
+  virtual task proc_intr_txthreshold; endtask
+  virtual task proc_intr_unexpstop; endtask
+  virtual task proc_intr_hosttimeout; endtask
+
+  // Target-mode interrupt handling routine.
+  //
+  // This routine loops through all wired-interrupts, and calls a handler routine
+  // for each one if active. It then does some checking based on 'expected_intr[]'.
+  //
   virtual task process_target_interrupts();
-    int delay;
-    int auto_ack_bytes;
     bit acq_fifo_empty;
     bit read_one = 0;
     while (!cfg.stop_intr_handler) begin
       @(posedge cfg.clk_rst_vif.clk);
-      if (cfg.intr_vif.pins[AcqStretch]) begin
-        bit [TL_DW-1:0] status;
-        bit ack_ctrl_stretch;
-        bit acq_full;
 
-        csr_rd(.ptr(ral.status), .value(status));
-        ack_ctrl_stretch = bit'(get_field_val(ral.status.ack_ctrl_stretch, status));
-        acq_full = bit'(get_field_val(ral.status.acqfull, status));
+      if      (cfg.intr_vif.pins[AcqStretch])  proc_intr_acqstretch();
+      else if (cfg.intr_vif.pins[TxStretch])   proc_intr_txstretch();
+      else if (cfg.intr_vif.pins[CmdComplete]) proc_intr_cmdcomplete();
+      else if (cfg.read_all_acq_entries)       read_acq_fifo(0, acq_fifo_empty);
+      else begin
 
-        if (ack_ctrl_stretch) begin
-          auto_ack_bytes = $urandom_range(1, 30);
-          csr_wr(.ptr(ral.target_ack_ctrl.nbytes), .value(auto_ack_bytes));
-        end
-
-        if (acq_full) begin
-          if (cfg.slow_acq) begin
-            acq_fifo_empty = 0;
-            while (!acq_fifo_empty) begin
-              delay = $urandom_range(50, 100);
-              #(delay * 1us);
-              read_acq_fifo(1, acq_fifo_empty);
-            end
-          end else begin
-            read_acq_fifo(0, acq_fifo_empty);
-          end
-        end
-      end else if (cfg.intr_vif.pins[TxStretch]) begin
-        proc_intr_txstretch();
-      end else if (cfg.intr_vif.pins[CmdComplete]) begin
-        proc_intr_cmdcomplete();
-      end else if (cfg.read_all_acq_entries) begin
-        read_acq_fifo(0, acq_fifo_empty);
-      end else begin
+        // The variable "expected_intr[]" is used to keep track of interrupts the testbench
+        // expects the possibility of seeing asserted. If an interrupt is asserted that is
+        // not marked in this way, throw an error.
         for (int i = 0; i < NumI2cIntr; i++) begin
           i2c_intr_e my_intr = i2c_intr_e'(i);
-          if (!expected_intr.exists(my_intr)) begin
-            if (cfg.intr_vif.pins[i] !== 0) begin
+          if ((!expected_intr.exists(my_intr)) && // Not expected
+              (cfg.intr_vif.pins[i] !== 0)) begin // Triggered
               `uvm_error("process_target_interrupts",
                          $sformatf("Unexpected interrupt is set %s", my_intr.name))
             end
           end
         end
       end // else: !if(cfg.intr_vif.pins[CmdComplete])
+
       // When bad command is dropped, expected rd_byte is adjust in 'write_tx_fifo'.
       // But for the last bad read command, we have to adjust expected rd_byte separately
       // because we will not call 'write_tx_fifo'.
@@ -1209,6 +1194,48 @@ class i2c_base_vseq extends cip_base_vseq #(
     // causes deadlock. Set verify_clear to 0 to avoid deadlock
     clear_interrupt(TxStretch, 0);
   endtask // proc_intr_txstretch
+
+  // The TB's interrupt handler should call this routine whenever the "acq_stretch"
+  // interrupt is pending.
+  //
+  virtual task proc_intr_acqstretch();
+    // 'AcqStretch' can mean two things
+    // 1) The N-Byte ACK counter is zero. (if this feature is enabled)
+    // 2) The ACQFIFO is full
+
+    bit [TL_DW-1:0] status;
+    bit             ack_ctrl_stretch;
+    bit             acq_full;
+
+    csr_rd(.ptr(ral.status), .value(status));
+    ack_ctrl_stretch = bit'(get_field_val(ral.status.ack_ctrl_stretch, status));
+    acq_full = bit'(get_field_val(ral.status.acqfull, status));
+
+    // 1) (if enabled) Add a random number to the N-Byte ACK counter, allowing the FSM to continue.
+    if (ack_ctrl_stretch) begin
+      int auto_ack_bytes = $urandom_range(1, 30);
+      csr_wr(.ptr(ral.target_ack_ctrl.nbytes), .value(auto_ack_bytes));
+    end
+
+    // 2) Read the ACQFIFO if it is currently full.
+    if (acq_full) begin
+      bit acq_fifo_empty = 0;
+      case (cfg.slow_acq)
+        // Read the fifo until empty.
+        1'b0: read_acq_fifo(0, acq_fifo_empty);
+        // Read the fifo until empty, but slowly (1-word at a time with delays inbetweeen).
+        1'b1: begin
+          while (!acq_fifo_empty) begin
+            int delay = $urandom_range(50, 100);
+            #(delay * 1us);
+            read_acq_fifo(1, acq_fifo_empty);
+          end
+        end
+        default:;
+      endcase
+    end
+
+  endtask: proc_intr_acqstretch
 
   // return target address between, address0, address1 and illegal address
   // illegal address can be return ony if cfg.bad_addr_pct > 0
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
@@ -947,50 +947,66 @@ class i2c_base_vseq extends cip_base_vseq #(
   virtual task proc_intr_unexpstop; endtask
   virtual task proc_intr_hosttimeout; endtask
 
-  // Target-mode interrupt handling routine.
+  // Target-mode interrupt handling.
   //
-  // This routine loops through all wired-interrupts, and calls a handler routine
-  // for each one if active. It then does some checking based on 'expected_intr[]'.
-  //
+  // This routine loops through all active target-mode wired interrupts, and
+  // randomly calls one of the handler routines.
   virtual task process_target_interrupts();
     bit acq_fifo_empty;
     bit read_one = 0;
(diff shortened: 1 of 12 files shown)


hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
+  // CLASS VARIABLES //
+  /////////////////////
+
+  protected bit timer_expired = 0;
+
+  typedef enum int {
+    stim_gen,
+    pre_hook,
+    stim,
+    post_hook
```
rswarbrick, line 40: I'd probably suggest changing these names to CamelCase. I spent a while believing that something that said `state == stim` was checking whether the state equalled the one given in a `stim` variable!

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_reset_acq_vseq.sv
```diff
@@ -50,14 +49,16 @@ class i2c_target_fifo_reset_acq_vseq extends i2c_target_runtime_base_vseq;
...
+    int acq_rd_cyc = 9 * (thigh + tlow);
 
     while (!cfg.stop_intr_handler) begin
-      delay = $urandom_range(0, 10);
+      int delay = $urandom_range(0, 10);
       cfg.clk_rst_vif.wait_clks(delay * acq_rd_cyc);
       // pop one all entries by 25% of chance.
       pop_all = ($urandom_range(0,3) > 0);
-      if (!pause_seq) read_acq_fifo(~pop_all, fifo_empty);
+      if ((runtime_state inside {stim})) read_acq_fifo(~pop_all, fifo_empty);
```
rswarbrick, line 61: No big deal, but I'd probably use `==` here: checking whether something is inside a singleton set always looks a bit weird to me! :-)

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
-  // one acq entry can be popped every 9 bit cycles (8bit + ack)
-  int acq_rd_cyc;
-  i2c_target_base_seq m_i2c_host_seq;
+  ///////////////////
+  // CONTROL KNOBS //
+  ///////////////////
+
+  // Sequence run time in us.
+  // Derived sequences should set this to an appropriate value in 'pre_start'.
+  int seq_runtime_us = 0;
```
rswarbrick, line 28: If my grepping is correct, I think the two sequences that set this both set it to 10000. Would it make sense to move that "sensible starting value" to the base sequence?

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_reset_tx_vseq.sv
```diff
@@ -14,31 +14,28 @@ class i2c_target_fifo_reset_tx_vseq extends i2c_target_runtime_base_vseq;
...
+    // flush acq
+    ral.fifo_ctrl.txrst.set(1'b1);
+    csr_update(ral.fifo_ctrl);
+
+    // clean up sb
+    cfg.scb_h.target_mode_rd_obs_fifo.flush();
+    cfg.scb_h.mirrored_txdata = '{};
+
+    // random delay before the next round
+    #($urandom_range(1, 5) * 10us);
```
rswarbrick, line 37: Before this tidyup, the code spent 10us in a state with `pause_seq = 0` and before the next event. I don't *think* that matters, but I can't see an equivalent here. Are we happy to drop that?

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_reset_tx_vseq.sv
```diff
@@ -49,6 +46,9 @@ class i2c_target_fifo_reset_tx_vseq extends i2c_target_runtime_base_vseq;
     bit fifo_empty, pop_all;
     int delay;
 
+    // one acq entry can be popped every 9 bit cycles (8bit + ack)
+    int acq_rd_cyc = 9 * (thigh + tlow);
```
rswarbrick, line 50: Before this commit, the number got calculated in the base sequence, in `stop_target_interrupt_handler` (I think). Do we need it here as well? (And can we get rid of either side?)

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
+
+  /////////////////////
+  // CLASS VARIABLES //
+  /////////////////////
+
+  protected bit timer_expired = 0;
```
rswarbrick, line 34: Maybe worth adding a 1-line documentation comment for this class variable?

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
     initialization();
-    `uvm_info("cfg_summary", $sformatf("target_addr0:0x%x target_addr1:0x%x num_trans:%0d",
-                             target_addr0, target_addr1, num_trans), UVM_MEDIUM)
-    acq_rd_cyc = 9 * (thigh + tlow);
+
     fork
       begin
         fork
-          while (timer_expired == 0) begin
+          while (!(timer_expired)) begin
```
rswarbrick, line 64: Nit: I don't think we need the inner parens here :-)

(It might be worth searching through this diff for similar points. I think I've seen at least one other)

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
+            // Construct the I2C Controller sequence that defines the agent stimulus
+            // for this test.
             `uvm_create_obj(i2c_target_base_seq, m_i2c_host_seq)
             create_txn(txn_q);
             fetch_txn(txn_q, m_i2c_host_seq.req_q);
+
+            runtime_state = pre_hook;
+
+            wait(runtime_state == stim);
```
rswarbrick, line 73: If I understand this correctly, we expect the FSM to be a cycle, so we'd always expect the change to be `pre_hook -> stim`. If that's right, I'd suggest changing this to `wait(runtime_state != pre_hook)`, followed by an assertion that `runtime_state` is now `stim`.

Similarly for the other places below. I'd probably stick this in a `wait_for_fsm_transition(old_state,new_state)` task.

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
         timer_expired = 1;
-        `DV_WAIT(cfg.m_i2c_agent_cfg.got_stop, "got_stop wait timeout occurred!",
-                 cfg.spinwait_timeout_ns, "target_runtime_base_vseq")
-        cfg.stop_intr_handler = 1;
-      end
-      begin
-        test_event();
+        // If we are currently paused at the end of stimulus, the previous txn may have already
+        // ended, and there won't be a stop condition to await.
+        if ((runtime_state inside {stim})) begin
```
rswarbrick, line 98: I'd probably write this as just `if (runtime_state == stim) begin`.

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
+  // (This definition overrides the base_vseq implementation)
+  //
+  // Desc:
+  //   If txfifo is not full, write more data into it.
+  //   (The scoreboard captures these writes to form its expectations)
+  virtual task proc_intr_txstretch();
+    uvm_reg_data_t curr_txlvl;
+    int num_writes, max_writes;
+
+    // one acqfifo entry can be popped every 9 bit cycles (8bit + ack)
```
rswarbrick, line 154: I wonder whether it's worth expanding this slightly to add "and a bit cycle takes thigh + tlow cycles of the underlying clock"?

rswarbrick (commented):
A few notes about the first commit. I haven't got through the following commits yet.

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -1210,6 +1195,48 @@ class i2c_base_vseq extends cip_base_vseq #(
     clear_interrupt(TxStretch, 0);
   endtask // proc_intr_txstretch
 
+  // The TB's interrupt handler should call this routine whenever the "acq_stretch"
+  // interrupt is pending.
+  //
+  virtual task proc_intr_acqstretch();
+    // 'AcqStretch' can mean two things
+    // 1) The N-Byte ACK counter is zero. (if this features is enabled)
```
rswarbrick, line 1202: Nit: I think you want "this feature"

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -958,25 +958,28 @@ class i2c_base_vseq extends cip_base_vseq #(
...
-                         $sformatf("Unexpected interrupt is set %s", my_intr.name))
-            end
-          end
+      if (cfg.intr_vif.pins[AcqThreshold]) proc_intr_acqthreshold(); // status
+      if (cfg.intr_vif.pins[CmdComplete])  proc_intr_cmdcomplete();  // event
+      if (cfg.intr_vif.pins[TxStretch])    proc_intr_txstretch();    // status
+      if (cfg.intr_vif.pins[TxThreshold])  proc_intr_txthreshold();  // status
+      if (cfg.intr_vif.pins[AcqStretch])   proc_intr_acqstretch();   // status
+      if (cfg.intr_vif.pins[UnexpStop])    proc_intr_unexpstop();    // event
+      if (cfg.intr_vif.pins[HostTimeout])  proc_intr_hosttimeout();  // event
```
rswarbrick, line 963: I wonder whether it would make sense to use `randcase` so that we don't have an opinion about the order these get handled in? I'm thinking something like this:
```systemverilog
      if (cfg.intr_vif.pins) begin
        randcase
          cfg.intr_vif.pins[AcqThreshold]: proc_intr_acqthreshold(); // status
          cfg.intr_vif.pins[CmdComplete]:  proc_intr_cmdcomplete();  // event
          cfg.intr_vif.pins[TxStretch]:    proc_intr_txstretch();    // status
          cfg.intr_vif.pins[TxThreshold]:  proc_intr_txthreshold();  // status
          cfg.intr_vif.pins[AcqStretch]:   proc_intr_acqstretch();   // status
          cfg.intr_vif.pins[UnexpStop]:    proc_intr_unexpstop();    // event
          cfg.intr_vif.pins[HostTimeout]:  proc_intr_hosttimeout();  // event
        endcase
      end
```
(Or maybe use a while loop if we don't want to do the stuff after this between handling two interrupts that were asserted at once)

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -1264,4 +1264,28 @@ class i2c_base_vseq extends cip_base_vseq #(
     return (addr == target_addr0 || addr == target_addr1);
   endfunction
 
+  // Check an interrupt matches our expectation (intr_state + wired)
+  //
```
rswarbrick, line 1264: Maybe it makes sense to expand this comment to explain how we're using `expected_intr` here (using it to qualify the interrupt pins, but not intr_state)?

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_acq_vseq.sv
```diff
@@ -0,0 +1,95 @@
...
+    cfg.max_data = I2C_ACQ_FIFO_DEPTH;
+
+    // Disable ACK-Control mode ('acqstretch' can only be due to acqfull)
+    cfg.ack_ctrl_en = 0;
+  endtask: pre_start
+
+  virtual task test_event();
+    fork
+      super.test_event();
+      count_edge_intr(AcqThreshold, acq_threshold_intr_poscnt, acq_threshold_intr_negcnt);
```
rswarbrick, line 43: Would it make sense to explicitly zero both of these counters at the start of the task, or maybe in `start_of_stim_hook`? This might be a bit easier than chaining the "setup" stuff to the end of `end_of_stim_hook`.

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_acq_vseq.sv
```diff
@@ -0,0 +1,95 @@
...
+    bit acq_fifo_empty;
+
+    // The ACQFIFO begins empty. We stimulated a transaction which should
+    // have generated at-least as much data as the configured threshold.
+
+    check_one_intr(AcqThreshold, 1);
+    `DV_CHECK_EQ(acq_threshold_intr_poscnt, 1, "acq_threshold interrupt did not trigger correctly!")
+    `DV_CHECK_EQ(acq_threshold_intr_negcnt, 0, "acq_threshold de-asserted when not expected!")
+
+    // If we filled the acqfifo, wait for it not to be full, that way we
```
rswarbrick, line 57: Grammar nit: I think this needs a full stop: "... full. That way we ..."

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
```diff
@@ -0,0 +1,107 @@
...
+//
+// Desc:
+//   Stimulate read transactions, filling the TXFIFO with data before starting
+//   driving the bus. By choosing a read-size which is larger than the
+//   configured tx_threshold watermark, we can see the "tx_threshold" interrupt
+//   assert when populating with data, then de-assert as the bus stimulus empties
+//   the fifo.
+//
+// (DUT/Agent == Target/Controller)
+// +use_intr_handler=1 -> Interrupt Driven
```
rswarbrick, line 16: I think I understand, but maybe this would be clearer if expanded to sentences?

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
```diff
@@ -0,0 +1,107 @@
...
+      count_edge_intr(TxThreshold, tx_threshold_intr_poscnt, tx_threshold_intr_negcnt);
+    join_any
+  endtask: test_event
+
+  virtual task start_of_stim_hook();
+    // At the start of stimulus, the txfifo should be empty.
+    // Unless the threshold is also configured to zero, this means the
+    // interrupt should be asserted.
+    csr_rd_check(.ptr(ral.target_fifo_status.txlvl), .compare_value(0));
+    check_one_intr(TxThreshold, 1);
```
rswarbrick, line 53: I think this needs something like "if threshold is positive" here?

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
```diff
@@ -0,0 +1,107 @@
...
+    `DV_CHECK_EQ(tx_threshold_intr_negcnt, 1, "tx_threshold de-asserted when not expected!")
+  endtask
+
+  virtual task end_of_stim_hook();
+    bit acq_fifo_empty;
+
+    // After the transaction completes, the txfifo should now be empty, and
+    // (unless threshold == 0) the watermark interrupt should be asserted again.
+
+    check_one_intr(TxThreshold, 1);
```
rswarbrick, line 72: Again, I think we just need to check this when threshold is positive?

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
```diff
@@ -0,0 +1,107 @@
...
+    `DV_CHECK_EQ(tx_threshold_intr_negcnt, 1, "acq_threshold de-asserted when not expected!")
+
+    // Empty out the ACQFIFO now.
+    read_acq_fifo(.read_one(0), .acq_fifo_empty(acq_fifo_empty));
+
+    #1ns; // Otherwise our cleanup races with the scoreboard checking the acqfifo read above.
+
+    // Cleanup for next iteration.
+    tx_threshold_intr_poscnt = 0;
+    tx_threshold_intr_negcnt = 0;
```
rswarbrick, line 83: Can we put this into `start_of_stim_hook`? I think the timing gets easier that way?

andreaskurth (commented):
This LGTM overall, thx @hcallahan-lowrisc! I focused my review on `i2c_target_runtime_base_vseq` and the new watermarks vseqs, which fulfill the corresponding testplan items as far as I can tell.

It would be best to get approvals from @rswarbrick and @a-will as well

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_reset_tx_vseq.sv
hcallahan-lowrisc: Yes that delay is unnecessary here.

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_reset_tx_vseq.sv
hcallahan-lowrisc: I pulled the `process_acq` routine up into the parent class, so both tests can share the routine.

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_acq_vseq.sv
hcallahan-lowrisc: Moved to a new 'setup_hook()' routine that is called at the top of the loop each time.

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
hcallahan-lowrisc: I think it's nice to be explicit in each sequence about how long it will run for, so I'm going to leave this as-is. I also added a check in the base seq that a non-zero timeout has been configured.

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
hcallahan-lowrisc: I refactored this into some tasks for clarity...
```
    advance_runtime_state(.to(StPreHook));
    wait_for_runtime_state(.await(StStim));
```
I left the wait-check to be blocking until the .await state is reached, as I added a couple of extra states to the enum, and it made the checks a little confusing as it was no-longer so rigidly cyclical. I think since this is an abstraction to help out DV authors during test development, it should be fine.

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
hcallahan-lowrisc: I reworked the checking to handle the threshold = 0 case.

a-will (approved):
LGTM, with just a nit from me.

I haven't had time to look thoroughly at i2c_target_runtime_base_vseq.sv (e.g. evaluating APIs and the overall flow), but the subclasses look sensible to me.

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
hcallahan-lowrisc: Yep this is a nicer way to do it, I have now pushed this change.

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
```diff
@@ -0,0 +1,107 @@
...
+
+// This test provides a basic check of the TXFIFO watermarks and watermark
+// interrupts.
+//
+// Desc:
+//   Stimulate read transactions, filling the TXFIFO with data before starting
+//   driving the bus. By choosing a read-size which is larger than the
+//   configured tx_threshold watermark, we can see the "tx_threshold" interrupt
+//   assert when populating with data, then de-assert as the bus stimulus empties
+//   the fifo.
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Yeah that's much clearer. Thanks!

hw/ip/i2c/dv/env/seq_lib/i2c_target_runtime_base_vseq.sv
```diff
@@ -2,91 +2,177 @@
...
-            // 'cfg.m_i2c_agent_cfg.got_stop = 1'
-            cfg.clk_rst_vif.wait_clks(1);
-            `DV_WAIT(pause_seq == 0, "pause_seq == 0 wait timeout occurred!",
-                     cfg.spinwait_timeout_ns, "target_runtime_base_vseq")
+
+            wait(runtime_state == stim_gen);
           end
+          // Process all target-mode interrupts until stopped by the timer expiring.
           process_target_interrupts();
         join
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Yeah the `cfg.stop_intr_handler` causes the other thread to terminate as well. I've refactored this logic now to bring the while loop exit condition outside the routine, so it should be a lot clearer now what is going on. Thanks

hcallahan-lowrisc:
@rswarbrick Would you mind taking another look?

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
```diff
@@ -0,0 +1,133 @@
...
+// interrupts.
+//
+// Desc:
+//   Stimulate read transactions, filling the TXFIFO with data before starting
+//   driving the bus. By choosing a read-size which is larger than the
+//   configured tx_threshold watermark, we can see the "tx_threshold" interrupt
+//   begin asserted when populating with data, de-assert once the TXFIFO is filled
+//   beyond the threshold, then assert again as the bus stimulus empties the fifo.
+//
+// (DUT:Agent == Target:Controller)
```
rswarbrick, line 15: Minor nit: I'm not sure I know how to parse this line of the comment :-)

rswarbrick (approved):
Other than an extremely minor question, this looks really sensible. Thanks for the reminder.

hw/ip/i2c/dv/env/seq_lib/i2c_target_fifo_watermarks_tx_vseq.sv
hcallahan-lowrisc: This is just a note to clarify the operating modes of the two endpoints for this test. The I2C spec uses Controller where OpenTitanCIP uses Host, and Target where OTCIP uses Device. It's mostly to orient the reader, as we test the DUT operating in both modes, it's easy to become disoriented as to where the data is moving in any given read or write txn. And by the reader, I mostly mean me :)

PR #23143, opened 2024-05-15
[manuf] remove HW_CFG measurement from UDS cert

timothytrippel:
This measurement is not relevant to the overall identity of the device. Moreover, the device ID (which is part of the HW_CFG partition) is already mixed into the keyladder.

This reduces the UDS cert size as well.


pamaury (approved):

timothytrippel:
CI failures are known issues on SiVal branch. Merging.

Issue #23149, opened 2024-05-16
[sensor_ctrl,OTP] Implement AST alert reset configuration and post-reset behavior


andreaskurth:
As described in [this implementation recommendation](https://docs.google.com/document/d/1KTAkLexrTPsEyJrMc6Auxiuxdv161GdEnyTbjLDY6Fw/edit): At reset we want all alerts in sensor_ctrl to be enabled and set to recoverable. During ROM boot and _before_ alert_handler is configured, we want ROM to read OTP values (from the `OWNER_SW_CFG` OTP partition) to disable specific alerts in sensor_ctrl or set them to fatal.

## TODOs
- [x] Change the reset value of sensor_ctrl’s ALERT_EN CSRs to `MuBiTrue`.
  --> #23152
- [x] Mention this design in sensor_ctrl’s documentation.
  --> #23152
- [x] Add a section in the `OWNER_SW_CFG` OTP partition to store the sensor_ctrl alert configuration as well as one section to store its digest (one for all LC states should be sufficient).
  --> #23155

Note that this issue is about changes to sensor_ctrl RTL and OTP partitions. See #23150 for the changes to ROM.

PR #23152, opened 2024-05-16
[sensor_ctrl] Change reset value of `ALERT_EN` CSRs to MuBiTrue

andreaskurth:
Part of #23149; see that issue for details.

PR #23155, opened 2024-05-16
[otp_ctrl_mmap] Add item to configure sensor_ctrl alerts

andreaskurth:
This is part of #23149; please see that issue for details.


hw/ip/otp_ctrl/data/otp_ctrl_mmap.hjson
```diff
@@ -335,6 +335,14 @@
                     name: "OWNER_SW_CFG_ROM_EXT_BOOTSTRAP_EN",
                     size: "4"
                 }
+                {
+                    name: "OWNER_SW_CFG_ROM_SENSOR_CTRL_ALERT_CFG",
+                    size: "4" // 2 bit per alert (disabled, fatal) for (currently) 11 alerts -> 32 bit suffice
+                }
```
andreaskurth, line 341: @timothytrippel Do you agree that 2 OTP bit per alert (one to disable it, one to make it fatal) are suitable from a provisioning and ROM perspective?

timothytrippel: That suffices, but do we want these to be mubis? Since the underlying CSRs in SENSOR_CTRL that this info will get copied to, perhaps it does not matter (in #23152)?

timothytrippel (approved):

hw/ip/otp_ctrl/data/otp_ctrl_mmap.hjson
andreaskurth: We do have integrity protection on the config bits in OTP (through the digest) and while getting sent over TL-UL (ECC), so I don't think they need to be MuBi in OTP.

andreaskurth: @vogelpi WDYT?

vogelpi: I think 2 bits per alert should suffice as we also have the digest. If we store the OTP values as MuBis, this would require quite some bits on top. I don't know how much we have left. How do we store similar configuration values in OTP?

andreaskurth: Alert classification is stored as 8-bit sparse values in OTP (as is alert class enable and escalation policies): https://github.com/lowRISC/opentitan/blob/2aca3e991a8d36d46d48f886a868aa87fdf1e781/sw/device/silicon_creator/lib/drivers/alert.h#L19-L41

andreaskurth: Changed to MuBi4

vogelpi: Thanks @andreaskurth , I think this makes sense.

PR #23156, opened 2024-05-16
[sw/fi] Add firmware code for OTP data read FI penetration test

wettermo:
This adds firmware code for FI pen-testing the OTP module.

The test sequence can be summarized in the following way:
  1. Read content from a selected OTP partition for comparison values
  2. Perform fault injection on nop instructions
  3. Read content from the same OTP partition again
  4. Compare output with previous output and check if values differ

The corresponding PR in lowRISC/ot-sca is https://github.com/lowRISC/ot-sca/pull/367


sw/device/tests/penetrationtests/firmware/fi/otp_fi.c
```diff
@@ -0,0 +1,320 @@
...
+  NOP1000;
+  NOP1000;
+
+  sca_set_trigger_low();
+
+  // Read OTP partition again to see if values changed
+  TRY(otp_vendor_test_dump(otp_read32_result_vendor_test_fi));
+
+  // Get registered alerts from alert handler.
+  reg_alerts = sca_get_triggered_alerts();
```
nasahlpa, line 119: Just a FYI - it might be useful to also register the alerts triggered inside Ibex. You could do something similar to here:
https://github.com/lowRISC/opentitan/blob/08d6cba5da6ced9d92be9a85b6c409e5a63dbd85/sw/device/tests/penetrationtests/firmware/fi/ibex_fi.c#L221-L223

Also applies to the test functions below.

sw/device/tests/penetrationtests/firmware/fi/otp_fi.c
```diff
@@ -0,0 +1,320 @@
...
+  for (uint32_t i = 0; i < OTP_CTRL_PARAM_VENDOR_TEST_SIZE / 4; i++) {
+    uj_output.vendor_test_comp[i] = otp_read32_result_vendor_test_comp[i];
+    uj_output.vendor_test_fi[i] = otp_read32_result_vendor_test_fi[i];
+  }
+  uj_output.otp_status_codes = status.codes;
+  memcpy(uj_output.otp_error_causes, (uint8_t *)status.causes,
+         kDifOtpCtrlStatusCodeHasCauseLast + 1);
+  uj_output.alerts[0] = reg_alerts.alerts[0];
+  uj_output.alerts[1] = reg_alerts.alerts[1];
+  uj_output.alerts[2] = reg_alerts.alerts[2];
```
nasahlpa, line 136:
```suggestion
  memcpy(uj_output.alerts, reg_alerts.alerts, sizeof(reg_alerts.alerts));
```
Would also work.

Also applies to the test functions below.

sw/device/tests/penetrationtests/firmware/fi/otp_fi.c
```diff
@@ -0,0 +1,320 @@
...
+  reg_alerts = sca_get_triggered_alerts();
+
+  // Get OTP CTRL status
+  dif_otp_ctrl_status_t status;
+  TRY(dif_otp_ctrl_get_status(&otp, &status));
+
+  // Send result & status codes to host.
+  otp_fi_hwcfg_partition_t uj_output;
+  for (uint32_t i = 0; i < OTP_CTRL_PARAM_HW_CFG_SIZE / 4; i++) {
+    uj_output.hw_cfg_comp[i] = otp_read32_result_hw_cfg_comp[i];
```
nasahlpa, line 218: Is there a reason why you are not doing the comparison on the chip instead of sending back all data?

In terms of performance, it could be faster to do the comparison on the chip to avoid the communication overhead?

sw/device/tests/penetrationtests/firmware/fi/otp_fi.c
```diff
@@ -0,0 +1,320 @@
...
+  // FI code target.
+  sca_set_trigger_high();
+
+  // Point for FI
+  NOP1000;
+  NOP1000;
+  NOP1000;
+  NOP1000;
+
+  sca_set_trigger_low();
```
nasahlpa, line 248: Did you check with the oscilloscope whether this trigger window stays constant when executing this function multiple times? If not, it might be worth to set the reseeding interval to max in the init, .e.g.:
https://github.com/lowRISC/opentitan/blob/08d6cba5da6ced9d92be9a85b6c409e5a63dbd85/sw/device/tests/penetrationtests/firmware/fi/otbn_fi.c#L520

sw/device/tests/penetrationtests/json/otp_fi_commands.h
```diff
@@ -0,0 +1,59 @@
...
+#define OTPFI_VENDORTEST_PARTITION(field, string) \
+    field(vendor_test_comp, uint32_t, 16) \
+    field(vendor_test_fi, uint32_t, 16) \
+    field(otp_status_codes, uint32_t) \
+    field(otp_error_causes, uint8_t, 10) \
+    field(alerts, uint32_t, 3)
+UJSON_SERDE_STRUCT(OtpFiVendortestPartition, otp_fi_vendortest_partition_t, OTPFI_VENDORTEST_PARTITION);
+
+#define OTPFI_OWNERSWCFG_PARTITION(field, string) \
+    field(owner_sw_cfg_comp, uint32_t, 200) \
```
nasahlpa, line 31: Where are these sizes coming from (e.g. 200)? Would it make sense to use variables here to make sure that we do not accidentally try sending too many / less data over UART back to the host?

nasahlpa (approved):
Thanks, LGTM apart from some smaller remarks.

sw/device/tests/penetrationtests/firmware/fi/otp_fi.c
nasahlpa: When doing measurements with the oscilloscope, I've noticed that sometimes the trigger windows (i.e., the time between `sca_set_trigger_high()` and `sca_set_trigger_low()`) is not constant. The root cause for this was that there were several entropy complex reseed that delayed the execution.

sw/device/tests/penetrationtests/json/otp_fi_commands.h
nasahlpa: That would be good, thanks!

wettermo:
Thanks Pascal for the review, I added everything. I also updated the PR in ot-sca (https://github.com/lowRISC/ot-sca/pull/367) accordingly.

PR #23157, opened 2024-05-16
[fpga] Disable FPGA-only SCA-specific optimizations on CW310 and CW340

vogelpi:
To facilitate pre-silicon SCA experiments and optimization of SCA countermeasures before tapeout, we've implemented a couple of FPGA-only optimizations for AES, KMAC and OTBN such as
- Trigger delays
- Skipping PRNG reseeds at OTBN application start
- Accepting writes to the KMAC message FIFO before triggering operations

Now that we're marching towards tapeout and we've concluded most of the SCA work, it's better to disable these optimizations to ensure that the FPGA targets are as close to the real chip in terms of entropy consumption, SW interface and timing behavior. If needed, these optimizations can still be turned on locally.


nasahlpa (approved):
LGTM!

timothytrippel (approved):

PR #23159, opened 2024-05-16
[jtag,dv] Handle trst_n at start of time in jtag_driver.sv

rswarbrick:
The existing code worked as long as the get_and_drive() task started after we had driven trst_n to zero AND we didn't ever want to drive it to zero afterwards.

It turns out that this wasn't the situation I was seeing in my simulations and the result was that the driver thought the JTAG FSM was in the RTI state, but it was actually in Test-Logic-Reset and everything got rather confused.

With this change, we will always notice if we see a trst_n reset, and the driver and bus shouldn't get out of sync in the same way.

(This PR contains just the first commit from #23097, a PR that's failing in CI. I'm hopeful that the failures there are caused by one of the later commits, and *this one* fixes a bunch of tests that failed on Tuesday night).


rswarbrick:
Darn, that's no good. Looks like this was the commit causing the problem. I'll debug and fix it now, but will close this PR and work on the original one instead.

PR #23160, opened 2024-05-16
[dv,otp_ctrl] Check ECC errors in VENDOR_TEST partition

### LINKED ISSUE #21265: [otp_ctrl/dv] Make sure the otp_ctrl tolerates spurious ECC errors on the VENDOR_TEST partition
### Description

In https://github.com/lowRISC/opentitan/pull/21256 we added a patch that filters ECC errors on partitions without integrity. This is done so that the updated OTP_CTRL design (with updates from Darjeeling) is still compatible with the previously implemented `prim_otp` wrapper from NTIL.

We need to test both in open source and with the closed source wrappers that uncorrectable ECC errors reported by the `prim_otp` wrapper do not cause the partition to report fatal alerts.

This is important as otherwise a VENDOR_TEST partition with garbage in the ECC bits may lock up the chip.

matutem:
Add a top level test that injects singe and double bit errors in the VENDOR_TEST partition, expecting no error to occur reading the otp_ctrl `STATUS` csr.

Fixes #21265

### DIFF
--- a/hw/ip/otp_ctrl/data/otp_ctrl_base_vseq.sv.tpl
+++ b/hw/ip/otp_ctrl/data/otp_ctrl_base_vseq.sv.tpl
@@ -352,9 +352,11 @@ class otp_ctrl_base_vseq extends cip_base_vseq #(
     end
   endtask
 
-  // This function backdoor inject error according to ecc_err.
-  // For example, if err_mask is set to 'b01, bit 1 in OTP macro will be flipped.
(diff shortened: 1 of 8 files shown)


vogelpi:
@sha-ron , are you happy if we move forward and merge this?

sha-ron:
> @sha-ron , are you happy if we move forward and merge this?

Yes please. Thanks.

PR #16909, opened 2022-12-21
[kmac] Expose the number and size of message FIFO entries.

jadephilipoom:
This change arose from a conversation with @eunchan about the KMAC message FIFO. Existing documentation recommended that software write messages by:
1. Reading the `FIFO_DEPTH` field from the status register.
2. Calculating the remaining space in the FIFO and writing that amount of input.
3. Repeat.

However, it was unclear from the documentation (a) what exactly `FIFO_DEPTH` meant (bytes? words? Space free or space occupied?) and (b) how software should calculate the remaining space. With @eunchan , we confirmed that `FIFO_DEPTH` means "the number of entries currently occupied", and therefore software actually needs more information in order to calculate the remaining space -- the number of entries occupied is not useful unless you also know the number of total entries (10) and their size (64b). Rather than hard-code these into the drivers, we thought it would be better to expose them through `kmac_regs.h` and include checks against SV parameters to make sure their values are correct.

This PR:
- Exposes FIFO entry number and size as parameters in `kmac.hjson`
- Adds checks that compare these parameters to specific SV variables
- Changes the depth value noted in the docs from 9 to 10 (it was previously incorrect)
- Updates the KMAC docs to be clearer about `FIFO_DEPTH` and include a code snippet showing how to write input
- Updates `dif_kmac_absorb()` to use the new parameters and match the recommended way of writing
  - Previously, it did not monitor the FIFO at all and would have simply continued writing if the FIFO became full

See also: https://github.com/lowRISC/opentitan/issues/16902 (possibilities for a future KMAC/software interface)

@ballifatih It may be a good idea to use these values in the cryptolib KMAC driver as well; what do you think?

PR #23162, opened 2024-05-16
[rom_ext] Add KMAC256 operations using software and hardware keys

cfrantz:
1. Add KMAC256 functionality to the silicon_creator driver.
2. Add NIST test vectors to the functest to ensure the kmac functions produce correct results.
3. Add the capability to select a hardware destination to the `keymgr` driver.

### DIFF
--- a/sw/device/silicon_creator/lib/drivers/BUILD
+++ b/sw/device/silicon_creator/lib/drivers/BUILD
@@ -375,6 +375,7 @@ opentitan_test(
         ":kmac",
         "//hw/top_earlgrey/sw/autogen:top_earlgrey",
         "//sw/device/lib/testing:entropy_testutils",
+        "//sw/device/lib/testing:hexstr",
         "//sw/device/lib/testing/test_framework:ottf_main",
         "//sw/device/silicon_creator/lib:error",
     ],
--- a/sw/device/silicon_creator/lib/drivers/keymgr.c
+++ b/sw/device/silicon_creator/lib/drivers/keymgr.c
@@ -203,8 +203,8 @@ static rom_error_t keymgr_wait_until_done(void) {
   return kErrorKeymgrInternal;
 }
 
-rom_error_t sc_keymgr_generate_key_otbn(
-    sc_keymgr_key_type_t key_type,
+rom_error_t sc_keymgr_generate_key(
+    sc_keymgr_dest_t destination, sc_keymgr_key_type_t key_type,
     sc_keymgr_diversification_t diversification) {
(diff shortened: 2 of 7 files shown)


timothytrippel (approved):

jadephilipoom:
Sorry for the post-merge comment (I was out sick for a bit and didn't see this earlier), but I think the implementation of `kmac_shake256_absorb` now needs to change for the reasons discussed in this comment:
https://github.com/lowRISC/opentitan/blob/c4666f429474acc507064e5448fa4d347fe5e611/sw/device/silicon_creator/lib/drivers/kmac.c#L331-L339

Basically, the previous implementation made use of the assumption that the KMAC block will never be in KMAC mode for ROM in order to make the SHAKE256 mode faster for SPHINCS+. But if we're allowing KMAC, we need to use a more conservative approach to avoid a potential deadlock scenario between entropy complex, Ibex, and KMAC.

cfrantz:
https://github.com/lowRISC/opentitan/blob/c4666f429474acc507064e5448fa4d347fe5e611/sw/device/silicon_creator/lib/drivers/kmac.c#L331-L339
> 
> Basically, the previous implementation made use of the assumption that the KMAC block will never be in KMAC mode for ROM in order to make the SHAKE256 mode faster for SPHINCS+. But if we're allowing KMAC, we need to use a more conservative approach to avoid a potential deadlock scenario between entropy complex, Ibex, and KMAC.

The document link in that comment is stale, but I think this is the correct link: https://opentitan.org/book/hw/ip/kmac/doc/theory_of_operation.html#fifo-depth-and-empty-status

Apart from describing how to write to the FIFO, that document doesn't give any indication what one is supposed to do if the KMAC is waiting for EDN.  IOW, will I simply transform a hardware wait condition into a software busy-loop?

jadephilipoom:
> Apart from describing how to write to the FIFO, that document doesn't give any indication what one is supposed to do if the KMAC is waiting for EDN. IOW, will I simply transform a hardware wait condition into a software busy-loop?

Sort of, yes. I found the [original discussion](https://github.com/lowRISC/opentitan/pull/16909#issuecomment-1361792331) from a couple of years ago; from my rough understanding of hardware design, there was concern about a rare corner case of interactions between KMAC, EDN, and Ibex that could cause a lock-up, and checking the FIFO depth in software avoids this possibility. However, I see now that one of those conditions is that SW doesn't set the entropy wait timer for KMAC, so perhaps if this timer is always set to a nonzero value in ROM we don't have to use the more complicated FIFO strategy.

PR #23164, opened 2024-05-17
[pattgen,sival] Port pattgen_ios to SiVal

pamaury:
This port required a number of changes to the C code: the DV test uses random parameters and only runs once. In SiVal, it's more effective to run it a loop so the code was changed to be able to run several several times. Also more synchronization was necessary between the host and device to achieve proper signal acquisition. I also had to change the way interrupts to handle OTTF flow control.


pamaury:
Failure is due to the FPGA having a difference peripheral clock than the silicon. I updated the test so that the device communicate the frequency to the host.

rules/signing.bzl
```diff
@@ -363,7 +363,7 @@ def _hsmtool_sign(ctx, tool, digest, ecdsa_key, rsa_key, spxmsg = None, spx_key
         inputs = [digest, rsa_key.file, tool.tool] + tool.data,
         arguments = [
             "--quiet",
-            "--lockfile=/tmp/hsmtool.lock",
+            "--lockfile=/tmp/pamaury.hsmtool.lock",
```
engdoreis, line 366: Seems like a local change that shouldn't be committed?

sw/device/tests/pattgen_ios_test.c
```diff
@@ -27,26 +30,50 @@
...
-OTTF_DEFINE_TEST_CONFIG();
+OTTF_DEFINE_TEST_CONFIG(.enable_uart_flow_control = true);
+
+enum {
+  kTestCmdTimeoutUsec = 5000000,
+};
+
+/* Macro to define and use variable that need to be stored in flash
+ * for DV tests and RAM for real tests, to enable backdoor access. */
+#define DEFINE_BACKDOOR_VAR(type, name, default_val)                       \
```
engdoreis, line 41: Should this be moved to macros.h?

pamaury:
After investigation, the CI failure is due to a glitch on the `pda1_tx` line that only seems to occur with the FPGA. Weirdly, the glitch is completely deterministic so I am investigating with a logic analyzer to confirm that this is an actual glitch on the line.

rules/signing.bzl
pamaury: yes

sw/device/tests/pattgen_ios_test.c
pamaury: I am not sure I want to make such a change in this PR. We have many instances spread over several tests, I think it's better to create a PR just for that.

pamaury:
I have updated the code. I marked the FPGA test as broken due to the output glitch fixed by #23206. Since the fix is only for the master branch, I will remove the broken tag when backporting.

sw/host/opentitanlib/src/test_utils/gpio_monitor.rs
```diff
@@ -148,6 +178,12 @@ impl<'a> GpioMon<'a> {
                 SYMBOLS[event.signal_index as usize]
             ));
         }
+        // Add a final time to make sure that the last edge is visible.
+        vcd.push_str(&format!(
```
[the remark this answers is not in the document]
pamaury: Yes it's the same because the vcd only contains a timestamp but no changes.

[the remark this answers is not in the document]
nbdd0121: It seems that there's a 1000 offset?

sw/host/opentitanlib/src/test_utils/gpio_monitor.rs
```diff
@@ -110,6 +118,10 @@ impl<'a> GpioMon<'a> {
         Ok(events.events)
     }
 
+    pub fn dump_on_drop(&mut self, dump_on_drop: bool) {
+        self.dump_on_drop = dump_on_drop
```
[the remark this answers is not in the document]
pamaury: I don't want to re-run the CI just to change that, but I'll change it next time I have the opportunity.

nbdd0121 (approved):

sw/device/tests/pattgen_ios_test.c
pamaury: yes that would be a good option. I'll look into this and create another PR.

pamaury:
This will need to manually ported to master with some changes after #23198 lands

Issue #22210, opened 2024-03-23
[cryptotest] //sw/device/tests/crypto/cryptotest:ecdh_kat


moidx:
### Hierarchy of regression failure

Block level

### Failure Description

```shell
[2024-03-23T04:43:43.099Z INFO  harness] Test counter: 332
[2024-03-23T04:43:43.099Z INFO  harness] vendor: wycheproof, test case: 332
[2024-03-23T04:43:43.099Z INFO  opentitanlib::test_utils::rpc] Sending: "Ecdh"
[2024-03-23T04:43:43.106Z INFO  opentitanlib::test_utils::rpc] Sending: "P256"
[2024-03-23T04:43:43.113Z INFO  opentitanlib::test_utils::rpc] Sending: {"d0":[177,101,82,146,221,80,170,112,109,135,24,140,227,103,29,83,28,124,200,238,162,37,236,242,2,68,226,233,184,229,201,55],"d0_len":32,"d1":[220,130,34,162,241,90,174,121,212,243,54,91,204,28,2,31,106,190,33,173,131,42,112,5,27,172,105,135,150,191,128,70],"d1_len":32}
[2024-03-23T04:43:43.410Z INFO  opentitanlib::test_utils::rpc] Sending: {"coordinate":[0],"coordinate_len":1}
[2024-03-23T04:43:43.450Z INFO  opentitanlib::test_utils::rpc] Sending: {"coordinate":[0],"coordinate_len":1}
[2024-03-23T04:43:43.491Z  console]E00002 ecdh.c:123] Unexpected status value returned from otcrypto_ecdh: 0x80006d89
[2024-03-23T04:43:43.498Z  console]RESP_ERR:{"Internal":["ECD",124]} CRC:1516886847

253:Finished test test_ecdh: Err(CRC didn't match received json body.
```

### Steps to Reproduce

- GitHub Revision: 6d3cb9a1f826838a6ac2d405affb6e3fdb903022 `earlgrey_es_sival`
- Bazel: //sw/device/tests/crypto/cryptotest:ecdh_kat_silicon_owner_prodc_rom_ext

Should be reproducible with `sival` sku as well.

### Tests with similar or related failures

- [ ] //sw/device/tests/crypto/cryptotest:ecdh_kat

moidx:
CC @milesdai @jadephilipoom

andreaskurth:
@moidx: Could be a problem in OTBN or with the entropy complex (not sure if entropy is required)

[...]

Issue #22209, opened 2024-03-23
[cryptotest]  //sw/device/tests/crypto/cryptotest:drbg_kat


moidx:
### Hierarchy of regression failure

Chip Level

### Failure Description

```shell
[2024-03-23T04:39:40.894Z INFO  harness] vendor: nist, test case: 6
[2024-03-23T04:39:40.894Z INFO  opentitanlib::test_utils::rpc] Sending: "Drbg"
[2024-03-23T04:39:40.901Z INFO  opentitanlib::test_utils::rpc] Sending: {"entropy":[162,1,5,244,225,238,117,198,137,137,54,127,78,201,156,115,245,190,4,192,94,80,9,157,75,10,210,38,42,237,23,149,178,63,213,133,143,186,150,200,177,154,0,181,103,253,18,227],"entropy_len":48,"personalization_string":[],"personalization_string_len":0,"reseed":1,"reseed_entropy":[154,146,38,164,179,128,157,42,63,159,48,213,91,159,205,31,104,159,112,109,193,130,233,235,103,37,241,17,222,48,180,243,63,142,32,65,238,189,177,38,234,98,68,13,28,193,110,23],"reseed_entropy_len":48,"reseed_additional_input":[],"reseed_additional_input_len":0,"additional_input_1":[],"additional_input_1_len":0,"additional_input_2":[],"additional_input_2_len":0,"output_len":64}
[2024-03-23T04:39:41.679Z  console]RESP_OK:{"output":[81,159,203,7,143,227,75,227,136,193,209,225,219,68,238,88,106,30,124,34,9,125,47,108,135,62,20,233,240,157,224,233,183,92,154,175,210,229,131,161,159,80,89,99,46,229,25,230,61,51,227,89,61,55,32,54,8,160,229,58,134,247,30,181],"output_len":64} CRC:321997622

[2024-03-23T04:39:41.702Z INFO  harness] Test counter: 8
[2024-03-23T04:39:41.702Z INFO  harness] vendor: nist, test case: 7
[2024-03-23T04:39:41.702Z INFO  opentitanlib::test_utils::rpc] Sending: "Drbg"
[...]

PR #23166, opened 2024-05-17
[cryptolib] Fix CSRNG driver

vogelpi:
This PR contains two commits, the really relevant one changes the entropy driver of cryptolib to also poll the CSRNG command ready bit for additional data. Previously, the command register ready bit was only polled for the additional data when writing commands to the SW register of EDN.

When writing commands to CSRNG instead, not polling the command ready bit could cause CSRNG to drop additional data provided by software while handling a hardware command. Once the hardware command was handled, the software instance would win arbitration and CSRNG would wait for software to provide the dropped additional data, thereby stalling the entire entropy complex. This would happen for example if a HW instance request would collide with a software request. Since the software command FIFO is only 2 entries deep, it can only be guaranteed to absorb 2 words if the software interface doesn't win arbitration immediately.

This is related to https://github.com/lowRISC/opentitan/issues/22209 and https://github.com/lowRISC/opentitan/issues/22210.

### DIFF
--- a/sw/device/lib/crypto/drivers/entropy.c
+++ b/sw/device/lib/crypto/drivers/entropy.c
@@ -320,6 +320,7 @@ static status_t csrng_send_app_cmd(uint32_t base_address,
   uint32_t cmd_reg_addr;
   uint32_t sts_reg_addr;
   uint32_t rdy_bit_offset;
+  uint32_t reg_rdy_bit_offset;
   uint32_t reg;
   bool ready;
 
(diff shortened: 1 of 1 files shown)


vogelpi:
FYI: I've generated a bitstream without the recent keymgr hardware configurations and then configured keymgr for maximum entropy consumption while software is using the SW instance of CSRNG. With this change, the several cryptotests are now passing which would fail otherwise. Examples include:
- //sw/device/tests/crypto/cryptotest:drbg_kat, see #22209 
- //sw/device/tests/crypto/cryptotest:ecdh_kat, see #22210

I believe this is the real fix we need for these issues. The keymgr hardware fixes just made the problem to disappear as CSRNG was no longer heavily loaded while SW was using it.

sw/device/lib/crypto/drivers/entropy.c
```diff
@@ -422,12 +403,14 @@ static status_t csrng_send_app_cmd(uint32_t base_address,
...
-    if (cmd_type == kEntropyCsrngSendAppCmdTypeEdnSw) {
+    // Before writing each word of additional data, the command ready or command
+    // reg ready bit needs to be polled if the command is issued to CSRNG or the
+    // SW register of EDN, respectively.
+    if (cmd_type == kEntropyCsrngSendAppCmdTypeCsrng ||
+        cmd_type == kEntropyCsrngSendAppCmdTypeEdnSw) {
       do {
-        reg = abs_mmio_read32(base_address + EDN_SW_CMD_STS_REG_OFFSET);
-        ready = bitfield_bit32_read(reg, EDN_SW_CMD_STS_CMD_REG_RDY_BIT);
+        reg = abs_mmio_read32(sts_reg_addr);
```
moidx, line 412: Since this code will be consumed at higher integration layers, it makes sense to not block forever. Would it make sense to add a timeout mechanism similart to what was done before with `kCsrngIdleNumTries`?

moidx (approved):

sw/device/lib/crypto/drivers/entropy.c
vogelpi: This sounds like a good idea @moidx . I've filed #23179 to track this.

PR #23170, opened 2024-05-17
[prim,fpv] Fix a couple of silly warnings in prim_fifo_sync

rswarbrick:
These shouldn't cause any change in behaviour, but it's nice to silence some of the cruft from the tool (Jasper) as it reads the code.


vogelpi (approved):
LGTM, thanks @rswarbrick !

marnovandermaas (approved):
Thanks.

PR #23171, opened 2024-05-17
[prim,fpv] Tweak how a parameter gets used in some assertions

rswarbrick:
The existing code wasn't strictly legal SystemVerilog because it contained a hierarchical reference in something that is being used as a constant (in the "[*0::xyz]" term).

This is not too difficult to fix, because we can just pass the parameter in explicitly constrain the width and then wire it up properly in the bind statement.

To make sure that the assertions don't change meaning, we add an extra assertion (which uses a hierarchical reference in an allowable context).


vogelpi (approved):
LGTM thanks, also for the explanation :-)

Issue #22182, opened 2024-03-21
[spi_host,dv] Check enable deassertion logic for Dual/Quad segments


hcallahan-lowrisc:
### Description

As described [in this comment](https://github.com/lowRISC/opentitan/pull/21338#issuecomment-1984908004) and fixed[ in this PR](https://github.com/lowRISC/opentitan/pull/22040#discussion_r1534369921), the TLT `chip_sw_spi_host_tx_rx` stimulated a failure when the SIO enable was used to mask the non-transmitting wire `sd_en_o[0]` in Standard Mode RxOnly segments ([original issue](https://github.com/lowRISC/opentitan/issues/14969)). 
This TLT generates stimulus that is significantly different to the block level DV environment, and it should be investigated if the same incorrect enable deassertion can be observed when using Dual or Quad mode segments. This will require block level DV enhancements to stimulate any degenerate behaviour.

antmarzam:
> @antmarzam Maybe you can help @hcallahan-lowrisc on this one?

Sure - Happy to help! :)

I just had a chat with Harry about this, and have now an idea about how to tackle this (typing it here for future reference):
1 - Roll spi_host RTL back before this fix
2 - Write an assertion and ensure it fires when running the test in https://github.com/lowRISC/opentitan/pull/21338#issuecomment-1984908004
3 - Place the assertion in parallel for each of the SPI lanes 
4 - Ensure we get stimulus in the block-level TB which fire the assertion (for lane=0)
5 - Update spi_host RTL to the latest and make sure assertion no longer fires.
6 - Generate the stimulus needed to reproduce the same issue for 2 or 4 lanes
7 - Add some FCOV to see these scenarios are hit by the block level TB

a-will:
For what it's worth, I've managed to spot check a few combinations (1-4-4 and 1-2-2 reads), and the waveforms looked fine.

antmarzam:
This can be closed since both DV and RTL changes are in master

PR #23174, opened 2024-05-17
Spi host ticket tb changes

antmarzam:
Takes care of #22182 :
- Checks RTL data signals stay stable during a whole cycle
- Added support for block-level TB to generate spi segments in random order


antmarzam:
The failure from the private CI is failing in the same way the existing failures from the regression fail:
`2024-05-17T15:34:55.7587397Z VERBOSE: [Launcher] UVM_FATAL @ 10005240574 ps: (csr_utils_pkg.sv:577) [csr_utils::csr_spinwait] timeout spi_host_reg_block.status.rxempty (addr=0x8a622454) == 0x0
`

hw/ip/spi_host/rtl/spi_host.sv
```diff
@@ -43,6 +43,75 @@ module spi_host
...
+  // 0    | 0    | 1       | 0       | -> XNOR(cpha, cpol): posedge
+  // 0    | 1    | 0       | 1       | -> NOR(cpha, cpol): negedge
+  // 1    | 0    | 0       | 1       | -> NOR(cpha, cpol): negedgesame
+  // 1    | 1    | 1       | 0       | -> XNOR(cpha, cpol): posedge
+  // ---------------------------------
+
+  reg                      posedge_enable, negedge_enable;
+  //HIGH when driven on pos-edge, sampled on the negedge
+  assign posedge_enable = (&cio_csb_o==0) & (configopts.cpha.q ^ configopts.cpol.q );
+  assign negedge_enable = (&cio_csb_o==0) & (configopts.cpha.q ^~ configopts.cpol.q);
```
rswarbrick, line 61: What does the chip-select part of this expression mean? I *think* we're only using the variables after a negedge of `csbs`, so isn't that part always true?

hw/ip/spi_host/rtl/spi_host.sv
```diff
@@ -43,6 +43,75 @@ module spi_host
...
+  // ---------------------------------
+  // CPHA | CPOL | posedge | negedge |
+  // ---------------------------------
+  // 0    | 0    | 1       | 0       | -> XNOR(cpha, cpol): posedge
+  // 0    | 1    | 0       | 1       | -> NOR(cpha, cpol): negedge
+  // 1    | 0    | 0       | 1       | -> NOR(cpha, cpol): negedgesame
+  // 1    | 1    | 1       | 0       | -> XNOR(cpha, cpol): posedge
+  // ---------------------------------
+
+  reg                      posedge_enable, negedge_enable;
```
rswarbrick, line 58: I think that Emacs has lined this up with line 41 above. We can probably lose some spaces here :-) (and similarly on the next few lines)

hw/ip/spi_host/rtl/spi_host.sv
```diff
@@ -43,6 +43,75 @@ module spi_host
...
+  //HIGH when driven on pos-edge, sampled on the negedge
```
rswarbrick, line 59: I think this comment documents `posedge_enable`, and `negedge_enable` is the other way around? Could you expand slightly to make it clearer what we're saying?

rswarbrick (commented):
I've taken a careful look at the first commit so far. To avoid touching the code in `spi_host/rtl`, would it be possible to bind the code instead? I think it might make the boundaries a bit more obvious?

hw/ip/spi_host/rtl/spi_host.sv
antmarzam: Sure - I added a comment to make it clearer, it's a bit confusing. 
Posedge_enable means: driven in pos-edge, sampled in neg-edge. 

I've reworded these signals to `sample_negedge_enable`, `sample_posedge_enable` respectively

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -72,25 +73,35 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
         join,
         @(negedge cfg.clk_rst_vif.rst_n or event_sw_rst)
       )
       `uvm_info(`gfn, "Restarting scoreboard checking due to reset event now.", UVM_LOW)
     end
   endtask : run_phase
 
+  virtual task get_plain_txn();
+    forever begin
+      plain_data_fifo.get(plain_item);
```
rswarbrick, line 86: The data flow is a bit hard to understand here. I *think* the point is that items get popped from `plain_data_fifo` here and then written to `plain_item`. Then, shortly later, `compare_tx_trans` will call `extract_data_for_segment`, which will use the data in `plain_item`.

Could we restructure things a bit? Maybe we could get rid of `get_plain_txn` entirely. Then pop the item from `plain_data_fifo` in `compare_tx_trans` and maybe pass it as an argument to `extract_data_far_segment`.

That will also let us get rid of the null check at the start of that task.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
       end
       `uvm_info(`gfn, $sformatf("\n successfully compared read transaction of %d ",
                                 tl_segment.command_reg.len+1), UVM_HIGH)
-      i++;
-    end
+      `uvm_info(`gfn, $sformatf("\n successfully compared read transaction of %d ",
```
rswarbrick, line 120: I think we've got a double copy of this message here.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+                                tl_segment.command_reg.len+1), UVM_LOW)
+    end // forever begin
   endtask
 
+  virtual task  extract_data_for_segment(spi_host_command_t  command_info,
+                                         output bit[7:0] host_data, output bit [7:0] device_data);
+    bit [4:0] bus_cycle;
+    int unsigned idx, loop_jump;
+    // segment.speed (mode)   : { standard: 0, dual: 1, quad: 2, reserved: 3).
+    // segment.direction: { dummy_cycles(None): 0, RX_only: 1, TX_only: 2, BiDir: 3)
```
rswarbrick, line 130: I think these comments probably aren't needed in the commit?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+
+    if(plain_item == null)
+      `uvm_fatal(`gfn, "Something's wrong: 'plain_item =  null'")
+
+    case(command_info.direction)
```
rswarbrick, line 135: Nit: Missing space after `case`

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+                                  , plain_item.sprint), UVM_DEBUG)
+        void'(plain_item.plain_data_q.pop_front());
+      end
+      default: begin
+        //RX/TX/Bidir
+        case(command_info.mode)
+          //Comparison against TX/RX fifo is carried out on a per-byte basis:
+          Standard: begin
+            wait(plain_item.plain_data_q.size >= 8);
+            loop_jump                         = 1;
```
rswarbrick, line 148: I'd probably suggest doing the unpacking into 8/4/2 and 1/2/4 in the case statement and then having just a single copy of the `wait` afterwards.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+          end
+          Dual: begin
+            wait(plain_item.plain_data_q.size >= 4);
+            loop_jump = 2;
+          end
+          Quad: begin
+            wait(plain_item.plain_data_q.size >= 2);
+            loop_jump = 4;
+          end
+          default: `uvm_fatal(`gfn, "Wrong Command.Speed = RSVD")
```
rswarbrick, line 158: I'd probably suggest using `command_info.mode.name()` here to avoid asserting that the only other option is RSVD (in case it changes!)

Similarly in the next `case` statement.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+        endcase // case (command_info.mode)
+
+        for(int i = 0; i < 8 ; i = i + loop_jump) begin
+          idx =  (command_info.direction != RxOnly ?
```
rswarbrick, line 162: I think this would probably be easier to read if you factored out a `bit bit_dir = ...` variable.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+            loop_jump = 2;
+          end
+          Quad: begin
+            wait(plain_item.plain_data_q.size >= 2);
+            loop_jump = 4;
+          end
+          default: `uvm_fatal(`gfn, "Wrong Command.Speed = RSVD")
+        endcase // case (command_info.mode)
+
+        for(int i = 0; i < 8 ; i = i + loop_jump) begin
```
rswarbrick, line 161: Nit: Missing space after `for` and stray space after "< 8".

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+          idx =  (command_info.direction != RxOnly ?
+                  cfg.m_spi_agent_cfg.host_bit_dir : cfg.m_spi_agent_cfg.device_bit_dir)
+                 ? i : (7 - i);
+          bus_cycle = plain_item.plain_data_q.pop_front();
+          case(command_info.mode)
```
rswarbrick, line 166: Nit: Missing space after `case`.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -106,17 +117,83 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+              host_data[idx]   = bus_cycle[0];
+              device_data[idx] = bus_cycle[1];
+            end
+            Dual: begin
+              host_data[idx -: 2] = bus_cycle[1:0];
+              device_data[idx -: 2] = bus_cycle[1:0];
+            end
+            Quad: begin
+              host_data[idx -: 4] = bus_cycle[4:0];
+              device_data[idx -: 4] = bus_cycle[4:0];
```
rswarbrick, line 177: This seems to be 5 bits: off by one error?

hw/ip/spi_host/dv/env/spi_transaction_item.sv
```diff
@@ -64,15 +65,27 @@ class spi_transaction_item extends uvm_sequence_item;
   // writeStd = 50/200, writeDual = 50/200 (overall write pct 50/100)
   // cmd_only = 2*5/200 (overall 5/100) etc.
   constraint cmd_c {
+    solve alt_weight before cmd;
+    alt_weight dist {
+       0        := 50,
+       [1:50]   := 20,
+       [51:100] := 30
+       };
```
rswarbrick, line 73: Nit: I think the usual style is to line the `}` up with `alt_weight` here. Similarly with the `cmd` constraint a few lines later.

hw/ip/spi_host/dv/env/spi_transaction_item.sv
```diff
@@ -64,15 +65,27 @@ class spi_transaction_item extends uvm_sequence_item;
...
+
+
+
```
rswarbrick, line 76: Stray newlines?

hw/ip/spi_host/dv/env/spi_transaction_item.sv
```diff
@@ -64,15 +65,27 @@ class spi_transaction_item extends uvm_sequence_item;
...
        ReadDual  := read_weight*dual_en,
        WriteDual := write_weight*dual_en,
        ReadQuad  := read_weight*quad_en,
-       WriteQuad := write_weight*quad_en };
+       WriteQuad := write_weight*quad_en,
+       AltCmd    := alt_weight
+              };
 
     read_weight + write_weight == 100;
+    //      alt_weight + read_weight + write_weight == 100;
```
rswarbrick, line 88: Isn't this supposed to be replacing the code on the previous line?

rswarbrick (commented):
This is just a review of the first commit so far. Even that might be easier to review (and get merged!) if it was split up a bit more. Next time, it might be worth splitting up the tidy-up changes from the new functionality.

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
+module spi_host_data_stable_sva
+  import spi_host_reg_pkg::*;
+  (input              rst_ni,
+   input             cio_sck_o,
+   input [NumCS-1:0] cio_csb_o,
+   input [3:0]       cio_sd_i,
+   input             spi_host_reg2hw_configopts_mreg_t configopts,
+   input             spi_device_pkg::passthrough_req_t passthrough_i
+   );
```
rswarbrick, line 9: Nit: This doesn't quite follow the standard indentation and I'd probably add `logic` types to everything that doesn't currently have them explicitly. I think you probably want:
```systemverilog
module spi_host_data_stable_sva
  import spi_host_reg_pkg::*;
(
  input logic              rst_ni,
  input logic              cio_sck_o,
  input logic [NumCS-1:0]  cio_csb_o,
  input logic [3:0]        cio_sd_i,

  input spi_host_reg2hw_configopts_mreg_t configopts,
  input spi_device_pkg::passthrough_req_t passthrough_i
);
```

I'd be inclined to avoid needing the import though. Maybe this would work too?
```systemverilog
module spi_host_data_stable_sva (
  input logic              rst_ni,
  input logic              cio_sck_o,
  input logic [NumCS-1:0]  cio_csb_o,
  input logic [3:0]        cio_sd_i,

  input spi_host_reg_pkg::spi_host_reg2hw_configopts_mreg_t configopts,
  input spi_device_pkg::passthrough_req_t                   passthrough_i
);
```

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+      end
+      POSEDGE_SAME_VALUE_CHECK_P: assert property (@(check_posedge) pos_value == neg_value) begin
+        `uvm_info("POSEDGE_SAME_VALUE_CHECK_P", $sformatf("[i=%0d] - ASSERTION PASSED %t",i, $time),
+                  UVM_DEBUG);
+      end
+      else begin
+        `uvm_error("POSEDGE_SAME_VALUE_CHECK_P", {$sformatf("[i=%0d] - ASSERTION FAILED",i),
+                                      $sformatf(" pos_value (0x%0x) != neg_value (0x%0x) - time=%t",
+                                                pos_value, neg_value, $time)})
+      end
```
rswarbrick, line 109: Hmm, it's a pity that the standard OT macros don't do what we want here. Maybe it would be worth defining something like `ASSERT_AT_EVENT` locally (by adapting a definition from `prim_assert_standard_macros.svh`).

I'd personally be keen on incorporating that into prim_assert, but a local version will split the "top-level discussion" from the simplification that we would get here.

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+  // 1    | 0    | 0       | 1       | -> XOR(cpha, cpol): drive posedge, sample negedge
+  // 1    | 1    | 1       | 0       | -> XNOR(cpha, cpol): drive negedge, sample posedge
+  // ---------------------------------
+
+
+  //Driven in pos-edge, sampled in neg-edge
+  reg                      sampled_negedge_enable;
+  //Driven in neg-edge, sampled in pos-edge
+  reg                      sampled_posedge_enable;
+  logic                    csbs;
```
rswarbrick, line 30: I think verilog-mode has followed the alignment of the macro definition above here :-) We can probably lose some spaces!

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+  assign csbs = &cio_csb_o;
+
+  //HIGH when driven on pos-edge, sampled on the negedge
+  assign sampled_negedge_enable = !csbs & (configopts.cpha.q ^ configopts.cpol.q);
+  assign sampled_posedge_enable = !csbs & (configopts.cpha.q ^~ configopts.cpol.q);
```
rswarbrick, line 35: I think I'd be inclined to write this as
```systemverilog
  assign sampled_negedge_enable = !csbs & (configopts.cpha.q ^ configopts.cpol.q);
  assign sampled_posedge_enable = !csbs & !sampled_negedge_enable;
```
to make the fact that both can't be true at once a bit more obvious.

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+
+  generate
```
rswarbrick, line 37: I don't think the `generate` is strictly necessary. Maybe drop it (and a level of indentation)?

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+    for(genvar i = 0; i < 4; i++) begin : g_cio_signal_stable_sva
```
rswarbrick, line 38: Nit: Missing space after "for"

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+  // 1    | 1    | 1       | 0       | -> XNOR(cpha, cpol): drive negedge, sample posedge
+  // ---------------------------------
+
+
+  //Driven in pos-edge, sampled in neg-edge
+  reg                      sampled_negedge_enable;
+  //Driven in neg-edge, sampled in pos-edge
+  reg                      sampled_posedge_enable;
+  logic                    csbs;
+  assign csbs = &cio_csb_o;
```
rswarbrick, line 31: I know that we only support a single CSB line in the design, but I wonder whether it would make sense to express it a bit more explicitly as `assign csb = cio_csb_o[0];`

I think weird stuff would happen if we had two lines and a negedge on `cio_csb_o[1]`, so maybe it's better to make it obvious that's not supported.

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+  assign sampled_negedge_enable = !csbs & (configopts.cpha.q ^ configopts.cpol.q);
+  assign sampled_posedge_enable = !csbs & (configopts.cpha.q ^~ configopts.cpol.q);
+
+  generate
+    for(genvar i = 0; i < 4; i++) begin : g_cio_signal_stable_sva
+      logic neg_value, pos_value;
+      event check_posedge, check_negedge;
+      initial begin
+        forever begin
+          @(negedge csbs);
```
rswarbrick, line 43: Would this work with `always @(negedge csbs)` ?

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+            neg_value <= cio_sd_i[i];
+          end
+        end
+      end
+
+      always @(negedge cio_sck_o or negedge rst_ni) begin
+        if(!rst_ni) begin
+          neg_value <= 0;
+        end
+        else if(!csbs) begin
```
rswarbrick, line 62: Nit: The convention is for `end else if` on one line here.

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,113 @@
...
+          end
+          if (sampled_negedge_enable) begin
+            @(posedge cio_sck_o);
+            neg_value <= cio_sd_i[i];
+          end
+        end
+      end
+
+      always @(negedge cio_sck_o or negedge rst_ni) begin
+        if(!rst_ni) begin
```
rswarbrick, line 59: Nit: Missing space after "if" (and in several other places. Probably worth doing a search for "if(")

rswarbrick (commented):
(Comments on commit 2)

hw/ip/spi_host/dv/sva/spi_host_sva.core
```diff
@@ -10,6 +10,8 @@ filesets:
       - lowrisc:tlul:headers
       - lowrisc:fpv:csr_assert_gen
     files:
+      - ../../rtl/spi_host_reg_pkg.sv
```
rswarbrick, line 13: This isn't quite what you want, I don't think. Probably the right approach is to add a dependency on `lowrisc:ip:spi_host` (the core that includes this file). But then we get the same thing as `files_formal`, so can get rid of that fileset as well.

rswarbrick (commented):
(Comment on commit 3)

hw/top_earlgrey/dv/env/chip_env.core
```diff
@@ -7,6 +7,7 @@ description: "CHIP DV UVM environmnt"
 filesets:
   files_dv:
     depend:
+      - lowrisc:dv:spi_host_sva
```
rswarbrick, line 10: Nit: I'd probably suggest moving this around so that stuff stays in alphabetical order.

hw/top_earlgrey/dv/chip_sim_cfg.hjson
```diff
@@ -24,7 +24,8 @@
   ral_spec: "{proj_root}/hw/top_earlgrey/data/top_earlgrey.hjson"
 
   // Add additional tops for simulation.
-  sim_tops: ["clkmgr_bind",
+  sim_tops: ["spi_host_bind",
```
rswarbrick, line 27: I'd probably move this down so that everything stays in alphabetical order.

rswarbrick (commented):
(Notes on commit 4)

rswarbrick (commented):
For the last commit, I think you should pull it out into a separate PR and get a proper designer (not me!) to take a proper look at it.

hw/dv/sv/spi_agent/spi_monitor.sv
```diff
@@ -48,12 +51,20 @@ class spi_monitor extends dv_base_monitor#(
...
-    host_item   = spi_item::type_id::create("host_item", this);
-    device_item = spi_item::type_id::create("device_item", this);
+    cfg.vif.sck_phase    = cfg.sck_phase[active_csb];
+    host_item            = spi_item::type_id::create("host_item", this);
+    device_item          = spi_item::type_id::create("device_item", this);
+    plain_item           = spi_item::type_id::create("plain_item",this);
+
+    plain_sampling_analysis_port.write(plain_item);
+    if(cfg.enable_plain_sampling)
+      csb_active_analysis_port.write(host_item);
```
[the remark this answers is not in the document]
antmarzam: They have different usages. 
For `plain_sampling_analysis_port`  This was needed for spi_host. This way, the monitor would just straight send what's in the bus, rather than interpreting the spi transaction segments in any given order.
The plain sampling AP just populates a data queue, which the spi_host SCB uses to check correctness.

For `csb_active_analysis_port`, this is used in spi_device, and it does sample in the "expected" segment order (opcode, [address, data]).
I'll add a comment for clarity

hw/dv/sv/spi_agent/spi_monitor.sv
```diff
@@ -104,6 +115,17 @@ class spi_monitor extends dv_base_monitor#(
     endcase // case (cfg.spi_func_mode)
   endtask : collect_trans
 
+  virtual protected task plain_sampling();
+    // for mode 1 and 3, get the leading edges out of the way
+    cfg.wait_sck_edge(LeadingEdge, active_csb);
+    forever begin : loop_forever
+      // wait for the sampling edge
+      cfg.wait_sck_edge(SamplingEdge, active_csb);
+      if(cfg.enable_plain_sampling)
```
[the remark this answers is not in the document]
antmarzam: I don't think that would work for cases where plain_sampling won't be used, because then that would kill the whole fork and kill other processes

hw/dv/sv/spi_agent/spi_monitor.sv
```diff
@@ -104,6 +115,17 @@ class spi_monitor extends dv_base_monitor#(
...
+        plain_item.plain_data_q.push_back(cfg.vif.sio[3:0]);
```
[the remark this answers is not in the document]
antmarzam: I'm not sure I follow, can you elaborate?
Currently, the spi_monitor packs data in different ways depending of the intended usage. For the spi_host case, we check that the segments sent into the block are the actual value written on the bus. For other cases we just sample the "golden spi transaction" that's expected to come in a given order.

I agree the agent can be refactored. Something along the lines like a layered agent where the bottom layer just deals with plain bus interface toggling and upper layers pack the data as needed would be helpful.

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
```diff
@@ -12,6 +12,9 @@
 class spi_host_env_cov extends cip_base_env_cov #(.CFG_T(spi_host_env_cfg));
   `uvm_component_utils(spi_host_env_cov)
 
+  spi_host_command_t dummy;
+  const int command_len_bits = $bits(dummy.len);
```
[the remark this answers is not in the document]
antmarzam: I've added a param to `spi_host_env_pkg.sv` where `spi_host_command_t` is defined, the field `len` in the struct is now derived from the parameter as well as the coverage bin

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
```diff
@@ -152,12 +155,58 @@ class spi_host_env_cov extends cip_base_env_cov #(.CFG_T(spi_host_env_cfg));
...
+      bins std_dual  = (Standard=>Dual);
+      bins std_quad  = (Standard=>Quad);
+
+      bins dual_dual = (Dual=>Dual);
+      bins dual_std  = (Dual=>Standard);
+      bins dual_quad = (Dual=>Quad);
+
+      bins quad_quad = (Quad=>Quad);
+      bins quad_std  = (Quad=>Standard);
+      bins quad_dual = (Quad=>Dual);
```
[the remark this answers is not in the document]
antmarzam: I don't think you can get transition FCOV without explicitly writing what transitions you're interested in.

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
```diff
@@ -152,12 +155,58 @@ class spi_host_env_cov extends cip_base_env_cov #(.CFG_T(spi_host_env_cfg));
...
+
+      bins Bi2None   = (Bidir => None);
+      bins Bi2Rx     = (Bidir => RxOnly);
+      bins Bi2Tx     = (Bidir => TxOnly);
+      bins Bi2Bi     = (Bidir => Bidir);
+    }
+    len_cp: coverpoint spi_cmd_reg.len{
+      bins zero = {0};
+      bins one = {1};
+      bins middle = {[2:2**command_len_bits - 1]};
```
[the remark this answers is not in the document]
antmarzam: I meant the middle range, values that aren't 0/1 or max. I renamed it to `middle_range_val`

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: That makes sense

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Ooops - Forget to remove. There was a `set_inst_override` in the JSON file, so the Scoreboard verbosity was silently being downgraded.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Ooops, I had declared the variable as `[4:0]` instead of `[3:0]` I fixed that too

hw/ip/spi_host/dv/env/spi_transaction_item.sv
antmarzam: Noted - I just let the editor (emacs) due the indentation on its own and thought was readable

hw/ip/spi_host/dv/env/spi_transaction_item.sv
antmarzam: That's what I thought doing at first, but the weights `read/write_weights` are taken directly from `spi_host_seq_cfg` rather than being randomized. I removed the line

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
antmarzam: I've added a comment indicating that currently we only support `NumCS=1`. 
I'm not sure we'd gain anything by hardcoding it to `cio_csb_o[0]` if we may support multiple lines in the future.

hw/ip/spi_host/rtl/spi_host.sv
rswarbrick: I think this question still stands: maybe the signals can be simplified slightly?

hw/dv/sv/spi_agent/spi_monitor.sv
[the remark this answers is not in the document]
rswarbrick: I think your comment looks sensible (and the discussion needn't hold up this PR).

My comment was about the fact that we write `plain_item` to `plain_sampling_analysis_port` at the start of a transaction and then push things into its `plain_data_q`  incrementally in the `plain_sampling` task here. I was wondering whether it would be possible to have a simple UVM item for each thing that gets sampled, instead of modifying a single UVM item through a "back door".

I wonder whether something that shape would make sense in a follow-up PR?

hw/ip/spi_host/rtl/spi_host.sv
antmarzam: This has been taken care in the latest is now written as:
```
assign sampled_negedge_enable = !csbs & (configopts.cpha.q ^ configopts.cpol.q);
  assign sampled_posedge_enable = !csbs & !sampled_negedge_enable;
```

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,109 @@
+module spi_host_data_stable_sva
+  (input logic                                               rst_ni,
+   input logic                                               cio_sck_o,
+   input logic [spi_host_reg_pkg::NumCS-1:0]                 cio_csb_o,
+   input logic [3:0]                                         cio_sd_i,
+   input spi_host_reg_pkg::spi_host_reg2hw_configopts_mreg_t configopts,
+   input spi_device_pkg::passthrough_req_t                   passthrough_i
+   );
```
rswarbrick, line 8: Sorry to come back to this (and maybe I said something silly last time). I think you probably want something like this:
```systemverilog
module spi_host_data_stable_sva (
  input logic                                               rst_ni,
  input logic                                               cio_sck_o,
  input logic [spi_host_reg_pkg::NumCS-1:0]                 cio_csb_o,
  input logic [3:0]                                         cio_sd_i,
  input spi_host_reg_pkg::spi_host_reg2hw_configopts_mreg_t configopts,
  input spi_device_pkg::passthrough_req_t                   passthrough_i
);
```

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,109 @@
...
+  // 0    | 0    | 1       | 0       | -> XNOR(cpha, cpol): drive negedge, sample posedge
+  // 0    | 1    | 0       | 1       | -> XOR(cpha, cpol): drive posedge, sample negedge
+  // 1    | 0    | 0       | 1       | -> XOR(cpha, cpol): drive posedge, sample negedge
+  // 1    | 1    | 1       | 0       | -> XNOR(cpha, cpol): drive negedge, sample posedge
+  // ---------------------------------
+
+
+  //Driven in pos-edge, sampled in neg-edge
+  reg    sampled_negedge_enable;
+  //Driven in neg-edge, sampled in pos-edge
```
rswarbrick, line 27: Nit: Missing space after "//"

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,109 @@
...
+  reg    sampled_posedge_enable;
+  logic  csbs;
+  // Current design support NumCS=1 only
+  assign csbs = &cio_csb_o;
+
+  //HIGH when driven on pos-edge, sampled on the negedge
```
rswarbrick, line 33: Nit: Missing space after "//"

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,109 @@
...
+  assign sampled_negedge_enable = !csbs & (configopts.cpha.q ^ configopts.cpol.q);
+  assign sampled_posedge_enable = !csbs & !sampled_negedge_enable;
+
+  for (genvar i = 0; i < 4; i++) begin : g_cio_signal_stable_sva
+    logic neg_value, pos_value;
+    event check_posedge, check_negedge;
+    initial begin
+      forever begin
+        @(negedge csbs);
+        //Initialising the neg/pos_value signals here, so they are ready for the first SPI cycle
```
rswarbrick, line 43: Nit: Missing space after "//"

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
```diff
@@ -0,0 +1,109 @@
...
+          pos_value <= cio_sd_i[i];
+        end
+        if (sampled_negedge_enable) begin
+          @(posedge cio_sck_o);
+          neg_value <= cio_sd_i[i];
+        end
+      end
+    end
+
+  always @(negedge cio_sck_o or negedge rst_ni) begin
```
rswarbrick, line 57: I think the indentation needs fixing here, adding two spaces to every line down to the final "end"

hw/ip/spi_host/rtl/spi_host.sv
rswarbrick: The thing I'm confused about is that we only ever read these variables when `!csbs` is true. So I'm not really convinced that the term is still needed, allowing the signal to be simplified. But it definitely doesn't matter either way :-)

hw/dv/sv/spi_agent/spi_monitor.sv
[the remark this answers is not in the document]
rswarbrick: I think we're currently sending a queue of things through `plain_item_q` and then waiting for those items to appear before sampling them. This is... exactly what UVM items are normally used for! So I'm suggesting that we could change `plain_sampling_analysis_port` to pass those <= 4-bit items instead.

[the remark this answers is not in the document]
antmarzam: I've tried this and the testbench starts falling over which probably means the scoreboard would need a refactor. 
We can do this in the future. I personally think it's probably best to leave it as is, since the `spi_item` will be populated for each time period the CSB is low. VS sending just the bus value with no reference as to where in the CSB it occurred.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -105,18 +119,84 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+    int unsigned idx, loop_jump;
+
+    if(plain_item == null)
+      `uvm_fatal(`gfn, "Something's wrong: 'plain_item =  null'")
+
+    case (command_info.direction)
+      None: begin
+        wait(plain_item.plain_data_q.size >= 1);
+        `uvm_info(`gfn, $sformatf("Dummy_cycle: Popping 1 item from 'plain_item.plain_data_q': \n%s"
+                                  , plain_item.sprint), UVM_DEBUG)
```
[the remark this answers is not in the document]
antmarzam: I'm limited due to the line length - it'd be nice to have longer lines in verification

hw/ip/spi_host/dv/env/spi_transaction_item.sv
```diff
@@ -19,6 +19,7 @@ class spi_transaction_item extends uvm_sequence_item;
   // NOTICE read_weight + write_weight <= 100
   int  read_weight                    = 45;
   int  write_weight                   = 50;
+  rand int  alt_weight                = 50;
```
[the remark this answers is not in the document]
antmarzam: The variables declared above and below do have 2 spaces after the variable type though

hw/ip/spi_host/dv/env/spi_transaction_item.sv
```diff
@@ -64,13 +65,21 @@ class spi_transaction_item extends uvm_sequence_item;
   // writeStd = 50/200, writeDual = 50/200 (overall write pct 50/100)
   // cmd_only = 2*5/200 (overall 5/100) etc.
   constraint cmd_c {
+    solve alt_weight before cmd;
+    alt_weight dist {
+       0        := 50,
+       [1:50]   := 20,
+       [51:100] := 30
+                    };
```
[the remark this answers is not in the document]
antmarzam: The alt_weight constraint has the same style as the cmd constraint. Curly braces are open after `dist`, and closing curly brace is aligned with the open brace

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
antmarzam: But I don't want to have 1-bin that covers a list all transitions possible, I want 1-bin per possible transition. PErhaps I misunderstood your suggestion. I did taake a look at 19.5.2 in the LRM but didn't find any auto_bins for transition coverage

hw/dv/sv/spi_agent/spi_monitor.sv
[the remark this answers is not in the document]
rswarbrick: Oh, I see. Yep, definitely fine to do this for now :-)

antmarzam:
I regressed the SVA check using VCS against the RTL in @a-will  PR #23579 and the assertion is working with no race conditions. The only failing test was the passthrough one which isn't related to the RTL change and there's already a ticket for it to get fixed ([passthrough failing test in VCS](https://github.com/lowRISC/opentitan/issues/23593))

hw/dv/sv/spi_agent/spi_agent_cfg.sv
```diff
@@ -21,6 +21,9 @@ class spi_agent_cfg extends dv_base_agent_cfg;
   bit decode_commands;  // Used in monitor if decoding of commands needed
   bit [2:0] cmd_addr_size = 4; // Address size for command
 
+  bit       enable_plain_sampling; // When set to 1 the monitor will sample the bus values directly
+                                   // without any configuration other than the leading edge
+
```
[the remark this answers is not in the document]
rswarbrick: I'm happy to leave this unchanged (and now see that other nearby fields are the same). But next time please could you try to respond or make the change instead of silently marking the conversation as resolved?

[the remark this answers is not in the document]
antmarzam: Oops - sorry didn't mean to ignore it. I thought I responded. The reason to leave unchanged was that there are plenty of variables which aren't initialised because the initial value is 0. Happy to explicitly change it to `=0` though but I'm not sure we gain anything

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
rswarbrick: I don't think that's quite right. I've just looked again at IEEE 1800 and I think we just need `[]`! On page 562, there's this example:
image
and then
image

I think you just need the equivalent of the `sb` example? So something like:
```systemverilog
    direction_transition_cp: coverpoint spi_cmd_reg.direction{
      bins Any2Any = ([None:Bidir] => [None:Bidir]);
    }
```
(untested) might work?

This needn't block this PR, but it might be nice to get straight.

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
rswarbrick: Ah, but I think `2**command_len_bits - 1` *is* the maximum value. Should it be `-2` at the top?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
rswarbrick: You've resolved this comment but I think we've still got the duplicated `wait` line. Definitely fine with leaving it the same, but it would be good to get some response.

hw/ip/spi_host/dv/env/spi_transaction_item.sv
rswarbrick: Hmm. I think that `verilog-mode` doesn't necessarily agree with our indentation opinions :-) Feel free to patch it... (It would make my life easier too!)

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
rswarbrick: I think this question is still outstanding. The final commit does a rename of the signal, but I still think we can get rid of the `forever begin wait_for_an_event; do_something; end` structure.

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
```diff
@@ -152,12 +152,58 @@ class spi_host_env_cov extends cip_base_env_cov #(.CFG_T(spi_host_env_cfg));
...
+    len_cp: coverpoint spi_cmd_reg.len{
+      bins zero = {0};
+      bins one = {1};
+      bins middle_range_val = {[2:2**SPI_HOST_COMMAND_LEN_SIZE_BITS - 1]};
+      bins max_val = { 2**SPI_HOST_COMMAND_LEN_SIZE_BITS - 1 };
+    }
+    //XCOV
+    speedXdirection_cp: cross speed_cp, direction_cp, csaat_cp;
+
+  endgroup // segment_speed_cg
```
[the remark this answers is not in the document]
rswarbrick: This still applies, I think.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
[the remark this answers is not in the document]
rswarbrick: Probably more obvious:
```systemverilog
        `uvm_info(`gfn,
                  $sformatf("Dummy_cycle: Popping 1 item from 'plain_item.plain_data_q':\n%s",
                            plain_item.sprint),
                  UVM_DEBUG)
```
(and I'm also dropping a trailing space from the literal)

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -105,18 +119,84 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+        void'(plain_item.plain_data_q.pop_front());
+      end
+      default: begin
+        //RX/TX/Bidir
+        case(command_info.mode)
+          //Comparison against TX/RX fifo is carried out on a per-byte basis:
```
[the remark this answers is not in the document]
rswarbrick: There is still a missing space after "case"

hw/ip/spi_host/dv/env/spi_transaction_item.sv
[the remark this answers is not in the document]
rswarbrick: Hmm, you're right. I have no idea why the original author did that! Eh, a tidy-up for another day :-)

hw/ip/spi_host/dv/env/spi_transaction_item.sv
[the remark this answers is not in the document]
rswarbrick: Are you following a style guide for this approach? It's pretty new to me (but maybe dictated by Emacs's `verilog-mode` auto-completion which is not particularly brilliant for this sort of thing).

If there's not a prevailing style like this, I suggest not starting one.

hw/dv/sv/spi_agent/spi_agent_cfg.sv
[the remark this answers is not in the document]
antmarzam: @rswarbrick - I'm having second thoughts. I don't see the point of adding `=0` for a variable that's already implicitly set to 0.  I've been looking at the guidelines, and I've not found that we recommend initialising variables to 0.

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
antmarzam: I've added this: 
```
    speed_trans_cp : coverpoint spi_cmd_reg.mode {
      // Creates a bin for each of the possible transitions
      bins Any2Any[] = ([None:Bidir] => [None:Bidir]);
    }

```
And also checked the names for the bins are still readable

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
antmarzam: Good spot - changing it

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
antmarzam: Do you mean the closing comment? I'll push that change

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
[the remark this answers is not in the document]
antmarzam: Is this in the guidelines? I'm not sure what we gain here any other than more lines for a printout that can take just 2 lines?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Sorry it's been a while and I can't remember why I resolved this comment - I thought I addressed the change. 
I'm not sure what you're asking here. I don't think we gain anything - we'd need to define a new variable to set to 8/4/2 to then move the wait further down before the `for` loop.
Maybe it's cos I wrote the code but it seems more readable as it is than by adding more complexity to it

hw/ip/spi_host/dv/sva/spi_host_data_stable_sva.sv
antmarzam: done!

hw/dv/sv/spi_agent/spi_agent_cfg.sv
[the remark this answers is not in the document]
rswarbrick: Ok, no strong feelings. Let's leave it :-)

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
rswarbrick: Great, thanks. And I'm pleased to hear that the bins come out looking reasonable! :-)

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
rswarbrick: I'm really sorry, but I think you want the -2 at the top of middle and just -1 in max_val.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
rswarbrick: I think something like this would be a bit simpler:
```systemverilog
      default: begin
        int unsigned bits_per_plain_item;

        // RX/TX/Bidir
        case (command_info.mode)
          Standard: bits_per_plain_item = 1;
          Dual:     bits_per_plain_item = 2;
          Quad:     bits_per_plain_item = 4;
          default: `uvm_fatal(`gfn, $sformatf("Wrong command.speed: %s",command_info.mode.name()))
        endcase
        wait(plain_item.plain_data_q.size >= 8 / bits_per_plain_item);

        for (int i = 0; i < 8; i = i + bits_per_plain_item) begin
          bit bit_dir = (command_info.direction != RxOnly ?
                         cfg.m_spi_agent_cfg.host_bit_dir : cfg.m_spi_agent_cfg.device_bit_dir);
          idx = bit_dir ? i : (7 - i);

          bus_cycle = plain_item.plain_data_q.pop_front();
          case (command_info.mode)
            Standard: begin
              host_data[idx]   = bus_cycle[0];
              device_data[idx] = bus_cycle[1];
            end
            Dual: begin
              host_data[idx -: 2] = bus_cycle[1:0];
              device_data[idx -: 2] = bus_cycle[1:0];
            end
            Quad: begin
              host_data[idx -: 4] = bus_cycle[3:0];
              device_data[idx -: 4] = bus_cycle[3:0];
            end
            default: `uvm_fatal(`gfn, "Command.mode = Reserved")
          endcase
        end
```

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -126,40 +206,37 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
       for (int i=0; i < exp_segment.command_reg.len+1; i++) begin
-        host_data_fifo.get(dut_item);
-        device_data_fifo.get(device_item);
+
+        wait(&cfg.m_spi_agent_cfg.vif.csb == 0);
+        // After the sampling edge the plain_item should've been populated
+        cfg.m_spi_agent_cfg.wait_sck_edge(SamplingEdge, cfg.m_spi_agent_cfg.vif.get_active_csb());
+
+        // Extract data based on segment: task below blocks if data has not yet been fully sampled
+        extract_data_for_segment(exp_segment.command_reg, host_data, device_data);
```
rswarbrick, line 215: Looking at the body of `extract_data_for_segment`, I think this is writing the same values to both `host_data` and `device_data`. If that's correct, maybe it would be easier to read the code if we changed this to something more like:
```systemverilog
      for (int i=0; i < exp_segment.command_reg.len+1; i++) begin
        bit [7:0] segment_data;

        wait(&cfg.m_spi_agent_cfg.vif.csb == 0);
        // After the sampling edge the plain_item should've been populated
        cfg.m_spi_agent_cfg.wait_sck_edge(SamplingEdge, cfg.m_spi_agent_cfg.vif.get_active_csb());

        // Extract data based on segment: task below blocks if data has not yet been fully sampled
        extract_data_for_segment(exp_segment.command_reg, segment_data);

        // process tx part of the transaction
        if (exp_segment.command_reg.direction inside {TxOnly, Bidir}) begin
          if (segment_data != exp_segment.spi_data[i]) begin
            txt = {txt, $sformatf("\n \t [%d] \t\t\t      %0h  \t\t\t %0h",
                                  i, segment_data, exp_segment.spi_data[i])};
            `uvm_fatal(`gfn,
                       $sformatf("\n\t WRITE: actual data did not match exp data \n len %d %s",
                                 exp_segment.command_reg.len+1, txt))
          end else begin
            txt = {txt, $sformatf("\n \t [%d] \t\t\t %0h \t\t\t %0h",
                                  i, segment_data, exp_segment.spi_data[i])};
          end
        end

        // process rx part of transaction
        // sorting step - will drop everything that is not stored in rx fifo
        if (exp_segment.command_reg.direction inside {RxOnly, Bidir}) begin
          rx_data_q.push_front(segment_data);
        end
      end // for (int i=0; i < exp_segment.command_reg.len+1; i++)
```
and `extract_data_for_segment` can be simplified to something like
```systemverilog
  virtual task  extract_data_for_segment(spi_host_command_t command_info, output bit[7:0] data);
    bit [3:0] bus_cycle;
    int unsigned idx;

    if (plain_item == null)
      `uvm_fatal(`gfn, "Something's wrong: 'plain_item =  null'")

    case (command_info.direction)
      None: begin
        wait(plain_item.plain_data_q.size >= 1);
        `uvm_info(`gfn, $sformatf("Dummy_cycle: Popping 1 item from 'plain_item.plain_data_q': \n%s"
                                  , plain_item.sprint), UVM_DEBUG)
        void'(plain_item.plain_data_q.pop_front());
      end
      default: begin
        int unsigned bits_per_plain_item;

        // RX/TX/Bidir
        case (command_info.mode)
          Standard: bits_per_plain_item = 1;
          Dual:     bits_per_plain_item = 2;
          Quad:     bits_per_plain_item = 4;
          default: `uvm_fatal(`gfn, $sformatf("Wrong command.speed: %s",command_info.mode.name()))
        endcase
        wait(plain_item.plain_data_q.size >= 8 / bits_per_plain_item);

        for (int i = 0; i < 8; i = i + bits_per_plain_item) begin
          bit bit_dir = (command_info.direction != RxOnly ?
                         cfg.m_spi_agent_cfg.host_bit_dir : cfg.m_spi_agent_cfg.device_bit_dir);
          idx = bit_dir ? i : (7 - i);

          bus_cycle = plain_item.plain_data_q.pop_front();
          case (command_info.mode)
            Standard: data[idx]      = bus_cycle[0];
            Dual:     data[idx -: 2] = bus_cycle[1:0];
            Quad:     data[idx -: 4] = bus_cycle[3:0];
            default: `uvm_fatal(`gfn, "Command.mode = Reserved")
          endcase
        end

        `uvm_info(`gfn, $sformatf("[%s] - Extracted byte-> 0x%0x",
                                  command_info.mode.name, data), UVM_DEBUG)
      end
    endcase

  endtask
```

hw/ip/spi_host/dv/env/spi_transaction_item.sv
rswarbrick: I definitely do not want to become the "style police" :-) If we don't have explicit guidelines committed somewhere then I guess my taste is probably not relevant! Let's keep this.

hw/ip/spi_host/dv/env/spi_host_env_cov.sv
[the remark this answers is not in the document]
rswarbrick: That's correct (and sorry for not being more explicit). Thanks.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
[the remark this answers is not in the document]
rswarbrick: Well, I am not going to approve something where the indentation is flat-out wrong (the leading comma here). If we don't have an explicit guideline forbidding that syntax, maybe another reviewer will be ok with it.

My proposed code was just a way to make things as clear as possible. But you're right: the results were a bit mad! Maybe something like this would avoid the long line and also avoid adding more newlines:
```systemverilog
        `uvm_info(`gfn, $sformatf("Dummy_cycle: Popped item from 'plain_item.plain_data_q': \n%s",
                                  plain_item.sprint), UVM_DEBUG)
```

hw/ip/spi_host/dv/env/spi_transaction_item.sv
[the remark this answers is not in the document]
rswarbrick: I think that code where the body of a `{...}` block is "outdented" from the braces looks quite strange to me. I don't think I've ever seen that style recommended in any language. But I might be wrong! In which case, maybe the right approach is to bring it up with some other engineers and generate a joint proposal (either convincing or overruling me :-D)

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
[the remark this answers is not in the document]
antmarzam: We're back at comment#1 - I can't do that because of the line-length limit. 

Hence, I tried to have some code which reads fine. rather than a one line per uvm_info parameter.

I'm just going to copy paste your 4-line suggestion just so we can get over this.

I want you two point me to where it says what I wrote is flat-out wrong though. 

As I can't find it in the guidelines: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#spacing
I've attached a screenshot of the guidelines below: 
Screenshot from 2024-07-01 09-39-41

hw/ip/spi_host/dv/env/spi_transaction_item.sv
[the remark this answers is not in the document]
antmarzam: Again, I will just copy paste your suggestion so we can get over this. 

But I want you to please point me to where in the guidelines indicate what I did is wrong. 
I can't find it anywhere, and if it's not forbidden anywhere I don't quite get how that's a reason not to approve it.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: I added this suggestion, only that I reused the variable `loop_jump` rather than defining a new one.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: This is not correct for Standard mode, only for Dual and Quad modes.

rswarbrick:
Thanks for taking the changes. Would you mind squashing them into the commits to avoid a trailing "Rupert's changes" commit?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
rswarbrick: Ok, I'm really not understanding something (sorry). As far as I can tell, `extract_data_for_segment` writes exactly the same things to both output arguments. So presumably the difference in standard mode is only visible in `compare_tx_trans`? As far as I can see, the two variables are used interchangeably there too. What am I missing?

antmarzam: It's not the same for both output arguments, you can see in `Std` mode:
https://github.com/lowRISC/opentitan/blob/c4c8fc345e7a3384c25c0dfc1bc3723afe0ade68/hw/ip/spi_host/dv/env/spi_host_scoreboard.sv#L162-L163

antmarzam:
> Thanks for taking the changes. Would you mind squashing them into the commits to avoid a trailing "Rupert's changes" commit?

Sure - I'll squash them once we're in agreement with the PR :)

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
rswarbrick: Ohh. I hadn't noticed that 0 != 1! :-D You're completely right, sorry for the repeated questions.

rswarbrick:
Thanks. I've just squashed and force-pushed (there should be an empty list of changes).

rswarbrick:
I think the only remaining discussion is about style. Thank you very much for putting up with all my nitty questions/comments.

rswarbrick (approved):

PR #23176, opened 2024-05-17
[pinmux,fpv] Fix check in MioAttrO_A property

rswarbrick:
This got broken when we added the `input_disable` field in f541cd92997. Tweak the masking to match what the design is doing.


hw/ip/pinmux/fpv/vip/pinmux_assert_fpv.sv
```diff
@@ -296,7 +296,7 @@ module pinmux_assert_fpv
   assign mio_pad_attr_mask.keep_en = 0;
   assign mio_pad_attr_mask.schmitt_en = 0;
   assign mio_pad_attr_mask.od_en = 0;
-  assign mio_pad_attr_mask.input_disable = 0;
+  assign mio_pad_attr_mask.input_disable = 1'b1;
```
andreaskurth, line 299: I agree that this is necessary because all pinmux-controlled pads now support input disabling; thx for this fix @rswarbrick.

The same two changes should be applied to the respective lines for DIOs below, right?

rswarbrick: You're completely right, but I hadn't got to the assertions that those broke :-)  I'll tweak to match.

marnovandermaas (approved):

PR #23167, opened 2024-05-17
[sival, rv_core_ibex] check for dbus address translation

nbdd0121:
Previously this test tests for ibus translation only despite setting up both ibus and dbus translation. This PR adds a dbus check.

PR #23178, opened 2024-05-17
[rtl,rv_core_ibex] Mask off lower bits of mapping address

GregAC:
Previously to this change software had to set the lower bits of the mapping address to 0 otherwise the address translation would produce an incorrect result.


GregAC:
Fixes https://github.com/lowRISC/opentitan/issues/22884

GregAC:
I've tested this with the updated rv_core_ibex_address_translation_test here: https://github.com/lowRISC/opentitan/pull/23167 with the addition of adding some junk into the lower bits (that gets masked off with this fix). I'll update that test with the junk bits once both PRs are merged.

cfrantz (approved):

marnovandermaas (approved):
Looks reasonable to me.

andreaskurth (approved):

andreaskurth:
@GregAC: I think the failing CI check is unrelated to the changes in this PR. Shall we go ahead and merge this?

PR #23182, opened 2024-05-17
[rtl,rv_core_ibex] Mask off lower bits of mapping address

GregAC:
Previously to this change software had to set the lower bits of the mapping address to 0 otherwise the address translation would produce an incorrect result.

Same fix as https://github.com/lowRISC/opentitan/pull/23178 but on integrated branch


Razer6 (approved):

andreaskurth (approved):

vogelpi:
CHANGE AUTHORIZED: hw/ip/rv_core_ibex/rtl/rv_core_addr_trans.sv

vogelpi (approved):

PR #23186, opened 2024-05-17
[i2c,dv] Target mode tx stretch ctrl test

hcallahan-lowrisc:
> Only final 3 commits are new, of which only the final 2 are meaningful. The previous commits come from #23142, upon which this PR is based, and will be merged before this.

See the testpoint and vseq header for more detailed descriptions, but in short
this feature (TX_STRETCH_CTRL) causes target-mode read transfers to halt and await software
intervention. The directed test checks for some descriptive components of the
testpoint, and also that the interrupt behaviour is in-line with the expectation
over the course of each transfer.


hcallahan-lowrisc:
CW310 CI failures are unrelated to this DV-only PR.

antmarzam (approved):
LGTM! :D

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -718,6 +718,7 @@ class i2c_base_vseq extends cip_base_vseq #(
     bit [7:0] wdata_q[$];
     i2c_item  txn;
     bit       rs_avl;
+    int       last_sr;
```
rswarbrick, line 721: Nit: Maybe worth explicitly zero-initialising this? (Otherwise, it's very noticeable that we read the variable before the first time it is written, so I think we're currently relying on the automatic zero initial value)

hw/ip/i2c/dv/env/seq_lib/i2c_target_tx_stretch_ctrl_vseq.sv
```diff
@@ -0,0 +1,165 @@
...
+// with data. It then awaits the tx_strech interrupt to be asserted, and checks
+// that 'TARGET_EVENTS.TX_PENDING' is set. It then clears the pending state to
+// allow the transfer to complete, and upon completion checks the pending
+// interrupt has not been asserted again.
+// We also check that once TX_PENDING is set, SCL remains low until the bit is
+// cleared.
+//
+//
+// (DUT/Agent == Target/Controller)
+// +use_intr_handler=1 -> Interrupt Driven
```
rswarbrick, line 17: Could you expand on these notes a bit? For example, is the second line saying that this vseq needs to be run with `use_intr_handler=1`? Or does the line tell me how to use this vseq in an interrupt driven context?

hw/ip/i2c/dv/env/seq_lib/i2c_target_tx_stretch_ctrl_vseq.sv
```diff
@@ -0,0 +1,165 @@
...
+              `uvm_fatal(`gfn, "SCL was released while TX_PENDING was set!")
+            end
+            begin
+              csr_spinwait(.ptr(ral.target_events.tx_pending), .exp_data(1'b0), .backdoor(1));
+            end
+          join_any
+          disable fork;
+        end: iso_fork join
+      end
+    join_any
```
rswarbrick, line 83: This looks slightly surprising to me. Do we really want to leave some background processes running? I'd probably expect some sort of `join` or `disable fork`?

hw/ip/i2c/dv/env/seq_lib/i2c_target_tx_stretch_ctrl_vseq.sv
```diff
@@ -0,0 +1,165 @@
...
+
+  virtual task start_of_stim_hook();
+    // Write all rdata for the next read transaction to the TXFIFO
+    while (read_rcvd.size() > 0) write_tx_fifo();
+  endtask: start_of_stim_hook
+
+  virtual task end_of_stim_hook();
+    // Allow the interrupt handlers to wrap-up before the next
+    // round of stimulus.
+    #10us;
```
rswarbrick, line 103: This looks reasonable, but is there anything we can wait for to avoid putting a magic time in the sequence?

rswarbrick (approved):
Nitty questions, but this looks very sensible to me.

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
hcallahan-lowrisc: Good idea. Done.

hw/ip/i2c/dv/env/seq_lib/i2c_target_tx_stretch_ctrl_vseq.sv
hcallahan-lowrisc: Sure, added a comment and a disable fork to kill the helper processes once super.body() has concluded.

hw/ip/i2c/dv/env/seq_lib/i2c_target_tx_stretch_ctrl_vseq.sv
hcallahan-lowrisc: I had a think about it, and couldn't immediately come up with something reliable. Hopefully in the future we will improve the reference model to become more accurate, and then it can do the checking rather than this directed test. Then this sort of arbitrary delay can be removed.

PR #23198, opened 2024-05-20
[pattgen] Add inactive level feature

### LINKED ISSUE #21924: [pattgen] Feature request to avoid glitches
### Description

ChromeOS intends to use the patterngen IP to temporarily take over control of two I2C wires, in order to generate a special waveform that puts certain ITE chips into bootloader mode.

Ideally, the switch back and forth between I2C IP controlling the wires and patterngen would be without glitches.  I have noticed two undesirable behaviors:

1) When reconfiguring the pinmux to let patterngen control the wires, before starting the patterngen clock, the signals will be driven down, rather than left at the default high level of an I2C bus.
2) When the pattern has run the requested number of iterations, the patterngen IP freezes in a state of driving the output according to the first sample of the repeating pattern, (when the channel enable bit is cleared, the outputs are then driven low.)

The second issue I could work around, by modifying the pattern so that the first sample is at a high level, and by reconfiguring pinmux before clearing the channel enable bit.  The first issue, though, is bound to generate a glitch on the wires.

My proposal is this:
1) Add additional bits to the CTRL register, such that while the pattern generator channel is not enabled, these new control bits will determine the "steady state" output on both clock and data. Four bits in total: PDA_LEVEL_CH0, PCL_LEVEL_CH0, PDA_LEVEL_CH1, PCL_LEVEL_CH1.  (If it proves difficult to support the new bits for clock signal, we could live with having only new steady state bits for the data signal, that is two new bits.  If so, ChromeOS would need to use the data signal of both channels, rather than data and clock signal of one channel.)
2) Make it such that upon completing the requested number of repetitions, the output will immediately go to the configured "steady state" level, rather than the first sample of the repeating pattern.  If this is most easily achieved by letting the hardware clear the ENABLE_CHn bit in the CTRL register, that is fine.

andreaskurth:
This feature allows to define the level that `pcl` and `pda` take when pattgen is not actively sending data  bits (i.e., when pattgen is disabled or all data bits have been sent).

The feature is not yet supported in the scoreboard, so the scoreboard gets disabled for the test that exercises this feature.  This limitation is tracked in a linked issue.  A manual inspection of waves showed that the main use case works as intended, though:
image
(see https://github.com/lowRISC/opentitan/issues/21924#issuecomment-2120534793 for details)

All existing tests retain their checks, coverage, and pass rate; i.e., when this feature is not used (`inactive_level_p`{`cl`,`da`}` = 0`), pattgen's behavior does not change.

This resolves #21924.

## TODOs (pre-merge)
- [x] Confirm with integration teams that this fulfills their requirements.
- [x] Document use case that led to this feature request in pattgen's Programmer's Guide.
- [x] Create issue to model the new feature in the scoreboard and track its coverage, and link that issue in the `TODO` comment in the code. --> #23219

### DIFF
--- a/hw/ip/pattgen/data/pattgen.hjson
+++ b/hw/ip/pattgen/data/pattgen.hjson
@@ -83,6 +83,9 @@
     { name: "PATTGEN.CONFIG.POLARITY",
       desc: "Each channel's clock can be inverted."
     },
+    { name: "PATTGEN.CONFIG.INACTIVE_LEVEL",
+      desc: "The inactive level of clock and data output of each channel can be configured separately."
+    }
     { name: "PATTGEN.COMPLETE",
       desc: "Once a pattern is finished, an interrupt is raised."
     }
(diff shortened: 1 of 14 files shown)


jesultra (approved):
Based on screenshots in issue #21924, I can confirm that this PR meets the requirements from the Ti50 team.

andreaskurth:
Great, thanks for your feedback! I'll finalize this PR and get it merged.

andreaskurth:
The failure of _CW310 SiVal ROM_EXT Tests_ seems unrelated to this PR; all other CI checks passed. Merging

PR #23201, opened 2024-05-20
[top_earlgrey,padring] Implement muxing for IOB[0:3]

andreaskurth:
In packages with reduced pin count, not all pads can be bonded out.  The
direct-mapped pads have to be bonded out, but Quad SPI may not always be
required.  For use cases where IOB[0:3] are not bonded out but
SPI_HOST_D{2,3} and SPI_DEV_D{2,3} are bonded out but are not used
(because Quad SPI is not used), this commit adds muxes that allow using
those Quad SPI ports via IOB[0:3].

The mux selectors are currently tied off (to the direct connection);
they will be wired up as part of https://github.com/lowRISC/opentitan/issues/23280.

### DIFF
--- a/hw/top_earlgrey/rtl/autogen/chip_earlgrey_asic.sv
+++ b/hw/top_earlgrey/rtl/autogen/chip_earlgrey_asic.sv
@@ -477,6 +477,7 @@ module chip_earlgrey_asic #(
   // This is only used for scan and DFT purposes
     .clk_scan_i   ( ast_base_clks.clk_sys ),
     .scanmode_i   ( scanmode              ),
+    .mux_iob_sel_i ('0), // TODO(#23280)
     .dio_in_raw_o ( dio_in_raw ),
     // Chip IOs
     .dio_pad_io ({
--- a/hw/top_earlgrey/rtl/padring.sv
+++ b/hw/top_earlgrey/rtl/padring.sv
@@ -27,6 +27,8 @@ module padring
   // This is only used for scan
   input                           clk_scan_i,
   prim_mubi_pkg::mubi4_t          scanmode_i,
+  // Mux selectors for IOB[0-3]
+  input logic               [3:0] mux_iob_sel_i,
   // RAW outputs used for DFT and infrastructure
   // purposes (e.g. external muxed clock)
   output logic     [NDioPads-1:0] dio_in_raw_o,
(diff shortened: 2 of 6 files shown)


hw/top_earlgrey/rtl/autogen/chip_earlgrey_asic.sv
```diff
@@ -477,6 +477,7 @@ module chip_earlgrey_asic #(
   // This is only used for scan and DFT purposes
     .clk_scan_i   ( ast_base_clks.clk_sys ),
     .scanmode_i   ( scanmode              ),
+    .mux_iob_sel_i ('0), // TODO
```
andreaskurth, line 480: This will need to be connected to CSRs inside ast or sensor_ctrl

hw/top_earlgrey/rtl/padring.sv
```diff
@@ -56,6 +58,36 @@ module padring
...
+          // Connect output towards muxed pad IOB3 to pad SPI_DEV_D3.
+          dio_out = mio_out_i[top_earlgrey_pkg::MioPadIob3];
+          dio_oe  = mio_oe_i[top_earlgrey_pkg::MioPadIob3];
+        end
+        default: begin
+          // Directly connect output signals to pad.
+          dio_out = dio_out_i[k];
+          dio_oe  = dio_oe_i[k];
+        end
+      endcase
```
andreaskurth, line 89: To be confirmed that this code is suitable or if it needs to be
- at a different location
- implemented with a muxing prim

hw/top_earlgrey/rtl/padring.sv
```diff
@@ -72,13 +104,43 @@ module padring
...
+          // Connect input coming from pad SPI_DEV_D3 to signals of muxed pad IOB3.
+          mio_in_o[k]     = dio_in_o[top_earlgrey_pkg::DioSpiDeviceSd3];
+          mio_in_raw_o[k] = dio_in_raw_o[top_earlgrey_pkg::DioSpiDeviceSd3];
+        end
+        default: begin
+          // Directly input signals to pad.
+          mio_in_o[k]     = mio_in;
+          mio_in_raw_o[k] = mio_in_raw;
+        end
+      endcase
```
andreaskurth, line 142: Same muxing code questions as above

hw/top_earlgrey/rtl/padring.sv
a-will: Recommend a muxing prim so the named hierarchy is accessible for any physical or timing constraints.

Probably a prim_clock_mux2 would work fine, I'd think.

andreaskurth: Thanks for the recommendation! I added a second commit that implements all muxes with `prim_clock_mux2`; see [this example](https://github.com/lowRISC/opentitan/pull/23201/files#r1607207957). If we go for that option, I'll squash the second commit into the first; if not, I'll drop it.

hw/top_earlgrey/rtl/padring.sv
```diff
@@ -56,6 +58,90 @@ module padring
...
+        .clk_o  (dio_out)
+      );
+      prim_clock_mux2 #(
+        .NoFpgaBufG(1'b1)
+      ) u_mux_dio_oe (
+        .clk0_i (dio_oe_i[k]),
+        .clk1_i (mio_oe_i[top_earlgrey_pkg::MioPadIob0]),
+        .sel_i  (mux_iob_sel_i[0]),
+        .clk_o  (dio_oe)
+      );
```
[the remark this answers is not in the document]
a-will: Looks good to me! (at least as good as this sort of exception can look...)

hw/top_earlgrey/rtl/padring.sv
andreaskurth: Received feedback from silicon team; mux prims are needed. Squashed the second commit into the first

hw/top_earlgrey/rtl/padring.sv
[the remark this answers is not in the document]
andreaskurth: Looks good to the PD team as well

andreaskurth:
Approved by PD team on 2024-05-23

vogelpi (approved):

PR #23202, opened 2024-05-20
[opentitantool] Factor JSON configuration files

jesultra:
Third-party developers, such as Chrome OS, who use Teacup boards for development and testing, will want to use how pins of HyperDebug are routed to pins of the OpenTitan chip as declared in hyperdebug_teacup.json.  However, depending on their test suites, they may want HyperDebug to drive e.g. IOA4 as a UART, as GPIO with some default level, or not drive in the "default" state prior to test execution.  Even within OpenTitan repository, one could imagine separate test suites wanting different default state.

In order to accomodate such differences, while still allowing re-use of the lengthy list of how e.g. IOA4 is connected to CN8_10 on the board (which will necessarily apply to all users of the Teacup board), this PR factors `hyperdebug_teacup.json` into two files.  All users will continue to use `hyperdebug_teacup.json`, which declares the signal routing.  Tests in OpenTitan repo will additionally use the new `hyperdebug_teacup_default.json` which sets up defaults on particular pins.  Other users will combine the former with their own configuration files.

### DIFF
--- a/sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
+++ b/sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
@@ -1,3 +1,20 @@
+/*
+ * This file declares the signal routing of the Teacup board.  This file is
+ * meant to be used with multiple iterations of the silicon chip (possibly
+ * requiring separate chip-specific JSON configuration files to declare where,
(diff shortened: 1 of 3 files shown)


jettr (approved):
This separation looks like the way to go!

cfrantz (approved):
LGTM

timothytrippel (approved):

PR #23205, opened 2024-05-20
[rescue] Support Slot B and explicit reboots

cfrantz:
1. Support rescue uploads to slot B.
2. Add a `WAIT` command that prevents automatic reboot after upload actions.

### DIFF
--- a/sw/device/silicon_creator/rom_ext/doc/rescue.md
+++ b/sw/device/silicon_creator/rom_ext/doc/rescue.md
@@ -43,7 +43,7 @@ Alternate modes are requested by sending the mode's 4-byte code followed by a ne
 
 The following sections detail each of the supported alternate modes.
 
-#### Firmware Rescue (`RESQ`)
+#### Firmware Rescue (`RESQ`, `RESB`)
 
 The firmware rescue mode may be requested with the 4-byte code `RESQ`.
 The ROM_EXT will acknowledge entry of this mode with the following message:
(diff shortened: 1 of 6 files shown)


sw/device/silicon_creator/rom_ext/doc/rescue.md
```diff
@@ -79,7 +82,7 @@ ok: send boot_svc request via xmodem-crc
 ```
 
 The ROM_EXT will then will prompt for the transfer to start by sending the Xmodem-CRC start character (which is the ASCII character `C`).
-After completing this action, the ROM_EXT will switch back to firmware rescue mode.
```
jesultra, line 82: Was this previous comment accurate?  If so, why are you reversing the default behavior?  (As opposed to keeping the convention that after firmware transfer, the chip goes back to rescue mode, with a "RESET"/"BOOT" command that can be sent afterwards.)

jesultra (approved):

sw/device/silicon_creator/rom_ext/doc/rescue.md
```diff
@@ -124,6 +127,18 @@ ok: reboot
...
+
+The user may request that the ROM_EXT disable the automatic reboot after upload actions with the 4-byte code `WAIT`.
+The ROM_EXT will acknowledge this request with the following message:
+
+```
+mode: WAIT
+ok: wait after upload
+```
+
+Once commanded to `WAIT`, the ROM_EXT will need an explicit reboot after an upload action.
```
jesultra, line 140: Does this apply only to the very next upload action or all uploads until next reset?  (It would be a strange use case for sending firmware twice in a single rescue session.)

sw/device/silicon_creator/rom_ext/doc/rescue.md
jesultra: I see.  I find it a little cleaner to have RESQ never reboot, and a separate "BOOT" command, which would by default be issued by opentitantool.  But this maintains backwards compatibility, and gets the job done.

sw/device/silicon_creator/rom_ext/doc/rescue.md
cfrantz: It disables automatic reboot after upload actions until the next reset.  A mutli-upload scenario is possible during an ownership transfer operation:

- Upload next_owner config.
- Upload next_owner firmware.
- Upload boot_svc OwnershipActivate command
- Reboot

cfrantz:
The commits in this PR were submitted with #23208.

PR #23206, opened 2024-05-20
[pattgen/rtl] Flop outputs to prevent glitches

andreaskurth:
Prior to this commit, both `pcl_o` and `pda_o` were driven from combinational logic.  In case of `pcl_o`, that logic was one mux (using `polarity_q` as selector) that switched between the `pcl_int_q` and its inverse.  In case of `pda_o`, that logic was a mux (using `!enable`, where `enable` comes from a CSR field) that selected between 0 and `data_q[bit_cnt_q]`.  Especially the latter term can be problematic, because during the `[bit_cnt_q]` indexing the different bits of `bit_cnt_q` can have different propagation delays to the mux structure that selects one bit from `data_q`, which has a significant probability of causing glitches on `pda_o`.

This commit prevents glitches by flopping both signals before outputting them.

### DIFF
--- a/hw/ip/pattgen/rtl/pattgen_chan.sv
+++ b/hw/ip/pattgen/rtl/pattgen_chan.sv
@@ -58,11 +58,21 @@ module pattgen_chan
     end
   end
 
+  // Drive pcl_o and pda_o directly from FF, so that they don't glitch.
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      pcl_o <= 1'b0;
+      pda_o <= 1'b0;
+    end else begin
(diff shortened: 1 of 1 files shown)


andreaskurth:
@pamaury: Could you please try this on FPGA? It should fix the glitch problem that you were observing on pattgen's PDA output.

andreaskurth:
CC @jesultra: In case you ever observed glitches while pattgen was active (i.e., not related to switching muxes in pinmux), this could be a likely reason. @pamaury did not observe such glitches in silicon AFAIK but on the FPGA.

pamaury:
I can confirm that I cannot reproduce the glitch with this change. I have run it on the problematic configuration and also many more random ones without any problem.

pamaury (approved):

Issue #21353, opened 2024-02-13
[rom,rom_ext] During owner firmware selection/verification handle flash ECC errors as non-fatal


jettr:
### Description

When ROM_EXT is deciding and verifying which slot of owner firmware should execute, any flash ECC error while reading the potential owner firmware image should be treated similar to a signature failure, i.e. the next slot should be considered. ECC flash errors during the read operation should not reset the chip in this case. Hopefully we can accomplish this with our current hardware.

moidx:
Hi @msfschaffner, marking this for PROD.M2 so that we can triage. The idea here is that during secure boot, we want to be able to avoid triggering a fatal escalation when reading the image to calculate its hash. We rather handle the alert, treat the impacted code partition as corrupted, and try to boot from the next partition. We want to consider this change to avoid getting into a potential boot failure loop.

This under the assumption that the probability of flash corruption with a hamming distance of >2 bits is non-zero; and, the corruption is due to environmental issues and not active attacks.

This change may also be applicable to partitions used to store data.

CC: @zi-v

zi-v:
I think that's a good idea. BTW, I guess you mean hamming distance > 1 bits, right?

msfschaffner:
I believe the current hardware implementation should support this already. 

--------------------------------------------------

First some background:

The `flash_ctrl` has two flavors of ECC:
1) reliablity ECC on 64bit blocks, added after encrypting two 32bit words with PRINCE. 
2) integrity ECC on 32bit words. this is part of the bus end-to-end integrity scheme. 

Both ECC flavors are enabled together on region/page boundaries with the `ECC_EN` bit. The first flavor is reliability ECC that is truly SECDED in the sence that 1bit errors are corrected. 2bit errors lead to an in-band access error, and trigger a fatal alert.

The second flavor is for detection only (no SECDED), although it is not truly end-to-end as in SRAM_CTRL for example, since the block size of the cipher and the amount of meta data bits available per 64bit block do not allow us to encrypt and store a full 7bit ECC tag alongside the each 32bit data word. Therefore, it is stripped away and a 4b checksum is calculated over 2x32bit words before encryption. If those bits are mismatched upon readout, any mismatch will trigger a fatal alert and lead to an in-band access error. 

[...]

PR #23207, opened 2024-05-20
[rom,e2e] add tests for flash ECC error handling during boot

timothytrippel:
This adds ROM E2E two test cases that corrupt the ROM_EXT manifest identifier in both slots A and B to prepare for the development of a ROM load access fault exception handler to address #21353. Note this just contains the test cases to tease out the flash ECC corruption error cases, but does not yet provide a solution for handling such in the ROM. A solution to handling the ECC error corruption (i.e., continue attempting to boot the next stage) will be developed in a follow up PR.

### DIFF
--- a/sw/device/silicon_creator/rom/e2e/boot_policy_flash_ecc_error/flash_ecc_error_test.c
+++ b/sw/device/silicon_creator/rom/e2e/boot_policy_flash_ecc_error/flash_ecc_error_test.c
@@ -30,6 +30,14 @@ static dif_flash_ctrl_region_properties_t kFlashFullAccessScrambledEcc = {
     .prog_en = kMultiBitBool4True,
     .rd_en = kMultiBitBool4True};
 
+/**
+ * ROM_EXT slots.
+ */
+typedef enum rom_ext_slot {
+  kRomExtSlotA = 0,
+  kRomExtSlotB = 1,
+} rom_ext_slot_t;
+
 enum {
   // Manifest Offsets; See `sw/device/silicon_creator/lib/manifest.h`.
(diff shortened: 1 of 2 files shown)


sw/device/silicon_creator/rom/e2e/boot_policy_flash_ecc_error/flash_ecc_error_test.c
```diff
@@ -0,0 +1,140 @@
...
+ * ROM_EXT slots.
+ */
+typedef enum rom_ext_slot {
+  kRomExtSlotA = 0,
+  kRomExtSlotB = 1,
+} rom_ext_slot_t;
+
+enum {
+  // Manifest Offsets; See `sw/device/silicon_creator/lib/manifest.h`.
+  kManifestIdOffest = 820,
```
moidx, line 43: nit: Consider adding an assertion to ensure this offset is always aligned.

moidx (approved):

sw/device/silicon_creator/rom/e2e/boot_policy_flash_ecc_error/flash_ecc_error_test.c
timothytrippel: Ah yeah, one exists here: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/manifest.h;drc=30b710a40e6a68786e6436e285ab8294104374ca;l=342

But I will add a duplicate here in my follow up PR that adds the example prototype and test enhancements so it is more obvious what code needs to be updated if this ever changes.

Issue #22943, opened 2024-05-02
[rom_ext] Feature request: Serial rescue at higher baud rates


jesultra:
### Description

Serial rescue at 115.2 kbaud is significantly slower than SPI bootstrapping, which can typically run at least at 8MHz.  This may impact ChromeOS, as our plan is to disable SPI bootstrapping on production chips.  It would be nice to have a way of doing faster serial rescue.

Running the console UART at 115.2 bkaud at boot is highly desirable, we do not want to change that.  However, perhaps the ROM_EXT rescue mode could be enhanced with a command to temporarily switch baud rate.

In order synchronize the switch on both OpenTitan and the debugger, I propose that a new "set baudrate" command would take an integer, and then still at the "old" baud rate acknowledge the reception of the command, and that the requested baud rate is supported.  Then in order to for the new baud rate to take effect, the debugger will generate a break condition, to which ROM_EXT responds with some serial prompt/banner at the new baud rate.  This protocol allows the host computer to abort if the switch is not possible, and gives it time to reconfigure the debugger to the new baud rate, before any prompt is sent with the new rate.

Also, a minor feature request for ROM_EXT would be that it prints some version number in the initial "banner" on the serial console, such that the host computer may know from the version what features are supported by the firmware.

jesultra:
Further thoughts here:  Extra robustness in the baud rate switch could possibly be achieved, if we made sure that the first character transmitted in the ROM_EXT banner (both when coming out of reset, and after switching baud rate) would be a "U" (0x55).  That particular value will result in five equidistant square pulses on the UART TX line, which the STM32 EC is able to use for automatic baud rate detection.
We could even say that at any time when ROM_EXT is active, a break condition should result in resetting the ROM_EXT state machine and printing a banner starting with 0x55, as means of recovering from lost protocol synchronization.  (This could also be achieved already by break + reset, though.)

cfrantz:
I think we can probably accommodate some kind of line-rate switching, but I don't want to add any variable-length arbitrary input handling to the code.  The current rescue modes are all handled by recognizing [FourCC](https://en.wikipedia.org/wiki/FourCC) type codes.  Mode switches are always <4 bytes><newline>.

To handle baud-rate switching, I'd like there to be a command (like `BAUD`) followed by an identifier that specifies a particular supported rate.  Offhand, I'd suggest the following:
- `115K`: 115200 bps
- `230K`: 230400 bps
- `460K`: 460800 bps
[...]

PR #23208, opened 2024-05-20
[rescue] Support faster UART speeds 

cfrantz:
Allow the rescue protocol to negotiate a speed faster than the standard 115200 bps.  The faster rates are the "extended standard" rates of 230400, 460800 and 921600 bps.

This addresses #22943.

This PR depends on #23205; you need only review the last 2 commits.

### DIFF
--- a/sw/device/lib/arch/device.h
+++ b/sw/device/lib/arch/device.h
@@ -123,15 +123,31 @@ extern const uint64_t kUartBaudrate;
  * the NCO width, use NCO = 0xffff for this case since the error is tolerable.
  * Refer to #4263
  */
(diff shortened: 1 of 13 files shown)


jesultra (approved):
This feature looks good to me, I look forward to seeing the improved rescue time on actual devices.  (Once this has gone into a ROM_EXT signed for engineering silicon.)

I wonder how fast we could go.  For higher speeds we should probably choose values more divisor-friendly to the 24MHz peripheral clock, such as 1.5Mbaud, 2.0Mbaud, 2.4Mbaud or even 3Mbaud.

cfrantz:
The CI test failures are unrelated to this change.

PR #23211, opened 2024-05-21
[cryptolib] Eliminate `status_create` symbol use

cfrantz:
The cryptolib avoid use of the general-purpose `status_t` creating macros like `OK_STATUS()` or `INVALID_ARGUMENT()` which call `status_create()`.  Instead, cryptolib defines its own status types that are compatible with `status_t` but do not invoke `status_create`.

A couple of the general-pupose invocations snuck in.  Eliminate them in favor of `OTCRYPTO_...` definitions.


jadephilipoom (approved):
Thanks, Chris!

moidx (approved):

PR #23214, opened 2024-05-21
Port `chip_sw_sleep_pin_mio_dio_val` to SiVal

luismarques:


sw/host/tests/chip/gpio/src/sleep_pin_mio_dio_val.rs
```diff
@@ -0,0 +1,252 @@
...
+use std::time::Duration;
+
+use opentitanlib::app::TransportWrapper;
+use opentitanlib::execute_test;
+use opentitanlib::io::gpio::PinMode;
+use opentitanlib::io::gpio::PullMode;
+use opentitanlib::test_utils::init::InitializeTest;
+use opentitanlib::uart::console::UartConsole;
+
+const MIO_PADS: &[&str] = &[
```
jwnrt, line 16: Please could you extract these to `sw/host/opentitanlib/src/test_utils`? They will be useful for a test I'm writing.

sw/host/tests/chip/gpio/src/sleep_pin_mio_dio_val.rs
```diff
@@ -0,0 +1,252 @@
...
+    "IOR6",  // MIO Pad 41
+    "IOR7",  // MIO Pad 42
+    "IOR10", // MIO Pad 43
+    "IOR11", // MIO Pad 44
+    "IOR12", // MIO Pad 45
+    "IOR13", // MIO Pad 46
+];
+
+// IOR6/7, IOR10-13
+const VALID_MIO_PADS_FPGA_CW310: &[bool] = &[
```
jwnrt, line 67: I would be a little worried about these arrays getting out of sync. What do you think of laying them out in structs? e.g.

```rust
struct Pad {
    name: &'static str,
    valid_cw310: bool,
    valid_silicon: bool,
}

const PADS: &[Pad] = &[
    Pad { name: "IOA0", valid_cw310: false, valid_silicon: true },
    Pad { name: "IOA1", valid_cw310: false, valid_silicon: true },
    // ...
];
```

sw/host/tests/chip/gpio/src/sleep_pin_mio_dio_val.rs
```diff
@@ -0,0 +1,252 @@
...
+    for _ in 0..get_valid_mio_pads(interface).len() {
+        let msg = UartConsole::wait_for(&*uart, r"MIO \[(\d+)\]: (\d+)", opts.timeout)?;
+        let index: usize = msg[1].parse()?;
+        expected_pin_values[index] = msg[2].parse()?;
+    }
+
+    UartConsole::wait_for(&*uart, r"Going to sleep", opts.timeout)?;
+
+    for (index, valid) in get_valid_mio_pads(interface).iter().enumerate() {
+        if *valid {
```
jwnrt, line 201: Unsure if this is actually clearer or not, so feel free to ignore

```suggestion
    for index in get_valid_mio_pads(interface)
        .iter()
        .enumerate()
        .filter_map(|(i, valid)| valid.then_some(i))
    {
```

jwnrt (approved):
LGTM, but would be nice to extract those valid pin arrays

sw/host/tests/chip/gpio/src/sleep_pin_mio_dio_val.rs
pamaury: I think the better solution would be to have a proper description of the pins somewhere in the tree (maybe a json file?) that opentitanlib can read. It can be more subtle than just valid or not, some pins have pull-up/down or are connected to peripheral which can make them unusuable for some tests but maybe not all.

jwnrt: Good point, is that what the `sw/host/opentitanlib/src/app/config/` JSON files are intended to be used for?

pamaury: Those files serve a different purpose at the moment but it feels like the right place to add more information indeed.

luismarques: After @a-will's suggestion [here](https://github.com/lowRISC/opentitan/pull/23275#issuecomment-2128307590) to go with a device tree, for now I've only incorporated @jwnrt's suggestion of the `struct Pad`. That should resolve this discussion, but please check if it.

jwnrt:
@luismarques Please could you manually port this to `master` (as above) and include https://github.com/lowRISC/opentitan/pull/23416 in the same PR? Thanks

PR #23216, opened 2024-05-21
[pinmux,rtl] Tweak pinmux.hjson to match RTL

rswarbrick:
The RTL code was copied from Earlgrey with commit 66dd1d2 and the hjson file no longer matched. I think the original intention with this bit of the hw/ip vs. hw/top_earlgrey divide was that the stuff in hw/ip/pinmux was a sort of template from which we derived versions for the different tops.

There's a reasonably long-term effort to switch to generating this block with ipgen (tracked as issue 8440), and doing so will realise that dream! In the meantime, we've got some tests that we want to do for Earlgrey that are currently run from the code in hw/ip/pinmux.

Tweak the numbers in the hjson so that these tests are testing the only chip we are currently building.

### DIFF
--- a/hw/ip/pinmux/data/pinmux.hjson
+++ b/hw/ip/pinmux/data/pinmux.hjson
@@ -234,25 +234,25 @@
     { name: "AttrDw",
       desc: "Pad attribute data width",
       type: "int",
-      default: "10",
+      default: "13",
       local: "true"
     },
     { name: "NMioPeriphIn",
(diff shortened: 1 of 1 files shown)


andreaskurth (approved):
LGTM, thanks for updating `hw/ip/pinmux/data/pinmux.hjson` to match the RTL I copied over from the top_earlgrey RTL on pinmux

PR #23217, opened 2024-05-21
[dv, chip]: Change tap_strap_value type

walaahattab:
Modifying tap_stap_value type.

Currently, the tap_strap_value check fails even though the tap_strap has reached the desired value, to fix this the type is changed to bit[1:0].

### DIFF
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_sw_base_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_sw_base_vseq.sv
@@ -763,7 +763,7 @@ class chip_sw_base_vseq extends chip_base_vseq;
       fork
         cfg.clk_rst_vif.wait_clks(max_cycles);
         forever begin
-          uvm_hdl_data_t tap_strap_value;
+          bit[1:0] tap_strap_value;
           string tap_strap_path = {"tb.dut.top_earlgrey.u_pinmux_aon.",
                                    "u_pinmux_strap_sampling.tap_strap"};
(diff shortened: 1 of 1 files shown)


walaahattab:
CC: @sha-ron

rswarbrick (changes requested):
I don't think this is the correct fix. `uvm_hdl_read` takes an output argument of type `uvm_hdl_data_t` (see [here](https://verificationacademy.com/verification-methodology-reference/uvm/docs_1.1c/html/files/dpi/uvm_hdl-svh.html#uvm_hdl_read)).

It's probably worth adding some `$display` or `uvm_info` statements to print out the strap values that you are actually seeing. My suspicion is that they are `'x`: but we don't want to conclude that things are connected before the signals get a sensible value.

rswarbrick:
Thanks for the explanation. It's quite surprising to me: it sounds like `uvm_hdl_read` is not touching the upper bits of its output argument, which was being initialised to `'x` by the simulator beforehand. I can't find that behaviour of `uvm_hdl_read` documented in a reference manual.

Can you point me at a test I can run to see it happen?

sha-ron:
I see also https://github.com/lowRISC/opentitan/issues/23257 that looks related.

sha-ron:
@rswarbrick can you please have a look again?

eshapira (approved):

rswarbrick (approved):
Thanks for helping with the debugging. I'll take a look at what's going on now, but this definitely looks like the right approach if it's fixing things.

PR #23225, opened 2024-05-21
[top_earlgrey/dv] Create regression set for nightly Xcelium runs

andreaskurth:
Some top-level tests depend on Xcelium to run correctly and have all checks active.  As top-level tests for Earlgrey run in VCS by default, all tests in the `nightly` regression set also run in VCS.  This commit creates a new `xcelium_nightly` regression set so that specific tests can be run in Xcelium during nightly regressions.


hcallahan-lowrisc (approved):

andreaskurth:
DV checks have passed; the remaining CI checks are not relevant for this DV-only change. Skipping them also to free up CI resources

PR #23227, opened 2024-05-21
[fpv] Tweak report headers to match Jasper version

rswarbrick:
Looking at documentation for the current version of Jasper, and also for Jasper version 2021.12, the reported numbers are not called stimuli/coi/proof any more. Tweak to match the names used in recent versions.

The most recent nightly run gave the following result:
image

This makes rather more sense if the "green column" is stimuli coverage (which essentially checks whether all the inputs of all the modules in the design can be 0/1).


marnovandermaas (approved):
Looks reasonable to me, although I haven't tested it.

PR #23228, opened 2024-05-21
[hmac, doc] Update HMAC context saving guide

ballifatih:
Preivously, it was suggested that KEY registers should be copied by SW as it saves the context. However KEY registers are swaccess=wo type, which means reading them returns 0. Therefore SW *must* keep a copy of the key itself, without relying on this register.


andreaskurth (approved):
Agreed, thx @ballifatih

andreaskurth:
Skipping remaining CI checks for this doc-only change (lint checks have already passed)

gdessouky:
LGTM thanks @ballifatih

Issue #22916, opened 2024-05-01
[hmac, rtl] Expose `Idle` status of HMAC


ballifatih:
### Description

Related to https://github.com/lowRISC/opentitan/issues/22731.

I am working on HMAC driver, and I realized that SW does not have a good visibility on what is going on in HMAC IP. In particular, I was not able to figure out:

- how to check if HMAC is still active,
- when active, if there is a way to discard the ongoing operation.

At the driver level, we usually have a function to check the availability of the IP before continuing with the operation (just to be safe). For instance, we have the following function, which is supposed to discard whatever is going on in HWIP, so that we can run HMAC with the current request:

https://github.com/lowRISC/opentitan/blob/e9925bdf1464e9cec8ab8931c108cee4acb2709b/sw/device/lib/crypto/drivers/hmac.c#L18-L40

Would it be possible to have `Idle` bit in `STATUS` register, as we have in [KMAC](https://opentitan.org/book/hw/ip/kmac/doc/registers.html#status--sha3_idle)? Also, what is the correct way to flush the ongoing operation in HMAC? As far as I see, deasserting `sha_en` as above does not work because CFG register is gated with internal FSM signals.

Not having visibility is a bit of problem, because with streaming, we are potentially looking at two HMAC calls that interleave each other, and we need to know if HMAC is occupied.

cc: @gdessouky @andreaskurth @martin-velay @jadephilipoom

martin-velay:
Hi Fatih,

What you are looking for might be the [hmac_done](https://opentitan.org/book/hw/ip/hmac/doc/registers.html#intr_state--hmac_done) interrupt. I have checked with a simulation and in the RTL: [hmac_core.sv](https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/hmac/rtl/hmac_core.sv;l=426-431)
Is that enough for your application?

The Idle is also transmitted as an output signal of the HMAC, I think for power gating purpose but I don't really know if this is made visible to the SW at any other stage as such.

BR,
Martin

ballifatih:
Hi @martin-velay. Yes I am aware of `hmac_done` interrupt. Although this is useful, it does not quite address what I need.

[...]

PR #23233, opened 2024-05-21
[hmac/rtl] Fix idle_o when told to stop

### LINKED ISSUE #23036: [hmac/spec] Output pin doesn't go to Idle in case of HMAC stop invoked
### Description

Seen in simulation while testing the Save and Restore feature, the output pin `idle_o` is not moving into idle state after a hash_stop has been triggered.

andreaskurth:
Prior to this commit, `hmac_core` and thus the entire `hmac` module would not raise the `idle_o` output when it got told to stop (#23036). This is problematic because then HMAC cannot be clock-gated during context switch (save & restore feature).  Clock gating would only work when a message stream is fully processed (after the *process* command).

This commit fixes the problem by keeping track of idleness inside `hmac_core`.  In addition to the conditions that existed before, this tracker goes to idle when a *stop* command is received and the current message block is completely processed.

This fixes #23036.

### DIFF
--- a/hw/ip/hmac/rtl/hmac.sv
+++ b/hw/ip/hmac/rtl/hmac.sv
@@ -660,6 +660,7 @@ module hmac
     .key_length_i  (key_length),
 
     .reg_hash_start_i    (hash_start),
+    .reg_hash_stop_i     (reg_hash_stop),
     .reg_hash_continue_i (hash_continue),
     .reg_hash_process_i  (packer_flush_done), // Trigger after all msg written
     .hash_done_o         (reg_hash_done),
--- a/hw/ip/hmac/rtl/hmac_core.sv
+++ b/hw/ip/hmac/rtl/hmac_core.sv
@@ -14,6 +14,7 @@ module hmac_core import prim_sha2_pkg::*; (
   input key_length_e  key_length_i,
 
   input        reg_hash_start_i,
+  input        reg_hash_stop_i,
   input        reg_hash_continue_i,
   input        reg_hash_process_i,
   output logic hash_done_o,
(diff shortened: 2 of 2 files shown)


ballifatih:
LGTM, but probably best to wait for @gdessouky's approval. I am not super familiar with the internal logic.

This is also the Idle signal that could be reused for SW I think. https://github.com/lowRISC/opentitan/issues/22916

andreaskurth:
> This is also the Idle signal that could be reused for SW I think.

Good idea; I'll mention it in #22916

martin-velay (approved):
LGTM, need to re-enable test as the piece of code from the SCB has been commented out with this tag `#23036`

gdessouky:
Thanks, this LGTM, if it doesn't break existing DV. I have a lot of PRs waiting to merged that touch DV with fixes.

gdessouky (approved):

andreaskurth:
The CW310 CI check failures seem to be unrelated to this PR; the other CI checks pass. Merging

gdessouky:
> > This is also the Idle signal that could be reused for SW I think.
> 
> Good idea; I'll mention it in #22916

I think we should actually be using the `idle_o` from `hmac.sv` (see https://github.com/lowRISC/opentitan/pull/23242/commits) not `hmac_core.sv `that was modified here.

andreaskurth:
> > > This is also the Idle signal that could be reused for SW I think.
> > 
> > 
> > Good idea; I'll mention it in #22916
> 
> I think we should actually be using the `idle_o` from `hmac.sv` (see https://github.com/lowRISC/opentitan/pull/23242/commits) not `hmac_core.sv `that was modified here.

Indeed :+1:

PR #23239, opened 2024-05-21
[dv,usdbev] usbdev_aon_wake sequences

alees24:
This PR introduces a single sequence that addresses the following test points:

- usbdev_aon_wake_resume
- usbdev_aon_wake_reset
- usbdev_aon_wake_disconnect

These are checking that the `usbdev_aon_wake` module responds to the appropriate newly-introduced USB bus events from the usb20_driver, and that it issues a 'wake request' in response. The request would ordinarily be received by power management hardware and cause the SoC to resume from Normal/Deep Sleep.

The PR also makes some minimal modifications to `usb20_monitor` to correct its decoding and handling of valid packet data on the USB, since that was found to be leading to failures in the new tests (as well as a number of the existing sequences). It does not attempt to fix up the monitor behavior for invalid traffic such as that presently being generated by the 'usbdev_bitstuff_err' sequence; that will be addressed in a later PR.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_driver.sv
+++ b/hw/dv/sv/usb20_agent/usb20_driver.sv
@@ -8,7 +8,14 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   `uvm_component_new
 
   // These delays are in terms of bit intervals
-  int usb_rst_time = 100_000;  // upto 10ms
+  //
+  // TODO: introduce variation in the duration of the signaling, and permit these intervals to be
+  // shortened for those sequences not specifically designed to test with real-world timings;
+  // it's just wasting simulation time for many of the sequences since the DUT must respond long
+  // before the protracted Reset/Resume Signaling concludes.
+
+  int usb_rst_time = 120_000;  // Bit intervals @ 12Mbps => 10ms
+  int usb_resume_time = 240_000;
   int usb_idle_clk_cycles = 5;
   int usb_pwr_good_clk_cycles = 100;  // arbitrary short delay post-VBUS assertion.
 
@@ -40,12 +47,42 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
         seq_item_port.get_next_item(req_item);
         $cast(rsp_item, req_item.clone());
         rsp_item.set_id_info(req_item);
-        case (req_item.m_pkt_type)
-          PktTypeToken:     prepare_token_packet(req_item, rsp_item);
-          PktTypeData:      prepare_data_packet(req_item, rsp_item);
-          PktTypeHandshake: prepare_handshake_packet(req_item, rsp_item);
-          PktTypeSoF:       prepare_sof_packet(req_item, rsp_item);
-          default: `uvm_fatal(`gfn, $sformatf("Pkt type %x not supported", req_item.m_pkt_type))
+        unique case (req_item.m_ev_type)
+          // Send Bus Reset to the DUT.
+          EvBusReset: begin
+            // No response required, and this is a lengthy operation; Reset Signaling is supposed
+            // to take >= 10ms under the USB 2.0 Protocol Specification.
+            seq_item_port.item_done(rsp_item);
+            bus_reset();
+          end
+          // Resume Signaling (exit from Suspended).
+          EvResume: begin
+            // No response required, and this is a lengthy operation; this must be at least 20ms
+            // for real USB hosts.
+            seq_item_port.item_done(rsp_item);
+            resume_signaling();
+          end
+          // Assert VBUS (connection to USB host).
+          EvConnect: begin
+            cfg.bif.drive_vbus = 1'b1;
+            seq_item_port.item_done(rsp_item);
+          end
+          // Deassert VBUS (disconnection from USB host).
+          EvDisconnect: begin
+            cfg.bif.drive_vbus = 1'b0;
+            seq_item_port.item_done(rsp_item);
+          end
+          // Transmit packet (and maybe receive a response).
+          EvPacket: begin
+            case (req_item.m_pkt_type)
+              PktTypeToken:     prepare_token_packet(req_item, rsp_item);
+              PktTypeData:      prepare_data_packet(req_item, rsp_item);
+              PktTypeHandshake: prepare_handshake_packet(req_item, rsp_item);
+              PktTypeSoF:       prepare_sof_packet(req_item, rsp_item);
+              default: `uvm_fatal(`gfn, $sformatf("Pkt type %x not supported", req_item.m_pkt_type))
+            endcase
+          end
+          default: `uvm_fatal(`gfn, "Invalid/unsupported USB event type")
         endcase
       end
     end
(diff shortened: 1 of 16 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
```diff
@@ -0,0 +1,159 @@
...
+  constraint stim_delay_clks_c {
+    stim_delay_clks >= -50 && stim_delay_clks <= 2000;
+  };
+
+  // Delay before optional reconnection of VBUS after disconnection.
+  rand int disconn_interval_clks;
+  constraint disconn_interval_clks_c {
+    // In practice the DUT will have capacitance on the VBUS line to stabilize it and to ignore
+    // transients. Additionally, the AON/Wake module operates on a much lower clock frequency than
+    // the DUT, so glitches will (deliberately) be filtered out.
```
rswarbrick, line 22: If I understand correctly, this is pointing out that we want to wait long enough before reconnecting that the AON/Wake module will see the gap? Presumably that explains the lower bound? Would it be reasonable to make this a bit more obvious for uninformed readers? (me!)

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
```diff
@@ -0,0 +1,159 @@
...
+    disconn_interval_clks >= 2000 && disconn_interval_clks <= 50000;
+  }
+
+  // Attempt to modify the pull-up enables whilst AON/Wake is activated?
+  rand bit perturb_enables;
+
+  // Attempt to disrupt the USB pull-up enables
+  //  (any changes should not be visible on the USB when the AON/Wake module has control).
```
rswarbrick, line 30: Grammar nit: Maybe "none of these changes should be visible..."?

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
```diff
@@ -0,0 +1,159 @@
...
+    // Hand over control of the USB to the `usbdev_aon_wake` module
+    // TODO: we ought to try to overlap the stimulus and the AON/Wake activation.
+    aon_wake_activate();
+
+    // Expected state of pull-up enables after deactivating the AON/Wake module later.
+    exp_dp_pullup = !phy_pinflip;
+    exp_dn_pullup =  phy_pinflip;
+
+    // AON/Wake module should now have control of the USB pull ups; try to break things
+    // (but not in every case because doing so introduces a delay)...
```
rswarbrick, line 69: Could you expand on this? I assume it's not just an efficiency issue, so maybe "... a delay, which we don't always want because..." ?

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
```diff
@@ -0,0 +1,159 @@
...
+    // Suspend signaling.
+    super.body();
+
+    // Report configuration.
+    `uvm_info(`gfn, $sformatf("do_resume_signaling %d do_reset_signaling %d do_vbus_disconnects %d",
+                              do_resume_signaling, do_reset_signaling, do_vbus_disconnects),
+              UVM_MEDIUM)
+    // `usbdev_base_vseq` shall collect bus stimuli to be provided; just check here that we've been
+    // asked for only one.
+    `DV_CHECK_EQ(32'(do_resume_signaling) + 32'(do_reset_signaling) + 32'(do_vbus_disconnects), 1)
```
rswarbrick, line 58: Can this just be implemented by using `case` instead of `casez` below and adding a default error case?

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -564,6 +593,64 @@ endtask
     finish_item(m_sof_pkt);
   endtask
 
+  // Perform Resume Signaling on the USB; this instructs the DUT to exit a Suspended state.
+  virtual task send_resume_signaling();
+    usb20_item item;
+    `uvm_create_on(item, p_sequencer.usb20_sequencer_h)
+    start_item(item);
+    item.m_ev_type = EvResume;
+    finish_item(item);
```
rswarbrick, line 602: Would it make sense to factor this out into a helper task which takes the desired `m_ev_type` as an argument? I think it would let us drop quite a lot of the repetition below.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -564,6 +593,64 @@ endtask
...
+    `uvm_create_on(item, p_sequencer.usb20_sequencer_h)
+    start_item(item);
+    item.m_ev_type = enabled ? EvConnect : EvDisconnect;
+    finish_item(item);
+  endtask
+
+  // Hand over control of the USB to the AON/Wake module.
+  virtual task aon_wake_activate();
+    // We half the cycle count because csr_rd takes at least two cycles.
+    uint delay_max = (aon_wake_control_wr_clks + aon_wake_events_rd_clks) / 2;
```
rswarbrick, line 627: I'd probably consider avoiding having to estimate the latency of `csr_rd` by using a `fork/join_any` and waiting for `ral.wake_events.module_active` to take the expected value. This can also be shared code with the stuff in `aon_wake_deactivate`, I think.

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
alees24: It could. I'd rather keep the configuration check up front, and this sequence may over time get extended such that it provides repeated stimuli and/or runs multiple iterations, choosing randomly from a set of permitted stimuli each iteration.

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
alees24: The phrasing was intentional, and aimed at capturing the fact that we may not be making any changes. I've clarified the expression, thanks.

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
alees24: Done.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -40,12 +47,42 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+            cfg.bif.drive_vbus = 1'b1;
+            seq_item_port.item_done(rsp_item);
+          end
+          // Deassert VBUS (disconnection from USB host).
+          EvDisconnect: begin
+            cfg.bif.drive_vbus = 1'b0;
+            seq_item_port.item_done(rsp_item);
+          end
+          // Transmit packet (and maybe receive a response).
+          default: begin
```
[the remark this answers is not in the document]
alees24: Okay. Changed.

alees24:
I've added another small commit just to fix up the 'link_in_err' sequence which I discovered has been failing quite intermittently.

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
rswarbrick: Oh! Now I understand! Thanks for tweaking :-)

hw/ip/usbdev/dv/env/seq_lib/usbdev_aon_wake_vseq.sv
rswarbrick: Ok, that makes sense. Thanks for explaining.

rswarbrick (approved):
Looks good to me!

Issue #23158, opened 2024-05-16
[hmac] Add endianness swap option to KEY registers


ballifatih:
### Description

For MSG_FIFO and DIGEST registers of HMAC, we have the option of changing the endianess, and the driver enables them to match Ibex's little endian addressing. Here is how we configure them at the moment:
https://github.com/lowRISC/opentitan/blob/349bab6601c24b1d01b7264296bfc1c3bcd6c631/sw/device/lib/crypto/drivers/hmac.c#L53-L56

Ibex is running on little endian, which means that when we have a key/msg vector as follows:

https://github.com/lowRISC/opentitan/blob/28643828cb74dde504cc6b4d17dc458e2c2130a4/sw/device/tests/crypto/hmac_sha256_functest.c#L28-L34

however this representation does not match HW's expectation. Would it be possible add a similar option for KEY registers so that the driver can enable endianness swap? My worry is that if we have to change endianness at driver level, it would be quite awful from SCA perspective, as it handles key bytes during the swapping.

cc: @gdessouky @martin-velay @andreaskurth

gdessouky:
That makes sense, and should be possible, but will require quite a bit of corresponding DV changes. @ballifatih, can it be the same `endian_swap_bit` we use for the digest registers, or do you see a need for independent bits? 

I would have this as a last priority change, especially since HMAC is not SCA-hardened anyway, or at least have the corresponding DV a last priority because of all the DV work happening at the moment.

ballifatih:
We do not need an explicit setting for key so we can reuse `endian_swap_bit` (or even `digest_swap`) option. I think it would be nice if it isn't adding too much DV work. One use case for HMAC is HMAC-DRBG so it would be nice to have at least some amount of SCA resistance, even though HMAC is not masked.

ballifatih:
I can actually push RTL changes for this one later, I made one-line modification for my driver PR. I think the question is more about whether there is too much DV effort invovled.

martin-velay:
Using the existing field digest/endian swap might be a source of confusion. I would prefer to have a dedicated field `key_swap`. 

About the DV part of it, I would also prefer to make this change a bit later, the time we clear the existing issues. But if we add a dedicated field as suggested above, the RTL change could be merged, @ballifatih can test it at his end, and we'll just add a simple constraint in the IP level DV.

ballifatih:
[...]

PR #23242, opened 2024-05-21
[hmac,regs/rtl] Add key_swap field and expose Idle status to SW

gdessouky:
This PR adds the `key_swap` field to the HMAC `CFG` CSR per this issue https://github.com/lowRISC/opentitan/issues/23158 and exposes the HMAC idle status via the `STATUS` CSR per this issue https://github.com/lowRISC/opentitan/issues/22916. Both changes would need to be covered by DV in M5 @martin-velay.

This PR depends on previous PRs; once they are merged, it will only be the last commit c5e80715a4e1c7b0dcdcb48bb44d56dfea1584f2.

### DIFF
--- a/hw/ip/hmac/data/hmac.hjson
+++ b/hw/ip/hmac/data/hmac.hjson
@@ -225,7 +225,20 @@
             "excl:CsrAllTests:CsrExclWrite"
           ]
         }
-        { bits: "7:4",
+        { bits: "4",
+          name: "key_swap",
+          desc: '''Key register byte swap.
+
+                If 1 the endianness of each KEY_* register is swapped. Default value (value 0) is big endian representation of the KEY_* CSRs.
+                ''',
+          resval: "0",
+          tags: [
+            // Don't enable/disable key swap in automated CSR tests because it will have side effects on the KEY_* CSRs that the automated tests don't know of.
+            // This field will be covered in functional tests instead.
+            "excl:CsrAllTests:CsrExclWrite"
+          ]
+        }
+        { bits: "8:5",
           name: "digest_size",
           resval: "0x08",
           desc: '''Digest size configuration.
@@ -263,7 +276,7 @@
           tags: [// Exclude fields with non-zero reset value from write-read checks.
                "excl:CsrNonInitTests:CsrExclWriteCheck"]
         }
-        { bits: "13:8",
+        { bits: "14:9",
           name: "key_length",
           resval: "0x20",
           desc: '''Key length configuration.
@@ -362,11 +375,16 @@
       hwext: "true",
       fields: [
         { bits: "0",
+          name: "hmac_idle",
+          desc: "HMAC idle status."
+          resval: "1"
+        }
+        { bits: "1",
           name: "fifo_empty",
           desc: "FIFO empty",
           resval: "1"
         }
-        { bits: "1",
+        { bits: "2",
           name: "fifo_full",
           desc: "FIFO full. Data written to the FIFO whilst it is full will cause back-pressure on the interconnect"
         }
--- a/hw/ip/hmac/rtl/hmac.sv
+++ b/hw/ip/hmac/rtl/hmac.sv
@@ -86,6 +86,7 @@ module hmac
   logic        hmac_en;
   logic        endian_swap;
   logic        digest_swap;
+  logic        key_swap;
 
   logic        reg_hash_start;
   logic        sha_hash_start;
@@ -134,6 +135,7 @@ module hmac
   assign hw2reg.status.fifo_full.d  = fifo_full;
   assign hw2reg.status.fifo_empty.d = fifo_empty;
   assign hw2reg.status.fifo_depth.d = fifo_depth;
+  assign hw2reg.status.hmac_idle.d  = idle;
 
   typedef enum logic [1:0] {
     DoneAwaitCmd,
@@ -206,7 +208,8 @@ module hmac
       // Allow updating secret key only when the engine is in Idle.
       for (int i = 0; i < 32; i++) begin
         if (reg2hw.key[31-i].qe) begin
-          secret_key_d[32*i+:32] = reg2hw.key[31-i].q;
+          // swap byte endianness per secret key word if key_swap = 1
+          secret_key_d[32*i+:32] = conv_endian32(reg2hw.key[31-i].q, key_swap);
         end
       end
     end
@@ -273,7 +276,8 @@ module hmac
   logic unused_cfg_qe;
   assign unused_cfg_qe = ^{cfg_reg.sha_en.qe,      cfg_reg.hmac_en.qe,
                            cfg_reg.endian_swap.qe, cfg_reg.digest_swap.qe,
-                           cfg_reg.digest_size.qe, cfg_reg.key_length.qe };
+                           cfg_reg.key_swap.qe,    cfg_reg.digest_size.qe,
+                           cfg_reg.key_length.qe };
 
   assign sha_en               = cfg_reg.sha_en.q;
   assign hmac_en              = cfg_reg.hmac_en.q;
@@ -316,6 +320,7 @@ module hmac
 
   assign endian_swap = cfg_reg.endian_swap.q;
   assign digest_swap = cfg_reg.digest_swap.q;
+  assign key_swap    = cfg_reg.key_swap.q;
 
   assign hw2reg.cfg.hmac_en.d     = cfg_reg.hmac_en.q;
   assign hw2reg.cfg.sha_en.d      = cfg_reg.sha_en.q;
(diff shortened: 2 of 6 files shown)


hw/ip/hmac/rtl/hmac.sv
```diff
@@ -222,41 +225,45 @@ module hmac
...
+        // digest SW -> HW
+        digest_sw[i][31:0] = conv_endian32(reg2hw.digest[i].q, digest_swap);
+        digest_sw_we[i]    = reg2hw.digest[i].qe;
+      end else if ((digest_size == SHA2_384) || (digest_size == SHA2_512)) begin
         // digest SW -> HW
-        digest_sw[i]    = '0;
-        digest_sw_we[i] = '0;
+        digest_sw[i][63:32]    = conv_endian32(reg2hw.digest[2*i].q, digest_swap);
+        digest_sw[i][31:0]     = conv_endian32(reg2hw.digest[2*i+1].q, digest_swap);
+        digest_sw_we[i]        = reg2hw.digest[2*i].qe | reg2hw.digest[2*i+1].qe;
```
ballifatih, line 247: If you follow `reg2hw.digest[i].q` wires for different `i` values in `hmac_reg_top.sv`, you will see that they all end up being wired to `reg_wdata[31:0]`. So essentially when `reg2hw.digest[2*i]` is updated by SW, we are actually writing the two copies of the same 32-bit into `digest_sw`.

Your intuition is correct that we need to set `digest_sw_we[i]` when either of `reg2hw.digest[2*i].qe` or `reg2hw.digest[2*i+1].qe` is asserted, but we need to be selecting the correct half to update.

hw/ip/hmac/data/hmac.hjson
```diff
@@ -316,6 +316,19 @@
           tags: [// Exclude fields with non-zero reset value from write-read checks.
                  "excl:CsrNonInitTests:CsrExclWriteCheck"]
         }
+        { bits: "14",
+          name: "key_swap",
+          desc: '''Key register byte swap.
+
+                If 1 the endianness of each KEY_* register is swapped.
```
ballifatih, line 323: We could also add that the default value (0) is big endian representation.

hw/ip/hmac/rtl/hmac.sv
```diff
@@ -222,41 +225,45 @@ module hmac
...
-          hw2reg.digest[i].d = inter_digest[63:32];
-          // digest SW -> HW
-          digest_sw[i/2]    = conv_endian64({reg2hw.digest[i].q,reg2hw.digest[i-1].q}, digest_swap);
-          digest_sw_we[i/2] = reg2hw.digest[i].qe | reg2hw.digest[i-1].qe;
-        end
+        // digest HW -> SW
+        // digest swap the entire 64-bit word
+        inter_digest           = conv_endian64(digest[i], digest_swap);
+        hw2reg.digest[2*i].d   = inter_digest[63:32];
+        hw2reg.digest[2*i+1].d = inter_digest[31:0];
```
ballifatih, line 262: I disagree with the order of two 32-bit halves. I think the previous implementation was correct (which I didn't have issue testing when `digest_swap = true`).

martin-velay:
This functionality will be tested later as mentioned. Here is the related issue: #23245

martin-velay (approved):

hw/ip/hmac/rtl/hmac.sv
gdessouky: see my comment here https://github.com/lowRISC/opentitan/pull/23197#discussion_r1609493429, I see it's just a debate on how the driver will interpret the CSRs to generate the 64-bit digest words. Either way will work, happy to adapt to the convention that makes the driver implementation nicer.

gdessouky:
> This functionality will be tested later as mentioned. Here is the related issue: #23245

Don't we also need to adapt DV when reading the STATUS CSR once the Idle bit is also exposed in there? DV is unforgiving.

martin-velay:
> > This functionality will be tested later as mentioned. Here is the related issue: #23245
> 
> Don't we also need to adapt DV when reading the STATUS CSR once the Idle bit is also exposed in there? DV is unforgiving.

What do you mean?

gdessouky:
> > > This functionality will be tested later as mentioned. Here is the related issue: #23245
> > 
> > 
> > Don't we also need to adapt DV when reading the STATUS CSR once the Idle bit is also exposed in there? DV is unforgiving.
> 
> What do you mean?

Only that when Idle is exposed in the `STATUS` CSR, we need to predict it throughout the scoreboard and compare it with the STATUS CSR when read out . Not a big issue.

hw/ip/hmac/rtl/hmac.sv
```diff
@@ -134,6 +135,7 @@ module hmac
   assign hw2reg.status.fifo_full.d  = fifo_full;
   assign hw2reg.status.fifo_empty.d = fifo_empty;
   assign hw2reg.status.fifo_depth.d = fifo_depth;
+  assign hw2reg.status.hmac_idle.d  = idle;
```
gdessouky, line 138: this exposes `idle_o` assigned in https://github.com/lowRISC/opentitan/blob/55e80651de027916be3010c0e6e57af669ea03d8/hw/ip/hmac/rtl/hmac.sv#L864 from hmac.sv not the idle_o from hmac_core.sv (https://github.com/lowRISC/opentitan/pull/23233#issuecomment-2123623418).

hw/ip/hmac/data/hmac.hjson
```diff
@@ -374,6 +387,10 @@
           name: "fifo_depth",
           desc: "FIFO entry count."
         }
+        { bits: "10",
```
andreaskurth, line 390: Wouldn't it make more sense to put this status bit into the first nibble (i.e., at bit 2), together with the FIFO status bits?

hw/ip/hmac/rtl/hmac.sv
andreaskurth: There's one more line in that assignment:
https://github.com/lowRISC/opentitan/blob/55e80651de027916be3010c0e6e57af669ea03d8/hw/ip/hmac/rtl/hmac.sv#L864-L865

so LGTM

hw/ip/hmac/rtl/hmac.sv
```diff
@@ -316,13 +320,16 @@ module hmac
...
+  assign key_swap    = cfg_reg.key_swap.q;
 
   assign hw2reg.cfg.hmac_en.d     = cfg_reg.hmac_en.q;
   assign hw2reg.cfg.sha_en.d      = cfg_reg.sha_en.q;
   assign hw2reg.cfg.digest_size.d = digest_mode_e'(digest_size);
   assign hw2reg.cfg.key_length.d  = key_length_e'(key_length);
   assign hw2reg.cfg.endian_swap.d = cfg_reg.endian_swap.q;
   assign hw2reg.cfg.digest_swap.d = cfg_reg.digest_swap.q;
+  assign hw2reg.cfg.key_swap.d    = cfg_reg.key_swap.q;
+
```
andreaskurth, line 332: Stray newline?

andreaskurth (commented):
Mostly LGTM, just left a few comments

hw/ip/hmac/data/hmac.hjson
gdessouky: I would move it to bit 0 then and have all the FIFO relevant fields (empty, full and depth) in sequence afterwards. This will probably break a couple of DV checks, but all quickly fixed. How does that sound?

hw/ip/hmac/data/hmac.hjson
```diff
@@ -316,6 +316,19 @@
           tags: [// Exclude fields with non-zero reset value from write-read checks.
                  "excl:CsrNonInitTests:CsrExclWriteCheck"]
         }
+        { bits: "14",
```
gdessouky, line 319: I had also just moved this now (in my next push) after `digest_swap` such that they are `endian_swap`, `digest_swap` and `key_swap` in sequence. @andreaskurth

gdessouky:
This now passes `smoke` at my end, but it may have broken some other block-level tests because `STATUS` has changed. But I would push fixes in another PR so as not to hold this one (with RTL changes) back.

hw/ip/hmac/data/hmac.hjson
andreaskurth: That makes sense, thx!

andreaskurth (approved):
LGTM now, thx @gdessouky :+1:

andreaskurth:
The failing CI check seems to be unrelated to this PR, all other tests have passed. Merging!

ballifatih:
`key_swap` option works with the driver, but I had some issues with `Idle` so far. I haven't had a change to look deeper into what the issue is with `Idle`. It could be that the driver has different assumptions on what `Idle` means, or when to check it.

`Idle` is not a blocker for the driver at the moment, so I would not fret over it.

Issue #15886, opened 2022-11-01
[RDC] Check if CSb can be related to any generated clocks from SCK


eunchan:
`SPI_DEV_{IN/OUT}_CLK` are generated clocks from `SPI_DEV_CLK`. Somehow, RDC tool reports `E_RST_METASTABILITY` for the path from flops resetting by CSb to the flops in the same clock category (`SPI_DEV*CLK`).

- [ ] Contact the tool vendor if the setup is wrong (sdc, env)

CC: @tjaychen @jeoongp 

> estimate 4
> remaining 2023-03-22 4

andreaskurth:
Triaged for `spi_device`. Assigning to M2.5 with https://github.com/lowRISC/opentitan/labels/Priority%3AP2.

hcallahan-lowrisc:
> estimate 4
> remaining 2023-03-22 4

johngt:
@cindychip - this is tagged as an RDC potential tooling error. Please suggest who might be able to look into this. TIA

a-will:
Agreed that it's a tool limitation. Our timing constraints should be set up so CSB is never permitted to change near either SPI_DEV_CLK edge--The two are not active at the same time (not for modes we support).

msfschaffner:
@moidx I think we can lower the priority on this one and defer it to M2.5.2.

andreaskurth:
Confirmed again with @a-will that CSB can never change near either SPI_DEV_CLK edge, so suggest to move this to Backlog

a-will:
It looks like I may have been more restrictive than the typical SPI flash datasheet. I'll post a PR that creates a reset synchronized for outclk domain usage. Then we might only need an exception for the reset assertion.

andreaskurth:
Closing as completed in #23244

PR #23244, opened 2024-05-21
[spi_device] Add CSB reset sync for outclk domain

a-will:
For flops that latch on the falling edge of SCK, an unsafe reset timing can occur if SCK is permitted to toggle while CSB toggles. SPI NOR flash datasheets typically only specify setup and hold requirements around the rising edge of SCK, so these flops are not protected (at least without OT-specific requirements on the datasheet).

Add a synchronizing flop that releases the CSB reset on the rising edge of SCK, and create a separate reset network for the outclk domain to fix the RDC issue.


a-will:
This should fix https://github.com/lowRISC/opentitan/issues/15886.

antmarzam:
LGTM!

antmarzam (approved):

hcallahan-lowrisc (approved):
This LGTM. Thanks @a-will

PR #23249, opened 2024-05-22
Fix pwrmgr sival

pamaury:
Even though the test plan says not to run `pwrmgr_normal_sleep_all_wake_ups` (and deep variant) and to prefer the random version, it is current in a weird state where it is enabled but does not specify a test harness which causes it to fail.


jwnrt (approved):
Thanks for the fix

engdoreis (approved):

PR #23253, opened 2024-05-22
[usb,sival] Various improvements and fixes

pamaury:
The main purpose of this PR is to fix the test harness that I broke by mistake in https://github.com/lowRISC/opentitan/commit/52714dcdce25352e48537c8dc7c9063ee7b261fc
The issue is that in certain test, the harness is needed to enable VBUS sensing but no USB device will actually appear on the bus so the harness must not wait for it. I took this opportunity to do some further cleanup/improvements.
The reason this failure was not caught by CI is that certain tests do not require the harness on the CW310 so it was not used and therefore the CI didn't break. This PR changes that to run the harness (even when almost useless) even on the CW310.

### DIFF
--- a/sw/device/tests/usbdev_mixed_test.c
+++ b/sw/device/tests/usbdev_mixed_test.c
@@ -46,7 +46,7 @@ Thoughts:
 #define NUM_STREAMS USBUTILS_STREAMS_MAX
 #endif
 
-#define TRANSFER_BYTES_SILICON (0x10U << 20)
+#define TRANSFER_BYTES_SILICON (4U << 20)
 #define TRANSFER_BYTES_FPGA (8U << 16)
 
 // This is appropriate for a Verilator chip simulation with 15 min timeout
--- a/sw/device/tests/usbdev_stream_test.c
+++ b/sw/device/tests/usbdev_stream_test.c
@@ -35,7 +35,7 @@
 #define NUM_STREAMS USBUTILS_STREAMS_MAX
 #endif
 
-#define TRANSFER_BYTES_SILICON (0x10U << 20)
+#define TRANSFER_BYTES_SILICON (8U << 20)
 // 16MiB takes about 256s presently with 10MHz CPU in CW310 FPGA and physical
 // USB with randomized packet sizes and the default memcpy implementation;
 // The _MEM_FASTER switch drops the run time to 187s. In CI, we want to keep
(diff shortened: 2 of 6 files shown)


engdoreis (approved):
LGTM.

nbdd0121 (approved):

sw/device/tests/usbdev_stream_test.c
```diff
@@ -35,7 +35,7 @@
 #define NUM_STREAMS USBUTILS_STREAMS_MAX
 #endif
 
-#define TRANSFER_BYTES_SILICON (0x10U << 20)
+#define TRANSFER_BYTES_SILICON (0x10U << 16)
```
alees24, line 38: Please don't reduce this value so far; it will barely be testing anything, and there's a danger that if something makes the streaming massively slower/less efficient, it will go completely unnoticed.

usbdev_stream_test is much faster than usbdev_mixed_test and usbdev_iso_test because it is using Bulk Transfers and running as fast as the bus can go. Just 1MiB completes in 4.8 seconds on my laptop, which is well below the 30s that you proposed originally.

Even the 16MiB originally that is specified for usbdev_stream_test should complete in 38s (and thus not time out with the 60s limit that is presently set).

sw/device/tests/usbdev_mixed_test.c
```diff
@@ -46,7 +46,7 @@
 #define NUM_STREAMS USBUTILS_STREAMS_MAX
 #endif
 
-#define TRANSFER_BYTES_SILICON (0x10U << 20)
+#define TRANSFER_BYTES_SILICON (0x10U << 16)
```
alees24, line 49: I see the previous 16MiB amount timing out on Si, but 4MiB takes about 20 seconds to complete.
usbdev_iso_test and usbdev_mixed_test should be about the same as each other, being limited by the 64 bytes/1ms polling rater of Isochronous and Interrupt endpoints.

alees24 (changes requested):
This all looks okay, except for the timeouts which run a risk of being reduced the point of doing nothing useful when they are all unilaterally divided by16 to prevent timeouts.

It is a guessing game exactly how much other traffic/activity there will be on a decidedly non-real time OS, but I would suggest the following amounts, please:
- usbdev_stream_test 8MiB (8U << 20) - estimated 19 secs
- usbdev_mixed_test 4MiB - 20 secs
- usbdev_iso_test 4MiB - 20 secs

pamaury:
I can reproduce your timing for the stream and mixed test but the iso test on crius seems to take significantly longer (40s), in particular it takes quite longer for the host to receive all the data. It  seems very consistent though so I think we should be fine.

alees24 (approved):
Thanks.

PR #23254, opened 2024-05-22
[sival, rv_core_ibex] make rv_core_ibex_mem_test work on silicon

nbdd0121:
Changes made:
* ROM check is skipped because it's locked by ROM_EXT.
* Since ROM_EXT locks address translation, the MMIO test location is
  switched to a rv_timer register instead.
* ROM_EXT locks down its own flash pages, so the page being tested is moved
  from page 0 to an accessible page. The default property config is also
  changed to a specific data region.

### DIFF
--- a/sw/device/tests/rv_core_ibex_mem_test.c
+++ b/sw/device/tests/rv_core_ibex_mem_test.c
@@ -17,6 +17,7 @@
  * Two MMIO registers from two different devices are written to and read from.
  */
 
+#include "sw/device/lib/arch/boot_stage.h"
 #include "sw/device/lib/arch/device.h"
 #include "sw/device/lib/base/csr.h"
 #include "sw/device/lib/dif/dif_flash_ctrl.h"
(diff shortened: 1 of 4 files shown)


sw/device/tests/rv_core_ibex_mem_test.c
```diff
@@ -158,24 +180,37 @@ static void setup_flash(void) {
...
+    // For the execution test we a specific `c.jr x1` (i.e. function return)
+    // instruction. Since the address can vary between ROM builds, we scan a
+    // small region to find it.
+    volatile uint16_t *test_loc;
+    for (test_loc = (uint16_t *)kRomTestLocStart;
+         test_loc < (uint16_t *)kRomTestLocEnd; test_loc++) {
+      if (*test_loc == kRomTestLocContent) {
+        break;
+      }
+    }
```
engdoreis, line 196: Nit. A tiny tweak to make it easier to read in my option. But feel free to ignore. 

```suggestion
    volatile uint16_t *test_loc = (uint16_t *)kRomTestLocStart;
    while(test_loc < (uint16_t *)kRomTestLocEnd && *test_loc != kRomTestLocContent) {
       test_loc++;
    }
```

engdoreis (approved):

sw/device/tests/rv_core_ibex_mem_test.c
nbdd0121: I personally prefer the current structure as I think it better reflects the "loop through all words and search" intent

PR #23255, opened 2024-05-22
[rom_ext] Select address tranlation based on link address

cfrantz:
1. Select address translation based on the link address of the `.text` section.  If the link address is in EFLASH, then address translation is `false`, else `true`.
2. Remove various `manifest_virtual` targets from `BUILD` files in favor of using the automatic selection in (1).


moidx (approved):

cfrantz:

PR #23262, opened 2024-05-22
[rom] Report the correct boot log `rom_ext_slot`

cfrantz:
The rom was incorrectly determining the value to place into the `rom_ext_slot` of the boot log.


milesdai (approved):
Oof -- thanks for catching that!

cfrantz:

Issue #23259, opened 2024-05-22
[rom_ext] Prevent incorrect MinimumSecurityVersion BootSvc call


jettr:
### Description

If ROM_EXT receives a SetMinimumSecurityVersion boot services message for a value that is higher than the actively running owner firmware image (e.g. `coalesce (nextBootSlot, primaryBootSlot)`), then the request should be ignored. This prevents help prevent a denial of service attack with minimum security version.

PR #23267, opened 2024-05-22
[rom_ext_e2e] Test the `MinBl0SecVer` request

cfrantz:
1. Verify the manifests of both slots to determine a maximum `min_sec_ver`.  We prevent advancing the `min_sec_ver` beyond the value associated with any valid firmware in flash.
2. Test the `MinBl0SecVer` request by advancing the minimum security version forward.  Attempt to go beyond the maximum allowed value and confirm that the request is rejected.  Attempt to go backwards and confirm that the request is rejected.

This addresses #23259.

### DIFF
--- a/sw/device/silicon_creator/rom_ext/rom_ext.c
+++ b/sw/device/silicon_creator/rom_ext/rom_ext.c
@@ -621,33 +621,48 @@ static rom_error_t boot_svc_min_sec_ver_handler(boot_svc_msg_t *boot_svc_msg,
   const uint32_t current_min_sec_ver = boot_data->min_security_version_bl0;
   const uint32_t requested_min_sec_ver =
       boot_svc_msg->next_boot_bl0_slot_req.next_bl0_slot;
-  const uint32_t diff = requested_min_sec_ver - current_min_sec_ver;
 
   // Ensure the requested minimum security version isn't lower than the current
   // minimum security version.
   if (launder32(requested_min_sec_ver) > current_min_sec_ver) {
     HARDENED_CHECK_GT(requested_min_sec_ver, current_min_sec_ver);
-    HARDENED_CHECK_LT(diff, requested_min_sec_ver);
+    uint32_t max_sec_ver = current_min_sec_ver;
 
-    // Update boot data to the requested minimum BL0 security version.
-    boot_data->min_security_version_bl0 = requested_min_sec_ver;
+    // Check the two flash slots for valid manifests and determine the maximum
+    // value of the new minimum_security_version.  This prevents a malicious
+    // MinBl0SecVer request from making the chip un-bootable.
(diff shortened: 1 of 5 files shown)


jettr (commented):
It is possible in our owner firmware for the FW update flow to put a  manifest in the other slot that is self consistent, but does not have a valid signature for the rest  of the data (our owner firmware does not validate the RSA signature when we are putting the fw image in the inactive slot -- we depends on ROM_EXT to perform that check).

If an attacker were to create a well formed manifest with a very high min security version, then call minsecurity version, the DoS attack still succeeds. The inactive slot isn't bootable since there isn't a valid RSA signature and now the previously active slot can't boot because the min security version is too high. Thoughts?

cfrantz:
> If an attacker were to create a well formed manifest with a very high min security version, then call minsecurity version, the DoS attack still succeeds. The inactive slot isn't bootable since there isn't a valid RSA signature and now the previously active slot can't boot because the min security version is too high. Thoughts?

The `boot_svc_min_sec_ver_handler` checks the signature on both slots and picks the maximum allowable security version only from slots that pass sigverify.  Malformed images are disqualified and do not influence the maximum.

jettr (approved):
> The `boot_svc_min_sec_ver_handler` checks the signature on both slots and picks the maximum allowable security version only from slots that pass sigverify. Malformed images are disqualified and do not influence the maximum.

I was trying to follow the code and I couldn't tell if it was just verifying the manifest layout or if is performed a full validation of each slot. So this boot command will take a few milliseconds to process then, which seem very reasonable.

moidx (approved):

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -621,33 +621,48 @@ static rom_error_t boot_svc_min_sec_ver_handler(boot_svc_msg_t *boot_svc_msg,
...
+    }
+    manifest = rom_ext_boot_policy_manifest_b_get();
+    error = rom_ext_verify(manifest, boot_data);
+    if (error == kErrorOk && manifest->security_version > max_sec_ver) {
+      max_sec_ver = manifest->security_version;
+    }
 
-    // Update boot data to the requested minimum BL0 security version.
-    boot_data->min_security_version_bl0 = requested_min_sec_ver;
+    if (requested_min_sec_ver <= max_sec_ver) {
```
moidx, line 645: Something we can consider from a hardening perspective is whether it makes sense to add a crc over all the boot services commands. This is to avoid a fault being able to modify command parameters. 

We should review if this is something needed on a case by case basis.

jettr: All Boot Service commands actually do have a "CRC" in the form of SHA256 digest: https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/lib/boot_svc/boot_svc_header.h#L41

PR #23274, opened 2024-05-23
[topgen] Generically filter for MMIO region visible devices

### LINKED ISSUE #14345: [top] "rom" interface of "rv_dm" is not a memory in top_earlgrey.hjson
Noticed this while working on #14336 and discussed offline with @tjaychen. Currently, the entry for rv_dm [here](https://github.com/lowRISC/opentitan/blob/1447b4853447dffbc0193b5aab27d96012f950ff/hw/top_earlgrey/data/top_earlgrey.hjson#L572) looks like this:

```
    { name: "rv_dm",
      type: "rv_dm",
      clock_srcs: {clk_i: "main"},
      clock_group: "infra",
      reset_connections: {rst_ni: "lc"},
      param_decl: {
        IdcodeValue: "jtag_id_pkg::JTAG_IDCODE",
      }
      // Note that this module also contains a bus host.
      base_addrs: {rom: "0x00010000", regs: "0x41200000"}
    },
```
which doesn't create a memory.

As a result, it is not included in the return value of `TopGenC::memories()` and required a minor workaround in #14336.

Razer6:
Remove hardcoded devices and interfaces for MMIO region filtering by adding two new optional module attributes cpu_visible and cpu_not_visible.

By unifying the implementation of the C and Rust generator, an inconsistency between the two implementations was uncovered. Ideally, `TopGenC` and `TopGenRust` will be refactored with a common base class to remove as much of the code duplication. I'm not sure though if that is the right PR to fix that.

This fixes #14345

### DIFF
--- a/hw/top_darjeeling/data/top_darjeeling.hjson
+++ b/hw/top_darjeeling/data/top_darjeeling.hjson
@@ -166,7 +166,62 @@
   // all peripherals, though not every peripheral will be accessible to every
   // host in that address space--Access privileges are separate from addresses.
   addr_spaces: [
-    { name: "hart", desc: "The main address space, shared between the CPU and DM"},
+    { name: "hart"
+      desc: "The main address space, shared between the CPU and DM"
+      subspaces: [
+        { name: "mmio",
+          desc: '''
+                MMIO region excludes any memory that is separate from the module configuration
+                space, i.e. ROM, main SRAM, and mbx SRAM are excluded but retention SRAM or
+                spi_device are included.
+                '''
+          nodes: [
+            "uart0",
+            "gpio",
+            "spi_device",
+            "i2c0",
+            "rv_timer",
+            "otp_ctrl",
+            "lc_ctrl.regs",
+            "alert_handler",
+            "spi_host0",
+            "pwrmgr_aon",
+            "rstmgr_aon",
+            "clkmgr_aon",
+            "pinmux_aon",
+            "aon_timer_aon",
+            "ast"
+            "sensor_ctrl",
+            "soc_proxy.core",
+            "sram_ctrl_ret_aon",
+            "rv_plic",
+            "aes",
+            "hmac",
+            "otbn",
+            "keymgr_dpe"
+            "csrng",
+            "edn0",
+            "edn1",
+            "sram_ctrl_main.regs",
+            "sram_ctrl_mbox.regs",
+            "rom_ctrl0.regs",
+            "rom_ctrl1.regs",
+            "dma",
+            "mbx0.core",
+            "mbx1.core",
+            "mbx2.core",
+            "mbx3.core",
+            "mbx4.core",
+            "mbx5.core",
+            "mbx6.core",
+            "mbx_jtag.core",
+            "mbx_pcie0.core",
+            "mbx_pcie1.core",
+            "rv_core_ibex"
+          ],
+        },
+      ]
+    }
     { name: "soc_mbx", desc: "SoC address space for mailbox access"},
     { name: "soc_dbg", desc: "SoC address space for debug module interfaces"},
   ]
--- a/util/topgen/lib.py
+++ b/util/topgen/lib.py
@@ -85,6 +85,27 @@ class Name:
         return Name([p for p in self.parts if p != part_to_remove])
 
 
+class MemoryRegion(object):
+    def __init__(self, name: Name, base_addr: int, size_bytes: int):
+        assert isinstance(base_addr, int)
+        self.name = name
+        self.base_addr = base_addr
+        self.size_bytes = size_bytes
+        self.size_words = (size_bytes + 3) // 4
+
+    def base_addr_name(self):
+        return self.name + Name(["base", "addr"])
+
+    def offset_name(self):
+        return self.name + Name(["offset"])
+
+    def size_bytes_name(self):
+        return self.name + Name(["size", "bytes"])
+
+    def size_words_name(self):
+        return self.name + Name(["size", "words"])
+
+
 def is_ipcfg(ip: Path) -> bool:  # return bool
     log.info("IP Path: %s" % repr(ip))
     ip_name = ip.parents[1].name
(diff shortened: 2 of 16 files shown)


hw/top_darjeeling/data/top_darjeeling.hjson
```diff
@@ -577,6 +578,7 @@
         regs: {hart: "0x21200000"},
         dbg:  {soc_dbg: "0x00000000"},
       },
+      cpu_not_visible: ["mem", "regs", "dbg"]
```
rswarbrick, line 581: It's a bit sad to have the names appearing twice. Could we switch it so that we end up with something like this?
```
      // Note that this module also contains a bus host.
      base_addrs: {
        mem:  {hart: "0x00040000", cpu_visible: false},
        regs: {hart: "0x21200000", cpu_visible: false},
        dbg:  {soc_dbg: "0x00000000", cpu_visible: false},
      },
```
and we could make it so a region has a default visibility attribute (true, presumably)

hw/top_darjeeling/data/top_darjeeling.hjson
```diff
@@ -553,6 +553,7 @@
         regs: {hart: "0x30500000"},
         ram:  {hart: "0x30600000"},
       },
+      cpu_visible: ['ram']
```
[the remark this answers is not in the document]
Razer6: That is totally right. This resulted from the  fact, that everything except memories was included by default. I removed that base, and explicitly excluded memories from the MMIO range (except for the retention RAM).

hw/top_darjeeling/data/top_darjeeling.hjson
Razer6: I changed the implementation according to that. It was required to add some special cases when iterating over that struct. Should we add a top-level validation, to ensure that there is no address space named `cpu_visible`?

Razer6:
> Would it make more sense to formalize memory region groups? The "MMIO region" is not a property of an individual IP instance. Instead, it's a contiguous grouping of nodes in the address space.
> 
> Validating memory region groups (...or "subspaces"? Not sure what would be a good name yet...) would mean ensuring that all nodes between the two ends are included in the named group, and we'd have to be explicit about each node that is intended to be in the group. topgen would need to error out if any node is missing, since that would be an inconsistent subspace.
> 
> I kind of feel like topgen is a dubious place to do this... but if we really want the internal check for minimizing ePMP entries, I guess it makes sense. If there's no DRC, then it might make more sense to kick this out to downstream software completely. Well... maybe it's also convenient to define regions with just the two ends and abandon the check /shrug

I'm not sure if downstream software actually uses those defines? If not, we maybe can remove that completely?

a-will:
> > Would it make more sense to formalize memory region groups? The "MMIO region" is not a property of an individual IP instance. Instead, it's a contiguous grouping of nodes in the address space.
> > Validating memory region groups (...or "subspaces"? Not sure what would be a good name yet...) would mean ensuring that all nodes between the two ends are included in the named group, and we'd have to be explicit about each node that is intended to be in the group. topgen would need to error out if any node is missing, since that would be an inconsistent subspace.
> 
> Not sure, if that is in the scope of this PR. I guess we can support multiple MMIO regions in a different PR if needed.

This is exactly the scope of what you are modifying, though. The "MMIO region" is merely a macro defined in a software header. The code changed here has no relationship with the CPU's visibility of node.

util/topgen/lib.py
```diff
@@ -566,3 +587,28 @@ def is_lc_ctrl(modules):
...
+    MMIO region excludes any memory that is separate from the module configuration
+    space, i.e. ROM, main SRAM, and flash are excluded but retention SRAM,
+    spi_device memory, or usbdev memory are included.
+    """
+    regions = []
+    for ((dev_name, if_name), region) in devices:
+        module = get_module_by_name(top, dev_name)
+      
+        if module['base_addrs'][if_name].get("cpu_visible", True):
+            regions.append(region)
```
a-will, line 605: Please no hard-coded address space name exceptions. This violates the data structure's layout.

a-will (approved):
I think we've got some misunderstandings of what the original code does here. It defines a contiguous region of the address map and calls it the "MMIO region" so software can program a single entry for it in the ePMP.

CPU visibility doesn't factor into it. Instead, there are hard-coded exceptions in topgen's output helpers that try to define the bounds of the region. If you want to remove those hard-coded exceptions, we would need to either explicitly define non-MMIO nodes for each addr_space, or we could generalize and define multiple such regions, with "MMIO" just being one of them.

a-will (changes requested):
Whoops, meant to request changes on that last review. Doh!

util/topgen/lib.py
Razer6: Exactly, I thought about that too. What about an address space validation as proposed in https://github.com/lowRISC/opentitan/pull/23274#discussion_r1618895737 ?

Razer6:
What about defining it like this:

`addr_group` defines the address range, the node is associated to. If nothing provided, it associates them to the default `mmio` group. If set to `none` or `false`, not sure what works, they will be excluded. The helper would then determine all `addr_group`s per address space generates the definitions for them. 

```
base_addrs: {
        mem:  {hart: "0x00040000", addr_range: none},
        regs: {hart: "0x21200000", addr_range: "mmio"},
        dbg:  {soc_dbg: "0x00000000", addr_range: none},
      },
```

a-will:
For the option where we generically define subspaces of an addr_space and explicitly include each node, it could look like this:

```hjson
  addr_spaces: [
    { name: "hart",
      desc: "The main address space, shared between the CPU and DM"},
      subspaces: [
        { name: "mmio",
          desc: "The memory region for peripherals",
          nodes: [
            "uart0", "gpio", "spi_device", "i2c0", "rv_timer",
            "otp_ctrl", // Shorthand for all otp_ctrl nodes
            "lc_ctrl.regs", // Can select specific nodes with a dot separator
            "alert_handler", "spi_host0", "pwrmgr_aon",  // And so on
          ],
        },
      ],
    },
    { name: "soc_mbx", desc: "SoC address space for mailbox access"},
    { name: "soc_dbg", desc: "SoC address space for debug module interfaces"},
  ]
```

Actually, we could even choose to move all the base_addr definitions into this three-level tree, and that would make us have compatible definitions with IP-XACT (where a "subspace" is the same as an IP-XACT "block"). SystemRDL seems to allow nesting addrmap definitions, so it's a little more free. But anyway, that would look like this:

```hjson
  addr_spaces: [
    { name: "hart",
      desc: "The main address space, shared between the CPU and DM"},
      subspaces: [
        { name: "mmio",
          desc: "The memory region for peripherals",
          nodes: {
            uart0:                "0x30010000",
            gpio:                 "0x30000000",
            spi_device:           "0x30310000",
            i2c0:                 "0x30080000",
            rv_timer:             "0x30100000",
            otp_ctrl.core:        "0x30130000",
            otp_ctrl.prim:        "0x30138000",
            lc_ctrl.regs:         "0x30140000",
            // And so on
          },
        },
      ],
      nodes: [ // Same as the "nodes" in the subspaces, but these are not in a named subspace.
      ],
    },
```

Razer6:
Looks good to me. The first one seems easy to me. Whats quite attractive to the second one is to have a base address table basically on one screen. It makes it much more readable for the base addresses, which already annoyed me a few times.

a-will:
> Looks good to me. The first one seems easy to me. Whats quite attractive to the second one is to have a base address table basically on one screen. It makes it much more readable for the base addresses, which already annoyed me a few times. However, the changes are of course much more intrusive though.

Would it make sense to start with the first one? We could choose to move to the second one later.

The first solution annoyingly adds some burden to update yet another place in the file when you assign new base addresses. That might motivate us to eventually get to the second solution instead of forever punting it, though, haha.

Razer6:
> Would it make sense to start with the first one? We could choose to move to the second one later.

That sounds good. The second one is an improvement on top of it, but touches a lot of places. 

> The first solution annoyingly adds some burden to update yet another place in the file when you assign new base addresses. That might motivate us to eventually get to the second solution instead of forever punting it, though, haha.

Totally agree. And it really shows you the memory map in one place. Quite hard to scroll down if there are many more devices ;-)

a-will (approved):
Oh, this looks great!

hw/top_darjeeling/data/top_darjeeling.hjson
```diff
@@ -166,7 +166,62 @@
...
+      desc: "The main address space, shared between the CPU and DM"
+      subspaces: [
+        { name: "mmio",
+          desc: '''
+                MMIO region excludes any memory that is separate from the module configuration
+                space, i.e. ROM, main SRAM, and mbx SRAM are excluded but retention SRAM or
+                spi_device are included.
+                '''
+          nodes: [
+            "uart0",
```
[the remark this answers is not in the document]
Razer6: Haha, no worries 😆

andreaskurth:
CHANGE AUTHORIZED: hw/top_darjeeling/data/top_darjeeling.hjson

PR #23275, opened 2024-05-23
[otlib] Document teacup pins in opentitanlib

pamaury:


sw/host/opentitanlib/src/app/config/structs.rs
```diff
@@ -145,6 +154,20 @@ pub struct I2cConfiguration {
...
+pub struct IoPeripheral {
+    /// The user-visible name of this pheripheral.
+    pub name: String,
+    /// List of pins that this peripheral is connected to.
+    pub pins: Vec<String>,
+    /// If present, this peripheral will only be enable if that particular pin is high.
+    pub enable_pin: Option<String>,
+    /// If set, the enable signal is inverted (i.e. active low).
+    #[serde(default)]
+    pub enable_low: bool,
```
nbdd0121, line 168: bikeshedding:
```suggestion
    pub enable_invert: bool,
```

engdoreis: The concept of inverted signal already exists in https://github.com/lowRISC/opentitan/blob/cbca064766066e90c06addfbcc7d2ee465d55ca7/sw/host/opentitanlib/src/app/config/opentitan_cw340.json#L8

a-will:
> I have updated the mapping, I think it's essentially complete but I edited the top comment to add some of the problems I found. The main issue seems to be that we have a list of i2c/spi devices but (as far as I can tell) no way to tell to which pins they are connected to. The fact that it has an `alias_of` seems to be used to send a hyperdebug command to get the required information.
> 
> One solution would be to add a `pins` fields to `i2c`, `spi` and `uart`. Alternatively, this field could be left empty and then needs to be filled by the transport implementation.

Haha, I'm not sure if anyone would agree with me, but instead of continuing to extend our ad hoc formats, I'd love to see us adopt the devicetree format to describe the targets and just have opentitanlib take it in and work with it. I've written the first bits of topgen support already, and devicetree does composition well.

luismarques:
> Haha, I'm not sure if anyone would agree with me, but instead of continuing to extend our ad hoc formats, I'd love to see us adopt the devicetree format to describe the targets and just have opentitanlib take it in and work with it. I've written the first bits of topgen support already, and devicetree does composition well.

That sounds reasonable to Amaury and me.

jesultra:
> * we have a [list of i2c devices](https://github.com/lowRISC/opentitan/blob/master/sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json#L420) with `alias_of` to names like `I2C1` and so on. I assume those are aliases to the transport-specific pins for i2c but I don't see anything that specifies to what OT pins this corresponds to?
It is true that the opentitanlib backends can support a set of named GPIOs and a set of named I2C ports, etc. but does not advertise which pins UARTS, I2C ports or SPI ports actually use.  On some backends such as UltraDebug, the pins that do SPI or UART cannot work as GPIO and have no GPIO names, on others such as Sam3X the SPI port can be moved to be at any set of pins, through a capable pinmux, on HyperDebug each of the three SPI ports can use any GPIO pin as chip select, but the other signals are fixed on particular pins, which must be put into "Alternate" mode, and the user is expected to just know which pins have what function.

> I am suggesting that we extend the current json configuration files with more information.

> Comments/Thoughts are welcome.
I am open to the idea, but I think it is going to be hard to completely describe everything on the Teacup board, in a way that test scripts can use.

Let me try to describe what we are doing in my team.  We use HyperDebug in at least 4 different hardware combinations, (Teacup, CW310+swizzle, and then two teacup-like boards with legacy GSCs).  We have some test scripts that do specific things on only one setup, and other test scripts that test basically the same thing across different hardware setups.

In order to manage this, we have a "hardware configuration" json file for each setup, that is `hyperdebug_teacup.json`, `hyperdebug_cw310.json` (both with some additional declarations) and then two other files, each of these declare among other things, aliases for the IOA0, IOA1, etc. of the chip being tested, a UART alias called "CONSOLE" which corresponds to the pins that the ROM on the given chip sets up for boot messages, rescue, etc., a SPI alias called "BOOTSTRAP", etc.

Then in addition to that we have a "software configuration" json file for each setup, which declares that our team has chosen to put the LID_OPEN function on IOA1 on OpenTitan, while the same LID_OPEN may be on IOA7 on one of the legacy GSCs.  This way, we can have test scripts that stimulate "LID_OPEN" while watching for particular log messages in response on the CONSOLE UART, and that script does not need to branch based on the chip being tested, as the alias resolution in opentitanlib takes care of it.

This breaks down a little, when some things that should be in the latter "software configuration" needs to also depend on hardware.  For instance, when we want HyperDebug to send TPM commands via SPI, we use a SPI port alias called "TPM", this could be an alias of QSPI or SPI2, depending on how the signals are routed in hardware.  But which pin to use for chip select could in theory be a software choice, and not all users may make the same choice.  However, we have had to put this alias into the hardware json file, along with a GPIO strapping that by convention must be applied before using the TPM SPI port, this strapping is empty for all hardware setups except Teacup, for which it contains a MUX control pin to assert.

What I am trying to say, is that choices such as the JTAG pins not being available for testing should probably go into a "software json", and different sets of tests could use different software configuration files, while sharing the hardware file declaring the routing of signals on the Teacup board. (Or actually, depending on whether the test run against ASIC, CW310 or CW340, it would use one of several hardware files, with its "own" software file.)  That was, the software file could chose values for the TAP_STRAPS which may or may not enable JTAG functionality, and hence may or may not enable the use of the JTAG pins for other functional testing.)

I am not saying that the above is necessarily how it has to be, but just that doing so has worked well for my team, and I suspect that as we have more hardware variants of OpenTitan, and also more kinds of tests that do different things, that we could benefit from something similar.

PR #23276, opened 2024-05-23
[bazel] Rename `cw310_params` to `fpga_params`

jwnrt:
This is the `earlgrey_es_sival` version of #22606. See that PR for more details.


engdoreis (approved):

rules/opentitan/defs.bzl
```diff
@@ -175,15 +175,15 @@ def opentitan_test(
...
       dv: Execution overrides for a DV-based test.
       silicon: Execution overrides for a silicon-based test.
       verilator: Execution overrides for a verilator-based test.
       kwargs: Additional execution overrides identified by the `exec_env` dict.
     """
     test_parameters = {
-        "cw310": cw310,
-        "cw340": cw340,
+        "cw310": fpga,
+        "fpga": fpga,
```
tommychiu-github, line 186: Shall we add a newline for
```
"cw340": fpga
```

jwnrt: No, the `cw310` line was included as a temporary measure to prevent existing PRs that have passed their CI from causing breakages when merged. It will be removed once the current set of PRs has cleared.

If you want a different set of parameters for your CW310 and CW340 environments, you can manually include two `fpga_params` and point your execution environments at them by name instead of using `None`.

Issue #22753, opened 2024-04-22
[entropy_complex] Debug hang issues


vogelpi:
### Description

While testing #22114 for M3 and investigating a failure of the `//sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup:ft_provision_fpga_cw310_sival` test, it was discovered that under certain circumstances, the entropy complex seems to hang (see https://github.com/lowRISC/opentitan/pull/22716#issue-2251855658 for a detailed description).

Changing the reseed rate allowed us to temporarily work around the issue but we need to track down the underlying bug and make sure it's fixed for TO. What we could / should do towards this:
- Locally revert 667b717d38890a04fb2fc29065d4fdd068b23897 and rerun //sw/device/silicon_creator/manuf/skus/earlgrey_a0/sival_bringup:ft_provision_fpga_cw310_sival to make sure the test fails.
- Make use of the debug infrastructure in the entropy complex added by @h-filali for M2 and M3. In particular, the EDNs and CSRNG now properly acknowledge commands and report their status.
- Check if either EDN or CSRNG trigger an internal FIFO error.
- Check the software implementation to see if maybe software doesn't properly wait with pushing commands until the previous command is acknowledged. 
- Experiment with different reseed intervals and generate lengths to see if we can trigger the error in a top-level test that can be simulated (to investigate waves).

FYI @timothytrippel @moidx

timothytrippel:
On the software front, we could created a separate testcase that just contains the parts of the `ft_provision` test that cause the entropy complex hang up if that would be helpful? I.e., we could create a top-level test that attempts to load an OTBN program (which triggers a secure wipe of the i/d-mems; thus requesting lots of entropy) and then attempt to do a keymgr operation which requests even more entropy. We could make this TLT device-code oonly, then there would be something to run DV.

Though if there is a block level DV test that could already be used to tease out the root problem that may be quicker.

vsukhoml:
I was trying to dump entropy from entropy_src through uart for analysis. And for the case when I just read 4000000 bits (500000 words) in few long reads, each about 80KB of data - everything works fine. But when I switch to `restart` test where I read chunks of 1024/4096 bits and print them via UART i start to observe dropped symbols - hex lines becomes ragged and it is very easily reproduced after about 300-500 seconds of work. Is there any reasonable path how disable/enable entropy source (with EDN disabled) can lead to this?

I use test `entropy_src_fw_observe_many_continous` as a basis and change:

```
[...]

PR #23277, opened 2024-05-23
[chip-level/sival] Tighten csrng_edn_concurrency test

vogelpi:
This PR contains a couple of commits to tighten the csrng_edn_concurrency test to stress the entropy complex more heavily (see #22753). Also, the PR carries over a fix from the cryptolib to the dif library to avoid hangs as a result of SW commands being missed / dropped by CSRNG due to handling HW commands.

I recommend reviewing the commit individually.

### DIFF
--- a/sw/device/lib/testing/csrng_testutils.c
+++ b/sw/device/lib/testing/csrng_testutils.c
@@ -7,6 +7,7 @@
 #include "sw/device/lib/base/memory.h"
 #include "sw/device/lib/dif/dif_csrng.h"
 #include "sw/device/lib/dif/dif_csrng_shared.h"
+#include "sw/device/lib/testing/rand_testutils.h"
 #include "sw/device/lib/testing/test_framework/check.h"
 
 #include "csrng_regs.h"  // Generated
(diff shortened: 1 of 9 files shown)


vogelpi:
I've tested this on FPGA and in simulation and investigated the waves to ensure that the different CSRNG contexts are really exercised in parallel and that interfaces experience stalls etc.

sw/device/lib/testing/csrng_testutils.c
```diff
@@ -17,6 +18,58 @@ enum {
...
+    dif_csrng_entropy_src_toggle_t entropy_src_en, unsigned int glen_val,
+    dif_csrng_seed_material_t *seed_material) {
+  CHECK_STATUS_OK(
+      csrng_testutils_seed_material_build(disable_rand, seed_material));
+
+  // If disable_rand is true, we set flag0 to the provided value.
+  // If disable_rand is false, we pick a value at random.
+  dif_csrng_entropy_src_toggle_t entropy_src_enable =
+      disable_rand                       ? entropy_src_en
+      : rand_testutils_gen32_range(0, 1) ? kDifCsrngEntropySrcToggleEnable
```
moidx, line 53: One thing to note here is that if for some reason this triggers an Ibex EDN0 request, we may end up in a lockup state. 

https://github.com/lowRISC/opentitan/blob/76b1c7498c102e9cdf2fc3f7fc5db2b229e8be27/sw/device/lib/testing/rand_testutils.c#L51-L57

Something you can consider is getting all the ranges needed before stopping the entropy complex, or we need a version of `rand_test_utils_gen32_range()` that won't trigger an lfsr reseed.

moidx (approved):

sw/device/lib/testing/csrng_testutils.c
vogelpi: Thanks @moidx for your review. The assumption here is that the entropy complex is running while we call this function, similar to 
https://github.com/lowRISC/opentitan/blob/76b1c7498c102e9cdf2fc3f7fc5db2b229e8be27/sw/device/lib/testing/edn_testutils.c#L29-L31

And in the function where we call this one, we have a `entropy_testutils_auto_mode_init();` at the top. However, this auto_mode_init call is non-ideal because it always disables and re-enables the entropy complex. Meaning it again takes a long time to get the first seed from ENTROPY_SRC (takes several ms) to actually compute these parameters. Ideally, we would infer whether the entropy complex is already running in auto mode. If we can keep the ENTROPY_SRC running throughout the test, it can fill up it's seed buffer leading to less serialization due to ENTROPY_SRC and thus more pressure on the CSRNG interfaces.

moidx: Thanks @vogelpi sounds good!

PR #23301, opened 2024-05-24
[crypto] Add key/seed side-loading to P-384 curve implementation

wettermo:
This PR adds key/seed side-loading functionality to the P-384 implementation.

For this it
- adds boolean-to-arithmetic masking conversion to P-384 OTBN code
- adds key/seed side-loading for key generation in P-384 OTBN code
- adds key side-loading for P-384 ECDH and ECDSA keygen to cryptolib code on Ibex
- adds side-loaded ECDSA sign in OTBN code and Ibex cryptolib
- adds the respective functests for side-loaded P-384 ECDH and ECDSA


sw/otbn/crypto/tests/p384_keygen_from_seed_test.s
```diff
@@ -0,0 +1,105 @@
...
+  bn.rshi   w21, w31, w21 >> 128
+  bn.rshi   w11, w11, w31 >> 128
+  bn.rshi   w11, w31, w11 >> 128
+
+  /* Calculate seed = s0 ^ s1
+     [w9,w8] <= [w21,w20] ^ [w11,w10] */
+  bn.xor    w8, w20, w10
+  bn.xor    w9, w21, w11
+
+  jal       x1, p384_key_from_seed
```
jadephilipoom, line 57: nit: comment above this line that explains the seed shares get written to DMEM by this call? I did a small double take while reviewing to figure out where d0 and d1 were getting stored.

sw/otbn/crypto/tests/p384_keygen_from_seed_test.exp
```diff
@@ -0,0 +1,3 @@
+# Expected values:
+w0 = 0x0000000000000000000000000000000000000000000000000000000000000000
+w1 = 0x0000000000000000000000000000000000000000000000000000000000000000
```
jadephilipoom, line 3: nit: I worry that an all-zero expected result might cause a bug (e.g. the program `ecall`s immediately) to get missed. Perhaps instead of URND the test could use a known value, and then compare to a full expected result? The URND from the OTBN simulator is just hardcoded constants anyway iirc, so it wouldn't affect coverage. Same comment applies to the `p384_boolean_to_arithmetic_test`.

I don't think this should block the PR, so if you don't want to address it here that's fine -- although it would be nice to open an issue so we can track it as a future improvement :slightly_smiling_face:

sw/otbn/crypto/p384_ecdsa_sign.s
```diff
@@ -39,26 +65,46 @@ ecdsa_sign:
...
+  bn.wsrr   w10, KEY_S1_L
+  bn.wsrr   w11, KEY_S1_H
+
+  /* Generate secret key d in shares.
+       dmem[d0] <= d0
+       dmem[d1] <= d1 */
+  jal       x1, p384_key_from_seed
+
+  /* Tail-call signature-generation routine. */
+  jal       x1, ecdsa_sign
```
jadephilipoom, line 97: nit: since this is a tail-call and `ecdsa_sign` will hit `ecall` instead of returning, there's no need to consume a call stack slot here:
```suggestion
  jal       x0, ecdsa_sign
```

jadephilipoom (approved):
Looks great, thanks Moritz! A few small comments but nothing that should block the PR :slightly_smiling_face:

sw/otbn/crypto/tests/p384_keygen_from_seed_test.exp
wettermo: In this case, this would not only be relevant for `p384_keygen_from_seed_test` and `p384_boolean_to_arithmetic_test`, but also for `p384_keygen_test`, `p384_arithmetic_to_boolean_mod_test` and `p384_arithmetic_to_boolean_test`. All these tests expect an all-zero result. 

Therefore, as this affects a couple of tests, I would suggest to open an issue and address this in a separate PR.

sw/otbn/crypto/p384_ecdsa_sign.s
wettermo: Missed that. Thanks for pointing out!

sw/otbn/crypto/tests/p384_keygen_from_seed_test.exp
wettermo: Actually, it's also the case for `p256_arithmetic_to_boolean_mod_test` and `p384_arithmetic_to_boolean_test`.

wettermo:
Thanks for the review Jade! I addressed the nits and opened an issue regarding the expected test results (https://github.com/lowRISC/opentitan/issues/23343).

PR #23310, opened 2024-05-24
[rom] More comprehensive ROM_EXT version selection

cfrantz:
The ROM should consider the security version as well as the formal version number in selecting which slot to boot:

- Select the slot with the greater security version.
- If equal, select the slot with the greater major version.
- If equal, select the slot with the greater minor version.
- If equal, select slot A.

This allows for easier testing of ROM_EXTs and conventional roll-out policies where a newer ROM_EXT can be qualified in a testing/staging fleet before committing to an anti-rollback event.


timothytrippel (approved):

cfrantz:
Fixes #22941.

PR #23313, opened 2024-05-24
[dv,usbdev] Max traffic, streaming and randomized bus stimuli

alees24:
This PR adds sequences to:
- check back-to-back streaming of packets to single/multiple endpoints (streaming_test)
- check operations under maximum USB utilization (max_usb_traffic)
- generate randomized bus stimuli such as Suspend-Resume Signaling, Bus Resets and VBUS Disconnects during traffic; to check that the DUT is recoverable.
- the PR also addresses some faults and inadequacies of the USB driver
  1. detect and check the SE0/EOP signaling, and additionally the Idle bit interval after the SE0 bits.
  2. prevent collision between reception of DUT traffic and transmission of host packets
  3. calculate and check the CRC16 on IN DATA packets

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_driver.sv
+++ b/hw/dv/sv/usb20_agent/usb20_driver.sv
@@ -7,15 +7,10 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
 
   `uvm_component_new
 
-  // These delays are in terms of bit intervals
-  //
-  // TODO: introduce variation in the duration of the signaling, and permit these intervals to be
-  // shortened for those sequences not specifically designed to test with real-world timings;
-  // it's just wasting simulation time for many of the sequences since the DUT must respond long
-  // before the protracted Reset/Resume Signaling concludes.
-
+  // These default times are in terms of bit intervals
   int usb_rst_time = 120_000;  // Bit intervals @ 12Mbps => 10ms
   int usb_resume_time = 240_000;
+  int usb_suspend_time = 37_573;  // Bit intervals @ 12Mbps => 3.1ms, allowing for max freq delta.
   int usb_idle_clk_cycles = 5;
   int usb_pwr_good_clk_cycles = 100;  // arbitrary short delay post-VBUS assertion.
 
@@ -50,17 +45,25 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
         unique case (req_item.m_ev_type)
           // Send Bus Reset to the DUT.
           EvBusReset: begin
-            // No response required, and this is a lengthy operation; Reset Signaling is supposed
-            // to take >= 10ms under the USB 2.0 Protocol Specification.
+            // This can be a lengthy operation but it requires exclusive access to the USB;
+            // Reset Signaling is supposed to take >= 10ms under the USB 2.0 Protocol Specification.
+            bus_reset(req_item.m_ev_duration_usecs);
             seq_item_port.item_done(rsp_item);
-            bus_reset();
+          end
+          // Suspend Signaling
+          EvSuspend: begin
+            // This must be a lengthy operation and it requires exclusive access to the USB;
+            // this must be at least 3ms because the timing is hardwired into the DUT logic
+            // presently.
+            suspend_signaling(req_item.m_ev_duration_usecs);
+            seq_item_port.item_done(rsp_item);
           end
           // Resume Signaling (exit from Suspended).
           EvResume: begin
-            // No response required, and this is a lengthy operation; this must be at least 20ms
-            // for real USB hosts.
+            // This can be a lengthy operation but it requires exclusive access to the USB;
+            // this must be at least 20ms for real USB hosts.
+            resume_signaling(req_item.m_ev_duration_usecs);
             seq_item_port.item_done(rsp_item);
-            resume_signaling();
           end
           // Assert VBUS (connection to USB host).
           EvConnect: begin
@@ -339,24 +342,53 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     end
   endtask
 
-  // USB Bus Reset Task
-  // -------------------------------
-  task bus_reset();
-    // Reset bus or drive 0 on both DP and DN for 10ms
-    repeat (usb_rst_time)
+  // USB Bus Reset
+  //
+  // The duration in microseconds may be specified explicitly to speed up testing.
+  // If the default of 0 is used then the minimum specification-compliant interval of 10ms will
+  // be used.
+  task bus_reset(int unsigned duration_usecs = 0);
+    int unsigned duration_bits = usb_rst_time;
+    if (duration_usecs != 0) duration_bits = duration_usecs * 12;
+    // Reset bus (drive 0 on both DP and DN) for the specified number of bit intervals.
+    repeat (duration_bits) begin
       drive_bit_interval(1'b0, 1'b0);
-    `uvm_info(`gfn, "Reset for 10ms completed", UVM_MEDIUM)
+    end
+    `uvm_info(`gfn, $sformatf("Reset for %d bits completed", duration_bits), UVM_MEDIUM)
     // After reset change state to IDLE
-    repeat(usb_idle_clk_cycles)
+    repeat(usb_idle_clk_cycles) begin
       drive_bit_interval(1'b1, 1'b0);
+    end
+  endtask
+
+  // Suspend Signaling
+  //
+  // The duration in microseconds may be specified explicitly to speed up testing.
+  // If the default of 0 is used then an interval of 3.1ms will be used to cause the DUT to enter
+  // a Suspended state, allowing for maximum host-device frequency delta.
+  task suspend_signaling(int unsigned duration_usecs = 0);
+    int unsigned duration_bits = usb_suspend_time;
+    if (duration_usecs != 0) duration_bits = duration_usecs * 12;
+    // K state signaling required for the specified number of bit intervals.
+    repeat (duration_bits) begin
+      drive_bit_interval(1'b1, 1'b0);
+    end
+    `uvm_info(`gfn, $sformatf("Suspend Signaling for %d bits completed", duration_bits), UVM_MEDIUM)
   endtask
 
   // Resume Signaling
-  task resume_signaling();
-    // K state signaling required for at least 20ms.
-    repeat (usb_resume_time)
+  //
+  // The duration in microseconds may be specified explicitly to speed up testing.
+  // If the default of 0 is used then the minimum specification-compliant interval of 20ms will
+  // be used.
+  task resume_signaling(int unsigned duration_usecs = 0);
+    int unsigned duration_bits = usb_resume_time;
+    if (duration_usecs != 0) duration_bits = duration_usecs * 12;
+    // K state signaling required for the specified number of bit intervals.
+    repeat (duration_bits) begin
       drive_bit_interval(1'b0, 1'b1);
-    `uvm_info(`gfn, "Resume Signaling for 20ms completed", UVM_MEDIUM)
+    end
+    `uvm_info(`gfn, $sformatf("Resume Signaling for %d bits completed", duration_bits), UVM_MEDIUM)
     // _Low_ Speed EOP (SE0 for two bit intervals, J for one); LS is 1.5Mbps,
     // so 8x longer than FS signaling.
     for (int unsigned i = 0; i < 24; i++) begin
--- a/hw/dv/sv/usb20_agent/usb20_driver.sv
+++ b/hw/dv/sv/usb20_agent/usb20_driver.sv
@@ -213,6 +213,13 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     @(posedge cfg.bif.clk_i);
   endtask
 
+  // Wait for the given number of host clks.
+  // TODO: we should probably have access to the host_clk_rst_if!
+  task wait_clks(int unsigned n);
+    for (int unsigned i = 0; i < n; i++)
+      @(posedge cfg.bif.clk_i);
+  endtask
+
   task drive_pkt(bit comp_pkt[]);
     bit nrzi_out[];
     bit bit_stuff_out[];
@@ -289,9 +296,8 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     return error;
   endfunction
 
-  // NRZI Encoding/Decoding Task
+  // NRZI Encoding/Decoding Tasks
   // -------------------------------
-
   task nrzi_encoder(input bit packet[], output bit nrzi_out[]);
     bit prev_bit = 1'b1;
     nrzi_out = new[packet.size()];
@@ -331,11 +337,13 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     @(posedge cfg.bif.rst_ni);
     if (cfg.bif.active) begin
       // Idle period post DUT reset.
-      repeat (usb_idle_clk_cycles)
+      repeat (usb_idle_clk_cycles) begin
         drive_bit_interval(1'b0, 1'b0);
+      end
       cfg.bif.drive_vbus = 1'b1;
-      repeat (usb_pwr_good_clk_cycles)
+      repeat (usb_pwr_good_clk_cycles) begin
         drive_bit_interval(1'b1, 1'b0);
+      end
       // Waitfor device active state
       `DV_SPINWAIT(wait(cfg.bif.usb_dp_pullup_o);, "timeout waiting for usb_pullup", 500_000)
       bus_reset();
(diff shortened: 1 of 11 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -0,0 +1,456 @@
...
+      send_prnd_setup_packet(ep);
+      // Sending a SETUP packet clears the data toggle.
+      exp_out_toggle[ep] = 1'b0;
+    end else begin
+      // Choose the packet length and content pseudo-randomly.
+      send_prnd_out_packet(ep, exp_out_toggle[ep] ? PidTypeData1 : PidTypeData0,
+                           1'b1, 0, ep_iso_enabled[ep]);
+      inter_packet_delay();
+    end
+    // An Isochronous endpoint returns no handshake.
```
rswarbrick, line 152: I understand, but maybe it makes sense to give a bit of the background. Something like this? "Check for an ACK response from the device. This is not checked if the endpoint is in isochronous mode, because we would not expect it to return a handshake."

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -0,0 +1,456 @@
...
+      supplied = 1'b1;
+    end
+    // Update the set of buffers that are available for later use, having not been supplied to
+    // the DUT.
+    buf_avail[buf_num] = !supplied;
+    `uvm_info(`gfn, $sformatf("usbstat was 0x%0x and supplied is %d", usbstat, supplied),
+              UVM_HIGH)
+  endtask
+
+  // Keep the 'Available Buffer' FIFOs topped up.
```
rswarbrick, line 248: Maybe this description could be slightly extended, with "until..."? When I read the comment at first, I expected to find a `forever` loop.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -0,0 +1,456 @@
...
+  endtask
+
+  task device_setup;
+    // Populate the 'Available Buffer' FIFOs.
+    uint buf_num = 0;
+    for (uint buf_num = 0; buf_num < NumBuffers; buf_num++) begin
+      bit supplied;
+      device_buf_supply(buf_num, supplied);
+    end
+    // Configure the endpoints; this should strictly be done on the device side really.
```
rswarbrick, line 267: This is just setting up endpoint zero. Do we need to rephrase the comment to something like "Configure endpoint zero"?

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -0,0 +1,456 @@
...
+              UVM_MEDIUM)
+    ral.configin[ep].rdy.set(1'b0);
+    csr_update(ral.configin[ep]);
+    // Read back to see whether the packet is being collected right now.
+    csr_rd(.ptr(ral.configin[ep]), .value(configin));
+    `DV_CHECK_EQ(get_field_val(ral.configin[ep].buffer, configin), prev_buf)
+    if (get_field_val(ral.configin[ep].sending, configin)) begin
+      // The IN packet is presently being collected; since the packet size is limited to 64 bytes,
+      // there is an upper bound on how long the transmission can take. With 4x oversampling, and
+      // bit stuffing it's around 'hA30 clock cycles.
```
rswarbrick, line 291: I guess this upper bound gets translated into the `'hC00` below. Can you extend the comment to explain? ("And we're rounding up a bit" would be fine :-D)

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -0,0 +1,456 @@
...
+  endtask
+
+  // Issue a Bus Reset to the DUT.
+  virtual task send_bus_reset(int unsigned duration_usecs = 0);
+    claim_driver();
+    super.send_bus_reset(duration_usecs);
+    // Both the USB host-side and the device-side code must be informed of bus resets, in order
+    // to update their states appropriately.
+    host_side_handle_bus_reset();
+    device_side_handle_bus_reset();
```
rswarbrick, line 419: Would it make sense to run these in parallel? I think the device side can consume time? Similarly in `vbus_disconnect` below, so maybe it makes sense to put it in a task?

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -0,0 +1,456 @@
...
+    release_driver();
+  endtask
+
+  // Disconnect VBUS for the specified number of microseconds.
+  virtual task vbus_disconnect(int unsigned duration_usecs = 0);
```
rswarbrick, line 424: What does the default value of 0 microseconds mean? Maybe it makes sense to extend the comment above to say?

hw/ip/usbdev/dv/env/seq_lib/usbdev_usb_host_vseq.sv
```diff
@@ -0,0 +1,169 @@
...
+  // Count of elapsed bus frames.
+  uint frames_done = 0;
+
+  // Bit intervals remaining within the current bus frame.
+  uint frame_bits_left = 0;
+
+  //
+  bit host_stopping = 0;
+
+  //
```
rswarbrick, line 29: Missing comments here?

hw/ip/usbdev/dv/env/seq_lib/usbdev_usb_host_vseq.sv
```diff
@@ -0,0 +1,169 @@
...
+          call_sof_seq(PidTypeSofToken, frame_num);
+          host_can_transmit = 1;
+          frame_num++;
+        end
+      join_any
+      disable fork;
+    end
+  endtask
+
+  // This sequence just perform a fixed number of bus frames.
```
rswarbrick, line 72: Can you extend the description slightly to explain the return value of the function?

hw/ip/usbdev/dv/env/seq_lib/usbdev_usb_host_vseq.sv
```diff
@@ -0,0 +1,169 @@
...
+
+  // Estimate the transmission time in clock cycles for the given packet.
+  //
+  // TODO: could make this less conservative at some point because we have the data
+  //       available.
+  function uint tx_bit_intervals(byte unsigned pkt[]);
+    uint crc_bits = 16;
+    uint pid_bits = 8;  // PID bits are subject to bit stuffing too.
+    uint sync_pre = 7;  // Final bit of SYNC is a 1 and does count along with the PID bits.
+    // Worst case bit stuffing leads to 7 * (bits / 6)
```
rswarbrick, line 96: Do we need to round up the division here?

hw/ip/usbdev/dv/env/seq_lib/usbdev_usb_host_vseq.sv
```diff
@@ -0,0 +1,169 @@
...
+
+  // Transmission of packet data
+  virtual task host_bus_traffic();
+    while (!traffic_complete()) begin
+      // Prevent collision between DATA packets and SOF packet at frame start.
+      while (!host_can_transmit) begin
+        cfg.host_clk_rst_vif.wait_clks(8);
+      end
+      host_can_transmit = 0;
+      while (host_can_transmit) begin
```
rswarbrick, line 108: I don't think this `while` loop will ever run at the moment?

hw/ip/usbdev/dv/env/seq_lib/usbdev_bus_rand_vseq.sv
```diff
@@ -0,0 +1,128 @@
...
+  int unsigned max_reset_duration = 1000;
+
+  // Min/max duration of VBUS Disconnections, in microseconds.
+  int unsigned min_disconnect_duration = 10;
+  int unsigned max_disconnect_duration = 100;
+
+  // Generate a Bus Reset during streaming.
+  //
+  // Note: Bus Resets from a real host would be more than 10ms, but we typically employ a shorter
+  // delay to avoid wasting simulation time.
```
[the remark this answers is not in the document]
alees24: The comments are attached to the event-generating tasks. There's a subtlety relating to the earlier comment about the Resume Signaling; it's an annotation about the Suspend Signaling that precedes Resume Signaling. We can vary the duration of Resume Signaling, but not the Suspend Signaling sadly.

hw/ip/usbdev/dv/env/seq_lib/usbdev_bus_rand_vseq.sv
```diff
@@ -0,0 +1,128 @@
...
+    // Generate bus events randomly until asked to stop.
+    while (!traffic_stop) begin
+      bit [17:0] event_delay;
+      bit [2:0] event_type;
+      `DV_CHECK_STD_RANDOMIZE_FATAL(event_delay)
+      fork
+        cfg.clk_rst_vif.wait_clks(event_delay);
+        wait (traffic_stop);
+      join_any
+      disable fork;
```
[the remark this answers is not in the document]
alees24: Done. I had decided it didn't warrant that complication, because it's a single task enabled from a newly-forked process.

hw/ip/usbdev/dv/env/seq_lib/usbdev_usb_host_vseq.sv
alees24: Apologies for wasting your time on this accidental inclusion of an incomplete file. Modeling of real host activity with bus framing etc is rather too much work to be undertaken right now. and incidental to the test points being addressed.

hw/ip/usbdev/dv/env/seq_lib/usbdev_usb_host_vseq.sv
alees24: That is worst case. Every complete run of 6 bits could become 7 bits, but not a final run of fewer than 6 '1' bits.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
alees24: There shouldn't really be a default in that case, thanks. It doesn't make sense. I have introduced a reset duration parameter as well, since the two intervals should not necessarily be the same, and for Bus Reset there is a default.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -50,17 +45,25 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
         unique case (req_item.m_ev_type)
           // Send Bus Reset to the DUT.
           EvBusReset: begin
-            // No response required, and this is a lengthy operation; Reset Signaling is supposed
-            // to take >= 10ms under the USB 2.0 Protocol Specification.
+            bus_reset(req_item.m_ev_duration_usecs);
+            // This can be a lengthy operation but it requires exclusive access to the USB;
+            // Reset Signaling is supposed to take >= 10ms under the USB 2.0 Protocol Specification.
+            seq_item_port.item_done(rsp_item);
```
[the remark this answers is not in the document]
alees24: That doesn't work for some of the packet handling, specifically Isochronous traffic. The packet-handling code can probably be simplified in time; there's still a lot of code replication there.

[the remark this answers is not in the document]
rswarbrick: Ah, I hadn't read through carefully enough :-) Makes sense to me.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -339,31 +342,61 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
 
-  // USB Bus Reset Task
-  // -------------------------------
-  task bus_reset();
-    // Reset bus or drive 0 on both DP and DN for 10ms
-    repeat (usb_rst_time)
+  // USB Bus Reset
+  //
+  // The duration in microseconds may be specified explicitly to speed up testing.
+  // If the default of 0 is used then the minimum specification-compliant interval of 10ms will
```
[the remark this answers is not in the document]
alees24: I regard that as an internal implementation detail of the task, and would prefer not to replicate the code in the comment if only for the risk of inconsistency.

rswarbrick:
GitHub has a rather unhelpful UI when there are lots of comments! Looking through quickly, I think there are 7 comments in my review from yesterday that aren't yet resolved. Would you mind taking a look through them?

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
alees24: It's setting up all of the endpoints. An unfortunate consequence of the way our 'multireg' works is that the endpoint configuration registers are represented as an array, even though each field is a single bit per endpoint, and they all fit within a single register.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
alees24: Comment elaborated.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
alees24: Migrated the comment from within the non-Iso handling block.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
alees24: It's an interesting idea, but I can't see any reason that the host-side would have to consume time.
It can't do anything on the bus at that point.

hw/ip/usbdev/dv/env/seq_lib/usbdev_usb_host_vseq.sv
alees24: As explained, I haven't even finished sketching out this sequence yet; it's non-essential and perhaps too ambitious to model real host behavior (bus allocation and timing) presently.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
a-will: The `compact` attribute could be added to the hjson register description to put them in one CSR if they fit. Or at least... that is what it is supposed to do.

But hey, now you don't have to do read-modify-write sequences for single bit changes to all these fields ;)

rswarbrick: Ahah! I hadn't noticed that this was a multireg. Looks sensible to me.

rswarbrick (approved):
Possibly still worth looking at https://github.com/lowRISC/opentitan/pull/23313#discussion_r1620870599 and reordering, but it's no big deal. This looks good to me.

hw/ip/usbdev/dv/env/seq_lib/usbdev_bus_rand_vseq.sv
```diff
@@ -0,0 +1,128 @@
...
+  // - Bus Resets.
+  // - VBUS Disconnections.
+  task generate_bus_events();
+    // The set of events from which we may choose.
+    bit [2:0] event_mask = {do_resume_signaling, do_reset_signaling, do_vbus_disconnects};
+
+    // Report the bus events that we shall generate, and collect the min/max durations if specified
+    // in the test point.
+    if (do_resume_signaling) begin
+      `uvm_info(`gfn, "Performing Suspend-Resume Signaling during sequence", UVM_LOW)
```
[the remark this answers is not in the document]
alees24: It's ordered that way deliberately, to keep the global configuration together; those 'do_...' switches are used in other sequences. Perhaps in time they should be moved into the usbdev_base_vseq whey are actually parsed, but that's for the future...
usbdev_bus_rand_vseq may, as noted, descend from a different parent at some point, but it also has a new in-progress child,so that we can inject bad traffic as well as bus events.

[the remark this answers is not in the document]
rswarbrick: Oh sorry: I think I expressed myself quite badly. I was wondering about doing something like this:
```systemverilog
      void'($value$plusargs("min_resume_duration=%d", min_resume_duration));
      void'($value$plusargs("max_resume_duration=%d", max_resume_duration));
      `uvm_info(`gfn,
                $sformatf({"Performing Suspend-Resume Signaling during sequence ",
                           "(min_resume_duration=%0d, max_resume_duration=%0d)"},
                          min_resume_duration, max_resume_duration),
                UVM_LOW)
```

PR #23314, opened 2024-05-24
[chip-test/sival] Optimize csrng_edn_concurrency test and run more seeds in nightlies

vogelpi:
This PR contains three more commits to optimize the `csrng_edn_concurrency` test to focus on the CSRNG and EDN interaction (and reducing wait times for ENTROPY_SRC during the test loop) and it increases the number of seeds we run in the nightlies as proposed #22753 .


moidx (approved):

vogelpi:
The FPGA tests all fail but it looks like an infrastructure issue. This PR touches a file only relevant for one specific test which isn't run in CI. But the test is run in the nightlies and I am merging this now such that the next nightly regression runs the modified test with more seeds. We need this for M4.

PR #23327, opened 2024-05-27
[usbdev,sival] Make config_host test more reliable

pamaury:
This test can fail because the harness waits for the USB device but the C side will only keep the USB device until the kernel sends a configuration request. Thus if the test harness is not quick enough, it might never detect the USB device.


jwnrt (approved):
Was this test working at all previously without the `usb_harness`?

pamaury:
Yes, the reason I added the harness for cw310 is to avoid regression, see #23253

engdoreis (approved):

PR #23329, opened 2024-05-27
[integrated,sw] Switch logging over to dbg_print

sboeuf:
Log traces with `dbg_print` instead of `rom_print`.
`dbg_print` comes from OpenTitan master branch and `rom_print` has been
deprecated.


sameo (approved):

rswarbrick (commented):
This looks reasonable to me, but I think it would probably make sense for a SW person to take a look. I'm adding Chris Frantz as a reviewer, because I think he wrote `dbg_print.c` in the first place.

cfrantz:
I have no objection to you making this change, but please consider the following caveats:

1. This printf implementation is not hardened.  With the exception of `%s`, it will never follow a pointer, but there is more that can be done to harden the implementation.  There will be a hardening exercise done on this implementation at some point in the future.
2. The Earlgrey ROM is careful to exclude any of the functions from the `dbg_print` module.  There is an exception for FPGA bitstreams (ie: printing `ROM:%x`), but because of (1), Earlgrey does not want this function in the ROM.
3. `dbg_printf` is around 500 bytes.  If you are size constrained and you don't need a general printing function, you're better off crafting your output functions from `uart_write` and/or `uart_putchar`.

sameo:
> I have no objection to you making this change, but please consider the following caveats:
> 
>     1. This printf implementation is not hardened.  With the exception of `%s`, it will never follow a pointer, but there is more that can be done to harden the implementation.

Right, and I believe this is not a regression compared to the `rom_print` version.
 
> There will be a hardening exercise done on this implementation at some point in the future.

Good, we will port that part as well.

>     3. `dbg_printf` is around 500 bytes.  If you are size constrained and you don't need a general printing function, you're better off crafting your output functions from `uart_write` and/or `uart_putchar`.

Thanks for the heads up. We are not that constrained. Yet...

andreaskurth (approved):
Approving based on comments above. Thx!

Issue #22731, opened 2024-04-19
[cryptolib] Move SHA-2/HMAC computations to HMAC HWIP


ballifatih:
### Description

I want to highlight how our current hybrid SHA-2/HMAC cryptolib+HW stack looks, so we can discuss how to move from SW implementation to HW implementation. Here is the current stack:
![SHA-2_HMAC stack(1)](https://github.com/lowRISC/opentitan/assets/7850700/f68573b9-da56-4e86-8e7e-3b32e2ad1b95)

The arrows indicate the use of functions, for example HMAC-256 (streaming) is implemented with calls to SHA-256 (streaming). This means, for SHA-2/HMAC calls made to cryptolib:
- Usually one-shot calls use {init, update, final} calls from its streaming equivalent in sequence.
- We are not using HMAC functionality of HMAC HW at the moment, but only hashing.
- Except one-shot SHA-256, all use cases lead to the use of OTBN, which might create bottleneck with other OTBN use cases. AFAIU, it is required to load run_sha256/512.s here, so the user might lose the context in OTBN. For instance, that might be problematic if HMAC-based SW DRBG runs in the middle of asymmetric operation.
- I am not sure about performance, but I would expect HMAC to more efficiently handle SHA-2 and HMAC loads. Plus, we already have the new context switching functionality.

This implies multiple changes with a goal to flatten this stack and move SHA-2/HMAC computations from OTBN to HMAC HWIP. It would imply:

- [ ] Add streaming support for SHA-256 to HMAC driver (hmac.c), so that we can use the context switching supported natively by HMAC HWIP.
- [ ] Extend HMAC driver (hmac.c) to support SHA-384 and SHA-512 (with streaming functionality) and connect these directly to crypto-API-level SHA-2 calls. https://github.com/lowRISC/opentitan/issues/22547
- [ ] Extend HMAC driver (hmac.c) to support HMAC-384 and HMAC-512 (with streaming functionality) and connect these to crypto-API-level HMAC calls.
- [ ] Redirect one-shot SHA-256/384/512 and HMAC-256/384/512 calls to HMAC driver (hmac.c).

HMAC HWIP also imposes certain restrictions on the key length that we also need to be aware of:
https://github.com/lowRISC/opentitan/blob/ec711524c05e92a306b3e28c0343522909a1b9a7/hw/ip/hmac/README.md?plain=1#L28

[...]

Issue #23191, opened 2024-05-19
[cryptolib] HMAC driver TODO list


ballifatih:
### Description

(this issue is WIP)
- [ ] Consider whether we want to have a timeout value for polling for interrupt signal. HMAC does not use EDN, so we should be able to determine a reasonable timeout value.

martin-velay:
The following is addressed by this PR (once merged): https://github.com/lowRISC/opentitan/pull/24839

"Consider whether we want to have a timeout value for polling for interrupt signal. HMAC does not use EDN, so we should be able to determine a reasonable timeout value."

johannheyszl:
IIUC, we have completed all tasks that are security hardening related, deprioritizing. cc @nasahlpa

nasahlpa:
Yes, the SCA and FI hardening of HMAC is in a good shape. No open security hardening tasks I am aware of.

johannheyszl:
Closing after discussion

PR #23335, opened 2024-05-27
[HMAC, cryptolib] Connect all SHA-256/384/512 and HMAC-256/384/512 to HMAC driver.

ballifatih:
This PR connects previously merged HMAC driver to cryptolib API. It means that all SHA-2/HMAC variants are routed to HMAC HWIP and they are not connected to the OTBN-based implementation. See #22731 for more context.

This PR includes multiple commits, where I arranged the commits with the hope to reduce reviewing effort:

- HMAC driver improvements:
  - Remove key endianness swapping and use freshly introduced swap_key flag instead (from #23242).
  - Remove redundant HMAC-driver-specific `hmac_digest_t`, `hmac_key_t` types and use generic `*word32_buf_t` structs instead.
  - Add placeholder oneshot call so that we can improve these later. This will later be replaced with a truly one-shat call implemented in the driver, that does not incur multiple pause/store/restore costs per block. At this point, it is there to have cleaner `otcrypto_hash` and `otcrypto_hmac` functions at cryptolib level.
- Added a function to `keyblob.h/c` that unmasks a given key. I am open to suggestions here, but I think it makes sense to have it because we have blinded->unblinded downgrading, when the underlying implementation does not support masking scheme.
- Connect `otcrypto_hash_*` and `otcrypto_hmac_*` calls to HMAC driver. The one-shot and streaming calls are simply wrappers to HMAC calls with similar names + few error checks.
  - I could use careful review on how I convert `otcrypto_hash_context_t`, `otcrypto_hmac_context_t` to driver-level `hmac_ctx_t`. Besides these structs, I implemented `hmac_ctx_save` and `hmac_ctx_restore` calls in `hmac.c` and `hash.c`. Unlike before, I decided to use the same context for all operations. This part is open to suggestions.
 - Testing support PR that includes python scripts to generate arbitrary-size HMAC vectors, as well as hardcoded hjson file generated from. We can drop this commit if needed, but I would keep it around for the sake of easier development, and follow-up bug fixing.


sw/device/lib/crypto/impl/keyblob.h
```diff
@@ -164,6 +164,26 @@ status_t keyblob_from_key_and_mask(const uint32_t *key, const uint32_t *mask,
 OT_WARN_UNUSED_RESULT
 status_t keyblob_remask(otcrypto_blinded_key_t *key, const uint32_t *mask);
 
+/**
+ * Unmask and return the effective key from blinded key struct.
+ *
+ * This function should be used only for situtations where the underlying
```
moidx, line 170: typo: situations.

sw/device/lib/crypto/impl/mac.c
```diff
@@ -9,23 +9,109 @@
...
-  HARDENED_TRY(otcrypto_hmac_update(&ctx, input_message));
-  return otcrypto_hmac_final(&ctx, tag);
+  // Validate key struct.
+  if (key == NULL || key->keyblob == NULL || tag.data == NULL) {
+    return OTCRYPTO_BAD_ARGS;
+  }
+  if (key->config.hw_backed != kHardenedBoolFalse) {
+    // TODO(#15590): Add support for sideloaded keys via a custom OTBN program.
+    return OTCRYPTO_NOT_IMPLEMENTED;
+  }
```
moidx, line 85: I assume this is no longer valid as we are now using the `hmac` block.

sw/device/lib/crypto/impl/mac.c
```diff
@@ -9,23 +9,109 @@
...
+  if (key->config.security_level != kOtcryptoKeySecurityLevelLow) {
+    // TODO: Harden SHA2 implementations.
+    return OTCRYPTO_NOT_IMPLEMENTED;
+  }
+
+  // Check for null input message with nonzero length.
+  if (input_message.data == NULL && input_message.len != 0) {
+    return OTCRYPTO_BAD_ARGS;
+  }
```
moidx, line 94: Is there a test vector for the empty string?

sw/device/lib/crypto/impl/mac.c
ballifatih: Removing sideload and masked implementation TODOs and adding the following comment line:
```
// HMAC HWIP is not masked and it does not have sideload support, so the
// following conditions return not implemented.
```

sw/device/lib/crypto/impl/mac.c
ballifatih: Yes, we have one here:
https://github.com/lowRISC/opentitan/blob/2a42ee82b39b4134dac8cba2e8e418e808bf906d/sw/device/tests/crypto/hmac_sha256_functest.c#L112-L122

ballifatih:
With the last push, `//sw/device/tests/crypto/cryptotest:hmac_kat_fpga_cw340_test_rom` is passing (with some local bazel changes for replacing cw310 with cw340).

sw/device/lib/crypto/impl/mac.c
```diff
@@ -9,23 +9,202 @@
...
+  // according to FIPS 198-1, Section 4, Step 2.
+  if (key->config.key_length > block_size * sizeof(uint32_t)) {
+    otcrypto_hash_digest_t key_digest = {
+        .mode = hash_mode,
+        .data = processed_key,
+        .len = digest_size,
+    };
+    otcrypto_const_byte_buf_t msg_buf = {
+        .len = key->config.key_length,
+        .data = (uint8_t *)unmasked_key,
```
jadephilipoom, line 150: nit: technically this violates strict aliasing, but `unsigned char *` is legal and equivalent:
```suggestion
        .data = (unsigned char *)unmasked_key,
```

sw/device/lib/crypto/impl/hash.c
```diff
@@ -9,125 +9,57 @@
 #include "sw/device/lib/base/hardened_memory.h"
 #include "sw/device/lib/crypto/drivers/hmac.h"
 #include "sw/device/lib/crypto/drivers/kmac.h"
-#include "sw/device/lib/crypto/impl/sha2/sha256.h"
-#include "sw/device/lib/crypto/impl/sha2/sha512.h"
```
jadephilipoom, line 13: I think these files can now be deleted.

sw/device/lib/crypto/impl/hash.c
```diff
@@ -9,125 +9,57 @@
...
  *
  * @param ctx Generic hash context to restore from.
- * @param[out] state Destination SHA-512 context object.
+ * @param[out] hmac_ctx Destination HMAC driver context object.
  */
-static void sha512_state_restore(const otcrypto_hash_context_t *restrict ctx,
-                                 sha512_state_t *restrict state) {
-  // As per the `hardened_memcpy()` documentation, it is OK to cast to
-  // `uint32_t *` here as long as `state` is word-aligned, which it must be
-  // because all its fields are.
```
jadephilipoom, line 128: I'd leave this comment in; that cast is normally unsafe, so it's good to record the justification for why it's OK here. Same for the comment in `save` above.

sw/device/lib/crypto/impl/mac.c
```diff
@@ -145,107 +324,21 @@ otcrypto_status_t otcrypto_hmac_init(otcrypto_hmac_context_t *ctx,
...
-  // Compute (K0 ^ opad).
-  uint32_t outer_block[message_block_words];
-  for (size_t i = 0; i < message_block_words; i++) {
-    outer_block[i] = k0[i] ^ 0x5c5c5c5c;
-  }
+  hmac_mode_t hmac_mode;
+  size_t block_size;
+  HARDENED_TRY(get_hmac_mode(key->config.key_mode, &hmac_mode, &block_size));
+
+  uint32_t procesed_key[block_size];
```
jadephilipoom, line 331: nit: typo
```suggestion
  uint32_t processed_key[block_size];
```

sw/device/tests/crypto/hmac_functest.c
```diff
@@ -0,0 +1,127 @@
...
+    default:
+      return OTCRYPTO_BAD_ARGS;
+  }
+  TRY_CHECK_ARRAYS_EQ(act_tag, current_test_vector->digest.data, digest_len);
+  return OTCRYPTO_OK;
+}
+
+OTTF_DEFINE_TEST_CONFIG();
+bool test_main(void) {
+  LOG_INFO("Testing cryptolib HMAC driver.");
```
jadephilipoom, line 116: This is really more testing the HMAC top-level API than the driver itself, no?

sw/device/lib/crypto/drivers/hmac.h
```diff
@@ -132,19 +124,20 @@ typedef enum hmac_mode {
...
  * @param[out] ctx Initialized context object for SHA2/HMAC-SHA2 operations.
  * @param hmac_mode Specifies the mode among SHA2-256/384/512, HMAC-256/384/512.
  * @param key HMAC key. The key to be used with HMAC calls.
- * @return Error status.
+ * @return Result of the operation.
  */
 OT_WARN_UNUSED_RESULT
 status_t hmac_init(hmac_ctx_t *ctx, const hmac_mode_t hmac_mode,
-                   const hmac_key_t *key);
+                   otcrypto_const_word32_buf_t *key);
```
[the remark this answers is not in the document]
ballifatih: Sounds fine, reverting to `(uint32_t*, size_t)` pairs for both key and the digest.

sw/device/lib/crypto/impl/mac.c
ballifatih: My understanding on the strict aliasing rules is limited, so I will go with your suggestion here. Thanks.

sw/device/lib/crypto/impl/hash.c
ballifatih: Do we want to keep them around, in case they are used for hardening SHA-2/HMAC through OTBN?

sw/device/lib/crypto/impl/hash.c
ballifatih: Added the same comment to all four functions (as there are duplicates of the same functions in `mac.c` and `hash.c`).

sw/device/tests/crypto/hmac_functest.c
ballifatih: Correct, updating this as `Testing cryptolib SHA-2/HMAC implementations.`.

sw/device/lib/crypto/impl/hash.c
ballifatih: Added to follow-up tracker: https://github.com/lowRISC/opentitan/issues/23191

moidx (approved):

PR #23336, opened 2024-05-27
[hmac/rtl] Fix idle when HMAC stopped

martin-velay:
- Allow HMAC to go into idle after a stop on block any boundaries

### DIFF
--- a/hw/ip/hmac/rtl/hmac_core.sv
+++ b/hw/ip/hmac/rtl/hmac_core.sv
@@ -79,6 +79,9 @@ module hmac_core import prim_sha2_pkg::*; (
 
   logic hmac_sha_rvalid;
 
+  logic idle_d, idle_q;
+  logic reg_hash_stop_d, reg_hash_stop_q;
+
   typedef enum logic [1:0] {
     SelIPad,
     SelOPad,
@@ -231,11 +234,12 @@ module hmac_core import prim_sha2_pkg::*; (
 
   assign sha_message_length_o = sha_msg_len;
 
+  // asserting txcnt_eq_blksz when txcount is a multiple of block size
   always_comb begin
     unique case (digest_size_i)
-      SHA2_256: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA256:0] == BlockSizeBSBSHA256);
-      SHA2_384: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA512:0] == BlockSizeBSBSHA512);
-      SHA2_512: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA512:0] == BlockSizeBSBSHA512);
+      SHA2_256: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA256-1:0] == '0) && (txcount != '0);
+      SHA2_384: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA512-1:0] == '0) && (txcount != '0);
+      SHA2_512: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA512-1:0] == '0) && (txcount != '0);
       default : txcnt_eq_blksz = '0;
     endcase
   end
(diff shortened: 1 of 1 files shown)


martin-velay:
It refers to this issue #22916

hw/ip/hmac/rtl/hmac_core.sv
```diff
@@ -233,9 +236,9 @@ module hmac_core import prim_sha2_pkg::*; (
 
   always_comb begin
     unique case (digest_size_i)
-      SHA2_256: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA256:0] == BlockSizeBSBSHA256);
-      SHA2_384: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA512:0] == BlockSizeBSBSHA512);
-      SHA2_512: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA512:0] == BlockSizeBSBSHA512);
+      SHA2_256: txcnt_eq_blksz = (txcount[BlockSizeBitsSHA256-1:0] == '0) && (txcount != '0);
```
gdessouky, line 239: I would only add a comment above `// asserting txcnt_eq_blksz at txcount == multiples of block size` or something like that

hw/ip/hmac/rtl/hmac_core.sv
```diff
@@ -434,15 +437,27 @@ module hmac_core import prim_sha2_pkg::*; (
...
   // Idle status signaling: This module ..
-  logic idle_d, idle_q;
   assign idle_d =
       // .. is not idle when told to start or continue
       (reg_hash_start_i || reg_hash_continue_i) ? 1'b0 :
       // .. is idle when the FSM is in the Idle state
       (st_q == StIdle) ? 1'b1 :
       // .. is idle when it has processed a complete block of a message and is told to stop
-      (st_q == StMsg && txcnt_eq_blksz && reg_hash_stop_i) ? 1'b1 :
+      (st_q == StMsg && txcnt_eq_blksz && reg_hash_stop_d) ? 1'b1 :
```
gdessouky, line 460: shouldn't this be `reg_hash_stop_q` here?

gdessouky:
thanks @martin-velay for catching, this LGTM modulo the comments I've left above, and so long as it doesn't break our existing block-level DV (it shouldn't from my inspection of the RTL changes).

hw/ip/hmac/rtl/hmac_core.sv
martin-velay: You're right, I'll

hw/ip/hmac/rtl/hmac_core.sv
martin-velay: I think `reg_hash_stop_d` is fine, I don't need the delayed version of it and it's using itself the memorized `reg_hash_stop_q` version. But maybe there is a coding rule or a good practice I ignore?

martin-velay:
My regression is still ongoing, but so far I don't see this issue in the log anymore.

hw/ip/hmac/rtl/hmac_core.sv
ballifatih: Can we not use `reg_hash_stop_i` directly here and remove `reg_hash_stop_d/q` signals?

```
...
(st_q == StMsg && txcnt_eq_blksz && reg_hash_stop_i) ? 1'b1
...
```

`hash_stop` bit from the register is hwext type, so this is basically a pass through. So when "stop" command is written through TL-UL, `reg_hash_stop_i` should give us a single clock cycle pulse.

https://github.com/lowRISC/opentitan/blob/1459afe788072bd95e185225f2b0179b7aed9065/hw/ip/hmac/rtl/hmac_reg_top.sv#L751-L764

I think it would be fair to force SW to first write full blocks of messages, and only then call "stop". Or do we want to support the case where "stop" is issued before the full block is complete?

Also, if we eventually add a transition `StMsg -> StIdle` in `st_q`'s FSM, then `st_q == StIdle` condition can make this particular condition line irrelevant (more context in https://github.com/lowRISC/opentitan/issues/23352).

ballifatih:
On the waveform, `Idle` signals looks correct for HMAC operations, but during SHA-384 operation, `Idle` goes to 0 for a single clock cycle only, which I found a bit weird. This is the clock cycle `start` is issued. It might be that `idle` signal is not correctly evaluated at `hmac.sv` level. 

Zoomed-out activity during SHA-384 operation:
image

hw/ip/hmac/rtl/hmac_core.sv
```diff
@@ -434,15 +438,27 @@ module hmac_core import prim_sha2_pkg::*; (
     endcase
   end
 
+  assign reg_hash_stop_d =
+      (reg_hash_stop_i == 1'b1) ? 1'b1 :
+      (idle_q          == 1'b1) ? 1'b0 :
```
andreaskurth, line 443: The `== 1'b1` on these two lines can be removed

hw/ip/hmac/rtl/hmac_core.sv
```diff
@@ -434,15 +438,27 @@ module hmac_core import prim_sha2_pkg::*; (
...
+      reg_hash_stop_q;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      reg_hash_stop_q <= 1'b1;
```
andreaskurth, line 448: Why is the reset value not zero? It will get set to zero after the first clock cycle, no?

hw/ip/hmac/rtl/hmac_core.sv
andreaskurth: Right, but if that pulse comes before `txcnt_eq_blksz` is true, this condition will not be met, right? Thus I also think we should latch `reg_hash_stop_i`.

Using the `_d` signal for this is fine.

Re SW having to write a full block before calling stop: Yes, this is the currently defined programming model. @martin-velay: Does the test that led you to this problem follow the programming model? Or does it call stop before having provided data for a full block?

ballifatih: I see, considering `txcnt_eq_blksz` might come delayed, it makes sense to latch `reg_hash_stop_i`.

andreaskurth:
There's a _quick lint_ error message saying
> The second line of a commit message must be empty, as it separates the summary from the long description

Once that's fixed and CI passes, and provided that this change passes a block-level regression (I'd recommend `-rx 0.2` at least), I think we can merge this

andreaskurth:
This PR got superseded by #23432, right? Should we close it?

PR #23344, opened 2024-05-28
[sival] Remove `pwrmgr_sleep_wake_5_bug_test`

jwnrt:
This test was created to reproduce a failure in the ES version of the silicon. It has been fixed in later versions of the chip and this test remains in those versions of the repository.

Removing from this branch since we don't want permanently failing tests.


moidx:
Discussed in SiVal sync. The recommendation is to remove the test case since it doesn't add any value to the `earlgrey_es_sival` branch.

moidx (approved):

Issue #22778, opened 2024-04-23
[Earlgrey-ES, sw, sensor_ctrl] Mask flaky alerts


andreaskurth:
sensor_ctrl's alerts 5 and 10 are known to be flaky (see #20678 and #20798) on Earlgrey-ES.

I see two options for dealing with this:
<ol type="a">
  <li>In each test/program for which these alerts have an impact, clear those alerts when they happen (might be at test initialization or repeatedly during the test). This has been done in #20812, for example, and is being proposed again in #22723. This means that multiple tests will contain ES-specific code.</li>
  <li>In ROM_EXT for ES -- and only for ES -- clear those alerts and make them fatal instead of the default recoverable, then disable and lock the SensorCtrlFatalAlert source in alert_handler (keeping the corresponding `alert_en` at MuBiFalse while clearing the corresponding `alert_regwen`). This means that ROM_EXT will contain ES-specific code but individual tests/programs don't have to deal with the flaky alerts. It also means we cannot test alerts classified as fatal from sensor_ctrl in SiVal, but as the logic in alert_handler is the same for all sources and those can still be tested, I don't think this is a problem.</li>
</ol>

andreaskurth:
I'm closing this as no longer relevant. If anyone disagrees, please re-open the issue.

PR #23345, opened 2024-05-28
[Sival, pwrmgr] Map test chip_sw_pwrmgr_sleep_sensor_ctrl_alert_wakeup

engdoreis:
This PR cherry-picks from https://github.com/lowRISC/opentitan/pull/22723 to fix the test `chip_sw_sensor_ctrl_ast_alerts` as described in https://github.com/lowRISC/opentitan/issues/22778.
I believe we should get this issue fixed on the test side, at least for now, and remove the change in case the fix is moved to the ROM_EXT.


moidx (approved):

jwnrt:
@engdoreis would you mind handling these backports?

PR #23349, opened 2024-05-28
[csrng/dv] Add DV for the force fips bit functionality

h-filali:
At some point we should probably add some coverage for the new registers. I can do so once I'm back.
Rarely there are still some failures I wasn't able to investigate yet. When I ran a 10% regression I found 1 failure due to the added DV in the csrng_cmds test.

This is the DV for #23324

### DIFF
--- a/hw/ip/csrng/dv/env/csrng_scoreboard.sv
+++ b/hw/ip/csrng/dv/env/csrng_scoreboard.sv
@@ -346,6 +346,8 @@ class csrng_scoreboard extends cip_base_scoreboard #(
           );
         end
       end
+      "fips_force": begin
+      end
       "hw_exc_sts": begin
       end
       "recov_alert_sts": begin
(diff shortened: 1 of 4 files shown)


vogelpi (approved):
Thanks @h-filali , this looks great! Once PR #23324 is merged, I'll rebase this one and can then merge it as well.

vogelpi:
Update: https://github.com/lowRISC/opentitan/pull/23324 got merged, I've rebased this PR now.

hw/ip/csrng/dv/env/csrng_scoreboard.sv
```diff
@@ -476,7 +480,8 @@ class csrng_scoreboard extends cip_base_scoreboard #(
     cfg.key[app] = 'h0;
     cfg.v[app]   = 'h0;
     cfg.reseed_counter[app] = 1'b0;
-    cfg.compliance[app]     = 1'b0;
+    cfg.compliance[app]     =  ((`gmv(ral.ctrl.fips_force_enable) == MuBi4True) &&
+                                 `gmv(ral.fips_force)[app]);
```
vogelpi, line 484: I only realized after merging that this change here is actually wrong. After an Uninstantiate command, the compliance bit is always 0 (and should be 0).

vogelpi: The PR to fix this is here: #23602

PR #23354, opened 2024-05-29
[rom] Initialize output argument when loading boot_data

### LINKED ISSUE #22927: [rom,refactor] Potentially uninitialized parameter when loading boot_data.
### Description

The `boot_data_read function` expects an output-only parameter `*boot_data`. 
https://github.com/lowRISC/opentitan/blob/605da270ddcf70b2af5573e9524f9613bc89383d/sw/device/silicon_creator/lib/boot_data.h#L181-L185

The pointer is then passed to `boot_data_page_info_update` which expects an in-out parameter, and the value is potentially uninitialized.
https://github.com/lowRISC/opentitan/blob/605da270ddcf70b2af5573e9524f9613bc89383d/sw/device/silicon_creator/lib/boot_data.c#L436-L443

---

One possible fix is to memset the `*boot_data` structure to all zeros in `boot_data_active_page_find`.
https://github.com/lowRISC/opentitan/blob/605da270ddcf70b2af5573e9524f9613bc89383d/sw/device/silicon_creator/lib/boot_data.c#L468-L473

sasdf:
Fix #22927.

Initialize the boot_data struct to all zeros before calling `boot_data_page_info_update` which expects an in-out argument.

### DIFF
--- a/sw/device/silicon_creator/lib/boot_data.c
+++ b/sw/device/silicon_creator/lib/boot_data.c
@@ -481,6 +481,7 @@ static rom_error_t boot_data_active_page_find(active_page_info_t *page_info,
 
   static_assert(kPageCount == 2,
                 "Number of pages changed, unrolled loop must be updated");
+  memset(boot_data, 0, sizeof(*boot_data));
   HARDENED_RETURN_IF_ERROR(
       boot_data_page_info_update(kPages[0], page_info, boot_data));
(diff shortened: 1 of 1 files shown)


a-will (changes requested):

sasdf:
I missed that precondition.  Thanks for reviewing!

PR #23356, opened 2024-05-29
[ast] adding spi or iob mux

meisnere:
[ast] adding spi or iob mux as part of #23201

Adding #23125 changes as it wasn't approved the first time for some reason.

### DIFF
--- a/hw/top_earlgrey/ip/ast/rtl/ast.sv
+++ b/hw/top_earlgrey/ip/ast/rtl/ast.sv
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 // -------- W A R N I N G: A U T O - G E N E R A T E D  C O D E !! -------- //
-// PLEASE DO NOT HAND-EDIT THIS FILE. IT HAS BEEN AUTO-GENERATED.
+// PLEASE DO NOT HAND-EDIT THIS FILE. IT HAS BEEN AUTO-GENERATED (by closed source generator).
 //
 //############################################################################
 // *Name: ast
@@ -18,7 +18,7 @@ module ast #(
   parameter int unsigned EntropyStreams  = 4,
   parameter int unsigned UsbCalibWidth   = 20,
   parameter int unsigned Ast2PadOutWidth = 9,
-  parameter int unsigned Pad2AstInWidth  = 9
+  parameter int unsigned Pad2AstInWidth  = 8
 ) (
   // tlul if
   input tlul_pkg::tl_h2d_t tl_i,              // TLUL H2D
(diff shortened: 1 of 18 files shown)


hw/top_earlgrey/ip/ast/rtl/ast_pkg.sv
```diff
@@ -36,8 +36,8 @@ parameter int unsigned Ot3Sel     = 8;
 parameter int unsigned Ot4Sel     = 9;
 parameter int unsigned Ot5Sel     = 10;
 //
-parameter int unsigned Lc2HcTrCyc = 102;  // ((99+1)+(3+1))x5 = 520 us
-parameter int unsigned Hc2LcTrCyc = 38;   // ((35+1)+(3+1))x5 = 200 us
+parameter int unsigned Lc2HcTrCyc = 112;  // ((109+1)+(3+1))x5 = 570 us
+parameter int unsigned Hc2LcTrCyc = 142;  // ((139+1)+(3+1))x5 = 720 us
```
andreaskurth, line 40: Is this change intended? It doesn't seem to be related to the SPI/IOB mux. If it's intended, I suggest creating a separate commit with a short description and move that to a separate PR

andreaskurth: Could this cause the `chip_sw_pwrmgr_smoketest` test to fail (see CI logs)?

andreaskurth (changes requested):

andreaskurth:
The linter run by kokoro currently fails because the port added to ast is not listed in the instance inside `top_earlgrey`. I can amend the commit to resolve this, but I'd prefer to first resolve the problem I mentioned above

hw/top_earlgrey/ip/ast/rtl/ast.sv
```diff
@@ -18,7 +18,7 @@ module ast #(
   parameter int unsigned EntropyStreams  = 4,
   parameter int unsigned UsbCalibWidth   = 20,
   parameter int unsigned Ast2PadOutWidth = 9,
-  parameter int unsigned Pad2AstInWidth  = 9
+  parameter int unsigned Pad2AstInWidth  = 8
```
andreaskurth, line 21: This change LGTM because `PAD2AST_WIRES`, which is used in the assignment to the `padmux2ast_i` port of `ast` contains 8 signals.

hw/top_earlgrey/ip/ast/rtl/ast_pkg.sv
andreaskurth: Resolved

PR #23357, opened 2024-05-29
[hw,dma] Sync DMA from integrated_dev to master

Razer6:
In the interest of syncing the `integrated_dev` branch back to master, I looked into the options. For changes, that are isolated to specific folders, git provides ways to cherry pick the whole history using the following command:
 
```bash
git format-patch -k --stdout integrated_dev...master -- hw/ip/dma | git am -3 -k
```
To reduce the divergence of both branches, I tried that for the DMA folder and it seems to work quite well.


a-will:
We'll probably want to do a little cleaning of the history, but that looks like a neat way to bring in the new IPs.

andreaskurth:
Interesting, thx for this exploration, @Razer6 :+1: I don't see an obvious blocker for this approach, but I currently can't review it in detail

Razer6:
Superseded by https://github.com/lowRISC/opentitan/pull/24434

PR #23366, opened 2024-05-29
[rv_dm,dv] Widen exclusion for dmi register in jtag_dtm_reg_block

rswarbrick:
The existing code already excluded the "op" field of the register, but we actually get other mismatches as well because the "address" and "data" fields have values based on the last DMI operation. Exclude the whole register.


marnovandermaas (approved):
This looks reasonable to me. I assume this `csr_excl` is something the DV code uses to automatically generate checks.

It also aligns with my reading of section: "6.1.5 Debug Module Interface Access (dmi, at 0x11)"

PR #23369, opened 2024-05-29
[util] Get rid of "type" field in Modes

rswarbrick:
The existing python code was using a string to tell the base class how to look up sub-modes for an instance of a given derived class. This was a bit complicated!

There are actually only two places where this does anything anyway, so duplicate them and get rid of the string magic.

### DIFF
--- a/util/dvsim/Regression.py
+++ b/util/dvsim/Regression.py
@@ -20,7 +20,6 @@ class Regression(Mode):
 
     def __init__(self, regdict):
         self.name = ""
-        self.type = ""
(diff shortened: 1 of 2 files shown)


hcallahan-lowrisc (approved):

andreaskurth:
@rswarbrick I think this isn't required to be part of the release branch, right?

rswarbrick:
That's correct: it's just a tidy-up (from May!) in dvsim. No particular relation to the release.

PR #23370, opened 2024-05-29
[prim,fpv] Avoid warning message from Jasper with ASSERT_AT_RESET

rswarbrick:
No functional change: the warning message from Jasper was saying that the $isunknown(__rst) was being treated as "0", which is the behaviour we're making explicit.

But now we don't get a warning message, which makes genuine problems a little more obvious.


andreaskurth (approved):
LGTM, thx for this tool-related fix, @rswarbrick :+1:

PR #23373, opened 2024-05-29
[rv_dm,dv] Improve the rv_dm_jtag_dmi_dm_inactive vseq

rswarbrick:
This isn't really meant to make big changes, but the vseq becomes a bit less magical (fewer magic hex constants), and it's a bit more obvious what it's doing. Descriptions in the commit messages (the last one is the main commit).


marnovandermaas (approved):
Cool bit of logic for the randomized writes while DMI is inactive.

PR #23374, opened 2024-05-29
[jtag,dv] Demote some driver messages to UVM_HIGH

rswarbrick:
This was very chatty and meant that rv_dm tests with -vm weren't very
easy to interpret. Drop the messages to UVM_HIGH accordingly.

(The PR looks bigger than it really is, because it depends on #23097. Once that is merged, this is just the last commit)


matutem (approved):
Yes, jtag was too chatty. Thanks.
LGTM

Issue #21630, opened 2024-02-22
[rom] Reading resistor straps should differentiate weak versus strong pulls


jettr:
If the ROM only performs a single read operation on the resistor strapping pins, then it will not be able to tell the difference between weak (e.g. 1M) and strong (e.g. 5.1k) pull resistors. This effectively eliminates the possibility of using the weak pulls values for the owner firmware if that aliases to a known ROM bootstrap combination.

https://github.com/lowRISC/opentitan/blob/2714197f3235f1737937ad20421a3f32ff4a7030/sw/device/silicon_creator/rom/bootstrap.c#L46

Right now there are 2 resistor straps combinations on IOC0, ICOC1, and ICO2:

[SW_STRAP_BOOTSTRAP](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/base/chip.h;drc=1ff6f6ea02a1711224a284a8855264406750eabd;bpv=1;bpt=1;l=98?gsn=SW_STRAP_BOOTSTRAP&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fgithub.com%2FlowRISC%2Fopentitan%3Flang%3Dc%252B%252B%3Fpath%3Dsw%2Fdevice%2Fsilicon_creator%2Flib%2Fbase%2Fchip.h%23SW_STRAP_BOOTSTRAP%2523m%25402500): Strong PUp,  Strong PUp,  Strong PUp

and 

[SW_STRAP_RMA_ENTRY](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/base/chip.h;drc=1ff6f6ea02a1711224a284a8855264406750eabd;bpv=1;bpt=1;l=89?gsn=SW_STRAP_RMA_ENTRY&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fgithub.com%2FlowRISC%2Fopentitan%3Flang%3Dc%252B%252B%3Fpath%3Dsw%2Fdevice%2Fsilicon_creator%2Flib%2Fbase%2Fchip.h%23SW_STRAP_RMA_ENTRY%2523m%25402241): Strong PDown, Strong PUp, Strong PUp

However if Owner Firmware wants to use  Weak PUp, Weak PUp, Weak PUp, then the ROM will incorrectly think that the `SW_STRAP_BOOTSTRAP` combination is getting triggered.

[...]

PR #23376, opened 2024-05-29
[rom] Document bootstrap detection

cfrantz:
Addresses #21630

### DIFF
--- a/sw/device/silicon_creator/rom/bootstrap.c
+++ b/sw/device/silicon_creator/rom/bootstrap.c
@@ -44,7 +44,9 @@ hardened_bool_t bootstrap_requested(void) {
   HARDENED_CHECK_NE(bootstrap_dis, kHardenedBoolTrue);
 
   // A single read is sufficient since we expect strong pull-ups on the strap
-  // pins.
+  // pins.  We assume pinmux has already been configured (by pinmux_init) to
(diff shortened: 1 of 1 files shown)


sw/device/silicon_creator/rom/bootstrap.c
```diff
@@ -44,7 +44,9 @@ hardened_bool_t bootstrap_requested(void) {
   HARDENED_CHECK_NE(bootstrap_dis, kHardenedBoolTrue);
 
   // A single read is sufficient since we expect strong pull-ups on the strap
-  // pins.
+  // pins.  We assume pinmux has already been configured (by pinmux_init) to
+  // enable the internal pull-downs on the strap pins.  As such, an external
+  // weak pull-up will not be detected as a logic-high.
   uint32_t res = launder32(kHardenedBoolTrue) ^ SW_STRAP_BOOTSTRAP;
```
moidx, line 50: nit: I believe we also want to condition entry on POR reset. Which we may already be doing. We could add this to the comment for future reference.

moidx (approved):

sw/device/silicon_creator/rom/bootstrap.c
cfrantz: Checking the wakeup state is already being done in `rom_main`.

cfrantz:

PR #23378, opened 2024-05-29
[rom] add load access fault exception handler

timothytrippel:
This adds a simple exception handler to the ROM that ONLY handles load access faults that are a result of an ECC integrity error; the exception handler should simply clear the fault and return (any loads that did not complete will have corrupted values that will poison signature verification).

Additionally, this enhances the existing `boot_policy_flash_ecc_error_test` to test corrupting:
    - all manifest fields that are accessed via pointer deferences in the ROM, and
    - code words in at the beginning, middle, and end of the code section of a ROM_EXT image

This partially addresses #21353.

Note: this depends on #23377, only review the last 5 commits.


sw/device/silicon_creator/rom/rom.c
```diff
@@ -672,11 +672,27 @@ void rom_interrupt_handler(void) {
   OT_UNREACHABLE();
 }
 
+void rom_exception_handler(void) {
+  uint32_t mcause;
+  CSR_READ(CSR_REG_MCAUSE, &mcause);
+  ibex_exception_code_t exception_code =
+      (ibex_exception_code_t)(mcause & kIbexExceptionCodeMax);
```
jettr, line 679: Should Max be name Mask instead?

sw/device/silicon_creator/rom/rom.c
```diff
@@ -672,11 +672,27 @@ void rom_interrupt_handler(void) {
...
+
+  if (launder32(exception_code) == kIbexExceptionCodeLoadAccessFault) {
```
jettr, line 681: Can you add a comment saying that this is an "expected" fault when reading flash with invalid ECC, and we just return execution back to code that tried to perform read operation.

jettr (commented):
I missed where the actual read is happening and the poison values present. Is that already landed?

sw/device/silicon_creator/rom/rom.c
cfrantz: The handling of the flash ECC error has been re-written as pure assembly.
- It's shorter and consumes fewer precious ROM bytes.
- It's faster.
- There are fewer jumps and no C setup/teardown.

sw/device/silicon_creator/rom/rom.c
cfrantz: It's both the Max exception code _and_ the Mask.  And moot, since the handler is now in assembly language :)

sw/device/silicon_creator/rom/rom_isrs.S
```diff
@@ -0,0 +1,85 @@
...
+  // Restore all registers from the stack.
+  lw   t0,  2 * OT_WORD_SIZE(sp)
+  lw   t1,  3 * OT_WORD_SIZE(sp)
+  lw   t2,  4 * OT_WORD_SIZE(sp)
+  lw   t3,  5 * OT_WORD_SIZE(sp)
+  addi sp, sp, OT_CONTEXT_SIZE
+  mret
+
+.L_not_a_flash_error:
+  j rom_interrupt_handler
```
jettr, line 83: Probably worth adding a comment that rom_interrupt_handler does or does not save the t0-t3 on the stack, but it doesn't matter since the chip it about to reset

cfrantz: Done.

sw/device/silicon_creator/rom/rom_isrs.S
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Exception handler.
+   */
+  .balign 4
+  .global rom_exception_handler_entry
+  .type rom_exception_handler_entry, @function
+rom_exception_handler_entry:
+  // Save all registers to the stack.
+  addi sp, sp, -OT_CONTEXT_SIZE
+  sw   t0,  2 * OT_WORD_SIZE(sp)
```
moidx, line 31: Do we have the shadow stack enabled in the ROM? If so, @luismarques can provide guidance on what would be the best way to use it in this routine.

sw/device/silicon_creator/rom/rom_isrs.S
```diff
@@ -0,0 +1,89 @@
...
+
+  // Check if there is a flash error.
+  li t3, TOP_EARLGREY_FLASH_CTRL_CORE_BASE_ADDR
+  lw t1, FLASH_CTRL_FAULT_STATUS_REG_OFFSET(t3)
+  andi t1, t1, PHY_ERRORS
+  beqz t1, .L_not_a_flash_error
+
+  // Clear the flash error.
+  sw t1, FLASH_CTRL_FAULT_STATUS_REG_OFFSET(t3)
+
```
moidx, line 69: is `FLASH_CTRL_FAULT_STATUS_REG_OFFSET` expected to read 0 after clear? If so, we can have a `beqz` into the rest of the sequence that leads to `mret` and have a few `unimp` entries after the branch statement.

sw/device/silicon_creator/rom/rom_isrs.S
```diff
@@ -0,0 +1,89 @@
...
+   * instruction is 16bits = 2bytes.
+   */
+  csrr t0, mepc
+  lh t2, 0(t0)
+  addi t0, t0, OT_HALF_WORD_SIZE
+  li t1, 0x3
+  and t3, t2, t1
+  bne t3, t1, .L_rom_16bit_trap_instr
+  // We already added one half word, so for a 32-bit instruction, add another.
+  addi t0, t0, OT_HALF_WORD_SIZE
```
moidx, line 52: We can add a check `t0 - mepc <= 4` using a separate register to harden this part of the code.

sw/device/silicon_creator/rom/rom_isrs.S
```diff
@@ -0,0 +1,89 @@
...
+.L_rom_16bit_trap_instr:
+
+  // Get the mcause, mask the reason and check that it is LoadAccessFault.
+  csrr t1, mcause
```
moidx, line 56: I  think this one is probably ok to leave unhardened since we are checking the current flash error.

sw/device/silicon_creator/rom/rom.ld
```diff
@@ -105,6 +105,20 @@ SECTIONS {
     . = ALIGN(4);
   } > rom
 
+  /**
+   * ROM ISRs text section, containing the exception handler entry/exit logic
+   * used for handling load access faults due to flash ECC errors during boot.
+   *
+   * This is placed next to the ".shutdown" section.
+   */
+  .rom_isrs : ALIGN(4) {
```
moidx, line 114: This LGTM, but open to discuss in case we want to collapse this into another region.

moidx (approved):

sw/device/lib/base/macros.h
```diff
@@ -550,4 +561,12 @@ class SignConverter {
     irq_global_ctrl(true);                             \
   }                                                    \
   irq_global_ctrl(true)
+
+/**
+ * Macros for implementing OT ISRs.
+ */
+#define OT_WORD_SIZE 4
+#define OT_HALF_WORD_SIZE (OT_WORD_SIZE / 2)
+#define OT_CONTEXT_SIZE (OT_WORD_SIZE * 30)
```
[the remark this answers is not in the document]
cfrantz: Done.

sw/device/silicon_creator/rom/rom_isrs.S
cfrantz: We only check for the two PHY_ERRORS bits (0x0180 hex).  Based on the error description, I don't think we should see any of the other errors.

sw/device/silicon_creator/rom/rom_isrs.S
cfrantz: Done.

sw/device/silicon_creator/rom/rom_isrs.S
cfrantz: Agree.

sw/device/silicon_creator/rom/rom.ld
cfrantz: Ok.  Lets leave it for now.

sw/device/silicon_creator/rom/rom_isrs.S
cfrantz: Done

moidx (approved):
Thanks @cfrantz! LGTM

sw/device/silicon_creator/rom/BUILD
```diff
@@ -176,6 +187,10 @@ opentitan_binary(
     linker_script = ":linker_script",
     deps = [
         ":mask_rom_lib",
+        # We depend on rom_isrs here rather than in mask_rom_lib so that the
+        # `rom_exception_handler` symbol remains unresolved and can be supplied
+        # by tests such as `rom_epmp_test`.
+        ":rom_isrs",
```
cfrantz, line 193: I had to move the `:rom_isrs` dependency here and adjust the symbol name to `rom_exception_handler` to allow the `rom_epmp_entry` test to pass.  That test supplies its own `rom_exception_handler` and we don't want the ECC handler to clash with it.

moidx: SGTM

sw/device/silicon_creator/rom/rom_isrs.S
```diff
@@ -0,0 +1,113 @@
...
+  // stack at `ram_end - 16` and the stack grows downwards from there, leaving
+  // 16 bytes at the top of RAM.  We need exactly 16 bytes to handle the flash
+  // exception, so we save `sp` into `mscratch` and use those 16 bytes to save
+  // the 4 registers we need to handle the flash exception.
+  csrw mscratch, sp
+  la   sp, (_stack_end - 16)
+  sw   t0,  0 * OT_WORD_SIZE(sp)
+  sw   t1,  1 * OT_WORD_SIZE(sp)
+  sw   t2,  2 * OT_WORD_SIZE(sp)
+  sw   t3,  3 * OT_WORD_SIZE(sp)
```
cfrantz, line 39: @moidx In order to allow `//sw/device/silicon_creator/rom/e2e/address_translation:rom_ext_a_flash_b` to function correctly, I needed to use a reserved exception stack.

The assembly-language startup of all of our next stages initializes SP with `la sp, (stack_end-16)`.  That pseudo-instruction translates into a PC-relative address load like this:
```
auipc sp, <some constant>
addi sp, sp, <another constant>
```
This PC-relative load fails in the above-referenced test because the test setup intentionally loads the binary into the wrong location in flash, thus causing SP to have a bad value.

moidx: This LGTM.

moidx:
LGTM after reviewing the reserved stack changes. Thanks @cfrantz!

cfrantz:

a-will:
FYI: Something about this PR didn't appear to mix well with https://github.com/lowRISC/opentitan/pull/23544. `//sw/device/silicon_creator/rom/e2e/boot_policy_flash_ecc_error:a_valid_b_corrupt_manifest_security_version_fpga_cw310_rom_with_fake_keys` is failing in CI as a result.

If we rebase on the commit prior (2c7d1e226096e62efd9e295c16a11fbf9d1ee3ea), that test passes. If we rebase on that PR's commit on master (3238897abe90a976ea45f78ed1c6eb39f9712d54), the test begins to fail.

PR #23384, opened 2024-05-29
[kmac,prim_sha2/rtl] Fix lint error in endian conversion functions

andreaskurth:
The linter flags an error when `conv_data` gets declared and initialized in one statement.  Splitting this into two statements removes the lint error.  This shouldn't make any functional difference inside a function, but it gets the file lint-clean without the need for a waiver.

### DIFF
--- a/hw/ip/kmac/rtl/kmac_pkg.sv
+++ b/hw/ip/kmac/rtl/kmac_pkg.sv
@@ -417,7 +417,8 @@ package kmac_pkg;
 
   // Endian conversion functions (32-bit, 64-bit)
   function automatic logic [31:0] conv_endian32( input logic [31:0] v, input logic swap);
-    logic [31:0] conv_data = {<<8{v}};
+    logic [31:0] conv_data;
+    conv_data = {<<8{v}};
     conv_endian32 = (swap) ? conv_data : v ;
   endfunction : conv_endian32
 
--- a/hw/ip/kmac/rtl/kmac_pkg.sv
+++ b/hw/ip/kmac/rtl/kmac_pkg.sv
@@ -421,9 +421,4 @@ package kmac_pkg;
     conv_endian32 = (swap) ? conv_data : v ;
   endfunction : conv_endian32
 
-  function automatic logic [63:0] conv_endian64( input logic [63:0] v, input logic swap);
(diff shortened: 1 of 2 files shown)


hw/ip/prim/rtl/prim_sha2_pkg.sv
```diff
@@ -113,12 +113,14 @@ package prim_sha2_pkg;
   };
 
   function automatic sha_word32_t conv_endian32(input sha_word32_t v, input logic swap);
-    sha_word32_t conv_data = {<<8{v}};
+    sha_word32_t conv_data;
+    conv_data = {<<8{v}};
     conv_endian32 = (swap) ? conv_data : v;
   endfunction : conv_endian32
 
   function automatic sha_word64_t conv_endian64(input sha_word64_t v, input logic swap);
```
gdessouky, line 121: I planned to remove `conv_endian64` altogether after our conclusion on how we will go about the endianness, we won't be needing it anymore. If you want to remove it already since you're at it, that would be great :-)

gdessouky:
Oh. Now that's a simple fix. Ascent Lint should be giving away hints to help us fix these little weird issue :-) Thanks!

gdessouky (approved):

hw/ip/kmac/rtl/kmac_pkg.sv
```diff
@@ -417,12 +417,14 @@ package kmac_pkg;
...
   // Endian conversion functions (32-bit, 64-bit)
   function automatic logic [31:0] conv_endian32( input logic [31:0] v, input logic swap);
-    logic [31:0] conv_data = {<<8{v}};
+    logic [31:0] conv_data;
+    conv_data = {<<8{v}};
     conv_endian32 = (swap) ? conv_data : v ;
   endfunction : conv_endian32
 
   function automatic logic [63:0] conv_endian64( input logic [63:0] v, input logic swap);
-    logic [63:0] conv_data = {<<8{v}};
```
ballifatih, line 425: +1 for the comment from @gdessouky about removing this function. I haven't found any calls to `conv_endian64`.

andreaskurth:
Thanks for your feedback! I'll remove `conv_endian64`

rswarbrick (approved):

PR #23386, opened 2024-05-29
[silicon_creator] make rnd32 work at owner stage

timothytrippel:
The ROM_EXT locks out OTP access, and the rnd32 function attempts to read a field in OTP, which was causing a load access fault when executed at the owner stage. This updates the rnd32 function to only read from OTP if it is executed at the ROM or ROM_EXT boot stages.


moidx (approved):

vsukhoml (approved):
sgtm! thank you.

cfrantz (approved):

Issue #23351, opened 2024-05-28
[pwrmgr] Deep/Normal sleep enter and WFI interacttion seems incorrect


jettr:
### Description

When entering enter deep sleep via `wfi` instruction (while performing all of the correct procedures), the core can wake up from the `wfi` instruction and proceed to execute the next instructions while the deep sleep mechanism is in the process of turning the core off. The deep sleep hardware will eventually turn off the core at some later point; this doesn't seem like the correct behavior.

I think either of the following are more consistent:
 * If the core returns from the `wfi` instruction, the deep sleep mechanism is cancelled
 * If the deep sleep mechanism has already started engaging, then the core should not be released to exectue anything after the `wfi` instruction.

---

This is not blocking us, since we can disable all of the "second-level" interrupt enable bits (e.g. `mie::msoft::CLEAR + mie::mtimer::CLEAR + mie::mext::CLEAR`) to ensure that `wfi` can never return back to the core.

jettr:
We actually are seeing this for normal sleep enter as well. The `wfi` instruction will return execution back to the core, but the machinery to enter normal sleep is asynchronously continuing to put the chip into a normal sleep mode. At some point in the future, the core is stopped for normal sleep (but it is many instructions after the `wfi`). This seems much worse than the deep sleep case as there is not as good of a firmware workaround for normal sleep other than just `no-op`'ing  in a loop after `wfi`, which does not seem great.

pamaury:
I am might be wrong since I am not a RISC-V expert but I think the spec makes no guarantee about the cirmcumstances in which a core wakes up from `wfi`. This is why the firmware should ideally call `wfi` in a loop and check it everytime if it was "woken up" for the intended reason. In fact, I believe that `nop` is a valid implementation of `wfi` so just calling `wfi` and hoping that the core will sleep before the next instruction is never correct.

Maybe @nbdd0121 or @andreaskurth can say more on this?

nbdd0121:
Yeah, wfi can spuriously wake up or be implemented as nop. Moreover, even with wfi implemented to not spuriously wake up, it also needs wake up (by spec) when there are interrupts pending but masked by global interrupt enables (MIE), so I wouldn't suggest rely on WFI stalling the processsor until deep sleep kicks in.

GregAC:
As @pamaury and @nbdd0121 say WFI doesn't really guarantee any specific behaviour. Though in the case of Ibex it should be ceasing execution until an interrupt occurs.

[...]

PR #23390, opened 2024-05-30
[pwrmgr,doc] Clarify how interrupts should be masked for low power

a-will:
Clarify that interrupt sources should be masked if they are not desired to contribute to a fall-through exit (or resumption of execution during a wait-for-interrupt exception).

CC @jettr 

Resolves https://github.com/lowRISC/opentitan/issues/23351

### DIFF
--- a/hw/ip_templates/pwrmgr/doc/programmers_guide.md
+++ b/hw/ip_templates/pwrmgr/doc/programmers_guide.md
@@ -19,6 +19,9 @@ Assume first the system has the power states described [above](theory_of_operati
 7. Set and poll [`CFG_CDC_SYNC`](registers.md#cfg_cdc_sync) to ensure above settings propagate across clock domains.
 8. Execute wait-for-interrupt instruction on the processing host.
 
+Note that entering low power mode requires that pwrmgr's `pwr_cpu_i.core_sleeping` input be at logic high long enough to be sampled.
+A wait-for-interrupt instruction does not guarantee entry into low power, since the CPU could immediately resume execution in some cases.
+
 ### Possible Exits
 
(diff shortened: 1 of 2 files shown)


a-will:
Possibly related issue: https://github.com/lowRISC/opentitan/issues/23351

nbdd0121:
Would a loop of WFI be sufficient?

a-will:
> Would a loop of WFI be sufficient?

No, interrupts stop low power entry (that is the fallthrough exit). It cancels the request, so the user needs to ensure all unwanted interrupt sources cannot trigger.

Technically, most IPs won't do anything without an active command from software, but it's probably simpler to just iterate through the half dozen interrupt enable entries in the PLIC and mask out whatever should be ignored.

jettr:
From the [fall through handling section](https://opentitan.org/book/hw/top_earlgrey/ip_autogen/pwrmgr/doc/theory_of_operation.html#fall-through-handling):

> For a normal low power entry, the fast FSM first checks that the low power entry conditions are still true. If the entry conditions are no longer true, the fast FSM “falls through” the entry handling and returns the system to active state, thus terminating the entry process.

I am not seeing the entry process getting terminated. We would be completely okay with that as it is consistent; however, we are seeing the fall through occur yet low power mode still engages at some later (but soonish) point in time.

a-will:
I should incorporate the changes from #23500 here.

a-will:
I added info on busy-polling CONTROL.LOW_POWER_HINT to the doc here, as a follow-up to #23500 .

vogelpi (approved):
Thanks @a-will for this update to the documentation. This looks perfect to me!

PR #23396, opened 2024-05-30
[dv] Minor tidy-up tweaks to tl_access_sub

rswarbrick:
No functional change, but it should make the code a little easier to read.


marnovandermaas (approved):
Looks like the same to me.

PR #19432, opened 2023-08-14
Transitioned the register tables of non topgen blocks to CMDGEN

HU90m:
Continuation of https://github.com/lowRISC/opentitan/pull/18694 and https://github.com/lowRISC/opentitan/pull/19266 .

This will fix https://github.com/lowRISC/opentitan/issues/19183 for the transitioned blocks.

The only blocks remaining are `ip_template`s (`alert_handler` and `rv_plic`). I'll likely have to make a small change to the `topgen` tool to transition these blocks.

Had to edit hjson files to remove trailing whitespace.

PR #23405, opened 2024-05-30
Initial Darjeeling ECDSA Support

sameo:
This is a PR for adding initial ECDSA support to Darjeeling.

This 89 (!) commits PR is, for most of it, a port from Earlgrey's `Earlgrey-PROD-M4-RC0` of the following folders:

* `sw/otbn/crypto`
* `hw/ip/otbn/doc`
* `hw/ip/otbn/data`
* `hw/ip/otbn/util`

There are two commits that are only partly based on Earlgrey's branch:

1. `[sw,silicon_creator,sigverify] Add ECDSA p256 verification support` (cda370739efcf0cebbbf0e2c3732af59b8fb8059)
2. `[sw,silicon_creator] Add OTBN boot services` (a4cf3b0f11f5eb37fcab925d09fd7096a421667f)

The first one adds the ECDSA p256 to the sigverify API, and is a copy paste of the Earlgrey code. I did not cherry-pick it directly as paths have diverged between the two branches.

The second one adds a reduced OTBN boot services support to Darjeeling. I removed the attestation part for now as the Earlgrey counter part relies on getting seeds from flash. We can not do that with Darjeeling and I thought it would be wiser to keep it out for now.


jadephilipoom:
Wow! In leiu of reviewing the diff line-by-line, I've reviewed the two commits
mentioned (LGTM) and checked out `integrated_dev` locally and attempted to
replicate the cherry-picks for the rest. Unfortunately, I wasn't really able to
do this cleanly, particularly for documentation changes. I think there's a bit
of a problem with isolation in the codebase and especially the documentation
that causes this to not quite be as easy as we'd like.

From my look at the changes, it appears that some commits have been reordered
with respect to their order on `master`, and most of them only cherry-pick
changes to the selected specific OTBN directories. But this has some unexpected
consequences, I think: for example, commit b348d949b5f2 (`[doc] Fixed links to
the getting started section.`) on master fixes documentation links throughout
different pages, including one OTBN page, after 7af944ccc moved the "getting
started" content into the main book. But this PR cherry-picks only the change
to the OTBN page from b348d9, not the commit that actually moves the content,
which I think will just break the link.

Here's my suggestion:
- Break this into at least two PRs -- one that handles documentation and
  `hw/ip/otbn/util`, which shouldn't depend on much the actual code (exceptions like the commit that makes IMEM 8kB, could be included in the docs/util PR), and one that adds the actual code (maybe even one for P256 and one for P384)
- Remove the first three commits in the chain (see below, they cancel each other out)
- Remove the `[doc] Fixed links to the getting started section.` commit, since
  as described above it breaks a link
- In all the PRs, re-order the cherry-picked commits by their commit date on
  `master` after selecting the commits, which should help them resolve properly.

Sorry to suggest such a big change of approach, but I want to make sure we can
keep things coherent and make future merges easier. Long-term, it would be good
to improve isolation in our documentation so that we don't get so many merge
conflicts.

--------------------------------------------------------------------------

Here are my longer-form notes if you want the gory details:

I suggest removing the first three commits in the chain:
- `[reggen] Add CIP_IDs and bump all major versions`
- `Revert "[reggen] Add CIP_IDs and bump all major versions"`
- `[reggen] Add initial support for version and cip_id hjson fields`

The second reverts the first, and the third has already been added on
`integrated_dev`
inhttps://github.com/lowRISC/opentitan/commit/538fdef42807462ff7701400bdb5088b9b5b06a6,
so the sum of the three commits shows no difference when I check it against
`integrated_dev`, and I think having the reversion and the extra copy of the
third commit is likely to confuse branch history a bit, especially since the
first two don't cleanly cherry-pick from `master` for me and differ
significantly from their `master` versions (due presumably to resolved
conflicts with 538fde).

When cherry-picking a9785f (`[otbn, rtl] Increase imem size to 8 KiB`), I
noticed that `hw/ip/otbn/doc/registers.md` did not exist on `integrated_dev`;
looks like it was added in a53d50 (`[doc] otbn registers and interfaces now use
CMDGEN`) and then modified in 963a50 (`[doc] Minor tweak to md sanitisation
code`). Both of those commits are cherry-picked later in the chain for this PR,
out of order compared to `master`, which means that the modifications from
a9785f (updating OTBN docs to say 8kB IMEM) get lost. I suggest moving the
cherry-picks of a53d50 and 963a50 earlier in the chain to fix this, which is
what I tried.

I had to remove a small block of text from
`hw/ip/otbn/doc/theory_of_operation.md` when cherry-picking a53d50 that got
removed on the `master` version but appears to have never existed on
`integrated_dev`. That was pretty solvable. However, when cherry-picking
963a50, I found a lot of conflicts relating to blocks that now use CMDGEN but
haven't been modified to do so on `integrated_dev`. Perhaps some more
cherry-picks are needed? Looks like a lot of blocks were moved to use CMDGEN in
https://github.com/lowRISC/opentitan/pull/19432, but the changes aren't totally
straightforward to merge onto `integrated_dev` because e.g.  paths in
documentation links to DIFs have diverged. For now I just dropped the changes
related to documentation files that didn't exist on `integrated_dev`.

After this the cherry-pick of a9785f worked automatically, so the duplicated
cherry-pick of this commit later in the chain is no longer needed (I checked).

Looking at the description of b348d949b5f2 I realized that this cherry-pick
does not make sense in isolation, as described up top. Skipped this one.

Had a few small conflicts in documentation links when cherry-picking 1c6ddf,
but they weren't hard to resolve.

Then, comparing to this PR, I realized there were still a lot of differences
compared to the state from the corresponding commits on the PR, because of
commits being re-ordered and changes getting auto-merged differently.
Particularly concerning is a variety of differences in `hw/ip/otbn/util`, since
that could affect functionality and not just documentation.

At this point I decided it was better to post my preliminary notes and agree on
a strategy than to continue hacking away. My WIP replication is at
https://github.com/jadephilipoom/opentitan/tree/verify-dj-otbn in case that's
useful.

sameo:
Hey @jadephilipoom 

> Wow! In leiu of reviewing the diff line-by-line, I've reviewed the two commits mentioned (LGTM)

Thanks!

>  and checked out `integrated_dev` locally and attempted to replicate the cherry-picks for the rest. Unfortunately, I wasn't really able to do this cleanly, particularly for documentation changes. I think there's a bit of a problem with isolation in the codebase and especially the documentation that causes this to not quite be as easy as we'd like.
> 

I agree. And FWIW we're also looking at syncing the opentitan tools and lib for roughly the same reasons. There are a lot more commits there, and we'll face the same issues I'm afraid.

> From my look at the changes, it appears that some commits have been reordered with respect to their order on `master`, and most of them only cherry-pick changes to the selected specific OTBN directories. But this has some unexpected consequences, I think: for example, commit [b348d94](https://github.com/lowRISC/opentitan/commit/b348d949b5f2f0a34ef33d5baf231070b7812986) (`[doc] Fixed links to the getting started section.`) on master fixes documentation links throughout different pages, including one OTBN page, after [7af944c](https://github.com/lowRISC/opentitan/commit/7af944ccc02a1a1f930c8deae58a0fa15bb041a9) moved the "getting started" content into the main book. But this PR cherry-picks only the change to the OTBN page from b348d9, not the commit that actually moves the content, which I think will just break the link.
> 
> Here's my suggestion:
> 
>     * Break this into at least two PRs -- one that handles documentation and
>       `hw/ip/otbn/util`, which shouldn't depend on much the actual code (exceptions like the commit that makes IMEM 8kB, could be included in the docs/util PR), and one that adds the actual code (maybe even one for P256 and one for P384)
> 
>     * Remove the first three commits in the chain (see below, they cancel each other out)
> 
>     * Remove the `[doc] Fixed links to the getting started section.` commit, since
>       as described above it breaks a link
> 
>     * In all the PRs, re-order the cherry-picked commits by their commit date on
>       `master` after selecting the commits, which should help them resolve properly.
>
> 
> Sorry to suggest such a big change of approach, but I want to make sure we can keep things coherent and make future merges easier. 

Please don't be sorry. I really appreciate the thorough review and the suggestion does make sense.

One comment though: If I split this PR into 2-3 of them, and order all commits chronologically in each of them, the final merge of all 3 PRs will not expose commits as they were chronologically merged in the `master` branch.

Thanks again for the feedback and suggestions.

> At this point I decided it was better to post my preliminary notes and agree on a strategy than to continue hacking away. My WIP replication is at https://github.com/jadephilipoom/opentitan/tree/verify-dj-otbn in case that's useful.

Yes, that makes sense to me as well.

sameo:
I am closing that PR as it will be split into multiple, smaller ones.

This is tracked through https://github.com/lowRISC/opentitan/issues/23486

PR #23409, opened 2024-05-30
Uart empty irq

GregAC:
This finishes the tasks in https://github.com/lowRISC/opentitan/issues/16693

@jettr @jesultra please note this will rename the existing `tx_empty` to `tx_done` and introduce a new `tx_empty` which is a true TX FIFO empty status type interrupt.


jesultra:
Thanks for the heads up. Will the new tx_done be asserted when the FIFO becomes empty or when the last (stop) bit leaves the shift register, while the FIFO is empty?
The latter is useful if the firmware wants to e.g. reset the chip after printing a message on the UART.

GregAC:
> Thanks for the heads up. Will the new tx_done be asserted when the FIFO becomes empty or when the last (stop) bit leaves the shift register, while the FIFO is empty?

It's the latter. It behaves exactly as the old `tx_empty` interrupt did (i.e. same behaviour you will observe with `tx_empty` in the engineering sample).

sw/device/lib/testing/uart_testutils.h
```diff
@@ -34,8 +34,9 @@ typedef struct uart_cfg_params {
   uint32_t base_addr;
   uint32_t peripheral_id;
   uint32_t irq_tx_watermark_id;
+  uint32_t irq_tx_empty2_id;
```
alees24, line 37: Is that name choice intentional or part of the renaming effort? It looks to me as though it will cause build failures.

hw/ip/uart/data/uart.hjson
```diff
@@ -61,10 +61,14 @@
       type:    "status",
       default: "1",
       desc:    "raised if the transmit FIFO is past the high-water mark."}
+    { name:   "tx_empty",
```
alees24, line 64: I understand wanting to keep the 'tx_' interrupts contiguous and separated from the 'rx_' ones although in this case I'd probably have been inclined to prioritize backwards compatibility and reduce the impact upon test software/SiVal needing to support multiple variants.
Perhaps agreement should be sought in this case? Renaming the existing interrupt helped with that, and it's unfortunate that adding another interrupt loses that advantage.

hw/ip/uart/data/uart.hjson
```diff
@@ -61,10 +61,14 @@
...
+      type:   "status",
+      default: "1",
```
alees24, line 66: I'm intrigued by the 'default' value of an interrupt, please?
I can't see it used anywhere other than the uart IP block; is this something that other IPs should be setting appropriately, perhaps for Status type interrupts?

hw/ip/uart/data/uart.hjson
```diff
@@ -61,10 +61,14 @@
...
+      desc:    "raised if the transmit FIFO has emptied."}
```
alees24, line 67: I think the description should say 'is '...FIFO is empty' otherwise it's likely to lead to confusion and demands for clarification.

alees24 (changes requested):
Thanks for this. It looks very nearly good to go; just a couple of clarifications and one - I think - mistake in 'uart_cfg_params_t', and it isn't it painful how much has to be changed when interrupts are modified/added?

hw/ip/uart/rtl/uart_core.sv
```diff
@@ -327,7 +327,7 @@ module uart_core (
   //
   // The alternative software fix is to disable tx_enable until it has a chance to
   // burst in the desired amount of data.
-  assign event_tx_empty     = ~tx_fifo_rvalid & ~tx_uart_idle_q & tx_uart_idle;
+  assign event_tx_done = ~tx_fifo_rvalid & ~tx_uart_idle_q & tx_uart_idle;
```
alees24, line 334: The comment above this driver still refers to 'The empty condition...' and should probably be updated in the presence of the new 'empty' interrupt.

hw/ip/uart/doc/theory_of_operation.md
```diff
@@ -165,13 +165,14 @@ and below if the system clock is over 31MHz.
...
+#### tx_watermark / tx_empty / rx_watermark
 If the TX FIFO level becomes smaller than the TX water mark level (configurable via [`FIFO_CTRL.RXILVL`](registers.md#fifo_ctrl) and [`FIFO_CTRL.TXILVL`](registers.md#fifo_ctrl)), the `tx_watermark` interrupt is raised to inform SW.
+If the TX FIFO is empty, the `tx_empty` interrupt is raised to inform SW.
 If the RX FIFO level becomes greater than or equal to RX water mark level (configurable via [`FIFO_CTRL.RXILVL`](registers.md#fifo_ctrl) and [`FIFO_CTRL.TXILVL`](registers.md#fifo_ctrl)), the `rx_watermark` interrupt is raised to inform SW.
 
-Note that the watermark interrupts are level-based status interrupts.
+Note that the watermark interrupts and the empty interrupt are level-based status interrupts.
 They will stay asserted for as long as the FIFO levels are in violation of the configured level and cannot be cleared by writing to the status register.
-This also means that the `tx_watermark` starts off in an asserted state (resets high).
+This also means that `tx_watermark` and `tx_empty` start off in an asserted state (resets high).
```
rswarbrick, line 175: Grammar nit: I think we want "reset" to match the plural.

hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
```diff
@@ -86,6 +86,37 @@ class uart_intr_vseq extends uart_base_vseq;
...
       end
 
+      TxEmpty: begin
+        if (!en_tx) return;
+        // Write a single byte to the TX FIFO which will immediately begin transmission
+        drive_tx_bytes(.num_bytes(1));
+        // Need a brief delay while the data byte is popped and the interrupt returns high
+        cfg.clk_rst_vif.wait_clks(1);
+        // Interrupt should now be high as TX FIFO will be empty (but TX of first byte on-going)
+        check_one_intr(.uart_intr(uart_intr), .exp(1));
```
rswarbrick, line 96: Would it make sense to also check that we were empty at the start, before we drove the first bytes? If something gets out of sync, that might be a bit easier to debug.

hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
```diff
@@ -41,7 +41,7 @@ class uart_intr_vseq extends uart_base_vseq;
         // tx may have unfinished transaction
         spinwait_txidle();
         dut_shutdown();
-        csr_wr(.ptr(ral.intr_state), .value('hff));
+        csr_wr(.ptr(ral.intr_state), .value((1 << NumUartIntr) - 1));
```
rswarbrick, line 44: Nit: Maybe it's worth pulling this cleanup into a separate commit (since it is correct even before we add the new empty interrupt). But no big deal.

sw/device/tests/uart_tx_rx_test.c
```diff
@@ -120,6 +120,8 @@ static volatile uart_cfg_params_t uart_cfg;
  */
 static volatile bool exp_uart_irq_tx_watermark;
 static volatile bool uart_irq_tx_watermark_fired;
+static volatile bool exp_uart_irq_tx_empty;
```
rswarbrick, line 123: This looks perfectly reasonable, but I wonder whether it makes sense to split the additions to the test into a separate commit? I don't think there's any reason they have to be atomic with the RTL change.

hw/ip/uart/data/uart.hjson
GregAC: I think it may be used by some of the autogen stuff for interrupt testing so it's aware this interrupt will be 1 out of reset rather than 0. Though if this is not being used elsewhere may be we've got another way to do this and I should just drop this line.

sw/device/lib/testing/uart_testutils.h
GregAC: It's a mistake. When I was first adding the new `tx_empty` I called it `tx_empty2` just to keep it distinct from the old `tx_empty`.

hw/ip/uart/data/uart.hjson
GregAC: Just checked it also employed on flash_ctrl at least

https://github.com/lowRISC/opentitan/blob/a182fcef27866fb502d8cceb810bdf76cfa3ef42/hw/ip_templates/flash_ctrl/data/flash_ctrl.hjson.tpl#L68-L77

It's used by autogen tests (e.g. so you know when you read from interrupt status after reset it will give a `1` for these interrupts).

hw/ip/uart/dv/env/seq_lib/uart_intr_vseq.sv
GregAC: A fair idea, but preferred not to do the commit edits

GregAC:
Thank you for the reviews @alees24 and @rswarbrick I believe I have addressed your points.

hw/ip/uart/rtl/uart_core.sv
GregAC: I decided to remove the whole thing, it didn't make a huge amount of sense now this become the done interrupt.

rswarbrick (approved):
Thanks for the tidy-ups. This looks good to me.

hw/ip/uart/data/uart.hjson
```diff
@@ -64,7 +64,7 @@
     { name: "rx_watermark",
       type: "status",
       desc: "raised if the receive FIFO is past the high-water mark."}
-    { name: "tx_empty",
+    { name: "tx_done",
       desc: "raised if the transmit FIFO has emptied and no transmit is ongoing."}
```
jesultra, line 68: I do not think we need further last-minute changes.  But I was trying to think how my code would use this, if trying to output "Rebooting..." on the console before doing a reset.

Naively, I would make sure to clear this interrupt status bit if set, then run a loop that puts the characters into the FIFO (while checking TXFULL bit), and then I would naively wait until seeing this tx_done status bit/interrupt before performing reset.

However, I came to think that if some other higher-priority interrupt happens during the loop, the FIFO could run empty and the tx_done flag be latched, before the loop continues, and then when the loop is done, my logic above would think that transmission is done when really the last half of the characters are still being actively output.  I would like to avoid having to clear the bit and put things into the FIFO while disabling interrupts, in order to guarantee that the FIFO cannot run empty "in the middle" of the operation.

As far as I can see, the above could be avoided if either this event type interrupt bit was automatically cleared upon writing any character to the TX FIFO, or if it too became a status type interrupt, which would be active exactly when the TX FIFO is empty and no transmit is ongoing.

As I said, I do not think the Ti50 team needs this change.  But I am just wondering how we are anticipating it could be used.

jesultra (approved):
This change looks good to me, with a minor comment about how tx_done would be used.

Issue #21355, opened 2024-02-13
[rom] Add another OTP setting that leave RESET_INFO register set


jettr:
### Description

Currently the owner firmware cannot read the [RESET_INFO] field since the ROM_EXT [always clears][1] that information. There is an OTP setting that instructs ROM_EXT to put that reset reason in the retention ram, but it would be nice the the owner firmware could also read the [RESET_INFO] register directly. Can we add an OTP setting that prevents ROM_EXT from clearing the register?

[RESET_INFO]:https://opentitan.org/book/hw/top_earlgrey/ip_autogen/rstmgr/doc/registers.html#reset_info

[1]:https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/rom/rom.c;l=187;drc=97ebef0bbefd6ec273856e993057d4b922d7a947

jettr:
This rom code should move to rom_ext, and we should re-purpose the `OTP_CTRL_PARAM_CREATOR_SW_CFG_RET_RAM_RESET` OTP flag to allow the reset reason register to remain uncleared for Owner Firmware (instead of just allowing the value to be stored in retention ram).

a-will:
Can someone write just a little more on the technical motivation here?

Right now, the text of the issue only amounts to an unsubstantiated opinion, and it'd be good to show the reasoning, even if it's not any deeper than, "The ROM was clearing reset reasons for convenience, but we'd like to see the accumulation of reset reasons in owner code, so we can see what happened without ROM hiding the old reasons."

cfrantz:
>  "The ROM was clearing reset reasons for convenience, but we'd like to see the accumulation of reset reasons in owner code, so we can see what happened without ROM hiding the old reasons."

You have correctly surmised the motivation: Some owners want an unaltered view of reset reasons.

johannheyszl:
@cfrantz should this move to ROM_EXT?

Issue #7887, opened 2021-08-23
[mask_rom] Should the mask ROM clear (and save) the reset reason?


mundaym:
The reset reason is 'sticky' and so at each reset reason bits accumulate until software clears them. For example, if after power-on the chip resets due to a watchdog timer bite both the POR and AON_TIMER reset reason bits will be set despite only the AON_TIMER being responsible for the last reset.

Unless there is already an existing plan for this I propose that mask ROM:

1. saves the current reset reasons into a slot in the retention SRAM
2. clears the reset reasons in the rstmgr ready for the next reset

Putting the reset reason into the retention SRAM means that we could save the reset reasons for the last N resets if we want. If we only store 1 reset reason then it could also just live in main SRAM.

mundaym:
@moidx @alphan @jon-flatley

PR #23411, opened 2024-05-30
[rom] Maybe don't clear the reset reason

cfrantz:
1. Add an OTP to control preserving the reset reason in rstmgr (default configuration: clear it).
2. Do not clear the reset reason if instructed to preserve it by OTP.
3. Add a test for the reset-reason-preserving OTP configuration.

Addresses #21355.

### DIFF
--- a/sw/device/silicon_creator/rom/rom.c
+++ b/sw/device/silicon_creator/rom/rom.c
@@ -210,7 +210,11 @@ static rom_error_t rom_init(void) {
 
   // Store the reset reason in retention RAM and clear the register.
   retention_sram_get()->creator.reset_reasons = reset_reasons;
-  rstmgr_reason_clear(reset_reasons);
+  if (otp_read32(
+          OTP_CTRL_PARAM_OWNER_SW_CFG_ROM_PRESERVE_RESET_REASON_EN_OFFSET) !=
+      kHardenedBoolTrue) {
+    rstmgr_reason_clear(reset_reasons);
+  }
 
   // This function is a NOP unless ROM is built for an fpga.
   device_fpga_version_print();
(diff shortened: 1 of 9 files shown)


sw/device/silicon_creator/rom/rom.c
```diff
@@ -217,7 +217,6 @@ static rom_error_t rom_init(void) {
 
   // Store the reset reason in retention RAM and clear the register.
   retention_sram_get()->creator.reset_reasons = reset_reasons;
-  rstmgr_reason_clear(reset_reasons);
```
a-will, line 220: Curious: Do we not want to clear that the POR happened, so we don't trigger any POR-specific code in ROM or ROM_EXT on deep sleep wakeups or NDM resets?

cfrantz: This is a request from the prodc customer to not clear the reset reason in ROM - they want to check the register and clear it in their firmware.  This draft PR is an experiment to see if there is any test fallout.

We'll then control this behavior via an OTP so the sku_creator can decide how they want the ROM to behave.

a-will: This sounds a little sketchy to me. However, I'll not stick my nose further into this, hehe.

Instead, I'd just tag @mundaym for visibility, since it seems like there might've been at least some discussion at one point: https://github.com/lowRISC/opentitan/issues/7887

a-will: Ah, was asked to tag @felixmiller too

a-will:
I wonder how many of those test failures simply come from tests that expected ROM to clear the reset reasons (since they now have to do it, else the reasons accumulate in retention SRAM). It looks like at least some fail because they reset and still see the POR reset reason, looping back to the same test phase without making forward progress.

The "lazy" thing where we move reset reason clearing into OTTF would probably clear the failures, but maybe that won't yield the desired testability for OTP controls, heh.

cfrantz:
> The "lazy" thing where we move reset reason clearing into OTTF would probably clear the failures, but maybe that won't yield the desired testability for OTP controls, heh.

Yeah, we discussed that possibility yesterday.  We can add something to the OTTF config structure to control whether or not to clear, which will allow tests that want an unmodified view of the reset reason register to get it.

moidx (approved):

PR #23415, opened 2024-05-31
[dv,usbdev] Usbdev DV cleanup

alees24:
This PR reduces the amount of replicated code and simplifies the code base, as well as using clearer naming for some base tasks.

_This PR is built on and includes #23313 which requires review and merge first._
The last 4 commits:

- Avoid the use of impicits endpoints ('endp' member in base class) by specifying the endpoint being addressed.
- Provide a randomized default endpoint to those sequences that address only one endpoint.
- Return 'timed out' indication in response item, rather than via agent configuration.
- Rename base tasks to be clearer.
- Modify start/finish_item creation of sequence items according to convention (no functional change).
- Introduce simplified base tasks for checking responses and IN packets, and for sending handshakes back to the DUT.

Pass rates unaffected, no functional changes. This is ongoing tidying and simplification.

_This PR is built on and includes #23313 which requires review and merge first._

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -226,7 +226,7 @@ endtask
   endtask
 
   // Construct and transmit a token packet to the USB device
-  virtual task call_token_seq(bit [3:0] ep, pid_type_e pid_type, bit inject_crc_error = 0);
+  virtual task send_token_packet(bit [3:0] ep, pid_type_e pid_type, bit inject_crc_error = 0);
     `uvm_create_on(m_token_pkt, p_sequencer.usb20_sequencer_h)
     m_token_pkt.m_ev_type  = EvPacket;
     m_token_pkt.m_pkt_type = PktTypeToken;
@@ -258,7 +258,7 @@ endtask
 
   // Construct and transmit a randomized DATA packet to the USB device, retaining a copy for
   // subsequent checks.
-  virtual task call_data_seq(input pid_type_e pid_type,
+  virtual task send_prnd_data_packet(input pid_type_e pid_type,
                              input bit randomize_length, input bit [6:0] num_of_bytes,
                              input bit isochronous_transfer = 1'b0);
     `uvm_create_on(m_data_pkt, p_sequencer.usb20_sequencer_h)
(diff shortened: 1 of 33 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -324,11 +324,15 @@ endtask
     check_tx_packet(in_data, pid_type, data);
   endtask
 
-  virtual task call_handshake_sequence(input pkt_type_e pkt_type, input pid_type_e pid_type);
+  // Send handshake to DUT after an appropriate turn-around delay.
+  virtual task send_handshake(input pid_type_e pid_type);
+    // Must delay for a few bit intervals before responding.
+    response_delay();
```
rswarbrick, line 330: I don't have strong feelings, but is this the right place to put the delay? I would probably have expected the delay to come "before the start of the handshake response", so maybe before the start of `send_handshake`?

That said, this change allows us to get rid of some repetition, so maybe it makes sense anyway?

hw/ip/usbdev/dv/env/seq_lib/usbdev_data_toggle_restore_vseq.sv
```diff
@@ -74,12 +74,12 @@ class usbdev_data_toggle_restore_vseq extends usbdev_base_vseq;
     // Respond as chosen to the DATA packet sent by the DUT.
     case (rsp)
       InResponseAck: begin
-        call_handshake_sequence(PktTypeHandshake, PidTypeAck);
+        send_handshake(PidTypeAck);
```
rswarbrick, line 77: If the timing of `send_handshake` is changing, do we want to drop the `wait_clks(20)` from line 72 above?

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: I wondered whether we may want at some point to suppress the delay using an optional parameter, perhaps so that we can test the response to an invalid delay. For now, though, it's fine.
A handshake packet is always the response to a packet just sent by the DUT, and there must be a 'response_delay' before sending a valid handshake; this is called 'bus turn-around time' because the direction of transmission is changing; contention must be avoided, voltage levels must stabilize and hubs must switch.

hw/ip/usbdev/dv/env/seq_lib/usbdev_data_toggle_restore_vseq.sv
alees24: Yes, thanks. There's been so much code replication that I'm sure I've missed one or two here and there.

rswarbrick (approved):
This looks good to me, thanks!

PR #23423, opened 2024-05-31
[rv_dm,dv] Improvements to rv_dm_tap_fsm_rand_reset

rswarbrick:
There are careful notes in the commit messages, but the overarching theme is to teach `rv_dm_tap_fsm_vseq` and the smoke sequence to allow injection of resets. This gets the pass rate from 2.5% (last night's run) to something more like 80% (where the failures are now all caused by timeouts).

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_smoke_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_smoke_vseq.sv
@@ -104,12 +104,19 @@ class rv_dm_smoke_vseq extends rv_dm_base_vseq;
         1: check_ndmreset();
         1: check_unavailable();
       endcase
+
+      spot_resets(should_stop);
+      if (should_stop) return;
+
       cfg.clk_rst_vif.wait_clks($urandom_range(1, 10));
     end
 
     repeat ($urandom_range(1, 5)) begin
       check_dmactive(should_stop);
+      spot_resets(should_stop);
+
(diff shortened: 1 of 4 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_smoke_vseq.sv
```diff
@@ -107,7 +108,8 @@ class rv_dm_smoke_vseq extends rv_dm_base_vseq;
     end
 
     repeat ($urandom_range(1, 5)) begin
-      check_dmactive();
+      check_dmactive(should_stop);
```
marnovandermaas, line 115: Why is seen reset the same as should stop? I would have thought it was inverted. If you set seen reset to one then it does not do the check so it does not stop.

marnovandermaas (approved):
From code inspection this looks reasonable to me. Just one question.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_smoke_vseq.sv
rswarbrick: The variable names are a bit confusing, sorry! The logic is that if we've seen a reset then we should stop this smoke vseq, rather than sending lots more items. In the `check_dmactive` task, the variable is called `seen_reset` (because that's what has happened). Down here in `body`, it's called `should_stop` because that's what the vseq should do :-)

Issue #22971, opened 2024-05-04
[manuf] bringup CP/FT flows on CW340 to better model entropy consumption


timothytrippel:
After #22957 merges, it will be more realistic to run these tests on the CW340, see: https://github.com/lowRISC/opentitan/pull/22957#issuecomment-2092739941

PR #23424, opened 2024-05-31
Add CW340 exec_env to CP/FT provisioning tests

tommychiu-github:
Link both CW310 and CW340 to the same definition for the tests.

Fixes: #22971


timothytrippel (approved):

timothytrippel:
Thanks @tommychiu-github !

PR #23427, opened 2024-05-31
[spi_host] Add ability to suppress clock edges

a-will:
Add a command segment bit that tells SPI host to hold SCK at the idle value. This enables software control of CSB, and it allows creating various signal patterns, such as the JESD252 serial flash reset request.

### DIFF
--- a/hw/ip/spi_host/data/spi_host.hjson
+++ b/hw/ip/spi_host/data/spi_host.hjson
@@ -410,6 +410,16 @@
       hwext: "true",
       hwqe: "true",
       fields: [
+        { bits: "14",
+          name: "CLKIDLE",
+          desc: '''Suppress output clock edges for this command.
+
+                   This keeps SCK at the idle value throughout this command.
+                   As a result, different patterns can be created on the pins without clocking data.
+                   This could be used for software control of the chip select, for example.
+                '''
+          resval: "0x0"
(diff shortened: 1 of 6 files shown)


hw/ip/spi_host/data/spi_host.hjson
```diff
@@ -410,6 +410,16 @@
       hwext: "true",
       hwqe: "true",
       fields: [
+        { bits: "14",
+          name: "CLKIDLE",
+          desc: '''Suppress output clock edges for this command.
+
+                   This keeps SCK at the idle value throughout this command.
+                   As a result, different patterns can be created on the pins without clocking data.
+                   This could be used for software control of the chip select, for example.
```
a-will, line 419: Here's how:

To set CSB, set CLKIDLE and CSAAT to 1, and set DIRECTION to 0.
To lift CSB, set CLKIDLE to 1, and set CSAAT and DIRECTION are 0.

a-will:
Hm... The JESD252 reset is possible, but it's a little wonky with just this bit. Either the single clock cycle between CSB rise and data output-enable dropping must be enough to satisfy the hold time, or the pad controls (for pull resistors) would have to be employed to make the waveform work.

Maybe this wasn't the best idea, heh.

PR #23428, opened 2024-05-31
[rom_e2e] Fix the e2e_weak_straps test

cfrantz:
This test should run an unsigned binary when executing under the `test_rom` and should run with empty flash when executing under the `mask_rom`.


cfrantz:
Addresses #21630.

moidx (approved):

PR #23429, opened 2024-05-31
[bazel] Experiment: add CW340 exec envs to default test list

jwnrt:
This PR:

* Adds the `fpga_cw340_sival_rom_ext` and `fpga_cw340_rom_with_fake_keys` envs to the default set.
* Adds those envs to tests which manually add the CW310 equivalents.
* Tidies up the exec envs of existing tests to use `EARLGREY_TEST_ENVS` and `EARLGREY_SILICON_OWNER_ROM_EXT_ENVS` where it looks appropriate.

There's a chance that the final point there could add extra environments which don't work. We'll see.


jwnrt:
Found one problem so far:

* The `binaries` parameter of `fpga_params` expects the key to be a single binary, which means adding a `cw340` `exec_env` to the `opentitan_binary` causes errors. Might be a way to point to the specific binary, not sure.

sw/device/lib/arch/device.h
```diff
@@ -216,6 +216,11 @@ uint64_t to_cpu_cycles(uint64_t usec);
  */
 void device_fpga_version_print(void);
 
+/**
+ * Returns whether or not the current device is an FPGA.
+ */
+bool device_type_fpga(void);
```
a-will, line 222: This is not a key that should be used for anything. Merely having an FPGA implementation does not logically imply specific properties. Let's not add this kind of shortcut when we'll need to directly describe capabilities for multi-top anyway.

jwnrt: Do you mean we should stop checking "FPGA or DV-sim or Verilator" and directly check the properties causing the special-casing, e.g. clock speed, available pads, etc?

a-will: > Do you mean we should stop checking "FPGA or DV-sim or Verilator" and directly check the properties causing the special-casing, e.g. clock speed, available pads, etc?

Yup!

> If so, I agree. I'm taking this shortcut for this experiment to save time, though.

No worries! I wanted to make sure, since we are carrying a lot of tech debt of this sort, and some of it is _just_ about coming due. xD

As we wrangle with that tech debt, we'll want to try to be more precise with the supported models and the mapping to a definition / query language.

jwnrt: I've opened https://github.com/lowRISC/opentitan/issues/23558 to track this and removed `device_type_fpga` from this PR.

jwnrt:
Here's a comparison of this PR's latest run to a baseline run that just happened in another PR:

CW310s:

* `CW310 Test ROM Tests`: Both 32 tests. 14m here, 6m there.
* `CW310 ROM Tests`: Both 477 tests. 50m here, 45m there.
* `CW310 SiVal non-ROM_EXT Tests`: Both 104 tests. 25m here, 22m there.
* `CW310 SiVal ROM_EXT Tests`: Both 202 tests. 24m here, 20m there.
* `CW310 Manufacturing Tests`: Both 70 tests. 23m here, 21m there.

In summary the same tests were run, this PR was slightly slower but seems to be within expected margins.

CW340s:

* `CW340 ROM Tests`:
  * 302 tests here in 21m.
  * 3 tests there in 7m.
* `CW340 SiVal ROM_EXT Tests`:
  * 279 tests here in 22m.
  * 3 tests there in 2m.
* `CW340 Manufacturing Tests`: Both 3 tests. 8m here, 8m there.

In summary, many more CW340 tests were run but fewer than the CW310 counterparts. Good start, though. Execution time is reasonable.

sw/device/tests/BUILD
```diff
@@ -3367,9 +3445,15 @@ opentitan_test(
 opentitan_test(
     name = "spi_host_winbond_flash_test",
     srcs = ["spi_host_winbond_flash_test.c"],
+    broken = fpga_params(tags = ["broken"]),
     exec_env = dicts.add(
         EARLGREY_TEST_ENVS,
         EARLGREY_SILICON_OWNER_ROM_EXT_ENVS,
+        {
+            # See #23558: needs porting to work on the CW340.
```
engdoreis, line 3453: Isn't this test and the above fixed already?

engdoreis (approved):
LGTM.

sw/device/tests/BUILD
jwnrt: Oh yes, I think we did fix this one, I'll re-enable in the next push

Issue #23352, opened 2024-05-28
[hmac, rtl] Reset hmac_core FSM to Idle after stop


ballifatih:
### Description

I am trying to interleave two HMAC streams. Namely:

- Stream A (HMAC-384) starts and feeds some of its messages.
- Stream A is stopped and the state is saved to SW context.
- Stream B (HMAC-256) starts and feeds some of its messages.
- After around ~1000 bytes or so, Stream B hangs forever, and the initial symptom is MSG_FIFO being full. The test eventually fails with a timeout.

I tried to track down the signals which fetch entries from MSG_FIFO, and I eventually stumbled upon the main FSM of `hmac_core`. When I issue `start` for Stream B, I was expecting `sha_hash_start_o` (of `hmac_core`) signal to be raised, which does not happen.

I am looking at waveforms from a high-level, so I am not sure if it helps much, but here is the clock cycle when `start` command is sent over TL-UL (note the vertical marker). `st_q` of `hmac_core` is at `StMsg` even though HMAC  should have a clear state.
![image](https://github.com/lowRISC/opentitan/assets/7850700/79d486f6-fd10-42d8-b6b7-624e96957d39)

Maybe we need a condition to go to `Idle` state from `StMsg` state, when `stop` is issued to HMAC HWIP. I do not understand the full interactions here to make a confident suggestion though.  

https://github.com/lowRISC/opentitan/blob/0e5093d7098daf6b9b848e5c1ca6242336e731d3/hw/ip/hmac/rtl/hmac_core.sv#L352-L370

cc: @martin-velay @gdessouky @andreaskurth

martin-velay:
I cannot try to reproduce it easily in simu at the moment. Would it be possible for you to take the latest RTL change (not merged into the master yet)?
https://github.com/lowRISC/opentitan/pull/23336/files

ballifatih:
Hi @martin-velay,
I don't think #23336 addresses this. In the test that I am running, the driver is actually not using the idle bit, but instead reads to interrupt to see if HMAC is stopped. Also the HWIP is not blocked after "stop" is called, but instead while the driver is feeding message bytes.

This is the exact function from the driver that waits for the interrupt:
https://github.com/ballifatih/opentitan/blob/33ec8044cdf293804aa338f25fd9fd25c9463649/sw/device/lib/crypto/drivers/hmac.c#L84-L99

I will re-run the test with #23336 to see if there is anything else that pops up.

gdessouky:
[...]

PR #23432, opened 2024-06-01
[hmac,rtl] HMAC FSM fixes for SAR

gdessouky:
This implements RTL changes to the HMAC FSM to handle context switching (transitions to `StDone` to wait on hash result, and then `StIdle` from where every context ought to start/continue) and aligns triggering of idle and done. With this change, the SAR smoke test passes on @martin-velay's SAR DV branch (https://github.com/lowRISC/opentitan/pull/23116) which tests out SAR across 2 contexts (HMAC + same digest size 256). 

This hasn't been tested out thoroughly, and I think other issues will come up as DV proceeds. I'm also running a block-level regression at my end. @ballifatih this should close this issue https://github.com/lowRISC/opentitan/issues/23352, but I wouldn't be surprised if something else comes up next.

### DIFF
--- a/hw/ip/hmac/rtl/hmac_core.sv
+++ b/hw/ip/hmac/rtl/hmac_core.sv
@@ -79,6 +79,9 @@ module hmac_core import prim_sha2_pkg::*; (
 
   logic hmac_sha_rvalid;
 
+  logic idle_d, idle_q;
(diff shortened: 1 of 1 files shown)


andreaskurth (approved):
LGTM, thx @gdessouky and @martin-velay

PR #23434, opened 2024-06-01
[rv_dm,dv] Increase rv_dm_jtag_dmi_csr_bit_bash timeout

rswarbrick:
We increase to to 150ms in the rv_dm base vseq, but this gets overridden by the command line argument from the hjson file. The whole point of that argument was to give a *higher* timeout, which isn't quite what is happening! Bump up to 150ms as an initial step to match.


rswarbrick:
I didn't test this properly and it doesn't quite do what we need. I've got a better version that I'll push in a second, but it's kind of different from this approach, so I think it makes more sense in a new PR.

PR #23435, opened 2024-06-01
[dv] Minor cleanups in the bit bash sequence

rswarbrick:
There shouldn't be any functional change, but the code hopefully becomes much easier to glance at. Since I wanted to make sure I understood it anyway, I thought this would be a chance to make it easier for the next person...

### DIFF
--- a/hw/dv/sv/csr_utils/csr_seq_lib.sv
+++ b/hw/dv/sv/csr_utils/csr_seq_lib.sv
@@ -350,72 +350,81 @@ class csr_bit_bash_seq extends csr_base_seq;
         continue;
       end
 
-      begin
-        uvm_reg_field   fields[$];
-        string          mode[`UVM_REG_DATA_WIDTH];
-        uvm_reg_data_t  dc_mask;  // dont write or read
-        uvm_reg_data_t  cmp_mask; // read but dont compare
-        int             n_bits;
-        string          field_access;
-        int             next_lsb;
+      bash_register(test_csrs[i]);
+    end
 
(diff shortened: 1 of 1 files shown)


hw/dv/sv/csr_utils/csr_seq_lib.sv
```diff
@@ -350,72 +350,80 @@ class csr_bit_bash_seq extends csr_base_seq;
...
+        mode[next_lsb] = field_access;
+        dc_mask[next_lsb] = dc;
+        cmp_mask[next_lsb] = cmp;
+        next_lsb++;
       end
     end
 
+    // Any unused bits on the left side of the MSB?
+    while (next_lsb < `UVM_REG_DATA_WIDTH)
+      mode[next_lsb++] = "RO";
```
hcallahan-lowrisc: I know this was already there, but I'm not sure I like the lack of begin/end here, with the expression wrapped onto the next line. Okay I looked, and the [coding style forbids it](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#begin--end:~:text=//%20Incorrect%20because%20a%20wrapped%20statement%20must%20have%20begin%20and%20end.). Would you mind fixing while you're here? Thanks

hcallahan-lowrisc (approved):
LGTM

hw/dv/sv/csr_utils/csr_seq_lib.sv
rswarbrick: Done!

PR #23439, opened 2024-06-02
[rv_dm,jtag,dv] Fix JTAG monitor getting out of sync with the TAP

rswarbrick:
There are some detailed notes in the commit messages. Honestly, I'm slightly surprised about the riscv-dbg TAP behaviour here and intend to raise a "huh, really?" issue on the pulp repo to check it's what they intend.


rswarbrick:
My logic was that the first commit is "generic jtag stuff" and the second is rv_dm specific, which is why is separated them out. (It also means that if I've messed up something chip-wide, the bisection won't find something rv_dm-specific!)

PR #23440, opened 2024-06-02
[ast] Updating Lc2HcTrCyc, Hc2LcTrCyc according to macro delay

meisnere:
This change impacts wakeup and power down time.


andreaskurth:
`chip_sw_pwrmgr_smoketest` currently fails because this bound is too low

https://github.com/lowRISC/opentitan/blob/4ffc4de67f949c45261ebc1be0ff3617d84cfa54/hw/top_earlgrey/dv/sva/clk_ctrl_and_main_pd_sva_if.sv#L31

I think that makes sense because this PR changes the number of clock cycles in power transitions. I'm thus increasing the bound -- 160 cycles seems to suffice. Amending the commit so as to keep CI passing.

andreaskurth (approved):

Issue #22954, opened 2024-05-03
[sival] rv_core_ibex_nmi_irq_test_silicon_owner_sival_rom_ext failure


mundaym:
### Description

Enabling alert_handler ping mechanism results in alert_handler NMI without any reported local of regular alerts. The only way to recover from this is by resetting the device.

pamaury:
I think @nbdd0121 has investigated the test in more depth and might comment here?

andreaskurth:
@moidx is currently investigating this. If there are any comments from others please provide them

moidx:
Configuring the alert_handler triggers nmi interrupt without any alert class association.

https://github.com/lowRISC/opentitan/blob/605da270ddcf70b2af5573e9524f9613bc89383d/sw/device/tests/rv_core_ibex_nmi_irq_test.c#L98-L101

The alert NMI is triggered even when there are no pending alerts. Verified this by removing the force alert.

https://github.com/lowRISC/opentitan/blob/605da270ddcf70b2af5573e9524f9613bc89383d/sw/device/tests/rv_core_ibex_nmi_irq_test.c#L203-L204

Disabling the ping timer removes the spurious NMI issue:

https://github.com/lowRISC/opentitan/blob/605da270ddcf70b2af5573e9524f9613bc89383d/sw/device/lib/testing/alert_handler_testutils.c#L139-L141

The following write in dif_alert_handler.c is what triggers the NMI from alert_handler:

```c
  // Lock the configuration.
  if (locked == kDifToggleEnabled) {
    mmio_region_write32(alert_handler->base_addr,
                        ALERT_HANDLER_PING_TIMER_REGWEN_REG_OFFSET, 0);
  }
```

moidx:
I tried switching the test to use a recoverable alert, different escalation sequences and a valid ping timeout timer configuration, but none of these changes helped to move the alert handler NMI. 

I will submit these test updates separately, but I think at this point I am going to flag this for CDC analysis. 

CC: @a-will @matutem @nbdd0121 who also took a look at this issue.

moidx:
I am going to measure the delay between enabling the ping mechanism and the NMI to try to determine if this is an issue with the reverse ping mechanism supported by the receivers. 

This was recommended by @msfschaffner.

moidx:
Running the test with the alerts and ping configured, and without triggering the alert results in the first alert NMI triggering between 500-4000 microseconds.

This as measured with rv_timer using an equivalent 1us tick.

[...]

PR #23441, opened 2024-06-02
[dv/top] Add alert_handler ping_ok test.

moidx:
Ensure the pinging mechanism does not trigger any alert handler NMI interrupts as a result of synchronization issues between ping senders and receivers.

This is part of https://github.com/lowRISC/opentitan/issues/22954. 

This test will be added to the GLS test plan.

### DIFF
--- a/sw/device/tests/alert_handler_ping_ok_test.c
+++ b/sw/device/tests/alert_handler_ping_ok_test.c
@@ -1,0 +1,247 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <assert.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "sw/device/lib/arch/device.h"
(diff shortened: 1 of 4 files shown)


sw/device/tests/alert_handler_ping_ok_test.c
```diff
@@ -0,0 +1,245 @@
...
+  irq_external_ctrl(false);
+  init_peripherals();
+
+  CHECK_DIF_OK(
+      dif_rv_core_ibex_enable_nmi(&rv_core_ibex, kDifRvCoreIbexNmiSourceAll));
+
+  alert_handler_ping_ok_test();
+
+  // Double check that no external interrupts have occurred.
+  CHECK(ext_irq_fired == false, "Unexpected external interrupt triggered.");
```
cfrantz, line 241: Do you also need to check that no NMIs have fired? (`counters.alert_raised`)

cfrantz (approved):

sw/device/tests/alert_handler_ping_ok_test.c
moidx: Good catch. Thanks! Fixed.

Issue #23348, opened 2024-05-28
[hmac, dv] `initiated` remains high during context switches


ballifatih:
### Description

I am running a driver-level streaming test where 6 different vectors (one for each algorithm) are streamed in an interleaving fashion. There is a particular issue that I encountered with `ValidHmacEnConditionAssert` assertion. Here is the important part from the test:
- SHA-384 is initiated and some message bits are fed.  
- SHA-384 is stopped (not completed), and context is saved in SW.
- HMAC-384 is initiated and some bits are fed.
- HMAC-384 is stopped and context is saved in SW.

However, in the waveforms, I can see that `initiated` bit is not cleared to 0, when SHA-384 is stopped. I think this bit should be reset to 0, when "stop" is issued.

https://github.com/lowRISC/opentitan/blob/d268f271f4f75aeb8f3bf9624a497ae5bfb9c47e/hw/ip/hmac/rtl/hmac.sv#L907-L924

I tagged this issue as DV issue because this signal seems to be only used by the assertion. I don't think it is very useful, but this is what I got as an error:

```
* `Offending '((!in_process) && (!initiated))'` has 1 failures:
    * Test chip_sw_hmac has 1 failures.
        * 0.chip_sw_hmac.1\
          Line 458, in log /workspaces/repo/scratch/mv_hmac_impl/chip_earlgrey_asic-sim-vcs/0.chip_sw_hmac/latest/run.log

                	Offending '((!in_process) && (!initiated))'
                UVM_ERROR @ 8908.259032 us: (hmac.sv:923) [ASSERT FAILED] ValidHmacEnConditionAssert
                UVM_INFO @ 8908.259032 us: (uvm_report_catcher.svh:705) [UVM/REPORT/CATCHER]
                --- UVM Report catcher Summary ---
                

```

cc: @martin-velay @gdessouky @andreaskurth

martin-velay:
I am currently making some updates in the DV to be able to switch between 2 different contexts. I'll try to highlight this issue then

gdessouky:
@ballifatih, do you still trigger this assertion if you change

`else if (hash_process)           initiated <= 1'b0; `  to      
`else if (hash_process || reg_hash_stop)  initiated <= 1'b0;` ?

gdessouky:
Closing as addressed by https://github.com/lowRISC/opentitan/pull/23456 and not blocking the TLT anymore.

PR #23456, opened 2024-06-03
[hmac, dv] Fix `ValidHmacEnConditionAssert`

ballifatih:
Related to #23348.

The fix is suggested by @martin-velay, but I am making this small PR so we can quickly merge it. This would unblock us from running driver tests as TLT (otherwise this assertion leads to trivial failures of the tests that I plan to upstream soon).

cc: @martin-velay @gdessouky @andreaskurth

### DIFF
--- a/hw/ip/hmac/rtl/hmac.sv
+++ b/hw/ip/hmac/rtl/hmac.sv
@@ -906,9 +906,9 @@ module hmac
 
   logic initiated;
   always_ff @(posedge clk_i or negedge rst_ni) begin
-    if (!rst_ni)                     initiated <= 1'b0;
-    else if (hash_start_or_continue) initiated <= 1'b1;
-    else if (hash_process)           initiated <= 1'b0;
+    if (!rst_ni)                              initiated <= 1'b0;
+    else if (hash_start_or_continue)          initiated <= 1'b1;
+    else if (hash_process || reg_hash_stop)   initiated <= 1'b0;
   end
 
   // the host doesn't write data after hash_process until hash_start_or_continue.


martin-velay:
Actually this fix is suggested by Ghada 😅

@ballifatih , have you tested it with a local change to be sure it solves your issue? Because I cannot reproduce it in simu for the moment, so I cannot say.

ballifatih:
Yes, with this change, TLT is not blocking anymore on my side. @martin-velay

martin-velay (approved):
I believe the modification is fine, but I haven't checked in simu. At least it makes sense to me.

gdessouky (approved):

hw/ip/hmac/rtl/hmac.sv
```diff
@@ -906,9 +906,9 @@ module hmac
 
   logic initiated;
   always_ff @(posedge clk_i or negedge rst_ni) begin
-    if (!rst_ni)                     initiated <= 1'b0;
-    else if (hash_start_or_continue) initiated <= 1'b1;
-    else if (hash_process)           initiated <= 1'b0;
+    if (!rst_ni)                              initiated <= 1'b0;
+    else if (hash_start_or_continue)          initiated <= 1'b1;
+    else if (hash_process || reg_hash_stop)   initiated <= 1'b0;
```
vogelpi, line 911: I think it would be cleaner to use `hash_done_event` which is set at the very end of a (stopped) hash operation. After the `reg_hash_stop` pulse, it may still take a while for the hash to complete.

However, this code here is just used for assertions and not synthesized. So I am fine to approve and merge this as is to unblock people.

vogelpi (approved):

hw/ip/hmac/rtl/hmac.sv
gdessouky: now that I look at this again, I understand that the intention of this assertion is to check that `hash_process` (or now `hash_stop` too) are only enabled after `hash_start` (or now `hash_continue`) have been enabled first, i.e., they are paired and toggled in that sequence. So I would say that we use ought to be using `reg_hash_stop` here, but instead change this  
``ASSERT(ValidHashProcessAssert, hash_process |-> initiated)``
to
``ASSERT(ValidHashProcessAssert, (hash_process || reg_hash_stop) |-> initiated)``

vogelpi: Thanks for your feedback both. I am against disabling the assertion (even temporarily). Disabling verification is never a good idea.

I am merging this now to unblock @ballifatih 's PR for the nightlies. But please feel free to file another PR @gdessouky to rework the assertion as suggested.

Issue #23460, opened 2024-06-03
[otbn,dv] Add tests to cover post-ES security features


rswarbrick:
### Description

There have been three RTL changes since ES that add security behaviour that probably needs testing. Write these tests.

- a6bbefa5be6445f593c33699772159c7a284ee5a
  + Ensure that we send an invalid `lc_escalate_en_i` signal and make
    sure that it is interpreted as a genuine escalation request.
- 5228dfccf508dbcd7823b4d2f2afe55d0a9332f9
  + Ensure that we send an invalid `lc_rma_req_i` signal and make sure
    that it is ignored.
- a4d9f164bb7495a5c4275c8b2094fd86f45a2e8c
  + Add error injection tests that raise the following signals when
    secure wipe is not running:
    - `sec_wipe_mod_urnd_i` in `otbn_alu_bignum`
    - `sec_wipe_zero_i` in `otbn_controller`
    - `sec_wipe_base` in `otbn_core`
    - `sec_wipe_wdr_q` in `otbn_core`
    - `sec_wipe_stack_reset_i` in `otbn_rf_base`

rswarbrick:
These three PRs will cover the features once they are all merged:
- #23498 
- #23508 
- #23545

Issue #22077, opened 2024-03-18
[otbn,dv] V2 signoff


rswarbrick:
### Description

Get OTBN ready for V2 signoff and sign it off.

rswarbrick:
I've spent a bit of time looking carefully at OTBN (and signoff for M4).  This is *logically* reasonably simple because we're only really tracking deltas since ES. But there have been quite a few changes! To start with, I'm looking at the changes needed to get to the V2 level.

The text below follows the V2 signoff checklist (but dropping the FPV
items, since they don't apply).

### DESIGN_DELTAS_CAPTURED_V2

Here is a list of the commits since the ES branch taken that might
touch the design or RTL. Each has a note about whether the change
might have a verification impact.

I generated them with `git log Earlgrey-M2.5.2-RC0..HEAD hw/ip/otbn`
(on 3/6/24) and walked through the list carefully, dropping items that
get reverted, are just DV changes or touch a Yosys config in the
pre_syn directory.

- e47df29f3e12a48c62f381a240c8cebf9658b91a
  + No functional change: just moves a check to a shared helper
    function.
- a6bbefa5be6445f593c33699772159c7a284ee5a
  + This slightly changes the design: if we send an invalid
    `lc_escalate_en_i signal`, it should now be interpreted as a
    genuine escalation, rather than triggering the `mubi_err` pathway,
    which is interpreted as a `bad_internal_state` error.
  + TASK: track this with a cover point and make sure we see it with
    testing.
- 31cbfd6793acc1e5c24d2b48019afb2143851196
  + Documentation change for users of the block. No DV change
    required.
- 1b16ca2122d956f4b1f909ba2e996b01ff22001e
  + Reggen change: no DV change required.
- 803d9aec77f18496007d493d7ad3b60bc15c734f
  + This will have an impact on the design, but is a change to a block
    that OTBN uses, so we shouldn't expect any impact on OTBN
    block-level DV.
- de31bdf1c26be2714ca787c6041a56eb56eb7d1e
  + Reggen change with no functional impact: no DV change required.
- fc8484601ef03a0bb05956c1d63142d4bfb90496
  + Reggen change with no functional impact: no DV change required.
- 61a237e1973d6a29d377c2be871139553c7d344f
  + Reggen change to change how substructs are encoded. No functional
    change intended in OTBN, and the behaviour will be seen by
    existing DV.
- 6bac48986c87ec909f0edb58b2b6159d48a26a38
[...]

PR #23463, opened 2024-06-03
[otbn,dv] Add post-ES sec_cm feature tests to testplan

rswarbrick:
These match the post-ES features discovered when analysing #22077. Implementing these tests is tracked as #23460.


rswarbrick:
Oops, that's embarrassing! I added the testpoints to the wrong testplan. Fixed now, I hope.

rswarbrick:
I've rewritten this PR. It now just adds a single testpoint (since all the testpoints that I had added were actually addressed by the same test) and links it to the newly written `otbn_partial_wipe` test.

PR #23465, opened 2024-06-03
[doc] Add V2S scoping to V2 checklist

rswarbrick:
The existing V2 checklist had an item (V3_CHECKLIST_SCOPED) to understand scope and estimate effort for V3. We added the V2S development stage since that was defined, and we're definitely going to need to estimate effort for that at some point!

Add the task to the V2 checklist accordingly. It will be a reasonably quick job for most blocks.


rswarbrick:
Thinking about this a bit more, I'm not really convinced this is worth separately tracking. We can always reconsider in the future if I turned out to be unexpectedly right :-)

PR #23471, opened 2024-06-03
[top-level] Update HMAC top level test plan.

moidx:
Add the following test cases:

1. `chip_sw_hmac_all_configurations`: Include coverage for 384 and 512 modes of operation.
2. `chip_sw_hmac_stream_mode`: Include coverage for context save and restore across multiple modes of operation.

### DIFF
--- a/hw/top_earlgrey/data/ip/chip_hmac_testplan.hjson
+++ b/hw/top_earlgrey/data/ip/chip_hmac_testplan.hjson
@@ -43,6 +43,49 @@
       bazel: ["//sw/device/tests:hmac_enc_idle_test"]
     }
     {
+      name: chip_sw_hmac_all_configurations
+      desc: '''Verify HMAC 256/384 and 512 modes of operation.
+
+            - Verify the following security configurations:
+              - SHA2-256
+              - SHA2-384
+              - SHA2-512
+              - HMAC-SHA2-256
+              - HMAC-SHA2-384
+              - HMAC-SHA2-512
+            - Verify input message size of 0 bytes.
+            - Verify multiple input message sizes.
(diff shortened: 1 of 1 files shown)


ballifatih (approved):
LGTM.

I plan to make small modification in the tests with an upcoming PR that will bring multistream test, so I will update the descriptions to match that. Based on my discussion with @gdessouky today, we converge on the following direction/priorities in terms of increasing testing coverage this week, and some of them will shape TLTs:

- Test one-shot calls (at the moment, oneshot calls are internally broken down to streaming calls).
- Enable SHA-2 tests from cryptotest
- Take a look at wycheproof vectors to see if  the corner cases that are interesting for TLT are covered. (e.g. those are mostly tests whose parameters are equal or close to block sizes)
- For hard-coded vectors, cherry-pick tests such that:
  - 6 variants are coverd (SHA-2/HMAC x 256/384/512)
  - 3 different message sizes (smaller than, equal to, larger than the block size)
  - 3 key sizes (smaller than, equal to, larger than the block size)
- Streaming vs one-shots: both cases should be covered (so 1 TLT for each)
- Test word at a time 32b writing and use of idle status flag.

gdessouky:
Thanks @ballifatih for capturing it here!

> * Take a look at wycheproof vectors to see if  the corner cases that are interesting for TLT are covered. (e.g. those are mostly tests whose parameters are equal or close to block sizes) 

& zero-length messages.



> * Streaming vs one-shots: both cases should be covered (so 1 TLT for each)

Would be nice if we can interleave streaming with one-shot variants in same TLT, e.g., context A is streaming and context B is using one-shot, etc. in same TLT.

cfrantz (approved):

PR #23474, opened 2024-06-04
[pwrmgr] Latch and synchronize WFI events

a-will:
Latch qualified WFI events in the CPU clock domain, so fast transitions there are not dropped. Synchronize those events to the pwrmgr fast FSM domain.

Prevent changes to low_power_hint while its 1->0 transition is clearing events in the CPU domain. Use cfg_cdc_sync to have software synchronize a low_power_hint high to enable CPU sleep monitoring (before issuing a WFI instruction).


hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr_cdc.sv
```diff
@@ -241,24 +262,168 @@ module pwrmgr_cdc import pwrmgr_pkg::*; import pwrmgr_reg_pkg::*;
     .q_o(usb_ip_clk_en_o)
   );
 
-  prim_flop_2sync # (
-    .Width(1)
-  ) u_sleeping_sync (
+  // Synchronize qualified core_sleeping events into the pwrmgr fast domain. Use a handshake with
```
a-will, line 265: @GregAC Maybe something like this could help with https://github.com/lowRISC/opentitan/issues/23351

It won't do anything for any potential WFI instructions that yield 0 cycles of core_sleeping, though.

a-will:
The thinking here is thay setting low_power_hint from software starts a new session of sorts. From there, pwrmgr only needs to know (1) if the core ever went to sleep and (2) if the core ever woke back up again. The former kicks off trying to enter low power mode, while the latter is checked later and causes a fall through exit. cdc_cfg_sync is used to bring that "new session" sense over to the CPU domain.

For cleaning up a session, the fast FSM will clear the low power hint, so we use that as the impulse to start tearing down. Since teardown needs to reset state in the CPU domain, we wait for notification of that completion before making the CONTROL CSR writeable again.

Thus, for software to enter low power mode, it would need to wait for CONTROL to unlock by polling the REGWEN, do the usual setup, set LOW_POWER_HINT, issue CDC_CFG_SYNC and wait for it to clear again, then do WFI.

hw/top_earlgrey/data/top_earlgrey.hjson
```diff
@@ -355,6 +355,10 @@
         clk_i: "io_div4",
         clk_slow_i: "aon",
         clk_lc_i: "io_div4",
+        clk_cpu_i: {
+          clock: "main",
+          group: "infra",
```
a-will, line 360: Not sure if these are the clock and reset groups that should be used. I'm not clear on infra vs. secure vs. powerup as choices here.

a-will:
Hm... maybe we can deal with any lack of core_sleeping output by allowing software to cancel requests? Software would check WAKE_INFO, see it is 0, check CONTROL and see LOW_POWER_HINT is still 1, then issue a LOW_POWER_CANCEL request.

If pwrmgr's FSM somehow got triggered to kick off the low power entry, it would ignore the cancellation. However, if it is still waiting for WFI, LOW_POWER_CANCEL would immediately bring LOW_POWER_HINT low and kick off the "session" teardown. It's only after the teardown completes that CONTROL becomes writeable again.

I need to run something in DV to grab waveforms first, to make sure there isn't something else broken. :smile:

a-will:
@GregAC It looks like zero cycles of core_sleeping on WFI is possible:

image

a-will:
If we add the ability to cancel, maybe we don't need this more complex CDC, though :thinking: 

The short pulse can be collapsed into the same space as the "absent pulse" variant. Software would request cancellation, and the fast FSM would only accept it if it hasn't moved from `FastPwrStateActive`.

a-will:
Changing tack: @matutem suggested deferring the CONTROL lock until the actual transition from `FastPwrStateActive`, which would make the behavior consistent with the docs, incidentally.

I'll change the wakeup process to involve writing 0 to LOW_POWER_HINT, then busy-polling until LOW_POWER_HINT becomes 0. This should ward off races with a core_sleep pulse that somehow makes it to the pwrmgr fast FSM (shouldn't happen with earlgrey's clocks, but... just in case).

PR #23476, opened 2024-06-04
[spi_host, rtl] Spi host sd_en_o stable through the entire cycle (fsm fix)

antmarzam:
The RTL has a bug in which if  `cmd_wr_en_q` is set, `sd_en_o` can be switching values mid-spi cyle.
The proposed change suggests only to allow to drive `sd_en_o` directly via `cmd_wr_en_q` when the edge is correct. Otherwise, `sd_en_o` is driven directly by it's latched value.

This PR aims to tackle the RTL side of #22182

### DIFF
--- a/hw/ip/spi_host/rtl/spi_host.sv
+++ b/hw/ip/spi_host/rtl/spi_host.sv
@@ -147,7 +147,7 @@ end
 
 `endif
 
-  
+
   spi_host_reg2hw_t reg2hw;
   spi_host_hw2reg_t hw2reg;
 
--- a/hw/ip/spi_host/rtl/spi_host.sv
+++ b/hw/ip/spi_host/rtl/spi_host.sv
@@ -43,6 +43,111 @@ module spi_host
 
   import spi_host_cmd_pkg::*;
 
+  // Note: this check is carried out via bind. The SVa is copied here in order to reproduce failures
+  //  in this branch, This commit will be dropped
+`ifndef SYNTHESIS
+  import uvm_pkg::*;
+ `include "uvm_macros.svh"
+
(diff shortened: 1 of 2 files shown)


antmarzam:
In order to reproduce the issue, you can run:
```
./util/dvsim/dvsim.py hw/ip/spi_host/dv/spi_host_sim_cfg.hjson --build-opts="-lwdgen"  -t xcelium -i spi_host_stress_all  -v n -r 1 -s 16594327444610937479871567872505438579723151996600861010153494694429326213120
``` 
You can see the assertion correctly fires with the error message:
```
xmsim: *E,ASRTST (/home/antonio/repos/opentitan_fix/opentitan/scratch/spi_host_fsm_fix/spi_host-sim-xcelium/default/src/lowrisc_ip_spi_host_1.0/rtl/spi_host.sv,137): (time 315851966 PS) Assertion tb.dut.g_cio_signal_stable_sva[0].POSEDGE_SAME_VALUE_CHECK_P has failed 
UVM_ERROR @ 315851966 ps: (spi_host.sv:144) [POSEDGE_SAME_VALUE_CHECK_P] [i=0] - ASSERTION FAILED pos_value (0x1) != neg_value (0x0) - time=315 852 000 ps
```

A screenshot of the waves to give more context about the issue:
Screenshot from 2024-06-04 08-57-28

One can see on the waveform above the signals  `cmd_wr_en_q` and `cmd_wr_en_last_bit` get set on the posedge, but the signal is meant to change on the neg-edge

a-will:
https://github.com/lowRISC/opentitan/pull/23579 may be an alternative.

antmarzam:
Closing this PR since this is handled via a better fix in #23579

PR #23491, opened 2024-06-04
[bazel] Refactor device test execution environment assignments

jwnrt:
This PR updates as many `opentitan_test`s as possible to use `EARLGREY_TEST_ENVS` and `EARLGREY_SILICON_OWNER_ROM_EXT_ENVS`.

* If a test isn't designed for an environment, we remove it with `dicts.omit`.
* If a test is supposed to work but is broken on an environment, we tag `broken`.

I have tried my best not to _remove_ any environments from tests, only add them.

@engdoreis and I discussed some tests in `//sw/device/tests` that do not need to run on `fpga_cw310_test_rom` so removed that environment from those tests.

### DIFF
--- a/sw/device/tests/pmod/BUILD
+++ b/sw/device/tests/pmod/BUILD
@@ -5,11 +5,13 @@
 load(
     "//rules/opentitan:defs.bzl",
     "EARLGREY_SILICON_OWNER_ROM_EXT_ENVS",
+    "EARLGREY_TEST_ENVS",
     "cw310_params",
     "fpga_params",
     "opentitan_test",
     "silicon_params",
 )
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
 
 package(default_visibility = ["//visibility:public"])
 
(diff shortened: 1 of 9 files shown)


engdoreis (approved):
LGTM

sw/device/tests/pmod/BUILD
```diff
@@ -311,18 +313,20 @@ opentitan_test(
...
     srcs = ["i2c_host_fram_test.c"],
-    exec_env = {
-        "//hw/top_earlgrey:fpga_cw310_sival": None,
-        "//hw/top_earlgrey:silicon_owner_sival_rom_ext": "silicon_owner",
-        "//hw/top_earlgrey:fpga_cw310_test_rom": None,
-    },
+    exec_env = dicts.add(
+        EARLGREY_TEST_ENVS,
+        EARLGREY_SILICON_OWNER_ROM_EXT_ENVS,
+        {
```
engdoreis, line 319: Would drop it.

engdoreis (approved):

sw/device/tests/pmod/BUILD
jwnrt: Will do in a follow up since there are many tests that may have unnecessary `fpga_cw310_sival` environments.

It's not doing any harm either since it's not run in CI if `fpga_cw310_sival_rom_ext` is present.

PR #23494, opened 2024-06-04
[earlgrey, hmac] Add TLT for SHA-2/HMAC multistream

ballifatih:
Reorganize SHA-2/HMAC tests into two:

- One shot calls -> hmac_functest
- Multistream (a.k.a. interleaving streaming) test -> hmac_multistream_functest

also update the test plan and add the second test as TLT, so that it runs on nightlies.

### DIFF
--- a/hw/top_earlgrey/data/ip/chip_hmac_testplan.hjson
+++ b/hw/top_earlgrey/data/ip/chip_hmac_testplan.hjson
@@ -46,15 +46,13 @@
       name: chip_sw_hmac_all_configurations
       desc: '''Verify HMAC 256/384 and 512 modes of operation.
 
-            - Verify the following security configurations:
+            - Verify the following security configurations for 6 hardcoded vectors:
               - SHA2-256
               - SHA2-384
               - SHA2-512
               - HMAC-SHA2-256
               - HMAC-SHA2-384
               - HMAC-SHA2-512
-            - Verify input message size of 0 bytes.
-            - Verify multiple input message sizes.
             - Use endianness configuration used in the crypto library.
             '''
       features: [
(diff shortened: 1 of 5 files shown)


sw/device/tests/crypto/hmac_multistream_functest.c
```diff
@@ -0,0 +1,232 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// #include "sw/device/lib/crypto/drivers/hmac.h"
```
moidx, line 5: Remove commented out code.

sw/device/tests/crypto/hmac_multistream_functest.c
```diff
@@ -0,0 +1,232 @@
...
+ * i) Each stream is instantiated (by init call).
+ * ii) Each stream receives the first half its message.
+ * iii) Each stream receives the second half of its message.
+ * iv) Each stream is concluded with a final call and the digest result is
+ * compared with the expected valud.
+ */
+static status_t run_test(void) {
+  // hash_contexts are also used as hmac_contexts, since their size is equal.
+  otcrypto_hash_context_t hash_contexts[ARRAYSIZE(kHmacTestVectors)];
+  for (size_t i = 0; i < ARRAYSIZE(kHmacTestVectors); i++) {
```
moidx, line 211: Would there be any value from randomizing all these loops from a DV coverage perspective? If so we can talk about how to do that in a follow up PR.

moidx (approved):

sw/device/tests/crypto/hmac_multistream_functest.c
ballifatih: Yes randomizing the order here would be great. I was not sure if there are available utils regarding randomness and where to get this randomness. I see `sw/device/lib/base/random_order.h` but the implementation looks like a place holder.

At the moment, this test does not configure EDN to save some time, but we can reenable it if it helps with the randomness.

ballifatih: An alternative would be to make some bazel changes so that the hardcoded testvector file `sw/device/tests/crypto/testvectors/hmac_testvectors_random.hjson` is re-generated randomly with the following:
```
./sw/device/tests/crypto/hmac_gen_random_testvectors.py 10 sw/device/tests/crypto/testvectors/hmac_testvectors_random.hjson
```
The generation script does not take seed at the moment, but I can add it. It would put less pressure on complication to TLT and we can avoid configuring EDN I think.

In any case, I think we can merge this once CI is clean and discuss how to improve the test. It is also possible to run this test with more vectors on FPGA/nightly.

hw/top_earlgrey/data/ip/chip_hmac_testplan.hjson
```diff
@@ -64,26 +62,37 @@
...
+            - Verify that HMAC block can stream 6 SHA-2/HMAC operations in parallel:
+              - One test vector for each configuration:
+                - SHA2-256
+                - SHA2-384
+                - SHA2-512
+                - HMAC-SHA2-256
+                - HMAC-SHA2-384
+                - HMAC-SHA2-512
+              - For each vector, message input is split into two chunks, and each
+              operation is completed with 2 update calls.
```
gdessouky, line 81: In a future refinement in an upcoming PR, can we randomize the number of update calls, e.g., to range between [1, 4] for the different vectors?

hw/top_earlgrey/data/ip/chip_hmac_testplan.hjson
```diff
@@ -64,26 +62,37 @@
...
+                - Initiate all 6 operations.
+                - Feed the first message chunk to each streams, by resuming,
```
gdessouky, line 83: minor typo: each stream*

gdessouky (approved):
thanks @ballifatih, this LGTM (only left a minor typo comment)! 

I think we can discuss more enhancements to the TLT for future PRs but not block this one. It would be great if we can randomize the number of update calls that the 6 vectors end up making, e.g.,  so we end up with a mix where 1 may eventually be one-shot, others are making 2-3 calls, and others more, etc.

ballifatih:
> I think we can discuss more enhancements to the TLT for future PRs but not block this one. It would be great if we can randomize the number of update calls that the 6 vectors end up making, e.g., so we end up with a mix where 1 may eventually be one-shot, others are making 2-3 calls, and others more, etc. WDYT?

As discussed offline with @gdessouky, I created an issue to track the TLT improvements for M5 https://github.com/lowRISC/opentitan/issues/23507

For M4, the focus is to have another PR to update HMAC driver to increase testing coverage.

martin-velay (approved):
I don't know so much SW, but LGTM

PR #23498, opened 2024-06-04
[otbn,dv] Collect coverage from lc_escalate_en_i signal

rswarbrick:
This signal is not really synchronous with anything else: an escalation request can come at any time. Collect very basic coverage of whether it is Off, On or undefined (treated as On).

The "bad" bin for this covergroup is true whenever an unknown value is sent on the lc_escalate_en_i. We expect this to be treated as an escalation request (because of the lc_tx_test_true_loose call in otbn_core_model.sv), so completing coverage for this covergroup will check that we do indeed treat unknown values on lc_escalate_en_i as good escalation requests.

This should satisfy the first item in #23460.


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #23500, opened 2024-06-05
[pwrmgr] Don't lock CONTROL until observing core_sleep

a-will:
To avoid race conditions with WFI and locking the CONTROL CSR prematurely, wait until the transition out of FastPwrStateActive that begins low power entry before locking the CONTROL CSR.
    
This allows software to clear CONTROL.LOW_POWER_HINT if an early interrupt caused the "core_sleep" pulse to be too short for pwrmgr to capture. Then, subsequent WFI instructions may still choose between full power and a low power mode.
    
When software wakes up, it must attempt to clear CONTROL.LOW_POWER_HINT, then busy-poll until it reads its value as 0. If pwrmgr triggered low power entry, the CPU will poll until pwrmgr completes the low power sequence. If pwrmgr did not trigger low power entry, the CPU's write will succeed, and the read will immediately return 0.

Also add a test that causes Ibex to decode WFI with an IRQ pending and not issue core_sleep (modified from a submission from @ecgh0). Check for correct recovery of the CONTROL CSR and that software can clear LOW_POWER_HINT.

### DIFF
--- a/hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr.sv
+++ b/hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr.sv
@@ -274,6 +274,7 @@ module pwrmgr
   prim_mubi_pkg::mubi4_t rom_ctrl_good;
 
   logic core_sleeping;
+  logic low_power_entry;
 
   ////////////////////////////
   ///  clk_slow_i domain declarations
@@ -339,7 +340,7 @@ module pwrmgr
       lowpwr_cfg_wen <= 1'b1;
     end else if (!lowpwr_cfg_wen && (clr_cfg_lock || wkup)) begin
       lowpwr_cfg_wen <= 1'b1;
-    end else if (low_power_hint) begin
+    end else if (low_power_entry) begin
       lowpwr_cfg_wen <= 1'b0;
     end
   end
@@ -577,6 +578,7 @@ module pwrmgr
   ////////////////////////////
 
   assign low_power_hint = reg2hw.control.low_power_hint.q == LowPower;
+  assign low_power_entry = core_sleeping & low_power_hint;
 
   pwrmgr_fsm u_fsm (
     .clk_i,
@@ -590,7 +592,7 @@ module pwrmgr
     .ack_pwrup_o         (ack_pwrup),
     .req_pwrdn_o         (req_pwrdn),
     .ack_pwrdn_i         (ack_pwrdn),
-    .low_power_entry_i   (core_sleeping & low_power_hint),
+    .low_power_entry_i   (low_power_entry),
     .reset_reqs_i        (peri_reqs_masked.rstreqs),
     .fsm_invalid_i       (fsm_invalid),
     .clr_slow_req_o      (clr_slow_req),
--- a/hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_sec_cm_ctrl_config_regwen_vseq.sv
+++ b/hw/ip_templates/pwrmgr/dv/env/seq_lib/pwrmgr_sec_cm_ctrl_config_regwen_vseq.sv
@@ -3,9 +3,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // Decription:
 // Create low power transition and wakeup a few times.
-// When PWRMGR.CONTROL.LOW_POWER_HINT is set,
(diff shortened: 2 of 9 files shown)


a-will:
@johannheyszl @vogelpi I tagged you because this modifies the behavior of the PWRMGR.CTRL.CONFIG.REGWEN countermeasure. Instead of locking immediately upon writing a 1 to CONTROL.LOW_POWER_HINT, the lock now happens when pwrmgr begins the low power entry sequence.

a-will:
Thanks, @matutem, for the great idea to defer the locking!

hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr.sv
```diff
@@ -339,7 +340,7 @@ module pwrmgr
       lowpwr_cfg_wen <= 1'b1;
     end else if (!lowpwr_cfg_wen && (clr_cfg_lock || wkup)) begin
       lowpwr_cfg_wen <= 1'b1;
-    end else if (low_power_hint) begin
+    end else if (low_power_entry) begin
```
a-will, line 343: If it's important to have the configuration locked for the synchronization into the slow domain, we can probably do something about that. I noticed that is one other visible bit that is dropped.

If it's not important, we can move on!

moidx (approved):

GregAC:
Nice work @a-will.

I'd just like to double check my understanding of the scenarios.

When you set the lower power hint and then execute the `WFI` you have two possibilities:

1. `low_power_entry_i` in `pwrmgr_fsm` is asserted for at least one cycle (low power hint has set and Ibex indicated sleep for enough cycles for a high value to make it over the CDC), so you transition out of `FastPwrStateActive` and clocks will be disabled. The hint will be cleared as part of this process. You then make your way back to `FastPwrStateActive` either due to a fall through or because you've actually slept.

2. `low_power_entry_i` in `pwrmgr_fsm` has not yet been asserted and will not be asserted until you execute another `WFI`. The hint will still be set as we haven't gone through the FSM path that clears it and there's no lock so software can instantly clear the lower power hint.

The tricky bit is in 1. after leaving `FastPwrStateActive` we begin to disable clocks but it is possible that Ibex could leave sleep before this is done and begin executing code following WFI. In this case we've got a fall through so the clocks will disable briefly and then re-enable. It's this the busy-poll deals with. If we've hit this case then the hint clear after WFI will be a no-op (it will be locked at this point) however we're going to fall through so it'll clear itself. The busy poll then effectively waits until the fallthrough has occurred and we're back to Active with no hint set.

Have I got that right?

a-will:
> Nice work @a-will.
> 
> I'd just like to double check my understanding of the scenarios.
> 
> When you set the lower power hint and then execute the `WFI` you have two possibilities:
> 
> 1. `low_power_entry_i` in `pwrmgr_fsm` is asserted for at least one cycle (low power hint has set and Ibex indicated sleep for enough cycles for a high value to make it over the CDC), so you transition out of `FastPwrStateActive` and clocks will be disabled. The hint will be cleared as part of this process. You then make your way back to `FastPwrStateActive` either due to a fall through or because you've actually slept.
> 2. `low_power_entry_i` in `pwrmgr_fsm` has not yet been asserted and will not be asserted until you execute another `WFI`. The hint will still be set as we haven't gone through the FSM path that clears it and there's no lock so software can instantly clear the lower power hint.
> 
> The tricky bit is in 1. after leaving `FastPwrStateActive` we begin to disable clocks but it is possible that Ibex could leave sleep before this is done and begin executing code following WFI. In this case we've got a fall through so the clocks will disable briefly and then re-enable. It's this the busy-poll deals with. If we've hit this case then the hint clear after WFI will be a no-op (it will be locked at this point) however we're going to fall through so it'll clear itself. The busy poll then effectively waits until the fallthrough has occurred and we're back to Active with no hint set.
> 
> Have I got that right?

Yup! That's right on all counts. :smile:

GregAC:
Oh and we should have a follow PR that updates the pwrmgr docs to explain the proper sleep proceedure

hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr.sv
vogelpi: I think it should be okay.

vogelpi (commented):
Post merge but this looks good to me, thanks @a-will and @matutem !

PR #23503, opened 2024-06-05
[flash/dv] Fix constraint error in flash_ctrl_config_regwen test

eshapira:
Fix constraints contradiction.
See constraint ctrl_num_c in file flash_ctrl_otf_base_vseq line 95 (at 6/05/2024).
   if (rand_op.partition == FlashPartData) ctrl_num == ctrl_data_num;

### DIFF
--- a/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_config_regwen_vseq.sv
+++ b/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_config_regwen_vseq.sv
@@ -43,6 +43,11 @@ class flash_ctrl_config_regwen_vseq extends flash_ctrl_otf_base_vseq;
 
     // Launch long operation
     flash_program_data_c.constraint_mode(0);
+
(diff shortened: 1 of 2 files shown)


sha-ron (approved):

rswarbrick (approved):
I guess that seems sensible, although it's sad to have to reimplement the constrained randomisation like this.

I'd probably suggest adding a comment that above the line, explaining that this is the equivalent of the `ctrl_num_c` constraint, because `rand_op.partition` is `FlashPartData`. I'd also move this line to be just after the randomisation of `rand_op`, which will make it a bit more obvious what's going on.

eshapira:
> I guess that seems sensible, although it's sad to have to reimplement the constrained randomisation like this.
> 
> I'd probably suggest adding a comment that above the line, explaining that this is the equivalent of the `ctrl_num_c` constraint, because `rand_op.partition` is `FlashPartData`. I'd also move this line to be just after the randomisation of `rand_op`, which will make it a bit more obvious what's going on.

In order to avoid contradiction, it has to be before the randomization, but I did removed the separating line and added an explanation.
Thanks!

PR #23504, opened 2024-06-05
[hmac,dv] Minor DV assertion fix

gdessouky:
This adds checking `reg_hash_stop` in the `ValidHashProcessAssert` assertion for completeness to make sure that `hash_process || reg_hash_stop `are toggled with `hash_start_or_continue`.

Block-level tests at the moment without SAR will not complain of this change. @ballifatih can you plz confirm this works the same at your end with your TLT (though this is a very last priority atm)?

### DIFF
--- a/hw/ip/hmac/rtl/hmac.sv
+++ b/hw/ip/hmac/rtl/hmac.sv
@@ -899,9 +899,9 @@ module hmac
   // are used in Assertion only
   logic in_process;
   always_ff @(posedge clk_i or negedge rst_ni) begin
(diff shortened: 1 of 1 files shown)


martin-velay (approved):

johannheyszl (approved):
Note that the RTL change is in a no synthesize section where the signal is only used in assertions and will never be synthesized in the chip.

vogelpi (approved):
LGTM. Sorry @gdessouky , I started reviewing this last week but then got interrupted and completely forgot about it again.

PR #23505, opened 2024-06-05
[ci] Introduce a test orchestrator written in bazel

pamaury:
Prototype of the test orchestrator as discussed in the software WG. The implementation follows what we have discussed: `opentitan_test` calls a `ci_orchetrator` function with the list of all execution environments. This function then decides which environment(s) to run in CI and returns this list. Then `opentitan_test` marks every test **not in that list** as `skip_in_ci`.

This PR also removes the old complicated logic in the azure pipeline and fpga runner script. This is done by moving the FPGA test job to its own template file to avoid duplication.


rules/opentitan/ci.bzl
```diff
@@ -0,0 +1,47 @@
...
+    "//hw/top_earlgrey:fpga_cw310_sival_rom_ext",
+    "//hw/top_earlgrey:fpga_cw310_rom_ext",
+    "//hw/top_earlgrey:fpga_cw310_sival",
+    "//hw/top_earlgrey:fpga_cw310_rom_with_fake_keys",
+    "//hw/top_earlgrey:fpga_cw310_test_rom",
+    "//hw/top_earlgrey:fpga_cw340_sival_rom_ext",
+    "//hw/top_earlgrey:fpga_cw340_sival",
+    "//hw/top_earlgrey:fpga_cw340_rom_ext",
+    "//hw/top_earlgrey:fpga_cw340_rom_with_fake_keys",
+    "//hw/top_earlgrey:fpga_cw340_test_rom",
```
jwnrt, line 25: I think we want to put the CW340s on top of the CW310s here

jwnrt (commented):
Where do you think would be the best place to document this?

I can imagine confusion if somebody adds a new test or execution environment and wonders why it's not running in CI

rules/opentitan/ci.bzl
pamaury: Do we have enough CW340 runners? Because this will switch almost the entire loads to the CW340

jwnrt:
How should the orchestrator interact with the `broken` tag? Should it look for this and try to run the next available environment that isn’t broken?

rules/opentitan/ci.bzl
jwnrt: I’ve added two more so we have 6 CW340s which should be enough.

pamaury:
> How should the orchestrator interact with the `broken` tag? Should it look for this and try to run the next available environment that isn’t broken?

Yes that's my thinking (although it's not implemented yet). Otherwise it wiuld require a lot of manual fiddling for broken tests.

rules/opentitan/ci.bzl
```diff
@@ -0,0 +1,51 @@
...
+            the_one_in_ci = env
+            break
+
+    if the_one_in_ci == None:
+        # FIXME is that an error?
+        print("WARNING: in test {}, no execution environment was found that runs in CI (except for those always enabled)")
+
+    run_in_ci = [env for env in _ALWAYS_RUN_IN_CI if sets.contains(exec_env_sets, env)]
+    if the_one_in_ci != None:
+        run_in_ci.append(the_one_in_ci)
```
engdoreis, line 50: Perhaps this should be an `else` of the previous `if`?

engdoreis (commented):
LGTM, will approve when is marked as Ready to review.

rules/opentitan/ci.bzl
```diff
@@ -0,0 +1,43 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+load("@bazel_skylib//lib:sets.bzl", "sets")
+
+# List of execution environments that are always enabled in CI.
+_ALWAYS_RUN_IN_CI = [
```
nbdd0121, line 8: Instead of returning a list of environments to run CI, perhaps it should return the skip_for_ci list? In that case we don't need this list.

azure-pipelines.yml
```diff
@@ -534,240 +534,41 @@ jobs:
...
-  - template: ci/load-bazel-cache-write-creds.yml
-  - bash: |
-      set -e
-      . util/build_consts.sh
-      module load "xilinx/vivado/$(VIVADO_VERSION)"
-      ci/scripts/run-fpga-tests.sh cw310 manuf,-cw310_sival,-broken || { res=$?; echo "To reproduce failures locally, follow the instructions at https://opentitan.org/book/doc/getting_started/setup_fpga.html#reproducing-fpga-ci-failures-locally"; exit "${res}"; }
-    displayName: Execute tests
-  - template: ci/publish-bazel-test-results.yml
-
+# CW310 FPGA jobs.
```
nbdd0121, line 537: This is no longer matrix and the commit message should be updated.

rules/opentitan/ci.bzl
pamaury: Thanks for the suggestion, I will implement that.

pamaury: I have updated the code.

pamaury:
One of the issue I am running into is that it can happen that some jobs end up having no tests to run due to the orchestration. Unfortunately, bazel will error out if the test tag filters produces 0 tests and I don't think this behaviour can be turned off.

jwnrt:
Is there a way to ask Bazel for a list of tests before running them? Then we can have a pre-step which exits early if there are none

jwnrt:
That sounds good, but I had actually thought that was what your current implementation does. Each test is tagged by its execution environment right? And the Azure jobs run the tests with that exec env tag -broken,-skip_in_ci.

Wouldn’t this still have the same problem though? The job applies these filters but gets zero tests back and the job fails?

pamaury:
Currently it is not quite the case that each job corresponds to an execution environment because we have manufacturing tests that run in their own job.

To make it clear what I am suggesting, there would be bazel function:
```python
def ci_fpga_job(env,tags):
   """Return the FPGA job on which this tests should run, based on its the execution environment
   and its tags.
  """
   if "cw340_sival_rom_ext" in tag and not "manuf" in tags:
      return "fpga_job_blabla"
   # and so on
```
and in `rules/defs.bzl` we could call this function to add the tag to the list. Then the azure file simply specifies the (unique) tag of each fpga job.

nbdd0121:
What's the reason that we currently run manufacturing tests as separate jobs? Probably just because we want to split into more jobs for parallelism?

jwnrt:
I think it might be because they have custom execution environments since they run at unusual OTP stages (`test_unlocked`, `rma`, etc.). If we mix them into other tests, we might be reprogramming bitstreams more times in the middle of the run which will slow things down slightly.

rules/opentitan/ci.bzl
jwnrt: I'll swap the order of the CW310s and CW340s later when I enable more tests.

ci/fpga-job.yml
```diff
@@ -0,0 +1,93 @@
...
+        - sw_build
+  - template: ./load-bazel-cache-write-creds.yml
+  # We run the update command twice to workaround an issue with udev on the container,
+  # where rusb cannot dynamically update its device list in CI (udev is not completely
+  # functional). If the device is in normal mode, the first thing that opentitantool
+  # does is to switch it to DFU mode and wait until it reconnects. This reconnection is
+  # never detected. But if we run the tool another time, the device list is queried again
+  # and opentitantool can finish the update. The device will now reboot in normal mode
+  # and work for the hyperdebug job.
+  # FIXME: only execute for hyper310 interface
```
jwnrt, line 64: I wonder if `opentitanlib` should accept `update-firmware` for all platforms but make it a no-op if not hyperdebug? Erroring seems a bit extreme.

Even then, I'm not sure how to manually specify `--interface hyperdebug_dfu` for these...

If the update step is really quick at detecting when no update is required, maybe we can move the update step into the commands for the execution environments?

jwnrt (approved):
This looks really good bar the hyperdebug update problem and the broken tests. Very nice work.

ci/fpga-job.yml
pamaury: Yes we discussed that in a previous PR and decided to do it later... This would be very convenient for this PR!

sw/device/tests/BUILD
```diff
@@ -4615,6 +4615,7 @@ test_suite(
             otp = test_cfg["otp"],
             tags = [
                 "lc_{}".format(test_cfg["lc_state"]),
+                "broken",
```
nbdd0121, line 4618: I think only the `delayed_enabled`, the other variants of this test is working fine.

pamaury: fixed

pamaury:
@cfrantz I have marked the relevant rom_ext e2e tests are broken in this PR. I suggest we mark them as unbroken when we backport your fixes in #24430

ci/fpga-job.yml
pamaury: I have now implemented the update step: it only runs for the `hyper310` interface. I will let the CI run to make sure that it works.

azure-pipelines.yml
```diff
@@ -518,275 +518,62 @@ jobs:
...
-  - bash: |
-      set -e
-      . util/build_consts.sh
-      module load "xilinx/vivado/$(VIVADO_VERSION)"
-      ci/scripts/run-fpga-tests.sh hyper310 cw310_sival_rom_ext || { res=$?; echo "To reproduce failures locally, follow the instructions at https://opentitan.org/book/doc/getting_started/setup_fpga.html#reproducing-fpga-ci-failures-locally"; exit "${res}"; }
-    displayName: Execute tests
-  - template: ci/publish-bazel-test-results.yml
+# CW310 FPGA jobs.
+- template: ci/fpga-job.yml
+  parameters: { job_name: execute_test_rom_fpga_tests_cw310, display_name: CW310 Test ROM Tests, bitstream: chip_earlgrey_cw310,
```
nbdd0121, line 523: Can this be multiple lines? It's not very easy to read using inline JSON syntax.

ci/fpga-job.yml
```diff
@@ -0,0 +1,93 @@
...
+      ci/scripts/run-bazel-test-query.sh \
+        "${{ parameters.target_pattern_file }}" \
+        "${{ parameters.tag_filters }}",-manual,-broken,-skip_in_ci \
+        //... @manufacturer_test_hooks//...
+      # Run FPGA tests.
+      if [ -s "${{ parameters.target_pattern_file }}" ]; then
+        ci/scripts/run-fpga-tests.sh "${{ parameters.interface }}" "${{ parameters.target_pattern_file }}" || { res=$?; echo "To reproduce failures locally, follow the instructions at https://opentitan.org/book/doc/getting_started/setup_fpga.html#reproducing-fpga-ci-failures-locally"; exit "${res}"; }
+      else
+        echo "No tests to run after filtering"
+      fi
```
nbdd0121, line 87: Can this logic be part of run-fpga-tests.sh?

azure-pipelines.yml
pamaury: It can be. I personally find that one item per line with the various jobs actually made it harder to read because it spreads many screens but I can change it.

ci/fpga-job.yml
pamaury: I would prefer not because this way, we have two scripts for two different purposes: run-bazel-test-query does the query, and run-fpga-tests run the tests. You can more easily repurpose the scripts this way. Or I would make it third script if you really want to (run-fpga-tests-query).

azure-pipelines.yml
pamaury: I have updated the code to have one item per line

nbdd0121 (approved):

PR #23508, opened 2024-06-05
[otbn,dv] Collect coverage for RMA requests

rswarbrick:
The point of this is to check that we're handling invalid RMA request values correctly (by ignoring them). Invalid values land in the "bad" bin for the covergroup and completing coverage for the covergroup will ensure that we have genuinely seen this invalid value.

The code in otbn_core_model.sv will ignore that request (using lc_tx_test_true_strict to check it), so the test will fail if the RTL acts on the request.

This should satisfy the second item in #23460.


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #23510, opened 2024-06-05
[hmac, driver] Improve HMAC driver

ballifatih:
Improves HMAC driver with the following changes, so that we have better coverage on HMAC RTL:

- Instead of writing messages byte at a time, try to write them in words as much as possible.
- Previously, oneshot HMAC/SHA-2 call was internally translated to `init->update->final` calls. With this PR, the oneshot calls do not force unnecessary context switches. They are done in one go as they should be.
- Recently fixed/added Idle status bit of HMAC is now used, instead of waiting `hmac_done` interrupt to fire. Both could be used, but we want to use Idle to make sure that it is working.
- Some code formatting + new functions to avoid duplicating same code in one-shot and streaming calls.

cc: @gdessouky

### DIFF
--- a/sw/device/lib/crypto/drivers/hmac.c
+++ b/sw/device/lib/crypto/drivers/hmac.c
@@ -198,11 +198,27 @@ static void context_save(hmac_ctx_t *ctx) {
 /**
  * Write given byte array into the `MSG_FIFO`. This function should only be
  * called when HWIP is already running and expecting further message bytes.
+ *
+ * @param message The incoming message buffer to be fed into HMAC_FIFO.
+ * @param message_len The length of `message` in bytes.
  */
-static void msg_fifo_write(const uint8_t *msg, size_t msg_len) {
-  // TODO(#23191): Improve message copying by writing them in 32b at a time.
-  for (size_t i = 0; i < msg_len; i++) {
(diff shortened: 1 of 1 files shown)


sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -65,52 +65,57 @@ OT_ASSERT_ENUM_VALUE(HMAC_DIGEST_14_REG_OFFSET, HMAC_DIGEST_13_REG_OFFSET + 4);
 OT_ASSERT_ENUM_VALUE(HMAC_DIGEST_15_REG_OFFSET, HMAC_DIGEST_14_REG_OFFSET + 4);
 
 /**
- * The following values are the only key sizes supported natively bw HWIP.
+ * The following values are the only key sizes supported natively by HMAC HWIP.
```
moidx, line 68: nit: Please update this comment. The key sizes are no longer defined below.

moidx (approved):

sw/device/lib/crypto/drivers/hmac.c
ballifatih: There was a small bug with oneshot calls. As I was fixing that I also removed this comment.

PR #23512, opened 2024-06-05
[prim/rtl] Allow for tech-specific impls of `prim_flop_2sync`

andreaskurth:
This allows for technology/library-specific implementations of `prim_flop_2sync`.  The generic implementation is based on two back-to-back `prim_flop`s, which is identical to how `prim_flop_2sync` was implemented before.

This PR doesn't add FPGA-specific implementations; this will be tracked in an issue and might get implemented later.

### TODOs
- [ ] Create issue to track FPGA impl and link this PR.

### DIFF
--- a/hw/ip/entropy_src/dv/env/entropy_src_env_cfg.sv
+++ b/hw/ip/entropy_src/dv/env/entropy_src_env_cfg.sv
@@ -271,7 +271,7 @@ class entropy_src_env_cfg extends cip_base_env_cfg #(.RAL_T(entropy_src_reg_bloc
        0: path = {path, ".i_sync_n"};
        1: path = {path, ".i_sync_p"};
      endcase
-     disabled_prim_cdc_rand_delays[i] = {path, ".u_prim_cdc_rand_delay"};
+     disabled_prim_cdc_rand_delays[i] = {path, ".gen_generic.u_impl_generic.u_prim_cdc_rand_delay"};
     end
   endfunction
 
--- a/hw/ip/lc_ctrl/dv/cov/lc_ctrl_volatile_unlock_disabled_unr_exclude.el
+++ b/hw/ip/lc_ctrl/dv/cov/lc_ctrl_volatile_unlock_disabled_unr_exclude.el
@@ -9,7 +9,7 @@
 // ExclMode: default
 //==================================================
 CHECKSUM: "1599929061 1913160869"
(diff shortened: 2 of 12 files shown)


a-will (approved):
LGTM

For the FPGAs, it's okay for the generic implementation to be used, I think.

moidx (approved):

a-will:
Looks like some FPGA constraints need updated paths.

andreaskurth:
> Looks like some FPGA constraints need updated paths.

Should be fixed now :crossed_fingers:

andreaskurth:
DV problems should now also be fixed

andreaskurth:
@meisnere @OTshimeon: This allows for `prim_flop_2sync` to have an implementation that is specific to a silicon technology and standard-cell library. Please implement the corresponding `prim_<techlib>_flop_2sync` on your side and don't hesitate to contact me if you have questions.

andreaskurth:
I cannot reproduce the remaining DV CI failures; the tests pass for me locally.

andreaskurth:
I actually was able to reproduce the failing OTBN test after purging the `scratch/` directory. I've now added a commit to fix that failing test. It's about `.core` files, which somehow seem to have changed in dependency due to the main commit in this PR.

moidx:
All green. Merging to pick this up in the next regression. Thanks @andreaskurth!

Razer6:
@andreaskurth  @a-will `integrated_dev` solved that particular issue differently. There is a tech-dependent `prim_TECH_sync_cell` that is instantiated in `prim_flop_2sync`. Would it make sense to backport those changes to not have another piece of hardware that diverges?

andreaskurth:
> @andreaskurth @a-will `integrated_dev` solved that particular issue differently. There is a tech-dependent `prim_TECH_sync_cell` that is instantiated in `prim_flop_2sync`. Would it make sense to backport those changes to not have another piece of hardware that diverges?

I think ideally we would instantiate a prim named `prim_flop_sync` which has a generic implementation and optional tech-specific implementations. `_2sync` would be a misnomer if the depth becomes a parameter (which is possible in the future). I'd keep `flop` in the module name to indicate that it's a FF-based synchronizer (rather than a synchronizer based on a FIFO and graypointers or a handshake).

I currently don't see the case for instantiating a tech-specific prim *inside* `prim_flop_2sync`, though. Is there a strong reason / preference for that?

Issue #23150, opened 2024-05-16
[ROM] Implement AST alert reset configuration and post-reset behavior


andreaskurth:
As described in [this implementation recommendation](https://docs.google.com/document/d/1KTAkLexrTPsEyJrMc6Auxiuxdv161GdEnyTbjLDY6Fw/edit): At reset we want all alerts in sensor_ctrl to be enabled and set to recoverable. During ROM boot and _before_ alert_handler is configured, we want ROM to read OTP values (from the `OWNER_SW_CFG` OTP partition) to disable specific alerts in sensor_ctrl or set them to fatal.

## TODOs

- [ ] Extend ROM to read the OTP values, verify their digest, and configure sensor_ctrl accordingly, then wait a few cycles (20 cycles in the slower clock domain of sender and receiver should be sufficient) for outstanding handshakes to complete before configuring alert_handler.
- [ ] Mention this design in ROM’s documentation.

Note that this issue is about changes to ROM. See #23149 for the changes to sensor_ctrl RTL and OTP partitions.

cfrantz:
> then wait a few cycles (20 cycles in the slower clock domain of sender and receiver should be sufficient) for outstanding handshakes to complete before configuring alert_handler.

I assume the slower clock domain is the AON domain.  Can I simply use the PWRMGR_CFG_CDC_SYNC register to perform this wait, or is it preferable to use one of the counters in the AON_TIMER peripheral?

andreaskurth:
Using PWRMGR_CFG_CDC_SYNC is probably simpler

andreaskurth:
Closing as completed in #23513, thx @cfrantz!

Issue #21871, opened 2024-03-06
[ROM] Consider `sensor_ctrl` alerts filter 


moidx:
### Description

https://github.com/lowRISC/opentitan/pull/21870 added support for filtering `sensor_ctrl` alerts to handle spurious alerts coming from AST. Since the ROM has the ability to configure the alert handler via OTP settings, we should consider adding a mechanism to setup the `sensor_ctrl` alerts as part of the process.

moidx:
HI @andreaskurth, is this issue now superseded by #23150? If so, please feel free to close this issue. Thanks

cfrantz:
ROM sensor-ctrl configuration was added by #23513.

PR #23513, opened 2024-06-05
Rom sensor ctrl

cfrantz:
Add a sensor_ctrl driver to configure the peripheral from OTP.
Configure sensor_ctrl during ROM bootup.

Addresses: #23150, #21871.

### DIFF
--- a/hw/ip/otp_ctrl/data/BUILD
+++ b/hw/ip/otp_ctrl/data/BUILD
@@ -234,6 +234,13 @@ otp_json(
                 "OWNER_SW_CFG_ROM_EXT_BOOTSTRAP_EN": otp_hex(CONST.HARDENED_FALSE),
                 "OWNER_SW_CFG_ROM_SRAM_READBACK_EN": otp_hex(CONST.MUBI4_FALSE),
                 "OWNER_SW_CFG_ROM_FLASH_ECC_EXC_HANDLER_EN": otp_hex(CONST.HARDENED_TRUE),
+                # By default, the sensor_ctrl should disable all sensors and mark
+                # alerts as recoverable.
+                "OWNER_SW_CFG_ROM_SENSOR_CTRL_ALERT_CFG": [
+                    otp_hex(0x69696969),
+                    otp_hex(0x69696969),
+                    otp_hex(0x69696969),
+                ],
             },
         ),
     ],
--- a/sw/device/silicon_creator/lib/drivers/pwrmgr.c
+++ b/sw/device/silicon_creator/lib/drivers/pwrmgr.c
@@ -6,6 +6,7 @@
 
 #include "sw/device/lib/base/abs_mmio.h"
 #include "sw/device/silicon_creator/lib/base/sec_mmio.h"
+#include "sw/device/silicon_creator/lib/drivers/ibex.h"
 
 #include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
 #include "pwrmgr_regs.h"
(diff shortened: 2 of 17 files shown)


hw/ip/otp_ctrl/data/BUILD
```diff
@@ -232,6 +232,13 @@ otp_json(
                 ),
                 # By default, ROM_EXT's bootstrap feature should be disabled.
                 "OWNER_SW_CFG_ROM_EXT_BOOTSTRAP_EN": otp_hex(CONST.HARDENED_FALSE),
+                # By default, the sensor_ctrl should disable all sensors and mark
+                # alerts as recoverable.
+                "OWNER_SW_CFG_ROM_SENSOR_CTRL_ALERT_CFG": [
```
moidx, line 237: Can you also add this to the `sival` and `prodc` SKUs?

sw/device/silicon_creator/lib/drivers/pwrmgr.c
```diff
@@ -23,13 +24,27 @@ static_assert(
...
+  cpu_cycle_timeout += cpu_cycle_timeout / 4;
+
+  while (n--) {
+    ibex_mcycle_zero();
+    abs_mmio_write32(kBase + PWRMGR_CFG_CDC_SYNC_REG_OFFSET, kSyncConfig);
+    while (abs_mmio_read32(kBase + PWRMGR_CFG_CDC_SYNC_REG_OFFSET)) {
+      if (ibex_mcycle32() > cpu_cycle_timeout)
+        break;
+    }
+  }
```
moidx, line 42: If there is a concern with an attacker bypassing synchronization, can you add something like this to try to detect this?

```c
if(launder32(abs_mmio_read32(kBase + PWRMGR_CFG_CDC_SYNC_REG_OFFSET))) {
  HARDENED_CHECK_GE(ibex_mcycle32() > cpu_cycle_timeout);
}
```

moidx (approved):

hw/ip/otp_ctrl/data/BUILD
cfrantz: Done.

sw/device/silicon_creator/lib/drivers/pwrmgr.c
```diff
@@ -23,13 +24,27 @@ static_assert(
     PWRMGR_RESET_EN_EN_FIELD_WIDTH == 1,
     "RESET_EN field width changed, kAllResetsEnable may need to be updated");
 
+void pwrmgr_cdc_sync(uint32_t n) {
+  // We want to timeout if the CDC bit doesn't self clear.  We calculate the
+  // number of CPU cycles of 1 clock in the AON domain and add an additional
+  // 25 percent.
+  uint32_t cpu_cycle_timeout =
+      (uint32_t)kClockFreqCpuHz / (uint32_t)kClockFreqAonHz;
+  cpu_cycle_timeout += cpu_cycle_timeout / 4;
```
a-will, line 33: That's not nearly long enough. It takes up to 3 AON cycles + 2 "fast domain" cycles after the store is delivered to pwrmgr for the sync to make it into the AON domain, then another 4 "fast domain" cycles + 1 AON cycle to go back in the other direction.

The "fast domain" cycles will likely be absorbed by the CPU and interconnect, but I'd probably go with at least the 4 AON cycles here (maybe 5, for additional margin for future tops).

sw/device/silicon_creator/lib/drivers/pwrmgr.c
```diff
@@ -23,13 +24,25 @@ static_assert(
...
 
+void pwrmgr_cdc_sync(uint32_t n) {
+  // We want to timeout if the CDC bit doesn't self clear.  It should take
+  // approximately 4 AON ticks to complete.  We wait 25% longer (5 ticks).
+  uint32_t cpu_cycle_timeout =
+      (uint32_t)kClockFreqCpuHz / (uint32_t)kClockFreqAonHz * 5;
+
+  while (n--) {
+    ibex_mcycle_zero();
+    abs_mmio_write32(kBase + PWRMGR_CFG_CDC_SYNC_REG_OFFSET, kSyncConfig);
```
moidx, line 35: To avoid hardware lockups, we will probably have to perform read before write to ensure we don't write a 1 if this register is already set.

sw/device/silicon_creator/lib/drivers/pwrmgr.c
cfrantz: Done.  There are 4 uses of the CDC_SYNC operation in the ROM:

1. One in rom_start.S during some early init.  It just sets the register and doesn't wait.
2. Watchdog configuration synchronizes twice.
3. Pwrmgr synchronizes once after enabling all reset sources.
4. Sensor_ctrl synchronizes after configuring the sensor severity.  I've been instructed that I need to wait "20 clocks in the slow domain".  As such, sensor_ctrl synchronizes 5 times.

sw/device/silicon_creator/lib/drivers/pwrmgr.c
cfrantz: Done.

cfrantz:
The CI test failure is unrelated to this test:

```
t="//sw/device/tests:spi_device_sleep_test_fpga_cw310_rom_with_fake_keys"
# On this branch and on master:
BITSTREAM=1459afe788072bd95e185225f2b0179b7aed9065 bazel test $t  # Fail
BITSTREAM=73c42574957fad3c99aee16cce2bfcbb720e0d04 bazel test $t  # Pass
```

PR #23515, opened 2024-06-05
[flash_ctrl/rtl] Enable SecFifoPtr for TL-UL FIFOs

nasahlpa:
This commit enables the .Secure() option of the prim_fifo_sync FIFOs inside the tlul_adapter_sram used by the flash_ctrl module.

On an error, the intg_error_o signal will be flagged.


johannheyszl (approved):
Thanks @nasahlpa !
This enables an additional security feature we already had for the blocks we can use it. 

Area impact should be super low because it concerns the very small fifo ctrs only.

Code change is small: parameter change to switch on transparent hardening; additional error signals; additional assertions

vogelpi (approved):
LGTM, thanks @nasahlpa !

PR #23520, opened 2024-06-05
[rom] Reclaim 662 bytes of unused space

cfrantz:
We have been padding the `.crt` section out to consume the whole of `epmp_reset_rx_size` (the size of the initial EPMP executable region coming out of reset).

Reclaim this unused space for ROM code.


moidx (approved):
We discussed offline that this padding is probably no longer needed as we enable the rest of the ROM region for execution, and a potential attacker would only need to wait for a little longer for the ePMP region to be expanded after ROM starts execution. 

CC: @johannheyszl for vis on this change.

mundaym (approved):
Seems reasonable to me. Longer term we might want to consider determining the reset values after the rom image is created as that would be a bit cleaner and we could then also get rid of some of the code setting up the regions.

PR #23521, opened 2024-06-05
[rom] Print a banner at startup 

cfrantz:
1. Print `OpenTitan:<lifecycle_hw_rev>` and `GIT:<git revision>` at startup.
2. The UART FIFO size was reduced from 128 to 32 for the TX FIFO, but the shutdown function assumed a 128 byte FIFO.
3. Because of the smaller FIFO, we cannot emit all of the fault info into the FIFO in one shot.  Move the UART wait function into shutdown_print.

### DIFF
--- a/sw/device/silicon_creator/lib/drivers/uart.c
+++ b/sw/device/silicon_creator/lib/drivers/uart.c
@@ -97,14 +97,30 @@ int uart_getchar(uint32_t timeout_ms) {
   return n ? (int)ch : -1;
 }
 
-size_t uart_write(const uint8_t *data, size_t len) {
-  size_t total = len;
+void uart_write(const void *data, size_t len) {
+  const uint8_t *d = (const uint8_t *)data;
   while (len) {
-    uart_putchar(*data);
-    data++;
+    uart_putchar(*d++);
     len--;
   }
-  return total;
+}
+
+void uart_write_hex(uint32_t val, size_t len, uint32_t after) {
+  HARDENED_CHECK_LE(len, sizeof(uint32_t));
+  static const uint8_t kHexTable[16] = "0123456789abcdef";
+  size_t i = len * 8;
+  do {
+    i -= 4;
+    uart_putchar(kHexTable[(val >> i) & 0xF]);
+  } while (i > 0);
+  uart_write_imm(after);
+}
+
+void uart_write_imm(uint64_t imm) {
+  while (imm) {
+    uart_putchar(imm & 0xFF);
+    imm >>= 8;
+  }
 }
 
 size_t uart_read(uint8_t *data, size_t len, uint32_t timeout_ms) {
@@ -143,5 +159,9 @@ hardened_bool_t uart_break_detect(uint32_t timeout_us) {
 
 size_t uart_sink(void *uart, const char *data, size_t len) {
   (void)uart;
-  return uart_write((const uint8_t *)data, len);
+  uart_write((const uint8_t *)data, len);
+  return len;
 }
+
+// Provide link locations for inline functions.
+extern void uart_write_newline(void);
--- a/sw/device/silicon_creator/rom/rom.c
+++ b/sw/device/silicon_creator/rom/rom.c
@@ -96,23 +96,49 @@ static sigverify_otp_key_ctx_t sigverify_ctx;
 // A ram copy of the OTP word controlling how to handle flash ECC errors.
 uint32_t flash_ecc_exc_handler_en;
 
+static inline bool rom_console_enabled(void) {
+  return otp_read32(OTP_CTRL_PARAM_OWNER_SW_CFG_ROM_BANNER_EN_OFFSET) !=
+         kHardenedBoolFalse;
+}
+
+/**
+ * Prints a banner during bootup.
+ */
+static void rom_banner(void) {
+  if (!rom_console_enabled()) {
+    return;
+  }
+  //                          a t i T n e p O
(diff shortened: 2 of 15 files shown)


sw/device/silicon_creator/lib/drivers/uart.c
```diff
@@ -97,14 +97,21 @@ int uart_getchar(uint32_t timeout_ms) {
   return n ? (int)ch : -1;
 }
 
-size_t uart_write(const uint8_t *data, size_t len) {
-  size_t total = len;
+void uart_write(const void *data, size_t len) {
+  const uint8_t *d = (const uint8_t *)data;
```
moidx, line 101: we should add some of the countermeasures we discussed offline to avoid enabling an attacker to dump the flash by manipulating the stack via faults. 

`data` should only point to ro in ROM or ROM_EXT, and `len` should be limited to the limit of ro. 

We can add a `public_strings` region if you want to make this more portable.

sw/device/silicon_creator/lib/drivers/uart.c
```diff
@@ -97,14 +97,21 @@ int uart_getchar(uint32_t timeout_ms) {
...
-    data++;
+    uart_putchar(*d++);
     len--;
   }
-  return total;
+}
+
+void uart_write_hex(uint32_t val, size_t len) {
+  static const uint8_t kHexTable[16] = "0123456789abcdef";
+  size_t i = len * 8;
```
moidx, line 110: Here it would also be good to constraint the length up to word size.

sw/device/silicon_creator/rom/rom.c
```diff
@@ -113,23 +113,40 @@ static rom_error_t rom_irq_error(void) {
   return kErrorInterrupt + mcause;
 }
 
+static inline bool rom_console_enabled(void) {
+  return otp_read32(OTP_CTRL_PARAM_OWNER_SW_CFG_ROM_CONSOLE_EN_OFFSET) !=
+         kHardenedBoolFalse;
+}
```
a-will, line 119: This is specifically for the banner message only, right? A silicon owner might want the few milliseconds* of boot time back for the happy path, yet want to keep UART messages for faults--At first glance, it looks like that's what the code does.

(Maybe a naming nit, hehe)

*I'm not sure if I've added the pieces up correctly, but it looks like the banner is 26 characters of blocking uart_putchar() calls.

sw/device/silicon_creator/rom/rom.c
```diff
@@ -113,23 +113,40 @@ static rom_error_t rom_irq_error(void) {
...
+static void rom_banner(void) {
+  if (!rom_console_enabled()) {
+    return;
+  }
+  static const uint8_t title[10] = "OpenTitan:";
+  static const uint8_t git[4] = "GIT:";
+  lifecycle_hw_rev_t hw;
+  lifecycle_hw_rev_get(&hw);
+  uart_write(title, sizeof(title));
+  uart_write_hex(hw.silicon_creator_id, sizeof(hw.silicon_creator_id));
```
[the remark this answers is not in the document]
cfrantz: Done.  The banner now looks like:
```
OpenTitan:4001:0002:01
GIT:4f70656e54697461
```

sw/device/silicon_creator/lib/drivers/uart.c
cfrantz: Rather than putting section limits on `uart_write`, I went the route of encoding the string as little-endian bytes in a uint64 so the whole chunk could be passed by value rather than by reference.

This means editing the banner message is somewhat tedious, but we all can find an ASCII table when we need one.

sw/device/silicon_creator/lib/drivers/uart.c
cfrantz: Done, although this is low risk: no pointer is followed, so corrupting the length will just result in a bunch of extra `00`s being sent to the UART.

sw/device/silicon_creator/rom/rom.c
cfrantz: Yes - this is to control the printing of the banner message in the ROM (and we'll presumably obey the same OTP in the ROM_EXT).  Disabling it allows the owner to reclaim the time spent printing the banner message (currently 46 characters including the newlines (Looks like about 4ms at 115200bps).

I renamed the OTP from `ROM_CONSOLE_EN` to `ROM_BANNER_EN`, as this OTP does not affect printing of boot fault information.

a-will: Awesome, thanks!

sw/device/silicon_creator/lib/drivers/uart.c
moidx: ack. That makes sense. I have been using https://github.com/cfrantz/asciitable :)

moidx (approved):

sw/device/silicon_creator/rom/rom.c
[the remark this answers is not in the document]
cfrantz: BTW,  I changed this from colons to dashes:
```
OpenTitan:4001-0002-01
GIT:4f70656e54697461
```

The reason is perhaps a little bit silly, but it makes all of the output regular as "key:value" items.  FWIW, the outputs are directly parsable by the yaml parser or a permissive json parser.  I don't have a specific use-case for that, other than "lets make machine readability easy".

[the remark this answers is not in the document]
moidx: I think this is a good call. Thanks!

Issue #21965, opened 2024-03-12
[chip-test] Late debug enablement


msfschaffner:
### Test point name

chip_sw_rv_dm_late_debug

### Host side component

SystemVerilog+Rust

### OpenTitanTool infrastructure implemented

Yes

### Silicon Validation (SiVal)

Unkown

### Emulation Targets

- [ ] None
- [X] CW310
- [X] Hyperdebug + CW310

### Contact person

@msfschaffner @a-will

### Checklist

This test is for covering the RFC #20829 that is implemented in #21964.
We should test the entire truth table below in order to ensure everything plays nicely together at the top-level:

LC State | DIS_RV_DM_LATE_DEBUG (OTP Fuse) | LATE_DEBUG_ENABLE (CSR) | RV_DM reachable via TAP | RV_DM fully ungated
------------|------------|------------|------------|------------
TEST_UNLOCKED* or RMA | x | x | yes | yes
PROD*, RAW or invalid states | x | x | no | no
DEV | kMuBi8True | x | yes | yes
DEV | not kMuBi8True | not kMuBi32True | yes | no
DEV | not kMuBi8True | kMuBi32True | yes | yes

Please fill out this checklist as items are completed. Link to PRs and issues as appropriate.
- [ ] Check if existing test covers most or all of this testpoint (if so, either extend said test to cover all points, or skip the next 3 checkboxes)
- [ ] Device-side (C) component developed
- [ ] Bazel build rules developed
- [ ] Host-side component developed
- [ ] HJSON test plan updated with test name (so it shows up in the dashboard)
- [ ] Test added to dvsim nightly regression (and passing at time of checking)
- [ ] For SiVal test cases, test is running relevant FPGA or silicon regression

msfschaffner:
@rswarbrick FYI

andreaskurth:
@moidx: Worth testing with real ROM

andreaskurth:
@moidx: Do you think this will fit into your workload for M4, or should we reprioritize or reassign? I think this is a more complex feature than the one getting tested by #21990, so maybe this should be priority?

vogelpi:
@moidx: work on track

PR #23522, opened 2024-06-06
[dv/tlt] Add rv_dm late debug enable test.

moidx:
Update the `chip_rv_dm_lc_disabled` test point to include coverage for the `rv_dm` late debug enablement feature.

Modify the chip_rv_dm_lc_disabled_vseq.sv to include randomized tests for the late debug enablement feature as well as fixed coverage for the DEV life cycle state to exercise this feature.

Fixes: https://github.com/lowRISC/opentitan/issues/21965

### DIFF
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
@@ -22,6 +22,9 @@ class chip_rv_dm_lc_disabled_vseq extends chip_stub_cpu_base_vseq;
   jtag_dmi_reg_block jtag_dmi_ral;
   rand lc_state_e lc_state;
 
+  rand bit dis_late_debug_en_otp;
+  rand bit late_debug_en_reg;
+
   constraint num_trans_c {
     num_trans inside {[5:10]};
   }
@@ -62,6 +65,8 @@ class chip_rv_dm_lc_disabled_vseq extends chip_stub_cpu_base_vseq;
     // the strap sampling pulse is released a few cycles after the ROM check completes.
     cfg.clk_rst_vif.wait_clks(100);
 
+    late_debug_en_reg_update();
+
     `uvm_info(`gfn, "Attempt to activate RV_DM via JTAG.", UVM_MEDIUM)
     // RV_DM needs to be activated for the registers to work properly.
     // We always attempt to write this via the JTAG - even in states where the RV_DM is locked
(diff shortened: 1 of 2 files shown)


hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
```diff
@@ -70,21 +75,55 @@ class chip_rv_dm_lc_disabled_vseq extends chip_stub_cpu_base_vseq;
...
+
+    // All other permutations should result in rv_dm disabled.
+    return 0;
+  endfunction
+
   // Returns true if the RV_DM is accessible in a certain life cycle state.
   virtual function bit allow_rv_dm_access();
     if (lc_state inside {LcStTestUnlocked0, LcStTestUnlocked1, LcStTestUnlocked2,
         LcStTestUnlocked3, LcStTestUnlocked4, LcStTestUnlocked5, LcStTestUnlocked6,
         LcStTestUnlocked7, LcStRma, LcStDev}) begin
```
a-will, line 123: Shouldn't LcStDev be removed from this branch? It would appear that the next branch, with allow_late_rv_dm_access(), would never be used.

moidx: Fixed.

hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
```diff
@@ -142,22 +181,63 @@ class chip_rv_dm_lc_disabled_vseq extends chip_stub_cpu_base_vseq;
...
       end
+
       // Check RV_DM gating via the TL-UL bus.
       if ($urandom_range(0, 1)) begin
+        bit gated = ~allow_rv_dm_access();
+
+        // When late debug is enabled only rv_dm JTAG access is disabled.
+        // Gating via TL-UL is still disabled, and thus we need the following
+        // to match the expected hardware behavior.
+        if (lc_state == LcStDev) gated = 0;
```
a-will, line 216: Don't we have cases in the DEV state when TL-UL accesses should succeed? It seems like `allow_rv_dm_access()` should still do the right thing.

hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
```diff
@@ -142,22 +181,63 @@ class chip_rv_dm_lc_disabled_vseq extends chip_stub_cpu_base_vseq;
...
+                  "Disable late debug enable in otp: %0d, enable debug in rv_dm: %0d",
+                  dis_late_debug_en_otp, late_debug_en_reg), UVM_LOW);
+        `uvm_info(`gfn, $sformatf("rv_dm enable: %0d", allow_late_rv_dm_access()), UVM_LOW);
+      end
+
       // Check the RV_DM gating via JTAG. Since the serial JTAG interface wires are gated with the
       // lc_dft_en signal, it is sufficient to check an RW'able register to verify whether the
       // gating works or not.
       if ($urandom_range(0, 1)) begin
         rv_dm_jtag_access_check(~allow_rv_dm_access());
```
a-will, line 206: Since `allow_rv_dm_access()` now reports the TL-UL level gating for the DEV state, we should change this JTAG check to always succeed for the DEV state, right?

hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
moidx: The original logic in this test is inverted. I believe `gated = 0` means that all TL-UL accesses should succeed.

hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
moidx: Done. I updated the function names to make the code easier to follow, and now have `gate_rv_dm_dmi_access()` and `gate_rv_dm_mem_access()` to better handle the late debug enable expected behavior.

hw/top_earlgrey/dv/env/seq_lib/chip_rv_dm_lc_disabled_vseq.sv
moidx: I had the logic inverted in my head. This should be fixed now.

moidx:
@a-will I think this now works.

@rswarbrick, do you know if we have any coverage for SBA transactions when `DIS_RV_DM_LATE_DEBUG == false && late_debug_enable = false`?  If not, I can look at adding additional checks for this in a follow up PR. 

Thanks

a-will (approved):

Issue #20379, opened 2023-11-16
[tools/fusesoc] Realign with upstream FuseSoC


olofk:
### Description

I would like to start the discussion about getting OpenTitan to use upstream FuseSoC. There are several users who gets confused by the OpenTitan fork and I believe some of the new features of FuseSoC would benefit OpenTitan as well. It's been a while since I last looked at this, but would it be possible to list the specific features currently used by OpenTitan to see if we can start this process?

jwnrt:
I'm not an expert, but I think it's mainly the ability to depend on core files produced by generators. We use this for primitives generated by [`primgen`](https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/util/primgen.py).

olofk:
I would consider myself an expert, and you are right. The dynamic dependency in the OpenTitan fork is a bit problematic. There are different ways to solve this, but @rswarbrick and I discussed this offline a while ago and concluded that the best way forward would be to just add explicit dependencies in the affected prim packages and avoid a lot of complexity that way.

The other, smaller issue is some syntax fixes that have been fixed with https://github.com/lowRISC/opentitan/pull/21898

a-will:
This is just a random aside, but unless I'm missing developments, support for generated cores also isn't complete in the OpenTitan fork. The whole tree added by the generated core's own dependencies isn't added to the requirements, so we have a very awkward graph, where some catch-all prim core tries to do the job of adding dependencies that might be needed by a dependency of a generated core.

Generating cores and inserting their subtrees to the main tree are some fundamental features for OpenTitan's use case. fusesoc's build graph management is specialized and not as mature as bazel's, and it'd be nice if we could get a better integration between them. I'm not sure what form _that_ would take, but it'd be nice to have one source of truth / not have to manually write dependencies in two different places.

While I'm on the subject of bazel interaction, we'd also need to ensure newer fusesoc versions can operate such that they don't write directories outside what is provided in arguments (or under the current working directory). Our builds are always sandboxed for reproducibility, and more recent fusesoc versions are writing to `$XDG_CACHE_HOME`, which is forbidden. Can that all be removed with configuration? (With the library_root and cache_root keys, it seems like so, but golly does the config file format need documentation, hehe.)

olofk:
[...]

Issue #6604, opened 2021-05-18
[primgen/fusesoc] Generated primitive cores cannot depend on other generated primitive cores


msfschaffner:
Our current `primgen` mechanism can't reliably handle generated primitives that rely on other generated primitives.

The problem seems to arise due to the fact that `primgen` is currently only called for primitives that are discovered in the dependency graph BEFORE executing all generators.

Hence, if a core file of a technology-specific primitive implementation references another core file of a generated primitive not used anywhere else, that second primitive will not be properly generated with `primgen`, as the dependency only becomes evident AFTER the `primgen` generator pass.

Example:

- `lowrisc:prim:otp` is a primitive that relies on `primgen` and has the two tech-specific implementations `lowrisc:prim_generic:otp` and `lowrisc:prim_techXY:otp`.
- the latter implementation (`lowrisc:prim_techXY:otp`) depends on `lowrisc:prim:clock_gating`, where `lowrisc:prim:clock_gating` is also a primitive with multiple tech-specific implementations.
- `lowrisc:prim:clock_gating` is not used anywhere else in the DUT to be compiled.

In this case, fusesoc calls `primgen` only for the primitives directly used in the DUT, but not on `lowrisc:prim:clock_gating`, which leads to a missing abstract primitive wrapper for `lowrisc:prim:clock_gating` even tough the dependency has been specified in the `lowrisc:prim_techXY:otp` core.

A solution to avoid this problem is to define dependencies on other generated primitives in the abstract wrapper core file (see e.g.: #6606) in order to enforce generation of the corresponding abstract primitive wrappers.
Not sure there is a better solution, since it is a sort of "Chicken/Egg" problem that requires multiple iterations until convergence.

PS: I am not sure this is a duplicate. If an issue for this already exists, please link it here and close this one.

imphil:
That's a known limitation, tracked upstream at https://github.com/olofk/fusesoc/issues/404

It's non-trivial to solve properly and will likely need a rethinking of the way we generate primitives right now (if we don't want to go for the "iterate dependency resolution until stable" approach, which is slow and error-prone). 

It's on my list of things to look at, but for now, the workaround is simply to add the dependencies one level up, as you did in #6606.

msfschaffner:
Ok good - thanks for linking that upstream bug.
I just wanted to make sure we track this somewhere in our OT issues as well.

rswarbrick:
[...]

PR #23523, opened 2024-06-06
[hw/prim] Adapted primitives to no longer require `primgen.py`

HU90m:
This change pulls in the previously generated primitive wrappers, so that the `primgen.py` generator is no longer needed.

With these changes, OpenTitan can be built with upstream FuseSoC (tested on version 2.2).

I think there is some reorganisation of the primatives that could be nice; these are simply the minimal changes to remove `primgen.py`.

Fixes: https://github.com/lowRISC/opentitan/issues/20379 https://github.com/lowRISC/opentitan/issues/6604

*@rswarbrick This follows [the initial plan](https://github.com/lowRISC/opentitan/issues/20379#issuecomment-1984403219) without any additional changes needed.*


hw/ip/prim/rtl/prim_clock_div.sv
```diff
@@ -0,0 +1,51 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// This file is auto-generated.
```
jwnrt: What are they autogenerated by, primgen?

Will new primitive wrappers need writing by hand?

hw/ip/prim/README.md
```diff
@@ -201,29 +183,15 @@ targets:
...
-a FuseSoC generator. This generator is the script `util/primgen.py`. As input,
-the script receives a list of all cores found by FuseSoC anywhere in its search
-path. The script then looks through the cores FuseSoC discovered and extracts
-a list of technology libraries out of it. It then goes on to create the
-abstract primitive (copying over the list of parameters and ports from the
-generic implementation), and an associated core file, which depends on all
-technology-dependent libraries that were found.
+1. Choose a name for the new technology library. Names are all lower-case.
+   To ease sharing of technology libraries it is encouraged to pick a very specific name, e.g. `tsmc40lp`, and not `asic`.
+2. Copy the `prim_generic` folder into an arbitrary location (can be outside of this repository). Name the folder `prim_YOURLIBRARYNAME`.
```
jwnrt, line 188: Ah, I see the idea is to copy from an existing prim.

marnovandermaas (commented):
Huge fan of being able to build with upstream FuseSoC!

I'm on board with the first commit so we don't depend on primgen in the build process. For the second commit, I wonder whether we should keep primgen around so that there is an automated way to update all the primitives through the templates. What do you think?

jwnrt:
Apologies for my SystemVerilog ignorance, but does the "abstract" file for these prims contain any real logic, or is it solely selecting which tech-specific prim to use?

If the latter, would the following work?

* Change the tech-specific prim modules from `prim_(generic|xilinx|xilinx_ultrascale)_NAME` to `prim_NAME` (which would conflict).
* Have fusesoc select which `*.sv` file to use depending on the tool flow, and not include the others at all.

HU90m:
> I'm on board with the first commit so we don't depend on primgen in the build process. For the second commit, I wonder whether we should keep primgen around so that there is an automated way to update all the primitives through the templates. What do you think?

I think if primgen is no longer used for builds it will start to bitrot, which would cause future devs pain. For the use case you are suggesting, primgen would need to be adapted to generate files in there expected locations in the repository. It can always be revived from the history, if we wish to do this.

Going forward there is a lot of cruft that could be removed and re-organised.

For example [`prim_xilinx_ultrascale_xor2.sv`](https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_xor2.sv) and [`prim_xilinx_xor2.sv`](https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim_xilinx/rtl/prim_xilinx_xor2.sv) are duplicate modules. One could have one module that is used for both:

```systemverilog
if ((Impl == prim_pkg::ImplXilinx) || (Impl == prim_pkg::ImplXilinx_ultrascale)) begin : gen_xilinx
    prim_xilinx_xor2 #(
      .Width(Width)
    ) u_impl_xilinx (
      .*
    );
end else begin : gen_generic
    prim_generic_xor2 #(
      .Width(Width)
    ) u_impl_generic (
      .*
    );
end
```

In fact, the only difference between `prim_xilinx_xor2` and `prim_generic_xor2` is an attribute, so these could potentially all be collapsed into one module. There is likely some directory restructuring we could do to make it more ergonomic for techlibs to share.

These sorts of changes would require further, significant changes to primgen.

GregAC (changes requested):
Thanks for this @HU90m this looks like good work, however we must hold off on merging for now to avoid any possible disruption to M4.

So marking a 'Request changes' to prevent an accidental merge. I think this would be better off as a draft PR for now to make it clear the intent is early visibility and review.

HU90m:
> Apologies for my SystemVerilog ignorance, but does the "abstract" file for these prims contain any real logic, or is it solely selecting which tech-specific prim to use?
> 
> If the latter, would the following work?
> 
>     * Change the tech-specific prim modules from `prim_(generic|xilinx|xilinx_ultrascale)_NAME` to `prim_NAME` (which would conflict).
> 
>     * Have fusesoc select which `*.sv` file to use depending on the tool flow, and not include the others at all.

I had considered this and we could still do this.

A reason not to do this now is that we want to make the changes to the `prim`s to be reasonably small and incremental. This approach would require wider changes.

A reason not to do this ever is that the current approach is pure system verilog and easy to understand. Using FuseSoC would make it harder for people with less familiar with FuseSoC to understand what is going on and would make it harder for people injesting the RTL into a different build flow to swap at the technology library without re-injesting the RTL.

hw/ip/prim/README.md
```diff
@@ -201,29 +183,15 @@ targets:
...
+3. Replace the word `generic` everywhere with the name of your technology library. This includes
+    - file and directory names (e.g. `prim_generic_ram1p.sv` becomes
+      `prim_tsmc40lp_ram1p.sv`),
+    - module names (e.g. `prim_generic_ram1p` becomes `prim_tsmc40lp_ram1p`), and
+    - all other references (grep for it!).
+4. Implement all primitives. Replace the module body of the generic implementation with a technology-specific implementation as needed.
+   Do *not* modify the list of ports or parameters in any way!
+5. Add the new technology library to the enum in `prim_pkg.sv` with a value in the form `prim_pkg::ImplTechlibname`.
```
a-will, line 196: This is what I would object to here. primgen's purpose is to allow out-of-tree primitives _without modifying the upstream repo_.

hw/ip/prim/prim_and2.core
```diff
@@ -6,11 +6,16 @@ CAPI=2:
 name: "lowrisc:prim:and2"
 description: "Generic 2-input and"
 filesets:
-  primgen_dep:
+  files_rtl:
     depend:
       - lowrisc:prim:prim_pkg
-      - lowrisc:prim:primgen
       - lowrisc:prim:assert
+      - lowrisc:prim_xilinx:and2:0
```
a-will, line 13: Explicit dependencies here is going to make this not scale.

a-will (changes requested):
I think this misses the reason we have primgen. What is it about primgen that makes it incompatible with upstream fusesoc?

An alternative that might not break the model would be to turn the abstract cores that were generated into virtual cores. Each tech lib would implement their own providers, and the particular tech lib would be chosen at the top level. This idea might need massaging for dealing with the chip-level cores that already specify the tech lib to use, though maybe _that_ could be the output of a simple fusesoc generator, where a simple VLNV string provides the tech lib to depend on.

jwnrt:
> A reason not to do this ever is that the current approach is pure system verilog and easy to understand. Using FuseSoC would make it harder for people with less familiar with FuseSoC to understand what is going on and would make it harder for people injesting the RTL into a different build flow to swap at the technology library without re-injesting the RTL.

I get what you mean. Just for the record, this is what I did when porting prims to Bender, so it's not necessarily FuseSoC specific, it just needs an equivalent in whatever system you're using to build hardware.

a-will:
I haven't tested this beyond running fusesoc --setup on a few chip-level cores, but this might show what I was talking about with virtual cores: https://github.com/lowRISC/opentitan/pull/23555

This won't work without a very recent commit of fusesoc, though: The fixes to how virtual cores work aren't in a fusesoc release yet. On top of that, a more recent fusesoc appears to change the build directory compared to what dvsim currently expects, so that'll need to be fixed up.

HU90m:
Closing in favor of exploring a virtual core approach. See https://github.com/lowRISC/opentitan/pull/23555

PR #23524, opened 2024-06-06
[dv] Fix ROM e2e JTAG tests.

moidx:
1. Add `opentitan_binary` ROM targets with the expected suffix: i.e. `mask_rom_sim_dv`, `mask_rom_sim_verilator`.
2. Update Bazel cc.bzl to export `elf` files in `opentitan_binary` builds.
3. Update sim.mk to query Bazel output files from the `files` group if the target kind is not `opentitan_test`. This works because the ROMs are built using `opentitan_binary` targets.


rswarbrick (approved):
Looks sensible to me, but it probably makes sense to get a thumbs up from someone in the SW world too.

cfrantz (approved):

Issue #23144, opened 2024-05-16
[RFC] SLH-DSA Updates for Earl Grey


moidx:
### Description

This [document](https://docs.google.com/document/d/1QtgKamt56NY_WYtpdsRfYMh5sBEimtbEvkYhCpo8SvU/edit) proposes updates to the SLH-DSA (Sphincs+) implementation used in secure boot to provide configuration options that will simplify integration with offline and cloud based code signing infrastructure, while providing additional performance improvements in secure boot.

jadephilipoom:
Here's some context that might be helpful:
- SPHINCS+ supports parameter sets with either SHAKE or SHA-2 as the underlying hash function.
- The current implementation uses SHAKE, although the reasons for that are not super strong. They include:
  - for signing or keygen in the future we'd want to use SHAKE, since it has SCA protection
  - NIST was previously considering dropping the SHA-2 parameter sets, but has since decided to keep them, so this no longer really matters
  - the algorithm instantiation is a little more ergonomic for SHAKE than SHA2; while the SHAKE version just uses SHAKE for everything, the SHA2 version needs HMAC and MFG1 as well.
- Now, there are some reasons emerging why we may want to use SHA-2 instead, or at least make it possible:
  - signing infrastructure might support SHA-2 faster than it supports SHAKE
  - SHA-2 is slightly faster than SHAKE on OpenTitan because the implementation is not masked (10.5ms vs 12.5ms based on some quick ballpark benchmarks)

In addition to the SHAKE vs SHA-2 question, there's a separate but related question: should we allow a "pre-hashed" option (via an OTP setting or similar) in which the message we sign with SPHINCS+ is the _hash_ of the ROM_EXT image, rather than the image directly? If we did this and used SHA-2, it would reduce boot time for large images because ECDSA and SPHINCS+ could use the same digest (although arguably it would also have an effect on the fault attack difficulty -- see the doc for some discussion).

I'm a little ambivalent on these options myself to be honest; I think there aren't a lot of terrible choices here, and the balance to strike is keeping ROM code as simple as possible while also retaining flexibility as much as we can. Either way, though, if we want to make it for ROM freeze we need to decide on this one pretty quickly. It's a not huge but also not trivial amount of implementation effort.

johannheyszl:
removing Hotlist label since approved by TC

timothytrippel:
Can this be closed now?

jadephilipoom:
[...]

PR #23527, opened 2024-06-06
[rom] Change the endianness of the digest for the HMAC driver.

jadephilipoom:

Previously, the HMAC driver emitted the digest in "little-endian" mode (bytes reversed within words); this change flips it to "big-endian" mode. Normally, we represent everything in ROM in little-endian, and in general I'm a fan of that. However, in the case of HMAC/SHA2, the output is more like a byte sequence than an integer.

Before, we were essentially flipping bytes twice to cancel out big-endianness. Standards love big-endian, so both ECDSA and RSA interpret the byte sequence as a big-endian integer. When we only needed those, little-endian mode for SHA2 kind of made sense; we produce the output with flipped bytes and then just flip them again when we interpret the integer in the way OTBN and most other processors prefer (little-endian) as opposed to the awkward way FIPS wants them to interpret it (big-endian).

However, the second you want to use a digest to be anything other than an integer (for example, in SPHINCS+, when you want it to be part of the input to another hash function), the flipped bytes cause issues; there's no integer interpretation to cancel it out. Conceptually, it makes more sense to me to not flip the bytes in the first place and simply represent the bytes in the same order on the processor as the SHA2 specification dictates. This means for ECDSA, we have to flip the bytes when we write them to OTBN, but that seems preferable to me both logically and performance-wise than flipping them in SPHINCS+. The cryptolib already does this for ECDSA; I copied the `set_message_digest` routine from there.

I also had to change the `gen-otp-rot-auth-json.py` script to flip the bytes for the OTP auth digest. I expected to remove a reversal here, not add one, since `Crypto.Hash.SHA256` should produce output in the same order as the HMAC block does with this change. But it looks like the script flips bytes for every _other_ field (e.g. when computing the hash) but previously didn't for the digest; now it needs to do that for the digest too.

Also added back a `"//hw/top_earlgrey:fpga_cw310_rom_with_fake_keys"` env for `otbn_boot_data_functest`; I think this was (mistakenly?) removed in b217f41a378e30cdef6685c1d26fb3e990db0b4e but I don't see why it shouldn't be running and would like to have the coverage. It passes in seconds.

### DIFF
--- a/sw/device/silicon_creator/lib/BUILD
+++ b/sw/device/silicon_creator/lib/BUILD
@@ -493,6 +493,7 @@ opentitan_test(
     srcs = ["otbn_boot_services_functest.c"],
     exec_env = {
         "//hw/top_earlgrey:fpga_cw310_sival": None,
+        "//hw/top_earlgrey:fpga_cw310_rom_with_fake_keys": None,
         "//hw/top_earlgrey:sim_dv": None,
         "//hw/top_earlgrey:sim_verilator": None,
     },
--- a/sw/device/silicon_creator/lib/drivers/hmac.c
+++ b/sw/device/silicon_creator/lib/drivers/hmac.c
@@ -23,12 +23,12 @@ void hmac_sha256_init(void) {
   abs_mmio_write32(TOP_EARLGREY_HMAC_BASE_ADDR + HMAC_INTR_STATE_REG_OFFSET,
                    UINT32_MAX);
 
+  // Configure the HMAC block to run SHA2-256 with a 256-bit key.
   uint32_t reg = 0;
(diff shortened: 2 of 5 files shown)


jadephilipoom:
After discussion with @cfrantz , closing this in favor of keeping the HMAC block in little-endian mode. While this is a little nicer conceptually, a lot of existing infrastructure relies on the little-endian mode and it's too last-minute to change it. The performance impact on SPHINCS+ can be managed by creating an alternate `hmac_sha256_final` function that reads the output in big-endian order with `__builtin_bswap32`. If the performance is too affected by the `bswap`, we can also add an argument to the `init` function to set the block configuration option, but let's see what the benchmarks say first.

PR #23528, opened 2024-06-06
[i2c,dv] I2C scoreboard rearchitect

hcallahan-lowrisc:
This PR cleans up the architecture of the I2C scoreboard to create a clean break between comparison (scoreboard) and modelling (reference model / predictor). This change is a precursor to further DV, which builds upon the new infrastructure to model more behaviour in the refModel and less in the vseqs. The I2C DV environment in it's current form is difficult to work on, and this change should go a long way towards improving it.

This is a big PR, and is therefore difficult to judge based on the diff only. I have regressed it locally, and only seen a difference to the "i2c_target_hrst_vseq" test, which is currently passing approx 30% of the time. This vseq is quite unpleasant in the way it interacts with the agent/scoreboard, and is very brittle to change, while mainly targeting an increase in FSM transition coverage via short-circuiting normal I2C transactions with early RSTART/STOP conditions. I have made some changes to the test to begin to clean it up, but it remains partially broken, and I propose to leave it this way and to come back and cleanup later.

Definitely best review commit-by-commit, with the final commit containing the most substantial changes, and the earlier 3 being more cleanup that functional change.

Commit message from final scoreboard refactor commit:
> The current state of the I2C DV environment is not great. Very little of the
> hardware is modelled, sequence items are used to represent vastly different
> things in different contexts, and checking in the scoreboard is deliberately
> limited to a subset of fields we believe to have modelled correctly. The vseq's
> frequently intertwine stimulus and checking, and often reach into other parts of
> the DV environment (e.g. agent's monitor, scoreboard) to reset state when our
> modelling is insufficient. This is not great.
> 
> This change begins to setup a new, principled foundation for DV by tackling the
> scoreboard. Currently, as often is done in OpenTitan DV, the scoreboard performs
> two functions, both the intended checking component, but also acts as a
> reference model to try and predict future DUT behaviour. This leads to a very
> dense and overloaded module, and makes it difficult to modify while maintaining
> a clean seperation.
> This commit splits the modelling components of the existing scoreboard into a
> seperate component "i2c_reference_model" (aka predictor), leaving the scoreboard
> as a more bare-bones comparator. The scoreboard maintains a handle to the
> reference model, and is responsible for calling "process_tl_access()" based on
> observed TL-UL bus activity. Ideally the reference model would communicate with
> the scoreboard only via TLM ports/fifo, but this is not yet implemented.
> 
> This change also routes stimulus seq_items, generated by the vseq's to drive the
> agent (particularly for DUT-Target tests), first into the reference model and
> then onto the scoreboard. In some cases the items are passed-on unchanged to the
> scoreboard, but sometimes we make modifications. This architecture is maybe not
> ideal, but does clean up what we have currently, and could be iterated on.m

### Old regression results (hash) : https://reports.opentitan.org/hw/ip/i2c/dv/2024.06.03_06.49.14/report.html
image

image

---

### New regression results (hash)

I2C Simulation Results
Thursday June 06 2024 09:45:36 UTC
GitHub Revision: [25ebe35d7a](https://github.com/lowrisc/opentitan/tree/25ebe35d7aceb7a065fb61fad89023cd052b4c48)
Branch: HEAD
[Testplan](https://opentitan.org/book/hw/ip/i2c/data/i2c_testplan.html)
Simulator: XCELIUM
image

---

### DIFF
--- a/hw/dv/sv/i2c_agent/i2c_item.sv
+++ b/hw/dv/sv/i2c_agent/i2c_item.sv
@@ -14,6 +14,7 @@ class i2c_item extends uvm_sequence_item;
   bit                      addr_ack;
   bit                      data_ack_q[$];
   // transaction control part
+  i2c_acq_byte_id_e        signal; // ACQDATA.SIGNAL
   bit                      nack;
   bit                      ack;
   bit                      rstart;
@@ -51,27 +52,28 @@ class i2c_item extends uvm_sequence_item;
   }
 
   `uvm_object_utils_begin(i2c_item)
-    `uvm_field_int(tran_id,                 UVM_DEFAULT)
-    `uvm_field_enum(bus_op_e, bus_op,       UVM_DEFAULT)
-    `uvm_field_int(addr,                    UVM_DEFAULT)
-    `uvm_field_int(num_data,                UVM_DEFAULT)
-    `uvm_field_int(start,                   UVM_DEFAULT)
-    `uvm_field_int(stop,                    UVM_DEFAULT)
-    `uvm_field_int(wdata,                   UVM_DEFAULT | UVM_NOCOMPARE)
-    `uvm_field_queue_int(data_q,            UVM_DEFAULT | UVM_NOPRINT)
-    `uvm_field_queue_int(fmt_ovf_data_q,    UVM_DEFAULT | UVM_NOCOMPARE)
-    `uvm_field_int(rdata,                   UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
-    `uvm_field_int(rstart,                  UVM_DEFAULT | UVM_NOCOMPARE)
-    `uvm_field_int(fbyte,                   UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
-    `uvm_field_int(ack,                     UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
-    `uvm_field_int(nack,                    UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
-    `uvm_field_int(read,                    UVM_DEFAULT | UVM_NOCOMPARE)
-    `uvm_field_int(rcont,                   UVM_DEFAULT | UVM_NOCOMPARE)
-    `uvm_field_int(nakok,                   UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
-    `uvm_field_enum(drv_type_e,  drv_type,  UVM_DEFAULT | UVM_NOCOMPARE)
-    `uvm_field_int(wait_cycles,             UVM_DEFAULT | UVM_NOCOMPARE)
-    `uvm_field_int(addr_ack,                UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
-    `uvm_field_queue_int(data_ack_q,        UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
+    `uvm_field_int(tran_id,                    UVM_DEFAULT)
+    `uvm_field_enum(bus_op_e, bus_op,          UVM_DEFAULT)
+    `uvm_field_int(addr,                       UVM_DEFAULT)
+    `uvm_field_int(num_data,                   UVM_DEFAULT)
+    `uvm_field_int(start,                      UVM_DEFAULT)
+    `uvm_field_int(stop,                       UVM_DEFAULT)
+    `uvm_field_int(wdata,                      UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_queue_int(data_q,               UVM_DEFAULT)
+    `uvm_field_queue_int(fmt_ovf_data_q,       UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_int(rdata,                      UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
+    `uvm_field_enum(i2c_acq_byte_id_e, signal, UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_int(rstart,                     UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_int(fbyte,                      UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
+    `uvm_field_int(ack,                        UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
+    `uvm_field_int(nack,                       UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
+    `uvm_field_int(read,                       UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_int(rcont,                      UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_int(nakok,                      UVM_DEFAULT | UVM_NOPRINT | UVM_NOCOMPARE)
+    `uvm_field_enum(drv_type_e,  drv_type,     UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_int(wait_cycles,                UVM_DEFAULT | UVM_NOCOMPARE)
+    `uvm_field_int(addr_ack,                   UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
+    `uvm_field_queue_int(data_ack_q,           UVM_DEFAULT | UVM_NOCOMPARE | UVM_NOPRINT)
   `uvm_object_utils_end
 
   `uvm_object_new
--- a/hw/dv/sv/i2c_agent/i2c_monitor.sv
+++ b/hw/dv/sv/i2c_agent/i2c_monitor.sv
@@ -95,7 +95,7 @@ class i2c_monitor extends dv_base_monitor #(
     end else begin
       mon_dut_item.rstart = 1'b1;
     end
-    num_dut_tran++;
+    mon_dut_item.tran_id = num_dut_tran++;
     mon_dut_item.start = 1'b1;
 
     target_address_thread();
@@ -120,7 +120,6 @@ class i2c_monitor extends dv_base_monitor #(
     bit rw_req = 1'b0;
 
     // sample address and r/w bit
-    mon_dut_item.tran_id = num_dut_tran;
     for (int i = cfg.target_addr_mode - 1; i >= 0; i--) begin
       cfg.vif.get_bit_data("host", cfg.timing_cfg, mon_dut_item.addr[i]);
       `uvm_info(`gfn, $sformatf("target_address_thread() address[%0d] %b",
@@ -255,22 +254,45 @@ class i2c_monitor extends dv_base_monitor #(
     end
   endtask : monitor_ready_to_end
 
-  // Handle agent reset and set stop bit to indicate completion of the current transaction
-  task handle_monitor_rst(input string task_name);
+  // Handle an externally-triggered monitor reset
+  // This routine is currently triggered externally by setting cfg.monitor_rst = 1 in the
+  // hrst_vseq. This is a hack to try and keep the dv env in sync with stimulus that
+  // short-circuits the current transaction with an early RSTART/STOP. It should be refactored
+  // and removed, preferably with proper modelling inside the scoreboard/refmodel.
+  //
+  // - Wait for the next edge on the bus
+  // - Clear temporary state about the currently monitored transaction
+  // - Set stop bit to indicate completion of the current transaction
+  task handle_rst(input string task_name);
     int wait_timeout_ns = 1_000_000; // 1 ms
-    if (cfg.monitor_rst) begin
-      @(cfg.vif.cb);
-      mon_dut_item.clear_all();
-      `DV_WAIT((!cfg.monitor_rst), , wait_timeout_ns, $sformatf("%0s:monitor reset de-assert",
-               task_name));
-      cfg.got_stop = 1;
-      `uvm_info(`gfn, $sformatf("monitor forceout from %0s", task_name), UVM_MEDIUM)
-    end
+    `uvm_info(`gfn, $sformatf("handle_rst() from task '%0s'", task_name), UVM_MEDIUM)
+
+    // Wait for the next bus activity via the clocking block
+    @(cfg.vif.cb);
+
+    // Clear the temporary item used to accumulate in-progress transactions
+    mon_dut_item.clear_all();
+
+    // Wait for cfg.monitor_rst to de-assert (within 1ms).
+    `DV_WAIT(// WAIT_COND_
+             (!cfg.monitor_rst),
+             // MSG_
+             ,
+             // TIMEOUT_NS_
+             wait_timeout_ns,
+             // ID_
+             $sformatf("handle_rst: reset failed to de-asserted from task '%0s'", task_name));
+
+    // Indicate the end of the current transaction with a stop condition.
+    cfg.got_stop = 1;
   endtask
 
   virtual protected task controller_collect_thread();
     i2c_item full_item;
     bit skip_the_loop = 0;
+
+    // Wait for the vseq to clear this at the start of the next stimulus round.
+    wait(cfg.got_stop == 0);
 
     cfg.valid_addr = 0;
     cfg.vif.drv_phase = DrvIdle;
@@ -281,7 +303,7 @@ class i2c_monitor extends dv_base_monitor #(
     end else begin
       mon_dut_item.rstart = 1'b1;
     end
-    mon_dut_item.tran_id = num_dut_tran;
+    mon_dut_item.tran_id = num_dut_tran++;
     mon_dut_item.start = 1'b1;
 
     controller_address_thread(skip_the_loop);
@@ -300,7 +322,6 @@ class i2c_monitor extends dv_base_monitor #(
       full_item.read = 1;
       analysis_port.write(full_item);
     end
-    num_dut_tran++;
     mon_dut_item.clear_data();
   endtask: controller_collect_thread
 
@@ -347,10 +368,11 @@ class i2c_monitor extends dv_base_monitor #(
              `DV_CHECK_CASE_EQ(r_bit, 1'b0)
            end
          end
-         begin // agent reset thread
-           bit rst_detected = 0;
-           wait(cfg.monitor_rst);
-           handle_monitor_rst("controller_address_thread()");
+         begin
+           begin
+             wait(cfg.monitor_rst);
+             handle_rst("controller_address_thread()");
+           end
            do_skip = 1; // Skip processing rest of the transaction
          end
        join_any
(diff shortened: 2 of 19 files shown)


hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -290,27 +321,31 @@ class i2c_monitor extends dv_base_monitor #(
            for (int i = cfg.target_addr_mode - 1; i >= 0; i--) begin
              cfg.vif.p_edge_scl();
              mon_dut_item.addr[i] = cfg.vif.cb.sda_i;
-             `uvm_info(`gfn, $sformatf("\nmonitor, address[%0d] %b", i, mon_dut_item.addr[i]),
-                       UVM_HIGH)
+             `uvm_info(`gfn, $sformatf("controller_address_thread() address[%0d] %b",
+               i, mon_dut_item.addr[i]), UVM_HIGH)
```
rswarbrick, line 341: Nit: indentation, and same on next uvm_info call, and again with the allow_bad_addr message below.

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -382,35 +418,43 @@ class i2c_monitor extends dv_base_monitor #(
...
-            cfg.vif.p_edge_scl();
+
+          begin
+            bit [6:0] data;
+            for (int i = 7; i >= 0; i--) begin
+              cfg.vif.p_edge_scl();
+              data[i] = cfg.vif.sda_i;
+            end
+            `uvm_info(`gfn, $sformatf("controller_write_thread() got data 0x%2x",
+              data), UVM_MEDIUM)
```
rswarbrick, line 465: Nit: indentation

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -382,35 +418,43 @@ class i2c_monitor extends dv_base_monitor #(
...
-          // check for ack
+
+          `uvm_info(`gfn, "controller_write_thread() waiting for A/N", UVM_MEDIUM)
           cfg.vif.p_edge_scl();
           r_bit = cfg.vif.cb.sda_i;
-          `uvm_info(`gfn, $sformatf("\nmonitor, target_write detect HOST %s",
-                                   (!r_bit) ? "ACK" : "NO_ACK"), UVM_MEDIUM)
-          // if nack is issued, next bit must be stop or rstart
+          `uvm_info(`gfn, $sformatf("controller_write_thread() saw %0s",
+            (!r_bit) ? "ACK" : "NACK"), UVM_MEDIUM)
```
rswarbrick, line 472: Nit: indentation

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -382,35 +418,43 @@ class i2c_monitor extends dv_base_monitor #(
...
     fork begin
       fork
         while (!mon_dut_item.stop && !mon_dut_item.rstart) begin
+
           mon_dut_item.drv_type = WrData;
-          for (int i = 7; i >= 0; i--) begin
-            cfg.vif.p_edge_scl();
+
+          begin
+            bit [6:0] data;
```
rswarbrick, line 459: What are we doing with the value that we're collecting here? All I can see is an assignment to a local variable. (Maybe that is for a future PR?)

hw/ip/i2c/dv/env/i2c_env_pkg.sv
```diff
@@ -78,22 +79,34 @@ package i2c_env_pkg;
   parameter uint NUM_ALERTS = i2c_reg_pkg::NumAlerts;
   parameter string LIST_OF_ALERTS[] = {"fatal_fault"};
 
-  function automatic i2c_item acq2item(bit[9:0] data);
+  function automatic i2c_item acq2item(bit[10:0] data);
```
rswarbrick, line 82: I'd consider expanding this out to something big and then adding an assertion that none of the higher bits are nonzero. The two callsites are passing a `uvm_reg_data_t` and a `bit [bus_params_pkg::BUS_DW-1:0]`.

rswarbrick (commented):
Some nitty comments about everything but the last commit (which I'm going to review properly now)

hw/dv/sv/i2c_agent/i2c_item.sv
```diff
@@ -14,6 +14,7 @@ class i2c_item extends uvm_sequence_item;
   bit                      addr_ack;
   bit                      data_ack_q[$];
   // transaction control part
+  i2c_acq_byte_id_e        signal;
```
antmarzam, line 17: if we wanted to compare the field or print it out, Should this field not be added to the `uvm_field*` macros?

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -1,7 +1,34 @@
...
+//
+// Read and Write transactions are collected into different ports, and also
+// different ports are used depending on if the Agent itself is configured as
+// either a Controller or Target.
+// Agent-Target
+// W -> 'wr_item_port'
+// R -> 'rd_item_port'
+// Agent-Controller
+// W ->
+// R -> 'analysis_port'
```
antmarzam, line 15: Is the default 'analysis_port' via inheriting from `uvm_subscriber` and used for the 'Agent-controller' variant not enough to reuse for the 'Agent-target'  R analysis port?

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -35,120 +62,86 @@ class i2c_monitor extends dv_base_monitor #(
   endtask : wait_for_reset_and_drop_item
 
   virtual task run_phase(uvm_phase phase);
+    wait(cfg.en_monitor);
     wait(cfg.vif.rst_ni);
     if (cfg.if_mode == Host) begin
```
antmarzam, line 67: Can the agent config change on the fly, or would we either have `cfg.if_mode == Host` or `cfg.if_mode == Device`?

If the config can't change I suggest having a child i2c_monitor_host, and i2c_monitor_device, then the run phase can have common behaviour and any host/device functionality could go in the child class run_phase?

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -0,0 +1,578 @@
...
+        // If stop was set, that concludes the in-progress write
+        // transfer. Push it to a queue for checking.
+        exp_wr_item.stop = stop;
+        if (stop) begin
+          push_complete_write_txn(exp_wr_item);
+          exp_wr_item = new("exp_wr_item");
+        end
+      end
+
+      // There is an in-progress read transfer
```
rswarbrick, line 382: Not a big deal, but maybe it's worth making this match with the `if` block above. If you're going for the "statement of fact" wording, I think it needs to be inside the `if` check.

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -0,0 +1,578 @@
...
+
+        end
+
+      end
+    end
+
+  endtask: fmtfifo_write
+
+  //
+  //
```
rswarbrick, line 429: I'd probably either fill this comment out or drop it. Similarly above `rdata_read`.

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -35,120 +62,86 @@ class i2c_monitor extends dv_base_monitor #(
...
   endtask : run_phase
 
-  // Collect transactions forever
-  virtual protected task collect_thread(uvm_phase phase);
+  virtual protected task target_collect_thread();
     i2c_item full_item;
-    wait(cfg.en_monitor);
     if (mon_dut_item.stop ||
        (!mon_dut_item.stop && !mon_dut_item.start && !mon_dut_item.rstart)) begin
       cfg.vif.wait_for_host_start(cfg.timing_cfg);
```
antmarzam, line 93: Not code you've touched now but figured it'd be worth asking.

Is there a reason why some of the code has some delays based on clocking blocks, but not all? I feel all the code should be referencing clocking blocks if we're using them.

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -35,120 +62,86 @@ class i2c_monitor extends dv_base_monitor #(
...
-            disable fork;
-          end: iso_fork
-        join
+        fork begin: iso_fork
+          fork
+            target_collect_thread();
+            begin // if (on-the-fly) reset is monitored, drop the item
+              wait_for_reset_and_drop_item();
+              `uvm_info(`gfn, $sformatf("Monitor is reset, dropping item now:\n%s",
+                                        mon_dut_item.sprint()), UVM_DEBUG)
```
antmarzam, line 80: Does it make sense to add this printout to the task above?

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -265,20 +254,78 @@ class i2c_monitor extends dv_base_monitor #(
     end
   endtask : monitor_ready_to_end
```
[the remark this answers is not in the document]
antmarzam: I'd personally suggest to reference all interface signals via the clocking block, including in the interface tasks

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -346,25 +398,26 @@ class i2c_monitor extends dv_base_monitor #(
...
+        // mon_rstart, when ack_stop_mon() task detects ack / stop event
         while (!mon_dut_item.stop && !mon_dut_item.rstart) begin
           // ask driver response read data
           mon_dut_item.drv_type = RdData;
           for (int i = 7; i >= 0; i--) begin
             cfg.vif.sample_target_data(cfg.timing_cfg, mon_data[i]);
-            `uvm_info(`gfn, $sformatf("\nmonitor, target_read, trans %0d, byte %0d, bit[%0d] %0b",
-                      mon_dut_item.tran_id, mon_dut_item.num_data+1, i, mon_data[i]), UVM_HIGH)
+            `uvm_info(`gfn,
+              $sformatf("controller_read_thread(), trans %0d, byte %0d, bit[%0d] %0b",
```
antmarzam, line 408: I wonder if you'd be happy to get rid of the "controller_read_thread()" printout via using %m instead?
Perhaps that printout won't be as neat as the one you're expecting. See table 21.2 in the LRM if you're interested, because I believe you may get "class_name.parent_task.child_task" type of printout

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: I think this is fine according to my reading of the [style guide section](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#line-wrapping)? (with the small caveat of `s/four/two/` for indentation depth as per opentitan style generally)

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: It was just for printing in the following uvm_info

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: We don't change on the fly, no. This code could definitely have a larger cleanup effort to reduce duplication, as you say. I'm going to leave that for a future task though.

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: This is cool, I did not know about this feature! Thanks. I'm going to leave the shorter name, just because using %m here results in very long log lines with our default printer configuration. But it's very useful to know about!
e.g.
```
UVM_INFO @  35049268 ps: (i2c_monitor.sv:325) uvm_test_top.env.m_i2c_agent.monitor [uvm_test_top.env.m_i2c_agent.monitor] i2c_agent_pkg::i2c_monitor.controller_address_thread.unmblk1 address[6] 1
```

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: Yes it's not ideal, but this is the existing architecture so I'm going to leave it alone for now. This block uses seq_items in different and inconsistent ways, and could definitely be simplified in the future.

hw/dv/sv/i2c_agent/i2c_item.sv
hcallahan-lowrisc: Indeed, added!

hw/ip/i2c/dv/env/i2c_env.sv
```diff
@@ -12,6 +12,8 @@ class i2c_env extends cip_base_env #(
 
   i2c_agent m_i2c_agent;
 
+  i2c_reference_model model;
```
martin-velay, line 15: Info for my understanding: is there a good reason behind using the words "reference model" instead of "predictor"? As that one is probably more commonly used in DV according to my own experience at least.

hw/dv/sv/i2c_agent/i2c_monitor.sv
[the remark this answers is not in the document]
hcallahan-lowrisc: Yeah good point, I will leave that as a follow up to this PR since it's not code I've touched here. I wonder if there is a way to setup a lint for that, as it would be a good default to take IMO!

hw/ip/i2c/dv/env/i2c_env.sv
hcallahan-lowrisc: No good reason, other than dv code in OTBN already uses the term model for this. I'm interested if predictor is a more familiar term in-general for DV engineers, as I'm not experienced enough to make a conclusion about that!

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -0,0 +1,578 @@
...
+    target_mode_wr_port = new("target_mode_wr_port", this);
+    target_mode_rd_port = new("target_mode_rd_port", this);
+  endfunction: build_phase
+
+  task run_phase(uvm_phase phase);
+    super.run_phase(phase);
+    fork
+      forever create_wr_exp_from_stim();
+      forever create_rd_exp_from_stim();
+    join_none
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Most scoreboards in OT (also derived from uvm_component) also do this, so I think it's fine for other static components in the hierarchy to do this as well.

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -0,0 +1,578 @@
...
+    // For any registers where hw can modify the register's value, and our refmodel / scoreboard
+    // are unable to predict this new value before the read takes place, the code below should
+    // ensure this bit is cleared so that no check takes place.
+    bit do_read_check = 1'b1;
+
+    bit write = item.is_write();
+    bit tl_get           = (!write && channel == AddrChannel);
+    bit tl_putdata       =  (write && channel == AddrChannel); // write
+    bit tl_accessackdata = (!write && channel == DataChannel); // read
+    bit tl_accessack     =  (write && channel == DataChannel);
```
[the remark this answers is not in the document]
hcallahan-lowrisc: These message names just came straight out of the TL spec, for the relevant TL-UL messages we care about (other than 'putdata', which covers both 'PutFullData' and 'PutPartialData')

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -0,0 +1,578 @@
...
+    end
+
+  endtask : process_tl_access
+
+
+  // Update our model and predictions based on the write event.
+  //
+  task update_on_write(bit [bus_params_pkg::BUS_DW-1:0] data, uvm_reg csr);
+    case (csr.get_name())
+      "ctrl":;
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Removed the empty cases. This method has been passed a csr handle which was ultimately sanitised in the scoreboard's process_tl_access() method, which checked the address was inside the expected range and the lookup did not return a null object, so I think this should be safe enough.

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -0,0 +1,578 @@
...
+      default:;
+    endcase
+  endtask: update_on_read
+
+  // Task to accumulate expected items based on writes to the FMTFIFO
+  //
+  task fmtfifo_write(bit [bus_params_pkg::BUS_DW-1:0] data);
+
+    bit [7:0] fbyte = get_field_val(ral.fdata.fbyte, data);
+    bit start = bit'(get_field_val(ral.fdata.start, data));
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Dropped the explicit casts (and a few other places too).

hw/ip/i2c/dv/env/i2c_reference_model.sv
hcallahan-lowrisc: Yeah good point, moved comments inside the appropriate blocks.

hw/ip/i2c/dv/env/i2c_reference_model.sv
hcallahan-lowrisc: Filled out both comments

rswarbrick (approved):
Thanks for responding to the nitty stuff. I'd probably suggest merging this mammoth PR when possible. Some of the other comments might be worth discussing, but maybe the best approach is to do that as a follow-up.

martin-velay (approved):
I have reviewed this PR, I think it could be merged as nothing is blocking. But I am expecting some answers to my comments and maybe some adjustments if judged as necessary

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -0,0 +1,578 @@
...
+    endcase
+
+    // Push the item to the scoreboard as our rd_exp item.
+    target_mode_rd_port.write(exp_rd);
+  endtask: create_rd_exp_from_stim
+
+
+  // Reset local fifos, queues and variables
+  virtual function void reset();
+    rd_pending_q = {};
```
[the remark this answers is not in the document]
hcallahan-lowrisc: No particular reason here, as I understand it the two are equivalent. We seem to use both throughout the codebase

hw/ip/i2c/dv/env/i2c_scoreboard.sv
```diff
@@ -9,783 +9,685 @@ class i2c_scoreboard extends cip_base_scoreboard #(
...
+            cov.intr_test_cg.sample(intr, data[i], intr_en[i], intr_exp[i]);
           end
         end
-        "timing0": begin
-          if (cfg.en_cov) begin
-            cov.thigh_cg.sample(`gmv(ral.timing0.thigh));
-            cov.tlow_cg.sample(`gmv(ral.timing0.thigh));
-          end
+        if (cfg.en_cov) begin
+          cov.interrupts_cg.sample(.intr_state(ral.intr_state.get_mirrored_value()),
```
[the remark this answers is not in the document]
hcallahan-lowrisc: I've standardized on the `gmv macro in this file now. Thanks

hcallahan-lowrisc:
Quite a few of the review comments are more 'meta' than specific to these particular changes, so I'm going to address as a follow up. Thanks for all the detailed reviews!

PR #23532, opened 2024-06-06
[sival, rv_core_ibex] Add junk bits to address translation test

GregAC:
The lower bits in the remap_addr field should be ignored. This adds some junk bits to the remap_addr field to ensure they don't effect translation.


sw/device/tests/rv_core_ibex_address_translation_test.c
```diff
@@ -58,15 +58,19 @@ const dif_rv_core_ibex_addr_translation_slot_t kSlots[] = {
...
 
 // Translation descriptions to use.
+// Junk bits (0xDEADBEEF) are added to the `remap_addr` fields to ensure they
+// are ignored by the translation mechanism.
 static const dif_rv_core_ibex_addr_translation_mapping_t kMakeLowerCaseMapping =
     {
         .matching_addr = (uintptr_t)remapped_function,
-        .remap_addr = (uintptr_t)make_lower_case,
+        .remap_addr = (uintptr_t)make_lower_case +
+                      (uintptr_t)(0xDEADBEEF & (kRemapAlignment - 1)),
```
nbdd0121, line 67: I wonder if we should test both with and without junk bits?

We can drop the `const` on these variables, do one test without junk bits, add junk bits of the mapping in `test_main` and repeat the test.

GregAC: I'm not sure that really adds anything to the test

nbdd0121 (approved):

sw/device/tests/rv_core_ibex_address_translation_test.c
GregAC: @nbdd0121 happy to approve without this change?

sw/device/tests/rv_core_ibex_address_translation_test.c
```diff
@@ -58,15 +58,19 @@ const dif_rv_core_ibex_addr_translation_slot_t kSlots[] = {
...
         .size = kRemapAlignment,
 };
 static const dif_rv_core_ibex_addr_translation_mapping_t kGetNameMapping = {
     .matching_addr = (uintptr_t)remapped_function,
-    .remap_addr = (uintptr_t)get_name,
+    .remap_addr =
+        (uintptr_t)get_name + (uintptr_t)(0xDEADBEEF & (kRemapAlignment - 1)),
```
marnovandermaas, line 73: Wouldn't it be better to randomize the lower bits instead of making them constant ones?

GregAC: Maybe though in general the chip-level tests haven't been using randomization so it seems reasonable to not randomize here as well.

marnovandermaas (approved):
Approving this based on a code review only.

sw/device/tests/rv_core_ibex_address_translation_test.c
marnovandermaas: Ok, I'll leave this up to you. I've approved it in the meantime to unblock

PR #23533, opened 2024-06-06
Fix `chip_sw_rom_access` / `openocd_debug_test_otp_*` LC states

luismarques:
We don't have JTAG access in `PROD`. Make the lifecycle states for these tests properly reflect that, instead of "passing when it fails" as was done before for `PROD` and `PROD_END` states.


jwnrt (approved):
Thanks!

Issue #23530, opened 2024-06-06
[hmac,rtl] Save & Restore FSM bug


gdessouky:
I am currently releasing some constraints after having in place the S&R with different contexts (not complete yet, still remain to add some additional checks). This is leading me into a potential issue where the HMAC Done is never triggered. I am currently trying to understand if it's truly an RTL bug or not.

To reproduce it with Xcelium:
```
git checkout b242d6c90327aed83a1f67fbf1518bf4734741d8
util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_smoke -t xcelium --fixed-seed 114185448211150958355807046043715683092785711447270148054926510481266628632264
```
With VCS, use seed: `83812634503382603449251114596099188419278643638756659410240524018062663081103`

@gdessouky please could you have a look?

_Originally posted by @martin-velay in https://github.com/lowRISC/opentitan/issues/23116#issuecomment-2150343359_

PR #23538, opened 2024-06-06
[hmac,rtl] Fix Save & Restore HMAC FSM bug

gdessouky:
This addresses the corner case where `stop` is triggered and block size is reached but there are data words buffered in the MSG FIFO that are equal to block size that still need to be push out. This should close https://github.com/lowRISC/opentitan/issues/23530.

cc @martin-velay


martin-velay (approved):

gdessouky:
The failing test is unrelated.

moidx (approved):

hw/ip/hmac/rtl/hmac_core.sv
```diff
@@ -367,10 +367,15 @@ module hmac_core import prim_sha2_pkg::*; (
...
           st_d    = StWaitResp;
-          hmac_sha_rvalid = 1'b0; // block
-          hash_process = (round_q == Outer);
-        end else if (txcnt_eq_blksz && reg_hash_stop_q && (round_q == Inner)) begin
-          st_d =  StWaitResp; // to wait on sha_hash_done_i
+
+          hmac_sha_rvalid = 1'b0; // block reading words from MSG FIFO
+          hash_process    = (round_q == Outer);
+        end else if (txcnt_eq_blksz && (txcount >= sha_message_length_o)
+                     && reg_hash_stop_q && (round_q == Inner)) begin
```
johannheyszl, line 374: fyi in case somebody would review, just checked, txcount and sha_message_length_o all seem to be in bits

johannheyszl (approved):
Thanks @gdessouky

moidx:
Ran the following tests by hand on this commit: 

1. `//sw/device/tests/crypto:hmac_functest`
2. `//sw/device/tests:hmac_smoketest`
3. `//sw/device/tests/crypto:hmac_multistream_functest`
4. `//sw/device/tests/crypto/cryptotest:sha256_kat`

moidx:
Also ran the following test that failed in CI: `//sw/device/tests/autogen:plic_all_irqs_test_0_fpga_cw310_test_rom`. Unable to reproduce, and issue does not seem related to this PR.

Issue #22932, opened 2024-05-02
[hmac/dv] Verify wider digest/configurable key length in hmac_test_vectors tests


gdessouky:
### Description

Extend `hmac_test_vectors_sha_vseq.sv `and `hmac_test_vectors_hmac_vseq.sv` tests to verify the newer digest sizes and configurable key lengths and use the same constraints on digest_size and key_length inputs from `hmac_base_vseq.sv`. ``

andreaskurth:
FYI @ballifatih is working on a similar test at the top-level (updating cryptolib driver for wider digest / configurable key length, then adding test vectors in TLTs)

ballifatih:
I think the more proper way to test it at the top-level would be to extend the TLT `hmac_enc_test`, but this is not exactly what I am doing. Instead, I have started to implement the driver and the functest of the driver directly to address this #22731. I defined this functest as top-level test so I could look at the waveforms directly.

So far I was able to stream for SHA-256 and HMAC-256 correctly for various-sized vectors. At the moment I am trying to stream SHA-512, which is not working right out of the box, so I have to dig deeper to see where the issue is (could also be the driver). My branch is here, but it is quite messy at the moment: https://github.com/ballifatih/opentitan/tree/mv_hmac_impl

I also implemented a script to generate arbitrary-sized vectors to get better coverage. Just mentioning these details in case there is anything useful from my branch for block or top-level testing.

ballifatih:
On the top-level this is covered with 2 top-level tests, so the focus of this issue is block-level, AFAIU. I am unassigning myself.

PR #23541, opened 2024-06-06
[hmac,dv] Block-level tests with SHA-2 NIST test vectors for new digest sizes 384/512

gdessouky:
This adds block-level testing with SHA-2 NIST test vectors for newer digest sizes 384/512, which addresses https://github.com/lowRISC/opentitan/issues/22932 in part. 
Block-level testing for HMAC with different key lengths is not yet added. Once in, this would close out https://github.com/lowRISC/opentitan/issues/22932.


hw/dv/sv/test_vectors/test_vectors_pkg.sv
```diff
@@ -17,10 +17,16 @@ package test_vectors_pkg;
...
                                    "vectors/sha/sha256/SHA256LongMsg.rsp"
                                   };
-  string hmac_file_list[]       = {"vectors/hmac/HMAC_RFC4868.rsp"};
+  string sha2_384_file_list[]   = {"vectors/sha/sha384/SHA384ShortMsg.rsp",
+                                   "vectors/sha/sha384/SHA384LongMsg.rsp"
+                                  };
+  string sha2_512_file_list[]   = {"vectors/sha/sha512/SHA512ShortMsg.rsp",
+                                   "vectors/sha/sha512/SHA512LongMsg.rsp"
+                                  };
+  string hmac_256_file_list[]   = {"vectors/hmac/HMAC_RFC4868.rsp"};
```
martin-velay, line 29: Why not the HMAC 384/512?

hw/dv/sv/test_vectors/test_vectors_pkg.sv
```diff
@@ -141,11 +147,11 @@ package test_vectors_pkg;
     end
   endfunction : open_file
 
-  // This function parses all NIST test vectors for SHA256/HMAC/SHA3SHAKE/CSHAKE/KMAC.
+  // This function parses all NIST test vectors for SHA256/SHA384/SHA512/HMAC/SHA3SHAKE/CSHAKE/KMAC.
```
martin-velay, line 150: Here maybe precise SHA2:
`SHA2-256/SHA2-384/SHA2-512/HMAC/SHA3SHAKE/CSHAKE/KMAC`

hw/dv/sv/test_vectors/test_vectors_pkg.sv
```diff
@@ -141,11 +147,11 @@ package test_vectors_pkg;
...
   //
   // The test vectors for these algorithms can be formatted in a variety of ways:
   //
-  // 1) This format is used for SHA256/SHA3 test vectors.
+  // 1) This format is used for SHA256/SHA384/SHA512/SHA3 test vectors.
```
martin-velay, line 154: Same as above: Here maybe precise SHA2:
`SHA2-256/SHA2-384/SHA2-512/HMAC/SHA3SHAKE/CSHAKE/KMAC`

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -54,23 +52,6 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
       1 :/ 10,  // SHA2_256/SHA2_384/SHA2_512/SHA2_None
       0 :/ 0    // Illegal -> should get casted to SHA2_None in HW
```
martin-velay, line 53: Maybe add a TODO here, as Illegal won't be hit and same for the digest size.
Do we need to raise an issue for that?

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -42,8 +42,6 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
     };
   }
 
-  // TODO: keep both blocks for key_length and digest size and toggle between both for WIP
-  // till DV stabilizes for all digest sizes and key lengths
   constraint key_digest_c {
```
martin-velay, line 45: I would prefer to have separated constraints, it will allow us more flexibility if we need to release one for a specific test:

```
  constraint key_length_c {
    $countones(key_length) == 1 dist {
      1 :/ 10,  // Key_128/Key_256/Key_384/Key_512/Key_1024/Key_None
      0 :/ 0    // Illegal -> should get casted to Key_None in HW
    };
  }

  constraint digest_size_c {
    $countones(digest_size) == 1 dist {
      1 :/ 10,  // SHA2_256/SHA2_384/SHA2_512/SHA2_None
      0 :/ 0    // Illegal -> should get casted to SHA2_None in HW
    };
  }

```

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_hmac_vseq.sv
```diff
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// This vseq will read hmac vectors and compare the expected value
+// This vseq will read HMAC test vectors and compare computed results with the expected value
```
martin-velay, line 5: As we are changing this comment, can precise from where those vectors are coming as I spent time to understand the purpose of this test. I didn't get at all that those vectors are actually provided by the NIST as a golden reference.

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_hmac_vseq.sv
```diff
@@ -13,7 +13,9 @@ class hmac_test_vectors_hmac_vseq extends hmac_test_vectors_sha_vseq;
   }
 
   task body();
-    vector_list = test_vectors_pkg::hmac_file_list;
+    vector_list_256 = test_vectors_pkg::hmac_256_file_list;
+    //vector_list_384 = test_vectors_pkg::hmac_384_file_list;
```
martin-velay, line 17: Why commented out? Maybe add a TODO with an issue #?

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -2,52 +2,30 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// This vseq will read a few NIST vector format sha or hmac test vectors
-// and compare the expected result
+// This vseq will read SHA-2 NIST vectors and compare computed results with the expected value
```
martin-velay, line 5: Same as above, can we say that those vectors and expected values are provided by the NIST as a golden reference?

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -2,52 +2,30 @@
...
   `uvm_object_utils(hmac_test_vectors_sha_vseq)
   `uvm_object_new
 
-  rand bit [31:0] key[]; // random because SHA-2 will not use the key
-  // TODO (#22932): remove this and use the constraints from hmac_base_vseq once
-  // this test is adapted for all digest sizes and key lengths
-  rand bit [5:0]  key_length;
-  rand bit [3:0]  digest_size;
+  bit [31:0] key[]; // SHA-2 tests will not use the key but the HMAC tests will
+  bit [5:0]  key_length;
```
martin-velay, line 12: I think `key_length` and `digest_size` are already declared into the `hmac_base_vseq`. So it should be removed from here.

Same for the `key`

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -97,9 +75,35 @@ class hmac_test_vectors_sha_vseq extends hmac_base_vseq;
         csr_rd(.ptr(ral.intr_state), .value(intr_state_val));
         csr_wr(.ptr(ral.intr_state), .value(intr_state_val));
         // read digest and compare with the expected result, scb will calculate and check too
-        compare_digest(parsed_vectors[j].exp_digest);
+        compare_digest(digest_size, parsed_vectors[j].exp_digest);
```
martin-velay, line 78: Can you swap `digest_size` with `parsed_vectors[j].exp_digest` tohave the same ordering as above? (and needs to be changed into the `base_seq` as well

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -97,9 +75,35 @@ class hmac_test_vectors_sha_vseq extends hmac_base_vseq;
...
       end
     end
+  endtask
+
+  task body();
+    // SHA-2/HMAC 256
+    digest_size = 4'b0001;
```
martin-velay, line 85: Maybe better to point to the pkg parameters:
`SHA2_256` and `Key_256`

Same for the 2 others below

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -97,9 +75,35 @@ class hmac_test_vectors_sha_vseq extends hmac_base_vseq;
...
+    key_length  = 6'b000010;
+    `uvm_info(`gfn, $sformatf("SHA-2/HMAC 256 NIST test vectors..."), UVM_LOW)
```
martin-velay, line 87: Add something like:
```
uvm_info(`gfn, $sformatf("Starting SHA-2/HMAC 256 NIST test vectors..."), UVM_LOW)
```

martin-velay (approved):
Could be merged once the comments have been taken into account

hw/dv/sv/test_vectors/test_vectors_pkg.sv
gdessouky: Here it is only the format for SHA-2 256/SHA-2 384/SHA-2 SHA512/HMAC/SHA3 test vectors. CSHAKE and KMAC test vectors format come below.

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
gdessouky: does it make a difference? `feed_vectors` is a different function and its parameter is the vector list unparsed, so it is not the same, but I am fine with either order :)

martin-velay: No it's more to keep the same logic and it's something I have in my PR https://github.com/lowRISC/opentitan/pull/23116 as it was causing me some troubles, and I took the order with digest at the end 😁

hw/dv/sv/test_vectors/test_vectors_pkg.sv
martin-velay: Oh yes, sure it's bad copy/paste from me

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
gdessouky: good catch, yes

gdessouky:
Thanks @martin-velay I've pushed again now but a single test run (when it's not smoke) is now very long since it runs through all test vectors for each digest size: SHA-2 256 then 384 then 512, and some time out. I think we should be splitting these up into 3 tests for each digest size and we shouldn't merge now.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
gdessouky: I tested them at some point in the weeks leading up to M4, to make sure they are correctly cast into `SHA2_None/Key_None` and predicted correctly in the scb, but I'll rework the constraint to hit these and run a block-level regression at my end to make sure.

gdessouky:
I've pushed again to this PR and split the test into 3 tests for each of the digest sizes (I think some timeouts might happen still but we'll see and adjust accordingly), and also released the constraints to hit illegal `digest size/key length`. I've tested/implemented changes for this before to make sure they are casted correctly in the HW and predicted correctly in the scoreboard, but I'm still running a local block-level regression at my end to confirm. Once merged, this should also close https://github.com/lowRISC/opentitan/issues/23592.

martin-velay (changes requested):

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -181,14 +181,12 @@ package hmac_env_pkg;
...
-      'h10: return 1024;  // 1024-bit secret key
-      'h20: return 0;     // Unsupported/invalid values and all-zero values are mapped to Key_None
-      default: `uvm_fatal("get_key_length", $sformatf("Invalid key length register value: %0h",
-                          key_length_reg))
+      'h01:    return 128;   // 128-bit secret key
+      'h02:    return 256;   // 256-bit secret key
+      'h04:    return 384;   // 384-bit secret key
+      'h08:    return 512;   // 512-bit secret key
+      'h10:    return 1024;  // 1024-bit secret key
+      default: return 0;     // Unsupported/invalid/all-zero values are mapped to Key_None
```
gdessouky, line 189: @martin-velay this is to also return 0 for all invalid input when we release the constraints to include invalid digest sizes/key lengths.

gdessouky:
@martin-velay, thanks for your feedback and comments, I've addressed them now in my last push after rebasing.
@vogelpi, can we go ahead and merge?

hw/ip/hmac/dv/env/hmac_env_pkg.sv
martin-velay: Yes, I had done this modif but it looks like it has been lost during my rebase somehow. Thanks and sorry!

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -460,24 +458,23 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
   // TODO (#23288): remove this check from the seq
   virtual task compare_digest(bit [7:0] exp_digest[], bit [3:0] digest_size_i);
     bit [TL_DW-1:0] act_digest[16];
-    bit [TL_DW-1:0] packed_exp_digest[8];
+    bit [TL_DW-1:0] packed_exp_digest[16];
+    int digest_index;
```
martin-velay, line 462: This `digest_index` is not used from what I can see

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -2,47 +2,32 @@
...
-    //   1 :/ 8,  // SHA2_256/SHA2_384/SHA2_512/SHA2_None
-    //   0 :/ 2   // Illegal -> should get casted to SHA2_None in HW
-    // };
-  }
-
   virtual task pre_start();
     cfg.save_and_restore_pct  = 0;    // Should not be triggered for this test
     do_hmac_init              = 1'b0;
+    // grab SHA-2 digest size from the command-line argument
+    void'($value$plusargs("sha2_digest_size=%s", digest_size_arg));
```
martin-velay, line 26: Here you should probably add a check like:
```
if (digest_size_arg == "") begin
  `uvm_fatal(`gfn, "This test sequence requires a command-line extra argument: sha2_digest_size=SHA2_256/SHA2_384/SHA2_512"
end
```

martin-velay (changes requested):
There is at least one change which will require a change in case the sequence is misused

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
gdessouky: true, not anymore, thanks for catching.

martin-velay (approved):
Thanks for the changes, now it LGTM!

rswarbrick (approved):
This looks sensible to me and I'm pleased to see that @martin-velay (who knows much more about HMAC than I do) agrees. Thanks!

gdessouky:
The failing CI tests are unrelated to the changes in this PR, can we go ahead and merge @rswarbrick?

PR #23545, opened 2024-06-06
[otbn,dv] Add a partial wipe test

rswarbrick:
The real test is in the final commit of this PR, but forcing some signals inside the design broke some DV code assumptions about what should happen in a trace. Those are weakened/fixed in the other commits in the PR.

### DIFF
--- a/hw/ip/otbn/dv/model/otbn_trace_checker.cc
+++ b/hw/ip/otbn/dv/model/otbn_trace_checker.cc
@@ -60,6 +60,20 @@ void OtbnTraceChecker::AcceptTraceString(const std::string &trace,
     return;
   }
 
+  // If the trace type is Stray then this is a change that happened at an
+  // unexpected time. We're expecting the RTL and model to lock shortly and can
+  // discard this trace entry otherwise.
+  //
+  // To avoid things being really mysterious, write this to stderr as well. It
+  // shouldn't happen very often, so we won't generate much text.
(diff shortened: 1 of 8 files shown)


vogelpi (approved):
Thanks @rswarbrick , I can't judge the changes to the tracer because it's out of my area of expertise but the newly created test looks excellent!

rswarbrick:
Well, that was rather embarrassing once I'd finally worked out what was going on. I intentionally made a change that would add the first line (the "header") to each entry by prepending at the end, rather than just appending it (to an empty string) at the start.

However, I hadn't realised that the stuff that we generate in `prepend_trace_header` might actually have more than one "trace header", because there was more than one item to write out. As such, if we were generating something like
```
append hdr1
append hdr2
append body2
```
(giving `hdr1\nhdr2\nbody2`), my initial version of the change was generating
```
append body2
prepend hdr1
prepend hdr2
```
But that's wrong! You get `hdr2\nhdr1\nbody2`.

The fix is just to switch around the order that things get prepended, so we get
```
append body2
prepend hdr2
prepend hdr1
```

I'm feeling a little silly about how long this took to debug...

Also, the names are a bit stupid! The function that I called `prepend_trace_header` now usually prepends a trace header. But it might also prepend an entire other trace item. That problem was there already: there was a `trace_header` function, which did something similar.

Probably a problem for a different PR.

PR #23546, opened 2024-06-06
[spi_device] Block flash write commands when busy

### LINKED ISSUE #21700: [spi_device] Disallow certain state changes when busy
### Description

On a real SPI flash device, many commands are ignored when the device is busy. We might want to consider adding this behavior to our IP as well.

Currently, command uploads, WREN, WRDI, EN4B, and EX4B all are allowed while the busy bit is set.

a-will:
Block uploads, WREN, WRDI, EN4B, and EX4B when the busy bit is set. The state of the device should not change until the busy bit is cleared.

This may be especially important to avoid the payload buffer changing its contents while software is processing a command.

Fixes #21700

### DIFF
--- a/hw/ip/spi_device/rtl/spi_cmdparse.sv
+++ b/hw/ip/spi_device/rtl/spi_cmdparse.sv
@@ -30,6 +30,10 @@ module spi_cmdparse
   // Read Status module as Index 0 is pre-assigned to Read Status.
   input cmd_info_t [NumTotalCmdInfo-1:0] cmd_info_i,
 
+  // Busy status at the beginning of the transaction.
+  // Used to block some commands.
+  input sck_status_busy_i,
+
   // control to spi_s2p
   output io_mode_e io_mode_o,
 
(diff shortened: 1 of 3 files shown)


moidx (approved):

moidx:
@antmarzam can you review as well? (even after post-merge in case this is merged before you can take a look).

moidx:
Merging now so that this can be picked up in the next regression. @antmarzam please still provide a review if you can. Thanks.

nbdd0121:
This pull request caused spi_passthru_test_fpga_cw310_rom_with_fake_keys to break (for this PR and all subsequent CI runs after it was merged).

Is it intentional that this change transitions the cmdparse state machine to StWait when a write command is received when busy? StWait is a terminal state that only reset can bring the state machine out.

a-will:
So it looks like the host side of spi_passthru doesn't actually read the flash status register after performing an upload command, and this PR made that a requirement for non-passthrough commands. Note that it was already required for passthrough commands. (Technically, the requirement is that some other command is issued, and it doesn't particularly matter which one, except it can't be one of these blocked ones.)

The host side code should be changed to exhibit the proper behavior. Also, yay, this feature got some unexpected coverage. ;)

PR #23548, opened 2024-06-06
[rom] Support both big- and little-endian digests in the HMAC driver.

jadephilipoom:
Replacement for https://github.com/lowRISC/opentitan/pull/23527

To allow SPHINCS+ to use HMAC in big-endian order without disrupting existing infrastructure, this adds a configuration option to the HMAC driver that lets the caller set the digest-endianness configuration bit. Because that bit only changes the order of bytes within words (not words themselves), we then have to check the bit in `final` to read the words in the right order. Thanks @cfrantz for your help on this one!

Overall the code size of `hmac.c` actually goes down as a result of this change (302B -> 284B) because `hmac_256` takes up less space. Compilers are mysterious beasts.


jadephilipoom:
Closing in favor of https://github.com/lowRISC/opentitan/pull/23544

PR #23549, opened 2024-06-06
[hmac,rtl] Coding style fixes

gdessouky:
This implements minor fixes to the coding style, e.g., defaults on top, etc. to align with our conventions with no functionality impact. Purely cosmetic :-)

### DIFF
--- a/hw/ip/hmac/rtl/hmac.sv
+++ b/hw/ip/hmac/rtl/hmac.sv
@@ -147,7 +147,7 @@ module hmac
   done_state_e done_state_d, done_state_q;
 
   always_comb begin
-    done_state_d = done_state_q;
+    done_state_d    = done_state_q;
     hash_done_event = 1'b0;
 
     unique case (done_state_q)
@@ -284,6 +284,8 @@ module hmac
 
   assign digest_size_supplied = digest_mode_e'(cfg_reg.digest_size.q);
   always_comb begin : cast_digest_size
+    digest_size = SHA2_None;
+
     unique case (digest_size_supplied)
       SHA2_256:  digest_size = SHA2_256;
       SHA2_384:  digest_size = SHA2_384;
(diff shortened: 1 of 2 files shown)


hw/ip/hmac/rtl/hmac_core.sv
```diff
@@ -318,6 +322,7 @@ module hmac_core import prim_sha2_pkg::*; (
     hash_start         = 1'b0;
     hash_continue      = 1'b0;
     hash_process       = 1'b0;
+    st_d               = st_q;
```
johannheyszl, line 325: thanks. this one especially I like because the case / if construction is complex and this way is more robust to avoid mistakes.

johannheyszl (approved):
Thanks @gdessouky . I like the improvements. As you said, this is not mandatory for M4.

As you said this is cosmetical - there should be no functional impact whatsoever (and if it is it would be good to catch because it would mean it was coded unintentionally).

hw/ip/hmac/rtl/hmac.sv
```diff
@@ -284,6 +284,8 @@ module hmac
 
   assign digest_size_supplied = digest_mode_e'(cfg_reg.digest_size.q);
   always_comb begin : cast_digest_size
+    digest_size = SHA2_None;
```
rswarbrick, line 287: This looks sensible to me. I wonder whether we should do the same as you did in `hmac_core.sv` and make the `default` case empty? Probably the comment about unsupported digest size values can also move to apply to this line?

Similarly for the next two case statement changes.

hw/ip/hmac/rtl/hmac.sv
```diff
@@ -823,7 +828,9 @@ module hmac
                    | hash_start_active | msg_push_not_allowed | invalid_config_atstart);
 
   always_comb begin
+    // default
```
rswarbrick, line 831: I'm not sure that I understand this comment. Do you want something like "The default situation (where none of the error flags is set) should be interpreted as NoError" ?

rswarbrick (approved):
This looks good. I had a couple of questions about ways the code could be "even more blindingly obvious", but am very happy with the current code too.

PR #23555, opened 2024-06-07
[prim] Remove primgen and replace with virtual cores

a-will:
Add back prim_pkg enum for legacy support, and mention that it is deprecated. Some IPs reference the enums directly.


a-will:
Interesting, it looks like mixing with a newer fusesoc might require at least Python 3.9 to work with the other dependencies.

a-will:
Newer fusesoc also has a different directory layout compared to the OT fork. This was the OT fork's layout:

- _{--build-root}/_
  - _sim-vcs/_
    - fusesoc-created metadata, Makefile, and file list
  - _src/_
    - Exported source files

But newer fusesoc has this directory layout:

- _{--work-root}/_
  - Contains fusesoc-created metadata, Makefile, and file list
  - _src/_
    - Exported source files

Accommodating the new tree needed fixes to various paths, including CFLAGS definitions in hjson files. In the current example, _{work-root}_ is set to dvsim's `{build_dir}/fusesoc-work`.

In addition, for ralgen, the `position` field was needed to be set to `prepend` for generated cores, else the files would come after the *_env_pkg.sv that would depend on them.

a-will:
This PR is way out of date now, but since @olofk has released fusesoc 2.4, I think we don't need my random development tag anymore. 2.4 should be able to handle virtual cores as needed.

olofk:
@a-will I think that depends on the OT requirements. If you are fine with pulling in cores from one vendor during a build, then virtual cores should do the trick.

However, as I understand it, OT wants to be able to select implementation at compile-time, and in that case you still need something like primgen.

With that said, I believe FuseSoC 2.4 should have all the functionality needed to implement what you need. I have done some experimenting implementing the equivalent to primgen as a FuseSoC filter instead of a generator and it looks promising.

Happy to discuss this further.

a-will:
> @a-will I think that depends on the OT requirements. If you are fine with pulling in cores from one vendor during a build, then virtual cores should do the trick.
> 
> However, as I understand it, OT wants to be able to select implementation at compile-time, and in that case you still need something like primgen.
> 
> With that said, I believe FuseSoC 2.4 should have all the functionality needed to implement what you need. I have done some experimenting implementing the equivalent to primgen as a FuseSoC filter instead of a generator and it looks promising.
> 
> Happy to discuss this further.

It's unclear to me if OT actually _needs_ that, though. All of our in-tree top-level fusesoc cores end up binding a specific prim library, and an out-of-tree integrator would almost certainly do the same for their actual chip. The build recipes for synthesis often can't be reused across technologies, and across integrations for the same opentitan IP, the top-level generally uses different RTL, so we end up with independent top-level cores anyway. We'd end up with different YAML for each real target, so sharing that core file across implementations doesn't seem to provide a benefit.

There may be a bit of a rub with gate-level simulation applications, though, especially if you just want to redo the existing block-level simulations with your own prim library. I'd guess that the full integration level still uses its own core file for GLS, but is doing block-level simulation with a different prim library a supported activity? I'm not sure.

We'd probably achieve the original setup's capabilities if only we could provide fusesoc (as a parameter to its invocation) which implementations of virtual cores to include for the specified top-level core file. For example, it could be arguments to provide additional VLNVs to add to the build (using their default target). Then the core file wouldn't need to explicitly pull them in.

a-will:
> For example, it could be arguments to provide additional VLNVs to add to the build (using their default target). Then the top-level core file wouldn't need to explicitly pull them in.

@olofk If that bit above is interesting, the specifying of additional VLNVs would be akin to how hierarchical synthesis flows work. The top-level (and other sub-cores) may have dangling references, but the missing netlists would get specified as additional sources, then linked in during the build. For fusesoc, this would mean writing in a "depends" arrow from the top-level VLNV to the additional VLNVs, and the solver should handle them as though they were written in the core file.

a-will:
Looks like fusesoc 2.4 isn't being used in the sim runs in CI, and something broke with the CI file changes. I guess the API is a bit different for github actions. Maybe `s/parameters/inputs`?

hw/ip_templates/flash_ctrl/dv/env/flash_ctrl_env.core
```diff
@@ -15,6 +15,8 @@ filesets:
       - lowrisc:dv:flash_phy_prim_agent
       - lowrisc:ip_interfaces:flash_ctrl_pkg
       - lowrisc:constants:top_pkg
+      # TODO: prim_pkg_legacy is deprecated
+      - lowrisc:prim:prim_pkg_legacy
```
marnovandermaas, line 19: Maybe good to put this in a separate commit, just to separate it from the prepend changes.

hw/ip/otbn/dv/uvm/env/otbn_env.core
```diff
@@ -73,6 +73,7 @@ generate:
     parameters:
       name: otbn
       ip_hjson: ../../../data/otbn.hjson
+    position: prepend
```
marnovandermaas, line 76: Should this go in the previous commit?

hw/ip/otp_ctrl/otp_ctrl.core
```diff
@@ -25,6 +25,8 @@ filesets:
       - lowrisc:ip_interfaces:pwrmgr_pkg
       - lowrisc:ip:edn_pkg
       - lowrisc:prim:sparse_fsm
+      # TODO: prim_pkg is deprecated
+      - lowrisc:prim:prim_pkg
```
marnovandermaas, line 29: Will we create an issue to track this?

hw/ip/prim_generic/rtl/prim_generic_flash_bank.sv
```diff
@@ -452,4 +452,4 @@ module prim_generic_flash_bank #(
   logic unused_he;
   assign unused_he = he_i;
 
-endmodule // prim_generic_flash
```
marnovandermaas, line 455: Should the module name itself also be changed?

hw/ip_templates/flash_ctrl/flash_ctrl.core.tpl
```diff
@@ -21,6 +21,8 @@ filesets:
       - lowrisc:prim:lfsr
       - lowrisc:prim:secded
       - lowrisc:prim:sparse_fsm
+      # TODO: prim_pkg_legacy is deprecated
+      - lowrisc:prim:prim_pkg_legacy
```
marnovandermaas, line 25: This change is similar to a change made in an earlier commit.

hw/top_earlgrey/cdc/cdc_waivers.data.tcl
marnovandermaas, line 1: Are these CDC waivers tested in CI? If not, have we tested this ourselves?

hw/top_earlgrey/chip_earlgrey_asic.core
```diff
@@ -14,6 +14,8 @@ filesets:
       - "fileset_partner ? (partner:systems:scan_role_pkg)"
       - "!fileset_partner ? (lowrisc:systems:ast)"
       - "!fileset_partner ? (lowrisc:systems:scan_role_pkg)"
+      - "!fileset_partner ? (lowrisc:prim:prim_pkg_legacy)"
```
marnovandermaas, line 17: Should this have a todo that this will be removed in the future?

marnovandermaas: Also true for other core files in top_earlgrey

hw/top_englishbreakfast/chip_englishbreakfast_cw305.core
```diff
@@ -12,6 +12,8 @@ filesets:
       - lowrisc:systems:ast
       - lowrisc:systems:topgen
       - lowrisc:systems:padring
+      - lowrisc:prim:prim_pkg_legacy
```
marnovandermaas, line 15: Add todo?

marnovandermaas (commented):
Amazing PR! Left a few comments from looking at the code.

olofk:
Regarding paths, I haven't checked exactly what the problem is, but I would recommend using `--work-root` instead of `--build-root` since that gives you more deterministic behaviour. You might also want to use `--system-name` to get a fixed name for the build artifacts.

olofk:
> > Regarding paths, I haven't checked exactly what the problem is, but I would recommend using `--work-root` instead of `--build-root` since that gives you more deterministic behaviour.
> 
> That is what I did in this PR: https://github.com/lowRISC/opentitan/pull/23555/files#diff-8ed9c1256f9282d31ab3a7f150e2e95eaae62a546151c7203117aaa70451129e

Great. Sounds like you're on top of things then.

a-will:
Hm... it appears fusesoc 2.4 performs unsafe actions with generators that aren't cached. Instead of generating files in a temporary directory owned by that specific call to fusesoc, it calls the generator from _the same directory_ that would've been used if it _were_ cached, and if the directory already existed, it simply _deletes_ the existing directory tree from the real cache root. This causes two simultaneous calls to fusesoc to race each other for access.

@olofk FYI, this causes problems for parallel builds: https://github.com/olofk/fusesoc/blob/8d3d6bca946c713c0804a138e03565014f53cb0f/fusesoc/edalizer.py#L721-L727

It looks like we can work around this either by using `--no-export` or by controlling the `cache_dir` config like we do with fusesoc calls from bazel.

HU90m:
## Status update

- Almost the whole of the public CI is passing. The exceptions are the CW305 build (caused by a simple typo) and the OTBN smoke test (see below).
- The private CI's main blocker is the [generator cache problem](https://github.com/lowRISC/opentitan/pull/23555#issuecomment-2513230085). The ascent lint reports are now being passed and just require a few waiver updates.

### Tasks Remaining 

#### OTBN Smoke Test

Starting with this because it's the biggest head scratcher.
For some reason the following lines, that appear in `VL_MODULE(Votbn_top_sim_otbn_loop_controller__Id)` (within `Votbn_top_sim_otbn_loop_controller__Id.h`) in master, no longer appear in after the FuseSoC update.

```cpp
class Votbn_top_sim_otbn_stack__S22_SB8;
class Votbn_top_sim_prim_flop__W8_Rz22;
class Votbn_top_sim_prim_count__W20_Rffffffff;
```

```cpp
    // CELLS
    Votbn_top_sim_otbn_stack__S22_SB8* loop_info_stack;
    Votbn_top_sim_prim_flop__W8_Rz22* u_loop_counter_err_flop;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__0__KET____DOT__u_loop_count;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__1__KET____DOT__u_loop_count;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__2__KET____DOT__u_loop_count;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__3__KET____DOT__u_loop_count;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__4__KET____DOT__u_loop_count;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__5__KET____DOT__u_loop_count;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__6__KET____DOT__u_loop_count;
    Votbn_top_sim_prim_count__W20_Rffffffff* g_loop_counters__BRA__7__KET____DOT__u_loop_count;
```

I'm not entirely sure what's causing this to happen at the moment.

#### Generator Cache problems

The remaining failures in Private CI's DV smoke tests are due to the generator cache problem outlined in https://github.com/lowRISC/opentitan/pull/23555#issuecomment-2513230085 .


#### Name collisions

The Verible RTL failures in lint (slow) are due to vlnv name collisions of cores such as `lowrisc:systems:padring`. There's a nice comment suggesting this could be fixed with virtual cores:

```yaml
# XXX: This name is currently required as global identifier until we have
# support for "interfaces" or a similar concept.
# Tracked in https://github.com/olofk/fusesoc/issues/235
name: "lowrisc:constants:top_pkg"
```

It sounds like this will be fixed by @andreaskurth  in https://github.com/lowRISC/opentitan/issues/25401 .


#### Waiver updates

The file names in the Ascent lint waivers need to updated.

The module hierarchies need to be updated in `hw/top_earlgrey/{syn,cdc}/`.
Then a CDC and SDC need to be run.


#### Full regressions

Still need to add an explicit dependency on the concrete `prim_generic` for `files_dv` in all `**/*_sim.core` files.
Then a full nightly DV run is needed to weed out any remaining path/hierarchy issues.

a-will:
> #### Full regressions
> Still need to add an explicit dependency on the concrete `prim_generic` for `files_dv` in all `**/*_sim.core` files. Then a full nightly DV run is needed to weed out any remaining path/hierarchy issues.

Perhaps in the future, this is the part where it might be nice if fusesoc provided a way to add core requirements on the command line beyond the "system" or top core. Then the prim implementation could be controlled by modes instead of separate core files.

a-will:
Another problem (which is more on the multi-top front) is that some of our utilities have very poor interfaces. Pulling from `flash_phy_pkg` in `mem_bkdr_util` causes all users to require defining some stuff for a flash_ctrl they don't use: https://github.com/lowRISC/opentitan/blob/bfeb51892f7c903a68f11815422060aab816c108/hw/dv/sv/mem_bkdr_util/mem_bkdr_util__flash.sv#L8-L10

I don't think OTBN should be needing to specify a particular flash_ctrl_pkg to simulate its block-level smoke tests.

a-will:
So, regarding OTBN's simulator software, its coding is simply wrong and violates abstraction boundaries:

https://github.com/lowRISC/opentitan/blob/bfeb51892f7c903a68f11815422060aab816c108/hw/ip/otbn/dv/verilator/otbn_top_sim.cc#L210

Directly grabbing at internal class members is not okay. Verilator's APIs must be used.

a-will:
Okay, fixed OTBN smoke test. `s/prim_generic_flop/prim_flop` for the "waiver" file that is actually supplying critical Verilator configuration to allow the simulator to compile.

olofk:
> Hm... it appears fusesoc 2.4 performs unsafe actions with generators that aren't cached. Instead of generating files in a temporary directory owned by that specific call to fusesoc, it calls the generator from the same directory that would've been used if it were cached, and if the directory already existed, it simply deletes the existing directory tree from the real cache root. This causes two simultaneous calls to fusesoc to race each other for access.

I'm working on a fix for this. Will try to have it fixed ASAP and let you know when I have something to test.

a-will:
I adjusted the work for the lint top-levels + added fixes for the new prim_xilinx_ram_1p + removed unneeded ultrascale cores.

Remember that the implementing prim libraries should only be incorporated in targets that represent a top level / system level (i.e. not the default target). Otherwise, we have a high risk of cyclic dependencies. :slightly_smiling_face:

HU90m:
> Remember that the implementing prim libraries should only be incorporated in targets that represent a top level / system level (i.e. not the default target). Otherwise, we have a high risk of cyclic dependencies. 🙂

Yeah I was just playing when I put the concrete implementation in `lowrisc:lint:common`. I started doing this properly last night before falling asleep. The cyclic dependency giving me trouble yesterday was the following, which it looks like you've beat me to pushing the fix for.

```python
{
    'lowrisc_ip_flash_ctrl_prim_reg_top':            ['lowrisc_opentitan_top_earlgrey_flash_ctrl_pkg'],
    'lowrisc_opentitan_top_earlgrey_flash_ctrl_pkg': ['lowrisc_opentitan_top_earlgrey_pwrmgr_pkg'],
    'lowrisc_opentitan_top_earlgrey_pwrmgr_pkg':     ['lowrisc_opentitan_top_earlgrey_pwrmgr_reg'],
    'lowrisc_opentitan_top_earlgrey_pwrmgr_reg':     ['lowrisc_ip_tlul'],
    'lowrisc_ip_tlul':                               ['lowrisc_tlul_sram2tlul'],
    'lowrisc_tlul_sram2tlul':                        ['lowrisc_prim_generic_all']
    'lowrisc_prim_generic_all':                      ['lowrisc_prim_generic_flash'],
    'lowrisc_prim_generic_flash':                    ['lowrisc_opentitan_top_earlgrey_flash_ctrl_pkg',
                                                      'lowrisc_ip_tlul',
                                                      'lowrisc_ip_flash_ctrl_prim_reg_top'],
}
```

I'll stop working on this PR to avoid hurt toes.

andreaskurth:
@a-will:
> Interesting, it looks like mixing with a newer fusesoc might require at least Python 3.9 to work with the other dependencies.
> 
> ```
> ERROR: Ignored the following versions that require a different python version: 1.0.0 Requires-Python >=3.9,<4.0; 1.1.0 Requires-Python >=3.9
> ERROR: Could not find a version that satisfies the requirement jsonschema2md==1.1.0 (from versions: 0.1.0, 0.1.1, 0.2.0.post1, 0.2.1, 0.3.0, 0.4.0, 0.9.0)
> ERROR: No matching distribution found for jsonschema2md==1.1.0
> ```

As [Python 3.9 is currently the oldest version of Python that still receives security fixes](https://devguide.python.org/versions/), I think we can reasonably increase the minimum required Python version for OT to 3.9. Should we change https://github.com/lowRISC/opentitan/blob/22168a3d42febfca725b36dd0d1554b68cf5783e/pyproject.toml#L9
accordingly?

olofk:
> @a-will:
> 
> > Interesting, it looks like mixing with a newer fusesoc might require at least Python 3.9 to work with the other dependencies.
> > ```
> > ERROR: Ignored the following versions that require a different python version: 1.0.0 Requires-Python >=3.9,<4.0; 1.1.0 Requires-Python >=3.9
> > ERROR: Could not find a version that satisfies the requirement jsonschema2md==1.1.0 (from versions: 0.1.0, 0.1.1, 0.2.0.post1, 0.2.1, 0.3.0, 0.4.0, 0.9.0)
> > ERROR: No matching distribution found for jsonschema2md==1.1.0
> > ```
> 
> As [Python 3.9 is currently the oldest version of Python that still receives security fixes](https://devguide.python.org/versions/), I think we can reasonably increase the minimum required Python version for OT to 3.9. Should we change
> 
> https://github.com/lowRISC/opentitan/blob/22168a3d42febfca725b36dd0d1554b68cf5783e/pyproject.toml#L9
> 
> 
> accordingly?

Curious from a FuseSoC perspective about how this happens. FuseSoC itself supports python 3.7+.

andreaskurth:
@olofk:
> Curious from a FuseSoC perspective about how this happens. FuseSoC itself supports python 3.7+.

I think that comes because [`jsonschema2md` seems to require Python 3.9+](https://github.com/sbrunner/jsonschema2md/blob/c13ebe5dd5d201659bde88ba060451f908981065/pyproject.toml#L32).

HU90m:
> @olofk:
> 
> > Curious from a FuseSoC perspective about how this happens. FuseSoC itself supports python 3.7+.
> 
> I think that comes because [`jsonschema2md` seems to require Python 3.9+](https://github.com/sbrunner/jsonschema2md/blob/c13ebe5dd5d201659bde88ba060451f908981065/pyproject.toml#L32)

The latest FuseSoC seems to work happily on 3.8.2 in CI. I had to add explicitly set `--python-version 3.8.2` in https://github.com/lowRISC/opentitan/pull/23555/commits/f049f9679db7c1715b447407993e599fc7857ff7 to stop too newer version of `jsonschema2md` from being installed.

FYI: I came across https://github.com/lowRISC/opentitan/pull/25459 the other day.

olofk:
> > Hm... it appears fusesoc 2.4 performs unsafe actions with generators that aren't cached. Instead of generating files in a temporary directory owned by that specific call to fusesoc, it calls the generator from the same directory that would've been used if it were cached, and if the directory already existed, it simply deletes the existing directory tree from the real cache root. This causes two simultaneous calls to fusesoc to race each other for access.
> 
> I'm working on a fix for this. Will try to have it fixed ASAP and let you know when I have something to test.

@a-will @HU90m I got this this PR to address the problem. https://github.com/olofk/fusesoc/pull/713 This always builds non-cached generators in work_root which seems to be a reasonable solution to me. Intend to follow up with another fix to lock the directory of cached generators before trying to write those, but wanted to get this to you first for comments.

HU90m:
## Status update

OTBN Smoke Test failure fixed thanks @a-will 

### Remaining tasks in for this PR
- Private CI should be all be passing with the patches in https://github.com/olofk/fusesoc/pull/713 thanks @olofk 
  - We just need to run check the patches do solve the problems we've been seeing.
- CW305 build was passing it's failing again, likely due to changes https://github.com/lowRISC/opentitan/pull/25533
  - Should be a simple fix that I missed when re-basing.
- Ascent lint failures
  - Good progress, unexpected dependency problems, now fixed
  - Only 7 lint errors remaining, no flow errors
- Full nightly DV regression
  - Explicit dependency on the concrete `prim_generic` for `files_dv` in all `**/*_sim.core` files has been added.
  - Haven't run a full regression to check all tests are passed, some remaining path/hierarchy issues expected.
- FPV, SDA, CDC
  - Problems encountered and fixed with Ascent lint flow errors should help a lot in takling these.
  - Files still need to be updated and tests run.

### Other failures
- Still expect name collision causing Verible RTL failures in lint (slow) to fail.
  - Fixed in #25401 .

Razer6:
@HU90m I just looked into the linter output of the CI task. This task is AFAIK run for top_earlgrey. This report yields warnings like: 

```
WARNING: Replacing lowrisc:dv:chip_ral_pkg:0 in /home/runner/work/opentitan/opentitan/build/regs-generated/dv/top_darjeeling with the version found in /home/runner/work/opentitan/opentitan/build/regs-generated/dv/top_earlgrey

WARNING: Replacing lowrisc:systems:topgen-reg-only:0 in /home/runner/work/opentitan/opentitan with the version found in /home/runner/work/opentitan/opentitan/bazel-opentitan

WARNING: Replacing lowrisc:tool:check_tool_requirements:0.1 in /home/runner/work/opentitan/opentitan with the version found in /home/runner/work/opentitan/opentitan/bazel-opentitan

WARNING: Replacing lowrisc:systems:topgen:0 in /home/runner/work/opentitan/opentitan with the version found in /home/runner/work/opentitan/opentitan/bazel-opentitan

WARNING: Replacing lowrisc:systems:physical_pads:0.1 in /home/runner/work/opentitan/opentitan/hw/top_earlgrey with the version found in /home/runner/work/opentitan/opentitan/hw/top_englishbreakfast

WARNING: Replacing lowrisc:constants:top_pkg:0 in /home/runner/work/opentitan/opentitan/hw/top_earlgrey with the version found in /home/runner/work/opentitan/opentitan/hw/top_englishbreakfast

WARNING: Replacing lowrisc:systems:padring:0.1 in /home/runner/work/opentitan/opentitan/hw/top_earlgrey with the version found in /home/runner/work/opentitan/opentitan/hw/top_englishbreakfast
```

It seems like it is seeing the English Breakfast files, and uses them instead of the Earlgrey ones. @andreaskurth this is the issue you and I also observed previously, right?

andreaskurth:
If FuseSoC uses `padring`, `physical_pads`, and `top_pkg` from Englishbreakfast instead of Earlgrey, that can be ignored because the former symlink to the latter.

The warnings in which FuseSoC uses cores under `bazel-opentitan` instead of those in the repo is a bit weird; maybe a `FUSESOC_IGNORE` in `bazel-opentitan` is missing?

The conflict between Darjeeling and Earlgrey for `chip_ral_pkg` is most relevant IMO, but it can be fixed in topgen.

None of this should cause the AscentLint errors, which are reported on AST, though

Razer6:
While I agree that it doesn't make a difference for Earlgrey and Englishbreakfast (for the mentioned reason), I am more concerned that the wrong packages are selected in general.

Is there a general issue here?

a-will:
> If FuseSoC uses padring, physical_pads, and top_pkg from Englishbreakfast instead of Earlgrey, that can be ignored because the former symlink to the latter.

I think we should get rid of those symlinks, though. We ought to eliminate duplicate _real_ VLNVs inside the repo. Virtual VLNVs can define common interfaces, and otherwise, englishbreakfast should become a proper top of its own.

> The warnings in which FuseSoC uses cores under bazel-opentitan instead of those in the repo is a bit weird; maybe a FUSESOC_IGNORE in bazel-opentitan is missing?

No, the bazel-opentitan symlink is pointing to a directory of symlinks that point to exactly the same set of files. That is a link to bazel's execroot, which includes all the files from the repo + the external repos pulled in and/or generated + the output directory. We cannot create a FUSESOC_IGNORE file there, and it also wouldn't have any practical effect.

If we really wanted to get rid of the confusion, though, we could play with where the symlinks get created. We could create a directory with this structure:

```
${REPO_TOP}/
    .bazel/
        FUSESOC_IGNORE
```

Commit that to the repo, then add `--symlink_prefix=.bazel/` to the common options. Then, all the symlinks will be created in the `.bazel/` folder alongside that FUSESOC_IGNORE, so non-bazel calls to fusesoc won't exhibit these warnings. (bazel calls do not exhibit these warnings, since fusesoc is run in a sandbox, not in the source directory)

In scripts, we probably foolishly use uncontrolled symlink locations to get to output files, though, so those would have to be fixed up.

> While I agree that it doesn't make a difference for Earlgrey and Englishbreakfast (for the mentioned reason), I am more concerned that the wrong packages are selected in general.
> 
> Is there a general issue here?

So far, nothing raised points to a general issue (except, arguably, that ascentlint should only be used in sign-off flows, since contributors get very low visibility into what's happening in CI with that tool).

a-will:
> I got this this PR to address the problem. https://github.com/olofk/fusesoc/pull/713 This always builds non-cached generators in work_root which seems to be a reasonable solution to me. Intend to follow up with another fix to lock the directory of cached generators before trying to write those, but wanted to get this to you first for comments.

@olofk Thanks for writing that! I created a test release for it (with wrong version numbers, surely) at https://github.com/a-will/fusesoc/releases/tag/2.4.1.dev0%2Ba-will , and the results look good here. :slightly_smiling_face:

a-will:
The ascentlint issues were just more of the missing changes from prim_generic_* to prim_* (in a couple waiver files).

For CI, we're probably very close now. I'll have to check what's happening with the CW305 bitstream--It seems to be failing somehow on the MMI generation, this time in the search for BRAM cells for otp_ctrl.

Even once we get CI to be all green, there will still be issues with the flows that are only activated for sign-off activities (e.g. FPV, CDC, RDC, UNR, coverage collection / waivers, etc.). At least some docs will need to be updated as well.

olofk:
> > I got this this PR to address the problem. [olofk/fusesoc#713](https://github.com/olofk/fusesoc/pull/713) This always builds non-cached generators in work_root which seems to be a reasonable solution to me. Intend to follow up with another fix to lock the directory of cached generators before trying to write those, but wanted to get this to you first for comments.
> 
> @olofk Thanks for writing that! I created a test release for it (with wrong version numbers, surely) at https://github.com/a-will/fusesoc/releases/tag/2.4.1.dev0%2Ba-will , and the results look good here. 🙂

I released FuseSoC 2.4.2 yesterday with this as well as the warning about non-deterministic virtual cores. Could it be that we have now reached the point where you can use a released version of upstream FuseSoC?

a-will:
This has been rebased on the latest commit in master, but there is still some catch-up work to do.

This PR should get significantly smaller with https://github.com/olofk/fusesoc/pull/727. Then we can have dvsim supply the desired virtual provider cores on the command-line (and allow it to be overridden by out-of-tree users!). That will eliminate the many file changes that add a dependency to `lowrisc:prim_generic:all`.

blueluna:
I have been working on a similar feature as olofk/fusesoc#727 for fusesoc under the guidance of @olofk. This feature adds a lock file to fusesoc, which guides the core selection during the fusesoc run. Guiding in selecting core version and virtual core implementations. I opened a draft pull request, olofk/fusesoc#728, for collaboration. We should collaborate on a guided virtual core selection feature.

HU90m:
Lock files are a brilliant addition to FuseSoC that I'm very pleased to see. However, I don't think it solves the same problem as https://github.com/olofk/fusesoc/pull/727 intends to.

Currently we have to add the generic primitives explicitly to a huge number files (e.g. `hw/ip/csrng/csrng.core`):

```yaml
  files_lint:
    depend:
      - lowrisc:prim_generic:all

# ...

    filesets_append:
      - files_lint
```

@a-will 's solution is to allow VLNVs to be added in the CLI instead:

```sh
fusesoc --add-vlnv lowrisc:prim_generic:all lint lowrisc:ip:csrng
```

--- 

Another solution to this problem that would keep the cli cleaner would be to allow users to **set default implementations for a given virtual core in the `fusesoc.conf`**.

olofk:
> Another solution to this problem that would keep the cli cleaner would be to allow users to **set default implementations for a given virtual core in the `fusesoc.conf`**.

Yes! This was perhaps not expressed clearly enough, but the lock files have two purposes. The first is pinning versions, and the other one is exactly what you describe above. Locking down virtuals to a concrete implementation. We felt it was so close in scope with the first purpose of the lock files so we decided to put it in the same file rather than in fusesoc.conf

This is very much developed with OpenTitan in mind, so let us know your thoughts. What we perhaps should consider adding is to allow expressing the virtual->concrete mapping on the command-line similar to what @a-will did with the --add-vlnv feature.

Razer6:
> This is very much developed with OpenTitan in mind, so let us know your thoughts. What we perhaps should consider adding is to allow expressing the virtual->concrete mapping on the command-line similar to what @a-will did with the --add-vlnv feature.

This is exactly what, we as Rivos and also other integrators need. There is the public `prim_generic` library. But integrators will have their own, e.g., a `prim_rivos`. We want to run the block level tests for example on our prim library. Pinning the virtual core provider of an IP specifically to for example `lowrisc:prim_generic:all` also creates a conflict. That is done on a higher level of the hierarchy.

olofk:
I suggest moving the discussion to https://github.com/olofk/fusesoc/pull/728 so that we can work out what is best for everyone involved.

a-will:
I've changed this PR to use a development release that includes the `--add-vlnv` option to show how we'd like that to work. I don't have complete working knowledge of dvsim, so the configs may be more verbose than necessary right now. (And only some of them are done, of course!)

a-will:
TODOs (WIP):

- [ ] Uniquify chip_ral_pkg VLNV. There are conflicts between different tops.
- [ ] Move the bazel symlinks to a subdirectory that contains a FUSESOC_IGNORE file. We get warnings for duplicate cores, but they're the exact same files!

HU90m:
I've auto squashed some fixups, removed some commits that were later reverted and added mappings. The old history can be found in my [fusesoc-compat-6 branch](https://github.com/HU90m/opentitan/tree/fusesoc-compat-6).

I still need to cherry pick and adapt https://github.com/HU90m/opentitan/commit/b02b99663b410f63b4f08fe85e9dd03363a4be71 & https://github.com/HU90m/opentitan/commit/3f414783de38b622162ea35453e3ce4f0b0e3f5b .

Then hopefully only some small changes to get CI back to green.

hw/ip_templates/otp_ctrl/rtl/otp_ctrl.sv.tpl
```diff
@@ -1593,8 +1593,8 @@ end
   `endif
   if (`PRIM_DEFAULT_IMPL == prim_pkg::ImplGeneric) begin : gen_reg_we_assert_generic
```
Razer6, line 1594: `PRIM_DEFAULT_IMPL == prim_pkg::ImplGeneric` that code should be removed at all, right?

hw/dv/sv/sec_cm/prim_sparse_fsm_flop_if.sv
```diff
@@ -23,7 +23,7 @@ interface prim_sparse_fsm_flop_if #(
   string msg_id = $sformatf("%m");
 
   string path = dv_utils_pkg::get_parent_hier($sformatf("%m"));
-  string signal_forced = $sformatf("%s.u_state_flop.q_o", path);
+  string signal_forced = $sformatf("%s.state_o", path);
```
rswarbrick, line 26: To avoid having a "known broken state" in the history, should this commit be merged with the switch to virtual prim cores?

.github/workflows/bitstream.yml
```diff
@@ -82,11 +82,21 @@ jobs:
         if: steps.strategy.outputs.bitstreamStrategy != 'cached'
         run: |
           . util/build_consts.sh
+
```
rswarbrick, line 85: If I understand correctly, this commit is all to do with the change to the new FuseSoc version. Again, maybe we need to squash with that commit so that (we think) everything works at each commit?

hw/ip_templates/flash_ctrl/rtl/flash_ctrl_top_specific_pkg.sv.tpl
```diff
@@ -297,7 +297,7 @@ package flash_ctrl_top_specific_pkg;
                  ecc_en:      MuBi4True,
                  he_en:       MuBi4True, // HW assumes high endurance
                  base:        '0,
-                 size:        NumBanks * PagesPerBank
+                 size:        10'(NumBanks * PagesPerBank)
```
rswarbrick, line 300: This commit looks like a candidate for splitting out into a separate PR?

hw/ip/hmac/dv/hmac_sim_cfg.hjson
```diff
@@ -109,14 +109,14 @@
     {
       name: hmac_test_sha384_vectors
       uvm_test_seq: hmac_test_vectors_sha_vseq
-      run_opts: ["+is_nist_test=1 +test_vectors_dir={build_dir}/src/lowrisc_dv_test_vectors_0 +test_timeout_ns=750_000_000 +sha2_digest_size=SHA2_384"]
+      run_opts: ["+is_nist_test=1 +test_vectors_dir={build_dir}/fusesoc-work/src/lowrisc_dv_test_vectors_0 +test_timeout_ns=750_000_000 +sha2_digest_size=SHA2_384"]
```
rswarbrick, line 112: I think there are quite a lot of commits starting here that should probably be squashed with the fusesoc version bump?

rswarbrick (approved):
Hi there!

This all looks rather sensible, but it looks like there are lots of changes that could land before switching tool version. The result would be that the "big switch" moment would be smaller.

Please feel free to tag me on the split versions!

rules/fusesoc.bzl
```diff
@@ -35,7 +35,7 @@ def _fusesoc_build_impl(ctx):
     # Vivado expects `HOME` environment variable to exist. Redirect it to a fake directory.
     home_dir = "{}/homeless-shelter".format(out_dir)
 
-    cache_dir = "{}/fusesoc-cache".format(out_dir)
+    cache_dir = "/tmp/fusesoc-cache"
```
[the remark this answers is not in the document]
pamaury: After some investigation, I think using such an absolute path will work in practice because `/tmp` in mapped in the sandbox. Here are some alternatives which might be better:
- assuming that one can use environment variables in the fusesoc cfg file, setting `cache_dir = "$TMPDIR"` should work because `$TMPDIR` gets set to a temporary directory for most bazel runners. This is not well documented however and I am not convinced by looking at the [fusesoc code](https://github.com/olofk/fusesoc/blob/ec1d6b9c973a200f5186878d5e956bee0dd091a6/fusesoc/config.py#L95) that it expands environment variables
- simply declare a temporary directory: `cache_dir = ctx.actions.declare_directory(".tmp");` which will always work and guarantee a clean cache directory on every run. I *think* that if you don't put the directory in the DefaultInfo, bazel will not store it on disk after running but I haven't actually tried it.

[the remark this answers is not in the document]
olofk: I was dead sure there was a `--cache-root` argument for FuseSoC to set this on the command-line, but apparently there isn't. There should be one however. Would that help this situation? FYI, It doesn't look like we can use env vars in fusesoc.conf

hw/ip/otp_macro/dv/env/otp_macro_env.core
```diff
@@ -9,6 +9,7 @@ filesets:
   files_dv:
     depend:
       - lowrisc:dv:ralgen
+      - lowrisc:dv:cip_lib
```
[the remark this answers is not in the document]
HU90m: :+1: done https://github.com/lowRISC/opentitan/pull/27284

hw/ip/prim_generic/prim_generic.core
```diff
@@ -28,6 +28,7 @@ filesets:
       - lowrisc:prim_generic:usb_diff_rx
       - lowrisc:prim_generic:xnor2
       - lowrisc:prim_generic:xor2
+      - lowrisc:prim_generic:flop_no_rst
```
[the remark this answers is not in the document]
HU90m: I've put the part I could into https://github.com/lowRISC/opentitan/pull/27284

hw/top_darjeeling/dv/chip_sim.core
```diff
@@ -11,8 +11,6 @@ filesets:
       - lowrisc:systems:top_darjeeling_pkg
       - "!gatelevel ? (lowrisc:systems:chip_darjeeling_asic)"
       - lowrisc:ibex:ibex_tracer
-    files:
```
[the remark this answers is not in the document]
HU90m: :+1: https://github.com/lowRISC/opentitan/pull/27284

hw/ip_templates/pwrmgr/pwrmgr_pkg.core.tpl
```diff
@@ -10,6 +10,7 @@ filesets:
     depend:
       - lowrisc:tlul:headers
   % if wait_for_external_reset:
+      - lowrisc:ip:lc_ctrl_pkg
```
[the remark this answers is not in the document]
HU90m: :+1: https://github.com/lowRISC/opentitan/pull/27284

hw/top_englishbreakfast/ip_autogen/flash_ctrl/rtl/flash_ctrl_top_specific_pkg.sv
```diff
@@ -297,7 +297,7 @@ package flash_ctrl_top_specific_pkg;
                  ecc_en:      MuBi4True,
                  he_en:       MuBi4True, // HW assumes high endurance
                  base:        '0,
-                 size:        NumBanks * PagesPerBank
+                 size:        10'(NumBanks * PagesPerBank)
```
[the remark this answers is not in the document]
HU90m: :+1: https://github.com/lowRISC/opentitan/pull/27284

hw/ip/prim_generic/prim_generic_flash.core
```diff
@@ -11,7 +11,7 @@ virtual:
 filesets:
   files_rtl:
     depend:
-      - lowrisc:ip:tlul
+      - lowrisc:tlul:headers
```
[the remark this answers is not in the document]
HU90m: :+1: https://github.com/lowRISC/opentitan/pull/27284

hw/ip_templates/flash_ctrl/rtl/flash_ctrl_top_specific_pkg.sv.tpl
HU90m: :+1: https://github.com/lowRISC/opentitan/pull/27284

rules/fusesoc.bzl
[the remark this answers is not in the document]
HU90m: Thank you @pamaury . Option 2 looks straight forward enough. I'll do that.

A `--cache-root` argument won't fix this problem, although it may be quite a nice convenience feature. This problem was due to `out_dir` being read-only.

hw/top_earlgrey/ip/ast/ast.core
```diff
@@ -18,9 +18,11 @@ filesets:
       - lowrisc:prim:lfsr
       - lowrisc:earlgrey_ip:pinmux_pkg
       - lowrisc:prim:assert
-      - lowrisc:prim:prim_pkg
       - lowrisc:prim:mubi
       - lowrisc:prim:multibit_sync
+      # TODO: prim_pkg is deprecated.
```
andreaskurth, line 23: We should probably create an issue to track this and mention the issue number here, right? (There are more prim_pkg-related TODO comments in other files)

hw/top_earlgrey/ip_autogen/flash_ctrl/flash_ctrl.core
```diff
@@ -19,6 +19,8 @@ filesets:
       - lowrisc:prim:lfsr
       - lowrisc:prim:secded
       - lowrisc:prim:sparse_fsm
+      # TODO: prim_legacy_pkg is deprecated
```
andreaskurth, line 22: And here as well (create an issue and mention its number)?

hw/top_earlgrey/chip_earlgrey_asic.core
```diff
@@ -10,10 +10,14 @@ filesets:
...
       - lowrisc:systems:top_earlgrey_pkg
       - lowrisc:systems:top_earlgrey_padring
+      - lowrisc:earlgrey_ip:flash_ctrl_prim_reg_top
       - "fileset_partner ? (partner:systems:top_earlgrey_ast)"
       - "fileset_partner ? (partner:systems:top_earlgrey_scan_role_pkg)"
       - "!fileset_partner ? (lowrisc:systems:top_earlgrey_ast)"
       - "!fileset_partner ? (lowrisc:earlgrey_systems:scan_role_pkg)"
+      - "!fileset_partner ? (lowrisc:prim:prim_legacy_pkg)"
+      - "!fileset_partner ? (lowrisc:prim_generic:all)"
+      - "!fileset_partner ? (lowrisc:prim_generic:flash)"
```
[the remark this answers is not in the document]
a-will: Personally, I think we should get rid of these flags. The mapping feature should handle replacements more directly, shouldn't it?

hw/ip_templates/rstmgr/rstmgr_reg.core.tpl
```diff
@@ -8,6 +8,7 @@ description: "Reset manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed two lines below.

hw/top_darjeeling/ip_autogen/clkmgr/clkmgr_reg.core
```diff
@@ -8,6 +8,8 @@ description: "Clock manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:ip:tlul
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed 4 lines below.

hw/ip_templates/clkmgr/clkmgr_reg.core.tpl
```diff
@@ -8,6 +8,8 @@ description: "Clock manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:ip:tlul
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed 4 lines below.

hw/top_darjeeling/ip_autogen/rstmgr/rstmgr_reg.core
```diff
@@ -8,6 +8,7 @@ description: "Reset manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed 2 lines below.

hw/top_earlgrey/ip_autogen/clkmgr/clkmgr_reg.core
```diff
@@ -8,6 +8,8 @@ description: "Clock manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:ip:tlul
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed 4 lines below.

hw/top_earlgrey/ip_autogen/rstmgr/rstmgr_reg.core
```diff
@@ -8,6 +8,7 @@ description: "Reset manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed 2 lines below.

hw/top_englishbreakfast/ip_autogen/clkmgr/clkmgr_reg.core
```diff
@@ -8,6 +8,8 @@ description: "Clock manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:ip:tlul
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed 4 lines below.

hw/top_englishbreakfast/ip_autogen/rstmgr/rstmgr_reg.core
```diff
@@ -8,6 +8,7 @@ description: "Reset manager registers"
 filesets:
   files_rtl:
     depend:
+      - lowrisc:prim:subreg
```
marnovandermaas: This one is listed 2 lines below.

hw/dv/sv/sec_cm/prim_sparse_fsm_flop_if.sv
marnovandermaas: Or reorder the commits so that this comes before the change to virtual cores.

hw/ip/prim_xilinx/rtl/prim_ram_1p.sv
marnovandermaas, line 1: Is the content of this block now copied from the generic ram 1p module? Mainly asking because it's nice to avoid code duplication.

hw/ip/prim_generic/prim_generic.core
```diff
@@ -32,6 +32,30 @@ filesets:
...
+  "lowrisc:prim:pad_attr"     : "lowrisc:prim_generic:pad_attr"
+  "lowrisc:prim:pad_wrapper"  : "lowrisc:prim_generic:pad_wrapper"
+  "lowrisc:prim:ram_1p"       : "lowrisc:prim_generic:ram_1p"
+  "lowrisc:prim:ram_1r1w"     : "lowrisc:prim_generic:ram_1r1w"
+  "lowrisc:prim:ram_2p"       : "lowrisc:prim_generic:ram_2p"
+  "lowrisc:prim:rom"          : "lowrisc:prim_generic:rom"
+  "lowrisc:prim:usb_diff_rx"  : "lowrisc:prim_generic:usb_diff_rx"
+  "lowrisc:prim:xnor2"        : "lowrisc:prim_generic:xnor2"
+  "lowrisc:prim:xor2"         : "lowrisc:prim_generic:xor2"
+  # Flash is a good canditate to be removed from the prims and become a macro like OTP.
```
marnovandermaas, line 57: Is there an issue tracking this work? If so can we include that issue number here?

hw/top_englishbreakfast/top_englishbreakfast.core
```diff
@@ -9,6 +9,8 @@ filesets:
     depend:
       - lowrisc:englishbreakfast_ip:pwrmgr_pkg
       - lowrisc:earlgrey_ip:alert_handler_pkg
+      - lowrisc:systems:top_englishbreakfast_ast_pkg
+      - lowrisc:earlgrey_constants:top_racl_pkg
```
marnovandermaas: Is there no English breakfast RACL?

hw/lint/tools/dvsim/common_lint_cfg.hjson
```diff
@@ -26,6 +26,7 @@
                "--target={flow}",
                "--tool={tool}",
                "--work-root={build_dir}/fusesoc-work",
+               "--mapping=lowrisc:prim_generic:all:0.1",
```
marnovandermaas, line 29: This can probably be squashed with the other mapping commit right?

hw/lint/tools/dvsim/common_lint_cfg.hjson
```diff
@@ -27,7 +27,7 @@
                "--tool={tool}",
                "--work-root={build_dir}/fusesoc-work",
                "--mapping=lowrisc:prim_generic:all:0.1",
-               "--mapping=lowrisc:systems:top_earlgrey:0.1",
+               "{additional_fusesoc_argument}",
```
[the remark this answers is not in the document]
marnovandermaas: Also probably unnecessary to change to top earl grey before going to the argument version.

andreaskurth:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_pkg.sv

^ Renamed from `hw/ip/prim/util/primgen/prim_pkg.sv.tpl` with effectively the same content.

CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_and2.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_buf.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_clock_buf.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_clock_div.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_clock_gating.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_clock_inv.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_clock_mux2.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_flash.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_flop.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_flop_2sync.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_flop_en.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_pad_attr.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_pad_wrapper.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_ram_1p.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_ram_1r1w.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_ram_2p.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_rom.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_usb_diff_rx.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_xnor2.sv
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_xor2.sv

^ Modules and files renamed to remove the `generic_` infix.

CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_generic_flash_bank.sv

^ Just a typo fix -- the module itself still (correctly) carries `generic_` in its name because it's internal to the `prim_generic` library rather than an implementation of a virtual core.

CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_and2.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_buf.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_clock_buf.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_clock_gating.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_clock_mux2.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_flop.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_flop_en.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_pad_attr.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_pad_wrapper.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_xnor2.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_xor2.sv

^ Modules and files renamed to remove the `generic_` infix.

hw/ip/prim_xilinx/rtl/prim_xilinx_rom.sv --> I think **this file still needs to be renamed to `prim_rom.sv`.**

CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_ram_1p.sv

^ Module and file renamed to remove the `generic_` infix, _and_ implementation code from the generic `prim_ram_1p` inserted.

CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_clock_buf.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_clock_div.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_clock_gating.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_clock_inv.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_pad_wrapper.sv

^ Modules and files renamed to remove the `generic_` infix.

hw/ip/prim_xilinx_ultrascale/rtl/prim_xor2.sv --> I think **this file still needs to be removed.**

CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_and2.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_buf.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_clock_mux2.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_flop.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_flop_en.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_pad_attr.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_ram_1p.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_xnor2.sv

^ Files removed because they are now implemented by the `prim_xilinx` (i.e., non-ultrascale) library.

CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/flash_ctrl/data/flash_ctrl.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/flash_ctrl/rtl/flash_ctrl.sv

^ `.gen_generic.u_impl_generic` removed from paths in code

CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/flash_ctrl/rtl/flash_ctrl_top_specific_pkg.sv

^ Added literal size cast to fix linting

hw/top_earlgrey/chip_earlgrey_asic.core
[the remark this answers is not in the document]
andreaskurth: Ah, so downstream integrators would invoke fusesoc with `--mapping partner:prim:all`? Do we already have everything in place to integrate this command-line argument into OT tooling (such as dvsim)? If not, I'd strongly suggest that in this PR we go with something that works *now*, and then improve as necessary in follow-up PRs. We really should get this merged soon.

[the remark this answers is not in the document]
HU90m: Will do Andreas' suggestion, but do agree that mappings are probably a nicer solution long term.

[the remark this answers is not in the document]
a-will: > Do we already have everything in place to integrate this command-line argument into OT tooling (such as dvsim)?

We should, or this PR wouldn't work. 😄 As far as I know, there's just fusesoc and dvsim's use of fusesoc. For dvsim, we currently use the `design_level` parameter as a terrible mechanism to select partner filesets. I haven't looked at the updates lately, but this PR should have a similar parameter for mappings that can be overridden or appended to.

The change should certainly be in a separate commit, and if you'd like to have it in another PR, that's totally fine with me. I'd recommend not leaving the confusing legacy stuff for too long, though, so there's just the one transition to work out with existing partner code bases (instead of two transitions). It's probably even better to get their opinion on that front, hehe.

HU90m:
I believe that's now all the review comments addressed. It only failed the lint (slow) last run due to a forgotten template file. That's now fixed so it should be ready to merge barring any further reviews.

*Not Verible will fail due to a review dog limitation. It passes locally.*

hw/BUILD
```diff
@@ -75,7 +75,7 @@ fusesoc_build(
     data = ["//hw/ip/otbn:rtl_files"],
     make_options = ":make_options",
     output_groups = {
-        "binary": ["lowrisc_dv_chip_verilator_sim_0.1/sim-verilator/Vchip_sim_tb"],
+        "binary": ["lowrisc_dv_top_earlgrey_chip_verilator_sim_0.1/sim-verilator/Vchip_sim_tb"],
```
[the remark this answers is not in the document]
a-will: This PR probably needed to be closed and reopened--There's so much discussion here that it's easy to miss the explanations, haha.

In any case, this changed because newer fusesoc changed the paths.

hw/dv/tools/dvsim/common_sim_cfg.hjson
```diff
@@ -13,6 +13,9 @@
                      "{dv_root}/tools/dvsim/bazel.hjson",
                      "{dv_root}/tools/dvsim/{tool}.hjson"]
 
+  sv_flist_gen_flags: ["--flag=fileset_{design_level}",
+                       "--mapping=lowrisc:prim_generic:all:0.1"]
+
```
a-will, line 18: If we're going to set a default mapping here, we should document how to override `sv_flist_gen_flags` + what is ordinarily set here. I feel iffy about putting prim selection in a generically-named property like this, though. It probably should be called out more specifically.

For dvsim, I think we should document the base configurations and explicitly require fusesoc as a dependency. If there were any hope of somehow abstracting away that dependency, that is gone now, hehe. Downstream user configuration files will explicitly have fusesoc-specific parameters in them.

This can all be done in future PRs.

hw/ip/prim_generic/prim_generic.core
a-will: Issue to include: https://github.com/lowRISC/opentitan/issues/27042

hw/top_darjeeling/lint/top_darjeeling_lint_cfgs.hjson
```diff
@@ -24,16 +24,19 @@
   use_cfgs: [{    name: mbx
                   fusesoc_core: lowrisc:ip:mbx
                   import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+                  additional_fusesoc_argument: "--mapping=lowrisc:systems:top_darjeeling:0.1"
```
a-will, line 27: `additional_fusesoc_argument` sounds a bit hokey. In a future PR, we should probably get more specific about prims and make the property names more obviously related to the intended use.

hw/top_darjeeling/physical_pads.core
```diff
@@ -24,8 +24,8 @@ filesets:
       # common waivers
       - lowrisc:lint:common
       - lowrisc:lint:comportable
-    files:
-      # - lint/physical_pads.waiver
+    # files:
+    #   - lint/physical_pads.waiver
```
a-will, line 28: nit: Should we simply remove these lines instead of commenting them out?

rules/fusesoc.bzl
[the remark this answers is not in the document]
a-will: We should create a TODO out of this and implement in a future PR. 😄

hw/ip/otp_ctrl/otp_ctrl.core
a-will: https://github.com/lowRISC/opentitan/issues/27347

hw/top_earlgrey/chip_earlgrey_asic.core
a-will: https://github.com/lowRISC/opentitan/issues/27347

hw/top_earlgrey/ip/ast/top_earlgrey_ast_top.core
```diff
@@ -13,6 +13,8 @@ filesets:
       - lowrisc:earlgrey_ip:rstmgr_pkg
       - lowrisc:earlgrey_ip:pinmux_pkg
       - lowrisc:systems:top_earlgrey_ast
+      - lowrisc:prim_generic:all
```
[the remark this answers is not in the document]
a-will: Eliminating this core looks to still be a TODO. For a future PR.

hw/ip/prim_xilinx/rtl/prim_ram_1p.sv
a-will: Unfortunately, yes. This is a consequence of virtual cores conflicting with each other. We'd need to change the hierarchy yet again if we wanted to create some module that could be instantiated (or else use a generator to change the module name).

We can work out deduplication in another PR, though. 😄

hw/top_englishbreakfast/top_englishbreakfast.core
a-will: There actually is such a package. This needs fixing.

hw/ip_templates/flash_ctrl/dv/env/flash_ctrl_env.core
a-will: File was removed.

hw/lint/tools/dvsim/common_lint_cfg.hjson
```diff
@@ -27,6 +27,7 @@
                "--tool={tool}",
                "--work-root={build_dir}/fusesoc-work",
                "--mapping=lowrisc:prim_generic:all:0.1",
+               "--mapping=lowrisc:systems:top_earlgrey:0.1",
```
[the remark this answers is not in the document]
a-will: Removed.

a-will:
It looks like comments have been addressed, either by making the change here or deferring to a future PR. 

Tests were reported passing, and my changes since then have almost entirely been comments. The only new bit was creating a core for top_englishbreakfast_racl_pkg and using it in the englishbreakfast systems (instead of earlgrey's). The CW305 build passes.

The failure for Darjeeling DV is a temporary network failure and is unrelated to this PR.

Merging now!

olofk:
Woohoo! Never thought I'd live to see the day :P Great work, everyone involved!

Now as for the FuseSoC-dvsim integration...

hw/top_earlgrey/chip_earlgrey_asic.core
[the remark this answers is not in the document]
hcallahan-lowrisc: > Personally, I think we should get rid of these flags. The mapping feature should handle replacements more directly, shouldn't it?
> 
> Now that the mapping feature exists, I feel like we should consolidate on that one mechanism to handle selection of prims (unless there is a good reason it can't work). Consistency will help simplify documentation and downstream integration.

Some more cleanup for the old way is here : https://github.com/lowRISC/opentitan/pull/27486

Issue #19789, opened 2023-09-29
[edn] FIFO error does not cause local escalation


h-filali:
### Description

Currently FIFO errors do not cause local escalation as can be seen in this snippet from the edn_core.sv:

https://github.com/lowRISC/opentitan/blob/e13ff43f1b951c00ea2e24cafa60f7ac11d04e01/hw/ip/edn/rtl/edn_core.sv?plain=1#L295-L307

Other errors like counter or state machine errors cause the SM to enter an error state, while after a FIFO error the SM continues its operation. 

According to secure hardware design guideline [5.2](https://opentitan.org/book/doc/security/implementation_guidelines/hardware/index.html#recommendation-5-alerts) we should "prefer to take a local action (clearing/randomizing state, cease processing) in addition to generating the alert".

There could very well be a reason for not escalating locally but I couldn't find any reason in the documentation or in the lines of code above.

Should this be changed in the design (local escalation on FIFO error) or documented why *no* local escalation is preferable in this case?

vogelpi:
Good catch @h-filali ! I think this is probably a bug. Unless if these errors can happen when disabling EDN which might trigger some FIFO operations? I think this should probably be fixed but I don't know if it's wise to fix this for PROD. We should probably investigate first what can trigger these FIFO errors.

vogelpi:
We've just discussed this in person and concluded that it doesn't make sense to not factor in those FIFO errors into the fatal alert signal and local escalation. @h-filali also confirmed to not have seen those FIFO errors occur during normal operation. I thus suggest to change the RTL accordingly once we can again accept RTL changes.

vogelpi:
@h-filali , I think we forgot about this one. I had a look at the RTL and I think we still don't lock up the FSM upon FIFO errors.

Maybe we should also check how we handle this in ENTROPY_SRC and CSRNG, and then align all three blocks. I've now put into M4.

vogelpi:
As for the other two blocks of the entropy complex:

- CSRNG: FIFO errors (no hardened primitives as all FIFOs have a depth of 1, i.e. read and write pointers are fixed - hardening would not have an effect) don't lead to local escalation. The only local escalation source are various hardened counters.
- ENTROPY_SRC: FIFO errors (also internal errors for hardened FIFOs) don't feed into a local escalation signal. The only local escalation source is a hardened counter.

[...]

PR #23539, opened 2024-06-06
[csrng] Make reseed counters always readable, allow locking down internal state access per instance

vogelpi:
As discussed here: https://github.com/lowRISC/opentitan/issues/21141#issuecomment-2082289550

PR #23560, opened 2024-06-07
[entropy_src] Locally escalate upon internal FIFO errors and FSM errors

vogelpi:
Previously, these errors would trigger a fatal alert but not cause a local escalation (i.e. lock up the main SM). Since all these errors are structural errors that can't occur e.g. due to how the module currently operates but instead due to FI attacks only, these errors should also cause a local escalation.

This change better aligns ENTROPY_SRC with EDN and CSRNG.

This resolves lowRISC/OpenTitan#19789.


moidx (approved):

vogelpi:
I am merging this. `spi_passthru_test_fpga_cw310_rom_with_fake_keys` keeps failing but I've also seen this failing for other PRs e.g. #23557 and #23539.

PR #23561, opened 2024-06-07
[csrng] Locally escalate upon FSM errors

vogelpi:
Previously, these errors would trigger a fatal alert but not cause a local escalation (i.e. lock up the main SM). Since these FSM errors are structural errors that can't occur e.g. due to how the module currently operates but instead due to FI attacks only, these errors should also cause a local escalation.

This change better aligns CSRNG with EDN and ENTROPY_SRC.

This is related to lowRISC/OpenTitan#19789.

### DIFF
--- a/hw/ip/csrng/rtl/csrng_core.sv
+++ b/hw/ip/csrng/rtl/csrng_core.sv
@@ -490,6 +490,17 @@ module csrng_core import csrng_pkg::*; #(
     .intr_o                 (intr_cs_fatal_err_o)
   );
 
+  // Counter and FSM errors are structural errors and are always active regardless of the
+  // functional state. main_sm_err_sum is not included here to prevent some tools from
+  // inferring combo loops.
+  logic fatal_loc_events;
+  assign fatal_loc_events = cmd_gen_cnt_err_sum ||
+                            cmd_stage_sm_err_sum ||
+                            drbg_gen_sm_err_sum ||
+                            drbg_updbe_sm_err_sum ||
+                            drbg_updob_sm_err_sum ||
+                            aes_cipher_sm_err_sum;
+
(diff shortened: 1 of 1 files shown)


moidx (approved):
LGTM.

vogelpi:
FPGA bitstream generation fails because Vivado infers a combo loop:
```
ERROR: [DRC LUTLP-1] Combinatorial Loop Alert: 3 LUT cells form a combinatorial loop. This can create a race condition. Timing analysis may not be accurate. The preferred resolution is to modify the design to remove combinatorial logic loops. If the loop is known and understood, this DRC can be bypassed by acknowledging the condition and setting the following XDC constraint on any one of the nets in the loop: 'set_property ALLOW_COMBINATORIAL_LOOPS TRUE [get_nets <myHier/myNet>]'. One net in the loop is top_earlgrey/u_csrng/u_csrng_core/u_csrng_main_sm/u_state_regs/u_state_flop/cs_main_sm_err. Please evaluate your design. The cells in the loop are: top_earlgrey/u_csrng/u_reg/u_err_code_test/g_intr_event.new_event_i_10__4, top_earlgrey/u_csrng/u_reg/u_err_code_test/q[0]_i_2__385, and top_earlgrey/u_csrng/u_csrng_core/u_csrng_main_sm/u_state_regs/u_state_flop/q[0]_i_3__199.
```
We shouldn't merge this before this is fixed.

vogelpi:
I've now removed the `main_sm_err_sum` signal from the `fatal_loc_events` fed back into the main SM. This should hopefully resolve the FPGA issue.

vogelpi:
spi_passthru_test_fpga_cw310_rom_with_fake_keys fails but it's a known issue not related to this PR.

PR #23566, opened 2024-06-07
[otbn,dv] Add missing include to otbn_trace_entry.h

rswarbrick:
This header uses uint32_t and builds with recent versions of GCC and Verilator don't get that included transitively, causing a build failure. Fix the silly bug.


vogelpi (approved):
LGTM, there are currently some CI failures and we should ensure CI passes before merging this.

rswarbrick:
I've just checked and the CI test failures are:
- CW310 ROM tests (at run time)
- A countermeasure check (build time) in Python for Englishbreakfast.
Neither of these failures could be caused by this change, which applies to C++ code that's used to build DV tooling (and is just adding a missing include file!)

PR #23571, opened 2024-06-07
[sw/host] Fix UART usage regarding control flow

pamaury:
These rust harnesses enable control flow on a UART object but then
immediately drop it and gets it again in each test function. Since
the UART object is reference counted, it will be destroyed in-between
tests and the control flow setting will be lost. The easiest fix is to
create the object once and pass it around.

### DIFF
--- a/sw/host/tests/chip/flash_ctrl/src/main.rs
+++ b/sw/host/tests/chip/flash_ctrl/src/main.rs
@@ -56,6 +56,7 @@ fn test_update_phase(
 
 fn test_end(opts: &Opts, end_test_address: u32, transport: &TransportWrapper) -> Result<()> {
     let uart = transport.uart("console")?;
+    uart.set_flow_control(true)?;
     let end_test_value = MemRead32Req::execute(&*uart, end_test_address)?;
     assert!(end_test_value == 0);
     MemWrite32Req::execute(&*uart, end_test_address, /*value=*/ 1)?;
--- a/sw/host/tests/chip/gpio/src/main.rs
+++ b/sw/host/tests/chip/gpio/src/main.rs
@@ -297,8 +297,7 @@ fn read_all_verify(
     Ok(())
 }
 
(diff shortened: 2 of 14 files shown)


jwnrt (approved):
Thanks for the fix

pamaury:
I have now added fixes for several harnesses which I think also suffer from this problem (in theory).

mundaym:
@cfrantz @moidx this looks like a promising fix for some of the flakiness issues we've been seeing on the FPGA runners. Might be worth merging today to see if it helps?

moidx (approved):

moidx:
Thanks for the fix @pamaury!

cfrantz:
> These rust harnesses enable control flow on a UART object but then immediately drop it and gets it again in each test function. Since the UART object is reference counted, it will be destroyed in-between tests and the control flow setting will be lost. The easiest fix is to create the object once and pass it around.

This is incorrect.  The UART object returned by the transport object is refcounted and _is held for the duration of the program_ by the transport object.  Once a hardware interface is created by opentitanlib (be it uart, spi, i2c, etc), that instance of the interface lives until the program ends.

This can be shown by adding a logging statement such as the following to various places in the test program
```
log::info!("uart = {:x?}:{}", Rc::as_ptr(&uart), Rc::strong_count(&uart));
```
Furthermore, one can show that the flow control setting doesn't change between the test functions by adding a `get_flow_control` method to the uart trait.

You will observe something like the following:
```
# Initial create in `main`
[2024-06-10T16:13:29.842Z INFO  spi_passthru] uart = 0x564017948740:2 fc=1

# Subsequent access to the uart object in the test functions.
[2024-06-10T16:13:31.463Z INFO  spi_passthru] uart = 0x564017948740:3 fc=1
[2024-06-10T16:13:31.489Z INFO  spi_passthru] uart = 0x564017948740:3 fc=1
```

You'll observe the initial refcount is 2: one Rc in the transport and one in `main`.  Subsequent refcounts are 3: transport, main and test function.  You'll also observe that the flow_control setting remains set for each access - indeed, they're all the same object.

PR #23574, opened 2024-06-07
[chip-test/sival] Modify config for csrng_edn_concurrency_reduced_freq

vogelpi:
We've recently observed some timeouts for this test in the nighlies. This commit thus speeds up the AST power-up time to reduce the test time per seed similar to the non-reduced frequency variant.


moidx (approved):

vogelpi:
I am merging this such that it is picked up by the nightly run. CI isn't through yet but this PR only changes the run argument of a single test not run in CI.

PR #23379, opened 2024-05-29
[opentitantool] Updated HyperDebug firmware

jesultra:
Updated HyperDebug firmware with support for 1.5Mbaud sustained UART throughput (on a single UART at a time), and a bugfix for JTAG on non-default pins.

PR #23576, opened 2024-06-07
[opentitantool] Arbitrary waveform generation via HyperDebug

jesultra:
HyperDebug has fast DACs, meaning that we can have them generate arbitrary waveforms on the analog output pins.

This is useful for verifying that the OpenTitan ADC filters have been configured as intended, by simulating e.g. USB-PD negotiation going on "on top of" the steady state CC voltage levels.

This feature builds on the existing support for streaming binary data for bit-banging, and handling it on HyperDebug via a timer interrupt.

Current HyperDebug firmware supports short bursts of up to 2000 samples at up to 1Msps, or streamed generation of arbitrary lengths at up to around 80ksps.  (If necessary, HyperDebug firmware could be modified to allocate larger buffers, such that more than 2000 samples could be emitted at the maximum speed.)

Example commands below.

Fast USB-PD-like communication "on top of" steady level of 0.25 volt (argument needs to be a lot longer for realistic packet, can be generated programmatically if using opentitanlib):
```
opentitantool --exec="gpio set CN7_9 --mode AnalogOutput --voltage 0.25" \
    gpio dac-bang --clock="600kHz" CN7_9 -s "0.5 0.0 0.5 0.5 0.0 0.5 0.0 0.0 0.5 0.0 0.5 0.25"
```

Gradual voltage drop from 0.25V down to zero in the course of 300 microseconds. (Will be implemented as steps at the given frequency, 180 steps in this example. If the waveform exceeds 2000 samples, HyperDebug may not be able to keep up):
```
opentitantool --exec="gpio set CN7_9 --mode AnalogOutput --voltage 0.25" \
    gpio dac-bang --clock="600kHz" CN7_9 -s "0.25 linear(300us) 0.0"
```

Pauses between bursts (this uses efficient encoding of pauses, so that they do not count towards the 2000 samples):
```
opentitantool --exec="gpio set CN7_9 --mode AnalogOutput --voltage 0.25" \
    gpio dac-bang --clock="600kHz" CN7_9 -s "0.5 0.0 0.5 0.25  100ms   0.5 0.0 0.5 0.25"
```

### DIFF
--- a/sw/host/opentitanlib/src/io/gpio.rs
+++ b/sw/host/opentitanlib/src/io/gpio.rs
@@ -44,6 +44,10 @@ pub enum GpioError {
     InvalidBitbangData(usize),
     #[error("Bitbang delay of zero, immediately preceding `await()`, or at end of sequence, not permitted")]
     InvalidBitbangDelay,
+    #[error("Dac-bang samples not a multiple of number of pins")]
+    InvalidDacBangData,
+    #[error("Dac-bang delay of zero or immediately adjacent to linear(), or at end of sequence, not permitted")]
+    InvalidDacBangDelay,
     #[error("Generic error: {0}")]
     Generic(String),
 }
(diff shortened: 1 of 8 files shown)


timothytrippel (approved):

jesultra:
With cherry pick PR#23398 now merged, this one might cleanly apply.

jesultra:
Oops, there was one more missing backport, #23379, before this one can apply.  Let me look into it...

PR #23579, opened 2024-06-08
[spi_host] Fix handling of CPHA=1 and IdleCSBActive

a-will:
Promote CPHA=1 to have its own fully-fledged logic, instead of
incorrectly chaining it to CPHA=0 cases. Leave only the timing of the
read enable with the simple delayed control style.

Also, ensure the state after IdleCSBActive still counts out a full
clkdiv so setup times are met. Previously, the launching state might
only last a single cycle.

Ensure that the data outputs and output-enables are held for the same
value continuously until the next launch edge. Use assertions to verify.

The commit introducing the assertions can be removed. It's merely to show it as an example.

### DIFF
--- a/hw/ip/spi_host/rtl/spi_host_fsm.sv
+++ b/hw/ip/spi_host/rtl/spi_host_fsm.sv
@@ -41,8 +41,6 @@ module spi_host_fsm
   logic [15:0]      clk_cntr_q, clk_cntr_d;
   logic             clk_cntr_en;
 
-  logic [1:0]       speed_cpha0, speed_cpha1;
-
   logic [CSW-1:0]   csid;
   logic [CSW-1:0]   csid_q;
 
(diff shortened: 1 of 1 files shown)


antmarzam:
Thank you Alex for putting a fix together so quickly :D 

I just regressed this fix in Xcelium and VCS. 
The results for xcelium :
```
          [   legend    ]: [Q: queued, D: dispatched, P: passed, F: failed, K: killed, T: total]                                       
00:00:23  [    build    ]: [Q: 000, D: 000, P: 001, F: 000, K: 000, T: 001] 100%               
00:42:52  [     run     ]: [Q: 000, D: 000, P: 373, F: 002, K: 000, T: 375] 100% 
```
The two failing tests are a recurring failure in the spi_host regression with failing signature:
```
UVM_FATAL @ 91698009931 ps: (csr_utils_pkg.sv:577) [csr_utils::csr_spinwait] timeout spi_host_reg_block.status.active (addr=0xac858f14) == 0x0

```

The results in VCS:
```
          [   legend    ]: [Q: queued, D: dispatched, P: passed, F: failed, K: killed, T: total]         
00:00:36  [    build    ]: [Q: 000, D: 000, P: 001, F: 000, K: 000, T: 001] 100%                         
00:46:15  [     run     ]: [Q: 000, D: 000, P: 358, F: 017, K: 000, T: 375] 100%            
```

From the VCS failures, one of them is the same as the ongoing spi_host regression failure in the nightly regression:
```
UVM_FATAL @ 25061256407 ps: (csr_utils_pkg.sv:576) [csr_utils::csr_spinwait] timeout spi_
host_reg_block.status.active (addr=0xe689af54) == 0x0
```
The other failures are related to the passthrough test, which is consistently failing for VCS with signature:
```
 UVM_FATAL @   4614591 ps: (spi_if.sv:59) [spi_if] Don't call this function - get_active_c
sb when there is no active CSB

```

With the numbers above, I'm confident this fix is working as intended. 

I'll file a ticket to fix the DV related passthrough failure in VCS

antmarzam (approved):
LGTM, plus the regression numbers back the change up as working well

rswarbrick (approved):
I think this makes sense to me. Thanks for the careful explanation.

a-will:
Dropped the second commit with the assertion per request. Merging, since RTL is the same.

PR #23580, opened 2024-06-08
backport-23511-to-earlgrey_es_sival

engdoreis:
This is a manual backport of https://github.com/lowRISC/opentitan/pull/23511.
There was a conflict on the file `sw/host/tests/chip/spi_device/BUILD`.

### DIFF
--- a/sw/device/tests/spi_device_sleep_test.c
+++ b/sw/device/tests/spi_device_sleep_test.c
@@ -1,0 +1,157 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/device/lib/base/mmio.h"
+#include "sw/device/lib/dif/dif_aon_timer.h"
+#include "sw/device/lib/dif/dif_pinmux.h"
+#include "sw/device/lib/dif/dif_pwrmgr.h"
+#include "sw/device/lib/dif/dif_rv_core_ibex.h"
+#include "sw/device/lib/dif/dif_rv_plic.h"
+#include "sw/device/lib/runtime/irq.h"
+#include "sw/device/lib/runtime/log.h"
+#include "sw/device/lib/testing/aon_timer_testutils.h"
+#include "sw/device/lib/testing/pwrmgr_testutils.h"
+#include "sw/device/lib/testing/rv_plic_testutils.h"
+#include "sw/device/lib/testing/spi_device_testutils.h"
+#include "sw/device/lib/testing/spi_flash_emulator.h"
+#include "sw/device/lib/testing/spi_flash_testutils.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+#include "sw/device/lib/testing/autogen/isr_testutils.h"
+
+OTTF_DEFINE_TEST_CONFIG();
+
+enum {
+  kWakeupTimeMicros = 500000,
+};
+static dif_spi_device_handle_t spid;
+static dif_pwrmgr_t pwrmgr;
+static dif_aon_timer_t aon_timer;
+static dif_rv_plic_t plic;
+static plic_isr_ctx_t plic_ctx = {.rv_plic = &plic,
+                                  .hart_id = kTopEarlgreyPlicTargetIbex0};
+static pwrmgr_isr_ctx_t pwrmgr_isr_ctx = {
+    .pwrmgr = &pwrmgr,
+    .plic_pwrmgr_start_irq_id = kTopEarlgreyPlicIrqIdPwrmgrAonWakeup,
+    .expected_irq = kDifPwrmgrIrqWakeup,
+    .is_only_irq = true};
+
+static void enter_low_power(void) {
+  uint64_t wkup_cycles = 0;
+  CHECK_STATUS_OK(aon_timer_testutils_get_aon_cycles_64_from_us(
+      kWakeupTimeMicros, &wkup_cycles));
+  CHECK_STATUS_OK(aon_timer_testutils_wakeup_config(&aon_timer, wkup_cycles));
+
+  dif_pwrmgr_domain_config_t pwrmgr_domain_cfg =
+      kDifPwrmgrDomainOptionMainPowerInLowPower;
+
+  LOG_INFO("SYNC: Entering lowpower mode");
+  CHECK_STATUS_OK(pwrmgr_testutils_enable_low_power(
+      &pwrmgr, kDifPwrmgrWakeupRequestSourceFive, pwrmgr_domain_cfg));
+  wait_for_interrupt();
+  LOG_INFO("SYNC: Wakening up");
+}
+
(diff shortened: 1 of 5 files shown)


sw/device/tests/spi_device_sleep_test.c
```diff
@@ -0,0 +1,157 @@
...
+    .pwrmgr = &pwrmgr,
+    .plic_pwrmgr_start_irq_id = kTopEarlgreyPlicIrqIdPwrmgrAonWakeup,
+    .expected_irq = kDifPwrmgrIrqWakeup,
+    .is_only_irq = true};
+
+static void enter_low_power(void) {
+  uint64_t wkup_cycles = 0;
+  CHECK_STATUS_OK(aon_timer_testutils_get_aon_cycles_64_from_us(
+      kWakeupTimeMicros, &wkup_cycles));
+  CHECK_STATUS_OK(aon_timer_testutils_wakeup_config(&aon_timer, wkup_cycles));
```
a-will, line 46: What role does aon_timer play here? It is merely to provide the wakeup trigger?

This looks like a race with the host, since it defines arbitrary time bounds for the tester to perform its work. For this case, it's probably plenty long, but... Would it be a more reliable / pertinent test if a pinmux-based wakeup were used here? Either a GPIO to advance the test or even a _slow_ SPI transaction, with CSB used to trigger it, could be used.

If CSB is used, the pulse width of CSB would need to be substantially longer than one AON clock period, so the SPI transaction's clock rate should probably be limited to something like 1 MHz. That'd be around 16 us for a ReadStatus1 or ReadJedec command, enough for 3 AON clock cycles.

sw/device/tests/spi_device_sleep_test.c
```diff
@@ -0,0 +1,157 @@
...
+
+  dif_pwrmgr_domain_config_t pwrmgr_domain_cfg =
+      kDifPwrmgrDomainOptionMainPowerInLowPower;
+
+  LOG_INFO("SYNC: Entering lowpower mode");
+  CHECK_STATUS_OK(pwrmgr_testutils_enable_low_power(
+      &pwrmgr, kDifPwrmgrWakeupRequestSourceFive, pwrmgr_domain_cfg));
+  wait_for_interrupt();
```
a-will, line 54: There's a bit of a race with the host here, since low power mode hasn't been enabled yet. It takes dozens of microseconds to set up the transition to low power mode, and the host could have processed that UART message and issued a SPI transaction before we ever reach WFI.

Do we need to have completed the transition to low power for this test? Or is it okay that we might be hitting spi_device while pwrmgr is still transitioning? The pinmux will take on the sleep values almost immediately after pwrmgr recognizes the WFI, so testing outputs during the transition should be fine, I think.

sw/device/tests/spi_device_sleep_test.c
```diff
@@ -0,0 +1,157 @@
...
+    .is_only_irq = true};
+
+static void enter_low_power(void) {
+  uint64_t wkup_cycles = 0;
+  CHECK_STATUS_OK(aon_timer_testutils_get_aon_cycles_64_from_us(
+      kWakeupTimeMicros, &wkup_cycles));
+  CHECK_STATUS_OK(aon_timer_testutils_wakeup_config(&aon_timer, wkup_cycles));
+
+  dif_pwrmgr_domain_config_t pwrmgr_domain_cfg =
+      kDifPwrmgrDomainOptionMainPowerInLowPower;
```
a-will, line 49: Should we be testing this for deep sleep?

@jesultra @jettr Would you have some idea of whether spi_device's responses in deep sleep are relevant?

jettr: We do not have a requirement for the spi_device to respond during deep sleep. We do have a requirement that activity on the SPI device should wake the device up from normal sleep, but it is not an expectation to capture the incoming SPI transaction while waking -- the spi host will resend the data packet if needed.

sw/device/tests/spi_device_sleep_test.c
engdoreis: I'm going to add another commit with the fixes you suggested making it easier to backport to master.

sw/device/tests/spi_device_sleep_test.c
engdoreis: I think it's ok for this test to issue the transaction before the core is sleeping. But if I want to test with core in sleep, what would be the solution? A delay on the host side?

a-will: You could make use of a GPIO indicator. You'd have some default value indicating the core is active, then use a deep sleep function in the pinmux to drive it the other direction.

For example, the host could pull up on a high-z MIO, then the pinmux's deep sleep mode for that MIO could be set to Tie-Low. Then the host can poll that MIO and wait to see 0.

engdoreis: Would that also work for normal sleep?

a-will: > Would that also work for normal sleep?

Oh, judging by the connectivity and RTL, yes. I think I must've somehow connected the pinmux CSRs with deep sleep, but that doesn't appear to be quite right. They are just "sleep modes" and pwrmgr gives the same "low_power_o" output for normal sleep and deep sleep.

sw/device/tests/spi_device_sleep_test.c
engdoreis: Done

sw/device/tests/spi_device_sleep_test.c
engdoreis: Done. I'm waking up the core with a gpio and checking a gpio to make sure it entered low power mode as suggested.

sw/host/tests/chip/spi_device/src/spi_device_sleep_test.rs
```diff
@@ -0,0 +1,112 @@
...
+    sleep_indicator.set_pull_mode(PullMode::PullUp)?;
+    let uart = transport.uart("console")?;
+    let spi = transport.spi(&opts.spi)?;
+    spi.set_max_speed(100_000)?;
+
+    wait_sleep(&*sleep_indicator, &*uart, opts.timeout)?;
+    println!("Reading jedec id");
+    // Should wakup when the cs goes low.
+    let jdec_id = SpiFlash::read_jedec_id(&*spi, 5)?;
+    assert_eq!(&jdec_id, &[0x17u8, 0x17, 0x74, 0x98, 0x22]);
```
a-will, line 73: This transaction is probably failing because it sets too unreasonable a burden on OT to wake up and clear sleep mode before the read data must start being returned.

A transaction that is used to wake up the chip would need to be discarded. If you just want CSB to wake the chip up, then you can probably do something like...

```rust
    // Wake up chip with CSB. Hold asserted until the chip releases sleep mode.
    let csb_hold = spi.assert_cs()?;
    wait_awake(); // Wait for IOA7 to return to high.
    drop(csb_hold);

    // Check that transactions now work.
    let jdec_id = SpiFlash::read_jedec_id(&*spi, 5)?;
    assert_eq!(&jdec_id, &[0x17u8, 0x17, 0x74, 0x98, 0x22]);
```

engdoreis: Should be fixed now by the configuration of the spi_device.

a-will (approved):
LGTM. Thanks for making this test more robust!

PR #23583, opened 2024-06-08
[otbn,dv] Try multiple times in otbn_ctrl_redun_vseq

rswarbrick:
This handles a situation when we don't actually find a time to inject an error (in inject_redun_err). Now we try a couple more times.


vogelpi (approved):
Nice work @rswarbrick ! I think this approach would also help for other blocks where we sometimes miss injecting errors.

PR #23587, opened 2024-06-08
[otbn,rtl] Avoid locking on corrupt RMA signal

### LINKED ISSUE #23585: [otbn,rtl] Incorrect handling of bogus values for rma request value
### Description

This is visible by running `otbn_escalate` with seed 2:
```
util/dvsim/dvsim.py hw/ip/otbn/dv/uvm/otbn_sim_cfg.hjson -i otbn_escalate --fixed-seed=2
```

As @vogelpi correctly described in the comments for 5228dfccf50, the CDC for this signal means that it might have a silly value for a cycle, so we shouldn't lock up if that happens. Unfortunately, the silly value might get stored into `rma_ack_q`, at which point the check on line 375 of `otbn_start_stop_control` fires and locks the block.

I think the fix might be as simple as deleting the `state_d` update.
```
    if (mubi4_test_invalid(rma_ack_q)) begin
      rma_ack_d = MuBi4False;
      mubi_err_d = 1'b1;
      state_d = OtbnStartStopStateLocked;
    end
```

@vogelpi: Would you mind taking a look?

rswarbrick:
The RMA signal comes into the block with an rma_req_i port and is encoded as lc_ctrl_pkg::lc_tx_t. If this happens to appear through a CDC then the 4-bit signal might be corrupted for a cycle on the way and the design is supposed to ignore it (until the value is strictly equal to lc_ctrl_pkg::On.

For the ES chip, there was a bug here, where we would lock OTBN if this happened. This bug was mostly fixed by commit 5228dfc, but that isn't quite a complete fix. The problem is that rma_req_i gets sampled and stored into rma_ack_d when the start/stop control block is in FSM state OtbnStartStopSecureWipeComplete.

This is then clocked into rma_ack_q. The existing code correctly squashed rma_ack_d back to MuBi4False if rma_ack_q was invalid. But it also locked OTBN, which is different from the intended behaviour (ignore invalid signals).

This commit is rather trivial and drops the code that was handling the invalid signals.

Fixes #23585. @vogelpi: would you mind taking a look? I *think* it's a reasonably trivial tidy-up, but I'd love a proper review from someone who knows how this is supposed to work.

### DIFF
--- a/hw/ip/otbn/rtl/otbn_start_stop_control.sv
+++ b/hw/ip/otbn/rtl/otbn_start_stop_control.sv
@@ -374,8 +374,6 @@ module otbn_start_stop_control
     end
     if (mubi4_test_invalid(rma_ack_q)) begin
       rma_ack_d = MuBi4False;
-      mubi_err_d = 1'b1;
-      state_d = OtbnStartStopStateLocked;
     end
   end
 


vogelpi:
Thanks @rswarbrick for debugging this and for filing the PR. The RTL behaves as intended what should be changed here is the test. Please allow me to explain why.

`rma_req_i` is only latched into `rma_ack_d/q` at the very end of the secure wipe in the `OtbnStartStopSecureWipeComplete` state of the FSM. The secure wipe itself takes in the order of 10 to 100 clock cycles so by the time it ends, `rma_req_i` has for sure stabilized to a good valid value (assuming the secure wipe was initialized due to a change in `rma_req_i`). 

This is documented in the code, see at around Line 360:
```
    // If the MuBi signals take on invalid values, something bad is happening. Put them back to
    // a safe value (if possible) and signal an error. The only exception is rma_req_i. This LC
    // signal may experience staggered transitions due to CDCs leading to invalid values. In
    // accordance with the spec, invalid values of non-escalation LC signals must be treated as
    // OFF. rma_ack_d/q is driven by rma_req_i but only at the end of the secure wipe. By that
    // time rma_req_i has for sure stabilized.
```

Now, what happens if we're performing a secure wipe due to another reason and we get an invalid `rma_req_i` signal (because it currently stabilizes) while being in that `OtbnStartStopSecureWipeComplete`? This is really bad luck, OTBN will lock up and we can't enter RMA. The chip has to be restarted and we need to retry.

Why we shouldn't change the design as you proposed in this PR: because it stops escalations upon invalid `rma_ack_q` signals. The complexity of glitching those 4 bits in succession vs. all at once is different. We shouldn't take risk here as this signal can be attacked to try putting the chip in to RMA mode before the secure wipe has finished. It's a security risk.

rswarbrick:
Ok, thanks for the explanation. I think I understand. I'll try to debug the situation further this morning. Probably the solution is just to only sample the RMA signal at the right time in the DV code (and keep the "lock on invalid signal" behaviour).

PR #23589, opened 2024-06-09
[spi_device] Mux in missing scan_rst_ni connectivity

a-will:
The newer out_clk-synchronized resets were missing direct connectivity for scan reset, as was the CSR-connected reset to the prim_fifo_async used for flash status register updates moving from SW to the SPI clock domain. Add the requisite muxes to patch that up.

### DIFF
--- a/hw/ip/spi_device/rtl/spi_device.sv
+++ b/hw/ip/spi_device/rtl/spi_device.sv
@@ -706,7 +706,7 @@ module spi_device
     .clk_o(rst_spi_n)
   );
 
-  logic rst_spi_in_n, rst_spi_out_n;
+  logic rst_spi_in_n, rst_spi_out_sync_n, rst_spi_out_n;
   assign rst_spi_in_n = rst_spi_n;
 
   // Synchronizes reset de-assertion to safely occur in the outclk domain.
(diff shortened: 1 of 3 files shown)


hw/ip/spi_device/rtl/spi_device.sv
```diff
@@ -739,7 +748,16 @@ module spi_device
...
     .d_i   (1'b1),
-    .q_o   (tpm_rst_out_n)
+    .q_o   (tpm_rst_out_sync_n)
+  );
+
+  prim_clock_mux2 #(
+    .NoFpgaBufG(1'b1)
+  ) u_tpm_rst_out_scan_mux (
+    .clk0_i (tpm_rst_out_sync_n),
+    .clk1_i (scan_rst_ni),
```
a-will, line 758: This felt a bit superfluous, since `tpm_rst_out_sync_n` would be `scan_rst_ni` with a synchronized de-assertion on `scan_clk_i`, but I've added them just in case. The tools complain about the missing mux, at least.

a-will: Oh, right, the flops that follow would no longer be sampling on the opposite edge, so there would be a new timing problem. Okay, these make sense.

hw/ip/spi_device/rtl/spid_status.sv
```diff
@@ -263,7 +267,15 @@ module spid_status
       status_fifo_clr_n <= !sys_update_clr_i;
     end
   end
-  assign status_fifo_rst_n = sys_rst_ni & status_fifo_clr_n;
```
vogelpi, line 266: `sys_rst_ni` seems no longer to be factored into `status_fifo_rst_n` but I assume this change is intended.

vogelpi (approved):
LGTM, thanks @a-will .

hw/ip/spi_device/rtl/spid_status.sv
a-will: Mostly, I figured it was unnecessary, since it provides the async reset to the flop driving status_fifo_rst_n. The extra AND gate seemed superfluous.

a-will:
Merging. The failure is unrelated and was fixed by a separate PR (already merged).

PR #23590, opened 2024-06-09
[dv,usbdev] Monitor improvements and error detection

alees24:
This PR mostly extends the usb20_monitor to support other bus configurations and non-packet signaling on the USB, and to detect and indicate the presence of errors in the received packets. The driver is also simplified slightly to leave only the 'mechanism' within the driver and leave the sequence to determine what activity shall be generated ('policy').

- Correct the collection of packet traffic to validate the End Of Packet.
- Generalize the collection of symbols from the USB so that the other bus configuration may readily be tested later.
- Implement the detection of invalid SYNC signals, invalid EOP and bit stuffing errors; flag these in the item that we will forward to the scoreboard.
- This also means that the agent is now using true tri-stated USB signaling.
- Support the detection and reporting of non-packet USB events (Bus Reset, Suspend-Resume Signaling) in the usb20_monitor.
- Modify the start up sequence of the usb20_driver to be passive, so that the sequence may determine whether to generate valid USB traffic. This then permits the usb20_monitor to start up appropriately or to remain inactive and not report invalid bus states.

Following the above changes in the agent, the sequences are simplified with regard to connection and device initialization, and the generation of invalid bus traffic shall now be detected and tolerated in the monitor.This should raise the overall DV test pass rate into the high 90s for the first time and all extant tests should now pass with all seeds.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_item.sv
+++ b/hw/dv/sv/usb20_agent/usb20_item.sv
@@ -15,6 +15,13 @@ class usb20_item extends uvm_sequence_item;
   // Indicates that a timeout occurred when awaiting a response from the device.
   bit timed_out = 1'b0;
 
+  // Validity indicators that apply to all packet types; used by the monitor at metadata for the
+  // scoreboard.
+  bit valid_sync;      // SYNC signal properly formed.
+  bit valid_length;    // Appropriate number of bits.
+  bit valid_stuffing;  // No bit stuffing violations.
+  bit valid_eop;       // End Of Packet correct.
+
   `uvm_object_utils_begin(usb20_item)
   `uvm_object_utils_end
 
--- a/hw/dv/sv/usb20_agent/usb20_monitor.sv
+++ b/hw/dv/sv/usb20_agent/usb20_monitor.sv
@@ -17,18 +17,26 @@ class usb20_monitor extends dv_base_monitor #(
    data_pkt       m_data_pkt;
    handshake_pkt  m_handshake_pkt;
 
-   bit bit_destuffed[$];
-   bit [2:0] eop_cnt = 3'b000;
-   bit [7:0] packet_type;
-
   `uvm_component_new
 
   function void build_phase(uvm_phase phase);
     super.build_phase(phase);
+    m_usb20_item = usb20_item::type_id::create("m_usb20_item", this);
     m_sof_pkt = sof_pkt::type_id::create("m_sof_pkt", this);
     m_token_pkt = token_pkt::type_id::create("m_token_pkt", this);
     m_data_pkt = data_pkt::type_id::create("m_data_pkt", this);
     m_handshake_pkt = handshake_pkt::type_id::create("m_handshake_pkt", this);
+    // Complete the event type for the packet objects because these never change.
+    m_sof_pkt.m_ev_type = EvPacket;
+    m_token_pkt.m_ev_type = EvPacket;
+    m_data_pkt.m_ev_type = EvPacket;
+    m_handshake_pkt.m_ev_type = EvPacket;
+    // TODO: presently we do not expect any invalid SYNC signals, but at some point we should be
+    // testing that the DUT correctly ignores such packets/interference.
+    m_sof_pkt.valid_sync = 1;
+    m_token_pkt.valid_sync = 1;
+    m_data_pkt.valid_sync = 1;
+    m_handshake_pkt.valid_sync = 1;
     if (!uvm_config_db#(virtual usb20_block_if)::get(this, "*.env.m_usb20_agent*", "bif",
                                                      cfg.bif)) begin
       `uvm_fatal(`gfn, "failed to get usb20_block_if handle from uvm_config_db")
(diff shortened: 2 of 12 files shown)


hw/dv/sv/usb20_agent/usb20_item.sv
```diff
@@ -12,6 +12,12 @@ class usb20_item extends uvm_sequence_item;
   pkt_type_e m_pkt_type;
   usb_transfer_e m_usb_transfer;
 
+  // Validity indicators that apply to all packet types.
```
rswarbrick, line 15: Would it make sense to add something to the comment to explicitly say that these are generated by the monitor, but the fields are ignored by the driver?

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -41,144 +49,335 @@ class usb20_monitor extends dv_base_monitor #(
...
+        USB20Sym_K:
+          if (suspended) begin
+            collect_resume_signaling();
+            suspended = 0;
+          end else begin
+            collect_packet();
+          end
+        // Detection of Bus Resets.
+        USB20Sym_SE0: begin
+          idle_time = 0;
```
rswarbrick, line 107: Nit: I don't think we actually need to write this local variable here? Similarly for where we clear `suspended` on line 101 (although that might be more informative for a reader, so maybe worth keeping)

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -41,144 +49,335 @@ class usb20_monitor extends dv_base_monitor #(
...
+    // slower than Full Speed signaling, but we have already collected the first symbol.
+    for (int i = 0; i < 23; i++) begin
+      collect_symbol(sym);
+      valid = valid & (sym == ((i >= 15) ? USB20Sym_J : USB20Sym_SE0));
+    end
+    // TODO: presently we are still relying upon the DUT driver enable to ascertain who is talking;
+    // the DUT shall NEVER be generating Resume Signaling.
+    `DV_CHECK_EQ(cfg.bif.usb_dp_en_o, 0, "Bus Resume Signaling generated by DUT")
+    // Check that the Resume Signaling from the driver is as expected.
+    `DV_CHECK_EQ(valid, 1, "Invalid Resume Signaling")
```
rswarbrick, line 194: Would it make sense to split this check up and and replace the current lines 183 and 188 with it? That might be a bit easier to debug if something invalid happens.

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -41,144 +49,335 @@ class usb20_monitor extends dv_base_monitor #(
...
-      packet.push_back(cfg.bif.usb_p);
-      @(posedge cfg.bif.clk_i);
-      @(posedge cfg.bif.clk_i);
-      @(posedge cfg.bif.clk_i);
-      @(posedge cfg.bif.clk_i);
+    // We never expect to receive invalid SYNC signals at present; at some point we may generate
+    // them from the driver, but the DUT should never generate them.
+    `DV_CHECK_GT(packet.size(), 8, "Incomplete SYNC signal detected")
+    for (int unsigned b = 0; b < 8; b++) begin
+      `DV_CHECK_EQ(packet[b], sync[b], "Invalid SYNC signal detected")
```
rswarbrick, line 235: I wonder whether it makes sense to print out the SYNC signal that we saw in hex here, which might be a bit easier to debug.

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -41,144 +49,335 @@ class usb20_monitor extends dv_base_monitor #(
...
-    // TODO: this does not even check the EOP!
-    // TODO: we should almost certainly be consuming the EOP interval here.
 
-    `uvm_info(`gfn, $sformatf("Complete monitored packet = %p ", packet), UVM_HIGH)
     nrzi_decoder(packet, monitored_decoded_packet);
-    bit_destuffing(monitored_decoded_packet, bit_destuffed);
-    classifies_packet();
+    bit_destuffing(monitored_decoded_packet, valid_stuffing, destuffed_packet);
+    classifies_packet(valid_eop, valid_stuffing, destuffed_packet);
     packet.delete();
```
rswarbrick, line 241: I think you can drop this line too, now that `packet` is a local variable.

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -41,144 +49,335 @@ class usb20_monitor extends dv_base_monitor #(
...
     // bits to binary conversion
-    sof_result = sof_p'(bit_destuffed);
-    m_sof_pkt.m_pid_type = pid_type_e'(packet_type);
-    m_sof_pkt.framenum = {<<{sof_result[18:8]}};
-    m_sof_pkt.crc5 = {<<{sof_result[7:3]}};
+    m_sof_pkt.m_pid_type = pid;
+    m_sof_pkt.valid_eop = valid_eop;
+    m_sof_pkt.valid_stuffing = valid_stuffing;
+    if (destuffed_packet.size() == 32) begin
+      sof_result = sof_p'(destuffed_packet);
```
rswarbrick, line 286: Nit: Maybe worth moving the declaration of `sof_result` into this block Similarly with `token_result` in the `token_packet` task below.

(I'd probably drop the typedefs as well, since they just seem to be a clunky way of saying "look at 32 bits" or similar. But maybe that's for a separate PR)

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -41,144 +49,335 @@ class usb20_monitor extends dv_base_monitor #(
...
+    end else m_token_pkt.valid_length = 0;
+
     req_analysis_port.write(m_token_pkt);
   endtask
 
 //------------------------------------------Data Packet-------------------------------------------//
-  virtual task data_packet();
-    bit [7:0] data_pid;
-    bit [7:0] data_temp[];
+  virtual task data_packet(pid_type_e pid, bit valid_eop, bit valid_stuffing,
```
rswarbrick, line 317: Can this be changed to a function? I don't think we consume any time.

Similar question for `handshake_packet` below.

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -200,15 +399,21 @@ class usb20_monitor extends dv_base_monitor #(
   endtask
 
   // Bit Destuffing
-  task bit_destuffing(input bit packet[], output bit packet_destuffed[$]);
+  task bit_destuffing(input bit packet[], output bit valid_stuffing,
+                      output bit packet_destuffed[$]);
```
rswarbrick, line 403: Definitely not for this PR, but things would probably get a bit easier to follow if we had a class representing "destuffed packet and validity bit". The end result would be fewer variables, I think.

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -236,25 +442,4 @@ class usb20_monitor extends dv_base_monitor #(
     // the negedge of the clk.
     cfg.clk_rst_if_i.wait_for_reset(0, 1);
   endtask
-
-  // TODO: is this used; what was it meant to do?!
```
rswarbrick, line 240: :-D

hw/ip/usbdev/dv/env/seq_lib/usbdev_phy_pins_sense_vseq.sv
```diff
@@ -10,6 +10,17 @@ class usbdev_phy_pins_sense_vseq extends usbdev_base_vseq;
   rand bit set_dp;
   rand bit set_dn;
 
+  task pre_start();
+    // Prevent normal device initialization.
+    //
+    // This sequence is designed to drive the USB lines without valid USB traffic, so the
+    // last thing we need is the driver and monitor connecting, and then rejecting the bus state
+    // or apparenty traffic.
```
rswarbrick, line 18: Nit: Typo in "apparent"?

rswarbrick (approved):
This looks like a dramatic improvement! Some nitty questions and suggestions, but I'm happy for this to go in either way :-)

hw/dv/sv/usb20_agent/usb20_monitor.sv
alees24: Sure. That's true throughout much of the codebase. Thanks.

hw/dv/sv/usb20_agent/usb20_monitor.sv
alees24: Ah, yes, I think I guess that was the intent and it worked on vcs I suspect. With xcelium it has been leading to errors and failed tests when the `destuffed_packet` bit array was longer. Goodbye ugly typedefs.

hw/dv/sv/usb20_agent/usb20_monitor.sv
alees24: Thanks. A consequence of code restructuring.

PR #23591, opened 2024-06-09
[dv/tlt] Fix ROM JTAG injection tests.

moidx:
The OTP images used in ROM e2e test cases set `EN_SRAM_IFETCH` to `True` inside the `HW_CFG1` partition. The `sram_ctrl` checks the otp value to determine to use the fetch enable signal provided by `lc_ctrl` versus the `sram_ctrl` `exec` register value.

This change enables SRAM execution by setting the `sram_ctrl` exec register. The `chip_sw_sram_ctrl_execution_main` test cases checks all the other OTP versus SRAM register permutations.

### DIFF
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_sw_rom_e2e_jtag_inject_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_sw_rom_e2e_jtag_inject_vseq.sv
@@ -24,7 +24,7 @@ class chip_sw_rom_e2e_jtag_inject_vseq extends chip_sw_base_vseq;
 
     // The steps in this test closely follow the gdb steps, which are encoded
     // in a host-side (Rust) test harness located at
-    // sw/host/tests/rom/e2e_openocd_debug_test/src/main.rs
+    // sw/host/tests/chip/jtag/src/sram_load.rs
 
     // Let the chip power up and allow the ROM to execute upto an arbitrary point.
     `DV_WAIT(cfg.chip_vif.pwrmgr_fast_pwr_state_active)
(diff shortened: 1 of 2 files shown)


hw/top_earlgrey/dv/env/seq_lib/chip_sw_rom_e2e_jtag_inject_vseq.sv
```diff
@@ -74,16 +74,28 @@ class chip_sw_rom_e2e_jtag_inject_vseq extends chip_sw_base_vseq;
...
+
+    // The dm package only has definitions for the first CFG and ADDR CSRs.
+    // This test uses the ibex_pkg definitions, and thus, we check that the
+    // offset values are the same for each respective index 0.
+    `DV_CHECK_EQ(dm::CSR_PMPCFG0, ibex_pkg::CSR_PMPCFG0);
+    `DV_CHECK_EQ(dm::CSR_PMPADDR0, ibex_pkg::CSR_PMPADDR0);
+
+    // NAPOT LX-R for PMP2   (ROM)
+    // TOR   LXWR for PMP1-0 (SRAM)
+    dbg_cmd_reg_write(.regno(ibex_pkg::CSR_PMPCFG0),   .value_q({'h009d8f00}));
```
a-will, line 94: Are we actually doing the configuration for the ROM? Or is the comment about PMP2 more about not changing the values?

(Since I didn't see a write for PMPADDR2)

a-will (approved):

hw/top_earlgrey/dv/env/seq_lib/chip_sw_rom_e2e_jtag_inject_vseq.sv
moidx: The hardware configures PMPADDR2. This test is taking PMPADDR1-0, so the comment is to clarify. I will add more details in a follow up change. Thanks!

Issue #23578, opened 2024-06-07
[otbn,dv] The `otbn_sec_cm` test is broken for all seeds


rswarbrick:
### Description

A bit of bisection shows that this error was caused by:

574ece386f * [flash_ctrl/rtl] Enable SecFifoPtr for TL-UL FIFOs

To run a test: `util/dvsim/dvsim.py hw/ip/otbn/dv/uvm/otbn_sim_cfg.hjson -i otbn_sec_cm --fixed-seed=123`, which will fail with a message like:
```
* `xmsim: *E,ASRTST (/home/rjs/work/opentitan/scratch/tmp/otbn-sim-xcelium/default/src/lowrisc_prim_fifo_*/rtl/prim_fifo_sync.sv,151): Assertion DataKnown_A has failed` has 1 failures:
    * Test otbn_sec_cm has 1 failures.
        * 0.otbn_sec_cm.123\
          Line 150, in log /home/rjs/work/opentitan/scratch/tmp/otbn-sim-xcelium/0.otbn_sec_cm/latest/run.log

                xmsim: *E,ASRTST (/home/rjs/work/opentitan/scratch/tmp/otbn-sim-xcelium/default/src/lowrisc_prim_fifo_0/rtl/prim_fifo_sync.sv,151): (time 1014314 PS) Assertion tb.dut.u_tlul_adapter_sram_imem.u_sramreqfifo.DataKnown_A has failed
                UVM_ERROR @   1014314 ps: (prim_fifo_sync.sv:151) [ASSERT FAILED] DataKnown_A
                UVM_INFO @   1014314 ps: (uvm_report_catcher.svh:705) [UVM/REPORT/CATCHER]
                --- UVM Report catcher Summary ---
```

Taking a quick look at a log, this error appears just after:
```
UVM_INFO @   1009051 ps: (prim_count_if.sv:38) [tb.dut.u_tlul_adapter_sram_imem.u_sramreqfifo.gen_normal_fifo.u_fifo_cnt.gen_secure_ptrs.u_rptr.u_prim_count_if] Forcing tb.dut.u_tlul_adapter_sram_imem.u_sramreqfifo.gen_normal_fifo.u_fifo_cnt.gen_secure_ptrs.u_rptr.cnt_q[0] from 0 to 1
```
so presumably we're telling a FIFO that it has entries, but they actually turn out to be 'X.

My guess is that the assertion needs turning off in this situation with an `$assertoff` call.

[...]

PR #23595, opened 2024-06-10
[prim_fifo_sync/rtl] Fix err_o behavior

nasahlpa:
When an err_o is raised by the u_fifo_cnt (there is a ptr error inside this module), we should not raise the rvalid_o signal of the FIFO and we should not forward rdata_o = rdata_int as this data is not valid.

Instead, set rdata_o = '0 and deasser the rvalid_o signal.

Closes lowRISC/opentitan#23578


GregAC:
I am slightly wary of killing the valid is the presence of an error. In particular I worry this will cause some issue where a bus response vanishes which could lock things up. I note you've just closed this so perhaps you're already doing a different fix anyway?

PR #23598, opened 2024-06-10
[test] Add test data for SPHINCS+ with SHA2 and generalize test infrastructure naming.

jadephilipoom:

This change doesn't alter any of the actual SPHINCS+ implementation, just does some preliminary setup so that we can test SHA2 parameter sets.

- First commit adds new test data that I generated from the SPHINCS+ reference implementation, following the procedure in `sw/device/tests/crypto/testvectors/sphincsplus_kat/README.md`.
- Second commit renames the test vector template for SPHINCS+ without changes (it was previously called `sphincsplus_shake_128s_testvectors.h.tpl` but doesn't have anything specific to the shake-128s parameter set, so I renamed it `sphincsplus_testvectors.h.tpl` to avoid confusion)
- Third commit fixes references to the testvectors template with the new name.


moidx (approved):
Thanks @jadephilipoom!. LGTM

PR #23599, opened 2024-06-10
[spi_device, dv] disallow certain commands if busy

antmarzam:
en/ex4b wren/wrdis and upload commands are ignored if busy bit is set 
FCOV has been added to know when a command is ignored due to flash status busy bit being set

### DIFF
--- a/hw/ip/spi_device/dv/env/spi_device_env_cov.sv
+++ b/hw/ip/spi_device/dv/env/spi_device_env_cov.sv
@@ -22,6 +22,14 @@ class tpm_read_hw_reg_cg_wrap;
   endfunction : sample
 endclass
 
+covergroup busy_blocks_command_cg with function sample (bit command_blocked);
+  option.per_instance = 1;
+  cp_blocked_or_allowed: coverpoint command_blocked {
+    bins allowed = {0};
+    bins blocked = {1};
+  }
+endgroup
+
 class spi_device_env_cov extends cip_base_env_cov #(.CFG_T(spi_device_env_cfg));
   `uvm_component_utils(spi_device_env_cov)
   tpm_read_hw_reg_cg_wrap tpm_read_hw_reg_cg_wraps[string];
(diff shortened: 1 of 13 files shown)


hw/ip/spi_device/dv/env/spi_device_env_cov.sv
```diff
@@ -307,6 +321,18 @@ class spi_device_env_cov extends cip_base_env_cov #(.CFG_T(spi_device_env_cfg));
     en4b_flash_cmd_clash_with_addr_mode_write_cg = new();
     ex4b_flash_cmd_clash_with_addr_mode_write_cg = new();
 
+
+    en4b_block_cmd_cg = new();
+    en4b_block_cmd_cg.option.name = "en4b_block_cmd_cg";
```
rswarbrick, line 326: Just because I'm interested: what does this change do? Does it improve the coverage report in some way?

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
```diff
@@ -197,116 +197,139 @@ class spi_device_scoreboard extends cip_base_scoreboard #(.CFG_T (spi_device_env
                 InternalProcessCfgCmd: begin
                   bit prev_wel = `gmv(ral.flash_status.wel);
                   if (`GET_OPCODE_VALID_AND_MATCH(cmd_info_en4b, item.opcode)) begin
-                    if (cfg.en_cov) begin
-                      cov.spi_device_addr_4b_enter_exit_command_cg.sample(
-                          .addr_4b_en(1), .prev_addr_4b_en(`gmv(ral.addr_mode.addr_4b_en)));
```
rswarbrick, line 202: I see that we're no longer sampling in `spi_device_addr_4b_enter_exit_command_cg`. Is that an intentional change? (I think the covergroup still exists)

hw/ip/spi_device/dv/env/spi_device_env_cov.sv
antmarzam: I don't think it does actually - It's typically used when you have an array of covergroups, so you can name them differently. But I think it's unnecessary in this case. Will remove. Thank you - good spot :)

hw/ip/spi_device/dv/env/spi_device_scoreboard.sv
antmarzam: We are, it's just a bit below on line -205

rswarbrick (approved):
Looks good to me! Before merge, please can you force-push and squash the two commits together? (This will avoid a chain of "do a thing; undo it" in the history)

Issue #23526, opened 2024-06-06
[edn] valid/ready protocol violations in case of ack errors and disablement


vogelpi:
### Description

I noticed a couple of issues between EDN / CSRNG related to the CSRNG CMD interface:
- In case of disablement and in case of ACK errors triggered inside CSRNG, EDN can de-asserted the valid without waiting for the ready. Similarly, it can send less additional data than announced in the first command word (CSRNG cannot handle that).
- In case of ACK errors, CSRNG pops the erroneous word from the command FIFO. But the command FIFO may already contain additional data belonging to that command which may be left in the FIFO. Instead of just popping the FIFO, the FIFO should probably be cleared.

Both things are non-ideal but not catastrophic as we can always get out of these situations by disabling / re-enabling both EDNs and CSRNG which is anyway recommended in the programmer's guide right now. However, it would be better if those protocol violations wouldn't happen and if we wouldn't have to shutdown all CSRNG contexts because of an issue triggered by a single EDN.

vogelpi:
@moidx as long as we have a way to restart the entropy complex it should be acceptable to de-prioritize this.

vogelpi:
@vogelpi will try to see if this can be handled tomorrow.

PR #23601, opened 2024-06-10
[csrng] Flush command FIFO after signaling ACK errors

vogelpi:
The decision whether a command is invalid is taken based on the command header but EDN might continue and send the additional data as well (it might already be in the FIFO). Previously, CSRNG would just pop the invalid command header from the FIFO and leave any additional data in the FIFO which can then erroneously be interpreted as the next command header.

Instead, we have to absorb the full invalid command before we can continue. This commit modifies the design to keep emptying the FIFO until the command interface is idle.

This is related to lowRISC/OpenTitan#23526.

### DIFF
--- a/hw/ip/csrng/rtl/csrng_cmd_stage.sv
+++ b/hw/ip/csrng/rtl/csrng_cmd_stage.sv
@@ -221,38 +221,39 @@ module csrng_cmd_stage import csrng_pkg::*; #(
   // state machine to process command
   //---------------------------------------------------------
   // Encoding generated with:
-  // $ ./util/design/sparse-fsm-encode.py -d 3 -m 10 -n 8 \
-  //      -s 170131814 --language=sv
+  // $ ./util/design/sparse-fsm-encode.py -d 3 -m 11 -n 8 \
+  //     -s 170131814 --language=sv
   //
   // Hamming distance histogram:
   //
   //  0: --
   //  1: --
   //  2: --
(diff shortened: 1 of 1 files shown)


hw/ip/csrng/rtl/csrng_cmd_stage.sv
```diff
@@ -221,38 +221,39 @@ module csrng_cmd_stage import csrng_pkg::*; #(
   // state machine to process command
   //---------------------------------------------------------
   // Encoding generated with:
-  // $ ./util/design/sparse-fsm-encode.py -d 3 -m 10 -n 8 \
-  //      -s 170131814 --language=sv
+  // $ ./util/design/sparse-fsm-encode.py -d 3 -m 11 -n 8 \
+  //     -s 170131814 --language=sv
```
rswarbrick, line 225: (super nit): I think we've dropped a space here. If you're re-pushing, it might be worth putting it back. If not, it *really* doesn't matter :-)

rswarbrick (approved):

vogelpi:
I am merging this, there is one FPGA test failing which is known to be flaky/broken atm (spi_passthrough).

hw/ip/csrng/rtl/csrng_cmd_stage.sv
vogelpi: Thanks for your review @rswarbrick ! I think there was actually one space too much before ;-)

PR #23602, opened 2024-06-10
[csrng/dv] Correct prediction of FIPS/CC compliance flag for UNI cmd

vogelpi:
This got accidentally modified with commit 3b47465d33e980e728ce96168cab07093efcf6ee. See also lowRISC/OpenTitan#23349.


rswarbrick (approved):
Thanks for explaining (and yep: it's pretty clear what went wrong when I look at the change we're undoing).

vogelpi:
Thanks for your review @rswarbrick ! This is a DV only change without impact on TLTs. I am thus merging this now.

PR #23603, opened 2024-06-10
[sram_ctrl/rtl] Use flop instead of prim_sparse_flop

nasahlpa:
This commit changes back the PRIM_FLOP_SPARSE_FSM to an ordinary flop.


vogelpi (approved):
This PR undoes a specific change introduced with https://github.com/lowRISC/opentitan/commit/ee5f4a1abf7dcbfca031bdad5e5aa4aafaefdb33#diff-ee7044ef51037db1e31cb795d5fa28525ca5705a5c0d8f7e46ba9888414862cf
to avoid a synthesis/elaboration issue. It seems some synthesis tools don't like the macro inside the if/else. LGTM!

vogelpi:
The only failure here is the known broken spi_passthru_test_fpga_cw310_rom_with_fake_keys . Merging.

PR #23604, opened 2024-06-10
[otbn,dv] Partially fix RMA request handling

rswarbrick:
This isn't completely working, because of the bug that I tried to fix in (the incorrect) PR #23587. With that incorrect fix working around the problem I was seeing there, I stumbled into some other issues, which this PR fixes properly.

- When we've taken an RMA request, we go down a path where we use the `stop` function to cancel any instructions that happen to be in flight. We want that behaviour, but we don't want the function to also forget about the RMA request. The first commit tidies that up.
- On an RMA request that happens before we run any operations, we don't want OTBN to actually do a secure wipe. In the design, the logic is that there's nothing that needs wiping anyway! The second commit makes the model use the same logic.
- When we take an RMA request, we need the STATUS register to jump to LOCKED. This register is "double flopped" in the model, to represent the exact timing that we expect. But the "I'm locking" logic ends up throwing away the partially committed change to the STATUS register. The third commit stops us from doing that.

### DIFF
--- a/hw/ip/otbn/dv/otbnsim/sim/sim.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/sim.py
@@ -175,11 +175,31 @@ class OTBNSim:
             self.state.ext_regs.write('INSN_CNT', 0, True)
 
         if self.state.init_sec_wipe_is_running():
-            # Wait for the URND seed. Once that appears, switch to WIPING_GOOD
-            # unless the FSM state is already LOCKED, in which case we change
-            # it to WIPING_BAD.
+            # Wait for the URND seed. If there is some genuine state to wipe
+            # (because self.state.has_state_to_wipe is true), then we switch to
+            # WIPING_GOOD unless the FSM state is already LOCKED, in which case
+            # we change it to WIPING_BAD.
+            #
+            # As a special case, there might be an RMA request and this might
+            # be just after reset. In that case, we don't need to worry about
+            # wiping state (because the entropy complex might not have been set
+            # up yet, we might not even be able to seed URND). In that case, we
+            # jump straight to the LOCKED state.
             if self.state.wsrs.URND.running:
-                if self.state.get_fsm_state() == FsmState.LOCKED:
+                cur_fsm_state = self.state.get_fsm_state()
+
+                # If URND finishes, it's probably the seed for a secure wipe.
+                # But there's an extra possibility: we might have seen an RMA
+                # request before any instructions get run. In that case, the
+                # rma_req flag will be high and the has_state_to_wipe flag will
+                # be low. Jump straight to the LOCKED state and skip the
+                # WIPING_BAD state (since there is nothing that needs wiping
+                # anyway).
+                if self.state.rma_req and not self.state.has_state_to_wipe:
+                    self.state.complete_init_sec_wipe()
+                    self.state.set_fsm_state(FsmState.LOCKED)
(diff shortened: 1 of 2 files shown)


hw/ip/otbn/dv/otbnsim/sim/state.py
```diff
@@ -353,7 +353,7 @@ def stop(self) -> None:
         # changes.
         insn_failed = self._err_bits and self._fsm_state == FsmState.EXEC
         if insn_failed or self.rma_req:
-            self._abort()
+            self._abort(not_status=True)
```
marnovandermaas, line 362: should this feed in the not_status argument?

rswarbrick: Yes, it definitely should! I just noticed it on my machine and jumped back to the PR to see whether I was doing the right thing. Duh...

hw/ip/otbn/dv/otbnsim/sim/sim.py
```diff
@@ -175,11 +175,29 @@ def _step_idle(self, verbose: bool) -> StepRes:
...
-            # unless the FSM state is already LOCKED, in which case we change
-            # it to WIPING_BAD.
+            # Wait for the URND seed. If there is some genuine state to wipe
+            # (because self.state.has_state_to_wipe is true), then we switch to
+            # WIPING_GOOD unless the FSM state is already LOCKED, in which case
+            # we change it to WIPING_BAD.
+            #
+            # As a special case, there might be an RMA request and this might
+            # be the start of time so we don't actually have any state to wipe.
+            # In that case, we jump straight to the LOCKED state.
```
vogelpi, line 185: I think the code does the right thing but the reasoning is a bit different:
- After a reset we ideally always reseed URND and then to do a secure wipe. There can be state because for FI hardening reasons, we've removed the resets from the wide register file (see https://opentitan.org/book/doc/security/implementation_guidelines/reset_vs_non-reset_flops/index.html#appendix-otbn---a-case-for-non-reset-flops to get some insight on the motivation).
- If we get an RMA request right after reset, we don't perform the URND reseed. Because the RMA request typically comes before the entropy complex is set up. So the URND reseed might never finish and thus the wipe never happen and thus RMA never entered.
- Not doing the full secure wipe (including URND) is acceptable from a security perspective in the case of RMA entry only because:
  - Triggering the RMA entry is itself hardened.
  - After RMA entry, the chip resets and while booting into RMA, another full secure wipe including URND reseed happens.

vogelpi (approved):
Thanks @rswarbrick for debugging and fixing this! I think this does the right thing. I've left some comments to give some background.

hw/ip/otbn/dv/otbnsim/sim/sim.py
rswarbrick: Thanks. I've tweaked the text to be a slightly looser version of what you're saying :-)

hw/ip/otbn/dv/otbnsim/sim/state.py
```diff
@@ -329,7 +335,7 @@ def start(self) -> None:
 
         self._urnd_client.request()
 
-    def stop(self) -> None:
+    def stop(self, not_status: bool) -> None:
```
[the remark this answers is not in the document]
rswarbrick: ... because I hadn't thought hard enough! Should be gone now.

PR #23605, opened 2024-06-10
[sival] Enable high UART baud rates on silicon

jwnrt:
Thanks to https://github.com/lowRISC/opentitan/pull/23379 we can now test high baud rates (1MBd, 1.5Mbd) so we enable this on silicon targets.

It's still disabled on the FPGA which cannot clock high enough to test these.

Thanks to @AlexJones0 for the help debugging.

### DIFF
--- a/sw/device/tests/uart_baud_rate_test.c
+++ b/sw/device/tests/uart_baud_rate_test.c
@@ -24,8 +24,10 @@ static const uint32_t kBaseAddrs[4] = {
     TOP_EARLGREY_UART2_BASE_ADDR,
     TOP_EARLGREY_UART3_BASE_ADDR,
 };
-static const uint32_t kBauds[9] = {4800,   9600,   19200,  38400, 57600,
-                                   115200, 230400, 128000, 256000};
+static const uint32_t kBauds[11] = {
+    4800, 9600, 19200, 38400, 57600, 115200, 230400, 128000, 256000, 1000000,
+    // This baud rate will only work on silicon, the FPGA cannot go fast enough:
+    1500000};
 
 enum {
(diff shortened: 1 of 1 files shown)


sw/device/tests/uart_baud_rate_test.c
```diff
@@ -24,11 +24,17 @@ static const uint32_t kBaseAddrs[4] = {
     TOP_EARLGREY_UART2_BASE_ADDR,
     TOP_EARLGREY_UART3_BASE_ADDR,
 };
-static const uint32_t kBauds[9] = {4800,   9600,   19200,  38400, 57600,
-                                   115200, 230400, 128000, 256000};
+static const uint32_t kBauds[11] = {4800,   9600,    19200,  38400,
+                                    57600,  115200,  230400, 128000,
+                                    256000, 1000000, 1500000};
```
alees24, line 29: Perhaps worth demarcating with a comment those baud rates which are available only on Si from those that may be used everywhere, to be a bit more robust against the insertion or appending of additional rates in future?

sw/device/tests/uart_baud_rate_test.c
```diff
@@ -24,11 +24,17 @@ static const uint32_t kBaseAddrs[4] = {
...
 
 enum {
   kTestTimeoutMillis = 500000,
+  // Number of Bauds to test from `kBauds`.
+  kBaudCountSilicon = 11,
```
alees24, line 34: I think you can use ARRAYSIZE() here and ARRAYSIZE() - 2 below, to make it a little less likely that the enum values and the array become inconsistent.

alees24 (approved):
This look good, and a very welcome improvement, thanks.

PR #23607, opened 2024-06-10
[edn] Hold csrng_req_valid until valid/ready handshake completes

vogelpi:
Previously, EDN could immediately de-assert the valid upon getting an ACK error thereby violating the valid/ready handshake protocol between EDN and CSRNG.

### DIFF
--- a/hw/ip/edn/dv/edn_sim_cfg.hjson
+++ b/hw/ip/edn/dv/edn_sim_cfg.hjson
@@ -95,6 +95,7 @@
       name: edn_alert
       uvm_test: edn_alert_test
       uvm_test_seq: edn_alert_vseq
+      reseed: 200
     }
 
     {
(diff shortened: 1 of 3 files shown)


vogelpi:
This resolves #23526.

vogelpi:
Just in case: please wait with merging this. I am not yet fully done with the DV side for this and would like to run some more tests. Just to be sure. I'll post an update later today.

moidx (approved):

vogelpi:
I've now updated the PR to also fix the updating of the hw_cmd_sts register (this is currently broken on master as additional data words sent to CSRNG update the cmd_type field with invalid data).

I've spent quite some time verifying this but the DV isn't ready yet. I'll file a separate PR tomorrow for the DV side. A block-level regression with a reduced number of seeds did not lead to test failures except for the edn_alert sequence which I am still adjusting. The sequences not triggering error cases all keep passing.

vogelpi:
CI passes except for `boot_policy_flash_ecc_error:a_valid_b_corrupt_manifest_security_version_fpga_cw310_rom_with_fake_keys` which I've seen also failing for other unrelated PRs. I will now push a small DV update to push up the FSM coverage and will then hit merge.

PR #23608, opened 2024-06-10
[test] Fix up spi_passthru test to check WIP

a-will:
test_write_status was missing the busy bit check, as there is no SpiFlash helper method for WriteStatus1.


a-will:
This became _required_ (for realz) after https://github.com/lowRISC/opentitan/pull/23546. Incidentally, those failures also gave us coverage of the command blocking, hehe.

cfrantz (approved):

pamaury (approved):

PR #23610, opened 2024-06-10
[dv/tlt] Fix power virus test.

moidx:
1. Add fast sim and rng_srate +plusargs to speed up simulation time. This may be removed by the backend integrator as needed.
2. Increase SW timeout.
3. Update entropy complex configuration to force more operations.

### DIFF
--- a/sw/device/tests/power_virus_systemtest.c
+++ b/sw/device/tests/power_virus_systemtest.c
@@ -118,8 +118,8 @@ enum {
    */
   kEdn0SeedMaterialNumWords = 0,
   kEdn1SeedMaterialNumWords = 12,
-  kEdn0ReseedInterval = 128,
-  kEdn1ReseedInterval = 32,
+  kEdn0ReseedInterval = 64,
+  kEdn1ReseedInterval = 16,
   /**
    * KMAC parameters.
    */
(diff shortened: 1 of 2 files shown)


moidx:
Hi @hcallahan-lowrisc, the test is now stuck because it seems the I2C agent is not reading all the words expected by the test. 

https://github.com/moidx/opentitan/blob/ca79273324b3e0b846e49fabce5686fc8be7f083/sw/device/tests/power_virus_systemtest.c#L1373-L1380

It is possible that this test has missed a few I2C required changes. Can you help us take a look? 

You may be much quicker in figuring out what changes are needed to unblock the test. Please feel free to take my commits and submit a new PR. 

Thanks!

hcallahan-lowrisc:
> It is possible that this test has missed a few I2C required changes. Can you help us take a look?

I'm looking into this now. Thanks

sw/device/tests/power_virus_systemtest.c
```diff
@@ -1273,9 +1284,18 @@ static void max_power(void) {
 
   // Request entropy during max power epoch. Since AES is so fast, realistically
   // we will only be able to request a single block of entropy.
+  csrng_wait_ready();
   mmio_region_write32(csrng.base_addr, CSRNG_CMD_REQ_REG_OFFSET,
                       csrng_instantiate_cmd_header);
 
+  uint32_t csrng_reseed_cmd = csrng_cmd_header_build(
+      kCsrngAppCmdReseed, kDifCsrngEntropySrcToggleEnable, /*cmd_len=*/0,
+      /*generate_len=*/0);
```
vogelpi, line 1293: For a moment I thought we maybe should build this command header outside of the max power epoch and just write it to CSRNG here. But then I noted that we trigger an instantiate with real entropy just before. So operations actually overlap which is nice.

vogelpi (approved):
LGTM, thanks @moidx !

moidx:
Thanks @vogelpi.

@hcallahan-lowrisc, I will merge this. You can submit the I2C fixes in a follow up PR. Thanks!

PR #23624, opened 2024-06-11
[dv,usbdev] Tidy up test plan and usb20_item handling

alees24:
This PR just collects a few uncontentious tidy ups and introduces constructors for use in subsequent code and to simplify/tidy the existing code in time. Presently objects are created and subsequently populated, which results in a lot of error-prone replication of code.


rswarbrick (approved):
Much nicer!

PR #23632, opened 2024-06-11
[opentitantool] Allow manifest updates with ELF info

jon-flatley:
Adds the --elf flag to the `image manifest update` subcommand that takes the corresponding ELF file for the binary being modified as input. When supplied, opentitantool will populate the `code_start`, `code_end` and `entry_point` fields in the manifest.

This is useful when building tock firmware, since defining the manifest as a static struct as in the ROM_EXT doesn't work as nicely in rust.


jon-flatley:
Closing this as obsolete since #23727 solves this by populating the manifest fields in the linker script.

PR #23634, opened 2024-06-11
[rom] Rewrite the flash exception handler 

cfrantz:
1. Use registers `a0-a3` instead of `t0-t3` to permit compressed
   instruction encodings in the exception handler itself.
2. Re-arrange the exception logic to check mstatus and the flash
   controller fault status first.
3. Decode the faulting instruction.  If the instruction is a load,
   extract the destination register.  Load all-ones into the
   destination register before returning from the exception.


cfrantz:
#23635 is a better solution.

PR #23640, opened 2024-06-12
[dv] Tweaks to improve results with sec_cm FI tests

rswarbrick:
There are detailed notes in the respective commits, but the basic idea is that these are tweaks that are needed if the FI test causes the device to start doing silly things on the TL bus.

### DIFF
--- a/hw/dv/sv/tl_agent/tl_host_driver.sv
+++ b/hw/dv/sv/tl_agent/tl_host_driver.sv
@@ -185,8 +185,6 @@ class tl_host_driver extends tl_base_driver;
     tl_seq_item rsp;
 
     forever begin
-      bit req_found;
-
       if ((cfg.vif.host_cb.d2h.d_valid && cfg.vif.h2d_int.d_ready && !reset_asserted) ||
           ((pending_a_req.size() != 0) & reset_asserted)) begin
         // Use the source ID to find the matching request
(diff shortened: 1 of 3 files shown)


nasahlpa (approved):
Thanks for working on this, LGTM!

I guess we still need manually list all the assertoffs in  [sec_cm_fI_ctrl_svas](https://github.com/lowRISC/opentitan/blob/d68f6beae7e0fb1ec6cbeb50510b47625a54e3f7/hw/ip/otbn/dv/uvm/env/seq_lib/otbn_common_vseq.sv#L153)? Not sure about OTBN but other modules use TL-UL assertions in the top level module.

hw/dv/sv/tl_agent/tl_host_driver.sv
```diff
@@ -208,16 +206,22 @@ class tl_host_driver extends tl_base_driver;
...
+        // If there was a matching request, we responded to it above. If not, the device seems to
+        // have sent a response without a request (and we won't have done anything yet).
+        //
+        // If we're in reset, we might have forgotten the request (and can ignore the response). If
+        // not, we wouldn't normally expect this to happen. This is a property that is checked in
+        // the tlul_assert module, which should be bound in. But it *might* happen if we're doing
+        // fault injection and disabling assertions in that module.
+        //
+        // Ignore the response either way: we're a driver and there is definitely no sequence that
+        // is waitign for the response here. If there's a bug in the design and we're generating
```
vogelpi, line 225: Nit: waiting

vogelpi (approved):
This looks good to me, thanks @rswarbrick !

PR #23650, opened 2024-06-13
[hjson,englishbreakfast] Do not declare aon_timer as templated

### LINKED ISSUE #23649: [tools,englishbreakfast] Do not tag aon_timer_aon as "templated" for englishbreakfast
### Description

Running topgen.py with the `--check-cm` flag causes countermeasures checks for a given top, and the checks are different for templated and all other modules. For templated modules the rtl files are collected differently, and the checks can fail if modules are erroneously tagged as templated.

The aon_timer_aon module is tagged incorrectly in https://github.com/lowRISC/opentitan/blob/master/hw/top_englishbreakfast/data/top_englishbreakfast.hjson#L321, and this causes countermeasure failures for englishbreakfast.

matutem:
The aon_timer module is a regular IP block, so it should not be marked as `attr: "templated"`, since that causes a failure in the countermeasures check.

Fixes #23649

### DIFF
--- a/hw/top_englishbreakfast/data/top_englishbreakfast.hjson
+++ b/hw/top_englishbreakfast/data/top_englishbreakfast.hjson
@@ -318,7 +318,6 @@
       reset_connections: {rst_ni: "sys_io_div4", rst_aon_ni: "sys_aon"},
       domain: ["Aon"],
       base_addr: "0x40470000",
-      attr: "templated",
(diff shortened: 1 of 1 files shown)


moidx (approved):

vogelpi (approved):
Thanks for fixing this @matutem !

PR #23653, opened 2024-06-13
[rom] Make the flash exception handler OTP controlled 

cfrantz:
Since the flash ECC exception handler is a complex bit of low-level
code, mitigate risk by enabling or disabling the handler via OTP.

1. Enable the handler in the default OTP configurations.
2. Test at exception entry whether or not we want to handle ECC errors.
3. Add a test that confirms that when flash ECC exception handling is
   disabled, an ECC error results in a classic Load Access Fault.

### DIFF
--- a/sw/device/silicon_creator/rom/rom_isrs.S
+++ b/sw/device/silicon_creator/rom/rom_isrs.S
@@ -4,6 +4,7 @@
 
 #include "hw/top_earlgrey/sw/autogen/top_earlgrey_memory.h"
 #include "sw/device/lib/base/macros.h"
+#include "sw/device/lib/base/hardened_asm.h"
 #include "flash_ctrl_regs.h"
 
 .equ LOAD_ACCESS_FAULT, 5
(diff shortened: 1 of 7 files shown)


sw/device/silicon_creator/rom/rom_isrs.S
```diff
@@ -25,15 +26,25 @@
...
   // Save `sp` into `mscratch` and use the 128 bytes reserved at the top of
   // RAM as the exception frame.
+  csrw mscratch, sp
+
+  // Before saving the stack frame, check if handling of flash exceptions is
+  // enabled.  Determine this with just the stack pointer so that if not enabled,
+  // we don't have to do any other register save/restore.
+  la   sp, flash_ecc_exc_handler_en
+  lw   sp, 0(sp)
+  xori sp, sp, HARDENED_BOOL_TRUE
```
moidx, line 42: For hardening, this will help the attacker bypass this check easier than to force it to be enabled. I think that is probably what we want in this case as disabling this handler via faults will not result in a weaker security.

moidx (approved):

sw/device/silicon_creator/rom/rom_isrs.S
cfrantz: That's what I was thinking as well.

PR #23656, opened 2024-06-13
[csrng/dv] Add the right err_test codes to fatal errors

h-filali:
For each error there is a corresponding error test signal. In the error test sequence not all of these error test signals were expected to cause a fatal error even though they do in the normal case. This commit adds for each fatal error, the corresponding err_test to the list of fatal errors.


vogelpi (approved):
This looks good to me, this seems to be a remainder of #23561. Thanks for fixing this @h-filali !

vogelpi:
I am merging this, all relevant CI checks passed already. The FPGA tests are not relevant for this DV-only PR.

h-filali:
Thanks a lot @vogelpi !

PR #23657, opened 2024-06-13
[csrng/rtl] Make the main SM error test signal fatal

h-filali:
The main SM error test signal is currently the only test signal where the actual error would be fatal but the test signal is not. This commit make adds the err_code_test_bit for the main SM error to fatal_loc_events to make the main SM error test fatal.

### DIFF
--- a/hw/ip/csrng/rtl/csrng_core.sv
+++ b/hw/ip/csrng/rtl/csrng_core.sv
@@ -495,14 +495,16 @@ module csrng_core import csrng_pkg::*; #(
 
   // Counter and FSM errors are structural errors and are always active regardless of the
   // functional state. main_sm_err_sum is not included here to prevent some tools from
-  // inferring combo loops.
+  // inferring combo loops. However, to include the state machine error for testing,
+  // we use the error code test bit (index 21) directly.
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
Thanks for flagging this @h-filali . This looks good to me!

The fix is very small and fully aligns EDN, CSRNG and ENTROPY_SRC with respect to error signaling. I think we should absorb this change but we'll need to discuss this with other leads.

vogelpi:
CHANGE AUTHORIZED hw/ip/csrng/rtl/csrng_core.sv

h-filali:
> Is there a reason not to use `main_sm_err_sum` here? That way, we don't have the magic "21" index in two different places.

Yes, there is a reason. `main_sm_err_sum` causes a loop, since it is set to high by the error signal that comes out of the main SM. And if `main_sm_err_sum` is high, it in turn causes the main SM error signal to go high. There is a comment above the lines I changed which tries to explain this. Should I maybe make the comment more clear?

rswarbrick:
CHANGE AUTHORIZED: hw/ip/csrng/rtl/csrng_core.sv

PR #23661, opened 2024-06-13
[flash/dv] Fix flash_ctrl_base_otf_vseq address overflow

eshapira:
This fixes some rare failures in test flash_ctrl_fs_sup caused by a possible overflow of the randomized operation by randomizing an address and number of words that exceeding the info partition size.

### DIFF
--- a/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_otf_base_vseq.sv
+++ b/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_otf_base_vseq.sv
@@ -128,8 +128,10 @@ class flash_ctrl_otf_base_vseq extends flash_ctrl_base_vseq;
     // transaction creates prog_win error.
     // To prevent that, make full size access always start from address[2:0] == 0.
     if (fractions == 16) rand_op.addr[2] == 0;
+    // Make sure that the operation will not cross the selected partition boundries.
     if (rand_op.partition != FlashPartData) {
-      rand_op.addr inside {[0:InfoTypeBytes[rand_op.partition>>1]-1]};
+      rand_op.addr inside
+        {[0:InfoTypeBytes[rand_op.partition>>1]-(FlashBankBytesPerWord*fractions)]};
       rand_op.prog_sel == 1;
     } else {
       rand_op.prog_sel == 0;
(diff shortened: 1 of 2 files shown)


hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_otf_base_vseq.sv
```diff
@@ -128,10 +128,13 @@ class flash_ctrl_otf_base_vseq extends flash_ctrl_base_vseq;
     // transaction creates prog_win error.
     // To prevent that, make full size access always start from address[2:0] == 0.
     if (fractions == 16) rand_op.addr[2] == 0;
+    // Make sure that the operation will not cross the selected partition boundries.
     if (rand_op.partition != FlashPartData) {
-      rand_op.addr inside {[0:InfoTypeBytes[rand_op.partition>>1]-1]};
+      rand_op.addr inside
+        {[0:InfoTypeBytes[rand_op.partition>>1]-1-(FlashBankBytesPerWord*fractions)]};
```
rswarbrick, line 134: I think this might still have a stray `-1`? (If blocks are 10 bytes and `InfoTypeBytes` is 100, I'd expect the maximum address to be 90, not 89)

hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_otf_base_vseq.sv
```diff
@@ -128,10 +128,13 @@ class flash_ctrl_otf_base_vseq extends flash_ctrl_base_vseq;
...
       rand_op.prog_sel == 1;
     } else {
+      rand_op.addr inside {[0:FlashSizeBytes-1-(FlashBankBytesPerWord*fractions)]};
```
rswarbrick, line 137: Is there a reason that we're getting a new constraint on the address in this case? I'm slightly surprised that it's needed.

eshapira: I guess it isn't really necessary. I added it because I felt it needed, but I don't really had issues with it so I'll remove it.

hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_otf_base_vseq.sv
eshapira: I think you're right, I'll fix it. Thanks!

rswarbrick (approved):
Looks good to me: thanks!

PR #23666, opened 2024-06-13
[otbn,dv] Restructure end of wipe round in sim.py

rswarbrick:
This should behave the same as the code that was there before, except that we no longer disable URND as we enter RMA (but this doesn't matter because we're going to lock anyway).

This makes the case analysis slightly simpler and there's now a comment that explains it clearly.

(This is motivated by me trying to tidy up the model's handling of escalation requests. This needed me to tidy up how the `rma_req` signal got passed around, which needed me to read this reasonably carefully).


vogelpi (approved):
LGTM!

PR #23667, opened 2024-06-13
[dv,usbdev] Add three link-related sequences

alees24:
This PR combines three small, similar, link-related sequences:

- host_lost signaling when there has been no SOF packet received for some time, despite there being USB activity.
- link reset signaling; check that software is notified of even short Bus Reset intervals.
- resume link active; important from resuming from Deep Sleep, this software-mediated change in the link state machine permits a return to Resuming and then Active after the device has powered up again.

As part of this PR, in order to facilitate the 'resume link active' sequence, the driver no longer _drives to Idle_ but rather releases the bus, as per the real USB, and the monitor is now able to cope with Resume Signaling that occurs without a full Suspend Signaling period, by bailing from packet collection if K state has been observed for too long.


hw/ip/usbdev/dv/env/seq_lib/usbdev_link_reset_vseq.sv
```diff
@@ -0,0 +1,30 @@
...
+  virtual task body();
+    clear_all_interrupts();
+    // Enable the interrupt to make it visible at the pin too.
+    csr_wr(.ptr(ral.intr_enable), .value(1 << IntrLinkReset));
+
+    // Leave the link in an Idle state for a while, to make the signaling changes clearer.
+    cfg.clk_rst_vif.wait_clks(1000);
+
+    // Perform a Bus Reset that lasts just 10 microseconds; the specification requires at least
+    // 10ms of reset signaling, but the device is permitted to respond after just 2.5 microseconds.
```
rswarbrick, line 20: I'd expand this comment slightly to add "... and we expect usbdev to do so."

rswarbrick (commented):
These are some comments on the first commit so far. It looks sensible, thanks.

One extra comment is that I wonder whether the `host_lost` testpoint description needs a slight tweak. When I first read the second bullet point, I thought it was saying that we were waiting exactly 4.096ms, which isn't quite right. Maybe something like "Wait for more than 4.096ms without sending a SOF from the host".

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -276,8 +276,8 @@ class usb20_monitor extends dv_base_monitor #(
     m_sof_pkt.valid_eop = valid_eop;
     m_sof_pkt.valid_stuffing = valid_stuffing;
     if (destuffed_packet.size() == 32) begin
-      m_sof_pkt.framenum = {<<{destuffed_packet[18:8]}};
-      m_sof_pkt.crc5 = {<<{destuffed_packet[7:3]}};
+      m_sof_pkt.framenum = {<<{destuffed_packet[8:18]}};
```
rswarbrick, line 295: This switch around is perfectly reasonable, but I'm a bit surprised! Is it just a style thing, or do you expect it to have any effect?

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -196,22 +201,33 @@ class usb20_monitor extends dv_base_monitor #(
...
     bit valid_stuffing;
     usb_symbol_e sym;
     bit valid_eop;
     bit packet[$];
 
+    resuming = 1'b1;
+
     // We have already detected the first K state of the SYNC signaling.
     packet.push_back(1'b0);
     // Keep collecting symbols until we detect something that is neither J nor K.
```
rswarbrick, line 220: Maybe this comment needs slightly generalising?

rswarbrick (commented):
A couple of nitty comments, but this looks good to me (commits 2,3).

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -195,6 +195,24 @@ endtask
     end
   endtask
 
+  // Wait for the DUT to assert one of the specified interrupts, with an optional timeout
+  //   (the default of -1 disables the timeout).
+  // Then optionally clear all of the interrupt that matched.
+  virtual task wait_for_interrupt(bit [31:0] snapshot, int timeout_clks = -1, bit clear = 1,
```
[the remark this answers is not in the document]
alees24: Oh, rather than invent a new name I've borrowed from the software/DIF world where a set of asserted interrupts is called a 'snapshot.'

hw/dv/sv/usb20_agent/usb20_monitor.sv
alees24: As discussed, the previous expression produced a warning in simulation.

alees24:
Superseded by PR #23671

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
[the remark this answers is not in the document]
rswarbrick: Sounds cool. Sorry, I don't think I new that use of the name. Thanks for the explanation.

PR #23671, opened 2024-06-13
[dv,usbdev] Add three FIFO interrupt-related sequences

alees24:
This PR adds three small FIFO interrupt-related sequences:

- av_empty checks the 'av_setup_empty' and 'av_out_empty' interrupts that shall be asserted when the corresponding FIFO is empty.
- av_overflow checks that an overflow of either Av Setup FIFO or Av OUT FIFO leads to an 'av_overflow' interrupt.
- rx_full fills the RX FIFO with SETUP packets and checks that the rx_full interrupt becomes asserted.

This PR builds on #23667 and presently includes it.


hw/ip/usbdev/dv/env/seq_lib/usbdev_av_empty_vseq.sv
```diff
@@ -0,0 +1,71 @@
...
+  endtask
+
+  task check_level(int unsigned exp_level);
+    uvm_reg_data_t level;
+    if (setup) csr_rd(.ptr(ral.usbstat.av_setup_depth), .value(level));
+    else csr_rd(.ptr(ral.usbstat.av_out_depth), .value(level));
+    `DV_CHECK_EQ(level, exp_level, "Unexpected FIFO level")
+  endtask
+
+  task check_intr_bit(bit exp, string msg);
```
rswarbrick, line 26: Nit: Maybe it's worth calling this `check_intr_state`? I got a bit confused when reading this because "intr_bit" is an index a few lines later.

rswarbrick (commented):
Just looking at the final commit, this looks good to me.

hw/ip/usbdev/dv/env/seq_lib/usbdev_av_empty_vseq.sv
alees24: I've renamed it to 'check_interrupt_state.'

hw/ip/usbdev/dv/env/seq_lib/usbdev_av_empty_vseq.sv
```diff
@@ -0,0 +1,71 @@
...
+
+  virtual task body();
+    uvm_reg_data_t intr_bit = 1 << (setup ? IntrAvSetupEmpty : IntrAvOutEmpty);
+    int unsigned depth = setup ? AvSetupFIFODepth : AvOutFIFODepth;
+
+    clear_all_interrupts();
+    // Enable the appropriate Av Empty interrupt.
+    csr_wr(.ptr(ral.intr_enable), .value(intr_bit));
+    check_level(0);
+
```
marnovandermaas, line 41: You could check interrupt is off here right?

marnovandermaas (approved):
Looks reasonable to me after just doing a code inspection.

hw/ip/usbdev/dv/env/seq_lib/usbdev_av_empty_vseq.sv
alees24: That is checked a few lines below, just before the USB transaction that should cause the FIFO to become empty, raising the interrupt.

PR #23676, opened 2024-06-13
Restore template build targets

a-will:
Restore BUILD file templates for clkmgr, flash_ctrl, pwrmgr, and rstmgr.
Properly give them .tpl suffixes so the template files are not picked up
by bazel.

Some new testplan items were missing a few fields in the rom_manual_testplan. Add them.


moidx (approved):

a-will:
Well, I messed something else up, apparently :rofl:

hw/top_earlgrey/ip_autogen/clkmgr/BUILD
```diff
@@ -6,7 +6,5 @@ package(default_visibility = ["//visibility:public"])
 
 filegroup(
     name = "all_files",
-    srcs = glob(["**"]) + [
-        "//hw/ip_templates/clkmgr/data:all_files",
-    ],
+    srcs = glob(["**"]) + ["data:all_files"],
```
a-will, line 9: We don't usually do labels with relative paths for anything outside the current directory. Let me know if you object and would prefer we do this differently!

a-will: Oh, I guess this is invalid. That's weird: I could've sworn I was able to write these in the past. And it builds locally /shrug

a-will:
Okay, changed things up: Now there is only the one BUILD.tpl file at the root of the IP template. All output files are then under the same package, and I've fixed up the labels representing the old targets for downstream targets.

a-will:
...and now fixed a full-path reference to an autogenerated regs file in ft_personalize.c. The current guidance is to add the directory for each generated file to the includes, then reference the file by just the basename.

a-will:
I'll let this CI run go through, then correct the one lint error. Buildifier took care of the issues with the generated BUILD file, but I forgot to bring the changes back to the BUILD.tpl file, leading to the lint error representing out-of-date generated files.

a-will:
//sw/device/tests:spi_device_sleep_test_fpga_cw310_sival_rom_ext seems to be some flaky test. It passes locally... but failed on the latest master commit.

a-will:
Merging, since the tests passed on the last commit (except the one flaky test), and the only change was deleting a blank line to satisfy in-depth line (which now passes).

PR #23682, opened 2024-06-14
[dv,flash_ctrl] Avoid double writes in flash_ctrl_rw_evict_vseq

### LINKED ISSUE #23681: [dv,flash_ctrl] Avoid double writes in flash_ctrl_rw_evict
### Description

When a flash word is updated after a prior write it is highly likely the updated data is not as intended because a 0 cannot be updated to 1, and will cause integrity errors when it is read. The set of candidate addresses for eviction need to be used only once, so it must be repopulated as the test progresses.

matutem:
The addresses to be programmed cannot repeat to avoid integrity errors.
Fix doc or comment typos in a separate independent commit.

Fixes #23681


vogelpi (approved):
Thanks @matutem , this looks good to me!

PR #23687, opened 2024-06-14
[hw,dma,rtl] Only set error code when entering the error state

Razer6:
Previously, the error code is written when entering the error state AND while being in there. Since the error code value is only availble in the transition into the error state, and further update overwrites the error code again with zero.

This change ensures that the error code only gets update in the transition into the error state, where the correct value is available


alees24 (approved):
Looks good.

andreaskurth (approved):

andreaskurth:
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma.sv

PR #23689, opened 2024-06-14
[otbn,dv] Avoid only writing zeros to FIFO memory before FI tests

rswarbrick:


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #23692, opened 2024-06-14
[usbdev] Tidy DP/DN output state when undriven

alees24:
Correct the handling of link reset to prevent it from inverting the output signals and to avoid leaving the signals inverted between the bus reset and the first transmission. Behavior should be unaltered on physical devices provided that the driver enable is honored.

Background: unexpected transitions of DP/DN were observed in block-level simulation relating to Bus Reset.

### DIFF
--- a/hw/ip/usbdev/rtl/usb_fs_tx.sv
+++ b/hw/ip/usbdev/rtl/usb_fs_tx.sv
@@ -402,14 +402,13 @@ module usb_fs_tx (
       end else begin
         usb_d_d = !usb_d_q;
       end
-
-      // Set to J state when OE=0 to avoid
-      // glitches
-      if (!oe_d) begin
-        usb_d_d = 1;
-      end
     end
 
+    // Set to J state when OE=0 to avoid
+    // glitches
+    if (!oe_d) begin
+      usb_d_d = 1;
+    end
   end
 
   always_ff @(posedge clk_i or negedge rst_ni) begin : proc_diff_reg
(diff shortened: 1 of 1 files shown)


alees24:
@a-will Re the placement of the 'oe' clause to achieve J state when the driver is not enabled, I think the altered placement was probably the real intent?

a-will (approved):
Whoops, looks like I wrote K instead of J in a few places. This LGTM, but as the RTL is now frozen, we'd have to seek authorization for the merge.

Because it doesn't appear like the issue addressed by these changes actually results in significant harm, these might end up being held back until the current earlgrey tape-out project gets its own git branch. (If an ECO is required, we'd have to do some paperwork and explain why this needs to be accepted.)

a-will:
> @a-will Re the placement of the 'oe' clause to achieve J state when the driver is not enabled, I think the altered placement was probably the real intent?

I think you are probably right about that, yes.

alees24:
Sure. Thanks for the response.

a-will:
CC @vogelpi @andreaskurth @moidx for decisions on this RTL change.

moidx (approved):

vogelpi (approved):
This looks good to me, thanks @alees24 for the fix and @a-will for the review.

vogelpi:
CHANGE APPROVED: hw/ip/usbdev/rtl/usb_fs_tx.sv

moidx:
CHANGE AUTHORIZED: hw/ip/usbdev/rtl/usb_fs_tx.sv

PR #23669, opened 2024-06-13
[dv,usbdev] Additional test sequences

alees24:
This PR adds test sequences to cover the following test points:

- link_resume
- phy_config_tx_osc_test_mode
- device_address
- disable_endpoint

It also updates av_buffer_seq following the division into the Available Buffer FIFO into two separated FIFOs for OUT and SETUP packets. It now exercises the Av SETUP FIFO too.

~~This PR is now based on #23671~~

PR #23693, opened 2024-06-14
[dv,usbdev] Usbdev setup priority sequence

alees24:
This PR adds a single sequence in the final commit:

- usbdev_setup_priority checks that the final slot of the RX FIFO accepts only SETUP packets and that OUT packets are NAKed. This is important in keeping the USBDEV always responsive to Control Transfers even under heavy traffic and/or high software latency.

~~_This PR is based upon and includes #23669 because it uses some common code, but **only the final commit is relevant.**_~~


rswarbrick (commented):
I've looked carefully at the last commit. Looks sensible to me!

Issue #22833, opened 2024-04-25
[component] Issues with Getting started


julienissuts:
### Description

Hello,
In the last couple of days I followed the getting started guide to simulate the opentitan design on Verilator.
I followed the instructions of https://opentitan.org/book/doc/getting_started/index.html (Getting started guide)
And then the Verilator guide: https://opentitan.org/book/doc/getting_started/setup_verilator.html (Verilator Setup)
I came to the chapter “Use Bazel to run software on a verilator simulation” and there I ran into some issues.

Firstly I had to install Bazel by myself, because there were no instructions how to install it and you never mentioned it before. I thought this was a little bit strange because the installation of the other tools are good documented, why not Bazel? 

Then I installed Bazelisk and tried to continue, but from the terminal output I figured out that fusesoc is necessary. I found that very strange because you didn’t document the installation of fusesoc neither. So I installed it and tried it again.

I installed Bazelisk as described on: https://github.com/bazelbuild/bazelisk
And fusesoc like described with: pip3 install --upgrade --user fusesoc

If I run :

> cd $REPO_TOP
> bazel test --test_output=streamed //sw/device/tests:uart_smoketest_sim_verilator

I get following error:

_

> _…
>   File "/home/JStuessi/.cache/bazel/_bazel_JStuessi/babbdb84fbabb5f19400ee8587d81374/sandbox/linux-sandbox/2/execroot/lowrisc_opentitan/bazel-out/k8-opt-exec-2B5CBBC6-ST-e846b08c7501/bin/util/fusesoc_build.runfiles/ot_python_deps/pypi__fusesoc/fusesoc/config.py", line 85, in __init__
>     os.makedirs(self.cache_root, exist_ok=True)
>   File "/home/JStuessi/.cache/bazel/_bazel_JStuessi/babbdb84fbabb5f19400ee8587d81374/execroot/lowrisc_opentitan/external/python3_x86_64-unknown-linux-gnu/lib/python3.9/os.py", line 225, in makedirs
>     mkdir(name, mode)
> OSError: [Errno 30] Read-only file system: '/home/JStuessi/.cache/fusesoc'
> INFO: Elapsed time: 25.326s, Critical Path: 5.24s
> INFO: 179 processes: 178 internal, 1 linux-sandbox.
> FAILED: Build did NOT complete successfully
> //sw/device/tests:uart_smoketest_sim_verilator                  FAILED TO BUILD
> 
> Executed 0 out of 1 test: 1 fails to build.

[...]

PR #23698, opened 2024-06-15
[rules] Specify fusesoc cache dir to be in the bazel tree

### LINKED ISSUE #23696: Issue with Building (and Testing) Software: Verilator Test Read-only File System Error
### Description

I am trying to run the Verilator test (./bazelisk.sh build //hw:verilator) but keep getting a Read-only file system error from the /home/dev/.cache/fusesoc directory. I have done the setup as described in the Getting Started section using both an Ubuntu 20.04 VM and the Docker image and have encountered the same error using both methods.  I have also tried setting my .cache directory permissions to 777 and still encounter the error.

Has anyone else also encountered this error, and if so, do you have a solution?

image

a-will:
Make the fusesoc cache dir be in the bazel tree, so fusesoc doesn't attempt to create directories outside, such as ~/.cache/fusesoc. That action can cause build failures, since fusesoc is sandboxed.

Fixes #23696

### DIFF
--- a/rules/fusesoc.bzl
+++ b/rules/fusesoc.bzl
@@ -31,7 +31,15 @@ def _fusesoc_build_impl(ctx):
     flags = [ctx.expand_location(f, ctx.attr.srcs) for f in ctx.attr.flags]
     outputs = [out_dir]
     groups = {}
+
+    cache_dir = "{}/fusesoc-cache".format(out_dir.path)
+    cfg_file_path = "build.{}.fusesoc_config.toml".format(ctx.label.name)
+    cfg_file = ctx.actions.declare_file(cfg_file_path)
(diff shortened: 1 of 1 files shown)


jwnrt (approved):
I think this would fix https://github.com/lowRISC/opentitan/issues/22833 and a bug that @AlexJones0 was having. It LGTM, was there something you wanted to change before marking as ready?

a-will:
> I think this would fix #22833 and a bug that @AlexJones0 was having. It LGTM, was there something you wanted to change before marking as ready?

It was merely marked draft because of the quiet period (since it changes something with tool flows, though it _should_ represent a fix, hehe).

a-will:
I've marked it ready, since the development is ostensibly done, but do check that the merge window for such changes is open before merging.

PR #23699, opened 2024-06-16
[flash/dv] Increase alert timeout in wr_intg test

eshapira:
Increase the timeout for the alert in the flash_ctrl_wr_intg, to make sure it will be enough for the slower extending enivronments.


eshapira:
Hi, can you please review this PR?

rswarbrick (approved):
Sorry for not reviewing this more quickly. It seems reasonable to me, but I'm definitely not an expert on `flash_ctrl`.

I'm approving the change (since there doesn't seem to be any risk in making it). But @matutem: could you take a quick look and check this matches what you'd expect?

PR #23701, opened 2024-06-17
[lint] Waive some more uncritical lint errors

vogelpi:


hw/ip/prim/lint/prim_lfsr.waiver
```diff
@@ -5,5 +5,5 @@
 # waiver file for prim_lfsr
 
 # TODO: this lint waiver may not be necessary in future AscentLint versions. check and remove if possible.
-waive -rules {LOOP_VAR_OP} -location {prim_lfsr.sv} -msg {Loop variable 'k' is in arithmetic expression '(k + shift) % gen_out_non_linear.NumSboxes' with non-constant terms} \
+waive -rules {LOOP_VAR_OP} -location {prim_lfsr.sv} -regexp {Loop variable 'k' is in arithmetic expression '\(k \+ shift\) % (gen_out_non_linear.|)NumSboxes' with non-constant terms} \
```
rswarbrick, line 8: Definitely not a problem, but I'm surprised this syntax is required by AscentLint. Don't they support `(optional_thing)?` ?

rswarbrick (approved):

vogelpi:
Thanks for your review @rswarbrick !

hw/ip/prim/lint/prim_lfsr.waiver
vogelpi: Ah, now I understand, thanks. This seems indeed to be working and I will do it this way next time.

Issue #23625, opened 2024-06-11
[spi_host] Allow SPI host tests to work with multiple flash chips


jwnrt:
### Description

The `spi_host_winbond_flash_test` checks the manufacturer and device IDs of a Winbond flash chip connected to the SPI host.

The device ID is hardcoded but the test should work with many kinds of flashes.

With the new CW340 FPGAs added to CI, we will need to add more flash chips. The kind we have are still manufactured by Winbond but have different IDs. We would like to remove the explicit device ID check from this test to allow it to work with our second Winbond flash part.

Opening this issue since it needs to wait until after the current quiet period.

a-will:
One way to do this would be to rely on the SFDP to configure the test (but report the JEDEC ID in the log, so you can identify the chip for failures).

General support for all the SFDP fields can be a semi-significant undertaking, though.

jwnrt:
> One way to do this would be to rely on the SFDP to configure the test (but report the JEDEC ID in the log, so you can identify the chip for failures).

Do you mean that the device ID is also in the SFDP and we can compare the JEDEC ID to that instead of our hardcoded value?

a-will:
> > One way to do this would be to rely on the SFDP to configure the test (but report the JEDEC ID in the log, so you can identify the chip for failures).
> 
> Do you mean that the device ID is also in the SFDP and we can compare the JEDEC ID to that instead of our hardcoded value?

No, I mean that the SFDP tells us how to drive the SPI flash chip (e.g. what the command opcodes are to do various things, whether a chip supports quad mode, how to enable, etc.).

A table of known JEDEC IDs isn't particularly scalable, so it might not make sense to check it. Instead, we'd know it should fail if you get all zeroes or all ones. We _do_ get the flash memory density from the SFDP, though, so maybe we could check one byte of the JEDEC ID.

jwnrt:
> No, I mean that the SFDP tells us how to drive the SPI flash chip (e.g. what the command opcodes are to do various things, whether a chip supports quad mode, how to enable, etc.).

@AlexJones0 and I checked the SFDP tables and it looks like you can get the op-code for quad reads but not writes from the standard tables. Our `sw/device/tests{/pmod}/spi_host_*_flash_test`s hard code these IDs, but reading them from the SFDP tables requires knowledge about the manufacturer-specific tables.

We could embed the tables for a few manufacturers and read the op-codes, but this doesn't help us much.

[...]

PR #23703, opened 2024-06-17
[sw] Remove `device_id` checks from spi host tests

jwnrt:
We would like more flexibility to change flash parts in CI without breaking these tests.

Removed the `device_id` check, but retain the manufacturer ID check.

I have not checked for `0x00` or `0xff` as signs of failure since these are legitimate device IDs.

### DIFF
--- a/sw/device/tests/spi_host_flash_test_impl.c
+++ b/sw/device/tests/spi_host_flash_test_impl.c
@@ -65,12 +65,9 @@ status_t test_read_sfdp(dif_spi_host_t *spi) {
   return OK_STATUS();
 }
 
-status_t test_read_jedec(dif_spi_host_t *spi, uint16_t device_id,
-                         uint8_t manufacture_id) {
+status_t test_read_jedec(dif_spi_host_t *spi, uint8_t manufacture_id) {
   spi_flash_testutils_jedec_id_t jdec;
   TRY(spi_flash_testutils_read_id(spi, &jdec));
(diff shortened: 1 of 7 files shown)


sw/device/tests/spi_host_flash_test_impl.c
```diff
@@ -65,12 +65,9 @@ status_t test_read_sfdp(dif_spi_host_t *spi) {
   return OK_STATUS();
 }
 
-status_t test_read_jedec(dif_spi_host_t *spi, uint16_t device_id,
-                         uint8_t manufacture_id) {
+status_t test_read_jedec(dif_spi_host_t *spi, uint8_t manufacture_id) {
   spi_flash_testutils_jedec_id_t jdec;
   TRY(spi_flash_testutils_read_id(spi, &jdec));
-  TRY_CHECK(jdec.device_id == device_id, "Expected %x, got %x!", device_id,
```
engdoreis, line 72: Could you instead check that device is neither 0x00 nor 0xff?

engdoreis (approved):
Besides my comment below, it LGTM?

sw/device/tests/spi_host_flash_test_impl.c
engdoreis: I see.

cfrantz (approved):

engdoreis:
Fix https://github.com/lowRISC/opentitan/issues/23625

Issue #19033, opened 2023-06-23
[csrng/dv] Tweak coverage configuration to close branch and expression coverage


andreaskurth:
Branch and expression coverage are reported as <90% for CSRNG. Examining the [coverage report](https://reports.opentitan.org/hw/ip/csrng/dv/latest/cov_report/index.html), however, one finds that the block coverage (which seems to include branch coverage) and expression coverage are well above 90% for all CSRNG RTL modules. In fact, the only CSRNG RTL modules that don’t have 100% block coverage are `csnrg_ctr_drbg_upd` and `csrng_main_sm`, but even those have a block coverage of >98%.

We should tweak the coverage tool configuration to report those higher numbers, which are more representative of CSRNG coverage than the currently reported overall numbers.

vogelpi:
Moving to M4. We need this to declare V2(S) which is required for M4 already.

vogelpi:
This is a duplicate of #19033.

vogelpi:
The work has been started but @h-filali is OoO this week. To be continued in M5.

PR #23707, opened 2024-06-17
[csrng/dv] Exclude AES core from coverage analysis

h-filali:
This commit excludes the AES cipher core from the coverage analysis. The exlusions file had to be revised since our coverage plummeted even though the coverage for CSRNG core was close to 100%. In the past we excluded a lot of paths from the AES core individually. However, excluding the AES core all together might be a more honest and transparent approach, since we now also exclude the paths where we have 100% coverage inside the AES core.

This resolves [#19033](https://github.com/lowRISC/opentitan/issues/19033)

These are the results after a full regression:

|  Score  |  Block  |  Branch  |  Statement  |  Expression  |  Toggle  |   Fsm    |  Assertion  |  CoverGroup  |
|:-------:|:-------:|:--------:|:-----------:|:------------:|:--------:|:--------:|:-----------:|:------------:|                                                                                               
| 97.15 % | 98.23 % | 95.85 %  |   98.86 %   |   96.43 %    | 91.77 %  | 100.00 % |   97.14 %   |   90.60 %    |


hw/ip/csrng/dv/cov/csrng_v2s_exclusions.vRefine
```diff
@@ -5,663 +5,18 @@
...
...odule prim_secded_inv_39_32_dec...;;// Black-box DV CDC module.;deselect_coverage -betfs -module prim_cdc_rand_delay;;// csr_assert_fpv is an auto-generated csr read assertion module. So only assertion coverage is;// meaningful to collect.;deselect_coverage -betf -module *csr_assert_fpv...;select_coverage -assert -module *csr_assert_fpv;;// Only enable assertion coverage;deselect_coverage -betf -module *tlul_assert...;select_coverage -assert -module *tlul_assert;;// Only collect toggle coverage on the DUT and the black-boxed IP (above) ports.;deselect_coverage -toggle -module ${DUT_TOP}...;select_coverage -toggle -module ${DUT_TOP};;select_coverage -toggle -module prim_alert_sender;select_coverage -toggle -module prim_alert_receiver;select_coverage -toggle -module prim_count;select_coverage -toggle -module prim_esc_sender;select_coverage -toggle -module prim_esc_receiver;select_coverage -toggle -module prim_onehot_check;select_coverage -toggle -module prim_prince;select_coverage -toggle -module prim_lfsr;select_coverage -toggle -module prim_secded_inv_64_57_dec;select_coverage -toggle -module prim_secded_inv_39_32_dec;" path="/home/dev/src/hw/dv/tools/xcelium/cover.ccf"></ccf-file>
```
vogelpi, line 13: Do you know where this line is coming from? Maybe we better remove this if it's not strictly needed.

vogelpi (approved):
Thanks @h-filali for doing this. I think excluding the whole AES cipher core is okay as we sign off the AES cipher core as part of the unmasked AES core (including coverage).

hw/ip/csrng/dv/cov/csrng_v2s_exclusions.vRefine
h-filali: I thought I'd ignore it since it's just in the info branch. But you're right it's probably best to just remove it.

PR #23710, opened 2024-06-17
[rom] Implement and test MGF1 and SHA2-based thash.

jadephilipoom:

These are preliminary steps for SHA2-based SPHINCS+; essentially the changes I can make without actually flipping the switch on the whole implementation. I've added an MGF1 implementation and a SHA2 version of the SPHINCS+ `thash` operation, and tests for both that I created using a known-good third-party implementation.

As a side change, I also added `test_rom` targets for all SPHINCS+ subcomponent tests. I found this useful for debugging, especially since it might be helpful to test SPHINCS+ subcomponents on commits/targets where there are issues with the ROM boot flow.

### DIFF
--- a/sw/device/silicon_creator/lib/sigverify/sphincsplus/BUILD
+++ b/sw/device/silicon_creator/lib/sigverify/sphincsplus/BUILD
@@ -69,7 +69,6 @@ cc_library(
     ],
 )
 
-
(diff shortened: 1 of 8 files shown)


moidx (approved):
LGTM

PR #23712, opened 2024-06-17
[otbn,dv] Tidy up RMA handling

rswarbrick:
This is still not beautiful, but the pass rate for the `otbn_escalate` test goes from 80% (last night's run) to `87.5%` (running 24 seeds) and the code now has comments that explain why it thinks it is doing the right thing.

There are careful comments on the individual commits. Clearly, this still isn't quite working right (and there's even a TODO message about an ugly workaround in the simulator), but this is much closer to where we want to be.

### DIFF
--- a/hw/ip/otbn/dv/otbnsim/sim/constants.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/constants.py
@@ -41,3 +41,20 @@ class ErrBits(IntEnum):
     LIFECYCLE_ESCALATION = 1 << 22
     FATAL_SOFTWARE = 1 << 23
     MASK = (1 << 24) - 1
+
+
+class LcTx(IntEnum):
+    r'''The same encoding as lc_tx_t in the RTL'''
+    ON = 0b0101
+    OFF = 0b1010
+    INVALID = 0
+
+
+def read_lc_tx_t(value: int) -> LcTx:
+    assert 0 <= value <= 15
+    if value == LcTx.ON:
+        return LcTx.ON
+    elif value == LcTx.OFF:
+        return LcTx.OFF
+    else:
+        return LcTx.INVALID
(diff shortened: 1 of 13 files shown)


hw/ip/otbn/dv/otbnsim/sim/constants.py
```diff
@@ -41,3 +41,20 @@ class ErrBits(IntEnum):
     LIFECYCLE_ESCALATION = 1 << 22
     FATAL_SOFTWARE = 1 << 23
     MASK = (1 << 24) - 1
+
+
+class LcTx(IntEnum):
+    r'''The same encoding as lc_tx_t in the RTL'''
```
hcallahan-lowrisc, line 47: Why is the raw string literal required here?

hw/ip/otbn/dv/otbnsim/sim/sim.py
```diff
@@ -253,6 +253,20 @@ def _step_exec(self, verbose: bool) -> StepRes:
             self.state.take_injected_err_bits()
             return (None, self._on_stall(verbose, fetch_next=True))
 
+        # If there is an RMA request, we treat it a bit like a fatal error, and
+        # abort any instruction that's currently running
+        if self.state.rma_req == LcTx.ON:
+            # TODO: Passing a nonzero value in ErrBits is incorrect: we don't
```
hcallahan-lowrisc, line 274: Is it worth tagging this todo with a issue number, or do you have a fix in mind for this already?

hcallahan-lowrisc (approved):
These changes seem sensible to me, but I'm definitely not an expert. Thanks for all the detailed comments!

hw/ip/otbn/dv/otbnsim/sim/constants.py
rswarbrick: Just a documentation comment :-)

hw/ip/otbn/dv/otbnsim/sim/sim.py
rswarbrick: Probably not worth creating a proper issue: I'm actively working on sorting this area out already :-) But thanks for noticing!

PR #23714, opened 2024-06-17
[prim] Waive AscentLint errors inside prim_generic_pad_wrapper

vogelpi:


a-will (approved):
LGTM

Random: It would probably be nice if we had a different pad type for clock inputs, though, particularly for DIOs that are dedicated for source synchronous I/O. Some pad attributes may make less sense for these. But that's a future release problem, hehe.

PR #23547, opened 2024-06-06
[top_earlgrey,prim_ram_*_pkg] Connect memory test port

andreaskurth:
This adds a memory test port and connects it to the AST interface added in PR #23531.

PR #23716, opened 2024-06-17
[chip_earlgrey_asic] Waive AscentLint error re ROM TEST port

vogelpi:
The ROM doesn't feature a TEST port like regular RAM primitives.

### DIFF
--- a/hw/top_earlgrey/lint/chip_earlgrey_asic.waiver
+++ b/hw/top_earlgrey/lint/chip_earlgrey_asic.waiver
@@ -98,6 +98,9 @@ waive -rules {CLOCK_USE} -location {chip_earlgrey_asic.sv}  -msg {'mio_in_raw[28
 waive -rules {HIER_NET_NOT_READ NOT_READ} -location {chip_earlgrey_asic.sv} -regexp {(Signal|Net) 'ast_pwst(\.vcc_pok|_h\.io_pok|_h\.main_pok|_h\.vcc_pok)' is not read from in module 'chip_earlgrey_asic'} \
       -comment "Not all POK signals are used inside top_earlgrey"
 
+waive -rules {HIER_NET_NOT_READ NOT_READ} -location {chip_earlgrey_asic.sv} -regexp {(Signal|Net) 'ast_rom_cfg.test' is not read from in module 'chip_earlgrey_asic'} \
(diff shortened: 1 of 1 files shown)


vogelpi:
This test PORT was only added recently with #23547.

hw/top_earlgrey/lint/chip_earlgrey_asic.waiver
```diff
@@ -98,6 +98,9 @@ waive -rules {CLOCK_USE} -location {chip_earlgrey_asic.sv}  -msg {'mio_in_raw[28
 waive -rules {HIER_NET_NOT_READ NOT_READ} -location {chip_earlgrey_asic.sv} -regexp {(Signal|Net) 'ast_pwst(\.vcc_pok|_h\.io_pok|_h\.main_pok|_h\.vcc_pok)' is not read from in module 'chip_earlgrey_asic'} \
       -comment "Not all POK signals are used inside top_earlgrey"
 
+waive -rules {HIER_NET_NOT_READ NOT_READ} -location {chip_earlgrey_asic.sv} -regexp {(Signal|Net) 'ast_rom_cfg.test' is not read from in module 'chip_earlgrey_asic'} \
+      -comment "The ROM doesn't feature a TEST port"
```
vogelpi, line 102: @meisnere , can you please confirm whether it's okay to waive this lint error or whether we should actually connect something here? I believe the Mask ROM itself doesn't involve logic and thus doesn't feature a TEST port.

a-will (approved):

hw/top_earlgrey/lint/chip_earlgrey_asic.waiver
meisnere: The ROM macro does have a TEST port and it should be connected as other RAMs.

vogelpi: Thanks for the feedback @meisnere . That's surprising! Also because it wasn't flagged during the review of #23547. I am going to work out a PR to change this but it will mean an RTL change / ECO.

vogelpi: @meisnere The PR with the RTL fix is now here:  #23718
FYI @moidx .

PR #23717, opened 2024-06-17
DRAFT - DO NOT MERGE - [usbdev] Increase robustness of packet reception

alees24:
Draft PR. As yet not confident of the proposed change and the ramifications for other signaling such as reporting of errors.

Modify end of packet detection so that it does not depend upon the validity of the SYNC signal. End of packet must always be signaled to the packet state machines even if the packet has been declared invalid, because otherwise the OUT side state machine will wait indefinitely for the end of a DATA packet.

This does not occur in normal use, but may occur in the presence of interference, signal integrity issues and/or improper sampling.

Background: fault injection testing in block level DV transmitted a valid OUT packet which was then followed by a DATA packet with a corrupted SYNC signal. The result is that 'packet_valid_q' does not become set for the DATA packet, although 'pkt_start_o' is asserted, and there is thus no 'pkt_end_o' assertion for the DATA packet. The OUT side state machine remains in the StRcvdDataStart state, from which there is no exit until the end of a subsequent packet.

The RTL change proposed in this draft PR permits the block level test to run to completion. (The test in question is generating invalid traffic whilst a streaming + checking process is running and keeping the DUT busy; the invalid traffic is expected to produce no response from the DUT but because of the above issue caused interference with the streaming operation.)

@a-will - for awareness and consideration.
@johngt


hw/ip/usbdev/rtl/usb_fs_rx.sv
```diff
@@ -277,7 +277,7 @@ module usb_fs_rx (
 
   // A bit of a misnomer: packet_start pulses when the PID begins, not SOP.
   assign packet_start = packet_valid_d & ~packet_valid_q;
-  assign packet_end   = ~packet_valid_d & packet_valid_q;
+  assign packet_end   = in_packet_q & ~in_packet_d;
```
a-will, line 280: This confuses me a bit. How do we get `in_packet_d == 0` but not `packet_valid_d == 0`. Isn't `see_eop` the trigger for both? Or is there a difference in how `packet_valid_q` and `in_packet_q` work that allowed the OUT FSM to get unstuck?

mdhayter: Is the real issue that assign pkt_start_o = see_sop; triggers the upper level machine even if there is not a valid SYNC (it just triggers on the J to K)? It won't trigger packet_valid here since there is no SYNC, thus won't trigger sending an end of packet to the upper layer using the original code.

For sure the fix of sending see_eop (via in_packet_d/q) as the end of packet seems correct (and should cover errors in the termination condition)

a-will: I feel like consumption of an explicit error indicator in the FSMs might help avoid ambiguity, though.

alees24: The FSMs already use 'valid_packet_o' from the _rx module, which is the (inverse of an) error indicator.

a-will:
For some reason, `//sw/device/tests:usbdev_stream_test_fpga_cw310_rom_with_fake_keys` failed with binding the driver (either due to a control transfer failure or test environment issue, surely, but it's hard to say which).

vogelpi:
We've discussed this issue in the tape-out sync yesterday (2024-06-24) and reached consensus to:

 1. Not merge this / integrate this as ECO now. The reasons are mostly that i) the issue isn't super critical, ii) normal operation shouldn't be affected (the fix improves things in the presence of interference and signal integrity issues), and iii) to increase confidence proper FPGA soak time would be needed which we don't have.
 2. Integrate the change post M5 after the TO branch has been taken to allow this USBDEV improvement for future releases / revisions of the design.

Integrating this improvement after taking the TO branch should be easy as it doesn't involve any software changes which could break the compatibility with Earlgrey-PROD A1.

PR #23720, opened 2024-06-18
[sival] Test inverted pads with GPIO

jwnrt:
This PR adds support for `opentitanlib` to configure the pad attributes of the device under test and extends the `gpio_pinmux` test to check inverted pads.

Only MIOs are tested. The DIOs could in theory be tested but it might be a bit harder.


engdoreis (approved):
LGTM.

PR #23723, opened 2024-06-18
[dv,usbdev] Fix up tri-state related block level tests

alees24:
This PR just addresses a couple of regressions in the block level DV tests following a change to the usb20_driver module to disable its drivers and let the bus float to the Idle state, rather than actively driving the bus to Idle.

- usbdev_phy_config_tx_osc_test_mode requires the DUT connected to take the USB out of a link reset state and permit the test mode oscillator to drive the bus.
- CSR test was failing because changing the PHY configuration was producing visible input changes to the packet reception logic,  causing it to raise errors. This modified the event counters, causing unexpected register changes in the CSR test(s) (_same_csr_outstanding).


rswarbrick (approved):
Looks good to me. I think you might be able to avoid some repetition by applying the tags to the register instead of the fields in the second commit (see e.g. `otbn.hjson`, line 454), but it definitely doesn't matter.

PR #23727, opened 2024-06-18
[tock] Enable Tock in ROM_EXT environments.

jrvanwhy:
This makes several changes:

1. Update Tock to the merge commit for PR 4034 (which makes its ePMP setup compatible with the ROM_EXT).
2. tock_image uses opentitantool to assemble the final Tock image, because tockloader was hardcoded to assume Tock is at address 0x20000000.
3. tock_image now produces a signed image.
4. The kernel linker script sets separate symbols for "start of flash" and "start of the Tock image", as required by the new ePMP setup logic.
5. The kernel linker script now fills in all the required manifest fields.
6. Removes TicKV from the Tock kernel as well as the flash controller reconfiguration, as they are not working on silicon. TicKV is unlikely to suit our needs.
7. Changes the Tock chip config to one that matches silicon.

### DIFF
--- a/rules/tock.bzl
+++ b/rules/tock.bzl
@@ -5,7 +5,11 @@
 """Rules for assembling Tock binaries.
 """
 
+load("//rules/opentitan:toolchain.bzl", "LOCALTOOLS_TOOLCHAIN")
+load("//rules:signing.bzl", "sign_binary")
 load("@rules_cc//cc:find_cc_toolchain.bzl", "find_cc_toolchain")
+load("@lowrisc_opentitan//rules/opentitan:exec_env.bzl", "ExecEnvInfo")
+load("@lowrisc_opentitan//rules/opentitan:util.bzl", "get_fallback")
 load(
     "//rules:rv.bzl",
     "rv_rule",
@@ -19,6 +23,7 @@ TockApplication = provider(
         "tab": "TAB file for this application",
         "tbf": "TBF file for this application",
         "elf": "ELF file for this application",
+        "flash_start": "Address at which the app image must start in flash",
     },
 )
 
@@ -58,7 +63,7 @@ def _tock_elf2tab_impl(ctx):
 
     outputs.append(elffile)
     return [
-        TockApplication(tab = tabfile, tbf = tbffile, elf = elffile),
+        TockApplication(tab = tabfile, tbf = tbffile, elf = elffile, flash_start = ctx.attr.flash_start),
         DefaultInfo(
             files = depset(outputs),
             data_runfiles = ctx.runfiles(files = outputs),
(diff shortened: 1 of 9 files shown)


rules/tock.bzl
```diff
@@ -115,56 +121,35 @@ def _tock_image_impl(ctx):
...
-        ],
-        executable = ctx.executable._tockloader,
+            "--rcfile=",
+            "image",
+            "assemble",
+            "--mirror=false",
+            "--output={}".format(image.path),
+            "--size=0x70000",
+            "{}@0".format(kernel_binary.path),
+        ] + ["{}@{}".format(app[TockApplication].tbf.path, app[TockApplication].flash_start - 0x20010000) for app in ctx.attr.apps],
```
cfrantz, line 134: I think this constant should also be expressed in a parameter (although, in this case, its probably ok for the parameter to have this value as a default - it's unlikely to change unless you're working with a chip that has a different flash arrangement or you're intentionally creating a bad image for a test case).

sw/device/silicon_owner/tock/tests/uart/BUILD
```diff
@@ -30,19 +30,17 @@ tock_elf2tab(
 
 tock_image(
     name = "image",
-    app_flash_start = 0x20040000,
     apps = [":tab"],
+    exec_env = "//hw/top_earlgrey:silicon_owner_sival_rom_ext",
     kernel = "//sw/device/silicon_owner/tock/kernel",
+    manifest = "//sw/device/silicon_owner:manifest",
```
cfrantz, line 36: FYI, the `exec_env` usually supplies the manifest (ie: the binary and test rules typically leave it blank and fallback to the `exec_env`'s manifest unless they're trying to do something special for a test).

cfrantz (approved):
LGTM with a few minor comments.

rules/tock.bzl
```diff
@@ -77,6 +82,7 @@ tock_elf2tab = rule(
         "src": attr.label(mandatory = True, allow_single_file = True, doc = "ELF binary to convert"),
         "disable": attr.bool(default = False, doc = "Mark the application as disabled"),
         "arch": attr.string(mandatory = True, doc = "Target architecture for the ELF binary (e.g., `rv32imc`)"),
+        "flash_start": attr.int(default = 0x20040000, doc = "Application starting address in flash"),
```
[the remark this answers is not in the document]
jrvanwhy: Good point, I made it mandatory.

In the future, the link address and app's linker script should probably be tied together somehow (put into the same provider?) so that it's harder to mess up.

rules/tock.bzl
jrvanwhy: I made the address range of the Tock image into new `image_start` and `image_end` parameters, to avoid the magic numbers this section currently has.

sw/device/silicon_owner/tock/tests/uart/BUILD
jrvanwhy: `manifest` is now optional and I've removed its specifications.

PR #23730, opened 2024-06-19
[opentitantool] SPHINCSPLUS refactoring

cfrantz:
Refactor the SPHINCSPLUS library to support multiple algorithm variants.

1. Patch the sphincsplus C reference implementation to give unique symbol names to each of the possible algorithm variants.  The unique names are mangled onto the function names through classic C preprocessor abuse.  We supply the "namespace" to our `cc_library` and `rust_bindgen` rules through the `defines` or `clang_flags` attributes of those rules.
2. Refactor the rust binding:
2a. Create an enum whose discriminants are the sphincsplus algorithm names.  Create a `macro_rules!` macro that can expand an implementation for each discriminant that invokes the corresponding per-algorithm functions supplied by (1).
2b. Create `Spx{Public,Secret}Key` structs to hold key material and their associated algorithm names.  Create nice associated functions on these key types that provide nice interfaces to the lower-level functions supplied by (2a).
2c. Add `{Encode,Decode}Key` traits that can serialize the key material to/from PEM files.  Save the algorithm name into the PEM file so that when keys are loaded from files, they key material will get associated with the correct implementation functions.
3. Plumb the new library into opentitanlib.  Eliminate most of the legacy `spx` module in favor of using the new `sphincsplus` crate directly.


cfrantz:
TODO:
- Test with the ROM
- Test with the HSM

jadephilipoom (approved):
This approach LGTM!

cfrantz:
I've added domain handling to the `sphincsplus` crate and have plumbed that through to opentitantool in the form of a `--domain` command-line option on the relevant sign/verify commands.

I've also rewritten the headers on the existing PEM files to identify the keys as `SPHINCS+_SHA2_128s_simple` keys.

Passing CI is an effective test with the ROM.

I've also tested with the HSM, although in order to do that, I need to change the version of the reference implementation we import to the ref-impl's `master` branch.  The HSM's current PQ implementation is still using the algorithms prior to the FORS indexing change.

timothytrippel (approved):

sw/device/silicon_creator/rom/keys/fake/spx/dev_key_0_spx.pem
```diff
@@ -1,4 +1,4 @@
------BEGIN RAW SPHINCS+ PRIVATE KEY-----
+-----BEGIN RAW:SPHINCS+_SHA2_128s_simple+ PRIVATE KEY-----
```
jadephilipoom, line 1: nit: It looks like there's an extra `+` in all of these. Is that necessary in some way for the `SphincsPlus::from_str` operation to work, or is it a typo?
```suggestion
-----BEGIN RAW:SPHINCS+_SHA2_128s_simple PRIVATE KEY-----
```

jadephilipoom (approved):

sw/device/silicon_creator/rom/keys/fake/spx/dev_key_0_spx.pem
cfrantz: Good catch.  I must have botched the `sed` expression to update the files.  Fixed.

cfrantz:

PR #23732, opened 2024-06-19
[rom] Switch SPHINCS+ to SHA2 parameters.

jadephilipoom:

This PR:
- Switches the SPHINCS+ device and host-side implementations to use SHA2 parameters
- Updates test data in a few places as needed (including keys)
  - This is most of the diff by LOC, the PR isn't as huge as it looks

After this change, #23144 is done except for introducing the pure/pre-hash OTP switch.

## Code size

This PR reduces the ROM size from  31,396 bytes on `master` (d47e7cf8cd8afaf9362d3022f9e475588e6ff15f) to 30,400 bytes. Most of that is because the KMAC driver (856 bytes) is no longer needed. I was also able to remove `HARDENED_RETURN_IF_ERROR` in some places because the HMAC driver doesn't produce errors.

## Performance

The final SPHINCS+ verification procedure is a bit slower than with SHAKE (14.7ms instead of 12.5ms) because of differences in address and public key seed encoding between the SHAKE and SHA2 variants. I have a local branch that uses the new HMAC save/restore feature to bring the performance back up so it's slightly faster than SHAKE (12.1ms) at the cost of 200B code size, but it may or may not be worth merging it.


cfrantz (approved):

moidx (approved):
🎉

moidx:
More than 2ms of saved boot time for 200 Bytes of code seem like a good deal to me, @cfrantz what are your thoughts on this?

PR #23733, opened 2024-06-19
[otbn,dv] Add unconditional return to otbn_model::step()

rswarbrick:
No functional change, but this avoids a lint warning from the C++ compiler (and I was looking at the log because I've got a genuine local bug as well)


hcallahan-lowrisc (approved):

rswarbrick:
CI failure is caused by Bazel failing to download OpenOCD (definitely not caused by the changes in this PR). Merging.

PR #23734, opened 2024-06-19
[dv,usbdev] Data toggle clear sequence

alees24:
Add a test sequence for exercising the data toggle clear functionality which was subsumed into the data toggle save/restore behavior of the new registers.

Data Toggle clearing is still the most common use case for these registers, since clearing the data toggle bit is essential for handling Bus Resets and some Control Transfers to endpoint zero, and is important in some protocols.
The more general save/restore functionality is for implementing resume from Deep Sleep.


rswarbrick (approved):

alees24:
CI failures unrelated.

Issue #18340, opened 2023-05-04
[aon_timer, dv] Pause in sleep feature for watchdog is untested


GregAC:
The aon_timer watchdog has a feature that allows the watchdog to be paused when in sleep. This isn't currently being tested in the testbench.

> estimate 4

GregAC:
P1 so as to discuss if we want to test this for M2.5.1. My view is it's a simple feature that looks correct by inspection of the RTL and isn't critical. Testing would be straight-forward but could end up taking a day or more of effort to implement and deal with any regression issues.

I could have a quick go at doing it and if there's lots of regression failures to wade through that don't appear to be RTL errors leave it til after M2.5.1

@msfschaffner @jdonjdon thoughts?

jdonjdon:
That looks reasonable to me.

Rdgs,

Jaedon

msfschaffner:
Sounds reasonable to me, but I wouldn't invest too much time.
IMO it's fine to punt this to M2.5.1, given that we already have several top-level tests.

GregAC:
Ok I think we'll move this to M2.5.2, no further action will be taken here for M2.5.1

moidx:
DV improvement, moving to M5

johngt:
@GregAC  / @andreaskurth / @moidx 
I think one of the areas integrations team are most concerned about is on sleep and resume from sleep.
@GregAC do we have coverage for this already or should we high grade this as a P1 to give us confidence that this is covered?

GregAC:
This remains a coverage gap and should be addressed for M5, @antmarzam is this something you could take on?

GregAC:
Moved to P1

antmarzam:
Sure - I'll take this on!

antmarzam:
Closing, this is now taken care of in PR #23743

PR #23743, opened 2024-06-19
[dv, aon_timer] Stimulus for wdog_ctrl.pause_in_sleep

antmarzam:
* aon_timer TB now randomizes wdog_ctrl.pause_in_sleep in aon_timer_smoke_vseq
* aon_timer SCB now stops WDOG counting in sleep mode when pause_in_sleep is set

I have run a local regression and the passing rate seems the same as the nightly regression. And the failures (2/102) have the same failing signature as the nightly one [(report from 11/06/2024)](https://reports.opentitan.org/hw/ip/aon_timer/dv/2024.06.12_04.38.22/report.html):
```
UVM_ERROR @ 23395289269 ps: (cip_base_vseq.sv:473) [uvm_test_top.env.virtual_sequencer.aon_timer_common_vseq] Check failed exp_val == act_val (0 [0x0] vs 1 [0x1]) when reading the intr CSRaon_timer_reg_block.intr_state
```

I merged the coverage from the local regression (with tweaks to Xcelium coverage config file, [since toggle coverage is only enabled for ports](https://github.com/lowRISC/opentitan/blob/5b17d34ccc8248e791c9d20f054ad52f9d778012/hw/dv/tools/xcelium/common.ccf#L36-L37))

I see the signal `pause_in_sleep` toggling:
Screenshot from 2024-06-19 16-54-12

### DIFF
--- a/hw/ip/aon_timer/dv/env/aon_timer_scoreboard.sv
+++ b/hw/ip/aon_timer/dv/env/aon_timer_scoreboard.sv
@@ -250,6 +250,10 @@ class aon_timer_scoreboard extends cip_base_scoreboard #(
     end // compute_num_clks
   endtask
 
+  virtual task wait_for_sleep();
+    wait ( !(wdog_pause_in_sleep & cfg.sleep_vif.sample_pin()));
+  endtask
+
   virtual task run_wkup_timer();
     event sample_coverage;
     forever begin
@@ -339,6 +343,7 @@ class aon_timer_scoreboard extends cip_base_scoreboard #(
           // logic.
           cfg.aon_clk_rst_vif.wait_clks(4);
           while (count < wdog_bark_num) begin
+            wait_for_sleep();
             cfg.aon_clk_rst_vif.wait_clks(1);
             // reset the cycle counter when we update the cycle count needed
             count = wdog_num_update_due ? 0 : (count + 1);
@@ -374,7 +379,7 @@ class aon_timer_scoreboard extends cip_base_scoreboard #(
       join_any
       disable fork;
     end
-  endtask
+  endtask : run_wdog_bark_timer
 
   virtual task run_wdog_bite_timer();
     event sample_coverage;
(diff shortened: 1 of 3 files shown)


antmarzam:
This takes care of #18340

hw/ip/aon_timer/dv/env/aon_timer_scoreboard.sv
```diff
@@ -339,6 +339,7 @@ class aon_timer_scoreboard extends cip_base_scoreboard #(
           // logic.
           cfg.aon_clk_rst_vif.wait_clks(4);
           while (count < wdog_bark_num) begin
+            wait ( !(wdog_pause_in_sleep & cfg.sleep_vif.sample_pin()));
             cfg.aon_clk_rst_vif.wait_clks(1);
```
rswarbrick, line 343: This looks really sensible. Do we have an opinion about what happens if `wdog_pause_in_sleep` or `sleep_mode_i` change value somewhere in the middle of an AON timer cycle?

Maybe it makes sense to pull these two lines out into a task? If we don't have an opinion (because it shouldn't happen), we could use a fork to make sure that the values stay constant.

(The same comment applies to lines 406,407 below)

antmarzam: I don't have a strong opinion about this. `sleep_mode_i` is latched on the aon_timer clock, whereas `wdog_pause_in_sleep` is latched on the TL-UL clock. So for `sleep_mode_i` it doesn't matter if it changes in the middle of a clock cycle.

CDC delays are strict in the scoreboard at the moment, and TL-UL writes are directly setting the state variables.

I don't follow - What do you mean by to pull out these two lines out into a task?

rswarbrick: My second paragraph didn't really make much sense, sorry. (And I don't completely understand it myself!)

What I think I was suggesting is that we could have something like
```systemverilog
task wait_for_sleep();
  wait (!(wdog_pause_in_sleep & cfg.sleep_vif.sample_pin()));
  cfg.aon_clk_rst_vif.wait_clks(1);
endtask
```
which makes the call site a bit more obvious to understand.

rswarbrick (approved):
Thanks for the tweaks. Looks good to me!

PR #23744, opened 2024-06-20
[rv_dm,dv] Tidy up and properly implement the jtag_dmi_cmderr_busy testpoint

rswarbrick:
The first commit tidies up the testpoint. The second commit writes an implementation that might actually do something.


rswarbrick:
Force-push tidies up the end of the vseq so that we correctly clear the busy status, and check that we've done so.

rswarbrick:
Force-push makes this test a little more fragile(!) so that it fails in a more understandable way if it follows another vseq that didn't tidy up after itself properly.

rswarbrick:
Well, this is a bit embarrassing! The latest force-push fixes yet another problem that shows up with the stress sequence (caused by the fact that I wasn't cleaning up the `cmderr` field at the end).

hcallahan-lowrisc (approved):
LGTM

PR #23746, opened 2024-06-20
[rv_dm,dv] Tidy up implementation of the ndmreset_req vseq

rswarbrick:
The testpoint description can be improved slightly (first two commits). Then the implementation needs re-doing, so that it gives us some confidence! That's the third commit.


vogelpi (approved):
LGTM, thanks for documenting the changes and motivation so well @rswarbrick !

PR #23747, opened 2024-06-20
[jtag,dv] Keep TCK running a short time after a transaction

rswarbrick:
This handles a rather confusing situation in the DMI frontdoor. The problem is that a DMI request has to travel from dmi_jtag to dm_top (involving a CDC from TCK to the main clock) and then back again (involving a CDC from the main clock back to TCK).

If we stop sending JTAG operations in that time, the CDC doesn't actually get anywhere. The wait is jtag_agent_cfg_h.vif.wait_tck(10), which ends up waiting based on elapsed time rather than clock edges.

The result is that when we send another JTAG operation to find out the result of the operation, the "busy" status only makes it through the CDC *after* the JTAG operation has finished and reported that the dm is idle.

Things then get rather confused when the "busy" status appears, which the agent interprets as a mysterious operation that has just come into being.

### DIFF
--- a/hw/dv/sv/jtag_agent/jtag_agent_cfg.sv
+++ b/hw/dv/sv/jtag_agent/jtag_agent_cfg.sv
@@ -28,6 +28,16 @@ class jtag_agent_cfg extends dv_base_agent_cfg;
   // This knob can bypass default 1 cycle initial delay of 'driver.drive_jtag_req()' task.
   // Use this knob only for the necessary tests.
   bit     min_rti = 0;
+
+  // This JTAG driver expects to control the JTAG interface clock (TCK), which it does by setting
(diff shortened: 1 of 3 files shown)


rswarbrick:
The fact that this window exists at all seems like a design error (or maybe a misunderstanding on the part of me / the DV team!)

I've raised a question about it on the Pulp project here: https://github.com/pulp-platform/riscv-dbg/issues/166.

hcallahan-lowrisc (approved):
This makes sense to me, but I don't know the problem space too well. Is it expected under normal use for TCK to continue running in the intermediate period to ensure data has time to propagate through the CDC?

rswarbrick:
Yep, that's exactly right. The propagation time is essentially measured in cycles of the two clocks. If one of the clocks stops, it takes a while :-D

Issue #21815, opened 2024-03-04
[hmac] Verify FIFO empty status interrupt


vogelpi:
### Description

When changing the FIFO empty interrupt type from "Event" to "Status" #21809, it turned out that the scoreboard previously tried to model this interrupt cycle accurate. Doing this for the status type interrupt is very challenging (and maybe not needed?). Thus, I've disabled verification of the FIFO empty interrupt (it's currently not being used by software anyway).

For M4/M5, we should verify this interrupt properly.

vogelpi:
Software is currently not using the fifo_empty interrupt and we may thus want to consider de-prioritizing this task @andreaskurth .

andreaskurth:
By @vogelpi: Also, HW is a lot faster at consuming inputs than SW can provide data. So most of the time the FIFO will be empty.
By @moidx: Some Si creator drivers don't even use this for performance reasons.

Agreed in triage to move to M5

vogelpi:
I am closing this as completed as both the DV and the RTL fix have been merged. There is nothing left to do here.

PR #23739, opened 2024-06-19
[hmac,rtl] FIFO Empty intr after FIFO seen full

martin-velay:
- fix RTL issue #23738
  - prevent to raise a FIFO empty interrupt when not required in case the FIFO is seen full while a stop is triggered.

PR #23749, opened 2024-06-20
[hmac,dv] FIFO empty interrupt and test mode pins

martin-velay:
- Verify FIFO empty interrupt register to tackle issue #21815
  - Add checks on interrupt pins when test mode is triggered via the dedicated register INTR_TEST

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_scoreboard.sv
+++ b/hw/ip/hmac/dv/env/hmac_scoreboard.sv
@@ -8,7 +8,7 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
   `uvm_component_utils(hmac_scoreboard)
   `uvm_component_new
 
-  bit             sha_en, fifo_empty, hmac_idle;
+  bit             sha_en, hmac_idle;
   bit [7:0]       msg_q[$];
   bit [7:0]       msg_part_q[$];  // Queue containing piece of the message if HASH is interrupted
   bit             hmac_start, hmac_process, hmac_stopped, hmac_continue;
@@ -20,6 +20,18 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
   bit [TL_DW-1:0] exp_digest[NUM_DIGESTS];
   bit [3:0]       previous_digest_size, expected_digest_size;
   bit             invalid_cfg;
+  bit [TL_DW-1:0] hmac_status_data;
+  bit [5:0]       hmac_fifo_depth;
+  bit             hmac_fifo_full;
+  bit             fifo_full_detected;
+  bit             hmac_fifo_empty;
+  bit             fifo_empty_intr;
+  bit             hmac_process_last;
+  bit             hmac_stopped_last;
+  bit             hmac_start_last;
+  bit             hmac_continue_last;
+  bit [TL_DW-1:0] intr_test;
+  bit [TL_DW-1:0] intr_state_exp;
 
   function void build_phase(uvm_phase phase);
     super.build_phase(phase);
@@ -31,6 +43,7 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
       hmac_process_fifo_status();
       hmac_process_fifo_wr();
       hmac_process_fifo_rd();
+      hmac_intr_test();
       monitor_cov();
     join_none
   endtask
@@ -191,14 +204,17 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
             end
           end
           "intr_test": begin // testmode, intr_state is W1C, cannot use UVM_PREDICT_WRITE
-            bit [TL_DW-1:0] intr_state_exp = item.a_data | `gmv(ral.intr_state);
+            intr_test = item.a_data;
+            // As INTR_STATE.hmac_err and INTR_STATE.hmac_done are RW1C, we should get back the
+            // current state
+            intr_state_exp = intr_test | `gmv(ral.intr_state);
             void'(ral.intr_state.predict(.value(intr_state_exp), .kind(UVM_PREDICT_DIRECT)));
             if (cfg.en_cov) begin
               bit [TL_DW-1:0] intr_en = `gmv(ral.intr_enable);
               hmac_intr_e     intr;
               intr = intr.first;
               do begin
-                cov.intr_test_cg.sample(intr, item.a_data[intr], intr_en[intr],
+                cov.intr_test_cg.sample(intr, intr_test[intr], intr_en[intr],
                                         intr_state_exp[intr]);
                 intr = intr.next;
               end while (intr != intr.first);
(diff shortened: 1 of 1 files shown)


martin-velay:
I still need to run a full regression to be sure nothing is broken, but you could start to review it

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -193,14 +206,16 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
             end
           end
           "intr_test": begin // testmode, intr_state is W1C, cannot use UVM_PREDICT_WRITE
-            bit [TL_DW-1:0] intr_state_exp = item.a_data | `gmv(ral.intr_state);
+            intr_test = item.a_data;
+            // As hmac_err and hmac_done are W1C, we should get back the current state
```
rswarbrick, line 210: I think the comment might be talking about the wrong registers here?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -537,21 +554,96 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+
+      // Delay FIFO empty signal to be aligned with the DUT behavior
+      fork
+        begin
+          cfg.clk_rst_vif.wait_clks(1);
+          fifo_empty_intr = pre_fifo_empty_intr;
         end
+      join_none
+
+      // Check wether FIFO full has been detected for the ongoing message but the retrictions cases
```
rswarbrick, line 605: Nit: Spelling of whether

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -537,21 +554,96 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+
+      // Compute FIFO level flags
+      hmac_fifo_depth = hmac_wr_cnt - hmac_rd_cnt;
+      hmac_fifo_full  = hmac_fifo_depth == HMAC_MSG_FIFO_DEPTH_WR;
+      hmac_fifo_empty = hmac_fifo_depth == 0;
+
+      // The FIFO empty interrupt is raised only if the message FIFO is actually writable by
+      // software, i.e., if all of the following conditions are met:
+      //   1- The HMAC block is not running in HMAC mode and performing the second round of
+      //      computing the final hash of the outer key as well as the result of the first round
```
rswarbrick, line 584: I'm not an HMAC expert but this looks surprising to me. Do you want to drop the "as well" here?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -537,21 +554,96 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+        pre_fifo_empty_intr = 0;
+      end
+
+      // Delay FIFO empty signal to be aligned with the DUT behavior
+      fork
+        begin
+          cfg.clk_rst_vif.wait_clks(1);
+          fifo_empty_intr = pre_fifo_empty_intr;
         end
+      join_none
```
rswarbrick, line 603: I can imagine this getting rather confusing if something surprising happens in the middle, like a reset stopping the clock for a bit. Would it make more sense to do this explicitly and copy `pre_fifo_empty_intr` to `fifo_empty_intr` at the start of this task? (since this task is the only one that currently uses it)

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -537,21 +554,96 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
+  endtask : hmac_process_fifo_status
+
+  // Check interrupt pins in test mode
+  virtual task hmac_intr_test();
+    bit [TL_DW-1:0] intr_en;
+    hmac_intr_e     intr;
+    forever @(intr_test) begin
+      intr_en = `gmv(ral.intr_enable);
+      foreach (intr_test[intr_i]) begin
+        fork begin
```
rswarbrick, line 623: The fork structure here is a bit hard to follow. I think it might be easier if you split this task into "one per bit", so `intr_i` ends up as an argument to the task. Then the chopped up task can wait with `@(intr_test[intr_i])` and all the fork stuff (and things like having to explicitly chop out `auto_intr_i`) ends up outside of that task.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Here I am referring to the INTR_STATE register as it is predicted while we are accessing the INTR_TEST register.
I have changed a bit my comment to make it clearer.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Sorry, that one doesn't want to stick in my mind!

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: This comment is almost a copy/paste of the description here:
https://opentitan.org/book/hw/ip/hmac/doc/registers.html#intr_state--fifo_empty

This description is from @vogelpi. What do you think?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: No, I can't do this copy at the beginning of this task it require the updated `pre_fifo_empty_int` value at this exact moment.
But I can do like the other tasks and implement a fork like:

```
            begin : reset_hmac
              wait(cfg.under_reset);
            end
          join_any
          disable fork;
```

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: OK, it's done. I hope it'll be easier to read and also that I haven't broken anything.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: I have pushed a commit which doesn't work yet! https://github.com/lowRISC/opentitan/pull/23749/commits/dbf5ec3c203d3f506c65af01c9fd0fe21fd8778e

I'll work on it later next week.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
vogelpi: Sorry, my description is probably confusing. What I meant here is:
If HMAC is running in HMAC mode (vs. the unkeyed SHA2 mode) and currently doing the second round, software cannot write the FIFO. During this second round, HMAC hashes the outer key together with the result of the first round. The first round result was produced using the inner key.

martin-velay:
@vogelpi I have added a temporary workaround to be able to merge without taking the related RTL fix from this PR https://github.com/lowRISC/opentitan/pull/23739.

Now, this PR could be merged into the master and won't break anything.

martin-velay:
@vogelpi, as it has been decided to merge the RTL fix from this PR #23739 , I have reverted the proposed workaround mentioned above.
If needed, it could be reestablished, see this commit: https://github.com/lowRISC/opentitan/commit/55e20274ed386a1ad7ff6aa453ea9440ff39f49b

rswarbrick:
Hi @martin-velay! Would you mind rebasing this over recently merged changes?

vogelpi (approved):
LGTM, when rebasing, would mind correcting the commit messages? One commit contains two sign-off lines.

rswarbrick:
Sorry, I jumped the gun! It turns out that the rebasing wasn't too hard. Happy for either of us to merge, but maybe it makes sense to wait long enough to check that I didn't accidentally break the compile.

vogelpi:
Thanks @rswarbrick . The DV checks passed already. What's left are the FPGA tests which are not affected by this. I am thus merging the PR now.

PR #23750, opened 2024-06-20
[rv_dm,dv] Tidy up "debug enable check"

rswarbrick:
The first commit tidies up some rather messy code (the original authors didn't use inheritance, which makes things rather easier). The second commit avoids an infinite wait if something goes wrong.


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
```diff
@@ -8,17 +8,8 @@
...
-
   `uvm_object_new
 
-  rand lc_ctrl_pkg::lc_tx_t pinmux_hw_debug_en;
-  constraint lc_hw_debug_en_c {
-    lc_hw_debug_en == lc_ctrl_pkg::On;
-  }
-  constraint pinmux_hw_debug_en_c {
-    pinmux_hw_debug_en == lc_ctrl_pkg::On;
-  }
```
marnovandermaas, line 20: I assume this is unecessary because it is already implied when `lc_hw_debug_en` is turned on.

marnovandermaas (approved):
Thanks for the code cleanup.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
rswarbrick: Nope, that was a mistake on my part. Thanks for spotting it! I'll drop that bit of the hunk.

rswarbrick: Huh. It's even stupider than that: the `pinmux_hw_debug_en` class variable has never actually done anything. I'm going to keep that class variable and the constraint and will follow up in a different PR with changes that make it actually do something. Sigh...

PR #23752, opened 2024-06-20
[rv_dm,dv] Clean up ndmreset value at end of smoke vseq

rswarbrick:
Failing to do this caused confusion in the stress_all test if the following vseq assumed we weren't in the middle of the ndmreset sequence.


eshapira (approved):
LGTM, thanks.

Issue #23724, opened 2024-06-18
[ROM] Improve boot failure error reporting


cfrantz:

cfrantz:
Improve the error reporting when the ROM fails to boot one of the slots.  Often, one of the slots has an image and the other is empty.  If there is an error with the image, the ROM will try the empty slot and then report the "Bad Identifier" error from the empty slot rather than the error associated with the bad image.

This should be done in a similar manner to #20965 in the ROM_EXT.

cfrantz:
I no longer believe this issue is relevant to the ROM.

In the ROM, we call boot_policy_manifests_get, which returns a sorted list of slots based on version numbers (secver, major, minor).

If a slot is erased, those values are all 0xFFFFFFF (greater), meaning that slot gets tried first, identified as bad, and then the other slot gets tried.  The second slot is the one that has the image, which also might be bad.  But, since it's tried second, it's BFV is what is emitted by the ROM, which is what we want.

In the ROM_EXT, Slot A and B are sorted based on `boot_data->primary_bl0_slot`, and so an empty slot might get tried second if the first is bad, yielding a wrong error code.

PR #23753, opened 2024-06-20
[rom] Improve boot failure error reporting

cfrantz:
Improve the error reporting when the ROM fails to boot one of the slots. Often, one of the slots has an image and the other is empty. If there is an error with the image, the ROM will try the empty slot and then report the "Bad Identifier" error from the empty slot rather than the error associated with the bad image.

Addresses #23724


moidx (approved):

moidx:
I am surprised we didn't have to update any test cases here :)

cfrantz:
I no longer believe this change is needed:
https://github.com/lowRISC/opentitan/issues/23724#issuecomment-2183547274

PR #23754, opened 2024-06-20
[rv_dm,dv] Properly implement the rv_dm_cmderr_not_supported_vseq test

rswarbrick:
The first commit tidies up the testpoint slightly. The second implements it, replacing a rather nonsensical mess that was there before.


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #23756, opened 2024-06-20
[tock] Run Tock tests on both the cw310 and silicon.

jrvanwhy:
This builds two Tock kernels: one for the cw310 and one for silicon. Each test app is built once, then combined with both kernels to produce two test images. The `opentitan_test` rule is configured to run image corresponding to its exec environment.

I manually tested this works on silicon.


cfrantz (approved):

jwnrt (approved):
I've been using the silicon half of this fine, and the FPGA side works on our CW310.

PR #23761, opened 2024-06-21
[rom] Speed up SPHINCS+ computations.

jadephilipoom:
Accelerates SPHINCS+ computations by changing the HMAC driver:
1. Separate out `init()` into `configure()` and `start()` so that the code can start a new hash operation of the same type without reconfiguring the hardware.
2. Separate `process()` from `final()` so that we can perform unrelated computations in between letting HMAC know the input is done and actually reading the digest, while HMAC is processing.
3. Use the new save/restore feature of the HMAC block to cache the state of the most performance-relevant computation after ingesting the first message block (the reference implementation also does this; I just didn't include it in #23732 because I wasn't sure we would want to commit to the code size tradeoff).

In terms of performance (average of 10 valid runs from `kat0`) and code size:
| Code  | Performance (cycles) | Code size (bytes) |
| ------------- | ------------- |----------------|
| `master` (a502432bf42811abea7132202bc001c1a9831199)  |  `1482599` | `30400`  |
| change (1)  |  `1428422` | `30420` |
| changes (1) and (2)  |  `1417290` | `30420` |
| changes (1), (2), and (3)  |  `1216347` | `30592` |
 
In summary, the changes together make the average verification speed **2.66ms faster** at the cost of **192 bytes** of code size. Most of both the speed benefit and the code size comes from change (3). With this change, the SHA2 variant becomes slightly faster than the SHAKE variant was before (~12.5ms).

### DIFF
--- a/sw/device/silicon_creator/lib/drivers/hmac.c
+++ b/sw/device/silicon_creator/lib/drivers/hmac.c
@@ -13,7 +13,7 @@
 #include "hmac_regs.h"  // Generated.
 #include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
 
-void hmac_sha256_init_endian(bool big_endian_digest) {
+void hmac_sha256_configure(bool big_endian_digest) {
   // Clear the config, stopping the SHA engine.
   abs_mmio_write32(TOP_EARLGREY_HMAC_BASE_ADDR + HMAC_CFG_REG_OFFSET, 0u);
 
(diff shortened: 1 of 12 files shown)


cfrantz (approved):

moidx (approved):

jadephilipoom:
Just to correct the archival record, I actually made a mistake in the measurements here -- SHA2 is actually significantly faster than it says here. I did the measurements on FPGA instead of Verilator, which overcounts the cycles significantly. The number I got from a recent run on Verilator was `935122` cycles, or 9ms.

PR #23762, opened 2024-06-21
[rom] Introduce domain separation prefix for SPHINCS+.

jadephilipoom:
For later compatibility with the FIPS 205 standard, introduces a domain separation prefix for the message that is always two zero bytes, indicating a "pure" (not pre-hashed) message and an empty context. For now, the pre-hashed variant is not supported (we may add it as a separate change, but it made sense to me to do this step first).

I chose to implement the domain separator outside of the main SPHINCS+ implementation to maintain KAT compatibility; the existing KATs don't include the domain separator, and NIST [has said](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/Mf2kemwwreY/m/oF8HfaNxAQAJ) that their KATs will include targets for the "internal" underlying operation that does not include the prefix.

Resolves https://github.com/lowRISC/opentitan/issues/21944


cfrantz (approved):

sw/device/silicon_creator/lib/sigverify/spx_verify.c
```diff
@@ -62,6 +62,19 @@ static const uint32_t kSpxVerifyShares[kSigverifySpxRootNumWords] = {
...
 
+/**
+ * Domain-separation prefix for SPHINCS+ with no prehashing.
+ *
+ * The domain separation prefix is the following byte sequence:
+ *   0x00 || len(ctx) || ctx
+ *
+ * In our case, `ctx` is always the empty string, so the length is 0.
+ */
+static const uint8_t kSpxVerifyPureDomainSep[] = {
```
moidx, line 73: This sgtm. The sequence of three arguments gets reduced to two due to `len(ctx) == 2`.  This matches the proposed specification.

moidx (approved):

PR #23765, opened 2024-06-21
[rom] Add "pure" and "pre-hash" domain separators for SPHINCS+.

jadephilipoom:
Includes commits from https://github.com/lowRISC/opentitan/pull/23762; draft until that PR is merged.

Adds an OTP parameter to toggle between "pure" and "pre-hashed" SPHINCS+ mode, and code in the ROM to toggle between the two.

I did not add tests for the pre-hashed mode because, to my knowledge, test data is not available. @moidx @cfrantz are you aware of any existing pre-generated tests or have any thoughts on how to test this other than to just make some more test data a la `spx_verify_functest` to check that the opentitantool and device code agree?


hw/ip/otp_ctrl/data/BUILD
```diff
@@ -94,6 +94,9 @@ otp_json(
                 # `kSigverifySpxDisabledOtp` and `kSigverifySpxEnabledOtp` in
                 # sw/device/silicon_creator/lib/sigverify/spx_verify.h.
                 "CREATOR_SW_CFG_SIGVERIFY_SPX_EN": otp_hex(0x8d6c8c17),
+                # Use pre-hashed SPX+ (if SPX+ enabled). See the definition of
+                # `hardened_bool_t` in sw/device/lib/base/hardened.h.
+                "CREATOR_SW_CFG_SIGVERIFY_SPX_PREHASH": otp_hex(CONST.HARDENED_FALSE),
```
cfrantz, line 99: I think we should consult with @moidx on this:  should the SPX+ variant be determined by OTP, or should we put something in the manifest to identify the signature variant?

cfrantz:
LGTM, pending the resolution of how we decide between Pure and Pre-Hashed.

hw/ip/otp_ctrl/data/BUILD
moidx: The specification recommends not using the same key in both pre-hashed and pure forms, so using this new OTP setting is ok. 

I prefer to use the existing `ROT_CREATOR_AUTH_CODESIGN_SPX_KEY_CONFIG0` so that we can make this decision at key provisioning time. 

https://github.com/lowRISC/opentitan/blob/6e698b4dfec2f9e61e091cbb9a866197681745f9/hw/ip/otp_ctrl/data/otp_ctrl_mmap.hjson#L430-L433

The key types can be augmented here: 

https://github.com/lowRISC/opentitan/blob/6e698b4dfec2f9e61e091cbb9a866197681745f9/sw/device/silicon_creator/lib/sigverify/spx_key.h#L56-L76

We can consider removing the config fields if we are not planning to use them, but take a look first. It is possible that the config field is part of the key struct used in spx sigverify.

cfrantz: As I understand it, the guidance is that one should not use the same key for creating pure vs. pre-hashed signatures.  If we follow that guidance, then like the algorithm variant, then pure vs. pre-hashed tells you how you can use the key.

With that in mind, recall the OTP structure that encodes the key:
```
SPX_KEY_TYPEn: a word associating a key with the TEST, DEV or PROD lifecycle states.
SPX_KEYn: 32 bytes of key material
SPX_KEY_CONFIGn: a word identifying the key algorithm (SHAKE-128s, SHA2-128s, etc).
```

I think we can dual-purpose the `CONFIGn` word to encode both the algorithm and key-usage constraint.  That is to say:
```
enum key_config_t {
  kShake_128s_pure,
  kShake_128s_prehashed,
  kSha2_128s_pure,
  kSha2_128s_prehashed,
  ...
}
```

Then plumb the [`config_id` field](https://github.com/lowRISC/opentitan/blob/6e698b4dfec2f9e61e091cbb9a866197681745f9/sw/device/silicon_creator/rom/sigverify_key_types.h#L223-L236) through and use it to call the verify function appropriately.

moidx: It is almost there... 

https://github.com/lowRISC/opentitan/blob/6e698b4dfec2f9e61e091cbb9a866197681745f9/sw/device/silicon_creator/rom/sigverify_key_types.h#L223-L236

sw/device/silicon_creator/lib/sigverify/spx_verify.c
```diff
@@ -62,20 +62,69 @@ static const uint32_t kSpxVerifyShares[kSigverifySpxRootNumWords] = {
...
+/**
+ * Domain-separation prefix for SPHINCS+ with SHA256 prehashing.
+ *
+ * The domain separation prefix is the following byte sequence:
+ *   0x01 || len(ctx) || ctx || OID(PH)
+ *
+ * In our case, `ctx` is always the empty string and PH (the pre-hashing
+ * function) is always SHA256.
+ */
+static const uint8_t kSpxVerifyPrehashDomainSep[] = {
```
moidx, line 87: I think this is the right byte ordering. Also the right OID identifier for `SHA-2-256`.

```
OID ← toByte(0x0609608648016503040201, 11) ▷ 2.16.840.1.101.3.4.2.1
```

moidx (approved):

jadephilipoom:
Moving out of draft since https://github.com/lowRISC/opentitan/pull/23762 is now merged

sw/device/silicon_creator/lib/sigverify/spx_verify.c
```diff
@@ -75,22 +75,51 @@ static const uint8_t kSpxVerifyPureDomainSep[] = {
...
+    0x01, 0x00, 0x06, 0x09, 0x60, 0x86, 0x48,
+    0x01, 0x65, 0x03, 0x04, 0x02, 0x01};
+
 rom_error_t sigverify_spx_verify(
     const sigverify_spx_signature_t *signature, const sigverify_spx_key_t *key,
-    lifecycle_state_t lc_state, const void *msg_prefix_1,
-    size_t msg_prefix_1_len, const void *msg_prefix_2, size_t msg_prefix_2_len,
-    const void *msg, size_t msg_len, uint32_t *flash_exec) {
+    const sigverify_spx_config_id_t *config, lifecycle_state_t lc_state,
```
cfrantz, line 93: The `config_id_t` is an integer; I don't think we need to pass a pointer to it here.

sw/device/silicon_creator/rom/rom.c
```diff
@@ -330,14 +330,15 @@ static rom_error_t rom_verify(const manifest_t *manifest,
       &ecdsa_key));
   // SPX+ key.
   const sigverify_spx_key_t *spx_key = NULL;
+  const sigverify_spx_config_id_t *spx_config = NULL;
```
cfrantz, line 333: Since this value is just an integer, we probably don't need a pointer to it here -- we can just get a copy of it from `spx_key_get` and pass-by-value to `spx_verify`.

If you're concerned with having an initial invalid value, zero or all-ones would be perfectly good as invalid values.

sw/host/opentitanlib/src/crypto/spx.rs
```diff
@@ -86,6 +94,13 @@ impl SpxKeypair {
         SpxSignature(sphincsplus::spx_sign(&self.sk, full_message).unwrap())
     }
 
+    /// Sign `message` using the secret key in "prehash" mode with an empty context.
+    pub fn sign_prehash(&self, oid: &[u8], message: &[u8]) -> SpxSignature {
```
cfrantz, line 98: How do you want to connect this up to the CLI?

Obviously, we could have a command-line switch that selects among raw, pure and pre-hash.  We could also attach some meta-data to the key to indicate its usage constraint.

jadephilipoom:
Resolves https://github.com/lowRISC/opentitan/issues/23144

In the TC we agreed to switch to SHA2 parameters and then implement an option that allowed us to select either SHA2 or SHAKE if there was enough time before ROM freeze and enough ROM space. However, I think at this point there is not enough time before ROM freeze -- I could _maybe_ implement the switching in time, but we'd be merging it at the very last second and I think we're better off focusing on testing this version instead. So once this is merged, the SPHINCS+ updates for EarlGrey prod should be complete.

sw/host/opentitanlib/src/crypto/spx.rs
jadephilipoom: Yes, I think a mode option with values raw, pure, and pre-hash, defaulting to pure, would make sense.

sw/host/opentitanlib/src/crypto/spx.rs
```diff
@@ -34,6 +34,14 @@ pub trait SpxPublicKeyPart {
         let full_message = &[domain_sep, message].concat();
         Ok(sphincsplus::spx_verify(self.pk(), &sig.0, full_message)?)
     }
+
+    // Verify a message signature in pre-hashed mode, returning Ok(()) if the
+    // signature matches.
+    fn verify_prehash(&self, oid: &[u8], message: &[u8], sig: &SpxSignature) -> Result<()> {
+        let domain_sep = &[1u8, 0u8];
```
cfrantz, line 41: I noticed that the `domain_sep` in the verify_prehash is different from the one in the `sign_prehash` function: `[1, 8]` vs `[0, 8]`.  Typo?

jadephilipoom: Yep, a typo. Fixed and I've now added a unit test here that makes sure `sign_prehash` and `verify_prehash` agree!

jadephilipoom:
I've added a new commit that just increases the timeout to "long" for the HMAC functest in silicon_creator; since I added a few tests, it seems like it's very close to the timeout limit and I saw it time out once here.

cfrantz (approved):

moidx (approved):
🎉🎉🎉🎉!

PR #23766, opened 2024-06-21
[rv_dm,dv] Implement sba_debug_disabled testpoint

### LINKED ISSUE #15668: [rv_dm,dv] Verify that access to SBA interface is disabled if lc_hw_debug_en is set to not strict true value (`sba_debug_disabled` BLT)
This issue tracks the implementation of the `sba_debug_disabled` block-level test, which is specified in the testplan for `rv_dm`.

>estimate 8

rswarbrick:
The actual test appears in the final commit, but there are some tidy-ups needed in the previous three. Detailed notes in commit messages.

Fixes #15668.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_sba_debug_disabled_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_sba_debug_disabled_vseq.sv
@@ -1,0 +1,25 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class rv_dm_sba_debug_disabled_vseq extends rv_dm_base_vseq;
+  `uvm_object_utils(rv_dm_sba_debug_disabled_vseq)
+  `uvm_object_new
+
+  // This overrides a constraint in the base vseq: we want to ensure that debug is *not* enabled for
+  // this vseq.
+  constraint debug_enabled_c {
+    lc_hw_debug_en != lc_ctrl_pkg::On;
+  }
+
+  task body();
+    sba_tl_device_seq_start();
+    repeat (10) begin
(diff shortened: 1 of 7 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_sba_debug_disabled_vseq.sv
```diff
@@ -0,0 +1,25 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class rv_dm_sba_debug_disabled_vseq extends rv_dm_base_vseq;
+  `uvm_object_utils(rv_dm_sba_debug_disabled_vseq)
+  `uvm_object_new
+
+  // This overrides a constraint in the base vseq: we want to ensure that debug is *not* enabled for
+  // this vseq.
```
vogelpi, line 10: According to the test plan, we also want to test that SBA works if debug is enabled. IIUC, this is not tested currently by this sequence. Is it necessary to add that or is it fine to add another test to the testpoint to verify the debug enabled case?

vogelpi (approved):
LGTM with one question, thanks @rswarbrick .

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_sba_debug_disabled_vseq.sv
rswarbrick: You're right that this is supposed to work, but it's a different item on the testplan. :-)

This test (`rv_dm_sba_debug_disabled`) addresses the `sba_debug_disabled` testpoint which, as you might expect, is all about making sure that SBA is disabled when debug is disabled.

Checking that SBA works when debug is *enabled* is the "sba" testpoint, which is addressed by `rv_dm_sba_tl_access` and `rv_dm_delayed_resp_sba_tl_access` (from squinting at the testplan).

It's possible I've missed something: please follow up if I have!

PR #23768, opened 2024-06-21
[dv,usbdev] Item CRC handling

alees24:
Expected CRC and checking of actual CRC against the expectation.
Informational reporting of CRC mismatches within the monitor. Scoreboard shall ultimately handle any mismatches.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_item.sv
+++ b/hw/dv/sv/usb20_agent/usb20_item.sv
@@ -63,6 +63,18 @@ class usb20_item extends uvm_sequence_item;
     return !(m_pid_type[3:0] ^ m_pid_type[7:4]);
   endfunction
 
+  // Check whether any CRC on this item is valid.
+  virtual function bit valid_crc();
+    `uvm_fatal(`gfn, "CRC relates only to packet items")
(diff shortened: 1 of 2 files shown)


hw/dv/sv/usb20_agent/usb20_item.sv
```diff
@@ -63,6 +63,16 @@ class usb20_item extends uvm_sequence_item;
     return !(m_pid_type[3:0] ^ m_pid_type[7:4]);
   endfunction
 
+  // Check whether any CRC on this item is valid.
+  virtual function bit valid_crc();
+    return 1'b1;
```
rswarbrick, line 68: Do you ever expect to use the base class implementation of `valid_crc` or `exp_crc`? If not, I'd suggest a fatal error here.

alees24: Fair point. That would, in practice, be an error; only packet events have CRCs.

rswarbrick: Cool, thanks.

rswarbrick (approved):

PR #23769, opened 2024-06-21
[dv,usbdev] Exercise all bus configurations

alees24:
There are three bus configuration options, leading to eight possible types of bus. Ensure that all are tested.

Modify the usb20_driver/monitor to be aware of pin-flipping when the TX side is using D/SE0 because this does invert the sense of the 'd' signal, since it is effectively driving DN now.


rswarbrick (approved):
Looks sensible to me!

Issue #22102, opened 2024-03-19
[hmac/dv] Verify wider digest & configurable key length


gdessouky:
### Description

Tentative List of TODOs:

- [x] Extend testplan and coverplan (add cover groups) to test and cover the additional digest sizes and key lengths.
- [x] Update the scoreboard and block-level tests to test the new digest sizes X key lengths and read digest CSRs between different digest size/key length configurations.
- [x] Verify the different digest sizes and key lengths in the test vectors tests https://github.com/lowRISC/opentitan/issues/22932
- ~[ ] Extend DIFs and TLTs to verify the additional digest sizes and key lengths supported~ https://github.com/lowRISC/opentitan/issues/22332#issuecomment-2149918855 --> we now have driver-based TLTs that accompanied the driver implementation and give us sufficient coverage at the top level. For M5, we plan to extend the DIFs and build DIF-based TLTs.
- [x] Verify unsupported digest sizes/key lengths once the error flagging to sw has been implemented.
- [x] Extend cryptoc_dpi to test wider digest sizes and different key lengths.
- [x] Stress test changing configuration digest size and key length while hashing.
- [ ] Evaluate coverage and run UNR when RTL is finalized and coverage is increased to generate the new exclusions list https://github.com/lowRISC/opentitan/issues/23683

gdessouky:
The key TODOs left here are:

`- [ ] Verify the different digest sizes and key lengths in the test vectors tests https://github.com/lowRISC/opentitan/issues/22932` which I plan to complete by today. I've already confirmed with @ballifatih that SHA-2 384/512 run with random test vectors interleaved, so I don't expect bugs requiring RTL fixes to surface here.

`- [ ] Extend testplan and coverplan (add cover groups) to test and cover the additional digest sizes and key lengths.` Cover groups have been added @martin-velay, and tests for this feature will be completed once I am done with the test vectors tests above. I only need to update the tests for this feature `wide_digest_configurable_key_length` in the testplan https://github.com/lowRISC/opentitan/blob/master/hw/ip/hmac/data/hmac_testplan.hjson#L119 also today.

[...]

Issue #23559, opened 2024-06-07
[hmac,dv] Fix hmac_stress_all_with_rand_reset V3 failing test


gdessouky:
### Description

Block-level `test hmac_stress_all_with_rand_reset`  test fails for HMAC because of testbench issues. The test exercises all block-level tests and triggers an almost random reset, but it waits for a number of cycles until there are no outstanding accesses in order to issue the reset here https://github.com/lowRISC/opentitan/blob/57f1155f7bb1e0c527ce0c40cdecf4d69aeb8a85/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv#L782

For smaller messages and tests, this passes, but once it exercises this for the very long msgs/test vectors, the test fails here https://github.com/lowRISC/opentitan/blob/57f1155f7bb1e0c527ce0c40cdecf4d69aeb8a85/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv#L828-L829 because it times out basically as it waits to trigger the reset, since there are always outstanding accesses.

I think we should be excluding tests that involve longer messages and test vectors for this one, given how it is implemented.

johngt:
Given that stress_all is a V3 item - moving this to M7.
Current work on coverage for V2 should be captured in other issues.

@moidx / @vogelpi / @andreaskurth

PR #23771, opened 2024-06-21
[hmac,dv] HMAC test vectors for block-level

gdessouky:
This adds HMAC NIST test vectors at block-level. Once merged, this would close https://github.com/lowRISC/opentitan/issues/22932 and thus also complete the tasks in issue https://github.com/lowRISC/opentitan/issues/22102.


hw/dv/sv/test_vectors/test_vectors.core
```diff
@@ -11,6 +11,7 @@ filesets:
     files:
       - test_vectors_pkg.sv
       - vectors/hmac/HMAC_RFC4868.rsp: {is_include_file: true}
+      - vectors/hmac/HMAC.rsp: {is_include_file: true}
```
martin-velay, line 14: I don't know so much in security yet, but don't you think that here we should precise which kind of HMAC we are talking about in the name itself (maybe add a suffix with SHA-2)? Maybe it's enough, I just want to double check with you

martin-velay (commented):
I think I have reviewed it a bit earlier, please could tag me once you've added the SV part of it, and I'll make a 2nd review round

hw/dv/sv/test_vectors/test_vectors.core
gdessouky: It was one file earlier with all test vectors for all digest sizes, but I decided to split it down now into 3 files, so now they are conveniently suffixed :-)

gdessouky:
I've rerun CI checks, but the fails here have occurred again and seem unrelated to this PR.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -468,13 +466,23 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
...
       // comparing all digest for SHA-2 512
       foreach (act_digest[i]) begin
-        if ((i  < 8) ||
-            ((i >= 8 && i < 12) && (digest_size_i == SHA2_384 || digest_size_i == SHA2_512)) ||
-            ((i >= 12)          && (digest_size_i == SHA2_512))) begin
-          `uvm_info(`gfn, $sformatf("Actual digest[%0d]: 0x%0h", i, act_digest[i]), UVM_HIGH)
-          `uvm_info(`gfn, $sformatf("Expected digest[%0d]: 0x%0h", i,
-                    packed_exp_digest[i]), UVM_HIGH)
-          `DV_CHECK_EQ(act_digest[i], packed_exp_digest[i], $sformatf("for index %0d", i))
+        if (hmac_en) begin // for HMAC test vectors only compare up to expected tag length
```
martin-velay, line 469: Instead of repeating the lines 472-475 with 412-484, maybe better to mutualize the if condition

rswarbrick:
I think the CI failure is "quick lint", which is failing because `hw/dv/sv/test_vectors/test_vectors_pkg.sv` doesn't have the standard licence header. (Easy enough to fix! You can just copy and paste from another file :-D)

hw/dv/sv/test_vectors/test_vectors_pkg.sv
```diff
@@ -1,4 +1,4 @@
-// Copyright lowRISC contributors (OpenTitan project).
+// Copyright lowRISC contributors (OpenTitan project)
```
rswarbrick, line 1: We've dropped a full stop here, which is what is failing the CI check.

hw/dv/sv/test_vectors/test_vectors_pkg.sv
```diff
@@ -334,6 +347,24 @@ package test_vectors_pkg;
           "KeyLen": begin
             vector.key_length_word = entry_data.atoi() / 32;
           end
+          "Klen": begin
+            // rounding up key length to nearest predefined key length
+            if (entry_data.atoi() <= 16) begin
```
rswarbrick, line 352: Minor nit: I'd suggest putting the integer value into a local variable to avoid parsing the string lots of times in a row.

martin-velay (approved):
Once the left comment has been taken into account and that regression is passing, this could be merged IMO

hw/dv/sv/test_vectors/test_vectors_pkg.sv
rswarbrick, line 352: It might also make sense to check that the string is "0" if `entry_data.atoi()` returns 0 (raising an error otherwise). Otherwise, I think you'll silently ignore things like "cabbage".

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -468,13 +466,23 @@ class hmac_base_vseq extends cip_base_vseq #(.CFG_T               (hmac_env_cfg)
...
+          // compute word index to compare up to (number of bytes/4)
+          if (i < (tag_len_byte/4)) begin
```
rswarbrick, line 471: Do we know that `tag_len_byte` is divisible by 4? If so, maybe we should actually pass something like `tag_length_words`. If not, I think you probably want explicit handling for when it is not.

gdessouky:
> I think the CI failure is "quick lint", which is failing because `hw/dv/sv/test_vectors/test_vectors_pkg.sv` doesn't have the standard licence header. (Easy enough to fix! You can just copy and paste from another file :-D)

oh yes thanks, this time I can spot this in the logs :D I was getting another issue in the last CI run.

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -9,9 +9,9 @@ class hmac_test_vectors_sha_vseq extends hmac_base_vseq;
   `uvm_object_utils(hmac_test_vectors_sha_vseq)
   `uvm_object_new
 
-  string vector_list_256[] = test_vectors_pkg::sha2_256_file_list;
-  string vector_list_384[] = test_vectors_pkg::sha2_384_file_list;
-  string vector_list_512[] = test_vectors_pkg::sha2_512_file_list;
+  string vector_list_256[]  = test_vectors_pkg::sha2_256_file_list;
+  string vector_list_384[]  = test_vectors_pkg::sha2_384_file_list;
+  string vector_list_512[]  = test_vectors_pkg::sha2_512_file_list;
```
rswarbrick, line 14: Nit: We seem to have added some extra spaces here. If you're force-pushing anyway, maybe it makes sense to drop them.

rswarbrick (commented):
Lots of nitty comments, but this looks sensible to me.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
gdessouky: the tag/digest length (whether for the HMAC test vectors or other test vectors) is in bytes and so I just passed it from the parsing in bytes still, and the tag length is word-aligned, and always divisible by 4, but can add a safety check just in case.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
gdessouky: I only want to print out the expected/actual digests when we actually compare them which is why I bundle the print statements after the check (and now in retrospect I would move them up before the check). Does that make sense?

gdessouky:
Thanks @rswarbrick  and @martin-velay for your reviews, I think I've addressed most of your comments and pushed again now and also running a block-level regression at my end.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
martin-velay: I meant something like that (please check if that's correct, I wrote it a bit quickly):
```
  virtual task compare_digest(bit [7:0] exp_digest[], int tag_len_byte, bit [3:0] digest_size_i);
    bit [TL_DW-1:0] act_digest[16];
    bit [TL_DW-1:0] packed_exp_digest[16];
    csr_rd_digest(act_digest);
    // `exp_digest` is guaranteed to always contain 16 words (64 bytes) of data
    // since HMAC digest size is max 512 bits.
    packed_exp_digest = {>>byte{exp_digest}};
    if (cfg.clk_rst_vif.rst_n) begin
      // comparing only digest[0] to digest [7] for SHA-2 256
      // comparing only digest[0] to digest [12] for SHA-2 384
      // comparing all digest for SHA-2 512
      foreach (act_digest[i]) begin
        // for HMAC test vectors only:
        //   - compare up to expected tag length
        //   - which is always divisble by 4 (word-aligned)
        //   - compute word index to compare up to (number of bytes/4)
        // For SHA in standalone (hmac_en=0) test vectors:
        //   - compare up to the desired digest size
        if (( hmac_en && (i < (tag_len_byte/4))) ||
            (!hmac_en && 
              ((i < 8) ||
              ((i >= 8 && i < 12) && (digest_size_i == SHA2_384 || digest_size_i == SHA2_512)) ||
              ((i >= 12)          && (digest_size_i == SHA2_512))))) begin
          `uvm_info(`gfn, $sformatf("Actual digest[%0d]: 0x%0h", i, act_digest[i]), UVM_HIGH)
          `uvm_info(`gfn, $sformatf("Expected digest[%0d]: 0x%0h", i,
                    packed_exp_digest[i]), UVM_HIGH)
          `DV_CHECK_EQ(act_digest[i], packed_exp_digest[i], $sformatf("for index %0d", i))
        end
      end
    end else begin
      `uvm_info(`gfn, "skipped comparison due to reset", UVM_LOW)
    end
  endtask
```

martin-velay (approved):

hw/ip/hmac/dv/hmac_sim_cfg.hjson
```diff
@@ -113,9 +113,21 @@
     }
 
     {
-      name: hmac_test_hmac_vectors
+      name: hmac_test_hmac256_vectors
```
martin-velay, line 116: Sorry, I have 2 more points to mention for this PR, as discussed together:
- I ran a regression and the timeout `+test_timeout_ns=1_000_000_000` for SHA 384/512 is too short, as a `uvm_fatal` is raised
- And also, those tests are quite long and according to my understanding the randomization scope is restricted. Indeed, if I am not wrong, only those variables are randomized:  message, wr_config_during_hash, wr_key_during_hash, key, and endian/digest/key_swap. Thus, I would suggest to reduce the `reseed` parameter to 10 for the tests with the NIST vectors. @gdessouky, @vogelpi, @andreaskurth: what do you think?

martin-velay (approved):

hw/dv/sv/test_vectors/test_vectors_pkg.sv
rswarbrick: Hmm, I'm not sure this is doing quite the right thing. How do I get to the trailing else in the `if` block now, representing a key bigger than 1024 bits?

hw/ip/hmac/dv/hmac_sim_cfg.hjson
martin-velay: Something else occurred to me, as `hmac_test_vectors_sha_vseq` and `hmac_test_vectors_hmac_vseq` are taking ages to run:

Do we need to run them from the `hmac_stress_all_vseq`? I think the added value is not that significant as tested in standalone and mixing it don't bring so much.

If you have a strong opinion to keep it, maybe we can reduce their weight? (to do so, I would suggest the use of a SV weighted `randcase`)

Again, what do you think: @gdessouky , @vogelpi and @andreaskurth ?

vogelpi: I agree that that 1) lowering the reseed parameter makes sense as the value of these tests is limited and 2) that we should also lower the weight of these tests in the stress_all sequence. These are both good proposals.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
gdessouky: Oh, should have clarified. I had added it here https://github.com/lowRISC/opentitan/pull/23771/files#diff-020f3f89cd4e24b16efe3707042675c21aa6d05c24a8c0a6e44ab0259448a2b4R402-R403 already while parsing, but can also do it in `hmac_base_vseq.sv` instead. Is there a preference/does it matter?

hw/dv/sv/test_vectors/test_vectors_pkg.sv
gdessouky: I'm not sure I follow, in the last pushed version of the code, if `key_length_byte > 128`, I would end up in the last trailing else block where `vector.key_length_word = 0;` - why do I not get to it?

gdessouky:
> This is failing CI because `hw/ip/hmac.h` has the wrong header. But it shouldn't be at that path either. Please could you move it somewhere sensible?

that wasn't supposed to get committed actually, that's why CI is so great :-)

hw/ip/hmac/dv/hmac_sim_cfg.hjson
gdessouky: @martin-velay I agree that we should go ahead and exclude them from `hmac_stress_all_vseq`.  In fact, this would actually resolve this issue as well https://github.com/lowRISC/opentitan/issues/23559 with the `hmac_stress_all_with_rand_reset` failing test, or at least partially. We might also need to exclude other tests with long messages for that one as well.

hw/ip/hmac/dv/hmac_sim_cfg.hjson
```diff
@@ -113,9 +113,21 @@
...
       uvm_test_seq: hmac_test_vectors_hmac_vseq
-      run_opts: ["+test_vectors_dir={build_dir}/src/lowrisc_dv_test_vectors_0 +sha2_digest_size=SHA2_256"]
+      run_opts: ["+test_vectors_dir={build_dir}/src/lowrisc_dv_test_vectors_0 +test_timeout_ns=1_000_000_000 +sha2_digest_size=SHA2_256"]
+    }
+
+    {
+      name: hmac_test_hmac384_vectors
+      uvm_test_seq: hmac_test_vectors_hmac_vseq
+      run_opts: ["+test_vectors_dir={build_dir}/src/lowrisc_dv_test_vectors_0 +test_timeout_ns=2_000_000_000 +sha2_digest_size=SHA2_384"]
```
gdessouky, line 124: @martin-velay I extend the timeout here for 384 and 512, I guess this should cover it

hw/dv/sv/test_vectors/test_vectors_pkg.sv
```diff
@@ -361,7 +397,13 @@ package test_vectors_pkg;
             // If a global output length is defined, this case will never be reached.
             vector.digest_length_byte = entry_data.atoi() / 8;
           end
-          "MD", "Output": begin
+          "Tlen": begin
+            vector.digest_length_byte = entry_data.atoi();
+            if (vector.digest_length_byte % 4 !=0) begin
```
gdessouky, line 402: @rswarbrick this is where I had added a safety check to check on the digest byte length, do you think this is OK or is it better placed in the vseq?

hw/ip/hmac/dv/hmac_sim_cfg.hjson
vogelpi: This sounds acceptable to me as well. On the other hand, increasing the wait time for outstanding accesses could be an easy fix for #23559.

martin-velay: Another think I would like to say after talking with Ghada this morning. I didn't realized but also at least the key and the message and the swap are not randomized as the NIST is providing fixed test vectors and not a model. So maybe even 5 seeds per test is enough, as few space is left for the randomization.

hw/dv/sv/test_vectors/test_vectors_pkg.sv
rswarbrick: My question was a bit silly, sorry! Embarrassingly, I'd forgotten that there are numbers greater than 128 :-) Your code looks perfectly sensible to me!

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
rswarbrick: Eh, maybe it's nice to move to the vseq, to avoid strange things happening if we mess up and cause the value to change after the check you've added (or use a code path that we hadn't thought about, setting the value in a completely different way and not doing the check).

But it doesn't really matter and I'm convinced this is good to go. Let's merge :-)

rswarbrick (approved):
Thanks for all the responses to my review. This looks good to me.

PR #23772, opened 2024-06-21
Fix i2c testutils

pamaury:
The PR fixes the following tests:
- `i2c_host_eeprom_test`
- `i2c_host_hdc1080_humidity_temp_test`
- `i2c_host_gas_sensor_test`
- `i2c_host_irq_test`

This is done by updating the `i2c_testutils_wait_transaction_finish` to handle halt events correctly. The test `i2c_host_irq_test` was completely broken since it didn't compile and required a small change (see commit).

TODO:
- [ ] `i2c_host_irq_test` is not part of any testplan: add it
- [ ] `i2c_host_fram_test` is still broken

### DIFF
--- a/sw/device/lib/dif/dif_i2c.c
+++ b/sw/device/lib/dif/dif_i2c.c
@@ -234,6 +234,17 @@ dif_result_t dif_i2c_compute_timing(dif_i2c_timing_config_t timing_config,
     config->scl_time_high_cycles = (uint16_t)lengthened_high_cycles;
   }
 
+  // For clock stretching detection to work, the SCL high and low time must be
+  // at leats 4 cycles.
+  // TODO The i2c IP has a parameter (InputDelayCycles) for that, use this when
+  // topgen supports exposing it in the headers (Issue#23786).
+  if (config->scl_time_high_cycles < 4) {
+    config->scl_time_high_cycles = 4;
+  }
+  if (config->scl_time_low_cycles < 4) {
+    config->scl_time_low_cycles = 4;
+  }
+
   return kDifOk;
 }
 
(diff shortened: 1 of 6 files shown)


a-will (approved):

jwnrt (approved):

a-will:
FYI, `i2c_host_clock_stretching_test` uses illegal timing values:

```
I00029 i2c_testutils.c:390] Setting i2c to FastPlus (1000kHz) mode.
I00030 i2c_testutils.c:414] period:166 nanos, cycles={fall=1, rise=3, hi=2, lo=4}
```

The minimum high time is 4 cycles, I believe.

pamaury:
> 
> The minimum high time is 4 cycles, I believe.

I think you are right, the Programmer's guide says:
```
To guarantee clock stretching works correctly in Controller-Mode, there is a requirement of THIGH >= 4. This constraint derives from the fact that there is a latency between the Controller FSM driving the bus and observing the effect of driving the bus. The implementation requires THIGH to be at least this large to guarantee that if the Target stretches the clock, we can observe it in time, and react accordingly.
```

By the way, this is a somewhat unexpected place to put this information because it's not in the Theory of Operation and not in the Register description.

pamaury:
After discussion with @a-will , he has created an issue #23786 about the fact that we can't access the i2c input delay parameter. Since this is not a priority, I have just hardcoded the minimum value of low/high time to 4 cycles per his advice.

pamaury:
It seems that the timing calculation changes has broken `i2c_host_fram_test.c` because the thoughput test doesn't take this minimum low/high cycle into account. I will need to fix it before merging.

pamaury:
I have updated the FRAM test. I went for the simplest option which is to slightly lower the expected throughput. In theory, a better solution would be to look at the computed i2c timing and compute the throughput from that but this is a lot more work but little benefits in this case I think.

pamaury:
Now the i2c_host_eeprom_test is failing in CI but I haven't been able to reproduce it locally yet.

pamaury:
I have merged despite the failure since it still fixes 3 tests. I will look specifically at why `i2c_host_eeprom_test` is flaky.

PR #23775, opened 2024-06-21
[entropy_src/rtl] Delay the main_sm_done signal

### LINKED ISSUE #23773: [entropy_src] Delay ht_done_pulse_i for the main SM by 1 cycle
### Description

Moving the 1-4-bit packer back in the pipeline #21626 introduced an issue where the `ht_done_pulse_i` can go high 1 cycle too early. This in combination with some other recent changes caused some unwanted words to be absorbed by SHA3.

When `ht_done_pulse_i` goes high this causes `process_i` for SHA3 to go high, which in turn leads to SHA3 not accepting any new inputs. If these signals go high 1 cycle too early SHA3 doesn't absorb the last word. The result were some test sequences that were failing due to a mismatch in the produced seeds.

I only observed it happening for the single lane mode, where one of the 4 random bit lanes is selected. This might be because the `post_ht` FIFO still receives the entropy without delay in normal operation.

Unfortunately I can't post a failing seed here for the `entropy_src_rng` test, since I had to change a lot in the DV environment to discover this bug.

h-filali:
Due to the relocation of the 1-4-bit packer FIFO, the main SM done signal arrived one cycle too early in the single lane mode. This caused the SHA3 process_i signal to go high one cycle too early. This in turn caused the issue that the last word from the precon FIFO is not absorbed properly by SHA3.
This commit fixes this by delaying the main_sm_done signal by 1 cycle in the single lane mode. The ht_error signal has to be delayed by 1 cycle as well to guarantee that the error signal arrives at the same time as the main_sm_done signal.

Resolves #23773

### DIFF
--- a/hw/ip/entropy_src/rtl/entropy_src_core.sv
+++ b/hw/ip/entropy_src/rtl/entropy_src_core.sv
@@ -463,6 +463,8 @@ module entropy_src_core import entropy_src_pkg::*; #(
 
   logic                    stale_seed_processing;
   logic                    main_sm_enable;
+  logic                    main_sm_done_pulse;
+  logic                    main_sm_ht_failed;
 
   logic                    unused_err_code_test_bit;
   logic                    unused_sha3_state;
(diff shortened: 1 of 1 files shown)


vogelpi:
CHANGE AUTHORIZED: hw/ip/entropy_src/rtl/entropy_src_core.sv

vogelpi (approved):
Thanks for debugging and fixing this @h-filali . This fixes exactly the identified issue but doesn't touch anything else. The change looks good to me.

In case we do another synthesis run, I am strongly in favor of including this.

moidx:
CHANGE AUTHORIZED: hw/ip/entropy_src/rtl/entropy_src_core.sv

moidx (approved):

PR #23776, opened 2024-06-21
[dv,usbdev] rx_pid_err sequence

alees24:
Include sequence to check that the 'rx_pid_err' interrupt is raised in response to invalid Packet IDentifiers being received. 
Applies to both token packets and data packets.


hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -516,6 +516,29 @@ endtask
     end
   endtask
 
+  // Decide upon one or more corrupted PID values when constructing an invalid OUT/SETUP
+  // transaction.
+  function void build_prnd_bad_pids(output bit [1:0] corrupt, output bit is_setup,
+                                    output bit [7:0] setup_out_pid, output bit [7:0] data_pid);
+    // Which PID(s) to corrupt?
+    corrupt = $urandom_range(1, 3); // bit 1 = corrupt data, bit 0 = setup/out
```
rswarbrick, line 524: I think it would probably make sense to give names to these values, with something like
```systemverilog
bit corrupt_out_pid = $urandom & 1;
bit corrupt_data_pid = (! corrupt_out_pid) || $urandom & 1;
```

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -516,6 +516,29 @@ endtask
     end
   endtask
 
+  // Decide upon one or more corrupted PID values when constructing an invalid OUT/SETUP
+  // transaction.
+  function void build_prnd_bad_pids(output bit [1:0] corrupt, output bit is_setup,
```
rswarbrick, line 521: Do you actually use the `corrupt` output argument? I don't think I could see its use at the two call sites?

rswarbrick (approved):
I think the function that does the corrupting could probably be a bit simpler (described in the comments), but this looks perfectly reasonable either way.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: usbdev_bad_traffic_vseq does use one of the bits; it needs to know whether a valid SETUP token packet was sent to the DUT.

Issue #22098, opened 2024-03-19
[ROM] Enable Schmitt trigger in pads (via pinmux) where beneficial


andreaskurth:
Earlgrey's pads feature a [Schmitt trigger](https://en.wikipedia.org/wiki/Schmitt_trigger), which can be enabled per [DIO](https://opentitan.org/book/hw/ip/pinmux/doc/registers.html#dio_pad_attr--schmitt_en) and [MIO](https://opentitan.org/book/hw/ip/pinmux/doc/registers.html#mio_pad_attr--schmitt_en) in pinmux and which can improve signal integrity for certain use cases / protocols.

We should go through our ROM code and check
- which DIOs and MIOs ROM uses
- for which DIOs and MIOs ROM locks the `PAD_ATTR` register (by setting the corresponding `REGWEN`)

and investigate if any of the uses / locks would benefit from having Schmitt trigger enabled. (One example that was mentioned might be SPI Clock.) Where it is beneficial, we should enable the Schmitt trigger.

We should check the timing tables in the datasheets of the relevant target devices and compare them to the characteristics of the Schmitt trigger in Earlgrey before we commit to this change. If the FPGA pads feature a comparable Schmitt trigger (to be checked), we should additionally test this with the relevant target devices connected to Earlgrey on an FPGA.

For DIOs and MIOs not used or locked by ROM, we can later enable the Schmitt trigger in mutable code if we find that the need arises.

andreaskurth:
To be resolved before ROM release --> assigned to M5.

cfrantz:
I think we should make the enabling of Schmitt triggers by ROM an OTP configuration option.

This means we don't have to decide now exactly which pads would benefit from Schmitt triggers and allows the sku_creator to decide what is appropriate for their environment.

a-will:
> and investigate if any of the uses / locks would benefit from having Schmitt trigger enabled. (One example that was mentioned might be SPI Clock.) Where it is beneficial, we should enable the Schmitt trigger.

SPI_DEV_CLK almost certainly should _not_ have the schmitt trigger enabled, not unless they have only a very short delay. This is a high-performance input in many cases, and delays are deleterious to transfer rates. Instead, the board design should handle signal integrity so the channel performs as desired.

cfrantz:
> > and investigate if any of the uses / locks would benefit from having Schmitt trigger enabled. (One example that was mentioned might be SPI Clock.) Where it is beneficial, we should enable the Schmitt trigger.
[...]

PR #23777, opened 2024-06-21
[rom] Configure Schmitt triggers on IO pads 

cfrantz:
1. Enhance the pinmux driver to read the Schmitt configuration from OTP
   and configure the corresponding pinmux registers.
2. Initialize the Schmitt triggers when initializing the pinmux in the ROM.

Addresses #22098


a-will (approved):
LGTM

I'm not sure how useful the extra noise margin will be for most of these, but maybe the cost isn't a big deal. The most critical signals (POR_N, JTAG TCK) already have schmitt triggers forced on.

The straps and UART probably don't need them. Maybe SPI bootstrap would have a little extra help if the board design is more marginal.

moidx (approved):

PR #23779, opened 2024-06-21
[d,alert_ping] Detect ping responses

### LINKED ISSUE #23644: [dv,alert_if] Detect alerts that are ping responses
### Description

Some top-level tests fail if they get incoming ping requests, because the outgoing active alert responding to the ping is mistaken for a fault. The problem is that the cip_lib's cip_base_vseq doesn't filter ping responses. This filtering should be provided by the alert_esc_if interface.

matutem:
Outgoing alerts can be in response to a ping request or real alerts. Make changes in alert_esc_if to detect ping responses so the cip infrastructure can avoid spurious alerts.

Change the chip_power_load test to re-enable pings.

Fixes #23644

### DIFF
--- a/hw/dv/sv/alert_esc_agent/alert_esc_if.sv
+++ b/hw/dv/sv/alert_esc_agent/alert_esc_if.sv
@@ -6,6 +6,8 @@
 // Alert interface
 // ---------------------------------------------
 interface alert_esc_if(input clk, input rst_n);
+  import uvm_pkg::*;
+
   wire prim_alert_pkg::alert_tx_t alert_tx;
   wire prim_alert_pkg::alert_rx_t alert_rx;
   wire prim_esc_pkg::esc_tx_t     esc_tx;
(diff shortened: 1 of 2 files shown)


hw/dv/sv/alert_esc_agent/alert_esc_if.sv
```diff
@@ -31,7 +33,41 @@ interface alert_esc_if(input clk, input rst_n);
...
   clk_rst_if              clk_rst_async_if(.clk(async_clk), .rst_n(rst_n));
 
-  // if alert sender is async mode, the clock will be drived in alert_esc_agent,
+  logic ping_pq;
+  always_ff @(posedge clk or negedge rst_n) begin
+    ping_pq <= alert_rx.ping_p;
+  end
+
+  logic ping_request_d;
+  always_comb ping_request_d = alert_rx.ping_p != ping_pq && !$isunknown(ping_pq);
```
rswarbrick, line 42: Thinking about it, we could avoid the "=== stuff" by writing this as something like `ping_request_d = alert_rx.ping_p ^ ping_pq;`

That said, the resulting X would pollute `state_d` and `state_q` and I think the FSM has no way to leave the X state. Maybe your code here is better, but I thought it was at least worth pointing out.

hw/dv/sv/alert_esc_agent/alert_esc_if.sv
```diff
@@ -94,7 +130,14 @@ interface alert_esc_if(input clk, input rst_n);
   endtask : wait_esc_complete
 
   function automatic bit get_alert();
-    get_alert = (alert_tx_final.alert_p === 1'b1 && alert_tx_final.alert_n === 1'b0);
+    if (alert_tx_final.alert_p === 1'b1) begin
+      `uvm_info("alert_esc_if::get_alert", $sformatf(
+                "alert_p:%b, alert_n:%b, ping_pending:%p",
+                alert_tx_final.alert_p, alert_tx_final.alert_n, ping_pending),
+                UVM_MEDIUM)
+    end
```
rswarbrick, line 138: After a quick grep for `get_alert` in the code base, I think this print will happen quite a bit. Is that intended?

rswarbrick (approved):
This looks sensible to me. A couple of questions, but I'm very happy for it to go in.

hw/dv/sv/alert_esc_agent/alert_esc_if.sv
matutem: I changed it so it will only print something for ping responses, not regular alerts.

PR #23785, opened 2024-06-24
[dv/chip] some required changes to chip_if and chip tb

eshapira:


sha-ron (approved):

hw/top_earlgrey/dv/tb/tb.sv
```diff
@@ -339,8 +351,13 @@ module tb;
     .BAUD('d1_000_000),
     .FREQ('d24_000_000)
   ) u_uartdpi0(
-    .clk_i(dut.top_earlgrey.u_clkmgr_aon.clocks_o.clk_io_div4_peri),
-    .rst_ni(dut.top_earlgrey.u_rstmgr_aon.resets_o.rst_lc_io_div4_n[rstmgr_pkg::Domain0Sel]),
+    `ifdef GATE_LEVEL
+       .clk_i(dut.top_earlgrey.u_clkmgr_aon.clocks_o_clk_io_div4_peri),
```
matutem, line 355: This clk_i is connected the same regardless of GATEL_LEVEL, so please take it out of the `ifdef.

eshapira: > This clk_i is connected the same regardless of GATEL_LEVEL, so please take it out of the `ifdef.

Thanks for your review!
Please note that the signal is a bit different:
u_clkmgr_aon.clocks_**o_c**lk_io_div4_peri  !=
u_clkmgr_aon.clocks_**o.c**lk_io_div4_peri

matutem: Got it, thanks.

hw/top_earlgrey/dv/tb/tb.sv
```diff
@@ -259,8 +259,13 @@ module tb;
 
   // The gpiodpi module allows the host to directly control gpio when enabled.
   gpiodpi u_gpiodpi(
-    .clk_i(dut.top_earlgrey.u_clkmgr_aon.clocks_o.clk_io_div4_peri),
-    .rst_ni(dut.top_earlgrey.u_rstmgr_aon.resets_o.rst_lc_io_div4_n[rstmgr_pkg::Domain0Sel]),
+    `ifdef GATE_LEVEL
```
matutem, line 262: I don't know if we have guidelines for ifdef indentation, but I tend to align them to the start of line, and indent by two if there are nested backtick directives. This is because backticks are global in scope.

matutem (approved):
Please consider the two suggesions. Otherwise LGTM.

hw/top_earlgrey/dv/tb/tb.sv
eshapira: Done. Thanks!

PR #23788, opened 2024-06-24
[rv_dm,dv] Implement sec_cm_lc_hw_debug_en_intersig_mubi testpoint

rswarbrick:
This builds on #23766, which is the first 4 commits. Only the final 3 commits are unique to this PR.

The first two commits are about strengthening the scoreboard slightly so that it can check the properties we want to test in the testpoint. The final commit implements most of the testpoint, falling back to rv_dm_sba_debug_disabled (which was implemented in commit 4) for one of the properties we want to assert.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_debug_disabled_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_debug_disabled_vseq.sv
@@ -1,0 +1,70 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class rv_dm_debug_disabled_vseq extends rv_dm_base_vseq;
+  `uvm_object_utils(rv_dm_debug_disabled_vseq)
+  `uvm_object_new
+
+  // Check that the "mem" TL interface is disabled as expected
+  task check_no_mem_if();
+    // Try to do an arbitrary TL access to the interface. We don't expect it to work, which is
+    // checked in the scoreboard (in predict_tl_err).
+    csr_wr(.ptr(tl_mem_ral.halted), .value(0));
+  endtask
+
+  // Check that some signal that comes out of dm_top doesn't make it as far as the rv_dm toplevel
+  task check_no_dm_output(string         dm_top_port,
+                          string         rv_dm_port,
+                          uvm_reg_data_t forced_input,
+                          uvm_reg_data_t expected_output);
+    string in_path = {"tb.dut.u_dm_top.", dm_top_port};
+    string out_path = {"tb.dut.", rv_dm_port};
+
+    `DV_CHECK(uvm_hdl_force(in_path, forced_input))
+    repeat (100) begin
+      uvm_reg_data_t rvalue;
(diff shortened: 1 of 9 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_debug_disabled_vseq.sv
```diff
@@ -0,0 +1,70 @@
...
+
+class rv_dm_debug_disabled_vseq extends rv_dm_base_vseq;
+  `uvm_object_utils(rv_dm_debug_disabled_vseq)
+  `uvm_object_new
+
+  // Check that the "mem" TL interface is disabled as expected
+  task check_no_mem_if();
+    // Try to do an arbitrary TL access to the interface. We don't expect it to work, which is
+    // checked in the scoreboard (in predict_tl_err).
+    csr_wr(.ptr(tl_mem_ral.halted), .value(0));
```
vogelpi, line 13: Do we need to randomize the address or data, read or write? I don't think so :-)

vogelpi (approved):
This looks good to me, thanks @rswarbrick !

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_debug_disabled_vseq.sv
rswarbrick: I wasn't planning to do that, so won't catch a situation where there's a particular address that works when others are correctly dropped. But that seems rather implausible, so I didn't think there was any point :-)

Of course, it's possible there's some clever reason that we might need to do different types of bus accesses. Follow up if so, and I'll work out how to generalise this (as a follow-up).

a-will:
Too quick with the merge button here. The tests hadn't run yet.

hw/ip/rv_dm/dv/sva/rv_dm_enable_checker.sv
```diff
@@ -0,0 +1,39 @@
...
+  // If debug is not enabled then the mem TL interface is disabled and will respond to everything
+  // with an error. This means that any response will have d_error=1.
+  `ASSERT(MemTLResponseWithoutDebugIsError_A,
+          mem_tl_d_o_i.d_valid && !lc_tx_test_true_strict(lc_hw_debug_en_i) |->
+          mem_tl_d_o_i.d_error)
+
+  // If debug is not enabled then the SBA TL interface is disabled and we will never generate a new
+  // TL transaction. As such, the a_valid signal will always be false.
+  `ASSERT(SbaTLRequestNeedsDebug_A,
+          sba_tl_h_o.a_valid |-> lc_tx_test_true_strict(lc_hw_debug_en_i))
```
a-will, line 38: Was this supposed to be...

```suggestion
  `ASSERT(SbaTLRequestNeedsDebug_A,
          sba_tl_h_o_i.a_valid |-> lc_tx_test_true_strict(lc_hw_debug_en_i))
```

rswarbrick: It was, thanks. And I'm really sorry for messing this up. Bizarrely, the incorrect code builds just fine with Xcelium. I think it must be that the checker module is getting bound into `rv_dm` (where `sba_tl_h_o` does exist and indeed is the source for that port).

Sorry for the noise. :-(

PR #23789, opened 2024-06-24
[rv_dm,dv] Connect up sec_cm_bus_integrity testpoint

rswarbrick:
This is implemented by a generic test (in
cip_base_vseq__tl_errors.svh) and the only work needed for rv_dm is to name it in the testplan. Do so.


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #23791, opened 2024-06-25
[pinmux,top_earlgrey] Add ScanClock role to SPI_DEV_CLK and filter from flop inputs during DFT

a-will:
Add a ScanClock role to the scan_role_e enum, then add the scan roles of all pinmux-connected DIOs and MIOs to the pinmux. Filter any DIO or MIO with the ScanClock role from connecting to the wakeup detector flops when scanmode is active.

Use SPI_DEV_CLK as the scan clock.

### DIFF
--- a/hw/ip/pinmux/rtl/pinmux.sv
+++ b/hw/ip/pinmux/rtl/pinmux.sv
@@ -543,17 +543,47 @@ module pinmux
   // Wakeup detectors //
   //////////////////////
 
+  // Wakeup detectors should not be connected to the scan clock, so filter
+  // those inputs.
+  logic [NDioPads-1:0] dio_wkup_no_scan;
+  for (genvar k = 0; k < NDioPads; k++) begin : gen_dio_wkup_filter
+    if (TargetCfg.dio_scan_role[k] == ScanClock) begin : gen_dio_scan
+      always_comb begin
+        dio_wkup_no_scan[k] = dio_in_i[k];
+        if (prim_mubi_pkg::mubi4_test_true_strict(scanmode_i)) begin
+          dio_wkup_no_scan[k] = 1'b0;
+        end
+      end
+    end else begin : gen_no_dio_scan
+      assign dio_wkup_no_scan[k] = dio_in_i[k];
+    end
+  end
+
+  logic [NMioPads-1:0] mio_wkup_no_scan;
+  for (genvar k = 0; k < NMioPads; k++) begin : gen_mio_wkup_filter
+    if (TargetCfg.mio_scan_role[k] == ScanClock) begin : gen_mio_scan
+      always_comb begin
+        mio_wkup_no_scan[k] = mio_in_i[k];
+        if (prim_mubi_pkg::mubi4_test_true_strict(scanmode_i)) begin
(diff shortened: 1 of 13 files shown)


meisnere (approved):

hw/ip/pinmux/rtl/pinmux.sv
```diff
@@ -543,17 +543,38 @@ module pinmux
...
 
+  // Wakeup detectors should not be connected to the scan clock, so filter
+  // those inputs.
+  logic [NDioPads-1:0] dio_wkup_no_scan;
+  logic [NMioPads-1:0] mio_wkup_no_scan;
+  always_comb begin
+    for (int k = 0; k < NDioPads; k++) begin
+      dio_wkup_no_scan[k] = dio_in_i[k];
+      if (prim_mubi_pkg::mubi4_test_true_strict(scanmode_i) &&
+          TargetCfg.dio_scan_role[k] == ScanClock) begin
```
msfschaffner, line 554: nit: to make it easier for DV to cover, we may want to formulate this as a generate loop for the static stuff, and then use a ternary pr always_comb if inside that for scanmode_i.
```systemverilog
for (int k = 0; k < NDioPads; k++) begin : gen_miodio
  if (TargetCfg.dio_scan_role[k] == ScanClock) begin : gen_dio_scan
    dio_wkup_no_scan[k] = prim_mubi_pkg::mubi4_test_true_strict(scanmode_i) ? 1'b0 : dio_in_i[k];   
  end else begin : gen_no_dio_scan
    dio_wkup_no_scan[k] = dio_in_i[k];   
  end
end
```

msfschaffner (approved):
LGTM

a-will:
@meisnere Note this could need changes in partner-provided `scan_role_pkg` and `prim_pad_wrapper`. `scan_role_pkg` would need the desired pad to be set to `ScanClock`, and I assume `ScanClock` would want to be handled like `ScanIn` in the pad wrapper.

hw/ip/pinmux/rtl/pinmux.sv
a-will: Done, thanks!

vogelpi:
CHANGE AUTHORIZED: hw/ip/pinmux/rtl/pinmux.sv
CHANGE AUTHORIZED: hw/ip/pinmux/rtl/pinmux_pkg.sv
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_pad_wrapper_pkg.sv

moidx:
CHANGE AUTHORIZED: hw/ip/pinmux/rtl/pinmux.sv
CHANGE AUTHORIZED: hw/ip/pinmux/rtl/pinmux_pkg.sv
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_pad_wrapper_pkg.sv

moidx (approved):

moidx:
Hi @meisnere, please sync with @a-will in case there is help needed to synchronize these changes in your instantiation of the `prim_pad_wrapper`.

a-will:
The RV_DM block-level test failures are unrelated. They should be fixed in https://github.com/lowRISC/opentitan/pull/23803

PR #23792, opened 2024-06-25
[rv_dm,dv] Support "late enable through register" mechanism

rswarbrick:
The main change here is the second commit, with the following message:

The logic for whether debug is enabled in the dut looks like:

```
  if ((otp_dis_rv_dm_late_debug_i is mubi true) ||
      (the late_debug_enable register is mubi true)) {
    enabled = lc_hw_debug_en_i;
  } else {
    enabled = lc_dft_en_i;
  }
```

Before this commit, we always set `otp_dis_rv_dm_late_debug_i` to true. This commit makes it possible to depend on the register because `otp_dis_rv_dm_late_debug_i` doesn't have the "true" value.

We still aren't handling the situation where neither is a mubi true value, but that will come in a following commit.

### DIFF
--- a/hw/ip/rv_dm/dv/env/rv_dm_scoreboard.sv
+++ b/hw/ip/rv_dm/dv/env/rv_dm_scoreboard.sv
@@ -322,6 +322,8 @@ class rv_dm_scoreboard extends cip_base_scoreboard #(
         case (csr.get_name())
           "alert_test": begin
           end
+          "late_debug_enable": begin
+          end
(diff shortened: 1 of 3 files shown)


rswarbrick:
Force-push simplifies things. Rather stupidly, I was reimplementing some of SystemVerilog's constrained randomisation functionality. Don't do that!

eshapira (approved):
LGTM, thanks.

rswarbrick:
The CI failure is FPGA tests, which definitely won't be touched by this PR. Merging.

PR #23793, opened 2024-06-25
[rv_dm,dv] Allow pinmux_hw_debug_en_i to be more independent

rswarbrick:
This builds upon #23792 but also splits out `pinmux_hw_debug_en_i` configuration. Only the final commit is unique to this PR.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -21,6 +21,10 @@ class rv_dm_base_vseq extends cip_base_vseq #(
   rand bit pin_late_debug_enable;
   rand bit reg_late_debug_enable;
 
+  // This flag controls whether the pinmux_hw_debug_en_i signal is set to On. This determines
+  // whether the JTAG interface is connected.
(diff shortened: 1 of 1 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
```diff
@@ -333,4 +347,15 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     csr_wr(.ptr(ral.late_debug_enable), .value(mubi_val));
   endtask
 
+  // Set the pinmux_hw_debug_en_i pin to the On/Off value matching bool_val.
```
hcallahan-lowrisc, line 350: I think this comment might be inaccurate, as the implementation appears to be something like "Set the en_i to lc_ctrl_pkg::On if bool_val == 1, else set it to a random, possibly invalid mubi value?"

hcallahan-lowrisc (approved):
LGTM, with one comment nit.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
rswarbrick: Good point! Fixed.

PR #23798, opened 2024-06-25
[dv,usbdev] Low speed traffic sequence

alees24:
This PR addresses the 'low_speed_traffic' test point by introducing the ability to generate low speed traffic on the USB.

Low speed traffic is propagated in the downstream direction by hubs when operating in a full/low-speed environment, so the DUT must be robust against the presence of such traffic. USBDEV has already been modified to prevent the reporting of spurious 'bit stuffing errors' when subjected to low speed traffic.

This PR also reduces the duration of the stress sequences because they take a long time to complete in the presence of other randomized bus events/stimuli.

For context the pertinent section of the USB 2.0 Protocol Specification is `8.6.5 Low-speed Transactions.`

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_driver.sv
+++ b/hw/dv/sv/usb20_agent/usb20_driver.sv
@@ -101,7 +101,9 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     pkt.endpoint = {<<{pkt.endpoint}};
     pkt.crc5 = {<<{pkt.crc5}};
     drive_packet("Token", pkt);
-    if (req_item.m_pid_type == PidTypeInToken) begin
+    // Attempt to collect a response only to full speed IN transactions; this should be DATA0|1
+    // or ACK/NAK.
+    if (req_item.m_pid_type == PidTypeInToken && !req_item.low_speed) begin
       device_response(rsp_item);
       seq_item_port.item_done(rsp_item);
       `uvm_info (`gfn, $sformatf("In drive afer In packet : \n %0s", rsp_item.sprint()), UVM_DEBUG)
@@ -121,7 +123,8 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     pkt.data = {<<{pkt.data}};
     pkt.crc16 = {<<{pkt.crc16}};
     drive_packet("Data", pkt);
-    if (req_item.m_usb_transfer == IsoTrans) begin
+    // Isochronous OUT transactions and low speed transactions do not yield a response.
+    if (req_item.m_usb_transfer == IsoTrans || req_item.low_speed) begin
       seq_item_port.item_done();
     end else begin
       device_response(rsp_item);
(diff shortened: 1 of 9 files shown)


hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -163,19 +168,23 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
       cfg.bif.drive_p = (sym == USB20Sym_J || sym == USB20Sym_Invalid);
       cfg.bif.drive_n = (sym == USB20Sym_K || sym == USB20Sym_Invalid);
     end
-    @(posedge cfg.bif.clk_i);
-    @(posedge cfg.bif.clk_i);
-    @(posedge cfg.bif.clk_i);
+    while (delay_clks > 0) begin
+      @(posedge cfg.bif.clk_i);
+      delay_clks--;
+    end
```
rswarbrick, line 174: No big deal, but can't this just be `repeat (delay_clks) @(posedge cfg.bif.clk_i);` ?

Similarly for `idle_bit_interval`

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -209,29 +221,69 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+    // Decide the duration of the hub setup interval.
+    int unsigned idle_bits = $urandom_range(3, 6);
+    bit [7:0] pre = PidTypePre;
+    bit driver_pkt[] = new [8];
+    for (int unsigned b = 0; b < 8; b++) driver_pkt[b] = pre[b];
+    drive_packet_body("PRE", valid_sync, driver_pkt);
+    // The hub setup interval (8.6.5) must consist of at least 4 full speed bit intervals of Idle
+    // signaling, with the first bit interval being actively driven.
+    drive_bit_interval(USB20Sym_J);
+    for (int unsigned i = 0; i < idle_bits; i++) idle_bit_interval();
```
rswarbrick, line 248: Nit: I think this probably would be easier with `repeat (idle_bits) idle_bit_interval();`

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -209,29 +221,69 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
+  task drive_packet(string pkt_type, usb20_item item);
+    bit valid_sync = item.valid_sync;
+    bit driver_pkt[];
+    if (item.low_speed) begin
+      // If instructed to generate an invalid SYNC, decide whether to corrupt the SYNC on the PRE
+      // PID or on the low speed packet itself.
+      bit valid_pre_sync = $urandom & 1;
+      // Low speed traffic is prefixed with a PREamble Packet IDentifier at full speed.
+      drive_pre(valid_sync | valid_pre_sync);
+      if (!valid_sync) valid_sync = !valid_pre_sync;
```
rswarbrick, line 261: I wonder whether it makes sense to change the names here slightly. At the moment, we can have `item.valid_sync = 1` and `valid_pre_sync = 0`. This won't do anything surprising because of the OR on line 260, but I think the code looks a bit confusing.

hw/dv/sv/usb20_agent/usb20_driver.sv
alees24: Fair enough. There are a number of such cases throughout the driver code now, and although I tend to forget that 'repeat' exists, I've now been through and replaced them.

hw/dv/sv/usb20_agent/usb20_driver.sv
alees24: I've slightly twiddled the logic to make it hopefully a bit clearer.

rswarbrick (approved):
Thanks for the tweaks. This looks great!

PR #23803, opened 2024-06-25
[rv_dm,dv] Fix port name typo in rv_dm_enable_checker

a-will:
The assertion referenced a non-existent port and was missing the direction suffix. Fix that up.

Fixes a typo from #23788


vogelpi (approved):
Thanks @a-will for root causing the CI failure and fixing this!

a-will:
The broken simulations are fixed, and no other tests should be affected. Merging to fix CI.

rswarbrick:
Thank you so much for sorting this out. I'm astonished that I managed to break things like this. The first I knew of your fix was that I couldn't rebase an equivalent one at my end :-D

Sorry for the hassle.

PR #23804, opened 2024-06-25
[rom_e2e,dv] fix ROM E2E DV test cases due to new ROM banner

timothytrippel:
A new ROM UART banner was added to print chip version information to the UART. This broke some DV E2E test cases that did not expect such information to be printed.

Note: this depends on #23802.

### DIFF
--- a/hw/top_earlgrey/dv/env/chip_common_pkg.sv
+++ b/hw/top_earlgrey/dv/env/chip_common_pkg.sv
@@ -24,6 +24,9 @@ package chip_common_pkg;
 
   parameter uint ROM_CONSOLE_UART = 0;
 
+  // ROM banner values.
+  parameter string ROM_BANNER   = "OpenTitan:4001-0002-01";
+
   // ROM Boot Fault Values, matches definitions in `rules/const.bzl`.
   parameter string ROM_BFV_BAD_IDENTIFIER       = "0142500d";
   parameter string ROM_BFV_BAD_ECDSA_SIGNATURE  = "07535603";
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_sw_rom_e2e_shutdown_exception_c_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_sw_rom_e2e_shutdown_exception_c_vseq.sv
@@ -9,7 +9,8 @@ class chip_sw_rom_e2e_shutdown_exception_c_vseq extends chip_sw_rom_e2e_base_vse
   virtual task body();
     super.body();
     connect_rom_uart_agent();
-    check_uart_output_msg($sformatf("BFV:%0s\x0d\nLCV:%0s\x0d\n",
+    check_uart_output_msg($sformatf("%0s\x0d\n", ROM_BANNER));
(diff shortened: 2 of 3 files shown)


hw/top_earlgrey/dv/env/seq_lib/chip_sw_rom_e2e_shutdown_exception_c_vseq.sv
```diff
@@ -9,7 +9,8 @@ class chip_sw_rom_e2e_shutdown_exception_c_vseq extends chip_sw_rom_e2e_base_vse
   virtual task body();
     super.body();
     connect_rom_uart_agent();
-    check_uart_output_msg($sformatf("BFV:%0s\x0d\nLCV:%0s\x0d\n",
+    check_uart_output_msg($sformatf("%0s\x0d\n", ROM_BANNER));
```
a-will, line 12: Will this need some dependence on the OTP value? Or do we always run these tests with the banner enabled?

timothytrippel: It is dependent on an OTP value, but these tests always use [OTP image](https://cs.opensource.google/opentitan/opentitan/+/master:hw/top_earlgrey/dv/chip_rom_tests.hjson;drc=a3db1863f6ae2e651d890801c9ef3509024736d2;l=30) that have it unprovisioned, and the [field needs to be kHardenedBoolFalse to disable, any other value (including unprovisioned) enables it](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/rom/rom.c;drc=be1c4a4f52396d6cdedb468e06760f585cd260b0;l=105).

moidx (approved):
LGTM. Thanks!

Is there a way we can reproduce this check in an FPGA based test so that we can cross-reference this in a CI test?

hw/top_earlgrey/dv/env/chip_common_pkg.sv
```diff
@@ -24,6 +24,9 @@ package chip_common_pkg;
 
   parameter uint ROM_CONSOLE_UART = 0;
 
+  // ROM banner values.
+  parameter string ROM_BANNER   = "OpenTitan:4001-0002-01";
```
cfrantz, line 28: This value will be (should be) different when run in the silicon partner's private backend repo.

cfrantz (approved):

timothytrippel:
> LGTM. Thanks!
> 
> Is there a way we can reproduce this check in an FPGA based test so that we can cross-reference this in a CI test?

sure thing, tracking this in #23850

hw/top_earlgrey/dv/env/chip_common_pkg.sv
timothytrippel: I think the current numbers match what is expected for the PROD tapeout no?
See: https://docs.google.com/document/d/1IuxBKj0jxMpnXUadO9lHBjj_HsdWnKlgnddyx_b7BEQ/edit#heading=h.j2nphfg0gb38

PR #23805, opened 2024-06-25
[dv,usbdev] Setup priority over STALL response vseq

alees24:
Ensure that transmission of SETUP to a STALLed endpoint clears the STALL condition. This is important in clearing protocol stall and permitting Control Transfers to the endpoint to continue.

This relates to issue #23806 and RTL changes in #23807 and #23812.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -417,7 +417,7 @@ endtask
           send_handshake(PidTypeAck);
         default: begin
           // We leave the caller to deal appropriately with any other response;
-          // could be a time out (invalid device/endpoint) or could be a NAK.
+          // could be a time out (invalid device/endpoint), a NAK or a STALL, for example.
         end
       endcase
     end
@@ -854,6 +854,22 @@ endtask
     ral.configin[ep].size.set(num_of_bytes);
     ral.configin[ep].buffer.set(buffer_id);
     csr_update(ral.configin[ep]);
+  endtask
+
+  // Enable/disable the STALL response to OUT transactions for a given endpoint.
+  virtual task configure_out_stall(bit [3:0] ep, bit out_stall = 1'b1);
+    uvm_reg_data_t stall;
+    csr_rd(.ptr(ral.out_stall[0]), .value(stall));
+    stall[ep] = out_stall;
+    csr_wr(.ptr(ral.out_stall[0]), .value(stall));
+  endtask
+
+  // Enable/disable the STALL response to IN transactions for a given endpoint.
+  virtual task configure_in_stall(bit [3:0] ep, bit in_stall = 1'b1);
+    uvm_reg_data_t stall;
+    csr_rd(.ptr(ral.in_stall[0]), .value(stall));
+    stall[ep] = in_stall;
+    csr_wr(.ptr(ral.in_stall[0]), .value(stall));
   endtask
 
   // Send 'Start Of Frame' packet (bus frame/timing referenace).
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
@@ -1,0 +1,93 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class usbdev_setup_priority_over_stall_response_vseq extends usbdev_base_vseq;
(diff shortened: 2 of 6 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -845,6 +845,24 @@ endtask
     csr_update(ral.configin[ep]);
   endtask
 
+  // Enable/disable the STALL response to OUT transactions for a given endpoint.
+  virtual task configure_out_stall(bit [3:0] ep, bit out_stall = 1'b1);
+    uvm_reg_data_t ep_mask = 1 << ep;
+    uvm_reg_data_t stall;
+    csr_rd(.ptr(ral.out_stall[0]), .value(stall));
+    stall = (stall & ~ep_mask) | (out_stall ? ep_mask : 0);
```
rswarbrick, line 853: This looks reasonable, but I wonder whether indexing would make it a bit more obvious. I think that 
something like this should work (and doesn't need the mask at all)
```systemverilog
stall[ep] = out_stall;
```

Similarly in `configure_in_stall` below.

hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
```diff
@@ -0,0 +1,89 @@
...
+  endtask
+
+  virtual task body();
+    // Properties of OUT transaction.
+    int unsigned num_of_bytes = 0;
+    bit randomize_length = 1'b1;
+    // Stimulus after OUT transaction.
+    bit bus_reset = 1'b0;
+
+    // Related to issue #23806 we need to be sure that we generate long packets that yield a
```
rswarbrick, line 27: Does this still apply now that #23812 is merged? (I think I'm echoing Andreas's question from the issue)

hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
```diff
@@ -0,0 +1,89 @@
...
+    // Note: this also supplies a buffer to the SETUP DATA packet.
+    configure_setup_trans(ep_default);
+
+    // Configure an endpoint to generate a STALL handshake in response to IN or OUT transactions.
+    configure_out_stall(ep_default);
+    configure_in_stall(ep_default);
+
+    check_stall_state(ep_default, .exp_out_stall(1'b1), .exp_in_stall(1'b1));
+
+    // Check that neither IN nor OUT transactions will be accepted.
```
rswarbrick, line 65: Maybe it makes sense to expand on this comment a bit? (It took me a few seconds to see where an IN packet was being sent)

hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
```diff
@@ -0,0 +1,89 @@
...
+    send_prnd_out_packet(ep_default, PidTypeData0, randomize_length, num_of_bytes);
+    check_response_matches(PidTypeStall);
+    // Issue a Bus Reset at this point?
```
rswarbrick, line 68: Maybe it's worth saying something like "Issue a bus reset instead of sending an IN packet?" (I think the point is that we no longer expect a stall after the reset?)

hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
alees24: The test remains relevant, although as noted in the ensuing comment it is alas presently impossible to induce the exact stimulus that would have exposed the bug fixed in #23812. I've closed that issue now, thanks.

hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
alees24: Done.

hw/ip/usbdev/dv/env/seq_lib/usbdev_setup_priority_over_stall_response_vseq.sv
alees24: We never got a stall after the reset actually; the point was that the logic got itself stuck and was unable to receive the next OUT/SETUP packet under two conditions:
- if STALLing was unable and OUT packet of length >= 62 bytes was transmitted.
- if OUT packet of length >= 62 bytes was transmitted and a link reset/VBUS disconnection occurred after that many bytes had been transmitted but before a handshake response was sent by the DUT. We'd have had to use rstmgr to recover reliably from this condition.

rswarbrick (approved):
Thanks for the careful answers. This looks very sensible to me.

alees24:
The present CI failures are unrelated; this is block level DV only.

PR #23807, opened 2024-06-25
[usbdev] Also reject OUT packet internally when STALLing

### LINKED ISSUE #23806: [usbdev] OUT side STALL response can leave device unable to capture next SETUP/OUT packet
### Description

An OUT packet that elicits a 'STALL' response is neither acknowledged nor rolled back; in `usb_fs_nb_out_pe` neither 'out_ep_acked_o' nor  'rollback_data' (= `out_ep_rollback_o`) is asserted.

This means that if the packet is 62 or more bytes in length, the byte counting logic in `usbdev_usbif` is not reset in preparation for the next OUT/SETUP packet to be received and the packet size recorded into the RX FIFO is incorrect; the true size of the packet is unknowable.

The STALLing mechanism is used within current test software to declare 'protocol stall' in response to the DEVICE_QUALIFIER requests, but this does not break anything because only the IN side receives a STALL response for that Control Transfer, I think.

To reproduce, fetch draft PR sequence #23805 and run with the following seed:
```
util/dvsim/dvsim.py hw/ip/usbdev/dv/usbdev_sim_cfg.hjson -i usbdev_setup_priority_over_stall_response --fi 71923730856705723741826963384972440023047469689158980899448095559948831305837 --wa shm --tool xcelium -v m
```
If that somehow passes, try specifying '.randomize_length(0), .num_of_bytes(64)' to the task enable 'send_prnd_out_packet(..)'

This issues needs further consideration and clarification of what else may be impacted.
To be clear, this is with ES and prod RTL; it does not relate to the draft proposed changes.

@a-will, @vogelpi, @andreaskurth

a-will:
The OUT packet engine was missing a signal to internal blocks about what happened to the just-received packet if it sent a STALL to the host. Fix this so the internal blocks correctly reset counters and free up resources given to the rejected packet.

Fixes #23806

### DIFF
--- a/hw/ip/usbdev/rtl/usb_fs_nb_out_pe.sv
+++ b/hw/ip/usbdev/rtl/usb_fs_nb_out_pe.sv
@@ -288,6 +288,7 @@ module usb_fs_nb_out_pe #(
           // Non-isochronous OUT transactions end here
           if (out_ep_stall_i[out_ep_index]) begin
             tx_pid_o = {UsbPidStall}; // STALL
+            rollback_data = 1'b1; // Packet not accepted
           end else if (nak_out_transaction | out_ep_full_i[out_ep_index]) begin
             tx_pid_o = {UsbPidNak}; // NAK -- the endpoint could not accept the data at the moment
             rollback_data = 1'b1;


a-will:
CHANGE AUTHORIZED: hw/ip/usbdev/rtl/usb_fs_nb_out_pe.sv

andreaskurth:
CHANGE AUTHORIZED: hw/ip/usbdev/rtl/usb_fs_nb_out_pe.sv

andreaskurth (approved):
Approving based on the description in #23806, where rollback is mentioned as option for this case

Issue #23810, opened 2024-06-26
[test-triage] HMAC rand seq with test vector tests


martin-velay:
### Hierarchy of regression failure

Block level

### Failure Description

UVM_FATAL (hmac_test_vectors_sha_vseq.sv:28) virtual_sequencer [hmac_test_vectors_hmac_vseq] This test sequence requires a command-line argument: sha2_digest_size=SHA2_*/SHA2_*/SHA2_*

@antmarzam thanks for the warning

### Steps to Reproduce

This issue is seen each time the random sequence will fall on `hmac_test_vectors_sha_vseq` as this test requires an extra argument to work. I'll create a PR to fix this DV issue.

PR #23811, opened 2024-06-26
[hmac,dv] SHA test vec uses arg/rand digest value

martin-velay:
- add possibility to use a randomized value of the digest_size in case the command line argument is not passed while calling the test seq. This is necessary as the stress seq cannot pass such argument.

Note: this PR is linked to this issue https://github.com/lowRISC/opentitan/issues/23810

### DIFF
--- a/hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
+++ b/hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
@@ -24,9 +24,14 @@ class hmac_test_vectors_sha_vseq extends hmac_base_vseq;
     do_hmac_init              = 1'b0;
     // grab SHA-2 digest size from the command-line argument
     void'($value$plusargs("sha2_digest_size=%s", digest_size_arg));
+    // When the command line argument is not defined then randomize the digest_size with valid data
+    // This is a safety but also necessary for the stress tests as extra argument cannot be passed
     if (digest_size_arg == "") begin
-      `uvm_fatal(`gfn, {"This test sequence requires a command-line argument: ",
-                        "sha2_digest_size=SHA2_256/SHA2_384/SHA2_512"})
+      `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(
+        digest_size,
+        digest_size inside {SHA2_256, SHA2_384, SHA2_512};
+      )
+      digest_size_arg = get_digest_size(digest_size);
     end
     super.pre_start();
   endtask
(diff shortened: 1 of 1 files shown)


hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -24,9 +24,14 @@ class hmac_test_vectors_sha_vseq extends hmac_base_vseq;
     do_hmac_init              = 1'b0;
     // grab SHA-2 digest size from the command-line argument
     void'($value$plusargs("sha2_digest_size=%s", digest_size_arg));
+    // When the command line argument is not defined then randomize the digest_size with valid data
+    // This is a security but also necessary for the stress tests as extra argument cannot be passed
```
gdessouky, line 28: nit: probably reword "a security" maybe to "safety net"? 
"security" can get ambiguous here :)

gdessouky (approved):
thanks @martin-velay this LGTM and would resolve the issue for `stress_all`

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
martin-velay: Ah true!

rswarbrick (commented):
This looks sensible to me but things still feel a bit wobbly afterwards. What happens if I make a mistake and invoke this with `sha2_digest_size=cabbage`?

Maybe it would make sense to use a case statement (with an option for "") and populate an enum value. Maybe just make the `digest_size_arg` variable the same type as `digest_size`, so it gives "the digest size we wanted (irrespective of hmac_en)" ?

martin-velay:
> This looks sensible to me but things still feel a bit wobbly afterwards. What happens if I make a mistake and invoke this with `sha2_digest_size=cabbage`?
> 
> Maybe it would make sense to use a case statement (with an option for "") and populate an enum value. Maybe just make the `digest_size_arg` variable the same type as `digest_size`, so it gives "the digest size we wanted (irrespective of hmac_en)" ?

@rswarbrick 
The `hmac_en` part will be changed here https://github.com/lowRISC/opentitan/pull/23771/files, it was temporary the time this PR will come. Nevertheless, your comment is valid if `sha2_digest_size=cabbage` , but in that case I would prefer to add an `else` condition in the `body` task to flag an error if that's the case.
Would it be acceptable for you?

martin-velay:
I ran a restricted regression with the concerned tests seems to be OK. I also have tested to provide a bad command-line arg.

This PR is ready to be merged from my POV once you are OK with my last changes @rswarbrick

antmarzam (approved):
LGTM!

martin-velay:
It appeared that my previous regression was not enough, sorry for the noise. I'll run a full regression over the night but hopefully this will be OK now.

rswarbrick (approved):
This makes sense, and thanks for explaining carefully above.

rswarbrick:
CI failure is an FPGA test, so definitely not caused by this vseq change.

PR #23812, opened 2024-06-26
[usbdev] Link reset must clear packet reception state.

alees24:
In the event that a link reset occurs at the end of receiving an OUT packet of 62 or more bytes, the same fault as observed with the OUT STALL behavior would occur, causing reception failure of the first packet post-reset.

Related to issue #23806 and the RTL change in #23807, but this additional change is from inspection/reasoning rather than (yet) being demonstrated in DV.


andreaskurth (approved):

andreaskurth:
Merging due to time criticality

PR #23813, opened 2024-06-26
[hmac,dv] SCB digest_idx issue in SHA2_384

martin-velay:
- index has to include 11 in mode SHA2_384


gdessouky (approved):
thanks for catching this!

rswarbrick (approved):

rswarbrick:
CI failure is an FPGA test which is definitely unaffected by this PR.

PR #23816, opened 2024-06-26
[dv,usbdev] DV tidying for USBDEV

alees24:
Now that the set of tests and sequences is starting to stabilize, this PR just tidies the handling of responses and sorts file lists, which was hard to achieve when the lists were in a state of continual flux.
No functional changes; test pass rate at 97.62% for '-i nightly'


rswarbrick (approved):
This looks good to me!

Issue #23245, opened 2024-05-22
[hmac, dv] Test `key_swap` functionality


ballifatih:
### Description

Test that `key_swap` correctly flips the endianness of KEY registers, as SW is loading key values. Related to https://github.com/lowRISC/opentitan/issues/23158

I will tag this as FutureRelease, as driver implementation should provide acceptable initial coverage.

PR #23818, opened 2024-06-26
[hmac,dv] Key swap register test

martin-velay:
This PR is related to this issue https://github.com/lowRISC/opentitan/issues/23245 and is linked with these RTL changes https://github.com/lowRISC/opentitan/issues/23158.

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_scoreboard.sv
+++ b/hw/ip/hmac/dv/env/hmac_scoreboard.sv
@@ -669,7 +669,19 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
     bit [3:0] digest_size = `gmv(ral.cfg.digest_size),
     bit [5:0] key_length  = `gmv(ral.cfg.key_length));
     bit [7:0] msg_tmp[];
+    bit [TL_DW-1:0] big_endian_key[NUM_KEYS];
     exp_digest = '{default:0}; // clear previous expected digest
+
+    // Swap the key when required according to the dedicated register field CFG.key_swap
+    if (`gmv(ral.cfg.key_swap)) begin
+      `uvm_info(`gfn, "Swap the key from little-endian to big-endian", UVM_HIGH)
+      foreach (big_endian_key[key_i]) begin
+        big_endian_key[key_i] = {<<8{key[key_i]}};
+      end
+    end else begin
+      `uvm_info(`gfn, "Keep the key in big-endian", UVM_HIGH)
+      big_endian_key = key;
+    end
 
     `uvm_info(`gfn, $sformatf("Computing digest prediction"), UVM_LOW)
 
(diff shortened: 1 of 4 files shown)


hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -669,50 +669,63 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
     bit [3:0] digest_size = `gmv(ral.cfg.digest_size),
     bit [5:0] key_length  = `gmv(ral.cfg.key_length));
     bit [7:0] msg_tmp[];
+    bit [TL_DW-1:0] key_swapped[NUM_KEYS];  // Keep big-endian when key_swap register is set to 0
```
rswarbrick, line 672: I'm sorry, but I don't completely understand this variable. Is the point that this is the big-endian representation of all the keys (which might be little-endian in the registers if `key_swap` is true)?

If so, I suggest:
- Rename the variable to something like `big_endian_keys`
- Drop the comment on this line
- Change the `uvm_info` line just below to "Swap the key from little-endian"

Or have I misunderstood something?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -669,50 +669,63 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
           if (key_length == Key_512) cryptoc_dpi_pkg::sv_dpi_get_hmac_sha512(
-                                                      key[0:15], msg_i, exp_digest);
+                                                      key_swapped[0:15], msg_i, exp_digest);
           if (key_length == Key_1024) cryptoc_dpi_pkg::sv_dpi_get_hmac_sha512(
-                                                      key[0:31], msg_i, exp_digest);
+                                                      key_swapped[0:31], msg_i, exp_digest);
         end
-        `uvm_info(`gfn, $sformatf("HMAC of key=%p, msg_i=%p: %p", key, msg_i, exp_digest), UVM_LOW)
+        `uvm_info(`gfn, $sformatf("HMAC of key=%p (key_swap=%1b), msg_i=%p: %p",
+                  key, `gmv(ral.cfg.key_swap), msg_i, exp_digest), UVM_LOW)
```
rswarbrick, line 728: I really think that it would be better if the indentation lined up here. `key` is the first argument to `$sformatf`, not `uvm_info`.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
rswarbrick: Ok, my understanding (which I think agrees with what you are saying):
- Keys might be provided in the registers in little-endian or big-endian format.
- Software tells the hardware which format is being used by setting a `key_swap` register to 1.
- The scoreboard needs to keep track of "the current key". It does so by snooping on writes to `key_0`,...`key_31` and updates its local mirror, stored in the `key` array.
- When predicting a digest, we look up the contents of each `key_*` register and swap it around as needed if `key_swap` is true.

This all looks sensible. I think I stand by my list of suggestions from yesterday. Here's an expanded version which should hopefully make a bit more sense:

- At the moment, `key_swapped` might be the same as the register value or it might be swapped around. I'd say this makes the name rather confusing! To make it less confusing, I suggest changing the name to something that's always true, like the `big_endian_keys` name I suggested. (I'd also suggest the plural, given that the array contains more than one key)
- The comment next to the variable is a little confusing. I think the  variable is always big-endian, regardless of the state of `key_swap`. With the current variable name, I agree it's meaning needs some explanation. With my proposed name change, we can probably drop the comment entirely.
- The `uvm_info` line is grammatically wrong. We're swapping the key **from** little-endian **to** big-endian and the message currently says the opposite. I suggest fixing it.

(Follow-up minor question, which isn't needed for this PR: Why do we have the `key` array at all? Wouldn't we expect its contents to be accessible with `gmv` as well?)

rswarbrick: Follow-up: Duh. It's a single key, which is more than 32 bits long so takes several registers. Ignore all my mentions of "key**s**" !

martin-velay: Oh OK I see, thanks for the detailed explanations, let me make the suggested changes.

The `key` is secret, otherwise the message can be decoded. So its access is `wo` and that's why we get the `key` value while writing it. For more info you can read the reg description [here](https://opentitan.org/book/hw/ip/hmac/doc/registers.html#key).

rswarbrick (approved):
Thanks for all the tidy-ups. This looks good to me.

PR #23820, opened 2024-06-26
[rom] Make the banner message non-blocking

cfrantz:
1. The ROM banner takes about 2 ms to be sent from the UART at 115200 bps.  Rather than wait for it to complete transmission, emit the characters into the UART FIFO and proceed with boot.  Hashing the FLASH contents and performing signature verification typically will take an additional 2.4 + 4.2 ms.  There is no need to delay boot waiting for the UART.
2. Fix the ROM's terminal states to wait for UART output to complete.


cfrantz:
As discussed offline: an boot-speed optimization for UART output.

timothytrippel (approved):

moidx (approved):

cfrantz:

Issue #23797, opened 2024-06-25
[flash_ctr] Decouple read buffer arbitration logic in `flash_phy_rd.sv` from Flash primitive


vogelpi:
### Description

As shared by @eshapira and @sha-ron in an email thread, it has been noted that the read buffer arbitration logic inside `flash_phy_rd.sv` always assumes the Flash primitive to immediately acknowledge new read requests. The open source simulation model seems to behave like that whereas some real Flash macros may not always acknowledge the requests immediately.

As a result, read buffer arbitration logic can skip certain buffers and only re-allocate new entries to odd buffers. It's possible that the even buffers don't get re-allocated unless prog or erase operations to the buffered addresses happen.

The functionality of the module is still correct (FLASH_CTRL always returns the requested data) but the number of effectively used buffers can be halved in practice which may have a performance impact.

This can potentially be fixed easily by factoring in the `ack_i` signal (together with the `req_o`) into the `ready_i` of the buffer arbiter. However, a full DV regression would need to be run to make sure we don't introduce a bug with this.

We've discussed this issue in the tape-out sync meeting on 2024-06-24 and concluded to not do an ECO for Earlgrey-PROD.M5 as there is still the Ibex ICACHE which very effectively reduces the pressure on FLASH_CTRL and there is not clear evidence that fixing this would notably improve performance to justify a potential risk of this ECO on the schedule.

However, we've agreed to look into this after M5 / after taking the TO branch to fix the design.

PR #23822, opened 2024-06-27
[flash_ctrl] Fix read buffer re-allocation for even buffers

vogelpi:
Previously, the read buffer re-allocation arbiter erroneously did not factor in the `ack_i` signal coming from the Flash macro. While the open source simulation model never really de-asserts `ack_i`, some real Flash macros may do so. As a result, `req_o` might be high for multiple clock cycles and the read buffer re-allocation may skip even read buffers. This effectively halves the number of usable read buffers which may have a negative performance impact.

The fix is easy and just involves factoring in `ack_i` to take the actual Flash macro request handshake for doing allocations.

This resolves lowRISC/OpenTitan#23797.

### DIFF
--- a/hw/ip_templates/flash_ctrl/rtl/flash_phy_rd.sv
+++ b/hw/ip_templates/flash_ctrl/rtl/flash_phy_rd.sv
@@ -179,7 +179,7 @@ module flash_phy_rd
     .idx_o(),
     .valid_o(),
     .data_o(),
-    .ready_i(req_o & no_match)
+    .ready_i(req_o & ack_i & no_match)
   );
 
   // which buffer to allocate upon a new transaction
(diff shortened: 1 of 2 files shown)


vogelpi:
Note: I am NOT suggesting to integrate this as an ECO. But I would like to have a PR ready just in case another opportunity opens up for integrating it.

I am currently running a full block-level regression for Flash and will post results here. @eshapira already confirmed that this is working fine with the closed source macro.

eshapira (approved):
LGTM, thanks!

andreaskurth:
CHANGE AUTHORIZED: hw/ip_templates/flash_ctrl/rtl/flash_phy_rd.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/flash_ctrl/rtl/flash_phy_rd.sv

If there's another synthesis window; otherwise we delay this (or have to revert).

vogelpi:
CHANGE AUTHORIZED: hw/ip_templates/flash_ctrl/rtl/flash_phy_rd.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/flash_ctrl/rtl/flash_phy_rd.sv

vogelpi:
I've locally run a full regression with the open source model and the pass rates are similar to the latest nightly regressions. There are no new failure signatures.

There was one failure `chip_sw_rstmgr_sw_req` in regression 1 with Xcelium. I've ran that locally and it passed.

I am thus merging this to create the next RC.

Issue #9999, opened 2022-01-11
pwm specification should comment on the supported range of clocks


rasmus-madsen:
The PWM specification should comment on what range of clocks are supported. 
it should also comment on the relationship between TL <-> core clock.

this is needed to know what range the clocks should be used to verify the CDC

> estimate 1

martin-lueker:
(Coming back to this after the V2 review)

The CDC is now managed by the reg-tool interface generator.  That said, do we need any special documentation for this?

cindychip:
@msfschaffner @tjaychen do you have any insight on this?

rasmus-madsen:
@martin-lueker I think this issue can be closed- the responsibility of verifying the CDC is now somewhere else

andreaskurth:
Triaged for `pwm`. Assigning to M2.5 with https://github.com/lowRISC/opentitan/labels/Priority%3AP3 because as per the last comment:
- "There's no clock frequency requirement anymore that we know of." So documentation needs to be updated but that could happen post M2.5.
- Block-level DV should check this: The clocks are being randomized, and we collect coverage for it in `clock_cg`, which is currently at 100%. However, the TL clocks are mapped to a single bin, so that bin and the cross with the core clock is not very meaningful. This should be improved to verify this more thoroughly.

msfschaffner:
This is low-prio and not needed for M2.5 hence deferring.

moidx:
Moving to M5 based on priority.

PR #23833, opened 2024-06-27
Pwm clk in doc and fcov fix

antmarzam:
This takes care of #9999


rswarbrick (approved):
Thanks, this makes sense and (after a bit of reading) I'm convinced that this handles the items discussed in #9999.

PR #23834, opened 2024-06-27
[rom] Improve hardened checks

cfrantz:
Change the hardened check to evaluate both the requested condition and its opposite.  Land on an illegal instruction if either of the tests evaluates incorrectly.

### DIFF
--- a/sw/device/lib/base/BUILD
+++ b/sw/device/lib/base/BUILD
@@ -233,6 +233,20 @@ cc_test(
     ],
 )
 
+opentitan_test(
+    name = "hardened_functest",
+    srcs = ["hardened_functest.c"],
+    exec_env = EARLGREY_TEST_ENVS,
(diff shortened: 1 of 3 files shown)


moidx (approved):
Should we add some unittests just as a sanity check given all the inverse operations?

cfrantz:
> Should we add some unittests just as a sanity check given all the inverse operations?

Would it surprise you to know we didn't have any tests of the hardened sequences (apart from the fact that the ROM itself executes successfully).

I've added an on-target test that checks each of the hardened sequences in both the good and bad cases.

moidx:
The functest makes me feel much better :) Thanks @cfrantz! 

LGTM

johannheyszl (approved):
Thanks @cfrantz and @aewag !

cfrantz:
The CI test failure is not related to the change in this PR.

PR #23836, opened 2024-06-27
[dv,usbdev] Frequency and phase delta sequence

alees24:
This PR adds frequency and phase delta testing in two commits:

_Commit 1:_
Adds a sequence for streaming max length packets, taking into consideration USB bit stuffing. Employing maximum length packets is critical in ensuring that the DUT continues to sample the USB signals correctly throughout the packet.

_Commit 2:_
Adds a sequence that employs the 'max length streaming' sequence to operate the DUT and host at different frequencies and phase relationships and check that all data is transferred correctly as expected, without any reliance upon retrying.
- Bus framing is performed by the host, ensuring that the SOF packets form a reliable heart beat signal every 1ms, according to the host clock frequency.
- Traffic may be sent to the DUT during each bus frame, away from the SOF packets, and optionally only once the 'reset recovery' interval has elapsed.
- The 'timing reference' signal from the DUT, in response to detected SOF packets from the host, may be used to adjust the oscillator frequency of the DUT.
- Host frequency may configured to drift during the test, to make the host clock/SOF packets a moving target, rather than tracking and converging only at the start of the sequence.
- For diagnostic/development purposes, the traffic may be disabled entirely, to ensure that the only USB traffic and receiver activity is that of the SOF packets/timing reference.

Passes with 10 seeds each for the 5 new freq/phase tests.

### DIFF
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -28,6 +28,8 @@ bit [4:0] setup_buffer_id = 5'd1;
 bit [4:0] out_buffer_id = 5'd7;
 // Current IN buffer number
 bit [4:0] in_buffer_id = 5'd13;
+// Current frame number
+bit [10:0] frame_number = 11'b0;
 
 // Chosen /default/ endpoint, for those sequences concerned with only a single endpoint.
 rand bit [3:0] ep_default;
@@ -92,6 +94,9 @@ uint aon_wake_events_rd_clks = usbdev_aon_wake_clk_ratio;
 // - some multi-stage operations must be performed atomically,
 //   eg. token packet and handshake response, or VBUS connection and Bus Reset.
 semaphore m_drv_lock;
+
+// Current period of the DUT clock, in picoseconds.
+int unsigned usb_clk_period_ps;
 
 `uvm_object_new
 
@@ -194,6 +199,25 @@ endtask
     // Create the lock controlling exclusive access to the driver.
     m_drv_lock = new(1);
 
+    // Use the configured DUT clock frequency.
+    //
+    // Retain the clock period in picoseconds for oscillator tracking calculations, rounding to
+    // nearest because we have no particular bias towards lower/higher frequency.
+    usb_clk_period_ps = (1000_000_000 + cfg.usb_clk_freq_khz / 2) / cfg.usb_clk_freq_khz;
+    cfg.clk_rst_vif.set_period_ps(usb_clk_period_ps);
+
+    // Use a sensible speed for the AON clock
+    cfg.aon_clk_rst_vif.set_freq_khz(cfg.aon_clk_freq_khz);
+
+    // Use the selected host clock frequency; this may differ from that of the DUT.
+    // The DUT is required to cope with a certain amount of disparity, and adjust itself
+    // to match the frequency of the host over time.
+    cfg.host_clk_rst_vif.set_freq_khz(cfg.host_clk_freq_khz);
+
+    // Report the clock frequencies.
+    `uvm_info(`gfn, $sformatf("usb_clk %dkHz, host_clk %dkHz",
+                              cfg.usb_clk_freq_khz, cfg.host_clk_freq_khz), UVM_MEDIUM)
+
     if (do_agent_connect) begin
       // Connect the USB20 agent and ensure that the USBDPI model is not connected.
       cfg.m_usb20_agent_cfg.bif.enable_driver(1'b1);
@@ -241,6 +265,24 @@ endtask
       `dv_fatal($sformatf("DUT did not generate any of the expected interrupt(s) 0x%x", snapshot))
     end
     if (clear) csr_wr(.ptr(ral.intr_state), .value(data & snapshot));
+  endtask
+
+  // Set the host clock to the given frequency (kHz).
+  virtual task set_host_clk(uint freq_khz);
+    `uvm_info(`gfn, $sformatf("Setting host clk to %dkHz", freq_khz), UVM_MEDIUM)
+    `DV_CHECK(freq_khz >= 47_880 && freq_khz <= 48_120);  // Sanity check, but also USB compliance.
+    cfg.host_clk_rst_vif.set_freq_khz(freq_khz);
+    cfg.host_clk_freq_khz = freq_khz;
+  endtask
+
+  // Adjust the DUT clock period by the given number of picoseconds.
+  virtual task adjust_usb_clk(int delta_ps);
+    // Ensure that the period will not become zero/negative when decreasing it.
+    `DV_CHECK(int'(usb_clk_period_ps) > -delta_ps);
+    usb_clk_period_ps += delta_ps;
+    `uvm_info(`gfn, $sformatf("Adjusting usb_clk_period by %dps to %d",
+                              delta_ps, usb_clk_period_ps), UVM_MEDIUM)
+    cfg.clk_rst_vif.set_period_ps(usb_clk_period_ps);
   endtask
 
   // Wait for the DUT to enter one of the given link states, with an optional timeout
@@ -873,16 +915,28 @@ endtask
   endtask
 
   // Send 'Start Of Frame' packet (bus frame/timing reference).
-  virtual task send_sof_packet(input pid_type_e pid_type);
+  virtual task send_sof_packet(input pid_type_e pid_type, int framenum = -1);
+    // If the caller did not request a specific frame number, then advance sequentially as per
+    // a real host; specific frame numbers should really only be specified to check awkward cases
+    // such as those which induce bit stuffing, making SOF packets longer.
+    if (framenum < 0) begin
+      // Starting a new bus frame.
+      framenum = frame_number + 1;
+      if (framenum > 11'h7ff) framenum = 0;
+    end else `DV_CHECK(framenum <= 11'h7ff, "Frame number is only 11 bits")
+    // Remember the current frame number.
+    frame_number = framenum;
     `uvm_create_on(m_sof_pkt, p_sequencer.usb20_sequencer_h)
     start_item(m_sof_pkt);
     m_sof_pkt.m_ev_type  = EvPacket;
     m_sof_pkt.m_pkt_type = PktTypeSoF;
     m_sof_pkt.m_pid_type = pid_type;
-    assert(m_sof_pkt.randomize());
+    // Randomization also completes the CRC5 value.
+    assert(m_sof_pkt.randomize() with {m_sof_pkt.framenum == framenum;});
     // Any fault injections requested?
     if (inject_invalid_token_sync) m_sof_pkt.valid_sync = 1'b0;
     if (inject_bad_token_crc5) m_sof_pkt.crc5 = ~m_sof_pkt.crc5;
+    `uvm_info(`gfn, $sformatf("Sending SOF packet with frame number 0x%03x", framenum), UVM_MEDIUM)
     finish_item(m_sof_pkt);
   endtask
 
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -872,7 +872,7 @@ endtask
     csr_wr(.ptr(ral.in_stall[0]), .value(stall));
   endtask
 
-  // Send 'Start Of Frame' packet (bus frame/timing referenace).
+  // Send 'Start Of Frame' packet (bus frame/timing reference).
   virtual task send_sof_packet(input pid_type_e pid_type);
     `uvm_create_on(m_sof_pkt, p_sequencer.usb20_sequencer_h)
     start_item(m_sof_pkt);
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
@@ -1,0 +1,278 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Frequency and phase delta between host and device.
+//
+// - Host and DUT frequency may be set independently, to opposite extremes of the permitted
+//   frequency range if desired.
+// - DUT frequency may be adjusted in response to the DUT timing reference signal to match the
+//   frequency of the host.
+// - Host frequency may be configured to drift throughout the test.
+// - Sequence may execute with or without traffic (max length packets to increase the likelihood
+//   of sampling errors from frequency/phase mismatch).
+// - There is a 10ms 'Reset Recovery' period guaranteed within the USB 2.0 protocol specification,
+//   during which the host controller is not permitted to send anything other than SOF packets
+//   to the DUT; this is important in guaranteeing that the DUT has time to measure and track the
+//   frequency of the host.
+class usbdev_freq_phase_delta_vseq extends usbdev_stream_len_max_vseq;
+  `uvm_object_utils(usbdev_freq_phase_delta_vseq)
+
+  `uvm_object_new
+
+  // Min/max DUT clock frequencies (scaled from 12Mbps to 48MHz clock) according to the
+  // USB 2.0 specification; Table 7-9.
+  localparam uint USB_CLK_FREQ_MIN_KHZ = 4 * 11_970;
+  localparam uint USB_CLK_FREQ_MAX_KHZ = 4 * 12_030;
+
+  // Target number of clock cycles per bus frame.
+  localparam int CYCLES_PER_FRAME = 48_000;
+
+  // Duration of test in 1ms bus frames.
+  rand uint num_frames;
+  constraint num_frames_c {
+    num_frames inside {[80:120]};
+  };
+
+  // Target clock frequency of host.
+  rand uint host_target_freq_khz;
+  constraint host_target_freq_khz_c {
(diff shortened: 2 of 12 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -243,6 +267,24 @@ endtask
...
+    `uvm_info(`gfn, $sformatf("Setting host clk to %dkHz", freq_khz), UVM_MEDIUM)
+    `DV_CHECK(freq_khz >= 47_880 && freq_khz <= 48_120);  // Sanity check, but also USB compliance.
+    cfg.host_clk_rst_vif.set_freq_khz(freq_khz);
+    cfg.host_clk_freq_khz = freq_khz;
+  endtask
+
+  // Adjust the DUT clock period by the given number of picoseconds.
+  virtual task adjust_usb_clk(int delta_ps);
+    // Ensure that the period will not become zero/negative.
+    `DV_CHECK(int'(usb_clk_period_ps) > -delta_ps);
```
rswarbrick, line 281: I'm convinced this is correct, but it took me a few seconds to convince myself! Would `int'(usb_clk_period_ps) + delta_ps > 0` work instead? I think it might be easier to read.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -857,16 +899,28 @@ endtask
   endtask
 
   // Send 'Start Of Frame' packet (bus frame/timing referenace).
```
rswarbrick, line 901: I know you're not changing this line but you're right next door... :-) Would you mind fixing the spelling of reference while you're here?

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -857,16 +899,28 @@ endtask
...
-  virtual task send_sof_packet(input pid_type_e pid_type);
+  virtual task send_sof_packet(input pid_type_e pid_type, int framenum = -1);
+    // If the caller did not request a specific frame number, then advance sequentially as per
+    // a real host; specific frame numbers should really only be specified to check awkward cases
+    // such as though which induce bit stuffing, making SOF packets longer.
```
rswarbrick, line 905: Do you mean "those which"?

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -857,16 +899,28 @@ endtask
...
+    if (framenum < 0) begin
+      // Starting a new bus frame.
+      framenum = frame_number + 1;
+    end
+    if (framenum > 11'h7ff) framenum = 0;
```
rswarbrick, line 910: Would it make sense to put this inside the `if` block? Maybe also stick an assertion that `-1 <= framenum < 7ff` at the top of the task?

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -857,16 +899,28 @@ endtask
...
+    // Remember the current frame number.
+    frame_number = framenum;
     `uvm_create_on(m_sof_pkt, p_sequencer.usb20_sequencer_h)
     start_item(m_sof_pkt);
     m_sof_pkt.m_ev_type  = EvPacket;
     m_sof_pkt.m_pkt_type = PktTypeSoF;
     m_sof_pkt.m_pid_type = pid_type;
-    assert(m_sof_pkt.randomize());
+    // Randomization also completes the CRC5 value.
+    assert(m_sof_pkt.randomize() with {m_sof_pkt.framenum == framenum;});
```
rswarbrick, line 919: Do we ever want to randomize `framenum`? If not, we could drop the "rand" in the item definition and could set `framenum` explicitly before the randomisation here, which might be a bit easier to understand.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+  rand uint num_frames;
+  constraint num_frames_c {
+    num_frames inside {[80:120]};
+  };
+
+  // Target clock frequency of host.
+  rand uint host_target_freq_khz;
+  constraint host_target_freq_khz_c {
+    // Maximum frequency extremes (scaled from 12Mbps to 48MHz clock). See USB 2.0 Table 7-9
+    host_target_freq_khz >= 47_880 && host_target_freq_khz <= 48_120;
```
rswarbrick, line 40: Would it be simpler to make these bounds `4*USB_CLK_FREQ_MIN_KHZ` and `4*USB_CLK_FREQ_MAX_KHZ`? It would avoid the extra literals.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+  };
+
+  // Configured initial frequency deltas from 48Mhz, given in Hz.
+  int host_freq_delta;
+  int usb_freq_delta;
+
+  // Are we require to track the host frequency by adjusting the DUT oscillator?
```
rswarbrick, line 47: Grammar nit: Do you mean "required"?

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+
+  // Shall the host clock frequency drift over the duration of the test?
+  bit host_drifting = 0;
+
+  // Shall there be any bus traffic during this test?
+  bit with_traffic = 1;
+
+  // Observe Reset Recovery period; do not transmit anything other than SOF packets within 10ms
+  // of Reset Signaling?
+  bit reset_recovery = 1;
```
rswarbrick, line 58: Do we ever want to test `reset_recovery=0`? If I understand correctly, usbdev isn't required to deal with that situation? (So maybe we can drop the plusarg)

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+
+  // Test only SOF reception, not maximum length packets; diagnostic/development switch can make
+  // it easier to study the frequency drifting/tracking.
+  bit sof_only = 0;
+
+  // Indicate that the test sequence is terminating.
```
rswarbrick, line 64: Nit: I'd suggest "Indicates". Just because I thought this was the same as the items just before, so initially read thsi as a flag that controlled whether we should indicate when we were terminating...

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+
+      // The mean delta in picoseconds must be amortized over all of the bits in the bus frame.
+      adjust_usb_clk(mean_delta_ps / CYCLES_PER_FRAME);
+    end
+  endtask
+
+  // Transmission of packet data
+  virtual task bus_traffic();
+    int unsigned frames = num_frames;
+    while (frames > 0) begin
```
rswarbrick, line 208: Nit: I think this might be simpler as just `repeat (num_frames)` because we don't need to keep track of which frame we're doing.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+      // Prevent collision between DATA packets and SOF packet at frame start.
+      while (!can_transmit) begin
+        cfg.host_clk_rst_vif.wait_clks(8);
+      end
+      can_transmit = 0;
```
rswarbrick, line 213: This looks reasonable, but we're essentially building a mutex between `bus_traffic` and `bus_framing`. Does it end up easier to read with a SystemVerilog semaphore? (Unless the steps by 8 are important?)

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+  endtask
+
+  virtual task pre_start();
+    // Vary the initial DUT clock frequency in these test sequences.
+    cfg.usb_clk_freq_khz_c.constraint_mode(0);
+    if ($value$plusargs("usb_freq_delta=%0d", usb_freq_delta)) begin
+      cfg.usb_clk_freq_khz = 48_000 + (usb_freq_delta + 500) / 1000;  // round to nearest
+    end else begin
+      // The DUT frequency is kept constant by default in other vseqs, but we want to vary it for
+      // freq/phase testing.
```
rswarbrick, line 229: Nit: I'm not sure that this comment only applies to this half of the if/else. Maybe it needs moving to join up with the comment at the start of the task?

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+      if (with_traffic & !reset_recovery & !sof_only) transmit_out_packets(10);
+      frames--;
+    end
+    // Signal to the other processes that the sequence is stopping.
+    stopping = 1;
+  endtask
+
+  virtual task pre_start();
+    // Vary the initial DUT clock frequency in these test sequences.
+    cfg.usb_clk_freq_khz_c.constraint_mode(0);
```
rswarbrick, line 224: I'm not sure whether you actually need to configure the constraints here, on the basis that we're not randomising `usb_clk_freq_khz` again: doing so would undo our work! :-) I'd probably either drop this line, or maybe it's worth avoiding the instance not matching its constraints. In which case, maybe there needs to be a comment about why we're disabling the constraint?

Similarly for the `host_clk_freq_khz_c` below.

rswarbrick (commented):
I realise that I've left a lot of comments (sorry), but this is really nice. I particularly enjoyed reasoning through the logic in `usbdev_freq_phase_delta_vseq` and love that we're thinking about `sum(1/2**n, n, 0, inf)` :-)

hw/ip/usbdev/dv/env/seq_lib/usbdev_stream_len_max_vseq.sv
```diff
@@ -0,0 +1,52 @@
...
+    // maximal bit stuffing occurs. Although bit stuffing is active throughout the DATA0 PID
+    // and the CRC16 too, it seems as though the maximum packet length is 614 bits and this is
+    // achieved with a simple sequence of 0xff bytes, yielding a CRC16 of 0x40fe.
+    for (int unsigned idx = 0; idx < MaxPktSizeByte; idx++) begin
+      data.push_back(8'hFF);
+    end
+    for (int unsigned txn = 0; txn < num_packets; txn++) begin
+      // TODO: perhaps we want to collect and check the packets in another process?
+      // This would stop the CSR/buffer accesses from affecting the USB traffic timing, which
+      // could be important in exercising all freq/phase extremes.
```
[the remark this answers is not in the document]
alees24: I think the parallel process would be preferable - at some point - because it's what happens on the real system, and because in this case the read from the RX FIFO does need to modify the hardware state.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: It's a somewhat pedantic point, and unlikely in this case ever to cause problems, but I tend to avoid additions/subtractions because of the potential for overflow/underflow to lead to the wrong decision. Only a comparison is actually required here.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: Fair enough; I'll pass the burden of ensuring the validity of the frame number to the caller.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: This is legacy really, but we may at some point want to generate the SOF packets at another site. The way we create and randomize the packets is not easily OT DV-compliant in other ways at present either.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
alees24: Changed, and annotated that presently the DV is using the same sampling scheme as the DUT. It may be preferable to change that at some point, the better to cope with arbitrary phase and frequency. It doesn't need to run purely on clocked logic like the DUT, and potentially run the risk of sampling errors.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
alees24: Typo, thanks.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
alees24: Well, USBDEV doesn't really rely upon or have any knowledge of Reset Recovery and that switch was only recently introduced into the DV, to see whether it would improve things with the known prim_cdc_rand_delay issue, ie. could we decode and use the SOF packets reliably enough to converge upon the host frequency before having to handle longer packets?
You're correct that USBDEV doesn't need to handle a lack of Reset Recovery, although 10ms is also a lot of wasted simulation time when it's not required and I would prefer to keep the flexibility in DV at least until such time as the CDC issues are resolved.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
alees24: Comment reworded and identifier renamed.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
alees24: The step size is not really significant, no, but it has always been the intent to have a more accurate model of real USB host behavior and bus timing, so it's been built using variables rather than sempahores to support additional processes and/or derived sequences.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+      while (!can_transmit) begin
+        cfg.host_clk_rst_vif.wait_clks(8);
+      end
+      can_transmit = 0;
+      // Transmit enough packets to fill about 2/3 of the bus frame.
+      if (with_traffic & !reset_recovery & !sof_only) transmit_out_packets(10);
+      frames--;
+    end
+    // Signal to the other processes that the sequence is stopping.
+    stopping = 1;
```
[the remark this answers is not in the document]
alees24: Ok. Now renamed to 'all_frames_done' since we aren't always generating bus traffic.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
alees24: Clarified the comment.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
alees24: Okay, thanks.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
```diff
@@ -0,0 +1,280 @@
...
+  endtask
+
+  // Parallel task generates SOF periodically, dividing the bus activity into discrete
+  // bus frames and indicating when transmission of regular bus traffic may occur.
+  virtual task bus_framing();
+    // Whatever the actual host frequency, we want to generate a SOF every 48_000 clock
+    // cycles, since this equates to 1ms in the host's time frame.
+    int unsigned elapsed_cycles = 0;
+    while (!stopping) begin
+      cfg.host_clk_rst_vif.wait_clks(48_000 - elapsed_cycles);
```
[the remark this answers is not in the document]
alees24: I don't think it makes any real difference either way, but one could argue that it somewhat harder for DUT and/or DV to have to wait a full bus frame before the first SOF with their frequencies mismatched so I'd prefer to leave it as is. I think the USB specification and host make no guarantee about the point within the bus frame at which connection/bus reset are performed.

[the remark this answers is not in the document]
rswarbrick: Ah, ok. That makes sense, thanks.

hw/ip/usbdev/dv/env/seq_lib/usbdev_freq_phase_delta_vseq.sv
rswarbrick: Ok, that's reasonable. Thanks for explaining the reasoning.

rswarbrick (approved):
Thanks for all the tidy-ups. This looks really good to me.

PR #23841, opened 2024-06-28
[rv_dm,dv] Teach SVA and scoreboard about late debug enable

rswarbrick:
Both were using lc_hw_debug_en_i to figure out whether debug was enabled. That's correct if late debug enable is true, but is wrong if it is false. Teach them to use the right control.

The background to this change is that I'm trying to generalise the DV virtual sequences so that we sometimes don't use late debug enable. After doing so, these checks fail very quickly! This fixes the checks.


hcallahan-lowrisc (approved):
LGTM

PR #23842, opened 2024-06-28
[entropy_src/dv] Exclude hard coded counter inputs for toggle coverage

h-filali:
This commit excludes all the hard coded inputs to counter blocks from toggle coverage. This should help us to bring back the toggle coverage to an acceptable level.


h-filali:
I ran a 10% regression and with the added changes we get a toggle coverage of around 87%.
I think for a 100% regression this should be enough to get above 90%.
I'll run a full regression over night to confirm.

vogelpi (approved):
LGTM, thanks @h-filali !

h-filali:
Okay looks like we're still missing 1.5%. I'll look into it and create a follow up PR.

Issue #18350, opened 2023-05-05
[csrng/dv] Improve testing and coverage of `glen`


andreaskurth:
The `glen` field of CSRNG commands currently has only very restricted randomization:
https://github.com/lowRISC/opentitan/blob/438c2431de941e0311786254d3879832a503bd26/hw/dv/sv/csrng_agent/csrng_item.sv?plain=1#L34-L37

This is problematic as `glen` values outside that range are legal but currently don't get tested. In particular, values > 32 are very relevant. Value 0 is not as relevant but should be tested to not cause problems either. (Value 0 currently seems to work on the RTL side but fails in DV checks.)

Coverage definition is also simplistic:
https://github.com/lowRISC/opentitan/blob/438c2431de941e0311786254d3879832a503bd26/hw/ip/csrng/dv/cov/csrng_cov_if.sv?plain=1#L315-L319

While this isn't as critical as actually testing the values, this coverage definition should be refined to include 0 as well as the maximum value (checking corner-case behavior) and ideally a few bins for value ranges between (e.g., small, medium, large values).

vogelpi:
Update: what's left to do here is to test the value 0. The other work has been done as part of #18356.

vogelpi:
It turns out that the RTL doesn't handle Generate commands with a glen equal 0 (generate NO randomness) correctly. But this behavior documented already. We could correct the design to signal a error status response instead, but this really isn't a priority as a Generate command that shouldn't generate entropy is a corner case and even the NIST spec isn't fully clear what should happen here. See https://github.com/lowRISC/opentitan/issues/23846 for details.

Thus, I've filed a PR to simply extend the coverage definition for the glen coverpoint here #23847.

Issue #23846, opened 2024-06-28
[csrng] Return error status response for GEN commands with glen = 0


vogelpi:
### Description

When taking care of #18350, I found out that CSRNG treats Generate commands with a glen = 0 similarly to commands with a glen = 1, i.e., it generates one single 128-bit block of random data instead of nothing.

I don't think this is a bug really because the [CSRNG documentation](https://opentitan.org/book/hw/ip/csrng/doc/theory_of_operation.html#command-header) specifies:
> Generate Length: Only defined for the generate command, this field is the total number of cryptographic entropy blocks requested. Each unit represents 128 bits of entropy returned. This field allows values between 1 and 4095. A value of 1 returns 1 * 128 bits of entropy. A value of 4095 returns 4095 * 128 bits of entropy, which is less than the 219 bits allowed by NIST (referenced to as max_number_of_bit_per_request). 

and further:
> The glen field defines how many 128-bit words are to be returned to the application interface. The glen field needs to be a minimum value of one.

In terms of the NIST spec, we are a bit in a gray area here. NIST SP 800-90Ar1 says 
![Screenshot from 2024-06-28 15-32-01](https://github.com/lowRISC/opentitan/assets/20307557/a7587243-003d-4232-9bec-71fc51b0e77b)

What the `returned_bits` actually is if `requested_number_of_bits` = 0 isn't clear.

Right now, CSRNG produces one 128-bit block and also the following CTR_DRBG_Update is done and the reseed counter is incremented. I think this behavior is not intended from an RTL perspective. When receiving the Generate command in the command stage, there is just a check for `glen = 1`:
https://github.com/lowRISC/opentitan/blob/9cfd621ac24593a8f9cdd5482c933b39e8e13cf1/hw/ip/csrng/rtl/csrng_cmd_stage.sv#L378-L391

And then also the `u_prim_count_cmd_gen_cntr` output is compared against 1 again. But there is no check against 0. I think the current behavior is a coincidence.

Now that we have the capability to signal error status responses, it would probably be better to change the design to return an error response when receiving a Generate command with glen = 0.

glaserf:
I think this can now be implemented with very little effort. We have to also adapt dv slightly to a) trigger this case and b) check that it indeed produces a status error response.

briann-bui:
[...]

PR #23847, opened 2024-06-28
[csrng] Extend coverage definition for glen

vogelpi:
It turns out that the RTL doesn't handle Generate commands with a glen equal 0 (generate NO randomness) correctly. But this behavior documented already. We could correct the design to signal a error status response instead, but this really isn't a priority as a Generate command that shouldn't generate entropy is a corner case and even the NIST spec isn't fully clear what should happen here. See https://github.com/lowRISC/opentitan/issues/23846 for details.

Thus, this PR simply takes care of extending the coverage definition for the glen coverpoint as outlined  in #18350. This resolves https://github.com/lowRISC/opentitan/issues/18350.

### DIFF
--- a/hw/dv/sv/csrng_agent/csrng_item.sv
+++ b/hw/dv/sv/csrng_agent/csrng_item.sv
@@ -33,7 +33,7 @@ class csrng_item extends uvm_sequence_item;
 
   constraint c_glen {
     glen dist {
-      // TODO(#18350): Add a bin for 0 (with 5% chance, like for 4095?).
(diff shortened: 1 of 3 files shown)


h-filali (approved):
LGTM!

rswarbrick (approved):
Looks sensible to me. (And I love the names for the bins. A clever way to avoid needing to realign things :-D)

vogelpi:
> Looks sensible to me. (And I love the names for the bins. A clever way to avoid needing to realign things :-D)

The main motivation was to increase readability of the crosses :-)

PR #23852, opened 2024-06-29
[otbn,dv] Avoid killing during TL read in otbn_stack_addr_integ_chk

### LINKED ISSUE #23691: [test-triage] `otbn_stack_addr_integ_chk` CSR read timeout

Block level

### Failure Description

This has started failing on 6 June. I ran the test and this is what it outputs:
```
UVM_INFO @ 0 ps: reporter [RNTST] Running test otbn_base_test...
UVM_INFO @   2829560 ps: (otbn_single_vseq.sv:29) uvm_test_top.env.virtual_sequencer [uvm_test_top.env.virtual_sequencer.otbn_stack_addr_integ_chk_vseq] Loading OTBN binary from `/home/mvdmaas/repos/opentitan/scratch/master/otbn-sim-xcelium/0.otbn_stack_addr_integ_chk/latest/otbn-binaries/56788973262516559838207322170164997838916271268665892627247913373781531586908.elf'
UVM_INFO @   2829560 ps: (otbn_base_vseq.sv:493) uvm_test_top.env.virtual_sequencer [uvm_test_top.env.virtual_sequencer.otbn_stack_addr_integ_chk_vseq] 
	 ----| Waiting for OTBN to finish (polling)
UVM_INFO @   3121228 ps: (otbn_base_vseq.sv:493) uvm_test_top.env.virtual_sequencer [uvm_test_top.env.virtual_sequencer.otbn_stack_addr_integ_chk_vseq] 
	 ----| Waiting for OTBN to finish (polling)
UVM_INFO @  11037958 ps: (otbn_base_vseq.sv:493) uvm_test_top.env.virtual_sequencer [uvm_test_top.env.virtual_sequencer.otbn_stack_addr_integ_chk_vseq] 
	 ----| Waiting for OTBN to finish (polling)
UVM_INFO @  25329739 ps: (otbn_base_vseq.sv:493) uvm_test_top.env.virtual_sequencer [uvm_test_top.env.virtual_sequencer.otbn_stack_addr_integ_chk_vseq] 
	 ----| Waiting for OTBN to finish (polling)
UVM_FATAL @ 2014204650 ps: (csr_utils_pkg.sv:380) [csr_utils::csr_rd] Timeout waiting to csr_rd otbn_reg_block.status (addr=0x30e10018)
```

It looks like it's failing when it gets into the `run_otbn` function. The CSR read that is failing originates from `otbn_base_vseq`.

### Steps to Reproduce

- GitHub Revision: c4657e4c100e98d9fc7ab04394238e162e196c09
`./util/dvsim/dvsim.py hw/ip/otbn/dv/uvm/otbn_sim_cfg.hjson -i otbn_stack_addr_integ_chk -r 1 --fixed-seed 56788973262516559838207322170164997838916271268665892627247913373781531586908`

- [ ] otbn_stack_addr_integ_chk

rswarbrick:
Diagnosing this took me ages! In this situation (with err_type = 1), the vseq works forking into two processes. One injects an error on every stack write. The other waits until it sees a read from the stack, which should trigger an alert, and then tells the model to go into a locked state.

Once the latter process finishes, we can release our forced signals and wait until the RTL and the model raise an alert at the same time. At this point, we use "disable fork"

The first process waited for reads from the stack using the wait_for_flag() task which, in turn, used wait_for_run_completion() to detect the end of a run (at which point, we weren't going to see the flag change, so should stop early).

Unfortunately, wait_for_run_completion would sometimes wait for the end of the run by polling the STATUS register. Doing so would normally involve a TL transaction. The test was failing because we started a TL transaction to read STATUS and then killed a process that was inside the sequencer, in wait_for_item_done(). This meant that the transaction blocked forever and we never passed the response back to the sequence. As a result, the csr_rd task timed out and the test failed.

The fix is rather simple in the end: if you want your process to be disabled, don't do a front-door TL transaction.

Fixes #23691.

### DIFF
--- a/hw/ip/otbn/dv/uvm/env/seq_lib/otbn_base_vseq.sv
+++ b/hw/ip/otbn/dv/uvm/env/seq_lib/otbn_base_vseq.sv
@@ -475,8 +475,10 @@ class otbn_base_vseq extends cip_base_vseq #(
 
   // Wait for OTBN to finish, either by polling or by waiting on the interrupt pins
   //
-  // When polling, this waits up to 1ms before failing with a timeout.
-  task wait_for_run_completion(input bit verbosity);
+  // When polling, this waits up to 1ms before failing with a timeout. If polling, we pass the
+  // backdoor flag to csr_rd, which is a way to avoid starting requests on the TL sequencer.
+  // Avoiding this means we can safely kill the process without leaving anything hanging.
+  task wait_for_run_completion(input bit verbosity, input bit backdoor=0);
     fork begin : isolation_fork
       fork
         begin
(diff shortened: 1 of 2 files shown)


hcallahan-lowrisc (approved):
This makes sense to me! Thanks @rswarbrick

PR #23854, opened 2024-06-30
[hw,dma,rtl] Set interrupt only when transition to done / error

Razer6:
Previously, clearing the DONE or ERROR interrupt before clearing STATUS.done would raise the same interrupt a second time, which could bring FW out of sync.

With this change, the DONE and ERROR interrupt are only sent once when transitioning into the DONE or ERROR state.

/cc @neeraj-rv @sameo

### DIFF
--- a/hw/ip/dma/data/dma.hjson
+++ b/hw/ip/dma/data/dma.hjson
@@ -634,7 +634,7 @@
         { bits: "1"
           name: "done"
           resval: 0x0
-          hwaccess: "hrw"
+          hwaccess: "hwo"
           desc: '''
                 Configured DMA operation is complete.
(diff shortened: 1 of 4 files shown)


alees24 (changes requested):
This change will cause the 'dma_done' and 'dma_error' interrupt to become asserted as soon as the DMA controller leaves its Idle state. I think you want to set the `intr_state` bits only when the 'set' clause of '...status.done.de' or '...status.error.de' is true.

Razer6:
Thats true indeed, lets AND the .d signal to it

alees24 (approved):

Razer6:
This PR is not necessary anymore because we switched to status-type interrupts.

PR #23855, opened 2024-07-01
[hmac,dv] Increase SHA2 test vector timeout

martin-velay:
Regression failed for some seeds due to timeout for SHA2-512 test only:

```
`UVM_FATAL (uvm_phase.svh:1512) [PH_TIMEOUT] Explicit timeout of * ps hit, indicating a probable testbench issue` has 1 failures:
    * Test hmac_test_sha512_vectors has 1 failures.
        * 3.hmac_test_sha512_vectors.108683676871538303991718868438741530266571739654349379393166068314816946587950\
          Line 132586, in log /home/dev/src/scratch/cov_test/hmac-sim-vcs/3.hmac_test_sha512_vectors/latest/run.log

                UVM_FATAL @ 1000000000000 ps: (uvm_phase.svh:1512) [PH_TIMEOUT] Explicit timeout of 1000000000000 ps hit, indicating a probable testbench issue
                UVM_INFO @ 1000000000000 ps: (uvm_report_catcher.svh:705) [UVM/REPORT/CATCHER]

```

### DIFF
--- a/hw/ip/hmac/dv/hmac_sim_cfg.hjson
+++ b/hw/ip/hmac/dv/hmac_sim_cfg.hjson
@@ -109,7 +109,7 @@
       name: hmac_test_sha512_vectors
       uvm_test_seq: hmac_test_vectors_sha_vseq
       // Increase timeout for all test iterations to pass
-      run_opts: ["+test_vectors_dir={build_dir}/src/lowrisc_dv_test_vectors_0 +test_timeout_ns=1_000_000_000 +sha2_digest_size=SHA2_512"]
+      run_opts: ["+test_vectors_dir={build_dir}/src/lowrisc_dv_test_vectors_0 +test_timeout_ns=1_200_000_000 +sha2_digest_size=SHA2_512"]
     }
 
     {


gdessouky (approved):
LGTM, I don't know if this is better or reducing the test vectors per run, but so long as there is no restriction on how far we can go with the timeout, I'm good with this, thanks!

rswarbrick (approved):
This looks perfectly sensible to me. Do we have a reason to believe that 1.2 seconds is a correct upper bound? (If so, brill! But I thought it was worth understanding where the estimate comes from)

martin-velay:
> This looks perfectly sensible to me. Do we have a reason to believe that 1.2 seconds is a correct upper bound? (If so, brill! But I thought it was worth understanding where the estimate comes from)

No I don't have any mathematical estimate, only watching the regression report. As 4 are passing over 5 test with 1 sec, I thought that 1.2 should be sufficient.

gdessouky:
> > This looks perfectly sensible to me. Do we have a reason to believe that 1.2 seconds is a correct upper bound? (If so, brill! But I thought it was worth understanding where the estimate comes from)
> 
> No I don't have any mathematical estimate, only watching the regression report. As 4 are passing over 5 test with 1 sec, I thought that 1.2 should be sufficient.

that's also how I experimentally increase these timeouts :-)

rswarbrick:
Fingers crossed then :-D

PR #23856, opened 2024-07-01
[hmac,dv] Fix secret wiping issues

gdessouky:
This undoes a minor DV change in https://github.com/lowRISC/opentitan/pull/23817 intended to address rare issues where wipe_secret_triggered is not pulled back to 0. This conflicts with another change that was on master that was meant to do the same, so the digest prediction fails at some points. We can either just merge this fix or unmerge that other PR. 

I am still running `stress_all` and `stress_all_with_rand_reset` at my end locally to make sure though.

### DIFF
--- a/hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
+++ b/hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
@@ -225,9 +225,6 @@ class hmac_smoke_vseq extends hmac_base_vseq;
       `uvm_info(`gfn, $sformatf("reading digest"), UVM_LOW)
       rd_digest();
     end
-    // Ensure that wipe secret flag is cleared for the next sequence, as it may cause digest
-    // comparison issue with the stress tests
-    clear_wipe_secret();
   endtask : body
 
 endclass : hmac_smoke_vseq


gdessouky:
`stress_all` regression runs fine at my end. Running `stress_all_with_rand_reset` now.

rswarbrick:
I'm not really an expert on this (yet). @martin-velay: would you mind writing something to explain what's going on?

martin-velay:
@rswarbrick 

In this PR I am explaining what I saw but unfortunately I didn't marked down the related seed.

Two months back Ghada has made a similar change to fix this, but it looks it has been lost somewhere for a reason I cannot explain, as I don't understand. Here below is the modif made by Ghada which should be kept instead of my change:
https://github.com/lowRISC/opentitan/blob/efbc2e1e786243b1ceff62a31f918651de671ccf/hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv#L96

Sorry for this mess and thanks @gdessouky for pointed your change.

martin-velay (approved):

rswarbrick (approved):
Thanks for tidying things up (and thanks, @martin-velay, for explaining carefully).

PR #23860, opened 2024-07-01
[entropy_src/dv] Close toggle coverage

h-filali:
This PR adds some exclusions for the toggle coverage analysis.

The signals that are included here are:
- cnt_after_commit_o since they aren't connected and the same signal is already excluded for most counters.
- decr_en_i, step_i for various FIFOs, since they are hard coded.
- commit_i for various counters, since they are also hard coded.

This is a follow up on #23842


vogelpi (approved):
LGTM!

vogelpi:
DV has passed except for the FPGA tests which are irrelevant for this PR. -> Merging.

Issue #23606, opened 2024-06-10
[test-triage] chip_sw_csrng_edn_concurrency_reduced_freq regularly times out


pamaury:
### Hierarchy of regression failure

Chip Level

### Failure Description

The test `chip_sw_csrng_edn_concurrency_reduced_freq` regularly times out.

### Steps to Reproduce

- GitHub Revision: [a182fcef27](https://github.com/lowrisc/opentitan/tree/a182fcef27866fb502d8cceb810bdf76cfa3ef42)
- dvsim invocation command to reproduce the failure, inclusive of build and run seeds:
  ./util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_csrng_edn_concurrency_reduced_freq --build-seed 115716131103921631007013649731972014580281041353363476420230431751664670300928 --fixed-seed 40308221990956139403360512504997844009802296242472870101329624554885491880045 --waves fsdb
- Kokoro build number if applicable

pamaury:
Fixed by #23620 ?

vogelpi:
> Fixed by #23620 ?

Unfortunately not :-/

vogelpi:
I've now found the issue and fixed it with #23861.

Basically, the test was waiting for the esfinal FIFO to fill up before starting the test loop. But in the reduced_freq configuration, it can be that the FIFO never fills up as the RNG produces raw entropy at a lower rate while the AST consuming part constantly consumes entropy from EDN0.

PR #23861, opened 2024-07-01
[chip-test/sival] Fix csrng_edn_concurrency_reduced_freq

vogelpi:
Inspecting the waves revealed that the reduced_freq variant of this test sometimes didn't make it across the initial EDN and CSRNG re-config section. The reason is that the RNG inside AST produces raw entropy at a lower rate while the entropy consuming part of AST still consumes a reasonable amount of entropy. As a result, the esfinal FIFO would never fill up and the reconfiguration section would never finish.

To fix this, this commit adds a timeout for the esfinal FIFO: If the FIFO hasn't filled up within 32 ms (under normal conditions this should suffice to fill the FIFO easily), we proceed without a full FIFO.

This fixes lowRISC/OpenTitan#23606.

### DIFF
--- a/sw/device/tests/csrng_edn_concurrency_test.c
+++ b/sw/device/tests/csrng_edn_concurrency_test.c
@@ -57,7 +57,15 @@ enum {
   /*
    * The number of entries of the esfinal FIFO of ENTROPY_SRC.
    */
-  kEntropySrcEsFinalFifoDepth = 3
+  kEntropySrcEsFinalFifoDepth = 3,
+  /*
+   * Parameters for controlling the waiting until the esfinal FIFO fills up.
+   * Under normal operating conditions, producing one seed takes roughly 4 ms.
+   * We wait for at most 32 ms for 6 seeds (2 for the EDNs, 1 for CSRNG, 3 to
+   * fill the esfinal FIFO).
+   */
+  kEsFinalFifoIterationTimeUs = 100,
+  kEsFinalFifoNumIterationsMax = 320
 };
 
 /**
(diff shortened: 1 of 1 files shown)


sw/device/tests/csrng_edn_concurrency_test.c
```diff
@@ -394,8 +402,21 @@ static void entropy_config(void) {
...
+  // Under normal operating conditions, producing one seed takes roughly 4 ms.
+  // We wait for at most 32 ms for 6 seeds (2 for the EDNs, 1 for CSRNG, 3 to
+  // fill the esfinal FIFO). In the reduced_freq varaint of the test, raw
+  // entropy may be generated at a lower rate. Depending on the configuration of
+  // the EDNs, the constant entropy consumption of AST may prevent the esfinal
+  // FIFO from filling up.
   dif_entropy_src_debug_state_t debug_state;
+  uint32_t iter_count = 0;
   do {
+    busy_spin_micros(kEsFinalFifoIterationTimeUs);
```
pamaury, line 415: The code looks ok but I think using `ibex_timeout_{init,check}` would be a bit cleaner.

pamaury (approved):

timothytrippel (approved):

sw/device/tests/csrng_edn_concurrency_test.c
vogelpi: Thanks @pamaury . In the interest of time, I leave this as is but I will investigate the `ibex_timeout_{init,check}` option next time I need something similar. Thanks for the pointer.

PR #23848, opened 2024-06-28
[hmac,dv] Optimize regression duration

martin-velay:
- reduce reseeding for HMAC SHA-2/HMAC vectors tests as the randomization space is quite narrow
  - remove these tests from the stress test as it doesn't bring anything
  - related to this issue #23844

PR #23862, opened 2024-07-01
[hmac,dv] Extend timeout on outstanding accesses for reset test

gdessouky:
This extends the timeout on outstanding accesses when issuing reset for `hmac_stress_all_with_rand_reset` to prevent the test timing out for messages/vectors where there is always outstanding accesses. This change, along with PR https://github.com/lowRISC/opentitan/pull/23848 which excludes the long HMAC/SHA-2 test vectors from this sequence, this should close https://github.com/lowRISC/opentitan/issues/23559.

This reduces to 1 commit 37cdc5fe4bc2a0d3675424be0ae11a1bbfdf541c once a previous PR gets merged.


gdessouky:
I'm still running the test locally to make sure this timeout is enough.

martin-velay (approved):
I don't really know how this works, but I can see similar stuff in some other DV IP. I guess it's fine then.

rswarbrick (approved):
Looks sensible to me, thanks.

PR #23864, opened 2024-07-01
[hmac,dv] Fix msg_length coverage

martin-velay:
- message length is expressed in bits but actually, the register granularity is in byte. The coverage bins have to be reviewed in consequence.

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_env_cov.sv
+++ b/hw/ip/hmac/dv/env/hmac_env_cov.sv
@@ -76,23 +76,24 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
                                               logic [TL_DW-1:0] msg_len_upper,
                                               logic [TL_DW-1:0] cfg          );
     hmac_en: coverpoint cfg[HmacEn];
+    // Message length is byte granularity
     msg_len_lower_cp: coverpoint (msg_len_lower) {
       bins len_0         = {0};     // Empty message
-      bins len_1         = {1};     // One byte message
-      bins len_511       = {511};   // One block in SHA-2 256, -1 byte
+      bins len_8         = {8};     // One byte message
+      bins len_504       = {504};   // One block in SHA-2 256, -1 byte
       bins len_512       = {512};   // One block in SHA-2 256
-      bins len_513       = {513};   // One block in SHA-2 256, +1 byte
-      bins len_1023      = {1023};  // One block in SHA-2 384/512 or two in SHA-2 256, -1 byte
+      bins len_520       = {520};   // One block in SHA-2 256, +1 byte
+      bins len_1016      = {1016};  // One block in SHA-2 384/512 or two in SHA-2 256, -1 byte
       bins len_1024      = {1024};  // One block in SHA-2 384/512 or two in SHA-2 256
-      bins len_1025      = {1025};  // One block in SHA-2 384/512 or two in SHA-2 256, +1 byte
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
This is reasonable, but I notice that the documentation says the bottom three bits of the `MESSAGE_LENGTH_LOWER` are ignored. Would it make more sense to change the sampling code in the scoreboard to look more like:
```
          if (cfg.en_cov) begin
            cov.msg_len_cg.sample(.msg_len_lower(item.d_data >> 3),
                                  .msg_len_upper('x),
                                  .cfg(`gmv(ral.cfg)));
          end
```
?

martin-velay:
@rswarbrick
As the documentation refers to block size in terms of bits, I would prefer to keep this in bits. This makes it easier to see what's covered and avoid to make some conversions.

rswarbrick (approved):
Cool, that makes sense to me.

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -76,23 +76,24 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
                                               logic [TL_DW-1:0] msg_len_upper,
                                               logic [TL_DW-1:0] cfg          );
     hmac_en: coverpoint cfg[HmacEn];
+    // Message length is byte granularity
     msg_len_lower_cp: coverpoint (msg_len_lower) {
       bins len_0         = {0};     // Empty message
-      bins len_1         = {1};     // One byte message
-      bins len_511       = {511};   // One block in SHA-2 256, -1 byte
+      bins len_8         = {8};     // One byte message
+      bins len_504       = {504};   // One block in SHA-2 256, -1 byte
```
gdessouky, line 83: If I understand correctly, these bins are used to measure coverage for these specific cases, correct?

If yes, then I would also measure the specific case when it is exactly a single block that will get processed. This happens when the message length is exactly = 448-bits, and then 64-bits are reserved for the message length, totaling 512-bit block for SHA-2 256.

This is also 896 bits for SHA-2 384/512 (1024-128 bits) where 128 is reserved to denote message length.

gdessouky: OK, merge was faster than my comment :) But still think it would make sense to look into this again still for an upcoming PR @martin-velay @rswarbrick WDYT?

martin-velay: Here we are sampling the message length data seen from the register itself.
In the scoreboard, to predict this value we are only taking the queue size which contains only the message:
https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/hmac/dv/env/hmac_scoreboard.sv;l=113
And we just split the bits according to the register map:
https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/hmac/dv/env/hmac_scoreboard.sv;l=828-832

So you are saying that our constraints are wrong too and we are actually not sure to cover exactly one block! I am referring to those constraints:
https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv;drc=32a1a99285e3da002c5de3c135b235cdb9a83dc4;l=35-45

We are probably not going to find bug there but still better to update the constraints and the coverage model. I completely missed this info from the spec ! This 512 and 1024 bits are actually matching with some HW sizes? I mean buffers, registers or anything else, right?

PR #23867, opened 2024-07-01
[lc_ctrl,dv] Add more staggering in rma signals

### LINKED ISSUE #21268: [lc_ctrl/dv] Improve coverage for RMA ACK signals
### Description

In #21267 we added additional RMA ACK channels to LC_CTRL so that the ACK signals from OTBN and FLASH_CTRL can be routed back to LC_CTRL in parallel. 

While the DV environment has been adapted to accommodate this change, and while CDC randomization does introduce some level of signal latency variations between the ACK channels, it would be good to improve the DV environment that some sequences assert the ACK channels at different times (several 10s or 100s of cycles appart) in order to better reflect reality.

rswarbrick:
Detailed notes in commits. This fixes #21268.

### DIFF
--- a/hw/ip/lc_ctrl/dv/env/lc_ctrl_if.sv
+++ b/hw/ip/lc_ctrl/dv/env/lc_ctrl_if.sv
@@ -133,11 +133,27 @@ interface lc_ctrl_if (
     clk_byp_ack_i = val;
   endtask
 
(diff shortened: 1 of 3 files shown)


hcallahan-lowrisc (approved):
LGTM, thanks Rupert

PR #23868, opened 2024-07-01
[flash_ctrl,dv] Drop some unused stuff from the smoke sequence

rswarbrick:
I think this predates the infrastructure that's now in `lc_ctrl_errors_vseq.sv`. When that was merged, the stuff in the smoke sequence seems to have been disabled but left in place unused. Remove it.


marnovandermaas (approved):
I think this is fine, but please do investigate the CI failures before merging.

PR #23870, opened 2024-07-01
[opentitantool]: add command line option for setting manifest timestamp

vbendeb:
The new command line option given to 'image manifest update' allows the user to set the manifest 'timestamp' field to the passed in 64 bit integer value.

Veified by invoking opentitantool with the new command line option and examining the resulting manifest header contents.

### DIFF
--- a/sw/host/opentitanlib/src/image/image.rs
+++ b/sw/host/opentitanlib/src/image/image.rs
@@ -506,6 +506,14 @@ impl Image {
     pub fn compute_digest(&self) -> Result<sha256::Sha256Digest> {
         self.map_signed_region(|v| sha256::sha256(v))
     }
+
+    pub fn update_timestamp(&mut self, timestamp:u64)->Result<()> {
+        let manifest = self.borrow_manifest_mut()?;
+
+        manifest.timestamp.timestamp_high = (timestamp >> 32) as u32;
+        manifest.timestamp.timestamp_low = timestamp as u32;
+        Ok(())
+    }
 }
 
 impl ImageAssembler {
--- a/sw/host/opentitantool/src/command/image.rs
+++ b/sw/host/opentitantool/src/command/image.rs
@@ -150,6 +150,9 @@ pub struct ManifestUpdateCommand {
     /// Filename to write the output to instead of updating the input file.
     #[arg(short, long)]
     output: Option<PathBuf>,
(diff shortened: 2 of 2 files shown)


timothytrippel (commented):
The way opentitantool handles the `manifest update` command is via an input JSON configuration file that contains all the manifest fields you want to fill, including the timestamp. Additionally, we have a [custom Bazel rule](https://cs.opensource.google/opentitan/opentitan/+/master:rules/manifest.bzl;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=167) that generates the manifest JSON input file for you. For example, if you build [this manifest bazel target](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/rom_ext/BUILD;drc=129185003b61f7444293d116f2e015576700f81a;l=267), with `bazel build //sw/device/silicon_creator/rom_ext:manifest`, and query to find the output JSON file (with `./bazelisk.sh outquery-all //sw/device/silicon_creator/rom_ext:manifest`), you can examine its contents to see what said file looks like.

With opentitantool's current implementation, you are free to create your own JSON files that define a manifest manually (with any timestamp you wish embedded in the JSON), or you can use your Bazel rule to create a target for you that can gen the file for you.

As is, it looks like the code you are proposing provides a second avenue to update the timestamp (i.e., overwriting what could potentially be in the JSON input file already). It seems this is a one-off parameter, as all other manifest parameters must be provided in the JSON file. Is there a reason the JSON input file format does not fit your needs regarding the timestamp value?

vbendeb:
I guess my am looking at this from the other security chips' workflow where the currently running image is updating the chip to the next firmware version, which is placed in the alternative flash section. Then the chip is restarted and the newer image is run. Both images have the same version, the newer image has more recent timestamp.

This is not relevant where the new image is always replacing the currently running image by ROM (or ROM_EXT), but this is not something we can use in prod, because in this case interrupted update would brick the chip. We (ti50) are working on adding update over USB or TPM interface, this is when the timestamp would become important.

But I think I saw another PR where the zero timestamp in the manifest would trigger plugging in the current time into the manifest section of the image - this would be sufficient, so this PR can be discarded.

vbendeb:
looks like placing the timestamp into manifest.json is the way to go.

Issue #22879, opened 2024-04-29
[flash_ctrl,dv] Lots of broken block-level tests in flash_ctrl


rswarbrick:
### Hierarchy of regression failure

Block level

### Failure Description

The nightly DV runs report lots of failing tests in `flash_ctrl`. They seem to have started breaking on 28th April, including the following tests:
- flash_ctrl_access_after_disable
- flash_ctrl_integrity
- flash_ctrl_intr_wr
- flash_ctrl_intr_wr_slow_flash
- flash_ctrl_oversize_error
- flash_ctrl_rw_derr
- flash_ctrl_sec_cm
- flash_ctrl_write_word_sweep
- flash_ctrl_rw_serr
- flash_ctrl_prog_reset
- flash_ctrl_rw_evict
- flash_ctrl_derr_detect
- flash_ctrl_serr_address
- flash_ctrl_serr_counter
- flash_ctrl_rw_evict_all_en
- flash_ctrl_fs_sup

@matutem: I think you landed a large flash_ctrl PR a few days ago. Would you mind taking a look at some of these: I think your PR might well be responsible for the failures.

### Steps to Reproduce

- GitHub Revision: ae68723071

vogelpi:
Those failures are all due to enabling ECC more widely in block-level DV some time ago. Before doing this, the nightlies all looked good. The last RTL change was already before that.

In the meantime, @matutem has been adjusting more block-level DV sequences to let them deal with ECC errors correctly. There are still some to do but I am not to worried about this.

vogelpi:
The current status (May 19 nightly regression) is as follows:
- 100 % pass rate for all tests except for
- flash_ctrl_rw_evict - 70%
- flash_ctrl_rw_evict_all_en - 85%
- flash_ctrl_oversize_error - 80%
- flash_ctrl_intr_rd - 95%
- flash_ctrl_intr_wr - 90%
- flash_ctrl_read_word_sweep_derr - 40%
- flash_ctrl_rw_derr - 60%
- flash_ctrl_integrity - 40%
- flash_ctrl_read_word_sweep_serr - 20%
- flash_ctrl_rw_serr - 80%
- flash_ctrl_wo - 85%
- flash_ctrl_rw - 95%
- flash_ctrl_fs_sup - 80%
- flash_ctrl_config_regwen - 80%

andreaskurth:
[...]

PR #23871, opened 2024-07-01
[dv,flash_ctrl] Fix some block level failures

matutem:
- Improve tracking of words with error injection so only one error of any kind is injected per word.
- Improve initialization of data in flash_ctrl_read_word_sweep. There is still room for improvements.

Addresses #22879


vogelpi (approved):
Thanks for working on this @matutem ! The PR looks good, except for the lint errors. Those should be fixed before merging.

Since this PR changes the environment config as well as the otf scoreboard, I was wondering whether we should run a full regression with coverage before merging this. I can imagine that there is a risk of these changes impacting many other tests. WDYT?

PR #23873, opened 2024-07-01
[opentitantool] Treat empty timestamp manifest field as now

jettr:
When updating an image with a manifest file, treat a non-present timestamp field as the current UTC time by default. This behavior can be overridden with `--empty_manifest_timestamp_is_now==false`

### DIFF
--- a/sw/host/opentitanlib/src/image/image.rs
+++ b/sw/host/opentitanlib/src/image/image.rs
@@ -10,6 +10,7 @@ use std::fs::File;
 use std::io::{Read, Write};
 use std::mem::{align_of, size_of};
 use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
 use thiserror::Error;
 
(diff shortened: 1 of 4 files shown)


timothytrippel (approved):

cfrantz (changes requested):
Please don't do this.  Timestamps break build reproducibility.

If you want to set a timestamp in your manifest, you must do it manually.

jettr:
Agree that timestamp break reproducible, so if one needs reproducible builds for a release, then that manifest.json for that build should include a specific timestamp.

I really think the default behavior if there is no timestamp present (meaning you don't care about reproducible builds) should be to use the current time. This is especially helpful for developers building new images locally without bumping the minor version to be able to update to the latest dev-built image (assuming version and timestamp are both under consideration for which image is newer).

Also if we don't have now as the default for absent timestamp fields, then we are going to end up with official builds with a timestamp of 0, which also seems incorrect.

Issue #21855, opened 2024-03-06
[entropy_src] Backpressure handling follow-up work


vogelpi:
### Description

What is left from #21799 for after M2:

- [x] Adding an SVA to ensure the FIFOs are indeed emptied before the conditioner runs, see #23877.
- [x] Documentation, see #22324.
- [x] Adjusting @pamaury 's observe mode TLT/SiVal test to check the new alert bit, see #23875.
- [ ] Ironing out remaining test failures, see #23893.

vogelpi:
We should do this for M5. Discussed with @h-filali that this one of the next things to investigate after coverage closure.

IIRC, I've only noted test failures due to backpressure in the rng_max_rate test.

PR #23875, opened 2024-07-01
[chip-test/sival] Check ENTRORPY_SRC_POSTHT_ENTROPY_DROP_ALERT_BIT

vogelpi:
When running in Firmware Override: Observe mode, this bit should be checked. If it's asserted, this means that health tested entropy bits have been dropped before the postht FIFO due to conditioner back pressure. This means entropy observed via observe FIFO may be non-contiguous.

This is related to lowRISC/OpenTitan#21855.

### DIFF
--- a/sw/device/tests/entropy_src_fw_observe_many_contiguous.c
+++ b/sw/device/tests/entropy_src_fw_observe_many_contiguous.c
@@ -73,6 +73,20 @@ static status_t let_observe_fifo_overflow(uint32_t timeout_usec) {
   return OK_STATUS();
 }
 
+/**
+ * Determine whether health tested entropy bits have been dropped before the
+ * postht FIFO due to conditioner back pressure.
+ */
+bool postht_entropy_dropped(void) {
+  uint32_t recov_alert_sts;
+  CHECK_DIF_OK(
+      dif_entropy_src_get_recoverable_alerts(&entropy_src, &recov_alert_sts));
+  bool dropped = bitfield_bit32_read(
+      recov_alert_sts,
+      ENTROPY_SRC_RECOV_ALERT_STS_POSTHT_ENTROPY_DROP_ALERT_BIT);
+  return dropped;
+}
+
 // Configure the entropy complex.
 static status_t entropy_config(
     dif_entropy_src_single_bit_mode_t single_bit_mode) {
@@ -136,6 +150,10 @@ status_t firmware_override_observe(
     // Collect samples from the the observe FIFO.
     uint32_t words_to_read = nr_sample_words;
     uint32_t *sample_buffer_ptr = sample_buffer;
+    // Clear POSTHT_ENTROPY_DROP_ALERT bit.
+    TRY(dif_entropy_src_clear_recoverable_alerts(
+        &entropy_src,
+        ENTROPY_SRC_RECOV_ALERT_STS_POSTHT_ENTROPY_DROP_ALERT_BIT));
     // Drain FIFO to make sure we get contiguous samples.
     LOG_INFO("drain observe FIFO overflow...");
     TRY(entropy_testutils_drain_observe_fifo(&entropy_src));
@@ -143,6 +161,10 @@ status_t firmware_override_observe(
     ibex_timeout_t tmo = ibex_timeout_init(timeout_usec);
     while (words_to_read > 0 && !ibex_timeout_check(&tmo)) {
       size_t len = words_to_read;
+      // Make sure no health tested entropy bits were dropped before the
+      // postht FIFO.
+      TRY_CHECK(!postht_entropy_dropped(),
+                "entropy bits dropped before postht FIFO");
       // Check FIFO did not overflow during collection.
       TRY_CHECK(!entropy_src_fifo_has_overflowed(),
                 "observe FIFO overflowed during collection");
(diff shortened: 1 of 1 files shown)


h-filali (approved):
LGTM!

rswarbrick:
As far as I can see, the CI check that just failed is an IT issue (so presumably not caused by the change in the PR)

sw/device/tests/entropy_src_fw_observe_many_contiguous.c
```diff
@@ -73,6 +73,20 @@ static status_t let_observe_fifo_overflow(uint32_t timeout_usec) {
   return OK_STATUS();
 }
 
+/**
+ * Determine whether health tested entropy bits have been dropped before the
+ * postht FIFO due to conditioner back pressure.
+ */
+bool postht_entropy_dropped(void) {
+  uint32_t recov_alert_sts;
+  CHECK_DIF_OK(
```
jwnrt, line 82: NIT: Generally we prefer to `TRY` in functions and return a `status_t`, but I know this makes the API harder to use.

jwnrt (approved):
LGTM, thanks

sw/device/tests/entropy_src_fw_observe_many_contiguous.c
jwnrt: You can probably leave it. There was a time when we wanted to move entirely over to `TRY` but we can do this in one larger effort later.

engdoreis (approved):

sw/device/tests/entropy_src_fw_observe_many_contiguous.c
```diff
@@ -136,13 +150,21 @@ status_t firmware_override_observe(
     // Collect samples from the the observe FIFO.
     uint32_t words_to_read = nr_sample_words;
     uint32_t *sample_buffer_ptr = sample_buffer;
+    // Clear POSTHT_ENTROPY_DROP_ALERT bit.
+    TRY(dif_entropy_src_clear_recoverable_alerts(
+        &entropy_src,
+        ENTROPY_SRC_RECOV_ALERT_STS_POSTHT_ENTROPY_DROP_ALERT_BIT));
     // Drain FIFO to make sure we get contiguous samples.
     LOG_INFO("drain observe FIFO overflow...");
     TRY(entropy_testutils_drain_observe_fifo(&entropy_src));
```
pamaury, line 159: Should the alert clearing be done after draining the FIFO? I think potentially one could be unlucky and the alert could trigger again between drain the FIFO and clearing the error?

pamaury (approved):
Looks good to me but I am not entirely sure the alert clearing is done at the right moment.

sw/device/tests/entropy_src_fw_observe_many_contiguous.c
vogelpi: Good question! Let me try to explain: This alert is independent of the observe FIFO. If the observe FIFO is full, we don't add further data to the observe FIFO but it continues to flow downstream into the conditioner (we're running Firmware Override: Observe mode here). In case of conditioner back pressure (it's very very unlikely to see this under normal operating conditions), entropy is being dropped at the input of the postht FIFO which is the FIFO before the observe FIFO. Draining the observe FIFO has zero impact on whether we're going to see such an alert.

If we first drained the observe FIFO and then cleared the alert bit, it could be that between draining the FIFO and clearing the alert bit, the conditioner exhibits back pressure and we drop a sample at the postht input. Since we already drained the observe FIFO and restarted collection, the collected entropy is no longer contiguous. Which is the issue we need to avoid using this alert bit.

vogelpi:
The failing TLT is `chip_sw_csrng_kat_test ` but the failure is due to an infrastructure issue:
```
* `Job returned non-zero exit code` has 1 failures:
    * Test chip_sw_csrng_kat_test has 1 failures.
        * 0.chip_sw_csrng_kat_test.1\
         Log /azp/agent/_work/1/s/scratch/HEAD/chip_earlgrey_asic-sim-xcelium/0.chip_sw_csrng_kat_test/latest/run.log

                			done; \
                		fi; \
                	fi; \
                done;
                Building SW image "//sw/device/tests:csrng_kat_test_sim_dv".
                Building "//sw/device/tests:csrng_kat_test_sim_dv" on network connected machine.
                Building with command: ./bazelisk.sh build  --define DISABLE_VERILATOR_BUILD=true --//hw/ip/otp_ctrl/data:img_seed=1 //sw/device/tests:csrng_kat_test_sim_dv
                Opening zip "/root/.cache/bazelisk/downloads/bazelbuild/bazel-6.2.1-linux-x86_64/bin/bazel (deleted)": open(): No such file or directory
                FATAL: Failed to open '/root/.cache/bazelisk/downloads/bazelbuild/bazel-6.2.1-linux-x86_64/bin/bazel (deleted)' as a zip file: (error: 2): No such file or directory
                make: *** [/azp/agent/_work/1/s/hw/dv/tools/dvsim/sim.mk:73: sw_build] Error 36
```

I just ran 10 seeds of this TLT locally and they all pass. I am thus merging this. Thanks for your reviews everybody!

sw/device/tests/entropy_src_fw_observe_many_contiguous.c
pamaury: Thanks for the explanation

vogelpi: No worries, thanks for your review!

PR #23877, opened 2024-07-01
[entropy_src] Add SVA to ensure FIFOs are empty when triggering SHA

vogelpi:
If any of the postht, distr or precon FIFOs are non-empty when triggering the conditioner, there is risk of leaving entropy bits tested as part of the current window inside the pipeline instead of absorbing them into the current seed. This would be problematic.

This is related to lowRISC/OpenTitan#21855.

### DIFF
--- a/hw/ip/entropy_src/rtl/entropy_src_core.sv
+++ b/hw/ip/entropy_src/rtl/entropy_src_core.sv
@@ -3465,6 +3465,20 @@ module entropy_src_core import entropy_src_pkg::*; #(
       precon_post_startup_exp_push_bit_cnt_q  <= precon_post_startup_exp_push_bit_cnt_d;
     end
   end
+
+  // When triggering the conditioner, the precon FIFO must be empty. The postht and distr FIFOs
+  // must have been empty in the cycle before. Otherwise, some entropy bits tested as part of the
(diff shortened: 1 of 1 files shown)


vogelpi:
This PR just adds an SVA to an RTL file which is okay.

CHANGE AUTHORIZED: hw/ip/entropy_src/rtl/entropy_src_core.sv

h-filali (approved):
LGTM! Very good idea to add this.

rswarbrick (approved):
Looks sensible, and thanks for the explanatory comment. It definitely helped me understand a bit more about what's going on.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/entropy_src/rtl/entropy_src_core.sv

vogelpi:
Thanks for your reviews both. I had to update the SVA to disable it in Firmware Override: Extract and Insert mode. Because in this mode firmware is responsible for operating the conditioner:
- The postht and distr FIFOs don't matter at all.
- If firmware triggers the conditioner without the precon FIFO being empty, an alert is raised in this mode.

vogelpi:
I am merging this. The relevant CI checks (lint and DV, FPGA build) have all passed.

PR #23878, opened 2024-07-02
[dv, chip]: Add pull delay and pad_in_dis_val

walaahattab:
Add the following updates to **chip_padctrl_attributes_vseq**:
- Pull delay after configuring the pad pull attributes.
- "pad_in_dis_val": As the value of the pad_in differs between open source and closed source when the input is disabled, with the current configuration we are failing when the pad’s input is disabled.


walaahattab:
@sha-ron

eshapira (approved):

Issue #23879, opened 2024-07-02
[hmac,dv] FIFO status to be cycle accurate


martin-velay:
### Description

Some tests might fail sometimes related to the status register as it's hard to be cycle accurate.

This PR https://github.com/lowRISC/opentitan/pull/23880 may help but doesn't solve all, as this is not really important this will be covered later.

martin-velay:
This issue can be closed as this PR https://github.com/lowRISC/opentitan/pull/25683 has been merged

PR #23880, opened 2024-07-02
[hmac,dv] Attempt to get cycle accurate status

martin-velay:
Related to this issue https://github.com/lowRISC/opentitan/issues/23879


martin-velay:
This solution has been preferred https://github.com/lowRISC/opentitan/pull/25683 and is now merged

PR #23887, opened 2024-07-02
[pinmux,fpv] Fix PinmuxTargetCfg initialiser in pinmux_tb.sv

rswarbrick:
The target_cfg_t gained a couple of fields with commit be4459d2b24 and Jasper (quite reasonably) fails if its input script doesn't match.

This fixes things up in hw/ip/pinmux/fpv/tb and also hw/top_earlgrey/ip/pinmux/fpv/tb/pinmux_tb.sv.

(This is kind of required because the nightly FPV runs are currently N/A for pinmux, due to the compilation failure)


a-will (approved):
Sorry I missed this!

vogelpi:
Thanks for catching and fixing this @rswarbrick !

PR #23889, opened 2024-07-02
[dv/dpi] tcpserver single connection and retry select on EINTR

kwalker27:
Two small patches to the `tcp_server` code, used by other transactors such as `jtagdpi` and `dmidpi`. These improve the usability of the transactors in a shared, remote environment such as an emulator.

Without this change, an interrupted select() call is treated as a connection error and the client is dropped. This makes connections very unstable.

### DIFF
--- a/hw/dv/dpi/common/tcp_server/tcp_server.c
+++ b/hw/dv/dpi/common/tcp_server/tcp_server.c
@@ -324,9 +324,6 @@ static void *server_create(void *ctx_void) {
     goto err_cleanup_return;
   }
 
-  // Initialise timeout
-  timeout.tv_sec = 0;
-
   // Initialise fd_set
 
   // Start waiting for connection / data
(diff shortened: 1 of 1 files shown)


kwalker27:
heads up @sameo

sameo (approved):

hw/dv/dpi/common/tcp_server/tcp_server.c
```diff
@@ -351,6 +351,13 @@ static void *server_create(void *ctx_void) {
     rv = select(mfd + 1, &read_fds, NULL, NULL, &timeout);
 
     if (rv < 0) {
+      if (errno == EINTR) {
+        // On interrupt we want to retry; according to the man page
+        // the timeout is now undefined.
+        timeout.tv_sec = 0;
```
rswarbrick, line 357: I'm not completely sure about this bit. If I understand the man page correctly, it is just saying that the function might have trashed timeout (see the section called "The timeout").

So I don't think we need to do this change, *but* we should probably set all of `timeout` every time, so just need to move the write to `tv_sec` inside the loop.

rswarbrick (commented):
Sorry for the very slow review. This looks sensible, but I think the second commit could be tweaked slightly.

hw/dv/dpi/common/tcp_server/tcp_server.c
rswarbrick: Duh. I've just re-read your change and now I understand what you're doing! I think it would still be much simpler to move this line to appear just before the write to `tv_usec`. I think the existing code is a bit silly (because `timeout.tv_sec` might get trashed even if the `select` call succeeds), but the fix is very easy.

kwalker27: Thanks for the feedback, I pushed an update.

kwalker27: (oops, pushed again to remove the now-extraneous initial init of `tv_sec`).

Razer6:
Closing this one as we merged them to master via the other PR.

PR #23892, opened 2024-07-02
[opentitantool] Declare additional aliases of HyperDebug pins

jesultra:
HyperDebug has no flexibility on which pins to use for SPI1, SPI2, UARTs, etc.  This CL declares a number of alises, that allows referring to pins such as "UART3_TX", which may be convenient from tests which want to use e.g. `GpioMonitoring` for logic analyzer functionality on pins used for UART, SPI or I2C.


jwnrt (approved):
This LGTM though I haven't checked the pin numbers are correct.
Thanks!

pamaury (approved):

Issue #23888, opened 2024-07-02
[rtl,escalation] Fix rst_ni connection of pwrmgr's receiver


matutem:
### Description

The alert_handler's escalation sender rst_ni is connected to rstmgr_aon_resets.rst_lc_io_div4_n[rstmgr_pkg::Domain0Sel] in top_earlgrey, but the pwrmgr's escalation_receiver is connected to rstmgr_aon_resets.rst_lc_io_div4_n[rstmgr_pkg::DomainAonSel]. Notice they are indexed by a different domain, so they are different nets.

It so happens that the behavior of the two resets (differing in the index) is almost the same, since the reset requests from pwrmgr always drives the same value for both domains. There may be minor cycle differences, but we don't expect this to cause reset divergences that can cause incorrect behavior.

Nevertheless, topearlgrey should be fixed to connect pwrmgr's rst_esc_ni to rstmgr_aon_resets.rst_lc_io_div4_n[rstmgr_pkg::Domain0Sel]. This needs to be done via the hjson files for the top.

PR #23894, opened 2024-07-02
[conn,alert_esc] Check escalation sender/receiver connectivity

### LINKED ISSUE #14887: [chip/prim_esc_receiver] check if prim_esc_receiver is out-of-sync
This discussion comes from PR: https://github.com/lowRISC/opentitan/pull/14663#issuecomment-1244540532

We would like some checking between `prim_esc_receiver` and `alert_handler`, and `pwrmgr` to make sure they are not out-of-sync.

For example - When alert_handler/pwrmgr reset, will `prim_esc_receiver` counters clear to 0?
Currently if the `prim_esc_receiver` did not reset, and we configured the alert_handler to ping after reset, it is hard to find this error.

So the current suggestions are:
- [ ] probing some internal signals, or write assertions.
- [x] have some long FPGA test to re-iterate this sequence, or wait for longer time after wake up, just to catch if any modules are out-of-sync.

@moidx @tjaychen @msfschaffner 

> estimate 12
> remaining 2023-05-04 8

matutem:
The escalation sender and receiver pairs need to connect to the same clock and reset signals, or they can malfunction.

This adds a connectivity configuration for these tests, but it is expected to fail due to the mis-connection of pwrmgr rst_ni described in #23888.

Fixes #14887

### DIFF
--- a/hw/top_earlgrey/formal/conn_csvs/alert_handler_esc.csv
+++ b/hw/top_earlgrey/formal/conn_csvs/alert_handler_esc.csv
@@ -1,0 +1,19 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Run these checks with:
+#  ./util/dvsim/dvsim.py hw/top_earlgrey/formal/chip_conn_cfg.hjson
+
+,NAME,SRC BLOCK,SRC SIGNAL,DEST BLOCK,DEST SIGNAL,,,,,,
+
+# escalation sender and receivers
+CONNECTION, ALERT_HANDLER_PWRMGR_ESC_CLK, top_earlgrey.u_alert_handler, clk_i, top_earlgrey.u_pwrmgr_aon.u_esc_rx, clk_i
+# TODO(lowrisc/opentitan#23888) Enable this check once this issue is fixed.
+# CONNECTION, ALERT_HANDLER_PWRMGR_ESC_RST, top_earlgrey.u_alert_handler, rst_ni, top_earlgrey.u_pwrmgr_aon.u_esc_rx, rst_ni
+CONNECTION, ALERT_HANDLER_LC_CTRL_ESC0_CLK, top_earlgrey.u_alert_handler, clk_i, top_earlgrey.u_lc_ctrl.u_prim_esc_receiver0, clk_i
+CONNECTION, ALERT_HANDLER_LC_CTRL_ESC0_RST, top_earlgrey.u_alert_handler, rst_ni, top_earlgrey.u_lc_ctrl.u_prim_esc_receiver0, rst_ni
+CONNECTION, ALERT_HANDLER_LC_CTRL_ESC1_CLK, top_earlgrey.u_alert_handler, clk_i, top_earlgrey.u_lc_ctrl.u_prim_esc_receiver1, clk_i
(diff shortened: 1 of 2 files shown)


hw/top_earlgrey/formal/conn_csvs/alert_handler_esc.csv
```diff
@@ -0,0 +1,19 @@
...
+# escalation sender and receivers
+CONNECTION, ALERT_HANDLER_PWRMGR_ESC_CLK, top_earlgrey.u_alert_handler, clk_i, top_earlgrey.u_pwrmgr_aon.u_esc_rx, clk_i
+# TODO(lowrisc/opentitan#23888) Enable this check once this issue is fixed.
+# CONNECTION, ALERT_HANDLER_PWRMGR_ESC_RST, top_earlgrey.u_alert_handler, rst_ni, top_earlgrey.u_pwrmgr_aon.u_esc_rx, rst_ni
+CONNECTION, ALERT_HANDLER_LC_CTRL_ESC0_CLK, top_earlgrey.u_alert_handler, clk_i, top_earlgrey.u_lc_ctrl.u_prim_esc_receiver0, clk_i
+CONNECTION, ALERT_HANDLER_LC_CTRL_ESC0_RST, top_earlgrey.u_alert_handler, rst_ni, top_earlgrey.u_lc_ctrl.u_prim_esc_receiver0, rst_ni
+CONNECTION, ALERT_HANDLER_LC_CTRL_ESC1_CLK, top_earlgrey.u_alert_handler, clk_i, top_earlgrey.u_lc_ctrl.u_prim_esc_receiver1, clk_i
+CONNECTION, ALERT_HANDLER_LC_CTRL_ESC1_RST, top_earlgrey.u_alert_handler, rst_ni, top_earlgrey.u_lc_ctrl.u_prim_esc_receiver1, rst_ni
+CONNECTION, ALERT_HANDLER_RV_CORE_IBEX_ESC_CLK, top_earlgrey.u_alert_handler, clk_i, top_earlgrey.u_rv_core_ibex.u_prim_esc_receiver, clk_i
+CONNECTION, ALERT_HANDLER_RV_CORE_IBEX_ESC_CLK, top_earlgrey.u_alert_handler, rst_ni, top_earlgrey.u_rv_core_ibex.u_prim_esc_receiver, rst_ni
```
vogelpi, line 19: This would need to be ALERT_HANDLER_RV_CORE_IBEX_ESC_RST instead of ALERT_HANDLER_RV_CORE_IBEX_ESC_CLK otherwise the tool throws an error due to a duplicated connection.

vogelpi:
@matutem , I've given your PR here a try with JasperGold. If I apply the fix proposed above and fix some outdated connections (will file a separate PR for that), all connections can be proven except for these two connections:
```
conn:PAD_AST
conn:AST_CLK_EXT_IN
```

Based on your message above, I think this is somewhat unexpected. We can also discuss this in more detail next week in one of our sync ups.

matutem:
I fixed the typo you flagged for ALERT_HANDLER_RV_CORE_IBEX_ESC_RST. I only expected a potential failure of ALERT_HANDLER_PWRMGR_ESC_RST.

PAD_AST and AST_CLK_EXT_IN are unexpected, but it seems the connectivity checks have not been run for a while, so those connections could have changed.

vogelpi (approved):

vogelpi:
Thanks @matutem , I'll file a follow-up for the remaining two connection failures.

PR #23896, opened 2024-07-03
[dv, spi_host] clkdiv FCOV fix

antmarzam:
Currently the covergroup sampling the clock divider was ignoring the upper 8-bit of the field.
This PR also includes the upper 8-bit in the sampling


rswarbrick (approved):
This looks sensible to me. Bizarrely, the upper end got dropped by 7be036f3ce1. As far as I can tell, the author just messed it up.

But I notice that before that change there was a coverpoint for `16'hffff`. Should we be restoring that? (instead?)

antmarzam:
> This looks sensible to me. Bizarrely, the upper end got dropped by [7be036f](https://github.com/lowRISC/opentitan/commit/7be036f3ce1c3b99b9eaafaf30207f9e62d916ef). As far as I can tell, the author just messed it up.
> 
> But I notice that before that change there was a coverpoint for `16'hffff`. Should we be restoring that? (instead?)

I don't have a strong preference. It depends of what we care about. IMO my suggestion gives greater coverage granularity over the entire 16-bit range. You could even argue for me to add an upper range bin at `ffff` to be consistent with the lowest bin.
The one in the commit that you mention would most likely make easier to close coverage at the cost of less granularity of stimulus

rswarbrick:
Well, would you mind throwing in a "max" bin as well? I don't think I can see anything that shows we definitely don't want it, and I strongly suspect that it was removed by accident. We can always drop it on purpose in future (with reasoning) if w change our minds!

antmarzam:
Sure -  change is in now!

Issue #23562, opened 2024-06-07
[hmac,dv] Save and Restore improvements


martin-velay:
### Description

This issue follows that one https://github.com/lowRISC/opentitan/issues/21708 where the main parts of the feature has been tested via this PR https://github.com/lowRISC/opentitan/pull/23116. Nevertheless, this PR also indicates that there are still some missing things to be added:
- [ ]  Verify context B with multi-streamed (= different context switching)
- [ ]  Verify intermediate digest when switching

FYI @gdessouky @andreaskurth @ballifatih : and if you feel that anything else as to be tested, please rise it here.

vogelpi:
Thanks for creating the issue @martin-velay .
> Verify context B with multi-streamed (= different context switching)

Right now, only context A is verified meaning we just check that context A can be saved and restored which is fine I believe.

> Verify intermediate digest when switching

This would allow detecting when exactly things get out of sync in case of issues. It would ease debugging but I think as long as the design passes verification, it's not a very high priority to allow more fine grained debugging.

So for me, both these items are not too important but rather nice to have. It's fine to defer this to M5 in my view.

gdessouky:
> [ ] Verify context B with multi-streamed (= different context switching)

And for sufficient assurance for M4, we cover this in the driver-based TLT for now.

andreaskurth:
I think we would do this as part of HMAC V3 work -- @martin-velay WDYT?

martin-velay:
@andreaskurth: OK

martin-velay:
This is covered at top level but it would be better to also cover this at block level as part of M7

martin-velay:
Doesn't required for D3-V3 signoffs as already tested in top level test with the SW and on FPGA

PR #23901, opened 2024-07-03
[hmac,dv] Test msg_length_upper

martin-velay:
- Take the advantage of Save and Restore feature to load a hypothetical context with a large msg_length to test the msg_length_upper register.
- Split the S&R into sub-tasks to make the code a bit easier to read.

Note: code should be slightly change after taking care of this issue https://github.com/lowRISC/opentitan/issues/23562 as the msg_length registers are checked in the vseq but it shouldn't be once the SCB will be able to check the context B.


martin-velay:
@gdessouky 
As discussed in the chat together, can you try to run this at your end and see if you understand what's wrong here.

With the current test as it is, you can run `hmac_smoke` test and I guess it should fail.

rswarbrick (approved):
This looks really good, thanks.

PR #23905, opened 2024-07-03
[hsmtool] fix ecdsa import/export commands

mvaquez:
Following the PKCS#11 specifications, the CKA_EC_POINT should be the DER-encoding of ANSI X9.62 ECPoint value, this was not the case.

Using softhsm it was possible to `import` a public key and `export` it back, but it wasn't possible to `verify` a signature using the imported public key. Moreover it wasn't possible to `export` a public key generated using the `generate` command.

This PR fixes these issues.


cfrantz (approved):
Thanks for the fix!

PR #23906, opened 2024-07-03
Spi host timeout fixes

antmarzam:
Modifying the csr_spinwait timeout based on the clock divider, but also constraining large clock divider values to be configured in `hw/ip/spi_host/dv/env/seq_lib/spi_host_upper_range_clkdiv_vseq.sv ` which would run 10 times less frequently in the overnight regression

Note: This is still WIP - I'm profiling the thoroughness of the new VSEQ and seeing whether any further constraining would be helpful, plus there is still another outstanding timeout for status.rxempty I'd like to fix in the same PR


hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -73,10 +73,16 @@ class spi_host_base_vseq extends cip_base_vseq #(
         spi_config_regs.csnidle[i] inside {[cfg.seq_cfg.host_spi_min_csn_latency :
                                           cfg.seq_cfg.host_spi_max_csn_latency]};
       }
-      foreach (spi_config_regs.clkdiv[i]) {
-        spi_config_regs.clkdiv[i] inside {[cfg.seq_cfg.host_spi_min_clkdiv :
-                                         cfg.seq_cfg.host_spi_max_clkdiv]};
-      }
+  }
+
+  // Separate constraint to allow easy overide in child sequences
```
rswarbrick, line 78: Nit: Two r's in override :-)

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -97,6 +103,20 @@ class spi_host_base_vseq extends cip_base_vseq #(
       };
   }
 
+  function void post_randomize();
+    super.post_randomize();
+    // We currently support 1 chip-select line only
+    case(spi_config_regs.clkdiv[0]) inside
+      // After randomization we set a different timeout based on the clock divider
+      [cfg.seq_cfg.host_spi_min_clkdiv : cfg.seq_cfg.host_spi_lower_middle_clkdiv]: ; // no change
```
rswarbrick, line 111: That's a good idea. Maybe worth adding a default: error case at the end?

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -109,6 +129,26 @@ class spi_host_base_vseq extends cip_base_vseq #(
     super.pre_start();
   endtask : pre_start
 
+  // csr_spinwait wrapper to allow finer control of task fields
+  virtual task automatic csr_spinwait(input     uvm_object          ptr,
```
rswarbrick, line 133: How does this override get called? (I don't think the task is defined in the base class?)

hw/ip/spi_host/dv/env/seq_lib/spi_host_speed_vseq.sv
```diff
@@ -33,10 +33,16 @@ class spi_host_speed_vseq extends spi_host_smoke_vseq;
...
       }
-      foreach (spi_config_regs.clkdiv[i]) {
-        spi_config_regs.clkdiv[i] inside {[cfg.seq_cfg.host_spi_min_clkdiv :
-                                         cfg.seq_cfg.host_spi_max_clkdiv]};
-      }
+  }
+
+  constraint spi_config_regs_clkdiv_c {
+    foreach (spi_config_regs.clkdiv[i]) {
+      // CLKDIV randomised not in the whole range since there's a dedicated VSEQ: seq_name
```
rswarbrick, line 40: Are we missing a name here?

hw/ip/spi_host/dv/env/spi_host_env_cfg.sv
```diff
@@ -27,6 +27,10 @@ class spi_host_env_cfg extends cip_base_env_cfg #(.RAL_T(spi_host_reg_block));
   // bumber of address bytes
   rand int    num_cmd_bytes;
 
+  // TB heavily relies on csr_spinwait, which causes timeouts when the clock divider is FF-ish or
+  // higher values
+  uint csr_spinwait_timeout_ns = csr_utils_pkg::default_spinwait_timeout_ns;
```
rswarbrick, line 32: If we're not going to configure it here, I'd probably suggest using the csr_utils_pkg item directly instead of copying it to a local flag: it might end up a bit more obvious what's going on?

hw/ip/spi_host/dv/env/seq_lib/spi_host_upper_range_clkdiv_vseq.sv
```diff
@@ -0,0 +1,35 @@
...
+
+
+  constraint spi_config_regs_clkdiv_c {
+    foreach (spi_config_regs.clkdiv[i]) {
+      // CLKDIV randomised not in the whole range since there's a dedicated VSEQ: seq_name
+      // which uses the upper range of clock divider values - this way we won't have super long
+      // tests when running this VSEQ
+      spi_config_regs.clkdiv[i] inside {[cfg.seq_cfg.host_spi_middle_clkdiv :
+                                         cfg.seq_cfg.host_spi_max_clkdiv]};
+    }
```
rswarbrick, line 18: Would it make sense to give a high weight to the maximum value (maybe 50% of the time)? That way, we'll see the "exciting value" more often, and the test runtimes won't be all that much bigger.

rswarbrick (commented):
This looks sensible to me, and thanks for introducing me to the "case inside" syntax: I've never used that! A couple of potential tweaks that might make it easier to read, but this looks good otherwise.

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
antmarzam: Done!

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
antmarzam: Done!

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
antmarzam: We discussed in person - but responding here too to be able to track the reasoning if needed. 

This is just called in the scope of `spi_host_base_vseq` and any sequence extending from this. 

In our chat we decided it was better to have a wrapper like this just so if we were using different components VSEQs (maybe in a multi-block TB) we could have a different timeout for the csr_spinwait calls in spi_host VSEQs VS any oher sequences. 

I've also added a comment to make it clearer

hw/ip/spi_host/dv/env/seq_lib/spi_host_speed_vseq.sv
antmarzam: good spot! done now - it was meant to have the name of the new seq (spi_host_upper_range_clkdiv_vseq)

hw/ip/spi_host/dv/env/spi_host_env_cfg.sv
antmarzam: We discussed it in person, but also commenting here for context.

That's just the default value, but then it get's set to a greater value depending on the clock divider value in the spi_host_base_vseq

hw/ip/spi_host/dv/env/seq_lib/spi_host_upper_range_clkdiv_vseq.sv
antmarzam: Thanks for the suggestion. I'm going to do that, but a lower weight (maybe 25-30%?) Just so we use the max sometime and get coverage for it. But we also take values in the range which would be useful to get more FCOV for that 0xFF to 0xFFFE range

rswarbrick: Looks good to me. I'm continually amazed by how much can be expressed in the constraint language :-)

rswarbrick (approved):
This looks really good, thanks!

rswarbrick:
Merging despite the broken ROM tests, since this is a DV-only change and definitely can't have broken them.

PR #23907, opened 2024-07-03
[otbn,dv] Try to improve model's support for RMA requests

rswarbrick:
There are some detailed notes in the commit messages, but this is all about trying to get the `otbn_escalate` test to work properly.


hcallahan-lowrisc (approved):
This seems sensible to me, thanks

Issue #23903, opened 2024-07-03
[otbn,rtl] Glitch in the STATUS register on way to LOCKED after an RMA


rswarbrick:
### Description

If an RMA request comes in when OTBN is idle, the STATUS register spends a cycle LOCKED (`0xff`), before jumping to secure wipe (`0x04`) for a while before it finally goes back to LOCKED.

It happens as follows:

- When OTBN is idle, `otbn_controller` will be in state `OtbnStateHalt` and `otbn_start_stop_control` will be in state `OtbnStartStopStateHalt`.
- Suppose an RMA request comes in.
- This appears to the controller as as `lc_rma_req_i` and the controller jumps to `OtbnStateLocked`.
- The controller exposes a `locking_o` signal, which essentially controls forces the `STATUS` register to be LOCKED. This is computed as follows:
```systemverilog
    assign locking_o = (state_d == OtbnStateLocked) & (~(secure_wipe_running_i | secure_wipe_req_o) |
                                                       urnd_reseed_err_i | mubi_err_d);
```
- The idea is that we want `STATUS` to become LOCKED once the secure wipe has finished running.
- Unfortunately, `secure_wipe_running_i` lags behind `state_d` here! It is computed by otbn_start_stop_control, where it is the value of `secure_wipe_running_q`.
- There's a cycle where state_d is `OtbnStateLocked` and `secure_wipe_running_d` is true, but `secure_wipe_running_q` (the registered value) hasn't yet gone high.
- The end result is that `STATUS` is LOCKED for a single cycle and then jumps to the secure wipe state for a while before going back to LOCKED. This contradicts the idea that the LOCKED status is terminal.

This is not a particularly critical bug, as it only exposes itself as a weird transient value in a status register. And I only found it because some of the DV code looks at that register every cycle. But I think it's probably worth fixing in a future version.

Here are some waves showing the problem. The final row is the value of `status_q` in `otbn.sv`, which is exposed directly as the contents of the STATUS register.
![image](https://github.com/lowRISC/opentitan/assets/104845/aa2ac148-9f05-43d4-97ff-bb403434638f)

vogelpi:
Thanks for flagging this @rswarbrick . I think your assessment that this isn't a critical bug is absolutely correct.. The check whether RMA entry is successful or not isn't done by software by checking the OTBN status register but rather by LC_CTRL which sends out the RMA requests and collects the acknowledgements from Flash and OTBN. After initiating the transition, the process is quite hardware driven.

[...]

PR #23908, opened 2024-07-03
[otbn,dv] Weaken some checks in DV code to allow RMA operations

rswarbrick:
This is to allow RMA operations, as found with the `otbn_escalate` test.

Commits 1 and 3 are working around the behaviour reported in #23903 (and leave TODO notes pointing back at it). Commit 2 fixes a genuine DV bug (we wrote the environment before RMA requests were in the design)


hcallahan-lowrisc (approved):
These changes seem sensible to me given the TODO item. Thanks :)

PR #23909, opened 2024-07-03
[dv, csr_spinwait] debug printout

antmarzam:
Adding printout to ease tracking csr_spinwait calls. When we get the timeout error we can now see

### DIFF
--- a/hw/dv/sv/csr_utils/csr_utils_pkg.sv
+++ b/hw/dv/sv/csr_utils/csr_utils_pkg.sv
@@ -542,18 +542,25 @@ package csr_utils_pkg;
                               input compare_op_e        compare_op = CompareOpEq,
                               input bit                 backdoor = 0,
                               input uvm_verbosity       verbosity = UVM_HIGH);
+    static int                      count;
+    count++;
+    `uvm_info($sformatf("%m()"), $sformatf(
+                "- (call_count=%0d, backdoor=%0d, exp_data=%0d, ptr=%s)",
+                count,backdoor,exp_data, ptr.get_name()), verbosity)
     fork
       begin : isolation_fork
         csr_field_t     csr_or_fld;
         uvm_reg_data_t  read_data;
         string          msg_id = {csr_utils_pkg::msg_id, "::csr_spinwait"};
+        automatic int lcount = count;
 
         csr_or_fld = decode_csr_or_field(ptr);
         if (backdoor && spinwait_delay_ns == 0) spinwait_delay_ns = 1;
         fork
           while (!under_reset) begin
             if (spinwait_delay_ns) #(spinwait_delay_ns * 1ns);
-            `uvm_info("csr_utils_pkg", "In csr_spinwait", verbosity)
+            `uvm_info("csr_utils_pkg", $sformatf("In csr_spinwait - call_count = %0d",lcount),
+                      verbosity)
             csr_rd(.ptr(ptr), .value(read_data), .check(check), .path(path),
                    .blocking(1), .map(map), .user_ftdr(user_ftdr), .backdoor(backdoor));
             `uvm_info(msg_id, $sformatf("ptr %0s == 0x%0h",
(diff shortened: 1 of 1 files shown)


hw/dv/sv/csr_utils/csr_utils_pkg.sv
```diff
@@ -573,8 +580,11 @@ package csr_utils_pkg;
             endcase
           end
           begin
-            `DV_WAIT_TIMEOUT(timeout_ns, msg_id, $sformatf("timeout %0s (addr=0x%0h) == 0x%0h",
-                ptr.get_full_name(), csr_or_fld.csr.get_address(), exp_data))
+            automatic int lcount = count;
+            `DV_WAIT_TIMEOUT(timeout_ns, msg_id,
+                             $sformatf("timeout %0s (addr=0x%0h) == 0x%0h - call_count = %0d",
```
rswarbrick, line 585: I always found this message a bit confusing because of the == bit. While you're changing this, would you mind turning it into something like this?
```
            `DV_WAIT_TIMEOUT(timeout_ns, msg_id,
                             $sformatf("timeout %0s (addr=0x%0h, exp_data=0x%0h, call_count=%0d)",
```

antmarzam: This is now in - I also added the comparison carried out for extra clarity. 
Had to do a bit of gymnastics due to line-length limits

hw/dv/sv/csr_utils/csr_utils_pkg.sv
```diff
@@ -542,18 +542,25 @@ package csr_utils_pkg;
                               input compare_op_e        compare_op = CompareOpEq,
                               input bit                 backdoor = 0,
                               input uvm_verbosity       verbosity = UVM_HIGH);
+    static int                      count;
+    count++;
+    `uvm_info($sformatf("%m"), $sformatf(
+                "csr_spinwait - call_count = %0d - backdoor=%0d - exp_data = %0d - ptr = %s",
```
hcallahan-lowrisc, line 548: One tiny suggestion, I usually add the parens when I'm putting the task/fun name in the log message.
e.g
```
csr_spinwait() - ...
```
This helps me quickly realize it's the name of a subroutine :)

hw/dv/sv/csr_utils/csr_utils_pkg.sv
```diff
@@ -542,18 +542,25 @@ package csr_utils_pkg;
...
+                count,backdoor,exp_data, ptr.get_name()), UVM_DEBUG)
```
hcallahan-lowrisc, line 549: Curious: Why `UVM_DEBUG` instead of `verbosity` here?

hcallahan-lowrisc (approved):
LGTM! Thanks for hopefully making debug easier in the future!

hw/dv/sv/csr_utils/csr_utils_pkg.sv
antmarzam: good spot! I missed the verbosity argument :/ 
Will push again!

hw/dv/sv/csr_utils/csr_utils_pkg.sv
antmarzam: I'm happy to add it. At the same time, I just realised the hardcoded "csr_spinwait" is kinda pointless since I added the "%m" already. Going to run a test and see how the "%m". Will paste the printout below.

Also for consistency, I'll probably print the fields separated by commas to be consistent with the fatal printout when the timer expires

antmarzam: OK - so different printouts would be:

1 - Original:
`[csr_utils_pkg::csr_spinwait] csr_spinwait - call_count = 1 - backdoor=0 - exp_data = 1 - ptr = ready`
2 - Adding parenthesis to csr_spinwait:
`[csr_utils_pkg::csr_spinwait] csr_spinwait() - (call_count = 1, backdoor=0, exp_data = 1, ptr = ready)`
3 - Not using %m, but hardcoding csr_spinwait():
`[csr_spinwait()] (call_count = 1, backdoor=0, exp_data = 1, ptr = ready)`
4 - Using %m and removing the hardcoded csr_spinwait:
`[csr_utils_pkg::csr_spinwait] - (call_count = 1, backdoor=0, exp_data = 1, ptr = ready)`

I don't have a strong opinion TBH. If something I guess I'd suggest using %m (option-4) just in case we ever change the name of the task so we minimise the amount of code which needs changing, but not a biggie regardless

hcallahan-lowrisc: Option 4 or 5 LGTM!

antmarzam: Cool - changes should be in now. I went for no-5:
`[csr_utils_pkg::csr_spinwait()] - (call_count=1, backdoor=0, exp_data=1, ptr=ready)`

PR #23910, opened 2024-07-03
[dv, spi_host] Using the factory to create objects

antmarzam:
The TB was barely using the factory to create objects. In addition there were a few places where a seq_item was created to them be overriden by popping an item from a queue.

### DIFF
--- a/hw/ip/spi_host/dv/env/seq_lib/spi_host_tx_rx_vseq.sv
+++ b/hw/ip/spi_host/dv/env/seq_lib/spi_host_tx_rx_vseq.sv
@@ -64,7 +64,7 @@ class spi_host_tx_rx_vseq extends spi_host_base_vseq;
 
   // Generate the TL-writes needed for the DUT to send all segments of a "spi_transaction_item" transaction.
   virtual task send_trans(spi_transaction_item trans, bit wait_ready = 1'b1);
-    spi_segment_item segment = new();
+    spi_segment_item segment;
     while (trans.segments.size() > 0) begin
       if (wait_ready) wait_ready_for_command();
       // lock fifo to this seq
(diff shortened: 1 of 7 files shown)


hcallahan-lowrisc (approved):
Great! Thanks for the cleanup ;)

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -60,9 +60,9 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   function void build_phase(uvm_phase phase);
     super.build_phase(phase);
     plain_data_fifo  = new("plain_data_fifo", this);
```
marnovandermaas, line 62: Can we use a similar construction to build this object?

hw/ip/spi_host/dv/env/seq_lib/spi_host_tx_rx_vseq.sv
```diff
@@ -64,7 +64,7 @@ class spi_host_tx_rx_vseq extends spi_host_base_vseq;
 
   // Generate the TL-writes needed for the DUT to send all segments of a "spi_transaction_item" transaction.
   virtual task send_trans(spi_transaction_item trans, bit wait_ready = 1'b1);
-    spi_segment_item segment = new();
+    spi_segment_item segment;
```
marnovandermaas, line 67: Nice catch!

marnovandermaas (approved):
Just a minor question but the current changes look good to me so approving.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: That's a great question. I had to go an look up why this is not done in the UVM. 
I presume the original reason was that this particular component is not expected to ever be overridden

I tried to create it via the factory but I was getting an error. The reason was that the component is not registered with the factory, hence can't be created that way...

Issue #14825, opened 2022-09-08
[i2c/dv] issues with i2c_agent


tjaychen:
This issue attempts to describe all the known issues with the existing `i2c_agent` and how they should be fixed. 

The `i2c_monitor` directly interprets the line content and creates response items (acknowledgement, read data) instead of just passively observing the line. 
The `i2c_driver`, instead of just receiving data and driving it onto the pins, has a method of directly creating randomized data

To address these issues, the following should be done:
- [ ] the "interpretation" done by the monitor should be passed onto the `base_seq` if possible. The base sequence can then be extended to emulate a memory model for more advanced kinds of testing, in addition to the "random" response it does now.  This also means the base sequence should become responsible for creating acknowledgements.
- [ ] the logic inside the i2c driver that creates return data should be removed
- [ ] There are many req_analysis_fifo's inside i2c agent, and those probably need to be consolidated (need to look at the interaction between monitor and scoreboard).

tjaychen:
@sriyerg 
@jdonjdon 

let me know if there are others you noticed. I admittedly didn't really look at the scoreboard and focused mostly on the driver / monitor / base_seq

tjaychen:
@jdonjdon i think we can postpone this to M3.
Because this issue is really about all the things in the i2c agent that need to be cleaned up, but perhaps it cannot be done in time for M2.

Let me know what you think.

jdonjdon:
> @jdonjdon i think we can postpone this to M3. Because this issue is really about all the things in the i2c agent that need to be cleaned up, but perhaps it cannot be done in time for M2.
> 
> Let me know what you think.

Agree

tjaychen:
actually @jdonjdon were some of these items already addressed? do you want to update the description to be more accurate?

jdonjdon:
> actually @jdonjdon were some of these items already addressed? do you want to update the description to be more accurate?

This issues are only applied to "host mode' rtl test bench. Unfortunately none of these are addressed during recent "target mode" rtl verification.

tjaychen:
do you think this is even worth fixing in V3?

jdonjdon:
I am neutral at this point. Can we revisit after v2?

hcallahan-lowrisc:
This is a V3 item really, hence moved to P4 and removed the PROD Candidate lable.

Issue #11, opened 2019-08-15
C/C++ style guide: Allow exceptions for manual whitespace formatting?


imphil:
In lowRISC/stwg-base#167 we initially discussed, and in lowRISC/stwg-base#862 we merged the C/C++ style guide, which is based on the Google C++ style guide. A rendered version is available here: https://bubble.servers.lowrisc.org/doc/rm/CCppCodingStyle.html. 

The style guide is enforced through [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html), a automatic code formatter. No exceptions to the formatting as produced by `clang-format` are allowed.

Using `clang-format` is uncontroversial. What is controversial, however, is the question of exceptions, more specifically: **Should the style guide allow exceptions for manual whitespace formatting?** Note that the discussion here is *only* about whitespace formatting, it's not about  other potential exceptions to the style guide.

`clang-format` can be told to ignore certain code blocks by using `clang-format off/on` comments around the affected code block ([documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code)). There have been no objections so far to require the use of these comments around manually indented code blocks.

In the discussion up to now many people have voiced their opinion. I'm trying to summarize the points as I understood them below.

Arguments against whitespace exceptions:

* Allowing any kind of exception opens the flood gates for more and more exceptions in the future.
* Instead of allowing exceptions, `clang-format` should be improved to give better results.
* Especially on auto-generated header files (like the `*_regs.h` files that regtool generates): "I don't tend to care about how pretty autogenerated headers look, since in general you don't need to look at them. I see even less point in doing this just for autogenerated stuff." (Ron)

Arguments for whitespace exceptions:

* In some pieces of code, especially in lists of registers or struct fields and when writing macros, manually adjusted whitespace helps readability. A full diff of changes produced by our initial run of clang-format is available [here](458be1593bbd502e671efbe9b889d17b581b6355) and [here](https://github.com/lowRISC/stwg-base/commit/5c510fdcd081c32441f8999473ef3c7e8e2fdc09), here are some examples:

   1. Manually formatted:
       https://github.com/lowRISC/stwg-base/blob/5c510fdcd081c32441f8999473ef3c7e8e2fdc09/sw/lib/usb_controlep.h#L46-L58

       clang-format:
[...]

Issue #10, opened 2019-08-16
[doc] document our primitives


sjgitty:
- [ ] add mention in Comportability the recommendation to follow utilization guides for existing IP primitives
  - [ ] pointer to a yet-to-be-written technical specification for all primitives (see below)
  - [ ] list requirements for primitives (RTL, lint, DV, etc)
  - [ ] suggest the protocol to add a primitive to our library (rationale as to why existing ones are insufficient)
- [ ] create said technical specification for all primitives; goal: concise but thorough, with rationale;
  - [ ] Question: one master doc, or one per? (leaning former)
  - [ ] Question: place under hw/ip/prim/doc/prim.md? or hw/ip/prim/README.md?
  - [ ] Sections, similar to other technical specifications: feature set, overview, block diagram, interfaces, current level of verification (mini D/V status), CDC implications (a bit redundant with hw/formal/README.md section on CDC that is coming), regtool recommendations, software recommendations

PR #23911, opened 2024-07-03
[i2c,dv] I2C transfer level TLM and capture

hcallahan-lowrisc:
> 12 commits will be / have been removed from the beginning of this PR, and were moved to #23919. Some review comments left here may refer to those commits, in case things look disconnected!

This very large PR fixes some architectural deficiencies with the I2C DV environment, with the goal of allowing easier modelling and checking of transfers which exercise the reactive components of the I2C protocol (ie. NACKing).

All commits except the last one are generally cleanups and uninteresting, and go towards supporting the final commit which contains the bulk of changes which need to 'touch-the-world' across the testbench environment and agent. Some of these commits could possibly be spread out into a different PR, but this whole thing came together a bit organically, and I don't think it's worth significantly more of my time to rework the commit history any further into smaller atomic blocks.

The final commit is the interesting one, and due to the low-level testbench changes, I couldn't easily find a way to break it into atomic pieces, hence it is very large. I have added a detailed commit message which explains some of the rationale. I've copied it to the bottom of this PR.

While the functionality isn't identical, this change should leave the block-level DV at or near parity with the previous results. I've linked some data below to show this.

\<INSERT REGRESSION COMPARISON HERE\>
 
> This large commit fixes a number of fundamental flaws and confusing design
> decisions around the way we model TLM transactions using the i2c_item.
>
> The major flaw this change initally set out to fix was that scoreboard
> expectations were being formed by passing the generated stimulus from the
> virtual sequences directly to the model/scoreboard. This is against UVM design
> patterns, and severely limits the ability to predict behaviour when the effect
> of the driver are fundamentally bypassed and unpredictable. In I2C, the receiver
> in any transaction can alter the progression of the traffic by choosing to ACK
> or to stretch SCL. In our I2C environment, where both endpoints are
> operated/tested as both Controllers and Targets, this means any reactivity
> in the Agent cannot be predictable, and sequences are therefore reduced to only
> a very small subset of behaviour. This problem was partially circumvented by a
> slew of configuration variables used to hold bits of state and signal events to other
> components, but this is unclean and inconvenient.
>
> To fix this problem, another intermediate issue needed to be addressed. The
> i2c_item was grossly overloaded in the way it was used to represent different
> things to different people across the testbench and agent. For read transfers,
> the seq_item was used to capture the entire transfer (data_q[$] held all the
> bytes of data), but for writes we would capture one item per entry in the
> ACQFIFO CSR to signal an I2C transaction was taking place. Different fields in
> the item were used for each use-case, and some were re-used with different
> meanings for each case.
> > There is also a third use of the i2c_item, again using different fields (with
> > some overlap), which is to control the i2c_driver in discrete steps depending
> > on the value of a control enum variable 'drv_type'. This use case remains, but
> > should definitely be removed in the future.
> To model at a transfer level, new fields 'rstart_front' and 'rstart_back' were
> added to the item, which indicate if the start or the end of the transaction was
> signalled using RSTARTS instead of STARTs and STOPs. The previous 'rstart' field
> would be ambiguous when used for modelling at a transfer-level.
>
> Additionally, this commit also importantly changes the way mis-addressed
> transfers are captured and checked. Previously, the i2c_agent contained copies
> of the DUT's configured Target-Mode addresses, and used these to change the way
> it captured observations and generated stimulus based on if the address given
> was known ahead of time to match the DUT. This is clearly a badly-designed
> dependence between the modules, and by removing this pre-filtering of
> transaction items, we now generate and check items which do not match the
> targets address. These items are effectively the I2C-Controller shouting into
> the void, which is a valid i2c transfer in terms of signalling, but in a single
> controller, single target system, a wholly uninteresting one.
>
> In short, this commit achieves the following:
> - Reworks the i2c_item to model full 'i2c_transfers', and updates the
>   i2c_monitor to generate seq_items at this level of abstraction, for both reads
>   and writes.
> - Change the reference_model/scoreboard to gather observations in this new
>   format from the monitor only (connections to the vseqr are now removed)
> - The i2c_monitor now generates valid transfer items for mis-addressed stimulus,
>   easily observed by checking if the ITEM_T.addr_ack field is equal to i2c_pkg::NACK.
> - Replaces the stimulus generation routines in the i2c_base_vseq with new,
>   concise versions that no longer need to generate transaction items to send to
>   the model/scoreboard. The previous code here was verbose, unclear and
>   difficult to modify.
>   - create_txn()/fetch_txn() now replaced by a single call to
>   generate_agent_controller_stimulus(), which makes better use of types,
>   targeted subroutines and randomization to achieve the same goal in a clear and
>   simpler way.
> - Renames variables across the env/agent cfgs, moved variables from the
>   class-scope into task/function scope where possible, and used user-defined
>   types where appropriate to aid readability.

- TODO : Fixup the hrst_vseq


hw/dv/sv/i2c_agent/seq_lib/i2c_target_base_seq.sv
```diff
@@ -7,18 +7,28 @@ class i2c_target_base_seq extends i2c_base_seq;
...
+    fork begin : iso_fork
+      fork
+        // Drive all items in the 'req_q'
+        while (req_q.size() > 0) begin
+          req = req_q.pop_front();
+          start_item(req);
+          finish_item(req);
+        end
+        // This process ends the stimulus generation if seq_stop() is called.
+        wait(stop);
```
rswarbrick, line 28: I know that this isn't new behaviour, but would it make sense to leave a note somewhere pointing out that we'll have trashed the request queue if this happens?

hw/dv/sv/i2c_agent/i2c_driver.sv
```diff
@@ -49,8 +49,11 @@ class i2c_driver extends dv_base_driver #(i2c_item, i2c_agent_cfg);
         begin: iso_fork
           fork
             begin
-              if (cfg.if_mode == Device) drive_device_item(req);
-              else drive_host_item(req);
+              case (cfg.if_mode)
+                Host: drive_host_item(req);
+                Device: drive_device_item(req);
+                default:;
```
rswarbrick, line 55: I'd probably stick a fatal error here: the enum is only supposed to have Host/Device as options.

hw/ip/i2c/dv/env/i2c_scoreboard.sv
```diff
@@ -573,7 +573,10 @@ class i2c_scoreboard extends cip_base_scoreboard #(
   // - start
   // - stop
   // - wdata (for non-rstart/stop items)
+  //
+  // #TODO : Refactor testbench such that we can compare full seq_items using A.compare(B).
```
rswarbrick, line 645: Would you mind opening an issue and tagging this with its number? Similarly for the TODO item on line 601.

rswarbrick (commented):
Notes on commit #10

rswarbrick (commented):
Minor note on commit #11.

rswarbrick:
So far, I've gone through all but the last commit. Some minor notes. It's probably worth factoring those commits into a parallel PR and just get those changes merged, without gating them on review of the final commit.

rswarbrick (commented):
Notes on the list commit, up to the end of i2c_item.sv:

hw/dv/sv/i2c_agent/seq_lib/i2c_target_base_seq.sv
```diff
@@ -7,18 +7,28 @@ class i2c_target_base_seq extends i2c_base_seq;
...
-    `DV_SPINWAIT_EXIT(wait (req_q.size() > 0);
-                      while (req_q.size() > 0) begin
-                        req = req_q.pop_front();
-                        start_item(req);
-                        finish_item(req);
-                      end
-                      , wait (stop);)
+    // A sequence of i2c_items are randomized in vseq and fed to req_q here.
+    // This isn't ideal, and hopefully will be rethough in the future.
+    // #TODO Randomize in the sequence, rather than the vseq. Or, compose a
```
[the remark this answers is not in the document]
hcallahan-lowrisc: I linked this to #14825, which is a bit of a bucket list for i2c_agent issues.

hw/ip/i2c/dv/env/seq_lib/i2c_host_fifo_reset_fmt_vseq.sv
```diff
@@ -9,47 +9,55 @@ class i2c_host_fifo_reset_fmt_vseq extends i2c_rx_tx_vseq;
...
+        `uvm_create_obj(i2c_item, fmt_item);
+        fmt_item.fbyte_c.constraint_mode(0);
+
+        `DV_CHECK_RANDOMIZE_WITH_FATAL(fmt_item,
+          fbyte == wr_data[j-1];
+          start == 1'b0;
+          read  == 1'b0;
+          // Randomize so at least one of format bits is non-zero.
+          // This ensures we have something at least to push into the fmt_fifo
+          $countones({|fbyte, rcont, nakok}) == 1;
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Indeed, good spot! Thanks :)

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -132,32 +155,76 @@ class i2c_monitor extends dv_base_monitor #(
...
-      `uvm_info(`gfn, $sformatf("Sending following full item to scb now:\n%s",
-                                full_item.sprint()), UVM_DEBUG)
+    // Capture the transfer
+    begin
+      target_address_thread();
+      if (mon_dut_item.bus_op == BusOpRead) target_read_thread();
+      else                                  target_write_thread();
+    end
+    // If the previous transfer ended via STOP (as opposed to RSTART), signal this to the driver.
+    if (mon_dut_item.stop) cfg.got_stop = 1;
```
rswarbrick, line 199: I know this isn't new to this commit, but it's a bit sad to be sending information from the monitor to the driver like this. I wonder whether it makes sense to create an issue tracking the possibility of breaking the link.

hw/dv/sv/i2c_agent/i2c_driver.sv
hcallahan-lowrisc: Done!

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -132,32 +155,76 @@ class i2c_monitor extends dv_base_monitor #(
...
-    if (cfg.vif.rst_ni && full_item.stop && full_item.start) begin
-      if (full_item.bus_op == BusOpRead) rd_item_port.write(full_item);
-      else                               wr_item_port.write(full_item);
-      `uvm_info(`gfn, $sformatf("Sending following full item to scb now:\n%s",
-                                full_item.sprint()), UVM_DEBUG)
+    // Capture the transfer
+    begin
+      target_address_thread();
+      if (mon_dut_item.bus_op == BusOpRead) target_read_thread();
+      else                                  target_write_thread();
```
rswarbrick, line 196: Based on the conversation we just had in person(!), would it make sense to turn this into a `case` statement?

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -132,32 +155,76 @@ class i2c_monitor extends dv_base_monitor #(
...
+
+    // The transaction has completed, send the completed item to the scoreboard.
+    begin
+      i2c_item full_item;
+      `downcast(full_item, mon_dut_item.clone());
+      case (full_item.bus_op)
+        BusOpRead:  begin
+          target_mode_rd_item_port.write(full_item);
+          // `uvm_info(`gfn, $sformatf("Wrote following item to target_mode_rd_item_port\n%s",
+          //                           full_item.sprint()), UVM_HIGH)
```
rswarbrick, line 209: Drop the commented code here?

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -267,21 +341,30 @@ class i2c_monitor extends dv_base_monitor #(
...
               mon_dut_item.data_q.push_back(mon_data);
               `uvm_info(`gfn, $sformatf("target_write_thread() data %2x num_data:%0d",
-                                        mon_data, mon_dut_item.num_data), UVM_HIGH)
-              // send device ack to host write
-              mon_dut_item.wdata = mon_data;
+                                        mon_data, mon_dut_item.num_data), UVM_FULL)
+
+              // Send data component of this write to the 'req_analysis_port', for consumption by
+              // reactive-agent sequences.
+              mon_dut_item.wdata = mon_data; // TODO Remove??? I don't think this is used...
```
rswarbrick, line 348: It's a very predictable comment from me but... maybe remove this? :-)

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -353,176 +436,223 @@ class i2c_monitor extends dv_base_monitor #(
...
-    cfg.valid_addr = 0;
-    if (mon_dut_item.stop ||
-        (!mon_dut_item.stop && !mon_dut_item.start && !mon_dut_item.rstart)) begin
+    `uvm_create_obj(i2c_item, mon_dut_item);
+
+    // Decode the start of the next transfer. As we are calling this task in a loop, this may be
+    // the start of a new transaction, or the start of new transfer following a repeated-start
+    // condition.
+
+    if ( prev_item != null && prev_item.rstart_back) begin
```
rswarbrick, line 449: Nit: Stray space after "("

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -353,176 +436,223 @@ class i2c_monitor extends dv_base_monitor #(
...
-    controller_address_thread(skip_the_loop);
-    if (skip_the_loop) begin
-      `uvm_info(`gfn, "Skipping the loop now.", UVM_MEDIUM)
-      return;
+    // Capture the transfer
+    // - If an external reset is requested, return early without pushing a sequence item.
+    begin
+      bit skip;
+      controller_address_thread(skip);
+      if (skip) begin
```
rswarbrick, line 478: Maybe this variable name could be improved. Based on the comment below, maybe "seen_mid_transfer_reset" or something?

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -353,176 +436,223 @@ class i2c_monitor extends dv_base_monitor #(
...
-    full_item.stop = 1'b1;
-    if (mon_dut_item.bus_op == BusOpRead) begin
-      full_item.read = 1;
-      analysis_port.write(full_item);
+    // If the previous transfer ended via STOP (as opposed to RSTART), signal this to the driver.
+    if (mon_dut_item.stop) cfg.got_stop = 1;
+
+    `uvm_info(`gfn, "controller_collect_thread() got end of transfer, pushing item now.", UVM_HIGH)
+
+    // The transaction has completed, send the completed item to the scoreboard.
```
rswarbrick, line 490: Nit: I think you probably want a ";" or "." instead of a comma here.

rswarbrick (commented):
Some notes as far as the end of `controller_collect_thread` in the monitor in the last commit.

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -353,176 +436,223 @@ class i2c_monitor extends dv_base_monitor #(
...
-                                    mon_dut_item.addr), UVM_MEDIUM)
+          dut_valid_addr = cfg.is_target_addr(mon_dut_item.addr);
 
-          cfg.vif.p_edge_scl();
-          r_bit = cfg.vif.cb.sda_i;
-          `uvm_info(`gfn, $sformatf("controller_address_thread() rw %d", r_bit), UVM_MEDIUM)
+          // Collect the R/W bit
+          begin
+            cfg.vif.p_edge_scl();
+            mon_dut_item.dir = rw_e'(cfg.vif.cb.sda_i);
```
rswarbrick, line 538: Would it make sense to encode this with a ternary like the following line? That makes it a bit easier to see the mapping at a glance.

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -353,176 +436,223 @@ class i2c_monitor extends dv_base_monitor #(
...
+            mon_dut_item.bus_op = (mon_dut_item.dir == i2c_pkg::READ) ? BusOpRead : BusOpWrite;
+          end
 
-          mon_dut_item.bus_op = (r_bit) ? BusOpRead : BusOpWrite;
-          cfg.valid_addr = is_target_addr(mon_dut_item.addr);
-          cfg.is_read = r_bit;
+          `uvm_info(`gfn, $sformatf(
+            "controller_address_thread() got byte 0x%2x (address=2'h%2x (%0s), rw=1'b%b (%0s))",
+            {mon_dut_item.addr, mon_dut_item.dir},
```
rswarbrick, line 544: I'm not sure about "byte" here: the address is up to 10 bits and we're appending an extra bit. (We're also overflowing the `%2x`)

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -353,176 +436,223 @@ class i2c_monitor extends dv_base_monitor #(
...
-    mon_dut_item.ack    = 1'b0;
-    mon_dut_item.nack   = 1'b0;
-    mon_rstart = 0;
-    target_read_phase = 1;
-    // Previous data collecting thread replied on nack / stop
-    // For ack / stop test, this thread need to be forked with
-    // separate ack_stop_monitor
-
-    fork begin
+    bit [7:0] mon_data;
```
rswarbrick, line 580: I think this is a dead variable?

hw/ip/i2c/dv/env/i2c_env.sv
```diff
@@ -36,39 +36,30 @@ class i2c_env extends cip_base_env #(
...
+      m_i2c_agent.monitor.target_mode_rd_item_port.connect(
         scoreboard.controller_mode_rd_obs_fifo.analysis_export);
-      m_i2c_agent.monitor.wr_item_port.connect(
+      m_i2c_agent.monitor.target_mode_wr_item_port.connect(
         scoreboard.controller_mode_wr_obs_fifo.analysis_export);
-      m_i2c_agent.monitor.analysis_port.connect(
+      m_i2c_agent.monitor.controller_mode_rd_item_port.connect(
         scoreboard.target_mode_rd_obs_fifo.analysis_export);
+      m_i2c_agent.monitor.controller_mode_wr_item_port.connect(
+        scoreboard.target_mode_wr_obs_fifo.analysis_export);
```
rswarbrick, line 48: This is much cleaner (and more symmetrical) than it was. It's a pity that the monitor now has a defunct `analysis_port` variable from the base class. Definitely not for this PR, but I wonder whether we can make it more obviously rubbish (maybe by setting the item class to some base type).

hw/ip/i2c/dv/env/i2c_env_cfg.sv
```diff
@@ -41,7 +41,7 @@ class i2c_env_cfg extends cip_base_env_cfg #(.RAL_T(i2c_reg_block));
   // any larger transaction. A transfer with no data, where the address byte is NACK'd,
   // is possible with our stimulus approach, and this can break checking in certain
   // circumstances.
-  int        min_xfer_len = 0;
+  int        min_xfer_len = 1;
```
rswarbrick, line 44: Does the comment above need tweaking here? I don't think we ever have `min_xfer_len=0` any more (a grep only finds it being overridden with 3)

hw/ip/i2c/dv/env/i2c_env_cfg.sv
```diff
@@ -57,24 +57,19 @@ class i2c_env_cfg extends cip_base_env_cfg #(.RAL_T(i2c_reg_block));
...
-  bit        read_rnd_data = 0;
-
   ////////////////////////////////
   // Helper Variables (not cfg) //
   ////////////////////////////////
 
-  int        sent_acq_cnt;
-  int        rcvd_acq_cnt;
-  bit [7:0]  lastbyte;
+  // These variables are used to track how many items we expect to produced at the ACQFIFO for some
```
rswarbrick, line 64: Nit: "to be produced"?

hw/ip/i2c/dv/env/i2c_env_cfg.sv
```diff
@@ -57,24 +57,19 @@ class i2c_env_cfg extends cip_base_env_cfg #(.RAL_T(i2c_reg_block));
...
+  // given stimulus. This is used by the testbench stimulus to know when it should start reading
+  // from the ACQFIFO, and also to know when it can stop to end the test.
+  // Note. REFACTOR
```
rswarbrick, line 67: Stray todo... :-)

hw/ip/i2c/dv/env/i2c_env_cfg.sv
```diff
@@ -57,24 +57,19 @@ class i2c_env_cfg extends cip_base_env_cfg #(.RAL_T(i2c_reg_block));
...
+  int        sent_acq_cnt; // How many ACQFIFO items we expect the be produced
+  int        rcvd_acq_cnt; // How many ACQFIFO items we have observed by reading it
 
-  // Flags for the 'ack_stop' test
-  int        sent_ack_stop = 0;
-  int        rcvd_ack_stop = 0;
+  // The i2c_host_fifo_fmt_empty_vseq uses this variable for checking purposes.
+  bit [7:0]  lastbyte;
```
rswarbrick, line 72: It's a little sad to have this, where the scoreboard is sending something to a particular vseq :-) I'm not sure whether it's worth it, but we could add a TODO and an issue to track that this is a bit nuts...

hw/ip/i2c/dv/env/i2c_env_cfg.sv
```diff
@@ -124,12 +119,10 @@ class i2c_env_cfg extends cip_base_env_cfg #(.RAL_T(i2c_reg_block));
     min_data = 1;
     max_data = 60;
     read_all_acq_entries = 0;
-    sent_ack_stop = 0;
-    rcvd_ack_stop = 0;
   endfunction : reset_seq_cfg
 
   task wait_fifo_not_empty(i2c_analysis_fifo fifo);
-    while (!fifo.is_empty()) begin
+    while (fifo.is_empty()) begin
```
rswarbrick, line 125: Ha! I claim this shows we didn't choose the best of names :-)

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -331,19 +314,27 @@ class i2c_reference_model extends uvm_component;
       // of the transfer. Upon further writes to the FMTFIFO, we will fill in the remaining
       // fields.
       case (fbyte[0])
-        1'b1: begin // read transfer
+        i2c_pkg::READ: begin // read transfer
```
rswarbrick, line 317: Not a problem, but you can probably drop the comment here and on line 328: you're naming it now :-)

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -387,12 +380,12 @@ class i2c_reference_model extends uvm_component;
...
-          exp_rd_item.read   = readb;
-          exp_rd_item.nakok  = nakok;
-          exp_rd_item.nack   = ~exp_rd_item.rcont;
-          exp_rd_item.rstart = (exp_rd_item.stop) ? 1'b0 : 1'b1;
+          exp_rd_item.stop        = stop;
+          exp_rd_item.rcont       = rcont;
+          exp_rd_item.read        = readb;
+          exp_rd_item.nakok       = nakok;
+          // exp_rd_item.nack        = ~exp_rd_item.rcont;
+          // exp_rd_item.rstart_back = (exp_rd_item.stop) ? 1'b0 : 1'b1;
```
rswarbrick, line 388: What should happen with this commented code?

hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -494,67 +489,114 @@ class i2c_reference_model extends uvm_component;
...
-      end
-      default:;
-    endcase
 
-    // Push the item to the scoreboard as our rd_exp item.
-    target_mode_rd_port.write(exp_rd);
-  endtask: create_rd_exp_from_stim
+  function bit is_target_addr(bit [6:0] addr);
+    return (addr == target_addr0 || addr == target_addr1);
+  endfunction
```
rswarbrick, line 599: I think this duplicates a function in `i2c_agent_cfg`. Maybe we can drop this one? Or is the point that the agent config should be about stimulus and this should be about modelling?

rswarbrick (commented):
Notes up to end of i2c_reference_model in final commit:

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -62,28 +62,44 @@ class i2c_base_vseq extends cip_base_vseq #(
...
-  int exp_wr_id = 0;
 
-  // read_txn_q usage: a single byte of read data and only wdata is valid.
-  i2c_item read_txn_q[$];
-  int tran_id = 0;
-  int sent_txn_cnt = 0;
+  // Start counter including restart.
+  // Starting from 1 to match rtl value and easy to trace.
+  int start_cnt = 1;
+  int read_txn_num = 0; // Duplicate of above?
```
rswarbrick, line 69: More than that: it's not actually used elsewhere in the repository :-) Can just delete it!

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -612,326 +639,413 @@ class i2c_base_vseq extends cip_base_vseq #(
...
+    // A local variable used to accumulate the new transaction as we construct it.
+    i2c_transaction txn;
+
+    // Randomize the number of data bytes total in the transaction.
+    `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(num_data_total,
+      num_data_total inside {[cfg.min_data : cfg.max_data]};
+    )
+
+    // Randomize the number of transfers in the transaction.
+    num_transfers = $dist_normal(/* seed */ seed,
```
rswarbrick, line 802: I'm a bit unconvinced by Verible here. But I guess you could do something ridiculous like using a step function. It could be something like "90% of the time, I expect to be roughly equal to the given fraction of num_data_total. The rest of the time, take anything possible". The resulting code might look like:
```systemverilog
    // Randomize the number of transfers in the transaction.
    begin
      int lo, hi;
      randcase
        9: begin
          int mean = int'(num_data_total * real'(cfg.rs_pct / 100.0));
          lo = int'(mean * real'(0.9));
          hi = int'(mean * real'(1.1));
        end
        1: begin
          lo = 1;
          hi = num_data_total;
        end
      endcase

      // Saturate into the possible range if we happen to fall outside of it.
      if (hi < lo) hi = lo;
      if (lo < 1) lo = 1;
      if (hi > num_data_total) hi = num_data_total;

      num_transfers = $urandom_range(lo, hi);
    end
```
(completely untested...)

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -612,326 +639,413 @@ class i2c_base_vseq extends cip_base_vseq #(
...
+            xfer.rstart_front = 1;
+            xfer.stop = 1;
+          end
+          default: begin // Any transfer not at the start/end of the txn must begin and end with
+            // rstart conditions.
+            xfer.rstart_front = 1;
+            xfer.rstart_back = 1;
+          end
+        endcase
+      end
```
rswarbrick, line 868: I think this could be a bit simpler:
```systemverilog
      xfer.start        = (i == 0);
      xfer.stop         = (i == num_transfers - 1);
      xfer.rstart_front = (i > 0);
      xfer.rstart_back  = (i < num_transfers - 1);
```
Isn't that equivalent?

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -612,326 +639,413 @@ class i2c_base_vseq extends cip_base_vseq #(
...
-          `downcast(exp_txn, txn.clone());
-          // Write payload
-          dst_q.push_back(txn);
-          if (valid_addr) begin
-            exp_txn.tran_id = this.tran_id++;
-            p_sequencer.target_mode_wr_exp_port.write(exp_txn);
-            cfg.sent_acq_cnt++;
+          BusOpRead: begin
+            // For read bytes, we wait 8 bit-periods then ack or nack (The driver behaviour
+            // for 'HostAck' and 'HostNAck' both include this 8-bit wait)
```
rswarbrick, line 964: I think "... driver behaviour for both 'HostAck' and 'HostNAck' includes ..."

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -1126,59 +1254,49 @@ class i2c_base_vseq extends cip_base_vseq #(
...
-    cfg.wr_pct = 0;
-
-    `uvm_create_obj(i2c_target_base_seq, m_i2c_host_seq)
-    create_txn(txn_q);
-    fetch_txn(txn_q, m_i2c_host_seq.req_q, 1);
-    m_i2c_host_seq.start(p_sequencer.i2c_sequencer_h);
-    sent_txn_cnt++;
-  endtask
 
+  virtual task send_ack_stop; endtask
```
rswarbrick, line 1258: I think this probably needs a documentation comment (possibly by nicking it from `i2c_target_ack_stop_vseq` and prefixing with "This could be..." or similar :-D)

rswarbrick (commented):
Oof, what an enormous PR! These comments get to the bottom of the last commit. I'd personally grateful if you could split out smaller PRs and send them earlier next time, even if you end up having to re-do stuff.

hw/dv/sv/i2c_agent/seq_lib/i2c_target_base_seq.sv
```diff
@@ -7,18 +7,28 @@ class i2c_target_base_seq extends i2c_base_seq;
...
-    // So, instead of randomizing i2c_item here, assuming i2c_item is randomized in
-    // vseq and fed to req_q here.
-    `DV_SPINWAIT_EXIT(wait (req_q.size() > 0);
-                      while (req_q.size() > 0) begin
-                        req = req_q.pop_front();
-                        start_item(req);
-                        finish_item(req);
-                      end
-                      , wait (stop);)
+    // A sequence of i2c_items are randomized in vseq and fed to req_q here.
```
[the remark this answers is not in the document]
rswarbrick: I think this is still missing the "that" :-

hcallahan-lowrisc:
The latest push removes the 12 commits at the beginning of this PR that were moved to https://github.com/lowRISC/opentitan/pull/23919, and rebases on top of those changes now it has been merged. Review feedback will be addressed in further pushes.

hw/dv/sv/i2c_agent/i2c_agent_cfg.sv
```diff
@@ -65,8 +66,6 @@ class i2c_agent_cfg extends dv_base_agent_cfg;
   // Set this bit when there is the possibility of generating Agent-Controller stimulus
   // transfers where the address does not match that configured into the DUT.
   bit       allow_bad_addr = 0;
```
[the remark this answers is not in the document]
hcallahan-lowrisc: True, removed

hw/dv/sv/i2c_agent/i2c_agent_cfg.sv
```diff
@@ -48,9 +47,11 @@ class i2c_agent_cfg extends dv_base_agent_cfg;
...
   int     host_scl_pause_cyc = 0;
 
-  // ack followed by stop test mode
-  bit     allow_ack_stop = 0;
-  bit     ack_stop_det = 0;
+  // ack-stop test mode
+  // This mode is used to generate stimulus that expects the assertion of the 'unexp_stop' irq
+  bit     allow_ack_stop = 0; // Enable the test mode, including generating the modified stimulus
+  bit     ack_stop_det = 0; // The monitor asserts this when an ack-then-stop stimulus has occurred
+  bit     force_ack = 0;
```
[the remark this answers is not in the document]
hcallahan-lowrisc: This bit wasn't used in the agent anymore, and is related to stimulus generation really. So I've moved it to the base_vseq and added a comment there.

hw/dv/sv/i2c_agent/i2c_item.sv
```diff
@@ -4,25 +4,34 @@
...
-  // transaction control part
-  bit                      nack;
-  bit                      ack;
-  bit                      rstart;
+  i2c_pkg::acknack_e       addr_ack; // ACK-bit for addr+dir byte
 
-  // queue dropped data due to fmt_overflow
-  bit [7:0]                fmt_ovf_data_q[$];
+  // Transaction data part
+  int                      num_data; // Number of valid data bytes (== data_q.size())
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Yes I'm definitely considering doing that in the future :)

hw/dv/sv/i2c_agent/i2c_item.sv
```diff
@@ -4,25 +4,34 @@
...
+  bit [7:0]                data_q[$]; // Queue of data bytes in the transfer
+  bit                      data_ack_q[$]; // Holds the corresponding ack/nack for data_q bytes
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Absolutely worth considering, I already have trialled this tweak in a separate changeset I'm working on for a multi-controller capable i2c_agent. Still not 100% sure how I want to break things down for the cleanest abstraction (as you say, the data byte is conceptually separate from it's corresponding ack bit, though an I2C Target Device can choose to ack or nack, or even clock stretch depending on the value of the data byte, so it's a decision point that requires some distinction)

hw/dv/sv/i2c_agent/i2c_item.sv
```diff
@@ -65,71 +74,109 @@ class i2c_item extends uvm_sequence_item;
...
+    str = {str, $sformatf("%s:bus_op       = %s\n",    pname, bus_op.name)};
+    str = {str, $sformatf("%s:addr         = 0x%2x\n", pname, addr)};
+    str = {str, $sformatf("%s:num_data     = %0d\n", pname, num_data)};
+    str = {str, $sformatf("%s:start        = %1b\n", pname, start)};
+    str = {str, $sformatf("%s:stop         = %1b\n", pname, stop)};
+    str = {str, $sformatf("%s:read         = %1b\n", pname, read)};
+    str = {str, $sformatf("%s:rstart_front = %1b\n", pname, rstart_front)};
+    str = {str, $sformatf("%s:rstart_back  = %1b\n", pname, rstart_back)};
     foreach (data_q[i]) begin
       str = {str, $sformatf("%s:data_q[%0d]=0x%2x\n", pname, i, data_q[i])};
```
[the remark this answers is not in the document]
hcallahan-lowrisc: I'll address this in a follow-up PR :)

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -34,17 +34,37 @@ class i2c_monitor extends dv_base_monitor #(
...
+  uvm_analysis_port #(i2c_item) controller_mode_rd_item_port;
+  uvm_analysis_port #(i2c_item) target_mode_wr_item_port;
+  uvm_analysis_port #(i2c_item) target_mode_rd_item_port;
+
+  // This item is used to capture and accumulate ongoing transaction control and data symbols.
+  // It is the main piece of state used by a monitor instance to keep track of the ongoing I2C
+  // transaction, across all subroutines. Once a transaction has completed, this item is cloned and
+  // published to the appropriate TLM ports.
+  // Note. that this item is only constructed once (in the class constructor), and it's state is
+  // partially reused/cleared by using i2c_methods clear_data() and clear_all() after cloning and
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Now that I re-read it, this whole comment is stale (it's described function is replaced by the variable 'prev_item'. Deleted.

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -132,32 +155,76 @@ class i2c_monitor extends dv_base_monitor #(
...
 
   virtual protected task target_collect_thread();
-    i2c_item full_item;
-    if (mon_dut_item.stop ||
-       (!mon_dut_item.stop && !mon_dut_item.start && !mon_dut_item.rstart)) begin
+
+    // Wait for the driver/vseq to clear this at the start of the next stimulus round.
+    wait(cfg.got_stop == 0);
+
+    `uvm_create_obj(i2c_item, mon_dut_item);
```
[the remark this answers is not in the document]
hcallahan-lowrisc: This sounds like a sensible safety net to me. I'm hoping to refactor a lot of this logic away in the future, but I'll file the suggestion under future improvements just so it doesn't get lost. Thanks

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -132,32 +155,76 @@ class i2c_monitor extends dv_base_monitor #(
...
-      `uvm_info(`gfn, "Detected START condition.", UVM_HIGH)
+      `uvm_info(`gfn, "target_collect_thread(): Saw START condition.", UVM_MEDIUM)
+
+      mon_dut_item.start = 1'b1;
+
     end else begin
-      mon_dut_item.rstart = 1'b1;
+      // If we got to the else-condition, it means our decoding above failed, and somehow the
+      // items are in an unexpected state. This may be a testbench bug.
+      `uvm_fatal(`gfn, "State at start of target_collect_thread() is unexpected!")
```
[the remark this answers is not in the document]
hcallahan-lowrisc: Eventually I hope to model at a level where the bus control conditions (start/rstart/stop) are handled at a lower level of abstraction, maybe part of a different sequence item or even maybe just driver/monitor state implementation details. In that case, the types / code here will likely change significantly.
Probably worth putting a monitor check like you suggest in at some point, though there is the complication that another invalid but checkable of-sorts ending condition for transfers is timeout (where the bus is released with no activity). Eventually that will need to be captured too...

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: Agreed! I hope to remove this too, filed under future improvements for now ;)

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: SGTM

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: Yep this was debug code that never got removed :)

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: Done!

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: yep

hw/ip/i2c/dv/env/i2c_env.sv
hcallahan-lowrisc: Yeah good point for a future improvement. Eventually we will condense everything down into one port/fifo for read and write transfers, as there is no real reason for them to be separate.

hw/ip/i2c/dv/env/i2c_env_cfg.sv
hcallahan-lowrisc: TODO tracked in #23920 :+1:

hw/ip/i2c/dv/env/i2c_reference_model.sv
hcallahan-lowrisc: First line is dead, so removed. Second is something I will be coming back to in follow up PRs.

hw/ip/i2c/dv/env/i2c_reference_model.sv
hcallahan-lowrisc: This is another thing I haven't yet completely cleared up yet. We still have the variable `i2c_agent_cfg.read_addr_q[$]`, which passes a "VALID/INVALID" bit for every observed Agent-Controller read transfer, and is used to adjust the stimulus generation in the vseqs. Obviously the i2c_agent having knowledge of what addresses the DUT is currently configured with is pretty ugly. Added a note to #23920 to come back and finish cleaning this up.

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
hcallahan-lowrisc: Gone!

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
hcallahan-lowrisc: I just added a Verible waiver here for now, as it seems pretty harmless to me. But I've added a note to #23920 to come back and consider a different implementation in the future.

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
hcallahan-lowrisc: Maybe, but I like mine more :p

hcallahan-lowrisc:
I've marked this ready for review, but it's not yet ready for merging as there are a couple more regression failures I need to cleanup first.
@rswarbrick when you have a moment, would you mind checking the changes based on your feedback? Thanks

hw/dv/sv/i2c_agent/i2c_item.sv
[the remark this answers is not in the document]
rswarbrick: Yeah, that makes sense. Maybe things will become more obvious as things crystallise otherwise :-)

hw/dv/sv/i2c_agent/i2c_monitor.sv
rswarbrick: Thanks, and I realise that I've thrown out a whole laundry list of possible future improvements. Would it make sense to stick them in a "big bucket" issue so that they are somewhere obvious? (Clearly not needed for this PR, but it might be a belt-and-braces way to avoid dropping a potential improvement)

hw/dv/sv/i2c_agent/i2c_monitor.sv
```diff
@@ -353,245 +428,266 @@ class i2c_monitor extends dv_base_monitor #(
...
+
     mon_dut_item.tran_id = num_dut_tran++;
-    mon_dut_item.start = 1'b1;
 
-    controller_address_thread(skip_the_loop);
-    if (skip_the_loop) begin
-      `uvm_info(`gfn, "Skipping the loop now.", UVM_MEDIUM)
-      return;
+    // Capture the transfer
+    // - If an external reset is requested, return early without pushing a sequence item.
```
rswarbrick, line 467: It might be worth moving this second line to come just before the `return` line a bit below. It took me a few seconds to understand how this all worked.

hw/dv/sv/i2c_agent/i2c_monitor.sv
rswarbrick: I think this is still being done with a cast. Would you mind turning it into a ternary?

hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
rswarbrick: Really sorry but I think we're still missing the 's' here (needed because behaviour is singular).

rswarbrick (commented):
Thank you very much for addressing so much. A couple of nitty follow-ups here, but it looks very sensible to me.

hw/dv/sv/i2c_agent/i2c_monitor.sv
hcallahan-lowrisc: I've switched this (and another cast in the same task) to use a ternary.

hcallahan-lowrisc:
The latest push addresses review comments, and fixes up a couple of bugs I noticed in the meantime. Hopefully it should be ready to merge at this point.

This change has affected the wider block level regression results, as many directed tests still use the old monitor abstractions. I will be fixing them up in follow up PRs to this one.
The same can be said for the chip-level I2C smoketests (`chip_sw_i2c_{host,device}_tx_rx`), which these changes also break. I have pushed a final commit which temporarily removes them from the smoketest regression. There will shortly be another PR to re-introduce them.

rswarbrick (approved):
This looks reasonable to me. Thanks for all the tidying up.

marnovandermaas (approved):
Looks reasonable to me thanks.

hcallahan-lowrisc:
> This change has affected the wider block level regression results, as many directed tests still use the old monitor abstractions. I will be fixing them up in follow up PRs to this one. The same can be said for the chip-level I2C smoketests (`chip_sw_i2c_{host,device}_tx_rx`), which these changes also break. To save this PR from becoming any larger, I have pushed a final commit which temporarily removes them from the smoketest regression. There will shortly be another PR to fix and re-introduce them.

This PR also broke the I2C component of the power_virus TLT, so I have temporarily removed the I2C stimulus from running during the max_power_epoch when running in DV. This change, like the one described above, is only temporary, and will be reverted once the appropriate changes to further fixup DV have landed. This PR is already big enough.

#23959 will re-enable the TLT-components mentioned above by reverting the two commits that disable them. This PR depends on a few other to land, so there might be a short delay. I've not tracked this point in an issue, as hopefully it will be very transient.

PR #23912, opened 2024-07-03
[dv] fix rom_raw_unlock test

timothytrippel:
This fixes the ROM raw unlock test by simply adding a `override_test_status_and_finish(.passed(1));` invocation to the end of sequence body, since there is no need to wait for a flash image to boot to confirm the chip has transitioned from raw to test_unlocked0 LC states.

Additionally, this adds some console prints to the sequence to aid in debugging in the future.


timothytrippel:
CC: @sha-ron

cfrantz (approved):

PR #23913, opened 2024-07-03
[rom_e2e,dv] fix ROM self hash test in DV

timothytrippel:
This changes the ROM self hash test to run in the PROD LC state and increases the timeout.

### DIFF
--- a/hw/top_earlgrey/dv/chip_rom_tests.hjson
+++ b/hw/top_earlgrey/dv/chip_rom_tests.hjson
@@ -788,8 +788,8 @@
       name: rom_e2e_self_hash
       uvm_test_seq: chip_sw_base_vseq
       sw_images: [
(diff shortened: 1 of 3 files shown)


cfrantz (approved):

vogelpi (approved):
LGTM, thanks @timothytrippel !

sha-ron (approved):
Will it work in TEST_UNLOCK0 life cycle?

timothytrippel:
It does not look like this alone will make the ROM self hash test pass in DV. I am looking deeper into the issue. (Note: the test is passing in DV).

timothytrippel:
It looks like the issue was that the test was missing a manifest when built for the sim_dv exec_env. I have fixed that now.

timothytrippel:
> Will it work in TEST_UNLOCK0 life cycle?

@sha-ron yes, the test is signed with PROD keys so it will work in any LC state if you change the LC state of the OTP image.

PR #23918, opened 2024-07-04
[i2c,dv] I2C SMBus maxlen directed test

### LINKED ISSUE #22252: [i2c,dv] Verify handling of max-length SMBus messages
In PR https://github.com/lowRISC/opentitan/pull/21789, we increased the depth of the ACQ FIFO so it can absorb max-length SMBus messages (see the linked PR for a description of such messages) without having to rely on SW to drain the FIFO. We should **add a directed (or tightly constrained) test to check that this holds**, first for the simple case in which the ACQ FIFO starts empty. We should then **add tests** (or extend the previous test) **to cover more complex scenarios**, in which the ACQ FIFO is already almost full and SW is already draining it or will soon start to drain it, and a max-length SMBus message arrives (SW should be modeled such that it drains the ACQ FIFO fast enough that it never overflows; otherwise the i2c IP would have to NACK, which is a separate issue).

hcallahan-lowrisc:
> Note. ~~This test depends on changes from #23911, which as been squashed into the first commit. Changes from this commit should be ignored while reviewing, as it will be rebased away when that PR is merged.~~

This directed test aims to check the I2C block can absorb a max-length SMBus message
without software intervention mid-message (e.g. to drain the ACQFIFO, or handle interrupts).

 - Stimulate a max-length transfer (Block-Write of 255 bytes)
 - Disable all interrupt/polling based handlers
 - Wait for the command-complete interrupt to signal the end of the transfer
 - Readback the transfer and check it was received correctly (done via scoreboard)

A new V2 testpoint is added to track this test against.

Closes #22252 

In the future, it might be desirable to confirm the intended behaviour via better constrained randomization and an appropriate coverage bin, rather than a specific vseq.

### DIFF
--- a/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
+++ b/hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
@@ -826,6 +826,8 @@ class i2c_base_vseq extends cip_base_vseq #(
   virtual function i2c_transaction create_txn();
     int seed = $urandom;
 
+    int min_num_transfers, max_num_transfers;
+
     // A temporary, used to randomize the total number of data bytes in the transaction.
     int unsigned num_data_total;
     // A temporary, used to randomize the total number of transfers in the transaction.
@@ -843,6 +845,12 @@ class i2c_base_vseq extends cip_base_vseq #(
       num_data_total inside {[cfg.min_data : cfg.max_data]};
     )
 
+    min_num_transfers = (cfg.min_num_transfers >= 1) ? cfg.min_num_transfers : 1;
+    max_num_transfers = (cfg.max_num_transfers == 0)             ? num_data_total :
+                        (cfg.max_num_transfers > num_data_total) ? num_data_total :
+                        cfg.max_num_transfers;
+    `DV_CHECK(min_num_transfers <= max_num_transfers)
+
     // Randomize the number of transfers in the transaction.
     // TODO(#23920) Consider re-implementing to avoid waiver.
     num_transfers = $dist_normal( // verilog_lint: waive invalid-system-task-function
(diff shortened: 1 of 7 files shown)


hw/ip/i2c/dv/env/seq_lib/i2c_base_vseq.sv
```diff
@@ -798,16 +800,21 @@ class i2c_base_vseq extends cip_base_vseq #(
       num_data_total inside {[cfg.min_data : cfg.max_data]};
     )
 
+    min_num_transfers = (cfg.min_num_transfers > 1) ? cfg.min_num_transfers : 1;
```
rswarbrick, line 803: Nit: I'd probably make this ">=1", since we want the first option to be taken unless something is out of range. Clearly not a big deal though!

hw/ip/i2c/dv/env/seq_lib/i2c_target_smbus_maxlen_vseq.sv
```diff
@@ -0,0 +1,64 @@
...
+    cfg.ack_ctrl_en = 0;
+
+    // Use a basic Agent sequence that drives all items it is passed.
+    i2c_base_seq::type_id::set_type_override(i2c_target_base_seq::get_type());
+
+  endtask
+
+  virtual task initialization();
+    super.initialization();
+    // The base-class enables all interrupts by default, so overwrite that behaviour here.
```
rswarbrick, line 49: Nit: I think you might want "override" here.

rswarbrick (commented):
The last two commits look really sensible (modulo some extremely nitty comments). Very happy for this to go in once the mammoth PR has landed :-)

hw/ip/i2c/dv/env/seq_lib/i2c_target_smbus_maxlen_vseq.sv
hcallahan-lowrisc: Done.

hw/ip/i2c/dv/env/seq_lib/i2c_target_smbus_maxlen_vseq.sv
```diff
@@ -0,0 +1,63 @@
...
+// - Wait for the command-complete interrupt to signal the end of the transfer
+// - Readback the transfer and check it was received correctly (done via scoreboard)
+//
+class i2c_target_smbus_maxlen_vseq extends i2c_target_runtime_base_vseq;
+  `uvm_object_utils(i2c_target_smbus_maxlen_vseq)
+  `uvm_object_new
+
+  // The maximum number of data-bytes in a SMBus Block Write
+  // 1 - command code
+  // 1 - byte count
```
marnovandermaas, line 19: Don't we need to somehow constrain the byte count to be equal to 255?

marnovandermaas (approved):
Looks very reasonable to me from a code inspection and reading Rupert's last review. Just one question that came to my mind but nothing blocking.

hw/ip/i2c/dv/env/seq_lib/i2c_target_smbus_maxlen_vseq.sv
hcallahan-lowrisc: Yep we do, in this case by constraining the randomization by setting both limits `cfg.min_data` and `cfg.max_data` to 258. From the perspective of our I2C block, this SMBus Block-Write message is just a completely normal write-transfer with 258 bytes of data (1 + 1 + 255 + 1). The meaning of those bytes is software-defined for us.

PR #23919, opened 2024-07-04
[i2c,dv] I2C transfer level TLM pre-cleanup

hcallahan-lowrisc:
This commit is split out from #23911, thanks for the review feedback @rswarbrick !

This contains 12 of the first 14 commits, which are mostly uninteresting cleanups that go towards supporting the bulk of the changes in the final commit in that PR. The following two commits weren't removed from that PR, as it turns out they weren't quite atomic from the final commit.
- 420bbf7796ceeee0a9ec38a01876a0f062718425 [i2c,dv] Change bad_addr_pct and rs_pct to range from 1\~100 instead of 1\~10
- 9152b7f57cfb91e6fe3a638a5dd591ee1ef660f7 [i2c,dv] Update acq2item helper method to return 'i2c_acqdata_item'


hcallahan-lowrisc:
Note that this PR doesn't yet incorporate the review feedback from #23911. I will push those changes now, so the diff can be easily seen in GH.

hcallahan-lowrisc:
Latest push addresses the review comments for these commits/files in #23911

hw/dv/sv/i2c_agent/seq_lib/i2c_target_base_seq.sv
```diff
@@ -7,18 +7,29 @@ class i2c_target_base_seq extends i2c_base_seq;
...
-    // So, instead of randomizing i2c_item here, assuming i2c_item is randomized in
-    // vseq and fed to req_q here.
-    `DV_SPINWAIT_EXIT(wait (req_q.size() > 0);
-                      while (req_q.size() > 0) begin
-                        req = req_q.pop_front();
-                        start_item(req);
-                        finish_item(req);
-                      end
-                      , wait (stop);)
+    // A sequence of i2c_items are randomized in the vseq and are fed into the req_q here.
```
rswarbrick, line 10: Nit: Missing "that"

rswarbrick (approved):
Tiny nit (which I think got dropped when extracting from the source PR). Assuming it's otherwise the same, this looks good to me.

rswarbrick:
I've just checked and the changes to the RTL directory are:
```
diff --git a/hw/ip/i2c/rtl/i2c_pkg.sv b/hw/ip/i2c/rtl/i2c_pkg.sv   
index c9ca6e19f0..b13a1bdfaa 100644
--- a/hw/ip/i2c/rtl/i2c_pkg.sv
+++ b/hw/ip/i2c/rtl/i2c_pkg.sv
@@ -49,4 +49,14 @@ package i2c_pkg;                                       
   // byte and an identifier defined by i2c_acq_byte_id_e.
   parameter int unsigned ACQ_FIFO_WIDTH = I2C_ACQ_BYTE_ID_WIDTH + 8;
                                                                                                                                                     
+  typedef enum bit {                                                     
+    WRITE = 1'b0,
+    READ = 1'b1                                                                                                                                     
+  } rw_e;                                                                                                                                           
+                                                                                                                                                    
+  typedef enum bit {                                                     
+    ACK = 1'b0,    
+    NACK = 1'b1
+  } acknack_e;      
+                       
 endpackage : i2c_pkg    
```
Addition of an enum is definitely not going to change the RTL (but doing this avoids needing a separate "utils" package or similar, to make the enums visible to the agent).

Similarly, there are changes reported in the `data` directory, but these are just documentation additions in `i2c.hjson` and `i2c_testplan.hjson`.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/i2c/data/i2c.hjson
CHANGE AUTHORIZED: hw/ip/i2c/rtl/i2c_pkg.sv

hw/dv/sv/i2c_agent/seq_lib/i2c_target_base_seq.sv
hcallahan-lowrisc: Yes it turns out my grammar skills are not so hot today :)

hcallahan-lowrisc:
CHANGE AUTHORIZED: hw/ip/i2c/data/i2c.hjson
CHANGE AUTHORIZED: hw/ip/i2c/rtl/i2c_pkg.sv

PR #23924, opened 2024-07-04
[sw/tests] Fix `rv_dm_access_after_wakeup` in DV

andreaskurth:
The C code kept waiting for `kSoftwareBarrier` to become 1, which the DV code did set via a backdoor memory write but was not visible to SW. This caused a timeout on the DV side (the spinwait on "Entering normal sleep.").

Qualifying `kSoftwareBarrier` with `const` in the C code puts the variable into flash instead of SRAM, which fixes the problem: the value updated from DV is now visible to SW.

### DIFF
--- a/sw/device/tests/rv_dm_access_after_wakeup.c
+++ b/sw/device/tests/rv_dm_access_after_wakeup.c
@@ -32,7 +32,7 @@ enum {
 };
 
 // This location will be update from SV
-static volatile uint8_t kSoftwareBarrier = 0;
+static volatile const uint8_t kSoftwareBarrier = 0;
 
 // Handle to the plic
 dif_rv_plic_t rv_plic;


timothytrippel (approved):
Thanks @andreaskurth !

andreaskurth:
Thanks for your review, @timothytrippel. As [discussed with you and @pamaury on Slack](https://opentitan.slack.com/archives/C01073VLYP4/p1720109050127979), the problem is that variables in (main) SRAM don't seem to get updated by `sw_symbol_backdoor_overwrite`, but variables in flash do. I created issue https://github.com/lowRISC/opentitan/issues/23926 to track this.

andreaskurth:
`rv_dm_access_after_wakeup` still fails on the FPGA, hence the _CW310 SiVal non-ROM_EXT Tests_ CI check fails. However, this DV-related fix should not negatively affect FPGA tests, thus I'm merging this PR.

nbdd0121:
I think we need to change the test to use const for DV and non-const for SiVal.

PR #23925, opened 2024-07-04
[dv,usbdev] Sequences for testing IN retraction and device timeouts

alees24:
This PR adds two sequences for the remaining test points:

- IN packet retraction, for Isochronous IN traffic primarily.
- Device timeout; checks that the device recovers from missing host handshake, corruption to/loss of DATA packet and delayed host handshake; USB 2.0 protocol specification stipulates the time window within which the device must timeout and become receptive to subsequent transactions, as part of the recovery mechanism.

Note that the first of these sequences relies upon RTL changes and is not yet expected to pass on the current RTL; the RTL change is yet to be soak-tested and ratified, but this is a refinement to the behavior of a feature not presently requested or required.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_driver.sv
+++ b/hw/dv/sv/usb20_agent/usb20_driver.sv
@@ -102,8 +102,10 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     pkt.crc5 = {<<{pkt.crc5}};
     drive_packet("Token", pkt);
     // Attempt to collect a response only to full speed IN transactions; this should be DATA0|1
-    // or ACK/NAK.
-    if (req_item.m_pid_type == PidTypeInToken && !req_item.low_speed) begin
+    // or ACK/NAK. For fault injection purposes we may sometimes want to avoid waiting for any
+    // response.
+    if (req_item.m_pid_type == PidTypeInToken && req_item.await_response &&
+        !req_item.low_speed) begin
       device_response(rsp_item);
       seq_item_port.item_done(rsp_item);
       `uvm_info (`gfn, $sformatf("In drive afer In packet : \n %0s", rsp_item.sprint()), UVM_DEBUG)
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -315,9 +315,10 @@ endtask
     m_drv_lock.put(1);
   endtask
 
-  // Construct and transmit a token packet to the USB device
+  // Construct and transmit a token packet to the USB device, and usually await a response from
+  // the driver. For fault injection purposes, awaiting the response is optional.
   virtual task send_token_packet(bit [3:0] ep, pid_type_e pid_type,
-                                 bit [6:0] target_addr = dev_addr);
+                                 bit [6:0] target_addr = dev_addr, bit await_response = 1'b1);
     `uvm_create_on(m_token_pkt, p_sequencer.usb20_sequencer_h)
     start_item(m_token_pkt);
     m_token_pkt.m_ev_type  = EvPacket;
@@ -325,6 +326,7 @@ endtask
     m_token_pkt.m_pid_type = pid_type;
     assert(m_token_pkt.randomize() with {m_token_pkt.address == target_addr;
                                          m_token_pkt.endpoint == ep;});
+    m_token_pkt.await_response = await_response;
     // Any fault injections requested?
     if (inject_invalid_token_sync) m_token_pkt.valid_sync = 1'b0;
     if (inject_bad_token_crc5) m_token_pkt.crc5 = ~m_token_pkt.crc5;
@@ -477,10 +479,13 @@ endtask
     check_tx_packet(in_data, pid_type, data);
   endtask
 
-  // Send handshake to DUT after an appropriate turn-around delay.
-  virtual task send_handshake(input pid_type_e pid_type);
-    // Must delay for a few bit intervals before responding.
-    response_delay();
+  // Send handshake to DUT after an appropriate turn-around delay; the delay may be overriden for
+  // the small number of sequences that must precisely control the handshake timing.
+  virtual task send_handshake(input pid_type_e pid_type, bit delay_first = 1);
+    if (delay_first) begin
+      // Must delay for a few bit intervals before responding.
+      response_delay();
+    end
     // Construct and send handshake response.
     `uvm_create_on(m_handshake_pkt, p_sequencer.usb20_sequencer_h)
     start_item(m_handshake_pkt);
--- a/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
+++ b/hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
@@ -571,6 +571,120 @@ endtask
     end
   endtask
 
+  // Clear the 'pend(ing)' indicator and the 'sent' status for the given endpoint, so that there is
+  // no lingering completion status from an earlier IN packet.
+  // The returned bits indicate which bit was set, if any; mirroring the hardware bits in `configin`
+  // and `in_sent`, hardware guarantees that both cannot be set simultaneously for a single packet.
+  task clear_pend_sent(bit [3:0] ep, uvm_reg_data_t configin, output bit sent, output bit pend);
+    pend = get_field_val(ral.configin[ep].pend, configin);
+    if (pend) begin
+      ral.configin[ep].sending.set(1'b1);
+      ral.configin[ep].pend.set(1'b1);
+      csr_update(ral.configin[ep]);
+      sent = 1'b0;
+    end else begin
+      uvm_reg_data_t in_sent;
+      csr_rd(.ptr(ral.in_sent[0]), .value(in_sent));
+      sent = in_sent[ep];
+      if (sent) csr_wr(.ptr(ral.in_sent[0]), .value(1 << ep));
+    end
+  endtask
+
+  // Attempt to retract the current packet from the given IN endpoint; if the packet is currently
+  // being collected by the host, then this task will wait until the packet has been acknowledged,
+  // or transmission is retired unsuccessfully.
+  //
+  // The outcome is indicated to the caller via _one_ of the following becoming set.
+  //
+  // - 'sent' means that the packet was successfully sent to the USB host.
+  // - 'pend' means that an event caused it to be aborted and it remains 'pending'.
+  // - 'retracted' means that it was neither collected nor aborted, but has been succesfully
+  //   retracted and is no longer available for collection.
+  //
+  // The returned buffer number indicates the location of the packet, so that the buffer may be
+  // released or reused by the caller.
+  task retract_in_packet(bit [3:0] ep, uvm_reg_data_t configin, output bit sent, output bit pend,
+                         output bit retracted, output int prev_buf);
+    bit sending;
+    // Retraction of IN packet.
+    `DV_CHECK_EQ(get_field_val(ral.configin[0].rdy, configin), 1, "No IN packet to retract")
+    prev_buf = get_field_val(ral.configin[0].buffer, configin);
+    `uvm_info(`gfn, $sformatf("Retracting IN packet from endpoint %0d, buffer %0d", ep, prev_buf),
+              UVM_MEDIUM)
+
+    // Use csr_wr on the register to ensure that 'csr_update' does not inadvertently clear the
+    // 'sending' or 'pend' bits because of rw1c behavior; 'update' does not permit us to specify
+    // 'do not change' and thus ensure that the rw1c behavior is actually written as '0.'
+    //
+    // This is particularly important in avoiding a sw/hw race here.
+    configin = get_csr_val_with_updated_field(ral.configin[ep].sending, configin, 1'b0);
+    configin = get_csr_val_with_updated_field(ral.configin[ep].pend, configin, 1'b0);
+    // Clear the RDY bit to initiate packet retraction.
+    configin = get_csr_val_with_updated_field(ral.configin[ep].rdy, configin, 1'b0);
+    csr_wr(.ptr(ral.configin[ep]), .value(configin));
+
+    csr_rd(.ptr(ral.configin[ep]), .value(configin));
+    `DV_CHECK_EQ(get_field_val(ral.configin[ep].rdy, configin), 1'b0)  // Just cleared.
+    `DV_CHECK_EQ(get_field_val(ral.configin[ep].buffer, configin), prev_buf)  // Buffer unchanged.
+    // Is it being sent?
+    sending = get_field_val(ral.configin[ep].sending, configin);
+    if (sending) begin
+      // The IN packet is presently being collected; since the packet size is limited to 64 bytes,
+      // there is an upper bound on how long the transmission can take. With 4x oversampling, and
+      // bit stuffing it's around 'hA30 clock cycles, but difficult to calculate exactly; we just
+      // need to ensure that we wait long enough...
+      for (int unsigned clks = 0; sending && clks < 'hC00; clks++) begin
+        csr_rd(.ptr(ral.configin[ep]), .value(configin));
+        sending = get_field_val(ral.configin[ep].sending, configin);
+      end
+      `DV_CHECK_EQ(sending, 0, "Packet was neither collected nor retracted")
+    end
+    // Ensure that there's no lingering completion status that could interfere with the next
+    // IN packet presented.
+    clear_pend_sent(ep, configin, sent, pend);
+    retracted = !sent & !pend;
+  endtask
+
+  // Present the IN packet for collection, optionally permitting the retraction of an existing
+  // packet.
+  // If a packet is retracted, then its allocated buffer number is returned such that it may be
+  // released by the caller.
+  task present_in_packet(bit [3:0] ep, uint buf_num, uint len, bit may_retract = 1'b1,
+                         output bit retracted, output int retracted_buf);
+    uvm_reg_data_t configin;
+    bit sent, pend;
+
+    `uvm_info(`gfn, $sformatf("EP %d presenting buf %d, %d byte(s)", ep, buf_num, len), UVM_HIGH)
+    // Remove the existing IN packet, if there is one (packet retraction).
+    csr_rd(.ptr(ral.configin[ep]), .value(configin));
+    if (get_field_val(ral.configin[0].rdy, configin)) begin
+      `DV_CHECK_EQ(may_retract, 1, "Not permitted to retract pending IN packet")
+      retract_in_packet(ep, configin, sent, pend, retracted, retracted_buf);
+      `DV_CHECK_EQ(sent | pend | retracted, 1, "Unexpected IN packet outcome")
+    end else begin
+      // No packet retraction required.
+      retracted = 1'b0;
+      // Clear completion status of any previous IN packets.
+      clear_pend_sent(ep, configin, sent, pend);
+    end
+    // Validate the buffer properties before supplying them to the DUT.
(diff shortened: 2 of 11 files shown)


hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -571,6 +571,120 @@ endtask
     end
   endtask
 
+  // Clear the 'pending' indicator and the 'sent' status for the given endpoint, so that there is
+  // no lingering completion status from an earlier IN packet.
+  task clear_pend_sent(bit [3:0] ep, uvm_reg_data_t configin, output bit cleared);
+    if (get_field_val(ral.configin[ep].pend, configin)) begin
+      // TODO: ARE we able to use csr_update here?
```
rswarbrick, line 583: I think we can! If not, maybe it's worth delving into why not. (rw1c fields? But I *think* they should work properly?)

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -571,6 +571,120 @@ endtask
     end
   endtask
 
+  // Clear the 'pending' indicator and the 'sent' status for the given endpoint, so that there is
+  // no lingering completion status from an earlier IN packet.
```
rswarbrick, line 580: Can you extend this documentation comment to explain what the output through `cleared` means?

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -571,6 +571,120 @@ endtask
...
+    prev_buf = get_field_val(ral.configin[0].buffer, configin);
+    `uvm_info(`gfn, $sformatf("Retracting IN packet from endpoint %0d, buffer %0d", ep, prev_buf),
+              UVM_MEDIUM)
+
+    // Use csr_wr on the register to ensure that 'csr_update' does not inadvertently clear the
+    // 'sending' or 'pend' bits because of rw1c behavior.
+    configin = get_csr_val_with_updated_field(ral.configin[ep].sending, configin, 1'b0);
+    configin = get_csr_val_with_updated_field(ral.configin[ep].pend, configin, 1'b0);
+    configin = get_csr_val_with_updated_field(ral.configin[ep].rdy, configin, 1'b0);
+    csr_wr(.ptr(ral.configin[ep]), .value(configin));
```
rswarbrick, line 616: Given that `configin` only has five fields, I wonder whether it would be easier to unpack the argument explicitly at the start of the task and then extract the fields we care about, creating a value for this sort of write afterwards. I think it might be shorter, and it will probably be easier to read.

(Also, are you sure that `csr_update` is broken for rw1c fields? That sounds like a bug in the underlying DV code)

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -571,6 +571,120 @@ endtask
...
+      bit cleared;
+      for (int unsigned clks = 0; clks < 'hC00; clks++) begin
+        csr_rd(.ptr(ral.configin[ep]), .value(configin));
+        if (!get_field_val(ral.configin[ep].sending, configin)) break;
+      end
+      // There are three outcomes at this point:
+      // - 'in_sent' bit becomes sent indicating that the packet was transmitted successfully.
+      // - 'pend' bit becomes sent indicating that the transaction was aborted.
+      // - there was no packet pending, so nothing neither of the above occurs, or the packet was
+      //   not transferred successfully; had was not retracted, a retry could have collected it
```
rswarbrick, line 637: Grammar nit: "had was not" doesn't look right here (and I'm not completely sure I understand!)

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -571,6 +571,120 @@ endtask
...
+    csr_rd(.ptr(ral.configin[ep]), .value(configin));
+    if (get_field_val(ral.configin[0].rdy, configin)) begin
+      `DV_CHECK_EQ(may_retract, 1, "Not permitted to retract pending IN packet")
+      retract_in_packet(ep, configin, retracted, retracted_buf);
+    end else begin
+      bit cleared;
+      // No packet retraction required.
+      retracted = 1'b0;
+      // Clear completion status of any previous IN packets.
+      clear_pend_sent(ep, configin, cleared);
```
rswarbrick, line 674: I wonder whether it makes sense to turn `cleared` into an optional argument that defaults to null (and then only write to it if it isn't null). It would avoid needing to create the dummy variable here.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -571,6 +571,120 @@ endtask
...
+    `DV_CHECK_FATAL(buf_num < NumBuffers)
+
+    // Use csr_wr on the register; we deliberately want to clear the 'pend' and 'sending' status
+    // indicators at this point.
+    configin = get_csr_val_with_updated_field(ral.configin[ep].size, configin, len);
+    configin = get_csr_val_with_updated_field(ral.configin[ep].buffer, configin, buf_num);
+    configin = get_csr_val_with_updated_field(ral.configin[ep].sending, configin, 1'b1);
+    configin = get_csr_val_with_updated_field(ral.configin[ep].pend, configin, 1'b1);
+    configin = get_csr_val_with_updated_field(ral.configin[ep].rdy, configin, 1'b0);
+    csr_wr(.ptr(ral.configin[ep]), .value(configin));
```
rswarbrick, line 687: Wouldn't it be easier to use `csr_update` after doing several iterations of `ral.configin[ep].<my_field>.set(...)`?

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -571,6 +571,120 @@ endtask
...
+    // Now set the RDY bit
+    configin = get_csr_val_with_updated_field(ral.configin[ep].rdy, configin, 1'b1);
+    csr_wr(.ptr(ral.configin[ep]), .value(configin));
```
rswarbrick, line 690: Again, I think this should work with `csr_update`?

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
```diff
@@ -0,0 +1,188 @@
...
+  // Packets retrieved, from the perspective of the host side.
+  bit pkts_rcvd[];
+  // Packets sent, from the perspective of the device side.
+  bit pkts_sent[];
+
+  // Indication that the device side has completed.
+  bit complete = 1'b0;
+
+  // Use isochronous endpoint/transfers?
+  // TODO: presently we are concerned only with Isochronous traffic.
```
rswarbrick, line 23: Can you tie this to an issue (possibly after opening it)?

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
```diff
@@ -0,0 +1,188 @@
...
+  // collection attempts and device-side retractions.
+  // We used max length packets, which take ca. 4 x 550-700 clock ticks to transmit.
+  int unsigned min_delay = 320;
+  int unsigned max_delay = 3200;
+
+  // Host side collected IN packets until instructed to stop.
+  task host_side();
+    int prev_seq = -1;
+    do begin
+      // TODO: experiment/justify.
```
rswarbrick, line 36: I think this needs expanding? (Or maybe you have already justified the code with an experimient :-D)

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
```diff
@@ -0,0 +1,188 @@
...
+      int unsigned delay = $urandom_range(1, max_delay - min_delay);
+      usb20_item reply;
+      // Randomized delay, to ensure that we miss some of the packets.
+      // TODO: wait on host clock, please.
```
rswarbrick, line 40: Again, can you tie this to an issue number?

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
```diff
@@ -0,0 +1,188 @@
...
+      cfg.clk_rst_vif.wait_clks(delay);
+      // Isochronous packets are not ACKnowledged.
+      retrieve_in_packet(ep_default, reply, .ack(!iso));
+      `DV_CHECK_EQ(reply.m_ev_type, EvPacket)
+      case (reply.m_pid_type)
+        PidTypeData0, PidTypeData1: begin
+          // Packets shall arrive in order, because it's a single endpoint.
+          data_pkt data;
+          $cast(data, reply);
+          // Isochronous endpoints without pending data shall return a Zero Length Packet.
```
rswarbrick, line 50: I think I understand, but maybe this comment could be expanded. The point is that it explains why we need the check at all?

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
```diff
@@ -0,0 +1,188 @@
...
+      int unsigned nwords;
+      int buf_num;
+
+      // Collect and populate a buffer; we use the sequence number as the packet content so that
+      // the host side can identify which packet it has collected.
+      buf_num = buf_alloc();
+      `DV_CHECK_GE(buf_num, 0, "No buffer available")
+
+      // Repeat the sequence number throughout the buffer to create a longer packet.
+      // TODO: investigate whether there's any benefit to varying this.
```
rswarbrick, line 118: This probably needs an issue number if it's going in?

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
```diff
@@ -0,0 +1,188 @@
...
+
+    // Size the packets sent and packets received bitmaps appropriately; all bits are initially
+    // reset, indicating that the corresponding packet has neither been sent not received.
+    pkts_rcvd = new [num_trans];
+    pkts_sent = new [num_trans];
+
+    // Enable endpoint for IN traffic.
+    csr_wr(.ptr(ral.ep_in_enable[0].enable[ep_default]),  .value(1'b1));
+    // Isochronous traffic?
+    // TODO: presently we are concerned only with Isochronous traffic.
```
rswarbrick, line 169: This probably wants an issue number. But isn't the vseq isochronous (based on its name)? Maybe the todo actually needs to be in the testplan?

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
```diff
@@ -0,0 +1,188 @@
...
+    // Check that the two parties agree on which packets were actually transferred successfully.
+    `uvm_info(`gfn, $sformatf("device side sent   %p", pkts_sent), UVM_LOW)
+    `uvm_info(`gfn, $sformatf("host side received %p", pkts_rcvd), UVM_LOW)
+    for (int unsigned seq = 0; seq < num_trans; seq++) begin
+      if (pkts_sent[seq] != pkts_rcvd[seq]) begin
+        `uvm_info(`gfn, $sformatf("mismatched seq %0d - sent %0d rcvd %0d", seq,
+                                  pkts_sent[seq], pkts_rcvd[seq]), UVM_MEDIUM)
+      end
+    end
+    if (pkts_rcvd != pkts_sent) `uvm_fatal(`gfn, "Device- and host-side do not agree")
```
rswarbrick, line 186: I think we might also need to set a "mismatch flag" in the loop, to make sure the two sides don't have a disagreement about *which* packets were sent/received.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -43,6 +43,8 @@ class usbdev_max_usb_traffic_vseq extends usbdev_base_vseq;
     ep_in_enabled == ep_out_enabled;
   }
 
+  // TODO: Support Isochronous endpoints at some point; presently there is no way to differentiate
```
rswarbrick, line 46: This probably wants an issue number.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
```diff
@@ -387,17 +335,27 @@ class usbdev_max_usb_traffic_vseq extends usbdev_base_vseq;
...
-                                 get_field_val(ral.rxfifo.size, rxfifo));
+        present_in_packet(get_field_val(ral.rxfifo.ep, rxfifo),
+                          get_field_val(ral.rxfifo.buffer, rxfifo),
+                          get_field_val(ral.rxfifo.size, rxfifo),
+                          .may_retract(1'b0),  // TODO: no isochronous support at present.
+                          .retracted(retracted), .retracted_buf(retracted_buf));
+        // Was a packet retracted?
+        if (retracted) begin
+          bit supplied;
+          device_buf_supply(retracted_buf, supplied);
```
rswarbrick, line 357: Again, I'd probably suggest allowing `device_buf_supply` to have a null output for `supplied`, which would avoid needing to make a fake variable here.

hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -103,7 +103,7 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     drive_packet("Token", pkt);
     // Attempt to collect a response only to full speed IN transactions; this should be DATA0|1
     // or ACK/NAK.
-    if (req_item.m_pid_type == PidTypeInToken && !req_item.low_speed) begin
+    if (req_item.m_pid_type == PidTypeInToken && req_item.await_response && !req_item.low_speed) begin
```
rswarbrick, line 106: Line length :-)

rswarbrick (commented):
Lots of nitty comments, but this looks sensible to me.

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
alees24: Oh, no, that final check is comparing the two bitmaps.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: In this specific case, it's safe, although as discussed using 'csr_update' on the `configin_X` registers is very risky.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: I've changed this task to be more explicit.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: As discussed offline, the concept of 'desired' state in UVM seems inadequate to handle what we need to achieve. In the case of '(r)w1c' we want "do not change" which is achieved by writing '0' and seems to necessitate the use of 'csr_wr.'

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: Sorry, the whole of that description is a little garbled, oweing to some confusion about RTL change (actual vs intended behavior), and having two different implementations of that task along the various sequences.
I've restructured, simplified and clarified it now, to describe and return explicitly the three possible outcomes of attempting to retract an IN packet.
This allows us to exercise non-Isochronous packet retraction too at some point.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: There's no real alternative but to have an unused output argument here, sadly.

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
alees24: I've removed the comments about non-iso support; at that point it wasn't clear whether we should exercise all endpoint types in that sequence. Isochronous is the more difficult case, so I built for that.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: Discussed elsewhere. I'm concerned about single-cycle races and the priority of 'set sending' and 'clear sending' when occurring in the same cycle here; it depends upon the detailed implementation of the DUT, and 'csr_wr' is safe.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
alees24: See earlier comment, please.

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
alees24: No longer relevant to this sequence. Let's leave it as isochronous-only.

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
alees24: Added the issue number.

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
alees24: Comment expanded.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
alees24: As discussed, unfortunately the unused output is required.

hw/ip/usbdev/dv/env/seq_lib/usbdev_max_usb_traffic_vseq.sv
alees24: #23932

hw/ip/usbdev/dv/env/seq_lib/usbdev_iso_retraction_vseq.sv
rswarbrick: Duh, I inferring from the names that these were just counters. Sorry: looks very sensible to me!

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
rswarbrick: This makes sense. Thanks!

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
rswarbrick: You're right (and being polite: this was a SystemVerilog error on my side. I hadn't twigged that "null" only makes sense for class types)

rswarbrick (approved):
Thanks for all the tweaks: this looks sensible to me now.

Issue #23930, opened 2024-07-05
[hmac,dv] Rare flakiness in hmac_stress_all


gdessouky:
### Description

`hmac_stress_all` test fails very rarely for some seeds in a non-reproducible way. It is root-caused to probably be due to when the  `cfg.wipe_secret_triggered`  is cleared in https://github.com/lowRISC/opentitan/blob/master/hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv#L211 and how the read digest CSRs are read and compared afterwards. Race conditions may sometimes be occurring between the clearing and reading of the digest CSRs which is the only explanation for the discrepancy between 2 exact logs for 2 exact seeds, where 1 fails and the other passes. 

Looking into it and testing out a fix now.

@martin-velay

gdessouky:
This is a minor DV issue, and no RTL is implicated :)

martin-velay:
Should be closed as PR #23931 has been merged some times back!

PR #23931, opened 2024-07-05
[hmac,dv] Fix clearing of wipe_secret_triggered flag

gdessouky:
This decouples the clearing of the `wipe_secret_triggered` flag from reading the digest CSRs and triggers it where necessary through the smoke sequence. This also captures the previous digest before a transaction is aborted to compare with it in the next transaction.

This should address issue #23930. It is not very reproducible but we will be able to tell eventually from the nightly regressions and also running a `-rx 2` regression locally.


rswarbrick (approved):

hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
```diff
@@ -91,8 +91,9 @@ class hmac_smoke_vseq extends hmac_base_vseq;
                 .key_length(key_length), .intr_fifo_empty_en(intr_fifo_empty_en),
                 .intr_hmac_done_en(intr_hmac_done_en), .intr_hmac_err_en(intr_hmac_err_en));
 
-      // always start off the transaction by reading previous digest to clear
-      // cfg.wipe_secret_triggered flag and update the exp digest val in scb with last digest
+      // always start off the transaction by first clearing cfg.wipe_secret_triggered flag
+      // and update the exp digest val in scb with last digest
+      clear_wipe_secret();
```
martin-velay, line 96: I wonder why do we clear this here? Instinctively I would do it after all of this, at the end of this loop (between l. 237-238).

hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
```diff
@@ -114,9 +118,15 @@ class hmac_smoke_vseq extends hmac_base_vseq;
         `uvm_info(`gfn, $sformatf("wiping before start"), UVM_HIGH)
         wipe_secrets();
         // Here the wipe secret will only corrupt secret keys and current digests.
-        // If HMAC is not enabled, check if digest is corrupted.
+        // If HMAC is not enabled, once we trigger start, the corruption does not affect the digest
+        // that will get computed. If HMAC is enabled, the key got corrupted, so the digest
+        // that will get computed will be corrupted. So only call call clear_wipe_secret()
```
martin-velay, line 123: Here just a little stuff without importance: "call call"

martin-velay (commented):
I have one question but I am not sure if what I am saying is valid. What do you think @gdessouky

hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
gdessouky: I would either clear here or after the loop completes, so after line 238, to account for discarded transactions where we do not trigger process because of invalid configuration, see line 159. This would skip the loop iteration and exit the loop and then start off a new transaction, so we should clear after the loop or with the beginning of every new iteration.

hw/ip/hmac/dv/env/seq_lib/hmac_smoke_vseq.sv
gdessouky: thanks, fixed and pushed now

martin-velay (approved):

vogelpi:
I am merging this as it doesn't affect the FPGA tests. All other tests passed.

PR #23935, opened 2024-07-05
[otbn,dv] Improve RMA handling in model

rswarbrick:
These changes massively improve the pass rate of the `otbn_escalate` test. The first commit tracks whether we're in the middle of a secure wipe (rather than waiting for the next seed) and the second commit uses that flag to figure out when we're getting an unexpected URND ack.

The final commit is just a code cleanup which I noticed because I was touching equivalent code. But it collides with the other two commits and I thought it made the most sense to "tailgate" them.

### DIFF
--- a/hw/ip/otbn/dv/otbnsim/sim/sim.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/sim.py
@@ -428,7 +428,9 @@ class OTBNSim:
                 # Also, set wipe_cycles to an invalid value to make really sure
                 # we've left the wiping code.
                 self.wipe_cycles = -1
+
                 self.state.first_round_of_wipe = True
+                self.state.incomplete_wipe = False
                 self.state.set_fsm_state(next_state)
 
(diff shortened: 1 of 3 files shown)


hw/ip/otbn/dv/otbnsim/sim/sim.py
```diff
@@ -253,9 +253,7 @@ def _step_pre_exec(self, verbose: bool) -> StepRes:
         # Correctly handle an RMA request when when we're still waiting to
```
hcallahan-lowrisc, line 253: Nit
```suggestion
        # Correctly handle an RMA request when we're still waiting to
```

hcallahan-lowrisc (approved):
Changes make sense to me, though I'm definitely not an expert!

hw/ip/otbn/dv/otbnsim/sim/sim.py
rswarbrick: Oops, thanks!

Issue #22471, opened 2024-04-08
[hmac] V2(S) Signoff


andreaskurth:

martin-velay:
### Commits since HMAC D2(S) Signoff (https://github.com/lowRISC/opentitan/issues/20996)
---
`git rev-parse HEAD`

https://github.com/lowRISC/opentitan/commit/ea5bce9ed692af726da4729039c4cbe1fba64ee7

`git log d7483116b80bf56ce2b8e665c679872de3f3794d..HEAD --oneline hw/ip/hmac`

99ce062a64 [hmac, dv] Fix `ValidHmacEnConditionAssert`
01a208901a [hmac,rtl/dv] DV synchronization and error handling fixes
f489ff703d [hmac,rtl] HMAC FSM RTL fixes to handle SAR
7d8ace24cb [dv] Remove default value for wait_to_issue_reset argument
6c51bc0da2 [hmac,dv] DV fixes for block-level tests
ee5f4a1abf [sram_ctrl] Add readback feature
f9f6fe5d06 [hmac,csrs/rtl] Add key_swap field and expose hmac idle status
aace3c0617 [hmac,rtl/dv] RTL and DV fixes for digest loading and error handling
76b1c7498c [hmac,dv] DV synchronization fixes for do_read_check
c1fa3221d0 [hmac,rtl] Minor HMAC fixes for secret key wiping
:...skipping...
99ce062a64 [hmac, dv] Fix `ValidHmacEnConditionAssert`
01a208901a [hmac,rtl/dv] DV synchronization and error handling fixes
f489ff703d [hmac,rtl] HMAC FSM RTL fixes to handle SAR
7d8ace24cb [dv] Remove default value for wait_to_issue_reset argument
6c51bc0da2 [hmac,dv] DV fixes for block-level tests
ee5f4a1abf [sram_ctrl] Add readback feature
f9f6fe5d06 [hmac,csrs/rtl] Add key_swap field and expose hmac idle status
aace3c0617 [hmac,rtl/dv] RTL and DV fixes for digest loading and error handling
76b1c7498c [hmac,dv] DV synchronization fixes for do_read_check
c1fa3221d0 [hmac,rtl] Minor HMAC fixes for secret key wiping
5a74a15fb1 [hmac/rtl] Fix idle_o when told to stop
e71da1ced8 [hmac/rtl] Wait for digest of complete block when stopping
[...]

PR #23939, opened 2024-07-05
[hmac,doc] Move into V2S stage

martin-velay:
To be able to close https://github.com/lowRISC/opentitan/issues/22471


vogelpi:

vogelpi (approved):
This looks good to me. As pointed out by @andreaskurth in https://github.com/lowRISC/opentitan/issues/22471#issuecomment-2211414212 the fifo_empty_status_interrupt testpoint should be mapped and the coverage exclusion for the uncovered but buggy FSM state transition should be added before merging this.

andreaskurth:

PR #23940, opened 2024-07-05
[i2c,dv] I2C Controller-Mode NACK modelling, mapping TP:host_mode_halt_on_nak

hcallahan-lowrisc:
> ~~The first two commits in this PR are squashes of other in-flight PRs, which we depend on here. Ignore for reviewing purposes~~.

Build on the changes in #23911 to start modelling and correctly-predicting transfers with a Target that may NACK our bytes.
The modelling in place in this PR is very rudimentary, and cannot yet make predictions when mid-transfer stimulus is applied, such as software clearing fifo state during NACK events which cause the DUT to halt. This modelling will be further fleshed-out in the future.

Map the test "i2c_host_may_nack" to the TP:host_mode_halt_on_nak.

The description and documentation of the "i2c_reference_model.sv" should be significantly fleshed-out in the future, as it will greatly increase in apparent complexity as we amend it to model and predict more and more DUT behaviours. In addition, we should strive to document under what limitations the model can predict, such as bounds for the validity of certain predictions. It would be greatly desirable to be explicit in capturing these model details.

### DIFF
--- a/hw/ip/i2c/dv/env/i2c_reference_model.sv
+++ b/hw/ip/i2c/dv/env/i2c_reference_model.sv
@@ -66,15 +66,22 @@ class i2c_reference_model extends uvm_component;
   i2c_reg_block ral;
 
   /////////////////////////////////
-  //
-  // Ports
-  // Inputs <- StimulusVseq/Monitor/DUT
-  // Outputs -> Scoreboard
 
   // Inputs
+  // - Note. that the tl_ul monitor inputs (i.e DUT CSR accesses) are input to the model via the
+  //   process_tl_access() method.
 
-  uvm_tlm_analysis_fifo #(i2c_item) target_mode_wr_obs_fifo;  // Input -> Monitor 'analysis_port'
-  uvm_tlm_analysis_fifo #(i2c_item) target_mode_rd_obs_fifo;  // Input -> Monitor 'analysis_port'
+  // Complete transfer items from the i2c_monitor
+  uvm_tlm_analysis_fifo #(i2c_item) controller_mode_wr_obs_fifo;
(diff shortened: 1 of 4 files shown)


hw/ip/i2c/dv/env/i2c_reference_model.sv
```diff
@@ -508,6 +533,56 @@ class i2c_reference_model extends uvm_component;
...
+    wait(inp_xfer.state == StAddrByteAckRcvd);
+
+    `uvm_info(`gfn, $sformatf(
+      "create_controller_exp() got inp_xfer : (address=2'h%2x, rw=1'b%b (%0s))",
+      inp_xfer.addr, inp_xfer.dir, inp_xfer.dir.name()), UVM_HIGH)
+
+    // If the address byte was NACK'd, we may just end up sending a STOP immediately, depending on
+    // if the software chose to clear any FMTFIFO state in response to the NACK.
+    // If not, we loop to capture all the data-byte ACKs from in-progress transfers.
+    fork
```
rswarbrick, line 558: If `inp_xfer.state` becomes `StStopped`, won't the other process carry on running (and trash `data_ack_q`)? Do we need an isolation fork and disable?

rswarbrick (commented):
I've looked at the final few commits and they look sensible. One question about processes.

hcallahan-lowrisc:
Some parts of this PR have been relocated to other PRs, and I want to implement some other things in here differently. Therefore, I'm going to move it to draft until the concerns have been separated properly.

rswarbrick (approved):
This looks good to me (if rather complicated!). Thanks for the cleanups.

hcallahan-lowrisc:
Thanks @rswarbrick for the review.

Issue #22341, opened 2024-03-29
[kmac,dv] Verify `fifo_empty` status interrupt


andreaskurth:
In PR #21657 we changed the `fifo_empty` interrupt from event to status. `kmac_scoreboard` currently doesn't model and check that interrupt; we should correct that.

vogelpi:
Software is currently not using the `fifo_empty` interrupt and we may thus want to consider de-prioritizing this task @andreaskurth .

andreaskurth:
Discussed in triage: same as for the similar HMAC issue, we move this to M5

vogelpi:
This would be good to get. Maybe something for @martin-velay after taking care of the same task for HMAC?

vogelpi:
@martin-velay started working on this and has a first draft PR ready here: #23942. Unfortunately, the work is not yet finished and Martin was OoO this week. Since this issue isn't super critical (see previous comments), I am moving this to M7 now.

martin-velay:
Closed as https://github.com/lowRISC/opentitan/pull/23942 has been merged

PR #23942, opened 2024-07-05
[kmac,dv] Check FIFO empty interrupt

martin-velay:
Linked to this issue #22341

I am not sure it works well yet. I have only checked it for one test, and it seems good from what I've seen.
This check is not modeling the FIFO level, instead it is using the FIFO empty/full from the DUT through a backdoor access. Then it does a prediction on the FIFO empty interrupt as described in the [KMAC register doc](https://opentitan.org/book/hw/ip/kmac/doc/registers.html#intr_state--fifo_empty).

### DIFF
--- a/hw/ip/kmac/dv/env/kmac_scoreboard.sv
+++ b/hw/ip/kmac/dv/env/kmac_scoreboard.sv
@@ -233,7 +233,7 @@ class kmac_scoreboard extends cip_base_scoreboard #(
           // does as many EDN fetches as necessary to fill up the required data bus size
           // of the "host".
           repeat (kmac_reg_pkg::NumSeedsEntropy) begin
-            `DV_SPINWAIT(edn_fifos[0].get(edn_item);, "Wait EDN request")
+            edn_fifos[0].get(edn_item);
           end
           `uvm_info(`gfn, "got all edn transactions", UVM_HIGH)
           set_entropy_fetch(0);
--- a/hw/ip/kmac/dv/env/kmac_scoreboard.sv
+++ b/hw/ip/kmac/dv/env/kmac_scoreboard.sv
@@ -65,6 +65,7 @@ class kmac_scoreboard extends cip_base_scoreboard #(
   bit in_edn_fetch = 0;
 
   // CMD fields
+  bit [KmacCmdIdx:0] kmac_cmd;
   kmac_cmd_e unchecked_kmac_cmd = CmdNone;
   kmac_cmd_e checked_kmac_cmd = CmdNone;
 
@@ -75,8 +76,17 @@ class kmac_scoreboard extends cip_base_scoreboard #(
   bit sha3_absorb;
   bit sha3_squeeze;
 
+  // FIFO status bits
+  bit cmd_process_triggered;
+  bit msgfifo_access;
+  bit fifo_empty_status;
+  bit fifo_full_status;
+  bit fifo_full_detected;
+  bit intr_fifo_empty_allowed;
+
   bit intr_kmac_done;
   bit intr_fifo_empty;
+  bit pre_intr_fifo_empty;
   bit intr_kmac_err;
 
   // Error tracking
@@ -168,6 +178,7 @@ class kmac_scoreboard extends cip_base_scoreboard #(
         process_kmac_app_req_fifo();
         process_kmac_app_rsp_fifo();
         process_sideload_key();
+        manage_fifo_empty_intr();
       join_none
     end
   endtask
@@ -216,19 +227,19 @@ class kmac_scoreboard extends cip_base_scoreboard #(
     forever begin
       wait(!cfg.under_reset);
       `DV_SPINWAIT_EXIT(
-          forever begin
-            @(posedge in_edn_fetch);
-            // Entropy interface is native 32 bits - prim_edn_req component internally
-            // does as many EDN fetches as necessary to fill up the required data bus size
-            // of the "host".
-            repeat (kmac_reg_pkg::NumSeedsEntropy) begin
-              `DV_SPINWAIT(edn_fifos[0].get(edn_item);, "Wait EDN request")
-            end
-            `uvm_info(`gfn, "got all edn transactions", UVM_HIGH)
-            set_entropy_fetch(0);
+        forever begin
+          @(posedge in_edn_fetch);
+          // Entropy interface is native 32 bits - prim_edn_req component internally
+          // does as many EDN fetches as necessary to fill up the required data bus size
+          // of the "host".
+          repeat (kmac_reg_pkg::NumSeedsEntropy) begin
+            `DV_SPINWAIT(edn_fifos[0].get(edn_item);, "Wait EDN request")
           end
-          ,
-          wait(cfg.under_reset);
+          `uvm_info(`gfn, "got all edn transactions", UVM_HIGH)
+          set_entropy_fetch(0);
+        end
+        ,
+        wait(cfg.under_reset);
       )
     end
   endtask
@@ -599,7 +610,6 @@ class kmac_scoreboard extends cip_base_scoreboard #(
 
     string csr_name = "";
 
-    bit msgfifo_access;
     bit share0_access;
     bit share1_access;
 
@@ -612,6 +622,11 @@ class kmac_scoreboard extends cip_base_scoreboard #(
     bit addr_phase_write  = (write && channel == AddrChannel);
     bit data_phase_read   = (!write && channel == DataChannel);
     bit data_phase_write  = (write && channel == DataChannel);
+
+    // Indicates that the msgfifo access is now over. Clear it at the beginning of this transaction
+    // as this flag is used somewhere else and needs to be actually raised for a non-null simulation
+    // time.
+    msgfifo_access = 0;
 
     // if access was to a valid csr, get the csr handle
     if (csr_addr inside {cfg.ral_models[ral_name].csr_addrs}) begin
@@ -757,7 +772,7 @@ class kmac_scoreboard extends cip_base_scoreboard #(
         //
         // ICEBOX - handle error cases
         if (addr_phase_write) begin
-          bit [KmacCmdIdx:0] kmac_cmd = item.a_data[KmacCmdIdx:0];
+          kmac_cmd = item.a_data[KmacCmdIdx:0];
 
           // Handle hash_cnt clear conditions
           if (item.a_data[KmacHashCntClrIdx]) `DV_CHECK(ral.entropy_refresh_hash_cnt.predict(0));
@@ -1093,8 +1108,6 @@ class kmac_scoreboard extends cip_base_scoreboard #(
           `uvm_info(`gfn, $sformatf("msg: %0p", msg), UVM_HIGH)
         end
       end
-      // indicate that the msgfifo access is now over
-      msgfifo_access = 0;
     end
 
     ///////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


martin-velay:
Hi @rswarbrick, @andreaskurth and @vogelpi,
This PR is ready to be reviewed.

Thanks in advance!

vogelpi (approved):
Thanks @martin-velay for your work on this! The PR looks good to me. I think I can follow and understand the code and it looks right to me. But I am obviously no DV expert.

Great catch also on the Xcelium-specific fixes btw!

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -613,6 +623,11 @@ class kmac_scoreboard extends cip_base_scoreboard #(
     bit data_phase_read   = (!write && channel == DataChannel);
     bit data_phase_write  = (write && channel == DataChannel);
 
+    // Indicate that the msgfifo access is now over. Clear it at the begining of the next
+    // transaction as this flag is used somewhere else and needs to be actually raised for
```
rswarbrick, line 627: I think this should probably be "... the beginning of this transaction ..." because you mean the current one, rather than the next call to `process_tl_access`?

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -65,6 +65,7 @@ class kmac_scoreboard extends cip_base_scoreboard #(
   bit in_edn_fetch = 0;
 
   // CMD fields
+  bit [KmacCmdIdx:0] kmac_cmd;
```
rswarbrick, line 68: I think the point of pulling this out is that we want to watch for when it changes in `manage_fifo_empty_intr`? I guess we expect `kmac_cmd` to change when the FIFO becomes empty? That's perfectly reasonable, but it isn't immediately obvious to me why that happens: can you explain?

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+    @(negedge cfg.under_reset);
+    forever begin
+      wait(!cfg.under_reset);
+      fork begin
+        fork
+          begin
+            // Trigger the task only when required to avoid to overload the simulation
+            forever @(msgfifo_access, in_kmac_app, sha3_absorb, kmac_cmd, sha3_idle) begin
+              // Loop also on each clock cycle when intr_fifo_empty_allowed or when msgfifo_access
+              // are high to avoid to miss some FIFO status changes
```
rswarbrick, line 1215: Grammar nit: I think you want "... to avoid missing some ..."

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+          begin
+            wait(cfg.under_reset);
+          end
+        join_any
+        disable fork;
+      end join
+    end
+  endtask : manage_fifo_empty_intr
+
+  virtual task predict_fifo_empty_intr();
```
rswarbrick, line 1237: Can you add a documentation comment above the task that explains roughly what it does? In particular, the timing (I think the task waits until `intr_state` is not busy)

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+            forever @(msgfifo_access, in_kmac_app, sha3_absorb, kmac_cmd, sha3_idle) begin
+              // Loop also on each clock cycle when intr_fifo_empty_allowed or when msgfifo_access
+              // are high to avoid to miss some FIFO status changes
+              do begin
+                // Do checks on the falling edge of the clock to ensure the DUT internal signals
+                // will be up to date
+                cfg.clk_rst_vif.wait_n_clks(1);
+                fork begin
+                  predict_fifo_empty_intr();
+                end join_none
```
rswarbrick, line 1222: If I understand correctly, this will be firing off a new process on each negedge of the clock. Can we restructure slightly so that `predict_fifo_empty_intr` gets split into two pieces? The first could return instantly. This would be enough for updating `intr_fifo_empty_allowed` (which is what we really need in the loop here).

Then we could fire off the second half and update `fifo_empty` outside of this loop?

I think this will also get rid of the second `wait_clks` line in the loop.

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+    end
+
+    // Manage interrupt
+    if (intr_fifo_empty_allowed && fifo_empty_status) begin
+      pre_intr_fifo_empty = 1;
+    end else begin
+      pre_intr_fifo_empty = 0;
+    end
+
+    // Delay FIFO empty signal to be aligned with the DUT behavior
```
rswarbrick, line 1276: I think the restructuring I'm suggesting above might make this a bit easier as well: won't we be able to avoid the `join_none` if we make this race "lexically" with the wait that comes just afterwards?

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+
+    // Detect when FIFO is not empty anymore
+    fifo_empty_status_neg = ~fifo_empty_status & fifo_empty_status_last;
+
+    // Latch Command Process flag
+    if (!cmd_process_triggered && (kmac_cmd == CmdProcess)) begin
+      cmd_process_triggered = 1;
+    end else if (sha3_idle) begin
+      cmd_process_triggered = 0;
+    end
```
rswarbrick, line 1253: I think this if/else block and the next few would all be simpler to read with something like
```systemverilog
    cmd_process_triggered = !cmd_process_triggered && (kmac_cmd == CmdProcess);
```

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+
+    // Check whether FIFO full has been detected for the ongoing message
+    if (fifo_full_status) begin
+      fifo_full_detected = 1;
+    end else if (sha3_idle || fifo_empty_status_neg) begin
+      fifo_full_detected = 0;
+    end
+
+    // Check if conditions are met
```
rswarbrick, line 1262: Which conditions?! :-) (I'd probably just drop the comment)

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+    fifo_empty_status_neg = ~fifo_empty_status & fifo_empty_status_last;
+
+    // Latch Command Process flag
+    if (!cmd_process_triggered && (kmac_cmd == CmdProcess)) begin
+      cmd_process_triggered = 1;
+    end else if (sha3_idle) begin
+      cmd_process_triggered = 0;
+    end
+
+    // Check whether FIFO full has been detected for the ongoing message
```
rswarbrick, line 1255: This block won't necessarily change `fifo_full_detected`, and it's not super obvious when glancing at the code. Assuming that behaviour is what you want, maybe it makes sense to extend the comment to mention that?

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+    end
+
+    // Check if conditions are met
+    if (!in_kmac_app && sha3_absorb && !cmd_process_triggered && fifo_full_detected) begin
+      intr_fifo_empty_allowed = 1;
+    end else begin
+      intr_fifo_empty_allowed = 0;
+    end
+
+    // Manage interrupt
```
rswarbrick, line 1269: Again, I'd probably just drop this comment. If not, it needs expanding a little bit.

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -1179,6 +1192,104 @@ class kmac_scoreboard extends cip_base_scoreboard #(
...
+        intr_fifo_empty = pre_intr_fifo_empty;
+      end
+    join_none
+
+    // Wait needed to avoid race condition for register access
+    wait(!ral.intr_state.is_busy());
+    // Update expected value
+    void'(ral.intr_state.fifo_empty.predict(.value(intr_fifo_empty), .kind(UVM_PREDICT_DIRECT)));
+    // Backdoor read the register and compare against mirrored value
+    csr_rd_check(.ptr(ral.intr_state.fifo_empty), .compare(1), .compare_vs_ral(1), .backdoor(1));
```
rswarbrick, line 1289: This is reasonable, but would it be easier the other way around? I think that `csr_rd` does the prediction anyway, so can't you use that and then check locally that the value matches `intr_fifo_empty`? The timing will be the same, but I think it might be easier to follow.

rswarbrick (commented):
Thanks for this work. It looks generally sensible but (since I'm reading it carefully) I think there are ways the first change could be a bit simpler. Of course, it's always possible that I've missed something: give me a shout if so!

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: Yes, I pulled this out to be able to see it from `manage_fifo_empty_intr`.

The interrupt could be raised only if "_iii) Software has not yet written the Process command to finish the absorption process_", as described in the [KMAC register spec](https://opentitan.org/book/hw/ip/kmac/doc/registers.html#intr_state--fifo_empty).

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: Ah yes, it was the "next" in my mind as it is for the next after this clear, which means the current actually 😅

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: I'll need time to re-run the simulation over the night then, to be sure it's not breaking anything as this was a bit tricky to get it to work

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: I have moved this comment:
```
  // The FIFO empty interrupt is raised only FIFO is actually empty, but also if all of the
  // following conditions are met:
  //   1- The KMAC block is not exercised by a hardware application interface.
  //   2- The SHA3 block is in the Absorb state.
  //   3- Software has not yet written the Process command to finish the absorption process.
  //   4- The message FIFO must also have been full previously. Otherwise, the hardware empties
  //      the FIFO faster than software can fill it and there is no point in interrupting the
  //      software to inform it about the message FIFO being empty.
```
above this task as it is more linked to that one. And about the `intr_state` wait, there is already a comment saying that a we need to wait until this is not accessed anymore, but this race condition is internal to the TB itself.

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: I don't really feel the same, I don't like this way of writing, it makes my brain heating even if the result is the same 😄

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: These conditions:
```
  //   1- The KMAC block is not exercised by a hardware application interface.
  //   2- The SHA3 block is in the Absorb state.
  //   3- Software has not yet written the Process command to finish the absorption process.
  //   4- The message FIFO must also have been full previously. Otherwise, the hardware empties
  //      the FIFO faster than software can fill it and there is no point in interrupting the
  //      software to inform it about the message FIFO being empty.
```
Now, this makes more sense as this comment is just right above this task name.
I'll write:
`// Check if FIFO empty interrupt conditions are met`

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: I'll drop that one

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: Sorry I am not sure I follow you. I don't fully understand our prediction way of doing. Here is what I though/seen, correct me if I am wrong:
- we trigger a read/write from the sequences
- which triggers `process_tl_access`
- we detect if we triggered a read/write operation and do an update of the RAL mirrored value by calling the `predict` method
- then, if the `do_read_check` is to '1' and it was a read operation, we do a check between the actual read value from the register against the mirrored value updated earlier

1. Is that right?
2. Then, I don't really know if a backdoor access will also trigger `process_tl_access`?

If the answer to these 2 questions is "yes", then I have another one:
3. Do you mean that I should remove the predict method call and calling only the `csr_read` task, and do the prediction somewhere within the `process_tl_access` task?

rswarbrick: Duh, good point! I hadn't thought about the fact that this is a backdoor access, so the answer to (2) is "it won't". I completely agree with your bullet points but I thought that my suggested change might be a bit easier to read. Unfortunately, it would also be wrong...

About the final question: (For non-backdoor accesses), there is a call to `predict` at the bottom of the task (line 1178, on master) and that's the call that I was hoping to use.

Your code here looks perfectly sensible to me, and thanks for the reply.

hw/ip/kmac/dv/env/kmac_scoreboard.sv
martin-velay: @rswarbrick 
When you suggest to move the "second half and update `fifo_empty` outside of this loop", I guess you are referring to the `do...while(...)` loop? If yes, this will cause to not check the FIFO full interrupt when it's actually the most important. I think I should keep the check on each clock cycles when `intr_fifo_empty_allowed` or `msgfifo_access` are high.

If that's OK, this PR could be merged now as the other comments have been already addressed.

rswarbrick: I'm not personally a fan of the approach this code takes. In particular, I really don't like the background process we fire off when updating `intr_fifo_empty` (because I worry that it ends up hard to reason about - what happens if there's a reset?) I'm willing to believe the code works, and I'm glad. And I hope I never have to maintain it :-)

Let's forget my restructuring notes: it's your code, not mine! I'm happy that the code looks correct: let's merge it.

rswarbrick (approved):
Thanks for all the responses to the review comments. Let's take this.

rswarbrick:
Merging despite the CI failure: it's pretty clearly an IT issue and not caused by this PR.
```
                Building SW image "//sw/device/tests:plic_sw_irq_test_sim_dv".
                Building "//sw/device/tests:plic_sw_irq_test_sim_dv" on network connected machine.
                Building with command: ./bazelisk.sh build  --define DISABLE_VERILATOR_BUILD=true --//hw/ip/otp_ctrl/data:img_seed=1 //sw/device/tests:plic_sw_irq_test_sim_dv
                Opening zip "/root/.cache/bazelisk/downloads/bazelbuild/bazel-6.2.1-linux-x86_64/bin/bazel (deleted)": open(): No such file or directory
                FATAL: Failed to open '/root/.cache/bazelisk/downloads/bazelbuild/bazel-6.2.1-linux-x86_64/bin/bazel (deleted)' as a zip file: (error: 2): No such file or directory
                make: *** [/azp/agent/_work/1/s/hw/dv/tools/dvsim/sim.mk:73: sw_build] Error 36

```

PR #23943, opened 2024-07-05
[top_earlgrey/dv] Fix `rom_raw_unlock` test

andreaskurth:
Prior to this PR, `chip_sw_lc_raw_unlock_vseq` would try to read a CSR via JTAG without checking if SBA is non-busy.  The `rom_raw_unlock` test failed (for some seeds) because the SBA was indeed busy, which caused an `sbbusyerror` that never got cleared.

With this, the pass rate of `rom_raw_unlock` is 100% (tested with `-rx 10`).


vogelpi (approved):
Nice work, this LGTM! Thanks for debugging and fixing this important test!

andreaskurth:
This PR changes only DV code, for which all CI checks have passed. Merging.

PR #23955, opened 2024-07-08
[i2c,dv] Capture progress of transfer into i2c_item.state field

hcallahan-lowrisc:
> For reviewing purposes, ignore the first commit. It is part of #23911, and will be rebased away when that PR is merged.

This change adds more granular transaction tracking into the i2c_item type. A new `.state` field is added, which is updated by the i2c_monitor as a transfer progresses.
This more-granular measurement allows:
1) For the agent's reactive sequences to monitor in-progress transfers directly, and to drive at the appropriate points.
2) For the I2C predictor/ref_model to make more accurate predictions about inputs to the system mid-transfer. Currently we make no use of this, but it will be built upon to start modelling DUT features more precisely.

### DIFF
--- a/hw/dv/sv/i2c_agent/i2c_agent_pkg.sv
+++ b/hw/dv/sv/i2c_agent/i2c_agent_pkg.sv
@@ -25,6 +25,20 @@ package i2c_agent_pkg;
     HostWait
   } drv_type_e;
 
+  // This enum is used to model the current state of an in-progress i2c transfer
+  typedef enum int {
+    StNone,
+    StStarted,
+    StAddrByte,
+    StAddrByteRcvd,
+    StAddrByteAckRcvd,
+    StDataByte,
+    StDataByteRcvd,
+    StDataByteAckRcvd,
+    StStopped,
+    StAborted
+  } transfer_state_e;
+
   // register values
   typedef struct {
     // derived parameters
--- a/hw/dv/sv/i2c_agent/i2c_item.sv
+++ b/hw/dv/sv/i2c_agent/i2c_item.sv
@@ -10,6 +10,11 @@ class i2c_item extends uvm_sequence_item;
   // Transfer ID
   int                      tran_id;
 
+  // Model the state of an in-progress i2c transfer. This can be used as a cheap proxy for
+  // breaking up the transfer into smaller sequence items (bytes, bits, etc.) when we need to
+  // update our predictions mid-transfer.
+  transfer_state_e         state;
+
   // Address / Direction
   bit [9:0]                addr; // enough to support both 7 & 10-bit target address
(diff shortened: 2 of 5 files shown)


hw/dv/sv/i2c_agent/i2c_item.sv
```diff
@@ -75,6 +80,7 @@ class i2c_item extends uvm_sequence_item;
   `uvm_object_utils_begin(i2c_item)
     `uvm_field_int(stim_id,                       UVM_DEFAULT | UVM_DEC)
     `uvm_field_int(tran_id,                       UVM_DEFAULT | UVM_DEC)
+    `uvm_field_enum(transfer_state_e, state,      UVM_DEFAULT               | UVM_NOCOMPARE)
```
marnovandermaas, line 83: Nit: Alignment looks a bit weird to me here.

hw/dv/sv/i2c_agent/i2c_agent_pkg.sv
```diff
@@ -25,6 +25,20 @@ package i2c_agent_pkg;
...
+    StNone,
+    StStarted,
+    StAddrByte,
+    StAddrByteRcvd,
+    StAddrByteAckRcvd,
+    StDataByte,
+    StDataByteRcvd,
+    StDataByteAckRcvd,
+    StStopped,
+    StAborted
```
marnovandermaas, line 39: It might be good to put a short comment on what each state means. For example, what is the difference between `StAddrByte` and `StAddrByteRcvd`?

marnovandermaas: After reading the next step, maybe this is not necessary. You could just reference the monitor logic.

marnovandermaas (approved):
Looks like a good change to me from a code review. Just a few small comments but none are blocking. I had a look at the CI failures and they don't look related but it would be good for another set of eyes to confirm that before merging.

hw/dv/sv/i2c_agent/i2c_agent_pkg.sv
hcallahan-lowrisc: Thanks Marno, I think the best thing to do here might be a diagram in the documentation with an annotated i2c waveform showing how each enum state lines up with phases of a transfer. I'm going to file as a future improvement and come back to it.

hw/dv/sv/i2c_agent/i2c_item.sv
hcallahan-lowrisc: Ah yes, it aligns with some other items just slightly further down the page :)

hcallahan-lowrisc:
The CI failures are unrelated to this PR. Hence I'm going to merge it now.

Issue #23920, opened 2024-07-04
[i2c,dv] Testbench issues and potential improvements


hcallahan-lowrisc:
### Description

This bucket issue is for tracking potential improvements to the I2C testbench at a block level. This may include more general sweeping comments that apply to the agent as well.

- [ ]  Remove `i2c_scoreboard.sv` checking restrictions where DUT-Target comparisons use limited-scope checking routines that ignore most of the fields in the sequence item. Ideally, we could use the UVM standard routine `A.compare(B)` here, which is used already for the DUT-Controller comparison code.

PR #23957, opened 2024-07-08
[i2c,dv] I2C agent sequences rework for transfer-level items

hcallahan-lowrisc:
> ~~The first two commits in this PR are squashes of other PRs, on which we are dependent. They should both be ignored for reviewing purposes, and will be rebased away when they land.~~
> Without these two commits : 313 insertions(+), 126 deletions(-)

Fixup the i2c_agent's target-mode sequences to use the new transfer-level TLM items from the i2c_monitor (functionality added in #23955). This removes the strange partial-item messages that were passed from monitor to seqr via the 'req_analysis_port', which was inflexible and used the existing i2c_item for yet another alternate purpose. With this change, the sequences can pickup the 'in-progress' transfer item from the monitor, and observe the state of the transaction as it progresses via the `i2c_item.state` field.

The base_seq has been amended to add this monitor-tracking behaviour, and now provides a number of callback routines that derived sequences can use to drive the bus at the correct time. 
The derived target-mode sequences have been cleaned up and reworked to use these new callbacks.

A significant fraction of the LoC changes are comments, to help me and the next person remember what is going on here.

### DIFF
--- a/hw/dv/sv/i2c_agent/seq_lib/i2c_base_seq.sv
+++ b/hw/dv/sv/i2c_agent/seq_lib/i2c_base_seq.sv
@@ -2,66 +2,207 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+// This base_vseq contains common subroutines for the agent to drive stimulus as either an
+// I2C-Controller (bus-initiator, or 'Host' in wider OpenTitan verbiage) or as an I2C-Target
+// (bus-receiver, or OpenTitan 'Device').
+//
+// - (cfg.if_mode == Device) As an I2C-Target, the default behaviour is to operate in a
+//   reactive-agent mode, awaiting the monitor to report observations via the
+//   TLM 'p_sequencer.in_progress_ports', and passing items to the driver based on this.
+// - (cfg.if_mode == Host) #FIXME - This configuration appears unused.
+//
 class i2c_base_seq extends dv_base_seq #(
     .REQ         (i2c_item),
     .CFG_T       (i2c_agent_cfg),
     .SEQUENCER_T (i2c_sequencer)
   );
+
+  ///////////////
+  // VARIABLES //
+  ///////////////
+
+  // Store a handle to the in-progress transfer item captured by the monitor.
+  i2c_item inp_xfer;
+
+  // This queue stores a sequence of items to be passed to the driver for stimulus creation.
+  // Typically, this variable is populated by the entity that creates the sequence before starting
+  // it. It is also possible to push_back() new items piecemeal while the sequence is running.
+  REQ req_q[$];
+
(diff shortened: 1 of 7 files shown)


hw/dv/sv/i2c_agent/seq_lib/i2c_base_seq.sv
```diff
@@ -2,66 +2,208 @@
...
+            fork
+              forever print_inp_xfer_state_changes();
+              send_device_mode_txn();
+            join_any
+            disable fork;
+          end : iso_fork join
+        end
+      end
+      Host: begin // Agent-Controller
+        // Pass a single randomized stimulus item to the driver (#FIXME Unused.)
```
marnovandermaas, line 74: This is the second fixme for the same issue. Maybe remove one or the other?

hw/dv/sv/i2c_agent/seq_lib/i2c_base_seq.sv
```diff
@@ -2,66 +2,208 @@
...
+      default: `uvm_fatal(`gfn, "Shouldn't get here!")
+    endcase
   endtask : body
 
+
+  // This task awaits a new item from the i2c_monitor to arrive in the corresponding TLM fifo.
+  //
+  virtual task get_in_progress_transfer_item(output i2c_item inp_xfer,
+                                             input i2c_analysis_fifo fifo);
+    fork begin: isolation_thread
```
marnovandermaas, line 86: What's the difference between an isolation thread and an isolation fork?

marnovandermaas (commented):
These changes look reasonable to me from code inspection. I'm not approving yet because there are merge conflicts and CI failures.

hcallahan-lowrisc:
Rebased

hw/dv/sv/i2c_agent/seq_lib/i2c_base_seq.sv
hcallahan-lowrisc: I changed the second FIXME to just `NOTE. Unused`. I've also added a note to track this fixme in #23920.

hw/dv/sv/i2c_agent/seq_lib/i2c_base_seq.sv
hcallahan-lowrisc: No difference, other than the name. I've changed the name to `iso_fork` instead, inline with other uses within the agent.

hcallahan-lowrisc:
Latest pushes address review feedback.

marnovandermaas (approved):
Approving this now based on my previous review. The conflicts have been resolved and the CI failures are only in the CW310 ROM test as far as I can see which should be unrelated to these changes.

hcallahan-lowrisc:
CI has some failures, mostly to do with bazel failing to fetch some dependencies, and some flaky CW310 ROM tests. I don't think any of that is related to this PR at all, and no I2C tests are failing. Hence merging now...

PR #23959, opened 2024-07-08
[i2c,dv] I2C chip-level vseq fixes for transfer-level TLM

hcallahan-lowrisc:
> ~~This commit builds upon #23911, #23955 and #23957 to align and fix the chip-level I2C tests with the new transfer-level monitor and agent abstractions. The first 3 commits are squashes of those PRs, so ignore for reviewing purposes, as they will be rebased away.~~
> Ignoring those changes, the true LoC change is : 6 files changed, 348 insertions(+), 236 deletions(-)

This PR aligns the chip-level I2C vseq's with the I2C DV architecture changes in #23911 to communicate at a transfer-level of abstraction. These vseq's, plus the power_virus_test, broke as a result of the original monitor changes, and were temporarily removed from the regression. This PR aligns all of this test code with the new TLM item format, fixing all 3 of the tests, and re-enables them in CI.

The changes to `chip_sw_i2c_tx_rx_vseq.sv` and `chip_sw_i2c_host_tx_rx_vseq.sv` are pretty minor, and are better categorized as cleanups. The changes required to fix the device_tx_rx_vseq were more substantial, and make up the majority of changes in this PR. As with previous PRs, a significant amount of the changes are new comments to explain the test implementations and structures.

More detailed notes can be seen in the commit messages.

Finally, the TLTs and the I2C component of the power_virus test are re-enabled in CI. They were originally disabled in #23911.


hw/top_earlgrey/dv/env/seq_lib/chip_sw_i2c_device_tx_rx_vseq.sv
```diff
@@ -63,24 +81,39 @@ class chip_sw_i2c_device_tx_rx_vseq extends chip_sw_i2c_tx_rx_vseq;
...
+
+    // Wait for TXFIFO to be pre-populated with any data for READ transfers.
+    `DV_WAIT(cfg.sw_logger_vif.printed_log == "Data written to fifo");
+
+    // Drive the i2c_agent to create the test stimulus.
+    drive_i2c_agent_stimulus(byte_count);
+  endtask
+
+
+  virtual function void configure_agent_timing_cfg();
```
marnovandermaas, line 111: I remember seeing this code in https://github.com/lowRISC/opentitan/issues/23957, any way we can reuse that?

marnovandermaas (commented):
From a code review this looks fine. Not approving this yet because of conflicts and CI failures.

hw/top_earlgrey/dv/env/seq_lib/chip_sw_i2c_device_tx_rx_vseq.sv
hcallahan-lowrisc: I don't think that code is duplicated in the other PR, but there are two routines in this PR with that name ; one each for the host and device chip-level vseqs. They are both subtly different, so we can't really get much de-duplication by factoring out the common bits just yet. 
Hopefully that can change in the future though. There is a comment in the base seq (`chip_sw_i2c_tx_rx_vseq`) pointing to #23920, asking if we can make the timing parameter calculations less static and more re-useable. Hopefully this sort of agent-configuration code can be refactored as part of a larger change like that. This could probably also link into #22476, when I get around to completing those changes.

hcallahan-lowrisc:
This is ready now, @marnovandermaas would you mind having another look? Thanks

hw/top_earlgrey/dv/env/seq_lib/chip_sw_i2c_device_tx_rx_vseq.sv
marnovandermaas: Ok, thanks

marnovandermaas (approved):
Approving now based on my previous code review and the since fixed conflicts and CI.

PR #23960, opened 2024-07-08
[otbn,lint] Fix silly Python lint error

rswarbrick:
This was caused by 597247f7614b6dcd661041a161fff19523c87b17 (and causes CI failures)


hcallahan-lowrisc (approved):

eshapira (approved):

rswarbrick:
Huh, the Python error came from #23935 and it looks like CI didn't properly run to completion (no red crosses, and I didn't notice the missing green ticks!).

rswarbrick:
Quick lint *has* now finished, so I think this should be safe to merge.

PR #23962, opened 2024-07-08
[hsmtool] Fix endianness reversal for ECDSA

cfrantz:
OpenTitan's ECDSA implementation expects the R and S components of an ECDSA signature to be in little-endian order.

1. Reverse the R and S components individually rather than reversing the entire signature.


timothytrippel (approved):

mvaquez (approved):

cfrantz:
The CI failure is not related to this change.

PR #23963, opened 2024-07-08
[opentitanlib] preprocess the manifest before signing

vbendeb:
The manifest usage_constraints field includes values which could be ignored by the signature validator, depending on the selector_bits setting.

This patch makes sure that the signer behaves the same, i.e. replaces the values disabled by the selector with the default value.

Tested by successfully starting the Ti50 image on the Cw310 FPGA board.

### DIFF
--- a/sw/host/opentitanlib/src/image/image.rs
+++ b/sw/host/opentitanlib/src/image/image.rs
@@ -11,6 +11,7 @@ use std::io::{Read, Write};
 use std::mem::{align_of, size_of};
 use std::path::{Path, PathBuf};
 use thiserror::Error;
+use zerocopy::AsBytes;
 
 use crate::crypto::ecdsa::{EcdsaPublicKey, EcdsaRawPublicKey, EcdsaRawSignature};
 use crate::crypto::rsa::Modulus;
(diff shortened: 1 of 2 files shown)


vbendeb:
When signing an image make sure that the constraints field values excluded by the selector bits are not fed into the hasher.


More tests might be added, this is mostly an RFC

sw/host/opentitanlib/src/image/image.rs
```diff
@@ -502,9 +506,40 @@ impl Image {
             ..self.borrow_manifest()?.signed_region_end as usize]))
     }
 
+    /// Operates on the altered signed region of the image: the fields of the
+    /// usage_constraint section not enabled by selector bits are set to the
+    /// default value. This matches the behavior of the chip when calculating
+    /// image hash for signature validation.
```
cfrantz, line 512: I think this is better handled with a logging message (ie `log::warn!` or `log::error!`) than trying to correct a user-error in the manifest.

We have tests that intentionally configure bad manifests to make sure that incorrect images are rejected.

vbendeb: The fact that the user needs to change two fields in the manifest instead of one seems a maintenance burden to me, especially since the embedded side ignores the masked off fields.

> We have tests that intentionally configure bad manifests to make sure that incorrect images are rejected.

Will those tests be working even if opentitantool  generates an error in case of this misconfiguration? A better test would be to demonstrate that the contents of the masked off fields are indeed ignored.

@timothytrippel WDYT?

timothytrippel: What @cfrantz is trying to say here is that we need tooling that does _exactly_ what the user asks, not what the tool thinks was "intended" for our ROM E2E tests. Essentially, we have a [suite of ROM E2E tests](https://cs.opensource.google/opentitan/opentitan/+/master:/sw/device/silicon_creator/rom/e2e/sigverify_usage_constraints/BUILD) where we intentionally craft improper manifests or conflicting manifest conditions to test that our ROM handles these scenarios correctly. If we make the change you are proposing, we can no longer test our ROM appropriately.

PR #23965, opened 2024-07-09
[ci, sival] Create a linter for the testplans

AlexJones0:
This PR is an updated version of a previous PR by @engdoreis found at https://github.com/lowRISC/opentitan/pull/20265, with merge conflicts resolved and with support for the `bazel` property of a testplan to be conditionally missing if the `si_stage` property is set to `NA`, as discussed [here](https://github.com/lowRISC/opentitan/pull/20265#discussion_r1541901278). That PR can be closed when this is merged. 

A JSON schema is used with the relevant python library to lint the testplans appropriately. This can be run using the below commands to verify that all testplans pass the linting.

```bash
./util/lint_testplan.py --schema hw/lint/sival_testplan_schema.json --dir hw/top_earlgrey/data/
```
```bash
./util/lint_testplan.py --schema hw/lint/sival_testplan_schema.json --dir hw/top_earlgrey/data/ip
```

Edit: The missing `si_stage` properties in `hw/top_earlgrey/data/standalone_sw_testplan.hjson` have been set to SV4 as suggested by @engdoreis, and support for SV4 and SV5 in the json schema has now been added.

I've also now added a Github actions CI Job to run the linting check on these two directories on any PR that is made to OpenTitan.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -91,6 +91,10 @@ jobs:
   - bash: ci/scripts/clang-format.sh $SYSTEM_PULLREQUEST_TARGETBRANCH
     condition: eq(variables['Build.Reason'], 'PullRequest')
     displayName: clang-format (C/C++ lint)
+    # Validate testplans with schema (json schema)
+  - bash: ci/scripts/validate_testplans.sh
+    condition: eq(variables['Build.Reason'], 'PullRequest')
+    displayName: Validate testplans with schema
     # Use clang-format to check C/C++ coding style
   - bash: ci/scripts/rust-format.sh $SYSTEM_PULLREQUEST_TARGETBRANCH
     condition: eq(variables['Build.Reason'], 'PullRequest')
--- a/ci/scripts/validate_testplans.sh
+++ b/ci/scripts/validate_testplans.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Use a JSON schema to validate the testplan hjson files have the correct structure.
+
+set -e
+
+dirs_with_testplan_files=(
+    "hw/top_earlgrey/data"
+    "hw/top_earlgrey/data/ip"
+)
+testplan_schema="hw/lint/sival_testplan_schema.hjson"
+
+for dir in "${dirs_with_testplan_files[@]}"; do
+    echo "Validating testplans under $dir:"
+    util/validate_testplans.py --dir "$dir" --schema "$testplan_schema" || {
+        echo -n "##vso[task.logissue type=error]"
+        echo "Failed testplan validation in ${dir}."
+        echo "::error::Failed testplan validation in ${dir}."
+        exit 1
+    }
+done
+
+exit 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,8 @@ jobs:
         run: ./ci/scripts/python-lint.sh "$GITHUB_BASE_REF"
       - name: Python (mypy)
         run: ./ci/scripts/mypy.sh
+      - name: Validate testplans with schema
+        run: ./ci/scripts/validate_testplans.sh
       - name: C/C++ formatting
         run: ./ci/scripts/clang-format.sh "$GITHUB_BASE_REF"
       - name: Rust formatting
--- a/ci/jobs/quick-lint.sh
+++ b/ci/jobs/quick-lint.sh
@@ -49,6 +49,9 @@ ci/scripts/python-lint.sh $tgt_branch || {
 echo -e "\n### Run Python lint (mypy)"
 ci/scripts/mypy.sh $tgt_branch
 
+echo -e "\n### Validate testplans with schema."
+ci/scripts/validate_testplans.sh
+
 echo -e "\n### Use clang-format to check C/C++ coding style"
 ci/scripts/clang-format.sh $tgt_branch
 
--- a/hw/lint/sival_testplan_schema.hjson
+++ b/hw/lint/sival_testplan_schema.hjson
@@ -58,6 +58,19 @@
             "type": "array"
           }
         },
+        // If the si_stage is not tagged NA, then require the bazel property be present.
+        "if": {
+          "properties": {
+            "si_stage": {
+              "not": {
+                "const": "NA"
+              }
+            }
+          }
+        },
+        "then": {
+          "required": ["bazel"]
+        },
         "required": ["name", "desc", "stage", "si_stage", "tests"]
       }
     }
--- a/hw/lint/sival_testplan_schema.hjson
+++ b/hw/lint/sival_testplan_schema.hjson
@@ -1,0 +1,65 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "additionalProperties": false,
+    "testpoints": {
+      "type": "array",
+      "items": {
+        "additionalProperties": false,
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "desc": {
+            "type": "string"
+          },
+          "stage": {
+            "enum": ["V1", "V2", "V2S", "V3"]
+          },
+          "si_stage": {
+            "enum": ["SV1", "SV2", "SV3", "SV4", "SV5", "NA", "None"]
+          },
+          "bazel": {
+            "type": "array"
+          },
+          "boot_stages": {
+            "type": "array",
+            "items": {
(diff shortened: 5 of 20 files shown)


jwnrt (commented):
Thanks for working on this

I ran this command:

```sh
./util/lint_testplan.py \
    --dir hw/top_earlgrey/data \
    --schema hw/lint/sival_testplan_schema.json
```

But I had to add `bazel: []` tags to `standalone_sw_testplan.hjson`, and tags and `si_stage`s to `chip_testplan.hjson` (as well as fix a `si_stage` which was `SV4`) to make it pass.

Maybe @engdoreis will know what `si_stages` to set the missing values.

I think we also want this script to run in CI (probably in `quick_lints`) to prevent regressions

util/lint_testplan.py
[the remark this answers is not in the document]
AlexJones0: Yes, I was running this exclusively through Python so never thought to check the actual permissions on the python script. Thanks for pointing it out

AlexJones0:
> But I had to add `bazel: []` tags to `standalone_sw_testplan.hjson`, and tags and `si_stage`s to `chip_testplan.hjson` (as well as fix a `si_stage` which was `SV4`) to make it pass.

Thanks for pointing this out - it seems that we were only ever running the linting on `hw/top_earlgrey/data/ip` and did not realise there were testplans defined outside of the `ip` folder, so I'll fix this up along with the other stuff

AlexJones0:
> But I had to add `bazel: []` tags to `standalone_sw_testplan.hjson`, and tags and `si_stage`s to `chip_testplan.hjson` (as well as fix a `si_stage` which was `SV4`) to make it pass.

Thanks - the linting errors in the `hw/top_earlgrey/data` directory have now also been fixed. 

> Maybe @engdoreis will know what `si_stages` to set the missing values.

After discussion we decided to set these as SV4 for now, and also add json schema support for both SV4 and SV5 to support future versioning.

> I think we also want this script to run in CI (probably in `quick_lints`) to prevent regressions

After some further discussion with @engdoreis I've added a new commit to add a github actions job that is run on PR to perform the linting. I can then make a separate PR with the ability to run through Bazel. Do you still think it should be added to `quick_lints` or is the Github action sufficient for now?

AlexJones0:
> If a quick lint fails, we block CI and don't move onto the more expensive FPGA or simulator tests. We can't do that blocking if some things are in GitHub Actions and some are in Azure.

Understood, I'll change that to a quick lint instead.

AlexJones0:
The most recent force push to this PR addresses the feedback gathered so far:
  - The PR has been rebased and linting has been applied to the most recent testplan changes.
  - Testplan linting has been incorporated as an Azure CI job, and added to `quick-lint.sh`, instead of using Github Actions.
  - The testplan schema has been converted to a `hjson` and given a comment and license header.
 
Again, you can test this PR using the following commands:
```sh
./util/lint_testplan.py --schema hw/lint/sival_testplan_schema.hjson --dir hw/top_earlgrey/data
./util/lint_testplan.py --schema hw/lint/sival_testplan_schema.hjson --dir hw/top_earlgrey/data/ip
```
or just by running 
```sh
./ci/scripts/testplans-lint.sh
```

@jwnrt @engdoreis for visibility now that I've updated this

util/lint_testplan.py
```diff
@@ -6,8 +6,8 @@
 """
 This script lint the various opentitan testplan.hjson
 usage:
-    ./util/lint_testplan.py --schema hw/lint/sival_testplan_schema.hjson \
-                            --dir hw/top_earlgrey/data/ip/
+    python3 ./util/lint_testplan.py --schema hw/lint/sival_testplan_schema.json \
```
rswarbrick, line 9: Is the name change intentional here? I think we might have lost an "h".

azure-pipelines.yml
```diff
@@ -91,6 +91,10 @@ jobs:
   - bash: ci/scripts/clang-format.sh $SYSTEM_PULLREQUEST_TARGETBRANCH
     condition: eq(variables['Build.Reason'], 'PullRequest')
     displayName: clang-format (C/C++ lint)
+    # Use json schema to check testplans formatting
```
rswarbrick, line 94: Isn't this going to run the script twice? Do we really need it in both `azure-pipelines.yml` and `quick-lint.sh`?

AlexJones0: From my understanding, `ci/scripts/quick-lint.sh` itself is not invoked as part of the quality (quick lint) pipeline. Instead, `quick-lint.sh` is just a script that can be used by developers to run linting locally before they PR (albeit, it has been broken for a while which I have a fix for in #24527, so it doesn't seem to be used much). Hence the comment to "keep them in sync" in this file. Please let me know if any part of my understanding is wrong here.

util/lint_testplan.py
AlexJones0: Good spot! Small oversight here on my part while rebasing and this slipped through.

azure-pipelines.yml
```diff
@@ -91,6 +91,10 @@ jobs:
...
+  - bash: ci/scripts/testplans-lint.sh
+    condition: eq(variables['Build.Reason'], 'PullRequest')
+    displayName: testplans lint (json schema)
```
nbdd0121, line 97: I think it's better to say the intention instead of implementation
```suggestion
    displayName: Validate testplans against schema
```

I general I think a lot of occurance of "lint" in this PR actually means validate.

AlexJones0: I agree with your comment about intention vs implementation, I'll make that change.

I also agree with your suggestion about the semantics of `lint` vs `validate` - what we are doing with the JSON schema here is a lot closer to validation than it is linting. Do you want me to just update the comments / display names, or do you think its worth replacing all references to `lint` in this PR with `validate` or some equivalent, for the sake of readability?

engdoreis (approved):

azure-pipelines.yml
nbdd0121: Ideally replace all references/file names.

AlexJones0:
I've just pushed changes to replace all mentions of `lint` with `validate`, and have changed the display name and comments of the azure job step as suggested by feedback. There should have been no changes to functionality in this last force push, CI notwithstanding.

ci/scripts/testplans-validate.sh
nbdd0121, line 1: Why is the util script called `validate_testplan.py` but this `testplans-validate.sh`? Better to use the same name.

azure-pipelines.yml
```diff
@@ -91,6 +91,10 @@ jobs:
   - bash: ci/scripts/clang-format.sh $SYSTEM_PULLREQUEST_TARGETBRANCH
     condition: eq(variables['Build.Reason'], 'PullRequest')
     displayName: clang-format (C/C++ lint)
+    # Validate testplans with schema (json schema)
+  - bash: ci/scripts/testplans-validate.sh
+    condition: eq(variables['Build.Reason'], 'PullRequest')
+    displayName: Validate testplans with schema
```
nbdd0121, line 97: @jwnrt recently added quick lint for GitHub CI. Maybe this job should be added there instead?

ci/scripts/testplans-validate.sh
AlexJones0: I think this is just a hold-over from the original PR, so I'm not sure if there was any specific reason why these were named this way. I'm happy to change it, I guess to `validate_testplans.{py,sh}` is probably the most suitable name.

azure-pipelines.yml
AlexJones0: Sure, this change was before that was added, I'm happy to look at moving this to the Github Actions CI instead.

(Ironically, I made a Github Actions workflow the first time I made this PR, but then changed it to an Azure job as we weren't using Github Actions for the quick lint job at the time :sweat_smile:)

ci/scripts/validate_testplans.sh
```diff
@@ -0,0 +1,25 @@
...
+dirs_with_testplan_files=(
+    "hw/top_earlgrey/data"
+    "hw/top_earlgrey/data/ip"
+)
+testplan_schema="hw/lint/sival_testplan_schema.hjson"
+
+for dir in "${dirs_with_testplan_files[@]}"; do
+    echo "Validating testplans under $dir:"
+    util/validate_testplans.py --dir "$dir" --schema "$testplan_schema" || {
+        echo -n "##vso[task.logissue type=error]"
```
nbdd0121, line 19: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message

AlexJones0: Ah, I see, thanks for the link. I wonder if I should retain this `##vso...` message for the Azure pipeline job to function correctly, and just add an additional error message for Github actions then? Something like:

```sh
echo -n "::error ::Failed testplan validation in ${dir}."
```

I wonder then if we need to change this in other scripts for the Github Actions CI to work as expected? See for example https://github.com/lowRISC/opentitan/blob/b120a77b03a706f32d64fdca6da429210b782252/ci/scripts/python-lint.sh#L48 and https://github.com/lowRISC/opentitan/blob/b120a77b03a706f32d64fdca6da429210b782252/ci/scripts/verible-lint.sh#L50 which I was looking at when I made this script originally.

AlexJones0: Great, thanks - that's very helpful.
I've added that now :+1:

nbdd0121 (approved):

nbdd0121:
The CI failure is not related.

PR #23966, opened 2024-07-09
[otbn,dv] Refactor FSM states in otbn simulator

### LINKED ISSUE #20835: [otbn] Correct model for handling staggered rma_req_i transitions
### Description

If the RMA signal (`lc_rma_req_i`) has an invalid value, this can cause OTBN to go into a locked state. I think the intention was that any value except for `MuBi4True` would be treated as `MuBi4False`.

The lock happens because the start/stop control sees its `rma_req_i` signal get an invalid value. This gets mirrored to `rma_ack_d`, then `rma_ack_q`. But the invalid contents of `rma_ack_q` then cause the state to go to `OtbnStartStopStateLocked`.

To see this going wrong, run
```
  util/dvsim/dvsim.py hw/ip/otbn/dv/uvm/otbn_sim_cfg.hjson -i otbn_escalate --fixed-seed=2
```
with the current head of the integrated-dev branch (df396017ee).

I suspect a correct fix would be to weaken the check for invalid `rma_ack_q` that's currently on line 373 in that file.

rswarbrick:
These changes are to get the `otbn_escalate` test working properly: there was a rather "bodged" model of how secure wipe should work and it didn't match the RTL (or make sense!)

The important change is the penultimate one, which splits the "wiping" state into two pieces. This means that we correctly track "waiting for seed" before "performing the wipe", which could end up being tracked backwards in some situations before.

But this doubles the number of "wiping" states and (many years ago) I unwisely differentiated between WIPING_GOOD and WIPING_BAD to distinguish whether we expected to lock when the wipe was complete. Rather than adding *two* more states, I've merged those two states back together (in the "Merge WIPING_GOOD, WIPING_BAD" commit) so the net change is zero!

Fixes #20835 (phew!)

### DIFF
--- a/hw/ip/otbn/dv/otbnsim/sim/sim.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/sim.py
@@ -183,12 +183,10 @@ class OTBNSim:
         if should_zero and new_zero:
             self.state.ext_regs.write('INSN_CNT', 0, True)
 
-        # If we are IDLE and rma_req is not a valid lc_ctrl signal, we
-        # expect a MUBI error inside the machine which causes us to jump to
-        # the locked state.
-        if self.state.rma_req == LcTx.INVALID and not is_locked:
+        if self.state.delayed_lock:
             self.state.set_fsm_state(FsmState.LOCKED)
             self.state.ext_regs.write('STATUS', Status.LOCKED, True)
+            is_locked = True
 
         # If we are IDLE and get an RMA request, we want to start a secure
         # wipe, which will eventually put us into the LOCKED state.
@@ -197,7 +195,7 @@ class OTBNSim:
             self.state.set_fsm_state(FsmState.WIPING_BAD)
             self.state.wipe_rounds_done = 0
 
-        if self.state.init_sec_wipe_is_running():
+        if self.state.init_sec_wipe_is_running() and not is_locked:
             # Wait for the URND seed. If there is some genuine state to wipe
             # (because self.state.has_state_to_wipe is true), then we switch to
             # WIPING_GOOD unless the FSM state is already LOCKED, in which case
@@ -417,15 +415,24 @@ class OTBNSim:
                 if self.state.wsrs.URND.running:
                     self.state.set_fsm_state(self.state.get_fsm_state())
             else:
+                # Set a flag that tells us to lock when we get to idle. This
+                # matches the cycle of delay that happens between the end of a
+                # secure wipe where we see an invalid RMA request and the
+                # status getting locked.
+                if self.state.rma_req != LcTx.OFF:
+                    self.state.delayed_lock = True
+
                 # This is the second wipe round or the only wipe round during
-                # an RMA request. If the wipe was good, set the next state to
-                # IDLE; otherwise set it to LOCKED.
-                if not locking:
+                # an RMA request. If there is an invalid RMA signal or we were
+                # in the WIPING_BAD state (so had the locking signal set) then
+                # we should lock. Otherwise, jump to IDLE.
+                if locking:
+                    next_state = FsmState.LOCKED
+                    self.state.ext_regs.write('STATUS', Status.LOCKED, True)
+                else:
                     next_state = FsmState.IDLE
                     if self.state.init_sec_wipe_is_running():
                         self.state.complete_init_sec_wipe()
-                else:
-                    next_state = FsmState.LOCKED
 
                 # Also, set wipe_cycles to an invalid value to make really sure
                 # we've left the wiping code. Leave wipe_rounds_done unchanged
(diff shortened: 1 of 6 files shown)


rswarbrick:
The latest force-push should sort out CI. It turns out that the previous version of my tweaking of `standalonesim.py` seeded URND on every cycle (with the same seed). It turns out that there are some crypto tests that fail if you have a constant random number generator! Should now be fixed.

hw/ip/otbn/dv/otbnsim/sim/sim.py
hcallahan-lowrisc, line 1: Nit. Commit message for 06c9386d0344af88974989079fcbe1133db1e290

hw/ip/otbn/dv/otbnsim/sim/sim.py
```diff
@@ -338,14 +338,21 @@ def _step_exec(self, verbose: bool) -> StepRes:
...
 
-        # Correctly handle an RMA request when we're already wiping by ensuring
-        # we're in the WIPING_BAD state (since we're going to lock when we're
-        # done)
-        if ((self.state.get_fsm_state() == FsmState.WIPING_GOOD and
-             self.state.rma_req == LcTx.ON)):
+        is_good = self.state.get_fsm_state() == FsmState.WIPING_GOOD
+        locking = self.state.rma_req == LcTx.ON or not is_good
+
+        # If there is ann RMA request, make sure we're in the WIPING_BAD state
```
hcallahan-lowrisc: Nit. ann

hw/ip/otbn/dv/otbnsim/sim/sim.py
```diff
@@ -428,8 +430,8 @@ def _step_wiping(self, verbose: bool) -> StepRes:
 
                 # This is the second wipe round or the only wipe round during
                 # an RMA request. If there is an invalid RMA signal or we were
-                # in the WIPING_BAD state (so had the locking signal set) then
-                # we should lock. Otherwise, jump to IDLE.
+                # in the going to lock after wipe (so had the locking signal
```
hcallahan-lowrisc: Nit. grammar here is a bit off

hw/ip/otbn/dv/otbnsim/sim/state.py
```diff
@@ -47,15 +48,15 @@ class FsmState(IntEnum):
...
+    instructions.
 
-    WIPING_GOOD and WIPING_BAD represent the time where we're performing a
-    secure wipe of internal state (ending in updating the STATUS register to
-    show we're done). The difference between them is that WIPING_GOOD goes back
-    to IDLE and WIPING_BAD goes to LOCKED.
+    WIPING represents the time where we're performing a secure wipe of internal
+    state (ending in updating the STATUS register to show we're done). The
+    difference we then goes back to IDLE or go to LOCKED (depending on the
+    lock_after_wipe flag).
```
hcallahan-lowrisc: Nit. grammar

hw/ip/otbn/dv/otbnsim/sim/state.py
```diff
@@ -125,10 +125,13 @@ def __init__(self) -> None:
         self.invalidated_imem = False
 
         # This is the number of cycles left for wiping. When we're in the
-        # WIPING_GOOD or WIPING_BAD state, this should be a non-negative
-        # number. Initialise to -1 to catch bugs if we forget to set it.
+        # WIPING state, this should be a non-negative number. Initialise to -1
+        # to catch bugs if we forget to set it.
         self.wipe_cycles = -1
 
+        # This controls which state we move to when the next secure wipe
```
hcallahan-lowrisc, line 130: ... ends?

hw/ip/otbn/dv/otbnsim/sim/sim.py
```diff
@@ -350,39 +390,17 @@ def _step_wiping(self, verbose: bool) -> StepRes:
         is_good = not self.state.lock_after_wipe
         locking = self.state.rma_req == LcTx.ON or not is_good
 
-        # If there is ann RMA request, make sure we're in the WIPING state and
-        # set lock_after_wipe (since we're going to lock when we're done)
-        if self.state.rma_req == LcTx.ON and is_good:
-            self.state.set_fsm_state(FsmState.WIPING)
+        # If there is ann RMA request, ensure lock_after_wipe is set (since
```
hcallahan-lowrisc, line 393: Nit. ann

hcallahan-lowrisc (approved):
The code looks reasonable to me, so I'm going to approve it. However, I'm not familiar enough with the system to confirm the new model matches the spec/intent. Given that this is about closing test corner-cases against the implementation which has been reviewed, discussed and documented, I think it's probably fine to merge.
Just some comment nits. This looked very tricky to get right! Nice job @rswarbrick :)

rswarbrick:
Thanks for the careful review, and for spotting the slightly spotty grammar from me!

PR #23967, opened 2024-07-09
[otbn,dv] Tidy up the types slightly in otbnsim

rswarbrick:
No functional change, but this avoids some warnings from mypy (which I saw when tracking down a genuine bug in a change I was trying to make).


rswarbrick:
The lack of this change is causing a mypy lint check to fail for #23966, so I'm moving it there (after fixing the missing blank line that caused lint to fail here). Closing this one now that it is moved.

PR #23968, opened 2024-07-09
[otbn,dv] Explicitly drop pending start request if we lock

rswarbrick:
This was occasionally a problem in the otbn_partial_wipe test, where we could see a start request which (correctly) got ignored because the model was locking itself instead.


eshapira (approved):
LGTM, thanks!

PR #23973, opened 2024-07-09
[rv_dm,dv] Don't use late_debug_enable through a CSR if scb disabled

rswarbrick:
Doing so breaks the rv_dm_tap_fsm_rand_reset test because it wants to check all our registers (including late_debug_enable) have their reset values.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -73,6 +73,14 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     late_debug_enable == 1;
   }
 
+  // If we are running without a scoreboard then avoid setting the late_debug_enable register to a
+  // value other than the default (false). The reason is that we have standard CSR check sequences
+  // that, when there is no scoreboard, assert the value matches the reset value. This doesn't work
+  // if we call set_late_debug_enable_with_reg(1) from dut_init.
+  constraint no_reg_late_debug_enable_when_no_scb_c {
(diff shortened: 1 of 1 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
```diff
@@ -73,6 +73,14 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     late_debug_enable == 1;
   }
 
+  // If we are running without a scoreboard then avoid setting the late_debug_enable register to a
+  // value other than the default (false). The reason is that we have standard CSR check sequences
+  // that, when there is no scoreboard, assert the value matches the reset value. This doesn't work
+  // if we call set_late_debug_enable_with_reg(1) from dut_init.
+  constraint no_reg_late_debug_enable_when_no_scb_c {
+    reg_late_debug_enable == 1'b0 || cfg.en_scb == 1'b1;
```
hcallahan-lowrisc, line 81: Throwaway comment : I might have written this as `(!cfg.en_scb) -> (!reg_late_debug_enable)`. I think that's equivalent, and it reads along with the first sentence of the comment a bit more intuitively IMO. But feel free to ignore me :)

hcallahan-lowrisc (approved):

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
rswarbrick: Hah! I think I wrote it as `|->` at first (which doesn't work because the syntax is for a sequence property rather than an expression) and I was lazy so didn't work out the correct rewording... Will fix it.

PR #23975, opened 2024-07-09
[opentitantool] Format `key show` for use in OTP configs

cfrantz:
Because the OTP creation tool uses python `int` for parsing integers and writing those values into the OTP configuration, some endianness transformations are needed to format key material correctly for OTP configurations.


timothytrippel (approved):

cfrantz:
The CI failure is not related to this change.

PR #23978, opened 2024-07-10
[hmac,dv] Extend timeouts

gdessouky:
This extends the timeout for some HMAC tests that time out and fail in the last nightly regression, and after some local runs by @martin-velay to try and determine the right timeout for the test vector tests.

### DIFF
--- a/hw/ip/hmac/dv/hmac_sim_cfg.hjson
+++ b/hw/ip/hmac/dv/hmac_sim_cfg.hjson
@@ -83,7 +83,7 @@
     {
       name: hmac_error
       uvm_test_seq: hmac_error_vseq
-      run_opts: ["+test_timeout_ns=300_000_000"]
+      run_opts: ["+test_timeout_ns=500_000_000"]
     }
 
     {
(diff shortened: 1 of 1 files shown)


gdessouky:
I'm also not sure we need to reseed the test vectors tests at all actually. It seems we run all the test vectors anyway for any iteration; we are not currently randomly picking some of the test vectors for every seed and I don't see any other randomization actually happening. What do we gain from reseeding? Can't we just run 1 seed? 
@martin-velay @rswarbrick

rswarbrick (approved):
Obviously, it would be nice to make the tests faster(!) but this looks perfectly reasonable to me. Thanks.

rswarbrick:
Do you mean the tests where we're currently running 5 seeds? I guess that's orthogonal to the question of how long each seed takes (which you are changing here). But burning less energy is always nice :-D

gdessouky:
> Do you mean the tests where we're currently running 5 seeds? I guess that's orthogonal to the question of how long each seed takes (which you are changing here). But burning less energy is always nice :-D

yes, it's orthogonal, we need the timeout extensions anyway. But wanted to also bring this up here before I change the seeding too (in maybe a followup forced push). Looking into the tests better, I'm not sure I see the use for reseeding 5 times when each run is the same except that the test vectors are shuffled so they are in a different order, but they're all fed in each time, unless I'm missing something.

rswarbrick:
I think I probably agree (but will wait to check that Martin feels the same). Let's decide on that in a separate PR to avoid holding up this improvement.

gdessouky:
> I think I probably agree (but will wait to check that Martin feels the same). Let's decide on that in a separate PR to avoid holding up this improvement.

sounds good, thx for my probably quickest merge ever ;), will spin out an issue to keep track.

PR #23799, opened 2024-06-25
[rv_dm,dv] Tidy up randomisation of "enable signals" in base vseq

rswarbrick:
This PR looks a bit massive, but it actually builds on some existing PRs.
- #23750: This contains commits 1,2
- #23792: This contains commits 3,4
- #23793: This contains commit 5

The remaining commits are:
- [rv_dm,dv] Explicitly control lc_dft_en_i
- [rv_dm,dv] Configure scanmode as a bit instead of a mubi4_t
- [rv_dm,dv] Config lc_hw_debug_en as a bit, not a 4-bit variable
- [rv_dm,dv] Get rid of unnecessary re-randomisation in base vseq

They have detailed commit messages, but the overriding aim is to simplify the way the underlying `rv_dm_base_vseq` controls the enable/disable of debug features.

PR #23981, opened 2024-07-10
[rv_dm,dv] Use the "right enable" in *_jtag_dmi_debug_disabled_vseq

rswarbrick:
This PR looks reasonably large because it depends on #23799. Once that has been merged, there are two remaining commits:
- *[rv_dm,dv] Make DMI bringup in base vseq depend on pinmux connection*
- *[rv_dm,dv] Use the "right enable" in \*_jtag_dmi_debug_disabled_vseq*

The first commit is a tidyup in the base vseq (which I noticed when trying to fix the vseq for the second commit). The second commit is the main point of this PR. Its commit message is as follows:

*[rv_dm,dv] Use the "right enable" in \*_jtag_dmi_debug_disabled_vseq*

The code here was a bit broken and I clearly hadn't previously
understood the difference between "pinmux enabled" and "debug
enabled".

The *whole point* of this test is supposed to be that DMI transactions
get blocked when pinmux is not enabled. After the changes here, we
actually test that property!

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -171,7 +171,7 @@ class rv_dm_base_vseq extends cip_base_vseq #(
 
     // TODO: Randomize the contents of the debug ROM & the program buffer once out of reset.
 
-    if (lc_hw_debug_en) begin
+    if (pinmux_hw_debug_en) begin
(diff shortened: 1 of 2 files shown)


vogelpi (approved):
This looks good as well, thanks @rswarbrick .

rswarbrick:
Thanks. I'm going to force-push after rebasing (because a bunch of CI(private) tests failed last time for unrelated IT issues. Assuming things look reasonable, I intend to merge this when they finish. (But it's an evening with children! If someone else gets here first, please feel free to hit the merge button! :-D)

Issue #22326, opened 2024-03-29
[entropy_src/dv] Extend `entropy_src_err_vseq` to test hardened FIFO pointers


vogelpi:
### Description

For M2 we've also enabled hardening for the `prim_fifo_sync` primitives used in ENTROPY_SRC (the hardening was added after reaching D2S for ES).

While enabling the countermeasure and connecting the alert signals is straight forward, we still should test the countermeasures in DV for M5. Ideally, we can just slightly extend the `entropy_src_err_vseq` sequence which already tests some FIFO related error conditions using signal forcing.

vogelpi:
This one should be straightforward but we also gain much here. Thus lowering priority.

PR #23984, opened 2024-07-10
[entropy_src/dv] Test hardened FIFO errors

h-filali:
This PR adds the new hardened FIFO errors to the error test sequence.

This resolves [#22326](https://github.com/lowRISC/opentitan/issues/22326)

While writing the code for this commit I discovered that the hardened counter errors don't cause the error state bits for FIFO state errors to go high. We might want to discuss whether this is intended or not.

Other than that the test results look good.
I also added coverage for the newly covered errors. However I counted them as state errors, so depending on what we decide here I need to change that.


h-filali:
@vogelpi thanks for your review. I can confirm that it only happens for the distr FIFO. I just discovered the same thing. Based on this I added a small change to this PR which also checks the `FIFO_STATE_ERR` field.

Issue #19036, opened 2023-06-23
[spi host] Test plan - 2.5.2 Remaining items


engdoreis:
### Description

This issue details what needs to be tested for the next deadline, and complements the previous [test plan](https://github.com/lowRISC/opentitan/issues/18710).

- [ ] **Features**:
  - [x] Byte-count.
  - [ ] ~32 bit big or little endian.~ --> It only make sense to be tested in DV as it is a RTL parameter.
  - [ ] ~Max speed of 30 Mhz.~ --> Due to environment constraints, It can be tested beyond 16MHz.
  - [x] Polarity phases (CPOL, CPHA) --> lower priority as there's only one mode used in practice.
     - [x] CPOL = 1
     - [x] CPHA = 1
  - [ ] Chip select automatic control.
    - [x] CSNLEAD, CSNTRAIL and CSNIDLE. (expect this to be covered in DV)
    - [ ] ~Mix transactions with two different devices with different clocks~.--> The spi_host only supports one csb.
    - [ ] ~Different number of CSBs~ --> The spi_host only supports one csb.

msfschaffner:
@engdoreis did we make any progress on this or the status still accurate?

engdoreis:
Yes, the status is accurate. Unfortunately, I haven't spent time on this, so no progress.

andreaskurth:
@engdoreis: Do you think you'll be able to complete this for M4? If not, which items do you think are critical to get tested for Earlgrey-PROD?

engdoreis:
It's not prioritized for SiVal, so unless @moidx changes the priority, it won't be completed before M4.

andreaskurth:
Discussed in triage: okay to move to M5

moidx:
@johngt  to follow up with the software team to update estimates and implementation plan.

johngt:
@luismarques - as discussed - believe we should see if any of the integration tests are failing on the BoBs, particularly around features that have changed in Spi_host. If they are all ok, suggest closing off this issue.
@a-will has suggested that we should have a task in the future to see if we can test either in FPGA or on a modified board to see if we can test multi-controller mode and the other features he worked on.

a-will:
> @a-will has suggested that we should have a task in the future to see if we can test either in FPGA or on a modified board to see if we can test multi-controller mode and the other features he worked on.

A little mix-up here--This note is for i2c, not spi_host.

johngt:
[...]

PR #23985, opened 2024-07-10
[Sival, Spi host]  Config test CPOL and CPHA

engdoreis:
### test:
1. Send four packets from the host to the device, each one with a different configuration for cpol and cpha.
2. The device should receive the packets and be able to decode them with the appropriated configuration.

Fix https://github.com/lowRISC/opentitan/issues/19036
Fix https://github.com/lowRISC/opentitan/issues/19844

### DIFF
--- a/sw/device/tests/spi_host_config_test.c
+++ b/sw/device/tests/spi_host_config_test.c
@@ -1,0 +1,105 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+#include <assert.h>
+
+#include "sw/device/lib/arch/device.h"
+#include "sw/device/lib/base/macros.h"
+#include "sw/device/lib/base/memory.h"
+#include "sw/device/lib/base/mmio.h"
+#include "sw/device/lib/dif/dif_pinmux.h"
+#include "sw/device/lib/dif/dif_spi_host.h"
+#include "sw/device/lib/runtime/hart.h"
+#include "sw/device/lib/runtime/log.h"
+#include "sw/device/lib/runtime/print.h"
+#include "sw/device/lib/testing/spi_device_testutils.h"
+#include "sw/device/lib/testing/spi_flash_testutils.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+#include "sw/device/lib/testing/test_framework/ottf_utils.h"
+#include "sw/device/tests/spi_host_flash_test_impl.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+static_assert(__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__,
+              "This test assumes the target platform is little endian.");
+
+OTTF_DEFINE_TEST_CONFIG();
+
+enum {
+  kDefaultTimeoutMicros = 50000,
+};
+
+uint8_t backdoor_cpha = UINT8_MAX;
+uint8_t backdoor_cpol = UINT8_MAX;
+uint32_t backdoor_data = UINT32_MAX;
+
+static status_t spi_config_test(dif_spi_host_t *spi);
+
+static status_t configure_pinmux(const dif_pinmux_t *pinmux);
+
+bool test_main(void) {
+  dif_spi_host_t spi_host;
+  mmio_region_t base_addr =
+      mmio_region_from_addr(TOP_EARLGREY_PINMUX_AON_BASE_ADDR);
+
+  dif_pinmux_t pinmux;
+  CHECK_DIF_OK(dif_pinmux_init(base_addr, &pinmux));
+  base_addr = mmio_region_from_addr(TOP_EARLGREY_SPI_HOST1_BASE_ADDR);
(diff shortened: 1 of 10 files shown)


sw/device/tests/spi_host_config_test.c
```diff
@@ -0,0 +1,105 @@
...
+  for (size_t i = 0; i < 4; ++i) {
+    CHECK_STATUS_OK(spi_config_test(&spi_host));
+  }
+  return true;
+}
+
+static status_t spi_config_test(dif_spi_host_t *spi) {
+  dif_spi_host_config_t config = {
+      .spi_clock = 40000,
+      .peripheral_clock_freq_hz = (uint32_t)kClockFreqUsbHz,
```
vogelpi, line 61: I first thought this seemed off, but since we're testing SPI_HOST1 which is running at 48 MHz (= USB clock) this is indeed correct. I think it would be worth to add a comment on this.

sw/device/tests/spi_host_config_test.c
```diff
@@ -0,0 +1,105 @@
...
+  configure_pinmux(&pinmux);
+  for (size_t i = 0; i < 4; ++i) {
+    CHECK_STATUS_OK(spi_config_test(&spi_host));
+  }
+  return true;
+}
+
+static status_t spi_config_test(dif_spi_host_t *spi) {
+  dif_spi_host_config_t config = {
+      .spi_clock = 40000,
```
vogelpi, line 60: Can you elaborate a bit on this number? 40k seems rather low compared to the values we use in other tests. Is this maybe a typo? I am also curious how this relates to previous comments in the issue regarding limitations on the max speed (16 MHz or 30 MHz). Sorry for asking this, I am no SPI expert :-/

vogelpi (approved):
Thanks for taking care of implementing this test @engdoreis ! This looks all good to me. I have some questions regarding the clock frequency configuration but I am happy to continue discussing this post merge. It's important to get this in and proceed.

vogelpi:
I've merge this such it can be picked up for the next nightly regression.

sw/device/tests/spi_host_config_test.c
a-will: Just an FYI: The correct enum reference in the current implementation is `kClockFreqHiSpeedPeripheralHz`. See https://github.com/lowRISC/opentitan/blob/8f2ad247ced6fde7db1b8ff3dcebe021c7bf8b0a/sw/device/lib/arch/device.h#L96

sw/device/tests/spi_host_config_test.c
engdoreis: I'm intentionally using 40kHz because of a limitation of the test platform, We don't have a spi device on the hyperdebug to test the OT SPI host, so I'm using bit banging instead.To ensure accurate signal capture and avoid aliasing, I've chosen a test frequency of 40kHz, adhering to the Nyquist limit with the Hyperdebug's 100kHz sampling rate.
Does that make sense?

sw/device/tests/spi_host_config_test.c
engdoreis: Thanks for catching that, I'll open a PR with a fix.

sw/device/tests/spi_host_config_test.c
jwnrt: Please could you lift this number up into a constant and document it if you get the chance?

sw/host/opentitanlib/src/test_utils/bitbanging/spi.rs
```diff
@@ -0,0 +1,155 @@
...
+    > {
+        pub cpol: bool,
+        pub cpha: bool,
+        pub data_mode: SpiDataMode,
+    }
+
+    impl<const D0: u8, const D1: u8, const D2: u8, const D3: u8, const CLK: u8, const CS: u8>
+        Decoder<D0, D1, D2, D3, CLK, CS>
+    {
+        /// Returns a sample when a raise or fall clock edge is detected depending on the cpol and cpha configuration.
```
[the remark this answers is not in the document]
engdoreis: You are right. I fix the comment in a follow-up PR.

sw/device/tests/spi_host_config_test.c
vogelpi: Thanks @engdoreis for providing more details. This makes perfect sense to me! It may be worth adding this information in a comment in case you decide to add a constant as @jwnrt recommended.

sw/device/tests/spi_host_config_test.c
engdoreis: I double-checked, and `kClockFreqHiSpeedPeripheralHz` is not the correct constant, according to this [diagram](https://opentitan.org/documentation/index.html), the SPI_HOST1 shares the same clock domain with the USB.
So, either we rename the constant or leave it as it is.

vogelpi: I think the correct thing to do here is:
```
kClockFreqHiSpeedPeripheralHz / 2
```
It happens to be 48 MHz as well but it may not be the same 48 MHz clock. For USB, we do have some clock tracking and tuning logic between USBDEV and AST. I think this doesn't apply to the SPI_HOST1 clock.

jwnrt:
Doesn't need cherry picking to `earlgrey_1.0.0`, I think this was merged before the branch point (the commits exist either way).

Issue #22108, opened 2024-03-19
[i2c] V2(S) Signoff


andreaskurth:
Ensure V2(S) signoff criteria are fulfilled.

This issue tracks V2(S) signoff for the M5-scoped components of the I2C hwip. As such, this is not a full signoff, but a signoff with qualifiers. The qualifying conditions can be seen in the following [linked thread comment](https://github.com/lowRISC/opentitan/issues/22108#issuecomment-2225759494). (And are also linked in the `i2c.hjson` file 'revisions' section under 'Notes'). The lightweight signoff template has also been completed in the [following comment](https://github.com/lowRISC/opentitan/issues/22108#issuecomment-2152850795) for the scoped features.

---

Signoff Meeting Notes (OpentitanACL)
- [Meeting Notes 06/06/24](https://docs.google.com/document/d/1559GiZe9Ada32IDraXT2mtv5NJawHpdcBVRh80zxBqQ/edit) 
- [Meeting Notes 05/07/24](https://docs.google.com/document/d/17qnLpGPr_dJs0denJ8p-iY8bJmJ2MpVAYv9e1tkxWhQ/edit#heading=h.4pl5f3fnl2sb)

---

Depends on items captured in the following issues to close:
- [x] #23077 

Effort tracked here is just for further signoff admin.

rswarbrick:
I've set up "pretend meeting review notes" [here](https://docs.google.com/document/d/1VXJUi7edxKygUOZwDiCh_1APMHRiICAMIhcKGdBj6Sk/edit), which is hopefully the leg-work needed for signing off the block as v1.

I believe the main work item remaining is to tweak the testplan to reflect design changes that have come in since the ES tapeout. This task is tracked as #22168.

moidx:
Discussed moving as P1 to M5 due to RTL risk analysis performed during last issue triage.

hcallahan-lowrisc:
The [above V2(S) review comment](https://github.com/lowRISC/opentitan/issues/22108#issuecomment-2152850795) follows the [lightweight signoff process](https://docs.google.com/document/d/1ZaM2SpZjsBO5Aaj0Lai7LE5wR5osp6StDQAEUQJPfDE/edit#heading=h.ee574932an4y) for a V2(S) level signoff, but as agreed and discussed, this signoff does not meet the full signoff critera due to the addition of new RTL features at a late point in the overall development timeline. 

**Importantly**, the following qualifying critera apply:
[...]

PR #23987, opened 2024-07-11
[i2c,dv] I2C regression fixes

hcallahan-lowrisc:
This is a bunch of smaller fixes, and cleanups that happened along the way, to get the overall I2C block-level regression back up to V2(S) levels.
Review commit-by-commit, see commit messages for more details about each failure. Most of these failures just came about from previous DV changes to model I2C transfers properly, but some are stragglers due to previous RTL changes.

The final commit removes the `i2c_target_ack_stop_vseq` from the target_stress_all test, as I believe this test is no longer functional due to a small RTL bug. See #24010 for more details.

--- 

`./util/dvsim/dvsim.py hw/ip/i2c/dv/i2c_sim_cfg.hjson -i all_once --tool vcs --reseed 10 --cov`

_home_harry_projects_opentitan_scratch_dev4_tmp_reports_hw_ip_i2c_dv_latest_report html

### DIFF
--- a/hw/dv/sv/i2c_agent/i2c_agent_cfg.sv
+++ b/hw/dv/sv/i2c_agent/i2c_agent_cfg.sv
@@ -30,6 +30,7 @@ class i2c_agent_cfg extends dv_base_agent_cfg;
   // test should not update timing parameter.
   // This variable indicates target DUT receives the end of the transaction
   // and allow tb to program a new timing parameter.
+  // TODO(#23920) Remove this.
   bit     got_stop = 0;
 
   int     rcvd_rd_byte = 0; // Increment each time the agent reads a byte.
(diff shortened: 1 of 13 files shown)


andreaskurth (approved):
Thanks for these comprehensive fixes delivered as well-separated and clearly documented (both in commit messages and in-line comments) changes, @hcallahan-lowrisc. LGTM!

hcallahan-lowrisc:
After running a small local regression, with the report shown in the PR description, I think the results are satisfactory. All tests have one seed passing, and the overall pass-percentage (ignoring the V3/unmapped tests) is 442/470 = 94%. The coverage results (excluding FSM coverage) are all sitting above or close to 90%, and I expect all of the metrics to rise above 90% when collecting coverage over a full 50-iterations-per-test regression run. Only 10 iterations per test were run here.
Note. FSM coverage is expected to be low, see the review note and further commentary in #22108.

PR #23989, opened 2024-07-11
[site] Remove landing page from repo

jwnrt:
The landing page for https://opentitan.org has been moved to a separate repo.

Removing it from this repo to prevent duplication. Also prevents this repo from overriding the landing page deployed by the separate repo.

* The block diagram was moved too.
* When building/serving locally, the book is built at the root rather than `/book/` so it's easier to browse without a landing page.


jwnrt:
I tested this using `./util/site/build-docs.sh serve` and browsing the local docs, and also `./util/site/deploy.sh staging` and checked that the staging site was correct (the landing page was not overwritten but the `/book/...` files were.

mundaym (approved):

hcallahan-lowrisc (approved):
LGTM! Thanks for this long-overdue change

PR #23991, opened 2024-07-11
[ci] Remove gcc-9 from packages in CI

jwnrt:
This was used to build Verilator which we now download from a bucket.

We're getting this old GCC from a separate repository which is currently having network issues. Removing this dependency saves time and makes CI less fragile.

### DIFF
--- a/ci/install-package-dependencies.sh
+++ b/ci/install-package-dependencies.sh
@@ -59,15 +59,6 @@ TMPDIR="$(mktemp -d)" || {
 }
 trap 'rm -rf "$TMPDIR"' EXIT
 
-# Install gcc-9 and set it as the default.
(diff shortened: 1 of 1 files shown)


mundaym:
@jwnrt I assume we're using LLVM to build the ROM so that wouldn't be affected by this?

mundaym (approved):

nbdd0121 (approved):

nbdd0121:
We're not using host side GCC to build device code anyway

jwnrt:
> @jwnrt I assume we're using LLVM to build the ROM so that wouldn't be affected by this?

Yep, the whole toolchain used for device software is handled separately within bazel.

PR #23992, opened 2024-07-11
[rv_dm,dv] Implement jtag_dmi_failed_op testpoint

### LINKED ISSUE #17034: [rv_dm,dv] Verify that a failed DMI op is reflected properly in dtmcs register (`jtag_dmi_failed_op` BLT)
This issue tracks the implementation of the `jtag_dmi_failed_op` block-level test, which is specified in the testplan for `rv_dm`.

> estimate 8

rswarbrick:
This was reasonably fiddly because causing the failure requires back-to-back JTAG writes to the dmi register, and the race window isn't very wide.

Fixes #17034.

### DIFF
--- a/hw/ip/rv_dm/data/rv_dm_testplan.hjson
+++ b/hw/ip/rv_dm/data/rv_dm_testplan.hjson
@@ -183,7 +183,7 @@
             - TBO - unclear how to generate other types of failed DMI transactions.
             '''
       stage: V2
-      tests: [] // TODO(#17034)
+      tests: ["rv_dm_dmi_failed_op"]
     }
     {
       name: jtag_dmi_dm_inactive
--- a/hw/ip/rv_dm/data/rv_dm_testplan.hjson
+++ b/hw/ip/rv_dm/data/rv_dm_testplan.hjson
@@ -180,7 +180,6 @@
             - Clear it by writing to dtmcs[dmireset] register.
             - Poll the DMI access for completion and verify that the first write completed
               successfully by doing a read-check.
-            - TBO - unclear how to generate other types of failed DMI transactions.
             '''
       stage: V2
       tests: ["rv_dm_dmi_failed_op"]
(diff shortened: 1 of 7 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_dmi_failed_op_vseq.sv
```diff
@@ -0,0 +1,85 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// smoke test vseq
```
andreaskurth, line 5: This comment seems misplaced :-)

hw/ip/rv_dm/data/rv_dm_testplan.hjson
```diff
@@ -183,7 +183,7 @@
             - TBO - unclear how to generate other types of failed DMI transactions.
```
andreaskurth, line 183: Can we get rid of this line? Or is it still open, in which case we should probably create a V3(?) issue to track it?

andreaskurth (approved):
LGTM overall, thx @rswarbrick!

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_dmi_failed_op_vseq.sv
rswarbrick: Oops. Guess where I copied this file from... :-) I'll drop it!

hw/ip/rv_dm/data/rv_dm_testplan.hjson
rswarbrick: Good point. It isn't open. I've just pushed an extra commit that drops the line (and explains the reasoning in the commit message). Would you mind taking a quick look to check that my change is reasonable?

andreaskurth: Sounds perfectly reasonable to me, thx :+1:

andreaskurth (approved):

PR #23994, opened 2024-07-11
[otbn,dv] Wipe INSN_CNT on an RMA that comes in when we're wiping

rswarbrick:
I'd put this logic into the PRE_WIPE state but hadn't noticed that the RMA might appear when we are in the middle of a round of wiping. Match it in the WIPING state as well.


vogelpi (approved):
LGTM, thanks @rswarbrick !

PR #23996, opened 2024-07-11
[reggen,dv] Remove some unused type: ignore comments

rswarbrick:
These had no effect but cause errors if you run mypy with

  make -C hw/ip/otbn/dv/otbnsim lint

I think I fixed the original problems with 6e37279.

### DIFF
--- a/util/reggen/field.py
+++ b/util/reggen/field.py
@@ -4,7 +4,7 @@
 
 from typing import Dict, List, Optional
 
-from design.mubi import prim_mubi  # type: ignore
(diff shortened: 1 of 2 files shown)


rswarbrick:
Hmm. That doesn't work at all! I think the problem is that we don't set up a path correctly when we run this in CI, which means that the mypy check for OTBN can't see the reggen code. The lines that I was removing here are telling the test not to worry about it.

I suspect we can sort out the paths properly without much work, but the first step in tidying this up is definitely not this PR.

PR #23998, opened 2024-07-11
[manuf] generalize perso flow extensions

timothytrippel:
This generalizes the perso flow extension infrastructure to enable extending any portion of the provisioning flow for pre and post silicon manufacturing testing and provisioning.


moidx (approved):
LGTM

PR #23999, opened 2024-07-11
[dv,usbdev] Scoreboard, functional model, register prediction and coverage

alees24:
This ~~draft~~ PR implements a number of improvements to the usb20_monitor functionality, as well as checking within the scoreboard against a functional model of USBDEV.

- All USB traffic is checked against the functional model.
- The vast majority of register reads are checked in the scoreboard against predictions formed from the functional model or the RAL.
- A loosely-timed prediction model is used for those register fields that change sooner in the functional model than in the DUT. The delay is register field-specific and calculations are based upon what is known about the clock relationships and the delays within the DUT.
- A checked process validates all register field predictions from the functional model against the changes within the DUT, ensuring that no predictions expire, whether or not there is a register read traffic on the TL-UL. This is achieved using backdoor access to the CSRs to avoid impacting the behavior/timing of the DUT and DV.

Additionally the PR introduces collection of functional coverage within the scoreboard.

The loosely-timed register model/predictions have been implemented in a manner that hopefully means they may be migrated/copied to benefit other IP blocks, keeping the USBDEV-specific information in a containing parent class `usbdev_timed_regs`.


hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
```diff
@@ -9,58 +9,64 @@ class usbdev_scoreboard extends cip_base_scoreboard #(
...
+    // Note: presently the functional model in the scoreboard does not have a concept of elapsed
+    // time and the capacity to change state when nothing has happened, eg. if a handshake from the
+    // USB host controller is not received within the timeout interval.
+    //
+    // It is therefore necessary to disable the read data checking of 'configin_' registers for
+    // the couple of sequences affected:
+    // - `usbdev_device_timeout`
+    // - `usbdev_timeout_missing_host_handshake`
+    cfg.en_scb_rdchk_configin = 1'b1;
+    void'($value$plusargs("en_scb_rdchk_configin=%0b", cfg.en_scb_rdchk_configin));
```
rswarbrick, line 48: For this and the next plusarg, is there a way for the vseq to disable things as it starts (and then re-enable them when it stops)? I'm thinking about things like stress sequences, where we want to temporarily disable stuff (and don't have a plusarg to point at)

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
```diff
@@ -69,509 +75,690 @@ class usbdev_scoreboard extends cip_base_scoreboard #(
...
+    if (!item.is_write()) return;
+
+    `uvm_info(`gfn, $sformatf("Writing 0x%0x to %s (index %0h)", wdata, csr_name, index),
+              UVM_MEDIUM)
+    if (csr_name != "buffer") void'(csr.predict(.value(wdata), .kind(UVM_PREDICT_WRITE)));
+
+    // Update the state of the model according to this write operation.
+    case (1)
+      // Interrupt handling.
+      (!uvm_re_match("intr_test", csr_name)): begin
```
rswarbrick, line 541: Do you actually need regular expressions here? Doesn't something like `"intr_test" == csr_name` work? (It's a pity that `case` doesn't support strings!)

hw/ip/usbdev/dv/env/usbdev_timed_regs.sv
```diff
@@ -0,0 +1,220 @@
...
+  virtual clk_rst_if clk_rst_vif;
+  // Access to DUT registers.
+  usbdev_reg_block ral;
+
+  timed_reg timed[timed_reg_e];
+
+  // Prediction ID; diagnostic purposes only.
+  int unsigned pred_id = 0;
+
+  // Monotonic, wrapping cycle count; used to detect when expectations have not been met.
```
rswarbrick, line 42: I'd be inclined to drop "wrapping" here. If we ever simulate 2**32 cycles, I suspect something is going to get confused when `latest_time` for some prediction is less than `time_now`.

rswarbrick (commented):
A couple of notes about the first commit, but it looks really good. I particularly like the timed prediction: it's really clean!

hw/ip/usbdev/dv/env/timed_reg.sv
```diff
@@ -0,0 +1,84 @@
...
+
+  `uvm_object_new
+
+  // Register field within DV/RAL.
+  dv_base_reg_field  field;
+
+  // Mask for this register field within the parent register.
+  uvm_reg_data_t     mask;
+  // Bit number of LSB of the register field within the parent register.
+  int unsigned       lsb;
```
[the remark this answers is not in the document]
alees24: In the function `add_field`, those functions are used to collect the values of mask and lsb, to make the places where they are used somewhat more compact and readable. Without retaining the mask, especially, the expressions could become very ugly:
```
  (reg >> ....field.get_lsb_pos()) & ((1 << ...field.get_n_bits()) - 1)
cf.
 (reg & ..mask) >> ( ... lsb)
```
There's an argument against replicating the data, of course, but I felt it was justified in the case.

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
alees24: As discussed there was precedent elsewhere for the one place that the wildcard matching is used (the bank of 'configin_<x>' registers). I'll try it with a different approach.

hw/ip/usbdev/dv/env/usbdev_timed_regs.sv
alees24: As discussed, the comparisons are done carefully, and - although I'm sure there are plenty of other things that would break after 2^32 cycles - I've tried to avoid adding another.

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
alees24: Good point. That will need some thought. These plusargs are used in the build phase presently, because they modify the set of fields present in some registers for forming predictions.

hw/ip/usbdev/dv/env/timed_reg.sv
```diff
@@ -0,0 +1,84 @@
...
+  uvm_reg_data_t unpred_mask = {`UVM_REG_DATA_WIDTH{1'b1}};
+
+  // Predicted fields within this register.
+  timed_reg_field fields[$];
+
+  // Add a field to the description of this register.
+  //
+  // `init_val` specifies the expected initial value of the field (ie. post-DUT reset).
+  // `max_delay` gives the maximum expected delay before a prediction is met (in DUT clock cycles).
+  function void add_field(ref dv_base_reg_field field, input uvm_reg_data_t init_val,
```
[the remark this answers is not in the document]
alees24: It's not so much about performance, although I don't really know how large those objects are, but rather the fact that's there no just no need to call by value here, since the function is going to have retain its own copy within the object that it creates.

hw/ip/usbdev/dv/env/timed_reg.sv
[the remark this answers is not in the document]
alees24: Please see updated implementation.

alees24:
I've refactored the handling of the loosely-timed registers in a manner that I hope you find more acceptable. This is perhaps a more object-oriented approach and has the distinct advantage of increased separation between that which is USBDEV-specific and that which is generic and reusable.

hw/ip/usbdev/dv/env/timed_reg.sv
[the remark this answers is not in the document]
rswarbrick: This looks really nice, thanks!

hw/ip/usbdev/dv/env/timed_reg.sv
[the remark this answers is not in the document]
rswarbrick: Ahah: this took me quite a while to find in the 1800 spec, but the important bit comes at the end of section 8.2 (top of page 169), which says that "objects" (i.e. class instances) are always passed by "object handle" (a pointer in other languages).

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
rswarbrick: This looks nice, thanks.

hw/ip/usbdev/dv/env/usbdev_timed_regs.sv
rswarbrick: I'm impressed :-D

rswarbrick (commented):
As a really nitty note, I think the first line of the first commit message has a stray square bracket. Possibly worth dropping on the next force-push.

hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -449,12 +478,14 @@ class usb20_monitor extends dv_base_monitor #(
...
-  function void token_packet(pid_type_e pid, bit valid_sync, bit valid_eop, bit valid_stuffing,
-                             ref bit destuffed_packet[$]);
+  function void token_packet(pid_type_e pid, bit low_speed, bit valid_sync, bit valid_eop,
+                             bit valid_stuffing, ref bit destuffed_packet[$]);
     m_token_pkt.m_pid_type = pid;
+    m_token_pkt.low_speed = low_speed;
     m_token_pkt.valid_sync = valid_sync;
     m_token_pkt.valid_eop = valid_eop;
     m_token_pkt.valid_stuffing = valid_stuffing;
+$display("token_packet 0x%0x low_speed %0d",m_token_pkt.m_pid_type, m_token_pkt.low_speed);
```
rswarbrick, line 488: Nit: indentation is off here

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
```diff
@@ -182,6 +188,91 @@ class usbdev_scoreboard extends cip_base_scoreboard #(
     end
   endtask
 
+  // Collection of functional coverage for SOF packets received by the DUT.
+  function void collect_sof_cov(ref sof_pkt sof);
+    if (!cfg.en_cov) return;
```
rswarbrick, line 193: This is nice! When I've written stuff in the past, I've made it so that you only call the function when coverage is enabled. But this way of doing things avoids some pointless repetition. I'll nick it! Thanks :-)

rswarbrick (approved):
Thanks for all the clean-ups. This looks really good to me.

hw/ip/usbdev/dv/env/timed_reg.sv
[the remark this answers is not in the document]
rswarbrick: I thought further: "Why are you allowed to pass objects by reference at all?". Presumably, the point is that you might want to adjust the handle itself, doing something like:
```systemverilog
function void foo(ref my_object_type obj);
  obj = null;
endfunction

function void bar();
  my_object_type x = new;
  assert(x != null);
  foo(x);
  assert(x == null);
endfunction;
```
(untested, but I think it would look something like that)

In C, that would be something like
```c
void foo(my_object_t **obj) {
 *obj = NULL;
}

void bar() {
  my_object_t item;
  my_object_t* ptr = &item;
  assert(ptr != NULL);
  foo(&ptr);
  assert(ptr == NULL);
}
```
This looks pretty ugly, but I *think* this is the only use for that SystemVerilog feature.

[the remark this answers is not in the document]
alees24: Thanks.

hw/dv/sv/usb20_agent/usb20_monitor.sv
alees24: Thanks. That line should have been deleted; it's my practise to outdent like that when something is temporary!

rswarbrick: Heh, I add my initials. If you ever see a PR with "rjs" in it then I've made a silly mistake... :-D

PR #24003, opened 2024-07-12
[entropy_src/dv] Fix edge case for predicting observe FIFO depth

h-filali:
This PR fixes an edge case for the prediction of the observe FIFO depth. When a new word is being pushed into the observe FIFO while we are in the address phase of reading the observe_fifo_depth register, the prediction will be off by one cycle. The reason for this is that the prediction can't be done while the register is being read, so the prediction has to wait until the register is not busy anymore. This commit adds two additional variables that allow for accurate prediction even in this edge case.


vogelpi (approved):
LGTM, thanks @h-filali !

vogelpi:
This is a DV-only PR without impact on SW or FPGA tests, all relevant checks passed already - Merging!

PR #24005, opened 2024-07-12
[top_earlgrey, conn] Update top-level connectivity specs

vogelpi:
This commit re-aligns the top-level connectivity specs with the RTL updates since Eargrey ES. These are:
- Update SPI_DEVICE and USBDEV memory paths for memory test port connections from AST.
- Update lc_rma_req/ack signals between LC_CTRL, FLASH_CTRL, OTBN. There are now dedicated connections and we no longer daisy chain things.


matutem (approved):
LGTM

PR #24008, opened 2024-07-12
[dv,chip] Delete legacy USBDEV-related test

alees24:
Delete testpoint because the functionality that it describes does not exist. There is no such CSR/signal.


andreaskurth (approved):
Confirmed that neither the `usb_state_debug_i` port nor the `wake_debug` CSR exist in usbdev. LGTM!

PR #24009, opened 2024-07-12
[chip-test,sival] Optimize ENTROPY_SRC config for ast_clk_rst_inputs

vogelpi:
This test previously configured the ENTROPY_SRC in Firmware Override: Extract and Insert mode and erroneously enabled the routing of entropy to firmware (ENTROPY_DATA register). Also the entropy source was reconfigured multiple times during the test.

This commit modifies the test to 1) use the Firmware Override: Observe mode in which entropy is observable from the Observe FIFO but also continues to flow through the pipeline to CSRNG and the EDNs, and 2) to not re-configure the entropy source but just drain the observe FIFO to restart the entropy re-collection.

In addition, the default_spinwait_timeout_ns value in the vseq is updated together with the run_timeout_mins to avoid random timeouts. Also a check that the observe FIFO is empty when we come out of deep sleep is removed. This check doesn't make sense as after deep sleep, a reset is performed which empties the observe FIFO anyway.

### DIFF
--- a/hw/top_earlgrey/dv/chip_sim_cfg.hjson
+++ b/hw/top_earlgrey/dv/chip_sim_cfg.hjson
@@ -2111,6 +2111,7 @@
       sw_images: ["//sw/device/tests/sim_dv:ast_clk_rst_inputs:1:new_rules"]
       en_run_modes: ["sw_test_mode_test_rom"]
       run_opts: ["+sw_test_timeout_ns=200_000_000"]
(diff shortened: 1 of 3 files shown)


vogelpi:
Background: this is a V2 test which has been very flaky for some time now.

andreaskurth (approved):
LGTM, thanks for these test improvements @vogelpi :+1:

PR #24012, opened 2024-07-14
[flash/dv] Add dependency in otf_sb when using vendor flash

eshapira:
When running otf tests with the vendor flash this check isn't valid so in this PR a dependency added in the use_vendor_flash variable.
This PR along with small PR in the foundry is fixing failing seeds of test flash_ctrl_disable.

### DIFF
--- a/hw/ip_templates/flash_ctrl/dv/env/flash_ctrl_otf_scoreboard.sv
+++ b/hw/ip_templates/flash_ctrl/dv/env/flash_ctrl_otf_scoreboard.sv
@@ -471,38 +471,41 @@ class flash_ctrl_otf_scoreboard extends uvm_scoreboard;
 
     // check if this is page erase
     if (exp.req.pg_erase_req | exp.req.bk_erase_req) begin
-      int cnt = 0;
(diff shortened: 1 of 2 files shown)


eshapira:
Hi @matutem , can you please review this PR?
Thanks

eshapira:
Hi @rswarbrick , can you please review this PR?
@matutem seems to be unavailable.

matutem:
@eshapira this avoids data checks in foundry, probably because it is not supported? Should we try to support these checks at some point to tighten up checking?

matutem (approved):
This change looks okay, but the real question is whether we plan to perform the corresponding checks against the foundry repo.

LGTM

eshapira:
@matutem thanks for your review!
As you maybe noticed this change is disabling checks that are already disabled in the foundry.
I think Jaedon disabled them when he worked on the flash IP DV, not sure why.
In order to support those checks there some work that should be done to make the otf_scoreboard more flexible, because currently its checks aren't compatible with the foundry flash model behavior.
We need to see how we can improve those checks.

PR #24013, opened 2024-07-15
[util] Use importlib.metadata in lintpy.py

rswarbrick:
This avoids a DeprecationWarning when running with a recent version of Python.


hcallahan-lowrisc (approved):
:+1:

PR #24016, opened 2024-07-15
[Sival, Spi_host] Improve comments on the config test

engdoreis:
This PR addresses the comments on this https://github.com/lowRISC/opentitan/pull/23985 PR.

### DIFF
--- a/sw/device/tests/spi_host_config_test.c
+++ b/sw/device/tests/spi_host_config_test.c
@@ -28,6 +28,10 @@ OTTF_DEFINE_TEST_CONFIG();
 
 enum {
   kDefaultTimeoutMicros = 50000,
+  // This test uses the hyperdebug bit banging. To ensure accurate signal
+  // capture and avoid aliasing, we run the spi at 40kHz, adhering to the
+  // Nyquist limit with the Hyperdebug's 100kHz sampling rate.
+  kSpiClock = 20000,
 };
 
 uint8_t backdoor_cpha = UINT8_MAX;
@@ -57,8 +61,8 @@ bool test_main(void) {
 
 static status_t spi_config_test(dif_spi_host_t *spi) {
   dif_spi_host_config_t config = {
-      .spi_clock = 40000,
-      .peripheral_clock_freq_hz = (uint32_t)kClockFreqUsbHz,
+      .spi_clock = kSpiClock,
+      .peripheral_clock_freq_hz = (uint32_t)kClockFreqHiSpeedPeripheralHz / 2,
   };
   backdoor_cpha = UINT8_MAX;
   backdoor_cpol = UINT8_MAX;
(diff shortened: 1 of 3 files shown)


sw/device/tests/spi_host_config_test.c
```diff
@@ -28,6 +28,10 @@ OTTF_DEFINE_TEST_CONFIG();
 
 enum {
   kDefaultTimeoutMicros = 50000,
+  // This test uses the hyperdebug bit banging.To ensure accurate signal capture
```
jwnrt, line 31:
```suggestion
  // This test uses the hyperdebug bit banging. To ensure accurate signal capture
```

jwnrt (approved):
LGTM, thanks

sw/device/tests/spi_host_config_test.c
```diff
@@ -57,8 +61,8 @@ bool test_main(void) {
 
 static status_t spi_config_test(dif_spi_host_t *spi) {
   dif_spi_host_config_t config = {
-      .spi_clock = 40000,
-      .peripheral_clock_freq_hz = (uint32_t)kClockFreqUsbHz,
+      .spi_clock = kSpiClock,
+      .peripheral_clock_freq_hz = (uint32_t)kClockFreqHiSpeedPeripheralHz,
```
vogelpi, line 65: I think this should be
```
.peripheral_clock_freq_hz = (uint32_t)kClockFreqHiSpeedPeripheralHz / 2,
```
like in `sw/device/tests/power_virus_systemtest.c`.

vogelpi: Ah, actually I am not sure, because on FPGA `kClockFreqHiSpeedPeripheralHz` is 24 MHz rather than 96 MHz.

engdoreis: The `kClockFreqUsbHz` constant on the FPGA is greater than the `kClockFreqCpuMhz`. Is that right?
If following ratio of Silicon, the first should be 12MHz, and therefore `.peripheral_clock_freq_hz = (uint32_t)kClockFreqHiSpeedPeripheralHz / 2,` would be correct.

vogelpi: > The `kClockFreqUsbHz` constant on the FPGA is greater than the `kClockFreqCpuMhz`. Is that right?

Yes that's right. The USB clock must always be 48 MHz but Ibex only runs at 24 MHz on the FPGA.

> If following ratio of Silicon, the first should be 12MHz, and therefore `.peripheral_clock_freq_hz = (uint32_t)kClockFreqHiSpeedPeripheralHz / 2,` would be correct.

Yes, I think we should just use `(uint32_t)kClockFreqHiSpeedPeripheralHz / 2`.

engdoreis: done

vogelpi (approved):
LGTM, thanks @engdoreis !

Issue #22048, opened 2024-03-15
[rv_dm,dv] Extend block-level testing to cover new NDM and WARL behaviour


rswarbrick:
### Description

Code to implement the behaviour in the design is in PR #22038. We need to:
- Extend the testplan to cover it
- Add functional coverage to make sure we're actually seeing the behaviour
- Strengthen the scoreboard to require the behaviour we expect.
- Write one or more sequences to actually trigger the feature

rswarbrick:
@johngt: As someone with more knowledge of PROD timescales, would you mind trying to figure out which milestone this should be tracked in? (My guess is M4, but I thought I should get a second opinion!)

moidx:
rv_dm DV work carried over to M5.

andreaskurth:
@rswarbrick: Is this planned for M5?

rswarbrick:
It's definitely part of the plan for M5 (honestly, this is part of the V2 verification stage). I'm not certain whether we'll meet the M5 targets for rv_dm, but I don't believe this is any different from the other ones for this block.

rswarbrick:
The main part of this should be completed with #24017. The name of the issue (my fault...) talks about the new WARL behaviour in rv_dm. I'm not entirely sure this is worth testing explicitly, and it *definitely* doesn't need to be part of M5.

rswarbrick:
Following up on this, I realise that the linked PR isn't of the form suggested in the issue description.

- There was an existing test which was basically doing what we needed, so I didn't think there was a need for a tweak to the testplan.
- We don't really have any functional coverage being used for this sort of thing in rv_dm. Honestly, the vseq here is more of a directed test, so fcov might not be the best tool for tracking it.
- Similarly with the scoreboard: this is really a directed test sequence.

rswarbrick:
About the other half of this, I just "bit the bullet" and implemented the obvious check where we write `'1` to `hartselhi` and `hartsello` then read the register back to get the number of harts.

This isn't quite working properly (because of DV problems: we aren't properly updating the RAL in the base classes), but I can look at the waves and I'm convinced that the RTL is handling this correctly.

PR #24017, opened 2024-07-15
[rv_dm,dv] Extend ndmreset test to check the reset pins are corrected properly

rswarbrick:
This is part of what is tracked in #22048, which wanted to make sure we were testing the new behaviour that we have added to rv_dm to make it track ndm resets properly.

The first commit is a "wiring tidyup" which makes sure that the signals are being driven properly. I'm not sure it was strictly needed, but this avoids some X's propagating through the design. The second commit strengthens the vseq and tries to check that the feature works properly.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
@@ -25,7 +25,20 @@ class rv_dm_ndmreset_req_vseq extends rv_dm_base_vseq;
     `DV_CHECK(dmstatus.allunavail)
   endtask
 
+  // Check that the ndmreset_pending_q signal in the design has the value we expect.
+  task check_ndmreset_pending(bit expected_value);
+    uvm_reg_data_t val;
+    string path = "tb.dut.ndmreset_pending_q";
+
+    `DV_CHECK(uvm_hdl_read(path, val));
+    `DV_CHECK_EQ(val, expected_value);
+  endtask
+
   task body();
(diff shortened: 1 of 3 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
```diff
@@ -46,8 +59,17 @@ class rv_dm_ndmreset_req_vseq extends rv_dm_base_vseq;
     // remains high: it will drop again when we disable lc_hw_debug_en below.
     `DV_CHECK(cfg.rv_dm_vif.cb.ndmreset_req)
 
-    // Pretend that the "rest of the system" has indeed gone into reset by asserting the
-    // unavailable_i input, which means that the hart is powered down.
+    // We should also expect to see the ndmreset_pending_q signal go high (tracking the fact that we
+    // are in the middle of an ndm reset)
+    check_ndmreset_pending(1'b1);
+
+    // Resetting the system would normally drop the rst_ni pin (restetting rv_dm as well). Because
```
andreaskurth, line 66: Small typo :-)
```suggestion
    // Resetting the system would normally drop the rst_ni pin (resetting rv_dm as well). Because
```

andreaskurth (approved):
LGTM (besides a small typo), thx @rswarbrick!

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
rswarbrick: Hah! I think I started with "rest" (of the system) and auto-completed. Duh...

PR #24022, opened 2024-07-15
[hmac,dv] remove invalid assertion

### LINKED ISSUE #23954: [test-triage] rom_e2e_self_hash

Chip Level

### Failure Description

 UVM_ERROR @ 200016.083823 us: (chip_sw_base_vseq.sv:306) uvm_test_top.env.virtual_sequencer [uvm_test_top.env.virtual_sequencer.chip_sw_base_vseq] SW TEST TIMED OUT. STATE: SwTestStatusBooted, TIMEOUT = 200000000 ns
  
  UVM_INFO @ 200016.083823 us: (uvm_report_catcher.svh:705) [UVM/REPORT/CATCHER]

### Steps to Reproduce

- GitHub Revision: 2e5d86c
  ./util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i rom_e2e_self_hash --build-seed 75221189197949424635294305394615322888112457483844341597147780944629972574676 --fixed-seed 71265896955482039545714391449479203836693477801414682517890286279159020201964

- [ ] rom_e2e_self_hash

timothytrippel:
This assertion was checking that a `process` or `stop` command happened only after a `start` or `continue` command had been issued. However, it was doing so by checking if the `process` or `stop` signals being high overlapped with the `initiated` signal being high (which simply latches whether a `start` or `continue` had been seen in the past). However, multiple `process` commands may be issued by SW before a `stop` or `done`, and the hardware does support this. Any ROM E2E test that enabled SPX+ signature verification does indeed trigger multiple `process` operations. This was the root cause behind #23954.

Since, what this assertion was checking was too restrictive, we opt for simply removing it.

This fixes #23954.

### DIFF
--- a/hw/ip/hmac/rtl/hmac.sv
+++ b/hw/ip/hmac/rtl/hmac.sv
@@ -928,8 +928,6 @@ module hmac
 
   // Below condition is covered by the design (2020-02-19)
   //`ASSERT(ValidHashStartAssert, hash_start_or_continue |-> !initiated)
-  // `hash_process` or `reg_hash_stop` should be toggled and paired with `hash_start_or_continue`
-  `ASSERT(ValidHashProcessAssert, (hash_process || reg_hash_stop) |-> initiated)
 
   // hmac_en should be modified only when the logic is Idle
   `ASSERT(ValidHmacEnConditionAssert,


vogelpi:
Thanks @timothytrippel for debugging this and for filing the PR. To me, the RTL behavior of the HMAC block with respect to accepting multiple `Process` commands per message is worrying. I am no SHA2 expert but AFAIK the `Process` command triggers the padding followed by the final absorption operation. After that, the hashing is completed and the final digest is ready. This is also what the programmer's guide of our HMAC module mentions: https://opentitan.org/book/hw/ip/hmac/doc/programmers_guide.html#triggering-hmacsha-2-engine

I took a look at the HMAC RTL and I don't understand why we're okay with multiple `Process` commands per message. It doesn't look like intended behavior. Can we please confirm the following things before moving forward?
1. The SHA2 engine really does the padding once and only once per message.
2. If it would be done more than once, we wouldn't end up with a correct digest. Are we 100% sure that the combination of software and hardware in this case is able to produce correct digests? Have we tested this particular combination with some test vectors?
3. If this indeed works correctly, does it work reliably or are does it just work because the timing is right and e.g. the hardware internally drops the process command when it arrives while the engine is busy? Even though this is ROM code and we assume things are pretty stable from a timing perspective, there are still some knobs who can change this: e.g. enabling SRAM ReadBack with an OTP fuse will make SRAM accesses slower. As a result, we may end up feeding HMAC slower and the `Process` commands might not get dropped anymore.

Sorry for slowing things down but since this is ROM code, I would like to be very careful.

timothytrippel:
Thanks @vogelpi for the pointers. Maybe I am missing something, but it looks like the padding states of the FSM [here](https://cs.opensource.google/opentitan/opentitan/+/master:hw/ip/hmac/rtl/hmac_core.sv;drc=d79a3bb4f80cb416fcf8740edc3a94572b92ef0f;l=328) are only entered via `reg_hash_start_i` not `hash_process` no? There is also this comment [here](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/drivers/hmac.c;drc=388b408d0800c6b589f223552ba7f67845ad646c;l=95) from @jadephilipoom it looks like?

gdessouky:
Thanks @timothytrippel for digging into this and @vogelpi for triggering the discussion.
The SHA-2 engine still does the padding and message absorption and finalizes the computation once `hash_process` is triggered, and this is not impacted if is triggered multiple times afterwards; the SHA-2 engine primitive does support that. 

At some point when we were working for integrated, I had implemented the SHA-2 primitive to capture `hash_process` once it is triggered and keep it captured in a flag till we receive the next `hash_start` (or `hash_continue` now), thus any more `hash_process` commands do not influence the operation - this change was implemented to cater for specific requirements that came up while integrating for DMA. 
https://github.com/lowRISC/opentitan/blob/a365ced52c2af6ee20a891d6e8f5ddbfbd2b25ce/hw/ip/prim/rtl/prim_sha2_32.sv#L188-L191

Having said that, we never verified this for HMAC (or we would have triggered that assertion already), because there was no SW usage need (that we were aware of)  that required supporting triggering `hmac_process` multiple times. Why is `hmac_process` triggered twice - is this a side effect or a constraint of some sort? I agree with @vogelpi that the safest way here would be to adhere to the programmer's guide and just trigger it once till the hashing completes.

vogelpi:
Thanks everybody for your input. After some sleep and going again through [Tim's comment here](https://github.com/lowRISC/opentitan/issues/23954#issuecomment-2229149325) I think the way the code exercises the hardware is okay.

The code of interest is here:
https://github.com/lowRISC/opentitan/blob/9a1c8516950ee4fbd7021928b6030e2f42b4a923/sw/device/silicon_creator/lib/sigverify/sphincsplus/wots.c#L46-L55

The important part is that software first feeds in all the data into the message FIFO before triggering the `Process` command. As explained in the comment, we send the `Process` command explicitly to be able to call the `spx_addr_hash_set()` while the hardware is busy. The second `Process` is then sent as part of the [hmac_sha256_final_truncated()](https://github.com/lowRISC/opentitan/blob/9a1c8516950ee4fbd7021928b6030e2f42b4a923/sw/device/silicon_creator/lib/sigverify/sphincsplus/wots.c#L54C5-L54C32) function, but we are not pushing additional data into the message FIFO in between.

What the SVA is trying to protect us from is that we send the `Process` command before having sent all data to the message FIFO. As @gdessouky explained above, this would work as long the hardware doesn't empty the FIFO in between receiving `Process` and the last data word. If the FIFO got emptied before receiving the last data word, the hardware would take the `Process` command and finish the digest. Additional data being received afterwards would not contribute to the digest even though if another `Process` command would be sent. I was afraid we could end up in this case depending on the performance of Ibex. This could really result in catastrophic failure of the chip.

Well, I am now convinced that the ROM code in this particular case won't trigger this behavior. Still the SVA is valid as it tries to prevent us from operating the hardware the wrong way. I think we should keep it. Instead, I am suggesting one of the following options:
1. Modify the ROM code to add an argument to `hmac_sha256_final_truncated()` deciding whether the `Process` command should be sent or not. This may trigger additional hardening efforts and is thus potentially less convenient at this stage.
2. Modify the ROM code to not call `hmac_sha256_process()` before `hmac_sha256_final_truncated()`, i.e., disable the performance optimization. This is still a ROM change but it's easier. In terms of performance we would loose around 20 clock cycles (according to the waves shared by Tim).
3. Alternatively, we could instead tweak the SVA to check that the between the first `Process` command and subsequent `Process` commands, no more data is fed into the message FIFO. This wouldn't involve ROM changes but may trigger other issues.

Personally, I am in favor of 1 or 2 if it's still doable from a schedule perspective.

jadephilipoom:
I'm coming to this a bit late since I just got back from vacation, but thanks everyone for figuring out the issue here! I thought that sending `process` twice would be fine because I didn't observe any difference in the results, but if it's not fine I think we can do a modification of Pirmin's option (1)

>     1. Modify the ROM code to add an argument to `hmac_sha256_final_truncated()` deciding whether the `Process` command should be sent or not. This may trigger additional hardening efforts and is thus potentially less convenient at this stage.

I'd say we can just remove the `process` command from `hmac_sha256_final_truncated` in all cases, and declare in the documentation of the `final` functions that you need to call `hmac_sha256_process` before calling them otherwise the code will hang. It means an extra line of code in one or two places, but I think it would be worth it. Most code that has a simple use case for SHA2 uses the `hmac_sha256` one-shot operation from `hmac.h` anyway, so I don't think there's a need to have an option -- it would also be simpler this way from a hardening perspective.

gdessouky:
@jadephilipoom was also going to suggest that. Decoupling triggering process from waiting on the operation to be done altogether might be best here, but was worried how much impact that would have throughout instances where this is called, and it's also an extra usage constraint to keep in mind as you point out. 

In any case, I'm definitely also in favor of this/any variant of this rather than triggering hmac_process() twice, even if we have enough confidence that it would work fine,  just because we haven't explicitly verified it for HMAC.

vogelpi:
This sound good to me, thanks everybody!

timothytrippel:
Thanks everyone!

> I'd say we can just remove the `process` command from `hmac_sha256_final_truncated` in all cases, and declare in the documentation of the `final` functions that you need to call `hmac_sha256_process` before calling them otherwise the code will hang. It means an extra line of code in one or two places, but I think it would be worth it. Most code that has a simple use case for SHA2 uses the `hmac_sha256` one-shot operation from `hmac.h` anyway, so I don't think there's a need to have an option -- it would also be simpler this way from a hardening perspective.

I think this seems like the best option. I will go ahead and make this change and tag a new ROM release candidate.

PR #24026, opened 2024-07-16
[rv_dm,dv] Add WARL check for hartsel

rswarbrick:
This is the extra check discussed in #22048. The second commit adds the check, but it didn't work at first! This turned out to be a silly DV bug, fixed in the first commit.


vogelpi (approved):
LGTM!

PR #24031, opened 2024-07-16
[dv] Correctly handle a reset when we're waiting to issue one

rswarbrick:


hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -806,6 +816,10 @@ class cip_base_vseq #(
     for (cycles_waited = 0;
          cycles_waited < wait_cycles || cycles_with_no_accesses > 0;
          ++cycles_waited) begin
+      // If we are actually in reset then there's no need to do any more waiting: the caller can
+      // "apply a reset now" (a no-op)
+      if (!cfg.clk_rst_vif.rst_n) return;
+
```
eshapira, line 822: Maybe it will be better to put the condition before this task is called, means that maybe it's better to move the calling of this task under this condition in the run_seq_with_rand_reset_vseq task?
See the other suggestions to complete it.
```suggestion
```

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -740,20 +740,30 @@ class cip_base_vseq #(
               wait(ongoing_reset == 0);
               `uvm_info(`gfn, $sformatf("\nFinished run %0d/%0d w/o reset", i, num_times), UVM_LOW)
             end
+            // This process waits a random time (until it's safe to issue a reset) and then issues a
+            // reset for a short time, ending when we leave reset again. If the dut goes into reset
+            // for another reason while we are waiting, this waits until that reset is finished.
             begin : issue_rand_reset
               wait_to_issue_reset(reset_delay_bound);
```
eshapira, line 748:
```suggestion
```

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -740,20 +740,30 @@ class cip_base_vseq #(
...
-              ongoing_reset = 1'b1;
-              `uvm_info(`gfn, $sformatf("\nReset is issued for run %0d/%0d", i, num_times), UVM_LOW)
-              apply_resets_concurrently();
-              do_read_and_check_all_csrs = 1'b1;
-              ongoing_reset = 1'b0;
+              if (!cfg.clk_rst_vif.rst_n) begin
+                // If we are in reset at this point then someone else must have put us there! Wait
+                // until we leave reset, at which point our job is done.
+                @(cfg.clk_rst_vif.rst_n);
+              end else begin
```
eshapira, line 753:
```suggestion
              if (!cfg.clk_rst_vif.rst_n) begin
                // If we are in reset at this point then someone else must have put us there! Wait
                // until we leave reset, at which point our job is done.
                @(cfg.clk_rst_vif.rst_n);
              end else begin
                wait_to_issue_reset(reset_delay_bound);
```

eshapira (approved):
See the comment with some connected suggestions, but regardless LGTM.
Thanks!

PR #24032, opened 2024-07-16
[dv,jtag] Don't constrain wait for trst_n to be de-asserted

rswarbrick:
Using the DV_WAIT macro causes a failure if the signal doesn't go high again in time. The driver shouldn't really be checking for things like this: it should just wait until we come out of reset.


hcallahan-lowrisc (approved):
Yep agree with the sentiment. I presume no use of the CB on this statement is intentional?

rswarbrick:
Hmm, it wasn't intentional: I just hadn't thought about that at all! But I think I still believe it: the `trst_n` signal is kind of asynchronous and isn't really tied to a clock at all.

Issue #23380, opened 2024-05-29
[hmac,dv] Verify HMAC idle status


gdessouky:
### Description

The idle status of HMAC is now exposed via the `STATUS` CSR in bit 0 (https://github.com/lowRISC/opentitan/issues/22916). This needs to be modeled and predicted in the scoreboard and verified against the STATUS CSR whenever read. For now, I read it from the CSR and update the prediction to match it, but we do not verify it yet.

gdessouky:
@andreaskurth I've labelled this as M5, since the driver implementation and ongoing DV work is already uncovering issues for now, so would guarantee sufficient assurance for now.

andreaskurth:
@martin-velay can provide more details, but when I look at the current `hmac_scoreboard` I think the `idle` bit of the `STATUS` register gets checked -- correct?

martin-velay:
@andreaskurth, I don't think this feature has been verified as of now unfortunately.
I don't know why, but I think it was not visible on the main board, probably a tag was missing.

The PR #23749 is only checking the pin in test mode.

I'll add this check in the SCB.

gdessouky:
It's not yet verified, we read it from the status register for now.
Issue is linked there - https://github.com/lowRISC/opentitan/blob/master/hw/ip/hmac/dv/env/hmac_scoreboard.sv#L444-L445

martin-velay:
Closed as https://github.com/lowRISC/opentitan/pull/24037 has been merged

PR #24037, opened 2024-07-17
[hmac,dv] Verify idle status register

martin-velay:
Linked to this issue https://github.com/lowRISC/opentitan/issues/23380

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_scoreboard.sv
+++ b/hw/ip/hmac/dv/env/hmac_scoreboard.sv
@@ -130,12 +130,12 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
               if (item.a_data[HashProcess] && (hmac_start || hmac_continue)) begin
                 {hmac_process, hmac_start} = item.a_data[1:0];
                 if (!invalid_cfg) begin
-                  check_idle_o(1'b0);
+                  check_idle(1'b0);
                   // only predict a new digest if configuration is valid and HMAC was indeed started
                   // otherwise previous digest is retained in CSRs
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
This looks nice, thanks!

martin-velay:
Thanks for your quick review @rswarbrick !

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -849,9 +847,14 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
...
 
-  virtual function void check_idle_o(bit val);
-    if (cfg.under_reset == 0) `DV_CHECK_EQ(cfg.hmac_vif.is_idle(), val)
-  endfunction : check_idle_o
+  // Check idle output pin and update status register predicted value
+  virtual function void check_idle(bit val);
+    if (cfg.under_reset == 0) begin
+      `DV_CHECK_EQ(cfg.hmac_vif.is_idle(), val)
+    end
+    hmac_idle = val;
```
gdessouky, line 855: Is there a reason why `hmac_idle = val` is outside the if condition? I would have had it inside but I may be missing something/or it makes no difference?

gdessouky (commented):
Thanks @martin-velay, LGTM, just left a small question

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: It shouldn't make any difference I'd say, but it's a good remark thanks!

PR #24043, opened 2024-07-17
[jtag,dv] Correct FSM driving after trst_n

rswarbrick:
The previous version of this code was assuming that we'd come out of trst_n in FSM state RunTest/IDLE. This *was* the behaviour of the debug module with a previous version of the RTL (which is presumably how this happened). But it's not what is supposed to happen, and it's not what the RTL does any more (since commit 663be282861).

Get things lined up properly.

### DIFF
--- a/hw/dv/sv/jtag_agent/jtag_driver.sv
+++ b/hw/dv/sv/jtag_agent/jtag_driver.sv
@@ -128,10 +128,16 @@ class jtag_driver extends dv_base_driver #(jtag_item, jtag_agent_cfg);
         $cast(rsp, req.clone());
         rsp.set_id_info(req);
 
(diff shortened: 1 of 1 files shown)


rswarbrick:
Well, that took me a long time to diagnose! The reason that this was failing CI is that we have some chip-level tests that merrily drive the jtag interface themselves (see `chip_tap_straps.vseq`). My proposed change made it so that the driver would react to a `trst_n` and try to get back to Run-Test/Idle. This didn't work at all when something else was controlling TCK.

The revised version is rather more simple and (I think) still fixes the original problem that this PR was trying to fix. It also has some documentation comments that explain our path through the FSM reasonably carefully. Because I got rather confused more than once.

rswarbrick:
And "driver tidyup" too! Thanks for noticing :-)

PR #24044, opened 2024-07-17
[rv_dm,dv] Weaken enable checker to allow "left over" items

rswarbrick:
These signals mostly need some response before they drop (e.g. the a_valid signal, which won't drop without a_ready). If one is asserted and then debug stops being enabled before we get a response, it's ok. The thing that isn't allowed is for one of those signals to *become* asserted when debug is not enabled.


hcallahan-lowrisc (approved):
This makes sense to me. Thanks :)

PR #24045, opened 2024-07-17
[rv_dm,dv] Properly clear cmderr field in rv_dm_cmderr_busy_vseq

rswarbrick:
This allows us to chain the vseq after other sequences that cause a command error.

I found this because of the failing `rv_dm_stress_all_with_rand_reset` test. That test works by writing random junk at the same time as an interesting sequence. The result is that the command error promptly gets set again! I'll address that in a separate PR, but this change should make this vseq more robust.


eshapira (approved):
LGTM

PR #24046, opened 2024-07-17
[dv] Add "writes_ignore_errors" flag

rswarbrick:
This is really for `rv_dm` (see the last commit). But the idea is that we've currently got some code in `cip_base_vseq__tl_errors.svh` that tries to generate bad writes by sending an invalid mask value. The `stress_all_with_rand_reset` sequence is supposed to be checking that those writes won't have any effect.

Unfortunately, they *do* have an effect for some registers in `rv_dm`. I don't *think* this behaviour is disallowed by the TileLink spec, so the correct response is probably to avoid depending on the writes being squashed.

### DIFF
--- a/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__tl_errors.svh
+++ b/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__tl_errors.svh
@@ -64,6 +64,11 @@ virtual task tl_write_less_than_csr_width(string ral_name);
     `DV_CHECK_FATAL($cast(csr, all_csrs[i]))
     msb_pos = csr.get_msb_pos();
     addr    = csr.get_address();
+
+    // If this is a register that might change on a write that causes an error then we shouldn't
+    // generate writes to it because we might update the register contents unexpectedly.
+    if (csr.writes_ignore_errors) continue;
+
     `create_tl_access_error_case(
         tl_write_less_than_csr_width,
         opcode inside {tlul_pkg::PutFullData, tlul_pkg::PutPartialData};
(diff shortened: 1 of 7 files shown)


rswarbrick:
I've just rejigged this dramatically. With the original change, the stress_all_with_rand_reset sequence was still failing, because it performed writes with bad masks on the abstract data registers, which *do* look at their byte enable signal. I wanted a weaker property that says something more like "Writing to this register might do something, even if the mask is bogus". The new version reflects that.

vogelpi (commented):
Thanks for the PR @rswarbrick , the regtool changes look very clean and good to me.

What I am not sure about is whether this isn't an RTL bug you've found here. Are we sure it's intended behavior that these registers update despite generating an error?

rswarbrick:
It's definitely not intended behaviour, and I raised an issue about it here: #24053.

The RTL bug was mistake when we patched the vendored file. Presumably not security-critical, so needn't affect the current tape-out (but shows that we probably should have done some verification of this module a bit earlier).

vogelpi:
Thanks @rswarbrick for filing the issue.

Knowing that this is bug and that we are going to fix it post M5 / for a future release, would it make sense to list the issue in the changes introduced with this PR such that no one starts to use the `write_ignore_errors` flag? I am thinking that we want to remove this again after fixing the bug. What is your view?

rswarbrick:
Good point. I've just updated to add a TODO message pointing at the issue. Thanks!

vogelpi:
CHANGE AUTHORIZED: hw/ip/rv_dm/data/rv_dm.hjson

vogelpi (approved):

vogelpi:
Thanks @rswarbrick for adding the TODO comment. This looks good to me now. You'll also need to add the CHANGE AUTHORIZED line. Once I think CI ran through we can merge this.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/rv_dm/data/rv_dm.hjson

Issue #18496, opened 2023-05-10
[sw/silicon_creator] Add manifest extensions


alphan:
### Description

- [ ] Add `signed_region_end`.
- [ ] Add `ext_table` and related types.
- [ ] Add extensions for SPHINCS+ key (signed) and signature (unsigned).
- [ ] Update `opentitantool` to make these fields optional (@jon-flatley )

PR #24048, opened 2024-07-17
opentitantool: implement spx signature verification

vbendeb:
This patch adds code verifying SPX+ signature of OT binary images, when requested. Some refactoring was required to reflect the fact that an image could be signed with either one or two keys.

Tested by successfully running 'opentitantool image manifest verify --spx <rom_ext.bin>'.


vbendeb:
Closing a hole left after #18496 was marked fixed, makes it possible to validate both main and optional TO image signatures.

timothytrippel (approved):

vbendeb:
fixed a problem in the pre-push script, updated the source code.

PR #24054, opened 2024-07-18
[otbn,bazel] Add a way to generate random-ish tests for OTBN.

jadephilipoom:
This is a slight expansion of the existing OTBN test infrastructure that allows us to generate test data based on a known-good implementation. This PR includes both the infrastructure and an example that tests our bignum `mul` implementation against test data that generates some numbers and then multiplies them in Python.

Since Bazel doesn't allow _actually_ random generation of numbers at build time, I included a "seed" argument that gets passed through from the `BUILD` file to seed the pseudorandom test-data generation. The test data for the first test (seed = 0) will always be the same on every build, and so will the test data for the second test, etc, as long as the test-generation script is the same. But they will be different from each other, so by repeating the target 10 times like in the example, you can get 10 different sets of generated data.

If there's a better way of increasing coverage here to run more truly random tests, I'm all ears -- I considered trying to e.g. get the digest of the OTBN binary so that changing the binary would trigger new data, but in the end I decided that this was too likely to lead to a situation where there's a failure in some corner case that's triggered by a specific seed value, and then someone updates the binary in a trivial way and thinks the bug is "fixed". So I decided to resist the temptation to over-engineer and just go with the dumb but transparent integer seed.

### DIFF
--- a/hw/ip/otbn/util/shared/testgen.py
+++ b/hw/ip/otbn/util/shared/testgen.py
@@ -1,0 +1,45 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Dict, TextIO
+
+_GPR_NAMES = [f'x{i}' for i in range(32)]
+_WDR_NAMES = [f'w{i}' for i in range(32)]
+
+
+def write_test_data(inputs: Dict[str, bytes], data_file: TextIO) -> None:
+    '''Write the test input data to DMEM.'''
+    data_file.write('.data\n')
+    for name in inputs:
+        data_file.write('\n')
+        if len(inputs[name]) <= 4:
+            # Buffer should be read as a 32-bit (4-byte) value.
+            align = 4
+            length = 4
+        else:
+            # Buffer should be read as a (sequence of) 256-bit (32-byte) values.
+            align = 32
+            length = ((len(inputs[name]) + 31) // 32) * 32
+        data_file.write(f'.balign {align}\n')
+        data_file.write(f'.globl {name}\n')
+        data_file.write(f'{name}:\n')
+        for i in range(0, length, 4):
+            word = int.from_bytes(inputs[name][i:i + 4], byteorder='little')
+            data_file.write(f'.word {word:#010x}\n')
+
+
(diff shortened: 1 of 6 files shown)


rswarbrick:
I might be missing something, but how does this differ from the stuff we do on the DV side using `hw/ip/otbn/dv/rig`? Isn't that tool doing something very similar? (but without the Bazel integration)

jadephilipoom:
> I might be missing something, but how does this differ from the stuff we do on the DV side using `hw/ip/otbn/dv/rig`? Isn't that tool doing something very similar? (but without the Bazel integration)

I didn't know about `rig`, but from a quick glance it seems like it generates random snippets of _instructions_ -- is that right? This setup is for tests where the instructions are always the same but the test data (DMEM inputs + expected final register values) is generated at build time. That way if we wanted to test, for example, SHA256, we can generate some random messages and use a third-party Python library to calculate the expected hashes.

rswarbrick:
Oh, I see! Yes, that's definitely different and now I understand: makes sense to me.

hw/ip/otbn/util/shared/testgen.py
```diff
@@ -0,0 +1,45 @@
...
+            length = 4
+        else:
+            # Buffer should be read as a (sequence of) 256-bit (32-byte) values.
+            align = 32
+            length = ((len(inputs[name]) + 31) // 32) * 32
+        data_file.write(f'.balign {align}\n')
+        data_file.write(f'.globl {name}\n')
+        data_file.write(f'{name}:\n')
+        for i in range(0, length, 4):
+            word = int.from_bytes(inputs[name][i:i + 4], byteorder='little')
```
rswarbrick, line 28: Oh, neat! I didn't know that Python allowed you to run off the end of an array like this.

rswarbrick (approved):
This looks sensible to me. Presumably the point of this is to check that we've written OTBN assembly correctly (with future testgen files), rather than testing the HW itself?

jadephilipoom:
> This looks sensible to me. Presumably the point of this is to check that we've written OTBN assembly correctly (with future testgen files), rather than testing the HW itself?

Yes, exactly. :slightly_smiling_face: So far the status quo is that the software developer does this offline, hardcodes the test data, and makes comments in the test about how to re-generate it. But this process is annoying and not super transparent, so it results in us running fewer tests on the crypto software than we hopefully will in the future with this new rule.

PR #24057, opened 2024-07-18
[dv, spi_host] Re-implementing command.CSAAT <-> CSB check

antmarzam:
Plus a bit of cleanup for unused variables and a warning from a VSEQ

### DIFF
--- a/hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
+++ b/hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
@@ -18,7 +18,6 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   // hold expected transactions
   spi_segment_item                  host_wr_segment;
   spi_segment_item                  host_rd_segment;
-  spi_item                          device_item;
   spi_item                          plain_item;
 
   // local variables
@@ -28,6 +27,11 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   // 'spi_segment_item.spi_data' are used.
   local spi_segment_item            read_segment_q[$];
   local bit [7:0]                   rx_data_q[$];
+  // state of the last segment CSAAT field. It's used to check CSB lines behaved correctly
+  // after each segment:
+  local bit                         csaat = 0;
+  // The configured half period in ps on the SPI bus
+  local int                         spi_clk_half_period;
 
   // interrupt bit vector
   local bit [NumSpiHostIntr-1:0]    intr_state = 2'b00;
@@ -45,7 +49,14 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   local spi_host_intr_state_t       spi_intr_state_reg;
   local spi_host_intr_enable_t      spi_intr_enable_reg;
   local spi_host_intr_test_t        spi_intr_test_reg;
+  // Holds latest written value
   spi_host_configopts_t             spi_configopts;
+  // Holds value modelling FSM in case reg is re-configured during in-flight spi txn
+  spi_host_configopts_t             curr_spi_configopts;
+  // Flag used to control when 'curr_spi_configopts' is updated in case there is a SPI-txn
+  // in flight when the configopts register is written
+  bit                               initialise_configopts = 1;
+
   // Tally-Counters
   int                               in_tx_seg_cnt      = 0;
   int                               checked_tx_seg_cnt = 0;
@@ -62,24 +73,121 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
     plain_data_fifo  = new("plain_data_fifo", this);
     host_wr_segment = spi_segment_item::type_id::create("host_wr_segment");
     host_rd_segment = spi_segment_item::type_id::create("host_rd_segment");
-    device_item = spi_item::type_id::create("device_item");
   endfunction
 
 
   task run_phase(uvm_phase phase);
     super.run_phase(phase);
+    // cfg.en_scb currently does not change mid-test
+    wait (cfg.en_scb == 1);
     forever begin
+      wait(cfg.clk_rst_vif.rst_n && !spi_ctrl_reg.sw_rst);
+      // Wait until CSB signal has the enable set
+      wait (cfg.spi_passthrough_vif.cio_csb_en_o === 1);
       `DV_SPINWAIT_EXIT(
         fork
           compare_tx_trans();
           compare_rx_trans();
           get_plain_txn();
+          check_csn_lead();
+          check_csn_idle();
         join,
         @(negedge cfg.clk_rst_vif.rst_n or event_sw_rst)
       )
       `uvm_info(`gfn, "Restarting scoreboard checking due to reset event now.", UVM_LOW)
     end
   endtask : run_phase
+
+  // Checks the minimum time between commands CSB stays high - which should equate
+  // to configopts.csnidle + 1
+  virtual task check_csn_idle();
+    int unsigned time_high, time_low;
+    int unsigned quarter_period;
+    int unsigned num_quarter_cycles;
+
+    `uvm_info(`gfn, $sformatf("%m - After reset"), UVM_DEBUG)
+    forever begin
+
+      wait (spi_ctrl_reg.sw_rst === 0);
+      @(posedge cfg.m_spi_agent_cfg.vif.csb[0]);
+      // Using quarter period for greater granularity
+      num_quarter_cycles = (curr_spi_configopts.csnidle + 1) * 2;
+      quarter_period = spi_clk_half_period / 2;
+      // Spec says 'configot.idle' is the minimum amount of half SCK cycles before the next command
+      for(int i = 0; i < num_quarter_cycles; i++) begin
+        #(quarter_period * 1ps);
+        if (cfg.m_spi_agent_cfg.vif.csb[0] === 0) begin
+          `uvm_error(`gfn, {$sformatf("(CONFIGOPTS.csnidle*2=%0d) > %0d 1/4 cycles since CSB=0x1.",
+                                      num_quarter_cycles, i),
+                            "... and CSB has just being deasserted which is against the spec"})
+        end
+      end
+      // After the time above, is ok for the RTL to set CSB low anytime
+    end // forever begin
+  endtask : check_csn_idle
+
+
+  virtual task check_csn_lead();
+    time start_time, end_time;
+    int unsigned quarter_period;
+
+    `uvm_info(`gfn, $sformatf("%m - After reset and CSB output enable"), UVM_DEBUG)
+    // Just a delay to ensure the X -> set/unset in clock is not detected after
+    // 'cio_csb_en_o' just becomes active.
+    cfg.clk_rst_vif.wait_clks(1);
+
+    // Threads set timestamps of moments in which CSB became active and first spi cycle.
+    // We then check the amount of spi half cycles isn't greater than the CSNLEAD register value
+    forever begin
+      @(negedge cfg.m_spi_agent_cfg.vif.csb[0]);
+      `uvm_info(`gfn, "CSB went low", UVM_DEBUG)
+      fork : iso_fork
+        begin
+          fork
+            begin
+              start_time = $time;
+              `uvm_info(`gfn, $sformatf("%m - NEGEDGE of CSB"), UVM_DEBUG)
+              @(posedge cfg.m_spi_agent_cfg.vif.csb[0]);
+              `uvm_info(`gfn, $sformatf("%m - POSEDGE of CSB: exiting thread"), UVM_DEBUG)
+            end
+            begin
+              wait_start_sck_edge();
+              // Wait a quarter period to ensure the sampled CSB value is correct with no races
+              quarter_period = spi_clk_half_period / 2;
+              #(quarter_period * 1ps);
+              end_time = $time;
+              `uvm_info(`gfn, $sformatf("%m - After clock tick. Exiting thread"), UVM_DEBUG)
+            end
+          join_any
+          disable fork;
+        end
+      join  : iso_fork
+
+      // Only interested for cases where there's been SCK cycles with the chip select being active
+      if (end_time > start_time) begin
+        int unsigned actual_clk_ticks = (end_time - start_time) / spi_clk_half_period;
+        if (actual_clk_ticks > (curr_spi_configopts.csnlead + 1)) begin
+          `uvm_fatal(`gfn, {"Since CSB become active, There's been",
+                            $sformatf(" %0d half SCKs ",
+                                      (end_time - start_time) / spi_clk_half_period ),
+                            $sformatf("VS (configopts.csnlead+1)=%0d",
+                                      curr_spi_configopts.csnlead + 1)})
+        end
+      end
+      else
+        `uvm_fatal(`gfn, "SPI Clock was ticking before CSB went low")
+
+    end
+  endtask : check_csn_lead
+
+  // SPI clock may "tick", as in being set to 0/1 depending on phase/polarity settings
+  // The clock driven in neg-edge if phase != polarity
+  task wait_start_sck_edge();
+    if (spi_configopts.cpol == spi_configopts.cpha)
+      @(negedge cfg.m_spi_agent_cfg.vif.sck);
+    else
+      @(posedge cfg.m_spi_agent_cfg.vif.sck);
+  endtask
 
   // RX spi raw txn. The monitor sampled the spi bus raw, and the data gets decoded in
   // 'compare_tx_trans' according to the segment configuration.
@@ -89,7 +197,7 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
       // Each time CSB goes low, SCB receives a "plain data txn"
       `uvm_info(`gfn, $sformatf("Received: plain_item=\n%s", plain_item.sprint), UVM_DEBUG)
     end
-  endtask
+  endtask : get_plain_txn
 
   // Receives spi_segment_item - this is then compared with the value on the bus which is extracted
   // in compare_tx_trans
@@ -121,7 +229,7 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
       `uvm_info(`gfn, $sformatf("\n successfully compared read transaction of %d ",
                                 tl_segment.command_reg.len+1), UVM_DEBUG)
     end // forever begin
-  endtask
+  endtask : compare_rx_trans
 
   virtual task  extract_data_for_segment(spi_host_command_t  command_info,
                                          output bit[7:0] host_data, output bit [7:0] device_data);
@@ -179,7 +287,7 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
       end
     endcase
 
-  endtask
+  endtask : extract_data_for_segment
 
   // Receives spi_segment_item - this is then compared with the value on the bus which is extracted
   // If the written segment's direction is RxOnly or Bidir, this then also populate the `rx_data_q`
@@ -187,9 +295,7 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   virtual task compare_tx_trans();
     spi_segment_item   exp_segment = spi_segment_item::type_id::create("exp_segment");
 
-    spi_item           dut_item, device_item;
     // indication that this is a new transaction
-    bit                prev_csaat = 0;
     string             txt = "";
     bit [7:0]          host_data, device_data, extracted_data;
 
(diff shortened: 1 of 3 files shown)


hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -238,15 +235,31 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
                                 exp_segment.command_reg.len+1), UVM_HIGH)
+
+      // Check the CSB line goes high/low after writing a segment
+      -> check_csaat_ev;
+
     end // forever begin
   endtask // compare_tx_trans
 
 
+  virtual task check_csaat();
```
rswarbrick, line 251: As far as I can tell, this spawns a background process each time `check_csaat_ev` gets triggered. That process waits until the `csb` matches the value that we've seen on the last register write. (I'm slightly surprised by the `!=`: don't we want to wait until this becomes equal?)

The `DV_WAIT` timeout is probably too long, and weird things are going to happen if we toggle `csaat` frequently: what happens if these overlap? Could you restructure this slightly so that instead of triggering an event, we start a process (probably in a task)? Something like this might be simpler: (untested!)
```systemverilog
  // Wait up to 10 cycles until the csb lines yield the expected csaat or until there is a reset.
  task wait_for_csaat(bit expected);
    fork begin : isolation_fork
      fork
        begin
          bit seen_match = 1'b0;
            if (&cfg.m_spi_agent_cfg.vif.csb == expected) begin
              seen_match = 1'b1;
              break;
            end
            cfg.clk_rst_vif.wait_clks(1);
          end
          `DV_CHECK(seen_match)
        end
        wait(!cfg.clk_rst_vif.rst_n)
      join_any
      disable fork;
    end join
  endtask
```

With that change, we could change the event trigger line to something like:
```systemverilog
      // Check the CSB line goes high/low after writing a segment
      fork wait_for_csaat(csaat);
      join_none
```

antmarzam: Thanks for the suggestion! This was helpful. 

In the end I realised the TB was not checking for `configopts.idle/trail/lead` times and incorporated a check for each of those times. 
I just pushed those changes.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -74,13 +79,99 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+            #(quarter_period * 1ps);
+            end_time = $time;
+            `uvm_info(`gfn, $sformatf("%m - After clock tick. Exiting thread"), UVM_DEBUG)
+          end
+        join_any
+        disable fork;
+        end
+      join  : iso_fork
+
+     // Only interested for cases where there's been SCK cycles with the chip select being active
```
rswarbrick, line 161: Nit: indentation of comment line

antmarzam: done

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -74,13 +79,99 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+        fork
+          begin
+            @ (negedge cfg.m_spi_agent_cfg.vif.csb[0]);
+            start_time = $time;
+            `uvm_info(`gfn, $sformatf("%m - NEGEDGE of CSB"), UVM_DEBUG)
+            @ (posedge cfg.m_spi_agent_cfg.vif.csb[0]);
+            `uvm_info(`gfn, $sformatf("%m - POSEDGE of CSB: exiting thread"), UVM_DEBUG)
+          end
+          begin
+            @ (cfg.m_spi_agent_cfg.vif.sck);
```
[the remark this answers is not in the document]
antmarzam: It depends on the clock phase and polarity, so it can be either

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -74,13 +79,99 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+
+
+  virtual task check_csn_lead();
+    time start_time, end_time;
+    int unsigned quarter_period;
+    // Wait until the design's been reset
+    wait ( cfg.clk_rst_vif.rst_n === 1);
+
+    // Just a delay to ensure the X -> set/unset in clock is not detected after
+    // 'cio_csb_en_o' just becomes active.
```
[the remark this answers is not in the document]
antmarzam: Yep - good spot. Moved it

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -74,13 +79,99 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+      // After the time above, is ok for the RTL to set CSB low anytime
+    end // forever begin
+  endtask : check_csn_idle
+
+
+  virtual task check_csn_lead();
+    time start_time, end_time;
+    int unsigned quarter_period;
+    // Wait until the design's been reset
+    wait ( cfg.clk_rst_vif.rst_n === 1);
```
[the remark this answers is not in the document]
antmarzam: done

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -74,13 +79,99 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+    // Just a delay to ensure the X -> set/unset in clock is not detected after
+    // 'cio_csb_en_o' just becomes active.
+    wait (cfg.en_scb == 1 && cfg.spi_passthrough_vif.cio_csb_en_o === 1);
+
+    `uvm_info(`gfn, $sformatf("%m - After reset and CSB output enable"), UVM_DEBUG)
+    #(cfg.clk_rst_vif.clk_period_ps * 1ps);
+
+
+    forever begin
+      fork : iso_fork
```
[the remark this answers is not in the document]
antmarzam: Sure - I'll add a comment for clarity.

I ended up adding an else with a fatal to the current `if (end_time > start_time) begin...` because what you're describing isn't legal spi_host behaviour. 
CSB drops first, and after `csnlead` half cycles spi_host start ticking the spi-clock.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -74,13 +79,99 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+    int unsigned quarter_period;
+    int unsigned num_quarter_cycles;
+
+    // Wait until the design's been reset
+    wait ( cfg.clk_rst_vif.rst_n === 1);
+    `uvm_info(`gfn, $sformatf("%m - After reset"), UVM_DEBUG)
+
+    forever begin
+
+      wait (spi_ctrl_reg.sw_rst === 0);
```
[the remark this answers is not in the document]
antmarzam: done

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -74,13 +79,99 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+
+  virtual task check_csn_lead();
+    time start_time, end_time;
+    int unsigned quarter_period;
+    // Wait until the design's been reset
+    wait ( cfg.clk_rst_vif.rst_n === 1);
+
+    // Just a delay to ensure the X -> set/unset in clock is not detected after
+    // 'cio_csb_en_o' just becomes active.
+    wait (cfg.en_scb == 1 && cfg.spi_passthrough_vif.cio_csb_en_o === 1);
```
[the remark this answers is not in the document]
antmarzam: done

antmarzam:
@rswarbrick - I believe I've addressed all your comments now. Please Let me know if there's anything I've missed!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -238,15 +332,118 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+            csr_rd(.ptr(ral.status), .value(status), .backdoor(1'b1));
+
+            if (spi_ctrl_reg.spien==0) begin
+              // IF control.Spien unset, we block until the block is re-enabled
+              `uvm_info(`gfn, $sformatf("%m - Blocked at spien"), UVM_DEBUG)
+              wait (spi_ctrl_reg.spien === 1);
+              if (i > 0) begin
+                // Taking an extra cycle if we're blocked during spien
+                // Otherwise, if spien=0x1 mid spi-cycle then we may fail the CSB check
+                i--;
```
rswarbrick, line 376: I'm convinced this is correct, but it might be clearer to set a "stall" flag here and then decrement `i` in one place at the end of the body of the loop. That way, it's obvious that we can't accidentally decrement `i` e.g. past zero!

I'd probably also do it the other way around and do something like:
```systemverilog
  int cycles_left;
  while (cycles_left > 0) begin
    bit stall = 1'b0;
   // ...
   // possibly set stall=1 here
   // ...
   if (!stall) cycles_left -= 1;
  end
```

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -238,15 +332,118 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+    // Spec states the number of half cycles will be 'curr_spi_configopts.csntrail[0] + 1'
+    // we add extra 1 because the monitor sometimes sends data 1/2 cycle earlier.
+    int num_times = curr_spi_configopts.csntrail[0] + 1 + 1;
+    spi_host_status_t status;
+    `uvm_info(`gfn, $sformatf("%m - carrying out CSAAT (%0d) check", csaat), UVM_DEBUG)
+
+    fork begin : iso_fork
+      fork
+        begin : csntrail_check
+          for (int i=0; i < num_times; i++) begin
```
rswarbrick, line 364: Can you add a comment here that explains what the loop is doing? I think it's something like "Wait (with a timeout) until csb[0] != csaat" ?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -238,15 +332,118 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+            end
+
+            if (cfg.m_spi_agent_cfg.vif.csb[0] !== csaat) begin
+              break;
+            end
+            // Consuming the delays programmed into configopts.CSNTRAIL
+            #(spi_clk_half_period * 1ps);
+          end
+
+          // We just delay by 1/4of a period in case there was a race in sampling CSB
```
rswarbrick, line 394: Nit: Missing space after 1/4

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -238,15 +332,118 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+            num_scks ++;
+            // We don't expect to continue sampling data, hence fire an error
+            // if the RTL keeps spitting data out (half a cycle is OK)
+            if (csaat == 0 && num_scks >= max_scks_edges)
+              break;
+          end
+          `uvm_fatal(`gfn, $sformatf("%m - Clock kept ticking"))
+        end
+        begin
+          // Only 1 csb is currently supported
```
rswarbrick, line 417: This comment applies to several other lines as well. Maybe it's best to expand it a bit and move it to the top of the task?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -238,15 +332,118 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+          forever begin
+            @ (cfg.m_spi_agent_cfg.vif.sck);
+            num_scks ++;
+            // We don't expect to continue sampling data, hence fire an error
+            // if the RTL keeps spitting data out (half a cycle is OK)
+            if (csaat == 0 && num_scks >= max_scks_edges)
+              break;
+          end
+          `uvm_fatal(`gfn, $sformatf("%m - Clock kept ticking"))
+        end
```
rswarbrick, line 415: Can't you do this more simply with a for/repeat loop?
```systemverilog
repeat (2) begin
  @(cfg.m_spi_agent_cfg.vif.sck);
  if (csaat == 0) break;
end
`uvm_fatal(...)
```

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -45,7 +46,12 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   local spi_host_intr_state_t       spi_intr_state_reg;
   local spi_host_intr_enable_t      spi_intr_enable_reg;
   local spi_host_intr_test_t        spi_intr_test_reg;
+  // Holds latest written value
   spi_host_configopts_t             spi_configopts;
+  // Holds value modelling FSM in case reg is re-configured during in-fliht spi txn
+  spi_host_configopts_t             curr_spi_configopts;
+  bit                               initialise_configopts = 1;
```
rswarbrick, line 53: Can you add a documentation comment explaining what this bit means?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -62,25 +68,115 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
     host_rd_segment = spi_segment_item::type_id::create("host_rd_segment");
-    device_item = spi_item::type_id::create("device_item");
   endfunction
 
 
   task run_phase(uvm_phase phase);
     super.run_phase(phase);
     forever begin
+      wait(cfg.clk_rst_vif.rst_n && !spi_ctrl_reg.sw_rst);
+      wait (cfg.en_scb == 1 && cfg.spi_passthrough_vif.cio_csb_en_o === 1);
```
rswarbrick, line 78: I don't think that `en_scb` changes in the middle of a test (and I'm sure that would break lots of other stuff!), so it's probably simpler just to move the wait for it outside of the loop.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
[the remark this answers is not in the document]
rswarbrick: Hmm, I see that we have also moved the wait for `cio_csb_en_o`. Is that correct? It seems to be something that comes out of the design, rather than something inside the DV code.

antmarzam:
@rswarbrick I believe I've addressed all your feedback in my latest commit. I had to refactor one of the checks to make it a function of `sys_clk` instead of `spi_clk`. 

I've regressed the changes heavily and didn't see any fallout. Let's wait and see what CI says :)

antmarzam:
I'm trying to respond to your comment https://github.com/lowRISC/opentitan/pull/24057#discussion_r1718326833
But for some reason I don't get a "reply" button. 

It is a design enable output, the one that determines when CSB is active. Would you suggest to model `cio_csb_en_o` instead and check the predicted vs actual value, and then only carry on with behavior if predicted `cio_csb_en_o` is set? 

Not many TBs do that kind of check, but it may be valuable. Although this enable signal is used to control the pads in the top-level and is tested there. 
In any case, if a check like this is worthwhile to add I suggest to make it part of a different PR.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: I forget to respond to the comment, but there's now a `stall_flag`!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Comment is now added: 
```
          // The loop below waits for configopts.csntrail spi half cycles (accounted as their
          // corresponding sys cycles) to see if the CSB line goes up as expected based on
          // the command.CSAAT

```

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: done!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: I've reshuffled the comments a bit and moved them towards the top of the task

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Done!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Done!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Done!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -62,25 +70,119 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+
+    `uvm_info(`gfn, $sformatf("%m - After reset and CSB output enable"), UVM_DEBUG)
+    // Just a delay to ensure the X -> set/unset in clock is not detected after
+    // 'cio_csb_en_o' just becomes active.
+    cfg.clk_rst_vif.wait_clks(1);
+
+    // Threads set timestamps of moments in which CSB became active and first spi cycle.
+    // We then check the amount of spi half cycles isn't greater than the CSNLEAD register value
+    forever begin
+      @ (negedge cfg.m_spi_agent_cfg.vif.csb[0]);
```
rswarbrick, line 139: Nit (sorry): I think the conventional layout in OpenTitan is not to have the space after the `@`.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -62,25 +70,119 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+              @ (posedge cfg.m_spi_agent_cfg.vif.csb[0]);
+              `uvm_info(`gfn, $sformatf("%m - POSEDGE of CSB: exiting thread"), UVM_DEBUG)
+            end
+            begin
+              // SPI clock may "tick", as in being set to 0/1 depending on phase/polarity settings
+              // Clock driven in neg-edge if phase != polarity
+              if (spi_configopts.cpol == spi_configopts.cpha)
+                @ (negedge cfg.m_spi_agent_cfg.vif.sck);
+              else
+                @ (posedge cfg.m_spi_agent_cfg.vif.sck);
```
rswarbrick, line 156: I think it might be better to pull this out into a separate task (which can then have a name and the comment that you've got here ends up as a "documentation comment" for the task.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -62,25 +70,119 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+              `uvm_info(`gfn, $sformatf("%m - After clock tick. Exiting thread"), UVM_DEBUG)
+            end
+          join_any
+          disable fork;
+        end
+      join  : iso_fork
+
+      // Only interested for cases where there's been SCK cycles with the chip select being active
+      if (end_time > start_time) begin
+        int unsigned actual_clk_ticks = (end_time - start_time)/ spi_clk_half_period;
```
rswarbrick, line 171: Nit: Missing space before `/`

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -62,25 +70,119 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+        if ( actual_clk_ticks > (curr_spi_configopts.csnlead + 1)) begin
```
rswarbrick, line 172: Nit: Stray space after first `(`

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -394,6 +615,7 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
           spi_configopts.csnidle  = get_field_val(ral.configopts.csnidle, item.a_data);
           spi_configopts.clkdiv   = get_field_val(ral.configopts.clkdiv, item.a_data);
           spi_configopts.csntrail = get_field_val(ral.configopts.csntrail, item.a_data);
+          update_configopts(spi_configopts);
```
rswarbrick, line 618: Is there a reason we need to mirror `spi_configopts` explicitly like this? Couldn't we use `gmv(ral.configopts.myfield)` instead of reading that value?

(Clearly for a different PR, but I wondered about it when I read the code here)

antmarzam: The reason is that the register configopts updates depending on what's going on with the RTL (ongoing segments). So the update_configopts tasks makes sure that `curr_spi_configopts` updates when it has to.

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -28,6 +27,8 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   // 'spi_segment_item.spi_data' are used.
   local spi_segment_item            read_segment_q[$];
   local bit [7:0]                   rx_data_q[$];
+  local bit                         csaat = 0;
+  local int                         spi_clk_half_period;
```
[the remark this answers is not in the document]
antmarzam: I've added a comment to both `csaat` and `spi_clk_half_period`. 

My preference would be to leave it as is, just so we save the assignment to `half_ticks` every time we wait a quarter period, and instead just assign it once every time configopts is written. 

Maybe there's "too many" of `quarter_period` calculations that can also be done just once each time configopts is written.

Yes, the reason for the hardcoded delays instead of using the spi clock was in case the clock died so we'd still carry out the checks

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Oops - I've just gone through the doc and removed the space in this and the rest of the occurrences :)

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Fixed now!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Done!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -238,15 +338,135 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
-      prev_csaat = exp_segment.command_reg.csaat;
+      csaat = exp_segment.command_reg.csaat;
+      fork
+        begin
+          check_csaat(.csaat(csaat));
+        end
+      join_none
+
+      // store CSAAT so we now if we are starting a new transaction
+      csaat = exp_segment.command_reg.csaat;
```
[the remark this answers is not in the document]
antmarzam: Removed the duplicate!

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: I've made a task called `wait_start_sck_edge` for this :)

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -28,6 +27,11 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
   // 'spi_segment_item.spi_data' are used.
   local spi_segment_item            read_segment_q[$];
   local bit [7:0]                   rx_data_q[$];
+  // state of the last segment CSAAT field. It's Use to check CSB lines behaved correctly
```
rswarbrick, line 30: Silly nit (sorry): I think you want "It's used to ..." (add a 'd' and drop a capital!)

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
```diff
@@ -62,25 +70,119 @@ class spi_host_scoreboard extends cip_base_scoreboard #(
...
+      wait (spi_ctrl_reg.sw_rst === 0);
+      @ (posedge cfg.m_spi_agent_cfg.vif.csb[0]);
+      // Using quarter period for greater granularity
+      num_quarter_cycles = (curr_spi_configopts.csnidle + 1) * 2;
+      quarter_period = spi_clk_half_period / 2;
+      // Spec says 'configot.idle' is the minimum amount of half SCK cycles before the next command
+      for(int i = 0; i < num_quarter_cycles; i++) begin
+        #(quarter_period * 1ps);
+        if (cfg.m_spi_agent_cfg.vif.csb[0] === 0) begin
+          `uvm_fatal(`gfn, {$sformatf("(CONFIGOPTS.csnidle*2=%0d) > %0d 1/4 cycles since CSB=0x1.",
```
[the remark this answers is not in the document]
rswarbrick: I agree, but does it have to kill the test instantly? I was imagining `uvm_error`?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
rswarbrick: Ah, that makes sense. Could we implement this by changing `update_configopts` to call predict on the RAL model value? That would avoid needing to duplicate the state, I think.

antmarzam: I've created an issue to make sure this is not forgotten (#27006) - Would it be OK if I do this as part of a separate PR?

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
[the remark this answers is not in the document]
antmarzam: I've changed it to `uvm_error`

It's probably not a biggie either way because spi_host simulations are not extremely long, but for things like this I'm not sure there's value in continuing the simulation if we know the RTL is "broken". I guess there can be an argument made into how "broken" it is

hw/ip/spi_host/dv/env/spi_host_scoreboard.sv
antmarzam: Updated the comment, thanks!

rswarbrick (approved):
Thanks for the tidy-up!

PR #24059, opened 2024-07-18
[rv_dm,dv] Tidy up reset length calculation

rswarbrick:
This was wrong twice in the same direction. The first commit tidies up a factor of 1000 in the rv_dm code, then the second commit tidies up another factor of up to 10 in the base dv code.


hcallahan-lowrisc (approved):
LGTM

PR #23475, opened 2024-06-04
[test] Add `rv_dm` late enable test 

moidx:
1. Add OTP image with `RV_DM` late debug feature enabled.
2. Introduce test case on FPGA.
3. Update selected test cases to run in `cw340` FPGA.

This is part of https://github.com/lowRISC/opentitan/issues/21965

PR #24061, opened 2024-07-18
[aes/dv] Only detect the final message end if the scoreboard is enabled

vogelpi:
Some automated CSR tests may actually trigger encryption operations without reading the result. Previously, this could cause the post_body() task to timeout. As the scoreboard is anyway disabled for these tests, meaning the produced finish_message variable isn't read, we can simply skip the detection of the final message end.

This resolves lowRISC/OpenTitan#23475.


rswarbrick (approved):
This sort of thing is really annoying! Looks sensible to me.

vogelpi:
Thanks for your review @rswarbrick .

PR #24062, opened 2024-07-18
[csrng/dv] Don't exit process_csrng_cmd_fifo() task upon disable

vogelpi:
Previously, we would erroneously exit the task when handling a disable while waiting for entropy. After this point, the scoreboard would no longer perform any checks on the command FIFO, no longer update the internal state, and no longer compare the generated bits of the corresponding app. This is bad in two ways: either it can go completely unnoticed or lead to failing checks when comparing internal states between the scoreboard and the DUT using check_internal_state().

Luckily, the fix is easy. We can just replace the `return` keyword by a `continue` to drop the current item of the queue and start over in case of a disable.


vogelpi:
I've run a full regression for this change and there is a positive impact on the test pass rates without decreasing coverage or introducing new failure signatures.

h-filali (approved):
Good catch! LGTM

rswarbrick (approved):

PR #24064, opened 2024-07-19
[dv] Track trst_n reset of IR in jtag_dmi_monitor.sv

rswarbrick:
This means that we monitor things correctly if we see a DR transaction immediately after a reset.

Honestly, this is a little silly: it would probably be better if the jtag monitor sent the IR for each transaction (so that we didn't have to implement IR tracking on everything downstream). But that's a rather bigger change.


hw/dv/sv/jtag_dmi_agent/jtag_dmi_monitor.sv
```diff
@@ -104,7 +104,7 @@ class jtag_dmi_monitor #(type ITEM_T = jtag_dmi_item) extends dv_base_monitor#(
           dmi_item.rdata = data;
         end
         `uvm_info(`gfn, $sformatf("Writing DMI item to analysis_port: %0s",
-                                  dmi_item.sprint(uvm_default_line_printer)), UVM_HIGH)
```
hcallahan-lowrisc, line 107: nit. debug logging levels

hcallahan-lowrisc (approved):

hw/dv/sv/jtag_dmi_agent/jtag_dmi_monitor.sv
rswarbrick: Gah! Fixed.

PR #24065, opened 2024-07-19
[rv_dm,dv] Weaken IDCODE check

rswarbrick:
I think this is a (very minor) bug in the riscv-dbg RTL. Reported upstream at https://github.com/pulp-platform/riscv-dbg/issues/170.


hcallahan-lowrisc (approved):
We discussed this in-person, and I'm not exactly sure what the intended behaviour should be if one reads past the length of a read-only register. But as for writing a useful scoreboarding check, I think only checking the number of bits in the register makes sense.

rswarbrick:
I agree. I'll follow up on that on the PULP side ("I'm not sure this is a bug, because..."). But the DV change here is definitely needed.

PR #24075, opened 2024-07-22
[dv] Add a quick exit in csr_rd_sub when in reset

rswarbrick:
If we are trying to read a CSR and are in reset, the operation isn't going to work. This change makes it so that we exit immediately (setting the status to UVM_NOT_OK) rather than blocking indefinitely.


eshapira (approved):
LGTM

PR #24076, opened 2024-07-22
[rv_dm,dv] Minor tweaks to rv_dm dv code to work with stress_all_with_rand_reset

rswarbrick:
These are small changes. The first one handles a situation where we've issued a CSR read which has stopped early because we are in reset (and obviously won't have seen the right value).

The other two concentrate on stopping the sequence if we wish to issue a reset. Honestly, I think the correct behaviour will be for the base classes to issue the resets unconditionally and for the IP classes to handle this. But that's a much bigger change and these are minimal tweaks to make the vseq work at all with the current structure.


hcallahan-lowrisc (approved):
Yeah that seems sensible to me.

Issue #24058, opened 2024-07-18
[test-triage] `lc_ctrl_volatile_unlock_disabled`:`lc_ctrl_sec_mubi`


elliotb-lowrisc:
### Hierarchy of regression failure

Block level

### Failure Description

```
  UVM_ERROR @ 274101392 ps: (lc_ctrl_errors_vseq.sv:768) [uvm_test_top.env.virtual_sequencer.lc_ctrl_sec_mubi_vseq] Check failed flash_rma_error_act == flash_rma_error_exp (0 [0x0] vs 1 [0x1])
```

### Steps to Reproduce

- GitHub Revision: aad711e1bc
- `./util/dvsim/dvsim.py hw/ip/lc_ctrl/dv/lc_ctrl_volatile_unlock_disabled_sim_cfg.hjson -i lc_ctrl_sec_mubi --build-seed 38297581588405771015102977205739484163575733999253050480227719508149669635625 --fixed-seed 25410439243959484991513546383319556044347456481080083384027716590735525914383`

vogelpi:
I've now started looking into this. This test had a 100% pass rate until 1st of July when we've merged #23867.

vogelpi:
The PR to fix this issue is here #24080

PR #24080, opened 2024-07-22
[lc_ctrl/dv] Lower max delay for flash_rma_ack in lc_ctrl_sec_mubi test

vogelpi:
It seems our DV currently always expects invalid MuBi values in the Flash RMA ack signal to arrive within 3 to 4 clock cycles after sending the request. 3 clock cycles are "spent" for CDC already. Thus, we need to constrain the delay lens for the set_flash_rma_ack() interface to be within 0 to 1 clock cycle.

This resolves lowRISC/OpenTitan#24058.


vogelpi:
This is a follow-up of #23867.

rswarbrick (approved):
Looks sensible to me. Thanks for the careful explanations.

PR #24083, opened 2024-07-23
[kmac,dv] Increase timeouts for some tests

martin-velay:
A bunch of tests are actually failing for simple timeout reason. I took the worst case to compute this new value.

### DIFF
--- a/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
+++ b/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
@@ -104,6 +104,7 @@
     {
       name: "{name}_burst_write"
       uvm_test_seq: kmac_burst_write_vseq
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
I think something has got a little mangled with the git history here and we've got some of another PR (which gets reverted in the final commit). Would you mind tidying it up?

rswarbrick (commented):
I assume this is supposed to be a single commit, which contains just the additions to `kmac_base_sim_cfg.hjson`. If so, that looks sensible. Would you mind adding a note to the commit message that explains why 500ms is the right number?

martin-velay:
@rswarbrick, all your comments have been addressed now. Thanks for your review

rswarbrick (approved):
Looks sensible to me, and thanks for the explanation.

PR #24084, opened 2024-07-23
[sw, lib] Add compile-time check for log arg size to prevent undefined behaviour in logging 

AlexJones0:
When an argument (to be formatted via format specifiers in a given string) is passed to the `LOG` function either directly or indirectly, this value is then later read from the `va_list` using the `va_arg(va_list ap, type)` function. In the case of DV logging via the C->SV backdoor (bypasses UART, speeding up DV simulation) this is always read as a `uint32_t`. In the case of core logging (on Ibex), this is either `uint32_t`, `size_t`, `uintptr_t` or `status_t` (a struct storing an `int32_t`).

As a result, when a 64 bit integer is passed into some debug logging functionality without appropriate explicit casting, no implicit casting is performed as may be expected, and instead the 64 bit integer argument is read as a 32 bit width argument.

As per N1570 7.16.1.1 paragraph 2: 
> ... if type is not compatible with the type of the actual next argument (as promoted according to the default argument promotions), the behaviour is undefined, except for the following cases:
> - one type is a signed integer type, the other type is the corresponding unsigned integer type, and the value is representable in both types.
> - one type is pointer to void and the other is a pointer to a character type.

Because there is no implicit cast, any such invocations of debug printing functionality without casting 64 bit arguments are undefined behaviour, and may cause unreliable and unexpected behaviour in logging. This can compound and mislead separate debugging efforts, and so should not be allowed to occur.

This PR contains a commit that introduces additional macro functionality in order to check the type of each of these variable arguments at compile time. As a result, any code that provides an incorrect argument to debug logging functionality will result in the code not compiling with an appropriate error message via `_Generic` checks and a static assertion. This is all at compile time, with no additional runtime cost added to perform these checks. As part of this functionality, this commit expands the base library of macros to include generic macros for invoking a macro per each argument in a variable list, as well as checking the size of a given argument.

This PR also includes fixes for 8 previous cases of this undefined behaviour that were found in the code base via compiling all software targets. The command `./bazelisk.sh build --keep_going --define bitstream=skip //sw/...` was used to build all possible software targets, and the result of this was then `grep`'d to find these error cases.

### DIFF
--- a/sw/device/lib/base/macros.h
+++ b/sw/device/lib/base/macros.h
@@ -270,6 +270,153 @@ extern "C++" {
   static_assert(var == expected_value, "Unexpected value for " #var)
 
 /**
+ * A variable-argument macro that expands to the number of arguments passed into
+ * it, between 1 and 32 arguments (1 action and 0-31 others).
+ *
+ * This macro accepts a macro name and a variable list of items, and will then
+ * expand to sequentially call the macro with each provided item in order. This
+ * can be useful for performing compile-time checks on arguments in variadic
+ * functions defined via macros.
+ *
+ * For example, OT_VAR_FOR_EACH(TEST, 5, 19, 32, 1, 4) would expand to be
+ * `do {TEST(5); TEST(19); TEST(32); TEST(1); TEST(4);} while (false)`
+ *
+ * @param action The name of the macro to invoke with each item individually.
+ * @param ... The variable args list to call the macro on.
+ */
+#define OT_VA_FOR_EACH(action, ...)                                        \
+  do {                                                                     \
+    OT_CAT(OT_CAT(OT_VA_FOR_EACH_, OT_VA_ARGS_COUNT(0, ##__VA_ARGS__)), _) \
+    (action, ##__VA_ARGS__)                                                \
+  } while (false)
+
+/**
+ * The following collection of `OT_VA_FOR_EACH` macros are used to construct
+ * the generic "for each" `OT_VA_FOR_EACH` macro.
+ */
+#define OT_VA_FOR_EACH_0_(action)
+#define OT_VA_FOR_EACH_1_(action, item, ...) \
+  action(item);                              \
+  OT_VA_FOR_EACH_0_(action, ##__VA_ARGS__)
+#define OT_VA_FOR_EACH_2_(action, item, ...) \
+  action(item);                              \
+  OT_VA_FOR_EACH_1_(action, ##__VA_ARGS__)
+#define OT_VA_FOR_EACH_3_(action, item, ...) \
+  action(item);                              \
+  OT_VA_FOR_EACH_2_(action, ##__VA_ARGS__)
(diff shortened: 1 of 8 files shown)


sw/device/lib/base/macros.h
```diff
@@ -269,6 +270,164 @@ extern "C++" {
...
+
+/**
+ * A macro that checks whether a specified argument is not a standard C integer
+ * type with a width greater than 32 bits. Useful when assuming that variable
+ * arguments are 32 bits integers.
+ *
+ * @param arg An argument/expression
+ */
+#if ULONG_LONG_MAX != UINT64_MAX
+#define OT_CHECK_LT_EQ_32_BITS(arg) \
```
nbdd0121, line 401: Probably better named `OT_CHECK_NOT_INT64`, since it doesn't actually  check for size.

sw/device/lib/base/macros.h
```diff
@@ -269,6 +270,164 @@ extern "C++" {
...
+  OT_VA_FOR_EACH_30_(action, ##__VA_ARGS__)
+
+/**
+ * A macro that checks whether a specified argument is not a standard C integer
+ * type with a width greater than 32 bits. Useful when assuming that variable
+ * arguments are 32 bits integers.
+ *
+ * @param arg An argument/expression
+ */
+#if ULONG_LONG_MAX != UINT64_MAX
```
nbdd0121, line 400: I think these are never hit, regardless 32-bit or 64-bit.

sw/device/lib/runtime/log.h
```diff
@@ -98,6 +98,25 @@ void base_log_internal_core(const log_fields_t *log, ...);
...
+
+/**
+ * A macro that checks the variable arguments of the `LOG` function are valid
+ * at compile time, by asserting that each of the argument is not a standard
+ * C type with a width greater than 32 bits. Any such invalid argument will
+ * cause a relevant error via a static assertion.
+ *
+ * @param ... the variable args list
+ */
+#define CHECK_VALID_LOG_ARGS(...) \
```
nbdd0121, line 117: Probably still name this after `OT_...`?

sw/device/lib/base/macros.h
AlexJones0: That makes sense for 32-bit and 64-bit. I added this to account for a hypothetical 128-bit case where `long long` is not defined to be the same size as a 64 bit integer, but I'll opt to remove this for now for coverage related reasons as this branch will indeed never be hit by current systems.

jwnrt (commented):
It's a shame we've added non-standard format specifiers otherwise we could use `__attribute__((__format__(__printf__, 1, 2)))` on our `printf` functions and let the compiler do more sophisticated checks.

jwnrt (approved):
Thanks @AlexJones0, this looks good to me. Nice to see it caught some real errors.

engdoreis (approved):
LGTM, thanks @AlexJones0 for find this issue and fixing it.

PR #24090, opened 2024-07-23
[rv_dm,dv] Clear ndmreset_pending_q at start of ndmreset req vseq

rswarbrick:
In the rv_dm_stress_all test, we might have written to the ndmreset field of dmcontrol in a previous vseq (the smoke vseq). It shouldn't be a problem if that has happened, but we need to clear the pending ndmreset before setting up our own.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
@@ -25,19 +25,46 @@ class rv_dm_ndmreset_req_vseq extends rv_dm_base_vseq;
     `DV_CHECK(dmstatus.allunavail)
   endtask
 
-  // Check that the ndmreset_pending_q signal in the design has the value we expect.
-  task check_ndmreset_pending(bit expected_value);
+  // Read the value of the ndmreset_pending_q flag.
+  task get_ndmreset_pending(output bit rvalue);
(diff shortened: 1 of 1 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
```diff
@@ -25,19 +25,46 @@ class rv_dm_ndmreset_req_vseq extends rv_dm_base_vseq;
...
+  endtask
+
+  // Clear the ndmreset_pending_q flag if it is currently set.
+  task clear_pending_ndmreset();
+    bit ndmreset_pending;
+    get_ndmreset_pending(ndmreset_pending);
+    if (!ndmreset_pending) return;
+
+    // If rv_dm currently thinks an ndmreset is pending, it probably means that we're following a
+    // vseq set the ndmreset field in dmcontrol. To clear this, we need to set the ndmreset_ack
```
alees24, line 51: ...vseq /that?/ set the..

alees24 (approved):
Looks sensible, thanks.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
rswarbrick: Oops, thanks!

PR #24092, opened 2024-07-23
[spi_host,dv] Correct variable name in spi_host_data_stable_sva.sv

rswarbrick:
This is flagged by a warning in the VCS build. The existing code was making a variable with no driver, which meant that lines like @(negedge sck_o) in whole_cycle_data_stable_signal_checker wouldn't ever complete.


antmarzam (approved):
LGTM! Thank you

eshapira (approved):
LGTM, good catch!

rswarbrick:
Hmm, CI is actually failing with this change. @antmarzam, would you mind taking a look? I *think* this change was fixing a typo and turning a test back on, so it would be nice to figure out why it's failing!

PR #24095, opened 2024-07-23
[dv] Remove pre_reset_dly_clks argument from clk_rst_if.apply_reset

rswarbrick:
This is never given a positive value and dates back to the initial import in 2019. Probably no benefit to keeping the implementation.


marnovandermaas (approved):
Looks reasonable to me.

PR #24097, opened 2024-07-23
[dv,jtag] Simplify tck enable/release in drive_jtag_test_logic_reset

rswarbrick:
This task is only being called in a situation where we're enabling tck just afterwards anyway. Simplify things (and avoid it looking like we might disable tck when it's in use) by formalising this.


eshapira (approved):
LGTM

PR #24098, opened 2024-07-23
[rv_dm,dv] Rewrite hard reset vseq

rswarbrick:
This hasn't really changed what it does, except for dropping an unnecessary loop, using all bits of some registers and dropping a random wait.

But I'm hoping the structure of the code is a bit more obvious now.


hcallahan-lowrisc (approved):
LGTM, thanks

PR #24099, opened 2024-07-23
[rv_dm,dv] Extend description of some sec_cm testpoints

rswarbrick:
Conveniently, these are actually implemented already!


hcallahan-lowrisc (approved):
These descriptions make sense to me, thanks @rswarbrick

rswarbrick:
I just checked and the CI failure is an IT problem (bazel being deleted when trying to run a top-level test). Definitely not caused by this PR.

PR #24101, opened 2024-07-24
[top,dma] Connect SPI device to the DMA

Razer6:
This PR connects the SPI device to the DMA to be able to be used with the DMA hardware handshake mode. Doing so, this PR adds the following changes:
* Add a watermark LSIO trigger to the SPI device
* Add a SPI device LSIO trigger to the SOC Proxy
* Re-generate top to hook up the new signals


hw/ip/spi_device/rtl/spi_device.sv
```diff
@@ -509,6 +512,15 @@ module spi_device
   assign intr_fwm_rxlvl = ~rxlvl && rxlvl_d;
   assign intr_fwm_txlvl = ~txlvl && txlvl_d;
 
+  // Flop trigger signal to avoid glitches on the output
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      lsio_trigger_o <= 1'b0;
+    end else begin
+      lsio_trigger_o <= intr_fwm_txlvl | intr_fwm_rxlvl;
```
a-will, line 520: Firmware mode is not supported in this IP version, and it is removed entirely in `master` and won't return in the same named IP.

rswarbrick: Is that true in the `integrated_dev` branch too?

rswarbrick (commented):
As a minor note, do the commits need squashing a bit? The final commit edits `top_darjeeling.hjson` and a bunch of autogenerated files, some of which are `top_earlgrey.*`. Do those come from a previous commit?

hw/ip/spi_device/rtl/spi_device.sv
Razer6: I see. Then it would probably be the `readbuf_watermark` signal I think?

rswarbrick:
The problem is that our "rebase and commit" setting in GitHub means that the commits of a PR don't end up grouped together. As such, if this got merged then we'd end up with a history where some intermediate commits have stale autogenerated files.

We should probably switch to a flow where every multi-commit PR gets a merge commit added. But we haven't done so yet. In the meantime, we can't really treat PRs as atomic like that.

hw/ip/spi_device/rtl/spi_device.sv
a-will: If you want DMA for that egress buffer, I guess `readbuf_watermark` could be useful, but that feature seems rather iffy in general.

The SPI flash interface is random access and has variable-sized commands; it does not map to a FIFO style (and that's why the read buffer feature seems like an awful map to me). Does it actually make sense to have an LSIO trigger here? Is there a DMA configuration that could do useful work just from the available interrupts?

It'd be a different story for the SPI TPM(-ish) module if it actually modeled the DATA_FIFO register's underlying FIFO, but that isn't a feature we have even in `master` (yet?).

PR #24104, opened 2024-07-24
[tlt, dv] spi_device/pinmux retention fix

antmarzam:
The CSB pulse needs to be long enough (at least a whole AON clock cycle) in order to be detected by pinmux and waking up the chip. 128 and over are sizes which suffice the pulse duration needed.


antmarzam:
Hi @engdoreis / @vogelpi  - My last TLT vseq had an issue in that it could create a spi_txn "short enough" so that the CSB pulse wouldn't be detected by pinmux. 
This should sort out the issue (tested with 128)

P.S. Thanks for flagging this up @engdoreis

engdoreis (approved):
LGTM. Thanks @antmarzam for the quick fix.

vogelpi (approved):

PR #24105, opened 2024-07-24
[dv/otp_ctrl] diconnecting external VPP when reset is asserted

dkabely-ot:
This PR solve an issue in CS where using EXT_VPP, some assertions in CS implementation will throw when reset is asserted and EXT_VPP is still active.

The PR is suggested in OS code and not in CS because adding another driving point for [`otp_ext_voltage_h_io`](https://github.com/lowRISC/opentitan/blob/8a6f1d44a8107ec7b80f0e476e7018c5d657b881/hw/ip/otp_ctrl/dv/tb.sv#L88) via DV will potentially result in a race condition.


dkabely-ot:
Resolved in CS. Closing.

PR #24106, opened 2024-07-24
[dv,usbdev] Monitor fixes

alees24:
This PR fixes decoding of packets in the `usb20_monitor` and tidies up object creation via the UVM factory.
It then simplifies the collection of DATA packets and then ensures that monitor listens to d/se0 only when the DUT is transmitting, because those lines are driven only by the DUT and not `usb20_driver` so the monitor was not seeing all of the traffic.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_item.sv
+++ b/hw/dv/sv/usb20_agent/usb20_item.sv
@@ -33,10 +33,10 @@ class usb20_item extends uvm_sequence_item;
   `uvm_object_utils_begin(usb20_item)
   `uvm_object_utils_end
 
-  function new(string name = "", pkt_type_e pkt_type = PktTypeEvent);
+  function new(string name = "");
     super.new(name);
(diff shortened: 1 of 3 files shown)


hw/dv/sv/usb20_agent/usb20_item.sv
```diff
@@ -76,7 +76,7 @@ class usb20_item extends uvm_sequence_item;
   // Check whether the Packet IDentifier passes its self-check (upper and lower nibbles are bitwise
   // complementary).
   function bit valid_pid();
-    return !(m_pid_type[3:0] ^ m_pid_type[7:4]);
+    return ((m_pid_type[3:0] ^ m_pid_type[7:4]) == 4'hf);
```
rswarbrick, line 79: I spent quite some time trying to work out why this was equivalent before realising that.. it isn't! (And I agree the new version is correct)

rswarbrick (approved):
Looks sensible to me!

PR #24108, opened 2024-07-24
[hw,dma,rtl] Add STATUS.chunk_done bit

Razer6:
This PR adds a new STATUS.chunk_done bit to the status register to determine the cause when the DMA stopped an operation. In multi-chunk mem2mem transfers, where each chunk is transferred with a separate SW-go, `chunk_done` goes high when finishing every chunk, except the last chunk, where `done` goes high. The `done` interrupt fires for every chunk for now.

### DIFF
--- a/hw/ip/dma/data/dma.hjson
+++ b/hw/ip/dma/data/dma.hjson
@@ -100,9 +100,11 @@
   interrupt_list: [
     { name: "dma_done"
       desc: "DMA operation has been completed."
+      type: "status"
     }
     { name: "dma_error"
       desc: "DMA error has occurred. DMA_STATUS.error_code register shows the details."
+      type: "status"
     }
   ]
   alert_list: [
--- a/hw/ip/dma/data/dma.hjson
+++ b/hw/ip/dma/data/dma.hjson
@@ -541,7 +541,6 @@
           resval: 0x0
           hwaccess: "hrw"
           swaccess: "ro"
-          hwaccess: "hrw"
           desc: '''
                 DMA operation is active if this bit is set.
                 DMA engine clears this bit when operation is complete.
@@ -581,6 +580,16 @@
           desc: '''
                 Indicates whether the SHA2_DIGEST register contains a valid digest.
                 This value is cleared on the initial transfer and set when the digest is written.
+                '''
+        }
+        { bits: "5"
+          name: "chunk_done"
+          resval: 0x0
+          hwaccess: "hrw"
+          desc: '''
+                Transfer of a single chunk is complete.
+                Only raised for multi-chunk memory-to-memory transfers.
+                Cleared automatically by the hardware when starting the transfer of a new chunk.
                 '''
         }
       ]
--- a/hw/ip/dma/dv/env/dma_scoreboard.sv
+++ b/hw/ip/dma/dv/env/dma_scoreboard.sv
@@ -42,6 +42,10 @@ class dma_scoreboard extends cip_base_scoreboard #(
 
   // Interrupt enable state
   bit [NUM_MAX_INTERRUPTS-1:0] intr_enable;
+  // Interrupt test state (contributes to `intr_state`).
+  bit [NUM_MAX_INTERRUPTS-1:0] intr_test;
+  // Hardware  interrupt state (contributes to `intr_state`).
+  bit [NUM_MAX_INTERRUPTS-1:0] intr_state_hw;
 
   // Prediction of the state of an interrupt signal from the DUT.
   typedef struct packed {
@@ -60,7 +64,7 @@ class dma_scoreboard extends cip_base_scoreboard #(
   uint num_fifo_reg_write;
   // Variable to store clear_intr_src register intended for use in monitor_lsio_trigger task
   // since ref argument can not be used in fork-join_none
-  bit[31:0] clear_intr_src;
+  bit [31:0] clear_intr_src;
   bit [TL_DW-1:0] exp_digest[16];
 
   // Allow up to this number of clock cycles from CSR modification until interrupt signal change.
@@ -474,6 +478,7 @@ class dma_scoreboard extends cip_base_scoreboard #(
     if (item.d_error && !tl_error_suppressed) begin
       `uvm_info(`gfn, "Bus error detected", UVM_MEDIUM)
       predict_interrupts(BusErrorToIntrLatency, 1 << DMA_ERROR, intr_enable);
+      intr_state_hw[DMA_ERROR] = 1'b1;
     end else if (got_dest_item) begin
       // Is this the final destination write?
       //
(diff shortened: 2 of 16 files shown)


hw/ip/dma/rtl/dma.sv
```diff
@@ -1252,8 +1258,10 @@ module dma
     hw2reg.control.abort.de = hw2reg.status.aborted.de;
     hw2reg.control.abort.d  = 1'b0;
 
-    // interrupt management
-    hw2reg.intr_state.dma_done.de = reg2hw.status.done.q | test_done_interrupt;
+    // Interrupt management
+    // Raise the done interrupt either when finishing finishing a single chunk or the whole
```
alees24, line 1262: Repeated word 'finishing' here.

hw/ip/dma/rtl/dma.sv
```diff
@@ -1252,8 +1258,10 @@ module dma
...
+    // transfer.
+    hw2reg.intr_state.dma_done.de = reg2hw.status.done.q | chunk_done | test_done_interrupt;
```
alees24, line 1264: Both the `dma_done` and the `dma_error` interrupts should be declared as Status type interrupts in the `dma.hjson` file since the only way to deassert these interrupts and clear their bits in the `intr_state` register is to clear the bits in the `status` register, because `dma_done|error.de` is driven by `reg2hw.status.done|error.q`:

```
  interrupt_list: [
    { name: "dma_done"
      desc: "DMA operation has been completed."
      type: "status"
    }
    { name: "dma_error"
      desc: "DMA error has occurred. DMA_STATUS.error_code register shows the details."
      type: "status"
    }
```
This is a requirement of comportable IP blocks, for the CIP base tests to pass.

The present behavior of the 'intr_test' register will also be subtly incorrect, in that Status type interrupts should remain assert whilst the 'intr_test' bit is asserted. It would be simpler/easier to use the `prim_intr_hw` submodule to implement the interrupt lines.

alees24 (changes requested):
I think this does essentially what's intended, but there are issues with the interrupt-handling at present. The OpenTitan project was changed a few months ago so that Status type interrupt are specified explicitly in the IP block description and all blocks were updated accordingly; see my comments below, please.

Razer6:
@alees24 Thanks for the review. Could you take another look? CI is currently failing because the `integrated_dev` branch is not able to correctly create the auto-generated unittests for status interrupts, which is fixed on master. The plan is to rebase that PR on top of master and merge it there once the DMA is in there.

hw/ip/dma/rtl/dma.sv
```diff
@@ -1119,8 +1143,6 @@ module dma
                             transfer_remaining_bytes : chunk_remaining_bytes;
 
   always_comb begin
-    hw2reg = '0;
-
```
alees24, line 1123: It is not safe to drop this assignment at present because the following signals will be left undriven/undefined:
```
hw2reg.control.initial_transfer.d
hw2reg.control.initial_transfer.de
hw2reg.sha2_digest[i].d
```
In particular the `control.initial_transfer` signals must be driven at all times; simulations fail presently because `initial_transfer.q` becomes undefined as soon as a transfer starts.

It is perhaps worth leaving the default assignment in place (I may have missed other signals) and annotating that it is required for driving the above signals, as well as avoiding latches/other undriven signals.

An alternative would be to ensure that the above signals are always driven explicitly.

hw/ip/dma/data/dma.hjson
```diff
@@ -558,11 +559,21 @@
                 '''
         }
         { bits: "2"
+          name: "chunk_done"
+          resval: 0x0
+          hwaccess: "hrw"
```
alees24, line 564: As per comments on other reviews, are you(pl) happy with breaking backwards compatibility by inserting a new status bit rather than keeping the existing bits in their previous locations?

alees24 (changes requested):
Thanks for making the changes to the interrupt handling, and using the `prim_intr_hw` module. That looks fine.

If you resolve the issue with the default `hw2reg` assignment and are happy to sign off the compatibility breakage by _inserting_ a bit into the status register then I will approve this.

hw/ip/dma/rtl/dma.sv
Razer6: Right, I had to explicitly drive these signals since the the `prim_intr`'s added a second driver to the `hw2reg`. Thus, the default statement would cause a conflicting driver.

hw/ip/dma/data/dma.hjson
Razer6: It was changed to keep the existing bit positions and add the new one at the end.

hw/ip/dma/dv/env/dma_scoreboard.sv
```diff
@@ -953,6 +949,14 @@ class dma_scoreboard extends cip_base_scoreboard #(
         index = get_index_from_reg_name(csr.get_name());
         dma_config.intr_src_wr_val[index] = item.a_data;
       end
+      "status": begin
+        bit done, error;
+        done = get_field_val(ral.status.done, item.a_data);
+        error = get_field_val(ral.status.error, item.a_data);
+        // Clearing the status bits also clears the status interrupt
+        predict_interrupts(CSRtoIntrLatency, done << DMA_DONE & `gmv(ral.intr_enable), 0);
+        predict_interrupts(CSRtoIntrLatency, error << DMA_ERROR & `gmv(ral.intr_enable), 0);
```
Razer6, line 958: With the new status interrupts, I had to re-predict the interrupts when writing the status register, ie., clerating the status.

Razer6:
@alees24 Thanks for the review. Could you take another look to get this merged? 

I updated the DV environment to deal with the status interrupts. All tests except the automated `dma_intr_test` (comfortable test) fail. Could you take a look into this one? Otherwise, maybe. @rswarbrick. Note, tests need to include https://github.com/lowRISC/opentitan/pull/24962 to pass.

alees24:
The RTL changes look okay, thanks, but I'm seeing more than just `dma_intr_test` failing presently:
```
0.dma_abort.88760512117758134865029955350941779515590544128422615357238515974507785526752\
0.dma_intr_test.67386857437561253008700698343085724055010273966100030887847157593273296727411\
0.dma_memory_region_lock.62058440537178885937077771562078581177386725377790603349600473204830182595305\
```
I'll try to investigate these shortly, but if you happen to get to it before I do, please try those seeds; busy with another project release right now.

Razer6:
It seems there is an issue on how predicted interrupts are getting cleared. When a DMA gets aborted, the interrupt prediction queue is left untouched. After some while, a previously predicted interrupt expectation gets asserted.

There is a second issue where the abort occurs just before a done. It seems there is an interrupt predicted although the abort has higher preiority.

Razer6:
Thanks @alees24 for looking into this. I cherry picked your changes into that PR.

alees24 (approved):
Thanks for the changes. LGTM.

Issue #22705, opened 2024-04-18
[entropy_src] Integration and top-level testing extensions


vogelpi:
### Description

During previous discussions it was mentioned that we extend the top-level / SiVal tests of ENTROPY_SRC / the entropy complex to cover the following things:
- [ ] In Firmware Override - Extract & Insert mode, test if things keep working in case a significant delay is added between extracting and inserting entropy, see https://github.com/lowRISC/opentitan/issues/20952#issuecomment-2027811199.
- [ ] In Firmware Override - Extract & Insert mode, test if things keep working if ENTROPY_SRC is disabled between producing seeds WITHOUT disabling CSRNG and the EDNs. See https://github.com/lowRISC/opentitan/issues/20952#issuecomment-2034273059 and https://github.com/lowRISC/opentitan/issues/22293#issuecomment-2210526844

Please add more stuff here if needed rather than creating new issues / re-open previously closed issues.

johngt:
@h-filali / @vogelpi - do you mind adding a rough estimate for time needed to resolve this?
Looking at the issue the related PRs are merged suggesting that this could be closed out?

vogelpi:
Moving to M6 for now as the risk here is low and we have other, more important items for M5.

PR #24109, opened 2024-07-24
[chip-level/sw] Extend the entropy_src_fw_override_test

h-filali:
This commit extends the entropy_src_fw_override_test to also test whether the entropy can be inserted with a significant delay after reading from the observe FIFO. Another extension is for testing whether the entropy_src can be reenabled without reenabling the rest of the entropy complex.

I tested it on FPGA and the test passes.
`bazel test //sw/device/tests:entropy_src_fw_override_test_fpga_cw310_sival_rom_ext --test_output=streamed --nocache_test_results`

I'm not sure what qualifies as a "significant" delay. I just did 128 micro seconds. For 100Mhz this should be ~12800 cycles.
That should probably be enough but it's still worth discussing.

Resolves [#22705](https://github.com/lowRISC/opentitan/issues/22705)

### DIFF
--- a/sw/device/tests/entropy_src_fw_override_test.c
+++ b/sw/device/tests/entropy_src_fw_override_test.c
@@ -18,6 +18,7 @@
 #include "sw/device/lib/testing/edn_testutils.h"
 #include "sw/device/lib/testing/entropy_testutils.h"
 #include "sw/device/lib/testing/kmac_testutils.h"
+#include "sw/device/lib/testing/rand_testutils.h"
 #include "sw/device/lib/testing/rv_core_ibex_testutils.h"
 #include "sw/device/lib/testing/test_framework/check.h"
 #include "sw/device/lib/testing/test_framework/ottf_main.h"
@@ -55,6 +56,16 @@ enum {
    * and blocks the ISR.
    */
   kMaxOutputWords = 16,
+  /**
+   * Number of micro seconds we wait between having extracted words from the
+   * observe FIFO and inserting words back into the entropy source.
+   */
+  kOutputDelayUs = 128,
+  /**
+   * The size of the buffer used in firmware to process the entropy bits in
+   * firmware override mode.
+   */
+  kEntropyFifoBufferSize = 12,
 };
 
 static dif_aes_t aes;
@@ -91,6 +102,8 @@ static size_t words_to_output = 0;
 
 static dif_edn_auto_params_t edn_params0;
 static dif_edn_auto_params_t edn_params1;
+
+static bool fw_ov_insert_wait_enabled = false;
 
 /**
  * Determine whether the observe FIFO has overflowed.
@@ -152,6 +165,10 @@ void ottf_external_isr(uint32_t *exc_info) {
 
     words_to_input = kFifoBufferSizeWords;
     words_to_output = kFifoBufferSizeWords;
+  }
+
+  if (fw_ov_insert_wait_enabled) {
+    busy_spin_micros(kOutputDelayUs);
   }
 
   if (words_to_output > 0) {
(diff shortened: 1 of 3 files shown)


sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -55,6 +56,11 @@ enum {
    * and blocks the ISR.
    */
   kMaxOutputWords = 16,
+  /**
+   * Number of micro seconds we wait between having extracted words from the
+   * observe FIFO and inserting words back into the entropy source.
+   */
+  kOutputdelay = 128,
```
vogelpi, line 63: Don't we usually add the unit for such constants? I.e., should this be `kOutputDelayUs`?

sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -258,13 +270,60 @@ static status_t entropy_config(
...
+          .fips_enable = true,
+          .fips_flag = true,
+          .rng_fips = true,
+          .route_to_firmware = false,
+          .bypass_conditioner = bypass_conditioner,
+          .single_bit_mode = single_bit_mode,
+          .health_test_threshold_scope = false,
+          .health_test_window_size = kEntropySrcHealthTestWindowSize,
+          .alert_threshold = UINT16_MAX},
+      kDifToggleEnabled));
```
vogelpi, line 310: Right now, you are reconfiguring the ENTROPY_SRC. It's seems to work but there is a risk of configuring it differently. Instead you can just do this:
```
TRY(dif_entropy_src_set_enabled(&entropy_src, kDifToggleDisabled));
```
and then
```
TRY(dif_entropy_src_set_enabled(&entropy_src, kDifToggleEnabled));
```
This will simply flip the enable field back and forth.

sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -258,13 +270,60 @@ static status_t entropy_config(
...
+
+  // Enable the interrupt handler to provide CSRNG with entropy.
+  irq_external_ctrl(true);
```
vogelpi, line 313: I think this routine should be changed to:
1. First disable the entropy source.
2. Do some Ibex requests until a new seed is requested.
3. Wait for the CSRNG interrupt (`cs_entropy_req`).
4. Handle the interrupt by re-enabling ENTROPY_SRC and inject the seed.

sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -377,10 +442,23 @@ bool test_main(void) {
 
   for (size_t i = 0; i < ARRAYSIZE(kModes); i++) {
     EXECUTE_TEST(test_result, firmware_override_extract_insert, kModes[i],
+                 false, false);
+    EXECUTE_TEST(test_result, firmware_override_extract_insert, kModes[i], true,
                  false);
-    EXECUTE_TEST(test_result, firmware_override_extract_insert, kModes[i],
-                 true);
   }
+  // Rerun the test with single bit mode disabled, this time we reenable the
```
vogelpi, line 449: Thanks for adding these two test cases on top of the existing test cases. I think it's good to keep testing the previous cases in their original form.

vogelpi (commented):
Thanks @h-filali for the PR. It looks mostly good, the enabling/disabling can actually be simplified quite a bit and I think it's worth trying to only re-enable the entropy source after CSRNG has requested a new seed.

Also, can you please update the test description in `chip_entropy_src_testplan.hjson`?

sw/device/tests/entropy_src_fw_override_test.c
h-filali: You're right, I'll add the unit.

sw/device/tests/entropy_src_fw_override_test.c
h-filali: Fair enough!

sw/device/tests/entropy_src_fw_override_test.c
h-filali: Okay, I'll change it.

sw/device/tests/entropy_src_fw_override_test.c
h-filali: Makes sense, thanks!

h-filali:
@vogelpi I added some checks that ensure that CSRNG requests data while Entropy_src is diabled. Feel free to have a look.

sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -258,13 +275,85 @@ static status_t entropy_config(
...
+    CHECK_DIF_OK(dif_entropy_src_fw_override_sha3_start_insert(
+        &entropy_src, kDifToggleDisabled));
+    // Clear CS_ENTROPY_REQ and check if it is reasserted.
+    // If not, we can disable Entropy_src and check if the entropy
+    // complex still produces entropy after reenabling Entropy_src.
+    mmio_region_write32(csrng.base_addr, CSRNG_INTR_STATE_REG_OFFSET,
+                        (uint32_t)(1 << CSRNG_INTR_STATE_CS_ENTROPY_REQ_BIT));
+    data = mmio_region_read32(csrng.base_addr, CSRNG_INTR_STATE_REG_OFFSET);
+    entropy_req_intr =
+        bitfield_bit32_read(data, CSRNG_INTR_STATE_CS_ENTROPY_REQ_BIT);
```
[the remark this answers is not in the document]
h-filali: Thanks, should be fixed now.

sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -258,13 +275,83 @@ static status_t entropy_config(
...
+    CHECK_DIF_OK(dif_entropy_src_fw_override_sha3_start_insert(
+        &entropy_src, kDifToggleEnabled));
+    for (int i = 0; i < kEntropyFifoBufferSize; i++) {
+      kInputMsg[i] = cnt++;
+    }
+    CHECK_DIF_OK(dif_entropy_src_fw_ov_data_write(
+        &entropy_src, (const uint32_t *)kInputMsg, kEntropyFifoBufferSize,
+        &words_written));
+    CHECK_DIF_OK(dif_entropy_src_fw_override_sha3_start_insert(
+        &entropy_src, kDifToggleDisabled));
```
vogelpi, line 315: Nit: this triggers the SHA3 operation which may take up to 48 (CS AES Halt interface) + 25 (padding) + 24 (SHA3 rounds) cycles to finish. If the CS_ENTROPY_REQ bit is still high after executing this DIF function it's most likely because the SHA3 hasn't finished yet. To avoid this, you could poll the FSM state register of the ENTROPY_SRC before acking and checking the interrupt again.

But it's not that dramatic, worst case you just run this loop here once more and "produce" one seed that is not necessary.

vogelpi (approved):
Thanks @h-filali , this looks good :-)

PR #24110, opened 2024-07-24
[otbn,dv] Avoid forcing between add/addc, sub/subb when carry=0

rswarbrick:
This sometimes happens randomly and the RTL doesn't spot the error. But we don't care because the operations are actually equivalent in this situation! Tighten the constraints on the input to avoid forcing the signal like that.


alees24 (approved):
Thanks for the helpful explanatory comments.

PR #24113, opened 2024-07-24
[otbn,dv] Factor out tasks in otbn_sec_wipe_err_vseq.sv

rswarbrick:
No functional change, but this should hopefully make the structure a little more obvious.


hcallahan-lowrisc (approved):
I didn't check every single line, but the overall reformatting looks clean and makes sense. Thanks!

PR #24114, opened 2024-07-24
[rom_ctrl,dv] Explicitly cast the return value of get_invalid_mubi4

rswarbrick:
This fixes an ENUMERR build failure when build with Xcelium.


eshapira (approved):
LGTM

PR #24116, opened 2024-07-24
Pull shared objects out of Hyperdebug Transport Inner member

zhangarino:
This is to remove the cyclic referenced between Hyperdebug Transport Targets and the Inner struct, allowing for Targets to be dropped if they and the Transport object both go out of scope.

This is useful for creating another Transport object after the previous one has been dropped. Otherwise the target objects stay alive forever preventing equivalents from being created the next time a Transport object is created.

### DIFF
--- a/sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
+++ b/sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
@@ -56,6 +56,7 @@ pub struct Hyperdebug<T: Flavor> {
     i2c_interface: Option<BulkInterface>,
     cmsis_interface: Option<BulkInterface>,
     uart_interfaces: HashMap<String, UartInterface>,
+    cached_io_interfaces: CachedIo,
     inner: Rc<Inner>,
     current_firmware_version: Option<String>,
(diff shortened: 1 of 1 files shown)


jesultra (changes requested):
Please see my comments on the other review:
https://github.com/lowRISC/opentitan/pull/24118

zhangarino:
Updated with the suggestions from #24118 by just moving the cached references into a new struct that `HyperdebugTransport` directly owns - agreed that the change is much cleaner this way, thanks.

If possible it would be nice to have the change cherry-picked back to earlgrey_es_sival, just because we're currently doing development using that branch. Although the patch is relatively small so it could just be something we maintain on our mirror on top of the upstream.

jesultra (approved):
Nice, one file changed instead of eight. Now it is straightforward to see everything that this CL does.

I have applied a tag to request automatic cherry-picking to earlgrey_es_sival as soon as this one is merged to master.

I will be on vacation Monday-Wednesday next week, but I want to give the other reviewers a chance to post comments.  I think I will merge this in the evening CET, unless I hear objections.

PR #24117, opened 2024-07-24
Pull shared objects out of Hyperdebug Transport Inner member

zhangarino:
This is to remove the cyclic referenced between Hyperdebug Transport Targets and the Inner struct, allowing for Targets to be dropped if they and the Transport object go out of scope.

This is useful for creating another Transport object after the previous one has been dropped. Otherwise the target objects stay alive forever preventing equivalents from being created the next a Transport object is created.


zhangarino:
Apologies for the noise, wrong branch.

PR #24118, opened 2024-07-24
Earlgrey es sival: Pull shared objects out of Hyperdebug Transport Inner member

zhangarino:
This is to remove the cyclic referenced between Hyperdebug Transport Targets and the Inner struct, allowing for Targets to be dropped if they and the Transport object both go out of scope.

This is useful for creating another Transport object after the previous one has been dropped. Otherwise the target objects stay alive forever preventing equivalents from being created the next a Transport object is created.

### DIFF
--- a/sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
+++ b/sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
@@ -52,6 +52,7 @@ pub use ti50::Ti50Flavor;
 /// Nucleo-L552ZE-Q.
 pub struct Hyperdebug<T: Flavor> {
     console: CommandHandler,
+    usb_device: Rc<RefCell<UsbBackend>>,
     spi_interface: BulkInterface,
     i2c_interface: Option<BulkInterface>,
     cmsis_interface: Option<BulkInterface>,
@@ -68,7 +69,7 @@ pub trait Flavor {
     fn gpio_pin(console: &CommandHandler, pinname: &str) -> Result<Rc<dyn GpioPin>>;
     fn spi_index(
         _console: &CommandHandler,
-        _inner: &Rc<Inner>,
+        _usb_device: &Rc<RefCell<UsbBackend>>,
         instance: &str,
     ) -> Result<(u8, u8)> {
         bail!(TransportError::InvalidInstance(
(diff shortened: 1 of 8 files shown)


sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
```diff
@@ -572,6 +572,17 @@ impl Inner {
     }
 }
 
+/// Internal state of the Hyperdebug struct, this struct is reference counted such that Gpio,
```
jesultra, line 575: This comment is no longer accurate.  As far as I can tell, nothing other than the main `Hyperdebug` struct will have reference to this new `Inner` struct, so there is no point in using `Rc<Inner>` up there.  In fact, I think that the five fields here should just become direct members of the `Hyperdebug` struct.

sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
```diff
@@ -404,29 +414,19 @@ impl<T: Flavor> Hyperdebug<T> {
...
         Ok(capabilities)
     }
 }
 
-/// Internal state of the Hyperdebug struct, this struct is reference counted such that Gpio,
-/// Spi and Uart sub-structs can all refer to this shared data, which is guaranteed to live on,
-/// even if the caller lets the outer Hyperdebug struct run out of scope.
-pub struct Inner {
+#[derive(Clone)]
+pub struct CommandHandler {
```
jesultra, line 425: I think this CL would become a lot smaller, if we left the three fields `console_tty`, `usb_device` and `selected_spi` as only three members of an `Inner` struct, which would be reference counted.  This would be according to a convention that the `Inner` struct contains the all driver state that needs to be shared between any instances of SPI `Target`, `GpioPin`, or similar, which may out-live the main `Hyperdebug` struct.  It would be a shame to have to touch all the same places again, if we in the future need a fourth mutable field to be shared this way.  (Also, see my comment below on the "new" `Inner` struct.)

jesultra (changes requested):
There could be some reason why what I suggest would not work out in practice, if so, please let me know.

jesultra:
I only now realize that this PR is for the eralgrey_es_sival branch, and duplicate of PR 24116.  Let's do the main review in the other PR, when that has been worked out, we can decide whether to use GitHub's cherry-pick feature, or whether a separate PR like this is warranted.  (And even whether we want the refactoring back-ported at all.)

jesultra:
As #24195 has merged, think PR should be abandoned.

PR #24124, opened 2024-07-25
[chip_earlgrey_asic] Enable pull-up setting for power-on reset

vogelpi:
As requested by the PD team.

This resolves lowRISC/OpenTitan#24132.

### DIFF
--- a/hw/top_earlgrey/rtl/autogen/chip_earlgrey_asic.sv
+++ b/hw/top_earlgrey/rtl/autogen/chip_earlgrey_asic.sv
@@ -1089,7 +1089,7 @@ module chip_earlgrey_asic #(
   assign manual_oe_otp_ext_volt = 1'b0;
 
   // Enable schmitt trigger on POR for better signal integrity.
-  assign manual_attr_por_n = '{schmitt_en: 1'b1, default: '0};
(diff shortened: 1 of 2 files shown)


msfschaffner (approved):
LGTM

meisnere (approved):

moidx:
Should we update the RTL blocker to pick up these paths? (This can be done in a follow up PR).

moidx:
We should also open an ECO issue to document this change. I can take care of this, but it is something that can be completed post-merge.

vogelpi:
> Should we update the RTL blocker to pick up these paths? (This can be done in a follow up PR).

Good catch! I'll take care of this.

vogelpi:
> We should also open an ECO issue to document this change. I can take care of this, but it is something that can be completed post-merge.

Yes, I agree. I've opened an issue here #24132 and this PR will again resolve it.

sha-ron:
@rswarbrick @matutem, now that there is internal pull-up on POR pad, shouldn't we remove the pull-up from DV test bench at https://github.com/lowRISC/opentitan/blob/master/hw/top_earlgrey/dv/env/chip_if.sv#L116 ?

PR #24130, opened 2024-07-25
[rom_ctrl,dv] Minor tidy-ups in smoke vseq

rswarbrick:
This was driven by me trying to fix (very sporadic) smoke test failures. Annoyingly, they still happen very sporadically after this change, but the test has got rather quicker and is slightly easier to read.


eshapira (approved):
LGTM, thanks!

PR #24133, opened 2024-07-25
[dv,usbdev] Fix usbdev_rand_suspends sequence

alees24:
For circa 10% of seeds the link state prediction lagged the observation and reporting of Resume Signaling by the DUT (an artefact of the fact that usb20_monitor is sampling on a fixed clock even throughout the Suspend Signaling).


rswarbrick (approved):
Thanks for the detailed comment! :-)

PR #24134, opened 2024-07-25
[rv_dm,dv] Tidy up cmderr sequence and testpoint

rswarbrick:
This should now be a bit quicker and (more importantly) it becomes a bit more obvious what it does.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -273,22 +273,7 @@ class rv_dm_base_vseq extends cip_base_vseq #(
       end
     join_none
   endtask
-  task write_chk (input uvm_object ptr,input int cmderr,input int command);
-    uvm_reg_data_t data;
-    uvm_reg_data_t rdata;
-    repeat ($urandom_range(1, 10)) begin
-      data = $urandom;
-      request_halt();
-      cfg.clk_rst_vif.wait_clks($urandom_range(0, 1000));
-      csr_wr(.ptr(jtag_dmi_ral.command), .value(command));
-      csr_wr(.ptr(ptr), .value(data));
-      cfg.clk_rst_vif.wait_clks($urandom_range(0, 1000));
-      csr_rd(.ptr(jtag_dmi_ral.abstractcs), .value(rdata));
-      cfg.clk_rst_vif.wait_clks($urandom_range(0, 1000));
-      `DV_CHECK_EQ(cmderr,get_field_val(jtag_dmi_ral.abstractcs.cmderr,rdata));
-      cfg.clk_rst_vif.wait_clks($urandom_range(1, 10));
-    end
-  endtask
+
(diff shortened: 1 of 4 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
```diff
@@ -405,4 +390,40 @@ class rv_dm_base_vseq extends cip_base_vseq #(
...
+  // Clear the cmderr field of abstractcs.
+  task clear_cmderr();
+    // To clear the field, we use uvm_reg_field.predict() to convince the model that it has an error
+    // that needs clearing, then call uvm_reg_field.set() with 3'b111 to set desired value to zero
+    // (passing 3'b111 rather than zero because the field is R/W1C), then finally call update.
+    uvm_status_e status;
+
+    `DV_CHECK_FATAL(jtag_dmi_ral.abstractcs.cmderr.predict(3'b111));
+    jtag_dmi_ral.abstractcs.cmderr.set(3'b111);
+    jtag_dmi_ral.abstractcs.update(.status(status));
```
vogelpi, line 409: Are you sure this is working? My understanding is that `update()` only performs a write if the value that `set` is different from the current value of the model. And the current value of the model is set to the same value above using `predict`. Of course, I am no expert and I might be mixing up things here.

vogelpi (approved):
This looks good @rswarbrick . I have a question regarding predicting, setting and then updating the RAL. I am probably mixing up something. It would appreciate if you could quickly clarify the situation :-)

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
rswarbrick: A good point, but it *is* actually correct, I think. The point is that this field is RW1C. There's a note a couple of lines above, mainly because this took me several goes (with waves) to get right when I initially did it, in commit 626ef9bb092a04c58e20dd2d4a2b501dd1846c0a. (This PR is just moving that code to the base sequence)

A thing that I only understood last week the exact meaning of `my_reg.set(1)`. It updates the desired value to be what you'd get if you wrote 1. If the register is RW1C, this means that it will clear the bit in the desired value. 

This is different from `my_reg.predict(1)`. There, the prediction "kind" matters. It defaults to `UVM_PREDICT_DIRECT` which means "updated the mirrored value to match the value given". (If it was `UVM_PREDICT_WRITE` it would mean "update the value as if we just wrote 1", which would set it to zero in this case. If it was `UVM_PREDICT_READ` it would mean the same as `UVM_PREDICT_DIRECT` in this setting).

Thus a RW1C clear register can be cleared with `predict(1); set(1); update()` because it means "Believe that the register has value 1. Then update the desired value to be what we'd get if we wrote 1. Now send a bus write to make the change happen."

Yes, this is quite confusing.

vogelpi: Ah, now this makes sense. Thanks a lot for explaining @rswarbrick !

PR #24136, opened 2024-07-25
[rv_dm,dv] Rewrite and simplify rv_dm_dataaddr_rw_access_vseq

rswarbrick:
This test is pretty trivial: it's just supposed to be checking that values written over TL to a certain pair of registers are visible over DMI.

This rewrite makes the code a bit easier to read. The behaviour is still the same.


eshapira (approved):
LGTM

PR #24137, opened 2024-07-25
[rv_dm,dv] Rewrite rv_dm_hart_unavail_vseq to make it easier to read

rswarbrick:
No functional change, but it should be a bit simpler to read what's going on.


hcallahan-lowrisc (approved):
Much easier to follow. Thanks

PR #24138, opened 2024-07-25
[rv_dm,dv] Expand the implementation of mem_tl_access_resuming

rswarbrick:
This expands the code a bit to make it clearer what's going on. It also drops the loop: that doesn't seem to make all that much sense (and nor does a randomised resume request).

Honestly, this is really a cut-down version of
rv_dm_halt_resume_whereto_vseq. That sequence tries to request an abstract command and then checks the whereto register contains the start of that command.

It probably makes sense to keep this (more simple) test as well, partly because the code is dramatically easier to understand.


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_mem_tl_access_resuming_vseq.sv
```diff
@@ -2,38 +2,69 @@
...
-    repeat ($urandom_range(1, 10)) begin
-      wdata = $urandom_range(0,1);
-      // Verify that writing to RESUMING results in anyresumeack and
-      // allresumeack to be set.When resumereq bit is set,
-      // it will clear the anyresumeack and allresumeack.
-      request_halt();
-      csr_rd(.ptr(jtag_dmi_ral.dmstatus), .value(rdata));
-      `DV_CHECK_EQ(1, get_field_val(jtag_dmi_ral.dmstatus.anyhalted, rdata))
-      `DV_CHECK_EQ(1, get_field_val(jtag_dmi_ral.dmstatus.allhalted, rdata))
-      cfg.clk_rst_vif.wait_clks($urandom_range(0, 1000));
```
marnovandermaas, line 25: How come you removed this random wait?

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_mem_tl_access_resuming_vseq.sv
```diff
@@ -2,38 +2,69 @@
...
-      csr_wr(.ptr(jtag_dmi_ral.dmcontrol.haltreq), .value(0));
-      csr_wr(.ptr(tl_mem_ral.resuming), .value(0));
-      cfg.clk_rst_vif.wait_clks($urandom_range(0, 1000));
```
marnovandermaas, line 28: How come you removed this random wait?

marnovandermaas (approved):
From my code inspection, the new implementation also does what the old one does without the randomization. I did leave a few comments on randomized delays, not sure if those are useful to add back in or not.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_mem_tl_access_resuming_vseq.sv
rswarbrick: This wait represents a time between the debugger seeing that a halt request has been passed through to the core and deciding to clear the halt request. I don't think it's really needed: the tightest timing is most likely to cause problems.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_mem_tl_access_resuming_vseq.sv
rswarbrick: I think this code was a bit broken before. In that version:
- The core responds to a resume request
- We wait a bit
- We make the resume request (that the core has already responded to)

With the new version, I've fixed the ordering. I don't think we need a wait here, but we can always revisit that in the future (now the logic is correct) if necessary.

PR #24140, opened 2024-07-25
[rv_dm,dv] Expand the rv_dm_progbuf_read_write_execute test

rswarbrick:
Not much of a functional change, except that I've made it so that now we write all the multiregisters before reading them all back (to make sure that they aren't all backed by a single register).

I've also added comments to make it as clear as possible what's going on.


hcallahan-lowrisc (approved):
Thanks for the nice comments, this LGTM

PR #24143, opened 2024-07-26
[BLOCKFILE] Add top_earlgrey.sv and chip_earlgrey_asic.sv files

vogelpi:
For some reason, we did not have `chip_earlgrey_asic.sv` in the BLOCKFILE nor the template used to generate it. Including the template is not needed as this also impacts the FPGA chip-level files. But we should at least include the generated ASIC file.


vogelpi:
CHANGE AUTHORIZED: BLOCKFILE

rswarbrick:
CHANGE AUTHORIZED: BLOCKFILE

rswarbrick (approved):

PR #24145, opened 2024-07-26
[ci] Create new nightly job for Verilator tests

AlexJones0:
This PR adds a new job to the nightly azure pipeline to run Verilator test targets. All (non-broken) Verilator tests are run nightly regardless of their individual timeout, with Bazel as the test orchestrator. The job timeout is tentatively set at 180 until a run might be able to reveal a more appropriate timeout value.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,10 +30,7 @@ trigger:
   tags:
     include:
     - "*"
-pr:
-  branches:
-    include:
-    - "*"
+pr: none
 
 jobs:
 - job: checkout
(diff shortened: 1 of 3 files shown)


AlexJones0:
I've temporarily changed this PR to disable CI and run nightlies. This will be changed before it can be merged, and is just for testing purposes.

AlexJones0:
Due to issues with the amount of memory available on Azure pipelines images not being enough to run Verilator for our testing purposes, I'm running all the nightly tests today on a `ci-public` agent overnight to see how long they take when we set them up properly on CI. If this is not completed by tomorrow morning then it will be judged whether these should be canceled or not based on CI load and the timing/progress showed by the logs.

PR #24148, opened 2024-07-26
[dv,usbdev] Coverage improvements

alees24:
This PR addresses a couple of small oversights in the coverage collection and extends two sequences to provide more extensive testing.
In changing the sequence `usbdev_data_toggle_restore` a subtle discrepancy between the functional model and the DUT was also discovered; the DUT routinely collects packet data into the buffer memory before deciding whether or not the packet can be accepted. This is because of the serial nature of the USB transmission and the absence of any internal storage beyond a few flip-flops and the packet buffer memory itself.

### DIFF
--- a/hw/ip/usbdev/dv/env/usbdev_env_cov.sv
+++ b/hw/ip/usbdev/dv/env/usbdev_env_cov.sv
@@ -194,10 +194,10 @@ class usbdev_env_cov extends cip_base_env_cov #(.CFG_T(usbdev_env_cfg));
       bins five  = {5};
       // Plus those where the additional bytes of the CRC16 cause the total byte count to near or
       // exceed 64.
-      bins sixty_one   = {61};
-      bins sixty_two   = {62};
-      bins sixty_three = {63};
-      bins sixty_four  = {64};
+      bins max_len_m3 = {MaxPktSizeByte-3};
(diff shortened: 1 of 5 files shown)


hw/ip/usbdev/dv/env/usbdev_env_cov.sv
```diff
@@ -215,7 +215,10 @@ class usbdev_env_cov extends cip_base_env_cov #(.CFG_T(usbdev_env_cfg));
       bins nak = {PidTypeNak};
     }
     cp_dir: coverpoint dir_in;  // 'dir_in' is set to indicate IN
-    cp_endp: coverpoint endp;
+    cp_endp: coverpoint endp {
+      // Device supports only 12 endpoints.
```
rswarbrick, line 219: Nit: Would it also make sense to say `NEndpoints` rather than 12 in the comment?

rswarbrick (approved):
This looks good to me!

One nitty comment and I notice that Verible is warning about line length in the last commit. But I'm happy otherwise.

hw/ip/usbdev/dv/env/usbdev_env_cov.sv
alees24: Thanks. I opted to leave the literal in the comment so that people don't have to look up its value. It's not really likely to change.

PR #24151, opened 2024-07-27
[rv_dm,dv] Improve the coverage we get from reading WhereTo address

rswarbrick:
These are minor tweaks in `rv_dm_halt_resume_whereto_vseq` to try to improve condition coverage numbers.

It feels a little bit silly to go much further, because the conditions and branches in `dm_mem` are all about deciding what instruction to manufacture for the core to execute. I don't think it would make much sense to put more effort into seeing lots of those instructions, only to ignore the values that come out (since `rv_dm` testing doesn't really care about what instructions get generated by the debug module).

A sensible approach might be to waive a large tract of the relevant conditional coverage and note that we're not really trying to test the internals of the pulp debug module. But that's definitely a problem for a different PR.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_halt_resume_whereto_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_halt_resume_whereto_vseq.sv
@@ -59,7 +59,8 @@ class rv_dm_halt_resume_whereto_vseq extends rv_dm_base_vseq;
     dm::command_t cmd;
 
     ar_cmd.aarsize  = 3'h2;
-    ar_cmd.postexec = 1'b1;
+    ar_cmd.postexec = $urandom_range(0, 1);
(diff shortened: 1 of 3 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_halt_resume_whereto_vseq.sv
```diff
@@ -71,51 +71,53 @@ class rv_dm_halt_resume_whereto_vseq extends rv_dm_base_vseq;
   task body();
     int unsigned hartsel = 0;
 
-    repeat (2) begin
```
vogelpi, line 73: Doing things twice or doing one more thing after the actual sequence can help detecting cases where the sequence leaves the DUT in a weird state. But I agree, we now have the stress_all sequence running which help to discover such issues. So this is not required anymore.

vogelpi (approved):
This looks good to me @rswarbrick , thanks for the useful explanations!

PR #24152, opened 2024-07-27
[rv_dm,dv] Send and abort a DMI operation

rswarbrick:
This is driven by a coverage hole (where the dap FSM doesn't jump from "Read" or "Write" to "Idle"). I'm not really sure how much this situation could ever matter, but it's not too difficult to exercise so let's just do it.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_dmi_failed_op_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_dmi_failed_op_vseq.sv
@@ -18,6 +18,36 @@ class rv_dm_dmi_failed_op_vseq extends rv_dm_base_vseq;
     return ret;
   endfunction
 
+  // Send a DMI operation and abort it by issuing trst_n. This will cancel the operation, so it will
+  // have no effect.
+  task send_interrupted_dmi_operation(jtag_dmi_op_req_e op);
+    bit [6:0] data0_addr = 4;
+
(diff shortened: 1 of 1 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_dmi_failed_op_vseq.sv
```diff
@@ -18,6 +18,33 @@ class rv_dm_dmi_failed_op_vseq extends rv_dm_base_vseq;
     return ret;
   endfunction
 
+    // Send a DMI operation and abort it by issuing trst_n. The operation should be cancelled with
```
hcallahan-lowrisc, line 21: Do we check that there is indeed no effect via scoreboarding or some other mechanism already?

nit. indent on comment

rswarbrick: We don't have any register model, so we don't have any "general proof" yet. Conveniently, I *am* certain in this particular case: the interrupted operation is writing a random value to `data0`, whose uncorrupted value we check later in the sequence. Or at least, I will be certain once I've switched the aborted write with the following `csr_wr`. Will do that now!

rswarbrick: ... and doing so showed that I wasn't cleaning up a DMI error flag (oops). That's all fixed in the latest force-push.

hcallahan-lowrisc (approved):
LGTM

PR #24153, opened 2024-07-27
[sphincsplus-rs] fix pre-hashed mode signing

vbendeb:
In SPHINCS+ pre-hashed signing mode a properly wrapped hash of the input message is passed to the signer, not the entire message.

Tested on CW310 by enabling SPX signing of the Ti50 image and verifying the image by the ROM_EXT.

### DIFF
--- a/sw/host/opentitantool/src/command/image.rs
+++ b/sw/host/opentitantool/src/command/image.rs
@@ -285,9 +285,16 @@ impl CommandDispatch for ManifestUpdateCommand {
         }
         // Sign with SPX+.
         if let Some(key) = spx_private_key {
-            image.add_manifest_extension(ManifestExtEntry::new_spx_signature_entry(
-                &image.map_signed_region(|buf| key.sign(self.domain, buf))??,
-            )?)?;
+            let sig_bytes = match self.domain {
+                SpxDomain::None | SpxDomain::Pure => {
+                    image.map_signed_region(|buf| key.sign(self.domain, buf))??
+                }
+                SpxDomain::PreHashedSha256 => {
+                    let digest = image.compute_digest()?.to_le_bytes();
+                    key.sign(self.domain, &digest)?
+                }
+            };
+            image.add_manifest_extension(ManifestExtEntry::new_spx_signature_entry(&sig_bytes)?)?;
         }
 
         // Offline signing takes place if signatures are provided.


vbendeb:
For Ti50 the plan is to be SPX signing in pre-hashed mode to facilitated distributed signer architecture, this patch fixes a signer bug.

vbendeb:
Updated the test case to match the code change.

sw/host/sphincsplus/key.rs
```diff
@@ -51,7 +52,16 @@ impl SpxDomain {
         match self {
             Self::None => message.into(),
             Self::Pure => [&[0u8, 0u8], message].concat().into(),
-            Self::PreHashedSha256 => [&Self::SHA256_DOMAIN, message].concat().into(),
+            Self::PreHashedSha256 => {
```
cfrantz, line 55: I don't think this is the proper place to perform the digest operation:
- The enumerated name is "PreHashed".  The message should _already_ be the sha256 hash.
- The signing flow in our build automation already computes the hash via the `opentitantool image digest` command.

Creating the SPX signature with prehashed data can be accomplished with the following sequence (or to be more precise, it _should_ work the following way):
1. `opentitantool image manifest update [...]` to set the manifest values.
2. `opentitantool image digest --output=digest.bin <input-file>` creates the sha256 digest.  This digest is appropriate for both ECDSA and SPX+ signatures.
3. `opentitantool spx sign --domain=PreHashedSha256 --message=digest.bin --output=signature.bin [...]` signs the message with the appropriate domain constants.
4. `opentitantool image manifest update --spx-signature=signature.bin [...]` to attach the signature onto the binary.

The `manifest update` command probably needs some updates to handle signing in a single invocation, but the above sequence needs to work to implement the detached signature flow.

I think the hash computation in your change more properly belongs in `opentitantool/src/command/image.rs` around line 288:
```
        if let Some(key) = spx_private_key {
            image.add_manifest_extension(ManifestExtEntry::new_spx_signature_entry(
// Here: `map_signed_region` for "Pure" or "None" schemes and
// `key.sign(&image.compute_digest())` for the pre-hashed scheme.
             &image.map_signed_region(|buf| key.sign(self.domain, buf))??,
            )?)?; 
        }
```

vbendeb: I thought about this too, but what seems inconsistent to me is this:

when the private key is passed to `opentititantool image manifest update ...` for all other signing modes the full image is passed as the singing input. 

What should be the correct invocation for `opentitantool image manifest update` when the passed in key is SPX and the signing domain is prehashed?

Or are we saying that for any other case when the private key is available we can use `image manifest update` but not for the case of SPX/prehashed?

cfrantz: When private keys are passed to `image manifest update`, the full image is _always_ passed to the CLI as an input argument.  The implementation of that command then invokes appropriate code to perform the signature over the relevant part of the image.
- For RSA and ECDSA keys, this is accomplished by digesting over the signed segments of the image by `image.compute_digest()`.
- For SPX+ "Pure" mode, `map_signed_region` creates a slice of the relevant portion of the image and that slice is passed to the signing function.
- For SPX+ "PreHashed" mode, the SPX signature is supposed to be over a SHA256 hash (ie: the same value computed by `image.compute_digest()`).  In the pre-hashed case, that value should be passed to the signing function.

The correct invocation for `image manifest update` when the passed in key is SPX and the domain is prehashed should be something like:
```
opentitantool image manifest update \
  --spx-key=/path/to/private_key.pem \
  --domain=PreHashedSha256 \
  [other options] \
  --output=my_image.signed.bin
  my_image.bin
```
However, the "sign in a single invocation" use case, while convenient, is not realistic. Most customers are going to be keeping their keys in an HSM or other signing infrastructure. In fact, we don't even use the "single invocation" version of signing in the OpenTitan codebase.

Because the detached signature use case is what customers will use, it what is currently implemented in OpenTitan's bazel signing rules.  The sequence of commands is the same the sequence I outlined in my first comment: update manifest properties, generate digest, sign digest, attach signature.  (This also allows us to easily swap the local "fake key" signature for a token provided signature for any build target by specifying cmdline args to bazel).

I think this is all that is needed to make this work (perhaps with a bit of massaging to ref-taking or error returns):
```
        if let Some(key) = spx_private_key {
            image.add_manifest_extension(ManifestExtEntry::new_spx_signature_entry(
              match self.domain {
                SpxDomain::None | SpxDomain::Pure => &image.map_signed_region(|buf| key.sign(self.domain, buf))??,
                SpxDomain::PreHashedSha256 => &key.sign(self.domain, image.compute_digest()?)?,
              }
            )?)?; 
        }
```

vbendeb:
```
        if let Some(key) = spx_private_key {
            image.add_manifest_extension(ManifestExtEntry::new_spx_signature_entry(
              match self.domain {
                SpxDomain::None | SpxDomain::Pure => &image.map_signed_region(|buf| key.sign(self.domain, buf))??,
                SpxDomain::PreHashedSha256 => &key.sign(self.domain, image.compute_digest()?)?,
              }
            )?)?; 
        }
```
looks like a good alternative (even though the fact that  the domain is both examined at call site and passed to the signing function is a bit odd).

There was also the issue of the digest endianness - the embedded code sha256 digest calculation did not seem to match the host side, verification started succeeding when I reverted the digest on the host.

But this all besides the point. I agree that the main use case would be detached signature case, but for a case when the private key is available (all kind of dev builds, FPGA, etc), we do want to be consistent when invoking opentitantool and passing the private key, right?

I can update the patch to make use of your suggestion, but we need to agree that the 'one shot signing' should be supported through the same CLI invocation for all signature schemes, right?

cfrantz: Right about the endianness - I don't recall which endian `compute_digest` returns; an endian reversal may be required.

One-shot signing should be supported for all signature schemes.  The change I've suggested should make that the case for SPX+ signatures when using the `PreHashedSha256` domain (and it should already be true for the Pure and None domains).  It also should not break the detached signature flow when using `opentitantool spx sign`.

vbendeb:
PTAL

sw/host/opentitantool/src/command/image.rs
```diff
@@ -285,8 +285,17 @@ impl CommandDispatch for ManifestUpdateCommand {
...
         if let Some(key) = spx_private_key {
+            // Just in case the pre-hashed domain is used, speculatively prepare
+            // the hash so that it is available at the right scope.
+            let digest = image.compute_digest().unwrap().to_le_bytes();
             image.add_manifest_extension(ManifestExtEntry::new_spx_signature_entry(
-                &image.map_signed_region(|buf| key.sign(self.domain, buf))??,
+                &image.map_signed_region(|buf| {
+                    let msg: &[u8] = match self.domain {
+                        SpxDomain::PreHashedSha256 => &digest,
+                        _ => buf,
```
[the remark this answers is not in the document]
vbendeb: I think using a single `key.sign()` invocation would be more readable, but not hashing when not necessary is also a good advantage, done.

cfrantz (approved):

cfrantz:

PR #24154, opened 2024-07-27
Rom ext spx

vbendeb:


vbendeb:
this is half baked, hopefully  could be used as a basis and extended with proper key management and test modifications.

cfrantz:
Here is an implementation (for `earlgrey_1.0.0`) that uses spx keys from the ownership configuration: #25083.  Specifically, this implements hybrid signing, where an application firmware needs to be signed with both an ECDSA key and SPHINCS+ key.

PR #24155, opened 2024-07-29
[DV] update flash vseq and aon c code file.

thouri:
Flash vseqs - small update for flash control mp region.
aon c code - update timer range to include upper and lower limits.

### DIFF
--- a/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_mp_regions_vseq.sv
+++ b/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_mp_regions_vseq.sv
@@ -248,6 +248,9 @@ class flash_ctrl_mp_regions_vseq extends flash_ctrl_base_vseq;
                                           flash_op_data.size == flash_op.num_words;)
     `uvm_info("do_mp_reg", $sformatf("flash_op: %p", flash_op), UVM_MEDIUM)
     bank = flash_op.addr[OTFBankId];
+    if  (flash_op.partition != FlashPartData) begin
+      illegal_trans |= check_info_part(flash_op, "check_info_tran");
+    end
     // For ctrl read, you have to validate per Qword
     if (flash_op.op == FlashOpRead) begin
       int is_odd = flash_op.addr[2];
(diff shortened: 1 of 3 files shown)


eshapira (approved):

andreaskurth:
@matute Could you PTAL at this change (even post merge)? TIA

vogelpi (commented):
This is only post merge but the C change (aon_timer_irq_test) looks good. For the flash_ctrl DV change I feel that I don't have sufficient context. @matutem could you take a look please?

matutem:
Regarding flash, I wonder if this change will work when a flash read spans more than one page since the page attributes are per page.

vogelpi:
@thouri @eshapira , we've discussed in the meeting with @sha-ron today that we should follow-up on this. Can you check whether this works if a flash read spans multiple pages with different attributes please?

matutem:
Or simply revert the change in flash_ctrl. I find it strange that this change was needed.

eshapira:
Hi, sorry I just respond now.
This change made because the method **check_info_part** is already taking into account the number of words of the read operation and in turn the end_page of the operation (see line 227):
https://github.com/lowRISC/opentitan/blob/e45ccd274ab25e38bf93ded01399110379e8e30a/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_base_vseq.sv#L225-L235

Following the above:

- I don't see why there should be any issue when reading across 2 pages, since that should be covered by the number of words and the end_page. The possible issue may rise only if reading through more than 2 pages in which the middle pages attributes will be skipped, means at least 514 words in single read operation.
In addition the only 2 things this method is checking is the special partitions of pages 1-3 and overflow, so following this I guess this is problematic only if the first page will be page 0 in bank0 and the last page will be page 4, which means the operation size should be at least 1538 bus words.
Is this scenario possible and allowed? If so, maybe we can change something in the method to catch this.

- Now, the 2 reasons why this change is needed:

1. Since the **check_info_part** method is already taking care of the end_page, no need to reevaluate the operation for every word.
2. The real issue that can cause false failures: Since the end_page is reevaluated for every word, its evaluation is overflowing and wrong.
For the first word: end_page = page(first_word+num_of_words-1).
For the second word: end_page = page(first_word+num_of_words)
For the last word: end_page = page(last_word+num_of_words-1) = page(first_words+2num_of_words - 2).

I did it after this test failed with the following scenario: address=0x4fe0 , num_of_words=6 (24bytes), While the info partition size is 0x5000 (10 pages of 256 8bytes words).
So with this condition, the last bus_word should be 0x4ff4 which is legal address within the same page as the start_adress.
However because of this issue, when evaluating address 0x4ff0, the **check_info_part** method reevaluated the end_page to be page(0x4ff0+24-1) = page(0x5007) = 11, which is bigger than the number of info pages, so the test set to expect the recov_err alert after this.

matutem:
Thanks for the explanation. I also noticed we check for access permissions separately in validate_flash_op, so all is covered.

PR #24156, opened 2024-07-29
[dv] Add GUI debug mode for Xcelium

martin-velay:
Add a GUI debug mode for Xcelium. Turn on debug features:
  - breakpoints
  - live data
  - transaction recording

And run simulation up until the end of the UVM elaboration phase to have access to the dynamic objects.

This mode is available through the dvsim command line arguments: `--gui-debug` or `-gd`

### DIFF
--- a/hw/dv/tools/common.tcl
+++ b/hw/dv/tools/common.tcl
@@ -21,6 +21,12 @@ if {[info exists ::env(GUI)]} {
   set gui "$::env(GUI)"
 }
 
+set gui_debug 0
+# Detect when GUI debug mode has been invoked to build UVM objects by default
+if {[info exists ::env(GUI_DEBUG)]} {
+  set gui_debug "$::env(GUI_DEBUG)"
+}
+
 set tb_top "tb"
 if {[info exists ::env(TB_TOP)]} {
   set tb_top "$::env(TB_TOP)"
--- a/hw/dv/tools/dvsim/vcs.hjson
+++ b/hw/dv/tools/dvsim/vcs.hjson
@@ -271,6 +271,13 @@
       run_opts: ["-gui", "-l {run_dir}/simv.log"]
     }
     {
+      name: vcs_gui_debug
+      // TODO as done for Xcelium see PR #24156
+      is_sim_mode: 1
+      build_opts: ["-debug_access+all+reverse"]
+      run_opts: ["-gui", "-l {run_dir}/simv.log"]
+    }
+    {
       name: vcs_waves
       is_sim_mode: 1
       build_opts: [// Enable generating Verdi Knowledge Database
--- a/hw/dv/tools/sim.tcl
+++ b/hw/dv/tools/sim.tcl
@@ -38,9 +38,17 @@ if {$simulator eq "xcelium"} {
   assertion -list -depth all -multiline -permoff $tb_top
 }
 
-# In GUI mode, let the user take control of running the simulation.
 global gui
-if {$gui == 0} {
+global gui_debug
+# In GUI mode, let the user take control of running the simulation.
+# In GUI debug mode, run only up until the end of the UVM elaboration phase to have access to the
+# dynamic objects.
+if {$gui_debug == 1} {
+  if {$simulator eq "xcelium"} {
+    uvm_phase -stop_at build -end
+    run
+  }
+} elseif {$gui == 0} {
   run
   if {$simulator eq "xcelium"} {
     # Xcelium provides a `finish` tcl command instead of `quit`. The argument '2' enables the
--- a/util/dvsim/Deploy.py
+++ b/util/dvsim/Deploy.py
@@ -137,7 +137,7 @@ class Deploy():
         """
         self._extract_attrs(self.sim_cfg.__dict__)
 
-        # Enable GUI mode.
+        # Enable GUI mode, also when GUI debug mode has been invoked.
         self.gui = self.sim_cfg.gui
 
         # Output directory where the artifacts go (used by the launcher).
--- a/hw/dv/tools/dvsim/common_modes.hjson
+++ b/hw/dv/tools/dvsim/common_modes.hjson
@@ -13,6 +13,11 @@
       en_build_modes: ["{tool}_gui"]
     }
     {
+      name: gui_debug
+      is_sim_mode: 1
+      en_build_modes: ["{tool}_gui_debug"]
+    }
+    {
       name: waves
       is_sim_mode: 1
       en_build_modes: ["{tool}_waves"]
--- a/hw/dv/tools/dvsim/common_sim_cfg.hjson
+++ b/hw/dv/tools/dvsim/common_sim_cfg.hjson
@@ -77,6 +77,7 @@
     { dv_root: "{dv_root}" },
     { SIMULATOR: "{tool}" },
     { GUI: "{gui}"},
+    { GUI_DEBUG: "{gui_debug}"},
     { WAVES: "{waves}" },
     { DUT_TOP: "{dut}" },
     { TB_TOP: "{tb}" },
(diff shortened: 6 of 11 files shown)


martin-velay:
This is not working with coverage collection, needs to be investigated

marnovandermaas:
This looks like a really cool PR!

martin-velay:
Anybody has time to review this?

hw/dv/tools/dvsim/vcs.hjson
```diff
@@ -270,6 +270,13 @@
       build_opts: ["-debug_access+all+reverse"]
       run_opts: ["-gui", "-l {run_dir}/simv.log"]
     }
+    {
+      name: vcs_gui_debug
+      // TODO as done for Xcelium see PR #24156
+      is_sim_mode: 1
+      build_opts: ["-debug_access+all+reverse"]
+      run_opts: ["-gui", "-l {run_dir}/simv.log"]
+    }
```
rswarbrick, line 279: If we don't think we have a special "gui debug mode" for VCS, I suggest we just drop this hunk?

hw/dv/tools/questa/sim.tcl
```diff
@@ -6,8 +6,8 @@
 # Remove leading "# " from the front of log file lines and run the test if not in gui mode.
 # This provides compatibility for log file error checking with other supported simulators within Opentitan.
 set gui 0
-if {[info exists ::env(GUI)]} {
-  set gui "$::env(GUI)"
+if {$::env(GUI)==1 || $::env(GUI_DEBUG)==1} {
+  set gui 1
```
rswarbrick, line 10: Is this change needed? Have you got a way to test it with a Questa licence? (I haven't!)

hw/dv/tools/sim.tcl
```diff
@@ -38,9 +38,17 @@ if {$simulator eq "xcelium"} {
...
-if {$gui == 0} {
+global gui_debug
+# In GUI mode, let the user take control of running the simulation.
+# In GUI debug mode, run only up until the end of the UVM elaboration phase to have access to the
+# dynamic objects.
+if {$gui_debug == 1} {
+  if {$simulator eq "xcelium"} {
+    uvm_phase -stop_at build -end
+    run
+  }
```
rswarbrick, line 50: Shouldn't the `run` line be outside of the inner `if` block?

hw/dv/tools/sim.tcl
```diff
@@ -38,9 +38,17 @@ if {$simulator eq "xcelium"} {
...
+} elseif {$gui == 0 && $gui_debug == 0} {
```
rswarbrick, line 51: I'd probably drop the `gui_debug` check here. At the moment, it's correct (but unnecessary) provided that `$gui_debug` is always 0 or 1. I'm not sure it does something sensible if `gui_debug` is 2...

If you drop that check here, we end up treating `gui_debug=cabbage` as `gui_debug=0`. Alternatively, we could change line 46 to be `$gui_debug != 0` (so treating silly values as 1) or add a check to make sure the value is something sensible.

util/dvsim/Deploy.py
```diff
@@ -137,8 +137,8 @@ def _set_attrs(self):
         """
         self._extract_attrs(self.sim_cfg.__dict__)
 
-        # Enable GUI mode.
-        self.gui = self.sim_cfg.gui
+        # Enable GUI mode, also when GUI debug mode has been invoked.
+        self.gui = self.sim_cfg.gui or self.sim_cfg.gui_debug
```
rswarbrick, line 141: Couldn't we do things more easily if we ensured `gui_debug` implied `gui`?

hw/dv/tools/sim.tcl
martin-velay: Yes you're right, I can drop it. I think `gui_debug` cannot be anything else than `0/1` as this is defined in this script `hw/dv/tools/common.tcl` and not by the user himself.

hw/dv/tools/dvsim/vcs.hjson
martin-velay: This will maybe be implemented in the future and it was to have a clearer error message from `util/dvsim/SimCfg.py` file:
```
        # GUI mode is only available for Xcelium for the moment.
        if (self.gui_debug) and (self.tool not in ['xcelium']):
            log.error("GUI debug mode is only available for Xcelium, please remove "
                      "--gui_debug / -gd option or switch to Xcelium tool.")
            sys.exit(1)
```
Do you still reckon to drop it?

util/dvsim/Deploy.py
martin-velay: That's would be better indeed. I have added this in `FlowCfg.py` file:
```
        if self.gui_debug:
            self.gui = 1
```
Is that OK for you?

hw/dv/tools/common.tcl
```diff
@@ -17,8 +17,15 @@ if {[info exists ::env(WAVES)]} {
 }
 
 set gui 0
-if {[info exists ::env(GUI)]} {
-  set gui "$::env(GUI)"
+# Detect when GUI mode has been invoked to prevent any auto-run
+if {$::env(GUI)==1} {
```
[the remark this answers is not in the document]
martin-velay: I have reverted this piece of code

hw/dv/tools/common.tcl
```diff
@@ -17,8 +17,15 @@ if {[info exists ::env(WAVES)]} {
...
+  set gui 1
+}
+
+set gui_debug 0
+# Detect when GUI debug mode has been invoked to build UVM objects by default
+if {$::env(GUI_DEBUG)==1} {
```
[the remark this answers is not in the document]
martin-velay: I have changed it to be similar to what is done in simple GUI mode

martin-velay:
@rswarbrick
Are you aware of Python rule deck changes? Because I have CI lint issue on some piece of code I haven't touch:
https://github.com/lowRISC/opentitan/actions/runs/11661355403/job/32465489675?pr=24156

hw/dv/tools/dvsim/vcs.hjson
rswarbrick: I think that if you drop the hunk in `vcs.hsjson`, the error message from `SimCfg.py` will still come out? That's probably the right behaviour, I think?

hw/dv/tools/sim.tcl
rswarbrick: Ah, ok, that makes sense. Thanks!

util/dvsim/Deploy.py
rswarbrick: That sounds like a good idea to me.

rswarbrick:
About the lint failure: how annoying! I'm not sure when the check changed, but the flake8 check only runs on changed files (which is why it's being triggered by this PR).

I'll file something in a minute that should fix these two items.

rswarbrick:
The change in #24998 should sort things out.

hw/dv/tools/dvsim/vcs.hjson
martin-velay: No, I am getting this error only:
`ERROR: [modes] Sub mode "vcs_gui_debug" added to mode "gui_debug" was not found!`

rswarbrick: Oh, I see. I guess the point is that we try to resolve all modes, even ones that aren't actually being used. In which case, maybe this is the right stub!

martin-velay:
@rswarbrick, thanks again for your review. I think all the comments have been addressed.

martin-velay:
> Cool. I've just merged the dvsim code that should fix the flake8 warnings. Can you rebase? (We should end up with green CI)

Thanks Rupert! Hopefully it will work now 🤞

martin-velay:
@rswarbrick, CI is passing now after the rebase. Thanks!

rswarbrick (approved):
Thanks, this looks good!

Issue #23455, opened 2024-06-03
[hmac,doc] Align documentation with changes


gdessouky:
### Description

Make a pass over the entire HMAC documentation and programmer's guide and make sure it is aligned with all the recent changes and new features added, e.g.,
- block diagrams
- fix latencies for different digest sizes
- context switching/one-shot calls 
- how `hmac_done` and `idle` are expected to work

andreaskurth:
Proposing to move to M6 to resolve this as part of D3

vogelpi:
@gdessouky , can you maybe give an update on the status of this item please?

PR #24157, opened 2024-07-29
[hmac,doc] Align documentation with HW changes

gdessouky:
This implements minor fixes to the HMAC documentation: minor wording edits and alignment with changes to the IP and its SW usage.

This should close https://github.com/lowRISC/opentitan/issues/23455 once merged.

### DIFF
--- a/hw/ip/hmac/README.md
+++ b/hw/ip/hmac/README.md
@@ -15,33 +15,34 @@ See that document for integration overview within the broader OpenTitan top leve
 
 ## Features
 
-- HMAC supporting multiple digest sizes: SHA-2 256/384/512 hashing algorithm
-- HMAC-SHA-2 and unkeyed SHA-2 dual mode
+- Two modes: SHA-2 | HMAC based on SHA-2
+- Multiple digest sizes supported (for both modes): SHA-2 256/384/512 hashing algorithm
 - Configurable key length 128/256/384/512/1024-bit secret key for HMAC mode
-- 32 x 32-bit message buffer
+- Support for context switching (via saving and restoring) across multiple message streams
+- 32 x 32-bit message FIFO buffer
 
 ## Description
 
 [sha256-spec]: https://csrc.nist.gov/publications/detail/fips/180/4/final
 
 The HMAC module is a [SHA-2][sha256-spec] hash-based authentication code generator to check the integrity of an incoming message and a signature signed with the same secret key.
-It supports SHA-2 256/384/512 and 128/256/384/512/1024-bit keys in HMAC mode, so long as the key length does not exceed the block size of the configured SHA-2 mode, i.e., 1024-bit keys are not supported for SHA-2 256 where the block size is 512-bit.
+It supports SHA-2 256/384/512 and 128/256/384/512/1024-bit keys in HMAC mode, so long as the key length does not exceed the block size of the configured SHA-2 digest size, i.e., 1024-bit keys are not supported for SHA-2 256 where the block size is 512-bit.
 It generates a different authentication code with the same message if the secret key is different.
 
 This HMAC implementation is not hardened against side channel or fault injection attacks.
 It is meant purely for hashing acceleration.
 If hardened MAC operations are required, users should use either [KMAC](../kmac/README.md) or a software implementation.
 
-The 1024-bit secret key is written in [`KEY_0`](doc/registers.md#key) to [`KEY_31`](doc/registers.md#key), and the key length relevant to the HMAC operation is configured in [`CFG.key_length`].
-For example, to use a 256-bit secret key, [`CFG.key_length`] should be configured to 0x02 and then only secret key registers [`KEY_0`](doc/registers.md#key) to [`KEY_7`] are read and relevant for the HMAC operation.
-The digest size required is configured in [`CFG.digest_size`].
-The message to authenticate is written to [`MSG_FIFO`](doc/registers.md#msg_fifo) and the HMAC generates a 256/384/512-bit digest value (depending on the digest size configuration provided) which can be read from [`DIGEST_0`](doc/registers.md#digest) to [`DIGEST_7`](doc/registers.md#digest) for SHA-2 256, or from [`DIGEST_0`] to [`DIGEST_12`] for SHA-2 384, or from [`DIGEST_0`] to [`DIGEST_15`] for SHA-2 512.
+The secret key is written in [`KEY_0-KEY_31`](doc/registers.md#key), and the key length relevant to the HMAC operation is configured in [`CFG.key_length`](doc/registers.md#cfg--key_length).
+For example, to use a 256-bit secret key, [`CFG.key_length`](doc/registers.md#cfg--key_length) needs to be configured (as per register documentation) and then only the relevant secret key registers, only [`KEY_0-KEY_7`](doc/registers.md#key) in this case, are consumed for the HMAC operation.
+The digest size required is configured in [`CFG.digest_size`](doc/registers.md#cfg--digest_size).
+The message to authenticate is written to [`MSG_FIFO`](doc/registers.md#msg_fifo) and the HMAC generates a 256/384/512-bit digest value (depending on the digest size configuration provided) which can be read from [`DIGEST_0-DIGEST_7`](doc/registers.md#digest) for SHA-2 256, or from [`DIGEST_0-DIGEST_12`](doc/registers.md#digest) for SHA-2 384, or from [`DIGEST_0-DIGEST_15`](doc/registers.md#digest) for SHA-2 512.
 The `hmac_done` interrupt is raised to report to software that the final digest is available.
 
 This module allows software to save and restore the hashing context so that different message streams can be interleaved; please check the [Programmer's Guide](doc/programmers_guide.md#saving-and-restoring-the-context) for more information.
 
 The HMAC IP can run in SHA-2 only mode, whose purpose is to check the correctness of the received message.
-The same digest registers above are used to represent the hash result.
+The same digest registers above are used to hold the final hash result.
 SHA-2 mode does not use the given secret key.
 It generates the same result with the same message every time.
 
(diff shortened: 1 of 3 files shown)


martin-velay (approved):
Thanks Ghada for your clarifications and details.
It's hard to say if something is missing, but it LGTM.

hw/ip/hmac/README.md
```diff
@@ -15,41 +15,42 @@ See that document for integration overview within the broader OpenTitan top leve
 
 ## Features
 
-- HMAC supporting multiple digest sizes: SHA-2 256/384/512 hashing algorithm
-- HMAC-SHA-2 and unkeyed SHA-2 dual mode
+- Dual modes: HMAC based on SHA-2 and unkeyed SHA-2
```
ballifatih, line 18: nit: I would drop the `unkeyed` term here. I think SHA-2 clearly refers to the hash function and in the standard definition there is no key input for a hash function.

hw/ip/hmac/README.md
```diff
@@ -15,41 +15,42 @@ See that document for integration overview within the broader OpenTitan top leve
...
 It generates a different authentication code with the same message if the secret key is different.
 
 This HMAC implementation is not hardened against side channel or fault injection attacks.
 It is meant purely for hashing acceleration.
 If hardened MAC operations are required, users should use either [KMAC](../kmac/README.md) or a software implementation.
 
-The 1024-bit secret key is written in [`KEY_0`](doc/registers.md#key) to [`KEY_31`](doc/registers.md#key), and the key length relevant to the HMAC operation is configured in [`CFG.key_length`].
-For example, to use a 256-bit secret key, [`CFG.key_length`] should be configured to 0x02 and then only secret key registers [`KEY_0`](doc/registers.md#key) to [`KEY_7`] are read and relevant for the HMAC operation.
+The secret key is written in [`KEY_0`](doc/registers.md#key) to [`KEY_31`](doc/registers.md#key), and the key length relevant to the HMAC operation is configured in [`CFG.key_length`].
+For example, to use a 256-bit secret key, [`CFG.key_length`] should be configured to 0x02 and then only secret key registers [`KEY_0`](doc/registers.md#key) to [`KEY_7`] are consumed and are relevant for the HMAC operation.
```
ballifatih, line 37: nit: Is the "are consumed and are relevant" part intentional. Two "are"s sounds a bit unnatural to me.

hw/ip/hmac/doc/programmers_guide.md
```diff
@@ -1,8 +1,9 @@
 # Programmer's Guide
 
-This chapter shows how to use the HMAC-SHA-2 IP by showing some snippets such as initialization, initiating SHA-2 or HMAC process and processing the interrupts.
+This chapter shows how to use the HMAC/SHA-2 IP by showing some snippets such as initialization, initiating SHA-2 or HMAC process and processing the interrupts.
 This code is not compilable but serves to demonstrate the IO required.
-More detailed and complete code can be found in the software under `sw/`, [ROM code](https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/lib/drivers/hmac.c) and [HMAC DIF](https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/dif/dif_hmac.c).
+More detailed and complete code can be found in the software under `sw/`, [ROM code](https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/lib/drivers/hmac.c) and [cryptolib code](https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/crypto/drivers/hmac.c).
```
ballifatih, line 5: I would actually drop the reference to the silicon creator driver here and only link the regular cryptolib link. The implementation from silicon driver is raw, meaning that there are no checks to enforce correct streaming expectations (i.e. feeding message bytes in block granularity). If people start using those as baseline for further development, they are going to have to spend time trying to debug unnecessary issues (this actually happened).

ballifatih (approved):
LGTM. My comments are all nits, so feel free to merge this without further changes. They could be useful to keep in mind for future edits.

hw/ip/hmac/README.md
andreaskurth: Maybe:
```suggestion
- Dual modes: SHA-2 / HMAC based on SHA-2
```
?

hw/ip/hmac/README.md
andreaskurth, line 37: I wouldn't mention the specific configuration value (`0x02`) in this file, because it duplicates information that can get stale. Instead, I suggest sth like:
```suggestion
For example, to use a 256-bit secret key, [`CFG.key_length`](doc/registers.md#cfg--key_length) needs to be configured (as per register documentation) and then only secret key registers [`KEY_0` to `KEY_7`](doc/registers.md#key) are consumed and relevant for the HMAC operation.
```

hw/ip/hmac/README.md
```diff
@@ -15,41 +15,42 @@ See that document for integration overview within the broader OpenTitan top leve
...
 It generates a different authentication code with the same message if the secret key is different.
 
 This HMAC implementation is not hardened against side channel or fault injection attacks.
 It is meant purely for hashing acceleration.
 If hardened MAC operations are required, users should use either [KMAC](../kmac/README.md) or a software implementation.
 
-The 1024-bit secret key is written in [`KEY_0`](doc/registers.md#key) to [`KEY_31`](doc/registers.md#key), and the key length relevant to the HMAC operation is configured in [`CFG.key_length`].
-For example, to use a 256-bit secret key, [`CFG.key_length`] should be configured to 0x02 and then only secret key registers [`KEY_0`](doc/registers.md#key) to [`KEY_7`] are read and relevant for the HMAC operation.
+The secret key is written in [`KEY_0`](doc/registers.md#key) to [`KEY_31`](doc/registers.md#key), and the key length relevant to the HMAC operation is configured in [`CFG.key_length`].
```
andreaskurth, line 36: ``[`CFG.key_length`]`` doesn't result in an implicit link, unfortunately. `(doc/registers.md#cfg--key_length)` needs to be appended. This problem also occurs in existing lines in this file; may I ask you to please fix them too in this PR?

hw/ip/hmac/README.md
```diff
@@ -15,41 +15,42 @@ See that document for integration overview within the broader OpenTitan top leve
 
 ## Features
 
-- HMAC supporting multiple digest sizes: SHA-2 256/384/512 hashing algorithm
-- HMAC-SHA-2 and unkeyed SHA-2 dual mode
+- Dual modes: HMAC based on SHA-2 and unkeyed SHA-2
+- Multiple digest sizes supported (for both modes): SHA-2 256/384/512 hashing algorithm
 - Configurable key length 128/256/384/512/1024-bit secret key for HMAC mode
+- Support for context switching (via saving and restoring) across multiple message streams
 - 32 x 32-bit message buffer
```
andreaskurth, line 22: How about
```suggestion
- 32 x 32-bit message FIFO buffer
```
because the next reference to this buffer links to `MSG_FIFO`?

hw/ip/hmac/README.md
andreaskurth: In this context, I would bundle "`KEY_0` to `KEY_31`" into a single link (and similar for "`DIGEST_0` to `DIGEST_15`" etc)

hw/ip/hmac/README.md
```diff
@@ -15,41 +15,42 @@ See that document for integration overview within the broader OpenTitan top leve
...
 SHA-2 mode does not use the given secret key.
 It generates the same result with the same message every time.
 
 The software does not need to provide the message length. The HMAC IP
 will calculate the length of the message received between **1** being written to
 [`CMD.hash_start`](doc/registers.md#cmd) and **1** being written to [`CMD.hash_process`](doc/registers.md#cmd).
 
-This version does not have many defense mechanisms but is able to wipe internal variables such as the secret key, intermediate hash results H, digest and the message FIFO.
+This version does not have many defense mechanisms but is able to wipe internal variables such as the secret key, intermediate hash results, digest and the internal message scheduling array.
 It does not wipe the software accessible 32x32b FIFO.
```
andreaskurth, line 54:
```suggestion
It does not wipe the message FIFO, which SW can write but not read.
```

hw/ip/hmac/doc/programmers_guide.md
andreaskurth: And I'd use relative links to files in the repository (e.g., `../../../sw/device/lib/crypto/drivers/hmac.c`) rather than links to the GitHub repo

hw/ip/hmac/doc/theory_of_operation.md
```diff
@@ -118,8 +119,10 @@ In case of SHA-2 256 with 256-bit key, as the digest result is 256-bit, it must
 
 In the second round, the message length is a fixed 768 bits (512-bit size of outer secret key + 256-bit first round digest size).
 
-HMAC supports secret key of length 256/384/512/1024-bit, so long as the key length does not exceed the block size of the configured digest, i.e., for SHA-2 256 a maximum length of 512-bit key is supported.
-It is up to the software to shrink the key to the supported key length (up to 512-bit for SHA-2 256 and up to 1024-bit for SHA-2 384/512) using a hash function when setting up the HMAC.
+HMAC supports a secret key of length 128/256/384/512/1024-bit, so long as the key length does not exceed the block size of the configured digest, i.e., for SHA-2 256 a maximum length of 512-bit key is supported.
+The byte order of the key registers is big-endian by default, can be swapped to little endian by setting [`CFG.key_swap`](registers.md#cfg) to 1.
+To support any arbitratry key length, the software should configure the HMAC to the next largest supported key length, e.g. for an 80-bit key, HMAC should be configured with an 128-bit key length and fed with the 80-bit key.
```
andreaskurth, line 124: Small typo :-)
```suggestion
To support any arbitrary key length, the software should configure the HMAC to the next largest supported key length, e.g. for an 80-bit key, HMAC should be configured with an 128-bit key length and fed with the 80-bit key.
```

andreaskurth (commented):
Couple of smaller comments, otherwise LGTM. Thx for these thorough updates, @gdessouky!

hw/ip/hmac/doc/programmers_guide.md
gdessouky: @ballifatih agree with your concern, but I  would keep the reference to the silicon creator just because it helps (me at least if I am onboarding into this) to know all the different instances in the project where the HW IP is being used by the SW, but I would explicitly mention that this is not the reference/baseline SW implementation, but to refer to the cryptolib for that.

gdessouky:
Thanks @ballifatih and @andreaskurth for all your comments, I think I addressed them all in my last force push!

andreaskurth (approved):
All good, thx @gdessouky and all reviewers!

PR #24158, opened 2024-07-29
[rv_dm,dv] Expand sec_cm_dm_en_ctrl_lc_gated testpoint and implement it

rswarbrick:
Detailed comments in individual commit messages. The second commit turns out to be necessary because the DV environment imagined "is debug enabled?" to be a yes/no question. Very reasonable! But not correct if we're forcing internal signals.

### DIFF
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_buffered_enable_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_buffered_enable_vseq.sv
@@ -1,0 +1,288 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+//
+//
+// This sequence forces a copy of the multi-bit debug enable signal to On, enabling the feature
+// enabled by that copy.
+//
+// Note that it's implausible that an attacker will be able to force this multi-bit signal
+// directly. But there might be a single bit somewhere downstream that is attackable. Allowing the
+// attacker to force at this level is a safe overapproximation of their capabilities.
+//
+// For some copies, the multi-bit signal gets divided into further buffered copies. For those
+// copies, allowing the top multi-bit signal to be forced is going to strictly over-approximate the
+// attacker's capabilities (which, of course, is perfectly safe)
+
+class rv_dm_buffered_enable_vseq extends rv_dm_base_vseq;
+  `uvm_object_utils(rv_dm_buffered_enable_vseq)
+  `uvm_object_new
+
+  // Override the constraint in rv_dm_base_vseq that enables debug (to disable it instead)
+  constraint debug_enabled_c {
+    lc_hw_debug_en == 1'b0;
+  }
+
+  // (Shortened versions of) the names of the buffered copies of the debug enable signal.
+  typedef enum { Fetch, Rom, Sba, DebugReq, ResetReq } copy_t;
+
+  // The index of the buffered copy of the signal to force.
+  rand copy_t tgt_copy;
+
+  function string enable_path_at(copy_t copy);
+    return $sformatf("tb.dut.u_lc_en_sync_copies.gen_buffs[%0d].lc_en_out", copy);
+  endfunction
+
+  task force_enable_at(copy_t copy);
+    `DV_CHECK(uvm_hdl_force(enable_path_at(copy), lc_ctrl_pkg::On))
+
+    // Wait a few cycles to allow any newly enabled copies to propagate through and for the LC gate
+    // FSMs to get back to Active.
+    cfg.clk_rst_vif.wait_clks(4);
+  endtask
+
+  task release_enable_at(copy_t copy);
+    `DV_CHECK(uvm_hdl_release(enable_path_at(copy)))
+
+    // Wait a few cycles to allow the (now cleared) enable flag for the given copy to propagate
+    // through (making sure that an operation that starts after this task is complete will be
+    // blocked)
+    cfg.clk_rst_vif.wait_clks(4);
+  endtask
+
+  // Try to access the functionality gated by the requested copy of the enable signal. Assert that
+  // the feature is accessible iff copy equals tgt_copy. Checking that a feature *is* enabled on a
+  // match is a way to check there is no bug in the test, so we would notice if we would see
+  // cross-enable behaviour if the two copies did not match.
+  task check_copy(copy_t copy);
+    case (copy)
+      Fetch: check_fetch();
+      Rom:   check_rom();
+      Sba:   check_sba();
+      DebugReq: check_debug_req();
+      ResetReq: check_reset_req();
+      default: `uvm_fatal(`gfn, "Unknown copy index")
+    endcase
+  endtask
+
+  // Check whether it is possible to do TL fetches from the debug module's ROM. This should only be
+  // possible if tgt_copy == Fetch
+  task check_fetch();
+    uvm_reg_addr_t mem_base_addr = tl_mem_ral.default_map.get_base_addr();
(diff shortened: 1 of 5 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_buffered_enable_vseq.sv
```diff
@@ -0,0 +1,283 @@
...
+    $asserton(0, tb.dut.enable_checker.DebugRequestNeedsDebug_A);
+  endtask
+
+  // Check whether it is possible to send an ndmreset request from the debug module to the rest of
+  // the system.
+  //
+  // As with check_debug_req, we're slightly cheating and force an output from dm_top and check that
+  // it doesn't make it to the top-level pin.
+  task check_reset_req();
+    // Note that the structureof the design means that an ndmreset request will only be exposed at
```
vogelpi, line 186: Typo: structureof -> structure of

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_buffered_enable_vseq.sv
```diff
@@ -0,0 +1,283 @@
...
+
+  // Check whether it is possible for an SBA access to cause the sba_tl_h_o signal to have
+  // a_valid=1. If the Sba copy of the enable flag is not On, this should never happen. If it is
+  // forced to On, it can happen. As a test of the check, we actually expect the access to complete
+  // in this situation (reading a register).
+  task check_sba();
+    sba_access_item item = sba_access_item::type_id::create("sba_req");
+    bit tl_finished, saw_a_valid;
+
+    // Construct a single arbitrary system bus access, reading from a known-good address in the
```
vogelpi, line 132: Interesting idea to use SBA to read a register of RV_DM itself :-)

vogelpi (approved):
Thanks for the PR and the useful explanations @rswarbrick ! This makes sense to me. I've just found some minor typos.

hw/ip/rv_dm/dv/tb.sv
```diff
@@ -130,4 +130,48 @@ module tb;
...
+  // saved_en_values. Together with the contents of force_enable, this feeds into the
+  // local_en_values signal. We then force the values in the design to match that signal until
+  // use_force drops again, at which point we release the force.
+
+  // Enable the forcing of the signals in the design, as described above. Set this from a vseq with
+  // uvm_hdl_deposit.
+  bit use_force;
+
+  // A snapshot of the values in lc_hw_debug_en_gated on the last posedge of use_force.
+  lc_ctrl_pkg::lc_tx_t [4:0] saved_en_values;
```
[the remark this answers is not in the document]
rswarbrick: Good point. I don't *think* there's a way to use the parameter with a hierarchical reference there: there's no equivalent using the module as a "namespace" like `rv_dm::NumEnableCopies` in SystemVerilog and you can't use something like `dut.NumEnableCopies` here because it expects a constant expression.

But you can write down the number in a localparam and then add an assertion that it's right. I'll do that :-)

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_buffered_enable_vseq.sv
rswarbrick: :-D And it's guaranteed to be around to reply!

rswarbrick:
Ah nuts. It looks like this syntax is not allowed by VCS, which is what we use for CI. I'll try to come up with a better solution tomorrow.

rswarbrick:
Hmm, this has been surprisingly difficult. It turns out that VCS *really* didn't like my attempt to read or force the debug enable copies and I think the root of the problem was that the number of copies (5) came from an enum value which isn't visible outside of `rv_dm.sv`.

This version seems to compile sensibly. I'll do a little more local testing to make sure everything is correct, but (obviously) we're touching the RTL. I don't *think* there's any way to achieve this test with VCS without such a change. I'm not sure what the right approach is at this point.

rswarbrick:
Huh. My idea for "an easier way" doesn't actually work because you then need to do a hierarchical reference from the vseq, which is not allowed (because it's starting from a package rather than some instance). I'm reasonably convinced that this is the cleanest approach I can do, so marking it ready for review.

hw/ip/rv_dm/dv/tb.sv
```diff
@@ -130,4 +130,56 @@ module tb;
...
+      force dut.lc_hw_debug_en_gated[1] = lc_ctrl_pkg::lc_tx_t'(local_en_values[1]);
+      force dut.lc_hw_debug_en_gated[2] = lc_ctrl_pkg::lc_tx_t'(local_en_values[2]);
+      force dut.lc_hw_debug_en_gated[3] = lc_ctrl_pkg::lc_tx_t'(local_en_values[3]);
+      force dut.lc_hw_debug_en_gated[4] = lc_ctrl_pkg::lc_tx_t'(local_en_values[4]);
+      @(negedge use_force);
+      release dut.lc_hw_debug_en_gated[0];
+      release dut.lc_hw_debug_en_gated[1];
+      release dut.lc_hw_debug_en_gated[2];
+      release dut.lc_hw_debug_en_gated[3];
+      release dut.lc_hw_debug_en_gated[4];
```
vogelpi, line 181: Can this maybe be simplified using the approach taken for AES? For reference, I recommend taking a look at these files:
- The sequence: https://github.com/lowRISC/opentitan/blob/master/hw/ip/aes/dv/env/seq_lib/aes_core_fi_vseq.sv
- The interface applying the force: https://github.com/lowRISC/opentitan/blob/master/hw/ip/aes/dv/err_injection_if/fi_core_if.sv

I don't have a strong opinion at all. Just sharing this as a potential source of inspiration for you :-)

hw/ip/rv_dm/rtl/rv_dm_pkg.sv
```diff
@@ -9,4 +9,6 @@ package rv_dm_pkg;
 
   parameter next_dm_addr_t NEXT_DM_ADDR_DEFAULT = '0;
 
+  localparam int unsigned NumDebugEnCopies = 5;
```
vogelpi, line 12: Would it be possible to defer this RTL change? Because this will change the output of synthesis and thus I strongly believe we can't port it into the release branch.

Would it be possible to define a constant in a DV-only package file and create an issue to unify the constants usage, i.e., to basically defer the RTL change to a separate PR we don't move over to the release branch?

hw/ip/rv_dm/dv/tb.sv
rswarbrick: Ah! That would work if the signals we were forcing weren't enums. Unfortunately, the fact that they are enums means that `uvm_hdl_force`/`uvm_hdl_release` (which the AES example uses) won't work. Xcelium fails at runtime when releasing the force, for some reason.

Doing the force/release in a loop without going through VPI doesn't work because you end up having something that looks like `force foo[i] = 123;`, which isn't allowed (because the index of the value to force needs to be known at compile time).

Trying to force the entire array doesn't work on VCS because it silently just forces the first item in the array. I suspect that's a tool bug.

Getting to the bottom of this was a VERY frustrating afternoon.

rswarbrick:
I've just force-pushed with a (much) simpler version. Thanks, @vogelpi and @andreaskurth for suggesting the non-enum signal that I could force in the middle. Much easier and doesn't touch the RTL.

vogelpi (approved):
This looks great, thanks @rswarbrick for going the extra mile to avoid touching the RTL!

rswarbrick:
Merging. I've checked and the CI failures are unrelated to this change:
- "Private CI / Run top_earlgrey DV test regression 1" seems to have failed because of an IT issue (failing to open a deleted zip file from Bazel)
- "CI (CW340's Earl Grey Bitstream)" failed to download OpenOCD.

PR #24161, opened 2024-07-29
[spi_host, doc] control.sw_rst clarification

antmarzam:


rswarbrick (commented):
This looks reasonable, but you'll actually have to make the change in `spi_host.hjson` and then update this file with `make -C hw`.

antmarzam:
> This looks reasonable, but you'll actually have to make the change in `spi_host.hjson` and then update this file with `make -C hw`.
@rswarbrick 
I did what you suggested, but that wasn't generating a `registers.md` file. 
Do I need to add more flags to the make command?

Will drop the first commit once this is ready to merge.

rswarbrick:
I think you've done it! :-) You can see the diff by hitting this button:
image

Otherwise, I'm very happy with this. Will ping a second committer to allow the change (because it touches a file where some changes should be blocked)

rswarbrick (approved):

rswarbrick:

hcallahan-lowrisc:

Issue #18797, opened 2023-06-02
[entropy_src/dv] Health tests sometimes fail in scoreboard but not in DUT


andreaskurth:
In very rare cases (based on how rarely this occurs even in full regression runs), `entropy_src`'s scoreboard predicts a recoverable alert due to a failed health tests but RTL does not produce this alert. This usually manifests itself as `UVM_ERROR @ * ps: (cip_base_scoreboard.sv:*) uvm_test_top.env.scoreboard [uvm_test_top.env.scoreboard] alert recov_alert did not trigger max_delay:*`.

This seems to affect bucket tests and Markov tests.

Given current evidence, I think this is a minor difference between scoreboard and DUT (e.g., of statistical nature) and does not pose a substantial problem. Nonetheless, we should resolve it for DV closure. I'm thus labeling it V3.

<details><summary>Here's a patch that can help with debugging:</summary>

```patch
diff --git a/hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv b/hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv
index 150130dca4..1419793a00 100644
--- a/hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv
+++ b/hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv
@@ -574,6 +574,7 @@ class entropy_src_scoreboard extends cip_base_scoreboard#(
                                               !total_scope, sigma_lo, fail_lo);
     end
 
+    `uvm_info(`gfn, $sformatf("evaluate_adaptp_test()=%b", fail_hi || fail_lo), UVM_LOW)
     return (fail_hi || fail_lo);
   endfunction
 
@@ -604,6 +605,7 @@ class entropy_src_scoreboard extends cip_base_scoreboard#(
                                               1'b0, sigma, fail);
     end
 
+    `uvm_info(`gfn, $sformatf("evaluate_bucket_test()=%b", fail), UVM_LOW)
     return fail;
   endfunction
 
@@ -636,9 +638,11 @@ class entropy_src_scoreboard extends cip_base_scoreboard#(
     update_watermark("markov_hi", fips_mode, total_scope ? value : maxval);
 
     fail_lo = check_threshold("markov_lo", fips_mode, total_scope ? value : minval);
+    `uvm_info(`gfn, $sformatf("fail_lo=%b", fail_lo), UVM_LOW)
     if (fail_lo) predict_failure_logs("markov_lo");
 
[...]

PR #24166, opened 2024-07-30
[entropy_src/dv] Ignore outstanding alerts after disablement

h-filali:
Some tests were failing because the scoreboard was expecting an alert after disabling the DUT. This PR adds a variable that when set leads the scoreboard to ignore any outstanding alerts.

Closes [#18797](https://github.com/lowRISC/opentitan/issues/18797)

### DIFF
--- a/hw/dv/sv/cip_lib/cip_base_scoreboard.sv
+++ b/hw/dv/sv/cip_lib/cip_base_scoreboard.sv
@@ -40,6 +40,7 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
 
   // This event is used to notify process_set_exp_alerts there are alerts to be handled.
   event new_exp_alert;
+  bit ignore_exp_alert = 0;
 
   // alert checking related parameters
   bit do_alert_check = 1;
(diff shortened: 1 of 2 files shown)


hw/dv/sv/cip_lib/cip_base_scoreboard.sv
```diff
@@ -297,7 +298,7 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
       cfg.clk_rst_vif.wait_n_clks(1);
       if (under_alert_handshake[alert_name] || cfg.under_reset) return;
     end
-    if (!cfg.en_scb) return;
+    if (!cfg.en_scb || ignore_exp_alert) return;
```
vogelpi, line 301: The intent for this change is pretty clear (see https://github.com/lowRISC/opentitan/issues/18797#issue-1737918919) but I fear this opens up a window for introducing bugs related to signaling fatal alerts - we should never be able to turn them off.

Thus, I suggest changing the PR to only take this bit into account for non-fatal alerts. Whether an alert is fatal can be checked using the `is_fatal` member variable. Something like this could work (to be tested):
```sv
if (!cfg.en_scb || (ignore_exp_alert && !expected_alert[alert_name].is_fatal)) return; 
```

vogelpi (approved):
Thanks for the PR @h-filali , I suggest only allowing to ignore expected non-fatal alerts. The rest looks good to me, thanks!

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
h-filali: What if we also want to ignore recoverable alerts? For example when a recoverable alert is to be expected and the DUT is turned off right before it is triggered. Can't we just say that `ignore_exp_alert` needs to be set only if `!expected_alert[alert_name].is_fatal` is true in the appropriate cases?

vogelpi: We've discussed this in person. There was a misunderstanding regarding the `!` in front of `expected_alert[alert_name].is_fatal`.

h-filali: We discussed offline and we agreed to add the additional is_fatal check, since fatal alerts should trigger even if the DUT is disabled.

h-filali:
@vogelpi thanks for your feedback. I changed the line to what you suggested and ran 100% of alert tests as well as 10% of a full regression and I didn't notice any unexpected failures. I think this can be merged now.

PR #24168, opened 2024-07-30
[quality,sw/otbn] Add SHA-3 and SHAKE OTBN SW implementations

dop-amin:
This PR adds software implementations of SHA-3 and SHAKE for OTBN. 
We provide a "static" test in the `.exp` file as well as auto-generated test infrastructure as implemented in #24054.

Right now, the `.exp`-tests for SHA-3 and SHAKE are inside the same file (`sw/otbn/crypto/tests/sha3-shake_test.s`). Do you prefer this to be split up into individual tests (maybe even per parameter set)?

Joint work with Pham Hoang Nguyen Hien (@phamhnh).


sw/otbn/crypto/sha3-shake.s
```diff
@@ -0,0 +1,753 @@
...
+  /* Copy context pointer */
+  add     x5, x0, x10 
+
+  /* Load necessary constants:
+     w11 <= (1 <<  64) - 1
+     w12 <= (1 << 256) - 1
+     w13 <= mask_top_1 */
+  li      x29, 12
+  la      x31, mask_bn 
+  bn.lid  x29, 0(x31++)
```
jadephilipoom, line 336: nit: If you just need an all-ones value, you can get it faster and without needing to store a constant using the all-zero register:
```suggestion
  bn.not  x29, w31 
```

sw/otbn/crypto/tests/generated/BUILD
```diff
@@ -33,3 +34,55 @@ py_binary(
...
+    otbn_autogen_sim_test(
+        name = "sha3_test" + str(i),
+        srcs = ["sha3_py_test.s"],
+        seed = i,
+        testgen = ":sha3_testgen",
+        deps = [
+            "//sw/otbn/crypto:sha3-shake",
+        ],
+    )
+    for i in range(100)
```
jadephilipoom, line 61: nit: let's keep it to `10`, since this will run in CI for every PR in the future.
```suggestion
    for i in range(10)
```

sw/otbn/crypto/tests/generated/BUILD
```diff
@@ -33,3 +34,55 @@ py_binary(
...
+    otbn_autogen_sim_test(
+        name = "shake_test" + str(i),
+        srcs = ["shake_py_test.s"],
+        seed = i,
+        testgen = ":shake_testgen",
+        deps = [
+            "//sw/otbn/crypto:sha3-shake",
+        ],
+    )
+    for i in range(100)
```
jadephilipoom, line 87:
```suggestion
    for i in range(10)
```

sw/otbn/crypto/tests/generated/sha3_testgen.py
```diff
@@ -0,0 +1,63 @@
...
+
+import argparse
+import random
+from typing import TextIO, Optional
+
+from Crypto.Hash import SHA3_224, SHA3_256, SHA3_384, SHA3_512
+
+from shared.testgen import write_test_data, write_test_exp
+
+DST_LEN = 64 # 28, 32, 48, 64
```
jadephilipoom, line 14: Should this be a random choice rather than a fixed value, so that variants other than SHA3-512 get tested?

sw/otbn/crypto/tests/generated/shake_testgen.py
```diff
@@ -0,0 +1,63 @@
...
+
+import argparse
+import random
+from typing import TextIO, Optional
+
+from Crypto.Hash import SHAKE128, SHAKE256
+
+from shared.testgen import write_test_data, write_test_exp
+
+DST_LEN = 32 # SHAKE128: 16, SHAKE256: 32
```
jadephilipoom, line 14: Similar comment as for SHA3, should this be randomly selected instead of fixed so that SHAKE128 also gets tested?

quality/BUILD.bazel
```diff
@@ -62,6 +63,8 @@ licence_test(
         "sw/otbn/crypto/rsa_verify.s",
         "sw/otbn/crypto/handwritten/rsa_verify_3072.s",
         "sw/otbn/crypto/sha512.s",
+        # Code based on tiny_sha3
+        "sw/otbn/crypto/symmetric.s",
```
jadephilipoom, line 67: Looks like the license checker is failing in CI because this file name has changed, so the exception no longer triggers:
```suggestion
        "sw/otbn/crypto/sha3-shake.s",
```

sw/otbn/crypto/sha3-shake.s
```diff
@@ -0,0 +1,753 @@
...
+mask_bn:
+  .dword 0xffffffffffffffff
+  .dword 0xffffffffffffffff
+  .dword 0xffffffffffffffff
+  .dword 0xffffffffffffffff
+mask_top_1:
+  .dword 0x8000000000000000
+  .dword 0x8000000000000000
+  .dword 0x8000000000000000
+  .dword 0x8000000000000000
```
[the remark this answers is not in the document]
phamhnh: Hi,
We ran `util/fix_trailing_whitespace.py` and this is fixed automatically.

sw/otbn/crypto/tests/generated/sha3_testgen.py
phamhnh: I initially thought that all 10 tests are run for each variant of SHA-3 to test its correctness, and thus the choice for the variant can be set manually. If I understood correctly, in case `DST_LEN` is set to be random, maybe only 2 tests for SHA3-512 are ran, while 4 tests for SHA3-224 are run, and maybe 0 test for SHA3-256, etc. Is this enough for each variant? Which test scenario that you want to have?

sw/otbn/crypto/sha3-shake.s
```diff
@@ -0,0 +1,753 @@
...
+    /* Get destination address */
+    add  x6, x10, x5 
+    addi x5, x5, 1 
+    /* Align destination address and load data */
+    and  x7, x6, x30
+    lw   x28, 0(x7)
+    /* Mask out the lowest 2 bits of destination address
+       to compute position of st[j] */
+    andi x29, x6, 0x3
+    slli x29, x29, 3
```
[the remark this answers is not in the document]
dop-amin: Thanks for the comment! Yes, we considered that but the code gets quite a bit longer and more complicated because the alignment needs to be ensured upon starting to read/write (so we need to work our way from byte-wise -> 4B-wise -> 32B-wise) and then when finishing, we need to do the reverse to deal with remainders smaller than 32B or 4B. 
I'll give the implementation a try and see how much code it will be later this week.

sw/otbn/crypto/tests/generated/sha3_testgen.py
jadephilipoom: Every time a PR is submitted to OpenTitan, the test will run via Bazel automatically as part of CI, and the test vectors will be generated by this script (i.e. only for SHA3-512). It would be nice if these automatic tests captured multiple variants, since they will be run much more often than anyone actually runs the tests manually. For that reason, I'd rather do it automatically instead of via arguments. I'm okay with either selecting it randomly or simply iterating through the variants as it generates the 10 tests. It's not strictly necessary to cover every variant every time; the manual `.exp` test already checks all the variants in turn, so in this case I think the random test is just an extra layer of checks to catch things the manual tests may have missed. :slightly_smiling_face:

PR #24171, opened 2024-07-30
[otbn,dv] Don't expect specific pre-decode checks in ctrl_redun_vseq

rswarbrick:


vogelpi (approved):
LGTM, thanks @rswarbrick !

vogelpi:
Upon request, I've tested the PR but it seems to only lead to a very minor improvement in the test pass rate from 97% to 98%. So we probably shouldn't merge this as is just yet.

rswarbrick:
Hi there! I've just done a bit of debugging at this end to track down the failures that you saw and have found a minor timing problem in the DV code, which is fixed by #24250. That's unrelated to the fix here.

To check, I just started at this version and cherry-picked #24250 and I'm pleased to say that I saw 100 seeds pass. (phew). I'll merge this PR now.

PR #24172, opened 2024-07-30
[sw/pentest] Update SCA library

nasahlpa:
This PR consists of several commits that update and use the new SCA lib that is utilized by the penetration testing framework. This PR moves code from my fork used in the testing to the upstream project.

New functionalities include:
- Returning the device ID. This allows the tester to differentiate between different chips/fpga used in the setup
- Configure and return the triggered alerts for FI tests

Moreover, this PR consists of the following cosmetical changes:
- Move function description from source file into header file to comply with the coding standard
- Remove 0 from OK_STATUS(0) as this is unnecessary
- Rename cryptotest to penetrationtest


vogelpi (approved):
This looks all good to me. Thanks @nasahlpa for splitting up the work into individual commits so cleanly! This made the review very easy.

PR #24179, opened 2024-07-31
[dv] Add plusarg to enable to skip external pull-up on POR pad

sha-ron:
Add an option to skip the external pull up from test bench on POR pad.
This following https://github.com/lowRISC/opentitan/pull/24124 that adds and internal pull up.


eshapira (approved):
LGTM

PR #24180, opened 2024-07-31
[DV/SPI] updated sleep retention vseq to avoid race condition

thouri:
Changed software log printed string in DV_WAIT from "SYNC: Awaked" to "PASS!"
If this DV_WAIT happens after the desired "SYNC: Awaked" is printed the test will fail on timeout.
"PASS!" is the last printed line to the software C log, I choose this line because the desired "SYNC: Awaked" happens almost at the end of the C code.


eshapira (approved):
LGTM

hcallahan-lowrisc (approved):

PR #24123, opened 2024-07-25
[entropy_src/dv] Randomize the simulation duration

h-filali:
This PR randomizes the duration of the simulation and sets the default for the mean time between failures to -1 to turn them off. The time between failures is already randomized for the rng_test and the fw_ov test.

Resolves [#18836](https://github.com/lowRISC/opentitan/issues/18836)

The values below the TODO comment are already random for certain tests.
@andreaskurth please tell me if I am misunderstanding something here.

PR #24181, opened 2024-07-31
[entropy_src/dv] Fix timeout failure for stress all test

h-filali:
This PR fixes an issue that arises for the stress all test whenever a ht failure occurs during the smoke vseq. If a HT failure happens the sequence infinitely waits for the DUT to produce a seed even though it will never be produced.
Now the vseq doesn't expect a seed if we get a HT failure.


vogelpi (approved):
This PR fixes some expected fallout from #24123 and it looks good to me. Thanks Hakim.

PR #24183, opened 2024-07-31
[dv,usbdev] Initial minor changes to improve coverage results

alees24:
This PR just implements a few initial minor changes to improve the coverage results:

- Disabling the tx oscillator test mode was not being performed at sequence end, so that state transition was never observed.
- Low Speed traffic would have been collected as per Full Speed traffic, when in reality all Low Speed packets have a PREamble PID and shall be ignored entirely by the device. This is just because the usb20_monitor is actually capable of decoding Low Speed traffic.


hcallahan-lowrisc (approved):
LGTM, thanks Adrian

PR #24185, opened 2024-07-31
[dv,usbdev] Endpoint types sequence

alees24:
This PR modifies the sequence `usbdev_spray_packets` to support Isochronous transfer types and the `set_nak_out` functionality, and then derives a simple, largely-unconstrained sequence `usbdev_endpoint_types` to ensure that all permutations of the configuration bits may be observed in coverage, and each endpoint configuration may see each of the possible token packets.
- Add support for sending PRE token packets (Low Speed traffic).
- Ensure that Isochronous configuration bits are appropriately qualified in the sequence and in the functional model; it is possible to 'misconfigure' an endpoint for _both_ Isochronous traffic and Control Transfers (SETUP packets), so ensure that the behavior is consistent throughout.

~~**This PR is built on 24183 and only the final commit is relevant to this PR; awaiting CI run before merging that PR.**~~

### DIFF
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -633,6 +633,20 @@
       tests: ["usbdev_disable_endpoint"]
     }
     {
+      name: endpoint_types
+      desc: '''
+            Exercise all permutations of endpoint configuration bits with all possible token packets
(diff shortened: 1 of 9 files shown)


andreaskurth (approved):
This looks sensible to me. (I'm not an expert on USB but you are, and your code changes and comments make sense to me.)

alees24:
CI failure unrelated.

PR #24187, opened 2024-07-31
[dv,usbdev] Support Iso transfers in max_usb_traffic

alees24:
Generalize the `usbdev_max_usb_traffic` sequence to support Isochronous transfers/endpoints too.
Introduce a derived sequence that constrains the testing to the much more common non-Isochronous traffic (Iso transfers are rarely used on the USB in practice).

_This PR requires the changes to `usbdev_bfm.sv` in #24185_ so that must be merged first otherwise there will be some test failures when Isochronous transfers are selected oweing to a discrepancy between the model and the DUT.

### DIFF
--- a/hw/ip/usbdev/data/usbdev_testplan.hjson
+++ b/hw/ip/usbdev/data/usbdev_testplan.hjson
@@ -1074,7 +1074,10 @@
             interrupts.
             '''
       stage: V2
-      tests: ["usbdev_max_usb_traffic"]
+      tests: [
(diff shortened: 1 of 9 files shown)


andreaskurth (approved):
This looks sensible to me. (I'm not an expert on USB but you are, and your code changes and comments make sense to me.)

alees24:
CI failures irrelevant to these changes. DV smoke test passed.

PR #24192, opened 2024-07-31
Move test token cshake hash calculations from device to host

### LINKED ISSUE #23895: [manuf] Enhance test tokens CP provisioning by sending its hash
### Description

Currently, the test unlock / exit tokens provisioned during CP phase are sent in cleartext to the device, and the device stores their hashes.

https://github.com/lowRISC/opentitan/blob/09b4736fce31af2e520b07fc4b1c5c66f5f17cf8/sw/device/silicon_creator/manuf/lib/individualize.c#L172-L177

 The flow can be improved by hashing the tokens on the host side, and only transferring their hashes to the device instead. With this approach, we can prevent passive eavesdropping on the token provisioning by the hash’s preimage resistance.

vbendeb:
This patch adds test token cShake hash calculations on the host side to make sure that the tokens are not passed in the clear.

fixes #23895


vbendeb:
never mind the update, just trying to see what else breaks in CI

Issue #24184, opened 2024-07-31
[verilator, tests] Timed-out Verilator test targets in CI


AlexJones0:
### Description

All Verilator test targets have been run in public CI in preparation for the addition of a weekly/nightly Verilator job. Of the current 390 targets, 67 tests timed out when run using `--test_timeout=2400,2400,3600,-1` (compared to the standard `60,300,900,3600`). All tests that timed out hence took at least 40 minutes, and so should probably not be run in a CI environment on Verilator. See related issue #24182 for Verilator tests that failed.

A separate PR will mark all of these Verilator test targets as `manual` (to avoid running in CI), but we need to determine for each test whether they genuinely ran out of time (i.e. allocating more time will allow the test to succeed) and as such should stay marked `manual`, or whether they timed out due to an error, and as such should be marked `broken` and/or fixed. The aim here is that any Verilator test target that is not marked as broken should be able to be manually run by overriding the test timeout, even if it might take hours to complete.

To run any of these tests, after setting up Verilator for Opentitan, run

```
./bazelisk.sh test -t- --test_output=streamed <TARGET-LOCATION>
```

The `--test_timeout=` Bazel flag can be used to manually override the timeout for these tests.

The 67 timed-out Verilator test targets that need checking are:

- [x] `sw/device/silicon_creator/lib:otbn_boot_services_functest_sim_verilator`
- [x] `sw/device/silicon_creator/lib/sigverify:mod_exp_ibex_functest_hardcoded_sim_verilator`
- [x] `sw/device/silicon_creator/lib/sigverify:mod_exp_ibex_functest_wycheproof_sim_verilator`
- [x] `sw/device/silicon_creator/lib/sigverify:rsa_verify_functest_sim_verilator`
- [x] `sw/device/silicon_creator/lib/sigverify:sigverify_dynamic_functest_hardcoded_sim_verilator`
- [x] `sw/device/silicon_creator/lib/sigverify:sigverify_dynamic_functest_wycheproof_sim_verilator`
- [x] `sw/device/silicon_creator/rom/e2e/weak_straps:sw_strap_value_sim_verilator`
- [x] `sw/device/tests:alert_handler_lpg_sleep_mode_pings_test_sim_verilator`
- [ ] `sw/device/tests:alert_handler_reverse_ping_in_deep_sleep_test_sim_verilator`
- [ ] `sw/device/tests:aon_timer_sleep_wdog_sleep_pause_test_sim_verilator`
[...]

Issue #24182, opened 2024-07-31
[verilator, tests] Failing Verilator test targets


AlexJones0:
### Description

All Verilator test targets have been run in public CI in preparation for the addition of a weekly/nightly Verilator job. Of the current 390 targets, 14 tests explicitly failed within their allocated timeout (`--test_timeout=2400,2400,3600,-1`). These failures have also been reproduced when executing the tests locally. Tracking the state of tests that timed out will be covered in a separate issue.

To reproduce any of these failures, after setting up Verilator for Opentitan, run 

```sh
./bazelisk.sh test -t- --test_output=streamed //sw/device/tests:<TARGET-NAME>
```

Of these 14 tests, the below 8 are failing for identifiable reasons, which will be addressed in a separate PR:

- `pwrmgr_all_reset_reqs_test_sim_verilator`: 
- `rv_core_ibex_mem_test_prod_sim_verilator`
- `usbdev_aon_wake_reset_test_sim_verilator`
- `spi_host_winbond_flash_test_sim_verilator`
- `spi_host_irq_test_sim_verilator`
- `spi_host_smoketest_sim_verilator` 
- `gpio_smoketest_sim_verilator`
- `example_mem_ujcmd_test_sim_verilator`

The remaining 6 tests are failing for reasons that are not clear to me. We should look into why these tests are failing specifically when run in the Verilator execution environment, and either aim to fix these tests or mark them as broken / not intended to be run in Verilator accordingly.

- [ ] `usbdev_setuprx_test_sim_verilator`
  - Error: `usbdev_setuprx_test.c:208] CHECK-STATUS-fail: DeadlineExceeded:["USB",111]`
- [ ] `usbdev_config_host_test_sim_verilator`
  - Error: `usbdev_config_host_test.c:98] CHECK-fail: success`
- [ ] `usbdev_logging_test_sim_verilator`
  - Error: ```Error: Vchip_sim_tb: ../src/lowrisc_dv_dpi_c_usbdpi_0.1/usbdpi.c:694: void getTestConfig(usbdpi_ctx_t*, uint16_t): 
    Assertion !"Device is stalled"' failed.
    [2024-07-29T23:36:53Z INFO  opentitantool] Command result: Broken pipe```
- [ ] `pwrmgr_usb_clk_disabled_when_active_test_sim_verilator`
  - Error: `pwrmgr_usb_clk_disabled_when_active_test.c:112] CHECK-fail: Unexpected hung address`
[...]

PR #24197, opened 2024-08-01
[verilator, tests] Fix and tag failing & timed-out Verilator test targets

AlexJones0:
This PR partially addresses issues #24182 and #24184.

All Verilator test targets were run using public CI overnight in preparation for the addition of a weekly/nightly Verilator job in PR #24145. Of the 390 targets (at the time of testing), 14 explicitly failed and 67 timed out (when using `--test_timeout=2400,2400,3600,-1`). This PR accordingly:

- Marks all 67 timed-out tests as "manual", as they take at least 40 (and many > 60) minutes to complete when run on our public CI infrastructure and as such are too long to run in any sort of regular CI. The confirmation of whether these tests truly take a long time or should be marked broken and/or fixed is tracked by #24184. All of these tags are documented with the reason for their inclusion, mentioning whether the timeout used was 40 or 60 minutes for future reference.
- Marks all 6 failing Verilator tests tracked in #24182 as `broken`, until they can be looked at further.
- Fixes the `example_mem_ujcmd_test` test for Verilator, which was incorrectly passing in the Firmware ELF as a keyword argument instead of a positional argument. This has been verified (`PASSED in 599.6s`) by running 
```
./bazelisk.sh test -t- --test_output=streamed --test_timeout=3600,3600,3600,3600 \
    //sw/device/tests:example_mem_ujcmd_test_sim_verilator
```
- Fixes the `spi_host_irq_test` target, which was failing on Verilator due to the SPI clock frequency being configured to 1 MHz, whereas Verilator's HiSpeed peripheral frequency is set to just 500 KHz. A conditional check is added and the SPI clock speed is reduced to 100 KHz if running on Verilator. This has been verified (`PASSED in 133.4s`) by running 
```
./bazelisk.sh test -t- --test_output=streamed --test_timeout=3600,3600,3600,3600 \
    //sw/device/tests:spi_host_irq_test_sim_verilator
```
- Correctly excludes the Verilator execution environment from `gpio_smoketest`, which previously erroneously had the `//hw/top_earlgrey:verilator` environment excluded instead of the correct `//hw/top_earlgrey:sim_verilator`. This occurrence is fixed in 3 other places in the `BUILD` file.
- Removes both the `sim_dv` and `sim_verilator` execution environments from the `usbdev_aon_wake_reset_test` Bazel test target, as [the test source already immediately fails if the device type is DV or Verilator](https://github.com/lowRISC/opentitan/blob/70142638925f66eae20f6232b5a7c1ffc51ff0e3/sw/device/tests/usbdev_aon_wake_reset_test.c#L120) - this lack of support for simulator environments is now made explicit in the Bazel parameters.
- Marks the `spi_host_winbond_flash_test` as being `manual` for the Verilator environment, as it requires access to Winbond flash, which is not currently available in CI.
- Removes the Verilator execution environment from `rv_core_ibex_mem_test_prod`, because [the `rv_core_ibex` tests are marked as being incompatible with test_rom](https://github.com/lowRISC/opentitan/blob/70142638925f66eae20f6232b5a7c1ffc51ff0e3/sw/device/tests/BUILD#L5756-L5757), which Verilator uses, and as such this environement should not be supported.
- Marks the `pwrmgr_all_reset_reqs_test` Verilator execution environment as broken alongside the existing broken `fpga_cw310_sival_rom_ext` environment, as the test provides the same error in both cases.
- Removes the Verilator execution environment from `spi_host_smoketest`, as that test requires a flash connected to the SPI bus which is not emulated by Verilator.

Aside from the few fixes, this PR only marks Verilator tests as manual or broken (or removes the Verilator execution environment). Existing Verilator tests that are running in CI are unaffected by this, as currently only `aes_smoketest` is being run on [English breakfast](https://github.com/lowRISC/opentitan/blob/70142638925f66eae20f6232b5a7c1ffc51ff0e3/ci/scripts/run-english-breakfast-verilator-tests.sh#L37), and [16 specific quick tests are being run on top_earlgrey](https://github.com/lowRISC/opentitan/blob/70142638925f66eae20f6232b5a7c1ffc51ff0e3/ci/scripts/run-verilator-tests.sh#L19C2-L34C64).

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -1421,7 +1421,7 @@ opentitan_test(
         timeout = "eternal",
         tags = ["manual"],
         test_cmd = " ".join([
-            "--firmware-elf=\"{firmware:elf}\"",
+            "\"{firmware:elf}\"",
         ]),
         test_harness = "//sw/host/tests/chip/mem",
     ),
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -2611,6 +2611,9 @@ opentitan_test(
         """,
         test_harness = "//sw/host/tests/chip/pwrmgr:sleep_all_resets",
     ),
+    # FIXME this test is also broken in Verilator for the same reason as in
+    # sival ROM_EXT above. Remove this line when fixed. See #21706.
+    verilator = verilator_params(tags = ["broken"]),
     deps = [
         "//hw/top_earlgrey/sw/autogen:top_earlgrey",
         "//sw/device/lib/arch:device",
@@ -3469,7 +3472,12 @@ opentitan_test(
     name = "spi_host_winbond_flash_test",
     srcs = ["spi_host_winbond_flash_test.c"],
     exec_env = dicts.add(
-        EARLGREY_TEST_ENVS,
+        dicts.omit(
+            EARLGREY_TEST_ENVS,
+            # This test requires an SPI flash on the bus, and therefore
+            # cannot be tested in Verilator where this is not available.
+            ["//hw/top_earlgrey:sim_verilator"],
+        ),
         EARLGREY_SILICON_OWNER_ROM_EXT_ENVS,
     ),
     deps = [
@@ -5990,7 +5998,10 @@ opentitan_test(
     name = "rv_core_ibex_mem_test_prod",
     srcs = ["rv_core_ibex_mem_test.c"],
     exec_env = dicts.add(
-        EARLGREY_TEST_ENVS,
+        dicts.omit(
+            EARLGREY_TEST_ENVS,
+            ["//hw/top_earlgrey:sim_verilator"],
+        ),
         EARLGREY_SILICON_OWNER_ROM_EXT_ENVS,
     ),
     deps = [
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -679,6 +679,8 @@ opentitan_test(
             "//hw/top_earlgrey:silicon_creator": None,
         },
     ),
+    # Broken on Verilator. See #24182.
+    verilator = verilator_params(tags = ["broken"]),
     deps = [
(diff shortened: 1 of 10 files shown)


jwnrt (approved):
Thanks, this LGTM if it passes CI. Thanks also for commenting the reasons why they're broken / manual.

AlexJones0:
CI failed because I didn't autogenerate the `plic` tests with the added manual verilator tag and comments. I've just pushed that change so hopefully CI should pass now with that

AlexJones0:
CI failed because I didn't autogenerate the `plic` tests with the added manual verilator tag and comments, and because of two instances of failed Bazel BUILD linting. I've just pushed those changes so hopefully CI should pass now with that.

sw/device/tests/BUILD
```diff
@@ -3439,6 +3442,7 @@ opentitan_test(
         EARLGREY_TEST_ENVS,
         EARLGREY_SILICON_OWNER_ROM_EXT_ENVS,
     ),
+    verilator = verilator_params(tags = ["manual"]),
```
engdoreis, line 3445: Why manual? Isn't this test that needs a spi flash on the bus and therefore can't be tests in verilator?

engdoreis (changes requested):
Overall it LGTM, thanks @AlexJones0 for the good work, I suggested some changes below.
I'm also blocking this PR with the "Request Changes" to avoid it merge before the end of the quiet period.

AlexJones0:
Now that quiet period has ended, I've rebased on the latest changes to `master`, and I've also addressed the remaining feedback on the comments, code style and `spi_host_winbond_flash_test`. 

Everything from the original PR should still be relevant, except for the comment on `spi_host_winbond_flash_test`. Rather than marking this as `manual` for the Verilator environment, the Verilator sim environment is now removed from this test entirely, as a result of the test requiring an SPI flash on the bus, which is not available in Verilator.

engdoreis (approved):

PR #24199, opened 2024-08-01
[dv,ast] Add timeunit for correct execution of delays in AST IF

sha-ron:
Adding \`timeunit and \`timeprecision to allow proper execution of '#1ps' inside 'GLITCH_VCMAIN_POK'.
Otherwise #1ps might be executed as 0 in certain cases.


eshapira (approved):
LGTM

PR #24201, opened 2024-08-01
[dv,usbdev] Improve testing of suspend transitions

alees24:
Some of the link transitions were previously not observed because (i) the driver/sequences typically do not model the USB by periodically sending the SOF packets when the bus is Idle. (ii) the Powered link state is transient and difficult to observe because the DP/DN both low state before VBUS connection acts as a Bus Reset. It therefore tends to transition quickly to ActiveNoSOF. Update the BFM to reflect the Powered -> PoweredSuspended transition.

Randomization is now required for the `usbdev_rx_full` sequence to be useful, since its scope has been extended.

### DIFF
--- a/hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
+++ b/hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
@@ -450,8 +450,6 @@
     {
       name: usbdev_rx_full
       uvm_test_seq: usbdev_rx_full_vseq
-      // No need for randomization.
-      reseed: 1
     }
     {
       name: usbdev_rx_pid_err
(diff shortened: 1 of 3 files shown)


hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
```diff
@@ -450,8 +450,6 @@
     {
       name: usbdev_rx_full
       uvm_test_seq: usbdev_rx_full_vseq
-      // No need for randomization.
-      reseed: 1
```
andreaskurth, line 454: Is this change intended to be part of this PR?

andreaskurth (approved):
This looks sensible to me; just one question

hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
alees24: Yes, as per the final comment. Perhaps you'd prefer it as a separate commit or PR? It should have been changed when that sequence was changed a couple of days ago.

andreaskurth: Ah indeed, sorry. Yes, I think a separate commit would be preferable. I'll leave the decision about keeping it in this PR or creating a new one to you

alees24: Okay, separated. I'll just let it run the basic checks again, thanks.

PR #24202, opened 2024-08-01
[dv,usbdev] Sequence to disrupt packet tx/rx with USB Bus Reset or IP block reset

alees24:
This PR adds a sequence that subjects the DUT to USB Bus Resets or IP block resets during the transmission or reception of packet data. Although the relevant logic in the module `usb_fs_tx.sv` is very simple the fact that bus resets were disjoint with packet transmission meant that a number of state transitions were not observed in coverage collection.

The sequence additionally uses the IP block reset and then reconfigures the DUT to ensure that it is possible to resume communication by receiving and transmitting packets. This is less of a concern than the idea that an inopportunely-timed USB Bus Reset might break packet reception/transmission logic or the OUT/IN packet engine state machines.

~~**The first commit belongs to PR #24206**~~

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_driver.sv
+++ b/hw/dv/sv/usb20_agent/usb20_driver.sv
@@ -103,10 +103,10 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     drive_packet("Token", pkt);
     // Attempt to collect a response only to full speed IN transactions; this should be DATA0|1
     // or ACK/NAK. For fault injection purposes we may sometimes want to avoid waiting for any
-    // response.
+    // response. Likewise for aborted transmissions.
     if (req_item.m_pid_type == PidTypeInToken && req_item.await_response &&
-        !req_item.low_speed) begin
-      device_response(rsp_item);
+        !req_item.low_speed && req_item.bits_to_transmit == 0) begin
+      device_response(rsp_item, req_item.bits_to_receive);
       seq_item_port.item_done(rsp_item);
       `uvm_info (`gfn, $sformatf("In drive afer In packet : \n %0s", rsp_item.sprint()), UVM_DEBUG)
     end else begin
@@ -125,11 +125,13 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
     pkt.data = {<<{pkt.data}};
     pkt.crc16 = {<<{pkt.crc16}};
     drive_packet("Data", pkt);
-    // Isochronous OUT transactions and low speed transactions do not yield a response.
-    if (req_item.m_usb_transfer == IsoTrans || req_item.low_speed) begin
+    // Isochronous OUT transactions, low speed transactions and aborted transmissions do not yield a
+    // response.
+    if (req_item.m_usb_transfer == IsoTrans || req_item.low_speed ||
+        req_item.bits_to_transmit != 0) begin
       seq_item_port.item_done();
     end else begin
-      device_response(rsp_item);
+      device_response(rsp_item, req_item.bits_to_receive);
       seq_item_port.item_done(rsp_item);
     end
   endtask
@@ -190,12 +192,14 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
   endtask
 
   // Drive out the main body of a packet, prefixing with the SYNC signaling, performing NRZI
-  // encoding and bit stuffing.
+  // encoding and bit stuffing. Optionally a non-zero value may be specified for 'bits_to_transmit'
+  // to truncate the packet.
   //
   // Note that since this task is used both to drive out the PRE PID and then subsequently the
   // low speed packet that follows it, the EOP is not emitted and must be appended by the caller
   // where appropriate.
-  task drive_packet_body(string pkt_type, bit valid_sync, bit driver_pkt[], bit low_speed = 1'b0);
+  task drive_packet_body(string pkt_type, bit valid_sync, bit driver_pkt[], bit low_speed = 1'b0,
+                         int unsigned bits_to_transmit = 0);
     bit comp_pkt[];
     bit nrzi_out[];
     bit bit_stuff_out[];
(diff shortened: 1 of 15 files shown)


hw/dv/sv/usb20_agent/usb20_driver.sv
```diff
@@ -231,8 +235,13 @@ class usb20_driver extends dv_base_driver #(usb20_item, usb20_agent_cfg);
...
       `uvm_info(`gfn, "Driver chose to corrupt bit 1 of SYNC signal (limited)", UVM_MEDIUM)
     end
-    // Drive out each bit in turn.
-    for (int i = 0; i < nrzi_out.size(); i++) begin
+    // Drive out each bit in turn, unless we've been asked to truncate the packet.
+    if (bits_to_transmit) begin
+      `uvm_info(`gfn, $sformatf("Driver asked to abort transmission after %0d bits",
+                                bits_to_transmit), UVM_MEDIUM)
+      if (bits_to_transmit > nrzi_out.size()) bits_to_transmit = nrzi_out.size();
+    end else bits_to_transmit = nrzi_out.size();
```
rswarbrick, line 243: I think this might be a bit confusing: if there are 128 bits to send and `bit_to_transmit = 129`, we'll send all the bits but then avoid sending the EOP. Maybe the best thing is slightly restructure to something like this:
```systemverilog
    // Drive out each bit in turn, unless we've been asked to truncate the packet.
    if (bits_to_transmit == 0 || bits_to_transmit >= nrzi_out.size()) begin
      bits_to_transmit = nrzi_out.size();
    end else begin
      `uvm_info(`gfn, $sformatf("Driver asked to abort transmission after %0d bits",
                                bits_to_transmit), UVM_MEDIUM)
    end
```
We'll still get the message about skipping EOP transmission.

hw/ip/usbdev/dv/env/seq_lib/usbdev_base_vseq.sv
```diff
@@ -50,6 +50,10 @@ bit inject_invalid_data_sync = 1'b0;
 bit inject_bad_token_crc5 = 1'b0;
 bit inject_bad_data_crc16 = 1'b0;
 
+// Trunctated transmission and reception of packets.
```
rswarbrick, line 53: Nit: Spelling of truncated

hw/ip/usbdev/dv/env/seq_lib/usbdev_tx_rx_disruption_vseq.sv
```diff
@@ -0,0 +1,129 @@
...
+        `uvm_info(`gfn, $sformatf("Setting device address to 0x%02x", dev_addr), UVM_MEDIUM)
+        usbdev_set_address(dev_addr);
+      end
+      default: `uvm_fatal(`gfn, $sformatf("Unexpected stimulus %p", stim_type))
+    endcase
+
+    // Try the smoke sequence again.
+    super.body();
+    // Choose another endpoint.
+    ep_default = NEndpoints - 1 - ep_default;
```
rswarbrick, line 122: This works because `NEndpoints - 1` is odd. But I had to think for a few seconds to convince myself this was always an involution. Maybe it makes sense to explicitly use `DV_CHECK_STD_RANDOMIZE_WITH_FATAL`?

hw/ip/usbdev/dv/env/timed_reg.sv
```diff
@@ -137,7 +140,7 @@ class timed_reg_field extends uvm_object;
                                   field.get_name(), pred.id, pred.val_new), UVM_HIGH)
         void'(pred_q.pop_front());
         obs_latest = act_val;
-      end else if (!(pred.latest_time - time_now > 0)) begin
+      end else if (!(pred.latest_time - time_now >= 0)) begin
```
rswarbrick, line 143: I agree that the change makes sense, but maybe it would be worth rephrasing to make it a bit more obvious. I think we're trying to say  there's a problem if `pred.latest_time` is before now ("we thought the change would have happened by now")? So maybe
```systemverilog
      end else if (pred.latest_time < time_now) begin
```
(which is equivalent) would be easier to understand?

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
```diff
@@ -102,19 +102,48 @@ class usbdev_scoreboard extends cip_base_scoreboard #(
...
+        handle_resets();
       join_none
     end
   endtask
 
+  // Handle DUT resets and ensure that the BFM is informed.
+  virtual task handle_resets();
+    forever begin
+      // Just ensure that we sample the reset line often enough that we never miss a reset;
+      // default reset width is 50-100 DUT clocks.
```
rswarbrick, line 121: This is reasonable, but wouldn't it be easier to just `wait(cfg.under_reset)` (and drop the `if` check)?

hw/ip/usbdev/dv/env/usbdev_timed_regs.sv
```diff
@@ -80,7 +80,7 @@ class usbdev_timed_regs extends uvm_object;
 
   // This process checks every prediction that is made, using backdoor csr_rd in zero time to avoid
   // interfering with actual CSR reads and the timing of the simulation.
-  task check_predictions();
+  task check_predictions(ref bit under_reset);
```
rswarbrick, line 83: I don't think this task writes `under_reset`. Maybe clearer to drop the `ref`?

hw/ip/usbdev/dv/env/seq_lib/usbdev_tx_rx_disruption_vseq.sv
alees24: Fair point. Unlikely that we'd ever have a design with an odd number of endpoints, but theoretically we support that as far as I know.

hw/ip/usbdev/dv/env/timed_reg.sv
alees24: It's not equivalent, because of the potential for wraparound.

hw/ip/usbdev/dv/env/usbdev_timed_regs.sv
alees24: It doesn't change it, but it does run forever and needs to be aware of resets that occur whilst running.

hw/ip/usbdev/dv/env/usbdev_scoreboard.sv
alees24: Okay; that was how the DV code evolved. I thought that there was a chance I may need to invoke 'dut_reset' a number of times throughout the reset signaling because the of other components still being active but have not found that to be necessary.

hw/dv/sv/usb20_agent/usb20_driver.sv
alees24: It's intended to inform of the request rather than the action of having truncated the packet, and the check was introduced just for robustness. In fact, since the 'bits_to_transmit' value is calculated and because the operation of bit stuffing can make the encoded packet longer, the condition you describe cannot actually occur. It would also lead to erroneous predictions if the DV proceeds to issue a Bus Reset condition in place of the dropped EOP, since they are one and the same to the DUT.

alees24:
I've made a couple more twiddles to this sequence to increase FSM coverage a little further.

hw/ip/usbdev/dv/env/timed_reg.sv
rswarbrick: Gah, good point. This looks sensible to me.

hw/ip/usbdev/dv/env/usbdev_timed_regs.sv
rswarbrick: Oh, I see! Yep, makes sense, thanks.

rswarbrick (approved):

PR #24204, opened 2024-08-02
[csrng/dv] Properly trigger the status error responses

### LINKED ISSUE #22869: [csrng/dv] Adjust csrng_intr test to properly test sw/hw_cmd_ack_sts
### Description

With #21981, CSRNG gained support for properly triggering and signaling command errors back to EDN as well as the software instance. This is great. However, the `csrng_intr` test still tests the corresponding status register bits by forcing them. And there is TODO in the DV code referring to the original issue https://github.com/lowRISC/opentitan/issues/16516

Now that CSRNG can actually signal these errors, the `csrng_intr` test should be reworked to properly trigger those errors instead of just forcing the corresponding status bits.

h-filali:
This PR changes the interrupt vseq to properly trigger the command error status responses instead of forcing the signals.

One thing we might need to discuss is that we run into issues where if a command with clen > 0 fails we don't absorb the whole command. This might be completely fine but still worth discussing.

Resolves #22869

### DIFF
--- a/hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
+++ b/hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
@@ -12,48 +12,7 @@ class csrng_intr_vseq extends csrng_base_vseq;
 
   csrng_item   cs_item;
   string       path1, path2, path3, path4, path_push, path_full, path_pop, path_not_empty, path;
-
-  // TODO(#22869): Instead of forcing ack_sts, actually generate the different error conditions.
-  task force_ack_and_ack_sts(int inst_idx, csrng_cmd_sts_e cmd_status);
-    bit [31:0] value_ack_sts;
-    string path_ack, path_ack_sts;
-
-    // Get paths
-    // ack is a 1D packed array - 1 bit per instance
-    path_ack = $sformatf("tb.dut.u_csrng_core.cmd_stage_ack[%0d]", inst_idx);
-    // ack_sts is a 1D packed array of enums -> a 2D packed array where the first packed dimension
-    // indexes the enum
-    path_ack_sts = "tb.dut.u_csrng_core.cmd_stage_ack_sts";
-
-    // Check paths and read the current ack_sts value
-    if (!uvm_hdl_check_path(path_ack)) begin
-      `uvm_fatal(`gfn, $sformatf("Path %s not found", path_ack))
-    end
-    if (!uvm_hdl_check_path(path_ack_sts)) begin
-      `uvm_fatal(`gfn, $sformatf("Path %s not found", path_ack_sts))
-    end else if (!uvm_hdl_read(path_ack_sts, value_ack_sts)) begin
-      `uvm_error(`gfn, $sformatf("Path %s could not be read", path_ack_sts))
-    end
-
-    // Inject the defined status bits at the correct positon.
-    // ack_sts is a 1D packed array of enums. This corresponds to a 2D packed array where the first
-    // packed dimension indexes the enum. We have to interpret this as a flat 1D packed array.
-    value_ack_sts =
-        value_ack_sts & ~({CSRNG_CMD_STS_WIDTH{1'b1}} << (CSRNG_CMD_STS_WIDTH * inst_idx));
-    value_ack_sts = value_ack_sts | ({cmd_status} << (CSRNG_CMD_STS_WIDTH * inst_idx));
-
-    // Force signals
-    if (!uvm_hdl_force(path_ack, 1'b1)) begin
-      `uvm_error(`gfn, $sformatf("Path %s could not be forced", path_ack))
-    end else begin
-      `uvm_info(`gfn, $sformatf("Path %s forced to 1", path_ack), UVM_MEDIUM)
-    end
-    if (!uvm_hdl_force(path_ack_sts, value_ack_sts)) begin
-      `uvm_error(`gfn, $sformatf("Path %s could not be forced", path_ack))
-    end else begin
-      `uvm_info(`gfn, $sformatf("Path %s forced to %b", path_ack_sts, value_ack_sts), UVM_MEDIUM)
-    end
-  endtask
+  bit [31:0]   backdoor_err_code_val;
 
   task release_ack_and_ack_sts(int inst_idx);
     string path_ack, path_ack_sts;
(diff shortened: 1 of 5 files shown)


hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
```diff
@@ -134,46 +92,92 @@ class csrng_intr_vseq extends csrng_base_vseq;
...
+    send_cmd_req(app, cs_item, .exp_sts(CMD_STS_INVALID_ACMD));
+
+    if (app != SW_APP) begin
+      // Expect/Clear interrupt bit
+      check_interrupts(.interrupts((1 << HwInstExc)), .check_set(1'b1));
+      cfg.clk_rst_vif.wait_clks(100);
+      // Make sure the interrupt bit is cleared
+      csr_rd_check(.ptr(ral.intr_state.cs_hw_inst_exc), .compare_value(1'b0));
+    end
+  endtask // trigger_invalid_acmd_sts_err
```
vogelpi, line 110: IIRC, we now correctly trigger these errors but we don't actually check the errors are reported anymore, right? Because previously, we did have the `csr_spinwait(.ptr(ral.intr_state.cs_hw_inst_exc), .exp_data(1'b1));` and `csr_spinwait(.ptr(ral.sw_cmd_sts.cmd_sts), .exp_data(CMD_STS_INVALID_CMD_SEQ));` calls etc. I can't similar calls anymore. Also, the calls to the sampling functions seem to be gone. Did you do a full coverage run with this PR?

vogelpi: Ah, now I understand: the checking happens inside the `send_cmd_req()` calls. So this is good. The calls to the sampling functions are still gone, did you check the coverage?

hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
```diff
@@ -134,46 +92,92 @@ class csrng_intr_vseq extends csrng_base_vseq;
...
-    cfg.clk_rst_vif.wait_clks(100);
-    cov_vif.cg_err_code_sample(.err_code(32'b0));
+    if (cfg.which_cmd_inv_seq != INS) begin
+      cs_item.randomize();
+      cs_item.acmd  = INS;
+      `uvm_info(`gfn, $sformatf("%s", cs_item.convert2string()), UVM_DEBUG)
+      send_cmd_req(app, cs_item);
+    end
+    `uvm_info(`gfn, $sformatf("Setting the reseed interval to %d", cfg.reseed_interval), UVM_DEBUG)
+    for (int i=0; i<cfg.reseed_interval; i++) begin
```
vogelpi, line 147: Nit: `    for (int i = 0; i < cfg.reseed_interval; i++) begin`

hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
```diff
@@ -357,11 +367,11 @@ class csrng_intr_vseq extends csrng_base_vseq;
...
     test_cs_entropy_req();
 
-    // Test cs_hw_inst_exc interrupt
-    test_cs_hw_inst_exc();
-
-    // Test faulty SW APP response by forcing sts output
-    test_cs_sw_cmd_sts();
+    // Test the command status response errors and for the HW apps
+    // the cs_hw_inst_exc interrupt.
+    for (int app=0; app<=SW_APP; app++) begin
```
vogelpi, line 372: Nit: `for (int app = 0; app <= SW_APP; app++) begin`

hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
```diff
@@ -134,46 +92,92 @@ class csrng_intr_vseq extends csrng_base_vseq;
...
+      cfg.clk_rst_vif.wait_clks(100);
+      // Make sure the interrupt bit is cleared
+      csr_rd_check(.ptr(ral.intr_state.cs_hw_inst_exc), .compare_value(1'b0));
+    end
+  endtask // trigger_reseed_interval_sts_err
 
-    // Release last error so that we're back to normal operation.
-    release_ack_and_ack_sts(1);
+  task test_cmd_sts_errs(uint app);
+    trigger_invalid_acmd_sts_err(app);
```
vogelpi, line 171: What is currently missing is the `INVALID_GEN_CMD` error. If I understand correctly, this error cannot be triggered under normal operating conditions as the `csrng_ctr_drbg_gen` module is only invoked for GEN commands and it then internally changes the ACMD from GEN to GENB and then GENU. The error is only set if the final ACMD is different from GENU. So to trigger this, one would need to perform FI. This feels a bit like the previous checks we had in the RTL that we removed like ensuring that after a UNI command the state is indeed 0. I think we should probably remove it for a future release.

Could you please create a new issue where we can discuss whether we want to completely remove this error (e.g. if we find the FI attack wouldn't be useful) or whether we should keep it? Thanks.

vogelpi (approved):
Thanks @h-filali for the PR, this looks good. I have some minor nits and questions.

hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
h-filali: Yes you are correct. I created the issue here: https://github.com/lowRISC/opentitan/issues/24224

vogelpi: Thanks!

hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
h-filali: Yes you understood correctly. I have now added the coverage sample calls.

h-filali:
@vogelpi I added some changes to how the coverage is collected for the cmd status response. And there were some failures whenever coverage was collected.

I also changed the coverage collection for EDN where necessary and I found a small coverage hole. The cg_cs_cmd_response_sample used to (and still does) sample only SUCCESS responses. We might want to create an issue for this.

h-filali:
I have now created a follow up PR for the EDN coverage #24244.

hw/ip/csrng/dv/env/seq_lib/csrng_intr_vseq.sv
vogelpi: Thanks, great work!

vogelpi:
Thanks for checking the coverage and for the additional changes. LGTM!

Issue #24208, opened 2024-08-02
[top_earlgrey,cdc] Complete CDC signoff


andreaskurth:
- [ ] Go through remaining errors and warnings in iDebug and review them with designers.
  - Preferably tune tool settings such that no waivers are needed.
  - Add waivers for remaining errors and warnings.
- [ ] Remove `ast_clks_byp` from the list of black-boxed modules.
- [ ] Analyze expressions that are currently black-boxed. Increase `ri_max_total_range_bits` if necessary, and ideally find a way to enable this on machines that have enough RAM.

Issue #24207, opened 2024-08-02
[dvsim] Implement report parsing for MeridianCDC


andreaskurth:
Currently dvsim cannot parse reports generated by MeridianCDC. Even for successful runs, dvsim exits with an error. The user can then use iDebug to analyze the results with a GUI. Once CDC for Earlgrey is clean, we may want to add CDC checks to nightly regressions, for which we would need dvsim to parse the reports and produce results (in HTML and JSON, as usual).

PR #24209, opened 2024-08-02
[top_earlgrey,cdc] Transition to MeridianCDC

### LINKED ISSUE #13091: [cdc] clean-up cdc constraints
with #13016. cdc constraints are now being introduced to the flow.
These constraints are separate from sdc constraints and need their own mechanism of sourcing (right now I am temporarily including it into `run-cdc.tcl`, which is not right).

Further, some of the cdc constraints should be made more specific, as it is too broad from async_fifo rdata output to ibex rvalid.

### LINKED ISSUE #16242: [CDC] Remove top specific config from `run-cdc-tcl`
[Current `run-cdc.tcl`](https://cs.opensource.google/opentitan/opentitan/+/2b989a42bad3aff373e024ce841b406eec1b3970:hw/cdc/tools/verixcdc/run-cdc.tcl;l=112-120;bpv=1;bpt=0) has **top_earlgrey** specific configs. It The part is better to be placed under `hw/top_earlgrey/cdc/` and source from `run-cdc.tcl` depending on the target top.

CC: @jeoongp @tjaychen @msfschaffner @eunchan

andreaskurth:
This transitions Earlgrey to MeridianCDC.  The files for VerixCDC get removed because we no longer support them.

Execute
```
util/dvsim/dvsim.py hw/top_earlgrey/cdc/chip_earlgrey_asic_cdc_cfg.hjson
```
to run CDC analysis on Earlgrey.

This resolves #13091 and resolves #16242 because the file that those issue references gets removed.

## Follow-up work (post merge)
- [ ] #24207
- [ ] #24208


hw/top_earlgrey/cdc/chip_earlgrey_asic_cdc_cfg.hjson
```diff
@@ -24,17 +24,11 @@
...
-
   // Main CDC waiver file
-  cdc_waiver_file: "{proj_root}/hw/top_earlgrey/cdc/cdc_waivers.tcl"
-
-  // CDC customized env file for ast integration
-  cdc_user_env: "{proj_root}/hw/top_earlgrey/cdc/top_user.env"
+  cdc_waiver_file: ""
 
-  // AST.lib
-  ast_lib: "{proj_root}/hw/top_earlgrey/ip/ast/lib/ast.lib"
```
vogelpi, line 37: Do you know what the purpose of this lib file was?

hw/cdc/tools/verixcdc/run-cdc.tcl
```diff
@@ -1,237 +0,0 @@
...
-read_sdc  $CONSTRAINT  -output_env constraints.sdc.env
-current_scenario env
-read_env constraints.sdc.env
-touch $USER_ENV_FILE
-read_env $USER_ENV_FILE
-
-# RI's recommended approach to filter out unnecessary errors on async_fifo
-catch {rm tmp.env}
-set outfile [open tmp.env  w]
-puts $outfile "set_data_clock_domain rdata_o -module prim_fifo_async -derived_from rvalid_o"
```
vogelpi, line 129: Is this something we should do in Meridian as well?

hw/cdc/tools/meridiancdc/run-cdc.tcl
```diff
@@ -0,0 +1,76 @@
...
+
+current_scenario env
+read_env $envs
+
+touch $ENV_FILE
+read_env $ENV_FILE
+
+analyze_intent
+
+verify_cdc
```
vogelpi, line 76: I am obviously no expert but comparing a bit the flow files for Verix and Meridian, it seems that there are some steps missing in the new flow like:
- setting the top modules
- reading in waivers
- writing out the report (I am aware that the parsing is currently missing and I believe that's fine at this point)

Does it make sense to add placeholders in the tcl script now (as the Verix reference is removed with this PR) or is adding these things anyway very obvious and tracked in an issue. What is your view?

vogelpi (approved):
Thanks @andreaskurth for your work on this. I do have a couple of question but nothing that should gate this PR.

hw/top_earlgrey/cdc/chip_earlgrey_asic_cdc_cfg.hjson
andreaskurth: This was a liberty model of AST, which isn't available in the open-source domain. As the CDC flow no longer uses closed-source models/code, this can now be removed

hw/cdc/tools/meridiancdc/run-cdc.tcl
andreaskurth: Thanks for these inputs!

- Setting the top modules: As we run CDC on the full chip (`chip_earlgrey_asic`), we don't need to set top modules other than `DUT` (which we do). The previous flow would have allowed to run CDC at the block level, which we currently don't intend to do, because most CDCs are between HW IP blocks.
- Reading in waivers: Currently there are no waivers, but at some point (and for generality), I agree that it would be useful to support loading a waiver file in the run script. I'll add it.
- Writing out reports: Right, I could as well add this to this PR already.

hw/cdc/tools/verixcdc/run-cdc.tcl
andreaskurth: Yes, added

vogelpi:
Thanks for addressing my comments this looks good. The AST library question we should probably discuss on Monday.

andreaskurth:
> Thanks for addressing my comments this looks good. The AST library question we should probably discuss on Monday.

Sounds good, but let's address that in a follow-up PR if necessary

PR #24215, opened 2024-08-05
[otbn,util] Add clobbered-register print mode for analyzer.

jadephilipoom:
Convenience mode that prints the "clobbered register:" and "clobbered flag groups:" lines as they should appear in the docstring for OTBN functions.

Usage looks like:
```
 $ hw/ip/otbn/util/analyze_information_flow.py p256_ecdsa.elf --subroutine p256_sign --clobbered
* clobbered registers: x2 to x3, x10, x16, x18 to x22, w0 to w31, acc, mod
* clobbered flag groups: FG0
```

I've been meaning to do this for a long while, since it's a straightforward extension of the information-flow analyzer we already have and is helpful for writing and checking OTBN docstrings :slightly_smiling_face:

### DIFF
--- a/hw/ip/otbn/util/analyze_information_flow.py
+++ b/hw/ip/otbn/util/analyze_information_flow.py
@@ -25,6 +25,11 @@ def main() -> int:
         action='store_true',
         help=('Print full control-flow and information-flow graphs.'))
     parser.add_argument(
+        '--clobbered',
+        action='store_true',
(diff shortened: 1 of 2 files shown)


rswarbrick (approved):
Sorry for not noticing this earlier. The change looks good to me.

jadephilipoom:
CI failure is pretty clearly not related.

Issue #19027, opened 2023-06-22
[edn/dv] Improve testing and coverage of endpoint requests during fatal alerts


andreaskurth:
See TODOs associated with this issue in the code for details.

vogelpi:
What we have in terms of comments is the following:
https://github.com/lowRISC/opentitan/blob/acd065ca7dab30b94a59ec33bdc6c167801fcfbd/hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv#L59-L84

I don't think this is going to be trivial and believe we have more critical stuff to take care off right now. Thus moving to M7.

PR #24217, opened 2024-08-05
[edn/dv] Add EDN endpoint requests while EDN is in error state

h-filali:
For further information check out the individual commits.

I didn't add multiple requests per endpoint, since lowering the req signal would be a protocol violation.
Furthermore, I could add a more in depth coverage which checks if entropy requests happen in each main SM state.
However, to do this I think I would have to create a proper interface for the endpoint req/res signals, which would be more work.
@vogelpi should we create an issue for this?

Resolves [#19027](https://github.com/lowRISC/opentitan/issues/19027)

### DIFF
--- a/hw/ip/edn/dv/cov/edn_cov_if.sv
+++ b/hw/ip/edn/dv/cov/edn_cov_if.sv
@@ -305,6 +305,16 @@ interface edn_cov_if (
     }
   endgroup : edn_hw_cmd_sts_cg
 
+  covergroup edn_endpoint_err_req_cg with function sample(uint endpoint);
+    option.name         = "edn_endpoint_err_req_cg";
+    option.per_instance = 1;
+
+    // Create one bin per endpoint of EDN.
+    endpoint_cg: coverpoint endpoint {
+      bins range[MAX_NUM_ENDPOINTS] = {[0:MAX_NUM_ENDPOINTS-1]};
+    }
+  endgroup : edn_endpoint_err_req_cg
+
   `DV_FCOV_INSTANTIATE_CG(edn_cfg_cg, en_full_cov)
   `DV_FCOV_INSTANTIATE_CG(edn_endpoints_cg, en_full_cov)
   `DV_FCOV_INSTANTIATE_CG(edn_cs_cmds_cg, en_full_cov)
@@ -313,6 +323,7 @@ interface edn_cov_if (
   `DV_FCOV_INSTANTIATE_CG(edn_cs_cmd_response_cg, en_full_cov)
   `DV_FCOV_INSTANTIATE_CG(edn_sw_cmd_sts_cg, en_full_cov)
   `DV_FCOV_INSTANTIATE_CG(edn_hw_cmd_sts_cg, en_full_cov)
+  `DV_FCOV_INSTANTIATE_CG(edn_endpoint_err_req_cg, en_full_cov)
 
   // Sample functions needed for xcelium
   function automatic void cg_cfg_sample(edn_env_cfg cfg);
@@ -373,5 +384,9 @@ interface edn_cov_if (
                                                    bit cmd_ack, csrng_pkg::acmd_e acmd);
     edn_hw_cmd_sts_cg_inst.sample(boot_mode, auto_mode, cmd_sts, cmd_ack, acmd);
   endfunction : cg_edn_hw_cmd_sts_sample
+
+  function automatic void cg_edn_endpoint_err_req_sample(uint endpoint);
+    edn_endpoint_err_req_cg_inst.sample(endpoint);
+  endfunction : cg_edn_endpoint_err_req_sample
 
 endinterface : edn_cov_if
--- a/hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
+++ b/hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
@@ -57,10 +57,11 @@ class edn_common_vseq extends edn_base_vseq;
   endtask
 
   virtual task send_edn_requests_during_fatal_alerts();
+    bit [edn_pkg::StateWidth-1:0] state;
+    bit req;
     push_pull_host_seq#(edn_pkg::FIPS_ENDPOINT_BUS_WIDTH) m_endpoint_pull_seq[MAX_NUM_ENDPOINTS];
+    bit [MAX_NUM_ENDPOINTS-1:0] send_edn_reqs = $urandom();
 
-    // TODO(#19027): can only send one request due to TB errors. Fix it.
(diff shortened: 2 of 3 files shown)


hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
```diff
@@ -57,29 +58,41 @@ class edn_common_vseq extends edn_base_vseq;
   endtask
 
   virtual task send_edn_requests_during_fatal_alerts();
+    bit [edn_pkg::StateWidth-1:0] state;
+    bit req;
     push_pull_host_seq#(edn_pkg::FIPS_ENDPOINT_BUS_WIDTH) m_endpoint_pull_seq[MAX_NUM_ENDPOINTS];
+    bit [MAX_NUM_ENDPOINTS-1:0] send_edn_reqs = {MAX_NUM_ENDPOINTS{1'b1}};
```
vogelpi, line 64: Would it better to randomize this?

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
```diff
@@ -57,29 +58,41 @@ class edn_common_vseq extends edn_base_vseq;
...
         fork
           m_endpoint_pull_seq[index].start(p_sequencer.endpoint_sequencer_h[index]);
+          begin
+            `DV_SPINWAIT(
+                do begin
+                  uvm_hdl_read($sformatf("%0s[%0d].edn_req", "tb.dut.u_edn_core.edn_i", index),
+                               req);
+                  cfg.clk_rst_vif.wait_clks(1);
+                end while (!req);,
+                $sformatf("Waiting for EDN endpoint request timeout")
```
vogelpi, line 86: This message seems not to make sense to me. From what I understand, we are waiting for the request to raise. The message will be printed only if we never see the request going up.

IIUC, this task should rather to the following and in this order:
1. Check that we are in the terminal error state.
2. Send the request. (The checking is done via SVA as mentioned.)
3. Sample the coverpoint. If the terminal error state isn't reached, we shouldn't send the request (there is no point). If we don't see the request going up, we shouldn't sample the covergroup.

Right now, we wait for the request to be high and then check the FSM state. If the SPINWAIT times out the request never got high but we still sample the covergroup.

Does this make sense to you?

vogelpi (commented):
Thanks for the PR @h-filali , I have some questions and I think the `send_edn_requests_during_fatal_alerts()` task should be modified to not sample the covergroup if the request is never seen (see comments).

vogelpi:
> I didn't add multiple requests per endpoint, since lowering the req signal would be a protocol violation.

This makes sense to me.

> Furthermore, I could add a more in depth coverage which checks if entropy requests happen in each main SM state. However, to do this I think I would have to create a proper interface for the endpoint req/res signals, which would be more work. @vogelpi should we create an issue for this?

That sounds like a good extension and it doesn't hurt to create the issue. I consider it a lower priority though - at the moment I am not sure if we gain a lot by that. But I can be convinced otherwise :-)

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
```diff
@@ -17,6 +17,7 @@ class edn_common_vseq extends edn_base_vseq;
   endtask
 
   virtual task body();
+    `uvm_info(`gfn, $sformatf("TESSSSSSSSSSSSSTTT"), UVM_LOW)
```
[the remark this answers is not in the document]
h-filali: Oh wow sorry I missed that.

hw/ip/edn/dv/cov/edn_cov_if.sv
```diff
@@ -305,6 +305,15 @@ interface edn_cov_if (
     }
   endgroup : edn_hw_cmd_sts_cg
 
+  covergroup edn_endpoint_err_req_cg with function sample(uint endpoint);
+    option.name         = "edn_endpoint_err_req_cg";
+    option.per_instance = 1;
+
+    endpoint_cg: coverpoint endpoint {
+      bins range[MAX_NUM_ENDPOINTS] = {[0:MAX_NUM_ENDPOINTS-1]};
+    }
```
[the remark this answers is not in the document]
h-filali: Yes this creates one bin per endpoint. I'll add a comment.

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
h-filali: I don't know if we gain much from randomizing this but yes I can randomize this.

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
h-filali: We already check if we are in the error state beforehand. So 1 and 2 already happen. If we don't wait for the request to go high, I think it would be possible for the request to go high when we are back in the idle state. Last I'm just checking the main SM again to make sure that the request happened in the error state. Otherwise we just don't sample for coverage. AFAIK if the timeout is reached a fatal alert will be triggered, so the sampling will not happen unless the request goes high in the error state.

h-filali:
@vogelpi thanks for your review. I implemented all your suggestions except for one comment where I need further feedback.
Regarding the extension, I don't feel strongly about it. It would just be a bit cleaner. I think I'll leave it be then.

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
vogelpi: Okay, thanks for explaining, this makes sense.

As for the timeout of the DV_SPINWAIT macro: you are totally right:
```
// Wait for a statement, but exit early after a timeout
`ifndef DV_SPINWAIT
`define DV_SPINWAIT(WAIT_, MSG_ = "timeout occurred!", TIMEOUT_NS_ = default_spinwait_timeout_ns, ID_ =`gfn) \
  `DV_SPINWAIT_EXIT(WAIT_, `DV_WAIT_TIMEOUT(TIMEOUT_NS_, ID_, MSG_);, "", ID_)
`endif
```
It calls DV_SPINWAIT_EXIT with DV_WAIT_TIMEOUT as argument. The latter is triggering the error. I overlooked that one. Sorry!

vogelpi (approved):
LGTM, thanks for the changes and explanations!

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
h-filali: Thanks for checking!

vogelpi:
I am merging this as the relevant CI steps ran through and all passed. This PR changes only block-level DV files and doesn't have an impact on SW or FPGA tests.

PR #24079, opened 2024-07-22
[dv,flash_ctrl] Fix many DV failures

matutem:
This PR has a number of commits:
- Fix tracking of injected errors
- Align legacy and otf rand_op constraints
- Fix tracking of addresses written
- Handle data errors in host reads
- Adjust number of transactions in some tests

With these fixes the flash_ctrl regression has much fewer errors left.

I intend to close the umbrella issue #22879, and open issues for more specific failures.

Fixes #22879

PR #24219, opened 2024-08-05
[flash/dv] Add back the exclude_info2 condition to legacy_base_vseq

eshapira:
Hi,
This PR added back the recently removed exclude_info2 condition to flash_ctrl_legacy_base_vseq.
This condition is required because info2 has special use in the extending environment in which it can't be tested in those tests, so the flag added to allow running those tests without using info2 partition.
Thanks!


vogelpi (approved):
This looks good to me. It's inline with the previous implementation before #24079 and this is required for the closed source DV. I am merging this now such that it gets into the release branch.

Thanks for the PR and the clear explanation @eshapira !

PR #24228, opened 2024-08-06
[dv,usbdev] Introduce VBUS disconnect/reconnect to `usbdev_link_suspend`

alees24:
This PR introduces randomized VBUS disconnection and reconnection to the sequence `usbdev_link_suspend` and its derived sequences to increase the coverage of FSM transitions in the `usbdev_linkstate` module of the DUT.

It also fixes a subtle fault in the Resume Signaling from the monitor to the scoreboard which for some clock frequencies and sampling phases could arrive too late to predict the change in the `usbstat.link_state` field. The result was a test failure for a very small number of seeds; this was observed with one simulator when running a larger number of seeds to check for regressions in the altered sequences.


hcallahan-lowrisc (approved):
LGTM based on the intent of the commit message descriptions. Thanks @alees24

Issue #24225, opened 2024-08-06
[test-triage] KMAC shake test vectors regression


jwnrt:
### Hierarchy of regression failure

Block level

### Failure Description

Failures line up with when an ECO was merged to KMAC.
I believe it was decided these failures were P2 since the tests just need updating.

Still generating the exact error message...

### Steps to Reproduce

- GitHub Revision: [e4c5daa580](https://github.com/lowrisc/opentitan/tree/e4c5daa580105299c3f45e0728376382265e0f41)
- dvsim invocation command to reproduce the failure, inclusive of build and run seeds:
  ```sh
  ./util/dvsim/dvsim.py hw/ip/kmac/dv/kmac_masked_sim_cfg.hjson \
    -i kmac_test_vectors_shake_128 \
    --build-seed 57478527486894479494471273459769404654835266620222125964939301612221385668501 \
    --fixed-seed 11025088414661302387259625850754552315971192149546710050530591896985944121480 \
    --waves fsdb
  ```

### Tests with similar or related failures

- [ ] kmac_test_vectors_shake_128
- [ ] kmac_test_vectors_shake_256
- [ ] Both masked and unmasked versions

vogelpi:
Thanks for flagging this @jwnrt . I've now had a close look. The failures are not related to the ECOs but related to adding DV support for verifying the FIFO empty interrupt (https://github.com/lowRISC/opentitan/pull/23942). This change quite heavily impacts the simulation run time of some long running tests (most tests are not affected though) which lead to many timeouts. I've now filed a PR to increase the timeout for these tests here: #24233.

For reference:
- Regression results w/o #23942: https://reports.opentitan.org/hw/ip/kmac_masked/dv/2024.07.27_09.35.21/report.html
- Regression results w/ #23942: https://reports.opentitan.org/hw/ip/kmac_masked/dv/2024.07.28_10.43.38/report.html

jwnrt:
Thank you for the fix @vogelpi

jwnrt:
The `_256` variants of these tests are now passing regressions, but the `_128` versions are still failing. Do those timeouts need extending further?

vogelpi:
I have been experimenting with a different approach. These tests don't contribute much to the overall coverage but are extremely expensive (as running for several) hours. I'll file a PR.

PR #24233, opened 2024-08-06
[kmac/dv] Extend timeout values for long running tests

vogelpi:
As of lowRISC/OpenTitan#23942, the FIFO empty status-type interrupt is predicted and verified which unfortunately negatively impacts the simulation performance in some cases. Most tests are not notably impacted but for some rather long running tests, the maximum observed job runtime increased by up to 50%. As a result, we started seeing timeouts for some of these tests in the nightly regressions after merging the FIFO empty interrupt verification PR.

Therefore, this commit increases the timeout for those long running tests where we observed timeouts.

This resolves lowRISC/OpenTitan#24225.

### DIFF
--- a/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
+++ b/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
@@ -95,6 +95,7 @@
       name: "{name}_long_msg_and_output"
       uvm_test_seq: kmac_long_msg_and_output_vseq
       run_opts: ["+test_timeout_ns=10_000_000_000"]
+      run_timeout_mins: 90
     }
     {
       name: "{name}_sideload"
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
Clearly for a separate PR, but a 3 hour test seems a little insane!

Looking at `test_vectors_pkg.sv`, I think that we've got two entries in most of the file lists. Would it make sense to change `kmac_test_vectors_base_vseq` so that it only runs one of the vectors? We might then need to increase the number of seeds to ensure we do the same stuff, but in parallel.

That said, I think it looks like we're actually running 50 seeds of each of these (directed) tests. So I can't really see any disadvantage in just halving it and not worrying about it.

vogelpi:
> Clearly for a separate PR, but a 3 hour test seems a little insane!

Thanks @rswarbrick , I fully agree!

> Looking at `test_vectors_pkg.sv`, I think that we've got two entries in most of the file lists. Would it make sense to change `kmac_test_vectors_base_vseq` so that it only runs one of the vectors? We might then need to increase the number of seeds to ensure we do the same stuff, but in parallel.

That sounds like a good approach.

> That said, I think it looks like we're actually running 50 seeds of each of these (directed) tests. So I can't really see any disadvantage in just halving it and not worrying about it.

For HMAC, we've noted the same and @martin-velay also noticed that these tests contribute very little to the coverage and thus reduced the number of seeds quite a bit. This would also work here I guess.

PR #24234, opened 2024-08-06
[dv,usbdev] Fix failure in `usbdev_bitstuff_err` sequence

alees24:
This PR addresses a single failing seed observed on a recent regression run.

The sequence `usbdev_bitstuff_err` could inject bit stuffing violations from which the current RTL does not properly recover and the DUT instead tries to interpret the remainder of the OUT token packet as a separate packet, leading to a `rx_pid_err` assertion too.

Instead, ensure that only known, recoverable bit stuffing violations are exercised with the current RTL, pending a more robust recovery mechanism in a later revision. Changing the RTL was deferred as a risk mitigation; bit stuffing violations are not expected in practice.

It has now passed with 500 random seeds.


hcallahan-lowrisc (approved):
Code changes makes sense to me. 

Is there an open issue / documentation about the current shortcomings with regards to what bit-stuffing violations the RTL cannot recover from? If not, we should probably open an issue and tag this constraint against it so it doesn't get lost in the future.

Issue #24226, opened 2024-08-06
[test-triage] `edn_sec_cm` failures


jwnrt:
### Hierarchy of regression failure

Block level

### Failure Description

Started failing on 25th of July, 100% -> 0%.

```
  UVM_ERROR @  16565539 ps: (csr_utils_pkg.sv:476) [csr_utils::csr_rd_check] Check failed obs == exp (193 [0xc1] vs 382 [0x17e]) Regname: edn_reg_block.main_sm_state reset value: 0xc1
  UVM_INFO @  16565539 ps: (uvm_report_catcher.svh:705) [UVM/REPORT/CATCHER]
  --- UVM Report catcher Summary ---
```

### Steps to Reproduce

- GitHub Revision: [e4c5daa580](https://github.com/lowrisc/opentitan/tree/e4c5daa580105299c3f45e0728376382265e0f41)
- dvsim invocation command to reproduce the failure, inclusive of build and run seeds:
  ```sh
  ./util/dvsim/dvsim.py hw/ip/edn/dv/edn_sim_cfg.hjson \
    -i edn_sec_cm \
    --build-seed 57478527486894479494471273459769404654835266620222125964939301612221385668501 \
    --fixed-seed 34299359568848372465292589661168344678023408838600054326005816702909329066132 \
    --waves fsdb
  ```

vogelpi:
Thanks for flagging this @jwnrt . This is related to the ECO fix (flopping prim_count error outputs). I've now filed a PR to fix this here: 24237.

PR #24125, opened 2024-07-25
[prim_count] Register err_o to avoid potential CDC issues downstream

vogelpi:
This is related to lowRISC/OpenTitan#24119.

Issue #23988, opened 2024-07-11
[entropy_src] Factor distr FIFO errors into ERR_CODE.FIFO_[STATE,READ,WRITE]_ERR


vogelpi:
### Description

Currently, errors triggered inside the distribution FIFO aren't factored into the ERR_CODE.FIFO_[STATE,READ,WRITE]_ERR fields. They still lead to a fatal alert and they still show up as error in the ERR_CODE CSR, but the type of error isn't reflected. This is an RTL bug but it's obviously not very critical.

The reason is simply that I forgot to include that signals into the `fifo_[write,read,status]_err_sum` signals. So it's very easy to fix. From `entropy_src_core.sv`:
```sv
  assign fifo_write_err_sum =
         sfifo_esrng_err[2] ||
         sfifo_observe_err[2] ||
         sfifo_esfinal_err[2] ||
         err_code_test_bit[28];
  assign fifo_read_err_sum =
         sfifo_esrng_err[1] ||
         sfifo_observe_err[1] ||
         sfifo_esfinal_err[1] ||
         err_code_test_bit[29];
  assign fifo_status_err_sum =
         sfifo_esrng_err[0] ||
         sfifo_observe_err[0] ||
         sfifo_esfinal_err[0] ||
         err_code_test_bit[30];
```

vogelpi:
Related to this, I also noted that the signaling of errors via ERR_CODE CSR is handled a bit differently between ENTROPY_SRC and CSRNG/EDN:
- CSRNG/EDN only signal errors via ERR_CODE if the module is enabled

https://github.com/lowRISC/opentitan/blob/e087179b7f2c41c843668125a06e3a66fb60c05b/hw/ip/csrng/rtl/csrng_core.sv#L632-L635
 
https://github.com/lowRISC/opentitan/blob/e087179b7f2c41c843668125a06e3a66fb60c05b/hw/ip/edn/rtl/edn_core.sv#L345-L347

- ENTROPY_SRC always signals errors via ERR_CODE register

https://github.com/lowRISC/opentitan/blob/e087179b7f2c41c843668125a06e3a66fb60c05b/hw/ip/entropy_src/rtl/entropy_src_core.sv#L961-L963

I think it would be better aligning ENTROPY_SRC with the other two blocks to ease integrations in the future.

vogelpi:
This has been resolved by #28054.

PR #24237, opened 2024-08-06
[edn/dv] Allow flopped error output of prim_count to propagate to CSRs

vogelpi:
With https://github.com/lowRISC/opentitan/pull/24125, we added a register to the error output of all prim_count instances. This adds an additional delay until errors become actually visible via CSRs (e.g. main_sm_state, err_code). This commit adds a corresponding wait statement to take this into account.

This resolves https://github.com/lowRISC/opentitan/issues/24226.

### DIFF
--- a/hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
+++ b/hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
@@ -36,7 +36,7 @@ class edn_common_vseq extends edn_base_vseq;
       cfg.clk_rst_vif.wait_n_clks(2);
     end
 
-    if (!uvm_re_match("*.cnt_q*", if_proxy.path)) begin
+    if (!uvm_re_match("*.u_prim_count_max_reqs_cntr", if_proxy.path)) begin
       csr_rd_check(.ptr(ral.err_code.edn_cntr_err), .compare_value(1'b1));
       if (cfg.en_cov) begin
         csr_rd(.ptr(ral.err_code), .value(backdoor_err_code_val));
--- a/hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
+++ b/hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
@@ -29,12 +29,18 @@ class edn_common_vseq extends edn_base_vseq;
 
   virtual task check_sec_cm_fi_resp(sec_cm_pkg::sec_cm_base_if_proxy if_proxy);
     bit [31:0] backdoor_err_code_val;
+    bit [3:0] edn_enable;
+    bit exp_fifo_err;
     // The error output of the prim_count primitive is flopped, leading to a total delay of 2 clock
     // cycles until error become visible via CSRs.
     if (!uvm_re_match("*.u_prim_count_max_reqs_cntr", if_proxy.path) ||
         !uvm_re_match("*.u_fifo_cnt.gen_secure_ptrs*", if_proxy.path)) begin
       cfg.clk_rst_vif.wait_n_clks(2);
     end
+
+    // FIFO errors are only signaled via the ERR_CODE CSR while EDN is enabled.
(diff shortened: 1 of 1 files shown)


hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
```diff
@@ -29,7 +29,14 @@ class edn_common_vseq extends edn_base_vseq;
...
     bit [31:0] backdoor_err_code_val;
-    if (!uvm_re_match("*.cnt_q*", if_proxy.path)) begin
+    // The error output of the prim_count primitive is flopped, leading to a total delay of 2 clock
+    // cycles until error become visible via CSRs.
+    if (!uvm_re_match("*.u_prim_count_max_reqs_cntr", if_proxy.path) ||
+        !uvm_re_match("*.u_fifo_cnt.gen_secure_ptrs*", if_proxy.path)) begin
+      cfg.clk_rst_vif.wait_n_clks(2);
+    end
+
+    if (!uvm_re_match("*.u_prim_count_max_reqs_cntr", if_proxy.path)) begin
```
vogelpi, line 39: @h-filali this is most likely needed to fill a small coverage hole. It might be related to what you found today.

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
```diff
@@ -29,7 +29,14 @@ class edn_common_vseq extends edn_base_vseq;
 
   virtual task check_sec_cm_fi_resp(sec_cm_pkg::sec_cm_base_if_proxy if_proxy);
     bit [31:0] backdoor_err_code_val;
-    if (!uvm_re_match("*.cnt_q*", if_proxy.path)) begin
+    // The error output of the prim_count primitive is flopped, leading to a total delay of 2 clock
+    // cycles until error become visible via CSRs.
+    if (!uvm_re_match("*.u_prim_count_max_reqs_cntr", if_proxy.path) ||
+        !uvm_re_match("*.u_fifo_cnt.gen_secure_ptrs*", if_proxy.path)) begin
+      cfg.clk_rst_vif.wait_n_clks(2);
+    end
```
vogelpi, line 37: @rswarbrick I think you would have to do something similar for otbn_sec_cm as well.

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
h-filali, line 39: I might as well be missing something but aren't we missing the read checks for the u_fifo_cnt now?

h-filali (approved):
I just have one question. Otherwise this change sounds reasonable to me!

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
h-filali: Thanks for tagging me here, but it looks like the issue I found seems to be something different.

rswarbrick (approved):
I think this looks sensible to me.

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
vogelpi: Sounds great, thanks @rswarbrick !

hw/ip/edn/dv/env/seq_lib/edn_common_vseq.sv
vogelpi: Yes, that's right. We don't do the read checks for the FIFO counters. But we also didn't do that before. Let me take another look.

vogelpi: I've now added a third commit to check those bits in the ERR_CODE CSR. The tricky thing here is that EDN only records the FIFO errors in ERR_CODE when the module is enabled. This is a bit different from how this is done in ENTROPY_SRC, I've added a comment here https://github.com/lowRISC/opentitan/issues/23988#issuecomment-2275655206 to keep track of aligning the behavior in the future.

Issue #23869, opened 2024-07-01
[rom_ext] BootLog incorrect for currently executing ROM_EXT slot


jettr:
### Description

It appears that the boot log that reports the active ROM_EXT is incorrect in some scenarios. Notice the following from the abridge firmware log after performing a firmware update that contains both a new ROM_EXT and new Owner Firmware.

 * `Starting ROM_EXT 0.2`
 * `0x406007a8: 5f5f4141 00000000 00000002 00010000` (memory dump of BootLog section start with [`rom_ext_slot`](https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/lib/boot_log.h#L33)
 * `Starting ROM_EXT 0.3`
 * `0x406007a8: 5f5f4141 00000000 00000003 00010000`
    *   This is especially concerning since the `rom_ext_slot` is reporting SlotA (`AA__`), but the version is reporting `0.3`, but SlotA contains 0.2, and SlotB contains 0.3. I am pretty sure 0.3 is what actually ran since it printed `Starting ROM_EXT 0.3` on the UART.

The boot log remains incorrect event after performing soft reboots from FW (via `reboot` console command). Only after I perform a chip reset with the `RESET_L` pin, do I see the boot log be correct.

Log from a ROM_EXT and owner firmware update:

```
Starting ROM_EXT 0.2
Warning: UDS certificate not valid.
0: 40130000 NAPOT L--- sz=00001000
1: 40480000 NAPOT L--- sz=00000400
2: a0010400 ----- ---- sz=00000000
3: a0071d0c   TOR -X-R sz=0006190c
4: a0000000 NAPOT ---R sz=00080000
5: 00000000 ----- ---- sz=00000000
6: 00000000 ----- ---- sz=00000000
7: 00000000 ----- ---- sz=00000000
8: 00000000 ----- ---- sz=00000000
9: 00000000 ----- ---- sz=00000000
10: 20000400 ----- ---- sz=00000000
11: 20008a78   TOR -X-R sz=00008678
12: 20000000 NAPOT ---R sz=00100000
13: 00000000 ----- ---- sz=00000000
14: 40000000 NAPOT --WR sz=10000000
15: 10000000 NAPOT --WR sz=00020000
mseccfg = 00000002
entry: 0xa0010730



[       0.001] Starting OpenTitan RO 0.0.2 RW 0.36.1
[...]

PR #24241, opened 2024-08-07
[rom_ext] Set the boot slot on every boot

cfrantz:
The function `boot_log_check_or_init` is meant to initialize the `boot_log` structure if it hasn't already been initialized.  This is meant to initialize the boot_log if the ROM doesn't (it doesn't on the earlgrey_es chip), however, it will also skip initialization if the boot_log is already initialized from a prior boot.

Initialize only portions of the boot_log that will be constant across boots.  Defer setting the `rom_ext_slot` to `rom_ext_start` and set it to the current boot slot on every boot.

Fixes: #23869

### DIFF
--- a/sw/device/silicon_creator/lib/boot_log.c
+++ b/sw/device/silicon_creator/lib/boot_log.c
@@ -71,8 +71,7 @@ rom_error_t boot_log_check(const boot_log_t *boot_log) {
   return kErrorBootLogInvalid;
 }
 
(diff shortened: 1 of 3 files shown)


jettr (approved):
I think we should call out the deep sleep resume (yet still a ROM_EXT boot) case a little more explicitly, but that can be another PR.

timothytrippel (approved):

PR #24242, opened 2024-08-07
[topgen] Support for multiple interrupt domains

Razer6:
This PR adds support for multiple interrupt domains to topgen. 

1. The top-level HJSON is added an entry `interrupt_domains`, defining all domains with support for a default one. Instantiating an IP in the HJSON allows you to assign the IP to one of the domains. If no domain is selected, the default interrupt domain is assigned to that IP.
2. The PLIC is ipgen-ed. For every interrupt domain, a dedicated PLIC is generated. The default interrupt domain has no name suiffix (to not change any existing RTL). Others have their suffix. FW collaterals such as IRQ mappings are generated for all domains (and address spaces).
3. The DIF is templated and generates a dedicated DIF for every PLIC.

Note, the first commit is coming from https://github.com/lowRISC/opentitan/pull/23122 that lays out some ground work.


Razer6:
Closing this as this superseeded by the changes from @qmn

PR #24243, opened 2024-08-07
[rv_dm,dv] Remove bogus TODO note from me

rswarbrick:
This demonstrates nicely that I didn't properly understand JTAG when I wrote it. I understand a bit more now and the structure with TDI flowing into a shift register before flowing back out of TDO is literally given in the spec. Oops.


hcallahan-lowrisc (approved):

andreaskurth:
Suggest we include this on the release branch to prevent merge conflicts that could occur later if we don't cherry-pick this

rswarbrick:
Good point. I *think* I've just done the git/github dance correctly to file a PR for that. (Let me know if not!)

PR #24246, opened 2024-08-07
[dv,usbdev] Delayed response to IN token packets in model

alees24:
This PR slightly delays the response to IN token packets within the device model, since the DUT does not respond immediately and there is a risk of CSR traffic introducing a discrepancy between the model and the DUT.

The timed-delay model is very simple at present but may later be extended to support other timed events such as detecting the absence of the host or violations of inter-packet delays.

The lack of temporal modeling was causing failures within the sequences `usbdev_iso_retraction` which heavily exercises CSR traffic and USB collection of Isochronous IN packets.

The PR also addresses the failure of that sequence to tidy up upon termination, which occasionally caused a failure in the interrupt clearing in `cip_base_vseq`. With this PR the sequence `usbdev_iso_retraction` has now been seen to pass successfully for 100 random seeds.

### DIFF
--- a/hw/ip/usbdev/dv/env/usbdev_bfm.sv
+++ b/hw/ip/usbdev/dv/env/usbdev_bfm.sv
@@ -116,6 +116,14 @@ class usbdev_bfm extends uvm_component;
   // within this bank are replaced when actual bytes are received from the USB. (usbdev_usbif.sv)
   bit [31:0] wdata_q;
 
+  // The decision on whether and how to respond to an IN request cannot be actioned imediately
+  // because the CSR state may change in the interim; we introduce a short delay before deciding
+  // how to respond to an IN request.
+  token_pkt in_token;
+
+  // Count of elapsed DUT clock ticks whilst timing (currently just for servicing IN requests).
+  int unsigned tick_cnt = 0;
+
   `uvm_component_new
 
   function void build_phase(uvm_phase phase);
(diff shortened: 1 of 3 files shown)


hw/ip/usbdev/dv/env/usbdev_bfm.sv
```diff
@@ -167,6 +175,30 @@ class usbdev_bfm extends uvm_component;
...
+  endfunction
+
+  // Advance by one clock tick and action any timed response that is now due.
+  function bit send_delayed_response(output usb20_item rsp);
+    bit send_rsp = 1'b0;
+    tick_cnt++;
+    `uvm_info(`gfn, $sformatf("BFM advanced tick count to 0x%0x", tick_cnt), UVM_MEDIUM)
+    if (tick_cnt >= 2 && in_token != null) begin
+      // Decide whether and how to respond to the IN token packet.
+      send_rsp = in_packet(in_token, rsp);
```
hcallahan-lowrisc, line 195: `in_packet()` always returns `1'b1`, so are we actually "deciding whether" to respond as per the comment in the line above? I guess this might change in the future, but maybe just tweak the comment for now?

hw/ip/usbdev/dv/env/usbdev_bfm.sv
```diff
@@ -501,7 +535,14 @@ class usbdev_bfm extends uvm_component;
...
         PidTypeInToken: begin
-          return in_packet(token, rsp);
+          // Basic check of whether we should do anything at all for this transaction.
+          bit [3:0] ep = token.endpoint;
+          if (token.address != dev_address || ep >= NEndpoints || !ep_in_enable[ep]) return 0;
+
+          // Remember the IN token packet and, after a short delay, decide whether and how to
+          // respond.
+          `downcast(in_token, token.clone());
+          tick_cnt = 0;
```
hcallahan-lowrisc, line 545: I think this is the only place `tick_cnt` gets cleared. Does this mean that after `in_token` gets cleared to null, we just keep ticking upwards until the next IN request?

hcallahan-lowrisc (approved):
This makes sense to me, though I'm can't say I understand the protocol details all that well. Two nitty comments, and one which is probably just me mis-understanding the bigger picture. 
Thanks!

hw/ip/usbdev/dv/env/usbdev_bfm.sv
alees24: We stop counting whilst 'in_token' is nul, so we only need to reset `tick_cnt` when `in_token` is assigned a valid handle.

hw/ip/usbdev/dv/env/usbdev_bfm.sv
```diff
@@ -167,6 +175,30 @@ class usbdev_bfm extends uvm_component;
     tx_ep = InvalidEP;
   endfunction
 
+  //------------------------------------------------------------------------------------------------
+  // Timing model.
+  //------------------------------------------------------------------------------------------------
+
+  // Returns 1 iff there is a need to track the passage of time in clock ticks; caller shall invoke
+  // `tick()` once per clock cycle of the DUT whilst this is the case.
```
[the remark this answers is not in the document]
alees24: Thanks. Consequence of hurried restructuring.

hw/ip/usbdev/dv/env/usbdev_bfm.sv
alees24: This function was split into two parts, and we no longer retain details of an IN token packet if no response shall be returned (endpoint disabled, for example). `in_packet()` may therefore be simplified somewhat now, since it's invoked only when the decision has already been made to return a packet.
Now done, thanks.

hw/ip/usbdev/dv/env/usbdev_bfm.sv
hcallahan-lowrisc: Thanks for explaining to me that we stop-ticking due to `bfm.timing()` now becoming false.

hw/ip/usbdev/dv/env/usbdev_bfm.sv
hcallahan-lowrisc: LGTM!

PR #24252, opened 2024-08-08
[rom_ctrl,dv] Use correct type of prediction in scoreboard

rswarbrick:
Commit 8478976 accidentally changed this from UVM_PREDICT_READ to UVM_PREDICT_DIRECT, which causes problems if there are register accesses currently in flight.


alees24 (approved):
Thanks.

PR #24253, opened 2024-08-08
Revert "[dv] Add a quick exit in csr_rd_sub when in reset"

### LINKED ISSUE #24214: [test-triage] Several pwrmgr block-level DV tests fail
### Description

Starting with https://github.com/lowRISC/opentitan/pull/24075, several pwrmgr block-level DV tests started failing. Since the referenced change didn't modify pwrmgr DV, I suspect this is a more general issue. Anyway, this needs to be fixed.

Screenshot from 2024-08-05 10-33-27

@rswarbrick once you're back, can you please investigate this?

rswarbrick:
This reverts commit b400b907d0891418ad4e12e05e6213e68a01391c.

The change there sort of made sense: if there's a reset when we're doing a CSR operation, it's probably more defensible to return with a silly answer than to block indefinitely. Unfortunately, this doesn't quite work with the pwrmgr DV code, which seems to set the under_reset flag well before it causes a reset (and the code doesn't seem to clear it again either).

Thinking harder about the problem, I think the correct behaviour should probably be that the driver/monitor are in charge of responding quickly when they see a reset on the interface in question. Revert the (badly thought out) initial change from me in the meantime.

Fixes #24214.

### DIFF
--- a/hw/dv/sv/csr_utils/csr_utils_pkg.sv
+++ b/hw/dv/sv/csr_utils/csr_utils_pkg.sv
@@ -348,16 +348,6 @@ package csr_utils_pkg;
       status = UVM_IS_OK;
       return;
     end
-
-    // If we are under reset then we want to respond instantly with a bad return value and
-    // UVM_NOT_OK, rather than blocking indefinitely on the bus.
-    if (under_reset) begin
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
Thanks @rswarbrick , this looks good!

PR #24254, opened 2024-08-08
[ci] add GitHub actions private CI

nbdd0121:
This adds GitHub action based private CI for:
* AscentLint (replacing kokoro)
* Xcelium/VCS (replacing Azure pipeline-based private CI)

Both AZP and GitHub CI would still be side-by-side for a while before we decide to turn one off.

See https://github.com/lowRISC/ibex/blob/master/.github/workflows/private-ci.yml for a working example.

### DIFF
--- a/.github/workflows/private-ci.yml
+++ b/.github/workflows/private-ci.yml
@@ -1,0 +1,51 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Private CI
+
+on:
+  push:
+    branches-ignore:
+      - "backport-*"
+    tags:
+      - "*"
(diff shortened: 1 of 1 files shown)


jwnrt:
I pushed this branch to lowRISC/opentitan here to test it:
https://github.com/lowRISC/opentitan/commit/5b1dbf39c2e2044f74fa9fad6c2620ad7d98ae03

jwnrt (approved):
LGTM if the test run passes

jwnrt:
The private CI runs passed and I can see the statuses were correctly reported to this PR

engdoreis (approved):
I'm not a specialist on GitHub actions, but the changes look reasonable to me.

jwnrt:
Merged to get AscentLint CI running again. We now have VCS running twice (GitHub Actions and Azure) for a short trial period. If the GHA version works then we'll disable the Azure version.

PR #24257, opened 2024-08-08
[dv,usbdev] Add coverage exclusions for dynamic USB behavior

alees24:
Manually add explicit coverage exclusions for transitions and branches that cannot be reached because of the dynamic behavior of the USB/DUT, eg. transmission and reception being impossible simultaneously on a shared, bidirectional bus.

### DIFF
--- a/hw/ip/usbdev/dv/cov/usbdev_manual_excl.el
+++ b/hw/ip/usbdev/dv/cov/usbdev_manual_excl.el
@@ -1,0 +1,35 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// The following exclusions have been written and added manually on account of the dynamic behavior
+// and understanding of the USB and USBDEV being unavailable to UNR tooling.
+
+// Since the signal `ev_reset` is asserted only at the end of Reset Signaling, but the entry to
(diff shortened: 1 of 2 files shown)


hw/ip/usbdev/dv/cov/usbdev_unr_excl.el
```diff
@@ -2,6 +2,37 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+// The following exclusions have been added manually on account of the dynamic behavior
+// and understanding of the USB and USBDEV being unavailable to UNR tooling.
```
andreaskurth, line 6: I think we place manual exclusions usually in a different file than UNR-generated ones -- so that when re-running UNR one can simply override the contents of one file (except for the license header). @rswarbrick WDYT?

andreaskurth (commented):
Thanks for composing and describing these exclusions, which make sense to me :+1: Just one comment about the file these manual exclusions should live in

hw/ip/usbdev/dv/cov/usbdev_unr_excl.el
alees24: I would have done that, but was mimicking what's been done in rv_dm

alees24: Separated out, since that's my preference too, and rechecked.

rswarbrick (approved):
This looks really sensible. Thanks for the careful annotations.

Issue #24236, opened 2024-08-06
[rom_ext] Add new boot service message to set both Primary and Next boot slot together


jettr:
### Description

Our owner firmware has run into a situation where it is running from the next slot, but it needs to set the current running slot as primary at the same time as setting the other slot as next. We could do this if there was a single boot service command that could set both primary and next slot within the same request.

PR #24258, opened 2024-08-08
[boot_svc] Consolidate boot_svc boot slot commands

cfrantz:
1. Give the `next_boot_bl0_slot` command the capability to set both the primary BL0 slot and the one-time next boot BL0 slot.
2. Eliminate the now-redundant `primary_bl0_slot` command.
3. Update opentitan{lib,tool} to reflect the boot services command changes.
4. Update the tests.

Fixes: #24236


sw/device/silicon_creator/lib/boot_svc/boot_svc_next_boot_bl0_slot.h
```diff
@@ -35,12 +35,18 @@ typedef struct boot_svc_next_boot_bl0_slot_req {
    * BL0 slot to boot on next reboot.
    */
   uint32_t next_bl0_slot;
+  /**
+   * BL0 slot to set as the primaryh boot slot.
```
timothytrippel, line 39:
```suggestion
   * BL0 slot to set as the primary boot slot.
```
?

timothytrippel (approved):

sw/device/silicon_creator/lib/boot_svc/boot_svc_next_boot_bl0_slot.h
cfrantz: Fixed.

sw/device/silicon_creator/lib/boot_svc/boot_svc_next_boot_bl0_slot.h
```diff
@@ -35,12 +35,18 @@ typedef struct boot_svc_next_boot_bl0_slot_req {
    * BL0 slot to boot on next reboot.
    */
   uint32_t next_bl0_slot;
+  /**
```
jettr, line 38: Can we change the kBootSvcNextBl0SlotReqType = 0x5458454e, to another value?

I am for dropping the two older versions of the commands, but maybe we should do it over a release. I can see this causing problems for the Owner firmware updates when both ROM_EXT and owner firmware change. The old owner firmware will call the old set next slot, and reboot. Then the new ROM_EXT will update but won't recognize the command.

Thoughts on a staggered rollout for the removal?

cfrantz: Normally, yes, but: the ROM_EXT is currently an unreleased product without a guarantee of a stable API and targeted at an engineering-sample chip.

You can distinguish between the old API and the new one by examining the ROM_EXT version in the boot_log (current API: rom_ext <= 0.3).  If it would be more convenient, I can update the ret-ram version identifier from `RR04` to `RR05`, but since this isn't a layout change, that's perhaps a bit of a misuse of that identifier.

jettr: We can definitely muddle through without a phased roll out since we are in a pre-released phase. We have the ability to use boot strap to get out of a bad state.

As I am thinking of this more: once we start officially releasing ROM_EXT. I don't think we can ever really change this command format -- even if we bump the version. The reason is, old owner firmware that doesn't know about the new ROM_EXT firmware, and the older owner FW is going to be the one sending this message to newer ROM_EXT during a typically firmware update that upgrades both ROM_EXT and Owner firmware.

We definitely need to think about best practice and recommendation since I would image the typical flow for updating owner firmware and ROM_EXT at the same time would have the ROM_EXT update first, but the older owner firmware issues a "legacy" boot service request.

PR #24259, opened 2024-08-08
[dv,usbdev] Ensure all endpoint types see PRE.

alees24:
The PRE special PID (used to shield Full Speed devices from Low Speed traffic) shall be seen by all endpoint types, and it shall produce no response/change in any.

Modify the existing sequences to ensure that all types of transactions may be sent as Low Speed traffic (with PREamble) to all types of endpoints, increasing testing and satisfying coverage points.

Low speed traffic sequence now derives directly from the max_usb_traffic sequence to ensure that Isochronous endpoints are also subjected to LS traffic.

There was also a failure to collect coverage for `data_tog_endp_cg` on handshake packets.


andreaskurth (approved):
Thanks for these clearly described enhancements, @alees24! LGTM

hcallahan-lowrisc (approved):

PR #24262, opened 2024-08-09
[dv,usbdev] Transition USBDEV to V2S

alees24:


hw/ip/usbdev/data/usbdev.hjson
```diff
@@ -21,7 +21,8 @@
   version:            "2.0.0",
   life_stage:         "L1",
   design_stage:       "D2S",
-  verification_stage: "V1",
+  verification_stage: "V2S",
+  commit_id:          "3f85737335e12afcff4e37d163c2849ab2e49846"
```
andreaskurth, line 25:
```suggestion
```
We no longer use the `commit_id` field, instead we would obtain the Git commit hash at signoff via tooling. Could you please remove that?

alees24: "When in doubt look at the UART." Does this want to be changed? -> https://github.com/lowRISC/opentitan/blob/master/hw/ip/uart/data/uart.hjson

andreaskurth: Yes, at some point :-)

andreaskurth (approved):

andreaskurth:
CHANGE AUTHORIZED: hw/ip/usbdev/data/usbdev.hjson

PR #24263, opened 2024-08-09
[dv,jtag] Teach jtag_dmi_reg_frontdoor to handle resets

rswarbrick:
This ensures that a DMI operation that is in flight will be aborted (rather than hanging) if a reset gets asserted.

### DIFF
--- a/hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
+++ b/hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
@@ -22,9 +22,8 @@ class jtag_dmi_reg_frontdoor extends uvm_reg_frontdoor;
 
   virtual task body();
     csr_field_t         csr_or_fld;
-    uvm_reg_data_t      wdata = 0, rdata;
+    uvm_reg_data_t      wdata = 0;
     jtag_dtm_reg_block  jtag_dtm_ral = jtag_agent_cfg_h.jtag_dtm_ral;
-    jtag_dmi_op_rsp_e   op_rsp;
 
     // Configure the JTAG agent to have a positive run-to-clear length, ensuring that DMI operations
     // make it from dmi_jtag to dm_top and back again before TCK stops.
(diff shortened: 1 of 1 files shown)


hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
```diff
@@ -75,41 +74,102 @@ class jtag_dmi_reg_frontdoor extends uvm_reg_frontdoor;
...
-            rw_info.value = new[1];
-            rw_info.value[0] = rdata;
-          end
+    // soon, it may end up setting the in progress sticky bit, causing more time wasted. If there is
+    // a reset in the middle of the wait, stop immediately.
+    fork begin
+      fork
+        jtag_agent_cfg_h.vif.wait_tck(10);
+        wait(jtag_agent_cfg_h.in_reset);
+      join_any
```
marnovandermaas, line 90: Why is this a join any? Since we only have one forked thread, can't this just be a join?

hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
```diff
@@ -75,41 +74,102 @@ class jtag_dmi_reg_frontdoor extends uvm_reg_frontdoor;
...
+      disable fork;
```
marnovandermaas, line 91: Excuse my ignorance but if there is just one fork and you have a join any then at this point all the forks are joined already right?

hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
```diff
@@ -75,41 +74,102 @@ class jtag_dmi_reg_frontdoor extends uvm_reg_frontdoor;
...
-            rdata = get_field_val(jtag_dtm_ral.dmi.data, rdata);
-            if (csr_or_fld.field != null) begin
-              rdata = get_field_val(csr_or_fld.field, rdata);
-            end
-            rw_info.value = new[1];
-            rw_info.value[0] = rdata;
-          end
+    // soon, it may end up setting the in progress sticky bit, causing more time wasted. If there is
+    // a reset in the middle of the wait, stop immediately.
+    fork begin
```
marnovandermaas, line 86: Maybe label this as an isolation fork?

hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
```diff
@@ -75,41 +74,102 @@ class jtag_dmi_reg_frontdoor extends uvm_reg_frontdoor;
...
+
+      // Otherwise, the operation was still in progress. Clear the (sticky) flag.
+      csr_wr(.ptr(jtag_dtm_ral.dtmcs.dmireset), .value(1), .blocking(1), .predict(1));
+
+      // Check there hasn't been a JTAG reset while we were clearing the "op in progress" flag. If
+      // not, we'll go around again (and read the DMI register again)
+      if (jtag_agent_cfg_h.in_reset) begin
+        rw_info.status = UVM_NOT_OK;
+        return;
+      end
```
marnovandermaas, line 171: Is there a reason why you decided to rewrite the if statement from the original code? I think I slightly prefer the syntax:
```
if
  ...
else
  ...
```

Over:
```
if
  ...
  return

...
```

marnovandermaas: Also if you do keep your construction, I tend to prefer if statements that use `==` rather than `!=` so I would switch the clauses.

marnovandermaas (commented):
A few comments from my end based on a code inspection.

hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
marnovandermaas: Never mind the two statements are the forked threads.

hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
marnovandermaas: Same reason as above.

marnovandermaas (approved):
Generally looks good to me based on a code review. Please do make sure the CI failures are unrelated to this PR before merging.

hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
rswarbrick: I think I did that at first but then dropped the name because it clashes with the next fork! But I guess it still makes sense to put a name on it. I'll be lame and call it "iso_fork" and rely on the fact that this is a different name from "isolation_fork"...

hw/dv/sv/jtag_dmi_agent/jtag_dmi_reg_frontdoor.sv
rswarbrick: Hmm. The logic is supposed to be something like:
```
operation_1;
if (that_failed) bail;
operation_2;
if (!that_succeeded) bail;
operation_3;
...
```

Chaining if/else statements doesn't really work (because e.g. operation_3 ends up indented twice). The `!=` bit is because there are two possible "done" values and I want to match either of them.

PR #24264, opened 2024-08-09
[dv] Exit quickly from check_no_fatal_alerts on reset

rswarbrick:
If we are under reset, there's no need to check whether alerts are in flight. Return from the task quickly (which avoids us spending ages in post_start).


eshapira (approved):
LGTM, thanks!

PR #24265, opened 2024-08-09
[rv_dm,dv] Make lots of changes to vseqs to allow reset during a run

rswarbrick:
This is tied in with changes I'm making locally to allow a reset to happen more easily in the middle of a run (for the stress_all_with_rand_reset testpoint). When the base vseq is configured to allow these resets to occur, the changes in this PR are needed to avoid things falling apart.

This PR has quite a few commits (sorry!). All of them are new except for the first one, which is from #24138 (which I hadn't been chasing reviews for: oops).

### DIFF
--- a/hw/ip/rv_dm/dv/env/rv_dm_scoreboard.sv
+++ b/hw/ip/rv_dm/dv/env/rv_dm_scoreboard.sv
@@ -400,6 +400,7 @@ class rv_dm_scoreboard extends cip_base_scoreboard #(
     tl_sba_a_chan_fifo.flush();
     tl_sba_d_chan_fifo.flush();
     selected_dtm_csr = cfg.m_jtag_agent_cfg.jtag_dtm_ral.idcode;
+    cfg.jtag_dmi_ral.reset(kind);
   endfunction
 
   function void check_phase(uvm_phase phase);
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -177,9 +177,18 @@ class rv_dm_base_vseq extends cip_base_vseq #(
       // takes effect. Fortunately, we can see that this has happened by looking at the trst_n
       // signal: it will go high once everything has been connected. *That* signal is exposed
       // through jtag_mon_if in the tb, which is visible through the jtag agent's mon_vif interface.
-      wait(cfg.m_jtag_agent_cfg.mon_vif.trst_n);
+      // Exit early if a system reset appears in the meantime.
+      fork begin : isolation_fork
+        fork
+          wait(cfg.m_jtag_agent_cfg.mon_vif.trst_n);
+          wait(!cfg.clk_rst_vif.rst_n);
+        join_any
+        disable fork;
+      end join
+      if (!cfg.clk_rst_vif.rst_n) return;
 
-      // "Activate" the DM to facilitate ease of testing.
+      // "Activate" the DM to facilitate ease of testing. This will exit early if there is a JTAG
+      // reset.
       csr_wr(.ptr(jtag_dmi_ral.dmcontrol.dmactive), .value(1), .blocking(1), .predict(1));
     end
 
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -201,8 +201,6 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     cfg.m_jtag_agent_cfg.vif.set_tck_period_ps(tck_period_ps);
     fork
       if (kind inside {"HARD", "TRST"}) begin
-        jtag_dtm_ral.reset("HARD");
-        jtag_dmi_ral.reset("HARD");
         cfg.m_jtag_agent_cfg.vif.do_trst_n();
       end
       if (kind inside {"HARD", "SCAN"}) apply_scan_reset();
--- a/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
+++ b/hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
@@ -260,9 +260,14 @@ class rv_dm_base_vseq extends cip_base_vseq #(
   endtask
 
   // Stop running the m_tl_sba_device_seq seq.
+  //
+  // This is a no-op if the sequence is actually null (because we never completed dut_init, which
+  // would have constructed the object),
   virtual task sba_tl_device_seq_stop();
-    m_tl_sba_device_seq.seq_stop();
-    `uvm_info(`gfn, "Stopped running m_tl_sba_device_seq", UVM_MEDIUM)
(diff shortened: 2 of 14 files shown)


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
```diff
@@ -237,7 +237,8 @@ class rv_dm_base_vseq extends cip_base_vseq #(
     sba_tl_device_seq_stop();
     // Check for pending rv_dm operations and wait for them to complete.
     // TODO: Improve this later.
```
marnovandermaas, line 237: Can we remove this TODO now?

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_mem_tl_access_halted_vseq.sv
```diff
@@ -18,11 +18,16 @@ class rv_dm_mem_tl_access_halted_vseq extends rv_dm_base_vseq;
...
     csr_rd(.ptr(jtag_dmi_ral.dmstatus), .value(r_data));
-    `DV_CHECK_EQ(1, get_field_val(jtag_dmi_ral.dmstatus.anyhalted, r_data))
-    `DV_CHECK_EQ(1, get_field_val(jtag_dmi_ral.dmstatus.allhalted, r_data))
+
+    // Assuming that we haven't entered reset (in which case r_data could be anything), check that
+    // anyhalted and allhalted are indeed set.
+    if (cfg.clk_rst_vif.rst_n) begin
+      `DV_CHECK_EQ(1, get_field_val(jtag_dmi_ral.dmstatus.anyhalted, r_data))
+      `DV_CHECK_EQ(1, get_field_val(jtag_dmi_ral.dmstatus.allhalted, r_data))
+    end
```
marnovandermaas, line 30: I'm not sure how often this occurs but maybe a good idea to put a UVM info in the else statement here in case it occurs when you're not running in the stress all mode.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
```diff
@@ -50,13 +50,20 @@ class rv_dm_ndmreset_req_vseq extends rv_dm_base_vseq;
...
+        wait (!cfg.clk_rst_vif.rst_n);
+        begin
+          cfg.clk_lc_rst_vif.drive_rst_pin(1'b0);
+          cfg.clk_rst_vif.wait_clks(8);
+          cfg.clk_lc_rst_vif.drive_rst_pin(1'b1);
+          cfg.clk_rst_vif.wait_clks(8);
+        end
+      join_any
+      disable fork;
+    end join
```
marnovandermaas, line 66: Should we still do the ndmreset pending check if reset is hit?

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
```diff
@@ -82,8 +89,10 @@ class rv_dm_ndmreset_req_vseq extends rv_dm_base_vseq;
     // field in dmcontrol.
     csr_wr(.ptr(jtag_dmi_ral.dmcontrol.ndmreset), .value(1));
 
-    // Check that the ndmreset request appears at top-level. Note that we don't assert that it
-    // remains high: it will drop again when we disable lc_hw_debug_en below.
+    // If we are in reset, stop immediately. Otherwise, check that the ndmreset request appears at
+    // top-level. Note that we don't assert that it remains high: it will drop again when we disable
+    // lc_hw_debug_en below.
+    if (!cfg.clk_rst_vif.rst_n) return;
```
marnovandermaas, line 95: Should we do a UVM info in this case?

hw/ip/rv_dm/dv/env/rv_dm_scoreboard.sv
```diff
@@ -393,6 +393,7 @@ class rv_dm_scoreboard extends cip_base_scoreboard #(
     tl_sba_a_chan_fifo.flush();
     tl_sba_d_chan_fifo.flush();
     selected_dtm_csr = cfg.m_jtag_agent_cfg.jtag_dtm_ral.idcode;
+    cfg.jtag_dmi_ral.reset("HARD");
```
marnovandermaas, line 396: Shouldn't this be `reset(kind)`?

marnovandermaas (commented):
From an initial code inspection this looks good. I'm not sure how much we gain from having lots of separate commits that just add if-reset-then-return statements. Maybe it's better to squash those ones?

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
rswarbrick: Hmm, I think so: yes! I'll drop it and the wait in a separate PR.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_mem_tl_access_halted_vseq.sv
rswarbrick: This is only going to happen in the stress-all-with-rand-reset mode and there are equivalent checks dotted all over the code. I don't think UVM info lines in each case would really make much sense. What do you think?

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
rswarbrick: Hmm, good point: it doesn't make much sense. Looking at `check_ndmreset_pending`, it won't matter (the `get_ndmreset_pending` task will complete in zero time and then we'll bail). But it's not very clear: I'll put an explicit return here.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_ndmreset_req_vseq.sv
rswarbrick: Like with many other places where I'm doing this, it's saying something like "if the design is in reset, stop the test". I don't think that we need to add a print in each case: what do you think?

hw/ip/rv_dm/dv/env/rv_dm_scoreboard.sv
rswarbrick: Good point, thanks.

rswarbrick:
About the millions of separate commits: it was a consequence of me doing stuff in bits (fixing a vseq at a time). I hadn't realised quite how ridiculous things had got! I've squashed the relevant commits together now.

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_base_vseq.sv
rswarbrick: Done! Now visible as #24282.

marnovandermaas: thanks

marnovandermaas (approved):
Looks reasonable to me. Just one comment on a change that might be good to have a separate commit.

PR #24266, opened 2024-08-09
[dv] Define a more robust "resetting procedure" for vseqs

rswarbrick:
This is intentionally opt-in and won't have any effect unless an IP block configures can_reset_with_csr_accesses=1.

### DIFF
--- a/hw/dv/sv/cip_lib/cip_base_env_cfg.sv
+++ b/hw/dv/sv/cip_lib/cip_base_env_cfg.sv
@@ -8,6 +8,15 @@ class cip_base_env_cfg #(type RAL_T = dv_base_reg_block) extends dv_base_env_cfg
   bit en_tl_err_cov      = 1;
   bit en_tl_intg_err_cov = 1;
 
+  // If this flag is set then we allow the stress_all_with_rand_reset task to apply a reset without
+  // waiting for CSR accesses to complete. This will only work if the IP block's vseqs
+  //
+  //  - Check for reset after performing a CSR read, skipping any check on the result if we are in
+  //  reset (because the result will be rubbish if we are in reset)
+  //
+  //  - Check for reset every so often (to run quickly to completion if a reset has been applied).
+  bit can_reset_with_csr_accesses = 1'b0;
+
   // Downstream agent cfg objects.
 
   // If the block supports only one RAL, just use `m_tl_agent_cfg`.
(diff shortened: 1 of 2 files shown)


hw/dv/sv/cip_lib/cip_base_env_cfg.sv
```diff
@@ -8,6 +8,12 @@ class cip_base_env_cfg #(type RAL_T = dv_base_reg_block) extends dv_base_env_cfg
   bit en_tl_err_cov      = 1;
   bit en_tl_intg_err_cov = 1;
 
+  // If this flag is set then we allow the stress_all_with_rand_reset task to apply a reset without
+  // waiting for CSR accesses to complete. This will only work if the IP block's vseqs check the
+  // current reset state after performing a CSR read (because the result will be rubbish if we are
+  // in reset) and every so often (to run quickly to completion if a reset has been applied).
```
hcallahan-lowrisc, line 14: nit: The second sentence is a little unclear to me (if you remove the content in parens it doesn't read that well). Would you mind rewording it? Thanks

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -763,6 +767,18 @@ class cip_base_vseq #(
               end
             end
           join_any
+
+          // If vseq_done is false then we have issued a reset (the second process in the fork) but
+          // the vseq that were racing against hasn't noticed the reset and stopped. Killing that
```
hcallahan-lowrisc, line 772: nit:
```suggestion
          // the vseq that we're racing against hasn't noticed the reset and stopped. Killing that
```

hcallahan-lowrisc (approved):
The idea here makes sense to me. I'm happy for it to be merged.
I think some documentation for this feature would be warranted too, maybe in the [Comportable IP Testbench Architecture](https://opentitan.org/book/hw/dv/sv/cip_lib/index.html#comportable-ip-testbench-architecture) page? I'm happy for that to be in a follow-up PR.

rswarbrick:
Thanks for the review. Good point about the potential documentation, but I think it probably makes more sense to document it once we've got at least one block that uses this properly :-D (I'm a little concerned that we'll put effort into documenting something and then implement something else...)

rswarbrick:
Documentation in #24301.

PR #24267, opened 2024-08-09
[hw,dma] Add a chunk_done interrupt

Razer6:
This PR adds a new chunk_done interrupt. This interrupt is raised for chunked mem2mem transfers when a single chunk is transferred and the SW needs to start the transfer of a new chunk. When transferring the last chunk, only the `done` interrupt is raised.

This PR is based upon https://github.com/lowRISC/opentitan/pull/24108

### DIFF
--- a/hw/ip/dma/data/dma.hjson
+++ b/hw/ip/dma/data/dma.hjson
@@ -102,6 +102,10 @@
       desc: "DMA operation has been completed."
       type: "status"
     }
+    { name: "dma_chunk_done"
+      desc: "Indicates the transfer of a single chunk has been completed."
+      type: "status"
+    }
     { name: "dma_error"
       desc: "DMA error has occurred. DMA_STATUS.error_code register shows the details."
       type: "status"
--- a/hw/ip/dma/doc/registers.md
+++ b/hw/ip/dma/doc/registers.md
@@ -71,55 +71,58 @@
 Interrupt State Register
 - Offset: `0x0`
 - Reset default: `0x0`
-- Reset mask: `0x3`
+- Reset mask: `0x7`
 
 ### Fields
 
 ```wavejson
-{"reg": [{"name": "dma_done", "bits": 1, "attr": ["ro"], "rotate": -90}, {"name": "dma_error", "bits": 1, "attr": ["ro"], "rotate": -90}, {"bits": 30}], "config": {"lanes": 1, "fontsize": 10, "vspace": 110}}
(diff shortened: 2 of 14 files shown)


alees24 (changes requested):
I think the logic here is correct, thank you, but please do heed my remarks about tightening the interrupt behavior so that it properly conforms to the Comportable IP requirements, otherwise the common, shared tests will fail for this block.

Interrupts are described [here](https://opentitan.org/book/doc/contributing/hw/comportability/index.html?highlight=interrupt#interrupt-handling).

Presently the tests will pass because the implementing intr_state/enable/test-related logic assumes that they are all Event type interrupts, and nothing is causing a Status-type (dma_done or dma_error) interrupt to be asserted by the device within those tests.

It is not clear whether the intention was for them to be Event type and rely upon the `status` register for the history, or if they're supposed to be Status type and the `intr_state.de` _should_ be driven by the `status.q` bit as presently.

Razer6:
@alees24 After the chunk_done status PR was merged, this one adds a status-type interrupt to the chunk_done. The PR intentionally uses interrupt index 1, where previously the memory limit IRQ was to not shuffle around the other IRQ.

andreaskurth (approved):
RTL changes LGTM; I'll defer to others for feedback on the DV changes. Thx @Razer6!

alees24 (changes requested):
There's a missing `chunk_done` assertion in the DmaShaWait state that leads to a premature assertion of `done`:
```
@@ -1005,6 +1005,7 @@ module dma
               // In non-hardware handshake mode, finishing one chunk should raise the done IRQ
               // and done bit, and release the go bit for the next FW-controlled chunk.
               clear_go     = !control_q.cfg_handshake_en;
+              chunk_done   = !control_q.cfg_handshake_en;
               ctrl_state_d = DmaIdle;
             end else begin
               ctrl_state_d = DmaAddrSetup;
```
There are more issues on the DV side, which has been somewhat lagging behind the RTL changes lately, since this block is unfortunately not yet running in regressions/CI. I'll supply the DV updates shortly.

Thanks.

Razer6:
Thank's for spotting this. I just added this change to the commit.

alees24 (approved):
I think this is good to merge now, thank you. Locally I'm just seeing the range_regwen_rd_A assertion failure on dma_stress_all and dma_memory_region_lock sequences, because of the missing changes in the autogen FPV register checking code.

alees24:
Merging because the salient CI tests have completed; DMA not present in Earl Grey.

PR #24268, opened 2024-08-09
[hw,dma,rtl] Flexible address increment options

Razer6:
This PR adds support for a flexible address increment mode. For the source and destination address, the DMA can be configured to perform:

* No address increment (FIXED). The DMA always reads/writes from the same address.
* Wrapped increment (WRAP). The DMA increments, and when finishing, one chunk wraps back to the starting address.
* Linear increment (LINEAR). The DMA always increments the address

These modes are independent of the hardware handshake mode. These changes would allow the DMA to be used for a memset operation, where a chunked source pattern is replicated in the destination memory range. 

This PR currently only does the RTL change without the DV updates.


hw/ip/dma/data/dma.hjson
```diff
@@ -464,35 +464,49 @@
...
           desc: '''
-                Used in conjunction with the hardware handshake enable.
-                0: Receive data from LSIO FIFO to memory buffer.
-                1: Send data from memory buffer to LSIO FIFO.
+                Defines the address increment behavior for the destination address.
                 '''
+          enum: [
+            { value: "0",
+              name: "FIX"
+              desc: "Do not increment the address, always write to the same address."
```
alees24, line 497: Small point but I think mode 0 should be named 'FIXED' rather than 'FIX'

hw/ip/dma/data/dma.hjson
```diff
@@ -464,35 +464,49 @@
...
-                Used in conjunction with the hardware handshake mode of operation.
-                Auto Increments the memory buffer address register by data size to point to the next memory buffer address.
-                Generate a warning (assert interrupt) if the auto-incremented address reaches close to the value set in limit address register to prevent destination buffer overflow.
-                Enables firmware to take appropriate action prior to reaching the limit.
-                '''
-        }
-        { bits: "6"
-          name: "fifo_auto_increment_enable"
+        { bits: "6:5"
+          name: "src_addr_increment"
```
alees24, line 468: These 'address increment' values sound like the numerical offset to be added to the src/dst address. Could they perhaps be renamed to 'src|dst_addr_mode' please? I think that's clearer.

hw/ip/dma/rtl/dma.sv
```diff
@@ -1130,37 +1125,12 @@ module dma
     // since multi-chunked transfers roundtrip via IDLE.
     hw2reg.cfg_regwen.d = prim_mubi_pkg::mubi4_bool_to_mubi(~reg2hw.status.busy.q);
 
-    // If we are in hardware handshake mode with auto-increment increment the corresponding address
-    // when finishing a DMA operation when transitioning from a data move state to the idle state
```
alees24, line 1134: Having removed the address update logic the 'LINEAR' address incrementing mode will not work as expected in the hardware handshaking mode.

If my memory/understanding are correct it can still be employed in the non-handshaking mode but only by relying upon firmware to provide the updated addresses.

alees24 (changes requested):
Thanks for this. Let me first say I strongly approve of the change to the register API because I've never liked the previous 3 configuration bits where the memory/fifo bits changed significance upon the confusing/poorly-named 'data_direction' bit.
The new API also permits some additional functionality for which I tried to advocate when we were reviewing the original design, making the controller more useful with 'chunked' operation.

There are a couple of faults with the current proposed implementation and we shall obviously need at least to introduce changes in the DV to map its present behavior onto the new programming API, even if we have to impose some additional constraints until the DV can be fully-updated.

Is any of the DV work in-progress or planned presently?

As per my review of the 'memset' PR, has consideration been given to whether introducing register changes that are less/not backwards-compatible is a concern?

Razer6:
@alees24 I re-designed the flexible addressing by introducing two different control registers for source and destination. Both implement a bit `increment` and `wrap` to implement the functionality described in the PR description. This leaves room for future extensions for src and destination and also keeps the configuration symmetry. 

Could you take another look into this one?

hw/ip/dma/rtl/dma.sv
Razer6: This implemented now differently. When doing a linear address increment, we need to update the register addresses when being in multi-chunk mode.

alees24 (approved):
Okay, I think this logic now does what's wanted, thank you.
Just a couple of errors in the comments.

hw/ip/dma/rtl/dma.sv
```diff
@@ -1170,27 +1157,29 @@ module dma
...
+    // When we would update the register, we would update it with the current transferred number of
+    // bytes of the current chunk
+    new_dst_addr = {reg2hw.dst_addr_hi.q, reg2hw.dst_addr_lo.q} +
+                    SYS_ADDR_WIDTH'(reg2hw.chunk_data_size.q);
+    new_src_addr = {reg2hw.src_addr_hi.q, reg2hw.src_addr_lo.q} +
+                    SYS_ADDR_WIDTH'(chunk_byte_q);
```
alees24, line 1165: Apologies that I did not notice the use of 'chunk_byte_q' here sooner....this may not have been updated at the point that 'new_src_addr' is used, and I don't see any reason for the disparity between the dst/src address updating anyway.
This was spotted in nascent DV changes; I'll contact you with a branch that you can use.

alees24 (changes requested):
Please see the comment about 'new_src_addr'

Also, whilst we're still finalizing this, could you perhaps change 'SRC|DST_CONTROL' to 'SRC|DST_CONFIG', please? I think these names are better because a control register typically causes actions to occur, plus additional configuration such as stride values may meaningfully be added to the registers later.

Lastly, as you may have seen, the linting is complaining that the CMDGEN documentation is out of date. (Run ` ./util/cmdgen.py -u '**/*.md' ` from the project root.)

andreaskurth (approved):

PR #24271, opened 2024-08-12
[rv_dm,dv] Fix constant width in vseq

rswarbrick:
This silences a build warning from Xcelium (because $urandom_range expects a (32 bit) int argument.


marnovandermaas (approved):
Looks reasonable to me.

PR #24275, opened 2024-08-12
[tl,dv] Remove d_error check from bus2reg in reg adapter

rswarbrick:
This was sitting in rather the wrong place: bus2reg shouldn't really be doing scoreboard checks. It was also failing for some rv_dm vseqs, which sent TL transactions when debug is disabled, which means they all get error responses.

This commit also removes the line that turns off this check in rv_dm_debug_disabled_vseq (because the check isn't there any more to be turned off)


hcallahan-lowrisc (approved):
LGTM, the reg_adapter definitely shouldn't be doing any checking over there...

PR #24279, opened 2024-08-12
[kmac/dv] Reduce test time spent on testing NIST test vectors

vogelpi:
Previously, these tests were testing up to three complete test vector sets per test run, which could take several hours per run. And for each of these tests, we were running 50 seeds in the nightlies per default.

As these tests don't contribute much to the test coverage, this commit reduces the number of seeds per test. In addition, only one test vector set is tested per test run to reduce the time spent per run.

This is a follow-up of lowRISC/OpenTitan#24233. This resolves lowRISC/OpenTitan#24225.

### DIFF
--- a/hw/ip/kmac/dv/env/seq_lib/kmac_test_vectors_base_vseq.sv
+++ b/hw/ip/kmac/dv/env/seq_lib/kmac_test_vectors_base_vseq.sv
@@ -18,6 +18,11 @@ class kmac_test_vectors_base_vseq extends kmac_smoke_vseq;
   task body();
     test_vectors_pkg::test_vectors_t vectors[];
 
+    // Randomly pick a single test vector set to limit test run time.
+    int test_idx = $urandom_range(0, test_list.size - 1);
+    test_list = {test_list[test_idx]};
+    `uvm_info(`gfn, $sformatf("test_idx = %0d", test_idx), UVM_MEDIUM)
+
     foreach (test_list[i]) begin
       // parse each test vector file
       test_vectors_pkg::get_hash_test_vectors(test_list[i], vectors);
--- a/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
+++ b/hw/ip/kmac/dv/kmac_base_sim_cfg.hjson
@@ -114,6 +114,7 @@
                  "+test_timeout_ns=5_000_000_000",
                  "+test_vectors_sha3_variant=224"]
       run_timeout_mins: 90
+      reseed: 5
     }
     {
       name: "{name}_test_vectors_sha3_256"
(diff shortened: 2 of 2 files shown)


hw/ip/kmac/dv/env/seq_lib/kmac_test_vectors_base_vseq.sv
```diff
@@ -18,6 +18,11 @@ class kmac_test_vectors_base_vseq extends kmac_smoke_vseq;
   task body();
     test_vectors_pkg::test_vectors_t vectors[];
 
+    // Randomly pick a single test vector set to limit test run time.
+    int test_idx = $urandom_range(0, test_list.size - 1);
+    test_list = {test_list[test_idx]};
```
rswarbrick, line 23: Oof, that's pretty horrible! :-)

I guess it avoids touching more of the rest of the code though: very clever!

rswarbrick (approved):
This looks very sensible. Can we also drop the timeout limits for the sequences? (If so, it would definitely make sense in a following commit/PR)

hw/ip/kmac/dv/env/seq_lib/kmac_test_vectors_base_vseq.sv
vogelpi: Yes, that was the intention: don't touch the rest of the vseq :-)

hw/ip/kmac/dv/env/seq_lib/kmac_test_vectors_base_vseq.sv
```diff
@@ -18,6 +18,11 @@ class kmac_test_vectors_base_vseq extends kmac_smoke_vseq;
...
+    `uvm_info(`gfn, $sformatf("test_idx = %0d", test_idx), UVM_MEDIUM)
+
     foreach (test_list[i]) begin
```
martin-velay, line 26: I guess this `foreach` loop is not needed anymore as only one `test_list` is retained

vogelpi:
Thanks for your review @rswarbrick , good idea re dropping the timeouts for the sequences. I will adapt the values once we see how long the tests really take in the next nightly regression.

martin-velay (approved):
I have one suggestion, but LGTM. I'll do a similar stuff for HMAC

hw/ip/kmac/dv/env/seq_lib/kmac_test_vectors_base_vseq.sv
vogelpi: Yes, that's right. It's pretty ugly but also super simple as I didn't have to change the rest of the sequence. Ideally, we would generate list of individual vectors from different sets and play them in a random order but all this would be more work and I don't have the time right now :-/

PR #24280, opened 2024-08-12
[rv_dm,dv] Add sparse FSM tests

rswarbrick:


hw/ip/rv_dm/dv/env/seq_lib/rv_dm_sparse_lc_gate_fsm_vseq.sv
```diff
@@ -0,0 +1,63 @@
...
+    `DV_CHECK(uvm_hdl_deposit(path, good_value ^ mask))
+
+    // Wait a cycle to allow the FSM flop to catch up
+    cfg.clk_rst_vif.wait_clks(1);
+
+    // Re-enable the FSM state assertions which would have failed when the state FSM and flop didn't
+    // match
+    $asserton(0, "tb.dut.u_tlul_lc_gate_sba.u_state_regs_A");
+    $asserton(0, "tb.dut.u_tlul_lc_gate_rom.u_state_regs_A");
+  endtask
```
vogelpi, line 51: Should we add a random delay here with a check that the FSM doesn't enter again any of the valid states, i.e., to ensure the default state is terminal?

vogelpi (approved):
LGTM with one question, thanks @rswarbrick !

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_sparse_lc_gate_fsm_vseq.sv
rswarbrick: Good point. I've just added a trailing "and stay there for 100 cycles" check.

PR #24282, opened 2024-08-12
[rv_dm,dv] Drop a wait from dut_shutdown

rswarbrick:
The first commit is from #24265, which should be merged first. The second commit's message is below.

I'm splitting the commits like this because I *think* the combination is correct (and needed), but the first commit is *definitely* safe. Splitting like this means that things are a bit easier to debug if something goes wrong.

**[rv_dm,dv] Drop a wait from dut_shutdown**

I think the logic behind this was that a vseq might have started some operations and we want to wait for them to run to completion. I don't believe that there *are* any such operations in rv_dm's sequences and this just made the code more complicated. Drop it.


hcallahan-lowrisc (approved):
The change makes sense to me. And CI agrees. Thanks!
Nit. needs a rebase

PR #24283, opened 2024-08-12
[rv_dm,dv] Mark sec_cm_exec_ctrl_mubi testpoint as being covered

rswarbrick:
It turns out that this testpoint is actually covered already by rv_dm_buffered_enable. Expand the description and mark it accordingly.


vogelpi (approved):
LGTM!

PR #24286, opened 2024-08-12
[dv,pwrmgr] Remove remaining TODOs

matutem:
Also mark ALL_TODOS_RESOLVED and a few others in the V3 checklist:
- X_PROP_ANALYSIS_COMPLETED: no issues shown in compilation xprop.log
- FPV_ASSERTIONS_PROVEN_AT_V3, FPV_CODE_COVERAGE_AT_100, and
  FPV_COI_COVERAGE_AT_100 do not apply to pwrmgr
- TB_LINT_COMPLETE: no issues shown in dashboards

### DIFF
--- a/hw/ip_templates/pwrmgr/data/pwrmgr_testplan.hjson
+++ b/hw/ip_templates/pwrmgr/data/pwrmgr_testplan.hjson
@@ -3,15 +3,12 @@
 // SPDX-License-Identifier: Apache-2.0
 {
   name: "pwrmgr"
-  // TODO: remove the common testplans if not applicable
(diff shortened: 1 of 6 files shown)


rswarbrick (approved):
I think I agree with these changes: thanks. For next time, it would be brilliant if you could split things into separate commits (possibly in the same PR). For example, there's no particular reason that an update to `checklist.md` should be atomic with a bugfix vseq change.

matutem:
@rswarbrick you are right about splitting commits. I guess the TODO I deleted from pwrmgr_escalation_timeout_vseq triggered me to update the checklist.

This should be cherry-picked to the ES prod branch since it impacts signoffs.

PR #24288, opened 2024-08-13
[dv] Manually apply specialisation for mubi32_cov

rswarbrick:
I strongly suspect that this is working around a VCS tool bug.


hcallahan-lowrisc (approved):
As discussed in person, this makes sense to me as a temporary fix. Though I'm still baffled as to why this seemingly-simple construct hits a bug!

PR #24293, opened 2024-08-13
[pentest] Update and add new OTBN FI tests

nasahlpa:
During the penetration testing, new OTBN FI tests and existing ones have been improved. As this work happened in a fork, this PR brings these changes upstream. The first commit two commits update the existing tests and the last one adds new tests.


johannheyszl (approved):
We had reviewed the tests in the private fork. The code is also successfully tested. Thanks Pascal!

vogelpi (approved):

For the record: This is not production code but code that was used for penetration testing of Earlgrey ES. We should not implement changes to these tests that might have an impact on the test results. Instead the aim is to integrate the code more or less as is to be able to reproduce results and to have it ready for future penetration testing campaigns. The first PR of this series (#24294) has been reviewed by the SW team. The suggested style changes have also been integrated into this PR.

PR #24284, opened 2024-08-12
[bazel,silicon] build VMEM files for flash images for the silicon exec_envs

timothytrippel:
This updates silicon exec_env to build VMEM images. This enables running specific tests (like the `rom_e2e_self_hash_no_bkdr_rom_load` test) in GLS with the images compiled for silicon platforms.

CC: @sha-ron @eshapira

PR #24299, opened 2024-08-13
[dv/chip] change condition order in resolve_sw_image_paths

eshapira:
Hi,
This PR changes the condition order in the **resolve_sw_image_path** method in the **chip_env_cfg** file to check for **silicon_creator** before checking for **new_rules**.
This required specifically for tests that defined with both **new_rules** and **silicon creator** like the following test, in which the flash image created with **_silicon_creator** as its suffix.
https://github.com/lowRISC/opentitan/blob/13e9917aca2db743554ec8f3e9a55128c115c78c/hw/top_earlgrey/dv/chip_rom_tests.hjson#L805-L810
This PR comes with PR #24284


timothytrippel (approved):
Thanks @eshapira , this LGTM. This should be cherry-picked over to the  `earlgrey_1.0.0` branch. I can do that as well.

timothytrippel:
#24306 cherry-picks this commit to the `earlgrey_1.0.0` branch.

PR #24278, opened 2024-08-12
[rv_dm,dv] Exclude DmiOPFailed state coverage in dap

rswarbrick:
Detailed notes in a comment above the exclusion.

PR #24304, opened 2024-08-13
[rv_dm,dv] Exclude DmiOPFailed state coverage in dap

rswarbrick:
Detailed notes in a comment above the exclusion.


rswarbrick:
Well, that's embarrassing. Turns out that this was a duplicate of #24278 which I had opened on the previous day!

PR #24307, opened 2024-08-13
[rv_dm,dv] Add a scanmode sequence

### LINKED ISSUE #23763: [rv_dm,dv] Exercise scanmode
### Description

The current tests don't exercise JTAG scanmode. Fundamentally, this is because the JTAG driver assumes that it is in control of TCK (and can pause it for a bit). We probably need to test this out.

rswarbrick:
This is a (trivial) test to make sure that JTAG works as expected in scanmode.

Fixes #23763 (and should cover some conditions that we weren't seeing otherwise)


hcallahan-lowrisc (approved):
Not sure I love driving the vif from the sequence, but given the simple requirements of this testcase I think I can accept it :)

PR #24311, opened 2024-08-14
[tpm_personalize] drop CEK and CIK certificates generation

vbendeb:
These certificates are a ChromeOS TPM extension, let's keep their generation in cros extension.

Tested by running bazel test --test_output=streamed \
 //sw/device/silicon_creator/manuf/base:ft_provision_tpm_ext_fpga_hyper310_rom_with_fake_keys

Observed proper generation of USD, CDI_0, CDI_1, and TPM_EK certificates.


vbendeb:
TPM_CEK and TPM_CIK are not part of the TPM speck they belong in the ChromeOS extension, not in the main tree.

pamaury (approved):
Just to be sure: those will continue to be generated but not in the open source repo? If so, are there any features used by the CEK and CIK certificates and not by the DICE? In this case, we need to make sure that those are properly unit tested to avoid any regression. Could you double check the cert unit tests are up-to-date?

timothytrippel:
Yes, that is correct @pamaury , since the CEK and CIK are not an official part of any spec, and they use a strict subset of the features the EK cert uses (namely the subject_alt_name is only part of EK, not CEK and CIK), I don't think it makes sense to test E2E gen/provisioning of them in our CI in this repo. Since TPM EK is a spec, I think it makes sense to demonstrate how this can be provisioned using our "provisioning extensions" mechanism. Since CEK and CIK use only the most basic cert gen fields, the unit tests that exist should suffice.

timothytrippel (approved):

vbendeb:
> Yes, that is correct @pamaury , since the CEK and CIK are not an official part of any spec, and they use a strict subset of the features the EK cert uses (namely the subject_alt_name is only part of EK, not CEK and CIK), I don't think it makes sense to test E2E gen/provisioning of them in our CI in this repo. Since TPM EK is a spec, I think it makes sense to demonstrate how this can be provisioned using our "provisioning extensions" mechanism. Since CEK and CIK use only the most basic cert gen fields, the unit tests that exist should suffice.

there is nothing special in CEK/CIK certificates, they are subsets of TPM EK, having TPM EK in the main tree should be enough for testing.

PR #24324, opened 2024-08-14
[rom_ctrl,dv] Allow words in the expected digest to not all be wrong

### LINKED ISSUE #24321: [rom_ctrl,dv] Tweak the mask for invalid digests in rom_ctrl_base_vseq
### Description

This will currently make all words of the digest mismatch. Practically speaking, that's probably the behaviour that we'd see! But we have a coverage point to see a word that matches after one has mismatched.

Changing the corruption mask to sometimes have fewer nonzero bits should hit this coverage point.

rswarbrick:
Practically speaking, we wouldn't expect differences to be local like this. But causing this to happen should allow us to hit a conditional coverage item in the compare module (where matches_q should remain false after a mismatch even if a later word matches).

Fixes #24321.

### DIFF
--- a/hw/ip/rom_ctrl/dv/env/seq_lib/rom_ctrl_base_vseq.sv
+++ b/hw/ip/rom_ctrl/dv/env/seq_lib/rom_ctrl_base_vseq.sv
@@ -194,9 +194,12 @@ class rom_ctrl_base_vseq extends cip_base_vseq #(
 
     if (!as_expected) begin
       // We want to choose a digest that doesn't match. To do so, start with the expected digest and
-      // then xor with a random nonzero value.
+      // then xor with a nonzero value. If single_bit is set, the digest will only be wrong at a
+      // single index, which means that most of the words in the digest will match the expected
+      // digest (but not all of them).
       bit [kmac_pkg::AppDigestW-1:0] mask;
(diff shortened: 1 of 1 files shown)


hw/ip/rom_ctrl/dv/env/seq_lib/rom_ctrl_base_vseq.sv
```diff
@@ -194,9 +194,12 @@ class rom_ctrl_base_vseq extends cip_base_vseq #(
 
     if (!as_expected) begin
       // We want to choose a digest that doesn't match. To do so, start with the expected digest and
-      // then xor with a random nonzero value.
+      // then xor with a nonzero value. If single_bit is set, the digest will only be wrong at a
+      // single index, which means that most of the words in the digest will match the expected
+      // digest (but not all of them).
       bit [kmac_pkg::AppDigestW-1:0] mask;
-      `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(mask, mask != 0;)
+      bit                            single_bit = $urandom_range(0, 1);
```
hcallahan-lowrisc, line 201: The code makes sense to me. Not knowing much of the context, does applying this new stimulus constraint 50% of the time when `(!as_expected)` to hit a single coverpoint detract from coverage elsewhere?

hcallahan-lowrisc (approved):
LGTM. One small question.

PR #24326, opened 2024-08-14
[rom_ctrl,dv] Avoid an injected FSM error causing a test failure

### LINKED ISSUE #24319: [rom_ctrl,dv] Fix invalid check in scoreboard
### Description

The scoreboard has a check that before we get to the check phase, we've seen `pwrmgr_data.done`. This isn't correct because the `Invalid` FSM state is terminal (and doesn't cause the signal to become true).

Weaken the check to allow:
- A reset to be in progress
- The current state to be `Invalid`.

rswarbrick:
This should fix an occasional failure in the
rom_ctrl_corrupt_sig_fatal_chk test.

Fixes #24319.


eshapira (approved):
LGTM, thanks!

Issue #24322, opened 2024-08-14
[rom_ctrl,dv] Make some assertions terminate


rswarbrick:
### Description

Assertion coverage for rom_ctrl has several assertions that don't run to completion. These are of the form `A |-> always B` which... obviously doesn't complete! Tweak them to fix.

PR #24328, opened 2024-08-14
[dv] Allow assertions from ASSERT_FPV_LINEAR_FSM to complete

rswarbrick:
The assertion is supposed to start and say that we don't come back to a state until reset. Unfortunately, the implementation was disabling the property when reset happened, which meant that the assertion never finished, so it got marked as incomplete in the coverage.

Change the call to `ASSERT() so that it doesn't kill the property any more, relying on the fact that the property will terminate on its own once we go into reset.

This is associated with #24322, but I'm not marking the PR as fixing that issue because the "rom_ctrl assertions terminating" is addressed by a different PR.


hcallahan-lowrisc (approved):

hcallahan-lowrisc:

rswarbrick:

PR #24329, opened 2024-08-14
[rom_ctrl,dv] Remove irrelevant coverpoint from rom_ctrl_tlul_cg

### LINKED ISSUE #17501: [rom_ctrl,dv] Increase score of `rom_ctrl_tlul_cg`
The covergroup mentioned in the title has a [score of only 75%](https://reports.opentitan.org/hw/ip/rom_ctrl/dv/latest/cov_report/grp10.html). The reason is that the bin `req_before_done` in `cp_regs_req_check` and the bin `check_invalid` in `cp_rom_invalid_condition` are not covered.

Tests should be extended to cover those bins.

> estimate 4

rswarbrick:
The check_invalid bin in the coverpoint is not hit, but it doesn't really matter. The pwrmgr_data_invalid signal means that the pwrmgr_data_o.good mubi signal that we are generating has an invalid encoding. This can never happen without an injected fault.

We also don't care about whether this signal is valid or not: it is not actually read by any code in the IP block, so it doesn’t really matter if an injected fault causes it to become invalid.

Fixes #17501.

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_if.sv
+++ b/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_if.sv
@@ -21,7 +21,6 @@ interface rom_ctrl_cov_if (
 
   bit en_full_cov = 1'b1;
   bit en_intg_cov = 1'b1;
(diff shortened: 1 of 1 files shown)


hcallahan-lowrisc (approved):
Makes sense to me! Thanks

PR #24330, opened 2024-08-15
Spi host timeout fixes and clkdiv testing

antmarzam:
Split the clk-divider high values to be tested in a different config which sets  a `define and decrements the pres-scaler in 16-units to quicken up simulation.

Tweaked TB knobs to reduce test duration in general.

Extra checks in VSEQs to avoid timeouts in corner case situations

### DIFF
--- a/hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
+++ b/hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
@@ -30,6 +30,9 @@ class spi_host_base_vseq extends cip_base_vseq #(
   // reactive sequences to run on spi_agent (configured as a Device)
   spi_device_cmd_rsp_seq m_spi_device_seq[SPI_HOST_NUM_CS];
 
+  // Used to control when csr_spinwaits are spinning
+  uvm_event start_stop_spinwait_ev = uvm_event_pool::get_global("start_stop_spinwait_ev");
+
   // constraints for simulation loops
   constraint num_trans_c {
     num_trans inside {[cfg.seq_cfg.host_spi_min_trans : cfg.seq_cfg.host_spi_max_trans]};
@@ -94,19 +97,31 @@ class spi_host_base_vseq extends cip_base_vseq #(
 
   function void post_randomize();
     super.post_randomize();
+    // We currently support 1 chip-select line only
+
+    // We set the timeout to it's default value every time upon entry here.
+    // This is because the sim may go through multiple resets and if we don't
+    // re-set it's value it may not be useful as a timeout value
+    cfg.set_default_csr_timeout();
     case(spi_config_regs.clkdiv) inside
       // After randomization we set a different timeout based on the clock divider
-      [cfg.seq_cfg.host_spi_min_clkdiv : cfg.seq_cfg.host_spi_lower_middle_clkdiv]: ; // no change
+      [cfg.seq_cfg.host_spi_min_clkdiv : cfg.seq_cfg.host_spi_lower_middle_clkdiv]: ;
       [cfg.seq_cfg.host_spi_lower_middle_clkdiv+1 : cfg.seq_cfg.host_spi_middle_clkdiv]: begin
         cfg.csr_spinwait_timeout_ns *= 1.5;
       end
-      [cfg.seq_cfg.host_spi_middle_clkdiv+1 : 16'hFFF] : cfg.csr_spinwait_timeout_ns *= 5;
-      [16'hFFF+1 : cfg.seq_cfg.host_spi_max_clkdiv] : cfg.csr_spinwait_timeout_ns *= 10;
-      default : begin
-        `uvm_fatal(`gfn, $sformatf("spi_config_regs.clkdiv=0x%0x is out range",
-                                   spi_config_regs.clkdiv))
+      [cfg.seq_cfg.host_spi_middle_clkdiv+1 : cfg.seq_cfg.host_spi_upper_middle_clkdiv]: begin
+        cfg.csr_spinwait_timeout_ns *= 3;
       end
+      [cfg.seq_cfg.host_spi_middle_clkdiv+1 : 16'hFFF] : begin
+        cfg.csr_spinwait_timeout_ns *= 5;
+      end
+      [16'hFFF+1 : cfg.seq_cfg.host_spi_max_clkdiv] : begin
+        cfg.csr_spinwait_timeout_ns *= 10;
+      end
+      default : `uvm_fatal(`gfn, $sformatf("spi_config_regs.clkdiv[0]=0x%0x is out range"))
     endcase
+    `uvm_info(`gfn, $sformatf("%m: CSR_SPINWAIT Timeout set to %0dns", cfg.csr_spinwait_timeout_ns),
+              UVM_DEBUG)
   endfunction
 
 
@@ -306,13 +321,30 @@ class spi_host_base_vseq extends cip_base_vseq #(
     csr_wr(.ptr(ral.intr_state), .value(intr_state));
   endtask : spi_host_init
 
+  // In addition to un/setting sw_rst also triggers a uvm_event (and uvm_object)
+  // so any csr_spinwaits can be halted until the reset is over
   virtual task program_spi_host_sw_reset(int drain_cycles = SPI_HOST_RX_DEPTH);
+    uvm_reg_data_t control_reg;
+    csr_spinwait_ctrl_object spinwait_ctrl_obj;
+    spinwait_ctrl_obj = csr_spinwait_ctrl_object::type_id::create("spinwait_ctrl_obj");
     ral.control.sw_rst.set(1'b1);
     csr_update(ral.control);
+
+    spinwait_ctrl_obj.stop = 1;
+    start_stop_spinwait_ev.trigger(spinwait_ctrl_obj);
+    `uvm_info(`gfn, "Triggered 'start_stop_spinwait_ev' due to SW_RST=1", UVM_DEBUG)
     // make sure data completely drained from fifo then release reset
     wait_for_fifos_empty(AllFifos);
-    ral.control.sw_rst.set(1'b0);
-    csr_update(ral.control);
+    // Backdoor read to avoid writing somethging different to what's there
+    csr_rd(.ptr(ral.control), .value(control_reg), .backdoor(1));
+    control_reg[30] = 0; // SW reset field
+    csr_wr(.ptr(ral.control), .value(control_reg));
+    spinwait_ctrl_obj.stop = 0;
+    start_stop_spinwait_ev.trigger(spinwait_ctrl_obj);
+    `uvm_info(`gfn, "Triggered 'start_stop_spinwait_ev' due to SW_RST=0", UVM_DEBUG)
+
+    start_stop_spinwait_ev.reset();
+    `uvm_info(`gfn, "Resetting 'start_stop_spinwait_ev'", UVM_DEBUG)
   endtask : program_spi_host_sw_reset
 
   virtual task program_spi_host_regs();
@@ -421,6 +453,9 @@ class spi_host_base_vseq extends cip_base_vseq #(
 
   // wait until fifos has available entries to read/write
   virtual task wait_for_fifos_available(spi_host_fifo_e fifo = AllFifos);
+    // Wait for control.sw_rst = 0 before checking for fifo availability, since
+    // a SW reset causes both TX/RX FIFOs to drain
+    csr_spinwait(.ptr(ral.control.sw_rst), .exp_data(1'b0));
     if (fifo == TxFifo || fifo == AllFifos) begin
       csr_spinwait(.ptr(ral.status.txfull), .exp_data(1'b0));
     end
--- a/hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
+++ b/hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
@@ -68,8 +68,9 @@ class spi_host_base_vseq extends cip_base_vseq #(
   // Separate constraint to allow easy override in child sequences
   constraint spi_config_regs_clkdiv_c {
     spi_config_regs.clkdiv dist {
-      [cfg.seq_cfg.host_spi_min_clkdiv : cfg.seq_cfg.host_spi_lower_middle_clkdiv]      :/80,
-      [cfg.seq_cfg.host_spi_lower_middle_clkdiv+1 : cfg.seq_cfg.host_spi_middle_clkdiv] :/20
+      [cfg.seq_cfg.host_spi_min_clkdiv : cfg.seq_cfg.host_spi_lower_middle_clkdiv]      :/90,
+      [cfg.seq_cfg.host_spi_lower_middle_clkdiv+1 : cfg.seq_cfg.host_spi_middle_clkdiv] :/7,
+      [cfg.seq_cfg.host_spi_middle_clkdiv+1 : cfg.seq_cfg.host_spi_upper_middle_clkdiv] :/3
     };
   }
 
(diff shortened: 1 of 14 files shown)


hw/dv/sv/csr_utils/csr_utils_pkg.sv
```diff
@@ -592,6 +601,107 @@ package csr_utils_pkg;
     join
   endtask
 
+  // Like csr_spinwait, only the timeout can be stop by using an event.
+  // In addition, one needs to pass the clock cycle period.
+  // This can be used instead of csr_spinwait for moments in which the timeout needs to be
+  // controlled due to RTL conditions such as SW_rest or the module being disabled.
+  task automatic csr_spinwait_stoppable(input   uvm_object ptr,
```
rswarbrick, line 608: Rather than adding this task, would it be easier to add an optional argument to `csr_spinwait` and make that task stoppable? I think this would avoid us having to duplicate the implementation here.

hw/ip/spi_host/dv/spi_host_sim_faster_prescaler.hjson
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+{
+  variant: faster_prescaler
+
+  // Import the base spi_device sim_cfg file
+  import_cfgs: ["{proj_root}/hw/ip/spi_host/dv/base_sim_cfg.hjson"]
+
+  // Enable the appropriate build mode for all tests
+  build_opts: ["+define+FAST_PRESCALER=1"]
```
rswarbrick, line 11: Rather than splitting the config, I think you can do this by adding an extra build mode. I think you can:
- Add an extra "build mode" called something like fast_prescaler
- This build mode should add the option in question
- Then use that build mode in the test by selecting it with `build_mode`.

Untested, but I think you can do something like add this near the top o the file
```
  build_modes: [
    {
      name: fast_prescaler
      build_opts: ["+define+FAST_PRESCALER=1"]
    }
  ]
```

Then the test can be something like
```
    {
      name: spi_host_upper_range_clkdiv
      uvm_test_seq: spi_host_upper_range_clkdiv_vseq
      reseed: 10 // Reduced reseed since this test may take ages
      build_mode: "fast_prescaler"
    }
```

hw/ip/spi_host/dv/env/seq_lib/spi_host_upper_range_clkdiv_vseq.sv
```diff
@@ -20,13 +20,35 @@ class spi_host_upper_range_clkdiv_vseq extends spi_host_speed_vseq;
...
+      }
     }
   }
 
+`ifdef FAST_PRESCALER
+  constraint spi_config_regs_faster_clkdiv_c {
+    // Only multiples of 16 to quicken up the clock - this constraint should be part of the clkdiv
+    // VSEQ
+    // TODO: disable/adjust csnlead/idle/trail checks when this constraint is available.
+    //       currently those checks are part of a PR not yet in master.
```
rswarbrick, line 42: Which PR? This one? :-)

hw/dv/sv/csr_utils/csr_utils_pkg.sv
antmarzam: I'm not sure that would be quite the same, because we need new arguments (like the clock cycle, the uvm_event name),  and csr_spinwait is used all over the place. 
The original csr_spinwait wait's for a timeout and fails after it finishes.
So everywhere where csr_spinwait is used would need to be amended (Unless we use the default `cfg.clk_rst_vif.clk_period_ps`

hw/ip/spi_host/dv/env/seq_lib/spi_host_upper_range_clkdiv_vseq.sv
antmarzam: This one: https://github.com/lowRISC/opentitan/pull/24057

hw/ip/spi_host/dv/spi_host_sim_faster_prescaler.hjson
antmarzam: Thanks - I'm going to give this ago. 
Can you elaborate on what this does? 
Does this mean DVSIM has two different snapshots? One with the `define set and the other one without?

hw/dv/sv/csr_utils/csr_utils_pkg.sv
antmarzam: Thinking a bit more about this. I think I agree with you overall, long term goal should be to have `csr_spinwait_stoppable` be the same as `csr_spinwait` but that'd require going around each TB and making sure we're passing the right clock period. 

I'm not sure how many TBs change the clk_period defined in `clk_rst_if`, as it's currently hardcoded rather than derived from a localparam or similar from a package:
https://github.com/lowRISC/opentitan/blob/bc90ee9322daea58dc71242a053c427b01e98f9c/hw/dv/sv/common_ifs/clk_rst_if.sv#L50

Perhaps another approach would be for the dv_base_test to put in `uvm_config_db` the clk_period, and then `csr_spinwait(_stoppable)` could just get it from there. The issue I see with this is that uvm_config_db calls are expensive and this would be done in each call to the task

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -118,26 +119,135 @@ class spi_host_base_vseq extends cip_base_vseq #(
...
+          end
+        join_any
+        disable fork;
+      end : isolation_fork
+    join
+  endtask
+
+  // wrapping csr_spinwait over csr_utils_pkg::csr_spinwait_stoppable so we can stop the timeout
+  // count in circumstances like the RTL being disable spien or sw_rst.
+  virtual task automatic csr_spinwait( input  uvm_object        ptr,
```
rswarbrick, line 242: We had a bit of a chat about this, and I think we have a plan which avoids having to copy-paste bits of the base class.
- Change this override so that it uses `fork/join` to call the superclass in parallel with the "chunked wait" that you added.
- But call the superclass with an insanely large timeout, so that the raw timeout will never fire and we will only get the new "chunked wait" doing the timeout.

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
```diff
@@ -160,6 +160,14 @@ package spi_host_env_pkg;
   parameter uint NUM_ALERTS = 1;
   parameter string LIST_OF_ALERTS[] = {"fatal_fault"};
 
+  class csr_spinwait_ctrl_object extends uvm_object;
+    `uvm_object_utils(csr_spinwait_ctrl_object)
+    bit            stop;
+    function new(string name = "csr_spinwait_ctrl_object");
```
rswarbrick, line 166: I know that it is standard elsewhere in the code, but I think we can drop the default value for the argument.

hw/ip/spi_host/dv/spi_host_sim_faster_prescaler.hjson
rswarbrick: I think `dvsim` will just end up with two different RTL simulations that it builds. When it decides to run the `spi_host_upper_range_clkdiv` test, it will do so by running the "fast one".

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
antmarzam: good point - done!

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
antmarzam: I implemented this suggestion and just pushed the changes :)

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
```diff
@@ -160,6 +160,14 @@ package spi_host_env_pkg;
   parameter uint NUM_ALERTS = 1;
   parameter string LIST_OF_ALERTS[] = {"fatal_fault"};
 
+  class csr_spinwait_ctrl_object extends uvm_object;
+    `uvm_object_utils(csr_spinwait_ctrl_object)
+    bit            stop;
```
rswarbrick, line 165: It's a boring comment from me, but would you mind adding a couple of lines of comment documentation before the class and/or this variable?

hw/ip/spi_host/dv/spi_host_sim_faster_prescaler.hjson
rswarbrick: Any luck trying the `build_mode` approach? I think it should make things somewhat simpler.

hw/ip/spi_host/dv/env/seq_lib/spi_host_upper_range_clkdiv_vseq.sv
rswarbrick: In which case, that's awesome (since it's now merged!). Can we drop the TODO?

hw/ip/spi_host/dv/spi_host_sim_faster_prescaler.hjson
antmarzam: Yes it works :) Thank you!

hw/ip/spi_host/dv/env/seq_lib/spi_host_upper_range_clkdiv_vseq.sv
antmarzam: Dropped it, and also pushed a commit with the SCB checks being disabled when `FAST_PRESCALER` is defined

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -118,26 +119,122 @@ class spi_host_base_vseq extends cip_base_vseq #(
...
-    csr_utils_pkg::csr_spinwait( .ptr(ptr), .exp_data(exp_data), .check(check), .path(path),
-                                 .map(map), .user_ftdr(user_ftdr),
-                                 .spinwait_delay_ns(spinwait_delay_ns),
-                                 .timeout_ns(timeout_ns), .compare_op(compare_op),
-                                 .backdoor(backdoor), .verbosity(verbosity));
+  // Like csr_spinwait, only the timeout can be stop by using an event.
+  // In addition, one needs to pass the clock cycle period.
+  // This can be used instead of csr_spinwait for moments in which the timeout needs to be
+  // controlled due to RTL conditions such as SW_rest or the module being disabled.
+  task automatic csr_spinwait_stoppable(input   uvm_object ptr,
```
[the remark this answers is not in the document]
antmarzam: Moved the first argument to the line below and removed the extra lines!

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -118,26 +119,122 @@ class spi_host_base_vseq extends cip_base_vseq #(
...
+    input   compare_op_e compare_op = CompareOpEq,
+    input   bit backdoor = 0,
+    input   uvm_verbosity verbosity = UVM_HIGH,
+    input   uint clk_cycle_period_ns,
+    input   string start_stop_ev_name =
+                                "start_stop_spinwait_ev"
+    );
+    static int     count;
+    count++;
+    `uvm_info($sformatf("%m()"), $sformatf(
```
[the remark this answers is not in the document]
antmarzam: Done!

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -118,26 +119,122 @@ class spi_host_base_vseq extends cip_base_vseq #(
...
+          begin: timeout_block
+            automatic csr_field_t csr_or_fld;
+            automatic int lcount = count;
+            automatic int unsigned clk_cycles = 0;
+            automatic csr_spinwait_ctrl_object obj;
+            automatic uvm_event start_stop_spinwait_ev;
+            start_stop_spinwait_ev = uvm_event_pool::get_global(start_stop_ev_name);
+
+            csr_or_fld = decode_csr_or_field(ptr);
+            // TODO: make a function/task out of this? Something like "controlled timeout"?
```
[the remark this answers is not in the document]
antmarzam: Moved it to its own task called `stoppable_timeout`

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -118,26 +119,122 @@ class spi_host_base_vseq extends cip_base_vseq #(
...
+            while ( (clk_cycles * clk_cycle_period_ns) <= timeout_ns) begin
+              clk_cycles++;
+              #(clk_cycle_period_ns * 1ns);
```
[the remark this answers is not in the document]
antmarzam: I prefer to keep it with the `while` loop. It reads easier that way to me.

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
antmarzam: Done! LMK if the comment makes sense :)

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
rswarbrick: Hmm: this seems to be defaulting to `""`. Is a default actually needed by e.g. `uvm_object_utils`? If not, maybe we should drop it entirely?

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
```diff
@@ -120,26 +121,133 @@ class spi_host_base_vseq extends cip_base_vseq #(
...
+                                        // block the count
+                                        .timeout_ns(timeout_ns*10000),
+                                        .compare_op(compare_op),
+                                        .backdoor(backdoor),
+                                        .verbosity(verbosity)
+                                        );
+          end
+          begin: timeout_block
+            automatic int lcount = count;
+            stoppable_timeout(clk_cycle_period_ns, timeout_ns, start_stop_ev_name, exp_data,
```
rswarbrick, line 185: This is nice! While we're at it, I think we can drop the automatic variable and can just call `stoppable_timeout` with `count`?

We also end up with just a single statement in each process (which is nice!). We *could* drop the begin/end pairs. I'm not sure how you feel about that though: it's probably how I'd do it, but I'm also happy for this layout :-)

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
rswarbrick: Looks great to me!

hw/ip/spi_host/dv/spi_host_sim_faster_prescaler.hjson
antmarzam: That makes sense! I removed the `base_sim_cfg.hjson` and moved the necessary bits back to `spi_host_sim_cfg.hjson`

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
antmarzam: As we discussed earlier, it seems you can't have an empty name in the constructor for UVM objects/components. 
I've left it with an empty string name.

hw/ip/spi_host/dv/env/seq_lib/spi_host_base_vseq.sv
antmarzam: I've dropped the automatic `lcount` variable. 
For the fork structure I think I prefer it as is since I think it's more readable that way :)

rswarbrick (commented):
I've got a suggestion that I think avoids us needing to bodge the RTL: worth a try?

hw/ip/spi_host/dv/spi_host_fsm_if.sv
```diff
@@ -0,0 +1,29 @@
...
+  // greater than 1 (currently hardcoded at 16 when in fast_mode)
+  bit fast_mode;
+
+  always @(posedge u_fsm.clk_i) begin
+    if (u_fsm.rst_ni && fast_mode) begin
+      if (!(u_fsm.sw_rst_i || !u_fsm.clk_cntr_en ||
+            u_fsm.new_command || u_fsm.is_idle || !u_fsm.clk_cntr_q)) begin
+        // This check matches the one defining clk_cntr_d in spi_host_fsm.sv. It expects to
+        // decrement by 1. If we are in "fast mode", we should decrement by 16 instead.
+        //        if (fast_mode) begin
```
rswarbrick, line 21: Hmm. Is this supposed to be commented?

hw/ip/spi_host/dv/env/spi_host_env_cfg.sv
```diff
@@ -31,6 +31,10 @@ class spi_host_env_cfg extends cip_base_env_cfg #(.RAL_T(spi_host_reg_block));
   // higher values
   uint csr_spinwait_timeout_ns = csr_utils_pkg::default_spinwait_timeout_ns;
 
+  // Used to provide a faster decrement via forcing internal RTL signals, so simulations
+  // finish more quickly. The same
```
rswarbrick, line 35: Two stray words at the end of the line :-)

rswarbrick (commented):
This looks really nice! A couple of minor tweaks needed, but I like it!

hw/ip/spi_host/dv/spi_host_fsm_if.sv
antmarzam: Oops, sorry! I Forget to remove it before pushing. Done now :)

hw/ip/spi_host/dv/env/spi_host_env_cfg.sv
antmarzam: Removed, thank you!

rswarbrick (approved):
Looks great, thanks!

PR #24333, opened 2024-08-15
[dv] Some cleanups/documentation for csr_spinwait

rswarbrick:
There should be no functional change, and the commit messages explain things further.


marnovandermaas (approved):
I'm reasonably convinced that there is no functional change here.

PR #24336, opened 2024-08-15
[rv_dm,dv] Hit minor conditional coverage hole for ndmreset signals

rswarbrick:
This took *ages* to figure out! I think this is the simplest way to see it happen.


hcallahan-lowrisc (approved):
Makes sense to me, and as you say it shouldn't have any other effect on the intended overall stimulus for this sequence.

Issue #24277, opened 2024-08-12
[cryptolib] Allow garbage collection when linking OTBN binaries to reduce code size


johannheyszl:
### Description

It would be useful to let the OTBN linker's garbage collection remove unused code sections to reduce size. This allows tailored OTBN binaries of reduced size and even combining different cryptographic functions into one binary though careful selection.

This will require:
- Adding` --gc-sections` to the linker script.
- Moving functions (or collections thereof) that are candidates for removal to separate code `sections` in respective `.asm `files. Alternatively, a split into multiple files is also an option.

Note that the linker will understand which functions are called e.g. from the[ interface layer](https://github.com/lowRISC/opentitan/blob/fd78577dc27a434550ff8bff7d0ed2439d425bcd/sw/otbn/crypto/p256_ecdsa.s#L35) where functions are called after switch-casing on the command.

johannheyszl:
Note: I have successfully tested this by adding  --gc-sections to the [linker script](https://github.com/lowRISC/opentitan/blob/fd78577dc27a434550ff8bff7d0ed2439d425bcd/hw/ip/otbn/util/otbn_ld.py#L69)

johannheyszl:
As @vsukhoml noted, it would make sense that we put actual function implementations and the start section ([example](https://github.com/lowRISC/opentitan/blob/fd78577dc27a434550ff8bff7d0ed2439d425bcd/sw/otbn/crypto/p256_ecdsa.s#L35)) into separate files.

jadephilipoom:
We already do this, see e.g. https://github.com/lowRISC/opentitan/blob/master/sw/otbn/crypto/p256_ecdsa.s and https://github.com/lowRISC/opentitan/blob/master/sw/otbn/crypto/rsa.s

vsukhoml:
It is close enough, but not exactly. I mean to put this part:

```
.section .text.start
.globl start
start:
  /* Read the mode and tail-call the requested operation. */
  la    x2, mode
  lw    x2, 0(x2)

  addi  x3, x0, MODE_KEYGEN
  beq   x2, x3, random_keygen

  addi  x3, x0, MODE_SIGN
  beq   x2, x3, ecdsa_sign

  addi  x3, x0, MODE_VERIFY
  beq   x2, x3, ecdsa_verify

  addi  x3, x0, MODE_SIDELOAD_KEYGEN
  beq   x2, x3, sideload_keygen

[...]

PR #24343, opened 2024-08-16
[crypto] Read mode constants from OTBN binaries.

jadephilipoom:
Small adjustment to the cryptolib OTBN routines that allows them to read mode constants directly from the OTBN binary, so that these do not have to be manually adjusted when they change. This is a minor cleanup I'd been planning on for a while, and had the impetus to fix because of a discussion in https://github.com/lowRISC/opentitan/issues/24277


rswarbrick (approved):
Huh, neat! I like the way that we can pass the information through like this.

Issue #21726, opened 2024-02-27
[manuf,ottf] implement spi_device console over mailbox 


timothytrippel:
Previously, in #17818, a spi_device OTTF console was added to facilitate manufacturing flows that require used of higher speed IO (than is available with UART). When generic mode was removed in #20856, this feature was also removed, since it made use of generic mode.

To facilitate manufacturing flows for Earlgrey PROD, a new spi_device OTTF console should be implemented using the SPI mailbox.

timothytrippel:
CC: @anthonychen1251

anthonychen1251:
The PRs associated with this task are listed below for the reference
- #24345
- #24366
- #24380
- #24413

PR #24345, opened 2024-08-16
[base,ottf,opentitanlib,test] Impl spi device ottf RX console

anthonychen1251:
This PR has three commits that:

1.  Implement a new spi_device Rx console by using the SPI device flash mode feature.
2.  Update the opentitanlib SPI console module (added in#17817) to support OTTF spi_device console by using SPI device flash mode feature instead of the removed generic mode .
3.  Add a test for the new OTTF spi_device console.

This PR addresses #21726 partially.  The TX (input) direction will be added in another PR.


timothytrippel:
Looks like you have some build errors. Trying seeing what they are locally by running: `bazel build //sw/host/opentitanlib`

anthonychen1251:
> Looks like you have some build errors. Trying seeing what they are locally by running: `bazel build //sw/host/opentitanlib`

Thanks for the instructions. Those build errors have been addressed in the new commits. Not sure why those errors didn't appear when I ran the newly added test for OTTF spi_device console though.

The `spi_host_config_test` failed in the most recent CI pipeline, but this test seems to be unrelated to the commits included in this PR.
https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=158355&view=logs&j=d6ba1a9c-e7a8-5acf-e6fb-86a516d856bf&t=e16cc888-fc04-5078-0841-ecafa9afe0ed&l=559

I also tried to run this test locally
`bazel test --test_output=streamed //sw/device/tests:spi_host_config_test_fpga_cw340_sival_rom_ext`
and the result was passed.

sw/host/opentitanlib/src/console/spi.rs
```diff
@@ -46,8 +78,33 @@ impl<'a> SpiConsoleDevice<'a> {
...
+            self.flash
+                .read(self.spi, data_address, &mut data[0..first_part_size])?;
+            self.flash.read(
+                self.spi,
+                0,
+                &mut data[first_part_size..data_len_bytes_w_pad],
+            )?;
+        } else {
+            self.flash.read(self.spi, data_address, &mut data)?;
+        }
```
timothytrippel, line 101: consider wrapping in a function since this code block is similar to the above?

sw/device/tests/spi_device_ottf_console_test.c
```diff
@@ -0,0 +1,24 @@
...
+
+OTTF_DEFINE_TEST_CONFIG(.console.type = kOttfConsoleSpiDevice,
+                        .console.base_addr = TOP_EARLGREY_SPI_DEVICE_BASE_ADDR,
+                        .console.test_may_clobber = false, );
+
+static const char test_str[] = "This is a SPI device OTTF console test.";
+
+bool test_main(void) {
+  for (int i = 0; i < 100; i++) {
+    LOG_INFO("%s round: %d", test_str, i);
```
timothytrippel, line 20: maybe this test should try writing varying size strings to the console? rather than the same length string each time? 

Also add some intentional corner cases, e.g.:
- string bigger than the spi flash egress buffer
- empty string
- string that wraps around buffer but is smaller than egress buffer, etc.

sw/host/opentitanlib/src/console/spi.rs
anthonychen1251: Good idea! Encapsulated the code snippets into `fn read_data` functions in the latest commit.

sw/device/lib/runtime/print.c
```diff
@@ -68,6 +71,151 @@ void base_set_stdout(buffer_sink_t out) {
...
+ * -----------------------------------|   Data   |
+ * |     0xFF Pad Bytes    | <4-bytes |          |
+ * -----------------------------------|----------|
+ */
+static size_t spi_device_send_frame(void *data, const char *buf, size_t len) {
+  dif_spi_device_handle_t *spi_device = (dif_spi_device_handle_t *)data;
+  const size_t kDataPacketSizeBytes = ((len + 3u) & ~3u);
+  const size_t kFrameSizeBytes =
+      kSpiDeviceFrameHeaderSizeBytes + kDataPacketSizeBytes;
+  uint8_t frame_bytes[kFrameSizeBytes];
```
pamaury, line 106: This is a potentially large buffer on the stack? Is it possible to directly write into the SPI buffer, at least for the payload? (the header could be constructed on the stack)

sw/device/lib/runtime/print.c
```diff
@@ -68,6 +71,151 @@ void base_set_stdout(buffer_sink_t out) {
...
+            (uint8_t *)(frame_bytes + first_part_size)) != kDifOk) {
+      return 0;
+    }
+  } else {
+    if (dif_spi_device_write_flash_buffer(
+            spi_device, kDifSpiDeviceFlashBufferTypeEFlash, next_write_address,
+            kFrameSizeBytes, frame_bytes) != kDifOk) {
+      return 0;
+    }
+  }
```
engdoreis, line 188: This can be slightly simplified to improve readability.
```suggestion
   size_t round_size = next_write_address + kFrameSizeBytes;
   round_size = round_size > kSpiDeviceReadBufferSizeBytes ? kSpiDeviceReadBufferSizeBytes
                                                           : round_size;
   TRY(dif_spi_device_write_flash_buffer(
           spi_device, kDifSpiDeviceFlashBufferTypeEFlash, next_write_address,
           round_size, frame_bytes) != kDifOk) );
   // Handle wrap-around.
   if ( round_size < kFrameSizeBytes){
     TRY(dif_spi_device_write_flash_buffer( spi_device, kDifSpiDeviceFlashBufferTypeEFlash,
           0, kFrameSizeBytes - round_size,
           (uint8_t *)(frame_bytes + round_size)) != kDifOk))
   }
  ```

File: sw/host/tests/chip/spi_device_ottf_console/src/main.rs [hunk 5]
```diff
@@ -0,0 +1,82 @@
...
+        ..Default::default()
+    };
+    let mut stdout = std::io::stdout();
+    let spi = transport.spi(&opts.console_spi)?;
+
+    let spi_console_device = SpiConsoleDevice::new(&*spi)?;
+    let _ = UartConsole::wait_for(&spi_console_device, r"Running [^\r\n]*", opts.timeout)?;
+
+    let result = console.interact(&spi_console_device, None, Some(&mut stdout))?;
+    match result {
```
engdoreis, line 50: This test can be improved by reading the strings from the elf and checking the equality on the host side.
i.e:
https://github.com/lowRISC/opentitan/blob/f1535c55404f8d314246805a3b1788e8efaa6173/sw/host/tests/chip/i2c_host_override/src/main.rs#L106

engdoreis (commented):
Thanks for the PR, we were look for forwarding for this feature.
I left a few comments below.

sw/device/lib/runtime/print.c
anthonychen1251: Yes, a ~2KB buffer will be allocated on the stack in the worst case. To optimize memory usage, the implementation has been modified to allow for the direct writing of aligned data into the SPI buffer.  A frame is split into three parts:
1. Header: A 12-byte array is constructed on the stack
2. Aligned Data: This data is written directly into the SPI buffer
3. Unaligned Data: A 4-byte array is constructed on the stack to hold this data.

sw/device/lib/runtime/print.c
```diff
@@ -68,6 +71,151 @@ void base_set_stdout(buffer_sink_t out) {
...
+ * |   Data Length (bytes) | 4-bytes  |          |
+ * -----------------------------------|----------|
+ * |      Data (word aligned)         |          |
+ * -----------------------------------|   Data   |
+ * |     0xFF Pad Bytes    | <4-bytes |          |
+ * -----------------------------------|----------|
+ */
+static size_t spi_device_send_frame(void *data, const char *buf, size_t len) {
+  dif_spi_device_handle_t *spi_device = (dif_spi_device_handle_t *)data;
+  const size_t kDataPacketSizeBytes = ((len + 3u) & ~3u);
```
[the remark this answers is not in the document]
anthonychen1251: Got it. Used snake case here in the latest commit.

sw/device/lib/runtime/print.c
```diff
@@ -68,6 +71,151 @@ void base_set_stdout(buffer_sink_t out) {
...
+  const size_t kFrameSizeBytes =
+      kSpiDeviceFrameHeaderSizeBytes + kDataPacketSizeBytes;
+  uint8_t frame_bytes[kFrameSizeBytes];
+
+  static uint32_t next_write_address = 0;
+
+  if (kFrameSizeBytes >= (kSpiDeviceReadBufferSizeBytes)) {
```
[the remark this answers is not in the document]
anthonychen1251: Removed, that was a leftover.

sw/device/lib/runtime/print.c
anthonychen1251: Thanks for the suggestion! If I understand the idea correctly, an additional line `round_size-=next_write_address;` needs to be added between line 215 and 216 in the suggestion. The implementation has been modified to align with the proposed suggestion, along with some additional modifications. Let me know if the new implementations meet your expectations.

anthonychen1251: I would like to add a note that the “TRY” macro wasn’t employed here because the function returns a `size_t` value.

sw/host/tests/chip/spi_device_ottf_console/src/main.rs
anthonychen1251: Thanks! The implementation has been modified to read the string from the elf and check the equality on the host side.

sw/device/lib/runtime/print.c
timothytrippel: Yes, let's avoid the use of TRY in this function as that is not the return type and this is a runtime lib. @engdoreis

sw/device/tests/BUILD
```diff
@@ -3593,6 +3593,29 @@ cc_library(
...
 )
 
+opentitan_test(
+    name = "spi_device_ottf_console_test",
+    srcs = ["spi_device_ottf_console_test.c"],
+    exec_env = dicts.add(
+        EARLGREY_CW340_TEST_ENVS,
+    ),
+    fpga = fpga_params(
+        tags = ["manual"],
```
timothytrippel, line 3603:
```suggestion
```
so this runs in CI

anthonychen1251: Removed.

sw/device/lib/testing/test_framework/ottf_console.c
```diff
@@ -118,6 +109,91 @@ void ottf_console_configure_uart(uintptr_t base_addr) {
...
+      {
+          // Slot 5: ReadNormal
+          .opcode = kSpiDeviceFlashOpReadNormal,
+          .address_type = kDifSpiDeviceFlashAddr3Byte,
+          .dummy_cycles = 0,
+          .payload_io_type = kDifSpiDevicePayloadIoSingle,
+          .payload_dir_to_host = true,
+      },
+  };
+  for (uint8_t i = 0; i < ARRAYSIZE(read_commands); ++i) {
```
timothytrippel, line 172: nit:
```suggestion
  for (size_t i = 0; i < ARRAYSIZE(read_commands); ++i) {
```

same for below.

sw/device/tests/BUILD
```diff
@@ -3593,6 +3593,28 @@ cc_library(
     ],
 )
 
+opentitan_test(
+    name = "spi_device_ottf_console_test",
+    srcs = ["spi_device_ottf_console_test.c"],
+    exec_env = dicts.add(
+        EARLGREY_CW340_TEST_ENVS,
```
timothytrippel, line 3600: can you add the first two exec environments here too (the `fpga_cw310*` ones): https://cs.opensource.google/opentitan/opentitan/+/master:rules/opentitan/defs.bzl;drc=8fdb25c8d96f8a7b267a3344058299b127278765;l=90 ?

sw/device/tests/spi_device_ottf_console_test.c
```diff
@@ -0,0 +1,114 @@
...
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+OTTF_DEFINE_TEST_CONFIG(.console.type = kOttfConsoleSpiDevice,
+                        .console.base_addr = TOP_EARLGREY_SPI_DEVICE_BASE_ADDR,
+                        .console.test_may_clobber = false, );
+
+static const char test_str[] = "This is a SPI device OTTF console test.";
```
timothytrippel, line 16:
```suggestion
static const char kTestStr[] = "This is a SPI device OTTF console test.";
```
see https://google.github.io/styleguide/cppguide.html#Constant_Names

applies for others below too

sw/host/tests/chip/spi_device_ottf_console/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Name of the SPI interface to connect to the OTTF console.
+    #[arg(long, default_value = "BOOTSTRAP")]
+    console_spi: String,
+
+    /// Path to the firmware's ELF file, for querying symbol addresses.
+    #[arg(value_name = "FIRMWARE_ELF")]
+    firmware_elf: PathBuf,
+}
+
+fn spi_device_console_test(opts: &Opts, transport: &TransportWrapper) -> Result<()> {
```
timothytrippel, line 37: Awesome test!

timothytrippel (approved):
Couple more minor improvements, otherwise this looks great! Thanks @anthonychen1251 . Once you make the last couple improvements, and this passes CI. I can merge it.

sw/device/tests/BUILD
anthonychen1251: Added. Since I don't have cw310 environment to run the test in advance, let's see if the test works as expected on the CI pipeline.

sw/device/tests/spi_device_ottf_console_test.c
anthonychen1251: Renamed these variables. Thanks for the reference.

sw/device/lib/testing/test_framework/ottf_console.c
anthonychen1251: Fixed.

sw/device/tests/BUILD
anthonychen1251: Looks like the test works as expected on cw310 https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=158483&view=logs&j=6d7ef521-d3a1-575b-de1b-a7342dcf1a8e&t=5fbf7d75-9dee-5f42-7084-95893df8eb52&l=14748

Some failures appear to be unrelated to this PR, though
https://github.com/lowRISC/opentitan/pull/24345/checks?check_run_id=29048541424
Regarding the error in private CI, I am not sure if it's related to this PR because I don't have permission to access private CI.

engdoreis:
Thanks for the changes.

After a discussion with @jwnrt @HU90m, we came up with a suggestion for the protocol that we believe would simplify the implementation and be more robust, by leveraging the spi_flash features to control the data flow.

1. The device would always write the header to the beginning of the buffer.
2. The Host must read the status register to check the busy bit is cleared before issuing a `kSpiDeviceFlashOpReadNormal` command.
3. After reading, the host must issue a `kSpiDeviceFlashOpSectorErase` to signal to the device that the frame has been read via the busy bit.
4. The device must check that the busy bit is set before writing a new frame to the buffer. 
5. The device writes a new frame to the buffer and clear the busy bit.

Does that make sense?

One downside thought is that we could have a deadlock in the device if the host does not poll the device. But that can be sorted with a timeout.

jwnrt:
> One downside thought is that we could have a deadlock in the device if the host does not poll the device. But that can be sorted with a timeout.

Minor nit: I don't think it would be a deadlock, it would just be blocked on the host doing its own thing

timothytrippel:
> Thanks for the changes.
> 
> After a discussion with @jwnrt @HU90m, we came up with a suggestion for the protocol that we believe would simplify the implementation and be more robust, by leveraging the spi_flash features to control the data flow.
> 
> 1. The device would always write the header to the beginning of the buffer.
> 2. The Host must read the status register to check the busy bit is cleared before issuing a `kSpiDeviceFlashOpReadNormal` command.
> 3. After reading, the host must issue a `kSpiDeviceFlashOpSectorErase` to signal to the device that the frame has been read via the busy bit.
> 4. The device must check that the busy bit is set before writing a new frame to the buffer.
> 5. The device writes a new frame to the buffer and clear the busy bit.
> 
> Does that make sense?
> 
> One downside thought is that we could have a deadlock in the device if the host does not poll the device. But that can be sorted with a timeout.

What is the benefit of 

> Thanks for the changes.
> 
> After a discussion with @jwnrt @HU90m, we came up with a suggestion for the protocol that we believe would simplify the implementation and be more robust, by leveraging the spi_flash features to control the data flow.
> 
> 1. The device would always write the header to the beginning of the buffer.
> 2. The Host must read the status register to check the busy bit is cleared before issuing a `kSpiDeviceFlashOpReadNormal` command.
> 3. After reading, the host must issue a `kSpiDeviceFlashOpSectorErase` to signal to the device that the frame has been read via the busy bit.
> 4. The device must check that the busy bit is set before writing a new frame to the buffer.
> 5. The device writes a new frame to the buffer and clear the busy bit.
> 
> Does that make sense?
> 
> One downside thought is that we could have a deadlock in the device if the host does not poll the device. But that can be sorted with a timeout.

This seems like it would lower the potential throughput always having to align the frame header at the beginning of the buffer no?

Let's get this merged before making any further improvements to it. It is blocking critical provisioning work.

sw/device/tests/BUILD
timothytrippel: I re-trigger CI; hopefully these should pass now.

jwnrt:
> This seems like it would lower the potential throughput always having to align the frame header at the beginning of the buffer no?

You mean because the current implementation can queue up multiple frames in the circular buffer without having to wait for the host to read them out?

timothytrippel:
Yes

jwnrt:
If we're sending lots of small frames, then no. If it's large amounts of data filling the buffer then it should be the same.

engdoreis:
I'm fine to merge as it is, and we can experiment other protocols later.

timothytrippel:
Looks like the test that was failing was a flaky test that provided no additional value to the test scenario it was exercising so I added a commit to this PR that should fix the issue and get this to pass CI.

anthonychen1251:
> Thanks for the changes.
> 
> After a discussion with @jwnrt @HU90m, we came up with a suggestion for the protocol that we believe would simplify the implementation and be more robust, by leveraging the spi_flash features to control the data flow.
> 
> 1. The device would always write the header to the beginning of the buffer.
> 2. The Host must read the status register to check the busy bit is cleared before issuing a `kSpiDeviceFlashOpReadNormal` command.
> 3. After reading, the host must issue a `kSpiDeviceFlashOpSectorErase` to signal to the device that the frame has been read via the busy bit.
> 4. The device must check that the busy bit is set before writing a new frame to the buffer.
> 5. The device writes a new frame to the buffer and clear the busy bit.
> 
> Does that make sense?
> 
> One downside thought is that we could have a deadlock in the device if the host does not poll the device. But that can be sorted with a timeout.

Thanks for your thoughtful suggestions! Before creating this PR, I did experiment with a similar approach where the host issued an upload command to signal the device that the frame has been read. In this protocol, each data chunk transfer is a blocking operation, even if there is free space in the read buffer to accommodate the next chunk. 

While the proposed protocol in this PR utilizes the `LAST_READ_ADDR` to coordinate Device and Host activities, so the device is able to push data chunks into the read buffer non-blocking and only block when there is insufficient space to accommodate the next chunk. 

To illustrate this potential advantage, I conducted a simple experiment by logging a test message `This is a SPI device OTTF console test.`  on the device side with varying iterations. While not a rigorous evaluation, the results suggest a throughput improvement with the non-blocking approach.
|   Iterations | blocking (seconds)   | non-blocking (seconds)   | ratio   |
|-------------:|:---------------------|:-------------------------|:----------|
|           50 | 0.381s               | 0.215s                   | 1.77     |
|          100 | 0.737s               | 0.426s                   | 1.73     |
|          500 | 3.735s               | 2.241s                   | 1.67     |
|         1000 | 7.496s               | 4.512s                   | 1.66     |
|         5000 | 37.598s              | 22.478s                  | 1.67     |

I acknowledge that the suggested protocol offers implementation simplicity. It seems there's a trade-off here between potential throughput performance and implementation simplicity/robustness.

PR #24346, opened 2024-08-16
[otbn,dv] Correct the timing for when we zero insn_cnt in model

rswarbrick:
There are two paths by which this can happen in the RTL (the insn_cnt_clear_int_o signal from otbn_start_stop_control and the way that otbn_controller can go into a locked state). These have slightly different timings!

Make sure to model them correctly.

This fixes failures in `otbn_escalate`. Last night had 5 failures in 60 runs, four of which were caused by this sort of mismatch. I've just run some seeds locally and I see this mismatch cause failures in 2 runs of 100, so the "pass rate" for this check goes from 93% to 98%, which should hopefully be worth it.

### DIFF
--- a/hw/ip/otbn/dv/otbnsim/sim/sim.py
+++ b/hw/ip/otbn/dv/otbnsim/sim/sim.py
@@ -137,12 +137,12 @@ class OTBNSim:
 
         return changes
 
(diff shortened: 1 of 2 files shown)


marnovandermaas (approved):
This looks reasonable to me assuming you've run these tests and are happy with the results.

PR #24347, opened 2024-08-16
[otbn,dv] Weaken an assertion about tracking busy/locking state

rswarbrick:
The NotBusyAndDone_A assertion wasn't quite correct. We can make it fail by injecting a fatal error when the core is in the middle of its initial wipe. If that happens, we don't bother doing a proper locking procedure (there's nothing important to wipe!) and the "busy_execute_d" signal stays high, but "locking" goes high.

These assertions were added by me (oops) last May and I don't think they really make much sense anyway. Drop the one that isn't true.


hcallahan-lowrisc (approved):
LGTM!

rswarbrick:

hcallahan-lowrisc:

PR #24354, opened 2024-08-18
[dv, gls] Update ibex PMP signals for gate-level

eshapira:
Cherry-pick one commit from master to earlgrey_1.0.0


andreaskurth:
This would cherry-pick #24353 to `earlgrey_1.0.0`. I agree with that, but we need cherry-pick commits to adhere to a common format (see previous commits on that branch for reference). I'll shortly create a PR that includes this one.

Issue #23237, opened 2024-05-21
[manuf] TPM provisioning sequence needs to include generation of a 512 bit secret


vbendeb:
### Description

This secret needs to be reproducible by the chip in the specific state of the key ladder, and a derivative of this secret (usually the 16th order of hmac) needs to be reported and stored by the host during personalization.

timothytrippel:
@vbendeb I think we can close this now?

timothytrippel:
Confirmed with @vbendeb ; closing.

PR #24355, opened 2024-08-18
[manuf] modify perso payload to support TLV objects

vbendeb:
This paves the way to support of including various objects in the set of provisioning data on top of X.509 certificates.

A bit of "manual' serialization/deserialization of u16 composites is being introduced, some functions and macors if ft_lib/lib.rs could be improved/converted to methods of their respective header structures. So this PR it he first draft.

This enables addressing #23237.


sw/device/silicon_creator/manuf/base/personalization_data.h
```diff
@@ -0,0 +1,80 @@
...
+#include <stdint.h>
+
+/**
+ * Personalization data is sent between the device and the host during the
+ * device provision. Personalization data is laid out as a sequence of
+ * concatenated LTV objects. The length and type field of the objects are packed
+ * in a 16 bit integer:
+ *  d15                                           d0
+ *  +-------------+--------------------------------+
+ *  | 4 bit type  |   12 bits total  object size   |
```
pamaury, line 16: Is there a risk of a certificate ever been bigger than 4kB? If so, this could become a real issue with this encoding.

pamaury:
That's a pretty big change, I had a look at it and I wonder if there is a reason for using such a compact encoding? It seems to add a lot of code for very little benefit. For example, why not using a single header for all objects like this:
```c
struct tlv_header {
   uint8_t type; // endorsed cert, unendorserd, ...
   uint8_t info; // for cert: indicate which certificate (instead of a using a name string).
   uint16_t length;
} tlv_header_t;
```
This would make the code for uniform, in particular it becomes easier to search for object when receiving data from the host. This is actually more efficient than the above because you don't have to store a string for each certificate and I don't think there is a real need for arbitrary strings for certificates?

sw/device/silicon_creator/manuf/base/personalization_data.h
vbendeb: I don't think this is likely at all: will we ever allow certificates spanning more than two INFO pages?

sw/device/silicon_creator/manuf/base/perso_blob.h
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param name the name of the certificate
+ * @param needs_endorsement defines the type of the LTV object the certificate
+ *              is wrapped into
+ * @param cert_body the actual certificate
+ * @param cert_size size of the certificate in bytes
+ *
+ * @return status of the operation.
+ */
+extern status_t prepare_cert_for_shipping(const char *name,
```
timothytrippel, line 25: I think it would be better if we just had one `perso_ltv_data.h` (or named something like that) lib that combines `personalization_data.h` with this `perso_blob.h`. This lib could contain all the formatting / parsing functions.

This function could take a perso_blob_t as an outparam.

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+#include "perso_blob.h"
+#include "personalization_data.h"
```
timothytrippel, line 6: We use full include paths.

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -327,6 +320,70 @@ static status_t log_hash_of_all_certs(ujson_t *uj) {
...
+  memcpy(perso_blob_to_host.body + perso_blob_to_host.next_free, data, size);
+  perso_blob_to_host.next_free += size;
+
+  return OK_STATUS();
+}
+
+status_t prepare_cert_for_shipping(const char *name, bool needs_endorsement,
+                                   const void *cert_body, size_t cert_size) {
+  // The header is a 16 bit object header followed by 16 bits cert wrapper
+  // header followed by the certificate name string.
```
timothytrippel, line 341: ASCII diagram would help make this clearer.

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -327,6 +320,70 @@ static status_t log_hash_of_all_certs(ujson_t *uj) {
...
+  } else {
+    set_field(Objh, Type, obj_header, kPersoObjectTypeTpmEndorsed);
+  }
+
+  set_field(Objh, Size, obj_header, obj_len);
+
+  set_field(Crth, Size, cert_header, wrapped_len);
+  set_field(Crth, NameSize, cert_header, name_len);
+
+  // Cert size will have to be blended in later.
```
timothytrippel, line 374: ?

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -405,12 +460,162 @@ static status_t personalize_gen_dice_certificates(ujson_t *uj) {
...
+typedef struct cert_block {
+  size_t obj_size;
+  size_t wrapped_cert_size;
+  const void *wrapped_cert_p;
+  char name[kCrthNameSizeFieldMask + 1];
+} cert_block_t;
+
+/**
+ *  Given the pointer to an LTV object, in case this is an endorsed certificate
+ *  set up the cert_block_t structure for it.
```
timothytrippel, line 481: nit: ```suggestion
 * Given the pointer to an LTV object, in case this is an endorsed certificate
 * set up the cert_block_t structure for it.
```

File: sw/device/silicon_creator/manuf/base/ft_personalize.c [hunk 7]
```diff
@@ -405,12 +460,162 @@ static status_t personalize_gen_dice_certificates(ujson_t *uj) {
...
+ *
+ * @param buf pointer to the LTV object storing the certificate
+ * @param max_room total number of bytes til the end of the buffer (the LTV
+ *               object is likely to be smaller, but can't be any bigger)
+ * @param block pointer to the block to set up.
+ *
+ * @return STATUS_OK on success, NOT_FOUND if the object is not an endorsed
+ *              certificate, or the error condition encountered.
+ */
+static status_t set_cert_block(const uint8_t *buf, size_t max_room,
```
timothytrippel, line 491: I think we should move all these functions to the `perso_ltv_data.h` lib.

sw/device/silicon_creator/manuf/base/personalization_data.h
```diff
@@ -0,0 +1,80 @@
...
+ *
+ *  d15                                           d0
+ *  +-------------+--------------------------------+
+ *  | 4 bit length|       12 bits total size       |
+ *  +-------------+--------------------------------+
+ */
+typedef enum cert_header_fields {
+  kCrthSizeFieldShift = 0,
+  kCrthSizeFieldWidth = 12,
+  kCrthSizeFieldMask = (1 << kCrthSizeFieldWidth) - 1,
```
timothytrippel, line 56: separate each field block with a comment of what it is. (same above for generic object)

Also, can we make this more generic, i.e., add another field to the header that includes the cert type (X509 or CWT). In the future, we would like to be able to support other cert formats besides X509 (i.e., CWT certs encoded in CBOR format)

sw/device/silicon_creator/manuf/base/personalization_data.h
```diff
@@ -0,0 +1,80 @@
...
+  kObjhTypeFieldMask = (1 << kObjhTypeFieldWidth) - 1,
+} obj_header_fields_t;
+
+/**
+ * The x509 certificate is prepeneded by a 16 bits header followed by the ASCII
+ * characters of the certificate name, followed by the certificate body.
+ *
+ * The certificate header includes 4 bits for the certificate name length then
+ * the full size of the certificate object (header size + name lenght +
+ * cetificate size).
```
timothytrippel, line 46:
```suggestion
 * the full size of the certificate object (header size + name length +
 * certificate size).
```

sw/device/silicon_creator/manuf/base/personalization_data.h
```diff
@@ -0,0 +1,80 @@
...
+ *  d15                                           d0
+ *  +-------------+--------------------------------+
+ *  | 4 bit type  |   12 bits total  object size   |
+ *  +-------------+--------------------------------+
+ *
+ * The following object types have been defined so far:
+ */
+typedef enum perso_object_type {
+  kPersoObjectTypeTpmNotEndorsed = 0,
+  kPersoObjectTypeTpmEndorsed = 1,
```
timothytrippel, line 23: can we generalize to:
- X.509 tbs cert
- X.509 cert
?

sw/device/silicon_creator/manuf/base/personalization_data.h
timothytrippel: 4k is just an arbitrary max payload size for all data exported. We can up the size, but realistically we want to keep this as small as possible to maximizing provisioning throughput.

sw/host/provisioning/ft_lib/src/lib.rs
```diff
@@ -186,9 +225,163 @@ fn send_rma_unlock_token_hash(
     Ok(())
 }
 
-const CERT_LABELS: [&str; 8] = [
-    "UDS", "CDI_0", "CDI_1", "TPM_EK", "CEK", "CIK", "Unknown", "Unknown",
-];
+// Helper functions used to pack LTV object and Certificate payload headers.
+fn make_obj_header(size: usize, otype: ObjType) -> Result<ObjHeaderType> {
```
timothytrippel, line 229: maybe we could move these to the ujson_lib: https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/ujson_lib/src/lib.rs ? since they are functions used to format data to read/write from/to the ujson LTV payload?

timothytrippel (commented):
Great, first pass @vbendeb ! I dropped a few suggestions. Let's iterate on this. I think will help make the perso process much more generic to support even more SKU owners.

timothytrippel:
CC: @tommychiu-github : just an FYI another refactoring is ongoing to help make the data sent between host and device during perso generic enough to support different types of certificate data and generic meta data.

sw/device/silicon_creator/manuf/base/ft_personalize.c
vbendeb: done

sw/device/silicon_creator/manuf/base/perso_blob.h
vbendeb: done

sw/host/provisioning/ft_lib/src/lib.rs
vbendeb: I can do it if you feel strongly about it, but this  does not seem a particularly good idea to me: so far all users of these structures and functions are in this file, moving these to the ujson lib will require a lot of exports/imports.

vbendeb (commented):
I think I addressed everything, it passes tests including local tpm_ext. I'll update the external one in the meantime.

sw/device/silicon_creator/manuf/base/ft_personalize.c
vbendeb: I might have missed this one, it's not clear what's suggested here.

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -76,28 +77,21 @@ static cert_key_id_pair_t cdi_1_key_ids = {
...
 static ecdsa_p256_public_key_t curr_pubkey = {.x = {0}, .y = {0}};
-static manuf_certs_t tbs_certs = {
-    .num_certs = 3,
-    .next_free = 0,
-    .tbs = {true, false, false, false, false, false, false, false},
-    .certs = {0},
-};
-static manuf_certs_t endorsed_certs;
+static perso_blob_t perso_blob_to_host;   // Perso data device => host.
+static perso_blob_t peso_blob_from_host;  // Perso data host => device.
```
pamaury, line 81: `peso` -> `perso`

pamaury:
> > That's a pretty big change, I had a look at it and I wonder if there is a reason for using such a compact encoding? It seems to add a lot of code for very little benefit. For example, why not using a single header for all objects like this:
> > ```c
> > struct tlv_header {
> >    uint8_t type; // endorsed cert, unendorserd, ...
> >    uint8_t info; // for cert: indicate which certificate (instead of a using a name string).
> >    uint16_t length;
> > } tlv_header_t;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This would make the code for uniform, in particular it becomes easier to search for object when receiving data from the host. This is actually more efficient than the above because you don't have to store a string for each certificate and I don't think there is a real need for arbitrary strings for certificates?
> 
> It seems to me object header needs to be generic, and then the objects could have a structure if necessary (in case of certs) or be just a flat number (in case of dev seed).
> 
> Certificate names in this context are pretty arbitrary and they end up not saved anywhere, I had this feeling that it is easy to get out of sync between the cert order in the names.
> 
> It is a good point that the code could be simpler if there was no need to pack stuff into 16 bit values, so maybe allow 16 bits for object sizes and one byte for object type/cert name length?

I think that if you can avoid having more structure in the data (ie just header + data) and no variable length headers, it's a lot simpler even if it wastes a few bytes. For example, I would use an enum for certificates names instead of strings, this way you don't get a variable length field.
```c
struct tlv_header {
   uint8_t type; // endorsed cert, unendorserd, ...
   uint8_t info; // for cert: indicate which certificate (instead of a using a name string).
   uint16_t length;
} tlv_header_t; // should be packed

typedef tlv_cert_type {
   kCertTypeUds,
   kCertTypeCdi0,
   kCertTypeCdi1,
} tlv_cert_type;
```
This will also make the rust side simpler by avoiding the use of a macro.

sw/device/silicon_creator/manuf/base/ft_personalize.c
vbendeb: done

vbendeb:
>  For example, I would use an enum for certificates names instead of strings, this way you don't get a variable length field.

How would we keep cert names in sync between C and Rust in this case, will bindgen allow to create shared arrays of strings? 

Honestly, I don't see a big problem with the fact that the string length is variable, the cert size is variable to start with and we need to deal with it. I'll let @timothytrippel weigh in on this.

sw/device/silicon_creator/manuf/base/perso_tlv_data.h
```diff
@@ -0,0 +1,112 @@
...
+  kCrthSizeFieldMask = (1 << kCrthSizeFieldWidth) - 1,
+
+  // Length of the certificate name immediately following the header.
+  kCrthNameSizeFieldShift = kCrthSizeFieldWidth,
+  kCrthNameSizeFieldWidth = sizeof(cert_header_t) * 8 - kCrthSizeFieldWidth,
+  kCrthNameSizeFieldMask = (1 << kCrthNameSizeFieldWidth) - 1,
+} cert_header_fields_t;
+
+// Helper macros allowing set or get various header fields.
+#define set_field(type_name, field_name, full_value, field_value)             \
```
timothytrippel, line 75: our style guides on naming macros:
https://google.github.io/styleguide/cppguide.html#Macro_Names
https://opentitan.org/book/doc/contributing/style_guides/c_cpp_coding_style.html#function-enum-struct-and-typedef-naming

applies below too

sw/device/silicon_creator/manuf/extensions/example_personalize_ext.c
```diff
@@ -16,7 +16,7 @@ status_t personalize_extension_pre_cert_endorse(
 }
 
 status_t personalize_extension_post_cert_endorse(
-    ujson_t *uj, manuf_certs_t *endorsed_certs,
+    ujson_t *uj, manuf_certs_t *perso_blob_from_host,
```
timothytrippel, line 19: `manuf_certs_t` type will no longer exist after this PR; replace with `perso_blob_t`?

sw/device/silicon_creator/manuf/base/perso_tlv_data.h
```diff
@@ -0,0 +1,112 @@
...
+
+#include "sw/device/lib/base/status.h"
+#include "sw/device/lib/testing/json/provisioning_data.h"
+
+/**
+ * Personalization data is sent between the device and the host during the
+ * device provision. Personalization data is laid out as a sequence of
+ * concatenated LTV objects. The length and type field of the objects are packed
+ * in a 16 bit integer:
+ *  d15                                           d0
```
timothytrippel, line 20:
```suggestion
 * Personalization data is sent between the device and the host during
 * device provisioning. Personalization data is laid out as a sequence of
 * concatenated LTV objects. The length and type field of the objects are packed
 * in a 16 bit integer:
 *  d15                                           d0
```

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -438,69 +660,89 @@ static status_t personalize_endorse_certificates(ujson_t *uj) {
...
+  while (perso_blob_from_host.num_objs)
+    TRY(extract_next_cert(&next_cert, &free_room));
+
   /*****************************************************************************
    * Save Certificates to Flash.
    ****************************************************************************/
-  size_t all_certs_offset = 0;
+  // This is where the certificates to be copied are stored, each one wrapped in
+  // with its name prepended to it.
+  uint8_t *certs = all_certs;
```
timothytrippel, line 701: Why reassign the all_certs buffer pointer to another pointer? can you not use `all_certs` pointer directly?

sw/host/provisioning/ft_lib/src/lib.rs
```diff
@@ -168,6 +169,44 @@ pub enum KeyWrapper {
...
+// u16.
+type CertHeaderType = perso_objects::cert_header_t;
+struct CertHeader<'a> {
+    // Total size the certificate takes in the buffer, header + hame length +
+    // cert size.
+    wrapped_size: usize,
+    cert_name: &'a str,
+    cert_body: Vec<u8>,
+}
+
```
timothytrippel, line 209: Convention in the OpenTitan repo so far is to group all of the bindgen'd assets into a separate lib and then import that lib where it is used. See an example of the otp_ctrl DIF bindgen'd lib: https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/opentitanlib/src/dif/otp_ctrl.rs;l=1 . I think we should do the same here so this file does not get bloated with a bunch of definitions of object types and enums. This lib is supposed be simply a set of functions required to drive an FT flow from the host side.

timothytrippel:
> I think that if you can avoid having more structure in the data (ie just header + data) and no variable length headers, it's a lot simpler even if it wastes a few bytes. For example, I would use an enum for certificates names instead of strings, this way you don't get a variable length field.
> 
> ```c
> struct tlv_header {
>    uint8_t type; // endorsed cert, unendorserd, ...
>    uint8_t info; // for cert: indicate which certificate (instead of a using a name string).
>    uint16_t length;
> } tlv_header_t; // should be packed
> 
> typedef tlv_object_type {
>    kObjectTypeCertTbs,
>    kObjectTypeCert,
>    kObjectTypeDevSeed,
> } tlv_cert_type;
> 
> // Only for kObjectTypeCertTbs and kObjectTypeCert
> typedef tlv_cert_info {
>    kCertTypeUds,
>    kCertTypeCdi0,
>    kCertTypeCdi1,
> } tlv_cert_type;
> ```
> 
> This will also make the rust side simpler by avoiding the use of a macro.

This is certainly a cleaner approach, but the issue is, we cannot enumerate all the valid certs a SKU owner may want to provision upstream here. We only know about DICE certs and TPM EK cert, but a OpenTitan SKU owner may have a special proprietary cert they want to provision in a provisioning extension. Thus the string approach makes this super flexible. It does not have to be strings, it could be a unique number, but strings a bit more readable.

sw/host/provisioning/ft_lib/src/lib.rs
vbendeb: done(I think)

vbendeb (commented):
Well, this was a massive second  iteration review, I think I addressed most of the comments, but  might still need some style compliance changes

The new patchset is quite different from the original one, sorry...

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -261,10 +262,32 @@ cc_library(
...
+
+cc_library(
+    name = "perso_tlv_headers",
+    hdrs = ["perso_tlv_data.h"],
+    deps = [
+        "//sw/device/lib/base:status",
+        "//sw/device/lib/testing/json:provisioning_data",
+    ],
+)
+
```
timothytrippel, line 290: Why there two libraries here? I think one would suffice:

```suggestion
cc_library(
    name = "perso_tlv_data",
    srcs = ["perso_tlv_data.c"],
    hdrs = ["perso_tlv_data.h"],
    deps = [
        "//sw/device/lib/base:status",
        "//sw/device/lib/runtime:log",
        "//sw/device/lib/testing/json:provisioning_data",
        "//sw/device/silicon_creator/lib/cert",
    ],
)
```

sw/host/provisioning/ft_lib/src/lib.rs
```diff
@@ -8,6 +8,7 @@ use std::time::Duration;
 
 use anyhow::{bail, Result};
 use arrayvec::ArrayVec;
+use bindgen::perso_tlv_objects;
```
timothytrippel, line 11: You still need this include?

timothytrippel (approved):
Just about there, thanks @vbendeb !

timothytrippel:
Looks like you need to rebase again too.

sw/device/silicon_creator/manuf/base/BUILD
vbendeb: This is how I started, but inclusion of  "//sw/device/silicon_creator/lib/cert" upsets  bindgen because of some OTBN stuff this pulls in (at least this is how it looked), so I separated headers, but forgot to clean up `perso_tlv_data` dependencies, which I did now.

sw/host/provisioning/ft_lib/src/lib.rs
vbendeb: yes, without it macro expansion chokes.

vbendeb:
reabsed and re-tested

pamaury:
> > I think that if you can avoid having more structure in the data (ie just header + data) and no variable length headers, it's a lot simpler even if it wastes a few bytes. For example, I would use an enum for certificates names instead of strings, this way you don't get a variable length field.
> > ```c
> > struct tlv_header {
> >    uint8_t type; // endorsed cert, unendorserd, ...
> >    uint8_t info; // for cert: indicate which certificate (instead of a using a name string).
> >    uint16_t length;
> > } tlv_header_t; // should be packed
> > 
> > typedef tlv_object_type {
> >    kObjectTypeCertTbs,
> >    kObjectTypeCert,
> >    kObjectTypeDevSeed,
> > } tlv_cert_type;
> > 
> > // Only for kObjectTypeCertTbs and kObjectTypeCert
> > typedef tlv_cert_info {
> >    kCertTypeUds,
> >    kCertTypeCdi0,
> >    kCertTypeCdi1,
> > } tlv_cert_type;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will also make the rust side simpler by avoiding the use of a macro.
> 
> This is certainly a cleaner approach, but the issue is, we cannot enumerate all the valid certs a SKU owner may want to provision upstream here. We only know about DICE certs and TPM EK cert, but a OpenTitan SKU owner may have a special proprietary cert they want to provision in a provisioning extension. Thus the string approach makes this super flexible. It does not have to be strings, it could be a unique number, but strings a bit more readable.

Ok that makes sense. I still think the header/object encoding is more complicated than necessary but I won't object to the current approach.

PR #24357, opened 2024-08-19
[manuf] Update pre-cert-endorsment perso extension hook

tommychiu-github:
Add one more parameter to the hook for accessing flash device state.

### DIFF
--- a/sw/device/silicon_creator/manuf/base/ft_personalize.c
+++ b/sw/device/silicon_creator/manuf/base/ft_personalize.c
@@ -523,7 +523,8 @@ static status_t personalize_endorse_certificates(ujson_t *uj) {
  */
 extern status_t personalize_extension_pre_cert_endorse(
     ujson_t *uj, manuf_certgen_inputs_t *certgen_inputs,
-    manuf_certs_t *tbs_certs, cert_flash_info_layout_t *cert_flash_layout);
+    manuf_certs_t *tbs_certs, cert_flash_info_layout_t *cert_flash_layout,
+    dif_flash_ctrl_state_t *flash_ctrl_handle);
 
 /**
  * A custom extension to the personalization flow.
@@ -545,7 +546,7 @@ bool test_main(void) {
   CHECK_STATUS_OK(personalize_otp_and_flash_secrets(&uj));
   CHECK_STATUS_OK(personalize_gen_dice_certificates(&uj));
   CHECK_STATUS_OK(personalize_extension_pre_cert_endorse(
-      &uj, &certgen_inputs, &tbs_certs, cert_flash_layout));
+      &uj, &certgen_inputs, &tbs_certs, cert_flash_layout, &flash_ctrl_state));
   CHECK_STATUS_OK(personalize_endorse_certificates(&uj));
   CHECK_STATUS_OK(personalize_extension_post_cert_endorse(&uj, &endorsed_certs,
                                                           cert_flash_layout));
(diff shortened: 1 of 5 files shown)


sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -523,7 +523,8 @@ static status_t personalize_endorse_certificates(ujson_t *uj) {
  */
 extern status_t personalize_extension_pre_cert_endorse(
```
vbendeb, line 524: this is a question for @timothytrippel I guess - why are these externs not in a .h file? 

I got burnt by the linker happily linking external modules which had these functions but a misplaced parameter order.

timothytrippel: Unfortunately, this was a result of bit rot / several refactorings. I agree, we can move this to a header file, but that can be done in a follow up PR.

timothytrippel (approved):

sw/device/silicon_creator/manuf/base/tpm_personalize_ext.c
```diff
@@ -135,7 +135,8 @@ static status_t personalize_gen_tpm_certificates(
 
 status_t personalize_extension_pre_cert_endorse(
     ujson_t *uj, manuf_certgen_inputs_t *certgen_inputs,
-    manuf_certs_t *tbs_certs, cert_flash_info_layout_t *cert_flash_layout) {
+    manuf_certs_t *tbs_certs, cert_flash_info_layout_t *cert_flash_layout,
+    dif_flash_ctrl_state_t *flash_ctrl_handle) {
```
timothytrippel, line 139: You will need to `#include "sw/device/lib/dif/dif_flash_ctrl.h"` at the top for this to compile. Same for the other extension files here.

timothytrippel (approved):

sw/device/silicon_creator/manuf/extensions/BUILD.bazel
```diff
@@ -28,6 +28,7 @@ cc_library(
         "@//sw/device/lib/testing/test_framework:status",
         "@//sw/device/lib/testing/test_framework:ujson_ottf",
         "@//sw/device/silicon_creator/lib/cert",
+        "@//sw/device/silicon_creator/manuf/lib:flash_info_fields",
```
timothytrippel, line 31: Looks like you included the wrong dep here. You should be including `@//sw/device/lib/dif:flash_ctrl`: https://github.com/lowRISC/opentitan/blob/34b8fc33e36b3ed2a32e6cd50b451a47f16b202d/sw/device/lib/dif/BUILD#L324

Same below.

sw/device/silicon_creator/manuf/extensions/BUILD.bazel
```diff
@@ -39,6 +40,7 @@ cc_library(
...
```
timothytrippel, line 43: See above comment.

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -265,6 +265,7 @@ cc_library(
         "//sw/device/silicon_creator/lib/cert:tpm_ek_template_library",
         "//sw/device/silicon_creator/lib/drivers:flash_ctrl",
         "//sw/device/silicon_creator/lib/drivers:hmac",
+        "//sw/device/silicon_creator/manuf/lib:flash_info_fields",
```
timothytrippel, line 268: you don't need this dep. you only need the `"//sw/device/lib/dif:flash_ctrl",` which is already included above.

timothytrippel (approved):

PR #24361, opened 2024-08-20
Update bazel dependencies

jwnrt:
### Summary

This PR updates certain Bazel dependencies to their most recent versions.

This includes:

* `rules_cc`
* `platforms`
* `rules_fuzzing`
* `rules_python`
* `buildifier`

One of the motivations for this PR is that our current versions of these dependencies do not support bzlmod (or do not support it well enough).

Another is that newer versions of Python (specifically 3.12) require a more recent `rules_python`.

### Non-bazel dependencies

This PR also updates `tockloader` to an upstream version which includes our custom patch. This is necessary due to `rules_python` breakage around Git dependencies. `tockloader`'s dependencies are now locked and hashed.

I've had to disable the `jsonschema` dependency from being used in the bitstream script when run under Bazel. This was used to check the layout of the `manifest.json` that comes from the bucket. Changes to `rules_python` have made it very difficult for rules to depend on `pip` dependencies. Outside of Bazel (or if we get it working again) the schema will be checked when `jsonschema` is available.

### Asides

`rules_rust` was not updated in this PR because recent versions require Bazel 7 or higher. We'll need to update this later, but deferring since updating Bazel is a lot of work.

All of these new versions have working `bzlmod` modules in the Bazel Central Registry. There is a commit at the end which renames our `abseil-cpp` repository to match the one provided by `bzlmod` to make this compatibility even closer.


pamaury (approved):
This looks great, thanks @jwnrt

rules/scripts/bitstreams_workspace.py
```diff
@@ -401,10 +399,6 @@ def _ConstructBazelString(self, build_file: Path, key: str) -> str:
             manifest["schema_version"] = MANIFEST_SCHEMA_VERSION
 
         if manifest["schema_version"] == 2:
-            schema_path = os.path.join(MANIFESTS_DIR, "bitstreams_manifest.schema.json")
```
a-will, line 404: Please leave this code in the script. If you need to make jsonschema optional for now, move the import here inside a try..catch, and gate the call behind a feature flag.

third_party/lint/repos.bzl
```diff
@@ -17,19 +17,15 @@ def lint_repos(lowrisc_lint = None):
     # that golang depends on.
     http_archive(
         name = "ot_hack_bazelbuild_buildtools",
```
cfrantz, line 19: If we no longer need to patch this, can we rename the repo to not be `ot_hack_...`?

cfrantz (approved):

rules/scripts/bitstreams_workspace.py
jwnrt: Sure, will do. Out of interest, do you use this script from outside of Bazel?

a-will: Yes, I use it outside bazel primarily for development and testing, but it can also be used to set up an offline cache.

jwnrt: I've added this back behind a try/except but wasn't sure what you mean about a feature flag. Is this so that you'll get an error message if you ask for it to check the schema but don't have `jsonschema` installed?

a-will: I just meant that you'd set some variable representing that schema validation is enabled, but since the only user is right here, maybe it's not necessary. I think this will do. Thanks! :)

third_party/python/pip.bzl
```diff
@@ -2,152 +2,12 @@
...
-        "quiet": attr.bool(
-            default = True,
-            doc = "If True, suppress printing stdout/stderr to the terminal.",
-        ),
-        "timeout": attr.int(
-            default = 300,
-            doc = "Timeout (in seconds) on the rule's execution duration.",
-        ),
-        "_gen_requirements_script": attr.label(
-            default = "//third_party/python:gen_requirements.sh",
```
timothytrippel, line 133: If we no longer need this hack of pre-processing the python-requirements.txt file to support airgapped builds, then I think we can also delete the gen_requirements.sh script too. @jwnrt

jwnrt: Ah yes, thanks, you're right
https://github.com/lowRISC/opentitan/pull/24591

timothytrippel:
Realized, #24372 needed to be cherry-picked over to the `earlgrey_1.0.0` branch first before the cherry-pick automation will work here. Once #24807 merges I will apply the automation label again here.

PR #24366, opened 2024-08-20
[opentitanlib] Refactor RPC module to enable supporting future console interfaces (e.g., SPI)

anthonychen1251:
This PR has a commit that refactors the RPC module to support additional console devices, e.g. the SPI device console.

This PR aims to support a future OTTF SPI  device console (#21726)

### DIFF
--- a/sw/host/opentitanlib/src/test_utils/rpc.rs
+++ b/sw/host/opentitanlib/src/test_utils/rpc.rs
@@ -16,7 +16,7 @@ use crate::uart::console::{ExitStatus, UartConsole};
 // Bring in the auto-generated sources.
 include!(env!("ottf"));
 
-pub trait UartSend<T>
+pub trait ConsoleSend<T>
 where
     T: ConsoleDevice + ?Sized,
 {
@@ -24,7 +24,7 @@ where
     fn send_with_crc(&self, device: &T) -> Result<()>;
 }
 
-impl<T, U> UartSend<T> for U
+impl<T, U> ConsoleSend<T> for U
 where
     T: ConsoleDevice + ?Sized,
     U: Serialize,
@@ -47,7 +47,7 @@ where
     }
 }
 
-pub trait UartRecv<T>
+pub trait ConsoleRecv<T>
 where
     T: ConsoleDevice + ?Sized,
 {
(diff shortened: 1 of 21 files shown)


sw/host/opentitanlib/src/test_utils/rpc.rs
```diff
@@ -9,46 +9,59 @@ use serde::Serialize;
...
+where
+    T: ConsoleDevice + ?Sized,
+{
+    fn send(&self, device: &T) -> Result<()>;
+    fn send_with_crc(&self, device: &T) -> Result<()>;
 }
 
-impl<T: Serialize> UartSend for T {
-    fn send(&self, uart: &dyn Uart) -> Result<()> {
+impl<T, U> UartSend<T> for U
```
timothytrippel, line 27: maybe this should be `ConsoleSend` eventually? not sure how invasive changing this would be? if too much of a pain we could do it in a separate PR

sw/host/opentitanlib/src/test_utils/rpc.rs
```diff
@@ -9,46 +9,59 @@ use serde::Serialize;
...
+    T: ConsoleDevice + ?Sized,
+{
+    fn recv(device: &T, timeout: Duration, quiet: bool) -> Result<Self>
     where
         Self: Sized;
 }
 
-impl<T: DeserializeOwned> UartRecv for T {
-    fn recv(uart: &dyn Uart, timeout: Duration, quiet: bool) -> Result<Self>
+impl<T, U> UartRecv<T> for U
```
timothytrippel, line 59: same here, `ConsoleRecv`?

timothytrippel (approved):

sw/host/opentitanlib/src/test_utils/rpc.rs
anthonychen1251: You're right. I initially planned to rename this in a separate PR, but after reviewing the codebase, the changes seemed less extensive than expected. I've now included a commit for the renaming in this PR.

sw/host/opentitanlib/src/test_utils/rpc.rs
anthonychen1251: Done.

PR #24372, opened 2024-08-20
[ci] Fix false-positive Bazel airgapped build test

jwnrt:
Usually bash sub-scripts inherit these flags from their parent shell, but that doesn't seem to be happening when run via `ip` and `sudo`.

We're seeing cases where the `cquery` fails due to OOM and generates no tests, but then the `build` succeeds because it only has to build the hard-coced `mask_rom` target. This success was used as the exit code for the whole script.

Adding `-e` and `-x` flags ensures the sub-script uses the same flags as the parent script.

### DIFF
--- a/ci/scripts/test-airgapped-build.sh
+++ b/ci/scripts/test-airgapped-build.sh
@@ -20,7 +20,7 @@ sudo ip netns exec airgapped ip addr add 127.0.0.1/8 dev lo
 sudo ip netns exec airgapped ip link set dev lo up
 
 # Enter the network namespace and perform several builds.
-sudo ip netns exec airgapped sudo -u "$USER" bash -c \
+sudo ip netns exec airgapped sudo -u "$USER" bash -e -x -c \
   "export BAZEL_BITSTREAMS_CACHE=$(pwd)/bazel-airgapped/bitstreams-cache;
   export BITSTREAM=\"--offline latest\";
   export BAZEL_PYTHON_WHEELS_REPO=$(pwd)/bazel-airgapped/ot_python_wheels;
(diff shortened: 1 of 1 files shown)


jwnrt:
FYI, I just checked the seven most recent passing CI jobs and every single one of them presented this false positive due to `bazel cquery` getting OOM killed.

I suspect every airgapped build has been failing like this.

When working on Bazel changes I've noticed that in my CI runs the airgapped build times out at two hours. I thought this was something I'd broken, but I think `cquery` is somehow being "fixed" and the job just takes too long when run properly.

If that's the case, we may need to limit the airgapped test to make it pass.

jwnrt:
I've updated this PR to make the test only build `mask_rom` since the `cquery` is not working. This is no change in coverage to our current state, but I don't know how useful it is.

If we want more coverage I think building something like a sival test would be useful since it also has to build `opentitanlib` with all its Rust dependencies and also splice a bitstream.

ci/scripts/test-airgapped-build.sh
```diff
@@ -19,27 +19,16 @@ sudo ip netns add airgapped
...
-  cat \"\${TARGET_PATTERN_FILE}\"
+sudo ip netns exec airgapped sudo -E -u "$USER" \
   bazel-airgapped/bazel build \
-    --distdir=$(pwd)/bazel-airgapped/bazel-distdir \
-    --repository_cache=$(pwd)/bazel-airgapped/bazel-cache \
+    --distdir="${PWD}/bazel-airgapped/bazel-distdir" \
+    --repository_cache="${PWD}/bazel-airgapped/bazel-cache" \
     --define DISABLE_VERILATOR_BUILD=true \
-    --target_pattern_file=\"\${TARGET_PATTERN_FILE}\""
+    //sw/device/silicon_creator/rom:mask_rom
```
nbdd0121, line 32: It might still make sense to keep a target pattern file despite since target being built currently. We previously have multiple manually added targets.

pamaury: I think removing the file is file unless we use a query.

pamaury (approved):
Looks good to me, would be good to double-check with other people if building mask_rom is a good enough test (I think it is).

jwnrt:
I'm going to merge this to unblock other work since it's strictly not worse than our current situation. We can add more to this test in the future.

PR #24373, opened 2024-08-21
[sram_ctrl,dv] Wait with reset after mem_init

nasahlpa:
Currently, some of the sram_ctrl_stress_all fail with a DataKnown_A assertion.

The reason for this failure is that the sram_ctrl_stress_all test, depending on the seed, immediately triggers a reset before the first test is started. However, this reset before the first test interferes with the hardware-based memory initialization feature.
image
More specifically, this reset currently happens when the last write of the mem init is issued, i.e., the write gets not finished. Hence, the memory at the last address is uninitialized, triggering the DataKnown_A assertions.

To circumvent this issue, this commit adds a wait_clks(2) before the reset.


nasahlpa:
I am now opening this PR as `./util/dvsim/dvsim.py hw/ip/sram_ctrl/dv/sram_ctrl_ret_sim_cfg.hjson -i sram_ctrl_stress_all -t vcs` now has 100% passing rate and also the failing seeds related to this issue now pass.

vogelpi (approved):
LGTM, thanks @nasahlpa !

PR #24375, opened 2024-08-21
[dv,usbdev] Changes to improve coverage

alees24:
This PR makes a couple of small changes to improve the coverage results:
- handshake packets were not being collected appropriately for contributing to the functional coverage results.
- manual exclusions are introduce for the line coverage to match the FSM transitions that had already been manually excluded as unreachable. Condition coverage exclusions are added for invalid FIFO operations that we _do not expect to see_.

The third commit simply increases the number of seeds used for one of the stress sequences because otherwise the pass rate will fluctuate between 80% and 100% if a single seed fails.


rswarbrick (approved):
Looks sensible to me.

PR #22948, opened 2024-05-02
[rv_core_ibex] Re-vendor Ibex and reduce number of PRINCE rounds for ICache scrambling

vogelpi:
This PR mainly reduced the number of PRINCE rounds for ICache scrambling to improve the timing. From a security perspective, this is acceptable. The scrambling of the other memory primitives (OTBN IMEM, DMEM, ROM etc.) is not touched.

To enable this change, Ibex needs to be re-vendored.

This is related to #22462.

PR #24376, opened 2024-08-22
[top_earlgrey] Reduce number of PRINCE half rounds for main SRAM scrambling

vogelpi:
This PR contains multiple commits to reduce the number of PRINCE half rounds for the main SRAM scrambling from 3 back to 2 as used for Earlgrey ES. This is done to reduce timing pressure.

The scrambling parameters of the other scrambled memory primitives (retention SRAM, ROM, OTBN IMEM & DMEM) are not touched by this PR.

Note that this PR is currently WIP. It will most likely not pass CI yet as I need to also update the tooling to support backdoor loading of memories using different numbers of scrambling rounds. But this PR is sufficient to allow the PD team kicking off another synthesis run. The RTL changes are okay and I pushed this through lint which hasn't flagged any warnings / errors.

For reference, the PRs where we increased the number of scrambling rounds in the past are:
- https://github.com/lowRISC/opentitan/pull/22425
- https://github.com/lowRISC/opentitan/pull/22948

### DIFF
--- a/hw/ip/sram_ctrl/data/sram_ctrl.hjson
+++ b/hw/ip/sram_ctrl/data/sram_ctrl.hjson
@@ -86,6 +86,13 @@
       expose:    "true",
       default:   "1"
     },
+    { name:      "NumPrinceRoundsHalf",
+      desc:      "Number of PRINCE half rounds for the SRAM scrambling feature",
+      type:      "int",
+      local:     "false",
+      expose:    "true",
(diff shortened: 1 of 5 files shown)


vogelpi:
This PR is superseded by #24447 which also contains the tooling changes. To allow comparing the RTL changes (must not change between this PR and #24447 because the PD team already absorbed this one), I decided to create a new one.

PR #24379, opened 2024-08-22
[third_party] switch OpenOCD to debian mirror

nbdd0121:
Sourceforge often has reliability issues which cause our CI to fail. Switch to debian source mirror as the primary URL to download it.

Debian only mirrors the bz2 compressed version, so the SHA is changed.


nbdd0121:
There was a small issue for the GitHub private CI which has been fixed now. Not sure why Azure pipelines don't run though.

Rebased and both seems to be running

cfrantz (approved):

PR #24382, opened 2024-08-22
[dv,usbdev] Fixes for usbdev_stress_usb_traffic

alees24:
Fix a couple of seed-dependent failures in the stress_usb_traffic sequence; in one case the BFM incorrectly predicted a LinkOutErr and in the other the inability of the DUT to detect and report a bit stuffing error exactly at the end of the received packet was being triggered by randomly-chosen address and endpoint, and the monitor did report a bit stuffing violation when the intention was to introduce a deliberate violation only in the subsequent data packet.

### DIFF
--- a/hw/dv/sv/usb20_agent/usb20_monitor.sv
+++ b/hw/dv/sv/usb20_agent/usb20_monitor.sv
@@ -667,7 +667,9 @@ class usb20_monitor extends dv_base_monitor #(
         else consecutive_ones_count = 0;
       end
     end
-    if (consecutive_ones_count >= 6) begin
+    // See explanation in `usb20_agent_cfg.sv` => the DUT presently does not detect bit stuffing
+    // violations that occur right at the end of the packet.
+    if (!cfg.rtl_limited_bitstuff_detection && consecutive_ones_count >= 6) begin
       `uvm_info(`gfn, $sformatf("Bit stuffing violation detected in %p at bit %d", packet, i),
                 UVM_LOW)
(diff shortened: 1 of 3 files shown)


hw/dv/sv/usb20_agent/usb20_monitor.sv
```diff
@@ -667,7 +667,9 @@ class usb20_monitor extends dv_base_monitor #(
         else consecutive_ones_count = 0;
       end
     end
-    if (consecutive_ones_count >= 6) begin
+    // See explanation in `usb20_agent_cfg.sv` => the DUT presently does not detect bit stuffing
+    // violations that occur right at the end of the packet.
+    if (!cfg.rtl_limited_bitstuff_detection && consecutive_ones_count >= 6) begin
```
rswarbrick, line 672: I'm a bit confused: where is this variable defined? My checkout sees `rtl_limited_bitstuff_recovery`...

alees24: '...bitstuff_recovery' was there before, whilst '...bitstuff_detection' is a new introduction in `usb20_agent_cfg.sv` because they are separate shortcomings in the package reception.

rswarbrick: Duh, serves me right for not reading the rest of the PR properly. Sorry!

rswarbrick (approved):
Thanks for the explanation and sorry for the silly question!

PR #24384, opened 2024-08-22
[dv,usbdev] Improve FIFO testing and coverage

alees24:
Complete the existing `fifo_rst` sequence so that it now exercises the Rx FIFO too.
Introduce an additional more targeted sequence that verifies the response to SETUP/OUT packets with randomized FIFO levels.


hw/ip/usbdev/data/usbdev_testplan.hjson
```diff
@@ -1217,6 +1217,26 @@
...
       tests: ["usbdev_tx_rx_disruption"]
     }
+    {
+      name: fifo_levels
+      desc: '''
+            Verify the DUT response to SETUP/OUT data packets under the various possible states of
+            the Available OUT/SETUP Buffer FIFOs and Rx FIFO.
+
+            - Choose levels for each of the FIFOs at random and populate the FIFOs accordingly,
+              ie. in the case of Av OUT/SETUP buffer FIFOs supply the appropriate number of buffers
```
rswarbrick, line 1227: Trivial nit: Maybe "i.e."?

hw/ip/usbdev/dv/env/seq_lib/usbdev_fifo_levels_vseq.sv
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  // FIFO selection.
+  typedef enum {
+    AvOutFifo,
+    AvSetupFifo,
+    RxFifo
+  } fifo_type_e;
+
+  // Populate a FIFOs with the given number of buffers.
```
rswarbrick, line 45: Nit: I think this should probably be "Populate **the** FIFOs"?

hw/ip/usbdev/dv/env/seq_lib/usbdev_fifo_levels_vseq.sv
```diff
@@ -0,0 +1,180 @@
...
+  // themselves shall never be used.
+  task populate_fifo(fifo_type_e fifo, uint nbuf);
+    for (uint b = 0; b < nbuf; b++) begin
+      bit [BufW-1:0] buffer;
+      // Choose a random buffer number
+      buffer = $urandom_range(0, NumBuffers - 1);
+      case (fifo)
+        AvOutFifo:   csr_wr(.ptr(ral.avoutbuffer.buffer), .value(buffer));
+        AvSetupFifo: csr_wr(.ptr(ral.avsetupbuffer.buffer), .value(buffer));
+        // Writing into the Rx FIFO require packet transmission from the host.
```
rswarbrick, line 57: Nit: I think this should be "requires"

hw/ip/usbdev/dv/env/seq_lib/usbdev_fifo_levels_vseq.sv
```diff
@@ -0,0 +1,180 @@
...
+      ResponseNone: check_no_response();  // Expects a response from the usb20_driver, but not DUT.
+      ResponseAck:  check_response_matches(PidTypeAck);
+      ResponseNak:  check_response_matches(PidTypeNak);
+      default:      `uvm_fatal(`gfn, "Invalid/unsupported response type")
+    endcase
+
+    // May as well check the FIFO levels too.
+    csr_rd(.ptr(ral.usbstat.av_setup_depth), .value(act_avsetup_level));
+    csr_rd(.ptr(ral.usbstat.av_out_depth), .value(act_avout_level));
+    csr_rd(.ptr(ral.usbstat.rx_depth), .value(act_rx_level));
```
rswarbrick, line 175: I think it probably makes sense to do this as a single read and then look at the fields of the result. Maybe mirror explicitly and then use the `gmv` macro?

hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
```diff
@@ -165,6 +165,11 @@
       // Short sequence, many possible endpoint configurations.
       reseed: 200
     }
+    {
+      name: usbdev_fifo_levels
+      uvm_test_seq: usbdev_fifo_levels_vseq
+      reseed: 500
```
rswarbrick, line 171: Is 500 really the count we need here?

hw/ip/usbdev/dv/env/seq_lib/usbdev_fifo_levels_vseq.sv
alees24: Ok.

hw/ip/usbdev/dv/env/seq_lib/usbdev_fifo_levels_vseq.sv
alees24: Well, 'Populate a FIFO with' since it's touching only one FIFO per invocation, thanks.

hw/ip/usbdev/dv/usbdev_sim_cfg.hjson
alees24: Well, it's relying upon randomization to hit coverage bins, and there were 192 possibilities. Each sequence itself is very short, but I've looked again at the coverage bins and noted that it was never the intent to exercise the RX FIFO with a level of 1 or the Av FIFOs (from which buffers are pulled by the DUT) with a level of 'depth - 1' since these are not corner cases,  so we're down to 81 possibilities. I've dropped the seed count to 160 accordingly.

rswarbrick (approved):
Sorry for the nitty comments! Looks good to me.

PR #24387, opened 2024-08-22
[ownership] Initialize SRAM & rescue according to the owner configuration

cfrantz:
1. Configure SRAM execution as specified in the owner configuration.
2. Configure rescue according to the owner configuration.
2a. Set bounds on the firmware rescue region in flash.
2b. Disallow rescue modes not allowed by the configuration.
2c. Disallow boot_svc commands not allowed by the configuration.
3. Correct an endian issue with the rescue-mode FourCC codes.  Rescue-modes are expressed in big-endian order, whereas all other FourCC codes are expressed in little-endian order (the reason is that rescue-modes should be thought of more as 4-byte strings than identifier words).

### DIFF
--- a/sw/device/silicon_creator/rom_ext/rescue.c
+++ b/sw/device/silicon_creator/rom_ext/rescue.c
@@ -6,11 +6,14 @@
 
 #include "sw/device/lib/arch/device.h"
 #include "sw/device/lib/base/memory.h"
+#include "sw/device/silicon_creator/lib/boot_data.h"
 #include "sw/device/silicon_creator/lib/dbg_print.h"
 #include "sw/device/silicon_creator/lib/drivers/flash_ctrl.h"
 #include "sw/device/silicon_creator/lib/drivers/retention_sram.h"
 #include "sw/device/silicon_creator/lib/drivers/rstmgr.h"
 #include "sw/device/silicon_creator/lib/drivers/uart.h"
+#include "sw/device/silicon_creator/lib/ownership/datatypes.h"
+#include "sw/device/silicon_creator/lib/ownership/owner_block.h"
 #include "sw/device/silicon_creator/lib/xmodem.h"
 
 #include "flash_ctrl_regs.h"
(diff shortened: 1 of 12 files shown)


sw/device/silicon_creator/rom_ext/rescue.c
```diff
@@ -32,32 +35,37 @@ rom_error_t flash_firmware_block(rescue_state_t *state) {
...
+      bootdata->ownership_state == kOwnershipStateLockedUpdate ||
+      bootdata->ownership_state == kOwnershipStateUnlockedEndorsed) {
+    HARDENED_RETURN_IF_ERROR(flash_ctrl_info_erase(
+        &kFlashCtrlInfoPageOwnerSlot1, kFlashCtrlEraseTypePage));
+    HARDENED_RETURN_IF_ERROR(flash_ctrl_info_write(
+        &kFlashCtrlInfoPageOwnerSlot1, 0,
+        sizeof(state->data) / sizeof(uint32_t), state->data));
+  } else {
+    dbg_printf("error: cannot accept owner_block in current state\r\n");
+  }
```
moidx, line 68: I think it would be good to have a bit of extra hardening here given that faulting this condition would result in a vulnerability. @johannheyszl can calculate the risk potential.

Adding redundant checks before performing flash erase seem a bit complicated from an FI perspective. What if we try to set a flag after performing the flash write, and then before returning at the end of the function perform a redundant check based on the state of the flag. 

```
hardened_bool_t updated_flash = kHardenedBoolFalse;

if (critical block) {
  // flash erase & write
  updated_flash = kHardenedBoolFalse;
}
...

if (launder32(updated_flash) == kHardenedBoolFalse) {
  // read back flash check.
  // clear flash if it doesn't match.
}
return kErrorOk;
```

sw/device/silicon_creator/rom_ext/rescue.c
```diff
@@ -282,8 +303,20 @@ static rom_error_t protocol(rescue_state_t *state) {
...
-rom_error_t rescue_protocol(void) {
-  rom_error_t result = protocol(&rescue_state);
+rom_error_t rescue_protocol(boot_data_t *bootdata,
+                            const owner_rescue_config_t *config) {
+  rescue_state.config = config;
+  if ((hardened_bool_t)config == kHardenedBoolFalse) {
+    // If there is no rescue config, then the rescue region starts immediately
+    // after the ROM_EXT and ends at the end of the flash bank.
+    rescue_state.flash_start = CHIP_ROM_EXT_SIZE_MAX;
+    rescue_state.flash_limit = kFlashBankSize;
```
moidx, line 313: This condition also seems good candidate for hardening, assuming the Owner may want to store critical data in the keep out area. The comparison itself is probably not subject to faults, but we may want to consider a redundant check in case a skip instruction can be mounted.

moidx (approved):

sw/device/silicon_creator/rom_ext/rescue.c
cfrantz: I'm game for hardening this, but we should understand if its really needed.

Uploading the owner config here is only a staging operation.  If an attacker is able to defeat the `if` condition and upload a configuration to page 1, it still has to be validated on next boot.

The potential ownership states are either `LockedOwner` or `Recovery` (if the chip were in one of the unlocked states, there'd be no reason to try to defeat this `if` statement).
- In `LockedOwner`, the owner page has to be validated by [`locked_owner_init`](https://github.com/lowRISC/opentitan/blob/earlgrey_es_sival/sw/device/silicon_creator/lib/ownership/ownership.c#L57-L98).  That function requires one of the two pages to be sealed to be valid, and if one of the pages is invalid, it erases the invalid page and copies the valid page there.  If neither page is sealed, the ownership state transitions to `Recovery`.
- In `Recovery` the chip refuses to honor the contents of the owner pages.

_If_ the owner can thunk the `ownership_state` in `ownership_init` on the next boot to one of the unlocked states, _then_ owner page 1 will be honored (if it is signed).  We should probably harden the switch/case in `ownership_init`.

sw/device/silicon_creator/rom_ext/rescue.c
cfrantz: Done.

sw/device/silicon_creator/rom_ext/rescue.c
moidx: Hi @cfrantz, this explanation makes sense. I agree with your assessment.

moidx (approved):
LGTM

timothytrippel (approved):

PR #24389, opened 2024-08-23
[personalization_ext] Improve personalization extensions API

vbendeb:
Declare interface functions in a .h file so that they can be validated at compile time and move all passed in parameters into the respective structures. Not all extesnions use all interface parameters, keeping paramters in structures will allow to add new interface parameters without modifying extensions which do not use the new parameters.

Tested by running various personalization end to end tests both local and with extensions.

### DIFF
--- a/sw/device/silicon_creator/manuf/base/BUILD
+++ b/sw/device/silicon_creator/manuf/base/BUILD
@@ -212,9 +212,16 @@ filegroup(
 )
 
 cc_library(
+    name = "personalize_ext",
+    hdrs = ["personalize_ext.h"],
+)
+
+cc_library(
     name = "ft_personalize_base",
     srcs = ["ft_personalize.c"],
     deps = [
+        ":perso_tlv_data",
+        ":personalize_ext",
         "//sw/device/lib/crypto/drivers:entropy",
         "//sw/device/lib/dif:flash_ctrl",
         "//sw/device/lib/dif:lc_ctrl",
@@ -241,7 +248,6 @@ cc_library(
         "//sw/device/silicon_creator/lib/drivers:hmac",
         "//sw/device/silicon_creator/lib/drivers:keymgr",
         "//sw/device/silicon_creator/lib/drivers:kmac",
-        "//sw/device/silicon_creator/manuf/base:perso_tlv_data",
         "//sw/device/silicon_creator/manuf/lib:flash_info_fields",
         "//sw/device/silicon_creator/manuf/lib:individualize_sw_cfg_earlgrey_sku_sival",
         "//sw/device/silicon_creator/manuf/lib:personalize",
@@ -252,6 +258,7 @@ cc_library(
     name = "tpm_perso_fw_ext",
     srcs = ["tpm_personalize_ext.c"],
     deps = [
+        ":personalize_ext",
         "//sw/device/lib/dif:flash_ctrl",
         "//sw/device/lib/runtime:log",
         "//sw/device/lib/testing/test_framework:status",
--- a/sw/device/silicon_creator/manuf/base/personalize_ext.h
+++ b/sw/device/silicon_creator/manuf/base/personalize_ext.h
@@ -1,0 +1,87 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef OPENTITAN_SW_DEVICE_SILICON_CREATOR_MANUF_BASE_PERSONALIZE_EXT_H_
+#define OPENTITAN_SW_DEVICE_SILICON_CREATOR_MANUF_BASE_PERSONALIZE_EXT_H_
+
+#include "sw/device/lib/base/status.h"
+#include "sw/device/lib/dif/dif_flash_ctrl.h"
+#include "sw/device/lib/testing/json/provisioning_data.h"
+#include "sw/device/silicon_creator/lib/cert/cert.h"
+
+/**
+ * Parameters passed to personalization extension function invoked before data
+ * is sent to the host for endorsement. Not all parameters are necessarily used
+ * by all extension implementations.
+ */
+typedef struct personalize_extension_pre_endorse {
+  /**
+   *  Serialization interface, marshals/unmarshals structures' fields and
+   *  communicates with the host.
+   */
+  ujson_t *uj;
+  /**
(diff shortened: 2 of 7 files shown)


vbendeb:
A noop change improving personalization extension API robustness.

vbendeb:
@tommychiu-github somehow I can't add Tommy to the reveiwers list explicitly, maybe this works.

sw/device/silicon_creator/manuf/extensions/example_personalize_ext.c
```diff
@@ -8,6 +8,7 @@
 #include "sw/device/lib/testing/test_framework/status.h"
 #include "sw/device/lib/testing/test_framework/ujson_ottf.h"
 #include "sw/device/silicon_creator/lib/cert/cert.h"
+#include "sw/device/silicon_creator/manuf/base/personalize_ext.h"
```
timothytrippel, line 11: I think you need to modify the inputs to the two perso extension functions in this file too.

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -214,6 +214,7 @@ filegroup(
 cc_library(
     name = "ft_personalize_base",
     srcs = ["ft_personalize.c"],
+    hdrs = ["personalize_ext.h"],
```
timothytrippel, line 217: I think what you really want to do is make the header file its own lib like:
```
cc_library(
    name = "personalize_ext",
    hdrs = ["personalize_ext.h"],
)
```

Then include it as a dep of `ft_personalize_base` and all the extension libs.

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -251,7 +252,9 @@ cc_library(
 cc_library(
     name = "tpm_perso_fw_ext",
     srcs = ["tpm_personalize_ext.c"],
+    hdrs = ["personalize_ext.h"],
     deps = [
+        ":ft_personalize_base",
```
timothytrippel, line 257: If you make my suggested change above, then this would be `personalize_ext`, and you would not need the hdrs bit above, like so:
```suggestion
    deps = [
        ":personalize_ext",
```

sw/device/silicon_creator/manuf/extensions/BUILD.bazel
```diff
@@ -29,6 +29,8 @@ cc_library(
         "@//sw/device/lib/testing/test_framework:status",
         "@//sw/device/lib/testing/test_framework:ujson_ottf",
         "@//sw/device/silicon_creator/lib/cert",
+        "@//sw/device/silicon_creator/manuf/base:ft_personalize_base",
+        "@//sw/device/silicon_creator/manuf/lib:flash_info_fields",
```
timothytrippel, line 33: After making the new cc_lib I mention above, then do:
```suggestion
        "@//sw/device/silicon_creator/manuf/base:personalize_ext",
```

sw/device/silicon_creator/manuf/extensions/BUILD.bazel
```diff
@@ -41,6 +43,8 @@ cc_library(
...
```
timothytrippel, line 47: same as above

sw/device/silicon_creator/manuf/base/personalize_ext.h
```diff
@@ -0,0 +1,87 @@
...
+   */
+  manuf_certgen_inputs_t *certgen_inputs;
+  /**
+   *  Pointer to the TLV data blob sent to the host for endorsement.
+   */
+  perso_blob_t *perso_blob_to_host;
+  /**
+   * Pointer to the flash layout table, the extension could enable disabled
+   * entries in the table and configure its INFO space use so that the caller
+   * knows where to endorsed objects received from the host.
```
timothytrippel, line 35:
```suggestion
   * knows where to place endorsed objects received from the host.
```

sw/device/silicon_creator/manuf/base/personalize_ext.h
```diff
@@ -0,0 +1,87 @@
...
+   *  communicates with the host.
+   */
+  ujson_t *uj;
+  /**
+   *  Pointer to the TLV data blob received from the host.
+   */
+  perso_blob_t *perso_blob_from_host;
+  /**
+   * Pointer to the flash layout table used previously when saving pre
+   * indorsement data..
```
timothytrippel, line 61:
```suggestion
   * endorsement data.
```

tommychiu-github:
> @tommychiu-github somehow I can't add Tommy to the reveiwers list explicitly, maybe this works.

Thanks for sharing me this change.
I'll need to update pixel sku to adopt this mechanism as well.

sw/device/silicon_creator/manuf/extensions/example_personalize_ext.c
vbendeb: done.

sw/device/silicon_creator/manuf/base/BUILD
vbendeb: done

sw/device/silicon_creator/manuf/base/BUILD
vbendeb: done

sw/device/silicon_creator/manuf/extensions/BUILD.bazel
vbendeb: I thought I included flash_info_fields for a reason, but seems to be not necessary...

done

sw/device/silicon_creator/manuf/extensions/BUILD.bazel
vbendeb: done

sw/device/silicon_creator/manuf/base/personalize_ext.h
vbendeb: done

vbendeb:
PTAL (and I understand that if I change an object in a BUILD file I might have to edit other build files ;)

timothytrippel (approved):
LGMT, thanks @vbendeb

timothytrippel:
Looks like you have some lint errors though.

vbendeb:
> Looks like you have some lint errors though.

I could find only one error even though the builder shows two. One of the other logs seems completely innocuous. 

I now added a pre-upload check for include guards, updated, let's see if there are more lint issues.

PR #24216, opened 2024-08-05
[experiment] Migrate to bzlmod

jwnrt:
Expecting tonnes of failures here but wanted to see what breaks in CI while it's quiet. I know of a few breakages already around airgapped builds, the bitstream cache, and the C toolchain.

This will be broken into a few PRs later, but it includes:

* Updating Bazel to 7.1.1.
* Updating most dependencies.
* Changing how airgapped builds work (due to changes in 7.1.1).
* Changing how Rust dependencies work (due to changes in `rules_rust`).
* Changing how Python dependencies work (due to changes in `rules_python`).
* Migrating `WORKSPACE` to `MODULE.bazel`.

The lowrisc toolchain and closed-source manufacturing hooks are not yet migrated.

Problems found so far by this PR:

* [x] `rules_python` now refuses to run as `root`, we may need to use rootless containers in CI.
  * For now I've added the `ignore_root_user_error = True` param which disables `.pyc` caching.
* [x] Bazel cannot find `serde_annotate_derive` for some reason. This was working before I rebased so I must have broken something.
  * Nevermind, this is a `rules_rust` regression https://github.com/bazelbuild/rules_rust/issues/2777.
* [x] Airgapped build takes way too long (2 hours before failure).
  * Turned out to be a problem with the airgapped build itself, see https://github.com/lowRISC/opentitan/pull/24372.
* [x] Bitstream rules cannot load python dependencies.
  * Worked around in #24361 by making the `jsonschema` dependency optional.
* [x] License checker rule is racing internally, sometimes the python executable is there and sometimes it's not.
  * Observed again in https://github.com/lowRISC/opentitan/pull/24392 which indicates it's something to do with Bazel 7 and not Bzlmod.
  * Fixed by https://github.com/lowRISC/misc-linters/pull/39.

PR #24392, opened 2024-08-23
[bazel] Update Bazel to version 7.3.1

jwnrt:
This is a major version upgrade from Bazel 6.1.1 to Bazel 7.3.1. This is the latest version.

This PR includes a couple of fixes required due to changes between versions:

1. The airgapped build no longer clones the Bazel Git repository to access its dependencies, it now uses the method described here in the Bazel documentation: <https://bazel.build/run/build#repository_cache_with_bazel_7_or_later>.
2. The `fusesoc_build` rule no longer outputs the whole of fusesoc's `$X.build/` directory, it now only exposes the parts explicitly listed in `output_groups`. This is required because Bazel 7 does not allow rules to output both a directory and its contents at once.

This PR is motivated in part because some of our dependencies (e.g. `rules_rust`) no longer support version 6. Migrating to Bzlmod will also require Bazel 7 since version 6 is missing a couple of features we need.

## Potential for breakage

CI should pick up any breakages, but we should also check that contributors using non-Ubuntu distributions can still use this new Bazel version.

Contributors using `./bazelisk.sh` should not need to manually install the new Bazel version, it should be installed automatically. Those using a native Bazel will have to update, but Bazel will read the `.bazelversion` file and tell them how to do this.


jwnrt:
I was too hasty in marking as ready for review. The strange `license_checker` breakage observed in #24216 has appeared here. I'll try work out what's going on.

rules/fusesoc.bzl
```diff
@@ -69,7 +68,7 @@ def _fusesoc_build_impl(ctx):
         "--setup",
         "--build",
     ])
-    args.add(out_dir.path, format = "--build-root=%s")
+    args.add("--build-root=%.")
```
a-will, line 71: Isn't this going to change the path to all the output files? `dirname` is now used for the output groups, but not fusesoc's actual build-root.

rules/fusesoc.bzl
```diff
@@ -26,12 +26,11 @@ def _corefiles2rootarg(core):
 
 def _fusesoc_build_impl(ctx):
     dirname = "build.{}".format(ctx.label.name)
-    out_dir = ctx.actions.declare_directory(dirname)
     flags = [ctx.expand_location(f, ctx.attr.srcs) for f in ctx.attr.flags]
-    outputs = [out_dir]
+    outputs = []
```
a-will, line 30: If we're getting rid of the directory here, we should probably add at least the routed design checkpoint (DCP) to all of the bitstream build targets' outputs. Otherwise, we won't be able to readily analyze the quality of results and adjust constraints.

rules/fusesoc.bzl
jwnrt: I've changed the implementation to leave the build directory alone and create symlinks into it from a separate `out` directory that the output groups can use.

Let me know what you think

jwnrt: Oops, my symlinks are pointing to the wrong place. I'll fix this in the next push.

jwnrt: You're right, I've worked around this and kept the path the same.

rules/fusesoc.bzl
jwnrt: I've found a couple more paths from the bitstream build that are used, e.g. `ci/fpga-template.yml` tries to read some `runme.log` files which are no longer present.

I'll think about what to do with these

jwnrt:
The `sw_build` job in CI is finding a bunch of problems with our C code now. Not sure what changed in Bazel 7 to make these identified or why only `sw_build` is finding them. They don't appear when I `bazel build` the targets separately.

Some examples of new warnings (which it treats as errors):

1. Missing newlines at the end of files (e.g. the generated `hw/ip/uart/data/uart_regs.h`.
2. Semicolons repeated inside and outside a macro (e.g. with `sw/device/lib/arch/device_fpga_cw305.c` using `CALCULATE_UART_NCO`).
3. Enum values being larger than an `int`. We do this *a lot*. Usually it's for `u32`s.
   * Apparently ISO C requires they fit into `int` but GCC promises to using `u32` if not signed. Not sure what Clang promises.
   * Not sure how to fix this. We use `enum`s everywhere for constants. Do we change the style guide? Ignore this warning?

a-will:
> The `sw_build` job in CI is finding a bunch of problems with our C code now. Not sure what changed in Bazel 7 to make these identified or why only `sw_build` is finding them. They don't appear when I `bazel build` the targets separately.
>

Wrong target, probably (or more accurately, the wrong configuration). This target triggers errors, for example:

```
./bazelisk.sh build //sw/device/silicon_creator/manuf/base:ft_provision_fpga_cw340_rom_with_fake_keys
```

> Some examples of new warnings (which it treats as errors):
> 
> 1. Missing newlines at the end of files (e.g. the generated `hw/ip/uart/data/uart_regs.h`.
> 2. Semicolons repeated inside and outside a macro (e.g. with `sw/device/lib/arch/device_fpga_cw305.c` using `CALCULATE_UART_NCO`).
> 3. Enum values being out of range for `int`. We do this _a lot_. Usually it's for `u32`s.
>    
>    * Apparently ISO C requires they fit into `int` but GCC promises to using `u32` if not signed. Not sure what Clang promises.
>    * Not sure how to fix this. We use `enum`s everywhere for constants. Do we change the style guide? Ignore this warning?

If you compare the subcommands, you'll notice that `-Wpedantic` got added. If we're going to use this style, we really should either update to C23 or use one of the GNU --std modes.

jwnrt:
@pamaury and I found the cause of the pedantic errors: Bazel has changed `--features` to only apply to the `exec` configuration and not `host`. We set `--features=-pedantic_warnings` in `.bazelrc`, but bindgen builds our software with `host` so this wasn't being applied.

Added a commit which adds `--host_features` flags to `.bazelrc.

We also found that the `rules_rust` we're currently using does not seem to properly support Bazel 7. `bazel query 'deps(//some/rust/target)'` was giving an unusual error:

```
$ bazel query 'deps(//sw/device/silicon_creator/manuf/base:ft_provision_fpga_cw340_rom_with_fake_keys)'

ERROR: Evaluation of query "deps(//sw/device/silicon_creator/manuf/base:ft_provision_fpga_cw340_rom_with_fake_keys)" failed: preloading transitive closure failed: no such package '@@rules_rust_util_import__syn-1.0.82//': The repository '@@rules_rust_util_import__syn-1.0.82' could not be resolved: Repository '@@rules_rust_util_import__syn-1.0.82' is not defined
```

To fix this, I have had to bundle the `rules_rust` upgrade into this PR which increases its size but hopefully we can still get this reviewed.

We cannot upgrade `rules_rust` first in a separate PR because new versions do not support Bazel 6. It's possible we could find a version that supports 6 and 7 which I will investigate if this PR is too large to review / merge.

jwnrt:
Sorry for the spam, we found a simpler way to fix the Rust issue without upgrading the toolchain just yet.

rules/fusesoc.bzl
jwnrt: I've replaced the implementation with a symlink one:

* The `build` directory is left alone and included in the outputs.
* Symlinks are created in an `out` directory pointing into `build`.
* The symlinks are used as the `output_groups`.

jwnrt:
Hmm, using symlinks doesn't seem to work because `ctx.actions.symlink` does not have an `inputs` parameter so Bazel just creates the symlink without building the thing it's linking to.

I guess `target_file` is supposed to be like `inputs`, but as previously mentioned we can't create a file to the targets because they're within a directory that is being output.

I can't think of a way to make this work, so reverting back to the previous implementation.

rules/fusesoc.bzl
jwnrt: Update: had to revert the symlink idea, will think of some other way to resolve this.

a-will: For what it's worth, it seems fine to get rid of the build directory as an output for DefaultInfo, but at each fusesoc_build() target, we'll want to be less lazy and add the other specific files we need to keep, hehe. I think the directory-as-output was intended more for generated files with unpredictable names, so we're probably abusing it.

We could also modify our Tcl scripts (and possibly fusesoc core file) to rename the output files to include the label in the Bazel BUILD file. That's just an aesthetic change, though.

jwnrt: I've added the `runme.log` and final DCP file to the outputs of our Vivado rules. Hopefully this is enough. I had problems with name collisions adding the `impl_1/runme.log` file so have removed this one for now. It will take some slightly larger changes to fix that.

hw/bitstream/vivado/BUILD
```diff
@@ -103,6 +107,10 @@ fusesoc_build(
         "bitstream": ["synth-vivado/lowrisc_systems_chip_earlgrey_cw310_hyperdebug_0.1.bit"],
         "rom_mmi": ["synth-vivado/rom.mmi"],
         "otp_mmi": ["synth-vivado/otp.mmi"],
+        "logs": [
+            "synth-vivado/lowrisc_systems_chip_earlgrey_cw310_hyperdebug_0.1.runs/synth_1/runme.log",
```
nbdd0121, line 111: Would it make sense to output the whole run directory?

pamaury: we could potentially make the output_group parameter more general so it's told if it's a file or a directory (or even a directory that you want to tar). But unless there is a specific usage for it, I don't know if there is much point in adding support for it with no user.

jwnrt: I've added this, seems to work fine

pamaury (approved):
That's great, thanks a lot @jwnrt for this work!

nbdd0121 (approved):

PR #24394, opened 2024-08-23
[ci] add a label to rerun workflows

nbdd0121:
This allow members with triage access but not commit access to apply a "CI:Rerun" label on pull request to trigger rerun of workflow.

This would trigger both public and prviate CI workflows.

### DIFF
--- a/.github/workflows/rerun.yml
+++ b/.github/workflows/rerun.yml
@@ -1,0 +1,62 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Rerun Failed Workflow
+on:
+  pull_request_target:
+    types: [labeled]
+
+permissions:
+  actions: write
+  contents: write # For repository dispatch
+  pull-requests: write # For label removal
+
+jobs:
+  rerun:
+    name: Rerun Failed CI
+    if: github.event.label.name == 'CI:Rerun'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Rerun failed GitHub actions
+        uses: actions/github-script@v7
+        with:
+          script: |
+            for await (const response of github.paginate.iterator(
+              github.rest.actions.listWorkflowRunsForRepo,
+              {
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                // Only return completed workflows, not queued and in_progress.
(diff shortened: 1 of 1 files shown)


nbdd0121:
I had a test PR here: https://github.com/nbdd0121/opentitan/pull/8 and you can see the rerun is triggered https://github.com/nbdd0121/opentitan/actions/runs/10524855578 and https://github.com/nbdd0121/opentitan-private-ci/actions/runs/10524857237 (check that the reruns are both triggered by bot)

.github/workflows/rerun.yml
```diff
@@ -0,0 +1,62 @@
...
+                await github.rest.actions.reRunWorkflowFailedJobs({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  run_id: run.id,
+                });
+              }
+            }
+
+            // Trigger a cross-repo-ci-rerun event for private CI
+            github.rest.repos.createDispatchEvent({
```
jwnrt, line 47: Does this receive a response code from the other repo if it failed to retrigger?

And if it did fail, does this throw an exception or anything?

nbdd0121: Repository dispatch is like a message queue, it's asynchronous and there's way to get back response.

The error will be visible in Google Cloud Error Reporting.

jwnrt: Hmm, okay, but the label will still be removed and there will be no error message in the pull request?

It's okay if this is difficult to add, but would like to know.

jwnrt:
Not to give you more work, but is there any way to receive a webhook when the GitHub UI's "rerun failed jobs" button is pressed?

I had a quick look in the docs and think you can get webhooks when an actual workflow has been triggered by the rerun button, but I'm guessing that's not going to fire when the failure was in the private repo.

nbdd0121:
What do you mean by "rerun failed jobs" button? Isn't it associated with a particular run? Is there a button that allows you rerun all workflow runs for a PR?

jwnrt:
Yes, but that may be a shortcut to sending a rerun to each individual workflow

jwnrt:
Actually I'm mistaken, it's one button to run all jobs within a workflow but there's no button to rerun all workflows.

Let's forget about this and just focus on the labels

jwnrt (approved):

engdoreis (approved):

jwnrt:
I've added the `CI:Rerun` label

PR #24395, opened 2024-08-23
[rom_ext/e2e] Define _manifest_address_translation in linker script

pamaury:
This symbol is expected by the code and the test fails to compile without it. This was not caught by the CI because it apparently does not run in CI.


cfrantz (approved):

cfrantz:
This probably needs to be cherry-picked to `earlgrey_es_sival` as well.

pamaury:
The ES branch doesn't seem to have the e2e tests and the manifest translation symbol

pamaury:
Apparently this symbol was already there on the ES branch. I guess the problem is that there was an error when backporting to master.

Issue #23339, opened 2024-05-28
[pattgen,dif] Add support for inactive levels


pamaury:
### Description

#23198 added support for inactive levels in the RTL. The DIF should be updated to support this capability.

timothytrippel:
@pamaury: Is this fixed by #24397?

pamaury:
Yes

PR #24397, opened 2024-08-23
[sival,pattgen] Port pattgen_ios test to master

pamaury:
This is a port of #23164 to master. The big difference between the ES and master is the inactive level feature was added in #23198 . I elected to cherry-pick the commits of #23164 (+some rebase fixes) and then implement the inactive feature on top of it to make the review easier.


pamaury:
Fixes #23339

sw/host/opentitanlib/src/test_utils/gpio_monitor.rs
```diff
@@ -95,6 +95,14 @@ impl<'a> GpioMon<'a> {
         self.initial_levels.clone()
     }
 
+    pub fn all_events(&self) -> Vec<MonitoringEvent> {
+        self.events.clone()
+    }
+
+    pub fn timestamp_to_ns(&self, timestamp: u64) -> u64 {
+        (timestamp - self.initial_timestamp) * 1000000000u64 / self.resolution
```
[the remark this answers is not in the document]
pamaury: If there are no other changes requested, I think I won't change it to avoid another run of the CI.

sw/device/tests/pattgen_ios_test.c
```diff
@@ -0,0 +1,276 @@
...
+
+OTTF_DEFINE_TEST_CONFIG(.enable_uart_flow_control = true);
+
+enum {
+  kTestCmdTimeoutUsec = 5000000,
+};
+
+/* Macro to define and use variable that need to be stored in flash
+ * for DV tests and RAM for real tests, to enable backdoor access. */
+#define DEFINE_BACKDOOR_VAR(type, name, default_val)                       \
```
timothytrippel, line 41: This seems like something that would be useful in `sw/device/lib/base/macros.h`, thoughts?

timothytrippel: (OK to do in follow up PR)

timothytrippel (approved):

sw/device/tests/pattgen_ios_test.c
pamaury: Yes that's a good idea, we probably want to revisit how the backdoor variables are handled, but it's better done in another PR to convert all tests.

PR #24401, opened 2024-08-23
[pentest] Add OTBN SCA tests

nasahlpa:
This PR consists of three commits:
- In the first commit, new OTBN SCA tests are added
- In the second commit, the sca_lib is renamed to pentest_lib
- Based on the second change, in the third change, the pentest_call_and_sleep function is added


johannheyszl (approved):
Thanks @nasahlpa - This has proven useful in testing

vogelpi (approved):

For the record: This is not production code but code that was used for penetration testing of Earlgrey ES. We should not implement changes to these tests that might have an impact on the test results. Instead the aim is to integrate the code more or less as is to be able to reproduce results and to have it ready for future penetration testing campaigns. The first PR of this series (https://github.com/lowRISC/opentitan/pull/24294) has been reviewed by the SW team. The suggested style changes have also been integrated into this PR.

PR #24380, opened 2024-08-22
[ottf,opentitanlib,test] Impl spi device ottf tx console

anthonychen1251:
This PR has three commits that:

1. Update the opentitanlib SPI console module to support TX direction
2. Add support for reading data from the Host via the OTTF SPI device console
3. Test the OTTF spi_device console in TX direction.

Workflow:
1. Device Ready Signal: Device initiates communication by sending a sync message to Host, signaling its readiness to receive data and its active wait state for an upload command.
2. Host Initiates Upload: Host responds by issuing an upload command. Following this, Host actively polls the status register, waiting for the BUSY bit to be cleared.
3. Device Data Processing: Upon receiving the upload command, Device reads data from the payload buffer. After processing, it clears the BUSY bit and enters a busy wait state, anticipating the next upload command.
4. Host Continues Upload: Host proceeds to send the subsequent upload command, repeating the process from step 2.

Transmission Completion:
- The write address of the last data chunk will be set to a special address, informing Device that the transmission is completed and Device can stop its busy wait state for further upload commands.

This PR addresses #21726 partially. The support for ujson_ottf will be added in another PR.

PR #24413, opened 2024-08-27
[base,ottf,test] Refactor ujson ottf console to support spi_device console

### LINKED ISSUE #21726: [manuf,ottf] implement spi_device console over mailbox

anthonychen1251:
This PR has four commits that:
1. Refactor ujson ottf console and implement `getc` for SPI device.
2. Test the new `getc` implementation.
3. Refactor mem module to support the new SPI device console
4. Add an example test for spi_device ujson ottf console.

This fixes #21726.

### DIFF
--- a/sw/device/lib/testing/test_framework/ottf_console.c
+++ b/sw/device/lib/testing/test_framework/ottf_console.c
@@ -49,6 +49,11 @@ enum {
 static dif_spi_device_handle_t ottf_console_spi_device;
 static dif_uart_t ottf_console_uart;
 
+// Function pointer to the currently active data sink.
+static sink_func_ptr sink;
+// Function pointer to a function that retrieves a single character.
+static status_t (*getc)(void *);
+
 // The `flow_control_state` and `flow_control_irqs` variables are shared between
 // the interrupt service handler and user code.
 static volatile ottf_console_flow_control_t flow_control_state;
@@ -61,6 +66,35 @@ void *ottf_console_get(void) {
     default:
       return &ottf_console_uart;
   }
+}
+
+static status_t uart_getc(void *io) {
+  const dif_uart_t *uart = (const dif_uart_t *)io;
+  uint8_t byte;
+  TRY(dif_uart_byte_receive_polled(uart, &byte));
+  TRY(ottf_console_flow_control(uart, kOttfConsoleFlowControlAuto));
+  return OK_STATUS(byte);
+}
+
+/*
+ * The user of this function needs to be aware of the following:
+ * 1. The exact amount of data expected to be sent from the host side must be
+ * known in advance.
+ * 2. Characters should be retrieved from the console as soon as they become
+ * available. Failure to do so may result in an SPI transaction timeout.
+ */
+static status_t spi_device_getc(void *io) {
+  static size_t index = 0;
+  static upload_info_t info = {0};
+  dif_spi_device_handle_t *spi_device = (dif_spi_device_handle_t *)io;
+  if (index == info.data_len) {
+    memset(&info, 0, sizeof(upload_info_t));
+    CHECK_STATUS_OK(spi_device_testutils_wait_for_upload(spi_device, &info));
+    index = 0;
+    CHECK_DIF_OK(dif_spi_device_set_flash_status_registers(spi_device, 0x00));
+  }
+
+  return OK_STATUS(info.data[index++]);
 }
 
 void ottf_console_init(void) {
@@ -76,9 +110,13 @@ void ottf_console_init(void) {
       }
 
       ottf_console_configure_uart(base_addr);
+      sink = get_uart_sink();
+      getc = uart_getc;
       break;
     case (kOttfConsoleSpiDevice):
       ottf_console_configure_spi_device(base_addr);
+      sink = get_spi_device_sink();
+      getc = spi_device_getc;
       break;
     default:
       CHECK(false, "unsupported OTTF console interface.");
@@ -324,3 +362,13 @@ size_t ottf_console_spi_device_read(size_t buf_size, uint8_t *const buf) {
 
   return received_data_len;
 }
+
+status_t ottf_console_putbuf(void *io, const char *buf, size_t len) {
+  size_t written_len = sink(io, buf, len);
+  if (len != written_len) {
+    return DATA_LOSS((int32_t)(len - written_len));
+  }
+  return OK_STATUS((int32_t)len);
+}
+
+status_t ottf_console_getc(void *io) { return getc(io); }
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -3276,6 +3276,36 @@ opentitan_test(
 )
 
 opentitan_test(
+    name = "spi_device_ujson_console_test",
+    srcs = ["spi_device_ujson_console_test.c"],
+    exec_env = dicts.add(
+        EARLGREY_CW340_TEST_ENVS,
+        {
+            "//hw/top_earlgrey:fpga_cw310_sival_rom_ext": None,
+            "//hw/top_earlgrey:fpga_cw310_rom_with_fake_keys": None,
+        },
+    ),
+    fpga = fpga_params(
+        test_cmd = """
+            --bootstrap="{firmware}"
+            "{firmware:elf}"
+        """,
+        test_harness = "//sw/host/tests/chip/spi_device_ujson_console_test",
+    ),
+    deps = [
+        "//hw/top_earlgrey/sw/autogen:top_earlgrey",
+        "//sw/device/lib/dif:spi_device",
+        "//sw/device/lib/runtime:log",
+        "//sw/device/lib/testing:spi_flash_testutils",
+        "//sw/device/lib/testing/json:command",
+        "//sw/device/lib/testing/json:provisioning_data",
+        "//sw/device/lib/testing/test_framework:check",
+        "//sw/device/lib/testing/test_framework:ottf_main",
+        "//sw/device/lib/testing/test_framework:ujson_ottf_commands",
+    ],
+)
+
+opentitan_test(
     name = "spi_device_sleep_test",
     srcs = ["spi_device_sleep_test.c"],
     exec_env = dicts.add(
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -1387,7 +1387,10 @@ opentitan_test(
 opentitan_test(
     name = "example_mem_ujcmd_test",
     srcs = ["example_mem_ujcmd.c"],
-    exec_env = EARLGREY_TEST_ENVS,
+    exec_env = dicts.add(
+        EARLGREY_TEST_ENVS,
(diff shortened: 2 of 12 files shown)


anthonychen1251:
cc @timothytrippel

timothytrippel:
I think you can rebase this on master now since https://github.com/lowRISC/opentitan/pull/24380 has been merged.

sw/device/lib/testing/test_framework/ottf_console.c
```diff
@@ -63,6 +68,37 @@ void *ottf_console_get(void) {
...
+  const dif_uart_t *uart = (const dif_uart_t *)io;
+  uint8_t byte;
+  TRY(dif_uart_byte_receive_polled(uart, &byte));
+  TRY(ottf_console_flow_control(uart, kOttfConsoleFlowControlAuto));
+  return OK_STATUS(byte);
+}
+
+/*
+ * The user of this function needs to be aware of the following:
+ * 1. The exact amount of data expected to be sent from the device side must be
```
timothytrippel, line 81: Do you mean:
```suggestion
 * 1. The exact amount of data expected to be sent from the host side must be
```
?

sw/device/lib/testing/test_framework/ottf_console.c
```diff
@@ -324,3 +364,13 @@ size_t ottf_console_spi_device_read(size_t buf_size, uint8_t *const buf) {
 
   return received_data_len;
 }
+
+status_t ottf_console_putbuf(void *io, const char *buf, size_t len) {
+  size_t written_len = sink(io, buf, len);
+  if (len != written_len) {
+    return DATA_LOSS((int32_t)written_len);
```
timothytrippel, line 371: wouldn't the data lost be `len - written_len`?

sw/device/lib/testing/test_framework/ottf_console.c
```diff
@@ -63,6 +68,37 @@ void *ottf_console_get(void) {
...
+ * available. Failure to do so may result in an SPI transaction timeout.
+ */
+static status_t spi_device_getc(void *io) {
+  static size_t index = 0;
+  static upload_info_t info = {0};
+
+  if (index == info.data_len) {
+    memset(&info, 0, sizeof(upload_info_t));
+    CHECK_STATUS_OK(
+        spi_device_testutils_wait_for_upload(&ottf_console_spi_device, &info));
```
timothytrippel, line 93: I think instead of `&ottf_console_spi_device` here you want to cast the `io` pointer to the spi_device DIF handle, like is done for the `uart_getc` function.

sw/device/tests/BUILD
```diff
@@ -3272,6 +3272,31 @@ opentitan_test(
     ],
 )
 
+opentitan_test(
+    name = "spi_device_ujson_ottf_test",
+    srcs = ["spi_device_ujson_ottf_test.c"],
+    exec_env = dicts.add(
+        EARLGREY_CW340_TEST_ENVS,
```
timothytrippel, line 3279: lets run this on cw310 too.

sw/device/tests/spi_device_ottf_console_test.c
```diff
@@ -143,5 +143,15 @@ bool test_main(void) {
       ottf_console_spi_device_read(sizeof(input_buf), input_buf);
   CHECK(received_data_len == sizeof(kTest4KbDataStr));
   CHECK_ARRAYS_EQ(input_buf, kTest4KbDataStr, ARRAYSIZE(kTest4KbDataStr));
+
```
timothytrippel, line 146: Add comment describing what this block is testing (since it is similar yet different from the other tests above)? i.e. underlying UJSON OTTF console data pipeline

sw/device/tests/spi_device_ujson_ottf_test.c
```diff
@@ -0,0 +1,48 @@
...
+#include "sw/device/lib/testing/test_framework/ujson_ottf.h"
+#include "sw/device/lib/ujson/ujson.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+OTTF_DEFINE_TEST_CONFIG(.console.type = kOttfConsoleSpiDevice,
+                        .console.base_addr = TOP_EARLGREY_SPI_DEVICE_BASE_ADDR,
+                        .console.test_may_clobber = false, );
+
+status_t dummy_sw_strap_read(ujson_t *uj) {
```
timothytrippel, line 23: This test has nothing to do with SW straps correct? If you need some UJSON Structs to send back and for testing, why don't you use this generic `MemRead` test command https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/json/command.h;l=25 , and send MemRead data payloads back and forth for one test: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/json/mem.h;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=24 (this should test data payloads of greater than 256 bytes).

Then you could also test sending a perso blob ujson struct to and from the host since it is > 4k in size which should excercise multi SPI buffer window transactions.

Also if you could add a block comment to the top describing what you are testing that would be helpful too.

sw/device/tests/spi_device_ottf_console_test.c
anthonychen1251: Okay, done.

sw/device/lib/testing/test_framework/ottf_console.c
anthonychen1251: Ah, you're right. Fixed the typo.

sw/device/lib/testing/test_framework/ottf_console.c
anthonychen1251: Good point! My original thought was to report the actual data written length along with a data loss error, but it seems to be reasonable to report the length of missing data.

sw/device/lib/testing/test_framework/ottf_console.c
anthonychen1251: Okay, done.

sw/device/tests/BUILD
anthonychen1251: Sure, done.

sw/device/tests/spi_device_ujson_ottf_test.c
anthonychen1251: Yes, the test has nothing to do with the functionality of SW straps. I did consider using the generic `Mem*` test commands but decided to use the SW straps command in the end. The tx transaction on SPI device console requires the device to send a sync message to notify the host it's ready for transmission but this limitation doesn't exist on UART console. The `mem` module encapsulates several tx transactions into a single `execute` function. To make the mem module support the workflow SPI device console, the existing `example_mem_ujcmd` test would be affected as well. I've submitted a new commit for refactoring the mem module. Let me know if that works for you.   I also added a test to send a perso blob ujson struct back and forth, thanks for the suggestion!

anthonychen1251: Thanks for the comment. The purpose is to ensure that the rx data has been read  by the host before host starts sending data to the device, so the device won't be blocked by the rx transmission if the read buffer is full and can be ready to receive the tx data. However, after I think about the use case here again, the sync operation in the `execute` seems to be unnecessary if a sync operation has already been performed in advance. Will double confirm on this and see if I can get rid of the `require_sync`

pamaury: Maybe this is a limitation of opentitanlib but I would have thought that the console would have a background thread so it can periodically read the SPI and therefore unblock the device if necessary?

nbdd0121: > The purpose is to ensure that the rx data has been read by the host before host starts sending data to the device, so the device won't be blocked by the rx transmission if the read buffer is full and can be ready to receive the tx data.

Couldn't the host just poll until the rx buffer is empty before sending data out? Would the device be sending data unsolicited to the host?

anthonychen1251: Removed the usage of `require_sync` in the new commit. Host poll until the rx buffer is empty before sending data out should work as well. Ideally, the device should not send unsolicited data to the host. I observed the issue when I tried to log plenty of debug messages on the device side but didn't handle those messages properly on the host side. After checking the behaviors on the existing ottf test code, I think the case I observed is an extreme case that should not occur in the real use scenario. The original purpose is not really necessary now.

anthonychen1251:
Upon pushing new commits, I observed that certain checks, like `CW310 ROM Tests`, were not executed again. Is there a way to rerun all the checks to ensure the latest commits do not break anything?

timothytrippel:
Hmm that is strange. IIRC, closing and re-opening the PR can retrigger CI. Let's try that.

sw/device/tests/BUILD
```diff
@@ -3272,6 +3279,36 @@ opentitan_test(
     ],
 )
 
+opentitan_test(
+    name = "spi_device_ujson_ottf_test",
+    srcs = ["spi_device_ujson_ottf_test.c"],
```
timothytrippel, line 3284: Can we rename this something like:
```suggestion
    name = "spi_device_ujson_console_test",
    srcs = ["spi_device_ujson_console_test.c"],
```

and likewise the `test_harness` below? 

The current name is confusing since there is also a `spi_device_ottf_console_test` that just tests the SPI device OTTF console, without sending ujson data payloads I believe.

timothytrippel (approved):

sw/device/tests/BUILD
anthonychen1251: Sure, done.

sw/device/tests/BUILD
```diff
@@ -1387,7 +1387,14 @@ opentitan_test(
 opentitan_test(
     name = "example_mem_ujcmd_test",
     srcs = ["example_mem_ujcmd.c"],
-    exec_env = EARLGREY_TEST_ENVS,
+    exec_env = dicts.add(
+        EARLGREY_TEST_ENVS,
+        EARLGREY_CW340_TEST_ENVS,
+        {
+            "//hw/top_earlgrey:fpga_cw310_sival_rom_ext": None,
```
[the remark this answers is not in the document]
anthonychen1251: Got it. Removed the cw310 part.

anthonychen1251:
The failed test `pattgen_ios_test_fpga_cw310_sival` in the most recent CI pipeline should be unrelated to the commits in the PR.
https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=159102&view=logs&j=5011a995-22d7-5073-e1ce-26fc382a8885&t=875997a6-7e2b-594d-05e6-ce6a085f3383

timothytrippel:
I retriggered CI. Hopefully the transient error should go away. Once it passes, I will merge.

PR #24414, opened 2024-08-27
[dv/flash] Increase fatal_err delay in flash_ctrl_phy_ack_consistency

eshapira:
Hi,
This small PR is increasing the delay for the fatal_err timeout in the flash_ctrl_phy_ack_consistency test because it seems to be insufficient in some seeds for the extending environment.

### DIFF
--- a/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_phy_ack_consistency_vseq.sv
+++ b/hw/ip_templates/flash_ctrl/dv/env/seq_lib/flash_ctrl_phy_ack_consistency_vseq.sv
@@ -21,7 +21,7 @@ class flash_ctrl_phy_ack_consistency_vseq extends flash_ctrl_phy_host_grant_err_
     cfg.scb_h.skip_alert_chk["recov_err"] = 1;
 
     cfg.scb_h.expected_alert["fatal_err"].expected = 1;
-    cfg.scb_h.expected_alert["fatal_err"].max_delay = 20000;
+    cfg.scb_h.expected_alert["fatal_err"].max_delay = 2_000_000;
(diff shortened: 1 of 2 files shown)


rswarbrick (commented):
This code is supposed to be saying "We expect an alert to come out, in at most 20_000 cycles". Here, you're weakening the expectation by a factor of 100. I'm willing to believe that the bound was too tight, but I think we need a bit more explanation of why the larger bound is correct, and we we don't have to worry about the change.

eshapira:
> This code is supposed to be saying "We expect an alert to come out, in at most 20_000 cycles". Here, you're weakening the expectation by a factor of 100. I'm willing to believe that the bound was too tight, but I think we need a bit more explanation of why the larger bound is correct, and we we don't have to worry about the change.

Thanks for your comment!

I understand the meaning of this timeout value, but the question is why its value is as it is.
The correct approach in all of those tests should have been to use overridable parameters for the timeouts, since this code should run on different types of design in which the time can be higher by many scales.
In this specific example I had seed of this test in which the fatal alert was asserted after 520,000 cycles.
The only reason I made it 2,000,000 is because I wanted to take extra time for maybe even slower scenarios that can appear.

Please correct me if I'm wrong, but I don't see a reason this change will cause problems.

rswarbrick:
I'll let @matutem address this properly (since he knows lots more about flash_ctrl than I do!). Amusingly, I see that it was also bumped up by a factor of 10 last July in aae501fe12a80fd5cd24d1160e692c1f3e8819bd.

eshapira:
@matutem Can you please take a look and review this PR?
Thanks!

matutem (approved):
This tests must be really slow for some tests.
LGTM

PR #24415, opened 2024-08-27
[rom] fix ROM_EXT immutable section to work with address translation

timothytrippel:
The original ROM_EXT immutable section configuration did not account for the address translation adjustments to the immutable section entry point. This meant that if address translation was enabled, the immutable section entry point would never get moved to the virtual address space, which gets marked RX. The physical address space of the current slot would only be marked R, and when jumping to the entry point an exception would occur. This slipped through b/c our tooling always calculated the correct offset to the immutable section base on symbols in the ELF, but meant that the same offset could not be used across different ROM_EXT slot locations (which is important since the immutable section offset is stored in fuses). This fixes that error and enhances the tests to ensure the same ROM_EXT mutable section offset can be used across tests that load a ROM_EXT in different slots (A, B, or Virtual slots).

### DIFF
--- a/sw/device/silicon_creator/rom/e2e/immutable_rom_ext_section/BUILD
+++ b/sw/device/silicon_creator/rom/e2e/immutable_rom_ext_section/BUILD
@@ -3,20 +3,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 load(
-    "@bazel_skylib//lib:dicts.bzl",
-    "dicts",
-)
-load(
     "//rules:const.bzl",
     "CONST",
-    "hex",
     "hex_digits",
 )
-load(
-    "//sw/device/silicon_creator/rom_ext:defs.bzl",
-    "ROM_EXT_VERSION",
-)
-load("//rules:manifest.bzl", "manifest")
 load(
     "//rules:otp.bzl",
     "STD_OTP_OVERLAYS",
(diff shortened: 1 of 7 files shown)


sw/device/silicon_creator/rom/e2e/immutable_rom_ext_section/BUILD
```diff
@@ -203,14 +222,14 @@ manifest(d = dicts.add(
...
+        name = "immutable_rom_ext_section_test_{}".format(s["name"]),
         testonly = True,
         srcs = ["immutable_rom_ext_section_test.c"],
         exec_env = [
             "//hw/top_earlgrey:fpga_cw310_sival",
         ],
-        linker_script = "//sw/device/lib/testing/test_framework:ottf_ld_silicon_creator_slot_{}".format(ld_script),
-        manifest = ":manifest_addr_trans_on" if ld_script == "virtual" else ":manifest_addr_trans_off",
+        linker_script = "//sw/device/lib/testing/test_framework:ottf_ld_silicon_creator_slot_{}".format(s["slot"]),
+        manifest = ":manifest_addr_trans_on" if s["slot"] == "virtual" else ":manifest_addr_trans_off",
```
cfrantz, line 232: This may not be required: the linker scripts in the code base (e.g. [ottf_common.ld](https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/testing/test_framework/ottf_common.ld#L56-L58) now automatically detect whether address translation is enabled and create a linker-defined constant that is consumed by the [manifest definition](https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/lib/manifest_def.c#L37).

You can probably eliminate the two explicit manifest rules in this BUILD file as well.

This address-translation auto-detect machinery was added to avoid having extraneous manifest definitions, particularly for ROM_EXT releases, but it should also help reduce complexity here.

timothytrippel:
Fixes #24416.

sw/device/silicon_creator/rom/e2e/immutable_rom_ext_section/immutable_rom_ext_section_test.c
```diff
@@ -64,6 +64,15 @@ bool test_main(void) {
       immutable_rom_ext_hash[3], immutable_rom_ext_hash[2],
       immutable_rom_ext_hash[1], immutable_rom_ext_hash[0]);
 
+  // Check the same immutabled section offset is used by every test to ensure it
```
cfrantz, line 67: nit (spelling): `s/immutabled/immutable/`

timothytrippel:
CHANGE AUTHORIZED: sw/device/silicon_creator/rom/e2e/release/rom_e2e_self_hash_test.c

sw/device/silicon_creator/rom/e2e/immutable_rom_ext_section/immutable_rom_ext_section_test.c
timothytrippel: fixed.

moidx:
CHANGE AUTHORIZED: sw/device/silicon_creator/rom/e2e/release/rom_e2e_self_hash_test.c

moidx (approved):

PR #24430, opened 2024-08-29
[rom_ext_e2e] Update CI to run ROM_EXT tests on hyper310 

cfrantz:
Use the `fpga_hyper310_rom_ext` environment for ROM_EXT end-to-end tests.
The hyperdebug serial ports support serial break out of the box, whereas
the CW310 serial ports require special SAM3x firmware to support serial
break.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -654,23 +654,23 @@ jobs:
     demands: BOARD -equals cw310
   timeoutInMinutes: 60
   dependsOn:
-    - chip_earlgrey_cw310
+    - chip_earlgrey_cw310_hyperdebug
     - sw_build
-  condition: succeeded( 'chip_earlgrey_cw310', 'sw_build' )
+  condition: succeeded( 'chip_earlgrey_cw310_hyperdebug', 'sw_build' )
   steps:
   - template: ci/checkout-template.yml
   - template: ci/install-package-dependencies.yml
   - template: ci/download-artifacts-template.yml
     parameters:
       downloadPartialBuildBinFrom:
-        - chip_earlgrey_cw310
+        - chip_earlgrey_cw310_hyperdebug
         - sw_build
   - template: ci/load-bazel-cache-write-creds.yml
   - bash: |
       set -e
       . util/build_consts.sh
       module load "xilinx/vivado/$(VIVADO_VERSION)"
-      ci/scripts/run-fpga-tests.sh cw310 fpga_cw310_rom_ext_tests || { res=$?; echo "To reproduce failures locally, follow the instructions at https://opentitan.org/book/doc/getting_started/setup_fpga.html#reproducing-fpga-ci-failures-locally"; exit "${res}"; }
+      ci/scripts/run-fpga-tests.sh hyper310 fpga_hyper310_rom_ext_tests || { res=$?; echo "To reproduce failures locally, follow the instructions at https://opentitan.org/book/doc/getting_started/setup_fpga.html#reproducing-fpga-ci-failures-locally"; exit "${res}"; }
     displayName: Execute tests
   - template: ci/publish-bazel-test-results.yml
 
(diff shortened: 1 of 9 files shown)


azure-pipelines.yml
```diff
@@ -670,7 +670,7 @@ jobs:
       set -e
       . util/build_consts.sh
       module load "xilinx/vivado/$(VIVADO_VERSION)"
-      ci/scripts/run-fpga-tests.sh cw310 fpga_cw310_rom_ext_tests || { res=$?; echo "To reproduce failures locally, follow the instructions at https://opentitan.org/book/doc/getting_started/setup_fpga.html#reproducing-fpga-ci-failures-locally"; exit "${res}"; }
+      ci/scripts/run-fpga-tests.sh hyper310 fpga_hyper310_rom_ext_tests || { res=$?; echo "To reproduce failures locally, follow the instructions at https://opentitan.org/book/doc/getting_started/setup_fpga.html#reproducing-fpga-ci-failures-locally"; exit "${res}"; }
```
a-will, line 673: Oops, it's missing the artifacts from (and dependency on) the chip_earlgrey_cw310_hyperdebug job. :)

rules/opentitan/defs.bzl
```diff
@@ -120,7 +120,9 @@ def _parameter_name(env, pname):
 def _hacky_tags(env):
     (_, suffix) = env.split(":")
     tags = []
-    if suffix.startswith("fpga_cw310_") or suffix.startswith("fpga_cw340_"):
```
pamaury, line 123: It is possible that master and es have diverged at this point but in master we have a slightly better way of doing hacky_tags (still hacky): https://github.com/lowRISC/opentitan/blob/cf32d0f1190b72144c52ffe1b226b4fe37eb3988/rules/opentitan/defs.bzl#L142

azure-pipelines.yml
pamaury: I am not a 100% sure I believe we have tests that are defined for fpga_cw310_rom_ext and not fpga_hyper310_rom_ext_tests (like the cryptotests). With this changes, those tests will not run anymore, but maybe they don't need to run on earlgrey_es?

cfrantz: There are some tests defined that use `fpga_cw310_rom_ext`.  They're located in `//sw/device/silicon_owner/bare_metal` and `//sw/device/tests/crypto`.  However, in the `run-fpga-tests` script, it appears that an `fpga_tags` argument (arg 2) of `fpga_{whatevs}_rom_ext_tests` sets the pattern to `//sw/device/silicon_creator/rom_ext/e2e/...`, so this pipeline stage was never running those other tests in the first place.

rules/opentitan/defs.bzl
cfrantz: They have diverged and from a first look, the function on master looks better.  I don't think I want to dive down the rabbithole of resolving this divergence.

timothytrippel (approved):

azure-pipelines.yml
cfrantz: AFAICT, the following tests are not currently being run in CI:
```
//sw/device/silicon_owner/bare_metal:rom_ext_virtual_bare_metal_virtual_boot_test
//sw/device/silicon_owner/bare_metal:rom_ext_virtual_ottf_bl0_virtual
//sw/device/tests/crypto:otcrypto_hash_test
//sw/device/tests/crypto:otcrypto_export_test
```

The change in this PR does not affect whether or not these tests run.

It would be nice to run them on `earlgrey_es_sival`.  We should make sure they run on `master`:
- The bare metal test is a post-ROM_EXT test that checks if we can run owner code that is mostly separate from our existing test framework.  We could argue about whether the this test truly provides value now that there are ROM_EXT E2E tests in place.
- The crypto tests check that the crypto library distributable build artifact is functional.

pamaury: If you want to run these jobs on the earlgrey_es_sival branch, I think a new FPGA job will be needed. I will double-check that they run on master (or at least will run once the CI orchestrator is merged)

PR #24432, opened 2024-08-29
[prim,fpv] Explicitly delay wready_o assertions by a cycle

rswarbrick:
The old version of these assertions were not quite true, because the under_rst signal (which qualifies wready_o) is only cleared a cycle after we come out of reset. Tweak the assertions to match.


hcallahan-lowrisc (approved):
LGTM, thanks @rswarbrick

PR #24434, opened 2024-08-29
[hw,dma,rtl] Sync DMA from integrated_dev to Master

Razer6:
This PR pulls over the DMA from the integrated_dev repository. It only includes changes from `hw/ip/dma` modulo some documentation changes to include the DMA into the list of supported repositories and adding the `REGWEN_MUBI` as a countermeasure to the infrastructure.

This PR was created with:
```
git format-patch -k --stdout integrated_dev...master -- hw/ip/dma | git am -3 -k
```
and a little bit of manual rewriting of the history for the aforementioned changes.


rswarbrick:
Hi there! I've done a bit of reviewing and taken "markdown notes" which I'm pasting here. There are some "review comments" at the bottom which I think probably need addressing before merge, then some "action items" which I suspect are on a list already (but I wanted to make sure I understood what was going on).

# List of commits with changes

The following list shows the commits that come up, together with a
note. To see the commit with a bit of context, the easiest trick seems
to be to run something like
```
  git log --reverse -p origin/master..Razer6/dma-sync-to-master
```
and then search in `less` for the "migration commit" SHAs

## [hw,dma,rtl] DMA skeleton for OT-integrated (5403885)

The diff to `hw/ip/README.md` here is because `ascon` was added to
`master` and `flash_ctrl` and `clkmgr` were converted to use `ipgen`
in the meantime.

The diff in `hw/top_darjeeling/lint/top_darjeeling_lint_cfgs.hjson`
shows that the "original commit" (the version in `integrated_dev`)
added `dma` to the lint configs. This gets dropped in this rebase
because lint configuration isn't yet set up for `top_darjeeling` on
master: this will happen in a follow-up PR (action item below).

## [dma,doc] Initial ingestion of specification (648a521)

This was linked into the top-level `SUMMARY.md` in the original commit
and the linking hasn't been migrated. This is probably an action item:
"link the DMA documentation in the main docs".

The original commit also wired up the DIF which hasn't yet been ported
at all. On the `integrated_dev` branch, the original commit to the
file in question was 52a43fb but that hasn't been included in the
porting flow yet. Action item: port that commit (and its successors)

## [regtool] Cast to mubi4_t explicitly (a184bed)

On `master`, this change already appears in `reg_top.sv.tpl` (as
commit 25bb576 from September 2023), but that change doesn't include
`dma_reg_top.sv` (of course!) and the tooling doesn't realise exactly
what's going on. This looks perfectly sensible to me.

## [reggen] Add DMA to makefile and make MuBi regwen width parametric (407e934)

The migrated commit doesn't actually add DMA to the Makefile: in fact,
it's just instantiating a stub `dma_sec_cm_testplan.hjson`. I'm not
sure if this is intentional: added as a question and an action item
below.

## [reggen] Add initial support for version and cip_id hjson fields (3778892)

This change conflicts dramatically with the automatically migrated
version, but the manual port looks correct to me.

## [hw,dma] Introduce *.REGWEN_MUBI security countermeasure (0ef14f1)

The tooling doesn't seem to have managed to follow the manual port of
this change. In both cases, the commit defines the "REGWEN_MUBI" name
in the project-level list of known countermeasures and then tags some
registers in `dma` with the name. These look like they match
correctly.

## [reggen] Add mubi support SWAccess that sets/clears a reg (4a4346c)

The diff here is just in autogenerated files, which looks a bit odd to
me. I think the initial version of `dma_reg_top.sv` needs regenerating.

## [hw,dma,dv] update sequence to use fifo instance (6fb8aeb)

It looks like the change that's being cherry-picked (db92b36) added
an `AddrWidth` parameter to `tl_device_seq`, which hasn't been ported.
I don't think that passing the parameter will work without.

## [hw,dma,dv] Fix compile errors and warnings (96d2c96)

The only difference in porting this (rather scattergun!) original
commit is that one bit of it was a tidy-up in `tl_device_seq.sv` that
I have performed separately since.

## [soc_proxy,dma] Increase number of external LSIO triggers to 8 (ae62546)

Only half of this commit is being ported (because `soc_proxy` is not
here!), which explains the diff. The ported version looks sensible to
me.

## [hw,dma,rtl] Rename RANGE_UNLOCK_REGWEN to RANGE_REGWEN for a better semantic (b28c182)

The diff here is because the commit also affects `dif_dma.c` which
hasn't yet been ported. That already has an action item below.

## [hw,dma,rtl] Add a register hole to cover all supported number of LSIOs (eba19c9)

The difference here is slightly surprising to me. The original version
of the commit (1ecb4e0) changed the address width of the block
(because of adding the `skipto`), which then got reflected in some
top-level files. The migrated version still changes the address width
but that doesn't seem to have any effect in the top-level structure.
I'm not yet sure that's right?

## [hw,dma,rtl] Enable TL-UL integrity checks on response (7ce3760)

This doesn't seem to have been reflected in
`top_darjeeling.gen.hjson`: I think it should have been?

## [hw,dma,rtl] Remove memory threshold interrupt from RTL and DIF (21bf224)

The only changes here are to the DIF, which hasn't been ported (and
already has an action item).

# Review comments

- Commit 407e934 doesn't seem to do anything much. Does the stub
  `sec_cm` testplan matter? If not, I propose dropping the commit and
  adding `dma` to the Makefile in a followup (listed as an action item
  below). Alternatively, it might be worth picking up more of this commit?

- Because the "mubi support" change in reggen has already landed on
  `master`, I think we can drop 4a4346c entirely but we should
  probably regenerate `dma_reg_top.sv` for the initial commit that
  adds the file, to avoid having commits where the autogenerated file
  isn't in sync with its source.

- I don't think that the incomplete port that gives 6fb8aeb will
  work: it's passing a parameter to a base class that doesn't expect
  it.

- Shouldn't "[hw,dma,rtl] Add a register hole to cover all supported
  number of LSIOs" cause a change in the top-level structure? I'd
  expect at least a change to `top_darjeeling.gen.hjson`.

# Action items

- Set up lint for `top_darjeeling` on `master`.
- Link the DMA documentation in the main docs (which might need a bit
  of high-level documentation restructuring)
- Add a DIF entry for the DMA ip.
- Add `dma` to the Makefile

Razer6:
Thank you for the very thorough review @rswarbrick! 

> * Commit [407e934](https://github.com/lowRISC/opentitan/commit/407e9340353871b763bc02e72abf09f230f2b42a) doesn't seem to do anything much. Does the stub
>   `sec_cm` testplan matter? If not, I propose dropping the commit and
>   adding `dma` to the Makefile in a followup (listed as an action item
>   below). Alternatively, it might be worth picking up more of this commit?

I added the DMA to the makefile. The `sec_cm` stub was generated when adding the DMA to the makefile, thus I propose leaving it as it is now.

> * Because the "mubi support" change in reggen has already landed on
>   `master`, I think we can drop [4a4346c](https://github.com/lowRISC/opentitan/commit/4a4346c9bebfd0e9e34005307a82e06c5670f4f5) entirely but we should
>   probably regenerate `dma_reg_top.sv` for the initial commit that
>   adds the file, to avoid having commits where the autogenerated file
>   isn't in sync with its source.

I dropped that commit and re-generated the reg_top as soon as it was possible to build without errors (after merging the CIP ID).

> * I don't think that the incomplete port that gives [6fb8aeb](https://github.com/lowRISC/opentitan/commit/6fb8aebfd758c7533d323d2f59db6715860ac418) will
>   work: it's passing a parameter to a base class that doesn't expect
>   it.

I added the parameter to `hw/dv/sv/tl_agent/seq_lib/tl_device_seq.sv`.

> * Shouldn't "[hw,dma,rtl] Add a register hole to cover all supported
>   number of LSIOs" cause a change in the top-level structure? I'd
>   expect at least a change to `top_darjeeling.gen.hjson`.

I am confused here? There is no top-level at this point in the master branch, thus no change?

Regarding the action points:

> * Set up lint for `top_darjeeling` on `master`.

Prepared a PR in https://github.com/lowRISC/opentitan/pull/24625

> * Link the DMA documentation in the main docs (which might need a bit
>   of high-level documentation restructuring)

DMA docs I thing need a general overhaul.

> * Add a DIF entry for the DMA IP.

Prepared a commit in https://github.com/lowRISC/opentitan/pull/24618

> * Add `dma` to the Makefile

Added in the particular commit.

hw/Makefile
```diff
@@ -19,6 +19,7 @@ IPS ?= aes           \
        csrng         \
        edn           \
        entropy_src   \
+       dma           \
```
andreaskurth, line 22: Suggest we keep this list sorted

Razer6: Ups, that was a bit late last night. Fixed the off-by-one error.

rswarbrick:
Sorry for the rather slow response. Working by items: (in hindsight, this is presumably why GitHub supports running multiple threads of discussion: oops!)

- Thanks for adding the DMA to the makefile. This looks good to me.
- Thanks for regenerating the reg_top as you describe. Definitely a pain point from the project's decision to check in autogenerated files...
- Thanks for tweaking the device sequence: this looks good to me.
- About the top-level note: I was asking a silly question! Please ignore! :-)

On the action items, I agree (and thanks for pointing to the other PRs as necessary). Thanks for the careful responses and I'm convinced this can go in! I'll approve and add magic "authorization comments" in a sec.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma.sv
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma_pkg.sv
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma_reg_top.sv

Razer6:
Awesome :) Who would be the second approver to get it merged?

mundaym:
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma.sv
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma_pkg.sv
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/dma/rtl/dma_reg_top.sv

PR #24435, opened 2024-08-29
[otbn] Allow sim to compare result to dmem

dop-amin:
This PR introduces the ability to compare the data memory dump after a OTBN simulator run to a file containing expected data memory results. 

In brief:
- The setup is generally similar to the comparison to .exp files.
- We call the files .dexp for "dmem expected".
- In a .dexp file, you write a label from the data section in memory followed by the expected data as bytes encoded as hex words.
- We parse the dmem dump, removing integrity protection bytes.
- We find the offset of the label in the elf file and read as many bytes as given in the .dexp file from this location from the dmem dump.
- We compare the expected bytes to the bytes read from memory.
- For mul_test, we integrate this so the test verifies the result in dmem. You can take a look at this for an example use-case.


hw/ip/otbn/util/otbn_sim_test.py
```diff
@@ -63,49 +81,89 @@ def main() -> int:
...
+            # case, give a special error message and exit rather than print all the
+            # mismatched registers.
+            if actual_err != 0:
+                err_names = ", ".join(get_err_names(actual_err))
+                result.err(
+                    f"OTBN encountered an unexpected error: {err_names}.\n"
+                    f"  {ERR_BITS}\t= {actual_err:#010x}\n"
+                    f"  {INSN_CNT}\t= {insn_cnt:#010x}\n"
+                    f"  {STOP_PC}\t= {stop_pc:#010x}"
+                )
```
jadephilipoom, line 135: It might make sense to move this logic out of the if statement, and just use an expected error value of 0 (no errors) if no expected-register file is provided. Right now, I think if there are no expected register values provided and the program encounters an error, this rule will only check the values in DMEM and won't say that there was an error at all. Like for registers, I think it also makes sense in that case not to print the mismatched DMEM values, since the end of the test was not reached.

hw/ip/otbn/util/shared/dmem_dump.py
```diff
@@ -0,0 +1,58 @@
...
+                             f'for {label}.')
+        out[label] = value
+
+    return out
+
+
+def parse_actual_dmem(dump: bytes) -> bytes:
+    '''Parse the dmem dump.
+
+    Returns the dmem bytes that except integrity info.
```
jadephilipoom, line 46: nit:
```suggestion
    Returns the dmem bytes except integrity info.
```

jadephilipoom (approved):
Couple of small comments but generally looks good to me, thanks Amin!

dop-amin:
> Couple of small comments but generally looks good to me, thanks Amin!

Thanks for the feedback, I addressed all of it.

jadephilipoom:
Great! Looks ready to merge once CI passes.

jadephilipoom:
It looks like a lot of OTBN crypto tests are failing now for some reason :face_with_diagonal_mouth: There's also a failure on the sival ROM_EXT tests, but that looks unrelated -- likely a flaky test.

PR #24443, opened 2024-08-29
[ot_certs] Add support for custom key usage

pamaury:
This allows certificate templates to specify whether they want a key usage extension at all, and if they do what the key usages are. I only added support for the 3 usages that seem the most

### DIFF
--- a/sw/host/ot_certs/src/template/mod.rs
+++ b/sw/host/ot_certs/src/template/mod.rs
@@ -75,6 +75,7 @@ pub struct Certificate {
     pub subject_key_identifier: Value<Vec<u8>>,
     // X509 basic constraints extension, optional.
     pub basic_constraints: Option<BasicConstraints>,
+    pub key_usage: Option<KeyUsage>,
     /// X509 Subject Alternative Name extension, optional.
     #[serde(default)]
     pub subject_alt_name: Name,
@@ -97,6 +98,13 @@ pub type Name = Vec<IndexMap<AttributeType, Value<String>>>;
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
 pub struct BasicConstraints {
     pub ca: Value<bool>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct KeyUsage {
+    pub digital_signature: Option<Value<bool>>,
+    pub key_agreement: Option<Value<bool>>,
+    pub cert_sign: Option<Value<bool>>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
@@ -466,6 +474,7 @@ mod tests {
                 },
                 authority_key_identifier: { var: "signing_pub_key_id" },
                 subject_key_identifier: { var: "owner_pub_key_id" },
+                key_usage: { key_agreement: true },
                 private_extensions: [
                     {
                         type: "dice_tcb_info",
@@ -562,6 +571,11 @@ mod tests {
             authority_key_identifier: Value::variable("signing_pub_key_id"),
             subject_key_identifier: Value::variable("owner_pub_key_id"),
             basic_constraints: None,
+            key_usage: Some(KeyUsage {
+                digital_signature: None,
+                key_agreement: Some(Value::literal(true)),
+                cert_sign: None,
+            }),
             subject_alt_name: vec![],
             private_extensions: vec![CertificateExtension::DiceTcbInfo(DiceTcbInfoExtension {
                 vendor: Some(Value::literal("OpenTitan")),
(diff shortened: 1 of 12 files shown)


sw/host/ot_certs/src/template/mod.rs
```diff
@@ -562,6 +571,11 @@ mod tests {
             authority_key_identifier: Value::variable("signing_pub_key_id"),
             subject_key_identifier: Value::variable("owner_pub_key_id"),
             basic_constraints: None,
+            key_usage: Some(KeyUsage {
```
vbendeb, line 574: why is this addition necessary (as opposed to `key_usage: None,` )?

Also, if you want to maintain backwards compatibility so that the default key usage field value does not change after this patch shouldn't `cert_sign` be set here  instead of `key_agreement`?

sw/host/ot_certs/src/x509/extension.rs
```diff
@@ -228,6 +228,31 @@ impl<'a> asn1::SimpleAsn1Readable<'a> for DiceTcbInfoFlags {
...
 }
 
+impl<'a> asn1::SimpleAsn1Readable<'a> for KeyUsage {
+    const TAG: asn1::Tag = asn1::OwnedBitString::TAG;
+    fn parse_data(_data: &'a [u8]) -> asn1::ParseResult<Self> {
+        let result = asn1::OwnedBitString::parse_data(_data)?;
+        let bs = result.as_bitstring();
+        // List of bits that this parser supports. Any bit set outside of
+        // these will be an error because we cannot record it.
+        const PARSED_BITS: &[usize] = &[0, 4, 5];
```
vbendeb, line 238: where do these bit values come from: 

This site does not seem in agreement: https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509keyusageflags?view=net-8.0

vbendeb (commented):
a couple of questions

sw/host/ot_certs/src/x509/extension.rs
timothytrippel: I believe they come from: https://www.rfc-editor.org/rfc/rfc5280.html#section-4.2.1.3

Maybe we should link the src in a comment?

timothytrippel: which agrees with another source here: https://docs.redhat.com/en/documentation/red_hat_certificate_system/9/html/administration_guide/standard_x.509_v3_certificate_extensions#Standard_X.509_v3_Certificate_Extensions-keyUsage

timothytrippel (approved):

sw/host/ot_certs/src/x509/extension.rs
pamaury: All those bits values are documented in the `push_key_usage_ext` function, I didn't want to duplicate the documentation there but maybe I'll just copy-paste it anyways. In any case, the bits in PARSED_BITS are simply the bits that the code right below handles. This is to make sure that the parse will fail if for any reason there is a bit set that cannot be represented by the KeyUsage structure.

Alternatively, we could handle all possibles bits, even if they are very unlikely to be used.

sw/host/ot_certs/src/template/mod.rs
pamaury: This just to make sure that the unittest is thorough (ie it can parse a key usage). I don't think this unit test should be use for the purpose of checking backward compatibility but maybe we could create another unittest whose sole purpose is to ensure that default values don't change?

pamaury:
I guess this raises another question: with this PR, I made a breaking where not specifying a key usage will not create a key usage extension at all. The alternative would be to say that not specifying a key usage extension revers to the previous behaviour (just cert_sign set). But what if someone wants to key usage extension at all? (here I am using key usage as an example, but the DICE extension would be another one). Maybe when we make changes like that we would want a tri-state:
- if `key_usage` is not specified, then it defaults to the previous behaviour for backward compatibility,
- if `key_usage` is specified, then it overrides the default
- there is a way to say that we don't want a key usage extension at-all, something like `key_usage: none`.

What do you think?

sw/host/ot_certs/src/x509/extension.rs
vbendeb: Maybe mention in the comments that bit values come from RFC5280?

vbendeb:
> I guess this raises another question: with this PR, I made a breaking where not specifying a key usage will not create a key usage extension at all. The alternative would be to say that not specifying a key usage extension reverts to the previous behaviour (just cert_sign set). But what if someone wants to key usage extension at all? (here I am using key usage as an example, but the DICE extension would be another one). Maybe when we make changes like that we would want a tri-state:
> 
> * if `key_usage` is not specified, then it defaults to the previous behaviour for backward compatibility,
> * if `key_usage` is specified, then it overrides the default
> * there is a way to say that we don't want a key usage extension at-all, something like `key_usage: none`.
> 
> What do you think?

I think it is fine to omit it if the user did not include it in the template.

vbendeb:
I don't see how I can vote on this, but FWIW it looks good to me

PR #24453, opened 2024-08-30
[WIP] Multi-top SW

pamaury:
**This is still a WIP, please refrain from commenting too much on the details at this point. High-level comments are welcome.**

This is a test PR for the multi-top SW effort. The goal of this PR is to see (1) how much is broken in CI, (2) allow for high-level comments on the approach.

**Explanation:**

TODO

**Important:** LTO is not enable in this PR.

**How to regenerate everything after making changes to templates:**
Run
```
util/topgen.py -t hw/top_earlgrey/data/top_earlgrey.hjson
util/make_new_dif.py --mode=regen --only=autogen
```

**How to test:**
Run
```
./bazelisk.sh test --test_output=streamed //sw/device/tests:aon_timer_irq_test_fpga_cw310_rom_with_fake_keys //sw/device/tests:sysrst_ctrl_outputs_test_fpga_cw310_sival
```
Build EB verilator
```
./bazelisk.sh build --config=english_breakfast  //hw:verilator
```
Build EB test_rom:
```
./bazelisk.sh build --config=english_breakfast --config=riscv32 //sw/device/lib/testing/test_rom:test_rom
```
Run the test_rom_test:
```
./bazelisk.sh test --test_output=streamed --config=english_breakfast //sw/device/lib/testing/test_rom:test_rom_test_sim_verilator
```

**How to look at the generated code:**

For top-level:
```
# build
./bazelisk.sh build //hw/top_earlgrey/sw/autogen/...
# then query files:
./bazelisk.sh cquery //hw/top_earlgrey/sw/autogen:top_earlgrey_dttool_dt_api
./bazelisk.sh cquery //hw/top_earlgrey/sw/autogen:top_earlgrey_dttool_dt_lib_hdrs
./bazelisk.sh cquery //hw/top_earlgrey/sw/autogen:top_earlgrey_dttool_dt_lib_srcs
```
For IP:
```
# build
./bazelisk.sh build //hw/ip/sysrst_ctrl/data:sysrst_ctrl_dt
# then query files:
./bazelisk.sh cquery //hw/ip/sysrst_ctrl/data:sysrst_ctrl_dt
```

TODOs:

- [ ] cleanup the templates (they are super messy and contain duplicated code)
- [ ] redo the LTO work from a clean state, see #22052 

**This is still a WIP, please refrain from commenting too much on the details at this point. High-level comments are welcome.**


a-will:
Just a reminder that Co-authored-by is a way to give attribution to code that wasn't only done by you. ;)

pamaury:
> Just a reminder that Co-authored-by is a way to give attribution to code that wasn't only done by you. ;)

Sure no worries, I'll add you back :) I just created this PR because several people didn't like looking at code in my branch (the GitHub UI is pretty poor). I have basically just squashed all commits into a few. The original code is here if you want the messy history: https://github.com/lowRISC/opentitan/compare/master...pamaury:opentitan:dt_for_multitop

util/make_new_dif/templates/dif_autogen.c.tpl
```diff
@@ -104,7 +121,7 @@ dif_result_t dif_${ip.name_snake}_init(
 
     static bool ${ip.name_snake}_get_irq_reg_offset(
       dif_${ip.name_snake}_intr_reg_t intr_reg,
-      dif_${ip.name_snake}_irq_t irq,
+      dt_${ip.name_snake}_irq_type_t irq,
```
engdoreis, line 124: Why the suffix `_type_t`? The `_t` already means `type`, no?

hw/top_earlgrey/sw/autogen/devicetables.c
```diff
@@ -0,0 +1,2181 @@
...
+  },
+};
+// Device tables for uart
+_Static_assert(kDtUartRegBlockCount == 1, "Reg block count mismatch");
+_Static_assert(kDtUartClockCount == 1, "Clock count mismatch");
+_Static_assert(kDtUartIrqTypeCount == 9, "IRQ count mismatch");
+_Static_assert(kDtUartCount == 4, "Number of uart modules mismatch");
+
+const dt_uart_t kDtUart[kDtUartCount] = {
+  // Properties for uart0
```
engdoreis, line 1654:
```suggestion
[kDtUart0] = 
```

hw/top_earlgrey/sw/autogen/devicetables.h
```diff
@@ -0,0 +1,162 @@
...
+extern const dt_sysrst_ctrl_t kDtSysrstCtrl[kDtSysrstCtrlCount];
+
+// Device tables for uart
+extern const dt_uart_t kDtUart[kDtUartCount];
+
+// Device tables for usbdev
+extern const dt_usbdev_t kDtUsbdev[kDtUsbdevCount];
+
+
+#endif // OPENTITAN_HW_TOP_EARLGREY_SW_AUTOGEN_DEVICETABLES_H_
```
engdoreis, line 162: Add new line

hw/top_earlgrey/sw/autogen/devicetables.c
```diff
@@ -0,0 +1,2181 @@
...
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+#include <stdint.h>
+
+// Device tables for adc_ctrl
+_Static_assert(kDtAdcCtrlRegBlockCount == 1, "Reg block count mismatch");
+_Static_assert(kDtAdcCtrlClockCount == 2, "Clock count mismatch");
+_Static_assert(kDtAdcCtrlIrqTypeCount == 1, "IRQ count mismatch");
+_Static_assert(kDtAdcCtrlCount == 1, "Number of adc_ctrl modules mismatch");
+
+const dt_adc_ctrl_t kDtAdcCtrl[kDtAdcCtrlCount] = {
```
engdoreis, line 26: Have you considered using a unified type for all the peripherals?
This type would have all possible fields, but some would be left empty when it does not make sense for a particular block.
This should be safe as we expect the tests to use the accessor functions to read from the device table, so all the checks will be done by autogenerated code.

I'm not entirely convinced this would be a batter approach, but it would reduce the amount of auto-generated code.
But it may prevent LTO to do a batter work?

hw/top_earlgrey/sw/cw310/dt_clock.c
engdoreis, line 1: Should this change be on the previous commit?

hw/top_earlgrey/sw/autogen/devicetables.c
pamaury: Yes, but I think it would have some drawbacks. It's still less type safe unless you had plenty of runtime checks, whereas by using a unique type, you literally can't write wrong code unless you really want to.
Also if you have some arrays (like `pins`) in the data then you would need either to point to array somewhere (more indirection and the data is not trivially copyable) or make the array of the maximal possible size.

Another advantage of using a unique type for each IP is that in the future, it is possible that some IP may want to embed more information in the DT and having a different type for each one makes that cleaner in my opinion.

hw/top_earlgrey/sw/cw310/dt_clock.c
pamaury: No because it depends on `dt_clock.c` that is a manually created file. I wanted the previous commit to be entirely autogenerated which is why I didn't include `dt_clock.c` in it. Possibly I could split `dt_clock.c` to its own commit?

sw/device/lib/dif/autogen/dif_adc_ctrl_autogen.c
```diff
@@ -25,6 +25,18 @@ dif_result_t dif_adc_ctrl_init(mmio_region_t base_addr,
   return kDifOk;
 }
 
+OT_WARN_UNUSED_RESULT
+dif_result_t dif_adc_ctrl_init_dt(const dt_adc_ctrl_t *dt,
```
engdoreis, line 29: This function is doing exactly the same as `dif_adc_ctrl_init`, I wonder if `dif_adc_ctrl_init` should be removed to avoid confusion.

engdoreis: we could also call this function `dif_adc_ctrl_from_dt`

pamaury: We need to keep both because we cannot change all tests are once and we don't want to break existing code. There is a comment in the autogen header:
>  DEPRECATED This function exists solely for the transition to dt-based DIFs and will be removed in the future.

I think you renaming suggestion is good.

sw/device/tests/aon_timer_irq_test.c
```diff
@@ -57,8 +60,7 @@ static_assert(configUSE_PREEMPTION == 0,
               "rv_timer may be initialized already by FreeRtos");
 
 static void tick_init(void) {
-  CHECK_DIF_OK(dif_rv_timer_init(
-      mmio_region_from_addr(TOP_EARLGREY_RV_TIMER_BASE_ADDR), &rv_timer));
+  CHECK_DIF_OK(dif_rv_timer_init_dt(rv_timer_dt, &rv_timer));
```
engdoreis, line 63: I didn't find the definition of `rv_timer_dt` maybe you mean `kDtRvTimer`?

pamaury: The definition is on line 34

util/make_new_dif/templates/dif_autogen.c.tpl
pamaury: done

sw/device/lib/dif/autogen/dif_adc_ctrl_autogen.c
pamaury: I have renamed it to `dif_adc_ctrl_init_from_dt`. I think it's important to keep the `init` to make it clear that this is creating a DIF.

hw/top_earlgrey/sw/autogen/devicetables.c
pamaury: At the moment, there is no enumeration for the indices of each block, ie there is no `kDtUart0`, there is just `kDtUart[0]`. Maybe we could create such an enumeration with per-top names but I don't know if that's really useful? The only instance where it could potentially be useful is if we had a top with two instances of a block and those instances are not obvious `0` and `1` but something more meaningful.

sw/device/tests/aon_timer_irq_test.c
```diff
@@ -21,19 +22,21 @@
...
 
-#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
-
 OTTF_DEFINE_TEST_CONFIG();
 
-static const uint32_t kPlicTarget = kTopEarlgreyPlicTargetIbex0;
+static const uint32_t kPlicTarget = 0;
 static const uint32_t kTickFreqHz = 1000 * 1000;  // 1Mhz / 1us
 static dif_rv_core_ibex_t rv_core_ibex;
+static const dt_rv_core_ibex_t *rv_core_ibex_dt = &kDtRvCoreIbex[0];
```
engdoreis, line 30: These are compile time constants, so should also be named using the pattern `kRvCoreIbexDt`.

pamaury: Good point, will address that in the next push

pamaury:
I have made a change to the file structure: now instead of generating some files using `reggen` and some using `topgen`, I am generating everything using `topgen` and putting it in `hw/top_earlgrey/sw/autogen`. The rationale for this change is that:
- it's simpler on both the generator and user side,
- it will be very useful for english breakfast where files are generated but not checked in.

pamaury:
I am now able to build the test_rom for english breakfast. For those interested:
```console
./bazelisk.sh build --config=english_breakfast --config=riscv32 //sw/device/lib/testing/test_rom:test_rom
```
Note that there is no test infrastructure to EB yet so to be able to use this, it's still a very manual process:
```
./bazelisk.sh build --config=english_breakfast --config=riscv32 //sw/device/lib/testing/test_rom:test_rom
./bazelisk.sh cquery --config=english_breakfast --config=riscv32 //sw/device/lib/testing/test_rom:test_rom_sim_verilator
# copy the file test_rom_sim_verilator.32.vmem to somewhere safe, e.g /path/to/safe/test_rom_sim_verilator.32.vmem
# Then download the verilator build for EB from a CI run and put it somewhere, e.g. /path/to/safe/Vchip_englishbreakfast_verilator
# finally run:
./bazelisk.sh run //sw/host/opentitantool -- --logging=info --interface=verilator --verilator-bin=/path/to/safe/Vchip_englishbreakfast_verilator --verilator-rom=/path/to/safe/test_rom_sim_verilator.32.vmem console
```

pamaury:
I have rebased the PR, notably this included a change to english breakfast which now has two UARTs. Also going through linting fixes.

hw/ip/adc_ctrl/data/BUILD
```diff
@@ -6,19 +6,11 @@ package(default_visibility = ["//visibility:public"])
 
 load(
     "//rules:autogen.bzl",
-    "autogen_hjson_c_header",
-    "autogen_hjson_rust_header",
+    "autogen_hjson_sw_headers",
```
matutem, line 9: I thought separating the c and rust headers was useful since different code depends on them, tock for rust and all others for c.

pamaury: They are separated, this is just a macro that creates several targets: one for C, one for rust and one for DT.

matutem:
Hey Amaury, where do you expect the per-top SW files to reside? Not just the parameters, but the actual functions that will be top-specific?

hw/ip/adc_ctrl/data/BUILD
matutem: Got it, thanks.

pamaury:
> Hey Amaury, where do you expect the per-top SW files to reside? Not just the parameters, but the actual functions that will be top-specific?

Hi @matutem, for now it's all in those targets:
# build
./bazelisk.sh build //hw/top_earlgrey/sw/autogen/...
# then query files:
./bazelisk.sh cquery //hw/top_earlgrey/sw/autogen:top_earlgrey_dttool_dt_api
./bazelisk.sh cquery //hw/top_earlgrey/sw/autogen:top_earlgrey_dttool_dt_lib_hdrs
./bazelisk.sh cquery //hw/top_earlgrey/sw/autogen:top_earlgrey_dttool_dt_lib_srcs

matutem:
Regarding my last question about generated SW, this is a drawback of not committing generated files: to see the changes I will need to copy the PR, build, and inspect the generated objects: not particularly smooth. And probably more problematic, when generation changes it may be harder to view the differences, let alone see the history of those changes. I'd say that is undesirable.

pamaury:
> Regarding my last question about generated SW, this is a drawback of not committing generated files: to see the changes I will need to copy the PR, build, and inspect the generated objects: not particularly smooth. And probably more problematic, when generation changes it may be harder to view the differences, let alone see the history of those changes. I'd say that is undesirable.

I am not sure that we should be checking those files: they are very verbose but most of the code is of no particular interest, also it makes diff very large when doing changes which is not particularly helpful. I believe that what is more useful than the code itself is the documentation so that people can see how to use it. Unfortunately, our documentation system currently cannot handle generated file but this is something that we should fix. Note that we are also not checking in the repo the IP register header and we have the same problem with the documentation.

pamaury:
Closed in favour of #25807

PR #24454, opened 2024-08-30
[prim,fpv] Get rid of ASSERT_KNOWN in an FPV context

rswarbrick:
Unknown signals aren't really supported by Jasper, which drops the assertion accordingly. Unfortunately, this shows up as a coverage hole (for a rather silly reason). Get rid of the assertion entirely if we are in this context (the signal is not X because the tool doesn't allow that to happen...)


alees24 (approved):

rswarbrick:

alees24:

rswarbrick:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_assert_standard_macros.svh

alees24:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_assert_standard_macros.svh

PR #24482, opened 2024-09-01
[opentitantool]: fix hashed SPX signature verification

vbendeb:
In case of SPX signing using the PrehashedSha256 domain the image payload needs to be hashed both before signing and before validating the signature.

The validation path was not hashing the image which was causing signature validation failures.

Tested by verifying pure and hashed SPX signing cases.

Change-Id: I055d2c5717b7280d9e5a11a93e54815ad9a707cd

### DIFF
--- a/sw/host/opentitanlib/src/image/image.rs
+++ b/sw/host/opentitanlib/src/image/image.rs
@@ -5,6 +5,7 @@
 use anyhow::{bail, ensure, Result};
 use memoffset::offset_of;
 use sphincsplus::{SphincsPlus, SpxDomain, SpxPublicKey};
+use std::borrow::Cow;
 use std::collections::HashSet;
 use std::convert::TryInto;
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


sw/host/opentitanlib/src/image/image.rs
```diff
@@ -83,7 +83,12 @@ impl SigverifyParams {
     // Verify the optional SPX+ signature.
     pub fn spx_verify(&self, b: &[u8], domain: SpxDomain) -> Result<()> {
         if let Some(spx) = &self.spx_sig_params {
-            spx.key.verify(domain, &spx.signature, b)?;
+            let hash = sha256::sha256(b).to_le_bytes();
```
cfrantz, line 86: The hash calculation should be inside the match case for `PreHashedSha256`.

You probably have to use a `Cow` (copy-on-write) reference to deal with the difference between the hash result (`Vec<u8>`) vs. the slice (`[u8]`):
```
let msg = match domain {
  PreHashedSha256 => Cow::from(sha256::sha256(b).to_le_bytes()),
  _ => Cow::from(b),
}
```

vbendeb: I tried to create the hash in the match case, but it did not have long enough life time, looks like `Cow::from()` takes care of this, done.

cfrantz (approved):

timothytrippel (approved):

Issue #24446, opened 2024-08-30
[bazel] Some unittest are not runnable on host by mistake


pamaury:
### Description

While working #24443, I realize that the unittest `//sw/device/silicon_creator/lib/cert:cert_unittest` had become unrunnable on host by mistake due to the following dependency:
```
Dependency chain:
    //sw/device/silicon_creator/lib/cert:cert_unittest (a54725)
    //sw/device/silicon_creator/lib/cert:cert (97da4a)
    //sw/device/silicon_creator/lib:otbn_boot_services (97da4a)   <-- target platform (@local_config_platform//:host) didn't satisfy constraint @platforms//cpu:riscv32
```
This is now fixed by #24440 but it illustrates that it is very easy to make a unittest unrunnable on host by mistake. Unfortunately, since we run host tests using `//...`, those incompatible tests will simply be skipped in CI so they are disable without anyone realizing.

Out of curiosity, I looked at the list of test that are currently incompatible with host:
```python
# incompatible.cquery
def format(target):
  if "IncompatiblePlatformProvider" in providers(target):
    return target.label
  return ""
```
run with
```console
$ bazel cquery 'tests(//sw/device/...)' --output=starlark --starlark:file=incompatible_targets.cquery
@//sw/device/silicon_creator/lib/cert:cert_unittest
@//sw/device/silicon_creator/lib/ownership:ownership_unlock_unittest
@//sw/device/silicon_creator/lib/sigverify:sigverify_unittest
@//sw/device/silicon_creator/rom_ext/keys/fake:fake_unittest
@//sw/device/silicon_creator/rom_ext/prodc/keys:keys_unittest
@//sw/device/silicon_creator/rom_ext/sival/keys:keys_unittest
```
This means that those tests have been disabled by mistake, I think all of them due to the extra dependency on `//sw/device/silicon_creator/lib:otbn_boot_services`.

**Short-term solution**
Fix the dependency and make those tests runnable again.

**Long-term solution**
We need to ensure that we don't disable such tests silently in CI. This could either be a lint in the CI, or maybe a way to mark some tests in bazel as "must run on host".

pamaury:
@moidx @cfrantz @timothytrippel

timothytrippel:
I think a lint check is probably easiest since its a simple query?

pamaury:
Done in #24484, could you look at the other unrunnable tests? I think they are all due to otbn boot services

PR #24484, opened 2024-09-02
[ci] Add a lint to check for unrunnable tests

pamaury:
We have discovered that some unittest have become unrunnable because we added my mistake dependencies that are marked as only compatible with the OT platforms. This means that those unittests are then skipped when running all tests using //...
This commit adds a lint to make sure that all tests are compatible with the host platform.

**NOTE:** I have currently marked this lint as a warning only because we still have 6 tests that are broken at the moment.

This partially addresses #24446

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -278,6 +278,13 @@ jobs:
         --target_pattern_file="${TARGET_PATTERN_FILE}"
     displayName: Build & test SW
   - template: ci/publish-bazel-test-results.yml
+  # We run this lint in the sw_test job instead of lints because it requires an expensive
+  # cquery. By running it after we have run the tests, we can save most of the preparation
+  # time of the query.
+  - bash: ci/scripts/check-unrunnable-tests.sh
+    displayName: Check for unrunnable tests
+    condition: succeededOrFailed()
+    continueOnError: True
 
 - job: chip_englishbreakfast_verilator
   displayName: Verilated English Breakfast (Build)
--- a/ci/scripts/check-unrunnable-tests.sh
+++ b/ci/scripts/check-unrunnable-tests.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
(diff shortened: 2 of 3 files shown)


rswarbrick (approved):
Looks good to me!

jwnrt (approved):
Looks good, thanks

ci/scripts/check-unrunnable-tests.sh
```diff
@@ -0,0 +1,18 @@
...
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+# Look for all tests are incompatible with host, i.e. unrunnable.
+ci/bazelisk.sh cquery 'tests(//sw/device/...)' \
+    --output=starlark \
+    --starlark:file=ci/scripts/incompatible_targets.cquery \
+    | sed '/^$/d' | uniq | sort > output.txt
```
luismarques, line 13: `uniq | sort` -> `sort | uniq`?

luismarques (approved):

pamaury:
The query seems to be very slow, I need to investigate why. It could be because doing a cquery will build a lot of things (rust crates, python wheels, etc).

ci/scripts/check-unrunnable-tests.sh
jwnrt: `sort -u`?

pamaury:
I have moved the query to the `sw_test` job, this way the graph is already built and it massively speeds up the query: it "only" takes 3min. For some reason that I do not understand, it still builds more things than I would expect, in particular some rust crates.

azure-pipelines.yml
```diff
@@ -278,6 +278,10 @@ jobs:
         --target_pattern_file="${TARGET_PATTERN_FILE}"
     displayName: Build & test SW
   - template: ci/publish-bazel-test-results.yml
+  - bash: ci/scripts/check-unrunnable-tests.sh
```
timothytrippel, line 281: Maybe just drop a comment here why this is not run during the lint job?

timothytrippel (approved):

azure-pipelines.yml
pamaury: done

timothytrippel:
Thanks @pamaury , and really great find!

Issue #24480, opened 2024-08-30
[sysrst_ctrl] Misleading documentation and reset default for PIN_ALLOWED_CTL


jesultra:
### Description

The section [Flash Write Protect Output](https://opentitan.org/book/hw/ip/sysrst_ctrl/doc/theory_of_operation.html#flash-write-protect-output) contains the following two sentences:
> [...] The software can release flash_wp_l_o explicitly by setting [PIN_OUT_CTL.FLASH_WP_L](https://opentitan.org/book/hw/ip/sysrst_ctrl/doc/registers.html#pin_out_ctl) to 0 when needed. [...] the value of flash_wp_l_o defaults to logic 0 when it is not explicitly driven via the override function. [...]

Keeping in mind that flash_wp_l is a low-active signal, the first sentence gives the impression that if the override control bit for flash_wp_l is disabled, that the signal will revert to its deasserted high level.  However, the second sentence states that in the absense of overriding, the signal will be logic 0, that is, asserted.

In practice, the Z1 silicon seems to behave as the second sentence describes, that is, contradicting the first one.

I think that it would make more sense that the "default" level of flash_wp_l, in absence of any override would be logic 1, that is, deasserted.  This would match how the other signal ec_rst_l behaves, in the sense that in absence of override, the level of that signal is controlled by the key combination detection logic, which defaults to logic 1.

Furthermore, the default value for the register PIN_ALLOWED_CTL is to have bits flash_wp_l_0 and ec_rst_l_0 set, and nothing else.  This seems consistent with the intention behind the first sentence in the documentation (which is contrary to how Z1 silicon behaves.)  If the default level of flash_wp_l is not changed to logic 1, then the PIN_ALLOWED_CTL ought to also have flash_wp_l_1 set by default.

pamaury:
I agree that the sentence 
> The software can release flash_wp_l_o explicitly by setting [PIN_OUT_CTL.FLASH_WP_L](https://opentitan.org/book/hw/ip/sysrst_ctrl/doc/registers.html#pin_out_ctl) to 0 when needed.

seems contrary to the rest of the documentation and how it actually works. 

It would indeed be more logical that the default value of `flash_wp_l` is 1 when there is no override, that would be consistent with the reset value of `PIN_OUT_CTL` and `PIN_ALLOWED_CTL`.

By the way, there is useful code in the repo in case in helps:
[...]

PR #24485, opened 2024-09-02
[sysrst_ctrl,doc] Fix documentation of the flash_wp_l pin

pamaury:
The documentation is incorrect: since the default logic value of the pin with no override is 0, the only way to release the pin is to use an override to 1, and NOT to disable the override.

Addresses #24480 since we cannot change the RTL at this point.

### DIFF
--- a/hw/ip/sysrst_ctrl/doc/theory_of_operation.md
+++ b/hw/ip/sysrst_ctrl/doc/theory_of_operation.md
@@ -159,9 +159,9 @@ Note that since the `sysrst_ctrl` does not have control over the pad open-drain 
 ## Flash Write Protect Output
 
 Upon reset, the `flash_wp_l_o` signal will be asserted active low.
-The software can release `flash_wp_l_o` explicitly by setting [`PIN_OUT_CTL.FLASH_WP_L`](registers.md#pin_out_ctl) to 0 when needed.
 The `flash_wp_l_o` signal does have a corresponding input signal `flash_wp_l_i` - but that one is mainly intended for pad observability and does not have a bypass path to `flash_wp_l_o`.
-Hence, the value of `flash_wp_l_o` defaults to logic 0 when it is not explicitly driven via the override function.
+Hence, the value of `flash_wp_l_o` defaults to asserted low when it is not explicitly driven via the override function.
+The software can release `flash_wp_l_o` explicitly by setting an override to 1, that is setting [`PIN_ALLOWED_CTL.FLASH_WP_L_1`](registers.md#pin_allowed_ctl) to 1, [`PIN_OUT_CTL.FLASH_WP_L`](registers.md#pin_out_ctl) to 1 and [`PIN_OUT_VALUE.FLASH_WP_L`](registers.md#pin_out_value) to 1.
 
 Note that since the `sysrst_ctrl` does not have control over the pad open-drain settings, software should properly initialize the pad attributes of the corresponding pad in the [pinmux configuration](../../pinmux/README.md) before releasing `flash_wp_l_o`.
 


hw/ip/sysrst_ctrl/doc/theory_of_operation.md
```diff
@@ -159,9 +159,9 @@ Note that since the `sysrst_ctrl` does not have control over the pad open-drain
 ## Flash Write Protect Output
 
 Upon reset, the `flash_wp_l_o` signal will be asserted active low.
-The software can release `flash_wp_l_o` explicitly by setting [`PIN_OUT_CTL.FLASH_WP_L`](registers.md#pin_out_ctl) to 0 when needed.
 The `flash_wp_l_o` signal does have a corresponding input signal `flash_wp_l_i` - but that one is mainly intended for pad observability and does not have a bypass path to `flash_wp_l_o`.
 Hence, the value of `flash_wp_l_o` defaults to logic 0 when it is not explicitly driven via the override function.
```
jettr, line 163:
```suggestion
Hence, the value of `flash_wp_l_o` defaults to 0V when it is not explicitly driven via the override function.
```

I think the phrase `logic 0` for a active low signal is a little ambious. Can we use `0V` or something else like that?

a-will: No, physical values would not be correct, since the voltage is an implementation-dependent signaling property. That does not belong in block-level documentation. I would probably go with "asserted low".

Really, though, the CSR documentation is unclear because it describes flash_wp_l controls as overrides, and that's nonsensical. It's just a GPIO that is assumed to be connected to an open-drain pad.

jettr: I think "asserted low" is a great phrase to use that is clear.

pamaury: I have replaced `logic 0` with `asserted low`.

a-will: For reference of the corresponding RTL, the "non-override" value is the constant here:

https://github.com/lowRISC/opentitan/blob/908eefb9a5888c55c9baac8a6a2780746c3afe4d/hw/ip/sysrst_ctrl/rtl/sysrst_ctrl_pin.sv#L99

hw/ip/sysrst_ctrl/doc/theory_of_operation.md
```diff
@@ -159,9 +159,9 @@ Note that since the `sysrst_ctrl` does not have control over the pad open-drain
 ## Flash Write Protect Output
 
 Upon reset, the `flash_wp_l_o` signal will be asserted active low.
-The software can release `flash_wp_l_o` explicitly by setting [`PIN_OUT_CTL.FLASH_WP_L`](registers.md#pin_out_ctl) to 0 when needed.
 The `flash_wp_l_o` signal does have a corresponding input signal `flash_wp_l_i` - but that one is mainly intended for pad observability and does not have a bypass path to `flash_wp_l_o`.
-Hence, the value of `flash_wp_l_o` defaults to logic 0 when it is not explicitly driven via the override function.
+Hence, the value of `flash_wp_l_o` defaults to asserted low when it is not explicitly driven via the override function.
+The software can release `flash_wp_l_o` explicitly by setting setting an override to 1, that is setting [`PIN_ALLOWED_CTL.FLASH_WP_L_1`](registers.md#pin_allowed_ctl) to 1, [`PIN_OUT_CTL.FLASH_WP_L`](registers.md#pin_out_ctl) to 1 and [`PIN_OUT_VALUE.FLASH_WP_L`](registers.md#pin_out_value) to 1.
```
a-will, line 164: typo / duplicate: `by setting setting`

pamaury: done

jettr (approved):

jesultra (approved):

PR #24486, opened 2024-09-02
[dv/flash] Allow invalid_op test to skip special case partitions

eshapira:
Hi,
This PR is fixing some failing seeds of flash_ctrl_invalid_op test when ran in the extending env.
It add the ability for extending env to skip the info2 partition or to skip programming or erasing of RO defined partitions.
The flags used in this PR are disabled by default so it shouldn't affect any existing regressions.
Thanks!


matutem (approved):
LGTM

PR #24487, opened 2024-09-02
[prim,fpv] Tweak the prim_alert_rxtx_async tests

rswarbrick:
These tests weren't actually doing anything because we were binding them into a nonexistent module (found by looking at formal coverage).

Fix that and fix most of the broken properties. Annoyingly, `AlertCheck0_A` still doesn't quite work. This property is supposed to be putting an upper bound on the time between someone asking the alert sender to do something and the alert receiver registering that it has happened. I suspect this just needs a small tweak (since the non-async version is working fine!). But it seems a bit fiddly and I'm hoping someone who understands this primitive better than me can help...

I think this change is worth taking either way because it enables a load of testing that we weren't previously running. And most of it passes!

### DIFF
--- a/hw/ip/prim/fpv/tb/prim_alert_rxtx_async_fatal_bind_fpv.sv
+++ b/hw/ip/prim/fpv/tb/prim_alert_rxtx_async_fatal_bind_fpv.sv
@@ -5,7 +5,7 @@
 
 module prim_alert_rxtx_async_fatal_bind_fpv;
 
-  bind prim_alert_rxtx_async_fpv
+  bind prim_alert_rxtx_async_tb
         prim_alert_rxtx_async_assert_fpv prim_alert_rxtx_async_assert_fpv (
     .clk_i,
     .rst_ni,
--- a/hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
+++ b/hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
@@ -78,11 +78,11 @@ module prim_alert_rxtx_async_assert_fpv
   // be parameterized accordingly if different clock ratios are to be used here.
   // TODO: tighten bounds if possible
   sequence FullHandshake_S;
-    $rose(prim_alert_rxtx_async_tb.alert_pd)   ##[3:5]
+    $rose(prim_alert_rxtx_async_tb.alert_pd)   ##[3:6]
     $rose(prim_alert_rxtx_async_tb.ack_pd)     &&
-    $stable(prim_alert_rxtx_async_tb.alert_pd) ##[3:5]
+    $stable(prim_alert_rxtx_async_tb.alert_pd) ##[3:6]
     $fell(prim_alert_rxtx_async_tb.alert_pd)   &&
-    $stable(prim_alert_rxtx_async_tb.ack_pd)   ##[3:5]
+    $stable(prim_alert_rxtx_async_tb.ack_pd)   ##[3:6]
     $fell(prim_alert_rxtx_async_tb.ack_pd)     &&
     $stable(prim_alert_rxtx_async_tb.alert_pd);
   endsequence
--- a/hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
+++ b/hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
@@ -125,7 +125,7 @@ module prim_alert_rxtx_async_assert_fpv
   `ASSERT(AlertCheck0_A,
       !ping_req_i [*10] ##1 ($rose(alert_req_i) || $rose(alert_test_i)) && sender_is_idle |->
       ##[3:5] alert_o,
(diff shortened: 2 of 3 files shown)


hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
```diff
@@ -78,11 +78,11 @@ module prim_alert_rxtx_async_assert_fpv
   // be parameterized accordingly if different clock ratios are to be used here.
   // TODO: tighten bounds if possible
   sequence FullHandshake_S;
-    $rose(prim_alert_rxtx_async_tb.alert_pd)   ##[3:5]
+    $rose(prim_alert_rxtx_async_tb.alert_pd)   ##[3:6]
```
vogelpi: Could this be related to some ECO changes impacting the signaling of alerts? It would probably be good if @matutem could cross check this quickly.

hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
```diff
@@ -109,13 +114,32 @@ module prim_alert_rxtx_async_assert_fpv
...
+  // This bound is relatively large as in the worst case, we need to resolve staggered differential
+  // signal patterns on all three differential channels.
+  //
+  // Note 1: The complete transmission of pings only happens when no ping handshake is in progress,
+  // so we only allow the property to start when the sender isn't in a ping handshake FSM state.
+  //
+  // Note 2: The testbench allows a clock skew between the p/n bits of the ping and ack signals from
+  // the receiver to the sender. If there is such a clock skew then an edge in either signal will be
+  // detected as a signal integrity error in the sender, which will ignore the ping. This is a
+  // property we want! But it won't allow a ping to go through if there is a skew, so we don't want
```
vogelpi, line 131: I am not sure about this. I thought the protocol has been designed to tolerate a little bit of skew to simplify the back end design. Can you please check this again?

hw/ip/prim/fpv/tb/prim_alert_rxtx_async_fatal_bind_fpv.sv
```diff
@@ -5,7 +5,7 @@
 
 module prim_alert_rxtx_async_fatal_bind_fpv;
 
-  bind prim_alert_rxtx_async_fpv
+  bind prim_alert_rxtx_async_tb
```
vogelpi, line 8: This could easily be combined with the first commit in the series. It addresses the very same issue in a related file.

vogelpi (commented):
Thanks @rswarbrick ! The changes look mostly good to me, thanks for the helpful comments in the code and the commit messages.

I do have one question regarding skew in the differential ping / alert signals.

hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
```diff
@@ -43,9 +43,9 @@ module prim_alert_rxtx_async_assert_fpv
 
   logic init_pending;
   assign init_pending = mubi4_test_true_strict(init_trig_i) ||
-                        prim_alert_rxtx_async_fpv.i_prim_alert_receiver.state_q inside {
```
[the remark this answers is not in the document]
rswarbrick: Hmm. Yes, I think it should! I'll do that and re-push.

hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
rswarbrick: I'm not sure. @matutem, would you mind taking a look?

hw/ip/prim/fpv/tb/prim_alert_rxtx_async_fatal_bind_fpv.sv
rswarbrick: Good point. I'd forgotten that these are using the same assertion file.

rswarbrick:
No change in the new version except that I've squashed together some commits (thanks, @vogelpi for suggesting it)

hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
rswarbrick: Hmm, I think you're right. I suspect that the `ping_skew_i` signal was changing in the middle of the run, which would cause the stuff in `prim_diff_decode` to choke. I'll try this out today and re-push with something more sensible.

hw/ip/prim/fpv/vip/prim_alert_rxtx_async_assert_fpv.sv
vogelpi: IIRC, @matutem is currently OoO. I think we can move forward with this now but it would be good to cross check this with Matute once he's back.

vogelpi (approved):

PR #24488, opened 2024-09-02
[hw,gpio,rtl] HW Strap Sampling  (integrated_dev to master)

Razer6:
This PR syncs the GPIO IO from integrated_dev to master. This PR implements the HW strap sampling mechanism for GPIOs, which is a parameterized feature which is disabled on Earlgrey.


Razer6:
@andreaskurth Not sure exactly why lint is failing. Locally I don't see any linting errors. Unfortunately, I don't have access to the private lint CI log.

hw/ip/gpio/data/gpio.hjson
```diff
@@ -27,10 +27,10 @@
       notes:              "",
     }
     {
-      version:            "1.1.0",
```
Razer6, line 30: Maybe undo that, given it is only an addition but not a change in behaviour.

rswarbrick:
**Conclusion: Yep, this looks perfectly good to me: thanks!**

I've just worked through the PR carefully (thank goodness it's short!).
Reviewing the migration, I tried to use the automated script as 
```
  ../ot-mig/mig_pr_review.py --source-branch origin/integrated_dev \
                             --target-branch origin/master \
                             --migration-branch Razer6/sync-gpio-strap-master \
                             --path hw/ip/gpio \
                             --show-diff 
```
but this didn't work very well (I think it struggled to find the right
commits to compare because some other PRs in `integrated_dev` touch
`hw/ip/gpio`).

# Manual commit comparison

Fortunately there aren't many commits! Pairing them up explicitly gives:

## [gpio] Add snapshot functionality

*Migration commit 030e2dc; Original commit b232c01*

- One apparent difference is because the autogenerated file from
  regtool wasn't touched in the original. The migration looks right to
  me.
- Another difference is in `gpio.sv`, but this turns out to be that a
  comment has been moved to the right place. The migration version
  looks correct.
- Finally, we lose lots of auto-generated changes in hw/top_darjeeling
  (because we aren't including those files in master yet). Perfectly
  reasonable!

## [gpio] Document snapshot feature

*Migration commit ba3b8fe; Original commit c95b70e*

Everything matches here.

## [gpio] Removed unnecessary cdc synchornizer on strap_en_i signal

*Migration commit 58133e7; Original commit e65178a*

Everything matches here.

# Other gpio commits on integrated_dev

These are the commits that `mig_pr_review.py` found to touch
`hw/ip/gpio`, but manually dropping the three that have been compared
above.

- 0045a91
  + This is Bazel and SW autogen support
  + It's a project-wide change which hasn't yet been done in master,
    but doesn't belong in this PR anyway.
- 6e588cd
  + A project-wide change to how results get reported (but one that
    touches hw/ip/gpio).
- 7d6167e
  + Project-wide change to the dif library.
  + The only change in gpio is to path names.
- 538fdef
  + Version and cip_id fields. The corresponding change was reverted again on master.
- b114c66
  + Adding the Mubi parameter to prim_subreg.
  + This has already been ported (in 1b16ca2122d)

rswarbrick:
CHANGE AUTHORIZED: hw/ip/gpio/data/gpio.hjson
CHANGE AUTHORIZED: hw/ip/gpio/rtl/gpio.sv
CHANGE AUTHORIZED: hw/ip/gpio/rtl/gpio_pkg.sv
CHANGE AUTHORIZED: hw/ip/gpio/rtl/gpio_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/gpio/rtl/gpio_reg_top.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip/xbar_peri/rtl/autogen/tl_peri_pkg.sv

rswarbrick:
Ah! This is currently failing CI because of an AscentLint problem. Specifically, it's noticing that `reg2hw.hw_straps_data_in` isn't being used. I think this is probably because `GpioAsHwStrapsEn` is false in EarlGrey? This probably just needs an "unused_blah" signal to read it?

hw/ip/gpio/data/gpio.hjson
Razer6: @rswarbrick I think that should be a `1.2.0` since it is backward compatible. What about the design and verification stage. Similar to the comment from @GregAC in https://github.com/lowRISC/opentitan/pull/24489#issuecomment-2405489959 - should we leave them as they are?

Razer6: SW register layout is different though.

rswarbrick:
Hmm, this is failing quick lint. I think we might need to regenerate gpio's registers.md?

Razer6:
> Hmm, this is failing quick lint. I think we might need to regenerate gpio's registers.md?

Oh dang. I did one change. I moved the new registers to the end to not mangle the existing registers layout.

rswarbrick (approved):

Razer6:
@rswarbrick Can we get that merged?

hw/ip/gpio/data/gpio.hjson
andreaskurth: I think this should be a minor version increment because the new registers are at the end of the address space of this IP -- so existing registers don't move and existing SW continues to work. Or am I missing something?

andreaskurth (commented):
LGTM except that I have the impression that this should be a minor (because backward-compatible) version bump

hw/ip/gpio/data/gpio.hjson
Razer6: Yes. I drop that from the commit - this was changed on the integrated_dev branch. The comment was from before we had the versioning discussion.

Razer6:
Thanks @andreaskurth for looking into it. changed to version to only do a minor bump.

andreaskurth (approved):

PR #24489, opened 2024-09-02
[hw,uart,rtl] Implement LSIO trigger (integrated_dev to master)

Razer6:
In this minimal implementation, the UART core asserts iff the TX FIFO fill level is smaller than the configured threshold or the RX FIFO fill level is greater than or equal to the configured threshold, independent of IRQ enablement. The trigger is flopped to avoid CDC issues between different clock domains.

This carries over the changes from the integrated_dev to master. The trigger is not used in Earlgrey.

### DIFF
--- a/hw/ip/uart/data/uart.hjson
+++ b/hw/ip/uart/data/uart.hjson
@@ -128,6 +128,17 @@
       '''
     },
   ]
+  inter_signal_list: [
+    { struct: "logic"
+      type:   "uni"
+      name:   "lsio_trigger"
+      desc: '''
+            Self-clearing status trigger for the DMA.
+            Set when RX or TX FIFOs are past their configured watermarks matching watermark interrupt behaviour.
+      '''
+      act:    "req"
+    }
+  ]
   countermeasures: [
     { name: "BUS.INTEGRITY",
       desc: "End-to-end bus integrity scheme."
--- a/hw/ip/uart/data/uart.hjson
+++ b/hw/ip/uart/data/uart.hjson
@@ -37,7 +37,7 @@
     notes:              ""
   }
   {
-    version:            "2.0.0",
+    version:            "2.1.0",
     life_stage:         "L1",
     design_stage:       "D2S",
     verification_stage: "V2S",
--- a/hw/ip/uart/doc/block_diagram.svg
+++ b/hw/ip/uart/doc/block_diagram.svg
@@ -8,7 +8,7 @@
    stroke-miterlimit="10"
    id="svg852"
    sodipodi:docname="block_diagram.svg"
-   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
+   inkscape:version="1.3.2 (091e20e, 2023-11-25)"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns="http://www.w3.org/2000/svg"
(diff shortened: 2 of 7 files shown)


Razer6:
@rswarbrick I wonder if we should use the new status type interrupts that the LSIOs provide for the trigger functionality. What do you think about that change?

rswarbrick:
I'm not entirely sure. This is essentially just the OR of two existing watermark interrupts, right?

Presumably the reason we actually *need* this extra output is that we want to pipe it to `soc_proxy` and might want the output even when the two source interrupts are disabled?

Do we ever expect the processor to want to use this as an interrupt? If not, maybe the current implementation is sensible, letting us avoid wiring things together more.

rswarbrick (commented):
This looks sensible to me (and I think we're just picking something from `integrated_dev`?). I'll assign someone with a better claim to be a designer than me :-)

HOWEVER: We don't currently document this signal anywhere, which isn't quite right. I think we should do *something* in this commit, even if it's adding something like "PLACEHOLDER: Document lsio_trigger_o" in the documentation somewhere. Similarly, I think this signal probably needs adding to the block diagram.

Razer6:
> HOWEVER: We don't currently document this signal anywhere, which isn't quite right. I think we should do _something_ in this commit, even if it's adding something like "PLACEHOLDER: Document lsio_trigger_o" in the documentation somewhere. Similarly, I think this signal probably needs adding to the block diagram.

I added a documentation entry to the HJSO definition. I tried to modify the block diagram, though opening that on my end with inkscape just shows "paths" but no text objects?

hw/ip/uart/data/uart.hjson
```diff
@@ -128,6 +128,17 @@
       '''
     },
   ]
+  inter_signal_list: [
+    { struct: "logic"
+      type:   "uni"
+      name:   "lsio_trigger"
+      desc: '''
+            Self-clearing status trigger for the DMA.
+            Triggers when the UART reaches the configured RX or TX watermark.
```
GregAC, line 137:
```suggestion
            Set when RX or TX FIFOs are past their configured watermarks matching watermark interrupt behaviour.
```

GregAC (approved):
Happy with the change though had a suggestion on a different way to describe the new functionality. Same suggestion applies to the other LSIO trigger changes.

GregAC:
Oh and this change should probably result in a version bump to v2.1.0 from v2.0.0 as this is a new feature but a minor one that doesn't break compatibility with anything.

Technically that would mean a bump back to V1/D1 as well but as this is a small isolated change I think I'd be comfortable leaving it at D2/V2 given V2 in particular allows for small changes beyond V2 along with small coverage gaps @rswarbrick what do you think?

Razer6:
@GregAC I updated the version number. Should there also be a bump in the commit id?

rswarbrick:
About the block diagram, this is a bit sad but not your fault! I've just spent an hour redrawing it in Inkscape with actual text. The redraw is in in #24770.

rswarbrick:
About the version bump and V1/D1: it's kind of annoying, but I suppose we probably should do it. Hopefully we can rattle through the signoff to get back to D2/V2 pretty quickly.

Razer6:
> About the version bump and V1/D1: it's kind of annoying, but I suppose we probably should do it. Hopefully we can rattle through the signoff to get back to D2/V2 pretty quickly.

I would like to hold-off from that at this point. In today's meeting regarding [block-level versioning](https://docs.google.com/document/d/1LxMgKAHMSMgyZHnP6Zhg2TirO7fPdwjM4TdUtGldtG4/edit#heading=h.aouraju8tye5), we agreed on leaving the design/verification stages when bumping the minor version and then during a release bump check the stages again and sign-off. That will be discussed next silicon WG meeting in detail. That said, I want to avoid the back and forth for now.

Razer6:
I added the update to the block diagram here.

andreaskurth:
CHANGE AUTHORIZED: hw/ip/uart/data/uart.hjson
CHANGE AUTHORIZED: hw/ip/uart/rtl/uart_core.sv

This is a backward-compatible minor change of the `uart` HW IP, where `top_earlgrey.sv` only gets changed to add the new LSIO trigger output as unconnected pin of the instances of the `uart` module.

vogelpi:
CHANGE AUTHORIZED: hw/ip/uart/data/uart.hjson
CHANGE AUTHORIZED: hw/ip/uart/rtl/uart_core.sv

I agree with @andreaskurth 's assessment above.

PR #24490, opened 2024-09-03
[fpv,prim] Fix PrimCountFpv define in prim_count tests

rswarbrick:
The existing code was wrong, and it was my fault: I had thought that the Jasper "-define" argument would turn into a pre-processor define. But it doesn't! It actually defines a TCL variable that (in our case) is visible in fpv.tcl.

Fortunately, the hjson/tcl machinery has a "defines" key, which common_formal_cfg.hjson translates into an FPV_DEFINES environment variable, which then gets picked up by fpv.tcl and passed in to the JasperGold "analyze" command.

This commit switches to using that.


vogelpi (approved):
LGTM!

PR #24491, opened 2024-09-03
[prim,fpv] Strengthen CntErrReported_A assertion in prim_count.sv

rswarbrick:
The err_d signal is flopped into err_q, which appears as err_o. The existing assertion (that err_d went high) was correct, but didn't check the err_q/err_o part. This eventually translated into a coverage hole where no property had a CoI that included the "err_q <= err_d" line.

Change things so that we use err_o instead. This feels a little "less internal" anyway, which is nice, and also includes the register stage.


vogelpi (approved):
LGTM, thanks @rswarbrick !

rswarbrick:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_count.sv

PR #24494, opened 2024-09-03
[prim,fpv] Tweak DataKnown_A assertion in prim_fifo_sync.sv

rswarbrick:
This doesn't change the meaning, but it means that we now use ASSERT_KNOWN_IF instead of explicitly talking about $isunknown. Doing so will allow FPV runs (that don't support $isunknown) to avoid needing to report unprocessed code.

Note that this won't make any difference without #24454.


vogelpi (approved):
LGTM, thanks @rswarbrick .

vogelpi:

This PR touches an SVA only and doesn't impact synthesis. This is fine.

PR #24497, opened 2024-09-03
[prim,fpv] Remove initialisation before a test

rswarbrick:
This initialisation isn't really needed because the array will come up as 'x and the SboxPermutationCheck_A assertion below would fail if the end result wasn't '1.

The existing code also causes a reported coverage hole in the FPV flow, which notes that assigning sbox_perm_test to zero has no effect.

Obviously, this is a little silly: it's inside ifndef SYNTHESIS so coverage doesn't really matter! But waiving this sort of coverage hole more generally might be tricky and this particular case can just be "fixed".

### DIFF
--- a/hw/ip/prim/rtl/prim_lfsr.sv
+++ b/hw/ip/prim/rtl/prim_lfsr.sv
@@ -419,7 +419,6 @@ module prim_lfsr #(
       // Check that the permutation is indeed a permutation.
       logic [LfsrDw-1:0] sbox_perm_test;
       always_comb begin : p_perm_check
-        sbox_perm_test = '0;
         for (int k = 0; k < LfsrDw; k++) begin
           sbox_perm_test[sbox_in_indices[k]] = 1'b1;
         end


hw/ip/prim/rtl/prim_lfsr.sv
```diff
@@ -419,7 +419,6 @@ module prim_lfsr #(
       // Check that the permutation is indeed a permutation.
       logic [LfsrDw-1:0] sbox_perm_test;
       always_comb begin : p_perm_check
-        sbox_perm_test = '0;
```
vogelpi, line 422: I think this is fine but I am wondering if this statement was initially added because some tools don't like it otherwise?

While taking a look at the file, I noted there is also another instance of such a permutation check:
```sv
  // Only check this if enabled.
  if (StatePermEn) begin : gen_perm_check
    // Check that the supplied permutation is valid.
    logic [LfsrDw-1:0] lfsr_perm_test;
    initial begin : p_perm_check
      lfsr_perm_test = '0;
      for (int k = 0; k < LfsrDw; k++) begin
        lfsr_perm_test[StatePerm[k]] = 1'b1;
      end
      // All bit positions must be marked with 1.
      `ASSERT_I(PermutationCheck_A, &lfsr_perm_test)
    end
  end
```
I am wondering: does the FPV tool here not complain because of the `initial` block or maybe this permutation is not enabled in the instance checked by FPV?

vogelpi (approved):
The PR looks okay to me but I noted we have another occurrence of such a check in the same file which seems to be fine for FPV. Does it make sense to just align to that second style?

Also, we should probably rebase this PR to re-trigger CI after 3 weeks before merging.

hw/ip/prim/rtl/prim_lfsr.sv
rswarbrick: I just checked and it's the latter: the parameter was only added in 9e760a48994 (09/2020) and the parameters in the FPV testbench date back to 45b94bb73d1 (11/2019).

I guess there's a more general question about whether we're testing our prims properly across all the parametrisations in use in the project. But that's probably not something we can resolve in the PR :-)

rswarbrick:
I'm rebasing the PR so that we can re-run CI now. I'm also explicitly "allowing" the change (since the whole point is that it has no effect :-D)

CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_lfsr.sv

vogelpi:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_lfsr.sv

This PR touches lines of an RTL file that are within a `ifndef SYNTHESIS` synthesis block, meaning it does not affect what is synthesized. This is okay.

PR #24498, opened 2024-09-03
[hw,spi_host,rtl] Implement LSIO trigger (integrated_dev to master)

Razer6:
In this minimal implementation, SPI Host asserts `lsio_trigger_o` iff the TX FIFO fill level is smaller than the configured watermark or the RX FIFO fill level is greater than or equal to the configured watermark, independent of IRQ enablement.

### DIFF
--- a/hw/ip/spi_host/rtl/spi_host.sv
+++ b/hw/ip/spi_host/rtl/spi_host.sv
@@ -37,6 +37,8 @@ module spi_host
   input  spi_device_pkg::passthrough_req_t passthrough_i,
   output spi_device_pkg::passthrough_rsp_t passthrough_o,
 
+  output logic             lsio_trigger_o,
+
   output logic             intr_error_o,
   output logic             intr_spi_event_o
 );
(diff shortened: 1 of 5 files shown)


hw/ip/spi_host/rtl/spi_host.sv
```diff
@@ -37,6 +37,8 @@ module spi_host
   input  spi_device_pkg::passthrough_req_t passthrough_i,
   output spi_device_pkg::passthrough_rsp_t passthrough_o,
 
+  output logic             lsio_trigger_o,
```
andreaskurth, line 40: This output doesn't seem to get driven in the commit that adds it

Razer6: I added an assignment to this commit to make it self contained. In the next commit, the assignment gets replaced with the output flop-

andreaskurth:

This is a backward-compatible minor change of the `spi_host` HW IP, where `top_earlgrey.sv` only gets changed to add the new LSIO trigger output as unconnected pin of the instances of the `spi_host` module.

vogelpi:

PR #24503, opened 2024-09-03
[hw,sram_ctrl,rtl] Sync sram_ctrl to master (integrated_dev to master)

Razer6:
Only adds a clarification comment related to word addresses and SRAM configs used by Darjeeling.


vogelpi (commented):
Thanks for the PR @Razer6 , this looks good. I do have some questions related to the configurations, some configs seems to be equal to the existing main and ret SRAM configs.

Razer6:
@vogelpi Thanks for the review. I dropped the redundant configurations.

vogelpi (approved):
Thanks @Razer6 , this looks good!

PR #24505, opened 2024-09-03
[hw,tlul] Make a_user.rsvd width configurable (integrated_dev to master)

Razer6:
Expose the TLUL user bits to the host adapter. In Earlgrey, these bits do not convey any information.


hw/ip/tlul/rtl/tlul_adapter_host.sv
```diff
@@ -104,7 +105,7 @@ module tlul_adapter_host
     a_source:  tl_source,
     a_address: {addr_i[31:WordSize], {WordSize{1'b0}}},
     a_data:    wdata_i,
-    a_user:    '{default: '0, data_intg: wdata_intg_i, instr_type: instr_type_i},
+    a_user:    '{default: '0, data_intg: wdata_intg_i, instr_type: instr_type_i, rsvd: user_rsvd_i},
```
vogelpi, line 108: I am always confused about the order of bits in this notation. Can you please confirm adding the `user_rsvd_i` bits at the end here, doesn't change the position of the other bits?

hw/ip/tlul/rtl/tlul_pkg.sv
```diff
@@ -33,6 +33,7 @@ package tlul_pkg;
   parameter int DataMaxWidth    = 32;
   parameter int DataIntgWidth   = 7;
   parameter int DataFullWidth   = DataMaxWidth + DataIntgWidth;
+  parameter int RsvdWidth       = 10;
```
vogelpi, line 36: I suggest reducing this to 5 for the time being to not change to not impact Earlgrey at this point.

vogelpi (commented):
Thanks for the PR @Razer6 . I don't mind these changes but have two comments regarding reducing impact on Earlgrey for the time being. I'll leave the final decision on this to @andreaskurth who has a better overview of when what should be merged into master.

hw/ip/tlul/rtl/tlul_pkg.sv
Razer6: Is there a way to make a top-level definition in the HJSON to materialize a top-specific constant for that?

vogelpi: I think not yet, but this is something we need in the future.

Razer6: Yep, needed indeed.

hw/ip/tlul/rtl/tlul_adapter_host.sv
Razer6: I need to check. I was adding it next to `instr_type` as it is placed next to it in the struct definition as well. Since it is named, it shouldn't matter?

hw/ip/tlul/rtl/tlul_pkg.sv
Razer6: There is already a `top_pkg` that defines the TLUL parameters. I used that to derive the size of the `rsvd` bits. Note, I had to increase the definition to match what was currently implemented in RTL.

```
DataIntgWidth (7)  + H2DCmdIntgWidth (7) + mubi4_t (4) + rsvd (4) = 22
```

Razer6:
@vogelpi Thanks for looking into this. Could you give it another look?

hw/ip/tlul/rtl/tlul_adapter_host.sv
vogelpi: Ah, yes you're right, this should work. We don't change the order of the fields, just the width of one field can change. I was worrying if in some case we don't index the struct via member names (we of course shouldn't do that but you never know). This looks good.

hw/top_earlgrey/rtl/top_pkg.sv
```diff
@@ -9,7 +9,7 @@ localparam int TL_AW=32;
 localparam int TL_DW=32;    // = TL_DBW * 8; TL_DBW must be a power-of-two
 localparam int TL_AIW=8;    // a_source, d_source
 localparam int TL_DIW=1;    // d_sink
-localparam int TL_AUW=21;   // a_user
+localparam int TL_AUW=22;   // a_user
```
vogelpi, line 12: Ouf, this was actually wrong for a long time! It turns out we were not using the parameter anywhere though. Thanks for fixing this!

BTW: I think it's still wrong right now. Because:
- The number of reserved bits was previously 4+1 = 5.
-  Now it's 22 - 4 - 7 - 7 = 4.

So, if we don't want to change the Earlgrey design, it should be 23 here.

vogelpi (commented):
Thanks @Razer6 for making the RsvdWidth parametrizable through the top package, this is nice!

I think right now, the TL_AUW parameter for Earlgrey is wrong. It should be increased by one - see my comment.

hw/ip/tlul/rtl/tlul_pkg.sv
vogelpi: Thanks, this is really great! Except that rsvd was 5 bits previously (it's 4 down to 0).

hw/top_earlgrey/rtl/top_pkg.sv
Razer6: Of course, it's bits. Although the above definition was wrong for a long time, it was never used at any place in the code.

Razer6:
@vogelpi Thanks for your review. I updated the PR and CI is green now.

vogelpi (approved):
Thanks @Razer6 , this is great!

vogelpi:
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl.sv
CHANGE AUTHORIZED: hw/ip/rv_core_ibex/rtl/rv_core_ibex.sv
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_adapter_host.sv
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_pkg.sv

This PR touches a couple of RTL files relevant for Earlgrey but doesn't change the functionality. It just allows passing reserved user bits into the TL-UL adapter but the bits are actually zero as before. This is fine.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl.sv
CHANGE AUTHORIZED: hw/ip/rv_core_ibex/rtl/rv_core_ibex.sv
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_adapter_host.sv
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_pkg.sv

PR #24507, opened 2024-09-04
[hw,lc_ctrl,rtl] Unified LC CTRL for JTAG TAP and TLUL-based DMI

Razer6:
This PR cherry picks the changes from Darjeeling and adds them under a dedicated parameter to select between the JTAG Tap and the TLUL-based DMI interface.

When merging this, we might bump the version number of the IP.


hw/top_earlgrey/sw/autogen/top_earlgrey.h
```diff
@@ -1775,8 +1793,8 @@ typedef enum top_earlgrey_hintable_clocks {
  * configuration space, i.e. ROM, main SRAM, and flash are excluded but
  * retention SRAM, spi_device memory, or usbdev memory are included.
  */
-#define TOP_EARLGREY_MMIO_BASE_ADDR 0x40000000u
-#define TOP_EARLGREY_MMIO_SIZE_BYTES 0x10000000u
+#define TOP_EARLGREY_MMIO_BASE_ADDR 0x0u
```
Razer6, line 1796: That change is because the DMI interface is not mapped in Earlgrey. That change will be reverted once https://github.com/lowRISC/opentitan/pull/24718 gets merged. This PR allows the generically add devices/interfaces to the MMIO region. For Earlgrey, the DMI won't be added and thus also not considered for the MMIO region computation.

Razer6: This is fixed on master.

hw/ip/lc_ctrl/dv/tb.sv
```diff
@@ -43,6 +43,10 @@ module tb;
   assign lc_ctrl_if.otp_vendor_test_ctrl_o = otp_vendor_test_ctrl;
   assign otp_vendor_test_status = lc_ctrl_if.otp_vendor_test_status_i;
 
+  // Used for JTAG DTM connections via TL-UL.
```
rswarbrick, line 48: I think that maybe this commit ("Update DV environment") should be squashed with the first one. Otherwise, you end up with an intermediate state where the tests won't actually build.

I was going to suggest putting some stub version in the first commit, but I've looked at this commit more carefully and I'm now convinced it's easiest just to squash it in.

hw/ip/lc_ctrl/data/lc_ctrl.hjson
```diff
@@ -689,726 +689,727 @@
...
+                  "excl:CsrAllTests:CsrExclCheck"],
+          fields: [
+            { bits: "31:0"
+            }
+          ]
+        }
+      }
+    ]
+  }
 }
```
rswarbrick, line 1410: Nit: I'm not sure the indentation is right here: the "registers: " line on line 692 hasn't moved horizontally but now the final closing brace is 2 characters to the left.

hw/ip/lc_ctrl/data/lc_ctrl.hjson
```diff
@@ -25,7 +25,7 @@
     {clock: "clk_kmac_i", reset: "rst_kmac_ni"}
   ]
   bus_interfaces: [
-    { protocol: "tlul", direction: "device" }
+    { protocol: "tlul", direction: "device", name: "regs" }
```
rswarbrick: Crikey, this commit was big! If the rest of this PR gets held up at all, it might be worth splitting out this commit and getting it merged. I think it looks right, but it was long and I'd rather not read through it again :-)

hw/dv/sv/jtag_agent/jtag_dtm_reg_block.sv
```diff
@@ -184,7 +184,11 @@ class jtag_dtm_reg_dtmcs extends jtag_dtm_base_reg;
       .access("RO"),
       .mubi_access("NONE"),
       .volatile(0),
+`ifdef USE_DMI_INTERFACE
```
rswarbrick, line 187: I wonder whether it would make sense to give the agent something like a "configure_use_dmi" function? Then we wouldn't need all the ifdefs: we could just call it with the correct value after construction, avoiding tangling the config so much with the construction.

Working through the items:
- For the `abits` field, I think we'd just need `set_reset`?
- For the `n_bits` default value override, this just goes through to the `uvm_reg` constructor. I think we can just use the larger value (with a comment)?
- For the `address` field, I *think* we could just use the larger value again. It's the top field in the register, so I don't think that would cause a problem.

hw/ip/lc_ctrl/dv/tb.sv
```diff
@@ -35,6 +35,8 @@ module tb;
   `include "uvm_macros.svh"
   `include "dv_macros.svh"
 
+
```
rswarbrick, line 38: Nit: Stray newlines?

hw/ip/lc_ctrl/dv/tb.sv
```diff
@@ -96,6 +98,7 @@ module tb;
 
   `DV_ALERT_IF_CONNECT()
 
+`ifdef USE_DMI_INTERFACE
```
rswarbrick, line 101: Would it make sense to pull this into a localparam further up in the file with something like
```systemverilog
`ifdef USE_DMI_INTERFACE
  localparam bit use_dmi_interface = 1'b1;
`else
  localparam bit use_dmi_interface = 1'b0;
`endif
```

That would cut down the amount of stuff that needed ifdef code.

hw/top_earlgrey/doc/design/README.md
```diff
@@ -404,6 +404,7 @@ For the purpose of `top_earlgrey`, the first option has been chosen to benefit s
 | otp_ctrl          | otp_ctrl      | 0x40130000 (core) |
 |                   |               | 0x40138000 (prim) |
 | lc_ctrl           | lc_ctrl       | 0x40140000 (regs) |
+|                   |               | 0 (dmi)           |
```
rswarbrick: It clearly doesn't matter, but maybe worth getting the "(dmi)" bit to line up with the other items? Maybe by expressing the address as `0x00000000`?

Razer6: I changed the address to 0x00000000 but it will stick there with the lcctrl.

hw/dv/sv/jtag_agent/jtag_dtm_reg_block.sv
Razer6: Could you help me in doing that on this PR?

hw/ip/lc_ctrl/data/lc_ctrl.hjson
rswarbrick: You're completely right: I think I must have misread the diff. Sorry!

hw/dv/sv/jtag_agent/jtag_dtm_reg_block.sv
rswarbrick: We chatted a bit about this in person earlier. Maybe the easiest approach would be to keep with the current code for now (since we know it works), but could you open an issue to track the fact that this could be tidied up?

hw/ip/lc_ctrl/dv/tb.sv
rswarbrick: Similarly, could you add a link to this note in the issue you create?

hw/dv/sv/jtag_agent/jtag_dtm_reg_block.sv
Razer6: I created https://github.com/lowRISC/opentitan/issues/24852 to track that improvements.

rswarbrick (approved):
This looks good to me, and thanks for the tweaks you've been adding after my questions :-)

rswarbrick:
Gah! But I think the lint check is failing (CMDGEN blocks). Would you mind sorting that out?

hw/dv/sv/jtag_agent/jtag_dtm_reg_block.sv
Razer6, line 187: As pointed out by @rswarbrick, the ifdef is not working in this way. When defining the variable with 0 or 1, both cases define the variable, thus always ending up in the true case.

andreaskurth: Do we need to create an issue to track this?

andreaskurth:
CHANGE AUTHORIZED: hw/ip/lc_ctrl/data/lc_ctrl.hjson
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl.sv
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl_dmi_reg_top.sv
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl_regs_reg_top.sv
CHANGE AUTHORIZED: hw/top_earlgrey/data/xbar_peri.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip/xbar_peri/rtl/autogen/tl_peri_pkg.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip/xbar_peri/rtl/autogen/xbar_peri.sv

The changes in this PR are implemented in a backward-compatible way for SW as well as for existing Earlgrey HW.

hw/dv/sv/jtag_agent/jtag_dtm_reg_block.sv
Razer6: Right now the JTAG config does not set the define, whereas the DMI config does. I think we are good?

rswarbrick:
CHANGE AUTHORIZED: hw/ip/lc_ctrl/data/lc_ctrl.hjson
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl.sv
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl_dmi_reg_top.sv
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/lc_ctrl/rtl/lc_ctrl_regs_reg_top.sv
CHANGE AUTHORIZED: hw/top_earlgrey/data/xbar_peri.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip/xbar_peri/rtl/autogen/tl_peri_pkg.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip/xbar_peri/rtl/autogen/xbar_peri.sv

PR #24508, opened 2024-09-04
[hw,rv_core_ibex,doc] Sync Rom Patching to master (indegrated_dev to master)

Razer6:
This PR syncs the Rom Patching documentation back from integrated_dev to master. It further increases the bounds of IbexIcacheScrambleKeyRequestAfterFence_i. The PMP configuration has not yet been changed since the configuration will be done differently after https://github.com/lowRISC/ibex/pull/2147 is merged and vendored.


hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,459 @@
...
+
+| Term  | Description                         |
+|-------|-------------------------------------|
+| SoC   | System On Chip                      |
+| DFx   | Design for Test and Debug           |
+| DFT   | Design for Test                     |
+
+## Scope
+
+This document covers the boot mechanism for the Integrated OpenTitan.
```
rswarbrick: It seems a bit strange to me that we're getting a copy of this in `top_earlgrey` and `top_englishbreakfast`! Is there anywhere sensible for the document to live that isn't in the template directory?

I realise that there are some technical problems with doing something like "Take a copy of this directory if XYZ is true" at the moment, but I'd suggest the better way to do things would be to only instantiate one copy (in a situation where it makes sense).

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,459 @@
...
+## Overview
+
+Integrated Open Titan is a root of trust for measurement.
+It shall be integrated within an SoC.
+It impacts the SoC level design SoC secure boot, debug authentication etc and provides the framework necessary to bootstrap the SoC computing elements in a secure & trusted fashion.
+This document covers the changes necessary for the OpenTitan to boot first while also allowing an SoC with integrated OpenTitan to add small customizations to the boot path configuration that fit with the SoC topology.
+A few examples are voltage ramp up sequence, clock settings etc.
+Such customizations are expected to
+
+- Be non-proprietary such that it can be made available in the open source environment for audit.
```
rswarbrick: Grammar nit: Should be "... such that they can be made ..."

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,459 @@
...
+- Involve minimal flow required to get the path to the external flash enabled and provide connectivity to the external flash under OpenTitan control.
```
rswarbrick: I actually don't understand this text. Maybe "Only require minimal flow to enable a path to external flash that is under OpenTitan control" ? (Or maybe something else! I'm guessing...)

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,459 @@
...
+
+Flows involving reset and bring up for the rest of the SoC, which may be potentially more complex and involve proprietary code, are expected to be handled by other computing element(s) that are bootstrapped by the integrated OpenTitan and running authenticated / measured code.
+
+This proposal introduces a second OpenTitan boot ROM partition with e-fuse/OTP based patching mechanism that enables each SoC to add such small customizations, provide critical survivability hooks while also removing the dependency on the integrated flash component.
```
rswarbrick: Grammar nit: Because there are just two items, I think it should be "... small customizations and provide critical ..."

rswarbrick (commented):
I started from the first commit... and then realised that there are lots of "fix-up" patches from me that come afterwards! (It's been a couple of years!)

My first comment (about not making copies that don't make sense) stands, and I really think it would be worth simplifying things in that way.

After that, I'd suggest squashing together all of my patches as part of the initial commit, assuming you are happy that the change describes the current state of affairs: I'm not sure that the detailed history of the document will be of any interest in the future.

Sorry, I realise that this makes porting it from the `integrated_dev` branch less of a trivial process, but hopefully it won't be too much work.

Razer6:
> After that, I'd suggest squashing together all of my patches as part of the initial commit, assuming you are happy that the change describes the current state of affairs: I'm not sure that the detailed history of the document will be of any interest in the future.
> 
> Sorry, I realise that this makes porting it from the `integrated_dev` branch less of a trivial process, but hopefully it won't be too much work.

When we agreed on the backporting a year ago, we decided we want to keep the history of the changes. I can squash them for the document though if needed.

rswarbrick:
Thanks for squashing the document items. I noticed a couple of minor fix-ups in the document (they've got comments). I'd probably suggest squashing the fix-ups into the document patch, but I don't have very strong feelings either way.

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
Razer6: Sounds good.

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
Razer6: At the moment that's the way ipgen IPs are organized. That may change in the future but I don't want to fix that in this docs PR at this point.

Razer6:
@rswarbrick Can we get that merged?

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+| DFT   | Design for Test                     |
+
+## Scope
+
+This document covers the boot mechanism for Integrated OpenTitan.
+This includes the execution of the OpenTitan ROM, associated patch support in efuses / OTP for survivability and the code fetch from external flash.
+
+## Overview
+
+The integrated Open Titan is a root of trust for measurement, integrated within the SoC.
```
rswarbrick, line 18: I think this should probably be "OpenTitan"?

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+This document covers the changes necessary for the OpenTitan to boot first while also allowing an SoC with integrated OpenTitan to add small customizations to the boot path configuration that fit with the SoC topology.
+A few examples are voltage ramp up sequence, clock settings etc.
+Such customizations are expected to
+
+- Be non-proprietary such that they can be made available in the open source environment for audit.
+- Only require minimal flow to enable a path to external flash that is under OpenTitan control.
+
+Flows involving reset and bring up for the rest of the SoC, which may be potentially more complex and involve proprietary code, are expected to be handled by other computing element(s) that are bootstrapped by the integrated OpenTitan and running authenticated / measured code.
+
+This proposal introduces a second OpenTitan boot ROM partition with efuse/OTP based patching mechanism that enables each SoC to add such small customizations and provide critical survivability hooks while also removing the dependency on the integrated flash component.
```
rswarbrick, line 29: Nit: I think we're missing an article. Should it be "... with an efuse/OTP based ..." ?

rswarbrick, line 29: Another nit: "... while also removing the dependency ..." sounds like we're talking about a particular dependency. I don't think I know an example at the moment? I think we can tweak the text and avoid a long sentence with something like this?
```
This proposal introduces a second OpenTitan boot ROM partition with an efuse/OTP based patching mechanism.
This combination enables each SoC to add such small customizations and provide critical survivability hooks without needing an integrated flash component.
```

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+Integrated flash technology may not be available for SoCs built using advanced process nodes.
+
+Two models for OpenTitan based ROM patch deployment are envisioned:
+
+- **Factory patching** - here the ROM patches are programmed into the eFuses in the factory.
+  This may be to accommodate post-silicon bug fixes without needing a new step of the silicon.
+  Performed through the secure manufacturing flow.
```
rswarbrick, line 36: Grammar nit: I think you probably want "Performed in the ..." here. Otherwise, it reads a bit like "Performed throughout", which is definitely not what you intended.

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+- Only require minimal flow to enable a path to external flash that is under OpenTitan control.
+
+Flows involving reset and bring up for the rest of the SoC, which may be potentially more complex and involve proprietary code, are expected to be handled by other computing element(s) that are bootstrapped by the integrated OpenTitan and running authenticated / measured code.
+
+This proposal introduces a second OpenTitan boot ROM partition with efuse/OTP based patching mechanism that enables each SoC to add such small customizations and provide critical survivability hooks while also removing the dependency on the integrated flash component.
+Integrated flash technology may not be available for SoCs built using advanced process nodes.
+
+Two models for OpenTitan based ROM patch deployment are envisioned:
+
+- **Factory patching** - here the ROM patches are programmed into the eFuses in the factory.
```
rswarbrick, line 34: Grammar nit: In the next item, you start a new sentence after the title and I think that probably looks nicer. The hyphen looks a bit "run-on" though.

Maybe something like this would be better?
```
- **Factory patching** Here, the ROM patches ...
```

The result:
image

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+  This may be to accommodate post-silicon bug fixes without needing a new step of the silicon.
+  Performed through the secure manufacturing flow.
+- **In-field patching** - The capability may be used to recover parts in the field but that requires an ability to burn eFuses in the field.
```
rswarbrick, line 37: Another grammar nit (sorry): The "... but that ..." section here comes as a bit of a surprise to the reader. I tried to see whether a comma helped, but the result was rather stilted. The repetition of "in the field" is also a little awkward. Would something like this work?
```
- **In-field patching** Given the ability to burn eFuses in the field, in-field patching can be used to recover parts after deployment.
```

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+Note that patch structure for both deployment models is the same.
+An infield patch can be viewed as a new revision of the OTP patch.
```
rswarbrick, line 40: Probably better to use "in-field" to match the text above.

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+The patch in eFuses will be signed using an appropriate PK signing algorithm
```
rswarbrick, line 41: A couple of things:
- Can you expand "PK" here? I assume it's meant to be "Public Key"? But we probably need to name an example.
- I was initially confused about this text because I thought "the patch in eFuses" was just one of the two options above. I'm not sure we need "in eFuses" in the sentence at all, do we? I think that's the topic of the whole document?

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+### Programming method:
+
+- If prior to the PROD lifecycle stage: Using the manufacturing efuse programming interface.
+- In PROD lifecycle stage: Secure, verified patch update programming managed by secure application running on OpenTitan Ibex core (e.g., update patch revision in factory after provisioning, backup in-field patching mechanism).
```
rswarbrick, line 46: Could you try to improve the structure of this text?
- Neither bullet point contains a sentence (and I think they should have at least one each!)
- There are lots of impressive words (secure, verified, secure again) associated with the second point. If there is an example, it needs describing. If not, I suggest dropping some of the adjectives.
- Please add a full stop or two. This is an currently a very long run-on-not-quite-sentence.

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+The factory provisioning model is expected to be the mainstream deployment model that covers bugs found throughout the product development lifecycle until product deployment phase.
+The in-field patching capability is primarily a backup mechanism used only in very rare cases to patch a critical ROM bug escape in the field, especially some hard to catch security or functional bugs.
```
rswarbrick, line 49: Very long sentence, and I had to read it several times. Can it split up like this?
```
The in-field patching capability is primarily a backup mechanism.
It will only be used in very rare cases to patch a critical ROM bug escape in the field.
An example of that might be a hard-to-catch security or functional bug.
```

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+**It is advised to avoid reliance on the in-field ROM patch as much as possible.**
+
+This document also contains a detailed specification for the OpenTitan ROM patching architecture, the programming interfaces, and example sequences for patching ROM firmware.
```
rswarbrick, line 52: The "also" here is a bit odd, because it looks like it applies to the advice on the previous line, which isn't what you mean. I think there's also a slight problem with the English (because it looks like we're claiming a detailed specification of example sequences). I'd suggest being a bit more explicit at the start and turning the list into bullets:

```
As well as describing the form of a ROM patch, this document contains:
- A specification of the OpenTitan ROM patching architecture.
- A description of how to use a ROM patch.
- Example sequences for patching ROM firmware.
```

(I'm not completely sure about item 2, though! This was me trying to guess what "the programming interfaces" meant. Since this ends up being a mini-TOC anyway, maybe a couple of internal links would help).

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+## ROM Boot & Patching Building Blocks
+
+![Integrated OpenTitan and Boot Building Blocks](building-blocks.svg)
+
+Following changes are envisioned to allow OpenTitan based boot customizations for an SoC to configure the access path to the external flash:
```
rswarbrick, line 58: I genuinely got a little lost when trying to get my head around the text from here. I think it would be a really good idea to add another "mini-TOC" here, to give a reader some chance of figuring out what's in this section.

Also, I think the heading levels might be a bit messed up. This is currently the same level as "Overview" and then we just have "###" items until we get to "OpenTitan ROM boot flow", which is "#".

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+- Base ROM design is common across all integrated OpenTitan instances.
+
+- Supports security hardened functions such as signature verification, verification & loading of second ROM partition patch, other hardened crypto functions required at boot etc.
+  Integrated OpenTitan reuses the majority of current base ROM implementation.
+
+- Base ROM is not patchable.
+  Guaranteed to be fully verified and to be free of bugs.
+  To achieve this goal, it is kept simple by design.
+
+- Following is a summary of a few additional changes required (detailed out in later sections)
```
rswarbrick, line 71: Additional to what? The text doesn't quite work here.

My guess is that this is supposed to be saying "Once the majority of the base ROM is complete, it must run the following tasks before jumping to the second ROM partition" ?

(Minor nit: Also, I think the text in parens could just be "detailed in ...")

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+  - Identify latest patch in OTP space.
+
+  - Signature verification of OTP patch.
```
rswarbrick, line 75: The previous and next items are commands but this is a noun. Maybe "Verify the signature of that OTP patch"?

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+- Supports security hardened functions such as signature verification, verification & loading of second ROM partition patch, other hardened crypto functions required at boot etc.
+  Integrated OpenTitan reuses the majority of current base ROM implementation.
+
+- Base ROM is not patchable.
+  Guaranteed to be fully verified and to be free of bugs.
+  To achieve this goal, it is kept simple by design.
+
+- Following is a summary of a few additional changes required (detailed out in later sections)
+
+  - Identify latest patch in OTP space.
```
rswarbrick, line 73: To make this a full sentence, we need an article. "Identify the latest patch in OTP space" ?

But what is "OTP space"? This either needs defining or we need a synonym. (Or possibly just drop "space"?)

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+  - Signature verification of OTP patch.
+
+  - Load & enable patch prior to second ROM partition execution.
```
rswarbrick, line 77: Again, this would benefit from an article. We also don't really need the ampersand. How about "Load and enable the patch".

I think the "prior to ..." text is probably not needed: isn't that implicit in what the bullet point is describing anyway?

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+### Second ROM partition
+
+- Separate ROM structure to house SoC specific boot customizations.
```
rswarbrick, line 81: I know that we're using bullet points rather than paragraphs in this document, but we need to at least put sentences on the bullets, otherwise things look rather sloppy.

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+- Late binding security functionality to be tested & hardened prior to incorporating in base ROM.
```
rswarbrick, line 83: Please turn this ampersand (and later ones) into "and".

rswarbrick, line 83: This doesn't quite work as English. I think the passive voice is probably the easiest way to avoid saying who is doing the incorporating:
```
- This contains more recent security functionality.
  Putting such code in the second partition allows it to be tested and hardened before it is incorporated in the base ROM.
```

hw/ip_templates/rv_core_ibex/doc/boot-rom-patching.md
```diff
@@ -0,0 +1,476 @@
...
+
+- Second ROM partition is patchable through efuse / OTP based code patches for survivability.
```
rswarbrick, line 85: Since this document describes the patch mechanism, I don't think we need to list the different categories of patches as part of this sentence fragment. Couldn't this bullet shorten to something like this?
```
- The second ROM partition is patchable.
  This provides hooks to fix critical functional (e.g. early reset sequence) and/or security bugs.
```

PR #24509, opened 2024-09-04
[prim,lint] Remove dead waivers for prim_fifo_*sync.sv

rswarbrick:
The first can never match (because the subtraction was changed from -3 to -2 in 7cbebc2e690 in June 2021).

The second *has* never matched: I think the variable in question has been called unused_decval_msb since the initial commit in the repository.

I think the third matched some width arithmetic that was done on pointer values until commit 64272d832be replaced the inline logic with a prim_fifo_sync_cnt instance. (The waiver came from an initial bunch of waivers in 2020 and the comment doesn't make sense: it looks like it was accidentally copied from the previous waiver)

### DIFF
--- a/hw/ip/prim/lint/prim_fifo.waiver
+++ b/hw/ip/prim/lint/prim_fifo.waiver
@@ -10,19 +10,7 @@ waive -rules {ONE_BIT_MEM_WIDTH} -location {prim_fifo_sync.sv} -msg {Memory 'gen
 waive -rules {INPUT_NOT_READ} -location {prim_fifo_sync.sv} -regexp {Input port '(clk_i|rst_ni)' is not read from, instance.*Depth=0\)} \
       -comment "In passthrough mode, clk and reset are not read form within this module"
 
-
-waive -rules {ASSIGN_SIGN} -location {prim_fifo_async.sv} -msg {Signed target 'i' assigned unsigned value 'PTR_WIDTH - 3'} \
-      -comment "Parameter PTR_WIDTH is unsigned, but integer i is signed. This is fine. Changing the integer to unsigned might \
-                cause issues with the for loop never exiting, because an unsigned integer can never become < 0."
-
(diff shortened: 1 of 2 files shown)


hw/ip/prim/lint/prim_fifo.waiver
```diff
@@ -10,21 +10,9 @@ waive -rules {ONE_BIT_MEM_WIDTH} -location {prim_fifo_sync.sv} -msg {Memory 'gen
 waive -rules {INPUT_NOT_READ} -location {prim_fifo_sync.sv} -regexp {Input port '(clk_i|rst_ni)' is not read from, instance.*Depth=0\)} \
       -comment "In passthrough mode, clk and reset are not read form within this module"
 
-
-waive -rules {ASSIGN_SIGN} -location {prim_fifo_async.sv} -msg {Signed target 'i' assigned unsigned value 'PTR_WIDTH - 3'} \
```
alees24, line 14: Is a similar waiver still required? The solution really is to introduce an explicit cast in the RTL, and the explanatory comment is correct that a signed type must be used here. Or the loop could emit dec_tmp[i-1] instead.

alees24 (approved):

hw/ip/prim/lint/prim_fifo.waiver
rswarbrick: Good point. Let's just make it more sensible: I'll force-push with a cleaner solution in a sec.

alees24 (approved):
Thanks.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_fifo_async.sv

The change to the code here is rewording a loop counter to avoid hitting the edge of the unsigned region. The change is "pretty obviously right" and the code will also be well tested: if there was a mistake, the "gray and back again" path would be broken and lots of SPI block tests would start failing (before any potential tapeout!)

PR #24518, opened 2024-09-05
[bazel] Ensure output files of outquery are unique

### LINKED ISSUE #23990: [bazel] Bazel outquery gives multiple repeated results for outquery/outquery-one after building a target
## Description
After building a target (via `./bazelisk.sh build`), `./bazeilsk.sh outquery` and `./bazelisk.sh outquery-one` will then return multiple duplicated results for the same target, which can cause unintended consequences in scripts that use this and rely on it only having a single output such as [`ci/scripts/target-location.sh`](https://github.com/lowRISC/opentitan/blob/39211701b5d2d8af1d52ab243b5b079bec36535f/ci/scripts/target-location.sh#L33C1-L37C36). This can cause some commands that we would expect to work to break.

## Steps to reproduce
Check for the correct initial state by running one of the following (these are identical implementations) at the repo top e.g.:
```bash
./bazelisk.sh outquery //sw/device/tests:uart_smoketest_fpga_cw340_rom_ext
./bazelisk.sh outquery-one //sw/device/tests:uart_smoketest_fpga_cw340_rom_ext
```
This should return one location, e.g. in my case
`bazel-out/k8-fastbuild-ST-2cc462681f62/bin/sw/device/tests/uart_smoketest_fpga_cw340_rom_ext.bash`

Then, run (make sure vivado is available first) e.g.
```bash
./bazelisk.sh build //sw/device/tests:uart_smoketest_fpga_cw340_rom_ext
```
After this builds correctly, once again run one of the outquery commands. The result should now be two repeated locations:
`bazel-out/k8-fastbuild-ST-2cc462681f62/bin/sw/device/tests/uart_smoketest_fpga_cw340_rom_ext.bash`
`bazel-out/k8-fastbuild-ST-2cc462681f62/bin/sw/device/tests/uart_smoketest_fpga_cw340_rom_ext.bash`

## The problem
Everything points to this being an issue with bazel's `cquery` (or the way we use it) when supplying the `--output=starlark --starlark:expr="$qexpr"` options here: https://github.com/lowRISC/opentitan/blob/39211701b5d2d8af1d52ab243b5b079bec36535f/bazelisk.sh#L105
Replacing this with an equivalent `--output=files` will return the correct single result as we would expect. However, the starlark postprocessing is being used to do some more complex output processing for other forms of `outquery`, such as `outquery-all`, `outquery.*`, `outquery-*`etc.

A few possible thoughts:
  - Figure out why this behaviour is happening with `--output=starlark` and not `--output=files`?
  - A simple solution that would work with all existing functionality would be to append ` | sort | uniq` to the end of the `cquery`. This should work just fine as far as I can tell but I'd need to check, and it could perhaps prevent the ability to intentionally duplicate files via starlark postprocessing (though I don't think the `starlark:expr` is ever exposed to user parameters here). 
  - Another way would be to replace with `--output=files` and replace the starlark post-processing with some other functionality
  - There might also be a way to stop bazel's `cquery` from finding its own built files to stop this behaviour happening?

AlexJones0:
This PR resolves #23990.

After building a software target using `./bazelisk.sh build`, sometimes an `opentitan_test` would be built in multiple configurations, causing Bazel's `cquery` to return the same file multiple times in its output, as a result of us using `--output=starlark` for starlark postprocessing instead of the newer `--output=files`.

Some scripts, such as [`ci/scripts/target-location.sh`](https://github.com/lowRISC/opentitan/blob/39211701b5d2d8af1d52ab243b5b079bec36535f/ci/scripts/target-location.sh#L33C1-L37C36), rely on the fact that outquery will return a unique result. This ensures that these scripts will not break if a test target is built, by ensuring that output files are unique and non-duplicated.

To test, run the following commands:
```sh
./bazelisk.sh build //sw/device/tests:uart_smoketest_fpga_cw340_rom_ext
./bazelisk.sh outquery //sw/device/tests:uart_smoketest_fpga_cw340_rom_ext
```
Before this PR, you would get a duplicated output. With this PR, there should be a single unique output.

### DIFF
--- a/bazelisk.sh
+++ b/bazelisk.sh
@@ -103,7 +103,8 @@ function do_outquery() {
 
     "$file" "${pre_cmd_args[@]}" cquery "$@" \
         --output=starlark --starlark:expr="$qexpr" \
-        --ui_event_filters=-info --noshow_progress
+        --ui_event_filters=-info --noshow_progress \
+        | sort | uniq
 }
 
 function main() {


AlexJones0:
CI failures should be unrelated - not 100% confident though so do let me know if that's not the case.

jwnrt (approved):
LGTM, thanks :)

FYI `sort` has a `—unique`/`-u` flag

AlexJones0:
> FYI `sort` has a `—unique`/`-u` flag

Interestingly, from a quick search it seems there is a minor difference in that `sort` uses the locale's collating algorithm (normally `strcoll`) whereas `uniq` checks byte-value identity, and so there are technically a few (cursed) characters that could be used in file names where I believe that `sort -u` would not work as expected whereas `sort | uniq` would. But in this case either should be appropriate.

pamaury (approved):
Looks good to me. For our use case, I don't think there is a difference between `sort -u` and `sort | uniq` but I will leave it to you.

jwnrt:
Oh please don’t feel the need to actually change it, was just giving a little tip

jwnrt:
Yep, CI failures look unrelated. Might be a good idea to check our hardware, that jtag inject test has failed in a couple of PRs now.

Issue #22912, opened 2024-05-01
[documentation] Bootstrapping the demo software does not work for CW340


AndySecqai:
### Description

Following the [guide](https://opentitan.org/book/doc/getting_started/setup_fpga.html#bootstrapping-the-demo-software) for bootstrapping the demo software on the CW340 card does not work as the CW340 is not a valid build target for the `hello_world` software:

```
./bazelisk.sh build //sw/device/examples/hello_world:hello_world_fpga_${BOARD}_bin
ERROR: Skipping '//sw/device/examples/hello_world:hello_world_fpga_cw340_bin': no such target '//sw/device/examples/hello_world:hello_world_fpga_cw340_bin': target 'hello_world_fpga_cw340_bin' not declared in package 'sw/device/examples/hello_world' defined by /home/admin-sqai/opentitan/sw/device/examples/hello_world/BUILD (Tip: use `query "//sw/device/examples/hello_world:*"` to see all the targets in that package)
WARNING: Target pattern parsing failed.
ERROR: no such target '//sw/device/examples/hello_world:hello_world_fpga_cw340_bin': target 'hello_world_fpga_cw340_bin' not declared in package 'sw/device/examples/hello_world' defined by /home/admin-sqai/opentitan/sw/device/examples/hello_world/BUILD (Tip: use `query "//sw/device/examples/hello_world:*"` to see all the targets in that package)
INFO: Elapsed time: 0.282s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded)
```

PR #24519, opened 2024-09-05
[docs,bazel] Update FPGA bootstrapping docs to match current build flow

AlexJones0:
This PR addresses issue #22912.

The documentation has fallen out of date with regards to the FPGA bootstrapping process - the CW340 target is not defined for the `hello_world` demo like is suggested, and the bootstrapping commands have changed since the documentation was written. This PR adds the extra build target, updates the commands to those that can be used to build the `hello_world` binary and execute it on a given board, and corrects a few semantic errors and adds some additional information in different parts of the documentation.

As part of this PR, the `target-location.sh` script has also been modified so that it can handle multiple output files, as the current build flow for `opentitan_binary` targets now produces multiple outputs and as such, these must all be translated to absolute paths. I have run a `rgrep -uuu target-location.sh` to confirm that the places that this script was being used in are unaffected by this change. These are:
- In the `azure-pipelines.yml` to find the Test ROM for the cw305
- In `get-bitstream-strategy.sh` to get the location of the bitstream manifest file)

To test this PR, you can follow the commands described in the updated documentation to run the `hello_world` demo as described on either the CW310 or CW340.

### DIFF
--- a/ci/scripts/target-location.sh
+++ b/ci/scripts/target-location.sh
@@ -30,8 +30,8 @@ then
   REDIR='/dev/null'
 fi
 
-REL_PATH=$(${REPO_TOP}/bazelisk.sh outquery "$@" 2>$REDIR)
+REL_PATH=$(${REPO_TOP}/bazelisk.sh cquery --output=files "$@" 2>$REDIR)
 readonly REL_PATH
 REPO_EXECROOT=$(${REPO_TOP}/bazelisk.sh info --show_make_env execution_root)
 readonly REPO_EXECROOT
-echo "${REPO_EXECROOT}/${REL_PATH}"
+echo "${REL_PATH}" | awk -v root="$REPO_EXECROOT" '{print root "/" $0}'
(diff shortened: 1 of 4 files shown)


ci/scripts/target-location.sh
```diff
@@ -30,8 +30,8 @@ then
   REDIR='/dev/null'
 fi
 
-REL_PATH=$(${REPO_TOP}/bazelisk.sh outquery "$@" 2>$REDIR)
+REL_PATH=$(${REPO_TOP}/bazelisk.sh cquery --output=files "$@" 2>$REDIR)
 readonly REL_PATH
 REPO_EXECROOT=$(${REPO_TOP}/bazelisk.sh info --show_make_env execution_root)
 readonly REPO_EXECROOT
-echo "${REPO_EXECROOT}/${REL_PATH}"
+echo "${REL_PATH}" | awk -v root="$REPO_EXECROOT" '{print root "/" $0}'
```
jwnrt, line 37: Do you remember what this change was for? It looks equivalent to me

jwnrt: We discussed offline, `awk` adds the root per line where the original prefixed the whole thing

PR #24521, opened 2024-09-05
[python] move from python-requirements.in to pyproject.toml

nbdd0121:
No changes have been made to the dependency versions, except `types-dataclasses` is removed (since we use Python 3.8+)


jwnrt:
I think Tock has its own requirements file for tockloader. Please could you look at that too?

I think I had to generate its hashed lockfile with `uv` in the end because `pip-compile` was broken on Ubuntu 24.04.

nbdd0121:
The tock requirement file is just a single line mentioning tockloader. I don't know how much value would it provide to convert it to pyproject.toml?

jwnrt (approved):
This is great, thanks

pamaury (approved):
Looks good to me.

PR #24524, opened 2024-09-05
[i2c,dif] Test i2c dif SCL high and low time cycle bounds on clock stretching

AlexJones0:
The i2c block requires a minimum value (for the current setup, at least 4 cycles) for the SCL high and low time in order for clock stretching to function. This functionality was previously added to the DIF (see #23772), and so this PR updates the relevant unit tests to also check for these conditional cases. Modified input clock periods are used that have been calculated to execute minimum boundary branch cases, and a static assertion before the test ensures that these calculated values will not fall out of sync if the minimum high/low time does change.

```sh
./bazelisk.sh test -t- --test_output=streamed //sw/device/lib/dif:i2c_unittest
```
which still passes after the changes as expected.

### DIFF
--- a/sw/device/lib/dif/dif_i2c.c
+++ b/sw/device/lib/dif/dif_i2c.c
@@ -235,14 +235,12 @@ dif_result_t dif_i2c_compute_timing(dif_i2c_timing_config_t timing_config,
   }
 
   // For clock stretching detection to work, the SCL high and low time must be
-  // at leats 4 cycles.
-  // TODO The i2c IP has a parameter (InputDelayCycles) for that, use this when
-  // topgen supports exposing it in the headers (Issue#23786).
-  if (config->scl_time_high_cycles < 4) {
-    config->scl_time_high_cycles = 4;
+  // at least 4 cycles.
+  if (config->scl_time_high_cycles < kDifI2cInputDelayCycles) {
+    config->scl_time_high_cycles = kDifI2cInputDelayCycles;
   }
-  if (config->scl_time_low_cycles < 4) {
-    config->scl_time_low_cycles = 4;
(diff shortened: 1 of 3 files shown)


jwnrt (approved):
I’m not familiar with clock stretching but this looks sensible, thanks

jwnrt:
Actually quick question: is the four cycles a property of i2c, our IP implementation, our clock speed…?

Will this test be easy to update if one of these things changes, e.g. for multitop?

engdoreis (approved):

engdoreis:
Multitop is orthogonal to this change.

a-will:
> Actually quick question: is the four cycles a property of i2c, our IP implementation, our clock speed…?

Our implementation. The high and low time must be long enough for the IP to unambiguously detect when another driver is pulling low and to respond to it. This is dependent on the number of cycles of delay for the IP to sense its own outputs (assuming 0 rise time, which is a different parameter). The IP needs to know how many cycles to wait before it can know that a sensed low signal actually came from some other driver (for the purpose of this clock stretching case, when the SCL input must be low due to stretching, instead of from our own delayed output).

The value is a property of a specific top (chip level, really), since the propagation delay through pads is part of it.

> Will this test be easy to update if one of these things changes, e.g. for multitop?

See the TODO that got moved to the header. There is a top-level parameter to aid software, but topgen doesn't currently expose it. I forget all the contributors at the moment, but it should be a constant value plus `InputDelayCycles`.

Most likely, we won't begin to see differences for other tops here unless we add support for a specific ASIC that needs a different `InputDelayCycles`.

jwnrt:
Thanks @a-will, that’s clear

PR #24525, opened 2024-09-05
[sw,lib] Add unit test for debug printing four character codes

AlexJones0:
The `printf` unit testing was missing tests for the custom debug formatting specifier `kFourCC`, which is used for formatting Four Character Codes. This specifier prints a `uint32_t` as four ASCII characters encoded in little endian order. Existing print unit tests did not check this specifier, and so this commit introduces three additional small unit tests to check that the implementation is working as intended.

```sh
./bazelisk.sh test -t- --test_output=streamed //sw/device/lib/runtime:print_unittest
```
which still passes after the changes as expected.


AlexJones0:
CI failure appears to be unrelated.

nbdd0121 (approved):

jwnrt (approved):

PR #24527, opened 2024-09-05
[ci] Update quick-lint & CI scripts after site move

AlexJones0:
After the building of the site / landing page was moved to a separate repository, there is still an existing reference to the old `./ci/scripts/build-site.sh` script in `./ci/jobs/quick-lint.sh`, which causes the quick lint job to error near the end, This PR removes that reference from the job.

This commit also excludes the `build-site` directory from the CI executable file and shell check scripts, as current building of the docs via mdbook creates many executable files and shell scripts which do not pass linting standards (but are temporary, and local).

I've tested that this still works by simply running `./ci/jobs/quick_lint.sh`.

### DIFF
--- a/ci/jobs/quick-lint.sh
+++ b/ci/jobs/quick-lint.sh
@@ -61,8 +61,5 @@ ci/scripts/rust-format.sh $tgt_branch
 echo -e "\n### Run shellcheck on all shell scripts"
 util/sh/scripts/run-shellcheck.sh
 
-echo -e "\n### Render landing site"
-ci/scripts/build-site.sh
-
 echo -e "\n### Check what kinds of changes the PR contains"
 ci/scripts/get-build-type.sh $tgt_branch PullRequest
--- a/ci/scripts/exec-check.sh
+++ b/ci/scripts/exec-check.sh
@@ -19,6 +19,7 @@ args=(
     -name .git -prune -o
     -name vendor -prune -o
     -name scratch -prune -o
+    -name build-site -prune -o
     # Filter to executable files.
     -type f
     -executable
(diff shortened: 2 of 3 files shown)


jwnrt (approved):
Thanks Alex, why was this not failing if the script doesn’t exist?

AlexJones0:
> Thanks Alex, why was this not failing if the script doesn’t exist?

I'm not sure about you (it would be super helpful if you could test this really quick), but if I got to my local `master` branch and run 
```sh
./ci/jobs/quick-lint.sh
```

then it in fact does fail with the following contents
```
...
### Render landing site
./ci/jobs/quick-lint.sh: line 65: ci/scripts/build-site.sh: No such file or directory
```

Meaning that the `get-build-type.sh` script that comes after it is never being run locally.

jwnrt:
Oh I see, this script is not run by CI but is supposed to replicate it. I’m not sure I like that, maybe I’ll try clean that up so that there’s a single source.

AlexJones0:
In theory each azure job step is specified via bash, and so with adequate setup it shouldn't be so hard to make one of the two locations a source of truth, and then convert it into the other. The easier option would probably be extracting the bash from the `Quality (quick lint)` job and concatenating it into a `ci/jobs/quick-lint.sh` script labelled with the display names as comments. That being said, I'm not a huge fan of the idea of parsing the azure file and auto-generating a script, so I'm curious if you have any better approaches in mind?

jwnrt:
In this case I might prefer a single Bazel test target that runs all the scripts. Then it’s one command for Azure and CLI. These lints don’t need to run in series, so the Bash isn’t really necessary.

AlexJones0:
That makes sense; I do like the atomicity of the steps in our current Azure pipeline job, as I think it makes it much nicer to see what part of the lint exactly failed when looking at the logs. I'm curious how this would look instead if we used the Bazel approach.

jwnrt:
Yes, I know what you mean. Bazel’s output leaves a lot to be desired and in this case the Azure/GHA split does a better job. It’s a trade-off I guess.

engdoreis (approved):

PR #24528, opened 2024-09-05
[entropy_src,verilator] Reduce entropy tests requirements on Verilator to fix failing test targets

AlexJones0:
This PR partially addresses #24184 (specifically, it addresses the two tests found to fail if given long enough to run in Verilator).

Entropy is generated much more slowly in Verilator, and as a result, two tests were timing-out (and if given long enough, failing) when run in Verilator. These are `sw/device/tests:entropy_src_fw_observe_many_contiguous_test_sim_verilator`, and `sw/device/tests:entropy_src_fw_override_test_sim_verilator`. This PR conditionally lowers the test requirements for the Verilator simulation environment to allow these tests to pass on Verilator.

For `fw_observe_many_contiguous_test`, the number of entropy samples to be observed is reduced from 1024 to just 8, as the full 1024 would take around 13-15 hours. For `fw_override_test`, the entropy consumers have been limited to just random number generation, and the test is terminated after this on Verilator. The timeout given to observe this Entropy is also increased 30x on Verilator.

Note that these code changes do not modify the operation of any execution environments other than Verilator; only Verilator has had conditional changes. If this is deemed not appropriate due to requirements of these tests, then instead the Verilator environment should be removed from these two tests.

```sh
./bazelisk.sh test -t- --test_output=streamed --test_timeout=3600,3600,3600,3600 \
    //sw/device/tests:entropy_src_fw_observe_many_contiguous_test_sim_verilator 
    //sw/device/tests:entropy_src_fw_override_test_sim_verilator
```
which now passes as expected:
```sh
//sw/device/tests:entropy_src_fw_observe_many_contiguous_test_sim_verilator PASSED in 2807.4s
//sw/device/tests:entropy_src_fw_override_test_sim_verilator             PASSED in 3358.1s
```

### DIFF
--- a/sw/device/tests/entropy_src_fw_override_test.c
+++ b/sw/device/tests/entropy_src_fw_override_test.c
@@ -66,6 +66,30 @@ enum {
    * firmware override mode.
    */
   kEntropyFifoBufferSize = 12,
+  /**
+   * The amount of random numbers to generate as entropy consumers in each test.
+   */
+  kRandomNumberCount = 16,
+  /*
+   * Timeout to generate a random number in micro seconds.
+   */
+  kRandomNumberTimeoutUsec = 100 * 1000,
+  /**
+   * Timeout to generate a random number in micro seconds when running in a
+   * Verilator simulation environment. Verilator observes entropy more slowly
+   * than other environments, and so is given a longer timeout.
+   */
+  kVerilatorRandomNumberTimeoutUsec = 48 * 100 * 1000,
+  /**
+   * Timeout to run AES testutils in micro seconds.
+   */
+  kAesTestutilsTimeoutUsec = 1 * 1000 * 1000,
+  /**
+   * Timeout to run AES testutils in micro seconds when running in a
+   * Verilator simulation environment. Verilator observes entropy more slowly
+   * than other environments, and so is given a longer timeout.
+   */
+  kVerilatorAesTestutilsTimeoutUsec = 48 * 1000 * 1000,
 };
 
 static dif_aes_t aes;
@@ -360,10 +384,17 @@ status_t firmware_override_extract_insert(
 
   // Generate some random numbers.
   LOG_INFO("Generating random numbers...");
+  uint32_t rnd_timeout_usec = kRandomNumberTimeoutUsec;
+  if (kDeviceType == kDeviceSimVerilator) {
+    // Simulation on Verilator generates entropy much more slowly than other
+    // execution environments. As a result, the timeout to generate random
+    // numbers is increased only when running on Verilator.
+    rnd_timeout_usec = kVerilatorRandomNumberTimeoutUsec;
+  }
   uint32_t data;
-  for (size_t i = 0; i < 16; i++) {
+  for (size_t i = 0; i < kRandomNumberCount; i++) {
     TRY(rv_core_ibex_testutils_get_rnd_data(&rv_core_ibex,
-                                            /*timeout_usec=*/100 * 1000,
+                                            /*timeout_usec=*/rnd_timeout_usec,
                                             &data));
   }
 
@@ -386,9 +417,16 @@ status_t firmware_override_extract_insert(
   };
 
   LOG_INFO("Running AES...");
+  uint32_t aes_timeout_usec = kAesTestutilsTimeoutUsec;
+  if (kDeviceType == kDeviceSimVerilator) {
+    // Simulation on Verilator generates entropy much more slowly than other
+    // execution environments. As a result, the timeout for AES testuils is
+    // increased only when running on Verilator.
+    aes_timeout_usec = kVerilatorAesTestutilsTimeoutUsec;
+  }
   TRY(aes_testutils_setup_encryption(transaction, &aes));
   AES_TESTUTILS_WAIT_FOR_STATUS(&aes, kDifAesStatusOutputValid, true,
-                                /*timeout_usec=*/1 * 1000 * 1000);
+                                /*timeout_usec=*/aes_timeout_usec);
   TRY(aes_testutils_decrypt_ciphertext(transaction, &aes));
 
   // Run KMAC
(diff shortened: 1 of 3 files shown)


pamaury (commented):
I would say that the main purpose of this test is to run on the FPGA or a real device. Running this test in verilator only makes sense for debugging, in which case it does not necessarily matter if it takes many hours to run. 
Therefore it's not clear to me if this tests should be just marked as `manual` for verilator instead. I think @vogelpi should have a look at this since he designed the testplan for that.

jwnrt:
I think we should apply this change *and* mark manual. @pamaury is right that it’s not useful in Verilator, but if you really really have to use it to debug something then it should at least work.

AlexJones0:
> I think we should apply this change _and_ mark manual.

I agree, that sounds sensible. Regardless of whether this change is applied or not, it makes sense to mark the Verilator target as manual if the environment stays around because, based on what has been described so far, the Verilator test target only really makes sense for debugging purposes.

sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -66,6 +66,24 @@ enum {
...
+  /*
+   * Timeout to generate a random number in micro seconds.
+   */
+  kRandomNumberTimeoutUsec = 100 * 1000,
+  /**
+   * Timeout to generate a random number in micro seconds when running in a
+   * Verilator simulation environment. Verilator observes entropy more slowly
+   * than other environments, and so is given a longer timeout.
+   */
+  kVerilatorRandomNumberTimeoutUsec = 3 * 1000 * 1000,
```
vogelpi, line 82: There is a factor of 30x difference here and this strongly suggests that we probably never cared about tuning the rate of the raw noise source (inside AST) for the Verilator simulation. For context, to speed up the Verilator simulation, we use different clock frequences and ratios as well as baud rates for Verilator. See e.g. sw/device/lib/arch/device_sim_verilator.c.

If you compare these values with the values for FPGA (for which this test was initially designed), you'll note an factor of 48x or similar difference. The RTL model used on FPGA and in Verilator is the same though. I don't remember exactly at which clock the AST model is running. I think what we should really do is tuning the AST model for the Verilator simulation. Would you mind creating a GitHub issue for this please?

sw/device/tests/entropy_src_fw_override_test.c
```diff
@@ -360,13 +378,29 @@ status_t firmware_override_extract_insert(
...
   }
 
+  if (kDeviceType == kDeviceSimVerilator) {
+    // If running on Verilator then entropy is generated much more slowly.
+    // It would take an impractical amount of time to generate the thousands
+    // of words of entropy that are required by all entropy consumers, and so
+    // since we are not concerned with the rate of entropy on Verilator, we
+    // stop at just generating some random numbers only.
+    return OK_STATUS();
+  }
```
vogelpi, line 402: I think this is non-ideal because it means we skip quite a big portion of the main test function in Verilator. So the Verilator simulation becomes less useful for debugging as a big part of the design is not exercised anymore.

Another way to speed up the Verilator simulation would be to reduce the number of modes tested. E.g. for Verilator only we could change:
```c
  // Test all modes.
  static dif_entropy_src_single_bit_mode_t kModes[] = {
      kDifEntropySrcSingleBitModeDisabled, kDifEntropySrcSingleBitMode0,
      kDifEntropySrcSingleBitMode1,        kDifEntropySrcSingleBitMode2,
      kDifEntropySrcSingleBitMode3,
  };
```
to
```c
  // Test all modes.
  static dif_entropy_src_single_bit_mode_t kModes[] = {
      kDifEntropySrcSingleBitModeDisabled, kDifEntropySrcSingleBitMode0,
  };
```
This would just check 2 instead of 5 modes (..Mode0 / .. ..Mode3 are almost identical and all of them are 4x slower compared to ...ModeDisabled).

Also, the final phase:
```c
  // Rerun the test with single bit mode disabled,
  // this time with an output delay.
  fw_ov_insert_wait_enabled = true;
  EXECUTE_TEST(test_result, firmware_override_extract_insert,
               kDifEntropySrcSingleBitModeDisabled, false, false);
  EXECUTE_TEST(test_result, firmware_override_extract_insert,
               kDifEntropySrcSingleBitModeDisabled, true, false);
```
Could be skipped for Verilator. We don't learn a lot from this.

vogelpi (commented):
Thanks for looping me in @AlexJones0 and @pamaury !

I've left some comments on how to change the fw_override_test in a way to not reduce the scope of the test too much but still speed it up for Verilator. Would you mind giving this a try please?

You are right: this test is mainly useful for SiVal purposes (non-Simulation platforms) but having this running in Verilator is very useful for debugging. So re-enabling this is valuable and the approach taken here is valid for the Verilator simlation environment. Thanks @AlexJones0 for doing this work!

sw/device/tests/entropy_src_fw_override_test.c
AlexJones0: That makes sense, thanks - I've made that issue here: #24585

Please do mention if there's anything that I've missed capturing in that issue. I had a quick look at the RTL of the ADC as well as the device clock/ratio definitions for the Verilator and FPGA environments and I suspect that as you say it is this 48x difference in clock speeds that necessitates this increase in timeout.

sw/device/tests/entropy_src_fw_override_test.c
AlexJones0: Yes, I agree, this was one thing that I was not sure on.

Making your suggested changes (just testing the first two modes and skipping the final phase on Verilator), alongside the same timeout changes as before, causes the test to pass in 6656.2 seconds (around 1 hour 50 minutes) in an environment similar to CI. It is not clear to me whether properly tuning the AST model as suggested would change this.

This seems fine to me for debug purposes - it is far better than the 15+ hours it was taking before. Ideally we would like it to run in under an hour, but given the purpose of this test I think it makes more sense to tag the Verilator environment as manual for `entropy_src_fw_override_test` with a comment explaining the runtime, alongside these changes. Does this sound reasonable to you @vogelpi?

sw/device/tests/entropy_src_fw_override_test.c
vogelpi: Thanks for creating the issue!

sw/device/tests/entropy_src_fw_override_test.c
vogelpi: Thanks for trying this out and reporting back. Tuning the AST model would help on top of this but we don't have to do this right now.

> This seems fine to me for debug purposes - it is far better than the 15+ hours it was taking before. Ideally we would like it to run in under an hour, but given the purpose of this test I think it makes more sense to tag the Verilator environment as manual for entropy_src_fw_override_test with a comment explaining the runtime, alongside these changes. Does this sound reasonable to you @vogelpi?

This would be perfect. Can you implement these changes directly in this PR?

vogelpi (approved):
LGTM, thanks @AlexJones0 !

AlexJones0:
I forgot to mark the Verilator environment for the `fw_override` test as manual in that push so I've just added that now - apologies!

AlexJones0:
Force push rebases on `main` to fix merge conflicts.

vogelpi:
Thanks @AlexJones0 ! CI passed except for the usbdev_iso FPGA test but this failure is unrelated to this PR and I 've seen it before. I am thus merging this PR now.

AlexJones0:
Should be able to automatically cherry pick after https://github.com/lowRISC/opentitan/pull/25138 is merged - I'll wait for that to be backported first to resolve any conflicts (and thus the need for manual backporting).

PR #24529, opened 2024-09-05
[rom_ext] Remove the SPI bootstrap code

cfrantz:
Initially we were considering the SPI bootstrap mechanism as the ROM_EXT rescue mechanism, however, we ended up going a different direction.

1. Remove the ROM_EXT SPI bootstrap code.


moidx (approved):

moidx:
For the future reader, this is the Rescue protocol that has been specified and verified: https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/rom_ext/doc/rescue.md

PR #24536, opened 2024-09-09
[opentitantool] improve user experience

vbendeb:
When 'opentitiantool image manifest update ...' fails due to any input or output file name mistyped, or any of the input files does not exist, the error message is the same:

  Error: No such file or directory (os error 2)

Which does not help much the operator. This patch adds some code which will print the name of the file causing the error.

Tested by invoking 'image manifest update' with incorrectly specified various input files. In case of the wrong SPX key file name the file name is printed twice, since the code is trying to read the file first trying to retrieve a private key, and then if unsuccessful - one more time trying to retrieve the public key.

This double print could be eliminated by proper return error processing, but is left as is at this time.

### DIFF
--- a/sw/host/opentitanlib/src/image/manifest_def.rs
+++ b/sw/host/opentitanlib/src/image/manifest_def.rs
@@ -8,7 +8,7 @@ use crate::util::bigint::fixed_size_bigint;
 use crate::util::num_de::HexEncoded;
 use crate::util::parse_int::ParseInt;
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use serde::{Deserialize, Serialize};
 use std::convert::{TryFrom, TryInto};
 use std::fmt;
@@ -96,7 +96,9 @@ macro_rules! manifest_def {
 
 impl ManifestSpec {
     pub fn read_from_file(path: &Path) -> Result<ManifestSpec> {
-        Ok(deser_hjson::from_str(&std::fs::read_to_string(path)?)?)
+        Ok(deser_hjson::from_str(
+            &std::fs::read_to_string(path).with_context(|| format!("Failed to open {path:?}"))?,
+        )?)
     }
 
     pub fn overwrite_fields(&mut self, other: ManifestSpec) {
--- a/sw/host/opentitantool/src/command/image.rs
+++ b/sw/host/opentitantool/src/command/image.rs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::{ensure, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use clap::{Args, Subcommand};
 use serde_annotate::Annotate;
 use std::any::Any;
@@ -26,7 +26,7 @@ use opentitanlib::image::manifest_def::ManifestSpec;
 use opentitanlib::image::manifest_ext::{ManifestExtEntry, ManifestExtId, ManifestExtSpec};
 use opentitanlib::util::file::{FromReader, ToWriter};
 use opentitanlib::util::parse_int::ParseInt;
-use sphincsplus::{DecodeKey, SpxDomain, SpxPublicKey, SpxSecretKey};
+use sphincsplus::{DecodeKey, SpxDomain, SpxError, SpxPublicKey, SpxSecretKey};
 
 /// Bootstrap the target device.
 #[derive(Debug, Args)]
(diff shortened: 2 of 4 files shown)


vbendeb:
A mostly cosmetic change making it easier to figure out the reason for opentitnatool invocation failures.

sw/host/opentitanlib/src/image/manifest_def.rs
```diff
@@ -96,7 +96,12 @@ macro_rules! manifest_def {
 
 impl ManifestSpec {
     pub fn read_from_file(path: &Path) -> Result<ManifestSpec> {
-        Ok(deser_hjson::from_str(&std::fs::read_to_string(path)?)?)
+        Ok(deser_hjson::from_str(
```
cfrantz, line 99: This kind of error information is better communicated by using the [`anyhow::Context`](https://docs.rs/anyhow/1.0.87/anyhow/trait.Context.html) trait:
```
use anyhow::Context;
...

  std::fs::read_to_string(path).with_context(|| format!("Failed to open {path:?}"))?
```

In the error case, this will cause the error message printed by `main` to look something like:

```
Error:
    Failed to open 'foo.json'
Caused by:
    File not found
```

vbendeb: done

sw/host/opentitantool/src/command/image.rs
```diff
@@ -233,11 +233,19 @@ impl CommandDispatch for ManifestUpdateCommand {
...
-                (SpxPublicKey::read_pem_file(key)?, None)
+            let (pk, sk) = match SpxSecretKey::read_pem_file(key) {
+                Ok(sk) => (SpxPublicKey::from(&sk), Some(sk)),
+                Err(SpxError::Io(_)) => {
+                    // Return error if file could not be read.
+                    bail!(std::io::Error::new(
+                        std::io::ErrorKind::NotFound,
+                        "File not found"
+                    ));
+                }
```
vbendeb, line 244: not sure if this is the most elegant way to do it, but this eliminates printing the same error twice when the key file does not exist

vbendeb:
I did not add `anyhow` to the sphincs library, seems more trouble than it'w worth

sw/host/opentitantool/src/command/image.rs
cfrantz: We could choose to not try to print an error in the sphincsplus lib and rely on the higher level code printing the error.  To pick this call-site as an example, if we were to write something like this:
```
    let (pk, sk) = if let Ok(sk) = SpxSecretKey::read_pem_file(key) {
        (SpxPublicKey::from(&sk), Some(sk))
    } else {
        (SpxPublicKey::read_pem_file(key).with_context(|| format!("Failed to open {key:?}"))?, None)
}
```

Then the "File not found" error is going to display as:
```
Error:
  Failed to open 'foo.pem'
Caused by:
  File not found
```

Whereas some other error (like a PEM parsing error) will display as:
```
Error:
  Failed to open 'foo.pem'
Caused by:
  PEM parse error
```

Another possibility is to consider that `anyhow::Context` is implemented on the core rust error type.  I think you can use `Context` without using the rest of `anyhow`.  In that case, you could (in `sphincspluis/lib.rs`) write something like:

```
use anyhow::Context;
...
let s = std::fs::read_to_string(filename)
    .map_err(SpxError::Io)
    .with_context(format!("Failed to open {filename:?}"))?;
```

(I'm not exactly sure how the Context gets attached to the error; I haven't studied that portion of the `anyhow` crate in detail).

vbendeb: I thought about removing error reporting from the SPHINCS library, but decided against it, because it could be invoked from multiple places, so each place would have to add error processing.

should this not be a concern?

cfrantz: The `eprintln!` is unconditional - I much prefer error reporting through the Error trait rather than extraneous prints buried in a library (IMHO, that style of error reporting or runtime state should be handled by a logging library).

I think in this case, the two best possibilities are:
1. Add the context object at the library call-sites.
2. Add the context object in the sphincsplus library.

In both cases, the helpful message describing which file had trouble gets presented to the user when the program terminates and prints its final result.

vbendeb: I am not really sure how to `Add the context object in the sphincsplus library`, may I suggest to leave this as a future exercise.

I improve the error reporting in `command/image.rs` so at least on manifest update invocations file name will be reported. 

Also added the fix to the bootstrap option - another place where I run into this error quite often.

PTAL

cfrantz: sgtm

cfrantz (approved):

PR #24537, opened 2024-09-09
[opentitantool] do not create duplicate extension entries

vbendeb:
It was observed that repetitive invoking 'opentitiantool image manifest update ...' passing an --spx_key and same manifest input file which included SPX key and signature extensions repetitively is causing addition of the SPX key and SPX signature space every time opentitantool is invoked.

This patch fixes the problem.

All opentitantool and opentitanlib tests pass, and running the same invocation does not cause additions of the SPX key and signature spaces any more.

### DIFF
--- a/sw/host/opentitanlib/src/image/manifest_def.rs
+++ b/sw/host/opentitanlib/src/image/manifest_def.rs
@@ -177,8 +177,23 @@ impl ManifestPacked<[ManifestExtTableEntry; CHIP_MANIFEST_EXT_TABLE_COUNT]>
 
     fn overwrite(&mut self, o: Self) {
         for i in 0..self.len() {
-            if !matches!(o[i].0, ManifestExtEntryVar::None) {
-                self[i].0 = o[i].0.clone();
+            match o[i].0 {
+                ManifestExtEntryVar::Name(other_id) => match self[i].0 {
+                    ManifestExtEntryVar::IdOffset {
+                        identifier: self_id,
+                        offset: _,
+                    } => {
+                        if self_id == other_id {
+                            // Do not overwrite existing entries with matching IDs.
+                            continue;
+                        } else {
+                            self[i].0 = o[i].0.clone()
+                        }
+                    }
+                    _ => self[i].0 = o[i].0.clone(),
+                },
+                ManifestExtEntryVar::None => (),
+                _ => self[i].0 = o[i].0.clone(),
             }
         }
     }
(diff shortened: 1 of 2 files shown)


vbendeb:
fixes the previously discussed issue of uncontrollable image growth when passing the same manifest file to `opentiatintool image manifest update ..`

vbendeb:
@pamaury @cfrantz any thought on this patch? 

I think the fact that the image grows with each successive invocation when manifest json file is supplied is a problem which needs to be fixed.

pamaury:
I need to look at the code a bit more, I am not super familiar with the image manifest. Probably @cfrantz knows a lot more than me. My most immediate question is: what is `opentitan manifest update` supposed to do with key? Is it adding a new one or updating one? Do we even support manifests with more than one RSA/ECDSA/SPX keys?

vbendeb:
> I need to look at the code a bit more, I am not super familiar with the image manifest. Probably @cfrantz knows a lot more than me. My most immediate question is: what is `opentitan manifest update` supposed to do with key? Is it adding a new one or updating one? Do we even support manifests with more than one RSA/ECDSA/SPX keys?

Yes, we chatted about this with @cfrantz, what happens now is that once opentitantool was invoked with a manifest including SPX key and signature extensions, and the SPX key, the tool allocates room for the signature to adjust the total image size, filling up the signature space with 0xff. But if one invokes opentitantool again with the same manifest, another space for signature is added, and this repeats indefinitely, the size keeps increased by the SPX signature size on every invocation.

sw/host/opentitanlib/src/image/manifest_def.rs
```diff
@@ -177,9 +177,20 @@ impl ManifestPacked<[ManifestExtTableEntry; CHIP_MANIFEST_EXT_TABLE_COUNT]>
 
     fn overwrite(&mut self, o: Self) {
         for i in 0..self.len() {
-            if !matches!(o[i].0, ManifestExtEntryVar::None) {
-                self[i].0 = o[i].0.clone();
-            }
+            match self[i].0 {
```
cfrantz, line 180: The logic seems reversed here: in the original, the `other` entry is checked that it is not `None` and then overwrites the corresponding entry in `self`.  In the new version, the value of `self` is checked to see if it should be overwritten with `other`.

It seems like `other` should be checked to see if there is something there to overwrite and then the current values in `self` and `other` should be examined to determine whether or not to perform the overwrite.

vbendeb: good point, fixed. Does not look very elegant, but I can't think of a better way.

vbendeb (commented):
PTAL

cfrantz (approved):

PR #24538, opened 2024-09-09
[dv, spi] Add smaller delay for spinwait on busy bit

sha-ron:
Sometimes SPI busy bit pulse might be smaller than the default spinwait interval so it can be missed.

### DIFF
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_sw_spi_passthrough_collision_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_sw_spi_passthrough_collision_vseq.sv
@@ -120,7 +120,7 @@ class chip_sw_spi_passthrough_collision_vseq extends chip_sw_spi_passthrough_vse
         `uvm_send(m_spi_host_seq);
         if (opcode == SpiFlashReadSts1 && m_spi_host_seq.rsp.payload_q[0][0] != 0) begin
           // Wait until software clears the WIP and WEL bits.
-          spi_host_wait_on_busy();
+          spi_host_wait_on_busy(.min_interval_ns(10));
         end
         host_rsp_q.push_back(m_spi_host_seq.rsp);
 


hcallahan-lowrisc:
@rswarbrick and I had a quick look at this, and don't quite follow the reasoning for the change.

https://github.com/lowRISC/opentitan/blob/9af9cadd1e40013845c232eee762e1933d9a9289/hw/top_earlgrey/dv/env/seq_lib/chip_sw_spi_passthrough_collision_vseq.sv#L121-L124
The test goes into this routine only if the busy bit is active (`m_spi_host_seq.rsp.payload_q[0][0] != 0`) when read by the previous upstream spi_agent 'SpiFlashReadSts1' transfer.

The routine `spi_host_wait_on_busy()` then spawns more 'SpiFlashReadSts1' transfers (with the configurable delay/timeout) from the upstream spi_agent that check the value of the busy bit, and returns as soon as the busy bit is read back as 0.

Even if the busy bit was only pulsed high for a very short time the first time it was read, by the time it is read again by the wait_on_busy() routine it will be read as zero. And then will remain at zero, as there is only the single thread of execution working through all the opcodes in the test.
Shortening the interval between reads then doesn't make sense to me in this context, as I don't think there is anything to 'miss'?

sha-ron:
Thank you @hcallahan-lowrisc for the review. Maybe I don't understand spi_host_wait_on_busy() and the polarity of m_spi_host_seq.rsp.payload_q[0][0].  m_spi_host_seq.rsp.payload_q[0][0] != 0 means SPI bus is busy, meaning there is a transaction on SPI bus? If yes, why spi_host_wait_on_busy() sends another transaction during this time?

hcallahan-lowrisc:
Hi @sha-ron, thanks for your message. The busy bit (read back as `m_spi_host_seq.rsp.payload_q[0][0]`) is a typical spi-flash feature which means that a (slow) internal write operation is in progress. The user of the flash (the upstream spi_host agent in this test) can then poll this bit to determine the current status of the flash, awaiting it to become ready for the next operation. 

The `spi_host_wait_on_busy()` is doing this polling operation. Reading back the bit as 1 means that the block is busy, and it needs to poll again to find out when this bit becomes 0. So it's not that the SPI bus is busy, but that write-opcodes in this test cause the OpenTitan SPI device block to automatically sets the busy bit, which remains set until OT software processes the uploaded command and then clears the busy bit.

sha-ron:
Closing this PR since it doesn't solve the issue.

PR #24543, opened 2024-09-10
[dv,alerts] Fix alert_sva_active to have correct reset polarity

eshapira:
This PR fix the reset polarization of the alert_sva_active flag in the alert_esc_if which was inverted.
The alerts SVA should be active only when the reset is deasserted, which means when rst_n = 1.
This means alert_sva_active should be with the same polarization as rst_n, hence the fix.


rswarbrick (approved):
Looks sensible to me!

eshapira:
Thanks!

sha-ron:
@andreaskurth can you please merge it also to earlgrey_1.0.0 branch?

PR #24545, opened 2024-09-11
[pentest] Add CSRNG Bias FI tests

nasahlpa:
This commit adds the following two new FI tests:
- CsrngBiasFWOverride
- CsrngBiasFWOverrideStatic

These tests can be used to target the entropy source and they allow to monitor the data over the CSRNG
interface.


vogelpi (approved):

For the record: This is not production code but code that was used for penetration testing of Earlgrey ES. We should not implement changes to these tests that might have an impact on the test results. Instead the aim is to integrate the code more or less as is to be able to reproduce results and to have it ready for future penetration testing campaigns. The first PR of this series (https://github.com/lowRISC/opentitan/pull/24294) has been reviewed by the SW team. The suggested style changes have also been integrated into this PR.

PR #24548, opened 2024-09-11
[third_party] Add open-dice dependency

tommychiu-github:
We'll need the cbor_reader and cbor_writer implementations from the open-dice repository, for the CWT CBOR DICE implementation in the ROM_EXT.

Test: bazel build @@open-dice//:cbor_reader_writer

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -149,6 +149,10 @@ sphincsplus_repos()
 load("//rules:bitstreams.bzl", "bitstreams_repo")
 bitstreams_repo(name = "bitstreams")
 
+# CBOR reader & writer from open-dice
+load("//third_party/open-dice:repos.bzl", "open_dice_repos")
+open_dice_repos()
+
 # Setup for linking in externally managed test and provisioning customizations
 # for both secure/non-secure manufacturer domains.
 load("//rules:hooks_setup.bzl", "hooks_setup", "provisioning_exts_setup", "secure_hooks_setup")
--- a/third_party/open-dice/repos.bzl
+++ b/third_party/open-dice/repos.bzl
@@ -1,0 +1,15 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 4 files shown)


third_party/open-dice/repos.bzl
```diff
@@ -0,0 +1,15 @@
...
+load("//rules:repo.bzl", "http_archive_or_local")
+
+def open_dice_repos(local = None):
+    http_archive_or_local(
+        name = "open-dice",
+        local = local,
+        build_file = Label("//third_party/open-dice:BUILD.open-dice.bazel"),
+        strip_prefix = "open-dice-main",
+        urls = ["https://github.com/google/open-dice/archive/refs/heads/main.zip"],
+        sha256 = "a802fa53c3147a1ee9a06c9aaeee6d2b49f20ca6d32877f1ccfa4f9487df7f5f",
```
cfrantz, line 14: Please pin this to an exact commit.  Referring to `refs/heads/main` means the archive will change every time there is a new commit to the upstream codebase and the `sha256` field will become invalid.

Use something like: `https://github.com/google/open-dice/archive/<git-hash>.tar.gz` as the URL reference.

WORKSPACE
```diff
@@ -30,6 +30,10 @@ google_repos()
 load("//third_party/google:deps.bzl", "google_deps")
 google_deps()
 
+# cboer reader & writer from open-dice
+load("//third_party/open-dice:repos.bzl", "open_dice_repos")
+open_dice_repos()
+
```
cfrantz, line 36: There is no specific problem with loading this here, but the early ordering of some deps can require some care.  As a general utility library, I'd suggest loading this later (certainly after the compilers and other language support).  I'd suggest somewhere after line 116 (FreeRTOS, etc).

WORKSPACE
```diff
@@ -149,6 +149,10 @@ sphincsplus_repos()
 load("//rules:bitstreams.bzl", "bitstreams_repo")
 bitstreams_repo(name = "bitstreams")
 
+# cboer reader & writer from open-dice
```
timothytrippel, line 152: nit:
```suggestion
# CBOR reader & writer from open-dice
```

timothytrippel (approved):

PR #24562, opened 2024-09-12
[dv,vcs] Tidy up how coverage is collected in assertion modules

rswarbrick:
- Stop collecting non-assertion coverage from the tlul_assert block (since that seems a bit pointless).

- Mirror this change in cover_reg_top.cfg

- Refactor stuff in the rv_dm specialisations of the two files because they are supposed to *extend* the base versions rather than reimplementing them.

- The resulting rv_dm version of cover.cfg ends up empty (which causes VCS to segfault!). Remove it.


eshapira (approved):
Sound good.
Thanks!

PR #24565, opened 2024-09-12
[spi_host, verible-lint] Adding spi_host to verible lint config file

antmarzam:


rswarbrick (approved):
Looks good to me. Merging since there's no way this could have broken FPGA CI tests.

PR #24566, opened 2024-09-12
[rv_dm,dv] Add rv_dm_scanmode_vseq to stress all list

rswarbrick:
This just means that we enable/disable scanmode in the middle of a test, which means a continuous assignment from the signal changes (fixing a coverage hole).


alees24 (approved):
Makes sense.

PR #24567, opened 2024-09-12
[rv_dm,dv] Manually waive line coverage for unreachable code

rswarbrick:
This didn't get picked up by UNR for some reason but definitely cannot run: it is the code that gets used for certain parameter values that aren't the ones in use.


eshapira (approved):

rswarbrick:
Bizarrely, I've just debugged something else and this doesn't actually work. Dropping the annotations makes it work though. (Huh?!)

I'm force pushing without the two added annotations and will merge assuming CI is happy with it.

PR #24568, opened 2024-09-12
[rv_dm,dv] Waive some impossible secded toggle coverage

rswarbrick:
Detailed notes about the waiver in a comment in the code.


alees24 (approved):
Thank you. It looks as though this will need replicating across the other IP blocks.

PR #24570, opened 2024-09-12
Cherry-pick from `master`: hsmtool fixes

cfrantz:
OpenTitan's ECDSA implementation expects the R and S components of an
ECDSA signature to be in little-endian order.

1. Reverse the R and S components individually rather than reversing the entire signature.
2. Following the PKCS#11 specifications, the CKA_EC_POINT should be the DER-encoding of ANSI X9.62 ECPoint value.

These cherry-picks are from #23962 and #23905.


moidx (approved):
LGTM

PR #24571, opened 2024-09-13
[util] Remove expand_seed routine.

jadephilipoom:
This routine wasn't currently affecting the logic, but would have weakened certain seeds that happened to have low values. Ideally this should all be refactored to use byte strings so it's clearer what range the value is sampled from, but for now just removing the harmful logic should suffice.

### DIFF
--- a/util/topgen/secure_prng.py
+++ b/util/topgen/secure_prng.py
@@ -173,23 +173,25 @@ class secure_prng():
         from previous instantiation.
         """
         if seed == 0:
-            entropy_input = seed.to_bytes(32, 'big')
             log.error("ERROR: PRNG seeded with 0.")
             sys.exit(1)
-        # Check if the provided seed is at least 32 bytes long.
-        elif sys.getsizeof(seed) < sys.getsizeof(1 << 255):
-            # Error out if seed is shorter than 256 bits.
-            log.error("ERROR: Seed shorter than 256 bits.")
-            sys.exit(1)
+
+        if seed.bit_length() < 250:
+            # Warn, but don't fail, because the DV logic always passes in 32-bit
+            # seeds and this can naturally happen about 1% of the time.
+            log.warn('PRNG seed is only {seed.bit_length()} bits long, which is '
+                     'unlikely for a sample from a 256-bit distribution. Please '
+                     'double-check the logic.')
         # Check if the seed is longer than 256 bits. Trim the excess bits and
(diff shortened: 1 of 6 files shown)


util/topgen/secure_prng.py
```diff
@@ -173,23 +173,25 @@ def reseed(self, seed):
...
-            entropy_input = seed.to_bytes(32, 'big')
             log.error("ERROR: PRNG seeded with 0.")
             sys.exit(1)
-        # Check if the provided seed is at least 32 bytes long.
-        elif sys.getsizeof(seed) < sys.getsizeof(1 << 255):
-            # Error out if seed is shorter than 256 bits.
-            log.error("ERROR: Seed shorter than 256 bits.")
-            sys.exit(1)
+
+        if seed.bit_length() < 250:
```
moidx, line 179: If this is to fail %1 of the time whenever we change seeds, how do you feel about reporting an error instead? I think the Makefile output is too verbose, so there is a good chance this warning will be missed.

moidx (approved):

util/topgen/secure_prng.py
jadephilipoom: There's two issues:
- Eliminating seeds below a certain lower bound would reduce the input distribution, so the actual seed would not be able to be fully randomly sampled. The loss in security bound is only marginal, but it's still not ideal; we want to sample from the full distribution of 0..2^256-1.
- The DV setup always passes a 32-bit seed. It uses the command-line override, though, so we could error only if the value is from HJSON.

The fix for this that I prefer long-term is to:
- make the values in the HJSON files byte-strings, so their length is easy to check and can include leading 0s, and error if the length is not exactly 32 bytes
- add a command-line argument for "allow small seeds" so the DV setup can override the error

However, when I was trying to make bigger fixes to the PRNG and surrounding logic the change quickly became pretty big, so here I wanted to start with as narrow a hotfix as possible to remove the biggest issue.

moidx: Hi @jadephilipoom, this explanation makes sense and I agree with your approach in this PR. Thanks for the additional details.

moidx:
LGTM

jadephilipoom:
I confirmed that the single test that is failing is also failing on `master` with the same error (consistently and locally as well as on CI). Since it's unrelated, I'll merge this now.

PR #24572, opened 2024-09-13
[sival/aes] Add AES Interrupt Encryption tests

rprakas-gsc:
@moidx  Checking in aes_interrupt_encryption_test.c code and modified the BUILD to include this test.

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -119,6 +119,31 @@ opentitan_test(
     ],
 )
 
+opentitan_test(
+    name = "aes_interrupt_encryption_test",
+    srcs = ["aes_interrupt_encryption_test.c"],
+    exec_env = dicts.add(
+        {
+            "//hw/top_earlgrey:sim_verilator": None,
+            "//hw/top_earlgrey:fpga_cw340_sival": None,
+        },
+    ),
+    verilator = verilator_params(
+        timeout = "long",
+    ),
+    deps = [
+        "//hw/ip/aes:model",
+        "//hw/top_earlgrey/sw/autogen:top_earlgrey",
+        "//sw/device/lib/base:memory",
+        "//sw/device/lib/base:mmio",
+        "//sw/device/lib/dif:aes",
+        "//sw/device/lib/dif:edn",
+        "//sw/device/lib/runtime:log",
+        "//sw/device/lib/testing:aes_testutils",
+        "//sw/device/lib/testing/test_framework:ottf_main",
+    ],
+)
+
 alias(
     name = "aes_smoketest_entropy_testutils",
     actual = select({
--- a/sw/device/tests/aes_interrupt_encryption_test.c
+++ b/sw/device/tests/aes_interrupt_encryption_test.c
@@ -1,0 +1,261 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "hw/ip/aes/model/aes_modes.h"
+#include "sw/device/lib/base/memory.h"
+#include "sw/device/lib/base/mmio.h"
+#include "sw/device/lib/base/multibits.h"
+#include "sw/device/lib/dif/dif_aes.h"
+#include "sw/device/lib/dif/dif_csrng.h"
+#include "sw/device/lib/dif/dif_csrng_shared.h"
+#include "sw/device/lib/dif/dif_edn.h"
+#include "sw/device/lib/runtime/log.h"
+#include "sw/device/lib/testing/aes_testutils.h"
+#include "sw/device/lib/testing/csrng_testutils.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+enum {
+  kTestTimeout = (1000 * 1000),
+};
+
+// The mask share, used to mask kKey. Note that the masking should not be done
+// manually. Software is expected to get the key in two shares right from the
+// beginning.
+
+static const unsigned char kAesModesPlainTextBlock12[32] = {
+    0x6b, 0xc1, 0xbe, 0xe2, 0x2e, 0x40, 0x9f, 0x96, 0xe9, 0x3d, 0x7e,
+    0x11, 0x73, 0x93, 0x17, 0x2a, 0xae, 0x2d, 0x8a, 0x57, 0x1e, 0x03,
+    0xac, 0x9c, 0x9e, 0xb7, 0x6f, 0xac, 0x45, 0xaf, 0x8e, 0x51,
+};
+
+static const unsigned char kAesModesPlainTextBlock34[32] = {
+    0x30, 0xc8, 0x1c, 0x46, 0xa3, 0x5c, 0xe4, 0x11, 0xe5, 0xfb, 0xc1,
+    0x19, 0x1a, 0x0a, 0x52, 0xef, 0xf6, 0x9f, 0x24, 0x45, 0xdf, 0x4f,
+    0x9b, 0x17, 0xad, 0x2b, 0x41, 0x7b, 0xe6, 0x6c, 0x37, 0x10};
+
+static const uint8_t kKeyShare1[] = {
+    0x0f, 0x1f, 0x2f, 0x3f, 0x4f, 0x5f, 0x6f, 0x7f, 0x8f, 0x9f, 0xaf,
+    0xbf, 0xcf, 0xdf, 0xef, 0xff, 0x0a, 0x1a, 0x2a, 0x3a, 0x4a, 0x5a,
+    0x6a, 0x7a, 0x8a, 0x9a, 0xaa, 0xba, 0xca, 0xda, 0xea, 0xfa,
+};
+
+OTTF_DEFINE_TEST_CONFIG();
+
+status_t execute_test(void) {
+  dif_aes_mode_t aes_mode;
+  dif_aes_iv_t iv_mode, iv;
+
+  for (int test_modes = 1; test_modes < 5; test_modes++) {
+    switch (test_modes) {
+      case 1:
+        LOG_INFO("Mode: CBC");
+        aes_mode = kDifAesModeCbc;
+        memcpy(iv_mode.iv, kAesModesIvCbc, sizeof(kAesModesIvCbc));
+        break;
+      case 2:
+        LOG_INFO("Mode: CFB-128");
+        aes_mode = kDifAesModeCfb;
+        memcpy(iv_mode.iv, kAesModesIvCfb, sizeof(kAesModesIvCfb));
+        break;
+      case 3:
+        LOG_INFO("Mode: OFB");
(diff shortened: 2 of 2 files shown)


sw/device/tests/aes_interrupt_encryption_test.c
```diff
@@ -0,0 +1,260 @@
...
+                                    (size_t)kAesNumBlocks_t2));
+
+  CHECK_DIF_OK(dif_aes_load_data(&aes2, plain_text2[0]));
+  LOG_INFO("Started AES trial Encrypt");
+  CHECK_DIF_OK(dif_aes_end(&aes2));
+
+  /** Not checking successful encryption here since its a trial transaction
+   *  Trial encryption ends here
+   *  Resume the Original Encryption below
+   */
```
luismarques, line 182: Nit: why is this using documentation-style comments?

sw/device/tests/BUILD
```diff
@@ -6279,3 +6279,24 @@ opentitan_test(
...
     ],
 )
+
+opentitan_test(
+    name = "aes_interrupt_encryption_test",
+    srcs = ["aes_interrupt_encryption_test.c"],
+    exec_env = {
+        #         "//hw/top_earlgrey:sim_verilator": None,
+        "//hw/top_earlgrey:fpga_cw340_sival": None,
+        #         "//hw/top_earlgrey:fpga_cw340_rom_with_fake_keys": None,
```
luismarques, line 6289: Please clean up the unsupported execution environments. Though perhaps you'd be able to add back verilator, etc?

sw/device/tests/aes_interrupt_encryption_test.c
```diff
@@ -0,0 +1,260 @@
...
+  } while (!ready);
+
+  // Disable the AES unit to no longer automatically start encryption/decryption
+
+  transaction.manual_operation = kDifAesManualOperationManual;
+
+  CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerDataOutClear));
+  CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerKeyIvDataInClear));
+
+  LOG_INFO(" De-initialized Automatic operation ");
```
luismarques, line 122: Nit: why the extra whitespace in the string?

sw/device/tests/aes_interrupt_encryption_test.c
```diff
@@ -0,0 +1,260 @@
...
+  memcpy(iv.iv, &iv_encrypt1, sizeof(iv.iv));
+  transaction.manual_operation = kDifAesManualOperationAuto;
+
+  CHECK_DIF_OK(dif_aes_start(&aes, &transaction, &key, &iv));
+
+  memcpy(plain_text[0].data, kAesModesPlainText_block34,
+         sizeof(kAesModesPlainText_block34));
+
+  CHECK_DIF_OK(dif_aes_get_status(&aes, kDifAesStatusInputReady, &ready));
+  LOG_INFO("Resuming Encrypt with the remaining 2 blocks of data ");
```
luismarques, line 195: Nit: ditto extraneous whitespace

luismarques (commented):
Can you add review / commit comments explaining what overall scenario this test tests? I don't suppose there's a testplan testpoint to reference?

sw/device/tests/aes_interrupt_encryption_test.c
```diff
@@ -0,0 +1,260 @@
...
+
+static const uint8_t kKeyShare1[] = {
+    0x0f, 0x1f, 0x2f, 0x3f, 0x4f, 0x5f, 0x6f, 0x7f, 0x8f, 0x9f, 0xaf,
+    0xbf, 0xcf, 0xdf, 0xef, 0xff, 0x0a, 0x1a, 0x2a, 0x3a, 0x4a, 0x5a,
+    0x6a, 0x7a, 0x8a, 0x9a, 0xaa, 0xba, 0xca, 0xda, 0xea, 0xfa,
+};
+
+OTTF_DEFINE_TEST_CONFIG();
+
+status_t execute_test(void) {
```
engdoreis, line 48: As this function returns `status_t`, you can replace all instances of `CHECK_DIF_OK` by `TRY`, which would make the code cleaner in my opinion.

sw/device/tests/aes_interrupt_encryption_test.c
```diff
@@ -0,0 +1,260 @@
...
+
+enum {
+  kTestTimeout = (1000 * 1000),
+};
+
+// The mask share, used to mask kKey. Note that the masking should not be done
+// manually. Software is expected to get the key in two shares right from the
+// beginning.
+
+static const unsigned char kAesModesPlainText_block12[32] = {
```
engdoreis, line 29: The name convention for `const` variables.
```suggestion
static const unsigned char kAesModesPlainTextBlock12[32] = {
```
The same would apply throughout the file.

sw/device/tests/aes_interrupt_encryption_test.c
rprakas-gsc: Ideally, I want the test code to report FAILURE & abort if the value doesn't get readout correctly , hence using `CHECK_DIF_OK`

rprakas-gsc:
> Can you add review / commit comments explaining what overall scenario this test tests? I don't suppose there's a testplan testpoint to reference?
There is.  I have added the below lines in the new Commit 
Added AES Interrupt Encryption tests as per test plan identified here:[[…](https://opentitan.org/book/hw/top_earlgrey/data/chip_testplan.html?highlight=chip_sw_aes_interrupt_encryption#chip_sw_aes_interrupt_encryption)
  and modified BUILD

sw/device/tests/aes_interrupt_encryption_test.c
rprakas-gsc: Fixed this in the updated aes_interrupt_encryption_Test.c code

sw/device/tests/BUILD
rprakas-gsc: Done!

sw/device/tests/aes_interrupt_encryption_test.c
rprakas-gsc: Fixed these comments with regular-style

sw/device/tests/aes_interrupt_encryption_test.c
rprakas-gsc: Fixed it - and amended the changes!

moidx:
Hi @nasahlpa, can you take a look at the test implementation to ensure it matches the intended test point description? Thanks!

nasahlpa:
> > Can you add review / commit comments explaining what overall scenario this test tests? I don't suppose there's a testplan testpoint to reference?
> > There is.  I have added the below lines in the new Commit
> > Added AES Interrupt Encryption tests as per test plan identified here:[[…](https://opentitan.org/book/hw/top_earlgrey/data/chip_testplan.html?highlight=chip_sw_aes_interrupt_encryption#chip_sw_aes_interrupt_encryption)
> > and modified BUILD

Could you please add the description to the git commit itself? Currently, no description is available in the git commit message. Furthermore, the contributing guideline also mentions that a [subject tag](https://opentitan.org/book/doc/contributing/detailed_contribution_guide/index.html#subject-tags) should be added to the git commit message.

rprakas-gsc:
> > > Can you add review / commit comments explaining what overall scenario this test tests? I don't suppose there's a testplan testpoint to reference?
> > > There is.  I have added the below lines in the new Commit
> > > Added AES Interrupt Encryption tests as per test plan identified here:[[…](https://opentitan.org/book/hw/top_earlgrey/data/chip_testplan.html?highlight=chip_sw_aes_interrupt_encryption#chip_sw_aes_interrupt_encryption)
> > > and modified BUILD
> 
> Could you please add the description to the git commit itself? Currently, no description is available in the git commit message. Furthermore, the contributing guideline also mentions that a [subject tag](https://opentitan.org/book/doc/contributing/detailed_contribution_guide/index.html#subject-tags) should be added to the git commit message.

The git commit has restrictions on word limit, if I include the entire link, i will  run out of the world limit and won't be able to commit successfully.

nasahlpa:
Understood. But you could add something like this:
```
This commit implements the chip_sw_aes_interrupt_encryption test that is defined in the chip testplan. The purpose of this test is to check, whether an AES encryption can be interrupted and restored.
```
For the header of the git commit message you could use:
```
[sival/aes] Add AES interrupt encryption test
```

rprakas-gsc:
Thanks Pascal.  I will incorporate these changes in the commit message now!

rprakas-gsc:
Just updated the PR with the commit messages as suggested!

nasahlpa:
Thanks! Would you mind looking into the review comments I've added? Especially the question regarding the decryption should be addressed.

moidx:
Hi @nasahlpa, I wasn't able to find your comments. Can you double check if they were sent? Thanks

nasahlpa (changes requested):

nasahlpa:
> Hi @nasahlpa, I wasn't able to find your comments. Can you double check if they were sent? Thanks

Oh sorry, now the comments should be public.

sw/device/tests/aes_interrupt_encryption_test.c
engdoreis: The final result would be the same, `TRY` would return immediately and then the callee would abort by using `CHECK_STATUS_OK`.

sw/device/tests/BUILD
engdoreis, line 1: Could you please re-order the commit message? The  `Signed-off` line should be after the description.

sw/device/tests/aes_interrupt_encryption_test.c
engdoreis, line 1: I suggest the following commit message:
[test, aes] Clean-up log message.

sw/device/tests/aes_interrupt_encryption_test.c
```diff
@@ -0,0 +1,261 @@
...
+    //  Trial encryption ends here
+    //  Resume the Original Encryption below
+
+    // Using the iv read from `iv_encrypt1` for resuming the encryption of
+    // remaining 2 blocks.
+
+    memcpy(iv.iv, &iv_encrypt1, sizeof(iv.iv));
+    transaction.manual_operation = kDifAesManualOperationAuto;
+
+    // "Convert" key share byte arrays to `dif_aes_key_share_t`.
```
nasahlpa, line 198: I think this is not necessary as `key` should still contain the key used for the first two blocks?

nasahlpa (approved):
Thanks @rprakas-gsc for changing the code based on the feedback. Apart from a small nit this PR looks good to me.

Maybe @engdoreis could have another look.

engdoreis (approved):
Thanks for the changes.

jwnrt:
@rprakas-gsc sorry we missed this in review, but would you mind filing a PR that assigns this test to the testpoint in the testplan please?

https://github.com/lowRISC/opentitan/blob/3a4d766b45e58af05dfccd81a861927724a4783f/hw/top_earlgrey/data/ip/chip_aes_testplan.hjson#L115

i.e.:
```
bazel: ["//sw/device/tests/aes_interrupt_encryption_test"]
```

Feel free to add me as a reviewer

rprakas-gsc:
Thanks James! I have create a separate PR request (https://github.com/lowRISC/opentitan/pull/25501) for this.

PR #24575, opened 2024-09-16
[doc] update cw340 jumper table references

anthonychen1251:
This updates the jumper table to correctly reference J23/J25 for the SPI device and J12/J13 for the JTAG connection on the CW340 base board.

The image of blue socket-to-socket jumpers on cw340 board is attached for the reference
7s2RvpzohtFdBLv


jwnrt (approved):
Thanks!

PR #24577, opened 2024-09-16
[dv,spi] Set strong drive strength and fast slew rate for SPI pads

sha-ron:
SPI pads should use strong drive strength and fast slew rate.

### DIFF
--- a/sw/device/tests/power_virus_systemtest.c
+++ b/sw/device/tests/power_virus_systemtest.c
@@ -32,6 +32,7 @@
 #include "sw/device/lib/testing/otbn_testutils_rsa.h"
 #include "sw/device/lib/testing/pinmux_testutils.h"
 #include "sw/device/lib/testing/spi_device_testutils.h"
+#include "sw/device/lib/testing/spi_host_testutils.h"
 #include "sw/device/lib/testing/test_framework/check.h"
 #include "sw/device/lib/testing/test_framework/ottf_macros.h"
 #include "sw/device/lib/testing/test_framework/ottf_main.h"
@@ -286,14 +287,6 @@ static const uint8_t kI2cMessage[] = {
     0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa,
     0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa,
     0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa,
-};
-
-static const top_earlgrey_direct_pads_t spi_host0_direct_pads[5] = {
-    kTopEarlgreyDirectPadsSpiHost0Sck,  // sck
-    kTopEarlgreyDirectPadsSpiHost0Sd3,  // sio[3]
-    kTopEarlgreyDirectPadsSpiHost0Sd2,  // sio[2]
-    kTopEarlgreyDirectPadsSpiHost0Sd1,  // sio[1]
-    kTopEarlgreyDirectPadsSpiHost0Sd0   // sio[0]
 };
 
 static void log_entropy_src_alert_failures(void) {
@@ -693,40 +686,9 @@ static void configure_pinmux_sim(void) {
   CHECK_DIF_OK(dif_pinmux_output_select(&pinmux, kTopEarlgreyPinmuxMioOutIor11,
                                         kTopEarlgreyPinmuxOutselPwmAonPwm0));
 
-  // Set fast slew rate and strong drive strengh for SPI host0 pads.
-  dif_pinmux_pad_attr_t out_attr;
-  dif_pinmux_pad_attr_t in_attr = {
-      .slew_rate = 1,
-      .drive_strength = 3,
-      // Set weak pull-ups for all the pads.
-      .flags = kDifPinmuxPadAttrPullResistorEnable |
-               kDifPinmuxPadAttrPullResistorUp};
-  dif_result_t res;
-  for (uint32_t i = 0; i <= ARRAYSIZE(spi_host0_direct_pads); ++i) {
-    res = dif_pinmux_pad_write_attrs(&pinmux, spi_host0_direct_pads[i],
-                                     kDifPinmuxPadKindDio, in_attr, &out_attr);
-    if (res == kDifError) {
-      // Some target platforms may not support the specified value for slew rate
-      // and drive strength. If that's the case, use the values actually
-      // supported.
-      if (out_attr.slew_rate != in_attr.slew_rate) {
-        LOG_INFO(
-            "Specified slew rate not supported, trying supported slew rate");
-        in_attr.slew_rate = out_attr.slew_rate;
-      }
-      if (out_attr.drive_strength != in_attr.drive_strength) {
-        LOG_INFO(
-            "Specified drive strength not supported, trying supported drive "
-            "strength");
-        in_attr.drive_strength = out_attr.drive_strength;
-      }
-      CHECK_DIF_OK(dif_pinmux_pad_write_attrs(&pinmux, spi_host0_direct_pads[i],
-                                              kDifPinmuxPadKindDio, in_attr,
-                                              &out_attr));
-      // Note: fallthrough with the modified `in_attr` so that the same
-      // attributes are used for all pads.
-    }
-  }
+  // Configure fast slew rate, strong drive strength, and weak pull-ups for SPI
+  // Host 0 pads.
+  CHECK_STATUS_OK(spi_host_testutils_configure_host0_pad_attrs(&pinmux));
 
   // Configure fast slew rate and strong drive strength for SPI device pads.
   CHECK_STATUS_OK(spi_device_testutils_configure_pad_attrs(&pinmux));
--- a/sw/device/tests/power_virus_systemtest.c
+++ b/sw/device/tests/power_virus_systemtest.c
@@ -288,6 +288,21 @@ static const uint8_t kI2cMessage[] = {
     0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa,
 };
 
+static const top_earlgrey_direct_pads_t spi_host0_direct_pads[5] = {
+    kTopEarlgreyDirectPadsSpiHost0Sck,  // sck
+    kTopEarlgreyDirectPadsSpiHost0Sd3,  // sio[3]
+    kTopEarlgreyDirectPadsSpiHost0Sd2,  // sio[2]
+    kTopEarlgreyDirectPadsSpiHost0Sd1,  // sio[1]
+    kTopEarlgreyDirectPadsSpiHost0Sd0   // sio[0]
+};
+
+static const top_earlgrey_direct_pads_t spi_device_direct_pads[4] = {
+    kTopEarlgreyDirectPadsSpiDeviceSd3,  // sio[3]
+    kTopEarlgreyDirectPadsSpiDeviceSd2,  // sio[2]
+    kTopEarlgreyDirectPadsSpiDeviceSd1,  // sio[1]
+    kTopEarlgreyDirectPadsSpiDeviceSd0   // sio[0]
+};
+
 static void log_entropy_src_alert_failures(void) {
   dif_entropy_src_alert_fail_counts_t counts;
   CHECK_DIF_OK(dif_entropy_src_get_alert_fail_counts(&entropy_src, &counts));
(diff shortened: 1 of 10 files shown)


moidx (approved):

moidx:
CI has the following error: 

```
E00003 power_virus_systemtest.c:1468] DIF-fail: dif_pinmux_pad_write_attrs(&pinmux, kTopEarlgreyDirectPadsSpiHost0Sck, kDifPinmuxPadKindDio, in_attr, &out_attr) returns 13
```

Code: https://github.com/lowRISC/opentitan/blob/7e34e67ade0717418930ecbfa4bee30ba44979ed/sw/device/lib/dif/dif_pinmux.c#L307-L309

This only happens in the `sival_rom_ext` configuration. Need to check if we are locking pinmux register in that case

a-will:
> CI has the following error:
> 
> ```
> E00003 power_virus_systemtest.c:1468] DIF-fail: dif_pinmux_pad_write_attrs(&pinmux, kTopEarlgreyDirectPadsSpiHost0Sck, kDifPinmuxPadKindDio, in_attr, &out_attr) returns 13
> ```
> 
> Code:
> 
> https://github.com/lowRISC/opentitan/blob/7e34e67ade0717418930ecbfa4bee30ba44979ed/sw/device/lib/dif/dif_pinmux.c#L307-L309
> 
> This only happens in the `sival_rom_ext` configuration. Need to check if we are locking pinmux register in that case

It's happening in all FPGA configurations because these pad attributes are not available. Pad attributes should not be written without proper matching against the configuration.

moidx:
Thanks @a-will for your quick response. 

@sha-ron, you will probably want to constraint this configuration to the follow device types: 

* `kDeviceSilicon`
* `kDeviceSimDV`

sha-ron:
Thank you @a-will, @moidx for the review and help with the CI failures. I added constrains on the configurations as suggested. Hopefully Ci will pass now.

sw/device/tests/power_virus_systemtest.c
```diff
@@ -1454,6 +1454,48 @@ bool test_main(void) {
   if (kDeviceType == kDeviceSimDV || kDeviceType == kDeviceSimVerilator) {
     configure_pinmux_sim();
   }
+
```
a-will, line 1457: These should probably be moved to be inside the `configure_pinmux_sim()` function.

sw/device/tests/power_virus_systemtest.c
```diff
@@ -684,6 +684,49 @@ static void configure_pinmux_sim(void) {
...
+                                   kDifPinmuxPadKindDio, in_attr, &out_attr));
+
+    CHECK_DIF_OK(
+        dif_pinmux_pad_write_attrs(&pinmux, kTopEarlgreyDirectPadsSpiHost0Sd2,
+                                   kDifPinmuxPadKindDio, in_attr, &out_attr));
+
+    CHECK_DIF_OK(
+        dif_pinmux_pad_write_attrs(&pinmux, kTopEarlgreyDirectPadsSpiHost0Sd3,
+                                   kDifPinmuxPadKindDio, in_attr, &out_attr));
+    // Spi_Device
```
timothytrippel, line 714: Aren't these attributes already set on the SPI Host 0 data pins? See lines 580 -- 601 in this file. Unless I am missing something here?

timothytrippel:
@sha-ron should this change be eventually cherry-picked over to the `earlgrey_1.0.0` branch too?

sw/device/tests/power_virus_systemtest.c
sha-ron: I think lines 580-601 only set pull up and not drive strength/slew rate.

sha-ron:
> @sha-ron should this change be eventually cherry-picked over to the `earlgrey_1.0.0` branch too?

Yes, it should be merged also to earlgrey_1.0.0 branch (it is required for GLS simulation).

sw/device/tests/power_virus_systemtest.c
timothytrippel: sure, but can you merge this config with the lines I mentioned above? So we don't set the attributes twice?

sha-ron: I tried, but dif_pinmux_pad_attr_flags https://github.com/lowRISC/opentitan/blob/ae8ac572e793f34f457d08c9ce03db3644c49f5a/sw/device/lib/dif/dif_pinmux.h#L140 doesn't have drive_strength and slew_rate fields. Am I missing something here?

pamaury: The slew is an attribute, not an attribute flag: https://github.com/lowRISC/opentitan/blob/ae8ac572e793f34f457d08c9ce03db3644c49f5a/sw/device/lib/dif/dif_pinmux.h#L158

timothytrippel: Sorry should have been more clear here. To do this properly would require an alignment between these two structs: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/pinmux_testutils.h;drc=30d7e787c753caaa03fe68a4a70da1bbcbc1d96f;l=99

and

https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/dif/dif_pinmux.h;l=168?q=dif_pinmux.h&ss=opentitan%2Fopentitan

Which is more refactoring. It is ok to leave as is in this PR and we can file a follow up PR to align these later.

timothytrippel (approved):

timothytrippel:
@sha-ron : looks like this breaks CI. Can you fix the failures? Then we can get this merged.

sha-ron:
Thank you @timothytrippel. It looks like CI is failing over FPGA checks. I use 'if (kDeviceType == kDeviceSilicon || kDeviceType == kDeviceSimDV)' so I am not sure if/how it is related to the changes in the PR. Do you understand what it is failing on?

timothytrippel:
@sha-ron It looks like there are also DV failures, see the "Private CI" jobs

sha-ron:
I see now:( It looks like it is related to the pad attribute definition in the open source https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim_generic/rtl/prim_generic_pad_attr.sv#L53 that is different. Not sure what is the reason for this definition but at stage maybe it will be better to add all the code under  some `ifdef. WDYT?

vogelpi:
You're right @sha-ron , depending on the platform (silicon, open source DV, or Verilator), a different number of drive strength bits are supported. Other bits are tied to zero. Some other pad attributes are treated similarly.

I think the real issue here is that these pad attribute registers are actually defined as WARL (write any, read legal) while the DIF treats them as RW (read and write): There is a check in the DIF which errors out in case an unsupported bit is written. This doesn't make sense for a WARL register. I've now filed #24688 to change this.

I also tested your PR on top of the #24688 and it works. So once #24688 is merged, we can rebase this one and it should pass CI.

sw/device/tests/power_virus_systemtest.c
```diff
@@ -684,6 +684,49 @@ static void configure_pinmux_sim(void) {
   //    Channel 0 on IOR11
   CHECK_DIF_OK(dif_pinmux_output_select(&pinmux, kTopEarlgreyPinmuxMioOutIor11,
                                         kTopEarlgreyPinmuxOutselPwmAonPwm0));
+
+  if (kDeviceType == kDeviceSilicon || kDeviceType == kDeviceSimDV) {
```
vogelpi, line 688: We can probably do the attribute configuration unconditionally once #24688 is in.

vogelpi (approved):
The changes look good to me @sha-ron , to get this through CI we will have to rebase this on top of #24688.

vogelpi:
> > CI has the following error:
> > ```
> > E00003 power_virus_systemtest.c:1468] DIF-fail: dif_pinmux_pad_write_attrs(&pinmux, kTopEarlgreyDirectPadsSpiHost0Sck, kDifPinmuxPadKindDio, in_attr, &out_attr) returns 13
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Code:
> > https://github.com/lowRISC/opentitan/blob/7e34e67ade0717418930ecbfa4bee30ba44979ed/sw/device/lib/dif/dif_pinmux.c#L307-L309
> > 
> > This only happens in the `sival_rom_ext` configuration. Need to check if we are locking pinmux register in that case
> 
> It's happening in all FPGA configurations because these pad attributes are not available. Pad attributes should not be written without proper matching against the configuration.

Ah, sorry @a-will , I only saw your message now. So the approach I took in #24688 is probably wrong then. Can someone please provide some guidance on how to match the attributes against the configuration? We could provide a platform-dependent mask similar to how we specify clock frequencies for FPGA, simulation and silicon. But the main issue here is that there are differences between the different pad implementations between closed and open source. I don't know how we should handle this if we want to keep the check referenced by Alex above.

a-will:
> > > CI has the following error:
> > > ```
> > > E00003 power_virus_systemtest.c:1468] DIF-fail: dif_pinmux_pad_write_attrs(&pinmux, kTopEarlgreyDirectPadsSpiHost0Sck, kDifPinmuxPadKindDio, in_attr, &out_attr) returns 13
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Code:
> > > https://github.com/lowRISC/opentitan/blob/7e34e67ade0717418930ecbfa4bee30ba44979ed/sw/device/lib/dif/dif_pinmux.c#L307-L309
> > > 
> > > This only happens in the `sival_rom_ext` configuration. Need to check if we are locking pinmux register in that case
> > 
> > 
> > It's happening in all FPGA configurations because these pad attributes are not available. Pad attributes should not be written without proper matching against the configuration.
> 
> Ah, sorry @a-will , I only saw your message now. So the approach I took in #24688 is probably wrong then. Can someone please provide some guidance on how to match the attributes against the configuration? We could provide a platform-dependent mask similar to how we specify clock frequencies for FPGA, simulation and silicon. But the main issue here is that there are differences between the different pad implementations between closed and open source. I don't know how we should handle this if we want to keep the check referenced by Alex above.

This is one of those special pieces of top-specific issues that we have been ignoring. Is this the first time we have had a mismatch between generic DV and GLS where the same code paths can't be used? In some ways, the target platform is closer to the "silicon" one than the generic "dv" one... Should this be a silicon device type for software, but tests run with DV tooling?

For now, I'd probably just choose to check the readback value under the DV device type and explicitly ignore mismatches for drive strength. Then add a TODO to improve this with multi-top. Pad attributes are going to be one of those gnarly areas to solve. sw/device/lib/arch/device.h is a kludge.

pamaury:
I wonder if this should be part of the [spi_device testutils](https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/testing/spi_device_testutils.h)?
This would avoid duplicating code in several places.

vogelpi:
Thanks a lot @a-will for your feedback, this is really appreciated!

@sha-ron , I took the liberty to directly update this PR with an implementation that follows the approach suggested by Alex and that we also use in other files. The idea is to configure the pad attributes and then compare the returned / accepted attributes against the specified ones. As previously discussed, different target platforms may use pad with different max slew rate and drive strength, so we can kind of expect that there might be a mismatch for these attributes. The implementation now accepts a mismatch in the drive strength and slew rate but not for other attributes.

After reworking these checks, I ran into issues with the power virus and the passthrough tests: I got assertion failures at the top-level and inside SPI host. It seems we should not enable the pull up for SPI device for these tests. I am no expert on this obviously, but does that make sense to you?

vogelpi:
> I wonder if this should be part of the [spi_device testutils](https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/testing/spi_device_testutils.h)? This would avoid duplicating code in several places.

I thought about that, too. But I think we should merge this PR and cherry pick it into the TO release branch as soon as possible to unblock @sha-ron . Maybe the restructuring can be done as a follow-up? I am happy for someone else to take this on but I should move on to another higher prio task.

pamaury:
Sure, just create an issue to remember to do it, it's not urgent to refactor now.

vogelpi:
@pamaury , I've now added two more comments to factor our that part of the pad attributes configuration which can be shared between tests into spi_host/device_testutils.

pamaury:
> @pamaury , I've now added two more comments to factor our that part of the pad attributes configuration which can be shared between tests into spi_host/device_testutils.

Thanks, it looks great. Is there any reason you have not converted `spi_host_tx_rx.c`?

vogelpi:
Thanks @pamaury , the configuration of SPI Host 0 in `spi_host_tx_rx.c` should now be shared. SPI Host 1 is a bit different in the sense that we set up pad attributes but also the muxing. I could probably share some of that logic as well, but it's not used in another test yet. So we probably don't gain much right now by doing this.

sha-ron:
> Thanks a lot @a-will for your feedback, this is really appreciated!
> 
> @sha-ron , I took the liberty to directly update this PR with an implementation that follows the approach suggested by Alex and that we also use in other files. The idea is to configure the pad attributes and then compare the returned / accepted attributes against the specified ones. As previously discussed, different target platforms may use pad with different max slew rate and drive strength, so we can kind of expect that there might be a mismatch for these attributes. The implementation now accepts a mismatch in the drive strength and slew rate but not for other attributes.
> 
> After reworking these checks, I ran into issues with the power virus and the passthrough tests: I got assertion failures at the top-level and inside SPI host. It seems we should not enable the pull up for SPI device for these tests. I am no expert on this obviously, but does that make sense to you?

Yes I don't see a reason to enable pull up for SPI device.

vogelpi:
Okay, thanks for the feedback @sha-ron !

PR #24581, opened 2024-09-16
[reggen,clkmgr] Add flops to deglitch shadow storage error

### LINKED ISSUE #24592: [clkmgr] Glitch in shadow register storage error
clkmgr has five shadowed registers:
- `IO_MEAS_CTRL_SHADOWED`
- `IO_DIV2_MEAS_CTRL_SHADOWED`
- `IO_DIV4_MEAS_CTRL_SHADOWED`
- `MAIN_MEAS_CTRL_SHADOWED`
- `USB_MEAS_CTRL_SHADOWED`

Combinational logic inside `prim_subreg_shadow` compares the value of the shadowed register to that of the committed register and asserts the `err_storage` output if there's a mismatch:
https://github.com/lowRISC/opentitan/blob/282d863055979595a24d393333fa6b8347a2a9e4/hw/ip/prim/rtl/prim_subreg_shadow.sv#L184

In `clkmgr` (and only there, at least in the current Earlgrey), this output goes through a CDC into the `IO_DIV4` (powerup) domain . Take the example of `IO_MEAS_CTRL_SHADOWED`:
https://github.com/lowRISC/opentitan/blob/282d863055979595a24d393333fa6b8347a2a9e4/hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr_reg_top.sv#L1345
https://github.com/lowRISC/opentitan/blob/282d863055979595a24d393333fa6b8347a2a9e4/hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr_reg_top.sv#L1294-L1302

The problem is that the input of the CDC flop is driven by combinational logic. That is, as a register changes and its new value ripples through the comparator, the comparator could temporarily have inequality as result, and if you're unlucky, the CDC flops just then.

To fix this, the output of the comparator (combinational logic in general) needs to be flopped in the source clock domain (where glitches can be prevented through STA), and the output of that flop then needs to go to the CDC flop.

andreaskurth:
Fixes #24592 on the `earlgrey_1.0.0` release branch.

### DIFF
--- a/hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr_reg_top.sv
+++ b/hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr_reg_top.sv
@@ -1288,7 +1288,18 @@ module clkmgr_reg_top (
     io_io_meas_ctrl_shadowed_we & io_io_meas_ctrl_shadowed_regwen;
   //   F[hi]: 9:0
   logic async_io_meas_ctrl_shadowed_hi_err_update;
-  logic async_io_meas_ctrl_shadowed_hi_err_storage;
+  logic async_io_meas_ctrl_shadowed_hi_err_storage, deglitched_io_meas_ctrl_shadowed_hi_err_storage;
+
+  // flop storage error to filter combinational glitches before sending it across CDC
+  prim_flop #(
+    .Width(1),
+    .ResetValue('0)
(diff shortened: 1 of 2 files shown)


andreaskurth:
Superseded by https://github.com/lowRISC/opentitan/pull/24622. Closing.

PR #24587, opened 2024-09-17
[bazel] Update `rules_rust` to upstream 0.49.3

jwnrt:
This PR updates `rules_rust` to the upstream version 0.49.3 in order to gain `bzlmod` support.

The change-blocker CI job will fail because too many files have changed, so we should check that manually. The Verible lint will also fail but I'm not touching any RTL.

I've had to include a bump to `rules_python` as well because there was a very strange bug in calling `tockloader` triggered by this PR. No idea why it was triggered or what was causing it...


third_party/rust/BUILD
```diff
@@ -28,7 +28,6 @@ rust_bindgen_toolchain(
         ":specify_bindgen_libstdcxx": "@bindgen_libstdcxx_linux//:libstdc++",
         "//conditions:default": None,
     }),
-    system_includes = "@bindgen_clang_linux//:system_includes",
```
pamaury, line 31: Do you know what the purpose of this was in the first place?

pamaury (approved):
Looks good, thanks @jwnrt

third_party/rust/BUILD
jwnrt: Yes, it was for adding the system toolchain's `include`s when installed in a different location. Since we patched this, `rules_rust` updated their bindgen toolchain to take the system includes from the hermetic C toolchain instead, so I don't think we need this anymore:

https://github.com/bazelbuild/rules_rust/commit/116f6db8fa6caa1f1cc7470d7f5a0d0e32b5352d

a-will: @timothytrippel might be interested in how this affects the air gapped environment's setup.

timothytrippel: I think this should be ok, but we will have to see if this breaks the nightly regressions.

jwnrt: Thanks, I think I'll break this PR up so we can revert the `rules_rust` update separately is regressions fail.

sw/host/hsmtool/acorn/acorn.rs
```diff
@@ -271,7 +271,7 @@ impl Acorn {
         unsafe {
             let public_key = acorn_bindgen::acorn_buffer {
                 // Transmute because the acorn API wants a mut ptr (but it wont mutate).
-                ptr: std::mem::transmute(public_key.as_ptr()),
+                ptr: std::mem::transmute::<*const u8, *mut u8>(public_key.as_ptr()),
```
nbdd0121, line 274: Why is this change needed? More over, this should be `pubic_key.as_ptr().cast_mut()`?

jwnrt: Clippy asked for it, but yes you’re right I should use that method

jwnrt:
I have reverted the update to the Rust toolchain and updated to a recent (but not latest) version of `rules_rust` instead. The latest Rust toolchain is incompatible with our LLVM-16 linker for RISC-V binaries, so I'm deferring that to another date.

nbdd0121 (approved):

PR #24590, opened 2024-09-17
[ci] Add `quick_lint` job to GitHub Actions CI workflow

jwnrt:
In Azure, we use these quick lints as a filter before we run the rest of the pipeline. We need this job running in parallel in GitHub Actions to start porting things over.

* I've dropped the "Determine build type" step as we don't need this right now and it's not a lint anyway.
* I also dropped the "Confirm no `.bazelrc-site`" step since it's already in `.gitignore` and this check seems superfluous.
* Otherwise, no changes.

This PR also ports the `ci/install-dependencies.{sh,yml}` files from Azure to GitHub (with some duplication). I'd still like to extract the Verilator and Verible versions to some external file. Same for the Python version ideally.

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -1,0 +1,67 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Install dependencies
+description: Install system dependencies needed for OpenTitan
+
+inputs:
+  verilator-version:
+    description: Verilator version to install
+    required: true
+    default: '4.210'
+  verilator-path:
+    description: Path at which to install Veriltator
+    required: true
+    default: /tools/verilator
+  verible-version:
+    description: Verible version to install
+    required: true
+    default: 'v0.0-3622-g07b310a3'
+  verible-path:
+    description: Path at which to install Verible
+    required: true
+    default: /tools/verible
+
+runs:
+  using: composite
+  steps:
+    - name: Install system dependencies
+      run: grep '^[^#]' apt-requirements.txt | xargs sudo apt install -y
+      shell: bash
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.8'
+        cache-dependency-path: python-requirements.txt
+        cache: pip
+
+    - name: Install Python dependencies
+      run: python3 -m pip install -r python-requirements.txt --require-hashes
+      shell: bash
+
+    - name: Install Verilator
+      run: |
+        VERILATOR_TAR="verilator-v${{ inputs.verilator-version }}.tar.gz"
+        VERILATOR_URL="https://storage.googleapis.com/verilator-builds/${VERILATOR_TAR}"
+        curl -f -Ls -o "/tmp/${VERILATOR_TAR}" "$VERILATOR_URL" || {
+          echo "failed to download Verilator from ${VERILATOR_URL}" >&2
+          exit 1
+        }
+        sudo mkdir -p "${{ inputs.verilator-path }}"
+        sudo tar -C "${{ inputs.verilator-path }}" -xvzf "/tmp/${VERILATOR_TAR}" --strip-components=1
+        echo "${{ inputs.verilator-path }}/bin" >> "$GITHUB_PATH"
+      shell: bash
+
+    - name: Install Verible
+      run: |
+        VERIBLE_TAR="verible-${{ inputs.verible-version }}-linux-static-x86_64.tar.gz"
+        VERIBLE_URL="https://github.com/chipsalliance/verible/releases/download/${{ inputs.verible-version }}/${VERIBLE_TAR}"
+        curl -f -Ls -o "/tmp/${VERIBLE_TAR}" "$VERIBLE_URL" || {
+          echo "failed to download Verible from ${VERIBLE_URL}" >&2
+          exit 1
+        }
+        sudo mkdir -p "${{ inputs.verible-path }}"
+        sudo tar -C "${{ inputs.verible-path }}" -xvzf "/tmp/${VERIBLE_TAR}" --strip-components=1
+        echo "${{ inputs.verible-path }}/bin" >> "$GITHUB_PATH"
+      shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0 # Bitstream cache requires all commits.
-      - name: Install system dependencies
-        run: grep '^[^#]' apt-requirements.txt | xargs sudo apt install -y
+      - name: Install dependencies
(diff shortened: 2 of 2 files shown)


pamaury (approved):
This looks good to me, thanks James

jwnrt:
GitHub's `cache` action seems a little broken:

* It doesn't have permission to restore `/tools/verilator`: https://github.com/actions/cache/issues/629.
* It isn't able to save caches under `$HOME`: https://github.com/actions/cache/issues/1332

I've moved the EDA tool install location to `/tmp/` and it seems to be working.

jwnrt:
Sorry, last force-push... The verible `PATH` was wrong.

.github/workflows/ci.yml
```diff
@@ -6,16 +6,57 @@ name: CI
 on:
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
nbdd0121, line 11:
```suggestion
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,98 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Enable caching for APT packages
+      uses: actions/cache@v4
+      with:
+        # We cache the directory where APT downloads `.deb` archives for packages.
+        path: ${{ inputs.apt-archive-path }}
+        key: ${{ runner.os }}-${{ hashFiles('apt-requirements.txt') }}
+      id: cache-apt
```
nbdd0121, line 39: Hmm. If we run this action on self-hosted runners then it would be downloading and uploading to Azure from Google Cloud?

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,98 @@
...
+          xargs sudo apt -o Dir::Cache::archives=${{ inputs.apt-archive-path }} install --download-only
+        grep '^[^#]' apt-requirements.txt | \
+          xargs sudo apt -o Dir::Cache::archives=${{ inputs.apt-archive-path }} install -y
+        # Allow the cache-saving step to read the directory contents.
+        sudo chmod -R +rw ${{ inputs.apt-archive-path }}
+      shell: bash
+
+    - uses: actions/setup-python@v5
+      with: # TODO: use `python-version-file` when we have `pyproject.toml`.
+        python-version: 3.9
```
nbdd0121, line 55: This should use 3.8

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,98 @@
...
+        cache-dependency-path: python-requirements.txt
+        cache: pip
+
+    - name: Install Python dependencies
+      run: python3 -m pip install -r python-requirements.txt --require-hashes
```
nbdd0121, line 60: Could change to uv

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,98 @@
...
+
+inputs:
+  verilator-version:
+    description: Verilator version to install
+    required: true
+    default: '4.210'
+  verilator-path:
+    description: Path at which to install Veriltator
+    required: true
+    default: /tmp/verilator
```
nbdd0121, line 16: Maybe `sudo mkdir -p /tools && sudo chown $USER: /tools` would allow the normal path while allowing caching?

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,98 @@
...
+          ${{ inputs.verilator-path }}
+          ${{ inputs.verible-path }}
+        key: ${{ runner.os }}-${{ inputs.verilator-version }}-${{ inputs.verible-version }}
+      id: cache-eda
+
+    - name: Install Verilator
+      if: ${{ steps.cache-eda.outputs.cache-hit != 'true' }}
+      run: |
+        VERILATOR_TAR="verilator-v${{ inputs.verilator-version }}.tar.gz"
+        VERILATOR_URL="https://storage.googleapis.com/verilator-builds/${VERILATOR_TAR}"
```
nbdd0121, line 76: This switches downloading from Google bucket to GitHub. Is there any significant saving by doing this? GitHub cache space is limited (and this would save a copy for each PR + branch).

.github/actions/install-deps/action.yml
```diff
@@ -0,0 +1,98 @@
...
+          echo "failed to download Verilator from ${VERILATOR_URL}" >&2
+          exit 1
+        }
+        mkdir -m 777 -p "${{ inputs.verilator-path }}"
+        tar -C "${{ inputs.verilator-path }}" -xvzf "/tmp/${VERILATOR_TAR}" --strip-components=1
+        echo "${{ inputs.verilator-path }}/bin" >> "$GITHUB_PATH"
+      shell: bash
+
+    - name: Install Verible
+      if: ${{ steps.cache-eda.outputs.cache-hit != 'true' }}
```
nbdd0121, line 87: This switches from downloading from GitHub release to GitHub cache. Probably not worth doing it?

.github/workflows/ci.yml
```diff
@@ -6,16 +6,57 @@ name: CI
...
+
 jobs:
+  quick_lint:
+    name: Lint (quick)
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v4
+        with: { fetch-depth: 0 } # Required so we can lint commit messages.
```
nbdd0121, line 19: I would still prefer use just normal YAML instead of embed JSON in yaml
```suggestion
        with:
          fetch-depth: 0 # Required so we can lint commit messages.
```

.github/actions/install-deps/action.yml
jwnrt: Should I disable this when self-hosted?

I can add `if ${{ runner.environment == 'github-hosted' }}` to the cache step

.github/actions/install-deps/action.yml
jwnrt: Changed back to `/tools/verilator` now the cache is removed

.github/actions/install-deps/action.yml
jwnrt: Probably not, I've removed the EDA tool cache

.github/actions/install-deps/action.yml
jwnrt: Actually, the apt cache doesn't seem to even give us much improvement since the downloading was not the slowest part, the installing is. I'll remove this as well.

.github/actions/install-deps/action.yml
jwnrt: Just tried this, but `uv` will only install into a `venv` and not the system. I don't want to have to `source .venv/bin/activate` and the start of every workflow step, and the speed increase wasn't very significant anyway.

nbdd0121 (approved):

jwnrt:
Changed `setup-python` to read the python version from `pyproject.toml`, but it chose to use `3.12`. This is supposedly supported by our `pyproject.toml` but doesn't work yet.

Getting OpenTitan working on 3.12 looks like a non-trivial amount of work, so deferring for another PR.

PR #24591, opened 2024-09-17
[bazel] Remove orphaned `gen_requirements.sh` script

jwnrt:
This was no longer needed after the `rules_python` upgrade but was not removed.


nbdd0121 (approved):

timothytrippel (approved):
Thanks @jwnrt !

PR #24594, opened 2024-09-18
[ci] Cut down `apt-requirements`

jwnrt:
These system dependencies don't seem to be required any more.

### DIFF
--- a/apt-requirements.txt
+++ b/apt-requirements.txt
@@ -9,26 +9,19 @@
 # and are maintained in yum-requirements.txt.
 #
 # Keep it sorted.
-autoconf
-bison
 brotli
 build-essential
 clang-format
-cmake
 curl
 doxygen
-flex
 g++
 git
-golang
-lcov
 libelf1
 libelf-dev
 libftdi1-2
 libftdi1-dev
 # A requirement of the prebuilt clang toolchain.
 libncursesw5
-libpcsclite-dev
 libssl-dev
 libudev-dev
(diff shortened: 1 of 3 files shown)


pamaury:
I see that we have a `cmake` rule [here](https://github.com/lowRISC/opentitan/blob/master/third_party/hsm/BUILD.softhsm2.bazel#L14).
I am not sure if this uses the system's cmake or not.

Our documentation recommends `autoconf` to build verilator [here](https://github.com/lowRISC/opentitan/blob/master//doc/getting_started/setup_verilator.md).

We also seem to depend on ninja in [airgapped](https://github.com/lowRISC/opentitan/blob/master//util/prep-bazel-airgapped-build.sh) builds and in [tool requirement checks](https://github.com/lowRISC/opentitan/util/check_tool_requirements.py) and in some [weird](https://github.com/lowRISC/opentitan/hw/ip/otbn/dv/verilator/run-some.py) files.

I would first double-check if those files are outdated or not. If we remove the tools for apt, then we should update the doc/files.

jwnrt:
Hmm, you're right. I was hoping CI would catch any problems but there are two reasons why it might not:

* GitHub / Azure hosted runners come with some tools pre-installed (like cmake and autoconf)
* Our self-hosted runners come with packages pre-installed in the container, but that container isn't updated automatically for this PR

jwnrt:
I'm going to drop this, don't think it's worth the work.

Issue #24595, opened 2024-09-18
[topgen.py]FileNotFoundError: ‘hw/top_earlgrey/data/ip/ast/data/ast.hjson’


lin-snail:
### Description
- run the topgen.py command line:
 util/topgen.py -t hw/top_earlgrey/data/autogen/top_earlgrey.gen.hjson -o hw/top_earlgrey/
- issue:
FileNotFoundError: [Errno 2] No such file or directory: 'hw/top_earlgrey/data/ip/ast/data/ast.hjson'
![filenotfound](https://github.com/user-attachments/assets/a06aca61-b7b9-44cd-80df-257526fe8a6a)

rswarbrick:
I just ran locally with `make -C hw top` and everything worked (which is probably a workaround for you). I can also see the error you report if I run the command you show above.

Looking at the difference in what runs in the two cases, I notice that the working version runs
```
util/topgen.py -t hw/top_earlgrey/data/top_earlgrey.hjson -o hw/top_earlgrey/
```
(notice the different path to the hjson file). The path you're giving as an input is actually an output from `topgen`.

But the error that comes out is a bit rubbish! Let's keep this issue open until we've sorted that out.

lin-snail:
Thanks for your help.
It is ok with the following command line:
- make -C hw top
- util/topgen.py -t hw/top_earlgrey/data/top_earlgrey.hjson -o hw/top_earlgrey/

rswarbrick:
I'm keeping this open to track the fact that we could improve the tooling messages to explain what's going on more clearly. That change is hopefully available in the linked PR.

PR #24598, opened 2024-09-18
[topgen] Fail more understandably if we start in an unexpected place

rswarbrick:
This should fix issue #24595. The problem that caused the issue is that we compute paths to the various IP blocks that should be included by following relative paths from the `-t` argument. This doesn't work very well if the top_earlgrey.hjson file moves around!

A sensible design would probably point more explicitly at where to find IPs, but this should hopefully be a little easier to debug.

### DIFF
--- a/util/topgen.py
+++ b/util/topgen.py
@@ -742,6 +742,16 @@ def _process_top(
         else:
             ip_hjson = hjson_dir.parent / f"ip/{ip}/data/{ip}.hjson"
 
+            # In this situation, we are computing the path to the IP's hjson
+            # file by walking a relative path from hjson_dir. That hjson file
+            # should exist at this point, but things get confused if we started
+            # from a top file in an unexpected place. Check now to make sure
+            # this hasn't happened.
+            if not ip_hjson.is_file():
+                log.error(f'Cannot find a file at {ip_hjson}, which was '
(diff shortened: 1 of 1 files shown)


pamaury:
I believe `topgen` has a option to point to places where to look for hjson files (`--hjson-path`), the doc says:
> If defined, topgen uses supplied path to search for ip hjson. This applies only to ip's with the `reggen_only` attribute. If an hjson is located both in the conventional path and the alternate path, the
                        alternate path has priority.

rswarbrick:
Yep, you're right. The issue that prompted the change is triggered by a block that isn't marked `reggen_only`. The code (before this change) is:
```python
        if reggen_only and args.hjson_path:
            ip_hjson = Path(args.hjson_path) / f"{ip}.hjson"
        else:
            ip_hjson = hjson_dir.parent / f"ip/{ip}/data/{ip}.hjson"
```
where `hjson_dir` is derived from the `-t` argument.

This change is basically just adding a somewhat helpful error message earlier if the derived directory is wrong.

rswarbrick:
I've come back to this and it sees *lots* of merge conflicts. And I wasn't completely convinced it was a change worth doing! Let's worry about it some other time :-)

Issue #24597, opened 2024-09-18
[SPI, dv] Missing SPI pass through test with two stages read pipeline


sha-ron:
### Description

SPI top level test (not unit level) that can run on VCS simulation and configures CMD_INFO.read_pipeline_mode to 0x2.
https://opentitan.org/book/hw/ip/spi_device/doc/registers.html#cmd_info--read_pipeline_mode

moidx:
@vogelpi is it possible to assess if we have any targeted test cases for this mode of operation?

moidx:
This is currently impacting `earlgrey_1.0.0`. We would like to be able to get coverage here to be able to provide feedback to Nuvoton as part of the characterization exercise. 

This will also be needed for gen2.

vogelpi:
@a-will and @engdoreis provided some very useful feedback on this. There are two relevant tests for this feature.

SiVal
---

The `//sw/device/tests:spi_passthru_test` **SiVal** test **exercises all pipeline modes**. This can be seen from the following files:
- The device side C code [`spi_passthru_test.c`](https://github.com/lowRISC/opentitan/blob/48f0fb8747cc7b0a078f90d7e6bad8aa5258be01/sw/device/tests/spi_passthru_test.c#L30) uses a volatile constant `kReadPipelineMode` set to 0 (this corresponds to 0 stages in [cmd_info--read_pipeline_mode](https://opentitan.org/book/hw/ip/spi_device/doc/registers.html#cmd_info--read_pipeline_mode).
- The host side code Rust code [`spi_passthru.rs`](https://github.com/lowRISC/opentitan/blob/fb62e4880551fb4bea4c02aff86b3a1d145f5987/sw/host/tests/chip/spi_device/src/spi_passthru.rs#L499) overwrites this constant and cycles through all the different settings.

This test has been passing on FPGA and A1/A2 silicon for quite some time now according to the SiVal spreadsheet.
 
DV
---

[...]

PR #24601, opened 2024-09-19
[SPI,dv] Remove 30MHz SCK usage when not using two_stages_full_cycle …

sha-ron:
…read pipeline.
SCK frequency bigger than 24MHz is supported only when CMD_INFO.read_pipeline_mode is set to 0x2.
Currently the test is using CMD_INFO.read_pipeline_mode=0x0.

### DIFF
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_sw_spi_passthrough_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_sw_spi_passthrough_vseq.sv
@@ -14,7 +14,11 @@ class chip_sw_spi_passthrough_vseq extends chip_sw_base_vseq;
   // Frequencies to use for testing the sequences.
   time sck_periods_ps[] = '{
     41_000, // 24 MHz
+`ifndef GATE_LEVEL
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
I'm not sure this is the right way to make the change: you end up not testing 30MHz SCK even when `read_pipeline_mode` has a value that would allow it.

Is it possible to change the code at the bottom of `body` so that it skips frequencies that aren't supported (by looking to see how `read_pipeline_mode` is currently configured)?

sha-ron:
Thank you for the review @rswarbrick.
CMD_INFO (that includes read_pipeline_mode field) is actually a set of registers and the test has internal loop over a few commands so I am not sure how to find the correct read_pipeline_mode configuration.
I opened https://github.com/lowRISC/opentitan/issues/24597 regarding missing configurations in SPI pass through mode. I guess that someone who knows better SPI will have to refactor this test anyway or create a new test for the missing configurations.

vogelpi:
Thanks @hcallahan-lowrisc for your review, and for looking into this together with @antmarzam . I'll now merge the PR and cherry-pick it over to earlgrey_1.0.0 together with some other changes.

Once the real solution is there, we can also merge and cherry-pick it over.

PR #24605, opened 2024-09-19
[python] fix compatibility for Python 3.12

### LINKED ISSUE #24360: [python] Add support for python 3.12
### Description

Currently, our repository is not totally hermetic when it comes to python. This is because `rules_python` needs to use the system python to install packages using `pip` (and only after that can it use the pinned python version). But due to a combination of using an old version of `rules_python` and some old version python packages, we cannot support python 3.12. The main issue is that 3.12 removed `distutils` on which old packages depend for installation.

Python 3.12 is now becoming the minimum supported version in major distributions like Debian so we should make sure that it works.

The #24216 by @jwnrt enables 3.12 as a side effect of updating all dependencies.

nbdd0121:
Fix #24360

pkg_resources is removed in Python 3.12 and replaced by `importlib.resources`. `importlib.resources` however is available in Python 3.9+, so use a polyfill library instead.

Bump a bunch of dependencies to the version that supports 3.12.


nbdd0121:
Passing CI: https://github.com/lowRISC/opentitan/actions/runs/10998725968/job/30537297981?pr=24605

jwnrt (approved):
LGTM, thanks

doc/contributing/sw/adding_python_depedencies.md
```diff
@@ -4,5 +4,5 @@ To comply with software supply chain security requirements of various project pa
 We accomplish this with the help of the `uv pip compile` command, which is part of the `uv` package.
 
 If you need to add another Python package to the project, do so by:
-1. adding the package and version number to the `pyproject.toml` file, in the form of `<package>==<version>`, and
+1. adding the package and version number to the `pyproject.toml` file, in the form of `<package> ~= <version>` (or `<pacakge> == <version>` in case a specific version is needed), and
```
Razer6, line 7:
```suggestion
1. adding the package and version number to the `pyproject.toml` file, in the form of `<package> ~= <version>` (or `<package> == <version>` in case a specific version is needed), and
```

nbdd0121: Fixed

pamaury (approved):
I am not a python expert but this looks reasonable to me

PR #24606, opened 2024-09-19
Excluded the continuous assignment block of rom_ctrl_scrambled_rom.sv…

KinzaQamar:
… on L81

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
+++ b/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
@@ -18,3 +18,9 @@ Transition Checking->RomAhead "341->917"
 ANNOTATION: "UNR"
 Fsm state_d "1410404563"
 Transition KmacAhead->ReadingHigh "629->181"
+
+// In rom_ctrl_scrambled_rom, the wire scr_nonce is having a fixed value and is a part of RHS of
+// the continuous assignment on line 81 and 82
+INSTANCE: tb.dut.gen_rom_scramble_enabled.u_rom
+ANNOTATION: "Waived the continuous assignment block as RHS is a fixed wire"
+Block 1 "3862690173" "assign data_scr_nonce = scr_nonce[(63 - Aw):0];"


hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -18,3 +18,11 @@ Transition Checking->RomAhead "341->917"
 ANNOTATION: "UNR"
 Fsm state_d "1410404563"
 Transition KmacAhead->ReadingHigh "629->181"
+
+// In rom_ctrl_scrambled_rom, the wire scr_nonce is having a fixed value and is a part of RHS of 
+// the continuous assignment on line 81 and 82  
+CHECKSUM: "2714069379 225975262"
+ANNOTATION: "ModuleName: rom_ctrl_scrambled_rom"
```
rswarbrick, line 25: I think we can probably drop these two lines? (The annotation won't really do anything, and I don't think the checksum is something we need to track: it doesn't match the one earlier, which I suspect shows we should drop both :-D)

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -18,3 +18,11 @@ Transition Checking->RomAhead "341->917"
 ANNOTATION: "UNR"
 Fsm state_d "1410404563"
 Transition KmacAhead->ReadingHigh "629->181"
+
+// In rom_ctrl_scrambled_rom, the wire scr_nonce is having a fixed value and is a part of RHS of
```
rswarbrick, line 22: We chatted earlier, but there are a couple of lines with trailing whitespace which cause the CI failure.

rswarbrick (commented):
Thanks for this: it looks really good!

A couple of nitty changes, but congratulations on your first OpenTitan PR :-)

rswarbrick (commented):
This looks great, thanks!

Two things to do before merging:
1. Squash the two commits together (to avoid the eventual history having them separate). The easiest way is probably to do an interactive rebase and use "fixup" as the first line in the second commit.
2. Tweak the commit message to avoid the overly long line and match the conventional style. Something like this maybe? "[rom_ctrl,dv] Exclude a continuous assignment from line coverage"

rswarbrick (approved):
This looks great, thanks!

PR #24611, opened 2024-09-20
[top_englishbreakfast] Add second UART

vogelpi:
For penetration testing, we've been using a second UART on the CW310 and the silicon platform such that the host can directly communicate with the target using the uJSON protocol instead of looping this communication through the simpleserial channel with the scope (available on Husky only anyways).

This commit now adds a second UART to the English Breakfast top level such that we can re-align the capture setups again. Ultimately, this will allow removing the simpleserial-based communication which will quite notably simplify code maintenance in the future.

For reference, see also https://github.com/lowRISC/ot-sca/issues/235.

### DIFF
--- a/hw/top_englishbreakfast/data/top_englishbreakfast.hjson
+++ b/hw/top_englishbreakfast/data/top_englishbreakfast.hjson
@@ -186,12 +186,27 @@
       // the ip.hjson will declare the clock port names
       // If none are defined at ip.hjson, clk_i is used by default
       clock_srcs: {clk_i: "io_div4"},
-
+      clock_group: "peri",
       // reset connections defines the port to top level reset connection
       // the ip.hjson will declare the reset port names
       // If none are defined at ip.hjson, rst_ni is used by default
-      reset_connections: {rst_ni: "sys_io_div4"},
+      reset_connections: {rst_ni: "lc_io_div4"},
       base_addr: "0x40000000",
+    },
+    { name: "uart1",    // instance name
+      type: "uart",     // Must be matched to the ip name in `ip.hson` (_reg, _cfg permitted)
+                        // and `hw/ip/{type}`
+
+      // clock connections defines the port to top level clock connection
+      // the ip.hjson will declare the clock port names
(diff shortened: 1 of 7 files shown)


vogelpi:
@nasahlpa , @andrea-caforio , I've successfully tested this using simpleserial on the FPGA connected to our shared workstation. To get the uJSON communication working in ot-sca CI, we'll also need to connect an additional FTDI to the IO5 and IO6 pins of the JP3 header. This is because on the CW305, the integrated of UARTs/USARTs of the SAM3U microcontroller are unfortunately not connected to the FPGA.

pamaury (approved):
The software side looks good to me

hw/top_englishbreakfast/data/top_englishbreakfast.hjson
```diff
@@ -186,13 +186,28 @@
       // the ip.hjson will declare the clock port names
       // If none are defined at ip.hjson, clk_i is used by default
       clock_srcs: {clk_i: "io_div4"},
-
+      clock_group: "peri",
       // reset connections defines the port to top level reset connection
       // the ip.hjson will declare the reset port names
       // If none are defined at ip.hjson, rst_ni is used by default
-      reset_connections: {rst_ni: "sys_io_div4"},
+      reset_connections: {rst_ni: "lc_io_div4"},
```
nasahlpa, line 193: Why are we now using `lc_io_div4` instead of `sys_io_div4`?

nasahlpa (approved):
LGTM, just one question.

hw/top_englishbreakfast/data/top_englishbreakfast.hjson
vogelpi: `rst_sys` resets the entire system where as `rst_lc` resets the system excluding rv_dm and a small portion of pinmux.  I noted that the same change was done a while ago for Earlgrey (but then we forgot to update English Breakfast in sync).

andrea-caforio (approved):

vogelpi:
Thanks for your reviews guys!

PR #24618, opened 2024-09-23
[hw,dma,dif] Sync DMA DIF from integrated_dev to Master

Razer6:
This PR syncs the DMA DIF from `integrated_dev` to `master`. 

This PR depends on https://github.com/lowRISC/opentitan/pull/24434


sw/ip/dma/test/utils/dma_testutils.c
```diff
@@ -0,0 +1,91 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/ip/dma/test/utils/dma_testutils.h"
+
+#include "hw/top_darjeeling/sw/autogen/top_darjeeling.h"
```
nbdd0121, line 7: I don't think this exists yet.

sw/ip/dma/dif/BUILD
a-will, line 1: I don't think we want to copy over the bazel targets directly (and possibly not the directory structure either, depending on the software team's strategy for multi-top).

The filegroup targets here were a means to avoid doing multi-top in integrated_dev (by creating top-specific build targets that depended on them). We should create the correct targets for master instead.

sw/ip/dma/test/utils/dma_testutils.c
Razer6: Indeed. I removed the testutils commit until the SW framework for multitop is in place.

Razer6:
@a-will Could you take another look?

sw/device/lib/dif/dif_dma.h
```diff
@@ -0,0 +1,414 @@
...
+#include "dma_regs.h"  // Generated.
+#include "sw/device/lib/dif/autogen/dif_dma_autogen.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+// Target Address space that the source address pointer refers to.
+typedef enum dif_dma_address_space_id {
+  /*OpenTitan 32 bit internal bus.*/
```
pamaury, line 25: nit: the spacing of the comment seems wrong, you can ignore it if it's too annyoing to change. It also happens on the next item.

sw/device/lib/dif/dif_dma.h
```diff
@@ -0,0 +1,414 @@
...
+
+  /*SoC control register address bus using 32 bit (or 64 bits if configured by
+     an SoC) CTN port.*/
+  kDifDmaSoCControlRegisterBus = 0x0a,
+
+  /* SoC system address bus using 64 bit SYS port.*/
+  kDifDmaSoCSystemBus = 0x09,
+} dif_dma_address_space_id_t;
+
+/*Address Space ID*/
```
pamaury, line 36: comment doesn't match the content of the enum

pamaury (approved):
The code looks ok to me, I left a couple of small comments.

Razer6:
@pamaury Thanks for looking into this. I fixed the above mentioned comments.

Issue #20941, opened 2024-01-23
[aes,dv] Randomly drive invalid operation values to fix coverage hole


vogelpi:
### Description

Fixing #20753 (erroneous covergroup sampling function declaration) revealed that there has been coverage hole in AES DV: we were not driving invalid operation values and thus not testing the inherent resolution to the encryption mode in case of an invalid value.

I am not worried at all about this, the logic is super simple an very similar to other config fields where we check the resolution and also have coverage. But we still should solve this for M5.

andreaskurth:
After discussing with @vogelpi, we're deferring this to V3 / M7 because it's about completing coverage under invalid operation

vogelpi:
The PR closing this is now here: #24623.

PR #24623, opened 2024-09-23
[aes/dv] Randomly drive invalid AES operation values

vogelpi:
This PR contains a couple of commits to enable closing a small coverage hole.

The main commit is the last one. It properly randomizes the operation values. Previously, the operation field of the main control register was only written with valid one-hot encoded values but not with invalid values leading to a small DV coverage hole.

This resolves https://github.com/lowRISC/opentitan/issues/20941.

### DIFF
--- a/hw/ip/aes/dv/env/aes_env_cfg.sv
+++ b/hw/ip/aes/dv/env/aes_env_cfg.sv
@@ -60,7 +60,7 @@ class aes_env_cfg extends cip_base_env_cfg #(.RAL_T(aes_reg_block));
   int                key_256b_weight            = 10;
 
   // reseed weigth
-  int                per8_weight                = 60;
+  int                per1_weight                = 60;
   int                per64_weight               = 30;
   int                per8k_weight               = 10;
 
--- a/hw/ip/aes/dv/env/aes_env_cfg.sv
+++ b/hw/ip/aes/dv/env/aes_env_cfg.sv
@@ -76,7 +76,7 @@ class aes_env_cfg extends cip_base_env_cfg #(.RAL_T(aes_reg_block));
   bit                fixed_data_en              = 0;
   // fixed operation
   bit                fixed_operation_en         = 0;
-  bit                fixed_operation            = 1'b0;
+  bit [1:0]          fixed_operation            = AES_ENC;
   // fixed iv (will set all to bits 0)
   bit                fixed_iv_en                = 0;
 
@@ -90,10 +90,14 @@ class aes_env_cfg extends cip_base_env_cfg #(.RAL_T(aes_reg_block));
   //   [3]: inject Lc_escalate
   error_types_t      error_types                = 4'b1111;
 
-  //   [2]: reseed error
-  //   [1]: mode error
-  //   [0]: key_len
-  cfg_error_type_t   config_error_type          = 3'b111;
+  //   [0]: op_error
+  //   [1]: reseed error
+  //   [2]: mode error
+  //   [3]: key_len
+  cfg_error_type_t   config_error_type_en       = '{key_len:  1'b1,
+                                                    mode:     1'b1,
+                                                    rsd_rate: 1'b1,
+                                                    op:       1'b1};
   int                config_error_pct           = 30;
 
   // min and max wait (clk) before an error injection
--- a/hw/ip/aes/dv/env/aes_message_item.sv
+++ b/hw/ip/aes/dv/env/aes_message_item.sv
@@ -68,7 +68,7 @@ class aes_message_item extends uvm_sequence_item;
   int    key_256b_weight      = 10;
 
   // reseed weight
-  int    per8_weight          = 60;
+  int    per1_weight          = 60;
   int    per64_weight         = 30;
   int    per8k_weight         = 10;
 
@@ -143,7 +143,7 @@ class aes_message_item extends uvm_sequence_item;
     solve cfg_error_type before reseed_rate;
     if (!(has_config_error && cfg_error_type.rsd_rate) ) {
       reseed_rate inside { 3'b001, 3'b010, 3'b100 };
-      reseed_rate dist   { 3'b001 :/ per8_weight,
+      reseed_rate dist   { 3'b001 :/ per1_weight,
                            3'b010 :/ per64_weight,
                            3'b100 :/ per8k_weight };
     } else {
--- a/hw/ip/aes/dv/env/aes_message_item.sv
+++ b/hw/ip/aes/dv/env/aes_message_item.sv
@@ -298,53 +298,66 @@ class aes_message_item extends uvm_sequence_item;
   virtual function string convert2string();
     string str;
     str = super.convert2string();
-    str = {str,  $sformatf("\n\t ----| \t\t AES MESSAGE ITEM   \t      |----\t ")
-           };
+    str = {str,  $sformatf("\n\t ----| AES MESSAGE ITEM")};
     str = {str, "\n\t ----| "};
-    str = {str,  $sformatf("\n\t ----| Mode: \t  \t \t %s                         \t ",
-                           aes_mode.name())};
+    str = {str,  $sformatf("\n\t ----| Message Length:       %0d", message_length)};
     str = {str,  $sformatf("\n\t ----| Operation:            %s",
                            aes_operation == AES_ENC ? "AES_ENC" :
                            aes_operation == AES_DEC ? "AES_DEC" : "INVALID")};
-    str = {str,  $sformatf("\n\t ----| has Configuration error:  %s  \t  \t        \t ",
-                           (has_config_error==1) ? "TRUE" : "FALSE")};
-    str = {str,  $sformatf("\n\t ----| Mode error en:  \t %d \n\t ----| Key_len error en: \t %d  \t         \t ",
-                           cfg_error_type.mode, cfg_error_type.key_len)};
-    str = {str,  $sformatf("\n\t ----| Message Length:  \t  \t %d             \t ",
-                           message_length)};
-
-    str = {str,  $sformatf("\n\t ----| Key Length:   \t \t %03b                             \t ",
-                           aes_keylen)};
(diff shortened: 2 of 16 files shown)


vogelpi:
Thanks @martin-velay for your help with debugging this :-)

hw/ip/aes/dv/env/aes_message_item.sv
```diff
@@ -119,6 +122,28 @@ class aes_message_item extends uvm_sequence_item;
...
+    } else {
+      has_config_error == 0;
+    }
+  }
+
+  constraint config_error_type_c {
+    solve has_config_error before cfg_error_type;
+    if (has_config_error) {
+      cfg_error_type inside {[1:15]};
+      config_error_type_en[0] == 0 -> cfg_error_type[0] == 0;
```
rswarbrick, line 138: Can't this be expressed with names? So something like `config_error_type_en.op == 0 -> cfg_error_type.op == 0`?

Even more tersely, I think you can do something like `!(cfg_error_type & ~config_error_type_en)`, but that's probably getting ridiculously hard to understand...

FOLLOW-UP: Ah! I've just seen the lines this came from! I think I stand by my comments, but they needn't gate this PR.

hw/ip/aes/dv/env/aes_message_item.sv
```diff
@@ -151,6 +192,20 @@ class aes_message_item extends uvm_sequence_item;
...
+    solve cfg_error_type before aes_operation;
+    if (!(has_config_error && cfg_error_type.op) ) {
+      aes_operation inside { 2'b01, 2'b10 };
+      aes_operation dist   { 2'b01 :/ 1,
+                             2'b10 :/ 1 };
+    } else {
+      !( aes_operation inside { 2'b01, 2'b10 });
+    }
+    if (fixed_operation_en) {
+      aes_operation == fixed_operation
```
rswarbrick, line 206: What happens if `fixed_operation_en` is true , `fixed_operation` is a valid operation, `has_config_error` is true and `cfg_error_type.op` is true?

I *think* this line will "beat" the previous ones in `operation_c` and the tool will merrily generate a valid operation. Presumably you don't really care about this situation, but maybe it makes sense to leave a note about what happens?

hw/ip/aes/dv/env/aes_message_item.sv
```diff
@@ -252,6 +260,14 @@ class aes_message_item extends uvm_sequence_item;
     // Check for invalid configuration values and resolve them if necessary. Illegal mode values
     // don't need to be handled here as they don't result in the DUT actually producing output
     // data.
+    if (item.operation inside {2'b01, 2'b10}) begin
+      this.aes_operation = item.operation;
+    end else begin
+      this.aes_operation = 2'b01; // resovle to AES-ENC
+      `uvm_info(`gfn,
+          $sformatf("\n\t ---| Illegal operation value detected, resolving to default AES-ENC"),
```
rswarbrick, line 268: Nit: I think you probably want "... detected. Resolving ..." to fix the grammar.

hw/ip/aes/dv/env/aes_message_item.sv
```diff
@@ -247,14 +247,26 @@ class aes_message_item extends uvm_sequence_item;
...
     end else begin
-      this.aes_keylen = 3'b100; // force to 256b
-      `uvm_info(`gfn, $sformatf("\n\t ---| Illegal key len detected reverting to default 256"),
-                UVM_MEDIUM)
+      this.aes_keylen = 3'b100; // resolve to AES-256
+      `uvm_info(`gfn,
+          $sformatf("\n\t ---| Illegal key length value detected, resolving to default AES-256"),
+          UVM_MEDIUM)
+    end
+    if (item.reseed_rate inside {3'b001, 3'b010, 3'b100}) begin
```
martin-velay, line 279: That's not really related to this PR, but in general I think it would be better to move to enumerated values. This would make reading easier. But it's not blocking for this PR of course.

hw/ip/aes/dv/env/aes_message_item.sv
```diff
@@ -281,53 +281,67 @@ class aes_message_item extends uvm_sequence_item;
...
-    str = {str,  $sformatf("\n\t ----| 192 Weight: %d       \t \t ", key_192b_weight)};
-    str = {str,  $sformatf("\n\t ----| 256 Weight: %d       \t \t ", key_256b_weight)};
+    str = {str,  $sformatf("\n\t ----| Sideload:             %0b", sideload_en)};
+    str = {str,  $sformatf("\n\t ----| Manual Operation:     %0b", manual_operation)};
     str = {str,  $sformatf("\n\t")};
-
     return str;
   endfunction // conver2string
 
+  function string cfg_error_string();
```
martin-velay, line 327: As the sequence item is not properly split originally, I agree with your change. But FYI, this [convert2string function is defined by the UVM](https://verificationacademy.com/verification-methodology-reference/uvm/docs_1.1c/html/files/base/uvm_object-svh.html#uvm_object.convert2string). That's convenient when you want to debug a sequence from anybody, you know that you can call this and print all the relevant info in a relevant format. In the future, we should break down the sequence items into sub-items and eventually gathering them into another sequence item.

martin-velay (approved):
Your changes are looking good, this PR can be approved according to me.

Note: the DV seems to be complicated to debug because of some bad practices, particularly regarding the sequence items which are gathering more information than it should. These objects should be broken up into multiple sequence items in the future. And also using enumerated values for some variables might help.

hw/ip/aes/dv/env/seq_lib/aes_base_vseq.sv
```diff
@@ -1039,8 +1039,12 @@ class aes_base_vseq extends cip_base_vseq #(
       end
       `downcast(cloned_message, aes_message.clone());
       message_queue.push_front(cloned_message);
-      `uvm_info(`gfn, $sformatf("\n\t ----| MESSAGE # %d \n %s",i, cloned_message.convert2string())
-               , UVM_MEDIUM)
+      `uvm_info(`gfn, $sformatf("\n\t ----| MESSAGE #%0d\n %s",
```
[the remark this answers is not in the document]
vogelpi: That is certainly not ideal but it was intentional. But after rethinking this again I realized that this is not necessary at all. It's changed now.

hw/ip/aes/dv/env/aes_env_cfg.sv
```diff
@@ -90,10 +90,11 @@ class aes_env_cfg extends cip_base_env_cfg #(.RAL_T(aes_reg_block));
...
 
-  //   [2]: reseed error
-  //   [1]: mode error
-  //   [0]: key_len
-  cfg_error_type_t   config_error_type          = 3'b111;
+  //   [0]: op_error
+  //   [1]: reseed error
+  //   [2]: mode error
+  //   [3]: key_len
+  cfg_error_type_t   config_error_type_en       = 4'b1111;
```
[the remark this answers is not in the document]
vogelpi: Good idea! I've implemented the change for lines / types touches in this PR.

hw/ip/aes/dv/env/aes_message_item.sv
vogelpi: I went with the first proposal to be consistent and improve readability.

hw/ip/aes/dv/env/aes_message_item.sv
vogelpi: This is mainly intended for DUT debugging and we don't use it in any of the sequences. I've left a comment to document this.

hw/ip/aes/dv/env/aes_message_item.sv
vogelpi: Yep, thanks!

hw/ip/aes/dv/env/aes_message_item.sv
vogelpi: I totally agree! I've changed this now for lines / member variables touches in this PR.

hw/ip/aes/dv/env/aes_message_item.sv
vogelpi: I agree with you. Thanks for providing some background on this Martin.

vogelpi:
Thanks @martin-velay and @rswarbrick for your review and the suggestions! I've implemented most of your feedback (see discussions). If you're happy with the changes, please feel free to approve & merge this @rswarbrick .

hw/ip/aes/dv/env/aes_env_cfg.sv
```diff
@@ -90,10 +90,14 @@ class aes_env_cfg extends cip_base_env_cfg #(.RAL_T(aes_reg_block));
...
   error_types_t      error_types                = 4'b1111;
 
-  //   [2]: reseed error
-  //   [1]: mode error
-  //   [0]: key_len
-  cfg_error_type_t   config_error_type          = 3'b111;
+  //   [0]: op_error
+  //   [1]: reseed error
+  //   [2]: mode error
+  //   [3]: key_len
```
rswarbrick, line 96: Nit: I don't think you need the comment about the indices any more?

Similarly for `cfg_error_type` below.

hw/ip/aes/dv/env/aes_seq_item.sv
```diff
@@ -7,7 +7,7 @@ class aes_seq_item extends uvm_sequence_item;
   `uvm_object_utils(aes_seq_item)
 
   aes_item_type_e  item_type;
-  aes_op_e operation;
+  bit [1:0] operation;
```
rswarbrick, line 10: Nit: I know that this is unchanged with this commit, but the indentation looks a bit weird with the previous line. Maybe we should either be dropping a space on line 9 or adding some more spaces before "operation"?

rswarbrick (approved):
This looks good to me. A couple of (rather nitty) tweaks that might be worth doing, but I'm happy either way.

vogelpi:
Thanks @rswarbrick for your approval. I am merging this now as I need to move on.

PR #24625, opened 2024-09-23
[top_darjeeling,lint,sim] Add lint and sim cfg for Darjeeling

Razer6:
This PR adds the block level lint and sim configuration for the Darjeeling top. It only adds the IPs that are used in Darjeeling, ignoring other IPs that are used in Earlgrey.

The PR enables the DMA, MBX, KeymgrDPE, soc_proxy IP.

### DIFF
--- a/hw/top_darjeeling/lint/top_darjeeling_dv_lint_cfgs.hjson
+++ b/hw/top_darjeeling/lint/top_darjeeling_dv_lint_cfgs.hjson
@@ -1,0 +1,220 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+{
+
+  // This is the primary cfg hjson for DV linting. It imports ALL individual lint
+  // cfgs of the IPs DV environments and the full chip DV environment for top_darjeeling.
+  // This enables to run them all as a regression in one shot.
+  name: top_darjeeling_dv_batch
+
+  import_cfgs:      [// common server configuration for results upload
+                     "{proj_root}/hw/data/common_project_cfg.hjson"
+                     // tool-specific configuration
+                     "{proj_root}/hw/lint/tools/dvsim/{tool}.hjson"]
+
+  flow: "lint"
+
+  // Different dashboard output path for each tool
+  rel_path: "hw/top_darjeeling/dv/lint/{tool}/summary"
+
+  use_cfgs: [{    name: dma
+                  fusesoc_core: lowrisc:ip:dma_sim
+                  import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+                  rel_path: "hw/ip/dma/lint/{tool}"
+             },
+             {    name: mbx
+                  fusesoc_core: lowrisc:ip:mbx_sim
+                  import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+                  rel_path: "hw/ip/mbx/lint/{tool}"
+             },
+             {    name: aes
+                  fusesoc_core: lowrisc:dv:aes_sim
+                  import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+                  rel_path: "hw/ip/aes/dv/lint/{tool}"
+             },
+          //    {    name: alert_handler
+          //         fusesoc_core: lowrisc:opentitan:top_darjeeling_alert_handler_sim
+          //         import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+          //         rel_path: "hw/top_darjeeling/ip_autogen/alert_handler/dv/lint/{tool}"
+          //    },
+          //    {    name: aon_timer
+          //         fusesoc_core: lowrisc:dv:aon_timer_sim
+          //         import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+          //         rel_path: "hw/top_darjeeling/ip/aon_timer/dv/lint/{tool}"
+          //    },
+          //    {    name: clkmgr
+          //         fusesoc_core: lowrisc:dv:clkmgr_sim
+          //         import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+          //         rel_path: "hw/top_darjeeling/ip_autogen/clkmgr/dv/lint/{tool}"
+          //    },
+             {    name: csrng
+                  fusesoc_core: lowrisc:dv:csrng_sim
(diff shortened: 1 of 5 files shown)


Razer6:
I see that those are not yet part of CI - should we already enable the configs for the other IP?

hw/top_darjeeling/lint/top_darjeeling_dv_lint_cfgs.hjson
```diff
@@ -0,0 +1,220 @@
...
+
+  // Different dashboard output path for each tool
+  rel_path: "hw/top_darjeeling/dv/lint/{tool}/summary"
+
+  use_cfgs: [{    name: dma
+                  fusesoc_core: lowrisc:ip:dma_sim
+                  import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+                  rel_path: "hw/ip/dma/lint/{tool}"
+             },
+          //    {    name: mbx
```
rswarbrick, line 26: Would it make sense to open an issue about the comment characters in this list so that we can't accidentally forget to turn on these lint configurations once we've merged the blocks / turned on the top-level autogen ones?

rswarbrick:
About the CI question: yes, probably :-) At the moment, the CI version of the lint flow runs through `ci/scripts/verible-lint.sh`, which just knows about Earlgrey. I think the way to do it properly would be to add an extra "toplevel" argument to the script and then use that when constructing the hjson paths. At the moment, we run the script three times (RTL, DV, FPV) in `slow-lint.sh`. Splitting it out like that, we'd have to bump that up to three times for each top-level.

hw/top_darjeeling/lint/top_darjeeling_fpv_lint_cfgs.hjson
```diff
@@ -0,0 +1,153 @@
...
+              import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+              rel_path: "hw/ip/alert_handler/alert_handler_esc_timer_fpv/lint/{tool}"
+             }
+             {
+              name: alert_handler_ping_timer_fpv
+              fusesoc_core: lowrisc:opentitan:top_darjeeling_alert_handler_ping_timer_fpv
+              import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+              rel_path: "hw/ip/alert_handler/alert_handler_ping_timer_fpv/lint/{tool}"
+             }
+            {
```
rswarbrick, line 32: Nit: Indentation is off here, I think.

hw/top_darjeeling/lint/top_darjeeling_fpv_lint_cfgs.hjson
```diff
@@ -0,0 +1,153 @@
...
+                     "{proj_root}/hw/data/common_project_cfg.hjson"
+                     // tool-specific configuration
+                     "{proj_root}/hw/lint/tools/dvsim/{tool}.hjson"]
+
+  // Different dashboard output path for each tool
+  rel_path: "hw/top_darjeeling/fpv/lint/{tool}/summary"
+
+  use_cfgs: [{
+              name: alert_handler_esc_timer_fpv
+              fusesoc_core: lowrisc:opentitan:top_darjeeling_alert_handler_esc_timer_fpv
```
rswarbrick, line 22: Hmm: do we have any way to do this yet? I think the Earlgrey version comes from `hw/top_earlgrey/ip_autogen/alert_handler/fpv`. I think this will probably need commenting until the Darjeeling equivalent lands?

Similarly for the ping timer.

hw/top_darjeeling/lint/top_darjeeling_fpv_lint_cfgs.hjson
```diff
@@ -0,0 +1,153 @@
...
+              import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+              rel_path: "hw/ip/alert_handler/alert_handler_ping_timer_fpv/lint/{tool}"
+             }
+            {
+               name: prim_arbiter_ppc_fpv
+               fusesoc_core: lowrisc:fpv:prim_arbiter_ppc_fpv
+               import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+               rel_path: "hw/ip/prim/prim_arbiter_ppc_fpv/lint/{tool}"
+             }
+             {
```
rswarbrick, line 38: This seems a little silly to me. All the blocks below are identical between the top-levels (I think) and they probably shouldn't have been put in a file with "top_earlgrey" in the name before. Can we split things out into some "shared_fpv_lint_cfgs" list? (Maybe as a separate PR that lands before this one)

hw/top_darjeeling/lint/top_darjeeling_lint_cfgs.hjson
```diff
@@ -0,0 +1,324 @@
...
+                 fusesoc_core: lowrisc:ip:dma
+                 import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+                 rel_path: "hw/ip/dma/lint/{tool}"
+            },
+            {     name: aes
+                  fusesoc_core: lowrisc:ip:aes
+                  import_cfgs: ["{proj_root}/hw/lint/tools/dvsim/common_lint_cfg.hjson"]
+                  rel_path: "hw/ip/aes/lint/{tool}"
+             },
+             {    name: alert_handler
```
rswarbrick, line 40: I don't think this can land until the corresponding autogenerated IP is in, can it? Maybe this item needs commenting? Similarly for quite a few other blocks below.

hw/top_darjeeling/lint/top_darjeeling_fpv_lint_cfgs.hjson
Razer6: Would adding that to the `import_cfgs` section work?

rswarbrick: I genuinely don't know! (And if it doesn't work, that might be a bug we should fix in dvsim...)

hw/top_darjeeling/lint/top_darjeeling_dv_lint_cfgs.hjson
Razer6: Done in https://github.com/lowRISC/opentitan/issues/24737

rswarbrick (approved):
This looks sensible to me, and thanks for splitting things out like this.

rswarbrick:
Oh, actually it looks like something has gone slightly wrong with the includes here:
image

Razer6:
@rswarbrick I'm not sure if nested configs are supported by the flow (nested in the sense of more than one child config)

rswarbrick:
Oh, that's a pity. But the change in the latest version (doing the expansion manually) looks like a reasonable approach for now, at least. Thanks.

Razer6:
Yep. Wanted to get the lint in, to unblock the other PRs.

Razer6:
@rswarbrick CI is green - can we get that merged?

PR #24627, opened 2024-09-23
[dv/dpi] Fix TCP server hang when ring buffer fills up

### LINKED ISSUE #24616: [dv/dpi] TCP server enters infinite loop if ring buffer is full
### Description

The `rptr` and `wptr` variables [here](https://github.com/lowRISC/opentitan/blob/ae8ac572e793f34f457d08c9ce03db3644c49f5a/hw/dv/dpi/common/tcp_server/tcp_server.c#L32) are missing `volatile`, so GCC will optimize the while loop in [tcp_server_put_byte](https://github.com/lowRISC/opentitan/blob/ae8ac572e793f34f457d08c9ce03db3644c49f5a/hw/dv/dpi/common/tcp_server/tcp_server.c#L65) to a branch and an infinite loop on -O2. as you can see [here](https://godbolt.org/z/czMK9qPd5).

duk-37:
These buffers are expected to be manipulated by multiple threads in these functions, so we need to mark them as volatile. GCC can and will optimize away things we don't want it to otherwise.

Closes #24616

### DIFF
--- a/hw/dv/dpi/common/tcp_server/tcp_server.c
+++ b/hw/dv/dpi/common/tcp_server/tcp_server.c
@@ -50,7 +50,7 @@ struct tcp_server_ctx {
   pthread_t sock_thread;
 };
 
-static bool tcp_buffer_is_full(struct tcp_buf *buf) {
+static bool tcp_buffer_is_full(volatile struct tcp_buf *buf) {
   if (buf->wptr >= buf->rptr) {
     return (buf->wptr - buf->rptr) == (BUFSIZE_BYTE - 1);
   } else {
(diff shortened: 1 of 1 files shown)


luismarques:
I don't think using `volatile` here properly addresses the underlying issue.

The `rptr`, `wptr`, and `buf` need to have consistent values, for all threads. Your use of `volatile`, combined with x86's strong memory model, happens to make it work, but it's not guaranteed. To understand that, consider the scenario where thread A puts data in (mutates `buf` and `wptr`) and thread B sees the write to `wptr` but not yet to `buf` (due to allowed memory reorderings between threads).

Probably the easiest way to fix this is to use a lock to synchronize accesses to the overall `struct tcp_buf`. Using atomics and fences might provide higher performance but unless this is particularly performance-sensitive I'd avoid the extra complexity.

duk-37:
> I don't think using `volatile` here properly addresses the underlying issue.
> 
> The `rptr`, `wptr`, and `buf` need to have consistent values, for all threads. Your use of `volatile`, combined with x86's strong memory model, happens to make it work, but it's not guaranteed. To understand that, consider the scenario where thread A puts data in (mutates `buf` and `wptr`) and thread B sees the write to `wptr` but not yet to `buf` (due to allowed memory reorderings between threads).

I mean, yes, this implementation as it stands is broken on any architecture with weak memory models to begin with. Using a blanket lock would probably not here since there's quite a bit of busy-looping going on. re: performance, the reason I found this to begin with was because the buffer _was_ filled before it could be emptied, which stalled JTAG emulation completely at some point.

Some combination of mutexes and a condition flag would probably work, but I'm currently having issues compiling on a home machine due to (I think) missing `vivado` tags on some tests, namely `hw/bitstream/universal`

PR #24630, opened 2024-09-24
[manuf] erase AST config data from flash after use

### LINKED ISSUE #24526: [manuf] wipe AST config data from flash info page 0 after it is committed to OTP
During CP, the AST configuration data is written to flash info page 0. During FT, the data is copied from flash to OTP, here. However, after copying the data to OTP, it should also be wiped from flash.

timothytrippel:
This updates the individualization firmware to erase the AST configuration data from flash info after committing it to OTP.

This fixes #24526.


timothytrippel:
CC: @benjbender

moidx (approved):

PR #24632, opened 2024-09-24
[sw/tests] Add OTP descrambling test

andreaskurth:
This adds a top-level test for the descrambling functionality/datapath of `otp_ctrl`.  All checks are done in SW; there's no need for any checks on the DV side.  The test works by reading out all three SECRET OTP partitions from SW (which is possible before they get locked) and confirm that the descrambled value that SW reads from OTP equals the expected value that is stored in the OTP image (the OTP image itself is scrambled).

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -1983,6 +1983,71 @@ opentitan_test(
     ],
 )
 
+otp_json(
+    name = "otp_ctrl_descrambling_otp_json",
+    partitions = [
+        otp_partition(
+            name = "SECRET0",
+            items = {
+                "TEST_UNLOCK_TOKEN": "117dbd7aca3f7de6749f1cb4a7daf4f5",
+                "TEST_EXIT_TOKEN": "1dc4b32c25ef2c31d6313735da1740e4",
+            },
+            lock = False,
+        ),
+        otp_partition(
+            name = "SECRET1",
+            items = {
+                "FLASH_ADDR_KEY_SEED": "434677540fdc9956038c1b4e7bd7394c8251c8921ed1a7fc8ed040844dde2dca",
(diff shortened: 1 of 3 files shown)


sw/device/tests/BUILD
```diff
@@ -1983,6 +1983,63 @@ opentitan_test(
...
+)
+
+opentitan_test(
+    name = "otp_ctrl_descrambling_test",
+    srcs = ["otp_ctrl_descrambling_test.c"],
+    dv = dv_params(
+        otp = ":otp_ctrl_descrambling_otp_image",
+    ),
+    exec_env = dicts.add(
+        EARLGREY_TEST_ENVS,
```
moidx, line 2032: nit: I would just target sim environments and maybe up to one FPGA target.

moidx (approved):

sw/device/tests/BUILD
andreaskurth: Thx, should I just write
```
"//hw/top_earlgrey:fpga_cw310_sival": None, 
"//hw/top_earlgrey:sim_dv": None,
"//hw/top_earlgrey:sim_verilator": None,
```
?

The OTP image has RMA as LC state and custom values in the SECRET* partitions; would this work with all of those targets?

andreaskurth: From CI results, it seems the test fails in _CW310 ROM Tests_ and _CW310 SiVal ROM_EXT Tests_

moidx: You probably need the following

```python
fpga = fpga_params(
  otp = ":otp_ctrl_descrambling_otp_image",
),
```

Try the `fpga_cw310_test_rom` target instead of the `fpga_cw310_sival` one. This is because we don't expect this test to run on a silicon configuration.

PR #24634, opened 2024-09-24
[rom_ext] Use the xmodem long timeout

cfrantz:
I have observed serveral test flakes at Xmodem packet boundaries (e.g. kErrorXModemTimeoutStart).  Always use the long timeout at packet boundaries.


cfrantz:
Committed as part of #24629.

PR #24676, opened 2024-09-26
[rom_ctrl,dv] Fixed rom_ctrl_kmac_err_chk_fix failure

KinzaQamar:
This PR fixes the `rom_ctrl_kmac_err_chk_fix` test failure.

rom_ctrl_scoreboard happens to fails the [pwrmgr_complete](https://github.com/lowRISC/opentitan/blob/ec74fd4b3fff1d9764cf1956c9f75057c00660e7/hw/ip/rom_ctrl/dv/env/rom_ctrl_scoreboard.sv#L254C1-L254C74) check a cycle earlier when the rom_ctrl_fsm is not in a done state. rom_ctrl_fsm was still processing through its post-reset flow and state registers after virtual sequence ends.

The fix here ensures that the run_phase doesn't complete until the [pwrmgr_complete](https://github.com/lowRISC/opentitan/blob/ec74fd4b3fff1d9764cf1956c9f75057c00660e7/hw/ip/rom_ctrl/dv/env/rom_ctrl_scoreboard.sv#L254C1-L254C74)  has happened.

### DIFF
--- a/hw/ip/rom_ctrl/dv/env/rom_ctrl_scoreboard.sv
+++ b/hw/ip/rom_ctrl/dv/env/rom_ctrl_scoreboard.sv
@@ -274,4 +274,20 @@ class rom_ctrl_scoreboard extends cip_base_scoreboard #(
     end
   endfunction
 
+  virtual function void phase_ready_to_end(uvm_phase phase);
+    if (phase.get_name() != "run") return;
+
+      // Raising an objection and waiting for the pwrmgr_complete to set. This will add an extra
+      // delay after the test finishes and rom_ctrl_fsm would be in a done state which will set
+      // pwrmgr_data_o.done.
+      phase.raise_objection(this, {`gfn, " objection raised"});
+      fork
+        begin
+          if (cfg.en_scb)
+            wait (pwrmgr_complete);
+          phase.drop_objection(this, {`gfn, " objection dropped"});
+        end
+      join_none
+  endfunction
+
 endclass


hw/ip/rom_ctrl/dv/env/rom_ctrl_scoreboard.sv
```diff
@@ -274,4 +274,18 @@ class rom_ctrl_scoreboard extends cip_base_scoreboard #(
     end
   endfunction
 
+  virtual function void phase_ready_to_end(uvm_phase phase);
+    if (phase.get_name() != "run") return;
+
```
rswarbrick, line 279: The CI check is failing because there is a tab here: I think we can probably drop the whitespace entirely.

hw/ip/rom_ctrl/dv/env/rom_ctrl_scoreboard.sv
```diff
@@ -274,4 +274,18 @@ class rom_ctrl_scoreboard extends cip_base_scoreboard #(
     end
   endfunction
 
+  virtual function void phase_ready_to_end(uvm_phase phase);
+    if (phase.get_name() != "run") return;
+	
+      // Raising an objection and waiting for the pwrmgr_complete to set. This will add an extra delay after the
+      // test finishes and rom_ctrl_fsm would be in a done state which will set pwrmgr_data_o.done.
```
rswarbrick, line 281: Although we don't check this in CI (it seems!), we try to keep to a 100 column width in SystemVerilog, so this needs wrapping. If it avoids you having to do it manually, I think it comes out as
```systemverilog
      // Raising an objection and waiting for the pwrmgr_complete to set. This will add an extra
      // delay after the test finishes and rom_ctrl_fsm would be in a done state which will set
      // pwrmgr_data_o.done.
```

rswarbrick (commented):
This looks good, thanks. There are a couple of nitty lint changes needed (I've added comments). Also, can you extend the commit message to explain what's going on? I'd probably go for something like "The rom_ctrl_kmac_err_chk test fails occasionally because the virtual sequence ends before the dut has run through it's post-reset flow. This causes a check in the scoreboard to fail (which is making sure that a test hasn't finished before it checks anything). Adding this objection ensures that the run phase doesn't complete until the check has happened."

rswarbrick (approved):
Looks good to me. Thanks!

PR #24679, opened 2024-09-26
[tlul,rtl] Modify readback check error reporting logic

### LINKED ISSUE #24698: [sram_ctrl] Fix alert signaling for Readback feature
### Description

While taking care of #23322, @nasahlpa discovered that under some circumstances, readback errors don't get signaled as alerts, meaning they get remain unnoticed which is bad.

nasahlpa:
Currently, only a readback error check is reported when the `error_i` signal is not raised. However, as this signal is also raised when there is no activity on the TL-UL bus, we might miss some readback failures.

This commit modifies the error reporting logic to also report the readback mismatch when the error signal is raised.

Closes #24698.

### DIFF
--- a/hw/ip/tlul/rtl/tlul_sram_byte.sv
+++ b/hw/ip/tlul/rtl/tlul_sram_byte.sv
@@ -262,8 +262,8 @@ module tlul_sram_byte import tlul_pkg::*; #(
             rdback_wait    = 1'b1;
             rdback_check_d = MuBi4False;
 
-            // Perform the readback check. Omit the check if the transaction contains an error.
-            if (!rdback_chk_ok && !error_i) begin
+            // Perform the readback check.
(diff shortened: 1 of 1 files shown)


vogelpi:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_sram_byte.sv

This is a confirmed ECO which is okay.

vogelpi (approved):
Thanks for flagging and fixing this @nasahlpa , great work!

nasahlpa:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_sram_byte.sv

Confirmed ECO.

moidx (approved):

moidx:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_sram_byte.sv

Confirmed ECO.

Issue #24641, opened 2024-09-25
[rom_ext] Migrate to ECDSA verification of application firmware


cfrantz:

cfrantz:
#24544 
#24643

PR #24684, opened 2024-09-27
[rom_ext] Migrate ProdA users to ECDSA

cfrantz:
1. Generate owner and application keys for the `ProdA` user in the CloudKMS instance.  Export the public components of those keys to the codebase (see `sw/device/silicon_creator/rom_ext/proda/keys/README.md`).
2. Add the keys to the appropriate exec_envs.

Addresses: #24641


hw/top_earlgrey/BUILD
```diff
@@ -433,15 +433,23 @@ silicon(
     name = "silicon_owner_proda_rom_ext",
     testonly = True,
     base = ":silicon_owner_sival_rom_ext",
-    ecdsa_key = CLEAR_KEY_SET,
+    # TODO(#24641): Simplify key selection after eliminating RSA keys.
+    ecdsa_key = select({
+        "//signing:owner_key_ecdsa": {"//sw/device/silicon_creator/rom_ext/prodc/keys:keyset": "appkey_test_0"},
+        "//signing:owner_key_rsa": CLEAR_KEY_SET,
+        "//signing:test_keys_ecdsa": {"//sw/device/silicon_creator/lib/ownership/keys/fake:app_test_ecdsa": "test_key_0"},
+        "//signing:test_keys_rsa": {"//sw/device/silicon_creator/lib/ownership/keys/fake:app_test_ecdsa": "test_key_0"},
```
timothytrippel, line 441: Is this use of RSA here just a bridge to migrate over to ECDSA fully?

timothytrippel (approved):

hw/top_earlgrey/BUILD
cfrantz: Its because there is no RSA test key (ie: fake key) if you've selected `--//signing:owner_key_type=rsa`, there is only the RSA key provided by the token.

Its arguable that trying to do an RSA sign with test keys is an error, but it seemed like I should just sign with the ECDSA key anyway.  All this mess goes away when we finally eliminate RSA signature verification, and looks more like:

```
ecdsa_key = select({
  "//signing:test_keys": {"path to fake key": "foo"},
  "//conditions:default": {"label of keyset": "pkcs11 label of key"},
}),
rsa_key = CLEAR_KEY_SET,
```

PR #24688, opened 2024-09-27
[pinmux/dif] Allow WARL behavior for pad attribute registers

vogelpi:
The MIO|DIO_PAD_ATTR registers are specified as WARL meaning certain bits written by software may get discarded by the hardware. For example the supported number of drive strength bits can depend on the hardware platform (simulation, FPGA, silicon).

Without this commit, the DIF treated these registers as RW: After writing pad attributes, it would perform a read back and error out in case of mismatches.

This is related to lowRISC/OpenTitan#24577.

---

The pinmux spec is probably a bit ambiguous with respect to the WARL/RW behavior of these registers:
Screenshot from 2024-09-27 23-25-12

Screenshot from 2024-09-27 23-25-35

I think the main issue is that the regtool does not support WARL inherently. But the registers are really WARL, depending on the hardware/simulation platform and pad implementation, a different number of drive strength bits can be accepted while unsupported bits are tied to zero.

### DIFF
--- a/sw/device/lib/dif/dif_pinmux.c
+++ b/sw/device/lib/dif/dif_pinmux.c
@@ -304,6 +304,12 @@ dif_result_t dif_pinmux_pad_write_attrs(const dif_pinmux_t *pinmux,
   uint32_t read_value = mmio_region_read32(pinmux->base_addr, reg_offset);
   *attrs_out = dif_pinmux_reg_to_pad_attr(read_value);
 
+  // Not all pads implement all attributes and not all target platforms support
+  // all attribute values. The underlying hardware registers implement Write-Any
+  // Read-Legal (WARL) semantics. If the specified attribute values are not
+  // supported by the hardware, return `kDifError`. The caller then needs to
+  // decide on how to resolve the situation. Unsupported attribute values can be
+  // identified by comparing `attrs_in` to `attrs_out`.
   if (reg_value != read_value) {
     return kDifError;
   }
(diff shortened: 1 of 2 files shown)


timothytrippel (approved):

a-will:
I'm not sure I really agree with the analysis here. I think attempting to write values that aren't possible _should_ be flagged as an error. It means the software is asking for something illegal, and this may have consequences for whether the intended target could ever work. Instead, I think the _caller_ ought to decide whether to ignore the error.

Don't blindly use CHECK_DIF_OK() if the error handling needs to be nuanced.

pamaury:
I am not sure what the right behaviour here is, but I am pretty sure this change this will break code such as:
https://github.com/lowRISC/opentitan/blob/47e16dffe65b8d505c877f1304efe80e6fab4b07/sw/device/lib/testing/sysrst_ctrl_testutils.c#L30

sha-ron:
As I understand it the "correct" solution is to update https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim_generic/rtl/prim_generic_pad_attr.sv to match the spec that has no limitation on drive_strength/slew_rate values. If for now we want a bypass we can the change suggested in this PR (assuming it won't break other tests) or I can put the code under `ifdef GATE_LEVEL or something like that.

vogelpi:
Thanks a lot for your input @a-will and also @pamaury ! This information is very helpful. I will propose a fix for the tests in the original PR (as suggested by analyzing the returned value and taking an appropriate action). This PR here, I will adapt to not change the behavior of the DIF but to document a bit what software should do in case the check fails.

vogelpi:
Update: I've now changed the PR to not alter the behavior of the DIF but to add some documentation to the DIF regarding why the check is done and that higher-level software needs to take care of the situation.

sha-ron (approved):

PR #24695, opened 2024-09-30
[ci] Move Verible lint action into CI workflow

jwnrt:


.github/workflows/ci.yml
```diff
@@ -63,3 +63,34 @@ jobs:
         run: ./util/prep-bazel-airgapped-build.sh
       - name: Build in the airgapped environment
         run: ./ci/scripts/test-airgapped-build.sh
+
+  verible-lint:
+    name: Verible lint
+    runs-on: ubuntu-20.04
```
nbdd0121, line 69: I'd prefer we don't downgrade OS. The old action runs on ubuntu 24.04 (latest) so please keep it that way.

jwnrt: Ah, sure. Would you prefer `-latest` or `-24.04`?

nbdd0121: Hmm, actually the `-latest` upgrade to 24 is still being rolled out: https://github.com/actions/runner-images/issues/10636.

Let's try ubuntu-24.04 first, if it doesn't work then we can use ubuntu-22.04.

jwnrt: Yep, works fine on `ubuntu-24.04`

nbdd0121 (approved):

PR #24696, opened 2024-09-30
[ci] Move in-depth lint to GitHub Actions

jwnrt:
This PR also includes a fix to the `install-deps` action for Verilator and Verible permissions.

It moves the "countermeasure implementation" check for `englishbreakfast` into the in-depth lint job for consistency.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -243,10 +243,6 @@ jobs:
       verilator --version
       verible-verilog-lint --version
     displayName: Display environment
-  # Check that all englishbreakfast countermeasures are implemented.
-  - bash: ci/scripts/check-countermeasures.sh englishbreakfast
-    displayName: Check countermeasures for top_englishbreakfast
-    continueOnError: True
   - bash: ci/scripts/build-chip-verilator.sh englishbreakfast
     displayName: Build simulation with Verilator
   - template: ci/upload-artifacts-template.yml
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -125,57 +125,6 @@ jobs:
     condition: eq(variables['Build.Reason'], 'PullRequest')
     displayName: Confirm no .bazelrc-site files
 
-- job: slow_lints
-  displayName: Quality (in-depth lint)
-  # Run code quality checks (in-depth lint)
-  dependsOn: lint
-  pool:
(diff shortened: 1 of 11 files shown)


jwnrt:
Sorry for the force-push spam. There was a permission problem with the Verilator and Verible installations that wasn't obvious outside of CI. Fixed now.

azure-pipelines.yml
```diff
@@ -125,57 +125,6 @@ jobs:
...
-  - bash: ci/scripts/check-countermeasures.sh earlgrey
-    displayName: Check countermeasures for top_earlgrey
-    continueOnError: True
-  # Bazel test suites are a common cause of problematic tags. Check test suites
-  # before checking for other tag issues.
-  - bash:  ci/scripts/check_bazel_test_suites.py
-    displayName: Check Bazel test suites (Experimental)
-    continueOnError: True
-  # #21973: Disabled until Verilator tags are fixed in our build tree.
-  #- bash: ci/scripts/check-bazel-tags.sh
```
nbdd0121, line 147: This is completely removed?

jwnrt: Oops, yes I did remove it. I'll add it back but commented out.

nbdd0121:
Can you check if scripts have used `##vso[]`, and if so, replace it with GitHub equivalent?

machshev (approved):

ci/scripts/verible-lint.sh
```diff
@@ -47,7 +47,7 @@ fi
 
 env DVSIM_MAX_PARALLEL="$mp" \
   util/dvsim/dvsim.py --tool=veriblelint "$dvsim_cfg" || {
-    echo -n "##vso[task.logissue type=error]"
-    echo "Verilog style lint of $human_desc sources with Verible failed. Run 'util/dvsim/dvsim.py -t veriblelint $dvsim_cfg' and fix all errors."
+    echo "::error::Verilog style lint of ${human_desc} sources with Verible failed."
+    echo "::error::Run 'util/dvsim/dvsim.py -t veriblelint ${dvsim_cfg}' and fix all errors."
```
nbdd0121, line 51: I think you probably still want these to be the same line to generate one annotation instead of two.

jwnrt:
That SiVal FPGA test failure is unrelated, merging

PR #24700, opened 2024-09-30
[csr_utils_pkg,dv] Avoid bug on reset in csr_spinwait

rswarbrick:
I broke this task for some pwrmgr tests with commit 5e0f0fe. The problem is that I changed the fork/join_any so that it would complete instantly on a reset and then use "disable fork" to kill the thing that was waiting.

Unfortunately, the thing that got killed might have been in the middle of a csr_rd and killing it doesn't allow the process to clean up properly.

Go back to waiting more carefully but (in a bid to get *some* benefit from the original rewrite) stop a wait between reads more quickly if there is a reset.

### DIFF
--- a/hw/dv/sv/csr_utils/csr_utils_pkg.sv
+++ b/hw/dv/sv/csr_utils/csr_utils_pkg.sv
@@ -578,11 +578,29 @@ package csr_utils_pkg;
         if (backdoor && spinwait_delay_ns == 0) spinwait_delay_ns = 1;
         fork
           // Repeatedly do reads of the value at path and stop when the value satisfies the supplied
-          // comparison. If it is positive, wait spinwait_delay_ns nanoseconds between reads.
+          // comparison. If it is positive, wait spinwait_delay_ns nanoseconds between reads. If we
+          // enter reset, stop immediately (once the current csr_rd call has finished).
           forever begin
-            if (spinwait_delay_ns) #(spinwait_delay_ns * 1ns);
             `uvm_info("csr_utils_pkg", $sformatf("In csr_spinwait - call_count = %0d", lcount),
                       verbosity)
+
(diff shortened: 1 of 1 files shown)


hw/dv/sv/csr_utils/csr_utils_pkg.sv
```diff
@@ -578,11 +578,31 @@ package csr_utils_pkg;
...
+            // and csr_rd has internal state that it must clean up without being killed.
+            if (spinwait_delay_ns) begin
+              fork begin : iso_fork
+                fork
+                  #(spinwait_delay_ns * 1ns);
+                  wait (under_reset);
+                join_any
+                disable fork;
+              end : iso_fork
+              join
```
hcallahan-lowrisc, line 600: Nit. I think you want this here to be consistent with other parts of the codebase
```suggestion
              end : iso_fork join
```

hw/dv/sv/csr_utils/csr_utils_pkg.sv
```diff
@@ -578,11 +578,31 @@ package csr_utils_pkg;
         if (backdoor && spinwait_delay_ns == 0) spinwait_delay_ns = 1;
         fork
           // Repeatedly do reads of the value at path and stop when the value satisfies the supplied
-          // comparison. If it is positive, wait spinwait_delay_ns nanoseconds between reads.
+          // comparison. If it is positive, wait spinwait_delay_ns nanoseconds between reads. If we
+          // enter reset, stop immediately (once the current csr_rd call has finished).
           forever begin
             if (spinwait_delay_ns) #(spinwait_delay_ns * 1ns);
```
hcallahan-lowrisc, line 584: I think this initial delay here is no longer required?

rswarbrick: Good point! I'd obviously intended to "copy, modify and delete" and had managed two out of three...

PR #24701, opened 2024-09-30
[dv,jtag] Change JTAG clock frequency to 24MHz for GLS

sha-ron:
Fix the code to match to the comments (24MHz for GLS, 50MHz for RTL).

Reverting the frequencies in [#24580](https://github.com/lowRISC/opentitan/pull/24580) so it will be correct.


rswarbrick (approved):
Thanks for sorting this out!

PR #24703, opened 2024-09-30
    [toolchain] add `--download-only` option

ngm0:
just download the archive; don't unpack.
    useful when the toolchain is externally managed
    in the development environment.


moidx (approved):
LGTM

PR #24709, opened 2024-10-01
[repo] Adjust BLOCKFILE to apply only to specific blocks

GregAC:
Previously the BLOCKFILE blocked all RTL changes to all IP blocks. This alters the file to block per IP block which allows us to release the block on certain IPs (DMA, Ascon and Key Manager DPE at time this commit was made).


Razer6 (approved):
LGTM

rswarbrick (approved):
This looks much less magic, which is nice.

BLOCKFILE
```diff
@@ -33,6 +31,48 @@ hw/vendor/lowrisc_ibex/rtl/*
...
+hw/ip/spi_device/rtl/*
+hw/ip/spi_host/rtl/*
+hw/ip/sram_ctrl/rtl/*
+hw/ip/sysrst_ctrl/rtl/*
+hw/ip/tlul/rtl/*
+hw/ip/uart/rtl/*
+hw/ip/usbdev/rtl/*
+
+hw/ip_templates/alert_handler/rtl/*
+hw/ip_templates/clkmgr/rtl/*
```
vogelpi, line 70: Especially for clkmgr and rstmgr but technically for all templated IP blocks here, we could drop blocker and instead enforce it on the top-level specific RTL and DV files. Because we will need changes to the templates and the generators to accommodate some of the changes required for Darjeeling (e.g., optionally generate SPI host / USB clocks etc.). But we don't really care about the template files changing as long as the generated RTL and DV files remain stable.

vogelpi (approved):
LGTM, thanks @GregAC ! I think we could also lift the blocker on the IP templates but we should enforce it on the generated RTL and DV files for Earlgrey.

BLOCKFILE
GregAC: Good point I'll drop the block on the templates

marnovandermaas:
CHANGE AUTHORIZED: BLOCKFILE

marnovandermaas (approved):
Authorizing changes for this.

Razer6 (approved):

jwnrt:
CHANGE AUTHORIZED: BLOCKFILE

HU90m (approved):

vogelpi (approved):
LGTM, thanks @GregAC for removing the blocker from the IP templates.

vogelpi:
CHANGE AUTHORIZED: BLOCKFILE

PR #24711, opened 2024-10-01
[ci] Enable GitHub Actions CI on pushes

jwnrt:
Before this PR, the CI only ran for pull requests and not, `master`, `earlgrey_es_sival`, etc.

This PR enables the CI workflow for all non-backport branches. Certain steps are skipped since they only lint diffs and not the whole repo (which isn't lint clean unfortunately). This should be fixed in the future but I don't have the patience right now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,11 @@
 name: CI
 on:
   pull_request:
+  push:
+    branches-ignore:
+      - "backport-*"
+    tags:
+      - "*"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,13 +37,14 @@ jobs:
       - name: Validate testplans with schema
         run: ./ci/scripts/validate_testplans.sh
       - name: C/C++ formatting
-        run: ./ci/scripts/clang-format.sh "$GITHUB_BASE_REF"
+        run: ./ci/bazelisk.sh test //quality:clang_format_check
       - name: Rust formatting
-        run: ./ci/scripts/rust-format.sh "$GITHUB_BASE_REF"
+        run: ./ci/bazelisk.sh test //quality:rustfmt_check
       - name: Shellcheck
         run: ./ci/bazelisk.sh test //quality:shellcheck_check
       - name: Header guards
         run: ./ci/scripts/include-guard.sh "$GITHUB_BASE_REF"
+        if: ${{ github.event_name == 'pull_request' }}
       - name: Trailing whitespace
         run: ./ci/scripts/whitespace.sh "$GITHUB_BASE_REF"
       - name: Broken links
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -37,13 +37,14 @@ jobs:
...
-        run: ./ci/scripts/clang-format.sh "$GITHUB_BASE_REF"
+        run: ./ci/bazelisk.sh test //quality:clang_format_check
       - name: Rust formatting
-        run: ./ci/scripts/rust-format.sh "$GITHUB_BASE_REF"
+        run: ./ci/bazelisk.sh test //quality:rustfmt_check
       - name: Shellcheck
         run: ./ci/bazelisk.sh test //quality:shellcheck_check
       - name: Header guards
         run: ./ci/scripts/include-guard.sh "$GITHUB_BASE_REF"
+        if: ${{ github.event_name == 'pull_request' }}
```
nbdd0121, line 55: This line goes to the wrong commit.

.github/workflows/ci.yml
```diff
@@ -37,13 +37,14 @@ jobs:
       - name: Validate testplans with schema
         run: ./ci/scripts/validate_testplans.sh
       - name: C/C++ formatting
-        run: ./ci/scripts/clang-format.sh "$GITHUB_BASE_REF"
+        run: ./ci/bazelisk.sh test //quality:clang_format_check
```
nbdd0121, line 48: This one takes ~25 seconds, so might worth consider optimizing in the future -- not blocking for this PR.

nbdd0121 (approved):
nits only

.github/workflows/ci.yml
jwnrt: I won't fix this for this PR, but in the future I hope to adopt `rules_lint` which will replace this and might be faster.

pamaury: I think this is not the same semantic? ``clang_format_check` checks the entire repo whereas the script only checks the diff? I think the issue is that we started linting after having some significant amount of code written so many files do not respect the style guide.

.github/workflows/ci.yml
```diff
@@ -24,28 +29,33 @@ jobs:
...
       - name: Python (mypy)
         run: ./ci/scripts/mypy.sh
       - name: Validate testplans with schema
         run: ./ci/scripts/validate_testplans.sh
       - name: C/C++ formatting
-        run: ./ci/scripts/clang-format.sh "$GITHUB_BASE_REF"
+        run: ./ci/bazelisk.sh test //quality:clang_format_check
       - name: Rust formatting
-        run: ./ci/scripts/rust-format.sh "$GITHUB_BASE_REF"
+        run: ./ci/bazelisk.sh test //quality:rustfmt_check
```
pamaury, line 50: I would maybe suggest to run `quality:rustfmt_fix` because contrary to `rustfmt_check`, it actually runs on the entire repository.

jwnrt: I've updated a new `rustfmt_test` rule that runs `rustfmt` but with the `--check` argument. Please could you review again?

nbdd0121:
Skip Azure CI since it's not relevant.

PR #24713, opened 2024-10-01
fix job status poll logic when timeout is not specified

venkatk-ot:
skip checking job_runtime_secs against timeout_mins when timeout_mins is None


rswarbrick (approved):
Thanks, this looks sensible to me.

venkatk-ot:
Thanks!

PR #24714, opened 2024-10-01
[ownership] Update ownership transfer documentation

cfrantz:
Update the ownership transfer documentation.
- Include a diagram of ownership state transitions.
- Add a high-level description of the ownership configuration.
- Describe the uses of the `update_mode`.
- Update struct definitions to reflect the current implementation.

### DIFF
--- a/sw/device/silicon_creator/rom_ext/doc/ownership.md
+++ b/sw/device/silicon_creator/rom_ext/doc/ownership.md
@@ -18,7 +18,7 @@ Ownership Transfer allows the owner of the chip to securely root an OpenTitan ch
 - Unlock Key: an ECDSA P-256 public/private key-pair used to authenticate an Unlock Owner command.
   This key is endorsed by the Owner key by virtue of being present in the owner configuration.
 - Ownership State: The state of a chip with respect to ownership transfer.
-  A chip may be in one of the following states: `LockedOwner`, `UnlockedSelf`, `UnlockedAny`, `UnlockedEndorsed`, `LockedNone`.
+  A chip may be in one of the following states: `LockedOwner`, `UnlockedSelf`, `UnlockedAny`, `UnlockedEndorsed`, `Recovery`.
 - Ownership Nonce: A 64-bit random integer used as a validation challenge for all ownership-related boot services requests.
 - Owner Page 0 & 1: The Owner Configuration is stored by the chip in a pair of flash INFO pages.
   Normally the two pages are identical and serve as redundant backups.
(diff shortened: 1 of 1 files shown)


sw/device/silicon_creator/rom_ext/doc/ownership.md
```diff
@@ -40,13 +40,60 @@ A chip may be in one of the following states:
...
+- An owner may want to provide an update mechanism for devices in the field that does not require the unlock and activate procedure.
+
+The update mode provides for three types of owner configuration update procedures:
+
+- `Open` mode allows the unlock key to unlock the device into any of the unlocked states.
+- `Self` mode allows the unlock key to unlock the device into only the `UnlockedSelf` state.
+- `NewVersion` mode forbids the unlock key from unlocking the device.
+  Instead, the device may accept an ownership configuration update at any time _if and only if_ the new configuration is signed by the self-same owner _and_ the configuration has a newer `config_version` number.
+
+When the update mode is either `Self` or `NewVersion`, the power of the unlock key is reduced to only allow ownership operations that do not change the owner.
```
jadephilipoom, line 95: Maybe worth an extra sentence here about how the power of the unlock key is reduced? My understanding is that in  `Self` or `NewVersion` the attacker would need to steal the owner key to take over ownership of the device, which contains the implicit assumption that the owner key is stored more securely than the unlock key. I think it's worth being explicit about this so that people know they need to protect the owner key in order to get the benefits of update mode.

sw/device/silicon_creator/rom_ext/doc/ownership.md
```diff
@@ -40,13 +40,60 @@ A chip may be in one of the following states:
...
+  - Enable or disable SRAM execution.
+  - Reset the BL0 minimum security version.
+
+#### Update Mode
+
+There are several allowed update modes for the ownership configuration.
+These update modes are motivated by two scenarios:
+
+- The Unlock Key is a very powerful key and owners may want to attenuate the power of the unlock key to mitigate the risk of the key leaking.
+- An owner may want to provide an update mechanism for devices in the field that does not require the unlock and activate procedure.
```
jadephilipoom, line 86: If I understand correctly, the main reason an owner might want this is that it would allow configuration updates without a device reset; it might be worth stating that to help people understand why owners would want to do this instead of a locked update.

sw/device/silicon_creator/rom_ext/doc/ownership.md
```diff
@@ -429,24 +493,44 @@ The Owner Configuration struct carries the required owner key materials and vali
...
+   * length: 2048.
+   */
+  tlv_header_t header;
+  /** Version of the owner struct.  Currently `0`. */
+  uint32_t struct_version;
+  /** SRAM execution configuration (DisabledLocked, Disabled, Enabled). */
+  uint32_t sram_exec_mode;
+  /** Ownership key algorithm (currently, only ECDSA is supported). */
+  uint32_t ownership_key_alg;
+  /** Configuraion version (monotonically increasing per owner) */
```
jadephilipoom, line 509:
```suggestion
  /** Configuration version (monotonically increasing per owner) */
```

sw/device/silicon_creator/rom_ext/doc/ownership.md
```diff
@@ -537,11 +684,23 @@ The rescue configuration describes the owner's desired configuration of the ROM\
...
+   * length: 16 + sizeof(command_allow).
+   */
+  tlv_header_t header;
+  /** The rescue type.  Currently, only `XMDM` is supported. */
+  uint32_t rescue_type;
+  /** The start offset of the rescue region in flash (in pages). */
+  uint16_t start;
+  /** The size of the rescue region in flash (in pages). */
+  uint16_t size;
+  /** An allowlist of rescue and boot_svc commands that may be invoked by the
```
jadephilipoom, line 700:
```suggestion
  /**
   * An allowlist of rescue and boot_svc commands that may be invoked by the
```

jadephilipoom (approved):
A few minor comments but LGTM :slightly_smiling_face: 

The current [ownership transfer docs](https://opentitan.org/book/doc/security/specs/ownership_transfer/index.html?highlight=ownership%20transfer#ownership-transfer) on the website are pretty outdated. Maybe we should consider simply updating `SUMMARY.md` to point to this page instead, and removing the old page?

sw/device/silicon_creator/rom_ext/doc/ownership.md
cfrantz: A reset is always required to update the owner config.  What the owner may want to avoid is the nonce-and-boot-services dance of unlock/update config/activate.

Full procedure:
1. Get nonce, sign unlock request (phone home), write to boot-svc.
2. Reset
3. Get updated owner config (phone home), write to owner page 1.
4. Reset
5. Get nonce, sign activate request (phone home), write to boot-svc.
6. Reset
7. Confirm operating under new config; back to operational.

Abbreviated procedure:
1. Get updated owner config (phone home, or included in a fw update).
2. Reset
3. Confirm operating under new config; back to operational.

sw/device/silicon_creator/rom_ext/doc/ownership.md
cfrantz: I added a section describing the owner key and included the example of storing the key in an offline HSM.

Can you clarify what you're asking for with "how the power of the unlock key is reduced"?  I think the text already describes the reduction in power of the unlock key:
- When in `Self` mode, the unlock key can only unlock to the `UnlockedSelf` state, meaning you can only update the config, not perform a full ownership transfer.
- When in `NewVersion` mode, you cannot move to an unlocked state at all.

jadephilipoom: I think I made a poor choice of words here -- I meant something like "how reducing the power of the unlock key increases security". What I was getting at is that if the owner key is no safer than the unlock key then you don't really win much, and the new section resolves this by making the expectation that the owner key is safer than the unlock key explicit. Thanks!

sw/device/silicon_creator/rom_ext/doc/ownership.md
jadephilipoom: Ah, I see, so it's more about replacing several round-trips and three resets with 0-1 round-trips and one reset? I initially misunderstood this, makes sense now.

cfrantz:
Documentation only; the CI failures are not related to this change.

PR #24715, opened 2024-10-02
[DO NOT MERGE] Test rules_python with local package

pamaury:
The purpose of this PR is to see if we can define local package using `pyproject.toml` and let `rules_python` use them as usual.

How to test:
- try it with:
```console
./bazelisk.sh run @ot_python_deps_hello//:rules_python_wheel_entry_point_hello
./bazelisk.sh run @ot_python_deps_hello//:rules_python_wheel_entry_point_bye
```

I added a patch so that rules_python watchs the local package's source code. You can try it by changing `hello.py` and see that if you re-run the commands above, the package is reinstall and the change is picked up.


pyproject.toml
```diff
@@ -84,8 +88,17 @@ dependencies = [
...
+
+  # OpenTitan python package
+  "opentitanlib @ file://${PROJECT_ROOT}/util/python",
 ]
 
 [tool.setuptools]
 # This is actually not a python project, we just use pyproject.toml to manage dependencies.
 py-modules = []
+
+[tool.uv.workspace]
```
machshev, line 103: Interesting concept! I've not seen that UV feature before... perhaps worth looking into.

util/hello/hello.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+
+from termcolor import colored
+
```
machshev, line 4: Would be interesting to see if you can use `opentitanlib.*` modules in here?

pamaury: Yes it works no problem. I have added an example in the last update

pamaury:
Note to self: I had to remove the workspace section of the top-level `pyproject.toml` because otherwise the nested `pyproject.toml` interpret that as the root of the workspace and leads to a very confusing behaviour when testing locally if one gets it wrong. I might revisit this workspace feature in the future because it sounds like it's still useful to pin dependencies for all sub-packages, if we decide to go down this path.

util/hello/WORKSPACE
jwnrt, line 1: Note: there is a .bazelignore file you can use for this instead

jwnrt (commented):
A few files are missing EOF newlines

PR #24717, opened 2024-10-02
[ci] add GitHub bitstream building job

nbdd0121:


nbdd0121:
A passing run: https://github.com/lowRISC/opentitan/actions/runs/11842465064/job/33001753065

jwnrt:
This looks good, but I think we should merge it disabled (with `if: false` or something) to prevent doubling our bitstream builds until we can move it off of Azure

jwnrt (approved):
LGTM, thanks

PR #24718, opened 2024-10-02
[topgen] Sync topgen from integrated_dev to master

Razer6:
This PR merges the topgen tool from `integrated_dev` to master. The major part of this PR is the support for multiple address spaces by @a-will. It further includes support for generically filtering MMIO regions, and changes related to Darjeeling's device structure (more than one ROM, no flash).

Work-around changes related to the clock manager are not yet merged, since it's not yet decided how that IP will look like in master eventually.


util/topgen/templates/chiplevel.sv.tpl
```diff
@@ -1157,6 +1167,16 @@ module chip_${top["name"]}_${target["name"]} #(
...
     .UsbdevRcvrWakeTimeUs(10000),
+  % if lib.num_rom_ctrl(top["module"]) > 1:
+    // TODO(opentitan-integrated/issues/251):
+    // Enable hashing below once the build infrastructure can
+    // load scrambled images on FPGA platforms. The DV can
+    // already partially handle it by initializing the 2nd ROM
+    // with random data via the backdoor loading interface - it
+    // can't load "real" SW images yet since that requires
+    // additional build infrastructure.
+    .SecRomCtrl1DisableScrambling(1),
```
vogelpi, line 1178: Is this still an issue?

util/topgen/validate.py
```diff
@@ -194,12 +195,14 @@
...
+    'base_addr': ['g', 'dict of address space mapped to the corresponding '
+                       'hex start address of the peripheral '
                        '(if the IP has only a single TL-UL interface)'],
-    'base_addrs': ['d', 'hex start addresses of the peripheral '
+    'base_addrs': ['g', 'hex start addresses of the peripheral '
                         ' (if the IP has multiple TL-UL interfaces)'],
     'memory': ['g', 'optional dict with memory region attributes'],
-    'param_decl': ['g', 'optional dict that allows to override instantiation parameters']
+    'param_decl': ['g', 'optional dict that allows to override instantiation parameters'],
+    'generate_dif': ['pb', 'optional bool to indicate if a DIF should be generated for that module']
```
pamaury, line 205: This setting does not seem to play a role in this PR (yet?)

vogelpi (approved):
Thanks for the PR @Razer6 . The hardware-related changes look good to me, the RTL isn't changed yet, just the hjson files and the tooling this is perfect.

For the software-related changes, I would like to get @pamaury's feedback as well. We should make sure these changes don't further complicate his work / if there is a way to make it go smoother.

vogelpi:
CHANGE AUTHORIZED: hw/top_earlgrey/data/xbar_main.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/data/xbar_peri.hjson
CHANGE AUTHORIZED: sw/device/silicon_creator/rom/e2e/release/rom_e2e_self_hash_test.c

This PR touches some files relevant for generating RTL for Earlgrey but the RTL itself doesn't change. This is fine. The changes to the ROM self hash test consist of renaming a variable only. This is fine.

util/topgen/templates/chiplevel.sv.tpl
Razer6: I think yes. In general, I don't want to make changes on these carry-over PRs.

andreaskurth:
CHANGE AUTHORIZED: hw/top_earlgrey/data/xbar_main.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/data/xbar_peri.hjson
CHANGE AUTHORIZED: sw/device/silicon_creator/rom/e2e/release/rom_e2e_self_hash_test.c

Agree with @vogelpi's assessment.

andreaskurth:
It looks like @pamaury has reviewed this and didn't request changes, though we should give @matutem a chance to chime in as well. Otherwise LGTM!

util/tlgen/item.py
```diff
@@ -36,7 +36,9 @@ class Node:
     # Other nodes have this fields set to -1.
     address_from: int = -1
     address_to: int = -1
+    addr_ranges = Dict[int, Tuple[int, int]]
```
matutem, line 39: Should this be
`addr_ranges: Dict[int, Tuple[int, int]] = {}`

But more curious is why these classes declare both instance and class variables. It would seem the class variables are unused. Was this code just created by @a-will or are you familiar enough with it?

Razer6: Oh yes, that's correct! 

This behavior is due to Python's typing system. The variables shown above are instance variables by default, not class variables. To explicitly declare a class variable with type hints, you need to use the `ClassVar` annotation.

I think the following documentation covers this:
https://typing.readthedocs.io/en/latest/spec/class-compat.html

util/topgen/validate.py
```diff
@@ -851,7 +855,7 @@ def check_modules(top, prefix):
                     if mem_type == "flash":
                         check_keys(value['config'], eflash_required, eflash_optional,
                                    eflash_added, "Eflash")
-                        flash = Flash(value['config'])
+                        flash = Flash(value['config'], m['base_addrs'][intf])
```
matutem, line 858: The check in line 866 should be performed before this line, probably in line 851.

util/topgen/top.py
```diff
@@ -4,7 +4,7 @@
 
 '''Code representing the entire chip for reggen'''
 
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Set
```
matutem, line 7: Union is used in line 71, so keep it. Also, the list of imports is sorted alphabetically, so reorder Set.

util/topgen/templates/xbar_tgl_excl.cfg.tpl
```diff
@@ -29,12 +29,16 @@
 %>\
 // [UNR] Exclude unused address bits based on IP address range. It is not possible to cover this.
 % for xbar in top["xbar"]:
+<%
+  # TODO: Add support for multiple ASIDs within an xbar.
```
matutem, line 33: Is this actually needed? Otherwise remove the TODO.

util/topgen/templates/toplevel_memory.ld.tpl
```diff
@@ -2,6 +2,8 @@
 /* Licensed under the Apache License, Version 2.0, see LICENSE for details. */
 /* SPDX-License-Identifier: Apache-2.0 */
 <%!
+import topgen.lib as lib
```
matutem, line 5: I don't see a use of `lib` in the modified code. Is this needed?

util/topgen/templates/xbar_env_pkg__params.sv.tpl
```diff
@@ -51,12 +51,16 @@
 // List of Xbar device memory map
 tl_device_t xbar_devices[$] = '{
 % for xbar in top["xbar"]:
+<%
+  # TODO: Add support for multiple ASIDs within an xbar.
```
matutem, line 55: Is this TODO really needed or just speculative? If the latter, please remove it.

util/topgen/lib.py
```diff
@@ -85,6 +85,40 @@ def remove_part(self, part_to_remove: str) -> "Name":
         return Name([p for p in self.parts if p != part_to_remove])
 
 
+class MemoryRegion(object):
```
matutem, line 88: Good replicated code removal :)

util/topgen.py
```diff
@@ -505,6 +508,7 @@ def generate_flash(topcfg: Dict[str, object], out_path: Path) -> None:
         "infos_per_bank": [10, 1, 2]
     })
 
+    params.pop('base_addrs', None)
```
matutem, line 511: I assume this was expected to come from line 503. Is it really necessary to pop this? Seems confusing.

util/topgen/templates/xbar_env_pkg__params.sv.tpl
a-will: An issue should still be created for the missing feature (and set to an appropriate priority level, even if it's P4 "probably never"), and the limitations should be documented. It needs to be obvious that the support for multiple address spaces is _only_ for entirely disjoint networks. With the current code, two hosts using different address spaces can't connect to the same device node.

Due to this limitation, IPs need to have a separate node defined for each address space, even if the hosts would actually access the same CSRs with the same privileges (tail-wagging-dog property flow). For the current set of known use cases, this seems to be fine, however.

util/tlgen/lib.py
```diff
@@ -150,7 +156,7 @@ def find_pow2_size(addr: Dict[Any, Any],
     should be less than or equal to 0x1_0000. Cannot be 0x4_0000. So, this
     case, the function returns the original min_size value 0x21000.
     """
-    base_addr = int(addr["base_addr"], 0)
+    base_addr = int(addr, 0)
```
matutem, line 159: This is weird: if addr is a dict then int(addr, 0) should give an error since int would expect addr to be a string. The caller seems to pass a string, so what is happening with the type annotation?

util/tlgen/lib.py
```diff
@@ -28,37 +32,37 @@ def simplify_addr(dev: Dict[Any, Any],
     {0x4000_0000, 0x1_0000}, {0x4008_0000, 0x1_0000} then it combines two
     entries into {0x4000_0000, 0x9_0000}
 
-    @param addrs List of Dict[Addr] : {'base_addr':,'size_byte':}
+    @param addrs List of Dict[Addr] : {'base_addr':Dict[ASID,Addr],'size_byte':}
```
matutem, line 35: What is this @param documenting? There is no addrs parameter, and Dict[ASID,Addr] seems an incorrect type.

util/tlgen/lib.py
```diff
@@ -28,37 +32,37 @@ def simplify_addr(dev: Dict[Any, Any],
...
                           0x100000000)
 
-            calc_size = int(e["base_addr"], 0) + int(e["size_byte"], 0) - int(
-                result[-1]["base_addr"], 0)
+            calc_size = int(e["base_addrs"][asid], 0) + int(e["size_byte"], 0) - int(
+                result[-1]["base_addrs"][asid], 0)
 
             # find power of 2 if possible
-            size_byte = find_pow2_size(result[-1], calc_size, next_value)
+            size_byte = find_pow2_size(result[-1]["base_addrs"][asid], calc_size, next_value)
```
matutem, line 65: The first arg of find_pow2_size seems to be a string?

util/topgen/templates/toplevel_memory.ld.tpl
Razer6: Not needed, the usage of `lib` was removed in this commit series. I removed the import

util/topgen/validate.py
Razer6: I moved the check above and changed it to raise an exception before reaching the usage of the entry.

util/topgen/top.py
Razer6: I agree that it is used as a typing hint. Though when importing `Union`, CI fails with:
```
  flake8 util/design/gen-top-docs.py util/reggen/gen_dv.py util/tlgen/elaborate.py util/tlgen/item.py util/tlgen/lib.py util/tlgen/validate.py util/tlgen/xbar.py util/topgen.py util/topgen/c.py util/topgen/c_test.py util/topgen/gen_dv.py util/topgen/intermodule.py util/topgen/lib.py util/topgen/merge.py util/topgen/rust.py util/topgen/top.py util/topgen/validate.py
	util/topgen/top.py:7:1: F401 'typing.Union' imported but unused
```

util/topgen/templates/xbar_env_pkg__params.sv.tpl
Razer6: I created https://github.com/lowRISC/opentitan/issues/24801 to track that.

util/topgen/templates/xbar_tgl_excl.cfg.tpl
Razer6: I created https://github.com/lowRISC/opentitan/issues/24801 to track that.

util/tlgen/lib.py
Razer6: Good point. I think the type annotation was added in the meanwhile on master. I updated the annotation to let `addr` be a string. The function is only used once.

util/tlgen/lib.py
Razer6: See the comment above.

util/tlgen/lib.py
Razer6: Lets drop that comment.

util/topgen/top.py
matutem: My mistake: Union is used within a comment, so should not be imported.

util/topgen/top.py
```diff
@@ -92,31 +94,39 @@ def __init__(self,
...
-                lst.append(inst_name)
-
-            # This should be guaranteed by the fact that we've already checked
-            # the existence of a device interface.
-            assert if_name in block.reg_blocks
-            reg_block = block.reg_blocks[if_name]
-
-            offset = base_addr + reg_block.offset
+        # One such dictionary per address space.
+        self.block_instances = {}  # type: Dict[str, Dict[str, List[str]]]
```
matutem, line 98: Please move the type comment into the declaration, here and in line 101.

util/topgen/validate.py
```diff
@@ -845,25 +849,25 @@ def check_modules(top, prefix):
                                      'nor extra configuration.  Unable to determine '
                                      'memory size')
 
+                # make sure the memory regions correspond to the TL-UL interfaces
+                if intf not in m['base_addrs']:
+                    raise ValueError("{} {} memory region {} does not "
```
matutem, line 854: Just a suggestion: replace <string>.format(...) by f-stings, since they are much more readable.

util/topgen/rust.py
```diff
@@ -214,6 +180,9 @@ def __init__(self, top_info, name_to_block: Dict[str, IpBlock], version_stamp: D
         self.regwidth = int(top_info["datawidth"])
         self.file_header = RustFileHeader(version_stamp)
 
+        # TODO: Don't hardcode the address space used for software.
+        self.addr_space = "hart"
```
matutem, line 184: Maybe this should be a constructor parameter?

util/topgen/rust.py
```diff
@@ -8,47 +8,13 @@
 from mako.template import Template
 from reggen.ip_block import IpBlock
 
-from .lib import Name, get_base_and_size
+from .lib import Name, get_base_and_size, MemoryRegion, init_subranges
```
matutem, line 11: Real nit: sort import items?

util/topgen/merge.py
```diff
@@ -213,6 +214,12 @@ def elaborate_instance(instance, block: IpBlock):
     if 'base_addr' in instance:
         del instance['base_addr']
 
+    # Default value if no value provided and otherwise convert string to bool
+    if 'generate_dif' not in instance:
+        instance['generate_dif'] = True
+    else:
+        instance['generate_dif'], _ = check_bool(instance['generate_dif'], False)
```
matutem, line 221: The second value in check_bool's returned tuple indicates an error, so it should be checked.

util/tlgen/item.py
```diff
@@ -66,7 +68,8 @@ def __init__(self,
         self.resets = [reset]
         self.us = []
         self.ds = []
-        self.addr_range = []
+        self.addr_ranges = {}
```
matutem, line 71: Nit, is this needed? The decl in line 39 sets it to {} already.

util/tlgen/validate.py
```diff
@@ -261,7 +266,7 @@ def validate(obj: Dict[Any, Any]) -> Optional[Xbar]:
     xbar.name = obj["name"].lower()
     xbar.clock = obj["clock"].lower()
     xbar.reset = obj["reset"].lower()
-    addr_ranges: List[Tuple[int, int]] = []
+    addr_ranges = {}
```
matutem, line 269: Add type, as in addr_ranges: Dict[str, [List[Tuple[int, int]]] = {}

util/topgen.py
```diff
@@ -603,10 +607,12 @@ def generate_top_ral(top: Dict[str, object], name_to_block: Dict[str, IpBlock],
         if t not in inst_to_block.values():
             del name_to_block[t]
 
-    chip = Top(regwidth, name_to_block, inst_to_block, if_addrs, mems, attrs)
+    addr_spaces = {addr_space["name"] for addr_space in top["addr_spaces"]}
+    chip = Top(regwidth, addr_spaces, name_to_block, inst_to_block, if_addrs, mems, attrs)
 
     # generate the top ral model with template
-    return gen_dv(chip, dv_base_names, str(out_path))
+    gen_dv(chip, dv_base_names, str(out_path))
```
matutem, line 614: You cannot ignore the return value from gen_dv: non-zero indicates an error.

util/topgen/c.py
```diff
@@ -111,6 +90,9 @@ def __init__(self, top_info, name_to_block: Dict[str, IpBlock]):
         # The .c file needs the .h file's relative path, store it here
         self.header_path = None
 
+        # TODO: Don't hardcode the address space used for software.
+        self.addr_space = "hart"
```
matutem, line 94: Same as rust.py: pass addr_space as an arg?

util/topgen/intermodule.py
```diff
@@ -1045,6 +1047,7 @@ def get_dangling_im_def(objs: OrderedDict) -> str:
 
     undriven_def = [obj for obj in objs if obj['end_idx'] > 0 and
                     (obj['act'] == 'req' and obj['suffix'] == 'rsp' or
-                     obj['act'] == 'rsp' and obj['suffix'] == 'req')]
+                     obj['act'] == 'rsp' and obj['suffix'] == 'req' or
+                     obj['act'] == 'rcv')]
```
matutem, line 1051: Curious: Is this a new possibility or was the code missing it?

a-will: I think you could argue either way. This is from 239f142fd09, which originally was part of https://github.com/lowRISC/opentitan-integrated/pull/280

util/topgen/rust.py
Razer6: This will already be fixed in a follow-up PR: https://github.com/lowRISC/opentitan/pull/23122 For now, I would hold-off changing that in this PR.

util/topgen/c.py
Razer6: Same comment as above.

util/topgen/merge.py
Razer6: `check_bool` already prints an error message. I also checked the error code and raise a `ValueError` if necessary.

util/topgen.py
Razer6: Changed that back to the original code.

util/topgen.py
```diff
@@ -630,7 +636,7 @@ def create_mem(item, addrsep, regwidth):
                          swaccess=swaccess)
 
 
-def generete_rust(topname, completecfg, name_to_block, out_path, version_stamp,
+def generate_rust(topname, completecfg, name_to_block, out_path, version_stamp,
                   src_tree_top, topgen_template_path):
     # Template render helper
     def render_template(template_path: str, rendered_path: Path, **other_info):
```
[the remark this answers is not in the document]
Razer6: Since this PR is not touching that particular code, I probably want to hold off with that. I created an issue with that instead: https://github.com/lowRISC/opentitan/issues/24822

Razer6:
@matutem Thank you for the thorough review. I changed the PR and/or added comments.

util/topgen.py
Razer6: 👍

util/topgen/merge.py
matutem: Thanks

util/tlgen/item.py
matutem: Cool

util/topgen.py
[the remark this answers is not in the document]
matutem: Okay

util/topgen/intermodule.py
matutem: Okay

matutem (approved):
Thanks for your patience :)
LGTM

Razer6:
@andreaskurth Can we please get that merged to unblock some upstream and downstream work?

Issue #20308, opened 2023-11-10
[crytpo] Suggested AES driver changes


vogelpi:
### Description

When investigating a [cryptotest failure for AES](https://github.com/lowRISC/opentitan/issues/20277) I had a brief look at the AES driver of the cryptolib and I noted the following things that should probably be changed:

---

- [ ] aes_write_key(): Write random bits for unused key words

The engine always loads 2x256 bits of key material into the cipher core (happens once per block). If half of the bits is 0 (for 128-bit keys) it maybe easier to extract key bits via profiling and SCA. Thus, it is preferred to fill with unused bits with randomness instead of 0. This is [documented in the programmers guide of AES](https://opentitan.org/book/hw/ip/aes/doc/programmers_guide.html#initialization):
> Anything can be written to the unused key registers of both shares, however, random data is preferred.

---

- [ ] otcrypto_aes(): overlap operations to improve performance and security (SCA)

This function internally uses the `aes_update()` function which itself follows the [recommended procedure in the programmers guide of AES](https://opentitan.org/book/hw/ip/aes/doc/programmers_guide.html#block-operation). However, `otcrypto_aes()` doesn't use `aes_update()` correctly. Namely, it should do

  //   aes_update(NULL, input[0]);
  //   aes_update(NULL, input[1]);
  //   aes_update(output[0], input[2]);
  //   aes_update(output[1], input[3]);

Instead of

  //   aes_update(NULL, input[0]);
  //   aes_update(output[0], input[1]);
  //   aes_update(output[1], input[2]);

With the former, the AES engine will process input[1] while software reads output[0] and provides input[2]. This gives the best peformance. With the latter (current implementation), software waits for output[0] to be ready, reads it, provides the new input[1] and then again waits for output[1]. Meaning always either software or the AES engine wait. But the two don't overlap their operations which is bad for SCA hardening and performance.

vogelpi:
FYI @jadephilipoom @johannheyszl

jadephilipoom:
Thanks Pirmin! Both of these changes seem like good ideas. I'll track them on the cryptolib milestone and assign myself for now (although I might not get to them immediately, so if anyone else wants to jump in feel free).

PR #24722, opened 2024-10-03
[crypto] Let otcrypto_aes() have up to 3 blocks in flight

vogelpi:
Previously, the hot loop inside otcrypto_aes() would only ever have 2 blocks in flight while the underlying driver function aes_update() as well as the hardware allows up to 3 blocks being in flight:
- Block x-1 can be retrieved from the data output registers by SW.
- Block x   is being processed by HW.
- Block x+1 can already be provided via the data input registers by SW.

Doing this is better for SCA hardening (more concurrency means more background noise) and allows achieving higher throughput.

This is related to lowRISC/OpenTitan#20308.

This is a cherry pick of commit 8f0f5003a4f7b0df446c718a71fc84b7bbc75476 to branch earlgrey_es_sival.

### DIFF
--- a/sw/device/lib/crypto/impl/aes.c
+++ b/sw/device/lib/crypto/impl/aes.c
@@ -355,40 +355,67 @@ otcrypto_status_t otcrypto_aes(const otcrypto_blinded_key_t *key,
       return OTCRYPTO_BAD_ARGS;
   }
 
-  // Perform the cipher operation for all full blocks (excluding last block).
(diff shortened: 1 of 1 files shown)


jadephilipoom (approved):

vogelpi:
Thanks for your approval @jadephilipoom .

CI passes except for the `sysrst_ctrl_in_irq_test_fpga_cw310_sival`  test. But I see the same test is failing also for other CI runs, e.g., #24712. I am thus merging this one here.

PR #24724, opened 2024-10-03
[kmac/dv] Add sideload_invalid test.

### LINKED ISSUE #22956: [kmac/dv] Extend DV to cover sideload key becoming invalid in all states
### Description

Factored out from #22794 and #16855. We should extend DV to cover the sideload key becoming invalid during operation in all possible states.

@andreaskurth and I discussed that this is not ultra critical (thus M5 rather than M4) but that it's still relevant because KEYMGR may clear sideload keys anytime it enters the invalid state / during an escalation. We should make sure this is covered sufficiently in DV to verify that KMAC does the right thing in this case.

nasahlpa:
This commit adds the new sideload_invalid test. The goal of this test is to increase coverage of the `StKeyMgrErrKeyNotValid` state of the kmac_app FSM.

In the test, during a random FSM state, the sideload_key.valid signal is forced to 0. The test checks whether the expected `ErrKeyNotValid` error is seen.

The following is now covered:
```console
StAppCfg->StKeyMgrErrKeyNotValid
StAppMsg->StKeyMgrErrKeyNotValid
StSw->StKeyMgrErrKeyNotValid
```
The other transitions are currently not covered because (_i_) the sequence does not trigger the error cases, (_ii_) or it seems not to be possible to reach this state (c.f. issue #24741).

----

For reference, this PR is related to #22956.


hw/ip/kmac/dv/env/kmac_env_pkg.sv
```diff
@@ -156,6 +156,23 @@ package kmac_env_pkg;
     StError                 = 10'b1101011101
   } kmac_app_st_e;
 
+  typedef enum bit [9:0] {
```
rswarbrick, line 159: It's a bit sad to have to copy this out of the design RTL. Since changes aren't actually blocked any more, maybe it would make more sense to pull the enum out of `kmac_app.sv` and put it in `kmac_pkg.sv` (and then use the values here) ?

hw/ip/kmac/dv/env/kmac_scoreboard.sv
```diff
@@ -576,7 +576,8 @@ class kmac_scoreboard extends cip_base_scoreboard #(
                     "in_kmac_app is not set, scoreboard has not picked up KMAC_APP request")
 
                 // Check app interface errors.
-                if (app_mode == AppKeymgr && cfg.enable_masking && !entropy_ready) begin
+                if (app_mode == AppKeymgr &&
+                    ((cfg.enable_masking && !entropy_ready) || cfg.key_invalidated)) begin
```
rswarbrick, line 580: I spent a couple of minutes thinking about this, and I think it can be slightly rejigged, which I think makes things a bit clearer. (But please check my comments! They might be wrong!)

```systemverilog
                // We expect an app interface error if app_mode is AppKeymgr (meaning that we are in
                // the mode where we are talking to the keymgr) and either there is no entropy to
                // perform the masking that is enabled or the key has been invalidated.
                app_intf_err = (app_mode == AppKeymgr &&
                                ((cfg.enable_masking && !entropy_ready) || cfg.key_invalidated));

                // Check app interface errors have been reported as expected.
                `DV_CHECK_FATAL(kmac_app_rsp.rsp_error == app_intf_err)

                // Check that digests have been zeroed if there was an interface error. If not,
                // extract the digests and (if configured) check they are correct.
                if (app_intf_err) begin
                  `DV_CHECK_FATAL(kmac_app_rsp.rsp_digest_share0 == 0,
                    "APP interface error, expect output to be all 0s")
                  `DV_CHECK_FATAL(kmac_app_rsp.rsp_digest_share1 == 0,
                    "APP interface error, expect output to be all 0s")
                end else begin

                  // assign digest values
                  kmac_app_digest_share0 = kmac_app_rsp.rsp_digest_share0;
                  kmac_app_digest_share1 = kmac_app_rsp.rsp_digest_share1;

                  if (do_check_digest) check_digest();
                end
```

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+    bit kmac_done;
+    bit invalidate_key;
+    key_sideload_set_seq sideload_seq;
+    bit [9:0] curr_state;
+    string key_valid_path = "tb.dut.keymgr_key_i.valid";
+    string app_fsm_path = "tb.dut.u_app_intf.st";
+    bit last_msg_endian = 1;
+    exp_kmac_err = '{valid: 1'b1,
+                     code: kmac_pkg::ErrKeyNotValid,
+                    info: '0};
```
rswarbrick, line 80: Nit: Indentation is wrong here, I think. Also, you can probably move the declaration of the variable to this line.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+    for (int i = 0; i < num_trans; i++) begin
+      bit [7:0] share0[];
+      bit [7:0] share1[];
+
+      `uvm_info(`gfn, $sformatf("iteration: %0d/%0d", i, num_trans), UVM_LOW)
+
+      `DV_CHECK_RANDOMIZE_FATAL(this)
+      // Make sure that cfg_shadow is different from the cfg_shadow used in the previous
+      // iteration. This is important as in kmac_init we do a csr_update which only
+      // updates the cfg_shadow register if there is a difference. If we would not doing
```
rswarbrick, line 96: Can't you use `csr_wr` here to force the write to actually happen?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+
+      set_prefix();
+
+      // Set the sideload key.
+      `uvm_create_on(sideload_seq, p_sequencer.key_sideload_sequencer_h);
+      `DV_CHECK_RANDOMIZE_WITH_FATAL(sideload_seq,
+                                     sideload_key.valid == 1;)
+      `uvm_send(sideload_seq)
+
+      // Write the SW key to the CSRs.
```
rswarbrick, line 119: Maybe worth referring to `en_app` here too? (We don't write the SW key if it's enabled, I think?)

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+      if (!en_app) begin
+        write_key_shares();
+      end
+
+      if (cfg.enable_masking && entropy_mode == EntropyModeSw) begin
+        provide_sw_entropy();
+      end
+
+      // Wait a few cycles to let the internal state settle in.
```
rswarbrick, line 128: Nit: I think you want "... state to settle in."

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+      cfg.clk_rst_vif.wait_clks($urandom_range(5, 10));
+
+      kmac_done = 0;
+      invalidate_key = 0;
+      cfg.key_invalidated = 0;
+      curr_state = Idle;
```
rswarbrick, line 134: Probably better to move these variables to be declared inside the `for` loop (and they can get values at that point)

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+      fork
+        begin : send_kmac_req
```
rswarbrick, line 136: As a general point, I'd probably suggest moving the big items in the `fork` into their own tasks. So you'd have one task for the body of `send_kmac_req`. I think the result might be easier to read.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+          if (en_app) begin
+            send_kmac_app_req(app_mode);
+
+            // Wait until we got the response.
+            // TODO(lowrisc/opentitan#24739): Currently, when invalidating the key when the
+            // app_i.last was received, exiting the StErrorAwaitApp is not possible anymore.
+            // After resolving this issue, we simply can wait until we got kmac_done.
+            wait(cfg.m_kmac_app_agent_cfg[app_mode].vif.kmac_data_req.valid &&
+                cfg.m_kmac_app_agent_cfg[app_mode].vif.kmac_data_req.last);
```
rswarbrick, line 145: Nit: Indentation off here.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,242 @@
...
+                cfg.clk_rst_vif.wait_n_clks(1);
+                `DV_CHECK(uvm_hdl_read(app_fsm_path, curr_state))
+              end
+              // Wait until the sideload key gets valid.
+              wait(cfg.keymgr_sideload_agent_cfg.vif.sideload_key.valid);
+              invalidate_key = 1;
+            end
+            ,
+            // If we are already finished, abort waiting for FSM state or key valid.
+            wait(kmac_done);
```
rswarbrick, line 191: Hmm. Isn't this going to kill the process that's doing a `uvm_hdl_read`? That could be ok if we used a backdoor read maybe?

rswarbrick (commented):
Lots of nitty comments, but this looks reasonably sensible to me. I think @martin-velay knows a bit more about the KMAC DV structure than I do: Martin, would you mind taking a look at this?

hw/ip/kmac/data/kmac_testplan.hjson
```diff
@@ -184,6 +184,23 @@
...
       tests: ["{name}_key_error"]
     }
+    {
+      name: sideload_invalid
+      desc: ''' Test kmac responses correctly sideloaded key is invalidated.
+            **Stimulus**:
+            - Configure kmac and send keymgr request.
+            - Wait until random FSM app state is entered.
+            - Invalidate the sideloaded key by forcing the .valid signal to 0.
+            - Repeat the above sequence a few times.
```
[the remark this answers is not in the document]
nasahlpa: In each of these iterations, a new random KMAC configuration is used and tested. Depending on the KMAC configuration, we are entering different FSM states which we want to test.

hw/ip/kmac/data/kmac_testplan.hjson
```diff
@@ -184,6 +184,23 @@
...
+
+            **Check**:
+            - Check error related registers including `err_code`, `status`, and `interrupt`.
+            - Check keymgr interface responses with all zero digests and error bit is set.
```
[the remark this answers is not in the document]
nasahlpa: Done, thanks.

hw/ip/kmac/dv/env/kmac_env_pkg.sv
nasahlpa: Yes, I agree - this is not good practice. In the first place, I hesitated pulling this out to the `kmac_pkg` but I think we should do this now. Done.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: Done thanks.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: Yes, this is going to kill ``DV_CHECK(uvm_hdl_read(app_fsm_path, curr_state))` above. Do you have a recommendation how this should be implemented? It seems that `uvm_hdl_read` does not support a backdoor parameter?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: You are referring to the comment or to the `uvm_send`? The sideload key is used in both cases.

hw/ip/kmac/dv/env/kmac_scoreboard.sv
nasahlpa: Thanks, I've added you suggestion as it increases the code readability.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: Ideally, the `csr_update` in the kmac_init task:
https://github.com/lowRISC/opentitan/blob/6e5b27f59be08e442daf27791f085f77c9695896/hw/ip/kmac/dv/env/seq_lib/kmac_base_vseq.sv#L340
gets replaced with a `csr_wr`. This should be equivalent, right?

Do I simply replace
https://github.com/lowRISC/opentitan/blob/6e5b27f59be08e442daf27791f085f77c9695896/hw/ip/kmac/dv/env/seq_lib/kmac_base_vseq.sv#L328C4-L328C43
with
`csr_wr(.ptr(ral.cfg_shadowed.kmac_en), .value(kmac_en));`
?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: Agree.

When moving this into a new task, how can I share the `kmac_done` bit?  Just use a global variable?

hw/ip/kmac/data/kmac_testplan.hjson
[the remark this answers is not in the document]
rswarbrick: That makes sense, maybe it would make sense to add something to make it more like "Repeat the above sequence a few times to try different configurations." ?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
rswarbrick: It is equivalent, but forces a write in all the other KMAC sequences even if it wasn't necessary. I thought the `csr_update` call you are talking about in this comment is the one a couple of lines later (line 103)?

Do you want to trigger a CSR write to `cfg_shadowed` that happens here? If so, something silly like ```csr_wr(.ptr(ral.cfg_shadowed), .value(`gmv(ral.cfg_shadowed)))``` would work (It means "write the value that we think the register already has").

But don't you want to configure some of the fields in `ral.cfg_shadowed` before this line anyway?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
rswarbrick: Ah! You've moved the comment inside the `if` block now and I understand! Thanks.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
rswarbrick: I'm really sorry but my original comment was completely wrong! Your text was correct and my suggestion has caused it to get mangled. I'm really sorry: would you mind going back to what you wrote in the first place?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
rswarbrick: Hmm, good point. I hadn't noticed that! I guess a class variable would work, but maybe that's more confusing than what you've got already. Maybe ignore my suggestion! :-)

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
rswarbrick: Duh. Your code is completely correct. I thought it was doing something like a `csr_rd`. The `uvm_hdl_read` macro *is* backdoor already. Looks good to me!

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,240 @@
...
+
+      `uvm_info(`gfn, $sformatf("iteration: %0d/%0d", i, num_trans), UVM_LOW)
+
+      `DV_CHECK_RANDOMIZE_FATAL(this)
+      // Make sure that cfg_shadow is different from the cfg_shadow used in the previous
+      // iteration. This is important as in kmac_init we do a csr_update which only
+      // updates the cfg_shadow register if there is a difference. If we would not doing
+      // this, the entropy_ready bit does not get set but the HW is waiting for the entropy
+      // ready pulse.
+      msg_endian = ~last_msg_endian;
```
martin-velay, line 101: I don't see where `msg_endian` or `last_msg_endian` are actually used?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
```diff
@@ -0,0 +1,240 @@
...
+      last_msg_endian = msg_endian;
+      csr_update(.csr(ral.cfg_shadowed));
+
+      kmac_init(.keymgr_app_intf(is_keymgr_app));
+      `uvm_info(`gfn, "kmac_init done", UVM_HIGH)
+
+      // Disable the scoreboard for the error injection test.
+      cfg.en_scb = 0;
```
martin-velay, line 109: What is checked if the SB is disabled? How do you ensure the DUT is doing what is expected?

martin-velay (approved):
This PR is looking good.

Note: nevertheless the KMAC looks to be a good example of what we shouldn't do in terms of DV, particularly the way the sequences (doing monitoring, driving and scoreboarding).

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: Changed the sentence back to: `Wait a few cycles to let the internal state settle in.`

hw/ip/kmac/data/kmac_testplan.hjson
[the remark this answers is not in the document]
nasahlpa: Done, thanks for the feedback.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: Ah right, thanks. Removed.

hw/ip/kmac/data/kmac_testplan.hjson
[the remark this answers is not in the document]
rswarbrick: I'm really sorry but I don't think it says anything about configurations yet?

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
rswarbrick: Thanks!

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: Thanks for the pointer to `csr_wr(.ptr(ral.cfg_shadowed), .value(`gmv(ral.cfg_shadowed));` This is exactly what I need here.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: I've now managed to create own tasks for the fork. In my opinion, it increases the readability.

hw/ip/kmac/dv/env/seq_lib/kmac_sideload_invalid_vseq.sv
nasahlpa: I've enabled the scoreboard again after the error injection. However, as discussed offline, you are right and it would be better to adapt the scoreboard accordingly.

hw/ip/kmac/data/kmac_testplan.hjson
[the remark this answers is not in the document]
nasahlpa: Should now be updated, I've just pushed the changes based on your feedback :-)

hw/ip/kmac/dv/env/kmac_env_pkg.sv
vogelpi: I agree that moving the enum out of `kmac_app.sv` is cleaner. But the consequence of the RTL change is that we won't be able to directly cherry-pick this PR into the TO branch. However, this should be fine as the current understanding is that the V3 work happens on the master branch only. If we still want to cherry-pick it over, the edit should be manageable.

vogelpi (approved):
Thanks @nasahlpa for your work on this, and @martin-velay as well as @rswarbrick for your reviews!

I think this is nice team work and we should okay the PR. It doesn't cover everything yet, but Pascal documented the missing transitions in issue #24741. So we now have something on which future efforts can be based on.

vogelpi:
CHANGE AUTHORIZED: hw/ip/kmac/rtl/kmac_app.sv
CHANGE AUTHORIZED: hw/ip/kmac/rtl/kmac_pkg.sv

This PR moves around some enum definitions in the RTL to simplify DV. We won't be able to directly cherry-pick this over into the TO branch but this is not needed. The important thing is that this PR doesn't alter the software interface or the behavior of the design. This is fine.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/kmac/rtl/kmac_app.sv
CHANGE AUTHORIZED: hw/ip/kmac/rtl/kmac_pkg.sv

PR #24730, opened 2024-10-04
Enable entropy_src block level tests to run using vcs simulator tool …

venkatk-ot:
…by fixing a couple of issues

Issue 1: csr_rd function didn't specify all required arguments. In this case, it was the value argument.
------------------------------------------------------------------- Error-[TFAFTC] Too few arguments to function/task call
                ../src/lowrisc_dv_entropy_src_env_0.1/seq_lib/entropy_src_rng_vseq.sv, 507
                vcs_paramclassrepository, "csr_utils_pkg::csr_rd(ral.fw_ov_wr_fifo_full.fw_ov_wr_fifo_full);"
                  The above function/task call is not done with sufficient arguments.
-------------------------------------------------------------------- Solution: declare a variable (call it unused for clarity that the value will not be sampled elsewhere in code) pass this 'unused' variable as the value argument of the csr_rd call -------------------------------------------------------------------- -------------------------------------------------------------------- Issue 2: code to disable assertion from within sequences using XMR doesnt compile This is because VCS prohibits SV class inside a package to have XMRs to code in module outside -------------------------------------------------------------------- Error-[SV-LCM-HRP] Hierarchical reference in package
                ../src/lowrisc_dv_entropy_src_env_0.1/seq_lib/entropy_src_rng_vseq.sv, 861
                entropy_src_env_pkg, "tb.dut.u_entropy_src_core.AtReset_EsrngFifoPushedIntoEsbitOrPosthtFifos_A"
                  Hierarchical reference from package 'entropy_src_env_pkg' into module 'tb'
                  is not allowed.
-------------------------------------------------------------------- Solution: move the functions containing assert_off tasks to the entropy_src_path_if call these functions as appropriate from the sequences using handle to the virtual interface

### DIFF
--- a/hw/ip/entropy_src/dv/env/entropy_src_path_if.sv
+++ b/hw/ip/entropy_src/dv/env/entropy_src_path_if.sv
@@ -4,6 +4,13 @@
 
 // This interface deals with the force paths in ENTROPY_SRC interrupt and error tests
 
+`define CORE \
+    tb.dut.u_entropy_src_core
+`define REPCNT \
+    u_entropy_src_repcnt_ht.u_prim_max_tree_rep_cntr_max
+`define BUCKET \
+    u_entropy_src_bucket_ht.u_prim_max_tree_bin_cntr_max
+
 interface entropy_src_path_if ();
   import uvm_pkg::*;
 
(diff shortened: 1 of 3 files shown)


rswarbrick (commented):
Hah, this is quite funny! The "assertoff in a vseq" code was actually my fault (6a75417). With this change, the net result is squashing together two interfaces. I wish I'd thought to do that in the first place!

As with the OTBN example, please could you split this into two commits (one for each change) and give each commit a helpful error message?

venkatk-ot:
split into two different commits done!

vogelpi (commented):
Thanks @venkatk-ot for your PR! The code looks good to me.

Can you please update the commit messages to follow [our guidance](https://opentitan.org/book/doc/contributing/detailed_contribution_guide/index.html?highlight=in-depth#what-should-commit-messages-look-like). This will help us tracking changes (relevant for porting across branches etc.). In particular, we would need you to:
- Add a tag to the subject line of the commit message, e.g., `[entropy_src/dv]` would be very suitable
- Wrap lines at 72 chars max. This holds for subject and body.
- It's great that you motivate why the changes are needed, but including the error message of the tool is not required.

To give you an example, the commit message of your second commit could be:
```
[entropy_src/dv] Add missing 'value' argument to the csr_rd call

This change fixes an issue seen when trying to run the entropy_src
block using VCS. A call of csr_rd() function did not specify all
required arguments. This commit solves the issue by declaring a
variable (it's called unused for clarity that the value will not be
sampled elsewhere in code) and passing this 'unused' variable as the
value argument of the csr_rd call.

Signed-off-by: Venkat Krishnan <venkateswarank.opentitan@gmail.com>
```

venkatk-ot:
Thanks @vogelpi . I have updated the commit messages per your suggestions. Please take a look.

vogelpi (approved):
Thanks @venkatk-ot for implementing the feedback. This looks good to me!

venkatk-ot:
@vogelpi , thanks for letting me know the errors and steps to reproduce them. 
I have fixed the lint warnings in the latest patch

moidx:
Picking this up before the next regression run. Thanks everyone.

PR #24731, opened 2024-10-04
Enable running the spi_host block level tests using vcs simulator by …

venkatk-ot:
…fixing a compile issue

Issue: No argument for format specification
Error-[MEAFFS] No argument for format specification
               ../src/lowrisc_dv_spi_host_env_0.1/seq_lib/spi_host_base_vseq.sv, 117
                 Missing or empty argument for format specification
                 Source info: $sformatf("spi_config_regs.clkdiv[0]=0x%0x is out range")


rswarbrick (commented):
Thanks for doing this: it looks right to me. But please could you force-push in a way that fixes the commit message? The first line should probably have tags to be something like "[spi_host,dv] Fix compile error in base vseq". Then (after a blank line) you might want to write something simple. Maybe something like "The previous code isn't quite legal, and causes a compile error with VCS."

venkatk-ot:
@rswarbrick , I have updated the commit message as per your suggestion. 
Please take a look

rswarbrick (approved):
Thanks!

Issue #24657, opened 2024-09-25
[rom_ext] Add an update_policy to the ownership block, allowing the owner to specify how ownership updates should be performed


cfrantz:
Examples:
- Full unlock allowed.
- Unlock to Self only.
- Accept signed updates from Self if config_version is greater.

PR #24745, opened 2024-10-07
[ownership] Add node-locked owner configs 

cfrantz:
Allow the owenrship block to specify that it is locked to a specific device ID.

1. Add `lock_constraints` and `device_id` fields to the owner config.
2. Use the device_id from the lifecycle controller in the cryptographic verification of the owner config.
3. Add tests to verify node locking.

Addresses: #24657

### DIFF
--- a/sw/device/silicon_creator/lib/ownership/BUILD
+++ b/sw/device/silicon_creator/lib/ownership/BUILD
@@ -106,6 +106,7 @@ cc_library(
         ":datatypes",
         ":ecdsa",
         ":owner_block",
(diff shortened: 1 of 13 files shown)


timothytrippel (approved):

cfrantz:
Note: I rewrote this to provide a usage_constraints-like lock on the entire device_id rather than just the lock on the device identification number.

This PR now depends on #24760.

cfrantz:
The CI test failures are not related to this PR.

PR #24748, opened 2024-10-08
[reggen] Allow a multireg's resval to depend on the index and use this in pinmux

### LINKED ISSUE #24621: [reggen] Add functionality to individualize resval for multireg instances
pinmux's MIO_PAD_ATTR is a hwext-implemented multireg, where the reset value (resval) is `0` for all instances except one. reggen currently only supports a homogeneous reset value over all instances of a multireg, however.

## Tasks

- [ ] Enhance reggen so that multiregs can have an instance-specific resval.
- [ ] Use this in pinmux's MIO_PAD_ATTR register and re-enable the automated reset value tests (change tag `CsrExclAll` back to `CsrExclWrite`).

rswarbrick:
Until the last two commits, this change is all about allowing a multireg to expand more cleanly so that e.g. it can have different resvals for the copies with indexes 0 and 1. There's a certain amount of code cleanup in the earlier commits, which I had to do in order to work out what on earth was going on in the code!

The last two commits use this new feature in pinmux.hjson. This resolves #24621, which motivated the change. Annoyingly, there's *still* a bit of a hack where we encode the index for the magic pad into `pinmux.hjson`, but we've got similar examples elsewhere in the repository and I think it's a bit cleaner than what we had.


rswarbrick:
Force-push fixes a silly bug where we didn't interpret the field resval for gpio's DATA_IN register. (It has reset value 'x' and is not hwext and my rewrite didn't handle that quite right. Now fixed)

rswarbrick:
Force-push fixes two problems:
1. The penultimate commit changes autogenerated files (intentionally!) and I hadn't updated them.
2. The "[reggen] Move Register.make_multi into MultiRegister" commit wasn't distinguishing between a field with resval x and empty resval. It should!

rswarbrick:
Well, that was rather painful. I've just rebased this past an enormous autoformatting rewrite. @andreaskurth: do you think you'll be able to take a look at this soon?

Razer6:
I also would like to see that feature to land as we have an immediate use case for that. @andreaskurth Can this prioritized? I will review this PR tomorrow.

util/reggen/field.py
```diff
@@ -82,11 +82,141 @@ def __init__(self, name: str, alias_target: Optional[str],
...
+        if resval is None:
+            # raw_value should be a string which we can evaluate as Python
+            # code.
+            if not isinstance(raw_value, str):
+                raise ValueError(f'{what} is not a bool or integer, so it '
+                                 f'should be a string containing a Python '
+                                 f'expression. Instead, it is {raw_value!r}')
+
+            try:
+                resval = eval(raw_value, bindings)
```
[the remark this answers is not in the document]
rswarbrick: I'm afraid it isn't:
```
>>> ast.literal_eval("1 == 2")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nix/store/xm56r94jg7mxcqdr6r2q8wki795mjgda-python3-3.11.11/lib/python3.11/ast.py", line 110, in literal_eval
    return _convert(node_or_string)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/xm56r94jg7mxcqdr6r2q8wki795mjgda-python3-3.11.11/lib/python3.11/ast.py", line 109, in _convert
    return _convert_signed_num(node)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/xm56r94jg7mxcqdr6r2q8wki795mjgda-python3-3.11.11/lib/python3.11/ast.py", line 83, in _convert_signed_num
    return _convert_num(node)
           ^^^^^^^^^^^^^^^^^^
  File "/nix/store/xm56r94jg7mxcqdr6r2q8wki795mjgda-python3-3.11.11/lib/python3.11/ast.py", line 74, in _convert_num
    _raise_malformed_node(node)
  File "/nix/store/xm56r94jg7mxcqdr6r2q8wki795mjgda-python3-3.11.11/lib/python3.11/ast.py", line 71, in _raise_malformed_node
    raise ValueError(msg + f': {node!r}')
ValueError: malformed node or string on line 1: <ast.Compare object at 0x7f3ad4f75a80>
```

The [literal_eval](https://docs.python.org/3/library/ast.html#ast-helpers) function doesn't allow operators and the only place we're using it in this PR is like this:
```
resval: (multireg_idx == 30)
```

rswarbrick:
(Nothing interesting in the force-push: it's just a trivial rebase, pulling in the changes on master since last time)

hw/top_darjeeling/ip_autogen/pinmux/data/pinmux.hjson
```diff
@@ -260,19 +260,25 @@
...
+                      resval: 0
                     },
                     { bits: "2",
                       name: "pull_en",
                       desc: "Enable pull-up or pull-down resistor."
+                      // Note: The special index (30) where this has a different resval depends on
+                      // the TargetCfg parameter that gets passed in. That, in turn, is configured
+                      // at chip level and it would be nicer if we could resolve it as we expanded
+                      // this template.
+                      resval: (multireg_idx == 30)
```
andreaskurth, line 281: Does this work for Darjeeling? I'm asking because for Earlgrey we have
https://github.com/lowRISC/opentitan/blob/6289ededa021225d7d04157a64e4bf022e63da76/hw/top_earlgrey/rtl/autogen/chip_earlgrey_asic.sv#L99
but for Darjeeling it is
https://github.com/lowRISC/opentitan/blob/6289ededa021225d7d04157a64e4bf022e63da76/hw/top_darjeeling/rtl/autogen/chip_darjeeling_asic.sv#L122

CC @Razer6

hw/ip_templates/pinmux/data/pinmux.hjson.tpl
```diff
@@ -572,19 +572,25 @@
...
+                      resval: 0
                     },
                     { bits: "2",
                       name: "pull_en",
                       desc: "Enable pull-up or pull-down resistor."
+                      // Note: The special index (30) where this has a different resval depends on
+                      // the TargetCfg parameter that gets passed in. That, in turn, is configured
+                      // at chip level and it would be nicer if we could resolve it as we expanded
+                      // this template.
+                      resval: (multireg_idx == 30)
```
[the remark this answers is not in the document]
rswarbrick: Ah, that's a bit sad. On the plus side, we can probably cheat: because pinmux is already templated, we can add the index as a parameter that gets used in the expansion. I'll pick that up today.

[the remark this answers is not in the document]
rswarbrick: Following up on this slightly:

Ok, I've finally understood properly :-) This seems to be (yet) another example of where SystemVerilog parameters don't really play very well with the reggen-style hjson flow.

The top-level configuration passes a `TargetCfg` parameter and one field of that parameter gives the index of the mio that should be used for strap0. The code in `pinmux.sv` then uses this to set some element of the constant `mio_pad_attr_q` array which, in turn, sets an element of the constant `hw2reg.mio_pad_attr`, affecting the reset value of some register.

This basically gives a path from a parameter to something that reggen cares about, which causes problems.

The laziest approach is probably:
- Add a "tap_strap0_idx" field to the pinmux Mako template
- Use that in `pinmux.hjson.tpl` instead of the literal 30
- Add a check in `pinmux.sv.tpl` to make sure that the parameter that comes in from the instantiation site matches the value that comes in through templating.

hw/top_darjeeling/ip_autogen/pinmux/data/pinmux.hjson
rswarbrick: I think the version I just force-pushed should sort this out.

hw/ip_templates/pinmux/data/pinmux.hjson.tpl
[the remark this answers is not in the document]
rswarbrick: (Lazy approach now done! :-D)

hw/top_darjeeling/ip_autogen/pinmux/data/pinmux.hjson
```diff
@@ -260,19 +260,27 @@
...
                     },
                     { bits: "1",
                       name: "virtual_od_en",
                       desc: "Enable virtual open drain."
+                      resval: 0
                     },
                     { bits: "2",
                       name: "pull_en",
                       desc: "Enable pull-up or pull-down resistor."
+                      // Note: The special index where this has a different resval depends on the
```
Razer6, line 277: The config should be an Earlgrey-only conifg and not render in Darjeelings pinmux instance.

util/topgen.py
```diff
@@ -480,10 +480,17 @@ def _get_pinmux_params(top: ConfigT) -> ParamsT:
                   pinmux["io_counts"]["dedicated"]["inputs"] +
                   pinmux["io_counts"]["dedicated"]["outputs"])
 
+    tap_strap0_idx = pinmux.get('tap_strap0_idx', 0)
+
     # Generation with zero MIO/DIO pads is currently not supported.
     assert (n_mio_pads > 0)
     assert (n_dio_pads > 0)
 
+    if tap_strap0_idx >= n_mio_pads:
```
Razer6, line 489: This is now pulling in an Earlgrey specific thing into the tooling. We should find a way to avoid that and provide the necessary information from the HJSONs and maybe ipgen parameters.

Razer6 (changes requested):

hw/top_darjeeling/ip_autogen/pinmux/data/pinmux.hjson
rswarbrick: Are you saying that there is no such pad in Darjeeling? If so, that's fine. We can just set that index to `-1`, I should think. But the current RTL does not do that!

util/topgen.py
rswarbrick: Hmm, I think this has the same response as the previous note. I suspect the right thing to do is to set `tap_strap0_idx` to `-1` for Darjeeling (and then this code is completely reasonable, although it would be worth me adding a couple of documentation comments to explain what's going on)

Is that right?

hw/top_darjeeling/ip_autogen/pinmux/data/pinmux.hjson
Razer6: Yes, that is correct. Straps are very integrator-specific. The EG/Nuvoton design is one integration, which needs that but that does not say anything about other designs. In fact DJ does not need since the TAP is outside of the top. even EG has efforts for changing that.

More importantly, this impacts any other TOP that uses a pinmux. They need to set an Earlgrey specific customization that is completely unrelated to them. Further, pulling EG-specific code into topgen is worse. It is topgen and not EGgen. How would you imagine this feature being used by someone else? Do they need to modify topgen for their variables?  

I think the right way moving forward is to pull in __all__ pinmux indices to ipgen such that the reserved value computation can access them in a clean fashion without adding a special case for one specific pin that only relates to one TOP.

rswarbrick: That makes sense. I'll set the default value to -1 (i.e. "not used") and override it for EarlGrey. I agree that we could definitely do this more cleanly, but that needn't gate the improvement that (I hope) is in this PR.

hw/ip_templates/pinmux/data/pinmux.hjson.tpl
```diff
@@ -572,19 +572,27 @@
...
                     { bits: "2",
                       name: "pull_en",
                       desc: "Enable pull-up or pull-down resistor."
+                      // Note: The special index where this has a different resval depends on the
+                      // TargetCfg parameter that gets passed in. That, in turn, is configured at
+                      // chip level and it would be nicer if we could resolve it as we expanded this
+                      // template. Instead, we expect the top that renders the template to give
+                      // matching parameters and we include a basic check in pinmux.sv.tpl to make
+                      // sure everything is in sync.
+                      resval: (multireg_idx == ${tap_strap0_idx})
```
Razer6, line 595: I think this code should only be rendered if `tap_strap0_idx ` > 0

util/topgen.py
Razer6: Can this check be moved to the parametrized pinmux template?

hw/ip_templates/pinmux/data/pinmux.hjson.tpl
rswarbrick: Honestly, I'm really unconvinced that this PR has the "right way" to solve the problem! I think the better approach is probably a bodge on the UVM side (where we squirt in some information to tell the register model the correct resval to expect)

I'm going on holiday next week, but intend to come back to this and make the change more carefully. That's part of the reason that I factored out the "obviously right" bit into #27289.

rswarbrick:
I'm closing this because it has rather bit-rotted and we haven't had an immediate use for it. Some of the machinery is now merged, so we should be able to implement the specific example in future if we need it.

PR #24749, opened 2024-10-08
[topgen] Make toplevel SV and chiplevel templates top-sepcific

Razer6:
This PR makes the toplevel verilog and chiplevel template top-specific to better scale with multiple tops. This avoids adding multiple cases to the template dealing with top-specic funcitonality.


rswarbrick (approved):
This looks sensible to me. Amusingly, I spent several minutes trying to work out what the diff on the new files meant before I figured out they were symlinks...

Razer6:
Hehe. I thought a symlink was a better idea to keep them in sync :)

PR #24753, opened 2024-10-09
[darjeeling] Add Darjeeling chip/toplevel SV template

Razer6:
This copies over the Darjeeling templates but removes the casing between the different tops.

### DIFF
--- a/hw/top_darjeeling/templates/chiplevel.sv.tpl
+++ b/hw/top_darjeeling/templates/chiplevel.sv.tpl
@@ -1,0 +1,1543 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+${gencmd}
+<%
+import re
+import topgen.lib as lib
+from copy import deepcopy
+
+# Provide shortcuts for some commonly used variables
+pinmux = top['pinmux']
+pinout = top['pinout']
+
+num_mio_inputs = pinmux['io_counts']['muxed']['inouts'] + \
+                 pinmux['io_counts']['muxed']['inputs']
+num_mio_outputs = pinmux['io_counts']['muxed']['inouts'] + \
+                  pinmux['io_counts']['muxed']['outputs']
+num_mio_pads = pinmux['io_counts']['muxed']['pads']
+
+num_dio_inputs = pinmux['io_counts']['dedicated']['inouts'] + \
+                 pinmux['io_counts']['dedicated']['inputs']
+num_dio_outputs = pinmux['io_counts']['dedicated']['inouts'] + \
+                  pinmux['io_counts']['dedicated']['outputs']
+num_dio_total = pinmux['io_counts']['dedicated']['inouts'] + \
+                pinmux['io_counts']['dedicated']['inputs'] + \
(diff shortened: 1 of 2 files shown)


andreaskurth (commented):
Thanks for this implementation proposal of Darjeeling's chip- and top-level templates, @Razer6.

I agree that the chip-level template of Darjeeling as integrated RoT can and should be implemented separately from that of Earlgrey as a discrete RoT. IMO this is supported by the diff of `hw/top_darjeeling/templates/chiplevel.sv.tpl` (proposed here) to `util/topgen/templates/chiplevel.sv.tpl` (which we probably want to move to `hw/top_earlgrey/templates/chiplevel.sv.tpl` if we accept this approach): there is a significant amount of different code, so a common template file would need many and complex case distinctions. I'd be interested to hear @a-will's opinion on this (including, but not limited to, the emulation perspective).

I don't think the same holds for the top-level template, however: With `topgen` and the data structures it uses and processes, one SV template IMO can and should suffice to generate the top-level SV code, IMO this is supported by the diff between ` hw/top_darjeeling/templates/toplevel.sv.tpl` (proposed in this PR) and `util/topgen/templates/toplevel.sv.tpl` -- which has 20-ish lines -- as well as the Git diff of `util/topgen/templates/toplevel.sv.tpl` between `master` and `integrated_dev` -- which also has 20-ish lines. I'd be interested to hear @matutem's and @hcallahan-lowrisc's opinion (also from a `topgen` perspective) and @vogelpi's opinion on this.

a-will:
> I don't think the same holds for the top-level template, however: With topgen and the data structures it uses and processes, one SV template IMO can and should suffice to generate the top-level SV code, IMO this is supported by the diff between  hw/top_darjeeling/templates/toplevel.sv.tpl (proposed in this PR) and util/topgen/templates/toplevel.sv.tpl -- which has 20-ish lines -- as well as the Git diff of util/topgen/templates/toplevel.sv.tpl between master and integrated_dev -- which also has 20-ish lines. I'd be interested to hear @matutem's and @hcallahan-lowrisc's opinion (also from a topgen perspective) and @vogelpi's opinion on this.

The fact that there _are_ differences would suggest to me that they might need to be independent. If we can come up with all the ways that the toplevel.sv might need to be generated using documented toplevel.hjson descriptions, then toplevel.sv.tpl can be shared. However, if there needs to be any `if toplevel is "mytop":`, shared templates generally foreclose out-of-tree support (i.e. the only permitted top_* configurations are the ones known to the opentitan tree). If there needs to be some top-specific stitching, then I'd suggest converting the current code into a library of generation functions and allowing a top-specific template to do the work.

We don't have to necessarily get to out-of-tree support now, though... However, we might find it easier to start with top-specific templates, then move common code back into topgen.

Razer6:
> (which we probably want to move to `hw/top_earlgrey/templates/chiplevel.sv.tpl` if we accept this approach)

That already happened on master.

On the top-level template, I agree, the diff is more minor. However, there is one. As summarized by Alex, a common template makes out-of-tree tops much harder. IMO, a split template gives the most flexibility also in the sense of not breaking backward compatibility with older tops such as Earlgrey. If templates are really the same, they can be symlinked. For example, English Breakfast's templates are symlinked to Earlgrey.

andreaskurth:
OK, SGTM as also discussed offline!

andreaskurth (approved):

PR #24755, opened 2024-10-09
[sival, keymgr] State-agnostic chip_sw_key_derivation_test

andrea-caforio:
`keymgr_key_derivation_test_fpga_cw310_sival_rom_ext test` was marked as broken in #21706. This commit fixes this and addresses #21500.

The source of the bug is due to the nature of the ROM_EXT that runs through a CDI attestation flow advancing the key manager state to "OwnerRootKey" in the process. This was not accounted for in the test, which assumed an active "CreatorRootKey" state (this works for sim_dv tests but not with ROM_EXT). The solution to this issue lies in rendering the test state-agnostic, meaning it will execute its verifications independently of the current key manager state. This approach thus mimics the `keymgr_sideload_{aes,kmac,otbn}` tests which follow a similar strategy.


sw/device/tests/keymgr_key_derivation_test.c
```diff
@@ -2,64 +2,63 @@
...
 #include "sw/device/lib/testing/keymgr_testutils.h"
-#include "sw/device/lib/testing/kmac_testutils.h"
-#include "sw/device/lib/testing/rstmgr_testutils.h"
-#include "sw/device/lib/testing/test_framework/check.h"
 #include "sw/device/lib/testing/test_framework/ottf_main.h"
 
-#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
-#include "keymgr_regs.h"  // Generated.
-#include "kmac_regs.h"    // Generated.
-
```
vogelpi, line 26: I think not all of these includes can be removed as a result of the changes in the test. Some could have already been removed before. If this is true, I would suggest factoring these removal out into a separate commit.

vogelpi (approved):
Thanks @andrea-caforio for working on this! The PR looks good, except that I would factor out the removal of unneeded includes into a separate commit. The reason is that even if the test code changes would need reverting in the future, we would want to keep the unneeded includes removed.

But you don't need to do this right now, I think this can be included with potential feedback from the software team. It would be great to also get this approved by e.g. @engdoreis (OoO today) or maybe @jwnrt . I'll ask him.

jwnrt (approved):
This looks great, thanks @andrea-caforio

sw/device/tests/keymgr_key_derivation_test.c
andrea-caforio: Indeed, all the removed includes could have been removed before the changes of this commit. I think they were just copy-pasted from another file when the test was initially written because `keymgr_sideload_{aes,kmac,otbn}_test.c` have almost the same set unused imports. I split the removal into a separate commit.

andrea-caforio:
Thank you for your reviews @vogelpi and @jwnrt. I split the commit into two parts to make the removal of the unused includes explicit.

jwnrt:
Thanks, I'll merge once CI has finished

Issue #24752, opened 2024-10-09
[rom_ctrl, dv] Uncovered conditional coverage in tlul_adapter_reg.sv


KinzaQamar:
Hi,

According to 2nd oct nightly regression coverage dashboard [here](https://reports.opentitan.org/hw/ip/rom_ctrl_64kB/dv/latest/cov_report/mod64.html#Cond) stating that a conditional coverage for [L208](https://github.com/lowRISC/opentitan/blob/29d22a60a272d26f8987bc39fa4ba0794f02a0e8/hw/ip/tlul/rtl/tlul_adapter_reg.sv#L208C72-L208C83) in tlul_adapter_reg.sv is uncovered. The tool never seen 0, 0, 0, 1, 0 and 0, 1, 0, 0, 0 transitions. That makes sense as **malformed_meta_err**, **tl_err** and **instr_error**  are dependent over instruction type being invalid. Which corresponds that if there is a malformed_meta_err there would be a  **tl_err** and **instr_error**. Hence, 01000 is a non-occuring case.
```             <Non-ocurring case> [tlul_adapter_reg.sv L208]
| addr_align_err | malformed_meta_err | tl_err | instr_error | intg_error
|  ------------- |  ----------------- |  ----- | ----------- | ---------
|        0       |          1         |   0    |      0      |     0
```
Regarding the next case: 
```             <uncovered case> [tlul_adapter_reg.sv L208]
| addr_align_err | malformed_meta_err | tl_err | instr_error | intg_error
|  ------------- |  ----------------- |  ----- | ----------- | ---------
|        0       |          0         |   0    |      1      |     0
```
This could only be occuring if instruction type is valid and instruction fetch is disabled and tl_i.a_opcode is **GET**
### Referencing files

- Case (01000)
1.  [tlul_adapter_reg](https://github.com/lowRISC/opentitan/blob/29d22a60a272d26f8987bc39fa4ba0794f02a0e8/hw/ip/tlul/rtl/tlul_adapter_reg.sv#L204C1-L211C58)
[...]

PR #24757, opened 2024-10-09
[tlul,rtl] Remove some unused error signals

rswarbrick:
This was found by @KinzaQamar when looking at conditional coverage in `rom_ctrl` (leading to #24752). Rather than messing around with waivers, we can simplify things a bit in the RTL.

### DIFF
--- a/hw/ip/tlul/rtl/tlul_adapter_reg.sv
+++ b/hw/ip/tlul/rtl/tlul_adapter_reg.sv
@@ -65,7 +65,6 @@ module tlul_adapter_reg
   logic             error_q, error, err_internal, instr_error, intg_error;
 
   logic addr_align_err;     // Size and alignment
-  logic malformed_meta_err; // User signal format error or unsupported
   logic tl_err;             // Common TL-UL error checker
 
   logic [IW-1:0]  reqid_q;
@@ -205,10 +204,7 @@ module tlul_adapter_reg
                        (prim_mubi_pkg::mubi4_test_true_strict(tl_i.a_user.instr_type) &
                         prim_mubi_pkg::mubi4_test_false_loose(en_ifetch_i));
 
-  assign err_internal = addr_align_err | malformed_meta_err | tl_err | instr_error | intg_error;
-
-  // Don't allow unsupported values.
-  assign malformed_meta_err = tl_a_user_chk(tl_i.a_user);
+  assign err_internal = addr_align_err | tl_err | instr_error | intg_error;
 
   // addr_align_err
   //    Raised if addr isn't aligned with the size
--- a/hw/ip/tlul/rtl/tlul_adapter_reg.sv
+++ b/hw/ip/tlul/rtl/tlul_adapter_reg.sv
@@ -199,10 +199,8 @@ module tlul_adapter_reg
   ////////////////////
 
   // An instruction type transaction is only valid if en_ifetch is enabled
-  // If the instruction type is completely invalid, also considered an instruction error
-  assign instr_error = prim_mubi_pkg::mubi4_test_invalid(tl_i.a_user.instr_type) |
-                       (prim_mubi_pkg::mubi4_test_true_strict(tl_i.a_user.instr_type) &
-                        prim_mubi_pkg::mubi4_test_false_loose(en_ifetch_i));
+  assign instr_error = prim_mubi_pkg::mubi4_test_true_strict(tl_i.a_user.instr_type) &
+                       prim_mubi_pkg::mubi4_test_false_loose(en_ifetch_i);
 
   assign err_internal = addr_align_err | tl_err | instr_error | intg_error;
 


hw/ip/tlul/rtl/tlul_adapter_reg.sv
```diff
@@ -200,15 +199,10 @@ module tlul_adapter_reg
...
-  assign instr_error = prim_mubi_pkg::mubi4_test_invalid(tl_i.a_user.instr_type) |
-                       (prim_mubi_pkg::mubi4_test_true_strict(tl_i.a_user.instr_type) &
-                        prim_mubi_pkg::mubi4_test_false_loose(en_ifetch_i));
+  assign instr_error = prim_mubi_pkg::mubi4_test_true_strict(tl_i.a_user.instr_type) &
+                       prim_mubi_pkg::mubi4_test_false_loose(en_ifetch_i);
 
-  assign err_internal = addr_align_err | malformed_meta_err | tl_err | instr_error | intg_error;
-
-  // Don't allow unsupported values.
-  assign malformed_meta_err = tl_a_user_chk(tl_i.a_user);
```
GregAC, line 211: `tl_a_user_chk` is an empty function so this seems reasonable.

Worrying it's a truly empty function doesn't actually define a return value! Guess we've lucked out in that that's been 0 by default so this hasn't caused us issues but a good reason to just drop its use entirely. I'm be tempted to just remove it from `tlul_pkg`. There's a couple of other usages 

https://github.com/lowRISC/opentitan/blob/6c15d058ca9ea0dac5f05b4dbdac125f9ab1249c/hw/dv/sv/sim_sram/tlul_sink.sv#L65
https://github.com/lowRISC/opentitan/blob/6c15d058ca9ea0dac5f05b4dbdac125f9ab1249c/hw/ip/tlul/rtl/tlul_adapter_dmi.sv#L170

that would need addressing too.

hw/ip/tlul/rtl/tlul_adapter_reg.sv
```diff
@@ -200,15 +199,10 @@ module tlul_adapter_reg
   ////////////////////
 
   // An instruction type transaction is only valid if en_ifetch is enabled
-  // If the instruction type is completely invalid, also considered an instruction error
-  assign instr_error = prim_mubi_pkg::mubi4_test_invalid(tl_i.a_user.instr_type) |
-                       (prim_mubi_pkg::mubi4_test_true_strict(tl_i.a_user.instr_type) &
-                        prim_mubi_pkg::mubi4_test_false_loose(en_ifetch_i));
+  assign instr_error = prim_mubi_pkg::mubi4_test_true_strict(tl_i.a_user.instr_type) &
+                       prim_mubi_pkg::mubi4_test_false_loose(en_ifetch_i);
```
GregAC, line 203: This alters the security properties so I'm not sure its sensible.

In particular if `tl_i.a_user.instr_type` has a fault injected in it (so it's neither a proper false nor a proper true value) we won't raise an error where previously we did.

Perhaps the checking for invalid values here is more than we need though, the injected fault should be picked up by the ECC check anyway and we check this again at the SRAM end anyway (instruction fetch only connects to ROM, Debug Module, SRAM and Flash and it's only SRAM we care about the fetch enable for).

One for @vogelpi to weigh in on.

GregAC:
Don't want to approve due to the security property change but would be happy to approve a modified version that just addressed the `malformed_meta_err` signal. Other change may also be reasonable but I'd like feedback from others first.

hw/ip/tlul/rtl/tlul_adapter_reg.sv
rswarbrick: I don't think I understand this comment properly. This is the function:
```systemverilog
  // Check user for unsupported values
  function automatic logic tl_a_user_chk(tl_a_user_t user);
    logic malformed_err;
    logic unused_user;
    unused_user = |user;
    malformed_err = prim_mubi_pkg::mubi4_test_invalid(user.instr_type);
    return malformed_err;
  endfunction // tl_a_user_chk
```
(and it has a return value...)

hw/ip/tlul/rtl/tlul_adapter_reg.sv
rswarbrick: If that happens then `tl_err` will also be true (which means we can't hit a conditional coverage point). I tried to explain this in the commit message, but I'm pretty certain there is no change to behaviour from the commit.

hw/ip/tlul/rtl/tlul_adapter_reg.sv
GregAC: Ah I blame github's codesearch UI that cuts lines out without making that particularly obvious:

GregAC (approved):
Apologies hadn't read the commit messages properly and hadn't realised `tl_err` is repeating the check. This change LGTM

rswarbrick:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_adapter_reg.sv

PR #24761, opened 2024-10-09
[dv, dvsim] stability fixes for the nc job scheduler

venkatk-ot:
Improve stability of batch runs using the nc job scheduler by specifying minimum cpu and ram requiements (CPU, CORES) and an option to skip directory validity check (-D) to prevent triggering false alarms about directory not accessible by all machines in the cluster.


rswarbrick (approved):
This looks sensible to me. I guess it's a bit sad to now be encoding an 8GiB RAM / 2 core dependency in the script. But I'd be very surprised if it ever caused a problem for anyone :-)

venkatk-ot:
Thanks. Perhaps in the future we can add an option in dvsim / hjson test list to specify min cpu and ram. When running simulations on cloud VMs, sometimes jobs land on 1 cpu 512MB ram VMs if this setting is not specified

Issue #15530, opened 2022-10-17
[kmac] EDN doesn't timeout on FPGA


ballifatih:
With FPGA, I cannot trigger timeout for KMAC's EDN handshake.

For example, I choose the smallest waiting time configuration for KMAC's EDN by setting entropy_prescaler = 0x00 and entropy_wait_timer=0x01, but EDN response still doesn't timeout. This issue is only on the FPGA side, and there is not an issue with verilator and dvsim.

The related chip_sw test is https://github.com/lowRISC/opentitan/pull/14637, and CI fails because of this FPGA test.

ballifatih:
With FPGA, I cannot trigger timeout for KMAC's EDN handshake. 

For example, I choose the smallest waiting time configuration for KMAC's EDN by setting `entropy_prescaler = 0x00` and `entropy_wait_timer=0x01`, but EDN response still doesn't timeout. This issue is only on the FPGA side, and there is not an issue with verilator and dvsim.

The related chip_sw test is #14637, and CI fails because of this FPGA test.

ballifatih:
I also tried this with a larger number of KMAC invocations (~100) as an experiment, and none of them times out. 

https://github.com/ballifatih/opentitan/blob/edn_timeout_fpga/sw/device/tests/kmac_entropy_test.c

msfschaffner:
CC @tjaychen @alphan @moidx @weicaiyang @cindychip 

I am not certain, but may be this is because of different clock ratios between req/ack domains on the FPGA.

Given that the DV test works I move this into M3 for now, but please LMK if you think this needs to be addressed as part of M2.

tjaychen:
i think it's fine to table. As long as we have a good feel for this on simulation for now.
The clock ratios on FPGA should actually maintain between edn and kmac.  However, in fpga, the entropy source i think is way faster (need to confirm this), so it just might be that by the time you request everything is already pre-generated...

alphan:
> i think it's fine to table. As long as we have a good feel for this on simulation for now. The clock ratios on FPGA should actually maintain between edn and kmac. However, in fpga, the entropy source i think is way faster (need to confirm this), so it just might be that by the time you request everything is already pre-generated...

Makes sense, thanks for the possible explanation @tjaychen! I also think that it's OK to move this into M3.

ballifatih:
Cross referencing an earlier (closed) issue that somewhat touched the same problem #14638, in case the discussions there might also be useful for later.

[...]

PR #24765, opened 2024-10-10
[sival,kmac] Add kmac_error_conditions_test

nasahlpa:
This adds the new `kmac_error_conditions_test` test that is defined in the `chip_sw_kmac_error_conditions` chip level test.

### DIFF
--- a/sw/device/tests/kmac_error_conditions_test.c
+++ b/sw/device/tests/kmac_error_conditions_test.c
@@ -1,0 +1,710 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/device/lib/base/mmio.h"
+#include "sw/device/lib/dif/dif_clkmgr.h"
+#include "sw/device/lib/dif/dif_keymgr.h"
+#include "sw/device/lib/dif/dif_kmac.h"
+#include "sw/device/lib/runtime/log.h"
+#include "sw/device/lib/testing/keymgr_testutils.h"
+#include "sw/device/lib/testing/kmac_testutils.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+#include "keymgr_regs.h"  // Generated.
+#include "kmac_regs.h"    // Generated.
+
+static dif_kmac_t kmac;
+static dif_keymgr_t keymgr;
+
+OTTF_DEFINE_TEST_CONFIG();
+
+/**
+ * KMAC test description.
+ */
+typedef struct kmac_test {
+  dif_kmac_mode_kmac_t mode;
+  dif_kmac_key_t key;
+
+  const char *message;
+  size_t message_len;
+
+  const char *customization_string;
+  size_t customization_string_len;
+
+  const uint32_t digest[16];
+  size_t digest_len;
+  bool digest_len_is_fixed;
+} kmac_test_t;
+
+/**
+ * A single KMAC example taken from:
+ * https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/KMAC_samples.pdf
+ */
+static const kmac_test_t kKmacTestVector = {
+    .mode = kDifKmacModeKmacLen256,
+    .key =
+        (dif_kmac_key_t){
+            .share0 = {0x43424140, 0x47464544, 0x4b4a4948, 0x4f4e4f4c,
+                       0x53525150, 0x57565554, 0x5b5a5958, 0x5f5e5d5c},
+            .share1 = {0},
+            .length = kDifKmacKeyLen256,
+        },
+    .message =
+        "\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0a\x0b\x0c\x0d\x0e\x0f"
+        "\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f"
+        "\x20\x21\x22\x23\x24\x25\x26\x27\x28\x29\x2a\x2b\x2c\x2d\x2e\x2f"
+        "\x30\x31\x32\x33\x34\x35\x36\x37\x38\x39\x3a\x3b\x3c\x3d\x3e\x3f"
+        "\x40\x41\x42\x43\x44\x45\x46\x47\x48\x49\x4a\x4b\x4c\x4d\x4e\x4f"
+        "\x50\x51\x52\x53\x54\x55\x56\x57\x58\x59\x5a\x5b\x5c\x5d\x5e\x5f"
+        "\x60\x61\x62\x63\x64\x65\x66\x67\x68\x69\x6a\x6b\x6c\x6d\x6e\x6f"
+        "\x70\x71\x72\x73\x74\x75\x76\x77\x78\x79\x7a\x7b\x7c\x7d\x7e\x7f"
+        "\x80\x81\x82\x83\x84\x85\x86\x87\x88\x89\x8a\x8b\x8c\x8d\x8e\x8f"
+        "\x90\x91\x92\x93\x94\x95\x96\x97\x98\x99\x9a\x9b\x9c\x9d\x9e\x9f"
+        "\xa0\xa1\xa2\xa3\xa4\xa5\xa6\xa7\xa8\xa9\xaa\xab\xac\xad\xae\xaf"
+        "\xb0\xb1\xb2\xb3\xb4\xb5\xb6\xb7\xb8\xb9\xba\xbb\xbc\xbd\xbe\xbf"
+        "\xc0\xc1\xc2\xc3\xc4\xc5\xc6\xc7",
+    .message_len = 200,
+    .customization_string = "My Tagged Application",
+    .customization_string_len = 21,
+    .digest = {0x1c73bed5, 0x73d74e95, 0x59bb4628, 0xe3a8e3db, 0x7ae7830f,
+               0x5944ff4b, 0xb4c2f1f2, 0xceb8ebec, 0xc601ba67, 0x57b88a2e,
+               0x9b492d8d, 0x6727bbd1, 0x90117868, 0x6a300a02, 0x1d28de97,
+               0x5d3030cc},
+    .digest_len = 16,
+    .digest_len_is_fixed = false,
+};
+
+/**
+ * Init key manager.
+ *
+ * @return OK or error.
+ */
+status_t setup_keymgr(void) {
+  // Initialize keymgr and advance to the state specified by the loaded ROM.
+  TRY(keymgr_testutils_initialize(&keymgr, &kmac));
+
+  return OK_STATUS();
+}
+
+/**
+ * Poll the SHA3_IDLE bit.
+ *
+ * @param kmac The KMAC handle.
+ */
+void poll_sha3_idle_bit(const dif_kmac_t *kmac) {
+  uint32_t reg;
+  do {
+    reg = mmio_region_read32(kmac->base_addr, KMAC_STATUS_REG_OFFSET);
+  } while (!bitfield_bit32_read(reg, KMAC_STATUS_SHA3_IDLE_BIT));
+}
+
+/**
+ * Poll the KMAC_DONE bit.
+ *
+ * @param kmac The KMAC handle.
+ */
+void poll_kmac_done_bit(const dif_kmac_t *kmac) {
+  uint32_t reg;
+  do {
+    reg = mmio_region_read32(kmac->base_addr, KMAC_INTR_STATE_REG_OFFSET);
+  } while (!bitfield_bit32_read(reg, KMAC_INTR_STATE_KMAC_DONE_BIT));
+}
+
+/**
+ * Check and clear the KMAC error.
+ *
+ * Performs the following:
+ * - Check whether error interrupt was seen.
+ * - Check whether the expected error was seen.
+ * - Clears the error interrupt.
+ * - Sets the error processed bit.
+ * - Reset the KMAC IP.
+ * - Waits until KMAC again gets ready.
+ * - Waits until the KMAC_DONE bit is seen.
+ *
+ * @param kmac The KMAC handle.
+ * @param exp_err The expected error.
+ * @param wait_done Wait for kmac_done bit. Not needed for tests that don't
+ * start KMAC.
+ * @return OK or error.
+ */
+status_t check_and_clear_error(const dif_kmac_t *kmac, dif_kmac_error_t exp_err,
+                               bool wait_done) {
+  dif_kmac_error_t err;
+  dif_kmac_operation_state_t kmac_operation_state;
+  uint32_t info;
+  bool error;
+
+  // Check whether the KMAC ERR bit was set in the interrupt state register.
+  TRY(dif_kmac_has_error_occurred(kmac, &error));
+  TRY_CHECK(error, "No error was triggered.");
+
+  TRY(dif_kmac_get_error(kmac, &err, &info));
+  TRY_CHECK(err == exp_err, "Unexpected error was triggered.");
+
+  // Clear the error IRQ.
+  TRY(dif_kmac_clear_err_irq(kmac));
+  // Set err_processed bit.
+  TRY(dif_kmac_err_processed(kmac));
+  // Reset the KMAC IP.
+  TRY(dif_kmac_reset(kmac, &kmac_operation_state));
+
+  // Wait until sha3_idle is again set.
+  poll_sha3_idle_bit(kmac);
+
+  // Wait until KMAC flags the done bit.
+  if (wait_done) {
+    poll_kmac_done_bit(kmac);
+  }
+
+  return OK_STATUS();
+}
+
+/**
+ * ErrWaitTimerExpired.
+ *
+ * This test checks whether the ErrWaitTimerExpired error code is
+ * returned by the hardware when KMAC timed out while waiting for EDN entropy.
+ * By configuring a small wait timer (=1), the timeout immediately will be
+ * triggered and we should see the expected error code.
+ *
+ * @return OK or error.
+ */
+status_t test_err_wait_timer_expired(void) {
+  LOG_INFO("Testing ErrWaitTimerExpired error.");
+
+  // Init the KMAC block.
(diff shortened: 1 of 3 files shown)


nasahlpa:
@vogelpi the test for the `ErrWaitTimerExpired` and `ErrIncorrectEntropyMode` is not working on the CW310 FPGA as we turn by default the masks off. I've marked the test as broken and referenced to #15530. Is this OK for you?

nasahlpa:
Thanks for your review, @engdoreis! As suggested, I've changed `CHECK_DIF_OK` to `TRY`.

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,447 @@
...
+}
+
+/**
+ * ErrorSoftwareCommandSequence.
+ *
+ * This test checks whether the kDifErrorSoftwareCommandSequence error code is
+ * returned by the hardware when sending software commands in a wrong order to
+ * the KMAC block. Note that this test is not exhaustive, i.e., the test does
+ * not trying to reach the ErrorSoftwareCommandSequence error state from each
+ * other state.
```
vogelpi, line 222: Thanks for the comment, I think this is okay but maybe it would be good for someone from the SiVal team to confirm @luismarques , @engdoreis ?

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,447 @@
...
+
+/**
+ * ErrSwPushedMsgFifo.
+ *
+ * This test checks whether the ErrSwPushedMsgFifo error code is
+ * returned by the hardware when writing to the message FIFO without first
+ * sending a START command. This is different to the test description in the
+ * test plan as writing to the message FIFO during a keymanager app request is
+ * not easy to implement. However, this test still checks whether the correct
+ * error code is generated.
```
vogelpi, line 345: Would this better be tested in a keymgr test?

I think we should test this somehow because it seems important to me to signal errors in this case. What is your view?

vogelpi (commented):
Thanks for the PR @nasahlpa !

The test implementation looks mostly good.

Can you please elaborate on
> ErrSwIssuedCmdInAppActive could not be implemented in SW as triggering a SW command while the app interface is active is not possible.

There are two error codes which we can't test related to this limitation. It would be good understand what the underlying issue is. I we agree to not test this, we should update the testplan accordingly.

Checking `kmac_pkg.sv` I noted there are some other currently untested error codes that we should actually be able to test:
- `ErrSwHashingWithoutEntropyReady` - triggering a hash operation without setting the entropy mode first. This probably doesn't work in ROM_EXT though.
- `ErrShadowRegUpdate` - triggering a shadow register update error. This should be super straight forward.

The remaining three are related to FI countermeasures and escalations. We probably can't test this using this test anyway.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: I've rewritten the test in order to comply to the test plan.

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,447 @@
...
+  size_t customization_string_len;
+
+  const uint32_t digest[16];
+  size_t digest_len;
+  bool digest_len_is_fixed;
+} kmac_test_t;
+
+/**
+ * A single KMAC example.
+ */
```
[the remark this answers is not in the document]
nasahlpa: I've added a reference where this example was taken from. It is just an example, no particular meaning behind it - I just need a valid config / inputs to trigger an operation.

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+};
+
+/**
+ * Init key manager.
+ *
+ * @return OK or error.
+ */
+status_t setup_keymgr(void) {
+  // Initialize keymgr and advance to the state specified by the loaded ROM.
+  CHECK_STATUS_OK(keymgr_testutils_initialize(&keymgr, &kmac));
```
engdoreis, line 87:
```suggestion
  TRY(keymgr_testutils_initialize(&keymgr, &kmac));
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+status_t check_and_clear_error(const dif_kmac_t *kmac, dif_kmac_error_t exp_err,
+                               bool wait_idle) {
+  dif_kmac_error_t err;
+  dif_kmac_operation_state_t kmac_operation_state;
+  uint32_t info;
+  bool error;
+
+  // Check whether the KMAC ERR bit was set in the interrupt state register.
+  TRY(dif_kmac_has_error_occurred(kmac, &error));
+  CHECK(error, "No error was triggered.");
```
engdoreis, line 128: This applies throughout this file.
```suggestion
  TRY_CHECK(error, "No error was triggered.");
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+ * @return OK or error.
+ */
+status_t test_err_shadow_reg_update(void) {
+  LOG_INFO("Testing shadow register update error.");
+
+  // Init the KMAC block.
+  TRY(dif_kmac_init(mmio_region_from_addr(TOP_EARLGREY_KMAC_BASE_ADDR), &kmac));
+
+  // Configure KMAC hardware.
+  CHECK_STATUS_OK(kmac_testutils_config(&kmac, false));
```
engdoreis, line 165:
```suggestion
  TRY(kmac_testutils_config(&kmac, false));
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+                      1 << KEYMGR_START_EN_BIT);
+  // Do not wait until operation has finished. Directly issue the command to
+  // trigger the ErrSwIssuedCmdInAppActive error.
+  mmio_region_write32(kmac.base_addr, KMAC_CMD_REG_OFFSET, cmd_start);
+
+  // Check whether we got the expected error.
+  check_and_clear_error(
+      &kmac, kDifErrorSoftwareIssuedCommandWhileAppInterfaceActive, true);
+
+  return OK_STATUS();
```
engdoreis, line 244:
```suggestion
  return check_and_clear_error(
      &kmac, kDifErrorSoftwareIssuedCommandWhileAppInterfaceActive, true);
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+      bitfield_bit32_write(cfg_reg, KMAC_CFG_SHADOWED_ENTROPY_READY_BIT, true);
+  cfg_reg = bitfield_bit32_write(cfg_reg, KMAC_CFG_SHADOWED_MSG_MASK_BIT,
+                                 kDifToggleEnabled);
+  mmio_region_write32_shadowed(kmac.base_addr, KMAC_CFG_SHADOWED_REG_OFFSET,
+                               cfg_reg);
+
+  // Check whether we got the expected error.
+  check_and_clear_error(&kmac, kDifErrorEntropyModeIncorrect, true);
+
+  return OK_STATUS();
```
engdoreis, line 376:
```suggestion
  return check_and_clear_error(&kmac, kDifErrorEntropyModeIncorrect, true);
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+  // provided SW key is ignored as we configured KMAC to use the sideloaded key.
+  dif_kmac_operation_state_t kmac_operation_state;
+  TRY(dif_kmac_mode_kmac_start(&kmac, &kmac_operation_state,
+                               kKmacTestVector.mode, 0, &kKmacTestVector.key,
+                               NULL));
+
+  // Check whether we got the expected error.
+  check_and_clear_error(&kmac, kDifErrorKeyNotValid, true);
+
+  return OK_STATUS();
```
engdoreis, line 430:
```suggestion
  return check_and_clear_error(&kmac, kDifErrorKeyNotValid, true);
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+  // Start the key manager operation.
+  mmio_region_write32(keymgr.base_addr, KEYMGR_START_REG_OFFSET,
+                      1 << KEYMGR_START_EN_BIT);
+  // Write to message FIFO while app interface operation is ongoing.
+  mmio_region_write8(kmac.base_addr, KMAC_MSG_FIFO_REG_OFFSET, 0xff);
+
+  // Check whether we got the expected error.
+  check_and_clear_error(&kmac, kDifErrorSoftwarePushedMessageFifo, true);
+
+  return OK_STATUS();
```
engdoreis, line 508:
```suggestion
  return check_and_clear_error(&kmac, kDifErrorSoftwarePushedMessageFifo, true);
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+
+  // Issue start command.
+  uint32_t cmd_reg =
+      bitfield_field32_write(0, KMAC_CMD_CMD_FIELD, KMAC_CMD_CMD_VALUE_START);
+  mmio_region_write32(kmac.base_addr, KMAC_CMD_REG_OFFSET, cmd_reg);
+
+  // Check whether we got the expected error.
+  check_and_clear_error(&kmac, kDifErrorIncorrectFunctionName, false);
+
+  return OK_STATUS();
```
engdoreis, line 564:
```suggestion
  return check_and_clear_error(&kmac, kDifErrorIncorrectFunctionName, false);
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+ *
+ * @return OK or error.
+ */
+status_t test_err_wait_timer_expired(void) {
+  LOG_INFO("Testing ErrWaitTimerExpired error.");
+
+  // Init the KMAC block.
+  TRY(dif_kmac_init(mmio_region_from_addr(TOP_EARLGREY_KMAC_BASE_ADDR), &kmac));
+
+  dif_kmac_config_t config = (dif_kmac_config_t){
```
engdoreis, line 263: Nit. this can be const

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+  LOG_INFO("Testing kDifErrorSoftwareCommandSequence error.");
+  uint32_t cmds[3] = {KMAC_CMD_CMD_VALUE_PROCESS, KMAC_CMD_CMD_VALUE_RUN,
+                      KMAC_CMD_CMD_VALUE_DONE};
+
+  for (int it = 0; it < ARRAYSIZE(cmds); it++) {
+    // Re-init KMAC for the test.
+    TRY(dif_kmac_init(mmio_region_from_addr(TOP_EARLGREY_KMAC_BASE_ADDR),
+                      &kmac));
+    // Configure KMAC hardware (using software key and software entropy).
+    CHECK_STATUS_OK(kmac_testutils_config(&kmac, false));
```
engdoreis, line 400:
```suggestion
    TRY(kmac_testutils_config(&kmac, false));
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+
+  return OK_STATUS();
+}
+
+status_t test_err_key_not_valid(void) {
+  LOG_INFO("Testing ErrKeyNotValid error.");
+  // Re-init KMAC for the test.
+  TRY(dif_kmac_init(mmio_region_from_addr(TOP_EARLGREY_KMAC_BASE_ADDR), &kmac));
+  // Configure KMAC to use the sideloaded key.
+  CHECK_STATUS_OK(kmac_testutils_config(&kmac, true));
```
engdoreis, line 418:
```suggestion
  TRY(kmac_testutils_config(&kmac, true));
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+  uint32_t misconfigured_strength[2] = {KMAC_CFG_SHADOWED_KSTRENGTH_VALUE_L128,
+                                        KMAC_CFG_SHADOWED_KSTRENGTH_VALUE_L224};
+  uint32_t mode[2] = {KMAC_CFG_SHADOWED_MODE_VALUE_SHA3,
+                      KMAC_CFG_SHADOWED_MODE_VALUE_SHAKE};
+
+  for (int it = 0; it < ARRAYSIZE(mode); it++) {
+    // Re-init KMAC for the test.
+    TRY(dif_kmac_init(mmio_region_from_addr(TOP_EARLGREY_KMAC_BASE_ADDR),
+                      &kmac));
+    CHECK_STATUS_OK(kmac_testutils_config(&kmac, false));
```
engdoreis, line 454:
```suggestion
    TRY(kmac_testutils_config(&kmac, false));
```

sw/device/tests/kmac_error_conditions_test.c
```diff
@@ -0,0 +1,583 @@
...
+ * returned by the hardware when using a wrong (i.e., other than "KMAC")
+ * function name when in KMAC mode.
+ *
+ * @return OK or error.
+ */
+status_t test_err_incorrect_fnc_name(void) {
+  LOG_INFO("Testing kDifErrorIncorrectFunctionName error.");
+  // Re-init KMAC for the test.
+  TRY(dif_kmac_init(mmio_region_from_addr(TOP_EARLGREY_KMAC_BASE_ADDR), &kmac));
+  CHECK_STATUS_OK(kmac_testutils_config(&kmac, false));
```
engdoreis, line 524:
```suggestion
  TRY(kmac_testutils_config(&kmac, false));
```

engdoreis (commented):
Thanks @nasahlpa for the PR, You are using the mmio API on the test code to change the registers directly, isn't it worth creating DIFs to abstracts these registers access

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: As far as I see, the difference between `CHECK` and `TRY_CHECK` is that the former aborts the test and the latter continues but returns a `status_t`? Wouldn't it be better to abort the test (so using `CHECK`)?

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Done, thanks.

nasahlpa:
Thanks for your review, @vogelpi and @engdoreis 

I've made quite some significant changes to this PR to address your review, @vogelpi:
- Executed the test on CW340 - this works now and the test also gets executed in CI. Thanks @jwnrt for setting this up.
- Worked on testing `ErrShadowRegUpdate`. I've noticed that this `ErrShadowRegUpdate` (and the corresponding `kDifErrorShadowRegisterUpdate`) is never used in HW and SW. Instead, according to the [documentation](https://opentitan.org/book/hw/ip/kmac/doc/interfaces.html#security-alerts), a shadow register update error is indicated by a recoverable alert. I've added a test to test whether the HW complies to the documentation. Do we want to remove `ErrShadowRegUpdate` and `kDifErrorShadowRegisterUpdate` from the HW and software in a separate PR?
- Implemented the `ErrSwIssuedCmdInAppActive` test.
- Implemented the `ErrSwPushedMsgFifo` test according to the test plan.
- Implemented the `ErrSwHashingWithoutEntropyReady` test.

@engdoreis I just have one question regarding the difference `CHECK` <-> `TRY_CHECK` (please see in the comment above). Regarding the DIFs, I don't think that we should add these register writes to the DIF as they are only required to trigger an error behavior. Adding them could cause some unintended misuse by others, hence, I prefer leaving it as it is.

sw/device/tests/kmac_error_conditions_test.c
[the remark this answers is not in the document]
vogelpi: This is perfectly fine, thanks @nasahlpa !

vogelpi:
This sounds all great, excellent work @nasahlpa !

About the `ErrShadowRegUpdate` error code: I see the corresponding alert condition is also signaled via STATUS register. So software will learn about having done something wrong, even if it's not via ERR_CODE register. And your test checks this which is very nice! I agree that we should remove `ErrShadowRegUpdate` and `kDifErrorShadowRegisterUpdate` from the HW and SW in a separate PR.

sw/device/tests/kmac_error_conditions_test.c
nasahlpa: Makes sense. Thanks for the explanation! Changed accordingly.

PR #24769, opened 2024-10-11
Add missing include to rv_plic template file

dpetrisko:
This provides `ASSERT definition that is otherwise missing

### DIFF
--- a/hw/ip_templates/rv_plic/rtl/rv_plic.sv.tpl
+++ b/hw/ip_templates/rv_plic/rtl/rv_plic.sv.tpl
@@ -14,6 +14,8 @@
 // Verilog parameter
 //   MAX_PRIO: Maximum value of interrupt priority
 
+`include "prim_assert.sv"
+
 module ${module_instance_name} import ${module_instance_name}_reg_pkg::*; #(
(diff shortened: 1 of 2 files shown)


rswarbrick (commented):
This looks right to me: thanks for tidying it up. I think you'll need to regenerate the earlgrey instance of the template though (which I think you can do with `make -C hw`)

rswarbrick:
Thanks for regenerating this. I think you'll probably want to squash the two commits (to avoid having a broken state in the history after merge). Also the quick lint test is currently failing because the commit messages aren't quite the required format. Would you mind rejigging them?

dpetrisko:
Hi @rswarbrick , I think everything is good to go now with the CI. Let me know if you need anything else. Thanks!

rswarbrick (approved):
This looks sensible to me (and sorry for the slow responses). Thanks!

rswarbrick:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic.sv

hcallahan-lowrisc:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic.sv

PR #24770, opened 2024-10-11
[uart,doc] Redraw the block diagram

rswarbrick:
This SVG now contains the text (rather than paths in the shape of the text), but should otherwise look the same as before.


Razer6 (approved):

jwnrt (approved):
Great, thanks

Issue #24767, opened 2024-10-10
[hmac] `HMAC_CMD_HASH_STOP_BIT` and `HMAC_CMD_HASH_PROCESS` commands causing hang conditions


moidx:
### Description

Certain delays between last message FIFO write and  `HMAC_CMD_HASH_STOP_BIT` and `HMAC_CMD_HASH_PROCESS` commands cause the `HMAC.STATUS` register to get stuck with `hmac_idle` bit cleared. 

Reproduction steps (pseudo-code): 

```c
 status = msg_fifo_write(data, len - leftover_len);
 
 // delay here, even  this small is enough
 for (size_t i = 0; i < 10; i = launder32(i + 1))
    ;

 // Time to tell HMAC HWIP to stop, because we do not have enough message
 // bytes for another round.
 uint32_t cmd_reg =
     bitfield_bit32_write(HMAC_CMD_REG_RESVAL, HMAC_CMD_HASH_STOP_BIT, 1);
 abs_mmio_write32(kHmacBaseAddr + HMAC_CMD_REG_OFFSET, cmd_reg);

 // Wait for HMAC HWIP operation to be completed.
 // In the error condition, this function will block forever.
 status = hmac_idle_wait();
```

During regular operating conditions, interrupts can fire at any time, causing delays similar to what is captured in the reproduction steps. This may cause hangs in the field that will be very difficult to debug / root cause.

Issue reported by @vsukhoml. 
CC: @vogelpi, @martin-velay, @johannheyszl, @gdessouky

vogelpi:
Thanks for reporting this @vsukhoml , and thanks for filing the issue @moidx .

Is it possible to workaround the issue by checking the FIFO empty status bit before triggering the stop? Alternatively, can we check the idle status bit before triggering the stop?

It would be good to understand what the status of the FIFO and the overall IP is when the issue occurs and when not. Emptying the FIFO takes just 16 clock cycles, and computing a block in order of 48 to 80 clock cycles. With 30 - 40 loop iterations I would expect that we signal the stop when the hardware is already idle.

vogelpi:
Update: I could reliably reproduce this on the FPGA using this command:
```
bazel test --test_output=streamed --cache_test_results=no //sw/device/tests/crypto:hmac_multistream_functest_fpga_cw310_sival_rom_ext
```
[...]

PR #24771, opened 2024-10-11
[hmac,rtl] Fix hang after stop #24767

martin-velay:
Related to issue #24767
  - Listen to msg_feed_complete when in FifoLoadFromFifo state to move into FifoIdle state.
  - And latch digest_on_blk when it happens in FifoLoadFromFifo state.

### DIFF
--- a/hw/ip/prim/rtl/prim_sha2.sv
+++ b/hw/ip/prim/rtl/prim_sha2.sv
@@ -332,6 +332,10 @@ module prim_sha2 import prim_sha2_pkg::*;
           // Wait until it is filled
           fifo_st_d          = FifoLoadFromFifo;
           update_w_from_fifo = 1'b0;
+          if (msg_feed_complete) begin
+            fifo_st_d       = FifoIdle;
+            hash_done_next  = 1'b1;
+          end
         end else if (w_index_q == 4'd 15) begin
           fifo_st_d = FifoWait;
           // To increment w_index and it rolls over to 0
@@ -443,12 +447,25 @@ module prim_sha2 import prim_sha2_pkg::*;
     if (!sha_en_i || hash_go) sha_st_d  = ShaIdle;
   end
 
+  logic update_digest_q, update_digest_d;
+
   // Determine whether a digest is being computed for a complete block: when `update_digest` is set,
   // this module is not waiting for more data from the FIFO, and `message_length_i` is zero modulo a
   // complete block (512 bit for SHA2_256 and 1024 bit for SHA2_384 and SHA2_512).
-  assign digest_on_blk_o = update_digest && (fifo_st_q == FifoIdle) && (
+  assign digest_on_blk_o = (update_digest || update_digest_q) && (fifo_st_q == FifoIdle) && (
       (digest_mode_flag_q == SHA2_256                 && message_length_i[8:0] == '0) ||
       (digest_mode_flag_q inside {SHA2_384, SHA2_512} && message_length_i[9:0] == '0));
+
+  assign update_digest_d = digest_on_blk_o ? 1'b0 :
+                           update_digest   ? 1'b1 : update_digest_q;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      update_digest_q <= 1'b0;
+    end else begin
+      update_digest_q <= update_digest_d;
+    end
+  end
 
   assign one_chunk_done = ((digest_mode_flag_q == SHA2_256 || ~MultimodeEn)
                           && (round_q == 7'd63)) ? 1'b1 :


Razer6:
Is there also an impact on Darjeeling's DMA that uses the prim_sha?

gdessouky:
> Is there also an impact on Darjeeling's DMA that uses the prim_sha?

@Razer6 it would be triggered (i.e., signalling the end of the hash operation would hang) but if there is a delay between issuing the last word to the SHA-2 engine and the process command that exceeds 64 cycles for SHA-2 256 and 80 for SHA-2 384/512, which as far as I recall was never an expected delay for DMA's inline hashing.

gdessouky:
> > Is there also an impact on Darjeeling's DMA that uses the prim_sha?
> 
> @Razer6 it would be triggered (i.e., signalling the end of the hash operation would hang) but only if there is a delay of more than 64 or 80 cycles (for SHA-2 256 or SHA-2 384/512 respectively) between issuing the last word to the SHA-2 engine and the process command. As far as I recall and see in the RTL, this was never an expected delay for DMA's inline hashing.

In fact, this specific issue is triggered only when issuing `hash_stop` (used to support context switching in HMAC). Issuing `hash_process` irrespective of the delay between it and last word works fine.

Razer6:
Thanks @gdessouky for the clarification.

hw/ip/prim/rtl/prim_sha2.sv
```diff
@@ -443,12 +447,27 @@ module prim_sha2 import prim_sha2_pkg::*;
...
+  assign pre_digest_on_blk = (update_digest && on_blk) ? 1'b1 :
+                             (fifo_st_q == FifoIdle)   ? 1'b0 :
+                             pre_digest_on_blk_q;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) pre_digest_on_blk_q <= 1'b0;
+    else         pre_digest_on_blk_q <= pre_digest_on_blk;
+  end
+
+  assign digest_on_blk_o = ~pre_digest_on_blk & pre_digest_on_blk_q;
```
vogelpi, line 470: @martin-velay and I discussed offline, this can maybe be reduced to something like:
```sv
  // Determine whether a digest is being computed for a complete block: when `update_digest` is set,
  // this module is not waiting for more data from the FIFO, and `message_length_i` is zero modulo a
  // complete block (512 bit for SHA2_256 and 1024 bit for SHA2_384 and SHA2_512).
  assign digest_on_blk_o = (update_digest_d || update_digest_q)  && (fifo_st_q == FifoIdle) && (
      (digest_mode_flag_q == SHA2_256                 && message_length_i[8:0] == '0) ||
      (digest_mode_flag_q inside {SHA2_384, SHA2_512} && message_length_i[9:0] == '0));

  assign update_digest_d = update_digest   ? 1'b1 :
                           digest_on_blk_o ? 1'b0 : update_digest_q;

    if (!rst_ni) begin
      update_digest_q <= 1'b0;
    end else begin
      update_digest_q <= update_digest_d;
    end
  end
```
which would mean something like 3 to 4 2-input logic gates and 1 FF.

martin-velay: @vogelpi, thanks for this alternative solution. It seems to work well in simulation! 👏
I've just pushed the changes on this PR.

vogelpi (approved):
The RTL change looks good to me, thanks @martin-velay !

But I agree that we should remove the DV change again.

We can merge this PR after the patch version increment went in (#25530).

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -558,6 +558,11 @@ task hmac_base_vseq::save_and_restore();
   // Wait until message transmission is on a block boundary (multiple of 512 bits in SHA-2 256
   // or 1024 bits SHA-2 384/512)
   sar_window.wait_trigger();
+
+  // Insert random delay to mimic the SW accesses which can take some time because of potential
+  // incoming interrupts.
+  cfg.clk_rst_vif.wait_clks($urandom_range(0, 100));
```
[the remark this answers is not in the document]
martin-velay: Yes you're right, thanks!

hw/ip/prim/rtl/prim_sha2.sv
```diff
@@ -443,13 +447,26 @@ module prim_sha2 import prim_sha2_pkg::*;
     if (!sha_en_i || hash_go) sha_st_d  = ShaIdle;
   end
 
+  logic update_digest_q, update_digest_d;
+
   // Determine whether a digest is being computed for a complete block: when `update_digest` is set,
   // this module is not waiting for more data from the FIFO, and `message_length_i` is zero modulo a
   // complete block (512 bit for SHA2_256 and 1024 bit for SHA2_384 and SHA2_512).
-  assign digest_on_blk_o = update_digest && (fifo_st_q == FifoIdle) && (
+  assign digest_on_blk_o = (update_digest_d || update_digest_q)  && (fifo_st_q == FifoIdle) && (
```
vogelpi, line 455: This should be
```sv
digest_on_blk_o = (update_digest || update_digest_q)  && (fifo_st_q == FifoIdle) && (
```
to avoid the combo loop flagged by AscentLint.

hw/ip/prim/rtl/prim_sha2.sv
```diff
@@ -443,13 +447,26 @@ module prim_sha2 import prim_sha2_pkg::*;
...
       (digest_mode_flag_q == SHA2_256                 && message_length_i[8:0] == '0) ||
       (digest_mode_flag_q inside {SHA2_384, SHA2_512} && message_length_i[9:0] == '0));
 
+  assign update_digest_d = update_digest   ? 1'b1 :
+                           digest_on_blk_o ? 1'b0 : update_digest_q;
```
vogelpi, line 460: And here we need:
```sv
  assign update_digest_d = update_digest && ~digest_on_blk_o ? 1'b1 :
                           digest_on_blk_o                   ? 1'b0 : update_digest_q;
```
or maybe even simpler:
```sv
  assign update_digest_d = digest_on_blk_o ? 1'b0 :
                           update_digest   ? 1'b1 : update_digest_q;
```
But this would need to be checked using DV.

martin-velay: Thanks Pirmin. Indeed, I'll check it !

andreaskurth:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_sha2.sv

This PR fixes an RTL bug.

PR #24775, opened 2024-10-12
[dvsim] Additional simulation env Altair - Changes to nc to improve stability improvements and cleanup

venkatk-ot:
multiple updates to improve stabilty and cleanup

remove using SNAPSHOTS to improve stability and performance
remove  post_cmd feature of nc_run which is not being used for any real purpose
move to using shared file descriptors across launcher and subprocesses that spawn off nc calls to avoid log file clobbering / collision and improve stability


johngt:
Hi @venkatk-ot . I've assigned this to @rswarbrick but to assist with reviews - could you please add the overall context in the title - e.g. I think this is a continuation of adding addition simulation environments, so would suggest adding to the title - 'Additional simulation env Altair - Changes to nc to improve stability improvements and cleanup'.

util/dvsim/NcLauncher.py
```diff
@@ -86,49 +85,49 @@ def _do_launch(self):
 
         self._dump_env_vars(exports)
 
+        # using os.open instead of fopen as this allows
+        # sharing of file descriptors across processes
+        fd = os.open(self.deploy.get_log_path(), os.O_WRONLY | os.O_CREAT)
```
rswarbrick, line 90: I'm not sure it really matters, but would it be easier to use `open` directly (with `closefd=False`, I think) and then use the file object's `fileno()` method to get the fd?

util/dvsim/NcLauncher.py
```diff
@@ -86,49 +85,49 @@ def _do_launch(self):
...
+        os.write(fd, byte_message)
+
         if self.deploy.sim_cfg.interactive:
             # Interactive: Set RUN_INTERACTIVE to 1
             exports['RUN_INTERACTIVE'] = '1'
-            interactive_flags = ['-I']
-            cmd_arr = self.get_submit_cmd(interactive_flags)
-            log.log(VERBOSE, '[Executing]:\n{}\n\n'.format(self.deploy.cmd))
+            cmd_arr = self.get_submit_cmd(['-I'])
+            buf_size = 1
```
rswarbrick, line 100: I think this is a change and we're now doing line buffering? That might be reasonable, but maybe it makes sense to leave a note explaining why?

util/dvsim/NcLauncher.py
```diff
@@ -86,49 +85,49 @@ def _do_launch(self):
...
+            std_out = subprocess.PIPE
+            std_err = subprocess.STDOUT
+        else:
+            cmd_arr = self.get_submit_cmd(['-wl'])
+            buf_size = 4096
```
rswarbrick, line 105: If 4k buffers aren't particularly important, is there a reason not to use `-1` (letting the system pick whatever it fancies)?

(I know this was in the previous code, but I'm wondering about it because the line is being touched :-D)

util/dvsim/NcLauncher.py
```diff
@@ -86,49 +85,49 @@ def _do_launch(self):
...
-                                                stderr=f,
-                                                env=exports,
-                                                cwd=self.deploy.odir)
-            except subprocess.SubprocessError as e:
-                raise LauncherError(f'IO Error: {e}\n'
-                                    f'See {self.deploy.get_log_path()}')
-            finally:
-                self._close_process()
+            if self.deploy.sim_cfg.interactive:
+                for line in self.process.stdout:
```
rswarbrick, line 121: If I understand correctly:
- In non-interactive mode, the subprocess just writes to `fd`.
- In interactive mode, the subprocess writes to some pipe that comes back to the Python process, which then does a "tee operation" and writes the data to the `fd` and also `stdout`.

I guess this makes sense. If you do the rewording I'm suggesting above, we'll have a file object floating around, so won't need the explicit call to `encode`. Maybe that would be marginally simpler?

util/dvsim/NcLauncher.py
```diff
@@ -86,49 +85,49 @@ def _do_launch(self):
...
+                    os.write(fd, line.encode())
+                    sys.stdout.write(line)
+                # Wait until the process exits
+                self.process.wait()
```
rswarbrick, line 125: Does this `wait` call ever do anything? I guess it only matters if the subprocess happens to close its `stdout` (ending the loop above)? If I've understood properly, maybe it makes sense to leave a note in the comment to stop the next person wondering :-)

rswarbrick (commented):
This looks sensible to me, and thank you for teaching me something about linux syscalls :-) A couple of nitty questions/comments.

venkatk-ot:
> Hi @venkatk-ot . I've assigned this to @rswarbrick but to assist with reviews - could you please add the overall context in the title - e.g. I think this is a continuation of adding addition simulation environments, so would suggest adding to the title - 'Additional simulation env Altair - Changes to nc to improve stability improvements and cleanup'.

Done

util/dvsim/NcLauncher.py
```diff
@@ -59,7 +58,6 @@ def get_submit_cmd(self, interactive_flags):
             license_args = ['-r', 'License:VCSRuntime_Net/1']
         else:
             license_args = []
-        license_args.extend(['-r', 'RAM/8192', '-r', 'CORES/2'])
```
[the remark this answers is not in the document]
venkatk-ot: Got it.  I have removed the commit that deletes the CPU and RAM requirements from the NcLauncher file.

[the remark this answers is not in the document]
venkatk-ot: I also removed the commit that disables the snapshot as it turns out it is needed to pass on the environment variables reliably to the processes forked off using the nc run command

util/dvsim/NcLauncher.py
venkatk-ot: done. added a comment

util/dvsim/NcLauncher.py
venkatk-ot: done. added a comment

util/dvsim/NcLauncher.py
venkatk-ot: done

util/dvsim/NcLauncher.py
venkatk-ot: I tried this, but the open command failed with this error message
Cannot use closefd=False with file name
Hence leaving the code as it is

util/dvsim/NcLauncher.py
venkatk-ot: done. added comment

util/dvsim/NcLauncher.py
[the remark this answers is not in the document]
venkatk-ot: I have added a new commit that overrides the TMPDIR variable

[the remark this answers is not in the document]
venkatk-ot: I have added a new commit that updates the nc run command's flag for interactive jobs

venkatk-ot:
@rswarbrick , sorry for the delay in my response.
I wanted to do a little more testing before I uploaded the latest patch
As mentioned in one of the reply above, I have deleted 2 commits and added 2 new commits to this PR. 
I think I have addressed all your comments above
Please take a look at the new changes

rswarbrick (commented):
One nitty question about environment variables (because I think we're not quite setting them?) but otherwise this looks sensible to me.

util/dvsim/NcLauncher.py
```diff
@@ -38,6 +38,8 @@ def create_run_sh(self, full_path, cmd):
                  'MY_FILEPATH=$(realpath "${BASH_SOURCE[0]}")',
                  'MY_DIR=$( dirname "${MY_FILEPATH}" )',
                  'cd $MY_DIR',
+                 'TMPDIR=$PWD/tmp',
+                 'mkdir -p $TMPDIR',
```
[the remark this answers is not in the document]
venkatk-ot: Thanks. I have fixed this and uploaded a patch

rswarbrick (approved):
This looks good to me, thanks!

PR #24777, opened 2024-10-14
[sival,kmac] Add kmac_endianess test

nasahlpa:
This commit adds a chip level test to test the endianess feature of KMAC. This includes the message as well as the state register.

The following configurations are covered:
- Little endian message, little endian digest.
- Little endian message, big endian digest.
- Big endian message, little endian digest.
- Big endian message, big endian digest.

Currently, only the KMAC mode is covered. However, as the big/ little endian feature should be independent of the underlying mode, this should be sufficient.

### DIFF
--- a/sw/device/tests/kmac_endianess_test.c
+++ b/sw/device/tests/kmac_endianess_test.c
@@ -1,0 +1,158 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/device/lib/arch/device.h"
+#include "sw/device/lib/base/macros.h"
+#include "sw/device/lib/base/mmio.h"
+#include "sw/device/lib/dif/dif_kmac.h"
+#include "sw/device/lib/runtime/log.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+OTTF_DEFINE_TEST_CONFIG();
+
+/**
+ * KMAC test description.
+ */
+typedef struct kmac_test {
+  dif_kmac_mode_kmac_t mode;
+  dif_kmac_key_t key;
+
+  const char *message_big_endian;
+  const char *message_little_endian;
+  size_t message_len;
+
+  size_t digest_len;
+} kmac_test_t;
+
+/**
+ * KMAC test.
+ *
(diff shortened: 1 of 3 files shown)


sw/device/tests/kmac_endianess_test.c
```diff
@@ -0,0 +1,207 @@
...
+  CHECK_DIF_OK(
+      dif_kmac_init(mmio_region_from_addr(TOP_EARLGREY_KMAC_BASE_ADDR), &kmac));
+
+  LOG_INFO("Testing big endian state.");
+  dif_kmac_config_t config = (dif_kmac_config_t){
+      .entropy_mode = kDifKmacEntropyModeSoftware,
+      .entropy_seed = {0xb153e3fe, 0x09596819, 0x3e85a6e8, 0xb6dcdaba,
+                       0x50dc409c, 0x11e1ebd1},
+      .entropy_fast_process = kDifToggleEnabled,
+      .output_big_endian = true,
```
vogelpi, line 97: Should you also set the `message_big_endian` to true? You're using the big endian messages for this test but I think if you don't configure the corresponding struct member to true, it might actually default to false.

sw/device/tests/kmac_endianess_test.c
```diff
@@ -0,0 +1,207 @@
...
+  return OK_STATUS();
+}
+
+/**
+ * Message endianess test.
+ *
+ * This test runs a KMAC operation twice, in the first run with
+ * message_big_endian true and the input message is in big ending. In the second
+ * run, the message_big_endian config is set to false and the input message is
+ * in little endian. In the end, the output digest should match.
```
vogelpi, line 144: It probably makes sense to also document which digest endianess you're using here.

vogelpi (commented):
Thanks @nasahlpa for working on this. The test implementation looks good to me. I have just some minor comments.

One more question: The test plan says that we should test all combinations of message and digest endianess. Assuming that you explicitly set the message endianess for the digest test to big endian, we test the following cases:
- digest test:
  - msg big, digest little
  - msg big, digest big
- msg test:
  - msg big, digest big
  - msg little, digest big

So we don't test  the combination msg little and digest little. Can you change the test to cover all combinations?

nasahlpa:
Thanks for the review, @vogelpi!

I tested this endianess configuration as internally the core anyway always operates on little endian and produces little endian output. In the big endian message case, the input is converted. In the little endian message case, the output is converted. So my reasoning was that testing all combinations shouldn't be necessary.

However, I see your point that we should test whether this behavior really is given, so I changed the test to cover all possible combinations. This also simplified the test :-)

vogelpi (approved):
This looks good, thanks @nasahlpa !

Issue #24610, opened 2024-09-20
[manuf] Preventing Re-execution of Perso Firmware


anthonychen1251:
### Description
Currently, the perso firmware remains executable even after personalization is complete. Moreover, the Perso firmware lacks an embedded immutable ROM extension, despite having a valid signature. If the immutable ROM extension feature is enabled and the corresponding hash is provisioned, running perso again may result in ROM self-shutdown due to an invalid immutable ROM extension hash. Consequently, ROM will not attempt to boot from an alternative slot, eliminating the possibility of booting into a valid ROM extension under these circumstances. This situation renders the device bricked until another image is bootstrapped. To mitigate this, we propose binding the creator manufacturing state within usage constraints, ensuring the perso firmware's signature is invalidated once the device is fully personalized, and thus preventing the perso firmware from being executed again.
 More details described [here](https://docs.google.com/document/d/174zeglo4yHyC9utqDss7Pod8_uv6DSTFWtyyLcxm5kE/edit?tab=t.0)

The goal is to implement the new FT flow below

1. Individualization:
 a. Skip the provisioning of `CREATOR_SW_CFG_IMMUTABLE_ROM_EXT_EN`, `CREATOR_SW_CFG_MANUF_STATE`, `CREATOR_SW_CFG_FLASH_DATA_DEFAULT_CFG`, `OWNER_SW_CFG_ROM_BOOTSTRAP_DIS`, and ~~`OWNER_SW_CFG_ROM_KEYMGR_OTP_MEAS_EN`~~.
2. Bootstrap Perso FW into flash slot A.
3. Personalization:
a. Configure the secret1 (flash and SRAM scrambling seeds) and provision `CREATOR_SW_CFG_FLASH_DATA_DEFAULT_CFG`.
b. Reboot and re-bootstrap (required when flash scrambling is enabled); this time loading Perso FW in flash slot A and ROM_EXT + Owner FW in flash slot B.
c. Personalize the device.
d. Fully provision the SW_CFG partitions, including `CREATOR_SW_CFG_IMMUTABLE_ROM_EXT_EN`, `OWNER_SW_CFG_ROM_BOOTSTRAP_DIS`, and ~~`OWNER_SW_CFG_ROM_KEYMGR_OTP_MEAS_EN`~~.
e. Transitions the `CREATOR_SW_CFG_MANUF_STATE` to `Personalized`.
f. Write-lock both OTP *_SW_CFG partitions.
[...]

PR #24783, opened 2024-10-14
[manuf] Finalize {Creator,Owner}SwCfg at end of perso flow 

anthonychen1251:
This PR has two commits that:

1. Move the calculation of OTP measurement from `dice_uds_tbs_cert_build` to `ft_personalize`
2. Modify the perso flow to complete the provisioning of the `CreatorSwCfg` and `OwnerSwCfg` OTP partitions at the end of the process.

   - Sets the expected values for fields in the OTP that are not provisioned until the final stages of personalization.
   - Compare the OTP measurement used during certificate generation with the value stored in the OTP.

This PR addresses #24610 partially.

### DIFF
--- a/sw/device/silicon_creator/manuf/base/ft_personalize.c
+++ b/sw/device/silicon_creator/manuf/base/ft_personalize.c
@@ -199,19 +199,18 @@ static status_t measure_otp_partition(otp_partition_t partition,
   otp_dai_read(partition, /*address=*/0, otp_state,
                kOtpPartitions[partition].size / sizeof(uint32_t));
 
+  // Sets the expected values for fields in the OTP that are not provisioned
+  // until the final stages of personalization.
+  if (partition == kOtpPartitionOwnerSwCfg) {
+    manuf_individualize_device_partition_expected_read(
+        kDifOtpCtrlPartitionOwnerSwCfg, (uint8_t *)otp_state);
+  } else if (partition == kOtpPartitionCreatorSwCfg) {
+    manuf_individualize_device_partition_expected_read(
+        kDifOtpCtrlPartitionCreatorSwCfg, (uint8_t *)otp_state);
+  }
+
   hmac_sha256(otp_state, kOtpPartitions[partition].size, measurement);
 
-  // Check the digest matches what is stored in OTP.
-  // TODO(#21554): remove this conditional once the root keys and key policies
-  // have been provisioned. Until then, these partitions have not been locked.
-  if (partition == kOtpPartitionCreatorSwCfg ||
-      partition == kOtpPartitionOwnerSwCfg) {
-    uint64_t expected_digest = otp_partition_digest_read(partition);
-    uint32_t digest_hi = expected_digest >> 32;
-    uint32_t digest_lo = expected_digest & UINT32_MAX;
-    HARDENED_CHECK_EQ(digest_hi, measurement->digest[1]);
-    HARDENED_CHECK_EQ(digest_lo, measurement->digest[0]);
-  }
   return OK_STATUS();
 }
 
@@ -225,18 +224,11 @@ static status_t personalize_otp_and_flash_secrets(ujson_t *uj) {
   if (!status_ok(manuf_personalize_device_secret1_check(&otp_ctrl))) {
     TRY(manuf_personalize_device_secret1(&lc_ctrl, &otp_ctrl));
   }
-  if (!status_ok(manuf_individualize_device_creator_sw_cfg_check(&otp_ctrl))) {
+  if (!status_ok(
+          manuf_individualize_device_flash_data_default_cfg_check(&otp_ctrl))) {
     TRY(manuf_individualize_device_flash_data_default_cfg(&otp_ctrl));
-    TRY(manuf_individualize_device_creator_sw_cfg_lock(&otp_ctrl));
     LOG_INFO("Bootstrap requested.");
     wait_for_interrupt();
-  }
-
-  // The last bootstrap process in the perso flow is done.
-  // Complete the provisioning of OTP OwnerSwCfg partition.
-  if (!status_ok(manuf_individualize_device_owner_sw_cfg_check(&otp_ctrl))) {
-    TRY(manuf_individualize_device_rom_bootstrap_dis_cfg(&otp_ctrl));
-    TRY(manuf_individualize_device_owner_sw_cfg_lock(&otp_ctrl));
   }
 
   // Provision OTP Secret2 partition and flash info pages 1, 2, and 4 (keymgr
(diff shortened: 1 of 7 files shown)


sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -162,6 +189,31 @@ static status_t config_and_erase_certificate_flash_pages(void) {
...
+  // Compute the digest.
+  otp_dai_read(partition, /*address=*/0, otp_state,
+               kOtpPartitions[partition].size / sizeof(uint32_t));
+
+  // Overwrites the CreatorSwCfg and OwnerSwCfg values read from OTP
+  // memory with the values from the OTP image. This is necessary because
+  // the CreatorSwCfg and OwnerSwCfg partitions are not fully provisioned
+  // in the OTP during the UDS certificate generation process.
+  if (partition == kOtpPartitionOwnerSwCfg) {
+    manuf_individualize_get_device_owner_sw_cfg((uint8_t *)otp_state);
```
timothytrippel, line 207: don't we only want to overwrite the AST values with something that is constant? like all ones?

additionally, the manuf_state field with its expected final value?

Whey do we read out the partition from OTP, then overwrite the whole thing instead of just the desired fields?

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c
```diff
@@ -91,6 +91,32 @@ static status_t otp_img_write(const dif_otp_ctrl_t *otp,
   return OK_STATUS();
 }
 
+static status_t otp_img_read(dif_otp_ctrl_partition_t partition,
```
timothytrippel, line 94: shouldn't this function just read out a specific field? why do we need to read out all fields. Don't we just want to read out the fields that are not provisioned until the end so we know what their expected values are?

Also, could we rename this `otp_expected_value_read`? or something more descript? currently, the name seems to indicate it is reading from OTP directly, not from what is expected to be written to OTP, i.e, what the values baked into the SW image itself.

sw/device/silicon_creator/manuf/base/ft_personalize.c
anthonychen1251: > don't we only want to overwrite the AST values with something that is constant? like all ones?

If I understand correctly, the AST values should be already provisioned at the individualization stage (pulled from a flash info page). https://github.com/lowRISC/opentitan/blob/202127f06e209d0366eede40b39e59f40b54e559/sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c#L197

If we overwrite the AST values with a constant here, then the otp measurement calculated at this stage won't be matched with the final otp measurement written into the flash.

> additionally, the manuf_state field with its expected final value?

Right, that will be covered in #24789.

> Whey do we read out the partition from OTP, then overwrite the whole thing instead of just the desired fields?
I've updated the commit to explicitly update the desired fields, let me know if it looks good to you. My original thought was that explicitly overwriting the desired fields would require extra efforts to update a desired list when a new field needs to be provisioned at the end of perso process, while overwriting the whole thing wouldn't. And it should be safe to overwrite the whole partitions with the expected values for calculating the expected measurement because we will compare the OTP measurement used during certificate generation with the actual measurement stored in the OTP.

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c
anthonychen1251: I've updated the commit to explicitly update the desired fields, let me know if it looks good to you. 

Renamed the function to `otp_img_expected_value_read`

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.h
```diff
@@ -68,6 +68,27 @@ OT_WARN_UNUSED_RESULT
...
+ *
+ * @param otp_ctrl OTP controller instance.
+ * @return OK_STATUS if the FLASH_DATA_DEFAULT_CFG field is provisioned.
+ */
+OT_WARN_UNUSED_RESULT
+status_t manuf_individualize_device_flash_data_default_cfg_check(
+    const dif_otp_ctrl_t *otp_ctrl);
+
+/**
+ * Reads the expected CREATOR_SW_CFG partition values from the OTP image.
```
timothytrippel, line 83:
```suggestion
 * Overwrites unprovisioned fields with their expected final values in a buffer representing the provided partition.
```

Can we create one function that can be used for both Creator and Owner SwCfg partitions?

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c
```diff
@@ -91,6 +91,44 @@ static status_t otp_img_write(const dif_otp_ctrl_t *otp,
...
+ * @param len Length of the `kv` array.
+ * @param buffer A buffer containing the entire target OTP partition.
+ * @return OK_STATUS if the expected OTP values are successfully written to the
+ * `buffer`.
+ */
+static status_t otp_img_expected_value_read(dif_otp_ctrl_partition_t partition,
+                                            const otp_kv_t *kv, size_t len,
+                                            uint8_t *buffer) {
+  for (size_t i = 0; i < len; ++i) {
+    if (kv[i].offset != OTP_CTRL_PARAM_OWNER_SW_CFG_ROM_BOOTSTRAP_DIS_OFFSET) {
```
timothytrippel, line 110: Maybe this offset could be a passed parameter? so this would more easily generalize in the future?

timothytrippel: Also, I feel like this function is necessarily complicated. Cant the function just take the following inputs:
- the buffer
- the kv item that should be overwritten
and just overwrite that field? why do you need to loop over the whole buffer? Am I missing something?

sw/device/silicon_creator/manuf/base/ft_personalize.c
timothytrippel: Ah right r.e. AST values, they will already be programmed. Sorry I that is for another purpose I am thinking of.

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c
anthonychen1251: > why do you need to loop over the whole buffer? Am I missing something?

We don't know where a given field is located in the kv array, so we need to loop over the kv array to find the corresponding kv item by checking the offset.
> Maybe this offset could be a passed parameter? so this would more easily generalize in the future?

I did consider passing the offset as a parameter, but since we have to loop over the kv array to find a target kv item, sending the offsets under the same partition separately will require looping over the kv array multiple times to update the buffer, while maintaining the offsets in this function only requires looping the kv array at most once  to update the buffer of a given partition.

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.h
anthonychen1251: Yes, we can create a single function that can be used for both Creator and Owner SwCfg partitions. I considered passing the partition as a parameter but finding that the implementation in `ft_personalize`  uses `enum otp_partition` to index a partition while the implementation in `individualize_sw_cfg` uses `enum dif_otp_ctrl_partition` to index a partition. This means that if a mapping between `enum otp_partition`  and `individualize_sw_cfg`  would be required if we pass the partition as a parameter. As a result, I decided to define two separate functions. I will check if it is possible to only use either  `enum otp_partition` or  `enum dif_otp_ctrl_partition`  on both sides.

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.c
anthonychen1251: Ah, I just realized that since we've already exposed the expected value of the fields that are not provisioned until the end of perso, it's actually unnecessary to get the expected value by accessing the kv item.

anthonychen1251: Done.

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.h
anthonychen1251: Done.

anthonychen1251:
The failed test runs `usbdev_iso_test_fpga_cw310_sival_rom_ext` and `usbdev_stream_test_fpga_cw310_sival_rom_ext` in the most recent CI pipeline should be unrelated to this PR.

timothytrippel (approved):
Thanks @anthonychen1251 , LGTM!

PR #24784, opened 2024-10-14
[opentitantool] Add vcd outfile option to gpio monitor start and read

jettr:
Augment the `gpio monitoring` `start` and `read` commands with the ability to also write a vcd file. The `start` and `read` commands do not block or hold the opentitantool interface, so they can be used in conjunction with other opentitantool commands. For example, the `gpio monitoring` `start` and `read` commands can be used with the `tpm i2c` commands to record all of the i2c traffic from hyperdebug and the DUT. The existing `gpio monitoring vcd` command holds the hyperdebug resource until it is finished, so it is currently not possible to sent tpm i2c commands to the DUT using opentitan tool while the `gpio monitoring vcd` is running.

### DIFF
--- a/sw/host/opentitantool/src/command/gpio.rs
+++ b/sw/host/opentitantool/src/command/gpio.rs
@@ -16,7 +16,9 @@ use std::time::Duration;
 
 use opentitanlib::app::command::CommandDispatch;
 use opentitanlib::app::TransportWrapper;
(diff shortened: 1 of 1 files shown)


jwnrt (approved):
This looks good to me, @pamaury could you take a look?

pamaury (approved):
Looks good to me

PR #24785, opened 2024-10-14
[rom_ext] Initialize boot_log after boot_svc

cfrantz:
A number of `boot_log` fields are copies of fields in `boot_data`. Since boot services can update `boot_data`, we need to initialize those `boot_log` entries after boot services completes.

Fixes: 24768


moidx (approved):

cfrantz:
The CI test failures are not related to this PR.

PR #24790, opened 2024-10-15
[hw,rv_dm,rtl] Unified RV_DM implementation for JTAG TAP and TLUL-based DMI

Razer6:


hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -409,7 +394,7 @@ module rv_dm
     //    errors during test / debug life cycle states so that the system can be debugged
     //    without triggering alerts.
     // 3) the error condition is hooked up to an error CSR that can be read out by the debugger
-    //    via JTAG so that bus integrity errors can be told appart from regular bus errors.
+    //    via JTAG or DMI so that bus integrity errors can be told appart from regular bus errors.
```
rswarbrick, line 400: Nit (since you're touching this line anyway): Maybe fix the spelling of "apart"?

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -440,65 +425,233 @@ module rv_dm
...
+      .lc_en_o(pinmux_hw_debug_en)
+    );
+
+    mubi4_t dmi_tlul_adapter_en;
+    // Convert LC signal from pinmux to MuBi4 value: As the multi-bit LC signal may cross clock
+    // domains, it may not have a valid LC value in every clock cycle.  This is not to be
+    // interpreted as an injected fault, so the conversion to a boolean tests for a strictly true
+    // LC value and interprets all other values as false.
+    assign dmi_tlul_adapter_en = prim_mubi_pkg::mubi4_bool_to_mubi(
+                                  lc_tx_test_true_strict(pinmux_hw_debug_en[PmEnDmiTlulAdapter]));
```
rswarbrick: I guess the downside of this is that we end up with a 1-bit signal that can be forced to give a blip with the wrong value. I don't think a cycle with a *bogus* value matters from the CDC matters: we're only using `dmi_tlul_adapter_en` as the `en_ifetch_i` signal in `tlul_adapter_reg` and that module just does a strict check for true (with no fault if the signal is nonsense).

Maybe a straight lc_tx -> mubi conversion without the 1-bit signal in the middle is the right path?

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -440,65 +425,233 @@ module rv_dm
...
+      .we_o        (dmi_we),
+      .addr_o      (dmi_addr),
+      .wdata_o     (dmi_req.data),
+      .be_o        (dmi_be),
+      .busy_i      (dmi_busy),
+      .rdata_i     (dmi_rsp.data),
+      .error_i     (dmi_error)
+    );
+    assign dmi_rsp_ready = 1'b1; // `tlul_adapter_reg` is always ready to receive responses.
+    assign dmi_req.addr = dmi_addr[DmiByteAddrWidth-1:2]; // Convert from byte to word addressing.
```
rswarbrick: Do we want to do anything with the bottom two bits? I think that we're just flooring the address at the moment. Another option might be to squash `dmi_req.valid` ("ignore attempts to access an address that doesn't exist") or even to fold this into `dmi_inject_error_d`?

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -440,65 +425,233 @@ module rv_dm
...
+          dmi_req.op = dm::DTM_READ;
+          dmi_req_valid = 1'b1;
+        end
+        default /* write & read */: begin
+          // Inject error after write & read.
+          dmi_inject_error_d = 1'b1;
+        end
+      endcase
+    end
+    assign dmi_busy = dmi_inject_error_d ? 1'b0           // when intercepting requets: never
```
rswarbrick: Nit: Spelling of "requests"

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -440,65 +425,233 @@ module rv_dm
...
+                                        : ~dmi_req_ready; // when forwarding requests: iff not ready
```
rswarbrick: Nit: Indentation of this line

hw/ip/rv_dm/rtl/rv_dm.sv
rswarbrick: Hmm. This code gets removed with the next commit too. Never mind :-)

hw/ip/rv_dm/dv/rv_dm_use_dmi_interface_sim_cfg.hjson
```diff
@@ -0,0 +1,15 @@
...
+
+// sim cfg file for the variant of RV_DM using the DMI interface
+{
+  // Name of the sim cfg variant
+  variant: use_dmi_interface
+
+  // Import additional common sim cfg files.
+  import_cfgs: ["{proj_root}/hw/ip/rv_dm/dv/rv_dm_base_sim_cfg.hjson"]
+
+  // Use the JTAG interface and not the DMI interace
```
rswarbrick, line 13: I think this comment needs reversing :-)

hw/ip/rv_dm/dv/tb.sv
```diff
@@ -25,11 +25,39 @@ module tb;
...
+    .rst_ni      (rst_n),
+    .jtag_i      ({jtag_if.tck, jtag_if.tms, jtag_if.trst_n, jtag_if.tdi}),
+    .jtag_o      ({jtag_if.tdo, jtag_tdo_oe}),
+    .scan_rst_ni (rv_dm_if.scan_rst_n),
+    .scanmode_i  (rv_dm_if.scanmode),
+    .tl_h2d_o    (dmi_tl_h2d),
+    .tl_d2h_i    (dmi_tl_d2h)
+  );
+`else
+  assign    = tlul_pkg::TL_H2D_DEFAULT;
```
rswarbrick, line 50: The syntax doesn't quite look right here. Should the LHS be `dmi_tl_h2d`?

hw/ip/rv_dm/dv/tb.sv
```diff
@@ -25,11 +25,39 @@ module tb;
   jtag_if jtag_if();
   rv_dm_if rv_dm_if(.clk(clk), .rst_n(rst_n));
 
+  // Used for JTAG DTM connections via TL-UL.
+  tlul_pkg::tl_h2d_t dmi_tl_h2d;
+  tlul_pkg::tl_d2h_t dmi_tl_d2h;
+
   `DV_ALERT_IF_CONNECT()
 
+`ifdef USE_DMI_INTERFACE
```
rswarbrick, line 34: I'm not sure this is right: I think you're defining it either way, but as 0 or 1?

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -452,7 +452,11 @@ module rv_dm
 `ifndef DMIDirectTAP
     tlul_pkg::tl_h2d_t dmi_tl_h2d_gated;
     tlul_pkg::tl_d2h_t dmi_tl_d2h_gated;
-    tlul_lc_gate u_tlul_lc_gate_dmi (
+
+    tlul_lc_gate #(
+      // If the DMI side is gated, the spec requires the DM to return a valid all-zero response.
+      .ReturnBlankResp(1)
+    ) u_tlul_lc_gate_dmi (
```
rswarbrick: Can't this be squashed into the commit that instantiates the gate in the first place? (This would avoid a "introduce bug; fix bug" history?)

hw/ip/rv_dm/data/rv_dm.hjson
```diff
@@ -178,6 +191,52 @@
...
+    { struct:  "lc_tx"
+      type:    "uni"
+      name:    "lc_escalate_en"
+      act:     "rcv"
+      default: "lc_ctrl_pkg::Off"
+      package: "lc_ctrl_pkg",
+      desc:    '''
+               Escalation enable signal coming from life cycle controller, used for invalidating
+               the latched lc_hw_debug_en state inside the strap sampling logic.
+               ''',}
```
rswarbrick, line 184: Nit: I realise that it has no effect, but I'm not a big fan of the ",}" syntax. Maybe drop the comma?

rswarbrick (commented):
This is a partial review, I'm afraid: I've only got half-way through "[hw,rv_dm,rtl] Enable RV_DM to check straps without the pinmux" so far. But hopefully it's a useful start.

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -445,114 +445,47 @@ module rv_dm
...
-    // Convert LC signal from pinmux to MuBi4 value: As the multi-bit LC signal may cross clock
-    // domains, it may not have a valid LC value in every clock cycle.  This is not to be
-    // interpreted as an injected fault, so the conversion to a boolean tests for a strictly true
-    // LC value and interprets all other values as false.
-    assign dmi_tlul_adapter_en = prim_mubi_pkg::mubi4_bool_to_mubi(
-                                  lc_tx_test_true_strict(pinmux_hw_debug_en[PmEnDmiTlulAdapter]));
+    // SEC_CM: DM_EN.CTRL.LC_GATED
+    assign dmi_en = lc_tx_test_true_strict(pinmux_hw_debug_en[PmEnDmiTlulAdapter]);
 
+    // Bound-in DPI module replaces the TAP
```
rswarbrick: Can you reverse this comment? It's sitting above the ifndef part, so maybe "The TAP is not being replaced by a bound-in DPI module" or similar?

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -435,59 +438,38 @@ module rv_dm
       PmEnLastPos
     } rv_dm_pm_en_e;
```
rswarbrick, line 439: I think this commit also gets rid of the only use of this enum, so we can drop that too?

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -435,59 +438,38 @@ module rv_dm
...
-      .dmi_req_o       (dmi_req),
-      .dmi_resp_valid_i(dmi_rsp_valid & dmi_en),
-      .dmi_resp_ready_o(dmi_rsp_ready),
-      .dmi_resp_i      (dmi_rsp)
+      .strap_en_override_i,
+      .strap_en_i,
+      .lc_hw_debug_en_i,
+      .lc_check_byp_en_i,
+      .lc_escalate_en_i,
+      .dbg_tl_h2d_win_i(               dmi_tl_h2d_i),
```
rswarbrick: I think we can probably trim some spaces from this aligned block of ports. The usual style is also to put the spaces before the opening paren, which I think is a little easier to read. (possibly only because I'm used to that style...)

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -517,9 +499,13 @@ module rv_dm
     assign jtag_in_int = (lc_tx_test_true_strict(pinmux_hw_debug_en[PmEnJtagIn]))  ? jtag_i : '0;
     assign jtag_o = (lc_tx_test_true_strict(pinmux_hw_debug_en[PmEnJtagOut])) ? jtag_out_int : '0;
 
+    logic dmi_en;
+
     // SEC_CM: DM_EN.CTRL.LC_GATED
     assign dmi_en = lc_tx_test_true_strict(pinmux_hw_debug_en[PmEnDmiReq]);
 
+    logic dmi_req_valid_raw, dmi_rsp_ready_raw;
+
```
rswarbrick, line 521: I'm not sure I have an opinion about the correct place for this (inside the ifndef or not), but I wouldn't expect the answer to change with this commit. Does one version (before or after!) need changing?

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -656,37 +644,37 @@ module rv_dm
...
+    .master_r_err_i        (host_r_err          ),
+    .master_r_other_err_i  (host_r_other_err    ),
+    .master_r_rdata_i      (host_r_rdata        ),
+    .dmi_rst_ni            (dmi_rst_n           ),
+    .dmi_req_valid_i       (dmi_req_valid       ),
+    .dmi_req_ready_o       (dmi_req_ready       ),
+    .dmi_req_i             (dmi_req             ),
+    .dmi_resp_valid_o      (dmi_rsp_valid       ),
+    .dmi_resp_ready_i      (dmi_rsp_ready       ),
+    .dmi_resp_o            (dmi_rsp              )
```
rswarbrick, line 691: There's a stray space before the ")" here, I think.

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -0,0 +1,280 @@
...
+        strap_en_override_q <= strap_en_override_d;
+      end
+    end
+
+    // Detect a change from 0 -> 1 on the override signal (it will stay at 1 afterwards).
+    assign strap_en = strap_en_i || (strap_en_override_d && !strap_en_override_q);
+
+    // The strap sampling override shall be set to high exactly once.
+    `ASSUME(LcCtrlStrapSampleOverrideOnce_A,
+        $rose(strap_en_override_i) |-> always strap_en_override_i)
```
rswarbrick, line 115: This took me a minute or two to understand! I think I would have got there quicker if there was some documentation above the `strap_en` declaration that explicitly says this is only supposed to be high for a pulse and the comment on line 110 was something like "Detect a posedge on the override signal and reflect it for just a cycle on strap_en (ignoring the fact that strap_en_override_i will stay high afterwards)."

Would that be sensible? (and is it correct?!)

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -0,0 +1,280 @@
...
+
+  end else begin : gen_no_strap_override
+    logic unused_strap_en_override;
+    assign unused_strap_en_override = strap_en_override_i;
+    assign strap_en = strap_en_i;
```
rswarbrick, line 120: As a debugging aid, would it make sense to also include an assertion here that says the strap override stays low? That might make it even less likely we forget to turn the override off for a production version.

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -0,0 +1,280 @@
...
+  end
+
+  /////////////////////////////
+  // LC_HW_DEBUG_EN Latching //
+  /////////////////////////////
+
+  // In order to keep a RV_DM JTAG debug session alive during NDM reset, we need to memorize the
+  // state of lc_hw_debug_en, since OTP/LC will be reset as part of NDM reset (the only parts not
+  // reset are in the pwr/rst/clkmgrs, RV_DM and this strap sampling module). We sample the life
+  // cycle signal state when the strap sampling pulse is asserted bu the PWRMGR. This pulse is
```
rswarbrick, line 130: Typo here: "bu the PWRMGR"

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -0,0 +1,280 @@
...
+    .clk_i,
+    .rst_ni,
+    .lc_en_a_i(lc_hw_debug_en_masked),
+    .lc_en_b_i(strap_hw_debug_en_q),
+    .lc_en_o  (hw_debug_en_set)
+  );
+
+  // Output ON if both lc_check_byp_en and lc_escalate_en are set to OFF.
+  lc_tx_t hw_debug_en_gating;
+  assign hw_debug_en_gating = lc_tx_inv(lc_tx_and_lo(lc_check_byp_en[0], lc_escalate_en[0]));
```
rswarbrick, line 166: I had to squint and think about deMorgan's laws, but I think this code is more like
```
hw_debug_en = !(byp_en && escalate_en) = !byp_en || !escalate_en
```
which doesn't match the comment. Do you mean "either ... or ... is"?

rswarbrick (commented):
Some initial comments about "[hw,rv_dm,rtl] Enable RV_DM to check straps without the pinmux". I haven't got to the end, but have to go and catch a train :-) Will try to write more when I'm back in the uk...

hw/ip/rv_dm/rtl/rv_dm.sv
Razer6: Sounds good.

hw/ip/rv_dm/rtl/rv_dm.sv
```diff
@@ -33,9 +34,12 @@ module rv_dm
   // SEC_CM: LC_DFT_EN.INTERSIG.MUBI
   // HW DFT lifecycle enable signal (live version from the life cycle controller)
   input  lc_ctrl_pkg::lc_tx_t lc_dft_en_i,
-  // HW Debug lifecycle enable signal (latched version from pinmux, only used for JTAG/TAP or DMI
-  // gating)
+  // HW Debug lifecycle enable signal (latched version from pinmux, only used for JTAG/TAP gating)
```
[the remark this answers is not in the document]
Razer6: Yes, for the JTAG TAP case. That one is unchanged, that's why this commit reverts it back to the original one. In the overall diff, there is no change on this line.

hw/ip/rv_dm/rtl/rv_dm.sv
Razer6: Yep.

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
Razer6: Sounds good.

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
Razer6: That's right.

hw/ip/rv_dm/rtl/rv_dm.sv
Razer6: perfect!

hw/ip/rv_dm/rtl/rv_dm.sv
Razer6: perfect

hw/ip/rv_dm/data/rv_dm.hjson
rswarbrick: Sorry to be annoying but I think the comma is still here on the first item (but thanks for dropping it on the later ones!)

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_tap_fsm_vseq.sv
```diff
@@ -17,6 +17,15 @@ class rv_dm_tap_fsm_vseq extends rv_dm_base_vseq;
...
 
+`ifdef USE_DMI_INTERFACE
+  constraint lc_hw_debug_en_c {
+    lc_hw_debug_en == lc_ctrl_pkg::On;
+  }
+  constraint scanmode_c {
+    scanmode == prim_mubi_pkg::MuBi4False;
+  }
+`endif
+
```
rswarbrick, line 28: These variables have changed types so this code isn't going to work, but the meaning that Jaedon was going for is actually now the default behaviour, so I don't think we need this commit at all!

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -183,7 +183,6 @@ module rv_dm_dmi_gate
   typedef enum logic [1:0] {
     StEnDmiTlulAdapter,
     StEnDmiTlulLcGate,
-    StEnDmiTlulZeroResponder,
```
rswarbrick: Can you squash this with "[hw,rv_dm,rtl] Enable RV_DM to check straps without the pinmux"? I don't think we need to keep the intermediate version.

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -266,12 +266,12 @@ module rv_dm_dmi_gate
...
       |->
       $past(lc_tx_test_true_strict(lc_hw_debug_en[0])))
-  // Check that latching ON can only occur if strap_en_i is set.
+  // Check that latching ON can only occur if strap_en is set.
   `ASSERT(LcHwDebugEnSetRev1_A,
       lc_tx_test_false_loose(strap_hw_debug_en_q) ##1
       lc_tx_test_true_strict(strap_hw_debug_en_q)
       |->
-      $past(strap_en_i))
+      $past(strap_en))
```
rswarbrick, line 274: Again, I think this can be squashed with "[hw,rv_dm,rtl] Enable RV_DM to check straps without the pinmux"?

hw/ip/rv_dm/dv/env/seq_lib/rv_dm_tap_fsm_vseq.sv
Razer6: Lets drop that then :)

rswarbrick:
Thanks for rebasing / squashing etc. I think there are still a few comments that are open:
- https://github.com/lowRISC/opentitan/pull/24790#discussion_r1815014196
- https://github.com/lowRISC/opentitan/pull/24790#discussion_r1815022088
- https://github.com/lowRISC/opentitan/pull/24790#discussion_r1817055309
- https://github.com/lowRISC/opentitan/pull/24790#discussion_r1817063612
- https://github.com/lowRISC/opentitan/pull/24790#discussion_r1817082352

Razer6:
I know. This PR is getting scheduled soon on my agenda again :-)

hw/ip/rv_dm/dv/tb.sv
Razer6: That's interesting, I was using the same pattern as in https://github.com/lowRISC/opentitan/pull/24507. Does that mean I'm testing the DMI case there?

hw/ip/rv_dm/data/rv_dm.hjson
Razer6: finally fixed.

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
Razer6: Yep added a better comment to both lines.

hw/ip/rv_dm/rtl/rv_dm.sv
Razer6: Added a better comment for both (JTAG and DMI)

hw/ip/rv_dm/dv/tb.sv
rswarbrick: Oops, looks like it!

hw/ip/rv_dm/rtl/rv_dm.sv
Razer6: I moved those signals in, same as the assignments below. The Bound-in module needs to define `dmi_req_valid`  and `dmi_rsp_ready`.

hw/ip/rv_dm/dv/tb.sv
rswarbrick: I think this still applies?

hw/ip/rv_dm/rtl/rv_dm.sv
rswarbrick: I'm really sorry but I think this might still be backwards? At the moment, we have:
```
    // If DMIDirectTAP is not defined, a bound-in DPI module replaces the TAP that s defined
    // within the ifndef block
```
Doesn't the DPI module replace the tap when it *is* defined? (Also, missing apostrophe in "that's")

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -0,0 +1,284 @@
...
+    // Check that once the strap_en_override_i is asserted to 1 it always stays at 1
+    `ASSUME(LcCtrlStrapSampleOverrideOnce_A,
+        $rose(strap_en_override_i) |-> always strap_en_override_i)
+
+  end else begin : gen_no_strap_override
+    logic unused_strap_en_override;
+    assign unused_strap_en_override = strap_en_override_i;
+    assign strap_en                 = strap_en_i;
+
+    // Ehsure that strap is never asserted
```
rswarbrick, line 124: Nit: I think this should be "Ensure"

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -0,0 +1,284 @@
...
+      end else begin
+        strap_en_override_q <= strap_en_override_d;
+      end
+    end
+
+    // Detect a posedge on the override signal and reflect it for just a cycle on strap_en
+    // (ignoring the fact that strap_en_override_i will stay high afterwards).
+    assign strap_en = strap_en_i || (strap_en_override_d && !strap_en_override_q);
+
+    // Check that once the strap_en_override_i is asserted to 1 it always stays at 1
```
rswarbrick, line 115: Nit: I think I'd probably choose a different verb. I realise that an `ASSUME` implies a check in simulation, but the implication is probably the other way around. Maybe just a bare statement? "Once strap_en_override_i is asserted, it will always stay high."?

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
```diff
@@ -0,0 +1,284 @@
...
+    `ASSUME(LcCtrlStrapSampleOverrideOnce_A,
+        $rose(strap_en_override_i) |-> always strap_en_override_i)
```
rswarbrick, line 117: It really doesn't matter, but isn't this an easier way to encode the statement? `strap_en_override_i |=> strap_en_override_i`

rswarbrick (commented):
A couple of nits about the fifth commit, but this looks pretty good to me. I think these two comments still remain?

- https://github.com/lowRISC/opentitan/pull/24790#discussion_r1815014196
- https://github.com/lowRISC/opentitan/pull/24790#discussion_r1817055309

hw/ip/rv_dm/dv/tb.sv
Razer6: Now it's fixed.

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
Razer6: I also changed the name of the assertion itself.

hw/ip/rv_dm/rtl/rv_dm.sv
Razer6: Sure 🤦

hw/ip/rv_dm/dv/tb.sv
rswarbrick: Ahah, cool: we're only defining it as 1 now. Looks good to me.

rswarbrick (approved):
Thanks for all the changes based on the review comments. This looks good to me.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/rv_dm/data/rv_dm.hjson
CHANGE AUTHORIZED: hw/ip/rv_dm/rtl/rv_dm_dbg_reg_top.sv
CHANGE AUTHORIZED: hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
CHANGE AUTHORIZED: hw/ip/rv_dm/rtl/rv_dm_reg_pkg.sv

The changes in `rv_dm` are carefully motivated (and shouldn't affect the existing setup). Changes in `top_earlgrey` are just setting parameters and turning the new feature off.

andreaskurth (approved):
LGTM, and the most substantial changes have already been reviewed on `integrated_dev`.

andreaskurth:
CHANGE AUTHORIZED: hw/ip/rv_dm/data/rv_dm.hjson
CHANGE AUTHORIZED: hw/ip/rv_dm/rtl/rv_dm_dbg_reg_top.sv
CHANGE AUTHORIZED: hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
CHANGE AUTHORIZED: hw/ip/rv_dm/rtl/rv_dm_reg_pkg.sv

Agree with @rswarbrick's assessment above

hw/ip/rv_dm/rtl/rv_dm_dmi_gate.sv
Razer6: @rswarbrick I am not sure this assertion makes sense. Why should it stay low?

rswarbrick: Hmm, good point. I think that maybe I took the name "`gen_no_strap_override`" to mean that we don't ever enable the straps. But that's completely wrong! Should we just be deleting `A_StrapNeverAsserted`?

PR #24791, opened 2024-10-15
[multitop] New bazel rules to describe HW

pamaury:
The goal of this experiment is to design new rules to better describe the HW and eventually get rid of our current usage of `all_files` targets. The idea is that each IP contains a new target created by `opentitan_ip` where the RTL, doc and hjson files are listed. Then, those files targets are collected by an `opentitan_top` rule that puts together the entire list of doc, RTL and hjson files. One this is done, we start replace all direct usages of HW files by reference to this new top target where we selectively extra only the files we need. Eventually this should apply to:
- all headers (C, rust, multi-top)
- fusesoc
- documentation

More is possible, for example we could point to testplans, we could separate RTL from DV, etc
This change will eventually be needed for multi-top because collecting all files as we currently do will not scale.

Finally, this PR contains another change: instead of pointing to //hw/ip/... for headers, we now collect everything in a new package //hw/top. There are several reasons for that:
- we can start treating all IPs the same way, whether they are in hw/ip or hw/top_<x>/ip/..., autogenerated or not
- this moves us closer to a bazel-configuration-based top selection which will be used for multi-top
- it solves the problem of IPs in hw/top that have top-specifiied parameters but those are not accessible when generating headers the way we currently do.

Since there are many IPs to convert and many files to patch, the approach in this PR is to create a script to automate most of the process. All commits before `[util] Add script to convert...` are manual. The rest is created automatically by the script to convert IPs and uses one by one, except for the **last** commit is manual.

**Important remark:** this does not duplicate any information that fusesoc has, this is completely orthogonal and only use coarse-grained globs to gather all files.

TODO:

- [ ] see if we need to handle the `node` attribute of previous C/rust headers to select the default register node, or if we need to generate several headers for some IPs (one per node)

TODO in follow-up PRs:

- cleanup the OTBN python: it's a complete mess, the way data files are handled and the the way it hack `sys.path` to point to `util/` is not great. I couldn't quite get rid of `all_files` due to that.


pamaury:
After discussing with @HU90m about the documentation, I have decided to drop the `doc` attribute of the rules for now. Indeed, it appears that the doc is mixed with the RTL and which files get included is based on what the `README.md` specifies, in particular it may not be in the `doc/` subdirectory.

I have also changed the `rtl` attribute of most IPS to include `glob(["**"])` instead of more fined grained dictories to keep as close as possible to the original behaviour, in case some downstream users were relying on this.

Razer6:
What's the plan of converting the new IPs such as:
* DMA
* MBX
* KeymgrDPE
* SocProxy

Razer6:
I don't think it's worth the effort. I plan to add the top_darjeeling.hjson by the end of the week.

pamaury:
Following the feedback on the RFC, I have renamed the `rtl` attribute to `files` to make it clearer that (1) this is more general than hardware files, (2) this is just a bazel "visibility" thing, we not trying to reimplement fusesoc here.

rules/autogen.bzl
```diff
@@ -87,6 +88,99 @@ autogen_hjson_c_header = rule(
     },
 )
 
+def _opentitan_ip_c_header_impl(ctx):
+    header = ctx.actions.declare_file("{}_regs.h".format(ctx.attr.ip))
```
matutem, line 92: Use f-strings: `...._file(f"{ctx.attr.ip}_regs.h")`

sw/device/lib/crypto/drivers/BUILD
```diff
@@ -24,7 +24,7 @@ cc_library(
     hdrs = ["aes.h"],
     deps = [
         ":entropy",
-        "//hw/ip/aes/data:aes_c_regs",
+        "//hw/top:aes_c_regs",
```
matutem, line 27: Too bad the dep below is top-specific. I think it could be expressed in a top-independent way?

sw/device/lib/dif/BUILD
```diff
@@ -37,7 +37,7 @@ cc_library(
     ],
     deps = [
         ":base",
-        "//hw/ip/adc_ctrl/data:adc_ctrl_c_regs",
```
matutem, line 40: Same as for lib/crypto/drivers. These are very regular rules for dif, drivers, and silicon_creator/drivers (at least).

sw/device/lib/testing/test_rom/BUILD
```diff
@@ -77,10 +77,10 @@ cc_library(
     # This should be built only for english breakfast and skipped if using wildcards.
     tags = ["manual"],
     deps = [
-        "//hw/ip/gpio/data:gpio_c_regs",
-        "//hw/ip/otp_ctrl/data:otp_ctrl_c_regs",
-        "//hw/ip/spi_device/data:spi_device_c_regs",
-        "//hw/top_earlgrey/ip_autogen/flash_ctrl:flash_ctrl_c_regs",
+        "//hw/top:flash_ctrl_c_regs",
```
matutem, line 80: Can we generate the test_rom_libs in a loop over tops?

sw/device/silicon_creator/lib/drivers/BUILD
```diff
@@ -497,9 +496,9 @@ cc_library(
     srcs = ["pinmux.c"],
     hdrs = ["pinmux.h"],
     deps = [
-        "//hw/ip/gpio/data:gpio_c_regs",
-        "//hw/ip/otp_ctrl/data:otp_ctrl_c_regs",
-        "//hw/top_earlgrey/ip_autogen/pinmux:pinmux_c_regs",
+        "//hw/top:gpio_c_regs",
+        "//hw/top:otp_ctrl_c_regs",
+        "//hw/top:pinmux_c_regs",
         "//hw/top_earlgrey/sw/autogen:top_earlgrey",
```
matutem, line 502: Could this top-specific autogen be selected via //hw/top: ?

sw/device/tock/BUILD
```diff
@@ -9,47 +9,47 @@ package(default_visibility = ["//visibility:public"])
...
-        "//hw/ip/rom_ctrl/data:rom_ctrl_rust_regs",
-        "//hw/ip/rv_core_ibex/data:rv_core_ibex_rust_regs",
-        "//hw/ip/rv_timer/data:rv_timer_rust_regs",
-        "//hw/ip/spi_device/data:spi_device_rust_regs",
-        "//hw/ip/spi_host/data:spi_host_rust_regs",
-        "//hw/ip/sram_ctrl/data:sram_ctrl_rust_regs",
-        "//hw/ip/sysrst_ctrl/data:sysrst_ctrl_rust_regs",
-        "//hw/ip/uart/data:uart_rust_regs",
-        "//hw/ip/usbdev/data:usbdev_rust_regs",
+        "//hw/top:adc_ctrl_rust_regs",
```
matutem, line 12: Could this list of _regs targets be provide by top?

util/rewrite_hw.py
```diff
@@ -0,0 +1,292 @@
...
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import subprocess
+import os
+import sys
+from pathlib import Path
+
```
matutem, line 11: Please add a doc string for this interesting script, so it can tell what it does.

util/rewrite_hw.py
```diff
@@ -0,0 +1,292 @@
...
+    return lines[:start_idx] + lines[end_idx + 1:], start_idx, lines[start_idx:end_idx + 1]
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v', '--verbose', action='store_true')
+    parser.add_argument('ip_dir', help='Path to the IP directory')
+    parser.add_argument('--root', help='Path to the project root (if not specified, assume CWD)')
+    parser.add_argument('-t', '--top', action='append', default=[],
+                        help='Path to the top to edit (optional,c an be specified several times)')
```
matutem, line 62: `... optional,c an...` -> `... optional, can...`

util/rewrite_hw.py
```diff
@@ -0,0 +1,292 @@
...
+
+    if hjson_path is not None:
+        assert hjson_path.exists() and hjson_path.is_file(), \
+            f"HSJON path {hjson_path} invalid (must exists and be a file)"
+        # Find the BUILD file for the hjson.
+        hjson_build_path = (hjson_path.parent / 'BUILD').resolve()
+        # We may need to go up several level for some ip_templates that don't have
+        # a BUILD file in data/
+        while not hjson_build_path.exists() and \
+                hjson_build_path.parent.parent.is_relative_to(project_root):
```
matutem, line 99: Maybe .parent.parent -> parents[2] ?

matutem (commented):
Some suggestions so far, and I hope you can figure out how to remove some of the top-specific dependencies and provide list of ips via top providers.

rules/autogen.bzl
pamaury: Unfortuntaley starlark doesn't support f-string

sw/device/lib/crypto/drivers/BUILD
pamaury: I am not entirely sure I understand your comment @matutem . Are you saying that you would prefer that `aes_c_regs` was not a top-specific dependency?

sw/device/silicon_creator/lib/drivers/BUILD
pamaury: Yes in the multi-top SW PR we have an alias `//hw/top:top_lib` for that. It's just not that in that PR.

sw/device/lib/testing/test_rom/BUILD
pamaury: The `test_rom` situation is a bit of mess: the englishbreakfast test_rom is very hacky and doesn't use the same files as earlgrey (e.g. it stubs out some some functions related to non-existent IPs) which makes this a very manual process. I think we will need to revisit how the `test_rom` is built and try to avoid having one `cc_library` per top if possible.

sw/device/tock/BUILD
pamaury: Yes we could create a single target that collects all rust_regs into on in `//hw/top/`

util/rewrite_hw.py
pamaury: I can add more doc, however this script will disappear once all IPs are converted, it might even be deleted as part of this PR....

pamaury:
Closed in favor of #25580

PR #24792, opened 2024-10-15
[sival,kmac] chip_sw_kmac_kmac_key_sideload

nasahlpa:
This commit adds the chip_sw_kmac_kmac_key_sideload test. In this test, a sideloaded key is used in KMAC. The output is stored in retention SRAM, a reset is performed, and it is checked whether the same KMAC output is generated.

### DIFF
--- a/sw/device/tests/kmac_kmac_key_sideload_test.c
+++ b/sw/device/tests/kmac_kmac_key_sideload_test.c
@@ -1,0 +1,216 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "sw/device/lib/base/mmio.h"
+#include "sw/device/lib/dif/dif_keymgr.h"
+#include "sw/device/lib/dif/dif_kmac.h"
+#include "sw/device/lib/dif/dif_rstmgr.h"
+#include "sw/device/lib/runtime/ibex.h"
+#include "sw/device/lib/runtime/log.h"
+#include "sw/device/lib/testing/keymgr_testutils.h"
+#include "sw/device/lib/testing/kmac_testutils.h"
+#include "sw/device/lib/testing/ret_sram_testutils.h"
+#include "sw/device/lib/testing/rstmgr_testutils.h"
+#include "sw/device/lib/testing/sram_ctrl_testutils.h"
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+#include "sw/device/silicon_creator/lib/drivers/retention_sram.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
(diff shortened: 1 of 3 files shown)


engdoreis (approved):

sw/device/tests/kmac_kmac_key_sideload_test.c
```diff
@@ -0,0 +1,217 @@
...
+                                         kCustomString, kCustomStringLen,
+                                         kKmacMessage, kKmacMessageLen,
+                                         kKmacOutputLen, output_sideload_bad,
+                                         /*capacity=*/NULL)));
+    LOG_INFO(
+        "Ran KMAC with an invalid sideload key and checked that it fails.");
+
+    // Verify that the output array did not get overwritten.
+    for (size_t i = 0; i < kKmacOutputLen; i++) {
+      CHECK(output_sideload_bad[i] == i);
```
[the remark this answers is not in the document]
nasahlpa: I think using randomness isn't strictly needed for this test. If ok, I would stick to the current implementation.

vogelpi, line 173: I think the test implementation currently doesn't fully match the testplan and the description above.
- IIUC, we check here that if the sideload key is invalid, KMAC doesn't produce an output and instead reports an error. 
- The testplan specifies we should check that the KMAC cannot produce the same output as before. The description above states the same (use a software key and ensure the output doesn't match the sideload key).

These are two different things. Maybe we should check both. But what is important is that test and testplan are aligned.

vogelpi (commented):
Thanks @nasahlpa for the PR. This looks mostly good, I have one question regarding the testing the output after clearing the sideload key.

Since @andrea-caforio is working on keymgr tests, it probably makes sense to also get his view on this.

sw/device/tests/kmac_kmac_key_sideload_test.c
nasahlpa: Yes, you  are right. Currently, I am only testing whether after clearing the sideload key, KMAC does not produce and output.

As this is not specified in the test description:
```
Clear sideload key from keymgr.
Verify that KMAC can no longer reproduce the same KMAC output.
```
What should I do between these two steps? Create a new sideload key with a different salt and check whether the output does not match? Or should I switch from a sideload key to a software key and do the comparison?

vogelpi: I thought of using a software key but using the sideload key with different salt sounds like a better idea to me! Do you think that can be done easily or would it be a lot of work?

nasahlpa: Done, I've also adapted the test plan description accordingly.

Testing whether after clearing the sideload key KMAC is no longer able to produce a valid output is actually already tested in `keymgr_sideload_kmac_test`.

vogelpi: Thanks @nasahlpa , this looks good now!

vogelpi (approved):

nasahlpa:
Thanks for the review @vogelpi & @engdoreis.

PR #24802, opened 2024-10-17
[dv, bazel] : run opentitantool with shared libs outside system lib path

venkatk-ot:
This commit enables installing opentitantool to find and link to its library dependencies like libftdi, libusb, libudev that are installed in
 user's work area. This is particularly useful when the user's compute
enironment does not have these libraries preinstalled as part of the system and the user doesn't have root permissions to install these libraries using syatem package managers. While the custom paths to the headers and shared libs needed for building the opentitantool can be specified by setting the PKG_CONFIG_PATH environment variable through the bazel build action_env property in site specific .bazelrc-site file, these paths are not available in the bazel sandbox that runs the opentitantool to generate the otp json.
for example, bazel build //hw/ip/otp_ctrl/data:otp_json_alert_digest_cfg. Attempts to pass in the user provided LD_LIBRARY_PATH using the bazel build action_env fail as this doesn't get propagated to the bazel subcommand that runs the opentitantool. Hence adding the LD_LIBRARY_PATH to the nonhermetic.bzl file to make this part of the ENV available to the
 opentitantool run sandbox.

### DIFF
--- a/rules/otp.bzl
+++ b/rules/otp.bzl
@@ -198,11 +198,20 @@ def _otp_alert_digest_impl(ctx):
     args.add_all(("--rcfile=", "otp", "alert-digest", ctx.file.otp_img))
     args.add("--output", file)
 
+    # By default, no environment variables are available to the action
+    # during bazel builds. This can not be changed with action_env.
+    # However, the opentitantool binary used in building the json target
+    # needs access to LD_LIBRARY_PATH to load shared libraries
+    # outside standard system paths.
+    # Setting use_default_shell_env to True allows passing LD_LIBRARY_PATH
+    # to the sandbox using --action_env switch
+
     ctx.actions.run(
         outputs = outputs,
         inputs = inputs,
         arguments = [args],
         executable = tc.tools.opentitantool,
+        use_default_shell_env = True,
     )
     return [DefaultInfo(files = depset([file]))]
 


venkatk-ot:
This fix is needed to run chip level sw tests in DV on our compute farm that doesn't have libftdi , libusb, libdev etc.. preinstalled as part of system libraries on the compute node and instead has these third_party libs installed on a central shared NFS location
Here's an example of what I mean by running a chip level SW test in DV
util/dvsim/dvsim.py hw/top_earlgrey/dv/chip_sim_cfg.hjson -i chip_sw_aes_smoketest

pamaury:
I will let @cfrantz have a look at this but I am a bit wary about doing this change: we have many rules that depend on opentitantool, so once we start making one nonhermetic for this reason, we could be down a slippery slope very quickly.
Is it possible that you might need to use [--host_action_env](https://bazel.build/reference/command-line-reference#flag--host_action_env) instead of `--action_env` in your case?

johngt:
Adding @jwnrt and @pamaury .
It looks like @pamaury has already looked at this (see comment above) but would like @jwnrt feedback on this too.

timothytrippel (changes requested):
Yes, I don't think this change should be required, and agree we should keep the build of opentitantool/lib hermetic as this is critical tooling/infrastructure for the project. @venkatk-ot let me see if I can help you solve your issue offline.

venkatk-ot:
I tried @pamaury suggestion about setting host_action_env but that didn't work
I dont know how to implement @jwnrt's suggestion as I am not familiar with bazel or rust. If there is a how-to guide, I can take a look. 
Meanwhile I debugged this issue with @timothytrippel offline. I uploaded a new patch with use_default_shell_env=True which then allows action_env LD_LIBRARY_PATH to be passed when running the opentitantool in bazel sandbox

venkatk-ot:
> I tried @pamaury suggestion about setting host_action_env but that didn't work I dont know how to implement @jwnrt's suggestion as I am not familiar with bazel or rust. If there is a how-to guide, I can take a look. Meanwhile I debugged this issue with @timothytrippel offline. I uploaded a new patch with use_default_shell_env=True which then allows action_env LD_LIBRARY_PATH to be passed when running the opentitantool in bazel sandbox

Is this acceptable ?

timothytrippel:
Yes, I worked with @venkatk-ot offline and this was the behavior we were noticing: https://groups.google.com/g/bazel-discuss/c/jHaj11mm6sU/m/vafDTu5wEgAJ

Essentially, for `bazel build ...` operations `action_env` and `host_action_env` are not passed through to the sandbox unless the `use_default_shell_env  = True` is added to the `ctx.actions.run()` invocation. However, this seems to also pass through the `PATH` envar by default on Linux. Sadly I could not come up with a better solution.

Another option would be to only pipe through LD_LIBRARY_PATH via the `env` [parameter](https://bazel.build/rules/lib/builtins/actions#run), but setting `use_default_shell_env` seemed most flexible.

rules/otp.bzl
```diff
@@ -203,6 +203,7 @@ def _otp_alert_digest_impl(ctx):
         inputs = inputs,
         arguments = [args],
         executable = tc.tools.opentitantool,
+        use_default_shell_env = True,
```
timothytrippel, line 206: Could you add a comment here as to why this is needed?

jwnrt (approved):
Thanks, this seems fine to me

rules/otp.bzl
venkatk-ot: Done. Thanks

pamaury:
> Yes, I worked with @venkatk-ot offline and this was the behavior we were noticing: https://groups.google.com/g/bazel-discuss/c/jHaj11mm6sU/m/vafDTu5wEgAJ
> 
> Essentially, for `bazel build ...` operations `action_env` and `host_action_env` are not passed through to the sandbox unless the `use_default_shell_env = True` is added to the `ctx.actions.run()` invocation. However, this seems to also pass through the `PATH` envar by default on Linux. Sadly I could not come up with a better solution.
> 
> Another option would be to only pipe through LD_LIBRARY_PATH via the `env` [parameter](https://bazel.build/rules/lib/builtins/actions#run), but setting `use_default_shell_env` seemed most flexible.

Thanks for the explanation @timothytrippel This is indeed a very curious behaviour. Using `use_default_shell_env` seems like a reasonable compromise in that case, even though I don't really like the idea of passing `PATH` by default.

moidx:
Hi @pamaury, I think it makes sense to support the `env` configuration approach you are proposing in your previous comment. I have merged this PR to unblock top level regressions in other environments. If it makes sense, lets open an issue to track your proposal, or you can also track it as part of the multi-top build system changes. 

Thanks!

PR #24803, opened 2024-10-17
[DO NOT MERGE][pinmux] Remove outdated file hw/ip/pinmux/data/pinmux.hjson

pamaury:
This file should not be used: the actual hsjon file is created by topgen based on a template. Therefore using this file could lead to using outdated data. There were only two references to it that this commit points to the correct location. This commit also removes two bazel targets for headers based on this file.


jwnrt (approved):

pamaury:
In a bizarre twist, it appears that `hw/ip/pinmux/data/pinmux.json` is used by topgen with a [huge hack](https://github.com/lowRISC/opentitan/blob/master/util/topgen.py#L758) where the first pass needs an hjson file that will later be generated. Since `pinmux` is not a proper IP template, this default hjson cannot be generated cleanly from the template so it uses this outdated file... I am honestly not use why this works and I am not super eager to find out. I will keep the file but remove all users except for topgen. The proper solution is to make pinmux a proper IP template. See #8440

pamaury:
This PR will probably be dropped/updated after #24808 is merged

PR #24812, opened 2024-10-18
[bazel] Move consts from legacy rule files

jwnrt:
This PR moves the remaining consts from the legacy `opentitan.bzl` and `opentitan_test.bzl` files to the `//rules/opentitan` package.

I've moved the key utils associated with these consts as well for simplicity.


pamaury (approved):
This looks good to me, thanks James

PR #24813, opened 2024-10-18
[bazel] Move remaining legacy rules to `//rules/opentitan` package

jwnrt:
These rules from `//rules:opentitan.bzl` are still used, and most have implementations duplicated in the `//rules/opentitan` package.

* `manual_test` moved to `//rules/opentitan:manual.bzl`
* `obj_transform`, `vmem_file`, and `scramble_flash_vmem` rules moved to `//rules/opentitan:cc.bzl` and made to use the implementations in `//rules/opentitan:transform.bzl`
* `elf_to_disassembly` is no longer used, so the `loads` are removed.

I'll remove `rules/opentitan.bzl` and `rules/opentitan_test.bzl` in a follow-up PR to keep this one simple.

### DIFF
--- a/sw/device/silicon_owner/tock/kernel/BUILD
+++ b/sw/device/silicon_owner/tock/kernel/BUILD
@@ -4,7 +4,6 @@
 
 load("@rules_rust//rust:defs.bzl", "rust_binary")
 load("//rules:linker.bzl", "ld_library")
-load("//rules:opentitan.bzl", "elf_to_disassembly")
 load("//rules/opentitan:legacy.bzl", "obj_transform")
 load("//rules/opentitan:defs.bzl", "OPENTITAN_CPU")
 
--- a/sw/device/silicon_owner/tock/upstream_kernel/BUILD
+++ b/sw/device/silicon_owner/tock/upstream_kernel/BUILD
@@ -4,7 +4,6 @@
 
 load("@rules_rust//rust:defs.bzl", "rust_binary")
 load("//rules:linker.bzl", "ld_library")
-load("//rules:opentitan.bzl", "elf_to_disassembly")
 load("//rules/opentitan:defs.bzl", "OPENTITAN_CPU")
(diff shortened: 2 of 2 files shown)


jwnrt:
CI failures are unrelated: the FPGA test failure (`//sw/device/tests:i2c_target_test_fpga_cw310_sival`) does not seem relevant, and the `Verify FPGA jobs` failure is due to a race in our pipeline (see #25164).

engdoreis (approved):
Looks sensible to me.

rules/opentitan/cc.bzl
```diff
@@ -477,3 +479,83 @@ opentitan_binary_assemble = rule(
     },
     toolchains = [LOCALTOOLS_TOOLCHAIN],
 )
+
+def _obj_transform_impl(ctx):
```
cfrantz, line 483: I'd like to see these functions moved to something like //rules/opentitan:legacy.bzl, as they each have exactly one use in the codebase and some future work should evaluate whether or not they can simply be eliminated.

rules/opentitan.bzl
```diff
@@ -50,41 +50,6 @@ PER_DEVICE_DEPS = {
 def flatten(l):
     return [item for ll in l for item in ll]
 
-def _obj_transform_impl(ctx):
```
cfrantz, line 53: Are there any uses of the remaining macro "rules" in this file (like `opentitan_binary`, `opentitan_rom_binary`, etc)?

They should be candidates for removal as well.

cfrantz (approved):

rules/opentitan.bzl
cfrantz: nvm: just re-read your PR description.

jwnrt:
CI failure (`//sw/device/tests:i2c_target_test_fpga_cw310_sival`) looks unrelated to me, I won't waste FPGA time rerunning it

PR #24817, opened 2024-10-18
[dv, dvsim] Add job queue/lic wait timeout in addition to run timeout

venkatk-ot:
The timeout_mins specified in the test list json file or on dvsim command
 line is intended to be a timeout on the wall time elapsed after the job
has started running.It is not intended to count in the time spent in waiting for a compute node and/or a license to become available. Despite best efforts to specify license resource strings, the availability of a compute node doesn't always guarantee the availability of a license as this scheme only works when all jobs from all projects running on the cluster adhere to this license requesting scheme which sadly can't be enforced in a fool proof manner.
In the proposed scheme, the file size as a proxy to determine if the job has really started running. While this is not one hundred percent accurate, it prevents false timeouts on jobs/tests that haven't even had a chance to start running because of unavailability of compute or license
 resources. The file size threshold to determine whether the job has
started running has been chosen after carefully analyzing log files of jobs executed successfuly and jobs that were killed while waiting

### DIFF
--- a/util/dvsim/NcLauncher.py
+++ b/util/dvsim/NcLauncher.py
@@ -71,6 +71,12 @@ class NcLauncher(Launcher):
                 license_args +
                 ['--', f'{odir}/run.sh'])
 
+    def _pre_launch(self):
+        # store the start_time (correspoinding to job wait time counter)
+        super()._pre_launch()
+        # set the nc_job_state to the initial state - waiting for resource
+        self.nc_job_state = 'waiting'
+
     def _do_launch(self):
         # Compute the environment for the subprocess by overriding environment
         # variables of this process with matching ones from self.deploy.exports
@@ -88,6 +94,10 @@ class NcLauncher(Launcher):
 
         self._dump_env_vars(exports)
 
+        # For reruns, delete the log file of the past run to avoid any race
+        # condition between the log file getting updated for the new run
+        # versus the logic that distinguishes the job wait versus run times.
+        rm_path(self.deploy.get_log_path())
         # using os.open instead of fopen as this allows
         # sharing of file descriptors across processes
         fd = os.open(self.deploy.get_log_path(), os.O_WRONLY | os.O_CREAT)
@@ -95,7 +105,7 @@ class NcLauncher(Launcher):
         os.set_inheritable(fd, True)
         message = '[Executing]:\n{}\n\n'.format(self.deploy.cmd)
         fobj.write(message)
-
+        fobj.flush()
         if self.deploy.sim_cfg.interactive:
             # Interactive: Set RUN_INTERACTIVE to 1
             exports['RUN_INTERACTIVE'] = '1'
@@ -138,6 +148,9 @@ class NcLauncher(Launcher):
 
         self._link_odir('D')
 
+    def minutes_since_start(self):
+        return (datetime.datetime.now() - self.start_time).total_seconds() / 60
+
     def poll(self):
         """Check status of the running process.
 
@@ -150,14 +163,44 @@ class NcLauncher(Launcher):
         """
 
         assert self.process is not None
-        elapsed_time = datetime.datetime.now() - self.start_time
-        job_runtime_secs = elapsed_time.total_seconds()
         if self.process.poll() is None:
-            timeout_mins = self.deploy.get_timeout_mins()
-            if timeout_mins is not None and not self.deploy.gui:
-                if job_runtime_secs > (timeout_mins * 60):
+            run_timeout_mins = self.deploy.get_timeout_mins()
+            if run_timeout_mins is not None and not self.deploy.gui:
+                wait_timeout_mins = 180  # max wait time in job / license queue
+                # We consider the job to have started once its log file contains
+                # something. file_size_thresh_bytes is a threshold: once the log
+                # file is bigger than this many bytes, the job must have started
+                file_size_thresh_bytes = 5120  # log file size threshold
+
+                # query the log file size
+                f_size = os.path.getsize(self.deploy.get_log_path())
+
+                if f_size >= file_size_thresh_bytes:
+                    if (self.nc_job_state == 'waiting'):
+                        # If the job log size is more than the threshold,
+                        # declare the job to have started running
+                        # capture the run start time
+                        self.start_time = datetime.datetime.now()
+                        self.nc_job_state = 'running'
+
+                if self.nc_job_state == 'waiting':
+                    # If we get here, we know the log size is less than the threshold.
+                    # check if wait_timeout_mins has elapsed.
+                    if self.minutes_since_start() > wait_timeout_mins:
+                        self.nc_job_state = 'wait_timeout'
+
+                if self.nc_job_state == 'running':
+                    if self.minutes_since_start() > run_timeout_mins:
+                        self.nc_job_state = 'run_timeout'
+
+                if self.nc_job_state in {'wait_timeout', 'run_timeout'}:
                     self._kill()
-                    timeout_message = f'Job timed out after {timeout_mins} mins'
+                    if self.nc_job_state == 'run_timeout':
+                        timeout_message = f'Job timed out after running ' \
+                                          f'{run_timeout_mins} mins'
+                    elif self.nc_job_state == 'wait_timeout':
+                        timeout_message = f'Job timed out after waiting ' \
+                                          f'{wait_timeout_mins} mins'
                     self._post_finish('K',
                                       ErrorMessage(line_number=None,
                                                    message=timeout_message,


johngt:
Assigning to @luismarques to forward assign. @rswarbrick is away.
There other person that may be able to look into this is @hcallahan-lowrisc . Whoever gets to it first and can check on it please remove the other from the reviewers.

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
 
         assert self.process is not None
-        elapsed_time = datetime.datetime.now() - self.start_time
-        job_runtime_secs = elapsed_time.total_seconds()
         if self.process.poll() is None:
-            timeout_mins = self.deploy.get_timeout_mins()
-            if timeout_mins is not None and not self.deploy.gui:
-                if job_runtime_secs > (timeout_mins * 60):
+            run_timeout_mins = self.deploy.get_timeout_mins()
+            wait_timeout_mins = 180  # max wait time in job queue / license
```
rswarbrick, line 162: Nit: I think you probably want "job / license queue" here?

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
+            file_size_thresh_bytes = 5000  # log file size threshold
```
rswarbrick, line 163: Probably worth leaving a note here to explain what's going on? (I'd probably also round it up to a multiple of 1024, but maybe that's just me :-D)

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
+            if run_timeout_mins is not None and not self.deploy.gui:
```
rswarbrick, line 164: I'm not sure about the gui term here. I wouldn't be happy with waiting 3 hours for the window to come up anyway :-)

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
+                job_waittime = 0
+                job_runtime = 0
```
rswarbrick, line 166: I'd be slightly inclined not to set these variables here. We don't have a meaningful value for them anyway.

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
+                # query the log file size
+                f_size = os.path.getsize(self.deploy.get_log_path())
+                if f_size < file_size_thresh_bytes:
+                    # if the job log size is less than the threshold, increment
+                    # the job wait time counter. hold the run time counter at 0
+                    self.start_run_time = 0
```
rswarbrick, line 172: Again, I don't think there's any need to talk about `start_run_time` here at all?

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
+                    elapsed_wait_time = (datetime.datetime.now() -
+                                         self.start_time)
+                    job_waittime = elapsed_wait_time.total_seconds() / 60
```
rswarbrick, line 175: This is duplicated across the two sides of the `if` block. Probably better to move it afterwards.

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
+                else:
+                    # if the job log size is more than the threshold, start the
+                    # the job run time counter if has not already started.
+                    # Hold the wait time counter at the last updated value
+                    if (self.start_run_time == 0):
+                        self.start_run_time = datetime.datetime.now()
```
rswarbrick, line 181: I'm not sure I properly understand this. Don't we want to pull this check to come at the start of this function, rather than sitting in the if/else block?

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +157,51 @@ def poll(self):
...
+                    elapsed_run_time = (datetime.datetime.now() -
+                                        self.start_run_time)
+                    job_runtime = elapsed_run_time.total_seconds() / 60
+                # if the job run has not started within the max time specified
+                # in this file or if the job has been running for longer than
+                # the time specified in the block json or via dvsim switches
+                # terminate it and send the appropriate error message
+                wait_timeout = (job_waittime > wait_timeout_mins)
+                run_timeout = (job_runtime > run_timeout_mins)
+                if run_timeout or wait_timeout:
```
rswarbrick, line 197: I believe this is correct, but it's pretty hard to read. Couldn't this be slightly restructured, to be a shape like this?
```python
  timeout_mins = None
  if self.start_run_time != 0:
    timeout_mins = run_timeout_mins
    phase_start_time = self.start_run_time
  elif self.start_wait_time != 0:
    timeout_mins = wait_timeout_mins
    phase_start_time = self.start_wait_time

  if timeout_mins is not None:
    elapsed = datetime.datetime.now() - phase_start_time
    elapsed_mins = elapsed.total_seconds() / 60
    if elapsed_mins > timeout_mins:
      timeout_message = '....'
      ...
```

util/dvsim/NcLauncher.py
```diff
@@ -71,6 +71,12 @@ def get_submit_cmd(self):
                 license_args +
                 ['--', f'{odir}/run.sh'])
 
+    def _pre_launch(self):
+        # start the start_time (correspoinding to job wait time counter)
```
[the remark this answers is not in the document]
venkatk-ot: done

util/dvsim/NcLauncher.py
```diff
@@ -71,6 +71,12 @@ def get_submit_cmd(self):
...
+        super()._pre_launch()
+        # hold the start_run_time at 0 (corresponding to job run time counter)
```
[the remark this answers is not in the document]
venkatk-ot: done

util/dvsim/NcLauncher.py
```diff
@@ -88,14 +94,15 @@ def _do_launch(self):
 
         self._dump_env_vars(exports)
 
+        rm_path(self.deploy.get_log_path())
```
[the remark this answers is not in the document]
venkatk-ot: yes, this is for rerun cases. added a comment. done

util/dvsim/NcLauncher.py
venkatk-ot: done

util/dvsim/NcLauncher.py
venkatk-ot: i have added a one line comment . and changed the threshold to 5120 / 5K

util/dvsim/NcLauncher.py
venkatk-ot: removed. done

util/dvsim/NcLauncher.py
venkatk-ot: removed. done

util/dvsim/NcLauncher.py
venkatk-ot: removed after rewrite . done

util/dvsim/NcLauncher.py
venkatk-ot: I have introduced a new class variable nc_job_state which acts like a state machine and keeps track of current state in a more user friendly manner instead of having to keep track of the actual wait/run time 0 or non zero value and infer the job state. This allows to remove the duplicated code in areas you have mentioned above at the small cost of having a new class variable

util/dvsim/NcLauncher.py
venkatk-ot: I didn't follow this approach as I had to change the logic to accommodate your other comments. Can you please check my new patch and let me know if I still need to reorder ?

util/dvsim/NcLauncher.py
rswarbrick: Oh, sorry: I was imagining something that gives the reader a bit of a hint about why this is being done. Would something like this work? "We consider the job to have started once its log file contains something. file_size_thresh_bytes is a threshold: once the log file is bigger than this many bytes, the job must have started."

util/dvsim/NcLauncher.py
rswarbrick: That seems like a good idea. Thanks!

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +160,49 @@ def poll(self):
...
+
+                if f_size >= file_size_thresh_bytes:
+                    if (self.nc_job_state == 'waiting'):
+                        # if the job log size is more than the threshold, start the
+                        # the job run time counter if has not already been started.
+                        self.start_time = datetime.datetime.now()
+                        self.nc_job_state = 'running'
+
+                if self.nc_job_state == 'waiting':
+                    # if the job log size is less than the threshold, increment
```
rswarbrick, line 183: I really like the structure of the if cases here! But I think this comment could be made slightly clearer. Something like this maybe? "If we get here, we know the log size is less than the threshold. Update the job wait time counter."

(And "increment" is probably the wrong verb: it's not like we've got `elapsed_wait_time += ...`)

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +160,49 @@ def poll(self):
...
+                    # the job wait time counter.
+                    elapsed_wait_time = (datetime.datetime.now() -
+                                         self.start_time)
+                    job_waittime_mins = elapsed_wait_time.total_seconds() / 60
```
rswarbrick, line 187: Thinking about it, it might be worth extracting this into a function that you can use for both the wait time and the run time. I'm imagining something like
```python
def minutes_since_start(self):
    return (datetime.datetime.now() - self.start_time).total_seconds() / 60
```

rswarbrick (commented):
A couple more small comments, but this makes sense to me.

Because GitHub "helpfully" collapses long lists of review comments, it's worth mentioning that I've got a note about the gui term on line 171.

util/dvsim/NcLauncher.py
venkatk-ot: I am not sure how my changes affect the gui mode as the 'if' statement is only entered when the gui mode is not set (and run_time_mins is set)
The problem with the original implementation was some tests would define run_time_mins as small as 1min and the polling logic didn't account for resource wait times
I have moved the setting of wait_timeout_mins and file_size_thresh_bytes and the associated comment inside the 'if' block

util/dvsim/NcLauncher.py
venkatk-ot: done . added clarity to the comment

util/dvsim/NcLauncher.py
venkatk-ot: Thanks for the suggestion. I have made this change

util/dvsim/NcLauncher.py
venkatk-ot: done. updated comment

rswarbrick: I'm really sorry to be annoying, but this probably needs to move down a line so that it's describing `file_size_thresh_bytes` instead of `wait_timeout_mins`.

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +163,44 @@ def poll(self):
...
+                        # If the job log size is more than the threshold,
+                        # declare the job to have started running
+                        # capture the run start time
+                        self.start_time = datetime.datetime.now()
+                        self.nc_job_state = 'running'
+
+                if self.nc_job_state == 'waiting':
+                    # If we get here, we know the log size is less than the threshold.
+                    # check if wait_timeout_mins has elapsed.
+                    if (self.minutes_since_start() > wait_timeout_mins):
```
rswarbrick, line 189: Sorry for the nitty comment, but I don't think you need the outer parens, so it can be
```python
                    if self.minutes_since_start() > wait_timeout_mins:
```
(and similarly on line 193)

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +163,44 @@ def poll(self):
...
+                        self.nc_job_state = 'wait_timeout'
+
+                if self.nc_job_state == 'running':
+                    if (self.minutes_since_start() > run_timeout_mins):
+                        self.nc_job_state = 'run_timeout'
+
+                if self.nc_job_state in ('wait_timeout', 'run_timeout'):
```
rswarbrick, line 196: Not a big deal, but I'd suggest using a list or set to avoid a reader thinking about membership of tuples. I just checked and this code does work (a surprise to me!) but it's not super obvious. I'd probably suggest `{ ... }`.

util/dvsim/NcLauncher.py
```diff
@@ -150,14 +163,44 @@ def poll(self):
...
                     self._kill()
-                    timeout_message = f'Job timed out after {timeout_mins} mins'
+                    if self.nc_job_state == 'run_timeout':
+                        timeout_message = f'Job timed out after running ' \
+                                          f'{run_timeout_mins} mins'
+                    if self.nc_job_state == 'wait_timeout':
```
rswarbrick, line 201: I realise that it's equivalent, but I'd suggest using `elif` here to make the structure a little more obvious.

rswarbrick (commented):
Some really minor comments, but this looks good to me.

util/dvsim/NcLauncher.py
venkatk-ot: done

util/dvsim/NcLauncher.py
venkatk-ot: done

util/dvsim/NcLauncher.py
venkatk-ot: done

util/dvsim/NcLauncher.py
venkatk-ot: done

rswarbrick (approved):
This looks really good, thanks!

PR #24823, opened 2024-10-20
[sival, keymgr] chip_sw_keymgr_derive_{sealing,attestation}

andrea-caforio:
This PR comprises two commits that ultimately lead to the implementation of `chip_sw_keymgr_derive_sealing` and `chip_sw_keymgr_derive_attestation`:

- Commit 1: The keymgr IP supports both the sealing and attestation CDI flows,
which is controlled by the `CDI_SEL` field in the `CONTROL_SHADOWED`
register. So far, this field was hardcoded to the sealing type in the
keymgr DIF. This commit parametrizes the field, making it possible to
also invoke the attestation flow through the DIF. In particular, the
commit comprises the following changes:

  - A new `dif_keymgr_cdi_type_t` enumeration type.
  - A new `dif_keymgr_identity_seed_params_t` struct such that the CDI
    flow can be chosen for the identity derivations.
  - A new `cdi_type` field in `dif_keymgr_versioned_key_params_t` to
    chose the CDI flow for software and sideload key derivations.
  - Added unit tests pertaining to the CDI type.
  - Updated some calls to this new API in other files.
- Commit 2: This commit adds the `chip_sw_keymgr_derive_{sealing,attestation}` tests
covering the complete testplan items, for the comprehensive set of
requirements see `hw/top_earlgrey/data/ip/chip_keymgr_testplan.hjson` in a
single file `keymgr_derive_cdi_test.c`.

### DIFF
--- a/sw/device/tests/keymgr_derive_cdi_test.c
+++ b/sw/device/tests/keymgr_derive_cdi_test.c
@@ -1,0 +1,439 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "sw/device/lib/testing/keymgr_testutils.h"
+#include "sw/device/lib/testing/otbn_testutils.h"
+#include "sw/device/lib/testing/ret_sram_testutils.h"
+#include "sw/device/lib/testing/rstmgr_testutils.h"
+#include "sw/device/lib/testing/sram_ctrl_testutils.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+#include "sw/device/silicon_creator/lib/drivers/retention_sram.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+static dif_keymgr_t keymgr;
+static dif_kmac_t kmac;
+static dif_otbn_t otbn;
+static dif_rstmgr_t rstmgr;
+static dif_sram_ctrl_t sram_ctrl;
+
+enum {
+  kKeymgrOutputSizeWords = 8,
+  kKeymgrOutputSizeBytes = 32,
+
+  kX2551PublicKeySizeBytes = 32,
+
+  // The retention SRAM testutils allocate some internal data as well as a
+  // number of counters; both of which should not be overwritten by this
+  // test. Hence, the actual base address is offset to account for this.
+  kRetSramBaseAddr = TOP_EARLGREY_RAM_RET_AON_BASE_ADDR +
+                     offsetof(retention_sram_t, owner) +
+                     4 * kRetSramTestutilsNumberOfCounters,
+
+// Compile-time directive to choose between CDI flow types. The default
+// derivation type is sealing.
+#ifndef DERIVE_ATTESTATION
+  kCdiType = kDifKeymgrSealingCdi
+#else
+  kCdiType = kDifKeymgrAttestationCdi
+#endif
+};
+
+/**
+ * Grouping of the three CDI output variants that can be generated in
+ * each key manager state (identity, versioned software key, sideload OTBN key).
+ *
+ * Note that the sideload OTBN key is not visible to software. In order to
+ * run the same verification steps as for the identity and software keys, a
+ * X25519 public key is generated in the OTBN and retrieved. For the sake of
+ * simplicity, and only in the confines of this test, we can assume that the
+ * X25519 public key and the sideload OTBN key refer to the same thing.
+ */
+typedef struct cdi_outputs {
+  dif_keymgr_output_t identity;
+  dif_keymgr_output_t sw_key;
+  uint32_t sideload_key[kX2551PublicKeySizeBytes];
+} cdi_outputs_t;
+
+// Symbols of the OTBN X22519 public key generation program.
+// See sw/otbn/crypto/x25519_sideload.s for the source code.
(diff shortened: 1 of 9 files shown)


vogelpi (commented):
Thanks @andrea-caforio for your PR. This is nice work!

I do have some questions in addition to the rather minor comments. They are all related to the similarities between this test and with the `chip_sw_keymgr_derive_attestation` test (according to the test plan).
1.  Given that quite some code can be shared, it's probably worth considering if the code can be structured to reduce the changes in the upcoming PR implementing the other test. WDYT?
2.  According to the testplan (see attestation test), we should also check the following:
> The softwre binding registers must be locked after configuration until a keymgr advance operation.
> The device initial state needs to be equivalent to the end state of the `manuf_ft_provision_rma_token_and_personalization` testpoint, otherwise they keymgr
              will fail to advance into operational states.
> Ensure the entropy complex is running in continuous mode, and that KMAC is configured to extract entropy from EDN.

Can you comment on whether this is happening already? I noted that the keymgr testutils indeed set up the entropy complex but I am not sure if KMAC is configured accordingly (probably yes).

andrea-caforio:
Thank you @vogelpi for the review. I split the PR into two commits in order to cover both sealing and attestation.

- ([24bbbe6](https://github.com/lowRISC/opentitan/pull/24823/commits/24bbbe6eb65343e67f5deac81abf8d91f3d38510)) This commit adds the possibility to choose the CDI type in the key manager DIF, which was hardcoded before to sealing. With this both `chip_sw_keymgr_derive_sealing` and `chip_sw_keymgr_derive_attestation` can be execute with the same file `keymgr_derive_cdi_test.c` with a compilation-time flag distinguishing some details.
- ([0e88cc2](https://github.com/lowRISC/opentitan/pull/24823/commits/0e88cc26a6afa2c663848e1ad9a1cbf6f0676851)) The second commit is then the implementation of `keymgr_derive_cdi_test.c` with your comments addressed.

Two points I'm unsure about:

- I agree that unmasking the output is awkward. However your method that checks whether two shares (pre and post reset) cancel out only works if the entropy used in both runs is identical. Is this always the case or did I misunderstand your technique?
-  Considering the equivalence to the `manuf_ft_provision_rma_token_and_personalization` testpoint. It seems that this is covered by the testutils, which correctly initializes the secrets in the OTP but I might be wrong. @timothytrippel can you confirm this? The `chip_sw_keymgr_sideload_kmac` testcase has a similar condition that is not explicitly tested either.

sw/device/tests/keymgr_derive_cdi_test.c
```diff
@@ -0,0 +1,448 @@
...
+ *   keymgr advance operation.
+ *
+ * This check is implemented in the keymgr DIF (see `dif_keymgr_advance_state`).
+ *
+ * - The device initial state needs to be equivalent to the end state of the
+ *   `manuf_ft_provision_rma_token_and_personalization` testpoint, otherwise
+ *   the keymgr will fail to advance into operational states.
+ *
+ * The provisioning of the device secrets in the OTP is handled by the keymgr
+ * testutils (see `keymgr_testutils_init_nvm_then_reset`).
```
timothytrippel, line 347: I would just reword the test point to say:
```suggestion
 * - OTP SECRET2 partition (which contains the CREATOR_ROOT_KEY) and flash info  
 *   pages 1 and 2 in partition 0 must be configured, otherwise the keymgr will fail to
 *   advance into operational states.
 *
 * The provisioning of the device secrets in the OTP SECRET2 partition is handled by the
 * Bazel execution environment, and the flash info secrets are handled by the keymgr
 * testutils (see `keymgr_testutils_init_nvm_then_reset`).
```
I think the test point description was a bit outdated/less concise than it should be, but this is what your requirement is (and you are correct in that the test already ensures this as described).

vogelpi:
> Thank you @vogelpi for the review. I split the PR into two commits in order to cover both sealing and attestation.
> 
>     * ([24bbbe6](https://github.com/lowRISC/opentitan/pull/24823/commits/24bbbe6eb65343e67f5deac81abf8d91f3d38510)) This commit adds the possibility to choose the CDI type in the key manager DIF, which was hardcoded before to sealing. With this both `chip_sw_keymgr_derive_sealing` and `chip_sw_keymgr_derive_attestation` can be execute with the same file `keymgr_derive_cdi_test.c` with a compilation-time flag distinguishing some details.
> 
>     * ([bf00cb9](https://github.com/lowRISC/opentitan/pull/24823/commits/bf00cb9f8ae46c6db3b77b1e35273ae56c46a942)) The second commit is then the implementation of `keymgr_derive_cdi_test.c` with your comments addressed.
> 
> 
> Two points I'm unsure about:
> 
>     * I agree that unmasking the output is awkward. However your method that checks whether two shares (pre and post reset) cancel out only works if the entropy used in both runs is identical. Is this always the case or did I misunderstand your technique?
> 
>     * Considering the equivalence to the `manuf_ft_provision_rma_token_and_personalization` testpoint. It seems that this is covered by the testutils, which correctly initializes the secrets in the OTP but I might be wrong. @timothytrippel can you confirm this? The `chip_sw_keymgr_sideload_kmac` testcase has a similar condition that is not explicitly tested either.

Thanks for making the changes @andrea-caforio ! We've discussed in person how to check the secrets without unmasking them. @timothytrippel gave feedback as well that you've already implemented and @jwnrt provided positive feedback on the compilation offline, too. So this looks all good to me.

Once the secret comparison is modified, we can merge this :-)

vogelpi:
@andrea-caforio , the automatic back porting to the earlgrey_1.0.0 branch failed, see the messages above. Would you mind investigating this and manually back porting this please?

andrea-caforio:
Manual backport PR is here: #25015

PR #24754, opened 2024-10-09
Dice cwt

stevenchtsai:

PR #24824, opened 2024-10-21
[dice,cwt] Add maunal implementations for CDI_0/1 builder

tommychiu-github:
Add some helper functions and a CDI_* builder manually.
Some of the changes in the builder will be replaced by a following RP on opentitantool > codegen feature.

### DIFF
--- a/rules/const.bzl
+++ b/rules/const.bzl
@@ -141,6 +141,7 @@ CONST = struct(
         ESC_PHASE_3 = 0x76,
     ),
     MANUF_STATE = struct(
+        PERSO_INITIAL = 0x00000000,
         SIVAL = 0x30305653,  # ASCII `SV00`.
         PRODC = 0x30304350,  # ASCII `PC00`.
     ),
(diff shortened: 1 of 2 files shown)


sw/device/silicon_creator/lib/base/util.c
```diff
@@ -22,3 +22,16 @@ void util_reverse_bytes(void *buf, size_t num_bytes) {
...
+
+// Returns hexdump character for the half-byte.
+static inline uint8_t hexdump_halfbyte(uint8_t half_byte) {
+  if (half_byte < 10)
+    return '0' + half_byte;
+  else
+    return 'a' + half_byte - 10;
+}
+
+void hexdump_byte(uint8_t byte, uint8_t *str) {
```
timothytrippel, line 34: style guide naming convention is
```suggestion
void util_hexdump_byte(uint8_t byte, uint8_t *str) {
```

timothytrippel (commented):
@tommychiu-github I thought we were going to implement these functions via the automatic code gen in #24754 ? why do we need a manual implementation as well? If you preferred a manual implementation, why did you already implement most of the codegen in #24754?

tommychiu-github:
> @tommychiu-github I thought we were going to implement these functions via the automatic code gen in #24754 ? why do we need a manual implementation as well? If you preferred a manual implementation, why did you already implement most of the codegen in #24754?

The thought is to have a reference implementation & output, for cross checking against the auto-codegen (#24754). 
BTW only the last commit will be replaced by the codegen output, the rest commits in this PR are requisites.

PR #24789, opened 2024-10-15
[manuf] Bind perso firmware with PERSO_INITIAL MANUF_STATE

anthonychen1251:
This PR has two commits that:

1. Separate the provisioning of the MANUF_STATE OTP field from the rest of the CREATOR_SW_CFG partition and provision it at the end of perso firmware.
2. Bind the personalization firmware with the `PERSO_INITIAL` manufacturing state.
   - Define `PERSO_INITIAL` creator manufacturing state.
   - Create a new `manifest_perso` with `Initial` manufacturing state binding.

This PR addresses #24610 partially.
Note: this depends on #24783. Only review the last 2 commits.

Issue #24617, opened 2024-09-23
[manuf] Enhance the re-entrant requirement for the provisioning flow


anthonychen1251:
### Description
There is a requirement for all steps in the provisioning flow: i.e., that each step (CP init, FT individualize, and FT personalize) is re-entrant, and can be re-run if it did not complete successfully.

The goal is to enhance OTP testutil/manuf lib to handle skipping an OTP word that was already programmed;

PR #24826, opened 2024-10-21
[bazel,otp,manuf] Support immutable ROM_EXT enabling during personalization 

anthonychen1251:
This PR has two commits that:

1. Update ROM_EXT immutable section OTP gen tooling
   - Consider `_rom_ext_start_address` when determining the manifest start address to support the common rom_ext binary
   - Updates the OTP  JSON file with data from an immutable ROM_EXT section only if the immutable ROM_EXT is enabled.
2. Delay IMMUTABLE_ROM_EXT_EN provisioning to end of personalization
   - Separate the provisioning of the IMMUTABLE_ROM_EXT_EN OTP field from the rest of the CREATOR_SW_CFG partition
   - Use the ROM_EXT immutable section OTP gen tool by default when building personalization binary

This PR addresses #24610 partially.

Note: this depends on #24783 and #24789. Only review the last 2 commits.

### DIFF
--- a/sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.h
+++ b/sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.h
@@ -17,6 +17,7 @@ extern const size_t kOtpKvCreatorSwCfgSize;
 extern const otp_kv_t kOtpKvCreatorSwCfg[];
 extern const uint32_t kCreatorSwCfgFlashDataDefaultCfgValue;
 extern const uint32_t kCreatorSwCfgManufStateValue;
+extern const uint32_t kCreatorSwCfgImmutableRomExtEnValue;
 
 /**
  * OTP Owner Software Configuration Partition.
@@ -83,6 +84,21 @@ status_t manuf_individualize_device_flash_data_default_cfg(
  */
 OT_WARN_UNUSED_RESULT
 status_t manuf_individualize_device_creator_manuf_state_cfg(
+    const dif_otp_ctrl_t *otp_ctrl);
+
+/**
+ * Configures the IMMUTABLE_ROM_EXT_EN field in the CREATOR_SW_CFG OTP
+ * partition.
+ *
+ * This must be called before `manuf_individualize_device_creator_sw_cfg_lock()`
+ * is called. The operation will fail if there are any pre-programmed words not
+ * equal to the expected test values.
+ *
+ * @param otp_ctrl OTP controller instance.
+ * @return OK_STATUS if the IMMUTABLE_ROM_EXT_EN field was provisioned.
+ */
+OT_WARN_UNUSED_RESULT
+status_t manuf_individualize_device_immutable_rom_ext_en_cfg(
     const dif_otp_ctrl_t *otp_ctrl);
 
 /**
--- a/util/design/gen-otp-immutable-rom-ext-json.py
+++ b/util/design/gen-otp-immutable-rom-ext-json.py
@@ -16,11 +16,16 @@ from elftools.elf import elffile
 _OTP_PARTITION_NAME = "CREATOR_SW_CFG"
 
 _OTTF_START_OFFSET_SYMBOL_NAME = "_ottf_start_address"
+_ROM_EXT_SATRT_OFFSET_SYMBOL_NAME = "_rom_ext_start_address"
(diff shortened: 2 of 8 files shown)


util/design/gen-otp-immutable-rom-ext-json.py
```diff
@@ -107,6 +149,10 @@ def main():
                         type=str,
                         metavar='<path>',
                         help='Output JSON file path.')
+    parser.add_argument('--hardened-true-value',
```
timothytrippel, line 152: I think this parameter is a bit overkill. The truth value for enablement is hardcoded in the ROM already, so I think we can hardcode in this script too, and just drop a comment saying it must the value in the ROM: https://github.com/lowRISC/opentitan/blob/05db1550680f82a5d555e402ad4454edf82a0228/sw/device/silicon_creator/rom/rom.c#L673

util/design/gen-otp-immutable-rom-ext-json.py
```diff
@@ -86,6 +112,22 @@ def update_json_with_immutable_rom_ext_section_data(self):
...
+        partition and compares it with the provided hardened true value.
+
+        Args:
+            hardened_true_value: The expected value for the enable field
+                                when immutable ROM extension is enabled.
+
+        Returns:
+            True if immutable ROM extension is enabled, False otherwise.
+        """
+        immutable_rom_ext_en_value = int(self.get_key_value(_ENABLE_FIELD_NAME), 0)
```
timothytrippel, line 128:
```suggestion
        immutable_rom_ext_en_value = int(self.json_data["partitions"][_OTP_PARTITION_NAME].get(_ENABLE_FIELD_NAME, 0))
```
should also work I believe.

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.h
```diff
@@ -85,6 +86,21 @@ OT_WARN_UNUSED_RESULT
...
+ *
+ * This must be called before `manuf_individualize_device_creator_sw_cfg_lock()`
+ * is called. The operation will fail if there are any pre-programmed words not
+ * equal to the expected test values.
+ *
+ * @param otp_ctrl OTP controller instance.
+ * @return OK_STATUS if the IMMUTABLE_ROM_EXT_EN field was provisioned.
+ */
+OT_WARN_UNUSED_RESULT
+status_t manuf_individualize_device_immutable_rom_ext_en_cfg(
```
timothytrippel, line 101: Just a thought here, since we now have a separate function for setting the:
- flash_data_default_cfg
- manuf_state
- immutable_rom_ext_en
- rom_bootstrap_dis
perhaps we could combine these into a one function that take as inputs:
- the OTP partition
- the field / offset to program
?

This would consolidate a lot of very similar functionality. Ok to do in a follow up PR, or additional commit on this PR if you would like.

timothytrippel (approved):
few comments, otherwise LGTM.

sw/device/silicon_creator/manuf/lib/individualize_sw_cfg.h
anthonychen1251: Thanks for the suggestion. That's a good improvement. If you don't mind,  I could include this in the PR for #24617

util/design/gen-otp-immutable-rom-ext-json.py
anthonychen1251: okay, done.

util/design/gen-otp-immutable-rom-ext-json.py
anthonychen1251: `self.json_data['partitions']` is a list so we can not directly access to a specific partition using its name as an index.

This is what `self.json_data['partitions']` looks like:
`[OrderedDict([('items', [OrderedDict([('name', 'CREATOR_SW_CFG_IMMUTABLE_ROM_EXT_EN'), ('value', '0x000001d4')])]), ('name', 'CREATOR_SW_CFG')])]`

timothytrippel: ah I see, never mind then.

anthonychen1251:
The CI pipeline was failing (task:`CI (Checkout repository) `) for this PR.  I've pushed a small update with reworded commit messages to see if that resolves the issue.

PR #24830, opened 2024-10-21
[rom_ctrl, dv] Fixed tlul_adapter_reg uncovered conditional coverage

KinzaQamar:
The nightly coverage dashboards were reporting two uncovered conditional coverage in tlul_adapter_reg.sv at L208. There is a conditional assignment on L208 for err_internal, which is getting assigned if any of the bits (addr_align_err, malformed_meta_err, tl_err, instr_error and intg_error) is set. The case 0 | 0 | 0 | 1 | 0 was uncovered as the assignment for tl_err, instr_error and malformed_meta_err are depending over common factors.
It turns out that instr_error could exist on its own if we have an opcode for A channel as GET, instr_type as MuBiTrue and fetch for register is disable.
This fix added a mechanism inside cip_base_vseq__tl_errors.svh to mimic the above behaviour and then let the scoreboard knows that it should expect an error and raise an invalid access.


hw/dv/sv/cip_lib/cip_base_scoreboard.sv
```diff
@@ -386,6 +386,16 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
     return 0;
   endfunction
 
+  //check if the tl_mapped address is a csr
```
rswarbrick, line 389: Really nitty, but I think the usual formatting is "// blah blah" (with a space after the forward slashes)

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
```diff
@@ -386,6 +386,16 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
...
+  virtual function bit is_tl_csr_read(tl_seq_item item, string ral_name);
+    cip_tl_seq_item cip_item;
+    `downcast(cip_item, item)
+    if (item.a_opcode == tlul_pkg::Get && cip_item.get_instr_type() == MuBi4True) begin
+      if ((is_tl_access_mapped_addr(item, ral_name) && !is_mem_addr(item, ral_name))) return 1;
+      else return 0;
```
rswarbrick, line 395: I don't think you need the if/else here: can't you just return the thing you're testing?

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__tl_errors.svh
```diff
@@ -232,13 +232,23 @@ task tl_write_ro_mem_err(string            ral_name,
   end
 endtask
 
+// Return the address of the csr at a random address for read
```
rswarbrick, line 235: I think you probably want "... the address of a random CSR ..." ? (Because a random address might not have a CSR!)

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__tl_errors.svh
```diff
@@ -232,13 +232,23 @@ task tl_write_ro_mem_err(string            ral_name,
...
+virtual function bit[BUS_AW-1:0] csr_read_addr (string ral_name, dv_base_reg_block block);
```
rswarbrick, line 236: I think I'd probably include a verb in the name here. Maybe "pick_random_csr_addr"? Or (if we don't fancy a verb) just "random_csr_addr"? I'm not sure I'd include the fact that we're planning to read it as part of the name :-)

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__tl_errors.svh
```diff
@@ -232,13 +232,23 @@ task tl_write_ro_mem_err(string            ral_name,
...
+  uvm_reg regs[$];
+  int index = $urandom_range(0, block.csr_addrs.size());
```
rswarbrick, line 238: Off by one error here, I think. The arguments are the max and min values, so the max value is too big by one.

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__tl_errors.svh
```diff
@@ -232,13 +232,23 @@ task tl_write_ro_mem_err(string            ral_name,
...
+  block.get_registers(regs);
```
rswarbrick, line 239: Is this actually needed? We're not using `regs`, so is it needed for some side effect? (I don't *think* so, but think I should check!)

rswarbrick (commented):
Some nitty comments, but this looks really good!

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__tl_errors.svh
KinzaQamar: Sorry, I wanted to write random "index". Does that look fine now?

rswarbrick (commented):
Other than the slight rejig that I'm suggestion for `is_tl_csr_read`, this looks really good to me.

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
```diff
@@ -386,6 +386,16 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
     return 0;
   endfunction
 
+  // Check if the tl_mapped address is a csr
+  virtual function bit is_tl_csr_read(tl_seq_item item, string ral_name);
```
rswarbrick, line 390: Oh nuts: I'm really sorry for not noticing this earlier. I think this function only returns true if `get_instr_type()` is `MuBi4True`. Shouldn't we call it something slightly more specific like `is_csr_fetch`?

rswarbrick (approved):
This looks really good to me: thanks!

PR #24833, opened 2024-10-22
[rom_ext] Correct the certificates written to flash

cfrantz:
1. Round the data lengths written to flash up to the next word.  This error was causing the certificates to fail their signature tests.
2. Eliminate the pathlen constraint from the fake CA certificate and allow certificate chains of arbitrary lenth.  This allows us to verify the certificate chain from CA->UDS->CDI0->CDI1.


sw/device/silicon_creator/manuf/keys/fake/fake_ca.conf
```diff
@@ -15,5 +15,5 @@ CN=Google Engineering ICA
 [v3_ca]
 subjectKeyIdentifier = hash
 authorityKeyIdentifier = keyid:always,issuer:always
-basicConstraints = critical,CA:true,pathlen:0
+basicConstraints = critical,CA:true,pathlen:3
```
timothytrippel, line 18: pathlen of 4 to enable CDI_1 to endorse another cert in test flows?

timothytrippel (approved):
nice find!

sw/device/silicon_creator/manuf/keys/fake/fake_ca.conf
cfrantz: https://stackoverflow.com/questions/6616470/certificates-basic-constraints-path-length

Perhaps we should just omit the pathlen constraint from the fake CA.

cfrantz: I eliminated the `pathlen` constraint.  We should still seek guidance on this when generating the real endorsement certificate to be used in production.

cfrantz:
The CI test failure is not related to this PR.

PR #24839, opened 2024-10-22
[hmac,sw] Workaround after hang

martin-velay:
- Implement a workaround in C cryptolib as HW is hanged after a stop in certain conditions.
- Refer to issue #24767

### DIFF
--- a/sw/device/lib/crypto/drivers/hmac.c
+++ b/sw/device/lib/crypto/drivers/hmac.c
@@ -67,39 +67,16 @@ OT_ASSERT_ENUM_VALUE(HMAC_DIGEST_15_REG_OFFSET, HMAC_DIGEST_14_REG_OFFSET + 4);
 enum {
   /* The beginning of the address space of HMAC. */
   kHmacBaseAddr = TOP_EARLGREY_HMAC_BASE_ADDR,
+  // TODO: should be removed when issue #24767 will be solved in the HW.
+  // Max number of iterations before the software decides that the hardware
+  // needs to be recovered after a stop command.
+  // At max 4 clock cycles are required to perform a read access to the
+  // register. As it should take less than 64 clock cycles in SHA2-256 and
+  // 80 clock cycles in SHA2-384/512, let's take some margin and consider
+  // that 100 loops are a way enough to see IDLE status. Otherwise we can
+  // start to attempt to recover the HW.
+  kNumIterTimeout = 100,
 };
-
-/**
- * Wait until HMAC becomes idle.
- *
- * It returns error if HMAC HWIP becomes idle without firing `hmac_done`
- * interrupt.
- *
- * TODO(#23191): It might be beneficial to have a timeout value for the polling.
- *
- * @return Result of the operation.
- */
-OT_WARN_UNUSED_RESULT
-static status_t hmac_idle_wait(void) {
-  // Verify that HMAC HWIP is idle.
-  // Initialize `status_reg = 0` so that the loop starts with the assumption
-  // that HMAC HWIP is not idle.
-  uint32_t status_reg = 0;
-  while (bitfield_bit32_read(status_reg, HMAC_STATUS_HMAC_IDLE_BIT) == 0) {
-    status_reg = abs_mmio_read32(kHmacBaseAddr + HMAC_STATUS_REG_OFFSET);
-  }
-
-  // Verify that HMAC HWIP raises `hmac_done` bit.
-  uint32_t intr_reg =
-      abs_mmio_read32(kHmacBaseAddr + HMAC_INTR_STATE_REG_OFFSET);
-  if (bitfield_bit32_read(intr_reg, HMAC_INTR_STATE_HMAC_DONE_BIT) == 0) {
-    return OTCRYPTO_FATAL_ERR;
-  }
-
-  // Clear the interrupt by writing 1, because `INTR_STATE` is rw1c type.
-  abs_mmio_write32(kHmacBaseAddr + HMAC_INTR_STATE_REG_OFFSET, intr_reg);
-  return OTCRYPTO_OK;
-}
 
 /**
  * Clear the state of HMAC HWIP so that further driver calls can use it.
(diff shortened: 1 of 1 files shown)


sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -77,28 +78,42 @@ enum {
...
+  //  | that 50 loops are a way enough to see IDLE status. Otherwise we can
+  //  | start to attempt to recover the HW.
   uint32_t status_reg = 0;
+  uint32_t attempt_cnt = 0;
   while (bitfield_bit32_read(status_reg, HMAC_STATUS_HMAC_IDLE_BIT) == 0) {
     status_reg = abs_mmio_read32(kHmacBaseAddr + HMAC_STATUS_REG_OFFSET);
+    attempt_cnt++;
+    if (attempt_cnt == 50) {
+      recover_hw_after_stop(ctx);
+      hw_recovered = true;
```
jadephilipoom, line 103: To write under the pointer, you might need:
```suggestion
      *hw_recovered = true;
```
Might be what's causing the compilation issue?

jadephilipoom:
The general direction of this looks reasonable to me. I'd maybe increase the number of attempts before timeout just in case, no need to have too narrow of a margin there.

martin-velay:
> The general direction of this looks reasonable to me. I'd maybe increase the number of attempts before timeout just in case, no need to have too narrow of a margin there.

Thanks for your inputs. I have changed it to 100, I don't really know what can be acceptable in SW.

martin-velay:
This last version is working:
image

martin-velay:
This change is working on the FPGA platform but in simulation as it causes some assertions to fail as we are doing something which is supposed to be done:
```
xmsim: *E,ASRTST (/home/dev/src/scratch/hmac_hang_c_lib/chip_earlgrey_asic-sim-xcelium/default/src/lowrisc_ip_hmac_0.1/rtl/hmac.sv,936): (time 4155662713 PS) Assertion tb.dut.top_earlgrey.u_hmac.ValidHmacEnConditionAssert has failed 
```
```
  // hmac_en should be modified only when the logic is Idle
  `ASSERT(ValidHmacEnConditionAssert,
          hmac_en != $past(hmac_en) |-> !in_process && !initiated)
```
And also:
```
xmsim: *E,ASRTST (/home/dev/src/scratch/hmac_hang_c_lib/chip_earlgrey_asic-sim-xcelium/default/src/lowrisc_ip_hmac_0.1/rtl/hmac.sv,927): (time 4155962713 PS) Assertion tb.dut.top_earlgrey.u_hmac.ValidWriteAssert has failed 
```
```
  // the host doesn't write data after hash_process until hash_start_or_continue.
  `ASSERT(ValidWriteAssert, msg_fifo_req |-> !in_process)
```
@vogelpi / @andreaskurth how can I do?

vogelpi (commented):
Thanks @martin-velay for your work on this!

I've left some comments but ultimately, @jadephilipoom should have the final say on this.

As for the assertions: we need to tune these to allow for the workaround in this particular case. I think tuning them should be pretty easy. As it's SVAs inside the top-level HMAC file hmac.sv, you can do hierarchical references into the lower modules to check whether we  are in the messed up state. Also a reference to the issue should be added to give background and to remember undoing the tuning once the RTL bug gets fixed.

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -397,17 +480,21 @@ status_t hmac_update(hmac_ctx_t *ctx, const uint8_t *data, size_t len) {
...
-  // Wait for HMAC HWIP operation to be completed.
-  HARDENED_TRY(hmac_idle_wait());
-
-  // Store context into `ctx`.
-  context_save(ctx);
+  // Wait for HMAC HWIP operation to be completed and store context into `ctx`
+  // only if not already done in case of hanged HW.
+  if (hmac_idle_wait(ctx).value != OTCRYPTO_RECOV_ERR.value) {
+    context_save(ctx);
+  }
```
jadephilipoom, line 497: This exact comparison will fail to trigger if `CRYPTO_STATUS_DEBUG` is set on the command line, since then every instance of `OTCRYPTO_RECOV_ERR` includes the line number on which the macro was called. It would be more robust to use `status_extract` and compare the resulting `code`:
https://github.com/lowRISC/opentitan/blob/da20a5887b2a48508cfc6334e8fa7377b40d7083/sw/device/lib/base/status.h#L221
It's just a tool for test debugging, so maybe not the most critical, but could make debugging harder. I'd be okay with not blocking the PR and creating an issue to fix this. There's more background on how cryptolib handles status codes here: https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/crypto/impl/status.h

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -256,6 +225,120 @@ static void msg_fifo_write(const uint8_t *message, size_t message_len) {
...
+  //  | 80 clock cycles in SHA2-384/512, let's take some margin and consider
+  //  | that 100 loops are a way enough to see IDLE status. Otherwise we can
+  //  | start to attempt to recover the HW.
+  uint32_t status_reg = 0;
+  uint32_t attempt_cnt = 0;
+  while (bitfield_bit32_read(status_reg, HMAC_STATUS_HMAC_IDLE_BIT) == 0) {
+    status_reg = abs_mmio_read32(kHmacBaseAddr + HMAC_STATUS_REG_OFFSET);
+    attempt_cnt++;
+    if (attempt_cnt == 100) {
+      recover_hw_after_stop(ctx);
```
jadephilipoom, line 321: I think you could simply return `OTCRYPTO_RECOV_ERR` here and exit, and then look for that case in the function that calls this one and run the `recover_hw_after_stop()` operation there. Then you wouldn't need to pass `ctx` to this function at all.

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -397,17 +480,21 @@ status_t hmac_update(hmac_ctx_t *ctx, const uint8_t *data, size_t len) {
   // Keep writing incoming bytes
   msg_fifo_write(data, len - leftover_len);
 
+  // delay here, even  this small is enough
+  for (size_t i = 0; i < 40; i = launder32(i + 1))
+    ;
```
jadephilipoom, line 485: We probably shouldn't merge this code; my understanding is that it's just here to force the bug to trigger so that we can see if the change works, but it shouldn't be in the final PR, right?

sw/device/lib/crypto/drivers/BUILD
```diff
@@ -183,6 +183,7 @@ cc_library(
         "//sw/device/lib/base:macros",
         "//sw/device/lib/base:memory",
         "//sw/device/lib/crypto/impl:status",
+        "//sw/device/lib/runtime:log",
```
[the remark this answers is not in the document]
jadephilipoom: This and the later instance of `log` imports are still present and I agree with Pirmin that they shouldn't be -- why was this comment resolved?

sw/device/lib/crypto/drivers/hmac.c
martin-velay: If I do so, the function will stop here and won't execute this remaining piece of code?
```
  // Verify that HMAC HWIP raises `hmac_done` bit.
  uint32_t intr_reg =
      abs_mmio_read32(kHmacBaseAddr + HMAC_INTR_STATE_REG_OFFSET);
  if (bitfield_bit32_read(intr_reg, HMAC_INTR_STATE_HMAC_DONE_BIT) == 0) {
    status = OTCRYPTO_FATAL_ERR;
  }

  // Clear the interrupt by writing 1, because `INTR_STATE` is rw1c type.
  abs_mmio_write32(kHmacBaseAddr + HMAC_INTR_STATE_REG_OFFSET, intr_reg);
  if (status.value != OTCRYPTO_RECOV_ERR.value) {
    status = OTCRYPTO_OK;
  }

  return status;
```
I think we should execute those read and write operations after the HW has been recovered or not.

martin-velay:
https://github.com/lowRISC/opentitan/pull/24839#discussion_r1815044118
@jadephilipoom, because I was doing it in parallel with you review, I was about to push my changes.

martin-velay:
This workaround is easier https://github.com/lowRISC/opentitan/pull/24944 and do not require to change the RTL assertions.

PR #24842, opened 2024-10-22
[topgen,template] Don't hardcode Earlgrey in toplevel_pkg

Razer6:
There was a hard coded reference to Earlgrey in the `toplevel_pkg` template. Remove that and fetch that information from the provided parameters.


rswarbrick (approved):
Looks sensible to me!

rswarbrick:
I don't believe that the failing FPGA tests can be caused by this commit (since it doesn't actually affect the generated RTL). Merging.

PR #24844, opened 2024-10-22
[rom_ext] Reject ownership with invalid flash configs

cfrantz:
1. If an ownership configuration contains a flash config that sets the ROM_EXT region ECC/Scrambling properties to mismatch with the default flash config, reject the configuration.
2. Add tests.


moidx (approved):

cfrantz:
The CI failure is not related to this PR.

PR #24853, opened 2024-10-23
Backport 24794 to earlgrey_es_sival

nmoroze:
The latest rev of the HyperTeacup board (v1.5.0) connects the active-low shutdown pin of the LED driver to HyperDebug pin CN10.15. The board pulls this pin low by default, so the HyperDebug needs to actively drive it high to enable the LED driver.

This commit updates opentitanlib's HyperTeacup configuration to drive this pin high by default, ensuring that the driver is always active.

CN10.15 is not connected on HyperTeacup rev 1.3.0, so this change should be backwards compatible with older boards.

Signed-off-by: Noah Moroze <noah@opentitan.org>
(cherry picked from commit cb5bf502a94ef3a7969aa6b43dfbfee92d377854)

### DIFF
--- a/sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
+++ b/sw/host/opentitanlib/src/app/config/hyperdebug_teacup.json
@@ -321,6 +321,16 @@
       "name": "VBUS_SENSE",
       "mode": "Input",
       "alias_of": "CN10_18"
+    },
+    {
+        // Connected to active-low shutdown pin on the LED driver. This is
+        // pulled low by the rev 1.5 Teacup board, so it must be driven high to
+        // activate the driver. This configuration makes a high output the
+        // default.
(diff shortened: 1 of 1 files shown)


nmoroze:
@jesultra I forgot to close the loop on backporting #24794 a while ago, but I figure we might as well get this in.

I think the CI failures are likely spurious, it seems like these same jobs are consistently failing on earlgrey_es_sival. The one exception is I don't see other recent failures of the SiVal ROM_EXT test job, but the failing test seems unrelated to my changes (//sw/device/tests:entropy_src_fw_override_test_fpga_cw310_sival_rom_ext).

jesultra (approved):
No objection to backporting from me.

nmoroze:
Thanks, would you mind merging? I don't have permission.

PR #24855, opened 2024-10-23
Have manifest rule convert timestamp hex int to array

zhangarino:
manifest json definition in opentitanlib's manifest_def.rs expects two uint32_t's for the timestamp.


zhangarino:
Sorry, there was an indenting issue

cfrantz (approved):

PR #24870, opened 2024-10-24
[ci] move Docker build test job to GitHub actions

nbdd0121:


jwnrt (approved):
Looks good

I know you copied the azure job, but why is one docker build `continue-on-error` and the other not?

jwnrt:
We do see failures occasionally and I don't want to block PRs on that. Let's leave it for now.

https://dashboards.lowrisc.org/grafana/d/ebe2c4c0-bb38-47ab-aac0-c04c9956898d/ci-dashboard?orgId=1&from=now-1y&to=now&viewPanel=298

Issue #24666, opened 2024-09-25
[rom_ext] Integration Specific Firmware Binding


cfrantz:
Supports ProdC use case
WIP Spec: https://docs.google.com/document/d/19oQtqgDdFU6qunbwgkwNaizYUNCzdLVqmnBpLaDvX5I/edit

jettr:
Looks like this feature might be close to done? Do we need to land it on the 1.0.0 branch though?

PR #24882, opened 2024-10-25
[opentitantool] Add ISFB manifest extensions.

moidx:
Add opentitantool image manifest support for Integrator Specfic Firmware Binding (ISFB) extensions.

The following manifest extensions were added:

1. isfb: Integration specific firmware binding manifest extension. Contains anti-rollback strike mask as well as device expecific product expression.
2. isfb_erase_policy: This extension is used to authorize erasing of the ISFB region. Intended for testing purposes. The implementation will require node-locking for this policy to be effective.

This change does not integrate the extensions into the build system. This will be done in a separate commit.

This is part of: https://github.com/lowRISC/opentitan/issues/24666

### DIFF
--- a/sw/host/opentitanlib/src/image/manifest.rs
+++ b/sw/host/opentitanlib/src/image/manifest.rs
@@ -12,6 +12,11 @@
 #![deny(unsafe_code)]
 
 use crate::with_unknown;
+use anyhow::Result;
+use byteorder::{LittleEndian, WriteBytesExt};
+use serde::{Deserialize, Serialize};
+use serde_annotate::Annotate;
+use std::io::Write;
 use zerocopy::{AsBytes, FromBytes, FromZeroes};
 
 // Currently, these definitions must be updated manually but they can be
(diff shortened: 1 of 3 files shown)


sw/host/opentitanlib/src/image/manifest.rs
```diff
@@ -154,6 +171,50 @@ pub struct ManifestExtSecVerWrite {
...
+pub struct ManifestExtIsfbProductExpr {
+    pub mask: u32,
+    pub value: u32,
+}
+
+/// Integrator Specific Firmware Binding manifest extension.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ManifestExtIsfb {
+    pub header: ManifestExtHeader,
+    pub strike_mask: [u32; 4],
```
cfrantz, line 185: Do we want to use `u128` here?

sw/host/opentitanlib/src/image/manifest.rs
```diff
@@ -104,12 +113,20 @@ impl Default for SigverifySpxSignature {
...
 }
 
+impl ManifestExtHeader {
+    pub fn write(&self, dest: &mut impl Write) -> Result<()> {
+        dest.write_u32::<LittleEndian>(self.identifier)?;
+        dest.write_u32::<LittleEndian>(self.name)?;
+        Ok(())
+    }
+}
+
```
cfrantz, line 129: Since the `ManifestExtHeader` implements `AsBytes`, this write function isn't really necessary (except that it provides a nice regular interface for use in `ManifestExtIsfb::write`.

I'd change this function to use the `AsBytes` implementation so as to not have two binary conversion functions:
```
pub fn write(&self, dest: &mut impl Write) -> Result<()> {
  dest.write(self.as_bytes())?;
  Ok(())
}
```

cfrantz (approved):

sw/host/opentitanlib/src/image/manifest.rs
moidx: Done. Thanks, this helped simplify the code.

sw/host/opentitanlib/src/image/manifest.rs
moidx: Done. I ended up having to use `dest.write_all(self.as_bytes())?;`. 

Tested changes.

PR #24884, opened 2024-10-25
[manuf] load the assembled image during perso

anthonychen1251:
This enables loading a combined image containing the personalization firmware in slot a and ROM_EXT in slot b during personalization flow. This also enhances the opentitan bootstrap test utils to support loading different binaries. 

This PR addresses https://github.com/lowRISC/opentitan/issues/24610 partially.

### DIFF
--- a/sw/device/silicon_creator/manuf/base/BUILD
+++ b/sw/device/silicon_creator/manuf/base/BUILD
@@ -26,6 +26,10 @@ load(
     "//sw/device/silicon_creator/rom_ext:defs.bzl",
     "ROM_EXT_VERSION",
 )
+load(
+    "//sw/device/silicon_creator/rom/e2e:defs.bzl",
+    "SLOTS",
+)
 
 package(default_visibility = ["//visibility:public"])
 
@@ -349,6 +353,7 @@ bool_flag(
 _FT_PROVISIONING_CMD_ARGS = """
   --elf={sram_ft_individualize}
   --bootstrap={ft_personalize}
+  --second-bootstrap={firmware}
 """ + FT_PROVISIONING_INPUTS + select({
     ":ckms_cert_endorsement_params": CLOUD_KMS_CERT_ENDORSEMENT_PARAMS,
     "//conditions:default": LOCAL_CERT_ENDORSEMENT_PARAMS,
@@ -366,15 +371,19 @@ _FT_PROVISIONING_HARNESS = "//sw/host/provisioning/ft"
         },
         fpga = fpga_params(
             timeout = "long",
+            assemble = "{ft_personalize}@{slot_a} {rom_ext}@{slot_b}",
             binaries =
                 {
                     ":sram_ft_individualize_{}".format(config["otp"]): "sram_ft_individualize",
                     ":ft_personalize_{}".format(sku): "ft_personalize",
+                    "//sw/device/silicon_creator/rom_ext/sival:rom_ext_fake_prod_signed_slot_b": "rom_ext",
                 },
             changes_otp = True,
             data = FT_PERSONALIZE_ENDORSEMENT_KEYS,
             needs_jtag = True,
             otp = "//hw/ip/otp_ctrl/data/earlgrey_skus/sival:otp_img_test_locked0_manuf_initialized",
+            slot_a = SLOTS["a"],
+            slot_b = SLOTS["b"],
             tags = [
                 "lc_test_locked0",
                 "manuf",
@@ -383,15 +392,19 @@ _FT_PROVISIONING_HARNESS = "//sw/host/provisioning/ft"
             test_harness = _FT_PROVISIONING_HARNESS,
         ),
         silicon = silicon_params(
+            assemble = "{ft_personalize}@{slot_a} {rom_ext}@{slot_b}",
             binaries =
                 {
                     ":sram_ft_individualize_{}".format(config["otp"]): "sram_ft_individualize",
                     ":ft_personalize_{}".format(sku): "ft_personalize",
+                    "//sw/device/silicon_creator/rom_ext/sival:rom_ext_fake_prod_signed_slot_b": "rom_ext",
                 },
             changes_otp = True,
             data = FT_PERSONALIZE_ENDORSEMENT_KEYS,
             interface = "teacup",
             needs_jtag = True,
+            slot_a = SLOTS["a"],
+            slot_b = SLOTS["b"],
             test_cmd = _FT_PROVISIONING_CMD_ARGS,
             test_harness = _FT_PROVISIONING_HARNESS,
         ),
--- a/sw/device/silicon_creator/rom_ext/sival/BUILD
+++ b/sw/device/silicon_creator/rom_ext/sival/BUILD
@@ -42,12 +42,14 @@ opentitan_binary(
     exec_env = [
         "//hw/top_earlgrey:silicon_creator",
         "//hw/top_earlgrey:fpga_cw310",
+        "//hw/top_earlgrey:fpga_cw340",
         "//hw/top_earlgrey:sim_dv_base",
         "//hw/top_earlgrey:sim_verilator_base",
     ],
     linker_script = "//sw/device/silicon_creator/rom_ext:ld_slot_a",
     linkopts = LINK_ORDER,
     manifest = ":manifest_sival",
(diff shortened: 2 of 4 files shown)


sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -349,6 +353,7 @@ bool_flag(
 _FT_PROVISIONING_CMD_ARGS = """
   --elf={sram_ft_individualize}
   --bootstrap={ft_personalize}
+  --bootstrap={firmware}
```
timothytrippel, line 356: we only want to bootstrap one image, not two, don't we? the combined image should be:
- Slot A: Perso fw
- Slot B: ROM_EXT + Owner FW (or just ROM_EXT for now).

sw/host/opentitanlib/src/test_utils/bootstrap.rs
```diff
@@ -16,15 +16,26 @@ pub struct Bootstrap {
...
+    pub bootstrap: Option<Vec<PathBuf>>,
 }
 
 impl Bootstrap {
     pub fn init(&self, transport: &TransportWrapper) -> Result<Option<Box<dyn Annotate>>> {
+        self.init_specific(transport, 0)?;
+        Ok(None)
+    }
+
+    pub fn init_specific(
```
timothytrippel, line 30: Why do you need to add a bootstrap function to be able to select from a list of items to bootstrap? We just want to bootstrap one image (as described above).

sw/device/silicon_creator/manuf/base/BUILD
anthonychen1251: When we do the 1st bootstrap, we just bootstrap perso fw; then when we do the 2nd bootstrap (after provisioning SECRET1 OTP partition) we bootstrap perso fw + ROM_EXT/Owner FW bundle; to save time loading big chunks of firmware into flash. That's why I sent a list of bootstrap items (perso fw only and the combined image) here,

sw/host/opentitanlib/src/test_utils/bootstrap.rs
anthonychen1251: It's because we only want to bootstrap perso fw when we do the first bootstrap to save time loading big chunks of firmware (combined image) into flash.

sw/device/silicon_creator/manuf/base/BUILD
tommychiu-github: Wondering if we didn't bootstrap Owner FW at this point, how could we push Owner FW after that?
Maybe we should confirm if the SPI rescue works on this version of ROM_EXT.
Could you help confirm that internally @anthonychen1251?

timothytrippel: Ah right, I forgot we settled on this approach @anthonychen1251.

@tommychiu-github : there are two bootstrap operations that happen during perso because the flash scrambling keys are [provisioned into OTP SECRET1 partition](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/base/ft_personalize.c;l=225?q=ft_personalize.c&ss=opentitan%2Fopentitan), and don't take effect until after a reset, at which point the contents of the flash will be garbage again if scrambling is turned on, so we bootstrap again to make it the rest of the way through perso. @anthonychen1251 approach is designed to be more efficient, since the first time around we know we need to bootstrap again, there is no need to bootstrap code that will never get used initially.

sw/host/opentitanlib/src/test_utils/bootstrap.rs
timothytrippel: Instead of creating a separate bootstrap function, can yo just add another command line arg like was done in the perso flow for the Z1 chip: https://github.com/lowRISC/opentitan/blob/3c72c0ed03608a03d1d0350c8e759a81ad55dfd7/sw/host/provisioning/ft/src/main.rs#L95 ?

Then you can just call the existing bootstrap function with this arg like so: https://github.com/lowRISC/opentitan/blob/3c72c0ed03608a03d1d0350c8e759a81ad55dfd7/sw/host/provisioning/ft_lib/src/lib.rs#L291

The earlgrey Z1 chip had 4 perso binaries due to size constraints set by the previous ROM. But those don't exist on the A1 chip so we just had one binary. But you can use the same approach we used before.

anthonychen1251: okay, I follow the same approach to add another command arg to get the path of the assemble image.

sw/device/silicon_creator/manuf/base/BUILD
anthonychen1251: @timothytrippel Thanks for the detailed explanation! I think @tommychiu-github was also wondering about the process for loading Owner FW onto flash if it's not done during perso. Given that the bootstrap isn't disabled in the bring-up SKU, flashing the Owner Software via bootstrap seems feasible. I'm not sure if the SPI interface support for the rescue protocol is ready though.

timothytrippel: Ah I see. If only Perso code is loaded during perso, and bootstrap was not disabled, then after perso, one can simply use the `opentitantool bootstrap ...` command to load a ROM_EXT + owner FW image.

SPI rescue is not supported yet.

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -366,15 +371,19 @@ _FT_PROVISIONING_HARNESS = "//sw/host/provisioning/ft"
         },
         fpga = fpga_params(
             timeout = "long",
+            assemble = "{ft_personalize}@{slot_a} {rom_ext}@{slot_b}",
```
vbendeb, line 374: while we are at it, why not include the RW as well? Especially since SPI rescue is not yet functional, how do we get to use the chip after personalization if the only valid image on it is ROM_EXT_B?

timothytrippel: I would prefer we made this a SKU configuration, i.e., add a field `transport_image` to [this dictionary](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/base/provisioning_inputs.bzl;drc=118a7713bbc060e58832437f071657abfe203f6f;l=28) that is the firmware image that will get placed in slot B. that way the Sival SKU can only place the ROM_EXT, and various other product SKUs can place a combo of ROM_EXT + their Owner firmware image. Does that make sense?

timothytrippel (approved):
LGTM, thanks @anthonychen1251 !

sw/device/silicon_creator/manuf/base/BUILD
timothytrippel: I think we can do this cleanup in a follow up PR.

timothytrippel: #24971 will fix this.

vbendeb: > I would prefer we made this a SKU configuration, 

sure, as long as there is a way to do this for the Cros case.

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -366,15 +371,19 @@ _FT_PROVISIONING_HARNESS = "//sw/host/provisioning/ft"
...
             binaries =
                 {
                     ":sram_ft_individualize_{}".format(config["otp"]): "sram_ft_individualize",
                     ":ft_personalize_{}".format(sku): "ft_personalize",
+                    "//sw/device/silicon_creator/rom_ext/sival:rom_ext_fake_prod_signed_slot_b": "rom_ext",
```
timothytrippel, line 379: I think this actually broke the `//sw/device/silicon_creator/manuf/base:ft_provision_prodc_fpga_hyper310_rom_with_fake_keys` target, because this hardcodes the `sival` ROM_EXT. But no worries, #24971 will fix it again. 

It also looks like the `ft_provision_*` flows are not being run in CI b/c they are marked as `long` tests... I will fix that in a follow up too.

timothytrippel: #24971 has been merged now so everything should be fixed.

timothytrippel: #24975 will ensure the FT flow runs in CI now too.

PR #24885, opened 2024-10-25
[gpiodpi] fix activation and the effective clock generation

tarikgraba:
Assignments on variables (reg/logic) are just initialization. Assignments wires are continuous assignment (assign).

### DIFF
--- a/hw/dv/dpi/gpiodpi/gpiodpi.sv
+++ b/hw/dv/dpi/gpiodpi/gpiodpi.sv
@@ -53,7 +53,8 @@ module gpiodpi
      gpiodpi_close(ctx);
    end
 
-   logic eff_clk = clk_i && active;
+   logic eff_clk;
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
Good point and thanks for the fix! Would you mind re-doing it with logic and splitting out a separate `assign` statement? The result is more obvious to someone who doesn't want to think about the difference between nets and variables :-)

rswarbrick:
Also, you'll need two tweak some stuff about the git commit message, which is currently causing quick lint to fail in CI.

tarikgraba:
Should be ok now,  forgotten about the sign-off thing.

rswarbrick (approved):
This is definitely correct. Sorry that we (I) have taken such a ridiculously long time to reply.

rswarbrick:
Hmm, something has come unstuck since this run through CI last year. I'm going to rebase and force-push.

PR #24896, opened 2024-10-25
[signing] Publish signing ceremony 2024-10-25

cfrantz:
1. Ceremony documentation.
2. Change owner_key_type to ecdsa.
3. ROM_EXT 0.6 binaries.

### DIFF
--- a/signing/logs/2024-10-25.md
+++ b/signing/logs/2024-10-25.md
@@ -1,0 +1,295 @@
+# Signing Ceremony 2024-10-25
+
+- Purpose:
+    - Create new a ROM\_EXT release with ownership transfer and a validatable DICE chain.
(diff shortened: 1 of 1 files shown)


timothytrippel (approved):

jettr:
There seems to be an issue with reporting the primary_slot in the boot log with this version. When I am booting as the NextBoot, the primary_slot field is incorrectly reporting the active slot (instead of the other slot).

cfrantz:
> There seems to be an issue with reporting the primary_slot in the boot log with this version. When I am booting as the NextBoot, the primary_slot field is incorrectly reporting the active slot (instead of the other slot).

This error will be corrected by #24897.

cfrantz:
As with release 0.5, I'm going to abort the release and re-sign binaries with the error corrected.

PR #24900, opened 2024-10-26
[manuf,otp] provision secure boot keys during individualization

### LINKED ISSUE #21554: [manuf] provision ROM secure boot keys and key usage policies
Recently, #21270 added OTP partitions for ROM secure boot keys and key policies.
These must be provisioned during either CP or FT.
Provisioning at FT, during individualization, seems logical since this will be when a chip is categorized to a SKU.

timothytrippel:
This updates the FT individualization provisioning stage to provision the first stage secure boot keys into the ROT_CREATOR_AUTH* partitions.

Doing so required updating the OTP C file constants generation to allow generating a C file that contains the secure boot key constants that are to be provisioned into OTP during individualization.

Additionally, this fixes a bug in the generation of the perso firmware that was always using the sival SKU OTP SW_CFG constants for the fields that cannot be provisioned during individualization. Now, a separate personalization binary is generated for each SKU.

Lastly, this simplifies how FT binary generation is done. Specifically, FT individualization and personalization binaries are generated based on a SKU. A SKU can be defined as a combination of the following:
- OTP constants
- DICE certificate format used
- personalization extension
This wraps these configuration settings in a dict that consolidates all SKU configuration data in one place, and generates one perso binary per SKU. Before, perso binaries that may not have been used by a SKU were being generated.

This fixes #21554.

### DIFF
--- a/sw/device/silicon_creator/manuf/base/provisioning_inputs.bzl
+++ b/sw/device/silicon_creator/manuf/base/provisioning_inputs.bzl
@@ -2,10 +2,41 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-EARLGREY_SKUS = [
+EARLGREY_OTP_CFGS = [
     "sival",
     "prodc",
 ]
+
+EARLGREY_SKUS = {
+    # OTP Config: SIVAL; DICE Certs: X.509; Additional Certs: None
+    "sival": {
+        "otp": "sival",
+        "dice_libs": ["//sw/device/silicon_creator/lib/cert:dice"],
+        "ext_libs": ["@provisioning_exts//:perso_fw_ext"],
+    },
+    # OTP Config: SIVAL; DICE Certs: CWT; Additional Certs: None
+    # TODO(#24281): uncomment when DICE CWT cert flows are fully supported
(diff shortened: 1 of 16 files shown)


sw/device/silicon_creator/manuf/base/provisioning_inputs.bzl
```diff
@@ -2,11 +2,42 @@
...
 
+EARLGREY_SKUS = {
+    # OTP Config: SIVAL; DICE Certs: X.509; Additional Certs: None
+    "sival": {
+        "otp": "sival",
+        "dice_libs": ["//sw/device/silicon_creator/lib/cert:dice"],
+        "ext_libs": ["@provisioning_exts//:perso_fw_ext"],
+    },
+    # OTP Config: SIVAL; DICE Certs: CWT; Additional Certs: None
+    # TODO(#24281): uncomment when DICE CWT cert flows are fully supported
```
timothytrippel, line 18: FYI: @tommychiu-github @stevenchtsai 

This flow does not work yet because the host test harness does not yet know how to parse CWTs and endorse them. You can uncomment this once you complete the CWT work.

tommychiu-github: Got it, thanks for the heads up.

moidx (approved):

PR #24905, opened 2024-10-28
[topgen] Remove code duplication from TopGenC/Rust

Razer6:
Both of these classes implement the same functionality, either for C or Rust by using the CEnum/CArrayMapping or RustEnum/RustArrayMapping classes. The business logic is identical though.

This PR removes this duplication by factoring out a common base class TopGen. This class is parametrized by using the types from before.


Razer6:
Not sure exactly why the import of `version_file` fails on CI. It is working locally?

a-will:
> Not sure exactly why the import of `version_file` fails on CI. It is working locally?

Is that dependency in pyproject.toml in the repo? (and maybe python-requirements.txt, but I am not sure if one is obsolete)

Razer6:
No, `version_file` is a local dependency to the `util` folder. Previously, `rust.py` imported that but this code now moved to `lib.py` in the same folder. Here, the import fails now.

a-will:
> No, `version_file` is a local dependency to the `util` folder. Previously, `rust.py` imported that but this code now moved to `lib.py` in the same folder. Here, the import fails now.

Ah, it failed during a bazel action, so the dependencies in the BUILD files aren't correct. You'll need to add a py_library target for `util/version_file.py` and have the topgen library depend on `//util:version_file` (or whatever you name that target). `util/version_file.py` isn't currently included in the set of dependent files.

util/topgen/BUILD
```diff
@@ -59,6 +59,7 @@ py_library(
         "//util/reggen:inter_signal",
         "//util/reggen:ip_block",
         "//util/reggen:validate",
+        "//util/version_file",
```
a-will, line 62:
```suggestion
        "//util:version_file",
```

Razer6: Thanks!

rswarbrick (approved):
This looks sensible to me. I was going to moan about the fact that the commit was so large but I'm now convinced that it would have been hard to do otherwise. I guess you could have factored out e.g. the code from `c.py` into `lib.py` (big but obvious), then made it more generic (small), then used that in `rust.py` (big but obvious). But that sounds like a right pain! This PR seems perfectly reasonable :-)

PR #24906, opened 2024-10-28
[dvsim] Add slurm launcher

Razer6:
Slurm is an open-source work management tool, similar to IBM LSF. We are slurm internally and extended dvsim with support for that. We factored all the configs out to environment variables to allow a generic use of it.

### DIFF
--- a/util/dvsim/SlurmLauncher.py
+++ b/util/dvsim/SlurmLauncher.py
@@ -34,8 +34,7 @@ class SlurmLauncher(Launcher):
     def _do_launch(self):
         # replace the values in the shell's env vars if the keys match.
         exports = os.environ.copy()
-        if self.deploy.exports:
-            exports.update(self.deploy.exports)
+        exports.update(self.deploy.exports)
 
         # Clear the magic MAKEFLAGS variable from exports if necessary. This
         # variable is used by recursive Make calls to pass variables from one
--- a/util/dvsim/SlurmLauncher.py
+++ b/util/dvsim/SlurmLauncher.py
@@ -1,0 +1,138 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+import logging as log
+import os
+import shlex
+import subprocess
+
+from Launcher import ErrorMessage, Launcher, LauncherError
+
+
+SLURM_QUEUE = os.environ.get("SLURM_QUEUE", "hw-m")
+SLURM_MEM = os.environ.get("SLURM_MEM", "16G")
+SLURM_MINCPUS = os.environ.get("SLURM_MINCPUS", "8")
+SLURM_TIMEOUT = os.environ.get("SLURM_TIMEOUT", "240")
+SLURM_CPUS_PER_TASK = os.environ.get("SLURM_CPUS_PER_TASK", "8")
+SLURM_SETUP_CMD = os.environ.get("SLURM_SLURM_SETUP_CMD", "")
+
+
+class SlurmLauncher(Launcher):
+    # Misc common SlurmLauncher settings.
+    max_odirs = 5
+
+    def __init__(self, deploy):
+        '''Initialize common class members.'''
+
+        super().__init__(deploy)
+
+        # Popen object when launching the job.
+        self.process = None
+        self.slurm_log_file = self.deploy.get_log_path() + '.slurm'
+
+    def _do_launch(self):
+        # replace the values in the shell's env vars if the keys match.
+        exports = os.environ.copy()
+        if self.deploy.exports:
+            exports.update(self.deploy.exports)
+
+        # Clear the magic MAKEFLAGS variable from exports if necessary. This
+        # variable is used by recursive Make calls to pass variables from one
+        # level to the next. Here, self.cmd is a call to Make but it's
+        # logically a top-level invocation: we don't want to pollute the flow's
+        # Makefile with Make variables from any wrapper that called dvsim.
+        if 'MAKEFLAGS' in exports:
+            del exports['MAKEFLAGS']
+
+        self._dump_env_vars(exports)
+
+        try:
+            with open(self.slurm_log_file, 'w') as out_file:
+                out_file.write("[Executing]:\n{}\n\n".format(self.deploy.cmd))
+                out_file.flush()
+
+                # Add a command delimiter if necessary
+                slurm_setup_cmd = SLURM_SETUP_CMD
+                if slurm_setup_cmd != '' and not slurm_setup_cmd.endswith(';'):
+                    slurm_setup_cmd += ';'
+
+                # Encapsulate the run command with the slurm invocation
+                slurm_cmd = f'srun -p {SLURM_QUEUE} --mem={SLURM_MEM} --mincpus={SLURM_MINCPUS} ' \
+                            f'--time={SLURM_TIMEOUT} --cpus-per-task={SLURM_CPUS_PER_TASK} ' \
+                            f'bash -c "{slurm_setup_cmd} {self.deploy.cmd}"'
+
+                log.info(f'Executing slurm command: {slurm_cmd}')
+                self.process = subprocess.Popen(shlex.split(slurm_cmd),
+                                                bufsize=4096,
+                                                universal_newlines=True,
+                                                stdout=out_file,
+                                                stderr=out_file,
+                                                env=exports)
+        except subprocess.SubprocessError as e:
+            raise LauncherError(f'IO Error: {e}\nSee {self.deploy.get_log_path()}')
+        finally:
+            self._close_process()
+
+        self._link_odir("D")
+
+    def poll(self):
+        '''Check status of the running process
+
+        This returns 'D', 'P' or 'F'. If 'D', the job is still running. If 'P',
+        the job finished successfully. If 'F', the job finished with an error.
+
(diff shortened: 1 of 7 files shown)


a-will:
Nice! I was hoping we'd get a slurm launcher eventually, but I didn't think I'd ever get the time to implement it. :smile:

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,146 @@
...
+
+from Launcher import ErrorMessage, Launcher, LauncherError
+
+
+SLURM_QUEUE = os.environ.get("SLURM_QUEUE", "hw-m")
+SLURM_MEM = os.environ.get("SLURM_MEM", "16G")
+SLURM_MINCPUS = os.environ.get("SLURM_MINCPUS", "8")
+SLURM_TIMEOUT = os.environ.get("SLURM_TIMEOUT", "240")
+SLURM_CPUS_PER_TASK = os.environ.get("SLURM_CPUS_PER_TASK", "8")
+SLURM_SETUP_CMD = os.environ.get("SLURM_SLURM_SETUP_CMD", "module purge; module load opentitan")
```
rswarbrick, line 17: This particular line seems to be pretty magic and presumably shows us something about the Rivos setup :-) Is there a way that the "setup command" can appear in a log file that isn't checked in?

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, deploy):
+        '''Initialize common class members.'''
+
+        super().__init__(deploy)
+
+        # Popen object when launching the job.
+        self.process = None
+
+    def _do_launch(self):
+        global job_name
```
rswarbrick, line 33: I don't think this is needed any more?

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,146 @@
...
+        # Update the shell's env vars with self.exports. Values in exports must
+        # replace the values in the shell's env vars if the keys match.
+        exports = os.environ.copy()
+        if self.deploy.exports:
```
rswarbrick, line 37: Is the `if` needed here?
```
Python 3.12.7 (main, Oct  3 2024, 15:15:22) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = {1: 1, 2: 2, 3: 3}
>>> a.update({})
>>> a
{1: 1, 2: 2, 3: 3}
```

(or is the point that it's `None` or something? In which case maybe `if self.deploy.exports is not None` ?)

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,146 @@
...
+        # level to the next. Here, self.cmd is a call to Make but it's
+        # logically a top-level invocation: we don't want to pollute the flow's
+        # Makefile with Make variables from any wrapper that called dvsim.
+        if 'MAKEFLAGS' in exports:
+            del exports['MAKEFLAGS']
+
+        self._dump_env_vars(exports)
+
+        try:
+            f = open(self.deploy.get_log_path() + '.slurm',
```
rswarbrick, line 51: Would it be possible to do this with `with open(...) as f:`? It might be slightly easier to read? You'd also be able to get rid of the `f.close` on line 77.

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,146 @@
...
+        This function must only be called after running self.dispatch_cmd() and
+        must not be called again once it has returned 'P' or 'F'.
+        '''
+
+        assert self.process is not None
+        if self.process.poll() is None:
+            return 'D'
+        # -------------------------------------
+        # copy Slurm job results to log file
+        if os.path.exists(self.deploy.get_log_path() + '.slurm'):
```
rswarbrick, line 94: I think you should probably define a local variable for this path: it gets used several times.

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,146 @@
...
+            file1.close()
+            f = open(self.deploy.get_log_path(),
+                     "a",
+                     encoding="UTF-8",
+                     errors="surrogateescape")
+            for line in Lines:
+                f.write(line)
+            f.flush()
+            os.remove(self.deploy.get_log_path() + '.slurm')
+            f.close()
```
rswarbrick, line 107: Is this just moving the file, but splatting encoding errors (depending on your current encoding)? That seems a little complicated to me: do we really need to transcode like this?

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,146 @@
...
+        # -------------------------------------
+
+        self.exit_code = self.process.returncode
+        status, err_msg = self._check_status()
+        self._post_finish(status, err_msg)
+        return status
+
+    def kill(self):
+        global job_name
```
rswarbrick, line 116: Again, I don't think this is needed?

util/dvsim/SlurmLauncher.py
Razer6: Most probably. Though it is using standard environment modules. What about leaving that as an empty string as default that a user can overwrite?

util/dvsim/SlurmLauncher.py
Razer6: agree.

util/dvsim/SlurmLauncher.py
Razer6: I was using the same code pattern as in other launchers. I traced it down, `deploy.exports` is coming from [here](https://github.com/lowRISC/opentitan/blob/master/util/dvsim/Deploy.py#L213) and always at least defines an empty dict.

What about a second commit, removing the `if` for all launchers?

util/dvsim/SlurmLauncher.py
Razer6: remmoved the encoding stuff.

Razer6:
@rswarbrick Thank you for looking into it. I addressed your comments, can you take another look?

util/dvsim/SlurmLauncher.py
rswarbrick: That's perfect: thanks!

rswarbrick (approved):
This looks really nice: thanks!

util/dvsim/SlurmLauncher.py
matutem: It is pythonesque to use python context managers since it avoids some error handling pitfalls. You are highly encouraged to use them everywhere you do I/O.

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,137 @@
...
+        status, err_msg = self._check_status()
+        self._post_finish(status, err_msg)
+        return status
+
+    def kill(self):
+        '''Kill the running process.
+
+        This must be called between dispatching and reaping the process (the
+        same window as poll()).
+
```
matutem, line 112: Remove this line?

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,137 @@
...
+        must not be called again once it has returned 'P' or 'F'.
+        '''
+
+        assert self.process is not None
+        if self.process.poll() is None:
+            return 'D'
+
+        # Copy slurm job results to log file
+        if os.path.exists(self.slurm_log_file):
+            with open(self.slurm_log_file, 'r') as slurm_file:
```
matutem, line 94: Use `try` to handle errors.

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,137 @@
...
+                lines = slurm_file.readlines()
```
matutem, line 95: This is unefficient: you are reading the full file, breaking the string at each `\n` into multiple strings, and then writing this line by line. You might as well just do something like out_file.write(slurm_file.read()), or better yet, shutil.copyfileobj(slurm_file, out_file).

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,137 @@
...
+            del exports['MAKEFLAGS']
+
+        self._dump_env_vars(exports)
+
+        try:
+            with open(self.slurm_log_file, 'w') as out_file:
+                out_file.write("[Executing]:\n{}\n\n".format(self.deploy.cmd))
+                out_file.flush()
+
+                # Add a command delimiter if necessary
```
matutem, line 54: I'd create the slurm_cmd outside the try block so the try just does IO and creates the process.

util/dvsim/SlurmLauncher.py
```diff
@@ -0,0 +1,137 @@
...
+                log.info(f'Executing slurm command: {slurm_cmd}')
+                self.process = subprocess.Popen(shlex.split(slurm_cmd),
+                                                bufsize=4096,
+                                                universal_newlines=True,
+                                                stdout=out_file,
+                                                stderr=out_file,
+                                                env=exports)
+        except subprocess.SubprocessError as e:
+            raise LauncherError(f'IO Error: {e}\nSee {self.deploy.get_log_path()}')
+        finally:
```
matutem, line 73: And an `except` for errors related to opening and writing to out_file?

matutem (commented):
Robert, could you look into these comments and send them to me for review if you decide to make some of these changes.

util/dvsim/SlurmLauncher.py
Razer6: Sounds good!

PR #24916, opened 2024-10-28
[hw,rstmgr,dv] Use Xcelium compatible syntax

Razer6:
This PR adds two changes to make the rstmgr simulation compatible with xcelium:

1. It removes the `@edge` and replaces it with a combination of `@posedge` and `@negedge`
2. Because of strong typing rules it's possible to directly read via `uvm_hdl_read` into a `prim_mubi_pkg::mubi4_t`. Read into a `logic` and cast the result to `prim_mubi_pkg::mubi4_t`


rswarbrick (approved):
This looks sensible to me. I'm slightly surprised that Xcelium doesn't support `@edge`: I didn't know the keyword but looked at the 1800 spec and found it immediately. But I agree that this translation matches :-)

PR #24919, opened 2024-10-28
[manuf] refactor FT flows to enable downstream OTP definitions

timothytrippel:
This refactors the FT provisioning Bazel build binary and test targets to enable defining downstream OTP configurations and ingesting them into FT provisioning flows.

Additionally, this removes the reliance on build configurations to select which personalization extension to use, and simply builds and tests all FT provisioning flows that can be enumerated at build time. This makes it easier to run and test all flows upstream only, with: `bazel test //sw/device/silicon_creator/manuf/base:ft_provision_cw310`, or all flows upstream and downstream with:
`PROV_EXTS_DIR=... bazel test //sw/device/silicon_creator/manuf/base:ft_provision_cw310`. As a result, this also simplifies the build graph by removing configuration transistions.


timothytrippel:
This fixes #24920.

PR #24921, opened 2024-10-29
[ci] configure Bazel cache for GitHub actions

nbdd0121:
This should not be merged yet until we have configured bucket permission so that workload identity federation can write to the cache properly.

### DIFF
--- a/.github/actions/prepare-env/action.yml
+++ b/.github/actions/prepare-env/action.yml
@@ -2,8 +2,8 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-name: Install dependencies
-description: Install system dependencies needed for OpenTitan
+name: Prepare environment
+description: Install dependencies and prepare environment needed for OpenTitan
 
 inputs:
   verilator-version:
@@ -22,6 +22,10 @@ inputs:
     description: Path at which to install Verible
     required: true
     default: /tools/verible
+  configure-bazel:
+    description: Configure Bazel to use remote cache
+    required: true
+    default: true
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


.github/actions/configure-bazel/action.yml
```diff
@@ -0,0 +1,39 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Populate ~/.bazelrc
+      run: |
+        cp ci/.bazelrc ~/.bazelrc
+        # Inject the OS version into a parameter used in the action key computation to
+        # avoid collisions between different operating systems in the caches.
+        # See #14695 for more information.
+        echo "build --remote_default_exec_properties=OSVersion=\"$(lsb_release -ds)\"" >> ~/.bazelrc
```
jwnrt, line 17: Hmm, I wonder if this is still an issue. Can Bazel really not use the cache between glibc versions? Does it link things like `opentitantool` against the system glibc?

jwnrt (approved):
Looks good to me, thanks

.github/actions/configure-bazel/action.yml
nbdd0121: A binary built against new glibc in general cannot be used with an old libc, so we cannot use binaries built on Ubuntu 22.04 on Ubuntu 20.04, for example.

jwnrt: I was hoping that Bazel's hermetic builds also included the libc, but it seems not

jwnrt:
I've added the WIF pricipal to the bucket (I think), can I test this by pushing to a branch on the repo?

jwnrt:
Good news is that it works, bad news is that some of the jobs are taking a lot longer:

* Verilated English Breakfast is very consistently 12m, but this run took ~16m.
* Lint (slow) is consistently 20m, but this run took over 30m just for the documentation build.

Any idea why it would be slowed down so much?

jwnrt:
Ran again and it took much less time. The first run was either a transient error or it had to repopulate the cache which took a long time.

https://github.com/lowRISC/opentitan/actions/runs/11661952950

jwnrt (approved):
LGTM, but could you bring back the `- name: Prepare environment` lines? They're easier to read than `./.github/actions/prepare-environment` in the UI

PR #24929, opened 2024-10-29
[pwm,dv] Very basic tidy-ups

rswarbrick:
This is a sequence of very basic tidy-ups in the DV code. There isn't supposed to be any functional change, but this gets rid of lots of duplicated state.


hcallahan-lowrisc (approved):
This LGTM, thanks for the cleanups.

PR #24932, opened 2024-10-29
[hw,keymgr,dv] Fix unsupported AND within an if

Razer6:
The AND within the if is not supported in xcelium. Factor this out into a dedicated signal.

### DIFF
--- a/hw/ip/keymgr/dv/env/keymgr_scoreboard.sv
+++ b/hw/ip/keymgr/dv/env/keymgr_scoreboard.sv
@@ -1236,14 +1236,14 @@ class keymgr_scoreboard extends cip_base_scoreboard #(
   virtual function keymgr_pkg::keymgr_working_state_e get_next_state(
       keymgr_pkg::keymgr_working_state_e cur = current_state);
     keymgr_pkg::keymgr_working_state_e next_linear_state = keymgr_env_pkg::get_next_state(cur);
+    logic creator_root_key_valid = cfg.keymgr_vif.otp_key.creator_root_key_share0_valid &
+                                   cfg.keymgr_vif.otp_key.creator_root_key_share1_valid;
 
     if (// If keymgr is not enabled in the current LC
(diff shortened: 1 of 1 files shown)


hw/ip/keymgr/dv/env/keymgr_scoreboard.sv
```diff
@@ -1234,16 +1234,16 @@ class keymgr_scoreboard extends cip_base_scoreboard #(
   endfunction
 
   virtual function keymgr_pkg::keymgr_working_state_e get_next_state(
-      keymgr_pkg::keymgr_working_state_e cur = current_state);
+    keymgr_pkg::keymgr_working_state_e cur = current_state);
     keymgr_pkg::keymgr_working_state_e next_linear_state = keymgr_env_pkg::get_next_state(cur);
+    logic shares_valid = cfg.keymgr_vif.otp_key.creator_root_key_share0_valid &
```
andreaskurth, line 1239:
```suggestion
    logic creator_root_key_valid = cfg.keymgr_vif.otp_key.creator_root_key_share0_valid &
```
I think that would be more descriptive

hw/ip/keymgr/dv/env/keymgr_scoreboard.sv
```diff
@@ -1234,16 +1234,16 @@ class keymgr_scoreboard extends cip_base_scoreboard #(
   endfunction
 
   virtual function keymgr_pkg::keymgr_working_state_e get_next_state(
-      keymgr_pkg::keymgr_working_state_e cur = current_state);
+    keymgr_pkg::keymgr_working_state_e cur = current_state);
```
andreaskurth, line 1237: Accidental change? I don't think this indentation should be reduced

andreaskurth (commented):
LGTM other than two small comments

Razer6:
Thanks for looking into this. I updated the commit for that.

PR #24941, opened 2024-10-30
[dvsim] Create symlink to source folder for xcelium DPI builds

Razer6:
Within the build directory, dvsim creates a folder `sim-vcs/src`, which contains the source fusesoc packages, e.g., `lowrisc_constants_top_pkg_0`.
Xcelium, however, expects a slightly different layout when it comes to building DPI models of IPs. It expects the folder containing the fusesoc packages to be available at `src`.

To overcome this different folder layout, this PR adds a symlink in the scratch directory between those folders when using Xcelium.

There might be a different solution to this problem, but this at least works.

### DIFF
--- a/util/dvsim/Deploy.py
+++ b/util/dvsim/Deploy.py
@@ -242,7 +242,7 @@ class Deploy():
         the final resolved 'cmd' & the exports. The 'name' field will be unique
         to 'item' and 'self', so we take that out of the comparison.
         """
-        if type(self) != type(item):
+        if not isinstance(self, type(item)):
             return False
 
         # Check if the cmd field is identical.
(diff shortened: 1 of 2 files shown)


rswarbrick:
I'm a bit surprised about this: I actually run most blocks' DV code with Xcelium at the moment already (without the symlink). What problem is this change solving?

util/dvsim/Deploy.py
```diff
@@ -242,7 +242,7 @@ def is_equivalent_job(self, item):
         the final resolved 'cmd' & the exports. The 'name' field will be unique
         to 'item' and 'self', so we take that out of the comparison.
         """
-        if type(self) != type(item):
+        if not isinstance(self, type(item)):
```
Razer6, line 245: That one was only needed to make flake8 happy

Razer6:
Closing that as a downstreaem issue.

PR #24944, opened 2024-10-30
[hmac,sw] Workaround to avoid hang - Stop CMD is not triggered

martin-velay:
- Implement a workaround in C cryptolib as HW is hanged
 after a stop in certain conditions.
- Refer to issue https://github.com/lowRISC/opentitan/issues/24767

### DIFF
--- a/sw/device/lib/crypto/drivers/hmac.c
+++ b/sw/device/lib/crypto/drivers/hmac.c
@@ -9,6 +9,7 @@
 #include "sw/device/lib/base/hardened.h"
 #include "sw/device/lib/base/memory.h"
 #include "sw/device/lib/crypto/impl/status.h"
+#include "sw/device/lib/runtime/ibex.h"
 
 #include "hmac_regs.h"  // Generated.
 #include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
@@ -67,6 +68,13 @@ OT_ASSERT_ENUM_VALUE(HMAC_DIGEST_15_REG_OFFSET, HMAC_DIGEST_14_REG_OFFSET + 4);
 enum {
   /* The beginning of the address space of HMAC. */
   kHmacBaseAddr = TOP_EARLGREY_HMAC_BASE_ADDR,
+  /* Timeout value for the polling
+   * As max 4 clock cycles are required to perform a read access to the
+   * register. And as it should take less than 64 clock cycles in SHA2-256
+   * and 80 clock cycles in SHA2-384/512, let's take some margin and consider
+   * that 100 loops are a way enough to see IDLE status.
+   */
+  kNumIterTimeout = 100,
 };
 
 /**
@@ -74,8 +82,6 @@ enum {
  *
  * It returns error if HMAC HWIP becomes idle without firing `hmac_done`
  * interrupt.
- *
- * TODO(#23191): It might be beneficial to have a timeout value for the polling.
  *
  * @return Result of the operation.
  */
@@ -85,8 +91,13 @@ static status_t hmac_idle_wait(void) {
   // Initialize `status_reg = 0` so that the loop starts with the assumption
   // that HMAC HWIP is not idle.
   uint32_t status_reg = 0;
+  uint32_t attempt_cnt = 0;
   while (bitfield_bit32_read(status_reg, HMAC_STATUS_HMAC_IDLE_BIT) == 0) {
     status_reg = abs_mmio_read32(kHmacBaseAddr + HMAC_STATUS_REG_OFFSET);
+    attempt_cnt++;
+    if (attempt_cnt >= kNumIterTimeout) {
+      return OTCRYPTO_FATAL_ERR;
+    }
   }
 
   // Verify that HMAC HWIP raises `hmac_done` bit.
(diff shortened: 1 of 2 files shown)


sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -256,6 +266,36 @@ static void msg_fifo_write(const uint8_t *message, size_t message_len) {
   }
 }
 
+/**
+ * Temporary workaround linked to issue #24767
+ *
+ * The HMAC HWIP is not tell to stop. This will cause the HW to be in a non
```
jadephilipoom, line 272: nit: I think there's a typo in the first sentence here.

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -256,6 +266,36 @@ static void msg_fifo_write(const uint8_t *message, size_t message_len) {
...
+ * SHA2-384/512).
+ *
+ * @param[out] ctx Context to which values are written.
+ */
+static void tmp_avoid_hw_hang(hmac_ctx_t *ctx) {
+  // Insert delay which should be equivalent to at least 80 clock cycles
+  for (size_t i = 0; i < 100; i = launder32(i + 1)) {
+    // NULL statement
+    ;
+  }
```
jadephilipoom, line 287: I think you can use the `IBEX_SPIN_FOR` macro here to ensure the compiler doesn't "helpfully" optimize away the loop:
```suggestion
uint64_t delay_cycles = 80; // or maybe 100 or so to be safe
uint64_t delay_usec = udiv64_slow(delay_cycles * 1000000, kClockFreqCpuHz, NULL);
IBEX_SPIN_FOR(delay_usec);
```

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -85,8 +90,13 @@ static status_t hmac_idle_wait(void) {
   // Initialize `status_reg = 0` so that the loop starts with the assumption
   // that HMAC HWIP is not idle.
   uint32_t status_reg = 0;
+  uint32_t attempt_cnt = 0;
   while (bitfield_bit32_read(status_reg, HMAC_STATUS_HMAC_IDLE_BIT) == 0) {
     status_reg = abs_mmio_read32(kHmacBaseAddr + HMAC_STATUS_REG_OFFSET);
+    attempt_cnt++;
+    if (attempt_cnt == kNumIterTimeout) {
```
jadephilipoom, line 97: nit: maybe just to be safe
```suggestion
    if (attempt_cnt >= kNumIterTimeout) {
```

jadephilipoom (approved):
Assuming that this reliably avoids the hang, I think it's a good simple workaround and the 80-ish cycles delay should be tolerable. Couple of nitpicks but LGTM.

sw/device/lib/crypto/drivers/hmac.c
vogelpi: Also, can we use the constant defined above for this, i.e., `kNumIterTimeout`?

vogelpi (approved):
Thanks @martin-velay for trying this out and implementing the PR! This looks good modulo the points raised by @jadephilipoom .

Thanks @jadephilipoom for your review and @vsukhoml for reporting the issue and for providing the simpler workaround, great team work everybody!

sw/device/lib/crypto/drivers/hmac.c
martin-velay: Thanks Jade for the trick, that's much better. And yes Pirmin, it makes sense to use this constant as it refers to the same HW constraint.

martin-velay:
@vogelpi regarding [this comment](https://github.com/lowRISC/opentitan/pull/24944#discussion_r1824091638), this for loop will go away and be replaced by `IBEX_SPIN_FOR` see [comment here](https://github.com/lowRISC/opentitan/pull/24944#discussion_r1822876763)

sw/device/lib/crypto/drivers/hmac.c
martin-velay: @jadephilipoom I have implemented something else based on what we've discussed together offline. Can you please have a look at these changes?

vsukhoml: this constant can be lower, as loop is 2 instructions, so 80 cycles needs only 40 iterations max and you already did something else after sending to FIFO (like return from function, storing/restoring regs), etc.

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -256,6 +267,38 @@ static void msg_fifo_write(const uint8_t *message, size_t message_len) {
...
+  timeout.start = ibex_mcycle_read();
+  while (!ibex_timeout_check(&timeout)) {
+    // NULL statement
+  }
+
+  // Save current context as it is updated after each block even if stop is not
+  // triggered
+  context_save(ctx);
+
+  // Trigger hash_process
```
vsukhoml, line 296: here you can reset HMAC_EN bit in CFG to make it faster, as otherwise `hash_process` will run HMAC's opad and take longer. and potentially with side-channel implications. `hash_process` serves only the purpose to bring HMAC IP into consistent state.

martin-velay: After discussing with @vogelpi, and doing some tests on the FPGA with some prints, this is not possible as the HMAC won't be in idle state at that time. This is a pre-requisite from the [spec](https://opentitan.org/book/hw/ip/hmac/doc/registers.html#cfg): "The register is updated when the engine is in Idle." otherwise the value is discarded. In our case, we'll move to the idle state only after the `hash_process` triggered will be done.
So I am suggesting to just remove this hmac_en=0, as it will be ignored.

vsukhoml: Thanks! I'll update my workaround as well. Consider now to not enable HMAC_EN if HW was already initialized, as it doesn't impact processing.

PR #24945, opened 2024-10-30
[flash_ctrl,dv] Correct syntax in flash_ctrl_phy_arb_redun_vseq

rswarbrick:
It seems that VCS accepts the code that was there but it isn't quite in accordance with the allowed syntax for structure literals (see IEEE 1800, section 5.10). Fortunately, the fix is pretty trivial!


Razer6 (approved):
I can confirm this works with xcelium.

marnovandermaas (approved):

PR #24946, opened 2024-10-30
[sival] Include cryptolib tests in earlgrey sival testplan

jwnrt:
These tests must pass for silicon validation, so we should include them in the testplan for Earl Grey.


moidx (approved):

jwnrt:
The CI failures here are unrelated. The FPGA test could not have been affected by this change, and the `Verify FPGA jobs` failure is due to a race in our pipeline (see #25164).

PR #24947, opened 2024-10-30
[ci] Prioritise CW340 jobs in CI over CW310

jwnrt:
This swaps the priority of CW310 and CW340 jobs. If a test can run on the CW340 then it will, else it will fall back to the CW310.

This is required so that we can check tests as they're moved over to running on the CW340s.


pamaury (approved):
As long as we have enough CW340 to handle the load, this seems fine with me

jwnrt:
We should do, there are currently 6 CW340s and 5 CW310s (two are currently broken/offline)

jwnrt:
CI runs look as expected: one or two tests moving to CW340s but not many assigned yet, so no large impact. I'm going to merge.

PR #24950, opened 2024-10-30
[sival] Add a nightly job on github actions

engdoreis:
This PR adds a new nightly Job using GitHub actions, this is meant to be the replacement of the nightly running on azure, but for now it only has the FPGA nightly for SiVal.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,62 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Nightly
+on:
+  schedule:
+    - cron: "00 04 * * *"
+
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to run the nightly'
+        required: true
+        type: string
+
+jobs:
+  FPGA-CW310-SiVal-Nightly:
+    name: Fpga cw310 sival tests
+    runs-on: [ubuntu-20.04-fpga, cw310]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Required for the bitstream cache to work.
+          ref: ${{ inputs.branch || 'earlgrey_1.0.0' }} # Schedule only work on the default branch, but we want to run on a different branch.
+
+      - name: Install dependencies
+        uses: ./.github/actions/install-deps
+
+      - name: Update hyperdebug
+        # We run the update command twice to workaround an issue with udev on the container.
+        # Where rusb cannot dynamically update its device list in CI (udev is not completely
+        # functional). If the device is in normal mode, the first thing that opentitantool
+        # does is to switch it to DFU mode and wait until it reconnects. This reconnection is
+        # never detected. But if we run the tool another time, the device list is queried again
+        # and opentitantool can finish the update. The device will now reboot in normal mode
+        # and work for the hyperdebug job.
+        run: |
+          ./bazelisk.sh run //sw/host/opentitantool -- --interface=hyperdebug_dfu transport update-firmware \
+          || ./bazelisk.sh run //sw/host/opentitantool -- --interface=hyperdebug_dfu transport update-firmware
+
+      - name: Run tests after ROM boot stage
+        run: |
+          module load xilinx/vivado
+          bazel_tests="$(mktemp)"
+          ./bazelisk.sh query 'attr("tags", "[\[ ]cw310_sival[,\]]", tests(//...))' \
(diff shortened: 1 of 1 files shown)


.github/workflows/sival_nightly.yml
```diff
@@ -0,0 +1,38 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Nigthly
```
jwnrt, line 5:
```suggestion
name: Nightly
```

.github/workflows/sival_nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  schedule:
+    - cron: "00 04 * * *"
+
+jobs:
+  FPGA-CW310-SiVal-Nightly:
```
jwnrt, line 11: nit: we typically use lower case for these IDs

.github/workflows/sival_nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+    name: FPGA CW310 SiVal nightly tests
```
jwnrt, line 12: nit: you don't need to repeat "nightly tests" since it will already show as "Nightly / FPGA CW310 SiVal"

.github/workflows/sival_nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: [ubuntu-20.04-fpga, cw310]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Required for the bitstream cache work.
+          ref: earlgrey_1.0.0 # schedule only work on the default branch, but we want to run another branch.
+
+      - uses: ./.github/actions/install-deps
```
jwnrt, line 20:
```suggestion
      - name: Install dependencies
        uses: ./.github/actions/install-deps
```

.github/workflows/sival_nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          module load xilinx/vivado
+          ./bazelisk.sh sync --configure
+          ./bazelisk.sh run //sw/host/opentitantool -- --interface=hyperdebug_dfu transport update-firmware
+          ./bazelisk.sh test --build_tests_only \
+          $(bash ./bazelisk.sh query 'attr("tags", "[\[ ]cw310_sival[,\]]", tests(//...))' \
+                                                    | grep -v examples \
+                                                    | grep -v third_party \
+                                                    | grep -v penetrationtests)
+      - name: ROMEXT boot stage
```
jwnrt, line 31:
```suggestion
      - name: ROM_EXT boot stage
```

.github/workflows/sival_nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Required for the bitstream cache work.
+          ref: earlgrey_1.0.0 # schedule only work on the default branch, but we want to run another branch.
+
+      - uses: ./.github/actions/install-deps
+      - name: ROM boot stage
+        run: |
+          module load xilinx/vivado
+          ./bazelisk.sh sync --configure
```
jwnrt, line 24: I don't think you need to do this if it's the first time the command runs, only if you change branches afterwards

.github/workflows/sival_nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+          ./bazelisk.sh run //sw/host/opentitantool -- --interface=hyperdebug_dfu transport update-firmware
+          ./bazelisk.sh test --build_tests_only \
+          $(bash ./bazelisk.sh query 'attr("tags", "[\[ ]cw310_sival[,\]]", tests(//...))' \
```
jwnrt, line 27: I think we should do this as a "bazel test pattern file" in case the list of tests gets too long to fit in command line arguments.

It would be something like

```sh
bazel_tests="$(mktemp)"
./bazelisk.sh query 'attr("tags", "[\[ ]cw310_sival[,\]]", tests(//...))' \
    | grep -v examples \
    | grep -v third_party \
    | grep -v penetrationtests \
    > "$bazel_tests"
./bazelisk.sh test --build_tests_only --target_pattern_file="$bazel_tests"
```

jwnrt (changes requested):
Thanks, good to see the FPGAs working in github actions.

You'll need to retarget this PR at the master branch to get the cron working, does GitHub let you edit that for an existing PR?

jwnrt (approved):
Looks fine to me, thanks

Would you mind adding a second trigger so that we can run this manually if we need to?

```yaml
on:
  workspace_dispatch:
```
is all you need to add.

Issue #24942, opened 2024-10-30
[manuf] Propose to store DICE certificates in the flash_info with the "Cert Header"


tommychiu-github:
### Test point name

NA

### Host side component

None

### OpenTitanTool infrastructure implemented

None

### Silicon Validation (SiVal)

Yes

### Emulation Targets

- [ ] None
- [ ] CW310
- [x] Hyperdebug + CW310

### Checklist

Please fill out this checklist as items are completed. Link to PRs and issues as appropriate.
- [x] Device-side (C) component developed

Currently, during ft_perso, the host write the RAW endorsed DICE certificate to flash_info pages.
When it needs to access them later on, the certificate length is calculated by parsing the header (1 or 2 bytes). This design works for X509 but not CWT-CBOR since there's no such information in some CBOR types (map, array specifically).
Ref: 
- https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/base/ft_personalize.c;l=627
- https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/base/ft_personalize.c;l=316

 I'm proposing to change the certificat storing format as below. The certificate is prepended by a 16 bits header followed by the ASCII  characters of the certificate name, followed by the certificate body.
This format is used already when transfering tbs & certificates from host vs. device.

The original X509 parsing logic can be reused for cross checking the result from the Cert Header.

```
   *  d15                                         d0
   * +-------------+--------------------------------+
   * | 4 bit       |            12 bits             |
   * +-------------+--------------------------------+
   * | name length |12 bits total cert payload size | <-- Cert Header
   * +-------------+--------------------------------+
   * |             cert name string                 |
   * +----------------------------------------------+
   * |                   cert                       |
   * +----------------------------------------------+
```

## Update
Move to the new storing format based on the [conclusion](https://github.com/lowRISC/opentitan/pull/24951#issuecomment-2449088327).
```
   *  d15                                         d0
   * +-------------+--------------------------------+
   * | 4 bit type  |   12 bits total object size    | <-- Object Header
   * +-------------+--------------------------------+
   * | name length |12 bits total cert payload size | <-- Cert Header
   * +-------------+--------------------------------+
   * |             cert name string                 |
   * +----------------------------------------------+
   * |                   cert                       |
   * +----------------------------------------------+
```

tommychiu-github:
> [@timothytrippel](https://github.com/timothytrippel) for viz

[...]

PR #24951, opened 2024-10-30
[manuf] Prepend a header to certificate before storing on flash

tommychiu-github:
Currently, during ft_perso, the host write the RAW endorsed DICE certificate to flash_info pages.
When it needs to access them later on, the certificate length is calculated by parsing the ASN.1 header (1 or 2 bytes).
This design works for X.509 (which are ASN.1 encoded) but not CWT-CBOR since there's no such information in some CBOR types (map, array specifically).

Proposed format:
```
   *  d15                                         d0
   * +-------------+--------------------------------+
   * | 4 bit type  |   12 bits total object size    | <-- Object Header
   * +-------------+--------------------------------+
   * | name length |12 bits total cert payload size | <-- Cert Header
   * +-------------+--------------------------------+
   * |             cert name string                 |
   * +----------------------------------------------+
   * |                   cert                       |
   * +----------------------------------------------+
```

Fixes: #24942
Test: //sw/device/silicon_creator/manuf/base:ft_provision_cw340

### DIFF
--- a/sw/device/silicon_creator/manuf/base/ft_personalize.c
+++ b/sw/device/silicon_creator/manuf/base/ft_personalize.c
@@ -307,13 +307,28 @@ static void compute_keymgr_owner_binding(manuf_certgen_inputs_t *inputs) {
 static status_t hash_certificate(const flash_ctrl_info_page_t *page,
                                  size_t offset, size_t *size) {
   uint8_t buffer[1024];  // 1K should be enough for the largest certificate.
+  perso_tlv_object_header_t objh;
+  uint16_t obj_size;
+  perso_tlv_object_type_t obj_type;
+  perso_tlv_cert_header_t crth;
   uint32_t cert_size;
-  uint32_t bytes_read;
+  uint16_t wrapped_cert_size;
+  uint16_t name_len;
+  uint16_t cert_header_and_name_len;
 
+  static_assert(sizeof(objh) <= 4, "Cert Header size too large");
+  static_assert(sizeof(crth) <= 4, "Object Header size too large");
   // Read the first word of the certificate which contains it's size.
   TRY(flash_ctrl_info_read(page, offset, 1, buffer));
-  bytes_read = sizeof(uint32_t);
-  cert_size = cert_x509_asn1_decode_size_header(buffer);
+  memcpy(&objh, buffer, sizeof(perso_tlv_object_header_t));
+  PERSO_TLV_GET_FIELD(Objh, Size, objh, &obj_size);
+  // Try to parse Cert Header, and calculate the real cert size
+  memcpy(&crth, &buffer[sizeof(objh)], sizeof(perso_tlv_cert_header_t));
+  PERSO_TLV_GET_FIELD(Crth, NameSize, crth, &name_len);
+  PERSO_TLV_GET_FIELD(Crth, Size, crth, &wrapped_cert_size);
+  cert_header_and_name_len = sizeof(crth) + name_len;
+  cert_size = wrapped_cert_size - cert_header_and_name_len;
+
   if (cert_size == 0) {
     LOG_ERROR("Inconsistent certificate header %02x %02x page %x:%x", buffer[0],
               buffer[1], page->base_addr, offset);
@@ -330,14 +345,24 @@ static status_t hash_certificate(const flash_ctrl_info_page_t *page,
     return DATA_LOSS();
   }
 
-  offset += bytes_read;
-  TRY(flash_ctrl_info_read(page, offset,
-                           util_size_to_words(cert_size - bytes_read),
-                           buffer + bytes_read));
-  hmac_sha256_update(buffer, cert_size);
+  TRY(flash_ctrl_info_read(page, offset, util_size_to_words(obj_size), buffer));
+  size_t cert_offset = sizeof(objh) + cert_header_and_name_len;
+  PERSO_TLV_GET_FIELD(Objh, Type, objh, &obj_type);
+  if (obj_type == kPersoObjectTypeX509Tbs ||
+      obj_type == kPersoObjectTypeX509Cert) {
+    uint32_t decoded_cert_size =
+        cert_x509_asn1_decode_size_header(&buffer[cert_offset]);
+    if (cert_size != decoded_cert_size) {
+      LOG_ERROR("Unexpected cert size %d instead of %d", decoded_cert_size,
+                cert_size);
+      return INTERNAL();
+    }
+  }
+
+  hmac_sha256_update(&buffer[cert_offset], cert_size);
 
   if (size) {
-    *size = cert_size;
+    *size = obj_size;
   }
   return OK_STATUS();
 }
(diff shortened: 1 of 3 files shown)


tommychiu-github:
@timothytrippel for review, thanks.

vbendeb:
Since we need to be able to store certs  in flash in different formats, why not store them along with the header containing the object type. This way we could just store the entire wrapped object received from the host, and then when reading from the flash use the header to tell what kind of object it is.

Also, please add the storage format to the patch description or to code comments.

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -616,15 +624,16 @@ static status_t personalize_endorse_certificates(ujson_t *uj) {
       // We just prepared the set of wrapped certificates, let's trust that the
       // data is correct and does not need more validation.
       memcpy(&crth, certs, sizeof(crth));
-      certs += sizeof(crth);
       PERSO_TLV_GET_FIELD(Crth, NameSize, crth, &name_len);
-      memcpy(name, certs, name_len);
+      memcpy(name, (void *)(certs + sizeof(crth)), name_len);
```
timothytrippel, line 628:
```suggestion
      memcpy(name, certs + sizeof(crth), name_len);
```
I don't think you need the cast here do you?

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -308,12 +308,22 @@ static status_t hash_certificate(const flash_ctrl_info_page_t *page,
...
+  uint16_t wrapped_cert_size;
+  uint16_t name_len;
+  uint16_t hdr_and_name_len;
 
   // Read the first word of the certificate which contains it's size.
   TRY(flash_ctrl_info_read(page, offset, 1, buffer));
-  bytes_read = sizeof(uint32_t);
-  cert_size = cert_x509_asn1_decode_size_header(buffer);
+  perso_tlv_cert_header_t crth;
+  memcpy(&crth, &buffer[0], sizeof(crth));
```
timothytrippel, line 318:
```suggestion
  memcpy(&crth, buffer, sizeof(crth));
```

Maybe we should also add a `static_assert()` some where that checks that `sizeof(crth) <= 4`, since we only read 1 word from the flash page on line 316 above (which is most efficient)?

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -308,12 +308,22 @@ static status_t hash_certificate(const flash_ctrl_info_page_t *page,
...
+  PERSO_TLV_GET_FIELD(Crth, NameSize, crth, &name_len);
+  // Try to parse Cert Header, and calculate the real cert size
```
timothytrippel, line 320: Is this too hard to add in this PR? Cant you just also keep the call to `cert_x509_asn1_decode_size_header(...)` and check its return value `== cert_size` below? You would need to also read in at least one byte of the cert itself, so maybe we do this check below on line 345 after reading in the entire cert?

timothytrippel (commented):
mostly LGTM, just a couple of suggestions. Thanks @tommychiu-github !

timothytrippel:
> Since we need to be able to store certs in flash in different formats, why not store them along with the header containing the object type. This way we could just store the entire wrapped object received from the host, and then when reading from the flash use the header to tell what kind of object it is.
> 
> Also, please add the storage format to the patch description or to code comments.

This sounds like a good idea to me as well (will simplify your future work @tommychiu-github), in such case, I take it the idea would be to add the following types to this [enum](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/base/perso_tlv_data.h;drc=32c42ca99db21abd0e2482ae1e681a0dac2616e0;l=28) (in a future PR that adds CWT cert formats)?
 - `kPersoObjectTypeCwtTbs`
 - `kPersoObjectTypeCwtCert`

tommychiu-github:
> > Since we need to be able to store certs in flash in different formats, why not store them along with the header containing the object type. This way we could just store the entire wrapped object received from the host, and then when reading from the flash use the header to tell what kind of object it is.
> > Also, please add the storage format to the patch description or to code comments.
> 
> This sounds like a good idea to me as well (will simplify your future work @tommychiu-github), in such case, I take it the idea would be to add the following types to this [enum](https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/manuf/base/perso_tlv_data.h;drc=32c42ca99db21abd0e2482ae1e681a0dac2616e0;l=28) (in a future PR that adds CWT cert formats)?
> 
> * `kPersoObjectTypeCwtTbs`
> * `kPersoObjectTypeCwtCert`

Yep, this makes sense to me as well. I'll push an update to include "Object Header" as well.
And for the 2 more enums for CWT, I'll put in another PR since it's more related to CWT implementation.
The storage format is in #24942 already, but I can copy it here as well.

sw/device/silicon_creator/manuf/base/perso_tlv_data.c
```diff
@@ -63,29 +57,27 @@ status_t perso_tlv_set_cert_block(const uint8_t *buf, size_t max_room,
...
-    return INTERNAL();
-
-  uint16_t wire_cert_size =
-      wrapped_cert_size - sizeof(perso_tlv_cert_header_t) - name_len;
-
-  if (cert_size != wire_cert_size) {
-    LOG_ERROR("Unexpected cert size %d instead of %d for cert %s", cert_size,
-              wire_cert_size, block->name);
-    return INTERNAL();
+  if (obj_type == kPersoObjectTypeX509Tbs ||
```
timothytrippel, line 67: I don't think you need to do the check twice. You can just do it when you hash the certs.

sw/device/silicon_creator/manuf/base/perso_tlv_data.c
```diff
@@ -27,33 +27,27 @@ status_t perso_tlv_set_cert_block(const uint8_t *buf, size_t max_room,
                                   perso_tlv_cert_block_t *block) {
   perso_tlv_object_header_t objh;
   uint16_t obj_size;
-  perso_tlv_object_header_t obj_type;
+  perso_tlv_object_type_t obj_type;
```
timothytrippel, line 30: nice find

timothytrippel (approved):

sw/device/silicon_creator/manuf/base/perso_tlv_data.c
tommychiu-github: Well I was thinking to validate the content wherever I change the parsing logic.
It's fine for me to remove the logic here.

timothytrippel: but when you hash the certs thats the final time a we are checking for correctness (after we have committed the certs to flash), so one check should be good enought right?

timothytrippel:
As discussed offline, I am working on updating this PR to fix how the ROM_EXT parses DICE certs from flash info to match this change. I will update this PR shortly.

timothytrippel:
@tommychiu-github when I was going over this to add the ROM_EXT update, I realized this could be improved quite a bit to reuse code and minimize the chance of errors. I refactored this in #24992, so I think we can close this. PTAL at #24992 when you have a moment.

PR #24958, opened 2024-10-31
[dvsim] Improve error handling on slurm launcher

Razer6:
Incorporate post-merge feedback from @matutem to improve the error handling and make try/catch block only deal with IO.

### DIFF
--- a/util/dvsim/SlurmLauncher.py
+++ b/util/dvsim/SlurmLauncher.py
@@ -5,6 +5,7 @@
 import logging as log
 import os
 import shlex
+import shutil
 import subprocess
 
 from Launcher import ErrorMessage, Launcher, LauncherError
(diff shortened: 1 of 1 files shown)


util/dvsim/SlurmLauncher.py
```diff
@@ -46,28 +47,30 @@ def _do_launch(self):
...
-
                 log.info(f'Executing slurm command: {slurm_cmd}')
                 self.process = subprocess.Popen(shlex.split(slurm_cmd),
                                                 bufsize=4096,
                                                 universal_newlines=True,
                                                 stdout=out_file,
                                                 stderr=out_file,
                                                 env=exports)
+        except IOError as e:
+            raise LauncherError(f'File Error: {e} when handling {self.slurm_log_file}')
```
rswarbrick, line 73: Nit: Maybe worth adding a '\n' here like the next error? I don't have experience of this error type but I *think* it's the same as OSError and you get more than a single word of text?

util/dvsim/SlurmLauncher.py
```diff
@@ -91,13 +94,18 @@ def poll(self):
...
+            try:
+                with open(self.slurm_log_file, 'r') as slurm_file:
+                    try:
+                        with open(self.deploy.get_log_path(), 'a') as out_file:
+                            shutil.copyfileobj(slurm_file, out_file)
+                    except IOError as e:
+                        raise LauncherError(f'File Error: {e} when handling '
+                                            f'{self.deploy.get_log_path()}')
+                # Remove the temporary file from the slurm process
+                os.remove(self.slurm_log_file)
```
rswarbrick, line 106: Isn't this essentially just moving the file? I don't think we need to open either copy, do we? Can we just use `shutil.move`?

util/dvsim/SlurmLauncher.py
```diff
@@ -46,28 +47,30 @@ def _do_launch(self):
 
         self._dump_env_vars(exports)
 
+        # Add a command delimiter if necessary
+        slurm_setup_cmd = SLURM_SETUP_CMD
+        if slurm_setup_cmd != '' and not slurm_setup_cmd.endswith(';'):
```
matutem, line 52: Nit: can just be "if slurm_setup_cmd and ...", since an empty string evaluates to False in a conditional.

matutem (approved):
I agree with @rswarbrick good point about shutil.move. Otherwise LGTM

util/dvsim/SlurmLauncher.py
Razer6: No it's not. It is appending the log files from all slurm jobs to a single log file.

Razer6:
Thank's for the review. I updated the two comments but not the shutil.move. The code is intentionally aggregating all slurm log files to a single log file.

util/dvsim/SlurmLauncher.py
rswarbrick: Ah, I see! Cool, that makes sense then :-)

rswarbrick: Oh, wait a sec! Does that mean we're doing something like:
```
cp acc a; rm a
cp acc b; rm b
cp acc c; rm c
```

Maybe your original code made sense (essentially doing a `cat`)?

Razer6: Yep, we are doing that. But shouldn't that work via shutil. copyfileobj()?

PR #24973, opened 2024-10-31
[ipgen,tpl,cmdgen] Remove earlgrey CMDGEN output from ipgen templates

matutem:
I verified the resulting generated md files are unchanged. This requires running make -C hw, followed by cmdgen.py for the affected generated files.


andreaskurth:
Thanks @matutem. I agree with the intention of this PR, but it doesn't seem to be aligned with what happens when `make -C hw` is run: In `hw/Makefile` (for the `top` target), topgen is run but cmdgen isn't.

Unfortunately it's not as simple as adding `${PRJ_DIR}/util/cmdgen.py -u hw/$*/**/*.md` there because the working directory then is wrong

andreaskurth:
> Unfortunately it's not as simple as adding `${PRJ_DIR}/util/cmdgen.py -u hw/$*/**/*.md` there because the working directory then is wrong

PR #24986 should solve this problem.

matutem:
I will file an issue and a separate PR for adding cmdgen in the Makefile target for top. I think it deserves better documentation since otherwise check-generated would fail. It is weird we didn't catch it for so long: a quick grep suggests we always have the top earlgrey text between brackets, which would break for multi-top, and perhaps we never checked this for englishbreakfast?  I'll loop you in for sure.

matutem:
Filed https://github.com/lowRISC/opentitan/issues/24991.

andreaskurth:
Force-pushed to rebase on `master` including PR https://github.com/lowRISC/opentitan/pull/24993

andreaskurth (approved):

pamaury:
Since this change, running topgen or make now seems to delete the top_earlgrey documentation which is problematic. Is that an unintentional consequence? Are we supposed to run another make command to restore the documentation?

hcallahan-lowrisc:
With this change, you would have to run `make -C hw top_and_cmdgen` to get back to the same clean repository state with cmdgen-rendered earlgrey documentation.
I think this API change was unintentional, I'll create a PR to change it back so the 'top' target gets us back to the clean state (unless anyone wants to advocate for an API change here for this development tool)

PR #24980, opened 2024-11-01
[rom_ext] encapsulate dice logics for moving to silicon_creator lib

### LINKED ISSUE #24368: [bazel,rom_ext,manuf] develop reference non-mutable ROM_EXT code section and tooling support
#23425 added a feature to the Earlgrey PROD ROM to support execution of a non-mutable ROM_EXT code section in the ROM before handing execution over to the ROM_EXT itself. To enable using this feature, we should develop:
1. a reference implementation of the most common use case for this feature (namely to perform UDS and CDI_0 key and certificate generation), and
2. Bazel build rules to enable building this code section as a separate CC binary and embedding it directly in the ROM_EXT binary
3. Bazel tooling to enable building an individualization binary with the proper code of hash of the expected non-mutable ROM_EXT to be deployed in a device's transport image

### LINKED ISSUE #25030: [rom_ext,dice] Private key is not cleared when the chain is not changed.
### Description

ROM_EXT will skip cert signing when the old cert in flash is valid for the current public key.

However, we managed the private key clear & save op in the cert signing library.
Skipping the call could result in signing with a wrong key and also key leakage.

To fix the issue, we’ll need to move the otbn_boot_attestation_key_save from dice lib to rom_ext, so the private key will always be overwritten with the current stage's key.

sasdf:
In order to have a better diff for code review, those changes are kept in rom_ext.c for now,
and we’ll move them to `silicon_creator/lib/cert/` in a follow-up PR.

---

This PR contains several bugfixes and refactoring of the DICE-related logics in rom_ext.

* Global variables are encapsulated in `dice_cert_t` and related code are encapsulated in `dice_chain_*` functions for moving.
* Includes more boundary checks on the cert buffer in RAM to avoid overflow.
* TLV object is parsed at where it will be consumed to handle parsing errors correctly.
* Stale TLV objects will be cleared if parsing fails.
* Certificate checking function are moved to dice library to handle different formats.
* TLV name will also be checked when validating the cert.
* Invalidate the remaining buffer before encoding the new TLV entry to ensure the alignment padding will also be cleared.
* Skip the first cert only if it’s UDS cert.
* Remove UDS cert check for simplicity, since rom_ext can't fix it even the cert is unmatched.
* Fixes #25030 by overwriting with the current stage's key.

This also partially addresses #24368.

### DIFF
--- a/sw/device/silicon_creator/rom_ext/rom_ext.c
+++ b/sw/device/silicon_creator/rom_ext/rom_ext.c
@@ -39,6 +39,7 @@
 #include "sw/device/silicon_creator/lib/drivers/rstmgr.h"
 #include "sw/device/silicon_creator/lib/drivers/uart.h"
 #include "sw/device/silicon_creator/lib/epmp_state.h"
+#include "sw/device/silicon_creator/lib/keymgr_binding_value.h"
 #include "sw/device/silicon_creator/lib/manifest.h"
 #include "sw/device/silicon_creator/lib/manifest_def.h"
 #include "sw/device/silicon_creator/lib/otbn_boot_services.h"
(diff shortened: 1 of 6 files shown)


sasdf:
Rebased, the PR is ready to review now.

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -99,31 +100,125 @@ const epmp_region_t kFlashRegion = {
...
+ * @return errors encountered during the operation.
+ */
+OT_WARN_UNUSED_RESULT
+rom_error_t dice_chain_attestation_silicon(void);
+
+/**
+ * Check the CDI_0 certificate and regenerate if invalid.
+ *
+ * @param attest_binding Pointer to the measurements to attest.
+ * @param owner_manifest Pointer to the current rom_ext manifest.
```
timothytrippel, line 197:
```suggestion
 * @param rom_ext_manifest Pointer to the current rom_ext manifest.
```

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -858,22 +1039,38 @@ static rom_error_t rom_ext_try_next_stage(boot_data_t *boot_data,
   return error;
 }
 
-static rom_error_t rom_ext_start(boot_data_t *boot_data, boot_log_t *boot_log) {
-  HARDENED_RETURN_IF_ERROR(rom_ext_init(boot_data));
-  const manifest_t *self = rom_ext_manifest();
-  dbg_printf("Starting ROM_EXT %u.%u\r\n", self->version_major,
-             self->version_minor);
+OT_WARN_UNUSED_RESULT
```
timothytrippel, line 1042: you dont need to mark OT_WARN_UNUSED_RESULT twice, once in the prototype is fine

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -317,61 +413,148 @@ static uintptr_t owner_vma_get(const manifest_t *manifest, uintptr_t lma_addr) {
           (uintptr_t)_owner_virtual_start_address + CHIP_ROM_EXT_SIZE_MAX);
 }
 
-static size_t dice_certs_buffer_space_remaining(void) {
-  return FLASH_CTRL_PARAM_BYTES_PER_PAGE -
-         (size_t)(dice_cert_obj.obj_p - dice_certs_page);
+// Get the size of the remaining tail space that is not processed yet.
+OT_WARN_UNUSED_RESULT
+static size_t dice_chain_get_tail_size(void) {
```
timothytrippel, line 418: could you group all the `dice_chain_*` functions together, and clearly comment which ones will be moved to a separate lib?

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -317,61 +413,148 @@ static uintptr_t owner_vma_get(const manifest_t *manifest, uintptr_t lma_addr) {
...
+      dice_chain.data));
+
+  // Resets the flash page status.
+  dice_chain.data_dirty = kHardenedBoolFalse;
+  dice_chain.tail_offset = 0;
+
+  return kErrorOk;
+}
+
+// Push the certfiicate to the tail with LTV header.
```
timothytrippel, line 532:
```suggestion
// Push the certificate to the tail with LTV header.
```

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -394,143 +577,141 @@ static rom_error_t rom_ext_attestation_silicon(void) {
...
+        "CDI_0", dice_chain.scratch_cert, updated_cert_size));
   } else {
-    HARDENED_RETURN_IF_ERROR(rom_ext_attestation_increment_cert_offset());
+    // Cert is valid, move to the next one.
+    HARDENED_CHECK_EQ(dice_chain.cert_valid, kHardenedBoolTrue);
+    dice_chain_next_cert_obj();
   }
+
+  // Replace UDS with CDI_0 key for endorsing next stage cert.
+  HARDENED_RETURN_IF_ERROR(otbn_boot_attestation_key_save(
```
timothytrippel, line 630: we only want to do this if the cert was valid, as there is already a call to `otbn_boot_attestation_key_save()` in `dice_cdi_0_cert_build()`

timothytrippel: same for CDI_1 below

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -394,143 +577,141 @@ static rom_error_t rom_ext_attestation_silicon(void) {
...
-        &dice_cert_obj));
-    HARDENED_RETURN_IF_ERROR(rom_ext_attestation_increment_cert_offset());
-    // If the CDI_1 cert is updated, it could be smaller than the previous CDI_1
-    // cert, so we make sure to clear out the rest of the buffer.
-    memset(dice_cert_obj.obj_p, 0, dice_certs_buffer_space_remaining());
+    size_t updated_cert_size = kScratchCertSizeBytes;
+    // TODO(#19596): add owner configuration block measurement to CDI_1 cert.
+    HARDENED_RETURN_IF_ERROR(
+        dice_cdi_1_cert_build((hmac_digest_t *)attest_binding->data,
+                              (hmac_digest_t *)zero_binding_value.data,
```
timothytrippel, line 663: The first parameter should be the BL0 (owner fw) measure as it was before: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/cert/dice.c;l=208?q=dice.c&ss=opentitan%2Fopentitan
The DiceTcbInfo extension of the cert contains both measurements that were combined to create the attest_binding measurement.

The second parameter can be 0 for now until we cherry pick this to `earlgrey_1.0.0` branch. On said branch the second parameter is the owner configuration block measurement, since that branch has the most up to date ROM_EXT with ownership transfer (https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/cert/dice.c;l=208?q=dice.c&ss=opentitan%2Fopentitan)

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -394,143 +577,141 @@ static rom_error_t rom_ext_attestation_silicon(void) {
...
+        "CDI_1", dice_chain.scratch_cert, updated_cert_size));
   } else {
-    HARDENED_RETURN_IF_ERROR(rom_ext_attestation_increment_cert_offset());
+    // Cert is valid, move to the next one.
+    HARDENED_CHECK_EQ(dice_chain.cert_valid, kHardenedBoolTrue);
+    dice_chain_next_cert_obj();
   }
 
+  // Replace CDI_0 with CDI_1 key for endorsing next stage cert.
+  HARDENED_RETURN_IF_ERROR(otbn_boot_attestation_key_save(
```
timothytrippel, line 676: we only want to do key save in the else branch above.

sw/device/silicon_creator/rom_ext/rom_ext.c
sasdf: Sure, renamed them to make it more clear.
The parameters are the same as before.

timothytrippel (approved):

timothytrippel:
@sasdf ever since merging this, it looks like every time the ROM_EXT runs, it is updating the DICE certificates. You can tell by running the following sequence of operations with your FPGA board:
1. `bazel run //sw/host/opentitantool -- --interface=hyper310 fpga clear-bitstream` (to clear the bitstream)
2. `bazel test --test_output=streamed --cache_test_results=no //sw/device/tests:aes_smoketest_fpga_cw310_sival_rom_ext` to run the AES smoketest at the owner stage after executing the ROM_EXT (this first execution of the ROM_EXT should trigger a cert update since CDI_0/1 are not provisioned since the FPGA has empty flash info tests
3. repeat step 2; you should NOT see the DICE CDI_* certs get updated again as they should be valid

This was the behavior before this PR was merged. Longer term, we need to get a test running in CI to detect this. In the interim, could you please investigate why this was broken? Thanks!

PR #24982, opened 2024-11-01
[rom_ctrl, dv] Uncovered conditional coverage for tlul_adapter_sram

KinzaQamar:
The nightly regression coverage dashboard thrown an uncovered conditional coverage hole on L138 and L131 in tlul_adapter_sram. To resolve the issue rom_ctrl_sec_cm test has been used. 

This PR fixes two kind of vulnerabilities.
1) The test rom_ctrl_sec_sm wasn't collecting the coverage for the wires that were uncovered by the report. Fixing this was straight forward to add the proxy binds in rom_ctrl_base_sim_cfg.hjson that were injecting faults to different blocks inside rom_ctrl and add the sec_cm_count test_plan in rom_ctrl_testplan.hjson. 
2) As the test inject faults to different blocks of rom_ctrl(FSM, SRAM, ...), the testbench should know what to do next if something bad has happened. For this particular coverage hole, the proxy_if was prim_count that pokes cnt_q in prim_count.sv and that generates an error. If the fifo was empty, that will lead alot of signals become 'X and the fifo
error being stuck at 'X. To avoid this, we splat the backing sram. Then there were something assertions in prim_fifo_sync and rom_ctrl that fails and needs to switched off during fault injection. Last to compare the respective fatal_alert_cause registers.

### DIFF
--- a/hw/ip/rom_ctrl/data/rom_ctrl_testplan.hjson
+++ b/hw/ip/rom_ctrl/data/rom_ctrl_testplan.hjson
@@ -10,6 +10,7 @@
                      "hw/dv/tools/dvsim/testplans/tl_device_access_types_testplan.hjson",
                      "hw/dv/tools/dvsim/testplans/sec_cm_fsm_testplan.hjson",
                      "hw/dv/tools/dvsim/testplans/stress_all_with_reset_testplan.hjson",
+                     "hw/dv/tools/dvsim/testplans/sec_cm_count_testplan.hjson",
                      "rom_ctrl_sec_cm_testplan.hjson"]
   testpoints: [
     {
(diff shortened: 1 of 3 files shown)


rswarbrick:
Hi @KinzaQamar! Thisk looks sensible but is currently failing lint because of the commits. I'd probably suggest just squashing the three together (and possibly tweaking the commit message afterwards). I realise that there are still some other holes, but maybe it makes sense to get something merged "on the way".

rswarbrick (commented):
A minor style question, but this looks good to me.

rswarbrick (approved):
This looks good to me. Happy to merge once CI has finished, but it looks like it's being slow. Give me a shout tomorrow if I've forgotten :-)

PR #24986, opened 2024-11-01
[util/cmdgen.py] Make paths in commands absolute

andreaskurth:
Commands may contain paths, which are relative to the root of the repository.  In order to make command invocation independent of the current working directory, we need to make these paths absolute.

### DIFF
--- a/util/cmdgen.py
+++ b/util/cmdgen.py
@@ -74,6 +74,23 @@ def cmdgen_rewrite_md(filepath: Path, dry_run: bool, update: bool) -> bool:
         match_start_linum = 1 + content[0:match_start.start(0)].count("\n")
         cmd = match_start.group('cmd').strip()
 
+        # The command may contain paths, which are relative to the root of the
+        # repository. In order to make command invocation independent of the
+        # current working directory, we need to make these paths absolute.
+        cmd_parts = cmd.split(' ')
+        for idx, part in enumerate(cmd_parts):
+            # If the command starts with a `#`, don't do anything because that
+            # command is to be treated as commented out.
+            if idx == 0 and part.startswith('#'):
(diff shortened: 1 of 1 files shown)


util/cmdgen.py
```diff
@@ -74,6 +74,19 @@ def cmdgen_rewrite_md(filepath: Path, dry_run: bool, update: bool) -> bool:
...
         cmd = match_start.group('cmd').strip()
 
+        # The command may contain paths, which are relative to the root of the
+        # repository. In order to make command invocation independent of the
+        # current working directory, we need to make these paths absolute.
+        cmd_parts = cmd.split(' ')
+        for idx, part in enumerate(cmd_parts):
+            # The first part (index 0) of a command is a path to an executable;
+            # all other parts are treated as paths if they contain a slash.
+            if idx == 0 or '/' in part:
```
andreaskurth, line 84: `'/i in part` is somewhat of a poor man's path detection, I think. If you guys have a better idea on reliably determining if a string is or could be a path, please suggest it.

rswarbrick (commented):
I'm not sure I understand this. At the moment, the examples in the documents look like this:
```
<!-- BEGIN CMDGEN util/regtool.py -d ./hw/ip/adc_ctrl/data/adc_ctrl.hjson -->
```
where all paths are relative to the root of the repository.

Rather than hacking on the paths, can't we run the command subprocess from the root of the repository? I think it would probably be simpler?

andreaskurth:
> Rather than hacking on the paths, can't we run the command subprocess from the root of the repository? I think it would probably be simpler?

Duh, indeed! Thanks for your suggestion -- I'll close this PR and create a new one to implement it!

PR #24987, opened 2024-11-01
[spi_device/dv] Remove unused variables

### LINKED ISSUE #20193: [spi_device] Improvements/fixes to SPI device constants
### Description

While reading the documentation and the code of the SPI device IP, I've found the following potential issues:

1. In `hw/ip/spi_device/dv/env/spi_device_env_pkg.sv`:
    ```sv
    parameter uint CMD_FIFO_SIZE                   = 32;
    parameter uint ADDR_FIFO_SIZE                  = 32;
    ```
    32 does not seem to match any units used elsewhere. Both bytes (64) and _depth_, _i.e._ 32-bit words (16) are used but `32` does not seem valid.

2. In `hw/ip/spi_device/rtl/spi_device_reg_pkg.sv`:
    ```sv
    parameter int unsigned        SPI_DEVICE_INGRESS_BUFFER_SIZE   = 'h 1a0;
    ```
    however in `hw/ip/spi_device/dv/env/spi_device_env_pkg.sv`
    ```sv
    parameter uint SRAM_INGRESS_SIZE               = 256 + 64 + 64; // 96 depth
    ```
     that is 384 = 0x180 not 0x1a0

3. In `hw/ip/spi_device/rtl/spi_device_pkg.sv`
    ```sv
    SramCmdFifoDepth = 16; // 16 x 1B for Cmd
    SramAddrFifoDepth = 16; // 16 x 4B for Addr
    SramTotalDepth = SramMsgDepth + SramMailboxDepth
               + SramSfdpDepth + SramPayloadDepth
               + SramCmdFifoDepth + SramAddrFifoDepth ;
    ```
     At first glance, it seems weird to add `SramCmdFifoDepth` and `SramAddrFifoDepth`, given the comments. The comment actually means that 1-byte Cmd is first extended to a 4-byte value before being added to the 32-bit command FIFO, which is itself only reflected as a 8-bit value in [UPLOAD_CMDFIFO](https://opentitan.org/book/hw/ip/spi_device/doc/registers.html#upload_cmdfifo) register.

andreaskurth:
See https://github.com/lowRISC/opentitan/issues/20193#issuecomment-2452097088. This resolves #20193.

### DIFF
--- a/hw/ip/spi_device/dv/env/spi_device_env_pkg.sv
+++ b/hw/ip/spi_device/dv/env/spi_device_env_pkg.sv
@@ -93,11 +93,6 @@ package spi_device_env_pkg;
   parameter uint SFDP_SIZE                       = 256;
   parameter uint PAYLOAD_FIFO_START_ADDR         = SFDP_START_ADDR + SFDP_SIZE; // 0xd00
   parameter uint PAYLOAD_FIFO_SIZE               = 256;
-  // CMD_FIFO_START_ADDR is 0xe00
-  parameter uint CMD_FIFO_START_ADDR             = PAYLOAD_FIFO_START_ADDR + PAYLOAD_FIFO_SIZE;
-  parameter uint CMD_FIFO_SIZE                   = 32;
-  parameter uint ADDR_FIFO_START_ADDR            = CMD_FIFO_START_ADDR + CMD_FIFO_SIZE; // 0xe20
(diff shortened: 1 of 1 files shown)


antmarzam (approved):
Thank you for the cleanup Andreas! 
LGTM as these params aren't used anywhere

rswarbrick (approved):

PR #24989, opened 2024-11-01
[util/cmdgen.py] Ensure working directory is repo root

andreaskurth:
Commands may contain paths, which are relative to the root of the repository.  In order to make command invocation independent of the current working directory, we set `cwd` of `subprocess.run` to the root of the repository.

This supersedes PR #24986.


matutem (approved):
LGTM

PR #24993, opened 2024-11-02
[ci,check-generated] Handle empty text in CMDGEN

### LINKED ISSUE #24991: [doc,cmdgen] Remove need for text bracketed by CMDGEN BEGIN END
### Description

The util/cmdgen.py tool runs a command to generate text that will be inserted between the line containing CMDGEN BEGIN and the one containing CMDGEN END. This means there is no need to have any text between these lines.

The usage of cmdgen is described at https://opentitan.org/book/doc/contributing/doc/index.html#cmdgen, and is currently done manually as it is not part of the make targets under hw/Makefile which generates a top.

Since we place all generated artifacts in git we need to make sure cmdgen is run so all md files are as expected.

The problem we face is that if the bracketed text is empty the ci/scripts/check-generated.sh will fail since it uses hw/Makefile to generate artifacts and there is no target for cmdgen. Apparently we currently deal with this by populating the lines between the CMDGEN brackets with what earlgrey expects, which is clumsy, and is unworkable in multi-top.

I propose we add a cmdgen target in that makefile and enhance the check-generated script to invoke that target as well.

matutem:
Change ci/scripts/check-generated.sh to run both top and cmdgen when checking for the topgen generated artifacts. This is to support template md files having no text between the BEGIN and END markers so they can handle multi-top.

Add a separate "cmdgen" target in the hw/Makefile and describe that in the README.md for contributing to documentation.

Fixes #24991

### DIFF
--- a/ci/scripts/check-generated.sh
+++ b/ci/scripts/check-generated.sh
@@ -41,7 +41,8 @@ gen_and_check_clean() {
 
 # A specialized version of gen_and_check_clean for targets of the Makefile in hw
 gen_hw_and_check_clean() {
-    gen_and_check_clean "$1" make -k -C hw "$2"
+    thing=$1; shift
+    gen_and_check_clean "$thing" make -k -C hw "$*"
 }
 
 # Check generated files are up to date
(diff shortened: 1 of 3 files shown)


matutem:
There is a CI failure from shellcheck because it wants $* to be within quotes in check-generated.sh. However, if it is quoted the "make" command will get a string with "top cmdgen", and that target is not found. I want to run make so it executes both targets and this is the simplest way I can think of. Perhaps someone can think of a simple alternative?

I know @andreaskurth suggested running cmdgen in the tops_gen target, but I find that unappealing.

andreaskurth (approved):
LGTM, thx @matutem!

jwnrt (approved):

HU90m (approved):

PR #25000, opened 2024-11-04
[pentest] Add RSA512 decryption SCA test

nasahlpa:
This commit adds the otbn.sca.rsa512_decrypt penetration test. The user needs to provide the mod, exp, and message which then gets decrypted using the rsa OTBN app. The result is transmitted back to the host. SCA measurements can be conducted during the trigger window.


nasahlpa:
@johannheyszl and @vogelpi could I also get an approval here, please? :)

vogelpi (approved):
LGTM, thanks @nasahlpa !

PR #25001, opened 2024-11-04
[ci] Move Verilated Earl Grey tests to GitHub Actions

jwnrt:
Fairly straight-forward move.


nbdd0121 (approved):
As a follow-up PR, we should address the Bazel warning and reduce parallelism to number of cores on the runner.

nbdd0121:
The added job passed the CI, merging.

PR #25002, opened 2024-11-04
[ci] Port software build job from Azure to GitHub

jwnrt:
Moves both the `sw_build` and `sw_test` jobs from Azure into one `sw_build_test` job in GitHub Actions.

There's a small improvement in execution time here for a cold build. I'll check the time when the results are cached once this merges.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,8 +242,8 @@ jobs:
         with:
           context: site/redirector/landing
 
-  sw_build:
-    name: Build all software
+  sw_build_test:
+    name: Build and test software
     runs-on: ubuntu-22.04-vivado
     timeout-minutes: 120
     needs: quick_lint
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -241,3 +241,72 @@ jobs:
...
+
+  sw_build_test:
+    name: Build and test software
+    runs-on: ubuntu-22.04-vivado
+    timeout-minutes: 120
+    needs: quick_lint
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Required so we can lint commit messages.
```
nbdd0121, line 253: Do we still need this for sw build & test?

jwnrt: The comment is wrong, but we do need this for the bitstream cache to work as well. I'm going to remove this and see if this job does any splicing.

.github/workflows/ci.yml
```diff
@@ -241,3 +241,72 @@ jobs:
...
+            --test_output=errors                                      \
+            --define DISABLE_VERILATOR_BUILD=true                     \
+            --test_tag_filters=-broken,-cw310,-verilator,-dv,-silicon \
+            --target_pattern_file="$target_pattern_file"
+      - name: Publish test results
+        uses: actions/upload-artifact@v4
+        with:
+          name: test_results
+          path: "**/test.xml"
+          overwrite: true
```
nbdd0121, line 309: This is what I currently have for FPGA jobs:

```yaml
- name: Publish Bazel test results
  uses: mikepenz/action-junit-report@ec3a351c13e080dc4fa94c49ab7ad5bf778a9668 # v5
  if: ${{ !cancelled() }}
  with:
    report_paths: 'bazel-out/**/test.xml'
```

Perhaps we can have a composite action that merge junit, upload the xml and publish junit report?

jwnrt: Yeah, that would be useful

nbdd0121 (approved):
Failing test seems unrelated

PR #25004, opened 2024-11-04
[topgen] Support for unmanaged external clocks

Razer6:
Tops may specify clocks that are completely managed externally, i.e., with a dedicated clock manager in the uncore environment of an integrated OpenTitan. Feeding that clock manager to the internal clock manager is not necessary and just adds complexity in the configuration.

To address this, this PR introduces support for unmanaged external clocks, allowing them to operate independently from the internal clock manager. To define an unmanaged clock, simply add an entry to the top-level unmanaged_clocks list. Once defined, these clocks can be used like any other clock for IP instantiation and crossbar mapping.

Additionally, when routing IP alerts to the alert manager, the alert manager requires a clock gating status signal to know if the clock source for the alert’s IP is currently gated. For unmanaged external alerts, it's assumed that the top-level external environment already has knowledge of the clock’s gating status. Topgen creates a dedicated LPG for every external clock. The clock gating status signal is routed at top-level and connected to the associated group to the alert manager.


hw/top_earlgrey/data/autogen/top_earlgrey.gen.hjson
andreaskurth, line 1: @Razer6 as discussed out-of-band, this generally LGTM, but `alert_lpgs` gets skewed during the generation of this file

util/topgen/merge.py
```diff
@@ -580,6 +580,14 @@ def _get_clock_group_name(clk: Union[str, OrderedDict],
     return group_name, src_name
 
 
+def is_unmanaged_clock(top: OrderedDict, clock: str):
+    clk = top['unmanaged_clocks']._asdict().get(clock)
```
matutem, line 584: The body of this function could be one line as `return clock in top['unmanaged_clocks']._asdict()`

util/topgen/merge.py
```diff
@@ -841,19 +853,27 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
         # 1) figure out the clock group assignment of the primary clock
         # Get the full clock name and split the hierarchy path, getting the last element
         clk = module['clock_connections'][block_clock.clock]
-        clk = clk.split(".")[-1]
-
-        # Discover what clock group we are related to
-        clock_group = clock_groups[clk]
+        # Unmanaged clocks are not part of the LPGs. Unmanaged clocks have the input signal
+        # identifier ('_i') directly in the signal name
+        unmanaged_clock = True if clk.endswith('_i') else False
```
matutem, line 858: I think it would be better to use the info in top['unmanaged_clks'].

util/topgen/merge.py
```diff
@@ -863,12 +883,16 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
             # indication signals, we can just use the clock / reset signals
             # of the first block that belongs to a new unique LPG.
             clock = module['clock_connections'][block_clock.clock]
-            top['alert_lpgs'].append({
+            lpg = {
                 'name': lpg_name,
-                'clock_group': clock_group,
```
matutem, line 868: Why was clock_group removed from this dict? That changes the logic for non-external clocks.

util/topgen/clocks.py
```diff
@@ -202,6 +214,18 @@ def hint_names(self) -> Dict[str, str]:
         return hint_names
 
 
+class UnmanagedClocks:
+    '''Unmanaged clock connections for the chip.'''
+
+    def __init__(self, raw: List[object]):
+        self.clks = {}
```
matutem, line 221: What about `self.clks = {clk['name']: Unmanaged(clk) for clk in raw}`?

hw/top_earlgrey/data/autogen/top_earlgrey.gen.hjson
matutem: Probably something broke in create_alert_lpgs.

Razer6: There is no functional diff in that generation. It re-shuffles the entries in the dict: So for example, the first entry became from:
```
{
      name: peri_lc_io_div4_0
      clock_group:
      {
        name: peri
        src: top
        sw_cg: yes
        unique: no
        clocks:
        {
          clk_io_div4_peri: io_div4
          clk_io_div2_peri: io_div2
          clk_io_peri: io
          clk_usb_peri: usb
          clk_aon_peri: aon
        }
      }
      clock_connection: clkmgr_aon_clocks.clk_io_div4_peri
      reset_connection:
      {
        name: lc_io_div4
        domain: "0"
      }
    }
    ```
    
    This:
    ```
    {
      name: peri_lc_io_div4_0
      clock_connection: clkmgr_aon_clocks.clk_io_div4_peri
      unmanaged_clock: false
      reset_connection:
      {
        name: lc_io_div4
        domain: "0"
      }
      clock_group:
      {
        name: peri
        src: top
        sw_cg: yes
        unique: no
        clocks:
        {
          clk_io_div4_peri: io_div4
          clk_io_div2_peri: io_div2
          clk_io_peri: io
          clk_usb_peri: usb
          clk_aon_peri: aon
        }
      }
    }
    ```

Razer6: Adding `unmanaged_clock: false` to that dict, shuffled it back to the original order

Razer6:
Thanks @matutem for the review. I think all of the comments are addressed and the diff in the generated HJSON is also removed due to the re-shuffeling.

andreaskurth:

No functional change

andreaskurth (approved):

vogelpi (approved):

vogelpi:

This PR touches the hjson to add a topgen feature required for Integrated but not used in Earlgrey. The PR doesn't lead to modified RTL for Earlgrey. This is fine.

PR #24966, opened 2024-10-31
[hmac,doc] Update Save and Restore SW guide

martin-velay:
- After this issue #24767 has been raised, and a workaround has been found #24944, the Programmer's Guide should be updated to reflect this.

PR #25012, opened 2024-11-05
[hmac,sw] Poll HMAC IDLE state

martin-velay:
- Add HMAC idle polling to be sure that the HW is ready after the temporary workaround for context switching.
- To complete https://github.com/lowRISC/opentitan/pull/24944 as already merged.
- This was missing and has been caught after updating the programmer's guide (see https://github.com/lowRISC/opentitan/pull/24966)

### DIFF
--- a/sw/device/lib/crypto/drivers/hmac.c
+++ b/sw/device/lib/crypto/drivers/hmac.c
@@ -280,7 +280,7 @@ static void msg_fifo_write(const uint8_t *message, size_t message_len) {
  *
  * @param[out] ctx Context to which values are written.
  */
-static void tmp_avoid_hw_hang(hmac_ctx_t *ctx) {
+static status_t tmp_avoid_hw_hang(hmac_ctx_t *ctx) {
   // Insert delay which should be equivalent to at least 80 clock cycles
   ibex_timeout_t timeout;
   timeout.cycles = kNumIterTimeout / 2;  // 1/2 as a loop is 2 instructions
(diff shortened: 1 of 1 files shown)


sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -458,6 +458,7 @@ status_t hmac_update(hmac_ctx_t *ctx, const uint8_t *data, size_t len) {
    */
 
   tmp_avoid_hw_hang(ctx);
+  HARDENED_TRY(hmac_idle_wait());
```
vogelpi, line 461: Would it be better to move this into the `tmp_avoid_hw_hang()`?

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -280,7 +280,7 @@ static void msg_fifo_write(const uint8_t *message, size_t message_len) {
  *
  * @param[out] ctx Context to which values are written.
  */
-static void tmp_avoid_hw_hang(hmac_ctx_t *ctx) {
+status_t tmp_avoid_hw_hang(hmac_ctx_t *ctx) {
```
engdoreis, line 283: Why the static was removed, do you plan to use this function outside this file?

sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -297,6 +297,11 @@ static void tmp_avoid_hw_hang(hmac_ctx_t *ctx) {
   uint32_t cmd_reg = abs_mmio_read32(kHmacBaseAddr + HMAC_CMD_REG_OFFSET);
   cmd_reg = bitfield_bit32_write(cmd_reg, HMAC_CMD_HASH_PROCESS_BIT, true);
   abs_mmio_write32(kHmacBaseAddr + HMAC_CMD_REG_OFFSET, cmd_reg);
+
+  // Wait for HMAC HWIP operation to be completed.
+  HARDENED_TRY(hmac_idle_wait());
+
+  return OTCRYPTO_OK;
```
engdoreis, line 304:
```suggestion
  return hmac_idle_wait();
```

sw/device/lib/crypto/drivers/hmac.c
martin-velay: No, I don't. Thanks.

sw/device/lib/crypto/drivers/hmac.c
martin-velay: That's better indeed.

engdoreis (approved):
Thanks for the fixes.

Issue #22710, opened 2024-04-18
Clean up pwrmgr test bench to drive external reset request


neeraj-rv:
This just feeds the internal reset request back into the DUT as external reset request, right? This may work to trick some of the existing tests, but it doesn't allow to verify the wait-for-external-reset feature, and a quick simulation shows failures in the following tests:
- pwrmgr_reset
- pwrmgr_stress_all
- pwrmgr_sec_cm_lc_ctrl_intersig_mubi

which don't fail in the latest nightly regression.

So we need to create an issue to fix this and properly verify it.

_Originally posted by @andreaskurth in https://github.com/lowRISC/opentitan/pull/22575#discussion_r1570803707_

neeraj-rv:
https://github.com/lowRISC/opentitan/pull/22575#discussion_r1571177536

neeraj-rv:
A loop back path from internal request to external SoC reset request was added in the testbench as a temporary fix. 
A more elaborate TB fix is needed to 

1. make sure that TB can control the timing of the generated external reset request after it sees an internal reset request
2. TB would also need to test scenarios where the external reset request is driven asynchronously by the SoC, with a pending internal reset request in flight.
3. Help with coverage closure

PR #25013, opened 2024-11-05
[pwrmgr] Take external SoC reset into account in FSM

Razer6:
1. Internally generated reset request passed to SoC.
2. Create a sticky assertion internally to log request. Sticky flag cleared only upon external reset deassertion.
3. SoC transfers this request to platform to assert external reset.
4. External SoC reset deasserts after some time.
5. Pwrmgr stays in FastPwrStateRstWait state until the external SoC reset deasserts.
6. Boot / reset sequence continues.
7. Also fixes the external reset signal glitch filter to use the por reset. This is required to make sure that the external reset state does not self-reset once asserted and faithfully forwards that state to pwrmgr.
8. Added some debug visibility signals for pwrmgr.
9. Added temporary reset loop back path in pwrmgr TB to translate internal reset req from OT --> ext reset from SoC to OT.  Needs a proper fix in testbench (issue #22710).


Razer6:
@matutem I ipgen'ed the change. Could you take a look into this? I currently rely on the default value and not populating a top-specific value in topgen.

util/ipgen/renderer.py
```diff
@@ -67,6 +67,11 @@ def get_template_parameter_values(
                         val_typed = int(val, 0)
                     else:
                         val_typed = val
+                elif template_param.param_type == 'bool':
```
matutem, line 70: The # type in line 64 should add bool to the options in the Union. Better yet would be to make the type part of the declaration instead of a comment.

Razer6: I added the type in line 64. The more holistic approach though looks more intrusve.

hw/ip_templates/pwrmgr/pwrmgr_pkg.core.tpl
```diff
@@ -11,6 +11,9 @@ filesets:
   files_rtl:
     depend:
       - ${instance_vlnv("lowrisc:ip:pwrmgr_reg")}
+% if wait_for_external_reset:
+      - lowrisc:ip:rom_ctrl_pkg
```
matutem, line 15: What triggers this dependency?

Razer6: I was thinking the `boot_status_o` output had some dependency, but that was wrong. I removed that dependency.

hw/ip_templates/pwrmgr/dv/tb.sv.tpl
```diff
@@ -47,6 +50,9 @@ module tb;
   );
 
   assign interrupts[0] = pwrmgr_if.intr_wakeup;
+% if wait_for_external_reset:
+  assign int_reset_req = tb.dut.internal_reset_req;
```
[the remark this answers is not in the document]
Razer6: This was just a hack to get that feature partially tested via that loopback path. There is #22710 to properly verify that feature from the DV side.

hw/ip_templates/pwrmgr/data/pwrmgr.hjson.tpl
andreaskurth, line 1: We should document the _external reset_ feature in the Theory of Operation of `pwrmgr`

andreaskurth (commented):
LGTM overall, thx @Razer6. There are a few points we should clarify before merging this, though

hw/ip_templates/pwrmgr/rtl/pwrmgr.sv
```diff
@@ -509,6 +516,21 @@ module pwrmgr
...
+  // TODO(#22711): Make this work also when `rstreqs` is structured differently.
+  assign internal_reset_req            =|(
+                                         slow_peri_reqs.rstreqs &
+                                         {{NumSwRstReq{1'b1}},      // SW driven reset
+                                          {NumDebugRstReqs{1'b1}},  // debugger reset
+                                          {NumIntRstReqs{1'b1}},    // {ESC reset, slow_fsm}
+                                          // exclude the external async reset
+                                          {1'b0, slow_reset_en[0]}
+                                         }
+                                        );
```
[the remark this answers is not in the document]
Razer6: No, this is gated and will only be active for Darjeeling. No impact on other tops.

hw/ip_templates/pwrmgr/pwrmgr_pkg.core.tpl
Razer6: It's `rom_ctrl_pkg::pwrmgr_data_t` in the boot status.

hw/ip_templates/pwrmgr/data/pwrmgr.hjson.tpl
Razer6: Created https://github.com/lowRISC/opentitan/issues/25086 to track that.

hw/ip_templates/pwrmgr/rtl/pwrmgr.sv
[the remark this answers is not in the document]
Razer6: So far yes. If changes are needed, we add them later on.

andreaskurth:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr_fsm.sv

These changes have no functional impact

vogelpi:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr_fsm.sv

This PR changes RTL source code for Earl Grey but the changes have no functional impact (a new output is added which is not connected in Earl Grey). This is fine.

util/ipgen/renderer.py
matutem: For future reference, I just checked you could add a variable declaration with type only above line 63.

hw/ip_templates/pwrmgr/pwrmgr_pkg.core.tpl
matutem: So this dependency is still needed, right?

matutem (approved):
I consider this work in progress. I think pwrmgr can be templetized deeper than this.
LGTM

PR #25014, opened 2024-11-05
[hw,otp_ctrl_darjeeling] Hard fork of OTP Ctrl from integrated_dev

Razer6:
To unblock the work for integrated, we decided to create a temporary hard fork of `otp_ctrl` based on the `integrated_dev` branch. This fork only lives until `otp_ctrl` is `ipgen`'ed and will be removed later on.

To avoid package conflicts, all fusesoc packages are renamed to `otp_ctrl_darjeeling(_XYZ)`. The CIP_ID remains the same for the time being.


matutem:
A better fix should be possible when this is ipgen-ed. See https://github.com/lowRISC/opentitan/issues/25019.

matutem:
@Razer6, this is probably okay, but we should make sure the CI failures are understood and addressed if needed.

andreaskurth:
> A better fix should be possible when this is ipgen-ed. See #25019.

I agree, I think this is the long-term solution we all want. At the same time, a "hard fork" of otp_ctrl for Darjeeling as implemented in this PR can unblock work, so it's a viable stepping stone.

> @Razer6, this is probably okay, but we should make sure the CI failures are understood and addressed if needed.

Agreed; probably some CI failures (hopefully most) can be resolved with a rebase on `master`. I'm taking the liberty to do so.

Razer6:
Closing in favor of the ipgen'ed  otp_ctrl from https://github.com/notifications

PR #25022, opened 2024-11-06
[doc,design] Mention the need to run cmdgen

matutem:
When generating top and top-dependent IPs it is necessary to run cmdgen.
Add a mention of this in doc/contributing/doc/README.md and
doc/contributing/hw/design.md.

### DIFF
--- a/doc/contributing/doc/README.md
+++ b/doc/contributing/doc/README.md
@@ -33,7 +33,7 @@ Commonly one will run the following, which will go through every markdown file i
 ./util/cmdgen.py -u '**/*.md'
 ```
 
-If run, the following snippet will have the register tables of the AES block (output by `util/regtool.py -d ./hw/ip/aes/data/aes.hjson`) inserted into it.
+For example, when cmdgen is run with the '-u' flag, the following snippet will result in the insertion of the AES register tables between the BEGIN and END lines.
 
 ````md
 # Registers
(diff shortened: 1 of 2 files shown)


doc/contributing/hw/design.md
```diff
@@ -107,9 +107,24 @@ The sections below show how to do items 2-4.
...
+These files are generated using cmdgen via make.
+
+```console
+$ make -C hw cmdgen
+```
+
+Generating both top and cmdgen can be done conveniently with
+
+```console
+$ make -C hw top_and_cmdgen
```
rswarbrick, line 125: How odd! Can't we just give `make -C hw top cmdgen` as the example (and get rid of the target in the Makefile)?

doc/contributing/doc/README.md
```diff
@@ -33,7 +33,7 @@ Commonly one will run the following, which will go through every markdown file i
 ./util/cmdgen.py -u '**/*.md'
 ```
 
-If run, the following snippet will have the register tables of the AES block (output by `util/regtool.py -d ./hw/ip/aes/data/aes.hjson`) inserted into it.
+When run, the following snippet will have the register tables of the AES block (output by `util/regtool.py -d ./hw/ip/aes/data/aes.hjson`) inserted into it.
```
rswarbrick, line 36: Since we're touching this line: the text is a bit silly here. It looks like the snippet is run! Probably "When that command is run, the following snippet...." ?

doc/contributing/hw/design.md
matutem: That was my first attempt, but unfortunately the ci/scripts/check-generated.sh works by running a single target at a time but only running both these targets can we recreate all generated artifacts.

rswarbrick: Boo :-)

rswarbrick (approved):

PR #25025, opened 2024-11-06
[hmac,dv] Adpot extern method declarations

martin-velay:
Edit all the HMAC DV files, to move tasks, functions and constraints and adopt the "extern" SV feature to enhance readability.


rswarbrick (approved):
I think this is from the output of your script, right? This looks sensible to me and I think we can be confident that (as it compiles!) it won't introduce any issues. Thanks.

martin-velay:
> I think this is from the output of your script, right? This looks sensible to me and I think we can be confident that (as it compiles!) it won't introduce any issues. Thanks.

And I have also ran a `all_once` regression, and it passes

PR #25029, opened 2024-11-07
[manuf,rom_ext] move UDS cert to separate info page

timothytrippel:
The DICE UDS certificate is endorsed offline at manufacturing time, and injected into a flash info page on the chip. It must remain stable for the lifetime of a chip.

Previously, the CDI_0 and CDI_1 DICE certificates were also written to the same flash info page. However, the ROM_EXT can update these certificates in the event that the ROM_EXT or Owner FW is ever updated (as this triggers a CDI0/1 key pair change). To update these certificates, the ROM_EXT buffered all certificates in RAM, erased the flash info page, and re-wrote it with the udpated certificates.

If there is a power outage during the certificate update, after the flash page was erased but before the certificates had been written back, there could be a moment when the UDS certificate was lost.

To prevent this from happening, the UDS certificate was moved to its own flash info page, that is never erased for the lifetime of the chip. Additionally, this updatest the ROM_EXT to never write a dummy UDS cert to flash in the event a chip is not provisioned, or we are running on a sim/FPGA platform, and instead, just write a warning message to the console.

### DIFF
--- a/sw/device/silicon_creator/lib/drivers/flash_ctrl.h
+++ b/sw/device/silicon_creator/lib/drivers/flash_ctrl.h
@@ -79,7 +79,7 @@ typedef enum flash_ctrl_partition {
   X(kFlashCtrlInfoPageOwnerReserved1,      0, 6) \
   X(kFlashCtrlInfoPageOwnerReserved2,      0, 7) \
   X(kFlashCtrlInfoPageOwnerReserved3,      0, 8) \
-  X(kFlashCtrlInfoPageCreatorReserved0,    0, 9) \
+  X(kFlashCtrlInfoPageFactoryCerts,        0, 9) \
   /**
    * Bank 1 information partition type 0 pages.
    */ \
@@ -87,7 +87,7 @@ typedef enum flash_ctrl_partition {
   X(kFlashCtrlInfoPageBootData1,           1, 1) \
   X(kFlashCtrlInfoPageOwnerSlot0,          1, 2) \
   X(kFlashCtrlInfoPageOwnerSlot1,          1, 3) \
-  X(kFlashCtrlInfoPageCreatorReserved1,    1, 4) \
+  X(kFlashCtrlInfoPageCreatorReserved0,    1, 4) \
   X(kFlashCtrlInfoPageOwnerReserved4,      1, 5) \
   X(kFlashCtrlInfoPageOwnerReserved5,      1, 6) \
   X(kFlashCtrlInfoPageOwnerReserved6,      1, 7) \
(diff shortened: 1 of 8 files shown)


moidx (approved):

sw/device/silicon_creator/manuf/base/perso_tlv_data.h
```diff
@@ -121,6 +121,10 @@ typedef struct perso_tlv_cert_obj {
    * LTV object size (in bytes).
    */
   size_t obj_size;
+  /**
+   * LTV object type.
```
moidx, line 125: nit: TLV object type.

moidx (approved):

sw/device/silicon_creator/manuf/base/perso_tlv_data.h
timothytrippel: Need to align comments across several files, will do in follow up if that is ok?

moidx: SGTM

sw/device/silicon_creator/lib/drivers/flash_ctrl.h
```diff
@@ -79,15 +79,15 @@ typedef enum flash_ctrl_partition {
...
+  X(kFlashCtrlInfoPageFactoryCerts,        0, 9) \
   /**
    * Bank 1 information partition type 0 pages.
    */ \
   X(kFlashCtrlInfoPageBootData0,           1, 0) \
   X(kFlashCtrlInfoPageBootData1,           1, 1) \
   X(kFlashCtrlInfoPageOwnerSlot0,          1, 2) \
   X(kFlashCtrlInfoPageOwnerSlot1,          1, 3) \
-  X(kFlashCtrlInfoPageCreatorReserved1,    1, 4) \
+  X(kFlashCtrlInfoPageCreatorReserved0,    1, 4) \
```
vbendeb, line 90: why does this have to be renamed?

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -117,11 +117,20 @@ static uint8_t all_certs[8192];
 static size_t cdi_0_offset;
 static size_t cdi_1_offset;
 static cert_flash_info_layout_t cert_flash_layout[] = {
+    {
+        // The DICE UDS cert is placed on this page since it must remain stable
+        // post manufacturing. This page should never be erased by ROM_EXT, nor
+        // owner firmware.
+        .used = true,
+        .group_name = "FACTORY",
```
vbendeb, line 125: could also be called DICE_UDS or some such

vbendeb (commented):
Could the problem be solved by saving UDS cert in a  different flash page before updating DICE certs and then copying it back?

sw/device/silicon_creator/manuf/base/ft_personalize.c
timothytrippel: I thought factory would be more encompasing in case we add other factory endorsed certs to the base perso firmware

sw/device/silicon_creator/lib/drivers/flash_ctrl.h
timothytrippel: Because `kFlashCtrlInfoPageCreatorReserved0` was consumed so we should start from 0 again.

timothytrippel:
> Could the problem be solved by saving UDS cert in a different flash page before updating DICE certs and then copying it back?

We thought about this. but then we still need a free page. So at that point, we might as well make the page just read only for the lifetime of the chip. it is simpler and more efficient.

vbendeb:
> We thought about this. but then we still need a free page. So at that point, we might as well make the page just read only for the lifetime of the chip. it is simpler and more efficient.

so, instead of having a spare flash page  (any page, INFO of regular) and reuse it in different cases we dedicate a separate INFO page (a more expensive resource) permanently, this is hardly more efficient..

PR #25035, opened 2024-11-07
[kmac, rtl/dv/sw] Remove unused error code

nasahlpa:
The `ErrShadowRegUpdate` code is never used in the RTL. Instead, according to the documentation, on a shadow register update error, a recoverable alert shall be triggered - this recoverable alert then is signaled in the STATUS register and not in the error code register.

Hence, this commit removes this error code as well as the corresponding SW code (`kDifErrorShadowRegisterUpdate`).

### DIFF
--- a/hw/ip/kmac/rtl/kmac_pkg.sv
+++ b/hw/ip/kmac/rtl/kmac_pkg.sv
@@ -456,9 +456,6 @@ package kmac_pkg;
     //  - Sw issues KMAC op without Entropy setting.
     ErrSwHashingWithoutEntropyReady = 8'h 09,
 
-    // Error Shadow register update
-    ErrShadowRegUpdate = 8'h C0,
-
     // Error due to lc_escalation_en_i or fatal fault
     ErrFatalError = 8'h C1,
 
--- a/hw/ip/kmac/dv/env/kmac_env_cov.sv
+++ b/hw/ip/kmac/dv/env/kmac_env_cov.sv
@@ -301,8 +301,7 @@ class kmac_env_cov extends cip_base_env_cov #(.CFG_T(kmac_env_cfg));
     kmac_err_code: coverpoint kmac_err {
       ignore_bins ignore = {kmac_pkg::ErrNone};
       // Covered by direct sequence, if scb enabled for those seq, can remove it from this list.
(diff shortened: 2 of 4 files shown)


hw/ip/kmac/rtl/kmac_pkg.sv
```diff
@@ -456,17 +456,14 @@ package kmac_pkg;
...
-    ErrFatalError = 8'h C1,
+    ErrFatalError = 8'h C0,
 
     // Error due to the counter integrity check failure inside MsgFifo.Packer
-    ErrPackerIntegrity = 8'h C2,
+    ErrPackerIntegrity = 8'h C1,
 
     // Error due to the counter integrity check failure inside MsgFifo.Fifo
-    ErrMsgFifoIntegrity = 8'h C3
+    ErrMsgFifoIntegrity = 8'h C2
```
vogelpi, line 466: I think the "right" way to handle this would be just remove `ErrShadowRegUpdate` without modifying the other encodings. Because touching the other encodings would mean an change to the software interface: software compliant with A0/A1 would only be partially compatible. This would mean a version increment. We should avoid this at this point.

Maybe it we should just remove it now but add a note somewhere (in an issue?) to edit also the other encodings next time we do an interface change anyway?

nasahlpa: Done.

My assumption was that SW from the master branch anyways is not compatible with A0/A1 (for different reasons). Hence, I decided to make the change more clean by also moving the error codes.

I created an issue (#25052) for that.

vogelpi: Thanks Pascal! You're right about this but we still want to be careful about introducing breaking changes into the IPs if possible.

vogelpi:
CHANGE AUTHORIZED: hw/ip/kmac/rtl/kmac_pkg.sv

This PR touches an RTL file to remove an unused error code. The used error codes are not touches. This is okay.

vogelpi (approved):

nasahlpa:
CHANGE AUTHORIZED: hw/ip/kmac/rtl/kmac_pkg.sv

This PR touches an RTL file to remove an unused error code. The used error codes are not touches. This is okay.

PR #25043, opened 2024-11-08
[imm_rom_ext] Add a new ROM_EXT target linking nop IM_EXT

lchiawei:
* Add a `nop.o` IMM_ROM_EXT file, which only contains a section called `.rom_ext_immutable` and a single `ret` command.
* Add a new ROM_EXT target to link the above IMM_ROM_EXT, and test it in `rom_ext/e2e/verified_boot`.

This PR partially addresses #24368.


sw/device/silicon_creator/imm_rom_ext/prebuilts/README.md
```diff
@@ -0,0 +1,4 @@
+## NOP IM_ROM_EXT
+
+* The `nop.o` is built by the following command:
```
timothytrippel, line 3: Perhaps you could also elaborate on the purpose of this "NOP immutable ROM_EXT"? I am assuming this is what would be used for testing?

sw/device/silicon_creator/imm_rom_ext/prebuilts/nop.S
```diff
@@ -0,0 +1,8 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+.section .rom_ext_immutable, "ax"
+.globl _start
+_start:
```
timothytrippel, line 7:
```suggestion
.globl _imm_rom_ext_start
_imm_rom_ext_start:
```

Let's make this symbol more specific since it is global.

sw/device/silicon_creator/imm_rom_ext/prebuilts/README.md
lchiawei: Done. I sightly updated the markdown file.

sw/device/silicon_creator/imm_rom_ext/prebuilts/nop.S
lchiawei: Done.

pamaury:
I assume we are checking in the repo the `.o` file because for testing it needs to have a known hash to put in OTP?

lchiawei:
@pamaury 

> I assume we are checking in the repo the `.o` file because for testing it needs to have a known hash to put in OTP?

If you add the `.o` file into the deps of rom_ext target, then the bazel rule `otp_json_immutable_rom_ext` can generate the OTP hash from the rom_ext target directly. The prebuilt `.o` file can be compiled during build time instead, and the OTP hash can still be calculated. The reasons that I check into the `.o` file are:
* I haven't sorted out a buildable imm_rom_ext target. Before the task is done, a `.o` file with a single `ret` command can be used as a minimum imm_rom_ext. By doing this, at least we can test the linker script and the boot flow (ROM -> IMM_ROM_EXT -> ROM -> ROM_EXT -> User Software) is working.
* In the future, the imm_rom_ext will be released as prebuilts, so this `nop.o` can be also considered as an example about how to deal with the prebuilt imm_rom_ext.

timothytrippel (approved):

PR #25046, opened 2024-11-08
[hw,pinmux,rtl] Make USB wakeup an ipgen parameter

Razer6:
Darjeeling does not have USB. Use ipgen to gate the USB wakeup feature.


hw/ip_templates/pinmux/data/pinmux.tpldesc.hjson
```diff
@@ -57,5 +57,11 @@
       type: "int"
       default: "8"
     }
+    {
+      name: "enable_usb_wakeup"
+      desc: "Enable USB wakeup"
+      type: "int"
```
Razer6, line 63: That will be changed to `bool` once https://github.com/lowRISC/opentitan/pull/25013 is merged.

hw/ip_templates/pinmux/data/pinmux.tpldesc.hjson
```diff
@@ -57,5 +57,17 @@
...
+      default: "1"
+    }
+    {
+      name: "enable_strap_sampling"
+      desc: "Set to true to instantiate the HW strap sampling and TAP selection submodule wake submodule"
+      type: "int"
```
Razer6, line 69: same comment as above.

andreaskurth (approved):
LGTM, I think this is a reasonable way to make large parts of existing pinmux functionality customizable, allowing tops that don't need it to opt out. I think this is out of scope for SV parameters. Thx @Razer6!

PR #25048, opened 2024-11-08
[rom_ext] Call immutable section first at startup

sasdf:
To minimize the divergence between SKUs that with and without immutable rom_ext feature enabled in ROM, we propose always linking the section to rom_ext, and call it first if it isn’t called by ROM.

With this change, the boot flows will be always (almost) the same, while the OTP flag `IMMUTABLE_ROM_EXT_EN` basically only controls whether the immutability is enforced by ROM.

SKUs without the immutability enforced enable should still link with the trunk/master imm_rom_ext section, but without the EN bit set in their OTP configuration.

This partially addresses #24368.

A placeholder is added to avoid breaking current tests, and should be replaced by a proper target after #25043 is merged.

### DIFF
--- a/sw/device/silicon_creator/imm_rom_ext/BUILD
+++ b/sw/device/silicon_creator/imm_rom_ext/BUILD
@@ -1,0 +1,11 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "main_lib",
+    srcs = ["imm_rom_ext.c"],
+    hdrs = ["imm_rom_ext.h"],
+)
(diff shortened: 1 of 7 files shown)


sw/device/silicon_creator/imm_rom_ext/BUILD
```diff
@@ -0,0 +1,11 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "main_lib",
```
timothytrippel, line 8: nit: if you named this `imm_rom_ext` then you could refer to it in build rules as `//sw/device/silicon_creator/imm_rom_ext`, instead of `//sw/device/silicon_creator/imm_rom_ext:main_lib`.

timothytrippel (approved):

sw/device/silicon_creator/imm_rom_ext/BUILD
timothytrippel: (also OK with merging this as is; and it can be cleaned up later due to CI times at the moment). Just lmk if you would like me to merge it.

sasdf: Sure, let's clean it up in a later CL.

PR #25056, opened 2024-11-08
[pwm,dv] Use out-of-block declarations in some of the DV code

rswarbrick:
This is not the entire DV code (because the list of commits was getting rather long!), but changes a reasonable tranche. There are five commits:

- [pwm,dv] Change rand_pwm_blink to return the randomised value
- [pwm,dv] Use out-of-block declarations in pwm_base_vseq
- [pwm,dv] Use out-of-block declarations in pwm_common_vseq
- [pwm,dv] Correctly set duty_cycle / blink in pwm_perf_vseq
- [pwm,dv] Use out-of-block declarations in pwm_perf_vseq


hcallahan-lowrisc (approved):
This LGTM, thanks!

PR #25057, opened 2024-11-08
[pwm,dv] Use out-of-block declarations in the rest of the DV code

rswarbrick:
This builds on top of #25056 (which should be merged first). Once that has been merged, there are only the following three commits:

- b1d4a6e906 [pwm,dv] Use out-of-block declarations in pwm_rand_output_vseq
- cad204d17c [pwm,dv] Use out-of-block declarations in pwm_smoke_vseq
- 40e83cf8c5 [pwm,dv] Use out-of-block definitions in pwm_stress_all_vseq

and this converts all of the PWM DV code to use out-of-block declarations.


hcallahan-lowrisc (approved):
LGTM

PR #25061, opened 2024-11-08
[pwm,dv] Lower some reseed counts

rswarbrick:
This should still give a reasonable confidence, but should complete rather faster.


matutem (approved):
The adequacy of this should be reflected in the coverage metrics. I assume you checked that, so LGTM.

rswarbrick:
Honestly, the testing we're doing for pwm is pretty rubbish, but the reseed counts that were checked in won't really improve things. I'm reasonably convinced this is an improvement (or makes no odds!)

PR #25062, opened 2024-11-08
[pwm,rtl] Rewrite a write-enable more explicitly

rswarbrick:
The behaviour will be the same, but it avoids an expression coverage hole being reported for `blink_en_i == 0` and `blink_ctr_q == 1`. This can't actually happen: `blink_ctr_q` can only be updated by being set to `blink_ctr_d` (on this line). If `blink_en_i` is false then `blink_ctr_d == 0`.

The only other way to reach the item would be to have incremented `blink_ctr_q` and then disabled blink. But disabling blink needs a register write, which will set a bit in `clr_blink_cntr` in `pwm_core`, which means that `clr_blink_cntr_i` is true, zeroing `blink_ctr_q`.

Coding this write-enable more explicitly gets rid of the expression coverage item.


vogelpi (approved):

But can you please combine the two commits? Both commits touch the very same file and they both implement a very similar change without functional impact. I don't think we would ever want to undo only one of them.

rswarbrick:
Good point: done. Would you mind doing the "unauthorized changes dance"?

rswarbrick:

vogelpi:

PR #25063, opened 2024-11-08
[pwm,rtl] Recode a 3-way selector to avoid a coverage hole

rswarbrick:
This is equivalent logic but avoids a coverage hole where the simulator complains that we don't hit blink_en_i=1 and htbt_en_i=0 in the second check. This doesn't happen because we don't actually get to the second check in that case.

### DIFF
--- a/hw/ip/pwm/rtl/pwm_chan.sv
+++ b/hw/ip/pwm/rtl/pwm_chan.sv
@@ -152,8 +152,12 @@ module pwm_chan #(
 
   assign duty_cycle_htbt = !dc_wrap_q ? dc_htbt_q : pos_htbt ? 16'hffff : '0;
 
-  assign duty_cycle_actual = (blink_en_i && !htbt_en_i) ? duty_cycle_blink :
-                             (blink_en_i && htbt_en_i) ? duty_cycle_htbt : duty_cycle_a_i;
+  always_comb begin
+    duty_cycle_actual = duty_cycle_a_i;
+    if (blink_en_i) begin
+      duty_cycle_actual = (htbt_en_i) ? duty_cycle_htbt : duty_cycle_blink;
+    end
+  end
 
   // For cases when the desired duty_cycle does not line up with the chosen resolution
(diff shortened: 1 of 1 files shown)


hw/ip/pwm/rtl/pwm_chan.sv
```diff
@@ -152,8 +152,10 @@ module pwm_chan #(
 
   assign duty_cycle_htbt = !dc_wrap_q ? dc_htbt_q : pos_htbt ? 16'hffff : '0;
 
-  assign duty_cycle_actual = (blink_en_i && !htbt_en_i) ? duty_cycle_blink :
-                             (blink_en_i && htbt_en_i) ? duty_cycle_htbt : duty_cycle_a_i;
+  assign duty_cycle_actual = blink_en_i ?
+                             (htbt_en_i ? duty_cycle_htbt : duty_cycle_blink) :
+                             duty_cycle_a_i;
```
vogelpi, line 157: Personally, I find nested ternaries always hard to read. Would this work as well?
```sv
  assign duty_cycle_actual = (blink_en_i && !htbt_en_i) ? duty_cycle_blink :
                             blink_en_i                 ? duty_cycle_htbt  : duty_cycle_a_i;
```

rswarbrick: Probably, but I guess I find that one a bit hard myself! :-) Maybe we should do it with an `always_comb`?
```systemverilog
  always_comb begin
    duty_cycle_actual = duty_cycle_a_i;
    if (blink_en) begin
      duty_cycle_actual = (htbt_en_i) ? duty_cycle_htbt : duty_cycle_blink;
    end
  end
```

vogelpi: SGTM!

vogelpi:

This changes an RTL file but the change has no behavioral impact. This is fine.

vogelpi (approved):

rswarbrick:

PR #25069, opened 2024-11-11
[rom_ext,dice] Move dice_chain code from rom_ext to lib/cert

sasdf:
This PR only moves the code with necessary headers / dependencies rewrite. All functionalities are unchanged.

---

To share the dice code between (imm) rom_ext, those code in rom_ext.c are refactored and move to the lib/cert/dice_chain* as a common library.

PR #24980 has already refactored the code, and this PR moves them out from the rom_ext.c to the library folder.


timothytrippel (approved):
Thanks @sasdf ! In a follow up PR, could you add unit tests for this `dice_chain` lib (perhaps not all functions are easily testable but most should be I think)? The unit tests are compiled and executed on host, and will need to mock some hardware bits. Here is an example: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/cert/cert_unittest.cc

timothytrippel:
Looks like you need to rebase now to deal with conflicts

PR #25070, opened 2024-11-11
[hmac,sw] Fix delay before context save

martin-velay:
- Increase the delay before saving the context. This has been changed from 100 to 200 IBEX cycles as issues have been observed in some cases.
- Linked to issue #24767


martin-velay:
Note, the measured number of clock cycles required for a register read access is:
- 3 clock cycles in simulation, which should be the same on the ASIC
- 5 clock cycles on the FPGA

That's the reason why the `kNumIterTimeout` has been increased to 200

vogelpi (approved):
This looks good to me, thanks @martin-velay for the thorough analysis (and for documenting that)!

PR #25071, opened 2024-11-11
[dv,alerts] Add ping checks to block-level DV

rswarbrick:
These are missing from the DV setup (presumably an oversight), which I realised because of a coverage hole. The first two changes are tidy-ups and the third adds these ping checks.


hw/dv/sv/alert_esc_agent/alert_receiver_driver.sv
```diff
@@ -63,35 +59,49 @@ class alert_receiver_driver extends alert_esc_base_driver;
...
   virtual task rsp_alert();
     forever begin
       alert_esc_seq_item req, rsp;
       wait(r_alert_rsp_q.size() > 0 && !under_reset);
       req = r_alert_rsp_q.pop_front();
+
+      // If we get here then something should have been pushed onto the alert queue in response to
+      // an alert being generated. Wait a short time to allow the alert to propagate through the
+      // clocking blocks.
+      repeat (4) begin
```
marnovandermaas, line 75: How do you know 4 is enough?

marnovandermaas (approved):
Looks reasonable to me and only touches DV code. Minor nit is that you have two sign-offs in the second commit.

hw/dv/sv/alert_esc_agent/alert_receiver_driver.sv
rswarbrick: Honestly, I think 1 is probably enough! I'll tweak the comment and re-push to make that clear.

PR #25072, opened 2024-11-11
[tmp] Attempt to fix reset handling

martin-velay:


martin-velay:
Hi @rswarbrick,
I am sorry for making you review this PR, my intent was only to share this work with Marcelo in case this can help him. This doesn't work at all yet. But thanks anyway, when I'll restart on it I'll take your comments into account.

hw/dv/sv/dv_lib/dv_base_sequencer.sv
marcelocarvalhoLowRisc, line 1: Could you please explain where this reset_done event is been used? Be careful using events in UVM, is quite tricky sometimes =)

marcelocarvalhoLowRisc: In my experience, sometimes when I used events can cause race conditions or deadlocks in scenarios involving forked processes, If multiple processes are waiting on the same event, there’s no guarantee of the order in which they will resume, causing unpredictable behaviour. And sometimes I wouldn't know where or what is causing it. Besides of the reusability perspective is not good, SystemVerilog events lack the structured logging and tracing features provided by UVM objects, making it harder to debug, trace, and modify the testbench when issues arise.

hw/dv/sv/dv_lib/dv_base_driver.sv
```diff
@@ -17,10 +17,27 @@ class dv_base_driver #(type ITEM_T     = uvm_sequence_item,
   `uvm_component_new
 
   virtual task run_phase(uvm_phase phase);
-    fork
-      reset_signals();
-      get_and_drive();
-    join
+    wait(under_reset);
```
antmarzam, line 20: Should this be `cfg.in_reset` instead, just so all agent components feed out of the same reset source?

martin-velay: That's possible. I was confused with `under_reset` or `cfg.in_reset`. I don't really understand why we do have 2 similar variables. This should be reviewed once we'll have a proper reset agent

PR #25073, opened 2024-11-11
[pwm,dv] Exclude impossible toggle coverage in u_reg.u_chk

rswarbrick:
Honestly, this isn't really a pwm-specific change so we'll probably want to move it eventually. Before this change, he reported toggle coverage misses bits 56:43 of data_i in tb.dut.u_reg.u_chk.u_chk. This instance is made with:

    prim_secded_inv_64_57_dec u_chk (
      .data_i({tl_i.a_user.cmd_intg, H2DCmdMaxWidth'(cmd)}),
      .data_o(),
      .syndrome_o(),
      .err_o(err)
    );

where cmd is 43 bits wide and H2DCmdMaxWidth is 57. As such, bits 56:43 are constant.


alees24 (approved):
Trusting you on the syntax of exclude files for xcelium, since I've only worked with vcs.
Thanks.

PR #25080, opened 2024-11-11
[ownership] Add datatypes for hybrid keys

cfrantz:

PR #25083, opened 2024-11-12
[rom_ext] Add SPX+ verification for owner firmware 

cfrantz:
1. Add fake SPX+ keys to the test_owner configuration.
2. Use key_ids to distinguish between plain and hybrid keys (hybrid
   key_ids are the xor of the individual key_ids).
3. Detect the SPX manifest extensions and update the search key_id
   accordingly.
4. Verify hybrid signatures by checking both ECDSA and SPX signatures.

Note: this PR depends on #25080.

### DIFF
--- a/sw/device/silicon_creator/lib/ownership/owner_block.c
+++ b/sw/device/silicon_creator/lib/ownership/owner_block.c
@@ -304,11 +304,15 @@ rom_error_t owner_block_info_apply(const owner_flash_info_config_t *info) {
 }
 
 rom_error_t owner_keyring_find_key(const owner_application_keyring_t *keyring,
-                                   uint32_t key_alg, uint32_t key_id,
-                                   size_t *index) {
+                                   uint32_t key_id, size_t *index) {
   for (size_t i = 0; i < keyring->length; ++i) {
-    if (keyring->key[i]->key_alg == key_alg &&
-        keyring->key[i]->data.id == key_id) {
+    uint32_t id = keyring->key[i]->data.id;
+    if ((keyring->key[i]->key_alg & kOwnershipKeyAlgCategoryMask) ==
+        kOwnershipKeyAlgCategoryHybrid) {
+      // The ID of a hybrid key is the xor of the IDs of each key.
+      id ^= keyring->key[i]->data.hybrid.spx.data[0];
+    }
+    if (id == key_id) {
       *index = i;
       return kErrorOk;
     }
--- a/sw/device/silicon_creator/rom_ext/rom_ext.c
+++ b/sw/device/silicon_creator/rom_ext/rom_ext.c
@@ -50,6 +50,7 @@
 #include "sw/device/silicon_creator/lib/sigverify/ecdsa_p256_key.h"
 #include "sw/device/silicon_creator/lib/sigverify/rsa_verify.h"
 #include "sw/device/silicon_creator/lib/sigverify/sigverify.h"
+#include "sw/device/silicon_creator/lib/sigverify/sphincsplus/verify.h"
 #include "sw/device/silicon_creator/manuf/base/perso_tlv_data.h"
 #include "sw/device/silicon_creator/rom_ext/rescue.h"
 #include "sw/device/silicon_creator/rom_ext/rom_ext_boot_policy.h"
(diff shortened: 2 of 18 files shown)


sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -256,11 +256,15 @@ rom_error_t owner_block_info_apply(const owner_flash_info_config_t *info) {
...
-                                   uint32_t key_alg, uint32_t key_id,
-                                   size_t *index) {
+                                   uint32_t key_id, size_t *index) {
   for (size_t i = 0; i < keyring->length; ++i) {
-    if (keyring->key[i]->key_alg == key_alg &&
-        keyring->key[i]->data.id == key_id) {
+    uint32_t id = keyring->key[i]->data.id;
+    if ((keyring->key[i]->key_alg & kOwnershipKeyAlgCategoryMask) ==
+        kOwnershipKeyAlgCategoryHybrid) {
+      // The ID of a hybrid key is the sum of the IDs of each key.
```
jadephilipoom, line 264: How do we know this won't overflow/that each hybrid key ID will be unique compared to other hybrid or non-hybrid keys?

sw/device/silicon_creator/rom_ext/rom_ext.c
```diff
@@ -257,18 +258,112 @@ void rom_ext_sram_exec(owner_sram_exec_mode_t mode) {
...
+                     (unsigned char *)digest->digest, sizeof(digest->digest),
+                     key->data, actual_root.data));
+      break;
+    default:
+      return kErrorSigverifyBadSpxConfig;
+  }
+  uint32_t result = 0;
+  for (--i; launder32(i) < kSigverifySpxRootNumWords; --i) {
+    result ^= expected_root.data[i] ^ actual_root.data[i];
+  }
```
jadephilipoom, line 330: The algorithm we use in ROM for this comparison is a little more complex and uses an additional mask variable `diff` https://github.com/lowRISC/opentitan/blob/16484156dd156f045ba2c6576c6a21cdca8e3674/sw/device/silicon_creator/lib/sigverify/spx_verify.c#L130

Does it make sense to consider a similar construction here? Or is it appropriate to be less conservative here than in ROM, and perhaps just highlight this to pentesters for evaluation?

jadephilipoom (approved):

sw/device/silicon_creator/lib/ownership/owner_block.c
cfrantz: We don't know that (any more than we know that the least significant word of any key will be unique).  I can change this operation to XOR instead of a sum to not lose the potential carry bit.  IMHO, we'll still have the same potential uniqueness problem.

Another alternative is to use a 64-bit integer and concatenate the two u32's together as the key ID.

WDYT?

sw/device/silicon_creator/rom_ext/rom_ext.c
cfrantz: The use of `diff` seems to be to clobber all of the bits in the result if there is an invalid signature.  I think that makes sense for clobbering the `flash_exec` "password".  We can probably add a similar construction here to clobber the result in a similar way.

sw/device/silicon_creator/lib/ownership/owner_block.c
moidx: Another option is to have the key look up operations implemented in a similar way to the ROM, and then add a separate function to verify the hybrid policy, to ensure that both keys are part of the same key slot. 

```
owner_keyring_hybrid_validate(ecdsa_key_id, spx_id)
```

I was thinking that this way we can avoid adding the new switch case statements added to the rom_ext to simplify the hardening.

cfrantz: I decided to xor the `key_ids`.  This allows the possibility of having an SPX-only signature in the future.

sw/device/silicon_creator/rom_ext/rom_ext.c
cfrantz: Done.

moidx (approved):

PR #25088, opened 2024-11-12
[pwm,dv] Manually exclude some coverage in prim_reg_cdc instances

rswarbrick:
This expression coverage is expecting to see a src_update signal equal to 1. That signal can only be nonzero if the DstWrReq parameter is 1, but it is wired to zero.

A working UNR flow would waive this but we don't have that set up properly in a way that I can run it right now, so I'm doing the manual exclusion instead.

### DIFF
--- a/hw/ip/pwm/dv/cov/manual_excl.vRefine
+++ b/hw/ip/pwm/dv/cov/manual_excl.vRefine
@@ -1,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<refinement-file-root>
+  <rules>
+    <!--
+     The blocks of code below exclude some expression coverage in prim_reg_cdc to waive values of
+     the expression on line 125 that depends on the src_update value being nonzero. This is only
+     possible when DstWrReq is one, but it is zero in the instantiations in pwm.
+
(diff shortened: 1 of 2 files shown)


matutem:
This is a bit problematic since it is specific to pwm while the exclusion should really apply to any instance of prim_reg_cdc with DstWrReq==0; it also doesn't address VCS, which is our sign-off.

I think this may be better handled in the RTL instead, if we connect the prim_reg_cdc_arb src_update_o output to a new net like src_update_cdc, and adding a generate like

if (DstWrReq == 0) begin
  // Potentially add an FPV check for src_update_cdc == 0.
  assign src_update = 1'b0;
end else
  assign src_update = src_update_cdc;
end

Of course, adequate comments explaining this code would be indispensable.

This sort of RTL change is something we have done in the past to avoid waivers for obvious coverage holes, since waivers have a lot of overhead, can easily be applied incorrectly, and don't cover all instantiations of prims.

rswarbrick:
I think you're right. Here, the exclusion is to line 125 of `prim_reg_cdc.sv`, which looks like:
```systemverilog
    end else if (src_busy_q && src_ack || src_update && !busy) begin
```

I think that the change you're describing won't quite work, but something similar should work. I'm going to close this PR and do that.

PR #25094, opened 2024-11-12
[ci] move GitHub runner to Ubuntu 22.04

nbdd0121:
GitHub action is moving latest to 24.04 and 20.04 may be deprecated soon. The Ubuntu 20.04 OS itself will also reach EOL early next year.


jwnrt (approved):
LGTM

Note that this effectively bumps our minimum supported OS to Ubuntu 22.04 since we'll no longer be testing Ubuntu 20.04 for everything.

PR #25095, opened 2024-11-12
[uart,dv] Tweak modelling of TX FIFO interrupts

elliotb-lowrisc:
Add a wait cycle between updates to the TX watermark/empty interrupts. This represents the single cycle that the interrupts may flip for between a TX item being added by the register interface and *almost* immediately popped off by the transmission logic.


rswarbrick (approved):
Looks sensible to me!

PR #25087, opened 2024-11-12
[bazel,otp] update manifest with immutable ROM_EXT hash

anthonychen1251:
This PR has two commits that:

1. Refactor `gen-otp-immutable-rom-ext-json.py` script to move
the logic for extracting the immutable ROM_EXT section data and
computing the OTP values to a new class `ImmutableSectionProcessor`
2. Add a new tool `update-manifest-json` that updates the `manuf_state_creator` field in a ROM_EXT manifest with the hash of the immutable ROM_EXT section data.

Note: this depends on https://github.com/lowRISC/opentitan/pull/24948. Only review the last 2 commits.
This PR addresses https://github.com/lowRISC/opentitan/issues/24610 partially.

PR #25106, opened 2024-11-13
[otp] decouple CREATOR_SW_CFG_MANUF_STATE from SKU

anthonychen1251:
This commit decouples the `CREATOR_SW_CFG_MANUF_STATE` OTP field from the SKU.

Previously, the `CREATOR_SW_CFG_MANUF_STATE` field was used to identify the SKU. However, we now leverage Device ID for SKU identification, eliminating the need to use `CREATOR_SW_CFG_MANUF_STATE` for this purpose. We can now embed the first four bytes of the immutable ROM_EXT hash into the `CREATOR_SW_CFG_MANUF_STATE` field.

The following changes were made to implement this decoupling:

- The `CREATOR_SW_CFG_MANUF_STATE` field in the `otp_ctrl` data structure is now set to a fixed value, `CONST.MANUF_STATE.PERSONALIZED`.
- The `MANUF_STATE` enum in the `const.bzl` file has been updated to remove the `SIVAL` value and add a `PERSONALIZED` value.
- The `update_creator_manuf_state_data` function in the `ImmutableSectionProcessor.py` file has been updated to no longer require the `creator_manuf_state` argument.
- The `update_manifest_json.py` script has been updated to no longer require the `manuf_state_creator` argument.

Note: this depends on https://github.com/lowRISC/opentitan/pull/24948 and https://github.com/lowRISC/opentitan/pull/25087. Only review the last commit.
This PR addresses https://github.com/lowRISC/opentitan/issues/24610 partially.

### DIFF
--- a/util/design/update-manifest-json.py
+++ b/util/design/update-manifest-json.py
@@ -8,11 +8,15 @@ import argparse
 import json
 
 import hjson
+
 from lib.ImmutableSectionProcessor import ImmutableSectionProcessor
 
 _USAGE_CONSTRAINTS_NAME = "usage_constraints"
 _MANUF_STATE_CREATOR_NAME = "manuf_state_creator"
+_SELECTOR_BITS_NAME = "selector_bits"
 _IDENTIFIER_NAME = "identifier"
+
+_SEL_MANUF_STATE_CREATOR = (1 << 8)
 
 # This must match the definitions in chip.h.
 _CHIP_ROM_EXT_IDENTIFIER = 0x4552544f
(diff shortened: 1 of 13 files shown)


util/design/update-manifest-json.py
```diff
@@ -0,0 +1,110 @@
...
+        Returns:
+            None
+        """
+
+        selector_bits = self.json_data[_USAGE_CONSTRAINTS_NAME][
+            _SELECTOR_BITS_NAME
+        ]
+
+        # Ensure the selector bit of `manuf_state_creator` field is set.
+        new_selector_bits = selector_bits | _SEL_MANUF_STATE_CREATOR
```
timothytrippel, line 42: Shouldn't the manifest bazel rule already do this: https://cs.opensource.google/opentitan/opentitan/+/master:rules/manifest.bzl;l=81?q=manifest.bzl&ss=opentitan%2Fopentitan

timothytrippel (approved):
LGTM, thanks @anthonychen1251 ! Waiting for @cfrantz to review too before merging.

util/design/update-manifest-json.py
anthonychen1251: yes, the selector bit of `manuf_state_creator` field will be set by the manifest bazel rule if the `manuf_state_creator` attribute is provided.

The operation here is to handle the case that the `manuf_state_creator` attribute wasn't specified when generating the manifest json file. Alternatively,  we can assume that `manuf_state_creator` should be specified when generating the manifest json file and validate its presence by checking the selector bit here. Let me know which approach works best for you.

timothytrippel: Ah got it ok that makes sense to me. We can keep as is then.

timothytrippel:
I rebased on #25087 to fix merge conflict. Will merge when this passes CI.

PR #22816, opened 2024-04-25
Backport 22701 to master

engdoreis:
This is a manual backport PR of #22701 to master.
The automatic backport failed because the master branch has new CRC functions for ujson.

PR #22729, opened 2024-04-19
[sival] Mark `i2c_target_test_fpga_cw310_sival` as broken

vogelpi:
I've seen this test failing in multiple PRs now that don't touch any of I2C. I think it's most likely broken.

Examples of where it failed CI:
- https://github.com/lowRISC/opentitan/pull/22599
- https://github.com/lowRISC/opentitan/pull/22588
- https://github.com/lowRISC/opentitan/pull/22621

PR #25110, opened 2024-11-13
[sival,i2c] Fix and re-enable I2C target test by adding config ujson CRC

AlexJones0:
The I2C target test was failing because the ujson `I2cTestConfig` command added for clock stretching (#22816) when backported from `earlgrey_es_sival` did not have a CRC added to it (#22277), causing the de-serialization to fail. This was not noticed as the test was marked broken at the time due to an unrelated issue (#22729). This commit adds this CRC, adds an additional CRC to the configuration payload itself, and marks the test as no longer being broken. 

This has been tested locally to pass on a CW310 using: `./bazelisk.sh test -t- --test_output=streamed --runs_per_test 10 //sw/device/tests:i2c_target_test_fpga_cw310_sival`.

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -5529,11 +5529,10 @@ opentitan_test(
     srcs = [
         "i2c_target_test.c",
     ],
-    broken = fpga_params(tags = ["broken"]),
     exec_env = dicts.add(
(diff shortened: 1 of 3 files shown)


jwnrt (approved):
Nice work, thanks

engdoreis (approved):

PR #25111, opened 2024-11-13
Backport 25065 from master to earlgrey_1.0.0 

engdoreis:
Update testplans and SiVal test suites

This Backport depends on https://github.com/lowRISC/opentitan/pull/25091 to be merged first.


hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
```diff
@@ -192,7 +192,7 @@
             SiVal: This is a connectivity test only, not suitable for sival.
             '''
       stage: V2
-      si_stage: NA
+      si_stage: SV3
```
matutem, line 195: The testpoint describes this test as a connectivity only, and the code in clkmgr_jitter_test.c is pretty lame, and doesn't take into account the jitter_regwen bit. In fact, I'd say that test is not useful. Instead, jitter can be tested via the testpoint chip_sw_clkmgr_jitter_cycle_measurements, and even that test doesn't take jitter_regwen into account.

The point about jitter_regwen is that it may not be possible to change the state of jitter_enable, so it may be frozen to either enable or disable: if enable we will never change and the frequencies will always correspond to the nominal frequency; if disabled we cannot switch it, and only then will the frequency measurements really be able to confirm it. This area seems weak: I filed https://github.com/lowRISC/opentitan/issues/25177 regarding these tests.

hw/top_earlgrey/data/ip/chip_flash_ctrl_testplan.hjson
```diff
@@ -23,7 +23,7 @@
       features: ["FLASH_CTRL.INIT.SCRAMBLING_KEYS", "FLASH_CTRL.INIT.ROOT_SEEDS",
                  "FLASH_CTRL.INFO.CREATOR_PARTITION", "FLASH_CTRL.INFO.OWNER_PARTITION"]
       stage: V2
-      si_stage: NA
+      si_stage: SV3
```
matutem, line 26: The //sw/device/tests:flash_ctrl_test does not cover this testpoint: it only tests info pages 5 and 9, and a data page. The tests are valid and useful but have nothing to do with this testpoint, rather it tests page attributes like scrambling and access permissions. The scrambling access is weak, since write followed by read should always result in identical data regardless of scrambling, so this only catches weird errors. Perhaps a better test is to write with some setting of scrambling and read with the opposite setting: this should always result in a fatal read error.

hw/top_earlgrey/data/ip/chip_pwrmgr_testplan.hjson
```diff
@@ -82,7 +82,7 @@
             SiVal: No need to run this, run chip_sw_pwrmgr_random_sleep_all_wake_ups instead.
             '''
       stage: V2
-      si_stage: NA
+      si_stage: SV3
```
matutem, line 85: Please notice line 82: there is a more comprehensive test that covers this functionality, and this will only add CI runtime.
If we really intend to run these tests, please remove line 82, here and in the other two pwrmgr tests that were marked as not needed.

matutem (changes requested):
I flagged some issues with clkmgr and pwrmgr tests, and filed an issue for the clkmgr testpoints mentioned. Also, we need to resolve if redundant tests in pwrmgr are really worth running and taking CI time.

Obviously these issues need to be resolved in master, and if some change is made they should be cherry-picked.

engdoreis:
Thanks @matutem for the review.
Do we never want to run these tests that you commented on in Silicon?
If yes, I should change the PR and keep them as NA, otherwise we merge it as it is, and then we open issues to adjust the tests as you suggested.

matutem:
@engdoreis I will be working on https://github.com/lowRISC/opentitan/issues/25177 now and expect to have `"//sw/device/tests:clkmgr_jitter_frequency_test"` ready and testing jitter frequency more thoroughly. I will change `"//sw/device/tests:clkmgr_jitter_test"` to be a test of jitter_regwen. The various bazel tests associated to chip_sw_clkmgr_jitter need to be examined individually: I noticed flash_ctrl_ops_test.c doesn't even turn on clock jitter, so it is surprising it made it into this testpoint.

As for the other changes, my concern about pwrmgr is weak, since it may be useful to have simpler tests than the more comprehensive one and the tests are quick, so they are good to go.

I am concerned about the chip_sw_flash_init testpoint: I think we need a more robust test in master before cherry picking it.

So we need a new issue to address the shortcomings of chip_sw_flash_init in master before cherry picking it.

Bottom line, all but the clkmgr and flash_ctrl testplan changes are okay. WDYT? @moidx?

hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
engdoreis: Reverted to NA

hw/top_earlgrey/data/ip/chip_flash_ctrl_testplan.hjson
engdoreis: Kept SV3

hw/top_earlgrey/data/ip/chip_pwrmgr_testplan.hjson
engdoreis: Removed line 82

engdoreis:
@matutem I made the suggested changes.

matutem (approved):
Thanks Douglas. LGTM

PR #25113, opened 2024-11-13
[sival,kmac] Add chip_sw_kmac_entropy test

nasahlpa:
Map `kmac_entropy_test` to `chip_sw_kmac_entropy` and include more execution targets.

### DIFF
--- a/hw/top_earlgrey/data/ip/chip_kmac_testplan.hjson
+++ b/hw/top_earlgrey/data/ip/chip_kmac_testplan.hjson
@@ -96,7 +96,7 @@
       si_stage: SV3
       lc_states: ["PROD"]
       tests: ["chip_sw_kmac_entropy"]
-      bazel: []
+      bazel: ["//sw/device/tests:kmac_entropy_test"]
     }
     {
       name: chip_sw_kmac_idle
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -2279,15 +2279,22 @@ opentitan_test(
 opentitan_test(
     name = "kmac_entropy_test",
     srcs = ["kmac_entropy_test.c"],
-    exec_env = EARLGREY_TEST_ENVS,
-    # TODO(#15530): EDN doesn't timeout on FPGA
-    fpga = fpga_params(tags = ["broken"]),
+    exec_env = dicts.add(
+        EARLGREY_SILICON_OWNER_ROM_EXT_ENVS,
+        {
+            "//hw/top_earlgrey:fpga_cw340_sival": None,
+            "//hw/top_earlgrey:sim_dv": None,
+            "//hw/top_earlgrey:sim_verilator": None,
+        },
+    ),
+    verilator = verilator_params(timeout = "long"),
     deps = [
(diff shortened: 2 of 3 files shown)


nasahlpa:
Not sure if the [chip_sw_kmac_entropy](https://opentitan.org/book/hw/top_earlgrey/data/chip_testplan.html?highlight=chip_sw_aes_interrupt_encryption#chip_sw_kmac_entropy) testpoint description needs to be improved.

The existing [kmac_entropy_test](https://github.com/lowRISC/opentitan/blob/master/sw/device/tests/kmac_entropy_test.c) already checks the following from the testpoint:
```
            Randomly configure `wait_timer` values (zero for disable, non-zero for timer expire).
              - Program `wait_timer` to a small value.
                Check if EDN timeout error occurs after issuing the hashing op.
              - Adjust `wait_timer` greater than expected EDN latency (with correct `prescaler`
                too). Then check if Digest is correct.
```
However, not sure if I really can check the following in SW:
```
            KMAC should send EDN request after `entropy_ready` is set.
            KMAC also should send out another request to EDN when either:
              - kmac hash counter hits the configured threshold (assuming it is non-zero)
              - Hash count exceeds the threshold.
            SW verifies that KMAC produces the correct digest value.
```
WDYT @vogelpi?

vogelpi:
At the moment, I see two possible options to test that KMAC sends out the EDN request (approximately) at the right time:
- Config EDN0 and CSRNG such that per 128-bit EDN request, an entropy request to the entropy source is made. Then, you can check the cs_entropy_req interrupt of CSRNG. It's going high whenever CSRNG requests a new seed from the entropy source.
- You disable the entropy complex and trigger KMAC operations until KMAC stalls (because it's waiting for entropy). Then you enable the entropy complex and you note that KMAC is now progressing again. This is what is done for AES SiVal IIRC.

In both cases, SW has to track the number of hashes to understand when the interrupt / stalling is expected.

nasahlpa:
Thanks @vogelpi!
I've now extended the test to also test the stall behavior of the KMAC core. I am disabling the entropy complex before a EDN request is expected and I am checking whether KMAC stalls. Afterwards, I enable the entropy complex again and check if we can continue with the operation.

nasahlpa:
Merging this now.

The failing test on CW310 SiVal: i2c_target_test_fpga_cw310_sival is unrelated to this PR as this test fails also in other CI runs ([1](https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=161731&view=logs&j=5011a995-22d7-5073-e1ce-26fc382a8885&t=875997a6-7e2b-594d-05e6-ce6a085f3383), [2](https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=161713&view=logs&j=5011a995-22d7-5073-e1ce-26fc382a8885&t=875997a6-7e2b-594d-05e6-ce6a085f3383), [3](https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=161705&view=logs&j=5011a995-22d7-5073-e1ce-26fc382a8885&t=875997a6-7e2b-594d-05e6-ce6a085f3383))

PR #25114, opened 2024-11-13
[rom_ctrl, dv] Coverage Collection for tlul_adapter_sram

KinzaQamar:
The coverage collection for tlul_adapter_sram for rom_ctrl_tl_intg_err test was disabled. The rom_ctrl_tl_intg_err test covers the intg_err signal in tlul_adapter_sram in rom_ctrl.
To let the nightly regression dashboard reporting intg_err as 1/0 being covered for the conditional statements present in rom_ctrl tlul_adapter_sram, this PR added a cover_reg_top.cfg file and overrides it in rom_ctrl_base_sim_cfg.hjson.

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/cover_reg_top.cfg
+++ b/hw/ip/rom_ctrl/dv/cov/cover_reg_top.cfg
@@ -1,0 +1,10 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Collect coverage for tlul_adapter_sram to resolve coverage hole for intg_err.
++tree tb.dut.u_tl_adapter_rom
+
+begin line+cond+fsm+branch+assert
+  +moduletree tlul_adapter_sram
+end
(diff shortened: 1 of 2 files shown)


hw/ip/rom_ctrl/dv/cov/cover_reg_top.cfg
```diff
@@ -0,0 +1,12 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
++node tb.dut *_tl_*
```
rswarbrick, line 5: It's not really a problem, but isn't this picked up by `hw/dv/tools/vcs/cover_reg_top.cfg`? I think we don't actually need the line?

hw/ip/rom_ctrl/dv/rom_ctrl_base_sim_cfg.hjson
```diff
@@ -57,6 +57,14 @@
       name: rel_path
       value: "hw/ip/{name}_{variant}/dv"
     }
+    {
+      name: default_vcs_cov_cfg_file
+      value: "-cm_hier {proj_root}/hw/dv/tools/vcs/cover.cfg"
+    }
```
rswarbrick, line 63: I think this is the value it had before? So we probably don't need this override?

PR #25120, opened 2024-11-13
[tlul,rtl] Rewrite definition of wr_attr_error

rswarbrick:
This code (from 7d6d53d11f2) was hard to read! It's not doing anything very clever, so let's rewrite it in a way that's easier to parse.


alees24 (approved):
Thanks for tidying this. Bizarre that it was left expressed in that manner.

rswarbrick:

This is equivalent logic, so should present no risk to the design.

PR #25124, opened 2024-11-13
[ot_certs] Various fixes

pamaury:
The primary goal of this PR is to fix some flakiness in the `ot_certs` parse test. See the commit for an explanation.

The short summary is that OpenSSL has an undocumented quirk: if the basic key usage is empty (and possibly in other circumstances?), it silently pretend that the certificate has no authority or subject key identifier. It is not clear whether this is intentional or a side-effect of a bug but the result is the same: a seemingly valid (but stupid) certificate will generate an error when parse with ot_certs. When we generate random certificates for parsing test, we can sometimes create some with no key usage set which triggers this behaviour.

To solve this, I have switch to manually parsing of the auth key id and subj key id. This is very easy and we already parse most extensions manually anyway.

### DIFF
--- a/sw/host/ot_certs/src/asn1/x509.rs
+++ b/sw/host/ot_certs/src/asn1/x509.rs
@@ -153,8 +153,12 @@ impl X509 {
                         if let Some(key_usage) = &cert.key_usage {
                             Self::push_key_usage_ext(builder, key_usage)?;
                         }
-                        Self::push_auth_key_id_ext(builder, &cert.authority_key_identifier)?;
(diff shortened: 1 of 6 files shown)


timothytrippel (approved):

jwnrt (approved):
Great, thanks

pamaury:
I realized that the making the random generator deterministic does not make the test reproducible because we use openssl to generate random keys. I have replaced the first commit with something much simpler: print the entire test data.

pamaury:
I also had a strange lint failure that seems unrelated to the PR

PR #25127, opened 2024-11-13
bazel: fix //quality:rustfmt_fix

vbendeb:
There is no bindgen module rom_error any more (or maybe not yet?), in either case this patch is needed to make sure that 'basel run //quality:rustfmt_fix' can actually execute.


timothytrippel (approved):

timothytrippel:
Looks like this module is used afterall ... @vbendeb

vbendeb:
@timothytrippel  can you run `basel run //quality:rustfmt_fix` in es100? It is failing for me...

PR #25128, opened 2024-11-13
[dv,adc_ctrl] Make dv improvements

matutem:
- Move SVA related code from tb.sv to adc_ctrl_sva_if.sv.
- Bind adc_ctrl_sva_if for sva only.
- Change module adc_ctrl_fsm_sva to interface adc_ctrl_fsm_sva_if.
- Fix some trivial typos, and remove useless functions.

### DIFF
--- a/hw/ip/adc_ctrl/dv/sva/adc_ctrl_sva_if.sv
+++ b/hw/ip/adc_ctrl/dv/sva/adc_ctrl_sva_if.sv
@@ -1,0 +1,123 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+// System verilog assertions for the adc_ctrl_fsm module
+
+interface adc_ctrl_sva_if
+  import uvm_pkg::*;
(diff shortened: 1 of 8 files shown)


hw/ip/adc_ctrl/dv/sva/adc_ctrl_sva_if.sv
```diff
@@ -0,0 +1,123 @@
...
+  end
+  // Pulse to check wake up counter
+  wire wakeup_time_chk = ~adc_o.pd & pd_prev;
+  // Model expects RTL to be in low power mode.
+  wire testmode_low_power = cfg_testmode inside {AdcCtrlTestmodeLowpower} && cfg_lp_mode;
+`endif
+
+  // Check the DUT enters low power
+  // In low power test mode, after falling edges on power down
+  // and the last ADC channel select, power down should be re-asserted within 10 clock cycles
```
rswarbrick, line 98: Is "10" the right number here? I think this corresponds to the `##[0:3]` in the consequent. Does one of the numbers need adjusting?

rswarbrick (approved):
This looks good to me. The question I left was before I realised the code was just moving, so it definitely doesn't need to gate the PR.

PR #25129, opened 2024-11-14
[util,topgen] Remove support for templated ips

matutem:
This has 4 commits with no breakages in between:
- Add cmdgen to "all" target in hw/Makefile
- Move type hints from comments to variable declarations in util/topgen.py
- Remove support for templated ips in topgen
- Improve ipgen docs, exception handling, and make minor implifications

Some are part of #8440

### DIFF
--- a/util/ipgen/README.md
+++ b/util/ipgen/README.md
@@ -1,4 +1,4 @@
-# Ipgen: Generate IP blocks from IP templates
+# Ipgen: Generate IP Blocks From IP Templates
 
 Ipgen is a tool to produce IP blocks from IP templates.
 
@@ -10,7 +10,6 @@ Ipgen is a command-line tool and a library.
 Users wishing to instantiate an IP template or query it for template parameters will find the command-line application useful.
 For use in higher-level scripting, e.g. within [topgen](../topgen/README.md) using ipgen as Python library is recommended.
 
-
 ## Anatomy of an IP template
 
 An IP template is a directory with a well-defined directory layout, which mostly mirrors the standard layout of IP blocks.
@@ -18,26 +17,25 @@ An IP template is a directory with a well-defined directory layout, which mostly
 An IP template directory has a well-defined structure:
 
 * The IP template name (`<templatename>`) equals the directory name.
-* The directory contains a file `data/<templatename>.tpldesc.hjson` containing all configuration information related to the template.
-* The directory also contains zero or more files ending in `.tpl`.
-  These files are Mako templates and rendered into a file in the same relative location without the `.tpl` file extension.
+* The directory contains a template description file `data/<templatename>.tpldesc.hjson` containing descriptions of the configurable parameters.
+  The "default" field of these descriptions are expected to be overriden via the actual configuration parameters.
+* The directory also contains some files ending in `.tpl`.
+  These files are Mako templates and are rendered into a file in the same relative location without the `.tpl` file extension.
 
 ### The template description file
 
 Each IP template comes with a description itself.
-This description is contained in the `data/<templatename>.tpldesc.hjson` file in the template directory.
-The file is written in Hjson.
+This description is contained in the hjson file `data/<templatename>.tpldesc.hjson` in the template directory.
 
-It contains a top-level dictionary, the keys of which are documented next.
-
-#### List of template parameters: `template_param_list`
-
-Keys within `template_param_list`:
+It contains a list of parameter objects.
+These objects are dictionaries with the following required keys:
 
 * `name` (string): Name of the template parameter.
 * `desc` (string): Human-readable description of the template parameter.
-* `type` (string): Data type of the parameter. Valid values: `int`, `str`
-* `default` (string|int|object): The default value of the parameter. The data type should match the `type` argument. As convenience, strings are converted into integers on demand (if possible).
+* `type` (string): Data type of the parameter. Valid values: `bool`, `int`, `str`, `object`.
+* `default` (bool|string|int|dict): The default value of the parameter.
+  The type of this should match the `type` argument.
+  For convenience, strings are converted into integers on demand (if possible).
 
 #### Example template description file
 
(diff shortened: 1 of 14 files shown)


util/topgen.py
```diff
@@ -475,15 +469,10 @@ def generate_top_only(top_only_dict: Dict[str, bool], out_path: Path,
...
-
-        if reggen_only and alt_hjson_path is not None:
-            hjson_dir = Path(alt_hjson_path)
-        else:
-            hjson_dir = (SRCTREE_TOP / "hw" / top_name / "ip" / ip / "data")
-
-        hjson_path = hjson_dir / f"{ip}.hjson"
-
-        genrtl_dir = out_path / "ip" / ip / "rtl"
+        log.info(f"generate_top_only: Ip {ip}, reggen {reggen_only} alt_hjson {alt_hjson_path}")
```
rswarbrick, line 472: I think we're no longer looking at `alt_hjson_path`. Is that intentional? (If so, it can probably be ripped out?)

util/ipgen/README.md
```diff
@@ -88,13 +91,13 @@ CAPI=2:
 name: ${instance_vlnv("lowrisc:ip:rv_plic")}
 ```
 
-After processing, the `name` key is set to e.g. `lowrisc:opentitan:top_earlgrey_rv_plic`.
+After processing, the VLNV could become `lowrisc:opentitan:top_earlgrey_rv_plic`.
```
rswarbrick, line 94: Nit: I think we probably want "would become" (since it's definitely what would happen)?

util/ipgen/renderer.py
```diff
@@ -99,6 +99,23 @@ def _get_mako_template_lookup(self) -> MakoTemplateLookup:
...
+        the version is optional. This is transformed as follows:
+        - The vendor string becomes `lowrisc`.
+        - The library string becomes `opentitan`.
+        - The name is processed as follows:
+          - If the `module_instance_name` parameter exists, the name must start
+            with it as a prefix, and that is replaced by the `instance_name`
+            value.
+          - If the name starts with the template name, it is replaced by the
+            `instance_name` value.
+          - Otherwise the `instance_name` is pefixed to the name.
```
rswarbrick, line 114: Nit: Spelling of "prefixed"

util/ipgen/renderer.py
```diff
@@ -108,13 +125,12 @@ def _tplfunc_instance_vlnv(self, template_vlnv_str: str) -> str:
...
-        # instance name 'my_instance_component' (instead of
-        # 'my_instance_rv_plic_component').
         if "module_instance_name" in self.ip_config.param_values:
-            assert template_core_name.startswith(
-                self.ip_config.param_values["module_instance_name"])
+            if not template_core_name.startswith(
+                    self.ip_config.param_values["module_instance_name"]):
+                raise TemplateRenderError(
+                    f'The name field in {template_vlnv_str} must start with '
+                    'self.ip_config.param_values["module_instance_name"]')
```
rswarbrick, line 133: Don't you want to insert the value of this module instance name here, rather than the name of the Python variable?

util/ipgen/renderer.py
```diff
@@ -226,116 +232,144 @@ class IpBlockRenderer(IpTemplateRendererBase):
...
+                    'Unable to delete the backup directory '
+                    f'{output_dir_existing_bak} of the overwritten data. '
+                    'Please remove it manually.')
+                raise TemplateRenderError(msg).with_traceback(e.__traceback__)
 
-        # Ensure that we operate on an absolute path for output_dir.
+    def render(self, output_dir: Path, overwrite_output_dir: bool) -> None:
+        """ Render the IP template into output_dir.
+
+        Generates the IP directory in a staging area and atomically move it
```
rswarbrick, line 282: Nit: To get the tense right, I think you want "moves" here.

util/ipgen/renderer.py
```diff
@@ -226,116 +232,144 @@ class IpBlockRenderer(IpTemplateRendererBase):
...
-            output_dir_existing_bak = output_dir.with_suffix(
-                '.bak~' + str(int(time.time())))
-            if do_overwrite:
-                os.rename(output_dir, output_dir_existing_bak)
+        if "module_instance_name" in self.ip_config.param_values:
+            hjson_base_name = (
+                self.ip_config.param_values["module_instance_name"]
+            )
+        else:
+            hjson_base_name = self.ip_template.name
```
rswarbrick, line 335: I think this is equivalent and a bit shorter:
```python
        hjson_base_name = self.ip_config.param_values.get('module_instance_name',
                                                          self.ip_template.name)
```

util/ipgen/README.md
matutem: It is conditional because this will only happen if the top is earlgrey.

util/topgen.py
matutem: It is intentional. There is no topgen invocation that uses it, and if someone really wanted to do it they can just overwrite the hjson with their alternate.

util/ipgen/renderer.py
matutem: Certainly, good catch.

rswarbrick:
Thanks for this! It looks good but we're failing lint because recent versions of `flake8` don't seem to notice uses through "# type: foo" comments. I'd probably suggest prefixing the PR with a commit that moves the types out of the comments.

matutem:
I inserted a commit before the other two that fixes the # type warnings.
PTAL

matutem:
Oops, there is more code in util that need cleaning up from templated ips. I'll make this a draft PR till ready.

matutem:
Ready to go, PTAL

rswarbrick (approved):
This looks great: thanks!

matutem:
Merging this since the changes cannot cause vivado failure: any RTL change would have been caught by ci/scripts/check-generated.sh.

PR #25132, opened 2024-11-14
[manuf] Verify manuf_state_creator consistency between ROM_EXT and OTP

anthonychen1251:
This adds the check to ensure the `manuf_state_creator` field in the ROM_EXT manifest matches the expected value that will be provisioned into the OTP. The ROM_EXT manifest is located in slot b and is flashed during the second bootstrap. The verification is done before the OTP is provisioned or the bootstrap is disabled, allowing for correction and re-running the personalization flow.

This PR partially addresses #24617

### DIFF
--- a/sw/device/silicon_creator/manuf/base/ft_personalize.c
+++ b/sw/device/silicon_creator/manuf/base/ft_personalize.c
@@ -32,6 +32,7 @@
 #include "sw/device/silicon_creator/lib/drivers/kmac.h"
 #include "sw/device/silicon_creator/lib/drivers/otp.h"
 #include "sw/device/silicon_creator/lib/error.h"
+#include "sw/device/silicon_creator/lib/manifest.h"
 #include "sw/device/silicon_creator/lib/otbn_boot_services.h"
 #include "sw/device/silicon_creator/manuf/base/perso_tlv_data.h"
 #include "sw/device/silicon_creator/manuf/base/personalize_ext.h"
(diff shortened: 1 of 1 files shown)


anthonychen1251:
The failed job `//sw/device/tests:i2c_target_test_fpga_cw310_sival` in the CI pipeline should be unrelated to this PR. I also spotted the same job failed in the other PRs.

sw/device/silicon_creator/manuf/base/ft_personalize.c
```diff
@@ -164,6 +165,27 @@ static void log_self_hash(void) {
...
+}
+
+extern const uint32_t kCreatorSwCfgManufStateValue;
+
+/*
+ * Check if the `manuf_state_creator` field in the ROM_EXT manifest (slot b)
+ * matches the expected value that will be provisioned into the OTP.
+ */
+static status_t check_manuf_state_creator_binding(void) {
+  HARDENED_CHECK_EQ(
```
timothytrippel, line 183: This does not need to a be a hardened check. It can just be an ordinary `TRY_CHECK(x == y)` https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/testing/test_framework/check.h;l=79?q=CHECK&ss=opentitan%2Fopentitan

anthonychen1251: Ok, I've also replaced the hardened check used in the `check_otp_measurement_{pre,post}_lock` with the ordinary check. This should be more informative when the mismatch occurs.

anthonychen1251: (LMK if you think a hardened check would be more suitable when checking against the locking status of OTP partition)

PR #25134, opened 2024-11-14
[SiVal, spi_host] Update execution environment of macronix test

engdoreis:
This will enable this test to run on the SiVal nightlies.


jwnrt (approved):

jwnrt:
Should we update the other `spi_host_*_flash_test`s in that file at the same time?

PR #25135, opened 2024-11-14
[pwm,dv] Exclude an unreachable item of block coverage

rswarbrick:
This cannot happen because of a constant wiring value.

~~Note: This PR depends on #25088. Once that is merged, this PR will only contain the last commit.~~ (No longer true)

### DIFF
--- a/hw/ip/pwm/dv/cov/manual_excl.vRefine
+++ b/hw/ip/pwm/dv/cov/manual_excl.vRefine
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<refinement-file-root>
+  <rules>
+    <!--
+        Waive block coverage for a conditional which is true if the 'we' signal is true for the
+        REGWEN in register in prim_subreg_arb. If the signal is true, this denotes a write to the
+        register from hardware but the instantiation in pwm_reg_top wires it to zero.
+    -->
+    <rule ccType="inst"
+          entityName="pwm/u_reg/u_regwen/wr_en_data_arb/gen_w0c/gen_non_mubi/2"
+          entityType="block"
+          line="135"
+          name="exclude"></rule>
+  </rules>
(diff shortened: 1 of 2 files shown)


rswarbrick:
I think the rejigged version of #25140 is a bit cleaner than the old version. It also no longer conflicts with this change! So I've squashed things together and the PR is now a clean single commit.

matutem:
This is specific to xcelium. It is troubling since our sign-off tool is VCS. Perhaps this hole doesn't show up in VCS so an exclusion is not needed there?

rswarbrick:
The signoff tool is defined by the block's sim_cfg hjson file. For `pwm`, it's xcelium.

matutem:
When reviewing the block for this I noticed the registers spec is missing regwens. I filed #25688 for this.

hw/ip/pwm/dv/cov/manual_excl.vRefine
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<refinement-file-root>
+  <rules>
+    <!--
+        Waive block coverage for a conditional which is true if the 'we' signal is true for the
+        REGWEN in register in prim_subreg_arb. If the signal is true, this denotes a write to the
+        register from hardware but the instantiation in pwm_reg_top wires it to zero.
```
matutem, line 7: Sorry Rupert, I don't understand this: in pwm_reg_top.sv the `we` input of `u_regwen` is connected to `regwen_we`, and from this comment I thought that is what you refer to. Could you clarify?

rswarbrick: I was busy writing a careful reply until I convinced myself that I'd got this wrong! What you write is completely correct. My suspicion is that the coverage hole was because the tests weren't actually writing to REGWEN at all.

Sorry for the noise!

PR #25136, opened 2024-11-14
[rom_ext,dice] Fix incorrect dice page id

sasdf:
There's one rebase conflict resolved incorrectly in #24980. It tries to load the CDI certs from the Factory page which is only for UDS cert. So the CDI certificate is always not found, making the rom_ext refreshing the cert chain on every boot.

#### Changes
* Specify flash pages explicitly, so each certificate's handler is independent without relying on previous states.
* Add the UDS certificate error message back so we can do e2e test with perso.
* An new e2e test is added to ensure we skip unnecessary update correctly.


timothytrippel:
I fixed both comments above and repushed this @sasdf . I will merge as soon as it passes CI. Thanks again!

PR #25142, opened 2024-11-14
[uart,dv] Mitigate dly from CSR reads in intr test

elliotb-lowrisc:
Two tweaks to avoid failures of `uart_intr_vseq`.

When testing RX Timeout does not occur when go close to but not exceeding the timeout, base delays off the last time the timeout was reset rather than a after performing additional CSR reads. Removing the added delay of the CSR reads avoids cases where it is enough to trigger the timeout.

Also, reorder interrupt checking so that pins are checked before CSR reads, seeing as the former are instant and the latter are not. Again, this is to avoid added delay that could allow the timeout to occur due to CSR read delays.


rswarbrick (approved):
This looks good to me, and thanks for the careful comments.

PR #25144, opened 2024-11-14
[tl,dv] Avoid silly hole in tl_a_chan_cov_cg

rswarbrick:
The cp_mask field expects to see a value of '1 (i.e. all byte lanes are enabled). Bizarrely, a full-width read doesn't match the coverpoint bin for some reason, yielding a hole in the cross at the bottom of the covergroup.

Calculate the "replicated ones" explicitly, which seems to solve the problem.


alees24 (approved):
I'm not very familiar with coverpoints still, but in normal (System)Verilog `{a}` makes 'a' a 'self-determined expression' so the '1' doesn't get widened. Whether it would work without the curly braces I know not, but surely the second pair is redundant in the revised expression?

Issue #23321, opened 2024-05-25
[sram_ctrl/dv] Adapt throughput tests for readback mode


nasahlpa:
### Description

With the introduction of the readback mode (c.f., #23212), the throughput tests need to be adapted to take the additional delay introduced by this feature into account. Currently, the readback feature is disabled for these tests.

vogelpi:
This seems doable without huge effort but the priority isn't very high as the feature can be disabled at run time.

vogelpi:
This doesn't seem like a priority to me. I suggest moving to M7.

PR #25145, opened 2024-11-14
[sram_ctrl,dv] Adapt throughput tests for readback feature

nasahlpa:
When the readback feature is enabled, reads and writes take longer. This commit adapts the cycle prediction to account for that delay.

Closes lowRISC/opentitan#23321

### DIFF
--- a/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
+++ b/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
@@ -26,6 +26,10 @@ class sram_ctrl_base_vseq #(
 
   int partial_access_pct = 10;
 
+  // Used by the throughput_w_readback test. When 1'b1, the readback feature is
+  // enabled at the beginning of the test.
+  bit init_w_readback = 1'b0;
+
   constraint readback_en_c {
     soft readback_en inside {MuBi4True, MuBi4False};
   }
@@ -34,6 +38,8 @@ class sram_ctrl_base_vseq #(
     super.pre_start();
     void'($value$plusargs("partial_access_pct=%0d", partial_access_pct));
     `DV_CHECK_LE(partial_access_pct, 100)
+    void'($value$plusargs("init_w_readback=%0d", init_w_readback));
+    `DV_CHECK_LE(init_w_readback, 1)
     // Wait for ram initialization to be done, since it blocks memory accesses.
     `DV_WAIT(cfg.in_init == 1'b0, "Timed out waiting for initialization done")
     // Make sure that the sram_readback_en task only runs once during the test.
@@ -51,11 +57,9 @@ class sram_ctrl_base_vseq #(
     if (!readback_running && do_readback_en) sram_readback_en();
   endtask
 
-  // Enable readback feature only for non-throughput and non-sec_cm tests. The
-  // readback feature is randomly initialized to on/off at the start of the test
-  // and randomly switched during the tests.
-  // TODO(#23321) Adapt the troughput tests to take the delay caused by the
-  // readback feature into account.
+  // Randomly disable or enable the readback feature during a test run.
+  // If a subclass wants to manually enable the readback feature, it should
+  // directly write to the readback enable register.
   virtual protected task sram_readback_en();
     readback_running = 1;
     if (uvm_re_match("*throughput*", get_type_name()) &&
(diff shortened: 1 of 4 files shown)


hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_throughput_vseq.sv
```diff
@@ -42,16 +55,32 @@ class sram_ctrl_throughput_vseq extends sram_ctrl_smoke_vseq;
 
       `uvm_info(`gfn, $sformatf("num_cycles: %0d, num_ops: %0d, num_partial_write: %0d",
                                 num_cycles, num_ops, num_partial_write), UVM_MEDIUM)
-
-      `DV_CHECK_EQ(num_cycles, num_ops + 1 + num_partial_write * 2);
+      if (init_w_readback) begin
+        // In throughput_w_readback test, if the last operation was a write, subtract
+        // one as the write already gets acknowledged over TLUL while the readback error
+        // still is doing the readback of the written value.
```
rswarbrick, line 61: Grammar nit: I think you want "is still doing" here.

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_throughput_vseq.sv
```diff
@@ -12,11 +12,20 @@ class sram_ctrl_throughput_vseq extends sram_ctrl_smoke_vseq;
 
   int num_partial_write;
 
+  int num_writes;
+  int num_reads;
+  int last_was_write;
```
rswarbrick, line 17: Can you add a documentation comment above this class variables? Probably just something like "Track the number of writes and reads and whether the most recent transaction was a write. This is updated in get_rand_mask."

hw/ip/sram_ctrl/dv/sram_ctrl_base_sim_cfg.hjson
```diff
@@ -108,6 +108,11 @@
       uvm_test_seq: sram_ctrl_throughput_vseq
       run_opts: ["+zero_delays=1 +partial_access_pct=20"]
     }
+    {
+      name: "{name}_throughput_w_readback"
+      uvm_test_seq: sram_ctrl_throughput_vseq
+      run_opts: ["+zero_delays=1 +partial_access_pct=0 +init_w_readback=1"]
```
rswarbrick, line 114: I wonder whether we should follow up with something that turns the list into the form `["+a=b", "+c=d"]`. I understand that the difference doesn't matter, but it feels a bit silly to be giving "arguments with spaces" :-)

nasahlpa: Either way works for me  :-) Happy to change that if you think it makes sense. If we want to do this, in a separate PR as this affects some other entries as well?

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_throughput_vseq.sv
nasahlpa: Thanks, I've corrected the sentence.

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_throughput_vseq.sv
nasahlpa: Done, thanks!

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
```diff
@@ -51,11 +57,10 @@ class sram_ctrl_base_vseq #(
...
 
-  // Enable readback feature only for non-throughput and non-sec_cm tests. The
-  // readback feature is randomly initialized to on/off at the start of the test
-  // and randomly switched during the tests.
-  // TODO(#23321) Adapt the troughput tests to take the delay caused by the
-  // readback feature into account.
+  // Randomly disable or enable the readback feature during a test run.
+  // The throughput and readback_err tests manually enable the readback feature
+  // at the beginning of the test as these tests require that the feature is
+  // enabled throughout the entire test.
```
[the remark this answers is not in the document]
nasahlpa: I've adapted the description accordingly.

hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_throughput_vseq.sv
rswarbrick: Thanks, but could the comment move up two lines, since it describes `num_writes` and `num_reads` too?

hw/ip/sram_ctrl/dv/sram_ctrl_base_sim_cfg.hjson
rswarbrick: Eh, definitely a follow-up if it's worth doing (and not in this PR). I'm not completely sure it is! :-D

rswarbrick (approved):
Brill, thanks!

Issue #25137, opened 2024-11-14
[rv_core_ibex] Enable SecureIbex config for FPGA


nasahlpa:
Currently, the `SecureIbex` configuration is disabled for CW310 & CW340 due to resource constraints. However, as `SecureIbex` is the default configuration (i.e., used in DV, Verilator, ASIC), we should also enable it on the FPGAs. Currently, we cannot test certain features that rely on this configuration (see #20119 for more details).

I checked, enabling `SecureIbex` on CW310 does not work due to resource constraints. However, I was able to build a bitstream for CW340 with the parameter enabled.

To close this issue, we should decide whether we want to have this parameter enabled by default on CW340. For this we should do an utilization analysis (e.g., are we still staying below a certain utilization level), run the existing tests to check whether enabling it does break something, and then create a PR with the enabled feature.

PR #2224, opened 2020-05-13
[sw, dif_plic] Remove redundant code and tidy up the code

silvestrst:
There is no need to calculate the multireg register count and register
width in bits, as these values are now auto generated.

MAX Priority is useful outside of the DIF, so it has been exposed to
the consumers of the DIF.

PR #25146, opened 2024-11-14
[rv_core_ibex, top_earlgrey] Enable SecureIbex for CW340

nasahlpa:
As the `SecureIbex` configuration is the one we use for the Earl Grey ASIC, we also should enable it for the FPGA boards.

Due to resource constraints, this is only possible for the CW340 but not the CW310.

**Report for CW340 with SecureIbex enabled:**
Resource utilization stays still well below 70%:
image

And timing is met:
image

**Report for CW340 with SecureIbex disabled:**
image
image

Closes lowRISC/opentitan#25137

### DIFF
--- a/hw/top_earlgrey/rtl/autogen/chip_earlgrey_cw310.sv
+++ b/hw/top_earlgrey/rtl/autogen/chip_earlgrey_cw310.sv
@@ -1079,9 +1079,9 @@ module chip_earlgrey_cw310 #(
     .KeymgrKmacEnMasking(0),
     .SecKmacCmdDelay(0),
     .SecKmacIdleAcceptSwMsg(1'b0),
+    .RvCoreIbexSecureIbex(0),
     .RomCtrlBootRomInitFile(BootRomInitFile),
     .RvCoreIbexRegFile(ibex_pkg::RegFileFPGA),
-    .RvCoreIbexSecureIbex(0),
     .SramCtrlMainInstrExec(1),
     .PinmuxAonTargetCfg(PinmuxTargetCfg)
   ) top_earlgrey (
--- a/hw/top_earlgrey/rtl/autogen/chip_earlgrey_cw340.sv
+++ b/hw/top_earlgrey/rtl/autogen/chip_earlgrey_cw340.sv
@@ -1063,9 +1063,9 @@ module chip_earlgrey_cw340 #(
     .KmacEnMasking(1),
     .KmacSwKeyMasked(1),
     .KeymgrKmacEnMasking(1),
+    .RvCoreIbexSecureIbex(1),
     .RomCtrlBootRomInitFile(BootRomInitFile),
     .RvCoreIbexRegFile(ibex_pkg::RegFileFPGA),
-    .RvCoreIbexSecureIbex(0),
     .SramCtrlMainInstrExec(1),
     .PinmuxAonTargetCfg(PinmuxTargetCfg)
   ) top_earlgrey (
--- a/hw/top_earlgrey/templates/chiplevel.sv.tpl
+++ b/hw/top_earlgrey/templates/chiplevel.sv.tpl
@@ -1148,6 +1148,7 @@ module chip_${top["name"]}_${target["name"]} #(
     .SecAesAllowForcingMasks(1'b1),
     .SecAesSkipPRNGReseeding(1'b1),
     .UsbdevStub(1'b1),
+    .RvCoreIbexSecureIbex(0),
 % else:
     .SecAesMasking(1'b0),
     .SecAesSBoxImpl(aes_pkg::SBoxImplLut),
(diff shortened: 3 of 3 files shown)


nasahlpa:
> Thanks @nasahlpa for looking into this. Do you know by how much the resource utilization increases roughtly?
> 
> The changes look mostly good, but there are FPGA CI failures and we should make sure all tests pass before merging this.

About that:

I get the same error on a local CW340 with the new bitstream:
`Command result: SFDP header contains incorrect signature: 0xffffffff`
as in CI.
I am a bit stuck here. Do you know how I can debug this?

nasahlpa:
> Thanks @nasahlpa for looking into this. Do you know by how much the resource utilization increases roughtly?
> 
> The changes look mostly good, but there are FPGA CI failures and we should make sure all tests pass before merging this.

I've updated the PR description with area and timing numbers before and after this PR.

vogelpi:
> > Thanks @nasahlpa for looking into this. Do you know by how much the resource utilization increases roughtly?
> > The changes look mostly good, but there are FPGA CI failures and we should make sure all tests pass before merging this.
> 
> About that:
> 
> I get the same error on a local CW340 with the new bitstream: `Command result: SFDP header contains incorrect signature: 0xffffffff` as in CI. I am a bit stuck here. Do you know how I can debug this?

Thanks for adding the area and timing numbers. I think the change is reasonable for the big board. About the SFDP header issue, I have no idea abou this. Maybe someone from the software team knows?

a-will:
> > > Thanks @nasahlpa for looking into this. Do you know by how much the resource utilization increases roughtly?
> > > The changes look mostly good, but there are FPGA CI failures and we should make sure all tests pass before merging this.
> > 
> > 
> > About that:
> > I get the same error on a local CW340 with the new bitstream: `Command result: SFDP header contains incorrect signature: 0xffffffff` as in CI. I am a bit stuck here. Do you know how I can debug this?
> 
> Thanks for adding the area and timing numbers. I think the change is reasonable for the big board. About the SFDP header issue, I have no idea abou this. Maybe someone from the software team knows?

That likely means bootstrapping failed at the very first step, and the chip gave no response to SPI. MISO was left high when the tester tried to read the SFDP table.

nasahlpa:
> @nasahlpa what test failed? Can't find the older CI logs and still awaiting the run for the newer one.
> 
> At a guess the bootloader has immediately fallen over because we get an instant alert from Ibex, maybe because the lockstep hasn't come out of reset properly? There's some prims involved that may have FPGA specific implementations that could explain the behaviour difference.

All of them. Not a single test on CW340 passed. Here are the [old](https://dev.azure.com/lowrisc/opentitan/_build/results?buildId=161570&view=results) results from CI.

nasahlpa:
I investigated this and discovered the following:

When enabling the `SecureIbex` paramter for CW340, a major Ibex alert is triggered by a register file ECC [check]([url](https://github.com/lowRISC/opentitan/blob/651fd2fcffeac52680bb1946db5d4a05da51326d/hw/vendor/lowrisc_ibex/rtl/ibex_core.sv#L926)). Hence, the cores does not boot up and we cannot bootstrap software.

When either setting the `RegFileECC = 1'b0` in the [ibex_top](https://github.com/lowRISC/opentitan/blob/651fd2fcffeac52680bb1946db5d4a05da51326d/hw/vendor/lowrisc_ibex/rtl/ibex_top.sv#L151) module or forcing the ECC error [signal](https://github.com/lowRISC/opentitan/blob/651fd2fcffeac52680bb1946db5d4a05da51326d/hw/vendor/lowrisc_ibex/rtl/ibex_core.sv#L926) to 0, the core boots up and I can successfully run SW on CW340 with the `SecureIbex` parameter on.

My assumption is that the `ibex_register_file_fpga` module does not correctly work with `RegFileECC = 1'b1`. I had a quick look into the FPGA register file - it though seems that we are correctly initializing the RF content with the ECC encoded 0.

In the long term, we should investigate this and fix it.

As a temporary solution, we could bypass the error:

- Disable the `RegFileECC` option.
  - Requires decoupling this option from the `SecureIbex` parameter, i.e., changes in the Ibex repository as well as in `top_earlgrey` as well as `chip_earlgrey_cw340` in the OpenTitan repository are needed.
- Use the FF register file.
  - When the `SecureIbex` parameter is enabled for CW340, use the FF instead of the FPGA register file
  - Probably the easiest solution for now as we only need to modify `chip_earlgrey_cw340` in the OpenTitan repository.

I tested both solutions and they work.

WDYT @GregAC?

GregAC:
@nasahlpa nice work on diagnosing the issue. I'd favour just using the FF register file in FPGA, assuming the FPGA implementation still works.

The reason to use the FPGA RF is efficient resource utilization I don't think switching to the FF version will have a meaningful effect given the full earl grey design is huge.

We should work out why the FPGA version is broken in this instance. I'm happy to take a look at this.

And having written that just seen the edit! If that's the quick fix then great go for it, otherwise switch to FF version would be my vote.

nasahlpa:
The error was indeed the zero value for `x0`.
This will be fixed once [#2224](https://github.com/lowRISC/ibex/pull/2224) in the Ibex repository is merged.

Thanks all for helping me debugging this issue.

hw/vendor/lowrisc_ibex/rtl/ibex_top.sv
```diff
@@ -148,7 +148,7 @@ module ibex_top import ibex_pkg::*; #(
   localparam bit          Lockstep              = SecureIbex;
   localparam bit          ResetAll              = Lockstep;
   localparam bit          DummyInstructions     = SecureIbex;
-  localparam bit          RegFileECC            = SecureIbex;
+  localparam bit          RegFileECC            = 1'b0;
```
vogelpi, line 151: This change should be removed before merging this.

vogelpi:
This got accidentally closed again by someone merging a related commit in the base repo. @nasahlpa , the reason is the "fixes #pr-no" in the commit message. I think it will happen again :-(

nasahlpa:
Issue is resolved (https://github.com/lowRISC/ibex/pull/2228) and CW340 with enabled `SecureIbex` configuration now successfully builds & runs test. Hence, I am merging this now

PR #25147, opened 2024-11-14
[rom_ext] Implement Integrator Specific FW Binding logic.

moidx:
Implements Integrator Specific FW Binding functionality to provide anti-rollback and product binding functionality to application firmware.

The following changes are included:

1. Add flash_ctrl helper function to generate the info flash page configuration based on the ownership ISFB config.
2. Implement isfb logic as a separate module to simplify ROM_EXT integration.
3. Add unittest cases to verify anti-rollback and product association functionality.
4. Add minimum level of hardening against FI.

The ROM_EXT integration and e2e test cases will be implemented in a follow up commit.


sw/device/silicon_creator/lib/ownership/isfb.h
```diff
@@ -0,0 +1,51 @@
...
+#include "sw/device/silicon_creator/lib/error.h"
+#include "sw/device/silicon_creator/lib/manifest.h"
+#include "sw/device/silicon_creator/lib/ownership/owner_block.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Processes the ISFB boot request.
```
vbendeb, line 19: Pardon my ignorance, but what is ISFB. Please explain it here and maybe in the patch description too.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,120 @@
...
+      (kExpectedStrikeBitCount * kFlashWordSize) / CHAR_BIT,
+  kExpectedStrikeRegionWordsCount =
+      kExpectedStrikeRegionBytesCount / sizeof(uint32_t),
+};
+
+rom_error_t isfb_boot_request_process(const manifest_ext_isfb_t *ext,
+                                      const owner_config_t *owner_config,
+                                      uint32_t *checks_performed_count) {
+  *checks_performed_count = UINT32_MAX;
+  if (ext->header.name == kManifestExtIdIsfb &&
```
vbendeb, line 28: what other values `ext->header.name` is allowed to have here?

shouldn't the check be

`(ext->header.name != kManifestExtIdIsfb || (hardened_bool_t)owner_config->isfb == kHardenedBoolFalse)` ?

vbendeb (commented):
some questions, maybe I'm missing something. 

It would be good to have a one pager explaining the design...

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,120 @@
...
+#include "sw/device/silicon_creator/lib/error.h"
+#include "sw/device/silicon_creator/lib/manifest.h"
+#include "sw/device/silicon_creator/lib/ownership/owner_block.h"
+
+enum {
+  kFlashWordSize = 64,
+
+  // The number of bits in the strike mask is fixed to 128. Each bit in the
+  // strike mask corresponds to a flash word.
+  kExpectedStrikeBitCount = 128,
```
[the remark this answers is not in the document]
moidx: 128 bits are mapped to 64bit words, so this should be a total of 1024 bytes. Both strikes and product mappings are captured in a single info flash page.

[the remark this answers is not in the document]
vbendeb: ah, my mistake - I counted bits, not bytes :P

sw/device/silicon_creator/lib/ownership/isfb.c
moidx: > what other values ext->header.name is allowed to have here?

All manifest extensions use the same header. The previous check was incorrect. Updated to match your recommendation. 

Done

sw/device/silicon_creator/lib/ownership/isfb.h
moidx: Updated commit description as well as stringdoc. 

Done.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,120 @@
...
+    return kErrorOwnershipISFBProductExpCnt;
+  }
+
+  flash_ctrl_perms_t perm = {
+      .read = kMultiBitBool4True,
+      .write = kMultiBitBool4False,
+      .erase = kMultiBitBool4False,
+  };
+  flash_ctrl_cfg_t cfg = {
+      .scrambling = kMultiBitBool4True,
```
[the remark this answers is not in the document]
moidx: Agreed. Fixed.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,120 @@
...
+    return kErrorOwnershipISFBNotPresent;
+  }
+  const owner_isfb_config_t *isfb = owner_config->isfb;
+  if (ext->product_expr_count * 2 > isfb->product_words) {
+    return kErrorOwnershipISFBProductExpCnt;
+  }
+
+  flash_ctrl_perms_t perm = {
+      .read = kMultiBitBool4True,
+      .write = kMultiBitBool4False,
```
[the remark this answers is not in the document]
moidx: Done. And yes, will submit the erase policy as a follow up change.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,120 @@
...
+      .erase = kMultiBitBool4False,
+  };
+  flash_ctrl_cfg_t cfg = {
+      .scrambling = kMultiBitBool4True,
+      .ecc = kMultiBitBool4True,
+      .he = kMultiBitBool4True,
```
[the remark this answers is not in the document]
moidx: I agree with you that this probably doesn't need high endurance. 

The tradeoff in this particular case is faster erase times in exchange for a lower number of erase cycles. I can give you more details offline.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,120 @@
...
+  HARDENED_RETURN_IF_ERROR(flash_ctrl_info_read(&isfb_info_page, /*offset=*/0,
+                                                ARRAYSIZE(data), data));
+
+  uint32_t strike_cnt_ok = 0;
+  uint32_t strike_cnt_bad = 0;
+  uint64_t *strike_word = (uint64_t *)data;
+  size_t i;
+  for (i = 0; i < ARRAYSIZE(data) / 2; ++i, ++strike_word) {
+    uint32_t strike_bit = ext->strike_mask[i >> 5] & (1 << (i & 31));
+    if (launder32(strike_bit) && *strike_word == 0) {
```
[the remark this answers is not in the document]
moidx: SGTM. Done.

sw/device/silicon_creator/lib/error.h
```diff
@@ -216,6 +216,11 @@ enum module_ {
   X(kErrorOwnershipInvalidDin,        ERROR_(13, kModuleOwnership, kInvalidArgument)), \
   X(kErrorOwnershipUnlockDenied,      ERROR_(14, kModuleOwnership, kPermissionDenied)), \
   X(kErrorOwnershipFlashConfigRomExt, ERROR_(15, kModuleOwnership, kInvalidArgument)), \
+  X(kErrorOwnershipISFBNotPresent,    ERROR_(16, kModuleOwnership, kNotFound)), \
+  X(kErrorOwnershipISFBProductExpCnt, ERROR_(17, kModuleOwnership, kOutOfRange)), \
+  X(kErrorOwnershipISFBStrikeMask,    ERROR_(18, kModuleOwnership, kInvalidArgument)), \
```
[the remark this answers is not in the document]
moidx: Done.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,120 @@
...
+    return kErrorOwnershipISFBStrikeMask;
+  }
+
+  // Read the product expressions from the ISFB info page.
+  HARDENED_RETURN_IF_ERROR(flash_ctrl_info_read(
+      &isfb_info_page, /*offset=*/1024, ARRAYSIZE(data), data));
+
+  uint32_t pexpr_cnt_ok = 0;
+  uint32_t pexpr_cnt_bad = 0;
+  for (i = 0; i < ext->product_expr_count; ++i) {
```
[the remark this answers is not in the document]
moidx: Done. Added a check for this at the beginning of the function.

sw/device/silicon_creator/lib/ownership/isfb.c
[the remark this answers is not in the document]
moidx: Updated the code to support `uint32_t` strike words. The maximum number of strikes is now `256`.

sw/device/silicon_creator/lib/manifest.h
```diff
@@ -494,14 +507,14 @@ typedef struct manifest_ext_isfb {
   /**
    * Strikeout mask.
    *
-   * The size of this array is equal to 128 strike bits packed into uint32_t
-   * words. Each strike bit corresponds to a flash word in the ISFB info flash
-   * page.
+   * The size of this array is equal to 256 strike bits packed into uint32_t
```
jettr, line 510: Can we keep the number of strike out bits at 128 (that should be all we need). I would like to keep the other space free for now.

Layout of info page would be
 * 0 - 512: Owner firmware rollback constraint section (128 4-byte "bits")
 * 512- 1024: Unused and available to owner firmware
 * 1024 - (1024 + owner-defind limit): product expression constraints
 * (1024 + owner-defind limit) - 2048: Unused and available to owner firmware

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,133 @@
...
+  kExpectedStrikeRegionWordsCount =
+      kExpectedStrikeRegionBytesCount / sizeof(uint32_t),
+};
+
+rom_error_t isfb_boot_request_process(const manifest_ext_isfb_t *ext,
+                                      const owner_config_t *owner_config,
+                                      uint32_t *checks_performed_count) {
+  *checks_performed_count = UINT32_MAX;
+  if (ext->header.name != kManifestExtIdIsfb ||
+      (hardened_bool_t)owner_config->isfb == kHardenedBoolFalse) {
```
jettr, line 35: I tried, but I cannot find the definition of `owner_isfb_config_t` any where. Can you point it out to me please?

Also why do we cast `owner_config->isfb` to `hardened_bool_t` then later to `owner_isfb_config_t`? That doesn't seem valid.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,133 @@
...
+  }
+  // Check loop completion and count consistency.
+  HARDENED_CHECK_EQ(strike_cnt_ok + strike_cnt_bad + i,
+                    kExpectedStrikeBitCount * 2);
+
+  if (launder32(strike_cnt_bad) > 0) {
+    return kErrorOwnershipISFBStrikeMask;
+  }
+
+  static_assert(sizeof(data) == KProductExprMaxCountBytes,
```
jettr, line 96: This doesn't seem like the correct comparisons. The data should be the size of the rollback space not the max product space (even though they happen to be the same right now)

sw/device/silicon_creator/lib/ownership/isfb.c
moidx: You can find the data structures in the ownership `datatypes.h` header in the `earlgrey_es_sival` branch. This will be merged to `earlgrey_1.0.0` once the feature is complete. 

Also take a look at `owner_config_default()` inside `owner_block.c`. Owner config entries are initialized to `kHardenedBoolFalse` to flag when an owner TLV entry is not present in the ownership blob.

sw/device/silicon_creator/lib/ownership/isfb.c
moidx: There is a separate check for the strikes region above. I will update this check again once we have the updated memory map you recommended in a separate comment.

sw/device/silicon_creator/lib/manifest.h
moidx: Updated to match suggested info page layout.

sw/device/silicon_creator/lib/ownership/isfb.c
moidx: This check is no longer here as I updated the info flash read to load the entire page.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,147 @@
...
+  flash_ctrl_info_cfg_set(&isfb_info_page, cfg);
+
+  // There are in total 128 bits in the strike mask. Each bit corresponds to a
+  // flash word. Each flash word is 64 bits.
+  uint32_t data[kFlashPageWords];
+  static_assert(sizeof(ext->strike_mask) * CHAR_BIT == kExpectedStrikeBitCount,
+                "Strike mask size mismatch");
+  static_assert(sizeof(data) >= kFlashPageBytes, "Unexpected data size");
+
+  HARDENED_RETURN_IF_ERROR(flash_ctrl_info_read(&isfb_info_page, /*offset=*/0,
```
jettr, line 88: What happens within `flash_ctrl_info_read` is there is an ECC failure? Does the chip reset or does it return an error?

How would we recovery this chip in a lab if that every occurred? We could make an owner firmware that has no restrictions?

jettr (commented):
LGTM, the only case to think about is the ECC failure recovery process

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,147 @@
...
+                                                ARRAYSIZE(data), data));
+
+  uint32_t strike_cnt_ok = 0;
+  uint32_t strike_cnt_bad = 0;
+  uint32_t *strike_word = &data[kIsfbWordOffsetStartStrikeRegion];
```
cfrantz, line 93: Since the strike words correspond to flash words, should this be `uint64_t` instead? (and use UINT64_MAX in the tests below)?

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,147 @@
...
+  flash_ctrl_perms_t perm = {
+      .read = kMultiBitBool4True,
+      .write = kMultiBitBool4True,
+      .erase = kMultiBitBool4False,
+  };
+  flash_ctrl_cfg_t cfg = {
+      .scrambling = kMultiBitBool4False,
+      .ecc = kMultiBitBool4True,
+      .he = kMultiBitBool4False,
+  };
```
cfrantz, line 73: Do we need to configure this here?  I'd imagined that the normal owner FlashInfoConfig block would handle the info page properties wanted.

sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -143,6 +144,12 @@ rom_error_t owner_block_parse(const owner_block_t *block,
           return kErrorOwnershipDuplicateItem;
         config->rescue = (const owner_rescue_config_t *)item;
         break;
+      case kTlvTagIntegrationSpecificFirmwareBinding:
+        HARDENED_CHECK_EQ(tag, kTlvTagIntegrationSpecificFirmwareBinding);
+        if ((hardened_bool_t)config->isfb != kHardenedBoolFalse)
+          return kErrorOwnershipDuplicateItem;
+        config->isfb = (const owner_isfb_config_t *)item;
```
cfrantz, line 151: We need to sanity check the ISFB configuration.  Notably, we need to ensure the selected flash bank & page are within the range designated as owner info pages.  There is a helper function `is_owner_page` below in this file.  It's currently tailored to the info config struct, but it could be made more generic.

sw/device/silicon_creator/lib/ownership/isfb_unittest.cc
```diff
@@ -0,0 +1,229 @@
...
+    kExpectedStrikeCount = 2,
+  };
+
+  // Sanity check to ensure that the strike mask only has the first word struck.
+  // Plase update the test if the strike mask is changed.
+  EXPECT_EQ(kManifestExtGood.strike_mask[0], 0xfffffffe);
+
+  // The state of device is such that the number of strikes is larger than what
+  // is defined in the manifest extension. This must result in an anti-rollback
+  // failure whic is mapped to the `kErrorOwnershipISFBStrikeMask` error code.
```
cfrantz, line 194: "which"

sw/device/silicon_creator/lib/ownership/isfb.c
moidx: Updated the documentation to be consistent with the implementation. We switched to 32bit strike words to compress the anti-rollback region. 

https://github.com/lowRISC/opentitan/pull/25147#discussion_r1844160952

sw/device/silicon_creator/lib/ownership/isfb.c
moidx: Removed. I will take care of orchestrating the right  order of operations when I connect this in the `rom_ext`.

sw/device/silicon_creator/lib/ownership/isfb_unittest.cc
moidx: Fixed.

sw/device/silicon_creator/lib/ownership/owner_block.c
moidx: Added sanity check function. Also added unittests to the `owner_block_unittest` target. I ended up finding a bug in `opentitanlib` which has now been fixed in a separate  commit included in this PR.

sw/device/silicon_creator/lib/ownership/isfb_unittest.cc
```diff
@@ -0,0 +1,212 @@
...
+            .name = kManifestExtIdIsfb,
+        },
+    .strike_mask =
+        {
+            // Only the first word is struck.
+            0xfffffffe,
+            0xffffffff,
+            0xffffffff,
+            0xffffffff,
+        },
```
cfrantz, line 87: Since strike mask is now 256 bits in size, you need 4 more values here.  I think the opentitantool manifest handling currently uses  a `u128` to represent this value (should be a 256-bit value).

I _think_ @jettr would like to represent this with the highest order bits going to zero first:
```
no strikes: 0xFFFFFFFF.......FF
1 strike:   0x7FFFFFFF.......FF
2 strikes:  0x3FFFFFFF.......FF
3 strikes:  0x1FFFFFFF.......FF
(etc)
```

I think in order to support that ordering of bits, the expression in isfb.c:66 would have to change to something like this:
```
uint32_t strike_bit = ext->strike_mask[i % 32] & (1 << (31 - i % 32));
```
In addition, we'd also have to take a close look at how the manifest handling packs the strike value into the manifest extension.

sw/device/silicon_creator/lib/ownership/isfb.c
```diff
@@ -0,0 +1,122 @@
...
+  if (ext->product_expr_count * 2 > isfb->product_words ||
+      ext->product_expr_count > kProductExprMaxCount) {
+    return kErrorOwnershipISFBProductExpCnt;
+  }
+
+  // The ISFB info page configuration is handled in the owner's
+  // `FlashInfoConfig` settings.
+  flash_ctrl_info_page_t isfb_info_page;
+  HARDENED_RETURN_IF_ERROR(flash_ctrl_info_type0_params_build(
+      isfb->bank, isfb->page, &isfb_info_page));
```
cfrantz, line 48: We should probably have a check somewhere to make sure that the owner FlashInfoConfig exists and has a configuration for this bank and page.

sw/device/silicon_creator/lib/ownership/isfb_unittest.cc
jettr: It is our intent to strike the bits starting from the lowest memory, where each bit is half of a flash word.

If the owner firmware strike section in INFO flash starts as follows (using `u32` words)

```0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF```

Then the first strike would change the INFO page to

```0x00000000 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF```

Then the next would be

```0x00000000 0x00000000 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF 0xFFFFFFFF```

This write pattern works with ECC enabled, but scramble disabled on that INFO page.

If we were to represent the INFO page in `u64` words, the write pattern would look like

```0xFFFFFFFFFFFFFFFF 0xFFFFFFFFFFFFFFFF 0xFFFFFFFFFFFFFFFF```
```0xFFFFFFFF00000000 0xFFFFFFFFFFFFFFFF 0xFFFFFFFFFFFFFFFF```
```0x0000000000000000 0xFFFFFFFFFFFFFFFF 0xFFFFFFFFFFFFFFFF```

Also we only plan on using 128 bits, but having more isn't a problem for us (as long as we don't have to use them or reserve space for them)

Comment (cfrantz) [reply sw/device/silicon_creator/lib/ownership/isfb_unittest.cc:87]: My concern was with how the strike mask is represented in the manifest, where each bit represents a strike word.  In that representation, I thought you wanted the most significant bit to represent strike word 0, and the next bit to represent strike word 1 and so on.

In that representation (a single u128), the strike mask values as follows, correct?
```
0 strikes: 0xFFFFFFFF_FFFFFFFF_FFFFFFFF_FFFFFFFF
1 strike:  0x7FFFFFFF_FFFFFFFF_FFFFFFFF_FFFFFFFF
2 strikes: 0x3FFFFFFF_FFFFFFFF_FFFFFFFF_FFFFFFFF
3 strikes: 0x1FFFFFFF_FFFFFFFF_FFFFFFFF_FFFFFFFF
4 strikes: 0x0FFFFFFF_FFFFFFFF_FFFFFFFF_FFFFFFFF
```

Comment (jettr) [reply sw/device/silicon_creator/lib/ownership/isfb_unittest.cc:87]: Ah, I see.. I like the mask value to match the layout in flash, so LE bits would make the most sense to me.

To make it explicit for discussion, a flash layout (in `u64` words) with increasing memory address:

`0xFFFFFFFF00000000 0xFFFFFFFFFFFFFFFF 0xFFFFFFFFFFFFFFFF` 

would correspond to logical `u128` strike value of `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE` where the least significant bit of the `u128` is `0`.

---

It is possible in the past I asked for BE versus LE representation as the current rollback scheme for our existing chip uses a LE strike pattern for owner firmware, yet a BE strike pattern for the bootloader stage. Depending on what I was looking most recently then, I could have confused myself.

File: sw/device/silicon_creator/lib/ownership/owner_block.c [hunk 20]
```diff
@@ -370,7 +395,7 @@ rom_error_t owner_block_info_apply(const owner_flash_info_config_t *info) {
   const owner_info_page_t *config = info->config;
   uint32_t crypt = 0;
   for (size_t i = 0; i < len; ++i, ++config, crypt += 0x11111111) {
-    if (is_owner_page(config) == kHardenedBoolTrue) {
+    if (is_owner_page(config->bank, config->page) == kHardenedBoolTrue) {
```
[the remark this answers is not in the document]
moidx: Done. Also fixed some issues in the `flash_ctrl_info_type0_params_build` function.

sw/device/silicon_creator/lib/ownership/isfb_unittest.cc
moidx: Updated the max number of strikes to 128 so that we can keep the u128 implementation in the manifest, aligning with the requirement described by @jettr. Also checked the following flash and bit strike layout.

Flash layout:

```
000: 0xFFFFFFFF00000000 
008: 0xFFFFFFFFFFFFFFFF 
010: 0xFFFFFFFFFFFFFFFF
```

Corresponding to strike pattern:

```
0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE --> least significant bit is 0.
```

sw/device/silicon_creator/lib/ownership/isfb.c
moidx: Let me add this check in a follow up PR. I need to think about how to implement checks that rely on other `TLV` entries.

cfrantz: Ack.

cfrantz (approved):

PR #25148, opened 2024-11-14
[opentitantool] Multi-byte read()/write() in serial console

jesultra:
The serial console library suffers from the need not to consume any additional bytes, if one of the regular expressions matches, for which we have found no better solution than to read a single byte at a time.

At Google, we often use `opentitantool console` and pipe the output into other test scripts (which may do regular expressions).  In such cases, the opentitanlib is not performing any regular expression matching, but it still reading a single byte at a time, and writing each byte using a separate `write()` operating system call.  This is highly inefficient, when the data goes through anonymous pipes, and is often further processed just one or a few bytes at a time by other processes.

This PR modifies the logic in `uart/console.rs` to recognize the case of there not being any regular expressions to match, and in that case read up to 1024 bytes at a time from the transport USB device.

I plan on combining this with below enhancements to HyperDebug firmware, to make it preferably compose USB packets with multiple characters captured on its UARTs.

https://chromium-review.googlesource.com/c/chromiumos/platform/ec/+/6020315

### DIFF
--- a/sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
+++ b/sw/host/opentitanlib/src/transport/hyperdebug/mod.rs
@@ -450,6 +450,9 @@ pub struct Conn {
 impl MaintainConnection for Conn {}
 
 impl Inner {
+    /// General timeout for response on the HyperDebug text-based USB command console.
+    const COMMAND_TIMEOUT: Duration = Duration::from_millis(3000);
+
     /// Establish connection with HyperDebug console USB interface.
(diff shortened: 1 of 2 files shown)


pamaury (approved):
This looks good to me but I think either @nbdd0121 or @jwnrt who know the serial console better than me should have a look.

jwnrt (commented):
Thanks, I think this is sensible

jesultra:
Backporting to earlgrey_es_1.0.0 is currently blocked by this other cherry-pick:
https://github.com/lowRISC/opentitan/pull/25498

PR #25149, opened 2024-11-14
[build] Introduce OTP emulation images.

moidx:
This change is in preparation for managing the SiVal silicon configuration using a separate OTP scaffolding. The `//sw/ip/otp_ctrl/data/earlgrey_skus/emulation` targets will contain OTP images used for simulation as well as FPGA emulation. FPGA emulation images will include both pre-silicon and post-silicon use cases.

Targets under `//sw/ip/otp_ctrl/data/earlgrey_skus/sival` will be used strictly for configuring silicon for Silicon Validation flows. We refer to this as the SiVal SKU.

The SiVal SKU will contain secure boot public keys backed by code-signing infrastructure as opposed to the the `emulation` configuration which will continue to use the test (i.e. fake) keys stored in the repository.


cfrantz (approved):

hw/ip/otp_ctrl/data/earlgrey_skus/emulation/BUILD
```diff
@@ -0,0 +1,371 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# SKU: SiVal Bringup. ASCII code `SV00`.
```
pamaury, line 5: Is this comment applicable to emulation targets?

timothytrippel:
Now we have the emulation OTP config, we can probably cleanup the extra images in sival: https://cs.opensource.google/opentitan/opentitan/+/earlgrey_1.0.0:hw/ip/otp_ctrl/data/earlgrey_skus/sival/BUILD

We only need the sw cfg partition and key partition definitions and the c_consts targets. All the extra in-between targets are unnecessary I think.

PR #25151, opened 2024-11-15
[provisioning] add initial orchestrator script for benchtop provisioning

timothytrippel:
This PR contains:
1. several bug fixes / cosmetic improvements to the CP and FT provisioning flows, and
2. and initial version of the new and improved orchestrator.py provisioning script to support A1 benchtop provisioning flows.

### DIFF
--- a/sw/host/opentitanlib/src/io/jtag.rs
+++ b/sw/host/opentitanlib/src/io/jtag.rs
@@ -23,6 +23,9 @@ pub struct JtagParams {
 
     #[arg(long, default_value = "1000")]
     pub adapter_speed_khz: u64,
+
+    #[arg(long, default_value = "false")]
+    pub log_stdio: bool,
 }
 
 impl JtagParams {
--- a/sw/host/provisioning/orchestrator/src/device_id.py
+++ b/sw/host/provisioning/orchestrator/src/device_id.py
@@ -1,0 +1,158 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+"""Module for computing OpenTitan device IDs."""
+
+import binascii
+import struct
+from dataclasses import dataclass
+
+from sku_config import SkuConfig
+from util import bytes_to_int, format_hex
+
+_RESERVED_WORD = 0
+
+
+@dataclass
+class DeviceIdentificationNumber:
+    """Class for storing Device Identification Number portion of Device ID."""
+    year: int = 0  # valid range: [0, 9]
+    week: int = 0  # valid range: [0, 51]
+    lot: int = 0  # valid range: [0, 999]
+    wafer: int = 0  # valid range: [0,99]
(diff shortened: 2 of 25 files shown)


sw/host/opentitanlib/src/io/jtag.rs
```diff
@@ -23,6 +23,9 @@ pub struct JtagParams {
 
     #[arg(long, default_value = "1000")]
     pub adapter_speed_khz: u64,
+
+    #[arg(long, default_value = "false")]
```
pamaury, line 27: I am not sure about the default value. We should definitely consider whether this is the right value for JTAG tests to enable debugging in CI. Maybe we want to explicitely enable logging in those tests. @nbdd0121 might have some opinion.

timothytrippel: Do we need the logs in CI? if a test is failing, it seems you would want to check for reproducibility locally, in which case you would enable the flag locally. Silencing the verbosity of the openocd logs in CI should save CI runtime to (admittedly on a little thought).

sw/host/provisioning/orchestrator/src/device_id.py
```diff
@@ -0,0 +1,158 @@
...
+        # - Week
+        # - Lot #
+        # - Wafer #
+        # - Wafer X coord.
+        # - Wafer Y coord.
+        self.din = din
+
+        # Build CRC32 over HW origin + DIN.
+        self.crc32 = binascii.crc32(
+            struct.pack("<IQ", self._hw_origin, self.din.to_int()))
```
cfrantz, line 108: Did we agree to do away with the CRC32 in the device id since the OTP partition already has an integrity check?

sw/host/provisioning/orchestrator/src/ot_dut.py
```diff
@@ -0,0 +1,181 @@
...
+            host_flags = host_flags.format("teacup")
+            host_flags += " --disable-dft-on-reset"
+            individ_elf = individ_elf.format(self.sku_config.name,
+                                             "silicon_creator")
+            perso_bin = perso_bin.format(self.sku_config.name,
+                                         "silicon_creator")
+            fw_bundle_bin = fw_bundle_bin.format(self.sku_config.name,
+                                                 "silicon_creator")
+
+        # Assemble bazel command.
```
cfrantz, line 142: Is this actually assembling a bazel command, or is this assembling a command line for the host `ft` binary?

cfrantz (approved):

sw/host/provisioning/orchestrator/src/ot_dut.py
timothytrippel: ah good catch. just a command. will fix.

sw/host/provisioning/orchestrator/src/device_id.py
timothytrippel: Fixed in #25167.

sw/host/provisioning/orchestrator/src/ot_dut.py
timothytrippel: Fixed in #25167.

PR #25152, opened 2024-11-15
[cherry-pick] manual cherry-pick of bug fix commits from #25151

timothytrippel:
Some bug and cosmetic fixes to the CP/FT provisioning flows were fixed in #25151. This cherry-picks them to the `master` branch.

### DIFF
--- a/sw/device/silicon_creator/manuf/base/BUILD
+++ b/sw/device/silicon_creator/manuf/base/BUILD
@@ -378,6 +378,19 @@ _FT_PROVISIONING_HARNESS = "//sw/host/provisioning/ft:ft_{}"
     for sku, config in EARLGREY_SKUS.items()
 ]
 
+filegroup(
+    name = "ft_personalize_all",
+    testonly = True,
+    srcs = [
+               ":ft_personalize_{}".format(sku)
+               for sku in EARLGREY_SKUS.keys()
+           ] +
+           [
+               ":ft_fw_bundle_{}".format(sku)
+               for sku in EARLGREY_SKUS.keys()
+           ],
+)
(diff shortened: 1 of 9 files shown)


sw/host/provisioning/ft/BUILD
```diff
@@ -38,3 +28,12 @@ package(default_visibility = ["//visibility:public"])
     )
     for sku in EARLGREY_SKUS.keys()
 ]
+
+filegroup(
+    name = "ft_all",
+    testonly = True,
```
cfrantz, line 34: Why testonly?  Because the srcs are testonly, but again, why?

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -378,6 +378,19 @@ _FT_PROVISIONING_HARNESS = "//sw/host/provisioning/ft:ft_{}"
     for sku, config in EARLGREY_SKUS.items()
 ]
 
+filegroup(
+    name = "ft_personalize_all",
+    testonly = True,
```
cfrantz, line 383: Does this need to be testonly?

cfrantz (approved):

sw/host/provisioning/ft/BUILD
timothytrippel: It is because all the FPGA exec_envs are marked as "testonly" so it forces this downstream.

sw/device/silicon_creator/manuf/base/BUILD
timothytrippel: same here.

PR #25157, opened 2024-11-15
[hw,keymgr_dpe,rtl] Remove flash seed input as not used currently

Razer6:
The flash seed input pulls in the flash_ctrl_pkg, which is ip-gened. Darjeeling, who is the only user of keymgr_dpe by now, does not have a flash_ctrl and so it does not use the flash seed input. To remove the dependency of flash_ctrl, remove the flash seed input from keymgr_dpe

### DIFF
--- a/hw/ip/keymgr_dpe/data/keymgr_dpe.hjson
+++ b/hw/ip/keymgr_dpe/data/keymgr_dpe.hjson
@@ -89,12 +89,6 @@
       act:     "rcv",
       package: "otp_ctrl_pkg",
     },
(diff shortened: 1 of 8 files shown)


andreaskurth (approved):
As keymgr_dpe doesn't use the flash interface, removing it LGTM as immediate solution for Darjeeling. In the longer term, I propose the solution tracked in the new issue #25159.

andreaskurth:
Skipping remainder of CI checks as they don't exercise `keymgr_dpe`

PR #25160, opened 2024-11-15
[fpv] Support for MuBi REGWEN fields

alees24:
This PR collects the MuBi REGWEN support from the private integrated repository, since it was not carried across and has been leading to DMA FPV assertions:

Introduce support for REGWEN fields that are specified as MultiBit values.

### DIFF
--- a/util/reggen/fpv_csr.sv.tpl
+++ b/util/reggen/fpv_csr.sv.tpl
@@ -53,6 +53,7 @@ module ${mod_base}_csr_assert_fpv import tlul_pkg::*;
   hro_idx_width = num_hro_regs.bit_length()
 %>\
 
+  import prim_mubi_pkg::*;
 `ifdef UVM
   import uvm_pkg::*;
 `endif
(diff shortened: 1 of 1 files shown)


rswarbrick:
Looks sensible: I think this is a port of 16fa702aeedbcceee169d708650669df345c3b98. Do we also need to cherry-pick 932a1b95f981bc1ce0f2bc9dd8f875fa48185403?

rswarbrick (approved):

alees24:
> Looks sensible: I think this is a port of [16fa702](https://github.com/lowRISC/opentitan/commit/16fa702aeedbcceee169d708650669df345c3b98). Do we also need to cherry-pick [932a1b9](https://github.com/lowRISC/opentitan/commit/932a1b95f981bc1ce0f2bc9dd8f875fa48185403)?

It seems you did that 3 weeks ago, so no. Unfortunately this earlier change had not made it across. I was surprised to discover that they're separable like that.

PR #25162, opened 2024-11-15
[ci] add GitHub actions FPGA job

nbdd0121:


.github/workflows/ci.yml
```diff
@@ -299,6 +299,91 @@ jobs:
           gcloud storage cp latest.txt $BUCKET_URI/latest.txt
           gcloud storage cp $BUCKET_URI/bitstream-${{ github.sha }}.tar.gz $BUCKET_URI/bitstream-latest.tar.gz
 
+  # CW310 FPGA jobs.
+  execute_test_rom_fpga_tests_cw310:
+    name: CW310 Test ROM Tests
+    needs: [chip_earlgrey_cw310, sw_build_test]
```
jwnrt, line 305: I think we discussed dropping the `sw_build_test` dependency, did we agree to keep it?

azure-pipelines.yml
jwnrt, line 1: I think you can remove the CW310 bitstream build from Azure now, right?

a-will, line 1: Shouldn't chip_earlgrey_cw310 be removed as well? It seems we're running duplicate CW310 build jobs right now.

nbdd0121: It's currently a dependency of the "Package & release" step. Although I think the artifact generated by that step is not being used anymore and we can probably remove it, I'll do it as a follow-up PR.

jwnrt (approved):
Updated version LGTM

PR #25163, opened 2024-11-15
[dv] Sample integrity errors more sensibly on the A channel

rswarbrick:
The existing code was wrong. I think the original author had imagined that you could count the number of injected errors by looking at the number of bit differences after re-calculating parity, but this is incorrect.

Do something more like what we'd expect, and base this properly on SECDED decode.


alees24 (approved):
LGTM.

PR #25164, opened 2024-11-15
[ci] Add checkout step to verify-fpga job

jwnrt:
Without this, Azure automatically checks out the repo at the time this job runs, not necessarily at the same commit as the previous FPGA jobs.

This can lead to a test not existing when the FPGA tests ran, but now existing when this verify job runs. The verify job then complains that you didn't run that test.


nbdd0121 (approved):

jwnrt:
FPGA failure (`//sw/device/tests:i2c_target_test_fpga_cw310_sival`) looks unrelated, merging

PR #25171, opened 2024-11-16
[ci] increase GitHub bitstream job timeout

nbdd0121:
The job itself doesn't take much time, but it needs time to wait for license to become available.


jwnrt (approved):
Failures unrelated, merging to fix CI

PR #25179, opened 2024-11-18
[topgen] Support for external alert routing

Razer6:
This PR implements the external alert routing according to the [RFC](https://docs.google.com/document/d/1LGQjasiB6WJ1dL42hTKbzScmEmRQDXurdG6n8ORtJQk/edit?tab=t.0#heading=h.ly4mwumf0y6l).


hw/top_earlgrey/templates/toplevel.sv.tpl
```diff
@@ -47,7 +48,7 @@ for m in top['memory']:
 last_modidx_with_params = lib.idx_of_last_module_with_params(top)
 
 %>\
-module top_${top["name"]} #(
+module top_${top["name"]} import top_${top["name"]}_pkg::*; #(
```
rswarbrick, line 51: Why does this import need moving? I think you're using the name for e.g. the `outgoing_alert_*` signals with a complete path?

hw/top_earlgrey/templates/toplevel.sv.tpl
```diff
@@ -455,7 +479,11 @@ max_intrwidth = (max(len(x.name) for x in block.interrupts)
   % if block.alerts:
 <%
 w = len(block.alerts)
-slice = str(alert_idx+w-1) + ":" + str(alert_idx)
+if 'outgoing_alert' in m:
+  slice = str(outgoing_alert_idx[m['outgoing_alert']]+w-1) + ":" + str(outgoing_alert_idx[m['outgoing_alert']])
+else:
+  slice = str(alert_idx+w-1) + ":" + str(alert_idx)
+endif
```
rswarbrick, line 494: Untested, but maybe this would repeat less?
```python
w = len(block.alerts)
if 'outgoing_alert' in m:
  outgoing_alert = m['outgoing_alert']
  lo = outgoing_alert_idx[outgoing_alert]
else:
  lo = alert_idx
slice = f"{lo+w-1}:{lo}"
```

util/topgen/merge.py
```diff
@@ -885,20 +889,36 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
...
                 'name': lpg_name,
                 'clock_group': None if unmanaged_clock else clock_group,
                 'clock_connection': clock,
                 'unmanaged_clock': unmanaged_clock,
                 'reset_connection': primary_reset
             })
-            num_lpg += 1
+
+        if 'outgoing_alert' in module:
+            alert_group = module['outgoing_alert']
```
rswarbrick, line 903: Probably cleaner:
```python
        alert_group = module.get('outgoing_alert')
        if alert_group is not None:
```

util/topgen/merge.py
```diff
@@ -885,20 +889,36 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
...
+            if lpg_name not in outgoing_lpg_dict[alert_group]:
+                outgoing_lpg_dict[alert_group].update({lpg_name: outgoing_num_lpg[alert_group]})
```
rswarbrick, line 905: I think this can just be:
```python
                outgoing_lpg_dict[lpg_name] = outgoing_num_lpg[alert_group]
```
?

rswarbrick, line 905: I think I'd probably avoid using `defaultdict`. If this is the only place we might be incrementing for the first time, maybe it makes sense to do something like this here?
```python
                outgoing_num_lpg.setdefault(alert_group, 0)
```
Then `outgoing_num_lpg` can just be initialised with `{}`.

util/topgen/merge.py
```diff
@@ -885,20 +889,36 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
...
+
+                # since the alert handler can tolerate timing delays on LPG
+                # indication signals, we can just use the clock / reset signals
+                # of the first block that belongs to a new unique LPG.
+                clock = module['clock_connections'][block_clock.clock]
+                append_to_lpg_dict(top['outgoing_alert_lpgs'][alert_group])
+                outgoing_num_lpg[alert_group] += 1
+        else:
+            if lpg_name not in lpg_dict:
+                lpg_dict.update({lpg_name: num_lpg})
```
rswarbrick, line 915: Again, I think you can just do:
```python
                lpg_dict[lpg_name] = num_lpg
```

util/topgen/merge.py
```diff
@@ -885,20 +889,36 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
...
+                # since the alert handler can tolerate timing delays on LPG
+                # indication signals, we can just use the clock / reset signals
+                # of the first block that belongs to a new unique LPG.
+                clock = module['clock_connections'][block_clock.clock]
+                append_to_lpg_dict(top['alert_lpgs'])
+                num_lpg += 1
```
rswarbrick, line 921: Can't we get rid of `num_lpg` and just use `len(lpg_dict)`?

util/topgen/merge.py
```diff
@@ -885,20 +889,36 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
...
+
+        if 'outgoing_alert' in module:
+            alert_group = module['outgoing_alert']
+            if lpg_name not in outgoing_lpg_dict[alert_group]:
+                outgoing_lpg_dict[alert_group].update({lpg_name: outgoing_num_lpg[alert_group]})
+
+                # since the alert handler can tolerate timing delays on LPG
+                # indication signals, we can just use the clock / reset signals
+                # of the first block that belongs to a new unique LPG.
+                clock = module['clock_connections'][block_clock.clock]
```
rswarbrick, line 910: The flow of information through the `clock` variable is pretty confusing. Can we just add it as an argument to `append_to_lpg_dict`?

util/topgen/merge.py
```diff
@@ -907,6 +927,13 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
                     'lpg_name': lpg_name,
                     'lpg_idx': lpg_dict[lpg_name]
                 })
+        for alert_group, alerts in top['outgoing_alert'].items():
+            for alert in alerts:
+                if alert['module_name'] == module['name']:
+                    alert.update({
+                        'lpg_name': lpg_name,
+                        'lpg_idx': outgoing_lpg_dict[module['outgoing_alert']][lpg_name]
+                    })
```
rswarbrick, line 936: I realise that this is copied from the lines above, but can't we just do this?
```python
        for alert_group, alerts in top['outgoing_alert'].items():
            for alert in alerts:
                if alert['module_name'] == module['name']:
                    alert['lpg_name'] = lpg_name
                    alert['lpg_idx'] = outgoing_lpg_dict[module['outgoing_alert']][lpg_name]
```

andreaskurth (commented):
Thanks for this work @Razer6. The underlying RFC is currently being reviewed, so I won't approve this PR yet. The changes in the PR seem to be aligned with the RFC, and in that sense they look good to me.

Please also add the RFC doc as Markdown to this PR. (You may want to wait with that until it's fully approved, though.)

hw/top_darjeeling/templates/toplevel.sv.tpl
```diff
@@ -396,6 +405,21 @@ for rst in output_rsts:
   assign lpg_rst_en[${k}] = ${rst_en};
 % endfor
 
+% for alert_group, lpgs in top['outgoing_alert_lpgs'].items():
+  // Outgoing LPGs for alert group ${alert_group}
+  % for k, lpg in enumerate(lpgs):
+  // ${lpg['name']}
+<%
+    cg_en = top['clocks'].hier_paths['lpg'] + lpg['clock_connection'].split('.clk_')[-1]
```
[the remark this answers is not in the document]
Razer6: That one is defined in [here](https://github.com/lowRISC/opentitan/pull/25179/files#diff-e4870f59f3d7b3080fabf9d5216c5a87104b49a1b98b1b9bb55b3157d1d4e612L679)

hw/top_earlgrey/templates/toplevel.sv.tpl
Razer6: I think that one can be removed. I though originally it was needed for the signal definitions.

util/topgen/merge.py
Razer6: Is there any reason not to use it?

util/topgen/merge.py
Razer6: No I don't think that would work. We need a mapping of LPG name to the index in the vector to be able to later associate the alerts to that.

hw/top_darjeeling/templates/toplevel.sv.tpl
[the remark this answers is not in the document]
rswarbrick: Oh! I see: the definitions are things like `clkmgr_aon_clocks.clk_io_div4_peri` and this wants to get `io_div4_peri` here. (A singleton but I hadn't twigged we weren't splitting on spaces)

If I've got it right, this is essentially the `clk` signal in `merge.py`, line 900? Is there any way that the generated json file can preserve more of the structure? If not, maybe it makes sense to be more explicit about how we're splitting things here? Since this split happens several times in the code, maybe it needs to be in a helper function.

util/topgen/merge.py
rswarbrick: I think I agree that we should keep the dictionary, but I'm reasonably sure that `num_lpg` is always equal to `len(lpg_dict)`, so I'm proposing that we get rid of the `num_lpg` variable.

util/topgen/merge.py
```diff
@@ -885,20 +889,36 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
 
         # if clock group is "unique", add some uniquification to the tag
         lpg_name = f"{module['name']}_{lpg_name}" if unique_cg else lpg_name
-        if lpg_name not in lpg_dict:
-            lpg_dict.update({lpg_name: num_lpg})
-            # since the alert handler can tolerate timing delays on LPG
-            # indication signals, we can just use the clock / reset signals
-            # of the first block that belongs to a new unique LPG.
-            clock = module['clock_connections'][block_clock.clock]
```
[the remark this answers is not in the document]
Razer6: That is refactored to use `clk`

util/topgen/merge.py
Razer6: Ahhh, sorry. I misread that. Got it now. Yeah, as we construct the dict, we can use its length to get the mapping index.

util/topgen/merge.py
Razer6: That one is now gone.

Razer6:
@rswarbrick I think I addressed all your comments. Could you take another look?

Razer6:
> Please also add the RFC doc as Markdown to this PR. (You may want to wait with that until it's fully approved, though.)

@andreaskurth I added an issue to track the documentation of that.

hw/top_darjeeling/templates/toplevel.sv.tpl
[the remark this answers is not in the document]
Razer6: In the last commit, I simplified that a lot I think.

andreaskurth:
> @andreaskurth I added an issue to track the documentation of that.

For reference, that would be https://github.com/lowRISC/opentitan/issues/25449

andreaskurth:
AUTHORIZE CHANGES: hw/top_earlgrey/rtl/autogen/top_earlgrey.sv

Just whitespaces; no functional changes.

vogelpi:

Just whitespaces; no functional changes.

hw/top_darjeeling/templates/toplevel.sv.tpl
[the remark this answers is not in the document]
rswarbrick: I really like it. Thanks!

util/topgen/merge.py
```diff
@@ -892,29 +894,41 @@ def create_alert_lpgs(top, name_to_block: Dict[str, IpBlock]):
...
                 'clock_group': None if unmanaged_clock else clock_group,
                 'clock_connection': clock,
                 'unmanaged_clock': unmanaged_clock,
                 'unmanaged_reset': is_unmanaged_reset(top, reset_name),
                 'reset_connection': primary_reset
             })
-            num_lpg += 1
+
+        alert_group = module.get('outgoing_alert')
+        if alert_group:
```
rswarbrick, line 913: Sorry to be annoying, but I'd probably suggest checking with `is not None` to avoid silly things happening if we accidentally have an alert group with an empty name.

util/topgen/merge.py
rswarbrick: Looks good, thanks!

rswarbrick (approved):
This looks good to me, thanks. One nitty comment about empty strings but it needn't really gate this going in.

rswarbrick:

As Andreas said, this is just a couple of blank lines: no functional change.

PR #25180, opened 2024-11-18
Enable KMAC error conditions test to partially run on CW310

AlexJones0:
This PR enables the `kmac_error_conditions_test` (added in #24765) in the CW310 FPGA execution environment, conditionally disabling the first three tests when running in this environment.

As mentioned in the code and in that PR, the CW310 lacks the masking features for these tests to run, but there is still value in running the remaining 7 tests in a CW310 environment. CI prioritisation now also prioritises CW340 over CW310 (#24947), meaning that the full set of tests will still be running in CI as usual.

This has been tested to run on a CW310 locally for 10 iterations: `./bazelisk.sh test -t- --test_output=streamed --runs_per_test=10 //sw/device/tests:kmac_error_conditions_test_fpga_cw310_sival`

@nasahlpa


nasahlpa (approved):
Thanks for the offline discussion @AlexJones0!
LGTM.

PR #25182, opened 2024-11-18
[ci] use service accounts instead of workload identity federation

nbdd0121:
It appears that Bazel cache performance suffers when using WIF. This is likely a bug in Bazel (or google-auth-library java) since both the ID token and the exchanged access token can be cached. I suppose when WIF is in use there're excessive amount of access token exchange happening which slows down all Google bucket accessese.

As a workaround, go back to use service account JSON instead of WIF.

### DIFF
--- a/.github/actions/prepare-env/action.yml
+++ b/.github/actions/prepare-env/action.yml
@@ -6,6 +6,9 @@ name: Prepare environment
 description: Install dependencies and prepare environment needed for OpenTitan
 
 inputs:
+  service_account_json:
+    description: Service account JSON for Google Cloud access
+    default: ''
   verilator-version:
     description: Verilator version to install
     required: true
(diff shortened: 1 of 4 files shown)


.github/actions/prepare-env/action.yml
```diff
@@ -6,6 +6,9 @@ name: Prepare environment
 description: Install dependencies and prepare environment needed for OpenTitan
 
 inputs:
+  service_account_json:
+    description: Service account JSON for Google Cloud access
+    default: ''
```
jwnrt, line 11: Can this default to `'${{ secrets.BAZEL_CACHE_CREDS }}'`? We're only ever going to pass this value from callers

nbdd0121: Actions can't have access to secrets directly, only workflows can.

jwnrt: I guess that makes sense

jwnrt (approved):

nbdd0121:
Merge to fix sw build & test CI, skipping bitstream build CI.

PR #25183, opened 2024-11-18
[test] Improve reliability of UART baud rate test

jwnrt:
The FPGA cannot clock high enough for 1MBd or 1.5MBd, but silicon can.

This change enables a wider range of baud rates and changes how the 1MBd and 1.5MBd are restricted to be better for multitop.

Backport of #23605


engdoreis (approved):

jwnrt:
Hm, it failed in CI. I guess it's still flaky.

PR #25190, opened 2024-11-18
[i2c, sival] Reduce I2C bitbanging flakiness by separating SDA and SCL pin writes

AlexJones0:
This PR reduces the flakiness seen in tests that use I2C over GPIO bitbanging - namely `i2c_target_test` where this was being used.

Currently, writes to SCL and SDA can be performed within a single GPIO bitbanging sample. This relied on the SDA value being set before the the SCL value, but in reality these changes occur almost simultaneously. Physical characteristics of the FPGA such as e.g. the capacitance on different pins could add delays of a couple of microseconds, causing the SDA variation to appear after the SCL variation. This could cause a valid write (setting SDA high before a rising edge on SCL) to transmit as a 'None' signal, which was causing a lot of flakiness in tests, varying across different FPGAs.

This commit fixes the I2C bitbanging logic to run at half the sampling rate, extending each sample into two samples - one that sets the new SDA value whilst the SCL is maintained, and one that sets the SCL value while the SDA is maintained. This ensures "atomic" sequential I2C GPIO pin writes which will be interpreted correctly, improving reliability.

This has been tested locally on FPGA on `//sw/device/tests:i2c_target_test_fpga_cw310_sival`.

### DIFF
--- a/sw/host/opentitanlib/src/test_utils/bitbanging/i2c.rs
+++ b/sw/host/opentitanlib/src/test_utils/bitbanging/i2c.rs
@@ -24,16 +24,36 @@ impl Symbol {
         Ok(Self::Broken(buffer))
     }
 
+    // We must extend the transaction: in a given sample, if data is written to SDA at the same time as SCL
+    // being driven high, then physical factors e.g. capacitance can cause the SDA write to appear before
+    // the SCL, even though the order of bitbanging is reversed. This can then cause a mistaken "STOP" signal
+    // to be received. We instead run at half frequency: on one cycle write SDA, and on the next write SCL.
+    // This makes GPIO bitbanging of I2C signals reliable.
+    // Assumes that SCL is low (0) when called.
     pub fn bitbanging<const SDA: u8, const SCL: u8>(&self, samples: &mut Vec<u8>) {
         match self {
+            // Each sample is translated into 2 samples: the first changes the SDA, and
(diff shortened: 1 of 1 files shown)


AlexJones0:
CI failure is a timeout in `usbdev_iso_test_fpga_cw310_sival_rom_ext` which is not related (does not use bitbanging, and not run in the same Azure job).

engdoreis (changes requested):
We should change the encoder to generate the new expected output.

sw/host/opentitanlib/src/test_utils/bitbanging/i2c.rs
```diff
@@ -43,21 +63,33 @@ impl Symbol {
 
     fn bitbanging_byte<const SDA: u8, const SCL: u8>(byte: u8, nack: bool, samples: &mut Vec<u8>) {
         let data: u16 = (byte as u16) << 1u16 | nack as u16;
-        samples.extend((0..9u8).rev().flat_map(|bit| {
-            [
+        let mut prev_scl: u16 = (samples.last().unwrap_or(&(0x01 << SCL)) & (0x01 << SCL)) as u16;
```
engdoreis, line 66: You don't need the last clock, similarly to the previous version, you can assume that the clock will always be left low by the last function that operated on samples. This makes the code simpler. And then you can keep the `flat_map`.

engdoreis (approved):

PR #25193, opened 2024-11-18
util: tool for parsing BFV codes

vbendeb:
Given the hex value reported in the `BFV: xxxxxxxx` entry of the crash log print the symbolic name of the error.

Usage example:

  > bazel run //util:fw_error_decode -- 50540d 0c4f5705 123
  ...
  0050540d: kErrorPersoTlvInternal
  0c4f5705: kErrorOwnershipKeyNotFound
  00000123: unknown error code

### DIFF
--- a/sw/host/opentitantool/BUILD
+++ b/sw/host/opentitantool/BUILD
@@ -10,6 +10,7 @@ package(default_visibility = ["//visibility:public"])
 rust_binary(
     name = "opentitantool",
     srcs = [
+        "src/command/bfv.rs",
         "src/command/bootstrap.rs",
         "src/command/certificate.rs",
         "src/command/clear_bitstream.rs",
@@ -53,6 +54,7 @@ rust_binary(
         "//sw/host/opentitanlib",
         "//sw/host/ot_certs",
         "//sw/host/sphincsplus",
+        "//util:bfv_decoder",
         "@crate_index//:anyhow",
         "@crate_index//:clap",
         "@crate_index//:directories",
--- a/sw/host/opentitantool/src/command/bfv.rs
+++ b/sw/host/opentitantool/src/command/bfv.rs
@@ -1,0 +1,55 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::{Context, Result};
+use clap::Args;
+use serde_annotate::Annotate;
+use std::any::Any;
+
+use opentitanlib::app::command::CommandDispatch;
+use opentitanlib::app::TransportWrapper;
+
+#[derive(Debug, Args)]
+/// Decode a raw status. Optionally accepts an ELF file to recover the filename.
+pub struct BfvCommand {
+    /// Hex BFV value as reported by device on failures.
+    bfv: Vec<String>,
+}
+
+extern "C" {
+    fn bfv_decoder(bfv: u32, buf: *mut u8, buf_size: usize) -> usize;
+}
+
+impl CommandDispatch for BfvCommand {
+    fn run(
+        &self,
+        _context: &dyn Any,
+        _transport: &TransportWrapper,
+    ) -> Result<Option<Box<dyn Annotate>>> {
+        for value in &self.bfv {
+            // Decode status.
+            let string_bfv = if value.starts_with("0x") {
+                &value.as_str()[2..]
+            } else {
+                value.as_str()
+            };
+
+            let bfv = u32::from_str_radix(string_bfv, 16)
+                .context(format!("\"{}\" is not a valid hex value", string_bfv))?;
+            let mut text = [0u8; 80];
+            // SAFETY:  the decodr function is guaranteed to size of the string
+            // written into text.
+            let size = unsafe { bfv_decoder(bfv, text.as_mut_ptr(), text.len()) };
+
+            println!(
+                "{:08x}: {}",
+                bfv,
+                std::str::from_utf8(&text[..size]).unwrap()
+            );
+        }
+        // Separate command output from the next shell prompt.
+        println!();
+        Ok(None)
+    }
+}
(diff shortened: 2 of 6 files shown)


pamaury:
This looks super useful! One of the most common uses of decoding BFV is when running programs through opentitanlib. I wonder if opentitanlib could gain this facility as well as, for example, automatically decode them on `exit_failure` when waiting for a BFV code? Or at least opentitanlib/tool could have a subcommand for it (because if you are already running a test with it, you might as well use the same tool).

There is also the specific BFV code for exception which is not just defined by the table but by looking at various bits of bit and is quite useful as well.

Also ideally this tool should appear in some documentation somewhere, I just don't know where?

vbendeb:
Hi Amaury, thank you for looking at this.

I did notice that some exception errors ('R' 'I') cannot be parsed using this approach, I can add a special case for these just showing the component values separately, or is there a better way?

> I wonder if opentitanlib could gain this facility as well as, for example, automatically decode them on exit_failure when waiting for a BFV code?

Can you please elaborate - I am not aware of the situations where opentitanlib is waiting for a BFV code. Also, opentitanlib is in Rust, and this utility is pure C for a good reason.

> Or at least opentitanlib/tool could have a subcommand for it (because if you are already running a test with it, you might as well use the same tool

Again, I thought about adding this to opentitanttool, but I don't see a good way of writing this in Rust, or are you suggesting interfacing opentitantool/opentitanlib with C?

pamaury:
> Hi Amaury, thank you for looking at this.
> 
> I did notice that some exception errors ('R' 'I') cannot be parsed using this approach, I can add a special case for these just showing the component values separately, or is there a better way?

I think you need a special case. Here is an example of one in the ROM:
https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/rom/rom_isrs.c#L26
There might be more

> 
> Can you please elaborate - I am not aware of the situations where opentitanlib is waiting for a BFV code. Also, opentitanlib is in Rust, and this utility is pure C for a good reason.
> 
Maybe I didn't express that well: whener we run a test using opentitantool, we usually use the console operation that waits for either an `exit_success` or `exit_failure` regex. The default `exit_failure` regex includes a BFV:
https://github.com/lowRISC/opentitan/blob/master/rules/opentitan/defs.bzl#L123

I was therefore wondering if you would make sense to have opentitanlib/tool detect that special case and decode the message, since it's quite common. But that's just a thought.

It is possible to the code in C like you did, and then use it from rust. For example a C API like this would work for rust:
```c
bool decode_bfv(uint32_t bfv, char *out_buf, size_t out_buf_size)
```

util/fw_error_decoder.c
```diff
@@ -0,0 +1,77 @@
...
+    const char *code = argv[i];
+    const error_descriptor_t *edesc;
+    long error_value = strtol(code, &end, 16);
+
+    if (*end) {
+      fprintf(stderr, "'%s' is not a hex value\n", code);
+      usage(1);
+    }
+
+    edesc = error_table;
```
engdoreis, line 62: Nit. If you move the declaration here, it would be easier to read.

util/fw_error_decoder.c
```diff
@@ -0,0 +1,77 @@
...
+    printf("%08lx: ", error_value);
+    while (edesc->value) {
```
engdoreis, line 64: I think you use the `sizeof` the table to stop the loop.

util/fw_error_decoder.c
```diff
@@ -50,7 +50,7 @@ int main(int argc, char **argv) {
 
   while (i < argc) {
     char *end;
```
engdoreis, line 52: You should squash this second commit.

engdoreis (commented):
Thanks @vbendeb for working on this very useful tool.

Please see my suggestions below.

cfrantz:
> I think you need a special case. Here is an example of one in the ROM: https://github.com/lowRISC/opentitan/blob/master/sw/device/silicon_creator/rom/rom_isrs.c#L26 There might be more
> 
> > Can you please elaborate - I am not aware of the situations where opentitanlib is waiting for a BFV code. Also, opentitanlib is in Rust, and this utility is pure C for a good reason.
> 
> Maybe I didn't express that well: whener we run a test using opentitantool, we usually use the console operation that waits for either an `exit_success` or `exit_failure` regex. The default `exit_failure` regex includes a BFV: https://github.com/lowRISC/opentitan/blob/master/rules/opentitan/defs.bzl#L123
> 
> I was therefore wondering if you would make sense to have opentitanlib/tool detect that special case and decode the message, since it's quite common. But that's just a thought.

There is code in opentitanlib to deal with `rom_error_t`, but it hasn't been cherry-picked to master yet.  The library `//sw/host/opentitanlib/bindgen` contains a `rom_error` module which has a `RomError` unit struct that defines all of the rom_error_t values and has the various Display, Debug, FromStr and serde traits implemented.

The existence of this particular bit of rust code doesn't negate the value of this simple C program, but it does make @pamaury's idea of enhancing the error case in the console easier.  It's worth pursing after the rest of the ROM_EXT and ownership transfer code gets cherry-picked onto master.

vbendeb:
Please take a look guys, also, can it go in as a standalone utility (which I think would be useful) or is it a must to be integrated with opentitantool/lib?

vbendeb:
also, I have some responses to comments marked on the page as `pending` but I don't see a way to publish them, is there some button I'm missing?

util/fw_error_decoder.c
```diff
@@ -0,0 +1,98 @@
...
+    unsigned char mod_high = (error_value >> 16) & 0xff;
+    unsigned char mod_low = (error_value >> 8) & 0xff;
+
+    if (((mod_high == 'R') && (mod_low == 'I')) ||
+        ((mod_high == 'I') && (mod_low == 'R'))) {
+      // This is an interrupt/exception error, retrieve the encoded mcause.
+      uint32_t mcause =
+          (error_value & (1 << 31)) | ((error_value >> 24) & 0x7f);
+
+      if (mod_high == 'R')
```
engdoreis, line 88: Per the style guide, `if` should always have braces.

timothytrippel:
@vbendeb since this patch is a couple weeks old, would you mind rebasing it on the tip of tree and re-uploading it just to ensure there is no merge SKU?  Then I will merge once it passes CI.

Once this merges on the master branch, it would be nice to cherry pick this to the earlgrey_1.0.0 branch as well.

vbendeb:
rebased and re-tested

Issue #24558, opened 2024-09-11
[ot_certs,rom_ext] optimize on-device certificate generation code-size


timothytrippel:
### Description

Currently, there exists opentitantool support (`opentitantool certificate codegen ...`) for generating device libraries that may be [used to generate X.509 certificate binary blobs](https://github.com/lowRISC/opentitan/blob/25b1acbf687e5535e8742c2aacb5b224ee77c80d/sw/device/silicon_creator/lib/cert/dice.c#L243). The libraries make use of a manually crafted [device library that can encode simple ASN.1 structures](https://github.com/lowRISC/opentitan/blob/25b1acbf687e5535e8742c2aacb5b224ee77c80d/sw/device/silicon_creator/lib/cert/asn1.h). While this mechanism is extremely flexible, i.e., new certificate extensions can be easily crafted, it comes at a cost of code-size. Specifically, when analyzing the size-breakdown of various components in the ROM_EXT, we can see that:
- the ASN.1 lib uses ~1.8kb
- the X.509 CDI_0 certificate gen code uses ~2.35kb
- the X.509 CDI_1 certificate gen code uses ~2.35kb
for a grand **total of ~6.5kb**.

![image](https://github.com/user-attachments/assets/c3a252d5-9e7d-4c81-a807-3ecdad703d94)

Moreover, [we are currently attempting to add support for other DICE certificate extensions, and encodings](https://github.com/lowRISC/opentitan/issues/24281#issuecomment-2316370519) (i.e., CBOR). If we continue to use the same approach, where we use host tools to generate device code that generates certificates on-device by calling into an additional CBOR device lib, then we could end up bloating the ROM_EXT (if we aim to provide a flexible ROM_EXT that can support multiple DICE certificate formats).

We should investigate a more efficient way to generate certificates on-device.

One approach could be to use an auto-generated fixed-size array template that various fields can be mem-copied into. The original rationale for not using such an approach was that doing so would require all fields in the TBS region to be fixed size, otherwise it is impossible to generate a fixed-sized template array. The biggest issue was that the serial number field in X.509 certs is encoded as an ASN.1 integer type, that does not allow 0 padding on the MSb side. One way around this is to just automatically set the MSb, since serial numbers are not crytpographic, rather just unique values (currently the serial number for OT certs is just a truncated SHA256 of the pub key). Other ideas/suggestions are also welcome.

[...]

PR #25198, opened 2024-11-19
[dice,x509] Reduce X509 template library code size

sasdf:
Our ROM_EXT is close to the limit,
and future changes (e.g. immutable rom_ext) will likely introduce 10~20k more bytes.

This refactor moves the error propagation to the end of asn1 builder (`asn1_finish`), reducing a huge amount of branches to save firmware size.

This patch reduced
* 1864 bytes from the ROM_EXT. (46180 to 44316)
* 3040 bytes from FT_perso. (92164 to 89124)

This PR partially address #24558.

---

### Technical details

More specifically, we added an internal error tracking variable `state->error`. When any operation results in error. The variable will be set, and all future operations will be no-op. This patch comes with the cost that we will execute many no-ops before the error is propagated. However, since the device shouldn’t trigger any of these errors during normal conditions, this cost seems affordable.

Also, the error guard `RETURN_IF_ASN1_ERROR(state);` is added when:
* The operation may raise a new error.
* The operation has side effects on its arguments.

### DIFF
--- a/sw/device/silicon_creator/lib/cert/asn1.c
+++ b/sw/device/silicon_creator/lib/cert/asn1.c
@@ -6,6 +6,31 @@
 
 #include <limits.h>
 
+/**
+ * Return if the given asn1 state has an active error.
+ *
+ * @param state Pointer to a user-allocated `asn1_state_t` state.
+ */
+#define RETURN_IF_ASN1_ERROR(state)   \
+  do {                                \
+    if ((state)->error != kErrorOk) { \
+      return;                         \
+    }                                 \
+  } while (false);
+
+/**
+ * Set the active error in asn1 state to `error_code` and return.
+ *
+ * @param state Pointer to a user-allocated `asn1_state_t` state.
+ */
+#define RAISE_ASN1_ERROR(state, error_code) \
+  do {                                      \
+    (state)->error = (error_code);          \
+    return;                                 \
+  } while (false);
+
+void asn1_clear_error(asn1_state_t *state) { state->error = kErrorOk; }
+
 rom_error_t asn1_start(asn1_state_t *new_state, uint8_t *buffer, size_t size) {
   // Make sure that the buffer is not too large to prevent overflows.
   if (new_state == NULL || buffer == NULL || size > PTRDIFF_MAX) {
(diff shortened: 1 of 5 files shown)


sw/device/silicon_creator/lib/cert/asn1.h
```diff
@@ -25,8 +25,17 @@ typedef struct asn1_state {
...
 } asn1_state_t;
 
+/**
+ * Clear the ASN1 builder internal error status.
+ *
+ * @param[out] state Pointer to a user-allocated state to be cleared.
+ */
+void asn1_clear_error(asn1_state_t *state);
+
 /**
```
pamaury, line 39: It think it's very important to write somewhere (maybe everywhere?) that every operation can update the `error` variable in the state.

sasdf: Sure things, added in all function comments.

sw/device/silicon_creator/lib/cert/asn1.c
```diff
@@ -6,6 +6,21 @@
...
 #include <limits.h>
 
+#define RETURN_IF_ASN1_ERROR(state)                      \
+  do {                                                   \
+    if ((state) == NULL || (state)->error != kErrorOk) { \
+      return;                                            \
+    }                                                    \
+  } while (false);
+
+#define RAISE_ASN1_ERROR(state, error_code) \
```
pamaury, line 16: Document as well.

pamaury (approved):
I tried to review everything. To be honest, with this approach, I find it harder to make sure that the code correctly handles all error conditions. This might be an instance where fuzzing could be very effective.

sw/device/silicon_creator/lib/cert/asn1.c
```diff
@@ -6,6 +6,21 @@
 
 #include <limits.h>
 
+#define RETURN_IF_ASN1_ERROR(state)                      \
```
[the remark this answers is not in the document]
sasdf: Done, and I also removed the NULL check.

The null check was for asn1_tag and bitstring, but adding the NULL check directly in those functions will save us some extra bytes.

timothytrippel (approved):

PR #25259, opened 2024-11-20
[opentitantool] Advanced TPM primitives in transport

jesultra:
Adapt TPM driver layer to use advanced transport features, if available.  These features allows asking the backend transport to perform a full TPM register read or write in a single USB roundtrip, including e.g. polling the SPI bus to know when the device is ready for the data transfer.

This change speeds up TPM communication between 2x and 4x when HyperDebug is used to communicate with GSC chips.

### DIFF
--- a/sw/host/opentitanlib/src/app/i2c.rs
+++ b/sw/host/opentitanlib/src/app/i2c.rs
@@ -2,11 +2,12 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::io::gpio::GpioPin;
 use crate::io::i2c::{self, Bus, DeviceStatus, Mode};
 use crate::transport::Transport;
 
 use anyhow::Result;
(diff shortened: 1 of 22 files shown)


jwnrt (approved):
Thanks @jesultra, this all looks good to me.

Aside, I believe we're in the middle of adding jumpers to our HyperDebug boards in CI to enable TPM tests, so we may be able to get better test coverage soon.

jesultra:
I will follow up with another PR to update the HyperDebug firmware to support the TPM operations in a single USB roundtrip.  For now, the `driver.rs` code will notice the lack of support, and fall back to the old polling logic.

> Aside, I believe we're in the middle of adding jumpers to our HyperDebug boards in CI to enable TPM tests, so we may be able to get better test coverage soon.

That is funny.  Just yesterday, I realized why a test of mine was working on the Teacup board in the ChromeOS lab, and the FPGA on my desk, but not the FPGA in the lab.  And the reason turned out to be lack of jumpers.

PR #25260, opened 2024-11-20
[manuf] enable use of two different root CAs during perso

### LINKED ISSUE #24955: [manuf] enable use of multiple CA keys in perso flow
Currently, a single CA key is used to endorse all certificates, including:
1. the UDS DICE cert, and
2. all SKU-specific certs.

However, the some SKU-specific certs have requirements over the `pathlen` constraint, as demonstrated in #23121, while the CA used to root the DICE chain should not have such a constraint (as pointed out in #24833).

To facilitate different SKU requirements, the perso flow should accept, as inputs, two different CAs:
1. one to endorse the UDS cert, and
2. one to endorse all SKU-specific certs.

timothytrippel:
This fixes #24955 by enabling the use of two different CAs during personalization:
- one for endorsing the UDS cert (i.e., the root of the DICE chain), and
- one for endorsing all SKU-specific certs provisioned in the personalization extensions.

As a byproduct, this also enables a flow where the CA private key is stored in a Cloud KMS keyring (or eventually a Nitrokey as well).


sw/host/provisioning/ft/src/main.rs
```diff
@@ -119,19 +112,44 @@ fn main() -> Result<()> {
...
-    // Individualization ujson payload.
+
+    // Parse and prepare individualization ujson data payload.
     let _ft_individualize_data_in = ManufFtIndividualizeData {
         device_id: hex_string_to_u32_arrayvec::<8>(opts.provisioning_data.device_id.as_str())?,
     };
-    // Personalization ujson payload.
+
+    // Parse and prepare CA key.
+    let mut ca_cfgs: HashMap<String, CaConfig> = from_str(
```
cfrantz, line 129: Since `from_str` is a bare function in the `serde_json` namespace, it'd be better to have the full "path" to the function here:
```
let foo = serde_json::from_str( ... )?;
```

Also, the docstring for `ca_config` says "CA HJSON config file", but by using `serde_json`, this is strict json.

Should do one of:
- Support hjson via `deser_hjson::from_str`
- Support json, json5 or hjson via `serde_annotate::from_str`
- Support only strict json and change the docstring.

sw/host/provisioning/ft/src/main.rs
```diff
@@ -2,15 +2,20 @@
...
 use std::path::PathBuf;
 use std::time::Duration;
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use clap::{Args, Parser};
+use elliptic_curve::pkcs8::DecodePrivateKey;
+use elliptic_curve::SecretKey;
+use p256::NistP256;
+use serde_json::from_str;
```
cfrantz, line 14: You can eliminate the `use` statement if you refer to the full path `serde_json::from_str` at the call site.

cfrantz (approved):

sw/host/provisioning/ft/src/main.rs
timothytrippel: Fixed. Went with serde_annotate for most flexibility in the future.

sw/host/provisioning/ft/src/main.rs
timothytrippel: fixed.

PR #25262, opened 2024-11-20
[topgen] Support for unmanaged external resets

Razer6:
Tops may specify resets that are completely managed externally, i.e., with a dedicated reset manager in the uncore environment of an integrated OpenTitan. Feeding that reset to the internal reset manager is not necessary and just adds complexity in the configuration.

To address this, this PR introduces support for unmanaged external resets, allowing them to operate independently from the internal reset manager. To define an unmanaged reset, simply add an entry to the top-level unmanaged_resets list. Once defined, these resets can be used like any other reset for IP instantiation and crossbar mapping.

Additionally, when routing IP alerts to the alert manager, the alert manager requires a reset status signal to know if the reset for the alert’s IP is currently active. For unmanaged external resets, it's assumed that the top-level external environment already has knowledge of the reset status. Topgen creates a dedicated LPG for every external reset. The reset status signal is routed at top-level and connected to the associated group to the alert manager.

### DIFF
--- a/hw/top_darjeeling/templates/toplevel.sv.tpl
+++ b/hw/top_darjeeling/templates/toplevel.sv.tpl
@@ -8,6 +8,7 @@ import topgen.lib as lib
 from reggen.params import Parameter
 from topgen.clocks import Clocks
 from topgen.resets import Resets
+from topgen.merge import is_unmanaged_reset
 
 num_mio_inputs = top['pinmux']['io_counts']['muxed']['inouts'] + \
                  top['pinmux']['io_counts']['muxed']['inputs']
@@ -127,6 +128,14 @@ module top_${top["name"]} #(
     % for clk in top['unmanaged_clocks']._asdict().values():
   input                        clk_${clk.name}_i,
   input prim_mubi_pkg::mubi4_t cg_${clk.name}_i,
+    % endfor
+  % endif
+  % if len(top['unmanaged_resets']._asdict().values()) > 0:
+
+  // Unmanaged external resets
+    % for rst in top['unmanaged_resets']._asdict().values():
+  input                        ${rst.signal_name},
+  input prim_mubi_pkg::mubi4_t ${rst.rst_en_signal_name},
     % endfor
   % endif
 
@@ -387,7 +396,7 @@ for rst in output_rsts:
     cg_en = 'cg_' + lpg['clock_connection'].split('clk_')[-1]
   else:
     cg_en = top['clocks'].hier_paths['lpg'] + lpg['clock_connection'].split('.clk_')[-1]
-  rst_en = lib.get_reset_lpg_path(top, lpg['reset_connection'])
+  rst_en = lib.get_reset_lpg_path(top, lpg['reset_connection'], False, None, lpg['unmanaged_reset'])
   known_clocks[cg_en] = 0
   known_resets[rst_en] = 0
 %>\
@@ -542,10 +551,16 @@ slice = str(alert_idx+w-1) + ":" + str(alert_idx)
       .${k} (${v}),
     % endfor
     % for port, reset in m["reset_connections"].items():
-      % if lib.is_shadowed_port(block, port):
-      .${lib.shadow_name(port)} (${lib.get_reset_path(top, reset, True)}),
-      % endif:
-      .${port} (${lib.get_reset_path(top, reset)})${"," if not loop.last else ""}
+<%
+      is_shadowed_port = lib.is_shadowed_port(block, port)
+      unmanaged_reset = is_unmanaged_reset(top, reset['name'])
+      reset_port = lib.get_reset_path(top, reset, False, unmanaged_reset)
+      shadowed_port = lib.get_reset_path(top, reset, True, unmanaged_reset)
+%>\
+    % if is_shadowed_port:
+      .${lib.shadow_name(port)} (${shadowed_port}),
+    % endif
+      .${port} (${reset_port})${"," if not loop.last else ""}
     % endfor
   );
 % endfor
--- a/hw/top_earlgrey/data/top_earlgrey.hjson
+++ b/hw/top_earlgrey/data/top_earlgrey.hjson
@@ -106,6 +106,13 @@
     ],
   },
 
+  // List of unmanaged external resets
+  //
+  // These should be listed as dictionaries that just give their names, in this form:
+  //
+  //    { name: "my_external_reset" }
+  // unmanaged_resets: []
+
   // This is the reset data structure of the design.
   // The hier path refers to the reset reference path (struct / port)
   //   - The top/ext desgination follows the same scheme as inter-module
--- a/hw/top_earlgrey/templates/toplevel.sv.tpl
+++ b/hw/top_earlgrey/templates/toplevel.sv.tpl
@@ -8,6 +8,7 @@ import topgen.lib as lib
 from reggen.params import Parameter
 from topgen.clocks import Clocks
 from topgen.resets import Resets
+from topgen.merge import is_unmanaged_reset
 
 num_mio_inputs = top['pinmux']['io_counts']['muxed']['inouts'] + \
                  top['pinmux']['io_counts']['muxed']['inputs']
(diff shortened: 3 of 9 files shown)


hw/top_darjeeling/templates/toplevel.sv.tpl
```diff
@@ -388,7 +397,10 @@ for rst in output_rsts:
     cg_en = 'cg_' + lpg['clock_connection'].split('clk_')[-1]
   else:
     cg_en = top['clocks'].hier_paths['lpg'] + lpg['clock_connection'].split('.clk_')[-1]
-  rst_en = lib.get_reset_lpg_path(top, lpg['reset_connection'])
+  if lpg['unmanaged_reset']:
+    rst_en =  top['unmanaged_resets']._asdict()[lpg['reset_connection']['name']].rst_en_signal_name
+  else:
+    rst_en = lib.get_reset_lpg_path(top, lpg['reset_connection'])
```
rswarbrick, line 403: I know that `lpg` is just a dict, but maybe it makes sense to check more sensibly to make sure `reset_connection` is defined here? (And I realise that this is unchanged in the commit, but I'm asking because you're touching the line :-D)

hw/top_darjeeling/templates/toplevel.sv.tpl
```diff
@@ -388,7 +397,10 @@ for rst in output_rsts:
     cg_en = 'cg_' + lpg['clock_connection'].split('clk_')[-1]
   else:
     cg_en = top['clocks'].hier_paths['lpg'] + lpg['clock_connection'].split('.clk_')[-1]
-  rst_en = lib.get_reset_lpg_path(top, lpg['reset_connection'])
+  if lpg['unmanaged_reset']:
+    rst_en =  top['unmanaged_resets']._asdict()[lpg['reset_connection']['name']].rst_en_signal_name
```
rswarbrick, line 401: Would it make sense to define something like this in `UnmanagedResets`?
```python
    def get(self, name: str) -> object:
        try:
            return self.resets['name']
        except KeyError:
            raise ValueError(f"No reset defined with name {name}") from None
```
It might be a way to push information out of this file and give better error messages if something doesn't get connected up properly.

hw/top_earlgrey/data/top_earlgrey.hjson
```diff
@@ -106,6 +106,11 @@
     ],
   },
 
+  // List of unmanaged external resets
+  unmanaged_resets: [
+    // { name: "my_ext" }
+  ]
+
```
rswarbrick, line 113: I don't really think we need to document the variable here (where it's getting defined for a particular toplevel). If we do want to document it, I'd probably suggest filling out the example more explicitly, but not in the list. Maybe something like this? (in the template)
```
  // List of unmanaged external resets
  //
  // These should be listed as dictionaries that just give their names, in this form:
  //
  //    { name: "my_external_reset" }
  unmanaged_resets: []
```

util/topgen/merge.py
```diff
@@ -592,6 +592,10 @@ def is_unmanaged_clock(top: OrderedDict, clock: str):
     return clock in top['unmanaged_clocks']._asdict()
 
 
+def is_unmanaged_reset(top: OrderedDict, reset: str):
+    return reset in top['unmanaged_resets']._asdict()
```
rswarbrick, line 596: I'd probably defined `__contains__` for the `UnmanagedResets` class. For example:
```python
class Foo:
    def __init__(self):
        self.d = {1: 2}

    def __contains__(self, k: int) -> bool:
        return k in self.d


foo = Foo()
print(1 in foo)
print(2 in foo)
```

util/topgen/resets.py
```diff
@@ -214,3 +214,29 @@ def has_shadowed_reset(self) -> bool:
                 return True
 
         return False
+
+
+class UnmanagedReset:
+    '''An unmanaged reset (input to the top-level).'''
+
+    def __init__(self, raw: Dict[str, object]):
+        self.name = str(raw['name'])
```
rswarbrick, line 223: Practically speaking, we're currently expecting the dict to be a singleton that just has a `name` key. We should probably check for that here.

andreaskurth (approved):
LGTM other than a few small comments. Thx @Razer6 :+1:

hw/top_earlgrey/data/top_earlgrey.hjson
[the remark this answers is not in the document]
Razer6: Gotcha. Currently it's a required field, I will change that. In a follow up also for the unmanaged clocks I think.

util/topgen/merge.py
Razer6: In a follow up, I'll do that for the clocks as well.

hw/top_earlgrey/data/top_earlgrey.hjson
Razer6: I removed that at all, though I'll add that for Darjeeling?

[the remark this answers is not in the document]
Razer6: Or should we add the following to know this feature is there?

```
  // List of unmanaged external resets
  //
  // These should be listed as dictionaries that just give their names, in this form:
  //
  //    { name: "my_external_reset" }
  // unmanaged_resets: []
```

Darjeeling would not use that but downstream tops.

util/topgen/resets.py
Razer6: I added a check.

hw/top_darjeeling/templates/toplevel.sv.tpl
Razer6: Do you really think that is needed there? There is a single place, where something is something added to the `lpg` dict:

```top['alert_lpgs'].append({
                'name': lpg_name,
                'clock_group': None if unmanaged_clock else clock_group,
                'clock_connection': clock,
                'unmanaged_clock': unmanaged_clock,
                'unmanaged_reset': is_unmanaged_reset(top, reset_name),
                'reset_connection': primary_reset
            })
```

That's already the full lpg

Razer6:
@andreaskurth @rswarbrick Thanks for the review. I updated the PR accordingly.

hw/top_darjeeling/templates/toplevel.sv.tpl
```diff
@@ -129,6 +130,14 @@ module top_${top["name"]} #(
   input prim_mubi_pkg::mubi4_t cg_${clk.name}_i,
     % endfor
   % endif
+  % if len(top['unmanaged_resets']._asdict().values()) > 0:
```
matutem, line 133: Nit: remove .values() since the length of a dict and its values is the same.

hw/top_earlgrey/templates/toplevel.sv.tpl
```diff
@@ -131,6 +132,14 @@ module top_${top["name"]} #(
   input prim_mubi_pkg::mubi4_t cg_${clk.name}_i,
     % endfor
   % endif
+  % if len(top['unmanaged_resets']._asdict().values()) > 0:
```
matutem, line 135: Nit: remove .values() since the length of the dict and its values is the same.

hw/top_darjeeling/templates/toplevel.sv.tpl
```diff
@@ -543,10 +552,19 @@ slice = str(alert_idx+w-1) + ":" + str(alert_idx)
...
     % endfor
     % for port, reset in m["reset_connections"].items():
-      % if lib.is_shadowed_port(block, port):
-      .${lib.shadow_name(port)} (${lib.get_reset_path(top, reset, True)}),
-      % endif:
-      .${port} (${lib.get_reset_path(top, reset)})${"," if not loop.last else ""}
+<%
+      is_shadowed_port = lib.is_shadowed_port(block, port)
+      if is_unmanaged_reset(top, reset['name']):
+        reset_port = top['unmanaged_resets'].get(reset['name']).signal_name
```
matutem, line 558: Nit: maybe add support for unmanaged resets in topgen.lib, as in lib.get_reset_path?

matutem (approved):
It seems you are assuming the clock of an unmanaged reset will never be off while the alert handler is active. If this is so, is it mentioned somewhere more prominently?

Other than nits, LGTM.

andreaskurth:

Only a comment is added; no functional change.

vogelpi:

Only a comment is added; no functional change.

PR #25265, opened 2024-11-20
[pwm,dv] Lots of tidy-ups in the scoreboard and monitor

rswarbrick:
I've spent some time looking at the scoreboard and monitor (because of a failing test) and there were lots of places they could be tidied up. This PR contains those tidy-ups (and is not really intended to change behaviour).

There *is* one known behaviour change, which comes from a8ee8ca ("Use the correct clock for pwm_if instances"). Local tests don't show this affecting very much, but it probably makes more sense if the interface uses the clock that drives its data signal.

### DIFF
--- a/hw/ip/pwm/dv/env/pwm_env_pkg.sv
+++ b/hw/ip/pwm/dv/env/pwm_env_pkg.sv
@@ -55,10 +55,44 @@ package pwm_env_pkg;
     bit [15:0]   A;
   } dc_blink_t;
 
-  // the index of multi-reg is at the last char of the name
-  function automatic int get_multireg_idx(string name);
-    string s = name.getc(name.len - 1);
-    return s.atoi();
+  // Split a name of the form "foo_bar_123" into a base name ("foo_bar") and an index (123). Returns
+  // 1'b1 if the split is possible (the name ends with an underscore and then some digits).
+  //
+  // If the split is not successful, index is set to 0, base_name is set to full_name and the
+  // function returns 1'b0.
+  function bit split_multireg_name(input string full_name,
+                                   output string base_name,
+                                   output int    index);
+    automatic int underscore_idx = full_name.len() - 1;
+    while (underscore_idx > 0) begin
+      automatic string ch = full_name.getc(underscore_idx);
+
+      if (ch == "_") break;
+
+      // If this isn't a digit, we should fail. Do so by setting the index to zero (so it looks like
+      // we walked past the whole string)
+      if (ch < "0" || "9" < ch) begin
+        underscore_idx = 0;
+        break;
+      end
+
+      underscore_idx--;
+    end
+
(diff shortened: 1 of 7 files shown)


hw/ip/pwm/dv/env/pwm_env_pkg.sv
```diff
@@ -55,10 +55,44 @@ package pwm_env_pkg;
...
+  // 1'b1 if the split is possible (the name ends with an underscore and then some digits).
+  //
+  // If the split is not successful, index is set to 0, base_name is set to full_name and the
+  // function returns 1'b0.
+  function automatic bit split_multireg_name(input string full_name,
+                                             output string base_name,
+                                             output int    index);
+    automatic int underscore_idx = full_name.len() - 1;
+    while (underscore_idx > 0) begin
+      automatic string ch = full_name.substr(underscore_idx, underscore_idx);
```
matutem, line 68: It seems this could just use full_name.getc(underscore_idx).

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -268,39 +268,41 @@ task pwm_scoreboard::compare_trans(int channel);
...
+      ignore_start_pulse[channel]--;
+      continue;
+    end
+
+    // After the state has switched to different state, settings will change. Comparison ignored for
+    // the first SettleTime pulses
+    `DV_CHECK_FATAL(ignore_state_change[channel] <= SettleTime)
+    if (ignore_state_change[channel]) begin
+      ignore_state_change[channel]--;
+      continue;
```
matutem, line 293: This seems redundant given lines 277-280.

hw/ip/pwm/dv/tb.sv
```diff
@@ -52,7 +52,7 @@ module tb;
   `ASSERT(PwmEnTiedHigh_A, cio_pwm_en == '1, clk, rst_n)
 
   for (genvar n = 0; n < PWM_NUM_CHANNELS; n++) begin: gen_pwm_if_conn
-    pwm_if pwm_if(.clk(clk), .rst_n(rst_n), .pwm(cio_pwm[n]));
+    pwm_if pwm_if(.clk(clk_core), .rst_n(rst_core_n), .pwm(cio_pwm[n]));
```
matutem, line 55: This changes behavior. Maybe notice it in the parent github PR.

hw/ip/pwm/dv/env/pwm_env_pkg.sv
rswarbrick: Good point! :-) Thanks!

hw/ip/pwm/dv/tb.sv
rswarbrick: Yep, I can do that: I'll update the text now.

PR #25266, opened 2024-11-20
[hmac,doc] Update for arbitrary key length

martin-velay:
- The HMAC supports arbitrary key lengths. Add precision on how to use it.

### DIFF
--- a/hw/ip/hmac/doc/registers.md
+++ b/hw/ip/hmac/doc/registers.md
@@ -184,7 +184,13 @@ If the software updates the register while the engine computes the hash, the upd
 Key length configuration.
 
 This is a 6-bit one-hot encoded field to configure the key length for HMAC.
-The HMAC supports key lengths of 128-bit, 256-bit, 384-bit, 512-bit and 1024-bit, as long as the key length is not greater than the block size: up to 1024-bit for SHA-2 384/512 and up to 512-bit for SHA-2 256.
+
+The HMAC can be programmed with the following key lengths: 128-bit, 256-bit, 384-bit, 512-bit and 1024-bit.
+But the HMAC supports any arbitrary key length: the software should configure the HMAC with the next largest supported key length and concatenate zeros to reach the programmed key length.
+The position of these zeros depends on the endianness, thus on the programmed [`CFG.key_swap`](registers.md#cfg--key_swap).
+For example, for an 80-bit key, HMAC should be configured with an 128-bit key length, fed with the 80-bit key and with 48 zero-bits.
+
+Note that the key length cannot be greater than the block size: up to 1024-bit for SHA-2 384/512 and up to 512-bit for SHA-2 256.
 The value of this register is irrelevant when only SHA-2 (not keyed HMAC) is configured.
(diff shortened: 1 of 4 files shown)


hw/ip/hmac/doc/registers.md
```diff
@@ -184,7 +184,13 @@ If the software updates the register while the engine computes the hash, the upd
 Key length configuration.
 
 This is a 6-bit one-hot encoded field to configure the key length for HMAC.
-The HMAC supports key lengths of 128-bit, 256-bit, 384-bit, 512-bit and 1024-bit, as long as the key length is not greater than the block size: up to 1024-bit for SHA-2 384/512 and up to 512-bit for SHA-2 256.
+
+The HMAC can be programmed with the following key lengths: 128-bit, 256-bit, 384-bit, 512-bit and 1024-bit.
+But the HMAC supports any arbitrary key length, to do so the software should configure the HMAC to the next largest supported key length and concatenate zeros to reach the programmed key length.
```
gdessouky, line 189: But the HMAC supports any arbitrary key length: the software should configure the HMAC with the next largest supported key length and concatenate zeros to reach the programmed key length.

hw/ip/hmac/doc/registers.md
```diff
@@ -184,7 +184,13 @@ If the software updates the register while the engine computes the hash, the upd
...
+The position of these zeros depends on the endianness, thus on the programmed [`CFG.key_swap`](registers.md#cfg--key_swap).
+E.g. for an 80-bit key, HMAC should be configured with an 128-bit key length, fed with the 80-bit key and with 48 zero-bits.
```
gdessouky, line 191: The position of these zeros depends on the endianness, thus on the programmed [`CFG.key_swap`](registers.md#cfg--key_swap), e.g., for an 80-bit...

gdessouky:
Thanks @martin-velay for catching and updating, LGTM! I left only minor suggestions for nitpicks. 

I think we also need to update `hw/ip/hmac/README.md` - I'd change this "Configurable key length 128/256/384/512/1024-bit secret key for HMAC mode" in the Features section to "Configurable key length up to 1024-bit secret key for HMAC mode", because otherwise now it reads more restrictive than what is effectively supports. Do you think you would update that too please while that PR is still open?

gdessouky (approved):

hw/ip/hmac/doc/registers.md
martin-velay: This sentence was too long to be read easily, that's why I've decided to split it.

martin-velay:
Hi @gdessouky, thanks for your review. I've just pushed with your suggestions (except https://github.com/lowRISC/opentitan/pull/25266#discussion_r1851721051)

hw/ip/hmac/doc/registers.md
gdessouky: Makes sense, perhaps just not start the sentence with E.g.? 

"For example, for an 80-bit key,...."

martin-velay: Done

martin-velay:
@vogelpi , CI is now passing, I think this can be merged. Thanks for your comments.

vogelpi:

This PR just touches documentation which is fine.

PR #25268, opened 2024-11-20
[bazel] Replace Bazel test suites in testplans with targets

AlexJones0:
Some testpoints in the Bazel testplans would link to manually defined test suites of Bazel targets, instead of linking to each individual test target. In this way, testpoints can recursively define a tree of tests via the introduction of test suites. This complicates metric and test tracking tooling, and makes it non-trivial to notice when a test included via a test suite is failing (you can enumerate the targets under a test suite, but this also includes the execution environment, which would have to be removed).

This commit replaces all instances of test suites being used in testplans with the individual `opentitan_test` targets they call (note: these are still generic over execution environments).

Also adds the missing `plic_all_irqs_test_{10,20}` Bazel targets to the `chip_sw_alert_handler_irqs` testpoint.

**Edit**: after review, also removes two unnecessary LC state tests from `//sw/device/tests:rv_dm_jtag_tap_sel`.


hw/top_earlgrey/data/chip_testplan.hjson
```diff
@@ -226,7 +226,12 @@
       lc_states: ["TEST_UNLOCKED", "DEV", "PROD", "RMA"]
       tests: ["chip_tap_straps_dev", "chip_tap_straps_prod", "chip_tap_straps_rma",
               "chip_tap_straps_testunlock0"]
-      bazel: ["//sw/device/tests:rv_dm_jtag_tap_sel"],
+      bazel: ["//sw/device/tests:rv_dm_jtag_tap_sel_test_unlocked1",
```
jwnrt, line 229: Nit: please could you make `lc_states` above match the lifecycle states here?

I'm not sure which one should be updated, @engdoreis what do you think?

jwnrt (approved):
LGTM, thanks

hw/top_earlgrey/data/chip_testplan.hjson
jwnrt: Testpoint descriptions that list `PROD` should almost always support `PROD_END` as well. It's the same as `PROD` except that it can't transition any further (e.g. to `RMA`).

jwnrt: And yes, you can remove them from the `BUILD` file if they're not mentioned in the testplan.

AlexJones0: Latest force push has made this change.

engdoreis (approved):

jwnrt:
Testplan changes only, skipping CI

jwnrt:
@AlexJones0 please could you handle the backport?

PR #25272, opened 2024-11-20
[alert,dv] Fix logic in alert_receiver_driver

rswarbrick:
This was "tightened up" in commit b1095ff, which actually broke things! The problem is that I was imagining the "continue" would continue in the surrounding "forever" loop. But it doesn't! It just skips an iteration of the "repeat" loop inside.

Make the logic a bit more explicit and correctly handle the case where we go into reset with the clock paused.

(@KinzaQamar: Thanks for finding and diagnosing this bug)


alees24 (approved):

rswarbrick:
It gets more embarrassing: my "fix" in this PR didn't actually work! I've just pushed a version that does (because it no longer drops alerts on a reset). Sorry about that.

matutem (approved):
LGTM

PR #25274, opened 2024-11-20
Add hello world imm rom ext

lchiawei:
* Add minimized C code, startup asm and linker script for a standalone IM_EXT build target.
* Add a bazel rule `imm_rom_ext_section`, which will transform binaries into object files and turn it to a target that can be directly used as deps of ROM_EXT's bazel targets.
* Reuse the rule `opentitan_binary` to build IM_EXT and add a attribute to skip the signing stage to the rule. The manifest given as None will turn into the default manifest and keep the signing operation so we need to skip it explicitly.
* Add a map for all IM_EXT target that can be used and make the ROM_EXT targets iterate the map to generate multiple targets.
* Update the ROM_EXT to use in e2e verified boot test.

---

This PR partially addresses https://github.com/lowRISC/opentitan/issues/24368.

The first commit is included in https://github.com/lowRISC/opentitan/pull/25043. I will remove it after it is reviewed and merged.

### DIFF
--- a/sw/device/silicon_creator/rom_ext/BUILD
+++ b/sw/device/silicon_creator/rom_ext/BUILD
@@ -12,6 +12,10 @@ load(
     "//sw/device/silicon_creator/rom_ext:defs.bzl",
     "ROM_EXT_VERSION",
 )
+load(
+    "//sw/device/silicon_creator/imm_rom_ext:defs.bzl",
+    "IMM_ROM_EXT_TARGETS",
+)
 
 package(default_visibility = ["//visibility:public"])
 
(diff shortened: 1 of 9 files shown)


rules/opentitan/cc.bzl
```diff
@@ -339,6 +339,7 @@ opentitan_binary = rv_rule(
             doc = "List of execution environments for this target.",
         ),
         "_cc_toolchain": attr.label(default = Label("@bazel_tools//tools/cpp:current_cc_toolchain")),
+        "skip_signing": attr.bool(default = False),
```
pamaury, line 342: Maybe use the `kind` attribute in `common_binary_attrs` would be more appropriate? There is currently `ram`, `rom` and `flash`. This could be a new one. Maybe @cfrantz can weight in on that?

pamaury:
My memory might be fuzzy but I think that `opentitan_binary` exports not only the signed binary but also the unsigned one, just a different output group. Therefore, I think you might not need a new attribute to skip signing, you just need to teach `imm_rom_ext_section` to look for the unsigned binary. Maybe @cfrantz can help with that, otherwise I can have a look to double-check if that's correct.

lchiawei:
> My memory might be fuzzy but I think that `opentitan_binary` exports not only the signed binary but also the unsigned one, just a different output group. Therefore, I think you might not need a new attribute to skip signing, you just need to teach `imm_rom_ext_section` to look for the unsigned binary (probably use use `ctx.attr.src` and look at `src.output_group["<exec_env>_binary"]`.

I did some experiments, if I reuse the exec env `//hw/top_earlgrey:silicon_creator`, then the signing will cause the following error:
```
ERROR: /usr/local/google/home/lchiawei/opentitan/sw/device/silicon_creator/imm_rom_ext/BUILD:40:17: PreSigningArtifacts sw/device/silicon_creator/imm_rom_ext/hello_world_binaries_silicon_creator.test_key_0.pre-signing failed: (Exit 1): opentitantool failed: error executing PreSigningArtifacts command (from target //sw/device/silicon_creator/imm_rom_ext:hello_world_binaries) bazel-out/k8-fastbuild-ST-13d3ddad9198/bin/sw/host/opentitantool/opentitantool '--rcfile=' --quiet image manifest update ... (remaining 4 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
Error: Image doesn't appear to contain a manifest, or the manifest is corrupted

Caused by:
    Condition failed: `manifest.signed_region_end <= len` (4294967295 vs 1048576)
```

> Alternatively, you might want to use the `kind` attribute (in `common_binary_attrs`) to introduce a new type of binary? And/or: if you introduce a new exec_env (which might make sense here), you can probably create one with no default manifest so it doesn't sign by default?

Actually I did found that the issues only happen when we use `"//hw/top_earlgrey:silicon_creator"` as the exec env, and using an empty manifest in that exec env can solve the issue. As currently all exec envs will derive same binaries and I plan to match the exec env between IM_EXT and ROM_EXT in the future, I will simply use `fpga_cw340` as the only exec env. The exec env for silicon creator with empty manifest will be add in the future. And I also remove the attribute to skip the signing process.

sw/device/silicon_creator/rom_ext/BUILD
```diff
@@ -327,29 +331,33 @@ opentitan_binary(
...
-        "//sw/device/silicon_creator/lib:manifest_def",
-        "//sw/device/silicon_creator/lib/ownership:test_owner",
-        "//sw/device/silicon_creator/lib/ownership/keys/fake",
-    ],
-)
+[
+    opentitan_binary(
+        name = "rom_ext_with_{}_imm_slot_virtual".format(name),
+        testonly = True,
+        # TODO(#22780): Integrate real keys for A1 flows.
```
timothytrippel, line 338: we can remove this todo, it is no longer relevant I think

timothytrippel (approved):
LGTM, thanks! I think we just need to rebase this again and fix the CI failure, then we can merge.

lchiawei:
Hi @timothytrippel ,

Thanks for your review. I've removed those TODOs and rebased my commit.

PR #25280, opened 2024-11-21
[rom_ext] Remove AST patch functionality.

moidx:
The original implementation was intended to be used during early silicon integration. Since then, the provisioning infrastructure and info flash memory map has changed and this funcitonality is no longer compatible.


timothytrippel (approved):

moidx:
CI Failing test `//sw/device/tests:i2c_target_test_fpga_cw310_sival` is not relevant to this PR.

PR #25281, opened 2024-11-21
[rom_ext,dice] build ROM_EXT with DICE CWT cert format support

timothytrippel:
This updates the ROM_EXT build configuration to build multiple ROM_EXTs that support each DICE cert format: X.509 and CWT. Additionally, this updates the `dice_cwt` lib to implement the `dice_cert_check_valid()` function for checking if a DICE CWT is valid or not.

### DIFF
--- a/sw/device/silicon_creator/lib/cert/dice_cwt.c
+++ b/sw/device/silicon_creator/lib/cert/dice_cwt.c
@@ -4,6 +4,7 @@
 
 #include <stdint.h>
 
+#include "include/dice/cbor_reader.h"
 #include "sw/device/lib/base/memory.h"
 #include "sw/device/silicon_creator/lib/base/util.h"
 #include "sw/device/silicon_creator/lib/cert/cbor.h"
@@ -25,6 +26,17 @@
 #include "otp_ctrl_regs.h"  // Generated.
 
 const dice_cert_format_t kDiceCertFormat = kDiceCertFormatCWTAndroid;
+
+enum {
+  // Must match the label used for the public key in
+  // `cwt_dice_chain_entry_payload.hjson`.
+  kDiceCwtSubjectPublicKeyLabel = -4670552,
+
+  // Must match the labels used for the public key X/Y coords in
+  // `cwt_cose_key.hjson`.
+  kDiceCwtCoseKeyXCoordLabel = -2,
+  kDiceCwtCoseKeyYCoordLabel = -3,
+};
 
 enum config_desc_labels {
   kSecurityVersionLabel = -70005,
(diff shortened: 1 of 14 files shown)


moidx (approved):

sw/device/silicon_creator/lib/cert/dice_cwt.c
```diff
@@ -310,10 +322,140 @@ rom_error_t dice_cdi_1_cert_build(hmac_digest_t *owner_measurement,
...
+
+  // Check if CBOR object is array or map. UDS is a COSE_Key which is a map.
+  // CDI_* are COSE_Sign1 objects, which are arrays.
+  size_t num_array_items = 0;
+  CborReadArray(&cbor_obj, &num_array_items);
+  size_t num_map_items = 0;
+  CborReadMap(&cbor_obj, &num_map_items);
+
+  // Extract the public key from the CBOR certificate object.
+  if (num_array_items > 0 && num_map_items == 0) {
```
sasdf, line 428: How about testing the `cert_obj->name == UDS` instead of this num check?

sw/device/silicon_creator/lib/cert/dice_cwt.c
```diff
@@ -310,10 +322,140 @@ rom_error_t dice_cdi_1_cert_build(hmac_digest_t *owner_measurement,
...
+  } else if (num_map_items > 0 && num_array_items == 0) {
+    RETURN_IF_ERROR(extract_pubkey_from_cose_key(cert_obj->cert_body_p,
+                                                 cert_obj->cert_body_size,
+                                                 &x_coord_p, &y_coord_p));
+  } else {
+    return kErrorDiceCwtCoseKeyNotFound;
+  }
+
+  // Compare the public key in the certificate to the public updated key.
+  for (size_t i = 0; i < kEcdsaP256PublicKeyCoordWords; ++i) {
```
sasdf, line 447: How about `memcmp`?

timothytrippel: The way I wrote it is slightly faster, since we only iterate over 32 byte array once.

sw/device/silicon_creator/lib/cert/dice_cwt.c
timothytrippel: I thought about that, however, in the future, we aim to support other DICE CWT profiles (see RFC: https://docs.google.com/document/d/1zBudM3YyasCoYhchRDLJUKKDEtlCQ4Yc5dx9HEE45HU/edit?usp=sharing), in which case the UDS might be a COSE_Sign1 object, not just. COSE_Key. so this is more flexible towards that.

PR #25288, opened 2024-11-21
[i2c,dv] Add missing "void" type to a function in i2c_monitor

rswarbrick:
The default type (bit) causes VCS to emit a warning about the fact we call the function without looking at the return value. Tell VCS that there isn't one!


matutem (approved):
Good catch, LGTM

PR #25291, opened 2024-11-21
[ci] Lint for "origin" lines in commit messages

jwnrt:
This line indicates whether a commit is original to this branch or was cherry-picked from another.

Signed-off-by: James Wainwright <james.wainwright@lowrisc.org>
(commit is original to earlgrey_1.0.0)


andreaskurth (approved):
LGTM technically, though please hold off with merging until all TZs have had a chance to read the announcement email and reply in case of concerns.

Thanks @jwnrt!

timothytrippel:
I would voice the opinion that having a special commit message for commit messages that originate on the `earlgrey_1.0.0` branch is overkill, provides little value, and is going to cause a lot of toil for two reasons:
1. we don't do this on any other branch (including master), and
2. there is no hook to automatically apply it when working on one branch, but not apply it when working on another.

Also what value does it provide? I understand the value of the `cherry-picked from ...` message, and there is automation to apply this.

andreaskurth (changes requested):
Let's provide an automation for the `(commit is original ..` message on the dev side; please hold off until then. I'll follow up offline

timothytrippel:
Thanks @jwnrt @andreaskurth; I discussed offline with @andreaskurth : I am OK with the commit message if there is a way for a git hook to automatically apply it based on what branch I am working on. Specifically, I am also OK if the target branch needs to be specified on the command line once per branch I create, e.g.,

```
git checkout -b contribution-to-bringup
git target-branch set earlgrey_1.0.0
```

jwnrt:
I've added a commit hook. Instead of requiring an extra config step, it looks to see if the current branch is a descendent of `earlgrey_1.0.0`.

timothytrippel:
awesome, thanks @jwnrt !

jwnrt:
@andreaskurth could I suggest that we change the format to `Commit-original-to: <branch>` (or similar)? This is the standard format for commit metadata and Git has built-in tooling for dealing with it.

This is what we use for `Signed-off-by: ...` and `Co-authored-by: ...` lines.

Contributors could add an alias like the following to their Git configs to explicitly put the line to the bottom of a commit:

```toml
[alias]
commit-original = "!git commit --trailer \"Commit-original-to:\"$(git rev-parse --abbrev-ref HEAD)\"\""
```

```console
$ git commit-original -s -m "[doc] Add images"

commit 8a50ee29c4314914cc4e67fb44fd47ecf2333c65 (HEAD -> earlgrey_1.0.0)
Author: James Wainwright <james.wainwright@lowrisc.org>
Date:   Sat Jan 18 21:26:04 2025 +0000

    [doc] Add images
    
    Signed-off-by: James Wainwright <james.wainwright@lowrisc.org>
    Commit-original-to: earlgrey_1.0.0
```

The drawback is that the message looks different to the `(commit cherry picked from $hash)` messages which arguably is a mistake from Git for not using the standard trailer format.

a-will (changes requested):
Can we please discuss what actual value this provides before we make development even more complicated?

I've seen nothing here that justifies people needing to do more than the ordinary sign-off. There have only been offline conversations, from what I can see, and adding more bespoke metadata makes me very grumpy. ;)

jwnrt:
IIRC the motivation is to find commits that needed to be ported to `master` but were forgotten.

FWIW I would also lean towards this being more troublesome than beneficial especially since there are many commits without this line already.

Presumably it has to go through RFC first (I don’t know if it has already) and this script was a PoC for that RFC.

PR #25294, opened 2024-11-21
[imm_rom_ext] Mint CDI_0 certs in immutable rom ext

sasdf:
Since CDI_0 attests the integrity of rom_ext, this PR moves the UDS & CDI_0 stage to the immutable rom_ext part, which will be executed before mutable rom_ext, to establish root trust in hardware.

SKU owners can choose whether the immutability should be enforced using the `CREATOR_SW_CFG_IMMUTABLE_ROM_EXT_EN` OTP field when the feature is ready. 

---

This PR also changes the ROM address translation e2e test from the `*_rom_with_fake_keys` exec env to `*_rom_ext` env, since the key manager is not functional without secret2.

To avoid ROM_EXT affecting ROM test results, we also modified the ROM_EXT boot message pattern to catch the immutable message, which will be printed first.


sw/device/silicon_creator/rom_ext/rom_ext_manifest.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "sw/device/lib/base/macros.h"
+#include "sw/device/silicon_creator/lib/manifest.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+/**
+ * Return the pointer to rom_ext manifest on the active flash bank.
```
timothytrippel, line 16: maybe you could drop a comment here why this function is different than https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/silicon_creator/lib/manifest_def.c;l=39?q=manifest_def_get&ss=opentitan%2Fopentitan  (therefore, why is is needed in addition to the current solution for getting the manifest of the currently executing slot)?

timothytrippel (approved):

sw/device/silicon_creator/imm_rom_ext/imm_rom_ext.c
```diff
@@ -4,7 +4,59 @@
...
+  // specific ROM version.
+  sec_mmio_check_values(rnd_uint32());
+
+  // Initialize Immutable ROM EXT.
+  sec_mmio_next_stage_init();
+  // Configure UART0 as stdout.
+  pinmux_init_uart0_tx();
+  uart_init(kUartNCOValue);
+
+  dbg_printf("IMM_ROM_EXT v0.1\r\n");
```
cfrantz, line 37: Please structure this print as "key:value".  e.g.: `IMM_ROM_EXT:0.1`.

(once we finish merging ROM_EXT changes over from `earlgrey_1.0.0`, all ROM_EXT prints are structured this way).

sw/device/silicon_creator/imm_rom_ext/imm_rom_ext.c
```diff
@@ -4,7 +4,59 @@
...
+
+  // Establish our identity.
+  const manifest_t *rom_ext = rom_ext_manifest();
+  HARDENED_RETURN_IF_ERROR(dice_chain_init());
+  // TODO: Move UDS cert check to mutable ROM_EXT.
+  HARDENED_RETURN_IF_ERROR(dice_chain_attestation_silicon());
+  HARDENED_RETURN_IF_ERROR(
```
cfrantz, line 44: In the ROM_EXT, we have to initialize ownership after the silicon stage but before the creator stage.  That is because we bind the ownership configuration to the chip with a keymgr derived KMAC key.

I think at a minimum, we need to generate a sealing key for KMAC here.  See `ownership_seal_init` on the `earlgrey_1.0.0` branch.

sw/device/silicon_creator/imm_rom_ext/e2e/boot_message/boot_test.c
```diff
@@ -0,0 +1,9 @@
+// Copyright lowRISC contributors (OpenTitan project).
```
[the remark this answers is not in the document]
sasdf: Nice idea! thanks.

sw/device/silicon_creator/imm_rom_ext/imm_rom_ext.c
sasdf: Added `ownership_seal_init` call between the silicon and the creator stages.

PR #25304, opened 2024-11-21
[provisioning] Encrypt the RMA unlock token

cfrantz:
1. During provisioning, encrypt the RMA unlock token for long term storage in a devices database.
2. Support generating a random RMA unlock token.
3. Add fake RSA keys to support encrypting the token during provisioning test flows.


moidx (approved):

timothytrippel (approved):

sw/host/provisioning/orchestrator/src/orchestrator.py
```diff
@@ -92,7 +92,6 @@ def main(args_in):
     )
     parser.add_argument(
         "--rma-unlock-token",
-        required=True,
```
timothytrippel, line 95: We should probably make this parameter optional in the orchestrator.py script too: https://cs.opensource.google/opentitan/opentitan/+/earlgrey_1.0.0:sw/host/provisioning/orchestrator/src/ot_dut.py;l=166

PR #25308, opened 2024-11-22
[openocd] fix openocd build on new versions of GCC

timothytrippel:
We configure GCC to treat all warnings as errors and newer versions warn when calloc args are tranposed. This adds a patch that fixes the issue.


moidx (approved):

cfrantz (approved):
Maybe pick this over to `earlgrey_es_sival` as well.

PR #25309, opened 2024-11-22
[ci] Only run FPGA tests on pull requests.

moidx:
This is to minimize the load on the CI infrastructure. Post-submit testing should be time bound to no more than a few  minutes on resources with very limited availability. This change takes the drastic measure of not running any FPGA tests, but a reduced test suite for post submit can be added in future (e.g. a `test_suite` with optimized coverage).


elliotb-lowrisc:
Closing as no longer relevant (Azure)

PR #25316, opened 2024-11-22
[ci] remove stale and broken release workflow

### LINKED ISSUE #25255: [ci] remove release job and workflows
### Description

`release/azure-pipelines-release.yml` seems to be constantly failing, and its usefulness should have already be replaced with bitstream cache.

deploy_release_artifacts of `azure-pipelines.yml` seems to be no longer uploading anything to tags, and the usefulness is also questionable.

We should remove them, and if they're still needed, recreate a properly working workflow.

nbdd0121:
This has been broken for a while, and nobody seems to be reporting it or fixing it, so remove the workflow all-together.

I fixed the Bazel release target (tested locally) regardless.

Close #25255


jwnrt (approved):

nbdd0121:
Skipping bitstream/FPGA builds since this PR won't affect these.

PR #25319, opened 2024-11-22
[darjeeling] Add missing padring.waiver

Razer6:
This file was missing. Together with that and the FUSESOC_IGNORE move you can lint Darjeeling.


andreaskurth (approved):

andreaskurth:
Skipping remaining CI checks as they aren't relevant for this change

PR #25322, opened 2024-11-22
[sival] Assign `gpio_pinmux_test` to two points

jwnrt:
This test checks various muxing configurations, plus one attribute (invert), which fulfils most of the testpoints here. More attributes can be added later if necessary.


luismarques (approved):

jwnrt:
FPGA failure unrelated to these testplan updates, skipping

PR #25325, opened 2024-11-22
[ci] Backport the Azure -> GitHub migration to `earlgrey_es_sival`

jwnrt:


jwnrt:
I think this is going to be too difficult to get working. We'll probably recreate the whole `.github` workflow in one go and delete `azure-pipelines.yml` in one go.

PR #25329, opened 2024-11-22
[crypto] Adjust error handling in P-384.

jadephilipoom:
This brings P-384 more in line with P-256, so that on invalid inputs it returns a bad status code instead of throwing a fault and locking OTBN. It also pushes point-validity checks closer to their use point, which helps with the status code construction as well as making misuse less likely.

This groundwork will help me consolidate the P-384 binaries, like I did for P-256 in https://github.com/lowRISC/opentitan/pull/25326


moidx (approved):
LGTM.

PR #25358, opened 2024-11-22
[opentitantool] Convenience for external programs

jesultra:
Add two new commands to `opentitantool`: `spi flashrom-arg` and `uart os-device`, which can be used for heling external programs access HyperDebug (or some other backend devices) without going through `opentitanlib`, while still using the `usb-serial` and port aliases from our configuration files to identify the debugger and port.

```
$ opentitantool --conf ... --usb-serial 2042355F3236 spi --bus=BOOTSTRAP flashrom-args
{
  programmer: "raiden_debug_spi:serial=2042355F3236,target=2"
}
$ opentitantool --conf ... --usb-serial 2042355F3236 uart --uart=EC os-device
{
  path: "/dev/ttyUSB6"
}
```

### DIFF
--- a/sw/host/opentitanlib/src/app/spi.rs
+++ b/sw/host/opentitanlib/src/app/spi.rs
@@ -205,6 +205,13 @@ impl Target for LogicalSpiWrapper {
         self.physical_wrapper.underlying_target.set_voltage(voltage)
     }
 
+    fn get_flashrom_programmer(&self) -> Result<String> {
+        self.apply_settings_to_underlying()?;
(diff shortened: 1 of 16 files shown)


jesultra:
> This looks good to me. I am not sure `os-device` name, maybe `device-path` would be clearer? It's a really small detail, feel free to ignore it.

The choice was not clear to me either, but "device" is such an overloaded word, I wanted to convey the fact that we are not talking about backend/interface/driver in the sense of opentitanlib, but something meaningful to the "outside" OS.

Issue #13098, opened 2022-06-07
[OpenTitan-tool] Bootstrapping should trigger a POR to ensure state is uniform at the start of tests


drewmacrae:
After bootstrapping if opentitan-tool triggers a POR, tests will always start with just the POR bit set. This allows for simple tests that use the reset_info to determine that they're running the test software for the first time, and could resolve #13096

a-will:
The answer from the chip-level test meeting was "yes"

drewmacrae:
I'm fairly certain it's currently mapped to a physical switch that we don't have control over. Should we remap it? Who should we ask about allocating an FPGA pin to this?

a-will:
> I'm fairly certain it's currently mapped to a physical switch that we don't have control over. Should we remap it? Who should we ask about allocating an FPGA pin to this?

If it is only mapped to the button right now, I'll take care of adding the GPIO to SAM3X.

alphan:
I'm not sure if we really need this. Is there a reason why tests can't just check if the POR is set, instead of requiring that **only** the POR bit is set? Once they start, they clear the reset reasons anyway.

a-will:
> I'm not sure if we really need this. Is there a reason why tests can't just check if the POR is set, instead of requiring that **only** the POR bit is set? Once they start, they clear the reset reasons anyway.

Because the other reset reasons are reserved for the test to modify, and we do not require bootstrapping to occur as part of the test methodology across all platforms.

Issuing a POR is an extremely simple thing to do, and this expectation has already been baked in to tests. Let's just get it done.

a-will:
> I think there are three unused pins here: https://cs.opensource.google/opentitan/opentitan/+/master:hw/top_earlgrey/data/pins_cw310.xdc;drc=a35f7687666385502591fafed977ab83475d19b6;l=71

For the pin, we'll likely be using whatever is the highest unused USB_A* pin, possibly with inverted polarity (so the default is to not be in reset). The section you've got there is for a header, which will be useful for the hyperdebug version later.

a-will:
It turns out the POR is already hooked up on CW310 via USB_A14. This pin was pretending to be a JTAG TAP reset, but because it is also used as a power-on reset, we're actually lacking a TAP reset pin. _That_ cleanup will happen with the rest of the JTAG pins, several of which currently are multiplexed on top of the SPI pins.

alphan:
#13277 should resolve this as a side effect.

a-will:
[...]

PR #25362, opened 2024-11-22
[sival,tests] Fix old TODOs and other enhancements

matutem:
This has a few independent commits:
- Remove workaround for #13098
- Remove workaround for #20798 and the test that reproduces the (fixed) bug
- Remove workaround for issue #20747 which requires updating the test template
- Update comments in rstmgr testplan to match tests' si_stage
- Remove TODO in clkmgr_off_trans and enhance it and related test to check crash dump's fault PC
- Remove some redundant comments


sw/device/tests/BUILD
```diff
@@ -3219,55 +3219,6 @@ opentitan_test(
     ],
 )
 
-# This test reproduces a silicon failure, but passes in all other environments.
-# TODO(lowrisc/opentitan#20798)  Remove this test and extend
-# pwrmgr_random_sleep_all_wake_ups to cover all wakeups once this issue is
-# fixed.
-opentitan_test(
```
engdoreis, line 3224: Don't we want to keep the test to the test Silicon and also to make sure that the bug will never show up again?

engdoreis: I see, thanks for explaining.

engdoreis (approved):
LGTM

PR #25363, opened 2024-11-23
[fpga] Add flow and constraints to dump the flash info array cell sites

a-will:
Adjust MMI dump to use a regex for mem type. Some memories in the FPGA can consist of both RAMB36 and RAMB18 types, depending on the width of the array. Use a regex to allow more flexible memory type detection.

Add KEEP_HIERARCHY to the prim_ram_1p instances representing the flash info arrays. This helps keep intelligible hierarchical paths to the memories, so we can readily select the correct cells. Because Vivado was previously combining all info types of the same bank into the same BRAMs, this increases BRAM utilization a bit.

~Add calls to dump an MMI file for each array represented by a `(bank,type)` tuple, and add them to the bitstream cache entry.~ Collect all MMI data into a single file, and update the manifest and bazel rules to use it. Instead of having separate files, each map has a separate Processor element, and the InstData is assigned the key that identifies the memory (instead of using "dummy").


a-will:
CC @moidx 

Still a WIP, and I haven't tested splicing any particular image. We might be able to create a more targeted Tcl proc that could aggregate arrays (e.g. concatenate the two banks for each type into a linearized address space).

rules/scripts/bitstreams_manifest_v2.example.json
a-will, line 1: Perhaps I ought to use these example files as test vectors, instead of making something else up inside the self-test Python script. But I'd punt that change for now ;)

rules/scripts/bitstreams_manifest.example.json
```diff
@@ -8,14 +8,12 @@
...
-        "rom": {
-          "file": "rom.mmi",
-          "build_target": "//hw/bitstream/vivado:fpga_cw310"
-        }
+        "file": "memories.mmi",
+        "build_target": "//hw/bitstream/vivado:fpga_cw310",
+        "memories": [
+          "otp",
+          "rom"
+        ]
```
a-will, line 16: This information isn't currently used in the bazel rules (not part of bitstreams_workspace.py), but I figured I'd make this info readily available in case it's useful.

The same info could be obtained from the MMI file directly, so if reviewers prefer I remove it from the manifest, I certainly can.

hw/top_earlgrey/util/vivado_hook_write_bitstream_pre.tcl
```diff
@@ -177,15 +185,60 @@ switch ${fpga_family} {
...
 # that its data input overruns the address space. The workaround is to pretend
 # the address space is 16 times larger than we would normally compute.
 set otp_brams [split [get_cells -hierarchical -filter " PRIMITIVE_TYPE =~ ${bram_regex} && NAME =~ *u_otp_ctrl*"] " "]
-generate_mmi "otp.mmi" $otp_brams "RAMB18" 0 16 2
+dict set memInfo otp brams $otp_brams
+dict set memInfo otp mem_type_regex $mem_type_regex
+dict set memInfo otp fake_word_width 0
+dict set memInfo otp addr_end_multiplier 16
+
+# The flash banks have 76-bit wide words. 64 bits are data, and 12 bits are metadata.
```
moidx, line 204: Will `updatemem` support 76-bit wide words per line? It seems like it should, but I am curious to know if you have spotted any quirks when looking at the synthesis results.

rules/scripts/bitstreams_manifest.example.json
moidx: Thanks, I think it is useful to know. In the future it would be interesting for us to consider creating a Bazel target that splices all memories in a single Bazel invocation.

moidx (commented):
LGTM. Thanks @a-will, these changes are very useful! Some nice refactorings as well.

moidx (approved):

hw/top_earlgrey/util/vivado_hook_write_bitstream_pre.tcl
a-will: I think it should, but I don't currently have an image that lets me check. Perhaps next we should cobble together something to test! :)

matutem (approved):
Alex, this seems to improve the MMI flow in addition to adding support for flash image uploading. I would only request you to make sure the docs about this area are adequate to navigate it for folks that need to work with images. This may be tricky since you are very familiar with it all, but things may look different if you pretend you don't :)

LGTM

PR #23472, opened 2024-06-03
Port `sleep_pin_wake` to SiVal

luismarques:

PR #25365, opened 2024-11-24
Port sleep_pin_wake to SiVal

luismarques:
This is a manual cherry-pick of https://github.com/lowRISC/opentitan/pull/23472


matutem (approved):
It seems the flow of the test diverged between simdv and sival: in simdv the CPU seems to determines the sequence, and in sival it is the host. Ideally in simdv it would be the vseq that chooses and communicates it to the vseq via backdoor symbol overwrites to the same variables that sival uses after ujson has sorted things out. Perhaps add an issue for this?

Otherwise LGTM

PR #25369, opened 2024-11-24
[ci] limit cherry-pick workflow concurrency per PR

nbdd0121:


.github/workflows/cherrypick.yml
```diff
@@ -10,6 +10,12 @@ on:
 permissions:
   contents: read
 
+# 2 backport jobs may be running simultaneously if the merge and label event happen in quick
+# succession. Serialize them based on PR number to avoid this.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
```
jwnrt, line 17: What happens when this is false, the old job continues and the new one never starts?

nbdd0121: The old job continues, the new one will start when the old one completes and the just do nothing.

jwnrt (approved):

PR #25370, opened 2024-11-25
[templates] Remove unused definitions

Razer6:
This PR removes unused variable definitions from the chip-level template.


andreaskurth (approved):
LGTM; generated files didn't change which should be enough proof that this code is indeed unused

PR #25377, opened 2024-11-25
[Sival pwm] Add a pwm smoketest

engdoreis:


rswarbrick (commented):
Some notes on the first two commits.

hw/top_earlgrey/data/ip/chip_pwm_testplan.hjson
```diff
@@ -0,0 +1,23 @@
...
+// SPDX-License-Identifier: Apache-2.0
+{
+  name: pwm
+  testpoints: [
+    {
+      name: chip_sw_pwm_smoketest
+      desc: '''Verify that the PWM outputs correct wave.
+
+              - Program the PWM to output waves of 1kHz, 5kHz and 10kHz.
+              - Program the duty cycle to 0%, 20%, 50%, 90%, and 100%.
```
alees24, line 12: I think these numbers need updating to match what the test actually achieves.

sw/device/tests/pwm_smoketest.c
```diff
@@ -0,0 +1,104 @@
...
+#include "sw/device/lib/testing/test_framework/check.h"
+#include "sw/device/lib/testing/test_framework/ottf_main.h"
+#include "sw/device/lib/testing/test_framework/ottf_utils.h"
+
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+
+OTTF_DEFINE_TEST_CONFIG();
+
+static volatile const uint8_t kClocksHz[] = {20, 50, 100, 200};
+static volatile const uint8_t kDutyCycles[] = {10, 30, 50, 70, 90};
```
alees24, line 22: If the duty cycles were chosen to by power-of-two values rather than decimal values (e.g. fixed-point fractions of 0x10000 as per the IP block specification) then the errors would be smaller.
I realize that the interface provided by the DIF mismatches that of the register-level API and hardware implementation, but the precision would be improved by sticking duty cycles that are integral numbers in base 2 (or 16).

sw/device/tests/pwm_smoketest.c
```diff
@@ -0,0 +1,104 @@
...
+
+static dif_pwm_config_t compute_clk_config(uint32_t pwm_clk) {
+  enum {
+    // Theres a trade off on the precision of the clock and the precision of the
+    // dutycycle, the higher the kDutyCycleResulution the more precise is the
+    // the dutycycle and less precise is the pwm frequency.
```
alees24, line 28: If you choose to stick with decimal percentages, then I think it's worth clarifying that the rounding issues result from using percentages, please.
Also, I think the sentence "By experimentation..." was meant to specify the relevant value of kDutyCycleResolution, presumably the value of 4 as per the code?

Spelling: _resolution_ in a few places, and duty cycle (two words), please.

sw/device/tests/pwm_smoketest.c
```diff
@@ -0,0 +1,104 @@
...
+  CHECK_DIF_OK(dif_pinmux_init(addr, &pinmux));
+  CHECK_DIF_OK(dif_pinmux_output_select(&pinmux, kTopEarlgreyPinmuxMioOutIoa8,
+                                        kTopEarlgreyPinmuxOutselPwmAonPwm0));
+
+  dif_gpio_t gpio;
+  addr = mmio_region_from_addr(TOP_EARLGREY_GPIO_BASE_ADDR);
+  CHECK_DIF_OK(dif_gpio_init(addr, &gpio));
+  CHECK_DIF_OK(dif_gpio_output_set_enabled_all(&gpio, 0x1));
+
+  LOG_INFO("Use IOA7 to let host know when sleep is active.");
```
alees24, line 56: I think this message maybe needs altering since this test doesn't enter sleep unlike the chip-level DV one. Also, if I'm understanding correctly, the host is using that GPIO signal to indicate to the device test code on the Ibex that it is sampling?

sw/device/tests/pwm_smoketest.c
```diff
@@ -0,0 +1,104 @@
...
+
+      CHECK_DIF_OK(dif_pwm_configure(&pwm, pwm_config));
+      CHECK_DIF_OK(
+          dif_pwm_configure_channel(&pwm, kDifPwmChannel0, channel_config));
+
+      // The goes low when the host is sampling.
+      bool not_sampling = true;
+      do {
+        CHECK_DIF_OK(dif_gpio_read(&gpio, 0, &not_sampling));
+        if (not_sampling) {  // Debauce
```
alees24, line 83: Debounce?

sw/device/tests/pwm_smoketest.c
```diff
@@ -0,0 +1,104 @@
...
+          .polarity = kDifPwmPolarityActiveHigh,
+          .blink_parameter_x = 0,
+          .blink_parameter_y = 0,
+      };
+
+      CHECK_DIF_OK(dif_pwm_configure(&pwm, pwm_config));
+      CHECK_DIF_OK(
+          dif_pwm_configure_channel(&pwm, kDifPwmChannel0, channel_config));
+
+      // The goes low when the host is sampling.
```
alees24, line 79: Here and below, I think this should have said something like 'GPIO line?'

alees24 (changes requested):
I think the device side code is largely correct, thank you, although it may be helped by choosing values that better match the hardware implementation. The comments in the source could use some proofreading and correction, please.
I'm probably not best-placed to review the Rust source.

sw/host/opentitanlib/src/test_utils/bitbanging/pwm.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    impl<const PIN: u8> Decoder<PIN> {
+        /// Loop sampling the pin until a rising edge is detected.
+        fn find_first_active_level<I>(&self, samples: &mut I) -> Result<()>
+        where
+            I: Iterator<Item = Sample<PIN>>,
+        {
+            let _ = samples
+                .by_ref()
+                .find(|sample| sample.pin() == Bit::Low)
```
nbdd0121, line 41: Should this be `!self.active_level` instead of hard-coded low?

sw/host/opentitanlib/src/test_utils/bitbanging/pwm.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        fn decode_period<I>(&self, samples: &mut I) -> Option<PwmPeriod>
+        where
+            I: Iterator<Item = Sample<PIN>>,
+        {
+            let num_active_samples =
+                samples.position(|sample| sample.pin() != self.active_level)? + 1;
+
+            let num_inactive_samples =
+                samples.position(|sample| sample.pin() == self.active_level)? + 1;
```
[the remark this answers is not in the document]
nbdd0121: The `find_first_active_level` is called first so the first `self.active_level` should be consumed already.

sw/host/opentitanlib/src/test_utils/bitbanging/pwm.rs
nbdd0121: Actually, can you just call `decode_period` once initially and discard the result?

sw/device/tests/pwm_smoketest.c
```diff
@@ -0,0 +1,99 @@
...
+
+OTTF_DEFINE_TEST_CONFIG();
+
+static volatile const uint8_t kClocksHz[] = {24, 48, 96};
+static volatile const uint8_t kDutyCycles[] = {11, 36, 50, 71, 91};
+
+static dif_pwm_config_t compute_clk_config(uint32_t pwm_clk) {
+  enum {
+    kDutyCycleResolution = 4,
+    kBeatsPerCycle = 2 << (kDutyCycleResolution + 1),  // 2 ^ (DC_RESN + 1)
```
alees24, line 27: The comment and code here are inconsistent; it should be `1 << (kDutyCycleResolution + 1)` or `2 << kDutyCycleResolution` for the same semantics as `dc_resn` in the HW IP documentation/register API. The behavior of the code is unaffected if you change '4' to '5' because the `...Resolution` value is not used elsewhere directly, but this would aid comprehension.

alees24 (approved):
Thanks for making the device-side changes. Good to see this working.

jwnrt (approved):
I think the Rust side LGTM, thanks

PR #25387, opened 2024-11-25
[provisioning] Log data from provisionging as json

cfrantz:
1. Collect personalization data and timing statistics during personalization and log it as json when provisioning is done.


cfrantz:
TODO:
- log the full device_id.
- log the encrypted rma unlock token.
- give the provisioning extensions the ability to log data.

timothytrippel (approved):

moidx (approved):

PR #25388, opened 2024-11-26
[manuf] refactor CP init to read out wafer data and device ID

### LINKED ISSUE #24600: [manuf] do not erase flash info page 0 in CP init
The CP init binary currently erases flash info page 0 and reconfigures its settings before writing the DeviceID, manuf state, and AST configuration data to it. It is possible however that some of this data may be written to flash info page 0 during CP calibration stage. Therefore, during CP init stage, flash info page 0 should be buffered to SRAM before reconfiguring and writing any additional data to the page.

timothytrippel:
When CP init runs, flash info page 0 will already be provisioned with wafer data from prior CP stages. This updates the CP init and subsequent FT provisioning stages to assume that is the case and pull appropriate data out of this flash info page accordingly.

This fixes #24600.


sw/host/provisioning/cp_lib/src/lib.rs
```diff
@@ -129,7 +120,13 @@ pub fn run_sram_cp_provision(
     )?;
 
     // Inject provisioning data into the device.
-    provisioning_data.send(spi_console)?;
+    data_in.send(spi_console)?;
+
+    // Wait to receive CP device ID.
+    let _ = UartConsole::wait_for(spi_console, r"Exporting CP device ID ...", timeout)?;
+    let _data_out = ManufCpProvisioningDataOut::recv(spi_console, timeout, true)?;
```
timothytrippel, line 127: @cfrantz @moidx We should update this to log (via JSON) the CP portion of the device ID so the orchestrator script can pick this up and update it with the FT portion (https://docs.google.com/document/d/1zVgZjLPMiBQhainWnyFjJx_fTnYx-8HpDJwLQHhltyo/edit?usp=sharing)

sw/device/silicon_creator/manuf/lib/flash_info_fields.h
```diff
@@ -21,20 +21,57 @@ typedef struct flash_info_field {
 
 enum {
   /**
-   * AST Calibration Data Size - Bank 0, Page 0
+   * Lot Name Data Start / Size - Bank 0, Page 0
+   */
+  kFlashInfoFieldLotNameStartOffset = 80,
```
timothytrippel, line 26: @benjbender please confirm these offsets

sw/host/provisioning/cp_lib/src/lib.rs
timothytrippel: to match #25387

sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -2,6 +2,8 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+load("@//rules:signing.bzl", "offline_presigning_artifacts", "offline_signature_attach")
+load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
```
cfrantz, line 6: Using this?

cfrantz (approved):

sw/device/silicon_creator/manuf/base/BUILD
timothytrippel: Ah looks like those just got moved around as a part of `bazel run //quality:buildifier_fix` running. Not sure why CI didn't catch that before. They were already in place though.

sw/device/silicon_creator/manuf/base/sram_cp_provision.c
```diff
@@ -61,99 +62,102 @@ static void manually_init_ast(uint32_t *data) {
...
-  TRY(flash_ctrl_testutils_write(
-      &flash_ctrl_state,
-      byte_address + kFlashInfoFieldAstCalibrationData.byte_offset,
-      kFlashInfoFieldAstCalibrationData.partition, ast_cfg_data,
-      kDifFlashCtrlPartitionTypeInfo,
+      &flash_ctrl_state, kFlashInfoFieldCpDeviceId.page,
+      kFlashInfoFieldCpDeviceId.bank, kFlashInfoFieldCpDeviceId.partition,
+      kFlashInfoPage0Permissions, /*offset=*/NULL));
+
+  // Read (wafer) lot name.
```
timothytrippel, line 72: @benjbender if you could enlighten me on the encoding of the lot name, I could address the "TODO" below on line 114, and validate the second word of the CP device ID contains the correct "year", "week", and "lot number" data. Thanks!

sw/device/silicon_creator/manuf/lib/flash_info_fields.h
timothytrippel: Synced with @benjbender offline to confirm these.

sw/device/silicon_creator/manuf/base/sram_cp_provision.c
timothytrippel: Synced with @benjbender offline to confirm the format of this; I have updated the flow to write the CP device ID to flash so it will not be expected to be there already.

sw/host/provisioning/cp_lib/src/lib.rs
cfrantz: How about something like this:
```
#[derive(Debug, Clone, Default, Serialize)]
struct ChipProbeResponse {
  pub device_id: String,
  // can put other stats in here if needed.
}

let mut response = ChipProbeResponse::default();
response.device_id = data_out
        .device_id
        .iter()
        .map(|v| format!("{v:08X}"))
        .collect::<Vec<String>>()
        .join("");

let doc = serde_json::to_string(&response)?;
println!("CHIP_PROBE_DATA: {doc}");
```

timothytrippel: Ah yes, thanks! updated. @moidx @cfrantz just a heads up the CP device ID will be logged as a little-endian string. I can update it if you prefer parsing a big-endian string, just let me know.

cfrantz: Lets transform the logged data to whichever format is like the command-line argument to the `ft` harness (which, btw, is the same form the FT logged data uses, and uses the same conversion code above).

timothytrippel: ah ok, I think that is big-endian format: https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/provisioning/util_lib/src/lib.rs;l=20?q=hex_string_to_u32_arrayvec&ss=opentitan%2Fopentitan

I updated the PR to log the CP device ID response in big-endian.

moidx (approved):

PR #25391, opened 2024-11-26
[crypto] Consolidate P-384 OTBN binaries.

jadephilipoom:
Analogous to the P-256 refactor in https://github.com/lowRISC/opentitan/pull/25326 but more dramatic, because P-384 was split into more pieces previously. With larger IMEM, we can use only one binary for P-384 operations, instead of the previous four (keygen/sign/verify/ecdh).

There's again a lot of code churn here, but the vast majority of it is only moved around, not changed. Like with P-256, some names had to be changed to fit the naming convention of definitions from the same C header starting with the same prefix (so e.g. "ecdsa_p384" became "p384_ecdsa").

This saves about 16.4 kB (!) of code size in the cryptolib.


moidx (approved):
Thanks @jadephilipoom. LGTM. 

I take it that this passes cryptotest?

jadephilipoom:
> I take it that this passes cryptotest?

P-384 isn't currently set up with cryptotest -- at the time Ryan went through and set up most of the existing tests, it wasn't fully hooked up and available in cryptolib. But for the analogous P-256 changes, yes, everything passes cryptotest :slightly_smiling_face:

PR #25393, opened 2024-11-26
[SiVal] Run broken tests in the sival nightlies

engdoreis:


engdoreis:
I removed the commit that was adding the name of the runner to the report name, we will add it inside the report.xml in a follow-up PR.

PR #25394, opened 2024-11-26
[regtool] Add inline veriblelint waivers for line-length rule in ral_pkg

vogelpi:
The ral_pkg files are all auto-generated and depending on the name and exact instantiation of shadow registers (potentially `hwext`), may contain pretty long path names. To avoid veriblelint flagging warnings about this (it feels silly as the files are auto-generated anyway), this commit adds a suitable inline waiver specifically for shadow registers paths.

### DIFF
--- a/util/reggen/uvm_reg_base.sv.tpl
+++ b/util/reggen/uvm_reg_base.sv.tpl
@@ -601,10 +601,10 @@ reg_field_name, field)">\
         assert 0
 %>\
       ${reg_inst}.add_hdl_path_slice(
-          "${shadowed_reg_path}.committed_reg.q",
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
Looks good to me.

I'm slightly surprised that we care though: why are we running `veriblelint` in the build tree?

vogelpi:
> Looks good to me.
> 
> I'm slightly surprised that we care though: why are we running `veriblelint` in the build tree?

Thanks for the feedback @rswarbrick . Yes, it's a bit surprising. But thinking about this again: we also run lint on the auto-generated register files and maybe also on auto-generated C code etc. The difference is just that we also check in those files while we don't check in the generated ral code.

PR #25400, opened 2024-11-26
[fpga] Add ability to issue POR to self via SW

a-will:
Co-opt the pad attributes in sensor_ctrl for FLASH_TEST_MODE0 to enable firmware on the FPGA to reset itself with a POR. The FPGA does not have the FLASH_TEST_MODE0 pad, and the ability to give itself a POR allows us to add firmware to fix up inconsistent starting points for the bitstream vs. a real provisioned chip.

This change is motivated by a need to fix up the flash info pages to reflect the expected chip state when in PROD. It is possible to splice VMEM images as is currently done for OTP and ROM, but calls to updatemem appear to consume an undesirable amount of time.

### DIFF
--- a/hw/top_earlgrey/rtl/autogen/chip_earlgrey_cw310.sv
+++ b/hw/top_earlgrey/rtl/autogen/chip_earlgrey_cw310.sv
@@ -876,6 +876,7 @@ module chip_earlgrey_cw310 #(
   assign pad2ast = '0;
 
   logic clk_main, clk_usb_48mhz, clk_aon, rst_n, srst_n;
+  logic fpga_eos;
   clkgen_xil7series # (
     .AddClkBuf(0)
   ) clkgen (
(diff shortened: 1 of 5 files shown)


a-will:
This just allows satisfying OTTF's "POR to start" guarantee, as many tests check for the POR reset reason to know they are at the beginning of the test.

The info pages would be written by ROM_EXT firmware specific to this FPGA execution environment, and then that firmware would trigger this internal POR.

matutem:
I thought splicing flash images was the plan-of-record. Is this draft PR still needed?

a-will:
> I thought splicing flash images was the plan-of-record. Is this draft PR still needed?

It's a draft PR: "Need" doesn't really enter.

However, I'll drop the review requests now, since it does look like we're going towards splicing after all. I'll keep this draft PR around in case we change our minds again. ;)

a-will:
This is not planned right now.

PR #25404, opened 2024-11-26
[sival,bazel] Add SiVal ROM EXT exec env without hyper310 for use in PMOD tests

AlexJones0:
Adds a SiVal ROM EXT environment local to the PMOD tests that overrides the `hyper310` bitstream and interface to use the `cw310` versions, so that the relevant pins required for PMOD are accessible during the test, and the test can be run in a ROM EXT environment with the local SiVal changes (e.g. OTP changes). This should work with the existing nightly runner as it uses the same `fpga_cw310_sival_rom_ext` tag, despite being a private exec env.

### DIFF
--- a/sw/device/tests/pmod/BUILD
+++ b/sw/device/tests/pmod/BUILD
@@ -12,19 +12,55 @@ load(
     "silicon_params",
     "verilator_params",
 )
+load("//rules:splice.bzl", "bitstream_splice")
+load(
+    "//rules/opentitan:defs.bzl",
+    "DEFAULT_TEST_FAILURE_MSG",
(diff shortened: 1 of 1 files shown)


jwnrt (commented):
LGTM, do we need to test this in CI or is it purely for sival nightlies?

AlexJones0:
> do we need to test this in CI or is it purely for sival nightlies?

I'm not sure, as this change was just to address the failing SiVal nightlies - perhaps @engdoreis would know?

engdoreis:
> do we need to test this in CI or is it purely for sival nightlies?

Only for nightlies.

engdoreis (approved):

PR #25406, opened 2024-11-26
[provisioning]: allow the owner to specify perso success condition

vbendeb:
Certain SKUs require personalization success validation which extends beyond confirming that the ROM_EXT in slot B has started.

Let's allow the user to specify a string which when issued on the chip console indicates perso success.

This will require longer timeouts after post perso reboot.

Tested the following:
  - when onwer string is not provided . success and failure due UDS cert problems are reported as before
  - when owner string is provided . failure due UDS cert problems is still reported . success when fw string is detected is reported . in case fw string is not detected timeout results in an error

### DIFF
--- a/sw/host/provisioning/ft_lib/src/lib.rs
+++ b/sw/host/provisioning/ft_lib/src/lib.rs
@@ -528,11 +528,12 @@ pub fn run_ft_personalize(
     Ok(())
 }
 
-pub fn check_rom_ext_boot_up(
+pub fn check_slot_b_boot_up(
     transport: &TransportWrapper,
     init: &InitializeTest,
     timeout: Duration,
     response: &mut PersonalizeResponse,
+    owner_fw_success_string: Option<String>,
 ) -> Result<()> {
     transport.reset_target(init.bootstrap.options.reset_delay, true)?;
     let uart_console = transport.uart("console")?;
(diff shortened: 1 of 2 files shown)


vbendeb:
a suggested way to provision SKU specific perso success condition. Once agreed upon will modify SKUs to provide the expected anchor strings.

sw/host/provisioning/ft_lib/src/lib.rs
```diff
@@ -548,26 +549,39 @@ pub fn check_rom_ext_boot_up(
...
+        });
 
-    let t0 = Instant::now();
-    let result = UartConsole::wait_for(
-        &*uart_console,
-        r".*UDS certificate not valid.",
-        UDS_CERT_INVALID_TIMEOUT,
-    );
     response.stats.log_elapsed_time("rom_ext-done", t0);
+    let fw_failure_anchor = r"Invalid UDS certificate detected!";
```
timothytrippel, line 566: nit (it is a ROM_EXT failure message):
```suggestion
    let rom_ext_failure_msg = r"Invalid UDS certificate detected!";
```

timothytrippel (approved):
Thanks @vbendeb !

sw/host/provisioning/ft_lib/src/lib.rs
vbendeb: done

PR #25409, opened 2024-11-27
[dv/dpi] tcpserver single connection and retry select on EINTR

Razer6:
This is pull over from https://github.com/lowRISC/opentitan/pull/23889 to `master`:

Two small patches to the tcp_server code, used by other transactors such as jtagdpi and dmidpi. These improve the usability of the transactors in a shared, remote environment such as an emulator.

Without this change, an interrupted select() call is treated as a connection error and the client is dropped. This makes connections very unstable.

### DIFF
--- a/hw/dv/dpi/common/tcp_server/tcp_server.c
+++ b/hw/dv/dpi/common/tcp_server/tcp_server.c
@@ -183,7 +183,6 @@ static int client_tryaccept(struct tcp_server_ctx *ctx) {
   int rv;
 
   assert(ctx->sfd > 0);
-  assert(ctx->cfd == 0);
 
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
Looks sensible (and unchanged from when I reviewed it on the `integrated_dev` branch)

kwalker27:
> Looks sensible (and unchanged from when I reviewed it on the `integrated_dev` branch)

I think I took your feedback and updated the patch, was waiting for you to take a second look.

Razer6:
> > Looks sensible (and unchanged from when I reviewed it on the `integrated_dev` branch)
> 
> I think I took your feedback and updated the patch, was waiting for you to take a second look.

I think he meant that I didn't touch your files, which you updated according to the review.

PR #25420, opened 2024-11-27
[sival] Fix rv_dm_jtag bazel entries

jwnrt:
This Bazel test is parametrised by its LC state, not the RV_DM states like other tests (e.g. `rv_dm_mem_`).


AlexJones0 (approved):
Oops - thanks for spotting this, this looks correct.

luismarques (approved):

PR #25421, opened 2024-11-27
[chip,dv] Correct a testplan item

elliotb-lowrisc:
Similar to commit 25b8f54

The testpoint was given a "tests" field that was supposed to be the "bazel" field.


engdoreis:
Thanks for your PR @elliotb-lowrisc. 
Sorry for missing it, but I fixed this same issue on this [PR](https://github.com/lowRISC/opentitan/pull/25428)

elliotb-lowrisc:
Ah well, at least it got done

PR #25424, opened 2024-11-27
[Darjeeling,lint] More lint fixes

Razer6:
More lint fixes for Darjeeling

### DIFF
--- a/hw/ip_templates/pinmux/rtl/pinmux.sv.tpl
+++ b/hw/ip_templates/pinmux/rtl/pinmux.sv.tpl
@@ -492,6 +492,11 @@ module pinmux
     assign mio_to_periph_o[k] = mio_mux[reg2hw.mio_periph_insel[k].q];
   end
 
+  if (AlignedMuxSize > 2**$clog2(NMioPads + 2)) begin : gen_align_unused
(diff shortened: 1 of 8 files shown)


hw/ip_templates/pinmux/rtl/pinmux.sv.tpl
```diff
@@ -492,6 +492,11 @@ module pinmux
     assign mio_to_periph_o[k] = mio_mux[reg2hw.mio_periph_insel[k].q];
   end
 
+  if (AlignedMuxSize > 2**$clog2(NMioPads + 2)) begin : gen_align_unused
+    logic unused_mio_mux_signals;
+    assign unused_mio_mux_signals = ^{mio_mux[AlignedMuxSize-1:2**$clog2(NMioPads + 2)]};
```
andreaskurth, line 497: Only the lowest `NMioPads + 2` bit of `mio_mux` are used, right? So shouldn't this be
```suggestion
    assign unused_mio_mux_signals = ^{mio_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
```
?

The same question also applies to `mio_wkup_mux` and `dio_wkup_mux` below

PR #25428, opened 2024-11-27
[SiVal, spi] Fix bazel target in testplan

engdoreis:


jwnrt:
Let's skip CI, testplan changes have no effect.

PR #25430, opened 2024-11-27
[alert,dv] Tidyups in the alert receiver driver

rswarbrick:
When debugging my error and fixing it in #25272, I tidied the code up a bit, using out-of-block definitions and documenting how things worked a little more carefully. This builds on that PR (which contains the first commit) and shouldn't change behaviour, but should make the code a bit easier to reason about.


matutem (approved):
LGTM

PR #25431, opened 2024-11-27
[alert,dv] Use priorities in alert agent arbitration

rswarbrick:
I was trying to be clever when I wrote a771270fab3, and set the ping sequence to run with a lower priority than the alert response sequence. But this doesn't work with the default fifo arbitration...


marnovandermaas (approved):
I think this makes sense and it only a small change to DV code.

PR #25432, opened 2024-11-27
[topgen] Only create pwrmgr/pinmux/alert_handler if there is an instance

Razer6:
Not all tops may have a pwrmgr, pinmux, alert_handler or clkmgr. Only create those IPs and the related connections if the top really specifies those IPs.


util/topgen.py
```diff
@@ -1152,14 +1154,17 @@ def render_template(template_path: str, rendered_path: Path,
...
-        for target in topcfg["targets"]:
-            target_name = target["name"]
-            render_template(top_template_path / "chiplevel.sv.tpl",
-                            out_path /
-                            f"rtl/autogen/chip_{topname}_{target_name}.sv",
-                            gencmd=gencmd,
-                            target=target)
+        # Render chip-level template only if top specifies one
+        chiplevel_tpl = top_template_path / 'chiplevel.sv.tpl'
+        if chiplevel_tpl.is_file():
```
matutem, line 1159: This seems to be inverted: I would expect topcfg["targets"] to be empty or not-such-entry if there is no chip-level.

Razer6: I dropped the commit. Let's leave `topcfg["targets"]` empty if you don't need an implementation target.

andreaskurth (approved):
LGTM in the sense that it doesn't negatively impact currently generated tops. Whether the result is correct for tops that make use of the added functionality I cannot say, though

Razer6:
I used it downstream and it works. The used top-level template though needs some tweaks, but I leave that as a downstream issue. One thing I need to change was the order of the alert and plic mappings in `toplevel.c.tpl`. The alert mappings where at the end of that file and because of gating it with an if, it caused mako to render two newlines. One for the mapping and one for the endif. This then caused a linting issue. The easy fix was to put the un-gated plic mapping at the end.

PR #25436, opened 2024-11-27
[dv/dpi] jtagdpi lookahead for 'R' on TCK posedge

kwalker27:
Clients performing a shift operation with data readback will be sending a sequence of commands that toggle TCK, providing the data to shift in, interleaved with read commands to sample the data shifted out. In the current implementation where a single command is processed on each jtagdpi tick, this leads to an uneven TCK duty cycle (i.e. 3 jtagdpi ticks per TCK cycle) whenever data read commands are present.

The JTAG spec dictates that TDO values change on the falling edge of TCK and so should be sampled around the rising edge. The command processing is changed to support a "lookahead" to attempt to pull a read command back to the same jtagdpi tick where a TCK rising edge command is handled. With this lookahead, a client that does (TCK falling edge, TCK rising edge, read TDO) will complete its TCK cycle in just 2 jtagdpi ticks.

In theory a client that generates a sequence like (TCK falling edge, read TDO, TCK rising edge) could also lookahead from the read command for a TCK rising edge. However, such a client is likely to read the data before sending the rising edge command - so a successful lookahead would incur an additional network round-trip to the client. On the other hand, the client-side change to take advantage of this optimization is trivial.

### DIFF
--- a/hw/dv/dpi/jtagdpi/jtagdpi.c
+++ b/hw/dv/dpi/jtagdpi/jtagdpi.c
@@ -22,7 +22,36 @@ struct jtagdpi_ctx {
   uint8_t tdo;
   uint8_t trst_n;
   uint8_t srst_n;
+  // Lookahead buffer - non-zero if valid
+  char cmd;
 };
+
+static bool lookahead(struct jtagdpi_ctx *ctx) {
+  // Look at the next command if available. Return true if it's an
+  // 'R', otherwise buffer it to return via get_cmd().
+  char cmd;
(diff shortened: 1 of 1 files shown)


Razer6 (approved):

kwalker27:
Hi @rswarbrick do you mind taking a look at this one too?

rswarbrick (approved):
This took me a bit of squinting to follow, but I think I understand it now. Thank you for the careful comments to explain what's going on!

PR #25437, opened 2024-11-27
[manuf] Extract device_id from CP stage.

moidx:
Extract device_id from CP to inject it into the FT stage. The individualize firmware is expected to  handle this at the device level as well.

### DIFF
--- a/sw/host/provisioning/orchestrator/src/device_id.py
+++ b/sw/host/provisioning/orchestrator/src/device_id.py
@@ -7,7 +7,7 @@ import struct
 from dataclasses import dataclass
 
 from sku_config import SkuConfig
-from util import bytes_to_int, format_hex
+import util
 
 _RESERVED_WORD = 0
 
@@ -61,6 +61,21 @@ class DeviceIdentificationNumber:
         din |= self.year
         return din
 
+    @staticmethod
+    def from_int(din: int) -> "DeviceIdentificationNumber":
+        year = din & 0xF
+        week = (din >> 4) & 0xFF
+        lot = (din >> 12) & 0xFFF
+        wafer = (din >> 24) & 0xFF
+        wafer_x_coord = (din >> 32) & 0xFFF
+        wafer_y_coord = (din >> 44) & 0xFFF
+        return DeviceIdentificationNumber(year=year,
+                                          week=week,
+                                          lot=lot,
+                                          wafer=wafer,
+                                          wafer_x_coord=wafer_x_coord,
+                                          wafer_y_coord=wafer_y_coord)
+
 
 class DeviceId():
     """An OpenTitan device ID.
@@ -89,7 +104,7 @@ class DeviceId():
         # - 16 bits Product ID
         self.si_creator_id = sku_config.si_creator_id
         self.product_id = sku_config.product_id
-        self._hw_origin = bytes_to_int(
+        self._hw_origin = util.bytes_to_int(
             struct.pack("<HH", self.si_creator_id, self.product_id))
 
         # Build Device Identification Number with:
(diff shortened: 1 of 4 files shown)


sw/host/provisioning/orchestrator/src/ot_dut.py
```diff
@@ -115,11 +114,37 @@ def run_cp(self) -> None:
         # Run provisioning flow and collect logs.
         res = run(cmd, f"{self.log_dir}/cp_out.log.txt",
                   f"{self.log_dir}/cp_err.log.txt")
```
moidx, line 116: We can probably stage the run in a tempdir, and then move the files to expected {device_id} folder after execution. wdyt?

sw/host/provisioning/orchestrator/src/ot_dut.py
```diff
@@ -115,11 +114,37 @@ def run_cp(self) -> None:
...
+            logging.error("CHIP_PROBE_DATA not found.")
+            confirm()
+
+        if "cp_device_id" not in chip_probe_data:
+            logging.error("cp_device_id found in CHIP_PROBE_DATA.")
+            confirm()
+
+        logging.info(
+            f"Updating device ID to: {chip_probe_data['cp_device_id']}")
+        self.device_id.from_hexstr(chip_probe_data["cp_device_id"])
```
timothytrippel, line 145: Note: this will just create the first 128-bits of the device ID (the "_base_uid" [portion](https://cs.opensource.google/opentitan/opentitan/+/earlgrey_1.0.0:sw/host/provisioning/orchestrator/src/device_id.py;l=121)); this needs to be combined with the SKU specific [part](https://cs.opensource.google/opentitan/opentitan/+/earlgrey_1.0.0:sw/host/provisioning/orchestrator/src/device_id.py;l=111) to complete the full Device ID that gets passed to the FT stage.

sw/host/provisioning/orchestrator/src/ot_dut.py
timothytrippel: yep that would probably be cleaner and simplify retries.

sw/host/provisioning/orchestrator/src/device_id.py
```diff
@@ -120,9 +121,14 @@ def __init__(self, sku_config: SkuConfig, din: DeviceIdentificationNumber):
         # Build full device ID.
         self.device_id = (self._sku_specific << 128) | self._base_uid
 
+    def from_hexstr(self, hexstr: str):
+        """Creates a DeviceId object from a hex string."""
+        # TODO: Update other device ID fields based on new device ID.
+        self.device_id = util.parse_hexstring_to_int(hexstr)
```
timothytrippel, line 127: Might be nice to call self.validate() here to ensure the hexstr is valid. Wdyt?

timothytrippel (approved):

sw/host/provisioning/orchestrator/src/ot_dut.py
moidx: Done.

sw/host/provisioning/orchestrator/src/device_id.py
moidx: Done. This should be taken care of by the DeviceId post_init.

sw/host/provisioning/orchestrator/src/ot_dut.py
moidx: Done. PTAL.

timothytrippel (approved):
LGTM. Thanks!

moidx:
Merge changes in #25469.

PR #25438, opened 2024-11-27
[sival] Enable `fpga_cw310_sival_rom_ext` exec env on cryptotests

milesdai:
This PR adds the `fpga_cw310_sival_rom_ext` execution environment to the cryptotests.

This will cause the nightly CI runs to pick up these tests; however, because some of these tests are very slow (i.e. > 30 min), this PR also adds a `skip_in_nightly_ci` bazel tag that allows excluding certain tests.

Currently, only three of the KAT tests repeatably run in under 5 minutes (HMAC-SHA256 is a bit close), so the rest of the tests are being excluded at the moment.

```console
//sw/device/tests/crypto/cryptotest:cshake_kat_fpga_cw310_sival_rom_ext (cached) PASSED in 21.6s
//sw/device/tests/crypto/cryptotest:kmac_kat_fpga_cw310_sival_rom_ext (cached) PASSED in 146.3s
//sw/device/tests/crypto/cryptotest:ecdh_kat_fpga_cw310_sival_rom_ext    PASSED in 248.5s
//sw/device/tests/crypto/cryptotest:hmac_sha256_kat_fpga_cw310_sival_rom_ext PASSED in 237.2s
```

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -64,7 +64,7 @@ jobs:
                                                     | grep -v examples \
                                                     | grep -v penetrationtests \
                                                     > "$bazel_tests"
-          ./bazelisk.sh test --build_tests_only --target_pattern_file="$bazel_tests"
+          ./bazelisk.sh test --build_tests_only --test_tag_filters=-skip_in_nightly_ci --target_pattern_file="$bazel_tests"
 
       - name: Run tests after ROM_EXT boot stage
         if: success() || failure()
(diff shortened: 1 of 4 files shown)


milesdai:
One thing I noticed is that the cryptotests seem to run about 6x slower in the `fpga_cw310_sival_rom_ext` exec env as compared to the previous `fpga_cw310_test_rom` environment. The most notable difference seems to be the change to the hyper310 interface. Since cryptotest leans very heavily on UART communication, is it possible that something in the UART handling stack is slower on the hyperdebug than on the CW310?

jwnrt:
I'm not aware of the HyperDebug UART being slower than the CW310/SAM3X one. Could the change from test ROM to real ROM have had any effect?

engdoreis (approved):

moidx (approved):
Thanks! Nice refactoring.

PR #25439, opened 2024-11-27
[dv/dpi] jtagdpi supports 2 new plusargs

kwalker27:
See the commit log messages, but the functionality is (1) override the `LISTEN_PORT` parameter and (2) assert `srst_n` from start of sim.


kwalker27:
Looks like `$test$plusargs` isn't allowed, so I'll fix that. It also looks like all-lowercase plusargs are the convention, so I'll move to that as well.

rswarbrick (approved):
This looks good (and I appreciate the clear coding style: it makes the review really easy!)

PR #25443, opened 2024-11-28
[dv,shadow_reg_errors] Test the last shadowed CSR

### LINKED ISSUE #25442: [sec_cm,shadow_reg_errors] Test the last shadowed CSR
### Description

The function run_shadow_reg_errors at https://github.com/lowRISC/opentitan/blob/master/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__shadow_reg_errors.svh#L5 has a bug in the loop over indices (https://github.com/lowRISC/opentitan/blob/master/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__shadow_reg_errors.svh#L65). It needs to iterate until i <= max_idx, since in the normal case max_idx is shadowed_csrs.size() - 1.

This was caught as a coverage hole in flash_ctrl, which has a single shadowed CSR , so no test was performed. Any block with more than one shadowed CSR would have tested them all since the test is run with 20 reseeds.

matutem:
Fixes #25442

### DIFF
--- a/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__shadow_reg_errors.svh
+++ b/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq__shadow_reg_errors.svh
@@ -62,7 +62,7 @@ virtual task run_shadow_reg_errors(int num_times, bit en_csr_rw_seq = 0);
 
     shadowed_csrs.shuffle();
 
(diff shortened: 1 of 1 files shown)


andreaskurth (approved):
Thx for finding and fixing this, @matutem! LGTM

vogelpi (approved):
Thanks for finding and fixing this @matutem !

PR #25447, opened 2024-11-28
[SiVal, spi_host] Fix macronix test tag

engdoreis:
This will make test run on the SiVal nightly job.

### DIFF
--- a/sw/device/tests/pmod/BUILD
+++ b/sw/device/tests/pmod/BUILD
@@ -41,7 +41,6 @@ fpga_cw310(
         "assemble": "{rom_ext}@0 {firmware}@0x10000",
     },
     rom_mmi = "//hw/bitstream:rom_mmi",
-    tags = ["cw310_sival_rom_ext"],
     visibility = ["//visibility:private"],
 )
 
@@ -65,6 +64,7 @@ opentitan_test(
     ),
     fpga = fpga_params(
         tags = [
+            "cw310_sival_rom_ext",
             "manual",
             "pmod",
         ],  # Requires the PMOD::BoB.


luismarques (approved):

sw/device/tests/pmod/BUILD
```diff
@@ -65,6 +64,7 @@ opentitan_test(
     ),
     fpga = fpga_params(
         tags = [
+            "cw310_sival_rom_ext",
```
AlexJones0, line 67: If I'm not misunderstanding, would doing this not make the `cw310_sival_rom_ext` tag appear on all tests that use `fpga`, including e.g. `fpga_cw310_test_rom`? If so, would it not be better to make a separate `fpga_params` entry for the `fpga_cw310_sival_rom_ext_no_hyper` environment?

AlexJones0, line 67: Would doing this not make the `cw310_sival_rom_ext` tag appear on all tests that use `fpga`, including e.g. `fpga_cw310_test_rom`? If so, would it not be better to make a separate `fpga_params` entry for the `fpga_cw310_sival_rom_ext_no_hyper` environment?

If I apply this change and run `./bazelisk.sh test --test_tag_filters=cw310_sival_rom_ext $(./bazelisk.sh query 'attr("tags", "[\[ ]pmod[,\]]", tests(//...))')` then I get the following tests:
```
//sw/device/tests/pmod:i2c_host_eeprom_test_fpga_cw310_sival_rom_ext
//sw/device/tests/pmod:i2c_host_fram_test_fpga_cw310_sival_rom_ext
//sw/device/tests/pmod:spi_host_macronix1Gb_flash_test_fpga_cw310_rom_with_fake_keys
//sw/device/tests/pmod:spi_host_macronix1Gb_flash_test_fpga_cw310_test_rom
//sw/device/tests/pmod:spi_host_macronix1Gb_flash_test_fpga_cw310_sival_rom_ext_no_hyper
```
which I assume is not intended?

I've looked into why the original `tags` on the environment were not working - turns out that this attribute does not apply the actual tags, and that is instead being done by an aptly named `_hacky_tags` function in [`rules/opentitan/defs.bzl`](https://github.com/lowRISC/opentitan/blob/4708d7bfe69ba8b8121ba4b3076e72b07113dafc/rules/opentitan/defs.bzl#L151-L176). One (hacky) work-around  that illustrates this is:
```patch
diff --git a/rules/opentitan/defs.bzl b/rules/opentitan/defs.bzl
index d96a346a28..c17bfc2c23 100644
--- a/rules/opentitan/defs.bzl
+++ b/rules/opentitan/defs.bzl
@@ -168,6 +168,8 @@ def _hacky_tags(env):
             tags.append("hyper310")
         elif subtag.startswith("cw340"):
             tags.append("cw340")
+        if subtag.startswith("cw310_sival_rom_ext_no_hyper"):
+            tags.append("cw310_sival_rom_ext")
     if suffix.startswith("silicon"):
         # We add the entire suffix for silicon exec environments to be able
         # to filter tests by them. "silicon_creator" and
```
Though an elegant solution would be nicer.

AlexJones0 (changes requested):

sw/device/tests/pmod/BUILD
AlexJones0: To avoid making `_hacky_tags` any more involved than it needs to be, I recommend probably taking the following approach to locally apply a fix to just this one test/environment combination:
```patch
diff --git a/sw/device/tests/pmod/BUILD b/sw/device/tests/pmod/BUILD
index 277d2ac5af..d63774aea8 100644
--- a/sw/device/tests/pmod/BUILD
+++ b/sw/device/tests/pmod/BUILD
@@ -41,7 +41,6 @@ fpga_cw310(
         "assemble": "{rom_ext}@0 {firmware}@0x10000",
     },
     rom_mmi = "//hw/bitstream:rom_mmi",
-    tags = ["cw310_sival_rom_ext"],
     visibility = ["//visibility:private"],
 )
 
@@ -60,9 +59,16 @@ opentitan_test(
             "//hw/top_earlgrey:fpga_cw310_test_rom": None,
             # Custom execution environment must be used for SiVal ROM EXT
             # to ensure that all the PMOD pins are available.
-            ":fpga_cw310_sival_rom_ext_no_hyper": None,
+            ":fpga_cw310_sival_rom_ext_no_hyper": "sival_rom_ext_no_hyper",
         },
     ),
+    sival_rom_ext_no_hyper = fpga_params(
+        tags = [
+            "cw310_sival_rom_ext",
+            "manual",
+            "pmod",
+        ],  # Requires the PMOD::BoB.
+    ),
     fpga = fpga_params(
         tags = [
             "manual",
```

AlexJones0:
Closing as superceded by #25457.

PR #25451, opened 2024-11-28
[dv] Add GUI debug mode for Riviera

dariuszst:
Hi,
I added a missing riviera_gui_debug mode required after f88269d2658dcf0d9c75b7a507d0bda160751c47 update.
This is an initial version allowing running tests with dvsim.py script.

### DIFF
--- a/hw/dv/tools/dvsim/riviera.hjson
+++ b/hw/dv/tools/dvsim/riviera.hjson
@@ -70,6 +70,13 @@
       run_opts: []
     }
     {
+      name: riviera_gui_debug
+      // TODO: Add options to setup Riviera gui debug environment and features (as in xcelium.hjson)
+      is_sim_mode: 1
+      build_opts: []
+      run_opts: []
+    }
+    {
       name: riviera_waves
       is_sim_mode: 1
     }


rswarbrick (commented):
This looks reasonable, but needs a bit of tidying up.
- For the TODO note, I think we should be able to explain what's going on without needing to call out to a merged PR.
- I think the two commits probably need squashing together.

dariuszst:
Thanks @rswarbrick  - I've updated my fork repository (removed this unnecessary commit) and I hope the change is now referenced correctly. I changed also the TODO note - is it OK now?

hw/dv/tools/dvsim/riviera.hjson
```diff
@@ -69,6 +69,13 @@
       build_opts: []
       run_opts: []
     }
+    {
+      name: riviera_gui_debug      
+      // TODO need to add options to setup Riviera gui debug environment and features (similary like in PR #24156 for Xelium)
```
rswarbrick, line 74: I think this would be better if we were pointing "internally". Can't you just say something like "Similar to how we do it for Xcelium: see xcelium.hjson." ?

dariuszst: Yes. It can be described in this way.

rswarbrick: Probably just merge the changes together.

rswarbrick:
I've just squashed things together and dropped the trailing whitespace. Thanks for looking at this.

rswarbrick (approved):

PR #25457, opened 2024-11-29
[sival,spi_host] Fix tag on Macronix flash SiVal test

AlexJones0:
This will properly tie the tag to the test (and not just to the execution environment, which does nothing, due to the implementation of `_hacky_tags` used in `opentitan_test`), so that the test will run in the nightly SiVal job. See discussion [here](https://github.com/lowRISC/opentitan/pull/25447#discussion_r1862072286) for more explanation.

Supercedes #25447 as per request by @engdoreis.


engdoreis (approved):

AlexJones0:
Force push is just a buildifier lint to pass CI.

AlexJones0:
CI failure is due to current CI outage issues, this change should only affect `//sw/device/tests/pmod:spi_host_macronix1Gb_flash_test` and so is unrelated.

PR #25459, opened 2024-11-29
[chore] Bump Python requirement to 3.12

jwnrt:
We'd like to decouple our Python version from the system Python.

The motivation is:

* Allowing us to update our dependencies.
* Allowing us to use projects that only support the latest Python (e.g. QEMU scripts).
* Allowing us to test against one fixed Python version in CI.

This shouldn't cause any issues within Bazel, however hardware engineers using Python scripts directly (e.g. through fusesoc) may need to start using virtual environments.


andreaskurth:
> We'd like to decouple our Python version from the system Python.
> (...)
> however hardware engineers using Python scripts directly (e.g. through fusesoc) may need to start using virtual environments.

I'm not generally opposing the changes proposed in this PR, just want to raise a concern that such changes should be properly discussed among teams and partners and that it's probably worth preparing a transition guide for integrators / downstream users.

If I understand correctly, there are two main changes:
- using a project-specific / virtual environment installation of Python instead of the system Python;
- updating from Python 3.8 to 3.12.

Is that accurate?

Going from Python 3.8 to 3.12 is basically a jump of four years (in terms of release date) which includes quite a few deprecations and subsequent removals. Not all OpenTitan partners may be able to make that change within a few hours, as Python code on their end may have to be changed.

As for using a project-specific / virtual environment installation of Python instead of the system Python, can this be bootstrapped on any Linux distribution (also consider older ones like CentOS 7) by a user in their local directories without superuser privileges? Does it also work for sandboxed machines (i.e., no internet access)?

This may be worth an RFC -- WDYT?

jwnrt:
Yes, I think you're right that this is too big of a jump. I have found a way to avoid having to do this update, so I will close this for now.

PR #25460, opened 2024-11-29
[rv_dm,sival] Fix `rv_dm_mem_access_delayed_enable`

AlexJones0:
The `rv_dm_mem_access` test was not working in the `dev_rv_dm_delayed_enable` test configuration. This is because the test was not written to handle the additional logic required for this configuration.

To fix this, before the host side attempts to connect via JTAG, it waits for a message ("`DEBUG_MODE_ENABLED`") from the device, which will configure the RV DM to enable the late debug feature (debug mode). This should only be available using the OTP image with the late debug feature enabled, in the `delayed_enable` test configuration.

This has been tested to pass consistently on a CW310 by running the following command:
`./bazelisk.sh test -t- --test_output=streamed --build_tests_only --runs_per_test=5 --test_tag_filters=cw310 $(./bazelisk.sh query "tests(//sw/device/tests:rv_dm_mem_access)")`

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -5072,7 +5072,7 @@ test_suite(
 [
     opentitan_test(
         name = "rv_dm_mem_access_{}".format(test_cfg["name"]),
-        srcs = ["example_test_from_flash.c"],
+        srcs = ["rv_dm_delayed_enable.c"],
         exec_env = {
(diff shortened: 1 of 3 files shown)


engdoreis (approved):

AlexJones0:
CI failure is due to current CI outage issues, this change should only affect the `//sw/device/tests/pmod:rv_dm_mem_access_*` tests.

PR #25461, opened 2024-11-29
[prim_fifo_sync,rtl] Specialize for Depth == 1

### LINKED ISSUE #25419: [rtl,prim_fifo_sync] Avoid counters when Depth == 1
### Description

The sync fifos generate logic for counters when Depth is 1, even though they are not used or needed. This creates coverage holes, and it is cleaner to simply remove the counters with an SV if-generate.

rswarbrick:
This case is pretty simple to reason about, and avoids needing proper read and write addresses (because there is only one element).

This has seen some FPV testing (`util/dvsim/dvsim.py hw/top_earlgrey/formal/top_earlgrey_fpv_prim_cfgs.hjson --select-cfg=prim_fifo_sync_fpv`) and passes happily. This was inspired by #25419 (or, more precisely, I was!) so I thought it made sense to implement.

Closes #25419.


rswarbrick:
Force-push wires up the `full_o` output (which wasn't tested by the FPV: a fact I only discovered from trying to fix some warnings from Jasper...)

rswarbrick:
Force-push wires up the other output that I'd missed...

Thinking about `err_o` makes me realise that this might not be hardened enough. We now have 1 bit of storage (`full_q`) which is not hardened, where it *was* hardened with the existing implementation. Adding the hardening wouldn't be hard, but I'd be interested to hear from someone who knows more about HW security. @andreaskurth, could you give me some pointers? :-)

hw/ip/prim/rtl/prim_fifo_sync.sv
```diff
@@ -55,6 +55,49 @@ module prim_fifo_sync #(
     // No error
     assign err_o = 1'b 0;
 
+  // FIFO has space for a single element (and doesn't need proper counters)
+  end else if (Depth == 1) begin : gen_singleton_fifo
```
matutem, line 59: It this is handled here, perhaps the code around line 150 should be wiped.

matutem (approved):
I left a couple suggestions, otherwise LGTM. Thanks for taking this one :)

hw/ip/prim/rtl/prim_fifo_sync.sv
rswarbrick: Good point. I've tidied that up now.

rswarbrick:
Thanks for the review, @matutem. I've addressed the comments (I hope) and force-pushed. I think there's still a question about security hardening and I'm hoping someone who knows more about security than me (@andreaskurth? @vogelpi?) would be able to weigh in.

vogelpi (commented):
I am not sure I understand the motivation of this PR. It doesn't mean I cannot support it but I would first like to understand before spending time on reviewing this. As noted in the discussion, this can have security impact, so a careful review is needed here.

@rswarbrick can you please clarify? My understanding is that with the current design and a depth of 0, we instantiate the `prim_fifo_sync_cnt` modules even though they are not used - there is a single module storage entry anyway and we always write that one and read from it. The `prim_fifo_sync_cnt` modules mean some waste of area due to the prim_flops inside and coverage cannot be closed for this as is. Is this correct?

rswarbrick:
Yes, I think that is correct, which is the reason for @matutem's original issue (I think).

vogelpi (commented):
The unhardened implementation looks good to me but I think adding a second flop (using `prim_flop`) in the case `Secure` is set to 1 would make sense.

Can you please also check if there are any instances in the design with `Depth == 1` and `Secure == 1`? Because for blocks using such instantiations we have very likely some FI tests that will break with this PR. Because the faults are typically injected into either the read or the write pointers which are no longer there with this PR.

rswarbrick:
I've just worked through the tree a bit to search for examples of this. And I've found at least some! The three fifos in `tlul_adapter_sram` are set to secure if the adapter is instantiated with `SecFifoPtr=1`. This happens in `otbn`, `rom_ctrl`, `sram_ctrl` and `flash_ctrl`.

rswarbrick:
I've just pushed up a version with the duplicated 1-bit counter that I think we'd need. Looking closely, I don't think this will actually break any testing. But that's because we do the tests by injecting things into instances of `prim_count` (which I would be removing).

Clearly, it *would* cause the testing to miss security holes! So I'm going to poke around now and see whether I can replicate the testing sensibly in this case.

rswarbrick:
Well, *that* was harder than I expected! I've just force pushed with a much larger commit (sorry) that I'm hoping will fix lots of the build failures with the previous version. Frustratingly, I can't really see an easy way to split it up.

hw/ip/prim/rtl/prim_fifo_sync.sv
```diff
@@ -55,6 +55,81 @@ module prim_fifo_sync #(
...
+    end else begin : gen_secure
+      logic inv_full;
+
+      prim_flop #(
+        .Width(1),
+        .ResetValue(1'b1)
+      ) u_inv_full (
+        .clk_i,
+        .rst_ni,
+        .d_i (^full_d),
```
vogelpi, line 115: Nit: should this be `~` instead? What is the definition of XORing all bits of a single-bit signal?

vogelpi (commented):
Thanks @rswarbrick for going the extra miles with the automated countermeasure tests, this helps keeping the nightlies healthy and avoids someone else debugging afterwards.

I think for the entropy complex (EDN, CSRNG, ENTROPY_SRC) we do have non-automated tests injecting errors into all FIFOs, see e.g. `csrng_err_vseq.sv`. I am not sure but I believe these sequences will now fail as the paths into the depth == 1 FIFOs no longer exists. This is only detected at runtime. Can you please run full regressions for these three blocks to be sure? I think a 10% regression should be sufficient to hit these cases.

vogelpi:
Ah and for the new DV interfaces and assertions, I would appreciate if @matutem could take another look please.

hw/ip/prim/rtl/prim_fifo_assert.sv
```diff
@@ -0,0 +1,47 @@
...
+
+// Use PRIM_COUNT_ERROR_TRIGGER_ALERT appropriately to check that the three prim_counts generated by
+// a prim_fifo_sync with depth at least two do indeed generate an alert if they detect an error.
+//
+// - NAME_ is used as the root of the names of the generated assertions.
+// - HIER_ is a hierarchical path to the prim_fifo_sync in question.
+// - ALERT_ is the name of the alert that should be generated.
+// - GATE_ is a signal that, if true, will cause an error to be ignored.
+// - MAX_CYCLES_ is the number of cycles allowed until the alert must be generated.
+`define ASSERT_PRIM_FIFO_SYNC_ERROR_TRIGGERS_ALERT(NAME_, HIER_, ALERT_, GATE_ = 0, MAX_CYCLES_ = `_SEC_CM_ALERT_MAX_CYC) \
```
matutem, line 18: Could you split this into multiple lines, and place arguments in line 19 in a separate line? Meaning, could you avoid very long lines via new-lines? This will look better and avoid the lint waiver.

matutem (approved):
Thanks Rupert. The extra flop seems to address the hardening concern, and I agree with you about creating a specialized proxy for depth==1, secure==1 seems right. LGTM

hw/ip/prim/rtl/prim_fifo_assert.sv
rswarbrick: Annoyingly, I don't think you can do that in the obvious way (sticking a backslash, newline) into the middle of the argument list. I think that SystemVerilog doesn't actually allow that. I looked at the 1800 spec and I think the interesting bit is section 22.5.1, but it isn't super clear whether the note about backslash, newline applies to the list of arguments. But tools generally choke on it :-)

I wondered about setting the default value for `MAX_CYCLES` to zero (which would be a silly value) and switching to the default value further down the chain of calls. The top-level macro becomes something like
```systemverilog
`define ASSERT_PRIM_FIFO_SYNC_ERROR_TRIGGERS_ALERT(NAME_, HIER_, ALERT_, GATE_ = 0, MAX_CYC_ = 0) \
```
which is now short enough (woo!)

This boils down to the third argument in following macro though:
```systemverilog
`define ASSERT_ERROR_TRIGGER_ERR(NAME_, HIER_, ERR_, GATE_, MAX_CYCLES_, ERR_NAME_, CLK_, RST_) \
  `ASSERT(FpvSecCm``NAME_``,                                                                    \
          $rose(HIER_.ERR_NAME_) && !(GATE_) |-> ##[0:MAX_CYCLES_] (ERR_),                      \
          CLK_, RST_)                                                                           \
  `ifdef INC_ASSERT                                                                             \
    assign HIER_.unused_assert_connected = 1'b1;                                                \
  `endif
```
I guess we probably could do the logic here by replacing `MAX_CYCLES_` by something like ``MAX_CYCLES ? MAX_CYCLES : `_SEC_CM_ALERT_MAX_CYC``, giving this:
```systemverilog
`define ASSERT_ERROR_TRIGGER_ERR(NAME_, HIER_, ERR_, GATE_, MAX_CYCLES_, ERR_NAME_, CLK_, RST_) \
  `ASSERT(FpvSecCm``NAME_``,                                                                    \
          $rose(HIER_.ERR_NAME_) && !(GATE_) |->                                                \
            ##[0:MAX_CYCLES ? MAX_CYCLES_ : `SEC_CM_ALERT_MAX_CYC] (ERR_),                      \
          CLK_, RST_)                                                                           \
  `ifdef INC_ASSERT                                                                             \
    assign HIER_.unused_assert_connected = 1'b1;                                                \
  `endif
```

It's pretty complicated though! So I'm going to file that as a separate PR that follows up.

rswarbrick:
I'm reasonably confident that this change is reasonable: I (and others) have looked at the RTL logic and I'm reasonably convinced that it's sensible. I've also done some extra testing for a few blocks that use lots of FIFOs internally (using 6b41f24).

For csrng, I ran `util/dvsim/dvsim.py hw/ip/csrng/dv/csrng_sim_cfg.hjson -i all --reseed-multiplier=0.1`. With the tip of `master`, this just saw a single failure (in `csrng_stress_all_with_rand_reset`, due to not managing to find a time to issue the reset). With this patch, there was the same single failure.

For entropy_src, I ran `util/dvsim/dvsim.py hw/ip/entropy_src/dv/entropy_src_sim_cfg.hjson -i all --reseed-multipli
er=0.1`. Upsettingly, this saw lots of failures (51 failures out of 259 tests). With the tip of master or the version with the patch, this saw essentially the same failure rate.

For edn, I ran `util/dvsim/dvsim.py hw/ip/edn/dv/edn_sim_cfg.hjson -i all -mp 16 --reseed-multiplier=0.1`. Two tests timed out (stress all with rand reset), but everything else passed.

vogelpi:
Thanks a lot Rupert! This sounds reasonable to me. I've checked ENTROPY_SRC and all the FIFOs used in there have a depth different from 1.

vogelpi:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_assert_sec_cm.svh
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_fifo_assert.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/flash_ctrl/rtl/flash_ctrl.sv

This PR adds a special case for FIFOs with a depth equal 1, primarily to enable coverage closure. The behavior of the RTL isn't changed. This is fine.

vogelpi:
There are currently lots of line length Lint warnings flagged by AscentLint in OTBN and FLASH_CTRL (e.g. `W   LINE_LENGTH:   prim_fifo_assert.sv:18`). Other than that this PR looks fine to me.

vogelpi (approved):

rswarbrick:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_assert_sec_cm.svh
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_fifo_assert.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/flash_ctrl/rtl/flash_ctrl.sv

This will only have an effect on FIFOs with depth 1, which have been checked reasonably closely. It also avoids some impossible coverage points, which should help with coverage closure.

rswarbrick:
Hmm. My trick to add the file to AscentLint waivers doesn't work. It turns out that this is because we end up waiving stuff on a file that isn't used (and thus isn't included) in most blocks, which causes a warning itself. Grr.

I'll change things to ensure that the file is always included, avoiding the problem.

hw/ip/prim/rtl/prim_fifo_assert.sv
a-will, line 1: nit: This should have the .svh extension, shouldn't it?

a-will:
> Hmm. My trick to add the file to AscentLint waivers doesn't work. It turns out that this is because we end up waiving stuff on a file that isn't used (and thus isn't included) in most blocks, which causes a warning itself. Grr.
> 
> I'll change things to ensure that the file is always included, avoiding the problem.

I wonder if that warning could also be waived (possibly for header / .svh files in general).

hw/ip/prim/rtl/prim_fifo_assert.sv
rswarbrick: I was intentionally matching `prim_assert.sv` (with no "h"!)

But I don't have strong opinions. Maybe switching to `svh` would work: I'll try it.

rswarbrick:
I don't think I know a way to "waive the waive warning". I agree that it would be a reasonable thing to do, but I'm not sure how to do it. I've just pushed up a version with the file renamed to `.svh` (which I agree makes more sense) but not telling Fusesoc that it's an include file (which insures it always gets included anyway, to avoid the warning from AscentLint).

hw/ip/prim/rtl/prim_fifo_assert.sv
a-will: > I was intentionally matching prim_assert.sv (with no "h"!)

I guess we don't exactly follow our own style guide, then :laughing: 

https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#file-extensions

rswarbrick:
Eugh. That doesn't work at all. I'll try to come up with something more sensible tomorrow.

rswarbrick:
Ok, my understanding of the problem:
- If you don't mark the file with `is_include_file` for fusesoc then you can't include it. (Not unreasonable!)
- But if you *do* mark it as such, then the file will only get shown to AscentLint if it happens to get included.
- AscentLint warns if it sees a waiver that applies to a file that it hasn't seen.

This doesn't cause a problem for `prim_assert.sv` because (I believe) every block uses at least one assertion, so the file always gets included. The other files with similar waivers (e.g. `prim_assert_sec_cm.svh`) are included by `prim_assert.sv`, so are always included as well.

Oof. My solution is to include the header file from `prim_fifo_sync.sv`, which is in the same core as it. This seems to work with some local testing so I'm *hoping* it will be the right way to get everything working.

vogelpi:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_fifo_assert.svh

(The authorization for prim_fifo_assert.sv earlier no longer works because I changed the commit to use the right file name!)

PR #25462, opened 2024-11-29
[SiVal, usbdev] Add harness to usbdev_test

engdoreis:


sw/host/tests/chip/usb/usbdev_smoketest.rs
```diff
@@ -0,0 +1,72 @@
...
+    let port = ports
+        .iter()
+        .find(|&port| {
+            if let serialport::SerialPortType::UsbPort(info) = &port.port_type {
+                info.vid == opts.usb.vid && info.pid == opts.usb.pid
+            } else {
+                false
+            }
+        })
+        .unwrap();
```
pamaury, line 48: I think it would be better to first create the list of ports and then print an error if:
- the list is empty (indicating that the serial port was not detected)
- the list is too long (indicating that several devices match)

Note that in CI, unfortunately, it is possible for several devices to match because we don't use a unique VID and PID, and all devices are visible (but not accessible) to every program. One possibility to avoid this issue might be to list all ports and for all matches to try to open them: the ones belonging to other containers cannot be opened and so should be discarded. Probably it's a good idea to print a message. So overall it would look like:
```rust
  let mut usb_uart = None;
  for port in ports {
    if let serialport::SerialPortType::UsbPort(info) = &port.port_type
      if info.vid == opts.usb.vid && info.pid == opts.usb.pid {
        let uart = SerialPortUart::open(&port.port_name, 115200);
        match uart {
          Ok(uart) => {
             match usb_uart {
               None => usb_uart = Some(uart),
               Some(_) => println!("Warning: several devices match, picked up the first one which might be wrong");
             }
           Err(err) => println!("Note: ignoring device {} which cannot be opended", port.port_name),
          }
        }
      }
    }
  }
//
```
It's quite ugly, probably can be written in a better way.

pamaury (approved):

PR #25463, opened 2024-11-29
[dv] Create and deploy reset agent

martin-velay:


martin-velay:
Hi DV team (@rswarbrick, @hcallahan-lowrisc, @antmarzam, @KinzaQamar)

Here is the PR I was talking about earlier today. I was expecting to have something in better to share by the end of this day but unfortunately things didn't went well. It compiles, elaborate, start to run and fail for a reason I cannot explain yet. I'll work on it later next week. This PR will be split into multiple later as I think these changes can happen incrementally. I haven't done that so far as I was not sure of the implementation.

You can start to have a look and give me feedback on some points which are supposed to be done at my end:
1. About the reset agent itself.
2. About the way I have started to make the reset shared with the other components through TLM interface.

-> Other comments are welcome either

What I still need to address:
1. Make sure the current implementation works.
2. Splitting the dv_lib into 3: dv_base_agent_lib, dv_base_env_lib and reset_agent.
4. Managing multiple reset agents, as discussed it's a must have for some env.
5. Trigger on the fly resets from the IP specific base test when rand_reset_vseq is called (not from the base test from dv_lib for the moment as not all the IP will move to this way directly, or maybe do it via a config bit somewhere).
6. Address all my comments tagged with "TODO MVy" (BTW if you have some comments on these I' be happy to have inputs on some).

-> Do you see anything else?

When all these points will be addressed and will work:
- Make all DV env transitioning to reset agent adoption.
- This should simplify some piece of code as highlighted with the "TODO MVy". And it will also uniformize some points.
- Remove the old reset triggering way.

-> Do you see anything else?

Note: @rswarbrick, I don't find Marcelo for some reason. Can you help me ?

rswarbrick:
@marcelocarvalhoLowRisc (I don't seem to be able to add Marcelo as a reviewer but I'm not entirely sure why. Maybe because he hasn't had any PRs merged yet?)

vogelpi:
> @marcelocarvalhoLowRisc (I don't seem to be able to add Marcelo as a reviewer but I'm not entirely sure why. Maybe because he hasn't had any PRs merged yet?)

@marcelocarvalhoLowRisc was a member of our GitHub org but not part of the OpenTitan GitHub team. I've now invited him.

On a different note: I've talked to @martin-velay and as outlined in his description above this is a draft PR not intended to be merged as is. There is still work to be done to get this in a mergeable state (including splitting it up into smalller, reviewable chunks) but Martin wanted to give people a chance to take a look asap and if possible to provide early feedback to check this makes sense to others as well. I think this is actually perfectly fine but of course, once the implementation becomes more mature and this needs real review and merging, the size has to be reduced. Martin is aware of this.

FYI @rswarbrick , @mundaym .

martin-velay:
Status update:

I have pushed all my changes. This PR is still not ready, the reset is not happening as it should yet. I need to pause this activity to be focused on V3 activities.

What should be done next: find a way of disabling the main process when a reset is happening and then restart. This should probably imply to have a fork with a reset and main thread as done in many components, but this be done in the main virtual sequence. But the this main sequence shouldn't be killed by using the stop_sequences from the sequencer, otherwise will run at all. Instead, a task should be spawned from the main vseq and this task should be killed by calling disable fork and then restarted. This should happen as many as defined by a control knob which says how many resets should happen. By default this number will be equal to 1 (for initial reset).

hw/dv/sv/reset_agent/reset_agent_cov.svh
```diff
@@ -0,0 +1,73 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+class reset_agent_cov extends dv_base_agent_cov #(reset_agent_cfg);
+  `uvm_component_utils(reset_agent_cov)
+
+  // Class should extends uvm_subscriber #(reset_item), not sure what to do here
+  // Option 1: class reset_agent_cov extends uvm_subscriber #(reset_item);
+  // Option 2: declare the following and extend dv_base_agent_cov
+  typedef uvm_subscriber #(reset_item) this_type;
+	uvm_analysis_imp #(reset_item, this_type) analysis_export;
```
antmarzam: I've never seen it done this way, what are you trying to do? 

Would it not be `uvm_analysis_imp # (reset_item, reset_agent_cov)`?

Another option would be to create a dv_base_subscriber, and have `dv_base_agent_cov` extend from `dv_base_subscriber`

hw/dv/sv/reset_agent/reset_agent.svh
```diff
@@ -0,0 +1,54 @@
...
+  .DRIVER_T       (reset_driver),
+  .SEQUENCER_T    (reset_sequencer),
+  .MONITOR_T      (reset_monitor),
+  .COV_T          (reset_agent_cov)
+);
+
+  `uvm_component_utils(reset_agent)
+
+  uvm_analysis_port #(reset_item) reset_tr_ap;
+  uvm_analysis_port #(reset_state_e) reset_st_ap;
```
antmarzam: What were you thinking with these APs? One is a a seq_item, the other one is an event? 

If so I think the type should be `event`. Having said that, I believe that a seq_item should be enough, and if we need events we can add it to the seq item?

On another point, Why are you declaring analysis ports in the agent? Do you plan  make the APs accessible at the agent level, so when we connect the agent with an environment you don't have to do: `agent.monitor.connect(..)`?

hw/dv/sv/reset_agent/reset_agent_cov.svh
```diff
@@ -0,0 +1,73 @@
...
+      bins bin_zero   = {0};
+      bins bin_one    = {1};
+      bins bin_short  = {[2:100]};
+      bins bin_long   = {[101:$]};
+    }
+
+    cp_assert_width: coverpoint assert_width {
+      bins bin_one    = {1};
+      bins bin_short  = {[2:100]};
+      bins bin_long   = {[101:$]};
```
antmarzam: I think the bins `zero` and `one` are OK, but we probably need a bit more granularity instead of short/long.

Maybe something like: `zero`, `one`, `less_than_ten`, `less_than_hundred`, `rest`?

hw/dv/sv/reset_agent/reset_driver.svh
```diff
@@ -0,0 +1,113 @@
...
+  reset_item tr;
+
+  phase.raise_objection(this);
+  // Build the initial transaction as the first reset should occur from the reset_phase
+  tr = reset_item::type_id::create("tr");
+  if (!tr.randomize()) begin
+    `uvm_fatal(`gfn, "Randomized failed !");
+  end
+  drive_trans(tr);
+  phase.drop_objection(this);
```
antmarzam: I'm not a big fan of the driver raising objections on it's own (there may be cases where it's necessary though)
I would prefer that the class handling the reset sequence is the one raising and dropping objections

hw/dv/sv/reset_agent/reset_interface.sv
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+interface reset_if ();
+  // Inputs
+  logic clk_i;
+  // Outputs
+  logic rst_o;
```
antmarzam: I'd be tempted to drop the `_i` `_o` since the direction for this will be dependant on how the agent is used?

hw/dv/sv/reset_agent/reset_item.svh
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class reset_item extends uvm_sequence_item;
+  rand int unsigned assert_delay;   // If reset is sync: number of clock cycles, async: time in ns
+  rand int unsigned assert_width;   // If reset is sync: number of clock cycles, async: time in ns
+
```
antmarzam: Maybe we should have a default "relaxed" constraint with "sensible" values, so that if we run  `reset_base_seq` we still don't get a delay/width of like `2**32 cycles`. Then each sequence can override it via in-line constraints.
Although if we just ensure we never use `reset_base_seq` and always use `reset_seq` we'd be OK I think

hw/dv/sv/dv_lib/reset_agent/reset_agent_cfg.svh
```diff
@@ -1,25 +1,32 @@
...
+  // Reset polarity: ActiveLow / ActiveHigh
+  polarity_e    polarity = ActiveLow;
+  // Reset assertion should be synchronous to the input clock
+  bit           assert_is_sync = 0;
+  // Reset deassertion should be synchronous to the input clock
+  bit           deassert_is_sync = 1;
+  // Initial reset delay
+  int unsigned  ini_assert_delay = 100;
+  // Initial reset duration
+  int unsigned  ini_assert_width = 100;
```
antmarzam: Can these maybe be `rand`? so we could in the dv_base_config randomise this class after we create it?

hw/dv/sv/dv_lib/reset_agent/reset_agent_cov.svh
```diff
@@ -1,73 +1,73 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+class reset_agent_cov extends dv_base_agent_cov #(reset_agent_cfg);
+  `uvm_component_utils(reset_agent_cov)
+
+  // Class should extends uvm_subscriber #(reset_item), not sure what to do here
+  // Option 1: class reset_agent_cov extends uvm_subscriber #(reset_item);
+  // Option 2: declare the following and extend dv_base_agent_cov
+  // typedef uvm_subscriber #(reset_item) this_type;
+  uvm_analysis_imp #(reset_item, reset_agent_cov) analysis_imp;
```
antmarzam, line 8: ignore my previous comment on this, I see that you changed it (I'm reviewing commit by commit)

hw/dv/sv/dv_lib/reset_agent/seq_lib/reset_seq.svh
```diff
@@ -1,45 +1,45 @@
...
-endtask : body
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+class reset_seq extends reset_base_seq;
+  `uvm_object_utils(reset_seq)
+
+  rand int unsigned assert_delay;
+  rand int unsigned assert_width;
```
antmarzam, line 9: I would move these to the base sequence

hw/ip/hmac/dv/tb/tb.sv
```diff
@@ -22,7 +22,8 @@ module tb;
   wire intr_hmac_err;
 
   // interfaces
-  clk_rst_if clk_rst_if(.clk(clk), .rst_n(rst_n));
+  clk_rst_if clk_rst_if(.clk(clk), .rst_n());
+  reset_interface reset_if();
```
antmarzam: I'm not sure what you're trying to do here. 
Maybe the `clk_rst_if` can now become a `clk_if` that also has an handle to `reset_interface`?

hw/dv/sv/dv_lib/reset_agent/reset_agent.svh
```diff
@@ -43,9 +42,10 @@ function void reset_agent::connect_phase(uvm_phase phase);
     driver.seq_item_port.connect(sequencer.seq_item_export);
   end
 
-  // Connect the monitor port
-  reset_tr_ap = monitor.reset_item_ap;
-  reset_st_ap = monitor.reset_state_ap;
+  // Connect the analysis ports from the monitor to the agent to simplify further usage
+  // while instanciating this agent.
+  monitor.reset_item_ap.connect(reset_tr_ap);
+  monitor.reset_state_ap.connect(reset_st_ap);
```
antmarzam, line 48: What do we achieve with this double AP? Can reset_state_e not be a field within the sequence item?

hw/dv/sv/dv_lib/dv_base_driver.sv
```diff
@@ -2,6 +2,10 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+// Use this UVM macro as we may need to implement multiple uvm_analysis_imp, which means
+// implemneting multiple write methods which is not possible with the same name.
+`uvm_analysis_imp_decl(_drv_reset)
```
antmarzam, line 7: I think using the macro is fine, but in which circumstances do you anticipate needing more than one? I can't think of one.

hw/dv/sv/dv_lib/dv_base_driver.sv
```diff
@@ -11,18 +15,68 @@ class dv_base_driver #(type ITEM_T     = uvm_sequence_item,
...
+    // After each reset, the current transaction should be dropped and get_and_drive should be
+    // restarted once the reset is being deasserted.
+    forever begin
+      // This isolation fork is needed to ensure that "disable fork" call won't kill any other
+      // processes at the same level from the base classes
+      fork begin : isolation_fork
+        fork
+          begin : main_thread
+            get_and_drive();
+            wait(0);  // Wait indefinitely to ensure the fork will end because of a reset detection
```
antmarzam, line 56: Are you expecting `get_and_drive` to have a `forever` loop with a `get_next_item()` inside it?
It may be easier this way since I believe this is what most drivers do, but it may make the driver code a bit tidier if we did here:
```
forever begin
  get_and_drive();
  wait (1); // I think this may be necessary in case `get_and_drive` was empty?
end
```

antmarzam:
@martin-velay Thanks a lot for putting this together!

I've left a few comments, but overall it feels this PR is taking things on the right direction.

The only thing I'm missing is updated to `dv_base_driver` and a new `seq_item` in the `dv_base_lib` like `dv_base_item` that we have all seq_items extend from.
In the DV driver, the task `get_and_drive` shouldn't be empty. They should be able to:

1.  get_item
2.  Do something with it (in the child base driver). But probably a `pure virtual` function in the dv_base_driver that will get populated in the child classes.
3. call item_done.
4. detect reset, and set a reset = 0/1 field in the seq_item, so that the UVM sequences can self-kill whenever reset is set.

P.S. You may need to rebase to the latest master when you get the chance!

PR #25464, opened 2024-11-29
[prim_fifo_sync,fpv+rtl] Add checks for the full_o output

rswarbrick:
When working on #25461, I noticed that we didn't have any assertions about the behaviour of the `full_o` output port. This PR adds those assertions in the second commit, and makes them true (in a trivial way) in the first.


hcallahan-lowrisc (approved):
Changes LGTM :+1:

PR #25465, opened 2024-11-29
[prim_fifo_sync,fpv] Add checks for the full_o output

rswarbrick:
This output doesn't really do anything in the FPV world (because it detects error injection which we don't model), but wiring it up in a trivial way avoids a warning in Jasper.

To avoid merge conflicts, this PR builds on #25464, which contains the first two commits. The only commit that is unique to this PR is the last commit.

### DIFF
--- a/hw/ip/prim/rtl/prim_fifo_sync.sv
+++ b/hw/ip/prim/rtl/prim_fifo_sync.sv
@@ -46,7 +46,7 @@ module prim_fifo_sync #(
 
     // host facing
     assign wready_o = rready_i;
-    assign full_o = rready_i;
+    assign full_o = 1'b1;
 
     // this avoids lint warnings
     logic unused_clr;
--- a/hw/ip/prim/fpv/tb/prim_fifo_sync_bind_fpv.sv
+++ b/hw/ip/prim/fpv/tb/prim_fifo_sync_bind_fpv.sv
@@ -29,6 +29,7 @@ module prim_fifo_sync_bind_fpv;
     .rvalid_o,
     .rready_i,
     .rdata_o,
+    .full_o,
     .depth_o
   );
 
@@ -47,6 +48,7 @@ module prim_fifo_sync_bind_fpv;
     .rvalid_o,
     .rready_i,
     .rdata_o,
+    .full_o,
     .depth_o
   );
 
@@ -65,6 +67,7 @@ module prim_fifo_sync_bind_fpv;
     .rvalid_o,
     .rready_i,
     .rdata_o,
+    .full_o,
     .depth_o
   );
 
@@ -83,6 +86,7 @@ module prim_fifo_sync_bind_fpv;
     .rvalid_o,
     .rready_i,
     .rdata_o,
+    .full_o,
     .depth_o
   );
 
@@ -101,6 +105,7 @@ module prim_fifo_sync_bind_fpv;
     .rvalid_o,
     .rready_i,
     .rdata_o,
+    .full_o,
     .depth_o
   );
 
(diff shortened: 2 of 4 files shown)


matutem (approved):
LGTM

hw/ip/prim/rtl/prim_fifo_sync.sv
```diff
@@ -46,7 +46,7 @@ module prim_fifo_sync #(
 
     // host facing
     assign wready_o = rready_i;
-    assign full_o = rready_i;
+    assign full_o = 1'b1;
```
vogelpi, line 49: I agree that this is much better than what we had before (it really didn't make sense). However, to avoid potential DV fallout and people spending time on debugging: would you mind checking we indeed have no prim_fifo_sync instance with a depth of 0 in the upstream repository, please?

rswarbrick: Good point. I've just done a (rather boring) search, looking at hits for `prim_fifo_sync #`. Notes below and I'm not sure whether there are any instances with depth zero but I'm certain that any such instances don't connect up the `full_o` signal.

- `csrng_block_encrypt.sv`: Depth 1.
- `csrng_cmd_stage.sv`: Depth 16.
- `csrng_ctr_drbg_cmd.sv`: Lots of instances but all have positive depth.
- `edn_core.sv`: Both instances have depth 13.
- `entropy_src_core.sv`:
  + `u_prim_fifo_sync_esrng`: Depth 2.
  + `u_prim_fifo_sync_distr`: Depth is DistrFifoDepth but the code for
    `sfifo_distr_err` wouldn't make much sense (before or after my
    change) if `sfifo_distr_full` was constant.
  + `u_prim_fifo_sync_observe`: Again, this comes from a parameter but
    the code for `sfifo_observe_gate_d` and `sfifo_observe_err`
    wouldn't make sense for a zero depth.
  + `u_prim_fifo_sync_esfinal`: As with the previous two items, this
    code wouldn't really make sense for an empty fifo.
- `i2c_fifo_sync_sram_adapter.sv`: Depths 2, 2
- `kmac_msgfifo.sv`: Depth is parameterised with
  `kmac_pkg::MsgFifoDepth`, which is at least 2.
- `otp_ctrl.sv`: Depth 2
- `prim_sram_arbiter.sv`: Depth 4.
- `prim_generic_flash_bank.sv`: Depth 2.
- `spi_tpm.sv`: Depth 1.
- `spid_readsram.sv`: Depths 1, 2.
- `spi_host_command_queue.sv`: `full_o` not connected.
- `spi_host_data_fifos.sv`: `full_o` not connected in either instance.
- `tlul_adapter_sram.sv`: `full_o` not connected in any of the four
  instantiations.
- `tlul_fifo_sync.sv`: `full_o` not connected in either instance.
- `tlul_sram_byte.sv`: `full_o` not connected in either instance.
- `uart_core.sv`: `full_o` not connected in either instance.
- `usbdev.sv`: Depths 4, 8, 8.
- `flash_ctrl.sv.tpl`: `full_o` not connected in either instance.
- `flash_phy.sv`: `full_o` not connected in either instance.
- `flash_phy_rd.sv`: `full_o` not connected in any of the 4 instances.
- `flash_ctrl.sv`: `full_o` is only connected for `u_sw_rd_fifo`. This
  has depth `RdFifoDepth` (a parameter), which is 16 for `top_earlgrey`.
- `dm_csrs.sv`: `full_o` is not connected.

rswarbrick:

rswarbrick:
The failing DV smoke test is because of an unrelated breakage (which was my fault, but not here!) which is now fixed. Rebasing locally makes the check pass.

PR #25466, opened 2024-11-29
[prim,fpv] Fix width warnings in prim_fifo_sync_assert_fpv

rswarbrick:


hw/ip/prim/fpv/vip/prim_fifo_sync_assert_fpv.sv
```diff
@@ -30,6 +30,9 @@ module prim_fifo_sync_assert_fpv #(
   input              err_o
 );
 
+  localparam [DepthW+2:0] WideOne = (DepthW + 3)'(1'b1);
```
marnovandermaas, line 33: I assume this is the solution?
```suggestion
  localparam logic [DepthW+2:0] WideOne = (DepthW + 3)'(1'b1);
```

rswarbrick: Oops! Thanks.

PR #25472, opened 2024-11-30
[hw,pwrmgr_common_pkg,rtl] Move common interface definitions out of ipgen

Razer6:
Some IPs depend on the power manager package to define their IO interfaces, for example the rv_core_ibex or flash control. However, some tops do not have a power manager, and therefore also not the ipgen'ed power manager package. This causes a dependency error.

To solve that problem, we put out common type definitions to a common `pwrmgr_common_pkg`, which is not part of ipgen. Thus, also tops that don't have a pwrmgr, can reference these definitions.

### DIFF
--- a/hw/ip/rv_core_ibex/data/rv_core_ibex.hjson
+++ b/hw/ip/rv_core_ibex/data/rv_core_ibex.hjson
@@ -153,11 +153,11 @@
       package: "lc_ctrl_pkg",
     },
 
-    { struct:  "pwr_cpu",
+    { struct:  "cpu_pwrmgr",
       type:    "uni",
       name:    "pwrmgr",
       act:     "req",
-      package: "pwrmgr_pkg",
+      package: "rv_core_ibex_pkg",
     },
 
     { struct:  "logic",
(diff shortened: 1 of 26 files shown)


matutem:
I think this can be addressed with a lot less shuffling if pwr_cpu_t was renamed cpu_pwr_t and its declaration was moved to rv_core_ibex_pkg instead. It is also sensible since this is from cpu to pwrmgr. This would also avoid "fragmenting" pwrmgr.

andreaskurth (approved):
LGTM, thx @Razer6 and @matutem :+1:

vogelpi:
CHANGE AUTHORIZED: hw/ip/rv_core_ibex/data/rv_core_ibex.hjson
CHANGE AUTHORIZED: hw/ip/rv_core_ibex/rtl/rv_core_ibex.sv
CHANGE AUTHORIZED: hw/ip/rv_core_ibex/rtl/rv_core_ibex_pkg.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/pwrmgr/data/pwrmgr.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/pwrmgr/rtl/pwrmgr_pkg.sv

This just renames the pwr_cpu struct to cpu_pwrmgr and moves it from pwrmgr_pkg to
rv_core_ibex_pkg. No functional change.

PR #25473, opened 2024-11-30
[darjeeling] Regenerate Darjeeling after tooling changes

Razer6:
Some tooling changes on the generated files are not yet reflected such as external alerts on so on. Re-generate it as a separate PR to avoid having that diff on a new feature PR.

### DIFF
--- a/hw/top_darjeeling/data/autogen/top_darjeeling.gen.hjson
+++ b/hw/top_darjeeling/data/autogen/top_darjeeling.gen.hjson
@@ -17205,6 +17205,7 @@
       lpg_idx: 11
     }
   ]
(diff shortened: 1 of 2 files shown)


rswarbrick (approved):
Looks sensible to me. I think this shows we're missing something in our CI though :-) Looking at `check-generated.sh`, my guess is that something isn't wired up properly for `make -C hw` calls.

What command are you running to regenerate these files?

Razer6:
Darjeeling is not yet in CI, therefore.

The "new" command is ` make -C hw top_and_cmdgen`. See also https://github.com/lowRISC/opentitan/pull/25022

PR #25478, opened 2024-12-02
[rom_ctrl, dv] tlul_adapter_sram conditional coverage

KinzaQamar:
There were bunch of statements in tlul_adapter_sram that needs to get excluded as the cases were non-occuring (refer rom_ctrl_cov_excl.el).

Moreover, there is a removal of an item (reqfifo_rvalid) in the design of tlul_adapter_sam.sv to cover a conditional coverage case as the item implies another item inside the conditional coverage.

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
+++ b/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
@@ -38,3 +38,77 @@ ANNOTATION: "Waived the assignment as the input is wired to a parameter"
 Block 1 "3318159292" "data_o = 39'(data_i);"
 INSTANCE: tb.dut.u_tl_adapter_rom.u_tlul_data_integ_enc_instr.u_data_gen
 Block 1 "3318159292" "data_o = 39'(data_i);"
+
+INSTANCE: tb.dut.u_tl_adapter_rom
+// wr_attr_error and wr_vld_error both depend on a_opcode != Get. wr_attr_error can exist without
+//wr_vld_error but ErrOnWrite is wired to 1 inside the instantiation of tlul_adapter_sram in
+//rom_ctrl.sv. Hence, if wr_attr_error is true then tl_i.a_opcode is PutFullData or PutPartialData
+//which automatically sets wr_vld_error.
+Condition 17 "3469950311" "(wr_attr_error | wr_vld_error | rd_vld_error | instr_error | tlul_error |
+                            intg_error) 1 -1" (7 "100000")
+
+// Because en_ifetch_i is hardwired to MuBi4True, instr_error can only be true if
+// tl_i.a_user.instr_type is not a valid mubi value. But that gets seen in u_err as instr_type_err,
+// causing tlul_error to be true in the adapter.
+Condition 17 "3469950311" "(wr_attr_error | wr_vld_error | rd_vld_error | instr_error | tlul_error |
+                            intg_error) 1 -1" (4 "000100")
+
+// For rspfifo_rdata.error, it is pushed in rspfifo_wdata.error. If the error field is true, it
+// means rerror_i[1] bit is true but the rerror_i bits are wired to 0 in the instantiation of the
+// tlul_adapter_sram in rom_ctrl.sv.
+Condition 4 "3638058042" "(rspfifo_rdata.error | reqfifo_rdata.error) 1 -1" (3 "10")
+
+// The 011 case is not possible.
+//
+// Since we are asking that rspfifo_rvalid = 1, we know that if reqfifo_rvalid is true then d_valid
+// will be true. As we ask that d_valid = 0, it follows that reqfifo_rvalid is false. This means
+// that u_reqfifo is empty.
+//
+// For rspfifo_rvalid = 1, we must either have a response arriving now (which is visible because
+// Pass=1) or have a stored response. In the first case, reqfifo_rvalid must be true (which is
+// checked with the rvalidHighReqFifoEmpty assertion). But we know that reqfifo_rvalid must be
+// false.
+//
+// For the second case, we must have had the request in the fifo and popped it before now. But this
+// isn't possible, because the ROM always responds in a single cycle: the same cycle we would be
+// popping the request from u_reqfifo.
+Condition 21 "3623514242" "(d_valid & rspfifo_rvalid & (reqfifo_rdata.op == OpRead)) 1 -1" (1 "011")
+
+// Condition 24 till 34 are excluded for the same reason. The requirements for the conditions are:
+// 1) reqfifo_rvalid --> 1
+// 2) rspfifo_rvalid --> 1
+// 3) reqfifo_rdata.op --> opread
+// 4) reqfifo_rdata.error --> 1
+//
+// We cannot see vld_rd_rsp && d_error. If d_error is true, we must have rspfifo_rdata.error or
+// reqfifo_rdata.error. The first cannot happen, because the items in the response fifo have an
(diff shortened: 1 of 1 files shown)


hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -38,3 +38,43 @@ ANNOTATION: "Waived the assignment as the input is wired to a parameter"
...
+
+// Condition 24 till 34 are excluded for the same reason. The requirements for the conditions are:
+// 1) reqfifo_valid --> 1
+// 2) rspfifo_valid --> 1
+// 3) reqfifo_rdata.op --> opread
+// 4) refifo_rdata.error --> 1
+// For condition 24 to 26, we don't expect a rspfifo_rvalid with a reqfifo_rdata.error. Thus it
+// restrict the case 10, 1, 11 and 01 for conditions 24 till 26 as non-ocurring. For condition 34,
+// d_error needs reqfifo_rvalid and reqfifo_rdata.error to get set which automatically set the
+// d_valid. So 01 for condition 24 is non-occuring.
```
rswarbrick, line 69: I think this would probably be easier to follow if you wrote some text about condition 24 (following the form you've got here) and then wrote something shorter for each of 25, 26, 34 that explains why the same argument works for it too.

For condition 24, it might be something like:
```
// We cannot see vld_rd_rsp && d_error. If d_error is true, we must have rspfifo_rdata.error or
// reqfifo_rdata.error. The first cannot happen, because the items in the response fifo have an
// error signal from rerror_i[1] and rom_ctrl connects rerror_i to zero. For the second to happen,
// we must have pushed in an item with error = 1, which means that error_internal was set. But that
// will have squashed the request, so req_o was low. As a result, the request fifo item will be
// popped again before anything appears in the response fifo.
```

For condition 25, I think we're talking about the situation where `vld_rd_rsp && !reqfifo_rdata.error`? I'm slightly surprised about that! (Wouldn't we expect to see that on every read?)

For condition 26, you could refer to condition 24. Or (maybe better!) use something like what I wrote above as a description of condition 26 and then make the description of condition 24 explain why it would imply 26, which can't happen.

For condition 34, I'm a bit surprised again. Here, we're talking about `d_valid && d_error`. The argument above shows that this can only happen when we have an error in the request. But don't we see this in the TL errors sequence?

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -38,3 +38,43 @@ ANNOTATION: "Waived the assignment as the input is wired to a parameter"
...
+Condition 24 "1059982851" "(vld_rd_rsp & ((~d_error))) 1 -1" (2 "10")
+Condition 25 "2807788926" "((vld_rd_rsp && reqfifo_rdata.error) ? error_blanking_integ :
+                            (vld_rd_rsp ? rspfifo_rdata.data_intg :
+                            prim_secded_pkg::SecdedInv3932ZeroEcc)) 1 -1" (2 "1")
+Condition 26 "561780173" "(vld_rd_rsp && reqfifo_rdata.error) 1 -1" (3 "11")
+Condition 34 "201396280" "(d_valid && d_error) 1 -1" (1 "01")
+
+// When access is granted to the bus and rom respond to the bus request leads the rvalid_i signal
+// go high and reqfifo_rvalid gets set on that cycle until the read happens. Hence, 10 is
+// non-occuring for the conditional statement.
```
rswarbrick, line 79: I had to re-convince myself about why this was correct. I'm not sure the text here makes it particularly obvious. Maybe something like this instead?
```systemverilog
// It is impossible to see rvalid_i && !reqfifo_rvalid. If rvalid_i is true, the ROM is responding
// to a request that was sent on the previous cycle. Since it always responds in exactly one cycle,
// the request will not already have been popped from the request fifo, so reqfifo_rvalid must be
// true.
```

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
KinzaQamar: > For condition 25, I think we're talking about the situation where vld_rd_rsp && !reqfifo_rdata.error? I'm slightly surprised about that! (Wouldn't we expect to see that on every read?)

No, the case here is to see the condition true i.e. (vld_rs_rsp && reqfifo_rdata.error) --> 1

> For condition 34, I'm a bit surprised again. Here, we're talking about d_valid && d_error. The argument above shows that this can only happen when we have an error in the request. But don't we see this in the TL errors sequence?

The case here is !d_valid && d_error.

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -38,3 +38,87 @@ ANNOTATION: "Waived the assignment as the input is wired to a parameter"
...
+//
+// We cannot see vld_rd_rsp && d_error. If d_error is true, we must have rspfifo_rdata.error or
+// reqfifo_rdata.error. The first cannot happen, because the items in the response fifo have an
+// error signal from rerror_i[1] and rom_ctrl connects rerror_i to zero. For the second to happen,
+// we must have pushed in an item with error = 1, which means that error_internal was set. But that
+// will have squashed the request, so req_o was low. As a result, the request fifo item will be
+// popped again before anything appears in the response fifo.
+Condition 24 "1059982851" "(vld_rd_rsp & ((~d_error))) 1 -1" (2 "10")
+
+// The reason for condition 24 explains why we can't get (vld_rd_rsp && reqfifo_rdata.error) ---> 1
```
rswarbrick, line 87: This text looks good to me, but should we use "=" instead of "--->"? (I'm not sure I know what "--->" means!)

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -38,3 +38,87 @@ ANNOTATION: "Waived the assignment as the input is wired to a parameter"
...
+// |       1        |      OpRead      |          1          |    1
+// |       1        |     !OpRead      |          1          |    1
+//
+// Excluded rspfifo_rdata.error as it is wired to 0. For rspfifo_rdata.error, it is pushed in
+// rspfifo_wdata.error. If the error field is true, it means rerror_i[1] bit is true but the
+// rerror_i bits are wired to 0 in the instantiation of the tlul_adapter_sram in rom_ctrl.sv.
+//
+// For d_error to set, the min requirement is to get reqfifo_rvalid = 1 and reqfifo_rdata.error = 1
+// but that also set d_valid.
+Condition 34 "201396280" "(d_valid && d_error) 1 -1" (1 "01")
```
rswarbrick, line 118: This took me some time to get my head around properly! But I think the following text is essentially the same proof and might be a bit easier to follow?
```
// We cannot see d_valid = 0 and d_error = 1. For d_error to be true, we need reqfifo_rvalid to be
// true. But then the only way for d_valid to be false is if rspfifo_rvalid is false (and
// reqfifo_rdata.op is OpRead).
//
// The ROM always services TL operations without back-pressure and responds in one cycle. As a
// result, after a request is be pushed into u_reqfifo, it is always popped again on the next cycle.
// As a result, if reqfifo_rvalid is true then rspfifo_wvalid will also be true and (because
// u_rspfifo is in pass-through mode) rspfifo_rvalid will be true.
```

rswarbrick (commented):
Thank you for all the work tidying this up!

rswarbrick (approved):

PR #25479, opened 2024-12-02
[kmac, sival] Add sival exec environment to kmac_app_rom_test

nasahlpa:
This commit adds the CW310 and CW340 SiVal execution environments to the kmac_app_rom_test.


sw/device/tests/BUILD
```diff
@@ -2308,6 +2308,7 @@ opentitan_test(
 _KMAC_APP_ROM_LIST = [
     ("//sw/device/lib/testing/test_rom:test_rom_sim_verilator_hashfile", "sim_verilator"),
     ("//sw/device/lib/testing/test_rom:test_rom_fpga_cw310_hashfile", "fpga_cw310_test_rom"),
+    ("//sw/device/lib/testing/test_rom:test_rom_fpga_cw310_hashfile", "fpga_cw310_sival"),
```
pamaury, line 2311: You probably need the real ROM here, not the test ROM. So I *think* it should be `//sw/device/silicon_creator/rom:mask_rom_fpga_cw310_hashfile`

nasahlpa: Ah yes, you are right. Thanks - done!

engdoreis (approved):

PR #25458, opened 2024-11-29
Update lowrisc_ibex to lowRISC/ibex@667fd20d

nasahlpa:
Update code from upstream repository
https://github.com/lowRISC/ibex.git to revision
667fd20d2ede51caececccbcbda3652074424ce2

* [rtl] Fix non-DSP reset in ibex_counter (Pascal Nasahl)
* Revert "[rtl] Fix counter reset value on FPGA" (Pascal Nasahl)
* [rtl] Fix counter reset value on FPGA (Pascal Nasahl)
* [ci] remove Azure Pipelines (Gary Guo)

PR #25480, opened 2024-12-02
[ibex,fpga] Apply Ibex FPGA patches

nasahlpa:
This PR applies two patches on top of Ibex that fix the following two FPGA related issues:

- Use `WordZeroVal` instead of `0` for initializing the FPGA register file
- Use async reset when not using a DSP for the Ibex FPGA counter

A patch is needed because:

- Without these fixes the SecureIbex parameter cannot be used on CW340 - this is needed for some SiVal tests
- These patches only touch the FPGA implementation. If we would vendor in the changes, we would get additional RTL changes that also affect the ASIC

### DIFF
--- a/hw/vendor/lowrisc_ibex/rtl/ibex_counter.sv
+++ b/hw/vendor/lowrisc_ibex/rtl/ibex_counter.sv
@@ -55,8 +55,12 @@ module ibex_counter #(
   localparam int DspPragma = CounterWidth < 49 ? "yes" : "no";
   (* use_dsp = DspPragma *) logic [CounterWidth-1:0] counter_q;
 
-  // DSP output register requires synchronous reset.
-  `define COUNTER_FLOP_RST posedge clk_i
+  if (CounterWidth < 49) begin : g_dsp_counter
+    // DSP output register requires synchronous reset.
+    `define COUNTER_FLOP_RST posedge clk_i
+  end else begin : g_no_dsp_counter
+    `define COUNTER_FLOP_RST posedge clk_i or negedge rst_ni
+  end
 `else
   logic [CounterWidth-1:0] counter_q;
 
--- a/hw/vendor/patches/lowrisc_ibex/rtl/0002-Fix-fpga-counter-reset.patch
+++ b/hw/vendor/patches/lowrisc_ibex/rtl/0002-Fix-fpga-counter-reset.patch
@@ -1,0 +1,19 @@
+diff --git a/ibex_counter.sv b/ibex_counter.sv
+index c78e510ee4..b4dc7ec347 100644
+--- a/ibex_counter.sv
++++ b/ibex_counter.sv
+@@ -55,8 +55,12 @@ module ibex_counter #(
+   localparam int DspPragma = CounterWidth < 49 ? "yes" : "no";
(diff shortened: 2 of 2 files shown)


vogelpi (approved):
This looks good to me, thanks @nasahlpa !

For awareness, I am also tagging @andreaskurth and @moidx . We need this FPGA-only RTL changes to enable using the lockstep core on the CW340 board which is required for SiVal. This change is unique to the earlgrey_1.0.0 branch and we will do a different fix (without patches) on the master branch.

moidx (approved):
I believe we still don't have bitstream caching enabled for `earlgrey_1.0.0`. @a-will can provide details on how to enable this.

a-will:
> I believe we still don't have bitstream caching enabled for `earlgrey_1.0.0`. @a-will can provide details on how to enable this.

Bitstreams are being cached, it looks like. The only thing to avoid would be using the `latest` hacky reference. bitstreams_workspace.py is currently hard-coded to look at `master/latest.txt` in the bucket for that. In general, though, the default value of "HEAD" should be used from a git repo.

vogelpi:
CHANGE AUTHORIZED: hw/vendor/lowrisc_ibex/rtl/ibex_counter.sv blocked
CHANGE AUTHORIZED: hw/vendor/lowrisc_ibex/rtl/ibex_register_file_fpga.sv

nasahlpa:
CHANGE AUTHORIZED: hw/vendor/lowrisc_ibex/rtl/ibex_counter.sv blocked
CHANGE AUTHORIZED: hw/vendor/lowrisc_ibex/rtl/ibex_register_file_fpga.sv

hw/vendor/lowrisc_ibex/rtl/ibex_counter.sv
```diff
@@ -55,8 +55,12 @@ module ibex_counter #(
...
   (* use_dsp = DspPragma *) logic [CounterWidth-1:0] counter_q;
 
-  // DSP output register requires synchronous reset.
-  `define COUNTER_FLOP_RST posedge clk_i
+  if (CounterWidth < 49) begin : g_dsp_counter
+    // DSP output register requires synchronous reset.
+    `define COUNTER_FLOP_RST posedge clk_i
+  end else begin : g_no_dsp_counter
+    `define COUNTER_FLOP_RST posedge clk_i or negedge rst_ni
+  end
```
andreaskurth, line 63: Just a note that this code relying on a pre-processor `` `define `` isn't generally correct, because the pre-processor would resolve the `` `define `` before elaboration resolves the parameterized `if`-`else` construct. However, this only affects Xilinx code (it's inside an `` `ifdef FPGA_XILINX ``) and @nasahlpa tested that this works as intended in Vivado.

On `master` a proper fix has been applied (#25458), but that would also modify ASIC RTL, which we don't allow on this branch.

andreaskurth:
CHANGE AUTHORIZED: hw/vendor/lowrisc_ibex/rtl/ibex_counter.sv
CHANGE AUTHORIZED: hw/vendor/lowrisc_ibex/rtl/ibex_register_file_fpga.sv

This only changes FPGA logic; ASIC logic isn't affected.

PR #25481, opened 2024-12-02
[prim,rtl] Rephrase a term in prim_reg_cdc.sv

rswarbrick:
This is trying to remove a coverage item that was a hole for some blocks that I was (wrongly) trying to waive with #25088. I hope this solution will be cleaner.

If DstWrReq is false then the src_update signal cannot be true. This is because it is driven by the src_update_o port on a prim_reg_cdc_arb instance. If DstWrReq is false then the instance wires src_update_o to zero (in the gen_passthru generate block).

This rewrite will behave equivalently in the same way as before but will no longer generate any conditional coverage terms that depend on src_update if DstWrReq is false.

### DIFF
--- a/hw/ip/prim/rtl/prim_reg_cdc.sv
+++ b/hw/ip/prim/rtl/prim_reg_cdc.sv
@@ -111,6 +111,24 @@ module prim_reg_cdc #(
   // This is the current destination value
   logic [DataWidth-1:0] dst_qs;
   logic src_update;
+
+  // The dst_to_src signal (on the src clock) means that data should be moving from the dst clock
+  // domain to the src clock domain on this cycle.
+  //
+  // This either means that an operation from the src side is being acknowledged or (if it's
+  // possible for the value to change on the dst side) it means that a register read response is
+  // coming from the dst clock domain. The register value may have changed from last time so should
+  // be copied back.
+  logic dst_to_src;
+  if (DstWrReq) begin : gen_wr_req
+    assign dst_to_src = src_busy_q && src_ack || src_update && !busy;
+  end else begin : gen_passthru
+    assign dst_to_src = src_busy_q && src_ack;
+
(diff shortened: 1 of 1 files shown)


rswarbrick:
Hi @vogelpi, Would you mind taking a look at this and checking the change is reasonable? If so, I'd really appreciate an explicit "change authorised" comment (and you are in a rather closer timezone than Matute!)

hw/ip/prim/rtl/prim_reg_cdc.sv
```diff
@@ -111,6 +111,21 @@ module prim_reg_cdc #(
...
+  // The dst_to_src signal (on the src clock) means that data should be moving from the dst clock
+  // domain to the src clock domain on this cycle.
+  //
+  // This either means that an operation from the src side is being acknowledged or (if it's
+  // possible for the value to change on the dst side) it means that a register read response is
+  // coming from the dst clock domain. The register value may have changed from last time so should
+  // be copied back.
+  logic dst_to_src;
+  if (DstWrReq) begin : gen_wr_req
+    assign dst_to_src = src_busy_q && src_ack | src_update && !busy;
```
vogelpi, line 124: The functionality looks good to me (thanks for the clarifying comment above!) as `src_update` can only ever be non-zero if `DstWrReq` is 1.

But:
- We should convert the bitwise OR `|` here to a boolean OR `||` for consistency. 
- `src_update` and `busy` are unused in case `DstWrReq` is 0. The latter is used in an assertion so lint may not realize this, the former should generate a lint error/warning that we should fix.

vogelpi (approved):

vogelpi:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_reg_cdc.sv

This doesn't implement a functional change but helps to improve coverage.

PR #25482, opened 2024-12-02
[rom_ctrl, rtl] Conditional coverage hole

KinzaQamar:
There is a removal of an item (reqfifo_rvalid) to remove the conditional coverage hole 1011. If reqfifo_rvalid is 0, then we won't see d_valid.


rswarbrick (commented):
I agree with the change, but can we expand the reasoning in the commit message a bit? (to explain it for people who weren't working through it with us!)

My reasoning: The only ways that `d_valid` can set to true require `reqfifo_rvalid` to be true (checked on line 282).

rswarbrick (approved):
This looks really good to me: thanks.

rswarbrick:

andreaskurth (commented):
I agree that the logical implication _`d_valid` requires `reqfifo_rvalid`_ holds in the current implementation, based on lines 282 to 292 of `tlul_adapter_sram.sv`. Thus `reqfifo_rvalid` in `d_valid & reqfifo_rvalid` is redundant and can be removed.

Maybe we break this implication in the future, though, so I think it would be valuable to add an assertion for this as we prune the logical AND above. If that makes sense to you, could you please expand your commit?

hw/ip/tlul/rtl/tlul_adapter_sram.sv
```diff
@@ -682,4 +682,6 @@ module tlul_adapter_sram
   `ASSERT_KNOWN(TlOutValidKnown_A, tl_o.d_valid)
   `ASSERT(TlOutKnownIfFifoKnown_A, !$isunknown(rspfifo_rdata) -> !$isunknown(tl_o))
 
+  // The definition of d_valid leads to the assertion below.
+  `ASSERT(reqfifo_rvalidImpliesdvalid, !reqfifo_rvalid -> !d_valid)
```
rswarbrick, line 686: Can you add a "_A" suffix to the name to match the other similar ones?

hw/ip/tlul/rtl/tlul_adapter_sram.sv
```diff
@@ -682,4 +682,6 @@ module tlul_adapter_sram
   `ASSERT_KNOWN(TlOutValidKnown_A, tl_o.d_valid)
   `ASSERT(TlOutKnownIfFifoKnown_A, !$isunknown(rspfifo_rdata) -> !$isunknown(tl_o))
 
+  // The definition of d_valid leads to the assertion below.
+  `ASSERT(ReqFifoRValidImpliesdvalid_A, !reqfifo_rvalid -> !d_valid)
```
rswarbrick, line 686: Sorry to be annoying, but the implication is the other way around from the name you have given it. Also, the expression `!A -> !B` is the same as `!B || !(!A)`, which equals `A || !B`, which is `B -> A`. So I think we probably want `d_valid -> reqfifo_rvalid` and could have a name like "DValidNeedsReqfifoRvalid_A".

andreaskurth:

No functional change, as ensured by the newly added assertion

PR #25484, opened 2024-12-02
[topgen] Use the number of alerts and not alert modules

Razer6:
Previously, the number of alert modules was used to determine the number of outgoing alerts. That's wrong. If a device has more than one alert, the number of alert modules and the actual number of alerts diverge.

This PR fixes that and uses the correct number.


andreaskurth (approved):

andreaskurth:
Skipping remaining CI checks as this template code isn't used in any upstream generated code.

PR #25492, opened 2024-12-03
[opentitanlib] remove 3b mode init of SPI console

timothytrippel:
SPI flash mode on hyperdebug defaults to 3 byte mode, and if we try to init to 3b mode if we are not in 4b mode the transaction hangs. This is a temporary fix until this can be root caused.


timothytrippel:
FYI @jesultra

milesdai (approved):

PR #25493, opened 2024-12-03
[provisioning] the DIN fields should be parsed as decimal values

timothytrippel:
The DIN fields should each be parsed as decimal values, even though they are reported as a hex string; there are just unused bits of each subfield.


sw/host/provisioning/orchestrator/src/device_id.py
```diff
@@ -61,14 +61,18 @@ def to_int(self) -> int:
         din |= self.year
         return din
 
+    @staticmethod
+    def _read_int_as_decimal_str(x: int) -> int:
```
milesdai, line 65:
```suggestion
    def _read_int_as_decimal_str(x: int) -> int:
        """Interpret a number by reading its hexadecimal representation as a decimal number.
        
        For example, `70` in hex is `0x47` and should be read as the decimal value 47.
        
        This encoding is a manufacturing equipment constraint.
        """
```

milesdai (approved):
Added a comment suggestion.

nmoroze (approved):
Good code :+1:

timothytrippel:
Sorry about that, it should be fixed in the latest version of `earlgrey_1.0.0` branch. Try pulling the latest.

Issue #12448, opened 2022-05-03
[bazel] fix `lld` system requirement


timothytrippel:
Since #12263, there is a new system requirement (the LLVM linker `lld`) needed to prevent the build from erroring out with:
```
[...]

PR #25494, opened 2024-12-03
[bazel] Add `rules_rust` `extra_rustc_toolchain_dirs` patch

jwnrt:
From the original commit:

Add build setting to enable specifying other tool locations

On some systems, the CC toolchain discovered by bazel uses a different path for the linker than where the linker (that is specified by the `-fuse-ld=` flag) actually lives on the system.

The issue stems from rules_rust deciding to pass `-fuse-ld=lld` a link arg, rather than just using what was detected by the CC toolchain. This results in the linker not being found, as reported by
https://github.com/lowRISC/opentitan/issues/12448

This is a workaround that enables telling rules_rust where other linker directories might be on your system. It is used by passing `--@rules_rust//:extra_rustc_toolchain_dirs=/path/to/somewhere/else` to `bazel ...` invocations.


jwnrt:
@Razer6 has kindly tested that this works for them.

pamaury (approved):

PR #25495, opened 2024-12-03
[rom_ctrl, rtl] Outstanding request fix

KinzaQamar:
** Fixed the outstanding parameter to 1 in rom_ctrl as only one request
   is accepted by rom_ctrl SRAM.

** Fixed the depth of the fifos from 2 to 1 in rom_ctrl_common_vseq due
   to the reason above.

### DIFF
--- a/hw/ip/rom_ctrl/dv/env/seq_lib/rom_ctrl_common_vseq.sv
+++ b/hw/ip/rom_ctrl/dv/env/seq_lib/rom_ctrl_common_vseq.sv
@@ -154,8 +154,8 @@ class rom_ctrl_common_vseq extends rom_ctrl_base_vseq;
     // their counters. This avoids a problem where we generate a spurious request when the FIFO was
     // actually empty and lots of signals in the design become X. This will let the fifos error
     // signal stuck at X. Zeroing the backing memory avoids that problem.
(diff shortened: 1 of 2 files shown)


rswarbrick (commented):
Thanks for this. I completely agree with the change (and didn't realise that I hadn't filed it myself: sorry!)

But could we tweak the commit message slightly to explain why this is correct? Since the ROM always responds on the next cycle, there is no way to push two items into the FIFO without one being popped in the meantime.

I guess it also makes sense to add an assertion that will quickly tell us if something goes wrong. If the ROM wasn't responding for some reason, the back pressure would get fed back to the TL interface by `a_ready` being false (because e.g. `reqfifo_wready` would be false in `tlul_adapter_sram.sv`, which feeds into `tl_o_int`, which `tl_sram_byte` uses to populate its `tl_o.a_ready`. This gets integrity bits added and then appears as `tl_o` in the adapter). So I *think* we can add an assertion like this? (Check that it works! :-D)
```systemverilog
  // Because ROM always responds in a cycle, it's not possible for it to cause back-pressure on its
  // TL interface. Such back pressure would appear as a false a_ready value.
  `ASSERT(NoBackPressure_A, rom_tl_o.a_ready)
```

KinzaQamar:
Closing it as there is more work required to make a single outstanding request to ROM.

PR #25497, opened 2024-12-03
[entropy_src, dv] Fix incorrect string concatenation

dariuszst:
The SystemVerilog LRM doesn't allow to use a  "+" operator for string literals concatenation. It should be replaced with correct form of concatenation expression (using {} brackets).

### DIFF
--- a/hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv
+++ b/hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv
@@ -917,10 +917,10 @@ class entropy_src_scoreboard extends cip_base_scoreboard#(
       if (main_sm_exp_alert_cond) begin
         if (!fw_ov_insert && !threshold_alert_active && !main_sm_escalates) begin
           if (dut_phase == STARTUP) begin
-            fmt =  "New alert anticpated with >= 2 failing windows." +
-                   "(supercedes count/threshold of %01d/%01d)";
+            fmt = {"New alert anticipated with >= 2 failing windows ",
+                   "(supersedes count/threshold of %01d/%01d)."};
           end else begin
-            fmt = "New alert anticpated! Fail count (%01d) >= threshold (%01d)";
+            fmt = "New alert anticipated! Fail count (%01d) >= threshold (%01d)";
           end
           threshold_alert_active = 1;
           `DV_CHECK_FATAL(ral.recov_alert_sts.es_main_sm_alert.predict(1'b1));


hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv
```diff
@@ -917,8 +917,8 @@ class entropy_src_scoreboard extends cip_base_scoreboard#(
       if (main_sm_exp_alert_cond) begin
         if (!fw_ov_insert && !threshold_alert_active && !main_sm_escalates) begin
           if (dut_phase == STARTUP) begin
-            fmt =  "New alert anticpated with >= 2 failing windows." +
-                   "(supercedes count/threshold of %01d/%01d)";
+		    fmt = {"New alert anticpated with >= 2 failing windows." ,
+                   "(supercedes count/threshold of %01d/%01d)"};
```
rswarbrick, line 921: Thanks for the change: this looks good.

But...
- Please could you indent the second line so that the " characters line up?
- Since we're touching the code, can we also fix the spelling of "anticipated" and "supersedes"?

Thanks!

dariuszst: I've made this corrections in a new commit. Is it OK?

dariuszst: OK. I've squashed it.

rswarbrick:
Annoyingly, it looks like lint has failed because of the commit message. I think there's a missing newline, and the commit message would be nice if it had the [square brackets] stuff at the front like the message on the PR. Could you do that?

hw/ip/entropy_src/dv/env/entropy_src_scoreboard.sv
```diff
@@ -917,10 +917,10 @@ class entropy_src_scoreboard extends cip_base_scoreboard#(
       if (main_sm_exp_alert_cond) begin
         if (!fw_ov_insert && !threshold_alert_active && !main_sm_escalates) begin
           if (dut_phase == STARTUP) begin
-            fmt =  "New alert anticpated with >= 2 failing windows." +
-                   "(supercedes count/threshold of %01d/%01d)";
+            fmt = {"New alert anticipated with >= 2 failing windows." ,
+                   "(supersedes count/threshold of %01d/%01d)"};
```
matutem, line 921:
```suggestion
            fmt = {"New alert anticipated with >= 2 failing windows ",
                   "(supersedes count/threshold of %01d/%01d)."};
```

matutem (changes requested):
Could you also make the commit message more succinct to help the folks reading this in the future? Something like "Fix incorrect string concatenation" seems adequate to describe the change.

I also made a change suggestion.

matutem:
Perhaps you didn't do a "git commit --amend" in your update of Dec 13, since this shows two commits.

matutem (approved):

rswarbrick:
I've just squashed the commits together and rebased onto the head of master. Assuming nothing falls over in CI, this should be good to go.

PR #25501, opened 2024-12-03
Added bazel target for aes_interrupt_encryption

rprakas-gsc:
Updating the aes_testplan.json file with bazel targets for Aes_interrupt_encryption_Test

Adding @jwnrt as reviewer

### DIFF
--- a/hw/top_earlgrey/data/ip/chip_aes_testplan.hjson
+++ b/hw/top_earlgrey/data/ip/chip_aes_testplan.hjson
@@ -112,7 +112,7 @@
       si_stage: SV3
       lc_states: ["PROD"]
       tests: []
-      bazel: []
+      bazel: ["//sw/device/tests:aes_interrupt_encryption_test"]
     }
     {
       name: chip_sw_aes_entropy


hw/top_earlgrey/data/ip/chip_aes_testplan.hjson
```diff
@@ -111,7 +111,8 @@
       stage: V2
       si_stage: SV3
       lc_states: ["PROD"]
-      tests: []
+      tests: ["chip_sw_aes_interrupt_encryption"]
```
engdoreis, line 114: This field is used for to link dv tests listed in the file `hw/top_earlgrey/dv/chip_sim_cfg.hjson`.
So in this case does not apply, can you please drop this change?

engdoreis (changes requested):

hw/top_earlgrey/data/ip/chip_aes_testplan.hjson
rprakas-gsc: Sure.  Just updated the file

hw/top_earlgrey/data/ip/chip_aes_testplan.hjson
```diff
@@ -112,6 +112,7 @@
       si_stage: SV3
       lc_states: ["PROD"]
       tests: []
+      bazel: ["//sw/device/tests/aes_interrupt_encryption_test"]
```
jwnrt, line 115:
```suggestion
      bazel: ["//sw/device/tests:aes_interrupt_encryption_test"]
```

rprakas-gsc:
@engdoreis - I have resolved the conflict, and force pushed the changes

engdoreis:
@rprakas-gsc the PR filed the linting because of the commit message. You have to squash both commits into one and make sure that the massage follows the [style gude](https://opentitan.org/book/doc/contributing/detailed_contribution_guide/index.html?highlight=commit%20message#what-should-commit-messages-look-like).

PR #25502, opened 2024-12-03
[provisioning] enable benchtop SIVAL cert endorsement with Nitrokeys

timothytrippel:
The DICE CA key may be stored on a Nitrokey, which can be used in a benchtop provisioning flow to endorse DICE certificates. This enables such flows by simply setting an envar (`PKCS11_MODULE_PATH`) and plugging in a Nitrokey to the host machine.

PR #25505, opened 2024-12-04
[manuf] Clean up runfile path handling in orchestrator

nmoroze:
This PR updates the provisioning orchestrator to use the `runfiles` library from rules_python to resolve paths to Bazel runfiles. This simplifies the instructions for running the packaged orchestrator. The previous instructions had a few small issues that caused some trouble.

Some hackery was required to preserve the scheme used previously for referring to the main repo vs external repos. We could remove this logic if openocd was the only external reference, but I have a hunch this may be required for supporting provisioning extensions.

### DIFF
--- a/sw/host/provisioning/orchestrator/README.md
+++ b/sw/host/provisioning/orchestrator/README.md
@@ -21,7 +21,8 @@ bazel run \
     --test-unlock-token="0x11111111_11111111_11111111_11111111" \
     --test-exit-token="0x22222222_22222222_22222222_22222222" \
     --fpga=${FPGA_TARGET} \
-    --non-interactive
+    --non-interactive \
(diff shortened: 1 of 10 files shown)


moidx (approved):
Nice change @nmoroze!

I think you still have to update some unittests, but otherwise LGTM.

nmoroze:
Thanks! I think those tests were previously failing, I've rebased now that #25502 is in and they should get cleaned up.

milesdai (approved):

timothytrippel (approved):

milesdai (approved):
LGTM!

timothytrippel:
@nmoroze can you rebase this on latest `earlgrey_1.0.0`?

nmoroze:
@timothytrippel done! Rebased and this is passing tests.

PR #25507, opened 2024-12-04
[manuf,ci] Add an E2E test of the provisioning orchestrator

milesdai:
The orchestrator.py script is intended to be deployed and run as a zip file in an offline environment. This CI check performs and e2e test that the zip file executes correctly.

### DIFF
--- a/apt-requirements.txt
+++ b/apt-requirements.txt
@@ -30,6 +30,7 @@ libftdi1-dev
 libncursesw5
 libpcsclite-dev
 libssl-dev
+libtool
 libudev-dev
 libusb-1.0-0
 lld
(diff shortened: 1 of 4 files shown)


nmoroze (approved):
LGTM!

apt-requirements.txt
```diff
@@ -30,6 +30,7 @@ libftdi1-dev
 libncursesw5
 libpcsclite-dev
 libssl-dev
+libtool
```
milesdai, line 33: #25517 adds the sc_hsm library to the ZIP package which has a non-hermetic system dependency on libtool. This wasn't exercised in CI previously, so the failure is showing up now in this PR.

timothytrippel (approved):

PR #25509, opened 2024-12-04
[aes,dv] Change the type of init_key variable - logic to bit assignments are not allowed in SV

dariuszst:
Sysytem Verilog LRM doesn't allow direct assigning logic type variables to bit one.  Although some tools extend standard (print warnings) in general this code causes compilation errors. To avoid this the type of "init_key" array elements could be probably changed to bit.

### DIFF
--- a/hw/ip/aes/dv/env/seq_lib/aes_wake_up_vseq.sv
+++ b/hw/ip/aes/dv/env/seq_lib/aes_wake_up_vseq.sv
@@ -11,7 +11,7 @@ class aes_wake_up_vseq extends aes_base_vseq;
 
 
   bit [3:0] [31:0]    plain_text       = 128'hDEADBEEFEEDDBBAABAADBEEFDEAFBEAD;
-  logic [255:0]       init_key [2]     = '{256'h0000111122223333444455556666777788889999AAAABBBBCCCCDDDDEEEEFFFF, 256'h0};
+  bit [255:0]         init_key [2]     = '{256'h0000111122223333444455556666777788889999AAAABBBBCCCCDDDDEEEEFFFF, 256'h0};
   bit [3:0] [31:0]    cypher_text;
   bit [3:0] [31:0]    decrypted_text;
   logic [3:0] [31:0]  read_text;


rswarbrick:
This is slightly surprising to me: I would have expected `logic foo; assign foo = 1'b1;` to be ok, and I think the code that you're touching looked like that.

Your commit message sounds like it's talking the about going the other way, so something like `bit bar; assign bar = 1'bx;`. Yep, that's definitely wrong! But I don't think we had that...

danielmlynek:
The issue  here is an unpacked array. 
Logic and bit types can be assigned to each other if they are scalar or vector.
But for array of logic vs array of bits rules for arrays comes into play.
Arrays to be assigned compatilbe must follow the rules "The element types of source and target shall be equivalent" and logic and bit are not equivalent types
module top;
	logic l[2];
	bit b[2];
	initial l= b; //error - arrays not assignment compatible - logic and bit are not equivalelnt
endmodule

rswarbrick (approved):
Oh, I see! I hadn't twigged that the unpacked array was important. Yep, this looks sensible to me.

dariuszst:
> This is failing CI because of the formatting in the commit message. Would you mind tweaking and re-pushing?

Just  to be sure - the empty line after the first one is always required even there is no detailed description and the second line is "Signed-of...." string?
Is there any possibility to do this in more convenient way than git amend/force push through my fork repository?

rswarbrick:
An amend and force-push is the simplest way. You can also avoid some long lines with something like:
```
[aes,dv] Fix type of the init_key class variable

The value that is assigned to the variable is a literal whose unpacked
type is an unpacked array of "bit [255:0]" elements. Although an
expression of type bit [255:0] can safely be considered as an
expression of type logic [255:0], this doesn't lift to unpacked
arrays. Rather than requiring this, use the more specific type.
```

rswarbrick:
Thanks for this. I rather dropped the review here (sorry), so have rebased and force-pushed. Assuming all goes through in CI, I'll come back and merge things.

rswarbrick:
The failing Verible check is not caused by this PR (we just touch a file that was a mess already). I'll push a PR that tidies that up in a few minutes.

rswarbrick:
Verible check should be fixed by #27138.

PR #25510, opened 2024-12-04
[cryptotest] Add P-384 targets to cryptotest.

jadephilipoom:
These were initially excluded because we didn't support P-384 in cryptolib yet, but now that P-384 support is present we should test it. Note that there were some tests with SHA-256 in the original test suite that was commented out, but I removed them: since cryptolib doesn't let you use a hash function with collision resistance weaker than the curve security level these will always fail.

Verified locally that the ECDH and ECDSA targets pass on FPGA.

Follow-up after noticing it was missing while making some changes, see https://github.com/lowRISC/opentitan/pull/25391

### DIFF
--- a/sw/device/tests/crypto/cryptotest/firmware/ecdh.c
+++ b/sw/device/tests/crypto/cryptotest/firmware/ecdh.c
@@ -5,6 +5,7 @@
 #include "sw/device/lib/base/memory.h"
 #include "sw/device/lib/base/status.h"
 #include "sw/device/lib/crypto/impl/ecc/p256.h"
+#include "sw/device/lib/crypto/impl/ecc/p384.h"
 #include "sw/device/lib/crypto/impl/integrity.h"
 #include "sw/device/lib/crypto/impl/keyblob.h"
 #include "sw/device/lib/crypto/include/datatypes.h"
@@ -15,6 +16,7 @@
 #include "sw/device/tests/crypto/cryptotest/json/ecdh_commands.h"
 
 static const int P256_KEY_BYTES = 32;
+static const int P384_KEY_BYTES = 48;
 
 status_t handle_ecdh(ujson_t *uj) {
   // Declare ECDH parameter ujson deserializer types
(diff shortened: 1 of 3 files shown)


sw/device/tests/crypto/cryptotest/firmware/ecdh.c
```diff
@@ -54,14 +59,37 @@ status_t handle_ecdh(ujson_t *uj) {
...
       key_config.key_length = P256_KEY_BYTES;
       shared_key_words = P256_KEY_BYTES / sizeof(uint32_t);
-      p256_masked_scalar_t private_key_masked;
-      memset(private_key_masked.share0, 0, kP256MaskedScalarShareBytes);
-      memcpy(private_key_masked.share0, uj_private_key.d0, kP256ScalarBytes);
-      memset(private_key_masked.share1, 0, kP256MaskedScalarShareBytes);
-      memcpy(private_key_masked.share1, uj_private_key.d1, kP256ScalarBytes);
-      private_key_masked_raw = (uint32_t *)&private_key_masked;
-      private_keyblob_length = sizeof(private_key_masked);
+      memset(private_key_masked_p256.share0, 0, kP256MaskedScalarShareBytes);
```
moidx, line 62: Would it be possible to write a function that takes the same arguments as `memset` but that fills the buffer with random data instead of 0s?

moidx (approved):

sw/device/tests/crypto/cryptotest/firmware/ecdh.c
jadephilipoom: Yes, and in fact we already have one: https://github.com/lowRISC/opentitan/blob/7e2221507aeb1db27b2879ec74d2ad19061b37ad/sw/device/lib/base/hardened_memory.h#L60

It's slightly different than memset because it only works with aligned pointers, but it's good practice to handle secret values in words rather than bytes anyway.

milesdai (approved):
LGTM! Thanks for resolving this!

jadephilipoom:
Backport depends on https://github.com/lowRISC/opentitan/pull/25391 which depends on https://github.com/lowRISC/opentitan/pull/25390

PR #25511, opened 2024-12-04
[aes,dv] Add a cast for the byte type return of getc() function to string variable assignment

dariuszst:
From SV LRM:

```
6.16.3 Getc()
function byte getc(int i);
— str.getc(i) returns the ASCII code of the ith character in str.
— If i < 0 or i >= str.len(), then str.getc(i) returns 0.
```
Thus the type cast should be used to assign it to string variable.

### DIFF
--- a/hw/ip/aes/dv/env/seq_lib/aes_base_vseq.sv
+++ b/hw/ip/aes/dv/env/seq_lib/aes_base_vseq.sv
@@ -503,7 +503,7 @@ class aes_base_vseq extends cip_base_vseq #(
 
   // the index of multi-reg is at the last char of the name
   virtual function int get_multireg_idx(string name);
-    string s = name.getc(name.len - 1);
+    string s = string'(name.getc(name.len - 1));
     return s.atoi();
   endfunction
 


rswarbrick:
I'm not sure this is needed. Isn't this just addressed by the last sentence here?

image

danielmlynek:
Reading the SV LRM  strictly getc return a byte and byte is not assignment compatible with string. 
You are trying to assign byte to whole string - so error is expected.
If you would try to assign it string element like below:
s [0] = name.getc(name.len - 1);
then this would be perfectly legal.
Some tools allow to suppress such error - but it is better to just fix the code.

rswarbrick:
Oh, that makes sense. How silly of me! It's like mixing up `char * ` and `char` in C (`char *my_str = 'a';`)

rswarbrick (approved):

rswarbrick:
Thanks for the change. Annoyingly, it's failing CI because of the commit message being badly formatted. Would you mind tweaking?

rswarbrick:
Thanks for tidying this up. Annoyingly, I didn't review the tidied version (sorry!) and the failing CI log has gone away in the meantime. I've just rebased the PR and force-pushed. I'll take another look once CI has run again.

PR #25513, opened 2024-12-04
[dv/dpi] add an active port to jtagdpi

kwalker27:
To avoid starting up a server when not needed, and spending time making the jtagdpi_tick call unnecessarily, an "active" input port is added. This is similar to what's done in gpiodpi already.

### DIFF
--- a/hw/dv/dpi/jtagdpi/jtagdpi.c
+++ b/hw/dv/dpi/jtagdpi/jtagdpi.c
@@ -176,14 +176,12 @@ void jtagdpi_close(void *ctx_void) {
 void jtagdpi_tick(void *ctx_void, svBit *tck, svBit *tms, svBit *tdi,
                   svBit *trst_n, svBit *srst_n, const svBit tdo) {
   struct jtagdpi_ctx *ctx = (struct jtagdpi_ctx *)ctx_void;
-
-  ctx->tdo = tdo;
-
-  // TODO: Evaluate moving this functionality into a separate thread
-  if (ctx) {
-    update_jtag_signals(ctx);
+  if (!ctx) {
+    return;
   }
 
+  ctx->tdo = tdo;
+  update_jtag_signals(ctx);
   *tdi = ctx->tdi;
   *tms = ctx->tms;
   *tck = ctx->tck;
--- a/hw/dv/dpi/jtagdpi/jtagdpi.sv
+++ b/hw/dv/dpi/jtagdpi/jtagdpi.sv
@@ -8,6 +8,7 @@ module jtagdpi #(
 )(
   input  logic clk_i,
   input  logic rst_ni,
+  input  bit   active,
 
   output logic jtag_tck,
   output logic jtag_tms,
@@ -31,8 +32,10 @@ module jtagdpi #(
 
   chandle ctx;
 
-  initial begin
+  function automatic void initialize();
     int port, assert_srst;
+
+    assert (ctx == null);
 
     // The listening socket port can be customized at runtime
     port = ListenPort;
@@ -43,6 +46,14 @@ module jtagdpi #(
     void'($value$plusargs("jtagdpi_assert_srst=%0d", assert_srst));
 
     ctx = jtagdpi_create(Name, port, assert_srst);
+  endfunction
+
+  initial begin
+    if (active) initialize();
+  end
+
+  always @(posedge active) begin
+    if (ctx == null) initialize();
   end
 
   final begin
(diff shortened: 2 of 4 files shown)


kwalker27:
Hi @rswarbrick I'm curious for your thoughts on the following:

One reason for adding this is that I'd like to make a `tcp_server` failure to open its listening socket be a fatal error instead of just printing and continuing. If there were never conflicts for the socket, I can see it being nice to just always have the server there in case one wants to connect a client. However, in the context of shared machines with multiple simulation jobs running, port conflicts can fly under the radar and lead to confusion if somebody _does_ fire up a client. In our testbenches, we'll wire this `active` input port to only be asserted if a certain plusarg is present at sim-time. This, in combination with the `jtagdpi_port` plusarg lets us fire up multiple jobs with unique ports and detect conflicts.

I'm prepping the `tcp_server` socket error handling change for a subsequent PR but I expect there would need to be discussion about whether there will be fallout from such a change. I don't see any active `jtagdpi` instances in this repo, but the `dmidpi` also uses `tcp_server` and unconditionally tries to start its server using a hard-coded port number - so CI sim jobs that land on the same machine (for example) might be affected by such a change in policy.

hw/dv/dpi/jtagdpi/jtagdpi.sv
```diff
@@ -43,6 +44,14 @@ module jtagdpi #(
...
 
     ctx = jtagdpi_create(Name, port, assert_srst);
+  endfunction
+
+  initial begin
+    if (active) initialize();
+  end
+
+  always @(posedge active) begin
+    if (ctx == null) initialize();
```
rswarbrick, line 54: This looks sensible but maybe we should also add an assertion in `initialize` that checks `ctx` is null when we are called.

hw/dv/dpi/jtagdpi/jtagdpi.c
```diff
@@ -176,14 +176,12 @@ void jtagdpi_close(void *ctx_void) {
 void jtagdpi_tick(void *ctx_void, svBit *tck, svBit *tms, svBit *tdi,
                   svBit *trst_n, svBit *srst_n, const svBit tdo) {
   struct jtagdpi_ctx *ctx = (struct jtagdpi_ctx *)ctx_void;
-
-  ctx->tdo = tdo;
```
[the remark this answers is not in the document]
kwalker27: Yeah, that was odd :)

rswarbrick:
Thinking about `jtagdpi` specifically, I wonder whether there is actually any use of the module in OpenTitan at the moment.
```
nix:rjs@sole:~/work/opentitan$ git grep 'jtagdpi[^a-z_]' -- '*.sv'
hw/dv/dpi/jtagdpi/jtagdpi.sv:module jtagdpi #(
hw/top_earlgrey/dv/verilator/chip_sim_tb.sv:  // jtagdpi u_jtagdpi (
hw/top_englishbreakfast/rtl/chip_englishbreakfast_verilator.sv:  // jtagdpi u_jtagdpi (
```

Do you guys have a use-case in mind for this inside OT? If not, maybe the best thing is to drop the module entirely! :-)

(@vogelpi: I know that you've touched this in EnglishBreakfast. Do you know anything about the topic?)

kwalker27:
> Thinking about `jtagdpi` specifically, I wonder whether there is actually any use of the module in OpenTitan at the moment.
> 
> ```
> nix:rjs@sole:~/work/opentitan$ git grep 'jtagdpi[^a-z_]' -- '*.sv'
> hw/dv/dpi/jtagdpi/jtagdpi.sv:module jtagdpi #(
> hw/top_earlgrey/dv/verilator/chip_sim_tb.sv:  // jtagdpi u_jtagdpi (
> hw/top_englishbreakfast/rtl/chip_englishbreakfast_verilator.sv:  // jtagdpi u_jtagdpi (
> ```

Right, I only spotted these two that are commented out, but I added the port there anyway.

> Do you guys have a use-case in mind for this inside OT? If not, maybe the best thing is to drop the module entirely! :-)

Inside OT... not really? We do instantiate it in several TBs to drive the JTAG pins for interactive debug activities.

kwalker27:
Rats, that merge was not intended, just fat-fingered the website. Will fix my branch.

hw/dv/dpi/jtagdpi/jtagdpi.sv
kwalker27: Done

andreaskurth:
Re @rswarbrick's question:
> Do you guys have a use-case in mind for this inside OT? If not, maybe the best thing is to drop the module entirely! :-)
>
> (@vogelpi: I know that you've touched this in EnglishBreakfast. Do you know anything about the topic?)

Given @kwalker27's response:
> Inside OT... not really? We do instantiate it in several TBs to drive the JTAG pins for interactive debug activities.

I think we should keep it.

rswarbrick (approved):
This looks good to me!

PR #25524, opened 2024-12-05
[formal,flash_ctrl] Add connectivity check for prim_flash recov_alert

### LINKED ISSUE #25523: [formal,flash_ctrl] Add connectivity check for prim_flash recov_alert
### Description

Verify the prim_flash recov_alert_o is connected to the flash_ctrl alert_senders[4] input. The recov_alert_o is not connected in the open source model so there is no simple way to check it will cause an outgoing alert. A connectivity check will confirm it should. This issue was flagged in the flash_ctrl tape-out review.

matutem:
Fixes #25523

### DIFF
--- a/hw/top_earlgrey/formal/conn_csvs/flash_ctrl.csv
+++ b/hw/top_earlgrey/formal/conn_csvs/flash_ctrl.csv
@@ -1,0 +1,11 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Run these checks with:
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
Doesn't this also need to be added to `hw/top_earlgrey/formal/chip_conn_cfg.hjson`?

vogelpi:
I agree with @rswarbrick that this new test should probably be added to the list, too.

I've tried to run this test but it currently fails for another reason. I am investigating this with @matutem offline and will eventually report back here.

vogelpi (approved):
Thanks @matutem , this looks good to me!

I've edited this PR to:
1. Add the newly added csv to the list of csvs as pointed out by Rupert.
2. Split the source signal from the source block in the newly added csv.

If I revert the commit adding the pull-up for the POR (see #25567), the connectivity test can be run successfully. I would recommend merging this PR and take care of the pull-up issue separately.

PR #25531, opened 2024-12-06
[opentitantool] Include updated firmware for HyperDebug

jesultra:
This update includes several improvements:

* New SPI/I2C features to perform TPM operations in a single USB roundtrip.

* Vastly improve efficiency of UART data forwarded via USB.

* Fix to bit-banging, which could previously incorrectly complain about "ongoing operation", if the most recent bit-banging had ended less than one clock-tick ago.  (This clock could be arbitrarily slow.  The problem showed up when bit-banging simulated key presses with a 20ms clock, and the next operation was attempted initiated less than 20ms after the previous one had applied its last sample.)

* ADC calibration did not entirely adhere to STM32 timing requirements, resulting in occasional crash during "reinit" (transport init).

Relevant CLs:
87399ae831 board/hyperdebug: Reduce USB fragmentation in CMSIS-DAP
261e3ec718 chip/stm32: L4/L5 UART driver makes use of RX timeout
ab72b9eaf8 chip/stm32: USB stream driver delays outbound USB packet until flush
19e0f86cd3 chip/stm32: Remove unused fields
3006f5bffb common: Introduce buffered mode and queue_flush()
46b9a8797e board/hyperdebug: Implement tailored logic for TPM I2C devices
5cbea1dba5 board/hyperdebug: Implement tailored logic for TPM SPI devices
8115337458 board/hyperdebug: Route all SPI requests through board code
d2a6c10ad5 board/hyperdebug: Support bit-banging in quick succession
7760f99e7d board/hyperdebug: ADC calibration robustness
513b83b195 chip/stm32: Correct ADC initialization timing
3024d3eb68 chip/stm32: Larger console UART buffer for STM32L5

### DIFF
--- a/third_party/hyperdebug/repos.bzl
+++ b/third_party/hyperdebug/repos.bzl
@@ -7,7 +7,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 def hyperdebug_repos():
     http_archive(
         name = "hyperdebug_firmware",
-        urls = ["https://github.com/lowRISC/hyperdebug-firmware/releases/download/20240621_01/hyperdebug-firmware.tar.gz"],
-        sha256 = "649a8cd6d183bc3fb286ea5895c752cfec3aa29b9990f44bb9e7621c0414c7de",
+        urls = ["https://github.com/lowRISC/hyperdebug-firmware/releases/download/20241206_01/hyperdebug-firmware.tar.gz"],
+        sha256 = "13612708d0d71e340babff07d14d67cd9d51f2eb3c370757c3aa905aeab13d1b",
         build_file = "@lowrisc_opentitan//third_party/hyperdebug:BUILD",
     )


jesultra:
A thought just occurred to me:  A little while ago, it was pointed out that opentitantool had not been kept in sync between our branches, and in particular that the version of HyperDebug firmware differered, which resulted in frequent upgrading/downgrading of firmware, as different CI runs were scheduled on the same physical testbeds.  Since the firmware version was updated in same PRs as chances in opentitantool Rust code, and some of those had merge conflicts,  as a stop-gap measure we modified opentitantool (in all branches) to never downgrade HyperDebug firmware.

This PR brings a new version of HyperDebug firmware, and even though I am not expecting issues, I cannot help thinking, that if this firmware is broken in some way, causing widespread test failures, then the damage is done, as some testbeds will have had their HyperDebug firmware updated, and subsequent test runs will keep using the newer firmware, even if this PR never merges.

We should probably either ensure that the HyperDebug firmware is kept in sync between branches, and then re-enable the automatic downgrading, in order to guarantee that damage from future PRs like this one are limited to its own CI run.

jwnrt (approved):

jesultra:
I have discovered a flaw in the way that the HyperDebug firmware waits for Google's proprietary "ready pulse" (on an additional pin).  This would not influence the ability of this new firmware to poll for SPI TPM signalling being ready by the value 0x01 according to the standard.

However, I dread having to first cherry-pink this to two release branches, and then soon cherry-pinking another HyperDebug firmware update.  So maybe I should revert this one, and then create a new PR with fixed firmware, which then becomes the only one to be cherry-picked.  (If not revering, I could also manually cherry-pick the second PR, resolving the conflict arising from the branches skipping a version.)

nbdd0121:
Given that the current HyperDebug update logic will not downgrade by default, we can also just not cherry-pick at all? Given some time all FPGAs will have their HyperDebug firmeware updated eventually.

Issue #23038, opened 2024-05-09
[FPGA] MASK_ROM/ROM_EXT exec_environments require info flash page initialization


moidx:
## Issue 

Currently, the flash_ctrl and keymgr may raise alerts due to the initial info flash configuration when booting with the `mask_rom` and `rom_ext` targets.

* The keymgr requires the Owner and Secret owner pages to be initialized.
* Ownership transfer requires the owner block to be initialized.

## Options

1. Modify all environments using the real `rom_ext` so that there is an initial bootstrap operation with the personalization image to initialize all required info partitions. Applicable to all FPGA configurations.
2. Register a weak functions in the `rom_ext` to initialize the owner partition for silicon targets already deployed in various environments. 

CC: @cfrantz @timothytrippel  @jwnrt

timothytrippel:
I think it may be more efficient to go with option 2, since it could be much lighter weight than running a full personalization on every boot, since we really just need to program those two pages.

jwnrt:
Related https://github.com/lowRISC/opentitan/issues/21225

timothytrippel:
yes, another option is splicing, the only downside with splicing is it will require a lot more infrastructure than a weak function in the ROM_EXT

Issue #21225, opened 2024-02-06
[fpga] Create splicing capability to init flash INFO pages


a-will:
The FPGA has no backdoor interface to supply initial values for the flash INFO pages. Because the creator/owner seeds can be in the INFO pages, we can't test from some OTP states without getting flash_ctrl errors.

We may want to create the ability to splice the flash INFO pages, so a test can start from a fully provisioned state, avoiding a requirement to go through a provisioning process and life cycle transitions first.

From observations by @jdonjdon:
https://github.com/lowRISC/opentitan/issues/20961#issuecomment-1930219405_

PR #25533, opened 2024-12-07
[fpga] Add prim features and tool flow to dump the flash info array cell sites

a-will:
### Collect MMI data into one file.
    
Reduce BUILD file boilerplate by placing all memory map info in a single file. Use a separate Processor element for each map, so the InstData attribute can be used as a key to identify the memory to be updated.  This construction also permits multiple memories to be spliced in a single call to updatemem, though that is not implemented in this commit.

Adjust MMI dump to use a regex for mem type and a dictionary for each memory's parameters. For the mem type, some memories in the FPGA can consist of both RAMB36 and RAMB18 types, depending on the width of the array. The regex allows more flexible memory cell detection.

Rev up the bitstream manifest schema to v3 so the memories contained in the MMI file are described. Add a function to collect MMI data from v2 cache entries and rewrite in the expected v3 format, with the MMI data in a single file.

### Add specialized ram_1p prim for Xilinx FPGAs
    
Add a specialized ram_1p prim for the prim_xilinx library. This prim adds some prim_xilinx-specific configuration ability to enable selecting particular layouts for embedded memories. The intention is to have the top-level specify the configuration via a special prim_xilinx_pkg, which is a virtual core. The prim instance will do a lookup to decide how to split the memories.

This feature is intended for use with memories that are too wide to fit in updatemem's capabilities (where splicing is required). The flash info pages are one such type, since updatemem can only handle up to 64-bit wide items, and the info page array is 76 bits wide. Synthesis chooses the most efficient layout of the memories, but the layout is awkward for handling splices.

Force earlgrey's flash info data and metadata into separate arrays for the FPGA.

### Dump flash info maps to the MMI file.
    
For earlgrey, export the flash info memory maps to the MMI file. Add knowledge of these maps existing to the bitstream cache entries. The
flash info arrays are identified by a key in this format:
    
        `flash<FlashBank>_info<InfoType>_<data|intg>`

This lays groundwork for a future PR where we can splice these memories.

This was tested by hand by manually splicing the uart_smoketest binary into flash bank 0, info partition 0 and running a slightly modified `//sw/device/tests:flash_ctrl_info_access_lc_prod_personalized_fpga_cw310_sival` (to prevent overwriting the data). Without the integrity bits in, the flash control read would get operation errors. With the integrity bits also spliced in, the reads would succeed, and we'd get comparison failures showing data from the MEM file.

This PR likely will supersede https://github.com/lowRISC/opentitan/pull/25363

Note: _The API for the prim_xilinx_pkg is known to be a terrible interface,_ but it's left this way to avoid alternatives that might change RTL for the ASIC's tree in earlgrey_1.0.0. An example of a more refined approach that changes that RTL might be to expose a top-level parameter for every major memory, so an ID-based lookup could be used to control prim library-specific configuration. That would yield instance-level control as needed.

Partially addresses #21225 and #23038


a-will:
If you were following along #25363 before, only the last commit is new.

a-will:
The SV was bugged anyway, but something seems a little off with our dvsim tool flow. prim_xilinx_ultrascale RTL files shouldn't be part of the file list for any of its top modules' build graphs.

a-will:
An alternative would've been to leave the memory as-is and require the bitstream cache to explicitly represent partitioning of memories (in practice, of splice-capable memories that are wider than 64 bits), then have tools that break images into the partitions and call updatemem on each of the partitioned MEM files.

If that is preferred for reviewers but the approach in this PR isn't _too_ objectionable, I would ask that we target that for the next round and let this one into the repo. We can handle a change to the bitstream cache schema, as is done here, and there is work that continues to be blocked while we lack some way to fix up the flash info page state.

a-will:
The failing CW340 CI flows appear to be an ongoing infrastructure failure, not a result of this PR.

moidx (approved):
Thanks @a-will,  great job simplifying the memory mapping infrastructure for `updatemem`. 

@andreaskurth Can you or someone on your end take a look at the FPGA prim changes included in this PR? 

Thanks

moidx:
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_xilinx_pkg.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_xilinx_ram_1p.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_ram_1p.sv

doc/contributing/fpga/debugging_with_ila.md
```diff
@@ -237,7 +237,7 @@ The path to the synthesis log file is usually `bazel-out/k8-fastbuild/bin/hw/bit
 
 Once Vivado has successfully generated a bitstream, locate its directory with `dirname $(./bazelisk.sh outquery-all //hw/bitstream:rom --define bitstream=vivado)` (it will usually be `bazel-out/k8-fastbuild/bin/hw/bitstream/vivado`).
 Append `/build.fpga_cw310/synth-vivado` to that path.
-In the resulting directory, you should find `otp.mmi`, `rom.mmi`, and `lowrisc_systems_chip_earlgrey_cw310_0.1.bit`.
+In the resulting directory, you should find `memories.mmi`, and `lowrisc_systems_chip_earlgrey_cw310_0.1.bit`.
```
andreaskurth, line 240: Nit :-)
```suggestion
In the resulting directory, you should find `memories.mmi` and `lowrisc_systems_chip_earlgrey_cw310_0.1.bit`.
```

hw/top_earlgrey/earlgrey_xilinx_prim_pkg.core
```diff
@@ -0,0 +1,19 @@
+CAPI=2:
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+name: "lowrisc:prim_xilinx:earlgrey_xilinx_prim_pkg"
+description: "Single port RAM"
+virtual:
+  - lowrisc:prim_xilinx:prim_xilinx_pkg
```
andreaskurth, line 9: This requires FuseSoC v2.4 to work properly, no?

hw/top_earlgrey/util/vivado_hook_write_bitstream_pre.tcl
```diff
@@ -200,43 +232,92 @@ dict set memInfo otp brams $otp_brams
...
+dict set memInfo flash0_info0_data fake_word_width 0
+dict set memInfo flash0_info0_data addr_end_multiplier 1
+dict set memInfo flash0_info0_data schema "MemoryArray"
+
+set flash_info_brams [split [get_cells -hierarchical -filter " PRIMITIVE_TYPE =~ ${bram_regex} && NAME =~ *u_flash_ctrl*gen_prim_flash_banks[0]*gen_info_types[1].u_info_mem*gen_xpm.gen_split[0].*"] " "]
+dict set memInfo flash0_info1_data brams $flash_info_brams
+dict set memInfo flash0_info1_data mem_type_regex $mem_type_regex
+dict set memInfo flash0_info1_data fake_word_width 0
+dict set memInfo flash0_info1_data addr_end_multiplier 1
+dict set memInfo flash0_info1_data schema "MemoryArray"
```
andreaskurth, line 250: Could this and the following code be deduplicated by using Tcl's `eval`?

hw/top_earlgrey/util/vivado_hook_write_bitstream_pre.tcl
```diff
@@ -201,6 +201,43 @@ dict set memInfo otp mem_type_regex $mem_type_regex
...
+dict set memInfo flash0_info0 brams $flash_info_brams
+dict set memInfo flash0_info0 mem_type_regex $mem_type_regex
+dict set memInfo flash0_info0 fake_word_width 0
+dict set memInfo flash0_info0 addr_end_multiplier 1
+
+set flash_info_brams [split [get_cells -hierarchical -filter " PRIMITIVE_TYPE =~ ${bram_regex} && NAME =~ *u_flash_ctrl*gen_prim_flash_banks[0]*gen_info_types[1].u_info_mem*"] " "]
+dict set memInfo flash0_info1 brams $flash_info_brams
+dict set memInfo flash0_info1 mem_type_regex $mem_type_regex
+dict set memInfo flash0_info1 fake_word_width 0
+dict set memInfo flash0_info1 addr_end_multiplier 1
```
andreaskurth: Also here: possibility for deduplication using `eval`?

hw/top_earlgrey/data/synth.xdc
```diff
@@ -0,0 +1,14 @@
+## Copyright lowRISC contributors (OpenTitan project).
+## Licensed under the Apache License, Version 2.0, see LICENSE for details.
+## SPDX-License-Identifier: Apache-2.0
+
+## Synthesis constraints
+
+# Prevent Vivado from combining these BRAMs, so the memories can be readily
+# spliced with updatemem.
+set_property KEEP_HIERARCHY TRUE [get_cells "top_earlgrey/u_flash_ctrl/u_eflash/u_flash/gen_generic.u_impl_generic/gen_prim_flash_banks[0].u_prim_flash_bank/gen_info_types[0].u_info_mem"]
+set_property KEEP_HIERARCHY TRUE [get_cells "top_earlgrey/u_flash_ctrl/u_eflash/u_flash/gen_generic.u_impl_generic/gen_prim_flash_banks[0].u_prim_flash_bank/gen_info_types[1].u_info_mem"]
```
andreaskurth: Another opportunity to deduplicate (probably not even using `eval`)?

andreaskurth (approved):
This looks great to me overall, thanks @a-will! One question and a few suggestions from my side

andreaskurth:
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_xilinx_pkg.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx/rtl/prim_xilinx_ram_1p.sv
CHANGE AUTHORIZED: hw/ip/prim_xilinx_ultrascale/rtl/prim_xilinx_ultrascale_ram_1p.sv

These are new Xilinx-specific prims, which should be functionally equivalent to the existing implementation from an I/O perspective.

hw/top_earlgrey/earlgrey_xilinx_prim_pkg.core
a-will: No, we use virtual cores with my fusesoc fix already. I added the satsolver instructions to our fork at the same time as upstream.

If we use upstream fusesoc, _then_ 2.4 is the first release that includes my commits.

hw/top_earlgrey/util/vivado_hook_write_bitstream_pre.tcl
a-will: I went with `apply` instead, but yes.

hw/top_earlgrey/data/synth.xdc
a-will: This actually gets added and removed again because of the changing ideas. If you'd like, I can go back in the history and get rid of its appearance altogether.

a-will:
The AscentLint job is failing, but the report is all green.

PR #25535, opened 2024-12-08
[ci] Create GitHub groups in log outputs for FPGA tests

pamaury:
The log are quite long and GitHub provides a feature where the output group can split lines into groups. This is achieved by using the --run_under command of bazel test. This allows to run a script that then invokes the test runner. This only thing this script does is print the grouping commands.


jwnrt (approved):
I think this is much easier to navigate, thanks

engdoreis (approved):
I also like it. Thanks @pamaury

PR #24276, opened 2024-08-12
[rv_dm,dv] Exclude StFlush state coverage in a tlul_lc_gate instance

rswarbrick:
Detailed notes in a comment above the exclusion.

PR #25552, opened 2024-12-09
[sram_ctrl,dv] Add exclusion for StFlush

nasahlpa:
Similar to lowRISC/opentitan#24276, this commit excludes state transitions from and to the `StFlush` state of the `tlul_lc_gate` FSM. As `flush_req_i` of this module is tied to 0, reaching this state is something we are not planning to see.


rswarbrick (approved):
This looks sensible to me.

Issue #24363, opened 2024-08-20
[hmac,dv] MSG_LENGTH registers write


martin-velay:
### Description

MSG_LENGTH registers should be writable only when CFG.sha_en=0, check also that they cannot be written when CFG.sha_en=1

martin-velay:
After being tested, this turned out to be a specification issue (in reality we are adapting the specification to match the actual RTL).
Finally, the spec will be updated to say that the SW should update the digest and the message length when the core is in IDLE otherwise the programmed value can cause unpredictable results. Which means that nothing should be further tested in DV.

martin-velay:
Can be closed as https://github.com/lowRISC/opentitan/pull/25557 has been merged

PR #25554, opened 2024-12-09
[hmac,dv] Write to msg_length when sha_en

martin-velay:
- this PR is linked to issue #24363
- it aims to test that write access to the message length registers is seemless when SHA core is enabled.
- add also a covergroup to track this from the fcov


martin-velay:
Converted back into draft as it reveals a bug:
The msg_length registers can be written when SHA core is enabled. This is not what the current spec says. After a discussion with @andreaskurth, we can consider this a specification bug and only modify the doc instead of the RTL. The specification will say that these registers can be updated only when the HMAC is in IDLE status. Another PR will be created soon and this PR will be modified to test this new behaviour.

martin-velay:
Cancelled as not a DV hole anymore. This was more a spec bug, which has been covered here:
https://github.com/lowRISC/opentitan/pull/25557

PR #25555, opened 2024-12-09
[dif] Fix dif_<ip>_irq_get_type to handle bad arguments

pamaury:
The functino performs an incorrect check on the irq number. In particular, this number could be negative or above max_irq+1. If this happens, the function will perform an out-of-bound array access.

One particular instance where this could happen is in the isr testutils where it can be called with an incorrect argument.

### DIFF
--- a/sw/device/lib/dif/autogen/dif_adc_ctrl_autogen.c
+++ b/sw/device/lib/dif/autogen/dif_adc_ctrl_autogen.c
@@ -72,8 +72,8 @@ OT_WARN_UNUSED_RESULT
 dif_result_t dif_adc_ctrl_irq_get_type(const dif_adc_ctrl_t *adc_ctrl,
                                        dif_adc_ctrl_irq_t irq,
                                        dif_irq_type_t *type) {
-  if (adc_ctrl == NULL || type == NULL ||
-      irq == kDifAdcCtrlIrqMatchPending + 1) {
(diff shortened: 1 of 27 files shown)


sw/device/lib/dif/autogen/dif_adc_ctrl_autogen.c
```diff
@@ -72,8 +72,8 @@ OT_WARN_UNUSED_RESULT
 dif_result_t dif_adc_ctrl_irq_get_type(const dif_adc_ctrl_t *adc_ctrl,
                                        dif_adc_ctrl_irq_t irq,
                                        dif_irq_type_t *type) {
-  if (adc_ctrl == NULL || type == NULL ||
-      irq == kDifAdcCtrlIrqMatchPending + 1) {
+  if (adc_ctrl == NULL || type == NULL || irq < 0 ||
+      irq >= kDifAdcCtrlIrqMatchPending + 1) {
```
jwnrt, line 76: Very nitty nit: can we use `irq > kDifAdcCtrlIrqMatchPending` instead?

jwnrt (approved):
LGTM, thanks

sw/device/lib/dif/autogen/dif_adc_ctrl_autogen.c
pamaury: Sure

PR #25557, opened 2024-12-09
[hmac,doc] S&R reg changes when IDLE

martin-velay:
- a bug has been found via PR #25554, but this has been considered as a specification bug. This PR fixes this issue. Now the specification says that DIGEST and MSG_LENGTH registers can be updated only when HMAC is in IDLE state (instead of when SHA core is disabled).
- linked to https://github.com/lowRISC/opentitan/issues/24363

### DIFF
--- a/hw/ip/hmac/data/hmac.hjson
+++ b/hw/ip/hmac/data/hmac.hjson
@@ -195,7 +195,7 @@
           desc: '''SHA-2 enable.
 
                  If 0, the SHA engine will not initiate compression, this is used to stop operation of the SHA-2 engine until configuration has been done.
-                 When the SHA-2 engine is disabled the digest is cleared, and the digest can be written to from SW which enables restoring context (to support context switching).'''
+                 When the SHA-2 engine is disabled the digest is cleared.'''
           tags: [// don't enable hmac and sha data paths - we will do that in functional tests
                  "excl:CsrNonInitTests:CsrExclWrite"]
         }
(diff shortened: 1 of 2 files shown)


andrea-caforio (approved):

martin-velay:
I have tested to write while the HMAC is not IDLE status. And I have found some period where it's actually possible:
image
Here we can see that a write is occurring while we are in IDLE state, but the value is propagated to the internal logic and used.

So, the specification should be reworded to say that the SW should update the message length and the digest only when the core is in IDLE state. Any changes outside of this window can lead to an unexpected state.
I'll update this PR consequently.

Do you approve this planned change @andreaskurth ?

martin-velay:
I have added the CherryPick:earlgrey_1.0.0 tag as this is also valid for the taped-out chip.

hw/ip/hmac/data/hmac.hjson
```diff
@@ -468,7 +470,7 @@
               For SHA-2 384, {DIGEST12-DIGEST15} are truncated; they are irrelevant and should not be read out.
 
               The digest gets cleared when `CFG.sha_en` transitions from 1 to 0.
-              When `CFG.sha_en` is 0, these registers can be written to by software.
+              When `STATUS.hmac_idle` is 1, these registers can be written to by software. Outside of this window, writes can cause unpredictable results.
```
andreaskurth, line 473: Slight rephrase but otherwise LGTM (also for the other two changes below)
```suggestion
              When `STATUS.hmac_idle` is 1, these registers may be written to by software. Outside of this window, writes can cause unpredictable behavior.
```

andreaskurth (approved):
LGTM with a slight rephrase suggestion. Thx @martin-velay.

Also I think the two commits are best squashed together

andreaskurth:

This fixes the spec; the implementation doesn't change.

vogelpi:

This fixes the spec; the implementation doesn't change.

hw/ip/hmac/data/hmac.hjson
vogelpi: And please use 1 sentence per line.

martin-velay: Thanks for your comments, all are in my latest commit.

PR #25560, opened 2024-12-09
[pwm] RTL modifications for specification compliance and predictable behavior

alees24:
This PR addresses a number of issues found within the PWM RTL as part of the D3 review process. Since the changes have ended up becoming pretty extensive, the PR is structured as a number of commits. Each of these, except perhaps the final one, should be easy to understand in isolation since it is either (i) small and self-contained or (ii) not a functional change.

Since the last involves significant changes to the blink mode and particularly heartbeat logic it is more involved.
Please see the commit messages for a breakdown of what each commit does.

Issues with the current PWM implementation:

 - phase counter configuration is accepted immediately; software could probably tolerate either immediate or deferred acceptance, but this PR changes the RTL to match the specification/documentation, allowing the configuration to be changed at any point and the phase counting to be disabled at any time. New configuration settings are accepted into the core logic only upon enabling the counter.

 - phase counter is not fully reset when the configuration was changed.

 - channel enable only gates the output on/off and does not start the internal logic from a defined state.

 - heartbeat logic is not reset properly when the configuration was changed, so to achieve synchronized heartbeat signaling (eg. tri-color LED operation) it is necessary first to configure the pwm channels for blinking operation and then quickly set the `htbt_en` bits to ensure that they start at the same time (not at the 5us clock-cycle level, but close enough for LEDs).

 - changing the `pwm_en` or `invert` register impacts the other channels, causing a reset of the blinking logic and a partial reset (loss of predictable timing) of heartbeat logic.

 - the duty cycle within the channel is updated at the end of the pulse cycle without any regard for the phase delay. A given output pulse may thus be formed from two different duty cycles.

 - there is a lot of combinational logic on the output of the PWM block which means that it is likely to suffer from glitches; this could be a problem for any application that is edge-sensitive or relies upon specific timing rather than just the overall average signal level.

### DIFF
--- a/hw/ip/pwm/rtl/pwm_chan.sv
+++ b/hw/ip/pwm/rtl/pwm_chan.sv
@@ -39,6 +39,29 @@ module pwm_chan #(
   assign blink_mode = pwm_en_i & blink_en_i;
   assign htbt_mode = pwm_en_i & blink_en_i & htbt_en_i;
 
+  // When there is a non-zero phase delay we cannot use the `cycle_end_i` signal directly
+  // to switch to a new duty cycle because we could be asserting the pwm output at that
+  // instant, so we shall need to delay this signal.
+  logic dc_update_pending;
+  logic set_dc_update;
+  logic clr_dc_update;
+  logic update;
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      dc_update_pending <= 1'b0;
+    end else if (clr_chan_cntr_i) begin
+      dc_update_pending <= 1'b0;
+    end else if (pwm_en_i) begin
+      dc_update_pending <= (dc_update_pending | set_dc_update) & ~clr_dc_update;
+    end
+  end
+  // We delay the update until the _next_ value of the phase counter passes the programmed phase
+  // delay. The new duty cycle could be zero, so we must know before the crossing point.
+  assign set_dc_update = (cycle_end_i & |phase_delay_masked);
+  assign clr_dc_update = (phase_ctr_next_i >= phase_delay_masked);
+  assign update = pwm_en_i & (|phase_delay_masked ? (dc_update_pending & clr_dc_update)
+                                                  : cycle_end_i);
+
   logic [CntDw-1:0] duty_cycle_actual;
   logic [CntDw-1:0] on_phase;
   logic [CntDw-1:0] off_phase;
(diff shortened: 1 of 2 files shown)


hw/ip/pwm/rtl/pwm_chan.sv
```diff
@@ -50,81 +73,56 @@ module pwm_chan #(
...
-        blink_ctr_q <= blink_ctr_d;
-      end
+      blink_ctr_q <= blink_ctr_d;
+      blink_y_phase_q <= blink_y_phase_d;
     end
   end
 
-  assign duty_cycle_blink = (blink_en_i && !htbt_en_i && (blink_ctr_q > blink_param_x_i)) ?
-                            duty_cycle_b_i : duty_cycle_a_i;
+  // Duty cycle for blinking toggles between 'a' and 'b', otherwise always 'a'.
```
rswarbrick, line 115: Maybe this comment could be expanded slightly. Something like "This is the duty cycle, taking blinking into account. If that is enabled, the duty cycle toggles between 'a' and 'b'. If not, the duty cycle is always 'a'." ?

hw/ip/pwm/rtl/pwm_chan.sv
```diff
@@ -136,69 +134,68 @@ module pwm_chan #(
   assign pattern_repeat = (duty_cycle_a_i == duty_cycle_b_i);
   assign neg_htbt = !pos_htbt & !pattern_repeat;
 
+  // We support either duty cycle 'a' or 'b' being the greater.
```
rswarbrick, line 137: Nit: I took a few seconds to understand this. Maybe "We support duty cycles where either 'a' or 'b' can be greater." ?

(I think my initial reading of your text was "... b being the greater of x, y, z" or something)

rswarbrick: Also, maybe this description should be above `pos_htbt`?

hw/ip/pwm/rtl/pwm_chan.sv
```diff
@@ -136,69 +134,68 @@ module pwm_chan #(
...
+    if (pattern_repeat) begin
+      {dc_wrap, dc_htbt_d} = {1'b0, duty_cycle_a_i};
+    end else if (htbt_falling) begin
+      {dc_wrap, dc_htbt_d} = {dc_wrap_q, dc_htbt_q} - (CntExtDw)'(blink_param_y_i) - 1'b1;
+    end else begin
+      {dc_wrap, dc_htbt_d} = {dc_wrap_q, dc_htbt_q} + (CntExtDw)'(blink_param_y_i) + 1'b1;
     end
   end
 
+  // Remember for 'x' pulse cycles whether an over/underflow occurred (dc_wrap was asserted).
```
rswarbrick, line 171: Maybe rephrase as "For 'x' pulse cycles, remember whether..." ? (I initially read this with 'x' being the number of pulse cycles)

hw/ip/pwm/rtl/pwm_chan.sv
```diff
@@ -26,7 +26,7 @@ module pwm_chan #(
   input                clr_chan_cntr_i,
   input [DcResnDw-1:0] dc_resn_i,
 
-  output logic pwm_o
+  output logic         pwm_o
```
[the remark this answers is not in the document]
alees24: Perhaps it isn't the only thing we care about; I could imagine a case where the PWM outputs are driven as multiple groups, say two independent tri-color LEDs, for different uses and driven by different bits of software.

This was one of the major motivations for changing the behavior of 'clr_chan_ctr_i'. Software workarounds can be built for the other issues, but not to prevent changes to one groups of signals interfering with another group that's already active.

hw/ip/pwm/rtl/pwm_core.sv
```diff
@@ -64,13 +68,28 @@ module pwm_core #(
...
+  end
+
+  assign cntr_en = reg2hw.cfg.cntr_en.qe ? reg2hw.cfg.cntr_en.q : cntr_en_q;
+  assign dc_resn = clr_phase_cntr        ? reg2hw.cfg.dc_resn.q : dc_resn_q;
+  assign clk_div = clr_phase_cntr        ? reg2hw.cfg.clk_div.q : clk_div_q;
 
   assign beat_ctr_d = (clr_phase_cntr) ? '0 :
                       (beat_ctr_q == clk_div) ? '0 : (beat_ctr_q + 1'b1);
-  assign beat_ctr_en = clr_phase_cntr | cntr_en;
+  assign beat_ctr_en = cntr_en;
```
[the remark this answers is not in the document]
alees24: Ok

hw/ip/pwm/rtl/pwm_core.sv
```diff
@@ -81,10 +100,10 @@ module pwm_core #(
     end
   end
 
-  // Only update phase_ctr at the end of each beat
-  // Exception: allow reset to zero whenever not enabled
+  // Only update phase_ctr at the end of each beat.
```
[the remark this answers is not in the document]
alees24: Yes, that comment was always a bit confused; either 'phase counter' or the net name would have been preferable.

hw/ip/pwm/rtl/pwm_chan.sv
rswarbrick: @alees24 has just pointed out that this comment didn't make sense. I was talking rubbish: please ignore! :-)

hw/ip/pwm/rtl/pwm_chan.sv
alees24: Reworded.

hw/ip/pwm/rtl/pwm_chan.sv
alees24: Done.

hw/ip/pwm/rtl/pwm_core.sv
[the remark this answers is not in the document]
rswarbrick: And now you have both! I like it :-)

rswarbrick (approved):
This looks good. Thank you very much for tidying up what was a bit messy before.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/pwm/rtl/pwm_core.sv

This has intentional functionality change (for the better) and has been carefully reasoned about and tested.

alees24:
CHANGE AUTHORIZED: hw/ip/pwm/rtl/pwm_core.sv

PR #25562, opened 2024-12-09
[bazel] Update bazelisk to 1.24.1

jwnrt:
This will fix master CI.

Required for bzlmod support. Older versions do not recognise `MODULE.bazel` as the root of the repo when searching for `.bazelversion`.


a-will:
Looks like the airgapped environment build script needs a little fix of some sort. Then this should be good to go!

jwnrt:
Airgapped build seems to work again. Merging to fix CI.

PR #25565, opened 2024-12-10
[corefiles] Fix 2 corefiles for Darjeeling

Razer6:
Ensure they are compliant with fusesoc 2.4

### DIFF
--- a/hw/top_darjeeling/lint/top_darjeeling.vbw
+++ b/hw/top_darjeeling/lint/top_darjeeling.vbw
@@ -3,4 +3,4 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # These lines are too long due to templating
-waive --rule=line-length --location=".*top_earlgrey.*"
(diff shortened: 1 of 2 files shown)


hw/top_darjeeling/top_darjeeling.core
```diff
@@ -87,10 +87,6 @@ filesets:
       # common waivers
       - lowrisc:lint:common
       - lowrisc:lint:comportable
-    files:
-      - lint/top_darjeeling.vbw
```
andreaskurth, line 91: Shouldn't we bring [this file over from `integrated_dev`](https://github.com/lowRISC/opentitan/blob/integrated_dev/hw/top_darjeeling/lint/top_darjeeling.vbw) rather than removing it here?

andreaskurth:
Verible and Verilator lint CI checks have passed, and other CI checks are not affected by this change. Merging

andreaskurth (approved):

Issue #23936, opened 2024-07-05
[prim_sha2_pad, hmac] Don't allow skipping the padding after stopping a hash operation


vogelpi:
### Description

During coverage closure for V2S, @martin-velay  and @gdessouky noted that there are some uncovered FSM transitions inside `prim_sha2_pad`:
- StLenHi -> StFifoReceive
- StLenLo -> StFifoReceive
- StPad00 -> StFifoReceive
- StPad80 -> StFifoReceive

An inspection of the RTL reveals that these transtions can be covered by something like the following sequence:
 1.  Trigger and message and make `prim_sha2_pad` go into any the `StPad80` / `StPad00` / `StLenHi` / `StLenLo` states. As a result of this `cfg_block` will assert and any future `reg_hash_start` / `reg_hash_continue` won't be forwarded to `prim_sha2_pad`.
 2. Trigger a `reg_hash_stop`. This will make `cfg_block` go low again in the next clock cycle.
 3. In the next clock cycle, trigger again a `reg_hash_start` / `reg_hash_continue`. This will trigger the uncovered transition.

The problematic thing with this is that after Step 2, the design should finish the hashing and then append the padding. By triggering another `reg_hash_start` or `reg_hash_continue`, software can cause the hardware to abort the padding and restart the hashing. This is not intended behavior, but software is also not expected to start another operation directly after stopping one. Instead, software is expected to wait for the `hmac_done` interrupt or to poll the status register to wait for the hashing and padding to finish before triggering further operations as outlined in the [HMAC programmer's guide](https://opentitan.org/book/hw/ip/hmac/doc/programmers_guide.html#saving-and-restoring-the-context) .

@andreaskurth and I concluded that this is NOT a terrible bug for Earlgrey-PROD and we will exclude the missing / buggy transitions from FSM coverage. But we should still fix this in a future release. The FSM should only be able to go back into the `StFifoReceive` state after finishing the padding.

vogelpi:
For more details, please refer to [this comment](https://github.com/lowRISC/opentitan/issues/23683#issuecomment-2208903846).

vogelpi:
PR #25590 which removes these undesirable FSM transitions has been merged into `master` now and I am thus closing this issue.

PR #25696 will remove the corresponding coverage exclusions on `master`. We have been doing these changes for the D3/V3 sign offs. 

[...]

PR #25568, opened 2024-12-10
[hmac,dv] Add new test to cover FSM transitions

martin-velay:
- New sequence to try to cover all the FSM transitions with random resets along with long message test.

### DIFF
--- a/hw/ip/hmac/data/hmac_testplan.hjson
+++ b/hw/ip/hmac/data/hmac_testplan.hjson
@@ -150,6 +150,14 @@
       stage: V3
       tests: ["hmac_smoke"]
     }
+    {
+      name: stress_reset
+      desc: '''Run multiple hmac_long_msg sequences with random on-the-fly reset to ensure that all
+            FSM transitions are exercised.
+            '''
+      stage: V3
+      tests: ["hmac_stress_reset"]
+    }
   ]
 
   covergroups: [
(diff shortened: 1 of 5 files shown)


andrea-caforio:
For context, the randomized reset signal in `hmac_stress_all_with_rand_reset` opens up FSM transition coverage holes in `prim_sha2_pad.sv`. These transitions are only triggered by the reset signal resetting the FSM state.

An example for such a reset-triggered transition is `StLenHi -> StIdle`, see below.
Screenshot from 2024-12-09 17-00-39

hw/ip/hmac/data/hmac_testplan.hjson
```diff
@@ -150,6 +150,14 @@
       stage: V3
       tests: ["hmac_smoke"]
     }
+    {
+      name: stress_reset
+      desc: '''Run multiple hmac_long_msg sequences with multiple random on-the-fly resets to ensure
```
rswarbrick, line 155: Nit: I don't think we need the second "multiple"? (Because each sequence will only see one reset)

rswarbrick (commented):
This looks really clear to me: I like it! Does this cover the FSM transition that was missing?

andrea-caforio:
> This looks really clear to me: I like it! Does this cover the FSM transitions that were missing?

I'll make another PR today that'll cover those missing transitions (referring to #23936.

hw/ip/hmac/data/hmac_testplan.hjson
martin-velay: Thanks!
True, I'll fix it.

And from what I saw in my short regression (only this test 50 times), I got 100% FSM coverage, whereas that wasn't always the case with a full regression. Let's see if now we systematically have these transitions covered.

rswarbrick (approved):

PR #25578, opened 2024-12-10
[bazel] Migrate `rules_rust` to bzlmod

jwnrt:
This PR migrates `rules_rust` to Bzlmod, using extensions for the toolchain, crates, and bindgen.

The large diff here is caused by additions to the lockfile.

Adjacent required changes:

* Rust crypto test harnesses reformatted. They were not formatted before for some reason.
* `RULES_RUST_BINDGEN_LIBSTDCXX_DIR` env var should no longer be needed. `libc++` is now always taken directly from LLVM.
* Tockloader executable moved from `actions.run_shell.inputs` to `.tools`. Bazel has stopped providing this script's runfiles when given as an input. Not sure why. `tools` seems correct, though.

A follow-up PR will remove `third_party/{rust,tock}/crates`. I haven't included the commit here since the number of file changes breaks part of CI.


nbdd0121 (commented):
Generally LGTM, but the commits need to be reordered to ensure that you can compile invididual commits:
* `.inputs` -> `.tools` fix and reformatting should be the first commits (or an separate PR that's merged first)
* .bazelignore changes should go with the commit that moves the corresponding crate universe to bazelmod.

Also, I think we'd better still have separate files for each third party dependency rather than having a giant MODULE.bazel file.

nbdd0121 (approved):
Overall LGTM, although I'm not experienced enough with Bazel to review the bindgen part. I'll leave it to @pamaury

third_party/rust/rust.MODULE.bazel
```diff
@@ -49,12 +49,14 @@ crate.annotation(
     crate = "libudev-sys",
     patch_args = ["-p1"],
     patches = ["@//third_party/rust/patches:libudev-sys-0.1.4.patch"],
+    repositories = ["crate_index"],
```
pamaury, line 53: I think it would be cleaner to add these lines in the previous commit. Even though they are not needed because there is a single repository, it makes the commit more atomic.

pamaury (approved):
This looks great, thanks @jwnrt :) 
I just suggest moving a few lines from one commit to another, see comment.

third_party/rust/rust.MODULE.bazel
jwnrt: You're right, updated.

PR #25579, opened 2024-12-10
[chip,dv] Fix $plusarg string

elliotb-lowrisc:
Incorrect inclusion of "+" at the start of the plusarg name string in a `$value$plusargs` call.

### DIFF
--- a/hw/top_earlgrey/dv/env/seq_lib/chip_common_vseq.sv
+++ b/hw/top_earlgrey/dv/env/seq_lib/chip_common_vseq.sv
@@ -23,7 +23,7 @@ class chip_common_vseq extends chip_stub_cpu_base_vseq;
 
   virtual task body();
     string csr_test_type;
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
Oops :-) This looks right to me.

rswarbrick:
Hmm, that said, the change that added that plus symbol was done by someone who presumably saw it work. I'm keen on this cleanup, but maybe some EDA tools discard the leading + here? (I just checked and the IEEE spec explicitly says not to use a +, but I guess an EDA tool can have this behaviour so long as no-one wants a plusarg that genuinely starts with a plus!)

elliotb-lowrisc:
A quick test in VCS showed it failing to pick up the plusarg when the superfluous "+" was present, and succeeding when it was removed. Perhaps other simulators behave differently, or the they got lucky with the randomisation of `num_trans` (which only selects between `1` and `2`).

PR #10982, opened 2022-02-19
[reggen/top] Initial support for register aliases 

msfschaffner:
This is a tooling extension that aims at supporting closed source registers in a more convenient way (#7688, #10324).

In particular, this adds support for an alias Hjson file that can override register names, fields, tags, descriptions and reset values of "generic" registers residing on the open-source side.
The requirement is that the register node structure of both the closed source and open source version is identical except for the attributes mentioned above.

To achieve this, the alias register file (residing in the closed repo) is validated against the structure of the open-source register file to ensure that the structure is identical, and then the overridable values are applied.

An alias file can be supplied to the `regtool` with the `--alias` switch. The actual code generation is then handled slightly differently for each backend:
- RTL: if the alias file is defined, reggen will produce a stripped down version of the `reg_pkg` where the register names have been overridden with closed source names and values. It also generates the associated `reg_top` with updated names so that the instantiating primitives can use the closed source names.
- DV: if the alias file is defined, reggen will produce a modified RAL model that has register/field handles with both generic and closed source names. This allows to mix and match both open and closed source register / field names when deriving closed source DV classes from open-source ones.
- SW: this is still TODO - however, the idea is to make this behave similarly to DV.

Open TODOs to be addressed after this PR has landed:
- update the foundry block-level environment to make use of this mechanism
- add SW backend
- integrate reggen changes into topgen so that the closed source chip-level DV can leverage this mechanism as well
- update the foundry chip-level simulation environment to make use of this mechanism

PR #25580, opened 2024-12-10
[multitop] Implement HW rules RFC

pamaury:
This PR implements the [Multitop HW desc rule RFC](https://docs.google.com/document/d/1J65grF7za5kHleVbqIxDfS6ASx_qmcGhegas9UNd3Qk/edit?usp=sharing). It replaces #24791 which is the original implementation.

There is one notable difference between this implementation and the one suggested in the RFC. 

# RFC suggested implementation

The RFC suggested to build the top description entire out of rules and providers:
```python
# In hw/ip/uart/BUILD:
# This target exports an OpenTitanIpInfo provider
opentitan_ip(
  name = "uart",
  hjson = "//path/to/uart.hjson"
)

# In hw/top_earlgrey/BUILD:
# This target exports an OpenTitanTopInfo provider
opentitan_top(
  name = "top_earlgrey_desc",
  hjson = "//path/to/top_earlgrey.gen.hjson",
  ips = [
    "//hw/ip/uart",
   ...
  ]
)
```
The information is then used to e.g. build headers:
```python
# In hw/top:
opentitan_ip_c_headers(
  name = "uart_c_regs",
  top = "//hw/top_earlgrey:top_earlgrey_desc",
  ip = "uart",
)
```
# Problem with this approach

While this approach works fine, we need to consider what happens when we mix this multitop. Suppose that we have a way to selecting a top though a bazel config and we use that to point to the relevant top description:
```python
# In hw/top
alias(
  name = "top_desc",
  actual = select({
    "//hw/top:is_earlgrey": "//hw/top_earlgrey:top_earlgrey_desc",
    "//hw/top:is_darjeeling": "//hw/top_earlgrey:top_darjeeling_desc",
  })
)

opentitan_ip_c_headers(
  name = "uart_c_regs",
  top = "//hw/top:top_desc",   # <-- NOTE: not pointing to earlgrey anymore
  ip = "uart",
)
```
This works fine because both Earlgrey and Darjeeling have a UART. Now consider an IP block that only exists in Earlgrey or Darjeeling:
```python
opentitan_ip_c_headers(
  name = "dma_c_regs",
  top = "//hw/top:top_desc", 
  ip = "dma",
)

opentitan_ip_c_headers(
  name = "usb_c_regs",
  top = "//hw/top:top_desc", 
  ip = "usb",
)
```
We run into an issue: the rule `opentitan_ip_c_headers` cannot work for `ip = dma` if run on earlgrey because Earlgrey does not have a DMA and therefore does not even know where the DMA hjson file is. One solution could be to output an empty header file but this would just push the problem onto the users: if they try to use it and depend on DMA register offsets, they will fail to compile. Therefore, the proper definition should be:
```python
opentitan_ip_c_headers(
  name = "dma_c_regs",
  top = "//hw/top:top_desc", 
  ip = "dma",
  target_compatible_with = opentitan_require_top("darjeeling") # Express that we want darjeeling
)

opentitan_ip_c_headers(
  name = "usb_c_regs",
  top = "//hw/top:top_desc", 
  ip = "usb",
  target_compatible_with = opentitan_require_top("earlgrey") # Express that we want earlgrey
)
```
Where `opentitan_require_top` is some unspecified-as-of-yet macro, for example:
```python
def opentitan_require_top(topname):
  return select({
    "//hw/top:is_{}".format(topname): [],
    "//conditions:default": ["@platforms//:incompatible"],
  })
```
But now run into another issue: we have to manually specify for each IP the list of tops that it is compatible with! This means that adding a top require adding 35+ conditions everywhere! Worse, if we change the definition of a top (e.g. Darjeeling which is supposed to be easily tweakable) then we have to change this as well. This defeats the entire point of have a top description generated by bazel.

# Formalizing the problem

Thinking back, what we really would like to write is this:
```python
opentitan_ip_c_headers(
  name = "dma_c_regs",
  top = "//hw/top:top_desc", 
  ip = "dma",
  target_compatible_with = opentitan_require_ip("dma") # Express that we only accept tops with the DMA IP
)
```
where `opentitan_require_ip` is a magical macro that somehow expands to the list of tops with the DMA block. In order to write such a macro, we need to remember that macros are expanded during Bazel's loading phase, i.e. before the build graph is even constructed. During the loading phase, the only "symbols" available to macros are:
- other macros defined in `.bzl` files
- variables defines in `.bzl` files

It is clear that in order to work, `opentitan_require_ip` needs access to a (possibly scaled-down) description of every top. Therefore the conclusion is that the **top descriptions needs to be available in the loading phase**. With the rule+provider approach from the RFC, it is only available in the analysis phase.

# Solution

The solution proposed in the RFC is therefore to still use the *idea* of the RFC but at the loading phase. Here `opentitan_ip` and `opentitan_top` become macros returning `struct`. Those need to be placed in newly created `.bzl` files to be usable in the loading phase:
```python
# In hw/ip/uart/defs.bzl:  (not a BUILD file)

UART = opentitan_ip(
  name = "uart",
  hjson = "//path/to/uart.hjson"
)

# In hw/top_earlgrey/defs.bzl:  (not a BUILD file)
load("//hw/ip/uart:defz.bzl", "UART")
EARLGREY = opentitan_top(
  name = "top_earlgrey_desc",
  hjson = "//path/to/top_earlgrey.gen.hjson",
  ips = [
    UART,
   ...
  ]
)

# In hw/top/defs.bzl:
load("//hw/top_earlgrey:defz.bzl", "EARLGREY")
load("//hw/top_darjeeling:defz.bzl", "DARJEELING")

ALL_TOPS = [EARLGREY, DARJEELING]
```
A simple implementation of those macros is to simply wrap the information in a `struct`:
```python
def opentitan_ip(name, hjson):
    return struct(
        name = name,
        hjson = hjson,
    )

def opentitan_top(name, hjson, ips):
    return struct(
        name = name,
        hjson = hjson,
        top_lib = top_lib,
        top_ld = top_ld,
        ips = ips,
    )
```
This allows use to implement `opentitan_require_ip` simply as follows:
```python
def opentitan_require_ip(ip):
    compatible_tops = []
    for top in ALL_TOPS:
        for _ip in top.ips:
            if _ip.name == ip:
                compatible_tops.append(top.name)
                break

    return select({
        "//hw/top:is_{}".format(top): []
        for top in compatible_tops
    } | {
        "//conditions:default": ["@platform//:incompatible"],
    })
```
Of course, we do not want to carry those struct around for the analysis phase so we still want to create a target that exports an `OpenTitanTopInfo` provider as before. This is implemented in the PR and a simplified version of the code is the following:
```python
# In hw/top/BUILD:

# Create one description per top
[
    describe_top(
        name = "top_{}_desc".format(top),
        all_tops = ALL_TOPS,
        top = top,
    )
    for top in ALL_TOP_NAMES
]

# In hw/top/defs.bzl:
def _describe_top(ctx):
    return [
        OpenTitanTopInfo(
            name = ctx.attr.topname,
            hjson = ctx.file.hjson,
            ip_hjson = ctx.attr.ip_hjson, # simplified: ctx.attr.ip_hjson has the wrong "type" and needs a bit of massaging...
        ),
    ]

describe_top_rule = rule(
    implementation = _describe_top,
    doc = """Create a target that provides the description of a top in the form of an OpenTitanTopInfo provider.""",
    attrs = {
        "hjson": attr.label(mandatory = True, allow_single_file = True, doc = "toplevel hjson file generated by topgen"),
        "ip_hjson": attr.label_keyed_string_dict(allow_files = True, doc = "mapping from hjson files to tops"),
        "topname": attr.string(mandatory = True, doc = "Name of the top"),
    },
)


def describe_top(name, all_tops, top):
    """
    Create a target that provides an OpenTitanTopInfo corresponding to the
    requested top.

    - all_tops: list of tops (created by opentitan_top).
    - top: name of the top to use.
    """

    # Details omitted: go through the all_tops struct and extract the relevant information

    describe_top_rule(
        name = name,
        hjson = top_hjson,
        ip_hjson = all_hjson,
        topname = top,
    )
```


hw/top/defs.bzl
```diff
@@ -14,3 +14,44 @@ ALL_TOP_NAMES = [
     top.name
     for top in ALL_TOPS
 ]
+
+def _all_ip_names():
+    names = {}
+    for top in ALL_TOPS:
+        for ip in top.ips:
+            names[ip.name] = {}
```
cfrantz, line 22: Would this be better as a comprehension? (also, maybe use an integer as the value, rather than empty dict?)
```
names = {ip.name: 1 for ip in top.ips for top in ALL_TOPS}
```

rules/host.bzl
```diff
@@ -23,6 +23,9 @@ def _host_tools_transition_impl(settings, attr):
         "//hw/bitstream/universal:rom": "//hw/bitstream/universal:none",
         "//hw/bitstream/universal:otp": "//hw/bitstream/universal:none",
         "//hw/bitstream/universal:env": "//hw/bitstream/universal:none",
+        # WARNING This is a horrible hack: when we transition to host, we pretend
+        # that this is earlgrey so opentitantool can compile...
```
cfrantz, line 27: Agree.  Longer term, we probably need to distinguish between chip-specific constants in opentitantool, stick them into something like a hashmap by chip, and add a `--chip=...` cmdline argument to select which set of constants we care about.

util/rewrite_all_hw.py
```diff
@@ -0,0 +1,310 @@
...
+
+def step1(project_root, commit):
+    new_files = []
+    for (_ippath, info) in all_ips.items():
+        ippath = project_root / Path(_ippath)
+        ip_name = ippath.name
+        is_template = info.get("is_template", False)
+
+        def_file_path = ippath / ("defs.bzl.tpl" if is_template else "defs.bzl")
+        # build_file_path = ippath / ("BUILD.tpl" if is_template else "BUILD")
```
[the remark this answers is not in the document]
pamaury: Yes some (all?) ipgen IP template `BUILD` file. Well technically they are not templates because their content does not depend on the template but we can't name them just `BUILD` because then bazel considers them as a package and this can cause all sorts of weird issues potentially. Therefore calling `BUILD.tpl` seems like an easy way to avoid this issue. Unless you have a better idea on how to handle this?

util/rewrite_all_hw.py
```diff
@@ -0,0 +1,310 @@
...
+    if args.step == "1":
+        step1(project_root, args.commit)
+    elif args.step == "2":
+        step2(project_root, args.commit)
+    elif args.step == "3":
+        step3(project_root, args.commit)
+    else:
+        assert False, "unknown step"
+
+
```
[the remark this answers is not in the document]
pamaury: Ah yes, although I should point out that this script will disappear at the end of the PR (I just forgot to add it yet). Still good practice though.

rules/host.bzl
pamaury: Yes that's a big issue. I would like to raise it in the SW WG tomorrow, there is clearly some work to be done here. I think the multitop SW PR is a good starting point to generate such a data structure.

a-will:
Out of curiosity, did you already consider a bazel module extension to create a repo that fills out the information from the hjson files? From the description, this starts to sound rather similar to a "foreign build system" or "foreign packaging" sort of problem. The hjson files seem to contain critical information about the build graph, such that it exists outside of bazel's BUILD files.

It looks like this PR goes the route of requiring the user to generate the graph with a separate call, but could we do it on the fly (in a bazel module extension) with what we already have with topgen, reggen, and the hjson files?

rules/host.bzl
a-will: One route: I could revive my devicetree generator, and opentitantool could consume its output for a description of the system. That would also leave open support for out-of-tree tops (instead of hard-coding entire known chips).

pamaury:
> Out of curiosity, did you already consider a bazel module extension to create a repo that fills out the information from the hjson files? From the description, this starts to sound rather similar to a "foreign build system" or "foreign packaging" sort of problem. The hjson files seem to contain critical information about the build graph, such that it exists outside of bazel's BUILD files.
> 
> It looks like this PR goes the route of requiring the user to generate the graph with a separate call, but could we do it on the fly (in a bazel module extension) with what we already have with topgen, reggen, and the hjson files?

Everything top related in this PR is generated by topgen (+one defs.bzl file per IP that needs to be created manually). I agree is not ideal but since we already need to run topgen to get the top library, top linker file and so on, adding one more file doesn't sound too bad and definitely was the simplest option (not the most elegant).

When I started this, it didn't really look like that and then evolved into this. If I was starting from scratch now, I would write it as a bazel extension ^^ Eventually we should probably replace the various `defs.bzl` with an extension, maybe taking the `top_<name>.gen.hsjon` files as input. However, I think it's not critical in the sense that it doesn't change the logic and the way it's used. Also I would like to make more changes to topgen before doing so.

hw/top/defs.bzl
```diff
@@ -0,0 +1,57 @@
...
+    top.name
+    for top in ALL_TOPS
+]
+
+def _all_ip_names():
+    names = {}
+    for top in ALL_TOPS:
+        for ip in top.ips:
+            names[ip.name] = {}
+    return names.keys()
```
a-will, line 23: Does this need to be sorted? Will it affect configuration hashes anywhere?

a-will (commented):
Just a few quick, probably ill-considered comments. Sorry I haven't had much time to look yet!

hw/top/BUILD
```diff
@@ -0,0 +1,108 @@
...
+    "opentitan_ip_rust_header",
+)
+
+package(default_visibility = ["//visibility:public"])
+
+# Use this flag to select the top.
+string_flag(
+    name = "top",
+    build_setting_default = "earlgrey",
+    values = ALL_TOP_NAMES,
```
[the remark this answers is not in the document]
pamaury: I thought about out-of-tree tops and I think it is ok: behind the scene, everything is just labels, the top and/or IPs can be out of tree. Potentially the only issue is the name itself: as you said it seems unlikely to have a collision but technically not impossible. If we wanted to make it really really future proof, we could replace the "names" (earlgrey, darjeeling) by a full bazel path and turn that into a `label_flag`, and maybe intro some aliases in `.bazelrc` like "earlgrey" and "darjeeling" for well-known tops.
A bazel extension might help with the registration step.

Regarding the last point, yes that's on purpose: with @cfrantz we want to make the exec_env more modular. Eventually we envision that a full description will be the top, the platform (fpga, sim_dv, probably platform is not good name, maybe call it target) and the board (hyperdebug, CW310). Then the exec_env would be replaced by a set of toolchain to provide the tools for the various building steps (compile, sign, transform, dispatch test). It's still at the design stage however, we only have some vague notes.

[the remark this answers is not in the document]
matutem: The life-cycle stage may also be part of ingredients that together create the exec env.

matutem:
Regarding the creation of the ips for a given top, the info is in the top hjson file. Perhaps we can at least enable the creation of a test to make sure all required IPS match between bazel and hjson, prior to having bazel  do it from the file itself?

util/rewrite_all_hw.py
```diff
@@ -0,0 +1,310 @@
...
+                "Remove old header targets",  # noqa: E231
+                "-m", "This commit was generated by ./util/rewrite_all_hw.py --step=3",
+            ],
+            check=True,
+            cwd = project_root,
+        )
+
+
+def main():
+    parser = argparse.ArgumentParser()
```
matutem, line 291: Adding args like this to ArgumentParser should print the doc string in the help message:
description=__doc__, formatter_class=RawTextHelpFormatter

util/rewrite_all_hw.py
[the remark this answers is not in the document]
matutem: Will we be able to dispense with the "all_files" target? That may make these BUILD and BUILD.tpl files unnecessary.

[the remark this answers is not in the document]
pamaury: Unfortunately, there was quite a lot of pushback against touching the "all_files" targets so I have given up on touching them for the time being.

hw/top/defs.bzl
pamaury: Sorting is not necessary for this PR but I guess it doesn't hurt to do it anyway if this ever gets used in a way where the order is important.

pamaury:
> Out of curiosity, did you already consider a bazel module extension to create a repo that fills out the information from the hjson files? From the description, this starts to sound rather similar to a "foreign build system" or "foreign packaging" sort of problem. The hjson files seem to contain critical information about the build graph, such that it exists outside of bazel's BUILD files.
> 
> It looks like this PR goes the route of requiring the user to generate the graph with a separate call, but could we do it on the fly (in a bazel module extension) with what we already have with topgen, reggen, and the hjson files?

I have looked into using a bazel extension and I think I can make it work with #25697. However, I think that for this particular PR, it is better to not use it. Indeed, the only file which gets autogenerated in the per-top `defs.bzl` containing the `opentitan_top` call. This file i generated by `topgen` when the top changes which is something that needs to be done anyway, so this does not change the status quo. The extra per-IP `defs.bzl` files calling `opentitan_ip` are created manually but are expected to not change, or very rarely so again I don't think there is a huge drawback to checking them for the moment.

The most important aspect is that even if we used a bazel extension to generated them, it would not change anything to how this information is used, i.e. it would generated the same things except that you would not see it. The only difference it would make is that the registration step is made a little bit easier.

Also regarding the `string_flag` versus `label_flag`, I don't think a label_flag is the best solution here. It's not really like a choosing a top is just "pointing something at something": it changes a lot of things in the build graph. From that perspective, selecting top is much closer to being a constraint value than a label flag. Unfortunately, constraint values cannot be specified on the command line which is why I went for a string flag. While it is true that this might create issues in case of name collision, I am not convinced that name collisions are a real issue here: it's trivial to just add any prefix/suffix/version to make the name different, and avoid the issue in the first place.

a-will:
> I have looked into using a bazel extension and I think I can make it work with #25697. However, I think that for this particular PR, it is better to not use it. Indeed, the only file which gets autogenerated in the per-top `defs.bzl` containing the `opentitan_top` call. This file is generated by `topgen` when the top changes which is something that needs to be done anyway, so this does not change the status quo. The extra per-IP `defs.bzl` files calling `opentitan_ip` are created manually but are expected to not change, or very rarely so again I don't think there is a huge drawback to checking them for the moment.
> 
> Longer term, I have some changes to topgen that will lead us to checking less software files into the repository. It might be a good point to revisit this choice.
> 
> The most important aspect is that even if we used a bazel extension to generated them, it would not change anything to how this information is used, i.e. it would generated the same things except that you would not see it. The only difference it would make is that the registration step is made a little bit easier.

I'd quibble with that point when we start to support out-of-tree tops. Right now, the topgen templates assume the definitions's labels are at particular paths in the main upstream bazel repo. At some point, the .bzl files (or whatever is used to assemble the graph) are going to need cross-repo paths to be filled in, and I suspect that will strongly indicate a desire to write the module extension.

The function of `data_defs.tpl` would end up getting absorbed into it--Otherwise, we're asking out-of-tree users to write those by hand, since that template likely can't handle the paths.

We don't have to add this support right now, though! But we might want to get that on a roadmap.

> Also regarding the `string_flag` versus `label_flag`, I don't think a label_flag is the best solution here. It's not really like a choosing a top is just "pointing something at something": it changes a lot of things in the build graph. From that perspective, selecting top is much closer to being a constraint value than a label flag. Unfortunately, constraint values cannot be specified on the command line which is why I went for a string flag. While it is true that this might create issues in case of name collision, I am not convinced that name collisions are a real issue here: it's trivial to just add any prefix/suffix/version to make the name different, and avoid the issue in the first place.

Yeah, that seems fine to me. I was "thinking aloud" and figured I'd raise the point for discussion. Thanks for considering it! :slightly_smiling_face:

pamaury:
> 
> I'd quibble with that point when we start to support out-of-tree tops. Right now, the topgen templates assume the definitions's labels are at particular paths in the main upstream bazel repo. At some point, the .bzl files (or whatever is used to assemble the graph) are going to need cross-repo paths to be filled in, and I suspect that will strongly indicate a desire to write the module extension.
> 
> The function of `data_defs.tpl` would end up getting absorbed into it--Otherwise, we're asking out-of-tree users to write those by hand, since that template likely can't handle the paths.
> 
> We don't have to add this support right now, though! But we might want to get that on a roadmap.

Agreed and thanks for your comments, it's definitely helpful to think about all that. One issue with the paths is that topgen can access a CLI particular to search for IPs outside of the repo but it has no way to know what bazel repo they are in (if any!). So yeah, I agree that generally speaking, bazel modules and extensions are the better way to go to support out of tree stuff, and avoid use of ugly environment variables. It might actually be worth thinking about this more generally: we have a similar issue for hooks and other things that could benefit from bazel extensions.

util/rewrite_all_hw.py
```diff
@@ -0,0 +1,310 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
pamaury: Since the script is removed at the end, I thought it was a bit too much to split it into three. I can still do it if you think it's really necessary though?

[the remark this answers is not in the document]
nbdd0121: I meant to only have the script in the commit message so it's never added or removed as part of repository. This makes the script and what it produces atomic in git history, and cause less confusion when reviewing by commits.

Add and then remove is also fine to me, I don't have too strong of an opinion about this.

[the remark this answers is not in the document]
pamaury: Ah I see, that makes a lot of sense. I will do that.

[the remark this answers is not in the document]
pamaury: done

pamaury:
@cfrantz @a-will are you happy with the current state? If so, can any of you approve?

hw/top/defs.bzl
```diff
@@ -0,0 +1,54 @@
...
+    """
+    compatible_tops = []
+    for top in ALL_TOPS:
+        for _ip in top.ips:
+            if _ip.name == ip:
+                compatible_tops.append(top.name)
+                break
+
+    return select({
+        "//hw/top:is_{}".format(top): obj
```
a-will, line 37: This top matching may be okay, too, but we could also add constraint values for the IPs directly and define platforms for the tops. Since it's all hidden underneath `opentitan_require_ip()`, I think we can defer that decision as well, though.

hw/top_darjeeling/data/autogen/defs.bzl
```diff
@@ -0,0 +1,83 @@
...
+load("//hw/ip/spi_device:defs.bzl", "SPI_DEVICE")
+load("//hw/ip/spi_host:defs.bzl", "SPI_HOST")
+load("//hw/ip/sram_ctrl:defs.bzl", "SRAM_CTRL")
+load("//hw/ip/uart:defs.bzl", "UART")
+
+DARJEELING = opentitan_top(
+    name = "darjeeling",
+    hjson = "//hw/top_darjeeling/data/autogen:top_darjeeling.gen.hjson",
+    top_lib = "//hw/top_darjeeling/sw/autogen:top_darjeeling",
+    top_ld = "//hw/top_darjeeling/sw/autogen:top_darjeeling_memory",
```
a-will, line 48: This might get a bit more interesting with support for different address spaces (depending on how software uses them), but we can iterate. There aren't terribly many tops at this time, so it's easy enough to update.

a-will (commented):
Ran out of time, so I'll have to come back: I still need to look at the particular implementation of the rules. However, the general way that this setup works seems good for now.

Thanks for the hard work!

hw/top/defs.bzl
pamaury: The decision to hide the details behind macros is a deliberate one so we are not bound by the current implementation and could switch to platforms in the future. We had a long discussion about platforms with @cfrantz and I am not convinced there are the way to go for tops, at least in their current state (there were some improvements in bazel 8.0). It would require one constraint value per IP and top, and in the future we might want more (like "does the top have an embedded flash"). All of this for no obvious benefit over the current implementation with a string flag?

It seems to me that platforms are more useful for toolchain related decisions, and therefore we might want to use them to replace execution environment. Constraint value could be very useful to represent the test platform/board.

hw/top_darjeeling/data/autogen/defs.bzl
pamaury: This file is autogenerated by topgen so changing is indeed very simple. For addresses spaces, we don't seem to have any use case yet where the software would like to know about address spaces other than its own. In particular, I imagine that if there were two address spaces for SW (e.g. one per CPU?) then you would compile your software for either one or the other and that would become part of the "top selection". All other cases could be handled by adding information to the DT but again I haven't yet seen this requirement.
Maybe @Razer6 has more information on this topic?

rules/autogen.bzl
```diff
@@ -60,25 +45,12 @@ def _hjson_c_header(ctx):
...
     attrs = {
-        "srcs": attr.label_list(allow_files = True),
-        "node": attr.string(
-            doc = "Register block node to generate",
-        ),
-        "alias": attr.label(
-            mandatory = False,
-            allow_single_file = True,
-            doc = "A path to an alias file",
-        ),
```
a-will, line 74: The alias feature might be used out-of-tree, to replace some of those integrator-specific CSRs with the real definitions. Dropping support for it _might_ be a problem. See https://github.com/lowRISC/opentitan/pull/10982 for the motivations.

Probably the Rust generation should support it as well, but that's a separate problem, haha.

a-will (approved):
LGTM, except I think the alias feature for C header generation may need to be preserved.

rules/autogen.bzl
pamaury: Thanks for the explanation, I had no idea what aliases were (I could not find any example or documentation). I can add it back to the rule and add an `alias` attribute to the `opentitan_ip` macros. This way integrators can just edit the `defs.bzl` files to add the alias if them want to.

a-will: This would be used in a separate header generation target out-of-tree, I think. In other words, in a separate bazel workspace, someone would write their own `opentitan_ip_c_header` target that uses our rule, and our rule's responsibility is simply to pass through the file to the command line option and into the sandbox.

I'll note that we have no uses of the alias attribute in our repo. One of our integrators might be able to help, but we probably ought to document this, haha.

pamaury: I have implemented the aliases. It's supported both in the `opentitan_ip_c_header` rule and in the `opentitan_ip` macro so that it's also possible to just add the alias attribute directly at the definition point.

hw/top_darjeeling/data/autogen/defs.bzl
Razer6: Right now, there would be only one address space, where software is compiled to, i.e., the one where Ibex sits. A system though might have different address spaces, that are not accessible via Ibex (or at least directly), i.e., the debug address space or the CTN/MBX address space in Darjeeling. I think that's good for now.

Razer6 (approved):
This looks good to me. Thanks.

pamaury:
The CI failures are unrelated (flaky tests, flaky CW305 bitstream build).

PR #25582, opened 2024-12-10
[sram_ctrl,dv] Increase fatal_error timeout

nasahlpa:
With the introduction of the SRAM readback mode (c.f., lowRISC/opentitan#23212) memory requests can be delayed. The SRAM DV tests are adapted such that the readback feature can be disabled or enabled anytime during a test.

The sram_ctrl_passthru_mem_tl_intg_err failed occasionally because the integrity alert was expected to arrive without any delay. This commit increases the max_delay parameter for the expected alert to take the readback mode delay as well as a potential enabling or disabling of the feature into account.


rswarbrick (approved):

nasahlpa:
Merging as the failing CW340 CI tests are unrelated to this DV-only change.

PR #25583, opened 2024-12-10
[rom_ctrl, dv] Removed the option of assert coverage collection

KinzaQamar:
tlul_adapter_sram contains fifos (prim_fifo_sync) and we are'nt required to collect assertion coverage. This will reduce the percentage of assert coverage being uncovered for the adapter.


rswarbrick (approved):
Thanks: this looks right to me (and I'm feeling a bit silly for not noticing it in the first place)

PR #25585, opened 2024-12-10
 [rom_ctrl, dv] Exclusion of non-occuring cases for tlul_adapter_sram 

KinzaQamar:
t is impossible to see !tl_i_int.a_valid & !error_internal. When a_valid is false, it is seen
in u_err inside tlul_adapter_sram.sv. This will make a_config_allowed inside u_err false as it
depends on addr_sz_chk, mask_chk and fulldata_chk. All of them are false if a_valid is false.
This means that the first term inside err_o definition in u_err is false which raised err_o.
Next, err_o become visible to error_det as tlul_error inside tlul_adapter_sram and it will
be true if tlul_error is true. error_det will then be seen as error_i to u_sram_byte inside
tlul_adapter_sram.
Since EnableIntg parameter is 0 for rom_ctrl, error_i comes out as error_o from u_sram_byte.
Then error_o would be seen as error_internal to tlul_adapter_sram and the conditional statement
is not able to cover 00 case for this reason.

It is impossible to get sram_ack & we_o. if sram_ack is true, then we have (req_o & gnt_i) = 1.
For req_o to become true requires error_internal as false. we_o become true when there is an
a_valid and a_opcode as Put. But when a_opcode is a Put, wr_vld_error in the adapter becomes
true as ErrOnWrite is true for the adapter in case of rom_ctrl. Which means that error_internal
is true and req_o is false, which leads to sram_ack being false.

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
+++ b/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
@@ -112,3 +112,26 @@ Condition 34 "201396280" "(d_valid && d_error) 1 -1" (1 "01")
 // the request will not already have been popped from the request fifo, so reqfifo_rvalid must be
 // true.
 Condition 43 "721931741" "(rvalid_i & reqfifo_rvalid) 1 -1" (2 "10")
+
+// It is impossible to see !tl_i_int.a_valid & !error_internal. When a_valid is false, it is seen
+// in u_err inside tlul_adapter_sram.sv. This will make a_config_allowed inside u_err false as it
+// depends on addr_sz_chk, mask_chk and fulldata_chk. All of them are false if a_valid is false.
+// This means that in u_err, the first term of err_o becomes true.
+// Next, err_o become visible to error_det as tlul_error inside tlul_adapter_sram and it will
+// be true if tlul_error is true. error_det will then be seen as error_i to u_sram_byte inside
+// tlul_adapter_sram.
+// Since EnableIntg parameter is 0 for rom_ctrl, error_i comes out as error_o from u_sram_byte.
+// Then error_o would be seen as error_internal to tlul_adapter_sram.
+//
+// To conclude the reasoning above, it is impossible to see error_internal = 0 when a_valid = 0 as
+// when a_valid appears as false to u_err will automatically raise err_o. Then err_o leads to
+// error_internal. Thus, we can't expect !tl_i_int.a_valid & !error_internal and for the reasons
+// above, the conditional statement is unable to cover case 011.
+Condition 37 "2164803938" "(tl_i_int.a_valid & reqfifo_wready & ((~error_internal))) 1 -1" (1 "011")
+
+// It is impossible to get sram_ack & we_o. we_o becomes true when there is an a_valid and a_opcode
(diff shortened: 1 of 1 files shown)


rswarbrick:
It looks like we diagnosed this wrongly, which is causing CI failures from e.g. the `sram_ctrl` smoke test. Would you mind taking a look to figure out what we got wrong?

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -112,3 +112,22 @@ Condition 34 "201396280" "(d_valid && d_error) 1 -1" (1 "01")
...
+// It is impossible to see !tl_i_int.a_valid & !error_internal. When a_valid is false, it is seen
+// in u_err inside tlul_adapter_sram.sv. This will make a_config_allowed inside u_err false as it
+// depends on addr_sz_chk, mask_chk and fulldata_chk. All of them are false if a_valid is false.
+// This means that in u_err, the first term of err_o becomes true.
+// Next, err_o become visible to error_det as tlul_error inside tlul_adapter_sram and it will
+// be true if tlul_error is true. error_det will then be seen as error_i to u_sram_byte inside
+// tlul_adapter_sram.
+// Since EnableIntg parameter is 0 for rom_ctrl, error_i comes out as error_o from u_sram_byte.
+// Then error_o would be seen as error_internal to tlul_adapter_sram and the conditional statement
+// is not able to cover 00 case for this reason.
```
rswarbrick, line 125: I'm not sure that 00 is right here. Should this be something like `0*1` ?

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -112,3 +112,22 @@ Condition 34 "201396280" "(d_valid && d_error) 1 -1" (1 "01")
...
+Condition 37 "2164803938" "(tl_i_int.a_valid & reqfifo_wready & ((~error_internal))) 1 -1" (1 "011")
+
+// It is impossible to get sram_ack & we_o. we_o become true when there is an a_valid and a_opcode
```
rswarbrick, line 128: Nit: I think this should be "becomes"

rswarbrick (commented):
Tiny nits about the comments, but thank you for the careful explanations. This looks good to me!

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
KinzaQamar: For me, 00 was !a_valid and !error_internal. But replaced it with 01 to avoid confusion

rswarbrick: Oh, I'm really sorry. I was imagining three bits, where we don't have an opinion about the middle one. This would correspond to the three bits in the conditional coverage item. I don't think that `0*1` is a good way to write that though!

Maybe we should be really explicit, so end with something like "If error_o is true then it will cause error_internal to be true as well. As a result, if tl_i_int.a_valid is false then error_internal will be true, so we can't see the 011 condition (or 001 either)".

rswarbrick (approved):
Thanks for this: looks good to me!

Issue #24692, opened 2024-09-30
[hmac] Code coverage exclusions


martin-velay:
### Description

Some code coverage holes should be added to the exclusion file:

- [hmac_core.sv](https://github.com/lowRISC/opentitan/blob/2e88b7aada39047590d662c0a7066b452cd82a07/hw/ip/hmac/rtl/hmac_core.sv):
  - Branch line 192 → the default case cannot be reached as all the sel_rdata possible values are explicitly specified and covered. This should be excluded.
  - Branch line 209-225 → any other value than SHA2_256/384/512 is not supposed to be programmed, otherwise an interrupt is raised. This non-existing “else” should be excluded.
- [prim_sha2.sv](https://github.com/lowRISC/opentitan/blob/2e88b7aada39047590d662c0a7066b452cd82a07/hw/ip/prim/rtl/prim_sha2.sv):
  - Branch line 124-132 → any other value than SHA2_256/384/512 is not supposed to be programmed, otherwise an interrupt is raised. This non-existing “else” should be excluded.
- [prim_sha2_pad.sv](https://github.com/lowRISC/opentitan/blob/2e88b7aada39047590d662c0a7066b452cd82a07/hw/ip/prim/rtl/prim_sha2_pad.sv):
  - Line 120 → any other value than SHA2_256/384/512 is not supposed to be programmed, otherwise an interrupt is raised. This should be excluded.
  - Line 206 → this requires to trigger `hash_start` or `hash_continue` while SHA core is disabled and also while the state machine is in Idle mode. This is outside the specification and won’t be easy to trigger, so we may as well exclude it.
  - Line 251 → any other value than SHA2_256/384/512 is not supposed to be programmed, otherwise an interrupt is raised. This should be excluded.
  - Line 251 and 257 when shaf_rready_i is equal to 0 → The StPad80 state gets entered through the StFifoReceive state, for which hash_start_i has to be 1. Outside of prim_sha2_pad, in prim_sha2, hash_start_i == 1 causes the FIFO FSM to enter the FifoLoadFromFifo state. In this state, prim_sha2 sets prim_sha2_pad’s shaf_rready_i input to 1 (through update_w_from_fifo) whenever shaf_rvalid is set. shaf_rvalid is driven by prim_sha2_pad to constant 1 in the StPad80 state. Thus shaf_rready_i == 1’b0 cannot occur and should be excluded.
[...]

PR #25590, opened 2024-12-10
[hmac, rtl] Do not skip padding after a hash stop command

andrea-caforio:
This commit removes some conditional FSM states transition that were deemed undesirable in `prim_sha2_pad`, see #23936. It made it possible to cancel an ongoing padding operation, which is disallowed by the HMAC module. As a side-effect, the removal plugs some existing FSM transition coverage holes in `prim_sha2_pad` whose manual exclusions can be removed as well.

This change makes it impossible to transition into the `StFifoReceive` state from either `StPad80`, `StPad00`, `StLenHi` or `StLenLo` by asserting the `hash_go` signal.

### DIFF
--- a/hw/ip/prim/rtl/prim_sha2_pad.sv
+++ b/hw/ip/prim/rtl/prim_sha2_pad.sv
@@ -337,8 +337,13 @@ module prim_sha2_pad import prim_sha2_pkg::*;
       end
     endcase
 
-    if (!sha_en_i)    st_d = StIdle;
-    else if (hash_go) st_d = StFifoReceive;
+    if (!sha_en_i) begin
+      st_d = StIdle;
+    // We do not allow the cancellation of an ongoing padding operation, i.e., reverting back to the
+    // `StFifoReceive` state while being in the states `StPad80`, `StPad00`, `StLenHi` or `StLenLo`.
+    end else if (hash_go && (st_q inside {StIdle, StFifoReceive})) begin
+      st_d = StFifoReceive;
+    end
   end
 
   // tx_count


rswarbrick (commented):
Splatting the exclusion list might be right (I haven't reviewed it carefully, but it seems reasonable). But doesn't the RTL change also imply that the UNR list might change?

andrea-caforio:
> Splatting the exclusion list might be right (I haven't reviewed it carefully, but it seems reasonable). But doesn't the RTL change also imply that the UNR list might change?

Thank you for the review @rswarbrick. You're right, the UNR file should be updated too. Unfortunately, we cannot generate these exclusion files with our VCS license on the Zurich workstation.

vogelpi (approved):
This looks good to me from an RTL perspective. Thanks for the PR @andrea-caforio !

Did you run a full regression for this and inspect the coverage yet?

andrea-caforio:
> This looks good to me from an RTL perspective. Thanks for the PR @andrea-caforio !
> 
> Did you run a full regression for this and inspect the coverage yet?

Thank you @vogelpi for the review. I ran a full regression and all the transitions are now properly covered (see screenshot). I further added a second commit that updates the exclusion file with the list in #24692. @martin-velay you might want to double-check this.

Screenshot from 2024-12-18 15-43-29

martin-velay:
Hi @andrea-caforio, I would prefer to have the exclusions in a separate PR as there is no direct link between this RTL fix and the exclusions. Thanks in advance !

andrea-caforio:
> Hi @andrea-caforio, I would prefer to have the exclusions in a separate PR as there is no direct link between this RTL fix and the exclusions. Thanks in advance !

Here you go #25696.

hw/ip/prim/rtl/prim_sha2_pad.sv
```diff
@@ -337,8 +337,13 @@ module prim_sha2_pad import prim_sha2_pkg::*;
       end
     endcase
 
-    if (!sha_en_i)    st_d = StIdle;
-    else if (hash_go) st_d = StFifoReceive;
+    if (!sha_en_i) begin
+      st_d = StIdle;
+    // We do not allow the cancellation of an ongoing padding operation, i.e., reverting back to the
+    // `StFifoReceive` state while being in the states `StPad80`, `StPad00`, `StLenHi` or `StLenLo`.
+    end else if (hash_go && (st_q == StIdle || st_q == StFifoReceive)) begin
```
andreaskurth, line 344:
```suggestion
    end else if (hash_go && (st_q inside {StIdle, StFifoReceive})) begin
```
would be a bit more concise, but please don't consider this a required change

andreaskurth:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_sha2_pad.sv

This PR removes some undesired FSM transitions that were previously uncovered. This is okay.

PR #25591, opened 2024-12-11
[sram_ctrl,dv] Fix readback_err test for write iteration 0

nasahlpa:
When injecting a fault into write transaction 0, we previously were writing 0 to memory. However, as the memory seems to be initialized with 0, the readback returned the init value 0. Hence, the readback check was not failing as expected.


rswarbrick (approved):
Ha! Yep, looks sensible to me.

PR #25593, opened 2024-12-11
[dv] Fix DVSim after GUI_DEBUG addition

martin-velay:
- GUI_DEBUG mode addition is creating an issue when tests are going into timeout. Because Deploy.py is managing GUI_DEBUG mode through GUI mode, but in LocalLauncher.py we were trying to access gui_debug attribute which doesn't exists.

### DIFF
--- a/util/dvsim/LocalLauncher.py
+++ b/util/dvsim/LocalLauncher.py
@@ -100,7 +100,7 @@ class LocalLauncher(Launcher):
         self.job_runtime_secs = elapsed_time.total_seconds()
         if self.process.poll() is None:
             if (self.timeout_secs and (self.job_runtime_secs > self.timeout_secs) and
-                    not (self.deploy.gui or self.deploy.gui_debug)):
+                    not (self.deploy.gui)):
                 self._kill()
                 timeout_message = 'Job timed out after {} minutes'.format(
                     self.deploy.get_timeout_mins())


martin-velay:
Here was the error:

```
Traceback (most recent call last):
  File "util/dvsim/dvsim.py", line 796, in <module>
    main()
  File "util/dvsim/dvsim.py", line 776, in main
    results = cfg.deploy_objects()
  File "/home/dev/src/util/dvsim/FlowCfg.py", line 397, in deploy_objects
    return Scheduler(deploy, get_launcher_cls(), self.interactive).run()
  File "/home/dev/src/util/dvsim/Scheduler.py", line 162, in run
    changed = self._poll(hms) or timer.check_time()
  File "/home/dev/src/util/dvsim/Scheduler.py", line 371, in _poll
    status = item.launcher.poll()
  File "/home/dev/src/util/dvsim/LocalLauncher.py", line 103, in poll
    not (self.deploy.gui or self.deploy.gui_debug)):
AttributeError: 'RunTest' object has no attribute 'gui_debug'
```

Hopefully this PR will fix it.

rswarbrick:
Looking at the original addition, should this have actually been looking at a `FlowCfg` object (instead of the `Deploy` object)?

martin-velay:
No I think it's fine the way we do. The `gui_debug` option can be considered as a simple `gui` option at this level as the behaviour we want to see from it doesn't differ.

rswarbrick (approved):

rswarbrick:
Thanks for tidying this up!

PR #25594, opened 2024-12-11
[hmac, sw] Revert hash stop hang workaround

andrea-caforio:
In line with the RTL fix in #24771 for the hash stop hang bug described in #24767, this commit reverts four commits that implemented a software workaround for the issue and only keeps the idle polling timeout implementation from #24944. This workaround and its corresponding documentation is now obsolete.

Reverted commits:
  - e869152 #25070
  - 2b072c4 #25012
  - 7f32449 #24966
  - f591bea #24944

### DIFF
--- a/sw/device/lib/crypto/drivers/hmac.c
+++ b/sw/device/lib/crypto/drivers/hmac.c
@@ -67,6 +67,15 @@ OT_ASSERT_ENUM_VALUE(HMAC_DIGEST_15_REG_OFFSET, HMAC_DIGEST_14_REG_OFFSET + 4);
 enum {
   /* The beginning of the address space of HMAC. */
   kHmacBaseAddr = TOP_EARLGREY_HMAC_BASE_ADDR,
+  /* Timeout value for the polling
+   * As min 3 clock cycles are required to perform a read access to the STATUS
+   * hmac_idle register. And as we should observe 240 cycles (80 for the inner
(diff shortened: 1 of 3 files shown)


sw/device/lib/crypto/drivers/hmac.c
```diff
@@ -89,6 +75,8 @@ enum {
  * It returns error if HMAC HWIP becomes idle without firing `hmac_done`
  * interrupt.
  *
+ * TODO(#23191): It might be beneficial to have a timeout value for the polling.
+ *
```
vogelpi, line 79: Doing a plain revert of the other commits is fine and the approach is good, it's easier to review this.

What is not ideal is that it means to re-introduce this TODO here which got resolved by the workaround. Would it be possible to do a separate commit or even PR (whatever you prefer @andrea-caforio ) to resolve this? You can take inspiration from the reverted commits on how to do this.

andrea-caforio: Thank you @vogelpi for the review and catching this issue. I reinstantiated the polling timeout implementation in a separate commit as to not spam the PR list with code that has already been review before.

vogelpi (approved):
This looks good to me, thanks @andrea-caforio !

martin-velay (approved):
This LGTM, thanks @andrea-caforio for taking care of it !

jadephilipoom (approved):

PR #25602, opened 2024-12-11
[sram_ctrl,dv] Increase sram_ctrl_bijection test timeout

nasahlpa:
This test is quite long and sometimes the default test timeout was not enough, letting the test fail in regression runs. Hence, this commit doubles the test timeout, letting previously failing tests pass again.


nasahlpa:
Good point @rswarbrick! I calculated an upper bound which almost matches my completely random guess.

Issue #23594, opened 2024-06-10
[hmac,dv] Analyze unreached feature coverage items


martin-velay:
### Description

The current feature coverage shows that some stuffs are not exercised:

- [x] msg_len_cg: long messages, and some others auto lengths (closed by https://github.com/lowRISC/opentitan/pull/23864 and https://github.com/lowRISC/opentitan/pull/23901)
- [x] pending_req_on_rst_cg.cp_req_pending (closed by https://github.com/lowRISC/opentitan/pull/25843)
- [x] sha2_invalid, sha2_none and key_invalid/none (closed by https://github.com/lowRISC/opentitan/pull/23843)
- [x] tl_errors_cg.cp_mem_byte_access_err/cp_mem_ro_err

andreaskurth:
@martin-velay: Should we move this to V3 and M7?

martin-velay:
@andreaskurth: I think we should keep it in M5 as it's quite important to ensure at least that, those items are covered when all the ongoing related PR will have been merged:
- msg_len_cg: long messages, and some others auto lengths
- sha2_invalid, sha2_none and key_invalid/none

martin-velay:
@rswarbrick 
Could you tell me if you have any idea how to cover the 2 remaining items? As those covergroups are coming from the common DV framework.

rswarbrick:
I misunderstood this message at first and thought you were talking about the remaining coverage points from your message on 26/7.

On those coverage points items:
- I think I agree with what you said, but...
- Looking at last night's run, `sha2_invalid`, `sha2_none` (in the `digest_size` coverpoint of `cfg_cg`) are hit.
- Similarly, `key_invalid` and `key_none` (in the `key_length` coverpoint of `cfg_cg`) are hit.
- The `msg_len_cg` results should be better after the fix that you merged this morning, thanks.
- But I notice that we haven't seen any hits in `msg_len_upper_cp` for some reason. Does that mean we never give a massive message length with a nonzero upper word? I don't think this has been fixed yet?

(I'll respond to the other items in a minute)

rswarbrick:
For `pending_req_on_rst_cg.cp_req_pending`:

I think this wants to see a reset when there is a pending request (so `pending_a_req` in `tl_monitor` is nonempty). This gets pushed to when we see a request on the A side (in `check_a_channel`) and then popped when we see the response (in `check_d_channel`).

[...]

PR #25604, opened 2024-12-11
[cip,dv] Explicit some fcov flags init val.

martin-velay:
~~add the option "goal" to 0 to exclude a coverpoint from the coverage calculation as VCS seems to ignore the weight option to 0 for some reason.~~
explicit has_mem_byte_access_err, has_wo_mem, has_ro_mem values as no default has been set, which can lead to different behaviour with different simulators I guess.
~~related to issue https://github.com/lowRISC/opentitan/issues/23594~~

### DIFF
--- a/hw/dv/sv/cip_lib/cip_base_pkg.sv
+++ b/hw/dv/sv/cip_lib/cip_base_pkg.sv
@@ -63,6 +63,10 @@ package cip_base_pkg;
     uvm_mem mems[$];
     reg_block.get_memories(mems);
 
+    has_mem_byte_access_err = 0;
+    has_wo_mem              = 0;
+    has_ro_mem              = 0;
+
     foreach (mems[i]) begin
       dv_base_mem dv_mem;
       `downcast(dv_mem, mems[i], , , msg_id)


hw/dv/sv/cip_lib/cip_base_pkg.sv
```diff
@@ -68,9 +68,19 @@ package cip_base_pkg;
       `downcast(dv_mem, mems[i], , , msg_id)
       if (!dv_mem.get_mem_partial_write_support() && dv_mem.get_access() == "RO") begin
         has_mem_byte_access_err = 1;
+      end else begin
+        has_mem_byte_access_err = 0;
```
rswarbrick, line 72: This isn't quite right: we're supposed to be calculating whether *any* memory might generate byte access errors. Similarly for `has_wo_mem` and `has_ro_mem`.

martin-velay: Oh yes you're right, thanks! I've just changed it to make the initialization prior the foreach loop.

martin-velay:
@rswarbrick , this PR is ready to be merged I think. Thanks!

rswarbrick (approved):
This looks sensible to me, thanks!

PR #25605, opened 2024-12-11
[ci] build and cache bitstream pre-merge

nbdd0121:


jwnrt (approved):

jwnrt:
This took a lot of tries. Doesn't seem like we'll be able to use it for `master` unless we bring the queue down right?

nbdd0121:
We can increase the timeout

PR #25606, opened 2024-12-11
[dv] Add option to ignore cov when test failed

martin-velay:
- Tests marked as "fail" are ignored for coverage score computation.


martin-velay:
I just found out that ignoring a failing is actually handled at DVSim level:
```
    def post_finish(self, status):
        if status != 'P':
            # Delete the coverage data if available.
            rm_path(self.cov_db_test_dir)
```
Sorry for this unnecessary PR!

PR #25608, opened 2024-12-11
[manuf] cleanup AST CSR init and program CreatorSwCfg last

timothytrippel:
This update the FT individualization firmware to:
1. not program the AST CSRs twice, as the sram_start.S assembly code that runs during SRAM program initialization already does so (thus reverting the changes in #25563), and
2. programs the CreatorSwCfg partition last, to enhance the reentrant nature of the FT individualization flow, since programming this partition enables ROM execution, which is better if done last.


moidx (approved):

moidx:
Merged in https://github.com/lowRISC/opentitan/pull/25609.

PR #25610, opened 2024-12-11
Fixed typo in print statement

tornupnegatives:
The original statement contained an interpolated string without the string interpolation token 'f'


rswarbrick (approved):
Oops! :-) Thanks for sorting this out.

PR #25611, opened 2024-12-11
[dv] Be more careful not to leave orphan processes in clk_rst_if

rswarbrick:
The existing code would always leave at least one process lying around when it ran wait_clks_or_rst. Avoid doing that.


matutem (approved):
Nice. LGTM

PR #25612, opened 2024-12-11
[dv] Avoid timing out if we call tl_access_sub lots in parallel

rswarbrick:
When we start a sequence to perform a CSR access, we wrap it in a DV_SPINWAIT with a timeout of e.g. tl_access_timeout_ns. This is a good idea! If the TL bus dies, we'd like the test to time out and fail.

But we might call e.g. csr_rd many times in parallel. If we do so, we'll end up enqueuing many copies of tl_seq. These each take some time to run and a very wide queue will cause the DV_SPINWAIT above to time out. This is a bit silly. There's no problem with the hardware: we're just doing a load of stuff, which takes a long time.

To avoid the problem, we set up an extra level of queueing. Getting through the first queue allows us access to the sequencer and causes the timer to be started.

### DIFF
--- a/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
+++ b/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
@@ -265,9 +265,22 @@ class cip_base_vseq #(
         data  == local::data;)
 
     csr_utils_pkg::increment_outstanding_access();
-    `DV_SPINWAIT(`uvm_send_pri(tl_seq, 100),
-                 $sformatf("Timeout waiting tl_access : addr=0x%0h", addr),
-                 tl_access_timeout_ns)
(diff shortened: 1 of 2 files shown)


rswarbrick:
Ah nuts: it looks like this is failing CI because it breaks something in an `rv_dm` test. I'm looking into that now.

rswarbrick:
Hmm. This really doesn't work. The problem is that you might have buses A, B and enqueue 1000 operations on bus A and immediately afterwards do an operation on bus B. The second operation now waits for ages (for rather silly reasons).

I'll restructure this more carefully.

rswarbrick:
Take 2 :-) I'm more convinced about this approach.

nasahlpa (approved):
Thanks that is really elegant :-)

PR #25613, opened 2024-12-11
[dv] Try harder for late randomisation in cip_base_vseq

rswarbrick:
This builds on the change in #25612. The first commit in this PR makes sure that late randomisation of CSR accesses actually happens late enough (if lots of CSR accesses get enqueued at the same time). The second commit is a slight cleanup, actually *doing* later randomisation for the TL sequences that actually do the CSR reads/writes. I'm not sure *that* will make much difference at the moment, but it seems sensible to be cleaner.


marnovandermaas (approved):
Looks good to me.

rswarbrick:
Hmm. I'm not completely certain what drove me to write this in December and the implementation here definitely doesn't work (because the "number of CSR accesses in flight" isn't split per interface). I'm going to drop this and come back to it afresh if necessary.

PR #25616, opened 2024-12-11
[rtl, aon_timer] Reset core.prescale_count on wkup_ctrl write

antmarzam:
The AON timer core has an internal counter `prescale_count_q` which is not reset upon each write to wkup_ctrl.
This causes issues if the WKUP timer is re-configured to a lower prescaler value than it was before, and the `prescale_count_q` variable is greater than the new prescaler value, forcing the `prescale_count_q` counter to overflow itself causing a non-spec compliant behavior.

In order to mitigate the above, the internal `prescale_count_q` is reset after wkup_ctrl write. Thus, the aon timer is forced to count from scratch the new intended prescaler value.

Docs PR is #25904

### DIFF
--- a/hw/ip/aon_timer/rtl/aon_timer_core.sv
+++ b/hw/ip/aon_timer/rtl/aon_timer_core.sv
@@ -45,6 +45,8 @@ module aon_timer_core import aon_timer_reg_pkg::*; (
   always_ff @(posedge clk_aon_i or negedge rst_aon_ni) begin
     if (!rst_aon_ni) begin
       prescale_count_q <= 12'h000;
+    end else if (reg2hw_i.wkup_ctrl.prescaler.qe) begin
+      prescale_count_q <= 12'h000;
     end else if (prescale_en) begin
       prescale_count_q <= prescale_count_d;
     end
--- a/hw/ip/aon_timer/rtl/aon_timer_reg_pkg.sv
+++ b/hw/ip/aon_timer/rtl/aon_timer_reg_pkg.sv
@@ -24,6 +24,7 @@ package aon_timer_reg_pkg;
   typedef struct packed {
     struct packed {
       logic [11:0] q;
+      logic        qe;
     } prescaler;
     struct packed {
       logic        q;
@@ -124,8 +125,8 @@ package aon_timer_reg_pkg;
 
   // Register -> HW type
   typedef struct packed {
-    aon_timer_reg2hw_alert_test_reg_t alert_test; // [247:246]
-    aon_timer_reg2hw_wkup_ctrl_reg_t wkup_ctrl; // [245:233]
+    aon_timer_reg2hw_alert_test_reg_t alert_test; // [248:247]
+    aon_timer_reg2hw_wkup_ctrl_reg_t wkup_ctrl; // [246:233]
     aon_timer_reg2hw_wkup_thold_hi_reg_t wkup_thold_hi; // [232:201]
     aon_timer_reg2hw_wkup_thold_lo_reg_t wkup_thold_lo; // [200:169]
     aon_timer_reg2hw_wkup_count_hi_reg_t wkup_count_hi; // [168:137]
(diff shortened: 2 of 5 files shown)


hw/ip/aon_timer/rtl/aon_timer_core.sv
```diff
@@ -43,7 +43,7 @@ module aon_timer_core import aon_timer_reg_pkg::*; (
                             lc_ctrl_pkg::lc_tx_test_false_strict(lc_escalate_en_i[0]);
 
   always_ff @(posedge clk_aon_i or negedge rst_aon_ni) begin
-    if (!rst_aon_ni) begin
+    if (!rst_aon_ni | reg2hw_i.wkup_ctrl.prescaler.qe) begin
```
Razer6, line 46: Please do not put logic in the reset path. This has to be cleared in a dedicated combinational path.

antmarzam: Thanks for the feedback! I've moved the condition to its own `else if` now

hw/ip/aon_timer/rtl/aon_timer_reg_pkg.sv
```diff
@@ -24,6 +24,7 @@ package aon_timer_reg_pkg;
   typedef struct packed {
     struct packed {
       logic [11:0] q;
+      logic        qe;
```
marnovandermaas, line 27: Isn't this generated using reggen? Should we change the hjson file for this instead?

hw/ip/aon_timer/rtl/aon_timer_reg_top.sv
```diff
@@ -644,7 +644,7 @@ module aon_timer_reg_top (
     .d      ('0),
 
     // to internal hardware
-    .qe     (),
+    .qe     (reg2hw.wkup_ctrl.prescaler.qe),
```
marnovandermaas, line 647: This is also generated by reggen right?

marnovandermaas:
Can you try this instead to fix the lint errors:
```
{ name: "WKUP_CTRL",
      desc: "Wakeup Timer Control register",
      swaccess: "rw",
      hwaccess: "hro",
      async: "clk_aon_i",
      fields: [
        { bits: "0",
          name: "enable",
          desc: "When set to 1, the wakeup timer will count",
        }
        { bits: "12:1",
          name: "prescaler",
          desc: "Pre-scaler value for wakeup timer count",
          hwqe: "true",
        }
      ],
    },
```

antmarzam:
> I think the RTL now looks right to me. Would you mind adding some documentation on the new behavior. Specifically that any write to the wkup_ctrl register now resets the prescalar including the enable bit.

Yep - I'll create a PR adding this new behavior to the doc. Will add you as a reviewer and link here.

Also, just pushed the changes adding `hwqe = true` to the field only. The behaviour is the same just that `hw2reg.wkup_ctrl.enable.qe` is now not connected.

Razer6 (approved):
HW looks good to me.

antmarzam:
@marnovandermaas I've moved the doc commit here as you suggested

marnovandermaas (approved):
This looks correct to me, thanks for pursuing this fix.

marnovandermaas:
CHANGE AUTHORIZED: hw/ip/aon_timer/data/aon_timer.hjson
CHANGE AUTHORIZED: hw/ip/aon_timer/rtl/aon_timer_core.sv
CHANGE AUTHORIZED: hw/ip/aon_timer/rtl/aon_timer_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/aon_timer/rtl/aon_timer_reg_top.sv

rswarbrick (approved):
This looks sensible to me. Thanks for sorting it out. (I wrote two versions of "isn't this wrong because...", and then realised that the change is correct!)

rswarbrick:
CHANGE AUTHORIZED: hw/ip/aon_timer/data/aon_timer.hjson
CHANGE AUTHORIZED: hw/ip/aon_timer/rtl/aon_timer_core.sv
CHANGE AUTHORIZED: hw/ip/aon_timer/rtl/aon_timer_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/aon_timer/rtl/aon_timer_reg_top.sv

The only change to behaviour comes from the change to `aon_timer_core.sv` in the first commit. This looks sensible and is tested (indeed, that's why the PR was needed!), so the risk to the taped out design is very low.

PR #25619, opened 2024-12-12
[pentest] Update EDN SCA tests

nasahlpa:
This commit updates the EDN SCA tests to use asm instead of C code such that the SCA target is more accurate.


vogelpi (commented):
Thanks for the PR @nasahlpa . The code looks good to me but I have a question regarding the analysis method:

Before this PR we had a fixed-vs-random command wheres with this PR we only have the command for measuring the transmission of one word over the bus. I assume we then do the trace grouping during the analysis based on individual subwords or even bits we want to analyze. Can you comment on this please?

nasahlpa:
Yes exactly, the analysis needs to be done at the host site. The reason for this change is that the lab decided that this will fit their setup better.

vogelpi (approved):

PR #25623, opened 2024-12-12
[ci] use GitHub Actions to build documentation site

nbdd0121:
The current Cloud Build setup is broken and is quite convoluted to fix. It's much easier to just use GitHub Actions to build and push to the site.

Demo run: https://github.com/lowRISC/opentitan/actions/runs/12285799612/job/34284759154


.github/workflows/ci.yml
```diff
@@ -123,6 +121,51 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Bitstream cache requires all commits.
+      - name: Prepare environment
+        uses: ./.github/actions/prepare-env
+        with:
+          service_account_json: '${{ secrets.BAZEL_CACHE_CREDS }}'
+      - name: Build documentation
+        run: util/site/build-docs.sh build
+      - uses: google-github-actions/setup-gcloud@v2
```
jwnrt, line 140: I think this conditional section where we do the upload only for `master` pushes is quite complex and would benefit from moving to a composite action. What do you think?

jwnrt (approved):
LGTM and the demo run looks correct. Thanks

.github/workflows/ci.yml
jwnrt: I mean the whole thing is quite complex. It's 28 lines with those four mixed in that have to be kept in sync.

nbdd0121: What do you mean by kept in sync?

jwnrt: When this is backported all four `master`s need to be changed

nbdd0121: How about:
* Make setup-gcloud be part of prepare-env and unconditionally executed for non-pull_request triggers (a couple of seconds, but only for pushes)
* Merge all bash steps into 1 so there's only single line with `master` as condition.

When backporting we should change the action on master to account for branch name and then backport as a whole.

jwnrt: Sure, that sounds fine

jwnrt (approved):

PR #25624, opened 2024-12-12
[bazel] Check Bazel version when found in PATH

jwnrt:
The `bazelisk.sh` script will try to use a real Bazel from the `PATH` if it can find it, but doesn't check its version against `.bazelversion`.

This change checks the version and falls back to downloading `bazelisk` if it doesn't match.


jadephilipoom (approved):
This worked locally for me, and fixed the (incompatible) Bazel 8.0.0 fetch I was seeing before. Thanks!

pamaury (approved):
I'm not a bash expert but this looks good to me.

PR #25626, opened 2024-12-12
[meridiancdc] Include SDC results in setup checks report

andreaskurth:
Without the `-scenario` argument, `report_policy` only reports checks for the current scenario, which is the ENV file at this point in the script.  By adding `sdc`, also SDC setup check results get reported.


vogelpi (approved):
LGTM

PR #25630, opened 2024-12-12
[manuf] enable toggling external clock on/off during FT individualize

timothytrippel:
This updates the FT individualize program to be able to select whether the external clock is enabled when running on silicon targets.


timothytrippel:
CC: @benjbender

moidx (approved):

PR #25632, opened 2024-12-13
[i2c] Add detection of SDA stuck-at-zero faults

a-will:
For multi-controller use cases, add a bus timeout condition that triggers when SCL has been held high and SDA has been held low for the bus timeout duration. This helps with detecting stuck-at-zero faults, following the recommendations of the SMBus spec. The duration is a bit longer than the SCL bus timeout, since the single counter is used for detecting idle clock first, then adds a fresh count to detect SDA stuck low.


a-will:
@hcallahan-lowrisc I tagged you for review, since this might still be interesting to someone. However, if this is to go in, somebody probably needs to take it over and make sure the implementation isn't dumb ;)

PR #25640, opened 2024-12-13
[sram_ctrl,dv] Increase wait_cycles for rand_reset tests

nasahlpa:
In the sram_dv_increase_timeout_reset test, we are waiting to issue a reset when no outstanding TL-UL are in the fly. This wait timed out in some of the tests. Increasing the wait timer lets the test pass.


rswarbrick:
This is kind of reasonable, I suppose but it's not beautiful (because the "wait until we're not doing anything before we issue a reset" logic is rather silly!)

@martin-velay knows quite a bit about this too, so it might be worth having a chat with him: there's probably a cleaner way to do this, and the resulting sequences will be quite a bit shorter.

rswarbrick (approved):
I think there's probably a better way to do this, but fixing the timeouts seems a sensible intermediate step.

nasahlpa:
> This is kind of reasonable, I suppose but it's not beautiful (because the "wait until we're not doing anything before we issue a reset" logic is rather silly!)
> 
> @martin-velay knows quite a bit about this too, so it might be worth having a chat with him: there's probably a cleaner way to do this, and the resulting sequences will be quite a bit shorter.

Yes I agee this requires more work in the future. Something similar is in this PR #25647 .

PR #25643, opened 2024-12-13
[ci] Move CW305 bitstream build to machine without license

jwnrt:
The CW305 uses an Artix-7 FPGA which does not require an enterprise Vivado license to synthesise for. We limit the number of machines on the `-bitstream` label to match our number of licenses, however CW305s are taking up these spots unnecessarily.

This commit moves them to the `-vivado` machines which have access to Vivado Standard edition only.


nbdd0121 (approved):
The vivado machine still have access to the license server, although it's on a different pool so wouldn't compete with bitstream builders.

jwnrt:
Ah, thanks for the clarification

PR #25645, opened 2024-12-13
[Opentitantool] Add a ftdi transport to the opentitanlib.

engdoreis:
This PR adds support for the Voyager board by adding a ftdi transport to opentitantool.

The ftdi transport may also be useful for the CW340.

Note: Voyager is a simple board lowRISC and NewAE designed to demonstrate the OpenTitan Earl Grey chip at conferences and other events.


engdoreis:
> Curious... What is the Voyager board? How is it associated with this project?

Sorry for letting that unexplained. I updated the description.

nasahlpa:
just wondering - what is the status of this PR? Please let me know if I can support you @engdoreis, I soon would need to have an `opentitantool` that works with Voyager.

engdoreis:
@nasahlpa is ready for review now.

thanks @jwnrt for letting me know.

sw/host/opentitanlib/src/transport/ftdi/gpio.rs
```diff
@@ -0,0 +1,83 @@
...
+        Ok(Self {
+            pin: Rc::new(RefCell::new(pin)),
+            pinname,
+        })
+    }
+}
+
+impl GpioPin for Pin {
+    fn read(&self) -> Result<bool> {
+        Ok(false)
```
jwnrt, line 60: Are you sure these should silently fail? Shouldn't they return some kind of `Error::Unsupported` instead?

engdoreis: I have implemented the read function.

jwnrt (approved):
This is looking good, thanks.

PR #25646, opened 2024-12-13
tcp_server_create changes

kwalker27:
I'm proposing to make it a fatal error to fail to open the listening socket rather than simply continuing the simulation after printing an error. My rationale is that it would be better to intentionally open the socket (and optionally provide a mechanism to make the port unique, as was recently done for `jtagdpi`), instead of the current "best effort" approach.

The changes here: (1) make the socket setup synchronous to the `create()` call while leaving the processing on a dedicated pthread and (2) assert if the setup fails.

I expect there could be fallout from this (in CI sim jobs, for example) since `dmidpi` currently starts the server unconditionally and uses a static port number (set via SV parameter). So, I expect there might need to be additional changes after discussion whether folks consider whether this policy change is sensible or not.


kwalker27:
The regression failure is what I was afraid of:

```
dmi0: Failed to bind socket: Address already in use (98)
dmi0: Unable to create TCP server on port 44853
```

So, I'd like to hear opinions on whether the suggestion in this PR makes sense or not - and if it does, this work needs to be deferred until (1) dmidpi server setup is made conditional and (2) port collisions are avoided somehow (dynamic port selection?) when a `dmidpi` client needs to be supported.

In the current state of things, a user running a single sim on a local machine gets the benefit of `dmidpi` always being there. But as soon as multiple sims are running on a machine, you can't guarantee the process you want to connect to will have won the race to claim the port. So I'd argue _some_ sort of change to the usage model is worth considering.

andreaskurth (commented):
I think this is sensible, though I admit this isn't my core area of expertise. Maybe @rswarbrick has experience or @luismarques can ask a C expert from the SW team to review this as well?

andreaskurth:
@kwalker27, I assume you're successfully using this change internally already?

kwalker27:
@andreaskurth yes, I've tested it locally but it's not deployed because I wanted to get through upstream review first. But using the `active` port on `jtagdpi` we can let the socket conflict be an error (and we don't instantiate `dmidpi` at all).

hw/dv/dpi/common/tcp_server/tcp_server.c
```diff
@@ -326,7 +326,10 @@ static void *server_create(void *ctx_void) {
   if (rv != 0) {
     fprintf(stderr, "%s: Unable to create TCP server on port %d\n",
             ctx->display_name, ctx->listen_port);
-    goto err_cleanup_return;
+    // Failing to create the listening socket is treated as a fatal
+    // error. If the creation of this socket is not important, it
+    // should not even be attempted.
+    assert(false);
```
rswarbrick: This is a genuine question (rather than a review comment), but why is it better to use `assert(false)` instead of e.g. `exit(1)` here?

On the review side: this looks completely sensible to me.

rswarbrick (approved):
This all looks good to me. Thanks for the carefully explained commits.

I'll give it until tomorrow to let @luismarques take a look too (on the basis that he knows rather a lot more about SW than me!)

hw/dv/dpi/common/tcp_server/tcp_server.c
a-will: Just piling on a little bit here, but the previous writers for this file got this wrong. `assert()` should _never_ be used for run time error handling. Its purpose is for _extra_ consistency checks, for scenarios that "can't" happen and thus can be removed with `-DNDEBUG`.

`exit()` would likely be the most appropriate, since this error still allows for an orderly program termination. If this were a bad bug in the program, then `abort()` would be the correct option.

andreaskurth: Thanks for pointing this out. @kwalker27 would you mind changing that, please?

kwalker27: Updated and rebased.

rswarbrick (dismissed):
Thanks: this looks good to me!

kwalker27:
> Thanks: this looks good to me!

OK, but I guess it can't get through CI as long as `dmidpi` is unconditionally activated and runs the risk of hitting the fatal port conflict.

rswarbrick:
I think you're right about the CI problem. Do you have an idea of how we might get it working?

kwalker27:
Yeah, this is the sticky part of my proposed change. I don't know what users of the `dmidpi` there are, and if it would be painful to switch it from being (1) always active and (2) having a static port - to being conditional based on some runtime flag and support more flexible port assignment. This is what I did a little while back for `jtagdpi`.

kwalker27:
But, also, if any CI tests actually relied on `dmidpi` I expect there would be occasional failures if multiple jobs landed on the same runner and only one of them (the "wrong" one?) won the race for the static port....

PR #25649, opened 2024-12-13
[rom_ctrl, dv] Assertion coverage collection for tlul_adapter_sram

KinzaQamar:
We are not required to collect assertion coverage for prim_fifo_sync instantiated in tlul_adapter_sram. Commit 0372c1f removed the assert coverage for these modules but still got assertion coverage holes reported by VCS.

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/cover_reg_top.cfg
+++ b/hw/ip/rom_ctrl/dv/cov/cover_reg_top.cfg
@@ -2,9 +2,13 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-// Collect coverage for tlul_adapter_sram to resolve coverage hole for intg_err.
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
I'm trying to make sure that I understand what's going on here. Is the point that line 6 turns on all coverage types for `tb.dut.u_tl_adapter_rom`, which implies assertion coverage for `prim_fifo_sync`?

Ignoring assertion coverage, how does the meaning of line 6 differ from the stuff that's at the bottom of this file? (Can we drop one of them and/or enclose everything in a set of coverage types?)

KinzaQamar:
Thanks for the review.

I agree that the exclusion file did contain some redundant lines. Fixed that by enabling all the coverage except assert for the adapter tree and then later disabled assertion coverage for the fifos in the adapter.

rswarbrick (approved):
Looks sensible to me: thanks!

PR #25650, opened 2024-12-13
[bazel] Migrate `rules_python` and dependents to bzlmod

jwnrt:
This PR migrates `rules_python` and the modules that depend on it to bzlmod.

These depending modules use python interpreters in their actual rule implementations, so they have to be migrated at the same time. The first commits add bzlmod support for each module before they're all switched over in the second-to-last commit.

`bazel_skylib` must also be migrated as it's needed for the Tock repositories (via `bare_repository`).

There is also a fix for the `INCLUDE` paths of linker scripts inside the `external` directory which is unstable and has to have its paths generated by Bazel.

The modules migrated are:

* `bazel_skylib`.
* `lowrisc_misc_linters` (license checker).
* `rules_fuzzing`.
* `rules_python`.


third_party/tock/repos.bzl
```diff
@@ -289,9 +291,8 @@ def tock_repos(tock = None, libtock = None, elf2tab = None):
         },
     )
 
-    http_archive_or_local(
+    http_archive(
         name = "elf2tab",
-        local = elf2tab,
```
jwnrt, line 293: We're not losing functionality here, bzlmod has a first-class feature for overriding repositories:

https://bazel.build/rules/lib/globals/module#override_repo

pamaury (approved):
This looks great. One small nit but I'll leave it up to you: would it make sense to split `[bazel] Fix external includes in linker scripts` into two: the bazel change, and the use of this change to fix linker scripts?

jwnrt:
Yeah they could be split. I think it's unlikely I'll need to revert just one so I'm fine with leaving it as one

PR #25655, opened 2024-12-15
[alert, sival] chip_sw_alert_handler_escalation

andrea-caforio:
This PR generalizes the existing `alert_handler_escalation_test` to all simulation targets to cover the
`chip_sw_alert_handler_escalation` testplan item (as specified in `hw/top_earlgrey/data/ip/chip_alert_handler_testplan.hjson`).

### DIFF
--- a/sw/device/tests/alert_handler_escalation_test.c
+++ b/sw/device/tests/alert_handler_escalation_test.c
@@ -23,18 +23,6 @@
 #include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
 #include "rv_core_ibex_regs.h"  // Generated.
 
-/*
-  - Verify the first escalation results in NMI interrupt serviced by the CPU.
-  - Verify the second results in device being put in escalate state, via the LC
-  JTAG TAP.
-  - Verify the third results in chip reset.
-  - Ensure that all escalation handshakes complete without errors.
-
-  The first escalation is checked via the entry of the NMI handler and polling
-  by dv. The second escalation is directly checked by dv. The third escalation
-  is checked via reset reason.
- */
-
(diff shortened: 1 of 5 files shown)


sw/device/tests/alert_handler_escalation_test.c
```diff
@@ -114,24 +91,79 @@ bool test_main(void) {
...
+  uint32_t phase0DurationCycles = 1000000;
+  if (kDeviceType == kDeviceSimVerilator) {
+    phase0DurationCycles /= 10;
+  } else if (kDeviceType == kDeviceSimDV) {
+    phase0DurationCycles /= 100;
+  }
+
+  dif_alert_handler_escalation_phase_t escalationProfiles[] = {
+      {.phase = kDifAlertHandlerClassStatePhase0,
+       .signal = 0,
```
jwnrt, line 115: @pamaury for multitop we might need to make these signal codes constants in the top since they're used semantically (i.e. we write 0 to mean NMI, 1 to mean escalation LC state, 3 to mean chip reset). Tops could change the meanings of these numbers.

jwnrt (approved):
This looks good to me, thanks @andrea-caforio

engdoreis (approved):

sw/device/tests/alert_handler_escalation_test.c
pamaury: Thanks for the reminder @jwnrt . I am aware of this problem and we discussed it with @matutem in the past. I don't remember the answer but I think it wasn't clear how/if this is configured in the Hjson or just how we turn out to implement it. In any case this is definitely something we want for multitop if we can abstract it nicely.

PR #25664, opened 2024-12-16
[RACL] Implement config parsing and racl_ctrl

Razer6:
This is the initial PR implementing RACL and consists of two commits:

1. Parse the RACL configuration file that is passed int the top-level HJSON file and render the `top_racl_pkg.sv` from it. If no RACL is used in the top, i.e., there is no RACL config passed, a default `top_racl_pkg.sv` is rendered to provide the minimal type information for all RACL usages in IPs. This package is a virtual core, and different Tops can generate their own version.

2. Implement the ipgen'ed `racl_ctrl` policy distributing IP. This IP renders all policies of a given policy group that is defined in the `racl.hjson` file. Currently, topgen only supports rendering the first RACL group's `racl_ctrl` IP due to some tooling limitations.


util/topgen.py
```diff
@@ -526,6 +527,40 @@ def generate_ac_range_check(topcfg: Dict[str, object], out_path: Path) -> None:
...
+        # If we have more RACL policy groups, uniquify the control IP
+        # if len(topcfg['racl']['policies']) > 1:
+        #     params['module_instance_name'] = f'racl_ctrl_{racl_group}'
+
+        # # Only render the RACL groups that are really instantiated in that top
+        # for m in topcfg['module']:
+        #     if m['name'] == params['module_instance_name']:
+        #         ipgen_render("racl_ctrl", topname, params, out_path)
+        #         break
+        # TODO: The obove code, would be the correct if ipgen correctly supports rendering multiple
```
andreaskurth, line 557: Do we already have an issue to track this limitation? If yes, please link it. If no, please create one and then link it :-)

hw/ip_templates/racl_ctrl/data/racl_ctrl.hjson.tpl
```diff
@@ -0,0 +1,163 @@
...
+  one_line_desc:      "Implements the RACL policy registers to distribute to subscribing IPs.",
+  one_paragraph_desc: '''
+  Implements the RACL policy registers to distribute to subscribing IPs.
+  '''
+  // Unique comportable IP identifier defined under KNOWN_CIP_IDS in the regtool.
+  cip_id:             "43",
+  design_spec:        "../doc",
+  dv_doc:             "../doc/dv",
+  hw_checklist:       "../doc/checklist",
+  sw_checklist:       "/sw/device/lib/dif/dif_rstmgr",
```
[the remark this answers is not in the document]
Razer6: Removed for now as there is no DIF yet.

hw/ip/dma/lint/dma.waiver
```diff
@@ -7,4 +7,4 @@
 waive -rules {ARITH_CONTEXT} -location {dma.sv} -regexp {.*Bitlength of arithmetic operation.*is self-determined in this context.*} \
       -comment "Carry issue when determining the bit length."
 waive -rules {HIER_NET_NOT_READ} -location {dma_reg_top.sv} -regexp {status_flds_we\[0|4\]' is not read from in module} \
-      -comment "Internal register is accepted to not be read. Tracked in #702."
+      -comment "Internal register is accepted to not be read. Tracked in #25663."
```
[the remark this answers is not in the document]
Razer6: I can move that out. I run into the same issue and knew that I already filed one, but this was in the old integrated repo. Lets remove that.

hw/ip_templates/racl_ctrl/data/racl_ctrl.tpldesc.hjson
```diff
@@ -0,0 +1,49 @@
...
+{
+  template_param_list: [
+    {
+      name: "topname"
+      desc: "Name of top-level design, e.g., 'darjeeling' or 'earlgrey'"
+      type: "string"
+      default: ""
+    }
+    {
+      name: "module_instance_name"
```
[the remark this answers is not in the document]
Razer6: `module_instance_name` is already used by the tooling to implement the uniquification. Changing that in topgen and ipgen is probably a different issue and PR.

hw/ip_templates/racl_ctrl/data/racl_ctrl.hjson.tpl
```diff
@@ -0,0 +1,163 @@
...
+          desc: '''
+                RACL source issuing the error.
+                '''
+        }
+        { bits: "${3 + nr_racl_bits + nr_ctn_uid_bits - 1}:${3 + nr_racl_bits}"
+          name: "ctn_uid"
+          resval: 0x0
+          desc: '''
+                CTN UID issuing the error.
+                '''
```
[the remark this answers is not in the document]
Razer6: I think the source_id is not the right thing, as it tracks only the request number if it supports multiple outstanding transfers. There would be no mapping from the integer to the device causing the error. There are 2 options:
1. Gate rendering that information via the racl.hjson
2. Define a device UID

2 probably makes sense in the long term. Short term, I would go for option 1.

hw/ip_templates/racl_ctrl/rtl/racl_ctrl.sv.tpl
```diff
@@ -0,0 +1,183 @@
...
+  // Bus Interface (device)
+  input  tlul_pkg::tl_h2d_t                                    tl_i,
+  output tlul_pkg::tl_d2h_t                                    tl_o,
+  // Alerts
+  input  prim_alert_pkg::alert_rx_t [NumAlerts-1:0]            alert_rx_i,
+  output prim_alert_pkg::alert_tx_t [NumAlerts-1:0]            alert_tx_o,
+  // Output policy vector for distribution
+  output top_racl_pkg::racl_policy_vec_t                       policies_o,
+  // RACL violation information.
+  input logic            [NumSubscribingIps-1:0]               racl_error_i,
```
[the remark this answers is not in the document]
Razer6: Let's deal with CDC later. I'll create an issue for both cases Error/Policy dsitribution.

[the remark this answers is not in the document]
Razer6: https://github.com/lowRISC/opentitan/issues/25672

hw/ip_templates/racl_ctrl/rtl/racl_ctrl.sv.tpl
```diff
@@ -0,0 +1,183 @@
...
+  input logic                                                  rst_shadowed_ni,
+% endif
+  // Bus Interface (device)
+  input  tlul_pkg::tl_h2d_t                                    tl_i,
+  output tlul_pkg::tl_d2h_t                                    tl_o,
+  // Alerts
+  input  prim_alert_pkg::alert_rx_t [NumAlerts-1:0]            alert_rx_i,
+  output prim_alert_pkg::alert_tx_t [NumAlerts-1:0]            alert_tx_o,
+  // Output policy vector for distribution
+  output top_racl_pkg::racl_policy_vec_t                       policies_o,
```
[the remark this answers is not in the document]
Razer6: https://github.com/lowRISC/opentitan/issues/25672

util/topgen.py
Razer6: Done https://github.com/lowRISC/opentitan/issues/25673

Razer6:
@andreaskurth Thanks for the initial feedback. I changed the docs or created issues for it. Can you take another look?

util/raclgen/lib.py
```diff
@@ -0,0 +1,51 @@
...
+    try:
+        with open(config_file, 'r') as f_racl_config:
+            racl_config = hjson.load(f_racl_config)
+    except OSError:
+        raise SystemExit(sys.exc_info()[1])
+
+    # TODO further error handling
+    error = check_keys(racl_config, racl_required, [], [], 'RACL Config')
+    if error:
+        raise SystemExit(f"Error occured while validating {config_file}")
```
rswarbrick, line 34: I'd strongly suggest putting a tiny bit more type checking in here. Something like extracting `roles` and `policies` as a list / dictionary, respectively (doing a basic type check) and maybe checking the item and key/value types for them respectively.

Extracting that properly will get rid of the dictionary lookups lower down in the function.

util/raclgen/lib.py
```diff
@@ -0,0 +1,51 @@
...
+
+# Default configuration to render the RACL package for systems that don't use RACL but need the
+# type definitions
+DEFAULT_RACL_CONFIG = {
+    'nr_policies': 1,
+    'policies': {},
+}
+
+
+def parse_racl_config(config_file: str):
```
rswarbrick, line 24: I'd be tempted to call this something like `config_path`, just to avoid a question about whether this is an open file object.

util/topgen.py
```diff
@@ -526,6 +527,15 @@ def generate_ac_range_check(topcfg: Dict[str, object], out_path: Path) -> None:
     ipgen_render("ac_range_check", topname, params, out_path)
 
 
+# Generate RACL collateral
+def generate_racl(topcfg: Dict[str, object], out_path: Path) -> None:
+    # Not all tops use RACL
+    if 'racl_config' not in topcfg:
+        return
```
rswarbrick, line 534: Wouldn't it make sense to set `topcfg['racl']` to `None` in this situation?

util/topgen.py
```diff
@@ -1249,6 +1262,12 @@ def render_template(template_path: str, rendered_path: Path,
                         out_path / f"rtl/autogen/{top_name}_rnd_cnst_pkg.sv",
                         gencmd=gencmd)
 
+        racl_config = completecfg['racl'] if 'racl' in completecfg else DEFAULT_RACL_CONFIG
```
rswarbrick, line 1292: Probably cleaner: `completecfg.get('racl', DEFAULT_RACL_CONFIG)`

util/topgen/templates/toplevel_racl_pkg.sv.tpl
```diff
@@ -0,0 +1,82 @@
...
+  // Number of CTN UID bits transferred
+  parameter int unsigned NrCtnUidBits = 8;
+
+  // RACL role type binary encoded
+  typedef logic [NrRaclBits-1:0] racl_role_t;
+
+  // CTN UID assigned the bus originator
+  typedef logic [NrCtnUidBits-1:0] ctn_uid_t;
+
+  // RACL permission and one-hot encoded role vector
```
rswarbrick, line 22: If I understand correctly, a RACL permission *is* a one-hot encoded role vector. Maybe this would be clearer as something like "RACL permission: a one-hot encoded role vector".

util/topgen/templates/toplevel_racl_pkg.sv.tpl
```diff
@@ -0,0 +1,82 @@
...
+  typedef logic [(2**NrRaclBits)-1:0] racl_role_vec_t;
+
+  // RACL policy vector containing a read and write permission
```
rswarbrick, line 25: I'm not sure that "policy vector" is the right name here: there are two vectors...

util/topgen/templates/toplevel_racl_pkg.sv.tpl
```diff
@@ -0,0 +1,82 @@
...
+
+  // RACL policy vector for distributing RACL policies from the RACL widget to the subscribing IP
+  typedef racl_policy_t [NrRaclPolicies-1:0] racl_policy_vec_t;
+
+  // RACL information logged in case of a denial
+  typedef struct packed {
+    racl_role_t racl_role;
+    ctn_uid_t   ctn_uid;
+    // 0: Write access, 1: Read access
+    logic       write_read;
```
rswarbrick, line 39: Not a big deal if this name appears in lots of other places. But something like `read_not_write` might be more obvious at a glance?

hw/ip_templates/racl_ctrl/doc/interfaces.md.tpl
```diff
@@ -0,0 +1,4 @@
+${"#"} Hardware Interfaces
+
+<!-- BEGIN CMDGEN util/regtool.py --interfaces ./hw/top_${topname}/ip_autogen/racl_ctrl/data/racl_ctrl.hjson -->
```
rswarbrick, line 3: Are there strange characters before "interfaces"? (Github's diff view shows them in red)

hw/ip_templates/racl_ctrl/lint/racl_ctrl.vlt.tpl
```diff
@@ -0,0 +1,5 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+//
+// waiver file for ${module_instance_name}
```
rswarbrick, line 5: Are you expecting to need Verilator waivers for the generated modules? If not, I'd be tempted to drop this file.

hw/ip_templates/racl_ctrl/lint/racl_ctrl.waiver.tpl
```diff
@@ -0,0 +1,8 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+#
+# waiver file for ${module_instance_name}
+
+waive -rules {HIER_NET_NOT_READ} -location {${module_instance_name}_reg_top.sv} -regexp {error_log_flds_we\[4:1\]' is not read from in module} ${"\\"}
+      -comment "Internal register is accepted to not be read. Tracked in #25663."
```
rswarbrick, line 8: Missing newline at the end of the file.

util/topgen.py
```diff
@@ -532,9 +532,36 @@ def generate_racl(topcfg: Dict[str, object], out_path: Path) -> None:
...
+        #     params['module_instance_name'] = f'racl_ctrl_{racl_group}'
+
+        # # Only render the RACL groups that are really instantiated in that top
+        # for m in topcfg['module']:
+        #     if m['name'] == params['module_instance_name']:
+        #         ipgen_render("racl_ctrl", topname, params, out_path)
+        #         break
+        # TODO(#25673): The obove code, would be the correct if ipgen correctly supports rendering
+        # multiple instances and allow topgen to instantiate right now. This support is not yet
+        # implemented properly. Therefore, simple render the first RACL group to the RACL control
```
rswarbrick, line 560: Nit: I think you want "simply" here.

util/topgen/templates/toplevel_racl_pkg.sv.tpl
Razer6: Let's drop that to policy.

hw/ip_templates/racl_ctrl/lint/racl_ctrl.vlt.tpl
Razer6: Let's drop that file for now and introduce it if needed later.

hw/ip_templates/racl_ctrl/doc/interfaces.md.tpl
Razer6: It seems it gets confused with being a diff control statement. Nothing in the source file.

util/reggen/reg_top.sv.tpl
```diff
@@ -115,7 +117,13 @@
 %>
 `include "prim_assert.sv"
 
-module ${mod_name} (
+module ${mod_name}${' (' if not racl_support else ''}
```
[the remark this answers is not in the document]
Razer6: The only reason for that was to keep the diff 0 when not enabling RACL on one IP. Otherwise we would generate a small whitespace diff for _alll_ regtops.

util/reggen/bus_interfaces.py
```diff
@@ -69,6 +72,9 @@ def from_raw(raw: object, where: str) -> 'BusInterfaces':
             hier_path = check_optional_str(ed.get('hier_path'),
                                            'hier_path field of ' + entry_what)
 
+            racl_support = check_optional_bool(ed.get('racl_support'),
+                                               'racl_support field of ' + entry_what)
```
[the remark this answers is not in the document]
Razer6: Well, the only reason was that `racl_support` is optional and not available in all definitions. Therefore the `optional` version. What do you think?

util/topgen.py
Razer6: The reason this was _not_ added is to make it easier to detect if no RACL is used, i.e., down below: `completecfg.get('racl', DEFAULT_RACL_CONFIG)`

util/raclgen/lib.py
Razer6: Created https://github.com/lowRISC/opentitan/issues/25690 to track this.

util/raclgen/lib.py
```diff
@@ -0,0 +1,51 @@
...
+
+# Default configuration to render the RACL package for systems that don't use RACL but need the
+# type definitions
+DEFAULT_RACL_CONFIG = {
+    'nr_policies': 1,
+    'policies': {},
+}
+
+
+def parse_racl_config(config_path: str):
```
matutem, line 24: Coud you add a return type, probably `...str) -> Dict[str, object]:` or some more specific type for the value in the Dict?

util/raclgen/lib.py
```diff
@@ -0,0 +1,51 @@
...
+    try:
+        with open(config_path, 'r') as f_racl_config:
+            racl_config = hjson.load(f_racl_config)
+    except OSError:
+        raise SystemExit(sys.exc_info()[1])
+
+    # TODO further error handling
```
matutem, line 31: Some hint about the missing checks?

util/raclgen/lib.py
```diff
@@ -0,0 +1,51 @@
...
+    if error:
+        raise SystemExit(f"Error occurred while validating {config_path}")
+
+    # Determine the maximum number of policies over all RACL groups for RTL
+    # RTL needs to create the vectors based on the largest group
+    racl_config['nr_policies'] = max(len(policies) for policies in racl_config['policies'].values())
+
+    for racl_group, policies in racl_config['policies'].items():
+        for policy in policies:
+            def compute_policy_value(permission: str):
```
matutem, line 42: Seems like the return value type delaration could be `... str) -> int:`

util/topgen.py
matutem: Detection of no RACL could be as  simple as `racl not in completecfg`, and we could have no trace of racl if it is not used. Perhaps the package could be avoided in those cases to keep it simple.

matutem (commented):
I thought the idea was to avoid any trace of RACL if it was not used in some top. Why do we need it at all in earlgrey for example?

Razer6:
It was discussed the we always create the ports and tie them off at reg_top level

util/raclgen/lib.py
Razer6: I changed the comment a little bit and linked the issue for that. https://github.com/lowRISC/opentitan/issues/25690

hw/ip_templates/racl_ctrl/README.md
```diff
@@ -0,0 +1 @@
+# RACL Control Permission IP
```
andreaskurth, line 1: Missing doc tracked in #25746

andreaskurth (approved):

PR #25665, opened 2024-12-16
[topgen] Deepcopy parameter when mangeling

Razer6:
Otherwise, the same parameter of a difference instance would be overwritten.

This fixes the issue in https://github.com/lowRISC/opentitan/pull/25625#discussion_r1882240404, where the same parameter was used in multiple parametrized port definitions.

### DIFF
--- a/util/topgen/intermodule.py
+++ b/util/topgen/intermodule.py
@@ -778,8 +778,20 @@ def check_intermodule(topcfg: Dict, prefix: str) -> int:
             err, rsp_struct = check_intermodule_field(rsp_struct)
             error += err
 
-            total_width += rsp_struct["width"]
-            widths.append(rsp_struct["width"])
+            if isinstance(rsp_struct["width"], Parameter):
+                param = rsp_struct["width"]
+                if param.expose:
+                    # If it's a top-level exposed parameter, we need to find definition from there
+                    module = lib.get_module_by_name(topcfg, req_m)
+                    width = int(module['param_decl'].get(param.name, param.default))
+                else:
+                    width = int(rsp_struct["width"].default)
+            else:
+                width = rsp_struct["width"]
+                assert isinstance(rsp_struct["width"], int)
+
+            total_width += width
+            widths.append(width)
 
             # Type check
             # If no package was declared, it is declared with an empty string
@@ -828,7 +840,7 @@ def check_intermodule(topcfg: Dict, prefix: str) -> int:
             if param.expose:
                 # If it's a top-level exposed parameter, we need to find definition from there
                 module = lib.get_module_by_name(topcfg, req_m)
-                width = int(module['param_decl'].get(param.name, req_struct["width"].default))
+                width = int(module['param_decl'].get(param.name, param.default))
             else:
                 width = int(req_struct["width"].default)
         else:
--- a/util/topgen/merge.py
+++ b/util/topgen/merge.py
@@ -173,6 +173,14 @@ def elaborate_instance(instance, block: IpBlock):
         if isinstance(s['width'], Parameter):
             for p in instance["param_list"]:
                 if p['name'] == s['width'].name:
+                    # When mangling the name, we first need to deep copy the param. Parameters in
+                    # signals have a reference to a parameter. If we have multiple instances of the
+                    # same IP, then their signals would reference the same single parameter. If we
+                    # would mangle that directly, we all signals of all IPs would reference to that
+                    # single mangled paramter. Since parameters are instance dependent, that would
+                    # fail. Therefore, copy the parameter first to have a unique paramter for that
+                    # particular signal and instance, which is safe to mangle.
+                    s['width'] = deepcopy(s['width'])
                     s['width'].name_top = p['name_top']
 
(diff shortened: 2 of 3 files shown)


util/topgen/merge.py
```diff
@@ -173,6 +173,11 @@ def elaborate_instance(instance, block: IpBlock):
         if isinstance(s['width'], Parameter):
             for p in instance["param_list"]:
                 if p['name'] == s['width'].name:
+                    # When mangeling the name, we first need to deep copy the param. Otherwise, we
```
rswarbrick, line 176: Nit: The spelling is "mangling" :-)

util/topgen/merge.py
```diff
@@ -173,6 +173,11 @@ def elaborate_instance(instance, block: IpBlock):
...
+                    # reference always the same single parameter. If there are multiple instances
+                    # having the same parameter, we would overwrite the previously mangled param
```
rswarbrick, line 178: I think this happens if there are multiple signals whose width is given by the same parameter? Could the text be expanded slightly? (In particular, I didn't immediately know what "having the same parameter" meant)

util/topgen/merge.py
```diff
@@ -173,6 +173,11 @@ def elaborate_instance(instance, block: IpBlock):
...
+                    # of another instance
+                    s['width'] = deepcopy(s['width'])
                     s['width'].name_top = p['name_top']
```
rswarbrick, line 181: Hmm, I'm a bit confused about this. I don't think that `Parameter` has a `name_top` member. Not necessarily for this PR, but would you mind tidying this up?

util/topgen/intermodule.py
```diff
@@ -778,8 +778,19 @@ def check_intermodule(topcfg: Dict, prefix: str) -> int:
...
-            total_width += rsp_struct["width"]
-            widths.append(rsp_struct["width"])
+            if isinstance(rsp_struct["width"], Parameter):
+                param = rsp_struct["width"]
+                if param.expose:
+                    # If it's a top-level exposed parameter, we need to find definition from there
+                    module = lib.get_module_by_name(topcfg, req_m)
+                    width = int(module['param_decl'].get(param.name, rsp_struct["width"].default))
+                else:
+                    width = int(req_struct["width"].default)
```
rswarbrick, line 788: Do we know that `req_struct["width"]` has type `Parameter`? If so, maybe add an assertion here?

util/topgen/intermodule.py
```diff
@@ -778,8 +778,19 @@ def check_intermodule(topcfg: Dict, prefix: str) -> int:
...
             error += err
 
-            total_width += rsp_struct["width"]
-            widths.append(rsp_struct["width"])
+            if isinstance(rsp_struct["width"], Parameter):
+                param = rsp_struct["width"]
+                if param.expose:
+                    # If it's a top-level exposed parameter, we need to find definition from there
+                    module = lib.get_module_by_name(topcfg, req_m)
+                    width = int(module['param_decl'].get(param.name, rsp_struct["width"].default))
```
rswarbrick, line 786: I think that we could use `param.default` here?

util/topgen/intermodule.py
```diff
@@ -778,8 +778,19 @@ def check_intermodule(topcfg: Dict, prefix: str) -> int:
...
+                else:
+                    width = int(req_struct["width"].default)
+            else:
+                width = rsp_struct["width"]
```
rswarbrick, line 790: I'd probably suggest adding an assertion here that `rsp_struct["width"]` is an integer.

util/topgen/intermodule.py
Razer6: There is a check `if isinstance(req_struct["width"], Parameter)` above or am I missing something?

util/topgen/merge.py
Razer6: Add that parameter to the `Parameter` class.

Razer6:
@rswarbrick Thanks for the review, I added the changes. Could you take another look?

util/topgen/merge.py
rswarbrick: Oh, I see. Yep, that makes sense (and thanks for explaining it!)

util/topgen/intermodule.py
rswarbrick: Do you mean line 781? That's looking at `rsp_struct`.

util/topgen/intermodule.py
Razer6: ha, that is more of an issue. Fixed that.

util/topgen/intermodule.py
Razer6: that is more of an issue. Thanks, that has been fixed.

util/topgen/intermodule.py
rswarbrick: Sorry but I don't think I can see that? I'm imagining something on the previous line that looks like `assert isinstance(rsp_struct["width"], int)`

Razer6: Hmm, that seems to be lost in the universe. I added the assertion.

rswarbrick (approved):
Sorry for the slow reply. This looks good to me.

PR #25667, opened 2024-12-16
Miscellaneous changes in preparation for multitop

pamaury:
These commits are preparatory commits for #25580 but can be reviewed independently. Most should are trivial except for the following changes:
- `isr_testutils`:there is a long explanation in the commit, basically currently we are in a weird state where isr_testutils contains code for IPs in Darjeeling but it's actually ealrgrey specific. This commit makes it completely earlgrey specific become it's simply not ready for multitop yet and it's better to be in a good state before moving to multitop.
- `modid_check`: the rule was still pointing directly to opentitantool instead of going through the "golden" toolchain. This PR changes that, which should also have the side benefit to avoiding an unecessary rebuild of opentitantool in a  second configuration and save CI/build time.


util/topgen.py
```diff
@@ -1260,6 +1260,8 @@ def render_template(template_path: str, rendered_path: Path,
         ]
         for idx, path in enumerate(out_paths):
             # C Header + C File + Clang-format file
+            gencmd_c = warnhdr + GENCMD.format(top_name=top_name)
+            gencmd_bzl = gencmd_c.replace("//", "#")
```
Razer6, line 1264: Is hat being used somewhere, or is this in preparation?

Razer6 (approved):

cfrantz (approved):

util/topgen.py
pamaury: It's used by the other PR but you are right that it should not be in that commit

pamaury: Actually not it's used further down in the file, I was confused

Razer6: Thanks!

matutem (approved):
LGTM

PR #25675, opened 2024-12-17
[bazel] Remove now-unused vendored Rust crate build files

jwnrt:
These files were used pre-bzlmod but are no longer needed since #25578.

This will fail the verible lint check and the authorized changes check due to the number of files changed.


jwnrt:
Waiving the two CI failures which are caused by the number of files changed. This PR does not touch RTL or any blocked files.

PR #25677, opened 2024-12-17
[bazel] Correct `bazel_skylib` version in `MODULE.bazel`

jwnrt:
Due to semantic version resolution, we're actually getting `bazel_skylib` version 1.7.1 due to a dependency asking for it.

Update the version listed in `MODULE.bazel` to match so the warning is supressed.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,7 +4,7 @@
 
 module(name = "lowrisc_opentitan")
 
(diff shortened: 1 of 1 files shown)


nbdd0121 (approved):

nbdd0121:
CI failure due to GitHub downtime, ignoring.

PR #25679, opened 2024-12-17
[bazel] Move buildifier to bzlmod

jwnrt:
Switch from building buildifier ourselves to using the `buildifier_prebuilt` module. This also means we don't have to build it from source using a Go toolchain.


nbdd0121 (approved):
Implementation LGTM.

It's good that we can get rid of the go dependency, although I'll leave it to others to decide if it's acceptable to add new binary dependency.

jwnrt:
Who would be best to get a review from?

For context, buildifier runs on the host for dev purposes (formatting Bazel code).

pamaury (approved):
Looks good. Switching to a prebuilt time will also save time because we routinely have to recompile it.

jwnrt:
> Would this be needed for Airgapped environment, or would the use of Bzlmod make this a lazy dependency and won't be needed if buildifier_fix is not to be executed?

It’s loaded lazily, so I don’t think it’s needed

jwnrt:
I think i’ll merge since the risk of a prebuilt Go binary on the host seems low (they don’t even link to libc do they?)

PR #25680, opened 2024-12-17
[bazel] Migrate Google dependencies to bzlmod

jwnrt:
Simple migrations so they can go directly in the main `MODULE.bazel`.


MODULE.bazel
```diff
@@ -4,16 +4,26 @@
 
 module(name = "lowrisc_opentitan")
 
+bazel_dep(name = "abseil-cpp", version = "20230802.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "googletest", version = "1.11.0")
+bazel_dep(name = "lowrisc_misc_linters")
```
nbdd0121, line 10: Move this line closer to the override?

jwnrt: Done

jwnrt:
Sorry for the notifications, this should have been a draft. It's blocked on the buildifier migration.

MODULE.bazel
jwnrt: Would you mind if I keep the dependencies sorted? If I try move this next to the override, `buildifier` tries to move it back into the list but now unsorted.

nbdd0121: Would buildifier try to move it back to the list even across comments?

jwnrt: I'd forgotten to use the `dev_dependency` parameter for this module so I've moved it to a separate section anyway. Now it's next to its override. Let me know if this is clearer nw.

nbdd0121 (approved):

Issue #24691, opened 2024-09-30
[hmac,dv] intr_state TB issue


martin-velay:
### Description

`UVM_ERROR (hmac_scoreboard.sv:481): intr_state`

Command to reproduce with [bc90ee9322](https://github.com/lowrisc/opentitan/tree/bc90ee9322):
`util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_stress_all_with_rand_reset -t vcs --fixed-seed 50137718212529458093890268432036893881721916430145661069749074293950074877432`

martin-velay:
Closed as https://github.com/lowRISC/opentitan/pull/25843 is merged

PR #25683, opened 2024-12-17
[hmac,dv] Fix fifo_empty interrupt

martin-velay:
- linked to issue #24691
- where the interrupt state for fifo_empty field from the RAL against the RTL were not matching. This was the case when all the empty conditions were filled but at the same time a full is coming. It was just a matter of changing the condition order in the TB.
- fix also FIFO depth status by allowing +1 of tolerance as this is not super critical to be cycle accurate.

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_scoreboard.sv
+++ b/hw/ip/hmac/dv/env/hmac_scoreboard.sv
@@ -113,7 +113,6 @@ task hmac_scoreboard::process_tl_access(tl_seq_item item, tl_channels_e channel,
   uvm_reg csr;
   string  csr_name;
   bit     do_read_check           = 1'b1;
-  bit     do_cycle_accurate_check = 1'b1;
   bit     write                   = item.is_write();
   bit [TL_AW-1:0] addr_mask       = ral.get_addr_mask();
   uvm_reg_addr_t  csr_addr        = cfg.ral_models[ral_name].get_word_aligned_addr(item.a_addr);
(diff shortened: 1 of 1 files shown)


vogelpi (approved):
This looks good to me, thanks @martin-velay !

martin-velay:
@rswarbrick, any feedback?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -607,60 +618,74 @@ endtask : hmac_process_fifo_wr
...
-    end
+    // Spawn a thread as an event from the list above might occurs during the time we wait
+    // for the clock cycle (later in this thread).
+    fork begin
+      // Store hmac_process and hmac_start to be able to detect when it goes up, as it could remain
+      // up for a while
+      bit hmac_process_posedge  = hmac_process & ~hmac_process_last;
+      bit hmac_start_posedge    = hmac_start & ~hmac_start_last;
+      bit hmac_stopped_posedge  = hmac_stopped & ~hmac_stopped_last;
+      bit hmac_continue_posedge = hmac_continue & ~hmac_continue_last;
```
rswarbrick, line 629: I think we're just using these four variables together (to decode whether to clear `fifo_full_detected`). Can't we just snapshot the 1-bit "should clear fifo full" value?

martin-velay: Sorry my brain is maybe a bit slow today, can you tell me how you suggest to do that ?

rswarbrick: I'm really sorry: I've just re-read the code and noticed that each `&` is between two different variables. So I think your code looks completely sensible already. D'oh!

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -478,8 +478,19 @@ task hmac_scoreboard::process_tl_access(tl_seq_item item, tl_channels_e channel,
...
+        if (!cfg.sar_skip_ctxt) begin
+          // We don't really need to be cycle accurate for this status register, instead we
+          // tolerate +1
+          if ((`gmv(ral.status.fifo_depth)   != rd_fifo_depth) &&
+              (`gmv(ral.status.fifo_depth)+1 != rd_fifo_depth)) begin
+            `uvm_error(`gfn, $sformatf({"Check failed, read status.fifo_depth is not matching ",
+                "expected value, not even with tolerance of +1: got 0x%0d but exp 0x%0h"},
+                rd_fifo_depth, `gmv(ral.status.fifo_depth)))
+          end else begin
+            do_read_check = 1;    // Still perform check as this register holds other fields
```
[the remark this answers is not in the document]
martin-velay: I have implemented a change to ensure that the check for the other fields will occur, but I have decided to take another path instead. I have talked with Antonio to understand the new feature he has implemented with the field mask, but I feel easier and less risky to make the way I've done:
I am simply updating the FIFO depth field value to match the register value before the check occurs.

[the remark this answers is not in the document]
rswarbrick: Ok, that seems reasonable. Thanks for the reply.

rswarbrick (approved):

PR #25684, opened 2024-12-17
[dif, sram_ctrl] Update wipe and key request description

nasahlpa:
With lowRISC/opentitan#21850, key requests that are issued during a memory init are ignored by the hardware. This has not yet been reflected in the dif_sram_ctrl_request_new_key and dif_sram_ctrl_wipe function description.


rswarbrick (commented):
This looks sensible, but I'll wait for a SW person to give it a once-over too.

engdoreis (approved):

PR #25310, opened 2024-11-22
[imm_rom_ext] Add immutable ROM_EXT ePMP reconfiguration

sasdf:
The immutable ROM_EXT data segment is also covered in ROM_EXT code region due to ROM limitation,
This PR let immutable ROM_EXT reconfigure ePMP first to remove executable permission on the data segment.

### Jumping from IMM_ROM_EXT -> ROM / mutable ROM_EXT

We propose to use the following new ePMP layout before jumping to the mutable ROM_EXT:
```
 *    0: ROM      ----- ----
 *    1: ROM        TOR LX-R
 *    2: ROM      NAPOT L--R
 *    3: -------  ----- ----
 *    4: -------  ----- ----
 *    5: -------  ----- ----
 *    6: IM_EXT   ----- ----
 *    7: IM_EXT     TOR LX-R
 *    8: MU_EXT   ----- ----
 *    9: MU_EXT     TOR LX-R
 *   10: VIRTUAL  NAPOT L--R
 *   11: STACK      NA4 L---
 *   12: RvDM     NAPOT LXWR
 *   13: FLASH    NAPOT L--R
 *   14: MMIO     NAPOT L-WR
 *   15: RAM      NAPOT L-WR
```

MU_EXT stands for mutable part of rom_ext text segment, and IM_EXT is the immutable part.

Slot 10 will be empty if address translation is not enabled, and stack guard
is added back to slot 11. 

Slot 12 and 13 are also swapped, so that the chip in production LCS has one more contiguous free entry. 

---

### Jumping from ROM_EXT -> Owner SW

Before jumping to Owner SW, all lock bits will be cleared, so all entries can
be reclaimed by Owner SW. Entries 8 ~ 11 ( &12 in prod LCS ) can be safely reconfigure by Owner SW.

```
 *    0: -------  ----- ----
 *    1: -------  ----- ----
 *    2: OWNER    ----- ----
 *    3: OWNER      TOR -X-R
 *    4: OWNER_V  NAPOT ---R
 *    5: -------  ----- ----
 *    6: -------  ----- ----
 *    7: -------  ----- ----
 *    8: MU_EXT   ----- ----
 *    9: MU_EXT     TOR -X-R
 *   10: VIRTUAL  NAPOT ---R
 *   11: STACK      NA4 ----
 *   12: RvDM     NAPOT -XWR
 *   13: FLASH    NAPOT ---R
 *   14: MMIO     NAPOT --WR
 *   15: RAM      NAPOT --WR
```

[...]

PR #25686, opened 2024-12-17
Link real IMM_ROM_EXT into ROM_EXT

lchiawei:
This PR mainly does the work about the IMM_ROM_EXT linking, including following commits:
* Refine the linker script and startup assembly for our real IMM_ROM_EXT.
* Make ROM_EXT use prebuilt IMM_ROM_EXT binary during the linking phase, instead of calling `imm_rom_ext_main()` in `rom_ext_main()`.
* Modify some bazel rules to make the ROM_EXT choose correct IMM_ROM_EXT binary according to the `exec_env`. For silicon creator, also add a new exec_env that does not contain a manifest to skip the signing phase.
* Support IMM_ROM_EXT for slot A & B besides original virtual slot usage.


sw/device/silicon_creator/imm_rom_ext/imm_rom_ext_epmp.c
```diff
@@ -0,0 +1,115 @@
...
+  // Reconfigure the ePMP MMIO region to be NAPOT region 14, thus freeing
+  // up an ePMP entry for use elsewhere.
+  epmp_set_napot(14, kMmioRegion, kEpmpPermLockedReadWrite);
+
+  // ePMP region 11 protects the stack from overflow.
+  // This stack guard was in ePMP region 14.
+  epmp_set_napot(11, kStackGuard, kEpmpPermLockedNoAccess);
+
+  // ePMP region 12 allows RvDM access.
+  // This RvDM region was in ePMP region 13.
```
timothytrippel, line 61: any reason this need to move down a slot? keeping in slot 13 aligns with what the ROM and ROM_EXT both expect on earlgrey_1.0.0 branch https://cs.opensource.google/opentitan/opentitan/+/earlgrey_1.0.0:sw/device/silicon_creator/rom_ext/rom_ext.c;l=945 I think?

sw/device/silicon_creator/imm_rom_ext/imm_rom_ext_epmp.c
```diff
@@ -0,0 +1,115 @@
...
+
+  // Clear mutable ROM_EXT entries (8 & 9).
+  epmp_clear(9);
+  epmp_clear(8);
+
+  // Immutable ROM_EXT TOR (6 & 7).
+  epmp_set_tor(6, kImmTextRegion, kEpmpPermLockedReadExecute);
+
+  // Clear entries from 5 ~ 3.
+  epmp_clear(5);
```
timothytrippel, line 90: wasn't slot 5 configured for flash read access by ROM? is there a reason for removing flash read access from the immutable rom_ext? is it just not needed?

hw/top_earlgrey/BUILD
```diff
@@ -426,6 +426,38 @@ silicon(
...
 
+silicon(
+    name = "silicon_creator_without_manifest",
+    args = [
+        "--rcfile=",
+        "--logging=info",
+        "--interface={interface}",
+    ],
+    design = "earlgrey",
+    # TODO: Add switch to enable privdate key.
```
timothytrippel, line 437: what do you mean here? this can be overridden already when used like here https://cs.opensource.google/opentitan/opentitan/+/earlgrey_1.0.0:sw/device/silicon_creator/rom_ext/BUILD;l=236

timothytrippel:
@lchiawei could you also rebase this on the latest tip of tree since it has been a few weeks since this was submitted (to prevent merge SKU)?

sw/device/silicon_creator/imm_rom_ext/imm_rom_ext_epmp.c
sasdf: We swap the Slot 12 and 13 (RvDM and Flash) so that the owner SW in production LCS has one more contiguous free entry. You could find the full layout here https://github.com/lowRISC/opentitan/pull/25310.

sw/device/silicon_creator/imm_rom_ext/imm_rom_ext_epmp.c
sasdf: Flash read access is moved to slot 13 by line 71.

timothytrippel:
I think this can be rebased now to fix the merge conflicts.

hw/top_earlgrey/BUILD
lchiawei: Hi @timothytrippel , sorry for the late reply.
Since I reuse `opentitan_binary` to build immutable ROM_EXT, we need to skip the signing phase when building. Thus, I introduce a new exec env as same as silicon_creator but without a manifest. It seems to me that giving an empty key with current existing exec env will make it fall back to the default key. It was discussed before in my previous PR: https://github.com/lowRISC/opentitan/pull/25274#issuecomment-2489735428, but I haven't address this issue yet. Do you have any suggestions?

lchiawei:
> I think this can be rebased now to fix the merge conflicts.

Thanks for reviewing my PR, I have rebased my commit chain now. However, the CI seems to be cancelled and we may need to trigger the tests again.

hw/top_earlgrey/BUILD
cfrantz: Look into `CLEAR_KEY_SET`.  I believe you can use that to override the keys in the binary rules to mean "no key".

lchiawei: @cfrantz Is it possible to clear the manifest as well?

lchiawei: Hi @timothytrippel , To clarify, this exec_env is used to clear / reset the manifest and make `opentitan_binary` skip the signing part. Using existing exec_env and leaving the manifest as None will still fall back to use the default manifest. I discussed with @sasdf yesterday and he suggested me to add a reserved target path which means an "empty" manifest explicitly, and we can skip the signing phase in the `opentitan_binary` if we use that empty manifest. Do you have any other suggestions?

timothytrippel: @lchiawei by "empty manifest" do you mean a manifest whose fields are all zeros? or do you mean skip attaching a manifest to the binary altogether?

hw/top_earlgrey/BUILD
```diff
@@ -426,6 +426,38 @@ silicon(
...
+        "//sw/device/lib/arch:boot_stage_rom_ext",
+        "//sw/device/lib/arch:silicon",
+    ],
+    linker_script = "//sw/device/lib/testing/test_framework:ottf_ld_silicon_creator_slot_a",
+    param = {
+        "interface": "teacup",
+        "exit_success": DEFAULT_TEST_SUCCESS_MSG,
+        "exit_failure": DEFAULT_TEST_FAILURE_MSG,
+    },
+    rom_scramble_config = "//hw/top_earlgrey/data:autogen/top_earlgrey.gen.hjson",
```
timothytrippel, line 452: I think you mean? (This is why this is failing CI)

```suggestion
    rom_scramble_config = "//hw/top_earlgrey/data/autogen:top_earlgrey.gen.hjson",
```

hw/top_earlgrey/BUILD
```diff
@@ -426,6 +426,38 @@ silicon(
...
+    # TODO: Add switch to enable privdate key.
+    ecdsa_key = {"//sw/device/silicon_creator/rom/keys/fake/ecdsa:test_key_0_ecdsa_p256": "test_key_0"},
+    exec_env = "silicon_creator",
+    extract_sw_logs = "//util/device_sw_utils:extract_sw_logs_db",
+    flash_scramble_tool = "//util/design:gen-flash-img",
+    libs = [
+        "//sw/device/lib/arch:boot_stage_rom_ext",
+        "//sw/device/lib/arch:silicon",
+    ],
+    linker_script = "//sw/device/lib/testing/test_framework:ottf_ld_silicon_creator_slot_a",
```
timothytrippel, line 446: I think if you want to eliminate the manifest section from the immutable ROM_EXT binary, you just need to change the linker script, which you arleady do below when you override it below with the `imm_rom_ext_*.ld` scripts you defined.

hw/top_earlgrey/BUILD
```diff
@@ -426,6 +426,38 @@ silicon(
     """,
 )
 
+silicon(
+    name = "silicon_creator_without_manifest",
```
timothytrippel, line 430: on second thought, I don' think you need this custom exec_env at all, I think if you don't want a manifest you just override the linker script of the default silicon_creator exec_env, and the key used with CLEAR_KEY_SET as Chris mentioned. Unless I am missing something?

sw/device/silicon_creator/imm_rom_ext/defs.bzl
```diff
@@ -2,8 +2,41 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+DEFAULT_EXEC_ENV = [
+    "//hw/top_earlgrey:fpga_cw310",
+    "//hw/top_earlgrey:fpga_cw340",
+    "//hw/top_earlgrey:sim_dv_base",
+    "//hw/top_earlgrey:sim_verilator_base",
+    "//hw/top_earlgrey:silicon_creator_without_manifest",
```
timothytrippel, line 10: I don't think you need this one do you? (see comments above); same for below

hw/top_earlgrey/BUILD
cfrantz: The CLEAR_KEY_SET value only clears keys, it does not clear the manifest.

As discussed in some of the other comments, if you need to omit the manifest from the binary altogether, you probably need to adjust your linker script to omit it.  Most binaries include a manifest via `//sw/device/silicon_creator/lib:manifest_def`.  This module includes a C-struct version of the manifest with certain values provided by linker constants.  In order to eliminate it completely, you need to make sure the linker discards/never includes the section.

hw/top_earlgrey/BUILD
lchiawei: Done. Thanks for pointing this out.

hw/top_earlgrey/BUILD
lchiawei: Sorry for the confusion caused. The linker script `imm_rom_ext_*.ld` to build IMM_ROM_EXT has been modified to not contain the manifest section in binary. The reason I introduced a new `exec_env` here is I would like to bypass the signing phase in the `_build_binary` rule here: https://cs.opensource.google/opentitan/opentitan/+/master:rules/opentitan/cc.bzl;l=191;drc=2df193412d42efaa00f4a7fe8708c0302306eca2

With current `silicon_creator` exec_env, even we do not provide the manifest, the manifest will fall back to the default one (as we did `manifest = get_fallback(ctx, "file.manifest", exec_env)` above) and we are not able to skip the signing phase. Hope this helps clarify the questions. 

So the question is that if there are any better approaches to skip this signing procedure without introducing the new exec_env.

hw/top_earlgrey/BUILD
lchiawei: You're right, this is not for eliminating the manifest section from the binary and actually I had added the dedicated linker scripts for IMM_ROM_EXT. It is used to skip the signing phase when building binary instead. (Please see the comments below.)

hw/top_earlgrey/BUILD
cfrantz: We should create a special label similar to the `CLEAR_KEY_SET` that indicates that we want no manifest (e.g. a `none` manifest).  We can then modify cc.bzl like this (pseudo code):
```
manifest = get_fallback(ctx, "file.manifest", exec_env)
if manifest == "<label of none manifest>":
    manifest = None
```

This will allow an opentitan_binary rule to use the `silicon_creator` environment, but override the default manifest and skip signing.

lchiawei: Hi @cfrantz , thanks for your suggestion. After some experiments, I am able to use this approach to clear the manifest. By the way, where will you suggest to put the `CLEAR_MANIFEST`? I don't think it is possible to put something like `CLEAR_MANIFEST = manifest(d = {"name": "none_manifest"})` in `rules/opentitan/defs.bzl` as bazel will complain that it cannot instantiate a rule when loading a .bzl file.

timothytrippel: I think once you make the change @cfrantz suggests above then you will no longer need this new exec_env

hw/top_earlgrey/BUILD
lchiawei: I have a dedicated linker script `imm_rom_ext_common.ld` and it does not include the manifest section. Let me resolve this conversation now.

hw/top_earlgrey/BUILD
lchiawei: Discussed with @timothytrippel in today's meeting. I added a `CLEAR_MANIFEST` in `hw/top_earlgrey/BUILD` and also removed this exec_env now.

sw/device/silicon_creator/imm_rom_ext/defs.bzl
lchiawei: Fixed. I used the original exec_env silicon_creator instead.

timothytrippel (approved):

PR #25693, opened 2024-12-18
Update setup_verilator.md

baskaranindia:
touching workspace is required, otherwise the bazel complains for missing workspace file.
Refer https://stackoverflow.com/questions/41791171/bazel-build-for-tensorflow-inception-model


nbdd0121:
We used bzlmod and does not use workspace file. You need to use an newer Bazel that matches our .bazel-version file.

PR #25300, opened 2024-11-21
[otbn/rsa] Add SCA hardening for sel_sqr_or_sqrmul

aewag:
As I cannot add reviewers I link you @jadephilipoom @johannheyszl @vogelpi 

This commit hardens two sources of leakage

- DMEM writeback: The selected temporary result, either sqr or sqrmul depending on the current exponent bit, is written to the DMEM. At this DMEM address the sqr result is stored. In the case of selecting the sqr result the HD of the selected result and the DMEM value is zero and in the other case a high value. This allows to distinguish exponent bits. The hardening overwrites the value at the DMEM address with a random value.
- SEL instruction: BN.SEL selects one of the two source WDRs based on the carry flag. The selection depends on the current exponent bit. This allows to distinguish exponent bits. The hardening randomizes the WDRs which contain the sqr or sqrmul results. The randomization is performed for each limb.

EDIT:
Grep into OTBN ISA SIM log to show functionality

```
000000ec | csrrs          x23, urnd, x0                  | [x23 = 0xb078d8e1, otbn.INSN_CNT = 0x000554cb]
000000f0 | andi           x23, x23, 1                    | [x23 = 0x00000001, otbn.INSN_CNT = 0x000554cc]
000000f4 | xor            x22, x22, x23                  | [x22 = 0x00000003, otbn.INSN_CNT = 0x000554cd]
000000f8 | xor            x9, x9, x23                    | [x09 = 0x00000002, otbn.INSN_CNT = 0x000554ce]
000000fc | xor            x11, x11, x23                  | [x11 = 0x00000003, otbn.INSN_CNT = 0x000554cf]
--
000000ec | csrrs          x23, urnd, x0                  | [x23 = 0xcb607848, otbn.INSN_CNT = 0x00055761]
000000f0 | andi           x23, x23, 1                    | [x23 = 0x00000000, otbn.INSN_CNT = 0x00055762]
000000f4 | xor            x22, x22, x23                  | [x22 = 0x00000003, otbn.INSN_CNT = 0x00055763]
000000f8 | xor            x9, x9, x23                    | [x09 = 0x00000003, otbn.INSN_CNT = 0x00055764]
000000fc | xor            x11, x11, x23                  | [x11 = 0x00000002, otbn.INSN_CNT = 0x00055765]
--
000000ec | csrrs          x23, urnd, x0                  | [x23 = 0x69d25f37, otbn.INSN_CNT = 0x0005576b]
[...]

PR #25694, opened 2024-12-18
[crypto] Harden RSA modexp subroutine.

jadephilipoom:
See https://github.com/lowRISC/opentitan/pull/25300 for context; this is an updated version of that hardening change based on the [coding guidelines](https://opentitan.org/book/doc/contributing/style_guides/otbn_style_guide.html?highlight=bn.sel#handling-secret-shares) from COCO-Alma.

Add some SCA hardening measures to the modexp subroutine that selects either the squared or squared-and-multiplied result in square-and-multiply-always modular exponentiation.

Although the rest of the code is not hardened, the hardening for RSA will likely preserve this subroutine, so hardening it is still useful.

### DIFF
--- a/sw/otbn/crypto/modexp.s
+++ b/sw/otbn/crypto/modexp.s
@@ -28,22 +28,28 @@
  * @param[in]  x11: pointer to temp reg, must be set to 2
  *
  * clobbered registers: x8, x21, w0, w2
- * clobbered Flag Groups: none
+ * clobbered Flag Groups: FG1
  */
 sel_sqr_or_sqrmul:
   /* iterate over all limbs */
-  loop      x30, 4
+  loop      x30, 8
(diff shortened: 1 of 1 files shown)


jadephilipoom:
Performance impact is minimal: RSA-2048 modexp goes from  `18889021` cycles (according to the last [benchmarks](https://opentitan.org/book/hw/ip/otbn/doc/otbn_intro.html#performance) to `18970942` cycles, a 0.4% increase.

johannheyszl (approved):
Thanks this is a great example for effective collab :)

nasahlpa (approved):
Thanks, LGTM!

This is the code snippet that was used for the SCA analysis @johannheyszl?

johannheyszl:
@nasahlpa this is correct

PR #25697, opened 2024-12-18
Ot py for repo

pamaury:


pamaury:
I should note that although I think this is a nice "hack", I still have mixed feelings about it because it depends on semi-documented repository names used by rules_python. Looking at the code of rules_python, I am not very worried because they have to run python script and run into the same problem: this is why the python_host repository exists in the first place and why there is one repo per package. However, the fact that those repo are visible and the names used does not seem to be documented as stable.

Another remark: with the extension, it's possible to to do several registrations steps so that each "python interpret" has its own set of packages enabled. What this means is that if we want to, we don't have to created one interpreter which has access to the union of all dependencies of all users. Still, the fact that we need to manually list all transitive dependencies makes this process cumbersome anyway.

PR #25700, opened 2024-12-18
[hsmtool] SPX cleanups

cfrantz:
1. It is legal for SPX key entities to be missing a PKCS#11 `id`.
2. Rename the `acorn` field to `spx`, as it represents a generic interface to SPX functions rather than a specific implementation.


moidx (approved):
LGTM.

jadephilipoom (approved):

PR #25701, opened 2024-12-18
[rv_plic] Adjust naming of PRIO regs by using multireg

a-will:
Now that we have non-compact multireg features, use them to eliminate the explicit numbering of CSRs from the template and improve Rust code generation.


hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic_reg_pkg.sv
```diff
@@ -797,192 +57,7 @@ package rv_plic_reg_pkg;
...
-    rv_plic_reg2hw_prio177_reg_t prio177; // [218:217]
-    rv_plic_reg2hw_prio178_reg_t prio178; // [216:215]
-    rv_plic_reg2hw_prio179_reg_t prio179; // [214:213]
-    rv_plic_reg2hw_prio180_reg_t prio180; // [212:211]
-    rv_plic_reg2hw_prio181_reg_t prio181; // [210:209]
-    rv_plic_reg2hw_prio182_reg_t prio182; // [208:207]
-    rv_plic_reg2hw_prio183_reg_t prio183; // [206:205]
-    rv_plic_reg2hw_prio184_reg_t prio184; // [204:203]
-    rv_plic_reg2hw_prio185_reg_t prio185; // [202:201]
+    rv_plic_reg2hw_prio_mreg_t [185:0] prio; // [572:201]
```
jadephilipoom, line 60: Same question here about the length and bit-length.

hw/top_darjeeling/ip_autogen/rv_plic/rtl/rv_plic_reg_pkg.sv
```diff
@@ -693,166 +57,7 @@ package rv_plic_reg_pkg;
...
-    rv_plic_reg2hw_prio151_reg_t prio151; // [192:191]
-    rv_plic_reg2hw_prio152_reg_t prio152; // [190:189]
-    rv_plic_reg2hw_prio153_reg_t prio153; // [188:187]
-    rv_plic_reg2hw_prio154_reg_t prio154; // [186:185]
-    rv_plic_reg2hw_prio155_reg_t prio155; // [184:183]
-    rv_plic_reg2hw_prio156_reg_t prio156; // [182:181]
-    rv_plic_reg2hw_prio157_reg_t prio157; // [180:179]
-    rv_plic_reg2hw_prio158_reg_t prio158; // [178:177]
-    rv_plic_reg2hw_prio159_reg_t prio159; // [176:175]
+    rv_plic_reg2hw_prio_mreg_t [159:0] prio; // [494:175]
```
jadephilipoom, line 60: At first glance, it looks to me like the `[159:0]` and the `// [494:201]` here disagree, since the comment indicates a 320-bit range. Is this maybe an array of two-bit multireg entries that has length 160 but _bit-length_ 320? But the multireg entry below has `[159:0]` and `// [174:15]`, which match exactly with length and bit-length 160. I'm not an RTL expert so I'm probably missing something here, but wanted to double-check :slightly_smiling_face:

jadephilipoom (approved):
Whew! Can't promise I read every line of the diff closely, but I at least opened all the files, read the first occurrence of each code-change pattern, and checked that the following lines looked like 160/185 repetitions of the same pattern. Thanks, Alex!

hw/top_darjeeling/ip_autogen/rv_plic/rtl/rv_plic_reg_pkg.sv
a-will: Yes, that's exactly right. This is a definition of an array of `rv_plic_reg2hw_prio_mreg_t`, which is defined above as a struct with a single 2-bit member (with the member named `q`).

a-will:
From a software perspective, the interesting bit here is how the output of `//hw/top_earlgrey:rv_plic_rust_regs` changes (and same for darjeeling).

With the multireg construction, reggen will generate an array for the Rust struct instead of the flattened variant. Without this, each priority register gets a named member in the IP's register definitions, which is just a little less nice to work with.

I noticed the incongruity with the other CSRs as I was fiddling with topgen for other reasons and figured, "Why not make this little tweak?" ;)

Razer6 (approved):
This looks good to me. Less technical debt. Always good!

a-will:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/data/rv_plic.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic_reg_pkg.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic_reg_top.sv

This PR only renames the PRIO CSRs. There is no functional change.

hw/ip_templates/rv_plic/doc/programmers_guide.md
```diff
@@ -6,7 +6,7 @@ After reset, RV_PLIC doesn't generate any interrupts to any targets even if
 interrupt sources are set, as all priorities and thresholds are 0 by default and
 all ``IE`` values are 0. Software should configure the above three registers.
 
-[`PRIO0`](../data/rv_plic.hjson#prio0) .. [`PRIO31`](../data/rv_plic.hjson#prio1) registers are unique. So, only one of the targets
+[`PRIO0`](../data/rv_plic.hjson#prio_0) .. [`PRIO31`](../data/rv_plic.hjson#prio_1) registers are unique. So, only one of the targets
```
andreaskurth, line 9: Probably a typo that existed before, but shouldn't it be
```suggestion
[`PRIO0`](../data/rv_plic.hjson#prio_0) .. [`PRIO31`](../data/rv_plic.hjson#prio_31) registers are unique. So, only one of the targets
```
?

andreaskurth (approved):
Multiregs FTW :-) Nice, thanks @a-will!

I just have one small suggestion/question

hw/ip_templates/rv_plic/doc/programmers_guide.md
a-will: This type of link appears to be broken in general, so I think I'll skip fixing this in this particular PR. alert_handler has a similar setup trying to reference data from the hjson, but it doesn't get transformed into a link that goes to the correct HTML page.

In addition, the programming guide should probably be templated to use the correct range of CSRs.

andreaskurth:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/data/rv_plic.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic_reg_pkg.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/rv_plic/rtl/rv_plic_reg_top.sv

This PR only renames the PRIO CSRs. There is no functional change.

PR #25703, opened 2024-12-18
[prim,ipgen] Fix up some hierarchies of fusesoc cores

a-will:
Ensure packages with constants are at the leaves of the dependency tree, so other cores may depend on them without creating cycles. Adjust the pwrmgr_reg core to depend on the pwrmgr_pkg core, with pwrmgr_reg_pkg moved to the pwrmgr_pkg core.

Also split the mubi RTL implementations from the package with declarations and standalone utility functions.

In addition, fix up the reg tops for various ipgen'd cores to include a dependency on `prim:subreg`.

### DIFF
--- a/hw/ip_templates/rstmgr/rstmgr_reg.core.tpl
+++ b/hw/ip_templates/rstmgr/rstmgr_reg.core.tpl
@@ -10,6 +10,7 @@ virtual:
 filesets:
   files_rtl:
     depend:
+      - lowrisc:prim:subreg
       - lowrisc:tlul:headers
     files:
       - rtl/rstmgr_reg_pkg.sv
--- a/util/design/data/prim_mubi_pkg.core.tpl
+++ b/util/design/data/prim_mubi_pkg.core.tpl
@@ -8,21 +8,14 @@ CAPI=2:
 #
 #    util/design/gen-mubi.py
 #
-name: "lowrisc:prim:mubi:0.1"
+name: "lowrisc:prim:mubi_pkg:0.1"
 description: "Multibit types and functions"
 filesets:
   files_rtl:
     depend:
       - lowrisc:prim:assert
-      - lowrisc:prim:buf
-      - lowrisc:prim:flop
-      - lowrisc:prim:mubi_pkg
     files:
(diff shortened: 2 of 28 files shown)


util/design/data/prim_mubi_pkg.core.tpl
```diff
@@ -0,0 +1,45 @@
+CAPI=2:
```
rswarbrick, line 1: This is really a question from my interest rather than a review comment, but can splitting out `prim_mubi_pkg` come as an initial commit? (Or does it depend on the change in `pwrmgr` as well?)

rswarbrick (commented):
This looks really sensible to me! Very happy with it once you're ready with the change.

hw/ip_templates/rstmgr/rstmgr_reg.core.tpl
```diff
@@ -10,6 +10,7 @@ virtual:
 filesets:
   files_rtl:
     depend:
+      - lowrisc:prim:subreg
       - lowrisc:tlul:headers
```
matutem, line 14: Funny this depends only on tlul:headers and clkmgr on ip:tlul. It seems it should only one of the two uniformly.

a-will: Since the reg_top is here, it should be ip:tlul. Nice catch!

Relatedly, I think I ought to move rstmgr_reg_pkg.sv over to the rstmgr_pkg core, so it can avoid pulling in RTL cores that can create dependency cycles. (similar to what has been done for the pwrmgr_pkg core)

util/design/data/prim_mubi_pkg.core.tpl
a-will: Done!

a-will:
Well, I'd love for CI to not get broken from this, but I can't debug it, due to the closed-source repo that's required just to call the tool. The terms for ascentlint are antithetical to open-source hardware development, and I feel like we ought to just chuck it from the CI flow.

a-will:
Superseded.

PR #25705, opened 2024-12-19
[dd, aon_timer cov_closure] CCOV exclusion file for aon_timer

antmarzam:
All the exclusions in this file are unreachable


hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
```diff
@@ -0,0 +1,182 @@
...
+//==================================================
+CHECKSUM: "1706182284 132761700"
+INSTANCE: tb.dut.u_reg.u_reg_if.u_rsp_intg_gen
+ANNOTATION: "tl_i.d_user is hardcoded to 0x0 "
+Block 1 "461445014" "assign rsp_intg = tl_i.d_user.rsp_intg;"
+ANNOTATION: "tl_i.d_user is hardcoded to 0x0"
+Block 2 "2643129081" "assign data_intg = tl_i.d_user.data_intg;"
+CHECKSUM: "2815520955 1214062232"
+INSTANCE: tb.dut.u_reg.u_wkup_count_hi_cdc.u_arb
+ANNOTATION: "Default statement can't be hit unless some wrong value is forced in the `state_d` signal"
```
rswarbrick, line 16: Hmm. It's a 2-bit signal whose higher bit is constant zero! The code comes from 23d3bb5cda0 and I wonder whether Tim was thinking about safety against fault injection (in which case we probably would have wanted values 01 and 10).

@vogelpi: As a security expert who might have a bit more background on this, what do you think is supposed to happen here? Are we worried about fault injection? (If so, I claim we should change the enum encoding). If not, maybe this should be a single bit?

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
```diff
@@ -0,0 +1,186 @@
...
+Branch 4 "3547459906" "gen_wr_req.state_q" (6) "gen_wr_req.state_q default,-,-,-,-"
+CHECKSUM: "530940107 4102526809"
+INSTANCE: tb.dut
+ANNOTATION: "intr_test_flds_we[0/1] is asserted at the same time whenever the intr_test register is written, so the combination of only one Qe being set is not possible."
+Condition 4 "594611319" "(reg2hw.intr_test.wkup_timer_expired.qe | reg2hw.intr_test.wdog_timer_bark.qe) 1 -1" (3 "10")
+Condition 4 "594611319" "(reg2hw.intr_test.wkup_timer_expired.qe | reg2hw.intr_test.wdog_timer_bark.qe) 1 -1" (2 "01")
+CHECKSUM: "1125143680 279071062"
+INSTANCE: tb.dut.u_reg.u_wkup_count_hi_cdc
+ANNOTATION: "src_ack is only 1 cycle pulse, and is an acknowledgment to a request from a read/write.
```
rswarbrick, line 57: This is another example where I think an RTL change is the right move. As your note says, there's even an assertion that checks the behaviour won't change :-)

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
```diff
@@ -0,0 +1,186 @@
...
+Condition 3 "86646938" "(src_busy_q && src_ack) 1 -1" (1 "01")
+CHECKSUM: "1125143680 1408670975"
+INSTANCE: tb.dut.u_reg.u_wdog_ctrl_cdc
+ANNOTATION: "src_ack is only 1 cycle pulse, and is an acknowledgment to a request from a read/write.
+src_busy_q feeds of the request, so by the time src_ack is high src_busy_q should also be high.
+See assertion SrcAckBusyChk_A for more details"
+Condition 3 "86646938" "(src_busy_q && src_ack) 1 -1" (1 "01")
+CHECKSUM: "2815520955 4109606122"
+INSTANCE: tb.dut.u_reg.u_wkup_count_hi_cdc.u_arb
+ANNOTATION: "It can't be hit because dst_update_ack won't be set prior to dst_update_req"
```
rswarbrick, line 117: I'm not sure that's quite correct. These signals come through `prim_sync_reqack`, where `dst_update_req` is visible as `src_req_i` and `dst_update_ack` is visible as `src_ack_o`.

I was slightly surprised that this module doesn't have an assumption that says `src_ack_o -> src_req_i`. But I don't think it's true! If `src_req_i` is asserted and acknowledged, what happens when `src_req_i` drops again?

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
```diff
@@ -0,0 +1,186 @@
...
+INSTANCE: tb.dut.u_reg.u_wdog_count_cdc.u_arb
+ANNOTATION: "Default statement can't be hit unless some wrong value is forced in the `state_d` signal"
+Branch 4 "3547459906" "gen_wr_req.state_q" (6) "gen_wr_req.state_q default,-,-,-,-"
+CHECKSUM: "2815520955 4051034043"
+INSTANCE: tb.dut.u_reg.u_wkup_cause_cdc.u_arb
+ANNOTATION: "Default statement can't be hit unless some wrong value is forced in the `state_d` signal"
+Branch 4 "3547459906" "gen_wr_req.state_q" (6) "gen_wr_req.state_q default,-,-,-,-"
+CHECKSUM: "530940107 4102526809"
+INSTANCE: tb.dut
+ANNOTATION: "intr_test_flds_we[0/1] is asserted at the same time whenever the intr_test register is written, so the combination of only one Qe being set is not possible."
```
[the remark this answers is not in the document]
antmarzam: Can you try to select the lines you're referring to? I have a hard time reading this exclusion file otherwise haha.
I think you mean lines 50-54? 

Your change would definitely make closing coverage easier. However, are we confident the register generator (or spec?) will stay as is? I'm not familiar with the tool. But we if we grew the interrupt vector, would that mean that `aon_timer.intr_test_q` would be ORed with all the separate `intr_test` fiedls? Is there a chance that if the `intr_test` was greater than 8-bit the RTL may generate a per-byte Qe? 

I guess your assertion would catch the difference if that will ever happen. 

Also, should we have a CCOV closure label for assertions that are just a logic check due to simplifying the logic to make closing coverage easier?

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
```diff
@@ -0,0 +1,186 @@
...
+src_busy_q feeds of the request, so by the time src_ack is high src_busy_q should also be high.
+See assertion SrcAckBusyChk_A for more details"
+Condition 3 "86646938" "(src_busy_q && src_ack) 1 -1" (1 "01")
+CHECKSUM: "2815520955 4109606122"
+INSTANCE: tb.dut.u_reg.u_wkup_count_hi_cdc.u_arb
+ANNOTATION: "It can't be hit because dst_update_ack won't be set prior to dst_update_req"
+Condition 3 "3080886878" "(gen_wr_req.dst_update_req && gen_wr_req.dst_update_ack) 1 -1" (1 "01")
+ANNOTATION: "It would need dst_req_i=1, dst_req_q=1, busy=1. However, that's not possible.
+
+The FSM stays in IDLE until there's a request, and  then stays in StWait 1-cycle. During StWait, dst_req_q is set to 0x1, but we are also transitioning back to StIdle where busy will be set to 0x0 making the missing condition coverage unhittable.
```
rswarbrick, line 121: I think we can see this happen?

- Suppose we are sitting in `StIdle` and nothing is going on.
- At cycle 0, suppose that `dst_req_i` is high.
- `dst_req_q` will get a nonblocking assignment of `1'b1`.
- Also (because `dst_req_i` is high) `dst_req` will be high too.
- As a result, `state_d` will be `StWait`.

- On cycle 1, `state_q` will be `StWait`
- As a result, `busy` will be true.
- Also, `dst_req_q` will be 1.

So we'll see a hit unless `dst_req_i` has dropped.

Is that correct? (And is it possible for this to happen?)

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
```diff
@@ -0,0 +1,186 @@
...
+"
+Condition 2 "1682497780" "(dst_req_i && ((!gen_wr_req.dst_req_q)) && gen_wr_req.busy) 1 -1" (2 "101")
+CHECKSUM: "2815520955 4109606122"
+INSTANCE: tb.dut.u_reg.u_wkup_count_lo_cdc.u_arb
+ANNOTATION: "It can't be hit because dst_update_ack won't be set prior to dst_update_req"
```
rswarbrick, line 126: I think this is the same comment as the one for line 117?

rswarbrick (commented):
I know that I haven't got to the bottom of the list, but I think there are a few items to work through so far :-)

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
antmarzam: This RTL change is in PR #25812 

I'll push the amended exclusion file in a sec

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
antmarzam: I've created a PR  for this: #25814

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
antmarzam: I don't think `dst_update_req` can't be unset until the `dst_update_ack` is set just based on how this FSM is working:
https://github.com/lowRISC/opentitan/blob/7780c6bf8aed527c1284c74e22d101b8bf256292/hw/ip/prim/rtl/prim_reg_cdc_arb.sv#L186-L228

Do you disagree?

rswarbrick: Looking now, I think I agree. But I guess the `prim_sync_reqack` module still couldn't have that assertion because it depends on the code around where the module is instantiated.

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
antmarzam: I'm still trying to convince myself of this case being unreachable. 

I don't think what you suggested can happen, because:

- `dst_req_i=1` can only occur for 1 cycle (even if you have back to back write accesses)
- If we're in `state_q = StWait` busy will be one, and that would set `dst_req_q` for the following cycle
- Once `dst_req_q` is set, `dst_req_i` is no longer set, but since `dst_req_q` is set, that would also set `dst_req`, which will then cause `dst_lat_d=1` and get `dst_req_q` cleared before the next `dst_req_i=1`.

Do you agree?

rswarbrick:
I think this is easy, but should these two commits be squashed?

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
rswarbrick: I think I probably agree. I have just spent ages working through it carefully myself and I *think* I'm convinced this can't happen.

Would you mind working through this logic with me and checking whether I'm missing something? I don't think it depends on `aon_timer` at all, so (if I'm right) I think the code that's currently:
```systemverilog
      end else if (dst_req_i && !dst_req_q && busy) begin
```
would be equivalent as
```systemverilog
      end else if (dst_req_i && busy) begin
```
(together with an assertion that says `dst_req_i |-> !dst_req_q`).

My logic is as follows:

- In `prim_reg_cdc`, we expect to see a request that comes from the SW
  as a signal on the `src` domain (either `src_we_i` or `src_re_i`).
- This gets synchronised to the destination clock by
  `u_src_to_dst_req`.
- This then appears to the arbiter as `dst_req_i`.
- If there is another request in flight (probably an update from HW),
  the SW request get stored in the `dst_req_q` register.
- The stored request gets cleared on the first posedge of the dst
  clock because it causes `dst_lat_d` to be high.
- To see `dst_req_i && dst_req_q`, we would need `prim_reg_cdc` to
  send another request before that happened.
- When the first SW request appears in `prim_reg_cdc`, we set `busy`
  until we see `src_ack`.
- This goes high when the SW request has been synchronised to the
  `dst` clock by `u_dst_update_sync`.
- As such, we can't see another request from `prim_reg_cdc` before
  `src_req_q` has been cleared. Clearing `src_req_q` takes exactly one
  clock crossing time and allowing `prim_reg_cdc` to send the second
  request takes the same time.

If you agree then we should probably file an issue to talk about this and get someone with more design experience than me to work through the reasoning as well.

antmarzam:
@rswarbrick I've removed the exclusions that have been addressed via RTL modifications and have squashed both commits together!

Also @vogelpi : Do you mind taking a look at [comment](https://github.com/lowRISC/opentitan/pull/25705#discussion_r1892272395) it would be helpful to get your two cents on it :)

hw/ip/aon_timer/dv/cov/aon_timer_unr_manually_excluded_excl.el
vogelpi: Sorry for the huge delay @rswarbrick and @antmarzam !

Thanks @rswarbrick for providing the pointer to the commit this originates from. I've briefly reviewed the original PR and the code and my conclusion is that using 2 bits for the two states is not related to some (incomplete) hardening approach or similar: There is no alert signal, no local lockup mechanism or anything.

I think the second state bit is just an oversight and I would recommend to remove that. But this means the unreachable `default` case will still be there. According to our style guide it must be there but in this particular case it can be empty as the section before the `unique_case` block contains defaults (keep the current state), see https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#case-statements for details.

antmarzam: Thanks for this! I've created a PR to simplify the state enum to be 1-bit wide only. 
https://github.com/lowRISC/opentitan/pull/26069

rswarbrick:
I've just worked through the discussion on this PR and I'm pretty convinced! :-)

@antmarzam: Do you think this exclusion list is correct for the current head of `master`? If so, I think this should be good to merge.

antmarzam:
Yeah I think it's ready to be merged :)

rswarbrick:
This is great, thanks!

rswarbrick (approved):

PR #25706, opened 2024-12-19
[rom_ctrl, dv] Excluded line coverage for prim_generic** blocks

KinzaQamar:


hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -39,6 +39,15 @@ Block 1 "3318159292" "data_o = 39'(data_i);"
 INSTANCE: tb.dut.u_tl_adapter_rom.u_tlul_data_integ_enc_instr.u_data_gen
 Block 1 "3318159292" "data_o = 39'(data_i);"
 
+INSTANCE: tb.dut.gen_rom_scramble_enabled.u_rom.u_rom.u_prim_rom.gen_generic.u_impl_generic
+ANNOTATION: "cfg_i is tied to 0 in tb"
```
rswarbrick, line 43: I think this needs to be `rom_cfg_i` (which is the name it has as a dut port). I'd probably also:
- Say something like "We expect this port to be wired to a constant."
- Add an assertion in `rom_ctrl.sv` that says this is constant. Maybe something like `ASSERT(CfgStable_A, $stable(rom_cfg_i))` ?

hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -39,6 +39,15 @@ Block 1 "3318159292" "data_o = 39'(data_i);"
...
+Block 1 "118728425" "assign unused_cfg = (^cfg_i);"
+
+INSTANCE: tb.dut.gen_rom_scramble_enabled.u_rom.u_seed_anchor.u_secure_anchor_buf.gen_generic.u_impl_generic
+ANNOTATION: "in_i and out_o are tied to parameters"
```
rswarbrick, line 47: I think this needs expanding a little to make it "less local". I'd probably suggest something like: "The u_seed_anchor module is a buffer whose inputs are constant because they are tied to parameters. As a result, this continuous assignment line cannot see execution."

rswarbrick (approved):
Thanks for the careful notes. This looks good to me.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/rom_ctrl/rtl/rom_ctrl.sv

The change in question is just adding a assertion, so there is no risk to the design.

vogelpi:
CHANGE AUTHORIZED: hw/ip/rom_ctrl/rtl/rom_ctrl.sv

The change in question is just adding a assertion. This is fine.

PR #25708, opened 2024-12-19
[SiVal] Map OTP and LC_CTRL tests in the testplans

engdoreis:


sw/device/silicon_creator/manuf/base/BUILD
```diff
@@ -380,6 +380,7 @@ filegroup(
         exec_env = {
             "//hw/top_earlgrey:fpga_hyper310_rom_with_fake_keys": None,
             "//hw/top_earlgrey:fpga_cw340_rom_with_fake_keys": None,
+            "//hw/top_earlgrey:fpga_cw340_sival": None,
```
moidx, line 383: We should keep an eye on this one as I am not sure it will work in `earlgrey_1.0.0`, but ok to merge as-is for now.

moidx (approved):

PR #25714, opened 2024-12-19
[bazel] Migrate all source repositories (non-Bazel aware) to bzlmod

jwnrt:
1. Remove two unused dependencies from `WORKSPACE.bzlmod`.
   * that should have been dropped in previous PRs but were missed.
2. Fix `#include "external/..."` paths to included directories in linker scripts.
   * These paths were never stable and have now changed with bzlmod enabled.
   * Directories are now added to the search path for includes.
3. Make `_repos()` functions private, run them from extensions, and extract repos.
4. Move patch files into `patches` subdirectories to match existing `third_party` convention.
5. Remove now-unused `http_archive_or_local` rule.
   * This is now implemented in bzlmod using `override_repo` and `inject_repo`.


third_party/llvm_compiler_rt/BUILD.llvm_compiler_rt.bazel
```diff
@@ -28,5 +28,8 @@ cc_library(
         "include/profile/InstrProfData.inc",
         "lib/profile/InstrProfiling.h",
     ],
-    includes = ["include"],
+    includes = [
+        "include",
+        "lib/profile",
```
pamaury, line 33: I feel like just using `includes` might make header names which are too generic. What about using `include_prefix` as well? For example if you add `include_prefix="external/llvm_compiler_rt"` and maybe "." to `includes` instead of `lib/profile` then you should be able to just keep the path intact.

jwnrt: Good idea, I didn't know about this parameter.

rules/repo.bzl
```diff
@@ -27,23 +27,14 @@ def http_archive_or_local(local = None, **kwargs):
...
-        # Although not particularly well documented, the `get_child` function will only add
-        # the workspace for relative paths, and simply ignore it if given an absolute path.
-        path = rctx.workspace_root.get_child(rctx.attr.local)
-
-        # NOTE By default, Bazel will automatically watch those directories and refetch
-        # the repository if they change.
-        for child in path.readdir():
-            rctx.symlink(child, child.basename)
-    elif rctx.attr.url:
+    if rctx.attr.url:
```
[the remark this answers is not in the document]
jwnrt: Ah yes, done

pamaury (approved):

PR #25717, opened 2024-12-19
[pwm] Correction to heartbeat mode

alees24:
Heartbeat direction reversal was premature in the event of passing the target duty cycle (non-wrapping case). The heartbeat direction shall only be changed at the end of the 'X'+1 pulse cycles for the current duty cycle.

This was found using the improved DV which is ongoing; the current DV environment overlooked this deviation from the specification.

### DIFF
--- a/hw/ip/pwm/rtl/pwm_chan.sv
+++ b/hw/ip/pwm/rtl/pwm_chan.sv
@@ -120,7 +120,7 @@ module pwm_chan #(
   logic [CntDw-1:0] dc_htbt_d;
   logic [CntDw-1:0] dc_htbt_q;
   logic dc_htbt_end;
-  // Asserted at the very end of the 'x' pulses cycles for which the heartbeat duty cycle
+  // Asserted at the very end of the 'x' pulse cycles for which the heartbeat duty cycle
   // remains unchanged; this is our signal to update the duty cycle in heartbeat mode.
   assign dc_htbt_end = htbt_mode & update & blink_end;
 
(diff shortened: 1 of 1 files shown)


hw/ip/pwm/rtl/pwm_chan.sv
```diff
@@ -139,19 +139,24 @@ module pwm_chan #(
...
   assign dc_htbt_min = pos_htbt ? duty_cycle_a_i : duty_cycle_b_i;
 
+  logic htbt_too_high;
+  logic htbt_too_low;
+  assign htbt_too_high = !htbt_falling & (dc_htbt_q >= dc_htbt_max);
+  assign htbt_too_low  =  htbt_falling & (dc_htbt_q <= dc_htbt_min);
+
+  // Does the heartbeat need to change direction for the next interval of 'x' pulse cycles?
+  logic htbt_reverse;
+  assign htbt_reverse = |{htbt_too_high, htbt_too_low, dc_wrap_q};  // Overflowed/underflowed?
```
rswarbrick, line 149: I agree with the code, but maybe it makes sense to expand this comment slightly and put it with the one two lines earlier. Otherwise it looks like it refers to all three terms in the OR instead of just the last one.

hw/ip/pwm/rtl/pwm_chan.sv
```diff
@@ -160,7 +165,7 @@ module pwm_chan #(
   always_comb begin
     if (pattern_repeat) begin
       {dc_wrap, dc_htbt_d} = {1'b0, duty_cycle_a_i};
-    end else if (htbt_falling) begin
+    end else if (htbt_falling ^ htbt_reverse) begin
```
rswarbrick, line 168: This is really nice! :-)

rswarbrick (commented):
A nitty comment, but this looks really nice to me. I particularly like the use of XOR, "taking reverse seriously" :-D

rswarbrick (approved):

rswarbrick:
CHANGE AUTHORIZED:  hw/ip/pwm/rtl/pwm_chan.sv

This causes a behaviour change, but only by fixing a bug! Looks like the right thing to me.

PR #25724, opened 2024-12-20
[dv,pattgen] Adjust pattern data distribution

elliotb-lowrisc:
Adjust the probabilities of the pattgen pattern data randomisation to reduce the chance of an all-zero half-pattern from 80% to 10%, and correspondingly increase the chance of a mixed half-pattern from 10% to 80%.

This should better exercise the block.


elliotb-lowrisc:
Apologies, I forgot to explore the effect this has on coverage numbers. I've converted it to draft for now.

elliotb-lowrisc:
Well, running the pattgen nightly tests locally seemed to show no change in coverage metrics, so I think I'll drop this for now.

PR #25727, opened 2024-12-20
[hw,spi_host,rtl] Remove configopts multi register

Razer6:
With a single configopts register, the register layout does not change when the user changes the `NumCS` parameter. Firmware, however needs to set up the config register ahead of a new transfer, but only when switching to a different device.

### DIFF
--- a/hw/ip/spi_host/data/spi_host.hjson
+++ b/hw/ip/spi_host/data/spi_host.hjson
@@ -317,81 +317,79 @@
              "excl:CsrAllTests:CsrExclCheck"]
 
     },
-    { multireg: { name: "CONFIGOPTS",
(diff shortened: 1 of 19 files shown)


a-will (commented):
This looks good to me! (modulo the little fixups that remain to get CI to pass)

a-will (approved):

andreaskurth:
CHANGE AUTHORIZED: hw/ip/spi_host/rtl/spi_host_reg_pkg.sv
CHANGE AUTHORIZED: hw/ip/spi_host/rtl/spi_host_reg_top.sv

Backward-incompatible changes aligned with current development stages for v3.0.0

PR #23200, opened 2024-05-20
[DRAFT} Suspend-sleep-resume tests for USBDEV

alees24:
This draft PR introduces a test framework and a set of top-level USBDEV tests to exercise all suspend/sleep states and transition stimuli from those states:

- Suspended - Resume Signaling on the USB causes exit from Suspended.
- Normal Sleep - Resume Signaling, Bus Reset or VBUS Disconnection cause exit from Normal Sleep.
- Deep Sleep - Resume Signaling, Bus Reset or VBUS Disconnection cause exit from Deep Sleep.

This code is intended to operate on CW310/340 FPGAs, in simulation with Verilator/DVsim (excluded from this PR for simplicity), and ES/Prod silicon (SiVal). Since this is based on the ES Sival branch, it is presently configured by a number of #defines for building and running on the ES Silicon.

In its earlier incarnation as draft PR #18564 (still open because of the simulation-related work that it contains, particularly the extensive changes to the USBDPI model), it exposed some issues with the USBDEV RTL and that has led to changes in the RTL for Prod. To run this test with the Prod RTL the #defines at the top of lib/testing/usb_testutils.h and lib/dif/dif_usbdev.h must be modified.

Test command for running the full test sequence:
```
bazel test --nocache_test_results --test_output=streamed --//signing:token=//signing/tokens:cloud_kms_sival //sw/device/tests:usbdev_suspend_full_test_silicon_owner_sival_rom_ext
```

The PR is presently a draft because it is felt that further tidying and testing is required before the 'usbdev_suspend_full_test' should be integrated into CI/SiVal automated testing, shielding both the host- and the device-side components from the variable timing of the other party.

PR #25728, opened 2024-12-20
[SiVal, usbdev] Suspend and Resume test.

engdoreis:
This PR is based on the PR: https://github.com/lowRISC/opentitan/pull/23200


jwnrt (approved):
Thanks.

The software team at lowRISC discussed this and decided that even though the diff is large and we couldn't review it all in depth, we're better off with the test coverage than without.

The non-test change to the DIF looks fine.

PR #25729, opened 2024-12-20
[reggen/doc] Clarify when accesses to unmapped addresses return errors

vogelpi:
Whether an error is typically returned depends on whether the requested address can be represented with the internally available address bits or not. This commit clarifies this with the help of an example.

This got reported by Amisha Srivastava from University of Texas at Dallas.

### DIFF
--- a/util/reggen/README.md
+++ b/util/reggen/README.md
@@ -637,7 +637,19 @@ The individual bundled wires are associated with the fields rather than the full
 
 ### Error responses
 
-Writes and reads that target addresses that are not represented within the register list typically return an error.
+Writes and reads that target addresses that are not mapped within the register list may return an error.
+Whether an error is returned depends on the target address and the number of address bits used internally:
+* Accesses to unmapped target addresses that can be fully represented with the internally used address bits typically return an error.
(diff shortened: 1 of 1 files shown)


util/reggen/README.md
```diff
@@ -637,7 +637,19 @@ The individual bundled wires are associated with the fields rather than the full
 
 ### Error responses
 
-Writes and reads that target addresses that are not represented within the register list typically return an error.
+Writes and reads that target addresses that are not mapped within the register list may return an error.
+Whether an error is returned depends on the target address and the number of address bits used internally:
+* Accesses to target addresses that can be fully represented with the internally used address bits typically return an error.
```
rswarbrick, line 642: Maybe add "unmapped" before target addresses here (because we don't return an error when it's a good address...)

rswarbrick (commented):
Looks sensible to me!

util/reggen/README.md
vogelpi: Thanks @rswarbrick ! This makes a lot of sense :-)

rswarbrick (approved):
Thanks!

PR #25731, opened 2024-12-20
[sram_ctrl,dv] Add cov exclusion for tlul_lc_gate

nasahlpa:
Excluding as we cannot reach the else condition in this module. Detailed description is available in the .el file.


rswarbrick (commented):
I'm really sorry but I'm not really convinced by this reasoning. I was trying to write up another version of the same thing and I *think* the coverage item might be attainable. I'm not convinced it's worth caring about, but...

My notes:
```systemverilog
// The condition we are excluding here is when outstanding_txn is nonzero when passing through the
// StErrorOutstanding state.
//
// For this to happen, we'd need to have an incomplete TL transaction pending (where the A side had
// happened but not the D side) when the gate became enabled, caused by the lc_en_i becoming On.
//
// The FSM state comes out of reset as StError but the lc_en_i signal is On at reset in sram_ctrl,
// so the FSM will follow the states StError (cycle 0), StErrorOutstanding (cycle 1), StActive
// (cycle 2). It *is* theoretically possible for a transaction to be enqueued in cycle zero because
// the TL inputs are top-level ports.
```

nasahlpa:
Thanks @rswarbrick! I assumed that issuing a transaction in the very first cycles is not possible. However, you are right - as we can control the input ports we certainly also can issue such a transaction (though not sure if this is possible with the current TL-UL DV environment). I've updated the description according to your suggestion.

rswarbrick:
I'm really sorry, but the text I put was me being lazy and copy-pasting from my notes when I was trying to prove the correctness of the waiver :-) Would you mind expanding on them a little and explaining why we're not worried about waiving it?

rswarbrick (approved):
Thanks for working through the reasoning so carefully. This looks great to me!

PR #25733, opened 2024-12-20
[bazel] Capture output of updatemem to make it less verbose

pamaury:
updatemem likes to print a lot of things which makes the CI output less readable for no particular benefit. This commit introduces a bash script that runs updatmem and captures the output. If an error occurs, it will print the path to the output and error logs. Those files are also added to the output groups of the bitstream_splice rules so that they can be queried if needed.

### DIFF
--- a/rules/opentitan/splice.bzl
+++ b/rules/opentitan/splice.bzl
@@ -37,6 +37,8 @@ def gen_vivado_mem_file(ctx, name, src, tool, swap_nibbles = True):
 
 def vivado_updatemem(ctx, name, src, instance, mmi, update, debug = False):
     spliced = ctx.actions.declare_file("{}.bit".format(name))
+    stdout = ctx.actions.declare_file("{}.out.log".format(name))
+    stderr = ctx.actions.declare_file("{}.err.log".format(name))
 
     # Vivado's `updatemem` only accepts bitstream filenames that end with `.bit`.
     # If the filename doesn't have that extension, symlink to a name that does.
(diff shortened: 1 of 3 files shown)


engdoreis (approved):

rules/opentitan/splice.bzl
```diff
@@ -57,17 +59,20 @@ def vivado_updatemem(ctx, name, src, instance, mmi, update, debug = False):
 
     ctx.actions.run(
         mnemonic = "SpliceBitstream",
-        outputs = [spliced],
+        outputs = [spliced, stdout, stderr],
```
a-will, line 62: If only they'd take care of https://github.com/bazelbuild/bazel/issues/5511

rules/scripts/run_and_capture.sh
```diff
@@ -0,0 +1,16 @@
...
+#
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Run the program and redirect the stdout and stderr to given files
+"$@" >"$RUN_CAPTURE_STDOUT" 2>"$RUN_CAPTURE_STDERR"
+res=$?
+if [ $res -ne 0 ]; then
```
a-will, line 11: Are you sure this actually works for updatemem? Do we know if it signals there are errors with a nonzero exit code?

I just ran it with arguments to files that don't exist, and the exit code was still 0.

pamaury: Indeed you are right, updatemem returns 0 even if the command line arguments are wrong... I was hoping this could handle both and the case where updatemem is not in the PATH but it seems it doesn't work. Unfortunately, this means that bazel sometimes only catches errors because the output file was not created.  I don't see a good way of handling this except for the script to itself check if the output file was created which seems... hackish?

Issue #24690, opened 2024-09-30
[hmac,dv] TB issue hmac_stress_all_with_rand_reset test "is_idle"


martin-velay:
### Description

`UVM_ERROR (hmac_base_vseq.sv:80): is_idle`
Command to reproduce with [bc90ee9322](https://github.com/lowrisc/opentitan/tree/bc90ee9322):
util/dvsim/dvsim.py hw/ip/hmac/dv/hmac_sim_cfg.hjson -i hmac_stress_all_with_rand_reset -t vcs --fixed-seed 50137718212529458093890268432036893881721916430145661069749074293950074877432

martin-velay:
Can be closed as [#25735](https://github.com/lowRISC/opentitan/pull/25735) has been merged

PR #25735, opened 2024-12-20
[hmac,dv] Move IDLE check from vseq to scb

martin-velay:
- this PR should fix this issue https://github.com/lowRISC/opentitan/issues/24690
- Move the IDLE pin check from the base sequence to the SCB as it makes more sense to perform check there.

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_scoreboard.sv
+++ b/hw/ip/hmac/dv/env/hmac_scoreboard.sv
@@ -88,6 +88,10 @@ task hmac_scoreboard::run_phase(uvm_phase phase);
       fork
         begin : main_thread
           fork
+            begin
+              cfg.clk_rst_vif.wait_clks(1);   // Wait one clk cycle to be sure IDLE state is set
+              check_idle(1'b1);               // Check IDLE after a reset has occured
+            end
             hmac_process_fifo_status();
             hmac_process_fifo_wr();
             hmac_process_fifo_rd();
--- a/hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
+++ b/hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
@@ -139,7 +139,6 @@ task hmac_base_vseq::dut_init(string reset_kind = "HARD");
   if (do_hmac_init) begin
     hmac_init();
   end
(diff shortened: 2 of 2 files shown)


vogelpi (approved):
This looks good to me!

Can this really fix the issues with the stress_all_with_rand_reset test?

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -82,6 +82,8 @@ task hmac_scoreboard::run_phase(uvm_phase phase);
   wait(cfg.under_reset);
   forever begin
     wait(!cfg.under_reset);
+    cfg.clk_rst_vif.wait_clks(1);   // Wait one clock cycle to be sure IDLE state will be set
+    check_idle(1'b1);               // Check IDLE after a reset has occured
```
rswarbrick, line 86: This looks sensible, but would it make more sense put put it in the fork below? That way, tasks like `hmac_process_fifo_status` don't have to wait before they start.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -139,7 +139,6 @@ task hmac_base_vseq::dut_init(string reset_kind = "HARD");
   if (do_hmac_init) begin
     hmac_init();
   end
-  `DV_CHECK_EQ(cfg.hmac_vif.is_idle(), 1'b1)
```
rswarbrick, line 142: I'm extremely enthusiastic about taking this sort of thing out of the sequences. Thanks!

martin-velay:
> This looks good to me!
> 
> Can this really fix the issues with the stress_all_with_rand_reset test?

It fixes only one of the issues, there are 1-2 other PR to handle some other parts . I ran 100 times this test with all the cherry-picked commits and no issue found.

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: Good point, thanks !

rswarbrick (approved):
Thanks: this looks much nicer to me.

marcelocarvalhoLowRisc (commented):
Totally agreed! Much better in the scoreboard.

PR #25738, opened 2024-12-23
[hmac,dv] Skip msg_length check when reset

martin-velay:
- when a reset a is happening, the HMAC DV env is not really robust. Some tasks are still ongoing and would require to be killed. A temporary fix has been proposed to skip the msg_length check when a Save and Restore task is ongoing when a reset is triggered. A better approach will be put in place later when a proper reset agent will be used.


rswarbrick (commented):
This looks sensible to me, but would it also make sense to open an issue and leave a TODO message here pointing at that issue?

martin-velay:
I have added the TODO and created an issue to track this : https://github.com/lowRISC/opentitan/issues/25809

rswarbrick (approved):

rswarbrick (approved):

PR #25739, opened 2024-12-23
[bazel] Update to version 7.4.1

pamaury:
Bazel 7.4 has the new [`string_keyed_label_dict`](https://bazel.build/versions/7.4.0/rules/lib/toplevel/attr#string_keyed_label_dict) attribute which could be useful for #25580. It can be replaced by `label_keyed_string_dict` as a work around but it's really the wrong data structure and I would prefer to avoid that we keep using the wrong one for the job.

There is also https://github.com/bazelbuild/bazel/pull/23585 could be very useful when we start adding bazel extensions.


jwnrt (commented):
LGTM but I’m surprised that the lockfile changed so much. Any ideas what caused that?

pamaury:
> LGTM but I’m surprised that the lockfile changed so much. Any ideas what caused that?

No idea, that's quite weird. The only thing that comes to mind is that they either updated the format or they changed something in the resolution algorithm?

jwnrt (approved):

PR #25740, opened 2024-12-23
[sram_ctrl,dv] Add post apply reset delay

nasahlpa:
After resetting, we should wait a bit before starting the new test to make sure that the tl_agent message queue really is emptied. If we are not doing this, we might see the following error:

```
./util/dvsim/dvsim.py hw/ip/sram_ctrl/dv/sram_ctrl_main_sim_cfg.hjson \
  -i sram_ctrl_passthru_mem_tl_intg_err -t vcs --fixed-seed \
  41020402185315011969996404882728929238760707320232330314609988937493640571488

UVM_ERROR @ 2116319532 ps: uvm_test_top.env.m_tl_agent_sram_ctrl_regs_reg_block
sequencer [uvm_test_top.env.m_tl_agent_sram_ctrl_regs_reg_block.sequencer] get_
next_item/try_next_item called twice without item_done or get in between
```

### DIFF
--- a/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
+++ b/hw/ip/sram_ctrl/dv/env/seq_lib/sram_ctrl_base_vseq.sv
@@ -30,6 +30,12 @@ class sram_ctrl_base_vseq #(
   // enabled at the beginning of the test.
   bit init_w_readback = 1'b0;
 
+  // The extra cycles to wait after reset before starting any test, required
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
This seems reasonable, but it's a bit sad that the TL agent queue doesn't get flushed as we *enter* reset, so it's quite surprising that this is needed.

Is there a good reason for that architecture? If not, maybe it makes sense to tag this with a TODO and open an issue about the fact that we've got left-over state in the TL agent that takes some time to clear.

PR #25744, opened 2024-12-23
[ci] Use alternative mechanisms to avoid github API file limits

a-will:
Due to the file number limit for `gh pr diff`, change to comparing the checkout against the base branch.

### DIFF
--- a/.github/workflows/pr_change_check.yml
+++ b/.github/workflows/pr_change_check.yml
@@ -22,7 +22,9 @@ jobs:
         run: |
           pr_url="https://github.com/${{ github.repository }}/pull/${{ github.event.number }}"
           echo $pr_url
-          gh pr diff $pr_url --name-only > $HOME/changed_files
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_change_check.yml
```diff
@@ -22,7 +22,9 @@ jobs:
         run: |
           pr_url="https://github.com/${{ github.repository }}/pull/${{ github.event.number }}"
           echo $pr_url
-          gh pr diff $pr_url --name-only > $HOME/changed_files
+          gh pr checkout $pr_url --name-only > $HOME/changed_files
```
nbdd0121, line 25: Would it be better to change the checkout sha above to
```
      with:
        ref: refs/pull/${{ github.event.pull_request.number }}/merge
        fetch-depth: 2 # include merge commit parents
```

which should allow direct diff without needing to checkout again?

a-will:
I must've run out of time for this, and it's probably obsolete now. Closing...

PR #25749, opened 2024-12-25
[RACL] Parse mappings and auto-connect RACL

Razer6:


util/raclgen/lib.py
```diff
@@ -50,3 +57,46 @@ def compute_policy_value(permission: str) -> int:
...
+                       if_name: Optional[str], ip_block: IpBlock):
+    mapping = _read_hjson(mapping_path)
+    parsed_register_mapping = {}
+
+    # Mapping must be a dict with a single entry:
+    # RACL_GROUP => register mapping
+    if len(mapping) != 1:
+        raise SystemExit('Mapping file must be a single-element dict mapping the RACL group to the '
+                         'register mapping')
+    racl_group = list(mapping.keys())[0]
```
rswarbrick, line 72: We might want to check that `racl_group` is a string too?

util/raclgen/lib.py
```diff
@@ -50,3 +57,46 @@ def compute_policy_value(permission: str) -> int:
...
+    register_mapping = list(mapping.values())[0]
```
rswarbrick, line 73: Since we know the length, this might be easier with something like `racl_group, register_mapping = mapping.items()[0]` ?

rswarbrick, line 73: Do we also need to check that `register_mapping` is a dict?

util/raclgen/lib.py
```diff
@@ -50,3 +57,46 @@ def compute_policy_value(permission: str) -> int:
...
+
+    # Special handling of the all star assignment:
+    # "*": POLICY
+    # Assigns all registers to a given policy
+    if len(register_mapping) == 1 and list(register_mapping.keys())[0] == "*":
```
rswarbrick, line 78: Possibly easier: `if list(register_mapping.keys()) == ["*"]` ?

util/raclgen/lib.py
```diff
@@ -50,3 +57,46 @@ def compute_policy_value(permission: str) -> int:
...
+        policy_idx = -1
+        for idx, policy in enumerate(racl_config['policies'][racl_group]):
+            if policy['name'] == policy_name:
+                policy_idx = idx
+                break
+
+        if policy_idx == -1:
+            raise SystemExit(f'RACL policy {policy_name} not defined in RACL config')
+
+        if if_name not in ip_block.reg_blocks:
```
rswarbrick, line 93: I'd probably be inclined to assign `ip_block.reg_blocks.get(if_name)` to something and then check for `None` here, which avoids needing to do the look-up again as the next operation.

util/topgen.py
```diff
@@ -528,13 +528,24 @@ def generate_ac_range_check(topcfg: Dict[str, object], out_path: Path) -> None:
...
     if 'racl_config' not in topcfg:
         return
 
+    # Read the top-level RACL information
     topcfg['racl'] = parse_racl_config(topcfg['racl_config'])
 
+    # Generate the RACL mappings for all subscribing IPs
+    for m in topcfg['module']:
+        if 'racl_mappings' in m:
+            for if_name, mapping_path in m['racl_mappings'].items():
```
rswarbrick, line 543: Maybe `for if_name, mapping_path in m.get('racl_mappings', {}).items()`, which will avoid a layer of indentation.

rswarbrick, line 543: Maybe `for if_name, mapping_path in m.get('racl_mappings', {}).items()`, which will avoid a layer of indentation.

util/topgen/intermodule.py
```diff
@@ -248,6 +248,32 @@ def autoconnect(topcfg: OrderedDict, name_to_block: Dict[str, IpBlock]):
     for xbar in topcfg["xbar"]:
         autoconnect_xbar(topcfg, name_to_block, xbar)
 
+    # Auto connect RACL subscribing IPs to the assocated racl_ctrl IP
+    if 'racl' in topcfg:
+        # Find the RACL control of the defined group. This currently only works for the default
+        # RACL ctrl module when there is a single instance. This limitiation currently comes from
```
rswarbrick, line 254: Nit: Spelling of "limitation"

util/topgen/merge.py
```diff
@@ -240,6 +240,29 @@ def elaborate_instance(instance, block: IpBlock):
...
 
+    # An instance can either have a 'racl_mapping' or 'racl_mappings' but can't have both.
+    # 'racl_mapping' is used when the device only has a single register interface and
+    # 'racl_mappings' there are more.
+    racl_mapping = instance.get('racl_mapping')
+    racl_mappings = instance.get('racl_mappings')
+
+    if racl_mapping and racl_mappings:
+        raise ValueError("Cannot specify both 'racl_mapping' and 'racl_mappings'")
+
```
rswarbrick, line 251: I'd suggest converting `racl_mapping` to a value for `racl_mappings` here, which will make the code below a bit easier to read.

rswarbrick (commented):
A partial review so far, sorry (just the first two commits)

hw/ip/pwm/rtl/pwm_reg_top.sv
```diff
@@ -3112,121 +3147,184 @@ module pwm_reg_top (
     addr_hit[20] = (reg_addr == PWM_BLINK_PARAM_3_OFFSET);
     addr_hit[21] = (reg_addr == PWM_BLINK_PARAM_4_OFFSET);
     addr_hit[22] = (reg_addr == PWM_BLINK_PARAM_5_OFFSET);
+
+    if (EnableRacl) begin : gen_racl_hit
+      racl_addr_hit_read [ 0] = addr_hit[ 0] & (|(racl_policies_i[RaclPolicySelVec[ 0]].read_perm  & racl_role_vec));
```
rswarbrick, line 3152: I've not tried this, but could we move the looping into the SystemVerilog instead of the Python? I think we end up with something like this, which is a bit more obvious to read, rather shorter and doesn't need line length waivers.
```systemverilog
    if (EnableRacl) begin : gen_racl_hit
      for (int unsigned slice_idx = 0; slice_idx < 23; slice_idx++) begin
        racl_addr_hit_read[slice_idx] =
            addr_hit[slice_idx] & (|(racl_policies_i[RaclPolicySelVec[slice_idx]].read_perm
                                     & racl_role_vec));
        racl_addr_hit_write[slice_idx] =
            addr_hit[slice_idx] & (|(racl_policies_i[RaclPolicySelVec[slice_idx]].write_perm
                                     & racl_role_vec));
      end
    end else begin : gen_no_racl
      racl_addr_hit_read  = addr_hit;
      racl_addr_hit_write = addr_hit;
    end
```

hw/ip/pwm/lint/pwm.vbl
```diff
@@ -0,0 +1,6 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# These lines are too long due to generated code
```
rswarbrick, line 5: As a note: if we can simplify the generated reg-top as I'm suggesting then we won't need this waiver entry.

hw/ip/pwm/rtl/pwm.sv
```diff
@@ -23,23 +26,35 @@ module pwm
...
     .rst_core_ni,
-    .tl_i       (tl_i),
-    .tl_o       (tl_o),
-    .reg2hw     (reg2hw),
+    .tl_i                  (tl_i),
+    .tl_o                  (tl_o),
+    .reg2hw                (reg2hw),
+    .racl_policies_i       (racl_policies_i),
+    .racl_error_o          (racl_error_o),
+    .racl_error_log_o      (racl_error_log_o),
```
rswarbrick, line 55: It's not a big issue, but do we really need to go that far to the right? I think I'd probably have expected just one space after the longest name (this one).

hw/ip/pwm/data/pwm.hjson
```diff
@@ -87,6 +87,37 @@
...
     }
   ]
+  inter_signal_list: [
+    { struct:  "racl_policy_vec",
+      type:    "uni",
+      name:    "racl_policies",
+      act:     "rcv",
+      package: "top_racl_pkg",
+      desc:    '''
+        Policy vector distributed to the subscribing RACL IPs.
```
rswarbrick, line 97: I wonder whether this text needs changing slightly. It looks a bit like PWM is the block that's distributing something (which is backwards). If I've got this right, PWM *is* a subscribing IP, so maybe something like "RACL policy vector, used by the IP to control register access permissions." ?

Razer6: True. Probably copied that from the racl_ctrl module. Changed that.

hw/ip_templates/ac_range_check/rtl/ac_range_check.sv.tpl
```diff
@@ -6,14 +6,20 @@ module ${module_instance_name}
   import tlul_pkg::*;
   import ${module_instance_name}_reg_pkg::*;
 #(
-  parameter logic [NumAlerts-1:0] AlertAsyncOn = {NumAlerts{1'b1}}
+  parameter logic [NumAlerts-1:0] AlertAsyncOn = {NumAlerts{1'b1}},
+  parameter bit                   RaclErrorRsp         = 1'b1,
```
rswarbrick, line 10: Nit: I'd probably line the `=` sign up with the one above. Presumably, the `${3 + 5*num_ranges}` expands to the right number of digits to line up downwards, but that seems a bit magical.

hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
```diff
@@ -0,0 +1,128 @@
...
+  );
+
+  if (EnableRacl) begin : gen_racl_role_logic
+    // Retrieve RACL role from user bits and one-hot encode that for the comparison bitmap
+    top_racl_pkg::racl_role_t racl_role;
+    assign racl_role = top_racl_pkg::tlul_extract_racl_role_bits(tl_i.a_user.rsvd);
+
+    top_racl_pkg::racl_role_vec_t racl_role_vec;
+    prim_onehot_enc #(
+      .OneHotWidth( $bits(prim_onehot_enc) )
```
rswarbrick, line 91: Is this correct? I don't think that `prim_onehot_enc` is a variable is it? Maybe this was supposed to be `racl_role_vec`?

But I'd probably use the type explicitly, avoiding use of the variable at all. Maybe `$bits(top_racl_pkg::racl_role_vec_t)`?

hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  // Not all RACL policies are used, even if RACL is enabled
+  logic unused_policy_sel;
+  assign unused_policy_sel = ^racl_policies_i;
+
+  assign we_o = wr_req & racl_write_allowed;
+  assign re_o = rd_req & racl_read_allowed;
+  // Mask read data in case of a RACL violation
+  assign rdata = racl_violation? '1 : rdata_i;
```
rswarbrick, line 126: Nit: I think ternaries normally have a space before the `?`.

util/raclgen/lib.py
```diff
@@ -50,3 +57,51 @@ def compute_policy_value(permission: str) -> int:
...
+    if not isinstance(racl_group, str):
+        raise SystemExit('RACL group must be a string')
+    if not isinstance(register_mapping, dict):
+        raise SystemExit('Register mapping must be a a dict')
+
+    # Special handling of the all star assignment:
+    # "*": POLICY
+    # Assigns all registers to a given policy
+    if list(register_mapping.keys()) == ["*"]:
+        policy_name = list(register_mapping.values())[0]
```
rswarbrick, line 98: I agree that this is correct, but `policy_name = register_mapping["*"]` is probably more obvious?

util/raclgen/lib.py
```diff
@@ -50,3 +57,51 @@ def compute_policy_value(permission: str) -> int:
...
+            raise SystemExit(f'RACL group {racl_group} not defined in RACL config')
+
+        policy_idx = -1
+        for idx, policy in enumerate(racl_config['policies'][racl_group]):
+            if policy['name'] == policy_name:
+                policy_idx = idx
+                break
+
+        if policy_idx == -1:
+            raise SystemExit(f'RACL policy {policy_name} not defined in RACL config')
```
rswarbrick, line 110: Doesn't this need to also say which RACL group it was searching?

util/topgen/merge.py
```diff
@@ -240,6 +240,29 @@ def elaborate_instance(instance, block: IpBlock):
...
+    if racl_mapping:
+        racl_mappings_dict[None] = racl_mapping
+
+    for if_name, racl_mapping in racl_mappings.items():
+        racl_mappings_dict[if_name] = racl_mapping
+
+    # Delete the racl_mapping entry as its converted to a unified racl_mappings entry
+    instance.pop('racl_mapping', None)
+    if racl_mappings:
+        instance['racl_mappings'] = racl_mappings_dict
```
rswarbrick, line 264: Thanks for tweaking this a bit, but I think it can be even simpler. Can't we do something like this? (completely untested!)
```python
racl_mapping = instance.get('racl_mapping')
if racl_mapping is not None:
  if instance.get('racl_mappings') is not None:
    raise ValueError("Cannot specify both 'racl_mapping' and 'racl_mappings'")
  del instance['racl_mapping']
  instance['racl_mappings'] = {None: racl_mapping}
```

util/topgen/validate.py
```diff
@@ -216,6 +216,9 @@
     'outgoing_alert': ['s', 'optional string to indicate alerts are routed externally to the named '
                             'group'],
     'incoming_interrupt': ['g', 'Parsed incoming interrupts (generated)'],
+    'racl_mapping': ['s', 'optional RACL mapping for that IP'],
+    'racl_mappings': ['g', 'dict of RACL mappings for all registers interfaces if there are more '
+                      'than one'],
```
rswarbrick, line 221: I'd probably list these in the other order, with something like:
- `racl_mappings`: dict that maps an interface to its associated RACL mapping.
- `racl_mapping`: A special case of racl_mappings. If specified, this is taken to represent a dict that associates all interfaces with the give mapping. It is an error to specify both this and racl_mappings.`

hw/ip/i2c/rtl/i2c.sv
```diff
@@ -25,6 +28,11 @@ module i2c
   input  prim_alert_pkg::alert_rx_t [NumAlerts-1:0] alert_rx_i,
   output prim_alert_pkg::alert_tx_t [NumAlerts-1:0] alert_tx_o,
 
+  // RACL interface
+  input  top_racl_pkg::racl_policy_vec_t  racl_policies_i,
```
rswarbrick, line 32: Nit: This clearly doesn't matter but I noticed it as I read through: I think there's a spare space before this list of indented ports. (But really don't bother thinking about this if you're not looking at the file anyway!)

hw/ip_templates/racl_ctrl/rtl/racl_ctrl.sv.tpl
```diff
@@ -4,7 +4,8 @@
 
 module ${module_instance_name} import ${module_instance_name}_reg_pkg::*; #(
   parameter logic [NumAlerts-1:0] AlertAsyncOn      = {NumAlerts{1'b1}},
-  parameter int unsigned          NumSubscribingIps = 1
+  parameter int unsigned          NumSubscribingIps = 1,
+  parameter bit                   RaclErrorRsp       = 1'b1
```
rswarbrick, line 8: Nit: I think there's a stray space here.

hw/ip_templates/ac_range_check/rtl/ac_range_check.sv.tpl
```diff
@@ -6,14 +6,20 @@ module ${module_instance_name}
...
+  parameter int unsigned          RaclPolicySelVec[${3 + 5*num_ranges}] = '{${3 + 5*num_ranges}{0}}
 ) (
   input  logic                                      clk_i,
   input  logic                                      rst_ni,
   input  logic                                      rst_shadowed_ni,
   // Alerts
   input  prim_alert_pkg::alert_rx_t [NumAlerts-1:0] alert_rx_i,
   output prim_alert_pkg::alert_tx_t [NumAlerts-1:0] alert_tx_o,
+  // RACL interface
+  input  top_racl_pkg::racl_policy_vec_t            racl_policies_i,
```
a-will, line 20: This discussion would be for a later PR, but I suspect at some point, we'll want to pass this type from a parameter, since it can vary in definition with the racl config's "nr_policies."

Razer6: I remember that can even be passed as a parameter 😀

Razer6:
@rswarbrick Can you take another look to get that one merged?

hw/ip_templates/ac_range_check/rtl/ac_range_check.sv.tpl
rswarbrick: Duh, sorry: That review comment was extremely hard to understand! What I mean is that the code is currently:
```systemverilog
  parameter logic [NumAlerts-1:0] AlertAsyncOn         = {NumAlerts{1'b1}},
  parameter bit                   RaclErrorRsp         = 1'b1,
  parameter int unsigned          RaclPolicySelVec[${3 + 5*num_ranges}] = '{${3 + 5*num_ranges}{0}}
```
which has presumably been done so that the equals signs all line up nicely. But of course, that depends on the number of digits in `3 + 5*num_ranges`.

I was trying to suggest a stupider approach, where we use one space for each of the first two parameters and then the third one isn't necessarily going to line up. Eh! Never mind!

If we want to be cleverer, we could do something like rendering the count to a string and then replicating spaces as an indent of the right length. But that seems overly complicated: I was trying to suggest the easy option :-)

Razer6: Let's be stupid about it. In the end, it's auto-generated code, and if that is our biggest problem.. you know ;-).
Ideally, we would pipe all auto-generated code through a tool like `verible-format` or what I learned this week, `clang-cormat` also supports SystemVerilog recently. Then, we don't have to deal with all that fancy code to deal with alignments.

hw/ip_templates/racl_ctrl/rtl/racl_ctrl.sv.tpl
```diff
@@ -123,28 +123,29 @@ module ${module_instance_name} import ${module_instance_name}_reg_pkg::*; #(
...
   // for the corresponding RACL role or Write/Read information.
+  `ASSERT(OneHotRaclError_A, $onehot0(racl_error_i))
+
   logic racl_error;
   assign racl_error = |racl_error_i;
 
   top_racl_pkg::racl_role_t racl_error_role;
   top_racl_pkg::ctn_uid_t racl_error_ctn_uid;
-  logic racl_error_write_read;
+  logic racl_error_read_not_write;
```
matutem, line 144: I don't think this change helps clarify what this bit means: 1 == read, 0 == write? If so, maybe call it read_access?

hw/top_darjeeling/templates/toplevel.sv.tpl
```diff
@@ -483,6 +483,14 @@ max_intrwidth = (max(len(x.name) for x in block.interrupts)
 %>\
   % if m["param_list"] or block.alerts:
   ${m["type"]} #(
+  % if 'racl_mappings' in m:
+    .EnableRacl(1'b1),
+    .RaclErrorRsp(${"1'b1" if top['racl']['error_response'] else "1'b0"}),
+    % for if_name in m['racl_mappings'].keys():
+<% if_prefix = f"_{if_name.upper()}" if if_name else "" %>\
```
matutem, line 490: This is used as a suffix, so please rename it if_suffix.

hw/top_earlgrey/templates/toplevel.sv.tpl
```diff
@@ -502,6 +502,14 @@ max_intrwidth = (max(len(x.name) for x in block.interrupts)
 %>\
   % if m["param_list"] or block.alerts:
   ${m["type"]} #(
+  % if 'racl_mappings' in m:
+    .EnableRacl(1'b1),
+    .RaclErrorRsp(${"1'b1" if top['racl']['error_response'] else "1'b0"}),
+    % for if_name in m['racl_mappings'].keys():
+<% if_prefix = f"_{if_name.upper()}" if if_name else "" %>\
```
matutem, line 509: Same: rename if_suffix.

util/raclgen/lib.py
```diff
@@ -22,13 +25,17 @@
...
 
 
-def parse_racl_config(config_path: str) -> Dict[str, object]:
+def _read_hjson(filename: str) -> Dict[str, object]:
     try:
-        with open(config_path, 'r') as f_racl_config:
-            racl_config = hjson.load(f_racl_config)
+        with open(filename, 'r') as f_racl_config:
+            return hjson.load(f_racl_config)
     except OSError:
```
matutem, line 34: Please also add "except ValueError", log.error so the error is printed, and raise SystemExit, unless  you are okay with a ValueError generating a backtrace.

util/raclgen/lib.py
```diff
@@ -50,3 +57,52 @@ def compute_policy_value(permission: str) -> int:
             policy['rd_default'] = compute_policy_value('allowed_rd')
             policy['wr_default'] = compute_policy_value('allowed_wr')
     return racl_config
+
+
+def parse_racl_mapping(racl_config: Dict[str, object], mapping_path: str,
+                       if_name: Optional[str], ip_block: IpBlock):
```
matutem, line 78: Add return type? -> Dict[str, int]:

util/raclgen/lib.py
```diff
@@ -50,3 +57,52 @@ def compute_policy_value(permission: str) -> int:
...
+    mapping = _read_hjson(mapping_path)
+    parsed_register_mapping = {}
+
+    # Mapping must be a dict with a single entry:
+    # RACL_GROUP => register mapping
+    if len(mapping) != 1:
+        raise SystemExit('Mapping file must be a single-element dict mapping the RACL group to the '
```
matutem, line 85: Apparently you have not run util/lintpy.py on this file: it will suggest reformatting this line, and a few others perhaps. It is good practice to run it since it makes the code compliant with the style guide.

hw/ip/uart/rtl/uart_reg_top.sv
```diff
@@ -15,6 +20,11 @@ module uart_reg_top (
   output uart_reg_pkg::uart_reg2hw_t reg2hw, // Write
   input  uart_reg_pkg::uart_hw2reg_t hw2reg, // Read
 
+  // RACL interface
+  input  top_racl_pkg::racl_policy_vec_t racl_policies_i,
```
matutem, line 24: I imagine there will be tops with different values of NrRaclPolicies, so this seems to make this (and any other IP) top-dependent. Could you clarify?

hw/ip_templates/racl_ctrl/rtl/racl_ctrl.sv.tpl
Razer6: Exactly. This encoding is documented in `top_racl_pkg.sv` at the struct definition. I renamed the field to access type to remove the confusion.

hw/top_darjeeling/templates/toplevel.sv.tpl
Razer6: For sure!

hw/ip/uart/rtl/uart_reg_top.sv
Razer6: Yes. It does in the short term. Long term, I plan to decouple that by adding the type as a parameter to the IP. Long term is next month, so probably not that long.

Razer6 (commented):
Thanks @matutem for the review. For the top dependency, I plan to decouple that but probably not in this PR.

hw/ip/uart/rtl/uart_reg_top.sv
matutem: Passing NrRaclPolicies as a parameter is probably all that is needed, and that can be handled as a regular parameter, meaning won't require ipgen. Indeed all CSR blocks will be top specific, but "mildly".

hw/top_englishbreakfast/rtl/autogen/top_racl_pkg.sv
```diff
@@ -39,12 +39,21 @@ package top_racl_pkg;
   // RACL policy vector for distributing RACL policies from the RACL widget to the subscribing IP
   typedef racl_policy_t [NrRaclPolicies-1:0] racl_policy_vec_t;
```
matutem, line 40: Could you just set NrRaclPolicies == 1 for earlgrey, so the width of RACL_POLICY_* becomes 4? Not necessary in this PR.

util/reggen/reg_top.sv.tpl
```diff
@@ -718,11 +735,11 @@ ${finst_gen(sr, field, finst_name, fsig_name, fidx)}
   assign racl_error_log_o.racl_role  = racl_role;
 
   if (EnableRacl) begin : gen_racl_log
-    assign racl_error_log_o.ctn_uid        = top_racl_pkg::tlul_extract_ctn_uid_bits(tl_i.a_user.rsvd);
-    assign racl_error_log_o.read_not_write = tl_i.a_opcode == tlul_pkg::Get;
+    assign racl_error_log_o.ctn_uid     = top_racl_pkg::tlul_extract_ctn_uid_bits(tl_i.a_user.rsvd);
+    assign racl_error_log_o.access_type = tl_i.a_opcode == tlul_pkg::Get;
```
matutem, line 739: Sorry, this name is still not clear enough. If access_type is boolean means for example access_type == false for a write ??

If you name this "read_access" then a bool would make sense, true for reads, false for writes. Ideally this would be an enum... up to you.

util/topgen/merge.py
```diff
@@ -240,6 +240,16 @@ def elaborate_instance(instance, block: IpBlock):
...
 
+    # An instance can either have a 'racl_mapping' or 'racl_mappings' but can't have both.
+    # 'racl_mapping' is used when the device only has a single register interface and
+    # 'racl_mappings' when there are more. Translate to always use unified racl_mappings entry.
+    racl_mapping = instance.get('racl_mapping')
+    if racl_mapping is not None:
+        if instance.get('racl_mappings') is not None:
+            raise ValueError("Cannot specify both 'racl_mapping' and 'racl_mappings'")
+        del instance['racl_mapping']
+        instance['racl_mappings'] = {None: racl_mapping}
```
matutem, line 251: A dictionary with "None" as a key can be confusing. Is there a better name for this, like "default" or some such.

util/topgen/templates/toplevel_racl_pkg.sv.tpl
```diff
@@ -31,12 +31,21 @@ package top_racl_pkg;
...
+  // Default ROT Private write policy value
+  parameter racl_role_vec_t RACL_POLICY_ROT_PRIVATE_WR = 16'h${f"{racl_config['rot_private_policy_wr']:x}"};
+
   // RACL information logged in case of a denial
   typedef struct packed {
     racl_role_t racl_role;
     ctn_uid_t   ctn_uid;
     // 0: Write access, 1: Read access
-    logic       read_not_write;
+    logic       access_type;
```
matutem, line 48: The comment suggests the name should be `read_access`.

util/reggen/reg_top.sv.tpl
Razer6: Fair enough. I changed to `read_access`. I left it as a `logic` atm since the only assignment to it is via `assign racl_error_log_o.read_access = tl_i.a_opcode == tlul_pkg::Get;`.

hw/top_englishbreakfast/rtl/autogen/top_racl_pkg.sv
Razer6: It's set to 1 or am I missing something? In the meanwhile, I added the RACL bit definition to the HJSON and defer all bit-widths out from that instead of hard coding them and where they are located in the TLUL user bit vector.

util/topgen/merge.py
Razer6: It can but this would add an assumption on how a TLUL interface can be named / or not. I think this one was done in the same way as for the base address. I am happy to change that in a follow up PR.

hw/ip/uart/rtl/uart_reg_top.sv
rswarbrick: What @Razer6 suggests seems reasonable. Another way we could do this (also definitely for a later PR) would be to template `top_racl_pkg` for the toplevel in question. I'm not completely sure I know which would be cleaner, but I suspect either would work.

rswarbrick (approved):
I've just worked through this again and I'm pretty convinced! I know that there are a couple of pending items from @matutem, but I'm very happy when we're ready with them.

util/topgen/merge.py
matutem: I hope you consider adding a default racl object in the top config, and when an IP doesn't assign one we give it the default. This has no name clash issues.

matutem (approved):
LGTM, but there are items that can be addressed in subsequent PRs.

util/topgen/merge.py
Razer6: We agreed on `core` being the default.

vogelpi:
CHANGE AUTHORIZED: hw/ip/i2c/data/i2c.hjson
CHANGE AUTHORIZED: hw/ip/i2c/rtl/i2c_reg_top.sv
CHANGE AUTHORIZED: hw/ip/pwm/data/pwm.hjson
CHANGE AUTHORIZED: hw/ip/uart/data/uart.hjson
CHANGE AUTHORIZED: hw/ip/uart/rtl/uart_reg_top.sv

This PR changes hardware blocks used in Earlgrey as well as Earlgrey but the functionality doesn't change as RACL is still disabled in Earlgrey, i.e., these changes have no functional impact.

matutem:
CHANGE AUTHORIZED: hw/ip/i2c/data/i2c.hjson
CHANGE AUTHORIZED: hw/ip/i2c/rtl/i2c_reg_top.sv
CHANGE AUTHORIZED: hw/ip/pwm/data/pwm.hjson
CHANGE AUTHORIZED: hw/ip/uart/data/uart.hjson
CHANGE AUTHORIZED: hw/ip/uart/rtl/uart_reg_top.sv

hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
```diff
@@ -0,0 +1,123 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+`include "prim_assert.sv"
+
+/**
+ * Tile-Link UL adapter for Register interface with RACL protection
+ */
+
+module tlul_adapter_reg_racl
```
vogelpi, line 11: https://opentitan.org/book/doc/project_governance/checklist/index.html#assert_known_added

vogelpi, line 11: This module doesn't seem to be used yet. Is this intentional?

Razer6: Yes. It will be used in the mailbox in conjunction with register windows. @davidschrammel is working on that, that will be a follow up when adding RACL to the mailbox.

hw/ip/i2c/rtl/i2c.sv
```diff
@@ -25,6 +28,11 @@ module i2c
   input  prim_alert_pkg::alert_rx_t [NumAlerts-1:0] alert_rx_i,
   output prim_alert_pkg::alert_tx_t [NumAlerts-1:0] alert_tx_o,
 
+  // RACL interface
+  input  top_racl_pkg::racl_policy_vec_t racl_policies_i,
+  output logic                           racl_error_o,
+  output top_racl_pkg::racl_error_log_t  racl_error_log_o,
```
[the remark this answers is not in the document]
vogelpi: Sounds good, thanks @Razer6 !

hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
davidschrammel: For reference: This is done in #26004

PR #25753, opened 2024-12-29
[hw,spi_host,rtl] Make NumCS an exposed  parameter

Razer6:
This moves the NumCS param from the reg_top_pkg to an instance level parameter. This allows instances to overwrite the parameter.

### DIFF
--- a/hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
+++ b/hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
@@ -15,9 +15,12 @@ package spi_host_env_pkg;
   import dv_base_reg_pkg::*;
   import spi_host_reg_pkg::*;
   import spi_host_ral_pkg::*;
+  import spi_host_env_cfg_pkg::*;
+
+  // Re-export SPI_HOST_NUM_CS from spi_host_env_cfg_pkg to make it available in DV
+  export spi_host_env_cfg_pkg::SPI_HOST_NUM_CS;
 
   // parameters
-  parameter uint SPI_HOST_NUM_CS         = spi_host_reg_pkg::NumCS;
   parameter uint SPI_HOST_TX_DEPTH       = spi_host_reg_pkg::TxDepth;
   parameter uint SPI_HOST_RX_DEPTH       = spi_host_reg_pkg::RxDepth;
   parameter uint SPI_HOST_CMD_DEPTH      = spi_host_reg_pkg::CmdDepth;
--- a/hw/ip/spi_host/dv/tb.sv
+++ b/hw/ip/spi_host/dv/tb.sv
@@ -28,8 +28,8 @@ module tb;
   lc_tx_t                       scanmode_i;
   logic                         cio_sck_o;
   logic                         cio_sck_en_o;
-  logic [NumCS-1:0]             cio_csb_o;
-  logic [NumCS-1:0]             cio_csb_en_o;
+  logic [SPI_HOST_NUM_CS-1:0]   cio_csb_o;
+  logic [SPI_HOST_NUM_CS-1:0]   cio_csb_en_o;
   logic [3:0]                   cio_sd_o;
   logic [3:0]                   cio_sd_en_o;
   logic [3:0]                   cio_sd_i;
@@ -49,7 +49,9 @@ module tb;
   `DV_ALERT_IF_CONNECT()
 
   // dut
-  spi_host dut (
+  spi_host # (
+    .NumCS(SPI_HOST_NUM_CS)
+  ) dut (
     .clk_i                (clk),
     .rst_ni               (rst_n),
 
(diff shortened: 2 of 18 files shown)


a-will (approved):

Razer6:
Odd that `NumCS` is used in `cmmand_t`. Moving out `csid` from that struct?

a-will:
> Odd that `NumCS` is used in `cmmand_t`. Moving out `csid` from that struct?

As long as `csid` still actually goes into the command FIFO, that seems fine to me. Would a parameterized interface work better than a struct, though? DV might need to be fixed up, too.

hw/ip/spi_host/rtl/spi_host.sv
```diff
@@ -216,7 +218,7 @@ module spi_host
   assign error_cmd_inval  = command_valid & ~command_busy &
                             (test_speed_inval | test_dir_inval);
 
-  assign command.csid = (test_csid_inval) ? '0 : reg2hw.csid.q[CSW-1:0];
+  assign command_csid = (test_csid_inval) ? '0 : reg2hw.csid.q[NumCS-1:0];
```
a-will, line 221: This should still use CSW, right? The CSID is an index, not a bit map. :)

Razer6: Indeed. Thanks for flagging!

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
```diff
@@ -15,9 +15,10 @@ package spi_host_env_pkg;
   import dv_base_reg_pkg::*;
   import spi_host_reg_pkg::*;
   import spi_host_ral_pkg::*;
+  import spi_host_env_cfg_pkg::*;
 
   // parameters
-  parameter uint SPI_HOST_NUM_CS         = spi_host_reg_pkg::NumCS;
+  parameter uint SPI_HOST_NUM_CS         = 1;
```
andreaskurth, line 21:
```suggestion
```

hw/ip/spi_host/dv/tb.sv
```diff
@@ -28,8 +28,8 @@ module tb;
   lc_tx_t                       scanmode_i;
   logic                         cio_sck_o;
   logic                         cio_sck_en_o;
-  logic [NumCS-1:0]             cio_csb_o;
-  logic [NumCS-1:0]             cio_csb_en_o;
+  logic [SPI_HOST_NUM_CS-1:0]   cio_csb_o;
+  logic [SPI_HOST_NUM_CS-1:0]   cio_csb_en_o;
```
andreaskurth, line 32: The simulator complains that `SPI_HOST_NUM_CS` isn't defined here. I'm not sure why, though, because this module imports `spi_host_env_pkg::*`, which in turn imports `spi_host_env_cfg_pkg::*`, which defines `SPI_HOST_NUM_CS`. Maybe try importing it directly?

a-will: `spi_host_env_pkg` would have to explicitly re-export the symbol, but it doesn't. `spi_host_env_pkg` would need a line like...

```
export spi_host_env_cfg_pkg::SPI_HOST_NUM_CS;
```

andreaskurth: Thanks for the explanation, Alex :+1:

Razer6: Thanks Alex. Added that!

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
```diff
@@ -15,9 +15,12 @@ package spi_host_env_pkg;
   import dv_base_reg_pkg::*;
   import spi_host_reg_pkg::*;
   import spi_host_ral_pkg::*;
+  import spi_host_env_cfg_pkg::*;
+
+  // Re-export SPI_HOST_NUM_CS from spi_host_env_cfg_pkg to make it available in DV
+  export spi_host_env_cfg_pkg::SPI_HOST_NUM_CS;
```
vogelpi, line 21: I've not come across this before but I think it should be fine. WDYT @rswarbrick ?

vogelpi (approved):
The RTL changes look all good to me, thanks @Razer6 !

hw/ip/spi_host/dv/env/spi_host_env_pkg.sv
Razer6: @vogelpi See the comment from @a-will [above](https://github.com/lowRISC/opentitan/pull/25753#discussion_r1902096281) for an explanation.

rswarbrick: Thanks for the explanation: I don't think I knew this before! :-)

rswarbrick:
CHANGE AUTHORIZED: hw/ip/spi_host/rtl/spi_host_cmd_pkg.sv
CHANGE AUTHORIZED: hw/ip/spi_host/rtl/spi_host_command_queue.sv
CHANGE AUTHORIZED: hw/ip/spi_host/rtl/spi_host_core.sv
CHANGE AUTHORIZED: hw/ip/spi_host/rtl/spi_host_fsm.sv
CHANGE AUTHORIZED: hw/ip/spi_host/rtl/spi_host_reg_pkg.sv

This won't have an effect on earlgrey, so should not have a risk.

PR #25754, opened 2024-12-30
[hw,darjeeling,rtl] Separate the RAM cfg ports

Razer6:
This allows to route differnt config parameters to different RAM instances.


andreaskurth (approved):

andreaskurth:
A CW310 FPGA CI check is failing, but this PR only changes Darjeeling, which we don't test in that check.

PR #25759, opened 2025-01-02
[test,do-not-merge] CI FPGA test Bazel execution log testing

AlexJones0:
Do not merge this PR, being used to extract Bazel execution logs from CI FPGA runs to determine potential caching issues.


rules/otp.bzl
```diff
@@ -297,6 +299,10 @@ otp_image = rule(
             default = False,
             doc = "Display progress messages from image-generation tool.",
         ),
+	"stamp": attr.bool(
```
pamaury, line 302: I would suggest to use the "standard" stamping attribute that bazel provides, see https://github.com/lowRISC/opentitan/blob/16076d679085f2bafc6bd5acee541549292554ce/rules/autogen.bzl#L140
This way this be automatically stamped when running with `bazel --stamp`

PR #25764, opened 2025-01-02
Fix questa hjson file with needed options

mostafaabdallah12:
We've been working for the last two months to get OpenTitan to work with Questa, and we have succeeded! 

We changed some setup files to add questa options and setups.

This is one of the setup files: hw/dv/tools/dvsim/questa.hjson

Once it's accepted in OpenTitan master branch we will push the remaining setup files.

### DIFF
--- a/hw/dv/tools/dvsim/questa.hjson
+++ b/hw/dv/tools/dvsim/questa.hjson
@@ -14,7 +14,7 @@
                 "-uvm -uvmhome {QUESTA_HOME}/verilog_src/uvm-1.2",
                 "-mfcu",
                 "-sv",
-								"-svext=hierbinsof",
+                "-svext=hierbinsof",
                 // Questa throws error on some LRM violations & restrictions. We demoted these errors since they will not impact the flow:
                   //2744: It shall be illegal to have an import statement directory within a class scope.
                   //      [DOC: IEEE Std 1800-2009 Verilog LRM - Section 26.3 "Referencing data in packages"]
--- a/hw/dv/tools/dvsim/questa.hjson
+++ b/hw/dv/tools/dvsim/questa.hjson
@@ -14,7 +14,21 @@
                 "-uvm -uvmhome {QUESTA_HOME}/verilog_src/uvm-1.2",
                 "-mfcu",
                 "-sv",
-				"-svext=hierbinsof",
+								"-svext=hierbinsof",
+                // Questa throws error on some LRM violations & restrictions. We demoted these errors since they will not impact the flow:
+                  //2744: It shall be illegal to have an import statement directory within a class scope.
+                  //      [DOC: IEEE Std 1800-2009 Verilog LRM - Section 26.3 "Referencing data in packages"]
+                  //2912: The specified port cannot be connected by name because it does not exist.
+                  //7070: An expression of string type can be assigned directly to a variable of string type.
+                  //      Values of integral type can be assigned to a string variable, but require a cast.
+                  //      [DOC: IEEE Std 1800-2012 Verilog LRM - 6.16 String Data Type]
+                  //7034: The strict rule for unpacked array assignment/comparison requires that the packed elements must match exactly as do the keys for associative arrays.
+                  //7033: In SystemVerilog, a variable assigned inside an always_comb, always_latch, or always_ff may not be assigned by any other process.
+                  //2244: An explicit static keyword shall be required when an initialization value is specified as part of a static variable's declaration to indicate the user's intent of executing that initialization only once at the beginning of simulation.
+                  //      The static keyword shall be optional where it would not be legal to declare the variables as automatic.
+                  //      [DOC: IEEE Std 1800-2009 Verilog LRM - 6.21 Scope and Lifetime] Variable initialized in its declaration in a static task, function or procedural block requires an explicit static keyword.
+                  //7041: The strict rule for string type assignment requires a cast if a string type is assigned to a packed type or a packed type is assigned to a string type.
+                  //      Note that a string literal is not the same as a string type.
                 "-suppress 2744,2912,7070,7034,7033,2244,7041",
                 "-f {sv_flist}",
                 // List multiple tops for the simulation. Prepend each top level with `-top`.
--- a/hw/dv/tools/dvsim/questa.hjson
+++ b/hw/dv/tools/dvsim/questa.hjson
@@ -2,26 +2,34 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 {
-  build_cmd:  "{job_prefix} {QUESTA_HOME}/questasim/linux_x86_64/qrun -optimize"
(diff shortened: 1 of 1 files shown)


hw/dv/tools/dvsim/questa.hjson
```diff
@@ -2,26 +2,34 @@
...
   build_opts: [ "-timescale 1ns/1ps",
                 "-outdir {build_dir}/qrun.out",
-                "-uvm -uvmhome {QUESTA_HOME}/questasim/verilog_src/uvm-1.2",
+				"+incdir+{QUESTA_HOME}/verilog_src/uvm-1.2/src",
+				"+incdir+{QUESTA_HOME}/verilog_src/uvm-1.2/src/mentor",
+                "{QUESTA_HOME}/verilog_src/uvm-1.2/src/uvm_pkg.sv",
+                "-uvm -uvmhome {QUESTA_HOME}/verilog_src/uvm-1.2",
                 "-mfcu",
+                "-sv",
+				"-svext=hierbinsof",
```
rswarbrick, line 17: Nit: I think the indentation is off here?

hw/dv/tools/dvsim/questa.hjson
```diff
@@ -2,26 +2,34 @@
...
+                "-suppress 2744,2912,7070,7034,7033,2244,7041",
```
rswarbrick, line 18: I'm willing to believe this is needed, but could it have a comment above it that says roughly we're suppressing? Before this change, `run_opts` had a comment like that for the 8323 message that was being suppressed.

I think we will need equivalent notes for all the other warnings that are being suppressed.

mostafaabdallah12:
@rswarbrick  I fixed the indentation issue. and added notes for the suppressed errors.

hw/dv/tools/dvsim/questa.hjson
```diff
@@ -2,26 +2,48 @@
...
 
   run_opts:   [ "-outdir {build_dir}/qrun.out",
                 "-sv_seed {seed}",
+                "-c",
+                "-sv_lib {QUESTA_HOME}/uvm-1.2/linux_x86_64/uvm_dpi",
                 // dv_macros.svh has a macro printing null using %0d
                 // format specifier, Questa throws an error on this
                 // we demote this error in Questa
-                "-suppress vsim-8323",
+                "-suppress 8323,7034,7033,7067,8111,16133,16132,7041",
```
rswarbrick, line 46: I'm really sorry, but would you mind adding notes above this line too?

hw/dv/tools/dvsim/questa.hjson
```diff
@@ -2,26 +2,48 @@
...
+				"+incdir+{QUESTA_HOME}/verilog_src/uvm-1.2/src",
+				"+incdir+{QUESTA_HOME}/verilog_src/uvm-1.2/src/mentor",
+                "{QUESTA_HOME}/verilog_src/uvm-1.2/src/uvm_pkg.sv",
+                "-uvm -uvmhome {QUESTA_HOME}/verilog_src/uvm-1.2",
                 "-mfcu",
+                "-sv",
+                "-svext=hierbinsof",
+                // Questa throws error on some LRM violations & restrictions. We demoted these errors since they will not impact the flow:
+                  //2744: It shall be illegal to have an import statement directory within a class scope.
+                  //      [DOC: IEEE Std 1800-2009 Verilog LRM - Section 26.3 "Referencing data in packages"]
```
rswarbrick, line 20: Eek! The waiver looks perfectly sensible, but is it easy for you to generate a list of examples that currently fail? I don't have a Questa licence, but I'd be very happy to tidy this up...

hw/dv/tools/dvsim/questa.hjson
```diff
@@ -2,26 +2,48 @@
...
+                  //2912: The specified port cannot be connected by name because it does not exist.
```
rswarbrick, line 21: This is very surprising to me! That would presumably be a bug in the code in question? Can you send examples that are currently failing this?

hw/dv/tools/dvsim/questa.hjson
```diff
@@ -2,26 +2,48 @@
...
+                  //7070: An expression of string type can be assigned directly to a variable of string type.
```
rswarbrick, line 22: Again, this looks rather surprising to me! I assume that `bit [7:0] foo = "12" won't set `foo` to `8'hc` :-) Can you send a list of examples that currently fail this?

rswarbrick (commented):
I started adding notes to this but realised that I was writing the same thing many times over :-)

Would you mind giving a list of examples for each waived rule? We might just be able to fix them, or otherwise should probably file issues to track fixing them properly.

mostafaabdallah12:
@rswarbrick  Can we proceed with the pull request and add the change and I will file issues for all suppressible errors we added for with more details

rswarbrick:
Sorry for not replying more quickly. I'd really appreciate at least a list of what the magic IDs mean. I don't have a Questa licence, so I don't know what the numbers mean at all. The output could be an issue (generated by me) to track the fact that this needs sorting out at our end.

mostafaabdallah12:
@rswarbrick  I thought we had added notes for all IDs. I updated the last commit to have notes for all the IDs. 
It's worth mentioning that it's not a must that it's an issue in code. Maybe Questa couldn't retrieve this info and throws an error, which makes sense to have these IDs as suppressible IDs. I will review from our side and make sure we file issues on github in case there's a real issue not questa issue.

hw/dv/tools/dvsim/questa.hjson
rswarbrick: I'm working (very slowly) through this list of waivers. I think the first one will be something we can remove once #26002 is merged. (If not, thank you for testing! Please tell me what I've missed!)

hw/dv/tools/dvsim/questa.hjson
rswarbrick: Again, please could you provide a list of examples where this doesn't work? (Or alternatively a pattern that will find them with grep?)

hw/dv/tools/dvsim/questa.hjson
```diff
@@ -2,26 +2,56 @@
...
   run_opts:   [ "-outdir {build_dir}/qrun.out",
                 "-sv_seed {seed}",
-                // dv_macros.svh has a macro printing null using %0d
-                // format specifier, Questa throws an error on this
-                // we demote this error in Questa
-                "-suppress vsim-8323",
+                "-c",
+                "-sv_lib {QUESTA_HOME}/uvm-1.2/linux_x86_64/uvm_dpi",
+                // Questa throws error on some LRM violations & restrictions. We demoted these errors since they will not impact the flow:
+                  //8323: dv_macros.svh has a macro printing null using %0d format specifier, Questa throws an error on this.
```
rswarbrick, line 44: Oof, that sounds like a genuine bug in some DV code ("passing a null pointer to something that wants an integer"). Can you give a list of examples of this? I'd love to just fix them!

rswarbrick (commented):
As a very general review, rather than adding waivers I'd *much* prefer to just fix the code.

If you can list examples of where we generate any of the errors, I'd be very happy to sort them out.

PR #25767, opened 2025-01-03
updated exclusions for GPIO

marcelocarvalhoLowRisc:
New updates for the exclusion file for GPIO.


rswarbrick:
I'll review this properly in a second, but it's failing the lint job because of a missing signoff line in the commit message.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,61 @@
...
-// Date: Wed Oct 30 17:22:12 2019
+// Date: Fri Dec 20 15:24:02 2024
 // ExclMode: default
 //==================================================
-CHECKSUM: "3109872245 3727953184"
-INSTANCE: tb.dut.u_reg.u_data_in
-ANNOTATION: "[UNR]"
-Block 1 "0" "assign wr_en = de;"
-CHECKSUM: "1352181506 3803838024"
+CHECKSUM: "1706182284"
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: Yes, it is generate using the verdi coverage. Strange, I don't understand what you mean by having effect.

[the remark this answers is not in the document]
rswarbrick: What I mean is that I think the file format uses `INSTANCE` to select some instance in the design but only *excludes* a coverpoint with a command like `Block`. This code seems to have lots of `INSTANCE` lines with associated checksums and (in some cases) annotations, but most don't actually do anything.

But it's possible that I'm missing something about how VCS works. Is there something clever going on?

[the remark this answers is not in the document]
marcelocarvalhoLowRisc: I have no idea why the VCS does that. Usually in my previous experience, I never care about the content of the .el files, usually I only look at it using the tool.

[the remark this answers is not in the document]
rswarbrick: Huh, how odd. But I still think something strange is going on. For example, this PR currently adds an annotation that starts "Becasuse of the data_in is a read-only register, t..." (line 18). What effect does that annotation have? (Maybe try to find the text in the UI? It's possible there's something clever in the file format that I've missed)

marcelocarvalhoLowRisc (commented):
File updated, I removed the strap related signals that were excluded before, now with the PR about strap verification test, those signals are suppose to toggle.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,48 @@
-// Copyright lowRISC contributors (OpenTitan project).
-// Licensed under the Apache License, Version 2.0, see LICENSE for details.
-// SPDX-License-Identifier: Apache-2.0
```
rswarbrick, line 3: Dropping this header is causing CI to fail.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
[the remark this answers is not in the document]
rswarbrick: I'm a little confused: I don't know what these annotations apply to:
image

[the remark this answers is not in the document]
marcelocarvalhoLowRisc: Fixed, let me know what do you think.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,55 @@
...
 //==================================================
-CHECKSUM: "3109872245 3727953184"
-INSTANCE: tb.dut.u_reg.u_data_in
-ANNOTATION: "[UNR]"
-Block 1 "0" "assign wr_en = de;"
-CHECKSUM: "1352181506 3803838024"
+CHECKSUM: "1706182284 132761700"
+INSTANCE: tb.dut.u_reg.u_reg_if.u_rsp_intg_gen
+ANNOTATION: "The parameter EnableRspIntgGen is zero, so the condition will never hit."
+Block 1 "461445014" "assign rsp_intg = tl_i.d_user.rsp_intg;"
```
rswarbrick, line 11: I'm not sure if this is correct. The code looks like this:
```systemverilog
  logic [D2HRspIntgWidth-1:0] rsp_intg;
  if (EnableRspIntgGen) begin : gen_rsp_intg
    tl_d2h_rsp_intg_t rsp;
    logic [D2HRspMaxWidth-1:0] unused_payload;

    assign rsp = extract_d2h_rsp_intg(tl_i);

    prim_secded_inv_64_57_enc u_rsp_gen (
      .data_i(D2HRspMaxWidth'(rsp)),
      .data_o({rsp_intg, unused_payload})
    );
  end else begin : gen_passthrough_rsp_intg
    assign rsp_intg = tl_i.d_user.rsp_intg;
  end
```

I think that the parameter means we end up in the second block. The reason that you don't see line coverage for this continuous assignment is that `tl_i.d_user.rsp_intg` never changes.

That makes sense because it's actually wired to zero in `gpio_reg_top`. But it's a bit ugly. I think this is probably a candidate for a (trivial) RTL change. The assignment could just be
```systemverilog
  if (EnableRspIntgGen) begin : gen_rsp_intg
    tl_d2h_rsp_intg_t rsp;
    logic [D2HRspMaxWidth-1:0] unused_payload;

    assign rsp = extract_d2h_rsp_intg(tl_i);

    prim_secded_inv_64_57_enc u_rsp_gen (
      .data_i(D2HRspMaxWidth'(rsp)),
      .data_o({rsp_intg, unused_payload})
    );
  end else begin : gen_passthrough_rsp_intg
    assign rsp_intg = 0;
  end
```
and we should add an assertion at the bottom:
```systemverilog
  `ASSERT(ZeroRspIntg_A, |tl_i.d_user.rsp_intg == 0)
```

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,55 @@
...
+ANNOTATION: "The parameter EnableRspIntgGen is zero, so the condition will never hit."
+Block 2 "2643129081" "assign data_intg = tl_i.d_user.data_intg;"
```
rswarbrick, line 13: The annotation points at the wrong parameter, which is a little dubious...

But I think the same thing as I said about the previous item (and this can be in a parallel PR). Better to wire it to zero and add an assertion.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,55 @@
...
+CHECKSUM: "2099741489 2073313596"
+INSTANCE: tb.dut.u_reg.u_data_in.wr_en_data_arb
+ANNOTATION: "Becasuse of the data_in is a read-only register, the signals wr_en and wr_data are not applicable. "
```
rswarbrick, line 16: This annotation is wrong. The missing line coverage means that `prim_subreg_arb.sv`, line 43 never excutes. Again, this is a continuous assignment, so it really means that `de` is constant. Here, `hw2reg.data_in.de` is wired to `1` (`gpio.sv`, line 112).

And that's fine: the hardware updates `data_in` on every cycle.

But the annotation needs fixing.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,55 @@
...
+CHECKSUM: "295805079 2786204945"
 INSTANCE: tb.dut.u_reg.u_masked_out_lower_mask
-ANNOTATION: "[UNR]"
-Block 1 "0" "assign qs = d;"
-CHECKSUM: "1352181506 3803838024"
-INSTANCE: tb.dut.u_reg.u_masked_out_upper_mask
-ANNOTATION: "[UNR]"
-Block 1 "0" "assign qs = d;"
-CHECKSUM: "1352181506 3803838024"
+ANNOTATION: "The masked_oe_lower_mask is only accesible by software. The signal related with d () is tied to zero. hw2reg.masked_out_lower.mask.d"
```
rswarbrick, line 20: The name is wrong at the start of this annotation: you want `masked_out_lower_mask`, I think.

Otherwise, I think I agree. Maybe this should say something like "The hardware always signals d=0 (through hw2reg.masked_out_lower.mask.d) so this continuous assignment will not execute."

And of course this applies to the few items too.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,55 @@
...
-ANNOTATION: "[UNR]"
-Block 1 "0" "assign qs = d;"
-CHECKSUM: "3109872245 1595801502"
+ANNOTATION: "The masked_oe_lower_mask is only accesible by software. The signal related with d () is tied to zero. hw2reg.masked_out_lower.mask.d"
+Block 2 "1004401251" "assign qs = d;"
+ANNOTATION: "The masked_oe_lower_mask is only accesible by software. The signal related with d () is tied to zero. hw2reg.masked_out_lower.mask.d"
+Block 1 "531057943" "assign ds = d;"
+CHECKSUM: "4255502330 2331745845"
+INSTANCE: tb.dut.u_reg.u_data_in
+ANNOTATION: "Becasuse of the data_in is a read-only register, the signals wr_en and wr_data are not applicable. "
```
rswarbrick, line 44: I think this needs tweaking a bit. Isn't the point that we don't see `wr_en=0`? That is because the `wr_en` signal is just `de` (as computed by a `prim_subreg_arb` with this register mode). But `de` is wired to `1` in `gpio.sv`.

The same applies to the following items, which need explaining a bit more carefully.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
marcelocarvalhoLowRisc: Thanks for the suggestion, I will update the annotation text.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
marcelocarvalhoLowRisc: Very true, I misunderstood the behavior of this prim_subreg_arb register instance. I will update the annotation and replicate to the other ones.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
rswarbrick: This comment still stands (and CI still fails)

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,59 @@
...
 // ExclMode: default
 //==================================================
-CHECKSUM: "3109872245 3727953184"
-INSTANCE: tb.dut.u_reg.u_data_in
-ANNOTATION: "[UNR]"
-Block 1 "0" "assign wr_en = de;"
-CHECKSUM: "1352181506 3803838024"
+CHECKSUM: "1706182284 132761700"
+INSTANCE: tb.dut.u_reg.u_reg_if.u_rsp_intg_gen
+ANNOTATION: "The signal intg_error_o is wired to zero in the gpio_reg_top.sv (line 104), which means no response from the u_rsp_intg_gen instance."
```
rswarbrick, line 10: The annotation here is now
```
The signal intg_error_o is wired to zero in the gpio_reg_top.sv (line 104), which means no response from the u_rsp_intg_gen instance.
```

That's not really correct: I think you might be looking at line 105 of
`gpio_reg_top.sv`, which sets a parameter with a different name to
zero. That isn't really related.

When diagnosing this myself:
- I noticed that the `tl_i` signal for this instance of
  `u_rsp_intg_gen` comes from `tl_o_pre` in `tlul_adapter_reg` (line
  171 of `tlul_adapter_reg.sv`)
- That structure has a zero for `d_user` (line 162 of `tlul_adapter_reg.sv`)
- Therefore, the right-hand side of the assignment is constant and we
  will never see it "execute".

hw/ip/gpio/dv/cov/gpio_cov_excl.el
rswarbrick: Oh, nuts. Sorry: I filed a separate comment because I didn't realise we'd already looked at this code. I think the explanation is better [here](https://github.com/lowRISC/opentitan/pull/25767#discussion_r1963830459).

hw/ip/gpio/dv/cov/gpio_cov_excl.el
rswarbrick: I think there's a proper discussion about this [here](https://github.com/lowRISC/opentitan/pull/25767#discussion_r1963830459). Let's sort things out in one place, so I'll mark this item resolved.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
rswarbrick: This looks more sensible, thanks. I'd probably suggest tweaking it slightly to avoid line numbers. Otherwise unrelated changes that e.g. add a line above will make this annotation false.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
marcelocarvalhoLowRisc: Aha! now I got it where this comes from. I will fix the annotation, sorry I take long to understand.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
marcelocarvalhoLowRisc: Annotation fixed =)

hw/ip/gpio/dv/cov/gpio_cov_excl.el
rswarbrick: I've just opened #26390 which should hopefully solve this on the RTL side later.

rswarbrick:
I think there is just one item remaining! [Here](https://github.com/lowRISC/opentitan/pull/25767#discussion_r1963849696)

hw/ip/gpio/dv/cov/gpio_cov_excl.el
marcelocarvalhoLowRisc: Please take a look again =) and let me know what you think

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,63 @@
...
+ANNOTATION: "The hardware always signals d=0 (through hw2reg.masked_out_lower.mask.d) so this continuous assignment will not execute."
+Block 2 "1004401251" "assign qs = d;"
+ANNOTATION: "The hardware always signals d=0 (through hw2reg.masked_out_lower.mask.d) so this continuous assignment will not execute."
+Block 1 "531057943" "assign ds = d;"
+CHECKSUM: "295805079 2786204945"
 INSTANCE: tb.dut.u_reg.u_masked_oe_lower_mask
-ANNOTATION: "[UNR]"
-Block 1 "0" "assign qs = d;"
-CHECKSUM: "1352181506 3803838024"
+ANNOTATION: "The hardware always signals d=0 (through hw2reg.masked_out_lower.mask.d). So this is continuous assignment will not execute."
```
rswarbrick, line 31: Being a little more specific (than I was [here](https://github.com/lowRISC/opentitan/pull/25767#discussion_r1969544036)), I think this annotation talks about `masked_out_lower` but it applies to `masked_ao_lower_mask`.

I think that's true for the other `oe` lines below as well.

hw/ip/gpio/dv/cov/gpio_cov_excl.el
```diff
@@ -1,34 +1,59 @@
-// Copyright lowRISC contributors (OpenTitan project).
-// Licensed under the Apache License, Version 2.0, see LICENSE for details.
-// SPDX-License-Identifier: Apache-2.0
```
rswarbrick, line 3: Dropping these lines breaks CI. If you're going to use a tool to generate the file, you'll need to add them back manually.

rswarbrick (changes requested):

rswarbrick (approved):
This looks good. Thanks!

PR #25768, opened 2025-01-03
[prim_generic] Depend on the tech ram_pkg within the tech lib

Razer6:
For generic usages, i.e.g, when using a ram_pkg in spi_device or in i2c, we rely on the fusesoc resolving skills. Once the virtual cors PR is merged, this gets explicitly set in the top-level core file.

### DIFF
--- a/hw/ip/prim_xilinx/prim_xilinx_ram_1p.core
+++ b/hw/ip/prim_xilinx/prim_xilinx_ram_1p.core
@@ -9,7 +9,6 @@ filesets:
   files_rtl:
     depend:
       - lowrisc:prim:assert
-      - lowrisc:prim:ram_1p_pkg
       # Note that prim_xilinx_pkg is a virtual core that the top should provide.
       # It maps parameters to instructions for how to split memories into
       # logical groups of bits. See prim_xilinx_default_pkg for an example.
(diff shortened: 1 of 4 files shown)


andreaskurth (approved):

hw/ip/prim_xilinx/prim_xilinx_ram_1p.core
```diff
@@ -9,7 +9,6 @@ filesets:
   files_rtl:
     depend:
       - lowrisc:prim:assert
-      - lowrisc:prim:ram_1p_pkg
```
vogelpi, line 12: Note for other reviewers: `lowrisc:prim_generic:ram_1p` is already included below.

vogelpi (approved):
LGTM.

hw/ip/prim_xilinx/prim_xilinx_ram_1p.core
a-will: This dependency also wouldn't work properly anymore anyway. Our fusesoc fork cannot run generators recursively, so the only way the prim library-specific cores of `lowrisc:prim:ram_1p_pkg` get included is if some non-generated core includes it.

PR #25769, opened 2025-01-03
[hw,entropy_src,rtl] Move RNG_BUS_WIDTH from entropy_src to top_pkg

Razer6:
This allows Tops to configure that parameter for their needs. Depending on the used source of of randomness, the bus width to that IP might be different.


Razer6:
Dropping that for now since it is fixed out of tree.

PR #25770, opened 2025-01-03
[rom_ctrl, dv] Improvements for V3

KinzaQamar:
Introduced extern method for declaring members of a class for better readibility


rswarbrick (approved):
Oof: this is quite a major rejig! I think it looks sensible though, thanks :-D

PR #25771, opened 2025-01-03
[sram_ctrl,dv] Cov. exclusion for tlul_sram_byte

nasahlpa:
This commit excludes a branch that cannot be reached in the current configuration and SRAM setup. A detailed explanation can be found in the code.

Screenshot from 2025-01-03 14-36-28

### DIFF
--- a/hw/ip/sram_ctrl/dv/cov/sram_ctrl_manual_cov_excl.el
+++ b/hw/ip/sram_ctrl/dv/cov/sram_ctrl_manual_cov_excl.el
@@ -1,0 +1,32 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// The following exclusions were manually added.
+
+// The `pending_txn_cnt` never can have a other value than 1 in the `StByteWrReadBackInit` state.
+// The FSM follows the following path: `StWaitRd` -> `StWriteCmd` -> `StByteWrReadBackInit`.
+// Observations:
+// - In all of these states, we are stalling the host with `stall_host = 1'b1`.
+// - When switching from `StWaitRd` -> `StWriteCmd`, `pending_txn_cnt == 1`.
+// - `pending_txn_cnt` is the depth of the `u_sync_fifo_a_size` FIFO
+//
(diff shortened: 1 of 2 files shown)


hw/ip/sram_ctrl/dv/cov/sram_ctrl_manual_cov_excl.el
```diff
@@ -0,0 +1,32 @@
...
+//
+// Popping the FIFO in the `StWaitRd` state is not possible as we set `rd_wait = 1'b1`, which
+// puts `tl_o.d_valid = 1'b0` and therefore also `d_ack = 1'b0`.
+// Popping the FIFO in the `StWriteCmd` state is not possible as we trigger a write in this
+// state and immediately get the `sram_a_ack` (the SRAM always can accept a read/write operation)
+// and the response `tl_sram_i.d_valid` cannot be in the same cycle. This again puts
+// `tl_o.d_valid = 1'b0` and therefore also `d_ack = 1'b0`.
+//
+// Hence, when we are entering the `StByteWrReadBackInit` state, `pending_txn_cnt == 1`. In this
+// state, we immediately switch to the `StByteWrReadBack` or `StByteWrReadBackDWait` state.
```
rswarbrick, line 26: This reasoning took me a while to understand, but I'm convinced, I think! I don't think this is specific to `sram_ctrl` though. Instead of excluding something, shouldn't we be: dropping the condition from `tlul_sram_byte.sv`, line 396 and replacing it with an assertion?

nasahlpa: Thanks for looking into this!

I also was thinking of removing this condition but I am a bit torn back and forth. Three points against removing it:

- If we change the underlying SRAM macro, this condition could actually make sense. Though the added assertion would prevent that things go wrong.
- `tlul_adapter_sram` is used in many other places. I only had a look in the SRAM controller usage. But again, the assertion would fire if there is a different behavior.
- SRAM controller is quite a functional critical IP - so I thought we try avoiding RTL changes?

But I have no strong feeling about this. I am happy to do the RTL change :-)

rswarbrick: I think I'm generally in favour of making the change. We're quite early in the "tapeout cycle", so we'd probably see something explode if we've forgotten something.

PR #25772, opened 2025-01-03
[bazel] Migrate manufacturer hooks to bzlmod

jwnrt:
The existing repository rules generate Bazel repositories at a path pointed to by an environment variable, falling back to a `dummy` directory if not given.

These rules can be merged into a single module extension. There is no change in functionality.

I have also migrated the nonhermetic repo and local toolchain registration since these are simple.


pamaury:
This is much cleaner than the workspace implementation because it makes all the environment variables appear in a single place. That being said, with the move to blzmod, maybe we should recommend that our partners override repositories instead of using env vars? In particular, the [`--override_repository`](https://bazel.build/reference/command-line-reference#flag--override_repository) CLI option could be very useful, especially if they are gathered in the .bazelrc file, or maybe in a `try-import`ed file?

pamaury (approved):

a-will:
> This is much cleaner than the workspace implementation because it makes all the environment variables appear in a single place. That being said, with the move to blzmod, maybe we should recommend that our partners override repositories instead of using env vars? In particular, the [`--override_repository`](https://bazel.build/reference/command-line-reference#flag--override_repository) CLI option could be very useful, especially if they are gathered in the .bazelrc file, or maybe in a `try-import`ed file?

FWIW, I'd agree with moving to using `--override_repository` instead of using environment variables. The `try-import` file option could continue being the `.bazelrc-site` we already support.

In terms of timeline, if we agree to go that route, it could come in a later reorganization that coincides with the introduction of ROM hooks, though.

jwnrt:
Rebased for conflicts.

I agree, we should change to using overrides in the future but will stay with the current env vars for now.

PR #19157, opened 2023-07-10
[alert_handler] Convert theory of operation to template

a-will:
The theory of operation document referenced a particular ~instance's~ generated core's parameters in a table. Change the table header to not explicitly refer to Earl Grey and populate some of the values from template parameters.

Note: The conversion to Mako template caused lines beginning with `#` to become comments, so I have changed the syntax to use the HTML variant.

PR #25773, opened 2025-01-04
[ipgen,topgen] Uniquify VLNVs and reorganize core hierarchies

a-will:
It'd be best to look at each commit individually, but the summary is that the changes move towards the following:

- earlgrey-specific VLNVs that looked very general have been converted to show that they are top-specific and unique.
   - The general VLNVs have become virtual VLNVs that each top has cores serving as providers.
- Non-templated IPs that depend on top-specific packages use the virtual VLNVs.
- Templated IPs created by topgen directly depend on the top-specific _real_ VLNVs.
- Various SV files providing constants have been moved out of cores that also have SV modules and collected into cores with similar package-providing files.
- Overall, every top-level core and ipgen core used with `fusesoc run` is should now (hopefully) provide a deterministic set of dependencies with no cycles in the graph.

Some minor changes have been made to begin making alert_handler optional, since English Breakfast doesn't use it. pwrmgr's use of `alert_pkg` has been changed to simple parameters, which currently only use the defaults (since alert_handler's original values are not actually parameterizable right now anyway). There is much more work that could be done across other IPs, but I'm not equipped to handle feature enablement / disablement for them. :smile: 

The use of globally-accessible constants is pervasive in our code base, and it probably ought to have a more critical eye checking for abuse. Also, the naming conventions here are likely to grate, considering we throw all the weight into the Name portion of the VLNV and barely use the Library. Suggestions for changes are expected, haha.

This takes some bits from https://github.com/lowRISC/opentitan/pull/25703 and https://github.com/lowRISC/opentitan/pull/25747, and hopefully it provides a more reviewable history and breakdown.


a-will:
Note that virtual VLNVs can be a trap, and we should be careful about using them in scenarios where we might want a second instance of something with different parameters. This PR reduces the number of these virtual VLNVs that are in our code base.

a-will:
Ah, probably need to remove bbca444, since it's more for englishbreakfast moving to the ordinary topgen flow, but I've removed those commits for this PR's sake.

hw/top_darjeeling/top_darjeeling.core
```diff
@@ -90,7 +90,7 @@ filesets:
       - lowrisc:lint:common
       - lowrisc:lint:comportable
     files:
-      - lint/top_darjeeling.vbl
```
Razer6, line 93: That line should be reverted. CI greps for `*.vbl` in the repo to build a global waiver:
 
```sh
echo "Concatenating Verible waivers"
find . -type f -name '*.vbl' -exec cat {} \; >> verible_waiver
```

a-will: Ah, I must have made a mistake during a rebase. That line change was definitely unintentional.

hw/top_darjeeling/FUSESOC_IGNORE
```diff
@@ -1 +0,0 @@
-
```
Razer6, line 1: ❤️

Razer6 (approved):
This really looks great. Thank you Alex! Just some minor questions.

util/topgen-fusesoc.py
```diff
@@ -1,164 +0,0 @@
-#!/usr/bin/env python3
```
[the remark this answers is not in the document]
a-will: Yes, English Breakfast's flow would call this script to generate a core file, and "fileset_topgen" would be the identifier for this flow.

I'll pull this commit out and move it to a PR focused on English Breakfast, though.

hw/top_darjeeling/top_darjeeling.core
a-will: Fixed that up.

hw/top_earlgrey/ip/ast/top_earlgrey_ast_top.core
```diff
@@ -0,0 +1,34 @@
+CAPI=2:
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:dv:top_earlgrey_ast_top:0.1"
+description: "Pseudo top-level for Earl Grey's ast"
```
vogelpi, line 6: Would you mind elaborating a bit on why this is needed? I see you have added a comment to the commit messages but I fear I am missing some background here. No need to update the comment but I would like to better understand this, just out of curiosity.

hw/top_darjeeling/ip/sensor_ctrl/sensor_ctrl.core
```diff
@@ -2,7 +2,7 @@ CAPI=2:
 # Copyright lowRISC contributors (OpenTitan project).
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
-name: "lowrisc:systems:sensor_ctrl:0.1"
+name: "lowrisc:systems:top_darjeeling_sensor_ctrl:0.1"
 description: "Sensor Control"
```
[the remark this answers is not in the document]
a-will: Nope, I don't think so! `sensor_ctrl` is a core representing an IP with modules, which should be a self-contained thing. There's no virtual "interface" to an anonymous sensor_ctrl IP that other cores would need to depend on. If sensor_ctrl were to export constants or types, there might be an argument for a virtual sensor_ctrl_pkg, though.

We should avoid introducing virtual cores unless it's a necessary model. There can only be one implementation in a fusesoc run, and there is overhead in maintaining top files to ensure the correct core is brought in.

hw/top_darjeeling/ip/sensor_ctrl/sensor_ctrl_pkg.core
```diff
@@ -2,16 +2,16 @@ CAPI=2:
 # Copyright lowRISC contributors (OpenTitan project).
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
-name: "lowrisc:systems:sensor_ctrl_pkg:0.1"
+name: "lowrisc:systems:top_darjeeling_sensor_ctrl_pkg:0.1"
 description: "Sensor Control Package"
```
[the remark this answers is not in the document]
a-will: Only if we intend to create a non-top-specific IP (i.e. something in `hw/ip`) that consumes constants or types from sensor_ctrl_pkg. As of right now, there are no users of a virtually-provided sensor_ctrl_pkg, so... nope! :smile:

hw/top_darjeeling/ip/sensor_ctrl/sensor_ctrl_reg.core
```diff
@@ -2,15 +2,15 @@ CAPI=2:
 # Copyright lowRISC contributors (OpenTitan project).
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
-name: "lowrisc:systems:sensor_ctrl_reg:0.1"
+name: "lowrisc:systems:top_darjeeling_sensor_ctrl_reg:0.1"
 description: "Sensor Control Package"
```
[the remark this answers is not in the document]
a-will: For the same reason as `lowrisc:systems:sensor_ctrl`, no. However, I've neglected to place `sensor_ctrl_reg_pkg.sv` in the `sensor_ctrl_pkg` core. We should do that if we actually want other IPs to use its definitions.

In general, I think SV modules (except for prim library cores!) should never be part of virtual VLNVs. IPs should always get a direct dependency to the real VLNV. It's the packages and macros (and maybe classes and such for DV) that could get reuse and may need virtual VLNVs for agreement on interfaces.

hw/top_earlgrey/ip/ast/top_earlgrey_ast_top.core
a-will: Ah, this is kind of an artifact of englishbreakfast sharing cores with earlgrey. Because the same exact `ast` core is used, earlgrey's `ast` is unable to depend directly on the _real_ earlgrey-specific VLNVs for the package files below (for alert_handler, clkmgr, pwrmgr, rstmgr, and pinmux). Consequently, it must depend on virtual VLNVs, so englishbreakfast can substitute its cores as dependencies.

Using virtual VLNVs has side effects, though. We don't want fusesoc to choose just _any_ clkmgr, pwrmgr, etc. package that is in the repo, so when we supply the system core to dvsim (e.g. in top_earlgrey_lint_cfgs.hjson), we still need a core that explicitly depends on the desired providers.

Side note: I haven't touched up darjeeling for any of this. Its ast core should really just depend on the real VLNVs, but it doesn't right now.

a-will:
@matutem @rswarbrick In case it helps cut down webpage loading time for a more DV-focused review, this is the relevant commit for changes to DV files (i.e. that aren't fusesoc core files):

[8b80208](https://github.com/lowRISC/opentitan/pull/25773/commits/8b8020869d965a1c2efe931a7b4083e6b233aba6)

If we end up doing a rebase and the commit hash changes, it's the one with this description:

```
[clkmgr,pwrmgr,rstmgr] Move the shared SVA interfaces out

Move the shared interfaces for SVAs out of the templated IP directories
to avoid duplicate cores.
```

hw/top_darjeeling/top_darjeeling_ibex_pmp_reset_pkg.core
```diff
@@ -8,6 +8,9 @@ CAPI=2:
 # Tracked in https://github.com/olofk/fusesoc/issues/235
 name: "lowrisc:constants:top_darjeeling_ibex_pmp_reset_pkg"
 description: "Top-level specific PMP reset settings for Darjeeling"
+virtual:
+  - lowrisc:constants:ibex_pmp_reset_pkg
```
matutem, line 12: Could you place a string implying it is virtual in this name?

hw/top_darjeeling/top_darjeeling_jtag_id_pkg.core
```diff
@@ -5,6 +5,9 @@ CAPI=2:
 
 name: "lowrisc:constants:top_darjeeling_jtag_id_pkg"
 description: "jtag id for top_darjeeling"
+virtual:
+  - lowrisc:constants:jtag_id_pkg
```
matutem, line 9: Same thing regarding a "virtual" hint in the name.

hw/ip_templates/alert_handler/data/alert_handler.tpldesc.hjson
```diff
@@ -51,5 +51,11 @@
       type: "object"
       default: []
     }
+    {
```
matutem, line 54: This should be unnecessary if we fix the VLNV convention :)  I guess the problem you face is that currently this name is not going through `instance_vlnv`?

hw/ip_templates/rstmgr/rstmgr.core.tpl
```diff
@@ -10,7 +10,7 @@ virtual:
 filesets:
   files_rtl:
     depend:
-      - lowrisc:ip:alert_handler_component
+      - lowrisc:ip_interfaces:alert_handler_pkg
```
matutem: This is a .tpl file: could you use ${instance_vlnv...} for this instead?

hw/top_darjeeling/ip/ast/ast.core
```diff
@@ -26,7 +26,7 @@ filesets:
       - lowrisc:ip:lc_ctrl_pkg
       - lowrisc:ip:edn_pkg
       - lowrisc:ip:entropy_src_pkg
-      - lowrisc:ip_interfaces:alert_handler_reg
+      - lowrisc:ip_interfaces:alert_handler_pkg
```
matutem, line 29: This is a top specific core file. Use the top_specific core instead of ip_interfaces?

hw/ip_templates/clkmgr/data/clkmgr.tpldesc.hjson
```diff
@@ -86,5 +86,23 @@
...
       default: "2"
     }
+    {
+      name: "with_alert_handler"
+      desc: "Generate outputs for a clkmgr that would connect to an alert handler"
+      type: "bool"
+      default: "1"
+    }
+    {
+      name: "pwrmgr_vlnv_prefix"
```
matutem, line 96: Another side-effect of the hopefully-to-be-replaced instance_vlnv implementation.

hw/ip_templates/clkmgr/data/clkmgr.tpldesc.hjson
```diff
@@ -86,5 +86,23 @@
...
+      desc: "Provides the VLNV prefix (for instance_vlnv) for the pwrmgr_pkg used by this clkmgr"
+      type: "string"
+      default: ""
+    }
+    {
```
matutem, line 101: Ditto regarding instance_vlnv.

hw/ip_templates/alert_handler/data/alert_handler.tpldesc.hjson
a-will: This was a conscious choice to not totally forbid running ipgen outside of topgen, actually.

At some point, this global package requirement should go away, though. It is not reasonable to require a globally defined `top_pkg::TL_<property>` to be consumed, since interconnection networks are not always uniform.

hw/ip_templates/clkmgr/doc/theory_of_operation.md.tpl
```diff
@@ -1,4 +1,4 @@
-# Theory of Operation
+<%text filter="n"># Theory of Operation
```
matutem, line 1: Are you doing this to avoid mako from interpreting ## as a comment? I tend to replace such by ${"##"} instead. The %text directive ends up affecting the full file and ends requiring placing </%text> around specific spots, so it over-reaches and can create confusion if they are not correctly paired. By the way, a single '#' is not a problem.

hw/ip_templates/rstmgr/rstmgr.core.tpl
a-will: Haha, you've probably already found that happens on a later commit ;)

In order to keep each commit focused, there needed to be some carrying of virtual dependencies until the commit that changed that specific template.

hw/ip_templates/clkmgr/doc/theory_of_operation.md.tpl
a-will: Yes, and I didn't much feel like going and modifying a ton of doc lines for this PR.

Personally, I'd just use HTML header tags (which is still markdown syntax) instead of markdown's special header syntax wrapped in a mako escape. Then mako won't get in the way.

a-will: It looks like the convention has been to wrap the header alone in the text directive (which I forgot from https://github.com/lowRISC/opentitan/pull/19157 ). I'll make that change.

If you'd like to change that style convention, I'd recommend raising it with the folks who have done focused work on maintaining the websites.

hw/ip_templates/clkmgr/data/clkmgr.tpldesc.hjson
a-will: They might change some, but I don't think we can remove these template variables altogether. At some point, we need to allow multiple specializations of clkmgr / pwrmgr / rstmgr in one topgen run.

hw/top_darjeeling/top_darjeeling_ibex_pmp_reset_pkg.core
a-will: I think I'm going to get rid of this virtual dependency. This package is only consumed in the top level template, so there is no need for a virtual VLNV.

hw/top_darjeeling/top_darjeeling_jtag_id_pkg.core
a-will: Same as the virtual ibex_pmp_reset_pkg. It turns out there are no users of a virtual `lowrisc:constants:jtag_id_pkg`, so let's get rid of the virtual VLNV. :smile:

hw/top_earlgrey/ip/ast/top_earlgrey_ast_top.core
vogelpi: Thanks for explaining, this makes sense.

hw/top_darjeeling/ip/ast/ast.core
a-will: Whoops, I think I accidentally went the opposite direction on the latest push and created another core to tie things together. I totally agree with your statement and will switch this to a direct dependency.

a-will:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/alert_handler/data/alert_handler.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/pwrmgr/data/pwrmgr.hjson

PR #25774, opened 2025-01-04
[top,darjeeling] Darjeelling uses 27 user bits

Razer6:
Lets change it. This is used to accommodate the RACL and CTN ID information.


vogelpi (approved):
LGTM!

PR #25775, opened 2025-01-06
Replace autogen DIF code by bazel rules

pamaury:
This PR changes the way we manage the autogenerated part of the DIFs: instead of checking-in the autogenerated files, which is not super compatible with multitop, move those to bazel rules. This requires to split `make_new_dif` into two: the actual autogen part, and the part generating actually new DIFs.


util/autogen_dif.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Render DIF templates.
+    template_path = REPO_TOP / "util/make_new_dif/templates"
+    # Render all templates
+    for filetype in [".h", ".c", "_unittest.cc"]:
+        # Build input/output file names.
+        template_file = template_path / f"dif_autogen{filetype}.tpl"
+        out_file = (args.outdir /
+                    f"dif_{ip.name_snake}_autogen{filetype}")
+        print(out_file)
```
Razer6, line 54: Consider dropping that print or adding a proper log message. Otherwise, the output might be confusing.

Razer6 (approved):
Thanks @pamaury. This looks good to me.

util/autogen_dif.py
pamaury: Good catch, I should have removed that.

a-will (approved):
LGTM

I like that some of our path assumptions are removed here :smile:

jwnrt (approved):
LGTM, thanks

util/make_new_dif/ip.py
```diff
@@ -89,7 +88,7 @@ class Ip:
     """
 
     def __init__(self, name_snake: str, name_long_lower: str,
-                 ipgen_modules: List[str], reggen_top_modules: List[str]) -> None:
+                 hjson_file: str) -> None:
```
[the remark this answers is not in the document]
pamaury: Done, good catch :)

util/autogen_testutils.py
```diff
@@ -64,9 +65,22 @@ def main():
         # (/path/to/dif_uart_autogen.c) and returns the IP name in lower
         # case snake mode (i.e., uart).
         ip_name_snake = Path(autogen_dif_filename).stem[4:-8]
+
+        # Load HJSON data.
+        if ip_name_snake in ipgen_modules:
+            data_dir = REPO_TOP / "hw/{}/ip_autogen/{}/data".format(
+                topcfg.name, ip_name_snake)
```
matutem, line 72: Should this be "top_<topcfg.name>"? topcgf name is like "earlgrey", not "top_earlgrey".

util/make_new_dif.py
```diff
@@ -55,6 +55,18 @@
 ALL_PARTS = ["header", "checklist", "autogen"]
 
 
+def get_ip_hjson_path(name_snake, ipgen_modules, reggen_top_modules):
+    if name_snake in ipgen_modules:
+        data_dir = REPO_TOP / "hw/top_earlgrey/ip_autogen/{0}/data".format(
```
matutem: Why the explicit top_earlgrey? Should topname be an argument or deduced from the topcfg?

util/make_new_dif.py
```diff
@@ -55,6 +55,18 @@
...
+            name_snake)
+    elif name_snake in reggen_top_modules:
+        data_dir = REPO_TOP / "hw/top_earlgrey/ip/{0}/data/".format(
```
matutem: Nit: it is better to break on / for a Path, like "hw" / "top.." since the path separator itself is platform dependent.

util/autogen_testutils.py
```diff
@@ -64,9 +65,22 @@ def main():
         # (/path/to/dif_uart_autogen.c) and returns the IP name in lower
         # case snake mode (i.e., uart).
         ip_name_snake = Path(autogen_dif_filename).stem[4:-8]
+
+        # Load HJSON data.
+        if ip_name_snake in ipgen_modules:
```
matutem, line 70: This should be a function in topgen lib (not in the PR if you prefer).

util/autogen_dif.py
```diff
@@ -0,0 +1,72 @@
...
+    template_path = REPO_TOP / "util/make_new_dif/templates"
+    # Render all templates
+    for filetype in [".h", ".c", "_unittest.cc"]:
+        # Build input/output file names.
+        template_file = template_path / f"dif_autogen{filetype}.tpl"
+        out_file = (args.outdir /
+                    f"dif_{ip.name_snake}_autogen{filetype}")
+
+        # Read in template.
+        template = Template(template_file.read_text(),
```
matutem, line 56: Add exception handling.

util/autogen_dif.py
```diff
@@ -0,0 +1,72 @@
...
+                            strict_undefined=True)
+
+        # Generate output file.
+        out_file.write_text(
```
matutem, line 60: Add exception handling.

rules/autogen.bzl
```diff
@@ -124,6 +124,86 @@ opentitan_ip_rust_header = rule(
...
+        "ip": attr.string(mandatory = True, doc = "Name of the IP for which to generate the DIF"),
+        "output_groups": attr.string_list_dict(
+            allow_empty = True,
+            doc = """
+                Mappings from output group names to lists of paths contained in
+                that group.
+            """,
+        ),
+        "_autogen_dif": attr.label(
+            default = "//util:autogen_dif",
```
matutem, line 180: Nt: autogen_dif is a misnomer since there is nothing "auto" about invoking a tool to generate something. "gen_dif" is better, but I know "autogen" is unfortunately pervasive in our project.

matutem:
Amaury, I left some comments. You can ignore some of the Nits :)

It is not clear this deals with the path change for ipgens, and for these did you fix the path in the href near the top in files like dif_clkmgr_autogen.h? It should refer to a specific top.

util/autogen_testutils.py
pamaury: done

util/autogen_testutils.py
pamaury: yes, done

util/make_new_dif.py
pamaury: fixed

util/make_new_dif.py
pamaury: It is my understanding that `Path` only works with `/`, on all platforms, but I am happy to change it anyway.

util/autogen_dif.py
pamaury: Done

util/autogen_dif.py
pamaury: done for render

rules/autogen.bzl
pamaury: This tool is invoked by bazel which I guess is why it was called `autogen` to start with. But generally I agree that gen would be better.

util/make_new_dif.py
matutem: The "/" operator on paths is universal, and you are right: Gemini confirms a slash embedded in a path string is universal.

util/autogen_dif.py
```diff
@@ -0,0 +1,85 @@
...
+from pathlib import Path
+
+from autogen_banner import get_autogen_banner
+from make_new_dif.ip import Ip
+from mako.template import Template
+from mako import exceptions
+
+# This file is $REPO_TOP/util/make_new_dif.py, so it takes two parent()
+# calls to get back to the top.
+REPO_TOP = Path(__file__).resolve().parent.parent
```
matutem, line 19:
```suggestion
REPO_TOP = Path(__file__).resolve().parents[1]
```

I think repo_top could be provided by a library like util/base/lib.py, since all sorts of tools get it in all sorts of ways, and we could make a single more reliable way to get it. It could, for example, at least check that REPO_TOP / "util" is_dir().

A potentially more stable way to get REPO_TOP is from git, as in https://github.com/lowRISC/opentitan/blob/master/util/dvsim/dvsim.py#L154, but that fires a new subprocess.

util/topgen/lib.py
```diff
@@ -368,6 +368,23 @@ def search_ips(ip_path):  # return list of config files
     return ips
 
 
+def get_ip_hjson_path(ip_name_snake: str, topcfg: Dict[str, object], repotop: Path) -> Path:
```
matutem, line 371: This will traverse through all modules twice and generate lists, once for ipgens, once for reggens, and then the "in" operator on lists is also O(N).

It is probably much better to get the module with lib.find_module(ip_name_snake) and then decide based on lib.is_ipgen and lib.is_top_reggen.

matutem (approved):
The REPO_TOP comment is up to you, but I think creating a library function to retrieve repo_top will be good.

I would encourage you to take the comment about get_ip_hjson_path more seriously. Topgen uses code like I suggest and it is way more efficient and clean.

Otherwise LGTM

util/autogen_dif.py
pamaury: I don't think relying on git is good idea because if someone downloads the code as a tarball checkout, this will make the tool fail for no good reason. I generally agree that this way of doing things is ugly but I think the issue that our python codebase is fundamentally broken in how it is organized.
Anyway, I have removed this particular `REPO_TOP` as it turns out I could do without.

util/topgen/lib.py
pamaury: done

pamaury:
> The REPO_TOP comment is up to you, but I think creating a library function to retrieve repo_top will be good.
> 
> I would encourage you to take the comment about get_ip_hjson_path more seriously. Topgen uses code like I suggest and it is way more efficient and clean.
> 
> Otherwise LGTM

I have address your comment about `get_ip_hjson_path`. I also fixed the one instance of `REPO_TOP` that you pointed to but I honestly don't know what the general solution for `REPO_TOP` should be so I prefer not to touch it for now. Maybe a solution would be:
- start from `__file__`
- walk up until you find `MODULE.bazel`

What do you think?

I also fixed an issue with `autogen_testutils.py`. I think this tool is not used by anything? It keeps breaking (it was broken before this PR) without noticing.

pamaury:
On second thought, I have introduce a new library, `repo_top.py` to find the top. It currently it only used in `autogen_dif.py`

matutem:
Thanks for adding the detection of repo_top. We can improve on this now, like recursing down till MODULE.bazel, BLOCKFILE, and/or util are found as you mention.

pamaury:
The CI failures are unrelated to this PR. The check for unauthorized changes is failing because the PR changes too many files but this PR does not actually change any unauthorized files.

Issue #24231, opened 2024-08-06
[entropy_src] Reference to main_sm_state in Stub mode uses incorrect path


calebofearth:
### Description

Issue is seen here:
https://github.com/lowRISC/opentitan/blob/3e8be7dd390bffc3ff26579f51c61f22259c3b1e/hw/ip/entropy_src/rtl/entropy_src.sv#L213
The code should say `stub_hw2reg.main_sm_state.d = 8'b01110110; `

calebofearth:
@howardtr

PR #25776, opened 2025-01-06
[entropy_src/rtl] Fix reg reference and default value for stubbed impl

vogelpi:
We currently don't use the stubbing anywhere in the codebase but some lint tools tend to flag issues despite the corresponding parameter being disabled.

This resolves lowRISC/OpenTitan#24231.


vogelpi:
CHANGE AUTHORIZED: hw/ip/entropy_src/rtl/entropy_src.sv

The RTL change of this PR has no impact as the affected parameterization is not used in any top level currently. This is fine.

nasahlpa:
CHANGE AUTHORIZED: hw/ip/entropy_src/rtl/entropy_src.sv

The RTL change of this PR has no impact as the affected parameterization is not used in any top level currently. This is fine.

PR #25778, opened 2025-01-06
[aes/rtl] Avoid double import of aes_pkg inside aes_wrap.sv

vogelpi:
The aes_pkg is already imported in the port declaration.

We don't use the aes_wrap.sv file (it's sole purpose is/was to enable simulation-based FI experiments), some lint tools flag this.


nasahlpa (approved):

vogelpi:
CHANGE AUTHORIZED: hw/ip/aes/rtl/aes_wrap.sv

The RTL file touches by this PR isn't used in any top level. This is fine.

PR #25779, opened 2025-01-06
Fix `alert_handler_lpg_sleep_mode_ping` tests for silicon (and other envs)

jwnrt:
* FPGA support is extended to the CW340.
* DV timings are corrected to match the DV.
* Silicon support is added.
* Unknown environments are now handled explicitly.

Note that I can't currently test this on the FPGA due to the flash splicing problem.


jwnrt:
Deferring the cherry-pick to master until we get feedback that this works.

PR #25780, opened 2025-01-06
[crypto] Fix the KMAC-KDF implementation to return the output in shares.

jadephilipoom:
Previously, the KMAC driver always unmasked the shares of the digest returned by the KMAC block. This is OK for hash functions like SHA-3, but problematic for KMAC-KDF where the output is key material. The implementation also did not modify the second half of the keyblob buffer in KMAC-KDF, which would actually cause correctness issues if those values were not initialized to zero.

This commit refactors the KMAC driver to allow returning the digest in masked form, and adjusts KMAC-KDF to use this new capability.

While I was messing with this code anyway, I also made a small tweak to the way that KMAC-KDF test vectors are formed; the previous setup defined the expected output as a full blinded key struct with most of the fields uninitialized, but this is unnecessary and confusing in my opinion given all we care about is the unmasked value. I changed it to a simple word buffer.

### DIFF
--- a/sw/device/lib/crypto/drivers/kmac.c
+++ b/sw/device/lib/crypto/drivers/kmac.c
@@ -595,20 +595,23 @@ static status_t kmac_write_key_block(kmac_blinded_key_t *key) {
  * The caller must ensure that `message_len` bytes (rounded up to the next 32b
  * word) are allocated at the location pointed to by `message`, and similarly
  * that `digest_len_words` 32-bit words are allocated at the location pointed
- * to by `digest`.
+ * to by `digest`. If `masked_digest` is set, then `digest` must contain 2x
+ * `digest_len_words` to fit both shares.
  *
  * @param operation The operation type.
  * @param message Input message string.
  * @param message_len Message length in bytes.
  * @param digest The struct to which the result will be written.
  * @param digest_len_words Requested digest length in 32-bit words.
+ * @param masked_digest Whether to return the digest in two shares.
  * @return Error code.
  */
 OT_WARN_UNUSED_RESULT
 static status_t kmac_process_msg_blocks(kmac_operation_t operation,
                                         const uint8_t *message,
                                         size_t message_len, uint32_t *digest,
-                                        size_t digest_len_words) {
+                                        size_t digest_len_words,
+                                        hardened_bool_t masked_digest) {
   // Block until KMAC is idle.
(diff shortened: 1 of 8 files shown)


sw/device/lib/crypto/drivers/kmac.c
```diff
@@ -691,14 +694,37 @@ static status_t kmac_process_msg_blocks(kmac_operation_t operation,
...
+        digest[idx + i] = abs_mmio_read32(kKmacStateShare0Addr +
+                                          (offset + i) * sizeof(uint32_t));
+      }
+      for (; launder32(idx) < digest_len_words && offset < keccak_rate_words;
+           offset++) {
+        digest[idx + digest_len_words] =
+            abs_mmio_read32(kKmacStateShare1Addr + offset * sizeof(uint32_t));
+        idx++;
+      }
+    } else {
```
milesdai, line 713: I'm having a bit of difficulty verifying the register-reading logic here with three different indices being changed (offset, idx, and i) and offset not actually being used in the first loop. What do you think about something like the following refactor that reinforces the parallel behavior of the two loops and just updates `idx` at the end after everything is done?

```C
if (launder32(masked_digest) == kHardenedBoolTrue) {
      HARDENED_CHECK_EQ(masked_digest, kHardenedBoolTrue);
      // Read the digest into each share in turn. Do this in separate loops so
      // corresponding shares aren't handled close together.
      for (offset = 0; launder32(idx + offset) < digest_len_words && offset < keccak_rate_words;
            offset++) {
            digest[idx + offset] = abs_mmio_read32(kKmacStateShare0Addr + offset * sizeof(uint32_t));
      }
      size_t word_count = 0;
      for (offset = 0; launder32(idx + offset) < digest_len_words && offset < keccak_rate_words;
           offset++) {
           digest[idx + digest_len_words + offset] =
                  abs_mmio_read32(kKmacStateShare1Addr + offset * sizeof(uint32_t));
            word_count++;
      }
      idx += word_count;
      // or just idx += offset - 1;
} else {
```

milesdai (approved):
LGTM modulo the comment!

sw/device/lib/crypto/drivers/kmac.c
jadephilipoom: You're right, that's better -- I replaced it with your version (the `idx += offset` variant, rather than introducing `word_count` separately).

PR #25783, opened 2025-01-07
[darjeeling] Feed I2C RAM config to toplevel interface

Razer6:
For completion, also include the I2C ram config interface to the top-level interface of darjeeling.

### DIFF
--- a/hw/top_darjeeling/data/autogen/top_darjeeling.gen.hjson
+++ b/hw/top_darjeeling/data/autogen/top_darjeeling.gen.hjson
@@ -906,6 +906,10 @@
           act: rcv
           width: 1
           inst_name: i2c0
+          default: ""
+          external: true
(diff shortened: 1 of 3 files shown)


rswarbrick (approved):
Looks sensible to me. I'm pleasantly surprised to see how small the change is!

As a nitty change, would it make sense to line up `i2c_ram_1p_cfg_i` and `i2c_ram_1p_cfg_rsp_o`?

Razer6:
> As a nitty change, would it make sense to line up `i2c_ram_1p_cfg_i` and `i2c_ram_1p_cfg_rsp_o`?

That is part of the code generator. The alignment here depends on the used type for the signal interface before. I don't think that this PR is the right place to unify that.

vogelpi (approved):
LGTM

PR #25787, opened 2025-01-07
[dv, base_reg] Adding a prediction flag to field and mask to reg

antmarzam:
The base scoreboard ignores register reads checks unless the prediction flag is set. The fields that are predicted in a given register can be then accessed via a mask in dv_base_reg

### DIFF
--- a/hw/dv/sv/dv_base_reg/dv_base_reg.sv
+++ b/hw/dv/sv/dv_base_reg/dv_base_reg.sv
@@ -73,6 +73,21 @@ class dv_base_reg extends uvm_reg;
     `downcast(get_dv_base_reg_block, get_parent())
   endfunction
 
+  function uvm_reg_data_t get_predicted_mask();
+    uvm_reg_data_t mask = 0;
+    dv_base_reg_field fields_q[$];
+    this.get_dv_base_reg_fields(fields_q);
+
+    foreach (fields_q[i]) begin
+      if (fields_q[i].has_prediction) begin
+        for (int j = 0; j < fields_q[i].get_n_bits(); j++)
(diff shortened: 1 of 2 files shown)


hw/dv/sv/dv_base_reg/dv_base_reg.sv
```diff
@@ -73,6 +73,20 @@ class dv_base_reg extends uvm_reg;
...
   endfunction
 
+  function uvm_reg_data_t get_predicted_mask();
+    uvm_reg_data_t mask = 0;
+    dv_base_reg_field fields_q[$];
+    this.get_dv_base_reg_fields(fields_q);
+
+    foreach (fields_q[i]) begin
+      if (fields_q[i].has_prediction) begin
+        mask[i] |= fields_q[i].has_prediction;
```
rswarbrick, line 83: Is this right? I think that `mask` is addressed by bits, but `fields_q` is addressed by field index. I don't think this will work if a field has length greater than one or if there is a gap?

antmarzam: Oops - good spot! Thank you. 
I was thinking of intr_state which is 1-bit per field but that's not always the case. Tweaking to be:
```
        for (int j = 0; j < fields_q[i].get_n_bits(); j++)
          mask[j+fields_q[i].get_lsb_pos] = 1'b1;

```

rswarbrick (commented):
This looks good to me, thanks.

rswarbrick (approved):
Happy to merge myself. Sorry, I don't remember why I didn't do so earlier!

PR #25788, opened 2025-01-07
Cip base scoreboard intr state check

antmarzam:
This PR depends on https://github.com/lowRISC/opentitan/pull/25787

The first commit will disappear after #25787 is merged on master

### DIFF
--- a/hw/dv/sv/cip_lib/cip_base_scoreboard.sv
+++ b/hw/dv/sv/cip_lib/cip_base_scoreboard.sv
@@ -210,23 +210,24 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
     if (!item.is_write()) begin
       uvm_reg csr = cfg.ral_models[ral_name].default_map.get_reg_by_offset(item.a_addr);
       if (csr != null) begin
-        if (!uvm_re_match("intr_state*", csr.get_name())) begin
-          // If intr_state
-          uvm_reg_data_t pred_mask;
-          dv_base_reg dv_reg;
-          `downcast(dv_reg, csr)
-          pred_mask = dv_reg.get_predicted_mask();
-          `DV_CHECK_EQ(dv_reg.get_mirrored_value() & pred_mask,
-                       item.d_data & pred_mask,
-                       $sformatf("Register read data for %0s (prediction mask %0h)",
-                       csr.get_full_name(),
-                       pred_mask))
-        end
+        dv_base_reg dv_reg;
+        `downcast(dv_reg, csr)
+        do_read_check(dv_reg, item);
       end
     end
-
     process_tl_access(item, DataChannel, ral_name);
   endtask
+
+  // This function implements the generic read checks, any TB can override this method to
+  // expand the checks further
+  virtual function void do_read_check(dv_base_reg dv_reg, tl_seq_item item);
+    uvm_reg_data_t pred_mask = dv_reg.get_predicted_mask();
+    `DV_CHECK_EQ(dv_reg.get_mirrored_value() & pred_mask,
+                 item.d_data & pred_mask,
+                 $sformatf("Register read data for %0s (qualified by prediction mask %0h)",
+                           dv_reg.get_full_name(),
+                           pred_mask))
+  endfunction
 
   virtual task process_alert_fifos();
     foreach (alert_fifos[i]) begin
--- a/hw/dv/sv/cip_lib/cip_base_scoreboard.sv
+++ b/hw/dv/sv/cip_lib/cip_base_scoreboard.sv
@@ -207,6 +207,24 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
 
     if (!cfg.en_scb) return;
 
+    if (!item.is_write()) begin
+      uvm_reg csr = cfg.ral_models[ral_name].default_map.get_reg_by_offset(item.a_addr);
+      if (csr != null) begin
+        if (!uvm_re_match("intr_state*", csr.get_name())) begin
+          // If intr_state
+          uvm_reg_data_t pred_mask;
+          dv_base_reg dv_reg;
+          `downcast(dv_reg, csr)
+          pred_mask = dv_reg.get_predicted_mask();
+          `DV_CHECK_EQ(dv_reg.get_mirrored_value() & pred_mask,
+                       item.d_data & pred_mask,
+                       $sformatf("Register read data for %0s (prediction mask %0h)",
+                       csr.get_full_name(),
+                       pred_mask))
+        end
+      end
+    end
+
     process_tl_access(item, DataChannel, ral_name);
   endtask
 
--- a/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
+++ b/hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
@@ -458,6 +458,7 @@ class cip_base_vseq #(
         uvm_reg_data_t data = $urandom();
         `uvm_info(`gfn, $sformatf("Write %s: 0x%0h", intr_csrs[i].`gfn, data), UVM_MEDIUM)
         csr_wr(.ptr(intr_csrs[i]), .value(data));
+        if (cfg.under_reset) break;
       end
 
       // Read all intr related csr and check interrupt pins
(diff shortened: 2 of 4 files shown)


rswarbrick (commented):
I'd suggest rebasing this on top of the PR that it depends on. Then the PR message can say something like ("Depends on #25787, which should be merged first. Once that has been merged, only the last two commits will remain").

That way, CI will do something useful :-)

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
```diff
@@ -207,6 +207,24 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
 
     if (!cfg.en_scb) return;
 
+    if (!item.is_write()) begin
+      uvm_reg csr = cfg.ral_models[ral_name].default_map.get_reg_by_offset(item.a_addr);
+      if (csr != null) begin
+        if (!uvm_re_match("intr_state*", csr.get_name())) begin
```
rswarbrick: Shouldn't this regex be something like `^intr_state[0-9]*$` ? (Otherwise, I think you'll match "intr_stat", "intr_stateeee" and "intr_statenotreally")

But shouldn't we just drop the check on CSR name entirely? I think we're planning to default to `pred_mask=0` for registers, so that won't cause any problems, and nothing inside this `if` block is special to `intr_state*`.

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
```diff
@@ -207,6 +207,24 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
...
+          // If intr_state
+          uvm_reg_data_t pred_mask;
+          dv_base_reg dv_reg;
+          `downcast(dv_reg, csr)
+          pred_mask = dv_reg.get_predicted_mask();
+          `DV_CHECK_EQ(dv_reg.get_mirrored_value() & pred_mask,
+                       item.d_data & pred_mask,
+                       $sformatf("Register read data for %0s (prediction mask %0h)",
```
rswarbrick: I think this might need tweaking slightly to something like "Register read data for %0s, qualified by prediction mask %0h" ? (Otherwise, the code might read `fe` when it expected `ff` and then generate a message talking about values `1` and `0` if the prediction mask happened to be `1` :-D)

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -474,24 +474,28 @@ class cip_base_vseq #(
           irq_ro_mask = intr_csrs[i].get_ro_mask();
         end
 
-        exp_val &= ~irq_ro_mask;
+        // we may need to update the exp_value here, since there may be a delay from the
+        // update above to ' exp_val'until here before we call the csr_rd
+        exp_val = `gmv(intr_csrs[i]) & ~irq_ro_mask;
```
rswarbrick, line 479: I'm not sure this is correct. The code above doesn't consume any time anyway. The `csr_rd` call *will* consume some time, though, so I think we should actually be computing `exp_val` just after we decide not to drop out after reset.

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -474,24 +474,28 @@ class cip_base_vseq #(
...
 
         csr_rd(.ptr(intr_csrs[i]), .value(act_val));
         act_val &= ~irq_ro_mask;
 
         if (cfg.under_reset) continue;
```
rswarbrick, line 484: I realise this code isn't new from you, but shouldn't this be `break` ? (Possibly worth adding a 1-line trivial commit to the PR)

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -474,24 +474,28 @@ class cip_base_vseq #(
...
-
-        // if it's intr_state, also check the interrupt pin value
-        if (!uvm_re_match("intr_state*", intr_csrs[i].get_name())) begin
-          interrupt_t exp_intr_pin = intr_csrs[i].get_intr_pins_exp_value();
-          interrupt_t act_intr_pin = cfg.intr_vif.sample();
-          act_intr_pin &= interrupt_t'((1 << cfg.num_interrupts) - 1);
-          `DV_CHECK_CASE_EQ(exp_intr_pin, act_intr_pin)
-        end // if (!uvm_re_match
+        if (intr_csrs[i].get_predicted_mask() == 0) begin
+          `DV_CHECK_EQ(exp_val, act_val, {"when reading the intr CSR", intr_csrs[i].`gfn})
```
rswarbrick, line 487: I realise that this was already dubious, but could you add a space after "CSR" here? Also (if you're touching the code) I'd call `get_full_name` explicitly rather than using the macro.

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
```diff
@@ -474,24 +474,28 @@ class cip_base_vseq #(
...
+
+          // if it's intr_state, also check the interrupt pin value
+          if (!uvm_re_match("intr_state*", intr_csrs[i].get_name())) begin
```
rswarbrick, line 490: I think this regex should be something like `^intr_state[0-9]*$`, to avoid matching "intr_stat", "intr_stateeee" and "intr_state_not_really".

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
```diff
@@ -207,27 +207,33 @@ class cip_base_scoreboard #(type RAL_T = dv_base_reg_block,
...
-    end
+    if (!item.is_write())
+      do_read_check(ral_name, item);
 
     process_tl_access(item, DataChannel, ral_name);
   endtask
 
+  // This function implements the generic read checks, any TB can override this method to
+  // expand the checks further
+  virtual function void do_read_check(string ral_name, tl_seq_item item);
```
rswarbrick, line 218: I think I'd suggest making this a bit narrower to something like `virtual function void on_register_read(dv_base_reg csr, tl_seq_item item);`

antmarzam: The reason why I was passing ral_name and item it was to derive the csr inside the function. I can derive the CSR in the task `process_tl_d_item` via defining it like this:
`    uvm_reg csr = cfg.ral_models[ral_name].default_map.get_reg_by_offset(item.a_addr);`

Would you prefer that? My thought process was to alter `process_tl_d_item` the least I could hence I embedded it in `do_read_check`. 

WRT the name of the function, I think having "check" in the name is helpful, but can change it to `on_register_read` if you disagree. I was trying to follow "uvm naming" with the "do_*" (do_read/write/predict/copy/clone...)

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
antmarzam: I think you're right. Given we have `pred_mask=0` for other registers this won't cause any issues. Fixed it. Will push changes shortly. 

WRT your regex suggestions. I tested it in online in a browser-based testing site and it worked, but for some reason it doesn't work for the UVM. Perhaps the semantics for regex in UVM are slightly different. 

I tried this:

```
        if (!uvm_re_match("intr_state*", csr.get_name())) begin
          if (!uvm_re_match("^intr_state[0-9]*$", csr.get_name())) begin
            $display("matching against: %s",csr.get_name());
          end
          else
            $display("not matching against: %s",csr.get_name());

          $fatal;

```
and get_name was returning `intr_state`

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
antmarzam: Done

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
antmarzam: Agreed - this was leftover comment from my previous attempted fix. I don't think it makes sense here anymore.

I don't think the call to `gmv()` can be after `csr_rd`, because by that time the mirrored value will be matching the returned value from the RTL.  This VSEQ tries to sample the mirrored value prior to the read, to then compare it against the actual value. 

I've basically reverted the VSEQ code to how it originally looked.

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
antmarzam: Done. 

Out of curiosity, why do you think it's better to use `get_full_name()` here instead of `gfn`?

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
antmarzam: I saw the same suggestion above, I've tested it in the uvm and it doesn't work

rswarbrick:
> I think I've rebased from PR#25787 correctly (I used magit). The commit from that PR is now showing here. Would that not mean PR#25787 is no longer needed?

This is completely correct. We just have to make sure we don't merge this one by accident before #25787, but a note (for next time) can avoid people wasting time reviewing the first commit, which comes from some other PR. Once #25787 has been merged, we can rebase the branch for this PR and the commit will disappear from the stuff that needs adding.

antmarzam:
> > I think I've rebased from PR#25787 correctly (I used magit). The commit from that PR is now showing here. Would that not mean PR#25787 is no longer needed?
> 
> The change you've made is completely correct. We just have to make sure we don't merge this one by accident before #25787, but a note (for next time) can avoid people wasting time reviewing the first commit, which comes from some other PR. Once #25787 has been merged, we can rebase the branch for this PR and the commit will disappear from the stuff that needs adding.

Noted! I've also added it to the first comment.

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
rswarbrick: About passing the register: yes, I think it makes sense to put that before the call. We're always going to want to do the same thing (I think) and it makes sense to avoid the duplication.

About the function name: that sounds completely reasonable. I didn't think about the name at all :-)

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
antmarzam: Does that mean we're leaving `uvm_re_match` as is, or rather creating a new PR to not define `UVM_REGEX_NO_DPI` and then changing this?

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
antmarzam: To avoid defining:
`uvm_reg csr = cfg.ral_models[ral_name].default_map.get_reg_by_offset(item.a_addr)`
Before doing:
 `do_read_check(csr);`
I'm just passing the csr directly as:
`      do_read_check(cfg.ral_models[ral_name].default_map.get_reg_by_offset(item.a_addr));`

hw/dv/sv/cip_lib/seq_lib/cip_base_vseq.sv
antmarzam: Cool - I've added another `break` to the write bit above

hw/dv/sv/cip_lib/cip_base_scoreboard.sv
rswarbrick: I think I'd suggest doing slightly more at the call site. Can't you downcast to pass the `dv_base_reg` (and move the null check outside too)?

antmarzam: Gotcha - I think it should be done now. The function receives `dv_base_reg` for the mirrored_value + predicted mask as well as `tl_seq_item` for the actual returned value

rswarbrick (approved):
Thanks for this. I'm convinced that this is the right way to go but even if we change our minds, it should be very easy to back out.

PR #25790, opened 2025-01-07
[WIP][topgen] Add class to load a complete topcfg properly

pamaury:
**Context:**

TODO

**Technical details:**

At the moment, topgen simplies serializes the complete topcfg using hjson.dumps() but this poses a few problems. The biggest one is that it is a mix of manual conversion to dict (when calling as_dict() in the various topgen functions) and automatic (when calling _asdict() from hjson.dumps). Furthermore, it turns out that we are missing some fields that probably were never added to _asdict().

This commit introduces a new class (CompleteTopCfg) whose sole purpose is to take the produced Hjson and reconstruct an in-memory topcfg that is *exactly* equivalent (in Pythonic types) to the ones that was dumped. This requires to sometimes reconstruct some classes, sometimes not. Classes that need to be reconstruct get a new method (fromdict) and the CompleteTopCfg does as much automatic deserializing as possible, then some manual fixing.

There is also a small tweak to how clock groups are handled: in alert_lpgs, the clock groups are all elements of the clocks.groups but because of the naive serializing, a full copy is made. This commit introduces GroupProxy that behaves like a Group for all purposes but serializes only to the group name. This makes the resulting hjson clearer and the deserialization easier.

Since this process is quite fragile, when topgen is running it will actually check that this works by dumping the Hjson, reloading it with the CompleteTopCfg and then checking that the two are equivalent.


util/topgen.py
```diff
@@ -500,12 +504,6 @@ def generate_flash(topcfg: Dict[str, object], out_path: Path) -> None:
         return
 
     params = vars(flash_mems[0]["memory"]["mem"]["config"])
-    # Additional parameters not provided in the top config.
```
matutem, line 503: How are these parameters passed to flash_ctrl?

util/topgen/clocks.py
```diff
@@ -20,6 +20,10 @@ def _bool_to_yn(val: bool) -> str:
     return 'yes' if val else 'no'
 
 
+def _bool_from_yn(val: str) -> bool:
```
matutem, line 23: reggen provides check_bool that does a more stringent check. It also provides some other checkers like check_int.

util/topgen/validate.py
```diff
@@ -336,16 +336,32 @@ def check_values(self):
...
+        # Do not include base_addrs as it will get removed later.
         return {
+            'class': 'Flash',
             'banks': self.banks,
             'pages_per_bank': self.pages_per_bank,
             'program_resolution': self.program_resolution,
             'pgm_resolution_bytes': self.pgm_resolution_bytes,
             'bytes_per_page': self.bytes_per_page,
             'bytes_per_bank': self.bytes_per_bank,
+            'data_width': self.data_width,
```
matutem, line 348: Same as above... These sizes need to be parameterized for flash_ctrl.

matutem (commented):
Some questions about flash parameters and I would suggest looking into the checkers in reggen, like check_bool and check_int: they check and error out if needed.

One more comment: I am having to deal with validate_top and found it is incomplete and inflexible. It is not directly along this PR but is "adjacent".

util/topgen.py
pamaury: Not sure I understand your question but that's probably because my knowledge of topgen is superficial. I don't really know what these parameters are used for but I noticed that they are set in https://github.com/lowRISC/opentitan/blob/master/util/topgen/validate.py#L300 and it doesn't make sense that `topgen.py` overwrites them here?

util/topgen/clocks.py
pamaury: Ok, I will look into using this possibly. Though this check is only used on a topgen generated file so we know that the value is either 'yes' or 'no'. also I put it here so that we have `_boot_to_yn` and `_boot_from_yn` next to each other, if that makes sense?

util/topgen/validate.py
pamaury: I would agree, I don't know why it's not already the case. They are hardcoded in the `Flash` class in `validate.py`.

pamaury:
@matutem I have done the fixes to flash_ctrl as we discussed. I also put them (and the resets and group proxy changes) in separate commits.

util/topgen/clocks.py
```diff
@@ -158,6 +187,14 @@ def _asdict(self):
             "name": self._grp.name
         }
 
+    @classmethod
+    def fromdict(cls, proxy: Dict[str, object], clocks: object) -> object:
+        proxy['_grp'] = clocks.groups[proxy["name"]]
```
Razer6, line 192: Nit: Can we use single-ticks for strings?

Razer6 (approved):
Apart from other comments, this looks good.

util/topgen/clocks.py
pamaury: Yes single-ticks are exactly like double-ticks in Python

util/topgen/clocks.py
pamaury: I have made the check more stringent. I can't use `check_bool` because it only accept `0`, `1`, `true`, `false` whereas we use `yes`/`no` in that part of the code...

util/topgen/clocks.py
Razer6: I just mean, can we avoid mixing it? Especially on a single line.

PR #25791, opened 2025-01-07
[ci] fix dependabot config and expand to Cargo and GitHub Actions

nbdd0121:
The additional `reviewers` field is preventing dependabot from running. Fix it, and also expand the scanned dependencies to Rust crates and GitHub Actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,9 +10,8 @@ updates:
       # Updates are scheduled at the beginning of the work week.
       interval: "weekly"
       day: "monday"
-      time: "02:00"
-      # US Pacific Time
-      timezone: "America/Los_Angeles"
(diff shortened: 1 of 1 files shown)


jwnrt (approved):
Looks good, thanks. Is there a way to do a test run?

Also, I double checked and dependabot doesn't support Bazel yet (though renovate does).

nbdd0121:
I'm not aware of a way to trigger a manual run

jwnrt:
Oh, I see. I can't see the advisories for your fork.

jwnrt:
@nbdd0121 do you remember the status of this? Would fixing the config still be useful even if we don't get Cargo advisories?

nbdd0121:
Closing as this doesn't help the dependabot advisory

PR #25792, opened 2025-01-07
[silicon_creator] Make the `nonce.h` header available separately

pamaury:
The `nonce.c` file has some requirements that pull a lot of code. On the other hand, some headers only require the definition of a nonce and not the implementation that goes with it. This commit refactors the dependencies so that it is possible to depend on the nonce and boot data definition without depending on the flash_ctrl driver.

This will be useful for Darjeeling test_rom because it depends on the nonce and boot data definition but does not have a flash controller.


Razer6 (approved):
LGTM!

a-will (approved):

PR #25795, opened 2025-01-07
[deps,bazel] Rust dependency updates for build determinism & `cargo audit`

AlexJones0:
This PR updates the rust dependencies found in `third_party/rust` to solve determinism and security issues. Primarily due to the usage of outdated versions of `mdbook` (due to patching) and the unused `rust-crypto` crate, we had a lot of transitive dependencies with security advisories / vulnerabilities. There is also a separate commit for directly pulling in updates to the `rustix` and `object` crates which had build determinism issues (see commit message).

It might be nice to do a more general `cargo update`, but I'm conservatively updating only what is necessary for now as I was running into build issues in CI that presumably need fixes.

See the commit messages for more details.

### DIFF
--- a/third_party/rust/Cargo.lock
+++ b/third_party/rust/Cargo.lock
@@ -477,7 +477,7 @@ dependencies = [
  "encode_unicode",
  "lazy_static",
  "libc",
- "unicode-width",
+ "unicode-width 0.1.10",
  "windows-sys 0.45.0",
 ]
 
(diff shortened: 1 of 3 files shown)


pamaury (approved):
Thanks @AlexJones0 for the great effort in addressing non-determinism in the build Regarding cargo audit, we should really have an automated test to make sure that we do not regress without realizing.

jwnrt (approved):
Looks good, thanks

PR #25796, opened 2025-01-07
[hw,pinmux,rtl] Also gate assertions to signals related to strap sampling

Razer6:
These signals are only available when enabling strap sampling. If not, gate the assertions.


rswarbrick (approved):
Looks good to me, thanks.

PR #25797, opened 2025-01-07
[rust, bazel] separate mdbook into its own crate index

nbdd0121:
mdbook is only used as binary and we don't use it as a library, so making it a separate crate index. This allows us to avoid mixing its dependencies with other OT dependencies.


third_party/mdbook/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "cargo_raze_deps"
```
jwnrt, line 6: Can we rename both of these packages please? We're not using Raze anymore and somebody was confused by this name today.

Maybe just `name = "cargo_mdbook"` and `name = "cargo_opentitan"`?

jwnrt (commented):
Looks fine to me, but this could be a short `crate.spec(...)` call in the `MODULE.bazel` file. Please remind me why we're using a full Cargo manifest?

nbdd0121:
A Cargo manifest makes it slightly easier to precisely track and pin dependencies, as there're more tools (e.g. cargo-audit or cargo-deny) that handles Cargo.lock compared to Bazel lock.

jwnrt (approved):
Seems fine to me

jwnrt:
@nbdd0121 is this okay to merge? I'd like the lockfile changes for another PR

PR #25798, opened 2025-01-07
[rom_ctrl, dv] Conditional coverage hole of rom_ctrl_mux

KinzaQamar:
rom_ctrl coverage report contains a hole occurred in rom_ctrl_mux.sv by a conditional statement (alert_d = sel_invalid | sel_reverted | sel_q_reverted) as 001 not covered.

This PR fixes corrupt_sig_fatal_vseq to force sel_bus_qq in rom_ctrl_mux to an invalid value to get sel_q_reverted = 1. Then to make sure that we inject something bad to sel_bus_qq before ROM check finishes, the fix adds force_early variable.

Lastly, changed the name err_point to inject_after_done for better readibility.


rswarbrick (commented):
This looks sensible, but don't we need to add a new possible value for `cm_id`? With this change on its own, we are no longer going to force `sel_bus_q`?

rswarbrick:
Duh, sorry! I'm convinced by this change after all. (For future me: There are three values A,B,C and we check A==B and B==C. We want to see each of those checks failing separately, so poking B won't help)

rswarbrick (approved):

PR #25799, opened 2025-01-07
[dv,pwm] Parameterized channel count

alees24:
Modify the DV environment to support other channel counts; until now a channel count of 6 has been assumed.
Minor reformatting changes within the scoreboard.


hw/ip/pwm/dv/env/pwm_env_pkg.sv
```diff
@@ -62,11 +62,14 @@ package pwm_env_pkg;
...
-  // than ten channels.
-  function automatic int get_multireg_idx(string name);
-    string s = name.getc(name.len - 1);
-    return s.atoi();
+  // The index of a multi-reg is given by the last character(s) of the name. The number of PWM
+  // channels is parameterized.
+  function automatic int unsigned get_multireg_idx(string name);
+    int str_len = name.len();
+    // Note: this extracts the final two characters which are either '_y' or 'xy',
+    //       and because '_' is permitted in (System)Verilog numbers, it works for 0-99
```
rswarbrick, line 70: That's really clever!

There's a bit of me that's sad about the behaviour when this isn't actually an integer (it will get parsed as zero). If we were writing this from scratch, I'd suggest `$sscanf`. But that fact doesn't change with this PR, so let's not worry about it! :-)

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -71,20 +71,15 @@ endfunction
...
 task pwm_scoreboard::run_phase(uvm_phase phase);
   super.run_phase(phase);
-
-  // TODO: Decide whether we want to leave the processes running.
-  forever begin
-    `DV_SPINWAIT_EXIT(
+  if (cfg.en_scb) begin
+    // Create one checker process for each PWM output.
+    foreach (item_fifo[channel]) begin
+      // Ensure that the newly-created process does not see changes to the 'channel' when starting.
```
rswarbrick, line 77: Nit: I'd suggest saying something like `'channel' variable`, since we're talking about the channel itself some of the time in this file.

rswarbrick (approved):
A nitty comment about a comment, but this looks good to me.

hw/ip/pwm/dv/env/pwm_env_pkg.sv
alees24: With the current code the returned index will only be used in the event that the accessed register name matches, ie. if it is terminated with a valid number.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: Fair enough, that's me failing to read what I've typed; the definite article wasn't really supposed to be present...."changes to 'channel' when ..." but I've reworded it.

PR #25801, opened 2025-01-07
[rust] update crate index after manifest changed

timothytrippel:
PR #25786 updated the Rust crates used the project but did not update the crate index, thus breaking opentitanlib builds. This runs `bazel run //third_party/rust:crate_index -- --repin` to fix the issue.


engdoreis (approved):
Thanks @timothytrippel for fix this.

engdoreis:
There are another issues caused because some changes to Opentitanlib were not backported to Earlgrey_1.0.0. Should I add a commit to this PR?

timothytrippel:
closing as this as the fix will be much more involved, and until then, the changes in #25786 should be reverted to fix the branch

PR #25804, opened 2025-01-08
[topgen] Fix LPG Map for incoming alerts

Razer6:
This PR fixes the rendering of the LPG map in the alert handler for incoming alerts. The LPG's are independent for every incoming alert group where their index is only relative within the alert group. Previously, the LPGs index offset was added after every alert, which is wrong. This only needs to be done after dealing with all alerts of one alert group and when switing to the next alert group.


vogelpi (approved):
LGTM

PR #25808, opened 2025-01-08
Revert (most of) #25795 due to broken mdBook sidebar links

AlexJones0:
Though the version bump to mdBook was only minor it seems to have broken the sidebar links in the documentation.

This reverts the PR so that the documentation will continue working while I try to figure out what broke. 

Note that this does not revert commit 4cce56ec8dcb7741c5c4da0e60d337dcdba65cb0 from that PR because that is unrelated to the mdBook changes and is useful for build determinism.

### DIFF
--- a/third_party/rust/Cargo.lock
+++ b/third_party/rust/Cargo.lock
@@ -60,9 +60,9 @@ dependencies = [
 
 [[package]]
 name = "ammonia"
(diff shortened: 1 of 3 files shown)


AlexJones0:
Closed as superseded by #25813 which directly fixes the issue instead.

PR #25811, opened 2025-01-08
[rtl, aon_timer] Simpler condition to aid coverage closure

antmarzam:
The condition originally looked like:
```
 assign intr_test_qe           = reg2hw.intr_test.wkup_timer_expired.qe |
                                 reg2hw.intr_test.wdog_timer_bark
```
Which causes a conditional coverage hole due to both Qe signals being connected to the same signal in aon_timer_reg_top. In order to avoid waving the hole, the RTL has been ammended, and an assertion has been added to ensure things work as expected in case the RTL (register top) changes in the future

### DIFF
--- a/hw/ip/aon_timer/rtl/aon_timer.sv
+++ b/hw/ip/aon_timer/rtl/aon_timer.sv
@@ -198,8 +198,12 @@ module aon_timer import aon_timer_reg_pkg::*;
   );
 
   // Registers to interrupt
-  assign intr_test_qe           = reg2hw.intr_test.wkup_timer_expired.qe |
-                                  reg2hw.intr_test.wdog_timer_bark.qe;
+  assign intr_test_qe           = reg2hw.intr_test.wkup_timer_expired.qe;
+  // All the fields of a register have the same value for their qe signals, so we just pick one
+  // of them. This assertion checks that we don't miss writes if the register top changes."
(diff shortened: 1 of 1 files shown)


rswarbrick (commented):
This looks sensible to me. I think I'd probably be inclined to shorten the comment a bit though. Maybe "All the fields of a register have the same value for their qe signals, so we just pick one of them. This assertion checks that we don't miss writes if the register top changes." ?

@vogelpi, would you mind taking a look at this?

antmarzam:
I've swapped the original comment with your suggestion!

rswarbrick (approved):

rswarbrick:
CHANGE AUTHORIZED:  hw/ip/aon_timer/rtl/aon_timer.sv

This looks completely safe to me (as checked by an assertion). @vogelpi: as another pair of eyes, does this make sense to you too?

vogelpi (approved):
This looks good to me but I think it might result in a lint error, no? At least in Verilator and in Vivado, the assert macro translates to nothing and the tools will generate warnings. Maybe we should add an unused signal to avoid this.

vogelpi:
CHANGE AUTHORIZED: hw/ip/aon_timer/rtl/aon_timer.sv

This change avoids a coverage hole. In case a future change would enable the two signals ever being different, an SVA will catch that and warn us.

PR #25814, opened 2025-01-08
[dd, prim] Simplyfing condition to ease coverage closure

antmarzam:
The acknoledgement can't be set unless the request is set. The condition has been simplified to only check for the acknoledgment.

In addition, the new assertion checks if the ack is set, the request must be set

### DIFF
--- a/hw/ip/prim/rtl/prim_reg_cdc_arb.sv
+++ b/hw/ip/prim/rtl/prim_reg_cdc_arb.sv
@@ -154,7 +154,7 @@ module prim_reg_cdc_arb #(
     always_ff @(posedge clk_dst_i or negedge rst_dst_ni) begin
       if (!rst_dst_ni) begin
         id_q <= SelSwReq;
-      end else if (dst_update_req && dst_update_ack) begin
+      end else if (dst_update_ack) begin
         id_q <= SelSwReq;
       end else if (dst_req && dst_lat_d) begin
         id_q <= SelSwReq;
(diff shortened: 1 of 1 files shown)


hw/ip/prim/rtl/prim_reg_cdc_arb.sv
```diff
@@ -154,7 +154,11 @@ module prim_reg_cdc_arb #(
     always_ff @(posedge clk_dst_i or negedge rst_dst_ni) begin
       if (!rst_dst_ni) begin
         id_q <= SelSwReq;
-      end else if (dst_update_req && dst_update_ack) begin
+      end else if (dst_update_ack) begin
+        // The condition above was originally `dst_update_req && dst_update_ack`.
+        // However, since there can't be an ACK without a REQ that means the condition
+        // can be simplified to `dst_update_ack` only.
+        // Assertion `DstAckReqChk_A` checks the above always happens
```
rswarbrick, line 161: I don't think we need this comment in the code?

hw/ip/prim/rtl/prim_reg_cdc_arb.sv
```diff
@@ -165,6 +169,10 @@ module prim_reg_cdc_arb #(
       end
     end
 
+    // dst_update_ack should only be sent if there was a dst_update_req.
+    // By association, whenever src_ack is seen, then src_busy must be high.
```
rswarbrick, line 173: I don't think you wanted to copy this comment as well :-)

rswarbrick (commented):
I think I agree with the changes to the code. Please could you tidy up the comments and force-push? Then we can move to merging things :-)

rswarbrick (approved):

rswarbrick:

No functional change (by careful reasoning and an assertion that gets added). This should give no risk to earlgrey.

rswarbrick:
As a note: If #25705 gets merged before this PR, we'll need to update the exclusion file to drop the exclusions that this RTL change gets rid of.

alees24 (approved):

alees24:

PR #25815, opened 2025-01-08
[dv,pwm] Synchronize to monitor pulses when blink state changes

alees24:
The changes in this PR better accommodate the uncertainty at the start of the DUT operation by increasing the initial SettleTime value (necessary for a very small number of seeds; one in 1500 observed to requires 4 pulse cycles, when the phase counter is changing most rapidly), and improves the robustness of the verification by detecting the end of the first blink phase (BLINK_PARAM.X + 1 pulse cycles) and synchronizing the DV to the monitor/DUT output at that point.

Once synchronized, no further discrepancy between observed pulse cycles and expectations is tolerated, until a further configuration change occurs.

With the changes in this PR 1500 seeds (-i all -rx 5) have been observed to pass successfully.

### DIFF
--- a/hw/ip/pwm/dv/env/pwm_scoreboard.sv
+++ b/hw/ip/pwm/dv/env/pwm_scoreboard.sv
@@ -144,12 +144,11 @@ task pwm_scoreboard::process_tl_access(tl_seq_item   item,
             blink_cnt[ii] = blink[ii].X;
             blink_state[ii] = CycleA;
             int_dc[ii] = duty_cycle[ii].A;
-
-            // Alas we presently must ignore the first couple of items because the PWM output is
-            // enabled with an arbitrary phase relationship to the shared phase counter.
-            ignore_state_change[ii] = SettleTime;
           end
-          txt = $sformatf("\n Channel[%0d] : %0b", ii, channel_en[ii]);
+          // Alas we presently must ignore the first couple of items because the PWM configuration
+          // is set up with an arbitrary phase relationship to the shared phase counter.
+          ignore_state_change[ii] = SettleTime;
+          txt = $sformatf("\n Channel[%d] : %0b", ii, channel_en[ii]);
         end
         `uvm_info(`gfn, $sformatf("Setting channel enables %s ", txt), UVM_HIGH)
       end
--- a/hw/ip/pwm/dv/env/pwm_scoreboard.sv
+++ b/hw/ip/pwm/dv/env/pwm_scoreboard.sv
@@ -18,9 +18,15 @@ class pwm_scoreboard extends cip_base_scoreboard #(
   typedef enum bit { LargeA = 1'b0, LargeB = 1'b1} dc_mod_e;
 
   // Every time the state is changed either from A to B or B to A, the initial transaction checking
-  // goes out of sync between checker and DUT. So, ignore_state_change is changed to 2 (SettleTime).
-  // This gives checker two pulses buffer to sync to DUT pwm_o pulse accurately.
-  localparam int SettleTime = 2;
+  // goes out of sync between checker and DUT. So, ignore_state_change is changed to `SettleTime`.
+  //
+  // This gives the checker a window of four pulse cycles to sync to the DUT `pwm_o` pulse
+  // accurately.
+  // Because the monitors do not respond to the assertion of the channel enable but rather the phase
+  // counter, if the phase counter is varying rapidly (ClkDiv=0, dcResn=0) the first blink phase can
+  // be up to 4 pulse cycles out, but synchronization typically occurs when the blinking state
+  // changes.
+  localparam int SettleTime = 4;
 
   // global settings
   bit                        regwen           =  0;
@@ -29,6 +35,11 @@ class pwm_scoreboard extends cip_base_scoreboard #(
   bit [PWM_NUM_CHANNELS-1:0] channel_en       = '0;
   bit [PWM_NUM_CHANNELS-1:0] invert           = '0;
   bit [PWM_NUM_CHANNELS-1:0] first_activation = '1;
+  // Have we synchronized yet? Once synchronized we tolerate no further deviations.
+  bit [PWM_NUM_CHANNELS-1:0] synchronized     = '0;
+  // Remember the previous item from the monitor so that we can spot blink state transitions and
+  // synchronize exactly.
+  pwm_item prev_item[PWM_NUM_CHANNELS];
 
   state_e                    blink_state[PWM_NUM_CHANNELS] = '{default:CycleA};
   int                        blink_cnt[PWM_NUM_CHANNELS]   = '{default:0};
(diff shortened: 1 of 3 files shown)


hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -138,12 +138,11 @@ task pwm_scoreboard::process_tl_access(tl_seq_item   item,
...
             int_dc[ii] = duty_cycle[ii].A;
-
-            // Alas we presently must ignore the first couple of items because the PWM output is
-            // enabled with an arbitrary phase relationship to the shared phase counter.
-            ignore_state_change[ii] = SettleTime;
           end
-          txt = $sformatf("\n Channel[%0d] : %0b", ii, channel_en[ii]);
+          // Alas we presently must ignore the first couple of items because the PWM output is
+          // enabled with an arbitrary phase relationship to the shared phase counter.
+          ignore_state_change[ii] = SettleTime;
```
rswarbrick, line 164: I'm not sure I understand this properly. Why is it no longer on the posedge of the enable signal?

(Also, maybe it makes sense to split this commit out from the vseq change?)

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -18,9 +18,14 @@ class pwm_scoreboard extends cip_base_scoreboard #(
   typedef enum bit { LargeA = 1'b0, LargeB = 1'b1} dc_mod_e;
 
   // Every time the state is changed either from A to B or B to A, the initial transaction checking
-  // goes out of sync between checker and DUT. So, ignore_state_change is changed to 2 (SettleTime).
-  // This gives checker two pulses buffer to sync to DUT pwm_o pulse accurately.
-  localparam int SettleTime = 2;
+  // goes out of sync between checker and DUT. So, ignore_state_change is changed to `SettleTime`.
+  //
+  // This gives the checker four pulses buffer to sync to DUT pwm_o pulse accurately.
```
rswarbrick, line 23: I know this text was there before, but maybe "... a four pulses buffer ..." ? I started by imagining we were giving the checker four things.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -247,7 +273,43 @@ function void pwm_scoreboard::reset(string kind = "HARD");
   end
 endfunction
 
-task pwm_scoreboard::compare_trans(int channel);
+// We allow a little difference in the timing of the blink state transitions until we've
+// synchronized to this PWM output.
```
rswarbrick, line 277: Have you duplicated the comment intentionally? I think my approach would be to only put it on the declaration (but possibly put "internal" comments about how the task works here if necessary).

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -54,7 +64,17 @@ class pwm_scoreboard extends cip_base_scoreboard #(
 
   // Run forever, comparing items from the monitor on the given channel with the items that we
   // expect to be generated (based on the scoreboard's model of the config registers).
-  extern task compare_trans(int channel);
+  extern task compare_trans(int unsigned channel);
+
+  // We allow a little difference in the timing of the blink state transitions until we've
+  // synchronized to this PWM output.
```
rswarbrick, line 70: Could you expand this slightly to explain what the function is doing? I think it might be something like "Return true if we are not yet synchronised to the PWM output for this channel." ?

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -247,7 +273,43 @@ function void pwm_scoreboard::reset(string kind = "HARD");
...
+// Adjust channel state to synchronize with the monitor/DUT output.
+function void pwm_scoreboard::synchronize_blink_state(int unsigned channel);
+  `uvm_info(`gfn, $sformatf("Synchronizing on channel %0d (ignore %0d blink_cnt %0d)", channel,
+                            ignore_state_change[channel], blink_cnt[channel]), UVM_MEDIUM)
+  if (ignore_state_change[channel] > 0) begin
+    // Expectation is ahead of the monitor/DUT, simply rewind the pulse cycle counter within the
+    // current blinking phase.
+    `uvm_info(`gfn, $sformatf("Synchronizing blink state for channel %0d", channel), UVM_MEDIUM)
+    case (channel_param[channel].HtbtEn)
+      1'b0: begin
```
rswarbrick, line 293: Verible has done something annoying with its lint check here. I'd probably do the stupid thing and make this an `if / else`.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -247,7 +273,43 @@ function void pwm_scoreboard::reset(string kind = "HARD");
...
+      1'b1: begin
+        blink_cnt[channel] = blink[channel].X;
+      end
+    endcase
+  end else begin
+    // Expectation is behind, so we need to advance to the next phase.
+    advance_blink_state(channel);
+  end
+  synchronized[channel] = 1'b1;
+  // We have now synchronized, reset `ignore_state_change` for clarity only; it should be ignored
```
rswarbrick, line 307: Nit. I think that the text from "reset" is a new sentence?

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -315,6 +406,76 @@ task pwm_scoreboard::compare_trans(int channel);
   end
 endtask : compare_trans
 
+// Advance the blink state at the end of the 'BLINK_PARAM.X+1' or 'BLINK_PARAM.Y+1' pulse cycles.
+function void pwm_scoreboard::advance_blink_state(int unsigned channel);
+  `uvm_info(`gfn, $sformatf("Advancing blink state for channel %0d", channel), UVM_HIGH)
+  // Unique case for violation report
+  case (channel_param[channel].HtbtEn)
+    1'b0: begin
```
rswarbrick, line 414: Again, I think you'll need to write this with an `if / else`.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: The settle time is required for disabled channels too because the pwm_monitor is no longer aware of the enable state and operates on disabled channels too. This vseq happened to expose that oversight in the scoreboard and depends upon it for some seeds to pass.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: Yes, it's my practice to describe what the function/task does before both the declaration and the definition so that it's immediately available, and the internals within the body of the function/task. If I'm honest I question the utility of separating the declaration and definition/body when the two are present in a single source file like this. It seems just to lead to repetition and potential inconsistency (warnings squirrelled away in build logs).

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: Reworded.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: Done.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: Yes, that structure has been inherited and cloned, without the `default` clause although the `case` always struck me as being of questionable utility; it's selecting using a single two-state bit. Restructured.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
rswarbrick: Fair enough. I'm not completely sure I have an opinion about what's best :-)

rswarbrick:
Cool, thanks. Modulo inconsistent doc comments, I think this looks good.

alees24:
> Cool, thanks. Modulo inconsistent doc comments, I think this looks good.

I certainly didn't meant to imply that comments associated with the declaration and with the definition must always be identical; where there's a useful separation, eg. the declaration is in a public header file, it should be directed towards external users, but the body comment may contain internal/implementation detail.

In this case it's such a simple function, and one that's only for internal use.

alees24:
Documentation point addressed, and subsumed into PR #25827

PR #25818, opened 2025-01-08
[dv,pwm] Test protection of configuration registers

alees24:
Introduce a test that attempts to modify the PWM configuration after the outputs have been set operating and the REGWEN bit has been cleared.

Model the REGWEN functionality in the scoreboard.

Tidy the reset handling within the scoreboard as required to support this sequence, since clearing the REGWEN shall only be achievable by a reset of the DUT. Normal shutdown via the configuration registers is not possible.

**Only the final commit is relevant to this PR,** but it is presently based on PR #25815 which should be reviewed and merged first.

### DIFF
--- a/hw/ip/pwm/dv/env/pwm_scoreboard.sv
+++ b/hw/ip/pwm/dv/env/pwm_scoreboard.sv
@@ -143,7 +143,8 @@ task pwm_scoreboard::process_tl_access(tl_seq_item   item,
     `uvm_fatal(`gfn, $sformatf("Access unexpected addr 0x%0h", csr_addr))
   end
 
-  if (addr_phase_write) begin
+  // The REGWEN register controls write access to all of the DUT registers, including itself.
+  if (addr_phase_write && `gmv(ral.regwen.regwen)) begin
     // if incoming access is a write to a valid csr, then make updates right away
     void'(csr.predict(.value(item.a_data), .kind(UVM_PREDICT_WRITE), .be(item.a_mask)));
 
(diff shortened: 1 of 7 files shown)


hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -268,8 +269,10 @@ endtask
 
 function void pwm_scoreboard::reset(string kind = "HARD");
   super.reset(kind);
-  for (int i = 0; i < PWM_NUM_CHANNELS; i++) begin
-    item_fifo[i].flush();
+  for (int unsigned channel = 0; channel < PWM_NUM_CHANNELS; channel++) begin
+    `uvm_info(`gfn, $sformatf("Reset flushing channel %0d", channel), UVM_MEDIUM)
```
rswarbrick, line 273: That's reasonable, but I'd be tempted to put "Reset flushing all channels" outside of the loop, which would be less chatty.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
```diff
@@ -321,6 +324,11 @@ task pwm_scoreboard::compare_trans(int unsigned channel);
     if (!cfg.under_reset) begin
       bit matched;
       item_fifo[channel].get(input_item);
+      if (cfg.under_reset) begin
+        // In the event that `regwen` has been sent it is not possible to shut down the DUT cleanly
+        // because the register cannot be altered; the DUT must be reset whilst it is operating.
```
rswarbrick, line 329: I'm not sure about this comment. Isn't the point that we might have entered reset while we were waiting on the item fifo? (I don't think I can see a connection with regwen here)

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: I like the irony here. The PWM DV historically has not been nearly chatty enough, having to switch to UVM_HIGH even to see that it was doing anything, let alone mismatching items.

hw/ip/pwm/dv/env/pwm_scoreboard.sv
alees24: Fair enough, the reasoning chain was incomplete. We definitely enter reset at the end of the sequence, since we have no other option really.

rswarbrick: Ah, now I understand! Thanks, that makes sense now.

rswarbrick (approved):
This looks good, thanks. Good to merge once #25815 has landed.

PR #25827, opened 2025-01-09
[dv,pwm] Phase relationship test sequence

alees24:
This PR introduces a sequence to drive the PWM outputs in a known phase relationship, with three of the signals driven at 120-degree offsets and the second group of three driven as their anti-phase counterparts using the `invert` functionality.

PWMphase

The above capture shows channels 0-2 being driven with phases of 0,120 and 240, and then channels 3-5 being their respective complements. It shows the point of transition from duty cycle A (1/4) to duty cycle B (3/4) and how the channels switch configuration only at the point at which it is safe to change the duty cycle, given their configured phase delays.

Note: **Only the final commit is relevant,** this is built on PR #25815.


rswarbrick (approved):
I like this! Good to rebase and merge once #25815 has landed.

PR #25828, opened 2025-01-09
[top_darjeeling] Overwrite default PMP config from package definition

Razer6:
This PR overwrites the default PMP configuration of Ibex with the one defined in `ibex_pmp_reset_pkg`. This allows integrators to customize the PMP config of Darjeeling.


vogelpi (approved):
LGTM!

PR #25829, opened 2025-01-09
Catch URL error if remote stats bucket does not exist

Razer6:
When creating a new IP or moving an IP to ipgen, there is no initial regression run. Consequently, fetching its stats fails as the remote path does not exist.

To unblock this error, this PR catches the URL error if the path does not exist and returns that 0 runs out of 0 are performed.


jwnrt (approved):
I think this is the right way to handle this, thanks

HU90m (approved):

jwnrt:
FPGA failures unrelated

PR #25831, opened 2025-01-09
[hw,rv_dm] Fix assertion path for FSM error trigger

Razer6:
This fixes the hierarchy path of the protected FSM registers for the RV DM LC gate


rswarbrick:

There is no risk to this change: it's correcting a path in some code currently will not compile (in a configuration that is not in earlgrey)

vogelpi:

There is no risk to this change: it's correcting a path in some code currently will not compile (in a configuration that is not in earlgrey)

PR #25832, opened 2025-01-09
[ci] Check lock files in CI

jwnrt:
This script regenerates the Bazel, Cargo, and Python lock files and checks for a diff.

### DIFF
--- a/ci/scripts/check-lock-files.sh
+++ b/ci/scripts/check-lock-files.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+REPO_TOP="$(git rev-parse --show-toplevel)"
+
+cd "$REPO_TOP"
+
+fail() {
+  file="$1"
+  command="$2"
+
+  echo "::error::${file} is not up to date"             >&2
+  echo "Regenerate this lock file using \`${command}\`" >&2
+
+  exit 1
+}
+
+# Regenerate `MODULE.bazel.lock` file.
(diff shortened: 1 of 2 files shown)


ci/scripts/check-lock-files.sh
```diff
@@ -0,0 +1,48 @@
...
+  command="$2"
+
+  echo "::error::${file} is not up to date"             >&2
+  echo "Regenerate this lock file using \`${command}\`" >&2
+
+  exit 1
+}
+
+# Regenerate `MODULE.bazel.lock` file.
+bazel mod deps
```
pamaury, line 24: Out of curiosity, where in this pipeline is the `bazel` alias defined? I assume it's alised to `./bazelisk.sh`?

pamaury:
Stupid question: with this check, does that mean that in certain cases, if a new version of package is available, the check will fail because cargo/pip/bazel will resolve to a different version? Here I am thinking about requirements that only specify a major.minor version for example and a new patch appears upstream.

nbdd0121:
The cargo update command has `-w` specified to only touch workspace packages without updating their dependencies.

Pip is already checked in ci/scripts/check-generated.sh, so it should be moved/removed from this PR, and we know that it won't be affected by upstream package upgrade.

Not sure about Bazel deps though.

jwnrt:
Yes, I added `-w` to prevent Cargo changing versions. I'm not sure about the `gen-python-requirements.sh` script or Bazel though... I wonder if there's a way to test this.

Both of them come back as up to date and it seems likely there have been new versions since the last time they were changed, so maybe we can assume that they're okay?

ci/scripts/check-lock-files.sh
jwnrt: Oh, good point, I'm actually using the Bazel that comes with the CI container here. I'm not sure whether that properly respects `.bazelversion` so I'll fix this.

PR #25833, opened 2025-01-09
[rom_ctrl, dv] Conditional coverage hole in tlul_adapter_sram fifos

KinzaQamar:


rswarbrick (commented):
Hmm. I have just thought a bit about this and I'm reasonably convinced that this isn't really a fact about `rom_ctrl`. My reasoning:

- The `under_rst` signal is only true when under reset or for the first cycle after we come out of reset.
- The `empty` signal is the `empty_o` port of a `prim_fifo_sync_cnt` instance, which is true if a read and write pointer are equal.
- If `Secure` is false because `SecFifoPtr` is false (not the case for `rom_ctrl`), this cannot be true because the two pointers are both reset to `{WrapPtrW{1'b0}}`.
- If `Secure` is true (the case for `rom_ctrl`), these pointers are driven by the `cnt_o` ports of two `prim_count` instances. These ports are driven with `cnt_q[0]` in each instance.
- This is a bit fiddly to follow, but comes from `cnt_unforced_q`, which is the output of a `prim_flop` in each case, that gets reset to `ResetValues[0]`. This, in turn, equals the `ResetValue` parameter, which has its default value (0).
- As a result, we come out of reset with `empty = 0`, regardless of the value of the `Secure` parameter.

That means that I think the right change is to adapt the RTL in the following way:
- Change the definition of `rvalid_o` to `assign rvalid_o = ~empty;`
- Add an assertion that says `under_rst -> empty` to check that this doesn't change in the future.

rswarbrick (commented):
That looks like the right approach to me: thanks! I think we probably need to drop the exclusion though and move the reasoning to the commit message.

hw/ip/prim/rtl/prim_fifo_sync.sv
```diff
@@ -159,7 +159,7 @@ module prim_fifo_sync #(
     // The latter can be '0' when under reset, while the former is an indication that no more
     // entries can be written.
     assign wready_o = ~full_o & ~under_rst;
-    assign rvalid_o = ~empty & ~under_rst;
+    assign rvalid_o = ~empty;
```
a-will, line 162: This doesn't look right to me for passthrough FIFOs. prim_fifo_sync should not pass through a valid data indicator if it is under reset. Without this gate, `wvalid_i` leaks through.

a-will (changes requested):

hw/ip/prim/rtl/prim_fifo_sync.sv
KinzaQamar: Thanks for the review.

This is the definition of empty [here](https://github.com/lowRISC/opentitan/blob/257ba6f4435c43502209652773bef11a325be3a9/hw/ip/prim/rtl/prim_fifo_sync_cnt.sv#L61). If the fifo is under reset then both wptr_wrap_cnt_q and rptr_wrap_cnt_q gets reset to 0. Which means empty=1 under reset keeps rvalid_o low.

Does that seems right to you?

a-will: That is _not_ the definition of `empty` for Pass = 1. It is here:

https://github.com/lowRISC/opentitan/blob/257ba6f4435c43502209652773bef11a325be3a9/hw/ip/prim/rtl/prim_fifo_sync.sv#L196

So, for Pass = 1, if wvalid_i is 1, empty is 0, and rvalid_o now becomes 1 with the change here. That's invalid behavior.

With good structuring of resets, this shouldn't happen, but I think we would invite new bugs if we don't make our FIFOs have consistent outputs. This change should not be applied for Pass = 1.

rswarbrick: Oh, I see! I was working on the basis that `empty |-> fifo_empty`, but it is *not* true that `~empty |-> ~fifo_empty`. Oops.

Would it make sense to put this in an if/else generate block that depends on `Pass`? Something like:
```systemverilog
    // Splitting like this avoids looking at the under_rst signal when Pass is false. If there is no
    // pass-through, it is not possible for there to be anything available through the fifo (so
    // empty is true) if under_rst is true.
    if (Pass == 1'b1) begin : gen_pass_rvalid
      assign rvalid_o = ~empty & ~under_rst;
    end else begin : gen_nopass_rvalid
      assign rvalid_o = ~empty;
    end

    `ASSERT(OnlyRvalidWhenNotUnderRst_A, rvalid_o -> ~under_rst)
```
(but moving the assertion to the bottom)

a-will: > Oh, I see! I was working on the basis that `empty |-> fifo_empty`, but it is _not_ true that `~empty |-> ~fifo_empty`. Oops.
> 
> Would it make sense to put this in an if/else generate block that depends on `Pass`? Something like:
> 
> ```systemverilog
>     // Splitting like this avoids looking at the under_rst signal when Pass is false. If there is no
>     // pass-through, it is not possible for there to be anything available through the fifo (so
>     // empty is true) if under_rst is true.
>     if (Pass == 1'b1) begin : gen_pass_rvalid
>       assign rvalid_o = ~empty & ~under_rst;
>     end else begin : gen_nopass_rvalid
>       assign rvalid_o = ~empty;
>     end
> 
>     `ASSERT(OnlyRvalidWhenNotUnderRst_A, rvalid_o -> ~under_rst)
> ```
> 
> (but moving the assertion to the bottom)

Yup, that sounds good to me!

rswarbrick: Thanks! @KinzaQamar, would you mind submitting the tweaked version?

rswarbrick (commented):
This looks really sensible to me, thanks. @a-will, does this look reasonable to you?

a-will (approved):

rswarbrick (approved):
Thanks for the careful tidy-up.

rswarbrick:

This change will have no effect on behaviour (after @a-will and I spent quite a long time thinking about it!) but will close a conditional coverage hole and, presumably make the design marginally smaller. The added assertion ensures that we will see test failures if anything changes or we got the reasoning wrong.

a-will:

PR #25840, opened 2025-01-10
[rom_ext] Enable Oz to reduce firmware size

sasdf:
This change enables Oz optimization for rom_ext to reduce code size (43084B -> 41868B).

Since some targets can't be built with LLD, this commit introduces a new attribute, `with_features`, to the `opentitan_binary` rule to allow specifying additional features for the target.

LTO is also ready to be enabled for those targets, but we found no size improvement (41868B off v.s. 41896B on), so we keep it off at the moment.

### DIFF
--- a/rules/rv.bzl
+++ b/rules/rv.bzl
@@ -19,10 +19,11 @@ PER_DEVICE_DEPS = {
 }
 
 def _opentitan_transition_impl(settings, attr):
+    features = settings["//command_line_option:features"] + attr.extra_bazel_features
     return {
         "//command_line_option:platforms": attr.platform,
         "//command_line_option:copt": settings["//command_line_option:copt"],
-        "//command_line_option:features": settings["//command_line_option:features"],
+        "//command_line_option:features": features,
         "//hw/bitstream/universal:rom": "//hw/bitstream/universal:none",
         "//hw/bitstream/universal:otp": "//hw/bitstream/universal:none",
         "//hw/bitstream/universal:env": "//hw/bitstream/universal:none",
@@ -58,6 +59,8 @@ def rv_rule(**kwargs):
     attrs = kwargs.pop("attrs", {})
     if "platform" not in attrs:
         attrs["platform"] = attr.string(default = OPENTITAN_PLATFORM)
+    if "extra_bazel_features" not in attrs:
+        attrs["extra_bazel_features"] = attr.string_list(default = [])
     attrs["_allowlist_function_transition"] = attr.label(
         default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
     )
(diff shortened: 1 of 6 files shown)


jwnrt:
Hi @sasdf, just to let you know I’ve moved the toolchain from CRT into this repository but the `minsize` feature should still be there and work in the same way, you may just need to drop the CRT update. Thanks.

luismarques:
Thanks for the PR. Not to block anyone else from reviewing this, but I can have a look at this later this week. I'm curious how -Oz compares with strategically marking functions as noinline, etc. (I'll compare).

rules/rv.bzl
```diff
@@ -58,6 +59,8 @@ def rv_rule(**kwargs):
     attrs = kwargs.pop("attrs", {})
     if "platform" not in attrs:
         attrs["platform"] = attr.string(default = OPENTITAN_PLATFORM)
+    if "with_features" not in attrs:
```
timothytrippel, line 62: nit: perhaps "with_features" would be better described with "extra_cc_features" or something like that to indicate these are compiler toolchain related features?

timothytrippel (commented):
mostly LGTM pending some comments in addition to others. Thanks @sasdf ! looking forward to getting this in to free up some precious ROM_EXT space.

hw/top_englishbreakfast/ip/pinmux/data/autogen/pinmux.hjson
```diff
@@ -0,0 +1,1165 @@
+// Copyright lowRISC contributors (OpenTitan project).
```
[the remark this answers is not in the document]
sasdf: Thanks, removed. It seems to be a rebase error.

rules/rv.bzl
sasdf: Sure, changed to `extra_bazel_features` since it's not cc toolchain only.

timothytrippel (approved):

pamaury:
Since this PR, I have noticed some rather suspicious warnings when compiling:

> INFO: From Linking sw/device/tests/aon_timer_irq_test_silicon_creator.elf:
/proc/self/cwd/external/_main~lowrisc_rv32imcb_toolchain~lowrisc_rv32imcb_toolchain/bin/riscv32-unknown-elf-ld: bazel-out/k8-fastbuild-ST-1df456420242/bin/sw/device/tests/aon_timer_irq_test_silicon_creator.elf: section `.bss' can't be allocated in segment 2
LOAD: .data .bss

and

> INFO: From PostSigningAttach sw/device/tests/uart_tx_rx_test_fpga_cw310_sival.prod_key_0.prod_key_0.signed.bin:
[2025-01-17T12:34:48Z WARN  opentitanlib::image::image] ROM_EXT is larger than 64k. Link offsets may need recalculating.

Is this expected? Is that a sign of something being wrong? I am particularly confused about the linker message because it still seems to succeed.

sasdf:
I have the same warnings on commit `734d9c1e33`, which is one commit before of this PR, so I think they are not related to this.

sasdf:
FYI, It seems the -Oz / -Os flag order becomes incorrect again during crt toolchain migration, where -Oz introduced in this PR got overwritten by the default -Os. #25928 will solve this ordering issue by making Os/Oz mutually exclusive instead.

PR #25657, opened 2024-12-16
Fix the gpio_stress_all with random reset 

marcelocarvalhoLowRisc:
New fix for the stress_all_gpio testcase with random reset. Basically included new if(reset) checks into the sequences to skip the flow when a reset comes, and replaced the wait_clks task to wait_clks_or_rst. This makes the gpio test pass with a reasonable number of seeds. It's a temporally fix until we have the reset agent on place.

PR #25843, opened 2025-01-10
[hmac,dv] Allow reset w/o CSR accesses complete

martin-velay:
- This PR is a temporary fix to allow the reset to complete without waiting for complete CSR accesses. This is done by setting the flag can_reset_with_csr_accesses to 1 in the env_cfg file. Ideally, this should go away once the a better reset approach will be in place via a reset agent.
- The approach here is to try to exit ASAP from all the tasks to skip the CSR accesses.

### DIFF
--- a/hw/dv/sv/dv_utils/dv_macros.svh
+++ b/hw/dv/sv/dv_utils/dv_macros.svh
@@ -387,6 +387,26 @@
   end
 `endif
 
+// Wait for one of two statements but stop early if the EXIT statement completes.
+//
+// Example usage:
+//
+//    `DV_SPINWAIT_EXIT_MULTI(do_something_time_consuming();,
+//                      do_something_else_time_consuming();,
+//                      wait(stop_now_flag);,
+//                      "The stop flag was set when we were working")
+`ifndef DV_SPINWAIT_EXIT_MULTI
+`define DV_SPINWAIT_EXIT_MULTI(WAIT_1_, WAIT_2_, EXIT_, MSG_ = "exit condition occurred!", ID_ =`gfn) \
+  `DV_SPINWAIT_EXIT(fork begin \
+                      fork \
+                        begin WAIT_1_ end \
+                        begin WAIT_2_ end \
+                      join_any \
+                      disable fork; \
+                    end join, \
+                    EXIT_, MSG_, ID_)
+`endif
+
 // macro that waits for a given delay and then reports an error
 `ifndef DV_WAIT_TIMEOUT
 `define DV_WAIT_TIMEOUT(TIMEOUT_NS_, ID_  = `gfn, ERROR_MSG_ = "timeout occurred!", REPORT_FATAL_ = 1) \
@@ -407,6 +427,12 @@
 `ifndef DV_WAIT
 `define DV_WAIT(WAIT_COND_, MSG_ = "wait timeout occurred!", TIMEOUT_NS_ = default_spinwait_timeout_ns, ID_ =`gfn) \
   `DV_SPINWAIT(wait (WAIT_COND_);, MSG_, TIMEOUT_NS_, ID_)
+`endif
+
+// Wait for one of two statements, but exit early after a timeout
+`ifndef DV_WAIT_MULTI
+`define DV_WAIT_MULTI(WAIT_COND_1_, WAIT_COND_2_, MSG_ = "wait timeout occurred!", TIMEOUT_NS_ = default_spinwait_timeout_ns, ID_ =`gfn) \
+  `DV_SPINWAIT_EXIT_MULTI(WAIT_COND_1_, WAIT_COND_2_, `DV_WAIT_TIMEOUT(TIMEOUT_NS_, ID_, MSG_);, "", ID_)
 `endif
 
 // Control assertions in the DUT.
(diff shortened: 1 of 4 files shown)


rswarbrick (approved):
Hmm, I'm not sure that's quite going to work. I think that e.g. `sar_stop_and_continue` calls `csr_spinwait`. What happens if a reset appears when we're half-way through the call?

I think the fix is probably to make it so that each of the `sar_*` functions exit early if they see a reset, then exit early after they have completed.

martin-velay:
> Hmm, I'm not sure that's quite going to work. I think that e.g. `sar_stop_and_continue` calls `csr_spinwait`. What happens if a reset appears when we're half-way through the call?
> 
> I think the fix is probably to make it so that each of the `sar_*` functions exit early if they see a reset, then exit early after they have completed.

@rswarbrick 
If I am not wrong, the `disable fork` will kill all the descendants, which means all the sub-tasks/functions initiated by one of these process. Is it clearing your doubt?

rswarbrick:
I think that will run into the original problem that caused all of this "outstanding access" tracking in the first place. The problem is that killing a sequence in the middle of a CSR operation leaves things in an inconsistent state.

martin-velay:
> I think that will run into the original problem that caused all of this "outstanding access" tracking in the first place. The problem is that killing a sequence in the middle of a CSR operation leaves things in an inconsistent state.

@rswarbrick, this fix is not linked to the "outstanding access" issue. The error seen here was something like: `UVM_ERROR @ 540230991 ps: (hmac_scoreboard.sv:534) [uvm_test_top.env.scoreboard] Check failed item.d_data == csr.get_mirrored_value() (4 [0x4] vs 0 [0x0]) intr_state`. Now this error is gone, but as you said maybe this will cause some "outstanding access". But I am not going to mix it with this PR.

rswarbrick:
Sorry: what I mean is that the processes in the fork will start CSR operations. If a reset happens to appear in the intervening time then the `wait` will complete in `reset_thread` and the `disable fork` will kill a sequence that's doing a TL access. This will cause things to come unstuck in rather confusing ways, which is why `cip_base_vseq` (without my flag) tries quite hard to avoid issuing the reset and killing the sequence when there is any TL operation in flight.

My proposed solution is a way to "complete the operation quickly". A lame temporary approach would be something like a delay just after the `wait`, which would let any pending CSR operation abort... (It's a bit lame though! I'm not keen! :-D)

marcelocarvalhoLowRisc:
I think this is a similar issue we had faced in the GPIO, during some rnd_reset_stress_all test.
https://github.com/lowRISC/opentitan/pull/25657

hw/dv/sv/dv_utils/dv_macros.svh
```diff
@@ -387,6 +387,36 @@
...
+          EXIT_ \
+          if (MSG_ != "") begin \
+            `dv_info(MSG_, uvm_pkg::UVM_HIGH, ID_) \
+          end \
+        end \
+      join_any \
+      disable fork; \
+    end join \
+  end
+`endif
```
rswarbrick, line 418: I think this can use `DV_SPINWAIT_EXIT`? Completely untested, but can't it be something like this?
```systemverilog
`ifndef DV_SPINWAIT_EXIT_MULTI
`define DV_SPINWAIT_EXIT_MULTI(WAIT_1_, WAIT_2_, EXIT_, MSG_ = "exit condition occurred!", ID_ =`gfn) \
  `DV_SPINWAIT_EXIT(fork begin \
                      fork \
                        begin WAIT_1 end \
                        begin WAIT_2 end \
                      join_any \
                      disable fork; \
                    end join, \
                    EXIT_, MSG_, ID_)
`endif
```

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -278,14 +278,18 @@ endfunction : clear_wipe_secret
...
     csr_rd(.ptr(ral.digest[i]), .value(digest[i]));
     `uvm_info(`gfn, $sformatf("digest[%0d]=32'h%08x", i, digest[i]), UVM_MEDIUM)
   end
 endtask : csr_rd_digest
 
 // Write digest value
 task hmac_base_vseq::csr_wr_digest(bit [TL_DW-1:0] digest[16]);
-  foreach (digest[i]) csr_wr(.ptr(ral.digest[i]), .value(digest[i]));
+  foreach (digest[i]) begin
+    if (cfg.under_reset) break;   // Skip if a reset is ongoing...
```
rswarbrick, line 290: I think this probably wants to be after the `csr_wr` (since that will return instantly anyway)

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
```diff
@@ -304,6 +308,7 @@ endtask : csr_rd_key_length
 // key_length determines how much of the key array is relevant for the HMAC operation
 task hmac_base_vseq::wr_key(bit [TL_DW-1:0] key[]);
   foreach (key[i]) begin
+    if (cfg.under_reset) break;   // Skip if a reset is ongoing...
```
rswarbrick, line 311: It doesn't really matter, but is this change really needed? I would have expected this code to complete in zero time if we went into reset.

hw/dv/sv/dv_utils/dv_macros.svh
martin-velay: Yes, it should work either.

hw/ip/hmac/dv/env/seq_lib/hmac_base_vseq.sv
martin-velay: I prefer to keep it as I am not sure and as it's temporary, it should go anyway later

PR #25847, opened 2025-01-10
[bazel] Disable hybrid mode

jwnrt:
This PR completes the transition to Bzlmod by disabling Hybrid mode.

This is a mostly non-functional change but allows Bazel to operate more efficiently without having to support pre-bzlmod features. These include:

* No more `//external:*` pseudo-package of repositories. Use `bazel mod graph` and friends instead.
* No more global namespace of repositories.

It also gives us access to bzlmod-only features like `bazel vendor` for airgapping.

## bazel sync

Unfortunately, once hybrid mode is disabled we cannot use the `sync` command any more and must use the (almost identical) `fetch` command instead. There will be many contributors with `bazelisk.sh sync --configure` in their `post-checkout` Git hooks whose workflows will break.

We cannot just ask everybody to change their hooks because `git fetch --configure` does not work on Bazel 6.2.0 which the hook will run on for old branches like `earlgrey_1.0.0`.

To resolve this, I have included a PR which causes `bazelisk.sh` to replace the `sync` command with `fetch` on this branch to allow contributors to continue using `sync` in scripts. This only happens when run from a script like a Git hook to prevent breakages. When run interactively, `bazelisk.sh` prints a deprecation error and exits.


jwnrt:
I've changed the `bazelisk.sh` patch to print a deprecation error when run interactively and forward `bazel sync` to `bazel fetch` when run from a script to keep Git hooks working.

pamaury (approved):
Your commit messages are strange, they include a list of environment variables

jwnrt:
Oh dear. I was dumping `env` in a Git hook to see what variables are set and forgot to remove it before rebasing. Luckily there's nothing sensitive.

nbdd0121 (approved):

PR #25851, opened 2025-01-13
[hw,alert_handler,rtl] Only read unused LPGs when there are more than 1

Razer6:
If there is only 1 LPG, no need to read the unused one. The one LPGs is associated to all alerts.


rswarbrick (approved):
This looks good to me.

rswarbrick:
Outside of lint fixes, this code doesn't actually have any effect before or after the change, so the risk of damaging the chip is zero.

CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/alert_handler/rtl/alert_handler_lpg_ctrl.sv

vogelpi:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/alert_handler/rtl/alert_handler_lpg_ctrl.sv

This RTL change targets configurations with a single LPG only. It doesn't have any functional impact on Earlgrey. This is fine.

PR #25852, opened 2025-01-13
[ci] make CI `$PATH` equal for push & PR to cache nonhermetic builds 

nbdd0121:
Change of `PATH` will cause change to hash of nonhermetic builds, which
prevents Verilator build cached in master to be reused in pull request
CI.

It appears that gcloud CLI is the one making difference, so fake a PATH
without installing it (as it's additional ~0.5G to download).


jwnrt (approved):
I'd rather not have this hack, it seems likely to break without us knowing. A real solution would be to fake the `PATH` inside Bazel to a proxy directory containing symlinks to the tools we need.

That said, I think the hack is harmless so I'll approve.

I would rather somebody tried the real fix though or at least opened an issue for it.

AlexJones0 (approved):
Thanks a lot @nbdd0121, this should fix the Verilator test caching issues. 

I agree with what @jwnrt said above, and think in long term we should move to better delineating exactly which parts of our build need to be non-hermetic to avoid more fragile hacks like this.

PR #25853, opened 2025-01-13
[hw,rv_plic,rtl] Fix uniquified instance name in VLNVs

Razer6:
The `module_instance_name ` was missing in some VLNVs from the latest changes.


a-will (approved):
Thanks for fixing those! I must've messed up a rebase :sweat_smile:

rswarbrick (approved):

PR #25858, opened 2025-01-14
[hw,soc_dbg_ctrl,rtl] Fix dependencies, improve type robustness, default values

Razer6:
This PR just fixes some issues that showed up during integration:

* Add all dependencies to the decode module that is used out-of-tree
* Explicitly convert the logic from flops back to the MUBI types to avoid linting errors
* Provide a default valid CAT 3 policy used for DV

### DIFF
--- a/hw/ip/soc_dbg_ctrl/rtl/soc_dbg_ctrl_pkg.sv
+++ b/hw/ip/soc_dbg_ctrl/rtl/soc_dbg_ctrl_pkg.sv
@@ -43,6 +43,16 @@ package soc_dbg_ctrl_pkg;
     prim_mubi_pkg::mubi4_t relocked;
   } soc_dbg_policy_t;
 
+  // TODO(#25880): This parameter is intended for DV so should be moved into a DV package file once
+  // that exists.
+  //
+  // Valid CAT 3 policy for DV purposes
+  parameter soc_dbg_policy_t SOC_DBG_POLICY_CAT3_VALID = '{
+    valid:    prim_mubi_pkg::MuBi4True,
+    category: DbgCategory3,
+    relocked: prim_mubi_pkg::MuBi4False
+  };
+
   // Encoding generated with:
   // $ ./util/design/sparse-fsm-encode.py -d 3 -m 6 -n 6 \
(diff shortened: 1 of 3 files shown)


hw/ip/soc_dbg_ctrl/rtl/soc_dbg_ctrl_pkg.sv
```diff
@@ -43,6 +43,13 @@ package soc_dbg_ctrl_pkg;
     prim_mubi_pkg::mubi4_t relocked;
   } soc_dbg_policy_t;
 
+  // Valid CAT 3 policy for DV purposes
```
rswarbrick, line 46: This seems reasonable but (assuming this is only used for DV) could you leave a TODO note here attached to an issue for creating a DV environment for the block?

Razer6: Created here: https://github.com/lowRISC/opentitan/issues/25880

rswarbrick: Cool, thanks. Then we can link to it here. I'm imagining something like
```systemverilog
  // TODO(#25880): This parameter is intended for DV so should be moved into a DV package file once
  // that exists.
  //
  // Valid CAT 3 policy for DV purposes
  parameter soc_dbg_policy_t SOC_DBG_POLICY_CAT3_VALID = '{
    valid:    prim_mubi_pkg::MuBi4True,
    category: DbgCategory3,
    relocked: prim_mubi_pkg::MuBi4False
  };
```

rswarbrick (commented):
One last bit of the dance to do in order to link to the issue from the code, but this looks good to me otherwise.

hw/ip/soc_dbg_ctrl/rtl/soc_dbg_ctrl_pkg.sv
Razer6: Fair enough. Directly linked the issue!

rswarbrick (approved):
Thanks! This looks good to me.

PR #25859, opened 2025-01-14
Ignore the utf8 errors from log

BenBenderOT:
The UART may have unsupported chars from firmware. These would crash the orchestrator.


timothytrippel:
Nice find! Unfortunately, your commit message is not formatted correctly, so this is failing CI. However, I included your fix in #25919. So I think we can close this.

BenBenderOT:
Perfect. I'll close the PR then.

BenBenderOT:
Perfect. Thanks Tim.

PR #25861, opened 2025-01-14
[bazel] Disable sandboxing for disassembly actions

jwnrt:
It's not clear why sandboxing was disabled here but it seems unnecessary.


nbdd0121 (approved):

AlexJones0 (approved):
LGTM if CI doesn't disagree, thanks @jwnrt.

jwnrt:
I just noticed my commit title is wrong, it should have been *enable* sandboxing. My bad.

PR #25862, opened 2025-01-14
[top_darjeeling] Instantiate AC Range checks

Razer6:
This PR instantiates the AC ranges for a single egress port. All Egress ports are muxed to a single egress port before leaving Darjeeling. An additional external miscellaneous port is added to the system that is coming from the external uncore but muxed  with same mux and going through the AC ranges.


rswarbrick (commented):
Crikey, these patches are enormous! (I understand that the code is autogenerated, but it's still an awful lot!)

Am I right in thinking these commits are orthogonal? If so, maybe it makes sense to split up the PR?

Razer6:
I don't think splitting up makes sense. The one big commit instantiates the ac ranges on the HJSON (5 lines or so) and this triggers ipgen to copy over the IP. If we split it up, there would be inconsistent IPs

hw/top_darjeeling/ip/soc_proxy/rtl/bat.sv
a-will, line 1: A thought for the future: Perhaps we'll want to formally introduce a TL-UL to TL-UL bus bridge with base address translation in hw/ip/tlul at some point. Then we can extend topgen / tlgen and do traversals to understand reachability.

hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
```diff
@@ -60,10 +66,39 @@ module soc_proxy
...
+  tlul_socket_m1 #(
+    .M         ( TLUL_HOST_CNT         ),
+    .HReqPass  ( {TLUL_HOST_CNT{1'b1}} ),
+    .HRspPass  ( {TLUL_HOST_CNT{1'b1}} ),
+    .HReqDepth ( {TLUL_HOST_CNT{4'd2}} ),
+    .HRspDepth ( {TLUL_HOST_CNT{4'd2}} ),
+    .DReqPass  ( 1                     ),
+    .DRspPass  ( 1                     ),
+    .DReqDepth ( 4'd4                  ),
+    .DRspDepth ( 4'd4                  )
```
vogelpi, line 96: I remember you mentioning that for inside the AC range checks a pipeline register might be needed for timing reasons. I think this is currently not present (but I might have missed it!). Alternatively, you could here set both `DReqPass` and `DRspPass` to 0. The `DReqDepth` and `DRspDepth` may be pretty expensive. They should at least be 1 if Pass is 0 but higher values don't help timing if Pass is 1.

vogelpi (approved):
This looks good, a few nits and you may want to consider tuning the TLUL MUX in SoC Proxy for timing and area. But this can also happen later. I think this probably needs to be rebased after #25986 gets merge.d

vogelpi:
Oh and CI flags a failing whitespace check:
```
Checking whitespace on files changed since 90a9ea57711d6b953edb1c7b655a4a2502eab573
Would have fixed 1 files.
Fixing file: "hw/top_darjeeling/ip_autogen/ac_range_check/data/ac_range_check.hjson"
```

hw/top_darjeeling/ip/soc_proxy/rtl/soc_proxy.sv
Razer6: Thanks good it. I changed the pass parameter to 0 for now.

hw/top_darjeeling/data/top_darjeeling.hjson
```diff
@@ -1014,6 +1014,20 @@
       },
       attr: "ipgen"
     },
+    { name: "ac_range_check",
```
vogelpi, line 1017: @Razer6 ac_range_check is not actually found by the tools as the dependency is missing in the top_darjeeling.core file. I've tried to add it but there is an issue with the actual core file of ac_range_check.

Razer6: Fixed that.

vogelpi: Thanks for fixing this @Razer6 . I see `ac_range_check.sv` has quite some syntax errors. Can you please take a look?

vogelpi:
Thanks for cleaning up ac_range_check.sv @Razer6 !

util/topgen.py
```diff
@@ -562,23 +562,26 @@ def generate_otp_ctrl(topcfg: Dict[str, object], cfg_path: Path, out_path: Path,
...
-        "nr_ctn_uid_bits": 5,
-        "num_ranges": topcfg['ac_range_check']['num_ranges'],
-        "module_instance_name": ac_ranges['type']
+        'num_ranges': ac_ranges['ipgen_param']['num_ranges'],
+        'module_instance_name': ac_ranges['type']
     }
+    params.update(racl_params)
 
-    ipgen_render("ac_range_check", topname, params, out_path)
+    ipgen_render('ac_range_check', topname, params, out_path)
```
matutem, line 584: Can we not have quote vs double quotes discrepancies? The overwhelming number of strings are double quoted, and we have the policy of preserving the prevailing convention. You can see topcfg/validate.py is mostly single quotes.

util/topgen/templates/toplevel_racl_pkg.sv.tpl
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 ${gencmd}
+<% import textwrap %>\
 <% racl_role_vec_len = 2 ** racl_config['nr_role_bits'] %>\
 
 package top_racl_pkg;
```
matutem, line 8: Weird this doesn't get some uniquifying name. The name top_racl_pkg can become confusing in multi-top.

PR #25863, opened 2025-01-14
[bazel] improve bitstream/verilator build cacheability

nbdd0121:
A few changes to accomplish the goal:
* Reduce nonhermetic envs by removing PYTHONPATH:
  * `PYTHONPATH` might be historically needed for airgapped build but testing shows that it's not needed anymore.
* Bazel `hw/ip/*/BUILD`, `hw/ip_templates/*/BUILD.tpl`, `hw/top_earlgrey/ip/*/BUILD` files to use exclude dv and doc files from glob:
  * They're excluded by existing filters but they don't work as well as just letting Bazel exclude them from inputs.
  * To achieve this I had to remove a few dependencies on DV core files that are not actually used.
* `hw/ip_templates`  is excluded as it's used by topgen, which is outside Bazel
* `hw/top_{darjeeling,englishbreakfast}` are excluded from Bazel as they aren't needed for earlgrey build.


jwnrt (approved):
Bazel changes LGTM, can't comment on the DV changes

pamaury (approved):
LGTM, I can't comment on the DV change either.

hw/ip/prim/prim_ram_1p_scr.core
```diff
@@ -15,7 +15,6 @@ filesets:
       - lowrisc:prim:buf
       - lowrisc:prim:cipher
       - lowrisc:prim:util_get_scramble_params
-      - lowrisc:dv_verilator:memutil_dpi_scrambled
```
vogelpi, line 18: This one looks good to me. The corresponding code is only used inside `hw/ip/otbn/dv/memutil/otbn_memutil.h` which has an explicit dependency on this core.

pamaury (approved):
The bazel changes look good to me.

PR #25864, opened 2025-01-14
[rom_ctrl,dv] Stress sequence tidyup

rswarbrick:
There are several commits in this PR, with detailed individual commit messages. The original motivation was actually a hole in toggle coverage, but this has ended up tidying up a couple of other items


rswarbrick:
I've just pushed a version with all but the last commit and will merge that :-)

PR #25865, opened 2025-01-14
[bazel] Change airgap method to `bazel vendor`

jwnrt:
Bazel now has a first-class airgap preparation command when using bzlmod. This greatly simplifies the steps required to prepare the airgapped environment to three:

1. Download Bazel itself.
2. Copy the latest bitstream somewhere local.
3. Run `bazel vendor`.

Using the airgapped environment is also simpler since only one flag is needed: `--vendor_dir`. The other flags are for bitstreams.

### DIFF
--- a/util/prep-bazel-airgapped-build.sh
+++ b/util/prep-bazel-airgapped-build.sh
@@ -3,170 +3,78 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
+# This script prepares a directory that can be used inside an airgapped
+# environment to build any Bazel target in this project without a network
+# connection.
+#
+# This includes the correct Bazel binary, all external dependencies, and the
+# latest valid bitstream.
+
 set -euo pipefail
+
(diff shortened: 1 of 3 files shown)


util/prep-bazel-airgapped-build.sh
```diff
@@ -3,170 +3,78 @@
...
+# Create directories
+mkdir -p "$BAZEL_AIRGAPPED_DIR"
+mkdir -p "$BAZEL_BITSTREAM_DIR"
+mkdir -p "$BAZEL_VENDOR_DIR"
+
+echo "Downloading Bazel binary..."
+
+curl --silent --location \
+  "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-x86_64" \
+  --output "${BAZEL_AIRGAPPED_DIR}/bazel"
```
jwnrt, line 47: Maybe `bazelisk` should do this download so it can check the hash and we don't have to read `.bazelversion` ourselves.

PR #25866, opened 2025-01-14
[bazel] build elf2tab using its own Cargo.toml/Cargo.lock

nbdd0121:
This lets bzlmod reads elf2tab's own Cargo.toml/Cargo.lock file and create an index, so we can remove them from OT's Rust dependencies.

This is good for crate auditing reasons as elf2tab uses somewhat outdated dependencies.


pamaury (approved):

jwnrt (approved):
Looks fine, thanks

Even though the order doesn't matter it might be good to put the `use_repo(tock, "elf2tab")` above the `elf2tab_index` initialization. Just a nitpick though, only do this if you make other changes.

nbdd0121:
In this case it's circular, as the extension relies on elf2tab_index being present too, so I don't think the order matters.

PR #25868, opened 2025-01-14
GPIO straps verif

marcelocarvalhoLowRisc:
New sequence to verify the straps feature and some additional description of the behaviour of this feature in the register document.

### DIFF
--- a/hw/ip/gpio/data/gpio_testplan.hjson
+++ b/hw/ip/gpio/data/gpio_testplan.hjson
@@ -145,5 +145,21 @@
       stage: V2
       tests: ["gpio_stress_all"]
     }
+    {
+      name: straps_data
+      desc: '''Verify the straps data/valid ouput expected values based on the strap_en and gpio_i inputs:
+      - Drive gpio_i input with random values.
+      - Set strap_en high for at least one clock cycle.
+      - Read the registers hw_straps_data_in and hw_straps_data_in_valid.
+      - Check the data read and sampled_straps_o in the scoreboard.
+      - Drive gpio_o and check that has no impact on straps registers.
+      - Apply a reset and ensure the strap registers are cleared.
+      - Read straps registers after reset.
+      - Iterate few times through the same flow again with new random values. Several iterations
+        will be done using the stress_all virtual sequence.
+      '''
+      stage: V3
+      tests: ["gpio_rand_straps"]
+    }
   ]
 }
--- a/hw/ip/gpio/dv/env/gpio_env_cfg.sv
+++ b/hw/ip/gpio/dv/env/gpio_env_cfg.sv
@@ -13,11 +13,16 @@ class gpio_env_cfg extends cip_base_env_cfg #(
   rand bit pulldown_en;
   // gpio virtual interface
   gpio_vif gpio_vif;
+  // gpio straps interface
+  straps_vif straps_vif_inst;
 
   constraint pullup_pulldown_en_c {pullup_en ^ pulldown_en;}
 
   `uvm_object_utils(gpio_env_cfg)
-  `uvm_object_new
+
+  function new(string name = "gpio_env_cfg");
+    super.new(name);
+  endfunction
 
   virtual function void initialize(bit [TL_AW-1:0] csr_base_addr = '1);
     list_of_alerts = gpio_env_pkg::LIST_OF_ALERTS;
--- a/hw/ip/gpio/dv/env/gpio_scoreboard.sv
+++ b/hw/ip/gpio/dv/env/gpio_scoreboard.sv
@@ -31,6 +31,8 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
   //  (i) indicate that write to INTR_STATE register just happened, and
   // (ii) store information of which all interupt bits were cleared
   bit [TL_DW-1:0] cleared_intr_bits;
+  // Flag to indicate that the strap was triggered
+  bit first_strap_triggered;
 
   // mask are WO, store the values in scb
   uvm_reg_data_t masked_out_lower_mask;
@@ -40,7 +42,9 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
 
   `uvm_component_utils(gpio_scoreboard)
 
-  `uvm_component_new
+  function new (string name = "gpio_scoreboard", uvm_component parent = null);
+    super.new (name, parent);
+  endfunction
 
   // Function: build_phase
   function void build_phase(uvm_phase phase);
@@ -54,6 +58,7 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
     fork
       monitor_gpio_i();
       monitor_gpio_interrupt_pins();
+      monitor_gpio_straps();
     join_none
   endtask
 
@@ -211,7 +216,6 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
       end // if (write)
     end else begin // if (channel == DataChannel)
       if (write == 0) begin
-        `uvm_info(`gfn, $sformatf("csr read on %0s", csr.get_name()), UVM_HIGH)
         // If do_read_check, is set, then check mirrored_value against item.d_data
         if (do_read_check) begin
           // Checker-2: Check if reg read data matches expected value or not
@@ -359,6 +363,50 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
       end
     end
   endtask : monitor_gpio_interrupt_pins
+
+  virtual task update_gpio_straps_regs(logic [NUM_GPIOS-1:0] gpio_i_sampled);
+    // Update data_in and data_in_valid ral register value based on result of input
+    `DV_CHECK_FATAL(ral.hw_straps_data_in.predict(.value(gpio_i_sampled),
+                                                  .kind(UVM_PREDICT_READ)));
+    `DV_CHECK_FATAL(ral.hw_straps_data_in_valid.predict(.value('b1),
+                                                        .kind(UVM_PREDICT_READ)));
+  endtask : update_gpio_straps_regs
+
+  // Task: monitor_gpio_straps
+  // The task monitors the gpio straps enable signal
+  // and checks the straps output signal after the first strap trigger
+  virtual task monitor_gpio_straps();
+    logic [NUM_GPIOS-1:0] gpio_i_sampled;
+    forever begin : monitor_gpio_straps
+      // Wait for going out of reset operation.
+      wait(!cfg.under_reset);
+      // Wait until the strap_en input be triggered
+      // if a reset comes in the middle, step-out of the loop.
+      while (!cfg.straps_vif_inst.tb_port.strap_en) begin
+        cfg.clk_rst_vif.wait_clks_or_rst(1);
+        if (cfg.under_reset) break;
+      end
+      // Step out to the next iteration if a reset happens.
+      if (cfg.under_reset) continue;
+      // Get the gpio_i input data from the pins interface.
+      gpio_i_sampled = cfg.gpio_vif.pins;
+      // Wait for one clock cycle to update the register model.
+      cfg.clk_rst_vif.wait_clks_or_rst(1);
+      // Step out from the loop if a reset comes.
+      if (cfg.under_reset) continue;
+      // Update the register model.
+      update_gpio_straps_regs(gpio_i_sampled);
+
+      // Checker: Compare actual values of gpio pins with straps register.
+      // Check the register hw_straps_data_in against gpio_i pins
+      `DV_CHECK_CASE_EQ(gpio_i_sampled, cfg.straps_vif_inst.tb_port.sampled_straps.data)
+      // Check the register hw_straps_data_in_valid
+      `DV_CHECK_CASE_EQ('b1, cfg.straps_vif_inst.tb_port.sampled_straps.valid)
+
+      // Wait for the next reset, if it happens.
+      wait(cfg.under_reset);
+    end
+  endtask : monitor_gpio_straps
 
   // Function: actual_gpio_i_activity
   function bit actual_gpio_i_activity();
@@ -561,6 +609,7 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
           ((|gpio_i_driven === 1'b1) && (actual_gpio_i_activity() == 1))) begin
         `DV_CHECK_CASE_EQ(pred_val_gpio_pins, cfg.gpio_vif.pins)
       end
+
     end
 
   endfunction : gpio_predict_and_compare
@@ -804,6 +853,7 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
     last_intr_update_except_clearing = '0;
     last_intr_test_event = '0;
     cleared_intr_bits = '0;
+    first_strap_triggered = 0;
   endfunction
 
   // Function: check_phase
--- a/hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
+++ b/hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
@@ -1,0 +1,142 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Verify the straps data/valid ouput expected values based on the strap_en and gpio_in inputs:
+//    - Drive gpio_in input with random values.
+//    - Set the strap_en high for at least one clock cycle.
+//    - Read the registers hw_straps_data_in and hw_straps_data_in_valid.
+//    - The data read and sampled_straps_o will be checked in the scoreboard.
+//    - Drive the gpio_out to make sure that has no impact on straps registers.
+//    - Read to make sure that if does not affect the straps registers after drive the gpio_out.
+//    - Apply reset and make sure the strap registers are clean.
+//    - Read straps registers after reset.
+//    - Iterate again the same flow, with new random values.
+class gpio_rand_straps_vseq extends gpio_base_vseq;
+
+  `uvm_object_utils(gpio_rand_straps_vseq)
+
+  // gpio input to drive
+  rand bit [NUM_GPIOS-1:0] gpio_in;
+  // gpio output to program in register
+  rand bit [NUM_GPIOS-1:0] gpio_out;
+  // gpio output enable to program in register
+  rand bit [NUM_GPIOS-1:0] gpio_oe;
+
+  function new(string name = "gpio_rand_straps_vseq");
+    super.new(name);
+  endfunction
+
+  // Read hw_straps_data_in and hw_straps_data_in_valid and
+  // check they match the expected value in
+  // the scoreboard
+  task csr_strap_read();
+    fork
+      begin
+        uvm_status_e status;
+        ral.hw_straps_data_in.mirror(status, UVM_CHECK);
+        `DV_CHECK_EQ(status, UVM_IS_OK)
+        `uvm_info(`gfn, $sformatf("Read the data_in status = %0d", status), UVM_HIGH)
+      end
+      begin
+        uvm_status_e status;
+        ral.hw_straps_data_in_valid.mirror(status, UVM_CHECK);
+        `DV_CHECK_EQ(status, UVM_IS_OK)
+        `uvm_info(`gfn, $sformatf("Read the data valid status = %0d", status), UVM_HIGH)
+      end
+    join
+  endtask : csr_strap_read
+
+  task test_straps_gpio_in();
+    // Drive the gpio_in
+    drive_gpio_in(gpio_in);
+
+    // Random wait to drive the strap_en
+    // The strap feature should work from zero clock cycles
+    // after driving the gpio_i inputs into the interface.
+    short_wait(0);
+
+    // Trigger the snapshot of gpio_in to be stored in the straps registers
+    cfg.straps_vif_inst.tb_port.strap_en = 1;
+    // Wait at least one clock cycle to update the strap register values.
+    short_wait(1);
+
+    // Read the hw_straps_data_in registers and check the expected value in the scoreboard
+    csr_strap_read();
+
+    // Random wait
+    short_wait(0);
+
+    // Stop driving gpio_in to make sure that, this is not affecting the strap_en registers
+    // so it will keep the same values were stored before.
+    undrive_gpio_in();
+
+    // Random wait
+    short_wait(0);
+
+    // Read to make sure that if does not affect the straps registers after undrive the gpio_in
+    csr_strap_read();
+
+  endtask : test_straps_gpio_in
+
+  task test_straps_gpio_out();
+
+    // Additional verification
+    // Drive the gpio_out to make sure that has no impact on straps registers.
+    // then read the gpio strap registers again
+    cfg.gpio_vif.drive_en('0);
+    csr_wr(ral.direct_out, gpio_out);
+    csr_wr(ral.direct_oe, gpio_oe);
+
+    // Random wait
+    short_wait(0);
+
+    // Read to make sure that if does not affect the straps registers after drive the gpio_out
+    csr_strap_read();
+
+  endtask : test_straps_gpio_out
+
+  // This task start the strap en test. First it will test the strap enable
+  // with the driven gpio_i. On a second step drive the gpio_out and check the strap
+  // registers based on that. And finally applies a second reset and check if the strap registers
+  // are reseted.
+  task start_test();
+
+    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_in)
+    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_out)
+    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_oe)
+
+    // User case to test the straps output, with gpio_in data randomised
+    test_straps_gpio_in();
+
+    // User case to test the straps output/registers, with gpio_out data randomised
+    // The gpio_out should not affect the straps output/registers.
+    test_straps_gpio_out();
+
+    // Random wait
+    short_wait(0);
+
+    // Set zero to the strap_en input.
+    cfg.straps_vif_inst.tb_port.strap_en = 0;
+    // Apply reset and make sure the strap registers are clean
+    apply_reset();
+
+    // Random wait
+    // At least one clock cycle required to take the updated register values.
+    short_wait(1);
+
+    // Read the straps registers after reset
+    csr_strap_read();
+
+  endtask : start_test
+
+  task body();
+    `uvm_info(`gfn, "Starting the test", UVM_HIGH)
+    // Just a minimum one clock cycle wait between more than one iteration if this
+    // test is executed into the stress_all virtual sequence.
+    short_wait(1);
+    start_test();
+    `uvm_info(`gfn, "End of the test", UVM_HIGH)
+  endtask : body
+
+endclass : gpio_rand_straps_vseq
--- a/hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
+++ b/hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
@@ -1,0 +1,21 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+// interface : gpio_straps_if
+`ifndef GPIO_STRAPS_IF_SV
+`define GPIO_STRAPS_IF_SV
+
+import gpio_pkg::*;
+
+// Interface definition
+interface gpio_straps_if(input logic clk, input logic rst_n);
+
+  logic          strap_en;       // Signal to enable straps
+  gpio_straps_t  sampled_straps; // Sampled gpio_i snapshot data from GPIO (DUT)
+
+  modport dut_port(input strap_en, output sampled_straps);
+  modport tb_port(output strap_en, input sampled_straps);
+
+endinterface
+
+`endif // GPIO_STRAPS_IF_SV
--- a/hw/ip/gpio/dv/tb/tb.sv
+++ b/hw/ip/gpio/dv/tb/tb.sv
@@ -11,8 +11,6 @@ module tb;
   import gpio_env_pkg::*;
   import gpio_test_pkg::*;
   import gpio_reg_pkg::*;
-  import gpio_pkg::*;
-
   // macro includes
   `include "uvm_macros.svh"
   `include "dv_macros.svh"
@@ -24,6 +22,8 @@ module tb;
   wire [NUM_GPIOS-1:0] gpio_oe;
   wire [NUM_GPIOS-1:0] gpio_intr;
   wire [NUM_MAX_INTERRUPTS-1:0] interrupts;
+  gpio_straps_t sampled_straps;
+  wire strap_en;
 
   `DV_ALERT_IF_CONNECT()
 
@@ -33,6 +33,10 @@ module tb;
     .rst_n(rst_n)
   );
   pins_if #(NUM_MAX_INTERRUPTS) intr_if (.pins(interrupts));
+  gpio_straps_if straps_if_inst (
+    .clk  (clk),
+    .rst_n(rst_n)
+  );
   tl_if tl_if (
     .clk  (clk),
     .rst_n(rst_n)
@@ -52,9 +56,8 @@ module tb;
     .clk_i (clk),
     .rst_ni(rst_n),
 
-    // TODO: need to test snapshot functionality
-    .strap_en_i(1'b0),
-    .sampled_straps_o(),
+    .strap_en_i(straps_if_inst.strap_en),
+    .sampled_straps_o(straps_if_inst.sampled_straps),
 
     .tl_i(tl_if.h2d),
     .tl_o(tl_if.d2h),
@@ -85,6 +88,7 @@ module tb;
     uvm_config_db#(intr_vif)::set(null, "*.env", "intr_vif", intr_if);
     uvm_config_db#(virtual tl_if)::set(null, "*.env.m_tl_agent*", "vif", tl_if);
     uvm_config_db#(virtual pins_if #(NUM_GPIOS))::set(null, "*.env", "gpio_vif", gpio_if);
+    uvm_config_db#(virtual gpio_straps_if)::set(null, "*.*", "straps_vif", straps_if_inst);
     $timeformat(-12, 0, " ps", 12);
     run_test();
   end
--- a/hw/ip/gpio/dv/tests/gpio_base_test.sv
+++ b/hw/ip/gpio/dv/tests/gpio_base_test.sv
@@ -1,16 +1,29 @@
 // Copyright lowRISC contributors (OpenTitan project).
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
-
 class gpio_base_test extends cip_base_test #(
   .ENV_T(gpio_env),
   .CFG_T(gpio_env_cfg)
 );
   `uvm_component_utils(gpio_base_test)
-  `uvm_component_new
+
+  straps_vif straps_vif_inst; // Virtual interface
+
+  function new(string name, uvm_component parent);
+    super.new(name, parent);
+  endfunction
+
+  task reset_phase(uvm_phase phase);
+    phase.raise_objection(this);
+    // Initialize inputs
+    straps_vif_inst.tb_port.strap_en = 0;
+    phase.drop_objection(this);
+  endtask
 
   virtual function void build_phase(uvm_phase phase);
     super.build_phase(phase);
+    if (!uvm_config_db#(straps_vif)::get(this, "*.*", "straps_vif", straps_vif_inst)) begin
+      `uvm_fatal("SEQ", "Virtual interface straps_vif_inst is not set")
+    end
   endfunction : build_phase
-
 endclass : gpio_base_test
--- a/hw/ip/gpio/dv/env/gpio_env.core
+++ b/hw/ip/gpio/dv/env/gpio_env.core
@@ -28,6 +28,7 @@ filesets:
       - seq_lib/gpio_stress_all_vseq.sv: {is_include_file: true}
       - seq_lib/gpio_intr_rand_pgm_vseq.sv: {is_include_file: true}
       - seq_lib/gpio_intr_with_filter_rand_intr_event_vseq.sv: {is_include_file: true}
+      - seq_lib/gpio_rand_straps_vseq.sv : {is_include_file: true}
     file_type: systemVerilogSource
 
 generate:
--- a/hw/ip/gpio/dv/env/gpio_env.sv
+++ b/hw/ip/gpio/dv/env/gpio_env.sv
@@ -17,6 +17,9 @@ class gpio_env extends cip_base_env #(
     if (!uvm_config_db#(gpio_vif)::get(this, "", "gpio_vif", cfg.gpio_vif)) begin
       `uvm_fatal(get_full_name(), "failed to get gpio_vif from uvm_config_db")
     end
+    if (!uvm_config_db#(straps_vif)::get(this, "", "straps_vif", cfg.straps_vif_inst)) begin
+      `uvm_fatal(get_full_name(), "Virtual interface straps_vif_inst is not set")
+    end
   endfunction
 
 endclass
--- a/hw/ip/gpio/dv/env/gpio_env_pkg.sv
+++ b/hw/ip/gpio/dv/env/gpio_env_pkg.sv
@@ -27,6 +27,7 @@ package gpio_env_pkg;
   parameter string LIST_OF_ALERTS[] = {"fatal_fault"};
 
   typedef virtual pins_if #(NUM_GPIOS) gpio_vif;
+  typedef virtual gpio_straps_if straps_vif;
   typedef class gpio_env_cfg;
   typedef class gpio_env_cov;
   typedef cip_base_virtual_sequencer #(gpio_env_cfg, gpio_env_cov) gpio_virtual_sequencer;
--- a/hw/ip/gpio/dv/env/seq_lib/gpio_base_vseq.sv
+++ b/hw/ip/gpio/dv/env/seq_lib/gpio_base_vseq.sv
@@ -124,4 +124,15 @@ class gpio_base_vseq extends cip_base_vseq #(
     end
   endtask : pgm_intr_regs
 
+  // Wait a few cycles. If force_positive is true, Wait at least one clock cycle.
+  task short_wait(bit force_positive);
+    int unsigned delay;
+    `DV_CHECK_FATAL(std::randomize(delay) with
+                    {
+                      delay dist {0 :/ 20, [1:5] :/ 40, [6:15] :/ 30, [20:25] :/ 10};
+                      force_positive -> delay > 0;
+                    })
+    cfg.clk_rst_vif.wait_clks(delay);
+  endtask
+
 endclass : gpio_base_vseq
--- a/hw/ip/gpio/dv/env/seq_lib/gpio_vseq_list.sv
+++ b/hw/ip/gpio/dv/env/seq_lib/gpio_vseq_list.sv
@@ -14,3 +14,4 @@
 `include "gpio_random_long_reg_writes_reg_reads_vseq.sv"
 `include "gpio_full_random_vseq.sv"
 `include "gpio_stress_all_vseq.sv"
+`include "gpio_rand_straps_vseq.sv"
--- a/hw/ip/gpio/dv/gpio_sim.core
+++ b/hw/ip/gpio/dv/gpio_sim.core
@@ -13,6 +13,7 @@ filesets:
     depend:
       - lowrisc:dv:gpio_test
       - lowrisc:dv:gpio_sva
+      - lowrisc:dv:gpio_if
     files:
       - tb/tb.sv
     file_type: systemVerilogSource
--- a/hw/ip/gpio/dv/gpio_sim_cfg.hjson
+++ b/hw/ip/gpio/dv/gpio_sim_cfg.hjson
@@ -180,6 +180,10 @@
       build_mode: en_cdc_prims
       run_opts: ["+no_pullup_pulldown=1"]
     }
+    {
+      name: gpio_rand_straps
+      uvm_test_seq: gpio_rand_straps_vseq
+    }
   ]
 
   // List of regressions.
--- a/hw/ip/gpio/dv/interfaces/gpio_if.core
+++ b/hw/ip/gpio/dv/interfaces/gpio_if.core
@@ -1,0 +1,18 @@
+CAPI=2:
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+name: "lowrisc:dv:gpio_if:0.1"
+description: "GPIO Interfaces"
+filesets:
+  files_dv:
+    depend:
+      - lowrisc:ip:gpio:0.1
+    files:
+      - gpio_straps_if.sv
+    file_type: systemVerilogSource
+
+targets:
+  default:
+    filesets:
+      - files_dv


hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,102 @@
...
+    for (uint tr_num = 0; tr_num < num_trans; tr_num++) begin
+      string msg_id = {`gfn, $sformatf(" Transaction-%0d", tr_num)};
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(delay)
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_i)
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_o)
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_oe)
+
+      cfg.clk_rst_vif.wait_clks(delay);
+      `uvm_info(msg_id, $sformatf("delay = %0d", delay), UVM_HIGH)
+      // Step-2 Drive the gpio_i
```
martin-velay, line 52: You should maybe add "Step-1" somewhere

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,102 @@
...
+      drive_gpio_in(gpio_i);
+
+      // If is the first strap triggered, it should be the gpio_i input, for the following interations
+      // keep the same value as the first strap already happened. No more updates in gpio_i should be take
+      // into consideration for the strap_data_in register.
+      exp_strap_data_in = straps_triggered ? rd_hw_straps_data_in : gpio_i;
```
martin-velay, line 58: Usually in DV the complete form with `if` is preferred

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,102 @@
...
+
+      // Set the expected straps_data_in with the gpio_i
+      ral.hw_straps_data_in.set(exp_strap_data_in);
+      // Update data_in ral register value based on result of input
+      void'(ral.hw_straps_data_in.predict(.value(exp_strap_data_in), .kind(UVM_PREDICT_DIRECT)));
```
martin-velay, line 63: I am surprised that we need to manually update the predicted value. Is there a particular reason here?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,102 @@
...
+      // Set the expected straps_data_in_valid
+      ral.hw_straps_data_in_valid.set('b1);
+      // Update data_in register value based on result of input
+      void'(ral.hw_straps_data_in_valid.predict(.value('b1), .kind(UVM_PREDICT_DIRECT)));
```
martin-velay, line 67: Same than above

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,102 @@
...
+
+      // Wait at least one clock cycle
+      `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(delay, delay >= 1;)
+
+      // Trigger the snapshot of gpio_i to be stored in the straps registers
+      cfg.straps_vif_inst.strap_en = 'b1;
```
martin-velay, line 73: Nit: all these `'b` are useless for single bit (IMO)

hw/ip/gpio/doc/registers.md
```diff
@@ -1,31 +1,35 @@
 # Registers
```
martin-velay, line 1: If I am not wrong, this file is generated based on the file `data/gpio.hjson`, but this file hasn't been pushed in this PR

martin-velay (changes requested):
I think this PR still requires some changes.

hw/ip/gpio/dv/tb/tb.sv
```diff
@@ -53,8 +55,8 @@ module tb;
     .rst_ni(rst_n),
 
     // TODO: need to test snapshot functionality
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: I can remove the TODO, the snapshot functionality is actually the straps featre I am testing in this PR.

hw/ip/gpio/dv/tb/tb.sv
```diff
@@ -85,6 +87,7 @@ module tb;
     uvm_config_db#(intr_vif)::set(null, "*.env", "intr_vif", intr_if);
     uvm_config_db#(virtual tl_if)::set(null, "*.env.m_tl_agent*", "vif", tl_if);
     uvm_config_db#(virtual pins_if #(NUM_GPIOS))::set(null, "*.env", "gpio_vif", gpio_if);
+    uvm_config_db#(virtual gpio_straps_if)::set(null, "*.*", "straps_if_inst", straps_if_inst);
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: very true! I miss that, I will fix it.

hw/ip/gpio/dv/env/gpio_env_cfg.sv
```diff
@@ -13,6 +13,10 @@ class gpio_env_cfg extends cip_base_env_cfg #(
   rand bit pulldown_en;
   // gpio virtual interface
   gpio_vif gpio_vif;
+  // gpio straps interface
+  straps_vif straps_vif_inst;
+  // Flag to indicate that the strap was trigered
+  bit first_strap_trigged;
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: yes it is only used in the scoreboard. it seems reasonable to move it there.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +365,35 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
     end
   endtask : monitor_gpio_interrupt_pins
 
+  // Task: monitor_gpio_straps
+  virtual task monitor_gpio_straps();
+    forever begin : monitor_gpio_intr
+      @(posedge cfg.clk_rst_vif.clk) begin
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: fixed, thnks

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: The scoreboard needs a fix to include this register in the predict function that is already there.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: Thanks for dropping that. Honestly, I think I generally use `1'b1` but I'm happy either way. But I think there's still a bare `'b1` a few lines later. Maybe switch that to `1` too?

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -478,6 +519,16 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+        "hw_straps_data_in_valid": begin
+          straps_data_in_valid = csr.get_mirrored_value();
+          `uvm_info(`gfn, $sformatf("straps_data_in_valid updated to 0x%0h", straps_data_in_valid),
+                    UVM_HIGH)
+        end
+        "hw_straps_data_in": begin
+          straps_data_in = csr.get_mirrored_value();
+          `uvm_info(`gfn, $sformatf("straps_data_in updated to 0x%0h", straps_data_in),
+                    UVM_HIGH)
+        end
```
rswarbrick, line 531: I'm not completely sure what this function is actually supposed to do, but I'm pretty certain this isn't doing anything useful :-)

Maybe you could spend a few minutes figuring out what on earth this function is supposed to be doing, then file a PR that adds a documentation comment explaining it?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// class : gpio_rand_straps_vseq
```
rswarbrick, line 5: Probably drop this magic line.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+// This gpio random test sequence performs following in each of multiple iterations:
+// - Drives gpio_i input with random values
+// - Set the expected strap_data_in based on the condition if is the first triggered strap_en or not
```
rswarbrick, line 8: Nit: I'm not sure how much it matters, but I'd probably be consistent on the tense here. Either "Drive" and "Set" or "Drives" and "Sets" (and similarly on the lines below)

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+// - Updates the RAL model with the expected strap_data_in and strap_data_in_valid
+// - Trigger the strap_en with one clock cycles
```
rswarbrick, line 10: I think I said this on the testplan, but this can probably be more general. The single clock cycle isn't needed, I don't think.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+// - Read the registers hw_straps_data_in and hw_straps_data_in_valid
+// - Check the expected behavior in the scoreboard and also on csr checks
```
rswarbrick, line 12: This isn't done by the sequence. Probably just drop the line.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+class gpio_rand_straps_vseq extends gpio_base_vseq;
+
+  `uvm_object_utils(gpio_rand_straps_vseq)
+
+  // gpio input to drive
+  rand bit [NUM_GPIOS-1:0] gpio_i;
+  // gpio output to program in register
+  rand bit [NUM_GPIOS-1:0] gpio_o;
```
rswarbrick, line 20: These names are a bit odd looking because of the `_i` and `_o` convention in the codebase. Maybe `_in` and `_out` ?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+  // gpio output enable to program in register
+  rand bit [NUM_GPIOS-1:0] gpio_oe;
+  // Flag to indicate that the first trigger straps happened
+  bit straps_triggered;
```
rswarbrick, line 24: I'd drop this bit and use `tr_num > 0` more explicitly in the body.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+  // Read straps_data_in
+  bit [NUM_GPIOS-1:0] rd_hw_straps_data_in;
+  // Read straps_data_in valid
+  bit                 rd_hw_straps_data_in_valid;
```
rswarbrick, line 28: I think this is basically duplicating the existing RAL model. Can't we use that instead?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+  // Expected strap data_in
+  bit [NUM_GPIOS-1:0] exp_strap_data_in;
```
rswarbrick, line 30: I believe this is dead code.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+  constraint num_trans_c {
+    num_trans inside {[20:200]};
+  }
```
rswarbrick, line 34: I think 3 is probably enough! Don't we expect to sample something on the first iteration and then just want to check it doesn't change? I'd probably force it to be exactly three.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+  task body();
+    `uvm_info(`gfn, $sformatf("num_trans = %0d", num_trans), UVM_HIGH)
+
+    for (uint tr_num = 0; tr_num < num_trans; tr_num++) begin
+      string msg_id = {`gfn, $sformatf(" Transaction-%0d", tr_num)};
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(delay)
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_i)
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_o)
+      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_oe)
```
rswarbrick, line 48: As a general question, why are these class variables? Can't they be local to the task?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+  constraint num_trans_c {
+    num_trans inside {[20:200]};
+  }
+
+  function new(string name = "gpio_rand_straps_vseq");
+    super.new(name);
+  endfunction
+
+  task body();
```
rswarbrick, line 40: I think this task could probably be simplified a bit by pulling out the body of the loop. Maybe something like this?
```systemverilog
  task check_transaction(string txn_desc, bit is_first);
    string msg_id = {`gfn, txn_desc};

    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_i)
    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_o)
    `DV_CHECK_MEMBER_RANDOMIZE_FATAL(gpio_oe)

    // Fill this in!
  endtask

  task body();
    `uvm_info(`gfn, $sformatf("num_trans = %0d", num_trans), UVM_HIGH)

    for (uint tr_num = 0; tr_num < num_trans; tr_num++) begin
      string txn_desc = $sformatf(" Transaction-%0d", tr_num);

      `DV_CHECK_MEMBER_RANDOMIZE_FATAL(delay)
      `uvm_info({`gfn, txn_desc}, $sformatf("delay = %0d", delay), UVM_HIGH)
      cfg.clk_rst_vif.wait_clks(delay);

      check_transaction(txn_desc, tr_num == 0);
    end
  endtask
```

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+      // If is not the first strap triggered, it should be the gpio_i input (that is the first one after reset),
+      // for the following interations keep the same value as the first strap already happened.
+      // No more updates in gpio_i should be taken
+      // into consideration for the strap_data_in register.
+      if(straps_triggered) begin
+        exp_strap_data_in = rd_hw_straps_data_in;
+      end else begin
+        exp_strap_data_in = gpio_i;
+      end
```
rswarbrick, line 63: This doesn't really belong in the sequence. We should be modelling the register value in the scoreboard, I think.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+      // Wait at least one clock cycle
+      `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(delay, delay >= 1;)
```
rswarbrick, line 66: I don't think this is what you expect it to do. It looks like the `delay` class member is supposed to be between transactions rather than in the middle of them.

And this code is just randomising the variable, not using it to wait.

I also can't see any reason to use a class member or indeed to force a nonzero delay here. And I think we should wait :-)

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+      // Step-2 Trigger the snapshot of gpio_i to be stored in the straps registers
+      cfg.straps_vif_inst.strap_en = 1;
+      cfg.clk_rst_vif.wait_clks(1);
+      cfg.straps_vif_inst.strap_en = 0;
```
rswarbrick, line 71: I really think this should be parallel and not necessarily a single cycle. I'd probably run a "gpio driver" task that drives random junk on the input pins and (in parallel) run "strap_en driver" task that sends pulses of 1s on `strap_en`.

These can both be done in parallel to the stuff here (and will be monitored in the scoreboard)

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+      // Step-3 Read the HW_STRAPS_DATA_IN register
+      // Reading straps data_in will trigger a check inside scoreboard
+
+      // Wait at least one clock cycle
+      `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(delay, delay >= 1;)
+
+      // Step-4 Read the hw_straps_data_in and check the expected value in the scoreboard
+      csr_rd(.ptr(ral.hw_straps_data_in), .value(rd_hw_straps_data_in));
+      // Read the hw_straps_data_in_valid and check the expected value in the scoreboard
+      csr_rd(.ptr(ral.hw_straps_data_in_valid), .value(rd_hw_straps_data_in_valid));
```
rswarbrick, line 83: I realise that `csr_rd` expects a `.value(...)` output argument, but that's just a sign that it was designed for directed tests :-)

Can't this use the `mirror` task? Something like this, maybe.
```systemverilog
  // Read the DATA_IN and DATA_IN_VALID registers
  //
  // The scoreboard will check that their values match the expected values sampled when strap_en_i
  // was first enabled.
  task check_straps_regs();
    uvm_state_e status;

    ral.hw_straps_data_in.mirror(.status(status), .check(UVM_CHECK));
    `DV_CHECK_EQ(status, UVM_IS_OK)

    ral.hw_straps_data_in_valid.mirror(.status(status), .check(UVM_CHECK));
    `DV_CHECK_EQ(status, UVM_IS_OK)
  endtask
```

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+      // Stop driving gpio_i
+      undrive_gpio_in();
+
+      // Step-5 Read to make sure that if does not affect the straps registers after undrive the gpio_in
+      csr_rd(.ptr(ral.hw_straps_data_in), .value(rd_hw_straps_data_in));
+      csr_rd(.ptr(ral.hw_straps_data_in_valid), .value(rd_hw_straps_data_in_valid));
```
rswarbrick, line 90: And of course this could just call the `check_straps_regs` task again here.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+
+      // Additional verification
+      // Drive the gpio_o to make sure that has no impact on straps registers.
+      // test gpio outputs
+      cfg.gpio_vif.drive_en('0);
+
+      ral.direct_out.set(gpio_o);
+      ral.direct_oe.set(gpio_oe);
+      csr_update(.csr(ral.direct_out));
+      csr_update(.csr(ral.direct_oe));
```
rswarbrick, line 100: This is something that I'd suggest running in a parallel task. Also, can't this use the `update` task directly (rather than relying on the magic CSR tasks) ?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,121 @@
...
+      // Step-6 Drive again the strap_en
+      cfg.straps_vif_inst.strap_en = 1;
+      cfg.clk_rst_vif.wait_clks(1);
+      cfg.straps_vif_inst.strap_en = 0;
+
+      `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(delay, delay >= 1;)
+
+      // Step-7 Read to make sure that if does not affect the straps registers after drive the gpio_o
+      csr_rd(.ptr(ral.hw_straps_data_in), .value(rd_hw_straps_data_in));
+      csr_rd(.ptr(ral.hw_straps_data_in_valid), .value(rd_hw_straps_data_in_valid));
```
rswarbrick, line 113: This could use `check_straps_regs` again.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// interface : gpio_straps_if
+`ifndef GPIO_STRAPS_IF_SV
+`define GPIO_STRAPS_IF_SV
+
+import gpio_pkg::*;
+
+// Interface definition
+interface gpio_straps_if (
+    inout                strap_en_i,     // Input signal to enable straps
```
rswarbrick, line 12: Nit: Stray space before "strap_en_i"

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,21 @@
...
+    input gpio_straps_t sampled_straps_o // Output signal of type gpio_straps_t
```
rswarbrick, line 13: I think something very strange is going on here. Do these need to be ports at all? I think probably not, and I definitely don't think an input port should have a name that ends in `_o`.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,21 @@
...
+);
+
+    logic strap_en;
+    assign strap_en_i = strap_en;
```
rswarbrick, line 17: Again, I'm not sure about the way this is using ports. Can't we just have an internal variable called `strap_en` and wire that up?

hw/ip/gpio/dv/tb/tb.sv
```diff
@@ -24,6 +22,8 @@ module tb;
   wire [NUM_GPIOS-1:0] gpio_oe;
   wire [NUM_GPIOS-1:0] gpio_intr;
   wire [NUM_MAX_INTERRUPTS-1:0] interrupts;
+  gpio_straps_t sampled_straps;
+  wire strap_en_i;
```
rswarbrick, line 26: This name doesn't quite work because the `_i` suffix is used for input ports in our style.

hw/ip/gpio/dv/tb/tb.sv
```diff
@@ -52,9 +54,8 @@ module tb;
     .clk_i (clk),
     .rst_ni(rst_n),
 
-    // TODO: need to test snapshot functionality
-    .strap_en_i(1'b0),
-    .sampled_straps_o(),
+    .strap_en_i(strap_en_i),
```
rswarbrick, line 57: Following on from the note I left in the interface, I think this could just use `straps_if_inst.strap_en`.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: Yes, I can do that

hw/ip/gpio/data/gpio.hjson
```diff
@@ -390,7 +390,9 @@
       ],
     },
     { name: "HW_STRAPS_DATA_IN",
-      desc: "GPIO Input data sampled as straps during cold boot read value",
+      desc: '''GPIO Input data sampled as straps during cold boot read value
+            The GPIO input data is sampled after the first one pulse cycle of straps_en input after reset.
+            ''',
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: Much better! Thanks!

hw/ip/gpio/data/gpio_testplan.hjson
```diff
@@ -145,5 +145,17 @@
       stage: V2
       tests: ["gpio_stress_all"]
     }
+    {
+      name: straps_data
+      desc: '''Verify the straps data/valid ouput expected values based on the strap_en and gpio_i inputs:
+      - Drives gpio_i input with random values
+      - Set the expected strap_data_in based on the condition if is the first triggered strap_en or not
+      - Updates the RAL model with the expected strap_data_in and strap_data_in_valid
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: very true, I forgot to remove this. And I have included some new comments/steps description.

hw/ip/gpio/data/gpio_testplan.hjson
```diff
@@ -145,5 +145,17 @@
...
+      - Trigger the strap_en with one clock cycle
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: fixed

hw/ip/gpio/data/gpio_testplan.hjson
```diff
@@ -145,5 +145,17 @@
...
+      - Read the registers hw_straps_data_in and hw_straps_data_in_valid
+      - Check the expected behavior in the scoreboard and also on csr checks'''
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: true, I will reformulate the phrase.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -31,6 +31,12 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
   //  (i) indicate that write to INTR_STATE register just happened, and
   // (ii) store information of which all interupt bits were cleared
   bit [TL_DW-1:0] cleared_intr_bits;
+  // hw straps_data_in_valid register
+  bit straps_data_in_valid;
+  // hw straps_data_in register
+  bit [NUM_GPIOS-1:0] straps_data_in;
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: you are right, I included those in the gpio_predict_and_compare() function, because was initial idea that this function it will be used in my case, but after some coding I realise that is not needed, but I forgot to remove it. I should done a proper clean up code, sorry.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
marcelocarvalhoLowRisc: If the interface is used primarily for encapsulation those signals, then you are correct, but in this case I will connect those ports signals to the GPIO module, otherwise I will have compilation issues.
You are right, besides the "input" explicitly in the sampled_straps_o feels weird/confuse, I kept the _o, just to signalise that this signal is a output from the DUT, but I can remove it, that is more reasonable, and put some comments.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
marcelocarvalhoLowRisc: In this case I am using the gpio_strap_en variable to initialise the strap_en_i with zero after the reset, otherwise I will have spreading X's to the input, so look at the gpio_base_test.sv

hw/ip/gpio/dv/tb/tb.sv
marcelocarvalhoLowRisc: ok, I will fix this =)

hw/ip/gpio/dv/tb/tb.sv
marcelocarvalhoLowRisc: Yes indeed, I would prefer to keep the single signal name instead, makes the code more clean, and the wire makes the "connection" between the interface and the DUT. But let me know if this does not make sense to you =)

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -31,6 +31,12 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+  // Flag to indicate that the strap was trigered
+  bit first_strap_trigged;
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: thanks =D, fixed

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +369,38 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+  // Task: monitor_gpio_straps
+  virtual task monitor_gpio_straps();
+    forever begin : monitor_gpio_intr
+      @(posedge cfg.clk_rst_vif.clk)
+        `uvm_info(`gfn, "Monitor gpio_straps on!", UVM_HIGH)
+        if (|gpio_i_driven === 1'b1) begin
+          @(negedge cfg.straps_vif_inst.strap_en_i) begin
+            if (first_strap_trigged == 0) begin
+              // Update data_in ral register value based on result of input
+              void'(ral.hw_straps_data_in.predict(.value(gpio_i_driven), .kind(UVM_PREDICT_DIRECT)));
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: I agreed, I just pick this from other part of the testbench so I thought that was the "common" way of using it. I am getting familiar with those macros, I will replace this and use the macro.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: done!

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: fixed

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: fixed

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: yes, it make sense to me as well

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: I removed the straps_triggered, don't need this in the sequence. Instead I have another flag in the scoreboard that makes more sense.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: Actually I don't need these at all, I have removed it.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: Also true, I removed it

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: Very true, make it cleaner code =)
I will cherry-pick that

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: I reviewed the entire sequence and replaced the <=1 to <=0 in the places the delay can be avoided.
I kept the delay <=1 after the strap_en trigger, to avoid a race condition with the prediction value of the register. And before and after the apply_reset() call.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: I never used this mirror task, I think this does not create a TL access, or am I wrong? anyway I kept the csr_rd but not using the read value back to the sequence, instead just doing like csr_rd(.ptr(ral.hw_straps_data_in));
And making sure the csr checks are done in scoreboard level.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: The same, I would prefer using TL access instead of backdoor methods to check the registers functionality.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: I rewrote the code and created a separated task for this. About the update, I thought that is a good approach to reuse what is already there, csr_* tasks. What do you mean about "use the update task directly" ? Can you give me an example?

hw/ip/gpio/dv/tests/gpio_base_test.sv
```diff
@@ -7,10 +7,22 @@ class gpio_base_test extends cip_base_test #(
   .CFG_T(gpio_env_cfg)
 );
   `uvm_component_utils(gpio_base_test)
-  `uvm_component_new
+  straps_vif straps_vif_inst; // Virtual interface
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: jumped back =)

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +369,38 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
     end
   endtask : monitor_gpio_interrupt_pins
 
+  // Task: monitor_gpio_straps
+  virtual task monitor_gpio_straps();
+    forever begin : monitor_gpio_intr
+      @(posedge cfg.clk_rst_vif.clk)
+        `uvm_info(`gfn, "Monitor gpio_straps on!", UVM_HIGH)
+        if (|gpio_i_driven === 1'b1) begin
+          @(negedge cfg.straps_vif_inst.strap_en_i) begin
```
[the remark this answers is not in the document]
rswarbrick: My proposal about writing some documentation still stands! I suggest switching to doing that while we wait to get this all merged.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +369,38 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+  // Based on strap_en input value, the straps register data should contains the snapshot of the data_in pins
+  function void check_straps();
+    // Checker: Compare actual values of gpio pins with straps register.
+    // Check the register hw_straps_data_in against gpio_i pins
+    `DV_CHECK_CASE_EQ(gpio_i_driven, cfg.straps_vif_inst.sampled_straps_o.data)
+    // Check the register hw_straps_data_in_valid
+    `DV_CHECK_CASE_EQ('b1, cfg.straps_vif_inst.sampled_straps_o.valid)
+    // Turn-off the checker after the first strap trigger.
+    first_strap_trigged = 'b1;
+  endfunction : check_straps
```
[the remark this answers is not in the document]
rswarbrick: I still think this suggestion applies!

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: I believe this "class: ..." line is still in the patch. Could you drop it please?

(I'd leave the blank line after the copyright header though)

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: This *should* be done by the scoreboard (and then you can just use `get_mirrored_value`, wrapped in the `gmv` macro). At a very quick glance, it doesn't *look* like the scoreboard actually does this. Boo!

But this probably needs adding, either for the `hw_straps_data_in` register in particular or (better) in something that looks up the register from the address. For an example, look at `otbn_scoreboard` line 324.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: Oh, I hadn't twigged that this works by applying a reset each time. Shouldn't this just do it once and then rely on `gpio_stress_all_vseq` to do the iteration?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: Ah, good point. That said, I think my note about resets probably applies here. If we just do one transaction per call, we won't have to re-randomise explicitly like this.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: Documentation for the `mirror` function: https://verificationacademy.com/verification-methodology-reference/uvm/docs_1.2/html/files/reg/uvm_reg-svh.html#uvm_reg.mirror

(found with a search engine...)

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: very true, removed.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: We could call it something else (`read_straps_regs`?), but what I'm suggesting *is* using TL accesses.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: My point about the output still applies though: you're currently updating some class variable which is "write-only". If it *must* exist, make it a local variable. But you could use the `mirror` function because you don't care (in the sequence) about the value of the variable.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: I was basing this on the fact that you used set+update. Some different ways of doing it:
```systemverilog
    // One option (using set and then the update task)
    ral.direct_out.set(gpio_out);
    ral.direct_out.update(.status(status));
    `DV_CHECK_FATAL(status == 1)

    // A second option (using the write RAL method)
    ral.direct_out.write(.status(status), .value(gpio_out));
    `DV_CHECK_FATAL(status == 1)

    // A third option (using csr_wr)
    csr_wr(.csr(ral.direct_out), .value(gpio_out));
```

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: This still stands: probably not sensible to repeat code like this.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: I think I missed the first comment, I will do the PR.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
rswarbrick: I think I don't know about the compilation issues you're talking about. Maybe you could explain what you tried and what didn't work?

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
rswarbrick: Can't `gpio_base_test` drop the internal value directly?

hw/ip/gpio/dv/env/gpio_scoreboard.sv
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: When you said:

> inline this into the single call site

means move the check_straps code to the monitor_gpio_straps() task? and drop the check_straps function?

hw/ip/gpio/dv/tb/tb.sv
rswarbrick: Hmm, I'm not really sure. I suppose one option would to add an output and an input port to the interface and wire them together on the inside. But that seems a little silly: I'm really not convinced that it matters whether the testbench looks at the internals of the interface?

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
marcelocarvalhoLowRisc: Yes it's possible, I removed the input/outputs and it's working fine. In my mind I always though like using interface, we are required to declare input/output ports, that is the interface by definition. In this case the interface behavior more like a wrapper for these signals to be used in the testbench, instead of driving directly from the testbench top level or some other way.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
marcelocarvalhoLowRisc: I have tried this:

```
// Interface definition
interface gpio_straps_if (
    input               strap_en,                       // Signal to enable straps
    output gpio_straps_t  sampled_straps // Sampled gpio_i snapshot data from GPIO (DUT)
);
```
 
But won't compile
But I resolved the way you recomended, without input/output ports.

hw/ip/gpio/dv/tb/tb.sv
marcelocarvalhoLowRisc: It's true, I did that way, straps_if_inst.strap_en
It seems more reasonable.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
marcelocarvalhoLowRisc: removed

hw/ip/gpio/dv/env/gpio_scoreboard.sv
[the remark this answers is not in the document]
rswarbrick: Yep, that's what I meant (and thanks for doing it!)

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
rswarbrick: You're right (it's basically now just a struct, which is pretty standard for OT DV so far). The "proper" way to do it is probably to use a `modport`. See IEEE 1800 section 25.5 for a description of how to use them.

hw/ip/gpio/dv/env/gpio_env_cfg.sv
```diff
@@ -27,6 +32,9 @@ class gpio_env_cfg extends cip_base_env_cfg #(
 
     // only support 1 outstanding TL item
     m_tl_agent_cfg.max_outstanding_req = 1;
+
+    // Used to allow reset operation during a stress all tests and check the CSR after that.
+    can_reset_with_csr_accesses = 1'b1;
```
rswarbrick, line 37: To avoid everything getting globbed together, can you pull this change into a separate commit (which can come first)?

That commit could be cherry-picked into a separate PR, that could be reviewed and merged without having to pin down the straps behaviour.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,20 @@
...
+
+import gpio_pkg::*;
+
+// Interface definition
+interface gpio_straps_if();
+
+    logic          strap_en;       // Signal to enable straps
+    gpio_straps_t  sampled_straps; // Sampled gpio_i snapshot data from GPIO (DUT)
+
+      modport straps_port(input strap_en, output sampled_straps);
```
martin-velay, line 16: Nit: alignment.

And for my knowledge, can you tell me why this `modport` is required?

hw/ip/gpio/dv/tests/gpio_base_test.sv
```diff
@@ -1,16 +1,25 @@
...
   .CFG_T(gpio_env_cfg)
 );
   `uvm_component_utils(gpio_base_test)
-  `uvm_component_new
-
+  straps_vif straps_vif_inst; // Virtual interface
+  function new(string name, uvm_component parent);
+    super.new(name, parent);
+  endfunction
+  task reset_phase(uvm_phase phase);
```
martin-velay, line 13: Nit: can you insert one line break between each task/function ? And same before `virtual function void build_phase(uvm_phase phase);`

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
rswarbrick: I *think* the idea was that such a port would allow one "actor" to see things in the correct direction.

But (for @marcelocarvalhoLowRisc):
- Doesn't the DV code need to sample through the port? I'm thinking about tasks like `monitor_gpio_straps`.
- But don't we need a port for things driving the `strap_en` state as well? (`gpio_rand_straps_vseq`)
- And the ports should probably be named in a way that relates to who uses them. At the moment, we have a straps port on a straps interface, which is slightly doubled!
- Finally: The indentation needs sorting out here.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +365,45 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+
+  // Task: monitor_gpio_straps
+  // The task monitors the gpio straps enable signal
+  // and checks the straps output signal after the first strap trigger
+  virtual task monitor_gpio_straps();
+    forever begin : monitor_gpio_straps
+      @(posedge cfg.clk_rst_vif.clk)
+      if(!cfg.under_reset) begin
+        if (|gpio_i_driven === 1'b1) begin
+          @(posedge cfg.straps_vif_inst.port_out.strap_en) begin
```
rswarbrick, line 387: Hmm, I don't think this is going to work very nicely: it will block the "every clock" logic that is there otherwise. I think you probably need to rephrase it slightly, updating local variables on every clock edge with the current value of `strap_en` and then checking each time for a posedge and looking two cycles back (`if (cfg.straps_vif_inst.port_out.strap_en && !strap_en_qq) ...`).

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +365,45 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+            // Wait for at least 1 clock cycle after strap_en is asserted, to allow the straps to be
+            // ready for sampling.
+            cfg.clk_rst_vif.wait_clks(1);
+            update_gpio_straps_regs();
```
rswarbrick, line 391: Doesn't this need to be conditional on whether it's the first time?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: I think this note still applies?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
rswarbrick: I think this note still applies?

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,153 @@
...
+  endfunction
+
+  task test_straps_gpio_in();
+
+    // Drive the gpio_in
+    drive_gpio_in(gpio_in);
+
+    // Wait at least one clock cycle to drive the strap_en
+    // Required because is required one clock cycle to update the gpio_in regsisters.
+    `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(delay, delay >= 1;)
```
rswarbrick, line 46: This is reusing `delay` (a base class variable that's supposed to be gaps between transactions) for something else. I'd just randomise a local variable.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,153 @@
...
+//    - Drive gpio_in input with random values.
+//    - Set the strap_en high for at least one clock cycle.
+//    - Read the registers hw_straps_data_in and hw_straps_data_in_valid.
+//    - The data read and sampled_straps_o will be checked in the scoreboard.
+//    - Drive the gpio_out to make sure that has no impact on straps registers.
+//    - Read to make sure that if does not affect the straps registers after drive the gpio_out.
+//    - Apply reset and make sure the strap registers are clean.
+//    - Read straps registers after reset.
+//    - Iterate again the same flow, with new random values.
+class gpio_rand_straps_vseq extends gpio_base_vseq;
```
rswarbrick, line 15: Thinking about this a bit, can't we drop this vseq entirely? Maybe the right approach is:
- Update the base vseq so that the `strap_en` signal gets driven at unexpected times.
- Teach the scoreboard when to sample the pins and store the results in the straps registers, updating the RAL.

Then I think tests like the CSR RW test should get all the behaviour you need?

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,21 @@
...
+import gpio_pkg::*;
+
+// Interface definition
+interface gpio_straps_if();
+
+  logic          strap_en;       // Signal to enable straps
+  gpio_straps_t  sampled_straps; // Sampled gpio_i snapshot data from GPIO (DUT)
+
+  modport port_in(input strap_en, output sampled_straps);
+  modport port_out(output strap_en, input sampled_straps);
```
rswarbrick, line 17: I'm not sure about the naming here: "in" and "out" both have one input and one output! :-)

Maybe name the ports after who is expected to do this sort of access?

hw/ip/gpio/rtl/gpio.sv
```diff
@@ -85,7 +85,6 @@ module gpio
     assign hw2reg.hw_straps_data_in.d        = data_in_d;
     assign sampled_straps_o.data             = reg2hw.hw_straps_data_in.q;
     assign sampled_straps_o.valid            = reg2hw.hw_straps_data_in_valid.q;
-    `ASSUME(StrapSampleOnce_A, ##1 $fell(sample_trigger) |-> always !sample_trigger)
```
rswarbrick, line 88: I'd probably pull this 1-line change into a separate commit and push it as a PR in parallel. It's obviously correct.

hw/ip/gpio/data/gpio_testplan.hjson
```diff
@@ -145,5 +145,20 @@
       stage: V2
       tests: ["gpio_stress_all"]
     }
+    {
+      name: straps_data
+      desc: '''Verify the straps data/valid ouput expected values based on the strap_en and gpio_i inputs:
+      - Drives gpio_i input with random values.
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: Fixed

marcelocarvalhoLowRisc (commented):
ready to be reviewed again

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,34 @@
...
+  logic          strap_en;       // Signal to enable straps
+  logic          strap_en_qq;
+  logic          strap_en_q;
+  gpio_straps_t  sampled_straps; // Sampled gpio_i snapshot data from GPIO (DUT)
+
+  modport dut_if(input strap_en, output sampled_straps);
+  modport tb_if(output strap_en, input sampled_straps);
+
+  
+    always_ff @(posedge clk or negedge rst_n) begin
```
martin-velay, line 22: Nit: I think there is still an indentation issue for this whole block.
And by convention, I don't know if usually we do "qq" or "q2" here at lowRISC, @rswarbrick ?

martin-velay (approved):
From what I can see most of the comments have been covered and this PR looks better now.  If the functionality part is OK, I am keen on having it merged.

hw/dv/sv/csr_utils/csr_utils_pkg.sv
marcelocarvalhoLowRisc, line 1: I created this csr_mirror, based on csr_read but with only the necessary parameters. That is more useful instead of doing this all uvm_reg decode and checks and I kept the increment_outstanding.... task as well, that I think is required if with run this with rnd_reset stress all for example...

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +365,44 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+  endtask : update_gpio_straps_regs
+
+  // Task: monitor_gpio_straps
+  // The task monitors the gpio straps enable signal
+  // and checks the straps output signal after the first strap trigger
+  virtual task monitor_gpio_straps();
+    forever begin : monitor_gpio_straps
+      @(posedge cfg.clk_rst_vif.clk)
+      if(!cfg.under_reset) begin
+        if (|gpio_i_driven === 1'b1) begin
```
rswarbrick, line 384: This looks a bit surprising to me: isn't `gpio_i_driven` the values that are being driven? I think we probably expect the `strap_en` feature to work even if all the inputs are zero...

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +365,44 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+                                                        .kind(UVM_PREDICT_READ)));
+  endtask : update_gpio_straps_regs
+
+  // Task: monitor_gpio_straps
+  // The task monitors the gpio straps enable signal
+  // and checks the straps output signal after the first strap trigger
+  virtual task monitor_gpio_straps();
+    forever begin : monitor_gpio_straps
+      @(posedge cfg.clk_rst_vif.clk)
+      if(!cfg.under_reset) begin
```
rswarbrick, line 383: I'm not sure this is going to work properly if any resets get injected in the middle of the test. I'd probably suggest this instead, which is much simpler:
```systemverilog
    forever begin
      wait(!cfg.under_reset);
      fork begin : isolation_fork
        fork
          wait(cfg.under_reset);
          forever begin
            @(posedge cfg.clk_rst_vif.clk);
            // The monitoring code here
          end
        join_any
        disable fork;
      end join
    end
```

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,151 @@
...
+  endfunction
+
+  task test_straps_gpio_in();
+    // Drive the gpio_in
+    drive_gpio_in(gpio_in);
+
+    // Wait at least one clock cycle to drive the strap_en
+    // Required because is required one clock cycle to update the gpio_in regsisters.
+    `DV_CHECK_MEMBER_RANDOMIZE_WITH_FATAL(m_delay, m_delay >= 1;)
+    cfg.clk_rst_vif.wait_clks(m_delay);
```
rswarbrick, line 47: Thanks for splitting this out of the class otherwise. But I think you can simplify further still. This is untested, but couldn't we use something like this?
```systemverilog
  // Wait a few cycles. If force_positive is true, Wait at least one clock cycle.
  task short_wait(bit force_positive);
    int unsigned delay;
    `DV_CHECK_FATAL(std::randomize(delay) with
                    {
                      delay dist {0 :/ 20, [1:5] :/ 40, [6:15] :/ 30, [20:25] :/ 10};
                      force_positive -> delay > 0;
                    })
    cfg.clk_rst_vif.wait_clks(delay);
  endtask
```

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
rswarbrick: I think I've seen `_qq` more, but I'm not sure there's a proper "convention".

martin-velay (approved):
I think it'll be OK when the comments will be taken into account, particularly the one from Rupert. So I am approving but please consider them.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
marcelocarvalhoLowRisc: True, I liked this way as well, I have tested and it works, thanks  for the suggestion!

marcelocarvalhoLowRisc (commented):
New updates based on the latest comments.

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,144 @@
...
+  // gpio input to drive
+  rand bit [NUM_GPIOS-1:0] gpio_in;
+  // gpio output to program in register
+  rand bit [NUM_GPIOS-1:0] gpio_out;
+  // gpio output enable to program in register
+  rand bit [NUM_GPIOS-1:0] gpio_oe;
+
+  constraint num_trans_c {
+    num_trans inside {[1:3]};
+  }
```
rswarbrick, line 28: I think that now this doesn't loop any more (after the discussion [here](https://github.com/lowRISC/opentitan/pull/25868#discussion_r1918912948), we don't need to mess around with `num_trans` at all?

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
rswarbrick: This still applies: I think the modports are called `dut_if` and `tb_if` here at the moment.

hw/ip/gpio/data/gpio_testplan.hjson
```diff
@@ -145,5 +145,21 @@
...
     }
+    {
+      name: straps_data
+      desc: '''Verify the straps data/valid ouput expected values based on the strap_en and gpio_i inputs:
+      - Drive gpio_i input with random values.
+      - Set strap_en high for at least one clock cycle.
+      - Read the registers hw_straps_data_in and hw_straps_data_in_valid.
+      - Check the data read and sampled_straps_o in the scoreboard.
+      - Drive gpio_o to make sure that has no impact on straps registers.
+      - Verify that driving gpio_o does not affect the straps registers.
```
rswarbrick, line 156: Nit: I think these can probably be collapsed into a single line.

hw/ip/gpio/data/gpio_testplan.hjson
```diff
@@ -145,5 +145,21 @@
...
+      - Apply a reset and ensure the strap registers are cleared.
+      - Read straps registers after reset.
+      - Iterate through the same flow again with new random values.
```
rswarbrick, line 159: I think this text now needs to say something about the fact we might be used in the stress sequence?

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +365,53 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+    forever begin : monitor_gpio_straps
+        wait(!cfg.under_reset);
+        fork begin : isolation_fork
+          fork begin
+            wait(cfg.under_reset);
+            forever begin
+              @(posedge cfg.clk_rst_vif.clk)
+              if(cfg.straps_vif_inst.tb_if.strap_en &&
+                !cfg.straps_vif_inst.strap_en_q     &&
+                !cfg.straps_vif_inst.strap_en_qq) begin
```
rswarbrick, line 389: I'm slightly surprised about this: we're looking at the first posedge after two zeros ("001"). Why do we care about it being low for two cycles? (What if the straps are enabled from the start of time? I wouldn't expect that to be a problem?)

hw/ip/gpio/dv/env/gpio_scoreboard.sv
```diff
@@ -360,6 +365,53 @@ class gpio_scoreboard extends cip_base_scoreboard #(.CFG_T (gpio_env_cfg),
...
+                // Wait for at least 1 clock cycle after strap_en is asserted,
+                // to allow the straps to be
+                // ready for sampling.
+                cfg.clk_rst_vif.wait_clks(1);
+                update_gpio_straps_regs();
```
rswarbrick, line 394: I don't think this is actually right, because you're going to update the register to sample the input on every clock. I think this is probably simpler (and a bit easier):
```systemverilog
  // Task: monitor_gpio_straps
  // The task monitors the gpio straps enable signal
  // and checks the straps output signal after the first strap trigger
  virtual task monitor_gpio_straps();
    forever begin : monitor_gpio_straps
        wait(!cfg.under_reset);
        fork begin : isolation_fork
          fork begin
            wait(cfg.under_reset);
            begin
              cfg.clk_rst_vif.wait_clks(1);
              // Wait until strap_en goes high after at least two cycles low. (TODO: Really?! This
              // looks rather odd to me)
              while (! (cfg.straps_vif_inst.tb_if.strap_en &&
                        !cfg.straps_vif_inst.strap_en_q    &&
                        !cfg.straps_vif_inst.strap_en_qq))
                cfg.clk_rst_vif.wait_clks(1);

              // Wait for at least 1 clock cycle after strap_en is asserted, to allow the straps to
              // be ready for sampling.
              cfg.clk_rst_vif.wait_clks(1);
              update_gpio_straps_regs();

              // Checker: Compare actual values of gpio pins with straps register.
              // Check the register hw_straps_data_in against gpio_i pins
              `DV_CHECK_CASE_EQ(gpio_i_driven, cfg.straps_vif_inst.tb_if.sampled_straps.data)
              // Check the register hw_straps_data_in_valid
              `DV_CHECK_CASE_EQ('b1, cfg.straps_vif_inst.tb_if.sampled_straps.valid)
            end
          end
          join_any
          disable fork;
        end join
    end
  endtask : monitor_gpio_straps
```

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,144 @@
...
+    short_wait(1);
+
+    // Read the hw_straps_data_in and check the expected value in the scoreboard
+    csr_strap_read();
+
+    // Random wait
+    short_wait(0);
+
+    // Stop driving gpio_in
+    undrive_gpio_in();
```
rswarbrick, line 71: I'm not sure about this. Don't we want to keep driving the inputs, but with different values? (`~gpio_in`?)

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,144 @@
...
+
+    // Random wait
+    short_wait(0);
+
+    // Read to make sure that if does not affect the straps registers after undrive the gpio_in
+    csr_strap_read();
```
rswarbrick, line 77: Can you check that this actually does anything at the moment? If I've understood your code, this is currently predicting something silly in the straps registers. It would be good to know that we're actually checking something!

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,144 @@
...
+
+    // Additional verification
+    // Drive the gpio_out to make sure that has no impact on straps registers.
+    // then read the gpio strap registers again
+    cfg.gpio_vif.drive_en('0);
+
+    ral.direct_out.set(gpio_out);
+    ral.direct_oe.set(gpio_oe);
+    csr_update(.csr(ral.direct_out));
+    csr_update(.csr(ral.direct_oe));
```
rswarbrick, line 91: If we're just setting the register and updating to match, it might be easier to use a `wr` method (`csr_wr`?)

hw/ip/gpio/dv/env/seq_lib/gpio_rand_straps_vseq.sv
```diff
@@ -0,0 +1,144 @@
...
+
+    // Random wait
+    short_wait(0);
+
+    // Read to make sure that if does not affect the straps registers after drive the gpio_out
+    csr_strap_read();
+
+  endtask : test_straps_gpio_out
+
+  task check_transaction(string txn_desc, bit is_first);
```
rswarbrick, line 101: Can you add a documentation comment that explains what this task does? I'm slightly surprised to see that it injects a reset in the middle, and also that it doesn't look at the `is_first` argument.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
rswarbrick: Argh! That wasn't quite right: you'll also need to wait for a reset after the fork, I think (to avoid looping the strap check)

hw/ip/gpio/data/gpio_testplan.hjson
marcelocarvalhoLowRisc: Fixed.

hw/ip/gpio/dv/env/gpio_scoreboard.sv
rswarbrick: This comment still applies. I know that you plan to refactor this, but we shouldn't merge something we know to be silly in the meantime...

rswarbrick:
Thanks, this looks really good. I think the only item remaining is [this one](https://github.com/lowRISC/opentitan/pull/25868#discussion_r1965203378) and then we can merge the change :-)

rswarbrick:
The current code looks like this:
image

I'm suggesting you change the modport names to something that doesn't look like an interface in *this* PR.

marcelocarvalhoLowRisc:
> The current code looks like this: image
> 
> I'm suggesting you change the modport names to something that doesn't look like an interface in _this_ PR.

ohhh, ok ok, got it, sorry my bad,  I misunderstood.

rswarbrick:
The private CI failure is a top-level run which definitely won't be affected by this change. Looks good: thanks!

rswarbrick (approved):

PR #25873, opened 2025-01-14
[orchestrator] fix bugs in orchestrator.py script

timothytrippel:
This fixes two bugs in the orchestrator.py script:
1. a bug in the computation and construction of the Device ID when parts are provisioned from a TestLocked* stage, i.e., if part has already previously run through CP and then runs through FT at a later stage. In this scenario, the there is no CP device ID returned to the Host (since CP stage is not run), so FT must trust what already exists in flash, and the final device ID is read out by the host over JTAG at the end of provisioning,
2. a device record was upserted into the registry database regardless if it already existed in the database, without flashing any warnings.

Additionally, this enhances the E2E testing of the orchestrator.py script catch these corner cases.

### DIFF
--- a/sw/device/silicon_creator/manuf/lib/individualize.c
+++ b/sw/device/silicon_creator/manuf/lib/individualize.c
@@ -75,7 +75,7 @@ static status_t hw_cfg1_enable_knobs_set(const dif_otp_ctrl_t *otp_ctrl) {
 status_t manuf_individualize_device_hw_cfg(
     dif_flash_ctrl_state_t *flash_state, const dif_otp_ctrl_t *otp_ctrl,
     dif_flash_ctrl_region_properties_t flash_info_page_0_permissions,
-    uint32_t *device_id) {
+    uint32_t *ft_device_id) {
   bool is_locked;
 
   // Provision HW_CFG0 if it is not locked.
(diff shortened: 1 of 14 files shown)


sw/device/silicon_creator/manuf/lib/individualize.c
```diff
@@ -75,71 +75,59 @@ static status_t hw_cfg1_enable_knobs_set(const dif_otp_ctrl_t *otp_ctrl) {
...
+    // On non-silicon targets, we expect the CP device ID from flash to be
+    // empty. In this case we set the HW origin portion of the CP device ID.
+    // Otherwise, we expect the CP device ID to be present and non-zero.
+    if (flash_cp_device_id_empty) {
+      if (kDeviceType != kDeviceSilicon) {
+        cp_device_id[0] = 0x00024001u;
+        memset(&cp_device_id[1], 0,
+               kFlashInfoFieldCpDeviceIdSizeIn32BitWords - 1);
+        memset(&cp_device_id[2], 0,
+               kFlashInfoFieldCpDeviceIdSizeIn32BitWords - 1);
```
cfrantz, line 117: I don't think this does what you intend.  The size argument to memset is in bytes.  I think this should be something more like:
```
memset(&cp_device_id[1], 0, sizeof(cp_device_id) - sizeof(cp_device_id[0]));
```

However, it's probably even better to memset the entire thing to zero and then fill in the one word that you care about:
```
memset(&cp_device_id, 0, sizeof(cp_device_id));
cp_device_id[0] = 0xWhatEvs;
```

cfrantz (approved):

moidx (approved):

sw/device/silicon_creator/manuf/lib/individualize.c
timothytrippel: ah, yes. nice catch! Fixed.

PR #25879, opened 2025-01-15
[bazel] update rules_rust to 0.56

nbdd0121:
This requires some fixes to Rust codebase to keep clippy happy.


jwnrt (approved):
LGTM, thanks. Surprised so much was dropped from the lock file but don't think it's worth working out why

nbdd0121:
rules_rust 0.56 split bindgen/wasm/protobuf into their own extensions and we're only pulling back bindgen.

PR #25885, opened 2025-01-15
[hw,dma,rtl] Change system port metadata to 3

Razer6:
Downstream interface changed, lets adapt that.


alees24 (approved):
I'll defer to you on this.

PR #25892, opened 2025-01-15
[rom_ext] Lock the correct ROM_EXT region

### LINKED ISSUE #25435: [rom_ext] Inactive ROM_EXT is being locked down via flashcontroller
### Description

With ROM 7.0.7, it appears that ROM_EXT is locking down the inactive ROM_EXT region instead of the active ROM_EXT via the flash controller.

```shell
$ opentitantool rescue get-boot-log
...
  "rom_ext_slot": "SlotA",
  "rom_ext_major": 0,
  "rom_ext_minor": 7,
  "rom_ext_size": 65536,
  "rom_ext_nonce": 0,
  "bl0_slot": "Unknown",
  "ownership_state": "LockedOwner",
...
```

Dump from owner firmware (which agrees that active ROM_EXT is `SlotA`):
```
> version
version
RO_A:  * 7.0.7/00000000
RO_B:    7.0.7/00000000
RW_A:    0.36.4/DBG/ti50_common_tot:v0.0.1623-249835b2
RW_B:  * 0.36.4/DBG/ti50_common_tot:v0.0.1623-249835b2
> md 0x41000050 20
md 0x41000050 20
0x41000050: 09666666 09669966 09666666 09669966
0x41000060: 06696666 06696666 09669966 06696666
0x41000070: 00004000 00004100 00019c20 00019d20
0x41000080: 000024ee 000023ee 000003ff 00001dee
0x41000090: 00966666 00000000 00000000 00000000
```

Notice that `region[1]` is `0x00004100` (which is page 256 to 288, i.e. SlotB ROM_EXT) has permission `0x09669966`, which means that program and erase are denied. However, `region[0]`, which corresponds to SlotA ROM_EXT, has permission `0x09666666`, which is fully writable.

cfrantz:
Ownership flash lockdown was protecting and locking all regions in the same slot that booted the owner code.  However, the ROM_EXT and owner code don't have to boot from the same side of the flash.

1. Protect the ROM_EXT and owner regions each according to their boot slot.
2. Test each combination of ROM_EXT slot and owner slot in the `flash_permission_test`.

Fixes #25435.


moidx (approved):

sw/device/silicon_creator/lib/ownership/owner_block.c
```diff
@@ -229,19 +235,47 @@ rom_error_t owner_block_flash_apply(const owner_flash_config_t *flash,
...
+      uint32_t pwp =
+          bitfield_field32_read(val, FLASH_CONFIG_PROTECT_WHEN_PRIMARY);
+      hardened_bool_t lock =
+          bitfield_field32_read(val, FLASH_CONFIG_LOCK) != kMultiBitBool4False
+              ? kHardenedBoolTrue
+              : kHardenedBoolFalse;
+      uint32_t region_start = config->start;
+      uint32_t region_end = region_start + config->size;
+
+      // When applying the flash configuration, a region is a ROM_EXT region if
```
jettr, line 247: Thank you for the detailed comment! The comment and the logic don't quite match either other. I agree with the comment. The logic to match the comment would be

```
kRomExtAStart <= region_start < kRomExtAEnd &&
kRomExtAStart < region_end <= kRomExtAEnd
```

(and maybe we need an explicit `region_start<=region_end` check too?)

jettr (changes requested):

sw/device/silicon_creator/lib/ownership/owner_block.c
cfrantz: :facepalm: How embarrassing.  A total logic fail.

I've fixed this and added unit tests to `owner_block_unittest.cc` to test the overlap/exclusivity checking functions.

We don't need to test `region_start <= region_end` because the bounds are computed from a start and size (which are `uint16_t`) and the `region_*` values are `uint32_t`.

I added a sanity check to the `end` value: if the end is beyond the total size of the flash, the configuration is invalid.

jettr (approved):

PR #25894, opened 2025-01-16
[topgen] Generate files for all address spaces

a-will:
Individually generate C and Rust collateral for all address spaces

Fix up alert and irq domains to enable generating collateral where the
alerts and irqs don't necessarily generate from devices in the same
address space as the CPU. Darjeeling has bus bridges that make servicing
those interrupts possible (indirect reachability to the IP).

The autogenerated tests will surely be bonkers, but that is to be fixed
up later.


a-will:
This PR reworks #23122 a bit and adds comments that explain what the functions are trying to do and where there are limitations.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.h
```diff
@@ -0,0 +1,341 @@
...
+extern "C" {
+#endif
+
+/**
+ * Peripheral base address for dmi device on lc_ctrl in top darjeeling.
+ *
+ * This should be used with #mmio_region_from_addr to access the memory-mapped
+ * registers associated with the peripheral (usually via a DIF).
+ */
+#define TOP_DARJEELING_LC_CTRL_DMI_BASE_ADDR 0x20000u
```
pamaury, line 37: I wonder if the types and defines in this file should all be prefixed by `<topname>_<addr>` ? Otherwise it makes it impossible for a file to include the header of two address spaces. I can see cases where it could make sense for a file to want to know the address of a peripheral on another address space than its own.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg_memory.ld
```diff
@@ -0,0 +1,58 @@
+/* Copyright lowRISC contributors (OpenTitan project). */
```
pamaury, line 1: I wonder if it makes sense to generate a linker script file for all address spaces? If the address space contains no Ibex, it's very unclear what purpose it could have.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.h
a-will: Yes, at some point, they should :)

I have not brought that in yet because it'll be another very large commit to change names. Then we'd do it over again for DIT multi-top work. For DIT, we'll probably need to throw in the notion of address spaces as a first-class citizen.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg_memory.ld
a-will: I kind of wonder if the linker script is even a sensible output for topgen to make in general. Seems like there would be a lot of assumptions.

hw/top_darjeeling/rtl/autogen/top_darjeeling_soc_dbg_pkg.sv
```diff
@@ -0,0 +1,442 @@
...
+
+  /**
+   * Peripheral size in bytes for jtag device on soc_dbg_ctrl in top darjeeling.
+   */
+  parameter int unsigned TOP_DARJEELING_SOC_DBG_CTRL_JTAG_SIZE_BYTES = 32'h20;
+
+
+  // Enumeration of alert modules
+  typedef enum int unsigned {
+    TopDarjeelingAlertPeripheralUart0 = 0,
```
Razer6, line 57: I think alert handler definitions etc should not be rendered in this address space? Same for the other files.

Razer6 (changes requested):
That looks good. Thanks for improving my works. A quick question, right now some definitions are rendered in all address space. I guess that should only be rendered in the appropriate address space.

hw/top_darjeeling/rtl/autogen/top_darjeeling_soc_dbg_pkg.sv
a-will: Ah, guess I didn't get all the files for checking for alert_handler presence. Will fix.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg_memory.ld
a-will: I changed linker scripts to only be generated for the "hart" address space for now. However, I think we ought to consider this be punted out of topgen. The top hjson-derived constants can be pulled in from the top or DIT header, since we're going to be preprocessing linker scripts.

Otherwise, this blurb from topgen makes opinionated / hard-coded assumptions about the software running on the chip. That integration layer would likely be best left to manually written code, I feel.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.h
a-will: I guess since the outputs have not existed (and thus have had no users) yet, I can switch it to have the address space embedded for all but the "hart" address space. We can fix up the special "hart" space later, since it ought not to have special handling.

To reclaim characters, we probably also ought to drop the top-specific naming.

In any case... done.

hw/top_darjeeling/rtl/autogen/top_darjeeling_soc_dbg_pkg.sv
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * Peripheral size in bytes for jtag device on soc_dbg_ctrl in top darjeeling.
+   */
+  parameter int unsigned TOP_DARJEELING_SOC_DBG_SOC_DBG_CTRL_JTAG_SIZE_BYTES = 32'h20;
+
+
+  // Enumeration of IO power domains.
+  // Only used in ASIC target.
+  typedef enum logic [0:0] {
```
Razer6, line 57: I don't think the pinmux config need to be rendered for the `soc_dbg` address space.

hw/top_darjeeling/rtl/autogen/top_darjeeling_soc_mbx_pkg.sv
```diff
@@ -0,0 +1,343 @@
...
+
+  /**
+   * Peripheral size in bytes for soc device on mbx_pcie1 in top darjeeling.
+   */
+  parameter int unsigned TOP_DARJEELING_SOC_MBX_MBX_PCIE1_SOC_SIZE_BYTES = 32'h20;
+
+
+  // Enumeration of IO power domains.
+  // Only used in ASIC target.
+  typedef enum logic [0:0] {
```
Razer6, line 107: The same for the `mbx` address space.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.h
```diff
@@ -0,0 +1,341 @@
...
+//  0 and 1 are tied to value 0 and 1
+#define NUM_MIO_PADS 12
+#define NUM_DIO_PADS 73
+
+#define PINMUX_PERIPH_OUTSEL_IDX_OFFSET 3
+
+/**
+ * Pinmux Peripheral Input.
+ */
+typedef enum top_darjeeling_pinmux_peripheral_in {
```
Razer6, line 117: Same for the headers.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.h
```diff
@@ -0,0 +1,341 @@
...
+  kTopDarjeelingMuxedPadsMio9 = 9, /**<  */
+  kTopDarjeelingMuxedPadsMio10 = 10, /**<  */
+  kTopDarjeelingMuxedPadsMio11 = 11, /**<  */
+  kTopDarjeelingMuxedPadsLast = 11, /**< \internal Last valid muxed pad */
+} top_darjeeling_muxed_pads_t;
+
+/**
+ * Power Manager Wakeup Signals
+ */
+typedef enum top_darjeeling_power_manager_wake_ups {
```
Razer6, line 282: Also for the `pwrmgr` definitions.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.h
```diff
@@ -0,0 +1,341 @@
...
+  kTopDarjeelingPowerManagerResetRequestsSocProxyRstReqExternal = 1, /**<  */
+  kTopDarjeelingPowerManagerResetRequestsLast = 1, /**< \internal Last valid pwrmgr reset_request signal */
+} top_darjeeling_power_manager_reset_requests_t;
+
+/**
+ * Clock Manager Software-Controlled ("Gated") Clocks.
+ *
+ * The Software has full control over these clocks.
+ */
+typedef enum top_darjeeling_gateable_clocks {
```
Razer6, line 315: And for the `clkmgr` definitions.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_mbx.h
```diff
@@ -0,0 +1,431 @@
...
+//  0 and 1 are tied to value 0 and 1
+#define NUM_MIO_PADS 12
+#define NUM_DIO_PADS 73
+
+#define PINMUX_PERIPH_OUTSEL_IDX_OFFSET 3
+
+/**
+ * Pinmux Peripheral Input.
+ */
+typedef enum top_darjeeling_pinmux_peripheral_in {
```
Razer6, line 207: Also here - probably because the same helper does not detect there is no `pinmux` and other IPs down below.

hw/top_darjeeling/sw/autogen/top_darjeeling_soc_dbg.h
```diff
@@ -0,0 +1,341 @@
...
+  kTopDarjeelingPowerManagerWakeUpsSensorCtrlWkupReq = 2, /**<  */
+  kTopDarjeelingPowerManagerWakeUpsSocProxyWkupInternalReq = 3, /**<  */
+  kTopDarjeelingPowerManagerWakeUpsSocProxyWkupExternalReq = 4, /**<  */
+  kTopDarjeelingPowerManagerWakeUpsLast = 4, /**< \internal Last valid pwrmgr wakeup signal */
+} top_darjeeling_power_manager_wake_ups_t;
+
+/**
+ * Reset Manager Software Controlled Resets
+ */
+typedef enum top_darjeeling_reset_manager_sw_resets {
```
Razer6, line 294: And `rstmgr` definitions.

hw/top_englishbreakfast/rtl/autogen/top_englishbreakfast_pkg.sv
```diff
@@ -272,62 +272,6 @@ package top_englishbreakfast_pkg;
   parameter int unsigned TOP_ENGLISHBREAKFAST_ROM_SIZE_BYTES = 32'h8000;
 
 
-  // Enumeration of alert modules
-  typedef enum int unsigned {
```
Razer6, line 276: That diff is because `EnglishBreakfast` does not have an alerthandler, right?

sw/host/opentitanlib/src/chip/autogen/darjeeling_soc_dbg.rs
```diff
@@ -0,0 +1,172 @@
...
+// Please do not modify content of this file directly.
+// File generated by using template: "host_toplevel.rs.tpl"
+// To regenerate this file follow OpenTitan topgen documentations.
+
+#![allow(dead_code)]
+
+use crate::with_unknown;
+
+with_unknown! {
+    pub enum PinmuxPeripheralIn: u32 [default = Self::End] {
```
Razer6, line 15: That definitions need to be dropped. No pinmux in this address space.

sw/host/opentitanlib/src/chip/autogen/darjeeling_soc_mbx.rs
```diff
@@ -0,0 +1,172 @@
...
+// Please do not modify content of this file directly.
+// File generated by using template: "host_toplevel.rs.tpl"
+// To regenerate this file follow OpenTitan topgen documentations.
+
+#![allow(dead_code)]
+
+use crate::with_unknown;
+
+with_unknown! {
+    pub enum PinmuxPeripheralIn: u32 [default = Self::End] {
```
Razer6, line 15: Same commend as above.

sw/host/opentitanlib/src/chip/autogen/top_darjeeling_soc_dbg.rs
```diff
@@ -0,0 +1,519 @@
...
+///
+/// This is the size (in bytes) of the peripheral's reserved memory area. All
+/// memory-mapped registers associated with this peripheral should have an
+/// address between #SOC_DBG_CTRL_JTAG_BASE_ADDR and
+/// `SOC_DBG_CTRL_JTAG_BASE_ADDR + SOC_DBG_CTRL_JTAG_SIZE_BYTES`.
+pub const SOC_DBG_CTRL_JTAG_SIZE_BYTES: usize = 0x20;
+
+// PERIPH_INSEL ranges from 0 to NUM_MIO_PADS + 2 -1}
+//  0 and 1 are tied to value 0 and 1
+pub const NUM_MIO_PADS: usize = 12;
```
Razer6, line 82: No pinmux in this address space, let's drop the definitions.

sw/host/opentitanlib/src/chip/autogen/top_darjeeling_soc_dbg.rs
```diff
@@ -0,0 +1,519 @@
...
+            11 => Ok(Self::Mio11),
+            _ => Err(val),
+        }
+    }
+}
+
+/// Power Manager Wakeup Signals
+#[derive(Copy, Clone, PartialEq, Eq)]
+#[repr(u32)]
+pub enum PowerManagerWakeUps {
```
Razer6, line 465: No pwrmgr in this address space, let's drop the definitions.

sw/host/opentitanlib/src/chip/autogen/top_darjeeling_soc_dbg.rs
```diff
@@ -0,0 +1,519 @@
...
+    AonTimerAonWkupReq = 1,
+    SensorCtrlWkupReq = 2,
+    SocProxyWkupInternalReq = 3,
+    SocProxyWkupExternalReq = 4,
+}
+
+/// Reset Manager Software Controlled Resets
+#[derive(Copy, Clone, PartialEq, Eq)]
+#[repr(u32)]
+pub enum ResetManagerSwResets {
```
Razer6, line 476: No rstmgr in this address space, let's drop the definitions.

sw/host/opentitanlib/src/chip/autogen/top_darjeeling_soc_mbx.rs
```diff
@@ -0,0 +1,589 @@
...
+pub const NUM_MIO_PADS: usize = 12;
+pub const NUM_DIO_PADS: usize = 73;
+
+pub const PINMUX_MIO_PERIPH_INSEL_IDX_OFFSET: usize = 2;
+pub const PINMUX_PERIPH_OUTSEL_IDX_OFFSET: usize = 3;
+
+/// Pinmux Peripheral Input.
+#[derive(Copy, Clone, PartialEq, Eq)]
+#[repr(u32)]
+pub enum PinmuxPeripheralIn {
```
Razer6, line 161: Same comments as above.

Razer6 (commented):
I think there still some output definitions that need to be dropped by adapting the IP detection in the template files.

a-will:
> I think there still some output definitions that need to be dropped by adapting the IP detection in the template files.

Those were not actually focus areas of this PR (which targets just the base addresses, interrupts, and alerts), but I could look at handling them here too.

hw/top_englishbreakfast/rtl/autogen/top_englishbreakfast_pkg.sv
a-will: Right.

Razer6 (approved):
Thanks Alex for removing the unneeded definitions. This looks good to me!

pamaury (approved):

jwnrt:
FYI I think there was some kind of merge skew here with Rust linting:

```console
$ bazel run //quality:rustfmt_fix
Target //quality:rustfmt_fix up-to-date:
  bazel-bin/quality/rustfmt_fix.bash
Error writing files: failed to resolve mod `top_darjeeling_soc_dbg`: opentitan/hw/top_darjeeling/sw/autogen/chip/top_darjeeling_soc_dbg.rs does not exist
```

jwnrt:
Can we revert and try again?

It looks like the `mod.rs` in `hw/top_darjeeling/sw/autogen/chip/mod.rs` expects to find `top_darjeeling_soc_dbg.rs` but this file is actually in `sw/host/opentitanlib/src/chip/autogen/top_darjeeling_soc_dbg.rs` and not included in the `mod.rs` there.

vogelpi:
I'll give it a try. Thanks @jwnrt for flagging.

util/topgen.py
```diff
@@ -731,29 +731,33 @@ def render_template(template_path: str, rendered_path: Path, **other_info):
...
                         helper=rs_helper)
 
-    # Generate Rust host-side files
-    rsformat_dir = src_tree_top / "sw/host/opentitanlib/src/chip/autogen"
-    rsformat_dir.mkdir(parents=True, exist_ok=True)
-    render_template(topgen_template_path / "host_toplevel.rs.tpl",
-                    rsformat_dir / f"{topname}.rs",
-                    helper=rs_helper)
+        # Generate Rust host-side files
+        rsformat_dir = src_tree_top / 'sw/host/opentitanlib/src/chip/autogen'
```
vogelpi, line 755: Note that this change of `rsformat_dir` inside the address space loop also affects the generation of the following device-side files which is wrong. I've now filed #25953 to fix this and to unblock CI again.

a-will: Nice catch! I didn't pay close enough attention to the rust outputs, apparently. That code could be cleaned up a little more...

vogelpi: Don't worry, it's working now :-)

Issue #25891, opened 2025-01-15
[dv,mem_bkdr_util] Move memory specific code to the corresponding IP


matutem:
### Description

The mem_bkdr_util class includes some specific code that is heavily dependent on the IP. For example, mem_kkdr_util.sv includes mem_bkdr_util__flash.sv (and similar files for sram and rom). This is undesirable since it makes the mem_bkdr_util less generic and can cause trouble for code that ends up being top-specific. This was the case for the otp_ctrl code, and it created problems for multi-top.

The solution for otp_ctrl is to move the code to hw/ip_templates/otp_ctrl/dv/env and create a separate core file for it, so chip_level environments can use it. The functions in that code used to be member functions of mem_bkdr_util, but they were simply change to package level functions that take a mem_bkdr_util handle (in this case, the handle to the otp_ctrl instance).

This code needs to be a template since the otp_ctrl is ipgen and the functions can access specific partitions.

It is quite possible the same solution could work for other types of memories.

Some tasks:
- [ ] Make this code a template since darjeeling has different parameters, traked by #25883

Move other memory specific code out of mem_bkdr_util using a similar mechanism:
- [ ] sram_ctrl
- [ ] rom_ctrl
- [ ] flash_ctrl

vogelpi:
Thanks @matutem for creating the issue and for working on this. @rswarbrick filed an issue specific for OTP here: #25883. In the meantime, we've also merged a hotfix to unblock things. But your proper fix in #25895 is still required (currently fails CI).

rswarbrick:
As far as I can tell, this was implemented the PR mentioned above (#25895), plus some more extensive work from @alees24. I don't believe there is any need to track it further in this issue.

PR #25884, opened 2025-01-15
[mem_bkdr_util] Temporarily fix build for an otp_ctrl parameter

rswarbrick:
This fixes the build again (currently broken!), leaving a TODO note to make sure it gets wired up properly for other top-levels.

@Razer6 fyi. Issue #25883 tracks needing a "proper" fix.

Fix #25860

PR #25895, opened 2025-01-16
[dv,mem_bkdr_util] Move otp specific code to otp_ctrl

matutem:
Change the code to take a handle to the OTP mem_bkdr_util instance,
and use that to perform plain reads and writes to the underlying
memory.

The otp_ctrl code in mem_bkdr_util was making this utility to become
top-specific, which is wrong for a generic utility, and caused build
failures.

Change the uses of the affected functions in top_earlgrey dv code.

The main change is the move of hw/dv/sv/mem_bkdr_util/mem_bkdr_util__otp.sv
to hw/ip_templates/otp_ctrl/dv/ev/otp_ctrl_mem_bkdr_util_pkg.sv, and
changing the functions to take a mem_bkdr_util handle. For some reason
git didn't detect this as a move.

Part of #25891


vogelpi:
Thanks for picking this up @matutem !

The PR currently fails CI (AscentLint and Run DV smoke test) and that's why I've merged @rswarbrick 's hotfix in the meantime (see #25884). You'll have to rebase on top of that.

hw/top_darjeeling/ip_autogen/otp_ctrl/dv/env/otp_ctrl_mem_bkdr_util.core
```diff
@@ -0,0 +1,25 @@
...
+virtual:
+  - lowrisc:ip_interfaces:otp_ctrl_mem_bkdr_util
+
+filesets:
+  files_dv:
+    depend:
+      - lowrisc:dv:mem_bkdr_util
+      - lowrisc:dv:crypto_dpi_present
+      - lowrisc:dv:lc_ctrl_dv_utils
+      - lowrisc:opentitan:top_darjeeling_otp_ctrl_top_specific_pkg:1.0
```
vogelpi, line 16: Question: now that we use a virtual core for this and that we have `top_darjeeling` in the core name, should we remove `_top_specific` from the name?

hw/top_darjeeling/ip_autogen/otp_ctrl/dv/env/otp_ctrl_mem_bkdr_util_pkg.sv
```diff
@@ -0,0 +1,221 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// This package has functions that perform mem_bkdr_util accesses to different partitions in OTP.
+// The functions take a handle to the otp_ctrl mem_bkdr_util class for reads and writes.
```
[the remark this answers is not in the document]
matutem: We create an an instance of the mem_bkdr_util class for each memory. The updated wording should be clearer.

hw/top_darjeeling/ip_autogen/otp_ctrl/dv/env/otp_ctrl_mem_bkdr_util.core
matutem: This is a bit verbose but we need to make a distinction between the otp_ctrl_pkg, which is generic and can be used by any IP, and the package that holds top specific stuff. With the existing VLNV convention if we name this "specific_otp_ctrl_pkg", the resulting name would be "top_<foo>_specific_otp_ctrl_pkg" which avoids repeating top, but when we change our VLNV convention this would be ...:<topname>_dv:specific_otp_ctrl_pkg, which is vague. We need to describe "top_specific" or "non_generic".

This reminded me to create an issue for this: https://github.com/lowRISC/opentitan/issues/25921.

vogelpi: Thanks for the explanation :-)

a-will:
For a future fix, but I'd recommend against any use of a virtual VLNV here. The files contained herein do not represent a virtual interface, and I suspect that ultimately, no ipgen core should be exposing virtual VLNVs. It prevents multiple specializations.

The cores that use it should always be able to bring in the dependency directly (instead of virtually).

matutem:
@a-will to be clear, this means you think invoking instance_vlnv is fine, but we don't need to declare this virtual since only templated cores should depend on this?

a-will:
> @a-will to be clear, this means you think invoking instance_vlnv is fine, but we don't need to declare this virtual since only templated cores should depend on this?

Yup, that's right! 😄

PR #25897, opened 2025-01-16
[hw,otp_ctrl,rtl] Provide a DFT config and response port

Razer6:
This port can be overwritten by the integrator with their tech lib. This allows to place custom DFT and response signals. In the future, existing status control and monitor ports should be incorporated to the cfg/cfg_rsp ports.


hw/ip_templates/otp_ctrl/rtl/otp_ctrl.sv
```diff
@@ -86,7 +86,10 @@ module otp_ctrl
   input prim_mubi_pkg::mubi4_t                       scanmode_i,
   // Test-related GPIO output
   output logic [OtpTestVectWidth-1:0]                cio_test_o,
-  output logic [OtpTestVectWidth-1:0]                cio_test_en_o
+  output logic [OtpTestVectWidth-1:0]                cio_test_en_o,
+  // DFT Cfg and response port
```
vogelpi, line 90: Nit: want to write out Cfg -> config?

hw/ip_templates/otp_ctrl/data/otp_ctrl.hjson.tpl
```diff
@@ -444,6 +444,21 @@ otp_size_as_uint32 = otp_size_as_bytes // 4
       package: ""
       desc:    "AST observability bus."
     }
+    // DFT CFG and RSP siognals
```
vogelpi, line 447: Nit: siognals -> signals

vogelpi (approved):
This looks good to me, thanks @Razer6 .

As you noted in an offline discussion, there are already the `test_xyz` signals for OTP. AFAIK, this are mostly used for the vendor test partition but I don't have any visibility into this so can't say for sure how this interplays with DFT. Adding a separate interface as you did now and inline with what we do for the SRAMs is very reasonable in my view. If really required, one could still combine the two interfaces later on.

@meisnere, @sha-ron , please note that you will have to add the same ports to your OTP macro wrapper.

vogelpi:
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_otp_cfg_pkg.sv

This PR adds a new interface to the OTP controller and the generic macro. Earlgrey functionality is not altered but partners will have to align their macro wrappers.

a-will (approved):
LGTM

Someday, I think it'd be good to evict the macros from the controllers, though. The dependency of a so-called "prim" on ipgen cores seems backwards, and the controller's `prim` interface's CSR definitions don't make sense to any but one integrator.

a-will:
CHANGE AUTHORIZED: hw/ip/prim_generic/rtl/prim_otp_cfg_pkg.sv

hw/ip_templates/otp_ctrl/rtl/otp_ctrl.sv
Razer6: hehe, I had it even in the uncommited stash but didn't add it to the commit 🥲

Razer6:
> @meisnere, @sha-ron , please note that you will have to add the same ports to your OTP macro wrapper.

Similar changes where added to the SRAMs.

Razer6:
> Someday, I think it'd be good to evict the macros from the controllers, though. The dependency of a so-called "prim" on ipgen cores seems backwards, and the controller's `prim` interface's CSR definitions don't make sense to any but one integrator.

Yes, this definitively comes from a first product and requires some generalization. But an adventure for a different time :)

PR #25899, opened 2025-01-16
[hmac,dv] Reset S&R skip context

martin-velay:
- reinitialize the Save and Restore skip context flag each time the HMAC is in reset.


rswarbrick (commented):
This looks reasonable to me, but I'm slightly surprised to see it in the base vseq when the "client" is actually the scoreboard. Shouldn't this be in the monitor?

martin-velay:
Both SCB and vseq need this signal, but that's a temporary thing as mentioned in a TODO #23562

rswarbrick:
Ahah, that makes sense. Thanks for explaining it (and I'm glad the tidy-up is being tracked).

rswarbrick (approved):

PR #25901, opened 2025-01-16
[hmac,dv] Reduce long_msg test duration

martin-velay:
- this test was taking too long to run and can cause a timeout, the max number or transactions has been divided by 2.


rswarbrick (approved):
This looks sensible to me. Thanks for explaining things in the commit message too.

PR #25902, opened 2025-01-16
[hmac,dv] Fix SCB fifo_full_detected signal

martin-velay:
- fifo_full_detected was going low and high without any change on the FIFO level. This was causing the SCB to generate a fifo_empty_intr when it was not expected.

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_scoreboard.sv
+++ b/hw/ip/hmac/dv/env/hmac_scoreboard.sv
@@ -29,6 +29,7 @@ class hmac_scoreboard extends cip_base_scoreboard #(.CFG_T (hmac_env_cfg),
   bit             hmac_stopped_last;
   bit             hmac_start_last;
   bit             hmac_continue_last;
+  bit             hmac_fifo_full_last;
   bit [TL_DW-1:0] intr_test;
   bit [TL_DW-1:0] intr_state_exp;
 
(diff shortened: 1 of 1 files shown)


hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -625,16 +626,20 @@ task hmac_scoreboard::hmac_process_fifo_status();
     fork begin
       // Store hmac_process and hmac_start to be able to detect when it goes up, as it could remain
       // up for a while
+      bit hmac_fifo_full_posedge;
       bit hmac_process_posedge  = hmac_process & ~hmac_process_last;
       bit hmac_start_posedge    = hmac_start & ~hmac_start_last;
       bit hmac_stopped_posedge  = hmac_stopped & ~hmac_stopped_last;
       bit hmac_continue_posedge = hmac_continue & ~hmac_continue_last;
+      bit hmac_any_posedge      = hmac_process_posedge || hmac_start_posedge ||
+                                  hmac_stopped_posedge || hmac_continue_posedge;
```
rswarbrick, line 635: I'm convinced by the code, but the naming is a bit confusing. Maybe something like `any_posedge_but_full` ? :-)

Or can we just drop the signal? I'm not sure I know why it's worth adding.

martin-velay: Yes, you're right. I've dropped it.

martin-velay:
I have made another little change, as my fix was not entirely working. I think it'll be alright now

PR #25903, opened 2025-01-16
[hmac,dv] Reset intr_test signal

martin-velay:
- Reset expected intr_test signal to 0 when a reset is triggered, as this was obviously causing the SCB comparison to fail.


rswarbrick (approved):
I'm slightly surprised by this, since the register is write-only. Looking at the rest of the scoreboard, this seems to be used for something other than mirroring a register value. Maybe it makes sense to follow up with a commit that renames it to something that isn't a register name. Something like `last_intr_test_write` or similar?

Anyway, I think I understand why this commit is going to do the right thing, so let's get it in in the meantime.

PR #25905, opened 2025-01-16
[tlul,rtl] Avoid an unnecessary wready signal in tlul_adapter_sram

rswarbrick:
When DataXorAddr is true, there are two fifos that hold SRAM requests (u_sramreqfifo and u_sramreqaddrfifo). These will always push and pop in sync (using the same signals in fact!).

As such, checking both wready signals is unnecessary. Drop one, avoiding an impossible conditional coverage item.

### DIFF
--- a/hw/ip/tlul/rtl/tlul_adapter_sram.sv
+++ b/hw/ip/tlul/rtl/tlul_adapter_sram.sv
@@ -262,7 +262,6 @@ module tlul_adapter_sram
   logic sramreqfifo_rready;
   sram_req_t sramreqfifo_wdata, sramreqfifo_rdata;
 
-  logic sramreqaddrfifo_wready;
   logic [SramBusBankAW-1:0] sramreqaddrfifo_wdata, sramreqaddrfifo_rdata;
 
   logic rspfifo_wvalid, rspfifo_wready;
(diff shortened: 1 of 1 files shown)


alees24 (approved):
The reasoning/logic is sound. Personally I'd prefer to see a single FIFO of parameterized-width in this case, to gain the area advantages that come from having a wider FIFO and reduced overhead.

vogelpi (approved):
LGTM, thanks @rswarbrick !

@alees24 I see your point, the challenge here is that one of the two FIFOs is only instantiated if the `DataXorAddr` parameter is true while the other one is always there.

vogelpi:
I've also asked @nasahlpa for a review (he did the implementation for the FI countermeasure). Pascal, my understanding is that removing the second wready is of no importance for the FI hardening right?

nasahlpa (approved):
LGTM!

Yes, there is no reason for checking both `wready` signals.

vogelpi:

This RTL change doesn't alter the behavior of the RTL (also not in the case of FI) but it removes a coverage hole.

alees24:

rswarbrick:
Thanks for the reviews. I think @alees24 is (sensibly) pointing out that if you have some items of width 2N then it's more efficient to use a FIFO of width 2N instead of two fifos of width N.

And I strongly agree! As such, this adapter could be improved quite a bit further (and waste less area). But I still stand by my change :-)

rswarbrick:
Grr. That failed CI because Ascentlint correctly noticed there was no longer anything looking at `sramreqaddrfifo_wready` (except the assertion). I've just pushed a version that feeds it into an "unused" signal.

vogelpi:
> Thanks for the reviews. I think @alees24 is (sensibly) pointing out that if you have some items of width 2N then it's more efficient to use a FIFO of width 2N instead of two fifos of width N.
> 
> And I strongly agree! As such, this adapter could be improved quite a bit further (and waste less area). But I still stand by my change :-)

Ah, yes, that's absolutely right, good point @alees24 !

rswarbrick:
Merging despite the failing lint job. The previous version passed everything except AscentLint. This version passes that check and is failing because the quick lint job is broken.

PR #25906, opened 2025-01-16
[hw,ac_range_check,rtl] Add RACL checks to the AC Range Checks

Razer6:
This PR adds RACL checks to the AC ranges. Now that RACL is (partly) merged, we can enable the RACL checks first.

RACL error logging is not yet fully implemented as it depends on determining the failing range.

### DIFF
--- a/hw/ip_templates/ac_range_check/data/ac_range_check.hjson.tpl
+++ b/hw/ip_templates/ac_range_check/data/ac_range_check.hjson.tpl
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 //
+<%
+import math
+%>\
 # AC Range Check register template
 {
   name:               "ac_range_check"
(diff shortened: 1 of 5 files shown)


hw/ip_templates/ac_range_check/data/ac_range_check.hjson.tpl
```diff
@@ -122,12 +125,17 @@
       swaccess: "ro"
       hwaccess: "hwo"
       fields: [
-        { bits: "22:18"
+        { bits: "${14+nr_role_bits+nr_ctn_uid_bits+math.ceil(math.log(num_ranges, 2))-1}:${14+nr_role_bits+nr_ctn_uid_bits}"
```
rswarbrick, line 128: This is a bit of a mouthful. Could you do this by defining some variables at the start, based on the template parameters. Maybe something like this?
```python
denied_ctn_uid_lsb = 14 + nr_role_bits
deny_range_index_lsb = denied_ctn_uid_lsb + nr_ctn_uid_bits
deny_range_index_size = math.ceil(math.log(num_ranges, 2))
```

hw/ip_templates/ac_range_check/rtl/ac_range_check.sv.tpl
```diff
@@ -101,14 +113,21 @@ module ${module_instance_name}
     assign addr_hit[i] = prim_mubi_pkg::mubi4_test_true_loose(reg2hw.range_perm[i].enable.q) &
                          tor_hit;
 
+    // Perform RACL checks - check if the incoming role matches with the configured policy
+    assign racl_read_hit [i] =  |(racl_role_vec & reg2hw.range_racl_policy_shadowed[i].read_perm.q)
```
rswarbrick, line 117: Nit: There's an extra space here! Probably drop one before the `|` and keep the one before the `[`?

Razer6:
@rswarbrick Thanks for looking into this. I fixed the nit and made the bit calculation more readable. Can you take another look?

vogelpi (approved):
LGTM, but you would need to rebase this on master @Razer6 .

Razer6:
> LGTM, but you would need to rebase this on master @Razer6 .

Thanks for look into this. I just rebased this PR on top of master.

PR #25908, opened 2025-01-16
[multitop] Add better support for linker scripts

pamaury:
Alternative approach to #25785 that preprocesses linker scripts instead of using bazel templates.


jwnrt (approved):
I've gone over the Bazel part a couple of times and it looks good. Nice to see the `get_tool_for_action` used. (aside, I realised that I missed this for the other part of our repo that uses the preprocessor here: [rules/ujson.bzl#L40](https://github.com/lowRISC/opentitan/blob/master/rules/ujson.bzl#L40)).

I think this is better than using Bazel substitution as discussed in the other PR.

LGTM, thanks

rules/linker.bzl
```diff
@@ -4,12 +4,72 @@
...
+        outputs = [output_script],
+        inputs = depset(
+            [ctx.file.script],
+            transitive = [compilation_context.headers, cc_toolchain.all_files],
+        ),
+        executable = cxx_compiler_path,
+        arguments = [
+            "-E",  # Preprocess only.
+            "-P",  # Avoid line markers in output.
+            "-xc",  # Force C language.
```
a-will, line 57: Might also want a `-C` in there too, so anything that looks like comments will not be removed.

rules/linker.bzl
```diff
@@ -20,31 +80,41 @@ def _ld_library_impl(ctx):
...
-        user_link_flags += [
-            "-Wl,-L,{}".format(include.dirname)
-            for include in ctx.files.includes
-        ]
+    files += ctx.files.includes
+    user_link_flags += [
+        "-Wl,-L,{}".format(include.dirname)
+        for include in ctx.files.includes
+    ]
+    print(user_link_flags)
```
a-will, line 88: nit: Debug output to be removed?

rules/linker.bzl
pamaury: Good idea

rules/linker.bzl
pamaury: Yes

pamaury:
One thing to note about this PR which may be confusing: it gives two ways to include files: `INCLUDE` and `#include`. The files which are `#include`-ed are obviously preprocessed but the files which are `INCLUDE`-ed are not. Maybe we should make that clear in the documentation (it's technically there, but might not be obvious).

a-will (approved):

engdoreis (approved):
I prefer this approach over the template.

hw/top/defs.bzl
```diff
@@ -52,3 +54,40 @@ def opentitan_require_ip(ip):
     )
     """
     return opentitan_if_ip(ip, [], ["@platforms//:incompatible"])
+
+def opentitan_select_top(values, default):
+    """
+    Select a value based on the top. If not top matches, a default
```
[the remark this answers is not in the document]
pamaury: It should be "no" instead of "any".

sw/device/lib/testing/test_framework/ottf_silicon_creator_a.ld
```diff
@@ -10,7 +10,7 @@
  * This linker script generates a binary to run rom.
  */
 
-INCLUDE hw/top_earlgrey/sw/autogen/top_earlgrey_memory.ld
+INCLUDE OPENTITAN_TOP_MEMORY_LD
```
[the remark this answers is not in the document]
pamaury: Will do

rules/linker.bzl
```diff
@@ -4,12 +4,73 @@
...
+        inputs = depset(
+            [ctx.file.script],
+            transitive = [compilation_context.headers, cc_toolchain.all_files],
+        ),
+        executable = cxx_compiler_path,
+        arguments = [
+            "-E",  # Preprocess only.
+            "-P",  # Avoid line markers in output.
+            "-C",  # Keep comments
+            "-xc",  # Force C language.
```
cfrantz, line 58: Does this mean that C++ comments (`// foo` ) are not permitted?

Considering that is already the case for assembly language files, its probably not an issue.

cfrantz (approved):

rules/linker.bzl
pamaury: Yes only C comments are supported. I think this is more consistent because linker scripts only allow `/* */` for comments when they are not preprocessed.

a-will: They should pass through the preprocessor with `-C`, so you don't _have_ to maintain that restriction, I think. Doesn't matter to me, though!

PR #25909, opened 2025-01-16
[bazel] Add QEMU execution environment

jwnrt:
### Summary

This PR adds a `sim_qemu_rom_with_fake_keys` execution environment and CI job.

QEMU comes from the in-development Earl Grey branch of our fork: https://github.com/lowRISC/qemu/tree/dev/ot-earlgrey-1.0.0-updates

In the future when Earl Grey and Darjeeling coexist in the same QEMU binary, this execution environment should support both tops in Bazel.

### Execution environment

The exec env currently runs `qemu-system-riscv32` directly. A future PR will add QEMU support to `opentitantool`. Until then, the exec env does not support tests that require custom harnesses and the test won't quit when `FAIL!` is printed, it will have to timeout.

`qemu_params` supports the `icount: <int>` and `globals: { <key>: <val> }` parameters for changing the Ibex clock scaling factor and global parameter values respectively.

### File formats and python scripts

QEMU requires specific files for the flash image, OTP, and an INI describing configuration parameters for the top. These are generated with `flashgen.py`, `otptool.py`, and `cfggen.py` scripts pulled from the QEMU release.

These scripts require Python 3.12, however we don't want to update everything in the repo from 3.9 just yet. I have updated `rules_python` to the latest version 1.1.0 which allows us to add `python_version = "3.12"` to specific binaries. This caused some problems with aspects, see the commit message for details.

### CI tests

I have added a CI job for `sim_qemu` tests and added `sim_qemu_rom_with_fake_keys` to three. Two of these require specific `-global` settings, so it's good to know that works. After some unrelated changes to QEMU we can also show off a couple of tests which require different `icount` values to pass.

EDIT: we don't want to block OpenTitan CI on QEMU failures in case the RTL here changes but QEMU does not. I have made the CI non-blocking.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -733,3 +733,19 @@ jobs:
       - name: Check for unrunnable tests
         run: ./ci/scripts/check-unrunnable-tests.sh
         continue-on-error: true
+
+  qemu_smoketest:
+    name: QEMU smoketest
+    runs-on: ubuntu-22.04-vivado
+    needs: quick_lint
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Prepare environment
+        uses: ./.github/actions/prepare-env
+        with:
+          service_account_json: '${{ secrets.BAZEL_CACHE_CREDS }}'
+      - name: Execute QEMU smoketest
+        run: |
+          ./bazelisk.sh test //sw/device/tests:rom_exit_immediately_sim_qemu_base
--- a/rules/opentitan/cc.bzl
+++ b/rules/opentitan/cc.bzl
@@ -416,9 +416,13 @@ def _opentitan_test(ctx):
         p = None
 
     executable, runfiles = exec_env.test_dispatch(ctx, exec_env, p)
+    if ctx.attr.test_harness != None:
+        harness_runfiles = ctx.attr.test_harness[DefaultInfo].default_runfiles
+    else:
+        harness_runfiles = ctx.runfiles()
     return DefaultInfo(
         executable = executable,
-        runfiles = ctx.runfiles(files = runfiles),
+        runfiles = ctx.runfiles(files = runfiles).merge_all([harness_runfiles]),
     )
 
 opentitan_test = rv_rule(
@@ -431,6 +435,7 @@ opentitan_test = rv_rule(
         "test_harness": attr.label(
             executable = True,
             cfg = "exec",
+            allow_files = True,
         ),
         "binaries": attr.label_keyed_string_dict(
             allow_files = True,
--- a/rules/opentitan/defs.bzl
+++ b/rules/opentitan/defs.bzl
@@ -109,6 +109,7 @@ EARLGREY_TEST_ENVS = {
     "//hw/top_earlgrey:fpga_cw310_rom_with_fake_keys": None,
     "//hw/top_earlgrey:sim_dv": None,
     "//hw/top_earlgrey:sim_verilator": None,
+    "//hw/top_earlgrey:sim_qemu_rom_with_fake_keys": None,
 }
 
 # The default set of test environments for Earlgrey.
--- a/rules/opentitan/defs.bzl
+++ b/rules/opentitan/defs.bzl
@@ -55,6 +55,11 @@ load(
     _sim_verilator = "sim_verilator",
     _verilator_params = "verilator_params",
 )
+load(
+    "@lowrisc_opentitan//rules/opentitan:qemu.bzl",
+    _qemu_params = "qemu_params",
+    _sim_qemu = "sim_qemu",
+)
 
 # The following definition is used to clear the key set in the signing
 # configuration for execution environments (exec_env) and opentitan_test
@@ -83,6 +88,9 @@ verilator_params = _verilator_params
 
 sim_dv = _sim_dv
 dv_params = _dv_params
+
+sim_qemu = _sim_qemu
+qemu_params = _qemu_params
 
 ecdsa_key_for_lc_state = _ecdsa_key_for_lc_state
 ecdsa_key_by_name = _ecdsa_key_by_name
@@ -144,6 +152,8 @@ def _parameter_name(env, pname):
             pname = "dv"
         elif "silicon" in suffix:
             pname = "silicon"
+        elif "qemu" in suffix:
+            pname = "qemu"
         else:
             fail("Unable to identify parameter block name:", env)
     return pname
@@ -160,17 +170,17 @@ def _hacky_tags(env):
         # label name.
         subtag = suffix[5:]
         tags.append(subtag)
-        if subtag.startswith("cw305"):
-            tags.append("cw305")
-        elif subtag.startswith("cw310"):
-            tags.append("cw310")
-        elif subtag.startswith("hyper310"):
-            tags.append("hyper310")
-        elif subtag.startswith("cw340"):
-            tags.append("cw340")
-    if suffix.startswith("silicon"):
-        # We add the entire suffix for silicon exec environments to be able
-        # to filter tests by them. "silicon_creator" and
+
+        # Also add the kind of FPGA by itself (e.g. "hyper310" or "cw340").
+        fpga_kind = subtag.split("_")[0]
+        tags.append(fpga_kind)
+    if suffix.startswith("sim"):
+        # Add the kind of simulator by itself (e.g. "verilator" or "qemu").
+        sim_kind = suffix.split("_")[1]
+        tags.append(sim_kind)
+    if suffix.startswith("silicon") or suffix.startswith("sim"):
+        # We add the entire suffix for silicon and sim exec environments
+        # to be able to filter tests by them. "silicon_creator" and
         # "silicon_owner_sival_rom_ext" have different target configurations.
         tags.append(suffix)
     return tags
@@ -195,6 +205,7 @@ def opentitan_test(
         dv = _dv_params(),
         silicon = _silicon_params(),
         verilator = _verilator_params(),
+        qemu = _qemu_params(),
         data = [],
         run_in_ci = None,
         **kwargs):
@@ -224,6 +235,7 @@ def opentitan_test(
       dv: Execution overrides for a DV-based test.
       silicon: Execution overrides for a silicon-based test.
       verilator: Execution overrides for a verilator-based test.
+      qemu: Execution overrides for a QEUM-based test.
       run_in_ci: Override the automatic selection of execution environments to run in CI and run exactly those environments.
       kwargs: Additional execution overrides identified by the `exec_env` dict.
     """
@@ -233,6 +245,7 @@ def opentitan_test(
         "dv": dv,
         "silicon": silicon,
         "verilator": verilator,
+        "qemu": qemu,
     }
     test_parameters.update(kwargs)
     kwargs_unused = kwargs.keys()
--- a/rules/opentitan/qemu.bzl
+++ b/rules/opentitan/qemu.bzl
@@ -1,0 +1,403 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+load(
+    "@lowrisc_opentitan//rules/opentitan:providers.bzl",
+    "SimQemuBinaryInfo",
+)
+load(
+    "@lowrisc_opentitan//rules/opentitan:util.bzl",
+    "get_fallback",
+    "get_override",
+)
+load(
+    "//rules/opentitan:exec_env.bzl",
+    "ExecEnvInfo",
+    "common_test_setup",
+    "exec_env_as_dict",
+    "exec_env_common_attrs",
+)
+load("//rules/opentitan:toolchain.bzl", "LOCALTOOLS_TOOLCHAIN")
+
+_TEST_SCRIPT = """#!/bin/bash
+set -e
+
+# QEMU requires mutable flash and OTP files but Bazel only provides RO
+# files so we have to create copies unique to this test run.
+cp {otp} {mutable_otp} && chmod +w {mutable_otp}
+if [ ! -z {flash} ]; then
+    cp {flash} {mutable_flash} && chmod +w {mutable_flash}
+fi
+
+echo Invoking test: {test_harness} {args} "$@"
+{test_harness} {args} "$@"
+"""
+
+def qemu_params(
+        tags = [],
+        timeout = "short",
+        local = True,
+        test_harness = "//rules/scripts:qemu_pass",
+        binaries = {},
+        rom = None,
+        rom_ext = None,
+        otp = None,
+        bitstream = None,
+        test_cmd = "",
+        data = [],
+        defines = [],
+        icount = 6,
+        globals = {},
+        **kwargs):
+    extra_params = {
+        "icount": str(icount),
+        # We have to stringify this dictionary here because `_opentitan_test` only accepts
+        # a dict with string values, not more dicts.
+        "globals": json.encode(globals),
+    }
+
+    return struct(
+        tags = tags,
+        timeout = timeout,
+        local = local,
+        test_harness = test_harness,
+        binaries = binaries,
+        rom = rom,
+        rom_ext = rom_ext,
+        otp = otp,
+        bitstream = bitstream,
+        test_cmd = test_cmd,
+        data = data,
+        param = kwargs | extra_params,
+        defines = defines,
+    )
+
+def gen_cfg(ctx, **kwargs):
+    """Generate a QEMU `readconfig` INI file containing OpenTitan RTL secrets"""
+    name = get_override(ctx, "label.name", kwargs)
+    cfggen = get_override(ctx, "executable.cfggen", kwargs)
+
+    otp_sv = get_override(ctx, "file.otp_sv", kwargs)
+    lc_sv = get_override(ctx, "file.lc_sv", kwargs)
+    top_hjson = get_override(ctx, "file.top_hjson", kwargs)
+
+    top_name = get_override(ctx, "attr.top_name", kwargs)
+
+    out = ctx.actions.declare_file(name + ".ini")
+
+    ctx.actions.run(
+        inputs = [otp_sv, lc_sv, top_hjson],
+        outputs = [out],
+        executable = cfggen[DefaultInfo].files_to_run,
+        arguments = [
+            "--out",
+            out.path,
+            "--top",
+            top_name,
+            "--topcfg",
+            top_hjson.path,
+            "--otpconst",
+            otp_sv.path,
+            "--lifecycle",
+            lc_sv.path,
+            ".",
+        ],
+        mnemonic = "CfgGen",
+    )
+    return out
+
+qemu_cfg = rule(
+    implementation = lambda ctx: [DefaultInfo(files = depset([gen_cfg(ctx)]))],
+    attrs = {
+        "top_name": attr.string(),
+        "top_hjson": attr.label(allow_single_file = True),
+        "otp_sv": attr.label(allow_single_file = True),
+        "lc_sv": attr.label(allow_single_file = True),
+        "cfggen": attr.label(
+            executable = True,
+            cfg = "exec",
+            allow_files = True,
+            default = Label("//third_party/qemu:cfggen"),
+        ),
+    },
+)
+
+def gen_otp(ctx, **kwargs):
+    """Generate a QEMU-compatible OTP image"""
+    name = get_override(ctx, "label.name", kwargs)
+    out = ctx.actions.declare_file(name + ".raw")
+
+    vmem = get_override(ctx, "file.vmem", kwargs)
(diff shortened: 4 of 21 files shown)


a-will:
How is this going to play out with PRs that bring in hardware changes? Doesn't QEMU need a fixed hardware target, and won't it always be trailing behind this repo?

This does not seem like the correct repo for QEMU tests. I feel like we ought to up our game and make this repo behave as a dependency of some other repo that gathers the supported hashes and runs tests.

moidx:
I think it is too cumbersome to require QEMU passing in the main branch. It should be ok to have it available on release branches where the RTL is not expected to change. 

I am not opposed to the approach @a-will is proposing. Perhaps this is something that can be discussed in the software WG meeting? 

At the same time, I am not opposed to providing some generic infrastructure to connect emulators.

jwnrt:
Yes, this definitely shouldn’t block CI, I’ll change that.

The motivation for having the CI here is that we can more easily see which changes need propagating to QEMU faster. There’s an argument for having it in the QEMU fork instead so that we can see tests pass as we fix the emulation. Maybe we run the tests in both repos but make it non blocking (or a nightly) here. They’re not too expensive to run.

good suggestion, maybe we can discuss this in the WG.

rivos-eblot:
> This PR adds a `sim_qemu_rom_with_fake_keys` execution environment and CI job.

I'm not really familiar with the execution environments in OT, but what is the "real" HW that it targets? QEMU emulates an existing target, up to now it was `fpga_cw310`, which defines both the chip and (some of) the board peripherals such as the SPI data flash. How does `sim_qemu` relate to `fpga_cw310`? This target has been initially chosen since no real chip existed, and we wanted to be sure to run the exact same binaries (ROMs, apps, ...) on the HW and on QEMU. At some point, I guess it would make sense to have a QEMU machine that emulates a real board with a real chip.

Maybe `sim_qemu` is required for dealing with Bazel build environment, but I think that the name should nevertheless keep the actual HW target, i.e. there could be cw310, cw340, real chip, ... machines with different capabilities and memory maps supported within the same QEMU binary.

> These scripts require Python 3.12

They should run fine with Python 3.9. Python [vermin](https://github.com/netromdk/vermin) reports two false positives for 3.11 (`log.py`) and 3.12 (`misc.py`). I may have miss some files though, please let me know so we can fix this.

>  the test won't quit when FAIL! is printed, it will have to timeout.

I read that some tests check the [FPGA_INFO](https://opentitan.org/book/hw/ip/rv_core_ibex/doc/registers.html#fpga_info) register. QEMU exposes a special value in this register - so that it is possible from the guest code to know it runs on QEMU. From there, it is possible to trigger an explicit QEMU termination.

jwnrt:
> I'm not really familiar with the execution environments in OT, but what is the "real" HW that it targets? QEMU emulates an existing target, up to now it was fpga_cw310

The QEMU device header file was copied from the CW310 one so it will report the clock speeds etc that I think QEMU was designed to emulate.

Some of our tests switch on the current platform and the CW310 special casing won’t trigger for QEMU, so there will be differences.

Long term I wonder if QEMU should only target the silicon. This would avoid “approximating an approximation” in which the QEMU emulation inherits shortcomings of the CW310 simulation.

> there could be cw310, cw340, real chip, ... machines with different capabilities and memory maps supported within the same QEMU binary.

This is interesting but I wonder how valuable it is. The differences between the FPGAs and silicon that I’m aware of are the clock speeds, the lack of KMAC masking on CW310, and the number of available pinouts. Would we gain coverage by emulating these limitations? I suppose it makes it easier to compare against the FPGAs whilst pre-silicon.

> I read that some tests check the [FPGA_INFO](https://opentitan.org/book/hw/ip/rv_core_ibex/doc/registers.html#fpga_info) register. QEMU exposes a special value in this register - so that it is possible from the guest code to know it runs on QEMU. From there, it is possible to trigger an explicit QEMU termination.

Interesting, I didn’t actually check any failing tests yet but maybe the ones using this already work. It might be useful even when `FAIL!` detection is working.

This reminds me that I also need to check whether QEMU exiting from a fatal reset or similar will cause the test to fail. I don’t know whether my backgrounding in the test wrapper script will mask that.

rivos-eblot:
> The QEMU device header file was copied from the CW310 one so it will report the clock speeds etc that I think QEMU was designed to emulate.

Ok. I think it is useful to be able to run the very same binary on QEMU and on the platform it emulates. If the binary hardcodes those values (i.e. they cannot be retrieved from the HW), and this is not tracked in the target name, this needs to be tracked in some docs. It really depends on how/why you want to use QEMU.

> Long term I wonder if QEMU should only target the silicon. This would avoid “approximating an approximation” in which the QEMU emulation inherits shortcomings of the CW310 simulation.

I guess the real silicon should be the primary target. If the memory map and features (ignoring some FPGA limitations) are the same, a single target would be needed for the EarlGrey version. I've been used to have FPGA memory maps that differ from real chip ones. From what you wrote, memory maps are all the same with OT platforms and my comment is therefore not really relevant. For the integrated versions (Darjeeling, ...), I'm not sure what is planned.

> This is interesting but I wonder how valuable it is. The differences between the FPGAs and silicon that I’m aware of are the clock speeds, the lack of KMAC masking on CW310, and the number of available pinouts.

Are there no other differences, for example on which external pins are connected the peripherals (SPI, I2C, ...) ?

> Would we gain coverage by emulating these limitations?

I guess it all boils down to what QEMU is used for. If it's only for CI, I don't think there is much value to emulate different targets you're right. If `sim_qemu_rom_with_fake_keys` only aims at running CI, the name is fine. If it used for example to track down a SW bug (using the introspection QEMU can provide on how the HW is used by the SW), the answer may be different.

> Interesting, I didn’t actually check any failing tests yet but maybe the ones using this already work. It might be useful even when `FAIL!` detection is working.

It can also be useful to report an early exception (before the UART is available/configured or in case there is no UART) for example.

jwnrt:
> Are there no other differences, for example on which external pins are connected the peripherals (SPI, I2C, ...)?

I'm not sure actually. I think they might be different because we ran out of pins on the CW310 to support both the PMODs and the HyperDebug at the same time, though maybe this was a limitation of all platforms.

jwnrt:
> I guess it all boils down to what QEMU is used for. If it's only for CI, I don't think there is much value to emulate different targets you're right. If sim_qemu_rom_with_fake_keys only aims at running CI, the name is fine. If it used for example to track down a SW bug (using the introspection QEMU can provide on how the HW is used by the SW), the answer may be different.

Ah sorry, I'm just usually thinking about CI and SiVal testing which doesn't make sense for an emulation like QEMU.

I guess the primary use case for QEMU is developing software, and the test coverage is only important to know how similar the emulation is to the target. With that view I think you're right that really we should be able to run QEMU with the binaries from any of the other platforms (FPGAs, silicon) just to see how similar they are.

jwnrt:
> > These scripts require Python 3.12
> 
> They should run fine with Python 3.9. Python [vermin](https://github.com/netromdk/vermin) reports two false positives for 3.11 (`log.py`) and 3.12 (`misc.py`). I may have miss some files though, please let me know so we can fix this.

Are you sure `misc.py` is a false positive for 3.12? I get this error when I try to run `cfggen.py` with 3.9:

```
~qemu_opentitan/python/qemu/ot/util/misc.py", line 16, in <module>
    from collections.abc import Buffer
ImportError: cannot import name 'Buffer' from 'collections.abc'
```

Maybe this error can be avoided on the QEMU side but the 3.12 toolchain is working here now so I don't mind supporting it.

rivos-eblot:
> Are you sure `misc.py` is a false positive for 3.12? I get this error when I try to run `cfggen.py` with 3.9:

There is an `ImportError` handler in this case, but I just realize that this handler is using itself a 3.10 syntax :-(
I'll push a patch. Sorry about that.

> Maybe this error can be avoided on the QEMU side but the 3.12 toolchain is working here now so I don't mind supporting it.

Ok. We try to support the oldest [Python version](https://devguide.python.org/versions/) which has not been EOLed so it would be nice to support 3.9.

rivos-eblot:
> Maybe this error can be avoided on the QEMU side but the 3.12 toolchain is working here now so I don't mind supporting it.

See https://github.com/lowRISC/qemu/pull/113/commits/9fdedf99c540df43b0cd9c342ef9c7502bca2d72

third_party/qemu/BUILD
```diff
@@ -0,0 +1,50 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+load("@qemu_python_deps//:requirements.bzl", "requirement")
+load("@rules_python//python:defs.bzl", "py_binary", "py_library")
+
+py_binary(
+    name = "cfggen",
+    srcs = ["@qemu_opentitan//:scripts/opentitan/cfggen.py"],
+    python_version = "3.12",
```
pamaury, line 13: I would suggest to create a variable for this version just locally in the file:
```python
QEMU_PY_VERSION="3.12"
```

rules/opentitan/defs.bzl
```diff
@@ -160,17 +170,17 @@ def _hacky_tags(env):
...
-        elif subtag.startswith("hyper310"):
-            tags.append("hyper310")
-        elif subtag.startswith("cw340"):
-            tags.append("cw340")
-    if suffix.startswith("silicon"):
-        # We add the entire suffix for silicon exec environments to be able
-        # to filter tests by them. "silicon_creator" and
+
+        # Also add the kind of FPGA by itself (e.g. "hyper310" or "cw340").
+        fpga_kind = suffix.split("_")[1]
```
pamaury, line 175: I find this code confusing (it was already confusing before this change). I would suggest to change that to `fpga_kind = subtag.split("_")[0]` to make it clear that the fpga_kind is the first part of the rest. Otherwise it makes the relationship between `subtag` and `fpga_kind` obscure.

third_party/qemu/BUILD.qemu_opentitan.bazel
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+package(default_visibility = ["//visibility:public"])
+
+exports_files(glob(["**"]))
+
+filegroup(
+    name = "ot",
+    srcs = glob([
+        "python/qemu/ot/**",
```
rivos-eblot, line 12: I do not know how bazel works, so maybe it is already declared somewhere else: 
- `ot.otp` depends on  the hjson PyPI module
- `ot.util.elf` depends on the pyelftools PyPI module
- `ot.dtm` depends on the jtag module from `python/qemu/jtag`

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,392 @@
...
+
+def gen_cfg(ctx, **kwargs):
+    name = get_override(ctx, "label.name", kwargs)
+    cfggen = get_override(ctx, "executable.cfggen", kwargs)
+
+    otp_sv = get_override(ctx, "file.otp_sv", kwargs)
+    lc_sv = get_override(ctx, "file.lc_sv", kwargs)
+    top_hjson = get_override(ctx, "file.top_hjson", kwargs)
+
+    top_name = get_override(ctx, "attr.top_name", kwargs)
```
pamaury, line 103: The top name is already part of the top_hjson file, is it really necessary to make that a separate attribute? The `cfggen` tool should be able to extra it from the top configuration file.

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,392 @@
...
+        ),
+        "flashgen": attr.label(
+            executable = True,
+            cfg = "exec",
+            allow_files = True,
+            default = Label("//third_party/qemu:flashgen"),
+        ),
+        "otp_sv": attr.label(
+            allow_single_file = True,
+            # TODO: should we really use Earl Grey as the default?
```
pamaury, line 257: Question I guess for @rivos-eblot: why is the tool taking an SV file as input? I would have expected the OTP mmap or something similar.

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,392 @@
...
+            default = Label("//hw/top_earlgrey/ip_autogen/otp_ctrl:rtl/otp_ctrl_part_pkg.sv"),
+        ),
+        "lc_sv": attr.label(
+            allow_single_file = True,
+            default = Label("//hw/ip/lc_ctrl:rtl/lc_ctrl_state_pkg.sv"),
```
pamaury, line 262: Same here

pamaury:
The code looks good to me overall. I just would not make the QEMU CI tests blocking, I think it's a bit of an open question what to do with them exactly but if I understand correctly, we could make them non-blocking and they would at least be gathered in the grafana dashboard?

third_party/qemu/BUILD.qemu_opentitan.bazel
jwnrt: Those two PyPI two dependencies are added to the `cfggen` and `flashgen` Python binaries in `third_party/qemu/BUILD`, but if they're actually used in the `ot` libraries maybe I should move them.

rules/opentitan/qemu.bzl
rivos-eblot: Last time I dig into the repo, the required variables were not defined in HJSON file, only in the SV file, ex:
```sh
qemu-eg> ./scripts/opentitan/cfggen.py ../opentitan-master | rg lc_state
  lc_state_first = "ee75b407d2314d2ef84185ac8c990f536071632c086d4c924070be92d2948d6228b2711e9b2d8c4d"
  lc_state_last = "ee75fe0ffe7b6f3ffc5f9ffd9ff96fdb7f736f6c9e6fdcd35277fef2d3bdcd6ffbb2f59fdf3fbedd"

qemu-eg> rg -l -i ee75 -g '*.hjson' -g '!*sphincsplus*' ../opentitan-master
```

rivos-eblot: If the same values can be extracted from a HJSON file please let me know, I would be happy to replace my ugly SV parser with a regular HJSON extraction. I may have just failed to locate the proper file(s).

pamaury: I unfortunately don't know too much about the OTP, probably @matutem knows but if you can extra it for the SV file then I think that's ok. I was just curious :)

rules/opentitan/qemu.bzl
rivos-eblot: See https://github.com/lowRISC/opentitan/pull/25909#discussion_r1927340137

rules/opentitan/qemu.bzl
rivos-eblot: We need to generate either "`eg`" for EarlGrey or "`dj`" for Darjeeling, and there is no real way to extract this information from the HJSON file I believe.
I guess we could create a map to convert the top name to these special strings. I'll have a look at it.

rivos-eblot: https://github.com/lowRISC/qemu/pull/117

jwnrt: Oh I've been using `ot-otp-e` so far (not `-eg`) and tests have just been working. I guess they haven't needed these constants yet?

Hopefully I can fix this when the QEMU patch lands.

pamaury (approved):

.github/workflows/ci.yml
```diff
@@ -707,3 +707,29 @@ jobs:
       - name: Check for unrunnable tests
         run: ./ci/scripts/check-unrunnable-tests.sh
         continue-on-error: true
+
+  execute_rom_qemu_tests:
+    name: QEMU ROM Tests
+    needs: quick_lint
+    continue-on-error: true
```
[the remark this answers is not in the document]
jwnrt: Ah okay, so it still showed as red in the status? That's not really what I want. I'll add it to the actual test step.

jwnrt:
The QEMU Python scripts now work with 3.9 so I've dropped my commits that add the Python 3.12 toolchain.

rules/opentitan/qemu.bzl
jwnrt: This is fixed now.

jwnrt:
I've made some changes since the last review:

1. QEMU tests can now be run in ROM position.
2. Tests now use a proper (though still temporary) test harness that checks for `PASS!` and QEMU's exit code.
3. `sim_qemu_rom_with_fake_keys` is now added to every test by defualt.
4. QEMU tests are no longer run in CI, even non-blocking, except:
5. There's now a test which runs in ROM position and exits immediately.
   * This is run **blocking** in CI so that we don't break the QEMU execution environment. 
   * It could break if there's an RTL change which moves the `rv_core_ibex.sw_fatal_err` changes base address or offset, which seems unlikely but could happen.

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,389 @@
...
+        data = [],
+        defines = [],
+        icount = 6,
+        globals = {},
+        **kwargs):
+    extra_params = {
+        "icount": str(icount),
+        # We have to stringify this dictionary here because `_opentitan_test` only accepts
+        # a dict with string values, not more dicts.
+        "extra_args": " ".join(["-global {}={}".format(key, val) for (key, val) in globals.items()]),
```
pamaury, line 57: Just FYI, a common way to work around this limitation is to take whatever object you want, serialize it with `json.encode` and then deserialize it with `json.decode` in the rule. Here it's not necessary though.

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,389 @@
...
+        mnemonic = "FlashGen",
+    )
+    return out
+
+qemu_flash = rule(
+    implementation = lambda ctx: [DefaultInfo(files = depset([gen_flash(ctx)]))],
+    attrs = {
+        "app": attr.label(allow_single_file = True),
+        "app_elf": attr.label(allow_single_file = True),
+        "bank": attr.string(default = "0"),
```
pamaury, line 190: this parameter seems to be unused in the implementation

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,389 @@
...
+            out.path,
+        ],
+        mnemonic = "FlashGen",
+    )
+    return out
+
+qemu_flash = rule(
+    implementation = lambda ctx: [DefaultInfo(files = depset([gen_flash(ctx)]))],
+    attrs = {
+        "app": attr.label(allow_single_file = True),
```
pamaury, line 188: Might be good idea to document what `app` is supposed to be (binary? signed binary?).
Also for the attributes here, it seems that it can only generates flash images with a single slot filled?

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,389 @@
...
+
+def _transform(ctx, exec_env, name, elf, binary, signed_bin, disassembly, mapfile):
+    if ctx.attr.kind == "rom":
+        default = elf
+        rom = elf
+    elif ctx.attr.kind == "ram":
+        default = elf
+        rom = None
+    elif ctx.attr.kind == "flash":
+        signed_bin = gen_flash(
```
pamaury, line 261: Maybe the generated flash image should not overwrite the `signed_bin` output. Otherwise it would be impossible to access the raw signed binary when doing bazel queries.

rules/opentitan/qemu.bzl
```diff
@@ -0,0 +1,389 @@
...
+            ctx,
+            flashgen = exec_env.flashgen,
+            app = signed_bin,
```
pamaury, line 264: I think it's possible for signed_bin to be `None` if signing is not enabled (not very likely but you never know what kind of abuse those rules will take), in which case you need to use `binary`. If any of the tools assumes that it is signed, then I would suggest to add an assertion on `signed_bin` not being `None`.

sw/device/silicon_creator/rom/BUILD
```diff
@@ -69,6 +69,11 @@ cc_test(
 
 ld_library(
     name = "linker_script",
+    includes = [
```
pamaury, line 72: why is this necessary? Those are already indirectly imported by the `deps`

rules/opentitan/cc.bzl
```diff
@@ -393,9 +393,12 @@ def _opentitan_test(ctx):
         p = None
 
     executable, runfiles = exec_env.test_dispatch(ctx, exec_env, p)
+    harness_runfiles = ctx.runfiles()
```
pamaury, line 396: Small nit, I find the flow a bit confusing, I would suggest:
```python
if test_harness != None:
   # bla
else:
   harness_runfiles = ctx.runfiles()
```

rules/opentitan/qemu.bzl
jwnrt: Removed

rules/opentitan/qemu.bzl
jwnrt: > Also for the attributes here, it seems that it can only generates flash images with a single slot filled?

Yes, I tried to get banks working but couldn't. I've documented that only single-bank images are supported.

sw/device/silicon_creator/rom/BUILD
jwnrt: I needed this for a different version of `rom_exit_immediately` to work but don't anymore, removed.

pamaury (approved):

PR #25910, opened 2025-01-16
[prim,rtl] Remove ~under_rst from the expression for fifo_incr_wptr

rswarbrick:
This term can only be necessary if wready_o is true. But wready_o is driven by the following continuous assignment:

    assign wready_o = ~full_o & ~under_rst;

This commit doesn't add an assertion (to check the behaviour doesn't change) because the behaviour is internal to the module.


vogelpi (approved):
LGTM!

vogelpi:

This is a non-functional RTL change which helps improving coverage. This is fine.

rswarbrick:

A change with no functional impact, so poses no risk to the design.

PR #25912, opened 2025-01-16
[rust] update clap, object, zerocopy

nbdd0121:
Now we've updated past Rust 1.74, we can unpin `clap` and update `object`. `object` has a trait lifetime update which requires minor adjustment.

`zerocopy` underwent a major update:
* `AsBytes` is renamed to `IntoBytes`
* `.as_bytes()` now require `Immutable`
* `FromBytes` now imply `FromZeroes` and doesn't need explicit derive
* method renames


sw/host/opentitanlib/bindgen/BUILD
```diff
@@ -254,7 +254,6 @@ rust_bindgen_library(
         "--allowlist-function=status_err",
         "--allowlist-function=status_ok",
         "--generate-inline-functions",
-        "--with-derive-custom=ot_status_create_record=zerocopy::FromZeroes",
```
jwnrt, line 257: Why don’t we need this anymore?

nbdd0121: I mentioned it in the PR body and commit message:
> `FromBytes` now imply `FromZeroes` and doesn't need explicit derive

jwnrt (approved):

PR #25919, opened 2025-01-17
[manuf] simplify testing of provisioning flows across all SKUs

timothytrippel:
This updates the E2E tests of all provisioning flows (across all SKUs) to be tested with the orchestrator script. This makes the `opentitan_test` targets that also ran said flows redundant and unneeded, thus we remove them. Additionally, we no longer need duplicate CA configuration files as these are generated on the fly by the orchestrator script from the same information that was duplicated in the orchestrator SKU configuration files. This simplifies our test infrastructure and provisioning flow configurations.


sw/host/provisioning/orchestrator/configs/skus/sival.hjson
```diff
@@ -10,6 +10,7 @@
   target_lc_state: "prod",
   otp: "sival",
   perso_bin: "sw/device/silicon_creator/manuf/base/binaries/ft_personalize_{sku}_{target}.signed.bin",
+  # owner_fw_boot_str: "Bare metal PASS!"
```
cfrantz, line 13: Is this meant to be commented out?

cfrantz (approved):

sw/host/provisioning/orchestrator/configs/skus/sival.hjson
timothytrippel: Yes, currently the bare metal image for the sival SKU is not signed properly so it boot faults. This should be cleaned up in a follow up.

moidx (approved):

PR #25922, opened 2025-01-17
[dv,mem_bkdr_util] Temporarily fix otp_ctrl mem_bkdr functions

matutem:
Put back the fix in otp_ctrl_mem_bkdr_util_pkg tadded in #25884.

otp_ctrl_mem_bkdr_util_pkg needs to be a template since it depends on the partitions. Until that is done this fix will do.


vogelpi (approved):
LGTM, thanks @matutem !

PR #25924, opened 2025-01-17
Remove legacy splice

### LINKED ISSUE #25876: Bazel software build hiccup
While I'm trying to run the bazel build software, I'm running into an issue regarding "updatemem"

The command that I am running is this
bazel build //sw/...

INFO: From GenVivadoImage hw/bitstream/universal/splice-rom.update.mem:
2025-01-15 02:43:05,410 [gen_vivado_mem_image.py] Generating updatemem-compatible MEM file for ROM.
ERROR: /home/meow/opentitan/hw/bitstream/universal/BUILD:29:17: SpliceBitstream hw/bitstream/universal/splice-rom.bit failed: (Exit 1): updatemem failed: error executing SpliceBitstream command (from target //hw/bitstream/universal:splice) updatemem --force --meminfo external/_main~_repo_rules~bitstreams/cache/4b2142e60443cfc0776d0fe6b3a26c1293a6024f/chip_earlgrey_cw310/memories.mmi --data ... (remaining 7 arguments skipped)
src/main/tools/process-wrapper-legacy.cc:80: "execvp(updatemem, ...)": No such file or directory
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.719s, Critical Path: 0.71s
INFO: 12 processes: 5 internal, 6 linux-sandbox, 1 local.

pamaury:
This PR remove all uses of the legacy split rules. As a side benefit, it should allow any target to be build built with `--define bitstream=skip` and so should fix #25876


jwnrt (approved):
Excellent, LGTM

hw/bitstream/BUILD
```diff
@@ -81,44 +80,6 @@ filegroup(
...
-    src = ":bitstream",
-    data = "//sw/device/lib/testing/test_rom:test_rom_fpga_cw310_scr_vmem",
-    instance = "rom",
-    meminfo = ":cw310_mmi",
-    tags = ["manual"],
-    update_usr_access = True,
-)
-
-bitstream_splice(
-    name = "mask_rom",
```
a-will, line 97: Doesn't the pkg_files target need some handling below? (should be missing deps now, right?)

These targets were part of the release or nightly CI flow, if memory serves. The cleanup might noy be done yet. ;)

pamaury: Ah damn, I have missed them when querying users. It might be because the target is marked as manual. Maybe I will only cleanup users in `sw/` and keep the legacy rule for `hw/`. Unfortunately the new splice rule is not exactly a replacement for the old one because it expects an execution environment so I'll need to look into this in more details.

pamaury: By the way, it's interesting that even with the missing dependencies, the following does *not* return an error:
```
./bazelisk.sh build --nobuild //hw/bitstream:package
```

pamaury: Also why do we package the `mask_rom` only for the CW310 and not for the CW340?

pamaury:
I have restored `rules/splice.bzl` and kept the `mask_rom*` splices which are used for packages.

hw/bitstream/BUILD
a-will: Those release flows have not really been updated. So far, we haven't formally created releases, and the flows for it are probably also legacy cruft.

pamaury: So do you think we should completely remove the package target?

nbdd0121: I think it was broken for quite a while now, probably we should just remove it. It could provide pre-built bitstreams but we have the bitstream cache for that purpose now anyway.

a-will: I think we can get rid of the targets right now.

When we come up with a formal release process, we'll want to consider what sorts of binaries should be distributed. Keep in mind that the tarball will no longer be a git repo, so the online bitstream cache won't necessarily work. (We could ship an offline cache entry, though!)

release/BUILD
nbdd0121, line 1: Probably we should just get rid of this file completely

pamaury: Yeah I am not sure what to do about this file. I guess we could delete it and resurect it later when we do a release flow. On the other hand, I think it might be useful because it points to various "packages" so if we remove it we will end up removing those because we think they are unused. It's not like it's really harmful to keep it anyway.

PR #25925, opened 2025-01-17
[tlul,rtl] Correct the calculation of data_intg in tlul_adapter_sram

rswarbrick:
This is the value that gets supplied as a user field that provides data integrity. If we are responding with an error response, we need to make sure we send integrity bits that correspond.

The logic that was previously here detected this condition with

    (vld_rd_rsp && reqfifo_rdata.error)

but that is wrong because it shouldn't depend on vld_rd_rsp. Imagine we send a TL write with a TL error. When we read the response, the d_error flag will be high (because u_reqfifo contains the faulty TL write) and d_data will be error_blanking_data. But the integrity bits will be SecdedInv3932ZeroEcc because vld_rd_rsp is false (we haven't seen a TL read at all!)

This is also possible to trigger by using only reads. Suppose we send an TL read with a TL error and then, a few cycles later, read the TL response.

When we read the response, the d_error flag will again be high (because u_reqfifo contains the faulty TL read). Again, d_data will (correctly) be error_blanking_data. Again, we should be using error_blanking_integ for error bits but we actually use SecdedInv3932ZeroEcc.

Dropping the vld_rd_rsp term will fix the behaviour in both cases.

So it remains in sync with the RTL, we also drop a conditional coverage exclusion for rom_ctrl. Tracking down how it was actually possible to see this happen led us to the design change.

@KinzaQamar: Thanks for helping me understand this in the first place.

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
+++ b/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
@@ -98,15 +98,6 @@ Condition 21 "3623514242" "(d_valid & rspfifo_rvalid & (reqfifo_rdata.op == OpRe
 // popped again before anything appears in the response fifo.
 Condition 24 "1059982851" "(vld_rd_rsp & ((~d_error))) 1 -1" (2 "10")
 
-// The reason for condition 24 explains why we can't get (vld_rd_rsp && reqfifo_rdata.error) = 1 as
-// if vld_rd_rsp = 1, this means that rspfifo_rvalid = 1. But when we have reqfifo_rdata.error, the
-// request from the req_fifo gets popped again before becoming visible to response fifo. This also
-// explains why we can't expect 1 and 11 for condition 25 and 26 respectively.
-Condition 25 "2807788926" "((vld_rd_rsp && reqfifo_rdata.error) ? error_blanking_integ :
-                            (vld_rd_rsp ? rspfifo_rdata.data_intg :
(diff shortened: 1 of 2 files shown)


vogelpi (approved):
Thanks @rswarbrick for the fix and the carefully written description, that's very useful!

IIRC, the issue also affected reads because during a read generating a TL error, the request is not actually sent out to the SRAM and thus no response is obtained from the SRAM and `u_rspfifo` remains empty, meaning `rspfifo_rvalid` and thus `vld_rd_rsp` remain deasserted. Is this correct?

vogelpi:
@nasahlpa according to the description of Rupert we have been generating ECC errors upon TL errors in the past for the SRAMs is this inline with your recollection?

vogelpi:

This PR fixes a bug in the error reporting. Previously, the adapater did insert an ECC error was generated when experiencing a TL-UL error. Fixing this is a good thing.

rswarbrick:
@vogelpi: Yep, I think so. Indeed, that's what I originally saw when reasoning about things. See the text starting with "This is also possible to trigger".

nasahlpa (approved):
Thanks for spotting & fixing this!

I've created an issue #25927 as we def. should check the error response of the SRAM controller more carefully.

marnovandermaas (commented):
I think I ran into this issue on Sonata at some point as well, I just didn't know how to fix it. Thanks for this!

marnovandermaas:

PR #25926, opened 2025-01-17
[bazel] Enable -Wtype-limits in RISC-V toolchain

jwnrt:
This catches problems like:

```c
for (uint8_t i = 5; i >= 0; --i) {
  printf("hello!\n");
}
```

```console
$ clang -Wtype-limits main.c
main.c:5:25: warning: result of comparison of unsigned expression >= 0 is always true [-Wtautological-unsigned-zero-compare]
  for (uint8_t i = 5; i >= 0; --i) {
                      ~ ^  ~
```

### DIFF
--- a/toolchain/BUILD
+++ b/toolchain/BUILD
@@ -245,6 +245,7 @@ cc_args(
         "-Wno-unused-parameter",
         "-Wstrict-prototypes",
         "-Wswitch-default",
+        "-Wtype-limits",
     ],
 )
 


pamaury:
I am a bit confused about the `irq<0` check: `dif_*_irq_t` is an enum without an underlying type so technically it's up to compilet to decide whether it should be signed or unsigned I think. This means that removing this check might potentially be unsound of the compiler chose a signed type (which it would be allowed to do)?

jwnrt:
Hmm, yes you’re right. I don’t have the spec but the reference says it can be an implementation defined signed or unsigned type.

I notice that C23 allows you to specify the type but I don’t know if we can support that yet.

@luismarques do you have any insight on this?

luismarques:
> Hmm, yes you’re right. I don’t have the spec but the reference says it can be an implementation defined signed or unsigned type.

Where does it say that? C99 guarantees it has type `int`.

jwnrt:
https://en.cppreference.com/w/cpp/language/enum

> Each enumerated type without a fixed underlying type(since C23) is [compatible](https://en.cppreference.com/w/c/language/type#Compatible_types) with one of: char, a signed integer type, or an unsigned integer type (excluding bool and the bit-precise integer types)(since C23). It is implementation-defined which type is compatible with any given enumerated type, but whatever it is, it must be capable of representing all enumerator values of that enumeration. For all enumerations with a fixed underlying type, the enumerated type is compatible with the underlying type of the enumeration.(since C23)

Is *all* of this since C23 and prior it’s always `int`?

luismarques:
> > Each enumerated type without a fixed underlying type(since C23) is [compatible](https://en.cppreference.com/w/c/language/type#Compatible_types) with one of: char, a signed integer type, or an unsigned integer type (excluding bool and the bit-precise integer types)(since C23). It is implementation-defined which type is compatible with any given enumerated type, but whatever it is, it must be capable of representing all enumerator values of that enumeration. For all enumerations with a fixed underlying type, the enumerated type is compatible with the underlying type of the enumeration.(since C23)
> 
> Is _all_ of this since C23 and prior it’s always `int`?

C99 also had similar text, but there's a distinction between the overall enum type and the type of the members. I thought you were talking about the members.

jwnrt:
In this case I think it’s the overall type that matters since we’re not using the constants in this comparison

jwnrt:
I read here that this warning has a special case exception for `irq < ${first_irq}` even when `${first_irq}` is zero. Seems to work, so I've used that instead.

I'm not 100% confident in this template syntax so would be good if somebody could double check that.

util/make_new_dif/templates/dif_autogen.c.tpl
```diff
@@ -204,12 +204,12 @@ dif_result_t dif_${ip.name_snake}_init(
     % if ip.irqs[-1].width == 1:
       if (${ip.name_snake} == NULL ||
           type == NULL ||
-          irq < 0 ||
+          irq < kDif${ip.name_camel}Irq${ip.irqs[0].name_camel} ||
```
pamaury, line 207: I think this code is not technically correct: you also need to do a case distinction on `ip.irqs[0].width` to see if it's `1` or bigger. Now I think that virtually all IPs have their first register which is not a multireg so you can probably just `assert ip.irqs[0].width == 1` in the template to at least verify this assumption.

If you want to solve it the proper way, I would suggest to create a name for this before the if:
```mako
<%
  first_irq_name = "kDif{}Irq{}".format(ip.name_camel, ip.irqs[0].name_camel)
  if ip.irqs[-1].width > 1:
    first_irq_name += "0"
>%
and then replace the test by `irq < ${first_irq_name}`
```

pamaury (approved):

PR #25929, opened 2025-01-17
[dv, rv_core_ibex] Fix broken rv_core_ibex_nmi_irq_test

ZhongYic00:
dif_rv_core_ibex_enable_nmi does a write rather than set, so to enable both wdog and alert NMI, kDifRvCoreIbexNmiSourceAll(=wdog|alert) should be used.

With current called twice impl, as shown below, nmi_enable.wdog_en will be cleared by second call enabling NMISourceAlert.
image

### DIFF
--- a/sw/device/tests/BUILD
+++ b/sw/device/tests/BUILD
@@ -5249,9 +5249,6 @@ opentitan_test(
             "//hw/top_earlgrey:silicon_creator": None,
         },
     ),
-    silicon = silicon_params(
-        tags = ["broken"],
-    ),
     deps = [
         "//hw/top_earlgrey/sw/autogen:top_earlgrey",
         "//sw/device/lib/dif:alert_handler",
--- a/sw/device/tests/rv_core_ibex_nmi_irq_test.c
+++ b/sw/device/tests/rv_core_ibex_nmi_irq_test.c
@@ -280,10 +280,9 @@ bool test_main(void) {
   init_peripherals();
 
   // Enable both NMIs.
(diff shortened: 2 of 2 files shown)


rswarbrick (approved):
This looks very sensible to me. I'll wait a little longer to give someone who knows more about Ibex SW than I do (just in case I'm missing something).

But I want to properly say thank you for the excellent PR, complete with a screenshot that shows the problem with the previous code.

jwnrt:
`NMI_ENABLE` is `rw1s` so you shouldn't be able to clear these bits by writing 0, right?

https://opentitan.org/book/hw/ip/rv_core_ibex/doc/registers.html?highlight=nmi_enable#nmi_enable

Are we looking at the wrong signals in this waveform, or is `reggen` not generating the correct RTL?

jwnrt (changes requested):
I want to block this for now. The actual software change is fine but should be semantically equivalent, so I want to investigate why this fixes the bug.

@ZhongYic00 did your waves come from VCS? If you happen to have the logs lying around then there should be two seeds printed in there somewhere (build seed and run seed) which would help us reproduce. Thanks!

This test is passing on FPGA so I'm slightly surprised it fails in sim.

ZhongYic00:
@jwnrt I'm sorry for mistakenly thinking this was a bug in the test code due to my incomplete verification, and for misunderstanding the meaning of the `silicon_params(tags = ["broken"])`. At the same time, I appreciate your thoughtful implementation of the design verification code, which reflects the `rw1s` feature of NMI_ENABLE very well.

I'm new to the OpenTitan project and am participating in a bug hunting event that uses an previous version & modified RTL code. Upon my further verification, this issue was caused by a bug in the RTL.

Thank you both for your prompt replies! 🥰

PR #25931, opened 2025-01-17
[hmac,dv] Rename intr_test to intr_test_rename

martin-velay:
- rename as intr_test is a register name, and this register is WO. This is a follow-up of a previous PR #25903.


rswarbrick (commented):
Thanks: this seem much clearer. And I think I prefer the name you chose in the code :-)

rswarbrick (approved):

rswarbrick:
Thanks! I think this needs rebasing over a change in the scoreboard?

rswarbrick (approved):

PR #25932, opened 2025-01-17
[pentest] Fix masking off for AES-SCA

nasahlpa:
This PR consists of three different commits:

- The first two commits tidy up the code base and prepare it for the third commit
- The third commit makes sure that we correctly (i.e., after a `dif_aes_start`()) do a reseed.

Before this commit, turning off the masks for AES gave us the following TVLA figure:
aes_masks_off_before
As shown in the figure, leakage is only detected in the first AES rounds. However, with masking disabled, we expect leakage in all rounds.

With these changes, the masking off now works again:
aes_masks_off_after

### DIFF
--- a/sw/device/sca/aes_serial.c
+++ b/sw/device/sca/aes_serial.c
@@ -245,7 +245,9 @@ static void aes_encrypt(const uint8_t *plaintext, size_t plaintext_len) {
   // Using the SecAesStartTriggerDelay hardware parameter, the AES unit is
   // configured to start operation 40 cycles after receiving the start trigger.
   // This allows Ibex to go to sleep in order to not disturb the capture.
+  pentest_set_trigger_high();
   pentest_call_and_sleep(aes_manual_trigger, kIbexAesSleepCycles, false, false);
+  pentest_set_trigger_low();
 }
 
 /**
@@ -293,9 +295,7 @@ static void aes_serial_single_encrypt(const uint8_t *plaintext,
     block_ctr = 1;
   }
 
-  pentest_set_trigger_high();
   aes_encrypt(plaintext, plaintext_len);
-  pentest_set_trigger_low();
 
   aes_send_ciphertext(false);
 }
@@ -374,12 +374,10 @@ static void aes_serial_batch_encrypt(const uint8_t *data, size_t data_len) {
     block_ctr = num_encryptions;
   }
 
-  pentest_set_trigger_high();
   for (uint32_t i = 0; i < num_encryptions; ++i) {
     aes_encrypt(plaintext_random, kAesTextLength);
     aes_serial_advance_random();
   }
-  pentest_set_trigger_low();
 
   aes_send_ciphertext(true);
 }
@@ -427,7 +425,6 @@ static void aes_serial_batch_alternative_encrypt(const uint8_t *data,
   // Set trigger high outside of loop
   // On FPGA, the trigger is AND-ed with AES !IDLE and creates a LO-HI-LO per
   // AES operation
-  pentest_set_trigger_high();
   dif_aes_data_t ciphertext;
   for (uint32_t i = 0; i < num_encryptions; ++i) {
     // Encrypt
@@ -444,7 +441,6 @@ static void aes_serial_batch_alternative_encrypt(const uint8_t *data,
     // Use ciphertext as next plaintext (incl. next call to this function)
     memcpy(batch_plaintext, ciphertext.data, kAesTextLength);
   }
-  pentest_set_trigger_low();
   // Acknowledge command
   simple_serial_send_status(0);
   // send last ciphertext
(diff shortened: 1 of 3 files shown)


sw/device/sca/aes_serial.c
```diff
@@ -563,12 +559,10 @@ static void aes_serial_fvsr_key_batch_encrypt(const uint8_t *data,
   num_encryptions = read_32(data);
   SS_CHECK(num_encryptions <= kNumBatchOpsMax);
 
-  pentest_set_trigger_high();
   for (uint32_t i = 0; i < num_encryptions; ++i) {
     aes_key_mask_and_config(batch_keys[i], kAesKeyLength);
     aes_encrypt(batch_plaintexts[i], kAesTextLength);
   }
-  pentest_set_trigger_low();
```
vogelpi, line 571: We moved those functions out of the encrypt to not have to repeatedly call it inside the hot loop during batch encryptions. I think it did have a notable performance impact but don't remember the numbers to be honest.

sw/device/sca/aes_serial.c
```diff
@@ -165,6 +165,37 @@ dif_aes_transaction_t transaction = {
...
+ * masking.
+ *
+ * @param key Key.
+ * @param key_len Key length.
+ */
+static void aes_sca_load_fixed_seed(void) {
+  aes_sca_wait_for_idle();
+  SS_CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerPrngReseed));
+  aes_sca_wait_for_idle();
+  SS_CHECK_DIF_OK(dif_aes_trigger(&aes, kDifAesTriggerDataOutClear));
```
vogelpi, line 195: I didn't remember this was needed but it turns out it really is! It maybe worth to add a comment why.

sw/device/tests/penetrationtests/firmware/sca/aes_sca.c
```diff
@@ -147,6 +148,28 @@ dif_aes_transaction_t transaction = {
...
+ * masking.
+ *
+ * @param key Key.
+ * @param key_len Key length.
+ */
+static status_t aes_sca_load_fixed_seed(void) {
+  AES_TESTUTILS_WAIT_FOR_STATUS(&aes, kDifAesStatusIdle, true, kTestTimeout);
+  TRY(dif_aes_trigger(&aes, kDifAesTriggerPrngReseed));
+  AES_TESTUTILS_WAIT_FOR_STATUS(&aes, kDifAesStatusIdle, true, kTestTimeout);
+  TRY(dif_aes_trigger(&aes, kDifAesTriggerDataOutClear));
```
vogelpi, line 167: Also here I would add a comment.

sw/device/tests/penetrationtests/firmware/sca/aes_sca.c
```diff
@@ -178,6 +201,13 @@ static status_t aes_key_mask_and_config(const uint8_t *key, size_t key_len) {
   }
   TRY(dif_aes_start(&aes, &transaction, &key_shares, NULL));
 
+#if !OT_IS_ENGLISH_BREAKFAST
+  if (transaction.force_masks) {
+    // Disable masking. Force the masking PRNG output value to 0.
+    TRY(aes_sca_load_fixed_seed());
+  }
+#endif
+
```
vogelpi, line 210: IIRC, it doesn't hurt to call this on the CW305. Because the trigger register is there but the reseeding is just ignored (there is no CSRNG though which is why we need to exclude the CSRNG setup).

vogelpi (approved):
LGTM, thanks @nasahlpa !

sw/device/sca/aes_serial.c
nasahlpa: I've checked whether this makes a big difference:

this PR:
simpleserial aes_fvsr_key_batch 20 batches: 615 traces / s
simpleserial aes_fvsr_key_batch 50 batches: 1444 traces / s

before this PR:
simpleserial aes_fvsr_key_batch 20 batches: 617 traces / s
simpleserial aes_fvsr_key_batch 50 batches: 1450 traces / s

So I think the change is reasonable :-)

sw/device/sca/aes_serial.c
nasahlpa: I quickly checked, we actually do not need to clear the data out registers with `dif_aes_trigger(&aes, kDifAesTriggerDataOutClear)`. Only triggering the Prng Reseed is sufficient and gives us the TVLA plot from above.

Added a comment & removed the data register clearing.

sw/device/tests/penetrationtests/firmware/sca/aes_sca.c
nasahlpa: Please see above.

sw/device/tests/penetrationtests/firmware/sca/aes_sca.c
nasahlpa: It seems when adding this also for CW305, the code does not compile anymore. So I've kept the if guards.

sw/device/sca/aes_serial.c
vogelpi: I agree! Thanks for feeding back the results!

Issue #25920, opened 2025-01-17
[tools,topgen] Split topgen into creating complete topcfg and the rest


matutem:
### Description

Topgen is a monolithic tool and is called multiple times to generate the full set of artifacts. Each time it generated the complete top configuration, and then performs a some subset of all the actions it is able to do. This is not efficient and is hard to maintain.

This calls for a change in the topgen flow so the complete top configuration is generated and written to disk by one tool. The actions that depend on the complete config can then be performed by smaller separate tools, and most importantly, they can be invoked only if the complete top configuration file has changed. This will allow a more incremental flow, and since most of the separate tasks depending on the complete config are independent, they can be done in parallel.

This is not only more efficient and simpler, but it simplifies the development of different tops, so can be very helpful for multi-top.

alees24:
It looks to me as though the 'filecmp' invocation in pursuing convergence is presently comparing the output from the current iteration with itself.
```
        if pass_idx > 0 and filecmp.cmp(
                f"/tmp/top{topname}cfg_{pass_idx}.hjson",
                "/tmp/top{}cfg_{}.hjson".format(topname, pass_idx),

```
Presumably fallout from changing the loop indexing.

Issue #8207, opened 2021-09-15
[top] No option in top_earlgrey.hjson to use <module>.hjson that is not under hw/ip or hw/top_earlgrey  


sha-ron:
Specifically AST attr in top_earlgrey.hjson is defined as "reggen_only" so ast.hjson is taken from hw/top_earlgrey/ip/ast/data.
When using the 'real' (not open source) AST need to use a private ast.hjson that is located somewhere else (under Nuvoton's private repository) .

tjaychen:
@sha-ron do you need for this for the full ral modeling at chip level? 
We can certainly add a flag to look elsewhere is that is the cleanest solution. 

@zi-v 
@Jacob-Levy 
Is there any way for you guys to expose a fully compatible (in terms of address map only) open source hjson?
You don't need to show any information on how any of it is used.  Ideally we would like the open source / closed source to at least match in terms of spacing / naming to test out our infrastructure.

tjaychen:
@msfschaffner this is a bit different from the prim_csr thing, since ast isn't sitting behind another hjson or anything.

sha-ron:
@tjaychen 
yes need this for the ral modeling at chip level. A flag can be great,

tjaychen:
is it at all possible to align the register names / address space between
open source / closed source? Because that would be the easiest way i think.

sha-ron:
Revealing the closed source names to the open source is problematic and using meaningless names in the closed makes it harder to debug and less readable.

msfschaffner:
CC @moidx

msfschaffner:
As @tjaychen noted, we would ideally just use one hjson in the open-source for this. 
We know that this is not ideal from a closed-source documentation perspective - but it actually makes things simpler from a tooling, testing and SW perspective since we would not have multiple hjson sources that need to be tested. See also the [solution for the prims](https://docs.google.com/document/d/1pnDU3-V2BwEbw2nxFmSClJTIODKlCGhPS4TFC8n7D04/edit#heading=h.l7qwxjui9ien) that we had discussed.

In any case - even if such a switch is implemented as proposed by Nuvoton - Nuvoton needs to guarantee that the open-source hjson is compatible with the closed-source one in terms of address ranges and CSR properties.
Our open-source ROM is going to use software constants generated from the open-source hjson, and this needs to work 100% with the closed source AST (CC @moidx).

Jacob-Levy:
Hi Michael,  

[...]

PR #25933, opened 2025-01-17
[tools,topgen] Enhance and simplify topgen

matutem:
This is a big change in topgen. Some salient changes:
- Regenerates from scratch the ipgen in-memory configurations (as IpBlock objects),
  so it is truly incremental: the previous flow reused the pre-existing generated ipgen
  hjson if found.
- Delays the actual ipgen file generation until the complete top config is generated.
  This is a step towards breaking topgen into smaller and mutually independent steps,
  thus enabling parallelizing them, and creating smaller and simpler tools to tackle each.
- It converges with a single pass of _process_top.

In order to accomplish this some important changes are made:
- Call some of the merge_top `functions` suitable for each specific ipgen prior to
  generating the in-memory hjson config. This means not all IpBlock objects are
  available, so these merge functions are instrumented to skip blocks missing.
- The generation of ipgens are ordered according to dependencies. This order created
  from the expected block functionality, so is not derived from a constructed graph, but
  converging in one pass means it is correct.
- Calling merge functions early means some of the objects in the in-memory top config
  are not plain dictionaries but are classes, so the code needs to handle either.

Part of #25920


matutem:
CHANGE_AUTHORIZED: hw/top_earlgrey/data/top_earlgrey.hjson

hw/top_englishbreakfast/data/top_englishbreakfast.hjson
```diff
@@ -158,24 +158,47 @@
...
+    int: [
+      {
+        "name": "MainPwr",
+        "desc": "main power glitch reset request",
+        "module": "pwrmgr_aon"
+      },
+      {
+        "name": "Esc",
+        "desc": "escalation reset request",
+        "module": "alert_handler"
```
a-will, line 191: Huh, but englishbreakfast doesn't have an alert_handler. Does this field not actually do anything?

util/topgen/validate.py
```diff
@@ -259,6 +298,99 @@
...
+interrupt_added = {}
+
+param_required = {
+    'name': ['s', 'the parameter name'],
+    'desc': ['s', 'the parameter description'],
+    'type': ['s', 'the data type of the parameter'],
+    'default': ['s', 'the default value of the parameter'],
+}
+param_optional = {
+    'expose': ['s', 'seems redundant TODO'],
```
a-will, line 362: This is whether the parameter bubbles up to the top of the generated toplevel.sv file. We probably need to redo some of the template stuff... but that's for another day!

util/topgen/validate.py
```diff
@@ -259,6 +298,99 @@
...
+    'local': ['s', 'whether it is a localparam, interpreted as boolean'],
+    'name_top': ['s', 'the name in the top-level'],
+    'randcount': ['d', 'TODO'],
```
a-will, line 365: This is apparently the number of bits requested for a random number inserted into toplevel_rnd_cnst_pkg.sv. If I'm not misreading the code, it looks like `randwidth` is added ...and is just equal to `randcount`.

util/topgen/validate.py
```diff
@@ -259,6 +298,99 @@
...
+        'whether the signal is unidirectional or part of a request-response '
+        'pair'
+    ],
+    'act': ['s', 'whether it is a request (req) or a response (rsp)'],
+    'width': ['d', 'the number of items of the signal for arrays'],
+}
+inter_sig_optional = {
+    'desc': ['s', 'the inter signal description'],
+    'inst_name': ['s', 'the instance this signal connects to'],
+    'index': ['d', 'not clear TODO'],
```
a-will, line 386: Added during elaboration. It's the index of an element from an inter_sig connection array.

For example, from this input...
https://github.com/lowRISC/opentitan/blob/3563d66411f24bd3017385dfb2636db2edaf5668/hw/top_earlgrey/data/top_earlgrey.hjson#L1168-L1170

comes output like...
https://github.com/lowRISC/opentitan/blob/3563d66411f24bd3017385dfb2636db2edaf5668/hw/top_earlgrey/data/autogen/top_earlgrey.gen.hjson#L19262-L19273

a-will (commented):
These are probably mostly useless comments and nits. I still have a lot more to look at. :smile:

util/topgen/merge.py
```diff
@@ -626,8 +625,11 @@ def extract_clocks(top: OrderedDict):
     group. However, it is possible to define the group attribute per clock
     if required.
     '''
+    if not isinstance(top['clocks'], Clocks):
+        top['clocks'] = Clocks(top['clocks'])
```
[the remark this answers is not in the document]
matutem: This behavior was not changed in this PR, rather this quirky behavior was exposed because we run merge operations more often. I agree it would be nice to keep the consisting of json objects only.

util/topgen/validate.py
matutem: I think they are not always the same, which is what triggered the TODO. I think we need a separate PR which can generate documentation about the format of the hjson files partly based on the key's documentation.

util/topgen/validate.py
matutem: Added to the documentation, thanks.

util/topgen/validate.py
```diff
@@ -259,6 +298,99 @@
 reset_connection_optional = {}
 reset_connection_added = {}
 
+reset_requests_required = {}
+reset_requests_optional = {
+    'int': ['s', 'internal request list'],
+    'debug': ['s', 'debug request list'],
+    'peripheral': ['s', 'peripheral request list'],
+}
```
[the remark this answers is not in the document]
matutem: Good point: I extended the description.

hw/top_englishbreakfast/data/top_englishbreakfast.hjson
matutem: TBD: notice the earlgrey and englishbreakfast use the same toplevel.sv.tpl, so there are plenty of questions about this.

util/topgen.py
```diff
@@ -367,102 +450,131 @@ def generate_pinmux(top: Dict[str, object], out_path: Path) -> None:
...
-        "number_of_clock_groups": len(clocks.groups),
-        "with_alert_handler": with_alert_handler,
+    with_alert_handler = lib.find_module(top['module'],
+                                         'alert_handler') is not None
+
+    topname = top["name"]
+    return {
+        "src_clks":
+        OrderedDict({name: vars(obj)
+                     for name, obj in clocks.srcs.items()}),
```
a-will, line 488: Is that the correct indentation for style / readability? When keys and values are on separate lines, the values don't get indented?

(just a question -- it feels a bit hard to read for me)

util/topgen.py
```diff
@@ -543,106 +659,103 @@ def generate_flash(topcfg: Dict[str, object], out_path: Path) -> None:
...
+    """Generate the regfile for top_only IPs."""
     log.info("Generating top only modules")
 
-    for ip, reggen_only in top_only_dict.items():
+    for ip in top_only_dict:
         ip_out_path = out_path / "ip" / ip
-        if reggen_only and alt_hjson_path:
-            hjson_path = Path(alt_hjson_path)
-        else:
-            hjson_path = ip_out_path / "data" / f"{ip}.hjson"
```
a-will, line 645: Are we sure we should be killing this feature? I don't know who uses it, but there doesn't seem to be a great reason to remove it.

That said... does it even work anymore? It looks like this was broken by https://github.com/lowRISC/opentitan/pull/25129, undoing the support that was completed for https://github.com/lowRISC/opentitan/issues/8207.

In a world where the tops in the open-source repo are reference tops and Nuvoton's top hjson is out-of-tree, I guess we don't really need that argument anymore, do we? If an integrator wanted to work with partners using a sort of "redacted" version of their internal top, I guess they could provide a separate toplevel.hjson that has the proprietary IPs replaced with shareable models.

a-will (approved):
I admit my eyes started to glaze over after awhile of reading the refactoring. Also, we probably could use a canonical config output order, so we don't get these huge diffs that simply move the same information around.

However, from what I have been able to read, LGTM

util/topgen.py
matutem: This flag is not even sensible: imagine we had multiple reggens, then they would all use the same hjson file? I vaguely recall talking to someone else about this and agreeing we don't need it anyway.

Razer6 (approved):
Thanks for doing this. So far, this looks really good and simplifies a lot. I will also apply these changes downstream and check if they work with other tops as well.

matutem:
@a-will regarding https://github.com/lowRISC/opentitan/pull/25933#pullrequestreview-2594800203, I expect the new hjson files will be stable. The changes for this change in topgen are the cause of the present difference, but I expect going forward there should be no gratuitous diffs. If there are we can use OrderedDicts for problematic dicts.

util/topgen.py
```diff
@@ -1252,15 +1508,53 @@ def main():
...
+                raise ValueError('Missing alias_target key '
+                                 'in alias file {}.'.format(alias))
+            alias_target = alias_cfg['alias_target'].lower()
+            if alias_target in alias_cfgs:
+                raise ValueError(f"Multiple alias targets for {alias_target}")
+            alias_cfgs[alias_target] = alias_cfg
+
+    topname = topcfg["name"]
+    cfg_copy = deepcopy(topcfg)
+    for pass_idx in range(1, maximum_passes + 1):
```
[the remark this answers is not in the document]
matutem: This was to make f-strings be easier, but that is too much of a detail. Anyway, python range has three ways to be called, and with two arguments (a, b) it yields a ... b-1, and with three a, b, s it does as with two except s becomes the increment.

I undid it anyway and also noticed process_top doesn't need pass_idx.

util/topgen.py
```diff
@@ -1211,8 +1469,6 @@ def main():
         log.error('Seed "rnd_cnst_seed" not found in configuration HJSON.')
         exit(1)
 
-    secure_prng.reseed(topcfg["rnd_cnst_seed"])
```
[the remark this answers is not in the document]
matutem: I moved it to where it actually matters, which is prior to each loop of _process_top: without running each pass with the same seed we could not compare the output file for stability.

util/topgen.py
```diff
@@ -787,218 +904,343 @@ def render_template(template_path: str, rendered_path: Path, **other_info):
...
+                        _get_flash_ctrl_params(topcfg))
+    if "otp_ctrl" in ipgen_instances:
+        insert_ip_attrs(ipgen_instances["otp_ctrl"][0]["type"],
+                        _get_otp_ctrl_params(cfg_path))
+    if "ac_range_check" in topcfg:
+        assert "ac_range_check" in ipgen_instances
+        insert_ip_attrs(ipgen_instances["ac_range_check"][0]["type"],
+                        _get_ac_range_check_params(topcfg))
+    # Pinmux depends on flash_ctrl and otp_ctrl
+    amend_pinmux_io(topcfg, name_to_block)
```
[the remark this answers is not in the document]
matutem: Done. I guess for tops without a pinmux we may need an alternative flow to potentially connect to pads... though I imagine you may never connect pads directly to these models.

util/topgen.py
```diff
@@ -143,102 +227,97 @@ def generate_xbars(top: Dict[str, object], out_path: Path) -> None:
...
         "n_classes": n_classes,
-        "n_lpg": n_lpg,
+        "n_lpg": n_lpgs,
         "lpg_map": lpg_map,
         "top_pkg_vlnv": f"lowrisc:constants:top_{topname}_top_pkg",
     }
 
-    ipgen_render("alert_handler", topname, params, out_path)
+
+def generate_alert_handler(top: Dict[str, object], module: Dict[str, object],
```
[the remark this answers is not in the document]
matutem: There are some subtleties: the get_params functions have slightly different signature so passing the get_params would be quirky; passing the params themselves won't work when and if we support different multiply-parameterized instances of ipgens. I'd rather punt on changing this in this PR.

util/topgen/merge.py
```diff
@@ -1343,6 +1432,43 @@ def amend_pinmux_io(top: Dict, name_to_block: Dict[str, IpBlock]):
...
+    for m in top_cfg['module']:
+        block = name_to_block.get(m['type'])
+        if block is None and allow_missing_blocks:
+            continue
+        racl_mappings = m.get('racl_mappings', {})
+        # Nothing to be done if there are no mappings or the mappings
+        # were expanded already.
+        # If racl_mappings is expanded the path is one of the fields
+        for if_name, mapping_path in racl_mappings.items():
+            if isinstance(mapping_path, dict):
```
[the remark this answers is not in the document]
matutem: amend_racl can be called multiple times, so this is simple and idempotent.

util/topgen.py
```diff
@@ -367,102 +450,131 @@ def generate_pinmux(top: Dict[str, object], out_path: Path) -> None:
...
+        OrderedDict({name: vars(obj)
+                     for name, obj in clocks.srcs.items()}),
+        "derived_clks":
+        OrderedDict(
+            {name: vars(obj)
+             for name, obj in clocks.derived_srcs.items()}),
+        "typed_clocks":
+        OrderedDict({ty: d
+                     for ty, d in typed_clks.items() if d}),
+        "hint_names":
```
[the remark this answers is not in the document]
matutem: I let lintpy.py decide on formatting, this way we don't end up with changes that will later be undone to clean up lint. I encourage you to do the same, or we will have flip-flops on plain formatting.

util/topgen.py
```diff
@@ -143,102 +227,97 @@ def generate_xbars(top: Dict[str, object], out_path: Path) -> None:
...
-    topname = top["name"]
 
+def generate_ipgen(topcfg: Dict[str, object],
+                   module: Dict[str, object],
+                   params: Dict[str, object],
+                   out_path: Path) -> None:
+    topname = topcfg["name"]
+    module_name = module["type"]
+    module_instance_name = params.get("module_instance_name")
+    assert not module_instance_name or module_instance_name == module_name
```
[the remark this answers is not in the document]
matutem: Yes, and it is important to keep that invariant.

Razer6:
CHANGE_AUTHORIZED: hw/top_earlgrey/data/top_earlgrey.hjson

a-will:

A property name changed, and some other properties were added. However, no functional changes to earlgrey were made (except potentially for the values of random numbers, but those aren't protected).

PR #25917, opened 2025-01-17
[opentitantool] Ship updated firmware for HyperDebug

jesultra:
Support 48Mhz clock generation, and bugfixes

Add ability to change HyperDebug core clock frequency.  Choosing 96 MHz will allow any PWM capable pin to produce a 48 MHz clock suitable for driving a NuvoTitan chip.  (CN10.31 is the only pin that can output 96 MHz.)

Also included are a few bugfixes to HyperDebug polling the "GSC ready" signal when passing through TPM operations.

Relevant CLs:
d090a07063 chip/stm32: Avoid stuck UART to USB forwarding
b3c959b5c5 board/hyperdebug: Avoid clobbering CCR_CFGR register
765b8a9d2b chip/stm32: Wait for core voltage to stabilize on STM32L4/5
6e9f274d2c board/hyperdebug: Improve GSC ready pulse detection
843d7ee9a6 board/hyperdebug: Expect GSC ready pulse after transaction
1b9f351528 board/hyperdebug: Properly set system timer tick rate
3962928140 board/hyperdebug: Allow runtime changes to system clock
821e9349d2 chip/stm32: Consolidate clock enums
e7f8842e93 chip/stm32: Rename STM32_HSE_CLOCK to CONFIG_STM32_CLOCK_HSE_HZ
a8fc0258fe board/hyperdebug: Fast clock output on CN10.31
6265770002 board/hyperdebug: Add support for PWM via low power timers
195271c11d board/hyperdebug: Refactor PWM
fbf583383f board/hyperdebug: Fix off-by-one and rounding errors

PR #25934, opened 2025-01-18
Revert "[opentitantool] Refuse downgrading HyperDebug firmware"

jesultra:
This reverts commit 7a0134a9b4d5620435b9750683083662acc7d6ab, such that opentitantool no longer requires the `--force` switch in order to downgrade firmware with `opentitantool transport update-firmware`. Instead, opentitantool will be back to the original behavior of overwriting firmware in any case when the version number does not match exactly.
    
The original reason for not downgrading was that our branches had diverged, such that they had different firmware versions, resulting in CI machines constantly upgrading/downgrading as CI runs from different branches were scheduled.
    
PR #25917, however introduced a faulty HyperDebug firmware with never version number, which is now on a number of CI machines, and will not be downgraded automatically, with the result that "innocent" other PRs will have CI failures.
    
I think it is clear that we should avoid such excessive flashing by keeping the HyperDebug firmware version in sync between branches.  (We do not need to keep all of opentitantool logic in sync, just the built-in firmware version.)  For now, in order to avoid CI failures, we should merge this PR, and accept some upgrading/downgrading.

### DIFF
--- a/sw/host/opentitanlib/src/transport/hyperdebug/dfu.rs
+++ b/sw/host/opentitanlib/src/transport/hyperdebug/dfu.rs
@@ -10,7 +10,6 @@ use regex::Regex;
 use serde_annotate::Annotate;
 use std::any::Any;
 use std::cell::RefCell;
-use std::cmp::Ordering;
 
 use crate::transport::{
     Capabilities, Capability, ProgressIndicator, Transport, TransportError, UpdateFirmware,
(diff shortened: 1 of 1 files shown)


nbdd0121:
What's the rationale of making the change?

nbdd0121 (approved):
The change LGTM and I think it probably makes sense to use exact version, although I think
1. before/after merging this change, keep HyperDebug firmware in sync (we haven't done so already I think)
2. In the PR body and commit message, articulate the rationale of making this change.

jesultra:
I have updated the commit and PR descriptions to mention that PR #25917 introduced flawed HyperDebug firmware with newer version number, which is a problem as once it has run on CI, those machines will retain the newer buggy version, causing CI failures for other unrelated PRs, unless this PR is merged.

nbdd0121:
integrated_dev isn't active so doesn't need backporting

PR #25935, opened 2025-01-18
[hw,alert_handler,rtl] Use full-scope package definition in toplevel

Razer6:
We use a stiching tool to automatically connect ports outside of the OpenTitan bounding box. This tool relies on the fact having the full-scope struct type available in the port list rather than an already imported type. This PR adds the full-scope port type to the alert and escalation interface of the alert handler, which we also use outside of OpenTitan.

### DIFF
--- a/hw/ip_templates/alert_handler/rtl/alert_handler.sv.tpl
+++ b/hw/ip_templates/alert_handler/rtl/alert_handler.sv.tpl
@@ -15,36 +15,36 @@ module ${module_instance_name}
   parameter lfsr_seed_t RndCnstLfsrSeed = RndCnstLfsrSeedDefault,
   parameter lfsr_perm_t RndCnstLfsrPerm = RndCnstLfsrPermDefault
 ) (
-  input                                    clk_i,
-  input                                    rst_ni,
-  input                                    rst_shadowed_ni,
-  input                                    clk_edn_i,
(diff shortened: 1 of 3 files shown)


a-will (approved):
This seems fine, but at some point, it sounds like we'll break the stitching tool when we introduce type parameters for some ports.

...or not, if the stitching tool only cared here because the relevant port went up to a "top-level" port. I notice that `alert_crashdump_t` was left alone.

Razer6:
> This seems fine, but at some point, it sounds like we'll break the stitching tool when we introduce type parameters for some ports.

Hehe, I'm just preparing one PR. Let me check on the `alert_crashdump_t`.

vogelpi:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/alert_handler/rtl/alert_handler.sv

This PR doesn't change the functionality of Earlgrey.

vogelpi (approved):
LGTM

a-will:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/alert_handler/rtl/alert_handler.sv

rswarbrick:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/alert_handler/rtl/alert_handler.sv

This won't change behaviour: it's just making some name resolution manual.

PR #25938, opened 2025-01-19
[hw,otp_ctrl,rtl] Parametrize the DFT config and response port

Razer6:
This parametrization allows the usage of prim_otp with different DFT config and response ports. One usage is for example when using macro tiling within the prim instance and have different instances with different configuration in the design. By passing the port type as a parameter we don't need to pull in all the tiling related parameters into the design.


a-will (approved):

vogelpi:

This PR doesn't change the functionality of OTP in Earlgrey.

a-will:

vogelpi (approved):

PR #25939, opened 2025-01-19
[hw,pinmux,rtl] Only depend on USBDEV if USB wakeup is enabled

Razer6:
No need to pull in the USB device if the USB wakeup is not enabled in the IPGEN config.


a-will (approved):

rswarbrick (approved):
Thanks for the tidy-up!

PR #25940, opened 2025-01-19
[bazel,hw] improve hermeticity of bitstream builds

nbdd0121:
* `HOME` is no longer used (except by hsmtool to find Google Cloud credentials)
* `XILINX_HLS` environment variable is removed (we don't use HLS anyway)
* Only relevant part of `PATH` is used, not the full path (this allows us to remove the CI hack)
* `updatemem` is now sandboxed

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -154,3 +154,7 @@ test --build_tests_only
 # these when running sandboxed tests.
 test --sandbox_explicit_pseudoterminal
 run --sandbox_explicit_pseudoterminal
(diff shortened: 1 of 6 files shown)


AlexJones0 (approved):
This all looks okay to me and should help remove the PATH hack in CI, thanks @nbdd0121.

pamaury (approved):
Looks great :)

pamaury (approved):

PR #25944, opened 2025-01-20
[python] bump dependencies with security alerts

nbdd0121:
This is limited to just packages with "High" security alerts


pyproject.toml
```diff
@@ -6,7 +6,7 @@
 name = "opentitan"
 version = "0.0.0"
 description = "Open-source hardware root-of-trust"
-requires-python = ">=3.8"
+requires-python = ">=3.8.1"
```
jwnrt, line 9: Was this required for our deps to work, or did `3.8` have a security advisory?

jwnrt (approved):
Looks fine, thanks

pyproject.toml
nbdd0121: Some deps specify 3.8.1 as minimum

AlexJones0 (approved):
Looks good, thanks @nbdd0121.

jwnrt:
I should have asked: was there a reason you only handled the high severities and not the moderate ones?

nbdd0121:
CyberEssentials compliance requires patching critical/high severity ones while the others may be delayed a bit.

I think the rest can be fixed by the effort of bumping Python dependency versions in general.

nbdd0121:
And I don't want to handle Python 3.8 & 3.12 issues again so ideally we'll update minimum Python to 3.9 before attempting that.

jwnrt:
Sure, thanks. I would say "open an issue" but those security advisories are already issues :)

PR #25946, opened 2025-01-20
[opentitanlib] get rid of bitvec dependency

nbdd0121:
This dependency hasn't been updated for a while, and given it's use is limited, just drop it.

This gets rid of 5 packages, and I think the readability/maintainability is not adversely affected.


jwnrt (approved):
I think this is actually more readable, thanks

PR #25947, opened 2025-01-20
[ci] support merge queues

nbdd0121:
With this change and merge queue enabled, PRs will only be merged to target branch once the bitstream is built and cached. This prevent RTL changes from causing all subsequent CI jobs to rebuild bitstream after merging and before bitstreams are built and cached.

This also has the advantage of reducing the likelihood of merge skews, if we move jobs from push to merge_group event. These are future possibilities, for now the merge queue is only gated on bitstream being built.

Implementation-wise, GitHub uses a single check name to gate both PR and merge queue, so I've created two dynamically named jobs to allow the PR to be gated only on `quick_lint` while the merge queue will be gate on bitstream being built and cached.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,12 @@ on:
   push:
     branches-ignore:
       - "backport-*"
+      - "gh-readonly-queue/*"
     tags:
       - "*"
+  merge_group:
+    types:
+      - checks_requested
 
 permissions:
   contents: read
@@ -81,6 +85,7 @@ jobs:
     name: Lint (slow)
     runs-on: ubuntu-22.04
     needs: quick_lint
+    if: ${{ github.event_name != 'merge_group' }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


jwnrt (commented):
The merge blocker jobs are clever, but I think they might be a bit hacky. What's the harm in blocking both regular CI and the merge queue on quick lints and the bitstreams?

Committers can bypass the requirement if needed.

I think preventing merge skew for `quick_lints` is a good idea. Anecdotally this job is rarely flaky and easy to fix if there's a problem.

jwnrt:
Discussed offline, making bitstream builds required for the *pull request* might delay crucial fixes because they tend to take a long time. Unfortunately bypassing this requirement also bypasses the merge queue which we want to avoid.

jwnrt:
We can reconsider in the future. Bitstream builds tend to be stable so it's okay to require them for CI.

Let's wait until cached bitstream loads don't need to run on our (resource constrained) bitstream builder machine. That should make most PRs very fast to get through quick lints and bitstreams.

PR #25948, opened 2025-01-20
[dv,pwm] PWM block has no interrupts

alees24:
PWM has no interrupt-related registers so drop `intr_test` because it does nothing useful.


rswarbrick (approved):
Thanks!

PR #25949, opened 2025-01-20
[darjeeling,sensor_ctrl] Remove sensor_ctrl instance

Razer6:
Darjeeling does not use sensor_ctrl. Let's remove that, make the crossbars simpler.


vogelpi (approved):
This looks good to me, thanks @Razer6 !

PR #25951, opened 2025-01-20
[tlul,rtl] Merge u_sramreqfifo and u_sramreqaddrfifo

rswarbrick:
Having two FIFOs with pointers that are kept in lock-step is a little silly. Turn them into a single wider FIFO.

Note that the two FIFOs had compatible parameters. The address FIFO had OutputZeroIfEmpty set to 1, which is the default value (so u_sramreqfifo got it too). Request it explicitly in case the default changes in future.

### DIFF
--- a/hw/ip/tlul/rtl/tlul_adapter_sram.sv
+++ b/hw/ip/tlul/rtl/tlul_adapter_sram.sv
@@ -246,9 +246,14 @@ module tlul_adapter_sram
     logic                      error ;
   } rsp_t ;
 
-  localparam int SramReqFifoWidth = $bits(sram_req_t) ;
+  localparam int SramReqWidth = $bits(sram_req_t);
+  localparam int SramReqFifoWidth = SramReqWidth + (DataXorAddr ? SramBusBankAW : 0);
   localparam int ReqFifoWidth = $bits(req_t) ;
   localparam int RspFifoWidth = $bits(rsp_t) ;
+
+  // An item in the SRAM request fifo is an SRAM request (of width SramReqWidth) plus, if
+  // DataXorAddr is true, some bits of the address that the request touches.
+  typedef logic [SramReqWidth-1:0] sram_req_fifo_item_t;
 
   // FIFO signal in case OutStand is greater than 1
   // If request is latched, {write, source} is pushed to req fifo.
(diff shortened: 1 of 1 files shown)


hw/ip/tlul/rtl/tlul_adapter_sram.sv
```diff
@@ -502,12 +519,12 @@ module tlul_adapter_sram
           // e.g., due to a fault, rdata now contains faulty data, which is detected by the
           // integrity mechanism.
           rdata_tlword = {
-              rdata_reshaped[sramreqfifo_rdata.woffset][DataWidth-1:top_pkg::TL_DW],
-              rdata_reshaped[sramreqfifo_rdata.woffset][top_pkg::TL_DW-1:0] ^
-                  {{(top_pkg::TL_DW-SramBusBankAW){1'b0}}, sramreqaddrfifo_rdata}
+              rdata_reshaped[sram_req_rdata.woffset][DataWidth-1:top_pkg::TL_DW],
+              rdata_reshaped[sram_req_rdata.woffset][top_pkg::TL_DW-1:0] ^
+                  {{(top_pkg::TL_DW-SramBusBankAW){1'b0}}, sram_addr_wdata}
```
nasahlpa, line 524: Not entirely sure but shouldn't this be `sram_addr_rdata` instead of `sram_addr_wdata`?

rswarbrick: Maybe that's why CI is failing :-) I'll take a look at it now!

rswarbrick:
Cool, this now works :-) I'd love reviews!

nasahlpa (approved):
Thanks @rswarbrick, this looks good to me!

hw/ip/tlul/rtl/tlul_adapter_sram.sv
```diff
@@ -246,10 +246,15 @@ module tlul_adapter_sram
...
-  localparam int SramReqFifoWidth = $bits(sram_req_t) ;
+  localparam int SramReqWidth = $bits(sram_req_t);
+  localparam int SramReqFifoWidth = SramReqWidth + (DataXorAddr ? SramBusBankAW : 0);
   localparam int ReqFifoWidth = $bits(req_t) ;
   localparam int RspFifoWidth = $bits(rsp_t) ;
 
+  // An item in the SRAM request fifo is an SRAM request (of width SramReqWidth) plus, if
+  // DataXorAddr is true, some bits of the address that the request touches.
+  typedef logic [SramReqWidth-1:0] sram_req_fifo_item_t;
+
```
vogelpi, line 257: It's a bit confusing that this type is name `sram_req_fifo_item_t` but it uses `SramReqWidth` rather than `SramReqFifoWidth`. But anyway, the type is not used and should probably be removed?

nasahlpa:

This PR merges two FIFOs that anyway were used in lock step. So this is fine.

vogelpi (approved):
LGTM with one comment, thanks @rswarbrick !

I would appreciate if also @nasahlpa could take a look please.

vogelpi:
> LGTM with one comment, thanks @rswarbrick !
> 
> I would appreciate if also @nasahlpa could take a look please.

And I see he already did :-D

vogelpi:

This PR merges two FIFOs that anyway were used in lock step. So this is fine.

PR #25952, opened 2025-01-20
[hw,sram_ctrl,rtl] Add flop stage on RAM output

Razer6:
This PR adds support for an optional flop stage on the RAM macro output. The support for this stage is already present but this PR adds the necesary parameters to sram_ctrl to be able to configure it. For Darjeeling, this flop stage is enabled on all sram_ctrl isntances. For the other tops it remains disabled.


a-will:
Does this actually work as-is?

I'm surprised to see no changes to any control signals (which also are frequently not connected, suggesting that there is an implicit dependence on single-cycle read latency).

Razer6:
Hmm the logic for the output stage is implemented in `prim_ram_1p_adv` enabled via the existing parameter `EnableOutputPipeline`. That delays the output logic, i.e., `rvalid`, `rdata`, and `error` via:

```verilog
f (EnableOutputPipeline) begin : gen_regslice_output
    // Put the register slices between ECC decoding to output

    // If no ECC or parity is used, do not use prim_flop to allow synthesis
    // tool to optimize the registers.
    if (EnableECC || EnableParity) begin : gen_prim_rvalid_flop
      prim_flop #(
        .Width(MuBi4Width),
        .ResetValue(MuBi4Width'(MuBi4False))
      ) u_rvalid_flop (
        .clk_i,
        .rst_ni,
        .d_i(MuBi4Width'(rvalid_d)),
        .q_o({rvalid_q})
      );
    end else begin: gen_no_prim_rvalid_flop
        if (!rst_ni) begin
          rvalid_q <= MuBi4False;
        end else begin
          rvalid_q <= rvalid_d;
        end
      end
    end

      if (!rst_ni) begin
        rdata_q  <= '0;
        rerror_q <= '0;
      end else begin
        rdata_q  <= rdata_d;
        // tie to zero if the read data is not valid
        rerror_q <= rerror_d & {2{mubi4_test_true_loose(rvalid_d)}};
      end
    end
  end else begin : gen_dirconnect_output
  ...
  end
```

a-will:
Ah, but that's only the generation. Don't forget to look at the consumers: Some of them ignore rvalid!

https://github.com/lowRISC/opentitan/blob/97151d2d2a8a7f5819a41c35f7eae291e4ec7ef1/hw/ip/prim/rtl/prim_ram_1p_scr.sv#L505

Single-cycle read latency is assumed in various parts of the code base, unfortunately.

Razer6:
Dang it. That changed a little bit. Will factor rvalid in such that it works independent of the latency. Flops are needed. They are needed everywhere 🚀

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -374,14 +376,14 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
   logic [Width-1:0] wdata_scr;
   assign wdata_scr = (mubi4_test_true_loose(write_pending_q)) ? wdata_scr_q : wdata_scr_d;
 
-  mubi4_t rvalid_q;
+  logic rvalid;
```
a-will, line 379: I suspect this was a mubi as a countermeasure. @vogelpi the mubi rvalid shouldn't be dropped, should it?

a-will: Hm... but prim_ram_1p_adv already does mubi for control signals, doesn't it? Maybe this is fine.

Razer6: Yes I thought about it. But that maybe needs a bit of discussion. In general the code uses a Mubi and then translates that back to a single bit to `rvalid_o`. Hmm It might be easier now to attack `rvalid_o` from the underlying `prim_ram_1p` instance, which is a flop compared to`rvalid_o` of `prim_ram_1p_scr` instance, which is combinational logic.

vogelpi: Hey both,
Let's try to keep this as MuBi and let's keep the flops below. If the new parameter is enabled, a second flop stage should be instantiated. Ideally the result of the output of the flop stage should be compared to the rvalid signal from the macro in in case of a mismatch we should factor this into `alert_o`. Also an alert should fire in case the flop stage doesn't hold a valid MuBi encoding. This is currently not done and I don't know why. @nasahlpa do you maybe remember? Did we try this and then ran into an issue?

vogelpi: We discussed this in a meeting and @nasahlpa brought up that the rvalid is buffered here to properly handle the address collision case (see `addr_collision_q`). This needs to be factored in as well when adding the flop stage. @Razer6 will look into this.

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -374,14 +376,14 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
...
-  mubi4_t rvalid_q;
+  mubi4_t rvalid;
   logic intg_error_r_q;
   logic [Width-1:0] wmask_q;
   always_comb begin : p_forward_mux
     rdata_o = '0;
     rvalid_o = 1'b0;
     // Kill the read response in case an integrity error was seen.
-    if (!intg_error_r_q && mubi4_test_true_loose(rvalid_q)) begin
+    if (!intg_error_r_q && mubi4_test_true_loose(rvalid)) begin
```
nasahlpa, line 386: I think this will not work when we have an address collision.

When we have (1) a write followed by a (2) read from the same address we actually do not forward the (2) read to the underlying `prim_ram_1p_adv` primitive. Instead we forward the data that is buffered in the holding register register.

However, with your code change, we still wait for the rvalid from the underlying `prim_ram_1p_adv` primitive.

So what we could do is to modify the condition here. If there is no address collision, use the `rvalid` from the underlying primitive. If there is an address collision, use the old buffered `rvalid`. I think we can leave the old `rvalid` buffer logic as it is because this should be independent of flop stages?

Razer6: I don't think that generating `rvalid` here is the right approach. That creates another assumption on the latency of the underlying RAM primitive.

nasahlpa: Oh sorry,  the second paragraph of my comment actually was referring to the case when we have an address collision. In the normal case, I also was thinking of forwarding the `rvalid` from the  underlying module. So yes, I agree with you :-)

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -184,8 +186,11 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
     .out_o(intg_error_buf)
   );
   logic macro_req;
-  assign macro_req   = ~intg_error_w_q & ~intg_error_buf &
-      mubi4_test_true_loose(mubi4_or_hi(mubi4_or_hi(read_en_buf, write_en_q), write_pending_q));
+  assign macro_req   = ~intg_error_w_q &  ~intg_error_buf &
+                        mubi4_and_hi(mubi4_test_false_loose(addr_collision_d),
```
nasahlpa, line 190: For this to work you need to add 
`import prim_mubi_pkg::mubi4_test_false_loose;`
at the beginning.

nasahlpa: And:
`mubi4_test_true_loose() `around the MuBi conditions.

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -184,8 +186,11 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
...
+                                     mubi4_test_true_loose(
+                                       mubi4_or_hi(mubi4_or_hi(read_en_buf, write_en_q),
+                                                   write_pending_q)));
```
nasahlpa, line 193:
```suggestion
  assign macro_req   = ~intg_error_w_q &  ~intg_error_buf &
                          mubi4_test_true_loose(mubi4_and_hi(mubi4_t'(~addr_collision_d),
                            mubi4_or_hi(mubi4_or_hi(read_en_buf, write_en_q),
                                          write_pending_q)));
```

Razer6: Could you elaborate on that? Doesn't that create an assumption on the encoding such that False is the inverse of true?

nasahlpa: Argh yes, true. Although we do this a couple of times in the code base, the following would be safer:
```suggestion
  assign macro_req   = ~intg_error_w_q &  ~intg_error_buf &
                          mubi4_test_true_loose(mubi4_and_hi(mubi4_bool_to_mubi(
                            mubi4_test_false_loose(addr_collision_d)),
                            mubi4_or_hi(mubi4_or_hi(read_en_buf, write_en_q),
                                          write_pending_q)));
```

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -374,29 +388,32 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
...
   always_comb begin : p_forward_mux
     rdata_o = '0;
     rvalid_o = 1'b0;
-    // Kill the read response in case an integrity error was seen.
-    if (!intg_error_r_q && mubi4_test_true_loose(rvalid_q)) begin
-      rvalid_o = 1'b1;
+    // Kill the read response in case an integrity error was seen. For integrity errors we just
+    // return zero
+    if (!intg_error_r_q) begin
+      // Regular readd from the RAM macro
```
nasahlpa, line 400:
```suggestion
      // Regular read from the RAM macro
```

nasahlpa (approved):
Thanks for the offline discussion @Razer6, this now looks good to me!

Please note that DV fails when setting `FlopRamOutput = 1`. Presumably because the tests assume a certain latency. However, for the default parameter `FlopRamOutput = 0` pass rate looks good so far:
`[Q: 013, D: 016, P: 092, F: 000, K: 000, T: 121]  76%  sram_ctrl:0.sram_ctrl_access`

rswarbrick:
As well as the notes that @nasahlpa has left, I think this needs some auto-generated files regenerating (see the slow lint CI failure).

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -406,8 +423,10 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
   ///////////////
   logic ram_alert;
 
-  assign alert_o = mubi4_test_invalid(write_en_q) | mubi4_test_invalid(addr_collision_q) |
-                   mubi4_test_invalid(write_pending_q) | mubi4_test_invalid(rvalid_q) |
+  assign alert_o = mubi4_test_invalid(write_en_q)       |
+                   mubi4_test_invalid(addr_collision_q) |
+                   mubi4_test_invalid(write_pending_q)  |
+                   mubi4_test_invalid(rvalid)           |
```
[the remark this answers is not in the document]
Razer6: This was done there before with the internal rvalid. Lets check the one we are testing now.

Razer6:
@andreaskurth Ping?

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -183,9 +185,20 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
...
+  // 1. There is no integrity error
+  // 2. There is no address collision (read after write to the same address). Here we are directly
+  //    returning the in-flight data that is waiting to be written. Note, inverting the
+  //    addr_collision_d works since True and False in the Mubi domain are complementary.
+  // 3. There is a non-conflicting read or staged write (we scrambled in between) or there is a
+  //    pending write.
   logic macro_req;
-  assign macro_req   = ~intg_error_w_q & ~intg_error_buf &
-      mubi4_test_true_loose(mubi4_or_hi(mubi4_or_hi(read_en_buf, write_en_q), write_pending_q));
+  assign macro_req = ~intg_error_w_q &  ~intg_error_buf &
```
andreaskurth, line 197: Nit: superfluous space?

hw/ip/prim/rtl/prim_ram_1p_scr.sv
```diff
@@ -183,9 +185,20 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
     .in_i(intg_error_i),
     .out_o(intg_error_buf)
   );
+
+  // Only perform a access request to the SRAM macro when:
```
andreaskurth, line 189:
```suggestion
  // Only perform a access request to the SRAM macro when all of the following hold:
```

andreaskurth (approved):
LGTM apart from two small suggestions. Thx @Razer6 :+1:

andreaskurth:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_ram_1p_adv.sv

A new output port, `rvalid_mubi_o`, gets added and driven from an existing flop.

CHANGE AUTHORIZED: hw/ip/i2c/rtl/i2c_fifos.sv
CHANGE AUTHORIZED: hw/ip/usbdev/rtl/usbdev.sv

In those files, the new `rvalid_mubi_o` output port gets added to the `prim_ram_1p_adv` instance.

CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_ram_1p_scr.sv

Structural changes:
- A new parameter, `FlopRamOutput`, gets added and connected to `prim_ram_1p_adv`'s `EnableOutputPipeline` parameter.
- The `rvalid` signal gets driven from a flop insice `prim_ram_1p_adv` instead of a separate flop in this module. The former can be optimized if parity and ECC are disabled.
- Response data muxing got simplified, potentially shortening that path.

Functional changes:
- The SRAM macro is no longer accessed for read after a write to the same address. 

CHANGE AUTHORIZED: hw/ip/sram_ctrl/data/sram_ctrl.hjson

A new parameter, `FlopRamOutput`, gets added and connected to `prim_ram_1p_scr`.

No functional change here; the `FlopRamOutput` parameter of both `sram_ctrl` instances is `0` (off).

Razer6:
There still seems to be an issue. It seems, the keystream needs to be delayed on read requests where the flopped RAM is enabled. Something like:

```systemverilog
--- a/hw/ip/prim/rtl/prim_ram_1p_scr.sv
+++ b/hw/ip/prim/rtl/prim_ram_1p_scr.sv
@@ -261,6 +261,7 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
   localparam int DataNonceWidth = 64 - AddrWidth;
   logic [NumParScr*64-1:0] keystream;
   logic [NumParScr-1:0][DataNonceWidth-1:0] data_scr_nonce;
+  logic keystream_valid;
   for (genvar k = 0; k < NumParScr; k++) begin : gen_par_scr
     assign data_scr_nonce[k] = nonce_i[k * DataNonceWidth +: DataNonceWidth];

@@ -284,7 +285,7 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
       .dec_i   ( 1'b0 ),
       // Output keystream to be XOR'ed
       .data_o  ( keystream[k * 64 +: 64] ),
-      .valid_o ( )
+      .valid_o ( keystream_valid )
     );

     // Unread unused bits from keystream
@@ -296,9 +297,27 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
   end

   // Replicate keystream if needed
-  logic [Width-1:0] keystream_repl;
+  logic [Width-1:0] keystream_repl, keystream_repl_q;
   assign keystream_repl = Width'({NumParKeystr{keystream}});

+  // Delay the keystream when data comes in delayed due to a flopped output
+  if (FlopRamOutput) begin
+    prim_flop #(
+      .Width(Width),
+      .ResetValue(0)
+    ) u_delay_keystream (
+      .clk_i,
+      .rst_ni,
+      .en_i(keystream_valid),
+      .d_i(keystream_repl),
+      .q_o(keystream_repl_q)
+    );
+  end else begin
+    logic unused_signals;
+    assign unused_signals = keystream_valid;
+    assign keystream_repl_q = keystream_repl;
+  end
+
   /////////////////////
   // Data Scrambling //
   /////////////////////
@@ -354,7 +373,7 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(

     // Apply Keystream, replicate it if needed
     assign rdata[k*DiffWidth +: LocalWidth] = rdata_xor ^
-                                              keystream_repl[k*DiffWidth +: LocalWidth];
+                                              keystream_repl_q[k*DiffWidth +: LocalWidth];
   end
```

nasahlpa:
> There still seems to be an issue. It seems, the keystream needs to be delayed on read requests where the flopped RAM is enabled. Something like:
> 
> ```diff
> --- a/hw/ip/prim/rtl/prim_ram_1p_scr.sv
> +++ b/hw/ip/prim/rtl/prim_ram_1p_scr.sv
> @@ -261,6 +261,7 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
>    localparam int DataNonceWidth = 64 - AddrWidth;
>    logic [NumParScr*64-1:0] keystream;
>    logic [NumParScr-1:0][DataNonceWidth-1:0] data_scr_nonce;
> +  logic keystream_valid;
>    for (genvar k = 0; k < NumParScr; k++) begin : gen_par_scr
>      assign data_scr_nonce[k] = nonce_i[k * DataNonceWidth +: DataNonceWidth];
> 
> @@ -284,7 +285,7 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
>        .dec_i   ( 1'b0 ),
>        // Output keystream to be XOR'ed
>        .data_o  ( keystream[k * 64 +: 64] ),
> -      .valid_o ( )
> +      .valid_o ( keystream_valid )
>      );
> 
>      // Unread unused bits from keystream
> @@ -296,9 +297,27 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
>    end
> 
>    // Replicate keystream if needed
> -  logic [Width-1:0] keystream_repl;
> +  logic [Width-1:0] keystream_repl, keystream_repl_q;
>    assign keystream_repl = Width'({NumParKeystr{keystream}});
> 
> +  // Delay the keystream when data comes in delayed due to a flopped output
> +  if (FlopRamOutput) begin
> +    prim_flop_en #(
> +      .Width(Width),
> +      .ResetValue(0)
> +    ) u_delay_keystream (
> +      .clk_i,
> +      .rst_ni,
> +      .en_i(keystream_valid),
> +      .d_i(keystream_repl),
> +      .q_o(keystream_repl_q)
> +    );
> +  end else begin
> +    logic unused_signals;
> +    assign unused_signals = keystream_valid;
> +    assign keystream_repl_q = keystream_repl;
> +  end
> +
>    /////////////////////
>    // Data Scrambling //
>    /////////////////////
> @@ -354,7 +373,7 @@ module prim_ram_1p_scr import prim_ram_1p_pkg::*; #(
> 
>      // Apply Keystream, replicate it if needed
>      assign rdata[k*DiffWidth +: LocalWidth] = rdata_xor ^
> -                                              keystream_repl[k*DiffWidth +: LocalWidth];
> +                                              keystream_repl_q[k*DiffWidth +: LocalWidth];
>    end
> ```

What exactely is not working?

Moreover, I think that also the [tlul_sram_byte](https://github.com/lowRISC/opentitan/blob/master/hw/ip/tlul/rtl/tlul_sram_byte.sv) adapter needs some adaptions. I recommend ignoring the readback countermeasure case (please add an assertion to the code base that Readback and the flop stage cannot be enabled at the same time) for now. But the partial word write logic needs some adaptions as we are making some assumptions there.

Razer6:
Darjeeling does not use the readback feature. 

> What exactely is not working?

The keystream becomes out of sync.

andreaskurth (changes requested):
I misunderstood that this PR works for your case and had focused my review on the changes that the PR contains rather than checking whether something might be missing.

To prevent an accidental merge of this PR, I'm flagging _Request changes_ for now. Let's get this working before we merge it.

Razer6:
closing as this is superseeded by https://github.com/lowRISC/opentitan/pull/26788

PR #25954, opened 2025-01-20
[englishbreakfast] Switch to 256-bit seed for compile-time random consts

vogelpi:
Previously, we were still using a shorter seed and topgen/secure_prng.py kept complaining about this. We don't really care for English Breakfast as this is never going to be taped out anyway that constantly seeing the warning is kind of annoying.

Just in case someone cares: the number added now has been generated with randomness from /dev/random using the following command:

od -vAn -N32 -tuL < /dev/random

This gives a random 32-byte number in the form of 4 64-bit unsigned integers which can then be shifted and added together.


nasahlpa (approved):
Thanks, LGTM!

PR #25957, opened 2025-01-21
[sw] Change cert template engine to pre-generated method.

sasdf:
The current certificate builder library consumes ~5KB on master (~7KB on earlgrey_1.0.0). Additionally, the immutable ROM extension duplicates the ASN.1/CBOR library, adding another ~1.5KB.

Given the proximity to the 64KB size limit, this change aims to provide a more lightweight certificate construction approach.

This PR saves about 2.6KB over 5KB, and the detailed design is described in [this doc](https://docs.google.com/document/d/1fYsukm2aqi3Dzai3Y4JJnEAhbbPc5OFJcfiHf-r803w/edit?tab=t.0).


sw/host/ot_certs/src/template/mod.rs
```diff
@@ -368,30 +368,109 @@ pub enum HashAlgorithm {
...
 #[serde(tag = "type", rename_all = "kebab-case")]
 pub enum VariableType {
     /// Raw array of bytes.
     ByteArray {
-        /// Length in bytes for this variable.
-        size: usize,
+        #[serde(flatten)]
+        size: SizeRange,
+        #[serde(flatten)]
+        int_size: Option<IntSizeRange>,
```
pamaury, line 416: what is the meaning of `int_size` for a byte array?

sw/host/ot_certs/src/template/mod.rs
```diff
@@ -368,30 +368,109 @@ pub enum HashAlgorithm {
...
     },
     /// Signed integer: such an integer is represented by an array of
     /// in big-endian.
     Integer {
-        /// Maximum size in bytes for this variable.
-        size: usize,
+        #[serde(flatten)]
+        size: SizeRange,
```
pamaury, line 422: why is there both a `size` and `int-size`? Isn't that redundant?

sw/host/ot_certs/src/template/mod.rs
sasdf: The integer and byte-array types are quite interchangeable at the moment.

Integers can be encoded as padded byte arrays (e.g., owner_intermediate_pub_key_ec_x), while byte arrays can also be encoded as big-endian integers. (e.g., owner_intermediate_pub_key_id).

We could have a follow-up PR later that either consolidates or makes a better distinction between the integer, bytes and string types.

pamaury:
Is it possible to get access to this design document? We apparently have not been in the loop of this refactoring.

sw/host/ot_certs/src/template/mod.rs
pamaury: Ok this makes sense. However I don't know if including 33 in the range makes sense: this is an implementation detail of the ANS1 encoding. If an integer can take all possible values up to 32 bytes then the implementation should internally translate that to 33 bytes of space at most?

sasdf:
Granted, I'll also check if I can make the doc available to all opentitan.org domain later.

sw/host/ot_certs/src/template/mod.rs
sasdf: For the `owner_intermediate_pub_key_id`, changing it to an integer type sounds okay, but it also introduces ambiguity with existing conversion functions (e.g., how to encode an integer as a hex string). 

```
serial_number: { var: "owner_intermediate_pub_key_id", convert: "big-endian" },
// ...
{ serial_number: { var: "owner_intermediate_pub_key_id", convert: "lowercase-hex"} },
```

OTOH, another approach would be removing the integer type altogether since both are represented as u8 byte arrays. We could then leave the `int-size` field unset if the variable will never be interpreted as an integer.

---

I agree that the current semantics are not ideal, while we still need both size fields for compile-time checks. As the signature r & s example below illustrates, the size ranges can differ.

timothytrippel:
Might also be good to update #24558 with a pointer to the doc as well.

sw/host/ot_certs/src/template/mod.rs
sasdf: Agree. However, distinguishing between 32\~33 and 32\~32 size range would be a bit tricky to achieve fixed-length template.

How about adding another bool field `tweak-msb`? when true, it would apply the MSb tweak; when false, it would increase the maximum size by one.

sasdf: Updated the design with the ‘tweak-msb’ flag, the annotation semantic would becomes more natural now.

Thanks for the helpful advice!

> In the following example, it claims that:
> 1. `owner_pub_key_id` always uses a 20-byte input, and will always be encoded as a 20-byte DER integer with its MSb set.
> 2. `cert_signature_r` always uses a 32-byte input, while the represented integer value has a negligible chance of going less than `2^(24*8)`.

```js
        owner_pub_key_id: {
            type: "byte-array",
            exact-size: 20,
            tweak-msb: true
        },
        cert_signature_r: {
            type: "integer",
            exact-size: 32,
            min-int-size: 24,
        },
```

pamaury:
I will go through the diff, it's quite large so it will take a bit of time. At a minimum, I think you need to split this into several atomic commits if possible instead of one large commit that does everything.

sw/device/silicon_creator/lib/cert/template.h
```diff
@@ -0,0 +1,160 @@
...
+extern "C" {
+#endif
+
+/**
+ * Initialize the template engine.
+ *
+ * @param out_buf Pointer to a user-provided output buffer.
+ * @param const_bytes Pointer to a pre-generated template bytes.
+ */
+#define template_init(out_buf, const_bytes)  \
```
pamaury, line 22: I don't know if using macros this way is a good idea: it's very unintuitive that a macro would create variables in the current scope who name is not even related to the input. Would it be terrible if those fields were stored in a struct and all those functions were `static inline`? It think the compiler should optimize everything anyway.

sw/device/silicon_creator/lib/cert/template.h
```diff
@@ -0,0 +1,160 @@
...
+    uint32_t _value = __builtin_bswap32((value));                       \
+    _ptr_out = template_asn1_integer_impl(_ptr_out, (tag), (tweak_msb), \
+                                          (uint8_t *)&_value, 4);       \
+  }
+
+uint8_t *template_asn1_integer_impl(uint8_t *out, uint8_t tag, bool tweak_msb,
+                                    const uint8_t *bytes_be, size_t size);
+
+/**
+ * Memo a location to be patch with the actual output size.
```
pamaury, line 129: nit: Memo*rize*, patch*ed*

sw/device/silicon_creator/lib/cert/template.h
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * The function will saved the address (last output byte + offset) for
+ * patching later, and it should be paired with a `template_patch_size_*`.
+ *
+ * @param offset Number of bytes after the last output byte.
+ */
+#define template_memo_size(offset) uint8_t *_ptr_tag = _ptr_out + (offset);
```
pamaury, line 136: Here it's very similar: this macro doesn't even look like a function because it creates a variable with a fixed name in the current scope. I think it would be cleaner if the template library defined a type and this function returned a value:
```c
typedef uint8_t *template_pos_t;
static inline template_pos_t template_save_pos(....);
```
and then the generator generates:
```
template_pos_t ptr_tag = template_save_pos(...);
/// and later.
template_patch_size_le(ptr_tag, ...);
```

sw/host/ot_certs/src/asn1/codegen.rs
```diff
@@ -105,86 +106,182 @@ impl Codegen<'_> {
     /// * `buf_name` - Name of the variable holding the pointer to the output buffer.
     /// * `buf_size_name` - Name of the variable holding the size of the output buffer.
     ///   This variable is updated by the code to hold the actual size of the data after production.
-    /// * `indent` - Identation string (one level).
-    /// * `indent_lvl` - Initial identation level.
-    /// * `constants` - Constant pool used to create necessary constants for the code.
     /// * `variables` - Description of the variable types used when producing the output.
     /// * `gen` - Closure generating the ASN1 document.
     pub fn generate(
```
pamaury, line 111: Can you document the returned tuple's content? (it was apparently forgotten in the original code)

sw/host/ot_certs/src/asn1/codegen.rs
```diff
@@ -105,86 +106,182 @@ impl Codegen<'_> {
...
+            .chunk_by(|inst| !inst.code.is_constant())
+        {
+            if !is_code {
+                let mut nbytes = 0;
+
+                for CodeChunkWithComment { code, comment } in chunk {
+                    nbytes += &code.len();
+                    match &code {
+                        CodeChunk::OpenBlock => output.push_str("{\n"),
+                        CodeChunk::CloseBlock => output.push_str("};\n"),
```
pamaury, line 142: you don't need the `;` after a scope I think

sw/host/ot_certs/src/asn1/codegen.rs
```diff
@@ -105,86 +106,182 @@ impl Codegen<'_> {
...
 
-    /// Register a constant byte array.
-    fn add_constant_byte_array(&mut self, name_hint: Option<String>, data: &[u8]) -> String {
-        // If constant already exist, do not recreate it
-        if let Some(name) = self.constants.get_var_name(&data.to_vec()) {
-            return name;
-        }
+    /// Close one level of nested block.
+    /// It should be paired with an `indent()` call.
+    fn dedent(&mut self) {
```
pamaury, line 235: nit: unindent?

sw/host/ot_certs/src/asn1/codegen.rs
```diff
@@ -626,29 +704,80 @@ impl Builder for Codegen<'_> {
...
-        self.push_str_with_indent(&format!(
-            "asn1_start_tag(&state, &{tag_name}, {});\n",
-            tag.codestring()
-        ));
-        self.push_str_with_indent("{\n");
-        self.indent_lvl += 1;
+
+        // Allocate placeholders in the output stream.
+        let len_idx = self.output.len();
+        self.push_const(None, &[]);
```
pamaury, line 718: what is the purpose of this empty const push?

sw/host/ot_certs/src/asn1/codegen.rs
```diff
@@ -626,29 +704,80 @@ impl Builder for Codegen<'_> {
...
+        let open_idx = self.output.len();
+        self.push_const(None, &[]);
```
pamaury, line 720: same here

sw/device/silicon_creator/lib/cert/template.h
timothytrippel: agreed here

timothytrippel (commented):
First pass mostly LGTM, thanks @sasdf. Agreed with @pamaury here, it would be much easier if this PR was broken into small PRs, or even a series of smaller commits for reviewing purposes. I will take another pass tomorrow.

sw/device/silicon_creator/lib/cert/template.h
sasdf: Cool, changed to inlined function, and compiler optimized to the same size.

BTW, I kept both inlined `template_push_hex` and the underlying `template_push_hex_impl`, which helps to save some bytes.

sw/host/ot_certs/src/asn1/codegen.rs
sasdf: I changed the long tuple to a struct which is more self-documented.

sw/host/ot_certs/src/asn1/codegen.rs
sasdf: This semicolon makes the formatter properly indent the blocks that only contain comments (i.e. whole tagged block compiled to constant bytes).

sw/device/silicon_creator/lib/cert/cert.c
```diff
@@ -102,8 +102,10 @@ rom_error_t cert_x509_asn1_check_serial_number(const uint8_t *cert_page_buffer,
          &cert_page_buffer[sn_bytes_offset], asn1_integer_length);
 
   // Check the serial number in the certificate matches what was expected.
+  // The first byte is skipped since the MSb might be modified for achieving
+  // fixed-length certificate.
   *matches = kHardenedBoolFalse;
-  for (size_t i = 0; i < kCertX509Asn1SerialNumberSizeInBytes; ++i) {
+  for (size_t i = 1; i < kCertX509Asn1SerialNumberSizeInBytes; ++i) {
```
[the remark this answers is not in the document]
sasdf: I plan to have a follow up PR that simplifies this `cert_x509_asn1_check_serial_number` function. Given how we generate the certificate, I think it's safe to just memcmp the offset. Also, the unused header size parsing can be removed. Let’s keep this approach for now.

sw/device/silicon_creator/lib/cert/cdi_0.hjson
```diff
@@ -10,43 +10,50 @@
         // (x and y) which are 32-bytes integers.
         owner_intermediate_pub_key_ec_x: {
             type: "integer",
-            size: 32,
+            exact-size: 32,
+            min-int-size: 24,
```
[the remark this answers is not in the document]
sasdf: This config claims the variable always uses a 32-byte input, while the represented integer value has a negligible chance of going less than 2^(24*8).

However, the variable is never encoded as an ASN1 integer, so it doesn’t matter here. I have removed them, Thanks!

sw/device/silicon_creator/lib/cert/dice.c
```diff
@@ -26,12 +26,10 @@ static ecdsa_p256_signature_t curr_tbs_signature = {.r = {0}, .s = {0}};
 static uint8_t cdi_0_tbs_buffer[kCdi0MaxTbsSizeBytes];
 static cdi_0_sig_values_t cdi_0_cert_params = {
     .tbs = cdi_0_tbs_buffer,
-    .tbs_size = kCdi0MaxTbsSizeBytes,
 };
 static uint8_t cdi_1_tbs_buffer[kCdi1MaxTbsSizeBytes];
 static cdi_1_sig_values_t cdi_1_cert_params = {
```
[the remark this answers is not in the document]
sasdf: There are also (uninitialized) signature r / s field in this param struct.

Also, I moved the initialization of r / s field to these global params, instead of the builder code to save some space.

sw/host/ot_certs/src/asn1/codegen.rs
pamaury: I see, good to know

pamaury:
Thanks for the changes @sasdf, I think with the improvements to `template.h` this is already in a good shape :)
Overall I like the approach of this PR, it's quite clever.

The downside though is that it's very easy to get it slightly wrong with an error in the generator. I would suggest that as part of a future PR, the `template.h` gains the ability to be optionally compiled in a debug mode. In this mode, the state would keep more information about the const and output buffer size, and it would check every access for out of bounds (same for all memory copy calls). Couple with some random unit tests, this would significantly increase the chances of catching subtle errors.

Final question: with this PR, is the old `asn1.{c,h}` library still useful? Should it be removed?

timothytrippel:
@pamaury I think the old asn1 lib may still be useful for compliance / testing. Wdyt?

sasdf:
@pamaury The unittest will check against the golden truth. Unexpected OoB access will likely yield a different output, so we’re mostly safe in this case.

For the input size checking, we currently rely on writing static assertions manually, which would be error-prone. I think it’s possible to create some setter-like macro/func that sets the field and asserts its size in one shot. Let me try this in a follow-up PR.

The asn1 library is mostly stripped from romext/manuf after linkage, so it could be removed later if there’s no other dependents.

sw/host/ot_certs/src/template/mod.rs
```diff
@@ -368,30 +368,143 @@ pub enum HashAlgorithm {
...
+        #[serde(flatten)]
+        int_size: Option<IntSizeRange>,
```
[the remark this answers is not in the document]
timothytrippel: where is min-int-size?

timothytrippel:
Thanks @sasdf ! Overall, I like the direction of this PR, but there are some drawbacks that must be addressed.

# Size Hints
First the additional size hints and ranges are confusing and make it very challenging to understand the HJSON files and for a new-comer to write new HJSON certificate description. I like to conceptualize this as follows. There are variables defined in the HJSON certificate config, that get mapped to templated certificate fields that are filled in at runtime by the device code that is generated:

Screenshot 2025-02-03 at 10 07 06 PM

There are four variable types supported in the HJSON:
- **byte-array**
- **integer** (which is just big-endian oriented byte array)
- string
- boolean

Only the first two (**bolded**) are of relevance to this comment for two reasons:
1. they are the only types for which substitution is used, AND
2. they have the potential to be mapped to an ASN.1 type that could have a variable size (i.e. an ASN.1 INTEGER type) that makes templating difficult.
(Note, substitution is used on boolean types, but these are not mapped to variable ASN.1 types).
Let's break them each down below.

### Byte-Array
The byte-array could be mapped to either:
1. a fixed size ASN.1 object, like a `UTF-8 STRING` or `OCTET STRING`, or
2. a variable sized ASN.1 object like an `INTEGER`. 
However, by always tweaking the MSB (i.e., ensuring it is set to 1), we can guarantee that if 2 occurs, the size is always predictable. 

### Integer
The integer type can be mapped to either:
1. a fixed size ASN.1 object, like an `OCTET STRING`, or
2. a variable sized ASN.1 object like an `INTEGER`. 
Case 1 is not an issue since since the target ASN.1 type is a fixed size. Case 2 is the only issue, but all we need to know at compile time is an upper bound. Why do we need to know a min bound?

### Concluding Questions / Requests
1. Can we revert to a single fixed size annotation on the byte-array template types?
2. Can we revert to a single upper-bound size annotation for integer template types?
3. Can we eliminate "exact-size" and just use a single "size" annotation as before that was either a fixed size or an upper bound? or rather for all types except integer have a fixed "size", and integer has a "max_size" annotation in the HJSON?

# PR Size
This PR is way too big to be a single commit PR. Like many projects, we prefer small PRs where possible (https://opentitan.org/book/doc/contributing/detailed_contribution_guide/index.html#what-should-a-pr-look-like-singlemultiple-commits). Small PRs get reviewed faster and often contain less mistakes (on both the submitter and the reviewer sides). This could have easily been broken into several small commits and PRs. Additionally, for complex changes, like this one, sometimes it can make more sense to build the new functionality alongside the old functionality, verify it works and is better than the old functionality, then tear the old functionality down. For example, one way this could have been broken down differently was:
- one commit / PR with just the new template.c/h library (and associated unittests)
- one commit / PR adding new size annotations to the various HJSON template variable types
- one commit / PR that adds the new codegen 2.0 code
- several commits that move each cert / HJSON template to the new code generation scheme
- one commit to remove the old codegen 1.0 code

Anywho, since this is mostly reviewed already, and close to acceptance, we can keep as is. But going forward, please submit smaller PRs.

sasdf:
> @timothytrippel: Case 2 is the only issue, but all we need to know at compile time is an upper bound. Why do we need to know a min bound?

We need to track the lower bound to ensure the length field is fixed-size at compile-time. In DER, elements with length 126 will have a 1-byte length field, where length 128 will be a 2-byte length field.

With the lower bound tracked, the generator can raise an error when the lower bound and upper bound encodes to different sizes, otherwise we’ll need to shift data during device runtime. This applies to both the “size” range field when encoded as bytes, or the “min-int-size” field when encoded as INTEGER. Also, tracking both upper and lower bound enables us to eliminate the padding code for the input array that is not long enough.

Based on your diagram, let’s make the following semantic changes:
1. Byte-array, when encoded as an integer, requires “tweak-msb” to be set. Otherwise it should be an integer type. This eliminates the confusing integer-related bounds on this type.
2. Integer type should not have the “tweak-msb” set, only the “min-int-size”.

That is:
```js
type: "byte-array",
exact-size: 20, // or size: [a, b],
[tweak-msb: true,]

type: "integer",
exact-size: 20, // or size: [a, b],
min-int-size: 20,
```

sw/device/silicon_creator/lib/cert/template.h
```diff
@@ -0,0 +1,217 @@
...
+ * Set the `bit_offset` bit of previous `byte_offset` byte.
+ *
+ * @param state Pointer to the template engine state.
+ * @param byte_offset Number of bytes before the last output bytes.
+ * @param bit_offset Index of the bit to be set.
+ * @param value Value of the bit.
+ */
+static inline void template_set_bit(template_state_t *state, int byte_offset,
+                                    int bit_offset, bool value) {
+  *(state->out_end - byte_offset) |= ((uint8_t) !!value) << bit_offset;
```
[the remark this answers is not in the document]
sasdf: Nice catch, the codegen will ensure the original bit is zero.

It seems more readable if the generated code calls “set_bit” instead of  “or_bit”, so I keep its name, but added a comment about the zero assumption.

sw/host/ot_certs/src/template/mod.rs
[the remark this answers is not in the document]
sasdf: Thanks, I reconstructed the settings model. Now, only ByteArray has the tweak-msb flag, while only Integer has the min-int-size hint.

timothytrippel:
> > @timothytrippel: Case 2 is the only issue, but all we need to know at compile time is an upper bound. Why do we need to know a min bound?
> 
> We need to track the lower bound to ensure the length field is fixed-size at compile-time. In DER, elements with length 126 will have a 1-byte length field, where length 128 will be a 2-byte length field.
> 
> With the lower bound tracked, the generator can raise an error when the lower bound and upper bound encodes to different sizes, otherwise we’ll need to shift data during device runtime. This applies to both the “size” range field when encoded as bytes, or the “min-int-size” field when encoded as INTEGER. Also, tracking both upper and lower bound enables us to eliminate the padding code for the input array that is not long enough.
> 
> Based on your diagram, let’s make the following semantic changes:
> 
> 1. Byte-array, when encoded as an integer, requires “tweak-msb” to be set. Otherwise it should be an integer type. This eliminates the confusing integer-related bounds on this type.
> 2. Integer type should not have the “tweak-msb” set, only the “min-int-size”.
> 
> That is:
> 
> ```js
> type: "byte-array",
> exact-size: 20, // or size: [a, b],
> [tweak-msb: true,]
> 
> type: "integer",
> exact-size: 20, // or size: [a, b],
> min-int-size: 20,
> ```

OK Thanks for the explanation this makes more sense r.e. the lower and upper bound encoding. But two things I still don't like (because all these extra fields are not very intutive, and it requires HJSON file writers to have intimate knowledge of ASN.1 DER encoding mechansims): 
 
Why cant we have the following:
```js
type: "byte-array",
size: 20, // where this is the "exact_size" always 
[tweak-msb-for-int: true,] // if destination could be an integer, we allow tweaking the msb

type: "integer",
size: [a, b], // where a is the "min-int" size and b is the "max" int size 
// No "min-int-size" nor "exact-size", and
// "exact-size" is encoded as "size: a" instead of "size: [a,b]" range syntax
```

Instead of what you have above. Will my suggestion work?

If so, could you please add a README.md to the dir that has the HJSON files, or to this https://cs.opensource.google/opentitan/opentitan/+/master:sw/host/ot_certs/README.md, to describe how to write an HJSON file and what these new size fields mean? So new folks to the project know how to create their own certs?

Thanks!

sasdf:
Sure, I'll add the documentation once we've finalized the field modeling.

#### Range variant
The variant [a, b] is currently used for the tbs variable and var-length string fields in tpm_ek. Maintaining this flexibility for potential var-length tbs seems valuable, though I agree on limiting use of this range variant.

#### `exact-size` v.s. `size`
Since `size` was referring to the upper bound previously, renaming it to `exact-size` (or potentially `exact-array-size`, and `array-size-range` for the range variant) would improve clarity, and the incompatibility also helps preventing errors during migrations or cherry-picks as compile-time errors.

The definition of this field is "the size of uint8_t array user will pass to the template engine".

#### Minimum integer value
To maintain consistency in the meaning of `size` across integer / byte-array / string types, I recommend keeping the minimum integer value as a separate field. While I'm open to better naming suggestions, current alternatives include:
* min-int-size: 24
* min-value-bytes: 24
* min-value: 0x1000000000000000000000000000000000000000000000000
* min-value-bits: 192

The definition of this field is "the minimum value of the represented integer user will pass".

timothytrippel:
Thanks for the discussion on this @sasdf. I would like to get the outstanding issues resolved so we can get this merged as it is blocking progress on the immutable ROM_EXT feature being cherry-picked to the `earlgrey_1.0.0` branch.

> Sure, I'll add the documentation once we've finalized the field modeling.
Awesome, thanks. Lets do this in a follow up PR.
 
> #### Range variant
> The variant [a, b] is currently used for the tbs variable and var-length string fields in tpm_ek. Maintaining this flexibility for potential var-length tbs seems valuable, though I agree on limiting use of this range variant.

Fair point. It is useful for shrinking the size certs that use byte-arrays for strings.

> #### `exact-size` v.s. `size`
> Since `size` was referring to the upper bound previously, renaming it to `exact-size` (or potentially `exact-array-size`, and `array-size-range` for the range variant) would improve clarity, and the incompatibility also helps preventing errors during migrations or cherry-picks as compile-time errors.

Yes, I could see this being beneficial as well, i.e., having an `exact-size` field in addition to a range field of `size: [a,b]`

> The definition of this field is "the size of uint8_t array user will pass to the template engine".
> 
> #### Minimum integer value
> To maintain consistency in the meaning of `size` across integer / byte-array / string types, I recommend keeping the minimum integer value as a separate field. While I'm open to better naming suggestions, current alternatives include:
> 
> * min-int-size: 24
> * min-value-bytes: 24
> * min-value: 0x1000000000000000000000000000000000000000000000000
> * min-value-bits: 192
> 
> The definition of this field is "the minimum value of the represented integer user will pass".

Thanks for your input here, I think you have made some compelling points above. However, I think we can reach a compromise here. At a minimum, I would like to see the following if it is possible:

```js
type: "byte-array",
exact-size: 20, // or `range-size: [a, b]` if size could be variable
[tweak-msb: true,] // if destination could be an integer

type: "integer",
exact-size: 20, // or `range-size: [a, b]`, where a is the "min-int" size and b is the "max" int size; this will only be used if the integer is variable sized.
```

The above should keep the meaning of the `range-size: [a,b]` consistent across `byte-array` and `integer` types, the `exact-size` field can be used to simplify what fields have an exact size vs when a range is possible. Please make these changes if they are possible so we can get this merged. If they are not possible, let's discuss offline so we can come to a resolution faster. Thanks!

sasdf:
Thanks @timothytrippel, I have changed the PR with the configuration format you proposed. Also, a short documentation about the variable typing has been added to sw/host/ot_certs/README.md.

timothytrippel (approved):

timothytrippel:
Thanks @sasdf , let's get this cherry-picked to the `earlgrey_1.0.0` branch.

PR #25958, opened 2025-01-21
[ujson,perso_tlv,dice] Fix potential signed integer flow, and add buffer checks

anthonychen1251:
This PR has three commits that:

1. Fixes a potential integer overflow in `ujson_parse_integer` by using an unsigned integer for parsing and adding a check before converting back to a signed integer.
2. Adds buffer size checks in `perso_tlv_data.c` to prevent potential memory corruption vulnerabilities.
3. Removes the unused cert_size output parameter in `dice_cert_check_valid`

### DIFF
--- a/sw/device/silicon_creator/lib/cert/dice.c
+++ b/sw/device/silicon_creator/lib/cert/dice.c
@@ -192,8 +192,7 @@ rom_error_t dice_cert_check_valid(const perso_tlv_cert_obj_t *cert_obj,
   // For X.509, we only check the serial_number but not public key contents.
   OT_DISCARD(pubkey);
 
-  size_t cert_size = cert_obj->cert_body_size;
-  return cert_x509_asn1_check_serial_number(cert_obj->cert_body_p, 0,
-                                            (uint8_t *)pubkey_id->digest,
-                                            cert_valid_output, &cert_size);
+  return cert_x509_asn1_check_serial_number(
+      cert_obj->cert_body_p, 0, (uint8_t *)pubkey_id->digest, cert_valid_output,
+      /*out_cert_size=*/NULL);
 }
--- a/sw/device/lib/ujson/ujson.c
+++ b/sw/device/lib/ujson/ujson.c
@@ -163,7 +163,7 @@ status_t ujson_parse_integer(ujson_t *uj, void *result, size_t rsz) {
     neg = true;
     ch = (char)TRY(ujson_getc(uj));
   }
-  int64_t value = 0;
+  uint64_t value = 0;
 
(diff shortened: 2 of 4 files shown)


sw/device/silicon_creator/lib/cert/dice.c
```diff
@@ -192,8 +192,7 @@ rom_error_t dice_cert_check_valid(const perso_tlv_cert_obj_t *cert_obj,
   // For X.509, we only check the serial_number but not public key contents.
   OT_DISCARD(pubkey);
 
-  size_t cert_size = cert_obj->cert_body_size;
-  return cert_x509_asn1_check_serial_number(cert_obj->cert_body_p, 0,
-                                            (uint8_t *)pubkey_id->digest,
-                                            cert_valid_output, &cert_size);
+  return cert_x509_asn1_check_serial_number(
```
sasdf, line 195: I think we should check the cert_body_size before calling the check_serial_number?

anthonychen1251: To clarify, are you proposing we compare the `cert_obj->cert_body_size` with the decoded size of the header using `cert_x509_asn1_decode_size_header(&cert_obj->cert_body_p[0])`?

sasdf: I'm referring a check that ensure the body size is at least the minimum size required by cert_x509_asn1_check_serial_number, since that function doesn't take a size argument.

anthonychen1251: Correct me if I'm wrong, but isn't that check already included in https://github.com/lowRISC/opentitan/blob/7240b029c24996c1dba56dcabd2a85c980e34bbb/sw/device/silicon_creator/lib/cert/cert.c#L54-L62 ? Although it will return `kErrorOk` when the size is not large enough to include a serial number.

The snippet below performs a sanity check on the certificate body size when populating the perso tlv object.
https://github.com/lowRISC/opentitan/blob/7240b029c24996c1dba56dcabd2a85c980e34bbb/sw/device/silicon_creator/manuf/base/perso_tlv_data.c#L62-L75

That's why I asked if you were proposing we compare `cert_obj->cert_body_size` with `cert_x509_asn1_decode_size_header(&cert_obj->cert_body_p[0])` (how cert_x509_asn1_check_serial_number() get the cert size actually ) here

timothytrippel (approved):
LGTM, thanks @anthonychen1251 

Can you rebase this on the latest tip of tree to ensure it passes CI checks (since it is a bit stale?), and then also cherry-pick this to the `earlgrey_1.0.0` branch when you have a moment?

PR #25959, opened 2025-01-21
[dv] Make it an error to run an intr_test sequence when there are no interrupts

rswarbrick:
The first commit tidies up the `clkmgr` hjson so that it doesn't ask us to do so (and the commit message explains what's going on). The second adds the error.

The second commit was originally part of #25864 but @alees24 (quite rightly!) thought it was worth checking it wouldn't break anything. It would have done, hence the first commit...


alees24 (approved):
LGTM.

PR #25960, opened 2025-01-21
[topgen] Properly uniquify the incoming alert names

Razer6:
Otherwise, we would end up in conflicts when adding incoming alerts from other domains, which have the same module and alert name.

### DIFF
--- a/util/topgen/lib.py
+++ b/util/topgen/lib.py
@@ -1142,8 +1142,9 @@ class TopGen:
 
         # Uniquify the incoming alert module list
         incoming_module_names = []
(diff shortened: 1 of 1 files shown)


sameo:
Should you run topgen and include the new generated files as part of this PR?

sameo (approved):

Razer6:
> Should you run topgen and include the new generated files as part of this PR?

As discussed offline, this code path is only hit when defining external alerts. Thus in upstream, there is no diff in the generated code.

vogelpi (approved):
LGTM, thanks @Razer6 !

PR #25961, opened 2025-01-21
[rom_ctrl, dv] Exclusions of non-occuring write cases in tlul_adapter…

KinzaQamar:
…_sram fifos

### DIFF
--- a/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
+++ b/hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
@@ -142,6 +142,28 @@ INSTANCE: tb.dut.u_tl_adapter_rom.u_rspfifo
 Condition 7 "1709501387" "(((~gen_normal_fifo.empty)) & ((~gen_normal_fifo.under_rst))) 1 -1"
                            (2 "10")
 
+// wvalid_i gets driven in u_tl_adapter_rom as reqfifo_wvalid by a_ack. a_ack requires
+// tl_o_int.a_ready as true to set itself true. But that signal requires wready_o=1, which is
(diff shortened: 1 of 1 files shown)


hw/ip/rom_ctrl/dv/cov/rom_ctrl_cov_excl.el
```diff
@@ -142,6 +142,28 @@ INSTANCE: tb.dut.u_tl_adapter_rom.u_rspfifo
 Condition 7 "1709501387" "(((~gen_normal_fifo.empty)) & ((~gen_normal_fifo.under_rst))) 1 -1"
                            (2 "10")
 
+// wvalid_i gets driven in u_tl_adapter_rom as reqfifo_wvalid by a_ack. a_ack requires
+// tl_o_int.a_ready as true to set itself true. But that signal requires wready_o, which is
+// reqfifo_wready in u_tl_adapter_rom. So, we can't get evalid_i without wready_o.
```
rswarbrick, line 147: I think I agree that this can't happen, but I'm not sure about your reasoning. You're right that `a_ack` requires `tl_o_int.a_ready` to be true. However, that *doesn't* imply anything about the `wready_o` signal from `u_reqfifo`. *That* signal is saying that the request FIFO is not full.

However, we only push something into the FIFO when we send a request to SRAM. We will pop it from the FIFO again on the following cycle (because the ROM always responds in a single cycle), so the request FIFO can never actually become full.

rswarbrick: I've just noticed that this is essentially the same argument as you need for the `u_rspfifo` item at the bottom. It probably makes sense to note in the comments that these are related.

rswarbrick (approved):
Thanks: this looks very clear to me!

PR #25962, opened 2025-01-21
[rom_ctrl, dv] Exclusion of a case where sramreqfifo_wready is false

KinzaQamar:
Merge this PR first as a dependency of [25869](https://github.com/lowRISC/opentitan/pull/25869)


rswarbrick (approved):
Thanks for working through this carefully. I agree :-)

PR #25963, opened 2025-01-21
[rom] Correct the hash order for SPX verify

cfrantz:
The hmac peripheral produces the SHA256 hash in reversed byte order. We need to byte-reverse the hash to put it into natural byte order before passing it to the `spx_verify` function.


jadephilipoom (approved):

cfrantz:
Addresses #25967

PR #25966, opened 2025-01-21
[topgen] Add support for instantiating uniquified ipgen IPs

Razer6:
This PR adds basic support to render and instantiate uniquified IPs from ipgen. The change is done in the following way:

1. It is assumed that all ipgen IP templates support the `module_instance_name` parameter. This is not true for all IPs there. Though this parameter was added to all `*.tpldesc` files, even it is not used.
2. In topgen, when you instantiate an ipgen'ed (`attr: "ipen"`), you specify three different fields:
   1. `template_type`: This is the basic template type that is used to render the IP from.
   2. `type`: This is the uniquified type.
   3. `name`: This is the actual instance name used in the top-level SystemVerilog file.

Let's make an example: To render a uniquified `rv_plic` you would set the following parameters in the HJSON:
```hjson
{
  template_type: "rv_plic"
  type:          "my_fancy_plic"
  name:          "my_plic_name"
  ...
}
```

This would render and create new module `my_fancy_plic` and the instantiation would look like:
```SystemVerilog
my_fancy_plic #(
   ...
 ) u_my_plic_name (
   ...
 );
 ```

Note, currently, only the AC ranges, RACL Ctrl, and the PLIC support this kind of uniquification. The alert handler would support it in theory. However, here, the template needs a few modifications as the uniquication there does create wrong file names.

### DIFF
--- a/hw/ip_templates/rstmgr/data/rstmgr.tpldesc.hjson
+++ b/hw/ip_templates/rstmgr/data/rstmgr.tpldesc.hjson
@@ -135,5 +135,11 @@
       type: "string"
       default: "lowrisc:constants:top_pkg"
     }
+    {
+      name: "module_instance_name"
+      desc: "instance name in case there are multiple pwrmgr instances. Not yet implemented."
+      type: "string"
+      default: "pwrmgr"
+    }
   ]
 }
--- a/util/topgen.py
+++ b/util/topgen.py
@@ -258,6 +258,9 @@ def generate_plic(top: Dict[str, object], out_path: Path) -> None:
     topname = top["name"]
     params = {}
 
+    # Get the PLIC instance
+    plic = lib.find_module(top['module'], 'rv_plic')
+
     # Count number of interrupts
     # Interrupt source 0 is tied to 0 to conform RISC-V PLIC spec.
     # So, total number of interrupts are the number of entries in the list + 1
@@ -267,6 +270,7 @@ def generate_plic(top: Dict[str, object], out_path: Path) -> None:
     # Target and priority: Currently fixed
     params["target"] = int(top["num_cores"], 0) if "num_cores" in top else 1
     params["prio"] = 3
+    params["module_instance_name"] = plic['type']
 
     ipgen_render("rv_plic", topname, params, out_path)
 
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

File: util/topgen.py [hunk 1]
```diff
@@ -583,29 +590,26 @@ def generate_racl(topcfg: Dict[str, object], out_path: Path) -> None:
...
-        # If we have more RACL policy groups, uniquify the control IP
-        # if len(topcfg['racl']['policies']) > 1:
-        #     params['module_instance_name'] = f'racl_ctrl_{racl_group}'
-
-        # # Only render the RACL groups that are really instantiated in that top
-        # for m in topcfg['module']:
-        #     if m['name'] == params['module_instance_name']:
-        #         ipgen_render("racl_ctrl", topname, params, out_path)
-        #         break
-        # TODO(#25673): The obove code, would be the correct if ipgen correctly supports rendering
```
a-will, line 603: Would you like this TODO with attached issue to remain in the code somewhere? (just wondering; no change needed)

a-will (approved):
LGTM

util/topgen.py
Razer6: This TODO is implicitly solved by this PR. If there are multiple RACL groups, it means that there are uniquified `racl_ctrl` instances, which was not possible before.

vogelpi: Sounds good to me.

vogelpi (approved):
LGTM

vogelpi:

This PR adds some now keys to the Earlgrey hjson but there is no RTL or functional impact of this. This is fine.

util/topgen/lib.py
```diff
@@ -855,6 +855,21 @@ def num_rom_ctrl(modules):
 
 
 def find_module(modules, type):
+    '''Returns the first module of a given type
+    For ipgen modules the base template type is used for matching
```
rswarbrick, line 859: This can be a follow-up PR, but I think the resulting functions are a bit odd here. Would it make more sense to add an optional `use_base_template_type` argument to `find_module`?

rswarbrick:

We gain some new keys in the hjson file, but these won't have any effect on the Earlgrey configuration.

util/topgen.py
Razer6: You are right ;-) Though I haven't fully decided on how to make the mental connection between the uniquified IP name and the RACL policy group. I am fully aware of that, and it will be solved properly in a follow-up PR.

vogelpi:
Thanks for your reviews, I am merging this to unblock follow-up work but it would still be good if @Razer6 could address the points you've raised.

hw/ip_templates/rstmgr/data/rstmgr.tpldesc.hjson
```diff
@@ -135,5 +135,11 @@
       type: "string"
       default: "lowrisc:constants:top_pkg"
     }
+    {
+      name: "module_instance_name"
+      desc: "instance name in case there are multiple pwrmgr instances. Not yet implemented."
```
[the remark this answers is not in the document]
Razer6: Addressed in https://github.com/lowRISC/opentitan/pull/25988

Issue #25967, opened 2025-01-21
[ROM] SLH-DSA pre-hash mode hash endianness


moidx:
### Description

The ROM implementation in `earlgrey_1.0.0` implements SLH-DSA in both pre-hash and pure modes. The pre-hash implementation requires a hash endianness that is not compatible with HSM and Cloud services. 

This issue tracks: 

- [ ] Update SLH-DSA implementation in the ROM.
- [ ] Update tooling to remove existing workarounds.

This issue proposes to implement the changes on `master` and `earlgrey_1.0.0` branches.

jadephilipoom:
I tried to root-cause this issue and I'm now a little confused where it's coming from and want to make sure we're certain the endianness is actually wrong and that flipping it in ROM is a good solution.

Originally, when I reviewed the PRs including workarounds for this, it didn't surprise me too much. Test vector suites for pre-hashed SPHINCS+ aren't available because it's too new, and ROM likes little-endian while NIST likes big-endian, so this sort of mismatch wouldn't be hard to believe.

However, when I look at the code, I see that the digest gets [passed as bytes](https://github.com/lowRISC/opentitan/blob/7240b029c24996c1dba56dcabd2a85c980e34bbb/sw/device/silicon_creator/lib/sigverify/spx_verify.c#L110), and that the OID is fed the same way and has a byte order that exactly matches FIPS-205. Now I'm confused! As far as `spx_verify` is concerned, the digest is just raw bytes, it's not treated any differently from the OID and it doesn't look like it's getting flipped in any way between then and when it's [passed to HMAC](https://github.com/lowRISC/opentitan/blob/7240b029c24996c1dba56dcabd2a85c980e34bbb/sw/device/silicon_creator/lib/sigverify/sphincsplus/hash_sha2.c#L81) later.

So my question is: how do we know that the byte order from HSM/Cloud services are correct, and could there be an error elsewhere in our tooling between that and ROM? Could I see an example message and the resulting hash from the HSM?  My preference would be for ROM to follow the byte order from the hash specifications, even if a particular HSM flips the bytes.

I'm also a little concerned that, even if the byte-order should be flipped, doing this in ROM is a little more risky than flipping the bytes in external tooling, since it involves moving around and overwriting the data that determines the digest we're checking the ROM_EXT signature against. I worry this might open an opportunity for fault attacks that, for example, block the write and cause one or more words of the digest to be zero so hash collisions are easier to find.

moidx:
[...]

PR #25969, opened 2025-01-22
[opentitantool] Correct the SHA256 hash ordering

### LINKED ISSUE #25870: [opentitantool] Fix sha256 byte ordering

cfrantz:
Opentitanlib was computing the SHA256 hash and storing it in a fixed-sized bigint.  This led to various problems dealing with the endianness of the resulting digest, including emitting digest in "little-endian" order. A SHA256 digest _is not_ an integer; it is a sequence of bytes.

1. Eliminate the bigint representation of the SHA256 digest.  Give the digest struct serialization, display and parsing traits to eliminate the need for dealing with conversions for those ues cases.  Add `to_vec` and `to_vec_rev` to produce the raw bytes when needed.
2. Add a `--spx-hash-reversal-bug` switch to opentitantool spx commands for cases where firmware erroneously uses a reversed digest.
3. Eliminate the copious digest reversals present in hsmtool.  Add a `Sha256HashReversed` input format for cases where firmware erroneously uses a reversed digest.
4. Fix the signing rules to use the reversed format only for keys tied to a firmware implementation with the reversal bug.

Fixes #25870


vbendeb:
I don't have any comments on the code, it would be good to see a test case when a signature is generated by an external entity and validated by hsmtool for all signature types or better yet by rom_ext when verifying owner's image signature.

cfrantz:
Addresses #25967

timothytrippel (approved):

cfrantz:
> I don't have any comments on the code, it would be good to see a test case when a signature is generated by an external entity and validated by hsmtool for all signature types or better yet by rom_ext when verifying owner's image signature.

I've added another commit that adds tests for opentitantool and hsmtool.  Opentitantool uses files on disk and hsmtool uses `softhsm2` as the HSM for these tests.

The tests create a signature with one tool and verify it with the other.  Interoperability between the tool-generated signatures and the firmware is tested by the `//sw/device/silicon_creator/rom_ext/e2e/verified_boot:keys` test suite and by `//sw/device/silicon_creator/rom/e2e/sigverify_spx/...`.

cfrantz:
I also manually tested the offline signing flow and the CloudKMS signing flow.

PR #25970, opened 2025-01-22
[rom_ctrl, cleanups] Minor convention changes

KinzaQamar:
** The assertion in tlul_adapter_sram was missing a _A suffix.

** The exclusion for some conditions in rom_ctrl_cov_excl.el
   didn't work as after some RTL changes their condition
   numbers got changed.


rswarbrick (commented):
I think this will need rebasing, but it looks good to me.

Could you also split it into two commits, so that the fix in `tlul_adapter_sram.sv` is separate from the stuff that's specific to `rom_ctrl`?

rswarbrick (approved):
Thank you for the tidy-up!

rswarbrick:

The change is just tweaking an assertion name to fit the style guide. This will have no function effect.

vogelpi (approved):
LGTM!

vogelpi:

The change is just tweaking an assertion name to fit the style guide. This will have no function effect.

PR #25971, opened 2025-01-22
Gpio straps doc

marcelocarvalhoLowRisc:
Documentation adjustments for GPIO Straps feature.

### DIFF
--- a/hw/ip/gpio/data/gpio.hjson
+++ b/hw/ip/gpio/data/gpio.hjson
@@ -110,7 +110,8 @@
       act:     "rcv",
       package: "",
       desc:    '''
-               This signal is pulsed high by the power manager after reset in order to sample the HW straps.
+               The strap enable signal tells gpio to take a snapshot of the input pins.
+               The behaviour of this signal after that event will have no effect.
                ''',
       default: "1'b0"
     },
@@ -410,7 +411,16 @@
       ],
     },
     { name: "HW_STRAPS_DATA_IN",
-      desc: "GPIO Input data sampled as straps during cold boot read value",
+      desc: '''
+        GPIO input data that was sampled as straps at most once after the block
+        came out of reset.
+
+        The behavior of this register depends on the GpioAsHwStrapsEn parameter.
+        - If the parameter is false then the register reads as zero.
+        - If the parameter is true then GPIO input data is sampled after reset
+          on the first cycle where the strap_en_i input is high. The
+          sampled data is then stored in this register.
+      ''',
       swaccess: "ro",
       hwaccess: "hrw",
       tags: [// Value in the register is determined by GPIO pin values that are sampled
--- a/hw/ip/gpio/doc/registers.md
+++ b/hw/ip/gpio/doc/registers.md
@@ -3,26 +3,26 @@
 <!-- BEGIN CMDGEN util/regtool.py -d ./hw/ip/gpio/data/gpio.hjson -->
 ## Summary
 
-| Name                                                       | Offset   |   Length | Description                                                   |
-|:-----------------------------------------------------------|:---------|---------:|:--------------------------------------------------------------|
-| gpio.[`INTR_STATE`](#intr_state)                           | 0x0      |        4 | Interrupt State Register                                      |
-| gpio.[`INTR_ENABLE`](#intr_enable)                         | 0x4      |        4 | Interrupt Enable Register                                     |
-| gpio.[`INTR_TEST`](#intr_test)                             | 0x8      |        4 | Interrupt Test Register                                       |
-| gpio.[`ALERT_TEST`](#alert_test)                           | 0xc      |        4 | Alert Test Register                                           |
-| gpio.[`DATA_IN`](#data_in)                                 | 0x10     |        4 | GPIO Input data read value                                    |
-| gpio.[`DIRECT_OUT`](#direct_out)                           | 0x14     |        4 | GPIO direct output data write value                           |
-| gpio.[`MASKED_OUT_LOWER`](#masked_out_lower)               | 0x18     |        4 | GPIO write data lower with mask.                              |
-| gpio.[`MASKED_OUT_UPPER`](#masked_out_upper)               | 0x1c     |        4 | GPIO write data upper with mask.                              |
-| gpio.[`DIRECT_OE`](#direct_oe)                             | 0x20     |        4 | GPIO Output Enable.                                           |
-| gpio.[`MASKED_OE_LOWER`](#masked_oe_lower)                 | 0x24     |        4 | GPIO write Output Enable lower with mask.                     |
-| gpio.[`MASKED_OE_UPPER`](#masked_oe_upper)                 | 0x28     |        4 | GPIO write Output Enable upper with mask.                     |
-| gpio.[`INTR_CTRL_EN_RISING`](#intr_ctrl_en_rising)         | 0x2c     |        4 | GPIO interrupt enable for GPIO, rising edge.                  |
-| gpio.[`INTR_CTRL_EN_FALLING`](#intr_ctrl_en_falling)       | 0x30     |        4 | GPIO interrupt enable for GPIO, falling edge.                 |
-| gpio.[`INTR_CTRL_EN_LVLHIGH`](#intr_ctrl_en_lvlhigh)       | 0x34     |        4 | GPIO interrupt enable for GPIO, level high.                   |
-| gpio.[`INTR_CTRL_EN_LVLLOW`](#intr_ctrl_en_lvllow)         | 0x38     |        4 | GPIO interrupt enable for GPIO, level low.                    |
-| gpio.[`CTRL_EN_INPUT_FILTER`](#ctrl_en_input_filter)       | 0x3c     |        4 | filter enable for GPIO input bits.                            |
-| gpio.[`HW_STRAPS_DATA_IN_VALID`](#hw_straps_data_in_valid) | 0x40     |        4 | Indicates whether the data in !!HW_STRAPS_DATA_IN is valid.   |
-| gpio.[`HW_STRAPS_DATA_IN`](#hw_straps_data_in)             | 0x44     |        4 | GPIO Input data sampled as straps during cold boot read value |
+| Name                                                       | Offset   |   Length | Description                                                             |
+|:-----------------------------------------------------------|:---------|---------:|:------------------------------------------------------------------------|
+| gpio.[`INTR_STATE`](#intr_state)                           | 0x0      |        4 | Interrupt State Register                                                |
+| gpio.[`INTR_ENABLE`](#intr_enable)                         | 0x4      |        4 | Interrupt Enable Register                                               |
+| gpio.[`INTR_TEST`](#intr_test)                             | 0x8      |        4 | Interrupt Test Register                                                 |
+| gpio.[`ALERT_TEST`](#alert_test)                           | 0xc      |        4 | Alert Test Register                                                     |
+| gpio.[`DATA_IN`](#data_in)                                 | 0x10     |        4 | GPIO Input data read value                                              |
+| gpio.[`DIRECT_OUT`](#direct_out)                           | 0x14     |        4 | GPIO direct output data write value                                     |
+| gpio.[`MASKED_OUT_LOWER`](#masked_out_lower)               | 0x18     |        4 | GPIO write data lower with mask.                                        |
+| gpio.[`MASKED_OUT_UPPER`](#masked_out_upper)               | 0x1c     |        4 | GPIO write data upper with mask.                                        |
+| gpio.[`DIRECT_OE`](#direct_oe)                             | 0x20     |        4 | GPIO Output Enable.                                                     |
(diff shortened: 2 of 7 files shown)


hw/ip/gpio/data/gpio.hjson
```diff
@@ -111,6 +111,7 @@
       package: "",
       desc:    '''
                This signal is pulsed high by the power manager after reset in order to sample the HW straps.
+               It can only be pulsed high one-time after reset. The following pulses will not be taken into consideration.
```
rswarbrick, line 114: I think this sounds like a requirement on the integrator. Don't we want something more like "This will only have an effect the first time it goes high after reset" ? (And I'm not sure about "pulse" in the line: we probably don't have an opinion about whether it has to drop again)

hw/ip/gpio/data/gpio.hjson
```diff
@@ -390,7 +391,14 @@
       ],
     },
     { name: "HW_STRAPS_DATA_IN",
-      desc: "GPIO Input data sampled as straps during cold boot read value",
+      desc: '''
+        GPIO input data that was sampled as straps as the block came out of reset.
```
rswarbrick, line 395: Nit: I'd avoid the double "as" with something like "... sampled as straps when the block ..."

hw/ip/gpio/data/gpio.hjson
```diff
@@ -390,7 +391,14 @@
...
+
+        This register depends on the GpioAsHwStrapsEn parameter.
```
rswarbrick, line 397: I guess the register exists either way, so maybe "The behaviour of this register depends ..." ?

hw/ip/gpio/data/gpio.hjson
```diff
@@ -390,7 +391,14 @@
...
+        - If the parameter is false then the register reads as zero.
+        - If the parameter is true then GPIO input data is sampled on the first cycle after reset
+          where the strap_en_i input is high. That sampled data is stored in this register.
```
rswarbrick, line 400: I'd suggest reordering this to "... is sampled after reset on the first cycle where the strap_en_i input is high."

That avoids it sounding like we only do this for special sorts of reset.

hw/ip/gpio/doc/theory_of_operation.md
```diff
@@ -73,14 +73,12 @@ the output driver is unable to switch the pin or during the delay imposed
...
-The GPIO controller also provides an additional optional feature to sample the GPIO input values as hardware configuration straps
-These input values are sampled only at initial cold boot reset deassertion and is not sampled thereafter.
-Thus any change in the input configuration information after power on reset deassertion will not be captured.
-A strap_en signal driven by the pwrmgr is used for this sampling mechanism. The strap_en pulse is a single cycle pulse of the io div4 clock
-It is expected to toggle only once at cold boot.
-
-The snapshot value of the GPIO inputs is available in the [`HW_STRAPS_DATA_IN`](../data/gpio.hjson#hw_straps_data_in) register for firmware read out
-Note that all 32 bits of input data to the GPIO controller are sampled in this mode regardless of the gpio output enable status
+
+The GPIO controller includes an optional feature to sample GPIO input values as hardware configuration straps. After the reset (cold boot), when the strap_en signal is asserted, the GPIO detects the rising edge of the signal and triggers the sampling of GPIO input values one clock cycle later. These sampled values are stored in the HW_STRAPS_DATA_IN register.
```
rswarbrick, line 77: A couple of style nits:

- The line is really long because there are missing newlines after the full stops.
- As with my comment on the hjson change, I think you might need to reorder things to avoid it sounding like "a reset when strap_en is asserted", which isn't what we want.
- And the sentence that starts "After the reset..." is rather long. Probably worth breaking.
- I'd also avoid talking about "After the reset...". Can't this be more operational? "After each reset, on the first cycle when strap_en is asserted, ..." ?
- Also "rising edge" is dangerous: it sounds a bit like we're talking about a clock.
- Don't talk about "cold boot" in this bit: that's a system-level concern. (I realise that the previous text did so, but we can do better!)

hw/ip/gpio/doc/theory_of_operation.md
```diff
@@ -73,14 +73,12 @@ the output driver is unable to switch the pin or during the delay imposed
...
+
+The strap_en signal is driven by the power manager (pwrmgr) and is expected to toggle only once after the reset process. Sampling occurs only during this period, and subsequent changes to the GPIO input configuration will not be captured.
```
rswarbrick, line 79: Is this chunk actually necessary? I think I'd be inclined to drop it, since that's very much a concern of stuff outside of the block.

rswarbrick:
Nuts: I clicked in the wrong place and didn't leave a summary note!

Thanks for documenting this carefully: I think it looks like a great improvement. Once it's stable, we can get someone from Rivos to check it matches the design they intended.

hw/ip/gpio/data/gpio.hjson
marcelocarvalhoLowRisc: Text updated.

hw/ip/gpio/data/gpio.hjson
```diff
@@ -110,7 +110,9 @@
       act:     "rcv",
       package: "",
       desc:    '''
-               This signal is pulsed high by the power manager after reset in order to sample the HW straps.
+               The strap enable signal is used to make the snapshot of the gpio_in.
```
rswarbrick, line 113: English nit:
- It's a signal to make a snapshot, rather than a mechanism by which to make the snapshot.
- gpio_in probably doesn't need a "the". (But I'd weaken the text slightly to...)

Maybe better: "The strap enable signal tells gpio to take a snapshot of the input pins" ?

hw/ip/gpio/data/gpio.hjson
```diff
@@ -110,7 +110,9 @@
...
+               It can trigger only one-time after the power-on reset.
```
rswarbrick, line 114: I don't think this is quite true: that sounds like a constraint on the behaviour of the port. I think you want to say something like "This only has an effect the first time it goes high".

hw/ip/gpio/doc/theory_of_operation.md
```diff
@@ -73,14 +73,20 @@ the output driver is unable to switch the pin or during the delay imposed
...
-A strap_en signal driven by the pwrmgr is used for this sampling mechanism. The strap_en pulse is a single cycle pulse of the io div4 clock
-It is expected to toggle only once at cold boot.
-
-The snapshot value of the GPIO inputs is available in the [`HW_STRAPS_DATA_IN`](../data/gpio.hjson#hw_straps_data_in) register for firmware read out
-Note that all 32 bits of input data to the GPIO controller are sampled in this mode regardless of the gpio output enable status
+
+The GPIO controller provides an optional feature to sample GPIO input values as hardware configuration straps.
+
+After each reset, on the first cycle when the `strap_en_i` signal is asserted, the GPIO detects its
+transition from low to high. One clock cycle later, it samples the GPIO input
```
rswarbrick, line 80: Style: Markdown files are written with one sentence to a line.

hw/ip/gpio/doc/theory_of_operation.md
```diff
@@ -73,14 +73,20 @@ the output driver is unable to switch the pin or during the delay imposed
...
+values and stores them in the `HW_STRAPS_DATA_IN` register.
+
+The `strap_en_i` signal transition from low to high is expected only once
+after the reset process. Sampling occurs exclusively during this time,
```
rswarbrick, line 84: Nit: I think you want "... exclusively at this time ..." (because the signal transition might be instantaneous, so you can't do something during the time)

hw/ip/gpio/doc/registers.md
```diff
@@ -358,7 +358,13 @@ Indicates whether the data in [`HW_STRAPS_DATA_IN`](#hw_straps_data_in) is valid
 |   0    |   ro   |   0x0   | HW_STRAPS_DATA_IN_VALID |               |
 
 ## HW_STRAPS_DATA_IN
-GPIO Input data sampled as straps during cold boot read value
+GPIO input data that was sampled as straps when the block came out of reset.
+
+The behavior of this register depends on the GpioAsHwStrapsEn parameter.
+- If the parameter is false then the register reads as zero.
+- If the parameter is true then GPIO input data is sample after reset
+on the first cycle where the strap_en_i input is high.
```
martin-velay, line 366: Nit: I can see here and in some other locations in most of the Markdown files of this IP, that some sentences have been broken over multiple lines. I thought the guideline/rule was to have "one sentence per line". This is something I plan to use for the vplan, I feel it would be better to arrange these files to be aligned with the "one sentence per line" guideline. But probably in another PR. @rswarbrick, what do you think?

martin-velay (approved):
Looks OK for me. I just have a remark about the line breaks, but this is not a comment for this PR, but more for the whole IP

hw/ip/gpio/data/gpio.hjson
rswarbrick: I think this suggestion is just adding "input" to my text? Either is fine, but something needs to change because the text currently in the PR says that the strap enable signal is the mechanism by which something makes a snapshot, rather than an instruction that tells something to make a snapshot.

hw/ip/gpio/data/gpio.hjson
```diff
@@ -110,7 +110,9 @@
       act:     "rcv",
       package: "",
       desc:    '''
-               This signal is pulsed high by the power manager after reset in order to sample the HW straps.
+               The strap enable signal is used to make the snapshot of the gpio_in.
+               This only has an effect the first time it goes high.
+               The following strap_en driven after that, it will not be taken into consideration.
```
rswarbrick, line 115: I think this text is a bit wonky. Can't you just say "The behaviour of this signal after that event will have no effect" ?

hw/ip/gpio/doc/registers.md
rswarbrick: I think this has now been reflowed (thanks, @marcelocarvalhoLowRisc!).

hw/ip/gpio/doc/theory_of_operation.md
```diff
@@ -73,14 +73,22 @@ the output driver is unable to switch the pin or during the delay imposed
...
-A strap_en signal driven by the pwrmgr is used for this sampling mechanism. The strap_en pulse is a single cycle pulse of the io div4 clock
-It is expected to toggle only once at cold boot.
-
-The snapshot value of the GPIO inputs is available in the [`HW_STRAPS_DATA_IN`](../data/gpio.hjson#hw_straps_data_in) register for firmware read out
-Note that all 32 bits of input data to the GPIO controller are sampled in this mode regardless of the gpio output enable status
+
+The GPIO controller provides an optional feature to sample GPIO input values as hardware configuration straps.
+
+After each reset, on the first cycle when the `strap_en_i` signal is asserted, the GPIO detects its
+transition from low to high.
```
rswarbrick, line 80: Three broken sentences from here "After each reset..."; "One clock cycle..."; "This mechanism samples..."

hw/ip/gpio/doc/theory_of_operation.md
```diff
@@ -73,14 +73,19 @@ the output driver is unable to switch the pin or during the delay imposed
...
-The GPIO controller also provides an additional optional feature to sample the GPIO input values as hardware configuration straps
-These input values are sampled only at initial cold boot reset deassertion and is not sampled thereafter.
-Thus any change in the input configuration information after power on reset deassertion will not be captured.
-A strap_en signal driven by the pwrmgr is used for this sampling mechanism. The strap_en pulse is a single cycle pulse of the io div4 clock
-It is expected to toggle only once at cold boot.
-
-The snapshot value of the GPIO inputs is available in the [`HW_STRAPS_DATA_IN`](../data/gpio.hjson#hw_straps_data_in) register for firmware read out
-Note that all 32 bits of input data to the GPIO controller are sampled in this mode regardless of the gpio output enable status
+
+The GPIO controller provides an optional feature to sample GPIO input values as hardware configuration straps.
```
rswarbrick, line 77: Thanks for avoiding the breaks inside the sentences. But I think now we have one sentence per paragraph :-) Can you group the lines up together and remove some blank lines?

rswarbrick (commented):
Thanks for the changes you have made. There's one more nitty change, and I think this needs rebasing over changes in `interfaces.md`?

hw/ip/gpio/doc/theory_of_operation.md
marcelocarvalhoLowRisc: fixed, let me know what you think =)

rswarbrick (approved):

rswarbrick:
CHANGE AUTHORIZED: hw/ip/gpio/data/gpio.hjson

This is just a documentation improvement: no risk to existing designs.

marcelocarvalhoLowRisc:
> Thanks, this looks good to me.

Thanks!!!

HU90m:
CHANGE AUTHORIZED: hw/ip/gpio/data/gpio.hjson

PR #25972, opened 2025-01-22
[doc,docker] update util/container to Ubuntu 22.04 and stop mentioning 20.04

nbdd0121:
Getting this fixed so people don't have new Ubuntu 20.04 installations anymore.

We'll need to deprecate the support of Ubuntu 20.04 before it goes EOL this April.


nasahlpa (approved):
Thanks!

PR #25973, opened 2025-01-22
[RACL] Add raclgen

davidschrammel:
This commit is based on #25749 and adds a command line utility for RACL.
The first 10 commits are from #25749 and will be dropped once that PR has been merged.


util/raclgen.py
```diff
@@ -0,0 +1,110 @@
...
+                        '-i',
+                        required=True,
+                        help='Path to IP block hjson file.')
+
+    parser.add_argument('--mapping',
+                        '-m',
+                        required=True,
+                        help='Path to RACL mapping hjson file.')
+
+    parser.add_argument(
```
Razer6, line 35: This argument alignment looks a little bit off.

util/raclgen.py
```diff
@@ -0,0 +1,110 @@
...
+    except ValueError as err:
+        raise SystemExit(
+            f'Failed to parse RACL config "{args.racl_config}". Error: {str(err)}'
+        )
+
+    parsed_register_mapping, racl_group, policy_names = parse_racl_mapping(
+        parsed_racl_config, args.mapping, args.if_name, ip_block)
+
+    template = """
+<%doc>
```
Razer6, line 60: Do we need that as a mako comment? Isn't a python comment enough?

Razer6: But probably to have the same chunk of template in both cases, with the same comment I guess.

Razer6 (approved):
Thanks @davidschrammel! This looks good to me.

util/raclgen.py
davidschrammel: It does seem off, but it was generated by `util/lintpy.py`.

util/raclgen.py
```diff
@@ -0,0 +1,99 @@
...
+    out = StringIO()
+    out.write(
+        Template(template).render(m=ip_block,
+                                  if_name=args.if_name,
+                                  register_mapping=parsed_register_mapping,
+                                  policy_names=policy_names,
+                                  module_name=ip_block.name,
+                                  racl_group=racl_group).rstrip())
+    print(out.getvalue())
+    out.close()
```
rswarbrick, line 95: I think I don't quite understand the StringIO bit here. What's the difference between this and `print(Template(template).render(...))` ?

util/raclgen/lib.py
```diff
@@ -101,11 +103,13 @@ def compute_policy_value(permission: str) -> int:
     return racl_config
 
 
-def parse_racl_mapping(racl_config: Dict[str, object], mapping_path: str,
-                       if_name: Optional[str],
-                       ip_block: IpBlock) -> Dict[str, int]:
+def parse_racl_mapping(
+        racl_config: Dict[str,
+                          object], mapping_path: str, if_name: Optional[str],
```
rswarbrick, line 108: My god, the automatic formatter makes a right mess of things like this. Eugh.

util/raclgen/lib.py
```diff
@@ -101,11 +103,13 @@ def compute_policy_value(permission: str) -> int:
...
+        ip_block: IpBlock) -> Tuple[Dict[str, int], str, List[str]]:
+
     mapping = _read_hjson(mapping_path)
-    parsed_register_mapping = {}
+    parsed_register_mapping = OrderedDict()
```
rswarbrick, line 112: I don't think I understand why this needs to change. What's the advantage of using `OrderedDict` here?

util/raclgen/lib.py
```diff
@@ -118,38 +122,34 @@ def parse_racl_mapping(racl_config: Dict[str, object], mapping_path: str,
         raise SystemExit('RACL group must be a string')
     if not isinstance(register_mapping, dict):
         raise SystemExit('Register mapping must be a a dict')
+    if racl_group not in racl_config['policies']:
+        raise SystemExit(f'RACL group {racl_group} not defined in RACL config')
+
+    policy_names = [
+        policy['name'] for policy in racl_config['policies'][racl_group]
+    ]
```
rswarbrick, line 130: I think this would be a bit cleaner as something like this:
```python
    racl_policy = racl_config['policies'].get(racl_group)
    if racl_policy is None:
        raise SystemExit(f'RACL group {racl_group} not defined in RACL config')

    policy_names = [policy['name'] for policy in racl_policy]
```

(I don't have an opinion about the variable name that I picked!)

util/topgen/templates/toplevel_racl_pkg.sv.tpl
```diff
@@ -89,15 +89,48 @@ package top_racl_pkg;
...
+   *   Register to policy mapping:
+        % for reg_name, policy_idx in register_mapping.items():
+   *     ${f"{reg_name}:".ljust(reg_name_len+1)} ${policy_names[policy_idx]} (Idx ${f"{policy_idx}".rjust(policy_idx_len)})
+        % endfor
+      % endif
    */
-<% if_suffix = f"_{if_name.upper()}" if if_name else "" %>\
-  parameter int unsigned RACL_POLICY_SEL_${m["name"].upper()}${if_suffix} [${len(mapping)}] = '{${", ".join(map(str, reversed(mapping.values())))}};
-  
+  parameter int unsigned RACL_POLICY_SEL_${m["name"].upper()}${group_suffix}${if_suffix} [${len(register_mapping)}] = '{${", ".join(map(str, reversed(register_mapping.values())))}};
```
rswarbrick, line 132: To avoid this line getting really long, it might be a good idea to generate some local variables in the Mako file.

rswarbrick (commented):
Some nitty Python comments, but this looks good. And thanks for finding and fixing a silly bug in `intermodule.py`.

util/raclgen.py
```diff
@@ -0,0 +1,99 @@
...
+        parsed_racl_config = parse_racl_config(args.racl_config)
+    except ValueError as err:
+        raise SystemExit(
+            f'Failed to parse RACL config "{args.racl_config}". Error: {str(err)}'
+        )
+
+    parsed_register_mapping, racl_group, policy_names = parse_racl_mapping(
+        parsed_racl_config, args.mapping, args.if_name, ip_block)
+
+    template = """
```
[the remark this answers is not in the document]
davidschrammel: That was the original intent.
However, initial testing showed that using `include` might not be trivial because the mako version that we're using doesn't support `include` and I couldn't get it to work even after updating it.
I plan to submit a follow-up PR to deduplicate the two templates when I get `include` working.

util/raclgen.py
davidschrammel: [Some other utils](https://github.com/lowRISC/opentitan/blob/220025a6cb7de10f60aaf9a84256a7d5a333eb52/hw/ip/rv_timer/util/reg_timer.py#L36) use the same code with a comment `# Determine output: if stdin then stdout if not then ??`. We can remove the StringIO for now and change it later when/if we decide that we want to be able to write the output to a file.

util/raclgen/lib.py
davidschrammel: Do we want to adhere to the automatic formatter in such cases or should I change it?

util/raclgen.py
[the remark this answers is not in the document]
rswarbrick: Oh, how annoying! It's a bit surprising though: I did a stupid search of the Mako changelog and can see mentions of the feature from back in 2016...

util/raclgen.py
rswarbrick: Thanks! Honestly, the code that did this in the first place seems rather silly. If they wanted to support something cleaner in future, couldn't they have used `sys.stdout.write(...)`? Anyway, probably worth forgetting about :-)

util/raclgen/lib.py
rswarbrick: I don't believe that there's a style requirement to use an automatic formatter. My personal preference is not to, but it's just a personal preference!

Sorry, you can ignore the comment. It was just a disgusted exclamation :-)

util/raclgen/lib.py
rswarbrick: Cool, that sounds sensible. Also, I think we'll have the same ordering anyway (`dict` keeps insertion order, since Python 3.6)

davidschrammel: Apparently `hjson.load()` is already using OrderedDicts (According to the ordering in the hjson file). So just using OrderedDict in lib.py should be sufficient and no other data structure changes are needed.

davidschrammel:
@rswarbrick I have updated the commits according to your comments and rebased them on master (now that #25749 has been merged). Could you take a look at this again?

rswarbrick (approved):
This looks good, thanks!

PR #25976, opened 2025-01-22
[manuf] build individualization binaries for sim_dv

timothytrippel:
This enables building the individualization binaries for the sim_dv exec_env to stand up ATE test flows.


cfrantz (approved):

sha-ron:
Thank you @timothytrippel. Does it mean that the code under '  if (kDeviceType == kDeviceSilicon' 
 won't be executed?  For example, https://github.com/lowRISC/opentitan/blob/earlgrey_1.0.0/sw/device/silicon_creator/manuf/base/sram_ft_individualize.c#L202

timothytrippel:
yes that is correct. I udpated that behavior in https://github.com/lowRISC/opentitan/pull/25985

LMK what you think.

PR #25979, opened 2025-01-22
Allow lowrisc_opentitan bazel module to be used as a non-root module by external workspaces.

korran:


a-will:
Thanks for contributing, Kor! This is a topic of interest these days, hehe.

You'll need to sign your git commits to get CI beyond the lint stage. (as in `git commit --signoff`, where it would append something like `Signed-off-by: Kor Nielson <user@domain>` to the end of the commit message)

jwnrt (commented):
This looks good to me, I'm personally fine with merging the `MODULE` files if it lets us be used as a dependency. Thanks a lot.

Would you mind reformatting `MODULE.bazel` so that the sections are:

1. All `bazel_dep`s at the top.
2. All overrides (`git_override`. and `single_version_override`s).
3. The rest of the extensions and toolchain registrations.

You can also drop the "original from `x.MODULE.bazel`" comments if you like, they won't be useful anymore. Feel free to add "Rust configuration", "Python configuration" headers to separate the sections if you like though.

rules/extensions.bzl
```diff
@@ -7,14 +7,16 @@ load("@bazel_tools//tools/build_defs/repo:local.bzl", "local_repository")
 def _hooks_impl(rctx):
     for mod in rctx.modules:
         for repo in mod.tags.repo:
-            dir = rctx.getenv(repo.env, repo.dummy)
-            local_repository(name = repo.name, path = dir)
+            d = rctx.getenv(repo.env, str(rctx.path(repo.dummy)))
+            # (hack) Make sure the dir is actually populated by reading a file
```
pamaury, line 11: Can you explain why this is necessary?

pamaury:
Thanks @korran for this contribution, this looks very useful. I don't know if we should make this part of this PR, but I think some kind of CI check to make sure that we do not regress would be very useful. Otherwise it's only a matter of time before we split another `@//something` instead of `@lowrisc_opentitan//` which will break the repo.

Ping @jwnrt @cfrantz on that.

rules/extensions.bzl
korran: I've found that when doing clean builds, intermittently bazel hasn't populated this directory before local_repository() is executed, and it fails with a file-not-found error.

FWIW, I don't think local_repository() should be abused to do what you want here, as it shouldn't be used to look at internal bazel-managed files. I think it's much safer to to iterate over the files in this label and manually put them into the repo with rctx.file().

korran:
> You'll need to sign your git commits to get CI beyond the lint stage. (as in `git commit --signoff`, where it would append something like `Signed-off-by: Kor Nielsen <user@domain>` to the end of the commit message)

I've added sign-offs to all the commits.

rules/extensions.bzl
korran: I've changed this logic. Calling local_repository() on files from a label ends up putting horrible absolute paths into MODULE.bazel.lock, which is unacceptable. Instead, I've changed it to use a custom repository_rule that references the label instead of a path.

korran: Also, I wonder if this might be a lot simpler if instead of doing these environment-variable tricks, we just used a bazel module for these things, letting users override on the command line with `--override_module=provisioning_exts=some/path`.

pamaury: Hi Korran, this rule is there for mostly for pre-blzmod backward compatibility. We discussed switch to module overrides [previously](https://github.com/lowRISC/opentitan/pull/25772#issuecomment-2569489175). Maybe we should accelerate the transition. @a-will @jwnrt

jwnrt: I know that somebody recently had issues with this being an extension since there was no way to add extra dependencies in the hook repo. I think making this an overridable module would solve that problem. Now would be a great time to do that.

a-will:
Looks like it just needs a...

```
./bazelisk.sh run //quality:buildifier_fix
```

...and then it might pass all the CI checks.

jwnrt:
@korran I'm interested in getting this merged, would you mind if I rebased it and fixed the conflicts?

cfrantz:
I had a go at rebasing this on the latest master.

- I think I've resolved the merge-skew with some of the now-inlined `MODULE.bazel` files.
- The `rom_hooks` configuration is not compatible with using lowrisc_opentitan as a dependency of another project.  `local_path_override`s are only respected in the root of the project.  We'll need to come up with a better solution.

cfrantz:
> Regarding the last point, I am not sure why `rom_hooks` is a bazel dependency in the first place, instead of using an overridable repo like we do for others hooks?

Should we change the `rom_hooks` into an overridable repo like the other hooks repos until we can implement something better?

pamaury:
> > Regarding the last point, I am not sure why `rom_hooks` is a bazel dependency in the first place, instead of using an overridable repo like we do for others hooks?
> 
> If you're talking about the hooks above, the problem is mostly around their use of environment variables, especially with generic identifiers that have a significant probability of collision. That still isn't really fixed with a generic `rom_hooks` module name either, though. 😆
> 
> An aside: To me, environment variables are a worse mechanism to pass configuration compared to command args (which, in bazel's case, can also be passed with a `.bazelrc` file). They require a separate environment setup script, and issues can be a little harder to debug, since the command invocation is typically more readily available.

The current use of environment variable is mostly for backward compatiblity since we didn't want to break the workflow of downstream, but as a matter of fact I believe they currently can be overridden using a bazel CLI argument instead of env vars already. Since we already moved the code to a bazel extension, so it should be relatively straighforward to modify the [code](https://github.com/lowRISC/opentitan/blob/master/rules/extensions.bzl) to allow registration using a local path directly. Coupled with using the repo as a dependency, this would allow registration of the rom hooks in an out-of-tree MODULE.bazel for example, or it could be override using a command-line flag.

cfrantz:
I've rebased this again and have made the following modifications:

1. I've fixed the filename/subdir calculation in `opentitan_top_dt_gen`.  The prior method of calculating the output directory does not work correctly when `lowrisc_opentitan` is a dependency of another bazel workspace.
2. I've changed the `rom_hooks` repo to use the existing `hooks` infrastructure.

rules/extensions.bzl
```diff
@@ -4,17 +4,35 @@
...
 
+def _copy_from_label_impl(rctx):
+    src_dir = rctx.path(rctx.attr.src)
+    for p in src_dir.readdir():
+        contents = rctx.read(p)
+        rel_path = str(p)[len(str(src_dir)):].lstrip("/")
+        rctx.file(rel_path, contents)
+
+
+copy_from_label = repository_rule(
```
pamaury: This rule should be documented

rules/extensions.bzl
```diff
@@ -4,17 +4,35 @@
 
 load("@bazel_tools//tools/build_defs/repo:local.bzl", "local_repository")
 
+def _copy_from_label_impl(rctx):
+    src_dir = rctx.path(rctx.attr.src)
+    for p in src_dir.readdir():
```
pamaury, line 9: This will only work on files (not directories), if you want to handle them, you could use symlinks, as in `bare_repository`.

rules/extensions.bzl
```diff
@@ -4,17 +4,35 @@
...
+        contents = rctx.read(p)
+        rel_path = str(p)[len(str(src_dir)):].lstrip("/")
```
pamaury: I think you can use `.basename` on `p` to achieve the same but I haven't tested it.

rules/autogen.bzl
```diff
@@ -209,15 +209,32 @@ def opentitan_autogen_dif(name, top, ip, deps = [], target_compatible_with = [])
...
     outputs = []
-    outdir = "{}/{}".format(ctx.bin_dir.path, ctx.label.package)
+    outdir = None
 
     groups = {}
     for group, files in ctx.attr.output_groups.items():
         deps = []
         for file in files:
-            deps.append(ctx.actions.declare_file(file))
+            f = ctx.actions.declare_file(file)
```
pamaury, line 221: Possibly an alternative to this hack is another construction which I used for doxygen:
https://github.com/lowRISC/opentitan/blob/master/rules/doxygen.bzl#L229
Essentially you prefix everything with `outdir` and declare a unused directory `outdir/_ignore` so that you can get a path to `outdir`. Feel free to ignore.

pamaury:
In relation to this PR, the issue of overrides could be tackled with a custom registry. I created an PR to experiment on it there: #27404. I don't have much time to work on it at the moment though

rules/autogen.bzl
cfrantz: I like that better.  Fixed.

rules/extensions.bzl
cfrantz: Fixed.

rules/extensions.bzl
cfrantz: Fixed.

rules/extensions.bzl
cfrantz: Fixed.

cfrantz:
> Requires regenerating the lock file with `cargo update -w --manifest-path third_party/mdbook/Cargo.toml`.

Done.


> Style-wise is there a reason to use `@lowrisc_opentitan//...` for some of these paths instead of `//...`? I don't think it's an issue, but would prefer consistency if they're the same.

Apparently, prefixing the patch labels with `@lowrisc_opentitan` is required when opentitan is used as a bazel dep of another project.  I've fixed all of the `patches = [ ... ]` clauses to use the repo name.

PR #25984, opened 2025-01-23
[rom] Examine the immutable section and skip if invalid 

cfrantz:
1. Add tests that check if a valid signed ROM_EXT with an invalid immutable section can boot.  The correct behavior is that the ROM should determine that the immutable section is invalid and try the other slot.
2. Refactor immutable section verification into a separate function.
3. When booting, examine the immutable section and fail the boot with an error if the immutable section is invalid.  If the failure occurs in the primary slot, the ROM should attempt to boot the secondary slot.

### DIFF
--- a/sw/device/silicon_creator/rom/rom.c
+++ b/sw/device/silicon_creator/rom/rom.c
@@ -542,7 +542,9 @@ static rom_error_t rom_measure_otp_partitions(
  * @return rom_error_t Result of the operation.
  */
 OT_WARN_UNUSED_RESULT
-static rom_error_t rom_boot(const manifest_t *manifest, uint32_t flash_exec) {
+static rom_error_t rom_boot(const manifest_t *manifest,
+                            uintptr_t imm_section_entry_point,
+                            uint32_t flash_exec) {
   CFI_FUNC_COUNTER_INCREMENT(rom_counters, kCfiRomBoot, 1);
   HARDENED_RETURN_IF_ERROR(sc_keymgr_state_check(kScKeymgrStateReset));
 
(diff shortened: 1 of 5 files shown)


sw/device/silicon_creator/rom/rom.c
```diff
@@ -706,17 +712,26 @@ static rom_error_t rom_boot(const manifest_t *manifest, uint32_t flash_exec) {
...
     for (size_t i = 0; i < kHmacDigestNumWords; ++i) {
-      HARDENED_CHECK_EQ(immutable_rom_ext_hash.digest[i],
-                        actual_immutable_section_digest.digest[i]);
+      if (immutable_rom_ext_hash.digest[i] !=
+          actual_immutable_section_digest.digest[i]) {
+        verify_result = kErrorRomImmSection;
+      }
+    }
+    // If address translation is enabled, adjust the entry_point.
+    if (launder32(manifest->address_translation) == kHardenedBoolTrue) {
```
timothytrippel, line 721: Just to confirm: this adjustment of the immutable section entry point based on the address translation enablement setting being moved to after the hashing of the immutable section is a nop right? I.e. moving of this block (on 721--727) from (682--689) should have no impact?

timothytrippel (approved):

sw/device/silicon_creator/rom/rom.c
cfrantz: No - its required.  The addresses to hash must be mapped.  Since we're now performing the imm_section check before `rom_boot` (which sets up the remap window and ePMP), we have to perform the digest over our view of the section in flash.

After we've checked the digest, we can then adjust the entry address for translation.  We won't attempt to access via the translated address (in `rom_boot` until after the remap window and ePMP are configured).

PR #25986, opened 2025-01-23
[RACL] Instantiate RACL for Darjeeling

Razer6:
This PR initially instantiates RACL for Darjeeling by:
1. Providing an initial RACL configuration
2. Routing the policy to the top-level (for the support of externally subscribed RACL IPs)

Note 1: This PR is based on https://github.com/lowRISC/opentitan/pull/25749 and needs to be rebased after that one is merged.

Note 2: This PR currently has one issue, the RACL error and error log would need to be rendered as an array and not as a single element.


hw/top_darjeeling/ip/soc_proxy/rtl/rivos_bat.sv
```diff
@@ -0,0 +1,76 @@
+// Copyright lowRISC contributors (OpenTitan project).
```
vogelpi, line 1: This file doesn't seem to belong to this PR as it's a different feature. Is it even required in this PR or can it be separated out?

vogelpi (commented):
I've just reviewed the last two commits (the other ones are part of another PR already). This looks mostly good but I have some questions.

hw/top_darjeeling/ip/soc_proxy/rtl/rivos_bat.sv
Razer6: That probably was mistakenly added during rebase.

hw/top_darjeeling/data/xbar_mbx.hjson
```diff
@@ -80,12 +80,19 @@
       reset:      "rst_mbx_ni",
       pipeline:   false,
     },
+    { name:       "racl_ctrl",
```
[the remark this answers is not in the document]
Razer6: We add racl_ctrl to the CTN facing port for the following reason:

* In a multi-chiplet design, it allows the primary ROT access and configure the secondary's ROT policy registers.

The same is true for the AC Ranges. After both instanciations are merged, I will rename the mbx crossbar to CTN crossbar.

hw/top_darjeeling/data/racl.hjson
```diff
@@ -0,0 +1,29 @@
...
+  error_response: true
+  ctn_uid_bit_lsb: 0
+  ctn_uid_bit_msb: 4
+  role_bit_lsb: 5
+  role_bit_msb: 8
+  roles: {
+    "ROT" :  { role_id: 0 }
+    "Role1": { role_id: 1 }
+    "SOC":   { role_id: 2 }
+  }
```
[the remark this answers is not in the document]
Razer6: Yes, for the upstream we need to agree on a default set. It's not necessary to define all. The integrator defines its own custom roles and policies though.

hw/top_darjeeling/rtl/autogen/top_darjeeling.sv
```diff
@@ -3030,6 +3067,10 @@ module top_darjeeling #(
     .tl_mbx_pcie1__soc_o(mbx_pcie1_soc_tl_d_req),
     .tl_mbx_pcie1__soc_i(mbx_pcie1_soc_tl_d_rsp),
 
+    // port: tl_racl_ctrl
+    .tl_racl_ctrl_o(racl_ctrl_tl_req),
+    .tl_racl_ctrl_i(racl_ctrl_tl_rsp),
+
```
vogelpi, line 3115: Should this additional newline here better be removed?

hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv
```diff
@@ -71,8 +71,73 @@ package top_racl_pkg;
...
+    rsvd[8:5] = racl_role;
+    rsvd[4:0] = ctn_uid;
+    return rsvd;
+  endfunction
+
+  /**
+   * RACL Roles
+   */
+  parameter racl_role_t RACL_ROLE_ROT   = 4'h0;
+  parameter racl_role_t RACL_ROLE_ROLE1 = 4'h1;
```
vogelpi, line 91: To align things and avoid confusion, should we enforce roles and policies to be ALL_CAPS everywhere? I see in the hjson, we have ROT, SOC and Role1 for the roles while the policies are ALL_CAPS in both cases. I would prefer to align things while we still can without touching too many files.

hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv
```diff
@@ -71,8 +71,73 @@ package top_racl_pkg;
...
+  parameter racl_role_t RACL_ROLE_SOC   = 4'h2;
+
+  /**
+   * Policies for group Null
+   */
```
vogelpi, line 96: Is this related to the additional Null entry in the hjson?

hw/ip_templates/racl_ctrl/rtl/racl_ctrl.sv.tpl
```diff
@@ -2,27 +2,34 @@
...
+  input  prim_alert_pkg::alert_rx_t [NumAlerts-1:0]                    alert_rx_i,
+  output prim_alert_pkg::alert_tx_t [NumAlerts-1:0]                    alert_tx_o,
   // Output policy vector for distribution
-  output top_racl_pkg::racl_policy_vec_t                       racl_policies_o,
+  output top_racl_pkg::racl_policy_vec_t                               racl_policies_o,
   // RACL violation information.
-  input logic            [NumSubscribingIps-1:0]               racl_error_i,
-  input top_racl_pkg::racl_error_log_t [NumSubscribingIps-1:0] racl_error_log_i
+  input logic                          [NumSubscribingIps-1:0]         racl_error_i,
+  input top_racl_pkg::racl_error_log_t [NumSubscribingIps-1:0]         racl_error_log_i,
```
vogelpi, line 29: Nit: you can drop `top_racl_pkg` here as it's important above already.

hw/top_darjeeling/data/top_darjeeling.hjson
```diff
@@ -1089,36 +1089,6 @@
...
-        { struct:  "racl_error_log",
-          type:    "uni",
-          name:    "racl_error_log",
-          act:     "req",
-          width:   "1"
-          package: "top_racl_pkg",
-          desc:    '''
-            Error log information from all IPs.
-          '''
-        }
```
vogelpi: Why are you removing those again here? It doesn't seem to belong to this commit which is on adding support for external racl errors.

hw/top_darjeeling/ip_autogen/racl_ctrl/rtl/racl_ctrl_reg_top.sv
```diff
@@ -603,6 +603,10 @@ module racl_ctrl_reg_top
       .en_i ( 1'b1          ),
       .out_o( racl_role_vec )
     );
+    // For the static RACL assignment for racl_ctrl only one role (ROT_PRIVATE) is used,
+    // leaving others unread. Intentionally read them to avoid lintint errors.
```
vogelpi, line 607: Nit: lintint -> linting / lint

hw/top_darjeeling/data/racl/racl.hjson
```diff
@@ -0,0 +1,33 @@
...
+  ctn_uid_bit_msb: 4
+  role_bit_lsb: 5
+  role_bit_msb: 8
+  roles: {
+    "ROT" :  { role_id: 0 }
+    "Role1": { role_id: 1 }
+    "SOC":   { role_id: 2 }
+  }
+ policies: {
+    Null: [
```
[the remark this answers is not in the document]
vogelpi: I think in the spec you're using "default_group" instead of "Null". Should we switch to that instead?

vogelpi (commented):
I've done a first (almost complete) pass and this mostly looks good to me @Razer6 . I have some comments regarding aligning things with the spec but nothing major.

What is left is reviewing the commit that hooks up the mailboxes. I want to carefully review the mappings to properly understand this. I'll get back to this later today in say 1h but don't want to block you.

hw/top_earlgrey/templates/toplevel.sv.tpl
```diff
@@ -502,15 +502,28 @@ max_intrwidth = (max(len(x.name) for x in block.interrupts)
...
-    .RaclPolicySelVec(top_racl_pkg::RACL_POLICY_SEL_${m["name"].upper()}${if_suffix}),
+<%
+        register_mapping = m['racl_mappings'][if_name]['register_mapping']
+        window_mapping = m['racl_mappings'][if_name]['window_mapping']
+        racl_group = m['racl_mappings'][if_name]['racl_group']
+        group_suffix = f"_{racl_group.upper()}" if racl_group and racl_group != "Null" else ""
+        if_suffix = f"_{if_name.upper()}" if if_name else ""
+        if_suffix2 = f"{if_name.title()}" if if_name else ""
+        policy_sel_name = f"RACL_POLICY_SEL_{m['name'].upper()}{group_suffix}{if_suffix}"
+%>\
```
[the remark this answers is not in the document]
davidschrammel: I believe this was just copied from [hw/top_darjeeling/templates/toplevel.sv.tpl](https://github.com/lowRISC/opentitan/blob/b63a8de29f0376c9980d08c8fb9e2d32d3c9237b/hw/top_darjeeling/templates/toplevel.sv.tpl#L502)

The code is indented 2 spaces further than the `for` above. So in my opinion this looks right. But I have not found many other templates in this code base with such nested `<% %>` blocks. Please do let me know if there is a better way to write such templates.

hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv
davidschrammel: I believe this is under specified at the moment. The RACL specification uses "Role1" in an example. But the templates enforce upper-casing: [toplevel_racl_pkg.sv.tpl](https://github.com/lowRISC/opentitan/blob/b63a8de29f0376c9980d08c8fb9e2d32d3c9237b/util/topgen/templates/toplevel_racl_pkg.sv.tpl#L88)

hw/top_darjeeling/data/racl/racl.hjson
```diff
@@ -0,0 +1,33 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+{
```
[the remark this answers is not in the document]
Razer6: I think there is a mix and match at so many places. I try to not use commas, as this is one of the advantages of HJSON vs JSON.

hw/ip_templates/racl_ctrl/rtl/racl_ctrl.sv.tpl
Razer6: Unfortunately not. For out-of-tree connections, we use a custom stiching-tool that relies on the full-scope package type (and we'll use that IP externally). I removed the import at the very top and added the import within the module body.

hw/top_darjeeling/data/top_darjeeling.hjson
Razer6: Previously, the _single_ error signal was routed to the top-level interface via this AST mechanism where you define a signal in the AST struct. See the connection from below:
```
      'ast.racl_policies'     : ['racl_ctrl.racl_policies']
      'ast.racl_error'          : ['racl_ctrl.racl_error']
      'ast.racl_error_log'    : ['racl_ctrl.racl_error_log']
```   
and the Verilog:
```
  assign ast_racl_policies = racl_policies_i;
  assign racl_error_o = ast_racl_error;
  assign racl_error_log_o = ast_racl_error_log;
  ```
This does now work properly. Instead, the internal error vector and the external error vector are now 2 different inputs to racl_ctrl (internally, they are combined again). This allows me to use the external connection directly from the racl_ctrl IP and not via this AST thing:
```
        'racl_ctrl.racl_policies'              : 'racl_policies',
        'racl_ctrl.racl_error_external'        : 'racl_error',
        'racl_ctrl.racl_error_log_external'    : 'racl_error_log'
```

File: hw/top_darjeeling/rtl/autogen/top_darjeeling.sv (hunk 4 above)
Comment (Razer6) [reply hw/top_darjeeling/rtl/autogen/top_darjeeling.sv:3115]: This is related to xbar template instantiation and not related to RACL. So I don't think that this is the right place to change that.

**Review by vogelpi — APPROVED** [review]
I've now also finished reviewing the rest. I do have more questions but no further requests for changes.

Sorry for all the questions, I am asking because I would like to better understand RACL and it's possibilities. I've always found it challenging to properly understand RACL. The more I reviewed this PR, the more I started to understand (or at least I believe to understand!). One thing that I for example got wrong all the time is what we need to do if we want to change ACLs while the system is running. Only now I understand that we don't modify the policy that a particular set of registers use (the policy selection is fixed at compile time), and we don't modify the roles of the hosts (only some of them can change the role based on e.g. debug being enabled or whatever), but what we change is the policy-specific permission bitmaps stored inside racl_ctrl.

With this in mind, it feels not very intuitive to call a non-default policy SOC_ROT which allows registers with this policy being read and written by the ROT and SOC roles exclusively (as in example in the spec). Because if those registers need to be written by another role (e.g. for initial configuration during boot up or for debugging), `racl_ctrl` would change the permission bitmap of the SOC_ROT policy to enable access by another role. Or in other words, the policy is not tied to the roles, but policies are tied to registers. The selection is done at compile time based on what we think will be needed under normal operating conditions in the field. And from that perspective, the name makes again sense. I think this is the point that I missed previously and studying this PR cleared up many many question marks :-)

File: hw/top_darjeeling/rtl/autogen/top_darjeeling.sv (hunk 4 above)
Comment (vogelpi) [reply hw/top_darjeeling/rtl/autogen/top_darjeeling.sv:3134]: I agree, thanks for explaining!

File: hw/top_darjeeling/data/top_darjeeling.hjson (hunk 8 above)
Comment (vogelpi) [reply hw/top_darjeeling/data/top_darjeeling.hjson]: This makes sense, thanks for explaining Robert.

File: hw/top_earlgrey/templates/toplevel.sv.tpl (hunk 11 above)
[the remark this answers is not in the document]
Comment (vogelpi) [reply hw/top_earlgrey/templates/toplevel.sv.tpl:526]: You're probably right. Maybe one could argue that the `% for` above starts with `%` so the indentation here is 4 instead of 2. I don't know. But let's not worry unless someone else comes along who knows more about this.

File: hw/top_darjeeling/data/racl/racl.hjson (hunk 12 above)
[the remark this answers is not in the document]
Comment (vogelpi) [reply hw/top_darjeeling/data/racl/racl.hjson:5]: Makes sense, and you're consistent with that in the new files. That sounds good to me.

File: hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv (hunk 5 above)
Comment (vogelpi) [reply hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv:91]: Thanks David. I think we want either ALL_CAPS or CamelCase in the SV file according to our style guide: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#constants

These parameters here are true constants and ALL_CAPS sounds right. I would recommend using the same spelling in the hjson to avoid confusion and align things.

File: hw/top_darjeeling/data/racl/racl.hjson (hunk 10 above)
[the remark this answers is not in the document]
Comment (Razer6) [reply hw/top_darjeeling/data/racl/racl.hjson:17]: Well, the spec does not cover those implementation details. `Null` is used to specify the default (unnamed) racl group. When specifying a name, this then yields a custom name. The current design relies on that but this is just naming. We can change that in the future.

File: hw/top_darjeeling/data/racl/soc_rot_mapping.hjson [hunk 13]
```diff
@@ -0,0 +1,9 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+{
```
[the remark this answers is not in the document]
davidschrammel: > mapping that applies to all registers in all instances

This is only true for modules where their `racl_mappings` field points to this mapping file, yes.

In my understanding, it is only "kind of a default" since we use this exact same file (or mapping) in multiple modules.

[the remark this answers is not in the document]
Razer6: Discussed also offline

hw/top_darjeeling/data/racl/all_rd_wr_mapping.hjson
```diff
@@ -0,0 +1,9 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+{
+  Null: {
+    "*": "ALL_RD_WR"
+  }
+}
```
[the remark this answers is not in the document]
Razer6: Discussed offline, we will agree on a reasonable public RACL config.

hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv
Razer6: Done.

hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv
Razer6: Discussed offline, there will be a follow up.

hw/top_darjeeling/data/racl/racl.hjson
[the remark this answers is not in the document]
Razer6: Discussed, there will be a follow up.

hw/top_darjeeling/rtl/autogen/top_racl_pkg.sv
vogelpi: Thanks!

hw/top_darjeeling/data/top_darjeeling.hjson
```diff
@@ -878,6 +878,9 @@
         core: {hart: "0x22000000"},
         soc:  {soc_mbx: "0x01465000"},
       },
+      racl_mappings: {
+        soc: 'top_darjeeling/data/racl/all_rd_wr_mapping.hjson'
+      }
```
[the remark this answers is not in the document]
vogelpi: Perfect, thanks for explaining @davidschrammel .

Razer6:
🎉

PR #25990, opened 2025-01-23
[Darjeeling] Configure Darjeeling clocks

Razer6:
This PR changes the Darjeeling clocks to:

* main: 1 GHz
* IO: 1 GHz
* USB (Not used): 1 GHz
* AON: 62.5 MHz

From IO, `clkmgr` derives a 500 MHz and 250 MHz Clock.

The `clkmgr` needed a tweak in the template. To measure the clock frequency, the clkmgr has some measurement circuitry. These regs are preloaded with defaults to (freq / f_aon) +- 10. For the 250 MHz IO_DIV4 clock, this ratio becomes 4 - 10, which is negative. This change ensures the lower value is clipped at 0.

### DIFF
--- a/hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
+++ b/hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
@@ -601,7 +601,7 @@
           bits: "${min_msb}:${max_msb+1}",
           name: "LO",
           desc: "Min threshold for ${src} measurement",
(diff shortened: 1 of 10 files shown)


alees24 (approved):
Thanks for this, and for clarifying the clocking scheme. Let's just wait until CI is happy.

Razer6:
One thing that is missing here is the clock generators within the AST need to be tweaked to output those frequencies.

PR #25995, opened 2025-01-23
[opentitantool] Erase header of other slot during rescue

jesultra:
I believe I have seen cases where the slot not being flashed during rescue still booted afterwards, meaning that the previous code flashing a zero-length image to the other slot may be ineffective.  This PR changes the default behavior to flashing an 4kB image consisting entirely of 0xFF to the other slot, in order to make sure that the header of any previous image is overwritten and invalidated.

### DIFF
--- a/sw/host/opentitantool/src/command/rescue.rs
+++ b/sw/host/opentitantool/src/command/rescue.rs
@@ -98,15 +98,15 @@ impl CommandDispatch for Firmware {
             rescue.set_baud(rate)?;
         }
         rescue.wait()?;
-        rescue.update_firmware(self.slot, payload)?;
         if self.erase_other_slot {
-            // Erase the other slot by programming an empty blob.
+            // Invalidate the other slot by overwriting its header.
             if self.slot == BootSlot::SlotB {
-                rescue.update_firmware(BootSlot::SlotA, &[])?;
+                rescue.update_firmware(BootSlot::SlotA, &vec![0xFF; 2048])?;
             } else {
-                rescue.update_firmware(BootSlot::SlotB, &[])?;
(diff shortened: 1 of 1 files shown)


sw/host/opentitantool/src/command/rescue.rs
```diff
@@ -100,11 +100,11 @@ impl CommandDispatch for Firmware {
         rescue.wait()?;
         rescue.update_firmware(self.slot, payload)?;
         if self.erase_other_slot {
-            // Erase the other slot by programming an empty blob.
+            // Invalidate the other slot by overwriting its header.
             if self.slot == BootSlot::SlotB {
-                rescue.update_firmware(BootSlot::SlotA, &[])?;
+                rescue.update_firmware(BootSlot::SlotA, &vec![0xFF; 4096])?;
```
timothytrippel, line 105: Is there a reason you write 4k? flash pages are 2k

timothytrippel (approved):

sw/host/opentitantool/src/command/rescue.rs
jesultra: No, I just pulled some number.  Let me change it to 2k.

Also, I wonder whether ROM_EXT will allow owner configuration to specify that rescue (of slot A) must be accompanied by erasing certain flash pages, either info, other data pages on slot A, or data pages of slot B?  Vadim B. mentioned that ChromeOS thinks it is a security concern to be able to rescue, without erasing filesystem or other areas.

If ROM_EXT automatically erases "the other" slot, then maybe it is a better idea if I move this optional erase step to happen *before* writing the actual image.

PR #25999, opened 2025-01-24
[assert] Add ASSERT_KNOWN asserts to multiple new outputs

Razer6:
This adds ASSERT_KNOWN assertions to many outputs of the newly introduced signals.

It also uncovered a bug. The RACL error may be `'x` because the internal address hit depends on the incoming TLUL address, which might be `'x` at some cases. To fix this, the `a_valid` signal is factored in to the RACL error computation.

### DIFF
--- a/hw/ip/i2c/rtl/i2c.sv
+++ b/hw/ip/i2c/rtl/i2c.sv
@@ -175,6 +175,9 @@ module i2c
   `ASSERT_KNOWN(IntrAcqStretchKnownO_A, intr_acq_stretch_o)
   `ASSERT_KNOWN(IntrUnexpStopKnownO_A, intr_unexp_stop_o)
   `ASSERT_KNOWN(IntrHostTimeoutKnownO_A, intr_host_timeout_o)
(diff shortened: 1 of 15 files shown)


vogelpi (approved):
Thanks @Razer6 this looks good to me!

Nits:
- Some comments mention that the outputs need to be known after reset. But AFAIK the SVA actually checks they're known all the time.
- One commit message has nbx in the title, I think it should be mbx.

vogelpi:

This PR only adds missing ASSERT_KNOWN assertions. This doesn't have functional impact.

Razer6:
> * Some comments mention that the outputs need to be known after reset. But AFAIK the SVA actually checks they're known all the time.
> * One commit message has nbx in the title, I think it should be mbx.

I changed the `mbx` commit message. As discussed offline, the comment about after the reset is correct. SVA's are disabled during reset.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/i2c/rtl/i2c_reg_top.sv
CHANGE AUTHORIZED: hw/ip/uart/rtl/uart_reg_top.sv

This PR is just adding some extra assertions (that we originally missed). No change in functional behaviour.

PR #26000, opened 2025-01-24
[hw,clkmgr,rtl] Properly size the measurement counter

Razer6:
Darjeeling has different clock ratios. This causes the registers to have a different size than the counter of the measurement circuit. That happens because the threshold value is exactly a power of two, which requires one bit more.
Change the width of the counter do be able to hold the maximum value.

### DIFF
--- a/hw/ip_templates/clkmgr/rtl/clkmgr.sv.tpl
+++ b/hw/ip_templates/clkmgr/rtl/clkmgr.sv.tpl
@@ -392,7 +392,9 @@ rg_srcs = list(sorted({sig['src_name'] for sig
 % for i, src in enumerate(rg_srcs):
 <%
  freq = all_srcs[src]['freq']
- cnt = int(freq*2 / aon_freq)
+ # One bit margin, same bit width as in the reg top
+ bit_width = int(freq / aon_freq).bit_length() + 1
+ cnt = 2**bit_width
  sel_idx = f"Clk{Name.from_snake_case(src).as_camel_case()}Idx"
 %>
   clkmgr_meas_chk #(
--- a/hw/top_darjeeling/ip_autogen/clkmgr/rtl/clkmgr.sv
+++ b/hw/top_darjeeling/ip_autogen/clkmgr/rtl/clkmgr.sv
@@ -549,7 +549,7 @@
   end
 
   clkmgr_meas_chk #(
-    .Cnt(8),
+    .Cnt(16),
     .RefCnt(1)
   ) u_io_div4_meas (
     .clk_i,
@@ -576,7 +576,7 @@
 
 
   clkmgr_meas_chk #(
-    .Cnt(32),
+    .Cnt(64),
     .RefCnt(1)
   ) u_main_meas (
     .clk_i,
@@ -603,7 +603,7 @@
 
 
   clkmgr_meas_chk #(
-    .Cnt(32),
+    .Cnt(64),
     .RefCnt(1)
   ) u_usb_meas (
     .clk_i,
--- a/hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr.sv
+++ b/hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr.sv
@@ -559,7 +559,7 @@
   end
 
   clkmgr_meas_chk #(
-    .Cnt(960),
+    .Cnt(1024),
     .RefCnt(1)
   ) u_io_meas (
     .clk_i,
@@ -586,7 +586,7 @@
 
 
   clkmgr_meas_chk #(
-    .Cnt(480),
+    .Cnt(512),
     .RefCnt(1)
   ) u_io_div2_meas (
     .clk_i,
@@ -613,7 +613,7 @@
 
 
   clkmgr_meas_chk #(
-    .Cnt(240),
+    .Cnt(256),
     .RefCnt(1)
   ) u_io_div4_meas (
     .clk_i,
@@ -640,7 +640,7 @@
 
 
   clkmgr_meas_chk #(
-    .Cnt(1000),
+    .Cnt(1024),
     .RefCnt(1)
   ) u_main_meas (
     .clk_i,
@@ -667,7 +667,7 @@
 
 
   clkmgr_meas_chk #(
-    .Cnt(480),
+    .Cnt(512),
     .RefCnt(1)
   ) u_usb_meas (
     .clk_i,
--- a/hw/top_englishbreakfast/ip_autogen/clkmgr/rtl/clkmgr.sv
+++ b/hw/top_englishbreakfast/ip_autogen/clkmgr/rtl/clkmgr.sv
@@ -554,7 +554,7 @@
   end
 
   clkmgr_meas_chk #(
-    .Cnt(960),
+    .Cnt(1024),
     .RefCnt(1)
   ) u_io_meas (
     .clk_i,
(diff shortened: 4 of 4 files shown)


vogelpi (approved):
This looks good to me, thanks @Razer6 !

`CntWidth` only changes for power-of-two values of `Cnt` which isn't the case for Earlgrey. `RefCntWidth` also doesn't change for Earlgrey (`RefCnt` is always 1).

vogelpi:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_clock_meas.sv

No functional change for Earlgrey.

rswarbrick:
Hmm, this prim looks pretty strange to me (unchanged by your PR). In particular, the way we update the counter:
```systemverilog
  logic cnt_ovfl;
  logic [CntWidth-1:0] cnt;
    if (!rst_ni) begin
      cnt <= '0;
      cnt_ovfl <= '0;
    end else if (!cnt_en && |cnt) begin
      cnt <= '0;
      cnt_ovfl <= '0;
    end else if (valid_o) begin
      cnt <= '0;
      cnt_ovfl <= '0;
    end else if (cnt_ovfl) begin
      cnt <= '{default: '1};
    end else if (cnt_en) begin
      {cnt_ovfl, cnt} <= cnt + 1'b1;
    end
  end
```
Won't this always blip `cnt` to zero for a cycle if it overflows? Then `fast_o` will have a cycle of being low.

@vogelpi (since you've also just looked at this file!): Am I missing something?

hw/ip/prim/rtl/prim_clock_meas.sv
```diff
@@ -19,8 +19,8 @@ module prim_clock_meas #(
   parameter int RefCnt = 1,
   parameter bit ClkTimeOutChkEn = 1,
   parameter bit RefTimeOutChkEn = 1,
-  localparam int CntWidth = prim_util_pkg::vbits(Cnt),
-  localparam int RefCntWidth = prim_util_pkg::vbits(RefCnt)
+  localparam int CntWidth = prim_util_pkg::vbits(Cnt + 1),
+  localparam int RefCntWidth = prim_util_pkg::vbits(RefCnt + 1)
```
rswarbrick, line 23: I'm not sure the second param needs changing. It's just used for the `cnt_ref` signal, which has to be able to count to `RefCnt-1`.

vogelpi:
> Hmm, this prim looks pretty strange to me (unchanged by your PR). In particular, the way we update the counter:
> 
> ```systemverilog
>   logic cnt_ovfl;
>   logic [CntWidth-1:0] cnt;
>   always_ff @(posedge clk_i or negedge rst_ni) begin
>     if (!rst_ni) begin
>       cnt <= '0;
>       cnt_ovfl <= '0;
>     end else if (!cnt_en && |cnt) begin
>       cnt <= '0;
>       cnt_ovfl <= '0;
>     end else if (valid_o) begin
>       cnt <= '0;
>       cnt_ovfl <= '0;
>     end else if (cnt_ovfl) begin
>       cnt <= '{default: '1};
>     end else if (cnt_en) begin
>       {cnt_ovfl, cnt} <= cnt + 1'b1;
>     end
>   end
> ```
> 
> Won't this always blip `cnt` to zero for a cycle if it overflows? Then `fast_o` will have a cycle of being low.
> 
> @vogelpi (since you've also just looked at this file!): Am I missing something?

I think you're correct. AFAIU, there are at least wo issues here:
1. The `valid_o` pulse is only generated while `cnt` is not zero. So assume `cnt` has any value and we get the `valid` signal (assuming `RefCnt == 1` which is true in our cases). Then `valid_o` asserts and we reset `cnt` to zero. The next reference pulse and value signal is expected to come in 2^`CntWidth` cycles later. At that point `cnt` will be zero again (it overflowed) and we miss the valid pulse / don't generate valid_o.
2. The `max_cnt` and `min_cnt` are also `CntWidth` bits wide. If `Cnt` is indeed a power of two. We don't have any bits to express a maximum value greater than `Cnt`. I think this is actually a real issue that is not solved yet by this PR. From the register file we get 10 bits for every measurement unit (independent of the actual ratio) and is sufficient for the current values in Darjeeling. However, `clkmgr_meas_chk` and `prim_clock_meas` then only take the `CntWidth` LSBs. That doesn't seem right. At least, this PR should also update the `CntWidth` definition in `clkmgr_meas_chk`.

@Razer6 can you confirm that this PR really fixes the issue you're seeing? I don't think the high threshold really works to be honest.

Razer6:
@vogelpi After investigating it further, I think that was not the right solution. The discrepancy in bit width came from a different computation of that for for the register in the reg top and the bit width of the counter. I unified this computation. This yields different maximum counter values for Earlgrey. But since they are in the same bit range, this is acceptable and does not change the width of the counter.

hw/ip_templates/clkmgr/rtl/clkmgr.sv.tpl
```diff
@@ -392,7 +392,9 @@ rg_srcs = list(sorted({sig['src_name'] for sig
 % for i, src in enumerate(rg_srcs):
 <%
  freq = all_srcs[src]['freq']
- cnt = int(freq*2 / aon_freq)
+ # One bit margin, same bit width as in the reg top
+ bit_width = int(freq / aon_freq).bit_length() + 1
+ cnt = 2**bit_width
```
rswarbrick, line 397: I'm trying to make sure I understand why this is the right calculation (for arbitrary frequencies). My understanding:
- We want the `clkmgr_meas_chk` to count how many ticks of the fast clock happen for each tick of the AON clock.
- We know roughly what the ratio of the two frequencies should be (which comes from `freq / aon_freq`)
- We're instantiating `clkmgr_meas_chk` with a reference count of 1, so the ratio is just measured by counting the faster clock.
- The desired ratio will give us some count that we want to program into `clkmgr_meas_chk` (using bounds in `src_min_cnt_i` and `src_max_cnt_i`)
- The counter in the module needs to be big enough to hold that value.
- This is done by taking the logarithm of the `Cnt` and `RefCnt` parameter (using `$vbits` in `prim_clock_meas.sv`)

If I've got that right, we want to be able to represent something like `freq / aon_freq * (1 + epsilon)` (presumably for epsilon a smallish positive real). Assuming that epsilon < 1, it's enough to be able to count to `2 * freq / aon_freq`.

I think that's where the factor of two in the old code come from. I'm not completely sure I understand the derivation for the new count. Could you expand the comment (or the commit message) to explain why this is the right width for the counter?

vogelpi (approved):
Thanks for taking another look at this @Razer6 and @rswarbrick . I've now done another review and the change as well as the explanation now looks good to me.

The change for the `Cnt` parameter in Earlgrey doesn't have any impact on the actual design. Because
- `Cnt` is only used to generate `CntWidth` through a `vbits()` call and the output of this function doesn't change for the change in this PR.
- Inside `prim_clock_meas.sv` `Cnt` is also used to compute the `ClkRatio` parameter which is used for a timeout check. However, inside the clkmgr variant in Earlgrey, these checks are turned of via the `RefTimeOutChkEn` parameter which is 0.

vogelpi:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr.sv

No functional RTL change for Earlgrey. Just parameters change but the functions using those parameters (vbits) don't change outputs for the new values.

rswarbrick:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr.sv

No functional RTL change for Earlgrey. Just parameters change but the functions using those parameters (vbits) don't change outputs for the new values.

PR #26002, opened 2025-01-24
[dv] Avoid imports into class scope

### LINKED ISSUE #26001: Some code in OpenTitan uses "import" at top-level in a class
### Description

This isn't allowed (see footnote 11 in section A.10 of the 1800 spec). It turns out that many EDA tools don't mind, which is why we hadn't noticed. But Questa generates an error.

I believe that there are just two instances of this, with potential examples found by searching for files that contain "class" and an "import" line:
```
git grep -l 'import ' '*.sv' >with-import.lst
grep <with-import.lst -v autogen | xargs grep -l '^\s*class' | sort -u >both.lst
```
There were 26 hits and I think two genuine examples.

rswarbrick:
There are detailed notes in the commits, but I *think* this fixes every example where we do the problem.

As such, this fixes #26001.

### DIFF
--- a/hw/ip/uart/dv/env/uart_env_cov.sv
+++ b/hw/ip/uart/dv/env/uart_env_cov.sv
@@ -5,12 +5,9 @@
 class uart_env_cov extends cip_base_env_cov #(.CFG_T(uart_env_cfg));
   `uvm_component_utils(uart_env_cov)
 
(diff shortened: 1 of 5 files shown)


marnovandermaas (approved):
Thank you for making this PR. Very nice to see these errors fixed instead of waived.

PR #26003, opened 2025-01-24
[tlul,rtl] Drop unused typedef from tlul_adapter_sram

rswarbrick:
This seemed sensible when I originally ported the code to use the shared fifo, but it turns out not to be needed. And Pirmin (quite reasonably) pointed out that the name is rather confusing. Forget about it!

(I mis-clicked when merging #25951. This PR tidies up again...)


nasahlpa (approved):

nasahlpa:

This PR merges only removes an unused typedef.

rswarbrick:

No functional change (this just removes an unused typedef)

PR #26006, opened 2025-01-24
[membkdr] Provide a tile adapter to access internal rows

Razer6:
This PR adds a custom row adapter that is being used by the memory backdoor util to access and manipulate rows of data. Integrators might are using SRAM primitives with a different internal organization than `Width x Depth`. They can provide a custom row adapter to the memory backdoor utility to for the needs of their memory prims.

### DIFF
--- a/hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
+++ b/hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
@@ -30,7 +30,7 @@ class mem_bkdr_util extends uvm_object;
   // The depth of a single SRAM tile.
   protected uint32_t tile_depth;
 
-  // The width of the memory.
+  // The logical width of the memory in bits, not including extra bits added by the row adapter.
   protected uint32_t width;
 
   // The number of subword entries in the whole memory
@@ -38,6 +38,10 @@ class mem_bkdr_util extends uvm_object;
 
   // Indicates the error detection scheme implemented for this memory.
   protected err_detection_e err_detection_scheme = ErrDetectionNone;
+
+  // Adapter to access the underlying memory organization
+  // Integrators can provide a custom row adapter for their SRAM primitives
+  protected mem_bkdr_util_row_adapter row_adapter;
 
   // Convenience macro to check if ECC / parity is enabled.
   `define HAS_ECC (!(err_detection_scheme inside {ErrDetectionNone, ParityEven, ParityOdd}))
@@ -88,6 +92,9 @@ class mem_bkdr_util extends uvm_object;
   //
   // Optional arguments:
   //
+  //  row_adapter              Adapter to access the internal row of a memory. Integrators can
+  //                           provide a custom adapter for a different memory architecture.
+  //
   //  num_prince_rounds_half   The number of rounds of PRINCE used to scramble the memory. This is
   //                           used for scrambled memories. This defaults to 3.
   //
@@ -113,6 +120,7 @@ class mem_bkdr_util extends uvm_object;
   //                           memory.
   function new(string name = "", string path, int unsigned depth,
                longint unsigned n_bits, err_detection_e err_detection_scheme,
+               mem_bkdr_util_row_adapter row_adapter = null,
                uint32_t num_prince_rounds_half = 3,
                uint32_t extra_bits_per_subword = 0, uint32_t system_base_addr = 0,
                string tiling_path = "", string tiling_suffix_fmt_str = ".gen_ram_inst[%0d].%s",
@@ -120,12 +128,18 @@ class mem_bkdr_util extends uvm_object;
     super.new(name);
     `DV_CHECK_FATAL(!(n_bits % depth), "n_bits must be divisible by depth.")
 
+    if (row_adapter != null) begin
+      this.row_adapter = row_adapter;
+    end else begin
+      this.row_adapter = new();
+    end
+
     this.path                   = path;
     this.tiling_path            = tiling_path;
     this.depth                  = depth;
     this.tile_depth             = tile_depth;
     this.tiling_suffix_fmt_str  = tiling_suffix_fmt_str;
-    this.width                  = n_bits / depth;
+    this.width                  = (n_bits / depth) - this.row_adapter.get_num_extra_bits();
     this.err_detection_scheme   = err_detection_scheme;
     this.num_prince_rounds_half = num_prince_rounds_half;
 
@@ -296,12 +310,13 @@ class mem_bkdr_util extends uvm_object;
   virtual function uvm_hdl_data_t read(bit [bus_params_pkg::BUS_AW-1:0] addr);
     bit res;
     uint32_t index, ram_tile;
-    uvm_hdl_data_t data;
+    uvm_hdl_data_t encoded_row, data;
     if (!check_addr_valid(addr)) return 'x;
     index    = addr >> addr_lsb;
     ram_tile = index / tile_depth;
-    res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
+    res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), encoded_row);
     `DV_CHECK_EQ(res, 1, $sformatf("uvm_hdl_read failed at index %0d", index))
+    data     = row_adapter.decode_row(encoded_row);
     return data;
   endfunction
 
@@ -410,11 +425,14 @@ class mem_bkdr_util extends uvm_object;
   // Updates the entire width of the memory at the given address, including the ECC bits.
   virtual function void write(bit [bus_params_pkg::BUS_AW-1:0] addr, uvm_hdl_data_t data);
     bit res;
+    uvm_hdl_data_t encoded_row;
     uint32_t index, ram_tile;
     if (!check_addr_valid(addr)) return;
-    index    = addr >> addr_lsb;
-    ram_tile = index / tile_depth;
-    res      = uvm_hdl_deposit($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
+    index       = addr >> addr_lsb;
+    ram_tile    = index / tile_depth;
+    encoded_row = row_adapter.encode_row(data);
+    res         = uvm_hdl_deposit($sformatf("%0s[%0d]", get_full_path(ram_tile), index),
+                                  encoded_row);
     `DV_CHECK_EQ(res, 1, $sformatf("uvm_hdl_deposit failed at index %0d", index))
   endfunction
 
@@ -466,9 +484,14 @@ class mem_bkdr_util extends uvm_object;
 
   // this is used to write 32bit of data plus 7 raw integrity bits.
   virtual function void write39integ(bit [bus_params_pkg::BUS_AW-1:0] addr, logic [38:0] data);
+    uvm_hdl_data_t rw_data;
     `_ACCESS_CHECKS(addr, 32) // this is essentially an aligned 32bit access.
     if (!check_addr_valid(addr)) return;
-    write(addr, data);
+    // Perform a read-modify-write to access the underlying memory architecture
+    rw_data = read(addr);
+    rw_data = row_adapter.write_row_data_39b(addr, data, rw_data);
+    // Note the write function takes care of interleaving, if used.
+    write(addr, rw_data);
   endfunction
 
   virtual function void write64(bit [bus_params_pkg::BUS_AW-1:0] addr, logic [63:0] data);
--- a/hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
+++ b/hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
@@ -1,0 +1,57 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Provide an abstract way to access the memory rows
+//
+// Integrators can subclass this to provide a specfic way of accessing a memory row
+// and incorporate the phyical architecture.
+//
+class mem_bkdr_util_row_adapter;
+
+  // A row might have additional extra bits
+  protected uint32_t num_extra_bits = 0;
+
+  // Return the number of extra bits of this row architecture
+  function uint32_t get_num_extra_bits();
+    return num_extra_bits;
+  endfunction
+
+  // Translates a raw encoded UVM data row from the memory in a contiguous
+  // row of memory.
+  //
+  virtual function uvm_hdl_data_t decode_row(uvm_hdl_data_t read_data);
+    return read_data;
+  endfunction
+
+  // Translates a contiguous UVM data row to the internal organization of a row
+  // that can be written to the memory.
+  //
+  virtual function uvm_hdl_data_t encode_row(uvm_hdl_data_t write_data);
+    return write_data;
+  endfunction
+
+  // Writes a 39 bit word into a decoded row data depending on the memory architecture
+
+  // Given decoded `row_data`, a 39-bit `data` word to be written, and an address, return the
+  // decoded row data with the data word at the correct position for the memory architecture and
+  // the given address.
+  virtual function uvm_hdl_data_t write_row_data_39b(bit [bus_params_pkg::BUS_AW-1:0] addr,
+                                                     logic [38:0] data,
+                                                     uvm_hdl_data_t row_data);
+    row_data[38:0] = data;
+    return row_data;
+  endfunction
+
+  // Reads a 39 bit word from decoded row data depending on the memory architecture
+
+  // Given decoded `row_data` and an address, return the 39-bit data from the correct position
+  // for the memory architecture and the given address.
+  virtual function logic [38:0] read_row_data_39b(bit [bus_params_pkg::BUS_AW-1:0] addr,
+                                                  uvm_hdl_data_t row_data);
+    logic data;
+    data = row_data[38:0];
+    return data;
+  endfunction
+
+endclass
--- a/hw/dv/sv/mem_bkdr_util/mem_bkdr_util.core
+++ b/hw/dv/sv/mem_bkdr_util/mem_bkdr_util.core
@@ -18,6 +18,7 @@ filesets:
       - lowrisc:ip:kmac_pkg
     files:
       - mem_bkdr_util_pkg.sv
+      - mem_bkdr_util_row_adapter.sv: {is_include_file: true}
       - mem_bkdr_util.sv: {is_include_file: true}
     file_type: systemVerilogSource
 
--- a/hw/dv/sv/mem_bkdr_util/mem_bkdr_util_pkg.sv
+++ b/hw/dv/sv/mem_bkdr_util/mem_bkdr_util_pkg.sv
@@ -40,5 +40,6 @@ package mem_bkdr_util_pkg;
   `include "dv_macros.svh"
 
   // sources
+  `include "mem_bkdr_util_row_adapter.sv"
   `include "mem_bkdr_util.sv"
 endpackage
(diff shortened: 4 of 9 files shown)


hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+`include "mem_bkdr_util_row_adapter.sv"
```
rswarbrick, line 5: Shouldn't this inclusion be in `mem_bkdr_util_pkg.sv`?

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -104,18 +113,20 @@ class mem_bkdr_util extends uvm_object;
...
 
-    this.path                  = path;
-    this.tiling_path           = tiling_path;
-    this.depth                 = depth;
-    this.tile_depth            = tile_depth;
-    this.width                 = n_bits / depth;
-    this.err_detection_scheme  = err_detection_scheme;
+    this.path                   = path;
+    this.tiling_path            = tiling_path;
+    this.row_adapter           = row_adapter;
```
rswarbrick, line 125: Nit: Indentation of '='

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -290,6 +301,7 @@ class mem_bkdr_util extends uvm_object;
     index    = addr >> addr_lsb;
     ram_tile = index / tile_depth;
     res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
+    data     = row_adapter.read_row(data);
```
rswarbrick, line 304: I'm slightly surprised about this: wouldn't it be nicer to put the `uvm_hdl_read` into the row adapter (and maybe change its name)?

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
```diff
@@ -0,0 +1,32 @@
...
+  // that can be written to the memory.
+  //
+  virtual function uvm_hdl_data_t write_row(uvm_hdl_data_t write_data);
+    return write_data;
+  endfunction
+
+  // Writes a 39 bit word into the row data depending on the memory architecture
+  virtual function uvm_hdl_data_t access_row_data39(bit [bus_params_pkg::BUS_AW-1:0] addr,
+                                                    logic [38:0] data,
+                                                    uvm_hdl_data_t row_data);
```
rswarbrick, line 28: This seems surprising to me. Why is `row_data` an argument at all?

hw/ip/otbn/dv/uvm/env/otbn_env_pkg.sv
```diff
@@ -15,7 +15,7 @@ package otbn_env_pkg;
   import push_pull_agent_pkg::*;
   import otbn_model_agent_pkg::*;
   import otbn_memutil_pkg::*;
-  import mem_bkdr_util_pkg::mem_bkdr_util;
+  import mem_bkdr_util_pkg::mem_bkdr_util_row_adapter;
```
rswarbrick, line 18: Is that really correct? I think that e.g. `otbn_env.sv` uses `mem_bkdr_util`.

hw/ip/otbn/dv/uvm/tb.sv
```diff
@@ -344,20 +345,24 @@ module tb;
     // Instantiate mem_bkdr_util objects to allow access to IMEM and DMEM
     //
     // Note that n_bits is the number of bits in the memory, including ECC check bits.
+    mem_bkdr_util_row_adapter row_adapter;
```
rswarbrick, line 348: I think it might be tidier to declare this next to `imem_util` and `dmem_util`?

hw/ip/rom_ctrl/dv/tb.sv
```diff
@@ -78,12 +79,14 @@ module tb;
...
   initial begin
+    mem_bkdr_util_row_adapter row_adapter = new();
     mem_bkdr_util m_mem_bkdr_util;
     m_mem_bkdr_util = new(.name  ("mem_bkdr_util"),
                           .path  (`DV_STRINGIFY(`ROM_CTRL_MEM_HIER)),
                           .depth ($size(`ROM_CTRL_MEM_HIER)),
                           .n_bits($bits(`ROM_CTRL_MEM_HIER)),
-                          .err_detection_scheme(mem_bkdr_util_pkg::EccInv_39_32));
+                          .err_detection_scheme(mem_bkdr_util_pkg::EccInv_39_32),
+                          .row_adapter(row_adapter));
```
rswarbrick, line 89: I wonder whether it makes sense to allow `row_adapter` to be null as a default argument to the constructor. That would avoid needing lots of these changes.

Razer6: That sounds good!

hw/ip/otbn/dv/uvm/tb.sv
Razer6: dropped that, by using a default adapter.

hw/ip/otbn/dv/uvm/env/otbn_env_pkg.sv
Razer6: Probably not.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
Razer6: I wanted to make the change as small as possible. `.read_row()` translates the internal encoded row into a contiguous chunk of data containing all subwords.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
Razer6: This function takes a contiguous data row and inserts the data word into it at the right subword index. The later `write()` command then translates the contiguous row first to the memory internal format, via `write_row()`.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
Razer6: Probably? I was not sure about that in general to make the type visible.

Razer6:
At the moment I am not sure why I get a Null error here

Razer6:
@rswarbrick I am not sure how to fix that. Can you help me on that?
The constructor has the following statement:

```systemverilog
   if (row_adapter != null) begin
      this.row_adapter = row_adapter;
    end else begin
      this.row_adapter = new();
    end
 ```
 
 And later on, the row adapter is used like:
 ```systemverilog
this.width                  = (n_bits / depth) - this.row_adapter.get_num_extra_bits();
```

However, this causes the following error (line 132 is  the line from above when calling `get_num_extra_bits()`) :

```
Error-[NOA] Null object access
                ../src/lowrisc_dv_mem_bkdr_util_0/mem_bkdr_util.sv, 132
                  The object at dereference depth 1 is being used before it was
                  constructed/allocated.
                  Please make sure that the object is allocated before using it.
 ```

**Comment by rswarbrick** [comment]
Hmm, that seems really odd to me: I'm not sure what's going on. Can I suggest debugging by adding some print statements inside the `if/else` block (possibly using `%p` to show the value of `this.row_adapter` at the end of it?

**Comment by Razer6** [comment]
> Hmm, that seems really odd to me: I'm not sure what's going on. Can I suggest debugging by adding some print statements inside the `if/else` block (possibly using `%p` to show the value of `this.row_adapter` at the end of it?

Stupid me! After the if/else, the default assignment from the parameter to the member variable was still there. This overwrote the previously allocated object with null again.  @rswarbrick  Fixed now, please take a look.

**Comment by rswarbrick** [comment]
Ha! I didn't notice it either :-) I'll have a proper look now.

File: hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv (hunk 3 above)
Comment (rswarbrick) [reply hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv:307]: That makes sense. I guess I was just surprised by an architecture different from what I imagined. Looks perfectly reasonable to me though.

File: hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv [hunk 8]
```diff
@@ -403,6 +418,7 @@ class mem_bkdr_util extends uvm_object;
     if (!check_addr_valid(addr)) return;
     index    = addr >> addr_lsb;
     ram_tile = index / tile_depth;
+    data     = row_adapter.write_row(data);
     res      = uvm_hdl_deposit($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
```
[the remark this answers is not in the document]
Razer6: I'll add another variable, that makes some debugging (if needed) a bit easier to make prints eetc..

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -290,6 +304,7 @@ class mem_bkdr_util extends uvm_object;
     index    = addr >> addr_lsb;
     ram_tile = index / tile_depth;
     res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
+    data     = row_adapter.read_row(data);
     `DV_CHECK_EQ(res, 1, $sformatf("uvm_hdl_read failed at index %0d", index))
     return data;
```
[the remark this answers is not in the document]
Razer6: Added encoded row

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
```diff
@@ -0,0 +1,41 @@
...
```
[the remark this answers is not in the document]
rswarbrick: Following up (after @Razer6 had a helpful phone call with me!): No, what I wrote there is nonsense. Please ignore :-D

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -285,11 +299,12 @@ class mem_bkdr_util extends uvm_object;
...
     bit res;
     uint32_t index, ram_tile;
-    uvm_hdl_data_t data;
+    uvm_hdl_data_t encoded_row, data;
     if (!check_addr_valid(addr)) return 'x;
     index    = addr >> addr_lsb;
     ram_tile = index / tile_depth;
-    res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
+    res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), encoded_row);
+    data     = row_adapter.decode_row(encoded_row);
```
rswarbrick, line 307: Now I've understood a bit more, I'm not sure this is quite right. I think this is returning several words of data packed together? But the `read` function is supposed to return just one word, isn't it?

Don't we need to extract just one lane here?

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -399,11 +414,14 @@ class mem_bkdr_util extends uvm_object;
...
     bit res;
+    uvm_hdl_data_t encoded_row;
     uint32_t index, ram_tile;
     if (!check_addr_valid(addr)) return;
-    index    = addr >> addr_lsb;
-    ram_tile = index / tile_depth;
-    res      = uvm_hdl_deposit($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
+    index       = addr >> addr_lsb;
+    ram_tile    = index / tile_depth;
+    encoded_row = row_adapter.encode_row(data);
```
rswarbrick, line 422: I think this has the same problem as the `read` function above. If we've got multiple words in the row, won't this need to do a read-modify-write?

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
```diff
@@ -0,0 +1,41 @@
...
+  endfunction
+
+  // Translates a raw read UVM data raw from the memory in a contigious
+  // row of memory.
+  //
+  virtual function uvm_hdl_data_t decode_row(uvm_hdl_data_t read_data);
+    return read_data;
+  endfunction
+
+  // Translates a contigious UVM data row to the internal organzation of a row
```
rswarbrick, line 27: Nit: Spelling of "organisation" (or possibly "organization")

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -26,7 +26,7 @@ class mem_bkdr_util extends uvm_object;
   // The depth of a single SRAM tile.
   protected uint32_t tile_depth;
 
-  // The width of the memory.
+  // The logical width of the memory in bits, not including extra bits added by the row adapter.
```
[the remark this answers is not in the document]
Razer6: Yes, `n_bits` includes all possible bits of the memory.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
```diff
@@ -0,0 +1,41 @@
...
+
+  // A row might have additional extra bits
+  protected uint32_t num_extra_bits = 0;
+
+  // Return the number of extra bits of this row architecture
+  function uint32_t get_num_extra_bits();
+    return num_extra_bits;
+  endfunction
+
+  // Translates a raw encoded UVM data row from the memory in a contigious
```
andreaskurth, line 20:
```suggestion
  // Translates a raw encoded UVM data row from the memory in a contiguous
```

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
```diff
@@ -0,0 +1,41 @@
...
+  // row of memory.
+  //
+  virtual function uvm_hdl_data_t decode_row(uvm_hdl_data_t read_data);
+    return read_data;
+  endfunction
+
+  // Translates a contigious UVM data row to the internal organization of a row
```
andreaskurth, line 27:
```suggestion
  // Translates a contiguous UVM data row to the internal organization of a row
```

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
andreaskurth: Based on how the functions and variables are currently named, I agree with @rswarbrick.  This code seems to decode a row and assign that to data.  @Razer6: Doesn't this function need to select a word from the decoded row based on `addr`?

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
```diff
@@ -0,0 +1,41 @@
...
+  // that can be written to the memory.
+  //
+  virtual function uvm_hdl_data_t encode_row(uvm_hdl_data_t write_data);
+    return write_data;
+  endfunction
+
+  // Writes a 39 bit word into a decoded row data depending on the memory architecture
```
andreaskurth, line 34:
```suggestion
  // Given decoded `row_data`, a 39-bit `data` word to be written, and an `addr`ess, return the decoded row data with the data word at the correct position for the memory architecture and the given address.
```

andreaskurth: Is my suggestion correct? It may be a bit long, but I think this needs a more precise description than is currently in the PR.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
Razer6: I think the later function `read8` (which calls `read`) needs a proper access function to select the right word?

andreaskurth: Ah, I see. I think what confused me is the short description of this function ("Read the entire word at the given address."). I suggest we change _word_ to _memory row_, because it may contain more than one word with parity bits, depending on the memory architecture. Does that make sense?

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
```diff
@@ -0,0 +1,45 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Provide an abstract way to access the memory rows
+//
+// Integrators can subclass this to privide a specfic way of accessing a memory row
```
matutem, line 7:
```suggestion
// Integrators can subclass this to provide a specfic way of accessing a memory row
```

matutem:
The interface provided by the row adapter is okay for a trivial adapter, but a more interesting one may need the address. As it stands it seems to only address memories with extra bits per word, and these bits are assumed to be in the most significant position. Is this all you intend to address? The description seems to suggest the ram may be very wide and fit multiple words. Assuming ECC is still per word then you would need to mux the data out for reads, and perhaps do read-modify-write for writes?

Perhaps you actually intend to implement redundant columns, as for addressing yield issues. In that case we would basically swap some portion of the word according to some setting derived from the errors detected by the ATE. Is this what you are trying to support? This would be simpler than the very large word organization mentioned above and seems to fit the interface you provide.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -285,11 +299,12 @@ class mem_bkdr_util extends uvm_object;
   virtual function uvm_hdl_data_t read(bit [bus_params_pkg::BUS_AW-1:0] addr);
     bit res;
     uint32_t index, ram_tile;
-    uvm_hdl_data_t data;
+    uvm_hdl_data_t encoded_row, data;
     if (!check_addr_valid(addr)) return 'x;
     index    = addr >> addr_lsb;
     ram_tile = index / tile_depth;
-    res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), data);
+    res      = uvm_hdl_read($sformatf("%0s[%0d]", get_full_path(ram_tile), index), encoded_row);
```
matutem, line 306: res is 0 if there was an error, so I suggest you wrap uvm_hdl_read in `DV_CHECK or some such to catch any errors.

matutem:
IMHO, this PR needs a better explanation of what it is really trying to accomplish or it will continue to linger. We need clarity if the "row" is meant to contain multiple words (thus needing RMW, or masking, or muxing on reads).

Also the whole mem backdoor utility is a bit wrong, and should ONLY perform dumb reads and writes of whatever width the mem_bkdr_util instance wires. So, for example, the read39 and write39 should NOT be part of the interface, since they only apply to very specific RAMs. The higher level objects (sram, otp, rom, flash, etc.) should perform all scrambling, ecc processing, encryption, whatever, and at the bottom perform ONLY read and write operations via the mem_bkdr_util class instances they create. Once this split is clean we should have read and write handle macro specific operations, like dealing with extra bits due to the chosen macro, column redundancy, etc.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util.sv
```diff
@@ -88,6 +92,9 @@ class mem_bkdr_util extends uvm_object;
   //
   // Optional arguments:
   //
+  //   row_adapter             Adapter to access the internal row of a memory. Integrators can
```
rswarbrick: Definitely not important, but if you force-push for another reason, I think there's a stray space before "row_adapter"

Razer6: CI is green before fixing the whitespace. Let's merge it.
Screenshot 2025-07-08 at 13 58 20

PR #26008, opened 2025-01-24
[racl] Flop TLUL to be in sync with Reg adapter

Razer6:
The reg adapter might introduce a flop stage. We need to be in sync with that when performing a RACL check. If the flop stage is enabled, also add a flop stage for the RACL related TLUL signals.


Razer6:
Closing in favour of https://github.com/lowRISC/opentitan/pull/26015

PR #26009, opened 2025-01-24
Gpio straps func cov

marcelocarvalhoLowRisc:
Functional coverage properties included in the gpio_straps_if.sv and some other minor fixes.


hw/ip/gpio/dv/env/gpio_env_cov.sv
```diff
@@ -50,6 +50,49 @@ class gpio_out_oe_cov_obj extends uvm_object;
...
 
+class gpio_straps_cov_obj extends uvm_object;
+  `uvm_object_utils(gpio_straps_cov_obj)
+
+  covergroup strap_snapshot_cg (string name) with function sample(bit strap_en, bit rst_n, uint data_in);
+    option.per_instance = 1;
+
+    // Coverpoint for strap_en trigger
+    cp_strap_en: coverpoint strap_en {
+      bins triggered = {1};
```
rswarbrick, line 61: I think we probably want to check the feature works, no matter what happens after the cycle the strap gets enabled.

One approach would be to sample something from a "slightly directed" bit of test that reads back the sampled value and cares about whether the pulse that caused the sample to happen is still high?

hw/ip/gpio/dv/env/gpio_env_cov.sv
```diff
@@ -50,6 +50,49 @@ class gpio_out_oe_cov_obj extends uvm_object;
...
+    }
+    // Coverpoint for the rst_n signal
+    cp_reset: coverpoint {rst_n, strap_en} {
+      // Bins to cover rst_n states
+      bins during_reset = {0, 0};          // rst_n is asserted, strap_en doesn't trigger a snapshot
+      bins after_reset_strap_en = {1, 1};  // rst_n is deasserted, and strap_en triggers a snapshot
+      bins illegal_during_reset = {0, 1};  // Illegal state: strap_en should NOT trigger during rst_n
```
rswarbrick, line 68: I'm not sure why this state would be illegal. Isn't `strap_en` an input signal for the block?

hw/ip/gpio/dv/env/gpio_env_cov.sv
```diff
@@ -50,6 +50,49 @@ class gpio_out_oe_cov_obj extends uvm_object;
...
+      bins all_values[] = {0, 32'hFFFFFFFF, default};
+    }
+    //cp_data_in_timing: coverpoint (strap_en ##1 data_in) {
+    //  bins strap_en_to_data_in[] = (0 => 1);
+    //  bins data_in_too_early[] = (1 => 1);
+    //}
+    //cp_reset_timing: coverpoint rst_n ##1 strap_en {
+    //  bins reset_to_strap_en = {0 => 1};  // Strap_en asserted 1 cycle after rst_n deassertion
+    //  bins strap_en_too_early = {1 => 1}; // Strap_en asserted while rst_n is still active
+    //}
```
rswarbrick, line 81: Commented code?

hw/ip/gpio/dv/env/gpio_env_cov.sv
```diff
@@ -50,6 +50,49 @@ class gpio_out_oe_cov_obj extends uvm_object;
...
+
+
+  // Cross coverage: strap_en with data_in
+  cross cp_strap_en, cp_data_in;
+  cross cp_reset, cp_data_in;
+
+  // Function: new
+  function new(string name="gpio_straps_cov_obj");
+    super.new(name);
+    strap_snapshot_cg = new(name);
```
rswarbrick, line 92: I'm not sure that's right. Usually, we don't really care about the `name` argument, but it should be a name for the object (which is being stored in `strap_snapshot_cg`), not the surrounding class.

hw/ip/gpio/dv/env/gpio_env_cov.sv
```diff
@@ -50,6 +50,49 @@ class gpio_out_oe_cov_obj extends uvm_object;
...
+    //  bins strap_en_too_early = {1 => 1}; // Strap_en asserted while rst_n is still active
+    //}
+  endgroup
+
+
+  // Cross coverage: strap_en with data_in
+  cross cp_strap_en, cp_data_in;
+  cross cp_reset, cp_data_in;
+
+  // Function: new
```
rswarbrick, line 89: Please drop this line. Probably also worth it to open a separate PR that drops analogous lines across `hw/ip/gpio/dv`.

rswarbrick (commented):
There are a few specific items here, but I think the commit structure also needs sorting out. You can probably do that by rebasing on top of any other PRs that are needed and then squashing together the commits that are unique to this PR.

hw/ip/gpio/dv/env/gpio_env_cov.sv
marcelocarvalhoLowRisc: Yes, I did this cover point but actually not very useful. My plan is to remove this and I did something better using sequences/properties in the strap interface.

hw/ip/gpio/dv/env/gpio_env_cov.sv
marcelocarvalhoLowRisc: Sorry, also not useful bins, I will remove this.

hw/ip/gpio/dv/env/gpio_env_cov.sv
marcelocarvalhoLowRisc: true, it was mistake from my side.

hw/ip/gpio/dv/env/gpio_env_cov.sv
marcelocarvalhoLowRisc: I will do a clean up.

hw/dv/tools/vcs/cover.cfg
```diff
@@ -9,6 +9,9 @@
 -module pins_if     // DV construct.
 -module clk_rst_if  // DV construct.
 
++tree tb
++module gpio_straps_if
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: fixed

rswarbrick (commented):
At the moment, I think this includes the changes from the PR that defines the rand straps vseq. I don't think there's any reason that they should be needed for a coverage change. Could you drop them from this PR?

marcelocarvalhoLowRisc:
> At the moment, I think this includes the changes from the PR that defines the rand straps vseq. I don't think there's any reason that they should be needed for a coverage change. Could you drop them from this PR?

Yes, it's fixed. I droped most of the items, I only kept the strap interface, where it is located the coverage properties.

hw/ip/gpio/dv/gpio_sim_cfg.hjson
```diff
@@ -180,6 +187,10 @@
       build_mode: en_cdc_prims
       run_opts: ["+no_pullup_pulldown=1"]
     }
+    {
+      name: gpio_rand_straps
+      uvm_test_seq: gpio_rand_straps_vseq
+    }
```
rswarbrick, line 193: I don't think this should be here?

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,63 @@
...
+
+// Interface definition
+interface gpio_straps_if();
+
+    logic          clk;
+    logic          rst_n;
+    logic          strap_en;       // Signal to enable straps
+    gpio_straps_t  sampled_straps; // Sampled gpio_i snapshot data from GPIO (DUT)
+
+    modport straps_port_in(input  clk,
```
rswarbrick, line 18: I find this sort of thing really hard to understand (not just from you!). Could you add a comment above it that explains who we're expecting to interact through this modport? ("This is the dut's view of the sampled straps" or "This is the way the outside world passes in straps" or ...)

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,63 @@
...
+                           input  rst_n,
+                           input  strap_en,
+                           output sampled_straps);
+
+    modport straps_port_out(output  clk,
+                            output  rst_n,
+                            output  strap_en,
+                            input sampled_straps);
+
+    // Sequence to capture the timing relationship between strap_en and strap_data
```
rswarbrick, line 28: I'm not sure I understand this. Normally a sequence models a 1-bit value changing over time (as used by a property or assertion). I don't think that `sampled_straps.data` is this.

marcelocarvalhoLowRisc: I have replaced the cover properties with macros and I wrote some comments.

hw/ip/gpio/dv/gpio_sim_cfg.hjson
marcelocarvalhoLowRisc: removed.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
```diff
@@ -0,0 +1,71 @@
...
+      end
+    end
+
+    // Sequence to capture the timing relationship between strap_en and strap_data
+    // This is used to cover that the straps are sampled after the strap_en signal is asserted
+    // The toggle coverage will make sure about all the possible transitions of the strap_data signal
+    `COVER(COV_STRAP_DATA_TIMING, strap_en ##1 sampled_straps.data)
+    `COVER(COV_STRAP_VALID_TIMING, strap_en ##1 sampled_straps.valid)
+
+    //sequence strap_data_timing_seq;
```
rswarbrick: Something looks rather strange here: the first commit adds some commented code and the second commit removes it again.

I'd suggest:
- Squash the two commits together locally.
- Then rebase on top of the tip of any PR that you depend on.
The resulting history (once any merge conflicts get sorted out) should have the stuff that needs reviewing in this PR as just the last commit.

hw/ip/gpio/dv/interfaces/gpio_straps_if.sv
rswarbrick: Thanks for tweaking the name: this makes much more sense! Would you mind changing the suffix too? I think you want something like `dut_port` instead of `dut_if` (since this is part of the interface, not an interface itself)

rswarbrick:
I'm looking through the commits and I think the first one is from another PR (Is that correct?) For the second one, the commit message is talking about PR structure, which doesn't really belong outside of github. That would probably be clearer in the conversation here.

hw/ip/gpio/dv/gpio_sim_cfg.hjson
```diff
@@ -47,8 +47,15 @@
...
+  overrides: [
+    {
+      name: default_vcs_cov_cfg_file
+      value: "-cm_hier {dv_root}/tools/vcs/cover.cfg+{proj_root}/hw/ip/gpio/dv/cov/gpio_cover.cfg"
+    }
+  ]
+
   // Enable cdc instrumentation.
-  run_opts: ["+cdc_instrumentation_enabled=1"]
+  run_opts: ["+cdc_instrumentation_enabled=1", "+interface", "+vcs+dumpinterface", "+define+UVM_DUMP_INTERFACES", "+cover=all"]
```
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: I thought that this line 58 changes was required for the gpio coverage visualisation. Is it not required? together with the lines 50/56?

marcelocarvalhoLowRisc:
> I'm looking through the commits and I think the first one is from another PR (Is that correct?) For the second one, the commit message is talking about PR structure, which doesn't really belong outside of github. That would probably be clearer in the conversation here.

Yes, it is from another PR. I have removed the comments about the PR structure. I agreed, that will probably be merged anyway, so won't need those confusing comments about what I included or not in the PR.

marcelocarvalhoLowRisc (commented):
I removed the COV properties and created these assertions instead as we discussed.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
```diff
@@ -0,0 +1,30 @@
...
+  input logic                 rst_ni,
+  input logic                 strap_en_i,
+  input logic                 strap_valid,
+  input logic [NUM_GPIOS-1:0] strap_data,
+  input logic [NUM_GPIOS-1:0] gpio_i
+);
+
+  // Check if when a strap_en_i is high,
+  // one clock cycle after it must have a valid signal high.
+  `ASSERT(strap_valid_assert,
```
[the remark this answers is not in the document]
rswarbrick: The stable assertions you've added don't last very long! My proposal was a property that started on every cycle and lasted a cycle. I don't think the code that you've added stops something like "strap_en_i goes high; strap_valid goes high; (wait ten cycles); strap_valid drops"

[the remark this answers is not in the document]
marcelocarvalhoLowRisc: The strap_valid does not drops if the strap_en_i drops. The only condition that drops the strap_valid is the reset.
So it should be stable until the next reset comes. Or the RTL should be something different?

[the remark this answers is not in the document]
marcelocarvalhoLowRisc: True, I didn't saw this case, I will fix it.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
```diff
@@ -0,0 +1,28 @@
...
+  strap_en_i |=> strap_valid, clk_i, !rst_ni);
+  // Check if strap output data is mirrored from gpio_i input data,
+  // one clock cycle after the strap_en is high.
+  `ASSERT(strap_data_assert,
+  strap_en_i && !strap_valid |=> (strap_data == $past(gpio_i)),
+  clk_i, !rst_ni);
+  // Check if the strap_valid never fell, only after reset.
+  `ASSERT(strap_valid_stable_assert, !$fell(strap_valid), clk_i, !rst_ni);
```
rswarbrick, line 27: As I said [here](https://github.com/lowRISC/opentitan/pull/26009/commits/1bebc63daf841eb6f45ec522ddb2af87b29767ad#r1973527271), this doesn't need the last two arguments.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
```diff
@@ -0,0 +1,28 @@
...
+);
+
+  // Check if when a strap_en_i is high,
+  // one clock cycle after it must have a valid signal high.
+  `ASSERT(strap_valid_assert,
+  strap_en_i |=> strap_valid, clk_i, !rst_ni);
+  // Check if strap output data is mirrored from gpio_i input data,
+  // one clock cycle after the strap_en is high.
+  `ASSERT(strap_data_assert,
+  strap_en_i && !strap_valid |=> (strap_data == $past(gpio_i)),
```
rswarbrick, line 24: Probably easier: can't the assertion be something like `$rose(strap_valid) -> (strap_data == $past(gpio_i))`?

hw/ip/gpio/dv/sva/gpio_strap_check.sv
```diff
@@ -0,0 +1,28 @@
...
+`include "prim_assert.sv"
+import gpio_env_pkg::NUM_GPIOS;
+
+module gpio_strap_check (
+  input logic                 clk_i,
+  input logic                 rst_ni,
+  input logic                 strap_en_i,
+  input logic                 strap_valid,
+  input logic [NUM_GPIOS-1:0] strap_data,
+  input logic [NUM_GPIOS-1:0] gpio_i
```
rswarbrick, line 14: Thinking about this, we might be able to avoid depending on the env package. Can we add a NUM_GPIOS parameter to this module and then pass it in the bind statement in `gpio_bind.sv`? There, I think we can use something like `$bits(cio_gpio_i)`, which will pick up the size in the binding site.

rswarbrick (commented):
A couple of nitty comments. Please could you fetch and rebase on the head of master? The result should only have one commit now.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
marcelocarvalhoLowRisc: No, then you gonna miss the check related with the strap_en_i. The strap_data and data_valid should be the consequence behaviour of triggering the strap_en_i.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
marcelocarvalhoLowRisc: cool, thanks! I forgot that has the default same names for clk and rst.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
rswarbrick: Hmm, I'm not sure that's correct. I guess I didn't assert that `strap_valid` only rises because of `strap_en_i`. Maybe we can strengthen things to `$rose(strap_valid) -> $past(strap_en_i) && (strap_data == $past(gpio_i))` ?

That way, we're saying "We only rise if we're enabled and, if so, we'll sample the right thing". The previous assertion says "We *do* rise if we are enabled".

hw/ip/gpio/dv/sva/gpio_bind.sv
```diff
@@ -20,4 +20,14 @@ module gpio_bind;
     .d2h    (tl_o)
   );
 
+  bind gpio gpio_strap_check
+    #(.NUM_GPIOS($bits(cio_gpio_i))) gpio_strap_assert (
```
rswarbrick, line 24: Nit: Oof, the indentation is a bit awkward here. Maybe it's better to put the newline one token later, so that `gpio_strap_check #(blah blah)` is together on a line.

rswarbrick (commented):
This looks pretty much good to go: thanks! Still slight questions about tidy-ups, but I'm pretty enthusiastic :-)

hw/ip/gpio/dv/sva/gpio_bind.sv
marcelocarvalhoLowRisc: I agreed, I will fix it.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
marcelocarvalhoLowRisc: But then, what happens if has a bug and the the strap_valid never rise? strap_valid is a consequence, not a pre-condition. We cannot rely on strap_valid because has no guarantee that it will rises. The only precondition we can rely on is the strap_en that is driven by the user.
Actually, should be something like this: $(strap_en_i) |=> (strap_data == $past(gpio_i)
Because again, has no guarantee that the strap_valid behaviour is correct. So if has a bug in the strap_valid, we will never trigger the assertion to check the data.

marcelocarvalhoLowRisc (commented):
Fixed the assertions, based on what we talked.
Just a minor change instead of $past(strap_valid) I did $past(strap_valid, 2) because $rose(strap_valid) and $past(strap_valid) collides in the same clock cycle. And to guarantee that strap_valid is never rises in any other circumstances, I have created another assertion, named strap_valid_no_other_rise_assert.

marcelocarvalhoLowRisc (commented):
Ready to be review again, after the ip_templates changes.

hw/ip_templates/gpio/dv/sva/gpio_bind.sv.tpl
```diff
@@ -20,4 +20,14 @@ module ${module_instance_name}_bind;
     .d2h    (tl_o)
   );
 
+  bind ${module_instance_name} ${module_instance_name}_strap_check #(.NUM_GPIOS($bits(cio_${module_instance_name}_i)))
```
rswarbrick, line 23: I'd be inclined to get rid of all the `${module_instance_name}` instances and turn them into just `gpio`. I think we then get rid of the need for templating, so this file can drop its `.tpl` suffix.

I'd suggest you add a commit before this one that makes that change in the existing version of this file.

hw/ip/gpio/dv/sva/gpio_strap_check.sv
rswarbrick: The point is that we check that in `strap_valid_assert`. You might have worked through this (and I see you've tweaked the assertion: thanks).

Unfortunately, it's not quite right. At the moment, you have `$rose(strap_valid) |=> (strap_data == $past(gpio_i))`. I think you want `->` instead (so that we do the check on the correct cycle if `gpio_i` is changing).

hw/ip_templates/gpio/dv/sva/gpio_strap_check.sv.tpl
```diff
@@ -0,0 +1,36 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+`include "prim_assert.sv"
+
+module ${module_instance_name}_strap_check #(parameter int NUM_GPIOS = 32)
+  (
```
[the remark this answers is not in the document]
rswarbrick: Thanks. Could you also drop the leading space from line 15 (before the `)`)?

marcelocarvalhoLowRisc (commented):
I think now everything is fixed

hw/ip/gpio/dv/sva/gpio_strap_check.sv
marcelocarvalhoLowRisc: Yes, that is true, I forgot that strap_valid is high on the same cycle that we have the data to be checked, instead of strap_en that was before...I will fix that.

hw/ip_templates/gpio/dv/sva/gpio_strap_check.sv
```diff
@@ -0,0 +1,30 @@
...
+  input logic [NUM_GPIOS-1:0] gpio_i
+  );
+
+  // Ensure `strap_valid` is high one clock cycle after `strap_en_i` is high.
+  `ASSERT(strap_valid_assert, strap_en_i |=> strap_valid);
+  // Check if strap output data is mirrored from gpio_i input data,
+  // one clock cycle after the strap_en is high.
+  `ASSERT(strap_data_assert, $rose(strap_valid) |-> (strap_data == $past(gpio_i)));
+  // Check if the strap_valid never fell, only after reset.
+  `ASSERT(strap_valid_stable_assert, !$fell(strap_valid));
```
rswarbrick, line 23: I'm really sorry to have got this wrong, but @KinzaQamar  demonstrated I was wrong about this on Wednesday (looking at some rv_plic code).

The problem is that this might fail on the first cycle, because the `x -> 0` transition satisfies "$fell". We can make sure that we're not looking at the first cycle after reset with a "1##" prefix:
```systemverilog
  `ASSERT(StrapValidOneWay_A, ##1 !$fell(strap_valid))
```

hw/ip_templates/gpio/dv/sva/gpio_strap_check.sv
```diff
@@ -0,0 +1,30 @@
...
+  input logic                 strap_valid,
+  input logic [NUM_GPIOS-1:0] strap_data,
+  input logic [NUM_GPIOS-1:0] gpio_i
+  );
+
+  // Ensure `strap_valid` is high one clock cycle after `strap_en_i` is high.
+  `ASSERT(strap_valid_assert, strap_en_i |=> strap_valid);
+  // Check if strap output data is mirrored from gpio_i input data,
+  // one clock cycle after the strap_en is high.
+  `ASSERT(strap_data_assert, $rose(strap_valid) |-> (strap_data == $past(gpio_i)));
```
rswarbrick, line 21: This looks good, but I guess we should also assert that it was genuinely sampled and doesn't change. How about this?
```systemverilog
  `ASSERT(StrapSamples_A, $rose(strap_valid) -> (strap_data == $past(gpio_i)));
  `ASSERT(StrapDataStable_A, strap_valid && $stable(strap_valid) -> $stable(strap_data));
```

hw/ip_templates/gpio/dv/sva/gpio_strap_check.sv
```diff
@@ -0,0 +1,30 @@
...
+  // Check if the strap_valid never fell, only after reset.
+  `ASSERT(strap_valid_stable_assert, !$fell(strap_valid));
+  // Check if we have a strap valid signal high, then the strap en
+  // must be high one clock cycle before.
+  `ASSERT(strap_valid_low_assert, $rose(strap_valid) |=> $past(strap_en_i));
```
rswarbrick, line 26: This is the same as `strap_data_assert`: I think you need `->` instead of `|=>`.

hw/ip_templates/gpio/dv/sva/gpio_strap_check.sv.tpl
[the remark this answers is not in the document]
rswarbrick: I think this comment still applies?

rswarbrick:
Thanks for all the re-working. Please could you address https://github.com/lowRISC/opentitan/pull/26009#discussion_r2009034762?

hw/ip_templates/gpio/dv/sva/gpio_strap_check.sv.tpl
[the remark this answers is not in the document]
marcelocarvalhoLowRisc: Sorry for the delay, I didn't realised that this comment was still pending. Now I think is fixed.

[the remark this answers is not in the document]
marcelocarvalhoLowRisc: I think now is fixed.

rswarbrick:
I this change has bit-rotted and would need quite a bit of work to make useful agani

PR #26010, opened 2025-01-24
opentitanlib: accept ECDSA signatures as ASN.1 blobs

vbendeb:
When processing ECDSA signatures consider the input file to be either a raw signature, with 32 bytes R and S values in little endian format concatenated into a single 64 bytes blob, or an ASN.1 encoded ECDSA signature, which is a sequence of two big endian big number values, which could be anywhere from one to 33 bytes long.

In case input is in ASN .1 format Convert the values into little endian and truncate/expand as necessary to generated the internal representation of R and S of exactly 32 bytes each.

Tested by attaching the Crypto4a HSM produced ECDSA signature to an owner binary image and verifying the signature, both using opentitantool.

### DIFF
--- a/sw/host/opentitanlib/src/crypto/ecdsa.rs
+++ b/sw/host/opentitanlib/src/crypto/ecdsa.rs
@@ -6,7 +6,9 @@ use anyhow::{anyhow, ensure, Context, Result};
 use ecdsa::elliptic_curve::pkcs8::{DecodePrivateKey, EncodePrivateKey};
 use ecdsa::elliptic_curve::pkcs8::{DecodePublicKey, EncodePublicKey};
 use ecdsa::signature::hazmat::PrehashVerifier;
-use p256::ecdsa::{Signature, SigningKey, VerifyingKey};
+use ecdsa::Signature;
+use p256::ecdsa::{SigningKey, VerifyingKey};
+use p256::NistP256;
 use rand::rngs::OsRng;
 use serde::{Deserialize, Serialize};
 use serde_annotate::Annotate;
@@ -97,7 +99,7 @@ impl Default for EcdsaRawSignature {
 impl TryFrom<&[u8]> for EcdsaRawSignature {
     type Error = Error;
     fn try_from(value: &[u8]) -> std::result::Result<Self, Self::Error> {
-        if value.len() != 64 {
+        if value.len() != Self::SIZE {
             return Err(Error::InvalidSignature(anyhow!(
                 "bad length: {}",
                 value.len()
@@ -109,6 +111,8 @@ impl TryFrom<&[u8]> for EcdsaRawSignature {
 }
 
 impl EcdsaRawSignature {
+    const SIZE: usize = 64;
+
     pub fn read(src: &mut impl Read) -> Result<Self> {
         let mut sig = Self::default();
         src.read_exact(&mut sig.r)?;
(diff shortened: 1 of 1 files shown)


vbendeb:
take two, after more thorough testing.

Support processing of ECDSA signatures represented as ASN.1 DER blobs.

vbendeb:
@timothytrippel  @cfrantz  what about this one guys  - build error seems unrelated here either.

sw/host/opentitanlib/src/crypto/ecdsa.rs
```diff
@@ -146,6 +164,46 @@ impl EcdsaRawSignature {
...
+        let mut r = [0u8; 32];
+        let mut s = [0u8; 32];
+
+        // Copy only what's necessary and ignore the leading zeros in case of
+        // 33 byte values.
+        let copy_size = r.len().min(raw_r.len());
+        r[0..copy_size].copy_from_slice(&raw_r[0..copy_size]);
+
+        let copy_size = s.len().min(raw_s.len());
+        s[0..copy_size].copy_from_slice(&raw_s[0..copy_size]);
```
cfrantz, line 200: Could we do away with this copying with [`Vec::resize`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize)?

sw/host/opentitanlib/src/crypto/ecdsa.rs
```diff
@@ -146,6 +164,46 @@ impl EcdsaRawSignature {
     pub fn is_empty(&self) -> bool {
         self.r.iter().all(|&v| v == 0) && self.s.iter().all(|&v| v == 0)
     }
+
+    fn size() -> usize {
+        let sig: EcdsaRawSignature = EcdsaRawSignature::default();
+
+        sig.r.len() + sig.s.len()
+    }
```
cfrantz, line 172: Currently, there is a fixed size requirement of 64 bytes in `EcdsaRawSignature::try_from`.  I'd suggest rather than computing the by creating a `default` object, we add a const to the implementation and use that both in `try_from` and `from_der`.
```rust
impl EcdsaRawSignature {
  const SIZE: usize = 64;
}
```

sw/host/opentitanlib/src/crypto/ecdsa.rs
```diff
@@ -120,7 +122,23 @@ impl EcdsaRawSignature {
     pub fn read_from_file(path: &Path) -> Result<EcdsaRawSignature> {
         let mut file =
             File::open(path).with_context(|| format!("Failed to read file: {path:?}"))?;
-        EcdsaRawSignature::read(&mut file)
+        let file_size = std::fs::metadata(path)
+            .with_context(|| format!("Failed to get metadata for {path:?}"))?
+            .len();
+
+        let raw_size = EcdsaRawSignature::size() as u64;
+        if file_size == raw_size {
```
cfrantz, line 130: Is there a better heuristic we can use here?  Are there some leading bytes that will be present in all ASN.1 encoded blobs? (e.g. `size != Self::SIZE && leading_bytes == whatever`)

vbendeb: done

sw/host/opentitanlib/src/crypto/ecdsa.rs
vbendeb: done

vbendeb:
PTAL

cfrantz (approved):

vbendeb:
@cfrantz @timothytrippel  guys, can you please merge this? We need to start using this in the cros project...

cfrantz (approved):

PR #26011, opened 2025-01-24
[hw,prim_subreg_shadow,rtl] Add Mubi Support to prim_subreg_shadow

Razer6:
Shadowed subregs lacked support for Multibit values. This PR feeds the multibit parameter down to all prims that are used within this one. 

The reason why this is needed now is that `soc_dbg_ctrl` uses shadowed mubi registers.

### DIFF
--- a/hw/ip/prim/rtl/prim_subreg_shadow.sv
+++ b/hw/ip/prim/rtl/prim_subreg_shadow.sv
@@ -77,7 +77,8 @@ module prim_subreg_shadow
 
   prim_subreg_arb #(
     .DW       ( DW       ),
-    .SwAccess ( SwAccess )
+    .SwAccess ( SwAccess ),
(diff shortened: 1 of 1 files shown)


rswarbrick (approved):
This looks sensible to me, and is basically just passing the flag downwards to the arbiter. The change risk to Earlgrey is zero (since the flag has its default value in that configuration)

rswarbrick:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_subreg_shadow.sv

The risk to Earlgrey from this change is zero (since the flag has its default value in that configuration)

alees24 (approved):
Thanks for this.

alees24:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_subreg_shadow.sv

PR #26012, opened 2025-01-24
[ci] Rename the ROM_EXT hyper310 workflow

cfrantz:


a-will (approved):
An itch you just _had_ to scratch? :smile: 

In any case, LGTM!

cfrantz:
> An itch you just _had_ to scratch? 😄
> 
> In any case, LGTM!

I suspect that since there were two CI stages with the name "CW310 ROM_EXT Tests" that one of them was being ignored and not running.  If I look at some different PRs for earlgrey_1.0.0, it seems like the stages downstream of creating the hyper310 bitstream don't include a "CW310 ROM_EXT Tests" stage.  _All_ of my ROM_EXT tests run under hyper310.

Perhaps you can independently look and tell me if I'm correct about that or not?

timothytrippel (approved):

cfrantz:
> Perhaps you can independently look and tell me if I'm correct about that or not?

Maybe I'm mistaken: https://github.com/lowRISC/opentitan/actions/runs/12955253508?pr=25969

a-will:
> > Perhaps you can independently look and tell me if I'm correct about that or not?
> 
> Maybe I'm mistaken: https://github.com/lowRISC/opentitan/actions/runs/12955253508?pr=25969

I did at least spot one PR where there were no tests run for the job changed here, but I don't know whether that is actually expected behavior.

cfrantz:
@nbdd0121, I think there may be a misconfiguration in CI, but I'm unable to determine that for certain.

It seems that we have two stages with a name of "CW310 ROM_EXT Tests".  One is configured for `fpga_cw310` and the other for `hyper310`.  Although they have unique keys in the yaml file, I can't tell if they're both running.  This PR renames the later to "Hyper310 ROM_EXT Tests".

Could you investigate and let me know if we have a configuration error?

moidx (approved):

nbdd0121:
The tests are still ran and checked (you'll find in the sidebar navigation), although I think it's hidden on the flow chart.

cfrantz:
> The tests are still ran and checked (you'll find in the sidebar navigation), although I think it's hidden on the flow chart.

Do you think we ought to rename user-visible name of the workflow so that it's easier to see it in the list of CI stages?

nbdd0121:
Oh we obviously should as this creates confusion, I am just saying that it was indeed running.

PR #26015, opened 2025-01-25
[hw,tlul_adapter_sram,rtl] Add support for RACL protected SRAM Adapter

Razer6:
This PR adds a RACL protected SRAM adapter. This works by acting like a shim around the existing TLUL SRAM adapter. In case of a RACL error, the error is responded via the TLUL loopback adapter. Currently, only a single RACL policy is supported, but this will be extended in the future.

This RACL protected SRAM adapter will be used in SPI host and device as well as in the SRAM Control.

The same approach is implemented for the Reg adapter. This circumvents dealing with any flop latencies in the adapter explicitly, since RACL is check for the TLUL request goes into the TLUL adapter.

### DIFF
--- a/hw/ip/tlul/adapter_reg_racl.core
+++ b/hw/ip/tlul/adapter_reg_racl.core
@@ -11,7 +11,8 @@ filesets:
       - lowrisc:prim:assert
       - lowrisc:prim:secded
       - lowrisc:tlul:common
-      - lowrisc:tlul:trans_intg
+      - lowrisc:tlul:adapter_reg
+      - lowrisc:tlul:request_loopback
       - lowrisc:systems:top_racl_pkg
     files:
       - rtl/tlul_adapter_reg_racl.sv
--- a/hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
+++ b/hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
@@ -52,33 +52,8 @@ module tlul_adapter_reg_racl
   // This can be a write or read error.
   input                    error_i
 );
-  logic racl_read_allowed, racl_write_allowed, racl_error;
-  logic rd_req, wr_req;
-  logic [RegDw-1:0] rdata;
-
-  tlul_adapter_reg #(
-    .CmdIntgCheck       (CmdIntgCheck),
-    .EnableRspIntgGen   (EnableRspIntgGen),
-    .EnableDataIntgGen  (EnableDataIntgGen),
-    .RegAw              (RegAw),
-    .RegDw              (RegDw),
-    .AccessLatency      (AccessLatency)
-  ) tlul_adapter_reg (
-    .clk_i,
-    .rst_ni,
-    .tl_i,
-    .tl_o,
-    .en_ifetch_i,
-    .intg_error_o,
-    .re_o(rd_req),
-    .we_o(wr_req),
-    .addr_o,
-    .wdata_o,
-    .be_o,
-    .busy_i,
-    .rdata_i(rdata),
-    .error_i(racl_error)
-  );
+  tl_h2d_t tl_h2d_filtered;
+  tl_d2h_t tl_d2h_filtered;
 
   if (EnableRacl) begin : gen_racl_role_logic
     // Retrieve RACL role from user bits and one-hot encode that for the comparison bitmap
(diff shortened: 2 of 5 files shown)


hw/ip/tlul/adapter_reg_racl.core
```diff
@@ -11,7 +11,8 @@ filesets:
       - lowrisc:prim:assert
       - lowrisc:prim:secded
       - lowrisc:tlul:common
-      - lowrisc:tlul:trans_intg
+      - lowrisc:tlul:adapter_reg
+      - lowrisc:tlul:request_loopback
```
rswarbrick, line 15: I think this looks reasonable to me. Am I right in thinking that we weren't actually using the `trans_intg` core directly even before this PR?

hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
```diff
@@ -94,30 +69,62 @@ module tlul_adapter_reg_racl
...
+    assign rd_req             = tl_i.a_opcode == tlul_pkg::Get;
     assign racl_read_allowed  = (|(racl_policies_i[RaclPolicySelVec].read_perm  & racl_role_vec));
     assign racl_write_allowed = (|(racl_policies_i[RaclPolicySelVec].write_perm & racl_role_vec));
-    assign racl_error_o       = (rd_req & ~racl_read_allowed) | (wr_req & ~racl_write_allowed);
-    // RACL only generates error responeses if enabled
-    assign racl_error         = racl_error_o & RaclErrorRsp;
+    assign racl_error_o       = (rd_req & ~racl_read_allowed) | (~rd_req & ~racl_write_allowed);
+
+    tlul_request_loopback #(
+      .ErrorRsp(RaclErrorRsp)
```
rswarbrick, line 79: Do we ever expect to instantiate this loopback module without setting `ErrorRsp`? I'm struggling to think of a use case for that. Maybe we should drop the parameter?

hw/ip/tlul/rtl/tlul_adapter_sram_racl.sv
```diff
@@ -0,0 +1,174 @@
...
+      .in_i ( racl_role     ),
+      .en_i ( 1'b1          ),
+      .out_o( racl_role_vec )
+    );
+
+    logic rd_req, racl_read_allowed, racl_write_allowed;
+    assign rd_req             = tl_i.a_opcode == tlul_pkg::Get;
+    assign racl_read_allowed  = (|(racl_policies_i[RaclPolicySelVec].read_perm  & racl_role_vec));
+    assign racl_write_allowed = (|(racl_policies_i[RaclPolicySelVec].write_perm & racl_role_vec));
+    assign racl_error_o       = (rd_req & ~racl_read_allowed) | (~rd_req & ~racl_write_allowed);
```
rswarbrick, line 102: It's not a big deal, but would this be slightly shorter and clearer?
```systemverilog
    assign racl_error_o       = !(rd_req ? racl_read_allowed : racl_write_allowed);
```
(I realise that this is already in the register adapter, so it would definitely be a follow-up PR if so)

rswarbrick (commented):
This looks sensible to me. A couple of minor questions before it lands.

hw/ip/tlul/adapter_reg_racl.core
Razer6: Exactly. The ordinary reg adapter was using that and that one has a direct dependency.

hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
Razer6: Yes, we defined RACL to support 2 forms of error behavior:
1. Return a TLUL error on a RACL block
2. Silently drop a write and return all 1 on read
For case number 2, we need the parametrized error response on the loopback.

hw/ip/tlul/rtl/tlul_adapter_sram_racl.sv
Razer6: Let's make a follow-up PR on that to unblock the others since CI is green at the moment.

hw/ip/tlul/rtl/tlul_adapter_reg_racl.sv
rswarbrick: Ahah, that makes sense. Thanks for explaining.

rswarbrick:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_request_loopback.sv

The protected files are all new for the RACL feature, so editing them (to make the feature work) doesn't really present any risk to existing designs.

vogelpi:
CHANGE AUTHORIZED: hw/ip/tlul/rtl/tlul_request_loopback.sv

The protected files are all new for the RACL feature, so editing them (to make the feature work) doesn't really present any risk to existing designs.

rswarbrick (approved):

PR #26017, opened 2025-01-25
[orchestrator] refactor SKU specific portion of device ID

timothytrippel:
This updates the SKU specific portion of the device ID to add the following subfields:
1. AST config version
2. OTP ID
3. OTP version
4. SKU specific version


timothytrippel:
CC: @benjbender

moidx (approved):

PR #26018, opened 2025-01-25
[RACL] Create TLUL user vector based on a role ctn uid

Razer6:
This PR adds a function to the package to compute the TLUL User RSVD bit vector given a RACL role and CTN UID bits. This function is used to create the user bit vector for TLUL agents such as DM, RV_DM and Ibex.


rswarbrick (commented):
Hmm, the generated code is a bit silly in the examples that we get, because they don't actually *use* racl. Should we condition the template that generates the function so that nothing comes out unless at least one msb is positive (which would demonstrate that they had been configured and weren't all on bit zero)?

Razer6:
I'm happy to gate that with `role_bit_msb  > 0 or ctn_uid_bit_msb > 0`

Razer6:
@rswarbrick I just updated the template to gate rendering the function unless there is an MSB > 0.

rswarbrick (approved):
Cool thanks. This looks sensible to me.

PR #26020, opened 2025-01-26
[hw,pinmux,lint] Read unused signals to avoid lint errors

Razer6:
Depending on the pin config of the pinmux, some signals of the MUX might be unused (It is aligned to the largest DIO size). Read the unused signals to avoid linting errors.

### DIFF
--- a/hw/ip_templates/pinmux/rtl/pinmux.sv.tpl
+++ b/hw/ip_templates/pinmux/rtl/pinmux.sv.tpl
@@ -492,6 +492,14 @@ module pinmux
     assign mio_to_periph_o[k] = mio_mux[reg2hw.mio_periph_insel[k].q];
   end
 
+% if n_dio_pads > n_mio_pads + 2:
+  // For configurations with NMioPads + 2 < NDioPads, mio_in is zero-extended to NDioPads bits for
+  // convenience. However, mio_periph_insel is sized to select the lowest NMioPads + 2 bits. Most
+  // of the zero bits cannot actually be selected. Tie them off to avoid lint warnings.
+  logic unused_mio_mux;
+  assign unused_mio_mux = ^{mio_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
+
+% endif
   //////////////////////
   // MIO Output Muxes //
   //////////////////////
--- a/hw/ip_templates/pinmux/data/pinmux.hjson.tpl
+++ b/hw/ip_templates/pinmux/data/pinmux.hjson.tpl
@@ -1136,7 +1136,7 @@
                   regwen_multi: "true",
                   cname:        "DETECTOR",
                   fields: [
-                    { bits: "${max(2 + n_mio_pads-1, n_dio_periph_in-1).bit_length()-1}:0",
+                    { bits: "${max(2 + n_mio_pads-1, n_dio_pads-1).bit_length()-1}:0",
                       name: "SEL",
                       resval: 0,
                       desc: '''Selects a specific MIO or DIO pad (depending on !!WKUP_DETECTOR configuration).
--- a/hw/top_darjeeling/ip_autogen/pinmux/data/pinmux.hjson
+++ b/hw/top_darjeeling/ip_autogen/pinmux/data/pinmux.hjson
@@ -824,7 +824,7 @@
                   regwen_multi: "true",
                   cname:        "DETECTOR",
                   fields: [
-                    { bits: "5:0",
+                    { bits: "6:0",
                       name: "SEL",
                       resval: 0,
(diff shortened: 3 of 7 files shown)


rswarbrick (commented):
I want to make sure that I understand this. If I've got it right, this mux is sized to be the first power of two that is at least `NMioPads+2` and `NDioPads`.

Presumably (for the configuration you're interested in), `NDioPads` is much larger so the index in this line:
```systemverilog
    assign mio_to_periph_o[k] = mio_mux[reg2hw.mio_periph_insel[k].q];
```
can never be all that big? (as checked by a lint tool)

Is that correct? If so, can you add a note to the commit message that explains things (and I'm very happy to take the change!)

vogelpi:
We've discussed that it would be good to add a comment / commit message and that ideally we can change the template to not insert the code if it is unneeded. I.e., we also won't need to authorize any code change for Earlgrey.

Razer6:
@vogelpi As discussed, I gated the read of the unused signals as part of the template rendering.

hw/ip_templates/pinmux/rtl/pinmux.sv.tpl
```diff
@@ -613,6 +618,13 @@ module pinmux
   // 1, respectively.
   assign mio_wkup_mux = AlignedMuxSize'({mio_wkup_no_scan, 1'b1, 1'b0});
 
+  % if n_dio_pads > n_mio_pads + 2:
+    logic unused_mio_wkup_mux_signals;
+    assign unused_mio_wkup_mux_signals = ^{mio_wkup_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
+    logic unused_dio_wkup_mux_signals;
+    assign unused_dio_wkup_mux_signals = ^{dio_wkup_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
+  % endif
```
nasahlpa, line 626:
```suggestion
% if n_dio_pads > n_mio_pads + 2:
  logic unused_mio_wkup_mux_signals;
  assign unused_mio_wkup_mux_signals = ^{mio_wkup_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
  logic unused_dio_wkup_mux_signals;
  assign unused_dio_wkup_mux_signals = ^{dio_wkup_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
% endif
```

hw/ip_templates/pinmux/rtl/pinmux.sv.tpl
```diff
@@ -613,6 +618,13 @@ module pinmux
   // 1, respectively.
   assign mio_wkup_mux = AlignedMuxSize'({mio_wkup_no_scan, 1'b1, 1'b0});
 
+% if n_dio_pads > n_mio_pads + 2:
+  logic unused_mio_wkup_mux_signals;
+  assign unused_mio_wkup_mux_signals = ^{mio_wkup_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
+  logic unused_dio_wkup_mux_signals;
+  assign unused_dio_wkup_mux_signals = ^{dio_wkup_mux[(AlignedMuxSize - 1):(NMioPads + 2)]};
+
+% endif
```
vogelpi, line 627: After taking another look at this I think this actually doesn't make sense. Either you have more MIOs or more DIOs, so it should only be required to waive either unused DIOs or unused MIOs but not both.

vogelpi (changes requested):
Thanks for making the proposed changes @Razer6 . 

After taking another look at this and also reading @rswarbrick comment and questions (thanks for the thorough review Rupert), I realized that once more the lint tool is right here and these waivers shouldn't be added in the first place.

What actually happens is that you have either 128 MIOs or DIOs (I haven't checked which one is bigger, it doesn't matter) but from the register file the padsel signals are only 6 bits wide. That's why the mux signals with indices 127 to 64 can never be selected.

What should be changed instead is the pinmux hjson for Darjeeling. At least the WKUP_DETECTOR_PADSEL register should be extend from 5:0 to 6:0 bits.

vogelpi:
@Razer6 , I've now implemented the change to correctly size the PADSEL registers. Interestingly, part of your lint waivers are still required. I've now updated your commit accordingly and added comments as requested by @rswarbrick .

Razer6:
@vogelpi Great. That seems it was a bug in the HJSON? but it never triggered because of the specific pad configuration of Earlgrey?

vogelpi:
> @vogelpi Great. That seems it was a bug in the HJSON? but it never triggered because of the specific pad configuration of Earlgrey?

Yes exactly!

vogelpi (approved):

PR #26021, opened 2025-01-26
[hw,darjeeling,rtl] Separate ROM ctrl cfg inputs

Razer6:
Downstream, the DFT signals for both ROM ctrls are different. Let's add separate top-level inputs for that and let the AST connect to both in the uncore environment.


rswarbrick (approved):
I think this makes sense and it looks like a pretty simple change (once I've unfolded the autogen stuff!). Seems reasonable to me.

PR #26022, opened 2025-01-26
[hw,prim alert,rtl] Add flop repeater for alert signals

Razer6:
Alerts are all collected by Darjeeling. IPs that are producing alerts are distributed in various places in the SoC, meaning there might a long distance to travel to Darjeeling. To cut the path, also when travelling across physical design partitions, it's necessary to flop signals, also alerts.

To make this easier, we add a repeater module tailored for alerts that flops both the TX and RX signal.

### DIFF
--- a/hw/ip/prim/rtl/prim_alert_repeater.sv
+++ b/hw/ip/prim/rtl/prim_alert_repeater.sv
@@ -1,0 +1,73 @@
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Staged flop repeater to cut the crirtical path on alert signals when travelling
+// in the larger SoC. Both TX and RX are staged with a single flop stage.
+
+module prim_alert_repeater
+  import prim_alert_pkg::*;
+(
+  input                             clk_i,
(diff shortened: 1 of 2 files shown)


hw/ip/prim/rtl/prim_alert_repeater.sv
a-will, line 1: Should this module be generally discouraged for use with alerts on asynchronous clock domains? Given all the individual flops and the single domain used for both directions, the module doesn't give me the impression that it was intended for async use.

hw/ip/prim/rtl/prim_alert_repeater.sv
```diff
@@ -0,0 +1,73 @@
...
+  // Flopped output interface
+  output prim_alert_pkg::alert_tx_t alert_staged_tx_o,
+  input  prim_alert_pkg::alert_rx_t alert_staged_rx_i
+);
+  prim_flop #(
+    .ResetValue(0)
+  ) u_tx_p (
+    .clk_i,
+    .rst_ni,
+    .d_i    (alert_tx_i.alert_p),
```
a-will, line 25: Do these need prim_sec_anchor_buf to protect from optimizations that eliminate the second wire, as in prim_alert_receiver?

Razer6:
Retiring this PR

PR #26024, opened 2025-01-26
[hw,soc_proxy,rtl] Fix assertion

Razer6:
This assertion was added recently but was using the wrong TLUL interface.


rswarbrick (approved):
Looks good to me (fixing a typo)

PR #26025, opened 2025-01-26
sphincsplus: accept PUB key as PEM encoded ASN.1 objects

vbendeb:
Some HSMs distribute SPX pub keys as PEM encoded ANS.1 objects. The approach so far has been to manually extract the 32 byte key material from the object. This patch adds parsing code for processing the ANS.1 key files.

The algorithm type is deduced from the ANN.1 OID, and is hardcoded to the value generated by Crypto4a HSM for SHA2 128S Simple encoding. This OID has not yet been standardized and could change in the future.

Tested on an opentitatn owner's firmware by passing to --spx-key first the raw SPX key and then the HSM generated PEM file, and observing the unchanged contents of the resulting image.


vbendeb:
support SPK key extraction from both raw and ANS.1 PEM files.

vbendeb:
guys, any comments on this?

sw/host/sphincsplus/key.rs
```diff
@@ -208,7 +209,34 @@ impl DecodeKey for SpxPublicKey {
...
+                    Ok((oid, key))
+                })
+            });
+            let (oid, key) = match r {
+                Ok((oid, key)) => (oid, key),
+                Err(_) => return Err(SpxError::ParseError(format!("Not a RAW key {label:?}"))),
+            };
+
+            // Semi standard OID used by Crypto4a HSM for DER representation of the SPX public key.
+            let expected = ObjectIdentifier::from_string("2.16.840.1.101.3.4.3.35").unwrap();
```
cfrantz, line 229: Can we make the OID a const using the `asn1::oid!` macro?
```rust
impl SpxPublicKey {
    const OID_PUBLIC_KEY: ObjectIdentifier = asn1::oid!(2, 16, 840, 1, ...);
}

// and in this block:
if oid == Self::OID_PUBLIC_KEY { ... }
```

sw/host/sphincsplus/key.rs
```diff
@@ -208,7 +209,34 @@ impl DecodeKey for SpxPublicKey {
             return Err(SpxError::ParseError(format!("not a public key: {label:?}")));
         }
         if !algorithm.starts_with("RAW:") {
-            return Err(SpxError::ParseError(format!("not a RAW key: {label:?}")));
+            // The PEM block is not a raw key, maybe it is an ASN.1 object.
```
cfrantz, line 212: Maybe this whole block could be a private helper function on `SpxPublicKey`:
```rust
if !algorithm.starts_with("RAW:") {
  return Self::parse_asn1_public_key(&key);
}
```

cfrantz (commented):
LGTM with a couple of nits.

sw/host/sphincsplus/key.rs
vbendeb: done

sw/host/sphincsplus/key.rs
vbendeb: done

vbendeb:
PTAL

vbendeb:
build error seems unrelated, can this be merged?

PR #26026, opened 2025-01-27
[opentitantool] do not create duplicate extension entries

vbendeb:
It was observed that repetitive invoking 'opentitiantool image manifest update ...' passing an --spx_key and same manifest input file which included SPX key and signature extensions repetitively is causing addition of the SPX key and SPX signature space every time opentitantool is invoked.

This patch fixes the problem.

All opentitantool and opentitanlib tests pass, and running the same invocation does not cause additions of the SPX key and signature spaces any more.

Signed-off-by: Vadim Bendebury <vbendeb@google.com>
(cherry picked from commit 7e34e67ade0717418930ecbfa4bee30ba44979ed)


vbendeb:
this seems to have slipped through the cracks, is in master but not in earlgreay_1.0.0

cfrantz (approved):

PR #26027, opened 2025-01-27
[rom] Update clkmgr jitter enable to match ROM specification.

moidx:
The ROM e2e plan captures the expected behavior of the clock jitter enable flow:

> `CREATOR_SW_CFG_JITTER_EN`: `kMultiBitBool4False` disables, all other values
>  enables. `CREATOR_SW_CFG_AST_INIT_EN` must also be `kMultiBitBool4True` to enable.

This change refactors the code in `rom_start.S` and `test_rom_start.S` to explicitly reduce any value different than MubiFalse to MubiTrue. The implementation relies on the register to be reset to MubiFalse by default.

### DIFF
--- a/sw/device/lib/testing/test_rom/test_rom_start.S
+++ b/sw/device/lib/testing/test_rom/test_rom_start.S
@@ -170,12 +170,17 @@ _start:
   andi t0, t0, 0x1
   beqz t0, .L_ast_done_loop
 
-  // Enable jittery clock if enabled in OTP.
+  // Enable jittery clock if not strictly disabled in OTP.
+  // Use hardcoded MULTIBIT_ASM_BOOL4_TRUE as the enable word to reduce all
+  // other possible OTP values to true.
   li   a0, (TOP_EARLGREY_OTP_CTRL_CORE_BASE_ADDR + \
             OTP_CTRL_SW_CFG_WINDOW_REG_OFFSET)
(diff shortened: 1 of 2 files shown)


vogelpi (approved):
This is inline with the confirmed ECO, LGTM. Thanks for making the change @moidx !

cfrantz (approved):

timothytrippel (approved):

sw/device/lib/testing/test_rom/test_rom_start.S
```diff
@@ -170,12 +170,17 @@ _start:
   andi t0, t0, 0x1
   beqz t0, .L_ast_done_loop
 
-  // Enable jittery clock if enabled in OTP.
+  // Enable jittery clock if not strictly disabled in OTP.
+  // Use hardcoded MULTIBIT_ASM_BOOL4_TRUE as the enable word to reduce all
```
matutem, line 174: Do you mean "Use MULTIBIT_ASM_BOOL4_FALSE as the disable word to reduce all other possible OTP values to enable."

moidx: I meant to use the `MULTIBIT_ASM_BOOL4_TRUE` word as the write value to reduce all other options to True (Enable).

PR #26029, opened 2025-01-27
[prim,clkmgr,rtl] Simplify prim_clock_meas by removing RefCnt

rswarbrick:
This parameter is only ever set to 1: we don't do things like testing a clock that should be 50% faster by setting Cnt=30, RefCnt=20. Simplify the code slightly by getting rid of the support.

This commit also expands a comment explaining Cnt: my initial read suggested that this was the maximum value of a counter, which gives the wrong width if Cnt is a power of two.

### DIFF
--- a/hw/ip/prim/rtl/prim_clock_meas.sv
+++ b/hw/ip/prim/rtl/prim_clock_meas.sv
@@ -13,10 +13,9 @@
 `include "prim_assert.sv"
 
 module prim_clock_meas #(
-  // Maximum value of input clock counts over measurement period
+  // Maximum number of input clock counts over measurement period. Note that this is 1-based
+  // indexing (so a hardware counter would only count up to Cnt-1)
   parameter int Cnt = 16,
-  // Maximum value of reference clock counts over measurement period
-  parameter int RefCnt = 1,
   parameter bit ClkTimeOutChkEn = 1,
   parameter bit RefTimeOutChkEn = 1,
   localparam int CntWidth = prim_util_pkg::vbits(Cnt),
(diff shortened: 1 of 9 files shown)


hw/ip/prim/rtl/prim_clock_meas.sv
```diff
@@ -155,27 +154,10 @@ module prim_clock_meas #(
...
-        cnt_ref <= '0;
-      end else if (!cnt_en && |cnt_ref) begin
-        cnt_ref <= '0;
-      end else if (cnt_en && valid) begin
-        cnt_ref <= '0;
-      end else if (cnt_en && valid_ref) begin
-        cnt_ref <= cnt_ref + 1'b1;
-      end
-    end
-  end
```
vogelpi, line 178: The following points make me think that it is fine to remove this option but I would like to get your input on this as well Rupert:

- There is no single instance in any design with `RefCnt != 0`. So this feature is not used anywhere.
- We don't have any DV for this feature (from the Git history, I can see that the `RefCnt == 1` special case was specifically added to ease coverage closure). So even if we would start using the feature, there would be additional DV work.
- With a `RefCnt` value >= 1, the module can count clock cycles across multiple reference clock cycles thereby allowing to average count values across multiple reference clocks. With `RefCnt == 1`, every reference clock edge clears the counting and triggers a comparison with the threshold registers. I would argue that since we have configurable low and high thresholds in clkmgr, we can just tune these to increase tolerance if needed. But I don't see a strong argument for averaging across multiple reference clocks. Or do I miss something?

vogelpi (approved):
I am okay with the change assuming I understand this all correctly (see my comment), sorry for the delay @rswarbrick .

Before merging this, would you mind rebasing this please and wait for CI to run through? Because the PR is 3 weeks old now.

hw/ip/prim/rtl/prim_clock_meas.sv
rswarbrick: I think I completely agree. Annoyingly, my "git grep" promptly found some stray uses of the parameter name! So I'll force-push in a sec to expunge them as well.

And I'll rebase and wait for CI.

vogelpi: Thanks Rupert!

vogelpi:
CHANGE AUTHORIZED: hw/ip/prim/rtl/prim_clock_meas.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr_meas_chk.sv

This PR removes an unused and untested feature. This has no functional impact.

rswarbrick:
Oops: I left a stray usage of a computed parameter (the reason for the CI failure). The force-push fixes that silliness.

vogelpi:
Hi @rswarbrick , IIUC @matutem suggested that we should actually keep the feature. Let's put this on hold and discuss in the issue here: #26385

hw/ip/prim/rtl/prim_clock_meas.sv
matutem: There is a valid use for RefCnt > 1 for darjeeling. See https://github.com/lowRISC/opentitan/issues/26385.

PR #26031, opened 2025-01-27
[darjeeling,lint] Fix lint config for Darjeeling

Razer6:
Some IPs changed.  Update the Lint config


rswarbrick (approved):
Thanks for this. I've gone through and I'm convinced:
- flash_ctrl probably shouldn't have been in the list in the first place.
- pinmux became a Darjeeling version on 21st Nov
- The "socdbg_ctrl" name hasn't ever been valid (on this branch in this repo...)
- sensor_ctrl was removed just over a week ago.

Thanks for the tidy-up.

Razer6:
I think Pirmin copied that file over from the `integrated_dev` branch, where all of the points where valid.

PR #26033, opened 2025-01-27
[aon_timer, dv] Update TB to match RTL behavior

antmarzam:
Now the RTL has been fix and resets the internal prescale count the TB removes all the backdoor reads it had to access the internal RTL counter.


rswarbrick (approved):
Well that's rather simpler :-) Thanks!

PR #26034, opened 2025-01-27
[flash_ctrl] Relax the hardening in `flash_ctrl_data_region_protect`

cfrantz:
The hardening in the function `flash_ctrl_data_region_protect` resulted in 1304 bytes of RV32 machine code.  This relaxed version results in 250 bytes of RV32 machine code.

### DIFF
--- a/sw/device/silicon_creator/lib/drivers/flash_ctrl.c
+++ b/sw/device/silicon_creator/lib/drivers/flash_ctrl.c
@@ -523,157 +523,59 @@ flash_ctrl_cfg_t flash_ctrl_data_default_cfg_get(void) {
   };
 }
 
-// This X macro helps to generate code that operates on each of the flash_ctrl
-// memory protection regions.
-#define FLASH_CTRL_MP_REGIONS(X) \
-  X(0)                           \
-  X(1)                           \
-  X(2)                           \
-  X(3)                           \
-  X(4)                           \
-  X(5)                           \
-  X(6)                           \
-  X(7)
-
-// Defines the bounds of the given memory protection region by setting the
-// MP_REGION_${region} register.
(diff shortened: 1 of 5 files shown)


sw/device/silicon_creator/lib/drivers/flash_ctrl.c
```diff
@@ -523,157 +523,52 @@ flash_ctrl_cfg_t flash_ctrl_data_default_cfg_get(void) {
...
                                     flash_ctrl_cfg_t cfg,
                                     hardened_bool_t lock) {
+  HARDENED_CHECK_GE(region, 0);
+  HARDENED_CHECK_LT(region, 8);
+  region *= sizeof(uint32_t);
+
   // Reset the region's configuration via the MP_REGION_CFG_${region} register.
   // This temporarily disables memory protection for the region.
-  flash_ctrl_mp_region_cfg_reset(region);
+  sec_mmio_write32(kBase + FLASH_CTRL_MP_REGION_CFG_0_REG_OFFSET + region,
```
moidx, line 537: We should consider adding a `SEC_MMIO_ASSERT_WRITE_INCREMENT()` to keep track of the `sec_mmio` counter increment.  

I am not sure if we are doing any sec_mmio checks in the rom_ext after ownership commands have been executed. This is something we can discuss as a separate issue.

moidx (approved):

sw/device/silicon_creator/lib/drivers/flash_ctrl.c
moidx: Sounds good. I think we can do this in a follow up PR once we review the `sec_mmio` hardening for the  ROM_EXT.

cfrantz: Done.

Issue #2254, opened 2020-05-18
[reggen] Adding index to register when there's only a single register


GregAC:
Currently in reggen where you have a multireg where the `count` isn't large enough that it needs to generate multiple registers it doesn't add an index to the register name when generating them. This causes an issue when trying to write something that uses the generated register definitions that is intended to work for any `count`.

As a concrete example in the padctrl DIF (currently in a PR so link is to my fork) `PADCTRL_DIO_PADS0_REG_OFFSET` is used and this becomes `PADCTRL_DIO_PADS_REG_OFFSET` depending upon the number of DIO pads:

https://github.com/GregAC/opentitan/blob/1b9ce558ddb0c440950525a890b1917773923576/sw/device/lib/dif/dif_padctrl.c#L64-L70

Fixing this in the C is a bit messy as it would require some `#ifdef` to determine if `PADCTRL_DIO_PADS0_REG_OFFSET` or `PADCTRL_DIO_PADS_REG_OFFSET` should be used.

Fixing this in reggen is easy enough, but disruptive, the python that does this is here: 

https://github.com/lowRISC/opentitan/blob/7a98ea292907689b15bcb7897f7163b8e6508680/util/reggen/validate.py#L1140-L1152

However removing this behaviour alters various autogened headers which breaks existing code. This can be fixed but I wanted to see what people thought before I went and did this. Perhaps there's a reason to preserve this behaviour I haven't considered?

As an alternative we could have reggen generated some additional define that gives you the 'base offset' who's name doesn't change (which in the example above would be equal to `PADCTRL_DIO_PADS0_REG_OFFSET`/`PADCTRL_DIO_PADS_REG_OFFSET`), ultimately code that can work with any `count` will want some base it can then compute an appropriate offset to so referring to something like `PADCTRL_DIO_PADS_REG_OFFSET_BASE` (as opposed to something that's meant to be a specific register) may make more sense.

GregAC:
@eunchan @msfschaffner @tjaychen any thoughts?

msfschaffner:
Thanks for bringing this up @GregAC. 

I do not know for sure, but I think the index omission in cases where all multireg fields fit within one physical register is only for aesthetical purposes.

Another place where this can cause issues is in our markdown documentation, where we sometimes reference registers by name (`!!REGNAME`). Those references can break as well if the name changes based on the `count`.

[...]

PR #26038, opened 2025-01-27
[multitop] Port OTTF and test_rom to multitop

pamaury:
This PR does several things:
- convert the OTTF, test_rom and several test, and uart_smoketest utils to multitop
- port those so that they work on darjeeling and (hopefully) english breakfast
- tweak `opentitan_{binary,test}` so that they become minimally multitop aware
- enable LTO on mask_rom

It should be noted that even though the purpose is to make the test_rom working on multitop, this requires to touch code in silicon_creator which is shared with the mask_rom. As a result, DT usage makes its way into the mask_rom which causes its size to increase. Therefore I also had to enable LTO in the mask_rom to optimize away the DT usage.

This PR is large but most commits are very small, I would recommend that reviewers all look at commits about the code that they are familiar with and skip the others. Although it's technically possible to split this PR, it would result in untested code being merged so I think it's not worth it.

Note: I haven't yet be able to check if everything works for english breakfast, probably the final fixes for EB will be done in a future PR.


pamaury:
The PR finally passes the CI. The only thing not done in this PR is wiring english breakfast which I suggest to postpone to another PR.

sw/device/lib/testing/pinmux_testutils.c
```diff
@@ -47,22 +69,24 @@ void pinmux_testutils_init(dif_pinmux_t *pinmux) {
...
+      pinmux, dt_uart_periph_io(kUart1Dt, kDtUartPeriphIoRx), kPadUart1Rx));
+  CHECK_DIF_OK(dif_pinmux_mio_select_output(pinmux, kPadUart1Rx,
+                                            kDtPeriphIoConstantHighZ));
+  // Configure UART1 TX output.
+  CHECK_DIF_OK(dif_pinmux_mio_select_output(
+      pinmux, kPadUart1Tx, dt_uart_periph_io(kUart1Dt, kDtUartPeriphIoTx)));
+#endif
+
+  // TODO convert to multitop
+#ifdef OPENTITAN_IS_EARLGREY
```
jwnrt, line 89: Should we have an `#else #warning` after this to make it more obvious it's unsupported for Darjeeling, or will that be too noisy?

sw/device/lib/testing/pinmux_testutils.c
```diff
@@ -47,22 +69,24 @@ void pinmux_testutils_init(dif_pinmux_t *pinmux) {
...
+#ifdef HAS_UART1
+  // Configure UART1 RX input.
+  CHECK_DIF_OK(dif_pinmux_mio_select_input(
+      pinmux, dt_uart_periph_io(kUart1Dt, kDtUartPeriphIoRx), kPadUart1Rx));
+  CHECK_DIF_OK(dif_pinmux_mio_select_output(pinmux, kPadUart1Rx,
+                                            kDtPeriphIoConstantHighZ));
+  // Configure UART1 TX output.
+  CHECK_DIF_OK(dif_pinmux_mio_select_output(
+      pinmux, kPadUart1Tx, dt_uart_periph_io(kUart1Dt, kDtUartPeriphIoTx)));
+#endif
```
jwnrt, line 86: Please could you add `#endif // HAS_UART1` or similar throughout this file? There's lots of conditional compilation and in some places it's hard to track which pairs match up.

sw/device/lib/testing/pinmux_testutils.c
```diff
@@ -138,6 +162,7 @@ const dif_pinmux_index_t kPinmuxTestutilsGpioMioOutPins[kDifGpioNumPins] = {
     kTopEarlgreyPinmuxMioOutIor6,  kTopEarlgreyPinmuxMioOutIor7,
     kTopEarlgreyPinmuxMioOutIor10, kTopEarlgreyPinmuxMioOutIor11,
     kTopEarlgreyPinmuxMioOutIor12, kTopEarlgreyPinmuxMioOutIor13};
+#endif
```
jwnrt, line 152: I can't actually find the corresponding `#if` for this, did this come from a future commit?

sw/device/lib/testing/pinmux_testutils.c
pamaury: done, I have added comments to all conditions in the file

sw/device/lib/testing/pinmux_testutils.c
pamaury: That will be too noisy, what the comment here means is that we should not check for earlgrey but instead probably check for the USB IP. On the other hand, the pin `IOC7` is hardcoded which is not ideal. This is why I currently left it as is.

sw/device/lib/testing/pinmux_testutils.c
pamaury: Yes I made a mistake, I will fix that

sw/device/lib/dif/dif_pinmux.c
```diff
@@ -338,6 +348,39 @@ dif_result_t dif_pinmux_pad_write_attrs(const dif_pinmux_t *pinmux,
...
+    case kDtPadTypeMio:
+      *type_out = kDifPinmuxPadKindMio;
+      *index_out = dt_pad_mio_pad(pad);
+      return kDifOk;
+    case kDtPadTypeDio:
+      *type_out = kDifPinmuxPadKindDio;
+      *index_out = dt_pad_dio_pad(pad);
+      return kDifOk;
+    default:
+      return kDifBadArg;
```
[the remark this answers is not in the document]
jwnrt: Yes, I think this would be better

engdoreis (commented):
Thanks for the atomic commits.
I left some comments.

sw/device/silicon_creator/lib/drivers/BUILD
```diff
@@ -807,7 +807,7 @@ dual_cc_library(
         device = [
             "//hw/top:flash_ctrl_c_regs",
             "//hw/top:spi_device_c_regs",
-            "//hw/top_earlgrey/sw/autogen:top_earlgrey",
+            "//hw/top:devicetables",
```
jwnrt: I think you need to squash the next commit into this one, this looks like the old name.

sw/device/silicon_creator/lib/drivers/flash_ctrl.c
```diff
@@ -32,12 +34,13 @@ static_assert(kFlashCtrlPartitionInfo1 >> 1 == 1,
...
 
-enum {
-  /**
-   * Base address of the flash_ctrl registers.
-   */
-  kBase = TOP_EARLGREY_FLASH_CTRL_CORE_BASE_ADDR,
-};
+/**
+ * Base address of the flash_ctrl registers.
+ *
```
jwnrt, line 39: Tiny nit: stray newline

sw/device/silicon_creator/lib/drivers/flash_ctrl.h
```diff
@@ -105,13 +105,13 @@ typedef struct flash_ctrl_info_page {
    */
   uint32_t base_addr;
   /**
-   * Config write-enable register address.
+   * Config write-enable register address (offset from register base).
    */
-  uint32_t cfg_wen_addr;
+  uint32_t cfg_wen_off;
```
jwnrt, line 110: Do we have precedent for using an `_off` suffix? I think I mostly see `_offset` in macros etc.

sw/device/lib/testing/test_rom/test_rom_start.S
```diff
@@ -2,17 +2,58 @@
...
+#include "sensor_ctrl_regs.h"
+#include "hw/top_earlgrey/sw/autogen/top_earlgrey_memory.h"
+#define OTP_CTRL_CORE_BASE_ADDR             TOP_EARLGREY_OTP_CTRL_CORE_BASE_ADDR
+#define AST_BASE_ADDR                       TOP_EARLGREY_AST_BASE_ADDR
+#define CLKMGR_AON_BASE_ADDR                TOP_EARLGREY_CLKMGR_AON_BASE_ADDR
+#define SRAM_CTRL_MAIN_REGS_BASE_ADDR       TOP_EARLGREY_SRAM_CTRL_MAIN_REGS_BASE_ADDR
+#define ENTROPY_SRC_BASE_ADDR               TOP_EARLGREY_ENTROPY_SRC_BASE_ADDR
+#define CSRNG_BASE_ADDR                     TOP_EARLGREY_CSRNG_BASE_ADDR
+#define EDN0_BASE_ADDR                      TOP_EARLGREY_EDN0_BASE_ADDR
+#define SENSOR_CTRL_BASE_ADDR               TOP_EARLGREY_SENSOR_CTRL_AON_BASE_ADDR
```
jwnrt, line 37: Do you think this will be a common pattern? Should we potentially extract these aliases to another file that can be included?

sw/device/lib/dif/dif_pinmux.c
```diff
@@ -338,6 +348,39 @@ dif_result_t dif_pinmux_pad_write_attrs(const dif_pinmux_t *pinmux,
...
+  }
+}
+
+dif_result_t dif_pinmux_pad_write_attrs_dt(const dif_pinmux_t *pinmux,
+                                           dt_pad_t pad,
+                                           dif_pinmux_pad_attr_t attrs_in,
+                                           dif_pinmux_pad_attr_t *attrs_out) {
+  dif_pinmux_index_t index;
+  dif_pinmux_pad_kind_t type;
+  dif_result_t res = dif_pinmux_pad_from_dt_pad(pad, &index, &type);
```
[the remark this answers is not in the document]
pamaury: done, thanks for the suggestion

sw/device/lib/testing/pinmux_testutils.c
```diff
@@ -13,29 +15,49 @@
 #include "sw/device/lib/runtime/hart.h"
 #include "sw/device/lib/testing/test_framework/check.h"
 
-#include "hw/top_earlgrey/sw/autogen/top_earlgrey.h"
+static const dt_gpio_t kGpioDt = kDtGpio;
+static const dt_uart_t kUart0Dt = kDtUart0;
+
+#if defined(OPENTITAN_IS_EARLGREY) || defined(OPENTITAN_IS_ENGLISHBREAKFAST)
+static const dt_pad_t kPadUart0Tx = kDtPadIoc4;
```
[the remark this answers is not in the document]
pamaury: yes I agree this is not an ideal solution. In the future I would like for those pins to be automatically set per top/per test board in some library but that's for a future PR.

sw/device/silicon_creator/lib/manifest.c
```diff
@@ -6,7 +6,18 @@
 
 #include "sw/device/silicon_creator/lib/base/chip.h"
 
+#if defined(OPENTITAN_IS_ENGLISHBREAKFAST)
+#include "hw/top_englishbreakfast/sw/autogen/top_englishbreakfast.h"
```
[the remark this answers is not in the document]
pamaury: Normally I would just use the DT to get the data but currently the DT only stores the start address of the memory block, not the size. The solution is to include the size in the DT. I will do that in the next PR.

sw/device/lib/testing/spi_device_testutils.c
```diff
@@ -4,11 +4,23 @@
...
 #include "sw/device/lib/dif/dif_spi_device.h"
 #include "sw/device/lib/testing/test_framework/check.h"
 
 #define MODULE_ID MAKE_MODULE_ID('s', 'd', 't')
 
+/*
+ * The SPI device pins for which `spi_device_testutils_configure_pad_attrs()`
+ * configures the pad attributes.
+ */
+static const dt_pad_t spi_device_direct_pads[4] = {
```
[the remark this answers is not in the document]
pamaury: done

sw/device/silicon_creator/lib/drivers/BUILD
pamaury: yes

sw/device/silicon_creator/lib/drivers/flash_ctrl.c
pamaury: done

sw/device/silicon_creator/lib/drivers/flash_ctrl.h
pamaury: I would think `off` is pretty clear but ok

sw/device/lib/testing/test_rom/test_rom_start.S
pamaury: At the moment this is the only place where it is needed. To be honest I am not a fan of how `test_rom_start.S` works, it has too much top specific init. It would be good to think about how we can reorganize that file for multitop in the future.

jwnrt (approved):
I've done a pass on the commits and discussed with @pamaury and I didn't spot any major issues. Some future work is tracked in https://github.com/lowRISC/opentitan/issues/26116.

LGTM, thanks @pamaury

cfrantz (approved):
The porting of test_rom and the silicon_creator drivers LGTM.

sw/device/lib/testing/test_framework/freertos_port.c
```diff
@@ -59,8 +56,7 @@ void vPortSetupTimerInterrupt(void) {
   LOG_INFO("Configuring timer interrupt ...");
 
   // Initialize and reset the timer.
-  CHECK_DIF_OK(dif_rv_timer_init(
-      mmio_region_from_addr(TOP_EARLGREY_RV_TIMER_BASE_ADDR), &timer));
+  CHECK_DIF_OK(dif_rv_timer_init_from_dt(kDtRvTimer, &timer));
```
a-will, line 59: Again, kDtRvTimer is a top-specific name.

sw/device/lib/testing/test_framework/ottf_console.c
```diff
@@ -129,7 +131,7 @@ void ottf_console_init(void) {
       // configured. The default is to use UART0.
       if (base_addr == 0) {
         CHECK(kOttfTestConfig.console.type == kOttfConsoleUart);
-        base_addr = TOP_EARLGREY_UART0_BASE_ADDR;
+        base_addr = dt_uart_primary_reg_block(kDtUart0);
```
a-will, line 134: nit: kDtUart0 is a top-specific name.

sw/device/lib/testing/test_framework/ottf_console.c
```diff
@@ -256,24 +258,17 @@ void ottf_console_configure_spi_device(uintptr_t base_addr) {
...
+      return dt_uart_irq_to_plic_id(uart, kDtUartIrqRxWatermark);
+    }
   }
+  return dt_uart_irq_to_plic_id(kDtUart0, kDtUartIrqRxWatermark);
 }
 
 void ottf_console_flow_control_enable(void) {
-  CHECK_DIF_OK(dif_rv_plic_init(
-      mmio_region_from_addr(TOP_EARLGREY_RV_PLIC_BASE_ADDR), &ottf_plic));
+  CHECK_DIF_OK(dif_rv_plic_init_from_dt(kDtRvPlic, &ottf_plic));
```
a-will, line 271: nit: Using top-specific names here.

sw/device/lib/testing/test_framework/ottf_isrs.c
```diff
@@ -79,10 +78,18 @@ void ottf_generic_fault_print(uint32_t *exc_info, const char *reason,
     }
     uint32_t *sp = exc_info + kExcWords;
     base_printf("\n");
-    uint32_t *ram_start = (uint32_t *)TOP_EARLGREY_SRAM_CTRL_MAIN_RAM_BASE_ADDR;
-    uint32_t *ram_end =
-        (uint32_t *)(TOP_EARLGREY_SRAM_CTRL_MAIN_RAM_BASE_ADDR +
-                     TOP_EARLGREY_SRAM_CTRL_MAIN_RAM_SIZE_BYTES);
+    uint32_t ram_base_addr =
+        dt_sram_ctrl_reg_block(kDtSramCtrlMain, kDtSramCtrlRegBlockRam);
```
a-will, line 82: nit: This one is kind of going to be an interesting problem for disambiguation, but... `kDtSramCtrlMain` is a top-specific name.

sw/device/lib/testing/test_framework/ottf_isrs.c
```diff
@@ -194,14 +201,12 @@ bool ottf_console_flow_control_isr(uint32_t *exc_info) { return false; }
...
+  const uint32_t kPlicTarget = 0;
   dif_rv_plic_irq_id_t plic_irq_id;
   CHECK_DIF_OK(dif_rv_plic_irq_claim(&ottf_plic, kPlicTarget, &plic_irq_id));
 
-  top_earlgrey_plic_peripheral_t peripheral = (top_earlgrey_plic_peripheral_t)
-      top_earlgrey_plic_interrupt_for_peripheral[plic_irq_id];
-
-  if (peripheral == kTopEarlgreyPlicPeripheralUart0 &&
+  dt_instance_id_t devid = dt_plic_id_to_instance_id(plic_irq_id);
+  if (devid == dt_uart_instance_id(kDtUart0) &&
```
a-will, line 209: nit: `kDtUart0` is a top-specific name.

sw/device/lib/testing/test_framework/ottf_main.c
```diff
@@ -153,8 +152,7 @@ void _ottf_main(void) {
 
   // Clear reset reason register.
   dif_rstmgr_t rstmgr;
-  CHECK_DIF_OK(dif_rstmgr_init(
-      mmio_region_from_addr(TOP_EARLGREY_RSTMGR_AON_BASE_ADDR), &rstmgr));
+  CHECK_DIF_OK(dif_rstmgr_init_from_dt(kDtRstmgrAon, &rstmgr));
```
a-will, line 155: nit: `kDtRstmgrAon` is a top-specific name, and this could be especially problematic for OTTF integration. Not every integration is going to have a rstmgr with this sort of AON / non-AON power domain split. We might need to think about how init might need to split off an integration layer specific to each top. (normal OS / operating environment stuff)

a-will (approved):
LGTM with nits.

I didn't bother continuing to nit-pick the top-specific names in every file, hehe. But do note that the changes have swapped the explicitly top-specific name parts with implicitly top-specific parts. I don't think you have to handle all of these right now, in this PR, but we should be cognizant that this isn't quite at the goal of removing top-specific identifiers.

engdoreis (approved):

pamaury:
> 
> I didn't bother continuing to nit-pick the top-specific names in every file, hehe. But do note that the changes have swapped the explicitly top-specific name parts with implicitly top-specific parts. I don't think you have to handle all of these right now, in this PR, but we should be cognizant that this isn't quite at the goal of removing top-specific identifiers.

Thanks for the review. You are correct that this PR does not "truly" make those files top-agnostic but the goal of that PR was to keep the semantic. I have added an item in the multitop tracking issue #26116 so that we can revisit this. Some are relatively easy (ie there should only be one rstmgr, so pick index 0), some are tricky (how do you distinguish the main ram from the aon ram?!). I think part of the solution here is to have per-top files that fill the gap to indicate what to do (same problem with linker scripts, and certain other things like `test_rom_start.S`).

sw/device/lib/testing/test_framework/ottf_isrs.c
pamaury: yup...

hw/ip/rom_ctrl/util/scramble_image.py
```diff
@@ -355,15 +440,16 @@ def add_hash(self, scr_mem: MemFile) -> None:
...
     parser.add_argument('outfile', type=argparse.FileType('w'))
     parser.add_argument('hashfile', type=argparse.FileType('w'))
 
     args = parser.parse_args()
-    scrambler = Scrambler.from_hjson_path(args.hjson, ROM_SIZE_WORDS, args.hashfile)
+    scrambler = Scrambler.from_hjson_path(args.hjson, args.mode, args.hashfile)
 
     # Load the input ELF file
-    clr_mem = MemFile.load_elf32(args.infile, 4 * ROM_BASE_WORD)
+    clr_mem = MemFile.load_elf32(args.infile, scrambler.rom_base)
```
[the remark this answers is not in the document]
pamaury: pinging @Razer6 @sameo, the ROM scrambling change was backported from integrated_dev. I agree that we should revisit this if possible to avoid hardcoding.

sw/device/lib/dif/dif_pinmux.c
```diff
@@ -48,11 +48,21 @@ static bool dif_pinmux_get_sleep_status_bit(dif_pinmux_pad_kind_t kind,
   switch (kind) {
     case kDifPinmuxPadKindMio:
       num_pads = PINMUX_PARAM_N_MIO_PADS;
+      // Only platforms with few MIOs, there is a single register with no index.
+#ifdef PINMUX_MIO_PAD_SLEEP_STATUS_0_REG_OFFSET
```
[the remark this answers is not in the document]
pamaury: I'll double check, I definitely ran into this issue in the multitop_dev branch but I didn't check when backporting if the issue still exists.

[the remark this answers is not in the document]
a-will: Ah, looks like I punted: https://github.com/lowRISC/opentitan/issues/2254#issuecomment-1379657502

PR #26040, opened 2025-01-28
[clkmgr] Align behavior of jitter_enable CSR

vogelpi:
This PR aligns RTL, block-level DV, DIF and doc to the modified behavior of the jitter_enable CSR. The corresponding ROM change is #26027.

### DIFF
--- a/hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
+++ b/hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
@@ -409,8 +409,7 @@
           name: "EN",
           resval: "1"
           desc: '''
-            When 1, the value of !!JITTER_ENABLE can be changed.  When 0, writes have no
-            effect.
+            This register has no effect.
           '''
         },
       ]
@@ -429,13 +428,14 @@
           name: "VAL",
           desc: '''
             Enable jittery clock.
-            A value of kMultiBitBool4False disables the jittery clock,
-            while all other values enable jittery clock.
+            At reset, this register reads as kMultiBitBool4False and the jittery clock is disabled.
+            Any write to the register turns the value to kMultiBitBool4True and enables the jittery clock.
+            The value written doesn't matter.
+            The value then remains kMultiBitBool4True until reset.
           ''',
           resval: false
-          // avoid writing random values to this register as it could trigger transient checks
-          // in mubi sync
-          tags: ["excl:CsrAllTests:CsrExclWrite"]
+          // Exclude this register from any automated CSR tests as the behavior is non-standard.
+          tags: ["excl:CsrAllTests:CsrExclAll"]
         }
       ]
     },
--- a/sw/device/lib/dif/dif_clkmgr.c
+++ b/sw/device/lib/dif/dif_clkmgr.c
@@ -88,8 +88,19 @@ dif_result_t dif_clkmgr_jitter_set_enabled(const dif_clkmgr_t *clkmgr,
     default:
       return kDifBadArg;
   }
-  mmio_region_write32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET,
-                      new_jitter_enable_val);
+
+  multi_bit_bool_t clk_jitter_enable_val =
+      mmio_region_read32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET);
+  if (clk_jitter_enable_val == kMultiBitBool4True &&
+      new_jitter_enable_val != kMultiBitBool4True) {
+    return kDifLocked;
+  }
+
+  if (new_jitter_enable_val == kMultiBitBool4True) {
+    mmio_region_write32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET,
+                        new_jitter_enable_val);
+  }
+
   return kDifOk;
 }
 
--- a/sw/device/lib/dif/dif_clkmgr_unittest.cc
+++ b/sw/device/lib/dif/dif_clkmgr_unittest.cc
@@ -29,14 +29,13 @@ class ClkMgrTest : public Test, public MmioTest {
 
 class JitterEnableTest : public ClkMgrTest {};
 
-// SetEnabled uses EXPECT_WRITE32 instead of EXPECT_MASK32 because
-// dif_clkmgr_jitter_set_enabled doesn't perform a read, just a write.
 TEST_F(JitterEnableTest, SetEnabled) {
-  // Disable jitter.
-  EXPECT_WRITE32(CLKMGR_JITTER_ENABLE_REG_OFFSET, kMultiBitBool4False);
+  // Attempt to disable jitter.
+  EXPECT_READ32(CLKMGR_JITTER_ENABLE_REG_OFFSET, kMultiBitBool4False);
   EXPECT_DIF_OK(dif_clkmgr_jitter_set_enabled(&clkmgr_, kDifToggleDisabled));
 
   // Enable jitter.
+  EXPECT_READ32(CLKMGR_JITTER_ENABLE_REG_OFFSET, kMultiBitBool4False);
   EXPECT_WRITE32(CLKMGR_JITTER_ENABLE_REG_OFFSET, kMultiBitBool4True);
   EXPECT_DIF_OK(dif_clkmgr_jitter_set_enabled(&clkmgr_, kDifToggleEnabled));
 }
--- a/util/reggen/fpv_csr.sv.tpl
+++ b/util/reggen/fpv_csr.sv.tpl
@@ -86,6 +86,14 @@ module ${mod_base}_csr_assert_fpv import tlul_pkg::*;
   // accesses).
   bit [${hro_idx_width-1}:0] hro_idx;
   bit [${addr_msb}:0] normalized_addr;
+  % if lblock == "clkmgr":
+    % for hro_reg in hro_regs_list:
+      % if hro_reg.name.lower() == "jitter_enable":
+  // The jitter_enable CSR needs some special handling below.
+  localparam bit [${hro_idx_width-1}:0] JitterEnableHroIdx = ${hro_map.get(hro_reg.offset)[0]};
+      % endif
+    % endfor
+  % endif
 
   `ifdef FPV_ON
     // For FPV, we restrict the pend_trans array by giving its size a smaller upper bound. This
@@ -191,7 +199,14 @@ module ${mod_base}_csr_assert_fpv import tlul_pkg::*;
       if (d2h.d_valid) begin
         if (pend_trans[d_source_idx].wr_pending == 1) begin
           if (!d2h.d_error && regwen[hro_idx]) begin
+            % if lblock == "clkmgr":
+            if (hro_idx == JitterEnableHroIdx) begin
+              // Any write to the jitter_enable CSR writes MuBi4True.
+              exp_vals[hro_idx] <= prim_mubi_pkg::MuBi4True;
+            end else if (access_policy[hro_idx] == FpvRw0c) begin
+            % else:
             if (access_policy[hro_idx] == FpvRw0c) begin
+            % endif
               // Assume FpvWr0c policy only has one field that is wr0c.
               exp_vals[hro_idx] <= exp_vals[hro_idx][0] == 0 ? 0 : pend_trans[d_source_idx].wr_data;
             end else begin
--- a/util/reggen/reg_top.sv.tpl
+++ b/util/reggen/reg_top.sv.tpl
@@ -824,6 +824,13 @@ ${rdata_gen(f, r.name.lower() + "_" + f.name.lower())}\
 % endif
 
   // Unused signal tieoff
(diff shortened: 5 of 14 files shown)


hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
```diff
@@ -429,13 +428,14 @@
...
+            Any write to the register turns the value to kMultiBitBool4True and enables the jittery clock.
+            The value written doesn't matter.
+            The value then remains kMultiBitBool4True until reset.
           ''',
           resval: false
-          // avoid writing random values to this register as it could trigger transient checks
-          // in mubi sync
-          tags: ["excl:CsrAllTests:CsrExclWrite"]
+          // Exclude this register from any automated CSR tests as the behavior is non-standard.
+          tags: ["excl:CsrAllTests:CsrExclAll"]
```
rswarbrick, line 438: Not a big deal, but is that exclusion change necessary? Which CSR test was failing?

sw/device/lib/dif/dif_clkmgr_unittest.cc
```diff
@@ -33,10 +33,11 @@ class JitterEnableTest : public ClkMgrTest {};
 // dif_clkmgr_jitter_set_enabled doesn't perform a read, just a write.
 TEST_F(JitterEnableTest, SetEnabled) {
   // Disable jitter.
-  EXPECT_WRITE32(CLKMGR_JITTER_ENABLE_REG_OFFSET, kMultiBitBool4False);
+  EXPECT_READ32(CLKMGR_JITTER_ENABLE_REG_OFFSET, kMultiBitBool4False);
   EXPECT_DIF_OK(dif_clkmgr_jitter_set_enabled(&clkmgr_, kDifToggleDisabled));
```
rswarbrick, line 37: This looks a bit strange to me: as far as I know, it's no longer possible to disable jitter after it has been enabled. Do we just need to update the comment on line 35?

vogelpi: Thanks for your feedback @rswarbrick ! This here is just the DIF unittest. It doesn't take into account the behavior of the hardware (the jitter not being disablable once enabled). What this checks is whether the DIF functions when called with different arguments generate the correct number of reads and writes. And the test does two calls to the corresponding DIF function: once to enable the jitter (a read and a write) and once to disable it (one read only).

hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
vogelpi: It turns out this is really required. Without the change, all CSR tests except for the csr_hw_reset test fail.

util/reggen/fpv_csr.sv.tpl
```diff
@@ -86,6 +86,14 @@ module ${mod_base}_csr_assert_fpv import tlul_pkg::*;
   // accesses).
   bit [${hro_idx_width-1}:0] hro_idx;
   bit [${addr_msb}:0] normalized_addr;
+  % if lblock == "clkmgr":
+    % for hro_reg in hro_regs_list:
+      % if hro_reg.name.lower() == "jitter_enable":
```
andreaskurth, line 91: Just a note here that while having code specific to a block and register hard-coded into reggen isn't a scalable or future-proof solution, I think it's totally appropriate for implementing this ECO on the `earlgrey_1.0.0` branch.

andreaskurth (approved):
LGTM as an ECO. Thanks @vogelpi!

andreaskurth:
CHANGE AUTHORIZED: hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/data/clkmgr.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr_reg_top.sv

Approved ECO

moidx:
CHANGE AUTHORIZED: hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/data/clkmgr.hjson
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/clkmgr/rtl/clkmgr_reg_top.sv

Approved ECO

moidx (approved):

hw/ip_templates/clkmgr/data/clkmgr.hjson.tpl
matutem: This CSR is declared RW for software, so automated tests will expect after a write you will read back what you wrote. I expect the cip_lib csr_rw family of tests to fail.

sw/device/lib/dif/dif_clkmgr.c
```diff
@@ -88,8 +88,19 @@ dif_result_t dif_clkmgr_jitter_set_enabled(const dif_clkmgr_t *clkmgr,
     default:
       return kDifBadArg;
   }
-  mmio_region_write32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET,
-                      new_jitter_enable_val);
+
```
matutem, line 91: I don't quite agree with this behavior: I would rather model it as
  new_jitter_enable_val = kMultiBitBool4True;
  mmio_region_write32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET,
                       new_jitter_enable_val);

Also the writes do occur, though not with the expected value, so I would not return kDifLocked. The real issue is that this is not a true RW CSR.

sw/device/lib/dif/dif_clkmgr_unittest.cc
matutem: You are right, the comment is misleading.

util/reggen/fpv_csr.sv.tpl
```diff
@@ -191,7 +199,14 @@ module ${mod_base}_csr_assert_fpv import tlul_pkg::*;
       if (d2h.d_valid) begin
         if (pend_trans[d_source_idx].wr_pending == 1) begin
           if (!d2h.d_error && regwen[hro_idx]) begin
+            % if lblock == "clkmgr":
```
matutem, line 202: Thanks so much for updating this: I would not have caught it, mostly because I can't run these tests.

matutem: By the way, do we check regwens in FPV?

util/reggen/reg_top.sv.tpl
```diff
@@ -904,13 +911,16 @@ ${bits.msb}\
...
 
       # when async, pick from the cdc handled data
-      wd_expr = f'{finst_name}_wd'
-      if reg.async_clk:
-        if field.bits.msb == field.bits.lsb:
-          bit_sel = f'{field.bits.msb}'
-        else:
-          bit_sel = f'{field.bits.msb}:{field.bits.lsb}'
-        wd_expr = f'{clk_base_name}{reg_name}_wdata[{bit_sel}]'
+      if lblock == "clkmgr" and reg_name == "jitter_enable":
```
matutem, line 914: These changes in reg_top.sv.tpl are ugly, but luckily they'll just stay in this branch.

sw/device/lib/dif/dif_clkmgr.c
```diff
@@ -88,8 +88,19 @@ dif_result_t dif_clkmgr_jitter_set_enabled(const dif_clkmgr_t *clkmgr,
...
+  multi_bit_bool_t clk_jitter_enable_val =
+      mmio_region_read32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET);
+  if (clk_jitter_enable_val == kMultiBitBool4True &&
+      new_jitter_enable_val == kMultiBitBool4False) {
```
matutem, line 95:
```suggestion
      new_jitter_enable_val != kMultiBitBool4True) {
```

matutem (approved):
LGTM

matutem:
Please consider the suggested changes.

vogelpi:
Thanks for your reviews everybody! I am happy to include the suggested changes later tonight. (I am currently traveling.)

util/reggen/fpv_csr.sv.tpl
vogelpi: The funny thing about this template is that it's used for generating `clkmgr_csr_assert_fpv.sv` which is also used during regular block-level DV (that's how I found out about the need for changing this file). Whether we check the regwens in FPV I don't know. TBH, I don't think we have an FPV setup for clkmgr, do we?

sw/device/lib/dif/dif_clkmgr.c
vogelpi: This will change with the next force push.

sw/device/lib/dif/dif_clkmgr_unittest.cc
vogelpi: Ah, now I understand. Sorry @rswarbrick , I've misread your comment earlier. I'll update this with the next push.

util/reggen/fpv_csr.sv.tpl
matutem: In any case the test would need to be created, since there is no way to infer which CSR a regwen is meant to affect. No magic bullets for these besides careful review.

sw/device/lib/dif/dif_clkmgr_unittest.cc
```diff
@@ -33,10 +33,11 @@ class JitterEnableTest : public ClkMgrTest {};
...
 
   // Enable jitter.
```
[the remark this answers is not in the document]
vogelpi: With the present implementation of `dif_clkmgr_jitter_set_enabled()` we cannot test this. Because the function checks the argument: if it doesn't match `kMultiBitBool4False` or `kMultiBitBool4True`, i.e., if it's an invalid MuBi value, then the function returns `kDifBadArg`. This part of the function did not change with this PR.

I am open to change this, too. But TBH, I don't think it really matters for the ECO. We should try to land this asap and create the tag. There will be time to make further updates to the software / DIF and the documentation afterwards.

vogelpi:
Update: I've now pushed those changes proposed by @matutem  and @rswarbrick which I was able to implement with reasonable effort. We should try to land this asap and I intend to merge this as soon as CI passes.

sw/device/lib/dif/dif_clkmgr.c
```diff
@@ -88,8 +88,19 @@ dif_result_t dif_clkmgr_jitter_set_enabled(const dif_clkmgr_t *clkmgr,
...
-                      new_jitter_enable_val);
+
+  multi_bit_bool_t clk_jitter_enable_val =
+      mmio_region_read32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET);
+  if (clk_jitter_enable_val == kMultiBitBool4True &&
+      new_jitter_enable_val != kMultiBitBool4True) {
+    return kDifLocked;
+  }
+
+  if (new_jitter_enable_val == kMultiBitBool4True) {
```
matutem, line 99: This "if" is incorrect: even if we try to write False it will write True. I propose you remove the If, and replace `new_jitter_enable_val` below by `kMultiBitBool4True`.

sw/device/lib/dif/dif_clkmgr.c
```diff
@@ -88,8 +88,19 @@ dif_result_t dif_clkmgr_jitter_set_enabled(const dif_clkmgr_t *clkmgr,
     default:
       return kDifBadArg;
   }
-  mmio_region_write32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET,
-                      new_jitter_enable_val);
+
+  multi_bit_bool_t clk_jitter_enable_val =
+      mmio_region_read32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET);
+  if (clk_jitter_enable_val == kMultiBitBool4True &&
+      new_jitter_enable_val != kMultiBitBool4True) {
```
matutem, line 95: This could simply be `new_state != kDifToggleEnabked`, and changing the write below to always be `kMultiBitBool4True` you realize `new_jitter_enable_val` is unused.

sw/device/lib/dif/dif_clkmgr.c
vogelpi: Sorry @matutem , I merged the PR before seeing this. Can we fix this in the follow-up containing also the new software test, please?

sw/device/lib/dif/dif_clkmgr_unittest.cc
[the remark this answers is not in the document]
matutem: That is an artifact of the dif code, and we could write bypassing dif, but I agree this is low priority.

sw/device/lib/dif/dif_clkmgr.c
matutem: I made the change in my test PR: https://github.com/lowRISC/opentitan/pull/26058

vogelpi: Thanks @matutem !

PR #26042, opened 2025-01-28
[RACL] Render RACL Roles in top_racl_pkg

davidschrammel:
This commit adds the racl roles as paramters to the top_racl_pkg.
These will be used later in a follow-up PR.

Example output:
```
  /**
   * RACL Roles
   */
  parameter role_t RACL_ROLE_ROT   = 4'h0;
  parameter role_t RACL_ROLE_ROLE1 = 4'h1;
  parameter role_t RACL_ROLE_SOC   = 4'h2;
```

### DIFF
--- a/util/topgen/templates/toplevel_racl_pkg.sv.tpl
+++ b/util/topgen/templates/toplevel_racl_pkg.sv.tpl
@@ -79,6 +79,16 @@ package top_racl_pkg;
   endfunction
 % endif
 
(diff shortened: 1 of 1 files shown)


rswarbrick:
Can we capitalise the names? I think that OT currently uses ALL_CAPS for the names of preprocessor macros?

rswarbrick (approved):

PR #26044, opened 2025-01-28
[RACL] Add RACL support for spi_device

davidschrammel:
This PR enables RACL support for `spi_device`.

~~The issues raised in #26004 also affect generated files here.
They are fixed in #26056 and I will rebase this PR on top assuming #26056 will be merged to master before this PR.~~


util/raclgen.py
```diff
@@ -55,6 +55,7 @@ def main():
 
     template = """
 <%doc>
+  Note: The paramter names must be generated identically in toplevel.sv.tpl
```
rswarbrick, line 58: Mis-spelling of "parameter". (While you're fixing it, can you change the commit title too?)

rswarbrick, line 58: Have we now got three copies (`toplevel.sv.tpl`, `raclgen.py` and `toplevel_racl_pkg.sv.tpl`)? Doesn't the comment need to mention them all?

hw/ip/spi_device/rtl/spi_device.sv
```diff
@@ -12,8 +12,13 @@ module spi_device
   import spi_device_reg_pkg::SPI_DEVICE_EGRESS_BUFFER_IDX;
   import spi_device_reg_pkg::SPI_DEVICE_INGRESS_BUFFER_IDX;
 #(
-  parameter logic [NumAlerts-1:0] AlertAsyncOn = {NumAlerts{1'b1}},
-  parameter spi_device_pkg::sram_type_e SramType = spi_device_pkg::DefaultSramType
+  parameter logic [NumAlerts-1:0] AlertAsyncOn         = {NumAlerts{1'b1}},
+  parameter spi_device_pkg::sram_type_e SramType       = spi_device_pkg::DefaultSramType,
+  parameter bit          EnableRacl                    = 1'b0,
+  parameter bit          RaclErrorRsp                  = 1'b1,
+  parameter int unsigned RaclPolicySelVec[73]          = '{73{0}},
```
rswarbrick, line 19: Not a comment about this PR, but I suddenly noticed that it's a bit awkward because we have to reflect the number of registers in the parameter type here.

I wonder whether there's a way this could be more automatic, or maybe some associative array (assuming those are allowed for parameters) keyed by the register name that selects policy zero if a register isn't in the array.

Clearly for a follow-up, but I'm interested in your opinion here.

util/raclgen.py
davidschrammel: I think the `toplevel.sv.tpl` exists two (or three) times (for darjeeling, earlgrey, englishbreakfast), so we might have up to 5 copies at the moment.
Yes, we will have to deduplicate this at some point.

hw/ip/spi_device/rtl/spi_device.sv
davidschrammel: That is a good point. I will try to look into this and see what's possible.

Razer6: Unfortunately, there is no constant provided by the corresponding `reg_pkg`. I guess it would be a fair change to let this package render a constant describing the number of registers?

davidschrammel:
@rswarbrick I have unified the notes across all template copies.

However, before merging I would still like to rebase this PR on top of #26056 so I can include those fixes as well.

util/raclgen.py
rswarbrick: Oof! Thanks for doing this though.

hw/ip/spi_device/rtl/spi_device.sv
rswarbrick: I think I agree. I've just opened #26074 to track it (but, honestly, I think it should be pretty easy to do!)

rswarbrick:
CHANGE AUTHORIZED: hw/ip/spi_device/data/spi_device.hjson
CHANGE AUTHORIZED: hw/ip/spi_device/rtl/spi_device.sv
CHANGE AUTHORIZED: hw/ip/spi_device/rtl/spi_device_reg_top.sv

The change to `spi_device.hjson` just adds the new ports for RACL signals. The change to `spi_device.sv` looks nontrivial, but that's just because it's wiring together various `racl_error` signals (which will all be both zero and unused in the earlgrey context. The change to the register top is just wiring in RACL support, which will have no effect if RACL is disabled. Finally, the change to `top_earlgrey.sv` is just stubbing the newly added ports.

In all, there is no significant risk to earlgrey from this change.

rswarbrick (approved):
Thanks for the work here. I'm happy that this is ready to come in!

vogelpi:
CHANGE AUTHORIZED: hw/ip/spi_device/data/spi_device.hjson
CHANGE AUTHORIZED: hw/ip/spi_device/rtl/spi_device.sv
CHANGE AUTHORIZED: hw/ip/spi_device/rtl/spi_device_reg_top.sv

As per Rupert's assessment above.

vogelpi (approved):

PR #26046, opened 2025-01-28
[ci] Rename the ROM_EXT hyper310 workflow

cfrantz:
This is a manual cherry-pick of #26012.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -440,7 +440,7 @@ jobs:
       timeout: 90
 
   execute_rom_ext_fpga_tests_cw310:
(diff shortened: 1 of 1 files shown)


a-will (approved):

nbdd0121:
I don't think this should be rename don master, we don't have other jobs named Hyper310, and there's no conflicts.

cfrantz:
> I don't think this should be rename don master, we don't have other jobs named Hyper310, and there's no conflicts.

I was going for consistency of user-visible names between `earlgrey_1.0.0` and `master`.  If you feel strongly about it, I can send revert PR.

PR #26054, opened 2025-01-29
[englishbreakfast] Add bazel targets for verilator and cw305

a-will:
Add bazel targets for verilator and cw305, and use them in CI.

Also, use an empty ROM for the CW305 bitstream, and require splicing for the bitstream to be usable. This removes the dependency arrow from hardware to software.


hw/ip/prim_xilinx/rtl/prim_xilinx_rom.sv
```diff
@@ -0,0 +1,68 @@
...
+  input  logic             req_i,
+  input  logic [Aw-1:0]    addr_i,
+  output logic [Width-1:0] rdata_o,
+  input rom_cfg_t          cfg_i
+);
+
+  logic unused_signals;
+  assign unused_signals = ^{cfg_i, rst_ni};
+
+  if (MemInitFile != "") begin : gen_generic
```
a-will, line 25: Why not use the xpm always? The macro rejects our .vmem files, and I was too lazy to figure out all of what was wrong and fix that up.

hw/BUILD
```diff
@@ -32,19 +32,33 @@ string_list_flag(
...
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "dv_common_files",
+    srcs = glob(["dv/sv/**"]),
+    visibility = ["//visibility:public"],
+)
+
 fusesoc_build(
```
pamaury, line 53: Should this be eventually moved to `hw/top_earlgrey/` like EB?

pamaury (approved):
I can't comment on the verilog change but the bazel looks good, thanks @a-will

hw/BUILD
a-will: Yes. I had started on that, then realized it was going to be a more significant upheaval and decided not to have this PR become too large.

hw/ip/prim_xilinx/rtl/prim_xilinx_rom.sv
```diff
@@ -0,0 +1,68 @@
...
+    );
+  end
+
+
+  ////////////////
+  // ASSERTIONS //
+  ////////////////
+
+  // Control Signals should never be X
+  `ASSERT(noXOnCsI, !$isunknown(req_i), clk_i, '0)
```
vogelpi, line 67: Nit: I think we usually put a `_A` suffix for SVAs and require camel case for the Name, i.e., `no` -> `No`.

vogelpi (approved):
This looks good to me, thanks @a-will  for taking care of this.

Question: I see the bitstream and Verilator builds are still in CI (so this works in principle) but we currently don't run any software on Englishbreakfast right? So it might be that we'll have to do some debugging related to this PR when setting up software again?

a-will:
> This looks good to me, thanks @a-will for taking care of this.
> 
> Question: I see the bitstream and Verilator builds are still in CI (so this works in principle) but we currently don't run any software on Englishbreakfast right? So it might be that we'll have to do some debugging related to this PR when setting up software again?

Yes, that may be true. In fact, the tests won't actually run on CW305 yet, since there is no mechanism or target to splice the ROM in. Since I can't build the ROM right now, I punted that part.

Our bitstream_splice rule might be a start, though it might have some hard-coding for earlgrey right now...

hw/ip/prim_xilinx/rtl/prim_xilinx_rom.sv
a-will: This was copied directly from prim_generic_rom.sv. Shall I change it in both places?

vogelpi: Don't worry about it then! Let's move forward with this :-)

PR #26056, opened 2025-01-29
[racl] Follow-up fixes to TLUL adapter and reg_top

Razer6:
This PR is a follow up to the review in https://github.com/lowRISC/opentitan/pull/26004#pullrequestreview-2579395912:

It adds:
1. Proper qualification of the read and write request with the necessary TLUL signals for the reg and sram adapter
2. Properly qualify the error log computation in the reg_top with the access control signal coming from the reg_adapter.

### DIFF
--- a/hw/ip/i2c/rtl/i2c.sv
+++ b/hw/ip/i2c/rtl/i2c.sv
@@ -12,7 +12,7 @@ module i2c
   parameter logic [NumAlerts-1:0] AlertAsyncOn         = {NumAlerts{1'b1}},
   parameter int unsigned          InputDelayCycles     = 0,
   parameter bit                   EnableRacl           = 1'b0,
-  parameter bit                   RaclErrorRsp         = 1'b1,
+  parameter bit                   RaclErrorRsp         = EnableRacl,
   parameter int unsigned          RaclPolicySelVec[32] = '{32{0}}
 ) (
   input                                    clk_i,
--- a/util/reggen/reg_top.sv.tpl
+++ b/util/reggen/reg_top.sv.tpl
@@ -730,9 +730,9 @@ ${finst_gen(sr, field, finst_name, fsig_name, fidx)}
 
   assign addrmiss = (reg_re || reg_we) ? ~|addr_hit : 1'b0 ;
 % if racl_support:
-  // A valid address hit but failed the RACL check
-  assign racl_error_o = tl_i.a_valid &
-                        (|addr_hit) & ~(|(addr_hit & (racl_addr_hit_read | racl_addr_hit_write)));
+  // A valid address hit, access, but failed the RACL check
+  assign racl_error_o = |addr_hit & ((reg_re & ~|racl_addr_hit_read) |
+                                     (reg_we & ~|racl_addr_hit_write));
   assign racl_error_log_o.racl_role  = racl_role;
 
   if (EnableRacl) begin : gen_racl_log
--- a/hw/ip/i2c/rtl/i2c_reg_top.sv
+++ b/hw/ip/i2c/rtl/i2c_reg_top.sv
@@ -3471,9 +3471,9 @@ module i2c_reg_top
   end
 
   assign addrmiss = (reg_re || reg_we) ? ~|addr_hit : 1'b0 ;
-  // A valid address hit but failed the RACL check
-  assign racl_error_o = tl_i.a_valid &
-                        (|addr_hit) & ~(|(addr_hit & (racl_addr_hit_read | racl_addr_hit_write)));
+  // A valid address hit, access, but failed the RACL check
+  assign racl_error_o = |addr_hit & ((reg_re & ~|racl_addr_hit_read) |
+                                     (reg_we & ~|racl_addr_hit_write));
   assign racl_error_log_o.racl_role  = racl_role;
 
   if (EnableRacl) begin : gen_racl_log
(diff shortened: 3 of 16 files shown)


util/reggen/reg_top.sv.tpl
```diff
@@ -731,8 +731,8 @@ ${finst_gen(sr, field, finst_name, fsig_name, fidx)}
   assign addrmiss = (reg_re || reg_we) ? ~|addr_hit : 1'b0 ;
 % if racl_support:
   // A valid address hit but failed the RACL check
-  assign racl_error_o = tl_i.a_valid &
-                        (|addr_hit) & ~(|(addr_hit & (racl_addr_hit_read | racl_addr_hit_write)));
+  assign racl_error_o = ~((reg_re & (|racl_addr_hit_read)) |
+                          (reg_we & (|racl_addr_hit_write)));
```
rswarbrick, line 735: This is a nice clean-up. I think I can see why it's now correct, and it's also simpler to read. Nice!

But please could you leave a note in the commit message that explains what's going on? Commit messages aren't really optional.

hw/ip/i2c/rtl/i2c.sv
```diff
@@ -12,7 +12,7 @@ module i2c
   parameter logic [NumAlerts-1:0] AlertAsyncOn         = {NumAlerts{1'b1}},
   parameter int unsigned          InputDelayCycles     = 0,
   parameter bit                   EnableRacl           = 1'b0,
-  parameter bit                   RaclErrorRsp         = 1'b1,
+  parameter bit                   RaclErrorRsp         = EnableRacl,
```
rswarbrick, line 15: I'm convinced that this is cleaner code, so it looks like the right thing to do. But does it actually change behaviour? I think the parameter is only used to qualify `racl_error_o`. This, in turn, is only true if we hit a register that doesn't have a matching racl permission and every register gets a permission if `EnableRacl` is false.

Is that correct? Probably worth explaining in the commit message why the change is needed (either cleaning stuff up or fixing some bug that Rupert didn't see!)

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
rswarbrick, line 1: Isn't this code from PR #26006? Is it needed in this PR? If so, this one is going to have to wait behind that one.

Razer6: Argh. Too many branches, open changes, and what not. I drop that.

hw/ip/i2c/rtl/i2c.sv
Razer6: No there is no change in behaviour. Is proposed by Adrian, this is jus a safeguard to not send out a spurious racl error if RACL is disabled. Nothing else.

util/reggen/reg_top.sv.tpl
Razer6: I changed the error message

hw/ip/i2c/rtl/i2c.sv
Razer6: I included that in the commit message.

hw/dv/sv/mem_bkdr_util/mem_bkdr_util_row_adapter.sv
Razer6: Removed.

util/reggen/reg_top.sv.tpl
```diff
@@ -731,8 +731,8 @@ ${finst_gen(sr, field, finst_name, fsig_name, fidx)}
   assign addrmiss = (reg_re || reg_we) ? ~|addr_hit : 1'b0 ;
 % if racl_support:
   // A valid address hit but failed the RACL check
-  assign racl_error_o = tl_i.a_valid &
-                        (|addr_hit) & ~(|(addr_hit & (racl_addr_hit_read | racl_addr_hit_write)));
```
alees24, line 735: This will error in the event that it is not a permitted read and not a permitted write, i.e. an error will also be raised when there is no access at all.
I think you're aiming for:
```
assign racl_error_o = |addr_hit & ((reg_re & ~|racl_addr_hit_read) | (reg_we & ~|racl_addr_hit_write));
```

hw/ip/i2c/rtl/i2c.sv
alees24: @rswarbrick _If the racl_error_o generation logic is correct,_ such that RACL errors cannot be raised spuriously, then it has no functional impact on non-RACL designs. It's belt'n'braces.

alees24 (changes requested):
Thanks for fixing up the rd_req/wr_req logic in the two adapters. Just the templated (+autogen) 'racl_error_o' logic that needs fixing now, I think.

util/reggen/reg_top.sv.tpl
Razer6: Right. Thanks for the help. I just rebased and updated.

alees24 (approved):
Thanks, LGTM.

alees24:
CHANGE AUTHORIZED: hw/ip/mbx/rtl/mbx.sv
CHANGE AUTHORIZED: hw/ip_templates/ac_range_check/rtl/ac_range_check.sv.tpl
CHANGE AUTHORIZED: hw/ip_templates/alert_handler/rtl/alert_handler.sv.tpl
CHANGE AUTHORIZED: hw/ip_templates/rv_plic/rtl/rv_plic.sv.tpl

alees24:
CI failures unrelated.

PR #26057, opened 2025-01-29
[pwm] Fix flakiness on smoketest

engdoreis:
Make the host to wait for the ottf to print the "running" message before start the test.

### DIFF
--- a/sw/host/tests/chip/pwm_smoketest/src/main.rs
+++ b/sw/host/tests/chip/pwm_smoketest/src/main.rs
@@ -13,6 +13,7 @@ use opentitanlib::io::gpio::GpioPin;
 use opentitanlib::io::gpio::PinMode;
 use opentitanlib::test_utils;
 use opentitanlib::test_utils::init::InitializeTest;
+use opentitanlib::uart::console::UartConsole;
 use std::borrow::Borrow;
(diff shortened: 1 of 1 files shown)


pamaury (approved):

sw/host/tests/chip/pwm_smoketest/src/main.rs
```diff
@@ -109,6 +110,10 @@ fn main() -> Result<()> {
     uart.clear_rx_buffer()?;
     transport.pin_strapping("RESET")?.remove()?;
 
+    // Wait until test is running.
+    let uart = transport.uart("console")?;
+    UartConsole::wait_for(&*uart, r"Running [^\r\n]*", opts.timeout)?;
+
```
luismarques, line 116: Since the PWM smoketest doesn't do too much between OTTF starting the test and that actual PWM testing logic, this seems like it should sufficiently synchronize the device and the host. An even more precise sync would have the device send a UART message (or toggle a pin, etc) when it's actually ready to start the PWM testing part, but that's probably overkill.

Arguably this could be moved to like 123.

luismarques (approved):

engdoreis:
The test failed in CI is not related to this PR.

PR #26058, opened 2025-01-29
[tests,clkmgr_jitter] Fix jitter tests to track RTL changes

matutem:
Fix clkmgr jitter enable tests to match the change that any write to jitter enable will set it to mubi-true.

### DIFF
--- a/hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
+++ b/hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
@@ -188,8 +188,13 @@
             jitter_o clkmgr output to toggle. Verify this output is connected to AST's
             clk_src_sys_jen_i input using formal.
 
+            Test the following properties of jitter enable:
+            - The reset value of jitter_enable is off.
+            - Any write to jitter_enable turns jitter on.
+            - jitter_regwen has no impact on jitter_enable.
+
             X-ref with various specific jitter enable tests.
-            SiVal: This is a connectivity test only, not suitable for sival.
+            SiVal: CPU must be enabled, but no other OTP or lifecycle dependencies.
             '''
       stage: V2
       si_stage: NA
@@ -211,7 +216,14 @@
       name: chip_sw_clkmgr_jitter_cycle_measurements
       desc: '''Verify jitter via clock cycle measurements after calibration.
             The clock count thresholds for main clk need to be set wider than when jitter is disabled.
-            SiVal: This is only useful on real silicon.
+            Check when jitter is enabled:
+            - Using jitter thresholds no errors are detected.
+            - Using normal thresholds results in recoverable errors.
+            - Don't test normal thresholds for FPGAs since they don't support jitter and no errors will occur.
+            Check when jitter is disabled:
+            - Using either sets of thresholds result in no errors.
+
+            SiVal: This is only useful on real silicon since FPGAs don't support jitter.
             Should be done after clock calibration.
             '''
       stage: V3
--- a/sw/device/lib/dif/dif_clkmgr.c
+++ b/sw/device/lib/dif/dif_clkmgr.c
@@ -42,6 +42,35 @@ static bool clkmgr_measure_ctrl_regwen(const dif_clkmgr_t *clkmgr) {
                              CLKMGR_MEASURE_CTRL_REGWEN_EN_BIT);
 }
 
+/**
+ * Checks if the jitter enable register is locked.
+ *
+ * Jitter enable register is locked by CLKMGR_JITTER_REGWEN.
+ */
+OT_WARN_UNUSED_RESULT
+static bool jitter_enable_register_is_locked(const dif_clkmgr_t *clkmgr) {
+  // Jitter enable register is locked when `CLKMGR_JITTER_REGWEN_EN_BIT` bit
+  // is 0.
+  return !bitfield_bit32_read(
+      mmio_region_read32(clkmgr->base_addr, CLKMGR_JITTER_REGWEN_REG_OFFSET),
+      CLKMGR_JITTER_REGWEN_EN_BIT);
+}
+
+/**
+ * Checks if the external clock control register is locked.
+ *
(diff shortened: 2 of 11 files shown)


sw/device/tests/clkmgr_jitter_frequency_test.c
```diff
@@ -36,6 +49,47 @@ enum {
...
+    // Set thresholds for jitter disabled expecting failures.
+    CHECK_STATUS_OK(
+        clkmgr_testutils_enable_clock_counts_with_expected_thresholds(
+            &clkmgr, /*jitter_enabled=*/false, /*external_clk=*/false,
+            /*low_speed=*/false));
+    busy_spin_micros(delay_micros);
+    dif_clkmgr_recov_err_codes_t err_codes;
+    CHECK_DIF_OK(dif_clkmgr_recov_err_code_get_codes(&clkmgr, &err_codes));
+    CHECK(err_codes != 0);
+  }
```
vogelpi, line 73: Should we add a print here for  FPGA platforms to highlight that this part of the test is skipped?

sw/device/tests/clkmgr_jitter_test.c
```diff
@@ -16,19 +16,30 @@ bool test_main(void) {
...
-    dif_toggle_t actual_state = state;
-    CHECK_DIF_OK(dif_clkmgr_jitter_set_enabled(&clkmgr, expected_state));
-    CHECK_DIF_OK(dif_clkmgr_jitter_get_enabled(&clkmgr, &actual_state));
-    CHECK(actual_state == expected_state);
-  }
+  CHECK(state == kDifToggleDisabled);
+  // Check that any write will enable it.
+  // A write to disable it will actually enable it.
+  dif_toggle_t actual_state = kDifToggleDisabled;
+  CHECK_DIF_OK(dif_clkmgr_jitter_set_enabled(&clkmgr, kDifToggleDisabled));
```
vogelpi, line 28: According to the test plan we should also test that the regwen actually doens't have an impact. Are these missing or did I overlook the calls to the newly defined functions for this?

vogelpi (commented):
Thanks @matutem for doing this!

This looks mostly good, I have a question regarding the behavior of the `set_enabled()` function and I think the testing of the regwen is missing from the jitter test despite the testplan saying it do this.

sw/device/tests/clkmgr_jitter_frequency_test.c
matutem: Done.

sw/device/tests/clkmgr_jitter_test.c
matutem: Good spotting: it was missing. Done.

hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
```diff
@@ -188,8 +188,13 @@
             jitter_o clkmgr output to toggle. Verify this output is connected to AST's
             clk_src_sys_jen_i input using formal.
 
+            Test the following properties of jitter enable:
+            - The reset value of jitter_enable is off.
+            - Any write to jitter_enable turns jitter on.
```
vogelpi, line 193: Nit: the test currently only writes the enable value and I think that's fine.

hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
```diff
@@ -188,8 +188,13 @@
...
+            - jitter_regwen has no impact on jitter_enable.
```
vogelpi, line 194: I think testing that the regwen has no impact is fine but what we really care about is that we cannot disable it again once it's set to enable and this is currently not done. I think it would be valuable to add this.

sw/device/lib/dif/dif_clkmgr.c
```diff
@@ -66,41 +114,26 @@ dif_result_t dif_clkmgr_jitter_get_enabled(const dif_clkmgr_t *clkmgr,
...
   }
-
-  if (new_jitter_enable_val == kMultiBitBool4True) {
-    mmio_region_write32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET,
-                        new_jitter_enable_val);
-  }
-
+  // Any write sets jitter enable to kMultiBitBool4True.
+  mmio_region_write32(clkmgr->base_addr, CLKMGR_JITTER_ENABLE_REG_OFFSET,
+                      kMultiBitBool4True);
```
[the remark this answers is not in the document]
vogelpi: Thanks for changing this @matutem !

hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
vogelpi: Ah, actually we cannot test this anymore as the dif in it's reworked form doesn't allow this. This should be alright then or what is your view @matutem ?

vogelpi (approved):
This looks good but I have one remaining question related to testing the disable of the jitter enable. Thanks @matutem for implementing the feedback!

moidx (approved):

hw/top_earlgrey/data/ip/chip_clkmgr_testplan.hjson
matutem: I added a test for this in clkmgr_jitter_test.c: first clear regwen and then set jitter enable and check that updates to enable go through. The dif code doesn't matter for this test.

matutem: By the way, I think jitter regwen getting stuck once cleared should be a unit test. There is a clkmgr_regwen_vseq.sv test that should have done this, but it only checks extclk_regwen and meas_ctrl_regwen, darn! I am pretty bummed about this bug!

I will add this check in that unit test in a separate PR. I think we can commit this separately, since it enables Nuvoton to check the jitter_enable fix at top-level.

matutem: A units test for jitter_regwen was added in #26259.

Issue #23980, opened 2024-07-10
[hmac,dv] Avoid reseeding for HMAC test vectors tests


gdessouky:
I'm also not sure we need to reseed the test vectors tests at all actually. It seems we run all the test vectors anyway for any iteration; we are not currently randomly picking some of the test vectors for every seed and I don't see any other randomization actually happening. What do we gain from reseeding? Can't we just run 1 seed? 
@martin-velay @rswarbrick

_Originally posted by @gdessouky in https://github.com/lowRISC/opentitan/issues/23978#issuecomment-2219947032_

martin-velay:
See this issue https://github.com/lowRISC/opentitan/issues/24303, as the way done for KMAC seems to be a good option to avoid too long tests

martin-velay:
Could be moved to M7 backlog as this won't give us more confidence

martin-velay:
Doesn't required for D3-V3 signoffs as it's only to reduce simulation time

Issue #23288, opened 2024-05-23
[hmac/dv] Remove/replace checks from the sequences


martin-velay:
### Description

The sequences are not supposed to perform checks, that's the job of the scoreboard. This constraint comes from the UVM itself, we should build reusable components, the SB can be disabled but not the seq.

Eg: `compare_digest(parsed_vectors[j].exp_digest);` from `hmac_test_vectors_sha_vseq`

martin-velay:
One part of this issue has been addressed by this PR https://github.com/lowRISC/opentitan/pull/26059, but it cannot be closed yet as some other checks are still performed in the vseq

PR #26059, opened 2025-01-29
[hmac,dv] Split NIST test vectors to avoid timeout

martin-velay:
- Split the NIST test vectors into shorter tests to avoid timeout issues. Linked to issue #23980.
- And move the NIST test vector checking from the sequence to the scoreboard to be more reusable (and UVM compliant). This addresses part of the issue #23288.
- Now, the seq is picking random vectors from the test vector file according to the desired configuartion. And the scoreboard is trying to find if one of the digest is matching. If yes, then the test is considered as passed and the coverage is collected. The NIST will be considred as covered if all of the vectors have been exercised.

### DIFF
--- a/hw/ip/hmac/dv/env/hmac_env_cov.sv
+++ b/hw/ip/hmac/dv/env/hmac_env_cov.sv
@@ -2,8 +2,52 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+// This covergroup requires to be wrapped into class to be replicated as much as needed and
+// be referenced in a generic way from the scoreboard
+class nist_cov_wrap;
+  // Usage of queue is supported since the set_covergroup_expression construct has been introduced
+  // in SystemVerilog LRM 1800-2012
+  int valid_vec_q[$];
+
+  covergroup nist_test_cg(string name) with function sample(int vec_i);
+    option.per_instance = 1;
+    option.name         = name;
+
+    vec_cg : coverpoint vec_i {
+      bins vector_list[] = valid_vec_q;
+    }
+  endgroup : nist_test_cg
+
+  function new(string name, string vec_type, test_vectors_pkg::test_vectors_t nist_vectors[]);
+    // Check if the nist_vectors is not empty
+    if (nist_vectors.size() == 0) begin
+      `uvm_fatal("hmac_env_cov", $sformatf("No vectors found for %0s", vec_type))
+    end
+
+    // Create a queue with only supported configurations to avoid unexpected holes
+    foreach (nist_vectors[j]) begin
+      string key_len;
+
+      // Get the key length name from the enum key_length_e
+      key_len = get_key_length_reg(nist_vectors[j].sha2_key_length).name();
+
+      // Check if the configuration is invalid (only for HMAC)
+      if ((vec_type == "SHA2_256") || (vec_type == "SHA2_384") || (vec_type == "SHA2_512") ||
+         ((key_len != "Key_None") && !(vec_type == "HMAC_SHA2_256" && key_len == "Key_1024"))) begin
+        valid_vec_q.push_back(j);
+      end else begin
+        `uvm_info("hmac_env_cov", $sformatf("Removed vector %0d from list %0s", j, name), UVM_DEBUG)
+      end
+    end
+
+    nist_test_cg = new(name);
+  endfunction : new
+endclass : nist_cov_wrap
+
 class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
   `uvm_component_utils(hmac_env_cov)
+
+  nist_cov_wrap nist_cov[string];
 
   covergroup cfg_cg with function sample (bit [TL_DW-1:0] cfg);
     hmac_en: coverpoint cfg[HmacEn] {
@@ -145,10 +189,14 @@ class hmac_env_cov extends cip_base_env_cov #(.CFG_T(hmac_env_cfg));
 
   // Standard SV/UVM methods
   extern function new(string name, uvm_component parent);
+
+  // Class specific methods
+  extern function void create_nist_cg(string vec_type, string vector_list[]);
 endclass : hmac_env_cov
 
 
 function hmac_env_cov::new(string name, uvm_component parent);
+  bit is_nist_test;
   super.new(name, parent);
   cfg_cg                      = new();
   status_cg                   = new();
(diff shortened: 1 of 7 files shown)


martin-velay:
This PR is ready to be reviewed I think, I am now playing a bit with the number of seeds and number or vector to be ran per test, but nothing else should change.

martin-velay:
Now this PR is fully ready and can be fully reviewed

hw/ip/hmac/dv/env/hmac_env_pkg.sv
```diff
@@ -211,6 +211,26 @@ package hmac_env_pkg;
...
+      assoc_array_name[i] = vector_list_path[i];
+    end
+    // Find the last slash and last dot positions in this array
+    last_slash = assoc_array_name.find_last_index with (item == byte'("/"));
+    last_dot   = assoc_array_name.find_last_index with (item == byte'("."));
+    // Finally, extract the list name based on the last slash and last dot positions
+    list_name  = vector_list_path.substr(last_slash[0]+1, last_dot[0]-1);
+    `uvm_info("hmac_env_pkg", $sformatf("NIST list_name is %0s", list_name), UVM_DEBUG)
+    return list_name;
+  endfunction : get_nist_list_name
```
antmarzam, line 231: I wonder if it makes sense to just get the `last_slash` and `last_dot` variables from the loop itself without having to convert it to an associative array of bytes, by just iterating from the vector in reverse order and breaking the loop sooner to avoid looping unnecessarily? Something like:

```
function automatic string get_nist_list_name(....)
  int last_slash=-1;
  int last_dot=-1;
  string list_name;
  for (int i = (vector_list_path.len()-1); i >= 0; i--) begin
    if (last_slash == -1 && vector_list_path[i] == "/") 
      last_slash = i;
    if (last_dot == -1 && vector_list_path[i] == ".") 
      last_dot = i;
    if (last_dot != -1 && last_slash != -1) 
      break;
  end
  // Then maybe checking that both were found, otherwise raise an error?
  if (last_dot == -1 || last_slash == -1)
    `uvm_fatal($sformatf("%m"), $sformatf("Vector format is incorrect: %0s",vector_list_path))  
```

I always try to avoid associative arrays if I can since they aren't very efficient in terms of performance, but maybe I misunderstood something in your code

hw/ip/hmac/dv/env/hmac_scoreboard.sv
```diff
@@ -906,6 +926,109 @@ function void hmac_scoreboard::predict_digest(
...
+        if (act_digest[dig_i] == exp_nist_digest[dig_i]) begin
+          matching_digest_cnt++;
+        end else begin
+          matching_digest_cnt = 0;
+          break;  // Break the for loop and jump to the next vector
+        end
+        // Sample which vectors were fully matching
+        if (cfg.en_cov && (matching_digest_cnt == nb_relevant_digest)) begin
+          cov.nist_cov[get_nist_list_name(vector_list[list_i])].nist_test_cg.sample(vec_i);
+        end
```
antmarzam, line 1013: Do you have to check the vectors were fully matching to sample coverage? 
If the vectors weren't matching that would cause a `uvm_error` and hence we won't be gathering coverage for that failing test right?

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
```diff
@@ -52,79 +58,87 @@ task hmac_test_vectors_sha_vseq::pre_start();
...
+    `uvm_info(`gfn, $sformatf("vector[%0d]: %0p", j, parsed_vectors[j]), UVM_MEDIUM)
+
+    case (parsed_vectors[j].sha2_key_length)
+      128:     key_length = Key_128;
+      256:     key_length = Key_256;
+      384:     key_length = Key_384;
+      512:     key_length = Key_512;
+      1024:    key_length = Key_1024;
+      default: key_length = Key_None;
+    endcase
```
antmarzam, line 94: Not a lot of duplication, but since you use the same in `hmac_env_cov` (lines-26-33) maybe it's worth moving this to the package into a function like: `sha_lenght2key_length` 
At the moment `key_length` here is a 6-bit vector and in the `hmac_env_cov` is a string though.

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -2,9 +2,54 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+// Require to be wrapped in a class to be replicated as much as needed and referenced easily from
+// the scoreboard
```
rswarbrick, line 6: English nit: Sorry, I'm not completely sure what this means. Is there any chance you could rephrase it?

hw/ip/hmac/dv/env/hmac_env_cov.sv
```diff
@@ -2,9 +2,54 @@
...
+    option.name         = name;
+
+    vec_cg : coverpoint vec_i {
+      bins vector_list[] = valid_vec_q;
+    }
+  endgroup : nist_test_cg
+
+  function new(string name, string vec_type, test_vectors_pkg::test_vectors_t nist_vectors[]);
+    // Create a queue with only supported configurations to avoid unexpected holes
+    foreach (nist_vectors[j]) begin
```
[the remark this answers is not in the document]
rswarbrick: Oh: is your point that we *want* to make sure we haven't made a mistake that causes this array to be empty? That does seem reasonable, I think: should it just be a check before the loop?

hw/ip/hmac/dv/env/hmac_env_pkg.sv
rswarbrick: I think I probably agree with the suggestion: it's probably simpler. And I think it has an error check that the original doesn't have. I think that e.g. `last_slash` might be an empty queue?

rswarbrick (commented):
This looks like an improvement. I think @antmarzam has some good points, but otherwise it's excellent!

hw/ip/hmac/dv/env/hmac_env_pkg.sv
martin-velay: It works, thanks for your suggestion Antonio !
And Rupert, are you suggesting to change the `last_dot` and `last_slash` declaration from integer to a queue?

hw/ip/hmac/dv/env/hmac_env_cov.sv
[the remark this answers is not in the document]
martin-velay: It sounds reasonable indeed, let me add this check

hw/ip/hmac/dv/env/hmac_scoreboard.sv
martin-velay: I need to keep this to avoid multiple useless samplings and I cannot move this out of the loop otherwise I won't have the `vec_i` value visible anymore

hw/ip/hmac/dv/env/seq_lib/hmac_test_vectors_sha_vseq.sv
martin-velay: I will extract the name from enum. It should work hopefully

hw/ip/hmac/dv/env/hmac_env_cov.sv
martin-velay: Is it better?

"This covergroup requires to be wrapped into class to be replicated as much as needed and be referenced in a generic way from the scoreboard"

antmarzam: The LRM refers to covergroups in classes as "embedded covergroups". How about something like:
"Embedded covergroup for NIST vectors. Each NIST vector is sampled separately and referenced in hmac_env_cov as part of a dynamic array addressed by a string for each vector."?

antmarzam (approved):
LGTM thank you! Before you pushed the latest changes I think CI was failing with HMAC errors though, not sure if due to this code or some other reason.

martin-velay:
Oh yes thanks, it was a constraint to be changed into "soft"

rswarbrick (approved):
This looks really good to me!

PR #26063, opened 2025-01-29
sphincsplus: accept PUB key as PEM encoded ASN.1 objects

vbendeb:
Some HSMs distribute SPX pub keys as PEM encoded ANS.1 objects. The approach so far has been to manually extract the 32 byte key material from the object. This patch adds parsing code for processing the ANS.1 key files.

The supported algorithm is slh-dsa-sha2-128s-with-sha256, its OID is defined in
https://csrc.nist.gov/projects/computer-security-objects-register/algorithm-registration#heading1725030685275_13

Tested on an opentitatn owner's firmware by passing to --spx-key first the raw SPX key and then the HSM generated PEM file, and observing the unchanged contents of the resulting image.

Signed-off-by: Vadim Bendebury <vbendeb@chromium.org>
(cherry picked from commit fb326fc736e955f3a6f2d8db44de03c430b9b9b1)

### DIFF
--- a/sw/host/sphincsplus/BUILD
+++ b/sw/host/sphincsplus/BUILD
@@ -60,6 +60,7 @@ rust_library(
     deps = [
         ":bindgen_sha2_128s_simple",
         ":bindgen_shake_128s_simple",
+        "@crate_index//:asn1",
         "@crate_index//:pem-rfc7468",
         "@crate_index//:serde",
         "@crate_index//:strum",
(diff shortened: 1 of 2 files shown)


vbendeb:
a straight cherry-pick from master

vbendeb:
I'm not sure if we want this in earlgreay_1.0.0, if we do please approve it @cfrantz @timothytrippel

cfrantz (approved):

PR #26068, opened 2025-01-30
[rules] Rename and change semantic of `extra_bazel_features`

pamaury:
The `extra_bazel_features` attribute was added #25840 as an attribute to `rv_rule` to transitively add features to the rule. However this is too general because when applied to opentitan_binary, it will change the features not only for the rule, but all its dependencies such as exec_env, tools,  etc which may have unintended effects.

This commit keeps essentially the same logic but instead of applying it to `rv_rule`, it is now only an attribute of `opentitan_binary` and `opentitan_test` and it transitions only features for the `deps`. This attribute is also renamed to `transitive_features`. The rationale for the rename is that every bazel rule has an implicit `features` argument but it only applies to the target, not its dependencies. This attribute does the same but also applies transitively.

### DIFF
--- a/rules/opentitan/cc.bzl
+++ b/rules/opentitan/cc.bzl
@@ -49,11 +49,13 @@ def ot_binary(ctx, **kwargs):
       (elf_file, map_file) File objects.
     """
     cc_toolchain = find_cc_toolchain(ctx)
+    transitive_features = [f for f in ctx.attr.transitive_features if not f.startswith("-")]
+    transitive_disabled_features = [f.removeprefix("-") for f in ctx.attr.transitive_features if f.startswith("-")]
     features = cc_common.configure_features(
         ctx = ctx,
         cc_toolchain = cc_toolchain,
-        requested_features = get_override(ctx, "features", kwargs),
-        unsupported_features = get_override(ctx, "disabled_features", kwargs),
+        requested_features = get_override(ctx, "features", kwargs) + transitive_features,
+        unsupported_features = get_override(ctx, "disabled_features", kwargs) + transitive_disabled_features,
     )
 
     compilation_contexts = [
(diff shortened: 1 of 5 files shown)


rules/opentitan/cc.bzl
```diff
@@ -49,11 +49,14 @@ def ot_binary(ctx, **kwargs):
...
     cc_toolchain = find_cc_toolchain(ctx)
+    transitive_features = [f for f in ctx.attr.transitive_features if not f.startswith("-")]
+    transitive_disabled_features = [f.removeprefix("-") for f in ctx.attr.transitive_features if f.startswith("-")]
     features = cc_common.configure_features(
         ctx = ctx,
         cc_toolchain = cc_toolchain,
-        requested_features = get_override(ctx, "features", kwargs),
-        unsupported_features = get_override(ctx, "disabled_features", kwargs),
+        # FIXME: is the use of get_override correct here? Or should it be additive?
+        requested_features = get_override(ctx, "features", kwargs) + transitive_features,
```
cfrantz, line 58: FYI: `get_overrride` will prefer the item in kwargs if it exists, else it will retrieve it from the ctx object.

Most of these functions are constructed with `get_override` so they can be naively called from rule implementations and work as expected, but, they can also be called with options in kwargs to customize their behavior when needed.

```
def _foo_impl(ctx):
  result = ot_binary(ctx)
  return make_providers_from(result)

## vs:

def something_special_impl(ctx):
  result = ot_binary(ctx,
        features = ["magical_option", "reduced_warnings"]
  )
    return make_providers_from(result)
```

sw/device/silicon_creator/rom_ext/BUILD
```diff
@@ -235,14 +235,14 @@ opentitan_binary(
...
-        "use_lld",
-    ],
     immutable_rom_ext_sections = SLOT_A_IMM_ROM_EXT_SECTIONS["main"],
     linker_script = ":ld_slot_a",
     manifest = ":manifest",
     spx_key = {"//sw/device/silicon_creator/rom/keys/fake/spx:prod_key_0_spx": "prod_key_0"},
+    transitive_features = [
+        "minsize",
+        "use_lld",
+    ],
```
cfrantz, line 245: Any issue with just enabling these globally in the rv32imc toolchain?

cfrantz (approved):

rules/opentitan/cc.bzl
pamaury: Ok

sw/device/silicon_creator/rom_ext/BUILD
pamaury: I haven't tried yet, I will create a PR to try it. I don't think there is any harm in explicitely specifying it for important targets such as rom_ext.

timothytrippel (approved):

timothytrippel:
@pamaury can you cherry-pick this to the `earlgrey_1.0.0` branch?

pamaury:
@timothytrippel #26543

PR #26069, opened 2025-01-30
[dd, prim_reg_cdc_arb] state signal width reduction

antmarzam:
Removing unused upper bit of state_q/d signals


vogelpi (approved):
LGTM, thanks @antmarzam !

For reference, this is as discussed in https://github.com/lowRISC/opentitan/pull/25705#discussion_r1935267779.

vogelpi:

This PR removes an unused state bit. There is no purpose in having that bit and it's fine to remove it.

rswarbrick (approved):

PR #26070, opened 2025-01-30
[otp_ctrl,doc] Fix some links to otp_ctrl documentation

rswarbrick:
These were broken when the block became templated. There are still a couple of hits for "ip/otp_ctrl" as an absolute path in the repository:

  - hw/syn/tools/dc/testsynth.tcl

    (I don't have a DC licence and I'm not sure how this gets run, so can't really fix it)

  - util/design/data/otp_ctrl_img.c.tpl

    This doesn't look quite right, but the generation is tied in with the Bazel SW build. I'm not entirely sure how the link should work.

@matutem: Would you mind glancing through this and checking that the renames (and explanatory comments) match how you think it should work?

@engdoreis: As someone with much more knowledge of the SW world, do you know how the `otp_ctrl_img.c.tpl` file gets used? Does it make more sense for the comment to point at the otp_ctrl template, or should this be some relative path?


matutem:
I suggest you make this change in otp_ctrl_img-c-tpl:
 // AUTOGENERATED. Do not edit this file by hand.
-// See the util/design/data hw/ip/otp_ctrl/data README for details.
+// See util/design/README.md section "OTP Preload Image Generator" for details.
 
As for the synthesis tcl script, it may only support earlgrey? If so, change the path from
../../../ip/otp_ctrl
to
../../../top_earlgrey/ip_autogen/otp_ctrl

If the script knew the topname it may be able to generate the right path.

matutem (approved):
Thanks Rupert. These look good. I have a couple suggestions about the missing paths you mention.
Also, I think it may be best to be able to refer to the html generated from the markup since it provides more anchors. This may be possible when the docs contain all tops (or do they already?). Perhaps we should add an issue to replace reference to md files to references to the generated html if and when docs are multi-top?

Please at least make the change in otp_ctrl_img.c.tpl, otherwise LGTM

rswarbrick:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/otp_ctrl/rtl/otp_ctrl_scrmbl.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/otp_ctrl/rtl/otp_ctrl_token_const.sv

The changes to these files lie in comments in each case, so cause no risk to the design.

rswarbrick:
@matutem: I think the docs already contain all tops. I'm not quite sure where this replacement would be relevant though. Can you point to an example?

matutem:
@rswarbrick my relevant comment is https://github.com/lowRISC/opentitan/pull/26070#issuecomment-2626891619

Specifically, I make a suggestion for a change in otp_ctrl_img.c.tpl, and I would make the synthesis tcl file either top_earlgrey-specific, or consult if the flow can take a topname parameter in case it actually makes sense. Another option is to move that script to hw/top_earlgrey/syn.

rswarbrick:
Good point. I'm being a bit lazy here, so I'm just adding a "caveat emptor" message. Hopefully, this covers things reasonably.

alees24:
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/otp_ctrl/rtl/otp_ctrl_scrmbl.sv
CHANGE AUTHORIZED: hw/top_earlgrey/ip_autogen/otp_ctrl/rtl/otp_ctrl_token_const.sv

PR #26071, opened 2025-01-30
[rom_ctrl,doc] Updates and tidy-ups

rswarbrick:
These commits are trying to tidy up the block documentation so that things are more fully specified. It might hopefully feed into a verification plan for the block. (But documenting what on earth it's supposed to do seems like it's sensible either way!)


hw/ip/rom_ctrl/README.md
```diff
@@ -8,15 +8,25 @@
 
 # Overview
 
-This document describes the ROM controller (`rom_ctrl`).
-This module attaches as a peripheral to the system bus, and thus follows the [Comportability Specification](../../../doc/contributing/hw/comportability/README.md).
+The ROM controller (`rom_ctrl`) is the connection between the chip and its ROM.
+It has three main tasks:
+- Pass read requests to the ROM and respond with the memory contents.
+- Generate a checksum of the ROM contents and feed its value to keymgr to be included in the device identity.
```
martin-velay, line 14: It'd make sense to add the [link to the keymgr](https://opentitan.org/book/hw/ip/keymgr/index.html) block at least one time. And for the next references as you are not saying the full name, maybe use this form `keymgr`

hw/ip/rom_ctrl/README.md
```diff
@@ -8,15 +8,25 @@
...
+- Inform pwrmgr once the checksum has been computed and looks reasonable, to allow the rest of the chip startup.
```
martin-velay, line 15: Same for here [pwrmgr](https://opentitan.org/book/hw/top_earlgrey/ip_autogen/pwrmgr/index.html)

hw/ip/rom_ctrl/README.md
```diff
@@ -8,15 +8,25 @@
...
 
-The ROM controller interfaces between the system bus and the ROM.
-This ROM has scrambled contents (scrambled with a fixed key, derived from a global constant).
-The controller is responsible for descrambling these contents on memory fetches.
+In order to fulfil the first task, the ROM controller attaches as a peripheral to the system bus.
+Because it does so, it follows the [Comportability Specification](../../../doc/contributing/hw/comportability/README.md).
```
martin-velay, line 18: Nit: it's maybe my English here, but this sentence sounds weird to me

hw/ip/rom_ctrl/README.md
```diff
@@ -8,15 +8,25 @@
...
+The contents of the ROM are also scrambled (with a fixed key, derived from a global constant).
```
martin-velay, line 19: Do we have somewhere an explanation on how the data are scrambled or a name of an algorithm mentioned?

hw/ip/rom_ctrl/doc/interfaces.md
```diff
@@ -53,10 +53,23 @@ Referring to the [Comportable guideline for peripheral device functionality](htt
...
-`RndCnstRomKey`             | (see RTL)             | (see RTL)    | Compile-time random default constant for scrambling key (used in `prim_prince` block).
-`RndCnstRomNonce`           | (see RTL)             | (see RTL)    | Compile-time random default constant for scrambling nonce (used in `prim_prince` block and the two S&P blocks).
+The top-level `rom_ctrl` module has several parameters, although some of them are only relevant for verification.
+Those parameters are described separately below.
+The parameters that have an effect on the generated block are:
+
+Parameter         | Default (Max)                 | Top Earlgrey | Description
+------------------|-------------------------------|--------------|---------------
+`AlertAsyncOn`    | True                          | True         | This is passed to the single `prim_alert_sender` instance and causes it to generate synchronization logic to support alert rx and tx being on different clocks.
+`RndCnstRomNonce` | `RND_CNST_SCR_NONCE` (define) | (see RTL)    | Compile-time random default constant for scrambling nonce (used in `prim_prince` block and the S&P block).
```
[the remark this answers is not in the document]
rswarbrick: Good point. It gets passed in as a parameter (with value `0123456789...`!)  in `tb.sv`. Probably the DV documentation is the right place to describe this properly, and I haven't written that yet!

But I'm about to force-push a tweak where we at least give the sizes of the values.

hw/ip/rom_ctrl/README.md
rswarbrick: Well... they aren't! They were before we dropped one of the stages. I've rewritten that text to be more correct and linked out to the right place this should be explained.

Honestly, it's not explained all that fully there either. Do you think there's more that we should be explaining with the new version?

rswarbrick:
Force push doesn't add anything (except for a couple of missing backticks) but it rebases onto the latest master.

hw/ip/rom_ctrl/README.md
rswarbrick: I agree. Both now done, I think.

hw/ip/rom_ctrl/README.md
rswarbrick: I agree. Annoyingly, this is top-specific. I've tried to write something reasonable, which ends up looking like:
image

martin-velay:
LGTM, thanks Rupert!

jwnrt (approved):
Looks great, thank you

PR #26072, opened 2025-01-30
[prim,rtl] Rewrite a case statement to avoid unreachable case

rswarbrick:
Instead of the default case being unreachable (unless state_q is X), rewrite things so that the last item is caught by the default case.

This is clearly safe at the moment (there are just two possible states!) but this adds the GoodState_A assertion to make sure no future RTL change can make this unsafe.


rswarbrick:
Note that this interacts with the coverage exclusions in #25705. If that PR lands before this one, we will need to amend this PR to drop some exclusions from the `aon_timer` manual exclusion list.

@antmarzam FYI (but I'm happy to do the leg work if it needs doing)

hw/ip/prim/rtl/prim_reg_cdc_arb.sv
```diff
@@ -219,19 +219,21 @@ module prim_reg_cdc_arb #(
...
-          state_d = StIdle;
-        end
       endcase // unique case (state_q)
     end // always_comb
 
+    // The case statement above uses default for StWait to avoid a coverage hole. This is safe, so
+    // long as the default case only runs when state_q is StWait. This assertion checks that is
+    // true.
+    `ASSERT(GoodState_A, state_q inside {StIdle, StWait})
+
```
antmarzam, line 236: CI is failing because you have to specify the clock in the assertion as the macro assumes `clk_i` 
I think you want to use: `clk_dst_i` and `negedge rst_dst_ni`

antmarzam:
> Note that this interacts with the coverage exclusions in #25705. If that PR lands before this one, we will need to amend this PR to drop some exclusions from the `aon_timer` manual exclusion list.
> 
> @antmarzam FYI (but I'm happy to do the leg work if it needs doing)

No worries, since you pushed this PR I'll clean up my other PR so we can get it merged on master!

hw/ip/prim/rtl/prim_reg_cdc_arb.sv
```diff
@@ -219,19 +219,21 @@ module prim_reg_cdc_arb #(
...
+        default: begin
           dst_hold_req = 1'b1;
           if (dst_update_ack) begin
             state_d = StIdle;
           end
         end
-
-        default: begin
-          state_d = StIdle;
-        end
```
a-will, line 231: Don't we keep these in case synthesis takes liberties with the state encoding? (e.g. converting to one-hot for some reason)

Maybe it's fine as long as there _is_ a default, but the examples in the style guide all show "recovery" circuits.

@vogelpi @andreaskurth to help me in case my memory / understanding has gone rotten. ;)

hw/ip/prim/rtl/prim_reg_cdc_arb.sv
rswarbrick: D'oh, thanks!

hw/ip/prim/rtl/prim_reg_cdc_arb.sv
rswarbrick: I *think* that we just need some default. The previous code was something like "If the state is silly then do X" and I propose change it to "If the state is silly, behave as if it is the following sensible value".

Clearly a bad idea if this is some sparsely encoded FSM in a security module! Here, we have exactly two states, represented by a single bit...

vogelpi: Can't we just put an empty default here, i.e., `default:;`. IMHO this would be more readable. Does this also create a coverage issue?

rswarbrick: Hmm. Good question: I'm not sure. I'll try it out.

rswarbrick: Sadly, it doesn't solve the problem. I think the point is that this is "branch coverage of the cond statement", so the question is whether we jump to the default case (instead of whether we run every line in that case).

I did the obvious rewrite:
image
and here are the results:
image

andreaskurth: The `default` case is functionally unreachable, so how about we add an assertion to the RTL to capture this, and then waive it from FSM branch coverage?

andreaskurth: On a second thought, shouldn't UNR take care of this exclusion?

antmarzam: I think UNR would automatically exclude this case. Although we don't have VCS UNR licenses. 

We do have a UNR waiver file from august: https://github.com/lowRISC/opentitan/blob/master/hw/ip/aon_timer/dv/cov/aon_timer_unr_excl.el

I think the reason why UNR didn't consider this case was unreachable may be that at the time UNR was run, the state signal was a 2-bit signal which perhaps made the tool believe the default state could be reached? 
state signal was simplified to 1-bit only [here](https://github.com/lowRISC/opentitan/commit/2f12c79f333484152d909d376f257ae758c7e175#diff-1769f50908f78ab4d29692c2439b87712c9e380d984771d6e7a7d90605dbd819L90)

rswarbrick: Thanks for looking into this carefully. Part of my motivation for avoiding needing UNR here is that this is a primitive. It would be nicer not to depend on UNR at each instantiation site.

andreaskurth: Ah, I see. And a waiver would also be cumbersome because it would have to be written for every instantiation, right?

andreaskurth: Since this FSM has just two states, how about we rewrite to make it simpler and remove the default case, along the lines of:
```sv
logic idle_d, idle_q;

always_comb begin
  idle_d = idle_q;
  // ... other default assignments ...
  if (idle_q) begin
    // ...
  end else begin // not idle / waiting
    // ...
  end
end
```

rswarbrick:
Oops. The second commit here was wrong: I reasoned that "because it was looking at something updated combinatorially, it should be based on `idle_d` instead of `idle_q`". But that's wrong! It was looking at something that is kept in sync with `idle_q`.

Hopefully the revised version is correct.

andreaskurth (approved):
LGTM, thanks @rswarbrick. I'm convinced this structural code change doesn't change the behavior, and if it removes a `default` that is problematic for DV, it's valuable.

andreaskurth:

Code refactoring without behavioral changes

PR #26075, opened 2025-01-30
[reggen] Render the number of registers

Razer6:
This PR renders a constant for the number of registers per interface. It's not yet used though in the individual IPs that use them. I want to wait until the naming is approved.

PR #26076, opened 2025-01-30
[rtl,racl] Add an automatic number of registers count and use it

rswarbrick:


davidschrammel:
@rswarbrick There appears to be some duplicated effort in #26075 .
The only difference seems to be that the other PR generates the number of registers per interface.
E.g., https://github.com/lowRISC/opentitan/pull/26075/files#diff-fef8df2479863db25308f6588ede458a42aea472fc246a7d768d965bc0fbf706R33

rswarbrick:
I'm pretty convinced that the "competing solution" from @Razer6 in #26075 is rather better. Let's use that :-)

